PR #7673, opened 2022-09-13
test: make exit code depends on checking test plan

ligurio:
TAP tests could be running by test-run.py, but it is often convenient to run these tests by tarantool only: tarantool test/app-tap/yaml.test.lua. Most TAP tests returns non-zero exit code when TAP asserts are failed. But exit code is not changed when TAP plan is bad (number of planned testcases is not equal to executed). Proposed patch adds test:check() to TAP tests so tests will return non-zero exit code when plan is bad.

NO_CHANGELOG=it is not a user-visible change
NO_DOC=tests


test/app-tap/cfg.test.lua
```diff
@@ -39,5 +39,4 @@ box.cfg{read_only=true}
 while f_ro:status() ~= "dead" do fiber.sleep(0.01) end
 test:is(f_ro:status(), "dead", "entered read-only mode")
 
-test:check()
-os.exit(0)
+os.exit(test:check() and 0 or 1)
```
tsafin, line 42: I'd rather introduce something common (something like `test:exit()`) which would be both:
- calling `test:check()` if it was not called yet;
- and exit properly via `os.exit`

Instead of those multiple changes in each relevant file. Though if we consider them pretty much ideomatic then it's ok for me.

tsafin (approved):
IMVHO, it's looking too mouthful, and I'd rather introduce some short-cut (see comment). But it's ok even in the current iteration.

ImeevMA (approved):
Hi! Thank you for the patch! LGTM.

I see that most tests now return `1` on error, but some of them return `-1`. Not sure if this matters, as both values are different from 0.

PR #7674, opened 2022-09-13
Make unit testing depending on source changes

ligurio:
Commit 9adedc ("test: add new `make` test targets") introduced new
    targets for running test-run.py with unit tests. However, this target
    doesn't depend on changes in tested libraries and changes in unit tests.
    Proposed patch introduces a function that creates a build targets for
    unit tests and fixes the described problem with dependencies.
    
    Additionally, patch added missed dependence for popen and popen-child.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -35,6 +35,8 @@ if(POLICY CMP0037)
     endif()
 endif(POLICY CMP0037)
 
+set(UNIT_TEST_TARGETS "")
+
 add_subdirectory(app)
 add_subdirectory(app-tap)
 add_subdirectory(box)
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -79,28 +96,18 @@ add_custom_target(test-func-force
...
-add_subdirectory(box-luatest)
-add_subdirectory(engine-luatest)
-add_subdirectory(sql-tap)
-add_subdirectory(sql-luatest)
-if(ENABLE_FUZZER)
-    add_subdirectory(fuzz)
-endif()
-add_subdirectory(unit)
 foreach(TEST_SUITE ${EXTRA_TEST_SUITES})
     add_subdirectory(
```
ylobankov, line 112: Shouldn't we move it as well?

ylobankov (approved):
LGTM, thanks!

PR #7686, opened 2022-09-14
memtx: fix TREE index `get` check for part count

### LINKED ISSUE #7685: memtx nullable TREE index `get` transaction management phantom reads
**Steps to reproduce**

```lua

local ffi = require('ffi')
local json = require('json')
local log = require('log')
local txn_proxy = require('txn_proxy')

ffi.cdef('bool flag;')

local s = box.schema.space.create('s')
s:create_index('pk')
s:create_index('sk', {parts = {{2, 'uint', is_nullable = true}}})

local tx = txn_proxy:new()
tx:begin()

ffi.C.flag = true
tx('box.space.s.index[1]:get{0}')
tx('box.space.s:replace{1, 1}')

s:replace{0, 0}
log.info('%s', json.encode(tx('box.space.s.index[1]:get{0}')))

os.exit()
```

**Actual behavior**

```console
[[0,0]]
```

**Expected behavior**

```console
""
```

CuriousGeorgiy:
**memtx: fix TREE index `get` check for part count**
If TREE index `get` result is empty, the key part count is incorrectly
compared to the tree's `cmp_def->part_count`, though it should be compared
with `cmp_def->unique_part_count`. But we can actually assume that by the
time we get to the index's `get` method the part count is equal to the
unique part count (partial keys are rejected and `get` is not
supported for non-unique indexes): change check to correct assertion.

Closes #7685

**memtx: refactor `index_def_new`**

Since `key_def_merge` sets the merged key definition's unique part count
equal to the new part count, the extra assignment in case the index is not
unique is redundant: remove it.

### DIFF
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -925,9 +925,9 @@ memtx_tree_index_get_internal(struct index *base, const char *key,
 		memtx_tree_find(&index->tree, &key_data);
 	if (res == NULL) {
 		*result = NULL;
-		if (part_count == cmp_def->part_count)
+		assert(part_count == cmp_def->unique_part_count);
 /********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
-			memtx_tx_track_point(txn, space, base, key);
+		memtx_tx_track_point(txn, space, base, key);
 /*********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND END**********/
 		return 0;
 	}
--- a/test/box-luatest/gh_7685_memtx_tree_idx_get_with_nullable_field_phantom_read_test.lua
+++ b/test/box-luatest/gh_7685_memtx_tree_idx_get_with_nullable_field_phantom_read_test.lua
@@ -1,0 +1,66 @@
+local server = require('test.luatest_helpers.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new{
+        alias   = 'dflt',
+        box_cfg = {memtx_use_mvcc_engine = true}
+    }
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.before_each(function(cg)
+    cg.server:exec(function()
+        local s = box.schema.space.create('s')
+        s:create_index('pk')
+        s:create_index('sk', {parts = {{2, 'uint', is_nullable = true}}})
+    end)
+end)
+
+g.after_each(function(cg)
+    cg.server:exec(function()
+        box.space.s:drop()
+    end)
+end)
+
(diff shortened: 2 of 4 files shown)


src/box/memtx_tree.cc
```diff
@@ -921,8 +921,8 @@ memtx_tree_index_get_internal(struct index *base, const char *key,
 		memtx_tree_find(&index->tree, &key_data);
 	if (res == NULL) {
 		*result = NULL;
-		if (part_count == cmp_def->part_count)
-			memtx_tx_track_point(txn, space, base, key);
+		assert(part_count == cmp_def->unique_part_count);
+		memtx_tx_track_point(txn, space, base, key);
```
drewdzzz, line 925: I don't get it. Doesn't `cmp_def` have the same value in `part_count` and `unique_part_count` fields?

test/box-luatest/gh_7685_memtx_tree_idx_get_with_nullable_field_phantom_read_test.lua
```diff
@@ -0,0 +1,56 @@
+local server = require('test.luatest_helpers.server')
```
drewdzzz, line 1: Test is passed locally without applying the patch.

src/box/memtx_tree.cc
CuriousGeorgiy: Please read this comment:
https://github.com/tarantool/tarantool/blob/8ee0e434610189f3e4bbca0264de60aa79f0696e/src/box/key_def.h#L191-L198

Please look at this snippet:
https://github.com/tarantool/tarantool/blob/8ee0e434610189f3e4bbca0264de60aa79f0696e/src/box/index_def.c#L100-L105

drewdzzz: As for me, if key_def is unique, it shouldn't be merged with pk_def, and cmp_def will have the same value in part_count and unique_part_count fields. I will ask the team about it.

test/box-luatest/gh_7685_memtx_tree_idx_get_with_nullable_field_phantom_read_test.lua
CuriousGeorgiy: It's a surpise for me that `get` is translated to `select` request over **IPROTO**, that's why the test does not work as expected: rewrote it using `txn_proxy`.

drewdzzz: You squashed changes with another commit.

CuriousGeorgiy: Sorry for the hush, fixed.

alyapunov (approved):

drewdzzz (approved):

Issue #6986, opened 2022-03-31
Introduce new FOREIGN KEY and CHECK constraints to SQL


ImeevMA:
These two constrains were added to BOX in the https://github.com/tarantool/tarantool/issues/6436 issue. We must now rework the SQL FOREIGN KEY and CHECK constraint using the new implementations.

PR #7651, opened 2022-09-04
sql: enhance drop constraint

ImeevMA:
This patch-set allows to use ALTER TABLE DROP CONSTRAINT to drop BOX field and tuple constraints.

Part of #6986

PR #7688, opened 2022-09-15
box: introduce interface for tuple constraints

ImeevMA:
This patch introduces drop_constraint(), drop_foreign_key(), create_constraint() and create_foreign_key() methods for space in Lua.

Needed for #6986

### DIFF
--- a/changelogs/unreleased/gh-6986-introduce-constraint-interface.md
+++ b/changelogs/unreleased/gh-6986-introduce-constraint-interface.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* The space object in Lua now has an interface to create or drop foreign keys
+  and constraints (gh-6986).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -89,6 +89,7 @@
 #include "mp_uuid.h"
 #include "flightrec.h"
 #include "wal_ext.h"
+#include "tuple_constraint_def.h"
 
 static char status[64] = "unknown";
 
@@ -2312,6 +2313,45 @@ box_set_txn_isolation(void)
 /* }}} configuration bindings */
 
 /**
+ * Encode variables from the variable argument list according to format string
+ * into a buffer allocated on region.
+ */
+static const char *
+mp_vformat_on_region(struct region *region, size_t *size, const char *format,
+		     va_list src)
+{
+	va_list ap;
+	va_copy(ap, src);
+	size_t buf_size = mp_vformat(NULL, 0, format, ap);
+	char *buf = (char *)region_alloc(region, buf_size);
+	va_end(ap);
+	if (buf == NULL) {
+		diag_set(OutOfMemory, buf_size, "region_alloc", "buf");
+		return buf;
+	}
+	va_copy(ap, src);
+	mp_vformat(buf, buf_size, format, ap);
+	va_end(ap);
+	*size = buf_size;
+	return buf;
+}
+
+/**
+ * Encode a sequence of values according to format string into a buffer
+ * allocated on region.
+ */
+static const char *
+mp_format_on_region(struct region *region, size_t *size, const char *format,
+		    ...)
+{
+	va_list ap;
+	va_start(ap, format);
+	const char *res = mp_vformat_on_region(region, size, format, ap);
+	va_end(ap);
+	return res;
+}
+
+/**
  * Execute a request against a given space id with
  * a variable-argument tuple described in format.
  *
@@ -2334,17 +2374,13 @@ boxk(int type, uint32_t space_id, const char *format, ...)
 	request.type = type;
 	request.space_id = space_id;
 	va_start(ap, format);
-	size_t buf_size = mp_vformat(NULL, 0, format, ap);
-	char *buf = (char *)region_alloc(&fiber()->gc, buf_size);
+	struct region *region = &fiber()->gc;
+	size_t size = 0;
+	const char *data = mp_vformat_on_region(region, &size, format, ap);
 	va_end(ap);
-	if (buf == NULL)
+	if (data == NULL)
 		return -1;
-	va_start(ap, format);
-	if (mp_vformat(buf, buf_size, format, ap) != buf_size)
-		assert(0);
-	va_end(ap);
-	const char *data = buf;
-	const char *data_end = buf + buf_size;
+	const char *data_end = data + size;
 	switch (type) {
 	case IPROTO_INSERT:
 	case IPROTO_REPLACE:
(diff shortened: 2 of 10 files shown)


src/box/lua/schema.lua
```diff
@@ -2661,6 +2708,88 @@ space_mt.run_triggers = function(space, yesno)
...
+    if builtin.box_field_constraint_drop(space.id, name, #name,
+                                         fieldno) ~= 0 then
+        box.error()
+    end
+end
+space_mt.create_constraint = function(space, name, func, field)
+    check_space_arg(space, 'create_constraint')
+    check_space_exists(space)
+    check_param(name, 'name', 'string')
+    local constr;
```
Gerold103, line 2731: Don't need `;`.

src/box/lua/schema.lua
```diff
@@ -2661,6 +2708,88 @@ space_mt.run_triggers = function(space, yesno)
...
+    if (func == nil) then
```
Gerold103, line 2732: Don't need `()`.

src/box/lua/schema.lua
```diff
@@ -2661,6 +2708,88 @@ space_mt.run_triggers = function(space, yesno)
...
+        constr = name
+    else
+        constr = {[name] = func}
+    end
+    constr = normalize_constraint(constr, "")
```
Gerold103, line 2737: I see other usages of this function employ `''` (single quotes). I suggest to do the same here and in `normalize_foreign_key()`.

src/box/schema.cc
```diff
@@ -487,3 +489,307 @@ schema_find_name(enum schema_object_type type, uint32_t object_id)
 	return NULL;
 }
 
+/**
+ * Drop the constraint or foreign key. If fieldno is equal to UINT32_MAX then
+ * the tuple constraint or tuple foreign key is dropped, otherwise the field
+ * constraint or field foreign key is dropped.
+ */
+int
```
Gerold103, line 497: Can be `static`. Doesn't seem to be used outside of this file.

src/box/schema.cc
```diff
@@ -487,3 +489,307 @@ schema_find_name(enum schema_object_type type, uint32_t object_id)
...
+	key_end = mp_encode_uint(key_end, space_id);
+	struct tuple *tuple;
+	if (box_index_get(BOX_SPACE_ID, 0, key, key_end, &tuple) != 0)
+		return -1;
+	assert(tuple != NULL);
+	const char *data = tuple_data(tuple);
+	const char *data_end = data;
+	mp_next(&data_end);
+
+	const char *str;
```
Gerold103, line 512: Why is it called `str` if it is not a string?

src/box/schema.cc
```diff
@@ -487,3 +489,307 @@ schema_find_name(enum schema_object_type type, uint32_t object_id)
...
+		constraint = str;
+	}
+	if (constraint_id == constraint_count) {
+		int errcode = type == CONSTR_FKEY ? ER_NO_SUCH_FOREIGN_KEY :
+			      ER_NO_SUCH_CONSTRAINT;
+		diag_set(ClientError, errcode, name,
+			 space_name(space_by_id(space_id)));
+		return -1;
+	}
+	const char *post = constraint;
```
Gerold103, line 567: Instead of all those tricky decodings could try to re-use `space_def_new_from_tuple()`. Although you would need to implement `space_def_to_tuple()` or `space_def_to_msgpack()`. Then you would decode a `space_def`, do the needed changes without the msgpack tricks, encode it back, and use `box_replace()` like now. Up to you if you want to try.

src/box/schema.cc
```diff
@@ -487,3 +489,307 @@ schema_find_name(enum schema_object_type type, uint32_t object_id)
 	return NULL;
 }
 
+/**
+ * Drop the constraint or foreign key. If fieldno is equal to UINT32_MAX then
+ * the tuple constraint or tuple foreign key is dropped, otherwise the field
+ * constraint or field foreign key is dropped.
+ */
+int
+constraint_drop(uint32_t space_id, const char *name, uint32_t name_len,
```
Gerold103, line 498: Having this new API in schema.h looks odd. Mostly because this file had nothing to do with the storage layer before (and this is why it didn't need box.h). All was in box.h/.cc like `box_insert()`, `box_sequence_next()`, `boxk()` and all the similar functions. I would consider moving it all there. Up to you.

src/box/schema.cc
```diff
@@ -487,3 +489,307 @@ schema_find_name(enum schema_object_type type, uint32_t object_id)
...
+		new_data_end = mp_encode_str0(new_data_end, constraint_type);
+		new_data_end = mp_encode_map(new_data_end, constraint_count);
+		uint32_t constraint_size = constraint - constraint_begin;
+		memcpy(new_data_end, constraint_begin, constraint_size);
+		new_data_end += constraint_size;
+	}
+	uint32_t post_size = data_end - post;
+	memcpy(new_data_end, post, post_size);
+	new_data_end += post_size;
+	int rc = box_replace(BOX_SPACE_ID, new_data, new_data_end, NULL);
```
Gerold103, line 594: You might think about `box_update()` with `#` operation. It supports JSON paths, so you can build msgpack operations like `{{'#', 'path.to.constraint.def'}}`. In case it won't fail on a not existing field (I don't remember if it does), then on success you can compare old and new tuples using `memcmp`. If they are equal, it means the update didn't do anything and the constraint wasn't found. `box_update()` could simplify this code a lot.

Even to build the operations msgpack you could use `mp_vformat()` so as not to encode anything manually.

Same for constraints creation.

Gerold103 (commented):
Thanks for the patch!

extra/exports
```diff
@@ -54,6 +54,10 @@ box_error_last
 box_error_message
 box_error_set
 box_error_type
+box_field_constraint_create
+box_field_constraint_drop
+box_field_foreign_key_drop
+box_field_foreign_key_create
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! I have dropped these functions from `exports` as you suggested below.

src/box/lua/schema.lua
```diff
@@ -64,6 +64,33 @@ ffi.cdef[[
     ssize_t
     box_index_count(uint32_t space_id, uint32_t index_id, int type,
                     const char *key, const char *key_end);
+    int
+    box_tuple_constraint_create(uint32_t space_id, const char *name,
+                                uint32_t name_len, uint32_t func_id);
+    int
+    box_tuple_constraint_drop(uint32_t space_id, const char *name,
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed!

src/box/lua/schema.lua
```diff
@@ -2514,6 +2541,26 @@ vinyl_index_mt.__ipairs = vinyl_index_mt.pairs
 memtx_index_mt.__pairs = memtx_index_mt.pairs
 memtx_index_mt.__ipairs = memtx_index_mt.pairs
 
+local function get_fieldno(space, field)
+    check_space_exists(space)
```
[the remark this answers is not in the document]
ImeevMA: Fixed.

src/box/lua/schema.lua
ImeevMA: Fixed.

src/box/lua/schema.lua
ImeevMA: Fixed.

src/box/lua/schema.lua
ImeevMA: Fixed.

src/box/schema.cc
ImeevMA: Fixed.

src/box/schema.cc
ImeevMA: I used `str` as a generic name here, but now I think `buf` or `pos` would be better. In any case, I dropped this part.

src/box/schema.cc
ImeevMA: I used your suggestion about `box_update()`.

src/box/schema.cc
ImeevMA: Thanks, moved.

src/box/schema.cc
ImeevMA: Thank you for the suggestion! The use of JSON paths helped a lot here - now the code is much simpler and a little shorter.

src/box/box.cc
```diff
@@ -4296,3 +4297,231 @@ box_read_ffi_enable(void)
...
+	}
+	struct region *region = &fiber()->gc;
+	uint32_t used = region_used(region);
+	size_t size = 64 + name_len;
+	char *raw = (char *)region_alloc(region, size);
+	if (raw == NULL) {
+		diag_set(OutOfMemory, size, "region_alloc", "raw");
+		return -1;
+	}
+	char *pos = mp_encode_array(raw, 1);
```
Gerold103, line 4346: Just in case you missed my first comment, you can also use `mp_vformat()`, it might be significantly simpler than manual encoding and it is not like we are running for perf here. Up to you.

Gerold103 (commented):
Hi! Thanks for the fixes! Now there is a new problem though, about `box_update()`, which I didn't think of before:
```Lua
box.cfg{}
body = "function(x) return true end"
box.schema.func.create('ck1', {is_deterministic = true, body = body})
func_id = box.func.ck1.id
fk0 = {one = {field = {a = 'a'}}, two = {field = {b = 'b'}}}
ck0 = {three = 'ck1', four = 'ck1'}
fk1 = {five = {field = 'a'}, six = {field = 'b'}}
ck1 = {seven = 'ck1', eight = 'ck1'}
fmt = {{'a', 'integer'}, {'b', 'integer'}}
fmt[1].constraint = ck1
fmt[2].foreign_key = fk1
def = {format = fmt, foreign_key = fk0, constraint = ck0}
s = box.schema.space.create('a', def)
t = box.space._space:get{s.id}
t = t:totable()
t[6].foreign_key = setmetatable({}, {__serialize = 'map'})
t[6].constraint = t[6].foreign_key
box.space._space:replace(t)
box.space.a:create_constraint('four', 'ck1')
```
In your code when you create a constraint, you assume that the corresponding map does not exist in case there were no constraints before. But it can exist as an empty map. In that case `!` operation would fail:
```
---
- error: 'Field ''flags.constraint'' UPDATE error: the key exists already'
...
```
Not sure what would be the best way to fix this. We certainly shouldn't make a `box_update()` like now and redo it if it fails as a duplicate. You could try to get the tuple first, then check if it has the needed map using `tuple_field_raw_by_full_path()` (so as not to bother with decoding things manually, or you can decode them manually too if want). If it does, then you insert into it, or otherwise insert the map itself too.

After fixing it I suggest to add tests on this. Deletion of constraints looks fine right now, the problem is just with their creation AFAICS.

src/box/box.cc
ImeevMA: Thank you for the review! I fixed the above bug. Also I refactored the code using `mp_vformat()`. I really missed that part of the comment.

src/box/box.cc
```diff
@@ -4296,3 +4297,264 @@ box_read_ffi_enable(void)
 	if (--box_read_ffi_disable_count == 0)
 		box_read_ffi_is_disabled = false;
 }
+
+/**
+ * Encode a sequence of values according to format string into a buffer
+ * allocated on region.
+ */
+static const char *
+mp_format_on_region(struct region *region, const char *format, ...)
```
Gerold103, line 4306: The same is done in `boxk()`. I suggest to use it there too. Should be easy given that it is in the same file (box.cc).

src/box/box.cc
```diff
@@ -4296,3 +4297,264 @@ box_read_ffi_enable(void)
...
+		return -1;
+	}
+
+	char key[16];
+	char *key_end = key + mp_format(key, 16, "[%u]", space_id);
+	struct region *region = &fiber()->gc;
+	uint32_t used = region_used(region);
+	const char *type_str = tuple_constraint_type_strs[type];
+	const char *path = fieldno != UINT32_MAX ?
+			   tt_sprintf("format[%d].%s", fieldno + 1, type_str) :
```
Gerold103, line 4367: `%d` -> `%u`. Same in `constraint_create()`.

src/box/box.cc
```diff
@@ -4296,3 +4297,264 @@ box_read_ffi_enable(void)
...
+						     TUPLE_INDEX_BASE);
+		is_empty = field == NULL;
+	}
+
+	struct region *region = &fiber()->gc;
+	uint32_t used = region_used(region);
+	const char *val;
+	if (type == CONSTR_FUNC) {
+		val = mp_format_on_region(region, "%u", value->func_id);
+	} else if (fieldno != UINT32_MAX) {
```
Gerold103, line 4488: I suggest to assert here that the type is either FUNC or FKEY.

src/box/box.cc
```diff
@@ -4296,3 +4297,264 @@ box_read_ffi_enable(void)
...
+	}
+	const char *ops;
+	if (is_empty) {
+		ops = mp_format_on_region(region, "[[%s%s{%.*s%p}]]", "!", path,
+					  name_len, name, val);
+	} else {
+		uint32_t size = strlen(path) + name_len + 2;
+		char *buf = (char *)region_alloc(region, size);
+		if (buf == NULL) {
+			diag_set(OutOfMemory, size, "region_alloc", "buf");
```
Gerold103, line 4505: Need to truncate the region here. Below there were done some `mp_format_on_region()` calls.

Gerold103 (changes requested):
Thanks for the fixes! Some last comments.

src/box/box.cc
ImeevMA: Thanks for the review! I tried to use `mp_format_on_region()` but it's quite problematic as we can't call one variadic function from another. We can use `ap_list` instead of `...` for `mp_format_on_region()`, but we have to create a new function that creates `ap_list` for `constraint_create()` and `constraint_drop()`. Not sure if this is needed.

I added size as a new output argument to mp_format_on_region().

src/box/box.cc
ImeevMA: Fixed.

src/box/box.cc
ImeevMA: Thanks, added.

src/box/box.cc
ImeevMA: Thanks, fixed.

src/box/box.cc
Gerold103: Try this:

```Diff
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -2312,6 +2312,41 @@ box_set_txn_isolation(void)
 
 /* }}} configuration bindings */
 
+/**
+ * Encode a sequence of values according to format string into a buffer
+ * allocated on region.
+ */
+static const char *
+mp_vformat_on_region(struct region *region, size_t *size, const char *format,
+		     va_list src)
+{
+	va_list ap;
+	va_copy(ap, src);
+	size_t buf_size = mp_vformat(NULL, 0, format, ap);
+	char *buf = (char *)region_alloc(region, buf_size);
+	va_end(ap);
+	if (buf == NULL) {
+		diag_set(OutOfMemory, buf_size, "region_alloc", "buf");
+		return buf;
+	}
+	va_copy(ap, src);
+	mp_vformat(buf, buf_size, format, ap);
+	va_end(ap);
+	*size = buf_size;
+	return buf;
+}
+
+static const char *
+mp_format_on_region(struct region *region, size_t *size, const char *format,
+		    ...)
+{
+	va_list ap;
+	va_start(ap, format);
+	const char *res = mp_vformat_on_region(region, size, format, ap);
+	va_end(ap);
+	return res;
+}
+
 /**
  * Execute a request against a given space id with
  * a variable-argument tuple described in format.
@@ -2335,17 +2370,13 @@ boxk(int type, uint32_t space_id, const char *format, ...)
 	request.type = type;
 	request.space_id = space_id;
 	va_start(ap, format);
-	size_t buf_size = mp_vformat(NULL, 0, format, ap);
-	char *buf = (char *)region_alloc(&fiber()->gc, buf_size);
+	size_t size = 0;
+	const char *data = mp_vformat_on_region(region, &size, format, ap);
 	va_end(ap);
-	if (buf == NULL)
+	if (data == NULL)
 		return -1;
-	va_start(ap, format);
-	if (mp_vformat(buf, buf_size, format, ap) != buf_size)
-		assert(0);
-	va_end(ap);
-	const char *data = buf;
-	const char *data_end = buf + buf_size;
+	const char *data_end = data + size;
 	switch (type) {
 	case IPROTO_INSERT:
 	case IPROTO_REPLACE:
@@ -4298,30 +4329,6 @@ box_read_ffi_enable(void)
 		box_read_ffi_is_disabled = false;
 }
 
-/**
- * Encode a sequence of values according to format string into a buffer
- * allocated on region.
- */
-static const char *
-mp_format_on_region(struct region *region, uint32_t *size, const char *format,
-		    ...)
-{
-	va_list ap;
-	va_start(ap, format);
-	size_t buf_size = mp_vformat(NULL, 0, format, ap);
-	char *buf = (char *)region_alloc(region, buf_size);
-	va_end(ap);
-	if (buf == NULL) {
-		diag_set(OutOfMemory, buf_size, "region_alloc", "buf");
-		return buf;
-	}
-	va_start(ap, format);
-	mp_vformat(buf, buf_size, format, ap);
-	va_end(ap);
-	*size = buf_size;
-	return buf;
-}
-
 /**
  * Drop the constraint or foreign key. If fieldno is equal to UINT32_MAX then
  * the tuple constraint or tuple foreign key is dropped, otherwise the field
@@ -4378,7 +4385,7 @@ constraint_drop(uint32_t space_id, const char *name, uint32_t name_len,
 		snprintf(buf, size, "%s.%*s", path, name_len, name);
 		path = buf;
 	}
-	uint32_t ops_size;
+	size_t ops_size;
 	const char *ops = mp_format_on_region(region, &ops_size, "[[%s%s%u]]",
 					      "#", path, 1);
 	const char *end = ops + ops_size;
@@ -4485,7 +4492,7 @@ constraint_create(uint32_t space_id, const char *name, uint32_t name_len,
 	uint32_t used = region_used(region);
 	const char *val;
-	uint32_t val_size;
+	size_t val_size;
 	assert(type == CONSTR_FUNC || type == CONSTR_FKEY);
 	if (type == CONSTR_FUNC) {
 		val = mp_format_on_region(region, &val_size, "%u",
@@ -4502,7 +4509,7 @@ constraint_create(uint32_t space_id, const char *name, uint32_t name_len,
 	if (val == NULL)
 		return -1;
 	const char *ops;
-	uint32_t ops_size;
+	size_t ops_size;
 	if (is_empty) {
```
</details>

Gerold103 (approved):
Thanks for the fixes! Technically looks good. See one remaining comment among the old ones where I've shown how to reuse 'mpformat on region'.

src/box/box.cc
ImeevMA: Thank you for the diff! Applied.

changelogs/unreleased/gh-6986-introduce-constraint-interface.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* The space object in Lua now has an interface to create or drop foreign keys
+  and constraints (gh-6986).
```
locker, line 4: We shouldn't modify or extend the existing API without an approval from the PM team. Anyway, I don't understand *why* you need these new Lua methods. The idea was to introduce the C API that could be used by SQL. Ideally, we should rewrite the *existing* Lua API using it if possible, not introduce *new* API.

locker (commented):
1. Please refrain from setting the `full-ci` label - the label is supposed to be set before merge, when all necessary approvals are obtained. If you need to test your PR on a certain platform (e.g. it modifies the platform-specific code)
2. Why do you need two PRs - this is confusing. Should you close this one and leave only #7651?

changelogs/unreleased/gh-6986-introduce-constraint-interface.md
ImeevMA: I convert this PR to draft for now.

ImeevMA:
It doesn't look like this PR will be continued anytime soon. Closing.

PR #7702, opened 2022-09-22
vinyl: implement transaction isolation levels

### LINKED ISSUE #5522: Vinyl tx manager dirty reads, and difference with memtx tx manager
Option `memtx_use_mvcc_engine` is supposed to turn on transaction isolation for memtx engine. And it seems to be working. So until a change is committed, it is not visible.

This seems not to be the case for vinyl. Here the dirty reads still happen. Here is an example using synchronous replication:

```Lua
-- Instance 1

-- Step 1
box.cfg{
    listen = 3313,
    replication = {'127.0.0.1:3313', '127.0.0.1:3314'},
    memtx_use_mvcc_engine = true,
    replication_synchro_quorum = 2,
    replication_synchro_timeout = 1000,
}

-- Step 3
s = box.schema.create_space('test', {is_sync = true, engine = 'vinyl'})
_ = s:create_index('pk')
s:replace{1}

-- Step 6
fiber = require('fiber')
f = fiber.create(function() s:replace{2} end)
f:status() -- Suspended

-- Step 7
s:get{2} -- !!! THIS SHOULD NOT RETURN ANYTHING, BUT IT RETURNS {2}.
```

```Lua
-- Instance 2

-- Step 2
box.cfg{
    listen = 3314,
    replication = {'127.0.0.1:3313', '127.0.0.1:3314'},
    read_only = true,
    memtx_use_mvcc_engine = true,
}

-- Step 4
box.space.test:get{1} -- Return {1}, ok.

-- Step 5
os.exit(1)
```
The bug is in the step 7. All works as expected when I change space engine to memtx. So for vinyl there is no any way to turn dirty reads off. Honestly, vinyl tx manager is so intricated, that I have no idea how would it behave if I would try to read {2} and change something in a transaction. Would it conflict or not? Anyway, the behaviour above seems to be a bug. It should be at least the same for vinyl and memtx, and ideally - without dirty reads.

The same happens on alive replicas, if there is more than 1.

locker:
This commit adds support of transaction isolation levels introduced earlier for memtx mvcc by commit ec750af66d9497207b358d0bc4a465163283f7ab. The isolation levels work exactly in the same way as in memtx.

Closes #5522

### DIFF
--- a/src/box/vy_tx.c
+++ b/src/box/vy_tx.c
@@ -109,6 +109,7 @@ vy_tx_manager_new(void)
 	}
 
 	rlist_create(&xm->writers);
+	rlist_create(&xm->prepared);
 	rlist_create(&xm->read_views);
 	vy_global_read_view_create((struct vy_read_view *)&xm->global_read_view,
 				   INT64_MAX);
@@ -164,13 +165,15 @@ vy_tx_manager_read_view(struct vy_tx_manager *xm)
 	 * Check if the last read view can be reused. Reference
 	 * and return it if it's the case.
 	 */
+	struct vy_tx *last_prepared_tx = rlist_empty(&xm->prepared) ? NULL :
+		rlist_last_entry(&xm->prepared, struct vy_tx, in_prepared);
 	if (!rlist_empty(&xm->read_views)) {
 		rv = rlist_last_entry(&xm->read_views, struct vy_read_view,
 				      in_read_views);
 		/** Reuse an existing read view */
-		if ((xm->last_prepared_tx == NULL && rv->vlsn == xm->lsn) ||
-		    (xm->last_prepared_tx != NULL &&
-		     rv->vlsn == MAX_LSN + xm->last_prepared_tx->psn)) {
+		if ((last_prepared_tx == NULL && rv->vlsn == xm->lsn) ||
+		    (last_prepared_tx != NULL &&
+		     rv->vlsn == MAX_LSN + last_prepared_tx->psn)) {
 
 			rv->refs++;
 			return  rv;
@@ -182,9 +185,9 @@ vy_tx_manager_read_view(struct vy_tx_manager *xm)
 			 "mempool", "read view");
 		return NULL;
 	}
-	if (xm->last_prepared_tx != NULL) {
-		rv->vlsn = MAX_LSN + xm->last_prepared_tx->psn;
-		xm->last_prepared_tx->read_view = rv;
+	if (last_prepared_tx != NULL) {
+		rv->vlsn = MAX_LSN + last_prepared_tx->psn;
+		last_prepared_tx->read_view = rv;
 		rv->refs = 2;
 	} else {
 		rv->vlsn = xm->lsn;
@@ -332,11 +335,14 @@ vy_tx_create(struct vy_tx_manager *xm, struct vy_tx *tx)
 	tx->psn = 0;
 	rlist_create(&tx->on_destroy);
 	rlist_create(&tx->in_writers);
+	rlist_create(&tx->in_prepared);
 }
 
 void
 vy_tx_destroy(struct vy_tx *tx)
 {
+	assert(rlist_empty(&tx->in_prepared));
+
 	trigger_run(&tx->on_destroy, NULL);
 	trigger_destroy(&tx->on_destroy);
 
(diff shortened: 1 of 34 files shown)


alyapunov:
It seems that in the commit message of `vinyl: implement transaction isolation levels` must be:
...
achieve that, we add a new flag is_prepared_ok. If the flag **not** is set,
...

test/unit/vy_mem.c
```diff
@@ -297,12 +297,280 @@ test_iterator_restore_after_insertion(void)
...
+		STMT_TEMPLATE(25, REPLACE, 200, 2),
+		STMT_TEMPLATE(MAX_LSN + 15, REPLACE, 300, 1),
+		STMT_TEMPLATE(MAX_LSN + 5, REPLACE, 400, 1),
+		STMT_TEMPLATE(MAX_LSN + 25, REPLACE, 400, 2),
+		STMT_TEMPLATE_FLAGS(10, REPLACE, VY_STMT_SKIP_READ, 500, 1),
+		STMT_TEMPLATE_FLAGS(15, REPLACE, VY_STMT_SKIP_READ, 500, 2),
+		STMT_TEMPLATE_FLAGS(5, REPLACE, VY_STMT_SKIP_READ, 600, 1),
+		STMT_TEMPLATE(10, REPLACE, 600, 2),
+		STMT_TEMPLATE(30, REPLACE, 600, 3),
+		STMT_TEMPLATE_FLAGS(15, REPLACE, VY_STMT_SKIP_READ, 600, 4),
```
alyapunov, line 395: It's a bit misleading that key=600 LSN=15 is placed below of key=600 LSN=30. There's no problem for the unit test but perhaps want to make a different test case?
As for me there should be both LSN=15 and (for example) LSN=45 statements with SKIP_READ flag..

src/box/vy_tx.c
```diff
@@ -813,28 +820,16 @@ vy_tx_rollback_after_prepare(struct vy_tx *tx)
...
-	 * concern all the prepared TXs, all of them will be rollbacked
-	 * and xm->last_prepared_tx must be set to NULL in the end.
-	 * Thus we can set xm->last_prepared_tx to NULL now and it will be
-	 * correct in the end of the cascading rollback.
-	 *
-	 * We must not change xm->last_prepared_tx in all other cases,
-	 * it will be changed by the corresponding TX.
+	 * In the first case, the transaction isn't in the list of prepared
+	 * transactions hence there's no assertion that the tx->in_prepared
+	 * link must be empty (rlist_del is a no-op in this case).
```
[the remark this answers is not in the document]
locker: Right. Fixed.

```diff
diff --git a/src/box/vy_tx.c b/src/box/vy_tx.c
index 937fce60db4e..10fb115993b7 100644
--- a/src/box/vy_tx.c
+++ b/src/box/vy_tx.c
@@ -827,7 +827,8 @@ vy_tx_rollback_after_prepare(struct vy_tx *tx)
 	 *
 	 * In the first case, the transaction isn't in the list of prepared
 	 * transactions hence there's no assertion that the tx->in_prepared
-	 * link must be empty (rlist_del is a no-op in this case).
+	 * link must not be an empty list head (rlist_del is a no-op if the
+	 * link is an empty head).
 	 */
 	rlist_del_entry(tx, in_prepared);
 
```

</details>

test/unit/vy_mem.c
```diff
@@ -297,12 +297,280 @@ test_iterator_restore_after_insertion(void)
 	footer();
 }
 
+static const char *
+lsn_str(int64_t lsn)
+{
+	char *buf = tt_static_buf();
+	if (lsn == INT64_MAX) {
+		strlcpy(buf, "INT64_MAX", TT_STATIC_BUF_LEN);
```
alyapunov, line 305: Nit: you could just return "INT64_MAX" here.

locker:
> It seems that in the commit message of `vinyl: implement transaction isolation levels` must be: ... achieve that, we add a new flag is_prepared_ok. If the flag **not** is set, ...

Right. Replaced the **if** with **unless**.

test/unit/vy_mem.c
locker: Right :) Fixed here and in `test/unit/vy_cache.c` as well.

alyapunov (approved):

PR #7310, opened 2022-06-27
Introduce new FK to SQL

ImeevMA:
This patch-set introduces new foreign keys to SQL.

PR #7704, opened 2022-09-22
Introduce core CHECK constraints to SQL

### LINKED ISSUE #6986: Introduce new FOREIGN KEY and CHECK constraints to SQL

ImeevMA:
This patch-set forces SQL to create core CHECK constraint instead of SQL CHECK constraint.

Part of #6986


src/box/sql/func.c
```diff
@@ -2389,40 +2389,37 @@ func_sql_expr_call(struct func *func, struct port *args, struct port *ret)
...
-	vdbe_field_ref_prepare_data(ref, data, mp_size);
-	ref->format = format;
-	if (sql_bind_ptr(stmt, 1, ref) != 0)
-		goto error;
+	if (pe->mp_format != NULL) {
+		/*
+		 * In SQL, we can only retrieve fields that have names. There is
+		 * no point to prepare slots for nameless fields.
+		 */
+		uint32_t count = pe->mp_format->min_field_count;
```
Gerold103, line 2411: `min_field_count` can be < named field count. It can be smaller if the last named fields are nullable. Is that ok for this code?

src/box/sql/build.c
```diff
@@ -799,9 +828,10 @@ sql_create_check_contraint(struct Parse *parser)
 			      space_id_reg);
 		vdbe_emit_ck_constraint_create(parser, ck_def, space_id_reg,
 					       space->def->name);
-		assert(sqlVdbeGetOp(v, v->nOp - 1)->opcode == OP_SInsert);
+		struct VdbeOp *op = sqlVdbeGetOp(v, v->nOp - 2);
+		assert(op->opcode == OP_SInsert);
 		sqlVdbeCountChanges(v);
-		sqlVdbeChangeP5(v, OPFLAG_NCHANGE);
+		op->p5 = OPFLAG_NCHANGE;
```
Gerold103, line 834: Probably the same needs to be done in the patch about fkeys - count the successful ops as `OPFLAG_NCHANGE`. `row_count == 1` when success, otherwise if it wasn't an error then `== 0`. Regardless of the number of actually affected rows, AFAIR.

src/box/sql/build.c
```diff
@@ -1125,34 +1155,49 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
 	/*
-	 * Occupy registers for 5 fields: each member in
-	 * _ck_constraint space plus one for final msgpack tuple.
+	 * Occupy registers for 20 fields: each member in _func space plus one
+	 * for final msgpack tuple.
 	 */
-	int ck_constraint_reg = sqlGetTempRange(parser, 7);
-	sqlVdbeAddOp2(v, OP_SCopy, reg_space_id, ck_constraint_reg);
-	sqlVdbeAddOp4(v, OP_String8, 0, ck_constraint_reg + 1, 0,
+	int regs = sqlGetTempRange(parser, 20);
```
Gerold103, line 1161: Given that the storage format has changed, and for fkeys too, is there a plan to update the upgrade.lua script to make the existing users' schemas up to date?

src/box/sql/build.c
```diff
@@ -1125,34 +1155,49 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + 4, 0, "SQL_EXPR", P4_STATIC);
+	char *body = sqlDbStrDup(db, ck_def->expr_str);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + 5, 0, body, P4_DYNAMIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + 6, 0, "function", P4_STATIC);
+	char *param_list = sqlDbMallocRawNN(db, 32);
+	char *buf = mp_encode_array(param_list, 0);
+	sqlVdbeAddOp4(v, OP_Blob, buf - param_list, regs + 7,
+		      SQL_SUBTYPE_MSGPACK, param_list, P4_DYNAMIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + 8, 0, "any", P4_STATIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + 9, 0, "none", P4_STATIC);
```
Gerold103, line 1177: This could look more understandable if you would use the names from schema_def.h:
```C
sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_RETURNS, 0, "any", P4_STATIC);
sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_AGGREGATE, 0, "none", P4_STATIC);
// ...
```
Up to you.

src/box/sql/vdbe.c
```diff
@@ -2853,19 +2877,32 @@ case OP_Sequence: {           /* out2 */
...
+	struct tuple *tuple;
+	char *key_end = mp_encode_array(key, 0);
+	assert(key_end - key == 1);
+	if (box_index_max(space_id, 0, key, key_end, &tuple) != 0)
+		goto abort_due_to_error;
+	if (tuple == NULL) {
+		mem_set_uint(res, 1);
+		break;
+	}
+	uint32_t fieldno = space_id == BOX_SEQUENCE_ID ? BOX_SEQUENCE_FIELD_ID :
```
Gerold103, line 2900: Fieldno could be stored in `p3`, then you wouldn't depend on the space ID and field ID constants in this opcode. Up to you.

Gerold103 (commented):
Thanks for the patchset! Frankly, I was hardly able to understand the last commit, especially what is happening in `sql_create_check_contraint()`. But I hope the tests are good enough. See some nits below.

src/box/sql/func.c
ImeevMA: Hi! Thank you for the review! I've added a new commit that updates existing spaces with constraints and made some changes to the commit that change the CHECK constraint due to changes in #7688.

src/box/sql/build.c
ImeevMA: Thanks, fixed. I dropped this code and added this functionality to opcode.

src/box/sql/build.c
ImeevMA: Added.

src/box/sql/build.c
ImeevMA: Fixed.

src/box/sql/vdbe.c
ImeevMA: Fixed.

ImeevMA:
Hi @Gerold103 ! Thank you for the review! I've added a new commit that updates existing spaces with constraints and made some changes to the commit that change the CHECK constraint due to changes in #7688.

test/sql-tap/alter2.test.lua
```diff
@@ -224,15 +224,20 @@ test:do_catchsql_test(
...
     "alter2-6.2",
     [[
         ALTER TABLE t1 ADD CONSTRAINT ck CHECK(id > 0);
-    ]], { 1, "CHECK constraint 'CK' already exists in space 'T1'" })
+    ]], { 1, 'Duplicate key exists in unique index "name" in space "_func" '..
+         'with old tuple - [66, 1, "check_T1_CK", 0, "SQL_EXPR", "id > 0", '..
+         '"function", [], "any", "none", "none", true, false, true, ["LUA"], '..
+         '{}, "", "", ""] and new tuple - [67, 1, "check_T1_CK", 0, '..
+         '"SQL_EXPR", "id > 0", "function", [], "any", "none", "none", true, '..
+         'false, true, ["LUA"], {}, "", "", ""]'})
```
Gerold103, line 240: I remember there was an explicit intention to keep the error messages human readable even when update system spaces under the hood. This is why we have `vdbe_emit_halt_with_presence_test()` IIRC. We should probably keep the nice error message.

src/box/lua/upgrade.lua
```diff
@@ -1236,8 +1236,39 @@ local function revoke_write_access_on__collation_from_role_public()
...
+    local _fk = box.space[box.schema.FK_CONSTRAINT_ID]
+    local _ck = box.space[box.schema.CK_CONSTRAINT_ID]
+    log.info("convert constraints from _ck_constraint and _fk_constraint")
+    for _, v in _fk:pairs() do
+        local space = box.space[v.child_id]
+        local mapping = setmap({})
+        for i, id in pairs(v.child_cols) do
+            mapping[id + 1] = v.parent_cols[i] + 1
+        end
+        local fk = space.foreign_key or {};
```
Gerold103, line 1249: Don't need `;`. Same below for `local ck = ...`.

test/sql-luatest/constraint_test.lua
```diff
@@ -407,3 +412,33 @@ g.test_constraints_5 = function()
...
+            fk_unnamed_T1_2 = {field = {[0] = 1, [1] = 0}, space = s.id}}
+        t.assert_equals(s1[6].foreign_key, fk2)
+        local ck = {}
+        ck.ck_unnamed_T2_1 = box.func.check_T2_ck_unnamed_T2_1.id
+        ck.ck_unnamed_T2_2 = box.func.check_T2_ck_unnamed_T2_2.id
+        ck.ck_unnamed_T2_3 = box.func.check_T2_ck_unnamed_T2_3.id
+        t.assert_equals(s2[6].constraint, ck)
+        t.assert_equals(s2[6].foreign_key, nil)
+    end)
+end
```
Gerold103, line 444: I suggest to add the script which you used to make the snapshot. Like `test/vinyl/upgrade/*/fill.lua`. Also are you use the folder should be 2.8.3? Last release was 2.10.0 if I remember correctly.

src/box/lua/upgrade.lua
Gerold103, line 1249: AFAIK in the upgrade script the internal alter triggers are disabled. Which means the Lua space objects are not updated on any DDL. And that means `space.foreign_key` remains unchanged this entire loop. Same for `space.constraint`. Or how does it work?

Gerold103 (changes requested):
Thanks for the fixes!

extra/mkkeywordhash.c
```diff
@@ -85,7 +85,7 @@ static Keyword aKeywordTable[] = {
   { "CROSS",                  "TK_JOIN_KW",     true  },
```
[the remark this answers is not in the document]
ImeevMA: Yes, here: https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition-for-foreign-keys

src/box/fk_constraint.h
```diff
@@ -42,31 +42,13 @@ extern "C" {
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! Fixed in #7310.

src/box/sql/vdbeaux.c
```diff
@@ -1832,30 +1832,6 @@ sqlVdbeCloseStatement(Vdbe * p, int eOp)
...
- * This function is called when a transaction opened by the database
- * handle associated with the VM passed as an argument is about to be
- * committed. If there are outstanding deferred foreign key constraint
- * violations, return -1. Otherwise, 0.
- *
- * If there are outstanding FK violations and this function returns
- * -1 and set an error.
- */
-int
-sqlVdbeCheckFk(Vdbe * p, int deferred)
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed in #7310.

test/sql-tap/table.test.lua
```diff
@@ -731,40 +731,18 @@ test:do_catchsql_test(
...
-        DROP TABLE t6;
-        CREATE TABLE t6(a INT REFERENCES t4 MATCH FULL ON UPDATE SET DEFAULT DEFAULT 1 primary key);
-    ]], {
-        -- <table-10.4>
-        0
-        -- </table-10.4>
-    })
-
 test:do_catchsql_test(
     "table-10.5",
```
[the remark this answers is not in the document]
ImeevMA: Dropped test 10.6, in #7310.

test/sql/transitive-transactions.result
```diff
@@ -12,133 +12,8 @@ test_run:cmd("setopt delimiter ';'")
...
-...
-box.space.CHILD:select();
----
-- []
-...
-box.space.PARENT:select();
----
-- - [1, 1]
-...
-box.space._session_settings:update('sql_defer_foreign_keys', {{'=', 2, true}})
```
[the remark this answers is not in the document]
ImeevMA: Fixed, added new commit to #7310.

src/box/sql/vdbeInt.h
```diff
@@ -290,6 +290,11 @@ struct Vdbe {
 	bft explain:2;		/* True if EXPLAIN present on SQL command */
 	bft changeCntOn:1;	/* True to update the change-counter */
 	bft runOnlyOnce:1;	/* Automatically expire on reset */
+	/**
+	 * Do not expire this statement because it is an expression and does not
+	 * use external resources other than bind variables.
+	 */
+	bft is_func : 1;
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

src/box/sql/vdbe.c
```diff
@@ -1976,7 +1976,17 @@ case OP_Column: {
...
+	struct Mem *mem = &aMem[pOp->p1];
+	struct Mem *res = vdbe_prepare_null_out(p, pOp->p3);
+	if (mem_is_bin(mem)) {
+		uint32_t unused;
+		if (mem_from_mp(res, mem->z, &unused) != 0)
+			goto abort_due_to_error;
+		break;
+	}
+	assert(mem->type == MEM_TYPE_PTR);
+	struct vdbe_field_ref *ref = mem->u.p;
```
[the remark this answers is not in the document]
ImeevMA: True, fixed.

src/box/sql/build.c
```diff
@@ -1125,34 +1136,63 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
+		      regs + BOX_FUNC_FIELD_PARAM_LIST, SQL_SUBTYPE_MSGPACK,
+		      param_list, P4_DYNAMIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_RETURNS, 0, "any",
+		      P4_STATIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_AGGREGATE, 0,
+		      "none", P4_STATIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_SQL_DATA_ACCESS,
+		      0, "none", P4_STATIC);
+	sqlVdbeAddOp2(v, OP_Bool, true, regs + BOX_FUNC_FIELD_IS_DETERMINISTIC);
+	sqlVdbeAddOp2(v, OP_Bool, false, regs + BOX_FUNC_FIELD_IS_SANDBOXED);
```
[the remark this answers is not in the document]
ImeevMA: Fixed.

test/sql-tap/alter2.test.lua
ImeevMA: Fixed.

src/box/lua/upgrade.lua
ImeevMA: Fixed.

test/sql-luatest/constraint_test.lua
ImeevMA: Thanks, done!

changelogs/unreleased/gh-6986-drop-fk-parsing-rules.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
```
Gerold103, line 1: Seems it is mostly about `sql`, so I would suggest `bugfix/sql`. Up to you.

src/box/sql.c
```diff
@@ -1581,3 +1574,122 @@ sql_constraint_drop(uint32_t space_id, const char *name)
...
+	const char *value = NULL;
+
+	struct region *region = &fiber()->gc;
+	uint32_t used = region_used(region);
+	if (mapping == NULL) {
+		count = child->def->fields[child_fieldno].constraint_count;
+		cdefs = child->def->fields[child_fieldno].constraint_def;
+		int len = snprintf(buf, 64, "format[%u].foreign_key",
+				   child_fieldno + 1);
+		assert(len > 0 && len < 64);
```
Gerold103, line 1643: Usually so as not to hardcode the values it is handy to do either this `64` -> `sizeof(buf)`, or:
```C
const int buf_size = 64;
char buf[buf_size];
// ...
snprintf(buf, buf_size, ...);
```
Same in `sql_constraint_create()` with `key[16]` variable and `sql_check_create()` with `buf[64]`.
Up to you.

changelogs/unreleased/gh-6986-introduce-sql-expr-functions.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
```
Gerold103, line 1: Seems like rather an SQL feature, not box.

src/box/sql/func.c
```diff
@@ -2414,40 +2414,33 @@ func_sql_expr_call(struct func *func, struct port *args, struct port *ret)
...
-	uint32_t count = format->min_field_count;
-	struct vdbe_field_ref *ref;
-	size_t size = sizeof(ref->slots[0]) * count + sizeof(*ref);
-	ref = region_aligned_alloc(region, size, alignof(*ref));
-	vdbe_field_ref_prepare_data(ref, data, mp_size);
-	ref->format = format;
-	if (sql_bind_ptr(stmt, 1, ref) != 0)
-		goto error;
+	if (pe->mp_format != NULL) {
+		uint32_t count = pe->mp_format->total_field_count;
```
Gerold103, line 2432: I see in the old code it was `min_field_count`. I remember this was fixed separately somewhere, no?

src/box/sql/vdbe.c
```diff
@@ -1968,15 +1968,27 @@ case OP_Column: {
...
  *
- * Interpret data P1 points at as an initialized vdbe_field_ref
- * object.
+ * Register P1 contains an initialized vdbe_field_ref object or encoded value.
+ * In the case of a value, the register type is BIN, otherwise the type is PTR.
  *
- * If P4 is not a field name, extract the P2th field from the tuple. Otherwise,
- * get the field with the given name. The retrieved value is stored in
- * register P3.
+ * In case of P1 being encoded value, decode this value to register P3.
```
Gerold103, line 1974: Can you explain how P1 gets an encoded value step by step? I am struggling to find it from the code.

src/box/sql.c
```diff
@@ -1693,3 +1673,45 @@ sql_foreign_key_create(const char *name, uint32_t child_id, uint32_t parent_id,
...
+		if (cdefs[i].type != CONSTR_FUNC)
+			continue;
+		if (strcmp(name, cdefs[i].name) == 0) {
+			diag_set(ClientError, ER_CONSTRAINT_EXISTS, "CHECK",
+				 name, space_name(space));
+			return -1;
+		}
+	}
+	int rc = sql_constraint_create(name, space_id, path, value);
+	return rc;
```
Gerold103, line 1716: Apparently you can also do `return sql_constraint_create(...)` without `rc` variable.

src/box/sql/build.c
```diff
@@ -1125,34 +1136,72 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
+	char *func_name = sqlMPrintf(db, "check_%s_%s", space_name,
+				     ck_def->name);
+	int name_reg = regs + BOX_FUNC_FIELD_NAME;
+	sqlVdbeAddOp4(v, OP_String8, 0, name_reg, 0, func_name, P4_DYNAMIC);
+	const char *err = tt_sprintf("Function for the check constraint '%s' "
+		"with name '%s' already exists", ck_def->name, func_name);
+	if (vdbe_emit_halt_with_presence_test(parser, BOX_FUNC_ID, 2, name_reg,
+					      1, ER_SQL_EXECUTE, err, 0,
+					      OP_NoConflict) != 0) {
+		sqlReleaseTempRange(parser, regs, 20);
```
Gerold103, line 1152: I would recommend not to hardcode 20 more than in one place, but rather store it in a const int. Up to you.

src/box/sql/build.c
```diff
@@ -1125,34 +1136,72 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
-					      ER_CONSTRAINT_EXISTS, error_msg,
-					      false, OP_NoConflict) != 0)
+	int regs = sqlGetTempRange(parser, 20);
+	char *func_name = sqlMPrintf(db, "check_%s_%s", space_name,
+				     ck_def->name);
+	int name_reg = regs + BOX_FUNC_FIELD_NAME;
+	sqlVdbeAddOp4(v, OP_String8, 0, name_reg, 0, func_name, P4_DYNAMIC);
+	const char *err = tt_sprintf("Function for the check constraint '%s' "
+		"with name '%s' already exists", ck_def->name, func_name);
+	if (vdbe_emit_halt_with_presence_test(parser, BOX_FUNC_ID, 2, name_reg,
```
Gerold103, line 1149: Might be better for readibility to do `2` -> `BOX_FUNC_FIELD_NAME` and `err, 0` -> `err, false` since the next to last arg is a boolean.

Gerold103 (commented):
Thanks for the fixes! Only minor comments are left.

src/box/sql/func.c
ImeevMA: You mentioned this previously: https://github.com/tarantool/tarantool/pull/7704#discussion_r981782286

Is that what you're asking?

changelogs/unreleased/gh-6986-drop-fk-parsing-rules.md
ImeevMA: Thank you the review! Fixed.

src/box/sql.c
ImeevMA: Thanks, fixed.

changelogs/unreleased/gh-6986-introduce-sql-expr-functions.md
ImeevMA: Fixed.

src/box/sql/vdbe.c
ImeevMA: I expanded the comment a bit, but I am not sure if this is enough.

Currently, the only case where this opcode can receive an encoded value is when the SQL_EXPR function is used as a field check constraint. In this case, the function gets the encoded value from the port instead. In case of tuple check constraint it takes an array containing all fields + format. I think we should rework the entire argument-taking system for SQL_EXPR functions, but I'm not sure if we should do that here, since SQL_EXPR functions currently can only work as constraints and nothing more.

src/box/sql.c
ImeevMA: True, fixed.

src/box/sql/build.c
ImeevMA: I have added a new variable and used the new value `box_func_field_MAX` in some places.

src/box/sql/build.c
ImeevMA: Thanks, fixed!

ImeevMA:
Hi @Gerold103! I pushed a new version, however there is a problem with permissions, described here: https://github.com/tarantool/tarantool/issues/7873

As stated, we can expect converted CHECK constraints to not work until the issue is fixed or the user has been explicitly granted the necessary permissions.

src/box/sql/func.c
Gerold103: Yeah. This looks like a separate bug, no? I can't see how `min_field_count` -> `total_field_count` change is related to introduction of `SQL_EXPR` functions. Am I missing something?

src/box/sql/vdbe.c
Gerold103: Ok, I see now. My concert is that it seems wrong to use `OP_Fetch` like that. In case of `SQL_EXPR` (field check) it doesn't really fetch anything. It only does `mem_from_mp()` on its single argument. The behaviour is too different depending on whether `p1` is bin or ptr.

What I propose below is up to you. It might simplify `OP_Fetch` (keep it like before I mean). AFAIU in `func_sql_expr_call()` you do `sql_bind_bin_static()` instead of using `vdbe_field_ref` with `sql_bind_ptr()` because  the column is not an array (so `vdbe_field_ref_prepare_data()` wouldn't work). Correct?

The proposal is to add `vdbe_field_ref_prepare_array()` which takes `uint32 field_count, const char *arr_begin, const char *arr_end`. It won't do `mp_decode_array()` inside. Then for CHECK constraints you can keep `func_sql_expr_call()` mostly like before but you call `vdbe_field_ref_prepare_array(ref, 1, data, mp_size)` for field checks. `1` here means an array with a single field. The field used in `CHECK`. We won't need `MP_ARRAY` header in the single field data for that. Tuple format won't be needed either. Because in `OP_Fetch` you will fetch by field number, not by name.

The overhead will be that we allocate `vdbe_field_ref` on region for field CHECKs, but win is that `OP_Fetch` will have less branches and code might be just simpler.

I would be glad if you could try the approach above, see if it works at all. And maybe it would be simpler.

src/box/sql/build.c
```diff
@@ -714,36 +741,12 @@ sql_create_check_contraint(struct Parse *parser)
...
-				ck_idx++;
-		}
-		name = tt_sprintf("ck_unnamed_%s_%u", space->def->name, ck_idx);
-	}
+	char *name;
+	if (name_token->n != 0)
+		name = sql_name_from_token(parser->db, name_token);
+	else
+		name = sql_ck_unique_name_new(parser, is_field_ck);
 	size_t name_len = strlen(name);
```
Gerold103, line 750: Yo, apparently `name` can be `NULL` here if `sql_ck_unique_name_new()` fails. Might be better to check for `NULL`. Alternatively you might make all sql allocations `panic()` when they can't alloc somewhere inside `sqlMalloc` or whatever. Like `xmalloc()` does.

src/box/sql/build.c
```diff
@@ -1125,34 +1136,75 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
-					      ER_CONSTRAINT_EXISTS, error_msg,
-					      false, OP_NoConflict) != 0)
+	const int reg_count = box_func_field_MAX + 1;
+	int regs = sqlGetTempRange(parser, reg_count);
+	char *func_name = sqlMPrintf(db, "check_%s_%s", space_name,
+				     ck_def->name);
+	int name_reg = regs + BOX_FUNC_FIELD_NAME;
+	sqlVdbeAddOp4(v, OP_String8, 0, name_reg, 0, func_name, P4_DYNAMIC);
+	const char *err = tt_sprintf("Function for the check constraint '%s' "
+		"with name '%s' already exists", ck_def->name, func_name);
```
Gerold103, line 1149: `func_name` might be `NULL` if `sqlMPrintf()` above fails.

src/box/sql/build.c
```diff
@@ -1125,34 +1136,75 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
+	sqlVdbeAddOp2(v, OP_Integer, 0, regs + BOX_FUNC_FIELD_SETUID);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_LANGUAGE, 0,
+		      "SQL_EXPR", P4_STATIC);
+	char *body = sqlDbStrDup(db, ck_def->expr_str);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_BODY, 0, body,
+		      P4_DYNAMIC);
+	sqlVdbeAddOp4(v, OP_String8, 0, regs + BOX_FUNC_FIELD_ROUTINE_TYPE, 0,
+		      "function", P4_STATIC);
+	char *param_list = sqlDbMallocRawNN(db, 32);
+	char *buf = mp_encode_array(param_list, 0);
```
Gerold103, line 1171: `param_list` might be `NULL` if `sqlDbMallocRawNN()` above fails.

Gerold103 (changes requested):
Huge thanks for the fixes! Almost done!

src/box/sql/func.c
ImeevMA: You are right. I moved this change to a new patch.

src/box/sql/vdbe.c
ImeevMA: Thanks for the suggestion! I applied your suggestion and it made the patch simpler. However, I'm not entirely sure that `vdbe_field_ref` was designed for such usage. However, I think at some point we will change the design of `vdbe_field_ref`, for example moving it from VDBE to BOX.

src/box/sql/build.c
ImeevMA: Thank you for the review! I added an `'if` for now. I will replace `malloc()` in  `sqlMalloc()` by `xmalloc()` and drop unnecessary checks for `NULL`, but not in this PR.

src/box/sql/build.c
ImeevMA: Fixed.

src/box/sql/build.c
ImeevMA: Fixed.

src/box/sql/build.c
```diff
@@ -1194,85 +1196,28 @@ vdbe_emit_fk_constraint_create(struct Parse *parse_context,
...
+	if (fk->is_field_fk) {
+		sqlVdbeAddOp2(vdbe, OP_Integer, fk->links->child_field,
+			      regs + 2);
+		sqlVdbeAddOp2(vdbe, OP_Integer, fk->links->parent_field,
+			      regs + 3);
 	} else {
-		sqlVdbeAddOp2(vdbe, OP_Integer, fk->parent_id,
-				  constr_tuple_reg + 2);
+		uint32_t size;
+		char *raw = fk_constraint_encode_links(fk, &size);
```
Gerold103, line 1279: If `raw` is `NULL`, we probably have to set `parse_context->is_aborted`.

Btw, I also see that all these `sqlVdbeAddOp...()` don't set `->is_aborted` too. They should do it in `growOpArray()` func. Can also leave it as is if we will port it to `xmalloc()` soon anyway.

src/box/sql/build.c
```diff
@@ -1124,35 +1139,85 @@ vdbe_emit_ck_constraint_create(struct Parse *parser,
...
+	int regs = sqlGetTempRange(parser, reg_count);
+	int name_reg = regs + BOX_FUNC_FIELD_NAME;
+	sqlVdbeAddOp4(v, OP_String8, 0, name_reg, 0, func_name, P4_DYNAMIC);
+	const char *err = tt_sprintf("Function for the check constraint '%s' "
+		"with name '%s' already exists", ck_def->name, func_name);
+	if (vdbe_emit_halt_with_presence_test(parser, BOX_FUNC_ID,
+					      BOX_FUNC_FIELD_NAME, name_reg,
+					      1, ER_SQL_EXECUTE, err, false,
+					      OP_NoConflict) != 0) {
+		sqlReleaseTempRange(parser, regs, reg_count);
```
Gerold103, line 1162: You probably don't need to release it. The parsing has failed anything and everything will be freed. No?

Gerold103 (commented):
Thanks for the fixes! Only minor ones are left. But I have a more important comment. See the older discussion above.

src/box/sql/vdbe.c
Gerold103: Now there is some kind of duplication. Might be better to have a single constructor for `vdbe_field_ref`, like this (I haven't tested it, and `slots[0]` makes no sense now):

```Diff
@@ -1374,17 +1374,14 @@ sql_template_space_new(Parse *parser, const char *name)
  */
 static void
 vdbe_field_ref_fill(struct vdbe_field_ref *field_ref, struct tuple *tuple,
-		    const char *data, uint32_t data_sz)
+		    uint32_t mp_count, const char *data, uint32_t data_sz)
 {
 	field_ref->tuple = tuple;
 	field_ref->data = data;
 	field_ref->data_sz = data_sz;
-
-	const char *field0 = data;
 	field_ref->format = NULL;
-	uint32_t mp_count = mp_decode_array(&field0);
 	field_ref->field_count = MIN(field_ref->field_capacity, mp_count);
-	field_ref->slots[0] = (uint32_t)(field0 - data);
+	field_ref->slots[0] = 0;
 	memset(&field_ref->slots[1], 0,
 	       field_ref->field_count * sizeof(field_ref->slots[0]));
 	field_ref->slot_bitmask = 0;
@@ -1395,30 +1392,28 @@ void
 vdbe_field_ref_prepare_data(struct vdbe_field_ref *field_ref, const char *data,
 			    uint32_t data_sz)
 {
-	vdbe_field_ref_fill(field_ref, NULL, data, data_sz);
+	const char *field0 = data;
+	uint32_t mp_count = mp_decode_array(&field0);
+	vdbe_field_ref_fill(field_ref, NULL, mp_count, field0,
+			    (uint32_t)(field0 - data) + data_sz);
 }
 
 void
 vdbe_field_ref_prepare_tuple(struct vdbe_field_ref *field_ref,
 			     struct tuple *tuple)
 {
-	vdbe_field_ref_fill(field_ref, tuple, tuple_data(tuple),
-			    tuple_bsize(tuple));
+	const char *field0 = tuple_data(tuple);
+	uint32_t data_sz = tuple_bsize(tuple);
+	uint32_t mp_count = mp_decode_array(&field0);
+	vdbe_field_ref_fill(field_ref, tuple, mp_count, field0,
+			    (uint32_t)(field0 - data) + data_sz);
 }
 
 void
 vdbe_field_ref_prepare_array(struct vdbe_field_ref *ref, uint32_t field_count,
 			     const char *data, uint32_t data_sz)
 {
-	ref->tuple = NULL;
-	ref->data = data;
-	ref->data_sz = data_sz;
-	ref->format = NULL;
-	ref->field_count = field_count;
-	ref->slots[0] = 0;
-	memset(&ref->slots[1], 0, ref->field_count * sizeof(ref->slots[0]));
-	ref->slot_bitmask = 0;
-	bitmask64_set_bit(&ref->slot_bitmask, 0);
+	vdbe_field_ref_fill(ref, NULL, field_count, data, data_sz);
 }
```
</details>

But I now noticed a bigger thing. Apparently another place of struggling is that you still have field names in those field-checks. That is why even in the new version you keep changing `OP_Fetch` to make it work differently depending on `format == / != NULL`. You can't properly resolve the names, because have no format. As a result we can write ridiculous stuff like below:
```Lua
box.cfg{}
def = {language = 'SQL_EXPR', is_deterministic = true, body = 'x + y == 4'}
box.schema.func.create('abc', def)
format = {{name = 'A', type = 'integer', constraint = 'abc'}}
s = box.schema.space.create('test', {format = format})
s:create_index('i')
s:replace{1} -- fails
s:replace{2} -- works
```

In the expr I used `x` and `y`, the field name was `A`, and all worked anyway. `A` was resolved to `x` and `y` at the same time. I suggest to clarify with products if this is the intended behaviour. Looks highly arguable :(

Otherwise it looks wrong tbh. We can try to think of some solution if products will confirm that this behaviour is wrong. Maybe sql_expr funcs should get complete tuple even for field checks.

src/box/sql/build.c
ImeevMA: Thank you for the review! Fixed.

src/box/sql/build.c
ImeevMA: Thanks, fixed.

src/box/sql/vdbe.c
ImeevMA: Thank you for the diff and the suggestion! As for diff - I applied it after a bit of change. As for suggestion - I discussed this with the product managers and we decided that for SQL_EXPR functions used as a field constraint from Lua, we would only check than number of arguments of the function is one. For SQL_EXPR functions used as a field constraint from SQL, we will check that the number of arguments is one and that the argument name is the same as the field name.

Also I reworked a bit `OP_Fetch` and introduced `OP_FetchByName`.

src/box/sql/expr.c
```diff
@@ -3850,10 +3850,12 @@ sqlExprCodeTarget(Parse * pParse, Expr * pExpr, int target)
 		}
 	case TK_ID:
 		assert(pParse->vdbe_field_ref_reg > 0);
+		const char *name = pExpr->u.zToken;
 		int reg = pParse->vdbe_field_ref_reg;
-		sqlVdbeAddOp4(v, OP_Fetch, reg, 0, target,
-			      sqlDbStrDup(pParse->db, pExpr->u.zToken),
-			      P4_DYNAMIC);
+		if (sql_vdbe_fetch_by_name(v, name, reg, target) != 0) {
```
Gerold103, line 3855: That is a strange name. The only other function having `sql_vdbe` prefix is `sql_vdbe_prepare`. So the name isn't consistent with other functions which emit code. It also looks like it fetches something from vdbe object by its name, which isn't true. It just appends a newly generated operation to the machine. I suggest to make its naming consistent with other code-emission functions.

src/box/sql/vdbeInt.h
```diff
@@ -280,6 +286,15 @@ struct Vdbe {
...
 	uint32_t res_var_count;
+	/**
+	 * Number of variables in SQL_EXPR function. Can be 0, 1 or 'Many'.
+	 */
+	enum sql_expr_arg_count arg_count;
+	/**
+	 * Name of the only SQL_EXPR function variable. Invalid if
+	 * arg_count != ARG_COUNT_ONE.
+	 */
+	const char *arg_name;
```
Gerold103, line 297: Why do you need all that? Can you just walk the `Vdbe.aOp` array when needed, find all those `OP_FetchByName`, and extract names from them? And fail if found more than one? IMO it would look simpler than these new highly specialized fields and how complicated it became to emit `OP_FetchByName`.

You should be able to do all that in `space_init_constraints()`. It validates that the functions exist, and you can also validate that the field constraints whose language is `SQL_EXPR` contain only one same name in all `OP_FetchByName` ops.

src/box/sql/vdbe.c
```diff
@@ -1988,6 +1987,18 @@ case OP_Fetch: {
...
+		 * If no format is specified, we assume that vdbe_field_ref
+		 * contains a single field for a field constraint.
+		 */
+		if (p->arg_count == ARG_COUNT_MANY) {
+			diag_set(ClientError, ER_SQL_EXECUTE,
+				 "Number of arguments in a field constraint "
+				 "function is greater than one");
+			goto abort_due_to_error;
+		}
+		id = 0;
```
Gerold103, line 2001: You should be able to validate that at attempt to create such field constraint. See my previous comment. (In case function body can't be altered.)

Gerold103 (changes requested):
Thanks for the fixes! Yes, we are closer and closer!

src/box/sql/vdbe.c
Gerold103: Ok, understandable. That would allow to use the same sql checks for multiple fields.

src/box/sql/expr.c
ImeevMA: Thank you for the review! I have dropped this function as per your suggestions below.

src/box/sql/vdbeInt.h
ImeevMA: Thank you for the suggestion! Done.

src/box/sql/vdbe.c
ImeevMA: Done.

src/box/ck_constraint.h
```diff
@@ -76,6 +76,10 @@ struct ck_constraint_def {
 	 * disabled (set to false) contraint won't be fired.
 	 */
 	bool is_enabled;
+	/** True if this is a field constraint, false otherwise. */
+	bool is_field_ck;
+	/** Fieldno of the field contains the constraint. */
+	uint32_t fieldno;
```
Gerold103, line 82: Sorry, I just noticed it now. These fields you apparently added to store here some temporary data during parsing. That is a problem because the def is also stored in `ck_constraint` itself and these fields are left uninitialized there. I.e. if I would fetch these cks from `space->format`, the flag `is_field_ck` will be false even for field checks.

The suggestions are:
- In SQL parser introduce your own `sql_ck_constraint_def` of some sort with any additional fields. Or reuse `ck_constraint_parse`;
- Store the field and tuple checks in separate rlists. Don't know how to resolve the fieldno though. I am not sure you really need it. Store it all in `create_column_def` somehow? Created columns are ordered. Which means you must be able to learn ck fieldno by its parent column fieldno.

Gerold103 (changes requested):
Thanks for the fixes! On the finish line already!

src/box/sql.h
```diff
@@ -416,10 +416,31 @@ void
 vdbe_field_ref_prepare_tuple(struct vdbe_field_ref *field_ref,
 			     struct tuple *tuple);
 
+/**
+ * Fill vdbe_field_ref instance with given data. The data will not be encoded as
+ * an array.
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! Fixed.

src/box/sql.h
```diff
@@ -416,10 +416,31 @@ void
...
+
 /** Initialize a new vdbe_field_ref instance. */
 void
 vdbe_field_ref_create(struct vdbe_field_ref *ref, uint32_t capacity);
 
+struct func;
+
+/**
+ * Return FALSE if the SQL_EXPR function has more than one arguments.
+ * Return True otherwise.
```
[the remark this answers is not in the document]
ImeevMA: Done. I replaced comment with `Check if SQL_EXPR func has single arg` which will expand a little in the next patch.

src/box/sql/func.c
```diff
@@ -2486,3 +2480,22 @@ static struct func_vtab func_sql_expr_vtab = {
 	.call = func_sql_expr_call,
 	.destroy = func_sql_expr_destroy,
 };
+
+bool
+is_func_sql_expr_single_arg(const struct func *base)
+{
+	struct func_sql_expr *func = (struct func_sql_expr *)base;
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

src/box/sql/vdbe.c
```diff
@@ -1988,6 +1987,14 @@ case OP_Fetch: {
...
 			goto abort_due_to_error;
 		}
+	} else {
+		/*
+		 * If no format is specified, we assume that the vdbe_field_ref
+		 * contains one field for the field constraint, and assume that
+		 * this was already checked when the field constraint was
+		 * created.
+		 */
+		id = 0;
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

src/box/sql.h
```diff
@@ -416,10 +416,31 @@ void
...
+ */
+void
+vdbe_field_ref_prepare_array(struct vdbe_field_ref *ref, uint32_t field_count,
+			     const char *data, uint32_t data_sz);
+
 /** Initialize a new vdbe_field_ref instance. */
 void
 vdbe_field_ref_create(struct vdbe_field_ref *ref, uint32_t capacity);
 
+struct func;
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

src/box/tuple_constraint_func.h
```diff
@@ -14,11 +14,12 @@ struct space;
 
 /**
  * Initialize constraint assuming that a func_name identifies a struct func
- * in func_cache.
+ * in func_cache. If this is a field constraint, @a is_complex must be false.
+ * If this is a complex constraint, @a is_complex must be true.
```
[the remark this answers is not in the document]
ImeevMA: Renamed.

src/box/space.c
```diff
@@ -144,7 +144,8 @@ space_init_constraints(struct space *space)
 		if (constr->check != tuple_constraint_noop_check)
 			continue;
 		if (constr->def.type == CONSTR_FUNC) {
-			if (tuple_constraint_func_init(constr, space) != 0)
+			if (tuple_constraint_func_init(constr, space,
+						       true) != 0)
```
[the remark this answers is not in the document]
ImeevMA: Sorry, I don't see how adding SQL headers would help here. The only reason I added this flag is to find out if this constraint is a field constraint. We can think of another way, but I think adding a third argument is the simplest way. Also the `tuple_constraint_fkey_init()` function has a similar third argument, so I figured that it wouldn't be a problem. I asked Vova, and he said that it was fine for him.

I think it's better to leave this function in `sql/func.c` since it works with the internals of `struct Vdbe`.

src/box/ck_constraint.h
ImeevMA: I do not think that this is a problem. I plan to drop all code of old FK and CK as a follow-up PR, where I will drop or move all these structures. I think that there is no need to rework them in this PR.

src/box/space.c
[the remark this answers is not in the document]
Gerold103: Ok, if you don't care about headers, then it is even simpler to reduce the complexity:

```Diff
diff --git a/src/box/space.c b/src/box/space.c
index 73c8afc34..b0df79c81 100644
--- a/src/box/space.c
+++ b/src/box/space.c
@@ -54,6 +54,7 @@
 #include "tuple_constraint_func.h"
 #include "tuple_constraint_fkey.h"
 #include "wal_ext.h"
+#include "sql/func.h"
 
 int
 access_check_space(struct space *space, user_access_t access)
@@ -144,8 +145,7 @@ space_init_constraints(struct space *space)
 		if (constr->check != tuple_constraint_noop_check)
 			continue;
 		if (constr->def.type == CONSTR_FUNC) {
-			if (tuple_constraint_func_init(constr, space,
-						       false) != 0)
+			if (tuple_constraint_func_init(constr, space) != 0)
 				return -1;
 		} else {
 			assert(constr->def.type == CONSTR_FKEY);
@@ -161,17 +161,27 @@ space_init_constraints(struct space *space)
 					   constr->def.type == CONSTR_FKEY;
 			if (constr->check != tuple_constraint_noop_check)
 				continue;
-			if (constr->def.type == CONSTR_FUNC) {
-				if (tuple_constraint_func_init(constr,
-							       space,
-							       true) != 0)
-					return -1;
-			} else {
-				assert(constr->def.type == CONSTR_FKEY);
+			if (constr->def.type == CONSTR_FKEY) {
 				if (tuple_constraint_fkey_init(constr,
 							       space, i) != 0)
 					return -1;
+				continue;
 			}
+			assert(constr->def.type == CONSTR_FUNC);
+			if (tuple_constraint_func_init(constr, space) != 0)
+				return -1;
+			const struct func *func =
+				constr->func_cache_holder.func;
+			if (func->def->language != FUNC_LANGUAGE_SQL_EXPR)
+				continue;
+			if (func_sql_expr_has_single_arg(func))
+				continue;
+			diag_set(ClientError, ER_CREATE_CONSTRAINT,
+				 constr->def.name, space->def->name, "Number "
+				 "of arguments in a SQL field constraint "
+				 "function is greater than one");
+			constr->space = NULL;
+			return -1;
 		}
 	}
 	space->has_foreign_keys = has_foreign_keys;
diff --git a/src/box/sql.h b/src/box/sql.h
index 7a0868a82..932677ca3 100644
--- a/src/box/sql.h
+++ b/src/box/sql.h
@@ -429,10 +429,6 @@ vdbe_field_ref_prepare_array(struct vdbe_field_ref *ref, uint32_t field_count,
 void
 vdbe_field_ref_create(struct vdbe_field_ref *ref, uint32_t capacity);
 
-/** Check if SQL_EXPR func has single arg. */
-bool
-func_sql_expr_has_single_arg(const struct func *base);
-
 #if defined(__cplusplus)
 } /* extern "C" { */
 #endif
diff --git a/src/box/sql/func.h b/src/box/sql/func.h
index 374a5d83d..f36b6b513 100644
--- a/src/box/sql/func.h
+++ b/src/box/sql/func.h
@@ -16,6 +16,10 @@ struct func_def;
 struct func *
 func_sql_expr_new(const struct func_def *def);
 
+/** Check if SQL_EXPR func has single arg. */
+bool
+func_sql_expr_has_single_arg(const struct func *base);
+
 #if defined(__cplusplus)
 } /* extern "C" */
 #endif /* defined __cplusplus */
diff --git a/src/box/tuple_constraint_func.c b/src/box/tuple_constraint_func.c
index 6cd7cba9a..996dc928d 100644
--- a/src/box/tuple_constraint_func.c
+++ b/src/box/tuple_constraint_func.c
@@ -42,8 +42,7 @@ tuple_constraint_func_find(struct tuple_constraint *constr)
  * used for better error message).
  */
 static int
-tuple_constraint_func_verify(struct tuple_constraint *constr,
-			     struct func *func, bool is_field)
+tuple_constraint_func_verify(struct tuple_constraint *constr, struct func *func)
 {
 	const char *func_name = func->def->name;
 
@@ -56,14 +55,6 @@ tuple_constraint_func_verify(struct tuple_constraint *constr,
 				    func_name));
 		return -1;
 	}
-	if (func->def->language == FUNC_LANGUAGE_SQL_EXPR && is_field &&
-	    !func_sql_expr_has_single_arg(func)) {
-		diag_set(ClientError, ER_CREATE_CONSTRAINT, constr->def.name,
-			 constr->space->def->name, "Number of arguments in a "
-			 "SQL field constraint function is greater than one");
-		constr->space = NULL;
-		return -1;
-	}
 	if (!func->def->is_deterministic) {
 		diag_set(ClientError, ER_CREATE_CONSTRAINT,
 			 constr->def.name, constr->space->def->name,
@@ -145,7 +136,7 @@ tuple_constraint_func_unpin(struct tuple_constraint *constr)
 
 int
 tuple_constraint_func_init(struct tuple_constraint *constr,
-			   struct space *space, bool is_field)
+			   struct space *space)
 {
 	assert(constr->def.type == CONSTR_FUNC);
 	constr->space = space;
@@ -158,8 +149,7 @@ tuple_constraint_func_init(struct tuple_constraint *constr,
 		assert(constr->check == tuple_constraint_noop_check);
 		return 0;
 	}
-	if (func == NULL ||
-	    tuple_constraint_func_verify(constr, func, is_field) != 0) {
+	if (func == NULL || tuple_constraint_func_verify(constr, func) != 0) {
 		constr->space = NULL;
 		return -1;
 	}
diff --git a/src/box/tuple_constraint_func.h b/src/box/tuple_constraint_func.h
index ff65f0865..291c3a4e7 100644
--- a/src/box/tuple_constraint_func.h
+++ b/src/box/tuple_constraint_func.h
@@ -19,7 +19,7 @@ struct space;
  */
 int
 tuple_constraint_func_init(struct tuple_constraint *constr,
-			   struct space *space, bool is_field);
+			   struct space *space);
 
 #ifdef __cplusplus
 } /* extern "C" */
```
</details>

The number of changed lines is exactly the same, but everything is done in a single file where it belongs. I also moved `func_sql_expr_has_single_arg()` to `sql/func.h`. Because you had its definition in `sql/func.c`, but somewhy had it declared in `box/sql.h` which is weird.

src/box/tuple_constraint_func.h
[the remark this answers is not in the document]
Gerold103: Oh, that just means `tuple_constraint_def_decode_fkey()` is crutchy as well.

src/box/ck_constraint.h
Gerold103: If you say so.

Gerold103 (approved):
Thanks for all the fixes! I've left one final message in a comment from the previous review, which you might want to apply. But it is not blocking.

src/box/space.c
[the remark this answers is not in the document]
ImeevMA: Hi! Thank you for the review! I'm not sure we should check the validity of the constraint outside of `tuple_constraint_func_verify()`. But I will ask the opinion of the second reviewer.

locker (approved):

PR #7706, opened 2022-09-23
xrow: fix crash on nested map/array update ops

Gerold103:
The patch fixes a crash in JSON update, #7705.

The test is in an existing diff test, because that test already has a good big tuple prepared for complex JSON updates and I don't want to copy-paste it. Probably better would be to port the entire test to luatest, but I feel reluctant to do it now as well.


locker (approved):

locker:
Cherry-picked to 2.10.

Issue #3714, opened 2018-10-04
Change console output format for zero ret values


Khatskevich:
Tarantool prints result of expression as yaml array in case ret_arg_cnt > 0:
```
tarantool> 1
---
- 1
...

tarantool> 1, 2
---
- 1
- 2
...

``` 
However for ret_arg_cnt == 0 it returns hardcoded value:
```
tarantool> 
---
...

```
Which is encoded as `null` according to yaml spec.

There are two proposals to make the output more consistent:
1. Return an empty array:
```
--- {}
...
```
2. Return blank string.

Totktonada:
Technically it would break backward compatibility, so we should revisit the issue and estimate cons and pros. I'll remove the 'good first issue' label, because the issue is the object for discussion.

Personally I dislike

```
--- []
...
```

as result of hitting Enter. However,

```
---
...
```

is not much better. I would prefer:

```
tarantool> <hit Enter>
tarantool>
```

However it may be different from a result of evaluating an expression with zero return values, right?

PR #7707, opened 2022-09-25
test: replace Lua assertions with luatest assertions in luatest tests

### LINKED ISSUE #7713: Replace `assert`s with luatest `assert`s in luatest tests
Some luatest framework tests use Lua `assert`s which are incomprehensible
when failed (the only information provided is 'assertion failed!'),
making debugging difficult: replace them with luatest `assert`s and their
context-specific varieties.

CuriousGeorgiy:
Some luatest framework tests use Lua assertions which are incomprehensible when failed (the only information provided is 'assertion failed!'), making debugging difficult: replace them with luatest assertions and their context-specific varieties.

Closes #7713

### DIFF
--- a/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
+++ b/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
@@ -50,7 +50,7 @@ local function tx_gc(server, steps, related_changes)
     if related_changes then
         table_apply_change(current_stat, related_changes)
     end
-    assert(table.equals(current_stat, server:eval('return box.stat.memtx.tx()')))
+    t.assert_equals(server:eval('return box.stat.memtx.tx()'), current_stat)
 end
 
 local function tx_step(server, txn_name, op, related_changes)
@@ -58,7 +58,7 @@ local function tx_step(server, txn_name, op, related_changes)
     if related_changes then
         table_apply_change(current_stat, related_changes)
     end
-    assert(table.equals(current_stat, server:eval('return box.stat.memtx.tx()')))
+    t.assert_equals(server:eval('return box.stat.memtx.tx()'), current_stat)
 end
 
 g.before_each(function()
@@ -76,7 +76,7 @@ g.before_each(function()
     -- CREATING CURRENT STAT
     current_stat = g.server:eval('return box.stat.memtx.tx()')
     -- Check if txm use no memory
-    assert(table_values_are_zeros(current_stat))
+    t.assert(table_values_are_zeros(current_stat))
 end)
 
 g.after_each(function()
@@ -167,8 +167,7 @@ g.test_simple = function()
     }
     tx_step(g.server, 'tx2', "s:replace{2, 2}", diff)
     tx_gc(g.server, 100, nil)
-    local err = g.server:eval('return tx2:commit()')
-    assert(not err[1])
+    t.assert_equals(g.server:eval('return tx2:commit()'), '')
     diff = {
         ["txn"] = {
             ["statements"] = {
@@ -194,8 +193,7 @@ g.test_simple = function()
         }
     }
     tx_gc(g.server, 10, diff)
-    err = g.server:eval('return tx1:commit()')
-    assert(not err[1])
+    t.assert_equals(g.server:eval('return tx1:commit()'), '')
     diff = {
         ["txn"] = {
             ["statements"] = {
@@ -225,7 +223,7 @@ g.test_simple = function()
         }
     }
     tx_gc(g.server, 100, diff)
-    assert(table_values_are_zeros(current_stat))
+    t.assert(table_values_are_zeros(current_stat))
 end
 
 g.test_read_view = function()
@@ -236,7 +234,7 @@ g.test_read_view = function()
     g.server:eval('s:replace{1, 1}')
     g.server:eval('s:replace{2, 1}')
     g.server:eval('box.internal.memtx_tx_gc(10)')
-    assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
+    t.assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
     g.server:eval('tx1("s:get(1)")')
     g.server:eval('tx2("s:replace{1, 2}")')
     g.server:eval('tx2("s:replace{2, 2}")')
@@ -283,7 +281,7 @@ g.test_read_view_with_empty_space = function()
     g.server:eval('s:replace{1, 1}')
     g.server:eval('s:replace{2, 1}')
     g.server:eval('box.internal.memtx_tx_gc(10)')
-    assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
+    t.assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
     g.server:eval('tx1("s:get(1)")')
     g.server:eval('tx2("s:delete(1)")')
     g.server:eval('tx2("s:delete(2)")')
@@ -361,7 +359,7 @@ g.test_conflict = function()
     g.server:eval('tx1 = txn_proxy.new()')
     g.server:eval('tx2 = txn_proxy.new()')
     g.server:eval('box.internal.memtx_tx_gc(10)')
-    assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
+    t.assert(table_values_are_zeros(g.server:eval('return box.stat.memtx.tx()')))
     g.server:eval('tx1:begin()')
     g.server:eval('tx2:begin()')
     g.server:eval('tx1("s:get(1)")')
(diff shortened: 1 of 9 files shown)


test/box-luatest/gh_6260_add_builtin_events_test.lua
```diff
@@ -13,7 +13,7 @@ g.test_subscriptions_outside_box_cfg = function()
         local result_no = 0
         local watcher = box.watch(val,
                                   function(name, state)
-                                      assert(name == val)
+                                      t.assert(name == val)
```
Totktonada, line 16: Why not `assert_equals()`?

test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
```diff
@@ -167,8 +167,7 @@ g.test_simple = function()
     }
     tx_step(g.server, 'tx2', "s:replace{2, 2}", diff)
     tx_gc(g.server, 100, nil)
-    local err = g.server:eval('return tx2:commit()')
-    assert(not err[1])
+    t.assert_equals(g.server:eval('return tx2:commit()'), '')
```
Totktonada, line 170: `box.commit()` actually returns zero return values and our console encodes it as `'---\n...\n'`. The reason that it is interpreted as an empty string is `yaml.decode()` in `tnx_proxy`. I think it is better to handle this case explicitly in `txn_proxy` and return no values.

There are some ideas around this console peculiar (see #3714), but I doubt that we'll change it. At least not unconditionally, because a script may lean on it.

test/box-luatest/gh_6260_add_builtin_events_test.lua
CuriousGeorgiy: That's a typo, thanks for noticing! Fixed.

test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
CuriousGeorgiy: > I think it is better to handle this case explicitly in txn_proxy and return no values.

I don't get how this can be applied in the context of this patch, could you please elabroate?

Totktonada:
```diff
diff --git a/test/box/lua/txn_proxy.lua b/test/box/lua/txn_proxy.lua
index 15b0e4add..02c6c7e19 100644
--- a/test/box/lua/txn_proxy.lua
+++ b/test/box/lua/txn_proxy.lua
@@ -22,7 +22,11 @@ local array_mt = { __serialize = 'array' }
 local mt = {
     __call = function(self, code_str)
         self.c1:put(code_str)
-        local res = yaml.decode(self.c2:get())
+        local res_str = self.c2:get()
+        if res_str == '---\n...\n' then
+            return
+        end
+        local res = yaml.decode(res_str)
         return type(res) == 'table' and setmetatable(res, array_mt) or res
     end,
     __index = {
```

It'll require updating of many usages, so deserves its own commit. But as result we'll get more straightforward behavior of `txn_proxy`.

Totktonada: Alternatively we can update the testing code step-by-step:

```lua
diff --git a/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua b/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
index df404d54c..49dda1577 100644
--- a/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
+++ b/test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
@@ -69,6 +69,7 @@ g.before_each(function()
     g.server:start()
 
     g.server:eval('txn_proxy = require("test.box.lua.txn_proxy")')
+    g.server:eval('txn_proxy.cfg({decode_empty_document_as_no_values = true})')
     g.server:eval('s = box.schema.space.create("test")')
     g.server:eval('s:create_index("pk")')
     -- Clear txm before test
@@ -167,8 +168,7 @@ g.test_simple = function()
     }
     tx_step(g.server, 'tx2', "s:replace{2, 2}", diff)
     tx_gc(g.server, 100, nil)
-    local err = g.server:eval('return tx2:commit()')
-    assert(not err[1])
+    t.assert_is(g.server:eval('return tx2:commit()'), nil)
     diff = {
         ["txn"] = {
             ["statements"] = {
@@ -194,8 +194,7 @@ g.test_simple = function()
         }
     }
     tx_gc(g.server, 10, diff)
-    err = g.server:eval('return tx1:commit()')
-    assert(not err[1])
+    t.assert_is(g.server:eval('return tx1:commit()', nil))
     diff = {
         ["txn"] = {
             ["statements"] = {
@@ -421,8 +420,7 @@ g.test_user_data = function()
         },
     }
     tx_step(g.server, 'tx', 'ffi.C.box_txn_alloc(' .. alloc_size .. ')', diff)
-    local err = g.server:eval('return tx:commit()')
-    assert(not err[1])
+    t.assert_is(g.server:eval('return tx:commit()', nil))
     diff = {
         ["txn"] = {
             ["user"] = {
diff --git a/test/box/lua/txn_proxy.lua b/test/box/lua/txn_proxy.lua
index 15b0e4add..e42d95d91 100644
--- a/test/box/lua/txn_proxy.lua
+++ b/test/box/lua/txn_proxy.lua
@@ -19,10 +19,24 @@ local console = require('console')
 
 local array_mt = { __serialize = 'array' }
 
+local cfg = {
+    -- Disabled by default for compatibility.
+    --
+    -- When disabled returns an empty string for
+    -- a command that doesn't return any value.
+    decode_empty_document_as_no_values = false,
+}
+
 local mt = {
     __call = function(self, code_str)
         self.c1:put(code_str)
-        local res = yaml.decode(self.c2:get())
+        local res_str = self.c2:get()
+        if cfg.decode_empty_document_as_no_values then
+            if res_str == '---\n...\n' then
+                return
+            end
+        end
+        local res = yaml.decode(res_str)
         return type(res) == 'table' and setmetatable(res, array_mt) or res
     end,
     __index = {
@@ -51,4 +65,15 @@ local function new_txn_proxy()
     }, mt)
 end
 
-return { new = new_txn_proxy }
+local function cfg_txn_proxy(new_cfg)
+    for k, v in pairs(new_cfg) do
+        assert(cfg[k] ~= nil)
+        assert(type(v) == 'boolean')
+        cfg[k] = v
+    end
+end
+
+return {
+    new = new_txn_proxy,
+    cfg = cfg_txn_proxy,
+}
```

However it looks overcomplicated.

test/replication-luatest/gh_5568_read_only_reason_test.lua
```diff
@@ -330,10 +329,13 @@ g.test_read_only_reason_synchro_no_uuid = function(g)
...
-                   box.info.replication[leader_id] == nil
-        end, {leader_id}))
+        t.assert_not_equals(g.replica:exec(function()
+            return box.info.synchro.queue.owner
+        end), 0)
+
+        t.assert_equals(g.replica:exec(function(leader_id)
+            return box.info.replication[leader_id]
+        end, {leader_id}), nil)
     end)
```
Totktonada, line 339: I don't know for sure, but maybe those two conditions are checked without yeilds in between intentionally.

test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
CuriousGeorgiy: I understand what you mean and I thought about this before too, but here I wanted to emphasize is that it is not applicable in scope of this patch: I suggest we open a separate ticket concerning this matter.

locker: > ```
>     g.server:eval('tx1:begin()')
>     g.server:eval('tx2:begin()')
>     g.server:eval('tx1("s:get(1)")')
>     g.server:eval('tx2("s:replace{1, 2}")')
>     g.server:eval("box.internal.memtx_tx_gc(10)")
> ```

The test is screwed: as long as it uses `eval` like that, replacing plain `assert` with `luatest.assert` is no better than putting lipstick on a pig. I think we should rewrite it using `exec` while we're at it or not touch it at all.

Totktonada: I think it is a kind of thing, which only has a chance to be made if someone wants it. Not really requires an issue and not really worth it. I don't mind ignoring it.

locker (approved):

test/box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua
CuriousGeorgiy: > Not really requires an issue and not really worth it.

No, I really believe it's also about code health: `txn_proxy` should ideally behave like streams and builtin operations.

PR #7711, opened 2022-09-26
box: support module as unit of reconfiguring

nshy:
Feedback daemon is reconfiguread as many times as number of options changed during single box.cfg call which is suboptimal. Add proper support for modules which configure all its options in a single call.

Part of https://github.com/tarantool/tarantool-ee/issues/200

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring

### DIFF
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -329,7 +329,6 @@ end
 -- load_cfg() may report an error, but box will be configured in
 -- fact.
 local dynamic_cfg = {
-    listen                  = private.cfg_set_listen,
     replication             = private.cfg_set_replication,
     log_level               = log.box_api.cfg_set_log_level,
     log_format              = log.box_api.cfg_set_log_format,
@@ -350,10 +349,6 @@ local dynamic_cfg = {
     checkpoint_wal_threshold = private.cfg_set_checkpoint_wal_threshold,
     wal_queue_max_size      = private.cfg_set_wal_queue_max_size,
     worker_pool_threads     = private.cfg_set_worker_pool_threads,
-    feedback_enabled        = ifdef_feedback_set_params,
-    feedback_crashinfo      = ifdef_feedback_set_params,
-    feedback_host           = ifdef_feedback_set_params,
-    feedback_interval       = ifdef_feedback_set_params,
     -- do nothing, affects new replicas, which query this value on start
     wal_dir_rescan_delay    = function() end,
     wal_cleanup_delay       = private.cfg_set_wal_cleanup_delay,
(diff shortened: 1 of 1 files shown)


src/box/lua/load_cfg.lua
```diff
@@ -398,6 +394,47 @@ local default_cfg_on_revert = {
     listen                  = nil,
 }
 
+-- The modules that can aplly all new options with single call. The
```
locker, line 397: Nit: Typo 'aplly'.

src/box/lua/load_cfg.lua
```diff
@@ -398,6 +394,47 @@ local default_cfg_on_revert = {
...
+local function build_option2module()
+    for module, info in pairs(config_modules) do
+        for option in pairs(info.options) do
+            option2module[option] = module
+        end
+    end
+end
+
+add_single_option_modules()
+build_option2module()
```
locker, line 436: I don't think we need these helper functions. Let's just inline the code here.

src/box/lua/load_cfg.lua
```diff
@@ -662,43 +699,58 @@ local function compare_cfg(cfg1, cfg2)
     return table.equals(cfg1, cfg2)
 end
 
+local function try_rollback_option(oldcfg, key)
```
locker, line 702: I don't like this function, because it requires for the option to be revertable (have an entry in the `dynamic_cfg_revert`). I'd rather keep it inline or introduce a different function instead, which would revert the option to the old value and call the 'revert' callback if it's available. Also, I'd put the revert function in the `config_modules` table.

src/box/lua/load_cfg.lua
```diff
@@ -662,43 +699,58 @@ local function compare_cfg(cfg1, cfg2)
...
-    for _, key in pairs(ordered_cfg) do
-        local val = newcfg[key]
-        local oldval = oldcfg[key]
-        if not compare_cfg(val, oldval) then
-            rawset(oldcfg, key, val)
-            local result, err = pcall(dynamic_cfg[key])
-            if not result then
+    for _, module in pairs(ordered_cfg) do
+        local oldvals = {}
+        for key in pairs(modules_cfg[module]) do
```
locker, line 735: I don't like the name of this local variable - it's easy to confuse with `module_cfg`. Do we really need it though? Maybe, iterator over `module_config[module].options` instead?

nshy: On f2f discussion decided to give better names to `modules_cfg` and `ordered_cfg`.

locker (approved):

locker:
integration/cartridge test is failing. PTAL.

> ```
> integration.feedback.test_feedback
> ...k/tarantool/tarantool/test/integration/feedback_test.lua:55: No feedback received
> expected: a table value, actual: type nil, value nil
> ```

src/box/lua/load_cfg.lua
```diff
@@ -832,20 +927,10 @@ local function load_cfg(cfg)
...
-                if log_cfg_option[key] ~= nil then
-                    val = log_cfg_option[key](val)
-                end
-                log.info("set '%s' configuration option to %s", key, json.encode(val))
-            end
-        end
-    end
+    --
+    -- FIXME actually is can raise an error. Even not all required checks
+    -- are done in private.cfg_check()
```
locker, line 932: s/actually is/actually it
s/Even not all/Even if all
?

src/box/lua/load_cfg.lua
```diff
@@ -382,24 +373,39 @@ local dynamic_cfg = {
...
+            listen = true,
+        },
+        skip_at_load = true,
+        revert_cfg = private.cfg_set_listen,
+        revert_fallback = {
+            listen = nil,
+        },
+    },
+    feedback = {
+        cfg = private.feedback_daemon.set_feedback_params,
```
locker, line 398: This won't work if feedback daemon is disabled at the compile time.

locker (approved):

PR #7407, opened 2022-07-13
box: add box.is_configured

nshy:
It shows whether box.cfg() is called or not.

Closes #7255

@TarantoolBot document
Title: add box.is_configured

It shows whether box.cfg() is called or not.

Issue #7715, opened 2022-09-26
don't call box.cfg() from test in gh_7288_console_flavours_vs_txn_test.lua


nshy:
Such usage causes issues with running tarantool tests using system luatest (#7647) and suggested to be baned (tarantool/luatest#245).

kyukhin:
Seems related to https://github.com/tarantool/luatest/issues/245

PR #7722, opened 2022-09-27
box: expose box.info() before box.cfg()

### LINKED ISSUE #7255: Allow box.info before box.cfg
There is a local variable, that is not exposed:
https://github.com/tarantool/tarantool/blob/a353ec918756c889e2c98f62617050a574a0168e/src/box/lua/load_cfg.lua#L739

Currently, the only way to check is to do
```
type(box.cfg) == 'function'
```

Proposed change:
```
tarantool> box.info
---
- error: "builtin/box/load_cfg.lua:726: Please call box.cfg{} first\nstack traceback:\n\tbuiltin/box/load_cfg.lua:726:
    in function '__index'\n\t[string \"return box.info\"]:1: in main chunk\n\t[C]:
    in function 'pcall'\n\tbuiltin/box/console.lua:404: in function 'eval'\n\tbuiltin/box/console.lua:710:
    in function 'repl'\n\tbuiltin/box/console.lua:759: in function <builtin/box/console.lua:747>"
...

tarantool> box.is_configured
---
- false
...

tarantool> box.cfg{}
...skip...
---
...

tarantool> box.is_configured
---
- true
...
```

nshy:
So one can easily check current box status.

See discussion in previous [PR](https://github.com/tarantool/tarantool/pull/7407) on why box.info() and not box.is_configured.

Closes #7255

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -90,7 +90,7 @@
 #include "flightrec.h"
 #include "wal_ext.h"
 
-static char status[64] = "unknown";
+static char status[64] = "unconfigured";
 
 /** box.stat rmean */
 struct rmean *rmean_box;
--- a/test/box-luatest/gh_7255_box_info_before_cfg_test.lua
+++ b/test/box-luatest/gh_7255_box_info_before_cfg_test.lua
@@ -1,0 +1,8 @@
+local t = require('luatest')
+
+local g = t.group()
+
+g.test_box_info_before_box_cfg = function()
+    local info = box.info()
+    t.assert_equals(info.status, "unconfigured")
+end
--- a/changelogs/unreleased/gh-7255-box-info-before-cfg.md
+++ b/changelogs/unreleased/gh-7255-box-info-before-cfg.md
@@ -1,0 +1,3 @@
+## feature/core
+
+* box.info() can be called before box.cfg() (gh-7255).
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -824,6 +824,7 @@ local box_cfg_guard_whitelist = {
     broadcast = true;
     txn_isolation_level = true;
     NULL = true;
+    info = true;
 };
 
 -- List of box members that requires full box loading.


test/box-luatest/gh_7255_box_info_before_cfg_test.lua
```diff
@@ -0,0 +1,4 @@
+local t = require('luatest')
+
+local info = box.info()
+t.assert_equals(info.status, "not_configured")
```
Totktonada, line 4: Shouldn't we add a test function? I guess, in current implementation if it'll fail we'll unable to run any test with `luatest -p <...>`, because the fail will occur on loading test files.

src/box/box.cc
```diff
@@ -90,7 +90,7 @@
 #include "flightrec.h"
 #include "wal_ext.h"
 
-static char status[64] = "unknown";
+static char status[64] = "not_configured";
```
Totktonada, line 93: I would like to hear @Mons opinion on this initial value. It is okay for me, but I don't know any similar API in tarantool and would re-check it.

Totktonada:
The code itself is okay for me. The main work is about validating that it doesn't lead to access to uninitialized memory (visually/using sanitizers). I assume that you already verified it at least visually and will not re-check.

I would add a test that read all `box.info()` values before `box.cfg()`, so if we'll run it with a memory sanitizer (like ASAN or valgrind), we'll have a chance to spot the memory access problem. It worth to write it in a way that would also cover future `box.info()` fields.

nshy:
Yeah, I looked thru the code to check for uninitialized memory.

nshy:
Discussed the suggestion to read box.info in the test and decided that calling box.info() is enough as it reads c-memory to return result. The result can have metatable with __index method so some c-memory can be read on reading result's fields but these fields are not visible in iterator and test can't be future-proof.

test/box-luatest/gh_7255_box_info_before_cfg_test.lua
```diff
@@ -0,0 +1,8 @@
+local t = require('luatest')
+
+local g = t.group()
+
+g.test_box_info_before_box_cfg = function()
+    local info = box.info()
+    t.assert_equals(info.status, "not_configured")
+end
```
Totktonada, line 8: BTW, it'll fail if we'll run it using luatest directly (without test-run) due to #7715. It shouldn't affect CI, but may be sometimes confusing. Not a blocker.

test/box-luatest/gh_7255_box_info_before_cfg_test.lua
Totktonada: Solved.

Totktonada (approved):

test/box-luatest/gh_7255_box_info_before_cfg_test.lua
Totktonada: Seems fixed now.

src/box/box.cc
ochaton: like to `not_configured` sounds like before `box.cfg`

Mons:
@nshym discussed the spelling for the `not_configured` status and decided to use `unconfigured` instead.
Please, update

alyapunov (approved):

PR #7632, opened 2022-09-01
Memtx tree pagination

drewdzzz:
This patchset adds helpers, needed for pagination, new options after and fetch_pos for select (both FFI and LuaC), and pagination to memtx tree. All the types of tree indexes (multikey, functional) are suported.

Documentation will be added in a later patchset.

Closes #7633
Closes #7635
Closes #7636

PR #7724, opened 2022-09-28
CI sandbox

drewdzzz:


locker:
Please don't set `full-ci` for testing so as not to overload CI - instead leave just those workflows that you need for debugging, like I did [here](https://github.com/tarantool/tarantool/pull/7632/commits/c7ab4a149353ce8b94adbd37641c3bf7d88f155c).

Issue #7618, opened 2022-08-26
TAP compatible tests (followups)


tsafin:
Followup to #5000, #6815

There are a number of unit-tests, which do `check_plan` as their final step, and might be converted to TAP-compatible format (via `UNIT_TAP_COMPATIBLE` as in #6815) but which still use redundant result-file. 

```
23:15 $ grep -l check_plan --include='*.c' . -r |wc -l
56

23:15 $ ls ./*.result | wc -l
95
```

This will not solve the entire problem of using result-files for unit-tests (as there are 95 result files, and only 56 potentially or actually TAP-compatible unit-tests) but we could dramatically reduce their number.

tsafin:
Here is suggested approach:
1. remove all result files for which there is check_plan in a corresponding c

```
23:19 $ grep -l check_plan --include='*.c' . -r |xargs -iXX basename -s.c XX|xargs -iXX rm XX.result
rm: cannot remove 'merger.test.result': No such file or directory
rm: cannot remove 'unit.result': No such file or directory
rm: cannot remove 'http_parser.result': No such file or directory
rm: cannot remove 'datetime.result': No such file or directory
rm: cannot remove 'trigger.result': No such file or directory

**Comment by tsafin** [comment]
I had to disable json patch, as if even fix TAP compatible output (headers and comments) the test in `json.c` has intentionally checking invalid Unicode sequences, which breaks `test-run` UTF8 parser:
```
======================================================================================
WORKR TEST                                            PARAMS          RESULT
---------------------------------------------------------------------------------
[001] unit/json.test
[001] TAP13 parse failed ('utf-8' codec can't decode byte 0xc2 in position 2613: invalid continuation byte).
[001]
[001] No result file (unit/json.result) found.
[001] Run the test with --update-result option to write the new result file.
[001] [ fail ]
[Main process] Got failed test; gently terminate all workers...
[001] Worker "001_unit" got failed test; stopping the server...
---------------------------------------------------------------------------------
```

The problem is in `test_errors()` for the code:
```cpp
void
test_errors()
{
	header();
	plan(22);
	const char *path;
	int len;
	struct json_lexer lexer;
	const struct path_and_errpos errors[] = {
		/* Double [[. */
		{"[[", 2},
....
		/* Invalid UNICODE */
		{"['aaa\xc2\xc2']", 6},
		{".\xc2\xc2", 2},
	};

```

<details>
[...]

PR #7753, opened 2022-09-30
 Tap-compatible unit-tests, step.2

tsafin:
> Part of #7618
> Follow-up to #7619 

Convert second bunch of TAP-ready unit-tests to become TAP-compatible, thus removing their result files.

Converted tests are:
- `latch.test`
- `light_view.test`
- `luaL_iterator.test`
- `luaT_tuple_new.test`
- `mhash.test`
- `mp_print_unknown_ext.test`
- `popen.test`
- `prbuf.test`
- `raft.test`
- `random.test`
- `ratelimit.test`
- `reflection_c.test`
- `reflection_cxx.test`
- ~~`say.test`~~
- `serializer.test`
- `sio.test`
- `stailq.test`
- `string.test`
- `swim.test`, `swim_errinj.test`, `swim_proto.test`
- `tt_sigaction.test`
- `tuple_bigref.test`
- `tuple_uint32_overflow.test`
- `uri.test`
- `uri_parser.test`
- `uuid.test`
- `vy_cache.test`, `vy_mem.test`, `vy_point_lookup.test`, `vy_write_iterator.test`
- `watcher.test`
- `xmalloc.test`


test/unit/latch.c
```diff
@@ -1,6 +1,8 @@
 #include "memory.h"
```
Totktonada, line 1: How it looks in `git log`:

image

BTW, shouldn't the commits be attached to some issue?

test/unit/say.result
```diff
@@ -1,67 +0,0 @@
-1..33
-# type: file
-# next:
```
Totktonada, line 3: We miss those checks after converting of the test. Checking of those `type`, `next`, `identity`, `facility` tokens seems to be the main thing that is verified by the test. Let's drop this commit from the patchset or convert `note()` calls to 'assertions' (`ok()`, `is()` and so on).

test/unit/swim_test_utils.h
```diff
@@ -32,7 +32,6 @@
  */
```
Totktonada, line 32: > temp - swim_proto.test

It is in the commit message. It seems, some forgotten note.

Totktonada: BTW,

> Updated swwim.test

Totktonada (approved):

test/unit/latch.c
tsafin: > BTW, shouldn't the commits be attached to some issue?

That was weak attempt to avoid thousands of references in the #7618 leading to stale commits, which would be replaced and force-pushed after then...

Now, I've restored links of `Part of #7618`.

test/unit/say.result
tsafin: Thanks for the catch! I'll drop this commit - it needs some more lave later.

Totktonada (approved):

Lord-KA:
image
Nit: dot at the end of the sentence. Maybe add detailed description only to the last commit?

Lord-KA (approved):
Overall, thanks for the patch. LGTM.

PR #7758, opened 2022-10-01
memtx: split `memtx_tx_story_unlink_both` logic in two separate contexts

### LINKED ISSUE #7757: memtx `space:drop` triggers assertion in transaction manager
**Steps to reproduce**

```lua

s:create_index('pk')

s:insert{0}
s:get{0}
s:drop()

os.exit()
```

**Actual behavior**

```console
Assertion failed: (story != NULL), function memtx_tx_track_read_story_slow, file memtx_tx.c, line 2754.
```

**Expected behavior**

Script successfully exits.

CuriousGeorgiy:
`memtx_tx_story_unlink_both` is called in two separate contexts: on space delete and on rollback. In the former case we need to simply unlink the story, while in the latter case we need to rebind read and gap trackers, and, perhaps do some other logic in the future. Calling `memtx_tx_story_unlink_both` in the former context can trigger assertion: split the function and its helpers into two separate functions for each case, grouping the common logic into third `*_common` functions.

Closes #7757

### DIFF
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -1052,41 +1052,11 @@ memtx_tx_story_link_top(struct memtx_story *new_top,
  * Unlink a @a story from history chain in @a index (in both directions),
  * where old_top was at the top of chain (that means that index itself
  * stores a pointer to story->tuple).
- * This is a light version of function, intended for the case when the
- * appropriate change in will be done later by caller.
- * In addition to unlinking from in list, this function also rebinds gap
- * records to the new top in history chain.
- */
-static void
-memtx_tx_story_unlink_top_light(struct memtx_story *story, uint32_t idx)
-{
-	assert(story != NULL);
-	assert(idx < story->index_count);
-	struct memtx_story_link *link = &story->link[idx];
-
-	assert(link->newer_story == NULL);
-
-	struct memtx_story *old_story = link->older_story;
-	if (old_story != NULL) {
-		memtx_tx_story_unlink(story, old_story, idx);
-
-		/* Rebind gap records to the new top of the list */
-		struct memtx_story_link *old_link = &old_story->link[idx];
-		rlist_splice(&old_link->nearby_gaps, &link->nearby_gaps);
-	}
-}
-
-/**
- * Unlink a @a story from history chain in @a index (in both directions),
- * where old_top was at the top of chain (that means that index itself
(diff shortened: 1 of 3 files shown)


src/box/memtx_tx.c
```diff
@@ -1128,8 +1098,100 @@ memtx_tx_story_unlink_top(struct memtx_story *story, uint32_t idx)
...
 		memtx_tx_unref_from_primary(story);
 	}
+}
+
+/**
+ * Unlink a @a story from history chain in @a index (in both directions),
+ * where old_top was at the top of chain (that means that index itself
+ * stores a pointer to story->tuple).
+ * This is a light version of function, intended for the case when the
+ * appropriate change in will be done later by caller.
```
drewdzzz, line 1108: appropriate change in *the index* will be done later by caller?

src/box/memtx_tx.c
```diff
@@ -1128,8 +1098,100 @@ memtx_tx_story_unlink_top(struct memtx_story *story, uint32_t idx)
...
+		memtx_tx_story_unlink(story, old_story, idx);
+}
+
+/**
+ * See description of `memtx_tx_story_unlink_top_light_common`.
+ * As opposed to `memtx_tx_story_unlink_top_light_on_space_delete`, also
+ * rebinds gap records to the new top in history chain.
+ */
+static void
+memtx_tx_story_unlink_top_light_on_rollback(struct memtx_story *story,
```
drewdzzz, line 1128: `memtx_tx_story_unlink_top_light_on_rollback` is a light version of `memtx_tx_story_unlink_top_on_rollback`, so `light` in the end of the method name would be a better variant. The same for all other methods.

drewdzzz (changes requested):

src/box/memtx_tx.c
CuriousGeorgiy: Yeah, good point, renamed as you suggested.

src/box/memtx_tx.c
CuriousGeorgiy: I just copy-pasted the comments, so I didn't really pay attention, thx for noticing: fixed.

drewdzzz (approved):

alyapunov (approved):

alyapunov:
Cherry-picked to 2.10.

PR #7759, opened 2022-10-01
memtx: fix use-after-free of successor in `tree_iterator_start`

### LINKED ISSUE #7755: Repeatable read violation with reverse iterator in memtx
**Steps to reproduce**
```lua

local json = require('json')
local log = require('log')
local txn_proxy = require('txn_proxy')

box.schema.space.create('s')
box.space.s:create_index('pk')

box.internal.memtx_tx_gc(1)

local tx1 = txn_proxy:new()
local tx2 = txn_proxy:new()
local tx3 = txn_proxy:new()

tx1('box.begin()')
tx2('box.begin()')
tx3('box.begin()')


tx1('box.space.s:insert{2}')
tx1('box.rollback()')

tx2('box.space.s:insert{0}')
log.info('%s', json.encode(tx3('box.space.s:select({1}, {iterator = "LE", fullscan = true})')))


box.space.s:insert{1}
log.info('%s', json.encode(tx3('box.space.s:select({1}, {iterator = "LE", fullscan = true})')))

os.exit()
```

**Actual behavior**

```console
 ---
[[[1]]]
...
```

**Expected behavior**

```console
 ---
[[]]
...
```

### LINKED ISSUE #7756: Crash on series of interactive transactions in memtx
**Steps to reproduce**

```lua

local netbox = require('net.box')

box.cfg{memtx_use_mvcc_engine = true, listen = 3301}

box.schema.space.create('s')
box.space.s:create_index('pk')

box.internal.memtx_tx_gc(3)

local conn = netbox.connect(3301)

local stream1 = conn:new_stream()
local stream2 = conn:new_stream()
local stream3 = conn:new_stream()

stream1:begin()
stream2:begin()
stream3:begin()


stream1.space.s:insert{2}
stream1:rollback()

stream2.space.s:insert{0}
stream3.space.s:select({1}, {iterator = "LE"})
box.space.s:insert{1}

os.exit()
```

**Actual behavior**

Segmentation fault

**Expected behavior**

Script successfully exits.

CuriousGeorgiy:
We assumed that the successor tuple's story could not get garbage collected on clarify of result tuple in `tree_iterator_start`, since they coincide in case of regular iterators. But this is not the case for reverse iterators: the result tuple is of-by-one from the successor, which means the successor's story can get garbage collected along with the tuple itself getting deleted, leading to use-after-free of successor: remove garbage collection from `memtx_tx_tuple_clarify` and call it manually.

The crash in #7756 revealed that the `put` in transaction manager's story hash table was performed incorrectly: fix it and add an assertion that nothing was replaced.

Closes #7755
Closes #7756

### DIFF
--- a/src/box/memtx_hash.cc
+++ b/src/box/memtx_hash.cc
@@ -152,6 +152,9 @@ name(struct iterator *iterator, struct tuple **ret)				\
 			return rc;						\
 		is_first = false;						\
 		*ret = memtx_tx_tuple_clarify(txn, space, *ret, idx, 0);	\
+/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/\
+		memtx_tx_story_gc();						\
+/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/\
 	} while (*ret == NULL);							\
 	return 0;								\
 }										\
@@ -172,8 +175,9 @@ hash_iterator_eq(struct iterator *it, struct tuple **ret)
 		return 0;
 	struct txn *txn = in_txn();
 	struct space *sp = space_by_id(it->space_id);
-/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
 	*ret = memtx_tx_tuple_clarify(txn, sp, *ret, it->index, 0);
+/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
+	memtx_tx_story_gc();
 /*********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND END**********/
 	return 0;
 }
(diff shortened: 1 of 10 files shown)


CuriousGeorgiy:
# :bangbang: ATTENTION :bangbang:

@alyapunov @drewdzzz  please not that first commit comes from #7758: we need to merge it first, otherwise the test in this PR will fail.

src/box/memtx_hash.cc
```diff
@@ -171,8 +171,9 @@ hash_iterator_eq(struct iterator *it, struct tuple **ret)
 		return 0;
 	struct txn *txn = in_txn();
 	struct space *sp = space_by_id(it->space_id);
-/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
 	*ret = memtx_tx_tuple_clarify(txn, sp, *ret, it->index, 0);
+/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
+	memtx_tx_story_gc();
```
drewdzzz, line 176: We should do the same for `WRAP_ITERATOR_METHOD` macro.

drewdzzz (changes requested):
The patch solves the set problems, but I am concerned about the current development vector - some methods categorically cannot call GC, and some methods on the contrary can do it implicitly (for example, gap trackers now should not call GC, but all other trackers can), and most importantly - during the process of development and support of transaction manager we must ensure that methods without GC do not call methods with it. I believe we need to redesign the transaction manager garbage collection system. The easiest option is to make ALL GC calls explicit.

src/box/memtx_hash.cc
CuriousGeorgiy: Thanks, missed that one: added garbage collection to `WRAP_ITERATOR_METHOD`.

drewdzzz (approved):
I've created an [issue](https://github.com/tarantool/tarantool/issues/7867) with my concern - this patch is OK.

alyapunov (approved):

alyapunov:
Backported to 2.10.

PR #7760, opened 2022-10-02
memtx: add assertion and test for gap tracker retention during rollback

CuriousGeorgiy:
Gap trackers are stored in the story at the top of the history chain (see `memtx_tx_story_{un}link_top_light`): following this logic, during rollback we need to rebind them to the older story. We already maintain this in 8c24259, since a story is not retained iff it has a newer story (which means it does not store trackers) or if it has an older story (in which case rebinding is done in `memtx_tx_story_{un}link_top_light`): add an assertion that a story deleted during rollback does not have any gap trackers and a test for this case.

Follow-up 8c24259

NO_CHANGELOG=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -2212,6 +2212,8 @@ memtx_tx_history_rollback_added_story(struct txn_stmt *stmt)
 	memtx_tx_story_unlink_added_by(story, stmt);
 	if (!story->rollbacked) {
 		assert(rlist_empty(&story->reader_list));
+		for (uint32_t idx = 0; idx < story->index_count; ++idx)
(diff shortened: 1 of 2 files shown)


drewdzzz (approved):

alyapunov:
Please change to in the commit message

CuriousGeorgiy:
@alyapunov thanks for spotting this, fixed.

CuriousGeorgiy:
@alyapunov sorry, working on a lot of PRs in parallel, losing my carefullness.

alyapunov:
Cherry-picked to 2.10.

PR #7765, opened 2022-10-03
memtx: deprecate HASH index 'GT' iterator type

### LINKED ISSUE #7231: Deprecation of 'GE' iterator type for HASH index
# Description

_HASH_ index has a dubious 'GE' iterator type[^GE]:
> The comparison operator is ‘>’ (greater than). If a hash of an index key is greater than a hash of a search value, it matches. Tuples are returned in ascending order by hash of index key, which will appear to be random. Provided that the space is not being updated, one can retrieve all the tuples in a space, N tuples at a time, by using {iterator=’GT’, limit=N} in each search, and using the last returned value from the previous result as the start search value for the next search.

One could argue it can be use this for pagination, but pagination seems to be useless with the restriction that the space cannot be updated during iteration. As opposed to ordered indexes which give a strong guarantee on iterators between yields (i.e., that the new state of the database will be observed in an orderly manner) — _HASH_ index can give completely undetermined results: for instance, the same value could be observed twice or more times due to bucket reordering.

# Proposal

Deprecate 'GE' iterator type for _HASH_ index.

# Alternatives

Leave things as is.

# Deliverable

Either drop 'GE' iterator type for _HASH_ index right away or at least deprecate it.

[^GE]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/pairs/

CuriousGeorgiy:
For reasons described in #7231 HASH index 'GT' iterator type is deprecated: print a warning exactly once about the deprecation.

Closes #7231

### DIFF
--- a/changelogs/unreleased/gh-7231-memtx-hash-idx-gt-iter-deprecation.md
+++ b/changelogs/unreleased/gh-7231-memtx-hash-idx-gt-iter-deprecation.md
@@ -1,0 +1,3 @@
+## bugfix/memtx
+
+* Deprecated HASH index 'GT' iterator type (gh-7231).
--- a/src/box/memtx_hash.cc
+++ b/src/box/memtx_hash.cc
@@ -443,7 +443,16 @@ memtx_hash_index_create_iterator(struct index *base, enum iterator_type type,
 	light_index_iterator_begin(&index->hash_table, &it->iterator);
 
 	switch (type) {
-	case ITER_GT:
+	case ITER_GT: {
+		static bool warn_once = false;
+		if (!warn_once) {
+			warn_once = true;
+			say_warn("HASH index 'GT' iterator type is deprecated "
+				 "since Tarantool 2.11 and should not be "
+				 "used. It will be removed in a future "
+				 "Tarantool release.");
+		}
+
 		if (part_count != 0) {
 			light_index_iterator_key(&index->hash_table, &it->iterator,
 					key_hash(key, base->def->key_def), key);
@@ -459,6 +468,7 @@ memtx_hash_index_create_iterator(struct index *base, enum iterator_type type,
 					 &index->base);
 /*********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND END**********/
 		break;
+	}
 	case ITER_ALL:
 		light_index_iterator_begin(&index->hash_table, &it->iterator);
 		it->base.next_internal = hash_iterator_ge;
--- a/test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
+++ b/test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
@@ -1,0 +1,40 @@
+local fio = require('fio')
+local server = require('test.luatest_helpers.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new {
+        alias   = 'dflt',
+        box_cfg = {memtx_use_mvcc_engine = true}
+    }
+    cg.server:start()
+    cg.server:exec(function()
+        local s = box.schema.create_space('s')
+        s:create_index('pk', {type = 'HASH'})
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+-- Checks that a deprecation warning is printed exactly once when using memtx HASH
+-- index 'GT' iterator type.
+g.test_memtx_hash_idx_iter_gt_deprecation = function(cg)
+    cg.server:exec(function()
+        box.space.s:select({}, {iterator = 'GT'})
+    end)
+    local deprecation_warning = "HASH index 'GT' iterator type is " ..
+                                "deprecated since Tarantool 2.11 and should " ..
+                                "not be used. It will be removed in a " ..
+                                "future Tarantool release."
+    t.assert_is_not(cg.server:grep_log(deprecation_warning, 256), nil)
+    local log_file = g.server:exec(function() return box.cfg.log end)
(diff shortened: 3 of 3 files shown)


src/box/memtx_hash.cc
```diff
@@ -443,7 +443,17 @@ memtx_hash_index_create_iterator(struct index *base, enum iterator_type type,
...
-	case ITER_GT:
+	case ITER_GT: {
+		static bool warn_once = false;
+		if (!warn_once) {
+			warn_once = true;
+			say_warn("HASH index 'GT' iterator type is deprecated "
+				 "since Tarantool 2.11 and should not be "
+				 "used.\n"
+				 "It will be removed in a future Tarantool "
+				 "release.");
```
locker, line 454: A few minor comments:
 - Multi-line strings don't look good in the log.
 - I think "It will be removed in a future release" sentence is redundant - it follows from the fact that the feature was deprecated.
 - Maybe mention that 'ALL' should be used instead?
 - Should we maybe backport this to 2.10? The ticket lacks any labels and isn't included into the team plan.

changelogs/unreleased/gh-7231-memtx-hash-idx-gt-iter-deprecation.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/memtx
+
+* Deprecated **HASH** index 'GT' iterator type (gh-7231).
```
locker, line 3: Nit: Why is **HASH** in bold?

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
```diff
@@ -0,0 +1,41 @@
...
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+--[[
+Checks that a deprecation warning is printed exactly once when using memtx HASH
+index 'GT' iterator type.
+]]
```
locker, line 26: Nit: We format multi-line comments like this usually:

```Lua
-- Checks that a deprecation warning is printed exactly once when using memtx HASH
-- index 'GT' iterator type.
```

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
```diff
@@ -0,0 +1,41 @@
...
+    local deprecation_warning = "HASH index 'GT' iterator type is " ..
+                                "deprecated since Tarantool 2.11 and should " ..
+                                "not be used."
+    t.assert_is_not(cg.server:grep_log(deprecation_warning, 256), nil)
+    local log_file = g.server:eval('return box.cfg.log')
+    fio.truncate(log_file)
+    cg.server:exec(function()
+        box.space.s:select({}, {iterator = 'GT'})
+    end)
+    t.assert_is(cg.server:grep_log(deprecation_warning, 256), nil)
```
locker, line 40: I don't understand what you check here after log truncation.

locker (changes requested):

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
CuriousGeorgiy: I check that the warning is not printed again.

changelogs/unreleased/gh-7231-memtx-hash-idx-gt-iter-deprecation.md
CuriousGeorgiy: To emphasize it's rather a name than an acronym.

src/box/memtx_hash.cc
CuriousGeorgiy: > Should we maybe backport this to 2.10?

Since the warning states `since Tarantool 2.11`, I think we shouldn't.

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
locker: Ah, it should be obvious :) Hope that log truncation doesn't break anything.

src/box/memtx_hash.cc
locker: Maybe, deprecate it in 2.10 *and* fix the log message :) Need to ask @alyapunov just in case.

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
CuriousGeorgiy: Okay, fixed: I will stick to this style in the future.

test/box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua
CuriousGeorgiy: > Hope that log truncation doesn't break anything.

I used the same technique for checking that warning for fullscan select is rate limited and it seems to be okay.

Ne raz'ebet :grin:

changelogs/unreleased/gh-7231-memtx-hash-idx-gt-iter-deprecation.md
CuriousGeorgiy: But all our [documentation](https://www.tarantool.io/en/doc/latest/concepts/data_model/indexes/) refers to index types as CAPS.

locker (approved):
Looks good to me, but @alyapunov has some reservations about whether we should deprecate GT iterator for hash index at all.

alyapunov (approved):

Issue #7761, opened 2022-10-02
Secondary index uniqueness violation


CuriousGeorgiy:
**Steps to reproduce**

```lua
os.execute('rm -rf *.snap *.xlog *.vylog 512')

local log = require('log')
local txn_proxy = require('txn_proxy')

box.cfg{memtx_use_mvcc_engine = true}

local s = box.schema.space.create('s')
s:create_index('pk')
s:create_index('sk2', {parts = {{2}}})

local tx1 = txn_proxy:new()
local tx2 = txn_proxy:new()

tx1:begin()
tx2:begin()

tx2('box.space.s:insert{0, 1}')
tx1('box.space.s:insert{1, 0}')
tx1('box.space.s:replace{1, 1}')

tx2:commit()
tx1:commit()

log.info(box.space.s:select({}, {fullscan = true}))

os.exit()
```

**Actual behavior**

```console
 ---
[[0,1],[1,1]]
...
```

**Expected behavior**

```console
 ---
[[0,1]]
...
```

PR #7767, opened 2022-10-03
memtx secondary index uniqueness violation

CuriousGeorgiy:
memtx: fix conflict handling of stories in secondary indexes

If we find a newer story in the secondary index, the statement of which
deletes a story added in the same transaction, there can be two cases: if
the secondary index is not unique, this is impossible (since it's `cmp_def`
is extended with the primary index's `key_def` and there can only be story
chains in the primary index). Otherwise, the prepared story conflicts the
newer one: remove the incorrect check.

Closes https://github.com/tarantool/tarantool/issues/7761


drewdzzz (approved):
> there can only be story chains in the primary index

Story chains are present in every index. Please elaborate.

alyapunov (approved):

alyapunov:
Cherry-picked to 2.10.

PR #7769, opened 2022-10-04
box: allow slab_alloc_granularity = 4 back

### LINKED ISSUE #7422: Fix `box.cfg.slab_alloc_granularity = 4`
[`box.cfg.slab_alloc_granularity`](https://github.com/tarantool/doc/issues/1911) sets the min alignment used for [`struct memtx_tuple`](https://github.com/tarantool/tarantool/blob/fd35f8c799317dcf0a569fac31f77bc802c9975b/src/box/memtx_allocator.h#L47). By default, it's set to the machine word size (8 bytes typically), but it can be set to 4 bytes to reduce memory usage.

The problem is that we add memtx tuples to a garbage collection list using the intrusive list pattern (basically, store a pointer to the next entry in the first 8 bytes of the `memtx_tuple` struct). If `box.cfg.slab_alloc_factor` is set to 4, the pointer will be unaligned while the compiler will generate the machine code assuming it is naturally aligned. For some architectures (x86), this works fine, but for others (ARM), this may result in a runtime failure.

We need to instruct the compiler that the list pointer stored in `memtx_tuple` may be unaligned.

nshy:
It was disabled in commit ccfef5fdaff8 ("memtx: ignore slab_alloc_granularity < machine word size") as at that time it was not supported safely on all platforms.

This granularity implies `memtx_tuple` objects are not aligned as usual. So let's use PACKED attribute so that compiler generate code for unaligned objects. And where this is not possible (at least I don't managed to) use memcpy to safely deal with unaligned pointers.

Closes #7422

NO_DOC=internal
NO_CHANGELOG=internal


test/unit/stailq.c
```diff
@@ -1,13 +1,15 @@
 #include "salad/stailq.h"
 #include <stdio.h>
 #include <stdarg.h>
+
+#define UNIT_TAP_COMPATIBLE 1
```
locker, line 5: This was done in another commit. Please rebase.

src/lib/salad/stailq.h
```diff
@@ -67,8 +75,16 @@ inline static void
...
-	if (item->next == NULL)
+	/*
+	 * Without curly braces gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
+	 * generates code such that statement is executed unconditionally.
+	 */
+	if (item->next == NULL) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Waddress-of-packed-member"
 		head->last = &item->next;
+#pragma GCC diagnostic pop
```
locker, line 86: Why not use memset here? The pragmas may not work with other compilers (e.g. clang).

src/trivia/util.h
```diff
@@ -148,6 +148,16 @@ strnindex(const char *const *haystack, const char *needle, uint32_t len,
 #  define __has_cpp_attribute(x) 0
 #endif
 
+/**
+ * Use instead of `*ptr = value` statement in case ptr is unaligned and
+ * compiler is not informed of unalignment.
+ */
+#define uptr_deref_set(ptr, value) \
```
locker, line 155: Let's rename to `unaligned_store`? I think we also need `unaligned_load` so that we can drop the pragmas.

test/unit/stailq.c
nshy: Done.

src/lib/salad/stailq.h
nshy: After discussion with @locker and @alyapunov we decided to try approach without `memcpy`. So it is not applicable anymore.

src/trivia/util.h
nshy: Same as above. Not applicable anymore.

src/box/txn.c
```diff
@@ -66,7 +66,7 @@ const char *txn_isolation_level_aliases[txn_isolation_level_MAX] = {
 };
 
 /* Txn cache. */
-static struct stailq txn_cache = {NULL, &txn_cache.first};
+static STAILQ_HEAD(txn_cache);
```
locker, line 69: Nit: Let's call it simply `STAILQ`? There's no such thing as `stailq_head` and we usually name definition macros after the struct they define.

nshy: Done.

locker (approved):

PR #7773, opened 2022-10-04
sql: fix assertion in JOIN using unsupported index

### LINKED ISSUE #5678: Assertion with select and join and hash index
I have Ubuntu 18.04.
I have Tarantool 2.6, pulled from source today.
I execute these requests:
```lua
box.schema.space.create('T')
box.space.T:create_index('I',{type='hash'})
box.space.T:format({{name='COLUMN_1',type='unsigned'}})
box.execute([[SELECT a.column_1 FROM t a,t;]])
```
Result:
```
tarantool: /home/pgulutzan/tarantool-2.6/src/box/sql/vdbeaux.c:813: vdbeVComment: Assertion `p->aOp == 0 || p->aOp[p->nOp - 1].zComment == 0 || p->db->mallocFailed' failed.
Aborted (core dumped)
```

ImeevMA:
This patch fixed the assertion when JOIN uses index of unsupported type.

Closes #5678

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-5678-assertion-in-JOIN-using-unsupported-index.md
+++ b/changelogs/unreleased/gh-5678-assertion-in-JOIN-using-unsupported-index.md
@@ -1,0 +1,3 @@
+## bugfix/sql
+
+* Fixed assertion in JOIN in case of using unsupported index (gh-5678).
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -866,7 +866,7 @@ sql_column_collation(struct space_def *def, uint32_t column, uint32_t *coll_id)
 	return field->coll;
 }
 
-int
+void
 vdbe_emit_open_cursor(struct Parse *parse_context, int cursor, int index_id,
 		      struct space *space)
 {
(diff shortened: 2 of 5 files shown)


src/box/sql/where.c
```diff
@@ -4634,8 +4634,10 @@ sqlWhereBegin(Parse * pParse,	/* The parser context */
 			if (op) {
 				uint32_t space_id = idx_def->space_id;
 				struct space *space = space_by_id(space_id);
-				vdbe_emit_open_cursor(pParse, iIndexCur,
-						      idx_def->iid, space);
+				if (vdbe_emit_open_cursor(pParse, iIndexCur,
```
Gerold103, line 4637: There are many other usages of this function which ignore the result. Would it be good to update all of them? Any why did it fail here? Is there code below which depends on having `OP_OpenSpace` and `OP_IteratorOpen` generated?

Gerold103 (commented):
Thanks for the patch!

src/box/sql/where.c
ImeevMA: Thank you for the review! I updated them all, however not sure if this was needed in most cases since in case of failure `is_aborted` is set to `true`. Maybe it would work even if the function would be made `void`.

The `assertion` in the issue was because there were two indexes retrieved - the PK and the secondary index. For PK the commented opcode was created, a and for secondary index no opcode was created, so assertion that checks that the created opcode does not have a comment fails.

Thank you for the review! I've updated them all, but I'm not sure if this is necessary in most cases as `is_aborted` is set to `true` in case of a failure. It might have worked even if the function had been made `void`.

The `assertion` in the issue fails was due to the fact that opcodes for two indexes were created - a PK and a secondary index. A commented opcode was generated for the PK, but no opcode was generated for the secondary index, so the assertion that the generated opcode has no comment fails because the assertion looked at opcode for PK.

Gerold103: Oh, I see. Maybe indeed make the function `void` then and simply drop the assertion `p->aOp[p->nOp - 1].zComment == 0` from `vdbeVComment()`?

ImeevMA: Agree. Fixed!

Gerold103 (approved):
Thanks for patiently trying different approaches!

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #7253, opened 2022-06-09
Old leader might confirm data not present on new leader


sergepetrenko:
<details><summary>10 step reproducer:</summary>

Run 3 tarantools in 3 consoles and execute the following commands in respective consoles:
```lua
--
-- Instance 1
--

-- Step 1
box.cfg{listen = 3301, replication = {3301,3302,3303}, election_mode = 'candidate', election_fencing_enabled = false, replication_timeout = 10}

-- Step 4
box.schema.user.grant('guest', 'replication')
box.schema.space.create('sync', {is_sync = true})
box.space.sync:create_index('pk')

-- Step 5
box.cfg{replication = ""}

-- Step 9
box.space.sync:insert{1}

--
-- Instance 2
--

-- Step 2
box.cfg{listen=3302, replication={3301,3302,3303}, election_mode='voter', replication_timeout = 10}

--
-- Instance 3
--

-- Step 3
box.cfg{listen=3303, replication={3301,3302,3303}, election_mode='voter', replication_timeout = 10}

-- Step 6
box.cfg{replication = {3302, 3303}}

-- Step 7
box.cfg{election_mode = 'candidate'}

-- Step 8
box.ctl.promote() box.error.injection.set("ERRINJ_WAL_DELAY", true) -- important to do this in one line, whitout yields.

-- Step 10
box.error.injection.set("ERRINJ_WAL_DELAY", false)
```

</details>

After you've performed these 10 steps:
 * Instance 3 is leader
 * Instance 1 and instance 2 output:
```lua
tarantool> box.space.sync:get{1}
---
- [1]
...
```
 * Instance 3 output:
```lua
tarantool> box.space.sync:get{1}
---
...
```

I guess it's already obvious what happened, but still some excerpts from the node WALs for convenience.
<details><summary>Instance 1 xlog</summary>

```
---
HEADER:
  lsn: 9
  replica_id: 1
  type: INSERT
  timestamp: 1654768508.6078
BODY:
  space_id: 512
  tuple: [1]
---
HEADER:
  lsn: 10
  replica_id: 1
  type: CONFIRM
  timestamp: 1654768508.608
BODY:
- null
- 1
- 9
...
```

</details>
<details><summary>Instance 2 xlog</summary>

```
---
HEADER:
  lsn: 9
  replica_id: 1
  type: INSERT
  timestamp: 1654768508.6079
BODY:
  space_id: 512
  tuple: [1]
---
HEADER:
  lsn: 10
[...]

PR #7776, opened 2022-10-04
Fix split-brain on a late CONFIRM from an old leader

Gerold103:
The patchset gets rid of a potential split-brain case, which could happen if an old leader would make a txn and CONFIRM it while a new term was already going on #7253.

In the last commit I update test 6036, but it doesn't seem related to #7785. The latter fails extremely rare and in logs I can see something about split brain.

Also the linearizable read test fails occasionally, it is a known problem.

### DIFF
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1374,6 +1374,7 @@ applier_signal_ack(struct applier *applier)
 		applier->ack_msg.txn_last_tm = (r == NULL ? 0 :
 						r->applier_txn_last_tm);
 		applier->ack_msg.vclock_sync = applier->last_vclock_sync;
+		applier->ack_msg.term = box_raft()->term;
 		vclock_copy(&applier->ack_msg.vclock, &replicaset.vclock);
 		cmsg_init(&applier->ack_msg.base, applier->ack_route);
 		cpipe_push(&applier->applier_thread->thread_pipe,
@@ -1397,6 +1398,7 @@ applier_thread_signal_ack(struct cmsg *base)
 	applier->thread.txn_last_tm = msg->txn_last_tm;
 	struct applier_heartbeat *ack = &applier->thread.next_ack;
 	ack->vclock_sync = msg->vclock_sync;
+	ack->term = msg->term;
 	vclock_copy(&ack->vclock, &msg->vclock);
 }
 
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -699,13 +699,11 @@ applier_register(struct applier *applier, bool was_anon)
 	/* Send REGISTER request */
 	struct iostream *io = &applier->io;
 	struct xrow_header row;
-
-	memset(&row, 0, sizeof(row));
-	/*
-	 * Send this instance's current vclock together
-	 * with REGISTER request.
-	 */
-	xrow_encode_register(&row, &INSTANCE_UUID, box_vclock);
+	struct register_request req;
+	memset(&req, 0, sizeof(req));
+	req.instance_uuid = INSTANCE_UUID;
+	vclock_copy(&req.vclock, box_vclock);
+	xrow_encode_register_xc(&row, &req);
 	row.type = IPROTO_REGISTER;
 	coio_write_xrow(io, &row);
 
@@ -732,8 +730,11 @@ applier_join(struct applier *applier)
 	struct iostream *io = &applier->io;
 	struct xrow_header row;
 	uint64_t row_count;
-
-	xrow_encode_join_xc(&row, &INSTANCE_UUID);
+	struct join_request req;
+	memset(&req, 0, sizeof(req));
+	req.instance_uuid = INSTANCE_UUID;
+	req.version_id = tarantool_version_id();
+	xrow_encode_join_xc(&row, &req);
 	coio_write_xrow(io, &row);
 
 	applier_set_state(applier, APPLIER_WAIT_SNAPSHOT);
@@ -1963,23 +1964,23 @@ applier_subscribe(struct applier *applier)
 	struct iostream *io = &applier->io;
 	struct ibuf *ibuf = &applier->ibuf;
 	struct xrow_header row;
-	struct tt_uuid cluster_id = uuid_nil;
 
-	struct vclock vclock;
-	vclock_create(&vclock);
-	vclock_copy(&vclock, &replicaset.vclock);
-
+	struct subscribe_request req;
+	memset(&req, 0, sizeof(req));
+	vclock_copy(&req.vclock, &replicaset.vclock);
 	ERROR_INJECT(ERRINJ_REPLICASET_VCLOCK, {
-		vclock_create(&vclock);
+		vclock_create(&req.vclock);
 	});
-
+	req.replicaset_uuid = REPLICASET_UUID;
+	req.instance_uuid = INSTANCE_UUID;
+	req.version_id = tarantool_version_id();
+	req.is_anon = replication_anon;
 	/*
 	 * Stop accepting local rows coming from a remote
 	 * instance as soon as local WAL starts accepting writes.
 	 */
-	uint32_t id_filter = box_is_orphan() ? 0 : 1 << instance_id;
-	xrow_encode_subscribe_xc(&row, &REPLICASET_UUID, &INSTANCE_UUID,
-				 &vclock, replication_anon, id_filter);
+	req.id_filter = box_is_orphan() ? 0 : 1 << instance_id;
+	xrow_encode_subscribe_xc(&row, &req);
 	coio_write_xrow(io, &row);
 
 	/* Read SUBSCRIBE response */
@@ -1995,29 +1996,29 @@ applier_subscribe(struct applier *applier)
 		 * In case of successful subscribe, the server
 		 * responds with its current vclock.
 		 *
-		 * Tarantool > 2.1.1 also sends its cluster id to
+		 * Tarantool > 2.1.1 also sends its replicaset UUID to
 		 * the replica, and replica has to check whether
-		 * its and master's cluster ids match.
+		 * its and master's replicaset UUIDs match.
 		 */
-		xrow_decode_subscribe_response_xc(&row, &cluster_id,
-					&applier->remote_vclock_at_subscribe);
+		struct subscribe_response rsp;
+		xrow_decode_subscribe_response_xc(&row, &rsp);
 		applier->instance_id = row.replica_id;
+		vclock_copy(&applier->remote_vclock_at_subscribe, &rsp.vclock);
 		/*
-		 * If master didn't send us its cluster id
+		 * If master didn't send us its replicaset UUID
 		 * assume that it has done all the checks.
-		 * In this case cluster_id will remain zero.
+		 * In this case replicaset UUID will remain zero.
 		 */
-		if (!tt_uuid_is_nil(&cluster_id) &&
-		    !tt_uuid_is_equal(&cluster_id, &REPLICASET_UUID)) {
+		if (!tt_uuid_is_nil(&rsp.replicaset_uuid) &&
+		    !tt_uuid_is_equal(&rsp.replicaset_uuid, &REPLICASET_UUID)) {
 			tnt_raise(ClientError, ER_REPLICASET_UUID_MISMATCH,
-				  tt_uuid_str(&cluster_id),
+				  tt_uuid_str(&rsp.replicaset_uuid),
 				  tt_uuid_str(&REPLICASET_UUID));
 		}
-
 		say_info("subscribed");
 		say_info("remote vclock %s local vclock %s",
-			 vclock_to_string(&applier->remote_vclock_at_subscribe),
-			 vclock_to_string(&vclock));
+			 vclock_to_string(&rsp.vclock),
+			 vclock_to_string(&req.vclock));
 	}
 	/*
 	 * Tarantool < 1.6.7:
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -207,10 +207,8 @@ applier_thread_writer_f(va_list ap)
 		try {
 			applier->thread.has_acks_to_send = false;
 			struct xrow_header xrow;
-			uint64_t vclock_sync = applier->thread.last_vclock_sync;
-			struct vclock *ack_vclock = &applier->thread.ack_vclock;
-			xrow_encode_heartbeat_xc(&xrow, ack_vclock,
-						 vclock_sync);
+			xrow_encode_applier_heartbeat_xc(
+				&xrow, &applier->thread.next_ack);
 			xrow.tm = applier->thread.txn_last_tm;
 			coio_write_xrow(&applier->io, &xrow);
 			ERROR_INJECT(ERRINJ_APPLIER_SLOW_ACK, {
@@ -799,7 +797,9 @@ applier_parse_tx_row(struct applier_tx_row *tx_row)
 			diag_raise();
 		}
 	} else if (type == IPROTO_OK) {
-		xrow_decode_heartbeat_xc(row, NULL, &tx_row->req.vclock_sync);
+		struct relay_heartbeat req;
+		xrow_decode_relay_heartbeat_xc(row, &req);
+		tx_row->req.vclock_sync = req.vclock_sync;
 	} else {
 		tnt_raise(ClientError, ER_UNKNOWN_REQUEST_TYPE, type);
 	}
@@ -1395,8 +1395,9 @@ applier_thread_signal_ack(struct cmsg *base)
 	fiber_cond_signal(&applier->thread.writer_cond);
 	applier->thread.has_acks_to_send = true;
 	applier->thread.txn_last_tm = msg->txn_last_tm;
-	applier->thread.last_vclock_sync = msg->vclock_sync;
-	vclock_copy(&applier->thread.ack_vclock, &msg->vclock);
+	struct applier_heartbeat *ack = &applier->thread.next_ack;
+	ack->vclock_sync = msg->vclock_sync;
+	vclock_copy(&ack->vclock, &msg->vclock);
 }
 
 /**
@@ -1883,7 +1884,7 @@ applier_thread_attach_applier(struct cbus_call_msg *base)
 	applier_thread_ibuf_init(applier);
 	applier_thread_msgs_init(applier);
 	applier_thread_fiber_init(applier);
-	applier->thread.last_vclock_sync = 0;
+	memset(&applier->thread.next_ack, 0, sizeof(applier->thread.next_ack));
 
 	return 0;
 }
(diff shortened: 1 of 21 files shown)


test/replication-luatest/gh_7253_election_long_wal_write_test.lua
```diff
@@ -59,12 +59,39 @@ local function server_wait_wal_is_blocked(server)
...
+        end
+        if box.info.synchro.queue.busy then
+            return
+        end
+        -- Allow 1 more WAL write.
+        box.error.injection.set('ERRINJ_WAL_DELAY_COUNTDOWN', 0)
+        box.error.injection.set('ERRINJ_WAL_DELAY', false)
+        error('Not busy yet')
+    end)
+end
```
sergepetrenko, line 85: Let's move this to luatest_helpers. It's quite obvious that this function is needed pretty often:
```
s.petrenko@spetrenko:~/Source/sptnt$ grep -r 'function play_wal_until_synchro_queue_is_busy' ./test
./test/replication-luatest/gh_7253_election_long_wal_write_test.lua:local function play_wal_until_synchro_queue_is_busy(server)
./test/replication-luatest/gh_7086_box_issue_promote_assert_test.lua:local function play_wal_until_synchro_queue_is_busy(server)
./test/replication-luatest/gh_6033_box_promote_demote_test.lua:local function play_wal_until_synchro_queue_is_busy(server)
./test/replication-luatest/gh_6842_qsync_applier_order_test.lua:local function play_wal_until_synchro_queue_is_busy(server)
```

test/replication-luatest/gh_7253_election_long_wal_write_test.lua
```diff
@@ -245,5 +276,267 @@ g.test_vote_during_txn_wal_write = function(g)
...
+    local server3_uri = server.build_instance_uri('server3')
+    local server4_uri = server.build_instance_uri('server4')
+    local server5_uri = server.build_instance_uri('server5')
+    --
+    -- Build the topology:
+    --   server4 <- fullmesh(server1, server2, server3)
+    --   server3 <-> server5
+    --
+    g.server4 = g.cluster:build_and_add_server({
+        alias = 'server4', box_cfg = table.deepcopy(g.server3.box_cfg)
```
sergepetrenko, line 403: Deepcopy is done by `cluster:build_server` anyway, you don't have to repeat it here

sergepetrenko (changes requested):
Hi! Thanks for the patch!

src/box/applier.cc
```diff
@@ -1375,6 +1375,7 @@ applier_signal_ack(struct applier *applier)
 		applier->ack_msg.txn_last_tm = (r == NULL ? 0 :
 						r->applier_txn_last_tm);
 		applier->ack_msg.vclock_sync = applier->last_vclock_sync;
+		applier->ack_msg.term = box_raft()->volatile_term;
```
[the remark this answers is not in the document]
Gerold103: I don't remember a discussion. Maybe you talked to Vova. My opinion here is that adding `tm` and `replica_id` to request wouldn't be better, because:
- It would be duplication of the members. You would have `tm` and `replica_id` both in `xrow_header` and in `..._request` structs.
- Duplication might lead to slightly bigger parent objects. For example, `relay.last_recv_ack` would store `tm` field even though it is not needed here. The timestamp is decoded and used in `relay_reader_f()` right away, doesn't need a long storage.

Doesn't mean it can't be done, but it would simply produce other questions. Can do it if you want anyway.

src/box/applier.cc
```diff
@@ -207,10 +207,8 @@ applier_thread_writer_f(va_list ap)
 		try {
 			applier->thread.has_acks_to_send = false;
 			struct xrow_header xrow;
-			uint64_t vclock_sync = applier->thread.last_vclock_sync;
-			struct vclock *ack_vclock = &applier->thread.ack_vclock;
-			xrow_encode_heartbeat_xc(&xrow, ack_vclock,
-						 vclock_sync);
+			xrow_encode_upstream_ack_xc(&xrow,
+						    &applier->thread.next_ack);
 			xrow.tm = applier->thread.txn_last_tm;
```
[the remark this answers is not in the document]
Gerold103: I don't remember a discussion. Maybe you talked to Vova. My opinion here is that adding `tm` and `replica_id` to request wouldn't be better, because:
- It would be duplication of the members. You would have `tm` and `replica_id` both in `xrow_header` and in `..._request` structs.
- Duplication might lead to slightly bigger parent objects. For example, `relay.last_recv_ack` would store `tm` field even though it is not needed here. The timestamp is decoded and used in `relay_reader_f()` right away, doesn't need a long storage.

Doesn't mean it can't be done, but it would simply produce other questions. Can do it if you want anyway.

src/box/applier.cc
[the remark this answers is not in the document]
Gerold103: The reason is simpler - I did it out of habit from usage inside the instance where we rely on the volatile term. A persistent term is surely not worse functionally and is consistent with existing broadcasts. Changed it to non-volatile term now.

test/replication-luatest/gh_7253_election_long_wal_write_test.lua
Gerold103: Thanks, done in a separate commit.

test/replication-luatest/gh_7253_election_long_wal_write_test.lua
Gerold103: Indeed, dropped the `deepcopy`.

src/box/applier.cc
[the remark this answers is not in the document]
sergepetrenko: I'm fine with the current approach, thanks for the answer.

src/box/applier.cc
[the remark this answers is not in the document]
sergepetrenko: Nice.

src/box/applier.cc
```diff
@@ -699,13 +699,12 @@ applier_register(struct applier *applier, bool was_anon)
...
 	struct xrow_header row;
-
 	memset(&row, 0, sizeof(row));
-	/*
-	 * Send this instance's current vclock together
-	 * with REGISTER request.
-	 */
-	xrow_encode_register(&row, &INSTANCE_UUID, box_vclock);
+	struct register_request req;
+	memset(&req, 0, sizeof(req));
```
sergepetrenko, line 702: You already do a memset inside `xrow_encode_replication_request`. Let's not duplicate it here and in other places below.

src/box/applier.cc
```diff
@@ -1995,29 +1997,30 @@ applier_subscribe(struct applier *applier)
...
-		 * Tarantool > 2.1.1 also sends its cluster id to
+		 * Tarantool > 2.1.1 also sends its replicaset UUID to
 		 * the replica, and replica has to check whether
-		 * its and master's cluster ids match.
+		 * its and master's replicaset UUIDs match.
 		 */
-		xrow_decode_subscribe_response_xc(&row, &cluster_id,
-					&applier->remote_vclock_at_subscribe);
+		struct subscribe_response rsp;
+		memset(&rsp, 0, sizeof(rsp));
```
sergepetrenko, line 2008: Let's move the memset to `xrow_decode_subscribe_response`. And do similarly for other `xrow_decode_...` helpers.

It would be nicer to have it in `xrow_decode_replication_request`, but it seems this can't be done now.

src/box/xrow.c
```diff
@@ -1705,70 +1705,79 @@ xrow_decode_ballot(const struct xrow_header *row, struct ballot *ballot)
...
-	data = mp_encode_vclock_ignore0(data, vclock);
-	assert(data <= buf + size);
-	row->body[0].iov_base = buf;
-	row->body[0].iov_len = (data - buf);
-	row->bodycnt = 1;
-	row->type = IPROTO_REGISTER;
-	return 0;
-}
+/**
+ * A template which can represent any replication request - join, register, ack,
```
sergepetrenko, line 1709: I'd rather not mention acks here. You create a separate struct for them right in the next commit.

src/box/xrow.c
```diff
@@ -1705,70 +1705,79 @@ xrow_decode_ballot(const struct xrow_header *row, struct ballot *ballot)
...
 		return -1;
 	}
+	uint32_t id_filter = req->id_filter != NULL ? *req->id_filter : 0;
+	uint32_t map_size = 0;
+	map_size += req->replicaset_uuid != NULL;
+	map_size += req->instance_uuid != NULL;
+	map_size += req->vclock != NULL;
+	map_size += id_filter != 0;
+	map_size += req->is_anon != NULL;
+	map_size += req->version_id != NULL;
```
sergepetrenko, line 1752: Fixmap is suitable for maps with up to 15 entries, so I would suggest reserving one byte for a map and then coming back and filling in the map size after the element encoding is done. So that you don't have to calculate map size separately.
We could return to your current approach when and if replication_request starts holding more than 15 fields.
WDYT?

src/box/xrow.c
```diff
@@ -1705,70 +1705,79 @@ xrow_decode_ballot(const struct xrow_header *row, struct ballot *ballot)
...
+	}
+	if (req->version_id != NULL) {
+		data = mp_encode_uint(data, IPROTO_SERVER_VERSION);
+		data = mp_encode_uint(data, *req->version_id);
+	}
+	if (req->is_anon != NULL) {
+		data = mp_encode_uint(data, IPROTO_REPLICA_ANON);
+		data = mp_encode_bool(data, *req->is_anon);
+	}
+	if (id_filter != 0) {
```
sergepetrenko, line 1776: Let's encode it unconditionally (when `req->id_filter` isn't NULL). This'd be consistent with other fields. I don't see why we are trying to save a couple of bytes on a subscribe request.

src/box/xrow.c
```diff
@@ -1798,19 +1806,18 @@ xrow_decode_subscribe(const struct xrow_header *row,
...
+	if (req->instance_uuid != NULL)
+		*req->instance_uuid = uuid_nil;
+	if (req->vclock != NULL)
+		vclock_create(req->vclock);
+	if (req->version_id != NULL)
+		*req->version_id = 0;
+	if (req->is_anon != NULL)
+		*req->is_anon = false;
+	if (req->id_filter != NULL)
+		*req->id_filter = 0;
```
sergepetrenko, line 1820: If you apply my suggestion with memsetting the `smth_request` in `xrow_decode_smth`, you won't have to nullify the fields here explicitly.

src/box/applier.cc
```diff
@@ -800,7 +798,9 @@ applier_parse_tx_row(struct applier_tx_row *tx_row)
 			diag_raise();
 		}
 	} else if (type == IPROTO_OK) {
-		xrow_decode_heartbeat_xc(row, NULL, &tx_row->req.vclock_sync);
+		struct downstream_ack_request last_recv_ack;
+		xrow_decode_downstream_ack_xc(row, &last_recv_ack);
```
sergepetrenko, line 802: Maybe switch to `struct relay_ack_request` and `struct applier_ack_request`?

Why does an applier decode a "downstream ack" when this ack came from the upstream?
IMO "relay ack" and "applier ack" would be simpler to understand. You even name the 'upstream ack' a 'downstream ack' by mistake in the last commit's mesage.

sergepetrenko: Or probably they should be 'relay_heartbeat' and 'applier_heartbeat'. Because relay doesn't have acks.

src/box/relay.cc
```diff
@@ -544,10 +539,10 @@ relay_final_join(struct iostream *io, uint64_t sync,
 
 	/*
 	 * Save the first vclock as 'received'. Because firstly, it was really
-	 * received. Secondly, recv_vclock is used by recovery restart and must
-	 * always be valid.
+	 * received. Secondly, last received vclock is used by recovery restart
```
sergepetrenko, line 544: By the way, we don't have `relay_restart_recovery` anymore, so you may shorten this comment.

src/box/relay.cc
```diff
@@ -1068,10 +1066,10 @@ relay_subscribe(struct replica *replica, struct iostream *io, uint64_t sync,
 	vclock_copy(&relay->local_vclock_at_subscribe, &replicaset.vclock);
 	/*
 	 * Save the first vclock as 'received'. Because firstly, it was really
-	 * received. Secondly, recv_vclock is used by recovery restart and must
-	 * always be valid.
+	 * received. Secondly, last received vclock is used by recovery restart
+	 * and must always be valid.
```
sergepetrenko, line 1082: Same about recovery restart.

sergepetrenko (changes requested):
Thanks for the fixes!

src/box/applier.cc
```diff
@@ -732,8 +731,11 @@ applier_join(struct applier *applier)
 	struct iostream *io = &applier->io;
 	struct xrow_header row;
 	uint64_t row_count;
-
-	xrow_encode_join_xc(&row, &INSTANCE_UUID);
+	struct join_request req;
```
sergos, line 732: I would rather propose to 
```
const struct join_request req = {
        .instance_uuid = INSTANCE_UUID,
        .version_id = tarantool_version_id()
}
```

src/box/applier.cc
Gerold103: I prefer to keep initialization with zeros. I think structs always have to be initialized either via memset or via `= {.field1 = ..., .field2 = ..., ...}`. Especially if they are used as in-parameters to a function (so the function expects them to be fully valid and initialized). Right now all fields are initialized explicitly indeed, but if a new field would be added, it wouldn't be so. You would need to walk all the usage places and initialize it with 0 explicitly if it is optional or with some meaningful value. If you forget to do so (code will compile anyway), you will get garbage in this field which will occasionally lead to failures.

src/box/applier.cc
Gerold103: It is not possible in C++ (which is applier's language).

src/box/applier.cc
Gerold103: Thanks, done. In `replication_request` it can be done, but it wouldn't be much better - pass `sizeof` as one of the params to make `memset` inside.

src/box/xrow.c
Gerold103: Removed.

src/box/xrow.c
Gerold103: I don't mind, changed to fixed-size map.

src/box/xrow.c
Gerold103: The intention was not to change the behaviour - previously we didn't encode the filter when it was 0. But I don't mind to alter it. Done.

src/box/xrow.c
Gerold103: It works for now because `uuid_nil` and `vclock_create` are zeros. I removed these initializations for now, but in future it might complicate things.

src/box/applier.cc
Gerold103: > Why does an applier decode a "downstream ack" when this ack came from the upstream?

It comes from downstream (`relay -> applier` is downstream).

> IMO "relay ack" and "applier ack" would be simpler to understand.

Apparently it is not much better. Besides, it is unclear whether `applier_ack` comes from applier or to applier?

> You even name the 'upstream ack' a 'downstream ack' by mistake in the last commit's mesage

It was a typo. Thanks for noticing, fixed.

> Or probably they should be 'relay_heartbeat' and 'applier_heartbeat'. Because relay doesn't have acks.

On the other hand applier doesn't have heartbeats. Anyway, I renamed them both to `..._heartbeat`.

src/box/relay.cc
Gerold103: Thanks for noticing, updates some comments in a separate commit.

src/box/relay.cc
Gerold103: Ditto.

src/box/applier.cc
sergepetrenko: > It comes from downstream (relay -> applier is downstream).

It's a downstream from the master's point of view. But for a replica, it's an upstream. And replica sends ACKs to upstream, and receives heartbeats from the upstream. So yeah, this is complicated regardless of the name we choose.

Anyway, thanks for the fixes!

src/box/xrow.c
sergepetrenko: > It works for now because uuid_nil and vclock_create are zeros. I removed these initializations for now, but in future it might complicate things.

Yep. Let's try to keep things this way then.

src/box/xrow.c
sergepetrenko: Thanks

src/box/applier.cc
sergepetrenko: > Is it the row memset you're referring to?

Yep. I got confused here.
Anyway, there is an extraneous `memset` 2 lines above. Not in other places though.

sergepetrenko (commented):
Thanks for the fixes!
One unresolved comment remains, everything else looks good.

src/box/applier.cc
Gerold103: Thanks, done.

sergepetrenko (approved):
Thanks for the patch!
LGTM.

src/box/applier.cc
sergos: oops. I tested it on MacOS and it appeared clang supports it back to `--std=c++98`, while g++ on Linux complains with `sorry, unimplemented: non-trivial designated initializers not supported`.

test/luatest_helpers/misc.lua
```diff
@@ -22,6 +22,7 @@ local function skip_if_enterprise()
 end
 
 return {
+    is_debug_build = is_debug_build,
```
sergos, line 25: Mmm? Shall we rework all ERRNIJ-based luatests with the same assertion?

sergos (approved):
Just a couple of minor comments.
Otherwise LGTM, thank you!

test/luatest_helpers/misc.lua
Gerold103: I wouldn't bother. Might ask to do it in new tests. Besides, it makes sense to do not in tests themselves, which anyway usually call `skip_if_not_debug()`, but rather in luatest helpers which use errinjs. From whose API it is not always clear that they work only in debug.

src/lib/raft/raft.c
```diff
@@ -1315,6 +1314,15 @@ raft_new_term(struct raft *raft)
 	raft_sm_schedule_new_term(raft, raft->volatile_term + 1);
 }
 
+void
+raft_process_term(struct raft *raft, uint64_t term, uint32_t source)
+{
+	if (term <= raft->volatile_term)
+		return;
+	say_info("RAFT: received a newer term from %u", (unsigned)source);
```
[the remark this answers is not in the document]
Gerold103: New term is reported by `raft_sm_schedule_new_term()` which is called on the next line. Old term was reported in older logs. And about *"replica"* word - I intentionally omitted it. Wanted to be consistent with many other logs in raft.c which refer to nodes just as `%u`. These, in turn, were written that way to keep the logs short but informative. Makes it easier to read them IMO.

Gerold103:
Not cherry-picked to 2.10, because it intersects with lineariazable reads changes, which also were not cherry-picked, because their tests depend on proxy luatest module.

Let me know if we want to integrate this all to 2.10.

PR #7777, opened 2022-10-05
New rules to determine type of result of NULLIF() and CASE

### LINKED ISSUE #6990: SQL SUM and AVG functions don't work for CASE expressions that may return NULL in their branches
**Bug description**

subj

* OS: Linux
* OS Version: Centos 8
* Architecture: amd64

Tarantool 2.10.0-beta2-207-ga5557281c
Target: Linux-x86_64-Release
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: /usr/lib64/ccache/cc /usr/lib64/ccache/c++
C_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type
CXX_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type

**Steps to reproduce**

Run tarantool, box.cfg{}

Run query:
```lua
box.execute([[select sum(case when 1 = 1 then 1 else null end)]])
```

and

```lua
box.execute([[select avg(case when 1 = 1 then 1 else null end)]])
```

**Actual behavior**

```
---
- null
- 'Failed to execute SQL statement: wrong arguments for function SUM()'
...
```

and

```
---
- null
- 'Failed to execute SQL statement: wrong arguments for function AVG()'
...
```

**Expected behavior**

```
---
- metadata:
  - name: COLUMN_1
    type: integer
  rows:
  - [1]
...
```

ImeevMA:
This patch-set introduces new rules to determine type of result of NULLIF() built-in function and CASE operator.

Closes #6990

### DIFF
--- a/changelogs/unreleased/gh-6990-type-of-CASE-operation.md
+++ b/changelogs/unreleased/gh-6990-type-of-CASE-operation.md
@@ -1,0 +1,3 @@
+## feature/sql
+
+* New rules are applied to determine the type of CASE operation (gh-6990).
--- a/changelogs/unreleased/gh-6989-type-of-NULLIF.md
+++ b/changelogs/unreleased/gh-6989-type-of-NULLIF.md
@@ -1,0 +1,4 @@
+## feature/sql
+
+* Now the type of result of NULLIF() is the same as the type of the first
+  argument (gh-6989).
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -45,6 +45,37 @@ static void exprCodeBetween(Parse *, Expr *, int,
 			    void (*)(Parse *, Expr *, int, int), int);
 static int exprCodeVector(Parse * pParse, Expr * p, int *piToFree);
 
+/**
+ * Determine the highest type between the given type and the type of the given
+ * expression.
+ */
+static enum field_type
+sql_highest_type(enum field_type a, struct Expr *expr)
+{
+	if (a == FIELD_TYPE_ANY || expr->op == TK_VARIABLE)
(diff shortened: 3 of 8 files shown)


changelogs/unreleased/gh-6990-type-of-CASE-operation.md
```diff
@@ -0,0 +1,3 @@
+## feature/sql
```
Gerold103, line 1: In the docbot requests you are saying *"after this patch will be ..."*. Doc requests are not bound to a patch. They just happen to be in the commit messages. The text in doc requests should be the same like if you would create a doc ticket manually without talking about any patches. Also they are created after stuff is done, so the facts should be described like they have already happened. For example, not *"new rules will be applied"*, but *"new rules are applied"*.

Gerold103 (approved):
Thanks for the patch!

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7782, opened 2022-10-06
app: fix crash on logrotate and immediate exit

### LINKED ISSUE #4450: test: log_destroy() sometimes segfaults on app-tap/logger.test.lua
Tarantool version: 2.3.0-61-g5b9f207df
OS version: Gentoo Linux.

How to reproduce:

```sh
$ cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_BACKTRACE=ON -DENABLE_DIST=ON && make -j
$ TEST_RUN_TESTS="$(yes app-tap/logger.test.lua | head -n 1000)" make test # several times
```

Backtrace:

```
(gdb) bt
#0  0x00007fe9392d9f1b in raise () from /lib64/libc.so.6
#1  0x00007fe9392c3535 in abort () from /lib64/libc.so.6
#2  0x0000560611c11ef7 in sig_fatal_cb (signo=11, siginfo=0x7fff252656f0, context=0x7fff252655c0) at /home/alex/p/tarantool-meta/r/t-2/src/main.cc:300
#3  <signal handler called>
#4  0x0000560611d52d13 in fiber_yield () at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/fiber.c:478
#5  0x0000560611d52ed1 in fiber_yield_timeout (delay=3153600000) at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/fiber.c:522
#6  0x0000560611d5b286 in fiber_cond_wait_timeout (c=0x5606122037a0 <log_std+96>, timeout=3153600000) at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/fiber_cond.c:107
#7  0x0000560611d5b354 in fiber_cond_wait (c=0x5606122037a0 <log_std+96>) at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/fiber_cond.c:117
#8  0x0000560611d50af4 in log_destroy (log=0x560612203740 <log_std>) at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/say.c:1178
#9  0x0000560611d4e9d5 in say_logger_free () at /home/alex/p/tarantool-meta/r/t-2/src/lib/core/say.c:714
#10 0x0000560611c12fda in tarantool_free () at /home/alex/p/tarantool-meta/r/t-2/src/main.cc:639
#11 0x0000560611c138cd in main (argc=1, argv=0x560612859f18) at /home/alex/p/tarantool-meta/r/t-2/src/main.cc:864
```

nshy:
`say_logrotate` does not rotate logs synchronously. It posts tasks to
coio which executes them in it's pool thread. On application exit we
destroy logs calling `log_destroy`. This function waits for rotate task
to finish because if it will free resources immediately then unfinished
rotate task can have use-after-free issues. Waiting crashes because at
this moment event loop is not running which is required for
`fiber_cond_wait` to work.

Note that if there is no crash then we will hang forever waiting in
`log_destroy` because event loop is not running and
`log_rotate_async_cb` will never be executed.

Let's use mutexes and conditions instead. It solves both problems with
crash and hang. Thanks Vladimir Davydov for idea.

Fixes #4450

NO_DOC=bugfix

### DIFF
--- a/src/lib/core/say.c
+++ b/src/lib/core/say.c
@@ -342,10 +342,15 @@ static int
 logrotate_cb(struct coio_task *ptr)
 {
 	struct rotate_task *task = (struct rotate_task *) ptr;
-	if (log_rotate(task->log) < 0) {
+	struct log *log = task->log;
+	if (log_rotate(log) < 0)
 		diag_log();
-	}
-	ev_async_send(task->loop, &task->log->log_async);
+	tt_pthread_mutex_lock(&log->rotate_mutex);
+	assert(log->rotating_threads > 0);
+	log->rotating_threads--;
+	if (log->rotating_threads == 0)
+		tt_pthread_cond_signal(&log->rotate_cond);
+	tt_pthread_mutex_unlock(&log->rotate_mutex);
 	return 0;
 }
 
@@ -356,16 +361,6 @@ logrotate_cleanup_cb(struct coio_task *ptr)
 	coio_task_destroy(&task->base);
 	free(task);
 	return 0;
-}
-
-static void
-log_rotate_async_cb(struct ev_loop *loop, struct ev_async *watcher, int events)
-{
-	(void)loop;
-	(void)events;
-	struct log *log = container_of(watcher, struct log, log_async);
-	log->rotating_threads--;
-	fiber_cond_signal(&log->rotate_cond);
 }
 
 void
@@ -385,8 +380,9 @@ say_logrotate(struct ev_loop *loop, struct ev_signal *w, int revents)
 			diag_log();
 			continue;
 		}
-		ev_async_start(loop(), &log->log_async);
+		tt_pthread_mutex_lock(&log->rotate_mutex);
 		log->rotating_threads++;
+		tt_pthread_mutex_unlock(&log->rotate_mutex);
 		coio_task_create(&task->base, logrotate_cb, logrotate_cleanup_cb);
 		task->log = log;
 		task->loop = loop();
@@ -672,8 +668,8 @@ log_create(struct log *log, const char *init_str, int nonblock)
 	log->format_func = NULL;
 	log->level = S_INFO;
 	log->rotating_threads = 0;
-	fiber_cond_create(&log->rotate_cond);
-	ev_async_init(&log->log_async, log_rotate_async_cb);
+	tt_pthread_mutex_init(&log->rotate_mutex, NULL);
+	tt_pthread_cond_init(&log->rotate_cond, NULL);
 	setvbuf(stderr, NULL, _IONBF, 0);
 
 	if (init_str != NULL) {
@@ -1252,8 +1248,10 @@ void
 log_destroy(struct log *log)
 {
 	assert(log != NULL);
+	tt_pthread_mutex_lock(&log->rotate_mutex);
 	while(log->rotating_threads > 0)
-		fiber_cond_wait(&log->rotate_cond);
+		tt_pthread_cond_wait(&log->rotate_cond, &log->rotate_mutex);
+	tt_pthread_mutex_unlock(&log->rotate_mutex);
 	pm_atomic_store(&log->type, SAY_LOGGER_BOOT);
 
 	if (log->fd != -1)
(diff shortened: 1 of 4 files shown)


src/lib/core/coio_task.c
```diff
@@ -150,11 +151,13 @@ coio_enable(void)
 	ev_async_init(&coio_manager.coio_async, coio_async_cb);
 
 	ev_async_start(loop(), &coio_manager.coio_async);
+	coio_is_running = true;
```
locker, line 154: `coio_enable` may be called from any thread, not just tx (note `coio_manager` is thread-local) so setting the global flag here is incorrect.

locker:
Discussed f2f. It looks like we don't really need `fiber_cond` here: a `pthread_cond` would be enough, because we signal it rarely (on log rotate) and check it only on shutdown. Let's try to protect `rotating_threads` with a `pthread_mutex` and use a `pthread_cond` instead of `fiber_cond` to signal when it hits zero.

src/lib/core/coio_task.c
nshy: Got rid of this flag in new version.

src/lib/core/say.c
```diff
@@ -1252,8 +1252,10 @@ void
 log_destroy(struct log *log)
 {
 	assert(log != NULL);
-	while(log->rotating_threads > 0)
-		fiber_cond_wait(&log->rotate_cond);
+	if (coio_is_running) {
```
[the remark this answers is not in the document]
nshy: Task is submitted before `coio_shutdown`. I just thought task can be not executed and we got a hang. Actually `eio` thread pool executes all tasks on shutdown.

src/lib/core/say.c
```diff
@@ -342,10 +342,14 @@ static int
...
 {
 	struct rotate_task *task = (struct rotate_task *) ptr;
-	if (log_rotate(task->log) < 0) {
+	struct log *log = task->log;
+	if (log_rotate(log) < 0)
 		diag_log();
-	}
-	ev_async_send(task->loop, &task->log->log_async);
+	tt_pthread_mutex_lock(&log->mutex);
+	log->rotating_threads--;
```
locker, line 349: Nit: It'd be nice to have an assertion here that `rotating_threads` never goes below 0.

src/lib/core/say.c
```diff
@@ -1252,17 +1247,17 @@ void
...
+	tt_pthread_mutex_unlock(&log->mutex);
 	pm_atomic_store(&log->type, SAY_LOGGER_BOOT);
 
 	if (log->fd != -1)
 		close(log->fd);
 	free(log->syslog_ident);
 	free(log->path);
 	rlist_del_entry(log, in_log_list);
-	ev_async_stop(loop(), &log->log_async);
-	fiber_cond_destroy(&log->rotate_cond);
```
locker, line 1265: You should destroy the mutex and the condition variable here.

src/lib/core/say.h
```diff
@@ -171,17 +171,12 @@ struct log {
...
-	ev_async log_async;
-	/**
-	 * Conditional variable securing variable below
-	 * from concurrent usage.
-	 */
-	struct fiber_cond rotate_cond;
 	/** Counter identifying number of threads executing log_rotate. */
 	int rotating_threads;
+	/** Mutex for critical sections. */
+	pthread_mutex_t mutex;
```
locker, line 177: Nit: Let's rename to `rotate_mutex` to match `rotate_cond` - after all it's used solely for log rotation. Also, please add a comment to `rotating_threads` that it's protected by this mutex.

locker (changes requested):

src/lib/core/say.c
nshy: Fixed.

src/lib/core/say.h
nshy: Done.

src/lib/core/say.h
```diff
@@ -172,16 +172,14 @@ struct log {
 	/* Application identifier used to group syslog messages. */
 	char *syslog_ident;
 	/**
-	 * Used to wake up the main logger thread from a eio thread.
+	 * Counter identifying number of threads executing log_rotate.
+	 * Protected by @mutex as it is accessed from different threads.
```
locker, line 176: Nit: `rotate_mutex`. Also, we don't usually use `@` to refer to variables and functions. Let's also elaborate the comment to `rotate_mutex` - "mutex for critical sections" sounds too vague.

nshy: Done.

locker (approved):

locker:
integration/tarantool-python failure is unrelated to this PR.

locker:
Cherry-picked to 2.10.

Issue #4231, opened 2019-05-20
box.execute should be immutable function 


R-omk:
Tarantool version:
version: 2.2.0-283-gd11b552e3

```
local box_execute = box.execute

box.cfg {
-- config
}

box_execute("SELECT 1")
```

it lead to error:
```
'builtin/box/load_cfg.lua:479: bad argument #1 to ''pairs'' (table expected, got  nil)'
```

https://github.com/tarantool/tarantool/blob/d11b552e3/src/box/lua/load_cfg.lua#L479

Gerold103:
You do not have guarantees about box method before box.cfg is called. You can only count on box.cfg, that is all.

kostja:
I asked @R-omk to report this since I believe the following should be done:
- we currently look at box.execute and change its type to implement automatic invocation of box.cfg{}. We should think of other way of doing this, not replacing box.execute()

Gerold103:
Ok, you believe, but what are the reasons?

Gerold103:
We can do that, just move current implementation of `box.execute` to something like `box.do_execute`, and inside `box.execute()` always call `do_execute`. The latter will be replaced on first `box.cfg`. But it is a perf penalty, obviously.

kostja:
We can well use our own internal flag to look at, and not make the life of a user more difficult just because we can.

kostja:
OK, first of all, we agreed a while ago that we should invoke box.cfg{} automatically and not just for SQL. So it's not only box.execute() that is affected. The automatic invocation as it is currently in 2.x is unfinished.

Totktonada:
One more similar case:

```
tarantool> box_cfg = box.cfg
--- 
... 
tarantool> box_cfg()
<...>
tarantool> box_cfg()
--- 
- error: 'builtin/box/load_cfg.lua:19: builtin/box/load_cfg.lua:547: bad argument
    #1 to ''pairs'' (table expected, got nil)'
... 
```

Totktonada:
[Cherry-picked](https://lists.tarantool.org/pipermail/tarantool-patches/2020-June/017944.html) box configuration fix for 1.10 and pushed it in [1.10.6-52-g9dd9b10f0](https://github.com/tarantool/tarantool/commit/9dd9b10f0047eeb31dea65615bef32577b22a4ae).

PR #7786, opened 2022-10-07
box: drop 'execute' field from uninitialized box

### LINKED ISSUE #4726: Disable implicit box.cfg{} in sql
In #3266 there were added implicit empty `box.cfg{}` to `box.execute`

At first, this behavior differs from other functions like `box.select` etc.
At second, this could be harmful, because calling empty `box.cfg` will start an instance without replication in readwrite mode, which, in case, could lead to unintended background data modification

I could not see a reason for implicit cfg inside sql and I insist on strict requirement of explicit `box.cfg` from `box.execute`

ImeevMA:
Prior to this patch, it was possible to call box.execute() before BOX was initialized, i.e. before calling box.cfg(). This, however, caused box.cfg() to be called automatically, which could be problematic as some parameters could not be changed after box.cfg() was called. After this patch, box.execute() will only be available when the BOX has been initialized.

Closes #4726

@TarantoolBot document
Title: box.execute() now available only after initialization of BOX Previously, it was possible to call box.execute() before the BOX was configured, in which case the BOX was configured automatically, which could lead to problems with BOX parameters. Now box.execute() can only be called after the BOX has been properly configured.


changelogs/unreleased/gh-4726-execute-field-in-box.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
```
Gerold103, line 1: I wouldn't call it a feature. Neither a bugfix really, but a bugfix fits better. Also this needs to be marked as a breaking change. Because there is at least one customer, who expected `box.execute` to be available before `box.cfg`: https://github.com/tarantool/tarantool/issues/4231. If his code still exists like that, then it will break.

Gerold103 (commented):
Thanks for the patch!

changelogs/unreleased/gh-4726-execute-field-in-box.md
Totktonada: I would ask @R-omk directly, but I guess that it was convenient for him to *capture* `box.execute` before `box.cfg()`, while the ability to *call* it before `box.cfg()` was irrelevant. I think we can keep ability to capture `box.execute` before `box.cfg()`, but raise an error instead of automatic box initialization.

I'm also curious, whether `local box_execute = box.execute` offers any benefits. Performance? Style consistency?

I saw considerable amount of old code, which doing something like `local gsub = string.gsub` at top of a module, but I doubt that it has any sense for LuaJIT. I saw [this](https://gitspartv.github.io/LuaJIT-Benchmarks/#test2) benchmark, but it counts difference in instructions outside of the loop. The real performance is the same (at least for this test). Curious fact: if I change `sin(3.14)` to `sin(i)`, the loops in both code variants are still equivalent.

src/box/lua/console.lua
```diff
@@ -278,6 +278,16 @@ local function set_language(storage, value)
         local msg = 'Invalid language "%s", supported languages: lua and sql.'
         return error(msg:format(value))
     end
+    if value == 'sql' then
+        local is_configured = false
+        for k, _ in pairs(box) do
+            is_configured = is_configured or k == 'execute'
+        end
+        if not is_configured then
+            local msg = "Unable to set language to 'sql' with unconfigured BOX"
```
Totktonada, line 287: Nit: I never seen `box` capitalized. It seems, it is unusual naming style for it.

changelogs/unreleased/gh-4726-execute-field-in-box.md
Totktonada: > I think we can keep ability to capture box.execute before box.cfg(), but raise an error instead of automatic box initialization.

BTW, it is not the only way to handle this case. We can also consider a yield till box initialization.

R-omk: > I think we can keep ability to capture box.execute before box.cfg(), but raise an error instead of automatic box initialization.

I agree with it, but it doesn't matter to me now.

src/box/lua/console.lua
```diff
@@ -278,6 +278,16 @@ local function set_language(storage, value)
         local msg = 'Invalid language "%s", supported languages: lua and sql.'
         return error(msg:format(value))
     end
+    if value == 'sql' then
+        local is_configured = false
+        for k, _ in pairs(box) do
+            is_configured = is_configured or k == 'execute'
+        end
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! I can't use `box.execute ~= nil` because accessing the `nil` field of `box` throws an error:
```
tarantool> box.something
---
- error: "builtin/box/load_cfg.lua:864: Please call box.cfg{} first\nstack traceback:\n\tbuiltin/box/load_cfg.lua:864:
    in function '__index'\n\t[string \"return box.something\"]:1: in main chunk\n\t[C]:
    in function 'pcall'\n\tbuiltin/box/console.lua:422: in function 'eval'\n\tbuiltin/box/console.lua:733:
    in function 'repl'\n\tbuiltin/box/console.lua:784: in function <builtin/box/console.lua:772>"
...

tarantool> box.something ~= nil
---
- error: "builtin/box/load_cfg.lua:864: Please call box.cfg{} first\nstack traceback:\n\tbuiltin/box/load_cfg.lua:864:
    in function '__index'\n\t[string \"return box.something ~= nil\"]:1: in main chunk\n\t[C]:
    in function 'pcall'\n\tbuiltin/box/console.lua:422: in function 'eval'\n\tbuiltin/box/console.lua:733:
    in function 'repl'\n\tbuiltin/box/console.lua:784: in function <builtin/box/console.lua:772>"
...
```

changelogs/unreleased/gh-4726-execute-field-in-box.md
ImeevMA: Thanks all for the review and comments! I still suggest to drop this method. This method doesn't something that can live without initialized `box`.

ImeevMA: > I wouldn't call it a feature. Neither a bugfix really, but a bugfix fits better. Also this needs to be marked as a breaking change. Because there is at least one customer, who expected `box.execute` to be available before `box.cfg`: #4231. If his code still exists like that, then it will break.

Thanks, fixed.

src/box/lua/console.lua
ImeevMA: Thank you, fixed.

src/box/lua/console.lua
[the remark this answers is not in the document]
Totktonada: `rawget()` should do the trick.

changelogs/unreleased/gh-4726-execute-field-in-box.md
Totktonada: I asked @Mons (as product manager). He said that `box.execute` should behave like anything else and so a user should get 'Please call box.cfg{} first' error at attempt to acquire this field before `box.cfg()`.

ImeevMA: Thanks for clarifying this issue!

src/box/lua/console.lua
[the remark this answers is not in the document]
ImeevMA: Thank you! Fixed.

src/box/lua/console.lua
Totktonada: That's minor, but I would also update it in the changelog entry.

changelogs/unreleased/gh-4726-execute-field-in-box.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* **[Breaking change]** Now the "execute" field is not available until BOX is
+  configured (gh-4726).
```
Totktonada, line 4: Shouldn't we also add somewhere that `box.execute()` does not automatically configures `box`? It seems as a more important fact for a user.

test/box-luatest/gh_4726_box_execute_before_cfg_test.lua
```diff
@@ -0,0 +1,15 @@
+local t = require('luatest')
+local g = t.group()
+
+g.test_box_execute_before_box_cfg = function()
+    t.assert_equals(rawget(box, 'execute') == nil, true)
```
Totktonada, line 5: Nit: I would rather use `t.assert_equals(rawget(box, 'execute'), nil)`. This way we'll get an actual value on the screen if the test case will fail.

Totktonada (approved):

src/box/lua/console.lua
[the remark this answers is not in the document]
Totktonada: *(I'm fine with the fix, but I left the discussion open, because it was opened by Vlad.)*

changelogs/unreleased/gh-4726-execute-field-in-box.md
Totktonada: *(I'm fine with the result of the discussion, but I left it open, because it was opened by Vlad.)*

changelogs/unreleased/gh-4726-execute-field-in-box.md
ImeevMA: I've added a doc-request that `execute()` cannot be called until the box has been initialized. I think this should be enough.

Totktonada: I meant the wording in the changelog entry, but anyway it is already too picky. Closing the discussion.

Gerold103 (approved):
Thanks for the fixes!

src/box/lua/console.lua
```diff
@@ -278,6 +278,12 @@ local function set_language(storage, value)
         local msg = 'Invalid language "%s", supported languages: lua and sql.'
         return error(msg:format(value))
     end
+    if value == 'sql' then
+        if rawget(box, 'execute') == nil then
+            local msg = "Unable to set language to 'sql' with unconfigured box"
+            return error(msg:format(value))
```
locker, line 284: Nit: `format()` is pointless here, please remove. Also, I'd put the whole condition in one `if`:

```Lua
if value == 'sql' and rawget(box, 'execute') == nil then
```

locker (changes requested):

src/box/lua/console.lua
ImeevMA: True, sorry. Fixed.

locker (approved):

changelogs/unreleased/gh-4726-execute-field-in-box.md
Totktonada: Filed https://github.com/GitSparTV/LuaJIT-Benchmarks/issues/3.

Issue #4804, opened 2020-03-11
Upgrade schema correctly at tarantool update within a release


Totktonada:
The overall situation is described in #4800.

I propose to implement necessary mechanics to do upgrades within one `x.y.z.*` release correctly. So it will be possible to implement an upgrade that does not block further upgrades within a release. It will not fix past upgrades.

Let's fix past upgrades (or provide a separate script to fix them) in the scope of #4800.

PR #7789, opened 2022-10-10
Revoke access of guest to `LUA` function

alyapunov:
Since the function is actually an eval, by default there should be no execute access right in public role.
    
Closes tarantool/security#14


tools/gen-bootstrap.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+
```
Totktonada, line 2: How about `set -exu -o pipefail` for strictness and preventing mistakes?

tools/gen-bootstrap.sh
```diff
@@ -0,0 +1,115 @@
...
+# In addition, it's so fun to play with hex!
+if ! head -n 6 ./bootstrap.snap | egrep "VClock: {.*}" > /dev/null; then
+    echo "Failed to patch VClock: is was not found."
+    exit 1
+fi
+head -n 6 ./bootstrap.snap > old_header.txt
+head -n 6 ./bootstrap.snap | sed "s/VClock: {.*}/VClock: {}/" > new_header.txt
+bootstrap_hex=$(xxd -p bootstrap.snap | tr -d '\n')
+old_header_hex=$(xxd -p old_header.txt | tr -d '\n')
+new_header_hex=$(xxd -p new_header.txt | tr -d '\n')
```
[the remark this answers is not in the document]
Totktonada: Ah, I see, `xxd -v` is already there. Sorry, overlooked it.

src/box/lua/upgrade.lua
```diff
@@ -1241,6 +1241,28 @@ local function upgrade_to_2_11_0()
 end
 --------------------------------------------------------------------------------
 
+--------------------------------------------------------------------------------
+-- Tarantool 2.11.1
```
ImeevMA, line 1245: I think we usually use the name of the current version for the upgrade, not the next one. I mean, since the `2.11.1` tag is still not set, and `2.11.0-entrypoint` already exists, it might be better to move this function to the `2.11.0` schema version. Although I may be wrong.

test/box/access.test.lua
```diff
@@ -841,6 +841,10 @@ _ = box.schema.func.call("LUA", "return box.space._space:count()")
 _ = box.schema.func.call("box.schema.user.info", 0)
 _ = box.schema.func.call("box.schema.user.info", 1)
 session.su('admin')
+c = net.connect(LISTEN.host, LISTEN.service)
+c:call('LUA', {'box.session.su("admin") return 1'})
```
ImeevMA, line 845: I suggest to use this instead:
```
c:call('LUA', {'return 1'})
```

I mean the result is the same, but the current line looks more defiant.

ImeevMA (approved):
Hi! Thank you for the patch! I only reviewed the patch `box: revoke access of guest to LUA function`.

src/box/lua/upgrade.lua
Totktonada: It seems, it should be backported to 2.10, so, maybe, 2.11.1 is not the best name (it seems, 2.10.4 fits better). OTOH, if existing 2.11.0 migration already applied, 2.10.4 will not be applied (see #4804). It shouldn't affect upgrades from a release to a release, but may hit developers a bit.

ImeevMA: I think in the case of 2.10 it just needs to be properly backported. I mean it should be done in another PR.

Totktonada: Good idea (2.11.0 in 2.11 and 2.10.4 in 2.10).

Regarding adding to existing 2.11.0 upgrade: it follows our practice. I don't like it (I would link #4804 here again), but we can't do more without refactoring. Or really disregard 3rd version number connection beetween tarantool runtime and tarantool schema and name the new upgrade as 2.11.1.

test/box/access.test.lua
alyapunov: OK. I just wanted to ensure that a user can't get admin rights.

tools/gen-bootstrap.sh
alyapunov: ok, if you say so )
BTW I prefer to check critical sections manually in order to display a good error message.

src/box/lua/upgrade.lua
alyapunov: Please check the updated version and verify that I understood you correctly.
Note that the last commit can be simply backported to 2.10.

src/box/lua/upgrade.lua
```diff
@@ -1238,6 +1260,8 @@ end
 
 local function upgrade_to_2_11_0()
     revoke_write_access_on__collation_from_role_public()
+    -- Revoke again if upgrade to 2.10.4 was skipped.
+    revoke_execute_access_to_lua_function_from_role_public()
```
locker, line 1264: I don't think we need to call this function here - it should be enough to call it only in `upgrade_to_2_10_4()`, because 2.11 hasn't been released yet so upgrade to 2.10.4 can't possibly be skipped.

src/box/lua/upgrade.lua
ImeevMA: I agree that this will work, although I'm a little uncomfortable creating an update to `2.10.4` after `2.11.0-entrypoint`. Also, I think to be able to backport to 2.10 you need to remove `revoke_execute_access_to_lua_function_from_role_public()` from `upgrade_to_2_11_0()` as suggested by @locker.

src/box/lua/upgrade.lua
alyapunov: Oh, removed.

src/box/lua/upgrade.lua
alyapunov: Thanks, removed.

tools/gen-bootstrap.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "Folder $TEMP_DIR was found, it seems that previous run was failed."
+    read -p "Drop the folder and proceed [Yn]?" yn
+    case ${yn:-y} in
+        [Yy]* ) rm -r "$TEMP_DIR";;
+        [Nn]* ) echo "OK, exiting."; exit 0;;
+        * ) echo "Wrong answer."; exit 1;;
+    esac
+fi
+
+if ! cmake --build . -j; then
```
locker, line 46: This would better be conditional - rebuild only if a flag is set. The flag may be set to true by default though.

tools/gen-bootstrap.sh
```diff
@@ -0,0 +1,117 @@
...
+if ! ../"$TNT_BINARY" -e "box.cfg{} box.internal.bootstrap() os.exit(0)"; then
+    echo "Failed to run script by tarantool, exiting."
+    exit 1
+fi
+
+mv $(ls *.snap | sort | tail -n 1) ./bootstrap.snap
+
+# Clear VClock field of the file.
+# Actually all the code below does only simple command:
+# sed -i "s/VClock: {.*}/VClock: {}/" ./bootstrap.snap
```
locker, line 63: The code below looks complex. Wouldn't it be better to use `sed` to replace the first match instead?

```
sed -i "s/VClock: {.*}/VClock: {}/"
```

alyapunov: You copy-pasted my line, but we discussed offline that there was special option for set..

locker: Sorry, I meant to paste this:

```
sed -i '0,/^VClock: {}$/{s//VClock: foo/}'
```

src/box/lua/upgrade.lua
Totktonada: As Vladimir stated below, there were no 2.11 releases, so it'll work. It is technically correct, so I technically agree. You know my opinion on given upgrade approach in general.

src/box/lua/upgrade.lua
```diff
@@ -1227,6 +1227,28 @@ local function upgrade_to_2_10_1()
...
+-- Tarantool 2.10.4
+--------------------------------------------------------------------------------
+local function revoke_execute_access_to_lua_function_from_role_public()
+    local _priv = box.space[box.schema.PRIV_ID]
+    if box.func.LUA then
+        local row = _priv:get{PUBLIC, 'function', box.func.LUA.id}
+        if row and bit.band(row.privilege, box.priv.X) ~= 0 then
+            local privilege = bit.bxor(row.privilege, box.priv.X)
+            privilege = bit.bor(privilege, box.priv.R)
+            log.info("revoke execute access to 'lua' function from public role")
```
Totktonada, line 1240: Nit: Function names as case sensetive, so technically we should say `'LUA' function` instead of `'lua' function'` here. Sorry for nits.

Totktonada (approved):
I'm not sure that I understand privileges machinery in tarantool well enough, but from what I know the PR looks okay. Thanks for the script!

src/box/lua/upgrade.lua
alyapunov: Thanks, fixes.

src/box/lua/upgrade.lua
```diff
@@ -1227,6 +1227,28 @@ local function upgrade_to_2_10_1()
...
+--------------------------------------------------------------------------------
+-- Tarantool 2.10.4
+--------------------------------------------------------------------------------
+local function revoke_execute_access_to_lua_function_from_role_public()
+    local _priv = box.space[box.schema.PRIV_ID]
+    if box.func.LUA then
+        local row = _priv:get{PUBLIC, 'function', box.func.LUA.id}
+        if row and bit.band(row.privilege, box.priv.X) ~= 0 then
+            local privilege = bit.bxor(row.privilege, box.priv.X)
+            privilege = bit.bor(privilege, box.priv.R)
```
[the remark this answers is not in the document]
alyapunov: Yes, you are right, I made it in order to make minimal necessary changes.
Add a comment that clarifies this.

tools/gen-bootstrap.sh
alyapunov: Thanks, I added you command to comment, but I don't want to drop hex/binary code because it was a challenge after which I decided to write the script.
So let's leave it as is and iff the code will need to be supported we can simply switch to sed version.

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7796, opened 2022-10-12
Pagination in pairs

### LINKED ISSUE #7638: Add pagination to pairs method
Implement option `after` in pairs method. If `after` is specified it must contain a tuple from which selection must continue or its position. More strictly:

- For tree index (memtx and vinyl) and forward iteration (EQ, GE, GT, ALL) pairs must internally skip all tuples less-or-equal-to tuple given in after in terms of that index.
- For tree index (memtx and vinyl) and reverse iteration (REQ, LE, LT) pairs must internally skip all tuples greater-or-equal-to tuple given in after in terms of that index.

Part of #7639

### LINKED ISSUE #7639: Introduce pagination
[PM link](https://www.notion.so/tarantool/After-in-iterators-0105eac5996b41e1ac7fd418de373b56)

Generalization of #5282.

Implement options `after` and `fetch_pos` in `select` and `pairs` methods. If `after` is specifiend it must contain a tuple from which selection must continue or its `position`. More strictly:

- For tree index (memtx and vinyl) and forward iteration (EQ, GE, GT, ALL) `select` and `pairs` must internally skip all tuples less-or-equal-to tuple given in after in terms of that index.
- For tree index (memtx and vinyl) and reverse iteration (REQ, LE, LT) `select` and `pairs` must internally skip all tuples greater-or-equal-to tuple given in after in terms of that index.
- For all other indexes that option must not be allowed (instead of rtree with NEIGHBOR iterator, but it is follow-up).
- Duplicates are not allowed.
- Pagination with `after` option must be more effective that limit-offset approach and it must not depend on the number of previously selected tuples.
- Performance of `select` and `pairs` without introduced options must not degrade.
- In Lua tuple must be table or cdata with C pointer to tuple, `position` must be a string, that was received from special methods - user is not allowed to create `position` by himself.
- In IPROTO tuple must be MP_ARRAY, `position` must be `MP_BIN`.
- Only those fields of the after tuple that participate in key definition of the index (and also, if the index is not unique, key definition of the primary index) are actually make sense; all other fields in the tuple are guaranteed to be ignored.
- If `box.NULL` or empty Lua string are passed to `after` option, it makes no effect (start position).

Option `fetch_pos` must return position descriptor of last selected tuple (or passed position descriptor, if no tuples were found) as the second return value.

Implement method `index:tuple_pos(tuple)` to allow get position by tuple.

Example:
```lua
local tuples, pos = index:select(key, {fetch_pos=true, limit = 10})
process_tuples(tuples)
tuples, pos = index:select(key, {fetch_pos=true, limit=10, after=pos})
process_tuples(tuples)
```

## Sub-tasks
- [x] #7633
- [x] #7634
- [x] #7635
- [x] #7636
- [x] #7637
- [x] #7638

drewdzzz:
- Introduce pagination to pairs
- Add changelog and doc request for pagination

Closes #7638
Closes #7639

### DIFF
--- a/changelogs/unreleased/pagination.md
+++ b/changelogs/unreleased/pagination.md
@@ -1,0 +1,5 @@
+## feature/core
+
+* Introduced pagination support for memtx and vinyl tree indexes. Now, it's
+  possible to resume `pairs` and `select` from the position where the last
+  call stopped (gh-7639).
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -42,6 +42,7 @@
 #include "memtx_tx.h"
 #include "box.h"
 #include "base64.h"
+#include "scoped_guard.h"
 
 struct rlist box_on_select = RLIST_HEAD_INITIALIZER(box_on_select);
 
@@ -429,8 +430,9 @@ box_index_count(uint32_t space_id, uint32_t index_id, int type,
 /* {{{ Iterators ************************************************/
 
 box_iterator_t *
-box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
-                   const char *key, const char *key_end)
+box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
+			 const char *key, const char *key_end,
+			 const char *packed_pos, const char *packed_pos_end)
 {
 	assert(key != NULL && key_end != NULL);
 	mp_tuple_assert(key, key_end);
@@ -449,19 +451,59 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
 	uint32_t part_count = mp_decode_array(&key);
 	if (key_validate(index->def, itype, key, part_count))
 		return NULL;
+	const char *pos, *pos_end;
+	char *pos_buf = NULL;
+	uint32_t pos_buf_size = 0;
+	auto pos_guard = make_scoped_guard([&pos_buf, &pos_buf_size] {
+		if (pos_buf != NULL)
+			runtime_memory_free(pos_buf, pos_buf_size);
+	});
+	if (packed_pos != NULL && packed_pos != packed_pos_end) {
+		pos_buf_size =
+			iterator_position_unpack_bufsize(packed_pos,
+							 packed_pos_end);
+		pos_buf = (char *)runtime_memory_alloc(pos_buf_size);
+		if (pos_buf == NULL) {
+			diag_set(OutOfMemory, pos_buf_size,
+				 "runtime_memory_alloc", "pos_buf");
+			return NULL;
+		}
+		if (iterator_position_unpack(packed_pos, packed_pos_end,
+					     pos_buf, pos_buf_size,
+					     &pos, &pos_end) != 0)
+			return NULL;
+		uint32_t pos_part_count = mp_decode_array(&pos);
+		if (exact_key_validate_nullable(index->def->cmp_def, pos,
+						pos_part_count) != 0)
+			return NULL;
+	} else {
+		pos = NULL;
+		pos_end = NULL;
+	}
 	box_run_on_select(space, index, itype, key_array);
 	struct txn *txn;
 	struct txn_ro_savepoint svp;
 	if (txn_begin_ro_stmt(space, &txn, &svp) != 0)
 		return NULL;
-	struct iterator *it = index_create_iterator(index, itype,
-						    key, part_count);
+	struct iterator *it = index_create_iterator_after(index, itype, key,
+							  part_count, pos);
 	txn_end_ro_stmt(txn, &svp);
 	if (it == NULL)
 		return NULL;
 	it->space = space;
+	it->pos_buf = pos_buf;
+	it->pos_buf_size = pos_buf_size;
+	pos_guard.is_active = false;
 	rmean_collect(rmean_box, IPROTO_SELECT, 1);
 	return it;
+}
+
+box_iterator_t *
+box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
+		   const char *key, const char *key_end)
+{
+	return box_index_iterator_after(space_id, index_id, type, key, key_end,
+					NULL, NULL);
 }
 
 int
@@ -534,6 +576,8 @@ iterator_create(struct iterator *it, struct index *index)
 	it->index_id = index->def->iid;
 	it->index = index;
 	it->space = NULL;
+	it->pos_buf = NULL;
+	it->pos_buf_size = 0;
 }
 
 /**
@@ -689,6 +733,11 @@ fail:
 void
 iterator_delete(struct iterator *it)
 {
+	if (it->pos_buf != NULL) {
+		runtime_memory_free(it->pos_buf, it->pos_buf_size);
+		it->pos_buf = NULL;
+		it->pos_buf_size = 0;
+	}
 	assert(it->free != NULL);
 	it->free(it);
 }
--- a/src/box/index.h
+++ b/src/box/index.h
@@ -254,6 +254,16 @@ box_tuple_extract_key(box_tuple_t *tuple, uint32_t space_id,
 /** \endcond public */
 
 /**
+ * Allocate and initialize iterator for space_id, index_id. If packed_pos is
+ * not NULL, iterator will start right after tuple with position, described by
+ * this argument. A returned iterator must be destroyed by box_iterator_free().
+ */
+box_iterator_t *
+box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
+			 const char *key, const char *key_end,
+			 const char *packed_pos, const char *packed_pos_end);
+
+/**
  * Get packed position of tuple in index to pass it to box_select
  * (multikey and func indexes are not supported). Returned position
  * is allocated on the fiber region.
@@ -316,6 +326,17 @@ struct iterator {
 	uint32_t space_id;
 	/** ID of the index the iterator is for. */
 	uint32_t index_id;
+	/**
+	 * Allocator that was used for allocation of fields pos_buf requires
+	 * its size on deallocation.
+	 */
+	uint32_t pos_buf_size;
+	/**
+	 * Pointer to a buffer which was allocated for start position with
+	 * runtime_alloc. Will be freed on iterator_delete.
+	 * Needed for box_index_iterator_after.
+	 */
+	char *pos_buf;
 	/**
 	 * Pointer to the index the iterator is for.
 	 * Guaranteed to be valid only if the schema
--- a/src/box/lua/index.c
+++ b/src/box/lua/index.c
@@ -35,7 +35,7 @@
 #include "box/box.h"
 #include "box/index.h"
 #include "box/lua/tuple.h"
-#include "box/lua/misc.h" /* lbox_encode_tuple_on_gc() */
+#include "box/lua/misc.h"
 #include "small/region.h"
 #include "fiber.h"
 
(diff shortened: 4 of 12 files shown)


src/box/index.cc
```diff
@@ -457,23 +458,58 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
 	uint32_t part_count = mp_decode_array(&key);
 	if (key_validate(index->def, itype, key, part_count))
 		return NULL;
+	const char *pos, *pos_end;
+	char *buf = NULL;
+	if (packed_pos != NULL && packed_pos != packed_pos_end) {
+		uint32_t buf_size =
+			iterator_position_unpack_bufsize(packed_pos,
+							 packed_pos_end);
+		buf = (char *)xmalloc(buf_size);
```
locker, line 467: Let's allocate the data from `runtime_alloc` - it should be much faster than `malloc`.

src/box/index.cc
```diff
@@ -457,23 +458,58 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
...
+	if (txn_begin_ro_stmt(space, &txn, &svp) != 0) {
+		free(buf);
 		return NULL;
-	struct iterator *it = index_create_iterator(index, itype,
-						    key, part_count);
+	}
+	struct iterator *it = index_create_iterator_after(index, itype, key,
+							  part_count, pos);
 	if (it == NULL) {
+		free(buf);
```
locker, line 494: Nit: Please use `make_scoped_guard` instead of inserting `free(buf)` before each return statement.

src/box/index.cc
```diff
@@ -457,23 +458,58 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
...
 		txn_rollback_stmt(txn);
 		return NULL;
 	}
 	it->space = space;
+	it->blessed_position_buf = buf;
```
locker, line 499: Nit: Why "blessed"? Let's name it simply `pos_buf`.

src/box/index.cc
```diff
@@ -457,23 +458,58 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
 	uint32_t part_count = mp_decode_array(&key);
 	if (key_validate(index->def, itype, key, part_count))
 		return NULL;
+	const char *pos, *pos_end;
+	char *buf = NULL;
```
locker, line 462: Nit: Please rename to `pos_buf` - there may be other buffers in this function.

src/box/lua/index.c
```diff
@@ -242,18 +242,30 @@ box_index_init_iterator_types(struct lua_State *L, int idx)
...
 	uint32_t space_id = lua_tonumber(L, 1);
 	uint32_t index_id = lua_tonumber(L, 2);
 	uint32_t iterator = lua_tonumber(L, 3);
 	size_t mpkey_len;
 	const char *mpkey = lua_tolstring(L, 4, &mpkey_len); /* Key encoded by Lua */
 	/* const char *key = lbox_encode_tuple_on_gc(L, 4, key_len); */
-	struct iterator *it = box_index_iterator(space_id, index_id, iterator,
-						 mpkey, mpkey + mpkey_len);
+	size_t packed_pos_size;
+	const char *packed_pos = lua_tolstring(L, 5, &packed_pos_size);
```
locker, line 258: Why don't you use `lbox_normalize_position` here, like `lbox_select` does?

src/box/index.h
```diff
@@ -327,6 +337,12 @@ struct iterator {
 	 * Don't access directly, use iterator_space().
 	 */
 	struct space *space;
+	/**
+	 * Pointer to a buffer which was allocated for start position with
+	 * malloc. Will be freed on iterator_delete.
+	 * Needed for box_index_iterator_after.
+	 */
+	char *blessed_position_buf;
```
locker, line 345: Maybe, store it in the iterator implementation struct?

Rationale:
 - Vinyl already copies the position internally.
 - The iterator may choose to free the position right after the first iteration, because it's no longer needed (Vinyl already does that).
 - The iterator may start the iteration on open and not store the position at all.

src/box/lua/schema.lua
```diff
@@ -1960,6 +1962,44 @@ end
 
 internal.check_iterator_type = check_iterator_type
 
+local function check_option_after(opts)
```
locker, line 1965: Let's introduce `check_pairs_opts` instead that would check all options of `pairs`, similarly to `check_select_opts`.

locker (changes requested):

changelogs/unreleased/pagination.md
```diff
@@ -0,0 +1,3 @@
+## feature/core
+
+* Memtx and vinyl tree indexes supports pagination now.
```
locker, line 3: - supports -> support
- Missing reference to the ticket.
- Let's be a bit more verbose - not everyone may now what "pagination" is:
   > Introduced pagination support for memtx and vinyl tree indexes. Now, it's possible to resume `pairs` and `select` from the position where the last call stopped (gh-XXXX).

src/box/index.cc
drewdzzz: Done

src/box/index.cc
drewdzzz: Done
Blessed was an analogue to `tuple_bless` function.

src/box/index.cc
drewdzzz: Done

src/box/lua/index.c
drewdzzz: I did it because I wanted to save unpacked position in Lua. It is not actual anymore, so I rewrote it.

src/box/index.h
drewdzzz: Good idea, but this brings up an API question - how should we pass this buffer?

drewdzzz: - A separate method looks weird.
- A new argument in `create_iterator` method looks ugly. Even more, if we will use runtime_alloc instead of malloc, we will need to store size of buffer - it is even more ugly.
Maybe we should keep it simple, especially because 99% of `pairs` usage is `for` loop.

src/box/lua/schema.lua
drewdzzz: Done

changelogs/unreleased/pagination.md
drewdzzz: Done

src/box/index.h
locker: Agree - passing the position size to the `create_iterator` callback would be a problem. Let's leave it as is for now.

src/box/index.cc
locker: `tuple_bless` is probably called this way, because it makes a tuple usable in the public C and Lua API by incrementing its ref counter (initially a tuple has ref counter of 0, which makes it unusable). Not applicable here.

src/box/index.cc
drewdzzz: Done - introduced `tuple_runtime_memory_alloc(free)` methods right in pagination commit.

src/box/index.cc
```diff
@@ -449,21 +451,61 @@ box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
 	uint32_t part_count = mp_decode_array(&key);
 	if (key_validate(index->def, itype, key, part_count))
 		return NULL;
+	const char *pos, *pos_end;
+	char *pos_buf = NULL;
+	uint32_t pos_buf_size = 0;
+	auto pos_guard = make_scoped_guard([&pos_buf, &pos_buf_size] {
```
locker, line 457: Nit: Why do use capture by reference (&) here? Capture by value should do - these are plain pointers. Also, we usually use "capture everything by value" in guards: `make_scoped_guard([=])`.

src/box/lua/schema.lua
```diff
@@ -2362,12 +2405,16 @@ base_index_mt.pairs_ffi = function(index, key, opts)
     local ibuf = cord_ibuf_take()
     local pkey, pkey_end = tuple_encode(ibuf, key)
     local itype = check_iterator_type(opts, pkey + 1 >= pkey_end);
+    local svp = builtin.box_region_used()
+    local after = check_pairs_opts(opts)
```
locker, line 2409: `check_pair_opts` should check and return *all* options passed to `pairs`, just like `check_select_opts`.

src/box/lua/schema.lua
```diff
@@ -34,10 +34,12 @@ ffi.cdef[[
...
 
-    /** \cond public */
     box_iterator_t *
-    box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
-                       const char *key, const char *key_end);
+    box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
+                             const char *key, const char *key_end,
+                             const char *packed_pos,
+                             const char *packed_pos_end);
+    /** \cond public */
```
locker, line 42: Nit: These comments (`cond public`) make no sense in `ffi.cdef`. Let's drop them all from this block while we are at it.

src/box/tuple.h
```diff
@@ -1449,6 +1449,18 @@ tuple_to_buf(struct tuple *tuple, char *buf, size_t size);
 size_t
 tuple_runtime_memory_used(void);
 
+/**
+ * Allocate size bytes on runtime arena for tuples.
+ */
+void *
+tuple_runtime_memory_alloc(size_t size);
```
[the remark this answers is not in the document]
drewdzzz: I named them likely to `tuple_runtime_memory_used()`

src/box/index.cc
locker: > But it does not work here - `pos_buf` is `NULL` initially

Right, missed that. Looks confusing, but I don't see how we could do better. Let's leave as is.

src/box/tuple.h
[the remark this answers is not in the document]
locker: I'm not sure we should account these allocations to `tuple_runtime_memory_used()`. After all, we don't account `struct iterator` allocations there, although they are allocated from the same runtime arena. Please discuss this question and the naming with @alyapunov. If we agree that this is in fact tuple memory and we do want to account it in `box.runtime.info().tuple`, then the current names should be fine.

test/engine-luatest/pagination_test.lua
```diff
@@ -19,9 +19,11 @@ local function start_server(server)
     end)
 end
 
+-- TODO(gh-7943): Enable Vinyl cache
 tree_g.before_all(function(cg)
     cg.server = server:new{
         alias   = 'default',
+        box_cfg = {vinyl_cache = 0},
```
locker, line 26: I just merged the fix for #7943. This isn't longer needed.

drewdzzz: Removed

src/box/lua/schema.lua
drewdzzz: Done

src/box/lua/schema.lua
drewdzzz: Added iterator type check to this function

src/box/tuple.h
```diff
@@ -1449,6 +1449,18 @@ tuple_to_buf(struct tuple *tuple, char *buf, size_t size);
...
+
+/**
+ * Free memory of size bytes allocated on runtime arena for tuples.
+ */
+void
+tuple_runtime_memory_free(void *ptr, size_t size);
```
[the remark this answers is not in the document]
drewdzzz: Discussed it with @alyapunov, we decided not to store size in a header in order to preserve this feature of the allocator and to avoid alignment problems.

src/box/tuple.h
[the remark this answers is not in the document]
drewdzzz: - Let's call them `runtime_alloc` and `runtime_free`.
- We don't really want to account these allocations in `box.runtime.info().tuple`, but it is already incorrect because of functional keys, so we decided that it's not a big problem if iterator position will be accounted. Especially, because our iterator position will not live long - 99% of pairs are used with `for` loop.
- We can account memory allocated by `runtime_alloc` in a separate counter and then just subtract it from allocator's used counter. But then we'll want to move there functional keys and so on, and probably add a new stats field. We decided there is no need to make a big deal out of it, and we can just turn a blind eye to a small inaccuracy in the statistics or use malloc to keep it simple.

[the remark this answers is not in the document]
locker: > * We don't really want to account these allocations in `box.runtime.info().tuple`, but it is already incorrect because of functional keys, so we decided that it's not a big problem if iterator position will be accounted. Especially, because our iterator position will not live long - 99% of pairs are used with `for` loop.

FYI functional keys are allocated as memtx tuples so they don't have anything to do with the runtime arena:

https://github.com/tarantool/tarantool/blob/d2d7f4de3a6ba2caa7f260860e750c733ee3201d/src/box/memtx_engine.cc#L1358-L1359

PR #7801, opened 2022-10-12
tools: add a srcipt that updates bootstrap.snap

alyapunov:
By design after patching update.lua, introducing new data version, a developer needs to update bootstrap.snap, which would have the new data version.

Generation of bootstrap.snap is a bit tricky, there's a manual: https://github.com/tarantool/tarantool/wiki/How-to-generate-new-bootstrap-snapshot

This commit presents a simple script that automates the process. Just run it in tarantool build directory and check its output. Check --help for script options.

Follow-up of tarantool/security#14

NO_DOC=internal tool
NO_TEST=internal tool
NO_CHANGELOG=internal tool

### DIFF
--- a/tools/gen-bootstrap.sh
+++ b/tools/gen-bootstrap.sh
@@ -1,0 +1,195 @@
+#!/bin/bash
+
+# The script is designed for simplification of new bootstrap.snap generation.
+# Just run it in tarantool build directory after updating update.lua (which
+# introduces new data version) and examine its output.
+# The script will try to generate bootstrap.snap in build directory (which is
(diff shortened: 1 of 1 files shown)


tools/gen-bootstrap.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
```
Totktonada, line 1: Typo in the commit message title/PR title: `srcipt`.

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #7447, opened 2022-07-20
log.cfg panic on incorrect syslog config


nshy:
```
Tarantool 2.11.0-entrypoint-210-g111663821
type 'help' for interactive help
 tarantool> log = require('log')
---
...

 tarantool> log.cfg{log='syslog:xxx'}
IllegalParams: bad option 'xxx'
failed to initialize logging subsystem
```
At the same time:
```
Tarantool 2.11.0-entrypoint-210-g111663821
type 'help' for interactive help
 tarantool> box.cfg{log='syslog:xxx'}
---
- error: 'Incorrect value for option ''log'': bad option ''xxx'''
...

 tarantool>
```
I guess we don't need to panic in the the first case as well.

olegrok:
Looks similar to https://github.com/tarantool/tarantool/issues/5130

nshy:
Duplicates #5130

nshy:
Closed as duplicate.

locker:
The PR that fixes this issue does quite a lot of refactoring of the configuration subsystem, which makes it difficult to backport it to 2.10. Since the issue is minor (exit on invalid config instead of raising an error) and it was reported by a developer, let's commit it to 2.11 only.

PR #7804, opened 2022-10-13
lua: add luaT_dostring helper

nshy:
It is same as luaL_dosting but also sets fiber diag.

Part of #7447

NO_DOC=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -500,6 +500,18 @@ luaT_call(struct lua_State *L, int nargs, int nreturns)
 }
 
 int
+luaT_dostring(struct lua_State *L, const char *str)
+{
+	if (luaL_loadstring(L, str) != 0) {
+		diag_set(LuajitError, lua_tostring(L, -1));
+		return -1;
+	}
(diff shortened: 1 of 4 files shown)


src/lua/utils.c
```diff
@@ -499,6 +499,18 @@ luaT_call(struct lua_State *L, int nargs, int nreturns)
 	return 0;
 }
 
+int
+luaT_dostring(struct lua_State *L, const char *str)
+{
+	if (luaL_loadstring(L, str) != 0) {
+		diag_set(LuajitError, lua_tostring(L, -1));
```
locker, line 506: This function pollutes the Lua stack on error. We should probably pop the string to be on a safe side.

Anyway, I don't understand why you need this function. AFAIU in the scope of #7447 you need to run a log configuration function so the Lua syntax is guaranteed to be correct. Typically, we don't check the result of `luaL_loadstring` in such cases, but even if we did, a panic or assertion should be enough. Please merge it into the PR that fixes #7447 if you really need it.

PR #7809, opened 2022-10-14
build: use relative paths in diagnostics and debugging information

### LINKED ISSUE #7808: Diagnostics and debugging information should provide relative paths instead of absolute ones
Originally raised in tarantool/security#15.

```lua

box.cfg{log_format = 'json'}

s:create_index('pk', {parts = {{1}, {2}}})

s:insert{1}

os.exit()

```

**Actual behavior**

```json
{"time": "2022-10-14T14:46:10.411+0300", "level": "ERROR", "message": "ER_FIELD_MISSING", "error": "Tuple field 2 required by space format is missing", "pid": 53796 , "cord_name": "main", "file": " \/Users\/georgiy.lebedev\/Work\/tarantool\/src\/box\/tuple_format.c", "line": 1250}
```

**Expected behavior**

```json
{"time": "2022-10-14T14:46:10.411+0300", "level": "ERROR", "message": "ER_FIELD_MISSING", "error": "Tuple field 2 required by space format is missing", "pid": 53796 , "cord_name": "main", "file": ".\/src\/box\/tuple_format.c", "line": 1250}
```

CuriousGeorgiy:
Since our diagnostics use the `__FILE__` macro, they provide absolute paths, which is kind of redundant and inconsistent: replace them paths with relative ones.

As for debugging information, replacing absolute paths with relative ones also requires an extra command to tell the debugger where to find the source files, which is not convenient for developers: provide a new `DEV_BUILD` option (turned off by default), which replaces absolute paths with relative ones in debugging information if turned off.

Closes #7808


CuriousGeorgiy:
# :bangbang: ATTENTION :bangbang:

A notice should be posted in all relevant development chats about the need to turn on the new `DEV_BUILD` option.

CMakeLists.txt
```diff
@@ -91,6 +91,23 @@ if (ENABLE_HARDENING)
...
+# Assuming the C and CXX compiler IDs are equal.
+# For clang `MATCHES` is used to match both Clang and AppleClang.
+if (CMAKE_C_COMPILER_ID MATCHES "Clang" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "10" OR
+    CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "8")
+    add_compile_flags("C;CXX" "-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=.")
+endif()
+
+# Assuming the C and CXX compiler IDs are equal.
+# For clang `MATCHES` is used to match both Clang and AppleClang.
+if (NOT DEV_BUILD AND
```
Gumix, line 105: This condition is FALSE when DEV_BUILD is FALSE and GNU >= 4.4.7, probably parentheses after AND are missed.

Gumix (changes requested):

CMakeLists.txt
```diff
@@ -91,6 +91,23 @@ if (ENABLE_HARDENING)
...
+    CMAKE_C_COMPILER_ID MATCHES "Clang" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "3.8" OR
+    CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "4.4.7")
+    add_compile_flags("C;CXX" "-fdebug-prefix-map=${CMAKE_SOURCE_DIR}=.")
```
Gumix, line 108: BTW these flags are not propagated to libunwind:
```
$ readelf --debug-dump=info ./build/src/tarantool | grep src
...
    <81dde0>   DW_AT_name        : (indirect string, offset: 0xbdcfc): /home/iverbin/git/tarantool/third_party/libunwind/src/mi/strerror.c
    <81dde4>   DW_AT_comp_dir    : (indirect string, offset: 0xbde00): /home/iverbin/git/tarantool/build/build/libunwind/work/src
    <81ed13>   DW_AT_name        : (indirect string, offset: 0xbdfdc): /home/iverbin/git/tarantool/third_party/libunwind/src/mi/Gget_accessors.c
...
```

Probably static build is also affected.

CMakeLists.txt
CuriousGeorgiy: Turned out that `AND` and `OR` logical operators have the same [precedence](https://cmake.org/cmake/help/latest/command/if.html#condition-syntax)...

> Binary logical operators [AND](https://cmake.org/cmake/help/latest/command/if.html#and) and [OR](https://cmake.org/cmake/help/latest/command/if.html#or), from left to right, without any short-circuit.

CMakeLists.txt
CuriousGeorgiy: Fixed: added these flags to new global variables used for passing compiler flags to dependencies. Also added these flags to static build.

CuriousGeorgiy:
@Gumix @locker I'm kind of lazy too add a test for this, but, if you insist, I can add one.

CuriousGeorgiy:
I am also a little confusing about the order of adding modules in CMake: `cmake/utils.cmake` now relies on `cmake/profile.cmake` and `cmake/os.cmake` (it uses variables set by them), but AFAIU, we only need the variables set by the time the module functions are invoked.

cmake/utils.cmake
```diff
@@ -138,3 +138,32 @@ function(set_hardening_flags)
...
+    # the source files, which is not convenient for developers, it is disabled when
+    # `DEV_BUILD` option is turned on.
+    if (NOT DEV_BUILD AND
+        ((CMAKE_C_COMPILER_ID MATCHES "Clang" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "3.8") OR
+        (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "4.4.7")))
+        # The AppleClang assembler does not support this option.
+        if (NOT APPLE)
+            set(AS_DEBUG_PREFIX_MAP_FLAG "-Wa,--debug-prefix-map=${CMAKE_SOURCE_DIR}=.")
+        endif()
+        set(DEBUG_PREFIX_MAP_FLAGS "-fdebug-prefix-map=${CMAKE_SOURCE_DIR}=. ${AS_DEBUG_PREFIX_MAP_FLAG}" PARENT_SCOPE)
```
Totktonada, line 167: There is no requirement to don't have (build) system paths in the executable or logs. It is not recommended to show them in outgoing messages. Let's strip it from them? This way we don't need to break `gdb.sh`, splitdebug installs, local debugging and likely other dev/user/admin flows that lean on usual paths in the debug information.

CuriousGeorgiy: We discussed this in person with @locker and @alyapunov: it looks like a reasonable default for production builds.

CuriousGeorgiy:
I don't quite understand why the release build has a lot of failed tests: the only significant change I made is passed hardening flags to all dependecies that were skipped previously.

cmake/utils.cmake
```diff
@@ -114,3 +116,25 @@ function(file_is_in_directory varname file dir)
         set(${varname} TRUE PARENT_SCOPE)
     endif()
 endfunction()
+
+#
+# Sets the `ENABLE_HARDENING_DEFAULT`, `HARDENING_FLAGS` and `HARDENING_LDFLAGS`
+# variables for passing hardening compiler flags.
+#
+# Uses `ENABLE_FUZZER` and `TARGET_OS_FREEBSD` variables.
+function(set_hardening_flags)
```
locker, line 125: I don't think this belongs in `utils.cmake`. Maybe better add a new file `hardening.cmake` for this.

cmake/utils.cmake
```diff
@@ -138,3 +138,32 @@ function(set_hardening_flags)
         endif()
     endif()
 endfunction()
+
+#
+# Sets the `MACRO_PREFIX_MAP_FLAG` and `DEBUG_PREFIX_MAP_FLAGS` used for `-fmacro-prefix-map` and `-fdebug-prefix-map`
+# with `-Wa,--debug-prefix-map` compiler flags respectively.
+#
+# Uses the `DEV_BUILD` option.
+function(set_prefix_map_flags)
```
locker, line 147: This doesn't look like an utility function. Why not set the flag in `compiler.cmake`?

static-build/CMakeLists.txt
```diff
@@ -53,6 +53,11 @@ set(DEPENDENCY_CFLAGS "${DEPENDENCY_CFLAGS} ${HARDENING_FLAGS}")
 set(DEPENDENCY_CXXFLAGS "${DEPENDENCY_CXXFLAGS} ${HARDENING_FLAGS}")
 set(DEPENDENCY_LDFLAGS "${DEPENDENCY_LDFLAGS} ${HARDENING_LDFLAGS}")
 
+option(DEV_BUILD "Enable build options that make debugging more convenient" OFF)
+set_prefix_map_flags()
+set(DEPENDENCY_CFLAGS "${DEPENDENCY_CFLAGS} ${MACRO_PREFIX_MAP_FLAG} ${DEBUG_PREFIX_MAP_FLAGS}")
+set(DEPENDENCY_CXXFLAGS "${DEPENDENCY_CXXFLAGS} ${MACRO_PREFIX_MAP_FLAG} ${DEBUG_PREFIX_MAP_FLAGS}")
```
locker, line 59: Why do we need it here? AFAIU we're only interested in stripping file names for the tarantool project so propagating `DEV_BUILD` to `ExternalProject_Add(tarantool)` below would be enough.

cmake/utils.cmake
Totktonada: With all my respect to the colleagues, I would like to hear arguments, not names. If it is reasonable, there should be a reason. Something like 'the alternative would be X, which has disadvantages Y and Z'.

I'll repeat my points.

* Is `tarabrt.sh` + `gdb.sh` flow becomes broken?
* Whether we decided to make user side debugging harder?

What exactly I should pass to `gdb` to make the new executable debuggable:

1. When building from sources.
2. When I installed a source package from a repository.

There is also a minor point regarding splitdebug (portage feature) -- I guess it'll not work too. But I can fix it in the ebuild.

CuriousGeorgiy: > Is tarabrt.sh + gdb.sh flow becomes broken?

Yes, it requires an extra command, `set substitute-path`.

> Whether we decided to make user side debugging harder?

What do you mean by 'user side debugging'?

> When I installed a source package from a repository.

What is a 'source package from a repository'? If I understand what you mean, I don't quite get how you debug it in the first place.

> When building from sources.

You can either turn on `DEV_BUILD` option or run the **GDB** `set substitute-path . <path to the root of tarantool source tree>`.

CuriousGeorgiy: > Something like 'the alternative would be X, which has disadvantages Y and Z'

The problem with absolute paths in debugging infromation is that they are a redundant build artefact for users who you use Tarantool in production: they are only needed by developers. 

The alternative would be to make the provided option have the opposite effect (and rename it to, let's say, `PROD_BUILD`: by default turned off, when turned on, all absolute paths in debugging information are replaced with relative ones.

CuriousGeorgiy: @locker @alyapunov RFC

Totktonada: > Yes, it requires an extra command, `set substitute-path`.

Let's add it so, otherwise we'll definitely introduce a problem. And ensure that `gdb.sh` works as with old debuginfo format as well as with the new one.

> What do you mean by 'user side debugging'?

Debugging by a user (on a package installed from our repositories). At least ability to run `gdb` and show backtraces/variables. It is sometimes helpful, when a power user tells you details about a problem on its own. I would consider non-working gdb integration as a bug. It works for other packages, why not for tarantool?

> What is a 'source package from a repository'?

https://wiki.debian.org/Packaging/SourcePackage

Totktonada: > The problem with absolute paths in debugging infromation is that they are a redundant build artefact for users who you use Tarantool in production: they are only needed by developers.

Ability to inspect an executable (including one installed from a package) using gdb without extra tarantool-specific magic is not redundant. Ability to use `tarabrt.sh` and `gdb.sh` is not redundant.

At least two bugs are introduced here.

CuriousGeorgiy:
Turned out release build is broken because of tarantool/test-run#352: blocked by tarantool/test-run#353.

cmake/utils.cmake
CuriousGeorgiy: > At least ability to run gdb and show backtraces/variables

This doesn't break **GDB** backtraces.

CuriousGeorgiy: > Ability to inspect an executable (including one installed from a package) using gdb without extra tarantool-specific magic is not redundant. Ability to use tarabrt.sh and gdb.sh is not redundant.
> 
> At least two bugs are introduced here.

@Totktonada please address your concerns to security team in [TNT-625](https://jira.vk.team/browse/TNT-625).

Totktonada: > AFAIC, this doesn't break GDB backtraces and variables, it only breaks stepping into source code.

Everything that need sources. I don't know how to inspect local variables in terms of C without sources. Anyway, gdb is much less useful without sources -- that's the point.

> @Totktonada please address your concerns to security team in [TNT-625](https://jira.vk.team/browse/TNT-625).

I asked questions regarding the patch from the author of the patch. It seems perfectly legal.

----

Let's step back to my initial questions:

> It is not recommended to show them in outgoing messages. Let's strip it from them?

> Something like 'the alternative would be X, which has disadvantages Y and Z'

>> The problem with absolute paths in debugging infromation is that they are a redundant build artefact for users who you use Tarantool in production: they are only needed by developers.

I asked to compare two approaches: your one and stripping path prefix in the error subsystem implementation. That's the alternative. Which disadvantages it has?

Regarding your answer: aside of usefulness of the 'redundant' information (see above) you incorrectly understood [CWE-497](https://cwe.mitre.org/data/definitions/497.html). There is no sense to protect system information from one, who able to read the executable file. The information shouldn't be shared with remote peers if it is unnecessary.

CuriousGeorgiy: > Anyway, gdb is much less useful without sources -- that's the point.

The point is, do you know of many users debugging tarantool and stepping into sources?

CuriousGeorgiy: > I asked questions regarding the patch from the author of the patch.

My patch is based on the interpetation our team has of the task.

CuriousGeorgiy: > I asked to compare two approaches: your one and stripping path prefix in the error subsystem implementation.

As you could read from the commit message, my patch addresses two separate concerns: 
1. replacing paths in diagnostics
2. replacing paths in debugging information

I can easily drop the second part, if we all agree on it.

Totktonada: > The point is, do you know of many users debugging tarantool and stepping into sources?

I know some. Do you know many users, which definitely didn't? This direction of the discussion will not give us any decision: we all are not representative.

> My patch is based on the interpetation our team has of the task.

That's nice (it is good to have an agreement within a working group first) and, moreover, it is legal to completely ignore my opinion here (I guess). However if you want to convince me, 'we agreed' doesn't work.

> <...> my patch addresses two separate concerns:
>
> 1. <...>
> 2. replacing paths in debugging information.

My point is that there is no reason for the second item. TNT-625 seems based on CWE-497 (or a similar recommendation) and has sense so. The second point should also have some reason under the hood. As we see, there are downsides. If there are no advantages, it is not worthful to do.

> It seems you didn't quite understand what my patch does.
> <...>
> The first part doesn't affect debugging, and I can easily drop the second part, if we all agree on it.

I don't see how, but okay. I propose to revisit a goal first.

CuriousGeorgiy: I just looked at `tools/gdb.sh`, and I don't quite understand what we are arguing about:
https://github.com/tarantool/tarantool/blob/98fcd437d9bec3f8f35304aad7ab1fdc3822ff45/tools/gdb.sh#L105

AFAIS, `set substitute-path` is used anyways, so let's just patch it, and add `-ex "set substitute-path ${SUBPATH} ." \`?

cmake/utils.cmake
CuriousGeorgiy: Fixed: moved the code to a new `cmake/hardening.cmake` module.

cmake/utils.cmake
CuriousGeorgiy: Fixed: moved the code to a new `cmake/prefix.cmake` module.

cmake/utils.cmake
CuriousGeorgiy: > As we see, there are downsides. If there are no advantages, it is not worthful to do.

As I said, the advantage is that we can make reproducible builds in the future.

CuriousGeorgiy: > I don't see how, but okay. I propose to revisit a goal first.

[`-fmacro-prefix-map`](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#Preprocessor-Options) flag is responsible for the first part and [`-fdebug-prefix-map`](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#Debugging-Options) and [`--debug-prefix-map`](https://linux.die.net/man/1/as) flags are reponsible for the second part: please visit the **GCC** and **GAS** documentation to get an idea of how they work.

locker: My 2c:
- Full paths (including the build directory) in a binary delivered to the user make no sense. Neither in error messages nor in debug info. They should be removed if possible.
- We should check that stripping the paths during build doesn't break debug info in our deb/rpm packages.
- Breaking our internal scripts is fine as long as we fix them.
- Are there any other projects that strip file paths on build?

locker:
> Turned out release build is broken because of [tarantool/test-run#352](https://github.com/tarantool/test-run/issues/352): blocked by [tarantool/test-run#353](https://github.com/tarantool/test-run/pull/353).

We can't merge this PR until the issue is fixed so I'm adding the "do not merge" label.

cmake/utils.cmake
CuriousGeorgiy: > Current implementation has its own price (broken integration with gdb) and there is no explicit plans regarding reproducible builds. Does it mean that in result we'll broke one thing and will not offer anything to a user instead?

From what I discussed with colleagues yesterday, everone say people use `set substitute-path` anyways and none really relies on the fact that source files mapping is correct. 

I'm kinda sick of this argument, I suggest we move it to server team chat.

CuriousGeorgiy: Actually, AFAIU from the regular expression here, `.` also matches it, and we don't break `tool/gdb.sh`...

https://github.com/tarantool/tarantool/blob/761574e33c514aa36f17baf426b06f67ba5c03c6/tools/gdb.sh#L96-L100

Totktonada: I setup a meeting. Anyway, there is no motivation to make this change. It doesn't give anything: neither reproducibility, nor security. I don't understand why you don't just do it for youself, nobody needs it.

> everone says people use `set substitute-path` anyways and noone really relies on the fact that source files mapping is correct.

It seems, all the discussion is about 'one person said', 'another person said' and so on. Okay. [Someone](https://stackoverflow.com/questions/67191647/using-ffile-prefix-map-breaks-debugging#comment118797400_67197241) said that it does not work. Let's say me that someone else said that it'll work. Of course, don't check it.

CuriousGeorgiy:
@ylobankov could you please verify 84e59c4 is okay?

ylobankov (approved):
Regarding update of test-run version, I am ok.

cmake/utils.cmake
CuriousGeorgiy: Tested: remapping from relative path to sources path in `tools/gdb.sh` works correctly:
Screen Shot 2022-10-20 at 13 42 26

Totktonada: Okay so. We discussed it (me, @CuriousGeorgiy, @locker, @kyukhin) and agreed to take the risk of the subtle changes and proceed with stripping of the build directory from the debug info.

locker (approved):

locker:
@Gumix Please review and approve if everything looks okay to you (I can't merge the PR without your approval).

Gumix:
Is this expected?

```sh
$ cmake .. -DCMAKE_BUILD_TYPE=Debug
-- Enabling LTO: FALSE
CMake Warning (dev) at cmake/hardening.cmake:9 (set):
  Cannot set "ENABLE_HARDENING_DEFAULT": current scope has no parent.
  CMakeLists.txt:75 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
[...]
```

Gumix:
```bash
tarantool/static-build$ cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
[...]
CMake Error at /home/iverbin/git/tarantool/cmake/profile.cmake:8 (check_library_exists):
  Unknown CMake command "check_library_exists".
  CMakeLists.txt:29 (include)
-- Configuring incomplete, errors occurred!
```

locker:
@CuriousGeorgiy don't forget to finish https://github.com/tarantool/tarantool-ee/pull/264.

PR #7816, opened 2022-10-17
box: fix fkey creation together with new field names

### LINKED ISSUE #7652: Wrong order of initialization in tuple foreign key
If a new foreign key is added along with a field that participates in that foreign field, the foreign key does not work correctly.

Reproducer:
```lua
s0 = box.schema.space.create('a', {format = {{'a', 'integer'}, {'b', 'integer'}}})
_ = s0:create_index('i0')
s1 = box.schema.space.create('b', {format = {{'a', 'integer'}}})
_ = s1:create_index('i1')
opts = {foreign_key = {one = {space = s0.id, field = {b = 'a'}}}}
format = {{name = 'a', type = 'integer'}, {name = 'b', type = 'integer'}}
box.space._space:replace({s1.id, 1, 'b', 'memtx', 0, opts, format})
```

Result:
```lua
tarantool> s1:insert({1,1})
---
- error: 'Foreign key constraint ''one'' failed: wrong local field name'
...
```

Another reproducer:
```lua
s0 = box.schema.space.create('a', {format = {{'a', 'integer'}, {'b', 'integer'}}})
_ = s0:create_index('i0')
s1 = box.schema.space.create('b', {format = {{'a', 'integer'}}})
_ = s1:create_index('i1')
opts = {foreign_key = {one = {space = s0.id, field = {b = 'a'}}}}
format = {{name = 'a', type = 'integer'}, {name = 'b', type = 'integer'}}
box.space._space:update(s1.id, {{'=', 7, format}, {'=', 6, opts}})
```

The error is the same.

If we split it into two updates, it works correctly:
```lua
s0 = box.schema.space.create('a', {format = {{'a', 'integer'}, {'b', 'integer'}}})
_ = s0:create_index('i0')
s1 = box.schema.space.create('b', {format = {{'a', 'integer'}}})
_ = s1:create_index('i1')
opts = {foreign_key = {one = {space = s0.id, field = {b = 'a'}}}}
format = {{name = 'a', type = 'integer'}, {name = 'b', type = 'integer'}}
box.space._space:update(s1.id, {{'=', 7, format}})
box.space._space:update(s1.id, {{'=', 6, opts}})
```

Gumix:
If a complex (tuple) foreign key is added along with a new field name that participates in that foreign key, the foreign key does not work correctly. This happens because the new name of a local field is added to `new_space->def->dict` only in `ModifySpace::alter()`, while before that, `new_space->def->dict` points to the old dictionary with old names (see `ModifySpace::alter_def()`). So when `local_field_no` is initialized earlier in `alter_space_do` -> `space_create` -> `tuple_constraint_fkey_init`, it's unable to find the new field number in the old dictionary.

Fix this by moving `new_def->dict = alter->old_space->def->dict;` from `ModifySpace::alter_def()` to `ModifySpace::alter()`. Note that, as before, we refer to the old dictionary from the new space (just put new names into it), because that dict is referenced by existing tuple formats.

Closes #7652

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -1058,44 +1058,43 @@ CheckSpaceFormat::prepare(struct alter_space *alter)
 /** Change non-essential properties of a space. */
 class ModifySpace: public AlterSpaceOp
 {
+	void space_swap_dictionaries(struct space_def *old_def,
+				     struct space_def *new_def,
+				     struct tuple_format *new_format);
 public:
 	ModifySpace(struct alter_space *alter, struct space_def *def)
-		:AlterSpaceOp(alter), new_def(def), new_dict(NULL) {}
+		: AlterSpaceOp(alter), new_def(def) {}
 	/* New space definition. */
 	struct space_def *new_def;
-	/**
-	 * Newely created field dictionary. When new space_def is
-	 * created, it allocates new dictionary. Alter moves new
-	 * names into an old dictionary and deletes new one.
-	 */
-	struct tuple_dictionary *new_dict;
 	virtual void alter_def(struct alter_space *alter);
 	virtual void alter(struct alter_space *alter);
 	virtual void rollback(struct alter_space *alter);
 	virtual ~ModifySpace();
 };
 
+void
+ModifySpace::space_swap_dictionaries(struct space_def *old_def,
+				     struct space_def *new_def,
+				     struct tuple_format *new_format)
+{
+	/*
+	 * When new space_def is created, it allocates new dictionary.
+	 * Move new names into an old dictionary, which already is referenced
+	 * by existing tuple formats. New dictionary object is deleted later,
+	 * in alter_space_delete.
+	 */
+	tuple_dictionary_unref(new_format->dict);
+	new_format->dict = old_def->dict;
+	tuple_dictionary_ref(new_format->dict);
+
+	SWAP(old_def->dict, new_def->dict);
+	tuple_dictionary_swap(old_def->dict, new_def->dict);
+}
+
 /** Amend the definition of the new space. */
 void
 ModifySpace::alter_def(struct alter_space *alter)
 {
-	/*
-	 * Unless it's online space upgrade, Use the old dictionary for the new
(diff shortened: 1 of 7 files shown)


src/box/alter.cc
```diff
@@ -2370,9 +2391,10 @@ on_replace_dd_space(struct trigger * /* trigger */, void *event)
 						     def->fields,
 						     def->field_count);
 		try {
-			(void) new CheckSpaceFormat(alter);
-			(void) new ModifySpace(alter, def);
-			(void) new RebuildCkConstraints(alter);
+			(void)new CheckSpaceFormat(alter);
+			(void)new ModifySpace(alter, def);
+			(void)new RebuildCkConstraints(alter);
+			(void)new RebuildTupleConstraints(alter);
```
locker, line 2397: Can we instead do this?
1. Use the new dictionary in `ModifySpace::alter_def` to create a new space.
2. Replace the space format dictionary with a new one **and** swap field names in `ModifySpace::alter`.

Gumix: Yes, done.

src/box/alter.cc
```diff
@@ -1079,23 +1079,8 @@ class ModifySpace: public AlterSpaceOp
...
-	 * For online space upgrade, all tuples fetched from the space will be
-	 * upgraded to the new format before returning to the user so it isn't
-	 * necessary. Moreover, the old tuples may be incompatible with the new
-	 * format so using the new dictionary for them would be wrong and could
-	 * result in error accessing fields by name from the space upgrade
-	 * function.
-	 */
-	if (new_def->opts.upgrade_def == NULL) {
+	if (new_def->opts.upgrade_def == NULL)
 		new_dict = new_def->dict;
```
locker, line 1083: I think we can drop the `new_dict` member, because we can check if `upgrade_def` is `NULL` right in `alter` and `rollback` methods.

Gumix: Done

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #5310, opened 2020-09-17
Some tuples missed when use sql `select` (seems degradation starts from 2.4)


filonenko-mikhail:
Tarantool version:
&gt;= 2.4

OS version:
centos

Bug description:

Some tuples missed when using `sql`

Steps to reproduce:
```lua
tarantool> box.space.test:select()
---
- - [1, 3]
  - [2, 2, 3, 4]
  - [3]
  - [4, 'asdasd', ['1', 2, true]]
...

tarantool> box.space.test:format({ { name = 'id', type = 'unsigned' } })
---
...

tarantool> \set lang sql
---
- true
...

tarantool> select * from "test";
---
- metadata:
  - name: id
    type: unsigned
  rows:
  - [1]
  - [2]
...
```

kyukhin:
Current behaviour
```lua
tarantool> box.execute([[select * from "test";]])
tarantool: src/box/sql/cursor.c:154: sqlCursorNext: Assertion `(pCur->curFlags & BTCF_TaCursor) || (pCur->curFlags & BTCF_TEphemCursor)' failed.
Aborted (core dumped)
```

PR #7819, opened 2022-10-18
sql: fix another cursor invalidation

### LINKED ISSUE #4666: sql: assertion fail on SELECT * FROM "_vfunc"
Tarantool version: 2.3.0-251-g6508ddb79.
OS version: Linux.

Clone tarantool and build:

```sh
cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_BACKTRACE=ON -DENABLE_DIST=ON -DENABLE_BUNDLED_LIBCURL=OFF && make -j
```

Run `./src/tarantool` and type in the interactive console:

```
tarantool> box.cfg{}
tarantool> \set language sql
tarantool> SELECT * FROM "_vfunc"
tarantool: /home/alex/p/tarantool-meta/r/t-1/src/box/sql/cursor.c:154: sqlCursorNext: Assertion `(pCur->curFlags & BTCF_TaCursor) || (pCur->curFlags & BTCF_TEphemCursor)' failed.
Aborted
```

Backtrace:

```
#0  0x00007f4192b5af1b in raise () from /lib64/libc.so.6
#1  0x00007f4192b44535 in abort () from /lib64/libc.so.6
#2  0x00007f4192b4440f in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#3  0x00007f4192b52722 in __assert_fail () from /lib64/libc.so.6
#4  0x000055eadb11e182 in sqlCursorNext (pCur=0x55eadcb08140, pRes=0x7f419217f3fc) at /home/alex/p/tarantool-meta/r/t-1/src/box/sql/cursor.c:154
#5  0x000055eadb0f09c3 in sqlVdbeExec (p=0x55eadcb082d8) at /home/alex/p/tarantool-meta/r/t-1/src/box/sql/vdbe.c:4252
#6  0x000055eadb0f4b6b in sqlStep (p=0x55eadcb082d8) at /home/alex/p/tarantool-meta/r/t-1/src/box/sql/vdbeapi.c:446
#7  0x000055eadb0f4c24 in sql_step (pStmt=0x55eadcb082d8) at /home/alex/p/tarantool-meta/r/t-1/src/box/sql/vdbeapi.c:474
#8  0x000055eadae3f220 in sql_execute (stmt=0x55eadcb082d8, port=0x7f419217fd60, region=0x7f4192000330) at /home/alex/p/tarantool-meta/r/t-1/src/box/execute.c:423
#9  0x000055eadae3f365 in sql_prepare_and_execute (sql=0x40898d98 "SELECT * FROM \"_vfunc\"", len=22, bind=0x0, bind_count=0, port=0x7f419217fd60, region=0x7f4192000330)
    at /home/alex/p/tarantool-meta/r/t-1/src/box/execute.c:451
#10 0x000055eadae5bca2 in lbox_execute (L=0x40878378) at /home/alex/p/tarantool-meta/r/t-1/src/box/lua/execute.c:260
#11 0x000055eadaebd9db in lj_BC_FUNCC ()
#12 0x000055eadaee07ab in lua_pcall (L=0x40878378, nargs=0, nresults=0, errfunc=0) at lj_api.c:1139
#13 0x000055eadae70451 in luaT_call (L=0x40878378, nargs=0, nreturns=0) at /home/alex/p/tarantool-meta/r/t-1/src/lua/utils.c:1036
#14 0x000055eadae67213 in lua_main (L=0x40878378, argc=1, argv=0x55eadbbd2f10) at /home/alex/p/tarantool-meta/r/t-1/src/lua/init.c:544
#15 0x000055eadae67845 in run_script_f (ap=0x7f41920003e8) at /home/alex/p/tarantool-meta/r/t-1/src/lua/init.c:643
#16 0x000055eadad48bc9 in fiber_cxx_invoke(fiber_func, typedef __va_list_tag __va_list_tag *) (f=0x55eadae6722c <run_script_f>, ap=0x7f41920003e8)
    at /home/alex/p/tarantool-meta/r/t-1/src/lib/core/fiber.h:762
#17 0x000055eadae8a13e in fiber_loop (data=0x0) at /home/alex/p/tarantool-meta/r/t-1/src/lib/core/fiber.c:830
#18 0x000055eadb09027f in coro_init () at /home/alex/p/tarantool-meta/r/t-1/third_party/coro/coro.c:110
```

Relevant assert is on the line 154:

```
149 int
150 sqlCursorNext(BtCursor *pCur, int *pRes)
151 {
152         assert(pRes != 0);
153         assert(*pRes == 0 || *pRes == 1);
154         assert((pCur->curFlags & BTCF_TaCursor) ||
155                (pCur->curFlags & BTCF_TEphemCursor));
156 
157         *pRes = 0;
158         return tarantoolsqlNext(pCur, pRes);
159 }
```

ImeevMA:
This patch fixes the issue described in issue #5310 when the tuple format has more fields than the space format. This solution is more general than the solution in 89057a21b.

Follow-up #5310
Closes #4666

NO_DOC=bugfix

### DIFF
--- a/src/box/ck_constraint.c
+++ b/src/box/ck_constraint.c
@@ -200,6 +200,7 @@ ck_constraint_on_replace_trigger(struct trigger *trigger, void *event)
 			 "field_ref");
 		return -1;
 	}
+	vdbe_field_ref_create(field_ref, space->def->field_count);
 	vdbe_field_ref_prepare_tuple(field_ref, new_tuple);
 
 	struct ck_constraint *ck_constraint;
--- a/changelogs/unreleased/gh-5310-fix-cursor-invalidation.md
+++ b/changelogs/unreleased/gh-5310-fix-cursor-invalidation.md
@@ -1,4 +1,4 @@
 ## bugfix/sql
 
 * Fixed a crash that could occur when tuples longer than specified in
-  the format were selected (gh-5310).
+  the space format were selected (gh-5310, gh-4666).
--- a/src/box/sql.c
+++ b/src/box/sql.c
@@ -1361,16 +1361,16 @@ sql_template_space_new(Parse *parser, const char *name)
 }
 
 /**
- * Initialize a new vdbe_field_ref instance with given tuple
- * data.
+ * Fill vdbe_field_ref instance with given tuple data.
+ *
  * @param field_ref The vdbe_field_ref instance to initialize.
  * @param tuple The tuple object pointer or NULL when undefined.
  * @param data The tuple data (is always defined).
  * @param data_sz The size of tuple data (is always defined).
  */
 static void
-vdbe_field_ref_create(struct vdbe_field_ref *field_ref, struct tuple *tuple,
-		      const char *data, uint32_t data_sz)
+vdbe_field_ref_fill(struct vdbe_field_ref *field_ref, struct tuple *tuple,
+		    const char *data, uint32_t data_sz)
 {
 	field_ref->tuple = tuple;
 	field_ref->data = data;
@@ -1379,13 +1379,7 @@ vdbe_field_ref_create(struct vdbe_field_ref *field_ref, struct tuple *tuple,
 	const char *field0 = data;
 	field_ref->format = NULL;
 	uint32_t mp_count = mp_decode_array(&field0);
-	if (tuple != NULL) {
-		assert(tuple_format(tuple) != NULL);
-		uint32_t field_count = tuple_format(tuple)->total_field_count;
-		field_ref->field_count = MIN(field_count, mp_count);
-	} else {
-		field_ref->field_count = mp_count;
-	}
+	field_ref->field_count = MIN(field_ref->field_capacity, mp_count);
 	field_ref->slots[0] = (uint32_t)(field0 - data);
 	memset(&field_ref->slots[1], 0,
 	       field_ref->field_count * sizeof(field_ref->slots[0]));
@@ -1397,15 +1391,22 @@ void
 vdbe_field_ref_prepare_data(struct vdbe_field_ref *field_ref, const char *data,
 			    uint32_t data_sz)
 {
-	vdbe_field_ref_create(field_ref, NULL, data, data_sz);
+	vdbe_field_ref_fill(field_ref, NULL, data, data_sz);
 }
 
 void
 vdbe_field_ref_prepare_tuple(struct vdbe_field_ref *field_ref,
 			     struct tuple *tuple)
 {
-	vdbe_field_ref_create(field_ref, tuple, tuple_data(tuple),
-			      tuple_bsize(tuple));
+	vdbe_field_ref_fill(field_ref, tuple, tuple_data(tuple),
+			    tuple_bsize(tuple));
+}
+
+void
+vdbe_field_ref_create(struct vdbe_field_ref *ref, uint32_t capacity)
+{
+	memset(ref, 0, sizeof(*ref) + capacity * sizeof(ref->slots[0]));
+	ref->field_capacity = capacity;
 }
 
 ssize_t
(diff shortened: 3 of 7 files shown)


src/box/ck_constraint.c
```diff
@@ -200,7 +200,8 @@ ck_constraint_on_replace_trigger(struct trigger *trigger, void *event)
 			 "field_ref");
 		return -1;
 	}
-	vdbe_field_ref_prepare_tuple(field_ref, new_tuple);
+	vdbe_field_ref_prepare_tuple(field_ref, space->def->field_count,
+				     new_tuple);
```
Gerold103, line 204: I don't understand. Why does it have to so complicated? How does `select *` fetch all fields? Why does it even go beyond `space->def->field_count` that you have to limit it here? Does it try to return all fields looking only at `ref->field_count`?

Gerold103 (commented):
Thanks for the patch!

src/box/ck_constraint.c
ImeevMA: Hi! Thank you for the review! The entire `struct vdbe_field_ref` is used for optimization and was introduced in commit https://github.com/tarantool/tarantool/commit/d24b3165904a69738c64456ce7ed05f06a8ae3f1. The `allocateCursor()` function allocates enough memory for `slots` that remember the shift of some fields of tuple. It is worth noting that the memory for cursor (`struct BtCursor`) is also allocated simultaneously and immediately after `slots`. In the `vdbe_field_ref_create()` function, which is called from `vdbe_field_ref_prepare_tuple()` and `vdbe_field_ref_prepare_data()`, the memory of the  `slots` is `memset()` to 0. However, the number of fields was determined from the number of fields of `msgpack`, and may be not the actual number of named fields. Thus, prior to https://github.com/tarantool/tarantool/commit/89057a21b5c1bb864f9f61adabd76b25af846a2c in case the number of named fields was less than the number of fields in `msgpack`, the `memset()` cursor was also `memset()` 0. However, in that patch, I used the number `tuple_format(tuple)->total_field_count`, which, as I found out yesterday, can also be greater than `space->def->field_count`.

The `field_count` field of the `struct vdbe_field_ref` structure must be equal to the minimum number of named fields and the number of fields in `msgpack`. It cannot be greater than the number of named fields, since in this case we will set the `memset()` cursor to 0, and it cannot be greater than the number of fields in `msgpack`, since in this case we can try to get a field that is not in `msgpack`.

According to the current logic, we cannot drop `memset()`. However, we can think of using the `slot_bitmask` field, so we don't need `memset()`. Do you think it's worth doing it?

Gerold103: Ok, I see. Thanks for the detailed explanation. It feels like we miss an important step: actual creation of a valid `vdbe_field_ref` when it is allocated. `allocateCursor()` allocates it and fills with zeros using `memset()`. But the fact is that already in `allocateCursor()` you know the max number of fields and slot count - it is `nField` parameter. The problem here is that `vdbe_field_ref` at creation doesn't save its capacity and hence you have to somehow pass same or smaller field count to `prepare_tuple/data()`.

With this implementation we can't even add assertions that we do not access slots out of borders.

I suggest to do a bit more generic rework:
- There is a new field `vdbe_field_ref.field_capacity`. `allocateCursor()` should correctly create `vdbe_field_ref` not just with `memset(0)`, but do `ref.field_capacity = nField`;
- `vdbe_field_ref_prepare_data()/_tuple()` remains like now except that it does `ref.field_count = MIN(mp_field_count, ref.field_capacity)`.

That way you don't need to pass space/format field count to `prepare()` functions and it gets safer, because the capacity is always known.

As for bitmask to avoid `memset()` - I don't mind, didn't think of that. Could be done separately as an optimization.

Gerold103 (approved):
Thanks for the patch, good work!

locker (approved):

locker:
@ImeevMA The patch doesn't apply to 2.10. Please prepare a PR for 2.10.

PR #7820, opened 2022-10-18
uri: optimize allocation of parameters and their values dynamic arrays

### LINKED ISSUE #7155: app: query parameters parsing is slow under ASAN
Version: 2.11.0-entrypoint-38-g7175f1c85.

Test case: [clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856.zip](https://github.com/tarantool/tarantool/files/8720697/clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856.zip).

Build:

```shell
CC=clang-13 CXX=clang++-13 cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_BACKTRACE=ON -DENABLE_DIST=ON -DENABLE_FEEDBACK_DAEMON=OFF -DENABLE_BUNDLED_LIBCURL=OFF -DENABLE_ASAN=ON -DENABLE_FUZZER=ON && make -j
```

Test:

```shell
$ ./test/fuzz/uri_fuzzer clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856 
<...>
Running: clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856
Executed clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856 in 25019 ms
<...>
```

Test from Lua:

```lua
tarantool> require('clock').bench(require('uri').parse, require('fio').open('clusterfuzz-testcase-minimized-uri_fuzzer-4766034870009856'):read())[1]
---
- 23.818751496
...
```

The test case has 132k query parameters with 33 unique names. Parsing of the parameters was added to `uri_create()` in 27cef5c80c8c88245591ff5d55d993ddb9602e9d (2.10.0-beta1-336-g27cef5c80). As I see from the parsing code, it intensively (re)allocates buffers. I suspect that allocations are considerably slow under ASAN.

@alyapunov asked me to file an issue on it to fix the query parameters parsing speed.

Originally reported by oss-fuzz in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44817 (will be disclosed 2022-05-19). You can find a bit more information here.

CuriousGeorgiy:
**uri: add warnings about exposure of structs in Lua via FFI**

`struct uri_param` and `struct uri` are exposed in Lua via FFI
(see src/lua/uri.lua): add warnings about the necessity of reflecting
changes to them in `ffi.cdecl`.

**uri: optimize allocation of parameters and their values dynamic arrays**

Allocation of URI parameters and their values dynamic arrays is done
inefficiently: they are reallocated each time a new parameter or parameter
value is added — grow them exponentially instead.

Closes #7155

### DIFF
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -10,9 +10,18 @@
 #define XSTRDUP(s)      (s != NULL ? xstrdup(s) : NULL)
 #define XSTRNDUP(s, n)  (s != NULL ? xstrndup(s, n) : NULL)
 
+/**
+ * WARNING: this structure is exposed in Lua via FFI (see src/lua/uri.lua): any
+ * change  must be reflected in `ffi.cdef`.
+ */
 struct uri_param {
 	/** Name of URI parameter. */
 	char *name;
+	/**
+	 * Capacity of URI parameter values dynamic array (used for exponential
+	 * reallocation).
+	 */
+	int values_capacity;
 	/** Count of values for this parameter. */
 	int value_count;
 	/** Array of values for this parameter. */
@@ -38,8 +47,13 @@ uri_find_param(const struct uri *uri, const char *name)
 static void
 uri_param_add_value(struct uri_param *param, const char *value)
 {
-	size_t size = (param->value_count + 1) * sizeof(char *);
-	param->values = xrealloc(param->values, size);
+	assert(param->value_count <= param->values_capacity);
+	if (param->value_count == param->values_capacity) {
+		param->values_capacity = param->values_capacity == 0 ? 16 :
+					 param->values_capacity * 2;
+		size_t size = param->values_capacity * sizeof(char *);
+		param->values = xrealloc(param->values, size);
+	}
 	param->values[param->value_count++] = xstrdup(value);
 }
 
@@ -65,6 +79,7 @@ uri_param_create(struct uri_param *param, const char *name)
 	param->name = xstrdup(name);
 	param->values = NULL;
 	param->value_count = 0;
+	param->values_capacity = 0;
 }
 
 /**
@@ -74,9 +89,10 @@ static void
 uri_param_copy(struct uri_param *dst, const struct uri_param *src)
 {
 	dst->name = xstrdup(src->name);
+	dst->values_capacity = src->values_capacity;
 	dst->value_count = src->value_count;
-	dst->values = (src->value_count == 0 ? NULL :
-		       xmalloc(src->value_count * sizeof(*dst->values)));
+	dst->values = (src->values_capacity == 0 ? NULL :
+		       xmalloc(src->values_capacity * sizeof(*dst->values)));
 	for (int i = 0; i < src->value_count; i++)
 		dst->values[i] = xstrdup(src->values[i]);
 }
@@ -99,9 +115,14 @@ uri_add_param(struct uri *uri, const char *name, const char *value)
 {
 	struct uri_param *param = uri_find_param(uri, name);
 	if (param == NULL) {
-		size_t size = (uri->param_count + 1) *
-			sizeof(struct uri_param);
-		uri->params = xrealloc(uri->params, size);
+		assert(uri->param_count <= uri->params_capacity);
+		if (uri->param_count == uri->params_capacity) {
+			uri->params_capacity = uri->params_capacity == 0 ? 16 :
+					       uri->params_capacity * 2;
+			size_t size =
+				uri->params_capacity * sizeof(struct uri_param);
+			uri->params = xrealloc(uri->params, size);
+		}
 		param = &uri->params[uri->param_count++];
 		uri_param_create(param, name);
 	}
@@ -161,9 +182,10 @@ uri_copy(struct uri *dst, const struct uri *src)
 	dst->query = XSTRDUP(src->query);
 	dst->fragment = XSTRDUP(src->fragment);
 	dst->host_hint = src->host_hint;
+	dst->params_capacity = src->params_capacity;
 	dst->param_count = src->param_count;
-	dst->params = (src->param_count == 0 ? NULL :
-		       xmalloc(src->param_count * sizeof(*dst->params)));
+	dst->params = (src->params_capacity == 0 ? NULL :
+		       xmalloc(src->params_capacity * sizeof(*dst->params)));
 	for (int i = 0; i < src->param_count; i++)
 		uri_param_copy(&dst->params[i], &src->params[i]);
 }
--- a/src/lib/uri/uri.h
+++ b/src/lib/uri/uri.h
@@ -15,6 +15,10 @@ extern "C" {
 
 struct uri_param;
 
+/**
+ * WARNING: this structure is exposed in Lua via FFI (see src/lua/uri.lua): any
+ * change  must be reflected in `ffi.cdef`.
+ */
 struct uri {
 	char *scheme;
 	char *login;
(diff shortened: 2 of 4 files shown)


src/lib/uri/uri.c
```diff
@@ -42,8 +47,13 @@ uri_find_param(const struct uri *uri, const char *name)
...
 {
-	size_t size = (param->value_count + 1) * sizeof(char *);
-	param->values = xrealloc(param->values, size);
+	assert(param->value_count <= param->values_capacity);
+	if (param->value_count == param->values_capacity) {
+		param->values_capacity = param->values_capacity == 0 ? 16 :
+					 param->values_capacity * 2;
+		size_t size = param->values_capacity * sizeof(char *);
+		param->values = xrealloc(param->values, size);
+	}
```
locker, line 56: We quite often use exponentially growing arrays. Even in this patch alone, there are two such places. Maybe, add a helper to `trivia/util.h` that would add an entry to such an array, relocating it if necessary? Or even a new data structure for dynamically growing arrays?

src/lib/uri/uri.h
```diff
@@ -15,6 +15,10 @@ extern "C" {
 
 struct uri_param;
 
+/**
+ * WARNING: this structure is exposed in Lua via FFI (see src/lua/uri.lua): any
+ * change  must be reflected in `ffi.cdef`.
+ */
```
locker, line 21: Nit: I see no point in factoring out this change into a separate patch (it touches the same structures that are patched in the main patch; it introduces no changes). Let's merge this change into the main patch.

src/lib/uri/uri.c
CuriousGeorgiy: I guess that's the burden of using C (as opposed to C++, where we could simply use `std::vector` everywhere).

I suppose we just open a refactoring ticket regarding this problem.

src/lib/uri/uri.h
CuriousGeorgiy: This change has no logical connection with the optimization patch, that's why I put it into a separate commit.

src/lib/uri/uri.c
```diff
@@ -74,6 +89,7 @@ static void
 uri_param_copy(struct uri_param *dst, const struct uri_param *src)
 {
 	dst->name = xstrdup(src->name);
+	dst->values_capacity = src->values_capacity;
 	dst->value_count = src->value_count;
 	dst->values = (src->value_count == 0 ? NULL :
 		       xmalloc(src->value_count * sizeof(*dst->values)));
```
Totktonada, line 95: Allocated size does not correspond to the capacity (see also a PoC patch in the issue).

src/lib/uri/uri.h
locker: My point is that there's no point in splitting this out - it doesn't ease code review, it doesn't ease cherry-picking, it doesn't help bisecting. You basically add comments to structures that you patch in the same commit. Splitting it out is useless.

src/lib/uri/uri.c
CuriousGeorgiy: Oh, I wasn't conscious enough about it, fixed it (and `uri_copy` too), thanks!

src/lib/uri/uri.h
CuriousGeorgiy: Merged it, though I can't say I completely agree with you.

src/lib/uri/uri.c
```diff
@@ -10,9 +10,18 @@
 #define XSTRDUP(s)      (s != NULL ? xstrdup(s) : NULL)
 #define XSTRNDUP(s, n)  (s != NULL ? xstrndup(s, n) : NULL)
 
+/**
+ * WARNING: this structure is exposed in Lua via FFI (see src/lua/uri.lua): any
+ * change  must be reflected in `ffi.cdef`.
+ */
```
Totktonada, line 16: Nit: `WARNING`, `CAUTION`, `READ IT`, `IMPORTANT` and so on are redundant for my taste. One either read comments or doesn't.

Totktonada: Feel free to resolve the discussion without any action, it is just about my personal feeling.

src/lib/uri/uri.c
locker: I don't think that using `ibuf` would make this code any more efficient, because we have to allocate the final result with malloc anyway.

Totktonada: I don't mean just place those arrays of pointers into an `ibuf` or a `region`. How about eliminating a 10^5 `strdup()` calls (on given test case)?

locker: Hmm, can we use [`grp_alloc`](https://github.com/tarantool/tarantool/blob/master/src/lib/salad/grp_alloc.h) here?

locker: > Hmm, can we use [`grp_alloc`](https://github.com/tarantool/tarantool/blob/master/src/lib/salad/grp_alloc.h) here?

Probably not.

> I don't mean just place those arrays of pointers into an `ibuf` or a `region`. How about eliminating a 10^5 `strdup()` calls (on given test case)?

I see your point. Those `strdup` look bad, and using a dynamic array here wouldn't help us eliminate them. Okay, I'm fine with leaving this code as is (with explicit realloc, that is). Maybe, we'll add a dynamic array some time later.

locker (approved):

src/lib/uri/uri.c
Totktonada: I'm not sure that it'll be optimal to implement a two phare parsing. I guess that a one phase parsing with ahead memory allocation (for all those small pieces of data) would win. At least (in my head) it should be good for the parsing itself (as opposite to how we return it for Lua).

Maybe we can do something even more exciting: eliminate string copying and just reference parts of the original strings. Somethings like https://github.com/rapidlua/jstr, but without destruction of the original string (track both start and end of a chunk). Or with a single copy before the destructive parsing.

To be honest, I don't know, whether we can do something smart for returning values into Lua. Maybe create strings on demand? Don't know.

Totktonada: (I didn't see the last your comment, when wrote mine one.)

locker: > Maybe we can do something even more exciting: eliminate string copying and just reference parts of the original strings.

The problem is that uri parameters can be added dynamically, after parsing the initial string (note that `uri_add_param` is a public method). This is used to pass extra uri parameters in a table.

Totktonada: > The problem is that uri parameters can be added dynamically, after parsing the initial string (note that `uri_add_param` is a public method). This is used to pass extra uri parameters in a table.

Good point. I guess we still can speed up parsing of a complete URI, but dynamic adding parameters will not be so optimal.

locker: >  I guess we still can speed up parsing of a complete URI.

Maybe later. This patch is already a good step forward, because it fixes the issue at hand. As of eliminating `strdup`, let's postpone it until somebody complains. After all, one would hardly pass thousands of uri parameters in real life. That said, I think this PR is ready for merge.

CuriousGeorgiy: > I don't mean just place those arrays of pointers into an ibuf or a region. How about eliminating a 10^5 strdup() calls (on given test case)?

Yeah, I actually thought that this impacts performance the most. Turned out not to be in this specific case, but it's still a performance issue.

Could you please open a supplementary ticket?

src/lib/uri/uri.c
Totktonada: Okay so, closing.

src/lib/uri/uri.c
Totktonada: > This patch is already a good step forward, because it fixes the issue at hand.

Of course, this PR is about one specific problem and it fixes the problem.

PR #5443, opened 2020-10-27
Change 'force_recovery' option behavior and makes tarantool to ignore some errors during snapshot recovery just like during xlog recovery

mechanik20051988:
Change 'force_recovery' option behavior and makes tarantool to ignore
some errors during snapshot recovery just like during xlog recovery.

Issue #5422, opened 2020-10-13
Snapshot force recovery


alyapunov:
There is an option `force_recovery` that makes tarantool to ignore some problems that happened during recovery. But that option is only used during xlog recovery - problems in snapshot cannot be ignored. That mean that if a snapshot is broken somehow - it cannot be loaded in any way even if the user truly understands the risks.
I think we should make another similar option that will allow loading of broken snapshot.

PR #7824, opened 2022-10-18
box: panic if snapshot has no system spaces during recovery

### LINKED ISSUE #7800: Assertion failed during recovery from snapshot
**Steps to reproduce**

1. Launch **tarantool** with [00000000000000000000.snap](https://github.com/tarantool/tarantool/files/9766558/00000000000000000000.snap.zip).
2. Type `box.cfg{}` into console.

**Actual behavior**

```console
Assertion failed: (e != NULL), function diag_raise, file diag.h, line 475.
```

**Expected behavior**

**tarantool** does not crash.

Gumix:
Currently, if a snapshot contains some correct entries, but doesn't include system spaces, Tarantool crashes with segmentation fault, or for Debug build:
```void diag_raise(): Assertion `e != NULL' failed.```
This happens because `memtx_engine_recover_snapshot` returns -1, while diag is not set. Let's panic instead of a crash.

Closes #7800

### CODE REVIEW

File: src/box/memtx_engine.cc [hunk 1]
```diff
@@ -278,6 +278,9 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 			say_error("snapshot `%s' has no EOF marker", cursor.name);
 	}
 
+	if (is_space_system < 0)
+		panic("snapshot `%s' has no system spaces", cursor.name);
```
locker, line 282: `is_space_system` < 0 means that the snapshot contains no spaces at all. If it contains at least one space (system or user), `is_space_system` will be >= 0, and the check will pass. IMO this check makes no sense. Looks like the intention of commit 0be1243ecc8a0b56b079d82d4c713982a05382f0, which added the check, was to fail recovery if at least one system space is missing, but adding such a check would be complicated. Let's just drop it? While we are at it, we should change the type of `is_space_system` to bool.

TBH I find the aforementioned commit dubious - according to its description (and #5422), the intention was to enable the `force_recovery` option for snapshots, but it was in fact already enabled. The only useful thing the commit did was enable recovery of truncated snapshots (without EOF marker). For some reason, it also added some broken checks that disabled the `force_recovery` option for system spaces. Maybe, just revert that part of the commit?

Gumix: `force_recovery` disabling for system spaces actually works ok, [it was requested here](https://github.com/tarantool/tarantool/pull/5443#issuecomment-717833930), so let's keep it as is.

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #7000, opened 2022-04-06
Introduce compat module


Lord-KA:
As a followup to #6200 and a result of #6912 it was decided to implement comparability module in Tarantool to simplify introduction of debatable behavior changes.

PR #7825, opened 2022-10-18
fiber: add channel graceful close option to compat

### LINKED ISSUE #7746: Introduce fiber.channel.shutdown
@Totktonada: The product task is [here](https://www.notion.so/tarantool/Fiber-channel-close-a6ead56417724da98fdf972912ed2393) (it is not public, sorry). I'll add details after internal discussions. The short problem statement is that `<fiber channel object>:close()` discards all messages and sometimes we want an action like 'close when all messages will arrive' (it is similar to how channels in Go works).

Lord-KA:
Before the change there was an unexpected behavior when using
channel:close(), as it closed the channel entirely and discarded all
unread events.

This commit introduces graceful channel close option in tarantool.compat
(https://github.com/tarantool/tarantool/issues/7000) that allowes to select new or old behavior.

With the new behavior `close()` marks channel as closed for writing.
Only when all events are extracted, the channel is closed entirely. If
there are no events in the channel, it is closed as usual.

Document that describes new API can be found on notion:
https://www.notion.so/fiber-channel-graceful-close-53b2788ed1f144598c4c0e1229c2eb69

Requires https://github.com/tarantool/tarantool/pull/7060
Closes https://github.com/tarantool/tarantool/issues/7746
See also https://github.com/tarantool/tarantool/issues/7000
NO_DOC=docs on tarantool.wiki

### DIFF
--- a/changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
+++ b/changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
@@ -1,0 +1,6 @@
+## feature/core/fiber
+
+* Fiber channel now has a compatibility option in tarantool.compat, that
+  provides transition from force channel `close()` to graceful `close()`, i.e.
+  closing channel for writing while existing messages can still be extracted
+  (gh-7746).
--- a/src/lib/core/fiber_channel.c
+++ b/src/lib/core/fiber_channel.c
@@ -50,12 +50,27 @@ struct ipc_wait_pad {
 	enum fiber_channel_wait_status status;
 };
 
+/**
+ * If fiber channel should work in a graceful close mode,
+ * i.e. become write-closed instead of destroyed. When buffer
+ * is emptied by recipient, the channel is finally destroyed.
+ */
+static enum fiber_channel_close_mode fiber_channel_close_mode =
+	FIBER_CHANNEL_CLOSE_FORCEFUL;
+
+void
+fiber_channel_set_close_mode(enum fiber_channel_close_mode mode)
+{
+	fiber_channel_close_mode = mode;
+}
+
 void
 fiber_channel_create(struct fiber_channel *ch, uint32_t size)
 {
 	ch->size = size;
 	ch->count = 0;
 	ch->is_closed = false;
+	ch->is_destroyed = false;
 	rlist_create(&ch->waiters);
 	if (ch->size) {
 		ch->buf = (struct ipc_msg **) (char *) &ch[1];
@@ -164,7 +179,7 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
 {
 	/*
 	 * Preconditions of waiting are:
-	 * - the channel is not closed,
+	 * - the channel is not destroyed,
 	 * - the current fiber has not been
 	 *   cancelled,
 	 * - the timeout has not expired.
@@ -173,7 +188,7 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
 	 * caller, since ev_now() does not get updated without
 	 * a yield.
 	 */
-	if (ch->is_closed) {
+	if (ch->is_destroyed) {
 		diag_set(ChannelIsClosed);
 		return -1;
 	}
@@ -191,7 +206,49 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
 void
 fiber_channel_close(struct fiber_channel *ch)
 {
-	if (ch->is_closed)
+	/*
+	 * When current value <fiber_channel_close_mode> equals to
+	 * FIBER_CHANNEL_CLOSE_FORCEFUL, just destroy the channel
+	 * with its contents. The channel is effectively dead as a
+	 * result of <fiber_channel_close>.
+	 */
+	if (fiber_channel_close_mode == FIBER_CHANNEL_CLOSE_FORCEFUL)
+		goto destroy;
+	/*
+	 * When current value <fiber_channel_close_mode> equals to
+	 * FIBER_CHANNEL_CLOSE_GRACEFUL, check whether the channel
+	 * is empty. In this case the channel can be also simply
+	 * destroyed. Otherwise, just mark the channel to make it
+	 * "read-only": no new messages can be added, but all
+	 * existing messages can be received.
+	 * All waiting writers receive an error in this case.
+	 */
+	if (fiber_channel_is_empty(ch))
+		goto destroy;
+	ch->is_closed = true;
+	/* Wake up all channel_put(), they will automatically cancel. */
+	struct fiber *f;
+	/*
+	 * There is a couple of reasons to process all waiters
+	 * unconditionally:
+	 * * A channel has either readers or writers (if any)
+	 *   simultaneously
+	 * * If the buffer has any message (see the condition
+	 *   above), the channel can't have a reader
+	 */
+	while (!rlist_empty(&ch->waiters)) {
+		f = rlist_first_entry(&ch->waiters, struct fiber, state);
+		fiber_channel_waiter_wakeup(f, FIBER_CHANNEL_WAIT_CLOSED);
+	}
+	return;
+destroy:
+	fiber_channel_destroy(ch);
+}
+
+void
+fiber_channel_destroy(struct fiber_channel *ch)
+{
+	if (ch->is_destroyed)
 		return;
 
 	while (ch->count) {
@@ -204,13 +261,9 @@ fiber_channel_close(struct fiber_channel *ch)
 		f = rlist_first_entry(&ch->waiters, struct fiber, state);
 		fiber_channel_waiter_wakeup(f, FIBER_CHANNEL_WAIT_CLOSED);
 	}
-	ch->is_closed = true;
-}
 
-void
-fiber_channel_destroy(struct fiber_channel *ch)
-{
-	fiber_channel_close(ch);
+	ch->is_closed = true;
+	ch->is_destroyed = true;
 }
 
 void
@@ -290,6 +343,11 @@ fiber_channel_put_msg_timeout(struct fiber_channel *ch,
 	ev_tstamp start_time = ev_monotonic_now(loop());
 
 	while (true) {
+		/* Exit if channel is already closed for writing. */
+		if (ch->is_closed) {
+			diag_set(ChannelIsClosed);
+			return -1;
+		}
 		/*
 		 * Check if there is a ready reader first, and
 		 * only if there is no reader try to put a message
@@ -304,10 +362,10 @@ fiber_channel_put_msg_timeout(struct fiber_channel *ch,
 			/*
 			 * There can be no reader if there is
 			 * a buffered message or the channel is
-			 * closed.
+			 * destroyed.
 			 */
 			assert(ch->count == 0);
-			assert(ch->is_closed == false);
+			assert(!ch->is_destroyed);
 
 			struct fiber *f = rlist_first_entry(&ch->waiters,
 							    struct fiber,
@@ -325,13 +383,11 @@ fiber_channel_put_msg_timeout(struct fiber_channel *ch,
 			 */
 
 			/*
-			 * Closed channels, are, well, closed,
-			 * even if there is space in the buffer.
+			 * Destroyed channels, are already closed,
+			 * thus, it will exit in the check above.
 			 */
-			if (ch->is_closed) {
-				diag_set(ChannelIsClosed);
-				return -1;
-			}
+			assert(!ch->is_destroyed);
+
 			fiber_channel_buffer_push(ch, msg);
 			return 0;
 		}
@@ -399,10 +455,10 @@ fiber_channel_get_msg_timeout(struct fiber_channel *ch,
 		if (ch->count > 0) {
 			/**
 			 * There can't be any buffered stuff in
-			 * a closed channel - everything is
-			 * destroyed at close.
+			 * a destroyed channel - everything is
+			 * destroyed too.
 			 */
-			assert(ch->is_closed == false);
+			assert(!ch->is_destroyed);
 
 			*msg = fiber_channel_buffer_pop(ch);
 
@@ -420,6 +476,8 @@ fiber_channel_get_msg_timeout(struct fiber_channel *ch,
 				fiber_channel_waiter_wakeup(f,
 					FIBER_CHANNEL_WAIT_DONE);
 			}
+			if (ch->count == 0 && ch->is_closed)
+				fiber_channel_destroy(ch);
 			return 0;
 		}
 		if (fiber_channel_has_writers(ch)) {
--- a/src/lib/core/fiber_channel.h
+++ b/src/lib/core/fiber_channel.h
@@ -104,15 +104,19 @@ ipc_value_new(void);
  *
  * A channel, once created is "open". I.e. anyone can read or
  * write to/from a channel. A channel can be closed at any time,
- * in which case, all messages currently buffered in a channel
- * are destroyed, waiting readers or writers awoken with an error.
+ * in which case, if `fiber_channel_close_mode` is set to
+ * FIBER_CHANNEL_CLOSE_GRACEFUL, it is marked as closed
+ * (read-only) and all messages currently buffered can be
+ * delivered, otherwise channel is marked as destroyed and all its
+ * contents are discarded, waiting readers or writers awoken with
+ * an error.
  *
  * Waiting for a message, a reader, or space in a buffer can also
  * return error in case of a wait timeout or cancellation (when the
  * waiting fiber is cancelled).
  *
  * Sending a message to a closed channel, as well as reading
- * a message from such channel, always fails.
+ * a message from destroyed channel, always fails.
  *
  * Channel memory layout
  * ---------------------
@@ -137,8 +141,16 @@ struct fiber_channel {
 	struct rlist waiters;
 	/** Ring buffer read position. */
 	uint32_t beg;
-	/* True if the channel is closed. */
+	/**
+	 * True if the channel is closed for writing and
+	 * is waiting to be destroyed or has been destroyed.
+	 */
 	bool is_closed;
+	/**
+	 * True if the channel forbids both reading and writing and
+	 * the buffer is no longer accessible.
+	 */
+	bool is_destroyed;
 	/** Channel buffer, if any. */
 	struct ipc_msg **buf;
 };
@@ -206,6 +218,55 @@ fiber_channel_is_empty(struct fiber_channel *ch)
 {
 	return ch->count == 0;
 }
+
+/** The fiber channel has two ways to end its life. */
+enum fiber_channel_close_mode {
+	/**
+	 * The forceful close mode means that fiber_channel_close()
+	 * discards all stored messages, wake ups all readers and
+	 * writers and forbids to read or write messages. The
+	 * channel is effectively died after closing.
+	 *
+	 * fiber_channel_close() works exactly as
+	 * fiber_channel_destroy() when this mode is in effect.
+	 *
+	 * This mode represents original behavior of the fiber
+	 * channel since its born and hence it is the default.
+	 */
+	FIBER_CHANNEL_CLOSE_FORCEFUL,
+	/**
+	 * The graceful close mode means that fiber_channel_close()
+	 * only marks the channel as read-only: no new messages can
+	 * be added, but all buffered messages can be received.
+	 * All waiting writers receive an error.
+	 *
+	 * A closed channel is automatically destroyed when all
+	 * buffered messages are read.
+	 *
+	 * This mode is considered as more safe and recommended for
+	 * users. The main idea behind it is to prevent accidental
+	 * data loss or, in other words, fit users' expectation that
+	 * all accepted messages will be delivered. This reflects
+	 * how channels work in other programming languages, for
+	 * example, in Go.
+	 */
+	FIBER_CHANNEL_CLOSE_GRACEFUL,
+};
+
+/**
+ * Choose how the channels end its life.
+ *
+ * Affects all channels.
+ *
+ * The behavior is unspecified for already created channels.
+ * Choose the mode at an early stage of application's
+ * initialization.
+ *
+ * @sa enum fiber_channel_close_mode
+ * @sa fiber_channel_close
+ */
+void
+fiber_channel_set_close_mode(enum fiber_channel_close_mode mode);
 
 /**
  * Check if the channel buffer is full.
@@ -351,15 +412,15 @@ fiber_channel_count(struct fiber_channel *ch)
 }
 
 /**
- * Close the channel. Discards all messages
- * and wakes up all readers and writers.
+ * Close the channel for writing. If `fiber_channel_close_mode` is
+ * FIBER_CHANNEL_CLOSE_FORCEFUL, discards all messages and wakes
+ * up all readers and writers.
  */
 void
 fiber_channel_close(struct fiber_channel *ch);
 
 /**
- * True if the channel is closed for both for reading
- * and writing.
+ * True if the channel is closed for writing or destroyed.
  */
 static inline bool
 fiber_channel_is_closed(struct fiber_channel *ch)
--- a/src/lua/compat.lua
+++ b/src/lua/compat.lua
@@ -32,6 +32,18 @@ with some existing applications that rely on the old style.
 https://github.com/tarantool/tarantool/wiki/compat%3Ayaml_pretty_multiline
 ]]
 
+local FIBER_CHANNEL_GRACEFUL_CLOSE_BRIEF = [[
+Whether fiber channel should be marked read-only on close, instead of being
+destroyed. The new behavior allows user to mark the channel as read-only. This
+way if a message was in the buffer before close, it will remain accessible after
+close. The read-only channel will be closed automatically when the buffer
+becomes empty, or will be deleted by GC if all links to it are lost, as before.
+The new behavior can break code that relies on `ch:get()` returning `nil` after
+channel close.
+
+https://github.com/tarantool/tarantool/wiki/compat%3Afiber_channel_close_mode
+]]
+
 -- Contains options descriptions in following format:
 -- * default  (string)
 -- * brief    (string)
@@ -57,6 +69,12 @@ local options = {
         obsolete = nil,
         brief = YAML_PRETTY_MULTILINE_BRIEF,
         action = internal.yaml_pretty_multiline_toggle,
+    },
+    fiber_channel_close_mode = {
+        default = 'old',
+        obsolete = nil,
+        brief = FIBER_CHANNEL_GRACEFUL_CLOSE_BRIEF,
+        action = internal.fiber_channel_close_mode_toggle,
     },
 }
 
--- a/test/app-luatest/gh_7746_fiber_channel_close_test.lua
+++ b/test/app-luatest/gh_7746_fiber_channel_close_test.lua
@@ -1,0 +1,133 @@
+local compat = require('tarantool').compat
+local t = require('luatest')
+local g = t.group()
+local fiber = require('fiber')
+
+g.test_channel_close_default = function()
+    -- No graceful close (old) by default.
+    local ch = fiber.channel(10)
+    ch:put('one')
+    ch:put('two')
+    t.assert_equals(ch:get(), 'one')
+    ch:close()
+    t.assert(ch:is_closed())
+    t.assert_not(ch:put('three'))
+    t.assert_equals(ch:get(), nil)
+end
+
+g.test_channel_close_old = function()
+    compat.fiber_channel_close_mode = 'old'
+    local ch = fiber.channel(10)
+    ch:put('one')
+    ch:put('two')
+    t.assert_equals(ch:get(), 'one')
+    ch:close()
+    t.assert(ch:is_closed())
+    t.assert_not(ch:put('three'))
+    t.assert_equals(ch:get(), nil)
+end
+
+g.test_channel_close_new = function()
+    -- Graceful close (new) selected.
+    compat.fiber_channel_close_mode = 'new'
+    local ch = fiber.channel(10)
+    ch:put('one')
+    ch:put('two')
+    ch:close()
+
+    t.assert(ch:is_closed())
+    t.assert_not(ch:put('three'))
+
+    t.assert_equals(ch:get(), 'one')
+    t.assert(ch:is_closed())
+
+    t.assert_equals(ch:get(), 'two')
+    t.assert(ch:is_closed())
+
+    t.assert_equals(ch:get(), nil)
+end
+
+g.test_close_graceful_on_empty = function()
+    compat.fiber_channel_close_mode = 'new'
+    local ch = fiber.channel(10)
+    ch:put('one')
+    ch:get()
+    ch:close()
+    t.assert(ch:is_closed())
+end
+
+local test_close_reader_payload = function(mode)
+    compat.fiber_channel_close_mode = mode
+    local ch = fiber.channel(0)
+    local reader = fiber.new(function()
+        t.assert_not(ch:is_closed(),
+                     'reader tries to read from the open channel')
+        -- Try to obtain the message from the zero-length
+        -- channel.
+        -- XXX: <reader> fiber hangs forever, until one of the
+        -- following occurs:
+        -- * <fiber_channel.put> is called from another fiber
+        -- * <channel> is closed from another fiber
+        -- For the latter case <fiber_channel.get> fails (i.e.
+        -- yields nil value).
+        t.assert_is(ch:get(), nil,
+                    'reader fails to read from the zero-length channel')
+        t.assert(ch:is_closed(), 'reader hangs until channel is closed')
+        t.assert_is(ch:get(0), nil,
+                    'reader fails to read from the closed channel')
+    end)
+    reader:set_joinable(true)
+    -- Yield to start tests in <reader> payload.
+    fiber.yield()
+    ch:close()
+    t.assert(ch:is_closed())
+    -- Wait until tests in <writer> payload are finished.
+    local status, err = fiber.join(reader)
+    t.assert(status, err)
+end
+
+g.test_close_reader_old = function()
+    test_close_reader_payload('old')
+end
+
+g.test_close_reader_new = function()
+    test_close_reader_payload('new')
+end
+
+local test_close_writer_payload = function(mode)
+    compat.fiber_channel_close_mode = mode
+    local ch = fiber.channel(0)
+    local writer = fiber.new(function()
+        t.assert_not(ch:is_closed(),
+                     'writer tries to write to the open channel')
+        -- Try to push the message into the zero-length
+        -- channel.
+        -- XXX: <writer> fiber hangs forever, until one of the
+        -- following occurs:
+        -- * <fiber_channel.get> is called from another fiber
+        -- * <channel> is closed from another fiber
+        -- For the latter case <fiber_channel.put> fails (i.e.
+        -- yields false value).
+        t.assert_not(ch:put('one'),
+                     'writer fails to write to the zero-length channel')
+        t.assert(ch:is_closed(), 'writer hangs until channel is closed')
+        t.assert_not(ch:put('one'),
+                     'writer fails to write to the closed channel')
+    end)
+    writer:set_joinable(true)
+    -- Yield to start tests in <writer> payload.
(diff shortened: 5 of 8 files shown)


changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
```diff
@@ -0,0 +1,5 @@
+## feature/fiber
```
Totktonada, line 1: > NO_DOC=docs on tarantool.wiki

There is [API description](https://www.tarantool.io/en/doc/latest/reference/reference_lua/fiber/#channels) on the website and it seems it should be updated after the patch. We should add a documentation request. Let's mention existing documentation inside.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
```diff
@@ -0,0 +1,5 @@
...
+
+* Fiber channel now has a compatibility option in tarantool.compat, that
+  provides transition from force channel `close()` to graceful `close()`, i.e.
+  closing channel for writing while existing messages can still be extracted.
```
Totktonada, line 5: Please, leave an issue number `<...> (gh-xxxx).` at the end.

src/lib/core/fiber_channel.h
```diff
@@ -351,15 +362,14 @@ fiber_channel_count(struct fiber_channel *ch)
 }
 
 /**
- * Close the channel. Discards all messages
- * and wakes up all readers and writers.
+ * Close the channel for writing. If not graceful, discards all
+ * messages and wakes up all readers and writers.
```
Totktonada, line 366: 'Is graceful' here without a context may be confusing. I would at least link a setter for the flag.

src/lib/core/fiber_channel.h
```diff
@@ -207,6 +211,13 @@ fiber_channel_is_empty(struct fiber_channel *ch)
 	return ch->count == 0;
 }
 
+/**
+ * If fiber channel should work in a graceful close mode,
+ * i.e. become write-closed instead of closed until buffer
+ * is emitted by recipient. Is set by tarantool.compat.
+ */
+extern bool fiber_channel_graceful_close;
```
Totktonada, line 219: I vote for getter and setter functions. Anyone up for it?

src/lib/core/fiber_channel.c
```diff
@@ -41,6 +41,8 @@ enum fiber_channel_wait_status {
 	FIBER_CHANNEL_WAIT_CLOSED /* Wait is aborted, the channel is closed. */
 };
 
+bool fiber_channel_graceful_close = false;
```
Totktonada, line 44: How about a unit test?

src/lua/compat.lua
```diff
@@ -15,14 +14,32 @@ local options_format = {
     action         = 'function/nil',
 }
 
+local FIBER_CHANNEL_GRACEFUL_CLOSE_BRIEF = [[
+Whether fiber channel should discard all messages in its buffer after being
+closed. New behavior allows user to get all messages that were put into buffer
+before close afterwards, by marking channel as read-only. The read-only channel
```
Totktonada, line 20: > before close afterwards

Oh my brain... Can you explain it like I'm five, please?

test/app-luatest/gh_7746_fiber_channel_close_test.lua
```diff
@@ -0,0 +1,62 @@
+local compat = require('tarantool').compat
```
Totktonada, line 1: What just comes into my mind: shouldn't we run all test cases that lean on `compat` in separate tarantool instances (using `server:exec()`)? A change in a global state shouldn't leak into other tests.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
```diff
@@ -0,0 +1,62 @@
...
+    ch:get()
+    ch:close()
+    t.assert(ch:is_closed())
+
+    -- Check explicitly true.
+    ch = fiber.channel(10)
+    ch:put('one')
+    ch:get()
+    ch:close()
+    t.assert(ch:is_closed())
```
Totktonada, line 61: I see no difference from the previous hunk. What is the point here?

src/lib/core/fiber_channel.h
Lord-KA: It is purely internal, why encapsulate one more time?

Totktonada: It is how we write the code in tarantool.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
Totktonada: A fiber or a fiber channel? TBH, I still don't understand. Can you point to those different branches in the code?

Lord-KA: Tipo, fiber channel. When there is a message in buffer and it is removed, the `channel_destroy()` is called from `get()` and when buffer is empty on `close()`, destroy is called from from `channel_close()`.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
Lord-KA: Done.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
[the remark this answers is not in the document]
Lord-KA: Changed to `feature/core/fiber`.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
Lord-KA: I misinterpreted your question, yest the latter part of the `on_empty` test is redundant. Removed it.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
Lord-KA: Created another [issue](https://github.com/tarantool/tarantool/issues/8033) for this.

src/lib/core/fiber_channel.h
Lord-KA:
```
  /**
   * Close the channel for writing. If not `fiber_channel_graceful_close`, 
   * discards all messages and wakes up all readers and writers.                 
   */                                                                            
```

src/lib/core/fiber_channel.h
Lord-KA: Added `fiber_channel_set_graceful_close(bool is_graceful)`.

src/lua/compat.lua
Lord-KA: Done.

src/lib/core/fiber_channel.c
```diff
@@ -190,6 +193,23 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
 
 void
 fiber_channel_close(struct fiber_channel *ch)
+{
+	if (!fiber_channel_graceful_close || ch->count == 0) {
+		/* Destroy if no graceful close or buffer is already empty. */
+		fiber_channel_destroy(ch);
+		return;
+	}
+	ch->is_write_closed = true;
```
Totktonada, line 202: Here I'm in doubt: in the new behavior `close()` sets `is_write_closed` and doesn't set `is_closed`. It feels lopsided. I have no certain recommendations how to make it more clean, but would appreciate ideas and ready to discuss variants.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
Lord-KA: Done:                                
>  Title: display availible compat options in doc                                                                                                                                                                
  With introduction of tarantool.compat, there will be added a lot of                               
  options that change user-visible behavior. Such changes are described in                          
  detail on tarantool.wiki, but it could be benefitial to have a link to                            
  them in the documentation of the current behavior.

Totktonada: Hm. This task is not measurable: it has no certain end, because the options are added from time to time. It'll be not clear, when one can close the task. I would rephrase it against the option introduced here.

src/lib/core/fiber_channel.c
Lord-KA: Done.

src/lib/core/fiber_channel.h
```diff
@@ -207,6 +211,13 @@ fiber_channel_is_empty(struct fiber_channel *ch)
 	return ch->count == 0;
 }
 
+/**
+ * If fiber channel should work in a graceful close mode,
+ * i.e. become write-closed instead of closed until buffer
+ * is emitted by recipient. Is set by tarantool.compat.
```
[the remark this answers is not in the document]
Lord-KA: Done

src/lib/core/fiber_channel.c
Lord-KA: Renamed `is_closed` -> `is_destroyed` and `is_write_closed` -> `is_closed`.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
Lord-KA: Updated:
```
    @TarantoolBot document
    Title: new compat option fiber_channel_graceful_close
    
    New behavior is gracefully closing fiber channel by marking it
    read-only, instead of destroying. Full API description can be found
    on notion (private):
    https://www.notion.so/tarantool/Fiber-channel-graceful-close-53b2788ed1f144598c4c0e1229c2eb69
```

Totktonada: Thank you!

src/lua/compat.h
```diff
@@ -17,6 +17,10 @@ struct lua_State;
 int
 tarantool_lua_compat_init(struct lua_State *L);
 
+/* Toggler for fiber channel graceful close change. */
+int
+fiber_channel_graceful_close_toggle(struct lua_State *L);
```
Totktonada, line 22: We don't need in the header: a static function in the .c file is enough.

I understand a wish to provide some declaration of provided functions/API description, but I doubt that we should achieve it this way. As for me, `struct luaL_reg` tables works good enough for this.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
```diff
@@ -0,0 +1,55 @@
...
+    ch:close()
+
+    t.assert(ch:is_closed())
+    t.assert_not(ch:put('three'))
+
+    t.assert_equals(ch:get(), 'one')
+    t.assert(ch:is_closed())
+
+    t.assert_equals(ch:get(), 'two')
+    t.assert(ch:is_closed())
```
Totktonada, line 45: We can do one more `:get()` to ensure that it'll not wait for writers.

Feel free to ignore.

src/lua/compat.h
Lord-KA: You are right, forgot to remove it like in other options. Fixed.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
Lord-KA: Good idea, done.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,103 @@ fiber_channel_get()
...
+	fiber_channel_graceful_close_false(channel);
+	channel = fiber_channel_new(10);
+	fiber_channel_set_graceful_close(true);
+	fiber_channel_graceful_close_true(channel);
+	channel = fiber_channel_new(10);
+	fiber_channel_set_graceful_close(false);
+	fiber_channel_graceful_close_false(channel);
+
+	footer();
+	status = check_plan();
```
Totktonada, line 168: `status = status && check_plan();` would be more correct.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,103 @@ fiber_channel_get()
...
+	plan(7 * 2 + 8);
+
+	struct fiber_channel *channel = fiber_channel_new(10);
+	fiber_channel_graceful_close_false(channel);
+	channel = fiber_channel_new(10);
+	fiber_channel_set_graceful_close(true);
+	fiber_channel_graceful_close_true(channel);
+	channel = fiber_channel_new(10);
+	fiber_channel_set_graceful_close(false);
+	fiber_channel_graceful_close_false(channel);
```
Totktonada, line 165: Stilystic proposals:

* `fiber_channel_test_graceful_close()` -> `fiber_channel_test_close()` (because we test here not only graceful close).
* `fiber_channel_graceful_close_false()` -> `fiber_channel_forceful_close()` (or, better, just `test_forceful_close()`).
* `fiber_channel_graceful_close_true()` -> `fiber_channel_graceful_close()` (or, better, just `test_graceful_close()`).
* Mark functions as `static`.
* Create channels inside the `test_*ful_close()` functions.
* Split the cases by empty lines:

  ```c
  test_forceful_close();

  fiber_channel_set_graceful_close(true);
  test_graceful_close();

  fiber_channel_set_graceful_close(false);
  test_forceful_close();
  ```

Isn't it look easier?

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,103 @@ fiber_channel_get()
...
+	ptr = NULL;
+	fiber_channel_get(channel, &ptr);
+
+	ok(ptr == NULL, "not fiber_channel_get(1)");
+
+	fiber_channel_delete(channel);
+}
+
+void
+fiber_channel_graceful_close_true(struct fiber_channel *channel)
```
Totktonada, line 117: AFAIS, the test cases differ only in one condition. So it looks worthful to parametrize it instead of copying.

test/unit/fiber_channel.cc
```diff
@@ -18,9 +18,9 @@ fiber_channel_basic()
 
 	ok(fiber_channel_count(channel) == 0, "fiber_channel_count()");
 
-	ok(fiber_channel_is_full(channel) == false, "fiber_channel_is_full()");
+	ok(!fiber_channel_is_full(channel), "fiber_channel_is_full()");
 
-	ok(fiber_channel_is_empty(channel) == true, "fiber_channel_is_empty()");
+	ok(fiber_channel_is_empty(channel), "fiber_channel_is_empty()");
```
Totktonada, line 23: > Don't do a change if it is not related to the patch.

https://github.com/tarantool/tarantool/wiki/Code-review-procedure#general-coding-points-to-check

src/lib/core/fiber_channel.h
```diff
@@ -137,7 +137,15 @@ struct fiber_channel {
 	struct rlist waiters;
```
Totktonada, line 137: Please, adjust the comment above. Quoted relevant parts:

> A channel, once created is "open". I.e. anyone can read or
> write to/from a channel. A channel can be closed at any time,
> in which case, all messages currently buffered in a channel
> are destroyed, waiting readers or writers awoken with an error.

> Sending a message to a closed channel, as well as reading
> a message from such channel, always fails.

src/lib/core/fiber_channel.h
```diff
@@ -137,7 +137,15 @@ struct fiber_channel {
...
+	/**
+	 * True if the channel forbids both reading and writing and
+	 * the buffer is no longer accessible.
+	 */
+	bool is_destroyed;
+	/**
+	 * True if the channel is closed for writing and
+	 * is waiting to be destroyed or has been destroyed.
+	 */
 	bool is_closed;
```
Totktonada, line 149: I would reorder the flags: the channel is closed first, then destroyed. The order opposite to one present now would be more natural.

src/lib/core/fiber_channel.h
```diff
@@ -207,6 +215,12 @@ fiber_channel_is_empty(struct fiber_channel *ch)
 	return ch->count == 0;
 }
 
+/**
+ * Setter for fiber_channel_graceful_close.
+ */
+void
+fiber_channel_set_graceful_close(bool is_graceful);
```
Totktonada, line 222: I thought on it once again and it seems I found how to naturally represent this duality.

```c
/** The fiber channel has two ways to end its life. */
enum fiber_channel_close_mode {
	/**
	 * The forceful close mode means that fiber_channel_close()
	 * discards all stored messages, wake ups all readers and
	 * writers and forbids to read or write messages. The
	 * channel is effectively died after closing.
	 *
	 * fiber_channel_close() works exactly as
	 * fiber_channel_destroy() when this mode is in effect.
	 *
	 * This mode represents original behavior of the fiber
	 * channel since its born and hence it is the default.
	 */
	FIBER_CHANNEL_CLOSE_FORCEFUL,
	/**
	 * The graceful close mode means that fiber_channel_close()
	 * only marks the channel as read-only: no new messages can
	 * be added, but all buffered messages can be received.
	 * All waiting writers receive an error.
	 *
	 * A closed channel is automatically destroyed when all
	 * buffered messages are read.
	 *
	 * This mode is considered as more safe and recommended for
	 * users. The main idea behind it is to prevent accidental
	 * data loss or, in other words, fit users' expectation that
	 * all accepted messages will be delivered. This reflects
	 * how channels work in other programming languages, for
	 * example, in Go.
	 */
	FIBER_CHANNEL_CLOSE_GRACEFUL,
};

/**
 * Choose how the channels end its life.
 *
 * Affects all channels.
 *
 * The behavior is unspecified for already created channels.
 * Choose the mode at an early stage of application's
 * initialization.
 *
 * @sa enum fiber_channel_close_mode
 * @sa fiber_channel_close
 */
void
fiber_channel_set_close_mode(enum fiber_channel_close_mode);
```

All comments are to illustrate the idea. Feel free to modify them to make better.

(I would highlight the unspecified behavior note. It should be declared explicitly.)

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,103 @@ fiber_channel_get()
...
+	status = check_plan();
+}
+
 int
 main_f(va_list ap)
 {
 	(void) ap;
 	fiber_channel_basic();
 	fiber_channel_get();
+	fiber_channel_test_graceful_close();
```
Totktonada, line 177: There is no any test that would verify waiting writter's behavior at channel closing.

The same is applicable for the Lua test.

src/lib/core/fiber_channel.c
```diff
@@ -41,6 +41,19 @@ enum fiber_channel_wait_status {
 	FIBER_CHANNEL_WAIT_CLOSED /* Wait is aborted, the channel is closed. */
 };
 
+/**
```
Totktonada, line 44: Let's keep some meaningful order of declarations and definitions. Usually it is structures-globals-functions. I see no reason to bend the rules.

src/lib/core/fiber_channel.c
```diff
@@ -325,10 +357,10 @@ fiber_channel_put_msg_timeout(struct fiber_channel *ch,
...
 			/*
-			 * Closed channels, are, well, closed,
+			 * Destroyed channels, are, well, destroyed,
 			 * even if there is space in the buffer.
 			 */
-			if (ch->is_closed) {
+			if (ch->is_destroyed) {
 				diag_set(ChannelIsClosed);
 				return -1;
 			}
```
Totktonada, line 366: You added an exit if `ch->is_closed` and there is no yield in between the check and this hunk. So the channel can't be destroyed at this point. Let's replace with an assertion.

src/lib/core/fiber_channel.c
```diff
@@ -191,7 +205,24 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
...
+		fiber_channel_destroy(ch);
+		return;
+	}
+	ch->is_closed = true;
+	/* Wake up all channel_put(), they will automatically cancel. */
+	struct fiber *f;
+	while (!rlist_empty(&ch->waiters)) {
+		f = rlist_first_entry(&ch->waiters, struct fiber, state);
+		fiber_channel_waiter_wakeup(f, FIBER_CHANNEL_WAIT_CLOSED);
+	}
```
Totktonada, line 219: AFAIU a channel with an empty buffer has either readers or writers, because a writer removes a reader from the list and vice versa (see comments in `fiber_channel_waiter_wakeup()`).

While thinking on it I found that we have no test case that gracefully close a channel with a waiting reader. It'll have no buffered messages, but anyway. Let's add a unit test case and a Lua test case.

Totktonada, line 219: AFAIU if the buffer is not empty, the destructors for bufferred messages will not be called. It seems incorrect.

Let's add a test case that verifies that a destructor is called for a bufferred message at graceful close. It is okay to add only a unit test case here. But if you want to add a Lua test case, you can define a destructor this way: `ffi.gc(ffi.new('void *'), function(_self) <...> end)`. Don't forget to call `collectgarbage()` before check that the destructor is invoked.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
```diff
@@ -0,0 +1,6 @@
+## feature/core/fiber
```
Totktonada, line 1: It would be nice to verify the new mode before we'll proceed with the patch. I collected a list of our modules that use channels.

----

Ones that I would verify in the new mode:

* tarantool/cartridge
* tarantool/moonwalker
* tarantool/kafka
* tarantool/mysql
* tarantool/pg
* tarantool/tracing

Where a channel is used in a test (but anyway interesting to check it):

* tarantool/expirationd
* tarantool/queue
* tarantool/sharded-queue
* tarantool/smtp
* tarantool/synchronized (beware, it is [flawed](https://github.com/tarantool/synchronized/issues/1))

Deprecated (but maybe worth to check):

* tarantool/pregel (no updates since 2016, but it is known to use `ch:close()`)

Deprecated (I would skip them):

* tarantool/zookeeper (last update in 2019, explicitly marked as community supported)
* tarantool/shard (deprecated several years ago)
* tarantool/sentry (archived)

----

My source data is two years old and I'm lazy to re-collect the repositories. I hope it gives a good test plan anyway.

You can ask me where to get help with each particular module.

src/lib/core/fiber_channel.c
```diff
@@ -173,7 +187,7 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
 	 * caller, since ev_now() does not get updated without
 	 * a yield.
 	 */
-	if (ch->is_closed) {
+	if (ch->is_destroyed) {
```
Totktonada, line 190: Let's update the comment above as well:

> Preconditions of waiting are:
> - the channel is not closed,

src/lib/core/fiber_channel.c
Totktonada: Argh, that's the point of graceful close. I need a fresh head to review.

Regarding the test case: it looks interesting anyway. However we should verify that the destructors are NOT called.

src/lib/core/fiber_channel.c
Totktonada: It looks better, thanks!

Totktonada:
Let's consider this review round as finished from my side: I need a fresh head to review it better and I have no such head today.

src/lib/core/fiber_channel.h
Lord-KA: Done.

src/lib/core/fiber_channel.c
Lord-KA: Done.

src/lib/core/fiber_channel.c
Lord-KA: Done.

src/lib/core/fiber_channel.c
Lord-KA: Fixed.

test/unit/fiber_channel.cc
Lord-KA: It does. Fixed.

src/lib/core/fiber_channel.h
Lord-KA: Done.

test/unit/fiber_channel.cc
Lord-KA: Done.

src/lib/core/fiber_channel.h
Lord-KA: I'm not a big fan of `enum` with just two values. Maybe keep `bool` but add more comments?

test/unit/fiber_channel.cc
Lord-KA: I think, it could be more confusing, if the tests are combined.

Totktonada: I don't insist. Will leave the discussion unresolved for Igor's glance.

src/lib/core/fiber_channel.h
Totktonada: (And in `fiber_channel_create()`. Feel free to ignore.)

src/lib/core/fiber_channel.c
Totktonada: The static variable is still in between of types declarations.

src/lib/core/fiber_channel.h
igormunkin: Also fixed.

src/lib/core/fiber_channel.c
```diff
@@ -191,7 +206,54 @@ fiber_channel_check_wait(struct fiber_channel *ch, ev_tstamp start_time,
...
+	 * channel is effectively died after closing.
+	 *
+	 * fiber_channel_close() works exactly as
+	 * fiber_channel_destroy() when this mode is in effect.
+	 *
+	 * This mode represents original behavior of the fiber
+	 * channel since its born and hence it is the default.
+	 */
+	if (fiber_channel_close_mode == FIBER_CHANNEL_CLOSE_FORCEFUL)
+		goto destroy;
```
Totktonada, line 222: Nit: It is strange to see a documentation comment inside a function body.

Also I would leave here only information that helps to understand the algorithm. Mode X do A, B, C. Mode Y do D, E, F.

Totktonada: I would also make assumption of the code explicit:

* A channel has either readers or writters (if any).
* If the buffer has any message, the channel can't have a reader.

src/lib/core/fiber_channel.h
Totktonada: Igor applied this proposal. Closing the discussion.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
```diff
@@ -0,0 +1,113 @@
...
+    local writer = fiber.create(function()
+        t.assert_not(ch:is_closed())
+        t.assert_equals(ch:get(), nil)
+        t.assert(ch:is_closed())
+        t.assert_equals(ch:get(0), nil)
+    end)
+    ch:close()
+    t.assert(ch:is_closed())
+    fiber.yield()
+end
```
Totktonada, line 71: There are several problems in this test case.

1. Typo: a reader fiber is named `writer`.
2. Assertions inside the fiber don't affect the test result. Let's check it:

   ```lua

   local t = require('luatest')
   local g = t.group()

   g.test_x = function()
       fiber.create(function()
           t.assert_equals(1, 2)
       end)
   end
   ```
3. `fiber.yield()` in the main fiber doesn't guarantee that the reader fiber will be executed. Let's wait for this event in a loop.
4. There are many assertions and it is not clear on the first glance what is the point of the test. Let's make it explicit somehow: either adjust the code or add a comment.
5. There is no difference in old/new behavior for this situation. It would be nice to highlight this fact to don't let anyone compare those two test cases line-by-line.

src/lua/compat.lua
```diff
@@ -58,6 +70,12 @@ local options = {
         brief = YAML_PRETTY_MULTILINE_BRIEF,
         action = internal.yaml_pretty_multiline_toggle,
     },
+    fiber_channel_graceful_close = {
+        default = 'old',
+        obsolete = nil,
+        brief = FIBER_CHANNEL_GRACEFUL_CLOSE_BRIEF,
+        action = internal.fiber_channel_graceful_close_toggle,
+    },
```
Totktonada, line 78:
```lua
compat.fiber_channel_graceful_close = 'old'
compat.fiber_channel_graceful_close = 'new'
```

It sounds a bit artificial. How about another name?

```lua
compat.fiber_channel_close_mode = 'old'
compat.fiber_channel_close_mode = 'new'
```

test/unit/fiber_channel.cc
Totktonada: Igor applied this proposal.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+	   graceful ? "not is_destroyed" : "is_destroyed");
+
+	ok(fiber_channel_put_timeout(channel, &msg_3, 0) != 0,
+	   "not fiber_channel_put(msg_3)");
+
+	ptr = NULL;
+	fiber_channel_get(channel, &ptr);
+
+	ok(ptr == (graceful ? &msg_2 : NULL),
+	   graceful ? "fiber_channel_get(1)" : "not fiber_channel_get(1)");
```
Totktonada, line 117: Nit: 1 -> 2.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+}
+
+static int
+reader_f(va_list ap)
+{
+	struct fiber_channel *channel = (struct fiber_channel *)fiber()->f_arg;
+	void *msg;
+	ok(!channel->is_closed, "reader_f tries to read from the open channel");
+	/*
+	 * Try to obtain the message into the zero-length channel.
```
Totktonada, line 130: Typo: obtain into. I guess, 'obtain from' was meant.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+	return 0;
+}
+
+static void
+fiber_channel_close_reader(enum fiber_channel_close_mode mode)
+{
+	fiber_channel_set_close_mode(mode);
+
+	bool graceful = mode == FIBER_CHANNEL_CLOSE_GRACEFUL;
+	struct fiber_channel *channel = fiber_channel_new(1);
```
Totktonada, line 154: This is not a zero-length channel.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+	   graceful ? "fiber_channel_get(1)" : "not fiber_channel_get(1)");
+	ok(channel->is_destroyed, "is_destroyed");
+
+	fiber_channel_delete(channel);
+}
+
+static int
+reader_f(va_list ap)
+{
+	struct fiber_channel *channel = (struct fiber_channel *)fiber()->f_arg;
```
Totktonada, line 126: Nit: An explicit casting `void *` to a specific pointer type is not necessary. Same as with `struct foo *x = malloc(sizeof(struct foo));`

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+	fail_if(reader == NULL);
+	reader->f_arg = channel;
+	fiber_wakeup(reader);
+	/* Yield to start tests in <reader> payload. */
+	fiber_sleep(0);
+
+	fiber_channel_close(channel);
+	ok(channel->is_closed, "is_closed");
+	/* Yield to finish tests in <reader> payload. */
+	fiber_sleep(0);
```
Totktonada, line 165: Can we at least assert that the `reader_f()` function reaches its end? We can either set a global variable or make the fiber joinable and join it.

Now we lean on particular fiber scheduling algorithm and it feels fragile.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+static void
+fiber_channel_close_reader(enum fiber_channel_close_mode mode)
+{
+	fiber_channel_set_close_mode(mode);
+
+	bool graceful = mode == FIBER_CHANNEL_CLOSE_GRACEFUL;
+	struct fiber_channel *channel = fiber_channel_new(1);
+	struct fiber *reader = fiber_new("reader", reader_f);
+	fail_if(reader == NULL);
+	reader->f_arg = channel;
```
Totktonada, line 157: BTW, there is `fiber_{get,set}_ctx()`.

test/unit/fiber_channel.cc
```diff
@@ -78,12 +78,167 @@ fiber_channel_get()
...
+
+	return 0;
+}
+
+static void
+fiber_channel_close_reader(enum fiber_channel_close_mode mode)
+{
+	fiber_channel_set_close_mode(mode);
+
+	bool graceful = mode == FIBER_CHANNEL_CLOSE_GRACEFUL;
```
Totktonada, line 153: Unused.

test/unit/fiber_channel.cc
Totktonada: Igor added it.

src/lib/core/fiber_channel.c
Totktonada: Igor added such test cases.

Totktonada: We discussed it with Igor and he thinks that the test case is more about general guarantees of the fiber channel and it has no strict relation to this patchset. I'll close the discussion so.

test/unit/fiber_channel.cc
igormunkin: FIxed, thanks!

test/unit/fiber_channel.cc
igormunkin: Much better, fixed, thanks!

test/unit/fiber_channel.cc
igormunkin: Removed this in both cases.

test/unit/fiber_channel.cc
igormunkin: My bad, fixed, thanks!

test/unit/fiber_channel.cc
igormunkin: Meh, copy-pasting artefacts. Fixed, thanks!

test/unit/fiber_channel.cc
igormunkin: Fixed, thanks!

test/unit/fiber_channel.cc
igormunkin: Returned back, since implicit conversion from `void *` is not allowed in C++.

Totktonada: Sorry. I missed that it is a `.cc` file.

test/unit/fiber_channel.cc
igormunkin: Sounds reasonable. Replaced this `fiber_sleep` with `fiber_join`.

src/lua/compat.lua
igormunkin: Also looks better to me. Fixed.

src/lib/core/fiber_channel.c
igormunkin: Fixed the aforementioned comment. Also added the new comment for the section below.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
igormunkin: Fixed everything mentioned above.

changelogs/unreleased/gh-7746-fiber-channel-graceful-close.md
igormunkin: I set the default value of `compat.fiber_channel_close_mode` to `new` and run integration testing that covers cartridge, mysql, expirationd, queue. Anyway, I suggest to verify all external modules when we decide to change the default.

test/app-luatest/gh_7746_fiber_channel_close_test.lua
```diff
@@ -0,0 +1,133 @@
...
+        t.assert(ch:is_closed(), 'reader hangs until channel is closed')
+        t.assert_is(ch:get(0), nil,
+                    'reader fails to read from the closed channel')
+    end)
+    reader:set_joinable(true)
+    -- Yield to start tests in <reader> payload.
+    fiber.yield()
+    ch:close()
+    t.assert(ch:is_closed())
+    -- Wait until tests in <writer> payload are finished.
```
Totktonada, line 84: Nit: writer -> reader.

Totktonada (approved):

igormunkin (approved):
I also add my formal LGTM, since initially I was the second reviewer.

PR #7827, opened 2022-10-19
box: unify errors about mismatch of password and login during auth

drewdzzz:
If we raise different errors in case of entering an invalid password and entering the login of a non-existent user during authorization, it will open the door for an unauthorized person to enumerate users. So let's unify raised errors in the cases described above.

Closes #tarantool/security#16


locker (approved):

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7833, opened 2022-10-21
fiber: initialize thread-local cord on demand

### LINKED ISSUE #7814: Allow to use diagnostic area in user threads
To use the diagnostic area (for error reporting) in a thread, one has to initialize it with [`cord_create`](https://github.com/tarantool/tarantool/blob/fd780129c4002a558cc547d0dd6826bf1afcffbe/src/lib/core/fiber.h#L791) and destroy it with [`cord_exit`](https://github.com/tarantool/tarantool/blob/fd780129c4002a558cc547d0dd6826bf1afcffbe/src/lib/core/fiber.h#L798) and [`cord_destroy`](https://github.com/tarantool/tarantool/blob/fd780129c4002a558cc547d0dd6826bf1afcffbe/src/lib/core/fiber.h#L801) at thread exit. This is done automatically by [`cord_start`](https://github.com/tarantool/tarantool/blob/fd780129c4002a558cc547d0dd6826bf1afcffbe/src/lib/core/fiber.h#L812), but if a thread is started directly with `pthread_create`, this has to be done manually, which is inconvenient. Currently, there's just one such place in our code - it's coio helper threads. To init/deinit cord there, we use the callbacks provided by eio:

https://github.com/tarantool/tarantool/blob/fd780129c4002a558cc547d0dd6826bf1afcffbe/src/lib/core/coio_task.c#L110-L136

The problem is that if a thread is started from the user code, there's no such callbacks available. This makes it impossible to use the diagnostic area in user threads, which is a pre-requisite for the read view C API.

To overcome this limitation, let's make the [`cord`](https://github.com/tarantool/tarantool/blob/66e388e86bd3810074bd79f1d8fa72518198d6db/src/lib/core/fiber.h#L786) function initialize the cord automatically on the first use and install a callback to deinitialize it at thread exit. Deinitialization can be done with C++ RAII, like in [`cxx_abi_demangle`](https://github.com/tarantool/tarantool/blob/302d91cf8e13e1a167424e4f116612113a012d12/src/lib/core/cxx_abi.cc#L47).

locker:
We're planning to introduce a basic C API for user read views (EE-only). Like all other box C API functions, the new API functions will use the existing box error C API for reporting errors. The problem is that a read view created using C API should be usable from user threads (started with the pthread lib) while the box error C API doesn't work in user threads, because those threads don't have the cord pointer initialized (a diagnostic area is stored in a cord object).

To address this issue, let's create a new cord object automatically on first use of `cord()` if it wasn't created explicitly. Automatically created object is destroyed at thread exit (to achieve that, we use the C++ RAII concept).

Closes #7814

### DIFF
--- a/src/lib/core/cord_on_demand.cc
+++ b/src/lib/core/cord_on_demand.cc
@@ -1,0 +1,53 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2022, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#include "cord_on_demand.h"
+
+#include <stdlib.h>
+
+#include "fiber.h"
+#include "trivia/util.h"
+
+/**
+ * RAII wrapper around a cord object.
+ *
+ * A thread-local cord object is created on the first call to the get() method
+ * and destroyed when the thread exits.
+ */
+class CordOnDemand final {
+public:
+	static cord *get() noexcept
+	{
+		thread_local CordOnDemand singleton;
+		return singleton.cord_ptr;
+	}
+
+private:
+	struct cord *cord_ptr;
+
+	CordOnDemand() noexcept
+	{
+		cord_ptr = static_cast<struct cord *>(
+				xcalloc(1, sizeof(*cord_ptr)));
+		cord_create(cord_ptr, "unknown");
+	}
+
+	~CordOnDemand()
+	{
+		cord_exit(cord_ptr);
+		cord_destroy(cord_ptr);
+		free(cord_ptr);
+	}
+
+	CordOnDemand(CordOnDemand &other) = delete;
+	CordOnDemand &operator=(CordOnDemand &other) = delete;
+};
+
+struct cord *
+cord_on_demand(void)
+{
+	return CordOnDemand::get();
+}
--- a/changelogs/unreleased/gh-7814-box-error-in-thread.md
+++ b/changelogs/unreleased/gh-7814-box-error-in-thread.md
@@ -1,0 +1,4 @@
+## feature/core
+
+* The box error C API (`box_error_set()`, `box_error_last()`, etc) can now be
+  used in threads started by user modules with the pthread library (gh-7814).
--- a/src/lib/core/CMakeLists.txt
+++ b/src/lib/core/CMakeLists.txt
@@ -43,6 +43,7 @@ set(core_sources
     tt_sigaction.c
     tt_strerror.c
     mp_util.c
+    cord_on_demand.cc
 )
 
 if (ENABLE_BACKTRACE)
--- a/src/lib/core/cord_on_demand.h
+++ b/src/lib/core/cord_on_demand.h
@@ -1,0 +1,25 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2022, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+struct cord;
+
+/**
+ * On the first call, creates and returns a new thread-local cord object.
+ * On all subsequent calls in the same thread, returns the object created
+ * earlier. The cord object is destroyed at thread exit.
+ */
+struct cord *
+cord_on_demand(void);
+
+#ifdef __cplusplus
+} /* extern "C" */
+#endif /* __cplusplus */
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -1481,7 +1481,7 @@ box_region_truncate(size_t size)
 void
 cord_create(struct cord *cord, const char *name)
 {
-	cord() = cord;
+	cord_ptr = cord;
 	slab_cache_set_thread(&cord()->slabc);
 
 	cord->id = pthread_self();
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -36,6 +36,7 @@
 #include <stdint.h>
 #include "tt_pthread.h"
 #include <tarantool_ev.h>
+#include "cord_on_demand.h"
 #include "diag.h"
 #include "trivia/util.h"
 #include "small/mempool.h"
@@ -783,7 +784,19 @@ struct cord {
 
 extern __thread struct cord *cord_ptr;
 
-#define cord() cord_ptr
+/**
+ * Returns a thread-local cord object.
+ *
+ * If the cord object wasn't initialized at thread start (cord_create()
+ * wasn't called), a cord object is created automatically and destroyed
+ * at thread exit.
+ */
+#define cord() ({							\
+	if (unlikely(cord_ptr == NULL))					\
+		cord_ptr = cord_on_demand();				\
+	cord_ptr;							\
+})
+
 #define fiber() cord()->fiber
 #define loop() (cord()->loop)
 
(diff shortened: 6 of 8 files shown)


Gumix (approved):

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * RAII wrapper around a cord object.
+ *
+ * A thread-local cord object is created on the first call to the get() method
+ * and destroyed when the thread exits.
+ */
+class CordOnDemand final {
+public:
+	static cord *get()
```
CuriousGeorgiy, line 22: Nit: this should be `noexcept`.

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+	{
+		static thread_local CordOnDemand singleton;
```
CuriousGeorgiy, line 24: Nit: `static` is redundant, `thread_local` already implies it, see https://timsong-cpp.github.io/cppwp/n3337/dcl.stc#4.

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+		return singleton.cord_ptr;
+	}
+
+private:
+	struct cord *cord_ptr;
```
CuriousGeorgiy, line 29: Nit: I think this is overengineering — this doesn't provide any encapsulation, since one can still get this pointer through `get`.

Better off making this member public and changing the `get` method to:
```C++
static CordOnDemand &instance() noexcept {
	thread_local CordOnDemand singleton;
	return singleton;
}
```

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+#include "fiber.h"
+#include "trivia/util.h"
+
+/**
+ * RAII wrapper around a cord object.
+ *
+ * A thread-local cord object is created on the first call to the get() method
+ * and destroyed when the thread exits.
+ */
+class CordOnDemand final {
```
CuriousGeorgiy, line 20: Nit: this class doesn't really provide any encapsulation — in this case I usually use `struct` declaration.

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+
+	~CordOnDemand()
+	{
+		cord_exit(cord_ptr);
+		cord_destroy(cord_ptr);
+		free(cord_ptr);
+	}
+
+	CordOnDemand(CordOnDemand &other) = delete;
+	CordOnDemand &operator=(CordOnDemand &other) = delete;
```
CuriousGeorgiy, line 46: Nit: I would move these to the public section.

src/lib/core/cord_on_demand.cc
locker: Added.

src/lib/core/cord_on_demand.cc
locker: Removed.

src/lib/core/cord_on_demand.cc
locker: > I think this is overengineering — this doesn't provide any encapsulation, since one can still get this pointer through `get`.

But one can't change the value of the `cord_ptr` pointer with `get` while with `instance` one can. Besides `CordOnDemand()::get()` looks better than `CordOnDemand()::instance().cord_ptr` IMO.

src/lib/core/cord_on_demand.cc
locker: I'm following the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes) here:

> Use a struct only for passive objects that carry data; everything else is a class.

src/lib/core/cord_on_demand.cc
locker: Why?

src/lib/core/cord_on_demand.cc
CuriousGeorgiy: Okay, sounds reasonable.

src/lib/core/cord_on_demand.cc
CuriousGeorgiy: If we would decide to discard the `get` method, this object would actually only carry data (static methods don’t account here, AFAIC).

src/lib/core/cord_on_demand.cc
CuriousGeorgiy: Let’s also make the `get` method const-qualified then.

CuriousGeorgiy: > Besides CordOnDemand()::get() looks better than CordOnDemand()::instance().cord_ptr.

Perhaps a matter of taste, but `_ptr` suffix looks ugly to me.

locker: > Let’s also make the `get` method const-qualified then.

[A static member function cannot have a cv-qualifier](https://stackoverflow.com/questions/19812786/why-cant-a-static-member-function-have-a-cv-qualifier).



> Perhaps a matter of taste, but `_ptr` suffix looks ugly to me.

Without the suffix, it wouldn't compile:

```
/home/vlad/src/tarantool/tarantool/src/lib/core/cord_on_demand.cc:29:15: error: declaration of ‘cord* CordOnDemand::cord’ changes meaning of ‘cord’ [-fpermissive]
   29 |  struct cord *cord;
      |               ^~~~
In file included from /home/vlad/src/tarantool/tarantool/src/lib/core/cord_on_demand.cc:11:
/home/vlad/src/tarantool/tarantool/src/lib/core/fiber.h:707:8: note: ‘cord’ declared here as ‘struct cord’
  707 | struct cord {
      |        ^~~~
```

We could use `_` suffix, as they usually add to private members, but we don't follow this rule in other parts of our code. Besides the `cord_ptr` name is consistent with [`cord_ptr`](https://github.com/tarantool/tarantool/blob/256da01046da8f4fbe81d629cf54ae32dc04a1fb/src/lib/core/fiber.h#L784) defined in the fiber module.

src/lib/core/cord_on_demand.cc
alyapunov: We have private constructor/destructor and that is part of interface in general meaning - you will not be able to create and destroy objects of this class. I think we should naturally have all the rule of three methods nearby.

src/lib/core/cord_on_demand.cc
```diff
@@ -0,0 +1,53 @@
...
+	struct cord *cord_ptr;
+
+	CordOnDemand() noexcept
+	{
+		cord_ptr = static_cast<struct cord *>(
+				xmalloc(sizeof(*cord_ptr)));
+		cord_create(cord_ptr, "unknown");
+	}
+
+	~CordOnDemand()
```
alyapunov, line 38: Nit: I would also mark it as noexcept.

alyapunov (approved):

src/lib/core/cord_on_demand.cc
alyapunov: Nevermind, destructors are noexcept by default.

locker:
Turns out, the cord object must be zeroed-out before calling `cord_create`, because `cord_create` doesn't set `cord->loop`. Caught by the release-asan workflow.

```diff
diff --git a/src/lib/core/cord_on_demand.cc b/src/lib/core/cord_on_demand.cc
index 7d9425c1e110..b0eb247d2430 100644
--- a/src/lib/core/cord_on_demand.cc
+++ b/src/lib/core/cord_on_demand.cc
@@ -31,7 +31,7 @@ private:
 	CordOnDemand() noexcept
 	{
 		cord_ptr = static_cast<struct cord *>(
-				xmalloc(sizeof(*cord_ptr)));
+				xcalloc(1, sizeof(*cord_ptr)));
 		cord_create(cord_ptr, "unknown");
 	}
 
```

</details>

PR #7840, opened 2022-10-24
build: fix `-Wa,--debug-prefix-map` compiler flag check

CuriousGeorgiy:
**build: revert usage of relative paths in debugging information**
The `-Wa,--debug-prefix-map` compiler flag breaks GNU-based LTO, and also
we cannot reliably test this feature.

Follow-up https://github.com/tarantool/tarantool/commit/256da01046da8f4fbe81d629cf54ae32dc04a1fb

**build: store compiler ID and version in COMPILER_INFO**
Currently, `tarantool.build.compiler` stores the `CMAKE_C_COMPILER` and
`CMAKE_CXX_COMPILER` strings from CMake, which are basically paths to the
compiler used for build, and are usually set to `/usr/bin/cc` and
`/usr/bin/c++` accordingly, which does not yield any useful information:
instead, set it to `${CMAKE_C_COMPILER_ID}-${CMAKE_C_COMPILER_VERSION}`,
assuming the same compiler is used for building C and C++.

Closes https://github.com/tarantool/tarantool/issues/7888

**test: add test for https://github.com/tarantool/tarantool/issues/7808**
Since this feature is compiler-dependent, we need to skip the test if the
server was build with a compiler that does not support the
`-fmacro-prefix-map` flag.

Follow-up https://github.com/tarantool/tarantool/issues/7808


cmake/prefix.cmake
```diff
@@ -20,9 +20,7 @@ if (NOT DEV_BUILD AND
     ((CMAKE_C_COMPILER_ID MATCHES "Clang" AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "3.8" OR CMAKE_C_COMPILER_VERSION VERSION_EQUAL "3.8")) OR
     (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "4.4.7" OR CMAKE_C_COMPILER_VERSION VERSION_EQUAL "4.4.7"))))
     # Clang integrated assembler does not support this option.
-    set(CMAKE_REQUIRED_FLAGS "-Wa,--debug-prefix-map")
-    check_c_source_compiles("int main(void) {}" HAVE_AS_DEBUG_PREFIX_MAP)
-    unset(CMAKE_REQUIRED_FLAGS)
+    check_c_compiler_flag(-Wa,--debug-prefix-map== HAVE_AS_DEBUG_PREFIX_MAP)
     if (HAVE_WA_DEBUG_PREFIX_MAP)
```
locker, line 24: Why do you check `HAVE_WA_DEBUG_PREFIX_MAP` while the macro above sets `HAVE_AS_DEBUG_PREFIX_MAP`?

cmake/prefix.cmake
```diff
@@ -20,9 +20,7 @@ if (NOT DEV_BUILD AND
     ((CMAKE_C_COMPILER_ID MATCHES "Clang" AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "3.8" OR CMAKE_C_COMPILER_VERSION VERSION_EQUAL "3.8")) OR
     (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "4.4.7" OR CMAKE_C_COMPILER_VERSION VERSION_EQUAL "4.4.7"))))
     # Clang integrated assembler does not support this option.
-    set(CMAKE_REQUIRED_FLAGS "-Wa,--debug-prefix-map")
-    check_c_source_compiles("int main(void) {}" HAVE_AS_DEBUG_PREFIX_MAP)
-    unset(CMAKE_REQUIRED_FLAGS)
+    check_c_compiler_flag(-Wa,--debug-prefix-map== HAVE_AS_DEBUG_PREFIX_MAP)
```
locker, line 23: Looks like this check fails on both clang and gcc:

```
$ clang -Wa --debug-prefix-map==
clang: error: unsupported option '--debug-prefix-map=='; did you mean '-fdebug-prefix-map=='?
clang: error: no input files
$ gcc -Wa --debug-prefix-map==
gcc: error: unrecognized command line option ‘-Wa’; did you mean ‘-W’?
gcc: fatal error: no input files
compilation terminated.
```

CuriousGeorgiy: Looks like you forgot the comma after `-Wa,`

locker: Looks like after a comma one can put anything:

```
$ gcc -Wa,--foobar==
gcc: fatal error: no input files
compilation terminated.
$ clang -Wa,--foobar==
clang: error: no input files
```

CuriousGeorgiy: Looks like this option is for assembler and is only recognized at link time.

locker: I see. Missed that. Looks okay then.

cmake/prefix.cmake
CuriousGeorgiy: Looks like a made a typo: fixed it.

locker:
Discussed f2f. Agreed that we need to add a test for this feature, at lest for Linux.

CuriousGeorgiy:
@Gumix @Totktonada, could you please look at the test and say if it's okay?

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,25 @@
+local bit = require('bit')
+local misc = require('test.luatest_helpers.misc')
+local netbox = require('net.box')
+local t = require('luatest')
+
+local g = t.group()
+
+g.test_rel_paths_in_diag = function(_)
+    local _, err = pcall(function() netbox.self:call('') end)
+    t.assert_not(err.trace[1].file:startswith('/'))
```
Totktonada, line 10: Will `box.error.new(box.error.UNKNOWN).trace[1].file` suit you here?

test/luatest_helpers/misc.lua
```diff
@@ -21,8 +26,15 @@ local function skip_if_enterprise()
...
+local function skip_if_macOS()
+    luatest.skip_if(is_macOS_build(), 'build target is not Linux')
+end
+
 return {
     is_debug_build = is_debug_build,
+    is_macOS_build = is_macOS_build,
     skip_if_not_debug = skip_if_not_debug,
     skip_if_enterprise = skip_if_enterprise,
+    skip_if_macOS = skip_if_macOS,
```
locker, line 39: This file is about to be removed. The helpers should be added to the luatest repository, see https://github.com/tarantool/luatest/pull/247.

test/luatest_helpers/misc.lua
```diff
@@ -21,8 +26,15 @@ local function skip_if_enterprise()
     luatest.skip_if(is_enterprise_package(), 'package is Enterprise')
 end
 
+-- Skips a running test if Tarantool build target is macOS.
+local function skip_if_macOS()
+    luatest.skip_if(is_macOS_build(), 'build target is not Linux')
```
locker, line 31: Nit: mixing CamelCase and underscores in one name looks ugly. Better name `is_mac_os_build` IMO.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,25 @@
...
+end
+
+g.test_rel_paths_in_debug_info = function(_)
+    misc.skip_if_macOS()
```
locker, line 14: There's also FreeBSD, where this test will probably not work, either. I think we should do `skip_if_not_linux` here.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,25 @@
...
+
+    local readelf = ("readelf --debug-dump=info %s"):format(arg[-1])
+    local grep_comp_dir = "grep DW_AT_comp_dir"
+    local awk = "awk '{print $9}'"
+    local grep_abs_path = "grep '^/'"
+    local rc = os.execute(table.concat({readelf, grep_comp_dir, awk, grep_abs_path}, '|'))
```
locker, line 20: Nit: I'd use `popen` here, because it'd allow for better error reporting (we could print which paths are not relative). Up to @Totktonada.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,25 @@
...
+    -- `os.execute` behaves like `system`.
+    rc = bit.rshift(bit.band(rc, 0xff00), 8)
+    -- `grep` returns 1 if no lines were selected.
+    t.assert_equals(rc, 1)
+end
```
Totktonada, line 25: Several suggestions:

1. If you need to run the test only on Linux, let's test for 'on linux' instead of 'not on mac os'. Tarantool works on FreeBSD and NetBSD, for example.
2. [jit.os](https://luajit.org/ext_jit.html#jit_os) provides simpler OS check and you'll not need a helper. If you'll update the helpers anyway, notify Yaroslav Lobankov: he is working on moving them to luatest itself (see https://github.com/tarantool/luatest/pull/247).
3. I would check for `readelf` existence (say, run `readelf --version`) and skip the test if is not exists.
4. I would use Lua's string functions instead of grep/awk for better portability.
5. You may be insterested in tarantool's built-in `popen` module or `io.popen`.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
CuriousGeorgiy: No, such a trace will show the test file, I need a diagnostic set from inside a source file.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
CuriousGeorgiy: `popen` does not have convenient helpers for reading.

CuriousGeorgiy:
@locker, @Totktonada the test rewritten with `io.popen` takes 40 seconds vs. the `os.execute` version taking 4 seconds :clown_face:

test/luatest_helpers/misc.lua
CuriousGeorgiy: Removed this helper in favor of testing for `readelf` availability.

test/luatest_helpers/misc.lua
CuriousGeorgiy: Removed this helper in favor of testing for `readline` availability.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
CuriousGeorgiy: Removed this helper in favor of testing for `readline` availability.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
CuriousGeorgiy: Points 1-3: applied your third suggestion and simply test for `readline` availability.

Points 4-5: rewrote the test to parse using `io.open` (`popen` does not have convenient functions for parsing) and `string.gmatch` — unfortunately, the test now takes 14 seconds compared to negligible in the `os.execute` version.

CuriousGeorgiy: Thanks a lot for your suggestions!

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,31 @@
...
+        local i = 1
+        for word in line:gmatch('%S+') do
+            if i ~= 2 and i ~= 9 then
+                goto continue
+            end
+            if i == 2 and word ~= "DW_AT_comp_dir" then
+                break
+            end
+            if i == 9 then
+                t.assert_str_matches(word, '^[^/]')
```
locker, line 25: This script passes with `DEV_BUILD=ON` - looks like this code never runs. Why not simplify path extraction? Something like this, maybe?

```lua
s:gsub('.*DW_AT_comp_dir.*:%s*(.*)$', '%1')
```

CuriousGeorgiy: > This script passes with DEV_BUILD=ON - looks like this code never runs. 

Screen Shot 2022-10-28 at 19 19 21

It doesn't pass on my machine — looks like you didn't reset your CMake cache.

CuriousGeorgiy: > Why not simplify path extraction? Something like this, maybe? 

Thanks for the suggestion, I couldn't think of an appropriate regular expression.

But initially I aimed for the best performance: with your approach the test now takes 112 seconds vs. 14 seconds of my approach.

locker: > It doesn't pass on my machine — looks like you didn't reset your CMake cache.

Now, *that you updated the script*, it doesn't pass on my machine, either. Anyhow, it *should pass* even if `DEV_BUILD` is enabled - the test should just be skipped in this case. I think you should be able to check the build options with `require('tarantool').build.options`.

> But initially I aimed for the best performance: with your approach the test now takes 112 seconds vs. 14 seconds with my approach.

Performance is a good point. If the test takes that long with regular expressions, then we need to do something about the check. Can we speed up the test like this?

```diff
diff --git a/test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua b/test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
index a03b870665ac..520f8c88b743 100644
--- a/test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
+++ b/test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
@@ -13,8 +13,8 @@ g.test_rel_paths_in_debug_info = function(_)
 
     local ph = io.popen('readelf --debug-dump=info ' .. arg[-1])
     for line in ph:lines() do
-        local _, _, path = line:find('.*DW_AT_comp_dir.*:%s*(.*)$')
-        if path ~= nil then
+        if line:find('DW_AT_comp_dir') then
+            local _, _, path = line:find('.*DW_AT_comp_dir.*:%s*(.*)$')
             t.assert_str_matches(path, '^[^/].*')
         end
     end
```

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,21 @@
...
+
+local g = t.group()
+
+g.test_rel_paths_in_diag = function(_)
+    local _, err = pcall(function() netbox.self:call('') end)
+    t.assert_not(err.trace[1].file:startswith('/'))
+end
+
+g.test_rel_paths_in_debug_info = function(_)
+    t.skip_if(bit.rshift(bit.band(os.execute('readelf --version'), 0xFF00), 8) == 127)
```
locker, line 12: This check looks hackish. `readelf` should be available on any Linux distribution that has a compiler installed. Let's change it to `require('jit').os ~= 'Linux'`?

CuriousGeorgiy: > This check looks hackish.

Sounds subjective. I use standard utilities to check availability of a program so that we can test a wider range of platforms. Moreover, relying on OS here is kind of strange.

locker: The check looks hackish, because of the bit manipulations with the return code. These manipulations are, in fact, OS-dependent according to the Lua manual:

[image](https://www.lua.org/manual/5.1/manual.html).

CuriousGeorgiy: Discussed f2f: decided to skip the test on `macOS` and with DEV_BUILD option enabled.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
CuriousGeorgiy: Applied your diff, thanks for the suggestion!

locker (approved):

locker:
The PR fails on several platforms. Please fix.

<details><summary>Examples</summary>

[OS X](https://github.com/tarantool/tarantool/actions/runs/3360523761/jobs/5569918791)

```
CMake Error at /Users/tntmac04.tarantool.i/actions-runner/_work/tarantool/tarantool/cmake/prefix.cmake:23 (check_c_compiler_flag):
  Unknown CMake command "check_c_compiler_flag".
```

[Ubuntu 16](https://github.com/tarantool/tarantool/actions/runs/3360523768/jobs/5569936957)

```
[060] app-luatest/gh_7808_rel_paths_in_diag_and_debu>                 [ fail ]
[060] Test failed! Output from reject file /tmp/t/rejects/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info.reject:
[060] TAP version 13
[060] 1..2
[060] # Started on Mon Oct 31 10:46:47 2022
[060] # Starting group: app-luatest.gh_7808_rel_paths_in_diag_and_debug_info
[060] not ok 1	app-luatest.gh_7808_rel_paths_in_diag_and_debug_info.test_rel_paths_in_diag
[060] #   ...uatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua:9: expected: false or nil, actual: true
[060] #   stack traceback:
[060] #   	...uatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua:9: in function 'app-luatest.gh_7808_rel_paths_in_diag_and_debug_info.test_rel_paths_in_diag'
[060] #   	...
[060] #   	[C]: in function 'xpcall'
[060] not ok 2	app-luatest.gh_7808_rel_paths_in_diag_and_debug_info.test_rel_paths_in_debug_info
[060] #   ...uatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua:20: Could not match pattern "^[^/].*" with string "/build/glibc-S7Ft5T/glibc-2.23/nptl"
[060] #   stack traceback:
[060] #   	...uatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua:20: in function 'app-luatest.gh_7808_rel_paths_in_diag_and_debug_info.test_rel_paths_in_debug_info'
[060] #   	...
[060] #   	[C]: in function 'xpcall'
[060] # Ran 2 tests in 5.126 seconds, 0 succeeded, 2 failed
[060] Test "app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua", conf: "None"
[060] 	failed, rerunning ...
```

</details>

changelogs/unreleased/gh-7888-store-compiler-id-and-version-in-compiler-info.md
```diff
@@ -0,0 +1,5 @@
+## feature/build
+
+* Set `COMPILER_INFO`, which is displayed in `tarantool.build.compiler` and
+  `tarantool --version`, to `CMAKE_C_COMPILER_ID-CMAKE_C_COMPILER_VERSION`
+  (gh-7888).
```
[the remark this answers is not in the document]
Totktonada: I don't know anything that leans on it, so it should be okay. It seems meaningful to provide this compiler information.

src/CMakeLists.txt
```diff
@@ -280,10 +280,8 @@ add_subdirectory(box)
 include_directories(${EXTRA_BOX_INCLUDE_DIRS})
 
 # Save CMAKE_XXX_FLAGS from this directory for config.h (used in --version)
-string(REPLACE  ${PREFIX_MAP_FLAGS} "" STRIPPED_CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
-set(TARANTOOL_C_FLAGS ${STRIPPED_CMAKE_C_FLAGS} PARENT_SCOPE)
-string(REGEX REPLACE ${PREFIX_MAP_FLAGS} "" STRIPPED_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
-set(TARANTOOL_CXX_FLAGS ${STRIPPED_CMAKE_CXX_FLAGS} PARENT_SCOPE)
```
Totktonada, line 286: `PREFIX_MAP_FLAGS` is still there (and contains `-fmacro-prefix-map=<source dir>=.`) and, to be honest, I don't understand two things:

* Why we strip it in the first place?
* Why now we don't do that?

This point looks minor, so feel free to ignore.

test/app-luatest/gh_7808_rel_paths_in_diag_and_debug_info_test.lua
```diff
@@ -0,0 +1,15 @@
+local netbox = require('net.box')
```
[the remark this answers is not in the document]
CuriousGeorgiy: > The 'fix #xxxx' style title does not describe the change that makes the title useless.

But this is not the title style you describe: I explicitly say that this commit adds a test for a specific issue (for brevity tagged with a number).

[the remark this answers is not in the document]
Totktonada: Well, to be concise: an issue ID is useless in a commit title and personally I consider it as a bad style.

cmake/prefix.cmake
```diff
@@ -8,23 +8,3 @@ if ((CMAKE_C_COMPILER_ID MATCHES "Clang" AND (CMAKE_C_COMPILER_VERSION VERSION_G
     (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "8" OR CMAKE_C_COMPILER_VERSION VERSION_EQUAL "8")))
     set(PREFIX_MAP_FLAGS "-fmacro-prefix-map=${CMAKE_SOURCE_DIR}=.")
 endif()
-
-# Assuming the C and C++ compiler IDs are equal.
-# For clang `MATCHES` is used to match both Clang and AppleClang.
-# Since this flag requires an extra command to tell the debugger where to find
-# the source files, which is not convenient for developers, it is disabled when
-# `DEV_BUILD` option is turned on.
-if (NOT DEV_BUILD AND
```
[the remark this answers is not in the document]
CuriousGeorgiy: Removed the redundant reference.

CMakeLists.txt
```diff
@@ -53,8 +53,6 @@ if(NOT EXISTS ${PROJECT_SOURCE_DIR}/src/lib/small/CMakeLists.txt)
     endif()
 endif()
 
-option(DEV_BUILD "Enable build options that make debugging more convenient" OFF)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Updated the changelog accordingly.

Gumix (approved):

locker (approved):

src/CMakeLists.txt
Totktonada: Okay.

I would like to have details that are not obvious from the code itself in a commit message, but I don't insist -- closing the discussion so.

CuriousGeorgiy: > I would like to have details that are not obvious from the code itself in a commit message.

I will keep that in mind for the future, thanks for your feedback!

locker:
Cherry-picked to 2.10.

PR #7852, opened 2022-10-27
lua/fio: clearer error msg for pread()

### LINKED ISSUE #4963: fio: vague error message when calling pread without arguments
Following snippet should be self explanatory:

```
tarantool> f = fio.open('out.json', {'O_RDONLY'})
---
...

tarantool> f:pread() -- Argument error should be thrown instead
---
- error: 'builtin/buffer.lua:123: attempt to perform arithmetic on ''char *'' and
    ''nil'''
...
```

Lord-KA:
fh:pread() will raise a clear error on wrong arguments.

Closes #4963

NO_DOC=bugfix
NO_CHANGELOG=bugfix

### DIFF
--- a/src/lua/fio.lua
+++ b/src/lua/fio.lua
@@ -100,9 +100,15 @@ end
 -- pread(buf, size, offset) -> len
 fio_methods.pread = function(self, buf, size, offset)
     local tmpbuf
-    if not ffi.istype(const_char_ptr_t, buf) then
+    local has_buf = ffi.istype(const_char_ptr_t, buf)
+    if not has_buf then
         offset = size
         size = buf
+    end
+    if type(size) ~= 'number' or (type(offset) ~= 'number' and offset ~= nil) then
+        error('Usage: fh:pread(buf, size[, offset]) or fh:pread(size[, offset])')
+    end
+    if not has_buf then
         tmpbuf = buffer.ibuf()
         buf = tmpbuf:reserve(size)
     end
--- a/test/app-luatest/gh_4963_fio_pread_err_msg_test.lua
+++ b/test/app-luatest/gh_4963_fio_pread_err_msg_test.lua
@@ -1,0 +1,10 @@
+local fio = require('fio')
+
+local t = require('luatest')
+local g = t.group()
+
+g.test_fio_pread_err_msg = function()
+    local fh = fio.open('/dev/null', {'O_RDONLY'})
+
+    t.assert_error_msg_contains("Usage: fh:pread(buf, size[, offset]) or fh:pread(size[, offset])", fh.pread, fh)
+end


src/lua/fio.lua
```diff
@@ -99,13 +99,21 @@ end
...
 fio_methods.pread = function(self, buf, size, offset)
-    local tmpbuf
+    local no_buf = false
     if not ffi.istype(const_char_ptr_t, buf) then
         offset = size
         size = buf
+        no_buf = true
+    end
+    if (type(size) ~= 'number' or (type(offset) ~= 'number' and offset ~= nil)) then
+        error('usage: fh:pread([buf,] size, offset = 0)')
```
Totktonada, line 109: Nit: Python style named parameter syntax may be confusing in the Lua context (is it positional or a table is assumed around?). I propose the following:

```lua
error('Usage: fh:pread(buf, size[, offset]) or fh:pread(size[, offset])')
```

src/lua/fio.lua
```diff
@@ -99,13 +99,21 @@ end
...
+    end
+    local tmpbuf
+    if (no_buf) then
```
Totktonada, line 112: Nit: Parentheses are not necessary here.

src/lua/fio.lua
```diff
@@ -99,13 +99,21 @@ end
...
 -- pread(buf, size, offset) -> len
 fio_methods.pread = function(self, buf, size, offset)
-    local tmpbuf
+    local no_buf = false
     if not ffi.istype(const_char_ptr_t, buf) then
         offset = size
         size = buf
+        no_buf = true
+    end
+    if (type(size) ~= 'number' or (type(offset) ~= 'number' and offset ~= nil)) then
```
Totktonada, line 108: Nit: Outermost parentheses are not necessary.

Nit: Personally I find this expression a bit too complex -- and/or, several comparisons, negation of what is expected. What if we would have one more parameter?

What do you think about a step-by-step validation?

```lua
local usage = '<...>'
if type(size) ~= 'number' then
    error(usage)
end
if offset ~= nil and type(offset) ~= 'number' then
    error(usage)
end
```

test/app-luatest/gh_4963_fio_pread_err_msg_test.lua
```diff
@@ -0,0 +1,10 @@
+local fio = require('fio')
+
+local t = require('luatest')
+local g = t.group()
+
+g.test_fio_pread_err_msg = function()
+    local fh = fio.open('/tmp/', {'O_RDWR', 'O_TMPFILE'})
```
Totktonada, line 7: Linux specific? I guess `/dev/null` will do the job on most Linux/Mac OS/*BSD systems.

src/lua/fio.lua
Lord-KA: Fixed.

src/lua/fio.lua
Lord-KA: Fixed.

src/lua/fio.lua
Lord-KA: I don't think that dividing the `if` is worth it. Maybe, if we explicitly compare with every format, it would be better:
``` lua
local usage = '...'
if type(size) == 'number' and type(offset) == 'number' then
    usage = nil
end
if type(size) == 'number' and offset == nil then
    usage = nil
end
if (usage) then
    error(usage)
end
```

But I feel that it is too much in this case.

test/app-luatest/gh_4963_fio_pread_err_msg_test.lua
Lord-KA: Good idea, thanks.

src/lua/fio.lua
Totktonada: Okay.

Totktonada (approved):

igormunkin:
Also backported to 2.10 in scope of 4ec07cb (2.10.3-65-g4ec07cb80) and to 1.10 in scope of f3b6af7 (1.10.14-27-gf3b6af7fb).

PR #7854, opened 2022-10-27
test: replication/transaction flakiness fix

sergos:
The test is expected the replication is broken by the time of the server switch. Apparently, switch won the race, so the expectation is changed for waiting.

Closes https://github.com/tarantool/tarantool-qa/issues/271

NO_CHANGELOG=testing
NO_DOC=testing
NO_TEST=testing

### DIFF
--- a/test/replication/transaction.result
+++ b/test/replication/transaction.result
@@ -62,6 +62,11 @@ test_run:cmd("switch replica")
 ---
 - true
 ...
+-- check replication status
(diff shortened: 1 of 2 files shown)


sergepetrenko (approved):
Thanks! LGTM.

By the way, you don't need the NO_TEST label here, because technically you touch a test, and checkpatch considers this ok.

NickVolynkin (approved):
Awesome!

PR #7855, opened 2022-10-27
luajit: proxy -j and -b flags

### LINKED ISSUE #5541: Introduce -j and -b flags to be proxied to LuaJIT environment
There are two nice flags I often use while debugging or testing LuaJIT:
* `-j` allowing to investigate the compiler behavior and manage JIT engine options.
* `-b` providing bytecode dumping and listing.

Unfortunately, when I try to check the same Lua code with the Tarantool, I need to adjust it, since there are no such options.

fckxorg:
There are two flags in the LuaJIT useful for debugging and runtime configuration purposes: `-j` and `-b`. However, if you want to check the same Lua code from the Tarantool, you will need to make some adjustments in the script itself, as those flags are not present in the Tarantool's CLI.

This patch introduces those flags to the Tarantool, so debugging is much more convenient now. Flags are working the same as they do in the LuaJIT.

Closes #5541

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -617,6 +617,8 @@ print_help(const char *program)
 	puts("  -v, --version\t\t\tprint program version and exit");
 	puts("  -e EXPR\t\t\texecute string 'EXPR'");
 	puts("  -l NAME\t\t\trequire library 'NAME'");
+	puts("  -j cmd\t\t\tperform LuaJIT control command");
+	puts("  -b ...\t\t\tsave or list bytecode");
 	puts("  -i\t\t\t\tenter interactive mode after executing 'SCRIPT'");
 	puts("  --\t\t\t\tstop handling options");
 	puts("  -\t\t\t\texecute stdin and stop handling options");
@@ -624,6 +626,9 @@ print_help(const char *program)
 	puts("Please visit project home page at https://tarantool.org");
 	puts("to see online documentation, submit bugs or contribute a patch.");
 }
+
+#define O_INTERACTIVE 0x1
+#define O_BYTECODE    0x2
 
 int
 main(int argc, char **argv)
@@ -636,7 +641,7 @@ main(int argc, char **argv)
 	fpconv_check();
 
 	/* Enter interactive mode after executing 'script' */
-	bool interactive = false;
+	uint32_t opt_mask = 0;
 	/* Lua interpeter options, e.g. -e and -l */
 	int optc = 0;
 	const char **optv = NULL;
@@ -649,9 +654,10 @@ main(int argc, char **argv)
 		{"version", no_argument, 0, 'v'},
 		{NULL, 0, 0, 0},
 	};
-	static const char *opts = "+hVvie:l:";
+	static const char *opts = "+hVvb::ij:e:l:";
 
 	int ch;
+	bool lj_arg = false;
 	while ((ch = getopt_long(argc, argv, opts, longopts, NULL)) != -1) {
 		switch (ch) {
 		case 'V':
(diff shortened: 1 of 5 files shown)


src/main.cc
```diff
@@ -643,68 +648,79 @@ main(int argc, char **argv)
...
 	};
-	static const char *opts = "+hVvie:l:";
+	static const char *opts = "+hVvbij:e:l:";
 
 	int ch;
+	bool lj_arg = false;
 	while ((ch = getopt_long(argc, argv, opts, longopts, NULL)) != -1) {
 		switch (ch) {
-		case 'V':
-		case 'v':
```
Totktonada, line 665: Please, don't do that. The semantic is the same as for the if-elseif-else chain and it should be indented similarly. There is no logical double nesting despite lexical double nesting. A code should represent a logic behind it, so the logic wins here.

test/box-py/args.test.py
```diff
@@ -59,4 +59,10 @@
                    script + \
                    " 1 2 3 --help")
 
+server.test_option("-b")
+server.test_option("-bl -e ''")
+
+server.test_option("-jon -e ''")
+server.test_option("-j on -e ''")
```
Totktonada, line 66: The checkpatch warning is about a trailing whitespace:

```diff
ERROR:TRAILING_NEWLINE: trailing newline
#444: FILE: test/box-py/args.result:182:
+tarantool -j on -e 
+
```

At least you can do something like the following:

```python
cmd = <...>
res = server.test_option_get(cmd, silent=True)
print(cmd)
print(res)
```

It uses [test_option_get](https://github.com/tarantool/test-run/blob/07a0c00c8b54e1e4bc2a458ce7e76b5cd790eefb/lib/tarantool_server.py#L1198) from test-run.

I would rewrite the test to luatest and drop `test_option` and `test_option_get` from test-run, but not necessarily in this patchset.

src/main.cc
```diff
@@ -670,29 +676,43 @@ main(int argc, char **argv)
 			return 0;
 		case 'i':
 			/* Force interactive mode */
-			interactive = true;
+			opt_mask |= O_INTERACTIVE;
+			break;
+		case 'b':
```
tsafin, line 681: There is one case left unhandled - when there was `-b` option without modifiers, and output file provided. Patched tarantool rejects this case, while original luajit does handle it correctly:

```
✘-1 ~/tarantool/build [tsafin/luajit-hack-0 L|…17⚑ 8]
11:37 $ ./src/tarantool -b ../debug-getsources-0.lua debug-getsources.out
Save LuaJIT bytecode: luajit -b[options] input output
  -l        Only list bytecode.
  -s        Strip debug info (default).
  -g        Keep debug info.
  -n name   Set module name (default: auto-detect from input name).
  -t type   Set output file type (default: auto-detect from output name).
  -a arch   Override architecture for object files (default: native).
  -o os     Override OS for object files (default: native).
  -e chunk  Use chunk string as input.
  --        Stop handling options.
  -         Use stdin as input and/or stdout as output.

File types: c h obj o raw (default)
✔ ~/tarantool/build [tsafin/luajit-hack-0 L|…17⚑ 8]
11:38 $ luajit -b ../debug-getsources-0.lua debug-getsources.out
```

tsafin: Also, we better modify the reported message to include tarantool name instead of luajit.

tsafin (changes requested):

src/main.cc
tsafin: Also I'd like to highlight, that I want this patch landed to master before #7929 where I modify all the same `init.c` for introduction of `-d` option, and given `-b` and `-j` changes from this pull request the possible resultant code with `-d` added similarly would look more generic (and shorter).

tsafin (approved):
Thanks, it's now handling a full set of possible `-b` options (including arg-less). Also, this patch is now part of debugger option code #7929 (I've rebased my PR upon the current one).

igormunkin:
Superseded by #8334. Closing.

PR #7863, opened 2022-10-28
build: enable unjustifiably disabled compiler warnings

### LINKED ISSUE #7862: Enable unjustifiably disabled compiler warnings
After revisiting the list of currently disabled compiler warnings it seems like the following are unjistifably disabled:
* `-Wunknown-pragmas`;
* `-Wstrict-aliasing`;
* `-Wunused-value`;
* `-Wchar-subscripts`;
* `-Wformat-truncation`;
* `-Wcast-function-type`.
 
Let's enable them for the sake of code health.

CuriousGeorgiy:
After revisiting the list of currently disabled compiler warnings it seems like the following are unjustifiably disabled:
* `-Wunknown-pragmas`;
* `-Wstrict-aliasing`;
* `-Wunused-value`;
* `-Wchar-subscripts`;
* `-Wformat-truncation`.

Enable them for the sake of code health.

Closes #7862

NO_CHANGELOG=build
NO_DOC=build
NO_TEST=build

### DIFF
--- a/cmake/compiler.cmake
+++ b/cmake/compiler.cmake
@@ -77,10 +77,6 @@ check_cxx_source_compiles("int main(void) {
 }" HAVE_OPENMP)
 set(CMAKE_REQUIRED_FLAGS "")
 
-if (NOT HAVE_OPENMP)
-    add_compile_flags("C;CXX" -Wno-unknown-pragmas)
-endif()
-
 #
 # GCC started to warn for unused result starting from 4.2, and
 # this is when it introduced -Wno-unused-result
@@ -92,8 +88,6 @@ check_c_compiler_flag("-Wno-parentheses-equality" CC_HAS_WNO_PARENTHESES_EQUALIT
 check_c_compiler_flag("-Wno-tautological-compare" CC_HAS_WNO_TAUTOLOGICAL_COMPARE)
 check_c_compiler_flag("-Wno-misleading-indentation" CC_HAS_WNO_MISLEADING_INDENTATION)
 check_c_compiler_flag("-Wno-varargs" CC_HAS_WNO_VARARGS)
-check_c_compiler_flag("-Wno-char-subscripts" CC_HAS_WNO_CHAR_SUBSCRIPTS)
-check_c_compiler_flag("-Wno-format-truncation" CC_HAS_WNO_FORMAT_TRUNCATION)
 check_c_compiler_flag("-Wno-implicit-fallthrough" CC_HAS_WNO_IMPLICIT_FALLTHROUGH)
 check_c_compiler_flag("-Wno-cast-function-type" CC_HAS_WNO_CAST_FUNCTION_TYPE)
 
@@ -225,8 +219,12 @@ macro(enable_tnt_compile_flags)
     add_compile_flags("C;CXX"
         "-Wall"
         "-Wextra"
-        "-Wno-strict-aliasing"
     )
+    if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_C_COMPILER_VERSION VERSION_LESS "8")
+        # Strict aliasing violation warnings cannot be disabled for libev with
+        # `system_header` pragma.
+        add_compile_flags("C" "-Wno-strict-aliasing")
+    endif()
 
     if (ENABLE_UB_SANITIZER)
         if (NOT CMAKE_COMPILER_IS_CLANG)
@@ -274,19 +272,6 @@ macro(enable_tnt_compile_flags)
 
             add_compile_flags("C;CXX" "${SANITIZE_FLAGS}")
         endif()
-    endif()
-
-    if (CMAKE_COMPILER_IS_CLANG AND CC_HAS_WNO_UNUSED_VALUE)
-        # False-positive warnings for ({ xx = ...; x; }) macroses
-        add_compile_flags("C;CXX" "-Wno-unused-value")
-    endif()
-
-    if (CC_HAS_WNO_CHAR_SUBSCRIPTS)
-        add_compile_flags("C;CXX" "-Wno-char-subscripts")
-    endif()
-
-    if (CC_HAS_WNO_FORMAT_TRUNCATION)
-        add_compile_flags("C;CXX" "-Wno-format-truncation")
     endif()
 
     if (CMAKE_COMPILER_IS_CLANG OR CMAKE_COMPILER_IS_GNUCXX)
--- a/src/box/xlog.c
+++ b/src/box/xlog.c
@@ -732,9 +732,26 @@ xdir_collect_inprogress(struct xdir *xdir)
 			continue;
 
 		char path[PATH_MAX];
-		snprintf(path, sizeof(path), "%s/%s", dirname, dent->d_name);
-		if (unlink(path) < 0)
-			say_syserror("error while removing %s", path);
+		int rc;
+		if ((size_t)snprintf(path, sizeof(path), "%s/%s", dirname,
+				     dent->d_name) >= sizeof(path)) {
+			/*
+			 * If we allocated a large enough buffer for the path,
+			 * unlink would fail with this error according to the
+			 * POSIX standard.
+			 */
+			errno = ENAMETOOLONG;
+			rc = -1;
+		} else {
+			rc = unlink(path);
+		}
+		if (rc < 0)
+			/*
+			 * In case the path got truncated, print the expected
+			 * path.
+			 */
+			say_syserror("error while removing %s/%s",
+				     dirname, dent->d_name);
 		else
 			say_info("removed %s", path);
 	}
--- a/third_party/tarantool_ev.h
+++ b/third_party/tarantool_ev.h
@@ -28,6 +28,9 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  */
+
+#pragma GCC system_header
+
 #include "trivia/config.h"
 #include <time.h>
 #include <sys/types.h>
--- a/src/box/tuple_compare.cc
+++ b/src/box/tuple_compare.cc
@@ -192,8 +192,12 @@ mp_compare_double_any_number(double lhs, const char *rhs,
 	 * Both NaNs. Compare signaling and quiet NaNs by
 	 * 'quiet bit'.
 	 */
-	uint64_t lqbit = *((uint64_t *)&lhs) & (uint64_t)0x8000000000000;
-	uint64_t rqbit = *((uint64_t *)&v) & (uint64_t)0x8000000000000;
+	uint64_t lqbit;
+	memcpy(&lqbit, &lhs, sizeof(lhs));
+	lqbit &= UINT64_C(0x8000000000000);
+	uint64_t rqbit;
+	memcpy(&rqbit, &v, sizeof(v));
+	rqbit &= UINT64_C(0x8000000000000);
 	/*
 	 * Lets consider the quiet NaN (fraction first bit == 1)
 	 * to be bigger than signaling NaN (fraction first
--- a/test/sql-luatest/gh_6572_nan_is_not_null_test.c
+++ b/test/sql-luatest/gh_6572_nan_is_not_null_test.c
@@ -17,7 +17,8 @@ get_nan(box_function_ctx_t *ctx, const char *args, const char *args_end)
 	char res[BUF_SIZE];
 	memset(res, 0, BUF_SIZE);
 	double d;
-	*(uint64_t *)&d = 0xfff8000000000000;
+	uint64_t val = 0xfff8000000000000;
+	memcpy(&d, &val, sizeof(val));
 	assert(isnan(d));
 	char *end = mp_encode_double(res, d);
 	box_return_mp(ctx, res, end);


cmake/compiler.cmake
```diff
@@ -276,19 +271,6 @@ macro(enable_tnt_compile_flags)
...
     endif()
 
-    if (CMAKE_COMPILER_IS_CLANG AND CC_HAS_WNO_UNUSED_VALUE)
-        # False-positive warnings for ({ xx = ...; x; }) macroses
-        add_compile_flags("C;CXX" "-Wno-unused-value")
-    endif()
-
-    if (CC_HAS_WNO_CHAR_SUBSCRIPTS)
-        add_compile_flags("C;CXX" "-Wno-char-subscripts")
-    endif()
```
Totktonada, line 286: It seems `CC_HAS_WNO_CHAR_SUBSCRIPTS` (defined on line 95) itself becomes unneeded. The same likely relevant for other such variables.

CuriousGeorgiy: Thanks for the suggestion, removed all such variables. 

We also have some unused compiler flag checks, but I will leave them for now.

CuriousGeorgiy:
Ignoring linter errors about **libev** sources, since I believe it's better to follow the existing code style.

src/box/xlog.c
```diff
@@ -731,7 +731,7 @@ xdir_collect_inprogress(struct xdir *xdir)
 		if (ext == NULL || strcmp(ext, inprogress_suffix) != 0)
 			continue;
 
-		char path[PATH_MAX];
+		char path[PATH_MAX * 2];
```
Gumix, line 734: Why?

third_party/libev/ev.h
```diff
@@ -696,16 +696,21 @@ EV_API_DECL void ev_resume  (EV_P) EV_NOEXCEPT;
 
 /* these may evaluate ev multiple times, and the other arguments at most once */
 /* either use ev_init + ev_TYPE_set, or the ev_TYPE_init macro, below, to first initialise a watcher */
-#define ev_init(ev,cb_) do {			\
-  ((ev_watcher *)(void *)(ev))->active  =	\
-  ((ev_watcher *)(void *)(ev))->pending = 0;	\
-  ev_set_priority ((ev), 0);			\
-  ev_set_cb ((ev), cb_);			\
+#define ev_init(ev,cb_) do { 							\
```
Gumix, line 699: Not sure whether the changes to libev should be placed in a separate commit or not.

third_party/libev/ev.h
```diff
@@ -696,16 +696,21 @@ EV_API_DECL void ev_resume  (EV_P) EV_NOEXCEPT;
...
 } while (0)
 
 #define ev_io_modify(ev,events_)             do { (ev)->events = (ev)->events & EV__IOFDSET | (events_); } while (0)
 #define ev_io_set(ev,fd_,events_)            do { (ev)->fd = (fd_); (ev)->events = (events_) | EV__IOFDSET; } while (0)
-#define ev_timer_set(ev,after_,repeat_)      do { ((ev_watcher_time *)(ev))->at = (after_); (ev)->repeat = (repeat_); } while (0)
+#define ev_timer_set(ev,at_,repeat_) do {       								\
+  ev_tstamp __ev_timer_set_at = (at_);          								\
+  memcpy ((char *)(ev) + offsetof(ev_watcher_time, at), &__ev_timer_set_at, sizeof(__ev_timer_set_at));		\
+  ev_tstamp __ev_timer_set_repeat = (repeat_);  								\
+  memcpy ((char *)(ev) + offsetof(ev_timer, repeat), &__ev_timer_set_repeat, sizeof(__ev_timer_set_repeat));	\
```
Gumix, line 712: In the original line ` (ev)->repeat = ` worked without type casting, is this `memcpy` really needed?

src/box/xlog.c
locker, line 734: A path shouldn't be longer than `PATH_MAX` by definition. Doubling the buffer makes no sense to me.

third_party/libev/ev.h
locker, line 712: I don't think that we should patch libev like this - it will complicate updating to a new version. Better live without strict aliasing IMO.

BTW checkpatch complains about mixed tabs and spaces in this patch.

src/box/xlog.c
CuriousGeorgiy: Discussed f2f: the compiler warning should be ignored.

third_party/libev/ev.h
CuriousGeorgiy: Discussed f2f: compiler warnings from libunwind sources should be ignored.

third_party/libev/ev.h
CuriousGeorgiy: Dropped this change.

CuriousGeorgiy: Discussed f2f: compiler warnings from libunwind sources should be ignored — dropped these changes.

src/box/xlog.c
CuriousGeorgiy: Discussed f2f: the compiler warning should be ignored.

Gumix (approved):

src/box/xlog.c
```diff
@@ -732,7 +732,14 @@ xdir_collect_inprogress(struct xdir *xdir)
 			continue;
 
 		char path[PATH_MAX];
+#pragma GCC diagnostic push
+		/*
+		 * We ignore this warning, since file paths are inherently
+		 * limited by `PATH_MAX`.
+		 */
+#pragma GCC diagnostic ignored "-Wformat-truncation"
```
locker, line 740: I find this warning useless, because it's clearly stated in the `snrpintf` manual page that this function may truncate output. We often truncate output when we write something to the log and this is expected. According to the [gcc manual](https://man7.org/linux/man-pages/man1/gcc.1.html), the warning is "based on heuristics" and "may also result in false positives". Also, the warning isn't implemented in clang. Taking this all into account, I'd rather leave it disabled. I'm leaving this decision up to @alyapunov.

Anyway, using the pragma to suppress the warning doesn't look nice. It'd be better to check the `snprintf` return value instead. I guess we should print an error if the resulting path doesn't fit in the buffer.

CMakeLists.txt
```diff
@@ -301,7 +301,7 @@ include_directories(${PROJECT_SOURCE_DIR}/src/lib)
 include_directories(${PROJECT_SOURCE_DIR}/src/lib/small/include)
 include_directories(${PROJECT_SOURCE_DIR}/src/lib/small/third_party)
 include_directories(${PROJECT_SOURCE_DIR}/src/lib/core)
-include_directories(${PROJECT_SOURCE_DIR}/third_party)
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party)
```
locker, line 304: Setting the `SYSTEM` option may have some extra side-effects aside ignoring warnings. According to the [doc](https://cmake.org/cmake/help/latest/command/include_directories.html), "these fixed-install system files not being considered in dependency calculations". I don't think we should set it for our internal files that may actually be updated. @Totktonada should know better.

CuriousGeorgiy: @locker you quoted out of context:

> Signalling this setting might achieve effects such as the compiler skipping warnings, or these fixed-install system files not being considered in dependency calculations - see compiler docs.

The key words here are _see compiler docs_. AFAI, what CMake does for [**GNU**](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html) is passing the the `-isystem` flag:
> Header files found in directories added to the search path with the -isystem and -idirafter command-line options are treated as system headers for the purposes of diagnostics.

Same goes for [**Clang**](https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers)

src/box/xlog.c
Totktonada: NB: The warning should disappear if the return value of `snprintf()` is checked. I don't remember precisely where I read it -- likely somewhere in a discussion in the GCC bug tracker.

CuriousGeorgiy: Yeah, we already discussed that actually this warning is a good point, and we should actually check the return value and log an error if the path got truncated: updated the patch accordingly.

src/box/xlog.c
```diff
@@ -732,12 +732,19 @@ xdir_collect_inprogress(struct xdir *xdir)
 			continue;
 
 		char path[PATH_MAX];
-		snprintf(path, sizeof(path), "%s/%s", dirname, dent->d_name);
+		int would_be_written = snprintf(path, sizeof(path), "%s/%s",
+						dirname, dent->d_name);
+		if ((size_t)would_be_written >= sizeof(path)) {
+			say_error("in-progress file's path got truncated: %s/%s",
+				  dirname, dent->d_name);
+			goto out_closedir;
```
locker, line 740: Nit:
- The error message looks confusing: a path was truncated - so what?
- No need to add a label - to break the loop, you can use `break`. Anyway, I don't see much point in breaking the loop.

Suggested diff:

```diff
diff --git a/src/box/xlog.c b/src/box/xlog.c
index cedbdff3bb23..bec53c7d4751 100644
--- a/src/box/xlog.c
+++ b/src/box/xlog.c
@@ -730,10 +730,16 @@ xdir_collect_inprogress(struct xdir *xdir)
 		char *ext = strrchr(dent->d_name, '.');
 		if (ext == NULL || strcmp(ext, inprogress_suffix) != 0)
 			continue;
-
+		int rc;
 		char path[PATH_MAX];
-		snprintf(path, sizeof(path), "%s/%s", dirname, dent->d_name);
-		if (unlink(path) < 0)
+		if ((size_t)snprintf(path, sizeof(path), "%s/%s",
+				     dirname, dent->d_name) >= sizeof(path)) {
+			errno = ENAMETOOLONG;
+			rc = -1;
+		} else {
+			rc = unlink(path);
+		}
+		if (rc != 0)
 			say_syserror("error while removing %s", path);
 		else
 			say_info("removed %s", path);
```

cmake/compiler.cmake
```diff
@@ -276,19 +269,6 @@ macro(enable_tnt_compile_flags)
         endif()
     endif()
 
-    if (CMAKE_COMPILER_IS_CLANG AND CC_HAS_WNO_UNUSED_VALUE)
-        # False-positive warnings for ({ xx = ...; x; }) macroses
-        add_compile_flags("C;CXX" "-Wno-unused-value")
-    endif()
```
Totktonada, line 282: It is interesting, because we actually have such code in small and tarantool itself.

I found a reason behind this suppression: https://github.com/llvm/llvm-project/issues/14119.

According to the LLVM [release dates](https://releases.llvm.org/), the last affected clang version was 3.1 and it was hopefully fixed in clang 3.2. AFAIS, neither of currently alive (or recently EOLed) distros offer such old clang versions.

Feel free to add this information into the commit message or ignore.

CMakeLists.txt
Totktonada: > @Totktonada should know better.

Hm. Someone called me a CMake professional? I'm definitely not one.

To be honest, personally I think that CMake is even worse than Bourne shell, but taking into account that it was born 20 years afterwards, it is purely insane thing. And it has worst documentation I ever saw in a popular project.

----

Vladimir said about cmake dependency resolution (rebuild everything that needs to be rebuilt after changing a file). It is a cmake's thing, not a compiler's one.

I'm a bit worried about that too, but not only about that.

There are libraries/headers in the `third_party` directory that are copied verbatim, there are ones that are patched at some points, there are ones that were rewritten significantly. There are ones with evolving upstreams and time-to-time pulldowns into our source tree, but there are ones, where upstreams changing rarely.

IOW, there are libraries of different kinds and it looks improper to disable warning in all of them due to one or two affected ones.

I see the following instruments around:

* Fix warnings. In context of libev it seems counter-productive, because it makes pulldowns more cumbersome.
* Use the `include_directories(SYSTEM <...>)` hack.
* Use a [precompiled headers hack](https://gitlab.kitware.com/cmake/cmake/-/issues/21219).
* Add pragmas for pushing/popping the diagnostics for specific headers that needs it.

The latter seems less intruisive as in affecting other third party libraries as well as in changing library's code itself.

src/box/xlog.c
CuriousGeorgiy: I would rather show the expected path in `say_syserror`, i.e.:
```C
say_syserror("error while removing %s/%s", dirname, dent->d_name);
```

CuriousGeorgiy: I also believe mimicing a system error here is kind of wrong: if you look at the POSIX specification of [`unlink`](https://pubs.opengroup.org/onlinepubs/007908799/xsh/unlink.html), it fails with `ENAMETOOLONG` under completely different conditions.

CuriousGeorgiy: I checked the POSIX standard and mimicking a system error here is indeed correct.

CuriousGeorgiy: Applied your diff with my nit above.

cmake/compiler.cmake
CuriousGeorgiy: Added your remark to the commit message, thanks for the research!

CMakeLists.txt
CuriousGeorgiy: Thanks for elaborating on the problem and for the suggested solutions: I chose to go with the latter.

third_party/tarantool_ev.h
```diff
@@ -28,6 +28,10 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  */
+
+/* Clang supports GCC's pragma. */
+#pragma GCC system_header
```
locker, line 33: Nit: It'd be nice to have a comment here explaining why we need to mark this header as system.

locker (approved):

third_party/tarantool_ev.h
CuriousGeorgiy: Added a brief comment about the need to ignore `-Wstrict-aliasing` compiler warnings from libev.

Totktonada (approved):

CMakeLists.txt
CuriousGeorgiy: Looks like ignoring `-Wstrict-aliasing` won't work at all: https://godbolt.org/z/Gzx1MG7Wh — I will try to report this bug on GCC.

CuriousGeorgiy: Yes, but the warning is supposed to be issued at the point where macro is defined — you can try out other warnings, and they will not be issued.

CuriousGeorgiy: > especially since old GNU compilers don't seem to support the system_header pragma well

Looks like there was also a bug in **GCC** with system headers that was only fixed in GCC 8: https://godbolt.org/z/xr54exzn7.> especially since old GNU compilers don't seem to support the system_header pragma well

Looks like there was also a bug in **GCC** with system headers that was only fixed in GCC 8: https://godbolt.org/z/xr54exzn7.

Totktonada: Warnings are for developers and developers are usually use relatively fresh compiler version. I see no problem in disabling a particular warning for the whole project for an old (but still used) compiler version. It is the same as we previously do for `-Wubused-value` and `-Wunknown-pragmas`.

I would even say that we can ignore the warnings, but we have `-Werror` builds and it would be inconvenient.

CuriousGeorgiy: I suggest the following solution:

* use `system_header` pragma;
* for GNU compilers older than 8 use `-Wno-strict-aliasing`.

CuriousGeorgiy: > I would prefer to disable only necessary warnings, because I would like to at least look at new warnings if they'll appear.

As I pointed out in the commit message, libev is a special case, where the practice is to simply ignore all warnings coming from it.

Totktonada: > I suggest the following solution:
> 
> * use `system_header` pragma for libev;
> * leave `-Wno-strict-aliasing` for GNU compilers older than 8.

Agreed.

> As I pointed out in the commit message, libev is a special case, for which the practice advocated even by library authors themselves is to simply ignore all compiler warnings coming from it.

I would still prefer to analyze new warnings explicitly disregarding on how self-confident authors of third party libraries are.

But since we have no working way to selectively disable this particular warning when it comes from a particular header, I'm okay with the proposed solution.

locker:
@CuriousGeorgiy The issue (#7862) isn't in the plan. According to the teamC policy, everything done by teamC has to be in the teamC plan. Please ask @alyapunov to add the issue to the plan and add a label to the ticket determining whether it should be backported to stable branches or not (label "2.11" or "2.10").

PR #7864, opened 2022-10-28
test: fix flakiness of replication/qsync_basic

sergos:
The async transactions should always wait for LSN from the master after switch to replica. Surprising, there was a place where replica expected to have data from master, that master rolls back.

Closes tarantool/tarantool-qa#274

NO_CHANGELOG=testing
NO_DOC=testing

### DIFF
--- a/test/replication/qsync_basic.result
+++ b/test/replication/qsync_basic.result
@@ -162,9 +162,12 @@ box.space.sync:select{}
  | ...
 
 -- Check consistency on replica.
-test_run:cmd('switch replica')
+test_run:switch('replica')
  | ---
  | - true
+ | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
  | ...
 box.space.sync:select{}
  | ---
@@ -256,6 +259,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
+ | ...
 box.space.test:select{5}
  | ---
  | - - [5]
@@ -292,6 +298,9 @@ box.space.sync:select{6}
 test_run:switch('replica')
  | ---
  | - true
+ | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
  | ...
 box.space.test:select{6}
  | ---
@@ -355,6 +364,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
+ | ...
 box.space.sync:select{8}
  | ---
  | - - [8]
@@ -409,6 +421,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
+ | ...
 box.space.sync:select{9}
  | ---
  | - []
@@ -459,6 +474,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
+ | ...
 box.space.sync:select{10}
  | ---
  | - - [10]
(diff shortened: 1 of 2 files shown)


test/replication/qsync_basic.result
```diff
@@ -222,6 +225,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
```
sergepetrenko, line 228: Strictly speaking this `wait_lsn` isn't needed. The `replace{4}` is replicated synchronously, so by the time the command returns on master it must've already been replicated to the replia.

You may leave that wait_lsn here if you want to make sure that `CONFIRM` reaches the replica as well. I wouldn't bother though.

test/replication/qsync_basic.result
```diff
@@ -270,7 +279,7 @@ test_run:switch('default')
  | ---
  | - true
  | ...
-box.cfg{replication_synchro_timeout = 0.001, replication_synchro_quorum = 3}
+box.cfg{replication_synchro_timeout = 0.1, replication_synchro_quorum = 3}
```
sergepetrenko, line 282: Better leave a tiny timeout here. Quorum is 3 so there is no chance the transaction below gets committed. No need to wait for extra `0.1` second there

test/replication/qsync_basic.result
```diff
@@ -374,7 +389,7 @@ test_run:switch('default')
...
```
sergepetrenko, line 392: Same as above, the smaller the timeout, the faster the test will finish. And there's no reason to wait more in this case

test/replication/qsync_basic.result
```diff
@@ -589,6 +616,9 @@ test_run:switch('replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
```
sergepetrenko, line 619: No need to do `wait_lsn` here. You wait until the transaction is committed on master, so the value must be visible on replica too (unless `memtx_mvcc` gets turned on by default in tests)

sergepetrenko (changes requested):
Thanks for the patch!

You said that the result waited for a rolled back value somewhere, but I couldn't find that place.

test/replication/qsync_basic.result
sergos: agree, removed

test/replication/qsync_basic.result
sergos: Yep, the idea behind was the async followed after sync can be either in or not in the limbo. That's not true, fixed.

test/replication/qsync_basic.result
sergos: fixed

test/replication/qsync_basic.result
sergos: right

test/replication/qsync_basic.result
```diff
@@ -409,6 +421,9 @@ test_run:switch('replica')
...
+ | ---
+ | ...
 box.space.sync:select{9}
  | ---
  | - []
```
sergepetrenko, line 429: AFAIR, this is the place we talked about (regarding `Surprising, there was a place where replica expected
to have data from master, that master rolls back.`)

Seems it's not here, so let's not mention it in the commit mesage.

sergepetrenko (changes requested):
Thanks. One nit left. LGTM otherwise.

test/replication/qsync_basic.result
sergos: Yep, you're right. At some point the following select from the test space resulted in [9]. Perhaps it was because of intermittent changes with a triggering change in relay. Removing.

sergepetrenko (approved):
Thanks! LGTM.

NickVolynkin (approved):

PR #7870, opened 2022-10-31
relay: fix loss of one of the two consecutive heartbeats

### LINKED ISSUE #7869: relay delays downstream vclock update for a `replication_timeout`
**Bug description**
Relay seems to postpone replica's ack delivery to tx thread, when there's another ack en route. Found while debugging `qsync_basic` test hanging on these lines:
https://github.com/tarantool/tarantool/blob/eed09192ba165055f7f82588e77f962d2f8835c1/test/replication/qsync_basic.test.lua#L83-L92
Due to `replication_timeout = 1000`, delaying the ack for a timeout leads to the test hanging on waiting for quorum of acks on the last line.

sergepetrenko:
When replica heartbeat is read while another heartbeat is en route, it's delivery is delayed for a full replication_timeout.

This can be seen in replication/qsync_basic.test.lua occasional hangs on these lines:
```lua
--
-- gh-5100: replica should send ACKs for sync transactions after
-- WAL write immediately, not waiting for replication timeout or
-- a CONFIRM.
--
-- on replica:
box.cfg{replication_timeout = 1000, replication_synchro_timeout = 1000}
test_run:switch('default')
box.cfg{replication_timeout = 1000, replication_synchro_timeout = 1000}
<factored-out>
-- Now commit something sync. It should return immediately even
-- though the replication timeout is huge.
box.space.sync:replace{4}

```

Change that so that a pending heartbeat is sent immediately upon status message return from tx thread.

Closes #7869

NO_TEST=hard to test
NO_DOC=bugfix
NO_CHANGELOG=not user-visible

### DIFF
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -561,6 +561,10 @@ relay_final_join(struct iostream *io, uint64_t sync,
 	});
 }
 
+/** Check if status update is needed and send it if possible. */
+static void
+relay_check_status_needs_update(struct relay *relay);
+
 /**
  * The message which updated tx thread with a new vclock has returned back
  * to the relay.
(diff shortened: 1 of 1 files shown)


sergos (approved):

Gerold103 (approved):
Thanks for the patch!

sergepetrenko:
@Gerold103, could you push this, please?

PR #7875, opened 2022-11-01
sql: fix error with new field in new index

### LINKED ISSUE #5183: bug: select is wrong
Tarantool version: 2.2.2-4-g4f8ac5999
OS version: Debian GNU/Linux 9
Bug description: select is wrong
Steps to reproduce:

format = {{'i', 'integer'}}
s = box.schema.space.create('A', {format = format})
_ = s:create_index('i')
box.execute('insert into a values (1), (2);')
format = {{'i', 'integer'}, {'a', 'integer', is_nullable = true}, {'b', 'integer', is_nullable = true}}
s:format(format)

box.execute('select * from a where "b" is NULL;')
- metadata:
  - name: i
    type: integer
  - name: a
    type: integer
  - name: b
    type: integer
  rows:
  - [1, null, null]
  - [2, null, null]

s:create_index('b',{parts={'b'}, unique = false})
box.execute('select * from a where "b" is NULL;')
- metadata:
  - name: i
    type: integer
  - name: a
    type: integer
  - name: b
    type: integer
  rows:
  - [1, null, null]

it is wrong,.. show only first record!

ImeevMA:
In case the space format is changed after inserting tuples, the new format contains new fields, and a new index is created that works with those new fields, it is possible to get an assertion during SELECT. This problem was fixed in this path.

Closes #5183

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-5183-fix-index-field-missing.md
+++ b/changelogs/unreleased/gh-5183-fix-index-field-missing.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* Creating index on newly added fields no longer leads to a wrong
+  select (gh-5183).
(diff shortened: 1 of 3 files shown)


Gerold103 (approved):
Thanks for the patch!

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7878, opened 2022-11-01
debugger: proper readline support, step.3

### LINKED ISSUE #7738: debugger followup: make debugger REPL compatible with Tarantool console
After the debugger.1 patch (#7593) we have situation when we could not load debugger from inside of Tarantol console REPL due to incompatibility with their readline implementation.
```
tarantool> dbg = require 'luadebug'
---
...

tarantool> dbg() << here we will hung

```

- we should reuse Tarantool console readline functionality for that (i.e. extracting the basic module into `internal.console` for example, and then reusing it both in `console.lua` and `luadebug.lua`);
- make sure debugger luatest is not fragile due to unpredictable popen behavior;

tsafin:
Implemented readline history and autocomplete via reusing of Tarantool console readline facilities. They were hidden once Lua 'console' module loaded, now they kept visible as 'console_internal'.

Followup #7593 
Closes #7738

### DIFF
--- a/changelogs/unreleased/gh-7593-console-debugger.md
+++ b/changelogs/unreleased/gh-7593-console-debugger.md
@@ -1,0 +1,3 @@
+## feature/debugger
+
+* Introduced readline support to tarantool debugger (gh-7738).
--- a/test/app-luatest/console_debugger_session_test.lua
+++ b/test/app-luatest/console_debugger_session_test.lua
@@ -1,7 +1,6 @@
 local t = require('luatest')
-local popen = require('popen')
-local tnt = require('tarantool')
 local fio = require('fio')
+local it = require('test.interactive_tarantool')
 
 local function unescape(s)
     return s and s:gsub('[\27\155][][()#;?%d]*[A-PRZcf-ntqry=><~]', '') or ''
@@ -11,17 +10,7 @@ local function trim(s)
     return s and s:gsub('%s+$', '') or ''
 end
 
-local function tarantool_path(arg)
-    local index = -2
-    -- arg[-1] is guaranteed to be non-null
-    while arg[index] do index = index - 1 end
-    return arg[index + 1]
-end
-
-local TARANTOOL_PATH = tarantool_path(arg)
-
 local DEBUGGER = 'luadebug'
-local dbg_header = tnt.package .. " debugger " .. tnt.version
 local dbg_prompt = DEBUGGER .. '>'
 local dbg_failed_bp = 'command expects argument in format filename:NN or ' ..
                       'filename+NN, where NN should be a positive number.'
@@ -146,17 +135,17 @@ local sequences = {
     12 | os.exit(0)
 ]],
         { ['\t'] = '' }, -- \t is a special value for start
-        { ['n'] = dbg_prompt },
-        { ['s'] = dbg_prompt },
-        { ['n'] = dbg_prompt },
-        { ['n'] = dbg_prompt },
+        { ['n'] = '' },
+        { ['s'] = '' },
+        { ['n'] = '' },
+        { ['n'] = '' },
         { ['p'] = 'expects argument, but none received' },
         { ['p obj'] = 'obj => {"tzoffset" = "+0300", "hour" = 3}' },
-        { ['n'] = dbg_prompt },
+        { ['n'] = '' },
         { ['p ymd'] = 'ymd => false' },
         { ['w'] = 'local hms = false' },
         { ['bogus;'] = 'is not recognized.' },
-        { ['h'] = dbg_prompt },
+        { ['h'] = '' },
         { ['t'] = '=> builtin/datetime.lua' },
         { ['u'] = '{file}:3 in chunk at' },
         -- FIXME (gh-8190) - we should not show calling side at luadebug.lua
@@ -165,15 +154,15 @@ local sequences = {
         { ['d'] = '{file}:3 in chunk at' },
         { ['d'] = 'Inspecting frame: builtin/datetime.lua' },
         { ['l'] = 'obj => {"tzoffset" = "+0300", "hour" = 3}' },
-        { ['f'] = dbg_prompt },
-        { ['n'] = dbg_prompt },
+        { ['f'] = '' },
+        { ['n'] = '' },
         { ['p T'] = 'T => 1970-01-01T03:00:00+0300' },
-        { ['n'] = dbg_prompt },
-        { [''] = dbg_prompt },
-        { [''] = dbg_prompt },
-        { [''] = dbg_prompt },
+        { ['n'] = '' },
+        { [''] = '' },
+        { [''] = '' },
+        { [''] = '' },
         { ['p S'] = 'S => "1970-01-01T0300+0300"' },
-        { ['c'] = '' },
+        { ['c'] = '<END_OF_EXECUTION>' },
     },
 
     -- partial sequence with successful breakpoints added
@@ -193,12 +182,12 @@ local sequences = {
     12 | os.exit(0)
 ]],
         { ['\t'] = '' }, -- \t is a special value for start
-        { ['b +9'] = dbg_prompt },
+        { ['b +9'] = '' },
         { ['c'] = '{file}:9' },
-        { ['n'] = dbg_prompt },
+        { ['n'] = '' },
         { ['p S'] = 'S => "1970-01-01T0300+0300"' },
         { ['p T'] = 'T => 1970-01-01T03:00:00+0300' },
-        { ['c'] = '' },
+        { ['c'] = '<END_OF_EXECUTION>' },
     },
 
     -- partial sequence with failed breakpoint addsitions
@@ -267,21 +256,23 @@ local sequences = {
     },
 }
 
-local function run_debug_session(cmdline, sequence, tmpfile, header)
+local function run_debug_session(sequence, tmpfile, flags)
     local short_src = fio.basename(tmpfile)
     --[[
         repeat multiple times to check all command aliases
     ]]
     for i = 1, MAX_ALIASES_COUNT do
-        local fh = popen.new(cmdline, {
-            stdout = popen.opts.PIPE,
-            stderr = popen.opts.PIPE,
-            stdin = popen.opts.PIPE,
-        })
-        t.assert_is_not(fh, nil)
-        -- for -d option we expect debugger banner in the stderr,
-        -- but for self invoke debugger there is no extra header.
-        local first = header ~= nil
+        local args
+        if flags then
+            args = {flags, tmpfile}
+        else
+            args = {tmpfile}
+        end
+        local expect_header = type(flags) == 'string' and flags:find("-d")
+        local child = it.new_debugger({args = args,
+                                       expect_header = expect_header})
+
+        debuglog(child:read_until_prompt(dbg_prompt))
         for _, row in pairs(sequence) do
             local cmd, expected = next(row)
             -- interpret template strings {file}
@@ -290,36 +281,22 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
                 if cmd ~= '' then
                     local key, arg = get_key_arg_pair(cmd)
                     arg = arg and (' ' .. arg) or ''
-                    cmd = get_cmd_alias(key, i) .. arg .. '\n'
-                else -- '' empty command - repeat prior one
-                    cmd = cmd .. '\n'
+                    cmd = get_cmd_alias(key, i) .. arg
                 end
-                fh:write(cmd)
-                debuglog('Execute command: "'..trim(cmd)..'"')
+                debuglog('Execute command: "' .. trim(cmd) .. '"')
+                child:execute_command(cmd)
             end
 
-            local result
-            local clean_cmd = trim(cmd)
-            -- there should be empty stderr - check it before stdout
-            local errout = fh:read({ timeout = 0.05, stderr = true})
-            debuglog('stderr output: ', trim(errout))
-            if first and errout then
-                -- we do not expect anything on stderr
-                -- with exception of initial debugger header
-                t.assert_str_contains(trim(errout), header, false)
-                first = false
+            if expected == '<END_OF_EXECUTION>' then
+                t.assert_error_msg_contains('Unexpected EOF',
+                    child.read_until_prompt, child)
             else
-                t.assert(errout == nil or trim(errout) == '')
-            end
-            repeat
-                result = trim(unescape(fh:read({ timeout = 0.5 })))
-                debuglog('stdout output:', result)
-            until result ~= '' or result ~= clean_cmd
-            if expected ~= '' then
-                t.assert_str_contains(result, expected, false)
+                local res = trim(unescape(child:read_until_prompt(dbg_prompt)))
+                debuglog(res)
+                t.assert_str_contains(res, expected)
             end
         end
-        fh:close()
+        child:close()
     end
     debuglog('Delete temporary file ', tmpfile)
     os.remove(tmpfile)
@@ -334,8 +311,7 @@ g.test_debugger = function(cg)
     local tmpfile = table.remove(sequence, 1)
     t.assert(type(tmpfile) == 'string')
 
-    local cmd = { TARANTOOL_PATH, '-d', tmpfile }
-    run_debug_session(cmd, sequences[scenario_name], tmpfile, dbg_header)
+    run_debug_session(sequences[scenario_name], tmpfile, '-d')
 end
 
 local g_self = t.group('self-debug')
@@ -352,10 +328,10 @@ local shorter_sequence = {
     8 | os.exit(0)
 ]],
     { ['\t'] = '' }, -- \t is a special value for start
-    { ['n'] = dbg_prompt },
-    { ['s'] = dbg_prompt },
-    { ['n'] = dbg_prompt },
-    { ['n'] = dbg_prompt },
+    { ['n'] = '' },
+    { ['s'] = '' },
+    { ['n'] = '' },
+    { ['n'] = '' },
     { ['p'] = 'expects argument, but none received' },
     { ['p obj'] = 'obj => {"tzoffset" = "+0300", "hour" = 3}' },
     { ['u'] = '{file}:5 in chunk at' },
@@ -363,7 +339,7 @@ local shorter_sequence = {
     { ['d'] = 'Inspecting frame: builtin/datetime.lua' },
     { ['l'] = 'obj => {"tzoffset" = "+0300", "hour" = 3}' },
     { ['f'] = '{file}:6 in chunk at' },
-    { ['c'] = '' },
+    { ['c'] = '<END_OF_EXECUTION>' },
 }
 
 -- `tarantool debug-self-target.lua`, where
@@ -372,6 +348,5 @@ g_self.test_debug_self_invoke = function()
     local tmpfile = table.remove(shorter_sequence, 1)
     t.assert(type(tmpfile) == 'string')
 
-    local cmd = { TARANTOOL_PATH, tmpfile }
-    run_debug_session(cmd, shorter_sequence, tmpfile)
+    run_debug_session(shorter_sequence, tmpfile)
 end
--- a/test/interactive_tarantool.lua
+++ b/test/interactive_tarantool.lua
@@ -6,6 +6,7 @@ local fiber = require('fiber')
 local log = require('log')
 local yaml = require('yaml')
 local popen = require('popen')
+local tnt = require('tarantool')
 
 -- Default timeout for expecting an input on child's stdout.
 --
@@ -17,6 +18,8 @@ local M = {}
 
 local mt = {}
 mt.__index = mt
+
+local dbg_header = tnt.package .. " debugger " .. tnt.version
 
 -- {{{ Instance methods
 
@@ -45,7 +48,14 @@ function mt._start_stderr_logger(self)
 end
 
 function mt._stop_stderr_logger(self)
-    self._stderr_logger:cancel()
+    if self._stderr_logger == nil then
+        return
+    end
+    -- The server could have finished already, thus
+    -- kill only what needs to be killed.
+    if self._stderr_logger:status() ~= 'dead' then
+        self._stderr_logger:cancel()
+    end
     self._stderr_logger = nil
 end
 
@@ -79,7 +89,7 @@ function mt.read_line(self, opts)
     return line
 end
 
--- Returns all readed lines (including expected one).
+-- Returns all read lines (including expected one).
 function mt.read_until_line(self, exp_line, opts)
     local opts = opts or {}
     local deadline = opts.deadline or (fiber.clock() + TIMEOUT)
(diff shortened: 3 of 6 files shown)


changelogs/unreleased/gh-7593-console-debugger.md
```diff
@@ -2,20 +2,5 @@
...
->...
->
->tarantool> dbg()
->---
->```
-> Users should call debugger activation only in their instrumented code, not
-> from the interactive console.
-
 * Introduced a new Lua API `tarantool.debug.getsources()` which allows
   seeing sources of builtin modules in any external debugger.
```
andreyaksenov, line 6: Minor suggestions regarding changelog (for example, builtin -> built-in):
```
* Introduced a new console debugger `luadebug.lua` for debugging external and
  built-in Lua modules.
* Introduced a new Lua API `tarantool.debug.getsources()`, which allows
  seeing sources of built-in modules in any external debugger.
```

andreyaksenov (approved):

p7nov (approved):
Please apply the changelog suggestion

changelogs/unreleased/gh-7593-console-debugger.md
Lord-KA: Applied.

third_party/lua/luadebug.lua
```diff
@@ -27,6 +27,7 @@
 local dbg
 
 local DEBUGGER = 'luadebug'
+local HISRTORYFILE = '.tdbg_history'
```
Totktonada, line 30: Nit: Should it actually be named as '**hisrtory** file'? :)

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -298,25 +298,38 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
-                debuglog('stdout output:', result)
-            until result ~= '' or result ~= clean_cmd
-            if expected ~= '' then
-                t.assert_str_contains(result, expected, false)
+
+            -- The output is returned in several chunks, first the prompt,
+            -- then the command, and finally the result. In most cases it
+            -- is can be read in one go, but to prevent flacking the reading
+            -- may be retried.
+            if (expected ~= '') then
```
Totktonada, line 324: Nit: Redundant parentheses.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -298,25 +298,38 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
+                t.helpers.retrying({timeout = 3, delay = 0.05},
+                function(expected)
+                    local result = trim(unescape(fh:read({ timeout = 0.5 })))
+                    t.assert_str_contains(result, expected, false)
+                    debuglog('stdout output:', result)
+                end, expected)
+            else
+                fh:read({ timeout = 0.5 })
```
Totktonada, line 332: What is purpose of this `:read()`?

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -298,25 +298,38 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
+            -- then the command, and finally the result. In most cases it
+            -- is can be read in one go, but to prevent flacking the reading
+            -- may be retried.
+            if (expected ~= '') then
+                t.helpers.retrying({timeout = 3, delay = 0.05},
+                function(expected)
+                    local result = trim(unescape(fh:read({ timeout = 0.5 })))
+                    t.assert_str_contains(result, expected, false)
+                    debuglog('stdout output:', result)
+                end, expected)
```
Totktonada, line 330: It doesn't solve the problem from the comment above:

> The output is returned in several chunks

For example, you're looking for `foo`. The first chunk is `fo`, the second is `o`. This code tries to find `foo` in `fo` and `o` instead of looking into the concatenated result.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -298,25 +298,38 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
-            local result
-            local clean_cmd = trim(cmd)
-            -- there should be empty stderr - check it before stdout
-            local errout = fh:read({ timeout = 0.05, stderr = true})
-            debuglog('stderr output: ', trim(errout))
-            if first and errout then
-                -- we do not expect anything on stderr
-                -- with exception of initial debugger header
-                t.assert_str_contains(trim(errout), header, false)
+            if first then
```
Totktonada, line 301: I see no direct relation of the test changes to the code changes. Please, either extract it into its own commit with a proper description of the solved problem or make this relation clear in the commit message.

third_party/lua/luadebug.lua
```diff
@@ -97,11 +98,19 @@ local stack_top = 0
...
 local stack_inspect_offset = 0
 
--- Default dbg.read function
-local function dbg_read(prompt)
-    dbg.write(prompt)
-    io.flush()
-    return io.read()
+-- Tarantool console compatible readline support.
+local function dbg_readline(prompt)
+    local console = require 'console.lib'
```
Totktonada, line 103: From the commit message:

> now they kept visible as 'console_internal'.

Does you mean `console.lib`?

Totktonada: I would add that it is changed in c432e9e918f1caa69a3e6050830c9f2734217ba7.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: To clear out the buffer for the followup reads. While we don't need to check the output in some rare cases, it seems like a good idea to separate it for different test cases by clearing the buffer.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: The comment explains what is ment by 'chunk' - one of the followup: prompt, cmd, result.

third_party/lua/luadebug.lua
Lord-KA: Yep, fixed.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Fixed.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: (I don't mind concatenating the results, but it doesn't seem necessary)

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Done.
>  The according test had to be ammended due to flaks on macOS.

third_party/lua/luadebug.lua
Lord-KA: Fixed:
>   Implemented readline history and autocomplete by reusing
  readline facilities of Tarantool console. They used to be
  being hidden once Lua 'console' module is loaded. With c432e9e 
  (lua: don't use public module name as internal one), now they 
  are available as 'console.lib'.

test/app-luatest/console_debugger_session_test.lua
Totktonada: Answering me here doesn't make the idea obvious from the code. Please, address such review comments in the patchset itself.

However, I'm not satisfied by the answer. As for me, it is wrong in two ways:

1. If some output is produced by a test case that is expected to don't produce any output, the problem will be hidden (instead of failing the next test case). At least, I see no check that this `:read()` call is timed out.
2. `read()` system call (and, so, our `:read()` from the pipe) has no guarantee like 'it returns everything that were written on the other side'. If two bytes were written, we can receive just one. So, in fact, it is 'usually cleans the pipe buffer, but sometimes doesn't'. It is common for byte streams. Attempt to work with a byte stream like it is a message stream is wrong.

My proposals:

* Let the next test case fail. (And properly comment how it works.)
* Trigger some response that is guaranteed to be written after the response that is expected to be empty.

test/app-luatest/console_debugger_session_test.lua
Totktonada: This is the behavior that you observe on some particular setup. It is not guaranteed by `read()`. I would prefer to lean on guarantees, especially when the work is about a test case stabilization.

If you're interested, whether it matters in practice: yes, it does. I've observed reading from a pipe with randomly splitted chunks during work on the interactive_tarantool.lua helper.

test/app-luatest/console_debugger_session_test.lua
Totktonada: Does it mean that the test starts to sporadically fail on macOS after the given changes in the code? I can't understand it from the comment. If it is so, I'm OK to place it into the same commit. However, if we would stabilize the test case first, it would look better, IMHO.

`checkpatch` doesn't impose on the `NO_TEST` annotation due to these changes. However, we know that the new behavior is left untested. I don't insist on testing it (it is cumbersome), but, please, leave the annotation with a reason.

Nit: there is no noun 'flak', at least it is not the same as adjective 'flaky'.

changelogs/unreleased/gh-7593-console-debugger.md
Totktonada: It looks irrelevant now. I'm closing the discussion.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: I rewrote the test using updated `interactive_tarantool`.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Removed the line and moved all test changes in separate commit.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -290,36 +279,21 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
-            local errout = fh:read({ timeout = 0.05, stderr = true})
-            debuglog('stderr output: ', trim(errout))
-            if first and errout then
-                -- we do not expect anything on stderr
-                -- with exception of initial debugger header
-                t.assert_str_contains(trim(errout), header, false)
-                first = false
-            else
-                t.assert(errout == nil or trim(errout) == '')
-            end
```
Totktonada, line 313: This commit is marked as refactoring, but actually it eliminates the stderr check. I don't know, whether it is important, but it should be mentioned in the commit message at least. If it is not important, the commit message should describe a reason.

test/interactive_tarantool.lua
```diff
@@ -94,6 +98,23 @@ function mt.read_until_line(self, exp_line, opts)
     return res
 end
 
+-- Returns all read lines (excluding one with expected prompt).
+function mt.read_until_prompt(self, exp_prompt, opts)
```
Totktonada, line 102: Can't we use `self.prompt` here?

test/interactive_tarantool.lua
```diff
@@ -263,18 +288,29 @@ function M.new(opts)
     local res = setmetatable({
         ph = ph,
         _readahead_buffer = '',
-        _prompt = 'tarantool> ',
+        _prompt = opts.prompt or 'tarantool> ',
```
Totktonada, line 291: Nit: I would place `local prompt = opts.prompt` next to `local args = opts.args` for consistency and to have a list of all the options at glance on the function beginning.

test/interactive_tarantool.lua
```diff
@@ -263,18 +288,29 @@ function M.new(opts)
...
     }, mt)
 
     -- Log child's stderr.
     res:_start_stderr_logger()
 
+    return res
+end
+
+function M.new_debugger(opts)
+    opts.prompt = opts.prompt or 'luadebug> '
```
Totktonada, line 301: Nit: I would add `local opts = opts or {}` like in `_new_internal()` to allow to call this function without arguments (if no options are needed).

test/interactive_tarantool.lua
```diff
@@ -138,11 +159,18 @@ end
...
+-- compare them for equality.
 local function _prepare_command_for_compare(x)
-    x = x:gsub(' ', ''):gsub(M.CR, '')
+    -- The terrifying regexp in the last gsub is meant to clean ANSI
+    -- color codes, for more details on the format see the doc:
+    -- https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+    x = x:gsub(' ', '')
+         :gsub(M.CR, '')
+         :gsub(M.TAB, '')
+         :gsub('[\x1b\x9b][][()#;?%d]*[A-PRZcf-ntqry=><~]', '')
```
Totktonada, line 173: If it is just about colors, it should be much simpler. According to the link something like `:gsub(M.ESC .. '%[[0-9;]*m', '')` should work. If this is not only about the colors, the comment seems misleading.

Totktonada: BTW, I would put such a regexp into the exposed module constants (close to M.EXP and others). It can be useful outside of the module itself.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -290,36 +279,21 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
-            repeat
-                result = trim(unescape(fh:read({ timeout = 0.5 })))
-                debuglog('stdout output:', result)
-            until result ~= '' or result ~= clean_cmd
-            if expected ~= '' then
-                t.assert_str_contains(result, expected, false)
+            if expected ~= '<END_OF_EXECUTION>' then
```
Totktonada, line 288: Shouldn't we expect EOF in the `else` branch?

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -290,36 +279,21 @@ local function run_debug_session(cmdline, sequence, tmpfile, header)
...
+                local res = trim(unescape(child:read_until_prompt(dbg_prompt)))
+                debuglog(res)
+                t.assert_str_contains(res, expected)
             end
         end
-        fh:close()
+        if child ~= nil then
+            child:close()
+        end
```
Totktonada, line 296: If I don't miss anything, `child` can't be `nil` here.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Is there a way to expect EOF with the current `it` helpers?

test/interactive_tarantool.lua
Lord-KA: Ok.

test/interactive_tarantool.lua
Lord-KA: It doesn't seem necessary, but I am not in the mood to argue about it.

test/interactive_tarantool.lua
Lord-KA: Done.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Added a remark in commit msg.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Removed the check.

test/interactive_tarantool.lua
Lord-KA: Good idea! Moved the `:gsub()` to `M.DECOLOR_RE`.

Lord-KA:
I've reworked a bit historyfile saving, now it is being saved to `$HOME/.tdbg_history` and only when `$HOME` env var is set.

test/app-luatest/console_debugger_session_test.lua
Totktonada: OK. I'm relying on your investigation here.

test/interactive_tarantool.lua
Totktonada: It looks much more accurate now.

Nit: I'm OK with regexp `M.DECOLOR_RE` or function `M.decolor`. But, please name/place it in sync with other functions/constants.

Lord-KA: Oh, my bad, will fix.

Totktonada (approved):
Visually I see no problems in the patchset.

Thank you for refactoring the test!

test/interactive_tarantool.lua
Lord-KA: Fixed.

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Added `t.assert_error_msg_contains()` to check expected EOF.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -1,7 +1,6 @@
 local t = require('luatest')
```
ligurio, line 1: commit "test: refactor dbgr test with it helper" breaks tests and next commit fixes it. please mention it in commit message

ligurio (commented):
Gleb, thanks for the patches! see my comments

test/interactive_tarantool.lua
```diff
@@ -138,11 +168,16 @@ end
...
--- This function drop duplicates that goes in row, strips <CR> and
--- spaces. Applying of this function to a source command and
--- readline's echoed command allows to compare them for equality.
+-- This function drop duplicates that goes in row, strips <CR>,
+-- spaces, tabs and ANSI color codes. Applying of this function
+-- to a source command and readline's echoed command allows to
+-- compare them for equality.
 local function _prepare_command_for_compare(x)
-    x = x:gsub(' ', ''):gsub(M.CR, '')
+    x = x:gsub(' ', '')
```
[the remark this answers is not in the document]
ligurio: I believe our goal is not preserving original style but producing patches with good quality.

test/interactive_tarantool.lua
```diff
@@ -119,6 +141,14 @@ function mt.assert_data(self, exp_data, opts)
     end
 end
 
+local function decolor(string)
+    assert(type(string) == 'string')
+    -- The gsub is meant to clean ANSI color codes, for more details on
+    -- the format see the doc:
+    -- https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
```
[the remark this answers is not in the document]
Lord-KA: Ok, replaced the doc with [1].

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -11,17 +10,7 @@ local function trim(s)
...
 
-local function tarantool_path(arg)
-    local index = -2
-    -- arg[-1] is guaranteed to be non-null
-    while arg[index] do index = index - 1 end
-    return arg[index + 1]
-end
-
-local TARANTOOL_PATH = tarantool_path(arg)
-
```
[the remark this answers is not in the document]
Lord-KA: It is a part of stderr check removal.

test/interactive_tarantool.lua
```diff
@@ -94,6 +98,24 @@ function mt.read_until_line(self, exp_line, opts)
     return res
 end
 
+-- Returns all read lines (excluding one with expected prompt).
```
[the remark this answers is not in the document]
Lord-KA: Ok...

test/interactive_tarantool.lua
```diff
@@ -263,18 +296,30 @@ function M.new(opts)
...
+
+function M.new(opts)
+    local res = M._new_internal(opts)
+
     -- Write a command and ignore the echoed output.
     --
     -- ReadLine 6 writes <ESC>[?1034h at beginning, it may hit
     -- assertions on the child's output.
     --
-    -- This sequence of charcters is smm ('set meta mode')
```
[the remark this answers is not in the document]
Lord-KA: >   * now there are several less tipos in comments

test/app-luatest/console_debugger_session_test.lua
Lord-KA: Done:
>   This patch breaks the test without debugger readline support patch,
  that is next in the patchset.

test/app-luatest/console_debugger_session_test.lua
```diff
@@ -267,21 +256,21 @@ local sequences = {
...
     for i = 1, MAX_ALIASES_COUNT do
-        local fh = popen.new(cmdline, {
-            stdout = popen.opts.PIPE,
-            stderr = popen.opts.PIPE,
-            stdin = popen.opts.PIPE,
-        })
-        t.assert_is_not(fh, nil)
-        -- for -d option we expect debugger banner in the stderr,
-        -- but for self invoke debugger there is no extra header.
-        local first = header ~= nil
```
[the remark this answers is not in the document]
Lord-KA: I removed the check because it is hard to implement using `interactive_tarantool`. `it` has stderr_logger being run in a different fiber and I couldn't get an after-test check for empty stderr working in reasonable time. As we don't expect anything from stderr, and get all the right stdout output, it seems like the test is comprehensive enough.

ylobankov (approved):

test/interactive_tarantool.lua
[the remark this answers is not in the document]
ligurio: > Please refer to a style-guide rule. If there isn't one, it seems like a matter of taste and can be left as is.

Actually, no, it is not a matter of taste.

Style guide cannot contain all possible rules. If we try to describe all the possible rules to follow and rules that no one should break, then we will be like lawmakers, but this is not our job.

Instead of referring to a certain rule in a style guide, I will appeal to common sense and my understanding of a good programming habits, to respecting expectations of people who will read my code and use these programs.

It could be unexpected when function mutates passed arguments. Function has no any comment so without reading source code you will never know that `x` was updated. Sometimes such behavior in public functions could be a reason of real bugs, see https://github.com/tarantool/crud/issues/192.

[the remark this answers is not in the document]
Lord-KA: The bug you referred to is about a different matter. In Lua, when string is passed as an argument, like in `_prepare_command_for_compare()`, it is being copied and local variable is modified, unlike the example in https://github.com/tarantool/crud/issues/192, where a table is passed by reference.

test/app-luatest/console_debugger_session_test.lua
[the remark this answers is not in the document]
Lord-KA: Ok, I've had another go at this, and now it seems to be working fine. `read_until_prompt()` in `interactive_tarantool` now expects empty stderr, while `new_debugger()` expects debugger header, similar to the old test.

test/interactive_tarantool.lua
```diff
@@ -94,6 +104,31 @@ function mt.read_until_line(self, exp_line, opts)
...
+                                                                1))
+    new_buffer = self._prompt .. new_buffer
+
+    self._readahead_buffer = new_buffer
+
+    local stderr, err = self.ph:read({timeout = 0.05, stderr = true})
+    if stderr ~= "" and not (stderr == nil and
+                             tostring(err) == "timed out") then
+        error(("Unexpected stderr output: %s"):format(stderr))
+    end
```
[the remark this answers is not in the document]
Lord-KA: No, I turned off `stderr_logger` for debugger.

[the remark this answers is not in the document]
Totktonada: I see now, thanks for the explanation.

test/app-luatest/console_debugger_session_test.lua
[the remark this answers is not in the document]
ligurio: right, not a dead code, it was used below

test/app-luatest/console_debugger_session_test.lua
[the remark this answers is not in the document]
ligurio: Thanks!

ligurio (approved):
LGTM, thanks, Gleb!

igormunkin:
@Lord-KA, thanks for the changes! Could you please add the reason of reverting the changes in scope of the last patch of the series. You can see an example in 6d414e7.

Lord-KA:
@igormunkin, done.

PR #7879, opened 2022-11-01
replication: say warning if before_replace trigger changed replica_id

### LINKED ISSUE #7846: `_cluster:before_replace()` crash if replica ID is altered
```Lua
--
-- Instance 1
--
-- Step 1
--
box.cfg{
    listen = 3313,
}
box.space._cluster:before_replace(function(old, new)
    return new:update({{'+', 1, 10}})
end)

--
-- Instance 2
--
-- Step 2
--
box.cfg{
    listen = 3314,
    replication = 3313,
}
```
Result:
```
tarantool> Assertion failed: (replica_by_uuid(uuid)->id == id), function box_register_replica, file /Users/gerold/Work/Repositories/tarantool/src/box/box.cc, line 3124.
Process 37607 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = hit program assert
    frame #4: 0x00000001001666ec tarantool`box_register_replica(id=2, uuid=0x000000010b8ffac0) at box.cc:3124:2
   3121		if (boxk(IPROTO_INSERT, BOX_CLUSTER_ID, "[%u%s]",
   3122			 (unsigned) id, tt_uuid_str(uuid)) != 0)
   3123			diag_raise();
-> 3124		assert(replica_by_uuid(uuid)->id == id);
   3125	}
   3126	
   3127	/**
Target 0: (tarantool) stopped.
```
Shouldn't crash. Just get the resulting replica ID from the replica object and that is it. Otherwise it might make sense to ban certain triggers on system spaces, like `before_replace()`, which allow to change data.

drewdzzz:
Currently, there is an assertion that checks if replica_id haven't been changed after it's registration. Let's replace this assertion with warning message about id being altered - new id will be used for affected replica.

Closes #7846

### DIFF
--- a/test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
+++ b/test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
@@ -1,0 +1,56 @@
+local server = require('test.luatest_helpers.server')
+local t = require('luatest')
+local urilib = require('uri')
+
+local g = t.group('before_replace alter',
+    t.helpers.matrix{ret_nil = {true, false}})
+
+g.test_before_replace_alter_replica_id = function(cg)
+    local ret_nil = cg.params.ret_nil
+    local server1 = server:new{
+        alias = 'server1',
+    }
+    server1:start()
+    server1:exec(function(ret_nil)
+        local trigger
+        if ret_nil == false then
+            trigger = function(_old, new)
+                return new:update({{'+', 1, 10}})
+            end
+        else
+            trigger = function(_old, _new)
+                return nil
+            end
+        end
+        box.space._cluster:before_replace(trigger)
+    end, {ret_nil})
+    local uri = urilib.parse(server1.net_box_uri)
+    local server2 = server:new{
+        alias = 'server2',
+        box_cfg = {replication = uri.unix}
+    }
+
+    if ret_nil == false then
+        server2:start()
+        local msg = server1:grep_log('Replica ID is changed by a trigger', 1024)
+        t.assert_not_equals(msg, nil)
+    else
+        -- We cannot wait until ready because server2 will not be started at all,
+        -- so we will retry until we find a log entry we need.
+        server2:start({wait_for_readiness=false})
+        t.helpers.retrying({}, function()
+            local msg = server1:grep_log('Replica ID is changed by a trigger', 1024)
+            if msg == nil then
+                error('Cannot find replica connection error')
+            end
+        end)
+    end
+
+    if ret_nil == false then
+        server2:drop()
+    else
+        -- Cannot drop server because it was not started.
+        server2:cleanup()
+    end
+    server1:drop()
+end
--- a/changelogs/unreleased/gh_7846_before_replace_crash_on_replica_id_alter.md
+++ b/changelogs/unreleased/gh_7846_before_replace_crash_on_replica_id_alter.md
@@ -1,0 +1,6 @@
+## bugfix/replication
+
+* Say a warning when replica_id is changed by before_replace trigger while
+  adding a new replica (gh-7846). There was an assertion checking this before.
+  Also, handle the case when before_replace set on space `_cluster` returns
+  nil - it caused segmentation fault before.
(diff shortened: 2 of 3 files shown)


Gerold103 (changes requested):
It still crashes if I do this:
```Lua
--
-- Instance 1
--
-- Step 1
--
box.cfg{
    listen = 3313,
}
box.space._cluster:before_replace(function(old, new)
    return nil
end)

--
-- Instance 2
--
-- Step 2
--
box.cfg{
    listen = 3314,
    replication = 3313,
}
```

Also the ticket mentions a more general problem with the triggers. They can break other system spaces too. For example, this is what happens in `_space`:
```Lua
box.cfg{}
box.space._space:before_replace(function(old, new)
    return nil
end)
box.schema.create_space('test')
```
```
tarantool> box.schema.create_space('test')
---
- null
- created
...
```
AFAIU this function is not supposed to return `null` ever. And what is `"created"`?

drewdzzz:
I think that we should merge this patch and create a new issue describing a global problem.
You suggested to ban `before_replace` trigger on system spaces, but it is a bad option - AFAIK advanced users of tarantool uses this "feature". So teamP should decide how to deal with this problem.

unera:
I think that we should merge the PR and check the other system space triggers' problems separately.

Gerold103 (changes requested):
Please, look at my previous comment again. `_cluster` still crashes on `before_replace` trigger even after your patch. Ok, you can move the other spaces into another ticket, but I suggest to fix this particular space properly here.

drewdzzz:
OK, I will fix all the problems with space `_cluster` that you describe, but I will not investigate them in the scope of the current problem. Please let me know if you find more problems.

Gerold103:
There is just one more bug with `_cluster`. It is already investigated - I gave you the second reproducer. The reproducer is different from the old one literally in a single Lua line. And the crash is in the same line in box.cc as the crash in the ticket. Do I really have to make a new ticket for each tiny bug in a single line of C++ code?

test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
```diff
@@ -0,0 +1,57 @@
+local server = require('test.luatest_helpers.server')
+local t = require('luatest')
+local urilib = require('uri')
+local fiber = require('fiber')
+
+local g = t.group('before_replace alter')
+
+g.test_replica_id_inc = function()
```
Gerold103, line 8: This and `test_replica_id_nullify` are almost identical. You could make them one function which would take the trigger's full body or a flag whether it should delete or replace the ID. And then call that one function 2 times. Would be ~x2 times shorter.

test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
```diff
@@ -0,0 +1,57 @@
...
+    local server2 = server:new{
+        alias = 'server2',
+        box_cfg = {replication = uri.unix}
+    }
+    -- We cannot wait until ready because server2 will not be started at all,
+    -- so let's just sleep here.
+    server2:start({wait_for_readiness=false})
+    fiber.sleep(1)
+
+    local msg = server1:grep_log('Replica ID is changed by a trigger', 1024)
```
Gerold103, line 53: Relying on 1 second being enough doesn't look good - 1) sometimes it won't be enough for sure, 2) most of the time it is too long. I would rather replace it with `t.helpers.retrying()` where you would wait on the log entry appearance. Also I am afraid that `server2:cleanup()` if done before the server is terminated might leave the server running. Wouldn't `:drop()` work?

Gerold103 (approved):
Thanks for the fixes! Looks fine except some stuff in the tests. You can proceed to a second review anyway.

test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
drewdzzz: `drop()` does not work because `stop()` method, which is called before `cleanup()`, throws an error, because the server was not started. So `cleanup()` is the best option I've found.

test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
drewdzzz: Done

test/box-luatest/gh_7846_crash_on_replica_id_alter_test.lua
drewdzzz: Replaced sleep with retry

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7881, opened 2022-11-02
Pagination followups

drewdzzz:
- Fix a bug when pagination is used with offset
- Redesign position descriptor - now it is base64 string


src/box/index.cc
```diff
@@ -631,29 +629,18 @@ iterator_position(struct iterator *it, const char **pos, uint32_t *size)
 	return it->position(it, pos, size);
 }
 
-/** Keys for iterator position map. All the keys must be uint. */
-enum {
-	ITERATOR_POSITION_KEY = 0,
-};
+static const int iterator_pos_b64_opts = BASE64_NOPAD | BASE64_NOWRAP;
```
locker, line 632: Nit: Please use a define instead: a static variable is allocated on bss, which makes no sense in this case.

src/box/index.cc
```diff
@@ -631,29 +629,18 @@ iterator_position(struct iterator *it, const char **pos, uint32_t *size)
...
 
 uint32_t
-iterator_position_pack_size(const char *pos, const char *pos_end,
-			    uint32_t part_count)
+iterator_position_pack_size(const char *pos, const char *pos_end)
```
locker, line 635: AFAIU It's not an *exact* pack size - it's the *maximal* size of packed data. If so, it's better to name this function `iterator_position_pack_size_max`.

src/box/index.h
```diff
@@ -375,48 +375,39 @@ iterator_next(struct iterator *it, struct tuple **ret);
...
+ */
+uint32_t
+iterator_position_unpack_size_max(const char *packed_pos,
+				  const char *packed_pos_end);
 
 /**
- * Unpack iterator position descriptor from MsgPack. Does not allocate memory,
- * so lifetime of returned position is the same as lifetime of passed ptr.
+ * Unpack iterator position descriptor from MsgPack to preallocated buffer.
+ * Returned position is a MsgPack key without MP_ARRAY header.
```
locker, line 400: I don't like this inconsistency between pack and unpack. Let's return the value *with* `MP_ARRAY` header and strip it in `box_select`.

src/box/index.cc
```diff
@@ -690,49 +681,23 @@ iterator_position_unpack(const char *packed_pos, const char *packed_pos_end,
...
-			break;
-		}
-	}
-	if (!has_position)
+	if (mp_typeof(*decoded_pos) != MP_ARRAY)
 		goto fail;
+	array = decoded_pos;
+	*part_count = mp_decode_array(&decoded_pos);
+	*pos = decoded_pos;
+	mp_next(&array);
```
locker, line 698: There's no need to use `mp_next` here - `mp_check` called above already advances the position to the data end. We should probably also check that the advanced position coincides with the end of decoded data (i.e. there's no garbage after the MsgPack array).

src/box/index.h
```diff
@@ -375,48 +375,39 @@ iterator_next(struct iterator *it, struct tuple **ret);
...
-iterator_position_pack_size(const char *pos, const char *pos_end,
-			    uint32_t part_count);
+iterator_position_pack_size(const char *pos, const char *pos_end);
 
 /**
- * Pack iterator position to preallocated buffer. Buffer length must be
- * more or equal than packed size of position.
+ * Pack (encode to base64 format) iterator position with MP_ARRAY header to
+ * preallocated buffer. Buffer length must be more or equal than packed size
+ * of position.
```
locker, line 385: The buffer size must be at least `iterator_position_pack_size`, not "more or equal than packed size of position". And this function should return the actual packed data size somehow AFAIU.

src/box/index.h
```diff
@@ -375,48 +375,39 @@ iterator_next(struct iterator *it, struct tuple **ret);
...
- * so lifetime of returned position is the same as lifetime of passed ptr.
+ * Unpack iterator position descriptor from MsgPack to preallocated buffer.
+ * Returned position is a MsgPack key without MP_ARRAY header.
  * Returns 0 on success, -1 on failure, diag is set.
  */
 int
 iterator_position_unpack(const char *packed_pos, const char *packed_pos_end,
-			 const char **pos, const char **pos_end,
-			 uint32_t *part_count);
+			 char *buf, const char *buf_end, const char **pos,
```
locker, line 405: Nit: We usually pass the buffer size explicitly as `size_t buf_size`, not as `const char *buf_end`.

locker (changes requested):

src/box/index.cc
locker: BTW if pack/unpack have a buf argument, it'd probably be better to name these functions `iterator_position_pack_bufsize`, `iterator_position_unpack_bufsize` to emphasize what they are supposed to be used for.

drewdzzz: Done

src/box/index.h
drewdzzz: Done

src/box/index.cc
drewdzzz: Done

src/box/index.h
drewdzzz: Done

src/box/index.h
drewdzzz: Done

locker (approved):

PR #7889, opened 2022-11-03
sql: throw error on creation of too many indexes

### LINKED ISSUE #5526: No error when no index is created
I have Ubuntu 18.04.
I have Tarantool 2.5, pulled from source today.
I execute these requests:

box.execute([[DROP TABLE add;]])
s=""
  s = "CREATE TABLE add (c0 INT PRIMARY KEY"
for i=1,128 do
  s = s .. ", c" .. i .. " INT UNIQUE"
end
s = s .. ");"
box.execute(s)
box.execute([[SELECT COUNT(*) FROM  "_vindex" WHERE "id" = (SELECT "id" FROM "_vspace" WHERE "name" = 'ADD');]])
box.execute([[INSERT INTO add (c0, c127) VALUES (1, 1);]])
box.execute([[INSERT INTO add (c0, c127) VALUES (2, 1);]])
box.execute([[INSERT INTO add (c0, c128) VALUES (3, 1);]])
box.execute([[INSERT INTO add (c0, c128) VALUES (4, 1);]])

The result of box.execute(s) is: okay.
The result of the SELECT is: 128.
The result of the INSERTs shows: c128 does not have a unique index.
In other words: I create a table with BOX_INDEX_MAX+1 unique columns.
There is no ER_MODIFY_INDEX error but no unique index on c128.

ImeevMA:
Prior to this patch, there was no error of creating too many indexes in SQL. This led to a situation where indexes with IDs greater than BOX_INDEX_MAX were not created, but no error was thrown. For example, in the case of CREATE TABLE, only indexes with an ID less than BOX_INDEX_MAX were created. Now an error is thrown when creating too many indexes.

Closes #5526

NO_DOC=bugfix

### DIFF
--- a/src/box/index_def.h
+++ b/src/box/index_def.h
@@ -400,11 +400,13 @@ index_def_cmp(const struct index_def *key1, const struct index_def *key2);
 /**
  * Check a key definition for violation of various limits.
  *
- * @param index_def   index definition
- * @param old_space   space definition
+ * @param index_def index definition
+ * @param old_space space definition
+ * @retval 0 Success.
+ * @retval -1 Error. Diag is set.
  */
-bool
-index_def_is_valid(struct index_def *index_def, const char *space_name);
+int
+index_def_check(struct index_def *index_def, const char *space_name);
 
 #if defined(__cplusplus)
 } /* extern "C" */
@@ -421,7 +423,7 @@ index_def_dup_xc(const struct index_def *def)
 static inline void
 index_def_check_xc(struct index_def *index_def, const char *space_name)
 {
-	if (! index_def_is_valid(index_def, space_name))
+	if (index_def_check(index_def, space_name) != 0)
 		diag_raise();
 }
 
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -2848,7 +2848,7 @@ sql_create_index(struct Parse *parse) {
 	}
 	index->def->key_def->part_count = new_part_count;
 
-	if (!index_def_is_valid(index->def, def->name)) {
+	if (index_def_check(index->def, def->name) != 0) {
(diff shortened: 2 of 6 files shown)


src/box/sql/build.c
```diff
@@ -2848,8 +2848,10 @@ sql_create_index(struct Parse *parse) {
 	}
 	index->def->key_def->part_count = new_part_count;
 
-	if (!index_def_is_valid(index->def, def->name))
+	if (!index_def_is_valid(index->def, def->name)) {
```
Gerold103, line 2851: As a follow-up patch I would suggest to rename this function to `index_def_check_is_valid()`. Because this function has a side effect - it sets the diag.

Gerold103 (approved):
Thanks for the patch!

src/box/sql/build.c
ImeevMA: Thank you for the review! Done.

src/box/index_def.h
```diff
@@ -404,7 +404,7 @@ index_def_cmp(const struct index_def *key1, const struct index_def *key2);
  * @param old_space   space definition
  */
 bool
-index_def_is_valid(struct index_def *index_def, const char *space_name);
+index_def_check_is_valid(struct index_def *index_def, const char *space_name);
```
locker, line 407: Nit: Conventionally, a function that sets a diag should return 0 on success, -1 on error. Please fix and update the comment.

Also, why not name it simply `index_def_check`? Note, we already have `index_def_check_xc`, which invokes this function and throws on error.

ImeevMA: Thank you for the review! Fixed.

locker (approved):

PR #7893, opened 2022-11-03
misc: get rid of fiber gc

### LINKED ISSUE #5665: Get rid of fiber_gc()
Tarantool version: 2.7.0

OS version: Ubuntu 20.04

Bug description: assert() fires in region_truncate, use-after-free

Steps to reproduce:
Allocate something from fiber region, then call C-function which works e. g. with R/W transactions. It would call fiber_gc() which empties region so assert() would fire later in the caller when it would call region_truncate().
One particular case of this for R/O transactions was filed as https://github.com/tarantool/tarantool/issues/5623 , was fixed in https://github.com/tarantool/tarantool/issues/5427 . Proper solution should be implemented: we should probably get rid of fiber_gc() completely, which requires careful analysis of all places were region_allocate() is used to ensure all such memory is eventually freed - that is now true now, there are even comments in code that some memory is not being freed hoping someone would call fiber_gc() soon enough.

Optional (but very desirable):
* backtrace

tarantool: /home/void/work/tarantool/src/lib/small/small/region.c:76: region_truncate: Assertion `region_used(region) >= used' failed.

Thread 1 "tarantool" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7171859 in __GI_abort () at abort.c:79
#2  0x00007ffff7171729 in __assert_fail_base (fmt=0x7ffff7307588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x555555bd3f5f "region_used(region) >= used", 
    file=0x555555bd3f28 "/home/void/work/tarantool/src/lib/small/small/region.c", line=76, function=<optimized out>) at assert.c:92
#3  0x00007ffff7182f36 in __GI___assert_fail (assertion=0x555555bd3f5f "region_used(region) >= used", file=0x555555bd3f28 "/home/void/work/tarantool/src/lib/small/small/region.c", line=76, 
    function=0x555555bd3fb0 <__PRETTY_FUNCTION__.4439> "region_truncate") at assert.c:101
#4  0x0000555555a536c6 in region_truncate (region=0x7ffff4c00390, used=1024) at /home/void/work/tarantool/src/lib/small/small/region.c:76
#5  0x00005555557729fd in func_c_call (base=0x555555d21c20, args=0x7ffff4dffb70, ret=0x7ffff4dffbc0) at /home/void/work/tarantool/src/box/func.c:610
#6  0x0000555555772e09 in func_call (base=0x555555d21c20, args=0x7ffff4dffb70, ret=0x7ffff4dffbc0) at /home/void/work/tarantool/src/box/func.c:696
#7  0x00005555556cc51c in lbox_func_call (L=0x40000378) at /home/void/work/tarantool/src/box/lua/call.c:879
#8  0x000055555587bc2f in lj_BC_FUNCC ()
#9  0x000055555583bcf6 in lua_pcall (L=0x40000378, nargs=0, nresults=0, errfunc=0) at lj_api.c:1158
#10 0x00007ffff7fc06bd in wrapper (L=0x40000378) at /home/void/work/tarantool/test/box-tap/gh5623helper.c:31
#11 0x000055555587bc2f in lj_BC_FUNCC ()
#12 0x000055555583bcf6 in lua_pcall (L=0x40000378, nargs=0, nresults=0, errfunc=0) at lj_api.c:1158
#13 0x00005555557bb347 in luaT_call (L=0x40000378, nargs=0, nreturns=0) at /home/void/work/tarantool/src/lua/utils.c:1022
#14 0x00005555557b2146 in lua_main (L=0x40000378, argc=1, argv=0x555555cd0f48) at /home/void/work/tarantool/src/lua/init.c:559
#15 0x00005555557b2631 in run_script_f (ap=0x7ffff4c00448) at /home/void/work/tarantool/src/lua/init.c:634
#16 0x00005555555c27c1 in fiber_cxx_invoke(fiber_func, typedef __va_list_tag __va_list_tag *) (f=0x5555557b215f <run_script_f>, ap=0x7ffff4c00448) at /home/void/work/tarantool/src/lib/core/fiber.h:882
#17 0x00005555557e1bac in fiber_loop (data=0x0) at /home/void/work/tarantool/src/lib/core/fiber.c:879
#18 0x0000555555a7ea47 in coro_init () at /home/void/work/tarantool/third_party/coro/coro.c:110

gh5623helper.c:
```
...
int wrapper(lua_State *L)
{
        size_t region_svp = region_used(region);

        void *ptr = region_alloc(region, 1024);
        (void)ptr;

        if (lua_pcall(L, 0, 0, 0) != LUA_OK)
                error(L, "error running function 'f': %s",
                      lua_tostring(L, -1));

        region_truncate(region, region_svp);
        return 0;
}
...
```
It is being called from Lua with an argument being Lua function which calls C-function which calls e. g. boxk(IPROTO_INSERT, ...)

nshy:
As it breaks sane usage of region as a data stack:

```
    size_t region_svp = region_used(&fiber()->gc);
    /* some allocation on fiber gc and usage of allocated memory. */
    region_truncate(&fiber()->gc, region_svp);
```

If in the above snippet one calls a function that in turn calls
`fiber_gc` then the snippet code may have use-after-free and later UB
on truncation.

For this reason let's get read of fiber_gc. However we need to make sure
we won't introduce leaks this way. So before actually removing
fiber_gc we make it perform leak check instead and only after fixing
all the leaks the fiber_gc was removed.

In order to find the leak easily the backtrace of the first fiber gc
allocation that is not truncated is saved and then reported.

In order to catch leaks that are not triggered by the current test suit
and to prevent introducing leaks in future patches the leak check is
added on fiber exit/recycle and for long living system fibers on every loop
iteration.

Leak check in release build is on but without leak backtrace info by
default for performance reasons. Backtrace can be provided by using
`fiber.debug_enable()` knob before starting leaking fiber.

Closes #5665


src/box/memtx_space.c
```diff
@@ -580,18 +585,19 @@ memtx_space_execute_upsert(struct space *space, struct txn *txn,
...
-			return -1;
-
-		new_tuple =
-			space->format->vtab.tuple_new(format, new_data,
-						      new_data + new_size);
+		if (new_data != NULL)
+			new_tuple =
+				space->format->vtab.tuple_new(format, new_data,
+							      new_data +
+								new_size);
```
locker, line 599: Nit: The indentation looks bad. Why not reformat like this?

```C
		if (new_data != NULL) {
			new_tuple = space->format->vtab.tuple_new(
					format, new_data, new_data + new_size);
		}
```

src/box/sql/vdbe.c
```diff
@@ -2069,6 +2069,7 @@ case OP_MakeRecord: {
 	char *tuple = mem_encode_array(pData0, nField, &tuple_size, region);
 	if (tuple == NULL)
 		goto abort_due_to_error;
+
```
locker, line 2072: Nit: Redundant change.

src/box/vy_log.c
```diff
@@ -2393,22 +2402,25 @@ vy_recovery_new_f(va_list ap)
...
 				break;
 			continue;
 		}
 		rc = vy_recovery_process_record(recovery, &record);
+		region_truncate(&fiber()->gc, region_svp);
 		if (rc < 0)
 			break;
-		fiber_gc();
 	}
-	fiber_gc();
```
locker, line 2411: Nit: Why not set `region_svp` outside this loop and simply replace these `fiber_gc` calls with `region_truncate`? This would look less intrusive.

src/lib/core/fiber.h
```diff
@@ -459,24 +460,13 @@ cord_slab_cache(void);
...
- * It is safe to call simple box APIs around tuples, key_defs and
- * so on -- they don't invalidate the allocated data.
- *
  * Each fiber has its own box region. It means that a call of,
  * say, <box_region_used>() will give its own value in different
  * fibers. It also means that a yield does not invalidate data in
  * the box region.
+ *
+ * Before version 2.10 one can only use a subset of API between
+ * box_region_used and box_region_truncate.
```
locker, line 469: I don't understand this comment. What "subset of API"?

Anyway, I'm not sure we should backport this to 2.10.

src/lib/core/fiber.h
```diff
@@ -584,8 +574,20 @@ struct fiber {
...
 	unsigned int stack_id;
-	/* A garbage-collected memory pool. */
+	/** A garbage-collected memory pool. */
 	struct region gc;
+#ifdef ENABLE_BACKTRACE
+	/**
+	 * Backtrace of the first fiber gc allocation that found
+	 * leaked on inspection.
+	 */
+	struct backtrace *leak_bt;
```
locker, line 584: It's not necessary the backtrace of a leak. It's the backtrace of the first allocation. Let's name it accordingly.

src/lib/core/fiber.h
```diff
@@ -584,8 +574,20 @@ struct fiber {
...
+#endif
+	/**
+	 * This much size of fiber gc at the beginning is used for
+	 * fiber internal purpuses.
+	 */
+	size_t gc_internal;
```
locker, line 590: Better name it `gc_initial_size` or something like this then.

src/lib/core/fiber.h
```diff
@@ -691,6 +693,7 @@ struct fiber {
 	/* Fiber parent's backtrace allocated on the 'gc' region. */
 	struct backtrace_lua *parent_bt;
 #endif /* ENABLE_BACKTRACE */
+	bool debug_enabled;
```
locker, line 696: Nit: Missing a comment.

src/lib/core/fiber.c
```diff
@@ -1209,6 +1288,31 @@ fiber_stack_create(struct fiber *fiber, struct slab_cache *slabc,
 	return 0;
 }
 
+void
+fiber_gc_checker_init(struct fiber *fiber)
```
locker, line 1292: Nit: This function should be static.

src/lib/core/fiber.c
```diff
@@ -1246,7 +1350,6 @@ fiber_new_ex(const char *name, const struct fiber_attr *fiber_attr,
 			    fiber->stack, fiber->stack_size);
 
 		region_create(&fiber->gc, &cord->slabc);
-
```
locker, line 1249: Nit: Unnecessary change.

src/lib/core/fiber.c
```diff
@@ -894,13 +888,97 @@ fiber_recycle(struct fiber *fiber)
...
+fiber_check_gc(void)
+{
+	struct fiber *fiber = fiber();
+
+	assert(region_used(&fiber->gc) >= fiber->gc_internal);
+	if (region_used(&fiber->gc) == fiber->gc_internal)
+		return;
+
+#ifdef ENABLE_BACKTRACE
+	if (fiber->debug_enabled) {
```
locker, line 955: Do you really need this flag? Can't you check if `leak_bt` is NULL instead?

test/unit/fiber.cc
```diff
@@ -374,6 +376,104 @@ cord_cojoin_test(void)
...
+		 */
+		s = strstr(s, "leaker_f");
+		fail_unless(s != NULL);
+		s = strstr(s, "fiber_cxx_invoke");
+		fail_unless(s != NULL);
+	} else {
+		const char *msg =
+			"Fiber gc leak is found. "
+			"Leak backtrace is not available. "
+			"Make sure fiber.enable_debug() is called before"
```
locker, line 448: The function is called `debug_enable`.

src/lib/core/fiber.c
```diff
@@ -1879,6 +1985,11 @@ fiber_init(int (*invoke)(fiber_func f, va_list ap))
 	main_cord.loop = ev_default_loop(EVFLAG_AUTO | EVFLAG_ALLOCFD);
 	cord_create(&main_cord, "main");
 	fiber_signal_init();
+#ifndef NDEBUG
+	fiber_debug_enabled = true;
+#else
+	fiber_debug_enabled = false;
+#endif
```
locker, line 1992: Nit: this would probably be done where this variable is defined.

src/lib/core/fiber.h
```diff
@@ -1121,6 +1121,27 @@ fiber_c_invoke(fiber_func f, va_list ap)
...
+ * Currenly in debug mode we collect fiber gc allocation backtrace and print
+ * it if leak is found. Otherwise we only print that leak is found.
+ */
+extern bool fiber_debug_enabled;
+
+/**
+ * If false then we don't abort in debug build if fiber gc leak is found.
+ * Default is true. Used in tests.
+ */
+extern bool fiber_abort_gc_leak;
```
locker, line 1136: The fact that this flag only affects debug builds looks confusing. I think it should affect release builds as well - we just need to initialize it to `false` then, similarly to `fiber_debug_enabled`.

locker, line 1136: The fact that this flag only affects debug builds looks confusing. I think it should affect release builds as well - we just need to initialize it to `false` then, similarly to `fiber_debug_enabled`.

Also, probably better name it `fiber_abort_on_gc_leak`.

src/lib/core/fiber.h
```diff
@@ -1178,8 +1199,6 @@ fiber_cxx_invoke(fiber_func f, va_list ap)
 	}
 }
 
-#endif /* defined(__cplusplus) */
-
 static inline void *
 region_aligned_alloc_cb(void *ctx, size_t size)
```
locker, line 1203: Looks like `region_aligned_alloc_cb` is never used so instead of hiding it from C code, better remove it.

locker (changes requested):

src/lib/core/backtrace.c
```diff
@@ -246,11 +246,15 @@ backtrace_frame_resolve(const struct backtrace_frame *frame,
...
+	 * Not too bad.
+	 */
+	char proc_name_buf[128];
 	int rc = acc->get_proc_name(unw_local_addr_space, (unw_word_t)frame->ip,
-				    proc_name_buf, TT_STATIC_BUF_LEN, offset,
-				    NULL);
-	if (rc != 0) {
+				    proc_name_buf, sizeof(proc_name_buf),
+				    offset, NULL);
+	if (rc != 0 && rc != UNW_ENOMEM) {
```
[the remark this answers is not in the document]
nshy: Checked and added a comment as discussed f2f. Name is truncated in case of UNW_ENOMEM.

It is also documented https://www.nongnu.org/libunwind/man/unw_create_addr_space(3).html

src/lib/core/util.c
```diff
@@ -284,6 +284,17 @@ strlcat(char *dst, const char *src, size_t size)
 }
 #endif
 
+char *
+astrcpy(char **buf, size_t *size, const char *src)
+{
+	size_t len = strlen(src);
+	if (len >= *size)
+		*buf = xrealloc(*buf, len + 1);
```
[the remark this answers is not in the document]
nshy: Made it inline.

src/lib/core/backtrace.h
```diff
@@ -69,8 +69,10 @@ backtrace_collect(struct backtrace *bt, const struct fiber *fiber,
 /*
  * Resolve C/C++ function name and `offset` from `frame`.
  *
- * Returns pointer to a temporary buffer storing the demangled function name:
- * the caller is responsible for making a copy of it.
+ * Returns pointer to demangled function name. Pointer is a record
+ * in procedure name cache with corresponding lifetime.
+ *
+ * May return NULL on resolving failure or OOM.
```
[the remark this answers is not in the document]
nshy: Ok.

src/box/memtx_engine.cc
```diff
@@ -1106,7 +1093,6 @@ memtx_join_f(va_list ap)
 						   data, size);
 			if (rc != 0)
 				break;
-			fiber_gc();
```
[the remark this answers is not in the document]
nshy: Added check instead. Currently there is no leaking gc allocations inside the loop.

src/box/memtx_engine.cc
```diff
@@ -1216,9 +1202,11 @@ memtx_engine_gc_f(va_list va)
 {
 	struct memtx_engine *memtx = va_arg(va, struct memtx_engine *);
 	while (!fiber_is_cancelled()) {
+		FiberGCChecker gc_check;
 		bool stop;
 		ERROR_INJECT_YIELD(ERRINJ_MEMTX_DELAY_GC);
 		memtx_engine_run_gc(memtx, &stop);
+		fiber_check_gc();
```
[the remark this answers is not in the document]
nshy: Removed.

src/box/memtx_space.c
nshy: Done.

src/box/sql/vdbe.c
nshy: Dropped.

src/box/vy_log.c
nshy: Replaced with truncation at the beginning of the loop and after exiting the loop.

src/lib/core/fiber.h
nshy: Rewrote the comment as discussed f2f.

src/lib/core/fiber.h
nshy: Renamed to `first_alloc_bt`.

src/lib/core/fiber.h
nshy: Done.

src/lib/core/fiber.h
nshy: Not applicable anymore since I dropped the field as you suggested below.

src/lib/core/fiber.c
nshy: Done.

src/lib/core/fiber.c
nshy: Dropped.

src/lib/core/fiber.c
nshy: Yep, dropped.

test/unit/fiber.cc
nshy: Fixed.

src/lib/core/fiber.c
nshy: Changed to

```c
#ifndef NDEBUG
bool fiber_debug_enabled = true;
bool fiber_abort_on_gc_leak = true;
#else
bool fiber_debug_enabled = false;
bool fiber_abort_on_gc_leak = false;
#endif
```

src/lib/core/fiber.h
nshy: Fixed.

nshy: Renamed.

src/lib/core/fiber.h
nshy: Removed.

src/lib/core/fiber.c
```diff
@@ -893,14 +892,96 @@ fiber_recycle(struct fiber *fiber)
...
 
+#ifdef ENABLE_BACKTRACE
+/**
+ * Called on allocation on region gc. Saves allocation caller backtrace
+ * to report it later if region gc leak is found.
+ */
+static void
+fiber_on_gc_alloc(struct region *region,
+		  size_t size,
+		  void *opaque)
```
locker, line 918: Nit: Usually, we don't place every argument on its own line:

```diff
diff --git a/src/lib/core/fiber.c b/src/lib/core/fiber.c
index 8ca22cb845d0..6c9ad05c8a56 100644
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -913,9 +913,7 @@ fiber_recycle(struct fiber *fiber)
  * to report it later if region gc leak is found.
  */
 static void
-fiber_on_gc_alloc(struct region *region,
-		  size_t size,
-		  void *opaque)
+fiber_on_gc_alloc(struct region *region, size_t size, void *opaque)
 {
 	(void)region;
 	(void)size;
```

</details>

src/lib/core/fiber.h
```diff
@@ -459,24 +460,14 @@ cord_slab_cache(void);
...
- * so on -- they don't invalidate the allocated data.
- *
  * Each fiber has its own box region. It means that a call of,
  * say, <box_region_used>() will give its own value in different
  * fibers. It also means that a yield does not invalidate data in
  * the box region.
+ *
+ * Before version 2.11 some API like executing a DML statement truncated
+ * fiber region to 0 so that missing box_region_truncate don't lead to leaks.
+ * Starting from 2.11 client code should truncates its allocations.
```
locker, line 470: Nit: Let's rephrase.

Prior to version 2.11, the box region was implicitly cleaned up on transaction commit (see <box_txn_commit>()) so that <box_region_truncate>() wasn't strictly necessary. Starting from version 2.11, it isn't true anymore, and the client code must free all its allocations explicitly.

src/lib/core/fiber.h
```diff
@@ -584,8 +575,20 @@ struct fiber {
 	struct fiber_slice max_slice;
 	/** Valgrind stack id. */
 	unsigned int stack_id;
-	/* A garbage-collected memory pool. */
+	/** A garbage-collected memory pool. */
 	struct region gc;
+#ifdef ENABLE_BACKTRACE
+	/**
+	 * Backtrace of the first fiber gc allocation that does not
+	 * truncated yet.
```
locker, line 583: Nit: Let's mention that it can be null if ... and then ...

src/lib/core/fiber.c
```diff
@@ -893,14 +892,96 @@ fiber_recycle(struct fiber *fiber)
...
+fiber_on_gc_truncate(struct region *region,
+		     size_t used,
+		     void *opaque)
+{
+	(void)region;
+	struct fiber *fiber = opaque;
+
+	assert(used >= fiber->gc_initial_size);
+	if (used == fiber->gc_initial_size)
+		fiber->first_alloc_bt->frame_count = 0;
```
locker, line 946: Why not reset `first_alloc_bt` to null here?

src/lib/core/fiber.h
```diff
@@ -1121,6 +1121,26 @@ fiber_c_invoke(fiber_func f, va_list ap)
...
+/**
+ * Whether debug mode is enabled on not.
+ *
+ * Currenly in debug mode we collect fiber gc allocation backtrace and print
+ * it if leak is found. Otherwise we only print that leak is found.
+ */
+extern bool fiber_debug_enabled;
+
+/**
+ * Whether we should abort if gc leak is found. Used in tests.
```
locker, line 1133: Nit: Let's say what it is equal to by default.

test/unit/fiber.cc
```diff
@@ -374,6 +376,105 @@ cord_cojoin_test(void)
...
+fiber_test_leak_modes()
+{
+	say_logger_init("log.txt", S_ERROR,
+			0 /* nonblock */, "plain");
+
+	fiber_test_leak(true /* backtrace_enabled */);
+	fiber_test_leak(false /* backtrace_enabled */);
+
+	say_logger_free();
+	unlink("log.txt");
```
locker, line 475: Nit: I don't think we need to delete the log file on failure, because it may be useful for investigation. Looks like other unit tests that create a log file don't bother to delete it. It should be fine, since when the test is started by test-run, the file will created in a test directory and deleted on the next test run.

locker (changes requested):

src/box/box.cc
```diff
@@ -328,6 +328,7 @@ box_process_rw(struct request *request, struct space *space,
 		return -1;
 	assert(iproto_type_is_dml(request->type));
 	rmean_collect(rmean_box, request->type, 1);
+	RegionGuard region_guard(&fiber()->gc);
```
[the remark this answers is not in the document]
nshy: Without this guard we have this fail:
```gdb

2022-11-15 11:47:38.157 [624587] main/110/main fiber.c:969 E> Fiber gc leak is found. First leaked fiber gc allocation backtrace:
#1  0x56386c25b219 in fiber_on_gc_alloc+167
#2  0x56386c4a3f5b in region_alloc+197
#3  0x56386c4a50f9 in char* tuple_extract_key_sequential_raw<false>(char const*, char const*, key_def*, int, unsigned int*)+587
#4  0x56386c4a4eac in char* tuple_extract_key_sequential<false>(tuple*, key_def*, int, unsigned int*)+205
#5  0x56386c0aaed2 in tuple_extract_key+52
#6  0x56386c0ab4a2 in request_rebind_to_primary_key+142
#7  0x56386c0afaf5 in space_execute_dml+559
#8  0x56386c0f6ca2 in box_process_rw+311
#9  0x56386c100228 in box_process1+405
#10 0x56386bfdd6ff in tx_process1(cmsg*)+101
#11 0x56386c263697 in cmsg_deliver+48
#12 0x56386c26489c in fiber_pool_f+588
#13 0x56386bfd1a78 in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*)+30
#14 0x56386c25b5b8 in fiber_loop+215
#15 0x56386c5c15cf in coro_init+124
```

src/box/box.cc
```diff
@@ -3581,6 +3585,7 @@ box_process_subscribe(struct iostream *io, const struct xrow_header *header)
 		 */
 		struct raft_request req;
 		box_raft_checkpoint_remote(&req);
+		RegionGuard region_guard(&fiber()->gc);
```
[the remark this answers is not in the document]
nshy: Ok. I added it to save some space on region. Removed.

src/box/box.cc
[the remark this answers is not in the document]
locker: This is a bug - looks like we overlooked it when we switched transactions to `txn->region`. We need to allocate these data from the `txn->region`, like all the other data used by transactions. Please fix in a separate preparatory commit and remove this guard.

src/box/relay.cc
```diff
@@ -742,6 +743,7 @@ relay_reader_f(va_list ap)
 	struct applier_heartbeat *last_recv_ack = &relay->last_recv_ack;
 	try {
 		while (!fiber_is_cancelled()) {
+			FiberGCChecker gc_checker;
```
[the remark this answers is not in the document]
nshy: Fixed.

src/box/sql/build.c
```diff
@@ -2606,6 +2607,7 @@ index_fill_def(struct Parse *parse, struct index *index,
 	index->def->iid = iid;
 	rc = 0;
 cleanup:
+	region_truncate(&fiber()->gc, region_svp);
```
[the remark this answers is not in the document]
nshy: We goto `cleanup` at the end of  `tnt_error` section.

src/box/sql/vdbe.c
```diff
@@ -3532,6 +3531,7 @@ case OP_Update: {
 	assert(rc == 0);
 	rc = box_update(space->def->id, 0, key_mem->z, key_mem->z + key_mem->n,
 			ops, ops + ops_size, 0, NULL);
+	region_truncate(&fiber()->gc, used);
```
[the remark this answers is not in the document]
nshy: Fixed.

src/lib/core/fiber.c
nshy: Ok. Used proper style.

src/lib/core/fiber.h
nshy: Ok.

src/lib/core/fiber.c
nshy: Yeah, it is a footprint of the zero version where `first_alloc_bt` was part of fiber structure. Fixed.

test/unit/fiber.cc
nshy: Done.

src/box/lua/index.c
```diff
@@ -195,12 +204,13 @@ lbox_index_max(lua_State *L)
 	uint32_t space_id = lua_tonumber(L, 1);
 	uint32_t index_id = lua_tonumber(L, 2);
 	size_t key_len;
+	size_t region_svp = region_used(&fiber()->gc);
 	const char *key = lbox_encode_tuple_on_gc(L, 3, &key_len);
```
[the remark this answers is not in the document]
locker: Filed a ticket for this issue: https://github.com/tarantool/tarantool/issues/7939

src/box/memtx_space.c
```diff
@@ -476,16 +477,17 @@ memtx_space_execute_update(struct space *space, struct txn *txn,
 	uint32_t new_size = 0, bsize;
 	struct tuple_format *format = space->format;
 	const char *old_data = tuple_data_range(decompressed, &bsize);
+	size_t region_svp = region_used(&fiber()->gc);
 	const char *new_data =
 		xrow_update_execute(request->tuple, request->tuple_end,
 				    old_data, old_data + bsize, format,
 				    &new_size, request->index_base, NULL);
-	if (new_data == NULL)
-		return -1;
```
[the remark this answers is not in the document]
locker: Filed a ticket for `xrow_update_execute` leaking memory on failure: https://github.com/tarantool/tarantool/issues/7940

Let's fix it in a separate commit. Note, there are a few places where we manually clean up the region after the function failure. We should remove this code in the scope of that patch.

src/box/lua/index.c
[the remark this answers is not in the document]
locker: Decided that we can postpone fixing this issue for later.

src/box/memtx_space.c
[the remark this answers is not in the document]
locker: Decided that it isn't really a bug, because the memory is freed by `fiber_gc` on transaction commit or rollback.

Still, better clean up the region in `xrow_update_execute` on failure, because it's conventional that a function leaves no traces if it fails. This can be done in the same patch.

src/lib/core/cxx_abi.cc
```diff
@@ -53,12 +53,21 @@ cxx_abi_demangle(const char *mangled_name)
...
-	return mangled_name;
+
+	size_t len = strlen(mangled_name) + 1;
+	if (DemangleBuf::instance().len < len) {
+		char *buf = (char *)xrealloc(DemangleBuf::instance().buf, len);
+		DemangleBuf::instance().len = len;
+		DemangleBuf::instance().buf = buf;
+	}
+
+	memcpy(DemangleBuf::instance().buf, mangled_name, len);
```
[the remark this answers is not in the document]
locker: `len` is incremented above so everything's fine.

src/box/vy_log.c
```diff
@@ -2391,8 +2400,10 @@ vy_recovery_new_f(va_list ap)
 
 	int rc;
 	struct xrow_header row;
+	size_t region_svp = region_used(&fiber()->gc);
 	while ((rc = xlog_cursor_next(&cursor, &row, false)) == 0) {
 		struct vy_log_record record;
+		region_truncate(&fiber()->gc, region_svp);
```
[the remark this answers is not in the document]
locker: Done like that, because there's a `continue` statement in the loop so it's more reliable to clean up the region in the beginning of the loop.

src/lib/core/fiber.c
locker: Done like that so that we don't have to reallocate the backtrace object on each first allocation.

src/box/memtx_engine.cc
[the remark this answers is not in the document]
nshy: Added a guard too.

src/lib/core/cxx_abi.cc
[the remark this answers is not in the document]
nshy: Byte for terminating null is already accounted in len.

src/lib/core/backtrace.c
```diff
@@ -246,11 +246,17 @@ backtrace_frame_resolve(const struct backtrace_frame *frame,
...
+	 * Not too bad.
+	 */
+	char proc_name_buf[128];
 	int rc = acc->get_proc_name(unw_local_addr_space, (unw_word_t)frame->ip,
-				    proc_name_buf, TT_STATIC_BUF_LEN, offset,
-				    NULL);
-	if (rc != 0) {
+				    proc_name_buf, sizeof(proc_name_buf),
+				    offset, NULL);
+	/* */
```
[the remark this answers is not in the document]
nshy: Done.

changelogs/unreleased/gh-5665-get-rid-of-fiber_gc.md
```diff
@@ -0,0 +1,6 @@
+## bugfix/core
+
+* **[Breaking change]** box\_region allocations not paired with
+box\_region\_truncate will produce leaks. Previously one can
+rely on some API (like executing a DML statement) truncating fiber
+region to 0. Now API stop truncating memory it does not own.
```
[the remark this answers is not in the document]
nshy: Fixed.

src/box/box.cc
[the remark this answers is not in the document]
nshy: We discussed f2f and decide to `request_handle_sequence` to allocate on txn region on this path.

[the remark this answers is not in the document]
nshy: Fixed as suggested in distinct patch.

src/box/vinyl.c
```diff
@@ -1934,21 +1936,27 @@ vy_update(struct vy_env *env, struct vy_tx *tx, struct txn_stmt *stmt,
...
 	const char *old_tuple = tuple_data_range(stmt->old_tuple, &old_size);
 	const char *old_tuple_end = old_tuple + old_size;
+	size_t region_svp = region_used(&fiber()->gc);
 	new_tuple = xrow_update_execute(request->tuple, request->tuple_end,
 					old_tuple, old_tuple_end,
 					pk->mem_format, &new_size,
 					request->index_base, &column_mask);
-	if (new_tuple == NULL)
+	if (new_tuple == NULL) {
+		region_truncate(&fiber()->gc, region_svp);
```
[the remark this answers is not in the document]
nshy: Fixed.

src/lib/core/fiber.h
nshy:
```с
        /**
         * Backtrace of the first fiber gc allocation that does not
         * truncated yet. NULL if backtrace is not supported by the
         * platform or fiber_debug_enabled == false.
         */
```

src/lib/core/fiber.h
nshy: Done.

src/box/request.h
```diff
@@ -54,7 +55,8 @@ struct tuple;
  */
 int
 request_create_from_tuple(struct request *request, struct space *space,
-			  struct tuple *old_tuple, struct tuple *new_tuple);
+			  struct tuple *old_tuple, struct tuple *new_tuple,
+			  struct region *region);
```
locker, line 59: Nit: You forgot to update the comment to this function.

src/box/txn.h
```diff
@@ -1038,6 +1038,20 @@ box_txn_savepoint(void);
...
+ */
+struct region *
+tx_region_acquire(struct txn *txn);
+
+/**
+ * Method to return region to txn and account new allocations.
+ */
+void
+tx_region_release(struct txn *txn, enum tx_alloc_type alloc_type);
+
```
locker, line 1054: Nit: Please move it before the box API function declarations.

.test.mk
```diff
@@ -74,7 +77,8 @@ test-release-asan: CMAKE_PARAMS = -DCMAKE_BUILD_TYPE=RelWithDebInfo \
                                   -DENABLE_WERROR=ON \
                                   -DENABLE_ASAN=ON \
                                   -DENABLE_UB_SANITIZER=ON \
-                                  -DENABLE_FUZZER=ON
+                                  -DENABLE_FUZZER=ON \
+                                  -DABORT_ON_LEAK=ON
```
locker, line 81: We may add more options like this in future, in which case we'd like to avoid patching `.test.mk`. Let's introduce a new meta-option `TEST_BUILD`, which would incorporate all similar options (currently, it'd just enable `ABORT_ON_LEAK`).

CMakeLists.txt
```diff
@@ -738,6 +740,11 @@ endif()
 set(TARANTOOL_BUILD "${CMAKE_SYSTEM_NAME}-${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_BUILD_TYPE}")
 set(TARANTOOL_C_COMPILER ${CMAKE_C_COMPILER})
 set(TARANTOOL_CXX_COMPILER ${CMAKE_CXX_COMPILER})
+if(ABORT_ON_LEAK)
+    set(TARANTOOL_OPTIONS "${TARANTOOL_OPTIONS} -DABORT_ON_LEAK=TRUE")
+else()
+    set(TARANTOOL_OPTIONS "${TARANTOOL_OPTIONS} -DABORT_ON_LEAK=FALSE")
+endif()
```
locker, line 747: Not sure if we want to export it to `tarantool.build.options`. Why do you do that?

locker (changes requested):

src/box/request.h
nshy: Fixed.

src/box/txn.h
nshy: Done.

.test.mk
nshy: Done.

changelogs/unreleased/gh-5665-get-rid-of-fiber_gc.md
```diff
@@ -0,0 +1,7 @@
+## bugfix/core
+
+* **[Breaking change]** `box_region allocations` not paired with
```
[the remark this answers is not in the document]
nshy: Fixed.

locker (approved):

locker:
@nshy Some tests [fail](https://github.com/tarantool/tarantool/actions/runs/3515415504/jobs/5891300731) on debug-aarch64. Please investigate.

<details>

```
[037] not ok 1	gh-6310-grant-rw-access-on-_session_settings-space-to-public-role.test_boostrap
[037] #   Peer closed
[037] #   stack traceback:
[037] #   	builtin/box/net_box.lua:817: in function '_request'
[037] #   	builtin/box/net_box.lua:878: in function 'eval'
[037] #   	...ccess_on__session_settings_space_to_public_role_test.lua:22: in function 'gh-6310-grant-rw-access-on-_session_settings-space-to-public-role.test_boostrap'
[037] #   	...
[037] #   	[C]: in function 'xpcall'
[037] upgrade | Segmentation fault
[037] upgrade |   code: SEGV_MAPERR
[037] upgrade |   addr: 0xc4200000000
[037] upgrade |   context: 0xffffa897c9b0
[037] upgrade |   siginfo: 0xffffa897c930
[037] upgrade | Current time: 1669044311
[037] upgrade | Please file a bug at https://github.com/tarantool/tarantool/issues
[037] upgrade | Attempting backtrace... Note: since the server has already crashed, 
[037] upgrade | this may fail as well
[037] upgrade | #1  0xaaaab1bd70ac in crash_collect+196
[037] upgrade | #2  0xaaaab1bd8c88 in crash_signal_cb+92
[037] upgrade | #3  0xffffac11778c in ??+0
[037] upgrade | #4  0xaaaab1f5a1f8 in access_mem+32
[037] upgrade | #5  0xaaaab1f577cc in _ULaarch64_step+540
[037] upgrade | #6  0xaaaab1f56e04 in unw_backtrace+340
[037] upgrade | #7  0xaaaab1be8ce0 in collect_current_stack+32
[037] upgrade | #8  0xaaaab1be8ddc in backtrace_collect+228
[037] upgrade | #9  0xaaaab1be3fec in fiber_on_gc_alloc+144
[037] upgrade | #10 0xaaaab1e08f68 in region_alloc+192
[037] upgrade | #11 0xaaaab1e0e644 in tuple_field_map_create_plain+716
[037] upgrade | #12 0xaaaab1e0eb04 in tuple_field_map_create+288
[037] upgrade | #13 0xaaaab1e05ba4 in runtime_tuple_new+256
[037] upgrade | #14 0xaaaab1e03b38 in tuple_new+44
[037] upgrade | #15 0xaaaab1e07cd0 in box_tuple_new+36
[037] upgrade | #16 0xaaaab19bb1ac in session_settings_iterator_next+260
[037] upgrade | #17 0xaaaab196b74c in iterator_next+128
[037] upgrade | #18 0xaaaab1a8a654 in box_select+1688
[037] upgrade | #19 0xaaaab1c42b00 in lj_vm_ffi_call+100
[037] upgrade | #20 0xc4200000000 in ??+0
[037] upgrade | #21 0xc4200000000 in ??+0
```

</details>

(In order not to overload the CI, please don't set the 'full-ci' label to run the test on debug-aarch64. Instead, just enable the test run without 'full-ci' in the [workflow config](https://github.com/tarantool/tarantool/blob/master/.github/workflows/debug_aarch64.yml).)

src/lua/fiber.c
```diff
@@ -931,6 +951,8 @@ static const struct luaL_Reg fiberlib[] = {
 	{"parent_backtrace_enable", lbox_fiber_parent_backtrace_enable},
 	{"parent_backtrace_disable", lbox_fiber_parent_backtrace_disable},
 #endif /* ENABLE_BACKTRACE */
+	{"leak_backtrace_enable", lbox_fiber_leak_backtrace_enable},
+	{"leak_backtrace_disable", lbox_fiber_leak_backtrace_disable},
```
locker, line 955: Shouldn't we define these only if `ENABLE_LEAK_BACKTRACE` is set? Let's be consistent and define all the code dealing with leak backtraces (including the `fiber_leak_backtrace_enable` variable and the Lua API) under the macro.

nshy: Done.

PR #7905, opened 2022-11-07
box: export `box_schema_version` to public API and Lua

### LINKED ISSUE #7904: Export `box_schema_version` to public API and Lua
# Problem
`box_schema_version` symbol is currently exposed via `exports` and `box.internal.schema_version`, but it is not part of the public API.

One of the use cases is #7901: **IPROTO** request handlers need to provide the schema version in their replies (see also [`IPROTO_SCHEMA_VERSION`](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/keys/#iproto-schema-version) and example below).

# Solution

## Interface
Export the server's schema version to C and Lua.

### C
```C
uint32_t
box_schema_version(void);
```

### Lua
```lua
box.info.schema_version
-- 81
```

The returned number is the version of the database schema, an unsigned number that goes up when
there is a major change in the schema, i.e., on DDL operations (see also (see also [`IPROTO_SCHEMA_VERSION`](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/keys/#iproto-schema-version)).

## Example
Mimicking an **IPROTO** response (see #7897 for `box.iproto.send`):
```lua
box.iproto.send(sid, {
    request_type = box.iproto.OK,
    sync = 10, 
    schema_version = box.info.schema_version,
}, {
    data = 1,
})
```

CuriousGeorgiy:
`box_schema_version` symbol is currently exposed via `exports`, but it is not part of the public API: export it to public API and Lua.

Closes #7904

### DIFF
--- a/changelogs/unreleased/gh-7904-export-box_schema_version-to-public-api-and-lua.md
+++ b/changelogs/unreleased/gh-7904-export-box_schema_version-to-public-api-and-lua.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Exported `box_schema_version` to public API and Lua via
+  `box.info.schema_version` (gh-7904).
--- a/src/box/lua/feedback_daemon.lua
+++ b/src/box/lua/feedback_daemon.lua
@@ -212,7 +212,7 @@ local cached_schema_version = 0
 local cached_schema_features = {}
 
 local function fill_in_schema_stats(features)
-    local schema_version = box.internal.schema_version()
+    local schema_version = box.info.schema_version
     if cached_schema_version < schema_version then
         local schema = {}
         fill_in_schema_stats_impl(schema)
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -28,7 +28,6 @@ local INT64_MAX = tonumber64('9223372036854775807')
 ffi.cdef[[
     extern bool box_read_ffi_is_disabled;
     struct space *space_by_id(uint32_t id);
-    extern uint32_t box_schema_version();
     void space_run_triggers(struct space *space, bool yesno);
     size_t space_bsize(struct space *space);
 
@@ -1924,7 +1923,14 @@ local function check_ck_constraint_arg(ck_constraint, method)
 end
 box.internal.check_ck_constraint_arg = check_ck_constraint_arg
 
-box.internal.schema_version = builtin.box_schema_version
+local internal_schema_version_warn_once = false
+box.internal.schema_version = function()
+    if not internal_schema_version_warn_once then
+        internal_schema_version_warn_once = true
+        log.warn('box.internal.schema_version will be removed, please use box.info.schema_version instead')
+    end
+    return box.info.schema_version
+end
 
 local function check_iterator_type(opts, key_is_nil)
     local opts_type = type(opts)
--- a/src/box/schema.cc
+++ b/src/box/schema.cc
@@ -69,7 +69,7 @@ struct rlist on_alter_func = RLIST_HEAD_INITIALIZER(on_alter_func);
 
 struct entity_access entity_access;
 
-uint64_t
+API_EXPORT uint64_t
 box_schema_version(void)
 {
 	return schema_version;
--- a/src/box/schema.cc
+++ b/src/box/schema.cc
@@ -57,7 +57,7 @@
 static struct mh_i32ptr_t *sequences;
 /** Public change counter. On its update clients need to fetch
  *  new space data from the instance. */
-uint32_t schema_version = 0;
+uint64_t schema_version = 0;
 
 /** Persistent version of the schema, stored in _schema["version"]. */
 uint32_t dd_version_id = 0;
@@ -69,7 +69,7 @@ struct rlist on_alter_func = RLIST_HEAD_INITIALIZER(on_alter_func);
 
 struct entity_access entity_access;
 
-uint32_t
+uint64_t
 box_schema_version(void)
 {
 	return schema_version;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -211,6 +211,7 @@ set(api_headers
     ${PROJECT_SOURCE_DIR}/src/box/field_def.h
     ${PROJECT_SOURCE_DIR}/src/box/tuple_format.h
     ${PROJECT_SOURCE_DIR}/src/box/schema_def.h
+    ${PROJECT_SOURCE_DIR}/src/box/schema.h
     ${PROJECT_SOURCE_DIR}/src/box/box.h
     ${PROJECT_SOURCE_DIR}/src/box/index.h
     ${PROJECT_SOURCE_DIR}/src/box/iterator_type.h
--- a/src/box/errcode.h
+++ b/src/box/errcode.h
@@ -161,7 +161,7 @@ struct errcode_record {
 	/*106 */_(ER_WRONG_INDEX_RECORD,	"Wrong record in _index space: got {%s}, expected {%s}") \
 	/*107 */_(ER_WRONG_INDEX_PARTS,		"Wrong index parts: %s") \
 	/*108 */_(ER_WRONG_INDEX_OPTIONS,	"Wrong index options: %s") \
-	/*109 */_(ER_WRONG_SCHEMA_VERSION,	"Wrong schema version, current: %d, in request: %u") \
+	/*109 */_(ER_WRONG_SCHEMA_VERSION,	"Wrong schema version, current: %d, in request: %llu") \
 	/*110 */_(ER_MEMTX_MAX_TUPLE_SIZE,	"Failed to allocate %u bytes for tuple: tuple is too large. Check 'memtx_max_tuple_size' configuration option.") \
 	/*111 */_(ER_WRONG_SPACE_OPTIONS,	"Wrong space options: %s") \
 	/*112 */_(ER_UNSUPPORTED_INDEX_FEATURE,	"Index '%s' (%s) of space '%s' (%s) does not support %s") \
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1793,7 +1793,7 @@ net_finish_destroy(struct cmsg *m)
 
 
 static int
-tx_check_schema(uint32_t new_schema_version)
+tx_check_schema(uint64_t new_schema_version)
 {
 	if (new_schema_version && new_schema_version != schema_version) {
 		diag_set(ClientError, ER_WRONG_SCHEMA_VERSION,
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -51,6 +51,7 @@
 #include "box/box.h"
 #include "box/raft.h"
 #include "box/txn_limbo.h"
+#include "box/schema.h"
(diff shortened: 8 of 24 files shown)


test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
```diff
@@ -0,0 +1,26 @@
...
+        alias   = 'dflt',
+    }
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+--[[
```
locker, line 17: Nit: We don't use multi-line comments in Lua.

test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
```diff
@@ -0,0 +1,26 @@
...
+Checks that `box.schema_version` returns a valid number.
```
locker, line 18: Let's also check that it grows with DDL. Also, you should add a C module API test.

src/box/lua/feedback_daemon.lua
```diff
@@ -212,7 +212,7 @@ local cached_schema_version = 0
 local cached_schema_features = {}
 
 local function fill_in_schema_stats(features)
-    local schema_version = box.internal.schema_version()
+    local schema_version = box.schema_version()
```
locker, line 215: Come to think of it, we usually return information like this to Lua via `box.info`.

CuriousGeorgiy: Exported schema version to Lua via `box.info.schema_version`.

test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
CuriousGeorgiy: > Also, you should add a C module API test.

Added a check to `test/app-tap/module_api.test.lua`.

test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
CuriousGeorgiy: Sorry, I copy-paste the test skelet from other tests and forget about this — fixed.

test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
CuriousGeorgiy: TBH, I'm kind of lazy to add checks for when the returned schema version is supposed to be 0 or shouldn't change.

changelogs/unreleased/gh-7904-export-box_schema_version-to-public-api-and-lua.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
```
Totktonada, line 1: The documentation request holds several requirements:

> 0 is returned before server configuration (i.e., box.cfg) or during
recovery and the old value of the schema version is returned in
transactions (i.e., DDL operations in transactions do not change the schema
version). Rolled back DDL operations also do not change the schema version.

As I see, neither of them are tested.

test/box-luatest/gh_7904_export_box_schema_version_to_public_api_and_lua_test.lua
locker: > This is already checked in other tests

Then I don't see any point in this test. Better drop it.

> Added a check to test/app-tap/module_api.test.lua.

The test doesn't belong in `app-tap`, because `box_schema_version` need `box.cfg` to work. Anyway, I agree with @Totktonada that we should also check it more extensively - at least, check that it's incremented on DDL.

CuriousGeorgiy: > Then I don't see any point in this test. Better drop it.

Then what’s the point of testing the C API?

CuriousGeorgiy: > at least, check that it's incremented on DDL.

I already wrote (see comment above) that we have such checks, not sure what you mean here.

locker: If the Lua API is tested, it doesn't mean that the C API doesn't have to be. For example, we have the `key_def` module available both in Lua and in C and we test both (`box-tap/key_def.test.lua`, `app-tap/module_api.test.lua`).

CuriousGeorgiy: > Then I don't see any point in this test. Better drop it.

So then why shouldn't the **Lua** API be tested?

locker: > So then why shouldn't the **Lua** API be tested?

I didn't mean it shouldn't be tested. What I meant to say is that, since the Lua API is already tested by `box/schema_reload.test.lua` and your test doesn't improve on the existing test, there's no point in it.

CuriousGeorgiy: > The test doesn't belong in `app-tap`, because `box_schema_version` need `box.cfg` to work.

As far as I can see from the test, it already uses `box.cfg`:
https://github.com/tarantool/tarantool/blob/a62e14a0106379b9cfd5198d9826e8d84c917827/test/app-tap/module_api.test.lua#L491-L498

CuriousGeorgiy: > If the Lua API is tested, it doesn't mean that the C API doesn't have to be. For example, we have the `key_def` module available both in Lua and in C and we test both (`box-tap/key_def.test.lua`, `app-tap/module_api.test.lua`).

On the contrary, we have, box API functions, such as `box_txn_set_timeout` or `box_txn_set_isolation` that are not tested.

CuriousGeorgiy: > I didn't mean it shouldn't be tested. What I meant to say is that, since the Lua API is already tested by `box/schema_reload.test.lua` and your test doesn't improve on the existing test, there's no point in it.

So what value would test of the C API add except for checking that this function is indeed exported?

locker: > As far as I can see from the test, it already uses `box.cfg`

Too bad, but then I guess it's okay to add one more test case in there that depends on `box.cfg`.

CuriousGeorgiy: > Too bad, but then I guess it's okay to add one more test case in there that depends on box.cfg.

Okay, did just so.

changelogs/unreleased/gh-7904-export-box_schema_version-to-public-api-and-lua.md
CuriousGeorgiy: Discussed with @Totktonada, @alyapunov, @unera: the this patch does not intend to change the behaviour of schema version number, so it remains unspecified with its monotonic growth being the only requirement held.

src/box/lua/schema.lua
```diff
@@ -1924,8 +1924,6 @@ local function check_ck_constraint_arg(ck_constraint, method)
 end
 box.internal.check_ck_constraint_arg = check_ck_constraint_arg
 
-box.internal.schema_version = builtin.box_schema_version
```
locker, line 1927: Nit: You should also remove `box_schema_version` definition from the `ffi.cdef` above.

src/box/schema.cc
```diff
@@ -69,7 +69,7 @@ struct rlist on_alter_func = RLIST_HEAD_INITIALIZER(on_alter_func);
 
 struct entity_access entity_access;
 
-uint32_t
+API_EXPORT uint32_t
```
locker, line 72: The documentation says that it only goes up while in fact it may wrap around. This is probably not critical, because a client is supposed to update the schema as soon as it changes, but perhaps we should turn it in `uint64_t` while we are at it to make its semantics clearer? We won't be able to do it in future, because it'd break the ABI.

src/box/lua/schema.lua
CuriousGeorgiy: Yeah, forgot about it, removed, thanks!

changelogs/unreleased/gh-7904-export-box_schema_version-to-public-api-and-lua.md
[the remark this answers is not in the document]
CuriousGeorgiy: I thought of it from the point that the symbol is exposed and usable, but is not part of the public API.

src/box/schema.cc
CuriousGeorgiy: I agree: added a commit changing the type of schema version and everything dependening on it to `uint64_t`.

CuriousGeorgiy:
Ignoring linter error for `src/box/sql/vdbeInt.h`, since fixing it would break the existing comment style of this file.

locker (approved):

CuriousGeorgiy:
@locker, @alyapunov: are we sure we need a separate function? Why don't we just export `schema_version`?

CuriousGeorgiy:
@alyapunov added your comments to the documentation request.

alyapunov (approved):

locker (approved):

locker:
@CuriousGeorgiy Please don't forget to update all our projects that use `box.internal.schema_version()`.
I [found](https://github.com/search?l=Lua&q=org%3Atarantool+box.internal.schema_version&type=code) four such projects:
- https://github.com/tarantool/tarantool-ee
- https://github.com/tarantool/vshard
- https://github.com/tarantool/document
- https://github.com/tarantool/ddl

Gerold103:
TBH I don't quite like how `box.info` is getting bloated. It is truly huge already. The schema version, how I use it in vshard, helps me to preserve a cache of spaces of some sort. I rebuild it when the schema version changes. With `box.info` growing more and more, eventually getting the schema version will become more expensive than each time collecting the spaces that I want.

It gets even further worse when some of our `box.info` members start yielding. For example:
```
fiber = require('fiber')
do
    c1 = fiber.self().csw()
    box.info.memory()
    c2 = fiber.self().csw()
end
```
Here `c2 > c1`. 

For now `box.info` itself does not seem to yield, but when it will, it will be unsafe to use in contexts like transactions. I would suggest to do something with it. Maybe not put more stuff into this table.

Totktonada:
@Gerold103 `box.info.schema_version` only gets the schema version, nothing more. It works through `__index` and `__serialize` metamethods (and `__call` for `box.info()`). So a cost of `box.info.schema_version` will be the same disregarding of amount of `box.info` entries.

Gerold103:
Just checked - indeed! Didn't know that. So apparently if I do `i = box.info` then it is actually obtained when I try to access the members of `i` (fine by me).

PR #7920, opened 2022-11-10
ci: add 'self-hosted' tag to runs-on section

igormunkin:
Fun fact: our self-hosted macOS runner has the same name as the one provided by GitHub. Hence sometimes when no self-hosted runners are available, the public GitHub one is chosen.

This patch enforces LuaJIT integration workflow to use only self-hosted runner by explicitly specifying this in runs-on section.

NO_DOC=ci
NO_TEST=ci
NO_CHANGELOG=ci

Signed-off-by: Igor Munkin <imun@tarantool.org>


.github/workflows/luajit-integration.yml
```diff
@@ -45,7 +45,7 @@ on:
 
 jobs:
   luajit-integration:
-    runs-on: ${{ inputs.host }}
+    runs-on: [self-hosted, '${{ inputs.host }}']
```
ylobankov, line 48: In such cases, we usually do so (add a space after `[` and before `]`):

```yaml
runs-on: [ self-hosted, '${{ inputs.host }}' ]
```

ylobankov (approved):
LGTM

.github/workflows/luajit-integration.yml
igormunkin: Fixed.

igormunkin:

> LGTM

Added your tag:

PR #7935, opened 2022-11-14
flaky: fix flakiness of the gh_7253 test

sergos:
It appears that we have to wait for commit completion on master before start waiting for it on replica.

NO_DOC=test fix
NO_CHANGELOG=test fix


sergepetrenko (approved):
Thanks for noticing and fixing this!
LGTM.

NickVolynkin (approved):

Issue #5272, opened 2020-09-04
Deprecate replication_connect_quorum


Gerold103:
The option is useless after bootstrap. All it can do is break a working instance, after it is already booted. People constantly have problems with this option.

The only purpose when it is useful is the cluster bootstrap time, when need to wait for at least specified number of replicas before creating replicaset UUID and registering all of them in `_cluster` on a selected leader. After bootstrap the option is useless. All installations set it to 0 as soon as possible.

Another reason why it should not exist is a conflict with synchronous replication options - there is `replication_synchro_quorum`. 

The option can be renamed to `replication_boot_quorum`, used at boot time like now, and then ignored in all subsequent instance restarts.

sergepetrenko:
To be done in scope of this ticket:
Introduce new `box.cfg` option: `bootstrap_strategy`, which receives 2 possible values `"auto"` -- default, and `"legacy"` -- compatibility with current `replication_connect_quorum` behaviour.

`bootstrap_strategy = "config"` and `bootstrap_strategy = "supervised"` will be done in scope of #7999.

Issue #3159, opened 2018-02-15
box.ctl._on_ctl_event


kostja:
Implement a trigger invoked on any control event:
- completion of local recovery -> should be used to set BEFORE REPLACE trigger
- switch to read only mode
- switch to read write mode
- changes in replica set

kostja:
See also https://github.com/tarantool/tarantool/issues/1607

opomuc:
Can you add `replica connection error/loss` event as well? (as not in a controlled shutdown, but rather unexpected situation)

GeorgyKirichenko:
I think there should be a bunch of events:

- system space creation event; it allows us to set a replace trigger on _space and then do anything what we want with user spaces (tracking a system space recovery)
- snapshot and xlog recovery events
- applier state event
- relay state event
- rw mode event
- exit event


replicaset changes might be tracked with _cluster replace trigger

PR #7938, opened 2022-11-15
Triggers on recovery state

sergepetrenko:
Introduce triggers which are run during various stages of box.cfg{}.
See the docbot request below for more info.

Part-of #5272
Part-of #3159

@TarantoolBot document
Title: new triggers -- `box.ctl.on_recovery_state`

`box.ctl.on_recovery_state` is a handle to register triggers that will
be executed on different stages of recovery/initial configuration.
A trigger registered this way is run on each of the supported recovery
stages, and receives the stage name as a parameter. The possible stages
are:
 - "snapshot_recovered": the node has recovered the snapshot file. This
   is the fist state seen by triggers.
 - "wal_recovered": the node has recovered all the .xlog files.
 - "indexes_built": the node has built secondary indexes for memtx spaces.
   Depending on a number of box.cfg options, "indexes_built" might come
   before or after "wal_recovered".
 - "synced": the node has synced with enough remote peers and has left
   orphan state. This is the last state seen by the triggers.

All these stages are passed during initial `box.cfg` call when recovering
from .snap and .xlog files.

When bootstrapping a fresh cluster with no data, all the instances in it
will still execute triggers on the same stages for consistency.
For example, "snapshot_recovered" and "wal_recovered" will
be run once the node finishes cluster bootstrap or finishes joining to an
existing cluster.

Anyway, by the time the initial `box.cfg` call finishes, regardless of
recovery or bootstrap, the triggers will be called with parameters
"snapshot_recovery", "wal_recovered" and "indexes_built". The triggers
on "synced" state might execute later, when the node leaves the orphan
mode.

Note, that you must set `box.ctl.on_recovery_state` triggers before the
initial `box.cfg` call for them to take effect.
Example:

NO_WRAP
```
Tarantool 2.11.0
---
...

tarantool> trigger_f = function(state) log.info('State "' .. state .. '" reached') end
---
...

tarantool> box.ctl.on_recovery_state(trigger_f)
---
- 'function: 0x0104662788'
...

tarantool> box.cfg{}
<stripped some messages>
2022-11-15 13:21:43.149 [98398] main/103/interactive I> recovering from `./00000000000000000004.snap'
2022-11-15 13:21:43.166 [98398] main/103/interactive I> cluster uuid 75c5c40e-c795-4c6f-b602-25fe5b1c9f3b
2022-11-15 13:21:43.187 [98398] main/103/interactive I> assigned id 1 to replica 1fbcb1d7-f600-4cc6-83b4-8d409d2937a9
2022-11-15 13:21:43.187 [98398] main/103/interactive I> update replication_synchro_quorum = 1
2022-11-15 13:21:43.187 [98398] main/103/interactive I> State "snapshot_recovered" reached
                                                               ^^^^^^^^^^^^^^^^^^
2022-11-15 13:21:43.187 [98398] main/103/interactive I> recover from `./00000000000000000004.xlog'
2022-11-15 13:21:43.188 [98398] main/103/interactive I> done `./00000000000000000004.xlog'
2022-11-15 13:21:43.188 [98398] main/103/interactive I> recover from `./00000000000000000006.xlog'
2022-11-15 13:21:43.188 [98398] main/103/interactive I> done `./00000000000000000006.xlog'
2022-11-15 13:21:43.188 [98398] main/103/interactive I> Building secondary indexes in space 'test'...
2022-11-15 13:21:43.188 [98398] main/103/interactive I> Adding 2 keys to TREE index 'sk' ...
2022-11-15 13:21:43.188 [98398] main/103/interactive I> Space 'test': done
2022-11-15 13:21:43.188 [98398] main/103/interactive I> State "indexes_built" reached
                                                               ^^^^^^^^^^^^^
2022-11-15 13:21:43.189 [98398] main/103/interactive I> State "wal_recovered" reached
                                                               ^^^^^^^^^^^^^
2022-11-15 13:21:43.189 [98398] main/103/interactive I> ready to accept requests
2022-11-15 13:21:43.189 [98398] main/103/interactive I> State "synced" reached
                                                               ^^^^^^
2022-11-15 13:21:43.189 [98398] main/103/interactive I> leaving orphan mode
<stripped some messages>
---
...

tarantool>

```
NO_WRAP

### DIFF
--- a/changelogs/unreleased/gh-3159-recovery-triggers.md
+++ b/changelogs/unreleased/gh-3159-recovery-triggers.md
@@ -1,0 +1,7 @@
+## feature/box
+
+* It is now possible to register triggers for various recovery stages: use
+  `box.ctl.on_recovery_state()` before initial `box.cfg()` call. The trigger
+  receives one parameter - a string resembling reached recovery stage. Possible
+  values are "snapshot_recovered", "wal_recovered", "indexes_built", "synced"
+  (gh-3159).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -166,6 +166,62 @@ static struct fiber_pool tx_fiber_pool;
  */
 static struct cbus_endpoint tx_prio_endpoint;
 
+RLIST_HEAD(box_on_recovery_state);
+
+/**
+ * Recovery states supported by on_recovery_state triggers.
+ * Are positioned in the order of appearance during initial box.cfg().
+ * The only exception are WAL_RECOVERED and INDEXES_BUILT, which might come in
+ * any order, since the moment when secondary indexes are built depends on
+ * box.cfg.force_recovery and box.cfg.memtx_use_mvcc_engine.
+ */
+enum box_recovery_state {
+	/**
+	 * The node has either recovered the snapshot from the disk, received
+	 * the snapshot from the remote master as part of initial join stage or
+	 * has bootstrapped the cluster.
+	 */
+	RECOVERY_STATE_SNAPSHOT_RECOVERED,
+	/**
+	 * The node has either recovered the local WAL files, received the WALs
+	 * from the remote master as part of final join or has bootstrapped the
+	 * cluster.
+	 */
+	RECOVERY_STATE_WAL_RECOVERED,
+	/**
+	 * The node has built secondary indexes for memtx spaces.
+	 */
+	RECOVERY_STATE_INDEXES_BUILT,
+	/**
+	 * The node has synced with remote peers. IOW, it has transitioned from
+	 * "orphan" to "running".
+	 */
+	RECOVERY_STATE_SYNCED,
+	box_recovery_state_MAX,
+};
+
+static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
+	/* [RECOVERY_STATE_SNAPSHOT_RECOVERED] = */
+	"snapshot_recovered",
+	/* [RECOVERY_STATE_WAL_RECOVERED] = */
+	"wal_recovered",
+	/* [RECOVERY_STATE_INDEXES_BUILT] = */
+	"indexes_built",
+	/* [RECOVERY_STATE_SYNCED] = */
+	"synced",
+};
+
+/** Whether the triggers for "synced" recovery stage have already run. */
+static bool recovery_state_synced_is_reached;
+
+static int
+box_run_on_recovery_state(enum box_recovery_state state)
+{
+	assert(state >= 0 && state < box_recovery_state_MAX);
+	return trigger_run(&box_on_recovery_state,
+			   (char *)box_recovery_state_strs[state]);
+}
+
 static void
 builtin_events_init(void);
 
@@ -407,6 +463,10 @@ box_do_set_orphan(bool orphan)
 {
 	is_orphan = orphan;
 	box_update_ro_summary();
+	if (!is_orphan && !recovery_state_synced_is_reached) {
+		box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
+		recovery_state_synced_is_reached = true;
+	}
 }
 
 void
@@ -3679,6 +3739,13 @@ box_free(void)
 		sql_built_in_functions_cache_free();
 		port_free();
 	}
+	trigger_destroy(&box_on_recovery_state);
+}
+
+static void
+box_on_indexes_built(void)
+{
+	box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
 }
 
 static void
@@ -3698,7 +3765,8 @@ engine_init()
 				    cfg_geti("strip_core"),
 				    cfg_geti("slab_alloc_granularity"),
 				    cfg_gets("memtx_allocator"),
-				    cfg_getd("slab_alloc_factor"));
+				    cfg_getd("slab_alloc_factor"),
+				    box_on_indexes_built);
 	engine_register((struct engine *)memtx);
 	box_set_memtx_max_tuple_size();
 
@@ -3767,6 +3835,12 @@ bootstrap_master(const struct tt_uuid *replicaset_uuid)
 	/* Make the initial checkpoint */
 	if (gc_checkpoint() != 0)
 		panic("failed to create a checkpoint");
+
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
+	box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
+	assert(!recovery_state_synced_is_reached);
+	box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
+	recovery_state_synced_is_reached = true;
 }
 
 /**
@@ -3816,6 +3890,8 @@ bootstrap_from_master(struct replica *master)
 					APPLIER_FINAL_JOIN;
 	applier_resume_to_state(applier, wait_state, TIMEOUT_INFINITY);
 
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
+
 	/*
 	 * Process final data (WALs).
 	 */
@@ -3849,6 +3925,8 @@ bootstrap_from_master(struct replica *master)
 	/* Make the initial checkpoint */
 	if (gc_checkpoint() != 0)
 		panic("failed to create a checkpoint");
+
+	box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
 
 	return true;
 }
@@ -4046,6 +4124,8 @@ local_recovery(const struct tt_uuid *instance_uuid,
 	 */
 	memtx_engine_recover_snapshot_xc(memtx, checkpoint_vclock);
 
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
+
 	engine_begin_final_recovery_xc();
 	recover_remaining_wals(recovery, &wal_stream.base, NULL, false);
 	if (wal_stream_has_tx(&wal_stream)) {
@@ -4148,6 +4228,8 @@ local_recovery(const struct tt_uuid *instance_uuid,
 			  tt_uuid_str(replicaset_uuid),
 			  tt_uuid_str(&REPLICASET_UUID));
 	}
+
+	box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
 }
 
 static void
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -71,6 +71,11 @@ extern double on_shutdown_trigger_timeout;
 extern struct rlist box_on_shutdown_trigger_list;
 
 /**
+ * Triggers invoked during initial `box.cfg` call on various recovery stages.
+ */
+extern struct rlist box_on_recovery_state;
+
+/**
  * Timeout during which the transaction must complete,
  * otherwise it will be rolled back.
  */
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -428,6 +428,7 @@ memtx_engine_begin_final_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
+		memtx->on_indexes_built_cb();
 	}
 	return 0;
 }
@@ -446,6 +447,7 @@ memtx_engine_begin_hot_standby(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
+		memtx->on_indexes_built_cb();
 	}
 	return 0;
 }
@@ -465,6 +467,7 @@ memtx_engine_end_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
+		memtx->on_indexes_built_cb();
 	}
 	xdir_collect_inprogress(&memtx->snap_dir);
 
@@ -632,7 +635,10 @@ memtx_engine_bootstrap(struct engine *engine)
 			break;
 	}
 	xlog_cursor_close(&cursor, false);
-	return rc < 0 ? -1 : 0;
+	if (rc < 0)
+		return -1;
+	memtx->on_indexes_built_cb();
+	return 0;
 }
 
 static int
@@ -1256,7 +1262,8 @@ struct memtx_engine *
 memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		 uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		 bool dontdump, unsigned granularity,
-		 const char *allocator, float alloc_factor)
+		 const char *allocator, float alloc_factor,
+		 memtx_on_indexes_built_cb on_indexes_built)
 {
 	int64_t snap_signature;
 	struct memtx_engine *memtx =
@@ -1362,6 +1369,8 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		panic("failed to create functional index key format");
 	}
 	tuple_format_ref(memtx->func_key_format);
+
+	memtx->on_indexes_built_cb = on_indexes_built;
 
 	fiber_start(memtx->gc_fiber, memtx);
 	return memtx;
--- a/src/box/memtx_engine.h
+++ b/src/box/memtx_engine.h
@@ -109,6 +109,9 @@ enum memtx_reserve_extents_num {
  */
 #define MEMTX_ITERATOR_SIZE (168)
 
+typedef void
+(*memtx_on_indexes_built_cb)(void);
+
 struct memtx_engine {
 	struct engine base;
 	/** Engine recovery state, see enum memtx_recovery_state description. */
@@ -126,6 +129,11 @@ struct memtx_engine {
 	 * needed to be able to cancel it on shutdown.
 	 */
 	struct cord *replica_join_cord;
+	/**
+	 * A callback run once memtx engine builds secondary indexes for the
+	 * data.
+	 */
+	memtx_on_indexes_built_cb on_indexes_built_cb;
 	/** Common quota for tuples and indexes. */
 	struct quota quota;
 	/**
@@ -209,7 +217,8 @@ struct memtx_engine *
 memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		 uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		 bool dontdump, unsigned granularity,
-		 const char *allocator, float alloc_factor);
+		 const char *allocator, float alloc_factor,
+		 memtx_on_indexes_built_cb on_indexes_built);
 
 int
 memtx_engine_recover_snapshot(struct memtx_engine *memtx,
@@ -341,13 +350,14 @@ static inline struct memtx_engine *
 memtx_engine_new_xc(const char *snap_dirname, bool force_recovery,
 		    uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		    bool dontdump, unsigned granularity,
-		    const char *allocator, float alloc_factor)
+		    const char *allocator, float alloc_factor,
+		    memtx_on_indexes_built_cb on_indexes_built)
 {
 	struct memtx_engine *memtx;
 	memtx = memtx_engine_new(snap_dirname, force_recovery,
-				 tuple_arena_max_size,
-				 objsize_min, dontdump,
-				 granularity, allocator, alloc_factor);
+				 tuple_arena_max_size, objsize_min, dontdump,
+				 granularity, allocator, alloc_factor,
+				 on_indexes_built);
 	if (memtx == NULL)
 		diag_raise();
 	return memtx;
--- a/test/replication-luatest/recovery_triggers_test.lua
+++ b/test/replication-luatest/recovery_triggers_test.lua
@@ -1,0 +1,111 @@
+local t = require('luatest')
+local cluster = require('test.luatest_helpers.cluster')
+local server = require('test.luatest_helpers.server')
+
+local g = t.group('gh-5272')
+
+local run_before_cfg = [[
+    rawset(_G, 'test_states', {})
+    box.ctl.on_recovery_state(function(state)
+        table.insert(test_states, state)
+    end)
+]]
+
+g.before_all(function(cg)
+    cg.cluster = cluster:new({})
+    cg.server = cg.cluster:build_and_add_server({
+        alias = 'server',
+        env = {
+            ['TARANTOOL_RUN_BEFORE_CFG'] = run_before_cfg,
+        },
+    })
+    cg.replica = cg.cluster:build_and_add_server({
+        alias = 'replica',
+        box_cfg = {
+            replication = server.build_instance_uri('server'),
+        },
+        env = {
+            ['TARANTOOL_RUN_BEFORE_CFG'] = run_before_cfg,
+        },
+    })
+end)
+
+g.after_all(function(cg)
+    cg.cluster:drop()
+end)
+
+local function is_state_reached(server, state)
+    return server:exec(function(state)
+        for _, s in pairs(_G.test_states) do
+            if s == state then
+                return true
+            end
+        end
+        return false
+    end, {state})
+end
+
+local function wait_state(server, state)
+    t.helpers.retrying({}, function()
+        t.assert(is_state_reached(server, state),
+                 'State ' .. state .. ' triggers fired')
+    end)
+end
+
+local function assert_state_not_reached(server, state)
+    t.assert(not is_state_reached(server, state),
+             'State ' .. state .. ' not reached')
+end
+
+local function wait_all_states(server)
+    wait_state(server, 'snapshot_recovered')
+    -- wal_recovered and indexes_built can come in any order.
+    wait_state(server, 'wal_recovered')
+    wait_state(server, 'indexes_built')
+    wait_state(server, 'synced')
+end
+
+g.test_recovery_stages = function(cg)
+    -- Test stages on bootstrap.
+    cg.server:start()
+    wait_all_states(cg.server)
+    cg.server:stop()
+    -- Test same stages on recovery.
+    cg.server:start()
+    wait_all_states(cg.server)
+    -- Test the stages on replica join.
+    cg.replica:start()
+    wait_all_states(cg.replica)
+    cg.replica:stop()
+    -- Test sync state.
+    cg.replica.box_cfg = {
+        replication_sync_timeout = 0.01,
+        replication_connect_timeout = 0.1,
+        replication_timeout = 0.1,
+        replication = {
+            server.build_instance_uri('server'),
+            server.build_instance_uri('non-existent-uri'),
+        },
+    }
+    cg.replica:start()
+    wait_state(cg.replica, 'snapshot_recovered')
+    wait_state(cg.replica, 'wal_recovered')
+    wait_state(cg.replica, 'indexes_built')
+    assert_state_not_reached(cg.replica, 'synced')
+    cg.replica:exec(function()
+        box.cfg{
+            replication_connect_quorum = 1,
+        }
+    end)
+    wait_state(cg.replica, 'synced')
+    -- Check that triggers on 'synced' stage are executed only once.
+    cg.replica:exec(function()
+        local t = require('luatest')
+        local num_events = #_G.test_states
+        local old_repl = box.cfg.replication
+        box.cfg{replication = ''}
+        box.cfg{replication = old_repl}
+        t.assert_equals(#_G.test_states, num_events,
+                        "Triggers on sync aren't re-run")
+    end)
+end
--- a/src/box/lua/ctl.c
+++ b/src/box/lua/ctl.c
@@ -83,6 +83,19 @@ lbox_ctl_on_schema_init(struct lua_State *L)
 }
 
 static int
+lbox_push_recovery_state(struct lua_State *L, void *event)
+{
+	lua_pushstring(L, (const char *)event);
+	return 1;
+}
+
+static int
+lbox_ctl_on_recovery_state(struct lua_State *L)
+{
+	return lbox_trigger_reset(L, 2, &box_on_recovery_state,
+				  lbox_push_recovery_state, NULL);
+}
+static int
 lbox_ctl_on_election(struct lua_State *L)
 {
 	return lbox_trigger_reset(L, 2, &box_raft_on_broadcast, NULL, NULL);
@@ -139,6 +152,7 @@ static const struct luaL_Reg lbox_ctl_lib[] = {
 	{"wait_rw", lbox_ctl_wait_rw},
 	{"on_shutdown", lbox_ctl_on_shutdown},
 	{"on_schema_init", lbox_ctl_on_schema_init},
+	{"on_recovery_state", lbox_ctl_on_recovery_state},
 	{"on_election", lbox_ctl_on_election},
 	{"promote", lbox_ctl_promote},
 	/* An old alias. */
--- a/test/instances/default.lua
+++ b/test/instances/default.lua
@@ -38,6 +38,12 @@ end
 -- such bad tests hang and fail.
 box.ctl.set_on_shutdown_timeout(TIMEOUT_INFINITY)
 
+local exec_arg = os.getenv('TARANTOOL_RUN_BEFORE_CFG')
+if exec_arg then
+    local run_before_cfg = loadstring(exec_arg)
+    run_before_cfg()
+end
+
 box.cfg(box_cfg())
 box.schema.user.grant('guest', 'super', nil, nil, {if_not_exists = true})
 
--- a/test/luatest_helpers/server.lua
+++ b/test/luatest_helpers/server.lua
@@ -321,7 +321,8 @@ function Server:grep_log(what, bytes, opts)
                     line = table.concat(buf)
                     buf = nil
                 end
-                if string.match(line, "Starting instance") and not noreset then
+                if string.match(line, "Tarantool %d+.%d+.%d+-.*%d+-g.*") and
+                   not noreset then
                     found = nil -- server was restarted, reset search
                 else
                     found = string.match(line, what) or found


src/box/box.cc
```diff
@@ -412,6 +415,8 @@ box_do_set_orphan(bool orphan)
 {
 	is_orphan = orphan;
 	box_update_ro_summary();
+	if (!is_orphan)
+		BOX_ON_RECOVERY_STATE("synced");
```
Gerold103, line 419: AFAIR the instance can become orphan again on `box.cfg` reconfig with new replication parameter. Is it true? If so, is it left out intentionally?

src/box/box_trigger_def.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
Gerold103, line 1: I would suggest to merge this entire file into box.h and turn `BOX_ON_RECOVERY_STATE` into a function. Separation into another file looks only inconsistent with other box-stuff. For example, `box_on_shutdown_trigger_list` is in box.h just fine. Same about the macro. It is not something perf-critical which needs to be inlined or macrosed.

If you made it a macro because of copying the strings, it should be fine to copy them on stack like:
```
const int max_size = 64;
char buf[max_size];
int len = strlen(state);
assert(len < max_size);
memcpy(buf, state, len + 1);
// run_triggers(buf);
```
Anyway we know all those strings and their max len.

test/replication-luatest/gh_5272_recovery_triggers_test.lua
```diff
@@ -0,0 +1,85 @@
+local t = require('luatest')
```
Gerold103, line 1: AFAIK we don't use `gh_####` prefix for tests about features. That prefix is only for bugfixes, no?

test/replication-luatest/gh_5272_recovery_triggers_test.lua
```diff
@@ -0,0 +1,85 @@
...
+local cluster = require('test.luatest_helpers.cluster')
+local server = require('test.luatest_helpers.server')
+
+local g = t.group('gh-5272')
+
+local run_before_cfg = [[
+    box.ctl.on_recovery_state(function(state)
+        require('log').info('State ' .. state .. ' reached')
+    end)
```
Gerold103, line 10: It might be easier to inspect the states not via logs but via a Lua table, like that:
```Diff
@@ -5,8 +5,9 @@ local server = require('test.luatest_helpers.server')
 local g = t.group('gh-5272')
 
 local run_before_cfg = [[
+    rawset(_G, 'test_states', {})
     box.ctl.on_recovery_state(function(state)
-        require('log').info('State ' .. state .. ' reached')
+        table.insert(test_states, state)
     end)
 ]]
```
Then in the test you can check `test_states` array without reading any logs. Up to you.

Gerold103 (changes requested):
Thanks for the patch! I will leave the state naming judgement to the products. Otherwise looks almost ready!

src/box/box.cc
sergepetrenko: > AFAIR the instance can become orphan again on box.cfg reconfig with new replication parameter

True.

> If so, is it left out intentionally?

What's left out? Once the instance syncs with new replication parameter, `'synced'` triggers will run once again.
Or do you mean there are no triggers on becoming an orphan?

src/box/box_trigger_def.h
sergepetrenko: > I would suggest to merge this entire file into box.h

That's what I did initially, but it lead to introducing box.h dependency to memtx_engine.c
I wanted to avoid that. But I'll merge everything into box.h if you want me to.

> If you made it a macro because of copying the strings, it should be fine to copy them on stack like:

Yep, I didn't want to make the copy, since the string's copied when passed to lua anyway. This would leave us with two copies of the same string.
Anyway, I'm fine with turning this to a function:

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 4a3339727..66f8c49a3 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -416,7 +416,7 @@ box_do_set_orphan(bool orphan)
 	is_orphan = orphan;
 	box_update_ro_summary();
 	if (!is_orphan)
-		BOX_ON_RECOVERY_STATE("synced");
+		box_run_on_recovery_state("synced");
 }
 
 void
@@ -3772,7 +3772,7 @@ bootstrap_master(const struct tt_uuid *replicaset_uuid)
 	if (gc_checkpoint() != 0)
 		panic("failed to create a checkpoint");
 
-	BOX_ON_RECOVERY_STATE("snapshot_recovered");
+	box_run_on_recovery_state("snapshot_recovered");
 }
 
 /**
@@ -3822,7 +3822,7 @@ bootstrap_from_master(struct replica *master)
 					APPLIER_FINAL_JOIN;
 	applier_resume_to_state(applier, wait_state, TIMEOUT_INFINITY);
 
-	BOX_ON_RECOVERY_STATE("snapshot_recovered");
+	box_run_on_recovery_state("snapshot_recovered");
 
 	/*
 	 * Process final data (WALs).
@@ -4054,7 +4054,7 @@ local_recovery(const struct tt_uuid *instance_uuid,
 	 */
 	memtx_engine_recover_snapshot_xc(memtx, checkpoint_vclock);
 
-	BOX_ON_RECOVERY_STATE("snapshot_recovered");
+	box_run_on_recovery_state("snapshot_recovered");
 
 	engine_begin_final_recovery_xc();
 	recover_remaining_wals(recovery, &wal_stream.base, NULL, false);
@@ -4358,7 +4358,7 @@ box_cfg_xc(void)
 
 	rmean_cleanup(rmean_box);
 
-	BOX_ON_RECOVERY_STATE("wal_recovered");
+	box_run_on_recovery_state("wal_recovered");
 
 	/* Follow replica */
 	replicaset_follow();
@@ -4399,8 +4399,8 @@ box_cfg_xc(void)
 	say_info("ready to accept requests");
 
 	if (is_bootstrap_leader) {
-		BOX_ON_RECOVERY_STATE("indexes_built");
-		BOX_ON_RECOVERY_STATE("synced");
+		box_run_on_recovery_state("indexes_built");
+		box_run_on_recovery_state("synced");
 	}
 	if (!is_bootstrap_leader) {
 		replicaset_sync();
diff --git a/src/box/box_trigger_def.h b/src/box/box_trigger_def.h
index 558890cf1..8bd90ad00 100644
--- a/src/box/box_trigger_def.h
+++ b/src/box/box_trigger_def.h
@@ -12,20 +12,24 @@
  */
 extern struct rlist box_on_recovery_state;
 
-/*
- * The macro is intended to simplify trigger invocation.
- *
- * on_recovery_state triggers expect a string, but passing a string literal to
- * them turns out quite ugly, because trigger_run takes void *. So one ends up
- * with trigger_run(trig, (char *)"string_literal");
- *
- * Compound literals solve this problem in a less ugly way, besides they really
- * can be written to: trigger_run(trig, (char[]){"string_literal"}). But
- * unfortunately they aren't supported in c++.
- *
- * So we are left with this:
- */
-#define BOX_ON_RECOVERY_STATE(str) ({\
-	char arr[] = str;\
-	trigger_run(&box_on_recovery_state, arr);\
-})
+/** Run box_on_recovery_state triggers with state represented by \a str. */
+static inline int
+box_run_on_recovery_state(const char *str)
+{
+	/*
+	 * on_recovery_state triggers expect a string, but passing a string
+	 * literal to them turns out quite ugly, because trigger_run takes
+	 * void *. So one ends up with
+	 * trigger_run(trig, (char *)"string_literal");
+	 *
+	 * Compound literals solve this problem in a less ugly way, besides they
+	 * can really be written to:
+	 * trigger_run(trig, (char[]){"string_literal"}).
+	 * But unfortunately they aren't supported in c++.
+	 * So we are left with this:
+	 */
+	char buf[64];
+	size_t len = strlcpy(buf, str, sizeof(buf));
+	assert(len < sizeof(buf));
+	return trigger_run(&box_on_recovery_state, buf);
+}
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 554094fcf..4049e3848 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -437,7 +437,7 @@ memtx_engine_begin_final_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		BOX_ON_RECOVERY_STATE("indexes_built");
+		box_run_on_recovery_state("indexes_built");
 	}
 	return 0;
 }
@@ -456,7 +456,7 @@ memtx_engine_begin_hot_standby(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		BOX_ON_RECOVERY_STATE("indexes_built");
+		box_run_on_recovery_state("indexes_built");
 	}
 	return 0;
 }
@@ -476,7 +476,7 @@ memtx_engine_end_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		BOX_ON_RECOVERY_STATE("indexes_built");
+		box_run_on_recovery_state("indexes_built");
 	}
 	xdir_collect_inprogress(&memtx->snap_dir);

```

</details>

test/replication-luatest/gh_5272_recovery_triggers_test.lua
sergepetrenko: Yep, you're right. Renamed.

test/replication-luatest/gh_5272_recovery_triggers_test.lua
sergepetrenko: Good point, done:
```diff
diff --git a/test/replication-luatest/recovery_triggers_test.lua b/test/replication-luatest/recovery_triggers_test.lua
index dc0c88574..d15d00816 100644
--- a/test/replication-luatest/recovery_triggers_test.lua
+++ b/test/replication-luatest/recovery_triggers_test.lua
@@ -5,8 +5,9 @@ local server = require('test.luatest_helpers.server')
 local g = t.group('gh-5272')
 
 local run_before_cfg = [[
+    rawset(_G, 'test_states', {})
     box.ctl.on_recovery_state(function(state)
-        require('log').info('State ' .. state .. ' reached')
+        table.insert(test_states, state)
     end)
 ]]
 
@@ -29,16 +30,27 @@ g.after_all(function(cg)
     cg.cluster:drop()
 end)
 
+local function is_state_reached(server, state)
+    return server:exec(function(state)
+        for _, s in pairs(test_states) do
+            if s == state then
+                return true
+            end
+        end
+        return false
+    end, {state})
+end
+
 local function wait_state(server, state)
     t.helpers.retrying({}, function()
-        t.assert(server:grep_log('State '.. state .. ' reached'),
+        t.assert(is_state_reached(server, state),
                  'State ' .. state .. ' triggers fired')
     end)
 end
 
 local function assert_state_not_reached(server, state)
-    t.assert_equals(server:grep_log('State ' .. state .. ' reached'), nil,
-                    'State ' .. state .. ' not reached')
+    t.assert(not is_state_reached(server, state),
+             'State ' .. state .. ' not reached')
 end
 
 local function wait_usual_state_order(server)
```

src/box/box_trigger_def.h
Gerold103: I wouldn't worry about the header dependency. It depends on libbox anyway, whatever you do. Lets better do it without new files.

About the copy - your current solution also does the copying on the stack. There wouldn't be copying if you would store it as const char * (I understand why you can't do it here), but when you do char[] the string literal is copied onto the stack.

src/box/box.cc
sergepetrenko: >  Btw, I just thought - why do we make these recovery states as triggers? Wouldn't a new built-in event work? Similar to "box.status".

The main difference between watchers and triggers is that triggers are run synchronously, and watcher callbacks are run asynchronously. So using watchers here would either mean that the events are delivered when box.cfg yields (which means we lose the exact moment when event happened), or we have to add more yields where the notification is sent.

So it seems triggers suit this particular use case more. For example, one might want to set up triggers on xlog recovery, without touching snapshot data.

> Yes. Maybe any of these states shouldn't be fired more than once?

I agree. Since the triggeres are called `on_recovery_state`, they should be run only during "recovery" or initial sync.
I'll add a run once guard for "synced" stage. All the other stages are already run once.

sergepetrenko: Here. I've also added an enumeration of all states, so calling a trigger with some specific state is less error-prone now.

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 66f8c49a3..1f0528a70 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -90,7 +90,6 @@
 #include "flightrec.h"
 #include "wal_ext.h"
 #include "mp_util.h"
-#include "box_trigger_def.h"
 
 static char status[64] = "unconfigured";
 
@@ -416,7 +415,7 @@ box_do_set_orphan(bool orphan)
 	is_orphan = orphan;
 	box_update_ro_summary();
 	if (!is_orphan)
-		box_run_on_recovery_state("synced");
+		box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
 }
 
 void
@@ -3772,7 +3771,7 @@ bootstrap_master(const struct tt_uuid *replicaset_uuid)
 	if (gc_checkpoint() != 0)
 		panic("failed to create a checkpoint");
 
-	box_run_on_recovery_state("snapshot_recovered");
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
 }
 
 /**
@@ -3822,7 +3821,7 @@ bootstrap_from_master(struct replica *master)
 					APPLIER_FINAL_JOIN;
 	applier_resume_to_state(applier, wait_state, TIMEOUT_INFINITY);
 
-	box_run_on_recovery_state("snapshot_recovered");
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
 
 	/*
 	 * Process final data (WALs).
@@ -4054,7 +4053,7 @@ local_recovery(const struct tt_uuid *instance_uuid,
 	 */
 	memtx_engine_recover_snapshot_xc(memtx, checkpoint_vclock);
 
-	box_run_on_recovery_state("snapshot_recovered");
+	box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
 
 	engine_begin_final_recovery_xc();
 	recover_remaining_wals(recovery, &wal_stream.base, NULL, false);
@@ -4358,7 +4357,7 @@ box_cfg_xc(void)
 
 	rmean_cleanup(rmean_box);
 
-	box_run_on_recovery_state("wal_recovered");
+	box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
 
 	/* Follow replica */
 	replicaset_follow();
@@ -4399,8 +4398,8 @@ box_cfg_xc(void)
 	say_info("ready to accept requests");
 
 	if (is_bootstrap_leader) {
-		box_run_on_recovery_state("indexes_built");
-		box_run_on_recovery_state("synced");
+		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+		box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
 	}
 	if (!is_bootstrap_leader) {
 		replicaset_sync();
@@ -4455,6 +4454,42 @@ box_cfg(void)
 	}
 }
 
+/**
+ * Whether the recovery state triggers for a specific state have already run.
+ */
+static bool recovery_state_reached[RECOVERY_STATE_MAX];
+
+static const char *recovery_state_names[RECOVERY_STATE_MAX] = {
+	[RECOVERY_STATE_SNAPSHOT_RECOVERED] = "snapshot_recovered",
+	[RECOVERY_STATE_WAL_RECOVERED] = "wal_recovered",
+	[RECOVERY_STATE_INDEXES_BUILT] = "indexes_built",
+	[RECOVERY_STATE_SYNCED] = "synced",
+};
+
+int
+box_run_on_recovery_state(enum box_recovery_state state)
+{
+	/* Don't run triggers for the same stage twice. */
+	if (recovery_state_reached[state])
+		return 0;
+	recovery_state_reached[state] = true;
+	/*
+	 * on_recovery_state triggers expect a string, but passing a string
+	 * literal to them turns out quite ugly, because trigger_run takes
+	 * void *. So one ends up with
+	 * trigger_run(trig, (char *)"string_literal");
+	 *
+	 * Copying the string turns out less ugly and more safe, because the
+	 * resulting string becomes writable.
+	 */
+	assert(state >= 0 && state < RECOVERY_STATE_MAX);
+	char buf[64];
+	size_t len = strlcpy(buf, recovery_state_names[state], sizeof(buf));
+	assert(len < sizeof(buf));
+	(void)len;
+	return trigger_run(&box_on_recovery_state, buf);
+}
+
 /**
  * box.coredump() forks to save a core. The entire
  * server forks in box.cfg{} if background=true.
diff --git a/src/box/box.h b/src/box/box.h
index acb76f425..4a7afd57a 100644
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -70,6 +70,47 @@ extern double on_shutdown_trigger_timeout;
 /** Invoked on box shutdown. */
 extern struct rlist box_on_shutdown_trigger_list;
 
+/**
+ * Triggers invoked during initial `box.cfg` call on various recovery stages.
+ */
+extern struct rlist box_on_recovery_state;
+
+/**
+ * Recovery states supported by on_recovery_state triggers.
+ * Are positioned in the order of appearance during initial box.cfg().
+ * The only exception are WAL_RECOVERED and INDEXES_BUILT, which might come in
+ * any order, since the moment when secondary indexes are built depends on
+ * box.cfg.force_recovery and box.cfg.memtx_use_mvcc_engine.
+ */
+enum box_recovery_state {
+	/**
+	 * The node has either recovered the snapshot from the disk, received
+	 * the snapshot from the remote master as part of initial join stage or
+	 * has bootstrapped the cluster.
+	 */
+	RECOVERY_STATE_SNAPSHOT_RECOVERED,
+	/**
+	 * The node has either recovered the local WAL files, received the WALs
+	 * from the remote master as part of final join or has bootstrapped the
+	 * cluster.
+	 */
+	RECOVERY_STATE_WAL_RECOVERED,
+	/**
+	 * The node has built secondary indexes for memtx spaces.
+	 */
+	RECOVERY_STATE_INDEXES_BUILT,
+	/**
+	 * The node has synced with remote peers. IOW, it has transitioned from
+	 * "orphan" to "running".
+	 */
+	RECOVERY_STATE_SYNCED,
+	RECOVERY_STATE_MAX,
+};
+
+/** Run the on_recovery_state_triggers with state \a state. */
+int
+box_run_on_recovery_state(enum box_recovery_state state);
+
 /**
  * Timeout during which the transaction must complete,
  * otherwise it will be rolled back.
diff --git a/src/box/box_trigger_def.h b/src/box/box_trigger_def.h
deleted file mode 100644
index 5e0fa438e..000000000
--- a/src/box/box_trigger_def.h
+++ /dev/null
@@ -1,36 +0,0 @@
-#pragma once
-/*
- * SPDX-License-Identifier: BSD-2-Clause
- *
- * Copyright 2022, Tarantool AUTHORS, please see AUTHORS file.
- */
-
-#include "trigger.h"
-
-/**
- * Triggers invoked during initial `box.cfg` call on various recovery stages.
- */
-extern struct rlist box_on_recovery_state;
-
-/** Run box_on_recovery_state triggers with state represented by \a str. */
-static inline int
-box_run_on_recovery_state(const char *str)
-{
-	/*
-	 * on_recovery_state triggers expect a string, but passing a string
-	 * literal to them turns out quite ugly, because trigger_run takes
-	 * void *. So one ends up with
-	 * trigger_run(trig, (char *)"string_literal");
-	 *
-	 * Compound literals solve this problem in a less ugly way, besides they
-	 * can really be written to:
-	 * trigger_run(trig, (char[]){"string_literal"}).
-	 * But unfortunately they aren't supported in c++.
-	 * So we are left with this:
-	 */
-	char buf[64];
-	size_t len = strlcpy(buf, str, sizeof(buf));
-	assert(len < sizeof(buf));
-	(void)len;
-	return trigger_run(&box_on_recovery_state, buf);
-}
diff --git a/src/box/lua/ctl.c b/src/box/lua/ctl.c
index 8d47fe69d..40d5ae4f2 100644
--- a/src/box/lua/ctl.c
+++ b/src/box/lua/ctl.c
@@ -40,7 +40,6 @@
 #include "lua/trigger.h"
 
 #include "box/box.h"
-#include "box/box_trigger_def.h"
 #include "box/schema.h"
 #include "box/engine.h"
 #include "box/memtx_engine.h"
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 4049e3848..6a1e05aa2 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -56,7 +56,7 @@
 #include "read_view.h"
 #include "memtx_tuple_compression.h"
 #include "memtx_space.h"
-#include "box_trigger_def.h"
+#include "box.h"
 
 #include <type_traits>
 
@@ -437,7 +437,7 @@ memtx_engine_begin_final_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state("indexes_built");
+		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
 	}
 	return 0;
 }
@@ -456,7 +456,7 @@ memtx_engine_begin_hot_standby(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state("indexes_built");
+		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
 	}
 	return 0;
 }
@@ -476,7 +476,7 @@ memtx_engine_end_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state("indexes_built");
+		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
 	}
 	xdir_collect_inprogress(&memtx->snap_dir);
 
diff --git a/test/replication-luatest/recovery_triggers_test.lua b/test/replication-luatest/recovery_triggers_test.lua
index 943146f15..870990032 100644
--- a/test/replication-luatest/recovery_triggers_test.lua
+++ b/test/replication-luatest/recovery_triggers_test.lua
@@ -32,8 +32,7 @@ end)
 
 local function is_state_reached(server, state)
     return server:exec(function(state)
-        -- luacheck: ignore
-        for _, s in pairs(test_states) do
+        for _, s in pairs(_G.test_states) do
             if s == state then
                 return true
             end
@@ -95,4 +94,14 @@ g.test_recovery_stages = function(cg)
         }
     end)
     wait_state(cg.replica, 'synced')
+    -- Check that triggers on 'synced' stage are executed only once.
+    cg.replica:exec(function()
+        local t = require('luatest')
+        local num_events = #_G.test_states
+        local old_repl = box.cfg.replication
+        box.cfg{replication = ''}
+        box.cfg{replication = old_repl}
+        t.assert_equals(#_G.test_states, num_events,
+                        "Triggers on sync aren't re-run")
+    end)
 end
```

</details>

src/box/box_trigger_def.h
sergepetrenko: > I wouldn't worry about the header dependency. It depends on libbox anyway, whatever you do. Lets better do it without new files.

Ok, I moved everything to `box.h` / `box.cc`.

> About the copy - your current solution also does the copying on the stack. There wouldn't be copying if you would store it as const char * (I understand why you can't do it here), but when you do char[] the string literal is copied onto the stack.

Yep, makes sense. It was just hidden.

Gerold103 (approved):
Thanks for the patch!

Totktonada:
I would highlight that unnamed triggers are hard to manage from a module side. A module should take care of the presence of others' triggers (do wrapping) and of hot reload to not add a trigger again (find and remove the old trigger somehow).

Sometimes it is even possible that the wrapping in a module is written correctly, but unwrapping/removing drops others' triggers, because it restores some old saved trigger function.

Totktonada:
Related discussion: https://github.com/tarantool/tarantool/discussions/7563

sergepetrenko:
@Totktonada
> I would highlight that unnamed triggers are hard to manage from a module side. A module should take care of the presence of others' triggers (do wrapping) and of hot reload to not add a trigger again (find and remove the old trigger somehow).

Yes, our trigger interface has this problem. However, I think we should introduce new behaviour for all triggers at once. And not try to do that for these particular triggers only.

sergepetrenko:
> Related discussion: https://github.com/orgs/tarantool/discussions/7563

Speaking of making stages comparable, this won't be possible for indexes_built and wal_recovered. They can come in any order.

test/replication-luatest/recovery_triggers_test.lua
```diff
@@ -0,0 +1,107 @@
...
+    cg.server = cg.cluster:build_and_add_server({
+        alias = 'server',
+        args = {run_before_cfg},
+    })
+    cg.replica = cg.cluster:build_and_add_server({
+        alias = 'replica',
+        box_cfg = {
+            replication = server.build_instance_uri('server'),
+        },
+        args = {run_before_cfg},
```
locker, line 25: Passing the code like this is unidiomatic. Please add a new properly named argument to the server constructor instead. Also, better to pass the code in an environment variable instead of the argument list, because environment variables have names.

src/box/memtx_engine.cc
```diff
@@ -56,6 +56,7 @@
 #include "read_view.h"
 #include "memtx_tuple_compression.h"
 #include "memtx_space.h"
+#include "box.h"
```
locker, line 59: Making `memtx_engine.cc` depend on `box.h` doesn't look good. Why not add a new header file for recovery triggers?

src/box/box.cc
```diff
@@ -4437,6 +4454,46 @@ box_cfg(void)
...
+	/* [RECOVERY_STATE_INDEXES_BUILT] = */
+	"indexes_built",
+	/* [RECOVERY_STATE_SYNCED] = */
+	"synced",
+};
+
+int
+box_run_on_recovery_state(enum box_recovery_state state)
+{
+	/* Don't run triggers for the same stage twice. */
```
locker, line 4476: Do I understand correctly that we need this only for the 'sync' state? If so, this looks rather ugly. IMO better add a flag to wherever you invoke the trigger instead.

src/box/box.cc
```diff
@@ -4437,6 +4454,46 @@ box_cfg(void)
...
+		return 0;
+	recovery_state_reached[state] = true;
+	/*
+	 * on_recovery_state triggers expect a string, but passing a string
+	 * literal to them turns out quite ugly, because trigger_run takes
+	 * void *. So one ends up with
+	 * trigger_run(trig, (char *)"string_literal");
+	 *
+	 * Copying the string turns out less ugly and more safe, because the
+	 * resulting string becomes writable.
```
locker, line 4487: I don't understand why we should bother at all - the trigger isn't going to modify the string, right? And `const` specifying in C is optional and doesn't really enforce constant-ness. So I see nothing wrong in writing:

```C
trigger_run(&box_on_recovery_state, (char *)recovery_state_names[state]);
```

I'm pretty sure we have a few other places where we do the same thing.

src/box/box.cc
```diff
@@ -4384,6 +4397,10 @@ box_cfg_xc(void)
 	title("running");
 	say_info("ready to accept requests");
 
+	if (is_bootstrap_leader) {
+		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
```
locker, line 4401: Why not do that somewhere in memtx engine callback instead so that we have all "indexes_built" events in the same file? What about `memtx_engine_bootstrap`?

src/box/box.cc
```diff
@@ -4346,6 +4357,8 @@ box_cfg_xc(void)
 
 	rmean_cleanup(rmean_box);
 
+	box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
```
locker, line 4360: Let's do this in `boostrap_master`, `bootstrap_from_master`, and `local_recovery` - the code will be easier to follow if "wal_recovered" is located after "snapshot_recovered" in the same function.

changelogs/unreleased/gh-3159-recovery-triggers.md
```diff
@@ -0,0 +1,7 @@
+## feature/box
+
+* It is now possible to register triggers for various recovery stages: use
+  `box.ctl.on_recovery_state()` before initial `box.cfg()` call. The trigger
```
locker, line 4: I can't find a design document for this feature. The closest thing I see is [replication_connect_quorum rework (RCQ)](https://www.notion.so/tarantool/replication_connect_quorum-rework-RCQ-a749e1e6b8934f3388b2176ab52c4300) doc, but there this new API is just mentioned, no detailed description provided. Besides, the function is called `on_recovery_stage` there... Since this is a part of the public API, please get an approval from the PM team regarding the names and behavior of this new feature.

src/box/memtx_engine.cc
sergepetrenko: That's what I did initially but @Gerold103 convinced me not to do so: https://github.com/tarantool/tarantool/pull/7938#discussion_r1024632116

changelogs/unreleased/gh-3159-recovery-triggers.md
sergepetrenko: Good point, I did get a verbal approval, just couldn't summon the products here.
@Mons, do you hear me?

src/box/memtx_engine.cc
locker: Minimizing dependencies reduces compilation time. OTOH `box.h` doesn't depend on much so it's probably fine. Let's ask @alyapunov as the main contributor to the memtx code.

changelogs/unreleased/gh-3159-recovery-triggers.md
locker: If @Mons could read the PR description and approve the PR if everything looks good to him, that'd be great.

In future, please describe all aspects of the proposed public API in details in the design doc so that the reviewer has no questions.

sergepetrenko: > In future, please describe all aspects of the proposed public API in details in the design doc so that the reviewer has no questions.

Ok

src/box/box.cc
sergepetrenko: Yep. Done:

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 543b25407..4eafe0d3a 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -404,13 +404,18 @@ box_wait_ro(bool ro, double timeout)
        return 0;
 }
 
+/** Whether the triggers for "synced" recovery stage have already run. */
+static bool recovery_state_synced_is_reached;
+
 void
 box_do_set_orphan(bool orphan)
 {
        is_orphan = orphan;
        box_update_ro_summary();
-       if (!is_orphan)
+       if (!is_orphan && !recovery_state_synced_is_reached) {
                box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
+               recovery_state_synced_is_reached = true;
+       }
 }
 
 void
@@ -4398,7 +4403,9 @@ box_cfg_xc(void)
 
        if (is_bootstrap_leader) {
                box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+               assert(!recovery_state_synced_is_reached);
                box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
+               recovery_state_synced_is_reached = true;
        }
        if (!is_bootstrap_leader) {
                replicaset_sync();
@@ -4453,11 +4460,6 @@ box_cfg(void)
        }
 }
 
-/**
- * Whether the recovery state triggers for a specific state have already run.
- */
-static bool recovery_state_reached[RECOVERY_STATE_MAX];
-
 static const char *recovery_state_names[RECOVERY_STATE_MAX] = {
        /* [RECOVERY_STATE_SNAPSHOT_RECOVERED] = */
        "snapshot_recovered",
@@ -4472,10 +4474,6 @@ static const char *recovery_state_names[RECOVERY_STATE_MAX] = {
 int
 box_run_on_recovery_state(enum box_recovery_state state)
 {
-       /* Don't run triggers for the same stage twice. */
-       if (recovery_state_reached[state])
-               return 0;
-       recovery_state_reached[state] = true;
        /*
         * on_recovery_state triggers expect a string, but passing a string
         * literal to them turns out quite ugly, because trigger_run takes
```

</details>

src/box/box.cc
sergepetrenko: Done. I've also simplifed the bootstrap sequence a bit.

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 4eafe0d3a..363452f30 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3779,6 +3779,11 @@ bootstrap_master(const struct tt_uuid *replicaset_uuid)
                panic("failed to create a checkpoint");
 
        box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
+       box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
+       box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+       assert(!recovery_state_synced_is_reached);
+       box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
+       recovery_state_synced_is_reached = true;
 }
 
 /**
@@ -3864,6 +3869,8 @@ bootstrap_from_master(struct replica *master)
        if (gc_checkpoint() != 0)
                panic("failed to create a checkpoint");
 
+       box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
+
        return true;
 }
 
@@ -4164,6 +4171,8 @@ local_recovery(const struct tt_uuid *instance_uuid,
                          tt_uuid_str(replicaset_uuid),
                          tt_uuid_str(&REPLICASET_UUID));
        }
+
+       box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
 }
 
 static void
@@ -4362,8 +4371,6 @@ box_cfg_xc(void)
 
        rmean_cleanup(rmean_box);
 
-       box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
-
        /* Follow replica */
        replicaset_follow();
 
@@ -4401,12 +4408,6 @@ box_cfg_xc(void)
        title("running");
        say_info("ready to accept requests");
 
-       if (is_bootstrap_leader) {
-               box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
-               assert(!recovery_state_synced_is_reached);
-               box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
-               recovery_state_synced_is_reached = true;
-       }
        if (!is_bootstrap_leader) {
                replicaset_sync();
        } else if (box_election_mode == ELECTION_MODE_CANDIDATE ||
```

</details>

src/box/box.cc
sergepetrenko: Please see my response in the conversation below. I've moved the whole trigger run sequence for bootstrap leader into `bootstrap_master`.

I don't think we should place INDEXES_BUILT into `memtx_engine_bootstrap` though. This would in turn require placing SNAPSHOT_RECOVERED there, because it should go before INDEXES_BUILT. And we'll have SNAPSHOT_RECOVERED in two files, memtx.cc and box.cc then.

src/box/box.cc
sergepetrenko: >  the trigger isn't going to modify the string, right?

Yes, it isn't going to modify the string, and actually the string is copied anyway for Lua triggers (via `lua_pushstring` in `lbox_push_recovery_state`). So your suggestion would work just fine now.

On the other hand, the trigger run function receives a `(void *)` argument, and nothing suggests that the argument can't be written to. That's why I want to remain on the safe side for the case when someone accidentally writes to the string in a C trigger. IMO if we promise that some argument is non-const, it should really be non-const.

What do you think?

test/replication-luatest/recovery_triggers_test.lua
sergepetrenko: Ok, what do you think of this? I don't see a point in a separate named argument, primarily because this feature depends on a specific instance file. So a named environment variable should do:
```diff
diff --git a/test/instances/default.lua b/test/instances/default.lua
index 8deda9d0b..e933f22d8 100755
--- a/test/instances/default.lua
+++ b/test/instances/default.lua
@@ -38,7 +38,7 @@ end
 -- such bad tests hang and fail.
 box.ctl.set_on_shutdown_timeout(TIMEOUT_INFINITY)
 
-local exec_arg = arg[1]
+local exec_arg = os.getenv('TARANTOOL_RUN_BEFORE_CFG')
 if exec_arg then
     local run_before_cfg = loadstring(exec_arg)
     run_before_cfg()
```

```diff
diff --git a/test/replication-luatest/recovery_triggers_test.lua b/test/replication-luatest/recovery_triggers_test.lua
index 870990032..755bf2523 100644
--- a/test/replication-luatest/recovery_triggers_test.lua
+++ b/test/replication-luatest/recovery_triggers_test.lua
@@ -15,14 +15,18 @@ g.before_all(function(cg)
     cg.cluster = cluster:new({})
     cg.server = cg.cluster:build_and_add_server({
         alias = 'server',
-        args = {run_before_cfg},
+        env = {
+            ['TARANTOOL_RUN_BEFORE_CFG'] = run_before_cfg,
+        },
     })
     cg.replica = cg.cluster:build_and_add_server({
         alias = 'replica',
         box_cfg = {
             replication = server.build_instance_uri('server'),
         },
-        args = {run_before_cfg},
+        env = {
+            ['TARANTOOL_RUN_BEFORE_CFG'] = run_before_cfg,
+        },
     })
 end)

```

src/box/box.cc
locker: > That's why I want to remain on the safe side for the case when someone accidentally writes to the string in a C trigger.

If someone accidentally writes to a constant string, they'll get SIGSEGV, because the section is read-only.

> IMO if we promise that some argument is non-const, it should really be non-const.

Only if it comes for free. Following the "non-const unless marked as const" rule just for the sake of it isn't worth complicating the code.

src/box/box.cc
```diff
@@ -4436,6 +4461,37 @@ box_cfg(void)
 	}
 }
 
+static const char *recovery_state_names[RECOVERY_STATE_MAX] = {
+	/* [RECOVERY_STATE_SNAPSHOT_RECOVERED] = */
```
locker, line 4465: Nit: You don't need to comment it. In modern C you could just as well write

```C
[RECOVERY_STATE_SNAPSHOT_RECOVERED] = "snapshot_recovered",
```

and we quite often do that.

src/box/box.h
```diff
@@ -70,6 +70,47 @@ extern double on_shutdown_trigger_timeout;
...
+	/**
+	 * The node has built secondary indexes for memtx spaces.
+	 */
+	RECOVERY_STATE_INDEXES_BUILT,
+	/**
+	 * The node has synced with remote peers. IOW, it has transitioned from
+	 * "orphan" to "running".
+	 */
+	RECOVERY_STATE_SYNCED,
+	RECOVERY_STATE_MAX,
```
locker, line 107: Nit: For `STR2ENUM` and `ENUM2STR` helpers to work, this should be called `box_recovery_state_MAX`.

test/replication-luatest/recovery_triggers_test.lua
locker: Looks good.

src/box/box.cc
locker: > I don't think we should place INDEXES_BUILT into `memtx_engine_bootstrap` though. This would in turn require placing SNAPSHOT_RECOVERED there, because it should go before INDEXES_BUILT.

Should it though? AFAIU "index_built" can be triggered before or after "wal_recovered". What's wrong if it's triggered before "snapshot_recovered"? In case of bootstrap, technically it seems to be correct.

src/box/box.cc
sergepetrenko: Yes, and I did that initially, but this file is C++, and some compilers error with "designated array initializers are a C extension" or something like that. It was caught with full-ci, AFAIR.

BTW, C++20 allows designated initializers for structs, but not for arrays.

It would look nicer this way, indeed.

locker: Oh, I missed that it's C++. Okay then.

src/box/box.cc
sergepetrenko: > If someone accidentally writes to a constant string, they'll get SIGSEGV, because the section is read-only.

True. But the trigger could write to the argument conditionally. And some condition might fire on the user side only.
Anyway, I'm just inventing possible scenarios here.

> Only if it comes for free. Following the "non-const unless marked as const" rule just for the sake of it isn't worth complicating the code.

As you wish:
```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 363452f30..2bda374f4 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4475,21 +4475,9 @@ static const char *recovery_state_names[RECOVERY_STATE_MAX] = {
 int
 box_run_on_recovery_state(enum box_recovery_state state)
 {
-       /*
-        * on_recovery_state triggers expect a string, but passing a string
-        * literal to them turns out quite ugly, because trigger_run takes
-        * void *. So one ends up with
-        * trigger_run(trig, (char *)"string_literal");
-        *
-        * Copying the string turns out less ugly and more safe, because the
-        * resulting string becomes writable.
-        */
        assert(state >= 0 && state < RECOVERY_STATE_MAX);
-       char buf[64];
-       size_t len = strlcpy(buf, recovery_state_names[state], sizeof(buf));
-       assert(len < sizeof(buf));
-       (void)len;
-       return trigger_run(&box_on_recovery_state, buf);
+       return trigger_run(&box_on_recovery_state,
+                          (char *)recovery_state_names[state]);
 }
 
 /**
```

src/box/memtx_engine.cc
alyapunov: I don't like this dependency either. Logically memtx_engine (and other engines) should be a small brick on which box stands. Perhaps it is far for ideal, but we should not make it worse. I don't think that compilation time will be affected, but that line of code directly introduces box and memtx_engine cross dependency.

src/box/box.cc
locker: > True. But the trigger could write to the argument conditionally. And some condition might fire on the user side only.

All possible code paths should be covered by tests. If a code path isn't covered, anything can happen.

> As you wish

Thanks :)

sergepetrenko: > All possible code paths should be covered by tests.

That's not what usually happens with our tests :)

src/box/box.cc
locker: > It seems to me we need some order for the triggers (unless we are talking about wal_recovered and indexes_built, which can come in any order).

There's a definite order: snap < wal < sync. The "indexes_built" event is already special: it's < sync, but may be < or > wal. Since it's special, there's nothing wrong in going further and saying that it may be < or > snap. Technically, we could introduce the mode in which indexes are built right before snapshot is recovered, e.g. we could do that in force_recovery mode.

> Maintaining the order seems cheap enough

It's not cheap - it makes the code more difficult to follow and, what is worse, prevents us from building indexes before snapshot is recovered in future.

> For example, @Totktonada suggested that the stages should be comparable in #7563.

Let's go from use cases: suppose the user knows that indexes_built is always fired after snap_recovered. What good does they get from that knowledge? I imagine that the user might want to know the moment when the database becomes usable (everything is recovered, and indexes are built), but currently there's no event for this except 'sync', which isn't quite what they might want; the user is supposed to wait for wal_recovered **and** indexes_built instead, which isn't quite convenient. Moving indexes_built before snap_recovered changes absolutely nothing from that point. Maybe, we should introduce a new event which would be fired when the database becomes fully functional, that is max(snap_recovered, indexes_built)?

src/box/memtx_engine.cc
sergepetrenko: Let's ask @Gerold103 as well, since his words were "Lets better do it without new files."

src/box/box.cc
sergepetrenko: > Technically, we could introduce the mode in which indexes are built right before snapshot is recovered, e.g. we could do that in force_recovery mode.

Anyway "indexes_built" implies that the data is accessible, beacause it's already recovered and there are indexes for the data.

Building the indexes before recovering the snap makes no sense -- since there is nothing to read yet. Even with "built" indexes.

> the user is supposed to wait for wal_recovered and indexes_built instead

This is true.

> Maybe, we should introduce a new event which would be fired when the database becomes fully functional, that is max(snap_recovered, indexes_built)?

The event you are speaking of should be max(wal_recovered, indexes_built).

After thinking this through, it seems to me we shouldn't care about the real moment when indexes are built, and instead always fire "indexes_built" after "wal_recovered". What do you think?

When indexes are built after local recovery completes, we aren't lying to the user and this stage has a use case: let the user know that the data is fully accessible (via secondary indices).

When indexes are built on top of an intermediate state (snap is recovered, but WAL isn't), there is no use for the user to know that the indexes are built. So it seems placing "indexes_built" after "wal_recovered" is good in this case too.

Besides, the implementation details will be hidden from the user, a full order will be established, and we will be free to build indexes even before the snapshot recovery, as you say.

locker: > Building the indexes before recovering the snap makes no sense -- since there is nothing to read yet. Even with "built" indexes.

Why not? This means that indexes are built, but the data isn't recovered yet. As we recover data, we will insert it into all indexes right away.

> After thinking this through, it seems to me we shouldn't care about the real moment when indexes are built, and instead always fire "indexes_built" after "wal_recovered". What do you think?

Sounds reasonable, **but** in the near future, the memtx mvcc mode will be enabled by default and indexes will be always built right after snapshot recovery so indexes_built will be fired right after wal_recovered. Do we even need this event then?

Moreover, thinking even further, in distant future, we may end up with *continuous* snapshotting, i.e. there will be no such thing as snapshot recovery at all and all the three events (snap_recovered, wal_recovered, indexes_built) will be fired one after another without any delay when recovery is fully completed.

Come to think of it, why do we need to separate snapshot_recovered from wal_recovered? Do you have a use case in mind? Why not just add one trigger - on_recovery_complete - which would fire after recovering snap and wal and building indexes?

sergepetrenko: I'll paste @Mons's words here:
```
indexes_built важен тем, что до того как они построены любая триггерная операция потенциально приводит к duplicate_key.
т.е. я вижу indexes_built не как строгий элемент порядка загрузки, а как момент, после которого можно полноценно пользоваться индексами.
```

and 
```
я бы не завязывался на предопределённом порядке
```

So it seems, "indexes_built" should mean what you are saying -- indexes are usable.
And there is a use case for "indexes_built" -- know if constraints are checked already or not (useful for before_replace triggers).

"recovery_finished" (which is wal_recovered now) is also needed: to catch the moment when local data is fully recovered (to access it via primary key at least).

I can't remember why "snapshot_recovered" is needed, but @ochaton asked for it explicitly.

Anyway, seems we don't care about the trigger invocation order for bootstrap. So I'll try to implement your suggestiong with "indexes_built" in `memtx_engine_bootstrap`.

src/box/memtx_engine.cc
Gerold103: I would agree with @alyapunov and @locker if the triggers wouldn't be in box.cc. But they are. Memtx engine depends on box.cc in both solutions. And that is fine IMO. They are a part of libbox both. Separating memtx engine outside of libbox might be possible, but the triggers are the smallest problem of all then. And our new trigger isn't even the first one like that.

So, in total, both solutions make memtx engine depend on box internals. But the current one (no new headers) is 1) smaller, 2) consistent with `box_on_shutdown_trigger_list`. Thus I still vote for not adding new headers.

However I won't block the patch either way.

src/box/box.cc
Totktonada: > It seems to me we need some order for the triggers (unless we are talking about wal_recovered and indexes_built, which can come in any order). For example, @Totktonada suggested that the stages should be comparable in #7563.
Maintaining the order seems cheap enough (it's the matter of one extra usage of "INDEXES_BUILT" in `bootstrap_master` compared to your solution), so I'd stick with ordered trigger invocation where it is possible.

When I wrote it I thought that the states appear in particular fixed order. If it doesn't reflect the world, let's ignore and go ahead with something that would reflect it better.

Regarding on the rest of the discussion: we can go from use cases (and improve particular user scenarios) or go from stages we already have in the code (improve tarantool's state observability). The latter may solve some use cases as well, but in fact I wouldn't consider it as the primary goal.

Gerold103:
Here is what I managed to craft for running code before box.cfg:
```Lua
local server = require('test.luatest_helpers.server')
local t = require('luatest')
local g = t.group()

g.before_all = function()
    g.server = server:new({
        alias = 'master',
        args = {
            '-e', 'rawset(_G, "test", true)',
        }
    })
    g.server:start()
end

g.after_all = function()
    g.server:stop()
end

g.test_sigterm_during_initial_snapshot = function()
    g.server:exec(function()
        local t = require('luatest')
        local log = require('log')
        local yaml = require('yaml')
        log.info(yaml.encode(arg))
        t.assert(rawget(_G, "test"))
    end)
end
```
The only problem is that while it works fine in a console (`_G.test` is true), in this example somewhy the command is not executed. Would be cool to find out why. (I need to run code before `box.cfg` to set an error injection in my own patch.)

Totktonada:
@Gerold103 I guess it is due to the order of arguments:

```shell
$ tarantool <(echo "print(42)") -e "print('43')"
42
$ tarantool -e "print('43')" <(echo "print(42)")
43
42
```

https://github.com/tarantool/luatest/blob/f2dc273f1ff884c1943f3003ff75eb6109a68f5b/luatest/server.lua#L137-L140

Server's `args` parameter is to pass arguments to a script:

```
$ tarantool <(echo "print(...)") -e "print('43')"
-e	print('43')
```

We can add a parameter to the server helper to run something at beginning of the default instance. The default instance will have `loadstring()` or `dostring()` so.

Totktonada:
Modification of a luatest helper is boring, let's use code injection instead:

```lua
tarantool> box.cfg{replication_synchro_quorum = [=[N / 2 + 1]] _G.test = true --[[]=]}
tarantool> test
---
- true
...
```

src/box/box.cc
sergepetrenko: Done:
```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index afa20ba4f..b75a82c1d 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3780,7 +3780,6 @@ bootstrap_master(const struct tt_uuid *replicaset_uuid)
 
        box_run_on_recovery_state(RECOVERY_STATE_SNAPSHOT_RECOVERED);
        box_run_on_recovery_state(RECOVERY_STATE_WAL_RECOVERED);
-       box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
        assert(!recovery_state_synced_is_reached);
        box_run_on_recovery_state(RECOVERY_STATE_SYNCED);
        recovery_state_synced_is_reached = true;
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index b5487d620..9a1becaed 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -636,7 +636,10 @@ memtx_engine_bootstrap(struct engine *engine)
                        break;
        }
        xlog_cursor_close(&cursor, false);
-       return rc < 0 ? -1 : 0;
+       if (rc < 0)
+               return -1;
+       box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+       return 0;
 }
 
 static int
diff --git a/test/replication-luatest/recovery_triggers_test.lua b/test/replication-luatest/recovery_triggers_test.lua
index 755bf2523..f1e9d3050 100644
--- a/test/replication-luatest/recovery_triggers_test.lua
+++ b/test/replication-luatest/recovery_triggers_test.lua
@@ -57,7 +57,7 @@ local function assert_state_not_reached(server, state)
              'State ' .. state .. ' not reached')
 end
 
-local function wait_usual_state_order(server)
+local function wait_all_states(server)
     wait_state(server, 'snapshot_recovered')
     -- wal_recovered and indexes_built can come in any order.
     wait_state(server, 'wal_recovered')
@@ -68,14 +68,14 @@ end
 g.test_recovery_stages = function(cg)
     -- Test stages on bootstrap.
     cg.server:start()
-    wait_usual_state_order(cg.server)
+    wait_all_states(cg.server)
     cg.server:stop()
     -- Test same stages on recovery.
     cg.server:start()
-    wait_usual_state_order(cg.server)
+    wait_all_states(cg.server)
     -- Test the stages on replica join.
     cg.replica:start()
-    wait_usual_state_order(cg.replica)
+    wait_all_states(cg.replica)
     cg.replica:stop()
     -- Test sync state.
     cg.replica.box_cfg = {
```

src/box/memtx_engine.cc
sergepetrenko: After a chat with @locker we decided to add a callback for memtx to call when indexes are built.
Now the triggers are local to box, and memtx receives a calback which calls the triggers from box.

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index b75a82c1d..20ca17896 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -168,6 +168,60 @@ static struct cbus_endpoint tx_prio_endpoint;
 
 RLIST_HEAD(box_on_recovery_state);
 
+/**
+ * Recovery states supported by on_recovery_state triggers.
+ * Are positioned in the order of appearance during initial box.cfg().
+ * The only exception are WAL_RECOVERED and INDEXES_BUILT, which might come in
+ * any order, since the moment when secondary indexes are built depends on
+ * box.cfg.force_recovery and box.cfg.memtx_use_mvcc_engine.
+ */
+enum box_recovery_state {
+	/**
+	 * The node has either recovered the snapshot from the disk, received
+	 * the snapshot from the remote master as part of initial join stage or
+	 * has bootstrapped the cluster.
+	 */
+	RECOVERY_STATE_SNAPSHOT_RECOVERED,
+	/**
+	 * The node has either recovered the local WAL files, received the WALs
+	 * from the remote master as part of final join or has bootstrapped the
+	 * cluster.
+	 */
+	RECOVERY_STATE_WAL_RECOVERED,
+	/**
+	 * The node has built secondary indexes for memtx spaces.
+	 */
+	RECOVERY_STATE_INDEXES_BUILT,
+	/**
+	 * The node has synced with remote peers. IOW, it has transitioned from
+	 * "orphan" to "running".
+	 */
+	RECOVERY_STATE_SYNCED,
+	box_recovery_state_MAX,
+};
+
+static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
+	/* [RECOVERY_STATE_SNAPSHOT_RECOVERED] = */
+	"snapshot_recovered",
+	/* [RECOVERY_STATE_WAL_RECOVERED] = */
+	"wal_recovered",
+	/* [RECOVERY_STATE_INDEXES_BUILT] = */
+	"indexes_built",
+	/* [RECOVERY_STATE_SYNCED] = */
+	"synced",
+};
+
+/** Whether the triggers for "synced" recovery stage have already run. */
+static bool recovery_state_synced_is_reached;
+
+int
+box_run_on_recovery_state(enum box_recovery_state state)
+{
+	assert(state >= 0 && state < box_recovery_state_MAX);
+	return trigger_run(&box_on_recovery_state,
+			   (char *)box_recovery_state_strs[state]);
+}
+
 static void
 builtin_events_init(void);
 
@@ -404,9 +458,6 @@ box_wait_ro(bool ro, double timeout)
 	return 0;
 }
 
-/** Whether the triggers for "synced" recovery stage have already run. */
-static bool recovery_state_synced_is_reached;
-
 void
 box_do_set_orphan(bool orphan)
 {
@@ -3691,6 +3742,12 @@ box_free(void)
 	trigger_destroy(&box_on_recovery_state);
 }
 
+static void
+box_on_indexes_built(void)
+{
+	box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+}
+
 static void
 engine_init()
 {
@@ -3708,7 +3765,8 @@ engine_init()
 				    cfg_geti("strip_core"),
 				    cfg_geti("slab_alloc_granularity"),
 				    cfg_gets("memtx_allocator"),
-				    cfg_getd("slab_alloc_factor"));
+				    cfg_getd("slab_alloc_factor"),
+				    box_on_indexes_built);
 	engine_register((struct engine *)memtx);
 	box_set_memtx_max_tuple_size();
 
@@ -4460,25 +4518,6 @@ box_cfg(void)
 	}
 }
 
-static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
-	/* [RECOVERY_STATE_SNAPSHOT_RECOVERED] = */
-	"snapshot_recovered",
-	/* [RECOVERY_STATE_WAL_RECOVERED] = */
-	"wal_recovered",
-	/* [RECOVERY_STATE_INDEXES_BUILT] = */
-	"indexes_built",
-	/* [RECOVERY_STATE_SYNCED] = */
-	"synced",
-};
-
-int
-box_run_on_recovery_state(enum box_recovery_state state)
-{
-	assert(state >= 0 && state < box_recovery_state_MAX);
-	return trigger_run(&box_on_recovery_state,
-			   (char *)box_recovery_state_strs[state]);
-}
-
 /**
  * box.coredump() forks to save a core. The entire
  * server forks in box.cfg{} if background=true.
diff --git a/src/box/box.h b/src/box/box.h
index af7f3b31e..33851c8a8 100644
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -75,42 +75,6 @@ extern struct rlist box_on_shutdown_trigger_list;
  */
 extern struct rlist box_on_recovery_state;
 
-/**
- * Recovery states supported by on_recovery_state triggers.
- * Are positioned in the order of appearance during initial box.cfg().
- * The only exception are WAL_RECOVERED and INDEXES_BUILT, which might come in
- * any order, since the moment when secondary indexes are built depends on
- * box.cfg.force_recovery and box.cfg.memtx_use_mvcc_engine.
- */
-enum box_recovery_state {
-	/**
-	 * The node has either recovered the snapshot from the disk, received
-	 * the snapshot from the remote master as part of initial join stage or
-	 * has bootstrapped the cluster.
-	 */
-	RECOVERY_STATE_SNAPSHOT_RECOVERED,
-	/**
-	 * The node has either recovered the local WAL files, received the WALs
-	 * from the remote master as part of final join or has bootstrapped the
-	 * cluster.
-	 */
-	RECOVERY_STATE_WAL_RECOVERED,
-	/**
-	 * The node has built secondary indexes for memtx spaces.
-	 */
-	RECOVERY_STATE_INDEXES_BUILT,
-	/**
-	 * The node has synced with remote peers. IOW, it has transitioned from
-	 * "orphan" to "running".
-	 */
-	RECOVERY_STATE_SYNCED,
-	box_recovery_state_MAX,
-};
-
-/** Run the on_recovery_state_triggers with state \a state. */
-int
-box_run_on_recovery_state(enum box_recovery_state state);
-
 /**
  * Timeout during which the transaction must complete,
  * otherwise it will be rolled back.
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 9a1becaed..52212d64e 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -429,7 +429,7 @@ memtx_engine_begin_final_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+		memtx->on_indexes_built_cb();
 	}
 	return 0;
 }
@@ -448,7 +448,7 @@ memtx_engine_begin_hot_standby(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+		memtx->on_indexes_built_cb();
 	}
 	return 0;
 }
@@ -468,7 +468,7 @@ memtx_engine_end_recovery(struct engine *engine)
 		memtx->state = MEMTX_OK;
 		if (space_foreach(memtx_build_secondary_keys, memtx) != 0)
 			return -1;
-		box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+		memtx->on_indexes_built_cb();
 	}
 	xdir_collect_inprogress(&memtx->snap_dir);
 
@@ -638,7 +638,7 @@ memtx_engine_bootstrap(struct engine *engine)
 	xlog_cursor_close(&cursor, false);
 	if (rc < 0)
 		return -1;
-	box_run_on_recovery_state(RECOVERY_STATE_INDEXES_BUILT);
+	memtx->on_indexes_built_cb();
 	return 0;
 }
 
@@ -1263,7 +1263,8 @@ struct memtx_engine *
 memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		 uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		 bool dontdump, unsigned granularity,
-		 const char *allocator, float alloc_factor)
+		 const char *allocator, float alloc_factor,
+		 memtx_on_indexes_built_cb cb)
 {
 	int64_t snap_signature;
 	struct memtx_engine *memtx =
@@ -1370,6 +1371,8 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
 	}
 	tuple_format_ref(memtx->func_key_format);
 
+	memtx->on_indexes_built_cb = cb;
+
 	fiber_start(memtx->gc_fiber, memtx);
 	return memtx;
 fail:
diff --git a/src/box/memtx_engine.h b/src/box/memtx_engine.h
index 2fbc74685..8f259c5ca 100644
--- a/src/box/memtx_engine.h
+++ b/src/box/memtx_engine.h
@@ -109,6 +109,9 @@ enum memtx_reserve_extents_num {
  */
 #define MEMTX_ITERATOR_SIZE (168)
 
+typedef void
+(*memtx_on_indexes_built_cb)(void);
+
 struct memtx_engine {
 	struct engine base;
 	/** Engine recovery state, see enum memtx_recovery_state description. */
@@ -126,6 +129,11 @@ struct memtx_engine {
 	 * needed to be able to cancel it on shutdown.
 	 */
 	struct cord *replica_join_cord;
+	/**
+	 * A callback run once memtx engine builds secondary indexes for the
+	 * data.
+	 */
+	memtx_on_indexes_built_cb on_indexes_built_cb;
 	/** Common quota for tuples and indexes. */
 	struct quota quota;
 	/**
@@ -209,7 +217,8 @@ struct memtx_engine *
 memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		 uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		 bool dontdump, unsigned granularity,
-		 const char *allocator, float alloc_factor);
+		 const char *allocator, float alloc_factor,
+		 memtx_on_indexes_built_cb cb);
 
 int
 memtx_engine_recover_snapshot(struct memtx_engine *memtx,
@@ -341,13 +350,13 @@ static inline struct memtx_engine *
 memtx_engine_new_xc(const char *snap_dirname, bool force_recovery,
 		    uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		    bool dontdump, unsigned granularity,
-		    const char *allocator, float alloc_factor)
+		    const char *allocator, float alloc_factor,
+		    memtx_on_indexes_built_cb cb)
 {
 	struct memtx_engine *memtx;
 	memtx = memtx_engine_new(snap_dirname, force_recovery,
-				 tuple_arena_max_size,
-				 objsize_min, dontdump,
-				 granularity, allocator, alloc_factor);
+				 tuple_arena_max_size, objsize_min, dontdump,
+				 granularity, allocator, alloc_factor, cb);
 	if (memtx == NULL)
 		diag_raise();
 	return memtx; 
```

</details>

src/box/memtx_engine.h
```diff
@@ -209,7 +217,8 @@ struct memtx_engine *
 memtx_engine_new(const char *snap_dirname, bool force_recovery,
 		 uint64_t tuple_arena_max_size, uint32_t objsize_min,
 		 bool dontdump, unsigned granularity,
-		 const char *allocator, float alloc_factor);
+		 const char *allocator, float alloc_factor,
+		 memtx_on_indexes_built_cb cb);
```
locker, line 221: Nit: Better name the argument `on_indexes_built`, because we may add other callbacks in future.

locker (approved):

src/box/memtx_engine.h
sergepetrenko: Ok:
```diff
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index bea6edda7..760f7c1ed 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -1263,7 +1263,7 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
                 uint64_t tuple_arena_max_size, uint32_t objsize_min,
                 bool dontdump, unsigned granularity,
                 const char *allocator, float alloc_factor,
-                memtx_on_indexes_built_cb cb)
+                memtx_on_indexes_built_cb on_indexes_built)
 {
        int64_t snap_signature;
        struct memtx_engine *memtx =
@@ -1370,7 +1370,7 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
        }
        tuple_format_ref(memtx->func_key_format);
 
-       memtx->on_indexes_built_cb = cb;
+       memtx->on_indexes_built_cb = on_indexes_built;
 
        fiber_start(memtx->gc_fiber, memtx);
        return memtx;
diff --git a/src/box/memtx_engine.h b/src/box/memtx_engine.h
index 8f259c5ca..76e6acec7 100644
--- a/src/box/memtx_engine.h
+++ b/src/box/memtx_engine.h
@@ -218,7 +218,7 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
                 uint64_t tuple_arena_max_size, uint32_t objsize_min,
                 bool dontdump, unsigned granularity,
                 const char *allocator, float alloc_factor,
-                memtx_on_indexes_built_cb cb);
+                memtx_on_indexes_built_cb on_indexes_built);
 
 int
 memtx_engine_recover_snapshot(struct memtx_engine *memtx,
@@ -351,12 +351,13 @@ memtx_engine_new_xc(const char *snap_dirname, bool force_recovery,
                    uint64_t tuple_arena_max_size, uint32_t objsize_min,
                    bool dontdump, unsigned granularity,
                    const char *allocator, float alloc_factor,
-                   memtx_on_indexes_built_cb cb)
+                   memtx_on_indexes_built_cb on_indexes_built)
 {
        struct memtx_engine *memtx;
        memtx = memtx_engine_new(snap_dirname, force_recovery,
                                 tuple_arena_max_size, objsize_min, dontdump,
-                                granularity, allocator, alloc_factor, cb);
+                                granularity, allocator, alloc_factor,
+                                on_indexes_built);
        if (memtx == NULL)
                diag_raise();
        return memtx;
```

src/box/box.cc
```diff
@@ -166,6 +166,62 @@ static struct fiber_pool tx_fiber_pool;
...
+	"indexes_built",
+	/* [RECOVERY_STATE_SYNCED] = */
+	"synced",
+};
+
+/** Whether the triggers for "synced" recovery stage have already run. */
+static bool recovery_state_synced_is_reached;
+
+int
+box_run_on_recovery_state(enum box_recovery_state state)
```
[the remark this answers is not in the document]
sergepetrenko: Sure:
```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 20ca17896..bd23f7ef4 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -214,7 +214,7 @@ static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
 /** Whether the triggers for "synced" recovery stage have already run. */
 static bool recovery_state_synced_is_reached;
 
-int
+static int
 box_run_on_recovery_state(enum box_recovery_state state)
 {
        assert(state >= 0 && state < box_recovery_state_MAX);
```

PR #7941, opened 2022-11-15
replication: fix local space writes failing when synchro queue isn't empty

### LINKED ISSUE #7592: Found uncommitted sync transactions from other instance with id 1
**Bug description**

While debugging a test we've encountered the error:

```
Found uncommitted sync transactions from other instance with id 1
```

It appears after the instance is restarted with an uncommitted sync transaction, during another transaction to a local space.

It looks related to https://github.com/tarantool/tarantool/issues/5440. We don't have to wait for rw, but we wait for leaving orphan mode.

Tarantool version: 2.10.0-29-g1ec2a9d15

**Steps to reproduce**

I only have a picodata-based repro: https://git.picodata.io/picodata/picodata/picodata/-/issues/129

The test is flaky. Sometimes it also reproduces another problem - https://github.com/tarantool/tarantool/issues/7590.

sergepetrenko:
Local spaces can be written to on any replica, even on a read-only one.
This makes sense, because local space data isn't replicated, so it can't
lead to a conflict or violate consistency anyhow.

However, when used together with synchronous replication, local spaces
can't be written to by anyone but the synchro queue owner:
```
tarantool> box.info.synchro.queue.len
---
- 1
...
tarantool> box.space.loc:replace{2}
---
- error: Found uncommitted sync transactions from other instance with id 1
...
```
Fix this and allow to put transactions touching local spaces to the
synchro queue even if it is claimed by someone else.

Note, we can't let local transactions bypass the synchro queue completely.
This would lead to consistency loss in a case when synchro queue
contains a transaction [sync_row, local_row], and another transcation
[local_row], probably based on the sync transaction, bypasses the limbo.

Closes #7592

NO_DOC=bugfix

### DIFF
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -112,7 +112,7 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
 	if  (limbo->owner_id == REPLICA_ID_NIL) {
 		diag_set(ClientError, ER_SYNC_QUEUE_UNCLAIMED);
 		return NULL;
-	} else if (limbo->owner_id != id) {
+	} else if (limbo->owner_id != id && !txn_is_fully_local(txn)) {
 		if (txn_limbo_is_empty(limbo)) {
 			diag_set(ClientError, ER_SYNC_QUEUE_FOREIGN,
 				 limbo->owner_id);
--- a/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
+++ b/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
@@ -1,0 +1,93 @@
+local t = require('luatest')
+local server = require('test.luatest_helpers.server')
+local cluster = require('test.luatest_helpers.cluster')
+local fiber = require('fiber')
+
+local g = t.group('gh-7592')
+
+g.before_each(function(cg)
+    cg.cluster = cluster:new{}
+    local box_cfg = {
+        replication_synchro_timeout = 1000,
+        replication_synchro_quorum = 3,
+        replication_timeout = 0.1,
+        election_mode = 'off',
+        replication = {
+            server.build_instance_uri('master'),
+            server.build_instance_uri('replica'),
+        },
+    }
+    cg.master = cg.cluster:build_and_add_server{
+        alias = 'master',
+        box_cfg = box_cfg,
+    }
+    cg.replica = cg.cluster:build_and_add_server{
+        alias = 'replica',
+        box_cfg = box_cfg,
+    }
+    cg.cluster:start()
+    cg.master:exec(function()
+        box.ctl.promote()
+        box.schema.space.create('sync', {is_sync = true})
+        box.space.sync:create_index('pk')
+        box.schema.space.create('loc', {is_local = true})
+        box.space.loc:create_index('pk')
+    end)
+    cg.replica:wait_vclock_of(cg.master)
+end)
+
+g.after_each(function(cg)
+    cg.cluster:drop()
+end)
+
+local function wait_synchro_queue_has_entries(server)
+    t.helpers.retrying({}, function()
+        local len = server:exec(function()
+            return box.info.synchro.queue.len
+        end)
+        t.assert(len > 0, 'Synchro queue is occupied')
+    end)
+end
+
+g.test_local_space_write_on_replica = function(cg)
+    -- Test that replica can write to local spaces when synchro queue is empty:
+    cg.replica:exec(function()
+        local t = require('luatest')
+        t.assert(box.info.synchro.queue.owner ~= 0, 'Queue is claimed')
+        t.assert(box.info.synchro.queue.owner ~= box.info.id,
+                 'Queue is foreign')
+        t.assert(box.info.synchro.queue.len == 0, 'Queue is empty')
+        local ok, _ = pcall(box.space.loc.insert, box.space.loc, {1})
+        t.assert(ok, 'Local space is writeable')
+    end)
+
+    -- Test that replica can write to local spaces when synchro queue is filled:
+    local f = fiber.new(function()
+        cg.master:exec(function()
+            box.space.sync:insert{1}
+        end)
+    end)
+    f:set_joinable(true)
+    wait_synchro_queue_has_entries(cg.replica)
+    local fid = cg.replica:exec(function()
+        local t = require('luatest')
+        t.assert(box.info.synchro.queue.len == 1, 'Queue is filled')
+        t.assert(box.info.synchro.queue.owner ~= box.info.id,
+                 'Queue is foreign')
+        local r_f = require('fiber').create(function()
+            box.space.loc:insert{2}
+        end)
+        r_f:set_joinable(true)
+        t.assert(box.info.synchro.queue.len == 2,
+                 'Local write is placed in the queue')
+        return r_f:id()
+    end)
+    cg.master:exec(function() box.cfg{replication_synchro_quorum = 2} end)
+    f:join()
+    cg.replica:exec(function(fid)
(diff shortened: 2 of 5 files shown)


kostja:
One can use a Lua table in a synchro transaction and thus "violate" consistency this way. It's a matter of the use case. In our use case local spaces should be always writable since they contain replica-local state, which we never mix with synchro state.

One way to support this use case is to allow all local spaces to bypass the synchro queue (my preference). Another is to have a switch on space or transaction level (always local or super local).

sergepetrenko:
> One can use a Lua table in a synchro transaction and thus "violate" consistency this way. It's a matter of the use case. In our use case local spaces should be always writable since they contain replica-local state, which we never mix with synchro state.

This is true, it's the matter of a use case. But the violation you put as an example is different. In your example the user is to blame. In case we let local space writes bypass the synchro queue it would be hard to blame the user for violating consistency by using local spaces. Basically we would have to say that local spaces can't be used in transactions with any other types of spaces. Writing "safe" code in tarantool would become even more complicated.

Another type of space -- "super local" sounds okay. But again, you won't be able to use "super local" spaces together with any other type of space in one transaction.

kostja:
You're just making it too difficult on yourself. The requirement you're talking about is imaginary - there is no user for it, specifically, nobody needs to be "saved" from consistency violation, there is no need to enforce consistency of transactions which use local spaces by forcing all local space transactions to go through the limbo queue. People know what they are doing. 
I'd suggest you do it when there is anyone asking you to do it - meanwhile leaving the path clear for the main use case, when local spaces are used for replica-local  writes which should always be availiable/reliable regardless of the cluster state.

kostja:
On a similar token, the current temporary table implementation today is broken - it replicates the temporary table definition, and so you can't create a temporary table on a read-only replica. A reasonable implementation would not replicate the definition and would work well in read only mode.

sergepetrenko:
>  meanwhile leaving the path clear for the main use case, when local spaces are used for replica-local writes which should always be availiable/reliable regardless of the cluster state.

But local space writes won't be reliable, see my example in https://github.com/tarantool/tarantool/issues/7592#issuecomment-1315357630.

And the rule "don't use local spaces together with any other type of space, if you have synchronous replication" sounds pretty complex. Even for people who know what they're doing.

Anyway, I'm not the user here. Are you saying you don't need the data from any other spaces when accessing local spaces?

For example, a on_replace trigger writing to local space on replica, would create a transaction touching both local and normal spaces. This looks like a quite likely use case for me.

sergepetrenko:
> A reasonable implementation would not replicate the definition and would work well in read only mode.

Good point about replicating the temp space definition.
BTW, you can write to temporary spaces just fine even when the synchro queue is non-empty.

src/box/txn_limbo.c
```diff
@@ -107,12 +107,12 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
...
 	}
-	if (id == 0)
-		id = instance_id;
+	/* id 0 is only possible for local transactions. */
+	assert(id != 0 || !txn_has_flag(txn, TXN_WAIT_ACK));
 	if  (limbo->owner_id == REPLICA_ID_NIL) {
 		diag_set(ClientError, ER_SYNC_QUEUE_UNCLAIMED);
 		return NULL;
-	} else if (limbo->owner_id != id) {
+	} else if (limbo->owner_id != id && id != REPLICA_ID_NIL) {
```
Gerold103, line 115: Wouldn't this patch be simpler?
```Diff
diff --git a/src/box/txn.h b/src/box/txn.h
index 9ac7441f3..b7252ed70 100644
--- a/src/box/txn.h
+++ b/src/box/txn.h
@@ -845,6 +845,13 @@ txn_commit_ro_stmt(struct txn *txn, struct txn_ro_savepoint *svp)
 bool
 txn_is_distributed(struct txn *txn);
 
+static inline bool
+txn_is_fully_local(const struct txn *txn)
+{
+       return txn->n_new_rows == txn->n_local_rows &&
+              txn->n_applier_rows == 0;
+}
+
 /**
  * End a statement. In autocommit mode, end
  * the current transaction as well.
index acdde6695..2d838413f 100644
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -112,7 +112,7 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
        if  (limbo->owner_id == REPLICA_ID_NIL) {
                diag_set(ClientError, ER_SYNC_QUEUE_UNCLAIMED);
                return NULL;
-       } else if (limbo->owner_id != id) {
+       } else if (limbo->owner_id != id && !txn_is_fully_local(txn)) {
                if (txn_limbo_is_empty(limbo)) {
                        diag_set(ClientError, ER_SYNC_QUEUE_FOREIGN,
                                 limbo->owner_id);
```
It allows to avoid origin id manipulations in txn.c.

Gerold103 (changes requested):
Thanks for the patch! I will leave it to the products to decide whether we should put local txns to the limbo or not. Reviewing only the tech patch. Also your commit title is a bit too long. Github shows it ugly-wrapped to a second line. I would suggest to make it <= 50 symbols.

src/box/txn_limbo.c
sergepetrenko: > Wouldn't this patch be simpler?

It would, thanks! Applied.

test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
```diff
@@ -0,0 +1,93 @@
...
+        box.space.loc:create_index('pk')
+    end)
+    cg.replica:wait_vclock_of(cg.master)
+end)
+
+g.after_each(function(cg)
+    cg.cluster:drop()
+end)
+
+local function wait_synchro_queue_is_busy(server)
```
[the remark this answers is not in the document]
sergepetrenko: > Besides, you don't really need this function at all. In its single usage case you could just do cg.replica:wait_vclock_ok(cg.master).

Not really. I'd also need to wait until lsn is bumped on master. So it would turn out uglier IMO.

Renamed the function though:
```diff
diff --git a/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua b/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
index 8cddd4a3f..a2af41de9 100644
--- a/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
+++ b/test/replication-luatest/gh_7592_local_write_with_syncro_test.lua
@@ -40,7 +40,7 @@ g.after_each(function(cg)
     cg.cluster:drop()
 end)
 
-local function wait_synchro_queue_is_busy(server)
+local function wait_synchro_queue_has_entries(server)
     t.helpers.retrying({}, function()
         local len = server:exec(function()
             return box.info.synchro.queue.len
@@ -68,7 +68,7 @@ g.test_local_space_write_on_replica = function(cg)
         end)
     end)
     f:set_joinable(true)
-    wait_synchro_queue_is_busy(cg.replica)
+    wait_synchro_queue_has_entries(cg.replica)
     local fid = cg.replica:exec(function()
         local t = require('luatest')
         t.assert(box.info.synchro.queue.len == 1, 'Queue is filled')  
```

sergepetrenko:
@Gerold103, thanks for the review!

Speaking of "going through the limo or not" here's @Mons's opinion:
> Мне кажется, нам не стоит нарушать принятые ранее гарантии
Если используется синхра, то все идут через лимб. Даже локальные спейсы.
В том порядке, в котором они создаются.

Gerold103 (approved):
Thanks for the patch! Looking good now.

sergepetrenko:
@Gerold103, could you push this to master and 2.10, please?

PR #7944, opened 2022-11-15
Make `ER_WRONG_INDEX_PARTS` and `ER_WRONG_SPACE_FORMAT` messages more verbose

### LINKED ISSUE #7933: Wrong index part number isn't reported in error message
Reproduced with Tarantool 2.11.0-entrypoint-699-g5d76efe708af. Tarantool 2.10 is affected as well.

This is a minor verbosity issue: if an index part supplied by the user is wrong, its number isn't reported, which complicates investigation:

```
tarantool> box.space.test:format({{'foo', type = 's'}}) -- error is ok: field no is present
---
- error: 'Wrong space format: field 1 has unknown field type'
...

tarantool> box.space.test:create_index('pk', {parts = {1, 's'}}) -- bad error: part no is missing
- error: 'Wrong index parts: unknown field type'
...
```

Reporting the wrong index part could be useful for pin pointing the error causing #7841.

locker:
Let's include part and field number into them to make it easier to pinpoint the problem.
To achieve that, we have to rework the way the `opts_decode` function sets errors.

Closes #7933

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -213,9 +213,11 @@ index_opts_decode(struct index_opts *opts, const char *map,
 		  struct region *region)
 {
 	index_opts_create(opts);
-	if (opts_decode(opts, index_opts_reg, &map, ER_WRONG_INDEX_OPTIONS,
-			region) != 0)
+	if (opts_decode(opts, index_opts_reg, &map, region) != 0) {
+		diag_set(ClientError, ER_WRONG_INDEX_OPTIONS,
+			 diag_last_error(diag_get())->errmsg);
 		return -1;
+	}
 	if (opts->distance == rtree_index_distance_type_MAX) {
 		diag_set(ClientError, ER_WRONG_INDEX_OPTIONS,
 			 "distance must be either 'euclid' or 'manhattan'");
@@ -389,8 +391,12 @@ space_opts_decode(struct space_opts *opts, const char *map,
 		  struct region *region)
 {
 	space_opts_create(opts);
-	return opts_decode(opts, space_opts_reg, &map, ER_WRONG_SPACE_OPTIONS,
-			   region);
+	if (opts_decode(opts, space_opts_reg, &map, region) != 0) {
+		diag_set(ClientError, ER_WRONG_SPACE_OPTIONS,
+			 diag_last_error(diag_get())->errmsg);
+		return -1;
+	}
+	return 0;
 }
 
 /**
@@ -3340,9 +3346,11 @@ func_def_new_from_tuple(struct tuple *tuple)
 		}
 		def->param_count = argc;
 		const char *opts = tuple_field(tuple, BOX_FUNC_FIELD_OPTS);
-		if (opts_decode(&def->opts, func_opts_reg, &opts,
-				ER_WRONG_SPACE_OPTIONS, NULL) != 0)
+		if (opts_decode(&def->opts, func_opts_reg, &opts, NULL) != 0) {
+			diag_set(ClientError, ER_WRONG_SPACE_OPTIONS,
+				 diag_last_error(diag_get())->errmsg);
 			return NULL;
+		}
 	} else {
 		/* By default export to Lua, but not other frontends. */
 		def->exports.lua = true;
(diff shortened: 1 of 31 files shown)


src/box/space_upgrade.c
```diff
@@ -18,23 +18,48 @@
...
+	 * a non-NULL ptr here and abort later, in space_upgrade_check_alter.
+	 */
+	mp_next(data);
+	return BAD_PTR;
+}
+
+struct space_upgrade_def *
+space_upgrade_def_dup(const struct space_upgrade_def *def)
+{
+	if (def == BAD_PTR)
```
Gumix, line 38: I'm not sure if it's legal in modern C.

[C11 § 6.5.9 paragraph 6](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf) (pointed by @drewdzzz):
> Two pointers compare equal if and only if both are null pointers, both are pointers to the same object (including a pointer to an object and a subobject at its beginning) or function, both are pointers to one past the last element of the same array object, or one is a pointer to one past the end of one array object and the other is a pointer to the start of a different array object that happens to immediately follow the first array object in the address space.)

src/box/alter.cc
```diff
@@ -3340,9 +3346,11 @@ func_def_new_from_tuple(struct tuple *tuple)
 		}
 		def->param_count = argc;
 		const char *opts = tuple_field(tuple, BOX_FUNC_FIELD_OPTS);
-		if (opts_decode(&def->opts, func_opts_reg, &opts,
-				ER_WRONG_SPACE_OPTIONS, NULL) != 0)
+		if (opts_decode(&def->opts, func_opts_reg, &opts, NULL) != 0) {
+			diag_set(ClientError, ER_WRONG_SPACE_OPTIONS,
```
Gumix, line 3350: Maybe it should be `ER_WRONG_FUNCTION_OPTIONS`?

src/box/space_upgrade.c
locker: Sounds fair. Replaced `BAD_PTR` with `dummy_def`.

```diff
diff --git a/src/box/space_upgrade.c b/src/box/space_upgrade.c
index 3c3fe6b42c82..6d6b8b59584b 100644
--- a/src/box/space_upgrade.c
+++ b/src/box/space_upgrade.c
@@ -18,7 +18,7 @@
 # error unimplemented
 #endif
 
-#define BAD_PTR ((void *)-1)
+static struct space_upgrade_def {} dummy_def;
 
 struct space_upgrade_def *
 space_upgrade_def_decode(const char **data, struct region *region)
@@ -29,22 +29,20 @@ space_upgrade_def_decode(const char **data, struct region *region)
 	 * a non-NULL ptr here and abort later, in space_upgrade_check_alter.
 	 */
 	mp_next(data);
-	return BAD_PTR;
+	return &dummy_def;
 }
 
 struct space_upgrade_def *
 space_upgrade_def_dup(const struct space_upgrade_def *def)
 {
-	if (def == BAD_PTR)
-		return BAD_PTR;
-	assert(def == NULL);
-	return NULL;
+	assert(def == NULL || def == &dummy_def);
+	return def != NULL ? &dummy_def : NULL;
 }
 
 void
 space_upgrade_def_delete(struct space_upgrade_def *def)
 {
-	assert(def == NULL || def == BAD_PTR);
+	assert(def == NULL || def == &dummy_def);
 	(void)def;
 }
 
@@ -53,7 +51,7 @@ space_upgrade_check_alter(struct space *space, struct space_def *new_def)
 {
 	assert(space->upgrade == NULL);
 	assert(new_def->opts.upgrade_def == NULL ||
-	       new_def->opts.upgrade_def == BAD_PTR);
+	       new_def->opts.upgrade_def == &dummy_def);
 	(void)space;
 	if (new_def->opts.upgrade_def != NULL) {
 		diag_set(ClientError, ER_UNSUPPORTED,
```
</details>

src/box/alter.cc
locker: Yep, it definitely shouldn't be `ER_WRONG_SPACE_OPTIONS`, but I don't want to change this in the scope of this refactoring. Also, this code isn't even covered... I filed a separate ticket to address this issue later, see #7972.

Gumix (approved):

PR #7950, opened 2022-11-17
box: fix rollback of read-only transaction statements

### LINKED ISSUE #5501: Transaction object corruption on rare errors
I managed to corrupt transaction object `struct txn` by imitation of an out-of-memory error when an iterator is created. Here is a patch to make it easy to reproduce:
```
diff --git a/src/box/memtx_tree.cc b/src/box/memtx_tree.cc
index 44bdc86e6..ed75ac25b 100644
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -1202,6 +1202,11 @@ memtx_tree_index_create_iterator(struct index *base, enum iterator_type type,
 		key = NULL;
 	}
 
+	ERROR_INJECT(ERRINJ_INDEX_ITERATOR_NEW, {
+		diag_set(ClientError, ER_INJECTION, "iterator fail");
+		return NULL;
+	});
+
 	struct tree_iterator<USE_HINT> *it = (struct tree_iterator<USE_HINT> *)
 		mempool_alloc(&memtx->iterator_pool);
 	if (it == NULL) {
index a075c9ef0..8cbd28f6b 100644
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -441,6 +441,7 @@ txn_commit_stmt(struct txn *txn, struct request *request)
 		}
 	}
 	--txn->in_sub_stmt;
+	assert(txn->in_sub_stmt >= 0);
 	return 0;
 fail:
 	txn_rollback_stmt(txn);
@@ -959,6 +960,7 @@ rollback:
 void
 txn_rollback_stmt(struct txn *txn)
 {
+	assert(txn->in_sub_stmt >= 0);
 	if (txn == NULL || txn->in_sub_stmt == 0)
 		return;
 	txn->in_sub_stmt--;
index 814c57c2e..df8b78a9b 100644
--- a/src/lib/core/errinj.h
+++ b/src/lib/core/errinj.h
@@ -149,6 +149,7 @@ struct errinj {
 	_(ERRINJ_AUTO_UPGRADE, ERRINJ_BOOL, {.bparam = false})\
 	_(ERRINJ_COIO_WRITE_CHUNK, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_APPLIER_SLOW_ACK, ERRINJ_BOOL, {.bparam = false}) \
+	_(ERRINJ_INDEX_ITERATOR_NEW, ERRINJ_BOOL, {.bparam = false}) \
 
 ENUM0(errinj_id, ERRINJ_LIST);
 extern struct errinj errinjs[];
```
The patch imitates an error in `memtx_tree_index_create_iterator` function.

Now here is a test to crash in one of the assertions above:
```
box.cfg{}
s1 = box.schema.create_space('test1')
_ = s1:create_index('pk')

s2 = box.schema.create_space('test2')
_ = s2:create_index('pk')

s2:on_replace(function() pcall(s1.select, s1) end)

function test()
    box.error.injection.set("ERRINJ_INDEX_ITERATOR_NEW", true)

    box.begin()
    s1:replace{1}
    s2:replace{2}
    box.commit()

    box.error.injection.set("ERRINJ_INDEX_ITERATOR_NEW", false)
end

test()
```

If we remove the assertions, the transaction fails with a very strange error from `box.commit()`: 
```
---
- error: Can not commit transaction in a nested statement
...
```

The issue is this whole idea of 'ro' transactions and `txn_begin_ro_stmt`/`txn_commit_ro_stmt` functions. Existence of `txn_commit_ro_stmt` assumes there is `txn_rollback_ro_stmt`, but it does not exist. Instead, in the code we use `txn_rollback_stmt`, which has nothing to do with 'ro' statements, and will just rollback the currently running normal statement.

As an example, consider `box_index_iterator`. It calls `txn_begin_ro_stmt`, which does not produce any artefacts in an existing `struct txn`. But in case of iteration creation fail it calls `txn_rollback_ro_stmt`, which will rollback the currently being executed 'rw' statement.

A solution would be to stop calling `txn_rollback_stmt` in case of an ro statement fail. However then it would become inconsistent, because there is `txn_commit_ro_stmt`, but no any rollback. So I think we also should drop `txn_commit_ro_stmt` and introduce `txn_end_ro_stmt`. It is supposed to be called regardless of ro statement result. Because there is no such thing as read-commit or read-rollback anyway.

The issue is related to https://github.com/tarantool/tarantool/issues/5427, and perhaps can be fixed in scope of that issue.

Gumix:
Existence of `txn_commit_ro_stmt` assumes there is `txn_rollback_ro_stmt`, but it does not exist. Instead, currently we use `txn_rollback_stmt`, which has nothing to do with 'ro' statements, and will just rollback the currently running normal statement.

As an example, consider `box_index_iterator`. It calls `txn_begin_ro_stmt`, which does not produce any artefacts in an existing `struct txn`. But in case of the iterator creation fail it calls `txn_rollback_stmt`, which will rollback the currently being executed 'rw' statement.

Drop `txn_commit_ro_stmt` and introduce `txn_end_ro_stmt`, which is called regardless of 'ro' statement result.

Closes #5501


test/engine-luatest/gh_5501_errinj_index_iterator_new_test.lua
```diff
@@ -0,0 +1,36 @@
...
+    local engine = cg.params.engine
+
+    cg.server:exec(function(engine)
+        local s = box.schema.space.create('test', {engine = engine})
+        s:create_index('pk')
+
+        -- This s:select{0} will fail because of the error injection.
+        s:on_replace(function() pcall(s.select, s, 0) end)
+
+        box.error.injection.set("ERRINJ_INDEX_ITERATOR_NEW", true)
```
locker, line 30: Can we make `iterator_create()` fail without introducing a new error injection? By passing an invalid key or iterator type, may be?

test/engine-luatest/gh_5501_errinj_index_iterator_new_test.lua
```diff
@@ -0,0 +1,36 @@
...
+-- or "Can not commit transaction in a nested statement".
+g.test_errinj = function(cg)
+    local engine = cg.params.engine
+
+    cg.server:exec(function(engine)
+        local s = box.schema.space.create('test', {engine = engine})
+        s:create_index('pk')
+
+        -- This s:select{0} will fail because of the error injection.
+        s:on_replace(function() pcall(s.select, s, 0) end)
```
locker, line 28: Let's add test cases for other functions, as well. We have just a just few of them and all of them seem to be affected to the very same issue and can be tested in a similar fashion: `min()`, `max()`, `get()`, `count()`, `pairs()`.

test/engine-luatest/gh_5501_errinj_index_iterator_new_test.lua
Gumix: No

locker (approved):

locker:
Cherry-picked to 2.10.

PR #7952, opened 2022-11-17
flaky: follow-up for qsync_basic test

sergos:
Two more places where replica didn't wait until data is replicated.

Follow-up to tarantool/tarantool-qa#274

NO_CHANGELOG=test fix
NO_DOC=test fix

### DIFF
--- a/test/replication/qsync_basic.test.lua
+++ b/test/replication/qsync_basic.test.lua
@@ -247,6 +247,7 @@ test_run:switch('default')
 box.cfg{replication_synchro_timeout = 1000}
 for i = 1, 100 do box.space.sync:replace{i} end
 test_run:cmd('switch replica')
+test_run:wait_lsn('replica', 'default')
 box.space.sync:count()
 -- Rows could be corrupted during WAL writes. Restart should
 -- reveal the problem during recovery.
@@ -255,6 +256,7 @@ box.space.sync:count()
 test_run:cmd('switch default')
 for i = 1, 100 do box.space.sync:delete{i} end
 test_run:cmd('switch replica')
+test_run:wait_lsn('replica', 'default')
 box.space.sync:count()
 
 --
--- a/test/replication/qsync_basic.result
+++ b/test/replication/qsync_basic.result
@@ -636,6 +636,9 @@ test_run:cmd('switch replica')
  | ---
  | - true
  | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
+ | ...
 box.space.sync:count()
  | ---
  | - 100
@@ -658,6 +661,9 @@ for i = 1, 100 do box.space.sync:delete{i} end
 test_run:cmd('switch replica')
  | ---
  | - true
+ | ...
+test_run:wait_lsn('replica', 'default')
+ | ---
  | ...
 box.space.sync:count()
  | ---


NickVolynkin:
Please fix the issue link in the commit message:

```diff
- tarantool-qa#274
+ tarantool/tarantool-qa#274
```

test/replication/qsync_basic.test.lua
```diff
@@ -247,6 +247,7 @@ test_run:switch('default')
 box.cfg{replication_synchro_timeout = 1000}
 for i = 1, 100 do box.space.sync:replace{i} end
 test_run:cmd('switch replica')
+test_run:wait_lsn('replica', 'default')
```
sergepetrenko, line 250: There's `replication_synchro_quorum = 2` left on master from the previous test case (see lines 228, 235).
So `for i = 1, 100 do box.space.sync:replace{i}` will succeed only after the data is really replicated to the replica.

There is no memtx mvcc, so replica should see all 100 entries even without this `wait_lsn`.
If it doesn't, the issue is somewhere else. Not in the missing `wait_lsn`.

test/replication/qsync_basic.test.lua
```diff
@@ -255,6 +256,7 @@ box.space.sync:count()
 test_run:cmd('switch default')
 for i = 1, 100 do box.space.sync:delete{i} end
 test_run:cmd('switch replica')
+test_run:wait_lsn('replica', 'default')
```
sergepetrenko, line 259: Same about this `wait_lsn`. The quorum's still 2 on master and no mvcc. Replica should see all 100 deletes as soon as they complete.

sergepetrenko (changes requested):
Thanks for the patch!

Although I've seen the failures you're trying to fix (count equal to 99 or less instead of 100), I don't understand how these changes could help. Please see more detailed comments below.

Also I've run `./test-run.py $(yes qsync_basic | head -n 512)` both with and without your patch, and saw no failures. So can't check if your patch really fixes them or not.

sergepetrenko:
By the way, judging by [monitoring](https://mon.tarantool.io/grafana/d/vQ35ub74k/test-statistics?orgId=1&var-bucket=multivac-tests&var-bucket_table=multivac-table&var-job_name=All&var-test_name=replication%2Fqsync_basic.test.lua&var-branch=All&var-os_version=All&var-architecture=All&var-gc64=All&var-compiler_version=All&var-libc_version=All&var-debug=All&var-configuration=All&var-test_type=All&var-test_subtype=All&var-test_attempt=All&var-period=1d&var-tag=_start&from=now-7d&to=now), it fails only on `vinyl` configuration. Seems my assertions stand for memtx. I wonder what could be so special about vinyl...

sergepetrenko (approved):
After noticing the test is flaky only in vinyl and consulting @locker, we learned that vinyl now uses 'best-effort' isolation by default, which's 'read-confirmed' (not 'read-committed') for read-only transactions. The corresponding change was made in 588170a77c9aba3d0cf898d0134286c6ffd46339. This explains why the test started failing more around these dates.

So the fix is correct. We need to wait until CONFIRM entry is replicated to the replica, hence the `wait_lsn` in both places.

test/replication/qsync_basic.test.lua
sergepetrenko: Never mind, I see now that the test only fails in vinyl. Because vinyl now has 'best-effort' isolation by default, so there are no dirty reads on replica. The fix's correct.

PR #7957, opened 2022-11-21
msgpack: add `__index` metamethod and `get` method to `msgpack.object`

### LINKED ISSUE #7898: `msgpack.object` indexation
Add indexation to `msgpack.object` from [`msgpack`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/msgpack/) similar to `tuple` from [`box.tuple`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_tuple/).

# Interface
New `msgpack.object:index` method with the following semantics:
* a number or string key can be passed just like for `tuple_object`;
* when indexing by number:
	* if the **MsgPack** is an array — return the element at index;
	* if the **MsgPack** is a map    — return the element with key;
* `nil` is returned when no element is found or the MsgPack type is not indexable;

# Examples
```lua
mp = msgpack.object{key = 1, [2] = 2, decode = 3}
mp['key']
-- 1

mp['does not exist']
-- nil

mp = msgpack.object{'a', 'b', 'c'}
mp[2]
-- 'b'

mp[0]
-- nil

mp[4]
-- nil

mp[-1]
-- nil

mp[2]
-- 2

mp[true]
-- nil

mp['decode']
-- function

mp:get('decode')
-- 3

mp = msgpack.object(1)

mp[1]
-- nil

mp['key']
-- nil
```

# Enhancements
* eager **MsgPack** decoding on the first indexation, such that next indexations do not require decoding;
* lazy breadth-first **MsgPack** decoding on indexation.

CuriousGeorgiy:
Closes #7898

### DIFF
--- a/changelogs/unreleased/gh-7898-msgpack-object-indexation.md
+++ b/changelogs/unreleased/gh-7898-msgpack-object-indexation.md
@@ -1,0 +1,6 @@
+## feature/lua/msgpack
+
+* Added `__index` metamethod and `get` method to `msgpack.object` which perform
+  indexation of MsgPack data stored in the object similar to `tuple` from
+  `box.tuple`: `__index` resolves collisions in favor of
+  `msgpack.object` methods, whereas `get` ignores methods (gh-7898).
--- a/src/lua/msgpack.c
+++ b/src/lua/msgpack.c
@@ -72,6 +72,13 @@ struct luamp_object {
 	const char *data;
 	/** Pointer to the end of msgpack data. */
 	const char *data_end;
+	/**
+	 * Upon first indexation the MsgPack data is completely decoded,
+	 * pushed to Lua stack and referenced: the Lua stack reference is saved
+	 * to this field.
+	 * Initially set to `LUA_NOREF`.
+	 */
+	int decoded_ref;
 };
 
 static const char luamp_object_typename[] = "msgpack.object";
@@ -665,6 +672,7 @@ luamp_new_object(struct lua_State *L, size_t data_len)
 	obj->data_ref = LUA_NOREF;
 	obj->data = (char *)obj + sizeof(*obj);
 	obj->data_end = obj->data + data_len;
+	obj->decoded_ref = LUA_NOREF;
 	luaL_getmetatable(L, luamp_object_typename);
 	lua_setmetatable(L, -2);
 	return obj;
@@ -766,6 +774,7 @@ luamp_object_gc(struct lua_State *L)
 	struct luamp_object *obj = luamp_check_object(L, 1);
 	luaL_unref(L, LUA_REGISTRYINDEX, obj->cfg_ref);
 	luaL_unref(L, LUA_REGISTRYINDEX, obj->data_ref);
+	luaL_unref(L, LUA_REGISTRYINDEX, obj->decoded_ref);
 	return 0;
 }
 
@@ -806,6 +815,67 @@ luamp_object_iterator(struct lua_State *L)
 	lua_setmetatable(L, -2);
 	lua_insert(L, 1);
 	it->source_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+	return 1;
+}
+
+/**
+ * Takes a `msgpack.object` and an indexation key as the arguments, indexes
+ * the MsgPack stored in the `msgpack.object` and pushes the result to Lua stack
+ * or, if the MsgPack data type is not indexable, pushes nil.
+ */
+static int
+luamp_object_get(struct lua_State *L)
+{
+	struct luamp_object *obj = luamp_check_object(L, 1);
+	enum mp_type type = mp_typeof(*obj->data);
+	if (type != MP_MAP && type != MP_ARRAY)
+		return luaL_error(L, "not an array or map");
+	if (obj->decoded_ref == LUA_NOREF) {
+		const char *data = obj->data;
+		luamp_decode(L, obj->cfg, &data);
+		assert(data == obj->data_end);
+		obj->decoded_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+	}
+	/* Pushes the decoded MsgPack data on top of the stack. */
+	lua_rawgeti(L, LUA_REGISTRYINDEX, obj->decoded_ref);
+	/* Pushes the indexing key on top of the stack. */
+	lua_pushvalue(L, -2);
+	/* Indexes the decoded MsgPack data and pops the key. */
+	lua_rawget(L, -2);
+	return 1;
+}
+
+/**
+ * Takes a `msgpack.object` and an indexation key as the arguments: if the key
+ * is of string type, first tries to match it with `msgpack.object` methods,
+ * and, in case the match occurs, pushes the matched method to Lua stack —
+ * otherwise, delegates indexation to `msgpack.object:get`.
+ */
+static int
+luamp_object_index(struct lua_State *L)
+{
+	luamp_check_object(L, 1);
+	if (lua_type(L, 2) != LUA_TSTRING)
+		return luamp_object_get(L);
+	lua_getmetatable(L, 1);
+	lua_pushvalue(L, 2);
+	lua_rawget(L, -2);
+	if (lua_isnil(L, -1)) {
+		/* Pop the nil and the metatable. */
+		lua_pop(L, 2);
+		return luamp_object_get(L);
+	}
+	return 1;
+}
+
+/**
+ * Push a table of `msgpack.object` methods for console autocompletion.
+ */
+static int
+luamp_object_autocomplete(struct lua_State *L)
+{
+	luamp_check_object(L, 1);
+	lua_getmetatable(L, 1);
 	return 1;
 }
 
@@ -1015,8 +1085,11 @@ luaopen_msgpack(lua_State *L)
 	static const struct luaL_Reg luamp_object_meta[] = {
 		{ "__gc", luamp_object_gc },
 		{ "__tostring", luamp_object_tostring },
+		{ "__index", luamp_object_index },
+		{ "__autocomplete", luamp_object_autocomplete },
 		{ "decode", luamp_object_decode },
 		{ "iterator", luamp_object_iterator },
+		{ "get", luamp_object_get },
 		{ NULL, NULL }
 	};
 	luaL_register_type(L, luamp_object_typename, luamp_object_meta);
--- a/test/app-luatest/msgpack_test.lua
+++ b/test/app-luatest/msgpack_test.lua
@@ -1,4 +1,5 @@
 local buffer = require('buffer')
+local console = require('console')
 local msgpack = require('msgpack')
 local ffi = require('ffi')
 local t = require('luatest')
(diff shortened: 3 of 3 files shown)


locker:
> @locker
> 
> Could we never overlap methods, but give a user an additional method to get any object from msgpack?
> 
> ```
> o.decode -- function
> o:get('decode') -- foo
> ```
> 
> I dislike constructor options because from time to time we need to build the object inside our API and don't know what variant is preferrable for a user.

@unera The `o:get()` method sounds fine to me. Do we need to override `__index` for fields that aren't defined in the MsgPack object metatable, so that we can use a MsgPack object passed to `box.iproto.override` (almost) like a plain table, without using `get`?

```Lua
box.iproto.override(box.iproto.SELECT, function(sid, header, body)
    box.iproto.send(sid, {
        request_type = box.iproto.OK,
        sync = header.sync, -- alias for header:get('sync')
    }, {data = {1, 2, 3}})
end)
```

That is,
 - `get` always returns the value from the MsgPack object.
 - `__index` returns the value from the MsgPack object metatable if it exists, otherwise calls `get`.

changelogs/unreleased/gh-7898-msgpack-object-indexation.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua/msgpack
+
+* Added `msgpack.object:index` method which performs indexation similar to
```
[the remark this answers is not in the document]
CuriousGeorgiy: > It'd also be nice to have the fields auto-completed in console:

Basically, AFAI, `__autocomplete` is equivalent to `__serialize` which, in turn, is equivalent to `decode` — wouldn't that be too expensive?

test/app-luatest/msgpack_test.lua
```diff
@@ -445,3 +445,67 @@ g.test_object_misc = function()
     t.assert_not(msgpack.is_object(it))
     t.assert_not(msgpack.is_object({mp}))
 end
+
+-- Checks that `msgpack.object:get` method works correctly.
+g.test_object_get = function()
+    local array = {nil, true, 1, '2', 3.0, {1, 2, 3}, {a = 1, b = 2, c = 3}, 'decode', 'iterator', 'get'}
```
locker, line 451: This isn't an array from Lua point of view - it's a map, because of the nil. The code testing arrays isn't covered at all: https://coveralls.io/jobs/110054477/source_files/28263573271#L819

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+		}
+		case MP_INT: {
+			int64_t probe = mp_decode_int(&it);
+			if (is_signed) {
+				if ((int64_t)key == probe) {
+					luamp_decode(L, obj->cfg, &it);
+					return 1;
+				}
+			} else {
+				if (probe >= 0 && key == (uint64_t)probe) {
```
locker, line 864: This code isn't covered: https://coveralls.io/jobs/110054477/source_files/28263573271#L864

locker (changes requested):

changelogs/unreleased/gh-7898-msgpack-object-indexation.md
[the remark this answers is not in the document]
locker: > AFAI, `__autocomplete` is equivalent to `__serialize` which, in turn, is equivalent to `decode`

No, we just need to iterate over all keys in the map.

> wouldn't that be too expensive?

It's okay, because it's only used in console. Autocompleting MsgPack fields may be optional, but we do need to autocomplete keys from the object metatable, because this works now:

```
tarantool> o = require('msgpack').object({})
---
...

tarantool> o.
o.decode(    o.iterator(
```

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+luamp_object_get_integer(struct lua_State *L, struct luamp_object *obj,
+			 uint64_t num, bool is_signed)
+{
+	switch (mp_typeof(*obj->data)) {
+	case MP_MAP:
+		return luamp_object_get_integer_map(L, obj, num, is_signed);
+	case MP_ARRAY:
+		return luamp_object_get_integer_array(L, obj, num, is_signed);
+	default:
+		return luaL_error(L, "not an array or map");
```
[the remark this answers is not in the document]
CuriousGeorgiy: > Looks like this error is never raised because of the check in `luamp_object_get`.

Yeah, I forgot to remove it, should be `unreachable`.

> Can you place all the checks at one level for the sake of symmetry?

Fixed: moved all type compatibility checks to `luamp_object_get`.

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+	uint32_t size = mp_decode_map(&it);
+	for (uint32_t i = 0; i < size; ++i, mp_next(&it)) {
+		if (mp_typeof(*it) != MP_STR) {
+			mp_next(&it);
+			continue;
+		}
+
+		uint32_t probe_len;
+		const char *probe = mp_decode_str(&it, &probe_len);
+		if (strncmp(key, probe, MIN(key_len, probe_len)) == 0) {
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed: added a check for the length equality and a test case for this.

test/app-luatest/msgpack_test.lua
CuriousGeorgiy: > it's a map, because of the nil

Didn't know about that, thanks!

> The code testing arrays isn't covered at all: [coveralls.io/jobs/110054477/source_files/28263573271#L819](https://coveralls.io/jobs/110054477/source_files/28263573271#L819)

Thanks for pointing this out, I will also check coverage next time when submitting a patch.

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+						key < 0);
+	}
+	case LUA_TSTRING: {
+		size_t len;
+		const char *key = lua_tolstring(L, 2, &len);
+		if (len == 0)
+			return 0;
+		return luamp_object_get_string(L, obj, key, len);
+	}
+	default:
```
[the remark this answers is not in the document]
CuriousGeorgiy: AFAIC, @unera only wants integers and strings. 

And `uuid` as a follow-up (issue about this to be opened).

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+		return luamp_object_get(L);
+
+	size_t key_len;
+	const char *key = lua_tolstring(L, 2, &key_len);
+	const struct luaL_Reg *reg = luamp_object_methods;
+	for (size_t i = 0; i < lengthof(luamp_object_methods); ++reg, ++i) {
+		if (strcmp("__", reg->name) == 0)
+			continue;
+		if (strcmp(reg->name, key) == 0) {
+			lua_pushcfunction(L, reg->func);
```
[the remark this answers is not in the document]
CuriousGeorgiy: AFAIC, getting looking up methods from the metatable should be sufficient — rewrote this snippet accordingly.

> Anyway, I don't understand why you implement this loop.

I guess I just don't have enough experience working with the Lua environment from C.

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+{
+	if (lua_type(L, 2) != LUA_TSTRING)
+		return luamp_object_get(L);
+
+	size_t key_len;
+	const char *key = lua_tolstring(L, 2, &key_len);
+	const struct luaL_Reg *reg = luamp_object_methods;
+	for (size_t i = 0; i < lengthof(luamp_object_methods); ++reg, ++i) {
+		if (strcmp("__", reg->name) == 0)
+			continue;
```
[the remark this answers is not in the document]
CuriousGeorgiy: I initially iterated over the the array of metamethods.

Rewrote this code to use metatable lookup.

src/lua/msgpack.c
CuriousGeorgiy: Fixed: added a test-case for this branch.

src/lua/msgpack.c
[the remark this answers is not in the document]
unera: SQL team has provided all supported types for similar operators.

But I think we should provide at least uuid. The other types can be placed on a wishlist.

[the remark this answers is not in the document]
unera: `msgpack.decode` returns usual lua object:
- scalar (string, number, nil
- or table

I think, that after decoding, we shouldn't do anything for keytypechecking.

A user ask you: give me a value 'foobar' from the table. 
We return him: table['foobar']

If the container has a wrong type (not table) or if the key has a wrong type (not scalar) we can catch an exception and return nil.

src/lua/msgpack.c
```diff
@@ -809,6 +811,190 @@ luamp_object_iterator(struct lua_State *L)
...
+ * the MsgPack stored in the `msgpack.object` and pushes the result to Lua stack
+ * or throws an error.
+ */
+static int
+luamp_object_get(struct lua_State *L)
+{
+	struct luamp_object *obj = luamp_check_object(L, 1);
+
+	enum mp_type type = mp_typeof(*obj->data);
+	if (type != MP_MAP && type != MP_ARRAY)
```
unera, line 933: lua_pushnil(L)

src/lua/msgpack.c
```diff
@@ -809,6 +811,190 @@ luamp_object_iterator(struct lua_State *L)
...
+		return luaL_error(L, "not an array or map");
+
+	switch (lua_type(L, 2)) {
```
unera, line 936: lua_rawget(L, -2) + catch error &  return nil if error

changelogs/unreleased/gh-7898-msgpack-object-indexation.md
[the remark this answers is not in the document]
CuriousGeorgiy: Added `__autocomplete` metamethod.

src/lua/msgpack.c
```diff
@@ -1013,10 +1196,9 @@ LUALIB_API int
 luaopen_msgpack(lua_State *L)
 {
 	static const struct luaL_Reg luamp_object_meta[] = {
-		{ "__gc", luamp_object_gc },
+		{ "__gc",       luamp_object_gc },
```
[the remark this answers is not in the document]
CuriousGeorgiy: Sorry, changed it by accident: dropped from diff.

src/lua/msgpack.c
```diff
@@ -809,6 +811,187 @@ luamp_object_iterator(struct lua_State *L)
...
+	struct luamp_object *obj = luamp_check_object(L, 1);
+
+	enum mp_type type = mp_typeof(*obj->data);
+	if (type != MP_MAP && type != MP_ARRAY)
+		return luaL_error(L, "not an array or map");
+
+	switch (lua_type(L, 2)) {
+	case LUA_TNUMBER: {
+		double key = lua_tonumber(L, 2);
+		if (ceil(key) != key && floor(key) != key)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Rewrote the patch, comparison is now delegated to Lua.

src/lua/msgpack.c
CuriousGeorgiy: We can simply `return 0;` here, which is equivalent.

Changed behaviour: now attempt to index non-indexable MsgPack data returns `nil`.

src/lua/msgpack.c
CuriousGeorgiy: `lua_rawget` does not throw errors. 

Rewrote the patch: now we do `lua_rawget` instead of explicit indexing.

src/lua/msgpack.c
```diff
@@ -72,6 +74,8 @@ struct luamp_object {
 	const char *data;
 	/** Pointer to the end of msgpack data. */
 	const char *data_end;
+	/** Reference to decoded MsgPack data. */
```
locker, line 77: Nit: From the comment it's unclear what kind of reference it is, what it equals to initially, and when it's set. Please clarify these points.

src/lua/msgpack.c
```diff
@@ -809,6 +815,80 @@ luamp_object_iterator(struct lua_State *L)
...
+	const char *data = obj->data;
+	luamp_decode(L, obj->cfg, &data);
+	assert(data == obj->data_end);
+	obj->decoded_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+}
+
+/**
+ * Takes a `msgpack.object` and an indexation key as the arguments, indexes
+ * the MsgPack stored in the `msgpack.object` and pushes the result to Lua stack
+ * or throws an error.
```
locker, line 836: Nit: Can it really throw an error? In what case?

src/lua/msgpack.c
```diff
@@ -809,6 +815,80 @@ luamp_object_iterator(struct lua_State *L)
...
+ * otherwise, delegates indexation to `msgpack.object:get`.
+ */
+static int
+luamp_object_index(struct lua_State *L)
+{
+	if (lua_type(L, 2) != LUA_TSTRING)
+		return luamp_object_get(L);
+
+	lua_getmetatable(L, 1);
+	lua_pushvalue(L, 2);
```
locker, line 870: This may crash if misused:

```
tarantool> require('msgpack').object(123).__index(123, 'foo')
tarantool: ./third_party/luajit/src/lj_api.c:840: lua_rawget: Assertion `(((t)->it) == (~11u))' failed.
Aborted
```

Please fix.

CuriousGeorgiy: Fixed: check that the first argument is `msgpack.object` in `__index` and `__autocomplete`.

Added test cases for this.

src/lua/msgpack.c
CuriousGeorgiy: Artifact from previous patch version: updated the comment to reflect the current behaviour.

src/lua/msgpack.c
```diff
@@ -809,6 +815,80 @@ luamp_object_iterator(struct lua_State *L)
 	return 1;
 }
 
+/**
+ * Decodes the MsgPack data stored in @a obj and saves a reference to it on the
+ * Lua stack.
+ */
+static void
+luamp_object_get_decode(struct lua_State *L, struct luamp_object *obj)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Okay, inlined it.

test/app-luatest/msgpack_test.lua
```diff
@@ -445,3 +446,91 @@ g.test_object_misc = function()
     t.assert_not(msgpack.is_object(it))
     t.assert_not(msgpack.is_object({mp}))
 end
+
+local test_array = {true, 1, '2', -2, 3.0, {1, 2, 3}, {a = 1, b = 2, c = 3}, 'decode', 'iterator', 'get'}
+local test_map = {a = nil, [1] = true, ['1'] = 1, b = '2', [-2] = 3.0, ['-2'] = 4, [2.5] = 5, [true] = 6,
+                  c = {a = 7, b = 8, c = 9}, decode = 10, iterator = 11, get = 12}
```
locker, line 452: Nit: Please keep lines within 80 characters in Lua files. I added a checkpatch check for this.

CuriousGeorgiy: This is questionable: according to luacheck [documentation](https://luacheck.readthedocs.io/en/stable/warnings.html?highlight=max%20line%20length#line-length-limits), the default line length limit is 120.

Anyways, you should rather patch [.luacheckrc](https://github.com/tarantool/tarantool/blob/master/.luacheckrc) than checkpatch.

locker: > This is questionable: according to luacheck [documentation](https://luacheck.readthedocs.io/en/stable/warnings.html?highlight=max%20line%20length#line-length-limits), the default line length limit is 120.

We follow the 80 character limit in most our Lua files. 120 characters don't fit in my screen:

image

> Anyways, you should rather patch [.luacheckrc](https://github.com/tarantool/tarantool/blob/master/.luacheckrc) than checkpatch.

I don't want to update luacheckrc now, because it's applied to all Lua files and we do have a few black sheep in the flock (tests mostly). I don't want to patch them now. I just want to prevent new files with long lines from appearing. Also, there may be exceptions when we do want to ignore this rule (e.g. long tables). Checkpatch failure may be ignored while luacheck must always pass, because it's run on the master.

CuriousGeorgiy: Fixed.

Regarding, the left `LONG_LINE_STRING` errors, please see my [comment](https://github.com/tarantool/tarantool/pull/7994#issuecomment-1334231161).

locker: You don't even need to break the line here:

```Lua
    t.assert_error_msg_content_equals(
        "bad argument #1 to '?' (msgpack.object expected, got number)",
```

Anyway, when a line is too long, you can break it by punctuation, e.g.:

```Lua
    t.assert_error_msg_content_equals("bad argument #1 to '?' " ..
                                      "(msgpack.object expected, got number)",
```

A dev will hardly try to grep the whole message. `grep "bad argument"` is more likely.

src/lua/msgpack.c
CuriousGeorgiy: Discussed with @unera: when a `msgpack.object` that does not contain an indexable data type is indexed, a `not an array or map` error is thrown.

test/app-luatest/msgpack_test.lua
CuriousGeorgiy: I agree with your points, fixed: split the lines as you suggest.

locker (approved):

PR #7958, opened 2022-11-21
box: inherit index collation from the space format

### LINKED ISSUE #5104: sql: collation is not used during index search
```
t = box.schema.create_space("test")
t:format({{name = "col1", type = "string", collation = "unicode_ci"}})
t:create_index('primary')
t:insert{"aaa"}
t:insert{"bbb"}
t:insert{"ccc"}
tarantool> box.execute("select * from \"test\"  where \"col1\" = 'AAA'")
---
- metadata:
  - name: col1
    type: string
  rows: []
```
Collation is defined in space's format, but it is not used since index iterator ignores it.

Gumix:
If `create_index('...', {parts = {...}})` is called without the `collation` option, then use collation from the space format.

Closes #5104

### DIFF
--- a/changelogs/unreleased/gh-5104-inherit-index-collation-from-format.md
+++ b/changelogs/unreleased/gh-5104-inherit-index-collation-from-format.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed a bug in `space_object:create_index()` when `collation` option is not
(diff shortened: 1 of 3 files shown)


locker (approved):

locker:
Cherry-picked to 2.10.

Issue #3853, opened 2018-12-03
log.info spoils input


rosik:
Tarantool version: 1.10.2-26-gb2ddd18a1

OS version: ubuntu 18.04

Bug description:
```
tarantool> log = require('log')
---
...

tarantool> a = {file='c://autorun.bat'}
---
...

tarantool> a
---
- file: c://autorun.bat
...

tarantool> log.info(a)
2018-11-30 17:18:37.047 [31946] main/101/srv.lua I> []
---
...

tarantool> a
---
- []
...
```

I do not expect `a` is modified.

rosik:
I know this feature is connected with json output.

Current behavior (string)
```
tarantool> log.info("INFO Message")
{"time": "2018-12-03T12:35:41.594+0300", "level": "INFO", "message": "INFO Message", "pid": 9346 , "cord_name": "main", "fiber_id": 101, "fiber_name": "interactive", "file": "[C]", "line": -1}
```

Current behavior (table):
```
tarantool> log.info({foo='bar'})
{"time": "2018-12-03T12:36:02.819+0300", "level": "INFO", "foo":"bar", "pid": 9346 , "cord_name": "main", "fiber_id": 101, "fiber_name": "interactive", "file": "[C]", "line": -1}
```

Proposed behavior:
```
tarantool> log.info({foo='bar'})
{"time": "2018-12-03T12:36:02.819+0300", "level": "INFO", "message":{"foo":"bar"}, "pid": 9346 , "cord_name": "main", "fiber_id": 101, "fiber_name": "interactive", "file": "[C]", "line": -1}
```

knazarov:
It works as designed. We need some place to report file information, and "file" in your json will clash with whatever is inserted by log module itself.

rosik:
I dislike inconsistency between formats:

With json format (they are the same):
```
tarantool> log.info({message = 'foobar'})
{"time": "2018-12-04T12:26:44.096+0300", "level": "INFO", "message":"foobar", "pid": 27657 , "cord_name": "main", "fiber_id": 101, "fiber_name": "interactive", "file": "[C]", "line": -1}
---
...

tarantool> log.info('foobar')
{"time": "2018-12-04T12:26:53.081+0300", "level": "INFO", "message": "foobar", "pid": 27657 , "cord_name": "main", "fiber_id": 101, "fiber_name": "interactive", "file": "[C]", "line": -1}
---
...
```

With plain format (they differ)
```
[...]

PR #7969, opened 2022-11-23
log: fix assertion on JSON message without string keys

### LINKED ISSUE #7955: Assertion in `say_format_json` failed on `log.info{'die'}`
Tarantool 1.10, 2.10 and 2.11 are affected:
```
Tarantool 1.10.14-28-gd6002844c
Target: Linux-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: /usr/bin/cc /usr/bin/c++
C_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror
CXX_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
```
```
Tarantool 2.10.4-2-g9e0da2a40
Target: Linux-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: GNU-8.3.0
C_FLAGS: -fexceptions -funwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/git/tarantool=. -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror
CXX_FLAGS: -fexceptions -funwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/git/tarantool=. -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
```

Steps to reproduce (on 1.10 `log.cfg{format='json'}` is not available, use `box.cfg{log_format='json'}` instead):
```lua
---
...

tarantool> log.cfg{format='json'}
---
...

tarantool> log.info{'die'}

tarantool: ./src/lib/core/say.c:890: say_format_json: Assertion `str_len > 2 && str[0] == '{' && str[str_len - 1] == '}'' failed.
zsh: abort (core dumped)  /git/tarantool/build/src/tarantool
```

`say_format_json` assumes that `str` is properly JSON-formatted and enclosed in curly braces:
https://github.com/tarantool/tarantool/blob/cba6f917004b571f20dab511561135bbf1889456/src/lib/core/say.c#L883-L892

It is formatted here:
https://github.com/tarantool/tarantool/blob/cba6f917004b571f20dab511561135bbf1889456/src/lua/log.lua#L275-L284

This works well for `log.info({abc=123})` which results in `{"abc":123}`,
however `log.info({'abc'})` results in `["abc"]` without curly braces, causing the crash.

Gumix:
If a Lua table, passed to `say()`, contains only integer indexes, `json.encode()` will encode it as an array rather than a map, thereafter `say_format_json()` will produce improperly formatted JSON string, or get assertion failure. Fix it by encoding all kinds of tables as maps.

Closes #7955


src/lua/log.lua
```diff
@@ -149,6 +149,8 @@ local function say(level, fmt, ...)
         for _, field in ipairs(special_fields) do
             fmt[field] = nil
         end
+        -- always encode tables as maps
+        setmetatable(fmt, json.map_mt)
```
Totktonada, line 153: We set the metatable to the original object passed by a user. We shouldn't modify a user provided object.

We already add fields into the table, though, so it'll not get much worse. OTOH, this is explicitly defined as a problem, see #3853.

I would just add a `table.copy()`. It is simple and solves the problem. (And performance considerations are barely applicable if the behavior is incorrect.)

Gumix: Done

Totktonada: AFAIU, now the assertion supposed to be solved by the following code:

```lua
-- set 'message' field if it is absent
if msg.message == nil then
    msg.message = msg[1]
    msg[1] = nil
end
```

However I guess it'll fail if `msg[1]` is `nil`:

```lua
tarantool> json.encode({nil, 2, 3})
---
- '[null,2,3]'
...
```

I would set the metatable unconditionally.

Totktonada: NB: Verify the old/the new behavior on a table with __serialize or __tostring in its metatable. Whether table.copy() assigns the metatable to the copy? If so, shouldn't we assign our own one only if there is no original one? What if __serialize returns a string like '[1, 2, 3]' or even invalid json?

Gumix: I've addressed this in a new patch, could you please take a look? `__tostring` is not used during logging.

Totktonada: Sorry, I missed your answer. Don't hesitate to ping me in a chat.

There are some doubts about current implementation, but I think that we should proceed now and maybe follow up something later.

For the record:

1. An object with `box.NULL` as a **key** (say `{[box.NULL] = 'x'}`) may pass the `next(msg) == nil` check. A correct check would use `type()`. However it doesn't matter much due to the next point.
2. There are objects that can't be encoded using just `json.encode(x)` (`userdata` for example). Those problems are usually solved using something like `local j = json.new() j.cfg({encode_use_tostring = true}) j.encode(x)`. However even this option doesn't help against `{[box.NULL] = 'x'}`.
3. The `__serialize` metamethod may return a table that is accessible for a user. We set a metatable to it and it may be considered as changing of a user provided object.
4. I'm not sure that removing of `msg[1]`, when it is placed to `msg.message`, is the good idea. What if a user pass an array intentionally? Maybe we should copy it into `msg.message`, not move. Maybe we should copy when the object is considered as an array, but move otherwise. Don't know.
5. I don't know, should we prefer `__tostring` or `__serialize`, when both are there. Now it is `__serialize` and it seems meaningful for a JSON logger, but I don't see a general rule.

I hope sooner or later we'll look into serialization and logging in a more general way. Given patch is tiny and solves several problems: I'm okay with it.

I would like someone else to look into the code and our discussion.

Totktonada (approved):

locker:
Cherry-picked to 1.10. and 2.10.

PR #7971, opened 2022-11-24
lua: load scripts or modules pointed by env var

### LINKED ISSUE #7714: Run a compatibility script if it was specified by environment variable
In order to provide compatibility of current tarantool version with future version we already need to run a specific compatibility script if a certain env variable is set.

Totktonada:
A script file or a module may be specified in the `TT_PRELOAD` environment variable to load before the main script. For example:

```shell
$ TT_PRELOAD=/path/to/foo.lua tarantool main.lua
```

The `foo.lua` script will be executed before `main.lua`.

```shell
$ TT_PRELOAD=foo.bar.baz tarantool main.lua
```

The `foo.bar.baz` module will be loaded before executing `main.lua`.

Several files or modules may be passed as a semicolon separated list:

```shell
$ TT_PRELOAD="/path/to/foo.lua;foo.bar.baz" tarantool main.lua
```

Don't forget quotes, a shell interprets a semicolon on its own.

If something goes wrong at script/module execution, tarantool reports the problem and exits.

A script receives a path to the file in `...`, a module gets the module name. `arg` is the same as for the main script.

Fixes #7714


Totktonada:
The same on Lua/C: 1cce318bd4ec5e76b37af8b39da0e00ad56db892. It looks more complicated, so I prefer the pure Lua implementation.

Totktonada:
* Added `TT_PRELOAD=-` negative case (reading stdin is deliberately unsupported).
* Added 'no script file' and 'no module' negative scenarios.

The negative scenarios are added with just ~50 SLoC of extra testing code. I would like to avoid bloating of the test, so I don't check stdout/stderr in case of an error -- just the exit code. I looks enough for me.

src/lua/init.lua
```diff
@@ -203,7 +203,31 @@ rawset(package, "search", search)
...
 
+-- Execute scripts or load modules pointed by TT_PRELOAD
+-- environment variable.
+local function run_preload()
+    local tt_preload = os.getenv('TT_PRELOAD') or ''
+    assert(type(tt_preload) == 'string')
+    if #tt_preload == 0 then
+        return
+    end
+    for _, script in ipairs(tt_preload:split(';')) do
```
Buristan, line 214: Side note: Can (and should) we support the format with `;` at the end:

```sh
$ TT_PRELOAD="src/test.lua;src.test2;" src/tarantool -e 'print(1)'
0       src/tarantool
-1      /home/burii/reviews/tarantool/Totktonada/src/tarantool
src/test.lua
0       src/tarantool
-1      /home/burii/reviews/tarantool/Totktonada/src/tarantool
src.test2
LuajitError: builtin/tarantool.lua:222: module '' not found:
        no field package.preload['']
        no file '/home/burii/reviews/tarantool/Totktonada/.lua'
        no file '/home/burii/reviews/tarantool/Totktonada//init.lua'
        no file '/home/burii/reviews/tarantool/Totktonada/.so'
        no file '/home/burii/reviews/tarantool/Totktonada/.rocks/share/tarantool/.lua'
        no file '/home/burii/reviews/tarantool/Totktonada/.rocks/share/tarantool//init.lua'
        no file '/home/burii/reviews/tarantool/.rocks/share/tarantool/.lua'
        no file '/home/burii/re
```

Maybe we can just skip empty `''` entries?

Buristan (approved):
Hi, Alexander!
Thanks for the patch!
LGTM, just a few nits and questions below.

test/app-luatest/preload_test.lua
```diff
@@ -0,0 +1,321 @@
...
+        end
+    end
+end
+
+-- Generate a script that follows SCRIPT_TEMPLATE and
+-- write it at given script file path in given directory.
+local function write_script(dir, script)
+    local script_abspath = fio.pathjoin(dir, script)
+    local flags = {'O_CREAT', 'O_WRONLY', 'O_TRUNC'}
+    local mode = tonumber('644', 8)
```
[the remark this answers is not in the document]
Totktonada: Hm, I use this hack several years and your message makes me curious: whether there is an issue to fix default `fio.open()` behavior on creating files. As I see, no. So I created it: #7981.

test/app-luatest/preload_test.lua
```diff
@@ -0,0 +1,321 @@
...
+--
+-- We should also inspect `arg` content to freeze certain
+-- behavior.
+local SCRIPT_TEMPLATE = [[
+print(require('json').encode({
+    ['script'] = '<script>',
+    ['...'] = {...},
+    ['arg[-1]'] = arg[-1],
+    ['arg[0]'] = arg[0],
+    ['arg[]'] = setmetatable(arg, {__serialize = 'seq'}),
```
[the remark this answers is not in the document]
Totktonada: `arg` contains `arg[-1]` and `arg[0]`. `arg[]` is supposed to say 'arg as an array, from 1 to the end'.

I can change it to something else if `arg[]` makes things more confusing rather than self-descriptive. After my comment, do you think I should leave it or change (and to what)?

src/lua/init.lua
Totktonada: We definitely can. Should we? Don't know.

This idea comes to me as well. I didn't find why one may request it.

Now I think a bit more and find that it may ease construction of this variable using concatenation. So maybe it worth it.

I'll raise the same discussion in the internal RFC document and will come here with updates.

src/lua/init.c
```diff
@@ -950,6 +950,19 @@ run_script_f(va_list ap)
...
+	/*
+	 * Execute scripts or modules pointed by TT_PRELOAD
+	 * environment variable.
+	 */
+	lua_getfield(L, LUA_GLOBALSINDEX, "package");
+	lua_getfield(L, -1, "loaded");
+	lua_getfield(L, -1, "tarantool");
+	lua_getfield(L, -1, "_internal");
+	lua_getfield(L, -1, "run_preload");
+	if (luaT_call(L, 0, 0) != 0)
```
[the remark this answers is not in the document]
Totktonada: (I was close to change the code to follow your suggestion, but grepped sources on `LUA_OK` and scratched my head.)

It is very usual for tarantool to check a return value against zero. Moreover, the [reference manual](https://www.lua.org/manual/5.1/manual.html#lua_pcall) says:

> The `lua_pcall` function returns 0 in case of success or one of the following error codes (defined in lua.h):
>
> `LUA_ERRRUN`: a runtime error.
> `LUA_ERRMEM`: memory allocation error. For such errors, Lua does not call the error handler function.
> `LUA_ERRERR`: error while running the error handler function.

Surprisingly, no `LUA_OK`. Across the whole manual.

And we have no checks against `LUA_OK` within the `src` directory.

I would stick with existing approach if you don't insist.

test/app-luatest/preload_test.lua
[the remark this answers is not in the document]
Buristan: We can leave at as is, but I suggest to add a comment about this. I missed a part, that `seq` serialization will omit `0` and `-1` indexes during JSON serialization. With comment it becomes clear.

src/lua/init.c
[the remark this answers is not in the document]
Buristan: OK, never mind.

src/lua/init.lua
Totktonada: Quote from the RFC:

> ### Whether to allow empty entries?
>
> From @Buristan and @Totktonada.
>
> If we’ll go with the direct solution (`:split(';')` without filtering empty entries), the behavior will be the following:
>
> ```bash
> TT_PRELOAD='foo;bar' tarantool  # correct: load 'foo' and 'bar'
> TT_PRELOAD='foo;bar;' tarantool # error
> TT_PRELOAD='foo;;bar' tarantool # error
> ```
>
> We can allow empty entries to ease construction of the environment valuable value using concatenation.
>
> This way it'll work in the following way:
>
> ```bash
> TT_PRELOAD='foo;bar' tarantool  # correct: load 'foo' and 'bar'
> TT_PRELOAD='foo;bar;' tarantool # correct: load 'foo' and 'bar'
> TT_PRELOAD='foo;;bar' tarantool # correct: load 'foo' and 'bar'
> ```
>
> Decision: allow.

@igormunkin's comment:

> Just consider empty string/path/module name as a noop. Hence there will be no issues with multiple/trailing separators. The only weak point of such approach is validation, but I doubt we have to validate empty argument in any other way than ignoring it.

@unera agreed too.

I'll update the patch.

Totktonada: <details>
<summary>The fixup</summary>

```diff
diff --git a/src/lua/init.lua b/src/lua/init.lua
index 723d16f44..77072e7c2 100644
--- a/src/lua/init.lua
+++ b/src/lua/init.lua
@@ -212,7 +212,13 @@ local function run_preload()
         return
     end
     for _, script in ipairs(tt_preload:split(';')) do
-        if script:endswith('.lua') then
+        if #script == 0 then
+            -- Ignore empty entries to allow duplicated semicolons
+            -- and leading/trailing semicolons.
+            --
+            -- It simplifies construction of the environment
+            -- variable value using concatenation.
+        elseif script:endswith('.lua') then
             local fn, err = loadfile(script)
             if fn == nil then
                 error(err)
diff --git a/test/app-luatest/preload_test.lua b/test/app-luatest/preload_test.lua
index c924f5866..cd8ba3d07 100644
--- a/test/app-luatest/preload_test.lua
+++ b/test/app-luatest/preload_test.lua
@@ -184,7 +184,10 @@ local function expected_output(scripts, env, args)
     -- TT_PRELOAD entries.
     for _, entry in ipairs(env['TT_PRELOAD']:split(';')) do
         local script
-        if entry:endswith('.lua') then
+        if #entry == 0 then
+            -- Ignore empty entries.
+            goto continue
+        elseif entry:endswith('.lua') then
             -- The entry is a script file.
             script = entry
         else
@@ -219,6 +222,8 @@ local function expected_output(scripts, env, args)
             ['arg[0]'] = args[1],
             ['arg[]'] = script_args,
         })
+
+        ::continue::
     end
 
     -- The main script.
@@ -296,6 +301,30 @@ for _, case in ipairs({
         env = {TT_PRELOAD = 'foo.lua;bar'},
         args = {'main.lua'},
     },
+    {
+        'test_leading_semicolon',
+        scripts = {'foo.lua', 'bar.lua', 'main.lua'},
+        env = {TT_PRELOAD = ';foo.lua;bar'},
+        args = {'main.lua'},
+    },
+    {
+        'test_trailing_semicolon',
+        scripts = {'foo.lua', 'bar.lua', 'main.lua'},
+        env = {TT_PRELOAD = 'foo.lua;bar;'},
+        args = {'main.lua'},
+    },
+    {
+        'test_duplicated_semicolon',
+        scripts = {'foo.lua', 'bar.lua', 'main.lua'},
+        env = {TT_PRELOAD = 'foo.lua;;bar'},
+        args = {'main.lua'},
+    },
+    {
+        'test_only_semicolons',
+        scripts = {'main.lua'},
+        env = {TT_PRELOAD = ';;;'},
+        args = {'main.lua'},
+    },
     -- Negative cases.
     {
         'test_nonexisting_script',
```
</details>

Updated the commit/PR description this way:

> Redundant (duplicated, leading, trailing) semicolons are ignored.

test/app-luatest/preload_test.lua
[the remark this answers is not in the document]
Totktonada: I have added a note:

```lua
-- Note: In order to serialize `arg` into JSON it is split into
-- `arg[-1]`, `arg[0]` and `arg[]`. The latter means an array of
-- all items from 1 to the end.
```

src/lua/init.lua
Totktonada: Ouch. Fixup of the fixup:

```diff
diff --git a/src/lua/init.lua b/src/lua/init.lua
index 77072e7c2..7f8b7c143 100644
--- a/src/lua/init.lua
+++ b/src/lua/init.lua
@@ -212,6 +212,7 @@ local function run_preload()
         return
     end
     for _, script in ipairs(tt_preload:split(';')) do
+        -- luacheck: ignore 542 empty if branch
         if #script == 0 then
             -- Ignore empty entries to allow duplicated semicolons
             -- and leading/trailing semicolons.
```

src/lua/init.lua
```diff
@@ -206,8 +206,39 @@ rawset(package, "setsearchroot", setsearchroot)
 local compat = require('compat')
 package.loaded.compat = nil
 
+-- Execute scripts or load modules pointed by TT_PRELOAD
+-- environment variable.
+local function run_preload()
+    local tt_preload = os.getenv('TT_PRELOAD') or ''
+    assert(type(tt_preload) == 'string')
```
igormunkin, line 213: Minor: I see no reason for this `assert`. Does it try to prevent the case when `os.getenv` result value type is changed?

src/lua/init.lua
```diff
@@ -206,8 +206,39 @@ rawset(package, "setsearchroot", setsearchroot)
...
+            -- and leading/trailing semicolons.
+            --
+            -- It simplifies construction of the environment
+            -- variable value using concatenation.
+        elseif script:endswith('.lua') then
+            local fn, err = loadfile(script)
+            if fn == nil then
+                error(err)
+            end
+            fn(script)
```
igormunkin, line 230: Minor: If one wonders what is more convenient way to write the code above, see my implementation:
```lua
local fn = assert(loadfile(script))
fn(script)
```
This can be also folded into one-liner, but it's less convenient and less readable IMHO.

Anyway, feel free to ignore.

src/lua/init.lua
```diff
@@ -206,8 +206,39 @@ rawset(package, "setsearchroot", setsearchroot)
...
+    end
+end
+
 return {
     uptime = uptime;
     pid = pid;
-    compat = compat
+    compat = compat;
+    _internal = {
+        run_preload = run_preload,
```
igormunkin, line 242: Minor: Why do we need to store this function? I guess we need to fire this function once and remove it completely when it finishes.

Feel free to ignore.

test/app-luatest/preload_test.lua
```diff
@@ -0,0 +1,354 @@
...
+
+local tempdirs = {}
+
+-- Remove all temporary directories created by the test
+-- unless KEEP_DATA environment variable is set.
+g.after_all = function()
+    local dirs = table.copy(tempdirs)
+    tempdirs = {}
+
+    local keep_data = (os.getenv('KEEP_DATA') or '') ~= ''
```
igormunkin, line 57: Minor: You can just write `local keep_data = not not os.getenv('KEEP_DATA')`.

Feel free to ignore.

igormunkin (approved):
@Totktonada, thanks for the patch! LGTM, with a couple of nits.

src/lua/init.lua
Totktonada: I don't like several things in such approach:

* We protect a user against *accidental* usage of an internal API. Here `internal` namespace is enough and `_internal` is even better, because it also removes the field from autocomplection.
* There is no sense to attempt to forbid reaching an internal API if a user (think that) (s)he really wants it. It is similar to how the class member visibility works in programming languages: it is possible to hack around, but hard to reach by a mistake.
* And, to be honest, if a user (think that) (s)he really wants to reach an internal API, it is better to make this access explicit (via `_internal`, not `debug`) and less fragile (same).
* I don't like the approach implemented for box fields: add them at early initialization, remove a bit later and then add back within `box.cfg()`. That's hell for one, who trying to decipher such code.
* At last, I don't like extra code, which doesn't add any value for a user (in my understanding of his/her needs).

This is my general approach for such things. Don't protect against intentional hacks, but protect from accidental usages of an internal API: i.e. don't mix private and public functions in one namespace (without a meaningful prefix at least).

src/lua/init.lua
Totktonada: Nope, it actually has no sense. Removed.

src/lua/init.lua
Totktonada: I don't mind, will do.

Totktonada:
`box-luatest/pagination_netbox_test.lua` fails in CI on `ubuntu_22_04_aarch64`. It is irrelevant to this patchset. It is tracked in https://github.com/tarantool/tarantool-qa/issues/288. I don't think that it should block this patchset.

PR #7973, opened 2022-11-24
test: fix flaky gh_6539_log_user_space_empty_or_nil_select_test

Gumix:
The test fails with:
```
not ok 2 box-luatest.gh_6539_log_user_space_empty_or_nil_select.test_log_entry_presence_for_memtx_user_space
gh_6539_log_user_space_empty_or_nil_select_test.lua:122: log must contain a critical entry about `box.space.test_memtx:select({0}, {limit = 1001, iterator = "ALL"})` call on a memtx user space
expected: a value evaluating to true, actual: nil
```
Looks like it happens, because `g.server:grep_log()` is executed before `g.server:eval(call)` completes writing to the log. In other tests `grep_log()` is wrapped into `t.helpers.retrying()` in order to avoid such issues, so do the same here.

May _close tarantool/tarantool-qa#264
Let's commit it and wait a bit to see whether it stops flaking.

### DIFF
--- a/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
+++ b/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
@@ -119,9 +119,11 @@ for _, eng in pairs{'memtx', 'vinyl'} do
         for _, call_fmt in pairs(dangerous_call_fmts) do
             local call = call_fmt:format(space)
             g.server:eval(call)
-            t.assert(g.server:grep_log(expected_log_entry, grep_log_bytes),
-                     ('log must contain a critical entry ' ..
-                      'about `%s` call on a %s user space'):format(call, eng))
+            t.helpers.retrying({}, function()
+                t.assert(g.server:grep_log(expected_log_entry, grep_log_bytes),
+                         ('log must contain a critical entry ' ..
+                          'about `%s` call on a %s user space'):format(call, eng))
+            end)
             fio.truncate(log_file)
         end
     end


test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
```diff
@@ -108,20 +112,24 @@ for _, eng in pairs{'memtx', 'vinyl'} do
...
-                        ('log must not contain a critical entry about ' ..
-                         'an empty or nil `select` call on a %s '..
-                         'user space with `fullscan = true` option '..
-                         'set'):format(eng))
+            t.helpers.retrying({}, function()
+                t.assert_not(g.server:grep_log(expected_log_entry, grep_log_bytes),
+                             ('log must not contain a critical entry about ' ..
+                              'an empty or nil `select` call on a %s '..
+                              'user space with `fullscan = true` option '..
+                              'set'):format(eng))
```
CuriousGeorgiy, line 120: Nit: while you are here, would be nice to concatenate all these strings.

> However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them. [^1]

[^1]: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#breaking-long-lines-and-strings

CuriousGeorgiy (approved):

test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
```diff
@@ -108,20 +112,24 @@ for _, eng in pairs{'memtx', 'vinyl'} do
         for _, call_fmt in pairs(explicit_fullscan_call_fmts) do
             local call = call_fmt:format(space)
             g.server:eval(call)
-            t.assert_not(g.server:grep_log(expected_log_entry, grep_log_bytes),
-                        ('log must not contain a critical entry about ' ..
-                         'an empty or nil `select` call on a %s '..
-                         'user space with `fullscan = true` option '..
-                         'set'):format(eng))
+            t.helpers.retrying({}, function()
+                t.assert_not(g.server:grep_log(expected_log_entry, grep_log_bytes),
```
locker, line 116: Wrapping `assert_not` in `retrying` doesn't make any sense: if the pattern is found in the log, then it'll also be found on the next check.

Gumix: Thanks, fixed.

locker (approved):

Issue #6647, opened 2021-11-25
Yield from a GC hook from a fiber serving an iproto request leads to assertion fail


Totktonada:
1.10.11-46-g5c8c025da

```lua
-- box.cfg, grants, requires, create net.box connection.
tarantool> box.cfg({listen = 3301})
tarantool> box.schema.user.grant('guest', 'super')
tarantool> ffi = require('ffi')
tarantool> fiber = require('fiber')
tarantool> net_box = require('net.box')
tarantool> c = net_box.connect(3301)

-- Reproducer itself.
tarantool> f = function() local t = {__gh_hook = ffi.gc(ffi.new('void *'), fiber.yield)} t = nil collectgarbage() end
tarantool> c:call('f')
tarantool: /home/alex/p/tarantool-meta/1.10/src/lua/utils.c:730: cord_on_yield: Assertion `L != NULL' failed.
Aborted
```

AFAIR, we can see an empty lua state in the fibr local storage on master too in some cases. It should be revisited carefully as for master as well as for 1.10.

Info for inspiration:

https://github.com/tarantool/tarantool/commit/9be50ae745a176da128f2244e1777b9b1e9afb6c
https://github.com/tarantool/tarantool/commit/b4e186fa57461820eb8b82306c505833048c40a2
https://github.com/tarantool/tarantool/commit/ec9b544a123f3e0614ab94b83a7ac4b395326f34
https://github.com/tarantool/tuple-merger/issues/18

kyukhin:
Reproduced on master (2.11.0-entrypoint-113-gede831).
Stack trace
```
#1  0x5e5f93 in tarantool_panic_handler+67
#2  0x5f1d4c in cord_on_yield+172
#3  0x618c70 in fiber_yield+192
#4  0x619c20 in fiber_sleep+288
#5  0x5e777c in lbox_fiber_yield.llvm.926667233294543737+12
#6  0x8613b3 in lj_BC_FUNCC+70
#7  0x6609f0 in gc_finalize.llvm.2025679844444349902+240
#8  0x6611ca in gc_onestep.llvm.2025679844444349902+970
#9  0x65a678 in lua_gc+504
#10 0x671534 in lj_cf_collectgarbage+148
#11 0x8613b3 in lj_BC_FUNCC+70
#12 0x5a3620 in execute_lua_call.llvm.15167021168380562020+512
#13 0x8613b3 in lj_BC_FUNCC+70
#14 0x5f0695 in luaT_call+69
[...]

PR #7976, opened 2022-11-26
[backport/1.10] fiber: fix assertion fail in `cord_on_yield`

mkokryashkin:
Background fibers (they serve binary protocol requests) do not store a Lua state in the fiber storage even for Lua call / eval requests.

Before the patch, fiber->storage.lua.stack is filled only for the main fiber and fibers created from Lua using fiber.create() or fiber.new(), but not for background fibers (which serve binary protocol requests). That may lead to a PANIC in `cord_on_yield`.

This patch fills fiber->storage.lua.stack for background fibers that serve a Lua call or eval.

For more, see https://github.com/tarantool/tarantool/commit/0d81bc5d0742ef6660fe001fcf3e9d2a2b0166dd

### DIFF
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -696,7 +696,7 @@ void cord_on_yield(void)
 		 * world and one can obtain the corresponding Lua
 		 * coroutine from the fiber storage.
 		 */
-		struct lua_State *L = fiber()->storage.lua.stack;
+		struct lua_State *L = gco2th(gcref(g->cur_L));
 		assert(L != NULL);
 		lua_pushfstring(L, "fiber %d is switched while running the"
 				" compiled code (it's likely a function with"
@@ -726,7 +726,7 @@ void cord_on_yield(void)
 	 * GC hook is active and the platform is forced to stop.
 	 */
 	if (unlikely(g->hookmask & HOOK_GC)) {
-		struct lua_State *L = fiber()->storage.lua.stack;
+		struct lua_State *L = gco2th(gcref(g->cur_L));
 		assert(L != NULL);
 		lua_pushfstring(L, "fiber %d is switched while running GC"
 				" finalizer (i.e. __gc metamethod)",
--- a/changelogs/unreleased/gh-6647-yield-gc-hook.md
+++ b/changelogs/unreleased/gh-6647-yield-gc-hook.md
@@ -1,0 +1,3 @@
+## bugfix/fiber
+
+* Fixed the assertion fail in `cord_on_yield` (gh-6647).
--- a/test/box-luatest/gh_6647_yield_gc_hook_test.lua
+++ b/test/box-luatest/gh_6647_yield_gc_hook_test.lua
@@ -1,0 +1,83 @@
+#!/usr/bin/env tarantool
+
+local t = require('luatest')
+local g = t.group('gh-6647-yield-gc-hook')
+local fio = require('fio')
+
+local tarantool_bin = arg[-1]
+local PANIC = 256
+local ERRMSG = 'fiber [0-9]+ is switched while running GC finalizer'
+
+local function check_log(t, filename, pattern)
+    local file = fio.open(filename, {'O_RDONLY'})
+    assert(file ~= nil)
+
+    local size = file:stat().size
+    assert(size ~= 0)
+
+    local log_contents = file:pread(size, 0)
+    file:close()
+    assert(log_contents ~= '')
+    t.assert_str_contains(log_contents, pattern, true, 'invalid error message')
+end
+
+local function check_case(t, code, logfile, case_name)
+    local dir = fio.tempdir()
+    local script_path = fio.pathjoin(dir, 'script.lua')
+    local script = fio.open(script_path, {'O_CREAT', 'O_WRONLY', 'O_TRUNC'},
+        tonumber('0777', 8))
+    script:write(code)
+    script:write("\nos.exit(0)")
+    script:close()
+    local cmd = [[/bin/sh -c 'cd "%s" && "%s" ./script.lua 2>&1 /dev/null']]
+    local res = os.execute(string.format(cmd, dir, tarantool_bin))
+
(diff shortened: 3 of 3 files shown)


Totktonada:
The problem is in the `cord_on_yield()` code:

https://github.com/tarantool/tarantool/blob/d4a4a74ee6264f85086de2d537b11dc69c1316fb/src/lua/utils.c#L885-L886

https://github.com/tarantool/tarantool/blob/d4a4a74ee6264f85086de2d537b11dc69c1316fb/src/lua/utils.c#L918-L919

This function assumes that `fiber()->storage.lua.stack` is always non-NULL. This assumption is wrong.

I guess it'll still wrong even after this patch. Consider the following fibers:

* `main` -- it's ok
* `fiber.create()` from Lua -- it's ok
* one that is serving `IPROTO_CALL` for a Lua function -- fixed here
* `fiber_new()` from C -- ??
* one that is serving `IPROTO_CALL` for a C function -- ??
* one that is serving `IPROTO_EXECUTE` for a query that calls a Lua function -- ??
* one that is serving `IPROTO_INSERT` for a space with a trigger -- ??

C code may execute a Lua function, which may yield in an inappropriate place.

It is just what appears in my head and likely not a complete list. The main idea here is that the assumption that `fiber()->storage.lua.stack != NULL` is wrong.

I would also note that a user can use several `struct lua_State` structures within one fiber. AFAIR, `yaml` module do so, but my point is that it is *possible* for a user side code.

So backporting of this patch doesn't fix the problem entirely. However I would backport it anyway, because different behavior of `fiber()->storage.lua.stack` on 1.10 and 2x is error-prone. We can do so as part of #6647 with explained motivation as to eliminate the 1.10 and 2x difference.

Possible solutions:

* Check `fiber()->storage.lua.stack` and fall back to `exit()`/`panic()` without a Lua stack trace if it is NULL.
* Use `cur_L` from LuaJIT (proposed by @igormunkin).

src/box/lua/call.c
```diff
@@ -490,13 +490,40 @@ box_process_lua(struct call_request *request, struct port *base,
 	port_lua_create(base, L);
 	((struct port_lua *) base)->ref = coro_ref;
 
+	/*
```
Totktonada, line 493: Regarding ec9b544a123f3e0614ab94b83a7ac4b395326f34 backport:

The commit message and the pull request description should cleanly state a motivation **of the backport**, when it is different from a motivation of the original commit. Otherwise a reader has no chance to understand why the change was backported. It is very sad, when you have no change to gain knowledge what may go wrong if related code will be changed. I would even say that it may lead to a kind of cargo cult, when some code is kept, because nobody knows, whether it could be changed.

We backport it due to two reasons:

* Make 1.10 and 2x code more similar to make further backports easier and more safe.
* Fill the fiber's Lua state in a frequent case to, again, don't accidentally run into a problem that doesn't exist on 2x.

The original commit was made as an optimization for merger. 1.10 has no merger module at all.

Feel free to choose any form of the description: you can remove the original commit message or keep it separately from backport's motivation. Just do it in a way that allows another developer to understand what is going on here. And don't forget about PR description: it seems, it doesn't reflect recent changes in the patchset.

While we're here:

> That may lead to a PANIC in `cord_on_yield`.

But the panic is a correct behavior. The difference is that before the patchset we had an assertion fail or a memory access problem in the case, but now tarantool prints a backtrace and panics.

test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -0,0 +1,43 @@
...
+    script:write("\nos.exit(0)")
+    script:close()
+    local cmd = [[/bin/sh -c 'cd "%s" && "%s" ./script.lua 2> /dev/null']]
+    local res = os.execute(string.format(cmd, dir, tarantool_bin))
+    fio.rmtree(dir)
+    return res
+end
+
+local code = [[
+box.cfg({listen = 3301})
```
Totktonada, line 26: All comments regarding a test from PR #8048 are applicable here.

test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -0,0 +1,43 @@
...
+fiber = require('fiber')
+net_box = require('net.box')
+c = net_box.connect(3301)
+local function f()
+    local t = {__gh_hook = ffi.gc(ffi.new('void *'), fiber.yield)}
+    t = nil
+    collectgarbage()
+end
+c:call('f')
+os.exit(0)
```
Totktonada, line 38: `run_script()` adds `os.exit()` on its own. It is harmless, but I would do it once.

I would leave `os.exit()` in the script itself, because this way no parts of the script will spread across testing code. However it is just my feelings, feel free to ignore if you feels it differently.

test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -22,7 +22,7 @@ local function run_script(code)
     return res
 end
 
-local code = [[
+local lua_iproto_call = [[
```
Totktonada, line 25: Nit: It is better to assign a name once rather than choose one and change it in a next commit.

src/lua/utils.c
```diff
@@ -696,7 +696,7 @@ void cord_on_yield(void)
 		 * world and one can obtain the corresponding Lua
 		 * coroutine from the fiber storage.
 		 */
-		struct lua_State *L = fiber()->storage.lua.stack;
+		struct lua_State *L = gco2th(gcref(g->cur_L));
```
Totktonada, line 699: This part is in review in PR #8048 (against `master`). I left this comment as reminder to wait for it and to add 'cherry-picked from <...>' later into the commit message.

igormunkin:
This PR is superseded by #8048 to totally resolve the issue in all long term branches. All the code health activity will be completed within a separate PR later (i.e. when it is needed).

Issue #7897, opened 2022-11-03
Sending arbitrary IPROTO packets 


CuriousGeorgiy:
# Interface
Sends an **IPROTO** packet over the session's socket with the given header (required) and body (optional), which are MsgPack.

```C
int 
box_iproto_send(uint64_t sid, char *header, char *header_end, char *body, char *body_end);
```

The header and body can be tables in which case they can contain normalized (i.e., lower case and without the `IPROTO_` prefix) **IPROTO** constants as keys.
```lua
function box.iproto.send(sid, header, body) end
```

# Errors
| Error                 | Condition                      | Format                              |
| --------------------- | ------------------------------ | ----------------------------------- |
| `ER_SESSION_CLOSED`     | closed session                 | see also [`src/box/errcode.h`](https://github.com/tarantool/tarantool/blob/master/src/box/errcode.h)                         |
| `ER_NO_SUCH_SESSION`    | session does not exist         | “Session " PRIu64 " does not exist” |
| `ER_MEMORY_ISSUE`       | OOM                            | см. src/box/errcode.h               |
| `ER_WRONG_SESSION_TYPE` | session type is not **IPROTO** | “PRIu64 ” is not an IPROTO session” |

# Examples
```lua
box.iproto.send(sid, {request_type = box.iproto.OK,
                      sync = 10, 
                      schema_version = box.schema_version()}), 
                {data = 1}))
```

```lua
box.iproto.send(sid, {[box.iproto.REQUEST_TYPE] = box.iproto.OK,
                       [box.iproto.SYNC] = 10, 
                       [box.iproto.SCHEMA_VERSION] = box.schema_version()}), 
                {[box.iproto.DATA] = 1}))
```

```lua
box.iproto.send(sid, {request_type = box.iproto.OK, 
                      sync = 10, 
                      schema_version = box.schema_version()}))
```

```lua
box.iproto.send(sid, {request_type = bit.bor(box.iproto.TYPE_ERROR, box.error.SPACE_EXISTS),
                      sync = 10, 
                      schema_version = box.schema_version()}), 
                {error_24 = "Space '_space' already exists",
                 error = box.error.last():unpack()}))
```

```lua
box.iproto.send(sid, msgpack.encode({[box.iproto.REQUEST_TYPE] = box.iproto.OK, 
                                     [box.iproto.SYNC] = 10, 
                                     [box.iproto.SCHEMA_VERSION] = box.schema_version()}), 
[...]

PR #7984, opened 2022-11-29
msgpack: add translation of first-level `MP_MAP` keys during encoding

CuriousGeorgiy:
Needed for #7897

### DIFF
--- a/src/lua/msgpack.c
+++ b/src/lua/msgpack.c
@@ -42,6 +42,7 @@
 #include <small/region.h>
 #include <small/ibuf.h>
 
+#include "core/assoc.h"
 #include "core/decimal.h" /* decimal_unpack() */
 #include "lua/decimal.h" /* luaT_newdecimal() */
 #include "mp_extension_types.h"
@@ -162,9 +163,29 @@ luamp_set_decode_extension(luamp_decode_extension_f handler)
 	}
 }
 
+/**
+ * Tries to translate MP_MAP key to an unsigned integer from the translation
+ * table.
+ */
+static void
+translate_map_key_field(struct luaL_field *field, struct mh_strnu32_t *translation)
+{
+	mh_int_t key = mh_strnu32_find_str(translation,
+					   field->sval.data,
+					   field->sval.len);
+	if (key != mh_end(translation)) {
+		field->type = MP_UINT;
+		field->ival = (int64_t)mh_strnu32_node(translation, key)->val;
+	}
+}
+
 enum mp_type
-luamp_encode_r(struct lua_State *L, struct luaL_serializer *cfg,
-	       struct mpstream *stream, struct luaL_field *field, int level)
+luamp_encode_with_translation_r(struct lua_State *L,
+				struct luaL_serializer *cfg,
+				struct mpstream *stream,
+				struct luaL_field *field,
+				int level,
+				struct mh_strnu32_t *translation)
 {
 	int top = lua_gettop(L);
 	enum mp_type type;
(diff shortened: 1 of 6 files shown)


test/unit/lua_msgpack.c
```diff
@@ -0,0 +1,117 @@
...
+ * Checks that translation of first-level `MP_MAP` keys is done correctly.
+ */
+static void
+test_translation_in_encoding(lua_State *L)
+{
+	plan(3);
+	header();
+
+	struct mh_strnu32_t *translation = mh_strnu32_new();
+	const char alias[] = "x";
```
locker, line 34: Nit: This looks weird. Why not
```C
const char *alias = "x";
```
and use `strlen(alias)` instead of `lengthof(alias) - 1`?

test/unit/lua_msgpack.c
```diff
@@ -0,0 +1,117 @@
...
+	lua_settable(L, -3);
+	luamp_encode_with_translation(L, luaL_msgpack_default, &stream, 1,
+				      translation);
+	mpstream_flush(&stream);
+	ok(strncmp(ibuf->buf, "\x81\x00\xc3", ibuf_used(ibuf)) == 0,
+	   "only keys with MP_STRING type are translated");
+	ibuf_reset(ibuf);
+	mpstream_reset(&stream);
+
+	cord_ibuf_drop(ibuf);
```
locker, line 87: Nit: `mh_strnu32_delete` is missing.

test/unit/lua_msgpack.c
```diff
@@ -0,0 +1,117 @@
...
+	   "only first-level MP_MAP key is translated");
+	ibuf_reset(ibuf);
+	mpstream_reset(&stream);
+
+	lua_createtable(L, 0, 1);
+	lua_pushnumber(L, 0);
+	lua_pushboolean(L, true);
+	lua_settable(L, -3);
+	luamp_encode_with_translation(L, luaL_msgpack_default, &stream, 1,
+				      translation);
```
locker, line 80: What happens if the Lua table contains both the alias and its value `t = {x = 10, [0] = 20}`?

src/lua/msgpack.c
```diff
@@ -288,8 +315,17 @@ luamp_encode_r(struct lua_State *L, struct luaL_serializer *cfg,
 }
 
 enum mp_type
-luamp_encode(struct lua_State *L, struct luaL_serializer *cfg,
-	     struct mpstream *stream, int index)
+luamp_encode_r(struct lua_State *L, struct luaL_serializer *cfg,
+	       struct mpstream *stream, struct luaL_field *field, int level)
+{
+	return luamp_encode_with_translation_r(L, cfg, stream, field,
```
locker, line 321: Nit: Could be static inline and reside it in the header IMO. The comment wouldn't be necessary then. Same comment for `luamp_encode`.

test/unit/lua_msgpack.c
CuriousGeorgiy: Discussed with @locker and @alyapunov: decided to declare this unspecified behavior.

test/unit/lua_msgpack.c
CuriousGeorgiy: Fixed: added missing destructor call.

test/unit/lua_msgpack.c
CuriousGeorgiy: Discussed with @locker: came to an agreement this approach is weird and not worth it — fixed.

locker (approved):

PR #7993, opened 2022-12-01
libunwind: use latest release v1.6.2 as a base

nshy:
Now we base on on some unreleased state of libunwind. This is by itself not a good practice. Yet the main motivation is that in the current version of libunwind fast path for backtrace detection on x86_64 does not work. I guess it is because of f67ef2867bc5 (" x86_64: Stop aliasing RSP and CFA"). See libunwind/libunwind#440. Fortunately this commit is not present it the latest release.

Using fast or slow path has a great impact on performance of debug build where collecting fiber gc allocation backtrace is turned on by default. It surely depends on frequency and pattern of allocation. Test sql/delete3 depends on backtrace performance most dramatically. On some installations collecting backtraces slowed down the test up to 10 times.

If fast path is available then collecting backtrace does not affect performance in a noticeable way.

I propose not to keep autotools products in the libunwind fork repo as it is done previously. LOG_CONFIGURE is removed because it somehow incompatible with using && in CONFIGURE_COMMAND command and not critical to the build.

Follow-up #5665

NO_DOC=internal
NO_TEST=internal
NO_CHANGELOG=internal

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -59,7 +59,7 @@
 [submodule "third_party/libunwind"]
 	path = third_party/libunwind
 	url = https://github.com/tarantool/libunwind.git
-	branch = libunwind-f67ef28-tarantool
+	branch = libunwind-1.6.2-tarantool
 [submodule "third_party/tz"]
 	path = third_party/tz
 	url = https://github.com/tarantool/tz.git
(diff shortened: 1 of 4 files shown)


locker (approved):

locker:
@nshy opensususe workflows [fail](https://github.com/tarantool/tarantool/actions/runs/3600134681/jobs/6064684161). Please take a look.

CuriousGeorgiy:
@nshy AFAIC, you should also update the branch in [.gitmodules](https://github.com/tarantool/tarantool/blob/master/.gitmodules):
https://github.com/tarantool/tarantool/blob/9adc4aa7cc477a1a1aad68181fb48a332009c065/.gitmodules#L59-L62

locker (approved):

Issue #7495, opened 2022-07-29
Wrong target printed by `--version` on M1


Gumix:
**Bug description**

Tarantool --version reports x86_64 target instead of arm64.

* OS: macOS
* OS Version: any
* Architecture: arm64

```
Tarantool 2.11.0-entrypoint-287-g822396d9b
Target: Darwin-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
C_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
CXX_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
```

**Steps to reproduce**

Build native arm64 binary for macOS.

**Actual behavior**

```
$ file ./tarantool
./tarantool: Mach-O 64-bit executable arm64

$ ./tarantool --version
Tarantool 2.11.0-entrypoint-287-g822396d9b
Target: Darwin-x86_64-Debug
```

**Expected behavior**

```
$ ./tarantool --version
Tarantool 2.11.0-entrypoint-287-g822396d9b
Target: Darwin-arm64-Debug
```

Discovered in #7484.

Gumix:
Bug is here: https://github.com/tarantool/tarantool/blob/master/cmake/arch.cmake#L28

PR #7997, opened 2022-12-02
cmake: fix wrong CPU architecture reported on M1/M2 Macs

Gumix:
This patch removes an old CMake kludge, which sets `CMAKE_SYSTEM_PROCESSOR` to `x86_64` on all 64-bit architectures, even on ARM (on Darwin).

Closes https://github.com/tarantool/tarantool/issues/7495


locker (approved):

locker:
Cherry-picked to 2.10.

PR #8001, opened 2022-12-05
box: export current session's identifier to C API

### LINKED ISSUE #7895: Export current session ID to C API
In order to send IPROTO packets using `box_iproto_send` (gh-7897) we need to have some session identifier source in the C API: for now, export an analog of `box.session.id`.

# Interface

Returns the current session's identifier (see also [`box.session.id`](https://github.com/tarantool/tarantool/blob/dec0e0221e183fa972efa65bb0fb658112f2196f/src/box/lua/session.c#L75-L88)).
```C
uint64_t
box_session_id();
```

CuriousGeorgiy:
Closes #7895

### DIFF
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -599,6 +599,12 @@ box_sequence_reset(uint32_t seq_id);
 API_EXPORT int
 box_session_push(const char *data, const char *data_end);
 
+/**
+ * \return current session's unique monotonic identifier (\sa box.session.id)
+ */
+API_EXPORT uint64_t
+box_session_id(void);
+
 /** \endcond public */
 
 /**
--- a/test/app-tap/module_api.c
+++ b/test/app-tap/module_api.c
@@ -2994,6 +2994,19 @@ test_box_schema_version(struct lua_State *L)
 
 /* }}} Helpers for schema version Lua/C API test cases */
 
+/* {{{ Helpers for current session identifier Lua/C API test cases */
+
+static int
+test_box_session_id(struct lua_State *L)
+{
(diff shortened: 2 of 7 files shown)


test/app-tap/module_api.c
```diff
@@ -3045,6 +3058,7 @@ luaopen_module_api(lua_State *L)
 		{"isdecimal", test_isdecimal},
 		{"isdecimal_ptr", test_isdecimal_ptr},
 		{"box_schema_version_matches", test_box_schema_version},
+		{"box_session_id", test_box_session_id},
```
locker, line 3061: Nit: Please rename "box_session_id" to "box_session_id_matches" here for consistency with "box_schema_version_matches".

src/box/box.h
```diff
@@ -599,6 +599,12 @@ box_sequence_reset(uint32_t seq_id);
 API_EXPORT int
 box_session_push(const char *data, const char *data_end);
 
+/**
+ * \return current session's unique monotonic identifier (\sa box.session.id)
+ */
+API_EXPORT uint64_t
+box_session_id();
```
locker, line 606: Nit: Should be defined/declared as `box_session_id(void)` (otherwise it can take an arbitrary number of arguments in C).

CuriousGeorgiy: Fixed: added `void` type parameter.

test/app-tap/module_api.c
CuriousGeorgiy: Fixed: rename all test related names similar to `box_schema_version`.

locker (approved):

PR #8004, opened 2022-12-06
zlib CE part

drewdzzz:
Since we are going to introduce zlib compression in enterprise version, we need to find zlib package there, so let's search for zlib package only if it has not been found before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -408,8 +408,18 @@ endif()
 #
 # ZLIB
 #
-set(ZLIB_FIND_REQUIRED ON)
-find_package(ZLIB)
+
+#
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -408,8 +408,10 @@ endif()
 #
 # ZLIB
 #
-set(ZLIB_FIND_REQUIRED ON)
-find_package(ZLIB)
+if (NOT ZLIB_FOUND)
```
locker, line 411: Nit: Please add a comment here explaining why this is necessary so that one doesn't delete this check by mistake.

drewdzzz: Done

locker (approved):

PR #8007, opened 2022-12-06
lua: exclude _ prefix loop vars from unused check

### LINKED ISSUE #8006: Disable warning for unused loop vars with prefix '_'
In #7261 we provided a way to disable warning for unused arguments in functions, to make the code more verbose. I think, that it is a good idea to do the same with loop vars, consider the example:
```lua
for _name, option in pairs(options) do
    ...
end
```

Lord-KA:
Update luacheckrc to suppress all warnings from lua loop vars with `_` prefix. This allows to make loops more verbose.

Closes #8006

NO_CHANGELOG=luacheckrc fix
NO_DOC=luacheckrc fix
NO_TEST=luacheckrc fix

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -23,6 +23,8 @@ ignore = {
     "432",
     -- Unused variable with `_` prefix.
     "212/_.*",
+    -- Unused loop variable with `_` prefix.
+    "213/_.*",
 }
 
 include_files = {


Totktonada (approved):

Totktonada:
Pushed to 2.10 as 2.10.4-19-g975a15e6f.

Issue #8011, opened 2022-12-06
`gh_6539_log_user_space_empty_or_nil_select_test` sometimes fails with JIT on


Gumix:
The test is flaky. Sometimes it fails, because the `Potentially long select [...]` message is not logged.
Usually this happens on the `box.space.test_memtx:select({0}, {limit = 1001, iterator = "ALL"})` testcase.
When it fails, this condition is false, because `itype` is **0 (ITER_EQ)**, although actually it is **2 (ITER_ALL)**.

https://github.com/tarantool/tarantool/blob/2ed06651c317f2d46631b0d55e4b88f2c8a6c0a0/src/box/lua/schema.lua#L2509-L2520

I tried to insert prints into `check_iterator_type` to understand why it becomes 0, but in that case it is always 2 and the test passes. Also note that it fails only on AArch64. Looks like a LuaJIT issue, this patch helped:

```diff
diff --git a/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua b/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
index 6a3975d4d..93c1f8276 100644
--- a/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
+++ b/test/box-luatest/gh_6539_log_user_space_empty_or_nil_select_test.lua
@@ -8,6 +8,7 @@ g.before_all(function()
     g.server = server:new{alias = 'dflt'}
     g.server:start()
     g.server:exec(function()
+        if jit.arch == 'arm64' then jit.off() end
         require("log").internal.ratelimit.disable()
     end)
 end)
```

PR #8012, opened 2022-12-06
test: fix flaky `gh_6539_log_user_space_empty_or_nil_select_test`

Gumix:
Turn LuaJIT off for this test on AArch64 in the scope of the issue [1].

Also increase `grep_log_bytes`. Currently the message is 235 byte long, which is pretty close to 256. If one more item is added to the traceback, the test will fail.

```
C> Potentially long select from space 'test_memtx' (512)
 stack traceback:
        builtin/box/schema.lua:2528: in function 'check_select_safety'
        builtin/box/schema.lua:2545: in function 'select'
        eval:1: in main chunk
        [C]: at 0x556e8ffd5c
```

[1] Related to #8011 
Closes tarantool/tarantool-qa#264


locker:
Cherry-picked to 2.10.

Issue #7020, opened 2022-04-12
Limit the number of make jobs on CI


locker:
Unlimited number of make jobs may result in OOM. We should use `make -j$(nproc)` instead of `make -j` on Linux. Note, on OS X and FreeBSD, the `nproc` command might be unavailable (this needs to be checked). Something else we could use instead:

- [FreeBSD](https://unix.stackexchange.com/a/477914)
  ```
  sysctl -n hw.ncpu
  ```
- [OS X](https://github.com/memkind/memkind/issues/33#issuecomment-316615308)
  ```
  sysctl -n hw.physicalcpu
  ```

We need to fix all occurrences of `make -j` in [`.travis.mk`](https://github.com/tarantool/tarantool/blob/master/.travis.mk).

locker:
Supposedly, this should fix the CI failure in #7018.

PR #8019, opened 2022-12-07
libunwind: improve incremental build/rebuild

nshy:
Currently we use ExternalProject_Add to build libunwind subproject. The usecases it supports are not quite aligned with our. Configuration and build steps are rerun in case download step sees updated sources. But we use git submodules to fetch third-party projects and thus subproject is not rebuild when it's files are changed on disk. Either because of we doing some sort of experiments locally or because the new sources are brought by git pull.

Actually libunwind has a CMakeList.txt so that we could try to build it just as project subdirectory. But it requires quite a fresh CMake version, currently supports only build on Visual Studio and probably is not up to date given it's package version variables. Keeping CMakeLists.txt in sync with main autotools build is additional maintenace burden. So I'd like to use main autotools build.

So ext_project_autotools function is added to provide nice build integration with third-party projects which are build using autotools. Actually incremental rebuild works only if CMake is at least 3.12 but I think this should be true for developer installations.

Hopefully we can reuse the function in case we need to bundle more subprojects with autotools builds.

Follow-up #5665

NO_DOC=build improvement
NO_TEST=build improvement
NO_CHANGELOG=build improvement

### DIFF
--- a/cmake/ext_project_autotools.cmake
+++ b/cmake/ext_project_autotools.cmake
@@ -1,0 +1,119 @@
+include(utils)
+include(ProcessorCount)
+
+# Build a third-party project that uses autotools to generate build system.
+# Basically it generates configure script, run configure and then make and
+# rerun them according to their dependants in case of incremental build.
+#
+# DIR - relative path to project directory in main project tree.
+# CONFIGURE - options for configure script.
+# BYPRODUCTS - products caller targets depend upon.
+#
+# Note that incremental rebuild in case autotools files or config files are
+# changed is available only if CMake version is at least 3.12. Otherwise
+# rebuild will not be triggered. Rebuild is only required for developers
+# which presumably has newer CMake versions.
+function(ext_project_autotools name)
+    set(sv_keywords
+        DIR
+    )
+    set(mv_keywords
+        CONFIGURE
+        BYPRODUCTS
+    )
+    set(options
+        ""
+    )
+    cmake_parse_arguments(
+        ARGS "${options}" "${sv_keywords}" "${mv_keywords}" ${ARGN}
+    )
+    if (NOT DEFINED ARGS_DIR)
+        message(FATAL_ERROR "DIR argument is mandatory")
+    endif()
+
+    # CONFIGURE_DEPENDS available since 3.12
+    # VERSION_GREATER_EQUAL is only since 3.7
+    if (${CMAKE_VERSION} VERSION_GREATER "3.12" OR
+        ${CMAKE_VERSION} VERSION_EQUAL "3.12")
+        set(autotool_files_standard
+            configure.ac
+            acinclude.m4
+        )
+        list_add_prefix(
+            autotool_files_standard "${ARGS_DIR}/" autotool_files_standard
+        )
+
+        file(
+            GLOB_RECURSE autotool_files_am
+            CONFIGURE_DEPENDS
+            RELATIVE "${CMAKE_SOURCE_DIR}"
+            "${ARGS_DIR}/*.am"
+        )
+
+        set(autotool_files ${autotool_files_am} ${autotool_files_standard})
+
+        file(
+            GLOB_RECURSE config_files_in
+            CONFIGURE_DEPENDS
+            RELATIVE "${CMAKE_SOURCE_DIR}"
+            "${ARGS_DIR}/*.in"
+        )
+    endif()
+
+    list_add_prefix(ARGS_BYPRODUCTS "${ARGS_DIR}/" byproducts)
+
+    file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/${ARGS_DIR})
+
+    add_custom_command(
+        OUTPUT ${CMAKE_SOURCE_DIR}/${ARGS_DIR}/configure
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/${ARGS_DIR}
+        COMMAND
+            autoreconf -i
+        COMMAND
+            # autoreconf do not touch configure on all paths. For example if
+            # only *.am are changed. We don't aim to follow precise dependency
+            # path of autotools thus let's just update configure in this case.
+            touch configure
+        DEPENDS ${autotool_files}
+    )
+
+    add_custom_command(
+        OUTPUT ${ARGS_DIR}/Makefile
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND ${CMAKE_SOURCE_DIR}/${ARGS_DIR}/configure ${ARGS_CONFIGURE}
+        DEPENDS ${ARGS_DIR}/configure
+    )
+
+    # Some *.in files are products of *.am files and some are not. For the
+    # latter we don't need to run autoreconf && configure as it will be
+    # overkill. We only need to run config.status.
+    add_custom_command(
+        OUTPUT
+            ${ARGS_DIR}/config.log
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND ./config.status
+        DEPENDS
+            ${config_files_in}
+            # This dependency reflects dependency on config.status.
+            # config.status and Makefile are both updated on configure run.
+            ${ARGS_DIR}/Makefile
+    )
+
+    ProcessorCount(nproc)
+    if (nproc)
+        set(make make -j${nproc})
+    else()
+        set(make make)
+    endif()
+
+    add_custom_target(${name}
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND
+            ${make}
+        DEPENDS
+            ${ARGS_DIR}/Makefile
+            ${ARGS_DIR}/config.log
+        BYPRODUCTS
+            ${byproducts}
+    )
+endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,7 +533,6 @@ if(ENABLE_BACKTRACE)
         endif()
 
         include(BuildLibUnwind)
-        libunwind_build()
         add_dependencies(build_bundled_libs
                          bundled-libunwind
                          bundled-libunwind-platform)
--- a/cmake/BuildLibUnwind.cmake
+++ b/cmake/BuildLibUnwind.cmake
@@ -9,100 +9,73 @@ Result Variables
   Include directory needed to use libunwind.
 ``LIBUNWIND_LIBRARIES``
   Libraries needed to link to libunwind.
-
-Cache Variables
-^^^^^^^^^^^^^^^
-``LIBUNWIND_INCLUDE_DIR``
-  The directory containing ``libunwind.h``.
-``LIBUNWIND_LIBRARIES``
-  The paths to the libunwind libraries.
 #]========================================================================]
 
-macro(libunwind_build)
-    set(LIBUNWIND_SOURCE_DIR ${PROJECT_SOURCE_DIR}/third_party/libunwind)
-    set(LIBUNWIND_BUILD_DIR ${PROJECT_BINARY_DIR}/build/libunwind)
-    set(LIBUNWIND_BINARY_DIR ${LIBUNWIND_BUILD_DIR}/work)
-    set(LIBUNWIND_INSTALL_DIR ${LIBUNWIND_BUILD_DIR}/dest)
-    set(LIBUNWIND_CFLAGS "${DEPENDENCY_CFLAGS} -g -O2")
-    set(LIBUNWIND_CXXFLAGS "-g -O2")
+include(ext_project_autotools)
 
-    include(ExternalProject)
-    ExternalProject_Add(bundled-libunwind-project
-                        TMP_DIR ${LIBUNWIND_BUILD_DIR}/tmp
-                        STAMP_DIR ${LIBUNWIND_BUILD_DIR}/stamp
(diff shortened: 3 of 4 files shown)


cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,118 @@
...
+    set(mv_keywords
+        CONFIGURE
+        BYPRODUCTS
+    )
+    set(options
+        ""
+    )
+    cmake_parse_arguments(
+        ARGS "${options}" "${sv_keywords}" "${mv_keywords}" ${ARGN}
+    )
```
Totktonada, line 28: 1. At the first glance it was not obvious, what `sv_keywords`, `mv_keywords` and `options` mean. We can name it, say, `opts_0`, `opts_1`, `opts_multi` and I guess it will be more obvious.

    ```cmake
    # Lists of options without arguments, with one argument and
    # with multiple arguments.
    set(opts_0 "")
    set(opts_1 DIR)
    set(opts_multi CONFIGURE BYPRODUCTS)
    ```
2. Isn't `ARGS_*` variables have are too common names (can clash with others)? Shouldn't we use something from the function name in the name of the variables?

    ```cmake
    # Lists of options without arguments, with one argument and
    # with multiple arguments.
    set(epa_opts_0 "")
    set(epa_opts_1 DIR)
    set(epa_opts_multi CONFIGURE BYPRODUCTS)

    cmake_parse_arguments(
        EPA_ARGS "${epa_opts_0}" "${epa_opts_1}" "${epa_opts_multi}" ${ARGN}
    )
    ```

Totktonada: https://cmake.org/cmake/help/latest/command/set.html

> Sets the given `<variable>` in the current function or directory scope.

Please, disregard the second point.

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,118 @@
...
+            ${config_files_in}
+            # Actually we depend on config.status but it updates
+            # it's own timestamp on every run.
+            ${ARGS_DIR}/Makefile
+    )
+
+    add_custom_target(${name}
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND
+            make -j
```
Totktonada, line 111: It is undesirable to run unlimited number of make jobs: the build may fail due to OOM on a machine with low amount of memory. See #7020 for example. See also https://github.com/tarantool/tarantool/pull/7354#discussion_r914740882.

Totktonada: Maybe [CMAKE_BUILD_PARALLEL_LEVEL](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html) should do the trick, but it is available since CMake 3.12.

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,118 @@
...
+        list_add_prefix(
+            autotool_files_standard "${ARGS_DIR}/" autotool_files_standard
+        )
+
+        file(
+            GLOB_RECURSE autotool_files_am
+            CONFIGURE_DEPENDS
+            RELATIVE "${CMAKE_SOURCE_DIR}"
+            "${ARGS_DIR}/*.am"
+        )
```
Totktonada, line 50: Whether the idea is to find all `*.am` recursively in the subproject or only ones that reside in the subproject root directory? If the former, I guess that `RELATIVE $X $Y/*.am` should be `RELATIVE $X/$Y *.am`. If the latter, I don't understand why `GLOB_RECURSE` is used, not `GLOB`.

(The same question about `*.in`.)

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,118 @@
...
+            GLOB_RECURSE config_files_in
+            CONFIGURE_DEPENDS
+            RELATIVE "${CMAKE_SOURCE_DIR}"
+            "${ARGS_DIR}/*.in"
+        )
+
+        # Makefile is used as dependency between configure and make thus we
+        # need to exclude it from dependency for config.status step as
+        # it cannot be output of two custom commands.
+        list(REMOVE_ITEM config_files_in ${ARGS_DIR}/Makefile.in)
```
Totktonada, line 64: The command is about `Makefile.in`, but the comment is about `Makefile`. TBH, I don't get the idea.

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,118 @@
...
+#
+# DIR - relative path to project directory in main project tree.
+# CONFIGURE - options for configure script.
+# BYPRODUCTS - products caller targets depend upon.
+#
+# Note that incremental rebuild in case autotools files or config files are
+# changed is available only if CMake version is at least 3.12. Otherwise
+# rebuild will not be triggered. Rebuild is only required for developers
+# which presumably has newer CMake versions.
+function(ext_project_autotools name)
```
Totktonada, line 15: (A chart to ease discussion.)

```mermaid
flowchart TD
    cmd_source_update([git submodule update --recursive --init])
    cmd_autoreconf([autoreconf -i])
    cmd_configure([./configure <...>])
    cmd_config_status([./config.status])
    cmd_make([make <...>])

    files_am[*.am]
    files_in[*.in]
    files_configure[configure]
    files_makefile[Makefile]
    files_config_stamp[config.stamp.tarantool]
    files_byproducts[$byproducts]

    cmd_source_update --> files_am
    cmd_source_update --> files_in
    files_am --> cmd_autoreconf
    files_in --> cmd_config_status
    cmd_autoreconf --> files_configure
    cmd_autoreconf -.-> files_in
    files_configure --> cmd_configure
    cmd_configure --> files_makefile
    files_makefile --> cmd_config_status
    cmd_config_status --> files_config_stamp
    files_makefile --> cmd_make
    files_config_stamp --> cmd_make
    cmd_make -.-> files_byproducts
```

cmake/ext_project_autotools.cmake
Totktonada: > ```
> /dir/*.py  - match all python files in /dir and subdirectories
> ```

https://cmake.org/cmake/help/latest/command/file.html#glob-recurse

So I was not right: it actually should match `*.am` in nested directories.

cmake/ext_project_autotools.cmake
nshy: `options`, `sv_keywords` and `mv_keywords` follow naming of `cmake_parse_arguments` arguments:

```
cmake_parse_arguments(<prefix> <options> <one_value_keywords>
                      <multi_value_keywords> <args>...)
```

`sv` and `mv` stands for single value and multiple values respectively. Also I use `set` with multiple lines formatting to be able in future extending options and keywords.

cmake/ext_project_autotools.cmake
nshy: Looks like better option for getting cpu number is [ProcessorCount](https://cmake.org/cmake/help/latest/module/ProcessorCount.html) module:

```
--- a/cmake/ext_project_autotools.cmake
+++ b/cmake/ext_project_autotools.cmake
@@ -1,4 +1,5 @@
 include(utils)
+include(ProcessorCount)

 # Build a third-party project that uses autotools to generate build system.
 # Basically it generates configure script, run configure and then make and
@@ -105,10 +106,17 @@ function(ext_project_autotools name)
             ${ARGS_DIR}/Makefile
     )

+    ProcessorCount(nproc)
+    if (nproc)
+        set(make make -j${nproc})
+    else()
+        set(make make)
+    endif()
+
     add_custom_target(${name}
         WORKING_DIRECTORY ${ARGS_DIR}
         COMMAND
-            make -j
+            ${make}
         DEPENDS
             ${ARGS_DIR}/Makefile
             ${ARGS_DIR}/config.stamp.tarantool
```

cmake/ext_project_autotools.cmake
nshy: This is a remnant from draft version where output of `config.status` custom command was list of `${config_files_in}` counterparts. So that root `Makefile` was output of both `config.status` and `configure` custom commands without this hunk which is invalid configuration.

Removed this hunk.

cmake/ext_project_autotools.cmake
nshy: Looks very similar to autotools graph at page [https://en.wikipedia.org/wiki/GNU_Autotools](https://en.wikipedia.org/wiki/GNU_Autotools) :)

cmake/ext_project_autotools.cmake
Totktonada: Okay.

cmake/ext_project_autotools.cmake
Totktonada: Nice!

igormunkin:
@nshy, does this PR also fix #8026?

nshy:
I didn't tested by it looks like so. Merely because a typo in BYPRODUCT list is fixed.

cmake/ext_project_autotools.cmake
Totktonada: We discussed the `*.in` -> `./config.status` arrow with Nikolay.

There are `*.in` files, which are part of the repository, and there are `*.in` files that are generated by `autoreconf -i`. Changing of the former ones don't obligate us to pass over the whole autoreconf->configure->make pipeline. We can just call `./config.status` (which is effectively 'repeat ./configure with previously passed options') and `make`, it is faster.

Changing of generated `*.in` files means that `*.am` files were changed, so we'll pass over the whole autoreconf->configure->make pipeline anyway.

So the graph looks logical for me.

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,121 @@
...
+    # Use artificial timestamp file to make a dependency with main target.
+    # We can't use just ${config_files_in} counterparts because
+    # depending on configure options not all counterparts could be generated
+    # by config.status and as a result it will be rerun unconditionally.
+    add_custom_command(
+        OUTPUT
+            ${ARGS_DIR}/config.stamp.tarantool
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND ./config.status
+        COMMAND touch config.stamp.tarantool
```
Totktonada, line 96: I usually use in-source build (`cmake . <...> && make -j`) and the `config.stamp.tarantool` file is generated in `third_party/libunwind`. In this case it causes the submodule to be marked as containing untracked content:

> modified:   third_party/libunwind (untracked content)

We can either choose a gitignored directory within libunwind's sources (say, `config`) or one in tarantool sources (say, `build`).

Totktonada (approved):
I see no problems visually and it seems working for me.

cmake/ext_project_autotools.cmake
nshy: Fortunately `config.log` timestamp is updated on `config.status` run thus I don't need artificial `config.stamp.tarantool`. `config.log` is in libunwind's `.gitignore` and other other sane third patry projects will have it in `.gitignore` too probably.

So modified this place the next way:
```diff
--- a/cmake/ext_project_autotools.cmake
+++ b/cmake/ext_project_autotools.cmake
@@ -84,20 +84,23 @@ function(ext_project_autotools name)
         DEPENDS ${ARGS_DIR}/configure
     )

     # Use artificial timestamp file to make a dependency with main target.
     # We can't use just ${config_files_in} counterparts because
     # depending on configure options not all counterparts could be generated
     # by config.status and as a result it will be rerun unconditionally.
     add_custom_command(
         OUTPUT
-            ${ARGS_DIR}/config.stamp.tarantool
+            ${ARGS_DIR}/config.log
         WORKING_DIRECTORY ${ARGS_DIR}
         COMMAND ./config.status
-        COMMAND touch config.stamp.tarantool
         DEPENDS
             ${config_files_in}
             # Actually we depend on config.status but it updates
             # it's own timestamp on every run.
             ${ARGS_DIR}/Makefile
     )

@@ -114,7 +117,7 @@ function(ext_project_autotools name)
             ${make}
         DEPENDS
             ${ARGS_DIR}/Makefile
-            ${ARGS_DIR}/config.stamp.tarantool
+            ${ARGS_DIR}/config.log
         BYPRODUCTS
             ${byproducts}
     )
```

cmake/ext_project_autotools.cmake
```diff
@@ -0,0 +1,124 @@
...
+        WORKING_DIRECTORY ${ARGS_DIR}
+        COMMAND ${CMAKE_SOURCE_DIR}/${ARGS_DIR}/configure ${ARGS_CONFIGURE}
+        DEPENDS ${ARGS_DIR}/configure
+    )
+
+    # Some *.in files are products of *.am files and some are not. For the
+    # latter we don't need to run autoreconf && configure as it will be
+    # overkill. We only need to run config.status.
+    #
+    # Use artificial timestamp file to make a dependency with main target.
```
Totktonada, line 91: This phrase was about `config.stamp.tarantool` and maybe now it worth to update it a bit.

nshy: Removed this paragraph altogether. I guess using `config.log` as dependency between commands looks sane enough so I don't need to explain why I don't used `*.in` files. As using them anyway would involve extra code to make a list of `*.in` files counterparts in comparison with `config.log`.

igormunkin:
> I didn't tested by it looks like so. Merely because a typo in BYPRODUCT list is fixed.

OK, thanks! I'll try your patch on my branch when it is merged into the trunk.

alyapunov (approved):

PR #8021, opened 2022-12-07
ci/cd: drop support for OpenSUSE 15.1/15.2

ylobankov:
It was decided not to proceed with support for these distros.

Closes tarantool/infra#162


kyukhin (approved):

kyukhin:
Removed from 2.10 as well.

Totktonada:
Let's also eliminate `%if (0%{?sle_version} >= 1500)` blocks from `rpm/tarantool.spec`. There is no sense to keep the dead code.

PR #8024, opened 2022-12-08
Drop code of SQL constraints

ImeevMA:
This patch-set drops code of SQL CK and FK constraints as they were replaced by BOX constraints. However, there is system space `_fk_constraint` and `_ck_constraint`, which will be dropped later.

This patch-set drops code of SQL CK and FK constraint as they have been replaced by BOX constraints. However, there is a system space `_fk_constraint` and `_ck_constraint` which will be dropped later.


src/box/sql/fk_constraint.c
```diff
@@ -1,633 +0,0 @@
...
-		 * transaction.
-		 */
-		assert(incr_count == 1);
-		const char *err = tt_sprintf(tnt_errcode_desc(ER_SQL_EXECUTE),
-					     "FOREIGN KEY constraint failed");
-		sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, err,
-			      P4_STATIC);
-		sqlVdbeAddOp1(v, OP_Halt, -1);
-	} else {
-		sqlVdbeAddOp2(v, OP_FkCounter, false, incr_count);
```
Gerold103, line 286: `OP_FkCounter` can be deleted.

src/box/sql/pragma.c
```diff
@@ -291,34 +289,9 @@ sql_pragma_index_list(struct Parse *parse, const char *tbl_name)
...
-					 parent->def->name, child_col,
-					 parent_col, "no_action",
-					 "no_action", "NONE");
-			sqlVdbeAddOp2(v, OP_ResultRow, 1, 8);
-		}
-		++i;
-	}
+	(void)parse_context;
+	(void)table_name;
+	return;
```
Gerold103, line 294: If this pragma is not documented, then why not drop it completely? Otherwise it should probably return the fkeys, like before.

src/box/ck_constraint.c
```diff
@@ -1,282 +0,0 @@
...
- * @retval -1 On error.
- */
-static int
-ck_constraint_resolve_field_names(struct Expr *expr,
-				  struct space_def *space_def)
-{
-	struct Parse parser;
-	sql_parser_create(&parser, sql_get(), default_flags);
-	parser.parse_only = true;
-	sql_resolve_self_reference(&parser, space_def, NC_IsCheck, expr);
```
Gerold103, line 91: AFAIS, `NC_IsCheck` is never set. Can probably drop it and all its checks.

Gerold103 (changes requested):
Thanks for the patch! A couple of nits. Also a question - why not delete `_fk_constraint` and `_ck_constraint` spaces now?

src/box/sql/fk_constraint.c
ImeevMA: Thanks! Dropped `OP_FkCounter` and `OP_FkIfZero`.

src/box/sql/pragma.c
ImeevMA: I've been asked to provide a way to show all FKs in a space, but I'm still not sure if it's going to be this pragma or something new. If we decide to do something new, I will drop this pragma.

src/box/ck_constraint.c
ImeevMA: Thanks! Dropped.

ImeevMA:
> Also a question - why not delete `_fk_constraint` and `_ck_constraint` spaces now?

Deletion of a system space is quite a problem, as far as I know. I know no proper way to drop system space.

Gerold103 (approved):
Thanks for the fixes!

src/box/alter.cc
```diff
@@ -5799,7 +5239,6 @@ TRIGGER(on_replace_sequence, on_replace_dd_sequence);
 TRIGGER(on_replace_sequence_data, on_replace_dd_sequence_data);
 TRIGGER(on_replace_space_sequence, on_replace_dd_space_sequence);
 TRIGGER(on_replace_trigger, on_replace_dd_trigger);
-TRIGGER(on_replace_fk_constraint, on_replace_dd_fk_constraint);
```
locker, line 5802: What happens when I update Tarantool from 2.10 to 2.11 *without* running `box.schema.upgrade()`. SQL constraints will stop working until I run `box.schema.upgrade()`?

ImeevMA: I asked @unera and we came to the conclusion that Tarantool 2.11 should not work properly with a snapshot from Tarantool 2.10 without a proper upgrade. I believe that partly to avoid such problems, we are working on the DOWNGRADE feature.

unera: I think it is OK

But I've just touched @Mons too, Maybe he has another other opinion

Mons: I don't think that's about downgrade feature.

In general, such a change is not ok, because online upgrade is one of the main scenarios of upgrade.
And loss of some functionality during an upgrade could be a problem.
But currently we have only a little usecases for SQL and we hope that almost nobody uses current sql constraints.
So, for this current case let's consider, that temporary loss of constraints is ok (due to lack of dev resources to support complete online upgrade and for the sake of code simplicity)

src/box/schema.cc
```diff
@@ -288,7 +288,7 @@ schema_init(void)
 	key_parts[1].fieldno = 1; /* child space */
 	key_parts[1].type = FIELD_TYPE_UNSIGNED;
 	sc_space_new(BOX_FK_CONSTRAINT_ID, "_fk_constraint", key_parts, 2,
-		     &on_replace_fk_constraint);
+		     NULL);
```
locker, line 292: Why don't you simply remove this system space definition now? It doesn't seem to be required. Also, why not drop the space in the upgrade script?

ImeevMA: Do you know a way to drop system space with primary index defined? Method `drop()` is not working:
```
tarantool> box.space._fk_constraint:drop()
---
- error: Can't drop the primary key in a system space, space '_fk_constraint'
...
```
I think we can add an internal function to remove system space. Should I do it in this PR?

locker: > Do you know a way to drop system space with primary index defined?

I guess you could update the check to allow removal of these two old spaces. Their ids should be removed too (replaced with 'RESERVED' values).

> I think we can add an internal function to drop system space.

A space is removed by deleting a record from a system space. I don't imagine how we could circumvent that.

> Should I do it in this PR?

Without it, this PR shouldn't be called "**Completely** drop SQL constraints" :)

ImeevMA: I will create a new PR where I'll drop these system spaces. For now, I've renamed the PR and edited the commit messages.

locker (approved):

Issue #7960, opened 2022-11-22
Crash in libunwind when collecting backtrace for fiber gc leak check


nshy:
CI run for workflow `debug-aarch64` for [PR 7893](https://github.com/tarantool/tarantool/pull/7893) fail with crash in libunwind:

```
[037] not ok 1	gh-6310-grant-rw-access-on-_session_settings-space-to-public-role.test_boostrap
[037] #   Peer closed
[037] #   stack traceback:
[037] #   	builtin/box/net_box.lua:817: in function '_request'
[037] #   	builtin/box/net_box.lua:878: in function 'eval'
[037] #   	...ccess_on__session_settings_space_to_public_role_test.lua:22: in function 'gh-6310-grant-rw-access-on-_session_settings-space-to-public-role.test_boostrap'
[037] #   	...
[037] #   	[C]: in function 'xpcall'
[037] upgrade | Segmentation fault
[037] upgrade |   code: SEGV_MAPERR
[037] upgrade |   addr: 0xc4200000000
[037] upgrade |   context: 0xffffa897c9b0
[037] upgrade |   siginfo: 0xffffa897c930
[037] upgrade | Current time: 1669044311
[037] upgrade | Please file a bug at https://github.com/tarantool/tarantool/issues
[037] upgrade | Attempting backtrace... Note: since the server has already crashed, 
[037] upgrade | this may fail as well
[037] upgrade | #1  0xaaaab1bd70ac in crash_collect+196
[037] upgrade | #2  0xaaaab1bd8c88 in crash_signal_cb+92
[037] upgrade | #3  0xffffac11778c in ??+0
[037] upgrade | #4  0xaaaab1f5a1f8 in access_mem+32
[037] upgrade | #5  0xaaaab1f577cc in _ULaarch64_step+540
[037] upgrade | #6  0xaaaab1f56e04 in unw_backtrace+340
[037] upgrade | #7  0xaaaab1be8ce0 in collect_current_stack+32
[037] upgrade | #8  0xaaaab1be8ddc in backtrace_collect+228
[037] upgrade | #9  0xaaaab1be3fec in fiber_on_gc_alloc+144
[037] upgrade | #10 0xaaaab1e08f68 in region_alloc+192
[037] upgrade | #11 0xaaaab1e0e644 in tuple_field_map_create_plain+716
[037] upgrade | #12 0xaaaab1e0eb04 in tuple_field_map_create+288
[...]

PR #8030, opened 2022-12-08
cmake: enable leak backtraces on AArch64

Gumix:
**cmake: add -fasynchronous-unwind-tables compiler flag**

This option enables emission of the DWARF CFI (Call Frame Information) directives to the assembler.
It is enabled by default on most compilers, but on GCC 7 for AArch64 and older it wasn't [^1], so turn it on explicitly.
When enabled, the compiler emits `.cfi_*` directives that are required for the stack unwinding, and defines `__GCC_HAVE_DWARF2_CFI_ASM`.

**cmake: enable leak backtraces on AArch64**

This feature was disabled due to a crash in libunwind. After commit https://github.com/tarantool/tarantool/commit/5b08d71ab47474f4278cdddf481bb8e50df074a5 ("libunwind: use latest release v1.6.2 as a base") the crash is gone.

Closes https://github.com/tarantool/tarantool/issues/7960

[^1]: https://gcc.gnu.org/pipermail/gcc-patches/2018-March/495549.html

### DIFF
--- a/cmake/compiler.cmake
+++ b/cmake/compiler.cmake
@@ -110,6 +110,20 @@ set (CMAKE_CXX_FLAGS_RELWITHDEBINFO
 
 unset(CC_DEBUG_OPT)
 
+# Set flags for all include files: those maintained by us and
+# coming from third parties.
+# Since we began using luajit, which uses gcc stack unwind
+# internally, we also need to make sure all code is compiled
+# with unwind info.
+add_compile_flags("C;CXX" "-fexceptions" "-funwind-tables")
+
+# Enable emission of the DWARF CFI (Call Frame Information) directives to the
+# assembler. This option is enabled by default on most compilers, but on GCC 7
+# for AArch64 and older it wasn't, so turn it on explicitly. When enabled, the
+# compiler emits .cfi_* directives that are required for the stack unwinding,
+# and defines __GCC_HAVE_DWARF2_CFI_ASM, which is checked below.
+add_compile_flags("C;CXX" "-fasynchronous-unwind-tables")
+
 check_c_source_compiles(
     "
     #if defined(__x86_64__) && !__has_attribute(force_align_arg_pointer)
@@ -157,16 +171,6 @@ if(BUILD_STATIC AND NOT TARGET_OS_DARWIN)
     # Static linking for c++ routines
     add_compile_flags("C;CXX" "-static-libstdc++")
 endif()
-
-#
-# Set flags for all include files: those maintained by us and
-# coming from third parties.
-# Since we began using luajit, which uses gcc stack unwind
-# internally, we also need to make sure all code is compiled
-# with unwind info.
-#
-
-add_compile_flags("C;CXX" "-fexceptions" "-funwind-tables")
 
 # In C a global variable without a storage specifier (static/extern) and
 # without an initialiser is called a ’tentative definition’. The
--- a/src/trivia/config.h.cmake
+++ b/src/trivia/config.h.cmake
@@ -75,10 +75,6 @@
  * Defined if configured with ABORT_ON_LEAK.
  */
 #cmakedefine ABORT_ON_LEAK 1
-/*
- * Defined if collecting leak backtrace have no issues.
- */
-#cmakedefine ENABLE_LEAK_BACKTRACE 1
 
 /*
  * Set if the system has bfd.h header and GNU bfd library.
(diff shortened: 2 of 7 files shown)


src/trivia/config.h.cmake
```diff
@@ -76,7 +76,7 @@
  */
 #cmakedefine ABORT_ON_LEAK 1
 /*
- * Defined if collecting leak backtrace have no issues.
+ * Defined if configured with ENABLE_BACKTRACE.
```
nshy, line 79: I would keep the coment unchanged. We introduce `ENABLE_LEAK_BACKTRACE` to be able to disable the feature when something go wrong in some situations. So the comment reflects this. After the patch `ENABLE_LEAK_BACKTRACE` will be an alias for `ENABLE_BACKTRACE` but tomorrow it can be changed again.

On the other hand `ENABLE_LEAK_BACKTRACE` seems to be a workaround for the this particular bug. May be we should better disable `ENABLE_BACKTRACE` for aarch64 at that moment instead and not introduce `ENABLE_LEAK_BACKTRACE` at all. In this case we should just replace `ENABLE_LEAK_BACKTRACE` with `ENABLE_BACKTRACE` in the code. @locker what do you think?

nshy (changes requested):

src/trivia/config.h.cmake
locker: It doesn't look like we need `ENABLE_LEAK_BACKTRACE` now, when the bug is fixed. Let's replace it with `ENABLE_BACKTRACE`. I agree that if we find another issue with libunwind on aarch64, we'll be better off disabling `ENABLE_BACKTRACE` on aarch64 altogether.

CuriousGeorgiy: > I agree that if we find another issue with libunwind on aarch64, we'll be better off disabling ENABLE_BACKTRACE on aarch64 altogether.

Sounds arguable. I agree with @nshy we should retain this option for a while (and leave the comment untouched).

CuriousGeorgiy: > I agree that if we find another issue with libunwind on aarch64, we'll be better off disabling ENABLE_BACKTRACE on aarch64 altogether.

Sounds arguable. I agree with @nshy we should retain this option for a while (and leave the comment untouched).

cmake/compiler.cmake
```diff
@@ -110,6 +110,13 @@ set (CMAKE_CXX_FLAGS_RELWITHDEBINFO
 
 unset(CC_DEBUG_OPT)
 
+# Enable emission of the DWARF CFI (Call Frame Information) directives to the
+# assembler. This option is enabled by default on most compilers, but on GCC 7
+# for AArch64 and older it wasn't, so turn it on explicitly. When enabled, the
+# compiler emits .cfi_* directives that are required for the stack unwinding,
+# and defines __GCC_HAVE_DWARF2_CFI_ASM, which is checked below.
+add_compile_flags("C;CXX" "-fasynchronous-unwind-tables")
```
CuriousGeorgiy, line 118: I would rather add this flag here, instead of -funwind-tables:
https://github.com/tarantool/tarantool/blob/05002edf80d44a4a82da8be0d5e7bcda5349d47a/cmake/compiler.cmake#L161-L169

CuriousGeorgiy (changes requested):

cmake/compiler.cmake
Gumix: This won't work because `add_compile_flags` must be called before `check_c_source_compiles`.

`-fexceptions` probably can be removed, I'm not sure.

CuriousGeorgiy: > -fexceptions probably can be removed, I'm not sure.

Nope, it can't be removed, since it is disabled by default for C:
> If you do not specify this option, GCC enables it by default for languages like C++ that normally require exception handling, and disables it for languages like C that do not normally require it.[^1]

[^1]: https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#Code-Gen-Options

CuriousGeorgiy: > This won't work because `add_compile_flags` must be called before `check_c_source_compiles`.

How about moving this check after adding necessary compile flags?

Gumix: Oops, I meant «`-funwind-tables` probably can be removed».

src/trivia/config.h.cmake
Gumix: Removed `ENABLE_LEAK_BACKTRACE`.

cmake/compiler.cmake
Gumix: > How about moving this check after adding necessary compile flags?

Done.

nshy (approved):

CuriousGeorgiy (approved):

locker (approved):

Issue #6998, opened 2022-04-06
libunwind submodule follow-up fixes


CuriousGeorgiy:
After reviewing #6877, @Totktonada proposed some suggestions and fixes:
```[tasklist]
### Tasks
- [x] return _autotools_ to _RHEL_ and _debian_ distribution dependencies;
- [x] remove _backtrace_ feature compiler dependency from [_rpm spec_](https://github.com/tarantool/tarantool/blob/master/rpm/tarantool.spec);
- [x] fix _libunwind_ submodule [changelog](https://github.com/tarantool/tarantool/blob/master/changelogs/unreleased/add-libunwind-submodule.md): license and _backtrace_ feature compiler dependency in _rpm spec_;
- [x] clarify compatibility of _libunwind_ and _Tarantool_ with compressed debug symbols;
- [x] add _libunwind_ license to [_debian/copyright_](https://github.com/tarantool/tarantool/blob/master/debian/copyright);
- [x] remove __libunwind__ package depencies from _rpm spec_ and [_debian/control_](https://github.com/tarantool/tarantool/blob/master/debian/control).
- [x] Re-verify elimination of `-DCMAKE_FIND_USE_CMAKE_SYSTEM_PATH=FALSE` (https://github.com/tarantool/tarantool/pull/6877#discussion_r936010410)
```

H9660:
I can fix this issue considering my expertise. Please assign it to me.

H9660:
Thanks for assigning me this

Totktonada:
@CuriousGeorgiy Is it planned for the upcoming release?

CuriousGeorgiy:
@Totktonada please address this to @alyapunov.

Totktonada:
We have bundled libunwind based on [this](https://github.com/libunwind/libunwind/commit/879ad87300a12f70c365fc16592c94cdf9f07d86) upstream revision. It is reported by `git describe --tags` as v1.5-rc1-112-g879ad873. The branch in our [downstream](https://github.com/tarantool/libunwind) is named `libunwind-1.6.2-tarantool`, but it does not seems to be v1.6.2.

@CuriousGeorgiy Can you clarify, whether it is a mistake or I missed something? Whether the idea was to bundle the latest release (v1.6.2)?

CuriousGeorgiy:
@Totktonada yes, the idea was to bundled the latest release.

CuriousGeorgiy:
@Gumix While you are here, could please fix this warning?

```console
[...]

PR #6877, opened 2022-03-01
libunwind: add new submodule

CuriousGeorgiy:
Investigation of GNU libunwind problems on the aarch64-linux-gnu
platform drive us to the conclusion that libunwind-1.2.1 provided by
major distribution packages is broken. Not to mention that its test
suite fails with SEGFAULTs.

Last but not least, some distributions, e.g. CentOS 8 (see #4611) do
not provide a libunwind package.

Hence, bundle libunwind: bundling is enabled by default on all
platforms, except for macOS — a system package can be used if its
version is greater or equal than 1.3.0 (minimal version that does not
seem to be broken on aarch64-linux-gnu).

* Add new submodule: bump it to current master
* Refactor libunwind package search logic out of compiler.cmake.
* Add CMake script for building bundled libunwind
* Add CMake script for extracting version of libunwind
* Re-enable backtrace for all RHEL distributions by default

Needed for #4002
Closes #4611

PR #8031, opened 2022-12-08
build: fix libunwind.h search in testing files

### LINKED ISSUE #8025: Tarantool doesn't build if libundwind.h is unavailable
Tarantool 2.11.0-entrypoint-764-g1b15fabbcf32

Build fails with:

```
In file included from /home/vlad/src/tarantool/tarantool/src/lib/core/fiber.h:47,
                 from /home/vlad/src/tarantool/tarantool/test/box-tap/check_merge_source.c:4:
/home/vlad/src/tarantool/tarantool/src/lib/core/backtrace.h:14:10: fatal error: libunwind.h: No such file or directory
   14 | #include "libunwind.h"
      |          ^~~~~~~~~~~~~
```

But it should build just fine, because we use a bundled libunwind.

Triggered by commit https://github.com/tarantool/tarantool/commit/19abfd2a39205de270836adda7f3a733476acaa4, but the root cause is that `test/box-tap/check_merge_source.c` is built as a module, but uses internal headers.

Totktonada:
## The problem

Case: a build host has no libunwind installed into the system. Dispite that tarantool has libunwind bundled as a git submodule (see PR #6877), the build fails in the case:

```c
In file included from <...>/src/lib/core/fiber.h:47,
                 from <..some test/**/*.{c,cc} file..>:
<...>/src/lib/core/backtrace.h:14:10: fatal error: libunwind.h: No such file or directory
   14 | #include "libunwind.h"
      |          ^~~~~~~~~~~~~
```

The problem appears when `ENABLE_BACKTRACE` and `ENABLE_BUNDLED_LIBUNWIND` CMake options are enabled. They're enabled by default for most targets.

The compilation fails for testing files, which include `fiber.h`. The `fiber.h` header includes `backtrace.h`, which includes `libunwind.h`. The `libunwind.h` header couldn't be found if appropriate include directory is not passed to a compiler.

## The solution

* Add libunwind's include directory to compilation of unit tests and fuzzers.
* Eliminate `fiber.h` usage in a merger's test using `tnt_internal_symbol()` (see 395c30e8f85d).
* Drop libunwind build dependency from rpm/deb packages to verify the fix (and because it is unnecessary after #6877).

See details in the commit messages.

Follows up #6877
Part of #6998
Fixes #8025

### DIFF
--- a/test/fuzz/CMakeLists.txt
+++ b/test/fuzz/CMakeLists.txt
@@ -3,6 +3,13 @@ include_directories(${PROJECT_SOURCE_DIR}/src)
 include_directories(${PROJECT_BINARY_DIR}/src)
 include_directories(${PROJECT_SOURCE_DIR}/src/box)
 
+# fiber.h may require libunwind.h. Add proper header search paths.
+function(add_libunwind_include_directory target)
+  if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+    target_include_directories(${target} SYSTEM PRIVATE ${LIBUNWIND_INCLUDE_DIR})
+  endif()
+endfunction()
+
 # A special target with fuzzer and sanitizer flags.
 add_library(fuzzer_config INTERFACE)
 
@@ -40,9 +47,11 @@ target_link_libraries(http_parser_fuzzer PUBLIC http_parser fuzzer_config)
 
 add_executable(swim_proto_member_fuzzer swim_proto_member_fuzzer.c)
 target_link_libraries(swim_proto_member_fuzzer PUBLIC swim fuzzer_config)
+add_libunwind_include_directory(swim_proto_member_fuzzer)
 
 add_executable(swim_proto_meta_fuzzer swim_proto_meta_fuzzer.c)
 target_link_libraries(swim_proto_meta_fuzzer PUBLIC swim fuzzer_config)
+add_libunwind_include_directory(swim_proto_meta_fuzzer)
 
 add_executable(datetime_parse_full_fuzzer datetime_parse_full_fuzzer.c)
 target_link_libraries(datetime_parse_full_fuzzer PUBLIC core fuzzer_config)
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -15,6 +15,15 @@ include_directories(${PROJECT_SOURCE_DIR}/src/box)
 include_directories(${PROJECT_SOURCE_DIR}/third_party)
 include_directories(${ICU_INCLUDE_DIRS})
 include_directories(${CURL_INCLUDE_DIRS})
+
+# There are many unit tests that include fiber.h, which may
+# transitively include libunwind.h. When libunwind is built as
+# part of tarantool's build process, we should add libunwind's
+# headers directory into the search path.
+if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+  include_directories(SYSTEM ${LIBUNWIND_INCLUDE_DIR})
+endif()
+
 include_directories(${EXTRA_CORE_INCLUDE_DIRS})
 
 include(CMakeParseArguments)
--- a/changelogs/unreleased/fix-libunwind-h-search.md
+++ b/changelogs/unreleased/fix-libunwind-h-search.md
@@ -1,0 +1,3 @@
+## bugfix/build
+
+* Fixed `libunwind.h` search in testing files (gh-6877).
--- a/debian/control
+++ b/debian/control
@@ -22,7 +22,6 @@ Build-Depends: cdbs (>= 0.4.100), debhelper (>= 9), dpkg-dev (>= 1.16.1~),
  libreadline-dev,
  libncurses5-dev,
  libssl-dev,
- libunwind-dev | libunwind8-dev | libunwind7-dev,
  libicu-dev,
 # libcurl build dependencies
  zlib1g-dev,
--- a/rpm/tarantool.spec
+++ b/rpm/tarantool.spec
@@ -75,7 +75,6 @@ Requires(preun): initscripts
 %undefine __cc
 
 %if %{with backtrace}
-BuildRequires: libunwind-devel
 #
 # Disable stripping of /usr/bin/tarantool to allow the debug symbols
 # in runtime. Tarantool uses the debug symbols to display fiber's stack
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -2096,3 +2096,9 @@ int fiber_stat(fiber_stat_cb cb, void *cb_ctx)
 	}
 	return 0;
 }
+
+struct lua_State *
+fiber_lua_state(struct fiber *f)
+{
+	return f->storage.lua.stack;
+}
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -1156,6 +1156,12 @@ extern bool fiber_abort_on_gc_leak;
 void
 fiber_check_gc(void);
 
+/**
+ * Return fiber's Lua state or NULL if it doesn't exist.
+ */
+struct lua_State *
+fiber_lua_state(struct fiber *f);
+
 #if defined(__cplusplus)
 } /* extern "C" */
 
--- a/src/symbols.c
+++ b/src/symbols.c
@@ -17,6 +17,11 @@
  * Tarantool 2.8.
  *
  * [1]: https://github.com/picodata/tarantool-module
+ *
+ * fiber_lua_state() is used in [2] to eliminate dependency from
+ * fiber.h. See gh-8025.
+ *
+ * [2]: test/box-tap/check_merge_source.c
  */
 
 extern void
@@ -45,6 +50,8 @@ extern void
 ipc_value_delete(void);
 extern void
 ipc_value_new(void);
+extern void
+fiber_lua_state(void);
 
 /** Symbol definition. */
 struct symbol_def {
@@ -68,6 +75,7 @@ static struct symbol_def symbols[] = {
 	{"fiber_channel_put_timeout", fiber_channel_put_timeout},
 	{"ipc_value_delete", ipc_value_delete},
 	{"ipc_value_new", ipc_value_new},
+	{"fiber_lua_state", fiber_lua_state},
 	{NULL, NULL}
 };
 
--- a/test/app-luatest/tnt_internal_symbol_test.lua
+++ b/test/app-luatest/tnt_internal_symbol_test.lua
@@ -24,6 +24,7 @@ g.test_fiber_channel = function()
         'fiber_channel_put_timeout',
         'ipc_value_delete',
         'ipc_value_new',
+        'fiber_lua_state',
     }
     -- Verify only that we can get the address. The functions are
     -- provided 'as is' and we don't guarantee anything regarding
--- a/test/box-tap/CMakeLists.txt
+++ b/test/box-tap/CMakeLists.txt
@@ -1,4 +1,1 @@
-# fiber.h requires tarantool_ev.h from third_party directory.
-include_directories(${PROJECT_SOURCE_DIR}/third_party)
-
 build_module(check_merge_source check_merge_source.c)
--- a/test/box-tap/check_merge_source.c
+++ b/test/box-tap/check_merge_source.c
@@ -1,10 +1,8 @@
+#include <assert.h>
 #include <lua.h>
 #include <lauxlib.h>
-#include "lib/core/diag.h"
-#include "fiber.h"
-#include "lua/utils.h"
+#include <module.h>
 #include "box/merger.h"
-#include "box/error.h"
 
 /**
  * Verify whether a temporary fiber-local Lua state has the same
@@ -25,6 +23,21 @@
  * module, but the stub that imitates usage of a merge source from
  * tarantool code.
  */
+
+/*
+ * Here we're going the dark way. We should verify a property of
+ * an object that is not reachable through the public C API.
+ */
+void *
+tnt_internal_symbol(const char *name);
+
+/*
+ * The idea of the `call_next()` check is to verify properties of
+ * the fiber's Lua state. Let's define a pointer to the accessor
+ * function.
+ */
+static struct lua_State *
+(*fiber_lua_state)(struct fiber *f) = NULL;
 
 /**
  * Extract a merge source from the Lua stack.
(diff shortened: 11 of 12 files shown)


Totktonada:
I also made two attempts to eliminate `box/merger.h` include from the merger's test file. Both are working, but I like neither of them. So I disregard it.

<details>
<summary>Attempt 1: use `tnt_internal_symbol()`.</summary>

```diff
diff --git a/src/box/merger.c b/src/box/merger.c
index fff12f9aa..f2582a141 100644
--- a/src/box/merger.c
+++ b/src/box/merger.c
@@ -47,6 +47,18 @@
 #include "box/key_def.h"      /* key_def_*(),
 				 tuple_compare() */
 
+
+/* {{{ Base merge source functions */
+
+int
+merge_source_next(struct merge_source *source, struct tuple_format *format,
+		  struct tuple **out)
+{
+	return source->vtab->next(source, format, out);
+}
+
+/* }}} */
+
 /* {{{ Merger */
 
 /**
diff --git a/src/box/merger.h b/src/box/merger.h
index 90ae175a9..8a360b90b 100644
--- a/src/box/merger.h
+++ b/src/box/merger.h
@@ -111,12 +111,9 @@ merge_source_unref(struct merge_source *source)
 /**
  * @see merge_source_vtab
  */
-static inline int
+int
 merge_source_next(struct merge_source *source, struct tuple_format *format,
-		  struct tuple **out)
-{
-	return source->vtab->next(source, format, out);
-}
+		  struct tuple **out);
 
 /**
  * Initialize a base merge source structure.
diff --git a/src/symbols.c b/src/symbols.c
index f352b44d2..c00d39aa8 100644
--- a/src/symbols.c
+++ b/src/symbols.c
@@ -18,8 +18,11 @@
  *
  * [1]: https://github.com/picodata/tarantool-module
  *
- * fiber_lua_state() is used in [2] to eliminate dependency from
- * fiber.h. See gh-8025.
+ * fiber_lua_state() and merge_source_next() are used in [2].
+ * The purpose is elimination of dependency from fiber.h (see
+ * gh-8025) and to distinguish usages of the public API and
+ * the internal API in an unambiguous matter in the testing
+ * code.
  *
  * [2]: test/box-tap/check_merge_source.c
  */
@@ -52,6 +55,8 @@ extern void
 ipc_value_new(void);
 extern void
 fiber_lua_state(void);
+extern void
+merge_source_next(void);
 
 /** Symbol definition. */
 struct symbol_def {
@@ -76,6 +81,7 @@ static struct symbol_def symbols[] = {
 	{"ipc_value_delete", ipc_value_delete},
 	{"ipc_value_new", ipc_value_new},
 	{"fiber_lua_state", fiber_lua_state},
+	{"merge_source_next", merge_source_next},
 	{NULL, NULL}
 };
 
diff --git a/test/app-luatest/tnt_internal_symbol_test.lua b/test/app-luatest/tnt_internal_symbol_test.lua
index 40d53cb01..280040d6e 100644
--- a/test/app-luatest/tnt_internal_symbol_test.lua
+++ b/test/app-luatest/tnt_internal_symbol_test.lua
@@ -25,6 +25,7 @@ g.test_fiber_channel = function()
         'ipc_value_delete',
         'ipc_value_new',
         'fiber_lua_state',
+        'merge_source_next',
     }
     -- Verify only that we can get the address. The functions are
     -- provided 'as is' and we don't guarantee anything regarding
diff --git a/test/box-tap/check_merge_source.c b/test/box-tap/check_merge_source.c
index f5c46fc3f..bba33bd81 100644
--- a/test/box-tap/check_merge_source.c
+++ b/test/box-tap/check_merge_source.c
@@ -2,7 +2,6 @@
 #include <lua.h>
 #include <lauxlib.h>
 #include <module.h>
-#include "box/merger.h"
 
 /**
  * Verify whether a temporary fiber-local Lua state has the same
@@ -39,6 +38,15 @@ tnt_internal_symbol(const char *name);
 static struct lua_State *
 (*fiber_lua_state)(struct fiber *f) = NULL;
 
+struct merge_source;
+
+/*
+ * XXX
+ */
+static int
+(*merge_source_next)(struct merge_source *source, struct tuple_format *format,
+		     struct tuple **out) = NULL;
+
 /**
  * Extract a merge source from the Lua stack.
  */
@@ -116,6 +124,9 @@ luaopen_check_merge_source(struct lua_State *L)
 	fiber_lua_state = tnt_internal_symbol("fiber_lua_state");
 	assert(fiber_lua_state != NULL);
 
+	merge_source_next = tnt_internal_symbol("merge_source_next");
+	assert(merge_source_next != NULL);
+
 	static const struct luaL_Reg meta[] = {
 		{"call_next", lbox_check_merge_source_call_next},
 		{NULL, NULL}
```
</details>

I don't want to make a short `foo->bar->baz(a, b, c)` function non-inline just to use it in a test.

<details>
<summary>Attempt 2: copy `merge_source_next()`.</summary>

```diff
diff --git a/test/box-tap/check_merge_source.c b/test/box-tap/check_merge_source.c
index f5c46fc3f..95eaa2e7b 100644
--- a/test/box-tap/check_merge_source.c
+++ b/test/box-tap/check_merge_source.c
@@ -2,7 +2,6 @@
 #include <lua.h>
 #include <lauxlib.h>
 #include <module.h>
-#include "box/merger.h"
 
 /**
  * Verify whether a temporary fiber-local Lua state has the same
@@ -39,6 +38,31 @@ tnt_internal_symbol(const char *name);
 static struct lua_State *
 (*fiber_lua_state)(struct fiber *f) = NULL;
 
+struct merge_source_vtab {
+	/** Free a merge source. */
+	void (*destroy)(struct merge_source *base);
+	/** Get a next tuple (refcounted) from a source. */
+	int (*next)(struct merge_source *base, struct tuple_format *format,
+		    struct tuple **out);
+};
+
+struct merge_source {
+	/* Source-specific methods. */
+	const struct merge_source_vtab *vtab;
+	/* Reference counter. */
+	int refs;
+};
+
+/**
+ * Copy of a box/merger.h function to don't include the whole header.
+ */
+static inline int
+merge_source_next(struct merge_source *source, struct tuple_format *format,
+		  struct tuple **out)
+{
+	return source->vtab->next(source, format, out);
+}
+
 /**
  * Extract a merge source from the Lua stack.
  */
```
</details>

Here we copy not only `merge_source_next()` definition, but also structures declarations. It doesn't look anyhow better than `#include`.

Totktonada:
Added a commit that eliminates libunwind-dev/devel dependency from RPM/Deb packages. It'll prevent the regression: there is no system `libunwind.h` at building rpm/deb packages anymore.

test/unit/CMakeLists.txt
```diff
@@ -30,6 +41,12 @@ function(create_unit_test)
   message(STATUS "Creating unit test ${UNIT_PREFIX}.test")
   add_executable(${UNIT_PREFIX}.test ${UNIT_SOURCES})
   target_link_libraries(${UNIT_PREFIX}.test ${UNIT_LIBRARIES})
+
+  # See LIBUNWIND_INCLUDE_DIR above for explanation.
+  if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+    add_dependencies(${UNIT_PREFIX}.test bundled-libunwind bundled-libunwind-platform)
```
CuriousGeorgiy, line 47: AFAIC, unit tests that use `fiber.h` depend either on `core` or `box` (which depends on `core`) library targets, so this is unnecessary:
https://github.com/tarantool/tarantool/blob/05002edf80d44a4a82da8be0d5e7bcda5349d47a/src/lib/core/CMakeLists.txt#L84-L89

test/unit/CMakeLists.txt
```diff
@@ -15,6 +15,17 @@ include_directories(${PROJECT_SOURCE_DIR}/src/box)
...
+
+# There are many unit tests that include fiber.h, which may
+# transitively include libunwind.h. It requires two actions:
+#
+# 1. Add the header directory into the search paths.
+# 2. Add the dependency to don't try to find libunwind.h before
+#    it'll appear in given directory.
+if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+  include_directories(SYSTEM ${LIBUNWIND_INCLUDE_DIR})
+endif()
```
CuriousGeorgiy, line 27: AFAIC, the included directories are inherited by all subdirectories:
> The include directories are added to the [INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_dir/INCLUDE_DIRECTORIES.html#prop_dir:INCLUDE_DIRECTORIES) directory property for the current CMakeLists file. They are also added to the [INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INCLUDE_DIRECTORIES.html#prop_tgt:INCLUDE_DIRECTORIES) target property for each target in the current CMakeLists file.[^1]

This snippet added to the beginning of this files yields the following result:
```cmake
get_property(INCLUDE_DIRECTORIES DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
message("INCLUDE_DIRECTORIES=${INCLUDE_DIRECTORIES}")
```
```console
INCLUDE_DIRECTORIES=/Users/georgiy.lebedev/Work/tarantool/src;/Users/georgiy.lebedev/Work/tarantool/build-release/src;/Users/georgiy.lebedev/Work/tarantool/src/lib;/Users/georgiy.lebedev/Work/tarantool/src/lib/small/include;/Users/georgiy.lebedev/Work/tarantool/src/lib/small/third_party;/Users/georgiy.lebedev/Work/tarantool/src/lib/core;/Users/georgiy.lebedev/Work/tarantool/third_party;/usr/local/opt/openssl@3/include;/Users/georgiy.lebedev/Work/tarantool/third_party/xxHash;/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include;/Users/georgiy.lebedev/Work/tarantool/third_party/luajit/src;/Users/georgiy.lebedev/Work/tarantool/build-release/third_party/luajit/src;/Users/georgiy
```

Perhaps, we should rather move all `include_directories` commands from [`src/CMakeLists.txt`](https://github.com/tarantool/tarantool/blob/05002edf80d44a4a82da8be0d5e7bcda5349d47a/src/CMakeLists.txt#L6-L18) to [`CMakeLists.txt`](https://github.com/tarantool/tarantool/blob/05002edf80d44a4a82da8be0d5e7bcda5349d47a/CMakeLists.txt#L293-L302)?

[^1]: https://cmake.org/cmake/help/latest/command/include_directories.html

test/fuzz/CMakeLists.txt
```diff
@@ -3,6 +3,15 @@ include_directories(${PROJECT_SOURCE_DIR}/src)
...
 include_directories(${PROJECT_SOURCE_DIR}/src/box)
 
+# fiber.h may require libunwind.h. Add proper header search paths
+# and dependencies.
+function(add_libunwind_dependency target)
+  if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+    target_include_directories(${target} SYSTEM PRIVATE ${LIBUNWIND_INCLUDE_DIR})
+    add_dependencies(${target} bundled-libunwind bundled-libunwind-platform)
+  endif()
+endfunction()
```
CuriousGeorgiy, line 13: Same comment as for the unit tests.

test/unit/CMakeLists.txt
```diff
@@ -30,6 +41,12 @@ function(create_unit_test)
   message(STATUS "Creating unit test ${UNIT_PREFIX}.test")
   add_executable(${UNIT_PREFIX}.test ${UNIT_SOURCES})
   target_link_libraries(${UNIT_PREFIX}.test ${UNIT_LIBRARIES})
+
+  # See LIBUNWIND_INCLUDE_DIR above for explanation.
+  if (ENABLE_BACKTRACE AND ENABLE_BUNDLED_LIBUNWIND)
+    add_dependencies(${UNIT_PREFIX}.test bundled-libunwind bundled-libunwind-platform)
+  endif()
```
CuriousGeorgiy, line 48: AFAIC, unit tests that use `fiber.h` depend either on `core` or `box` (which depends on `core`) library targets, so this is unnecessary:
https://github.com/tarantool/tarantool/blob/05002edf80d44a4a82da8be0d5e7bcda5349d47a/src/lib/core/CMakeLists.txt#L84-L89

test/fuzz/CMakeLists.txt
```diff
@@ -3,6 +3,15 @@ include_directories(${PROJECT_SOURCE_DIR}/src)
 include_directories(${PROJECT_BINARY_DIR}/src)
```
CuriousGeorgiy, line 3: Is it really worth making two separate commits? I would rather squash them.

CuriousGeorgiy (changes requested):

test/unit/CMakeLists.txt
Totktonada: Your point looks valid, but I actually saw the fail without this dependency. I'll look around and return here with results.

test/fuzz/CMakeLists.txt
Totktonada: I have no strict preference here. I'll do.

test/unit/CMakeLists.txt
Totktonada: I thought on it, but I'm afraid to make major changes in our build scripts without some consistent vision on it. IOW, I would proceed this way for all bundled libraries if we'll be sure that it is the right thing to do. But I wouldn't make it as a fix of particular build problem.

test/unit/CMakeLists.txt
Totktonada: The reason is that `test/unit/uri.c` includes `fiber.h`, but doesn't use it. It has no dependency on `core` and it was the reason of the build fail I saw.

I updated the patchset by eliminating this `add_dependencies()` directive and removing the `fiber.h` include from `test/unit/uri.c`.

I visually verifed that all the unit tests (that fail without `libunwind.h`) have direct or indirect dependency on `core`. I tested the parallel build locally a couple of times and everything looks fine.

Thank you for spotting it!

test/fuzz/CMakeLists.txt
Totktonada: I did.

test/fuzz/CMakeLists.txt
Totktonada: I verified visually that the fuzzers have an indirect dependency on `core` and removed the `add_dependencies()` directive. I built the project with fuzzers locally several times (with `make -j`) and everything looks fine. (All the builds were without system's `libunwind`.)

CuriousGeorgiy (approved):

locker (approved):

locker:
I'm **not** cherry-picking it to 2.10, because in 2.10 `fiber.h` doesn't include `libunwind.h` so everything builds just fine.

PR #8035, opened 2022-12-09
test: tarantoolctl.test flakiness fixup

sergos:
The `recursive_rmdir` can start execution while tarantool instance started by the tarantoolctl is in the process of shutdown. As a result the instance pid file can be available at the name collection, but by the time of its removal the instance deletes it by itself.

NO_DOC=test
NO_CHANGELOG=test


ylobankov:

PR #8040, opened 2022-12-12
cd: eliminate OpenSUSE related code from RPM spec

ylobankov:
Support for OpenSUSE 15.1/15.2 has been recently dropped. There is no sense to keep the dead code. So removing it.

Follows up #8021


Totktonada (approved):

ylobankov:

PR #8041, opened 2022-12-12
ci: fix running full testing on any label when `full-ci` is set

ylobankov:
Previously, setting any label on a pull request with the `full-ci` label
triggered running the full testing. The reason was in not enough strict
condition for running workflows.

Now the full testing will be run only when pull requests are labeled by
the 'full-ci' label, or it already exists, but event's activity type is
not related to labeling.

### DIFF
--- a/.github/workflows/centos_7.yml
+++ b/.github/workflows/centos_7.yml
@@ -34,10 +34,16 @@ concurrency:
 jobs:
   centos_7:
     # Run on push to the 'master' and release branches of tarantool/tarantool
-    # or on pull request if the 'full-ci' label is set.
-    if: github.repository == 'tarantool/tarantool' &&
-        ( github.event_name != 'pull_request' ||
-          contains(github.event.pull_request.labels.*.name, 'full-ci') )
+    # or on pull request if the 'full-ci' label is set or already exists.
+    if: |
+      github.repository == 'tarantool/tarantool' && (
+        github.event_name != 'pull_request' || (
+          ( github.event.action == 'labeled' &&
+            github.event.label.name == 'full-ci' ) ||
+          ( github.event.action != 'labeled' &&
+            contains(github.event.pull_request.labels.*.name, 'full-ci') )
+        )
+      )
 
     runs-on: ubuntu-20.04-self-hosted
 
--- a/.github/workflows/centos_7_aarch64.yml
+++ b/.github/workflows/centos_7_aarch64.yml
@@ -34,10 +34,16 @@ concurrency:
 jobs:
   centos_7_aarch64:
     # Run on push to the 'master' and release branches of tarantool/tarantool
-    # or on pull request if the 'full-ci' label is set.
-    if: github.repository == 'tarantool/tarantool' &&
-        ( github.event_name != 'pull_request' ||
-          contains(github.event.pull_request.labels.*.name, 'full-ci') )
+    # or on pull request if the 'full-ci' label is set or already exists.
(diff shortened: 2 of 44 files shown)


NickVolynkin:
@ylobankov please check what happens if you open a pull request with a label. If I remember correctly, both events will trigger at once: opened and labeled.

NickVolynkin:
Also, you can use https://github.com/tarantool/tarantool/tree/no-ci as a base to avoid triggering too many workflows at once.

ylobankov:
> @ylobankov please check what happens if you open a pull request with a label. If I remember correctly, both events will trigger at once: opened and labeled.

Yes, you are right. But one of them will be cancelled due to `cuncurrency`. I think we should fix the `pull_request` trigger and leave only `[ synchronize, labeled ]`.

ylobankov:
> > @ylobankov please check what happens if you open a pull request with a label. If I remember correctly, both events will trigger at once: opened and labeled.
> 
> Yes, you are right. But one of them will be cancelled due to `cuncurrency`. I think we should fix the `pull_request` trigger for workflows running on 'full-ci' and leave only `[ synchronize, labeled ]`.

Or change the condition 
```
( github.event.action != 'labeled' && 
  contains(github.event.pull_request.labels.*.name, 'full-ci') )
```
to 
```
( github.event.action == 'synchronize' && 
  contains(github.event.pull_request.labels.*.name, 'full-ci') )
```

NickVolynkin:
So, we have two options:
* Labeled with https://github.com/tarantool/tarantool/labels/full-ci
* Synchronized or reopened while having label https://github.com/tarantool/tarantool/labels/full-ci

ylobankov:
> So, we have two options:
> 
> * Just labeled with 
>   [
>     full-ci](https://github.com/tarantool/tarantool/issues?q=is%3Aopen+label%3Afull-ci+sort%3Aupdated-desc)
>     Enables all tests for a pull request
> * Synchronized or reopened while having label 
>   [
>     full-ci](https://github.com/tarantool/tarantool/issues?q=is%3Aopen+label%3Afull-ci+sort%3Aupdated-desc)
>     Enables all tests for a pull request
>   
>   ```yaml
>   ( contains('synchronize, reopened', github.event.action) && 
>   contains(github.event.pull_request.labels.*.name, 'full-ci') )
>   ```

I considered your variant before, but after some time thinking about it, I realized that it is not convenient when I don't have skipped checks when a pull request opens:

1. A person may think that all checks already run and there are no other checks. Skipped checks give a clue that there are tests that should be run as well.
2. From time to time, we need to update almost all workflows for some reason, and sometimes we make a mistake, for example, in the workflow syntax. Skipped workflows help catch such cases, because broken workflows don't appear in the check's list, and it means that something is wrong.

NickVolynkin:
> Skipped workflows help catch such cases, because broken workflows don't appear in the check's list, and it means that something is wrong.

That's correct. 

> Skipped checks give a clue that there are tests that should be run as well.

Screenshot 2022-12-22 at 13 48 08

Indeed, they do. It's probably good to see that there's a lot more checks to run. What we really need is to avoid re-running jobs when they should not re-run. Like when we add some custom label like https://github.com/tarantool/tarantool/issues?q=is%3Aopen+label%3Ado-not-merge+sort%3Aupdated-desc, but there's already https://github.com/tarantool/tarantool/issues?q=is%3Aopen+label%3Afull-ci+sort%3Aupdated-desc on the PR.

Issue #3682, opened 2018-09-11
URL percent-encoding


bigbes:
cURL easy API contains functions `curl_easy_escape` and `curl_easy_unescape` that are respectively analogues of `url_encode` and `url_decode`:

* https://curl.haxx.se/libcurl/c/curl_easy_escape.html
* https://curl.haxx.se/libcurl/c/curl_easy_unescape.html

Old implementation of these functions in `http.server` are slow and inefficient.

RFC: https://www.notion.so/tarantool/http-add-percent-encoding-decoding-of-query-string-in-request-76a2425a4c4744a1a72643527a4fe7f7

PR #7254, opened 2022-06-09
app/uri: add url encode and decode methods

ligurio:
https://datatracker.ietf.org/doc/html/rfc3986.html

Fixes #3682

PR #8042, opened 2022-12-12
http: escape encoded params

### LINKED ISSUE #7931: http client: properly encode query parameters
The HTTP client now has `params` option, where a user may provide a table of query parameters (#6832). Those parameters are encoded into a `?foo=bar&baz=fiz` string verbatim. If a name or a value of a query parameter contains `&`, `=` (or another symbol with specific meaning in the URI query component), the query may be interpreted incorrectly by a server.

We're going to implement encoding for URI components (so called percent-encoding) in the scope of #3682. After this we should enable it for query parameters in the HTTP client (for names and values both).

ligurio:
**Requires commits in #3682 and thus blocked by #3682**

Commits to be reviewed:

```
commit 1950807fdf18fadac99a592eaf9ee238d4bf745a (HEAD -> ligurio/gh-7931-encode-query-params)
Date:   Mon Dec 12 15:50:42 2022 +0300

    http: escape encoded params
    
    @TarantoolBot document
    Title: Document a percent-encoding of params passed to http client
    
    The HTTP client has `params` option, where a user may provide a table of
    query parameters (added in #6832). Those parameters are encoded into a
    `?foo=bar&twedledoo=tweedledee` string verbatim. If a name or a value of
    a query parameter contains `&`, `=` (or any another symbol with specific
    meaning in the URI query component), the query may be interpreted
    incorrectly by a server. Now key and values passed in a table as
    `params` option are percent-encoded and then encoded to a query string.
    This will be made automatically. Percent-encoding depends on used HTTP
    method: with `GET`, `HEAD` and `DELETE` parameters `uri.QUERY_STRING`
    are used and with other HTTP method `uri.FORM_URLENCODED` is used.
    
    Closes #7931
```

```
commit 6166d650d25ac9361df56670c209e053d7a99eeb
Date:   Fri Dec 9 18:23:05 2022 +0300

    uri: escape params
    
    NO_CHANGELOG=internal
    NO_DOC=internal
    
    Needed for #7931
```

Fixes #7931

### DIFF
--- a/changelogs/unreleased/gh-7931-httpc-params-escaping.md
+++ b/changelogs/unreleased/gh-7931-httpc-params-escaping.md
@@ -1,0 +1,6 @@
+## feature/lua/http client
+
+* Performed automated percent-encoding of query params passed into URI using a
+  using the params option. Parameters encoded using `uri.QUERY_PART` when `GET`,
+  `DELETE` or `HEAD` HTTP methods are used and `uri.FORM_URLENCODED` in other
+  cases (gh-7931).
--- a/src/lua/httpc.lua
+++ b/src/lua/httpc.lua
@@ -311,9 +311,18 @@ local function decode_body(response)
     return res
 end
 
-local function encode_url_params(params)
-    local query_encoder = require("uri")._internal.params
-    local ok, res = pcall(query_encoder, params)
+local function encode_url_params(params, http_method)
+    assert(http_method ~= nil)
+    local uri = require("uri")
+    local uri_escape_opts = uri.FORM_URLENCODED
+    if http_method == "GET" or
+       http_method == "HEAD" or
+       http_method == "DELETE" then
+        uri_escape_opts = uri.QUERY_PART
+    end
+
+    local query_encoder = uri._internal.params
+    local ok, res = pcall(query_encoder, params, uri_escape_opts)
     if not ok then
         error(res)
     end
@@ -410,7 +419,7 @@ curl_mt = {
 
             local encoded_params
             if opts.params then
-                encoded_params = encode_url_params(opts.params)
+                encoded_params = encode_url_params(opts.params, method)
             end
             local url_with_params = url
             if encoded_params then
--- a/src/lua/uri.lua
+++ b/src/lua/uri.lua
@@ -289,41 +289,45 @@ local function unescape(buf, opts)
     return ffi.string(dst, dst_size)
 end
 
-local function encode_kv(key, values, res)
+local function encode_kv(key, values, res, escape_opts)
     local val = values
     if type(val) ~= "table" then
         val = { val }
     end
+    local key_escaped= escape(tostring(key), escape_opts)
 
     -- { a = {} } --> "a"
     if next(val) == nil then
-        table.insert(res, key)
+        table.insert(res, key_escaped)
     end
 
     -- { a = { "b" } } --> "a=c"
     for _, v in pairs(val) do
-        local kv = ("%s=%s"):format(key, v)
+        local val_escaped = escape(tostring(v), escape_opts)
+        local kv = ("%s=%s"):format(key_escaped, val_escaped)
         table.insert(res, kv)
     end
 end
 
--- Encode map to a string.
-local function params(opts)
+-- Encode map to a string and perform escaping of keys and values in
+-- parameters.
+local function params(opts, escape_opts)
     if opts == nil then
         return ""
     end
     if type(opts) ~= "table" then
-        error("Usage: uri.params(table)")
+        error("Usage: uri.params(table[, escape_opts])")
     end
     if next(opts) == nil then
         return ""
     end
     local res = {}
     for key, value in pairs(opts) do
-        if type(key) ~= "string" then
-            error("uri.params: keys must have a type 'string'")
+        if type(key) ~= "string" and
+           type(key) ~= "number" then
+            error("uri.params: keys must have a type 'string' or 'number'")
         end
-        encode_kv(key, value, res)
+        encode_kv(key, value, res, escape_opts)
     end
 
     return table.concat(res, '&')
--- a/test/app-luatest/http_client_test.lua
+++ b/test/app-luatest/http_client_test.lua
@@ -986,3 +986,39 @@ g.test_http_params_request_post = function(cg)
     t.assert_equals(resp.body, "k=2&v=3&v=4")
     t.assert_equals(resp.headers['content_type'], "application/x-www-form-urlencoded")
 end
+
+g.test_http_params_escaped_request_post = function(cg)
+    local http = client.new()
+    t.assert(http ~= nil, "client is created")
+
+    local opts = table.deepcopy(cg.opts)
+    opts.params = {
+        [2] = 4,
+        ["k&"] = "&2",
+        ["v&"] = uri.values("3", "4", "5=", "6 ")
+    }
+
+    -- HTTP method in uppercase.
+    local resp = http:request("POST", cg.url, nil, opts)
+    t.assert_equals(resp.status, 200)
+    local expected_body = "2=4&k%26=%262&v%26=3&v%26=4&v%26=5%3D&v%26=6+"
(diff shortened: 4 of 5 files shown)


ImeevMA (commented):
Hi! Thank you for the patches. I reviewed the last three patches (`uri: reorder functions in module`, `uri: escape params` and `http: escape encoded params`) and I have a few comments. Also, could you please write a proper commit-messages for each of these three patches? And, please drop the patch `Revert "libunwind: use latest release v1.6.2 as a base"`.

src/lua/httpc.lua
```diff
@@ -311,9 +311,18 @@ local function decode_body(response)
...
-local function encode_url_params(params)
-    local query_encoder = require("uri")._internal.params
-    local ok, res = pcall(query_encoder, params)
+local function encode_url_params(params, http_method)
+    print(require("json").encode(params), http_method)
+    local uri = require("uri")
+    local uri_escape_opts = uri.FORM_URLENCODED
+    if http_method == "GET" or
+       http_method == "HEAD" or
+       http_method == "DELETE" then
```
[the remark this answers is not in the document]
ImeevMA: Is it possible to make `http_method` to be `nil`? I suggest adding a test if possible. Otherwise assert() should be enough. Sorry for misleading.

changelogs/unreleased/gh-7931-httpc-params-escaping.md
```diff
@@ -0,0 +1,5 @@
+## feature/lua/http client
+
+* Perform automated percent-encoding of query params passed into URI using a Lua table.
+  Parameters encoded using `uri.QUERY_PART` when `GET`, `DELETE` or `HEAD` HTTP
+  methods are used and `uri.FORM_URLENCODED` in other cases. (gh-7931)
```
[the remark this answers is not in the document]
ImeevMA: I still suggest moving the doc request after `Closes #7931` as I think `Closes #7931` will also appear in the doc issue that will be generated.

ImeevMA (commented):
Hi! Thank you for the patches! I still have a few comment left, please see them.

src/lua/uri.lua
```diff
@@ -258,31 +258,43 @@ local function unescape(buf, opts)
...
 end
 
-local function encode_kv(key, values, res)
+local function encode_kv(key, values, res, escape_opts)
     local val = values
     if type(val) ~= "table" then
         val = { val }
     end
 
+    local key_escaped = escape(key, escape_opts)
```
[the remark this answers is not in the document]
ImeevMA: I still think that using
```
    if escape_opts ~= nil then
        key = escape(key, escape_opts)
    end
```
would be simpler and with fewer `if`'s, but the current version works for me.

ImeevMA (approved):
Thank you or the fixes!

src/lua/uri.lua
```diff
@@ -258,31 +258,45 @@ local function unescape(buf, opts)
...
+        key_escaped = escape(key, escape_opts)
+    end
     -- { a = {} } --> "a"
     if next(val) == nil then
-        table.insert(res, key)
+        if escape_opts ~= nil then
+            table.insert(res, key_escaped)
+        else
+            table.insert(res, key)
+        end
```
Totktonada, line 277: If we'll add a capture

```lua
-- Unlike escape() it doesn't perform escaping by default.
--
-- It is needed to <...>.
local function _escape(str)
    if escape_opts == nil then
        return str
    end
    return escape(str, escape_opts)
end
```

those if-else blocks will gone. They will be just:

```lua
table.insert(res, _escape(key))
<...>
local kv = ("%s=%s"):format(_escape(key), _escape(v))
```

It makes the code more straightforward.

However I don't see a reason why don't call just `escape()`. The functions of the question are internal, we don't need to keep any compatibility here.

test/app-luatest/http_client_test.lua
```diff
@@ -986,3 +986,36 @@ g.test_http_params_request_post = function(cg)
     t.assert_equals(resp.body, "k=2&v=3&v=4")
     t.assert_equals(resp.headers['content_type'], "application/x-www-form-urlencoded")
 end
+
```
Totktonada, line 989: Regarding tests:

1. The motivating example is `&` or `=` in a key or a value. It would be nice to have such a case.
2. It would be nice to see a difference between `GET` and `POST`: have a space in the test case.
3. We have no test case that would show that a key name is encoded.

changelogs/unreleased/gh-7931-httpc-params-escaping.md
```diff
@@ -0,0 +1,6 @@
+## feature/lua/http client
+
+* Performed automated percent-encoding of query params passed into URI using a
+  Lua table. Parameters encoded using `uri.QUERY_PART` when `GET`, `DELETE` or
+  `HEAD` HTTP methods are used and `uri.FORM_URLENCODED` in other cases.
+  (gh-7931)
```
Totktonada, line 6: Nits on the changelog entry.

1. 'using a Lua table' -> 'using the `params` option'.
2. 'other cases. (gh-xxxx)' -> 'other cases (gh-xxxx).'

src/lua/uri.lua
```diff
@@ -272,39 +305,6 @@ local function values(...)
     return {...}
 end
 
--- Encodes a string into its escaped hexadecimal representation.
```
Totktonada: PR #7254 doesn't land ATM, so we can place the functions at needed place from scratch. Or use a forward declaration.

```lua
local escape

local function foo()
   <..use escape..>
end

escape = function(<...>)
    <...>
end
```

Feel free to ignore, but in this case I beg you to give extra care at future rebase, because it may leave old version of the code after the movement. The conflict detection will not complain against the added code, only about removed.

changelogs/unreleased/gh-7931-httpc-params-escaping.md
Totktonada: > > 'using a Lua table' -> 'using the params option'.
>
> What's wrong here? We pass a number of parameters as a Lua table.

That's correct, but a bit vague. I highlighted it, because I stuck here myself for some time. Specifying certain parameter name should make things clear.

test/app-luatest/http_client_test.lua
Totktonada: > > 3. We have no test case that would show that a key name is encoded.
>
> Key contains a `&` now that is escaped.

Keys are `k` and `v` in this test and they don't contain `&`.

src/lua/uri.lua
```diff
@@ -292,31 +292,34 @@ local function unescape(buf, opts)
     return ffi.string(dst, dst_size)
 end
 
-local function encode_kv(key, values, res)
+local function encode_kv(key, values, res, escape_opts)
     local val = values
     if type(val) ~= "table" then
         val = { val }
     end
+    local key_escaped= escape(key, escape_opts)
```
Totktonada, line 300: That's interesting: you call `tostring()` on values, but not on keys. I think that we should follow the same rule here. I guess that the idea behind this was to accept numbers in values. Moreover, it is more important for keys, because `{'a', 'b', 'c'}` has numeric keys.

- [ ] Let's accept them in values and keys both (call `tostring()` for both).
- [ ] Let's add test cases with numeric keys and values.

changelogs/unreleased/gh-7931-httpc-params-escaping.md
```diff
@@ -0,0 +1,6 @@
+## feature/lua/http client
```
Totktonada, line 1: Doc request:

> with `GET`, `HEAD` and `DELETE` parameters `uri.QUERY_STRING`
> are used and with other HTTP method `uri.FORM_URLENCODED` is used.

Typo: `uri.QUERY_STRING` -> `uri.QUERY_PART`.

src/lua/uri.lua
Totktonada: Ouch, I didn't know that everything except strings is forbidden for keys. However allowing at least numbers seems correct for me.

Totktonada (approved):

PR #8047, opened 2022-12-12
httpc: introduce stream input/output interface

### LINKED ISSUE #7845: Stream http.client support

We were unable to communicate [with a third-party API using http.client](https://github.com/tarantool/etcd-client/issues/2#issuecomment-1288860236). It requires not a regular POST-request workflow (open a connection, send a request, receive a response, close the connection), but a kind of a stream HTTP-request (open a connection, send/receive data many times, close the connection).

The cURL library has options [CURLOPT_WRITEFUNCTION](https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html) and [CURLOPT_READFUNCTION](https://curl.se/libcurl/c/CURLOPT_READFUNCTION.html). With that callbacks you can control both output and input data streams. If there was a way to set up these callbacks. It would help to solve the problem.

I would suggest to extend the [http.client](https://www.tarantool.io/ru/doc/latest/reference/reference_lua/http/) API. We could add a `stream` field for a request options:

```Lua
local response = client:post(url, first_data_chunk or nil, {stream = true}
```

In this case, the response will not have a `body` field, but two methods for `read` and `write` streamed data:
```Lua
response:write(data_chunk)
local data_chunk = response:read()
```
and the `close` method to finish the request:
```Lua
response:close()
```

Also it may have the HTTP-header "Transfer-Encoding: chunked" instread of `Content-Length: ХХ`.

In all other things it could be a response object with the same fields and settings.

We could act in a cURL way and add posibility to a user specify write/read function callbacks:

```Lua
local response = client:post(url, first_data_chunk or nil, {
     read_callback = read_callback_function,
     write_callback = write_callback_function,
})
```

**Additional context**

[gRPC -> REST implementation](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/handler.go#L18-L99) Used headers:
```
Transfer-Encoding: chunked
Content-Type: application/json
```
[See also.](https://hackmd.io/PjlwhbzuQR6x9nCM0MIHUQ?view#Long-response-chunked)

**Definition of Done**

A user can send and receive data via a streamed HTTP-request.

oleg-jukovec:
Add a streaming data input/output object for http.client. The input/output object can be created using the same methods and the same options as a normal request, but with a new option {chunked = true}.

An uncompleted io object has only headers and cookies fields. A completed io object has the same fields as a normal request, but without the `body` field.

The io object interface looks like the socket object interface and should have the same description:

```
io_object:read(chunk[, timeout])
io_object:read(delimiter[, timeout])
io_object:read({chunk = chunk, delimiter = delimiter}[, timeout])
io_object:write(data[, timeout])
```

The difference is in the method `finish`. Unlike socket:close() it has an optional parameter `timeout`:

```
io_object:finish([timeout])
```

Be careful, the call may yield a fiber. The idea is to wait until a HTTP connection is finished by the server-side or force finish the connection from client-time after a timeout value.

The default timeout value is 10 seconds for all methods.

Usage example:

```lua
local io = httpc:post(url, nil, {chunked = true})

local write_chan = fiber.channel()
fiber.new(function()
    fiber.name("write to " .. url)
    while true do
        local data = write_chan:get()
        if data == nil then
            break
        end
        io:write(data, 1)
    end
end)

local recvdata
while recvdata = io:read('\r\n', 1) do
    local decoded = json.decode(recvdata)
    if condition(decoded) then
        write_chan:put(data)
    end
    if condition(decoded) then
        io:finish(1)
    end
end
write_chan:close()
```

See also:
* https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#lua-function.socket_object.read
* https://github.com/tarantool/tarantool/issues/7845#issuecomment-1298538412
* https://github.com/tarantool/tarantool/issues/7845#issuecomment-1298821779

Closes #7845

### DIFF
--- a/changelogs/unreleased/gh-7845-httpc-stream-io-interface.md
+++ b/changelogs/unreleased/gh-7845-httpc-stream-io-interface.md
@@ -1,0 +1,3 @@
+## feature/lua/http client
+
+* Introduce stream input/output interface for http.client (gh-7845).
--- a/src/curl.c
+++ b/src/curl.c
@@ -91,6 +91,8 @@ curl_multi_process(CURLM *multi, curl_socket_t sockfd, int events)
 			errinj->bparam = false;
 #endif
 		fiber_cond_signal(&request->cond);
+		if (request->done_handler)
+			request->done_handler(request->done_handler_arg);
 	}
 }
 
@@ -277,6 +279,8 @@ curl_request_create(struct curl_request *curl_request)
 	curl_request->in_progress = false;
 	curl_request->code = CURLE_OK;
 	fiber_cond_create(&curl_request->cond);
+	curl_request->done_handler = NULL;
+	curl_request->done_handler_arg = NULL;
 	return 0;
 }
 
@@ -288,15 +292,42 @@ curl_request_destroy(struct curl_request *curl_request)
 	fiber_cond_destroy(&curl_request->cond);
 }
 
+/**
+ * Set diag if mcode != CURLM_OK.
+ */
+static inline void
+curl_diag_set_merror(CURLMcode mcode)
+{
+	if (mcode != CURLM_OK) {
+		switch (mcode) {
+		case CURLM_OUT_OF_MEMORY:
+			diag_set(OutOfMemory, 0, "curl", "internal");
+			break;
+		default:
+			errno = EINVAL;
+			diag_set(SystemError, "curl_multi_error: %s",
+				 curl_multi_strerror(mcode));
+		}
+	}
+}
+
 CURLMcode
-curl_execute(struct curl_request *curl_request, struct curl_env *env,
-	     double timeout)
+curl_request_start(struct curl_request *curl_request, struct curl_env *env)
 {
 	CURLMcode mcode;
 	curl_request->in_progress = true;
 	mcode = curl_multi_add_handle(env->multi, curl_request->easy);
-	if (mcode != CURLM_OK)
-		goto curl_merror;
+	curl_diag_set_merror(mcode);
+
+	return mcode;
+}
+
+CURLMcode
+curl_request_finish(struct curl_request *curl_request, struct curl_env *env,
+		    double timeout)
+{
+	CURLMcode mcode;
+
 	ERROR_INJECT_YIELD(ERRINJ_HTTP_RESPONSE_ADD_WAIT);
 	/* Don't wait on a cond if request has already failed or finished. */
 	if (curl_request->code == CURLE_OK && curl_request->in_progress) {
@@ -307,20 +338,8 @@ curl_execute(struct curl_request *curl_request, struct curl_env *env,
 		--env->stat.active_requests;
 	}
 	mcode = curl_multi_remove_handle(env->multi, curl_request->easy);
-	if (mcode != CURLM_OK)
-		goto curl_merror;
+	curl_diag_set_merror(mcode);
+	curl_request->in_progress = false;
 
-	return CURLM_OK;
-
-curl_merror:
-	switch (mcode) {
-	case CURLM_OUT_OF_MEMORY:
-		diag_set(OutOfMemory, 0, "curl", "internal");
-		break;
-	default:
-		errno = EINVAL;
-		diag_set(SystemError, "curl_multi_error: %s",
-			 curl_multi_strerror(mcode));
-	}
 	return mcode;
 }
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -45,7 +45,34 @@ static_assert(MAX_HEADER_LEN < SMALL_STATIC_SIZE,
 /** The HTTP headers that may be set automatically. */
 #define HTTP_ACCEPT_HEADER	"Accept:"
 #define HTTP_CONNECTION_HEADER	"Connection:"
+#define HTTP_CONTENT_LENGTH_HEADER "Content-Length:"
 #define HTTP_KEEP_ALIVE_HEADER	"Keep-Alive:"
+
+/**
+ * libcurl callback for CURLOPT_READFUNCTION
+ * @see https://curl.haxx.se/libcurl/c/CURLOPT_READFUNCTION.html
+ */
+static size_t
+curl_easy_io_read_cb(char *buffer, size_t size, size_t nitems, void *ctx)
+{
+	struct httpc_request *req = (struct httpc_request *)ctx;
+	const size_t buffer_size = size * nitems;
+
+	size_t ibuf_len = ibuf_used(&req->send);
+	if (ibuf_len == 0) {
+		fiber_cond_broadcast(&req->io_send_cond);
+		if (req->io_send_closed)
+			return 0;
+		return CURL_READFUNC_PAUSE;
+	}
+
+	size_t read_len = ibuf_len < buffer_size ? ibuf_len : buffer_size;
+	memcpy(buffer, req->send.rpos, read_len);
+	req->send.rpos += read_len;
+
+	fiber_cond_broadcast(&req->io_send_cond);
+	return read_len;
+}
 
 /**
  * libcurl callback for CURLOPT_WRITEFUNCTION
@@ -57,12 +84,17 @@ curl_easy_write_cb(char *ptr, size_t size, size_t nmemb, void *ctx)
 	struct httpc_request *req = (struct httpc_request *) ctx;
 	const size_t bytes = size * nmemb;
 
-	char *p = region_alloc(&req->resp_body, bytes);
+	char *p = region_alloc(&req->recv, bytes);
 	if (p == NULL) {
 		diag_set(OutOfMemory, bytes, "ibuf", "httpc body");
 		return 0;
 	}
+
 	memcpy(p, ptr, bytes);
+
+	if (req->io)
+		fiber_cond_signal(&req->io_recv_cond);
+
 	return bytes;
 }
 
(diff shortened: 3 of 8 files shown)


R-omk:
from the commit text:

`chan:readable()`

fiber channel has no method  'readable'

changelogs/unreleased/gh-7845-httpc-stream-io-interface.md
```diff
@@ -0,0 +1,3 @@
+## feature/httpc
```
Totktonada, line 1: Nit: Usually `## feature/lua/http client`.

src/lua/httpc.lua
```diff
@@ -320,6 +326,122 @@ local function encode_url_params(params)
...
+
+local function io_write(self, data, timeout)
+    timeout = timeout or IO_TIMEOUT_DEFAULT
+    return self._internal.io:write(data, #data, timeout)
+end
+
+local function io_close(self, timeout)
+    timeout = timeout or IO_TIMEOUT_DEFAULT
+    self["status"], self["reason"] = self._internal.io:close(timeout)
+end
```
Totktonada, line 443: From a user perspective it looks unusual for me to see a timeout in a close method. I briefly looked around the code and, it seems, it waits for all in-fly write requests from other fibers. It is correct? I would like to see some clarification in the documentation (doc request?) about this timeout: what exactly it waits? I would also want to know, what is a state of the `io` if `:close()` fail on a timeout. Is it closed or open?

A side question: whether `:close()` is idempotent? IOW, whether `:close()` on a closed object is no-op and successful?

I think that it is good property for :close()` to never fail, because it is unclear what to do with a failed close. (`:close()` may report an error for diagnostic purposes, but still release resources.)

oleg-jukovec: > From a user perspective it looks unusual for me to see a timeout in a close method. I briefly looked around the code and, it seems, it waits for all in-fly write requests from other fibers. It is correct?

In this case, we are waiting for a server to close the connection after we have already done:
https://curl.se/libcurl/c/CURLOPT_READFUNCTION.html

If you stop the current transfer by returning 0 "pre-maturely" (i.e before the server expected it, **like when you have said you will upload N bytes and you upload less than N bytes), you may experience that the server "hangs" waiting for the rest of the data that will not come.** 

A similar case may be with `Transfer-Encoding: chunked` (we think that we already done, but a server don't).

Those. actually it can be used like `it waits for all in-fly write requests from other fibers`, but in fact the main idea to force close the connection from the client-side.

changelogs/unreleased/gh-7845-httpc-stream-io-interface.md
oleg-jukovec: Thank you, updated.

src/lua/httpc.lua
oleg-jukovec: > A side question: whether :close() is idempotent? IOW, whether :close() on a closed object is no-op and successful?

Yes, since the connection is closed, we have nothing to wait for. So we just finish the call immediately.

> I think that it is good property for :close() to never fail, because it is unclear what to do with a failed close. (:close()` may report an error for diagnostic purposes, but still release resources.)

Now (after the update) the error is returned only if timeout < 0. It looks clear, but I can do to remove the error:
```C
timeout = timeout >= 0 ? timeout : 0;
```
But it seems to me better with an error in this case.

Totktonada: >> I think that it is good property for :close() to never fail <...>
>
> Now (I make a little update) the error is returned only if timeout < 0. <...>

Of course, if incorrect arguments are passed, the method should report an error. I meant all other situations except incorrect arguments. I think that if we're going to make `:close()` always-successful in given sense we should also handle fiber cancelling.

BTW, `:close()` should fail all subsequent `:write()` calls. Oleg said that there is a window after `:close()`, when a next `:write()` may succeed. I would look on it and estimate the time frame (one event loop iterator? a network trip?). I would also take care to effort that is needed to resolve it. Can't say now, whether it worth it.

Regarding waiting in `:close()` for the server to close its side: got it, need some time to think around.

LeonidVas (approved):
Hi! Thank you for the patch.

src/httpc.c
```diff
@@ -390,8 +435,192 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+
+	if (len > 0) {
+		ibuf_reset(&req->send);
+		char *buf = ibuf_alloc(&req->send, len);
+		memcpy(buf, ptr, len);
+	} else {
+		req->io_send_closed = true;
+	}
+
+	curl_easy_pause(req->curl_request.easy, CURLPAUSE_SEND_CONT);
```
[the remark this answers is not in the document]
oleg-jukovec: https://curl.haxx.se/libcurl/c/CURLOPT_READFUNCTION.html
https://curl.se/libcurl/c/curl_easy_pause.html

`Your function must return the actual number of bytes that it stored in the data area pointed at by the pointer buffer. Returning 0 will signal end-of-file to the library and cause it to stop the current transfer. `

`The callback can return CURL_READFUNC_PAUSE to cause reading from this connection to pause. See [curl_easy_pause](https://curl.se/libcurl/c/curl_easy_pause.html) for further details.`

`Using this function ( curl_easy_pause ), you can explicitly mark a running connection to get paused, and you can unpause a connection that was previously paused.`

I return `CURL_READFUNC_PAUSE` if there are no data to send yet from `curl_easy_io_read_cb`.  So we need resume sending data after receiving the next batch of data with `CURLPAUSE_SEND_CONT`.

src/lua/httpc.lua
```diff
@@ -329,6 +335,122 @@ local function encode_url_params(params, http_method)
...
+        return 0
+    end
+    local rbuf = self._internal.rbuf
+    if rbuf:size() == 0 then
+        return nil
+    end
+
+    local shortest
+    for _, eol in ipairs(eols) do
+        local data = ffi.C.memmem(rbuf.rpos, rbuf:size(), eol, #eol)
```
0x501D, line 356: memmem(3) is not specified in POSIX.1. Is it OK for us?

src/lua/httpc.lua
```diff
@@ -459,13 +584,22 @@ curl_mt = {
...
+            if not opts.chunked then
+                resp.url = url_with_params
+                return setmetatable(resp, {
+                    __index = {
+                        decode = decode_body,
+                    },
+                })
+            else
+                return setmetatable(resp, {
+                    __index = {
```
LeonidVas, line 596: It might be nice to add some flag like `is_stream` to distinguish a different type of  `resp` in client code.

src/lua/httpc.lua
oleg-jukovec: The code is copy-pasted from socket.lua. So I think it is OK here.

src/httpc.c
```diff
@@ -390,8 +435,192 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+	fiber_cond_broadcast(&req->io_send_cond);
+	req->io_send_closed = true;
+}
+
 int
-httpc_execute(struct httpc_request *req, double timeout)
+httpc_set_io(struct httpc_request *req, const char *method)
+{
+	if (req->io)
+		return -1;
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, added.

src/curl.c
```diff
@@ -307,20 +338,19 @@ curl_execute(struct curl_request *curl_request, struct curl_env *env,
...
-			 curl_multi_strerror(mcode));
-	}
+	curl_diag_set_merror(mcode);
+	curl_request->in_progress = false;
+
 	return mcode;
 }
+
+CURLMcode
+curl_execute(struct curl_request *curl_request, struct curl_env *env,
```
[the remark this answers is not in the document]
oleg-jukovec: Removed.

src/httpc.h
```diff
@@ -404,6 +426,86 @@ httpc_set_follow_location(struct httpc_request *req, long follow);
...
+ * @see https://en.wikipedia.org/wiki/Chunked_transfer_encoding
+ * @see https://curl.haxx.se/libcurl/c/CURLOPT_UPLOAD.html
+ */
+int
+httpc_set_io(struct httpc_request *req, const char *method);
+
+/**
+ * Read a next chunk of data.
+ *
+ * @param request Reference to request object with enabled io.
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed.

0x501D (approved):

oleg-jukovec:
Issues discussed with @Totktonada @LeonidVas :

1. Rename request:close() -> request:finish().
2. A unit-test with finish(timeout) expired.
3. request:write() must return an error if a request not PUT/POST/PATCH.
4. Mark that request:finish() yields() with docbot.

oleg-jukovec:
> New issues discussed with @Totktonada @LeonidVas :
> 
>     1. Rename request:close() -> request:finish().
> 
>     2. A unit-test with request:finish(timeout) expired.
> 
>     3. request:write() must return an error if a request method not PUT/POST/PATCH.
> 
>     4. Add a note that request:finish() may yield to the commit message (document part).

Thank you, fixed.

src/lua/httpc.lua
Totktonada: We agreed to rename `:close()` to `:finish()` due to its semantic with waiting for an answer. See below.

Totktonada (approved):
Thank you! Impressive work!

While there are areas to polish (illegal parameters reporting, io object serialization, buffer allocation/data copying) it is OK for inclusion into upcoming RC.

PR #8048, opened 2022-12-12
lua: use cur_L in cord_on_yield

fckxorg:
Before the patch, fiber->storage.lua.stack is used for `panic` calls. However, some fibers don't have any Lua state saved in their storage (for example, space
triggers).

After the patch, the Lua state pointed by `cur_L` is used to make those calls, as it is always present.

As proposed by @Totktonada in https://github.com/tarantool/tarantool/pull/7976, I've checked a few cases, that could lead to the same behavior.
- fiber_new() from C -- Can't occur here, either no lua_State at all, or it is invoked from a lua fiber.
- IPROTO_CALL for a C function -- Again, no lua_State here at all.
- IPROTO_EXECUTE for a query that calls a Lua function -- Handled well for 2.x.x, not present for 1.10
- IPROTO_INSERT for a space with a trigger -- fails, added corresponding test

### DIFF
--- a/changelogs/unreleased/gh-6647-yield-gc-hook.md
+++ b/changelogs/unreleased/gh-6647-yield-gc-hook.md
@@ -1,0 +1,3 @@
+## bugfix/fiber
+
+* Fixed the assertion fail in `cord_on_yield` (gh-6647).
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -882,7 +882,7 @@ void cord_on_yield(void)
 		 * world and one can obtain the corresponding Lua
 		 * coroutine from the fiber storage.
 		 */
-		struct lua_State *L = fiber()->storage.lua.stack;
+		struct lua_State *L = gco2th(gcref(g->cur_L));
 		assert(L != NULL);
 		snprintf(buf, sizeof(buf),
 			 "fiber %llu is switched while running the"
@@ -915,7 +915,7 @@ void cord_on_yield(void)
 	 */
 	if (unlikely(g->hookmask & HOOK_GC)) {
 		char buf[128];
-		struct lua_State *L = fiber()->storage.lua.stack;
+		struct lua_State *L = gco2th(gcref(g->cur_L));
 		assert(L != NULL);
 		snprintf(buf, sizeof(buf),
 			 "fiber %llu is switched while running GC"
--- a/test/box-luatest/gh_6647_yield_gc_hook_test.lua
+++ b/test/box-luatest/gh_6647_yield_gc_hook_test.lua
@@ -1,0 +1,80 @@
+local t = require('luatest')
+local g = t.group('gh-6647-yield-gc-hook')
+local fio = require('fio')
+
+local tarantool_bin = arg[-1]
+local PANIC = 256
+local ERRMSG = 'fiber [0-9]+ is switched while running GC finalizer'
+
+local function check_log(t, filename, pattern)
+    local file = fio.open(filename, {'O_RDONLY'})
+    assert(file ~= nil)
+
+    local size = file:stat().size
+    assert(size ~= 0)
+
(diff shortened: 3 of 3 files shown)


test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -0,0 +1,48 @@
...
+    script:write("\nos.exit(0)")
+    script:close()
+    local cmd = [[/bin/sh -c 'cd "%s" && "%s" ./script.lua 2> /dev/null']]
+    local res = os.execute(string.format(cmd, dir, tarantool_bin))
+    fio.rmtree(dir)
+    return res
+end
+
+local space_trigger = [[
+box.cfg({listen = 3301})
```
Totktonada, line 26: Our testing is parallel and there is no guarantee (or even a good probability) that any particular TCP port is free. There are two well-known ways to overcome it:

* Use a Unix domain socket.
* Bind on zero TCP port (look https://github.com/tarantool/tt/pull/112 for details).

I suggest the former, because it is simpler. It has its own limitation: 103 (Mac OS) or 107 (Linux) symbols in the socket path. However it unlikely will hit us here.

test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -0,0 +1,48 @@
...
+
+local function run_script(code)
+    local dir = fio.tempdir()
+    local script_path = fio.pathjoin(dir, 'script.lua')
+    local script = fio.open(script_path, {'O_CREAT', 'O_WRONLY', 'O_APPEND'},
+        tonumber('0777', 8))
+    script:write(code)
+    script:write("\nos.exit(0)")
+    script:close()
+    local cmd = [[/bin/sh -c 'cd "%s" && "%s" ./script.lua 2> /dev/null']]
```
Totktonada, line 19: How about writing stderr to a file and check that it contains `'fiber [0-9]\+ is switched while running GC finalizer'`? Or, maybe better, set `box.cfg({log = 'script.log'})` and find the pattern in the file.

Totktonada (approved):
I'm not a LuaJIT professional, so I can't say much there. IOW, if `cur_L` works here, it is probably fine. I would lean on @igormunkin here as on the main reviewer.

test/box-tap/gh-6647-yield-gc-hook.test.lua
```diff
@@ -0,0 +1,48 @@
...
+test:plan(1)
+
+local tarantool_bin = arg[-1]
+local PANIC = 256
+
+local function run_script(code)
+    local dir = fio.tempdir()
+    local script_path = fio.pathjoin(dir, 'script.lua')
+    local script = fio.open(script_path, {'O_CREAT', 'O_WRONLY', 'O_APPEND'},
+        tonumber('0777', 8))
```
Totktonada, line 15: The file unlikely exists, but `O_TRUNC` here looks more appropriate.

igormunkin (approved):
@fckxorg, thanks for the patch! See no flaws, so LGTM, but I agree with almost all @Totktonada comments regarding the test.

Furthermore, I'd rather close #7976, or convert it to draft at least, since originally it hardly relates to the original problems described in #6647. Now, when the original issue is resolved, all refactoring is out of scope of this particular issue. I will simply backport this PR to all other long term branches. All the work done within #7976 should be done either in scope of  the more related issue or as a refactoring.

igormunkin:
For the history: @fckxorg has tried to rewrite both test in a more luatest way via `server:exec` (see d479240), but he's failed since luatest doesn't allow to crash a server instance on purpose. Test are finally reverted back to the last valid revision using luatest, but running the code via `os.execute`.

igormunkin:
Also backported to 2.10 in scope of 9f4d315 and to 1.10 in scope of 64c72d0.

PR #8049, opened 2022-12-13
sql: fix possible memleak in lemon.c

ImeevMA:
This patch fixes possible memleak.

NO_DOC=Possible bug
NO_TEST=No proper way to test
NO_CHANGELOG=Possible bug

### DIFF
--- a/extra/lemon.c
+++ b/extra/lemon.c
@@ -2785,6 +2785,7 @@ void Parse(struct lemon *gp)
   filebuf = (char *)malloc( filesize+1 );
   if( filesize>100000000 || filebuf==0 ){
     ErrorMsg(ps.filename,0,"Input file too large.");
+    free(filebuf);
     gp->errorcnt++;
     fclose(fp);
     return;


extra/lemon.c
```diff
@@ -2785,6 +2785,7 @@ void Parse(struct lemon *gp)
   filebuf = (char *)malloc( filesize+1 );
   if( filesize>100000000 || filebuf==0 ){
     ErrorMsg(ps.filename,0,"Input file too large.");
+    free(filebuf);
```
Gerold103, line 2788: An alternative option was to 1) use `xmalloc()`, 2) check file size before the allocation. But I am fine with this solution too.

Gerold103 (approved):
Thanks for the patch!

extra/lemon.c
ImeevMA: Thanks for the review! I think `xmalloc()` needed a new header and we needed `free()` anyway in case `malloc()` failed, so I decided to use `free()` in both cases.

locker (approved):

Issue #3534, opened 2018-07-14
Panic on malloc/mmap failure


locker:
There are two kinds of memory allocations in tarantool: those that are used for storing and indexing tuples and those that are used for storing internal objects. The primary difference between them is that the max amount of memory that can be used up by the former is limited by various configuration options (such as `box.cfg.memtx_memory`) while allocations used for internal objects are not limited at all and can only fail due to malloc/mmap failure.

Obviously, it totally makes sense to handle allocation failures caused by configuration parameters, because they may happen due to misconfiguration, in which case the admin can recover the system by tweaking the corresponding parameter. However, handling malloc/mmap failures doesn't make much sense, because modern operating systems have quite lenient memory overcommit policies (in response to modern application demands). For instance, if you have a 4 GB Linux box, you can't allocate 5 GB with mmap or malloc (it will return NULL/ENOMEM), but you can allocate 1 GB 5 times (10, 100, 1000 times) as long as you don't actually use the allocated virtual memory. If an application tries to use more memory than there's available on the host, the OS will do swapping and if that doesn't help, it will kill the application. That's it. No graceful termination. No malloc/mmap failure. Just SIGKILL.

Nevertheless, we diligently strive to bail out gracefully in case malloc returns NULL even though all the code written for handling those failures is never executed (see the previous paragraph) hence is untested and likely to have numerous bugs lurking in it. What's worse it often complicates the code workflow and forces us to write even more code. The proposal is to dump that dead weight and simply panic on any mmap/malloc failure.

Totktonada:
I think we should either find a way to test this code or (more likely) remove it. I agree.

PR #8057, opened 2022-12-13
vinyl: fix potential null-ptr dereference in `vy_read_view_merge`

locker:
The `copy` ptr is checked for `NULL` after dereferencing. Found by PVS.

https://jira.vk.team/browse/TNT-698

It can hardly result in a null-ptr dereference, because `vy_stmt_dup()` uses `malloc()`, which doesn't normally fail (instead, the OOM killer just kills the process if there's a severe shortage of memory, see #3534). Still, PVS rightfully complains about it so let's fix it.

Fixes commit 902d212b6e544.


Gumix (approved):

locker:
Cherry-picked to 2.10.

PR #8058, opened 2022-12-13
ci: update submodule to HEAD of remote branch

NickVolynkin:
Workflow submodule_update used the HEAD of local branch on the runner as a git reference for updating the submodule. If the repository was already checked out on the runner before, it had the same old `master` branch from the first checkout.
Now the workflow will force action tarantool/actions/update-submodule to use the HEAD of remote branch. This remote branch must be up-to-date.


NickVolynkin:
#8060 solves the same problem and more, so I'm closing this PR.

PR #8059, opened 2022-12-13
log: call log_write_flightrec regardless of the log level

Gumix:
Before commit 243234483ae8 ("log: add log.new() function that creates a new logger"), `log_write_flightrec()` was called from `log_vsay()` regardless of the log level. After, the log level is checked in Lua, so `log_vsay()` may not be called.
This patch restores the original behaviour by moving `log_write_flightrec()` calls to `say_default()` and `say()`.

Part of https://github.com/tarantool/tarantool-ee/issues/320

### DIFF
--- a/src/lib/core/say.c
+++ b/src/lib/core/say.c
@@ -58,9 +58,9 @@ int log_level = S_INFO;
 /**
  * Log level of flight recorder. Log is passed to flight recorder only if
  * it's log level not less that this. -1 is used if we should not pass logs
- * to flight recorder (it is currently disabled).
+ * to flight recorder (it is currently disabled). Used in Lua via FFI.
  */
-static int log_level_flightrec = -1;
+int log_level_flightrec = -1;
 /**
  * This function is called for every log which log level is not less than
  * log_level_flightrec.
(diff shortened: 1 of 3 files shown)


src/lib/core/say.c
```diff
@@ -60,7 +60,7 @@ int log_level = S_INFO;
  * it's log level not less that this. -1 is used if we should not pass logs
  * to flight recorder (it is currently disabled).
  */
-static int log_level_flightrec = -1;
+int log_level_flightrec = -1;
```
locker, line 63: Nit: Please add a comment saying that it isn't static, because it's used in Lua via FFI.

src/lib/core/say.c
```diff
@@ -1051,15 +1054,13 @@ say_internal(int level, bool check_level, const char *module,
...
 {
-	int errsv = errno;
 	int total = log_vsay(log_default, level, check_level, module, filename,
 			     line, error, format, ap);
 	if (total > 0 &&
 	    level == S_FATAL && log_default->fd != STDERR_FILENO) {
 		ssize_t r = safe_write(STDERR_FILENO, say_buf, total);
 		(void)r;
 	}
-	errno = errsv; /* Preserve the errno. */
```
locker, line 1062: Strictly speaking, `log->format_func`, which is called by `log_vsay`, may set `errno` (in `snprintf`) so I wouldn't move it to `safe_write`.

Gumix: `log_vsay` itself saves/restores `errno`, am I missing something?

src/lib/core/say.c
Gumix: Done.

src/lib/core/say.c
locker: Ah, okay, missed that.

locker: Still, decided to not move `errno` saving code from `say_internal` to `safe_write`, because `safe_write` is called from `log_vsay` so this change doesn't make the code any better.

locker (approved):

PR #8061, opened 2022-12-14
Eliminate dependency of core lib headers on `libunwind.h`

### LINKED ISSUE #8025: Tarantool doesn't build if libundwind.h is unavailable

locker:
The only reason why `libunwind.h` is included into `backtrace.h` (and then transitively into `fiber.h`) is use of `unw_word_t` type. Let's replace it with `uintptr_t` and use `unw_word_t` only for interacting with the unwind library.

This commit partially reverts commit 6d088b56ac34, because we don't need to include libunwind.h into tests anymore.

Fixes #8025

### DIFF
--- a/src/lib/core/backtrace.c
+++ b/src/lib/core/backtrace.c
@@ -15,6 +15,8 @@
 #include <dlfcn.h>
 #endif /* __APPLE__ */
 
+#include "libunwind.h"
+
 #include "cxx_abi.h"
 #include "proc_name_cache.h"
 
@@ -239,18 +241,19 @@ skip_frames:
 
 const char *
 backtrace_frame_resolve(const struct backtrace_frame *frame,
-			unw_word_t *offset)
+			uintptr_t *offset)
 {
-	const char *name = proc_name_cache_find((unw_word_t)frame->ip, offset);
+	const char *name = proc_name_cache_find(frame->ip, offset);
 	if (name != NULL)
 		return name;
 #ifndef __APPLE__
 	unw_accessors_t *acc = unw_get_accessors(unw_local_addr_space);
 	assert(acc->get_proc_name != NULL);
 	char proc_name_buf[128];
+	unw_word_t unw_offset;
 	int rc = acc->get_proc_name(unw_local_addr_space, (unw_word_t)frame->ip,
 				    proc_name_buf, sizeof(proc_name_buf),
-				    offset, NULL);
+				    &unw_offset, NULL);
 	/*
 	 * In case result does not fit into buffer we got UNW_ENOMEM and
 	 * result will be truncated.
@@ -260,6 +263,7 @@ backtrace_frame_resolve(const struct backtrace_frame *frame,
 			  "%s", unw_strerror(rc));
 		return NULL;
 	}
+	*offset = (uintptr_t)unw_offset;
 #else /* __APPLE__ */
 	Dl_info dli;
 	if (dladdr(frame->ip, &dli) == 0) {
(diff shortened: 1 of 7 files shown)


CuriousGeorgiy (commented):
Could you please verify that a compiler error will be issued if `uintptr_t` and `unw_word_t` are not the same?

Also, I would still want to see the tests unnecessarily including `src/core/fiber.h` in EE fixed (if there are).

CuriousGeorgiy (changes requested):
AFAIC, you can use `unw_word_t` instead of `uintptr_t` in `src/lib/core/backtrace.c` — that way we can statically guarantee that `uintptr_t` and `unw_word_t` are the same.

Also, I would still want you to investigate the tests unnecessarily including `src/core/fiber.h` in EE and fix them, if there are.

locker:
> Could you please verify that a compiler error will be issued if `uintptr_t` and `unw_word_t` are not the same?

Why do they need to be the same? We always pass `unw_word_t` to libunwind functions.

> Also, I would still want to see the tests unnecessarily including `src/core/fiber.h` in EE fixed (if there are).

All the inclusions of `fiber.h` into EE unit tests are required.

locker:
> > Why do they need to be the same? We always pass unw_word_t to libunwind functions.
> 
> But now you are casting `unw_word_t` to `uintptr_t` in your patch.

What's wrong with it? The variable of type `unw_word_t` that I cast to `uintptr_t` stores an offset inside a function that will surely fit in `uintptr_t`. BTW the current code actually casts `unw_word_t` to `uint64_t` implicitly in `proc_name_cache.cc` and doesn't care.

> Besides, IMHO, the patch will be simpler if you apply my suggestion (i.e., you won't need to patch types in `src/lib/core/backtrace.c` and `src/lib/core/proc_name_cache.cc`).

My primary goal isn't making the *patch* simpler, but eliminating the `libunwind.h` dependency and making the code easier to read. If we use `uintptr_t` in headers, we should use it in the implementation as well. The whole purpose of `unw_word_t` is to provide a platform-independent API to libunwind. I don't see why we should stick to it.

Totktonada (approved):
Nice idea!

alyapunov:
Thanks for summon! Interesting discussion.

I think that the main problem is that libunwind should use `uintptr_t` in its API but by some reason it doesn't. Perhaps we could statically check that `unw_word_t` is exactly the same as `uintptr_t` and simply use the latter but generally it's not a portable approach and/or a bit not polite to libunwind authors.

I mean that the library convention states that there's a type `unw_word_t`, that seems to be an unsigned integral type of platform-depended size. By this convention we should not expect this type to be exactly `uintptr_t` or `unsigned long long' or anything else because in that case the author would use simple internal type. It seems to me that stating that "the library author(s) wanted to use another type but didn't" is to impugn the author competency. As always we may think that somebody is stupid but it's not polite to behave ourselves as we think so :).

Generally if me make our own wrapper and want to use `uintptr_t` we should simply convert the values between `uintptr_t` and the library type. That's an no-op in expected case and works perfect in unexpected. That's all we need.

alyapunov (approved):

PR #8062, opened 2022-12-14
replication: replicaset state machine assert fail

### LINKED ISSUE #7590: Assertion `replicaset.applier.connected > 0' failed
**Bug description**

While debugging a test we've encountered the error:

```
/home/rosik/w/picodata/tarantool-sys/src/box/replication.cc:502: void replica_on_applier_disconnect(replica*): Assertion `replicaset.applier.connected > 0' failed.
```

Tarantool version: 2.10.0-29-g1ec2a9d15

**Steps to reproduce**

I only have a picodata-based repro: https://git.picodata.io/picodata/picodata/picodata/-/issues/129

Serpentian:
Currently replicaset state machine tracking the number of connected,
loading and synced appliers may perform unnecessary decrementing of
their count. On debug version this may lead to assertion failure.
Here's the way it may happen:
  1. Any kind of exception occurs in applier thread and leads to
     invoking its destructor (applier_thread_data_destroy), which
     is set with scoped guard;
  2. Cbus call is made in order to remove the corresponding applier
     from the thread. According to the fact that cbus_call is
     synchronous, we yield, waiting for the result from the applier
     thread.
  3. During yielding user calls reconfiguration, which invokes
     replicaset_update. Old appliers are pruned: for every replica
     trigger on changing state machine counter is deleted after which
     we stop fiber and wait its join.
  4. If the first replica in replicaset_foreach is not the errored
     one and the errored fiber wakes up during yielding with
     fiber_join, then zero decrementing happens.

Let's clear the above mentioned triggers for all replicas at the
first place and only after that stop and join their applier fibers.

Closes https://github.com/tarantool/tarantool/issues/7590

NO_DOC=bugfix


Serpentian:
Here's the way, the failed assertion on updating of the replicaset's state machine occurs:

1. Any kind of destructing exception occur in one of applier fiber, which works in applier thread, not tx.
2. According to the `applier_subscibe` `applier_thread_data_destroy` is called:
https://github.com/tarantool/tarantool/blob/eb0a4c901fb98390b03bbf5a51c99d858a4ea889/src/box/applier.cc#L2073-L2075
3. In order to notify applier thread about the death of one applier `applier_thread_destroy` push message with `cbus_call` and waits, yielding until the tx knows, applier thread succeded.
4. While yielding in `cbus_call` we invoke reconfiguration, `replicaset_update` is called:
https://github.com/tarantool/tarantool/blob/eb0a4c901fb98390b03bbf5a51c99d858a4ea889/src/box/replication.cc#L652-L680
5. State machine is fully nullified. If the first replica in `replicaset_for_each` is not the one, which is yielding on the `cbus_call`, then reconfiguration yield on `fiber_join`:
https://github.com/tarantool/tarantool/blob/eb0a4c901fb98390b03bbf5a51c99d858a4ea889/src/box/applier.cc#L2309-L2319
6. `cbus_call` gets the message, triggers on changing of appliers state are called, decrementing of zerio happens.

If the first applier in `replicaset_for_each` is the one, which is trying to be destroyed, then trigger, which decrement states, is deleted before we yield on `fiber_join`. Bug is not reproduced

Serpentian:
There are two ways to solve the probem: from applier's point of view, and from the `replicaset_update`. Both of them have downsides:

1. **Applier**.  We may consider replacing `cbus_call` with `cpipe_push` (which is used inside `cbus_call`). Instead of yielding we may wait for some condition to become true (AFAIK, we cannot set second hop for cbus message push, as it won't be executed, if we don't yield. So, it may be waiting for `applier->thread.reader` to become `NULL`). The problem with such solution, that it breaks consistency of applier configuration functions (`applier_thread_detach/attach_applier`, `applier_thread_data_destroy/applier_thread_data_create`). All of them use `applier_cfg_msg`, and we'll have to use `applier_msg` e.g.

2. **replicaset_update**:
```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index b74f84e1b..5f874c938 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -592,6 +592,8 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 	RLIST_HEAD(anon_replicas);
 	struct replica *replica, *next;
 	struct applier *applier;
+	struct applier *appliers_for_delete[VCLOCK_MAX] = {};
+	int applier_for_delete_count = 0;
 
 	auto uniq_guard = make_scoped_guard([&]{
 		replica_hash_foreach_safe(&uniq, replica, next) {
@@ -657,7 +659,6 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 	replicaset_foreach(replica) {
 		if (replica->applier == NULL)
 			continue;
-		ERROR_INJECT_YIELD(ERRINJ_REPLICASET_UPDATE_DELAY);
 		struct replica *other = replica_hash_search(&uniq, replica);
 		if (keep_connect && other != NULL &&
 		    (replica->applier->state == APPLIER_FOLLOW ||
@@ -677,6 +678,11 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 			replica_clear_applier(replica);
 			replica->applier_sync_state = APPLIER_DISCONNECTED;
 		}
+		appliers_for_delete[appliers_for_delete_count++] = applier;
+	}
+
+	for (int i = 0; i < appliers_for_delete_count; ++i) {
+		applier = appliers_for_delete[i];
 		applier_stop(applier);
 		applier_delete(applier);
 	}
``` 

The problem is in test. We must not implicit yield in cycle, but skip the replica, if it's uuid equals to the one, which is yielding on `cbus_call` now. How can we do it, if we want that solution? We may save some bits of uuid in `ERRINJ`, but we're going to change uuid sytle in the nearest future.

We may also use not the test, which fails 100% on debug, but sometimes. I don't like that solution either.

sergepetrenko:
Good job with finding a stable reproducer! This was a hard one.

Now on the variants you propose:
1. Regarding replacing `cbus_call` with `cpipe_push`. It seems we don't have to wait for anything. The `applier_thread_detach_applier` may be done asynchronously. That is, applier might simply push the destruction message to cpipe and exit immediately. Maybe some extra changes to the code are needed, but this is doable, I think. Still, the 2nd variant seems better.
2. Regarding clearing triggers without yields in `replicaset_update`. This variant sounds like the right one. As soon as replicaset_update is run, we should ignore old triggers trying to change replicaset state.
For example, if we choose variant 1, we might introduce some new yields in applier later on, and the same problem will show up once again. This won't happen if we choose variant 2.
If you're choosing between a 100% reproducer and a good fix, better always choose the good fix.
3. Regarding the test: I think there's some way to make it work for the 2nd variant. It seems you don't need the ERRINJ_REPLICASET_UPDATE_DELAY. Just need to create a test with 3 replicas and make it so that the first replica's applier stop is delayed, making replicaset_update yield in fiber_join. Maybe even ERRINJ_APPLIER_DESTROY_DELAY will do the job. It will delay both  appliers (the errored one and the one exiting normally). Need to find a way to wake up the errored applier first. You might need 2 error injections for that: ERRINJ_APPLIER_DESTROY_DELAY and ERRINJ_APPLIER_NORMAL_STOP_DELAY, or something similar. So that you control which applier is delayed separately.

Long story short, I like the variant with replicaset_update more. I have a raw idea about how you can make a stable test then, but even if this doesn't work out, it's fine. The test is less important than the fix.

Serpentian:
> Need to find a way to wake up the errored applier first. You might need 2 error injections for that: ERRINJ_APPLIER_DESTROY_DELAY and ERRINJ_APPLIER_NORMAL_STOP_DELAY, or something similar. So that you control which applier is delayed separately.

Actually, the problem is only partially in the order, appliers're waked up. The main problem is in deleting a trigger in `replica_clear_applier`. So, if the first replica in `replicaset_foreach` is the errored one, then, firstly, it's trigger is deleted, and after that it is waked up. Zero decrementing doesn't happen. We cannot

Serpentian:
Will transfer from the Draft PR to the ordinary one as soon everything will be ready

Serpentian:
Smth hangs on release testing. Will check it tomorrow

src/box/applier.cc
```diff
@@ -1936,6 +1936,18 @@ applier_thread_data_destroy(struct applier *applier)
 	msg.applier = applier;
 	cbus_call(&thread->thread_pipe, &thread->tx_pipe, &msg.base,
 		  applier_thread_detach_applier);
+	ERROR_INJECT(ERRINJ_APPLIER_DESTROY_DELAY, {
+		struct errinj *inj = errinj(ERRINJ_WAL_IO, ERRINJ_BOOL);
+		if (inj->bparam) {
+			inj->bparam = false;
+			say_warn("applier failure destroy is delayed");
+			ERROR_INJECT_YIELD(ERRINJ_APPLIER_FAIL_DESTROY_DELAY);
```
sergepetrenko, line 1944: These three error injections are extremely interconnected.

Maybe move the ERRINJ_APPLIER_FAIL_DESTROY_DELAY here?
https://github.com/tarantool/tarantool/blob/dfe79b9aa30e9aad9b105fc559df7a9f08b8fe5b/src/box/applier.cc#L2229-L2232

sergepetrenko (changes requested):
Thanks for working on this!

The fix looks good. As for the test -- let's simplify these error injections. Maybe then it'll be easier to make the test 100% stable

src/box/applier.cc
```diff
@@ -1936,6 +1936,18 @@ applier_thread_data_destroy(struct applier *applier)
...
+			inj = errinj(ERRINJ_APPLIER_DESTROY_DELAY, ERRINJ_BOOL);
+			inj->bparam = false;
```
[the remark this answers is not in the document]
Serpentian: You're right. Great idea! Let's forbid relay of the first instance to send anything at all. Now, the commit without changes in src/box/replication.cc is the 100% reproducer.

src/box/applier.cc
Serpentian: > Or maybe even better, leave APPLIER_FAIL_DESTROY_DELAY in applier_thread_data_destroy, but move APPLIER_DESTROY_DELAY to the place where FiberIsCancelled is caught.

`applier_thread_data_destroy` works for all appliers, it's really difficult to manipulate connected error injections, which yields on different "levels" of applier destruction. Unfortunately, I couldn't make this work.  Here's the problem.

Changes you proposed can indeed reproduce the problem, but it's not capable of testing the patch. As soon as `APPLIER_FAIL_DESTROY_DELAY` stops yielding, triggers're run, assertion fails. After applying patch (src/box/replication.cc) the test must end successfully. But it won't because all other appliers should be joined successfully, but they still yields. So we must set `APPLIER_DESTROY_DELAY` manually to `false` right after `APPLIER_FAIL_DESTROY_DELAY`.

These two error injections are still pretty connected, but it's much better now. They're both in `applier_f` and look pretty consistent.

src/box/applier.cc
```diff
@@ -2227,6 +2227,20 @@ applier_f(va_list ap)
...
+					ERRINJ_APPLIER_FAIL_DESTROY_DELAY);
+					/*
+					 * All fibers, stopped with fiber_cancel
+					 * can end their work now.
+					 */
+					struct errinj *inj = errinj(
+						ERRINJ_APPLIER_DESTROY_DELAY,
+						ERRINJ_BOOL);
+					inj->bparam = false;
+				});
```
Serpentian, line 2243: This looks pretty ugly due to the requirements of `checkpatch` for indenting. But there's nothing I can do about it(

src/box/applier.cc
```diff
@@ -2227,6 +2227,13 @@ applier_f(va_list ap)
 				goto reconnect;
 			} else {
 				/* Unrecoverable errors */
+				ERROR_INJECT(
+					ERRINJ_APPLIER_FAIL_DESTROY_DELAY, {
+					say_warn("applier fail delay destroy");
```
Serpentian, line 2232: If we don't want to grep log, there's an alternative: we may also create incrementing error injection: 0 - nothing happened, 1 - yielding started, 2 - yielding ended, But in general it will be just the same, just no log output and inability to use `ERROR_INJECT_YIELD` macros

src/box/applier.cc
Serpentian: Now error injections are not connected at all. But one of them just output to log some info

Serpentian:
@Gerold103, should we wait for you?

src/box/replication.cc
```diff
@@ -591,6 +591,8 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 	RLIST_HEAD(anon_replicas);
 	struct replica *replica, *next;
 	struct applier *applier;
+	struct applier *appliers_for_delete[VCLOCK_MAX] = {};
+	int appliers_for_delete_count = 0;
```
sergepetrenko, line 595: nit: let's declare these variables lower, where they're actually used.

test/replication-luatest/gh_7590_replicaset_assert_test.lua
```diff
@@ -0,0 +1,91 @@
...
+        server.build_listen_uri('r1'),
+        server.build_listen_uri('r2'),
+        server.build_listen_uri('r3'),
+    }
+
+    g.r1 = g.cluster:build_and_add_server({alias = 'r1', box_cfg = {
+        -- We need to specify uuid explicitly, so that rb tree traversal
+        -- wouldn't take errored applier (applier from r2 to r3) first.
+        instance_uuid = '00000000-0000-0000-0000-000000000001',
+        replication = replication_cfg,
```
sergepetrenko, line 19: Let's also set `replication_timeout = 0.1`, like it's done in other tests.
This speeds them up quite a bit. For example, your test would run for ~1 second instead of ~3 on my machine.

test/replication-luatest/gh_7590_replicaset_assert_test.lua
```diff
@@ -0,0 +1,91 @@
...
+
+g.after_all(function(g)
+    g.cluster:drop()
+end)
+
+g.test_replicaset_state_machine_assert_fail = function(g)
+    g.r1:exec(function()
+        -- Forbid r1 to send any kind of data in order to avoid
+        -- triggering destruction of the applier from r1 to r3.
+        box.error.injection.set('ERRINJ_RELAY_SEND_DELAY', true)
```
sergepetrenko, line 48: The test still hangs occasionally when run for 256 times in 32 jobs.
I think the reason is that RELAY_SEND_DELAY forbids relay to send heartbeats, and sometimes r2 destroys applier due to time out.

My first idea was to replace `ERRINJ_RELAY_SEND_DELAY` with `ERRINJ_WAL_DELAY`. This way relay continues to write heartbeats, but doesn't send any actual data, because no data can be written on r1.
But this would lead to timeouts between `r1` and `r2`.

Let's simply reconfigure r1 to not replicate from r2 after bootstrap. No error injections needed.

sergepetrenko: P.S. even after trying ERRINJ_WAL_DELAY and reconfiguring replication the test hangs for a couple out of 256 runs. Please, check it out on your side.

src/box/applier.cc
sergepetrenko: Yep, ugly indeed.
Try this:
```c
struct errinj *inj = errinj(ERRINJ_APPLIER_FAIL_DESTROY_DELAY);
applier_log_error(...);
applier_disconnect(...);
if (inj == NULL)
        return -1;
/* all the error injection stuff goes here. */
```

src/box/applier.cc
sergepetrenko: grepping the log seems just fine IMO.

sergepetrenko (changes requested):
Thanks for the fixes!

Almost done, just some minor comments left.

Gerold103:
> @Gerold103, should we wait for you?

Yes.

src/box/applier.cc
Serpentian: We must yield before running triggers. This one won't work. We may replace macro with explicit code. But in general it's so unimportant

src/box/applier.cc
Serpentian: Yep, seems fine for me too

src/box/replication.cc
Serpentian: Fixed

test/replication-luatest/gh_7590_replicaset_assert_test.lua
Serpentian: Fixed

test/replication-luatest/gh_7590_replicaset_assert_test.lua
Serpentian: Tried to reproduce hang locally on my laptop and desktop. Unfortunately, it doesn't hang at all, no matter, how many tests are run or how many threads are used. Don't have access to dev machines now too(

src/box/applier.cc
sergepetrenko: > We must yield before running triggers.

Yes, indeed. Missed that.

test/replication-luatest/gh_7590_replicaset_assert_test.lua
```diff
@@ -0,0 +1,93 @@
...
+    end)
+
+    t.helpers.retrying({timeout = 5}, function()
+        if not g.r3:grep_log('applier fail destroy continue') then
+            error("Applier destruction haven't continued yet")
+        end
+    end)
+
+    g.r3:exec(function()
+        box.error.injection.set('ERRINJ_APPLIER_DESTROY_DELAY', false);
```
sergepetrenko, line 91: Let's unset "ERRINJ_WAL_IO" as well, just for consistency.

sergepetrenko (approved):
Thanks for the patch!

Side note: please, don't forget to re-request the review once you're ready.

changelogs/unreleased/gh-7590-replicaset-assert.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/replication
+
+* Fixed a bug when a replicaset state machine, tracking the number of
+  appliers according to their states, could become inconsistent during
+  reconfiduration (gh-7590).
```
Gerold103, line 5: `reconfiduration` -> `reconfiguration`.

src/box/applier.cc
```diff
@@ -2258,6 +2267,7 @@ applier_f(va_list ap)
 			applier_disconnect(applier, APPLIER_LOADING);
 			goto reconnect;
 		} catch (FiberIsCancelled *e) {
+			ERROR_INJECT_YIELD(ERRINJ_APPLIER_DESTROY_DELAY);
```
Gerold103, line 2270: The problem with both those error injections is that they are added to impossible places. Meaning that without those injections there couldn't be any yields at those lines. Basically, you introduce a bug with the new yields and then fix it, alongside apparently fixing the original bug too.

Instead, error injections should reflect what can happen in reality. Since you needed only a delay for applier fiber join, you could add the injection right before or after that fiber join. Or add a usleep() injection inside the applier thread where needed. These would actually simulate something that can happen in reality.

src/box/replication.cc
```diff
@@ -653,6 +653,10 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 	replicaset.applier.connected = 0;
 	replicaset.applier.loading = 0;
 	replicaset.applier.synced = 0;
```
Gerold103, line 655: Why do we even nullify those counters? Wouldn't it be more natural to let them be updated via the normal flow? Not break it by this nullification? So we keep the counters unchanged, destroy the appliers, they will themselves properly decrement all the counters, and all will be fine. Then you wouldn't need to rely on which functions yield and which not. For example, currently your fix won't work anymore if `replica_clear_applier` or `replica_delete` would start yielding.

test/replication-luatest/gh_7590_replicaset_assert_test.lua
```diff
@@ -0,0 +1,93 @@
...
+    g.r1:wait_for_vclock_of(g.r2)
+end)
+
+g.after_all(function(g)
+    g.cluster:drop()
+end)
+
+g.test_replicaset_state_machine_assert_fail = function(g)
+    g.r1:exec(function(cfg)
+        -- Let's drop replication from r2 instance so that it cannot
```
Gerold103, line 45: The comment is extremely misleading. Or rather the node naming. Somewhy your order is [r1, r3, r2] instead of [r1, r2, r3]. I spent quite some time trying to understand why `dropping r2` means `replication[3] = nil`. Can you change it?

test/replication-luatest/suite.ini
```diff
@@ -2,4 +2,4 @@
 core = luatest
 description = replication luatests
 is_parallel = True
-release_disabled = gh_5295_split_brain_test.lua gh_6036_qsync_order_test.lua gh_6842_qsync_applier_order_test.lua gh_6033_box_promote_demote_test.lua gh_7253_election_long_wal_write_test.lua gh_7086_box_issue_promote_assert_test.lua
+release_disabled = gh_5295_split_brain_test.lua gh_6036_qsync_order_test.lua gh_6842_qsync_applier_order_test.lua gh_6033_box_promote_demote_test.lua gh_7253_election_long_wal_write_test.lua gh_7086_box_issue_promote_assert_test.lua gh_7590_replicaset_assert_test.lua
```
Gerold103, line 5: AFAIU, we should prefer `t.tarantool.skip_if_not_debug()` right in the test code instead of the `release_disabled` list which is a remnant of diff-tests. But that is not 100%.

Gerold103 (approved):
Thanks for the patch! To be clear - it looks good enough already. But I think it could be done much clearer. Up to the author and other reviewers if want to improve it.

test/replication-luatest/suite.ini
sergos: @ylobankov could you please comment?

ylobankov: I think it is better to use luatest tools to skip such tests. When you run this test via luatest directly, it will be skipped as well.

test/replication-luatest/gh_7590_replicaset_assert_test.lua
Serpentian: It was done above, but all right, le's put it here, it doesn't affect anything

changelogs/unreleased/gh-7590-replicaset-assert.md
Serpentian: My bad. Fixed

test/replication-luatest/gh_7590_replicaset_assert_test.lua
Serpentian: Replication array cannot be {rep_a, nil, rep_b}. It was done to workaround this one. Fixed, it's clearer now.

src/box/applier.cc
Serpentian: Finally manged to write a 100% reproducer with good error injections. Thanks for the hint)

src/box/replication.cc
Serpentian: The short answer is that it's too difficult to implement that fix for such a rare bug. Here's the explanation.

Every applier has a list of trigger, the second of which is the one, we're interested in. It's set as soon as we created replica and assign applier to it. Applier cannot belong to several replicas at the same time.

On configuration the list of appliers is created in `box_sync_replication`. Appliers are started, and only after that they go down, to `replicaset_update`, where replicas are created, applier are assigned to them. By this time applier can have any state already. So, triggers didn't update state machine. 

The fix could be to update state machine right from applier, but not from triggers. But it would break consistency of the code.

src/box/replication.cc
```diff
@@ -653,6 +653,11 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
...
+
+	struct applier *appliers_for_delete[VCLOCK_MAX] = {};
+	int appliers_for_delete_count = 0;
+	int csw = fiber()->csw;
```
Serpentian, line 659: Which method of initialization of such variables is preferable: to wrap `csw` in `ifndef NDEBUG` or just to leave everything as it is.

Excess memory is used in release for a variable, we don't use.  But AFAICS, we don't wrap such things as I proposed in general.

sergepetrenko: I'd leave it as is, no need to worry

test/replication-luatest/gh_7590_replicaset_assert_test.lua
sergepetrenko: > Replication array cannot be {rep_a, nil, rep_b}.

You could use table.remove for that. Anyway, your current version looks fine as well.

sergepetrenko (approved):
Thanks! Looks good.

src/box/applier.cc
Gerold103: Thank you!

Gerold103 (approved):
Thanks for the fixes! I am really surprised you didn't chuck my errinj comment, thanks for that :)

sergepetrenko:
Will backport to 2.10 in scope of #8682

PR #8063, opened 2022-12-14
Don't replace `stdin` with `/dev/urandom` after fork

### LINKED ISSUE #7886: flaky test: `app-luatest/gh_5747_crash_multiple_args_test.lua`
The test frequently fails:
```
[024] app-luatest/gh_5747_crash_multiple_args_test.l>
[024] TAP13 parse failed ('utf-8' codec can't decode bytes in position 162-163: invalid continuation byte).
[024]
[024] No result file (app-luatest/gh_5747_crash_multiple_args.result) found.
```

It fails because the result file contains binary data:
```sh
TAP version 13
1..1
# Started on Wed Nov  2 18:45:11 2022
# Starting group: app-luatest.gh_5747_crash_multiple_args
LuajitError: stdin:1: malformed number near '1é³s<9e>e'
ok     1        app-luatest.gh_5747_crash_multiple_args.test_multiple_args
# Ran 1 tests in 0.083 seconds, 1 succeeded, 0 failed
```

~~The error means that the argument of the `-e` option is somehow corrupted.~~ Here is an example of a similar error message:
```
$ tarantool -e 1xxx
LuajitError: (command line):1: malformed number near '1xxx'
```

~~Maybe it's a bug in `getopt_long`, or its usage.~~ Valgrind and ASan don't detect anything.

According to the dashboard, the test fails on centos_7, centos_8, debian_9, debian_10, ubuntu_16_04, ubuntu_18_04.
But it always pass on fedora_34, fedora_35, fedora_36, opensuse_15_1, opensuse_15_2, ubuntu_20_04, ubuntu_22_04.
Therefore it looks like a `glibc < 2.31` issue.

Gumix:
**box: do not close xlog file descriptors in the atfork handler**

Use `O_CLOEXEC` flag instead.

If Tarantool is forked before executing `box.cfg{}`, e.g. using `io.popen()`, the child process could start with `stdin` linked to `/dev/urandom`. This happens because `wal_writer_singleton` and `vy_log_writer` are not yet initialized, i.e. `fd` fields are 0, then atfork child handler `wal_atfork()` is called. It checks that xlog is opened (the check succeeded as 0 != -1) and closes its fd 0, in fact closing `stdin` (twice). Next, Tarantool opens the file `/dev/urandom` during initialization, and it receives the lowest unused file descriptor, which is 0. Then `luaL_loadfile()` loads `stdin` as a Lua chunk, in effect reading random numbers.

This happens on glibc 2.28 and older, as newer versions do not invoke atfork handlers during `io.popen()`: https://sourceware.org/bugzilla/show_bug.cgi?id=17490

Closes #7886

### DIFF
--- a/changelogs/unreleased/gh-7886-dont-close-xlog-fd-atfork.md
+++ b/changelogs/unreleased/gh-7886-dont-close-xlog-fd-atfork.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when Tarantool could execute random bytes as a Lua code after fork
+  on systems with glibc version < 2.29 (gh-7886).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4490,16 +4490,6 @@ box_cfg(void)
 	}
 }
 
-/**
- * box.coredump() forks to save a core. The entire
- * server forks in box.cfg{} if background=true.
- */
-void
-box_atfork(void)
-{
-	wal_atfork();
-}
-
 int
 box_checkpoint(void)
 {
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -107,12 +107,6 @@ box_cfg(void);
 bool
 box_is_configured(void);
 
-/**
- * A pthread_atfork() callback for box
- */
-void
-box_atfork(void);
-
 /** Check if the slice of main cord has expired. */
 int
 box_check_slice_slow(void);
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -1519,19 +1519,3 @@ wal_notify_watchers(struct wal_writer *writer, unsigned events)
 	rlist_foreach_entry(watcher, &writer->watchers, next)
 		wal_watcher_notify(watcher, events);
 }
-
-
-/**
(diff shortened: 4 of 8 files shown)


changelogs/unreleased/gh-7886-dont-close-xlog-fd-atfork.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when Tarantool could execute random bytes as a Lua code after fork
+  (gh-7886).
```
locker, line 4: Nit: I'd explicitly mention in the changeog that it was only possible with glibc <= 2.28.

locker (approved):

changelogs/unreleased/gh-7886-dont-close-xlog-fd-atfork.md
Gumix: Done

alyapunov (approved):

alyapunov:
backported to 2.10.

PR #8064, opened 2022-12-15
sql: introduce SEQSCAN to SELECT

### LINKED ISSUE #7747: Add SEQSCAN keyword for SELECT queries that do not use indexes
This issue suggest to throw an error in case keyword SEQSCAN was not used in select where indexes were not used.

SEQSCAN example:
```
SELECT a FROM SEQSCAN t;
```

The default use of SEQSCAN can be disabled by setting the new `seq_scan` session option to `true`. The default value of the option is `false`.

ImeevMA:
This patch introduces new keyword SEQSCAN and new restrictions on SELECTs.

Closes #7747

@TarantoolBot document
Title: SEQSCAN

Now scanning SELECT will not run and will throw an error if the new SEQSCAN keyword is not used for scanned spaces. This change only affects SELECT and does not affect UPDATE and DELETE. A SELECT is recognized as a scanning SELECT if `EXPLAIN QUERY PLAN SELECT ...` indicates that the SELECT `scans` rather than `searches`.

For example, if we have spaces created with these queries:
```
CREATE TABLE t(i INT PRIMARY KEY, a INT);
CREATE TABLE s(i INT PRIMARY KEY, a INT);
```

Then these queries will throw an error:
```
SELECT * FROM t;
SELECT * FROM t WHERE a > 1;
SELECT * FROM t WHERE i + 1 = 5;
SELECT * FROM t, s;
SELECT * FROM t JOIN s;
```

And these will not:
```
SELECT * FROM t WHERE i > 1;
SELECT * FROM SEQSCAN t;
SELECT * FROM SEQSCAN t WHERE i + 1 = 5;
SELECT * FROM SEQSCAN t, SEQSCAN s;
SELECT * FROM SEQSCAN t JOIN SEQSCAN s;
```


test/sql-luatest/seq_scan_test.lua
```diff
@@ -0,0 +1,177 @@
...
+
+        res, err = box.execute([[SELECT * FROM SEQSCAN t WHERE i + 1 > 2;]])
+        t.assert(res ~= nil and err == nil)
+
+        res, err = box.execute([[SELECT * FROM SEQSCAN t WHERE a > 2;]])
+        t.assert(res ~= nil and err == nil)
+    end)
+end
+
+g.test_seq_scan_joins = function()
```
Gerold103, line 81: From the joins test it seems that JOINs never work without SEQSCAN. Aren't there cases when JOINs do search, not scanning?

Gerold103 (approved):
Thanks for the patch!

test/sql-luatest/seq_scan_test.lua
ImeevMA: Thank you for the review! There is. Added a few tests.

src/box/session_settings.c
```diff
@@ -49,6 +49,7 @@ const char *session_setting_strs[SESSION_SETTING_COUNT] = {
 	"sql_recursive_triggers",
 	"sql_reverse_unordered_selects",
 	"sql_select_debug",
+	"sql_seq_scan",
```
locker, line 52: I don't see any information about this session setting key in the doc bot request.

ImeevMA: Thank you for the review! Added.

changelogs/unreleased/gh-7747-seqscan.md
```diff
@@ -0,0 +1,4 @@
+## feature/sql
+
+* A new SEQSCAN keyword has been introduced and a new restrictions on scanning
+  SELECTs has been applied (gh-7747).
```
locker, line 4: Shouldn't it be marked as a breaking change?

src/box/errcode.h
```diff
@@ -305,6 +305,7 @@ struct errcode_record {
 	/*250 */_(ER_INVALID_AUTH_DATA,		"Invalid '%s' data: %s") \
 	/*251 */_(ER_INVALID_AUTH_REQUEST,	"Invalid '%s' request: %s") \
 	/*252 */_(ER_WEAK_PASSWORD,		"Password doesn't meet security requirements: %s") \
+	/*253 */_(ER_SEQ_SCAN,			"Scanning is not allowed for %s") \
```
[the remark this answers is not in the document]
ImeevMA: Changed to `ER_SQL_SEQ_SCAN`. Rebased.

changelogs/unreleased/gh-7747-seqscan.md
ImeevMA: Added.

locker (approved):

locker:
A few integration tests fail:
- [tarantool-c](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108505)
- [tarantool-python](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385109147)
- [php-client](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108983)
- [go-tarantool](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108836)

Please fix.

ImeevMA:
> A few integration tests fail:
> 
>     * [tarantool-c](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108505)
> 
>     * [tarantool-python](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385109147)
> 
>     * [php-client](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108983)
> 
>     * [go-tarantool](https://github.com/tarantool/tarantool/actions/runs/3757505591/jobs/6385108836)
> 
> 
> Please fix.

I set the old behavior to be default and created issue #8096. I also removed the words **Breaking change** from the changelog since the current patch doesn't break anything.

test/sql/bind.test.lua
```diff
@@ -1,5 +1,6 @@
 netbox = require('net.box')
 test_run = require('test_run').new()
+box.execute([[SET SESSION "sql_seq_scan" = true;]])
```
locker, line 3: Why do you set the value to true if it's true by default anyway?
Anyway, a proper way to fix the tests would be to use the new `SEQSCAN` keyword, wouldn't it?

ImeevMA: Even if default value is `true`, it is not set to stone and setting it expilictly allow to not think about default value. As for SEQSCAN in all tests - why? These test were added not for testing SEQSCAN, so why not just use the option where I can?

locker: AFAIU the point of this new option is compatibility, that is in the long run, we might deprecate the option and require all users to use the `SEQSCAN` keyword instead. (Note that after a few releases, compat module options are marked obsolete and can't be changed anymore. Actually, if we used the compat module, I wouldn't even bother introducing a per session option at all, but I think it shouldn't hurt.) Our tests should always use the proper most up-to-date approach. Otherwise, other people might add more tests using this session option instead of the `SEQSCAN` keyword as a workaround for full scan.

I'm leaving this up to you, since you develop sql mostly by yourself. If you're still sure that using this session option in tests like that is fine, reply so in this thread and reassign the PR back to me - I'll merge it as is.

ImeevMA: No, the option is essential. We decided to leave users the ability to use SQL the way they used it before. It has nothing to do with compat. This was properly described in the document and I see no reason to believe that oprion can be dropped. Compat, in my opinion, is just what is not needed here, but now I had to use it anyway.

Using the option to avoid using SEQSCAN is a valid use case.

locker (approved):

locker:
integration/tarantool-python [fails](https://github.com/tarantool/tarantool/actions/runs/3782507027/jobs/6430487453). @ImeevMA please check if it's caused by this PR.

PR #8079, opened 2022-12-19
test: skip box-luatest/gh_7743_term_initial_cfg_snap_test

locker:
The test is extremely flaky, especially on coverage builds. Let's skip it until https://github.com/tarantool/tarantool-qa/issues/297 is fixed.


locker:
This test is really annoying - CI [failed](https://github.com/tarantool/tarantool/actions/runs/3731928290/jobs/6331266708) after spending more than 40 minutes trying to pass coverage:

image

locker:
The test should be fixed properly in a few hours so this PR isn't needed.

Issue #8083, opened 2022-12-20
Crash report is collected for SIGSEGV only in the main thread


Gumix:
**Bug description**

If segmentation fault occur in any non-main thread, the crash report is not printed, because currently SIGSEGV has a handler only in the main thread. In other threads SIGSEGV are blocked, however according to [man sigprocmask](https://man7.org/linux/man-pages/man2/sigprocmask.2.html):
>     If SIGBUS, SIGFPE, SIGILL, or SIGSEGV are generated while they
>     are blocked, the result is undefined, unless the signal was
>     generated by kill(2), sigqueue(3), or raise(3).

* OS: Linux
* OS Version: any
* Architecture: x86_64

```
Tarantool 2.10.4-36-g9a97707b9
Target: Linux-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: GNU-8.3.0
C_FLAGS: -fexceptions -funwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/gi
t/tarantool=. -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror
CXX_FLAGS: -fexceptions -funwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/
git/tarantool=. -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -We
rror
```

**Steps to reproduce**

Apply the patch from #8023, then run `box.cfg{}` in a console without a debugger attached.

**Actual behavior**

```
tarantool> box.cfg{}
2022-12-20 18:37:09.621 [12549] main/103/interactive I> Tarantool 2.10.4-36-g9a97707b9
2022-12-20 18:37:09.621 [12549] main/103/interactive I> log level 5
2022-12-20 18:37:09.621 [12549] main/103/interactive I> wal/engine cleanup is paused
[...]

Issue #8023, opened 2022-12-07
MacOS threads are immortal when block signals?! - very cringe


Gerold103:
Apply this patch on MacOS:
```Diff
diff --git a/src/lib/core/fiber.c b/src/lib/core/fiber.c
index 8d1dbd76d..61bae6072 100644
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -1730,6 +1730,8 @@ void *cord_thread_func(void *p)
 	ct_arg->is_started = true;
 	tt_pthread_cond_signal(&ct_arg->start_cond);
 	tt_pthread_mutex_unlock(&ct_arg->start_mutex);
+	void (*callback)(void) = (void *)0x5050505050;
+	callback();
 	void *res = f(arg);
 	/*
 	 * cord()->on_exit initially holds NULL. This field is
```
then run `box.cfg{}` in a console without a debugger attached. The process will hang, but will not crash.

The reason is that whatever signal this diff raises (`SIGSEGV` or `SIGABRT`), it is not delivered to the main thread. Apparently the reason is that all non-main threads have all the signals blocked to force their delivery via the main thread.

On Mac in such case (again, according to experiments) there is only a **"mach exception"** thrown. Normally when signals are not blocked, they automatically turn into needed signals. But when the thread sigmask is full, the crash handling is stuck at the mach exceptions stage.

There is no a simple fix like `try-catch` the whole thread or set some kind of a callback to handle all exceptions.

I don't know an ideal solution, but the possible directions to investigate:
- Maybe there is just one Mac-specific signal, whose blockage causes normal signals being stuck? Look at the list of Mac signals and their purpose. Try to unblock some of them which are not critical to handle only in the main thread.
- Try to make a global exception handler. In the end there is diff attached and some links how to do that. In the diff the problem is that (1) when a crash happens in the main thread, it hangs, (2) crashes in non-main thread during their handling in the main-thread's signal handler probably won't see a correct callstack, but it wasn't checked, (3) it might affect users who have their own mach exception handlers installed.
- Try to make a thread-local exception handler. This might be possible from what I managed to find.
- Unblock the crashing signals on Mac in all threads. Or maybe not just on Mac. Need to see how that would affect core/crash module.

---
Links
* Something about mach threads, perhaps there is a way to install a per-thread exception handler: http://gnu.ist.utl.pt/software/hurd/gnumach-doc/mach_7.html;
[...]

PR #8088, opened 2022-12-22
core: unblock fatal (crashing) signals in all threads

Gumix:
**crash: handle all unblockable signals**

There are 4 fatal signals that cannot be blocked by `sigmask()` if they are caused by the CPU exception (rather than kill, sigqueue or raise): SIGILL, SIGBUS, SIGFPE, SIGSEGV. Currently the crash module handles only SIGSEGV and SIGFPE.
This patch adds handlers for SIGBUS and SIGILL. SIGBUS is usually raised by macOS on access to the unmapped memory, and SIGILL is possible, for example, while running AVX version of memcpy on a CPU without AVX support.

Faulting address `siginfo->si_addr` is valid for all these signals, so print it unconditionally.

**core: unblock fatal (crashing) signals in all threads**

Currently all non-main threads have all the signals blocked, however according to `man pthread_sigmask`:

> If SIGBUS, SIGFPE, SIGILL, or SIGSEGV are generated while they
> are blocked, the result is undefined, unless the signal was
> generated by kill(2), sigqueue(3), or raise(3).

On macOS they are actually blocked, causing the faulting instruction to loop indefinitely. While on Linux they are not blocked, however the signal handler registered by `sigaction()` is not executed. Don't block them.

Closes https://github.com/tarantool/tarantool/issues/8023
Closes https://github.com/tarantool/tarantool/issues/8083


Gerold103 (approved):
Thanks for the great fix!

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #5983, opened 2021-04-12
Brew install Apple M1 chip error


yuriysergeev:
**Bug description**
Tarantool doesn't install from 'brew install tarantool' or 'brew install --build-from-source tarantool' on Apple M1
* OS: macOS 
* OS Version: Darwin Kernel Version 20.3.0
* Architecture: arm64

**Steps to reproduce**

Run two ways tarantool installations via homebrew:

1. Type 'brew install tarantool' in console: 
`Installing tarantool --HEAD
==> cmake . -DCMAKE_INSTALL_MANDIR=/opt/homebrew/Cellar/tarantool/HEAD-b88df4d_1/share/doc/tarantool -DCMAKE_INSTALL_S
==> make
Last 15 lines from /Users/ys/Library/Logs/Homebrew/tarantool/02.make:
[  3%] Building C object CMakeFiles/zstd.dir/third_party/zstd/lib/compress/hist.c.o
[...]

PR #8093, opened 2022-12-24
test: stabilize jit.dump() smoke test

Totktonada:
A jiggle in tarantool's initialization code can lead to trace numbers
above 9. We can either:

* accept trace numbers above 9 in the test
* or drop traces from the initialization code in the test and assume
  that the new trace will have number 1.

Igor Munkin suggested to stick with the second approach to avoid
dependency on the initialization code.

### DIFF
--- a/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua
+++ b/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua
@@ -24,11 +24,11 @@ if #arg == 0 then
   local proc = io.popen(cmd)
   local got = proc:read('*all'):gsub('^%s+', ''):gsub('%s+$', '')
   local expected = table.concat({
-      '---- TRACE %d start',
-      '---- TRACE %d IR',
-      '---- TRACE %d mcode',
-      '---- TRACE %d stop',
-      '---- TRACE %d exit',
+      '---- TRACE 1 start',
+      '---- TRACE 1 IR',
+      '---- TRACE 1 mcode',
+      '---- TRACE 1 stop',
+      '---- TRACE 1 exit',
   }, '.+')
 
   test:like(got, expected , 'jit.dump smoke tests')
@@ -40,5 +40,7 @@ end
 require('jit.dump').start('+tbisrmXaT')
 -- Tune JIT engine to make the test faster and more robust.
 jit.opt.start('hotloop=1')
+-- Purge all existing traces generated on Tarantool startup.
+jit.flush()
 -- Record primitive loop.
 for _ = 1, 3 do end


igormunkin (changes requested):
@Totktonada, thanks for the patch!

We discussed with @Buristan these particular changes and came to the next solution: since JIT is quite an unpredictable and hardly manageable mechanism, it's better to call `jit.flush()` prior to test assertions **in all JIT specific tests** (you can see my recent struggle within `LUAJIT_ENABLE_CHECKHOOK`-related issue). As a result the JIT tests will be less fragile and less changes can break them (I hope @Buristan will drop a few words regarding hot counting).

All in all here is the diff I suggest to apply instead of the current changes:
```diff
diff --git a/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua b/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua
index 72caec2f9..3c671f0ce 100755
--- a/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua
+++ b/test/app-tap/gh-5983-jit-library-smoke-tests.test.lua
@@ -24,11 +24,11 @@ if #arg == 0 then
   local proc = io.popen(cmd)
   local got = proc:read('*all'):gsub('^%s+', ''):gsub('%s+$', '')
   local expected = table.concat({
-      '---- TRACE %d start',
-      '---- TRACE %d IR',
-      '---- TRACE %d mcode',
-      '---- TRACE %d stop',
-      '---- TRACE %d exit',
+      '---- TRACE 1 start',
+      '---- TRACE 1 IR',
+      '---- TRACE 1 mcode',
+      '---- TRACE 1 stop',
+      '---- TRACE 1 exit',
   }, '.+')
 
   test:like(got, expected , 'jit.dump smoke tests')
@@ -40,5 +40,7 @@ end
 require('jit.dump').start('+tbisrmXaT')
 -- Tune JIT engine to make the test faster and more robust.
 jit.opt.start('hotloop=1')
+-- Purge all existing traces generated on Tarantool startup.
+jit.flush()
 -- Record primitive loop.
 for _ = 1, 3 do end
 ```
 
 It works on my local machine.

Totktonada:
@igormunkin It works for me too. Applied your variant of the fix.

igormunkin (approved):
@Totktonada, thanks for the changes! LGTM now.

igormunkin:
Also backported to 2.10 in scope of d4f6e79.

Buristan:
Hi, Alexander, Igor. Thanks for the patch and discussion!
LGTM. 

As we discussed offline with @igormunkin, that too many traces on Tarantool startup may spoil JIT hotcount, so some traces will be recorded earlier than expected. But it is not so probable for now. 
Also, some background fibers for similar tests (but with `fiber.yield()`) may silently record a trace.

Issue #5997, opened 2021-04-14
box: get rid of _schema.max_id


Korablev77:
The only purpose of this thing is to generate increasing space ids in `box.schema.create_space()` - it is neither transactional (max_id is increased before invocation of `on_replace_dd` trigger and in case creation of space fails, its value is not rolled back) nor safe (in case creation of space takes place via direct `box.space._space:replace()` - value of max_id is not changed; so in this case next call of `box.schema.create_space()` may fail due to duplicate space_id).
The suggestion is to replace _schema.max_id with `box.space._space.index[0]:max()[1]` or attached sequence.

kostja:
If you use max() you'll break replication by creating a space with id of a dropped one.
Using an attached sequence should be fine.

PR #8097, opened 2022-12-27
box: use maximal space id instead of _schema.max_id

### LINKED ISSUE #8036: Failed to mix space create with and without explicit id
**Bug description**

If some space was created with `box.schema.space.create` with explicit id, following `CREATE TABLE` SQL query may try to use the same id and fail with `Duplicate key exists in unique index "primary" in space "_space"`.

* OS: Linux
* OS Version: Ubuntu 22.04
* Architecture: amd64

Tarantool Enterprise 2.10.3-0-g3c268adfd-r510-nogc64

**Steps to reproduce**

```
tarantool> box.cfg{}
---
...

tarantool> box.schema.space.create('space', {format = {{name = 'id', type = 'unsigned'}}, id = 512})
---
- engine: memtx
  before_replace: 'function: 0x4134ecf0'
  field_count: 0
  is_sync: false
  ck_constraint: []
  on_replace: 'function: 0x4134ee58'
  temporary: false
  index: []
  is_local: false
  enabled: false
  name: space
  id: 512
- created
...

tarantool> box.execute("CREATE TABLE t1(a INT PRIMARY KEY, b INT, c INT);")
---
- null
- 'Duplicate key exists in unique index "primary" in space "_space" with old tuple
  - [512, 1, "space", "memtx", 0, {}, [{"name": "id", "type": "unsigned"}]] and new
  tuple - [512, 1, "T1", "memtx", 3, {}, [{"name": "A", "type": "integer", "is_nullable":
  false, "nullable_action": "abort"}, {"name": "B", "type": "integer", "is_nullable":
  true, "nullable_action": "none"}, {"name": "C", "type": "integer", "is_nullable":
  true, "nullable_action": "none"}]]'
...

```

**Actual behavior**

Failed to autogenerate a valid id for a space on `CREATE TABLE`.

**Expected behavior**

`CREATE TABLE` succeeds disregarding of what ids are used for existing spaces.

drewdzzz:
Currently, `_schema.max_id` is used to generate sequentially growing
space ids. The main drawback of this approach is that generated space id
can be not unique if one mixes implicit and explicit space ids. Let's use
actual maximal space id to generate a new one, and scan for free id if
overflow happened.

Also, the patch removes `max_id` from bootstrap snapshot, adds upgrade
and downgrade scripts.

Closes https://github.com/tarantool/tarantool/issues/8036
Closes https://github.com/tarantool/tarantool/issues/5997


src/box/lua/schema.lua
```diff
@@ -812,18 +812,25 @@ box.schema.space.create = function(name, options)
...
                 id = box.schema.SYSTEM_ID_MAX
             end
-            max_id = _schema:insert{'max_id', id + 1}
+            max_id = _schema:replace{'max_id', id + 1}
         end
         id = max_id.value
+    else
+        local max_id_tup = _schema:get{'max_id'}
+        if max_id_tup ~= nil then
+            _schema:update({'max_id'}, {{'=', 2, math.max(max_id_tup[2], id)}})
```
locker, line 830: This worked before, now it doesn't:

```
tarantool> box.schema.space.create('test', {id = 1ULL})
---
- error: 'builtin/box/schema.lua:830: bad argument #2 to ''max'' (number expected,
    got cdata)'
...
```

src/box/lua/schema.lua
```diff
@@ -812,18 +812,25 @@ box.schema.space.create = function(name, options)
...
         end
     end
+    local _schema = box.space._schema
     local id = options.id
     if not id then
-        local _schema = box.space._schema
         local max_id = _schema:update({'max_id'}, {{'+', 2, 1}})
-        if max_id == nil then
-            id = _space.index.primary:max().id
+        id = _space.index.primary:max().id
```
locker, line 819: This won't work if there's a space with the max id:

```
tarantool> box.schema.space.create('test', {id = 2147483647})
---
- engine: memtx
  before_replace: 'function: 0x416a02a0'
  field_count: 0
  is_sync: false
  on_replace: 'function: 0x416a0278'
  temporary: false
  index: []
  is_local: false
  enabled: false
  name: test
  id: 2147483647
- created
...

tarantool> box.schema.space.create('test2')
---
- error: 'Failed to create space ''test2'': space id is too big'
...
```

drewdzzz: Reworked

src/box/lua/schema.lua
drewdzzz: It's OK now, using this way of creation in the test.

test/box-luatest/gh_8036_create_space_with_explicit_and_implicit_id_test.lua
```diff
@@ -0,0 +1,63 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group('Create space with explicit and implicit id', {
+    {sql = false},
+    {sql = true}
```
locker, line 6: Parameterization by `sql = true/false` looks weird. Better duplicate the code - the test isn't that big.

test/box-luatest/gh_8036_create_space_with_explicit_and_implicit_id_test.lua
```diff
@@ -0,0 +1,63 @@
...
+g.before_each(function(cg)
+    cg.server = server:new{
+        alias   = 'default',
+        box_cfg = {}
+    }
+    cg.server:start()
+end)
+
+g.after_each(function(cg)
+    cg.server:drop()
```
locker, line 18: Recreating the test server after each test case extends the test time. We should avoid it if possible. In this case, it'd be better to drop the test spaces instead.

test/box-luatest/gh_8036_create_space_with_explicit_and_implicit_id_test.lua
drewdzzz: Done

src/box/sql.c
```diff
@@ -868,21 +868,46 @@ tarantoolsqlIncrementMaxid(uint64_t *space_max_id)
...
+	char *tuple_end = mp_encode_uint(ops_eq + 5, *space_max_id);
+	memset(&request, 0, sizeof(request));
+	request.tuple = ops_eq;
+	request.tuple_end = tuple_end;
+	request.key = key;
+	request.key_end = key + sizeof(key);
+	request.type = IPROTO_UPDATE;
+	request.space_id = BOX_SCHEMA_ID;
+	if (box_process1(&request, &res) != 0 || res == NULL)
+		return -1;
```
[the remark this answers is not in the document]
drewdzzz: 1. `boxk` does not return result. We will need to read `max_id` manually, which requires read access to space `_schema` - it is not required now, and we decided not to ask for new privileges for space creation.
2. Done
3. Done

drewdzzz:
Added test cases for corner cases - there is no `max_is` in space `_schema` and when it gets overflowed.

src/box/lua/schema.lua
```diff
@@ -84,6 +84,9 @@ ffi.cdef[[
     box_txn_savepoint_t *
     box_txn_savepoint();
 
+    int
+    box_advance_max_id(void);
```
locker, line 88: This doesn't match the actual function signature.

src/box/box.cc
```diff
@@ -4662,6 +4662,72 @@ box_read_ffi_enable(void)
...
+	bool max_id_is_system = false;
+	if (max_id <= BOX_SYSTEM_ID_MAX) {
+		max_id = BOX_SYSTEM_ID_MAX + 1;
+		max_id_is_system = true;
+	}
+	uint32_t available_max_id = space_available_id(max_id);
+	/** Try again if overflowed. */
+	if (available_max_id > BOX_SPACE_MAX && !max_id_is_system) {
+		max_id = BOX_SYSTEM_ID_MAX + 1;
+		available_max_id = space_available_id(max_id);
```
locker, line 4718: AFAIU the function may fail to find a vacant id, but you never check that.

src/box/space_cache.h
```diff
@@ -97,6 +97,13 @@ space_by_id(uint32_t id);
 struct space *
 space_by_name(const char *name);
 
+/**
+ * Find minimal available id, which is it lower than current_id.
```
locker, line 101: I don't understand the comment.

src/box/space_cache.h
```diff
@@ -97,6 +97,13 @@ space_by_id(uint32_t id);
...
+ * If there is no available id, BOX_SPACE_MAX + 1 is returned.
+ */
+uint32_t
+space_available_id(uint32_t current_id);
```
locker, line 105: Rather bad name - it's not really clear what the function does. Also, it's named like it works with a space object while in fact it doesn't. Better change it to `space_cache_find_next_unused_id(min_id)`.

src/box/sql.c
```diff
@@ -853,36 +853,10 @@ tarantoolsqlIdxKeyCompare(struct BtCursor *cursor,
 int
 tarantoolsqlIncrementMaxid(uint64_t *space_max_id)
```
locker, line 854: This change makes the function name and the op code obsolete - the function doesn't simply increment the max id anymore, instead it generates a space id. Also, this change turns this function into a one-liner so better just inline it into the op code handler. Anyway, this change should be reviewed by @ImeevMA.

src/box/box.cc
drewdzzz: The logic is if we overflowed the second time - overflowed value is returned and the space creation will fail with an error. I did so because the error is more informative - in this function we don't know the new space name.

src/box/box.h
```diff
@@ -630,6 +630,12 @@ box_process_rw(struct request *request, struct space *space,
 int
 boxk(int type, uint32_t space_id, const char *format, ...);
 
+/**
+ * Increase schema max_id to the minimum available one.
+ */
+API_EXPORT int64_t
+box_advance_max_id(void);
```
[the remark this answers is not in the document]
drewdzzz: 1. The function is used in ffi - it used much easier to return value that have an output argument
2. `INT64_MAX` is a kind of infinity
3. `box_generate_space_id` does not reflect the fact `_schema.max_id` is changed. Should we reflect it at all?

src/box/space_cache.h
drewdzzz: Done

src/box/space_cache.h
drewdzzz: Me too. Rewrote it.

src/box/box.h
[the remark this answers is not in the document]
locker: 1. IMO it's better to use a Lua C wrapper then.
3. I don't think that we should reflect this fact - it's implementation detail. You could say a word about it in a comment though.

src/box/box.cc
```diff
@@ -4662,6 +4662,72 @@ box_read_ffi_enable(void)
...
+	/*
+	 * Use the first available id greater than BOX_SYSTEM_ID_MAX if max_id
+	 * is overflowed.
+	 */
+	if (max_id > BOX_SPACE_MAX)
+		max_id = BOX_SYSTEM_ID_MAX;
+	bool max_id_is_system = false;
+	if (max_id <= BOX_SYSTEM_ID_MAX) {
+		max_id = BOX_SYSTEM_ID_MAX + 1;
+		max_id_is_system = true;
```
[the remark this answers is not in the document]
locker: Let's update `max_id` unconditionally for the sake of simplicity?

src/box/box.cc
locker: What kind of error? Is it tested? IMO an explicit error indicating that we ran out of space ids would be better.

src/box/box.h
```diff
@@ -630,6 +630,12 @@ box_process_rw(struct request *request, struct space *space,
 int
 boxk(int type, uint32_t space_id, const char *format, ...);
 
+/**
+ * Increase schema max_id to the minimum available one.
+ */
+API_EXPORT int64_t
```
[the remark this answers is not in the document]
drewdzzz: Thought it is needed for FFI - removed.

src/box/lua/schema.lua
drewdzzz: Removed it (use LuaC now)

src/box/box.h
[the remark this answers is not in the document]
drewdzzz: Done (for all points)

src/box/box.cc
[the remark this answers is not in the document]
drewdzzz: Done

src/box/box.cc
drewdzzz: Discussed f2f, use panic here

src/box/sql.c
drewdzzz: Dropped the function.

drewdzzz:
I've added writability and recovery checks to `box_generate_space_id` (relevant check for `_schema` space from `box_process1`) - and annoying relocation test `gh_5568_read_only_reason_test.lua` stopped failing in release.

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
 		box_read_ffi_is_disabled = false;
 }
 
+int
+box_generate_space_id(uint64_t *new_space_id)
```
locker, line 5035: Overall, the function code looks confusing. I'd expect `space_cache_find_next_unused_id(id)` to return an id *strictly greater* than the given one (because of *next* in its name) while in fact it returns a value >= id. I assume this is because we use `boxk` to update and get the max id. I think we could rewrite the code like this to make it look more straightforward:

```C
struct tuple *tuple;
if (box_index_get(BOX_SPACE_MAX, 0, key, key_end, &tuple) != 0)
	return -1;
uint32_t max_id = 0;
if (tuple != NULL && tuple_field_u32(tuple, 1, &max_id) != 0)
	return -1;
if (max_id > BOX_SPACE_MAX || max_id < BOX_SYSTEM_ID_MAX)
	max_id = BOX_SYSTEM_ID_MAX;
*new_space_id = space_cache_find_next_next_unused_id(max_id);
if (*new_space_id > BOX_SPACE_MAX) {
	*new_space_id = space_cache_find_next_next_unused_id(
					BOX_SYSTEM_ID_MAX);
	if (*new_space_id > BOX_SPACE_MAX)
		panic("...");
}
if (boxk(IPROTO_REPLACE, BOX_SCHEMA_ID, NULL,
	 "[%s%lu]", "max_id", (unsigned long)*new_space_id) != 0)
	return -1;
return 0;
```

To avoid permission errors, we could change credentials to admin's for `box_index_get()`, like we do here:

https://github.com/tarantool/tarantool/blob/56a43147444141ada97d5f03faa0dd218dbbcc89/src/box/box.cc#L3404-L3410

src/box/sql/vdbe.c
```diff
@@ -4340,7 +4340,7 @@ case OP_IncMaxid: {
 	assert(pOp->p1 > 0);
 	pOut = vdbe_prepare_null_out(p, pOp->p1);
 	uint64_t u;
-	if (tarantoolsqlIncrementMaxid(&u) != 0)
+	if (box_generate_space_id(&u) != 0)
```
locker, line 4343: I think you should also change the op code name, because `OP_IncMaxid` doesn't reflect what this operation actually does.

locker (changes requested):

src/box/box.cc
```diff
@@ -2945,17 +2945,18 @@ box_set_txn_isolation(void)
  * a variable-argument tuple described in format.
  *
  * @example: you want to insert 5 into space 1:
- * boxk(IPROTO_INSERT, 1, "[%u]", 5);
+ * boxk(IPROTO_INSERT, 1, NULL, "[%u]", 5);
```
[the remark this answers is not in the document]
drewdzzz: Outdated: I don't use boxk, so this change is removed.

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
...
 
+int
+box_generate_space_id(uint64_t *new_space_id)
+{
+	assert(new_space_id != NULL);
+	struct space *space_schema = space_cache_find(BOX_SCHEMA_ID);
+	if (space_schema == NULL)
+		return -1;
+	if (box_check_writable() != 0)
+		return -1;
```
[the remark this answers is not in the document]
drewdzzz: Removed it.

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
...
+	if (memtx_space_is_recovering(space_schema)) {
+		diag_set(ClientError, ER_UNSUPPORTED, "Snapshot recovery",
+			 "space id generation, use "
+			 "box.ctl.is_recovery_finished() "
+			 "to check that snapshot recovery was completed");
+		diag_log();
+		return -1;
```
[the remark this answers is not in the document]
drewdzzz: Removed it (instance is read_only until recovery is finished, so the check is redundant).

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
...
+	}
+	struct tuple *res = NULL;
+	uint64_t schema_max_id = 0;
+	if (boxk(IPROTO_UPDATE, BOX_SCHEMA_ID, &res, "[%s][[%s%d%d]]",
+		 "max_id", "+", 1, 1) != 0)
+		return -1;
+	if (res != NULL) {
+		if (tuple_field_u64(res, 1, &schema_max_id) != 0)
+			return -1;
+	}
```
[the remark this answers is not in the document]
drewdzzz: Done

src/box/lua/misc.cc
```diff
@@ -428,6 +428,19 @@ lbox_tuple_format_new(struct lua_State *L)
 	return lbox_push_tuple_format(L, format);
 }
 
+/**
+ * Generate space_id for non-system space. Updates _schema.max_id.
+ */
+static int
+lbox_generate_space_id(lua_State *L)
```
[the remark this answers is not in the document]
drewdzzz: Didn't noticed the section, thanks. Moved the function to misc section.

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
...
+	uint64_t max_id = schema_max_id;
+	if (max_id > BOX_SPACE_MAX || max_id <= BOX_SYSTEM_ID_MAX)
+		max_id = BOX_SYSTEM_ID_MAX + 1;
+	uint32_t available_max_id = space_cache_find_next_unused_id(max_id);
```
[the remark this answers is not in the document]
drewdzzz: Use `uint32_t` now.

src/box/box.cc
```diff
@@ -5030,6 +5031,58 @@ box_read_ffi_enable(void)
...
+		 * memory, and its pretty hard to test it, so let's just panic
+		 * if we've run out of ids.
+		 */
+		if (available_max_id > BOX_SPACE_MAX)
+			panic("Space id limit is reached");
+	}
+	*new_space_id = available_max_id;
+	/** Resfresh _schema.max_id if it is stale. */
+	if (schema_max_id != *new_space_id &&
+	    boxk(IPROTO_REPLACE, BOX_SCHEMA_ID, NULL, "[%s%llu]", "max_id",
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/box.cc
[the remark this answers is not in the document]
drewdzzz: I will never pass 0 because of `if` statement above. Anyway, the function is reworked, so the comment is outdated.

src/box/box.cc
drewdzzz: Reworked the function in a more straightforward manner.

src/box/box.cc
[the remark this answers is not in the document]
drewdzzz: Also, added a test for this case.

src/box/box.cc
drewdzzz: Also, updated changeling and commit message - I considered it necessary to note the weakening of access requirements.

ImeevMA (commented):
Hi! Thank you for the patch! I have a couple of questions about the issue, but since I haven't seen discussions on this subject, maybe I just don't know something.

src/box/box.cc
```diff
@@ -5030,6 +5030,50 @@ box_read_ffi_enable(void)
 		box_read_ffi_is_disabled = false;
 }
 
+int
+box_generate_space_id(uint32_t *new_space_id)
```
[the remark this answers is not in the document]
drewdzzz: Sorry, I don't really understand what you mean. Manipulations with space `_schema` are not enough - we need to scan `space_cache` for an available id.

src/box/sql/vdbe.c
```diff
@@ -4330,17 +4330,16 @@ case OP_Init: {          /* jump */
 	goto jump_to_p2;
 }
 
-/* Opcode: IncMaxid P1 * * * *
+/* Opcode: GenSpaceid P1 * * * *
```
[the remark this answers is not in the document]
drewdzzz: Added your synopsis:
<details>
<summary>diff</summary>

```diff
diff --git a/src/box/sql/vdbe.c b/src/box/sql/vdbe.c
index faecdcae8..611914ab0 100644
--- a/src/box/sql/vdbe.c
+++ b/src/box/sql/vdbe.c
@@ -4331,6 +4331,7 @@ case OP_Init: {          /* jump */
 }

 /* Opcode: GenSpaceid P1 * * * *
+ * Synopsis: r[P1]=new space ID
  *
  * Generate an id for space and store it in register specified by first
  * operand. It is system opcode and must be used only during DDL routine.
```
</details>

src/box/box.cc
```diff
@@ -5030,6 +5030,50 @@ box_read_ffi_enable(void)
...
+	fiber_set_user(fiber(), orig_credentials);
+	if (rc != 0)
+		return -1;
+	uint32_t max_id = 0;
+	if (res != NULL && tuple_field_u32(res, 1, &max_id) != 0)
+		return -1;
+	if (max_id > BOX_SPACE_MAX || max_id < BOX_SYSTEM_ID_MAX)
+		max_id = BOX_SYSTEM_ID_MAX;
+	*new_space_id = space_cache_find_next_unused_id(max_id);
+	/** Try again if overflowed. */
```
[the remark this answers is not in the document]
locker: It looks like `max_id` was first persisted in `_schema` because it was required by the Sophina engine: https://github.com/tarantool/tarantool/commit/dec8c277e80f65eb006c05192c34c271e020ed9e. The Sophia engine was renamed to Vinyl, and nowadays Vinyl doesn't depend on space ids growing monotonically. Let's try to drop `max_id` and see if the tests pass.

src/box/box.cc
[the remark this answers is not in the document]
ImeevMA: You are right, thanks.

src/box/box.cc
```diff
@@ -5031,6 +5031,43 @@ box_read_ffi_enable(void)
 		box_read_ffi_is_disabled = false;
 }
 
+int
+box_generate_space_id(uint32_t *new_space_id)
+{
+	assert(new_space_id != NULL);
+	static const char empty_key[] = {
```
locker, line 5038: Nit: There's no need to make the buffer static. Also, since this isn't a hot path, it's better to encode MsgPack explicitly instead of using hard-coded constants.

test/box-luatest/gh_8036_create_space_with_explicit_and_implicit_id_test.lua
```diff
@@ -0,0 +1,71 @@
...
+g.after_each(function()
+    g.server:exec(function()
+        box.space.SPACE:drop()
+        box.space.SPACE2:drop()
+    end)
+end)
+
+g.test_create_space_with_explicit_and_implicit_id = function()
+    g.server:exec(function()
+        local t = require('luatest')
```
locker, line 26: Nit: There's no need to `require('luatest')` in `exec`.

src/box/lua/misc.cc
```diff
@@ -213,6 +213,17 @@ port_msgpack_dump_lua(struct port *base, struct lua_State *L, bool is_flat)
...
 }
 
+/** Generate unique id for non-system space. */
+static int
+lbox_generate_space_id(lua_State *L)
+{
+	uint32_t ret = 0;
+	if (box_generate_space_id(&ret) != 0)
+		return luaT_error(L);
+	lua_pushinteger(L, ret);
```
locker, line 223: Nit: `lua_pushinteger()` works with 32-bit signed integers. Even though a space id shouldn't exceed `BOX_SPACE_MAX`, which equals `INT32_MAX`, it's still defined as `uimnt32_t` so it's better to use `lua_pushnumber()` instead.

changelogs/unreleased/gh_5997_get_rid_of_schema_max_id.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* Tarantool does not use `_schema.max_id` anymore, so this field is
+  dropped (gh-5997).
```
locker, line 4: `max_id` is [documented](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_schema/) so we need to file a ticket to update the doc. Please add a doc bot request.

src/box/lua/upgrade.lua
```diff
@@ -1408,6 +1420,7 @@ local handlers = {
     {version = mkversion(2, 10, 4), func = upgrade_to_2_10_4},
     {version = mkversion(2, 10, 5), func = upgrade_to_2_10_5},
     {version = mkversion(2, 11, 0), func = upgrade_to_2_11_0},
+    {version = mkversion(2, 11, 1), func = upgrade_to_2_11_1},
```
locker, line 1423: Please add a downgrade handler and a test as well.

locker (changes requested):

src/box/box.cc
drewdzzz: Encode it explicitly

src/box/lua/misc.cc
drewdzzz: Changed to `lua_pushnumber`

changelogs/unreleased/gh_5997_get_rid_of_schema_max_id.md
drewdzzz: Done

src/box/lua/upgrade.lua
drewdzzz: The absence of `max_id` is handled correctly but is not tested, so I wrote downgrade handler and test.

locker (approved):

locker:
@nshy Please review the upgrade/downgrade part.

nshy (approved):

ImeevMA (approved):
Hi! Thank you for the fixes! I see no problem with SQL code.

I still have one question: do we really want to push this change to 2.11? This is a change in behavior and might be worth pushing into 3.0.

PR #8101, opened 2022-12-28
sql: remove unused variables

ImeevMA:
This patch removes unused variables that were not caught by the compiler due to MAYBE_UNUSED or conversion to void.

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring


Gerold103 (approved):
Thanks for the patch!

locker (approved):

Issue #6620, opened 2021-11-18
Setup integration testing with CRUD module


ligurio:
CRUD [1] is one of the popular modules in Tarantool's ecosystem.
It would be nice to trigger tests for CRUD module with latest Tarantool version built from master branch. The same as we did for other modules and connectors, see #6056.

1. https://github.com/tarantool/crud

Pull requests:
* https://github.com/tarantool/crud/pull/335/
* https://github.com/tarantool/tarantool/pull/8056

slavakirichenko:
Before merge this PR, first we need to mergre [335 at trarntool/crud](https://github.com/tarantool/crud/pull/335/)

Issue #6619, opened 2021-11-18
Setup integration testing with DDL module


ligurio:
DDL [1] is one of the popular modules in Tarantool's ecosystem.
It would be nice to trigger tests for DDL module with latest Tarantool version built from master branch. The same as we did for other modules and connectors, see #6056.

1. https://github.com/tarantool/ddl

Pull requests:
* https://github.com/tarantool/ddl/pull/103
* https://github.com/tarantool/tarantool/pull/8056

PR #8105, opened 2022-12-29
ci: add integration testing for ddl and crud

NickVolynkin:
Add workflow for integration testing of ddl
and crud modules.

Resolves tarantool/tarantool#6619
Resolves tarantool/tarantool#6620


ylobankov (approved):
LGTM

PR #8109, opened 2022-12-30
lua: speedup uri.unescape()

Totktonada:
Use an array to convert a hex digit into a number instead of `isdigit()` and `tolower()`.

Eliminate the `isxdigit()` check by reusing the same array lookup.

Results on perf/lua/uri_escape_unescape.lua show 2.4x average boost (from 2.1x to 2.5x).

Noise in the results: 4% in baseline, 11% in the new version (the maximal difference between min and max results). The average linear/standard deviation is within 3%. 

Measured on i7-10510U, acpi-cpufreq driver (no intel_pstate), no TurboBoost, userspace governor, 1.80GHz, HyperThreading. The workload is bound to a single logical core using `taskset`.

### DIFF
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -3,7 +3,6 @@
  *
  * Copyright 2010-2021, Tarantool AUTHORS, please see AUTHORS file.
  */
-#include <ctype.h>
 #include "uri.h"
 #include "uri_parser.h"
 #include "trivia/util.h"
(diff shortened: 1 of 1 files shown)


Totktonada:
If you're interesting in reproducing the results I suggest to take care to [these](https://github.com/tarantool/go-tarantool/blob/13b9f8e6d89668fcaa6d399b42c07be634699062/CONTRIBUTING.md#recommendations-for-how-to-achieve-stable-results) recommendations for stabilizatoin of performance measurements.

igormunkin (approved):
@Totktonada, thanks for such a neat patch! LGTM.

PR #8112, opened 2023-01-09
ci: change macos runners labels

lastoCHka42:
Change runner label 'macos-11' to 'macos-11-self-hosted' and 'macos-12' to 'macos-12-self-hosted' to distinguish between self-hosted and GitHub-hosted runners. We want to use only self-hosted macOS runners because of test-run problems with python3.11 on GitHub-hosted macOS runners.

NO_DOC=ci
NO_TEST=ci
NO_CHANGELOG=ci

### DIFF
--- a/.github/workflows/osx_11.yml
+++ b/.github/workflows/osx_11.yml
@@ -39,7 +39,7 @@ jobs:
         ( github.event_name != 'pull_request' ||
           contains(github.event.pull_request.labels.*.name, 'full-ci') )
 
-    runs-on: macos-11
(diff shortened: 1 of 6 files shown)


NickVolynkin:
message: use new labels to distinguish between github-owned and self-hosted runners. These workflows should always run on self-hosted runners.

ylobankov (approved):
LGTM

NickVolynkin (approved):

ylobankov:

PR #8113, opened 2023-01-09
box: dynamically adjust xlog readahead size

### LINKED ISSUE #8108: Increase `XLOG_READ_AHEAD`
`xlog` and `snap` files are compressed, `xlog_cursor_next` reads them from HDD by ~50 KB chunks:

https://github.com/tarantool/tarantool/blob/182034dc979b3125f0ad66227997b58b400b6e2e/src/box/xlog.c#L1898-L1902

https://github.com/tarantool/tarantool/blob/182034dc979b3125f0ad66227997b58b400b6e2e/src/box/xlog.c#L1548

However `XLOG_READ_AHEAD` is only 16 KB, which doesn't make any sense.

https://github.com/tarantool/tarantool/blob/182034dc979b3125f0ad66227997b58b400b6e2e/src/box/xlog.c#L1557-L1558

Here are CPU usage graphs for Tarantool recovery from a 900 MB snapshot (see #5787 "trees with hints") from a HDD with 300 MB/sec read speed (buffered, uncached).

The first run (`# sync; echo 1 > /proc/sys/vm/drop_caches`) took 25 sec:

Screenshot 2022-12-29 at 21 26 43

The second run (with hot page cache) took 12 sec:

Screenshot 2022-12-29 at 21 27 26

The first run is obviously I/O bound, it should be up to 2x faster with a big enough `XLOG_READ_AHEAD`.

Gumix:
`XLOG_READ_AHEAD` defines the number of bytes added to the `count` argument
of the `pread` syscall, currently it equals to 16 KB. However xlog and snap
files are written by 128 KB chunks of data, which turn into ~80 KB chunks
after compression (in average, depending on the data), so the 16 KB read-
ahead doesn't make any sense.

According to performance experiments, 8 MB readahead gives the best results
for large files. However, 8 MB read buffers would increase memory
consumption of the replication relays, which usually read small portions of
data and does not need such a big buffers. For this reason, dynamically-
sized read buffer is implemented by this patch. The minimum readahead is
now 128 KB, and the maximum is 8 MB. As a result, the recovery time of a
900 MB snapshot decreased from 25 sec to 13 sec (when reading from HDD).
Performance of a recovery from SSD is not affected.

Closes #8108

**Do not merge until tarantool/small#58 is merged!**

### DIFF
--- a/changelogs/unreleased/gh-8108-adjust-xlog_read_ahead.md
+++ b/changelogs/unreleased/gh-8108-adjust-xlog_read_ahead.md
@@ -1,0 +1,4 @@
+## feature/core
+
+* Reduced recovery time from a snapshot by up to 2x on the systems with a hard
+  disk drive (gh-8108).
--- a/src/box/xlog.c
+++ b/src/box/xlog.c
@@ -1526,7 +1526,13 @@ xlog_close(struct xlog *l, bool reuse_fd)
 
 /* {{{ struct xlog_cursor */
 
-#define XLOG_READ_AHEAD		(1 << 14)
+enum {
+	/**
+	 * Min and max values for xlog_cursor::read_ahead.
+	 */
+	XLOG_READ_AHEAD_MIN = XLOG_TX_AUTOCOMMIT_THRESHOLD,
+	XLOG_READ_AHEAD_MAX = 8 * 1024 * 1024,
+};
 
 /**
  * Ensure that at least count bytes are in read buffer
@@ -1545,7 +1551,7 @@ xlog_cursor_ensure(struct xlog_cursor *cursor, size_t count)
 		return 1;
 
 	size_t to_load = count - ibuf_used(&cursor->rbuf);
-	to_load += XLOG_READ_AHEAD;
+	to_load += cursor->read_ahead;
 
 	void *dst = ibuf_reserve(&cursor->rbuf, to_load);
 	if (dst == NULL) {
@@ -1567,9 +1573,25 @@ xlog_cursor_ensure(struct xlog_cursor *cursor, size_t count)
 			 cursor->name);
 		return -1;
 	}
-	/* ibuf_reserve() has been called above, ibuf_alloc() must not fail */
 	assert((size_t)readen <= to_load);
+	/* ibuf_reserve() has been called above, ibuf_alloc() must not fail */
 	ibuf_alloc(&cursor->rbuf, readen);
+	/* Shrink the read buffer to reduce the memory consumption. */
+	if (cursor->need_rbuf_shrink) {
+		ibuf_shrink(&cursor->rbuf);
+		cursor->need_rbuf_shrink = false;
+	}
+	/*
+	 * Grow readahead size if the requested number of bytes was successfully
+	 * read, and decrease it to the minimum otherwise.
+	 */
+	if ((size_t)readen == to_load) {
+		if (cursor->read_ahead * 2 <= XLOG_READ_AHEAD_MAX)
+			cursor->read_ahead *= 2;
+	} else {
+		cursor->need_rbuf_shrink = true;
+		cursor->read_ahead = XLOG_READ_AHEAD_MIN;
+	}
 	cursor->read_offset += readen;
 	return ibuf_used(&cursor->rbuf) >= count ? 0: 1;
 }
@@ -1982,6 +2004,7 @@ xlog_cursor_openfd(struct xlog_cursor *i, int fd, const char *name)
 {
 	memset(i, 0, sizeof(*i));
 	i->fd = fd;
+	i->read_ahead = XLOG_READ_AHEAD_MIN;
 	ibuf_create(&i->rbuf, &cord()->slabc,
 		    XLOG_TX_AUTOCOMMIT_THRESHOLD << 1);
 
--- a/test/unit/xlog.c
+++ b/test/unit/xlog.c
@@ -1,0 +1,161 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#define UNIT_TAP_COMPATIBLE 1
+#include "unit.h"
+#include "xlog.h"
+#include "xrow.h"
+#include "crc32.h"
+#include "random.h"
+#include "memory.h"
+#include "iproto_constants.h"
+
+/**
+ * Keep in sync with src/box/xlog.c!
+ */
+enum {
+	XLOG_READ_AHEAD_MIN = 128 * 1024,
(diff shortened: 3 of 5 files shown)


src/box/xlog.c
```diff
@@ -1526,7 +1526,7 @@ xlog_close(struct xlog *l, bool reuse_fd)
 
 /* {{{ struct xlog_cursor */
 
-#define XLOG_READ_AHEAD		(1 << 14)
+#define XLOG_READ_BUF_SIZE	(8 * 1024 * 1024)
```
locker, line 1529: xlog cursor is used not only for recovery, but also for replication so raising the read ahead size up to 8 MB may lead to memory consumption increase by up to 256 MB per instance with registered replicas and even more with anonymous replicas. I'm not sure we can afford that. Maybe, increase the buffer size only for recovery? @alyapunov WDYT?

Gumix: Fixed by a dynamically-sized read buffer.

changelogs/unreleased/gh-8108-adjust-xlog_read_ahead.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* Improved snapshot recovery time by up to 2x on the systems with a hard disk
```
locker, line 3: Nit: "Improved recovery time" sounds weird to me (I'm not sure if it's incorrect though). I'd say "Reduced recovery time".

test/unit/xlog.c
```diff
@@ -0,0 +1,183 @@
...
+};
+
+/**
+ * Create a temporary directory, initialize it as xdir, and create a new xlog.
+ */
+static void
+create_xlog(struct xlog *xlog, char *dirname)
+{
+	if (mkdtemp(dirname) == NULL)
+		panic("mkdtemp() failed");
```
locker, line 31: Nit: We have `fail` helper for unit tests - it should be more appropriate here. You could write:

```C
fail_if(mkdtemp(dirname) == NULL);
```

test/unit/xlog.c
```diff
@@ -0,0 +1,183 @@
...
+write_1k(struct xlog *xlog)
+{
+	/* Each uint64_t is encoded as 9 bytes + the array header. */
+	const size_t num_elements = 114;
+	const size_t data_size = num_elements * 9 + 3;
+
+	char data[data_size];
+	char *data_end = mp_encode_array(data, num_elements);
+	for (size_t i = 0; i < num_elements; i++)
+		data_end = mp_encode_uint(data_end, xoshiro_random());
```
locker, line 83: Nit: You could encode `MP_BIN` to avoid all the computations above.

test/unit/xlog.c
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * Test that the size of the read buffer dynamically increased while reading a
+ * large file, and shrunk when xlog is written/read by small portions of data.
+ */
+static void
+test_dynamic_sized_ibuf(void)
+{
+	plan(4);
```
locker, line 97: Nit: I think each test case should have `header()` and `footer()`. At least, the output looks better that way.

test/unit/xlog.c
```diff
@@ -0,0 +1,183 @@
...
+	ok(capacity >= start_capacity && capacity < 2 * start_capacity,
+	   "ibuf capacity decreased to [%zu..%zu)",
+	   start_capacity, 2 * start_capacity);
+
+	xlog_cursor_close(&cursor, false);
+	if (xlog_close(&xlog, false) < 0)
+		panic("xlog_close() failed");
+
+	char rmdir_cmd[sizeof("rm -rf ") + sizeof(dirname) + 1] = "rm -rf ";
+	strlcat(rmdir_cmd, dirname, sizeof(rmdir_cmd));
```
locker, line 165: `rmdir_cmd` is never used. Anyway, `rm -rf` looks dangerous. Let's explicitly remove only those files that we created with `unlink(2)` and `rmdir(2)`.

changelogs/unreleased/gh-8108-adjust-xlog_read_ahead.md
Gumix: Done

test/unit/xlog.c
Gumix: Done

test/unit/xlog.c
Gumix: Done

test/unit/xlog.c
Gumix: Done

test/unit/xlog.c
Gumix: Done

src/box/xlog.c
```diff
@@ -1571,6 +1577,18 @@ xlog_cursor_ensure(struct xlog_cursor *cursor, size_t count)
...
 	cursor->read_offset += readen;
+	/*
+	 * Grow readahead size if the requested number of bytes was successfully
+	 * read, and lower it otherwise.
+	 */
+	if ((size_t)readen == to_load) {
+		if (cursor->read_ahead * 2 <= XLOG_READ_AHEAD_MAX)
+			cursor->read_ahead *= 2;
+	} else if (cursor->read_ahead / 2 >= XLOG_READ_AHEAD_MIN) {
+		cursor->read_ahead /= 2;
```
locker, line 1588: AFAIR we decided to set `read_ahead` to the min immediately if we read less than could. This would also be consistent with shrinking the bufffer.

Gumix: Changed to:
```c
{
        cursor->need_rbuf_shrink = true;
        cursor->read_ahead = XLOG_READ_AHEAD_MIN;
}
```

locker (approved):

alyapunov (approved):

Gumix:
Updated the `src/lib/small` submodule and the unit test accordingly:
```diff
diff --git a/test/unit/xlog.c b/test/unit/xlog.c
index ee9d52971..ae4d920ef 100644
--- a/test/unit/xlog.c
+++ b/test/unit/xlog.c
@@ -131,14 +131,10 @@ test_dynamic_sized_ibuf(void)
                fail_if(row.lsn != prev_lsn + 1);
                prev_lsn = row.lsn;
        }
-       size_t capacity = ibuf_capacity(&cursor.rbuf);
-       size_t start_capacity = cursor.rbuf.start_capacity;

        is(cursor.read_ahead, XLOG_READ_AHEAD_MIN,
           "read_ahead decreased to %d", XLOG_READ_AHEAD_MIN);
-       ok(capacity >= start_capacity && capacity < 2 * start_capacity,
-          "ibuf capacity decreased to [%zu..%zu)",
-          start_capacity, 2 * start_capacity);
+       ok(ibuf_capacity(&cursor.rbuf) == 0, "ibuf capacity decreased to 0");

        xlog_cursor_close(&cursor, false);
        fail_if(xlog_close(&xlog, false) < 0);
```

Issue #4691, opened 2019-12-18
Unable to connect with net.box to 2.2 if box.schema.upgrade() has not been called.


Kasen:
Tarantool version:
2.2.1-137

OS version:
CentOS Linux release 7.6.1810
Bug description:
If I start 2.2 as a read-only replica of 1.10, then I can't connect to this node with error:
```
error: Space '277' does not exist
```

Steps to reproduce:
1. Start tarantool 1.10
2. Start the second instance of 1.10
3. Configure replication between these instances. The second instance must be read-only.
4. Restart the second instance with tarantool 2.2
5. Try to connect to the second instance with net.box

rosik:
@kyukhin could you please set a milestone and help with fixing that.

Totktonada:
It is the next incarnation of #4307. We missed the case when a remote instance is run on a new tarantool version, but don't upgraded yet and so lack of `_vcollation` view.

A trivial fix is possible: we can just ignore an error of fetching _vcollation. We may try to strength this way with some kind of verification whether the error is transient or not. Don't sure this way is right, but I don't see any other one for now.

Issue #4307, opened 2019-06-23
Unable to call a function on tarantool-1.10* from tarantool-2* with net.box


Totktonada:
Tarantool versions:

* Server: 1.10.3-73-g154d4f526 Linux-x86_64-Debug.
* Client (net.box): 2.2.0-423-gd3b2bcb66 Linux-x86_64-Debug.

OS version: Gentoo Linux.

## server.lua

```lua
#!/usr/bin/env tarantool

box.cfg({listen = 3301})

box.once('init', function()
    box.schema.user.grant('guest', 'read,write,execute,create,alter,drop',
                          'universe')
end)

function echo(...)
    return ...
end
```

## client.lua

```lua
#!/usr/bin/env tarantool

local net_box = require('net.box')
local json = require('json')

local connection = net_box.connect('localhost:3301')
local response = {connection:call('echo', {1, 2, 3})}
print(json.encode(response)) -- [1,2,3]
os.exit()
```

When both server and client are run with 2.2 or 1.10 the client prints `[1,2,3]` as expected. When the server is on 1.10, but the client is on 2.2 the output is `Space '277' does not exist`. 277 is ID of `_vcollation` system space.

Looks as the regression from of #3941.

cyrillos:
@romanhabibov This patch fixes similar problem for me.

PR #8119, opened 2023-01-09
test: verify net.box on a schema of version 2.1.3

### LINKED ISSUE #5482: test for 'net.box: fix fetching of schema of an old version'
Add a test for fix of #4691.

Totktonada:
See #4691 for the problem description. In brief: before the fix net.box expected _vcollation presence if tarantool's runtime on the server is new enough disregarding a server's schema version. The old schema version has no this view, so net.box was unable to connect. After #4691 net.box tolerates lack of the view.

The fix was pushed without a test (see [here][1]) and now I want to pay this debt.

[1]: https://lists.tarantool.org/pipermail/tarantool-patches/2020-April/016207.html

Follows up #4307
Follows up #4691
Fixes #5482


Totktonada:
@Gerold103 I remember that we had an [agreement](https://lists.tarantool.org/pipermail/tarantool-patches/2020-April/016177.html) about data directory names, but I see that new tests are added with `test/box-luatest/gh_xxxx_data` directories. I followed this new approach.

Unlike my initial test the new one has all the needed data written inside the snapshot.

test/box-luatest/gh_4691_data/fill.lua
```diff
@@ -0,0 +1,61 @@
...
+local function sourcefile()
+    local info = debug.getinfo(3, 'S')
+    local source = info and info.source
+    return source and source:startswith('@') and source:sub(2) or nil
+end
+
+local function sourcedir()
+    local source = sourcefile()
+    return fio.abspath(source and source:match('(.*)/') or '.')
+end
```
Gerold103, line 16: Nit: this seems excess. Below anyway you do
```Lua
local snaps = fio.glob('*.snap')
assert(#snaps == 1)
```
which assumes the work dir == WAL dir. Hence you could also just drop `workdir` arg and this code.

test/box-luatest/gh_4691_data/fill.lua
```diff
@@ -0,0 +1,61 @@
...
+
+-- }}} Helpers to determine the script directory
+
+local destdir = sourcedir()
+
+box.cfg({
+    -- Place the resulting snapshot near to this fill.lua script.
+    work_dir = destdir,
+    -- We need only the last snapshot.
+    checkpoint_count = 1,
```
Gerold103, line 26: Nit: is it needed really? This code is executed like once per universe existence to care about unnecessary params. Which, on the other hand, might confuse into thinking that they are actually a pare of the test.

test/box-luatest/gh_4691_data/fill.lua
```diff
@@ -0,0 +1,61 @@
...
+-- Ensure that the snapshot is generated with certain schema
+-- version.
+--
+-- The gh-4691 test needs the schema of the 2.1.3 version.
+local schema_version_tuple = box.space._schema:get({'version'})
+assert(schema_version_tuple ~= nil)
+assert(schema_version_tuple[1] == 'version')
+assert(schema_version_tuple[2] == 2)
+assert(schema_version_tuple[3] == 1)
+assert(schema_version_tuple[4] == 3)
```
Gerold103, line 43: Somewhy I thought the idea is to checkout target (in this case - 2.1.3) source code, build it, make the snap, and use it as is. Not try to simulate the old schema on master tarantool. But if it covers the case, then alright.

Btw, you will still have `_vcollation` here, right? Which means netbox would connect fine regardless of the schema version and the original bug is not covered then. Or is it?

test/box-luatest/gh_4691_data/fill.lua
```diff
@@ -0,0 +1,61 @@
...
+    parts = {
+        {field = 1, type = 'string', collation = 'unicode_ci'}
+    },
+})
+
+-- Bump the snapshot.
+box.snapshot()
+local snaps = fio.glob('*.snap')
+assert(#snaps == 1)
+print('Created ' .. fio.abspath(snaps[1]))
```
Gerold103, line 59: Nit: isn't it printed in the logs anyway which are by default going to the standard output?

test/box-luatest/gh_4691_net_box_connect_schema_2_1_3_test.lua
```diff
@@ -0,0 +1,60 @@
...
+    -- presence is required by the client, this call will fail
+    -- after 60 seconds of attempts to connect.
+    g.schema_2_1_3:wait_until_ready()
+    local connection = g.schema_2_1_3.net_box
+
+    -- Just is case: the connection is alive.
+    t.assert(connection:is_connected())
+
+    -- Just in case: the instance has an old schema version, which
+    -- has no _vcollation view.
```
Gerold103, line 50: Why aren't you testing that indeed `_vcollation` is missing?

Gerold103 (commented):
Thanks for the patch!

test/box-luatest/gh_4691_data/fill.lua
Totktonada: It seems, I missed your point. I want to generate the snapshot in the directory, where the script resides. I don't want to manually copy the snapshot from a current directory.

It is opinionated, but I like that the script doing all the needed work itself. Not some random part of the work, while leaving some other random part of the work to a human. That's disgusting. Let a machine doing that.

test/box-luatest/gh_4691_data/fill.lua
Totktonada: My goal is to make the script self-sufficient. I would like to eliminate a manual work about moving/removing snapshots.

There is the comment, why the option is needed. It says nothing about necessarity of the option for the test.

There is the 'Add data necessary for the gh-4691 test' section below. It looks quite obvious that it contains the part, which relates to the test (not some automation).

Of course, it is run once. However if we provide a script it is better to let it automate all the steps and hold all the contracts with the test. Otherwise I would prefer `fill.md` with instructions for a human how to generate the snapshot manually, where to place it and so on, because an interation with a human is anyway needed.

I would also highlight that the script may be copied with replacing of the workload: it'll save some human time. If we'll copy it several times, maybe it'll worth to extract common parts into a tool. However I wouldn't bother now.

test/box-luatest/gh_4691_data/fill.lua
Totktonada: It doesn't try to simulate the old schema on the new master. Those assertions ensure that the script is run using an older version of tarantool.

There is no `_vcollation` on given schema version. So it is verified transitively.

Regarding automatic checkout+build: it is a bit fragile. We need to backport [two patches](https://github.com/tarantool/tarantool/wiki/User:Totktonada#build-old-tarantool-version) (see 2.5.1 instruction) to build tarantool 2.1.3 on a modern OS. In a future the set of the patches may be larger. Moreover, a build command may vary from OS to OS. This is not what I would like to automate.

However this work is done in [tt install tarantool](https://github.com/tarantool/tt/issues/91). If we'll want to automate it I would reuse an existing solution.

test/box-luatest/gh_4691_data/fill.lua
Totktonada: Almost so. `.inprogress` filename is in the logs. And it has no full path, it is relative to the workdir. I found it convenient to have a path on the terminal to copy-paste it to, say, `tarantoolctl cat /path/to/snap --show-system`.

test/box-luatest/gh_4691_net_box_connect_schema_2_1_3_test.lua
Totktonada: Given schema version doesn't provide the `_vcollation` view. So it is verified transitively. Since the check is not actually part of the test, but rather re-verification of correctness of the necessary files, I would minimize amount of such checks.

test/box-luatest/gh_4691_data/fill.lua
Gerold103: I somewhy thought that `workdir` won't change workdir of the whole process, all is fine.

test/box-luatest/gh_4691_data/fill.lua
Gerold103: > why the option is needed.

This comment just narrates the next line, which isn't too helpful. It doesn't say why it is necessary really. What would change if I delete it, if you make just one snapshot anyway. I checked it myself, apparently the zero snapshot remains without this option, which I didn't know, and which could be in the comment like `otherwise 0 snap remains`. I mean, until I checked it manually, this option looked as useless as trying to set `memtx_memory` - it will work of course, but wouldn't change anything.

> Otherwise I would prefer fill.md with instructions

It isn't such a complicated place ... There are 3 more folders which just contain binary files and people were able to create them without instructions. Besides, when instructions are made and nobody controls their execution all the time, we get what we get - no `fill.lua` in other dirs, naming and structure is new and random and each time someone wants an upgrade test - one way in `box-luatest/`, another in `sql/`, third in `sql-luatest/` (it has `fill.lua`), fourth in `vinyl/` and `xlog/` (ticket-id-folder inside version folder). `xlog/` has an `.md` file with instructions but nobody cares, frankly.

> I would also highlight that the script may be copied with replacing of the workload

Hopefully. This one looks like a good basis. However I would bet on a next upgrade test we will see from someone a fifth way of doing it instead :)

test/box-luatest/gh_4691_data/fill.lua
Gerold103: I wasn't attentive enough, somewhy thought these lines were setting the schema, not checking it.

test/box-luatest/gh_4691_data/fill.lua
Gerold103: > Almost so. .inprogress filename is in the logs. 

Weird, indeed. I was sure we log the final name, but no.

Gerold103 (approved):
Thanks for the patch!

test/box-luatest/gh_4691_data/fill.lua
Totktonada: >  I checked it myself, apparently the zero snapshot remains without this option, which I didn't know, and which could be in the comment like `otherwise 0 snap remains`.

The default is two snapshots, not zero. I guess you missed some error.

I wrote what will going on without the option.

```diff
diff --git a/test/box-luatest/gh_4691_data/fill.lua b/test/box-luatest/gh_4691_data/fill.lua
index b828a3c13..3a876ce1c 100755
--- a/test/box-luatest/gh_4691_data/fill.lua
+++ b/test/box-luatest/gh_4691_data/fill.lua
@@ -23,6 +23,9 @@ box.cfg({
     -- Place the resulting snapshot near to this fill.lua script.
     work_dir = destdir,
     -- We need only the last snapshot.
+    --
+    -- The option has default value 2, so if we omit it, then two
+    -- snapshots will be stored in the directory.
     checkpoint_count = 1,
     -- Our goal is to generate a snapshot and we don't need WALs.
     wal_mode = 'none',
```

Redarding the rest of the discussion: I understood you this way: you don't mind this patch as a proof of concept of given approach, but you want to coalesce all the ways into one in a future. Correct?

Gerold103: > The default is two snapshots, not zero. I guess you missed some error

It was a typo, I meant `2 snaps`, sorry.

> Correct?

Yes. But I also understand it is hardly possible.

alyapunov (approved):
Great test, LGTM.

Totktonada:
Rebased on top of current `master` just in case.

PR #8128, opened 2023-01-10
console: make prompt bookkeeping thread-safe

### LINKED ISSUE #8124: Crash when logging to interactive console from multiple IPROTO threads
**Bug description**

tarabrt crash artifacts: https://drive.google.com/file/d/1nrnXr2fDgRN3k-PvLL-OEc-mtVceKa_j/view?usp=sharing

Backtrace:
Screen Shot 2023-01-10 at 15 57 06

**Steps to reproduce**

1. Limit the number of file descriptors;
2. Run a Tarantool instance in interactive console mode;
3. Open a lot of connections to Tarantool instance, until the file descriptors are exhausted (e.g., using go connector or k6);
4. Observe `accept, called on fd 28, aka 0.0.0.0:8080: Too many open files` messages spammed to console (sic!) and, finally, the crash.

### LINKED ISSUE #8136: Hide/show prompt feature does not handle errors thrown from `print `
Consider the following snippet:
```lua
require('fiber').create(function() while true do pcall(function() print(setmetatable({}, {__tostring = error})) end) require('fiber').sleep(1) end end)
```

This code constantly erases the prompt and, AFAIC, the memory allocated in `console_hide_prompt` is also leaked, since `after_cb` never gets called:
https://github.com/tarantool/tarantool/blob/a82c2462897f56d5073ca2b00467ddb0b2293b55/src/lua/print.lua#L8-L14

https://github.com/tarantool/tarantool/blob/a82c2462897f56d5073ca2b00467ddb0b2293b55/src/box/lua/console.c#L340-L366

CuriousGeorgiy:
Closes #8136
Closes #8124

### DIFF
--- a/src/box/lua/console.c
+++ b/src/box/lua/console.c
@@ -274,6 +274,9 @@ console_sigint_handler(ev_loop *loop, struct ev_signal *w, int revents)
 /*
  * The idea is borrowed from
  * https://metacpan.org/dist/AnyEvent-ReadLine-Gnu/source/Gnu.pm
+ *
+ * Since this feature is not thread-safe, it will work only when logging occurs
+ * from main (transaction) thread.
  */
 
 static char *saved_prompt = NULL;
@@ -311,7 +314,7 @@ console_can_hide_show_prompt(void)
 static void
 console_hide_prompt(void)
 {
-	if (!console_can_hide_show_prompt())
+	if (!console_can_hide_show_prompt() || !cord_is_main())
 		return;
 
 	if (rl_prompt == NULL) {
@@ -343,7 +346,7 @@ console_hide_prompt(void)
 static void
 console_show_prompt(void)
 {
-	if (!console_can_hide_show_prompt())
+	if (!console_can_hide_show_prompt() || !cord_is_main())
 		return;
 
 	rl_set_prompt(saved_prompt);
--- a/src/lua/print.lua
+++ b/src/lua/print.lua
@@ -8,9 +8,12 @@ _G.print = function(...)
     if M.before_cb ~= nil then
         M.before_cb()
     end
-    M.raw_print(...)
+    local ok, err = pcall(M.raw_print, ...)
     if M.after_cb ~= nil then
         M.after_cb()
+    end
+    if not ok then
+        error(err)
     end
 end
 
--- a/test/app-luatest/gh_8136_fix_print_wrapper_not_handling_errors_thrown_from_print_test.lua
+++ b/test/app-luatest/gh_8136_fix_print_wrapper_not_handling_errors_thrown_from_print_test.lua
@@ -1,0 +1,32 @@
+local it = require('test.interactive_tarantool')
+
+local t = require('luatest')
+local g = t.group()
+
+-- Checks that `print` throwing an error is handled correctly.
+g.test_basic_print_with_exception = function()
+    local child = it.new({args = {'-l', 'fiber'}})
+
+    child:execute_command([[
+        _ = fiber.create(function()
+            fiber.name('print_flood', {truncate = true})
+            while true do
+                pcall(function()
+                          print(setmetatable({}, {__tostring = error}))
+                      end)
+                print('flood')
+                fiber.sleep(0.01)
+            end
+        end)
+    ]])
+    child:assert_empty_response()
+
+    local exp_line = it.PROMPT .. it.CR .. it.ERASE_IN_LINE ..
+                     it.PROMPT .. it.CR .. it.ERASE_IN_LINE ..
+                     'flood'
+    for _ = 1, 10 do
+        child:assert_line(exp_line)
+    end
+
+    child:close()
+end


Totktonada:
Wow, weird and interesting.

CuriousGeorgiy:
FYI: I tested the patch against the setup, and now it doesn't crash.

Actually, AFAIC, the setup can be simplified: the only requirement is big number of connections, and errors in IPROTO thread (i.e., sendinging invalid IPROTO packets) — not sure if this can be done using `net.box`, but this can surely be done using the golang connector.

If you wish to debug the original setup, DM me, and I will provide you access to the development servers and instruct you on the reproduction steps.

src/box/lua/console.c
```diff
@@ -314,6 +320,8 @@ console_hide_prompt(void)
 	if (!console_can_hide_show_prompt())
 		return;
 
+	tt_pthread_mutex_lock(&prompt_mutex);
```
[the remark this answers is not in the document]
Totktonada: If my guess is right, I see two possible solutions:

* Make a call to `raw_print` (src/lua/print.lua) protected (use `pcall()`).
* Track whether prompt is hidden and skip hiding in the case. The same for show.

Don't know what is better.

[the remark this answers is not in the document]
CuriousGeorgiy: You are right, I missed the second usage context of these functions. 

I guess we should wrap the ones used in the logger with synchronization then?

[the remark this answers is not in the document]
CuriousGeorgiy: AFAIC, the state use describe is impossible: when by the time a console line is evaluated, the prompt callbacks are reset:
https://github.com/tarantool/tarantool/blob/6336dfa1d2eef2cdfc914540162140cfea5e9054/src/box/lua/console.c#L566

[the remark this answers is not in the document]
CuriousGeorgiy: Actually, the original patch was not safe to errors thrown from `print`, e.g., this code would constantly erase the prompt:
```lua
require('fiber').create(function() while true do pcall(function() print(setmetatable({}, {__tostring = error})) end) require('fiber').sleep(1) end end)
```

I suggest we go for the first approach, file another ticket regarding this problem, and add a test for it.

Gerold103 (approved):
If this only affects the console, then I don't see any problems with using mutexes in any way. Also I don't understand the point about `errors or prints in IPROTO threads` in the latest comment above. If as `print` is meant Lua `print()`, then it is impossible to do from any thread except the main one. We don't have Lua runtime in any other thread. It doesn't matter which client you use to connect to tarantool.

src/box/lua/console.c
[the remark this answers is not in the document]
Totktonada: The solution with `pcall()` around `raw_print()` looks okay.

Totktonada (approved):

src/box/lua/console.c
locker, line 323: AFAICS readline doesn't use any mutexes internally, e.g. see [`rl_replace_line`](http://git.savannah.gnu.org/cgit/readline.git/tree/text.c#n212). Doesn't it mean that there may be a race between the tx thread calling an `rl_*` function while processing a console event and another thread calling another `rl_*` function from this callback?

CuriousGeorgiy: https://lists.gnu.org/archive/html/bug-readline/2006-11/msg00014.html

So, you suggest wrapping all work with readline with this mutex?

locker: I'd rather suggest reverting the original patch, because TBCH I don't quite understand the benefit of it. Logs should be written to a file/socket, not to stderr. As for Lua `print`, we should probably just redirect `print` output to a separate file/socket. Also, AFAIR there was an idea to override `print` to be handled properly by a console client, both local and remote.

Wrapping a call to every single `rl_*` function would look ugly IMO, besides we'd take and release the mutex multiple times per each console event. We could probably protect the whole `lbox_console_readline` with the mutex... Anyhow, this would be an anti-pattern, because we'd protect code sections with a mutex, not data. It's easy to overlook a code section, which should be protected. Even in this commit, we protect code instead of data, which doesn't look good. It'd probably be best if we followed the "Tarantool way" and, instead of using mutexes, sent an event to the tx thread.

CuriousGeorgiy: @Totktonada RFC

Totktonada: Personally I meet the problem while playing with https://github.com/tarantool/crud/blob/master/doc/playground.lua (try to run it and type some commands). It is not a production environment, where we have logs redirected to syslog/journald/file, but a developer playground, where we want all the information present on a single terminal. It would be nice to don't mix prompt/stdout/stderr if it is simple to achieve. However:

* The fix is partial, because it doesn't handle the incremental search (Ctrl+R).
* Even this partial fix is not really simple, rather hacky.
* This code is now known to lead to problems (which you're fixing here) and it lets me suspect that there may be more hidden problems.
* It becomes even more complex with the fixes (disregarding of a way to fix).

In brief: I disagree that there are no benefits, but I agree that they're not so large comparing to risks and maintenance costs.

I would also ask @Mons as a person who initially highlighted the problem.

Totktonada: IOW, I'm okay with attempt to fix it, but I'm also okay if we'll decide to don't support it.

(TBH, I'm wonder why terminal emulators have no API for a status line/user input/prompt and so on. It would eliminate a need to hack it around in readline; at least for some terminals -- other ones would follow it later. The output and prompt/input/status would be completely separate.)

changelogs/unreleased/gh-8136-fix-print-wrapper-not-handling-errors-thrown-from-print.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/lua
+
+* Fixed `print` function wrapper not handling errors thrown from `print`: this
+  could render the hide/show prompt feature (gh-7169) to be useless and leak
+  memory (gh-8136).
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed: removed the changelogs.

test/box-luatest/gh_8136_fix_print_wrapper_not_handling_errors_thrown_from_print_test.lua
```diff
@@ -0,0 +1,32 @@
+local it = require('test.interactive_tarantool')
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed: changed the test suite to app-luatest.

src/box/lua/console.c
```diff
@@ -276,6 +276,12 @@ console_sigint_handler(ev_loop *loop, struct ev_signal *w, int revents)
  * https://metacpan.org/dist/AnyEvent-ReadLine-Gnu/source/Gnu.pm
  */
 
+/**
+ * Multiple threads can attempt to write to stderr, thus the prompt bookkeeping
+ * needs to be synchronized.
+ */
+pthread_mutex_t prompt_mutex;
```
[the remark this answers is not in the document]
CuriousGeorgiy: Dropped this change.

src/box/lua/console.c
CuriousGeorgiy: Discussed with @locker, @alyapunov, @Mons, @igormunkin, @Totktonada: decided that fixing this feature to work for multiple thread is not acceptable right now — this feature will only work in main (transaction) thread.

locker (approved):

src/box/lua/console.c
```diff
@@ -311,7 +314,7 @@ console_can_hide_show_prompt(void)
 static void
 console_hide_prompt(void)
 {
-	if (!console_can_hide_show_prompt())
+	if (!console_can_hide_show_prompt() || !cord_is_main())
```
Totktonada, line 317: I looks important to do this check from inside the callback (as you implemented it). Otherwise one thread may set a callback and another will call it. I would leave a short note here. Just to don't step into the problem later.

Totktonada (approved):
Visually looks ok.

src/box/lua/console.c
CuriousGeorgiy: > Otherwise one thread may set a callback and another will call it. 

TBH, I don't understand what you mean. Could you please elaborate and provide an example?

Totktonada: My first question was 'why don't do this in `luaT_console_setup_write_cb()` and `luaT_console_cleanup_write_cb()`?' But if we would go this way, we would get the situation described above.

CuriousGeorgiy: Now I get it... 

But AFAIC, only the main thread can set up callbacks? Actually, I think it would be better to check for this in the setup functions.

@locker RFC

Totktonada: > But AFAIC, only the main thread can set up callbacks?

I guess so. But when the callback is set from the main thread, another thread may call it. So I think that your current implementation is the right one.

CuriousGeorgiy: Yeah, I just wrote some nonsense, sorry. Tuesday evening… Well then I guess it’s pretty intuitive that the check needs to be done in the callback itself. If you don’t mind, I would skip the comment, as I can’t come with one, seems obvious at first glance.

Totktonada: Okay.

Issue #5029, opened 2020-05-27
Get rid of instance_uuid/replicaset_uuid in favor of unique strings


Mons:
Currently we have an option to set up instance_uuid/replicaset_uuid during start

Within such configurations these identifiers commonly made by hand and created as pure mapping from distinguishable instance name.

Example:
* Cluster on 256 shards with 3 nodes in each replicaset
* Name formed as `<name>_<XXX>_<YY>`, where XXX are numbers from 001 to 256 and YY are 01 to 03. Each instance is of course a member of replicaset `<name>_<XXX>`
* For every of that name there are strict mapping to instance UUID's: `1457A4CE-0<XXX>-00<YY>-0000-000000000000` and Every cluster name has strict mapping to `C1457312-0<XXX>-0000-0000-000000000000`

So, considering that case there is a question: why should we use uuid, when we may just save mapping `vclock-id => instance_name` instead of `vclock-id => uuid` into _cluster?

Common, human friendly names are shorter than 36 bytes of stringified uuid, they are distinguishable by human and it seems, that they could be used in that cases, when they are provided by initial configuration.

It is possible to make some limitations on name/cluster (for ex to 36 bytes) if it is required by some current implementations

tsafin:
Yup, in general, there is no much need for check validity of an instance uuid via `tt_uuid_from_string` unless it's unique string.

From the practical prospective we simply need to use something similar to FQDN names, and they are usually unique (iff DNS is correctly working at this network)

Mons:
It's not worth using DNS, since we may move some node from one host to another
It may change its fqdn while still have it's name (by role in some replicaset)

tsafin:
> It's not worth using DNS, since we may move some node from one host to another

That's why I said "something _similar_ to FQDN" :)

R-omk:
May be related https://github.com/tarantool/tarantool/issues/3805

PR #8159, opened 2023-01-13
box: rename cluster to replicaset where visible

Gerold103:
In a few places visible to users and in iproto naming the term "cluster" really means "replicaset". One of those places is a part of public API - box.iproto.key.CLUSTER_UUID - which is not yet released.

The commit renames "cluster" in those places as a preparation for introduction of actual "cluster", like a set of replicasets. It will start from introduction of cluster name in addition to replicaset uuid/name.

Part of #5029

NO_DOC=original was not released yet
NO_CHANGELOG=same
NO_TEST=refactoring

Sent this PR out of 5029 patchset to push it asap before we release `iproto.key.CLUSTER_UUID`.


src/box/alter.cc
```diff
@@ -3897,7 +3897,7 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
 			return -1;
 		REPLICASET_UUID = uu;
 		box_broadcast_id();
-		say_info("cluster uuid %s", tt_uuid_str(&uu));
+		say_info("replicaset uuid %s", tt_uuid_str(&uu));
```
locker, line 3900: What about `box.space._schema:get('cluster')`, `box.space._cluster`, `box.info.cluster`? Are there any plans to rename those as well? Otherwise, it looks inconsistent...

src/box/iproto_constants.c
```diff
@@ -82,7 +82,7 @@ const unsigned char iproto_key_type[IPROTO_KEY_MAX] =
 	/* 0x22 */	MP_STR, /* IPROTO_FUNCTION_NAME */
 	/* 0x23 */	MP_STR, /* IPROTO_USER_NAME */
 	/* 0x24 */	MP_STR, /* IPROTO_INSTANCE_UUID */
-	/* 0x25 */	MP_STR, /* IPROTO_CLUSTER_UUID */
+	/* 0x25 */	MP_STR, /* IPROTO_REPLICASET_UUID */
```
locker, line 85: `IPROTO_CLUSTER_UUID` is documented [here](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/keys/). We should probably add a doc bot request to update the name there.

src/lib/core/crash.c
```diff
@@ -323,7 +323,7 @@ crash_report_feedback_daemon(struct crash_info *cinfo)
 	tt_uuid_to_string(&INSTANCE_UUID, uuid_buf);
 	snprintf_safe("\"server_id\":\"%s\",", uuid_buf);
 	tt_uuid_to_string(&REPLICASET_UUID, uuid_buf);
-	snprintf_safe("\"cluster_id\":\"%s\",", uuid_buf);
+	snprintf_safe("\"replicaset_id\":\"%s\",", uuid_buf);
```
locker, line 326: I think this should match the value in the [feedback daemon](https://github.com/tarantool/tarantool/blob/1a678a5e88b0c0ba7565d0027bf819e5c47f8999/src/box/lua/feedback_daemon.lua#L110). Should we update it as well? Or maybe leave them both as is for now? After all, it's cluster_*id*, not cluster_*uuid*.

src/box/alter.cc
Gerold103: Those can not be renamed so simple. Safely we can only rename texts like logs and not yet released features like `box.iproto.key.CLUSTER_UUID`. For the rest there is a backward-compatible plan, which I make live in the rest of #5029 (you can revisit the plan in the instance/replicaset/cluster names RFC).

I am trying to submit this patch asap because once `CLUSTER_UUID` is released soon, we won't be able to safely rename it either.

src/box/iproto_constants.c
Gerold103: Ok, added a request.

src/lib/core/crash.c
Gerold103: You are right, I didn't notice that this was not just logs, it is a JSON request. Reverted this line.

locker (approved):

locker:
@Gerold103 The integration/cartridge test [fails](https://github.com/tarantool/tarantool/actions/runs/3937050223/jobs/6734605041). Looks like the test [scans](https://github.com/tarantool/cartridge/blob/c13701540037ea987126a408c4e996ba9561a830/test/integration/logger_test.lua#L123) Tarantool logs for "cluster uuid". Please either fix the cartridge test so that it accepts both "replicaset uuid" and "cluster uuid" or revert the change regarding logging.

src/lib/core/crash.c
Totktonada: It seems it is not a big problem to update it in https://github.com/tarantool/feedback (to accept both keys as a replicaset id). Let's file an issue if you're interested.

PR #8160, opened 2023-01-16
Make static-build cmake config reusable for EE

locker:
EE basically duplicates most of the code of the CE static-build cmake config. Let's make the CE cmake config reusable and include it into the EE config instead.

Needed for https://github.com/tarantool/tarantool-ee/issues/185

EE part: https://github.com/tarantool/tarantool-ee/pull/340


Gumix (approved):

locker:
Cherry-picked to 2.10.

Issue #7980, opened 2022-11-28
`unwinding error: unw_step failed with status: -6542` spammed on Mac


sergepetrenko:
Tarantool version: master, pulled from source today:
```
➜  sptnt git:(master) ✗ ./src/tarantool -v
Tarantool 2.11.0-entrypoint-723-ge4f1c891b
Target: Darwin-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=TRUE
Compiler: Clang-14.0.0.14000029
C_FLAGS: -fexceptions -funwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/Users/s.petrenko/Source/sptnt=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
CXX_FLAGS: -fexceptions -funwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/Users/s.petrenko/Source/sptnt=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror
```

Simply using tarantool produces the following output on almost any action:
```
2022-11-28 15:41:24.945 [89466] snapshot/101/main backtrace.c:81 E> unwinding error: unw_step failed with status: -6542
```

`-6542` seems to be
```c
UNW_EBADREG            = -6542,        /* bad register number */
```
according to [apple's libunwind headers](https://opensource.apple.com/source/libunwind/libunwind-35.3/include/libunwind.h.auto.html)

The issue seems to be uncovered by 19abfd2a39205de270836adda7f3a733476acaa4, since now backtrace is taken on each `fiber_gc()` attempt. Although the mentioned patch didn't change anything about how libunwind is used on mac.

<details><summary>beware of huge output</summary>

```
Tarantool 2.11.0-entrypoint-723-ge4f1c891b
type 'help' for interactive help
tarantool> box.cfg{}
'strip_core' is set but unsupported
2022-11-28 15:41:24.869 [89466] main/103/interactive I> Tarantool 2.11.0-entrypoint-723-ge4f1c891b
2022-11-28 15:41:24.870 [89466] main/103/interactive I> log level 5
[...]

PR #7994, opened 2022-12-01
box: sending arbitrary IPROTO packets

CuriousGeorgiy:
Closes #7897

PR #8166, opened 2023-01-17
core: silence libunwind errors

sergepetrenko:
Every libunwind error during backtrace collection is reported with `say_error`. Since commit 19abfd2a3920 ("misc: get rid of fiber_gc") backtraces are collected on each fiber gc allocation, of which there are plenty.

For some reason (https://github.com/tarantool/tarantool/issues/7980) each unw_step fails on mac, and an error is spammed to instance logs, even though the backtrace is actually collected.

Silence the errors, since there is no much use for them anyway. And silence all of them just to be consistent.

This doesn't close https://github.com/tarantool/tarantool/issues/7980, because that issue still needs a proper fix. Although its severity is ameliorated now.

In-scope-of https://github.com/tarantool/tarantool/issues/7980

NO_DOC=bugfix
NO_CHANGELOG=bugfix
NO_TEST=nothing to test

### DIFF
--- a/src/lib/core/backtrace.c
+++ b/src/lib/core/backtrace.c
@@ -57,13 +57,13 @@ collect_current_stack(struct backtrace *bt, void *stack)
 	unw_context_t unw_ctx;
 	int rc = unw_getcontext(&unw_ctx);
 	if (rc != 0) {
-		say_error("unwinding error: unw_getcontext failed");
+		say_debug("unwinding error: unw_getcontext failed");
 		return stack;
 	}
 	unw_cursor_t unw_cur;
 	rc = unw_init_local(&unw_cur, &unw_ctx);
 	if (rc != 0) {
-		say_error("unwinding error: unw_init_local failed");
+		say_debug("unwinding error: unw_init_local failed");
 		return stack;
 	}
 	for (unsigned frame_no = 0; frame_no < BACKTRACE_FRAME_COUNT_MAX;
@@ -71,7 +71,7 @@ collect_current_stack(struct backtrace *bt, void *stack)
 		unw_word_t ip;
 		rc = unw_get_reg(&unw_cur, UNW_REG_IP, &ip);
 		if (rc != 0) {
-			say_error("unwinding error: unw_get_reg failed with "
+			say_debug("unwinding error: unw_get_reg failed with "
 				  "status: %d", rc);
 			return stack;
 		}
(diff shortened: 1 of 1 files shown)


src/lib/core/backtrace.c
```diff
@@ -57,29 +57,34 @@ collect_current_stack(struct backtrace *bt, void *stack)
 	unw_context_t unw_ctx;
 	int rc = unw_getcontext(&unw_ctx);
 	if (rc != 0) {
-		say_error("unwinding error: unw_getcontext failed");
+		/*
+		 * XXX: turn this and other say_debug occurrences back to
+		 * say_error once the error stops being spammed on mac.
+		 * (https://github.com/tarantool/tarantool/issues/7980).
+		 */
```
locker, line 64: I don't think we should ever revert it, because there may be other errors resulting in libunwind failures. I think that switching permanently to `say_debug` is fine, because if there's an error in libunwind, the user will see it, because backtraces will be missing. In this case, they may raise the logging level to 'debug' to find out the error reason.

src/lib/core/backtrace.c
```diff
@@ -57,29 +57,34 @@ collect_current_stack(struct backtrace *bt, void *stack)
 	unw_context_t unw_ctx;
 	int rc = unw_getcontext(&unw_ctx);
 	if (rc != 0) {
-		say_error("unwinding error: unw_getcontext failed");
+		/*
+		 * XXX: turn this and other say_debug occurrences back to
```
CuriousGeorgiy, line 61: Please change the comment start to `FIXME(gh-7980)` (see also https://github.com/tarantool/tarantool/pull/7994#discussion_r1055363750), so it will be easier to find, and also please mention this in #7980.

CuriousGeorgiy (commented):
While you are here, how about making these messages rate-limited?

src/lib/core/backtrace.c
sergepetrenko: Ok, I don't mind.

sergepetrenko:
@CuriousGeorgiy
> While you are here, how about making these messages rate-limited?

I'd rather not. I think rate limit is mostly for user-visible messages. Like the ones on level `warn`, `info` and maybe `verbose`.

`debug` is spammed with a whole bunch of messages, and that's its main purpose: to reflect everything that happened. So it might even be harmful to suppress some of them.

CuriousGeorgiy (approved):

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8172, opened 2023-01-17
libunwind: fix for EE build

nshy:
Commit 14f93aee86b8 ("libunwind: improve incremental build/rebuild") break EE build. We need fix the commit up to use current cmake source and build dir instead of top level ones as in `ext_project_autotools` arguments we specify `DIR` relative to current position in tree.

Follow-up #5665

NO_DOC=build fix
NO_TEST=build fix
NO_CHANGELOG=build fix


nshy:
Closed as superseded by https://github.com/tarantool/tarantool/pull/8174 PR. That one is better.

Issue #7774, opened 2022-10-04
Builtin modules overloading


igormunkin:
@Totktonada: The idea is to offer something similar to Perl's dual-file modules. We should elaborate package loader rules to allow convenient writing of built-in libraries replacements.

Totktonada:
Fixed by the following pull requests:

* https://github.com/tarantool/tarantool/pull/8173
* https://github.com/tarantool/tarantool/pull/8253
* https://github.com/tarantool/tarantool/pull/8250
* https://github.com/tarantool/tarantool/pull/8300
* https://github.com/tarantool/tarantool/pull/8335
* https://github.com/tarantool/tarantool/pull/8348
* https://github.com/tarantool/tarantool/pull/8351

Landed to 2.11.0-rc1.

PR #8173, opened 2023-01-18
lua: preliminary refactoring for built-in modules overloading [1/2]

Totktonada:
This patchset starts a series, which refactors built-in module registration mechanisms. The series aims several goals:

* Offer a standard tarantool specific way to register a built-in module. It allows to change the registration process for all built-in modules at once.
* Make the API of such functions simple and clean.
* Eliminate known problems of the existing approach to register built-in modules.

Details are described in the commit messages, but main changes are the following:

* Implement `luaT_newmodule()` as partial (and more clean) replacement for `luaL_register()` and `luaL_register_module()`.
* Use `luaT_setfuncs()` instead of `luaL_register(L, NULL, funcs)` for readability.
* Don't use a public module name for an internal part of the module.
* Use `internal.compat` and `internal.compat.lib` for modules behind `require('tarantool').compat`.
* Use `require()` for accessing built-in modules instead of direct indexing of `package.loaded`.

The most changes here are around Lua/C built-in modules and Lua/C parts of Lua built-in  modules. Changes around built-in modules, which are written in Lua, will follow as the second pull request. I plan to eliminate `package.loaded.foo = <...>` writes, extract loaders code from init.lua, move tarantool specific loaders to early stage of initialization, add more self-checks into the loading process.

Part of #7774


src/lua/utils.c
```diff
@@ -285,6 +286,19 @@ luaL_setcdatagc(struct lua_State *L, int idx)
 	lua_pop(L, 1);
 }
 
+int
+luaT_setfuncs(struct lua_State *L, const struct luaL_Reg *funcs)
```
igormunkin, line 290: It looks like `luaL_setfuncs` with `nup` being 0, doesn't it?

src/lua/utils.c
```diff
@@ -328,7 +342,43 @@ luaL_register_module(struct lua_State *L, const char *modname,
...
+	/* Copy the module table. */
+	lua_pushvalue(L, -1);
+
+	/* package.loaded[modname] = <module table> */
+	lua_setfield(L, -3, modname);
+
+	/* Stack: package.loaded, module table. */
+
+	/* Drop package.loaded. Stack: module table. */
+	lua_remove(L, -2);
```
igormunkin, line 378: Lua will do this automatically, later (see `return 1` below): there is no need in this line.

igormunkin (changes requested):
[PATCH 1/6] lua: add built-in module registration function
@Totktonada, thanks for the patch! There are some major comments I've left regarding the patch. I would ask you to fix them before I proceed with the review.

src/lua/utils.c
igormunkin: Discussed with @Totktonada: this `lua_remove` is required here, since `luaT_newmodule` is used mostly in C, hence the guest stack should be consistent after its return.

Ask Sasha to drop a comment regarding this.

test/app-luatest/http_client_unit_test.lua
```diff
@@ -10,12 +10,10 @@ local ffi = require("ffi")
...
 
-package.loaded.http = {
-    client = {
-        new = function() return end
-    }
+package.loaded['http.client'] = {
+    new = function() return end
 }
-local driver = package.loaded.http.client -- luacheck: no unused
+local driver = package.loaded['http.client'] -- luacheck: no unused
```
igormunkin, line 16: Well, now I see that tests are also adjusted. So, feel free to disregard [this review comment]().

igormunkin (approved):
**[PATCH 2/6] lua: eliminate luaL_register*() usages**

@Totktonada, thanks for the patch! LGTM, since mostly all modules are flat that makes substitution quite trivial. Anyway, please consider a couple of non critical comments below.

src/lua/utils.c
Buristan: Side note: Another one feature from 5.2:)

src/lua/utils.c
```diff
@@ -328,7 +342,43 @@ luaL_register_module(struct lua_State *L, const char *modname,
...
+	/* Get package.loaded. */
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
+
+	/* Verify that the module is not already registered. */
+	lua_getfield(L, -1, modname);
+	assert(lua_isnil(L, -1));
+	lua_pop(L, 1);
+
+	/* Create a module table. */
+	lua_newtable(L);
```
Buristan, line 364: Side note: I wondering: can we use `lua_createtable()` instead (as far as we can count necessary array length). But `luaT_newmodule()` doesn't look like performance critical function, so it's OK to use `lua_newtable()` here.

src/lua/utils.h
```diff
@@ -249,6 +249,14 @@ luaL_maplen(struct lua_State *L, int idx)
 
 extern int luaL_nil_ref;
 
+/**
+ * Fill a table on top of the Lua stack with functions.
+ *
+ * Leave the table on top of the stack.
```
Buristan, line 255: Don't get this comment. Am I right that this is the same table, that is filled?

Buristan:
Hi, Alexander!
Thanks for the patch-set!
I'll try to keep format of LGTM-s like Igor, as far as I can't approve one specific patch via GitHub. 

**[PATCH 1/6] lua: add built-in module registration function**

LGTM, after fixes mentioned by Igor.

**[PATCH 2/6] lua: eliminate luaL_register\*() usages**

LGTM. (I suppose that `luaT_setfuncs()` will be replaced with `luaL_setfuncs()`, am I correct?)

src/lua/utils.h
igormunkin: Yes. In Lua Reference manual notation it means `[0, 0]`.

Buristan:
**[PATCH 3/6] box/lua: eliminate luaL_register\*() usages in box** 
LGTM, except a single nit below.

I see you use `LUALIB_API const char *luaL_findtable(lua_State *L, int idx, const char *fname, int szhint)`. The last `szhint` parameter is the size of the hash-part of the table to create if necessary.
So, for example, for `box_lua_call_init()` routine, its value should be the following:
```
p sizeof(boxlib_internal)/sizeof(boxlib_internal[0]) - 1
$2 = 3
```

I suggest to create the corresponding macros to fill this parameter correct.

Buristan:
**[PATCH 4/6] lua: don't use public module name as internal one**
LGTM.

Buristan:
**[PATCH 5/6] lua: don't tweak package.loaded.compat**
LGTM, except 2 typos in the commit message.
> The module can't be pulled as `require('compat')`, because it is
tarantool specific tuning mechanism, not a general compatibility module.

Typo: s/tarantool specific tuning mechanism/the tarantool specific tuning mechanism/

> Before this patch the module was loaded as `package.loaded.compat`,
captured and then removed from `package.loaded`. However it would be

Typo: s/However/However,/

Buristan:
**[PATCH 6/6] lua: use require() to pull built-in modules**

LGTM, except a few insignificant nits.

> However we still have a guarantee that a registered module is accessible

Typo: s/However/However,/

Also, I suggest to add the comments about print/log callbacks order in the code too.

Also, we can use 
```c
if (luaT_call(L, 1, 1) != LUA_OK)
```
instead `0` status to be more readable. But it maybe contradict Tarantool's code style. So,  feel free to ignore.

igormunkin (approved):
**[PATCH 3/6] box/lua: eliminate luaL_register\*() usages in box**

@Totktonada, thanks for the patch! LGTM and I agree with @Buristan suggestion [here](https://github.com/tarantool/tarantool/pull/8173#issuecomment-1405074304).

src/lua/uri.c
```diff
@@ -301,7 +301,7 @@ luaT_uri_create_internal(lua_State *L)
 {
 	struct uri *uri = (struct uri *)lua_topointer(L, 1);
 	if (uri == NULL)
-		luaL_error(L, "Usage: uri.internal.uri_create(string|table)");
+		luaL_error(L, "Usage: uri_lib.uri_create(string|table)");
```
igormunkin, line 304: Looks like a typo: `s/uri_lib/uri.lib/`.

src/lua/uri.c
```diff
@@ -312,7 +312,7 @@ luaT_uri_set_create_internal(lua_State *L)
 {
 	struct uri_set *uri_set = (struct uri_set *)lua_topointer(L, 1);
 	if (uri_set == NULL)
-		luaL_error(L, "Usage: uri.internal.uri_set_create(string|table)");
+		luaL_error(L, "Usage: uri_lib.uri_set_create(string|table)");
```
igormunkin, line 315: Ditto.

igormunkin (approved):
**[PATCH 4/6] lua: don't use public module name as internal one**

@Totktonada, thanks for the patch! LGTM except two nits below and the next thought: I am confused with the lib name. There is a patch were you start using `internal.print` for quite the similar reason, and now you decided to use `lib` suffix. Could you please share the stats regarding `lib` usage you've mentioned in the commit message?

igormunkin (approved):
**[PATCH 5/6] lua: don't tweak package.loaded.compat**

@Totktonada, thanks for the patch! LGTM as trivial (hope you'll fix the typos, mentioned by @Buristan).

Anyway, still confused with the naming. Why is `internal` prefix used here?

src/box/lua/console.c
```diff
@@ -421,18 +421,19 @@ luaT_console_setup_write_cb(struct lua_State *L)
...
 		return;
 
-	say_set_stderr_callback(console_hide_prompt, console_show_prompt);
-
-	lua_getfield(L, LUA_GLOBALSINDEX, "package");
-	lua_getfield(L, -1, "loaded");
-	lua_getfield(L, -1, "internal.print");
+	lua_getfield(L, LUA_GLOBALSINDEX, "require");
+	lua_pushstring(L, "internal.print");
+	if (luaT_call(L, 1, 1) != 0)
```
igormunkin, line 426: Regarding `LUA_OK`: I don't agree with @Buristan, since `luaT_call` yields `0` on success and `1` on failure.

igormunkin (approved):
**[PATCH 6/6] lua: use require() to pull built-in modules**

@Totktonada, thanks for the patch! LGTM (hope you'll add the comments and fix the typos, mentioned by @Buristan).

src/lua/utils.c
Totktonada: Yes, it is quite similar to `luaL_setfuncs()` from Lua 5.2 (and from LuaJIT). I'll explain why I added my own function and we'll make a quorum decision :)

I want a helper function, which is easy to understand, when you look at code that uses it[^1]. It should have a description and a simple source, so I can go to declaration/definition and spot in seconds what exactly it is doing. The LuaJIT's `luaL_setfuncs()` has neither of those properties: it has no description in the source code and it has a bit non-obvious source due to the `nup` argument handling.

(I would also note that it is not very clear that one should find the documentation in the Lua 5.2 manual -- if we'll ignore the fact that this description is out of the sources and you can't go to it in a click.)

So, for my taste, presence of a technically satisfying function in LuaJIT doesn't mean that using it makes my code better comparing to using a simple and well documented own helper.

Dixi.

[^1]: Yes, I got tired by deducing `luaL_register()`'s behavior from its source code.

src/lua/compat.c
```diff
@@ -85,6 +85,6 @@ static const struct luaL_Reg internal_compat[] = {
 int
```
Totktonada, line 85: Quoted @Buristan's comment to discuss it in a thread:

> > The module can't be pulled as require('compat'), because it is
> > tarantool specific tuning mechanism, not a general compatibility module.
>
> Typo: s/tarantool specific tuning mechanism/the tarantool specific tuning mechanism/

src/lua/utils.c
```diff
@@ -285,6 +286,19 @@ luaL_setcdatagc(struct lua_State *L, int idx)
 	lua_pop(L, 1);
 }
 
+int
+luaT_setfuncs(struct lua_State *L, const struct luaL_Reg *funcs)
+{
+	assert(lua_gettop(L) >= 1);
+	assert(lua_istable(L, -1));
+
+	for (const struct luaL_Reg *f = funcs; f->name != NULL; ++f) {
```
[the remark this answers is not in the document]
Totktonada: Added `assert(funcs != NULL);`.

src/lua/utils.c
```diff
@@ -328,7 +342,43 @@ luaL_register_module(struct lua_State *L, const char *modname,
...
+{
+	say_debug("%s(%s)", __func__, modname);
+	assert(modname != NULL && funcs != NULL);
+
+	/* Get package.loaded. */
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
+
+	/* Verify that the module is not already registered. */
+	lua_getfield(L, -1, modname);
+	assert(lua_isnil(L, -1));
```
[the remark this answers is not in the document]
Totktonada: Okay, done.

src/lua/utils.c
Totktonada: I looked on it once again. The function can't be called from Lua directly, because it has arguments beyond `struct lua_State *`. So it would be a bit strange to highlight a fact, which is anyway considered as a [good practice](https://www.lua.org/manual/5.1/manual.html#lua_call). And it is part of the function contract.

Anyway, I changed the comment to make it more obvious that I'm doing that deliberately:

```c
/*
 * Drop package.loaded to leave the stack in its original
 * state.
 */
lua_remove(L, -2);
```

src/lua/httpc.lua
```diff
@@ -29,8 +29,7 @@
 --  SUCH DAMAGE.
 --
 
-local driver = package.loaded.http.client
-package.loaded.http = nil
+local driver = require('http.client')
```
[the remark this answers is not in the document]
Totktonada: I would prefer to no, because:

1. If we'll look on it as is, without taking into account further commits, this `package.loaded` entry is replaced with another table in `tarantool_lua_init()` (see line 874):

   https://github.com/tarantool/tarantool/blob/ed1362557f9dba58a7bc781ae8ca2a0dcd5dd85b/src/lua/init.c#L862-L882
2. I'm eliminating direct `package.loaded` maniputations here and there and will go into this direction in further patches.
3. If we'll take the *[4/6] lua: don't use public module name as internal one* commit into account, it replaces the name of the internal module with `http.client.lib` and it is done for many other modules without removing from `package.loaded`. (Moreover, I eliminate such tweak in *[5/6] lua: don't tweak package.loaded.compat* for another module.) I see no sense to add code purely to complicate access to internals: it is anyway possible to access them.

I've added a note to the commit message:

> Dropping of the `package.loaded['http.client']` field in unnecessary,
> because it is anyway replaced by the resulting module table in
> `tarantool_lua_init()`

It is a bit excessive, for my taste, but it is harmless, so let it be.

src/lua/utf8.c
```diff
@@ -459,8 +459,8 @@ tarantool_lua_utf8_init(struct lua_State *L)
 	unicode_ci_coll = coll_new(&def);
 	if (unicode_ci_coll == NULL)
 		goto error_coll;
-	luaL_register(L, "utf8", utf8_lib);
-	lua_pop(L, 1);
+	luaT_newmodule(L, "utf8", utf8_lib);
+	lua_setfield(L, LUA_GLOBALSINDEX, "utf8");
```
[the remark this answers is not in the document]
Totktonada: Added the comment:

```c
/* Assign to _G.utf8 for compatibility with Lua 5.3. */
lua_setfield(L, LUA_GLOBALSINDEX, "utf8");
```

test/app-luatest/http_client_unit_test.lua
Totktonada: Okay.

src/lua/utils.c
Totktonada: At cost of caller's code complication (either by adding the argument or by adding the argument and wrapping it into a macro to hide the complication). For my taste, it is better to keep it simple.

src/lua/utils.h
Totktonada: > Am I right that this is the same table, that is filled?

Yes.

> May be it is better to drop the second sentence, or specify that this is the filled table.

I've rewrote it a bit:

```c
/**   
 * Fill a table on top of the Lua stack with functions.
 *
 * The Lua stack remains unchanged after calling this function.
 */
```

src/box/lua/console.c
Totktonada: I agree with myself (and Igor) here: we use just 0 in tarantool for checking `luaT_call()` and `lua_pcall()` return values, so I'll keep it consistent. Moreover, technically `lua_pcall()` is described as a function that returns zero at success (not `LUA_OK` constant). And, yes, it is `luaT_call()`, which has no good description, but its source code has `return 0;` at the end.

src/box/lua/call.c
```diff
@@ -1145,7 +1145,8 @@ box_lua_call_init(struct lua_State *L)
 	trigger_add(&luaL_msgpack_default->on_update,
 		    &call_serializer_no_error_ext.update_trigger);
 
-	luaL_register(L, "box.internal", boxlib_internal);
+	luaL_findtable(L, LUA_GLOBALSINDEX, "box.internal", 0);
+	luaT_setfuncs(L, boxlib_internal);
```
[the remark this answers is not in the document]
Totktonada: There is `lengthof()` in `src/trivia/util.h`.

Regarding your proposal: I have several answers here.

The first answer is that I don't like this parameter. It only makes things more complex for illusion of performance. I'll not use `strchr()` on a string literal on a hot code path. So the only purpose of the parameter here is to make code less readable.

The second answer is specific for the `box.internal` table. It is creaccessed from several places and there is no sense to set the parameter to any particular value above zero. So it is zero in all places.

The third answer is that a non-zero value looks like a thing that we should remember to update at changes and it is developer's stress factor. So I should use zero or `lengthof(arr)`.

The fourth answer is that everything written above makes me close to be convinced that our own function without the parameter would be a good choice.

Let's summarize possible solutions and order them from conformant (like I don't see that the parameter is senseless) to radical (like I'm highlighting that the parameter is senseless):

1. Use `lengthof(arr)`, but zero for `box.internal`.
2. Always use zero.
3. Use own function without the parameter.

If we'll go the first way, I'll carry the lines to fit to 80 symbols:

```c
void
box_lua_ctl_init(struct lua_State *L)
{
	luaL_findtable(L, LUA_GLOBALSINDEX, "box.ctl",
		       lengthof(lbox_ctl_lib) - 1);
	luaT_setfuncs(L, lbox_ctl_lib);
	lua_pop(L, 1);
}
```

--will add `#include "trivia/util.h"` into each such file and all that new code gives neither simplicity nor performance.

So I'll go the second way.

Added an explanation to the commit message:

> One may wonder why fourth argument of `luaL_findtable()` (size hint for 
> the hash part of the table) is always zero in the added calls. It is an
> effect of the following points:
>
> * Those calls are not on hot paths and, moreover, there is no sense to
>   use `luaL_findtable()` with a string literal as the second argument on
>   a hot path (no sense to parse the string literal in runtime if a
>   performance is the goal).
> * Using `lengthof()` macro would add more code to read for a human (the
>   argument itself, a line carry to fit 80 symbols, an inclusion of
>   trivia/util.h) at no gain.
> * A hardcoded array length is a developer's stress factor: it looks like
>   it requires an update at a change of the array. So zero is better.

src/lua/compat.c
Totktonada: Added the indefinite artile. It fits better here AFAIU.

src/lua/uri.c
Totktonada: Not exactly. I assume something like `local uri_lib = require('uri.lib')`. If I'll write `uri.lib` it would mean that there is a table `uri`, which has the `lib` field. But there is no such table.

`uri_lib` is not really nice, because there is nothing named `uri_lib` in the public API. I hope that this error shouldn't occur at any more-or-less valid usage. However, if you have some idea how to make the error message better, I'm open for it.

src/lua/uri.c
Totktonada: Let's discuss above. If we'll change something I'll update both occurences. Closing this discussion.

src/box/lua/console.c
```diff
@@ -886,7 +886,7 @@ tarantool_lua_console_init(struct lua_State *L)
 		{"run_on_eval",		lbox_console_run_on_eval},
 		{NULL, NULL}
 	};
-	luaT_newmodule(L, "console", consolelib);
+	luaT_newmodule(L, "console.lib", consolelib);
```
Totktonada, line 902: Quoted @igormunkin's comments to discuss in it a thread:

> I am confused with the lib name. There is a patch were you start using `internal.print` for quite the similar reason, and now you decided to use `lib` suffix. Could you please share the stats regarding `lib` usage you've mentioned in the commit message?

> **[PATCH 5/6] lua: don't tweak package.loaded.compat**
> <...>
>
> Anyway, still confused with the naming. Why is `internal` prefix used here?

Totktonada: To make it clear, my current approach is the following.

1. Use `internal.foo` for a Lua module that is not exposed into the public API. Existing examples: `internal.argparse`, `internal.trigger`, `internal.print` (the latter is mine).
2. Use `foo.lib` for a Lua/C module that is not exposed into the public API and used in the `foo` module. Existing examples: `net.box.lib`. Hm, just one.

The compat module is a good example. It is not exposed directly as a module (`require('compat')` shouldn't return it). And it has both Lua and Lua/C parts. So the Lua part is named `internal.compat` according to the rule 1. And so the Lua/C part is named `internal.compat.lib` according to the rule 2.

The hack with the same named Lua/C and Lua modules was the most used approach. Now I need something else and I stick with what I saw around. It is self-consistent at least.

If there is another proposal, I'm open for it.

Totktonada:
Answered into all threads and created threads for discussions started at top level to ease the tracking. Fixed all the comments, where I have no doubts.

src/box/lua/call.c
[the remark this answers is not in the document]
igormunkin: Meh. OK, zero also works for me.

src/lua/utils.c
igormunkin: I totally agree with you regarding `luaL_register`, so I said no word regarding `luaT_newmodule`. However, I still can't agree with introducing `luaT_setfuncs`. As you mentioned, the latter function has quite the same source code and (IMHO) handling `nup` parameter makes it as much intricate as `lua_pushcclosure` usage is more complex than `lua_pushcfunction`.

Generally, I'm OK with introducing helpers, that provides similar functionality but results to quite a different effect (`luaL_register` and `luaL_register_module` are just a hell), but introducing the function with 2 parameters instead of using the one provided by LuaJIT C API with 3 parameters looks overkill to me..

src/lua/utils.c
igormunkin: Thanks a huge!

src/box/lua/call.c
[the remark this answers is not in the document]
Buristan: Zero is OK for me, as far as this code is called only once on bootstrap.

src/lua/utils.c
Buristan: I'll try to suggest the solution, that may help all here to stay satisfied...
Can we introduce our own helper `luaT_setfuncs(L, L_Reg)` which simply calls `luaL_setfuncs(L, L_Reg, 0)`?
Here, we can add the comment inside our helper, where we can take a look to the documentation and the description of the last parameter.

src/lua/utils.c
Buristan: Yes, its totally fine for me.

src/lua/utils.h
Buristan: Nice, thanks!

src/box/lua/console.c
Buristan: OK, my bad.

src/lua/uri.c
igormunkin: I looked over `src/` directory and found a big variety of names, depending on the function type: public/private, function/method, etc. All in all, I'm OK with your choice, so feel free to ignore.

src/box/lua/console.c
igormunkin: OK, that works for me. Anyway, it would be great to see this in our style guide (or something similar).

src/lua/utils.c
igormunkin: @Buristan, this is not a compromise at all. I'm for using LuaJIT interfaces in Tarantool; @Totktonada wants to use the functions that have the clear signature and are easy to understand.

Your proposal contradicts both points :)

src/box/lua/call.c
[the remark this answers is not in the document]
Totktonada: It seems, we all agreed (and the implementation conforms to that). Closing the discussion.

src/lua/httpc.lua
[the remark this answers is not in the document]
igormunkin: Thanks!

src/lua/utils.c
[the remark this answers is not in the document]
igormunkin: Neat, thanks!

src/lua/utils.c
Totktonada: I'm motivated by two points:

* Documentation.
* Simple source code.

If I'll use `luaL_setfuncs()` in `luaT_setfuncs()`, the latter point would not be in effect IMHO. And introducing the helper just because of one of those points looks not motivated enough.

Okay. Let's go with `luaL_setfuncs(<...>, 0)`. It is not so hard to understand as `luaL_register()` and I can agree.

Sooner or later we should do something to improve in-source LuaJIT's API documentation.

src/box/lua/console.c
Totktonada: I filed a stub issue: https://github.com/tarantool/doc/issues/3308.

src/lua/uri.c
Totktonada: Let me know if you'll have a proposal that looks definitely better than what I implemented now (I see no one currently). Closing the discussion for now.

igormunkin (approved):
@Totktonada, I've looked over all changes once more and resolved all my comments except the thread related to `luaT_setfuncs`. I believe this is my final review here, so I will not insist on replacing your helper with `luaL_setfuncs`, if @Buristan is OK with leaving it.

**TL;DR:** The PR itself LGTM, and I leave my vote in `luaT_setfuncs` thread to @Buristan.

src/lua/utils.c
Totktonada: I've replaced `luaT_setfuncs()` with `luaL_setfuncs()` and updated commits messages (mainly the first commit).

src/lua/utils.c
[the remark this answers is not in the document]
Totktonada: (Irrelevant now, because `luaT_setfuncs()` was removed after a discussion in [this](https://github.com/tarantool/tarantool/pull/8173#discussion_r1083928591) thread.)

Buristan (approved):
Thanks for the fixes!
LGTM.

src/lua/utils.c
Totktonada: Seems resolved now. Closing the discussion.

Totktonada:
@igormunkin Apline fails in CI are expected, see PR #8232. I can rebase the pull request and it will re-run all the jobs, but I propose to ignore the fails on Alpine and proceed.

Issue #8026, opened 2022-12-08
ninja couldn't find a target for building build/libunwind/dest/lib/libunwind-aarch64.a' on AArch64


ligurio:
**Bug description**

It is not possible to build Tarantool on AArch64 using `ninja-build`.

* OS: Linux
* Architecture: AArch64 (graviton)

Version: b2dab5f4270da1d1a0759245c8a1d93cd8616941

**Steps to reproduce**

```
cmake -S . -B .  -G Ninja -DCMAKE_BUILD_TYPE=Debug -DLUAJIT_ENABLE_GC64=ON
...
-- Configuring done
-- Generating done
-- Build files have been written to: /opt/actions-runner/_work/luajit/luajit

cmake --build . --parallel 8
ninja: error: 'build/libunwind/dest/lib/libunwind-aarch64.a', needed by 'src/tarantool', missing and no known rule to make it
make: *** [.test.mk:42: build] Error 1
Error: Process completed with exit code 2.
```

**Actual behavior**

`ninja` couldn't find a target for building `build/libunwind/dest/lib/libunwind-aarch64.a'`

**Expected behavior**

Building finished successfully.

Job "[Tarantool (Linux/aarch64) Debug GC64:ON / luajit-integration](https://github.com/tarantool/luajit/actions/runs/3437429234/jobs/5732216487#logs)"

igormunkin:
Fixed by #8019. Closing.

PR #8175, opened 2023-01-18
Maintenance of LuaJIT integration CI

igormunkin:
Small changes to remove the old code from LuaJIT integration CI:
* Removed the obsolete `inputs.host` parameter
* Enabled `CMAKE_EXTRA_PARAMS` for .test.mk (since #8026 has been fixed)

---

P.S. LuaJIT CI is [green](https://github.com/tarantool/luajit/commit/52ff4c3) as a result of this patch.

### DIFF
--- a/.github/workflows/luajit-integration.yml
+++ b/.github/workflows/luajit-integration.yml
@@ -2,7 +2,7 @@
 # implement integration testing without manual LuaJIT submodule
 # bump and draft PR creation.
 # It bumps the given LuaJIT revision for the given Tarantool
-# release, builds Tarantool with the given buildtype and GC64
+# release, builds Tarantool with the given CMAKE_EXTRA_PARAMS
 # option and runs Tarantool testing suite.
 
 name: LuaJIT integration testing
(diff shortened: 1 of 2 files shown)


.test.mk
```diff
@@ -275,10 +275,10 @@ test-coverity: build-coverity
 # LuaJIT integration testing #
 ##############################
 
-test-luajit-linux-x86_64: build run-luajit-test run-test
+test-luajit-Linux-x86_64: build run-luajit-test run-test
```
ylobankov, line 278: Actually, I like this naming :) Can we keep it further?

.github/workflows/luajit-integration.yml
```diff
@@ -2,25 +2,18 @@
...
+        required: true
         type: string
-        # XXX: Temporary ugly hack: use 'self-hosted' as a default
-        # value to fool GitHub Actions machinery choosing the
-        # appropriate runner for the workflow (since the empty
-        # string is not a valid parameter).
-        default: self-hosted
       # FIXME: Remove this inputs entry when transition to
       # CMAKE_EXTRA_PARAMS usage is finished.
       buildtype:
```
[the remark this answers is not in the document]
igormunkin: Unfortunately, not yet: `buildtype` and `GC64` are still being used in the long-term LuaJIT branches. They will be removed when this PR is merged into the trunk.

.test.mk
igormunkin: We use this values within LuaJIT matrix and I, actually, don't want to make additional tweaks to obtain the values for the current naming.

I suggest to try the new naming and return to this question, if we face issues with it (either Makefile targets mislookup or its complex maintenance).

ylobankov (approved):
LGTM

igormunkin:

> LGTM

Added your tag:

PR #8177, opened 2023-01-18
box: handle SIGBUS in flightrec (CE Part)

nshy:
Flight recorder uses mmapped file for performance reasons. Unfortunately if for some reason mapping is not possible we got SIGBUS on accessing mmapped memory.

We already handle this issue on flight recorder start in [1]. But if we got SIGBUS in the meanwhile we currently got a crash. Let's just panic.

[1] https://github.com/tarantool/tarantool-ee/commit/e68b84768f3ae76aa9f03d36dd6f0587b884e1bb

EE Part: https://github.com/tarantool/tarantool-ee/pull/344

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -2898,6 +2898,13 @@ box_feedback_report_crash(struct crash_info *cinfo)
 static void
 box_crash_callback(struct crash_info *cinfo)
 {
+	if (cinfo->signo == SIGBUS &&
+	    flightrec_is_mmapped_address(cinfo->siaddr)) {
+		fprintf(stderr, "error accessing flightrec file\n");
+		fflush(stderr);
+		_exit(EXIT_FAILURE);
+	}
+
 	crash_report_stderr(cinfo);
 
 	if (box_feedback_crash_enabled &&
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -32,6 +32,9 @@
 
 #include "trivia/config.h"
 
+#include <sys/utsname.h>
+#include <spawn.h>
+
 #include "lua/utils.h" /* lua_hash() */
 #include "fiber_pool.h"
 #include <say.h>
@@ -91,6 +94,7 @@
 #include "flightrec.h"
 #include "wal_ext.h"
 #include "mp_util.h"
+#include "small/static.h"
 
 static char status[64] = "unconfigured";
 
@@ -226,6 +230,15 @@ static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
 
 /** Whether the triggers for "synced" recovery stage have already run. */
 static bool recovery_state_synced_is_reached;
+
+/** PATH_MAX is too big and 2K is recommended limit for web address. */
+#define BOX_FEEDBACK_HOST_MAX 2048
+
+/** Feedback host URL to send crash info to. */
+static char box_feedback_host[BOX_FEEDBACK_HOST_MAX];
+
+/** Whether sending crash info to feedback URL is enabled. */
+static bool box_feedback_crash_enabled;
 
 static int
 box_run_on_recovery_state(enum box_recovery_state state)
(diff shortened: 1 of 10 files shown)


src/box/box.cc
```diff
@@ -2727,20 +2745,214 @@ box_set_prepared_stmt_cache_size(void)
...
+	char *tail = &p[SMALL_STATIC_SIZE];
+	char *e = &p[SMALL_STATIC_SIZE];
+	char *head = p;
+
+	int total = 0;
+	(void)total;
+	int size = 0;
+
+#define snprintf_safe(...) SNPRINT(total, snprintf, p, size, __VA_ARGS__)
+#define jnprintf_safe(str) SNPRINT(total, json_escape, p, size, str)
```
locker, line 2770: Nit: Let's undef these two macros as soon as we don't need them anymore.

src/lib/core/crash.h
```diff
@@ -9,23 +9,124 @@
...
+#ifdef HAS_GREG
+	/**
+	 * Registers contents.
+	 */
+	struct crash_greg greg;
+#endif
+	/**
+	 * Timestamp in seconds (realtime).
+	 */
+	time_t timestamp_rt;
```
locker, line 103: Why did you change the type of this field from `uint64_t` to `time_t`?

src/main.cc
```diff
@@ -721,8 +721,22 @@ main(int argc, char **argv)
...
+	/*
+	 * We need to keep clock data locally to report uptime without binding
+	 * to libev and etc. Because we're reporting information at the moment
+	 * when crash happens and we are to be independent as much as we can.
+	 */
+	struct timespec ts;
+	if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0)
+		box_start_time = ts.tv_sec;
+	else
+		say_syserror("fail to get box_start_time, ignore");
```
locker, line 737: Nit: Let's change to `panic`? We don't expect a failure here, and if the clock doesn't work, we'll probably panic somewhere else anyway.

src/lib/core/crash.h
```diff
@@ -9,23 +9,124 @@
 extern "C" {
 #endif /* defined(__cplusplus) */
 
-/**
- * PATH_MAX is too big and 2K is recommended
- * limit for web address.
- */
-#define CRASH_FEEDBACK_HOST_MAX 2048
+#if defined(TARGET_OS_LINUX) && defined(__x86_64__)
```
locker, line 12: You need to include `trivia/config.h`, otherwise `TARGET_OS_LINUX` may not be defined.

src/lib/core/crash.h
```diff
@@ -9,23 +9,124 @@
...
- * Initialize crash subsystem.
+#ifdef HAS_GREG
+/*
+ * Values of x86 registers.
  */
-void
-crash_init(const char *tarantool_bin);
+struct crash_greg {
+	/* r8 register value */
+	uint64_t r8;
```
locker, line 22: Nit: Please include `stdint.h` and other headers needed to compile this one.

src/lib/core/crash.h
```diff
@@ -9,23 +9,124 @@
...
+	 */
+	char backtrace_buf[1024];
+#endif
+};
+
+typedef void
+(*crash_callback_f)(struct crash_info *cinfo);
+
+/* Callback to call on crash. */
+extern crash_callback_f crash_callback;
```
locker, line 129: Nit: Please improve on the comment: can it be NULL? what's it set to by default?

src/main.cc
```diff
@@ -721,8 +721,22 @@ main(int argc, char **argv)
...
+
+	crash_callback = box_crash_callback;
```
locker, line 739: IMO, better set it in `box_init` - anyway, feedback won't be enabled until the box is configured.

src/box/box.cc
```diff
@@ -2925,6 +2925,13 @@ box_feedback_report_crash(struct crash_info *cinfo)
 void
 box_crash_callback(struct crash_info *cinfo)
 {
+	if (cinfo->signo == SIGBUS &&
+	    flightrec_is_mmapped_address(cinfo->siaddr)) {
+		fprintf(stderr, "error accessing flightrec file\n");
+		fflush(stderr);
+		_exit(EXIT_FAILURE);
```
locker, line 2932: The callback is invoked after dumping the stack trace to stderr. I think it should be called before, because a stack trace looks pretty scary to users. AFAIR we agreed to *replace* the default crash callback instead of adding an extra handler and add make the default handler (the one that dumps the stack trace) available outside crash.c so that it can be reused in box.cc.

locker (changes requested):

src/main.cc
locker: Agreed to leave as is, because the old code used `pr_syserr`.

Let's use a more human-readable error message (without reference to `box_start_time`).

src/box/box.cc
```diff
@@ -227,6 +231,20 @@ static const char *box_recovery_state_strs[box_recovery_state_MAX] = {
 /** Whether the triggers for "synced" recovery stage have already run. */
 static bool recovery_state_synced_is_reached;
 
+char tarantool_path[PATH_MAX];
+time_t box_start_time;
```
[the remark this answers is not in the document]
nshy: Done.

src/box/lua/cfg.cc
```diff
@@ -484,7 +484,7 @@ box_lua_cfg_init(struct lua_State *L)
 		{"cfg_set_replication_anon", lbox_cfg_set_replication_anon},
 		{"cfg_set_net_msg_max", lbox_cfg_set_net_msg_max},
 		{"cfg_set_sql_cache_size", lbox_set_prepared_stmt_cache_size},
-		{"cfg_set_crash", lbox_cfg_set_crash},
+		{"cfg_set_feedback_crash", lbox_cfg_set_feedback_crash},
```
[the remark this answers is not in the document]
nshy: Done.

src/box/box.cc
nshy: Done.

src/lib/core/crash.h
nshy: We don't use nanosecond precision actually. Moved this and related change to a distinct patch.

src/main.cc
nshy: Done. `"failed to get start time, ignore"`

src/lib/core/crash.h
nshy: Fixed.

src/lib/core/crash.h
nshy: Fixed.
```
+#include <stdint.h>
+#include <time.h>
```

src/lib/core/crash.h
nshy: Done.

src/main.cc
nshy: Done.

src/box/box.cc
nshy: Issue with reporting crash info on flightrec SIGBUS is fixed.

As to error message - leave "as is" as discussed f2f.

src/lib/core/crash.h
```diff
@@ -5,34 +5,149 @@
...
 
 /**
  * Initialize crash signal handlers.
  */
 void
 crash_signal_init(void);
 
+/**
+ * Report crash information to the stderr (usually a current console).
+ */
```
locker, line 147: Nit: Let's also drop this comment from the function definition in the C file to avoid duplication.

src/main.h
```diff
@@ -39,6 +40,16 @@ extern "C" {
...
 
+/**
+ * Path to Tarantool binary of the process.
+ */
+extern char tarantool_path[PATH_MAX];
+
+/**
+ * Application start time on monotonic clocks.
+ */
+extern time_t start_time_safe;
```
locker, line 51: Nit: `_safe` suffix looks confusing. Also, it'd be nice to have the `tarantool_` prefix for consistency with `tarantool_path`. Let's rename to `tarantool_start_time`?

src/main.h
```diff
@@ -39,6 +40,16 @@ extern "C" {
 
 double tarantool_uptime(void);
 
+/**
+ * Path to Tarantool binary of the process.
+ */
+extern char tarantool_path[PATH_MAX];
```
locker, line 46: Nit: AFAIK you don't need to specify the array length in the declaration. Let's replace with `extern char tarantool_path[]` to eliminate dependency on `limits.h`.

src/box/flightrec.h
```diff
@@ -67,6 +67,17 @@ box_set_flightrec(void)
 	return 0;
 }
 
+/**
+ * This function is called in SIGBUS handler to check whether accessed address
+ * belongs to flightrec file.
+ */
+static inline bool
```
locker, line 74: Nit: Let's add `stdbool.h` to the header list.

locker (changes requested):

src/lib/core/crash.c
```diff
@@ -115,18 +115,11 @@ static struct crash_info {
...
-static inline uint64_t
-timespec_to_ns(struct timespec *ts)
-{
-	return (uint64_t)ts->tv_sec * 1000000000 + (uint64_t)ts->tv_nsec;
-}
+static time_t timestamp_mono = 0;
 
 static char *
-ns_to_localtime(uint64_t timestamp, char *buf, ssize_t len)
+ns_to_localtime(time_t sec, char *buf, ssize_t len)
```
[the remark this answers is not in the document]
locker: I see that in the next patch you drop this helper function. Maybe, better drop it in the first patch then?

src/lib/core/crash.c
```diff
@@ -90,7 +90,7 @@ static struct crash_info {
 	/**
 	 * Timestamp in nanoseconds (realtime).
 	 */
-	uint64_t timestamp_rt;
+	time_t timestamp_rt;
```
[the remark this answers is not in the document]
nshy: Fixed to `/** Timestamp in seconds (realtime). */` as in next patch.

Changed type to `long` as discussed f2f.

src/box/box.cc
```diff
@@ -4900,6 +5115,7 @@ box_init(void)
 	 * supported from box.cfg not being called yet.
 	 */
 	builtin_events_init();
+	crash_callback = box_crash_callback;
```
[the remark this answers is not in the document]
nshy: Discussed f2f and decided not to reset the callback. This keeps existent behavior of reporting crashes to feedback URL even on application shutdown.

src/box/box.h
```diff
@@ -306,12 +306,17 @@ void box_set_replication_skip_conflict(void);
...
 int box_set_txn_timeout(void);
 int box_set_txn_isolation(void);
 int box_set_auth_type(void);
 int box_set_bootstrap_strategy(void);
 
+/**
+ * Applies box.cfg feedback parameters.
+ * On success, returns 0. On error, sets diag and returns -1.
+ */
+int box_set_feedback(void);
```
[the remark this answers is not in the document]
nshy: Fixed.

src/main.h
nshy: Done.

src/main.h
nshy: Done.

src/box/flightrec.h
nshy: Done.

locker (approved):

locker:
@nshy Compilation fails on a few platforms. Please fix.

locker:
Cherry-picked to 2.10.

PR #8178, opened 2023-01-18
sql: drop unused modules

ImeevMA:
Some of the SQL modules have not been used for a long time. This patch drops these modules.

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring


Gerold103 (approved):
Thanks for the patch!

locker (approved):

Issue #8122, opened 2023-01-10
Excess conflict in mvcc engine


alyapunov:
The simple reproduce:
```lua
box.cfg{memtx_use_mvcc_engine = true}
_ = box.schema.space.create('test'):create_index('pk')
box.space.test:replace{1}
require('fiber').create(function() box.space.test:delete{1} end)
box.space.test:delete{1} -- this command fails with error: Transaction has been aborted by conflict
```
There's no reason for the second transaction to be aborted.

PR #8183, opened 2023-01-19
applier: fix isolation level of applied transactions

### LINKED ISSUE #8121: mvcc shouldn't work on a replication stream
Reproduced on latest master (2.11.0-entrypoint-850-g182034dc9)

**Bug description**
Run 2 Tarantool instances in console:
```lua
-- Instance 1.
-- Step 1.
box.cfg{
    listen = 3301,
    memtx_use_mvcc_engine = true,
    txn_isolation = 'read-confirmed',
}

box.once('bootstrap', function()
    box.space.test:insert{1}
end)

-- Step 3.
box.space.test:delete{1}
box.space.test:delete{1}

-- Instance 2.
-- Step 2.
box.cfg{replication = 3301, memtx_use_mvcc_engine = true, txn_isolation = 'read-confirmed'}
box.cfg{replication = ""}

-- Step 4.
box.cfg{replication = 3301}
```

After these steps the replica will disconnect from master with an `ER_TRANSACTION_CONFLICT` error.
It's box.info.replication will read:
```
---
- 1:
    id: 1
    uuid: 95005d25-21b3-42cb-8eb8-9aed62f35771
    lsn: 8
    upstream:
      peer: '3301'
      lag: 0.00046181678771973
      status: stopped
      idle: 4.4793854899999
      message: Transaction has been aborted by conflict
  2:
    id: 2
    uuid: e00fad89-c417-4d39-af1c-e6b7197ea38b
    lsn: 0
...

```

The reason for this is that from master's point of view the deletes were run sequentially, but for replica they were concurrent: that's because replica applies transactions in parallel: commits them asynchronously not waiting for a wal write to complete.

It seems mvcc shouldn't work on a replication stream. At least when this stream comes from a single master. Because having transactions in master's WAL means there were no conflicts when master committed them.

CuriousGeorgiy:
Closes #8121

### DIFF
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1378,6 +1378,7 @@ apply_plain_tx(uint32_t replica_id, struct stailq *rows,
 	struct applier_tx_row *item;
 	if (txn == NULL)
 		 return -1;
+	txn->isolation = TXN_ISOLATION_READ_COMMITTED;
 
 	stailq_foreach_entry(item, rows, next) {
 		struct xrow_header *row = &item->row;
--- a/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
+++ b/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
@@ -1,0 +1,70 @@
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g = t.group(nil, t.helpers.matrix({engine = {'memtx', 'vinyl'}}))
+
+g.before_all(function(cg)
+    cg.replica_set = replica_set:new()
+    cg.master = cg.replica_set:build_and_add_server(
+            {
+                alias = 'master',
+                box_cfg = {
+                    replication_timeout = 0.1,
+                },
+            })
+    cg.replica = cg.replica_set:build_and_add_server(
+            {
+                alias = 'replica',
+                box_cfg = {
+                    replication = server.build_listen_uri('master'),
+                    replication_timeout = 0.1,
+                    memtx_use_mvcc_engine = true,
+                    txn_isolation = 'read-confirmed',
+                }
+            })
+    cg.replica_set:start()
+    cg.master:exec(function(engine)
+        local s = box.schema.space.create('s', {engine = engine})
+        s:create_index('pk')
+        s:insert{0}
+    end, {engine = cg.params.engine})
+    t.helpers.retrying({}, function()
+        cg.replica:assert_follows_upstream(cg.master:get_instance_id())
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+    cg.replica:exec(function()
+        box.cfg{replication = ""}
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.replica_set:drop()
+end)
+
+-- Checks that replication stream does not get transaction conflict errors.
+g.test_replication_stream_transaction_conflict_errors = function(cg)
+    cg.replica:exec(function()
+        local t = require('luatest')
+
+        t.assert_equals(box.space.s:select{}, {{0}})
+    end)
+    cg.master:exec(function()
+        box.space.s:delete{0}
+        -- Delete statement sees {0} deleted by another prepared statement and
+        -- must not conflict with it.
+        box.space.s:delete{0}
+    end)
+    cg.replica:exec(function(master_uri)
+        box.cfg{replication = master_uri}
+    end, {server.build_listen_uri('master')})
+    t.helpers.retrying({}, function()
+        cg.replica:assert_follows_upstream(cg.master:get_instance_id())
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+    cg.replica:exec(function()
+        local t = require('luatest')
+
+        t.assert_equals(box.space.s:select{}, {})
+    end)
+end
--- a/changelogs/unreleased/gh-8121-memtx-mvcc-replication-stream.md
+++ b/changelogs/unreleased/gh-8121-memtx-mvcc-replication-stream.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed possibility of transaction conflict errors during replication stream
+  (gh-8121).


src/box/applier.cc
```diff
@@ -1378,6 +1378,7 @@ apply_plain_tx(uint32_t replica_id, struct stailq *rows,
 	struct applier_tx_row *item;
 	if (txn == NULL)
 		 return -1;
+	txn->isolation = TXN_ISOLATION_READ_COMMITTED;
```
sergepetrenko, line 1381: I'd probably use `box_txn_set_isolation` here. Then we'll be safe in case setting txn isolation becomes more complex than just setting a variable someday.

Or even better, to avoid the extra checks `box_txn_set_isolation` performs, you can extract the part which sets the isolation in a separate routine and use it here.
And also use it in `box_txn_set_isolation` after the checks.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,57 @@
...
+    cg.replica = cg.cluster:build_server(
+            {
+                alias = 'replica',
+                box_cfg = {
+                    replication = server.build_listen_uri('master'),
+                    memtx_use_mvcc_engine = true,
+                    txn_isolation = 'read-confirmed',
+                }
+            })
+    cg.cluster:add_server(cg.master)
```
sergepetrenko, line 19: BTW, you can use `build_and_add_server` instead of `build_server`, `add_server`.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,57 @@
...
+    cg.cluster:add_server(cg.replica)
+    cg.cluster:start()
+    cg.master:exec(function()
+        local s = box.schema.space.create('s')
+        s:create_index('pk')
+        s:insert{0}
+    end)
+    t.helpers.retrying({}, function()
+        cg.replica:assert_follows_upstream(cg.master:get_instance_id())
+    end)
```
sergepetrenko, line 29: Better use `cg.replica:wait_for_vclock_of(cg.master)` here.
the fact that replica follows upstream doesn't prove anything here. It might follow master, but it might not have the same vclock yet.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,57 @@
...
+end)
+
+-- Checks that replication stream does not get transaction conflict errors.
+g.test_replication_stream_transaction_conflict_errors = function(cg)
+    cg.replica:exec(function()
+        box.cfg{replication = ""}
+    end)
+    cg.master:exec(function()
+        box.space.s:delete{0}
+        box.space.s:delete{0}
```
sergepetrenko, line 43: I see that the test fails on the current master, which's good.

But wasn't the case with 2 deletes fixed in #8122 ?
For example, when I try to run the original reproducer with 2 deletes, it doesn't cause a conflict.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,57 @@
+local t = require('luatest')
+local cluster = require('luatest.replica_set')
```
sergepetrenko, line 2: Nit: since the module was renamed to `replica_set`, let's call it replica_set in new tests.

sergepetrenko (changes requested):
Thanks for the patch!

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: > For example, when I try to run the original reproducer with 2 deletes, it doesn't cause a conflict.

But at the same time you say:
> I see that the test fails on the current master, which's good.

> But wasn't the case with 2 deletes fixed in https://github.com/tarantool/tarantool/issues/8122 ?

No, AFAIC, it wasn't — I run the original reproducer with 2 deletes and it causes conflict on master.
Screen Shot 2023-01-20 at 11 30 04

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Fixed.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Changed to `cg.master:wait_for_downstream_to(cg.replica)`.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Thanks for the suggestion, applied!

src/box/applier.cc
CuriousGeorgiy: Not sure about this. `box_txn_set_isolation` is really general: even if we factor out the part that sets the passed transaction's isolation level, it still performs additional checks, e.g., whether the isolation level is default, and whether the transaction needs to be linearized.

@alyapunov RFC

alyapunov: I agree with @CuriousGeorgiy, box_txn_set_isolation is a part of API but we do some internal thing here and there's no reason to do additional check.

Perhaps one day we'll need a wrapper like `txn_set_isolation` but it seems to be an overkill today.

CuriousGeorgiy: Besides, we didn't do this this in vinyl: https://github.com/tarantool/tarantool/blob/a24714c55305280838576fa6dec9a9781763e893/src/box/vy_tx.c#L344.

CuriousGeorgiy: @locker what do you think about this?

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,61 @@
...
+                alias = 'replica',
+                box_cfg = {
+                    replication = server.build_listen_uri('master'),
+                    memtx_use_mvcc_engine = true,
+                    txn_isolation = 'read-confirmed',
+                }
+            })
+    cg.replica_set:start()
+    cg.master:exec(function()
+        local s = box.schema.space.create('s')
```
locker, line 21: Let's make this space multi-engine.

src/box/applier.cc
locker: Vinyl inherits the isolation level from `struct txn`:

https://github.com/tarantool/tarantool/blob/02e2b2920e2c99c349bcfc246976a0f3e0c2a199/src/box/vinyl.c#L2433

Still, it'd be nice to have a test that makes sure this patches fixes the issue for both engines.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Made the test multi-engine.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,61 @@
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g = t.group(nil, t.helpers.matrix({eng = {'memtx', 'vinyl'}}))
```
locker, line 5: Nit: please don't use confusing abbreviations like 'eng' - 'engine' is short enough.

CuriousGeorgiy: Ok, fixed.

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,61 @@
...
+    cg.master = cg.replica_set:build_and_add_server({alias = 'master'})
+    cg.replica = cg.replica_set:build_and_add_server(
+            {
+                alias = 'replica',
+                box_cfg = {
+                    replication = server.build_listen_uri('master'),
+                    memtx_use_mvcc_engine = true,
+                    txn_isolation = 'read-confirmed',
+                }
+            })
```
sergepetrenko, line 18: Don't forget to lower `replication_timeout`. We usually set 0.1, and it speeds the tests up quite a bit. For example for this test the time went down from ~ 3.4 seconds to ~ 1.6 on my machine.
```diff
diff --git a/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua b/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
index 1157a287c..7debcda00 100644
--- a/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
+++ b/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
@@ -6,12 +6,19 @@ local g = t.group(nil, t.helpers.matrix({eng = {'memtx', 'vinyl'}}))
 
 g.before_all(function(cg)
     cg.replica_set = replica_set:new()
-    cg.master = cg.replica_set:build_and_add_server({alias = 'master'})
+    cg.master = cg.replica_set:build_and_add_server(
+            {
+                alias = 'master',
+                box_cfg = {
+                    replication_timeout = 0.1,
+                },
+            })
     cg.replica = cg.replica_set:build_and_add_server(
             {
                 alias = 'replica',
                 box_cfg = {
                     replication = server.build_listen_uri('master'),
+                    replication_timeout = 0.1,
                     memtx_use_mvcc_engine = true,
                     txn_isolation = 'read-confirmed',
                 }
```

sergepetrenko (changes requested):

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Ok, fixed. Why isn't this the default value then?

sergepetrenko (approved):
LGTM!

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```diff
@@ -0,0 +1,68 @@
...
+
+-- Checks that replication stream does not get transaction conflict errors.
+g.test_replication_stream_transaction_conflict_errors = function(cg)
+    cg.replica:exec(function()
+        local t = require('luatest')
+
+        t.assert_equals(box.space.s:select{}, {{0}})
+    end)
+    cg.master:exec(function()
+        box.space.s:delete{0}
```
kyukhin, line 53: Could you please add a comment describing why there're two deletes?

kyukhin (changes requested):

test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
CuriousGeorgiy: Actually, this is just a reproducer from the ticket, but I also added a comment as you asked for.

kyukhin:
CI for alpine seems to be broken due to other reasons. So, let's ignore them.

kyukhin (approved):

PR #8198, opened 2023-01-23
applier: name ballot_watcher fiber properly

sergepetrenko:
Currently all ballot_watcher fibers for different appliers have identical names: "applier_ballot_watcher". It's hard to distinguish them in logs and to know which ballot_watcher belongs to which applier, and that complicates debugging quite a bit.

Let's fix this and name the fibers properly. With master's uri at the end.

NO_DOC=no behaviour changed
NO_CHANGELOG=no behaviour changed
NO_TEST=doesn't need one

### DIFF
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -700,8 +700,9 @@ applier_connect(struct applier *applier)
 	if (method_default == NULL)
 		method_default = AUTH_METHOD_DEFAULT;
 
-	applier->ballot_watcher = fiber_new_system("applier_ballot_watcher",
-						   applier_ballot_watcher_f);
+	applier->ballot_watcher = applier_fiber_new(applier, "ballot_watcher",
+						    applier_ballot_watcher_f,
+						    false);
 	applier->ballot_watcher->f_arg = applier;
 	fiber_wakeup(applier->ballot_watcher);
 
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -170,13 +170,14 @@ applier_check_sync(struct applier *applier)
  * but doesn't start it.
  */
 static struct fiber *
-applier_fiber_new(struct applier *applier, const char *name, fiber_func func)
+applier_fiber_new(struct applier *applier, const char *name, fiber_func func,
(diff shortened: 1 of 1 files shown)


sergepetrenko:
@Gerold103, @grafin, sorry for a ton of small PRs, guys. PTAL once you have time.

src/box/applier.cc
```diff
@@ -170,13 +170,14 @@ applier_check_sync(struct applier *applier)
  * but doesn't start it.
  */
 static struct fiber *
-applier_fiber_new(struct applier *applier, const char *name, fiber_func func)
+applier_fiber_new(struct applier *applier, const char *name, fiber_func func,
+		  bool is_joinable)
```
grafin, line 174: IMO this patch should be split into 2 commits: one adding is_joinable argument to applier_fiber_new, second changing fiber_system_new to applier_fiber_new for ballot_watcher in applier_connect.

grafin (changes requested):
Thanks for the patch! Consider splitting this patch into two commits, more info in comment.

src/box/applier.cc
sergepetrenko: I thought the change was too tiny to split it into 2 commits, but I don't mind. Done.

grafin (approved):
Thanks for quick fixes! LGTM.

src/box/applier.cc
```diff
@@ -699,8 +700,9 @@ applier_connect(struct applier *applier)
 	if (method_default == NULL)
 		method_default = AUTH_METHOD_DEFAULT;
 
-	applier->ballot_watcher = fiber_new_system("applier_ballot_watcher",
-						   applier_ballot_watcher_f);
+	applier->ballot_watcher = applier_fiber_new(applier, "ballot_watcher",
+						    applier_ballot_watcher_f,
+						    false);
```
Gerold103, line 705: Need to cover this with a test!

Gerold103 (changes requested):

Gerold103 (approved):
Nah, just kidding :)

PR #8199, opened 2023-01-23
ci: remove libtool from build requirements

nshy:
The commit 5b08d71ab474 ("libunwind: use latest release v1.6.2 as a base") started using autotools for building bundled libunwind. In particular it added libtool as in CI pipeline build requirements. We need libtool because of use `make install` in ExternalProject_Add macro.

Later we switch to our own macro in the commit 14f93aee86b8 ("libunwind: improve incremental build/rebuild") which do not use `make install`. So let's drop libtool dependency.

Follow-up #5665

NO_TEST=ci
NO_DOC=ci
NO_CHANGELOG=ci

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -33,7 +33,7 @@ Build-Depends: cdbs (>= 0.4.100), debhelper (>= 9), dpkg-dev (>= 1.16.1~),
 # needed for datetime tests
  tzdata,
 # For third-party subprojects that are built with autotools.
- autoconf, automake, libtool,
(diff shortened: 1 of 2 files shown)


nshy:
With this PR I wanted to test whether we still need libtool after https://github.com/tarantool/tarantool/commit/14f93aee86b81d21d207f669e29983f181a8c179 ("libunwind: improve incremental build/rebuild") or not. Unfortunately we do.

PR #8200, opened 2023-01-23
test: add test for #7394 (`space:count` bug in memtx MVCC)

### LINKED ISSUE #7394: `space:count()` with mvcc returns not the same as `space:select()`
Here is what I observe in one of vshard tests exclusively when mvcc is enabled:
```
578: ilt.assert_equals(s:select{}, {{1, 1}, {2, 2}, {3, 1}, {4, 2}})
579: ilt.assert_equals(s:count(), 4)
```
the first line passes, the second one doesn't. I couldn't reproduce it on bare tarantool with some simple script, sorry. But it reproduces 100% on this vshard branch: https://github.com/tarantool/vshard/tree/gerold103/mvcc-count-bug if you run `python test-run.py storage-luatest/garbage_collector_test.lua`. The result file shows:
```
not ok 13	bucket_gc.engine:"memtx".memtx_use_mvcc_engine:true.test_unit_bucket_delete_garbage
#   ...s/vshard/test/storage-luatest/garbage_collector_test.lua:579: expected: 4, actual: 3
```
So as you can see, `count()` returns less than `select()` at line 579 of the test. There are no yields between those lines. It didn't fail at the line 578 which checks 4 tuples. Tarantool version was the latest at the moment of doing the test: `2.11.0-entrypoint-232-ged629224f`.

The unusual things which I have here are `on_replace` triggers in `_bucket` space which install an `on_commit` trigger. Could be related. Also the bug doesn't happen without the last commit of the branch. Can't imagine how could it affect the behaviour though.

CuriousGeorgiy:
Closes #7394

# :bangbang: NOTA BENE :bangbang:
If you want to verify that this test fails, you need to checkout commit 88203d4f or older.

# :bangbang: NOTA BENE :bangbang:

### DIFF
--- a/test/box-luatest/gh_7394_memtx_mvcc_space_cnt_bug_test.lua
+++ b/test/box-luatest/gh_7394_memtx_mvcc_space_cnt_bug_test.lua
@@ -1,0 +1,38 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new {
+        alias   = 'dflt',
+        box_cfg = {memtx_use_mvcc_engine = true}
+    }
+    cg.server:start()
+    cg.server:exec(function()
+        local s = box.schema.create_space('s')
+        s:create_index('pk')
+        box.internal.memtx_tx_gc(100)
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+-- Checks that memtx transaction manager MVCC invariant violation does not lead
+-- to incorrect space count (see also gh-7490 test).
+g.test_memtx_mvcc_space_cnt_with_invariant_violation = function(cg)
+    cg.server:exec(function()
+        local t = require('luatest')
+
+        for _ = 1, 5 do
+            box.begin()
+            box.space.s:replace{1}
+            box.space.s:replace{2}
+            box.commit()
+        end
+
+        t.assert_equals(box.space.s:count(), #box.space.s:select{})
+    end)
+end


drewdzzz (commented):
Is there a way to delete only top stories? Maybe a handle that sets gc in the beginning of story list would be helpful? Then the test would be more reliable - now it depends on gc constants AFAIU.

test/box-luatest/gh_7394_memtx_mvcc_space_cnt_bug_test.lua
```diff
@@ -0,0 +1,38 @@
+local server = require('luatest.server')
```
Gerold103, line 1: The test seem to belong to `engine-luatest/`, because it is memtx-specific. Otherwise looks good.

Gerold103 (approved):
Thanks for the patch!

drewdzzz:
I think that this test needs randomization - let's make it nondeterministic to always cover the case, even with other gc parameters

test/box-luatest/gh_7394_memtx_mvcc_space_cnt_bug_test.lua
CuriousGeorgiy: I guess you are right, but we traditionally add memtx transaction manager tests to `box` suites, so I think we should stick to it. BTW, this test can actually be made multi-engine.

alyapunov (approved):

alyapunov:
CI alpine is always read, but it should not stop us.

alyapunov:
Cherry-picked to 2.10.

Issue #8051, opened 2022-12-13
`log_modules` option can not be set thru env var on box.cfg{}


nshy:
`box.cfg{}` [currently can read](https://github.com/tarantool/tarantool/blob/435e5efe982025233977effa3dab27e729c57c8a/src/box/lua/load_cfg.lua#L968) only arrays from env variable if option type is `table`:

```lua
    if param_type:find('table') and raw_value:find(',') then
        assert(not param_type:find('boolean'))
        local res = {}
        for i, v in ipairs(raw_value:split(',')) do
            res[i] = tonumber(v) or v
        end
        return res
```
Thus `log_modules`  option for example can not be set thru env var:

```console
$ TT_LOG_MODULES=aaa=info,bbb=error ./build-dev/src/tarantool
Tarantool 2.11.0-entrypoint-779-g5784a14ed
type 'help' for interactive help
*tarantool> box.cfg{}
---
- error: 'Incorrect value for option ''module name'': should be of type string'
...
 tarantool>
```

Also there is a different issue with setting table with single value:

```console
$ TT_LOG_MODULES=aaa=info ./build-dev/src/tarantool
Tarantool 2.11.0-entrypoint-779-g5784a14ed
type 'help' for interactive help
 tarantool> box.cfg{}
---
- error: 'builtin/box/load_cfg.lua:20: builtin/box/load_cfg.lua:1001: assertion failed!'
...
 tarantool>
```

I expect in both cases `log_modules` should be set successfully to `{aaa="info", bbb="error"}` and `{aaa="info"}` respectively.

Mons:
In the #5602 there were no support for key-value paramaters, since Tarantool didn't contain any nested key-value options earlier.
So, key-value should also be supported.

To maintain simplicity for basic options and ability to pass complex options we should allow passing such an options with JSON.
So, the rules should be the following:

If param_type is a `table` and first symbol of value is `{` or `[`, then parse it as JSON.

And after splitting by "`,`":
If the param is a table and the value contains `=`, then split value into two parts by splitting it by first =

PR #8203, opened 2023-01-23
Embed metrics

### LINKED ISSUE #7725: Embed tarantool/metrics in core tarantool
The series of tickets about tarantool/metrics embedding is just a beginning of the discussion started with possible Product Team plans related to Flight Recorder development. Expected features may be reevaluated, estimation is preliminary.

[tarantool/metrics](https://github.com/tarantool/metrics) is a Lua+ffi module to collect and expose [Tarantool](https://tarantool.io/)-based applications metrics.

The ability to collect metrics in a single place and expose them is useful. It seems that it could be even more useful if module would be embedded to core tarantool. It could decrease the inhomogeneity in core tarantool handles and metrics exported with the module, remove the time lag between releasing new observation handles in core and supporting them in tarantool/metrics. Some potential features like exposing metrics in a separate thread are more likely to implement in core tarantool rather then separate module. It could also simplify integrating another Lua modules with `metrics`. At last, integrating `metrics` with FlightRec would be possible.

On the other hand, it is likely to also bring several inconveniences. Providing new versions would be much harder since now new metrics release would be bound to core Tarantool release. (It is partially would be covered by new `tarantool/metrics-extensions` module, see #7727). It is also a drawback that core tarantool would became a much complicated monolith. Moreover, if it is planned to enable metrics by default (as we discussed), it would affect performance.

Integrating `metrics` with existing core features may vary from most simplest (just put a bunch of Lua code to the core) to time-consuming. For example, most default metrics handles is actually about extracting something from core handle call and putting it to a  Lua table. Since now the original values are in our hands, it could be optimized in some way. Global metrics registry may also be reworked to be more efficient. Not only internals could be reworked -- external API is highly likely to get reworked too. For example, it seems expected to expose some configuration to `box.cfg`, especially if we will enable metrics collect by default. On the other hand, we shouldn't remove old handles to support compatibility with applications that uses `metrics` as a module. If it is planned to use `metrics` as a replacement for `box.info` and many other calls, additional handles would be needed.

One may also want to separate `cartridge`-related features from the module.

Embedding [tarantool/checks] is also a prerequirement to embedding `metrics`: it is a rock dependency.

**Definition of Done**

Minimal: any application that uses the module can use `require('metrics')` and another features without any additional rocks installation.

Additional: internal optimization, convenient API for configuration.

DifferentialOrange:
tarantool/metrics [1] is a lua module (distributed as a separate rock) for metrics aggregation and export. After this patch, it will be a part of the tarantool binary.

`metrics.cfg` is a module configuration entrypoint [2]. After this patch, `box.cfg{metrics = cfg}` triggers `metrics.cfg`. Parameter is dynamic.

This patch also enables all metrics by default. To disable them, use `box.cfg{metrics = {include = 'none'}}`. Since callbacks are triggered manually (when user calls `invoke_callbacks()` or `collect{invoke_callbacks=true}`), it should not affect the performance.

1. https://github.com/tarantool/metrics
2. https://www.tarantool.io/en/doc/latest/book/monitoring/api_reference/#lua-function.metrics.cfg

Closes #7725


src/box/lua/load_cfg.lua
```diff
@@ -377,6 +383,8 @@ local template_cfg = {
     net_msg_max           = 'number',
     sql_cache_size        = 'number',
     txn_timeout           = 'number',
+
+    metrics = 'table',
```
DifferentialOrange, line 387: Using hierarchic cfg instead of flattened one (like in log, flightrec opts) was proposed here: https://github.com/tarantool/tarantool/issues/7725#issuecomment-1410045326

test/metrics-luatest/helper.lua
```diff
@@ -0,0 +1,13 @@
+-- Assert running tests with built-in metrics.
+local rock_utils = require('third_party.metrics.test.rock_utils')
+rock_utils.assert_builtin('metrics')
```
Totktonada, line 3: Will the tests fail if a recent metrics module (with built-in overriding support, after https://github.com/tarantool/metrics/pull/440) is installed into the system?

test/metrics-luatest/helper.lua
```diff
@@ -0,0 +1,13 @@
...
+
+-- Workaround paths to reuse existing submodule tests.
+local function workaround_requires(path)
+    package.preload[path] = function()
+        return require('third_party.metrics.' .. path)
+    end
+end
+
+workaround_requires('test.utils')
+workaround_requires('test.psutils_linux_test_payload')
```
Totktonada, line 13: Raw idea: can we walk over `test/metrics-luatest/*_test.lua` and `third_party/metrics/test/**/*_test.lua` and perform some check that all the tests are included into tarantool? We can write the checks as a test, for example. Now it looks easy to miss a new test at the submodule update.

test/metrics-luatest/helper.lua
DifferentialOrange: Well, yeah. Let's workaround it.

DifferentialOrange: Added handle call and comment

DifferentialOrange:
Metrics 0.17.0 has been released, PR has been rebased. No new updates has planned.

test/metrics-luatest/helper.lua
Totktonada: Thanks!

DifferentialOrange:
See https://github.com/tarantool/tarantool/issues/8494 and https://github.com/tarantool/tarantool/issues/8051 about possible `TT_METRICS` support

Totktonada (approved):
Thank you for the patches!

I looked visually across the new code. It seems logical. I didn't perform manual testing.

LGTM.

Issue #7739, opened 2022-09-29
`engine-luatest/pagination_test.lua` fails on aarch64 if Lua JIT is enabled


locker:
Tarantool version: 2.11.0-entrypoint-531-ga6766b9c3946

[`engine-luatest/pagination_test.lua`](https://github.com/tarantool/tarantool/blob/master/test/engine-luatest/pagination_test.lua) randomly fails on aarch64 CI workflows. All other CI workflows do fine. The test was added by https://github.com/tarantool/tarantool/pull/7632. Disabling JIT with `require('jit').off()` eliminates the failures so it was decided to disable JIT in this test only when run on aarch64. To reproduce the issue, remove the [corresponding code block](https://github.com/tarantool/tarantool/blob/ec1a71ffed2aa506f693c991b9ca81387a4a238a/test/engine-luatest/pagination_test.lua#L9-L15) and run the test on any aarch64 CI workflow.

<details><summary>failure log</summary>

```
[029] engine-luatest/pagination_test.lua                              [ fail ]
[029] Test failed! Output from reject file /tmp/t/rejects/engine-luatest/pagination.reject:
[029] TAP version 13
[029] 1..25
[029] # Started on Wed Sep 28 21:48:18 2022
[029] # Starting group: Memtx tree func index tests
[029] not ok 1	Memtx tree func index tests.test_func_index_pagination
[029] #   ...trypoint.534.dev/test/engine-luatest/pagination_test.lua:617: expected: {[6, 6], [7, 6]}
[029] #   actual: {[8, 6], [9, 6]}
[029] #   stack traceback:
[029] #   	...trypoint.534.dev/test/engine-luatest/pagination_test.lua:561: in function 'Memtx tree func index tests.test_func_index_pagination'
[029] #   	...
[029] #   	[C]: in function 'xpcall'
[029] not ok 2	Memtx tree func index tests.test_func_multikey_index_pagination
[029] #   ...trypoint.534.dev/test/engine-luatest/pagination_test.lua:693: expected: {[1, 1], [1, 1]}
[029] #   actual: {[2, 2], [3, 3]}
[029] #   stack traceback:
[029] #   	...trypoint.534.dev/test/engine-luatest/pagination_test.lua:653: in function 'Memtx tree func index tests.test_func_multikey_index_pagination'
[029] #   	...
[029] #   	[C]: in function 'xpcall'
[029] # Starting group: Tree index tests.disable_ffi:false.engine:"memtx"
[...]

PR #8206, opened 2023-01-24
box: make ffi select work on M1

### LINKED ISSUE #7946: Select FFI does not work on M1
If we call select on memtx space (or another function that calls `select_ffi` under the hood) on M1, it will throw an error `NYI: cannot call this C function (yet)`, just try this simple script:
```lua
box.cfg{}
s = box.schema.create_space('test')
s:create_index('pk')
s:replace{0, 0}
s:select{}
```
The snippet from `schema.lua` that throws an error:
```lua
nok = builtin.box_select(index.space_id, index.id, iterator, offset, limit,
                         key, key_end, iterator_pos, iterator_pos_end,
                         fetch_pos, port) ~= 0
``` 
The problem is in LuaJIT: https://github.com/LuaJIT/LuaJIT/issues/205
We can fix the issue or simply create a wrapper with reordered arguments for `box_select`.

drewdzzz:
Recently, ffi select was broken on M1 - it turned out that ffi on M1 poorly supports a big quantity of arguments, which was used for pagination. Fortunately, there is a workaround - we can pass only 64-bit integer arguments beyond the 8th argument. Let's do it.

Closes #7946

### DIFF
--- a/extra/exports
+++ b/extra/exports
@@ -101,7 +101,7 @@ box_replace
 box_return_mp
 box_return_tuple
 box_schema_version
-box_select
+box_select_ffi
 box_sequence_current
 box_sequence_next
 box_sequence_reset
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -118,11 +118,11 @@ ffi.cdef[[
                              const char **pos, const char **pos_end);
 
     int
-    box_select(uint32_t space_id, uint32_t index_id,
-               int iterator, uint32_t offset, uint32_t limit,
-               const char *key, const char *key_end,
-               const char **after, const char **after_end,
-               bool update_pos, struct port *port);
+    box_select_ffi(uint32_t space_id, uint32_t index_id, const char *key,
+                   const char *key_end, const char **packed_pos,
+                   const char **packed_pos_end, bool update_pos,
+                   struct port *port, int64_t iterator, uint64_t offset,
+                   uint64_t limit);
 
     enum priv_type {
         PRIV_R = 1,
@@ -2537,9 +2537,9 @@ base_index_mt.select_ffi = function(index, key, opts)
     check_select_safety(index, key_is_nil, iterator, limit, offset, fullscan)
     local region_svp = builtin.box_region_used()
     normalize_position(index, after, iterator_pos, iterator_pos_end)
-    nok = builtin.box_select(index.space_id, index.id, iterator, offset, limit,
-                             key, key_end, iterator_pos, iterator_pos_end,
-                             fetch_pos, port) ~= 0
+    nok = builtin.box_select_ffi(index.space_id, index.id, key, key_end,
+                                 iterator_pos, iterator_pos_end, fetch_pos,
+                                 port, iterator, offset, limit) ~= 0
     if not nok and fetch_pos and iterator_pos[0] ~= nil then
         new_position = ffi.string(iterator_pos[0],
                                   iterator_pos_end[0] - iterator_pos[0])
--- a/static-build/test/exports.test.lua
+++ b/static-build/test/exports.test.lua
@@ -114,7 +114,6 @@ local check_symbols = {
     'coio_getaddrinfo',
     'luaT_call',
     'box_txn',
-    'box_select',
     'clock_realtime',
     'string_strip_helper',
 
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3075,6 +3075,22 @@ fail:
 	return -1;
 }
 
+/**
+ * A special wrapper for FFI - workaround for M1.
+ * Use 64-bit integers beyond the 8th argument.
+ * See https://github.com/LuaJIT/LuaJIT/issues/205 for details.
+ */
+extern "C" int
+box_select_ffi(uint32_t space_id, uint32_t index_id, const char *key,
+	       const char *key_end, const char **packed_pos,
+	       const char **packed_pos_end, bool update_pos, struct port *port,
+	       int64_t iterator, uint64_t offset, uint64_t limit)
+{
+	return box_select(space_id, index_id, iterator, offset, limit, key,
+			  key_end, packed_pos, packed_pos_end, update_pos,
+			  port);
+}
+
 API_EXPORT int
 box_insert(uint32_t space_id, const char *tuple, const char *tuple_end,
 	   box_tuple_t **result)


drewdzzz:
I don't have my own M1, so @Gumix will check manually if the patch is OK.

drewdzzz:
If you don't like this soultion, I came up with another idea - we can simply disable ffi on M1.

src/box/lua/schema.lua
```diff
@@ -2537,9 +2537,9 @@ base_index_mt.select_ffi = function(index, key, opts)
     check_select_safety(index, key_is_nil, iterator, limit, offset, fullscan)
     local region_svp = builtin.box_region_used()
     normalize_position(index, after, iterator_pos, iterator_pos_end)
-    nok = builtin.box_select(index.space_id, index.id, iterator, offset, limit,
-                             key, key_end, iterator_pos, iterator_pos_end,
-                             fetch_pos, port) ~= 0
+    nok = builtin.box_select_ffi(index.space_id, index.id, key, key_end,
```
locker, line 2540: Is there a ticket for this? #7739? Should we re-enable LuaJIT fro the tests disabled in the scope of this issue?

https://github.com/tarantool/tarantool/blob/e02bc87feba73ce42e00719f1d40f2c0d363dc72/test/engine-luatest/pagination_test.lua#L10-L19

Should we file a ticket to revert this change once the LuaJIT issue is fixed and reference it in the code?

Anyway, I think that this fix should be reviewed by someone from the Lango team.

Gumix: No, #7739 is different. This workaround is for #7946

Gumix (commented):
I can confirm that it fixes #7946 on Apple M1.

src/box/lua/schema.lua
drewdzzz: This PR will close #7946

drewdzzz: > Anyway, I think that this fix should be reviewed by someone from the Lango team.

Requested a review from @igormunkin

extra/exports
```diff
@@ -101,7 +101,7 @@ box_replace
 box_return_mp
 box_return_tuple
 box_schema_version
-box_select
```
Gumix, line 104: Shouldn't we keep the old version for compatibility?

locker: `box_select` isn't a part of the public API so I guess not.

Gumix: It is present here:
https://github.com/tarantool/tarantool/blob/ca58d6c92a10c4ab80f07425c51f20225e5611e8/static-build/test/exports.test.lua#L104-L119

locker: Maybe by mistake, but we need to check if it's actually used by any of our modules (e.g. tdg).

drewdzzz: We've changed `box_select` signature (added arguments for pagination), so the compatibility with old `box_select` is already broken.

drewdzzz: I've searched for its usage across `tarantool` organization - it is not used outside the core part.

src/box/lua/schema.lua
igormunkin: > Should we re-enable LuaJIT fro the tests disabled in the scope of this issue?

JIT engine is unrelated for this case. But I wonder, whether we can run all M1 tests with JIT disabled... I'll check if there is nothing left related to JIT for M1 in the queue. All in all, one can just add `-e jit.off() jit.flush()` to the command launching the Tarantool.

> Should we file a ticket to revert this change once the LuaJIT issue is fixed and reference it in the code?

Up to you. Then there will be `N+1` M1-related tickets on GitHub. I have no idea when they are scheduled to RnD plans. Furthermore, I'd rather believe Mike Pall regarding the potential fix for this problem (without M1 ABI spec that is more verbose than [this one](https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms)).

extra/exports
igormunkin: Agree with @locker here.

igormunkin (approved):
@drewdzzz, thanks for the fix! LGTM.

drewdzzz:
Fix static build test:
```diff
diff --git a/static-build/test/exports.test.lua b/static-build/test/exports.test.lua
index 0f6fef439..c503ab53d 100755
--- a/static-build/test/exports.test.lua
+++ b/static-build/test/exports.test.lua
@@ -81,6 +81,7 @@ local check_symbols = {
     'space_run_triggers',
     'space_bsize',
     'box_schema_version',
+    'box_select_ffi',

     'crypto_ERR_error_string',
     'crypto_ERR_get_error',
@@ -114,7 +115,6 @@ local check_symbols = {
     'coio_getaddrinfo',
     'luaT_call',
     'box_txn',
-    'box_select',
     'clock_realtime',
     'string_strip_helper',
```

extra/exports
drewdzzz: AFAIK, `box_select` was in public API until it became to use `port`.

static-build/test/exports.test.lua
```diff
@@ -81,6 +81,7 @@ local check_symbols = {
     'space_run_triggers',
     'space_bsize',
     'box_schema_version',
+    'box_select_ffi',
```
locker, line 84: This function isn't public so it shouldn't be checked. Please remove.

src/box/box.h
```diff
@@ -372,6 +372,17 @@ box_select(uint32_t space_id, uint32_t index_id,
...
+ * A special wrapper for FFI - workaround for M1.
+ * Use 64-bit integers beyond the 8th argument.
+ * See https://github.com/LuaJIT/LuaJIT/issues/205 for details.
+ */
+int
+box_select_ffi(uint32_t space_id, uint32_t index_id, const char *key,
+	       const char *key_end, const char **packed_pos,
+	       const char **packed_pos_end, bool update_pos, struct port *port,
+	       int64_t iterator, uint64_t offset, uint64_t limit);
+
```
locker, line 385: There's no need to add a function declaration to the header if it'll only be used via Lua FFI. Please remove.

locker (changes requested):

static-build/test/exports.test.lua
drewdzzz: But there is plenty of function that aren't part of public API. I see that all symbols are checked, even ones that needed for FFI only.

src/box/box.h
drewdzzz: Done
<details>
<summary> diff </summary>

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 09ad27010..362b3df02 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3075,7 +3075,12 @@ fail:
        return -1;
 }

-int
+/**
+ * A special wrapper for FFI - workaround for M1.
+ * Use 64-bit integers beyond the 8th argument.
+ * See https://github.com/LuaJIT/LuaJIT/issues/205 for details.
+ */
+extern "C" int
 box_select_ffi(uint32_t space_id, uint32_t index_id, const char *key,
               const char *key_end, const char **packed_pos,
               const char **packed_pos_end, bool update_pos, struct port *port,
diff --git a/src/box/box.h b/src/box/box.h
index 8d53aaba2..3661322c2 100644
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -372,17 +372,6 @@ box_select(uint32_t space_id, uint32_t index_id,
           const char **packed_pos, const char **packed_pos_end,
           bool update_pos, struct port *port);

-/**
- * A special wrapper for FFI - workaround for M1.
- * Use 64-bit integers beyond the 8th argument.
- * See https://github.com/LuaJIT/LuaJIT/issues/205 for details.
- */
-int
-box_select_ffi(uint32_t space_id, uint32_t index_id, const char *key,
-              const char *key_end, const char **packed_pos,
-              const char **packed_pos_end, bool update_pos, struct port *port,
-              int64_t iterator, uint64_t offset, uint64_t limit);
-
 /** \cond public */

 /*
 ```
 
 </details>

static-build/test/exports.test.lua
locker: I assume they're checked because they are used by some tdg modules (static build was initially imported from tarantool-enterprise). Apparently, `box_select_ffi` can't be used by any modules, and we don't want it to ever be used so we don't need to check it.

drewdzzz: OK, removed it

locker (approved):

PR #8208, opened 2023-01-25
box/lua: disallow changing box options directly

### LINKED ISSUE #2867: Raise on raw modifications of box.cfg values
```sh
tarantool
> box.cfg{}
> box.cfg.read_only
false
> box.cfg.read_only = true
> box.cfg.read_only
true
```

I'd expect 3rd command should raise and cfg table at all would be readonly

Updated (25.01.2023): RFC - https://www.notion.so/Raw-modifications-of-box-cfg-values-9f440d321f984389a22bcb3f81366ea7

ligurio:
Tarantool has a special table 'box.cfg' that includes configuration
parameters. User could view that table, but it's direct modification has
no effect - after assigning a new value it is actually "updated" but
actual value remains the same. Such behaviour is a counterintuitive for
our users and provides a bad experience.

Proposed patch change this behaviour: new value assigned to parameter
via direct access to table box.cfg raise an error.

Before the patch:
```
tarantool> box.cfg{}
<snipped>
tarantool> box.cfg.read_only=true
---
...

tarantool>
```

After the patch:

```
tarantool> box.cfg{}
<snipped>
tarantool> box.cfg.read_only=true
---
- error: 'builtin/box/load_cfg.lua:973: Use function box.cfg for modifications'
...

tarantool>
```

NO_DOC=bugfix

Closes #2867


Gumix:
Hi, here is something wrong with the validation of the options:
```lua
tarantool> box.cfg{}
-- skipped --

tarantool> box.cfg.log_format = 'xxx'
---
- error: 'Incorrect value for option ''log_format'': expected ''plain'' or ''json'''
...

tarantool> box.cfg.log_format = 'xxx'
---
...

tarantool> box.cfg.log_format = 'yyy'
---
...

tarantool> box.cfg.log_format
---
- yyy
...

tarantool> box.cfg.log_level = 6
---
- error: 'Incorrect value for option ''log_format'': expected ''plain'' or ''json'''
...
-- ??? why log_format?
```

Here is another issue, `box.cfg.log_level` is updated correctly, while `box.cfg` still prints the old value:
```lua
tarantool> box.cfg{}
-- skipped --

tarantool> box.cfg.log_level
---
- 5
...

tarantool> box.cfg.log_level = 6
2023-01-25 10:53:43.764 [339315] main/103/interactive/box.load_cfg I> set 'log_level' configuration option to 6
---
...

tarantool> box.cfg.log_level
---
- 6
...

tarantool> box.cfg
---
- replication_sync_timeout: 0
  -- skipped --
  log_level: 5
  -- skipped --
```

igormunkin:
@Gumix, nice catch, thanks for noticing! By the way, I've added you to reviewers list, since we touch box-related interface. I believe you're the best candidate, 'cause you're the only one from Core team, who has already touched the patch :)

src/box/lua/load_cfg.lua
```diff
@@ -962,13 +964,40 @@ local function load_cfg(cfg)
...
+            __index = raw_cfg,
+            __newindex = function(self, key) -- luacheck: no unused args
+                if template_cfg[key] ~= nil then
+                    error('Use function box.cfg for modifications')
+                end
                 error('Attempt to modify a read-only table')
             end,
-            __call = locked(reload_cfg),
+            __call = function(self, ...)
+                -- assert(next(proxy_table) ~= nil,
```
Gumix, line 994: Commented-out code.

Gumix (approved):
OK for me, but I want to invite @nshy

test/box-luatest/gh_2867_raise_on_raw_box_cfg_modifications_test.lua
```diff
@@ -0,0 +1,45 @@
...
+    local ok, err = g.server:eval([[
+        return pcall(function() box.cfg.LANGO_TEAM=true end)]])
+    t.assert_equals(ok, false)
+    local err_msg = "Attempt to modify a read-only table"
+    t.assert_str_contains(tostring(err), err_msg)
+    t.assert_equals(g.server:eval("return box.cfg.LANGO_TEAM"), nil)
+end
+
+g.test_box_cfg_set_nil = function()
+    t.skip("Skipped")
```
nshy, line 38: Do we need to drop this this test as it is unconditionally skipped?

src/box/lua/load_cfg.lua
```diff
@@ -962,13 +964,40 @@ local function load_cfg(cfg)
...
+                --        "FATAL: length of proxy table is not zero")
+                return locked(reload_cfg)(raw_cfg, ...)
+            end,
+            __serialize = function()
+                return raw_cfg
+            end,
```
nshy, line 1000: Great we finally going to fix this issue! I have a couple of concerns though.

1. We have config options of type table. Examples are `wal_ext` or `log_level_modules`. Can we make table proxies recursively so that changing nested tables directly is also prohibited?

2. We probably better make same proxy for log.cfg in this patch or this patch series too. log.cfg have same issue.

nshy (changes requested):

src/box/lua/load_cfg.lua
nshy: Looks like we can just apply same approach with empty proxy table recursively on box.cfg. This way we make box.cfg readonly on all levels. Then we can put this functionality info function and apply to log.cfg too.

nshy: Discussed f2f and decided to create a new task for the above issues.

nshy (approved):

test/box-luatest/gh_2867_raise_on_raw_box_cfg_modifications_test.lua
```diff
@@ -0,0 +1,35 @@
...
+    local ok, err = g.server:eval([[
+        return pcall(function() box.cfg.read_only=true end)]])
+    t.assert_equals(ok, false)
+    local err_msg = "Use function box.cfg for modifications"
+    t.assert_str_contains(tostring(err), err_msg)
+    t.assert_equals(g.server:eval("return box.cfg.read_only"), false)
+end
+
+g.test_box_cfg_set_unknown_option = function()
+    t.assert_equals(g.server:eval("return box.cfg.LANGO_TEAM"), nil)
```
igormunkin, line 28: Side note: nice "unknown" option ;)

src/lua/init.lua
```diff
@@ -31,6 +31,18 @@ local LUA_TEMPLATES = { '?.lua', '?/init.lua' }
 local ROCKS_LIB_TEMPLATES = { ROCKS_LIB_PATH .. '/?.'..soext }
 local ROCKS_LUA_TEMPLATES = { ROCKS_LUA_PATH .. '/?.lua', ROCKS_LUA_PATH .. '/?/init.lua' }
 
+-- A proxy for pairs function that allows to use pairs for nested table in
+-- box.cfg.
+local builtin_pairs = rawget(_G, 'pairs')
+local function patched_pairs(object)
+    local mt = getmetatable(object)
+    local iterable = (mt and mt.__name == 'box_cfg') and
+                     debug.getmetatable(object).__index or
```
igormunkin, line 40: Well, this part is not quite right.

1. It's unsafe, since `getmetatable` yields the value of `__metatable` field if one is set (that can be e.g. string[^string]). You need either to check the type of `mt` variable or to use `debug.getmetatable` to obtain `object`'s metatable. I'd rather use the latter approach.
[^string]: Fortunately, all Lua strings share the metatable with overloaded index, so `mt.__name` in case of strings just yields `nil`, but anyway in general case this is not safe.
2. There is no need in fetching `object`'s metatable one more time in the line above (it's already stored in `mt`, isn't it?) so just use `mt.__index` for this.

src/box/lua/load_cfg.lua
```diff
@@ -962,13 +964,40 @@ local function load_cfg(cfg)
             box[k] = v
         end
     end
+
     setmetatable(box, nil)
-    box.cfg = setmetatable(cfg,
+    -- There is only a single way for setting a value to a box option -
+    -- call 'box.cfg{ option_name=option_value }'. Direct modifications
+    -- of table box.cfg are not allowed. On attempt to make a direct
+    -- modifications (setting a new option or updating an existed one) we raise
```
igormunkin, line 972: Typo: s/existed/existing/.

src/box/lua/load_cfg.lua
```diff
@@ -962,13 +964,40 @@ local function load_cfg(cfg)
...
+            -- tarantoolctl (extra/dist/tarantoolctl.in) expects a metatable
+            -- with metamethods, not a string. Thus we couldn't use __metatable
+            -- instead.
+            __name = 'box_cfg',
+            __index = raw_cfg,
+            __newindex = function(self, key) -- luacheck: no unused args
+                if template_cfg[key] ~= nil then
+                    error('Use function box.cfg for modifications')
+                end
                 error('Attempt to modify a read-only table')
```
igormunkin, line 991: Why do we need to differ these two cases? I believe `box.cfg` modification is forbidden, whether the key exists in the table or not.

src/box/lua/load_cfg.lua
```diff
@@ -801,7 +801,7 @@ local function reconfig_modules(module_keys, oldcfg, newcfg, log_basecfg)
             oldvals = {}
             for key in pairs(keys) do
                 oldvals[key] = oldcfg[key]
-                rawset(oldcfg, key, newcfg[key])
+                oldcfg[key] = newcfg[key]
```
igormunkin, line 804: Why all these three changes above are required? AFAIU, `oldcfg` parameter is `raw_cfg` as a result of your patch and this table has no metatable, hence the change has no effect. However, the code in `load_cfg.lua` is not quite clear, so I might be wrong.

src/box/lua/load_cfg.lua
```diff
@@ -1091,7 +1120,7 @@ end
 -- log.cfg of log module for example).
 local function update_cfg(option, value)
     if box_is_configured then
-        rawset(box.cfg, option, value)
+        raw_cfg[option] = value
```
igormunkin, line 1123: The same as for the three similar hunks above.

igormunkin (changes requested):
@ligurio, thanks for the patch! Please consider my comments below. 

Furthermore, there are two confusing `NOWRAP` entries in the commit message: if these are just typos, please fix it either.

I also guess that adding a docbot request will be nice, since such behaviour change should be added to the docs.

Totktonada:
I glanced over the patch out of curiosity and would give a few suggestions about the code layout.

* I would move our own `pairs()` into its own module src/lua/pairs.lua (just like we do for `print()`). src/lua/init.lua is bloated with many things unrelated to each other and I think that it should be decomposed in a future anyway.
* I would eliminate the `__name` field and would the tables itself as keys in a weak table... well, it is easier to explain with code:

```lua
-- pairs.lua
local M = {}

M.raw_pairs = pairs

-- table -> iterator generator
M.tracked = setmetatable({}, {__mode = 'k'})

_G.pairs = function(t)
    local iter_gen = M.tracked[t] or M.raw_pairs
    return iter_gen(t)
end

function M.add(t, iter_gen)
    M.tracked[t] = iter_gen
end

function M.remove(t)
    M.tracked[t] = nil
end

return M
```

Usage:

```lua
local internal_pairs = require('internal.pairs')
local real_table = {}
local proxy_table = {_iterate_over = real_table}
internal_pairs.add(proxy_table, function(self)
    return pairs(self._iterate_over)
end)

-- or

local internal_pairs = require('internal.pairs')
local real_table = {}
local proxy_table = {}
internal_pairs.add(proxy_table, function(_self)
    return pairs(real_table)
end)
```

Feel free to ignore, I just pass here over.

src/box/lua/load_cfg.lua
nshy: `rawset` is not required anymore because internally we start to use `raw_cfg`. So I would drop `rawset` too.

src/box/lua/load_cfg.lua
igormunkin: Yeah, nice change, thanks.

src/box/lua/load_cfg.lua
igormunkin: Great, thanks!

src/box/lua/load_cfg.lua
igormunkin: Yeah, that one is definitely required. Anyway, I see the second patch totally removes `rawset` from `load_cfg.lua`, and that's neat. Thanks!

src/box/lua/load_cfg.lua
```diff
@@ -1011,13 +1013,45 @@ local function load_cfg(cfg)
...
-            __newindex = function(table, index) -- luacheck: no unused args
+            -- tarantoolctl (extra/dist/tarantoolctl.in) expects a metatable
+            -- with metamethods, not a string. Thus we couldn't use __metatable
+            -- instead.
+            __name = 'box_cfg',
+            __index = raw_cfg,
+            __newindex = function(self, key, value) -- luacheck: no unused args
+                if template_cfg[key] ~= nil then
+                    local v = '<...>'
+                    if type(value) ~= 'table' then
```
igormunkin, line 1039: Minor: The only thing that bothers me here is the condition.

Is it possible to pass other "compex" types (such as cdata/udata/function) as a `box.cfg` parameter value? If not, then everything is fine; if there are doubts we can just leave it as is and fix if/when one is found: yeah, it's bad when there is garbage in the error message, but not critical.

Feel free to ignore.

igormunkin (approved):
@ligurio, thanks for all changes made! LGTM now (but I'll wait for your answer for my question below prior to merging the PR).

PR #8213, opened 2023-01-25
Revert "box: forbid DDL operations until box.schema.upgrade"

Gumix:
https://www.notion.so/tarantool/box-forbid-DDL-operations-until-box-schema-upgrade-f753883922a34159bcb94c6b33c13696


src/box/errcode.h
```diff
@@ -300,23 +300,22 @@ struct errcode_record {
...
+	/*254 */_(ER_WRONG_SESSION_TYPE,	"Session '%s' is not supported") \
+	/*255 */_(ER_PASSWORD_EXPIRED,		"Password expired") \
+	/*256 */_(ER_AUTH_DELAY,		"Too many authentication attempts") \
+	/*257 */_(ER_AUTH_REQUIRED,		"Authentication required") \
+	/*258 */_(ER_SQL_SEQ_SCAN,		"Scanning is not allowed for %s") \
+	/*259 */_(ER_NO_SUCH_EVENT,		"Unknown event %s") \
+	/*260 */_(ER_BOOTSTRAP_NOT_UNANIMOUS,	"Replica %s chose a different bootstrap leader %s") \
+	/*261 */_(ER_CANT_CHECK_BOOTSTRAP_LEADER, "Can't check who replica %s chose its bootstrap leader") \
+	/*262 */_(ER_BOOTSTRAP_CONNECTION_NOT_TO_ALL, "Some replica set members were not specified in box.cfg.replication") \
+	/*263 */_(ER_NIL_UUID,			"Nil UUID is reserved and can't be used in replication") \
```
locker, line 318: Please don't move all these error codes. Instead replace `ER_DDL_NOT_ALLOWED` with `ER_UNUSED` like it was done in [this commit](https://github.com/tarantool/tarantool/commit/d026b546fa6b437c90c6c7849e3a0dfe9ae76f92#diff-3ab8a0e9527079f220221deb0f67e18ea14b35d814ab2300055745626e7b314f) for example. Later we'll reuse it for something else.

test/box-luatest/gh_7149_forbid_ddl_until_box_schema_upgrade_test.lua
```diff
@@ -21,62 +21,3 @@ g.test_schema_access = function()
         box.schema.upgrade()
     end)
 end
-
```
locker, line 24: This test now looks pointless. Let's drop it altogether?

locker (changes requested):

src/box/errcode.h
Gumix: Done.

test/box-luatest/gh_7149_forbid_ddl_until_box_schema_upgrade_test.lua
Gumix: Let's keep it, there is a check for the another minor issue.

locker (approved):

PR #8223, opened 2023-01-27
Changelog verbosity for `bootstrap_strategy` and compat option for `replication_sync_timeout`

### LINKED ISSUE #8209: Restore old default value of `replication_sync_timeout`, add a compat option for the new value
Commit 67cb4e4e72a85da50b85cfc0309970d68bb8d3df changed default `replication_sync_timeout` value to 0 (it was 300 seconds before). This is a breaking change, which's stated in the changelog: a user expecting his initial `box.cfg` call to return **after** the master becomes writable will now see that master is read-only ("orphan") for some period of time.

Let's fix this and put the new behavior under compat.

sergepetrenko:
As part of our efforts to reduce the number of breaking changes in the upcoming release, add a compatibility option for the default value of `replication_sync_timeout` instead of setting it unconditionally and reword the changelog for another breaking change -- `bootstrap_strategy = "auto"` introduction.

Closes #8209


src/box/lua/load_cfg.lua
```diff
@@ -187,6 +188,56 @@ local default_cfg = {
...
+local replication_sync_timeout_brief = [[
+Sets the default value for box.cfg.replication_sync_timeout.
+Old is 300 seconds, new is 0 seconds. New behaviour makes
+box.cfg{replication = ...} call exit without waiting for
+synchronisation with all the remote nodes. This means that the node
+might be in 'orphan' state for some time after the box.cfg{} call
+returns.
+
+Set before first box.cfg{} call in order for the option to take effect.
+]]
```
Totktonada, line 212: Please, add https://github.com/tarantool/tarantool/wiki/compat%3Areplication_sync_timeout link to end of the brief description (see `src/lua/compat.lua` for examples) and add it into the list of the options: https://github.com/tarantool/tarantool/wiki/compat

src/box/lua/load_cfg.lua
```diff
@@ -187,6 +188,56 @@ local default_cfg = {
...
+        newval = 0,
+        obsolete = nil
+    },
+}
+
+for _, option in ipairs(compat_options) do
+    local option_name = 'box_cfg_' .. option.name
+    compat.add_option({
+        name = option_name,
+        default = option.obsolete and 'new' or 'old',
```
Totktonada, line 229: Those are three stages:

* old by default
* new by default
* obsolete (always new)

changelogs/unreleased/gh-5272-bootstrap-strategy-auto.md
```diff
@@ -2,7 +2,18 @@
...
+  `replication_connect_quorum`, `bootstrap_strategy` is automatically set to
+  "legacy" for backward compatibility reasons.
+
+  Examples of changed behaviour (when neither `bootstrap_strategy` nor
+  `replication_connect_quorum` are passed explicitly) include:
+  1) bootstrapping a new replica in an existing cluster will only succeed if all
+  the cluster members are listed in replica's `box.cfg.replication`.
+  2) when reconfiguring `box.cfg.replication`, the node will not stay orphan
+  even if some of the remote peers are unavailable. The node will transition
+  from orphan to running as soon as all the alive peers are synced (gh-5272).
```
Totktonada, line 19: A few questions on the new changelog entry.

1. We agreed to split breaking and non-breaking changes. This entry is marked as breaking at whole. I would split it into two entries (we can add several entries within one file).
2. Are the examples in the last paragraph present exhaustive list of breaking changes? If so, it worth to rephrase it not as examples, but as the list of breaking changes.

src/box/lua/load_cfg.lua
sergepetrenko: Judging by `compat.lua`, one sets `option.default` to 'new' and `option.obsolete` to true at the same time, to make the option "obsolete". Am I missing something here?
https://github.com/tarantool/tarantool/blob/db6245f8b9d2720fb585cb16a6af8175d50c1bd2/src/lua/compat.lua#L158-L170

sergepetrenko: Never mind. I see what you mean.
The option may be "new" by default, but not yet obsolete.

sergepetrenko: Fixed:
```diff
diff --git a/src/box/lua/load_cfg.lua b/src/box/lua/load_cfg.lua
index 6bed23cad..0fca98f5b 100644
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -218,7 +218,8 @@ local compat_options = {
         brief = replication_sync_timeout_brief,
         oldval = 300,
         newval = 0,
-        obsolete = nil
+        obsolete = nil,
+        default = 'old',
     },
 }
 
@@ -226,7 +227,7 @@ for _, option in ipairs(compat_options) do
     local option_name = 'box_cfg_' .. option.name
     compat.add_option({
         name = option_name,
-        default = option.obsolete and 'new' or 'old',
+        default = option.default,
         obsolete = option.obsolete,
         brief = option.brief,
         action = function(is_new)
```

src/box/lua/load_cfg.lua
sergepetrenko: Done. Thanks!

changelogs/unreleased/gh-5272-bootstrap-strategy-auto.md
sergepetrenko: 1. Ok
2. Yes, the list is exhaustive.

Now the changelog looks like:
```markdown
## feature/replication

* Introduce new configuration option `bootstrap_strategy`. The default value of
  this option - "auto" - brings new replica behaviour on replica set bootstrap,
  recovery and on replication reconfiguration. The user may set the option to
  "legacy" to preserve old behaviour.

  The new option value - "auto" - will be in effect only when no value for
  `replication_connect_quorum` is passed. If the user passes any value of
  `replication_connect_quorum`, `bootstrap_strategy` is automatically set to
  "legacy" for backward compatibility reasons (gh-5272).

* **[Breaking change]** If options `bootstrap_strategy` and
  `replication_connect_quorum` are not passed explicitly, the following
  behaviour will change:
  1) bootstrapping a new replica in an existing cluster will only succeed if all
  the cluster members are listed in replica's `box.cfg.replication`.
  2) when reconfiguring `box.cfg.replication`, the node will not stay orphan
  even if some of the remote peers are unavailable. The node will transition
  from orphan to running as soon as all the alive peers are synced (gh-5272).
```

changelogs/unreleased/gh-5272-sync-timeout-default.md
```diff
@@ -0,0 +1,13 @@
+## feature/replication
+
+* Added a compat option "box_cfg_replication_sync_timeout" to control the
+  default value of "replication_sync_timeout" box.cfg option. Old default is 300
+  seconds and new default is 0. The noticeable difference in new behavior is
+  that `box.cfg{replication = ""}` call returns before the node is synced with
+  remote instances. If one expects the node to be writable once `box.cfg`
+  returns, this can be achieved with new behavior by calling `box.ctl.wait_rw()`
+  after `box.cfg{replication=...}`.
```
Totktonada, line 9: I don't know, whether it is important, but is there a workaround for `box.cfg({read_only = true, replication = <...>})`?

sergepetrenko: I suppose no. The main difference is whether the node is writable or not after `box.cfg{}` call returns.
If the config contains `read_only = true`, the node won't be writable anyway.

Totktonada: It seems, it is not a question for those changelog/compat option changes, but rather about the change itself, but I'll summarize it just in case.

Let's assume I have a read only node, which serves read requests (to reduce load on writeable nodes). Technically I should handle a possibility to move into an orphan state due to a network problems to don't feed stale data to clients (and should also handle large replication lags). However, if I have a good network, it is possible that I neven meet this situation and don't handle it carefully.

After the change I'll meet the situation after a restart with a good probability. So, technically, there should be a way to fail iproto requests, when the data is stale (orphan mode or large replication lag). Preferably as a box option, because it is simpler than a middleware for all iproto functions and triggers for all spaces. I would also add an option to the http server to fail http requests in this case.

However it is definitely out of scope here. I'm closing the discussion.

src/box/lua/load_cfg.lua
```diff
@@ -187,6 +188,57 @@ local default_cfg = {
...
+    compat.add_option({
+        name = option_name,
+        default = option.default,
+        obsolete = option.obsolete,
+        brief = option.brief,
+        action = function(is_new)
+            local val = is_new and option.newval or option.oldval
+            default_cfg[option.name] = val
+            pre_load_cfg[option.name] = val
+        end,
```
Totktonada, line 237: We can raise an error on attempt to set the option after/during first `box.cfg()`. It should work better than a caution in the help message :)

It seems, it should be something like this:

```lua
local function is_locked()
    return lock:is_empty()
end

if box_is_configured or is_locked() then
    error(<...>)
end
```

sergepetrenko: Good idea, done:

```diff
diff --git a/src/box/lua/load_cfg.lua b/src/box/lua/load_cfg.lua
index 0c5bca6f5..4330ca00d 100644
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -25,6 +25,10 @@ local function locked(f)
     end
 end
 
+local function is_locked()
+    return lock:is_full()
+end
+
 --
 -- When a feature is disabled, every single mentioning of it should
 -- be eliminated. Even box.cfg{} should not accept any parameters
@@ -200,6 +204,17 @@ local pre_load_cfg = table.copy(default_cfg)
 -- of nil values.
 local pre_load_cfg_is_set = {}
 
+
+-- Whether box is loaded.
+--
+-- `false` when box is not configured or when the initialization
+-- is in progress.
+--
+-- `true` when box is configured.
+--
+-- Use locked() wrapper to obtain reliable results.
+local box_is_configured = false
+
 local replication_sync_timeout_brief = [[
 Sets the default value for box.cfg.replication_sync_timeout.
 Old is 300 seconds, new is 0 seconds. New behaviour makes
@@ -231,6 +246,10 @@ for _, option in ipairs(compat_options) do
         obsolete = option.obsolete,
         brief = option.brief,
         action = function(is_new)
+            if is_locked() or box_is_configured then
+                error("The compat  option '" .. option_name .. "' takes " ..
+                      "effect only before the initial box.cfg() call")
+            end
             local val = is_new and option.newval or option.oldval
             default_cfg[option.name] = val
             pre_load_cfg[option.name] = val
@@ -952,16 +971,6 @@ setmetatable(box, {
      end
 })
 
--- Whether box is loaded.
---
--- `false` when box is not configured or when the initialization
--- is in progress.
---
--- `true` when box is configured.
---
--- Use locked() wrapper to obtain reliable results.
-local box_is_configured = false
-
 local function load_cfg(cfg)
     -- A user may save box.cfg (this function) before box loading
     -- and call it afterwards. We should reconfigure box in the
```
</details>

changelogs/unreleased/gh-5272-bootstrap-strategy-auto.md
sergepetrenko: **upd**: after our verbal discussion, changed the changelog:
```markdown
## feature/replication

* Introduce new configuration option `bootstrap_strategy`. The default value of
  this option - "auto" - brings new replica behaviour on replica set bootstrap,
  recovery and on replication reconfiguration. The user may set the option to
  "legacy" to fully return to old behaviour.

  The new option value - "auto" - will be in effect only when no value for
  `replication_connect_quorum` is passed. If the user passes any value of
  `replication_connect_quorum`, `bootstrap_strategy` is automatically set to
  "legacy" for backward compatibility reasons.

  Note, that if you leave the options untouched (that is, `bootstrap_strategy`
  defaults to "auto"), the following behaviour will noticeably change: during
  the recovery from local files and during replication reconfiguration the
  replica will not stay "orphan" even if some of the remote peers listed in
  `box.cfg.replication` are unavailable. The node will transition from "orphan"
  to "running" as soon as it syncs with all the reachable peers (unlike before,
  when not setting `replication_connect_quorum` would mean that each remote peer
  connection is necessary). Note, the node will try to establish connections
  with all the peers listed in `box.cfg.replication` for a period of
  `replication_connect_timeout` seconds (gh-5272).

* **[Breaking change]** If neither of the configuration options
  `bootstrap_strategy` and `replication_connect_quorum` is passed explicitly, or
  if `bootstrap_strategy` is set to "auto", bootstrapping a new replica in an
  existing replica set will only succeed if all the replica set members are
  listed in replica's `box.cfg.replication`. For example, when joining a fresh
  replica to a replica set of 3 nodes, all 3 node URIs must be present in
  replica's `box.cfg.replication` parameter. When joining 2 new replicas to a
  single master, both replicas must have each other's URIs (alongside with
  master's URI) in their `box.cfg.replication` (gh-5272).
```

sergepetrenko: And one more iteration:
```markdown
## feature/replication

* Introduce new configuration option `bootstrap_strategy`. The default value of
  this option - "auto" - brings new replica behaviour on replica set bootstrap,
  replica join to an existing replica set, recovery and replication
  reconfiguration. The user may set the option to "legacy" to fully return to
  old behaviour.

  The new option value - "auto" - will be in effect only when no value for
  `replication_connect_quorum` is passed. If the user passes any value of
  `replication_connect_quorum`, `bootstrap_strategy` is automatically set to
  "legacy" for backward compatibility reasons.

  Note, that if you leave the options untouched (that is, `bootstrap_strategy`
  defaults to "auto"), the following behaviour will noticeably change: during
  the recovery from local files and during replication reconfiguration
  `box.cfg{replication = ...}` will not fail even if some (or all) of the remote
  peers listed in `box.cfg.replication` are unavailable. Instead the node will
  try to establish a connection to everyone for a period of
  `replication_connect_timeout` and then transition to
  `box.info.status == "running"` as soon as it syncs with all the reached peers.
  (gh-5272).

* **[Breaking change]** If neither of the configuration options
  `bootstrap_strategy` and `replication_connect_quorum` is passed explicitly, or
  if `bootstrap_strategy` is set to "auto", bootstrapping a new replica in an
  existing replica set will only succeed if all the replica set members are
  listed in replica's `box.cfg.replication`. For example, when joining a fresh
  replica to a replica set of 3 nodes, all 3 node URIs must be present in
  replica's `box.cfg.replication` parameter. When joining 2 new replicas to a
  single master, both replicas must have each other's URIs (alongside with
  master's URI) in their `box.cfg.replication` (gh-5272).
```

Totktonada: Thank you for the updates! I think we reached a good balance between verbosity and simplicity.

Totktonada (approved):

Issue #6200, opened 2021-07-07
json.encode() escapes '/'


vasilky3:
Приветствую, столкнулся с неожиданным экранированием в json модуле.
```
localhost:3501> t = {url = "https://srv:7777"}
---
...
localhost:3501> require('json').encode(t)
---
- '{"url":"https:\/\/srv:7777"}'
...
```

`Json.encode` экранирует `/`, по спеке это не обязательно. [rfc4627]

```
 All Unicode characters may be placed within the
   quotation marks except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).
```

https://github.com/tarantool/tarantool/blob/4b2016ab3d35b505ea4712998458c2cd553b0271/third_party/lua-cjson/lua_cjson.c#L135

**Ожидаемое поведение:**
```
localhost:3501> t = {url = "https://srv:7777"}
---
...
localhost:3501> require('json').encode(t)
---
- '{"url":"https://srv:7777"}'
...
```

vasilky3:
Up. На стороне клиента также cjson. При decode() экранирование убирается.
Таким образом проблема решилась сама собой.

Totktonada:
Please, share a reason why do you closed the issue if it was intentional. The problem is there.

Totktonada:
A bit more context:

* The issue https://github.com/mpx/lua-cjson/issues/66 in the upstream repository, where the example with an URL (`https://` -> `https:\/\/`) is shown and the author of the library asked for a specific use case, where something breaks due the forward slash escaping.
* The pull request https://github.com/mpx/lua-cjson/pull/57, where the unconditional fix for the `char2escape` table is proposed and a discussion appears. Interesting points:
  * The author asked again for specific cases, which are broken by the forward slash escaping.
  * Aside of the URL example, a token example was provided.
  * However I don't see a self-contained / full scenario, where something becomes broken. Quite interesting. I also can't propose one.
* There is another pull request https://github.com/mpx/lua-cjson/pull/76, which implements a global option to choose whether to escape a forward slash, but the author asked to lower this option to `cjson.new()` (IOW, to make it encoder instance specific). I guess the reason is to don't introduce too many global options.
* OpenResty's cjson fork has a global option for that: https://github.com/openresty/lua-cjson/#encode_escape_forward_slash.

----

What is also interesting: does anyone verify the `char2escape` table to spot all cases, where it escapes symbols that are optional for escape? Are there other such cases?

----

[...]

PR #8224, opened 2023-01-27
Introduce the tweaks subsystem and switch the compat module to it

### LINKED ISSUE #7883: Internal config registry
[Desgin document](https://www.notion.so/tarantool/Tarantool-tweaks-0b42aca6d98149b6b32cd1e243a5c96a)

It'd be nice to have a registry available both in C and Lua that would store internal (dev-only) options.

Use cases:
- Conditionally enable an experimental behavior.
  Example: we could use it to try a new algorithm in test and staging environments.
- Switch to a new behavior, but provide an option to switch back in case something goes wrong.
  Example: after successfully trying a new algorithm, we could switch to it by default, but leave the option to switch back.
- The compatibility module could use the internal registry to provide a user-friendly way to change Tarantool behavior between versions, see #7000.

We could also use the internal configuration registry under the hood in `box.cfg`.

### LINKED ISSUE #8117: Simplify addition of new compat options from C code
[Desgin document](https://www.notion.so/tarantool/Tarantool-tweaks-0b42aca6d98149b6b32cd1e243a5c96a)

In the scope of #7000, a new module "compat" was introduced. The goal of this new module is to provide a compatibility configuration layer to make Tarantool updates less painful for users. In a nutshell, when a breaking change is about to be introduced, we should add a new compat option for it and set it to 'old', which means the old behavior (i.e. no breaking change actually happens). When the major version is bumped, we should:
 - update all 'old' compat options to 'new', but allow to reset them to 'old' manually;
 - update all 'new' compat options to 'deprecated' and disable/drop the corresponding code.

The problem is that the new module is Lua-oriented. Adding an option from C is complicated, because:
- One has to write some boilerplate C and Lua code to add a Lua wrapper for the new option (e.g. see [here](https://github.com/tarantool/tarantool/blob/182034dc979b3125f0ad66227997b58b400b6e2e/src/lua/compat.c#L58-L71) and [here](https://github.com/tarantool/tarantool/blob/182034dc979b3125f0ad66227997b58b400b6e2e/src/lua/compat.lua#L73-L78)).
- The code should be added to two different files.
- All options should be either declared or included into `src/lua/compat.c`.

It'd be awesome if the process was simplified somehow. Ideally, it should be enough to just call one 'register' function per compat option in any place of the C code while the rest should be done automatically by the compat module internals.

Maybe related with #7883.

locker:
A tweak is basically an internal C variable that can be switched from Lua, for example:

```Lua
local tweaks = require('internal.tweaks')
tweaks.json_escape_forward_slash = true
```

A tweak may be boolean, integer, or enum. Defining a tweak in C is as simple as using one macro, for example:

```C
static bool my_tweak;
TWEAK_BOOL(my_tweak);
```

Besides introducing the tweaks subsystem, this PR also switches the compat module completely to it and drops `lua/compat.c` wrappers.

This PR also bumps the msgpuck submodule as per [this](https://github.com/tarantool/tarantool/pull/8224#discussion_r1092911459) discussion.

Closes #7883
Closes #8117


src/lua/tweaks.c
```diff
@@ -0,0 +1,132 @@
...
+tarantool_lua_tweaks_init(struct lua_State *L)
+{
+	const struct luaL_Reg mt[] = {
+		{"__index", luaT_tweaks_index},
+		{"__newindex", luaT_tweaks_newindex},
+		{"__serialize", luaT_tweaks_serialize},
+		{"__autocomplete", luaT_tweaks_serialize},
+		{NULL, NULL},
+	};
+	luaL_register(L, "internal.tweaks", NULL);
```
Totktonada, line 127: What do you think about https://github.com/tarantool/doc/issues/3308?

Maybe replace `foo.lib` in my PR (see [here](https://github.com/tarantool/tarantool/pull/8173#discussion_r1088439359)) with `internal.foo` (and use some temporary name for the Lua/C part of compat)? If we'll agree on it, I'll replace `net.box.lib` with `internal.net.box` as well.

src/lua/tweaks.c
```diff
@@ -0,0 +1,132 @@
...
+	lua_newtable(L);
+	luaL_register(L, NULL, mt);
```
Totktonada, line 129: I suggest to use `luaL_setfuncs()`. I plan to eliminate `luaL_register()` usages to set functions into a table in PR #8173 and use `luaL_setfuncs()` for that.

src/lua/compat.lua
```diff
@@ -69,12 +68,7 @@ local options = {
         obsolete = nil,
         brief = JSON_ESCAPE_BRIEF,
         run_action_now = true,
-        action = function(is_new)
-            local esc_slash = not is_new
-            require('json').cfg{encode_escape_forward_slash = esc_slash}
-            internal.json_escape_forward_slash_toggle(esc_slash)
-            internal.msgpuck_escape_forward_slash_toggle(esc_slash)
```
Totktonada, line 64: How about code that uses `mp_snprint()` (and so `mp_char2escape` from `src/lib/msgpuck/hints.c`)?

src/lua/tweaks.c
locker: > Maybe replace `foo.lib` in my PR (see [here](https://github.com/tarantool/tarantool/pull/8173#discussion_r1088439359)) with `internal.foo` (and use some temporary name for the Lua/C part of compat)? If we'll agree on it, I'll replace `net.box.lib` with `internal.net.box` as well.

There's a difference between `net.box.lib` and `internal.tweaks`. The former is a helper module that is supposed to be used by the net.box implementation only while the latter is a final module that can be used by end users. We just don't want the tweaks module to be public. So no, I don't think we should rename `foo.lib` to `internal.foo` in your PR. I think of `require('tarantool')._internal.tweaks` as an alias for `require('internal.tweaks')`. I can't get rid of `internal.tweaks` completely (e.g. by fixing `package.loaded`), because it's used by the `compat` module which is loaded before the `tarantool` module. Maybe, we should drop the `require('tarantool')._internal.tweaks` alias altogether and use `internal.tweaks` instead? Come to think of it, maybe name the new module simply `tweaks`? It sounds "internal" enough.

As for the Lua/C part of compat, I drop it in the last path of the series, because it's not needed anymore.

src/lua/tweaks.c
locker: There are two calls to `luaL_register` in this commit. AFAIU I can replace the second one with `luaL_setfuncs`, but to eliminate the first one, I need the new `luaT_newmodule` function that is introduced in your PR. So I'll wait for your PR to be merged.

src/lua/compat.lua
locker: AFAICS `mp_snprint` is an internal function used for debugging and logging only. It dumps MsgPack data in a human-readable format, which isn't JSON. I don't see why we should enable escaping of the forward slash for it. AFAIU it was enabled solely because of `json_encode` (which is used by the JSON logger) using the `mp_char2escape` table internally. This doesn't make any sense, anyway - `json_encode` should use the same escape table as our JSON encoder, not the one from the MsgPack library. I tried to explain that in the last commit message.

src/lua/tweaks.c
Totktonada: I meant the second `luaL_register()` call, but in fact #8173 now ~has all approves and will go soon~merged. So it worth to rebase it.

src/lua/tweaks.c
Totktonada: Regarding `foo.lib` and `internal.foo`: okay, let's assume this is our new naming rule.

Regarding eliminating of the `internal.tweak` module: I don't think that it worth to do at all. We need some mechanism to pass values over built-in modules and `require()` machinery fits good here. It makes our built-in modules similar to usual ones and reduces amount of practices a developer should be aware of. I wouldn't change it. At least I don't see a reason for now.

Regarding entrypoint to the `tweaks` API: I don't mind any of the options, but I would prefer to have 'internal' word somewhere or start the name from the underscore symbol.

src/lua/compat.lua
Totktonada: `log.info()` is also used for debugging and logging only, however it is deliberately changed by the compat option. I'm worrying about the inconsistency. When the new behavior will be enabled in compat, some messages will have escaped forward slash after this patchset (while some will conform to the option).

src/lib/core/tweaks.h
```diff
@@ -0,0 +1,226 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
```
Totktonada, line 4: Just curious: `Copyright 2010-2023` here and `Copyright 2022-2023` in src/lua/tweaks.c. What is more correct or better in some sense?

src/lua/tweaks.c
locker: Okay, let's leave `internal.tweaks` for the module name and drop the `tarantool._internal.tweaks` alias, as it isn't used anywhere anyway and users can just as well call `require('internal.tweaks')` instead.

```diff
diff --git a/src/lua/init.lua b/src/lua/init.lua
index 226afb83c8e7..c8f5e1eef915 100644
--- a/src/lua/init.lua
+++ b/src/lua/init.lua
@@ -356,6 +356,5 @@ return {
         strip_cwd_from_path = strip_cwd_from_path,
         module_name_from_filename = module_name_from_filename,
         run_preload = run_preload,
-        tweaks = require('internal.tweaks'),
     };
 }
```

</details>

src/lua/tweaks.c
locker: Done.

```diff
diff --git a/src/lua/tweaks.c b/src/lua/tweaks.c
index d71e1094c004..b8ae1ba38048 100644
--- a/src/lua/tweaks.c
+++ b/src/lua/tweaks.c
@@ -117,16 +117,19 @@ luaT_tweaks_serialize(struct lua_State *L)
 void
 tarantool_lua_tweaks_init(struct lua_State *L)
 {
-	const struct luaL_Reg mt[] = {
+	const struct luaL_Reg module_funcs[] = {
+		{NULL, NULL},
+	};
+	const struct luaL_Reg module_mt_funcs[] = {
 		{"__index", luaT_tweaks_index},
 		{"__newindex", luaT_tweaks_newindex},
 		{"__serialize", luaT_tweaks_serialize},
 		{"__autocomplete", luaT_tweaks_serialize},
 		{NULL, NULL},
 	};
-	luaL_register(L, "internal.tweaks", NULL);
+	luaT_newmodule(L, "internal.tweaks", module_funcs);
 	lua_newtable(L);
-	luaL_register(L, NULL, mt);
+	luaL_setfuncs(L, module_mt_funcs, 0);
 	lua_setmetatable(L, -2);
 	lua_pop(L, 1);
 }
```

</details>

src/lib/core/tweaks.h
locker: My bad. I blindly copied the copyright header from `src/lua/compat.h`, which used `2022-2022` for some reason. It looks like we use `2010-2023` everywhere else. Fixed.

```diff
diff --git a/src/lua/tweaks.c b/src/lua/tweaks.c
index b8ae1ba38048..8a3c8045fcc8 100644
--- a/src/lua/tweaks.c
+++ b/src/lua/tweaks.c
@@ -1,7 +1,7 @@
 /*
  * SPDX-License-Identifier: BSD-2-Clause
  *
- * Copyright 2022-2023, Tarantool AUTHORS, please see AUTHORS file.
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
  */
 #include "lua/tweaks.h"
 
diff --git a/src/lua/tweaks.h b/src/lua/tweaks.h
index 5ad429dde0fb..1e5455727b29 100644
--- a/src/lua/tweaks.h
+++ b/src/lua/tweaks.h
@@ -1,7 +1,7 @@
 /*
  * SPDX-License-Identifier: BSD-2-Clause
  *
- * Copyright 2022-2023, Tarantool AUTHORS, please see AUTHORS file.
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
  */
 #pragma once
 
```

</details>

src/lua/compat.lua
locker: `log.info` doesn't use `mp_snprint`. It uses the `json.encode` Lua function if its argument is a table. It also uses the `json_encode` C function internally in `say.c` if the log format is JSON. Both functions now don't have anything to do with `mp_char2escape` and whether they escape forward slashes or not is configured with the new tweak.

BTW the effect that the compat option has on `log.info` is tested:

https://github.com/tarantool/tarantool/blob/d4e928091b18406cd966e54f9eb207d3682ae730/test/app-luatest/gh_6200_lua_cjson_escapes_slash_test.lua#L72-L127

locker: Discussed with @alyapunov and @Totktonada and agreed that it's okay to remove forward slash escaping from `mp_snprint` unconditionally. We don't need to do it under the compat option, because we only use it for presenting MsgPack data in *unspecified* human-readable format (e.g. for logging) - we never promised that it won't change (we didn't even promise that it's a valid JSON).

I opened a PR for libmsgpuck: https://github.com/tarantool/msgpuck/pull/25

alyapunov (approved):

src/trivia/util.h
```diff
@@ -573,6 +573,20 @@ int clock_gettime(uint32_t clock_id, struct timespec *tp);
...
+ * encoded in JSON or NULL otherwise.
+ */
+static inline const char *
+json_escape_char(char c)
+{
+	extern const char *const json_char2escape[];
+	extern bool json_escape_forward_slash;
+	if (c == '/' && json_escape_forward_slash)
+		return "\\/";
+	return json_char2escape[(unsigned char)c];
```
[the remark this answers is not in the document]
locker: Since we're going to drop this change after two major releases anyway, I think it's better to keep it simple, and one condition check looks simpler to me than an extra escape table. (Let's assume we need to unescape one more character. Then we'd have to add two more tables!)

test/app-luatest/gh_6200_lua_cjson_escapes_slash_test.lua
```diff
@@ -53,25 +53,21 @@ local function server_test_json_new_encode()
     -- Test that '/' is escaped with 'old' setting.
     t.assert_equals(json.encode('/'), [["\/"]])
 
-    -- Create new serializer and check that it has correct defaults
-    -- and doesn't change behavior.
+    -- Test that a serializer created with json.new() follows
```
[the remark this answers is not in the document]
locker: This feature (configuring forward slash escape behavior per json serializer) looks unintended to me. I reckon it was introduced solely for the compat's sake, because the `encode_escape_forward_slash` JSON serializer option wasn't documented or even mentioned anywhere, not even in the commit message (see commit 1436601cbf8a6796bc0c9493bafd7206fbf50ed8).

Since we assume that escaping the forward slash character is a bug and are going to drop this functionality completely after a couple major releases, I don't think we should preserve this serializer option.

src/lua/tweaks.c
```diff
@@ -0,0 +1,135 @@
...
+		{"__index", luaT_tweaks_index},
+		{"__newindex", luaT_tweaks_newindex},
+		{"__serialize", luaT_tweaks_serialize},
+		{"__autocomplete", luaT_tweaks_serialize},
+		{NULL, NULL},
+	};
+	luaT_newmodule(L, "internal.tweaks", module_funcs);
+	lua_newtable(L);
+	luaL_setfuncs(L, module_mt_funcs, 0);
+	lua_setmetatable(L, -2);
```
Totktonada, line 133: > The module is implemented as a Lua userdata object

It seems, it isn't.

src/lua/tweaks.c
```diff
@@ -0,0 +1,135 @@
...
+ * Looks up a tweak value by name (2nd arg) and pushes it to Lua stack.
+ * Returns nothing if there's no such tweak.
+ */
+static int
+luaT_tweaks_index(struct lua_State *L)
+{
+	const char *name = luaL_checkstring(L, 2);
+	struct tweak *t = tweak_find(name);
+	if (t == NULL)
+		return 0;
```
Totktonada, line 49: Nit: I would `lua_pushnil()` + `return 1` to mimic a table behavior more precisely.

src/lua/tweaks.c
```diff
@@ -0,0 +1,135 @@
...
+	case LUA_TBOOLEAN:
+		v.type = TWEAK_VALUE_BOOL;
+		v.bval = lua_toboolean(L, 3);
+		break;
+	case LUA_TNUMBER:
+		v.type = TWEAK_VALUE_INT;
+		v.ival = lua_tointeger(L, 3);
+		if (lua_tonumber(L, 3) != v.ival)
+			goto invalid_value;
+		break;
```
Totktonada, line 79: Nit: I guess I can pass a value above 2^32 and I'll not receive an error.

src/lib/core/util.c
```diff
@@ -346,17 +346,65 @@ utf8_check_printable(const char *start, size_t length)
...
 	for (int i = 0; i < data_len; i++) {
-		unsigned char c = (unsigned char ) data[i];
-		if (c < 128 && mp_char2escape[c] != NULL) {
-			/* Escape character */
-			SNPRINT(total, snprintf, buf, size, "%s",
-					  mp_char2escape[c]);
+		char c = data[i];
+		const char *escstr = json_escape_char(c);
+		if (escstr != NULL) {
+			SNPRINT(total, snprintf, buf, size, "%s", escstr);
```
Totktonada, line 407: https://github.com/tarantool/tarantool/wiki/compat%3Ajson_escape_forward_slash

> You can select new/old behavior in tarantool.compat. It will affect global serializer instance and all others created with `json.new()` afterwards.

Now it also affects serializers created before setting the compat option. IOW, all the serializers.

src/lib/core/fiber_channel.c
```diff
@@ -58,6 +59,13 @@ struct ipc_wait_pad {
...
 
+static const char *const fiber_channel_close_mode_strs[] = {
+	"forceful",
+	"graceful",
+};
+
+TWEAK_ENUM(fiber_channel_close_mode, fiber_channel_close_mode);
+
 void
 fiber_channel_set_close_mode(enum fiber_channel_close_mode mode)
```
Totktonada, line 70: > Note that we don't drop fiber_channel_set_close_mode function, because
> it's still used in unit tests.

Shouldn't tweaks be easily accessible from unit tests? Or I missed some point why the function should be used in the unit test and not a tweak setter?

changelogs/unreleased/gh-6200-lua_cjson-encode-escapes-slash.md
```diff
@@ -1,5 +1,6 @@
 ## bugfix/lua/json
 
-* Now the forward slash (`/`) is not escaped in `json.encode()` and msgpack.
-  A new `tarantool.compat` option `json_escape_forward_slash` is added for
-  switching to the new behavior (gh-6200).
+* A new `tarantool.compat` option `json_escape_forward_slash` was added. This
+  option configures whether the internal JSON encoder escapes the forward slash
+  character (old behavior) or not (new behavior). This option affects the
+  `json.encode()` Lua function and the JSON logger (gh-6200).
```
Totktonada, line 6: > Part of #8117

Nit: Shouldn't we close the issue here? I understand that it'll be closed when the pull request will land, but usually we close an issue at the last commit in a series. So one may understand when the problem is considered entirely solved by looking over `git log`.

src/lua/tweaks.c
Totktonada: This comment and all the comments below are minor and can be safely ignored/postponed. Feel free to proceed.

Totktonada (approved):

src/lua/tweaks.c
locker: Right. Updated.

<details><summary>New commit message</summary>

```
lua: add tweaks module

This commit adds the new 'internal.tweaks' Lua module for toggling
internal tweaks. The module is implemented as a Lua table with the
__index and __newindex metamethods overridden, which makes it possible
to get/set a tweak value with a simple assignment, for example:

  tarantool> tweaks = require('internal.tweaks')
  ---
  ...

  tarantool> tweaks.fiber_channel_close_mode = 'graceful'
  ---
  ...

  tarantool> tweaks.fiber_channel_close_mode
  ---
  - graceful
  ...

(Note, currently there are no tweaks available; tweaks are added in the
following commits.)

Getting a value of an unknown tweak returns nil. Setting a value of an
unknown tweak raises 'No such option' error. Setting an invalid value
for an existing tweak raises 'Invalid value' error. The module also
implements __serialize and __autocomplete metamethods that return all
tweaks and their values in a table, for example:

  tarantool> tweaks
  ---
  - fiber_channel_close_mode: forceful
    yaml_pretty_multiline: false
    json_escape_forward_slash_default: true
  ...

Part of #7883
Needed for #8117

NO_DOC=internal
NO_CHANGELOG=internal
```

</details>

src/lua/tweaks.c
locker: Makes sense. Fixed.

```diff
diff --git a/src/lua/tweaks.c b/src/lua/tweaks.c
index 8a3c8045fcc8..40342652fc15 100644
--- a/src/lua/tweaks.c
+++ b/src/lua/tweaks.c
@@ -38,15 +38,17 @@ luaT_push_tweak_value(struct lua_State *L, const struct tweak_value *v)
 
 /**
  * Looks up a tweak value by name (2nd arg) and pushes it to Lua stack.
- * Returns nothing if there's no such tweak.
+ * Returns nil if there's no such tweak.
  */
 static int
 luaT_tweaks_index(struct lua_State *L)
 {
 	const char *name = luaL_checkstring(L, 2);
 	struct tweak *t = tweak_find(name);
-	if (t == NULL)
-		return 0;
+	if (t == NULL) {
+		lua_pushnil(L);
+		return 1;
+	}
 	struct tweak_value v;
 	tweak_get(t, &v);
 	return luaT_push_tweak_value(L, &v);
```

</details>

src/lua/tweaks.c
locker: You will receive an error if you pass a value that doesn't fit into `int`. I added a test case for this.

```diff
diff --git a/test/unit/lua_tweaks.c b/test/unit/lua_tweaks.c
index bad4f993fb53..28d0caed656e 100644
--- a/test/unit/lua_tweaks.c
+++ b/test/unit/lua_tweaks.c
@@ -63,7 +63,7 @@ test_index(void)
 static void
 test_newindex(void)
 {
-	plan(6);
+	plan(8);
 	header();
 	is(luaT_dostring(L, "tweaks.no_such_var = 1"), -1, "unknown option");
 	ok(!diag_is_empty(diag_get()) &&
@@ -81,6 +81,12 @@ test_newindex(void)
 	   strcmp(diag_last_error(diag_get())->errmsg,
 		  "Value must be boolean, integer, or string") == 0,
 	   "check error");
+	is(luaT_dostring(L, "tweaks.bool_var = 8589934592"), -1,
+	   "invalid value - big int");
+	ok(!diag_is_empty(diag_get()) &&
+	   strcmp(diag_last_error(diag_get())->errmsg,
+		  "Value must be boolean, integer, or string") == 0,
+	   "check error");
 	footer();
 	check_plan();
 }
```

</details>

src/lib/core/util.c
locker: [Updated](https://github.com/tarantool/tarantool/wiki/compat:json_escape_forward_slash/_compare/25b0416f6cff6ed75105dfa14e95f5e54dcece2b...950a3faca9df71d5328ad581461973994bf31918) the page.

src/lib/core/fiber_channel.c
locker: Yes, tweaks are accessible from unit tests, but I don't want to refactor the test in the scope of this commit (note, I said "we **don't** drop", not "we **can't** drop"). If we decide to drop `fiber_channel_set_close_mode`, we have two options:
1. Use the tweaks API in the test (`tweak_find`, `tweak_set`). This would look cumbersome IMO, taking into account the fact that we just want to toggle the value of a C variable from C code, i.e. we could just as well toggle it directly, which brings us to the second option.
2. Move the `fiber_channel_close_mode` variable to the header and toggle it directly in the test. This would bloat the patch with an unrelated changed.

IMO the code looks fine as it is so I decided to leave it as is.

changelogs/unreleased/gh-6200-lua_cjson-encode-escapes-slash.md
locker: Makes sense. Replaced `Part of #XXXX` with `Closes #XXXX` statements where appropriate.

src/lua/tweaks.c
Totktonada: `tweaks.bool_var` is a typo?

src/lib/core/fiber_channel.c
Totktonada: Okay, thanks for the explanation.

src/lua/tweaks.c
locker: No, it's not a typo. It doesn't make any difference which variable I use here, because I check the error message, and this error message is raised by `lua/tweaks.c`. If the value passed this check, the error would be different ("Invalid value, expected boolean"). I use `bool_var` for all other checks in this function.

locker:
Fixed compilation of `STATIC_ASSERT_VAR_TYPE` macro by adding the `MAYBE_UNUSED` attribute to the generated function:

```diff
commit 6dcc83f7ed6f0549fa4980a2df14cbfba6befde7
Author: Vladimir Davydov <vdavydov@tarantool.org>
Date:   Wed Feb 8 11:56:15 2023 +0300

    fix

diff --git a/src/trivia/util.h b/src/trivia/util.h
index 87015444562a..1ebeb29bb9ca 100644
--- a/src/trivia/util.h
+++ b/src/trivia/util.h
@@ -711,7 +711,7 @@ strtolowerdup(const char *s);
 
 /** Asserts that a global variable has a given type. */
 #define STATIC_ASSERT_VAR_TYPE(var, type)				\
-static inline void							\
+MAYBE_UNUSED static inline void						\
 var##_assert_type(void)							\
 {									\
 	const type *p = &(var);						\
```

</details>

Issue #7594, opened 2022-08-22
Add `__serializer` option to skip private (starting with `_`) fields of Lua object


locker:
It's conventional in Lua to prefix private fields with an underscore (`_`). However, the default Lua serializer outputs these fields to the admin console just like any other public field. As a result, one often has to define a `__serialize` method that filters out private fields. For example, here's what we do for net.box connection object:

https://github.com/tarantool/tarantool/blob/02fae15a3adb8ea450ebbe3c250a4846cf1cca69/src/box/lua/net_box.lua#L151-L165

This is tedious and error-prone, because one may easily forget to patch the `__serialize` method when a new field is added. Also, this approach results in duplication of the output in case an object is referenced twice in a table (e.g. by name and by id):

```
tarantool> t = {foo = o, [1] = o}
---
...

tarantool> t
---
- foo: &0
    a: 1
    _x: 2
  1: *0
...

tarantool> o = setmetatable({a = 1, _x = 2}, {__serialize = function(self) return {a = self.a} end})
---
...

tarantool> t = {foo = o, [1] = o}
---
...

tarantool> t
---
- foo:
    a: 1
  1:
    a: 1
...
```

We already have 'map' and 'array' options for prettifying table output (used like this: `__serialize = 'map'`):

https://github.com/tarantool/tarantool/blob/02fae15a3adb8ea450ebbe3c250a4846cf1cca69/src/lua/serializer.c#L334-L350

The suggestion is to add a special serializer option to automatically filter out private fields ('class', 'object'?).

This would be consistent with the console autocompletion handler that doesn't suggest fields that start with an underscore unless the typed string starts with an underscore as well.

Totktonada:
Until it is fixed I'll go with module local hacks case-by-case: PR #8227.

locker:
Come to think of it, `__serialize = 'object'` doesn't make much sense. It's probably better to add a helper Lua function to some utility module (maybe `table`?).

PR #8227, opened 2023-01-27
lua: hide tarantool._internal at serialization

Totktonada:
So the field will not be shown to a user in the interactive console.

Reported-by: Mons Anderson (@Mons).


src/lua/init.lua
```diff
@@ -349,13 +349,31 @@ local function run_preload()
...
 
-return {
-    uptime = uptime;
-    pid = pid;
-    compat = compat;
+-- Extract all fields from a table except ones that start from
+-- the underscore.
+--
+-- Useful for __serialize.
+local function hide_internal_fields(t)
```
igormunkin, line 356: Minor: I only wonder, why this function is not simply called `serialize`, but I'm not going to argue regarding the naming here. Anyway, my two cents here:
1. The function doesn't work in-place, but IMHO "hide" implies this.
2. You **get** the new table containing only the **public fields**.

It looks like `get_public_fields` fits more here. Feel free to ignore.

test/app-luatest/tarantool_module_no_internal_fields_test.lua
```diff
@@ -0,0 +1,33 @@
...
+    -- Collect fields that start from the underscore.
+    local internal_keys = {}
+    for k, _ in pairs(response) do
+        if k:startswith('_') then
+            table.insert(internal_keys, k)
+        end
+    end
+
+    -- Ensure that there are no such fields.
+    t.assert_equals(internal_keys, {})
```
igormunkin, line 32: Minor: I'd rather check that `next(internal_keys)` yields `nil`, since I'm not sure whether `assert_equals` compares tables by values (I've already checked the comment [here](https://github.com/tarantool/luatest/blob/master/luatest/assertions.lua#L173)). Feel free to ignore.

igormunkin (approved):
@Totktonada, thanks for the patch! LGTM, but still left two nits below.

test/app-luatest/tarantool_module_no_internal_fields_test.lua
Totktonada: I would do this in a project's code, but I like the luatest's deep comparison assertion due to its visual simplicity and verbose error reporting (when the check fails). Given way ideally fits for testing.

src/lua/init.lua
Totktonada: > I only wonder, why this function is not simply called `serialize`, but I'm not going to argue regarding the naming here.

I've named the function according to what is does, not how it is used. It seems meaningful for me.

> 1. The function doesn't work in-place, but IMHO "hide" implies this.

Maybe. I'll think here a bit more.

> 2. You **get** the new table containing only the **public fields**.
>
> It looks like `get_public_fields` fits more here.

I don't like `get_` functions -- it looks like one failed to find an appropriate name to describe what it does. I wouldn't say that all the remaining fields are definitely public. I know that `_<...>` ones are internal and don't know anything about others.

Totktonada: Renamed to `filter_out_private_fields()` (it also follows a terminology from #7594 and such fields are actually called 'private' in other languages).

src/lua/init.lua
```diff
@@ -349,13 +349,31 @@ local function run_preload()
     end
 end
```
Lord-KA, line 350: Nit: commit msg is a bit confusing, as one has to read title as a part of the continuing sentence (but the second line starts with a capital for whatever reason).  Feel free to ignore, but it would be much easier to read if reworded.

Lord-KA (approved):
LGTM, besides a nit below.

src/lua/init.lua
Totktonada: Removed 'so'.

Totktonada:
The replication-luatest/linearizable_test.lua fails in the freebsd-13 job, but it is irrelevant to the patchset and is known as flaky: #7737. I'll ignore this fail.

Issue #4577, opened 2019-10-18
Stop using lua_ and luaL_ namespaces for tarantool functions


PersDep:
Rename lua_ and luaL_ interfaces in tarantool code base (utils.h, etc..) so that we do not interfere with LuaJIT code base.

Totktonada:
We use luaL_ prefix for functions that don't operate on tarantool specific structure (at least in its API) and luaT_ for ones that do. There are exceptions from this rule across our codebase, but when we thought about that with Vladimir Davydov (@locker), we end with this as the rule.

Also keep in the mind that we are unable to change names for exported functions (at least when they are exposed via module.h): it is part of our public API.

Let's discuss certain name conflicts you found.

igormunkin:
@Totktonada, as discussed offline, I propose the following naming policy:
* stop spoiling `lua_` and `luaL_` namespaces with new interfaces and leave it to API defined in Lua Reference Manual
* continue using `luaT_` prefix for functions relied on tarantool specifiсs
* introduce a new header, similar to [this one](https://github.com/iponweb/luavela/blob/master/src/lextlib.h) to be provided by our LuaJIT fork and move all Lua extensions defined [here](https://github.com/tarantool/tarantool/blob/master/src/lua/utils.h) to a new separate namespace e.g. `luaE_`. Then deprecate usage of the corresponding signatures from `module.h`, and proxy them to the introduced header preventing compatibility breakage

igormunkin:
There is another prefix suggested by @Totktonada: `luaM_`. The explanations are the following (feel free to choose one or both):
* Original one by @Totktonada: **M** is simply the next after **L** (LOL, looks like `dd` naming case)
* My one: **M** stands for **M**iscellaneous

IMHO `luaM_` is also a good one.

Totktonada:
I vote to don't remove anything just for the same of sanity or formal accuracy. We'll hurt users and module developers without any, even artificial, reward. Keep Python 2 -> Python 3 story in the mind (all those replace `import foo` with `import bar`). Always.

We can recommend to avoid usage of old aliases and keep them in a separate section in the documentation.

It seems, we agreed on the policy for new names:

* `luaT_` for tarantool specific functions: ones that are work with tarantool's own structures or logically depend on tarantool in some other way.
* `luaM_` for general purpose Lua functions including LuaJIT specific ones (say, when it works with `cdata`).

We should proceed with the following steps:

* Update our code style guide with this agreement.
[...]

PR #8229, opened 2023-01-27
lua/utils: use `index2adr()` instead of index arithmetic

### LINKED ISSUE #8249: `luaL_checkcdata` et al. do not work with upvalues
**Bug description**

This code performs calculations to obtain the address from the given index, but it doesn't work for pseudo-indexes (e.g. upvalue indexes):
https://github.com/tarantool/tarantool/blob/352fe0c7d774ebaf91b165c99b8c67b9528a2cca/src/lua/utils.c#L171-L176

https://github.com/tarantool/tarantool/blob/352fe0c7d774ebaf91b165c99b8c67b9528a2cca/src/lua/utils.c#L182-L183

**Steps to reproduce**

```c
static int
closure(lua_State *L)
{
	uint32_t ctypeid;
	luaL_checkcdata(L, lua_upvalueindex(1), &ctypeid);
	return 0;
}

static void
test_upvalueindex(lua_State *L)
{
	uint32_t pvoid_ctypeid = luaL_ctypeid(L, "void *");
	*(void **) luaL_pushcdata(L, pvoid_ctypeid) = NULL;
	lua_pushcclosure(L, &closure, 1);
	lua_call(L, 0, 0);
}
```

**Actual behavior**

Lua error `expected cdata as -10002 argument` raised.

**Expected behavior**

No errors.

Needed for #7356

Gumix:
We have a code that performs calculations to obtain the address from the
given index, the address is then passed to `cdataV()` to get cdata value from
the stack. But this doesn't work for pseudo-indexes (e.g. upvalue indexes).
This patch brings `index2adr()` from `luajit/src/lj_api.c`, which accepts all
kinds of indexes, so that the calculations are no longer needed. Also the
helper function `luaL_tocpointer()` is introduced.

Closes #8249

### DIFF
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -469,11 +469,11 @@ luaL_tofield(struct lua_State *L, struct luaL_serializer *cfg, int index,
 		return 0;
 	case LUA_TCDATA:
 	{
-		GCcdata *cd = cdataV(L->base + index - 1);
-		void *cdata = (void *)cdataptr(cd);
+		uint32_t ctypeid;
+		void *cdata = luaL_tocpointer(L, index, &ctypeid);
 
 		int64_t ival;
-		switch (cd->ctypeid) {
+		switch (ctypeid) {
 		case CTID_BOOL:
 			field->type = MP_BOOL;
 			field->bval = *(bool*) cdata;
@@ -534,19 +534,19 @@ luaL_tofield(struct lua_State *L, struct luaL_serializer *cfg, int index,
 			/* Fall through */
 		default:
 			field->type = MP_EXT;
-			if (cd->ctypeid == CTID_DECIMAL) {
+			if (ctypeid == CTID_DECIMAL) {
 				field->ext_type = MP_DECIMAL;
 				field->decval = (decimal_t *) cdata;
-			} else if (cd->ctypeid == CTID_UUID) {
+			} else if (ctypeid == CTID_UUID) {
 				field->ext_type = MP_UUID;
 				field->uuidval = (struct tt_uuid *) cdata;
-			} else if (cd->ctypeid == CTID_CONST_STRUCT_ERROR_REF) {
+			} else if (ctypeid == CTID_CONST_STRUCT_ERROR_REF) {
 				field->ext_type = MP_ERROR;
 				field->errorval = *(struct error **)cdata;
-			} else if (cd->ctypeid == CTID_DATETIME) {
+			} else if (ctypeid == CTID_DATETIME) {
 				field->ext_type = MP_DATETIME;
 				field->dateval = (struct datetime *)cdata;
-			} else if (cd->ctypeid == CTID_INTERVAL) {
+			} else if (ctypeid == CTID_INTERVAL) {
 				field->ext_type = MP_INTERVAL;
 				field->interval = (struct interval *)cdata;
 			} else {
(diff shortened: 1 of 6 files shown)


src/lua/utils.c
```diff
@@ -53,6 +53,37 @@ uint32_t CTID_UUID;
 uint32_t CTID_DATETIME = 0;
 uint32_t CTID_INTERVAL = 0;
 
+/** A copy of index2adr() from luajit/src/lj_api.c */
+static TValue *
+luaL_index2adr(lua_State *L, int idx)
```
igormunkin, line 58: Generally, we try to avoid using `luaL` prefix in function naming (see #4577 for more info), but this still respects our current guidelines and this function is not exported, so can be freely renamed later. Feel free to ignore.

src/lua/utils.c
```diff
@@ -554,17 +576,13 @@ luaT_tolstring(lua_State *L, int idx, size_t *len)
 	return lua_tolstring(L, -1, len);
 }
 
-/* Based on ffi_meta___call() from luajit/src/lib_ffi.c. */
```
igormunkin, line 557: Typo: According to our [guidelines](https://www.tarantool.io/en/doc/latest/dev_guide/c_style_guide/#chapter-8-commenting) dots are required in comments. Same for the comment above.

igormunkin (changes requested):
@Gumix, thanks for the patch! Besides the comments below, I also grepped all sources to find all `L->base` occurrences. There is one in `src/lua/utils.h` and two in `src/lua/serializer.c`.

Considering these usages, I'd rather suggest to introduce the helper to obtain cdata payload.

src/lua/utils.c
Gumix: Renamed to `index2adr()`.

src/lua/utils.c
Gumix: Fixed.

Gumix:
> @Gumix, thanks for the patch! Besides the comments below, I also grepped all sources to find all `L->base` occurrences. There is one in `src/lua/utils.h` and two in `src/lua/serializer.c`.
> 
> Considering these usages, I'd rather suggest to introduce the helper to obtain cdata payload.

Added `get_cdata()`.

> P.S. I also think, that changelog is required, since you change the semantics of the public interfaces.

Added changelog and testcase.

src/lua/utils.c
```diff
@@ -54,6 +54,43 @@ uint32_t CTID_UUID;
...
+		} else {
+			idx = LUA_GLOBALSINDEX - idx;
+			return idx <= fn->c.nupvalues ?
+			       &fn->c.upvalue[idx - 1] : niltv(L);
+		}
+	}
+}
+
+GCcdata *
+get_cdata(struct lua_State *L, int idx)
```
igormunkin, line 89: OK, this is not quite the thing I meant in the previous review. I propose to implement a new function `luaL_tocpointer(lua_State *L, int idx, uint32_t *ctypeid)` with the signature similar to `lua_tolstring` and implementing `luaL_checkcdata` without actual check. I left an example below.

```c
void *
luaL_tocpointer(lua_State *L, int idx, uint32 *ctypeid) {
        TValue *o = index2adr(L, idx);
        if (!tviscdata(o)) {
                if (ctypeid != NULL) *ctypeid = 0;
                return NULL;
        }
        GCcdata *cd = cdataV(o);
        if (ctypeid != NULL) *ctypeid = cd->ctypeid;
        return cdataptr(cd);
}
```

IMHO, such function can be used widely in Tarantool to work with cdata without additional checks.

igormunkin (changes requested):
@Gumix, thanks for the changes! I've left some more comments below.

The main purpose of introducing `luaL_tocpointer` is avoid raw access to `L->base`. For other enhancements (if there is something you want to change) I suggest to open a new PR.

BTW, I'd rather split introducing of `index2adr` helper and `luaL_tocpointer` function into 2 separate commits. This is just recommendation, so feel free to ignore.

src/lua/serializer.c
```diff
@@ -447,9 +465,6 @@ int
 luaL_tofield(struct lua_State *L, struct luaL_serializer *cfg, int index,
 	     struct luaL_field *field)
 {
-	if (index < 0)
-		index = lua_gettop(L) + index + 1;
-
```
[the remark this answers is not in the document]
Gumix: Reverted.

src/lua/utils.c
Gumix: Done.

igormunkin (approved):
@Gumix, thanks for the fixes! LGTM.

PR #8231, opened 2023-01-30
feedback: send default metrics

### LINKED ISSUE #8192: Send metrics in feedback daemon
[PM link](https://www.notion.so/tarantool/metrics-min-avg-max-percentile-50-95-99-56433be43d5f4e55a10c81d70fd0ad4e)

Introduce new cfg options (defaults):
`feedback_send_metrics`(true) - collect and send metrics only if the flag is set,
`feedback_metrics_collect_interval`(60) - period of metrics collection, in seconds,
`feedback_metrics_limit`(1048576) - memory limit for metrics, in bytes.

If feedback is enabled and feedback_send_metrics is true, a background fiber must be started that collects metrics every feedback_metrics_collect_interval seconds. It must call `require('metrics').collect()` but it must do nothing if metrics module is not found. All the metrics must be appended to a huge array, but the approximate size exceeds feedback_metrics_limit bytes, some earlier collected metrics must be dropped.

If feedback_send_metrics is true, all the collected metrics must be send with other feedback to feedback server. At this moment all collected metrics must be dropped.

I think that the most complex part of this issue is to test it: we should somehow create a feedback server and check that all the metrics are successfully delivered to it.

drewdzzz:
The patch extends feedback_daemon functionality - now it collects default metrics if module metrics of required version (>= 0.16.0) is installed. The metrics are sent as a part of feedback. Feedback version is bumped.

Closes #8192

Manually checked that it works with actual `metrics` module.


src/box/lua/feedback_daemon.lua
```diff
@@ -348,6 +365,78 @@ local function feedback_loop(self)
...
+            size = size + obj_approx_size(k)
+            size = size + obj_approx_size(v)
+        end
+        return size
+    elseif type(obj) == 'string' then
+        return 17 + #obj
+    elseif type(obj) == 'boolean' then
+        return 1
+    else -- Numbers and cdata
+        return 8
```
CuriousGeorgiy, line 406: Where did you get the numbers for approximate type sizes from?

src/box/lua/feedback_daemon.lua
```diff
@@ -357,7 +446,17 @@ local function guard_loop(self)
...
-        local st = pcall(fiber.sleep, self.interval)
+        if self.send_metrics and
+           get_fiber_id(self.metrics_collect_fiber) == 0 then
+            self.metrics_collect_fiber =
+                fiber.create(metrics_collect_loop, self)
+            log.verbose("%s restarted", METRICS_PREFIX)
+        end
+        local interval = self.interval
+        if self.send_metrics then
+            interval = math.min(interval, self.metrics_collect_interval)
```
CuriousGeorgiy, line 457: Not sure this is correct. Since the intervals are different, technically, I guess there should be two guard loops.

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
```diff
@@ -0,0 +1,141 @@
...
+        t.assert(fb.metrics == nil or #fb.metrics < 5)
+    end)
+end
+
+g.test_no_metrics = function()
+    g.server:exec(function()
+        local t = require('luatest')
+        local fiber = require('fiber')
+        local has_metrics, _ = pcall(require, 'metrics')
+        t.skip_if(has_metrics)
```
CuriousGeorgiy, line 61: How can there be `metrics`, if you explicitly unload it in `g.before_each`?

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
```diff
@@ -0,0 +1,141 @@
...
+        fiber.sleep(0.3)
+        local daemon = box.internal.feedback_daemon
+        local fb = daemon.generate_feedback()
+        -- This check may fail because of assertion failure in collect method.
+        t.assert_type(fb.metrics, 'table')
+        t.assert_almost_equals(#fb.metrics, 30, 5)
+        fb = daemon.generate_feedback()
+        -- Check is metrics were reset. Fiber yields on feedback generation, so
+        -- metrics may be not empty.
+        t.assert(fb.metrics == nil or #fb.metrics < 5)
```
CuriousGeorgiy, line 52: Why don't you set a big enough `feedback_metrics_collect_interval` to test both cases?

src/box/lua/feedback_daemon.lua
```diff
@@ -7,15 +7,24 @@ local http  = require('http.client')
 local fio = require('fio')
 
 local PREFIX = "feedback_daemon"
+local METRICS_PREFIX = "metrics_collector"
```
CuriousGeorgiy, line 10: I would add a separate patch adding some prefix, e.g., `feedback` to `PREFIX`, `fiber`, etc., because now it becomes unclear what they refer to.

CuriousGeorgiy (changes requested):
Is there any integration test with `tarantool/feedback` or `tarantool/metrics` planned to test this patch?

src/box/lua/load_cfg.lua
```diff
@@ -181,6 +181,9 @@ local default_cfg = {
     feedback_crashinfo    = ifdef_feedback(true),
     feedback_host         = ifdef_feedback("https://feedback.tarantool.io"),
     feedback_interval     = ifdef_feedback(3600),
+    feedback_send_metrics = ifdef_feedback(true),
+    feedback_metrics_collect_interval = ifdef_feedback(60),
```
CuriousGeorgiy, line 185: But the PRD [says](https://www.notion.so/tarantool/metrics-min-avg-max-percentile-50-95-99-56433be43d5f4e55a10c81d70fd0ad4e#fdd6a1690d1a4710afe50dff6de9d87e) it should be 3600?

src/box/lua/feedback_daemon.lua
```diff
@@ -348,6 +365,78 @@ local function feedback_loop(self)
...
+local function insert_metric(self, new_metric)
+    local new_metric_size = obj_approx_size(new_metric)
+    self.metrics_size = self.metrics_size + new_metric_size
+    table.insert(self.metrics, new_metric)
+    table.insert(self.metrics_sizes, new_metric_size)
+    while #self.metrics > 0 and self.metrics_size > self.metrics_limit do
+        self.metrics_size = self.metrics_size - self.metrics_sizes[1]
+        table.remove(self.metrics_sizes, 1)
+        table.remove(self.metrics, 1)
+    end
```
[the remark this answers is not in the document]
CuriousGeorgiy: Your implementation also contradicts PRD.

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
drewdzzz: From the most popular website for Lua programmers - [Gamepedia](https://wowpedia.fandom.com/wiki/Lua_object_memory_sizes)!
But seriously, it is the only place where I can get approximate sizes of objects. Another place is Lua source code, but it is much harder to extract such information from there.

drewdzzz:
@CuriousGeorgiy Integration test is not planned

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
drewdzzz: > What if metrics_size is always larger than the size of one metric? Shouldn't we keep at least one metric?

As for me - no. If we never fit the limit, we should not store metrics.

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
drewdzzz: `require` does more than reading from `package.loaded` - search for the module file in current and user dirs, for example. So, unloading the module does not help here.

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
drewdzzz: Feedback daemon is cleared on reconfiguring, so I will need to wait for a long time before the first collection.1

src/box/lua/load_cfg.lua
drewdzzz: > `box.cfg.feedback_metrics_collect_interval = 3600`
Позволяет задавать интервал сбора метрик. 
По умолчанию, раз в минуту.

3600 is not a minute.

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
CuriousGeorgiy: @artur-barsegyan RFC

src/box/lua/feedback_daemon.lua
drewdzzz: Could you elaborate, please? Why do I need a separate patch? It is just name of the fiber for logs.

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
drewdzzz: Agree

src/box/lua/feedback_daemon.lua
drewdzzz: I'm not even sure that we need this guard loop.

src/box/lua/feedback_daemon.lua
CuriousGeorgiy: After this patch we will the daemon will have fields like `fiber` and `metrics_collect_fiber` — I suggest you add the prefix `feedback` to the latter ones for clarity.

src/box/lua/feedback_daemon.lua
drewdzzz: I don't see its value in reconfiguration. I assume that it is needed to restart the fiber if it fails with an error. However, I don't see any calls that can cause an error.

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
drewdzzz: Discussed with @igormunkin - generally it is OK for approximate size count. But if gc64 is enabled, trivial types consumes no memory - they are stored right in table slot. Added this logic.

<details>
<summary>diff</summary>

```diff

diff --git a/src/box/lua/feedback_daemon.lua b/src/box/lua/feedback_daemon.lua
index 4c8e292c9..58ac8c64b 100644
--- a/src/box/lua/feedback_daemon.lua
+++ b/src/box/lua/feedback_daemon.lua
@@ -5,6 +5,7 @@ local json  = require('json')
 local fiber = require('fiber')
 local http  = require('http.client')
 local fio = require('fio')
+local ffi = require('ffi')
 
 local PREFIX = "feedback_daemon"
 local METRICS_PREFIX = "metrics_collector"
@@ -386,11 +387,12 @@ local function collect_default_metrics()
     end
 end
 
+-- Trivial objects are stored right in table slot or stack with gc64
+local trivial_obj_size = ffi.abi('gc64') and 0 or 8
+
 -- Calculate approximate amount of occupied memory
 local function obj_approx_size(obj)
-    if type(obj) == nil then
-        return 0
-    elseif type(obj) == 'table' then
+    if type(obj) == 'table' then
         local size = 40
         for k, v in pairs(obj) do
             size = size + 16
@@ -400,10 +402,8 @@ local function obj_approx_size(obj)
         return size
     elseif type(obj) == 'string' then
         return 17 + #obj
-    elseif type(obj) == 'boolean' then
-        return 1
-    else -- Numbers and cdata
-        return 8
+    else -- Number, boolean and nil
+        return trivial_obj_size
     end
 end
 
diff --git a/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua b/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
index e0fa83612..3dc08e8df 100644
--- a/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
+++ b/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
@@ -120,17 +120,16 @@ g.test_memory_limit = function()
     g.server:exec(function()
         local t = require('luatest')
         local fiber = require('fiber')
-        box.cfg{feedback_metrics_limit = 8 * 5}
+        box.cfg{feedback_metrics_limit = 20 * 5}
         local m = {}
         m.collect = function(_)
-            return 0
+            return "abc"
         end
         m._VERSION = "0.16.0"
         package.loaded["metrics"] = m

```
</details>

src/box/lua/feedback_daemon.lua
drewdzzz: I've added `send` prefix to some fields in a separate commit.

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
CuriousGeorgiy: Thanks for the research!

src/box/lua/feedback_daemon.lua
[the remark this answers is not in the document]
drewdzzz: Discussed with @artur-barsegyan - we should always fit the limit.

drewdzzz:
@CuriousGeorgiy 
I decided to change preemption algo - now I keep old ones. Turned out that it does not matter which ones to keep, but keeping the old ones is simplier, does not require additional array of sizes and we does not need to move elems in huge array of metrics.

src/box/lua/feedback_daemon.lua
```diff
@@ -12,13 +12,13 @@ local METRICS_PREFIX = "metrics_collector"
```
CuriousGeorgiy, line 12: Nit, feel free to resolve and ignore: I would put this commit first.

src/box/lua/feedback_daemon.lua
```diff
@@ -348,6 +363,73 @@ local function feedback_loop(self)
...
+            size = size + obj_approx_size(k)
+            size = size + obj_approx_size(v)
+        end
+        return size
+    elseif type(obj) == 'string' then
+        return 17 + #obj
+    else -- Number, boolean and nil
+        return trivial_obj_size
+    end
+end
```
CuriousGeorgiy, line 405: Nit, feel free to resolve and ignore: I would open a ticket for refactoring this part into `src/lua/utils.h`, as this may be useful in other cases.

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
CuriousGeorgiy: I don't get your point, could you please elaborate? AFAIU, metrics are collected every `feedback_metrics_collect_interval`, and you explicitly retrieve them by calling `daemon.generate_feedback` — so why will you have to wait long?

CuriousGeorgiy (changes requested):

src/box/lua/feedback_daemon.lua
drewdzzz: I've taken a close look at this function. And I still don't see any value in guard fiber.

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
drewdzzz: Because I wait until metrics are collected, and only then I check that they are empty (or almost empty) - the point of this test case is to check if metrics are reset after retrieving.

src/box/lua/feedback_daemon.lua
CuriousGeorgiy: Sorry, my bad, I missed the fact `reload` also cancels the guard loop fiber.

But anyways, AFAIC, `json.encode` can throw an error inside the `feedback_loop` (e.g., on OOM) or `send_fiber` can get accidentally canceled by the user. The latter can also happen with the `metrics_collect_loop`.

CuriousGeorgiy: Leaving it up to @alyapunov.

CuriousGeorgiy (approved):

src/box/lua/feedback_daemon.lua
alyapunov: I think one guard is enough (and perhaps even one guard is more than needed).

alyapunov (approved):

drewdzzz:
@CuriousGeorgiy @alyapunov
Integration test failed because cartridge has its own [feedback wrapper](https://github.com/tarantool/cartridge/blob/master/cartridge/feedback.lua). So let's not rename fields for compatibility and file an issue in cartridge after this PR is merged - their feedback will be expired.

PR #8244, opened 2023-02-02
Fixes syslog facility

### LINKED ISSUE #8269: log: wrong priority value in syslog on Alpine and OpenBSD
Tarantool versions: all present versions are affected (since 1.7.4-460-ga362eff6a, which implements #2466).

A syslog message contains a number that represents facility and severity. The number is called priority (and the severity value also sometimes called priority). The [RFC 5424](https://www.rfc-editor.org/rfc/rfc5424) defines it this way (see section 6):

> ```
> PRI             = "<" PRIVAL ">"
> PRIVAL          = 1*3DIGIT ; range 0 .. 191
> ```
>
> <...>
>
> The Priority value is calculated by first multiplying the Facility number by 8 and then adding the numerical value of the Severity.

Code that implement this logic is the following:

https://github.com/tarantool/tarantool/blob/b6159217153c4312c93eedfd53927f8565232045/src/lib/core/say.c#L1003-L1009

It uses libc's `LOG_MAKEPRI` macro, which is non-standard and historically defined differently:

```c
#define LOG_MAKEPRI(facility, severity) (((facility) << 3) | (severity))
```

```c
#define LOG_MAKEPRI(facility, severity) ((facility) | (severity))
```

Our code assumes that it is implemented as `facility|severity`, but it is not always so. In particular:

* [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/syslog.html) has no `LOG_MAKEPRI` macro, but it defines an argument of the `syslog()` function this way:

  > Values of the priority argument are formed by OR'ing together a severity-level value and an optional facility value.

  It means that `LOG_LOCAL7` and other facility values are already bit-shifted.

  So I assume that `LOG_MAKEPRI` is defined correctly when it has no the bit-shift operation.
* FreeBSD: the macro is [fixed](https://svnweb.freebsd.org/base/head/sys/sys/syslog.h?r1=23252&r2=23251&pathrev=23252) in 1997 (so it is likely [released](https://en.wikipedia.org/wiki/FreeBSD_version_history) in FreeBSD 3).
* OpenBSD: it seems, there is no such macro (according to 643084843ec54037f498c71c2bc4dd323e320dd0). It either was deleted or was not present at all. We define it in our code, but incorrectly. So our behavior on OpenBSD is incorrect.
* Glibc: the macro was [fixed](https://sourceware.org/bugzilla/show_bug.cgi?id=14347) in 2.17 (the [change](https://github.com/gentoo/glibc/commit/c68257729de8a2165e161b187cf866897f71c502) in a git mirror). CentOS 6 has 2.12, CentOS 7 has 2.17. All (or at least most of) the actual Linux distros has correct definition so.
* Musl: the macro is [incorrect](https://git.musl-libc.org/cgit/musl/tree/include/syslog.h#n21) (despite that it was [reported](https://www.openwall.com/lists/musl/2015/10/06/11) in 2015). (I guess it was forgotten, because the project has no bugtracker.) Alpine is the most popular musl based Linux distribution.

To sum up: OpenBSD and Alpine are affected. CentOS 6 as well. Others likely not.

We shouldn't use the `LOG_MAKEPRI` macro.

Reported by @kolsys in PR #8244.

kolsys:
~Fixes syslog facility. Multiplying does not need there because `LOG_MAKEPRI` will make it byself or defined constants already have bit offset, depending on the OS.~

See the linked issue for the problem description. Affects musl based Linux distros (Apline), distros with glibc prior to 2.17 (CentOS 6) and OpenBSD.

Fixes #8269

### DIFF
--- a/changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
+++ b/changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
@@ -1,0 +1,3 @@
+## bugfix/log
+
+* Fixed incorrect facility value in syslog on Alpine and OpenBSD (gh-8269).
--- a/src/lib/core/say.c
+++ b/src/lib/core/say.c
@@ -1001,12 +1001,8 @@ format_syslog_header(char *buf, int len, int level,
 	int total = 0;
 
 	/* Format syslog header according to RFC */
-	int prio = level_to_syslog_priority(level);
-#if defined(__OpenBSD__)
-#define LOG_MAKEPRI(fac, pri)        (((fac) << 3) | (pri))
-#endif
-	SNPRINT(total, snprintf, buf, len, "<%d>",
-		LOG_MAKEPRI(8 * facility, prio));
+	int prio = (facility << 3) | level_to_syslog_priority(level);
+	SNPRINT(total, snprintf, buf, len, "<%d>", prio);
 	SNPRINT(total, strftime, buf, len, "%h %e %T ", &tm);
 	SNPRINT(total, snprintf, buf, len, "%s[%d]: ", ident,
 		getpid());


Totktonada:
It looks incorrect.

[RFC 5424](https://www.rfc-editor.org/rfc/rfc5424) says (section 6.2.1):

> The Priority value is calculated by first multiplying the Facility
> number by 8 and then adding the numerical value of the Severity.

It is what tarantool is doing here.

The facility values are not pre-multiplied:

https://github.com/tarantool/tarantool/blob/d31796079c477aa7454e6942997efa88550b69f8/src/lib/core/say.h#L118-L144

And the macro doesn't multiply it at least on my OS (Gentoo Linux):

```sh
$ grep '#define.LOG_MAKEPRI' /usr/include/sys/syslog.h 
#define	LOG_MAKEPRI(fac, pri)	((fac) | (pri))
```

Also, there are fails in CI. We don't accept pull requests with failed CI without at least an analysis of the fails. Please, consider reading the [contributing guide](https://github.com/tarantool/tarantool/blob/master/CONTRIBUTING.md).

I suggest to desribe a problem from a user perspective in [issues](https://github.com/tarantool/tarantool/issues) and give us all the details about your system and how you observe a behavior that looks wrong. For now, we don't know anything about the problem you're trying to solve here. A good user story description is a half of the deal.

kolsys:
In Gentoo it is already pre-multiplied.

https://github.com/gentoo/glibc/blob/6ef1b536aa12fb39fcc1f7376f38275d985ed75c/misc/sys/syslog.h#L107

Totktonada:
> In Gentoo it is already pre-multiplied.
> 
> [gentoo/glibc@`6ef1b53`/misc/sys/syslog.h#L107](https://github.com/gentoo/glibc/blob/6ef1b536aa12fb39fcc1f7376f38275d985ed75c/misc/sys/syslog.h#L107)

Of course, that's why the macro is defined this way. But those are not the constants we pass to the macro. You can find them in my message above.

Totktonada:
The `LOG_MAKEPRI` macro was fixed (by eliminating the bit shift) in BSD in 1997 (so the fix likely appears in FreeBSD 3), in glibc 2.17, but the problem is still present in musl. You either use CentOS 6 (unlikely) or Alpine (likely).

Your attempt to fix the problem is incorrect: it breaks the behavior on systems, where the macro has no bit shift (all modern glibc based systems).

I'll file an issue regarding the problem with all the details.

kolsys:
Yes, it's my bad. Tarantool don't uses syslog.h constants. Looks like it don't need to use LOG_MAKEPRI macro in same reasons.
The problem is reproducing in official Docker Alpine container when I see double facility shifting in real syslog packets.

src/lib/core/say.c
```diff
@@ -1001,12 +1001,9 @@ format_syslog_header(char *buf, int len, int level,
 	int total = 0;
 
 	/* Format syslog header according to RFC */
-	int prio = level_to_syslog_priority(level);
-#if defined(__OpenBSD__)
-#define LOG_MAKEPRI(fac, pri)        (((fac) << 3) | (pri))
-#endif
-	SNPRINT(total, snprintf, buf, len, "<%d>",
-		LOG_MAKEPRI(8 * facility, prio));
+	int prio = (((facility) << 3) | level_to_syslog_priority(level));
```
Totktonada, line 1004: Nit: The parentheses around `facility` and around the whole rhs (right hand statement) are redundant.

changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/log
```
Totktonada, line 1: Nits regarding the commit message:

* Subject formatting and wording:
  * `log: Fixed` -> log: fix` (title case -> lower case, past tense -> imperative mood).
  * `in syslog in docker and other cases` -> `in syslog on Alpine and OpenBSD`
  * But the subject line is anyway too long, so I would rephrase it:

    > log: fix syslog facility for musl and OpenBSD
* Typo: `Alipine` -> `Alpine`.
* `bug with syslog severity` -> `bug with syslog priority`.
* `Follow-up #8244` -- it should be removed, GitHub links the PR from the commit on its own.

changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
```diff
@@ -0,0 +1,4 @@
...
+
+* Fixed incorrect facility value in syslog on Alipine, OpenBSD and other
+  cases (gh-8269).
```
Totktonada, line 4: Typo: Alipine.

changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
Totktonada: Fixed.

src/lib/core/say.c
Totktonada: Fixed.

changelogs/unreleased/gh-8269-log-syslog-incorrect-facility.md
Totktonada: Fixed.

Totktonada (approved):

nshy (approved):

Totktonada:
Pushed to master (2.11.0-entrypoint-1025-gacca6d7ac), cherry-picked to 2.10 (2.10.5-entrypoint-132-g2df10ef1d) and 1.10 (1.10.14-71-g640c3d238).

PR #8245, opened 2023-02-02
Fix alias detection in YAML serializer in case `__serialize` method is available

### LINKED ISSUE #8240: YAML serializer doesn't detect aliases in table returned by `__serialize` meta method
Reproduced with Tarantool 2.11.0-entrypoint-959-gfb8d98614652. Tarantool 2.10 is affected as well.

YAML serializer (used by default in Tarantool console) supports aliasing to produce more compact output:

```
tarantool> x = {}
---
...

tarantool> {a = x, b = x}
---
- a: &0 []
  b: *0
...
```

This is useful when dumping space indexes to the console:

```
tarantool> box.space._schema.index
---
- 0: &0
    unique: true
    parts:
    - type: string
      is_nullable: false
      fieldno: 1
    hint: true
    id: 0
    type: TREE
    space_id: 272
    name: primary
  primary: *0
...
```

However, this feature doesn't seem to work for a table returned by the `__serialize` meta method:

```
tarantool> setmetatable({}, {__serialize = function() return {a = x, b = x} end})
---
- a: []
  b: []
...
```

Funny fact: aliasing is detected if the aliased table is referenced in the original table:

```
tarantool> setmetatable({c = x}, {__serialize = function() return {a = x, b = x} end})
---
- a: []
  b: []
...

tarantool> setmetatable({c = x, d = x}, {__serialize = function() return {a = x, b = x} end})
---
- a: &0 []
  b: *0
...
```

locker:
YAML format supports aliases, which are useful in case the same object is referenced more than once (for instance spaces are referenced by name and id in the `box.space` table):

```
tarantool> x = {}
---
...

tarantool> {a = x, b = x}
---
- a: &0 []
  b: *0
...
```

The implementation has two bugs:

1. Aliases aren't detected in objects returned by the `__serialize` method:
   ```
   tarantool> x = {}
   ---
   ...
   
   tarantool> setmetatable({}, {
            >     __serialize = function() return {a = x, b = x} end,
            > })
   ---
   - a: []
     b: []
   ...
   ```

2. An object that implements the `__serialize` method isn't aliased:
   ```
   tarantool> x = setmetatable({}, {
            >     __serialize = function() return {} end,
            > })
   ---
   ...
   
   tarantool> {a = x, b = x}
   ---
   - a: []
     b: []
   ...
   ```

This PR fixes both issue.

Needed for https://github.com/tarantool/tarantool-ee/issues/221
Closes #8240

### DIFF
--- a/third_party/lua-yaml/lyaml.cc
+++ b/third_party/lua-yaml/lyaml.cc
@@ -616,6 +616,8 @@ static int yaml_is_flow_mode(struct lua_yaml_dumper *dumper) {
    return 0;
 }
 
+static void find_references(struct lua_yaml_dumper *dumper);
+
 static int dump_node(struct lua_yaml_dumper *dumper)
 {
    size_t len = 0;
@@ -635,6 +637,8 @@ static int dump_node(struct lua_yaml_dumper *dumper)
 
    int top = lua_gettop(dumper->L);
    luaL_checkfield(dumper->L, dumper->cfg, top, &field);
+   if (field.serialized)
+      find_references(dumper);
    switch(field.type) {
    case MP_UINT:
       snprintf(buf, sizeof(buf) - 1, "%" PRIu64, field.ival);
--- a/third_party/lua-yaml/lyaml.cc
+++ b/third_party/lua-yaml/lyaml.cc
@@ -511,6 +511,8 @@ usage_error:
 static int dump_node(struct lua_yaml_dumper *dumper);
 
 static yaml_char_t *get_yaml_anchor(struct lua_yaml_dumper *dumper) {
+   if (lua_type(dumper->L, -1) != LUA_TTABLE)
+      return NULL;
    const char *s = "";
    lua_pushvalue(dumper->L, -1);
    lua_rawget(dumper->L, dumper->anchortable_index);
@@ -540,11 +542,9 @@ static yaml_char_t *get_yaml_anchor(struct lua_yaml_dumper *dumper) {
    return (yaml_char_t *)s;
 }
 
-static int dump_table(struct lua_yaml_dumper *dumper, struct luaL_field *field){
+static int dump_table(struct lua_yaml_dumper *dumper, struct luaL_field *field,
+                      yaml_char_t *anchor){
    yaml_event_t ev;
-   yaml_char_t *anchor = get_yaml_anchor(dumper);
-
-   if (anchor && !*anchor) return 1;
 
    yaml_mapping_style_t yaml_style = (field->compact)
       ? (YAML_FLOW_MAPPING_STYLE) : YAML_BLOCK_MAPPING_STYLE;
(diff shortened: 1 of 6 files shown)


locker:
@Totktonada noticed that this PR doesn't fix the case when there are inter-references between a table and a nested table with the `__serialize` meta method defined:

```
tarantool> t = {}
---
...

tarantool> {a = t, b = setmetatable({}, {__serialize = function() return {t, t} end})}
---
- a: []
  b:
  - &0 []
  - *0
...
```

However, this would be complicated to fix and we don't really need it for https://github.com/tarantool/tarantool-ee/issues/221 so we agreed to file a separate ticket for this case and fix it later: https://github.com/tarantool/tarantool/issues/8310.

third_party/lua-yaml/lyaml.cc
```diff
@@ -623,6 +618,9 @@ static int dump_node(struct lua_yaml_dumper *dumper)
    bool unused;
    (void) unused;
 
+   yaml_char_t *anchor = get_yaml_anchor(dumper);
+   if (anchor && !*anchor) return 1;
```
Totktonada, line 624: Shouldn't we do it only for `lua_type(dumper->L, -1) == LUA_TTABLE` for performance matters? It doesn't look worthful to check the `anchortable_index` table for numbers, strings and so on (and we can have quite lot of them to serialize).

Totktonada, line 624: (It looks a bit strange that `return` is on the same line as `if`.)

Totktonada (approved):

third_party/lua-yaml/lyaml.cc
locker: You're right. Looking up anything but a table in `anchorable_index` is pointless, because `find_references` stores only tables in it:

https://github.com/tarantool/tarantool/blob/d67962828c0eee224935fbe2af2495c7a122c33b/third_party/lua-yaml/lyaml.cc#L781-L784

I updated `get_yaml_anchor` to bail out if the top element isn't a table:

```diff
diff --git a/third_party/lua-yaml/lyaml.cc b/third_party/lua-yaml/lyaml.cc
index 2137ce665f63..bbbfb677063a 100644
--- a/third_party/lua-yaml/lyaml.cc
+++ b/third_party/lua-yaml/lyaml.cc
@@ -511,6 +511,8 @@ usage_error:
 static int dump_node(struct lua_yaml_dumper *dumper);
 
 static yaml_char_t *get_yaml_anchor(struct lua_yaml_dumper *dumper) {
+   if (lua_type(dumper->L, -1) != LUA_TTABLE)
+      return NULL;
    const char *s = "";
    lua_pushvalue(dumper->L, -1);
    lua_rawget(dumper->L, dumper->anchortable_index);
```

locker: It's because I copied-and-pasted this piece of code. Fixed:

```diff
diff --git a/third_party/lua-yaml/lyaml.cc b/third_party/lua-yaml/lyaml.cc
index bbbfb677063a..762d85bca59b 100644
--- a/third_party/lua-yaml/lyaml.cc
+++ b/third_party/lua-yaml/lyaml.cc
@@ -632,7 +632,8 @@ static int dump_node(struct lua_yaml_dumper *dumper)
    (void) unused;
 
    yaml_char_t *anchor = get_yaml_anchor(dumper);
-   if (anchor && !*anchor) return 1;
+   if (anchor && !*anchor)
+      return 1;
 
    int top = lua_gettop(dumper->L);
    luaL_checkfield(dumper->L, dumper->cfg, top, &field);
```

locker:
Cherry-picked to 2.10.

PR #8247, opened 2023-02-02
sql: add missing malloc failure handling

ImeevMA:
This patch fixes some possible bugs that may occur due to malloc
failure.

Closes https://github.com/tarantool/security/issues/65
Closes https://github.com/tarantool/security/issues/66
Closes https://github.com/tarantool/security/issues/68

### DIFF
--- a/src/box/sql/select.c
+++ b/src/box/sql/select.c
@@ -739,9 +739,16 @@ addWhereTerm(Parse * pParse,	/* Parsing context */
 	assert(pSrc->a[iRight].space != NULL);
 
 	struct Expr *pE1 = sql_expr_new_column(db, pSrc, iLeft, iColLeft);
-	struct Expr *pE2 = sql_expr_new_column(db, pSrc, iRight, iColRight);
-	if (pE1 == NULL || pE2 == NULL)
+	if (pE1 == NULL) {
 		pParse->is_aborted = true;
+		return;
+	}
+	struct Expr *pE2 = sql_expr_new_column(db, pSrc, iRight, iColRight);
+	if (pE2 == NULL) {
+		sqlDbFree(db, pE1);
+		pParse->is_aborted = true;
+		return;
+	}
 	pEq = sqlPExpr(pParse, TK_EQ, pE1, pE2);
 	if (pEq && isOuterJoin) {
 		ExprSetProperty(pEq, EP_FromJoin);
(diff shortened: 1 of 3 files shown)


src/box/sql/select.c
```diff
@@ -740,8 +740,10 @@ addWhereTerm(Parse * pParse,	/* Parsing context */
 
 	struct Expr *pE1 = sql_expr_new_column(db, pSrc, iLeft, iColLeft);
 	struct Expr *pE2 = sql_expr_new_column(db, pSrc, iRight, iColRight);
-	if (pE1 == NULL || pE2 == NULL)
+	if (pE1 == NULL || pE2 == NULL) {
```
locker, line 743: AFAIU if `pE1` allocation succeeds but `pE2` allocations fails, `pE1` will leak. This is hardly possible in reality, but static analyzers may still complain about it.

src/box/sql/select.c
```diff
@@ -2188,22 +2192,25 @@ sqlColumnsFromExprList(Parse * parse, ExprList * expr_list,
 				if (zName[j] == '_')
 					nName = j;
 			}
-			zName =
-			    sqlMPrintf(db, "%.*z_%u", nName, zName, ++cnt);
+			zName = sqlMPrintf(db, "%.*z_%u", nName, zName, ++cnt);
```
locker, line 2195: Just curious: doesn't it leak memory allocated for `zName` with `sqlDbStrDup` above?

src/box/sql/select.c
```diff
@@ -2188,22 +2192,25 @@ sqlColumnsFromExprList(Parse * parse, ExprList * expr_list,
...
+		if (sqlHashInsert(&ht, zName, field) == field) {
+			sql_free(zName);
 			sqlOomFault(db);
-		space_def->fields[i].name = region_alloc(region, name_len + 1);
+			goto cleanup;
+		}
+		space_def->fields[i].name = region_alloc(region, nName + 1);
 		if (space_def->fields[i].name == NULL) {
 			sqlOomFault(db);
 			goto cleanup;
```
locker, line 2210: `zName` isn't freed...

locker (changes requested):
1. Missing references to the tickets you fix. Please add to both the PR and commits.
2. `NO_CHANGLOEG=bugfix` isn't a good excuse not to add a changlog. Let's instead write that we're unlikely to run into this bug in practice.
3. "sql: fix some possible errors" is too vague for a commit title. Let's change it so that it's clear what exactly we're trying to fix here ("sql: add missing malloc failure handling" maybe?)

src/box/sql/select.c
ImeevMA: True, fixed.

src/box/sql/select.c
ImeevMA: No, these values are added to a hash table and removed during cleanup below.

src/box/sql/select.c
ImeevMA: No, this value is added to a hash table and removed during cleanup below.

locker (approved):

PR #8250, opened 2023-02-03
lua: preliminary refactoring for built-in modules overloading [2/2]

Totktonada:
This patchset continues the [series][1] of preliminary commits for implementing the built-in module overriding (so called dual-life modules, see #7774 for the problem statement).

The core idea of the future functionality is that if there is the `override.foo` module on the filesystem, it automatically replaces the built-in module `foo`.

Several changes are needed to implement the overriding:

* Move Lua loaders implementation into its own file `loaders.lua`.
* Split `fio` to `minifio` (needed by the loaders) and `fio`.
* Move the loaders into an early initialization stage.
* Eliminate direct `package.loaded` assignments in src/box/lua/*.lua files.
* Hide built-in module registration process behind a function (to modify it later).

The overriding itself will go as a separate pull request.

Part of #7774

[1]: https://github.com/tarantool/tarantool/pull/8173

### DIFF
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -128,44 +128,62 @@ extern char session_lua[],
 	console_lua[],
 	merger_lua[];
 
+/**
+ * List of box's built-in modules written using Lua.
+ *
+ * Each module is defined as a triplet:
+ *
+ * 1. A file name (without the .lua extension).
+ *
+ *    It is used for prepending error messages and filling
+ *    debug.getinfo() information.
+ * 2. A module name for require().
+ *
+ *    NULL means 'do not register as a module, just execute'.
+ *    Such code shouldn't return any value.
+ *
+ *    Typical NULL usage: code that define functions in
+ *    box or box.internal.
+ * 3. A Lua source code of the module.
+ */
 static const char *lua_sources[] = {
-	"box/session", session_lua,
-	"box/tuple", tuple_lua,
-	"box/schema", schema_lua,
+	"box/session", NULL, session_lua,
+	"box/tuple", NULL, tuple_lua,
+	"box/schema", NULL, schema_lua,
 #if ENABLE_FEEDBACK_DAEMON
 	/*
 	 * It is important to initialize the daemon before
 	 * load_cfg, because the latter picks up some values
 	 * from the feedback daemon.
 	 */
-	"box/feedback_daemon", feedback_daemon_lua,
+	"box/feedback_daemon", NULL, feedback_daemon_lua,
 #endif
 #if ENABLE_SPACE_UPGRADE
 	/*
 	 * Must be loaded after schema_lua, because it redefines
 	 * box.schema.space.upgrade.
 	 */
-	"box/space_upgrade", space_upgrade_lua,
+	"box/space_upgrade", NULL, space_upgrade_lua,
 #endif
 #if ENABLE_AUDIT_LOG
-	"box/audit", audit_lua,
+	"box/audit", "audit", audit_lua,
 #endif
 #if ENABLE_FLIGHT_RECORDER
-	"box/flightrec", flightrec_lua,
+	"box/flightrec", "flightrec", flightrec_lua,
 #endif
 #if ENABLE_READ_VIEW
-	"box/read_view", read_view_lua,
+	"box/read_view", NULL, read_view_lua,
 #endif
 #if ENABLE_SECURITY
-	"box/security", security_lua,
+	"box/security", NULL, security_lua,
 #endif
-	"box/xlog", xlog_lua,
-	"box/upgrade", upgrade_lua,
-	"box/net_box", net_box_lua,
-	"box/console", console_lua,
-	"box/load_cfg", load_cfg_lua,
-	"box/key_def", key_def_lua,
-	"box/merger", merger_lua,
+	"box/xlog", "xlog", xlog_lua,
+	"box/upgrade", NULL, upgrade_lua,
+	"box/net_box", "net.box", net_box_lua,
+	"box/console", "console", console_lua,
+	"box/load_cfg", NULL, load_cfg_lua,
+	"box/key_def", "key_def", key_def_lua,
+	"box/merger", "merger", merger_lua,
 	NULL
 };
 
@@ -574,20 +592,44 @@ box_lua_init(struct lua_State *L)
 	luamp_set_encode_extension(luamp_encode_extension_box);
 	luamp_set_decode_extension(luamp_decode_extension_box);
 
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
+
 	/* Load Lua extension */
-	for (const char **s = lua_sources; *s; s += 2) {
-		const char *modname = *s;
-		const char *modsrc = *(s + 1);
+	for (const char **s = lua_sources; *s; s += 3) {
+		const char *modfile_raw = *s;
+		const char *modname = *(s + 1);
+		const char *modsrc = *(s + 2);
+
 		const char *modfile = lua_pushfstring(L,
-			"@builtin/%s.lua", modname);
+			"@builtin/%s.lua", modfile_raw);
 		if (luaL_loadbuffer(L, modsrc, strlen(modsrc), modfile) != 0 ||
-		    lua_pcall(L, 0, 0, 0) != 0)
+		    lua_pcall(L, 0, 1, 0) != 0)
 			panic("Error loading Lua module %s...: %s",
-			      modname, lua_tostring(L, -1));
+			      modname != NULL ? modname : modfile,
+			      lua_tostring(L, -1));
+
+		if (!lua_isnil(L, -1)) {
+			/* package.loaded.modname = t */
+			assert(modname != NULL);
+			lua_setfield(L, -3, modname);
+		} else {
+			lua_pop(L, 1); /* nil */
+		}
+
 		lua_pop(L, 1); /* modfile */
-		builtin_modcache_put(modname, modsrc);
+
+		/*
+		 * TODO: Use a module name (as written in a
+		 * require() call) as the
+		 * tarantool.debug.getsources() parameter.
+		 *
+		 * For example, "net.box" instead of
+		 * "box/net_box".
+		 */
+		builtin_modcache_put(modfile_raw, modsrc);
 	}
 
+	lua_pop(L, 1); /* _LOADED */
 	assert(lua_gettop(L) == 0);
 }
 
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -592,8 +592,6 @@ box_lua_init(struct lua_State *L)
 	luamp_set_encode_extension(luamp_encode_extension_box);
 	luamp_set_decode_extension(luamp_decode_extension_box);
 
-	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
-
 	/* Load Lua extension */
 	for (const char **s = lua_sources; *s; s += 3) {
 		const char *modfile_raw = *s;
@@ -608,12 +606,16 @@ box_lua_init(struct lua_State *L)
 			      modname != NULL ? modname : modfile,
 			      lua_tostring(L, -1));
 
-		if (!lua_isnil(L, -1)) {
-			/* package.loaded.modname = t */
-			assert(modname != NULL);
-			lua_setfield(L, -3, modname);
+		/*
+		 * Register a built-in module if the module name
+		 * is provided. Otherwise ensure that no value is
+		 * returned.
+		 */
+		if (modname == NULL) {
+			assert(lua_isnil(L, -1));
+			lua_pop(L, 1);
 		} else {
-			lua_pop(L, 1); /* nil */
+			luaT_setmodule(L, modname);
 		}
 
 		lua_pop(L, 1); /* modfile */
@@ -629,7 +631,6 @@ box_lua_init(struct lua_State *L)
 		builtin_modcache_put(modfile_raw, modsrc);
 	}
 
-	lua_pop(L, 1); /* _LOADED */
 	assert(lua_gettop(L) == 0);
 }
 
--- a/src/lua/fio.lua
+++ b/src/lua/fio.lua
@@ -1,5 +1,6 @@
 -- fio.lua (internal file)
 
+local minifio = require('internal.minifio')
 local fio = require('fio')
 local ffi = require('ffi')
 local buffer = require('buffer')
@@ -238,51 +239,20 @@ fio.open = function(path, flags, mode)
     return res
 end
 
-fio.pathjoin = function(...)
-    local i, path = 1, nil
-
-    local len = select('#', ...)
-    while i <= len do
-        local sp = select(i, ...)
-        if sp == nil then
-            error("fio.pathjoin(): undefined path part "..i)
-        end
-
-        sp = tostring(sp)
-        if sp ~= '' then
-            path = sp
-            break
-        else
-            i = i + 1
-        end
-    end
-
-    if path == nil then
-        return '.'
-    end
-
-    i = i + 1
-    while i <= len do
-        local sp = select(i, ...)
-        if sp == nil then
-            error("fio.pathjoin(): undefined path part "..i)
-        end
-
-        sp = tostring(sp)
-        if sp ~= '' then
-            path = path .. '/' .. sp
-        end
-
-        i = i + 1
-    end
-
-    path = path:gsub('/+', '/')
-    if path ~= '/' then
-        path = path:gsub('/$', '')
-    end
-
-    return path
+-- Expose several functions implemented as part of the minifio
+-- module.
+for name, func in pairs(minifio.expose_from_fio) do
+    fio[name] = func
 end
+
+-- List of functions exposed from fio, but defined in minifio.
+--
+-- Without the list a reader should be aware of the fio/minifio
+-- splitting in order to find an implementation of a function by
+-- its name. Let's make a reader's life a bit easier.
+assert(type(fio.cwd) == 'function')
+assert(type(fio.pathjoin) == 'function')
+assert(type(fio.abspath) == 'function')
 
 fio.basename = function(path, suffix)
     if type(path) ~= 'string' then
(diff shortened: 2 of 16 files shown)


Buristan:
Hi, Alexander!
Thanks for the patch-set!
I started to review a draft, AFAICS it's ready for the review:).


**[PATCH 1/7] lua: extract lua loaders code into its own file**

> This commit continues the series of preliminatory commits for

Do you mean "preliminary" here?

src/lua/fio.lua
```diff
@@ -238,51 +239,14 @@ fio.open = function(path, flags, mode)
...
-    if path ~= '/' then
-        path = path:gsub('/$', '')
-    end
-
-    return path
+-- Expose several functions implemented as part of the minifio
+-- module.
+for name, func in pairs(minifio.expose_from_fio) do
+    fio[name] = func
 end
```
Buristan, line 246: Do you move it out of cycle to test that these specific functions are available?

src/lua/minifio.lua
```diff
@@ -0,0 +1,113 @@
+local ffi = require('ffi')
+local minifio = require('internal.minifio')
+
+-- Several file manipulation functions without dependencies on
+-- tarantool's built-in modules to use at early initialization
+-- stage.
+
+ffi.cdef([[
+    char *
```
Buristan, line 9: Minor:
As for me, the following format is more readable:
```c
char *dirname(char *path);
```
Feel free to ignore.

Buristan (commented):
**[PATCH 2/7] lua: split fio to minifio and fio**
LGTM, except a few nits below.

> cwd(), pathjoin() and abspath() are moved to `minifi`o and exposed from

Typo: s/`minifi`o/`minifio`/

Buristan (commented):
**[PATCH 3/7]  lua: use minifio instead of fio in lua loaders**
LGTM, as trivial.

Buristan (commented):
**[PATCH 4/7]  lua: setup loaders before loading built-in modules**
LGTM.

Buristan (commented):
**[PATCH 5/7] lua: assign box modules retvals to package.loaded**
LGTM.

Buristan (commented):
**[PATCH 6/7]  lua: eliminate package.loaded write in box modules**
LGTM, except a few nits regarding the commit message.

> There are several reason to do so:

Typo: s/reason/reasons/

> 3. src/lua/*.lua files are already return module tables instead of

Typo: s/are already return/are already returned/

Buristan (approved):
**[PATCH 7/7]  lua: add second built-in module registration func**
LGTM.

src/lua/fio.lua
Totktonada: Do you mean the assert right below the cycle?

```lua
assert(type(fio.cwd) == 'function')
assert(type(fio.pathjoin) == 'function')
assert(type(fio.abspath) == 'function')
```

It aims two goals:

* Ensure that the functions are there.
* Make them searchable/greppable, when one is looking over the `fio.lua` file: it is not obvious that they're implemented in another place.

src/lua/minifio.lua
Totktonada: I think I should follow our [C Style Guide](https://www.tarantool.io/en/doc/latest/dev_guide/c_style_guide/#chapter-6-functions):

> Note that we place the function return type on the line before the name and signature.

Totktonada:
Rebased on top of current master, fixed the typos pointed by Sergey.

igormunkin (approved):
**[PATCH 1/7] lua: extract lua loaders code into its own file**

@Totktonada, thanks for the patch! LGTM as trivial with one nit below.

src/lua/minifio.lua
```diff
@@ -0,0 +1,113 @@
...
+    return '/' .. table.concat(path_tab, '/')
+end
+
+-- }}} Functions exposed by fio
+
+-- {{{ Functions replaced by fio
+
+-- Similar to fio.dirname, but it doesn't use
+-- cord_ibuf_take()/cord_ibuf_put().
+function minifio.dirname(path)
```
igormunkin, line 91: There is one more thing I can't figure out: why both functions (`cwd` and `dirname`) can't be implemented via the one mechanism (either Lua C or FFI)?

I'd rather move `dirname` to minifio.c either in scope of this patch.

src/lua/fio.lua
igormunkin: I suggest to add the comment with the text of the second bullet to warn future generations (or at least me and @Buristan) of moving these assertions to the loop later.

igormunkin (changes requested):
**[PATCH 2/7] lua: split fio to minifio and fio**

@Totktonada, thanks for the patch! Please, consider my comments below.

---

P.S. There is also a typo in commit message: `s/manipulatoin/manipulation/`.

igormunkin (approved):
**[PATCH 3/7] lua: use minifio instead of fio in lua loaders**

@Totktonada, thanks for the patch! LGTM as trivial.

src/lua/init.c
```diff
@@ -869,28 +912,11 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
...
-		lua_call(L, 1, 1);
-		if (!lua_isnil(L, -1)) {
-			lua_setfield(L, -3, modname); /* package.loaded.modname = t */
-		} else {
-			lua_pop(L, 1); /* nil */
-		}
-		lua_pop(L, 1); /* chunkname */
-		builtin_modcache_put(modname, modsrc);
-
+		luaT_set_module_from_source(L, modname, modsrc);
```
igormunkin, line 911: Looks like you're loading `internal.minifio` and `internal.loaders` once more here, doesn't it?

igormunkin (approved):
**[PATCH 4/7] lua: setup loaders before loading built-in modules**

@Totktonada, thanks for the patch! LGTM, but please consider my comment regarding repeated module loading below.

src/lua/init.c
igormunkin: PEBKAC, I overlooked their deletion from `lua_modules`. Disregard this comment.

src/lua/loaders.lua
```diff
@@ -0,0 +1,166 @@
+local fio = require("fio")
+
+local soext = (jit.os == "OSX" and "dylib" or "so")
```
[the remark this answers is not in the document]
Totktonada: I don't think that I should enhance this code at spliting into separate files. Moreover, any way to obtain `'dylib'` or `'so'` string is likely wrong. AFAIR, we should search as for `.dylib` as well as `.so` on Mac OS, because some modules are compiled into an `.so` file on Mac OS due to a PUC-Rio Lua 5.1.0 problem (it was fixed in later versions, but I guess that some modules are still shipped as `.so`):

> For consistency with other platforms the modules should be installed with the `.so` extension on Mac OS X, too, and this assumed by the default definitions of the package cpath.

http://lua-users.org/wiki/BuildingModules

So if we'll change it, we'll change it more dramatically :)

There are more things we likely should change around the loaders. We aware of them, but I hope we'll do it later in scope of its own tasks.

src/lua/minifio.c
```diff
@@ -0,0 +1,58 @@
...
+ */
+static int
+lbox_minifio_cwd(struct lua_State *L)
+{
+	char *buf = (char *)lua_newuserdata(L, PATH_MAX);
+	if (!buf) {
+		errno = ENOMEM;
+		return luaT_minifio_pushsyserror(L);
+	}
+	if (getcwd(buf, PATH_MAX) == NULL)
```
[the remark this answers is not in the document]
Totktonada: Your points look correct on a first glance, but I would answer here as [above](https://github.com/tarantool/tarantool/pull/8250#discussion_r1102729134): let's work on it somehow separately. Here I'm moving the code. I understand the 'while we're here' reason, but actually I don't know why it is written in the way it is written. I should look down to the history and test different variants. It is not like 'fix a typo in a comment', at least for me.

src/lua/minifio.h
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+struct lua_State;
```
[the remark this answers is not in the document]
Totktonada: I need the type name defined to define a pointer argument. It isn't worth to pull the definition of the type and all the content of the `lua.h` header into this header file. It is usual for header files to define types this way to keep things more clean and to speedup compilation. IOW, I have no goal to avoid this forward declaration, but I think it is a good practice to avoid `#include <lua.h>` here.

src/box/lua/init.c
```diff
@@ -129,43 +129,43 @@ extern char session_lua[],
 	merger_lua[];
 
 static const char *lua_sources[] = {
```
igormunkin, line 131: Minor: It's worth to add a comment regarding this not such obvious structure. However, everybody live without it for many years already, so free to ignore.

igormunkin: Furthermore, what is the rule for setting either name or `NULL`?

src/lua/minifio.lua
Totktonada: And we'll end with the question: why both `dirname()` functions can't be implemented via the one mechanish? :)

...and you'll ask me to rewrite `fio.dirname()` as well. But since it is in C anyway, we can just use `cord_ibuf_*()` in C and implement it once in minifio. But will it better this way? We should perform several performance tests: jit on/off, amd64/arm64.

We should stop at some point to actually solve current task instead of some other one and my copy-and-replace-two-calls variant looks as a good candidate to stop on it.

I see a small gain about existing `minifio.dirname()` implementation: it is very similar to `fio.dirname()`, so they're easy to compare visually.

src/box/lua/init.c
```diff
@@ -574,20 +574,44 @@ box_lua_init(struct lua_State *L)
...
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
+
 	/* Load Lua extension */
-	for (const char **s = lua_sources; *s; s += 2) {
-		const char *modname = *s;
-		const char *modsrc = *(s + 1);
+	for (const char **s = lua_sources; *s; s += 3) {
+		const char *modfile_raw = *s;
+		const char *modname = *(s + 1);
+		const char *modsrc = *(s + 2);
```
igormunkin, line 581: Why do you refuse to introduce a separate function for this also (like you've done in the previous patch)?

igormunkin (approved):
**[PATCH 5/7] lua: assign box modules retvals to package.loaded**

@Totktonada, thanks for the patch! Since all comments below are minor, LGTM.

igormunkin (approved):
**[PATCH 6/7] lua: eliminate package.loaded write in box modules**

@Totktonada, thanks for the patch! LGTM as trivial.

src/box/lua/init.c
Totktonada: The code in src/lua/init.c is used several times, so I moved it into a function. Here we need this code once, so I don't.

I think that I'll end up with a common function in a future, but there are small differences to investigate and eliminate (call/pcall and ones marked with TODO).

src/lua/utils.c
```diff
@@ -348,6 +348,54 @@ luaT_newmodule(struct lua_State *L, const char *modname,
...
+
+	/*
+	 * Verify that the module is not already registered with
+	 * another value.
+	 */
+	assert(lua_isnil(L, -1));
+	lua_pop(L, 1);
+
+	/* Swap package.loaded and the module table. */
+	lua_insert(L, -2);
```
igormunkin, line 387: Technically, it's better to use `lua_pushvalue(L, -2)` here to avoid copying of the top Lua stack slots. As a result you need to pop 2 elements instead of 1 (that is simple decrement of `L->top`) at the end.

src/box/lua/init.c
```diff
@@ -590,12 +588,11 @@ box_lua_init(struct lua_State *L)
 			      modname != NULL ? modname : modfile,
 			      lua_tostring(L, -1));
 
-		if (!lua_isnil(L, -1)) {
-			/* package.loaded.modname = t */
-			assert(modname != NULL);
-			lua_setfield(L, -3, modname);
+		if (modname == NULL) {
```
igormunkin, line 591: I finally understand this condition, but it was kinda challenge for me. Could you please add a comment with clarification here?

igormunkin (approved):
**[PATCH 7/7] lua: add second built-in module registration func**

@Totktonada, thanks for the patch! LGTM, but please consider my nits below.

src/lua/loaders.lua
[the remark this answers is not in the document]
igormunkin: OK, I agree with you now. IIRC, so/dylib problem is in our TODO list for Application Server, so let's solve it there. Resolving this comment.

---

> > For consistency with other platforms the modules should be installed with the `.so` extension on Mac OS X, too, and this assumed by the default definitions of the package cpath.

What a bullshit.

src/lua/minifio.c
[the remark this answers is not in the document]
igormunkin: Alright, that works for me too.

What about "why it is written in this way", I doubt it's necessary to allocate an auxiliary buffer instead of using the static one, but `lua_newuserdata` **never** yields `NULL`, so the latter is simply a bug.

All in all, resolving.

src/lua/minifio.lua
igormunkin: OK, let's leave it as is for now.

I guess we need to collect all these points somewhere near Application Server TODO list, so everyone can make this little maintenance from time to time.

Resolving this also.

src/box/lua/init.c
igormunkin: That works for me, resolving then.

src/lua/minifio.lua
Totktonada: > I guess we need to collect all these points somewhere near Application Server TODO list, so everyone can make this little maintenance from time to time.

I like this idea. I will do something toward this direction.

src/lua/fio.lua
Totktonada: I agree, added a note.

```diff
diff --git a/src/lua/fio.lua b/src/lua/fio.lua
index 6c5c51656..ea66aee22 100644
--- a/src/lua/fio.lua
+++ b/src/lua/fio.lua
@@ -244,6 +244,12 @@ end
 for name, func in pairs(minifio.expose_from_fio) do
     fio[name] = func
 end
+
+-- List of functions exposed from fio, but defined in minifio.
+--
+-- Without the list a reader should be aware of the fio/minifio
+-- spliting in order to find an implementation of a function by
+-- its name. Let's make a reader's life a bit easier.
 assert(type(fio.cwd) == 'function')
 assert(type(fio.pathjoin) == 'function')
 assert(type(fio.abspath) == 'function')
```

src/box/lua/init.c
Totktonada: Fair point. Added:

```diff
diff --git a/src/box/lua/init.c b/src/box/lua/init.c
index 79c6cce17..5eccd6902 100644
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -128,6 +128,24 @@ extern char session_lua[],
        console_lua[],
        merger_lua[];
 
+/**
+ * List of box's built-in modules written using Lua.
+ *
+ * Each module is defined as a triplet:
+ *
+ * 1. A file name (without the .lua extension).
+ *
+ *    It is used for prepending error messages and filling
+ *    debug.getinfo() information.
+ * 2. A module name for require().
+ *
+ *    NULL means 'do not register as a module, just execute'.
+ *    Such code shouldn't return any value.
+ *
+ *    Typical NULL usage: code that define functions in
+ *    box or box.internal.
+ * 3. A Lua source code of the module.
+ */
 static const char *lua_sources[] = {
        "box/session", NULL, session_lua,
        "box/tuple", NULL, tuple_lua,
```

src/box/lua/init.c
Totktonada: I agree, it is not obvious. Added a note.

```diff
diff --git a/src/box/lua/init.c b/src/box/lua/init.c
index 5eccd6902..bdf0aafac 100644
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -606,6 +606,11 @@ box_lua_init(struct lua_State *L)
                              modname != NULL ? modname : modfile,
                              lua_tostring(L, -1));
 
+               /*
+                * Register a built-in module if the module name
+                * is provided. Otherwise ensure that no value is
+                * returned.
+                */
                if (modname == NULL) {
                        assert(lua_isnil(L, -1));
                        lua_pop(L, 1);
```

src/lua/utils.c
Totktonada: Hm. Okay.

I'm bother about readability a bit, but both variants are more or less same in this aspect.

```diff
diff --git a/src/lua/utils.c b/src/lua/utils.c
index ffbbf293f..8becd8da7 100644
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -383,14 +383,17 @@ luaT_setmodule(struct lua_State *L, const char *modname)
        assert(lua_isnil(L, -1));
        lua_pop(L, 1);
 
-       /* Swap package.loaded and the module table. */
-       lua_insert(L, -2);
+       /* Stack: module table, package.loaded. */
 
        /* package.loaded[modname] = <module table> */
+       lua_pushvalue(L, -2);
        lua_setfield(L, -2, modname);
 
-       /* Pop package.loaded. */
-       lua_pop(L, 1);
+       /*
+        * Drop all the temporary values and the module table from
+        * the stack.
+        */
+       lua_pop(L, 2);
 
        say_debug("%s(%s): success", __func__, modname);
        return 0;
```

Totktonada: It seems, it is resolved.

src/lua/loaders.lua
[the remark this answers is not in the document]
Totktonada: I re-verified, it [is](https://www.notion.so/tarantool/so-dylib-package-cpath-contains-only-dylib-on-mac-1a828cc9b4d6490db86e37a91e489572) in the list.

src/lua/minifio.c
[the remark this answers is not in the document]
Totktonada: I [added](https://www.notion.so/tarantool/fio-cwd-implementation-oddity-a9314c7951bd495281da6de4beffd4f0) it to the wishlist.

src/lua/minifio.lua
Totktonada: Tracked [here](https://www.notion.so/tarantool/fio-s-approaches-C-vs-Lua-allocation-mechanism-f508ed23eea34732aa97237100aa9b05) now.

PR #8253, opened 2023-02-03
lua: drop luaL_register_module()

Totktonada:
The function was replaced by luaT_newmodule(). See commit e3cf5a5d636f ("lua: add built-in module registration function") for the motivation and details.

Now all luaL_register_module() usages are eliminated (including Tarantool Enterprise source code) and we can safely drop it.

Part of #7774
Follows up PR #8173

### DIFF
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -307,30 +307,6 @@ luaL_register_type(struct lua_State *L, const char *type_name,
 	lua_pop(L, 1);
 }
 
-void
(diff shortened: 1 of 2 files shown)


igormunkin (approved):
@Totktonada, thanks **a huge** for the patch! LGTM, as trivial.

Buristan (approved):
Hi, Alexander!
Thanks for the patch!
LGTM!

PR #8257, opened 2023-02-06
memtx: fix force recovery handling

### LINKED ISSUE #7974: `box.cfg.force_recovery` doesn't work for snap file if there are no user spaces
Commit 0be1243ecc8a0b56b079d82d4c713982a05382f0 "improved" the behavior of [`box.cfg.force_recovery`](https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-binary-logging-snapshots-force-recovery) so that now it simply doesn't work if the snap file doesn't have user spaces. This is bad. Consider the use case:
1. The user runs Tarantool version X. The server is used as a pure application server so there's no user spaces in the snap file.
2. The user decides to upgrade to Tarantool version Y, where we added a new request type, which is written to the snap file for some reason (this could be [`IPROTO_RAFT_PROMOTE`](https://github.com/tarantool/tarantool/blob/e4f1c891bc1cf50828bbeedbbe1b102b2273e618/src/box/iproto_constants.h#L254)). After upgrading, the server runs for a while, writing snap files periodically, as scheduled. The user does **not** call [`box.schema.upgrade()`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/upgrade/) just in case they decide to downgrade.
3. Suddenly, the user decides that Tarantool version Y performs worse than Tarantool version X and decides to downgrade. However, Tarantool version X fails to recover from the snap file written by Tarantool version Y because of the new request type. The user tries to start with `box.cfg.force_recovery` set to true, but the server still fails.

How to reproduce:
1. Create a new server using Tarantool 2.7.2
   ```Lua
   box.cfg{}
   box.space._schema:insert{'foo', 'bar'} -- bump lsn to make a snap
   ```
2. Restart with Tarantool 2.8+ in read-only mode (to avoid auto call of `box.schema.upgrade()`) and make a snapshot
   ```Lua
   box.cfg{read_only = true}
   box.snapshot()
   ```
3. Try to restart with Tarantool 2.7.2 from the snapshot written by Tarantool 2.8+

Result:
```
Tarantool 2.7.2-0-g4d8c06890187
tarantool> box.cfg{force_recovery = true}
2022-11-25 19:55:31.684 [1406905] main/103/interactive C> Tarantool 2.7.2-0-g4d8c06890187
2022-11-25 19:55:31.684 [1406905] main/103/interactive C> log level 5
2022-11-25 19:55:31.684 [1406905] main/103/interactive I> wal/engine cleanup is paused
2022-11-25 19:55:31.685 [1406905] main/103/interactive I> mapping 268435456 bytes for memtx tuple arena...
2022-11-25 19:55:31.685 [1406905] main/103/interactive I> Actual slab_alloc_factor calculated on the basis of desired slab_alloc_factor = 1.044274
2022-11-25 19:55:31.685 [1406905] main/103/interactive I> mapping 134217728 bytes for vinyl tuple arena...
2022-11-25 19:55:31.695 [1406905] main/103/interactive I> instance uuid 97b8d9c4-4744-46e0-92cd-b6c2770b9278
2022-11-25 19:55:31.695 [1406905] main/103/interactive I> instance vclock {1: 1}
2022-11-25 19:55:31.696 [1406905] main/103/interactive I> recovery start
2022-11-25 19:55:31.696 [1406905] main/103/interactive I> recovering from `./00000000000000000001.snap'
2022-11-25 19:55:31.718 [1406905] main/103/interactive I> cluster uuid 29853d59-6814-429b-813c-a0bf363d0ec7
2022-11-25 19:55:31.734 [1406905] main/103/interactive I> assigned id 1 to replica 97b8d9c4-4744-46e0-92cd-b6c2770b9278
2022-11-25 19:55:31.734 [1406905] main/103/interactive memtx_engine.c:236 E> ER_UNKNOWN_REQUEST_TYPE: Unknown request type 31
2022-11-25 19:55:31.734 [1406905] main/103/interactive F> can't initialize storage: Unknown request type 31
```

(Note, `box.cfg.force_recovery` doesn't help)

Should you create a new user space at step 1 and write something to it (instead of bumping lsn by writing to a system space), the server still prints the error to the log, but successfully recovers in the `force_recovery` mode.

CuriousGeorgiy:
Closes #7974


test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,82 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+-- Checks that force recovery works with snapshot containing no user spaces
+-- and an unknown request type.
+g.test_uknown_request_type_force_recovery = function()
+    local datadir = 'test/box-luatest/gh_7974_data/unknown_request_type'
```
locker, line 9: What bothers me is that it's unclear what these snapshot files contain and how they were created so one wouldn't know what to do if they want to add another test case or update the snapshots for some reason. It'd be nice to have a script that would regenerate these files, like [here](https://github.com/tarantool/tarantool/tree/master/test/xlog/upgrade/2.1.3/gh-4771-upgrade-sequence), or at least a human-readable instruction. The problem is that we don't have an xlog writer Lua API so that would require writing the generator in C. I'm fine to ignore this if everyone else thinks it's okay to leave it as is. I'd ask what @Totktonada thinks, because he worked a lot on the tests in the past.

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,82 @@
...
+            })
+    t.assert_error_msg_contains(
+            'Timed out to wait for "server is ready" condition for server',
+            function()
+                s:start()
+            end)
+    s:clean()
+end
+
+-- Checks that force recovery does not work with snapshot containing one valid
```
Gumix, line 70: ~does not work~ works

CuriousGeorgiy: Fixed, thanks!

Gumix (approved):

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
Totktonada: I'm out of context here. Can we generate the snapshot and post-process it somehow? If so, I would point to my recent [fill.lua](https://github.com/tarantool/tarantool/blob/c066f63b5f40d0a3e4c2f6d1d5232924cd7671f4/test/box-luatest/gh_4691_data/fill.lua) as an example that I feel good. If it is complicated, I would vote for an instruction for a human.

CuriousGeorgiy: @Totktonada we need snapshots with invalid rows to test force recovery. Basically, one needs to manually patch `checkpoint_f` to generate an invalid one. @locker suggests writing a snapshot (i.e., xrow) generation utility instead).

Totktonada: It would be nice to have such a tool. While we there is no one I would write an instruction for a human.

CuriousGeorgiy: @Totktonada thanks for your insight!

CuriousGeorgiy: @locker I added snapshot generation instructions to each of the test cases, please review.

locker (changes requested):

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,129 @@
...
+-- Checks that force recovery does not work with snapshot containing one
+-- corrupted user request.
+--
+-- Snapshot generation instruction:
+-- 1. Patch this place to change the user space's ID to an invalid one (e.g.,
+--    777):
+-- luacheck: no max comment line length
+-- https://github.com/tarantool/tarantool/blob/cba4e20aa2600713b55bdaad6fd84396257835f2/src/box/memtx_engine.cc#L856.
+-- For instance, add this snippet:
+-- `if (space_rv->id == 512) space_rv->id = 777;`;
```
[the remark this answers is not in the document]
CuriousGeorgiy: Discussed with @locker and @alyapunov: decided to distinguish invalid requests from corrupted requests — an invalid request to non-system space allows force recovery.

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,129 @@
...
+-- 4. Call `box.snapshot`.
+g.test_second_user_space_request_force_recovery = function()
+    local datadir = 'test/box-luatest/gh_7974_data/second_user_space_request'
+    local s = server:new(
+            {
+                alias = 'dflt',
+                box_cfg = {force_recovery = true},
+                datadir = datadir,
+            })
+    s:start()
```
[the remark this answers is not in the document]
CuriousGeorgiy: Done.

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,129 @@
...
+g.test_first_non_insert_request_force_recovery = function()
+    local datadir = 'test/box-luatest/gh_7974_data/first_non_insert_request'
+    local s = server:new(
+            {
+                alias = 'dflt',
+                box_cfg = {force_recovery = true},
+                datadir = datadir,
+            })
+    t.assert_error_msg_contains(
+            'Timed out to wait for "server is ready" condition for server',
```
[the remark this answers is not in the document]
CuriousGeorgiy: Thanks for pointing this technique out — fixed.

src/box/memtx_engine.cc
```diff
@@ -333,7 +346,7 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 	RegionGuard region_guard(&fiber()->gc);
 	if (xrow_decode_dml(row, &request, dml_request_key_map(row->type)) != 0)
 		return -1;
-	*is_space_system = (request.space_id < BOX_SYSTEM_ID_MAX);
+	*is_system_space_row = request.space_id < BOX_SYSTEM_ID_MAX;
```
Gumix, line 349: That check was a bit incorrect, it should compare with BOX_SYSTEM_ID_MIN too.

CuriousGeorgiy: Fixed, thanks for noticing!

src/box/memtx_engine.cc
```diff
@@ -282,7 +293,7 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 	 * Snapshot entries are ordered by the space id, it means that if there
 	 * are no spaces, then all system spaces are definitely missing.
 	 */
-	if (is_space_system < 0)
+	if (!any_system_space)
```
locker, line 296: I don't understand: does this mean that if we recovered at least one system space, we assume everything's fine? I thought we wanted to ensure that we recovered at least one record unrelated to system spaces.

Anyway, it'd probably be better to move the flag update completely to `memtx_engine_recovery_snapshot_row`. I mean something like this:

```diff
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 8fdcb1b78ea6..001a5467687d 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -218,7 +218,8 @@ memtx_engine_shutdown(struct engine *engine)
 
 static int
 memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
-				  struct xrow_header *row, int *is_space_system);
+				  struct xrow_header *row,
+				  bool *recovered_system_spaces);
 
 int
 memtx_engine_recover_snapshot(struct memtx_engine *memtx,
@@ -238,16 +239,16 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 	int rc;
 	struct xrow_header row;
 	uint64_t row_count = 0;
-	int is_space_system = -1;
 	bool force_recovery = false;
+	bool recovered_system_spaces = false;
 	/*
 	 * In case when we read system space, we can't ignore errors.
 	 */
 	while ((rc = xlog_cursor_next(&cursor, &row, force_recovery)) == 0) {
 		row.lsn = signature;
-		rc = memtx_engine_recover_snapshot_row(memtx, &row,
-						       &is_space_system);
-		force_recovery = is_space_system == 0 ?
+		rc = memtx_engine_recover_snapshot_row(
+			memtx, &row, &recovered_system_spaces);
+		force_recovery = recovered_system_spaces ?
 				 memtx->force_recovery : false;
 		if (rc < 0) {
 			if (!force_recovery)
@@ -282,7 +283,7 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 	 * Snapshot entries are ordered by the space id, it means that if there
 	 * are no spaces, then all system spaces are definitely missing.
 	 */
-	if (is_space_system < 0)
+	if (!recovered_system_spaces)
 		panic("snapshot `%s' has no system spaces", cursor.name);
 
 	return 0;
@@ -317,10 +318,12 @@ memtx_engine_recover_synchro(const struct xrow_header *row)
 
 static int
 memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
-				  struct xrow_header *row, int *is_space_system)
+				  struct xrow_header *row,
+				  bool *recovered_system_spaces)
 {
 	assert(row->bodycnt == 1); /* always 1 for read */
 	if (row->type != IPROTO_INSERT) {
+		*recovered_system_spaces = true;
 		if (row->type == IPROTO_RAFT)
 			return memtx_engine_recover_raft(row);
 		if (row->type == IPROTO_RAFT_PROMOTE)
@@ -333,7 +336,9 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 	RegionGuard region_guard(&fiber()->gc);
 	if (xrow_decode_dml(row, &request, dml_request_key_map(row->type)) != 0)
 		return -1;
-	*is_space_system = (request.space_id < BOX_SYSTEM_ID_MAX);
+	if (reqeust.space_id < BOX_SYSTEM_ID_MIN ||
+	    request.space_id > BOX_SYSTEM_ID_MAX)
+		*recovered_system_spaces = true;
 	struct space *space = space_cache_find(request.space_id);
 	if (space == NULL)
 		goto log_request;
```

</details>

src/box/memtx_engine.cc
```diff
@@ -216,9 +216,17 @@ memtx_engine_shutdown(struct engine *engine)
 	free(memtx);
 }
 
+/**
+ * Recovers one xrow from snapshot.
+ *
+ * @retval -1 general error, diagnostic set
+ * @retval 0 success
+ * @retval 1 unknown request type, diagnostic set
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed, now this function has only two possible return values.

src/box/memtx_engine.cc
CuriousGeorgiy: > does this mean that if we recovered at least one system space, we assume everything's fine

No, this flag is needed for the cases we don't find any system spaces in the snapshot, please see @Gumix's patch e0a9aed4.

CuriousGeorgiy: Thanks for the idea, the third flag is indeed redundant, applied your suggestion!

src/box/space.h
```diff
@@ -545,7 +545,7 @@ space_is_vinyl(const struct space *space)
  * meaning for tarantool and has predefined insert/remove triggers.
  */
 bool
-space_is_system(const struct space *space);
+space_is_system(uint32_t space_id);
```
locker, line 548: Judging by its name, `space_is_system()` is supposed to take a space, not id. Anyway, this function isn't very convenient to use in case we have a space. It's better to introduce a new helper function `space_id_is_system()` instead of changing the signature of `space_is_system`. This change had better be squashed into the patch that needs this function.

locker (changes requested):

src/box/space.h
CuriousGeorgiy: > This change had better be squashed into the patch that needs this function.

Why? This an independent change.

src/box/memtx_engine.cc
CuriousGeorgiy: Again, your patch doesn't work — there can be a snapshot that only has user spaces, and storage initialization will not fail. We need a separate flag to track if we recovered a system space.

src/box/space.h
locker: It's okay to leave it in a separate patch. However, come to think of it, the new function belongs in `schema_def.h`, because it only depends on `BOX_SYSTEM_ID_MIN` and `BOX_SYSTEM_ID_MAX` constants, not on `struct space`. Please move it to `schema_def.h` and make it inline.

CuriousGeorgiy: Good point! Done.

src/box/memtx_engine.cc
locker: A enum and a flag still look ugly. I think it's possible to use only enum. What about something like this?

```diff
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 8fdcb1b78ea6..43bc4234c278 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -216,9 +216,16 @@ memtx_engine_shutdown(struct engine *engine)
 	free(memtx);
 }
 
+enum snapshot_recovery_phase {
+	SNAPSHOT_RECOVERY_NOT_STARTED,
+	RECOVERING_SYSTEM_SPACES,
+	DONE_RECOVERING_SYSTEM_SPACES,
+};
+
 static int
 memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
-				  struct xrow_header *row, int *is_space_system);
+				  struct xrow_header *row,
+				  enum snapshot_recovery_phase *phase);
 
 int
 memtx_engine_recover_snapshot(struct memtx_engine *memtx,
@@ -238,17 +245,16 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 	int rc;
 	struct xrow_header row;
 	uint64_t row_count = 0;
-	int is_space_system = -1;
 	bool force_recovery = false;
+	bool snapshot_recovery_phase phase = SNAPSHOT_RECOVERY_NOT_STARTED;
 	/*
 	 * In case when we read system space, we can't ignore errors.
 	 */
 	while ((rc = xlog_cursor_next(&cursor, &row, force_recovery)) == 0) {
 		row.lsn = signature;
-		rc = memtx_engine_recover_snapshot_row(memtx, &row,
-						       &is_space_system);
-		force_recovery = is_space_system == 0 ?
-				 memtx->force_recovery : false;
+		rc = memtx_engine_recover_snapshot_row(memtx, &row, &phase);
+		if (phase >= DONE_RECOVERING_SYSTEM_SPACES)
+			force_recovery = true;
 		if (rc < 0) {
 			if (!force_recovery)
 				break;
@@ -277,14 +283,6 @@ memtx_engine_recover_snapshot(struct memtx_engine *memtx,
 		else
 			say_error("snapshot `%s' has no EOF marker", cursor.name);
 	}
-
-	/*
-	 * Snapshot entries are ordered by the space id, it means that if there
-	 * are no spaces, then all system spaces are definitely missing.
-	 */
-	if (is_space_system < 0)
-		panic("snapshot `%s' has no system spaces", cursor.name);
-
 	return 0;
 }
 
@@ -315,12 +313,26 @@ memtx_engine_recover_synchro(const struct xrow_header *row)
 	return txn_limbo_process(&txn_limbo, &req);
 }
 
+static inline int
+complete_system_space_recovery(enum snapshot_recovery_phase *phase)
+{
+	if (*phase < RECOVERING_SYSTEM_SPACES) {
+		diag_set(ClientError, ER_MISSING_SYSTEM_SPACES);
+		return -1;
+	}
+	*phase = DONE_RECOVERING_SYSTEM_SPACES;
+	return 0;
+}
+
 static int
 memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
-				  struct xrow_header *row, int *is_space_system)
+				  struct xrow_header *row,
+				  enum snapshot_recovery_phase *phase)
 {
 	assert(row->bodycnt == 1); /* always 1 for read */
 	if (row->type != IPROTO_INSERT) {
+		if (complete_system_space_recovery(phase) != 0)
+			return -1;
 		if (row->type == IPROTO_RAFT)
 			return memtx_engine_recover_raft(row);
 		if (row->type == IPROTO_RAFT_PROMOTE)
@@ -333,7 +345,9 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 	RegionGuard region_guard(&fiber()->gc);
 	if (xrow_decode_dml(row, &request, dml_request_key_map(row->type)) != 0)
 		return -1;
-	*is_space_system = (request.space_id < BOX_SYSTEM_ID_MAX);
+	if (!space_id_is_system(request.space_id) &&
+	    complete_system_space_recovery(phase) != 0)
+		return -1;
 	struct space *space = space_cache_find(request.space_id);
 	if (space == NULL)
 		goto log_request;
```

</details>

src/box/errcode.h
```diff
@@ -318,6 +318,7 @@ struct errcode_record {
 	/*263 */_(ER_BOOTSTRAP_CONNECTION_NOT_TO_ALL, "Some replica set members were not specified in box.cfg.replication") \
 	/*264 */_(ER_NIL_UUID,			"Nil UUID is reserved and can't be used in replication") \
 	/*265 */_(ER_WRONG_FUNCTION_OPTIONS,	"Wrong function options: %s") \
+	/*266 */_(ER_MISSING_SYSTEM_SPACES,	"snapshot has no system spaces") \
```
locker, line 321: Nit: All other error messages start with a capital letter. Please capitalize this one too for consistency.

src/box/memtx_engine.cc
```diff
@@ -333,7 +363,12 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 	RegionGuard region_guard(&fiber()->gc);
 	if (xrow_decode_dml(row, &request, dml_request_key_map(row->type)) != 0)
 		return -1;
-	*is_space_system = (request.space_id < BOX_SYSTEM_ID_MAX);
+	if (!space_id_is_system(request.space_id)) {
+		if (complete_system_space_recovery(state) != 0)
+			return -1;
```
locker, line 368: This return statement isn't covered by any tests. Please add a test for the case when a snapshot starts with a user space row.

src/box/memtx_engine.cc
```diff
@@ -333,7 +363,12 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
...
+	} else {
+		*state = RECOVERING_SYSTEM_SPACES;
```
locker, line 370: The code'd look nicer if the state was updated in one place, because one could see the state-diagram and expectations by looking at just one small function:

```diff
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 2fdaf29fd3f0..ce8f687ce633 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -231,6 +231,29 @@ enum snapshot_recovery_state {
 	DONE_RECOVERING_SYSTEM_SPACES,
 };
 
+/** Checks and updates the snapshot recovery state. */
+static int
+snapshot_recovery_state_update(enum snapshot_recovery_state *state,
+			       bool is_system_space_row)
+{
+	switch (*state) {
+	case SNAPSHOT_RECOVERY_NOT_STARTED:
+		if (!is_system_space_row) {
+			diag_set(ClientError, ER_MISSING_SYSTEM_SPACES);
+			return -1;
+		}
+		*state = RECOVERING_SYSTEM_SPACES;
+		break;
+	case RECOVERING_SYSTEM_SPACES:
+		if (!is_system_space_row)
+			*state = DONE_RECOVERING_SYSTEM_SPACES;
+		break;
+	case DONE_RECOVERING_SYSTEM_SPACES:
+		break;
+	}
+	return 0;
+}
+
 /**
  * Recovers one xrow from snapshot.
  *
@@ -328,20 +351,6 @@ memtx_engine_recover_synchro(const struct xrow_header *row)
 	return txn_limbo_process(&txn_limbo, &req);
 }
 
-/**
- * The snapshot must contain system spaces.
- */
-static inline int
-complete_system_space_recovery(enum snapshot_recovery_state *state)
-{
-	if (*state == SNAPSHOT_RECOVERY_NOT_STARTED) {
-		diag_set(ClientError, ER_MISSING_SYSTEM_SPACES);
-		return -1;
-	}
-	*state = DONE_RECOVERING_SYSTEM_SPACES;
-	return 0;
-}
-
 static int
 memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 				  struct xrow_header *row,
@@ -349,7 +358,8 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 {
 	assert(row->bodycnt == 1); /* always 1 for read */
 	if (row->type != IPROTO_INSERT) {
-		if (complete_system_space_recovery(state) != 0)
+		if (snapshot_recovery_state_update(
+				state, /*is_system_space_row=*/false) != 0)
 			return -1;
 		if (row->type == IPROTO_RAFT)
 			return memtx_engine_recover_raft(row);
@@ -363,12 +373,9 @@ memtx_engine_recover_snapshot_row(struct memtx_engine *memtx,
 	RegionGuard region_guard(&fiber()->gc);
 	if (xrow_decode_dml(row, &request, dml_request_key_map(row->type)) != 0)
 		return -1;
-	if (!space_id_is_system(request.space_id)) {
-		if (complete_system_space_recovery(state) != 0)
-			return -1;
-	} else {
-		*state = RECOVERING_SYSTEM_SPACES;
-	}
+	bool is_system_space_row = space_id_is_system(request.space_id);
+	if (snapshot_recovery_state_update(state, is_system_space_row) != 0)
+		return -1;
 	struct space *space = space_cache_find(request.space_id);
 	if (space == NULL)
 		goto log_request;
@@ -666,10 +673,10 @@ memtx_engine_bootstrap(struct engine *engine)
 		return -1;
 
 	int rc;
-	enum snapshot_recovery_state unused;
 	struct xrow_header row;
+	enum snapshot_recovery_state state = SNAPSHOT_RECOVERY_NOT_STARTED;
 	while ((rc = xlog_cursor_next(&cursor, &row, true)) == 0) {
-		rc = memtx_engine_recover_snapshot_row(memtx, &row, &unused);
+		rc = memtx_engine_recover_snapshot_row(memtx, &row, &state);
 		if (rc < 0)
 			break;
 	}
```

</details>

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -0,0 +1,189 @@
...
+    s = server:new(
+        {
+            alias = 'fail',
+            box_cfg = {force_recovery = false},
+            datadir = datadir,
+        })
+    s:start{wait_until_ready = false}
+    local log = fio.pathjoin(s.workdir, s.alias .. '.log')
+    t.helpers.retrying({}, function()
+        t.assert_not_equals(s:grep_log("can't initialize storage: " ..
```
locker, line 37: Nit: You could use the plain assert here:

```Lua
t.assert(s:grep_log(...))
```

locker (changes requested):

test/box-luatest/gh_7974_force_recovery_bugs_test.lua
CuriousGeorgiy: As per our Lua style guide[^1]:

> Avoid implicit casting to boolean in if conditions like `if x then` or `if not x then`. Such expressions will likely result in troubles with `[box.NULL]`(https://www.tarantool.io/ru/doc/latest/reference/reference_lua/box_null/). Instead of those conditions, use `if x ~= nil` then and `if x == nil` then.

[^1]: https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/#idioms-and-patterns

locker: `grep_log` never returns `box.NULL` (it returns nil or a string) so I don't see much point to be verbose here. We use plain `t.assert()` with `grep_log` in other places.

CuriousGeorgiy: It's a matter of consistency.

src/box/errcode.h
CuriousGeorgiy: Fixed.

src/box/memtx_engine.cc
CuriousGeorgiy: Added two more test cases. Actually, removing the check at the end of `memtx_engine_recovery_snapshot` was wrong — it misses the cases when the snapshot is completely empty.

src/box/memtx_engine.cc
CuriousGeorgiy: Thanks, applied your suggestion!

CuriousGeorgiy:
@locker checkpatch errors are FP.

locker:
> @locker checkpatch errors are FP.

Added snap and xlog files to `.gitattributes` to fix this.

locker (approved):

locker:
ASAN workflow fails. Please fix.

locker:
> ASAN workflow fails. Please fix.

Never mind - the fix is trivial:

```diff
diff --git a/src/box/memtx_engine.cc b/src/box/memtx_engine.cc
index 713ad0838498..934b1353b5ad 100644
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -679,10 +679,10 @@ memtx_engine_bootstrap(struct engine *engine)
 		return -1;
 
 	int rc;
-	enum snapshot_recovery_state unused;
 	struct xrow_header row;
+	enum snapshot_recovery_state state = SNAPSHOT_RECOVERY_NOT_STARTED;
 	while ((rc = xlog_cursor_next(&cursor, &row, true)) == 0) {
-		rc = memtx_engine_recover_snapshot_row(memtx, &row, &unused);
+		rc = memtx_engine_recover_snapshot_row(memtx, &row, &state);
 		if (rc < 0)
 			break;
 	}
```

I amended the last commit.

CuriousGeorgiy:
@locker thanks for the fix!❤️

locker:

PR #8258, opened 2023-02-06
datetime: fix precision overlflow

Serpentian:
We didn't take into consideration the fact, that precision
value passed to control the width of nanoseconds part in
datetime_object:format could be more than maximum positive
value, integer may have. Currently it leads to segfault.

```
tarantool> require('datetime').new{}:format('%2147483648f')
```

We should check errno in order to find out, if overflow
occurs. The problem is the fact, that `width` variable must
have int type due to snprintf requirements ("%*d") and
strtol returns long. Errno won't be set if returned value
is in bounds [INT_MAX, LONG_MAX], but it will overflow
int resulting in inconsistent behavior.

So, let's save the result of strotl to the temp value.
If this value doesn't belong to the above-mentioned set,
or errno was set, we assign to `width` maximum value, it
may have: 9.

Closes https://github.com/tarantool/security/issues/31

NO_DOC=bugfix

### DIFF
--- a/src/lib/tzcode/strftime.c
+++ b/src/lib/tzcode/strftime.c
@@ -41,6 +41,7 @@
 #include "timelocal.h"
 
 #include <assert.h>
+#include <limits.h>
 #include <locale.h>
 #include <stdbool.h>
 #include <stdio.h>
@@ -277,8 +278,21 @@ _fmt(char *buf, ssize_t size, const char *format, const struct tnt_tm *t,
 						format - mod_len);
 					continue;
 				}
-				width = (int)strtol(format - mod_len,
-						    (char **)NULL, 10);
+				/*
+				 ** We cannot cast returned value of strtol
+				 ** to int in place, as the returned value has
+				 ** long type: so, when width is more then
+				 ** INT_MAX but less then LONG_MAX,
+				 ** no errno will be set, we won't distinguish
+				 ** if overflow happens.
+				 */
+				errno = 0;
+				long tmp_width = strtol(format - mod_len,
+							(char **)NULL, 10);
+				width = (int)tmp_width;
+				if (tmp_width < INT_MIN || tmp_width > INT_MAX ||
+				    errno != 0 || width < 0 || width > 9)
+					width = 9;
 				/* fallthru */
 			case 'f': {
 				int nsec = t->tm_nsec;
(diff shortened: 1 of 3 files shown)


src/lib/tzcode/strftime.c
```diff
@@ -300,7 +300,15 @@ _fmt(char *buf, ssize_t size, const char *format, const struct tnt_tm *t,
...
-					width = width >= 9 ? 9 : width;
+					/*
+					 ** If width is more then some value
+					 ** (most often it equals to 2^31 - 1)
+					 ** then overflow happens. Let it fail
+					 ** and assign to width maximum value
+					 ** it may have in order to be consistent
+					 ** with e.g '%50f'.
+					 */
+					width = width >= 9 || width < 0 ? 9 : width;
```
locker, line 311: Looks inconsistent:

```
$ tarantool -e "print(require('datetime').new{}:format('%64424509541f'))"
000000000
$ tarantool -e "print(require('datetime').new{}:format('%64424509441f'))"
0
```

Let's detect overflow by looking at `strtol` return code and/or errno, like we do here, for example:

https://github.com/tarantool/tarantool/blob/d31796079c477aa7454e6942997efa88550b69f8/src/lib/vclock/vclock.c#L100-L103

locker (changes requested):

src/lib/tzcode/strftime.c
Serpentian: > Looks inconsistent

That's really strange if it works like that on your machine. I get the same output:

```
┌─> Patched version
└─>  | ./tarantool -e "print(_TARANTOOL); print(require('datetime').new{}:format('%64424509541f'))"
2.11.0-entrypoint-969-g87f60dd42
000000000

┌─> Non-patched
└─>  | tarantool -e "print(_TARANTOOL); print(require('datetime').new{}:format('%64424509541f'))"
2.11.0-entrypoint-968-gd31796079
000000000
```

Here's the explanation:
```
; 64424509541 - 2^32 * 15
	101
```

So, the width equals to 101, old part of code works, replacing width with 9. Inconsistency is not possible, AFAICS. 

However, I won't insist. If we want to process errno, I suppose we shouldn't throw an error, but just replace `width` with 9. Am I right?

locker: I ran the *same* patched build, but with different inputs (note 541 vs 441):

```
%64424509541f
%64424509441f
```

I'd expect it to return `000000000` in both cases, but it returns `0` in the second case. That's what I meant by "inconsistent behavior".

> If we want to process errno, I suppose we shouldn't throw an error, but just replace `width` with 9. Am I right?

Yes.

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8259, opened 2023-02-06
sql: fix NULL dereference in where.c

Lord-KA:
There was a ambitious pointer deref in where_loop_builder_shortcut(), that wasn't visible due to dead code.

Removed the dead code and added assert before dereference.

NO_TEST=refactoring
NO_DOC=refactoring
NO_CHANGELOG=refactoring

### DIFF
--- a/src/box/sql/where.c
+++ b/src/box/sql/where.c
@@ -4036,8 +4036,6 @@ where_loop_builder_shortcut(struct WhereLoopBuilder *builder)
 		return 0;
 	assert(where_info->pTabList->nSrc >= 1);
 	struct SrcList_item *item = where_info->pTabList->a;
-	struct space_def *space_def = item->space->def;
-	assert(space_def != NULL);
 	if (item->fg.isIndexedBy)
 		return 0;
 	int cursor = item->iCursor;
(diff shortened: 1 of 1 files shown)


src/box/sql/where.c
```diff
@@ -4059,22 +4057,17 @@ where_loop_builder_shortcut(struct WhereLoopBuilder *builder)
...
-				if (!idx_def->opts.is_unique)
-					continue;
-				if (where_loop_assign_terms(loop, clause,
-							    cursor, space_def,
-							    idx_def) == 0)
-					break;
-			}
-		} else {
-			/* Space is ephemeral. */
-			assert(space_def->id == 0);
```
ImeevMA, line 4075: This line looks rather strange. I looked at earlier commits and found that `space` was first obtained using [`space_cache_find()`](https://github.com/tarantool/tarantool/commit/a40287051f4d80851c770bfb3d160b85d94f193f) which would return `NULL` for ephemeral spaces. This was later changed to the [current version](https://github.com/tarantool/tarantool/commit/ff75878ab235119bfa067fbf342ceb5773203492), however the new version does not properly check `space` for being ephemeral. So most likely the code is not working as intended. I think a more appropriate solution would be to check if the `space->def->id` is `0` instead of a `space! = NULL`. However, all the mentioned changes happened 5 years ago, so it is entirely possible that the original design will not be applicable. It should probably be investigated.

ImeevMA (changes requested):
Hi! Thank you for the patch! I have a few comments below. Also, I think the current commit title is not appropriate as `fix` assumes the bug will be fixed, which is definitely not a refactoring. Also, I think the code became dead due to an error, so it might be better to restore it instead of removing.

PR #8262, opened 2023-02-07
bitset: fix index size calculation

### LINKED ISSUE #5809: Segmentation fault or error message if bitset index exists
I have Ubuntu 20.04. I have Tarantool 2.7 pulled from source today.

I make these requests:
```lua
box.execute([[CREATE TABLE j (s1 INT PRIMARY KEY, s2 VARBINARY NOT NULL);]])
box.space.J:create_index('J',{type='bitset'})
box.execute([[SELECT * FROM j NOT INDEXED WHERE s2 > x'00';]])
```
Result: Segmentation fault.

Then I destroy the database, restart, and make these requests:
```lua
box.execute([[CREATE TABLE k (s1 INT PRIMARY KEY, s2 VARBINARY NOT NULL);]])
box.space.K:create_index('K',{type='bitset'})
box.execute([[INSERT INTO k VALUES (1,X'00');]])
box.execute([[SELECT * FROM k NOT INDEXED WHERE s2 > x'00';]])
```
Result: Error message:
"SQL does not support using non-TREE index type. Please, use INDEXED BY clause to
force using proper index."

This is better than a segmentation fault but I believe Tarantool should
regard NOT INDEXED as a variant of the INDEXED BY clause, else I cannot do the query.
I did not see this error message when I used a hash index.

CuriousGeorgiy:
Closes #5809

### DIFF
--- a/test/unit/bitset_index.c
+++ b/test/unit/bitset_index.c
@@ -45,6 +45,13 @@ void test_size_and_count(void)
 
 	struct tt_bitset_index index;
 	tt_bitset_index_create(&index, realloc);
+	/*
+	 * Checks size of empty, i.e., uninitialized bitset index works
+	 * correctly.
+	 */
+	fail_unless(tt_bitset_index_size(&index) == 0);
+	/* Checks bit number overflow works correctly. */
+	fail_unless(tt_bitset_index_count(&index, UINT64_MAX) == 0);
 
 	enum { P = 10, SIZE = (1 << P) + 1 };
 	for(size_t i = 0; i < SIZE; i++) {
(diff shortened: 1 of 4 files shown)


test/unit/bitset_index.c
```diff
@@ -45,6 +45,13 @@ void test_size_and_count(void)
 
 	struct tt_bitset_index index;
 	tt_bitset_index_create(&index, realloc);
+	/*
+	 * Checks size of empty, i.e., uninitialized bitset index works
+	 * correctly.
+	 */
+	fail_unless(tt_bitset_index_size(&index) == 0);
+	/* Checks bit number overflow works correctly. */
+	fail_unless(tt_bitset_index_count(&index, UINT64_MAX) == 0);
```
locker, line 54: Please also add a Lua test that checks that the reported issue was fixed.

locker (changes requested):

test/unit/bitset_index.c
CuriousGeorgiy: Added a Lua test.

locker (approved):

locker:
Cherry-picked to 2.10.

Issue #8100, opened 2022-12-27
Consider dropping of VDBE


ImeevMA:
We are currently using VDBE to build any SQL query before execution, but this is not necessary. All requests have their own properly described syntax. In this regard, we only need to get the necessary properties of the request using the parser and execute the query.

ImeevMA:
This is actually bad idea.

Issue #3243, opened 2018-03-16
Set iterator type for each index parts


NERON:
Having index: 
`box.space.feeds:create_index('primary',
       { type = 'TREE',unique = true, parts = {1, 'unsigned',2,'unsigned',3,'unsigned'}})`
I'm using pairs for iterating over index, but how can i set different iterator type for each parts: (EQ parts[1], LE parts[2],GE[3])?
In SQL we can set "order by" order for each field, but how we can do it by tarantool without writing special lua functions?

kostja:
This is actually a very good idea, and we need it in SQL as well.

Khatskevich:
Should this iterator walk only over a continuous range of tuples? Or skip unsuitable tuples?

E.g imagine data:
```
1. [1, 2]
2. [1, 3]
3. [2, 1]
4. [2, 2]
5. [2, 3]
```
What should the iterators return?:
1. [GE=1, EQ=2] ([[1,2]] or [[1,2][2,2]]?)
2. [GE=1, LE=1] ([[2,1]] or [[]]?)
3. [GE=1, LE=2] ([[1,2]] or [[1,2],[2,2]]?)
4. Some other (more complex) cases.

locker:
> Should this iterator walk only over a continuous range of tuples? Or skip unsuitable tuples?
> 
> E.g imagine data:
> 
> ```
> 1. [1, 2]
> 2. [1, 3]
> 3. [2, 1]
> 4. [2, 2]
> 5. [2, 3]
> ```
> What should the iterators return?:
> 
> 1. [GE=1, EQ=2] ([[1,2]] or [[1,2][2,2]]?)

[[1,2][2,2]]

> 2. [GE=1, LE=1] ([[2,1]] or [[]]?)

[[2,1]]

> 3. [GE=1, LE=2] ([[1,2]] or [[1,2],[2,2]]?)

[[1,2],[2,1],[2,2]]

In other words, a multi-directional iterator should return all tuples matching the given criteria, even if they don't constitute a singe range. That would match the behavior of SQL SELECT statement.

PR #8272, opened 2023-02-08
sql: drop ASC/DESC in index creation statements

ImeevMA:
Prior to this patch, the SQL parser supported the sort order in the index definition during index creation. However, all of our indexes are ascending until #3243 is implemented, so the specified sort order was simply ignored. To avoid a drastic change in behavior due to the implementation of multidirectional indexes, we have decided to drop the sort order from the index definition until #3243 is implemented.

Part of #8100

@TarantoolBot document
Title: ASC/DESC during index creation

Now the SQL parser does not support ASC/DESC when creating an index. This used to be supported by the parser, but never worked.

An example of statements that are no longer supported:
```
CREATE TABLE t(i INT PRIMARY KEY ASC);
CREATE TABLE t(i INT PRIMARY KEY, a STRING, UNIQUE(a DESC));
CREATE INDEX i1 ON t(a ASC);
```


Gerold103 (approved):
Thanks for the patch!

changelogs/unreleased/gh-8100-parser-changes.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/sql
+
+* Unsupported sort order is dropped from index creation statements (gh-8100).
```
locker, line 3: Looks like a breaking change. Please check with @Totktonada if we should mark it as breaking in the changelog or add a compat option for it.

ImeevMA: Decided to leave as it is now.

PR #8276, opened 2023-02-08
core: check event loop creation in fiber_init()

sergepetrenko:
The main cord's event loop is initialized by fiber_init(), but for some reason successful initialization is only checked in main() after other initialization code might try to use the event loop already.

For example, some of the loop users are coio_enable(), signal_init(), tarantooL_lua_init(), and they are all run before we actually check that loop is not NULL.

Closes https://github.com/tarantool/security/issues/28

### DIFF
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -2027,6 +2027,8 @@ fiber_init(int (*invoke)(fiber_func f, va_list ap))
 	fiber_invoke = invoke;
 	main_thread_id = pthread_self();
 	main_cord.loop = ev_default_loop(EVFLAG_AUTO | EVFLAG_ALLOCFD);
+	if (main_cord.loop == NULL)
+		panic("can't init event loop");
 	cord_create(&main_cord, "main");
 	fiber_signal_init();
 }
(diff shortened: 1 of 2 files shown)


src/lib/core/fiber.c
```diff
@@ -2027,6 +2027,8 @@ fiber_init(int (*invoke)(fiber_func f, va_list ap))
 	fiber_invoke = invoke;
 	main_thread_id = pthread_self();
 	main_cord.loop = ev_default_loop(EVFLAG_AUTO | EVFLAG_ALLOCFD);
+	if (!main_cord.loop)
```
Gerold103, line 2030: Aren't we using explicit `== NULL`?

Gerold103 (changes requested):
Thanks for the patch!

src/lib/core/fiber.c
locker, line 2030: Missing link to the ticket.

sergepetrenko: Yep, we are. My bad. Fixed.

locker (approved):

Gerold103 (approved):
Thanks for the fix!

locker:
Cherry-picked to 2.10.

PR #8282, opened 2023-02-09
log: print build target triplet alongside with the version

Gumix:
This is useful for example for the analysis of performance complaints from users, when they claim that one version of Tarantool is slower than another, in fact comparing debug and release builds.

The log format is approved by @Mons and @unera.


Totktonada (approved):

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8283, opened 2023-02-09
compat: introduce sql_seq_scan_default

### LINKED ISSUE #8096: Introduce SEQSCAN to COMPAT
Generally speaking, [SEQSCAN](https://github.com/tarantool/tarantool/issues/7747) is breaking change. However, for now, we have enabled the old default behavior, so no breaking changes have been made. To gradually introduce new behavior, we need to use the COMPAT module.

ImeevMA:
This patch introduces new sql_seq_scan_default compat option.

Closes #8096


src/lua/compat.lua
```diff
@@ -44,6 +44,13 @@ channel close.
 https://github.com/tarantool/tarantool/wiki/compat%3Afiber_channel_close_mode
 ]]
 
+local SQL_SEQ_SCAN_DEFAULT_BRIEF = [[
+Whether seq_seq_scan session setting should be set to true or false during
+initialization or new session creation.
+
+https://github.com/tarantool/tarantool/wiki/compat%3Asql_seq_scan_default
```
Totktonada, line 51: We plan to change it to `https://tarantool.io/compat/<option_name>`. See #8325.

ImeevMA: I think the PR that gets pushed first wins. Another PR should be corrected accordingly :).

Totktonada: Okay :) I'll left this comment open, because it is open question, which PR will win.

src/lua/compat.lua
```diff
@@ -83,6 +90,12 @@ local options = {
         action = tweak_action('fiber_channel_close_mode',
                               'forceful', 'graceful'),
     },
+    sql_seq_scan_default = {
```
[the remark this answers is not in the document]
Totktonada: Don't know. Compat options are anyway for setting at very beginning of an application initialization. The option re JSON slash escaping initially affects only newly created json serializers (it was changed later, but it was not the goal). The fiber channel option has undefined behavior for already created channels, so it should be called before creating any channel. I would consider that it is not quite valid to set a compat option somewhere in a middle of the application lifespan (except for, maybe, experimenting).

test/sql-luatest/gh_8096_seqscan_compat_test.lua
```diff
@@ -0,0 +1,70 @@
+local server = require('luatest.server')
```
Gerold103, line 1: Feature tests shouldn't start with `gh_xxxx` prefix, if I remember correctly.

Gerold103 (changes requested):
Thanks for the patch!

src/box/sql.c
```diff
@@ -76,15 +77,22 @@ sql_expr_delete_cb(struct Expr *expr)
...
 void
 sql_init(void)
 {
 	tuple_format_expr_compile = sql_expr_compile_cb;
 	tuple_format_expr_delete = sql_expr_delete_cb;
 
-	default_flags |= default_sql_flags;
-
-	current_session()->sql_flags |= default_sql_flags;
+	current_session()->sql_flags = sql_init_flags();
```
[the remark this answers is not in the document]
ImeevMA: It is expected that sessions created prior to box.cfg{} will not receive tweaked SQL flags, as this worked previously. The problem, however, is that in the current patch, these sessions will get their SQL flags initialized to something other than 0, which could be a problem. I fixed it, these sessions will now have 0 as SQL flags.

src/box/sql.c
```diff
@@ -56,13 +56,14 @@
...
 #include "mp_util.h"
+#include "tweaks.h"
 
 static sql *db = NULL;
 
 static const char nil_key[] = { 0x90 }; /* Empty MsgPack array. */
 
-static const uint32_t default_sql_flags = SQL_EnableTrigger | SQL_AutoIndex |
-					  SQL_RecTriggers | SQL_SeqScan;
+bool sql_seq_scan_default = true;
```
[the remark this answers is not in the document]
ImeevMA: It can be, fixed.

test/sql-luatest/gh_8096_seqscan_compat_test.lua
ImeevMA: Fixed, renamed the test.

ImeevMA:
@Gerold103 thank you for the review!

src/box/sql.c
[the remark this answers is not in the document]
Gerold103: Well, if that is intentional, then ok.

Gerold103 (approved):
Thanks for the fixes!

changelogs/unreleased/gh-8096-sql-seq-scan-default.md
```diff
@@ -0,0 +1,3 @@
+## feature/compat
+
+* Compat option sql_seq_scan_default was introduced (gh-8096).
```
locker, line 3: The SEQSCAN feature was never released so we shouldn't add a separate changelog entry. Also, there's no such section 'feature/compat'. You should update the original changelog entry for SEQSACAN instead. PTAL at how other compat options are described in changelogs.

Also, the doc bot request isn't required, since you've already documented it [here](https://github.com/tarantool/tarantool/wiki/compat%3Asql_seq_scan_default).

src/box/sql.c
```diff
@@ -76,15 +79,24 @@ sql_expr_delete_cb(struct Expr *expr)
 	sql_expr_delete(expr);
 }
 
+uint32_t
+sql_flags(void)
+{
+	if (default_sql_flags == 0)
+		return 0;
```
locker, line 86: What is this check for? Why do you need this variable, anyway? Can't you use `SQL_DEFAULT_FLAGS` here instead?

Also, this function had better be called `sql_default_flags()`, because it returns the *default* value of session flags, not just *all* flags.

src/box/sql/tokenize.c
```diff
@@ -624,7 +624,7 @@ struct sql_trigger *
 sql_trigger_compile(const char *sql)
 {
 	struct Parse parser;
-	sql_parser_create(&parser, default_flags);
+	sql_parser_create(&parser, SQL_DEFAULT_FLAGS);
```
locker, line 627: Why don't you use the `sql_flags()` function here and in other places above? I'd expect you use the same default flags everywhere. If `sql_flags()` are solely for session, the function should be renamed appropriately, e.g. to `sql_default_session_flags()` or something like this.

locker (changes requested):

src/lua/compat.lua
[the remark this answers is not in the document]
Totktonada: @Gerold103, @locker, if you're comfortable with current option name, I'll resolve this discussion. Ok?

changelogs/unreleased/gh-8096-sql-seq-scan-default.md
locker: Reminder for this particular option? What about other compat options? I think it's better to create a separate ticket for this task manually.

ImeevMA: Removed doc-request and added `NO_DOC`.

src/lua/compat.lua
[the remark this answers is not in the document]
locker: Ok

src/box/sql.c
ImeevMA: Prior to this patch, all sessions created before `box.cfg{}` had `sql_flags` == 0. Changing them to something else would be a behavior change. Generally speaking, I should add another boolean variable to indicate that `default_sql_flags` has been initialized, but I think we expect default initialized flags to not be 0.

src/box/sql/tokenize.c
ImeevMA: We decided that SEQSCAN changes the behavior of SELECT and should not affect other statements, so there is no need to call `sql_flags()`. Renamed the function to `sql_default_session_flags()`.

src/box/sql.c
locker: > Prior to this patch, all sessions created before `box.cfg{}` had `sql_flags` == 0.

Why? Looks like a bug... Only local sessions can be created before `box.cfg` so I think it'd be fine to fix this along the way.

```
tarantool> box.session.settings
---
- sql_full_metadata: false
  sql_reverse_unordered_selects: false
  sql_full_column_names: false
  sql_vdbe_debug: false
  sql_select_debug: false
  sql_parser_debug: false
  sql_default_engine: memtx
  sql_seq_scan: false
  sql_recursive_triggers: false
...

tarantool> box.cfg{log_level = 'warn'}
---
...

tarantool> box.session.settings
---
- sql_full_metadata: false
  sql_reverse_unordered_selects: false
  sql_full_column_names: false
  sql_vdbe_debug: false
  sql_select_debug: false
  sql_parser_debug: false
  sql_default_engine: memtx
  sql_seq_scan: true
  sql_recursive_triggers: true
...
```

locker: SQL can't be used before `box.cfg` so I don't think we should preserve this behavior.

src/box/sql/tokenize.c
locker: What "other statements"?

src/box/sql.c
ImeevMA: Fixed.

src/box/sql/tokenize.c
ImeevMA: Statement to create `SQL_EXPR` function, statement to create trigger and statement to create view.

changelogs/unreleased/gh-8096-sql-seq-scan-default.md
ImeevMA: As for `feature/compat` - I saw commit headers that started with `compat:`, so I thought `compat` was the new section.

src/lua/compat.lua
Totktonada: I pushed #8325. Please, adjust the link.

Totktonada: Seems resolved. @ImeevMA, thanks!

changelogs/unreleased/gh-7747-seqscan.md
```diff
@@ -3,3 +3,4 @@
 * Introduced a new `SEQSCAN` keyword for usage in scanning `SELECT` queries.
   A new session setting `sql_seq_scan` can be used to allow or restrict scanning
   queries without `SEQSCAN` (gh-7747).
+* Compat option sql_seq_scan_default was introduced (gh-8096).
```
locker, line 6: There's no need to add a new entry for this change or mention the ticket. Let's change to this:

```
* Introduced the new keyword `SEQSCAN` for SQL `SELECT` queries. You may now
  use a scanning SQL `SELECT` query without the `SEQSCAN` keyword only if the
  `sql_seq_scan` session setting is set to true. The default session setting
  value is is controlled by the new `compat` option `sql_seq_scan_default`
  (gh-7747).
```

src/box/sql/sqlInt.h
```diff
@@ -1014,6 +1014,8 @@ struct sql {
 					 */
 enum {
 	SQL_SeqScan = 0x00000008,
+	SQL_DEFAULT_FLAGS = SQL_EnableTrigger | SQL_AutoIndex |
+			    SQL_RecTriggers | SQL_SeqScan,
```
locker, line 1018: Nit: Why don't you use a macro for these constants, like for other SQL flags? It'd look more consistent.

locker (changes requested):

src/box/sql/sqlInt.h
ImeevMA: I'm planning to remake the other flags into `enum` instead of a macro. Not now, though.

locker: Okay. TBH I dislike using enum for bitwise flags. Is there a good reason for that?

ImeevMA: Nothing special, I think. Just prefer enum over macro in such cases.

locker (approved):

PR #8285, opened 2023-02-09
icu: fix potential out-of-bounds access

psergee:
Added bounds check after conversion of a string key to int to avoid potential out-of-bounds access.

Closes tarantool/security#45


locker (approved):

locker:
Cherry-picked to 2.10.

PR #8286, opened 2023-02-09
static-build: fix potential NULL deref in openssl

0x501D:
tls_construct_ctos_session_ticket() has a potential NULL pointer dereference.

Closes tarantool/security#54

NO_DOC=security
NO_TEST=security
NO_CHANGELOG=security

### DIFF
--- a/static-build/cmake/AddDependencyProjects.cmake
+++ b/static-build/cmake/AddDependencyProjects.cmake
@@ -74,8 +74,10 @@ ExternalProject_Add(openssl
         --libdir=lib
         no-shared
     INSTALL_COMMAND ${CMAKE_MAKE_PROGRAM} install_sw
-    PATCH_COMMAND patch -d <SOURCE_DIR> -p1 <
+    PATCH_COMMAND cat
         "${PATCHES_DIR}/openssl-111q-gh-18720.patch"
+        "${PATCHES_DIR}/openssl-tarantool-security-54.patch" |
+        patch -d <SOURCE_DIR> -p1
 )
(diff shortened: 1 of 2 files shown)


static-build/openssl-tarantool-security-54.patch
```diff
@@ -0,0 +1,10 @@
+--- openssl.old/ssl/statem/extensions_clnt.c	2023-02-09 16:18:41.231053686 +0300
++++ openssl/ssl/statem/extensions_clnt.c	2023-02-09 17:33:09.257282077 +0300
+@@ -266,6 +266,7 @@
+         return EXT_RETURN_NOT_SENT;
+ 
+     if (!WPACKET_put_bytes_u16(pkt, TLSEXT_TYPE_session_ticket)
++            || !s->session 
+             || !WPACKET_sub_memcpy_u16(pkt, s->session->ext.tick, ticklen)) {
```
locker, line 8: I think that we should just skip the call to `WPACKET_sub_memcpy_u16` if `s->session == NULL`, not fail the function.

locker (changes requested):
Please rebase on the latest master.

static-build/openssl-tarantool-security-54.patch
locker: Agreed to leave this is as is, because we shouldn't hit this code path anyway and checking `s->session` is needed solely to suppress the SVACE warning so it doesn't really matter how we fix it.

0x501D: rebased

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8290, opened 2023-02-10
mpstream: fix NULL dereference in `mpstream_encode_double`

CuriousGeorgiy:
Closes tarantool/security#63

### DIFF
--- a/src/lib/mpstream/mpstream.c
+++ b/src/lib/mpstream/mpstream.c
@@ -1,37 +1,17 @@
 /*
- * Copyright 2010-2018, Tarantool AUTHORS, please see AUTHORS file.
+ * SPDX-License-Identifier: BSD-2-Clause
  *
- * Redistribution and use in source and binary forms, with or
- * without modification, are permitted provided that the following
- * conditions are met:
- *
- * 1. Redistributions of source code must retain the above
- *    copyright notice, this list of conditions and the
- *    following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above
- *    copyright notice, this list of conditions and the following
- *    disclaimer in the documentation and/or other materials
- *    provided with the distribution.
- *
- * THIS SOFTWARE IS PROVIDED BY <COPYRIGHT HOLDER> ``AS IS'' AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
- * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
- * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
- * <COPYRIGHT HOLDER> OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
- * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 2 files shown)


test/unit/security_gh_63_mpstream_encode_double_null_dereference.c
```diff
@@ -0,0 +1,61 @@
...
+test_encode_double_oom(void)
+{
+	plan(1);
+	header();
+
+	struct mpstream mps;
+	bool err;
+	mpstream_init(&mps, NULL, reserve_mock, alloc_mock, err_mock, &err);
+	err = false;
+	mpstream_encode_double(&mps, 0);
```
locker, line 41: The test looks incomplete to me (why does it test only `mpstream_encode_double`?). Please either test OOM conditions for all mpstream functions or drop the test. If you decide to keep the test, please name it simply `test/uni/mpstream.c` so that we can add more test cases to it later.

test/unit/CMakeLists.txt
```diff
@@ -600,3 +600,8 @@ create_unit_test(PREFIX lua_tweaks
                            ${LIBYAML_LIBRARIES}
                            ${READLINE_LIBRARIES}
 )
+
+create_unit_test(PREFIX security_gh_63_mpstream_encode_double_null_dereference
+                 SOURCES security_gh_63_mpstream_encode_double_null_dereference.c core_test_utils.c
```
locker, line 605: Nit: the line is longer than 80 characters.

locker (changes requested):
Please rebase on the latest master (for the `static-build-ci` label).

test/unit/CMakeLists.txt
CuriousGeorgiy: We do not have a line length limit for CMake files in our code style. And we don't have one because CMake files tend to have long lines.

test/unit/security_gh_63_mpstream_encode_double_null_dereference.c
CuriousGeorgiy: > The test looks incomplete to me, because it doesn't check the same condition in other mpstream methods.

This should've been tested in the patch that introduced MsgPack streams in the first place, I'm just adding a reproducer for a specific bug.

Okay, dropping the test, at your risk.

CuriousGeorgiy: /cc @alyapunov

locker: image

How very nice of you... I asked you to **either fix the test or drop it**, but you wrote that I "insisted on dropping" it. You could just as well say "author refused to fix the test and decided to drop it".

Anyway, neither sounds like a valid justification for absence of a test so I changed it to "unlikely to happen because malloc shouldn't normally fail, and we don't test other mpstream methods for OOM either".

locker:
Cherry-picked to 2.10.

test/unit/security_gh_63_mpstream_encode_double_null_dereference.c
CuriousGeorgiy: > Either fix the test

You ignored my arguments:
> This test checks a specific bug condition. Do we test all possible conditions when we write tests for bug tickets?

> This should've been tested in the patch that introduced MsgPack streams in the first place, I'm just adding a reproducer for a specific bug.

We have a certain a workflow and I believe this discussion is wayward. The SOP document says:
> For a bug fix, please create a separate file with an autonomous test case in a > corresponding test suite. Name should follow this format:
<suite name>/gh-XXX-two-three-word-title
>
> Try to make a test case as small as possible. A test is written once and is run, read and maintained many times by many people: keep this in mind and avoid adding obscure, undocumented, and unjustifiably long tests.

PR #8292, opened 2023-02-10
static-build: move all patches to patches sub-directory

locker:
We're going to add a whole bunch of them. Putting them all in a sub-directory will help keeping the file tree organized.

Note, we have to update .gitignore so that the patches/ sub-directory is ignored only at the top level (it's used by quilt).

Related EE PR: https://github.com/tarantool/tarantool-ee/pull/371


nshy (approved):

locker:
Cherry-picked to 2.10.

PR #8293, opened 2023-02-10
ci: run static-build workflows on 'static-build-ci' label

locker:
Sometimes, we only need to test static-build, e.g. when we apply a patch to a third party sub-project. Let's introduce a new label to run the ci checks faster in this case.


ylobankov (approved):
LGTM

locker:
Cherry-picked to 2.10.

Issue #7223, opened 2022-05-31
Invalid key part count error is not descriptive enough


Totktonada:
```yaml
- error: Invalid key part count in an exact match (expected 1, got 0)
```

It is hard to spot what exactly raises the error. There are different ways to make it better:

* Give more details: space name, index name.
* Include a code line number into the message itself.
* Change something about error reporting to not miss details.

Inspired by @Mons. He said that he met it as result of a maintenance script run and was angry.

Mons:
Let's add for now space and index names as a simple solution to not dive into errors refactoring

alyapunov:
Blocked by new error subsystem refactoring.

Issue #15, opened 2013-05-07
tarantool console incorrectly prints some strings


IlyaM:
Reproduceable with 1.4.9.

When inserting tuples with string data that contains numbers from [1000,9999] range afterwards the same data is printed incorrectly in the shell if I select corresponding tuples. As far as I can tell the actual data in the storage is correct (because I can see correct values if I use python client library to retrieve it)

I.e. example from console session; the following selects end up in incorrect data printed

localhost> insert into t0 values ('1', '1000')
Insert OK, 1 rows affected
localhost> select \* from t0 where k0='1'
Select OK, 1 rows affected
[1, 808464433]
localhost> replace t0 values ('1', '2000')
Insert OK, 1 rows affected
localhost> select \* from t0 where k0='1'
Select OK, 1 rows affected
[1, 808464434]
localhost> replace t0 values ('1', '9999')
Insert OK, 1 rows affected
localhost> select \* from t0 where k0='1'
Select OK, 1 rows affected
[1, 960051513]

but these selects look ok

localhost> replace t0 values ('1', '10000')
Insert OK, 1 rows affected
localhost> select \* from t0 where k0='1'
Select OK, 1 rows affected
[1, 10000]
localhost> replace t0 values ('1', '100')
Insert OK, 1 rows affected
localhost> select \* from t0 where k0='1'
Select OK, 1 rows affected
[1, 100]

Issue #14, opened 2013-03-18
Confusing "Could not find 'script_dir' option" error message


IlyaM:
Hello,

I've just downloaded ubuntu package for tarantool as I wanted to test this database. When trying to start tarantool with sample config (I've copied /etc/tarantool/instances.available/example.cfg to /etc/tarantool/instances.enabled/) it fails with confusing error message:

> ilya@denmark:~$ sudo /etc/init.d/tarantool start
> tarantool: Staring instances
> /usr/bin/tarantool_box: check_config FAILED
> - Could not find 'script_dir' option

Interestingly enough the message goes away if I comment out script_dir line from the config so the error text is obviously incorrect: I would expect such error text if script_dir was required and was missing in the config. Here the error is happening because somehow tarantool is not liking the value of script_dir in my config but the error mesage doesn't explain why (i.e. for example creating an empty directory on script_dir path doesn't remove the error).

Issue #13, opened 2013-02-28
Compile on MacOS 10.7.5 failed with undefined IOV_MAX 


svaroqui:
error: 'IOV_MAX' undeclared (first use in this function)
Quick fixe on src/sio.m 

39,41d38
< #if defined(**FreeBSD**) || defined(**APPLE**)
< #define IOV_MAX 1024
< #endif

PR #8296, opened 2023-02-10
box: add detailed invalid-key-part error

Lord-KA:
Add ER_EXACT_MATCH_DETAILED ("Invalid key part count in an exact match") error code, as old version wasn't informative enough in some cases. Now it displays space and index names too.

Refactor exac_key_validate() and exac_key_validate_nullable() to receive wider context to provide names for the error msg. Now they are specialised to primary key and user-defined key accordingly.

NO_DOC=error is not described in doc

### DIFF
--- a/changelogs/unreleased/gh-7223-update-invalid-key-part-error.md
+++ b/changelogs/unreleased/gh-7223-update-invalid-key-part-error.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Updated "invalid key part count" error to include current space and index
+  names when possible (gh-7223).
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -142,13 +142,17 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
 }
 
 int
-exact_key_validate(struct key_def *key_def, const char *key,
+exact_key_validate(struct index_def *index_def, const char *key,
 		   uint32_t part_count)
 {
 	assert(key != NULL || part_count == 0);
+	struct key_def *key_def = index_def->key_def;
 	if (key_def->part_count != part_count) {
-		diag_set(ClientError, ER_EXACT_MATCH, key_def->part_count,
-			 part_count);
+		struct space *space = space_by_id(index_def->space_id);
+		diag_set(ClientError, ER_EXACT_MATCH,
+			 tt_sprintf("in space \"%s\", index \"%s\" ",
+				    space_name(space), index_def->name),
+			 key_def->part_count, part_count);
 		return -1;
 	}
 	const char *key_end;
@@ -156,13 +160,17 @@ exact_key_validate(struct key_def *key_def, const char *key,
 }
 
 int
-exact_key_validate_nullable(struct key_def *key_def, const char *key,
+exact_key_validate_nullable(struct index_def *index_def, const char *key,
 			    uint32_t part_count)
 {
 	assert(key != NULL || part_count == 0);
+	struct key_def *key_def = index_def->cmp_def;
 	if (key_def->part_count != part_count) {
-		diag_set(ClientError, ER_EXACT_MATCH, key_def->part_count,
-			 part_count);
+		struct space *space = space_by_id(index_def->space_id);
+		diag_set(ClientError, ER_EXACT_MATCH,
+			 tt_sprintf("in space \"%s\", index \"%s\" ",
+				    space_name(space), index_def->name),
+			 key_def->part_count, part_count);
 		return -1;
 	}
 	const char *key_end;
@@ -298,7 +306,7 @@ box_index_get(uint32_t space_id, uint32_t index_id, const char *key,
 	}
 	const char *key_array = key;
 	uint32_t part_count = mp_decode_array(&key);
-	if (exact_key_validate(index->def->key_def, key, part_count))
+	if (exact_key_validate(index->def, key, part_count))
 		return -1;
 	box_run_on_select(space, index, ITER_EQ, key_array);
 	/* Start transaction in the engine. */
@@ -473,7 +481,7 @@ box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
 					     &pos, &pos_end) != 0)
 			return NULL;
 		uint32_t pos_part_count = mp_decode_array(&pos);
-		if (exact_key_validate_nullable(index->def->cmp_def, pos,
+		if (exact_key_validate_nullable(index->def, pos,
 						pos_part_count) != 0)
 			return NULL;
 	} else {
--- a/src/box/index.h
+++ b/src/box/index.h
@@ -460,22 +460,27 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
 
 /**
  * Check that the supplied key is valid for a search in a unique
- * index (i.e. the key must be fully specified).
+ * index, i.e. the key must be fully specified and correspond to
+ * index_def->key_def.
  * @retval 0  The key is valid.
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate(struct key_def *key_def, const char *key,
+exact_key_validate(struct index_def *index_def, const char *key,
 		   uint32_t part_count);
 
 /**
  * Check that the supplied key is valid for representing iterator
- * position (i.e. the key must be fully specified, but nulls are allowed).
+ * position, i.e. the key must be fully specified and correspond to
+ * index_def->cmp_def, where nulls are allowed.
+ * Note that since this function is designed to check a key for
+ * precise iterator position, it relies on index_def->cmp_def,
+ * in contrast to `key_validate` which relies on index_def->key_def.
  * @retval 0  The key is valid.
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate_nullable(struct key_def *key_def, const char *key,
+exact_key_validate_nullable(struct index_def *index_def, const char *key,
 			    uint32_t part_count);
 
 /**
--- a/src/box/session_settings.c
+++ b/src/box/session_settings.c
@@ -343,7 +343,10 @@ session_settings_space_execute_update(struct space *space, struct txn *txn,
 	const char *new_key, *key = request->key;
 	uint32_t new_size, new_key_len, key_len = mp_decode_array(&key);
 	if (key_len == 0) {
-		diag_set(ClientError, ER_EXACT_MATCH, 1, 0);
+		diag_set(ClientError, ER_EXACT_MATCH,
+			 tt_sprintf("in space \"%s\", index \"%s\" ",
+				    space_name(space), pk_def->name),
+			 1, 0);
 		return -1;
 	}
 	if (key_len > 1 || mp_typeof(*key) != MP_STR) {
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3146,7 +3146,7 @@ box_select(uint32_t space_id, uint32_t index_id,
 					     &pos, &pos_end) != 0)
 			return -1;
 		uint32_t pos_part_count = mp_decode_array(&pos);
-		if (exact_key_validate_nullable(index->def->cmp_def, pos,
+		if (exact_key_validate_nullable(index->def, pos,
 						pos_part_count) != 0)
 			return -1;
 	} else {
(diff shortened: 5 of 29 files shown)


alyapunov:
As I understood you have fixed only a small part of the problem:
```
tarantool> box.space._space:get({1, 1})
---
- error: Invalid key part count in an exact match in space _space, index primary (expected
    1, got 2)
...

tarantool> box.space._space:select({1, 1})
---
- error: Invalid key part count (expected [0..1], got 2)
...

tarantool> box.space._space:select({'a'})
---
- error: 'Supplied key type of part 0 does not match index part type: expected unsigned'
...
```
I fear this patch introduces some chaos: in some places there will be details, in some - not..

alyapunov:
I also think that we should fix all places with ER_EXACT_MATCH.

changelogs/unreleased/gh-7223-update-invalid-key-part-error.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Added detailed version of "invalid key part count" error with current
+  space and index names (gh-7332).
```
alyapunov, line 4: It must be 7223

alyapunov:
We discussed offline, it would be much better if we use stack diagnostics.

Mons:
> As I understood you have fixed only a small part of the problem:

It is better to fix at least part of problems, than not to fix them at all and leave it in LTS release as is for several years.

alyapunov:
`exac_key_validate` -> `exact_key_validate` in the commit message (twice).

alyapunov:
> > As I understood you have fixed only a small part of the problem:
> 
> It is better to fix at least part of problems, than not to fix them at all and leave it in LTS release as is for several years.

It must be fixed a least for `select` too, not for `get` alone.

src/box/index.h
```diff
@@ -459,23 +459,23 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
 	     const char *key, uint32_t part_count);
 
 /**
- * Check that the supplied key is valid for a search in a unique
- * index (i.e. the key must be fully specified).
+ * Check that the supplied primary index key is valid for a search in
```
alyapunov, line 462: That is not necessary a primary index, so you should not change the comment.

src/box/index.h
```diff
@@ -459,23 +459,23 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
...
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate(struct key_def *key_def, const char *key,
+exact_key_validate(struct index_def *index_def, const char *key,
 		   uint32_t part_count);
 
 /**
- * Check that the supplied key is valid for representing iterator
+ * Check that the supplied user-defined key is valid for representing iterator
```
alyapunov, line 472: Can't catch why you changed the comment here. I'm not sure that the function is used only for user-defined key, perhaps it is or will be used for developer-defined keys.

src/box/session_settings.c
```diff
@@ -343,7 +343,8 @@ session_settings_space_execute_update(struct space *space, struct txn *txn,
 	const char *new_key, *key = request->key;
 	uint32_t new_size, new_key_len, key_len = mp_decode_array(&key);
 	if (key_len == 0) {
-		diag_set(ClientError, ER_EXACT_MATCH, 1, 0);
+		diag_set(ClientError, ER_EXACT_MATCH_DETAILED,
+			 space_name(space), pk_def->name, 2, 0);
```
alyapunov, line 347: Why did you change `1` to `2`? There must be only one part, not two. It's a pity that we don't have a test for it. Perhaps we should add one.

alyapunov (changes requested):
Thanks for the patch, see several comments below.

Generally I don't like that we fix only ER_EXACT_MATCH error, leaving lots of similar error aside. But I can agree that it is better that nothing.

But the thing that bothers me most is that we create another error ER_EXACT_MATCH_DETAILED along with the original ER_EXACT_MATCH. There's no logical difference between them, only message. For a stupid example, what will we do later if we need the same error, but without known instance of a space, just with index? Create another error ER_EXACT_MATCH_LESS_DETAILED?
I think we should add an extra optional token (%s) to the existing error and fill it with space and index when they are known, leaving as empty string when they are not known. That is not perfect but more flexible solution.

alyapunov:
Note also that when we finally use stacked errors, and these changes in `exact_key_validate` and `exact_key_validate_nullable` will be rolled back.

src/box/index.h
Lord-KA: As we've discussed privately, I changed `exact_key_validate_nullable()` to check `index_def->cmp_def` that is according to description a user defined key.

```C
         /**
           * User-defined key definition, merged with the primary
           * key parts. Used by non-unique keys to uniquely identify
           * iterator position.
           */
          struct key_def *cmp_def;
```

src/box/session_settings.c
Lord-KA: Fixed.

src/box/index.h
Lord-KA: Ok, we've discussed it once more and I reworded the comments.

```diff
 /**
- * Check that the supplied key is valid for a search in a unique
- * index (i.e. the key must be fully specified).
+ * Check that the index_def->key_def is valid for a search in
+ * a unique index (i.e. the key must be fully specified).
  * @retval 0  The key is valid.
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate(struct key_def *key_def, const char *key,
+exact_key_validate(struct index_def *index_def, const char *key,
                   uint32_t part_count);
 
 /**
- * Check that the supplied key is valid for representing iterator
+ * Check that the index_def->cmp_def is valid for representing iterator
  * position (i.e. the key must be fully specified, but nulls are allowed).
  * @retval 0  The key is valid.
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate_nullable(struct key_def *key_def, const char *key,
+exact_key_validate_nullable(struct index_def *index_def, const char *key,
```

src/box/index.h
Lord-KA: Fixed. See diff below.

changelogs/unreleased/gh-7223-update-invalid-key-part-error.md
Lord-KA: Fixed:

```md
## feature/box

* Updated "invalid key part count" error to include current space and index
  names when possible (gh-7223).
```

Lord-KA:
Removed `DETAILED` version.

src/box/index.h
```diff
@@ -459,23 +459,23 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
 	     const char *key, uint32_t part_count);
 
 /**
- * Check that the supplied key is valid for a search in a unique
- * index (i.e. the key must be fully specified).
+ * Check that the index_def->key_def is valid for a search in
```
alyapunov, line 462: That's wrong again.
The supplied key - is `key` argument.
I think you should write something like:
```
 * Check that the supplied key is valid for a search in a unique
 * index, i.e. the key must be fully specified and correspond to
 * index_def->key_def.
```

src/box/index.h
```diff
@@ -459,23 +459,23 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
...
  * @retval -1 The key is invalid.
  */
 int
-exact_key_validate(struct key_def *key_def, const char *key,
+exact_key_validate(struct index_def *index_def, const char *key,
 		   uint32_t part_count);
 
 /**
- * Check that the supplied key is valid for representing iterator
+ * Check that the index_def->cmp_def is valid for representing iterator
```
alyapunov, line 472: That's wrong again.
The supplied key - is key argument.
I think you should write something like:
```
* Check that the supplied key is valid for representing iterator
* position, i.e. the key must be fully specified and correspond to
* index_def->cmp_def, with nulls are allowed.
* Note that since this function is designed checking a key for
* precise iterator position, it relies on index_def->cmp_def,
* in contrast to `key_validate` which relies on index_def->key_def.
* 
```

alyapunov (changes requested):

src/box/index.h
Lord-KA: Done.

src/box/index.h
Lord-KA: Done.

alyapunov (approved):

src/box/index.cc
```diff
@@ -142,27 +142,35 @@ key_validate(const struct index_def *index_def, enum iterator_type type,
...
 	assert(key != NULL || part_count == 0);
+	struct key_def *key_def = index_def->key_def;
 	if (key_def->part_count != part_count) {
-		diag_set(ClientError, ER_EXACT_MATCH, key_def->part_count,
-			 part_count);
+		struct space *space = space_by_id(index_def->space_id);
+		diag_set(ClientError, ER_EXACT_MATCH,
+			 tt_sprintf("in space \"%s\", index \"%s\" ",
+				    space_name(space), index_def->name),
+			 key_def->part_count, part_count);
```
locker, line 155: 1. Updated the submodule in [Tarantool EE](https://github.com/tarantool/tarantool-ee) to this PR.

2. Applied the following patch to make the project compile.

   <details>

   ```diff
      diff --git a/src/box/index_read_view.c b/src/box/index_read_view.c
   index 7c6c562fef9a..d4dfa56b29f4 100644
   --- a/src/box/index_read_view.c
   +++ b/src/box/index_read_view.c
   @@ -190,7 +190,7 @@ box_index_read_view_get(struct index_read_view *rv, const char *key,
    		return -1;
    	}
    	uint32_t part_count = mp_decode_array(&key);
   -	if (exact_key_validate(rv->def->key_def, key, part_count) != 0)
   +	if (exact_key_validate(rv->def, key, part_count) != 0)
    		return -1;
    	if (index_read_view_get(rv, key, part_count, result) != 0)
    		return -1;
   diff --git a/src/box/raw_read_view.c b/src/box/raw_read_view.c
   index 4e83bbd8850f..a0038bf464a4 100644
   --- a/src/box/raw_read_view.c
   +++ b/src/box/raw_read_view.c
   @@ -318,7 +318,7 @@ box_raw_read_view_get(const box_raw_read_view_index_t *index_,
    		return -1;
    	}
    	uint32_t part_count = mp_decode_array(&key);
   -	if (exact_key_validate(index->def->key_def, key, part_count) != 0)
   +	if (exact_key_validate(index->def, key, part_count) != 0)
    		return -1;
    	return index_read_view_get_raw(index, key, part_count, data, size);
    }
   ```

   </details>

3. Ran the following script.

   <details>

   ```Lua
   box.cfg({log_level = 'warn'})
   box.schema.create_space('test')
   rv = box.read_view.open()
   box.space.test:drop()
   rv.space.test:get({1, 2})
   ```

   </details>

4. Got a segmentation fault.

   <details>

   ```
   Segmentation fault
     code: SEGV_MAPERR
     addr: 0xc0
     context: 0x7facfedff540
     siginfo: 0x7facfedff670
     rax      0x0                0
     rbx      0x55ff7b7240a0     94555776106656
     rcx      0x14               20
     rdx      0x100000           1048576
     rsi      0x100000           1048576
     rdi      0x0                0
     rsp      0x7facfedffc00     140380987194368
     rbp      0x7facfedffc00     140380987194368
     r8       0x7facfedfffd0     140380987195344
     r9       0x0                0
     r10      0x7facff03b000     140380989534208
     r11      0x10               16
     r12      0x1                1
     r13      0x0                0
     r14      0x40dfc088         1088405640
     r15      0x55ff7a7bd4b0     94555759957168
     rip      0x55ff7a57bd07     94555757591815
     gs       0x0                0
     fs       0x0                0
     cr2      0xc0               192
     err      0x4                4
     oldmask  0x0                0
   Current time: 1676469433
   this may fail as well
   #1  0x55ff7a7f42d6 in crash_collect+256
   #2  0x55ff7a7f4ce8 in crash_signal_cb+100
   #3  0x7fad016da420 in __restore_rt+0
   #4  0x55ff7a57bd07 in space_name+12
   #5  0x55ff7a57cdc5 in exact_key_validate+257
   #6  0x55ff7a7b8365 in box_index_read_view_get+588
   #7  0x55ff7a7bd591 in lbox_index_read_view_get+225
   #8  0x55ff7a8589f7 in lj_BC_FUNCC+70
   #9  0x55ff7a863875 in lua_pcall+849
   #10 0x55ff7a7d2f78 in luaT_call+45
   #11 0x55ff7a7c8c58 in lua_main+189
   #12 0x55ff7a7c9200 in run_script_f+1423
   #13 0x55ff7a565cca in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*)+34
   #14 0x55ff7a80016d in fiber_loop+210
   #15 0x55ff7aaca17d in coro_init+120
   Aborted
   ```

   </details>

locker (changes requested):

Lord-KA:
It was decided to all error enhancements in one go via stack diagnostics.

PR #8300, opened 2023-02-13
lua: allow to override a built-in module

### LINKED ISSUE #7774: Builtin modules overloading

Totktonada:
This patchset continues the series (#8173, #8250) of commits about overriding built-in modules (#7774). It finally solves the main task. The changes:

* Use `luaT_setmodule()` to register serializers.
* Use another table instead of `package.loaded` to store built-in modules (and add a loader that looks into this table).
* Set `_G.arg` earlier to make it accessible from override modules.
* Added `gentree` and `justrun` testing helpers (see commits for details).
* Added `override.<module_name>` loader.
* Added an ability to enable/disable the override loader.

Several things are technically out of scope of the issue, but they should be solved together with it:

* Add a mechanism for discarding assignment to `package.loaded` for a module.
* Add a flag to indicate, whether tarantool's initialization is finished.
* Search for `.rocks` next to the main script file.

I'll send them in separate pull requests.

Fixes #7774


igormunkin (approved):
**[PATCH 1/7] lua: register serializers using luaT_setmodule()**

@Totktonada, thanks for the patch! LGTM as trivial.

test/luajit-test-init.lua
```diff
@@ -1,9 +1,12 @@
 -- Disable strict for Tarantool.
 require("strict").off()
 
+local loaders = require('internal.loaders')
+
 -- XXX: lua-Harness test suite uses it's own tap.lua module
 -- that conflicts with the Tarantool's one.
 package.loaded.tap = nil
```
igormunkin, line 8: Minor: I believe, that as a result of the corresponding patches removing builtin modules from `package.loaded`, this line can be removed, right?

src/lua/init.c
```diff
@@ -839,6 +839,13 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
 	}
 	luaL_openlibs(L);
 
+	/*
+	 * Create a table for storing loaded built-in modules.
+	 * Similar to _LOADED (package.loaded).
+	 */
+	lua_createtable(L, 0, 128);
```
igormunkin, line 858: Minor: Why 128? I would rather use "`sizeof(lua_modules)`" for this.

Buristan (commented):
Hi, Alexander!
Thanks for the patch-set!
I'll review it per patch.

**[PATCH 1/7] lua: register serializers using luaT_setmodule()**
LGTM!

src/lua/minifio.c
```diff
@@ -53,6 +53,12 @@ tarantool_lua_minifio_init(struct lua_State *L)
 		{NULL, NULL}
 	};
 
+	/*
+	 * The loaders.builtin loader is not in effect yet.
+	 * Set the minifio module into package.loaded manually.
+	 */
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
```
igormunkin, line 60: Minor: why do we need use `_LOADED` here instead of `_PRELOAD`?

test/unit/luaT_tuple_new.c
```diff
@@ -187,6 +187,10 @@ main()
 	struct lua_State *L = luaL_newstate();
 	luaL_openlibs(L);
 
+	/* luaT_newmodule() assumes that this table exists. */
```
igormunkin, line 190: Side note: I want to share here the approach used in [uJIT tests](https://github.com/luavela/luavela/blob/master/tests/impl/uJIT-tests-C/suite/test_common_lua.h#L27-L35), but I don't ask to fix it within this patchset. If this works for you, feel free to file an issue to move all unit tests for this machinery.

test/unit/lua_tweaks.c
```diff
@@ -241,7 +241,30 @@ main(void)
...
+	 */
+	lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
+	lua_getfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
+	lua_getfield(L, -1, "internal.tweaks");
+	lua_setfield(L, -3, "internal.tweaks");
+	lua_pop(L, 2);
+
 	tarantool_lua_error_init(L);
 	fail_unless(luaT_dostring(
 			L, "tweaks = require('internal.tweaks')") == 0);
```
igormunkin, line 270: I just wonder, whether this line doesn't solve the problem described above?

src/lua/minifio.c
```diff
@@ -53,6 +53,12 @@ tarantool_lua_minifio_init(struct lua_State *L)
...
 	luaT_newmodule(L, "internal.minifio", minifio_methods);
+	lua_setfield(L, -2, "internal.minifio");
 	lua_pop(L, 1);
```
igormunkin, line 63: `lua_pop` looks excess here (`lua_setfield` removes the module field from the stack).

igormunkin (changes requested):
**[PATCH 2/7] lua: don't set built-in modules to package.loaded**

@Totktonada, thanks for the patch! Please consider my comments and questions below.

src/lua/minifio.c
Totktonada: It pops `_LOADED`.

igormunkin (approved):
**[PATCH 3/7] lua: define _G.arg earlier**

@Totktonada, thanks for the patch! Looks legit, LGTM.

src/lua/minifio.c
Totktonada: I have several answers here.

1. Why should we use `_PRELOAD` if we need only `_LOADED`? If we would use `_PRELOAD` here, the result would be setting the value into `_LOADED` in a bit less direct way.
2. If there would be the builtin loader at this stage, `require()` would find the module and store it in `_LOADED`. We imitate this behavior.
3. `_PRELOAD` is accessible by a user and I prefer to don't touch it in the platform if possible.

src/lua/minifio.c
igormunkin: PEBKAC. Anyway, comment desired, if you don't mind.

src/lua/init.c
Totktonada: `lua_modules` is mostly irrelevant here. There are modules written using C. There are box's modules.

I did some calculations and got a number around 90 or 100 for this table. So 128 looks good to avoid rehashing and to don't have reasons to touch this for a long time. I can set it to zero if it is important. I definitelly will not write anything more complex than a constant here.

test/luajit-test-init.lua
Totktonada: Technically, I guess, yes. However it would be fragile.

If there is no `require('tap')` call during tarantool's initialization, `package.loaded.tap` is `nil`. We only need `loaders.builtin.tap = nil` to choose an external module at next `require('tap')`.

However if some built-in module will depend on `tap` (and so call `require('tap')`), the built-in `tap` module will be written to `package.loaded.tap` and future require() will return it.

So the old `package.loaded.foo = nil` hack is now `package.loaded.foo = nil` plus `loaders.builtin.foo = nil`.

igormunkin (approved):
**[PATCH 4/7] test: add a generator for a tree of Lua files**

src/lua/minifio.c
igormunkin: 1. `package.loaded[module] = nil` is a standard idiom for module "unload"; in general case `package.preload` allows to reload the module if this is needed.
2. Yeah, but see the reason above.
3. Don't get this point. `package.loaded` is accessible too, isn't it?

src/lua/init.c
igormunkin: OK, I get it now, thanks. Anyway, I would like to hear @Buristan opinion here. The most bothering thing here is the raw constant, but I can try to live with it :)

igormunkin (approved):
**[PATCH 5/7] test: add a simple tarantool runner/output catcher**

Buristan (commented):
**[PATCH 2/7] lua: don't set built-in modules to package.loaded**
Generally LGTM, just few minor suggestions below.


> If `foo` is already assigned to `package.loaded`, the loaders will not
called at all, so we can't check the file system.

Typo: s/will not called/will not be called/

>   built-in tap module. However now it is not enough to remove it from

Typo: s/However/However,/

Buristan (commented):
**[PATCH 3/7] lua: define _G.arg earlier**
> However nothing prevents us from providing the `arg` global variable at

Typo: s/However/However,/

test/treegen.lua
```diff
@@ -0,0 +1,129 @@
...
+    log.info(('Writing a script: %s'):format(script_abspath))
+    fh:write(template:gsub('<(.-)>', {
+        script = script,
+    }))
+
+    fh:close()
+end
+
+
+function treegen.before_all_default(g)
```
Buristan, line 64: `before_all_default()`/`after_all_defalt()` is slightly misleading to me.
Maybe `before_all_init()`/`after_all_clean()` is better?
Feel free to ignore.

Buristan (commented):
**[PATCH 4/7] test: add a generator for a tree of Lua files**

LGTM, except a few insignificant nits below.

> based test. Also it gain ability to provide several templates for Lua

Typo: s/Also it gain/Also, it gains/

Buristan (commented):
**[PATCH 5/7] test: add a simple tarantool runner/output catcher**

> However there is nothing like 'just run tarantool with given args and

Typo: s/However/However,/

test/unit/lua_tweaks.c
Totktonada: Agreed to add a bit more info into the comment above.

src/lua/init.c
```diff
@@ -839,6 +839,13 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
...
+	lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
```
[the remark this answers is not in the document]
Totktonada: I think it would be a bit cumbersome to include a header just to use the name. If you don't mind, I'll pass it over.

test/treegen.lua
Totktonada: Agreed to name it `init` and `clean`.

src/lua/minifio.c
Totktonada: We agreed that we can't reach an agreement here, so I don't know. I'll leave it as is.

test/luajit-test-init.lua
```diff
@@ -100,6 +100,45 @@ rawset(_G, 'pairs', pairs_M.builtin_pairs)
...
+--
+-- And the Lua code:
+--
+--  | package.path = 'libs/?.lua;libs/bar.lua'
+--  | local foo = require('foo')
+--
+-- The test case expects that `foo` contains a return value of
+-- libs/foo.lua.
+--
+-- However when the override loader is enabled, the following
```
Buristan, line 125: Typo: s/However/However,/

test/luajit-test-init.lua
```diff
@@ -100,6 +100,45 @@ rawset(_G, 'pairs', pairs_M.builtin_pairs)
...
+-- tries `libs/?.lua` first -- it means `libs/override/foo.lua`
+-- and there is no such file. Then the loader tries `libs/bar.lua`
+-- and succeeds.
+--
+-- So `foo` contains a return value of `libs/bar.lua`.
+--
+-- Unlikely there is a valid user scenario for a package.path
+-- entry without the interrogation mark. We consider this override
+-- loader behavior as valid. The loader is disabled to pass the
+-- test suite, but it is likely would be more correct to discard
```
Buristan, line 136: Typo? s/it is likely would be/likely it would be/

Buristan (commented):
**[PATCH 6/7] lua: allow to override a built-in module**
Generally, LGTM, just minor nits and questions below.

Buristan (approved):
**[PATCH 7/7] lua: enable/disable override loader**
First of all want to say that I really love `loaders` internal functions. They look very nice (`conditional_loader`, `chain_loader` are state of art for me) :).

Ignoring value of `TT_OVERRIDE_BUILTIN` when it can't be parsed looks fragile, but OK for me as far as it is more like for internal usage.

Thanks for the patch-set! Nice work!
LGTM!

test/app-luatest/override_test.lua
```diff
@@ -0,0 +1,146 @@
...
+-- The list is collected from loaders.builtin and package.loaded
+-- keys. Many modules are excluded deliberately:
+--
+-- - json -- it is used in the test itself
+-- - bit, coroutine, debug, ffi, io, jit, jit.*, math, os,
+--   package, string, table -- LuaJIT modules
+-- - misc -- tarantool's module implemented as part of LuaJIT.
+-- - *.lib, internal.*, utils.* and so on -- tarantool internal
+--   modules
+-- - memprof.*, misc.*, sysprof.*, table.*, timezones -- unclear
```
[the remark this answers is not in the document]
Totktonada: Seems so.

```lua
tarantool> do local modules = {} for m in pairs(package.loaded) do if m:startswith('misc.') then table.insert(modules, m) end end table.sort(modules) return modules end
---
- - misc.memprof
  - misc.sysprof
...
```

test/app-luatest/override_test.lua
```diff
@@ -0,0 +1,146 @@
...
+    'datetime',
+    'digest',
+    'error',
+    'fun',
+    'help',
+    'http.client',
+    'iconv',
+    'key_def',
+    'luadebug',
+    'memprof',
```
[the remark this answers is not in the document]
Totktonada: It works, so it is ok to have there. If you're about `memprof.*` exclusion, it is glob, not regex :)

src/lua/init.c
Totktonada: Changed to `lua_newtable(L)`. No magic constants anymore.

test/unit/luaT_tuple_new.c
Totktonada: Something like this?

```diff
diff --git a/test/unit/luaT_tuple_new.c b/test/unit/luaT_tuple_new.c
index 0b3847347..42643464c 100644
--- a/test/unit/luaT_tuple_new.c
+++ b/test/unit/luaT_tuple_new.c
@@ -14,6 +14,7 @@
 
 #define UNIT_TAP_COMPATIBLE 1
 #include "unit.h"             /* plan, header, footer, is, ok */
+#include "lua_test_utils.h"
 
 /*
  * This test checks all usage cases of luaT_tuple_new():
@@ -184,12 +185,7 @@ main()
 	memory_init();
 	fiber_init(fiber_c_invoke);
 
-	struct lua_State *L = luaL_newstate();
-	luaL_openlibs(L);
-
-	/* luaT_newmodule() assumes that this table exists. */
-	lua_newtable(L);
-	lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
+	struct lua_State *L = luaT_newteststate();
 
 	box_init();
 	tarantool_lua_error_init(L);
diff --git a/test/unit/lua_msgpack.c b/test/unit/lua_msgpack.c
index 106097638..628ad1259 100644
--- a/test/unit/lua_msgpack.c
+++ b/test/unit/lua_msgpack.c
@@ -14,6 +14,7 @@
 
 #define UNIT_TAP_COMPATIBLE 1
 #include "unit.h"
+#include "lua_test_utils.h"
 
 static void
 mpstream_error_mock(void *ctx)
@@ -151,16 +152,11 @@ main(void)
 	plan(2);
 	header();
 
-	struct lua_State *L = luaL_newstate();
-	luaL_openlibs(L);
+	struct lua_State *L = luaT_newteststate();
 	tarantool_L = L;
 	memory_init();
 	fiber_init(fiber_c_invoke);
 
-	/* luaT_newmodule() assumes that this table exists. */
-	lua_newtable(L);
-	lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
-
 	tarantool_lua_serializer_init(L);
 	luaopen_msgpack(L);
 	lua_pop(L, 1);
diff --git a/test/unit/lua_test_utils.h b/test/unit/lua_test_utils.h
new file mode 100644
index 000000000..1ed5dcb36
--- /dev/null
+++ b/test/unit/lua_test_utils.h
@@ -0,0 +1,36 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#pragma once
+
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+/**
+ * Incapsulates steps we should do to create a Lua state suitable
+ * to initialize some built-in Lua modules.
+ */
+static inline struct lua_State *
+luaT_newteststate(void)
+{
+	struct lua_State *L = luaL_newstate();
+	luaL_openlibs(L);
+
+	/* luaT_newmodule() assumes that this table exists. */
+	lua_newtable(L);
+	lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
+
+	return L;
+}
+
+#if defined(__cplusplus)
+} /* extern "C" */
+#endif /* defined(__cplusplus) */
diff --git a/test/unit/lua_tweaks.c b/test/unit/lua_tweaks.c
index 56aa2b229..c5aa46253 100644
--- a/test/unit/lua_tweaks.c
+++ b/test/unit/lua_tweaks.c
@@ -15,6 +15,7 @@
 
 #define UNIT_TAP_COMPATIBLE 1
 #include "unit.h"
+#include "lua_test_utils.h"
 
 static struct lua_State *L;
 
@@ -239,13 +240,8 @@ main(void)
 {
 	memory_init();
 	fiber_init(fiber_c_invoke);
-	L = luaL_newstate();
-	luaL_openlibs(L);
-
-	/* luaT_newmodule() assumes that this table exists. */
-	lua_newtable(L);
-	lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_BUILTIN");
 
+	L = luaT_newteststate();
 	tarantool_lua_tweaks_init(L);
 
 	/*
```

Added.

src/lua/minifio.c
Totktonada: No problem, added.

```diff
diff --git a/src/lua/minifio.c b/src/lua/minifio.c
index 952ad86e2..7795408a8 100644
--- a/src/lua/minifio.c
+++ b/src/lua/minifio.c
@@ -60,5 +60,5 @@ tarantool_lua_minifio_init(struct lua_State *L)
        lua_getfield(L, LUA_REGISTRYINDEX, "_LOADED");
        luaT_newmodule(L, "internal.minifio", minifio_methods);
        lua_setfield(L, -2, "internal.minifio");
-       lua_pop(L, 1);
+       lua_pop(L, 1); /* _LOADED */
 }
```

src/lua/init.c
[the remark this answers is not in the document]
Totktonada: Changed to `lua_newtable(L)`.

test/unit/luaT_tuple_new.c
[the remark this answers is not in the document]
Totktonada: I added a helper named `luaT_newteststate()`. I'm not sure that it is what you asked, but it makes the code more clean.

test/unit/lua_tweaks.c
Totktonada: Added:

> In particular, tarantool_lua_init() funtcion is not called in a unit test.

test/app-luatest/override_test.lua
```diff
@@ -0,0 +1,146 @@
...
+    return {
+        exit_code = 0,
+        stdout = res,
+    }
+end
+
+-- A couple of test cases with overriding built-in modules.
+--
+-- In a general case it is not a trivial task to correctly
+-- override a built-in module. However there are modules that
```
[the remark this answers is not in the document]
Totktonada: Fixed. Resolved the discussion as trivial.

test/luajit-test-init.lua
Totktonada: Fixed. Resolved the discussion as trivial.

test/luajit-test-init.lua
Totktonada: Seems so. Fixed. Resolved the discussion as trivial.

test/treegen.lua
Totktonada: Renamed.

test/app-luatest/override_test.lua
```diff
@@ -0,0 +1,146 @@
...
+    'luadebug',
+    'memprof',
+    'msgpack',
+    'net.box',
+    'pickle',
+    'popen',
+    'pwd',
+    'socket',
+    'swim',
+    'sysprof',
```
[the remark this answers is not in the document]
Buristan: It’s OK, see [reply in this discussion](https://github.com/tarantool/tarantool/pull/8300#discussion_r1109540739).

igormunkin (approved):
**[PATCH 6/7] lua: allow to override a built-in module**

igormunkin (approved):
**[PATCH 7/7] lua: enable/disable override loader**

Totally agree with @Buristan regarding `conditional_loader` :)

Totktonada:
box-luatest/gh_8192_feedback_daemon_metrics_test.lua fails on the coverage job in CI, but passes locally on a debug build. I'm looking into it.

test/luajit-test-init.lua
igormunkin: Alright, let's leave it as is then.

src/lua/init.c
igormunkin: Thanks!

test/unit/luaT_tuple_new.c
igormunkin: Precisely, thanks!

test/unit/lua_tweaks.c
igormunkin: Great, thanks!

src/lua/minifio.c
igormunkin: Thanks!

Totktonada:
> box-luatest/gh_8192_feedback_daemon_metrics_test.lua fails on the coverage job in CI, but passes locally on a debug build. I'm looking into it.

I decided to adjust a test. See details below.

@drewdzzz, please take a look if time permits.

I have added one commit. Copy-pasted its description.

> ### test: expect slow metrics collection on ASAN build
>
> The feedback daemon test case fails constantly in CI after recent changes for override modules. I found no direct correspondence between the changes and the feedback daemon code/test case.
>
> The test case runs metrics collection for 0.3 seconds with 0.01 interval. It accepts [22; 38] samples as a correct result. I got 7-22 samples on ASAN build in CI.
>
> It means that the tested code is working, but slower. The tested code don't run the newly added code, but the new code is run at tarantool initialization. I think that the fails in CI are due to influence of the changed initialization code in composition with LuaJIT's unstable performance: some traces are collected differently, GC runs on different points and we're here.
>
> I performed a brief comparison of amount of fails in the test in parallel run (8 logical cores, 16 runs in parallel) of baseline (master) and changed code with and without ASAN. I can't achieve results that would be stable enough for judgements. At least, I can't do it without a long running testing.
>
> I changed the feedback daemon test case to accept [5; 55] samples as a correct result, when tarantool is built with ASAN. The old [22; 38] range is accepted for other builds, as before

<details>
<summary>Patch (collapsed)</summary>

```diff
diff --git a/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua b/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
index 1c278c7b5..c5508040c 100644
--- a/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
+++ b/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
@@ -27,6 +27,7 @@ g.test_metrics = function()
     g.server:exec(function()
         local t = require('luatest')
         local fiber = require('fiber')
+        local tarantool = require('tarantool')
         local m = {}
         -- Method checks that we pass required arguments to collect defaults.
         -- Note that if one of assertions in the method fails, you will not
@@ -45,8 +46,12 @@ g.test_metrics = function()
         local fb = daemon.generate_feedback()
         -- This check may fail because of assertion failure in collect method.
         t.assert_type(fb.metrics, 'table')
-        -- Huge margin to prevent flakiness on slow environments like coverity.
-        t.assert_almost_equals(#fb.metrics, 30, 8)
+        -- Huge margin to prevent flakiness.
+        --
+        -- Even bigger for the ASAN build, which is known as slow: just several
+        -- samples are okay for this build type.
+        local margin = tarantool.build.flags:find('-fsanitize') and 25 or 8
+        t.assert_almost_equals(#fb.metrics, 30, margin)
         fb = daemon.generate_feedback()
         -- Check is metrics were reset. Fiber yields on feedback generation, so
         -- metrics may be not empty.
```
</details>

Totktonada:
`checks` integration test needs https://github.com/tarantool/checks/pull/51.

Totktonada:
Update regarding the failed test: I refused the idea to conditionally extend the range of allowed samples count. The test fails on macOS in CI too and it seems that the only way to proceed with the patchset now is to accept nearly any amount of metrics samples in the test on any build type.

PR #8302, opened 2023-02-13
static-build: don't use cat to apply multiple patches

locker:
The problem is if `cat` fails, because a patch file doesn't exist `PATH_COMMAND` written like this won't detect it, because the last command (`patch`) will complete successfully (apply existing patches found by `cat`):

```
cat XXX.patch YYY.patch | patch -p1
```

The proper way is to use `PATCH_COMMAND` continuation:

```
PATCH_COMMAND patch -p1 -i XXX.patch
COMMAND       patch -p1 -i YYY.patch
```


nshy (approved):

locker:
Cherry-picked to 2.10.

PR #8303, opened 2023-02-13
gdb: fixpack

### LINKED ISSUE #8263: tarantool-gdb.py tt-list doesn't work
Reproduced with Tarantool 2.11.0-entrypoint-970-g98dd8e69e014 and gdb 9.2-0ubuntu1~20.04.1.

Dump a core:

```
tarantool - &>/dev/null <<EOF
box.cfg{}
os.execute('kill -s 11 ' .. require('tarantool').pid())
EOF
```

Open the core in gdb and run `tt-list engines` (this is an example from `help tt-list`):

```
gdb tarantool core
source src/tarantool-gdb.py
tt-list engines
```

Result:

```
(gdb) source src/tarantool-gdb.py
No struct type named checkpoint_entry.
No struct type named memtx_join_entry.
No struct type named ck_constraint.
No struct type named fk_constraint.
No struct type named ck_constraint.
No struct type named fk_constraint.
(gdb) tt-list engines
Python Exception <class 'TypeError'> 'address' is an invalid keyword argument for this function:
Python Exception <class 'TypeError'> 'address' is an invalid keyword argument for this function:
$1 = {prev = 0x55a66d716bd8, next = 0x55a66d7176e8}
```

### LINKED ISSUE #8265: tarantool-gdb.py tt-tuple doesn't print unicode characters
Reproduced with Tarantool 2.11.0-entrypoint-970-g98dd8e69e014 and gdb 9.2-0ubuntu1~20.04.1.

Dump a core:

```
tarantool - &>/dev/null <<EOF
box.tuple.new('foo', 'баррэ')
os.execute('kill -s 11 ' .. require('tarantool').pid())
EOF
```

Open the core in gdb and run print a tuple:

```
gdb tarantool core
source src/tarantool-gdb.py
p *box_tuple_last
```

Result:

image

(Unprintable characters break the terminal.)

Without `tt-tuple` gdb prints unicode strings properly:

image

### LINKED ISSUE #8266: tarantool-gdb.py rlist containers are out of sync
Reproduced with Tarantool 2.11.0-entrypoint-970-g98dd8e69e014 and gdb 9.2-0ubuntu1~20.04.1.

Dump a core:

```
tarantool - &>/dev/null <<EOF
os.execute('kill -s 11 ' .. require('tarantool').pid())
EOF
```

Open the core in gdb and include tarantool-gdb.py:

```
(gdb) source src/tarantool-gdb.py
No struct type named checkpoint_entry.
No struct type named memtx_join_entry.
No struct type named ck_constraint.
No struct type named fk_constraint.
No struct type named ck_constraint.
No struct type named fk_constraint.
```

These structs were removed from recent Tarantool. It isn't enough to just update the list to fix this issue. We need to come up with an automated procedure to keep the container list in sync.

### LINKED ISSUE #8299: tarantool-gdb.py tt-mp doesn't support spaces in MsgPack expression
Command 'tt-mp' expects that MspPack expression is specified as a single argument, so it works incorrectly when there is a space in expression

(gdb) tt-mp (char *)box_tuple_last+6
Python Exception <class 'gdb.error'>: A syntax error in expression, near `'.
Error occurred in Python: A syntax error in expression, near `'.

Need to adjust argument parsing so it works correctly in such a case

'tt-tuple' has the same problem for its tuple expression

elhimov:
- Support unicode in MsgPack MP_STR type

- Display MP_BIN type in base64

- Don't worry (and don't warn user) if some structures from the list of possible rlist entries don't exist in the current instance of Tarantool (put these messages into 'debug' level). In fact it is expected that some of them might not exist

- Don't use 'address' keyword argument in 'gdb.Value.format_string' method because gdb of versions 9 and 10 doesn't support it

- Reimplement argument parsing for 'tt-mp' and 'tt-tuple' commands so MsgPack and Tuple expressions are parsed correctly when contains spaces

NO_DOC=gdb extension
NO_CHANGELOG=gdb extension
NO_TEST=gdb extension

### DIFF
--- a/src/tarantool-gdb.py
+++ b/src/tarantool-gdb.py
@@ -4,7 +4,10 @@ To use, just put 'source <path-to-this-file>' in gdb.
 """
 
 import gdb.printing
+import gdb.types
 import argparse
+import base64
+import logging
 import struct
 import itertools
 import re
@@ -15,6 +18,11 @@ slice = itertools.islice
 if sys.version_info[0] == 2:
     filter = itertools.ifilter
     zip = itertools.izip
+elif sys.version_info[0] == 3:
+    unicode = str
+
+logger = logging.getLogger('gdb.tarantool')
+logger.setLevel(logging.WARNING)
 
 def dump_type(type):
     return 'tag={} code={}'.format(type.tag, type.code)
@@ -47,12 +55,6 @@ def cast_ptr(dest_type, ptr, offset):
 
 def container_of(ptr, container_type, field):
     return cast_ptr(container_type, ptr, -(container_type[field].bitpos // 8)).dereference()
-
-def type_has_field(type, field_name):
-    for field in type.fields():
-        if field.name == field_name:
-            return True
-    return False
 
 INT32_MAX = 2**31 - 1
 INT32_MIN = -2**31
@@ -368,7 +370,7 @@ class MsgPack(object):
                 k -= 1
                 continue
             elif l > cls.MP_HINT:
-                k -= l;
+                k -= l
                 k -= 1
                 continue
             else:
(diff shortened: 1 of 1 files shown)


locker (changes requested):
It looks like there are three independent fixes in this commit - one for `tt-list` and and two for `tt-mp`. Please split, otherwise it's difficult to review.

Also, closing #8266 is incorrect, because we it's not just about warnings. As I wrote in the issue description,

> We need to come up with an automated procedure to keep the container list in sync.

elhimov:
> Also, closing https://github.com/tarantool/tarantool/issues/8266 is incorrect, because we it's not just about warnings

I'm totally agree that it would be much more convenient to have automatic sync mechanism, but I'm afraid it would require pretty much efforts so I doubt it worth it.

locker:
> > Also, closing #8266 is incorrect, because we it's not just about warnings
> 
> I'm totally agree that it would be much more convenient to have automatic sync mechanism, but I'm afraid it would require pretty much efforts so I doubt it worth it.

Then don't close the issue.

PR #8304, opened 2023-02-13
box: replace malloc with xmalloc in key_def_dup

ImeevMA:
This patch replaces malloc() with xmalloc() in key_def_dup() to avoid the possibility of skipping the malloc() return value check.

Closes tarantool/security#81

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring


locker (approved):

locker:
Cherry-picked to 2.10.

PR #8305, opened 2023-02-13
sql: properly check result of sql_get_coll_seq()

ImeevMA:
This patch fixes an issue with checking the result of sql_get_coll_seq() in sql_expr_coll(). This fix only changes the error if the collation combination is invalid because sql_get_coll_seq() sets the is_aborted flag and error will be thrown in any case.

Closes tarantool/security#80

NO_DOC=change of returned error in rare case
NO_CHANGELOG=change of returned error in rare case


locker (approved):

locker:
Cherry-picked to 2.10.

PR #8306, opened 2023-02-13
box: check tuple_field_by_path first argument type

ImeevMA:
This patch adds a type check of the first argument of the tuple_field_by_path() function.

Closes tarantool/security#82

NO_DOC=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/box/lua/tuple.c
+++ b/src/box/lua/tuple.c
@@ -649,20 +649,17 @@ lbox_tuple_transform(struct lua_State *L)
 static int
 lbox_tuple_field_by_path(struct lua_State *L)
 {
-	struct tuple *tuple = luaT_istuple(L, 1);
-	/* Is checked in Lua wrapper. */
-	assert(tuple != NULL);
-	assert(lua_isstring(L, 2));
+	struct tuple *tuple = luaT_checktuple(L, 1);
 	size_t len;
-	const char *field = NULL, *path = lua_tolstring(L, 2, &len);
+	const char *path = luaL_checklstring(L, 2, &len);
 	if (len == 0)
 		return 0;
-	field = tuple_field_raw_by_full_path(tuple_format(tuple),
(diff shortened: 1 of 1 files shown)


src/box/lua/tuple.c
```diff
@@ -649,7 +649,7 @@ lbox_tuple_transform(struct lua_State *L)
 static int
 lbox_tuple_field_by_path(struct lua_State *L)
 {
-	struct tuple *tuple = luaT_istuple(L, 1);
+	struct tuple *tuple = luaT_checktuple(L, 1);
```
locker, line 652: Nit: The assertion for `tuple` is no longer needed. Let's remove.

Let's also remove the assertion for `path` and instead write:

```C
const char *path = luaL_checklstring(L, 2, &len);
```

test/box-luatest/func_test.lua
```diff
@@ -120,3 +120,12 @@ g.test_gh_7822_vfunc_format = function()
         t.assert_equals(box.space._vfunc:format(), box.space._func:format())
     end)
 end
+
+g.test_ghs_82_tuple_field_by_path = function()
```
locker, line 124: Why do you think we should add this test case to this file? They look completely unrelated.

BTW I think that it's okay to commit this fix without a test, because testing internal functions isn't really necessary.

locker (changes requested):

src/box/lua/tuple.c
ImeevMA: Fixed.

test/box-luatest/func_test.lua
ImeevMA: I decided it because of the test-file name.

Dropped test.

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8307, opened 2023-02-13
box: don't check return values of some cfg_gets() for NULL

Gumix:
Sometimes the return value of `cfg_gets()` is checked for NULL, and sometimes not. Actually this is intended, although a bit confusing. If an option can have a nil value, it must be checked for NULL, but if it can't be nil, there is no sense in it. The nil value can be assigned only by default, it cannot be set via `box.cfg{}`.

This patch removes the NULL checks for `cfg_gets("election_mode")` and `cfg_gets("election_fencing_mode")` because they are not nil by default. All other non-nil options (e.g. `cfg_gets("bootstrap_strategy")`) are already implemented without the NULL checks.

Follow-up tarantool/security#75

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -935,9 +935,6 @@ static enum election_mode
 box_check_election_mode(void)
 {
 	const char *mode = cfg_gets("election_mode");
-	if (mode == NULL)
-		goto error;
-
 	if (strcmp(mode, "off") == 0)
 		return ELECTION_MODE_OFF;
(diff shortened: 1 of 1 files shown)


sergepetrenko (approved):
I have no objections. Seems the checks are unnecessary.

Please note when cherry-picking to 2.10: string `election_fencing_mode` used to be a bool `election_fencing_enabled`.

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8309, opened 2023-02-13
NULL dereferences in `enumEitherTrie` and `unum_clone`

Serpentian:
Closes tarantool/security#59
Closes tarantool/security#61


Serpentian:
I'd rather don't apply patch for https://github.com/tarantool/security/issues/59, it seems to be false-positive. It's not supposed to happen: all code of the function rely on the assumption that `idx` and `data32` cannot be NULL at the same time

locker (approved):

locker:
Cherry-picked to 2.10.

PR #8324, opened 2023-02-15
datetime: implement comparison operations of datetime to string

### LINKED ISSUE #7445: datetime behaviour is not consistent with decimal and uuid
```lua
-- Can compare decimal with strings and numbers
tarantool> decimal.new(123) == '123'
---
- true
...

tarantool> decimal.new(123) == 123
---
- true
...
```

```lua
-- Can compare uuid with strings
tarantool> uuid.fromstr('744c5225-a962-40d2-a301-d401c43789eb') == '744c5225-a962-40d2-a301-d401c43789eb'
---
- true
...
```

```lua
--- Can't compare datetime with strings
tarantool> datetime.parse('2022-07-20T11:51:57.531005+0300') == '2022-07-20T11:51:57.531005+0300'
---
- false
...

```

tsafin:
Both `uuid` and `decimal` implement comparison operations of their objects with string literals, but `datetime` object was incompatible with that.

For consistency reasons, implement it similarly:
- equality operators (== and ~=) are silently compared with strings and if we could parse then we compare datetime values, if we could not parse then silently return false;
- all other operators (<, <=, >, >=) are _raising_, i.e. they will raise error if they could not parse given operand according to the extended ISO-8601 format.

```
tarantool> datetime.parse('2022-07-20T11:51:57.531005+0300') ==
'2022-07-20T11:51:57.531005+0300'
---
- true
...
```

Closes #7445

### DIFF
--- a/src/lua/datetime.lua
+++ b/src/lua/datetime.lua
@@ -276,6 +276,9 @@ local function check_range(v, from, to, txt, extra)
     end
 end
 
+-- Forward declaration.
+local datetime_parse_full
+
 local function dt_from_ymd_checked(y, M, d)
     local pdt = date_dt_stash_take()
     local is_valid = builtin.tnt_dt_from_ymd_checked(y, M, d, pdt)
@@ -399,6 +402,32 @@ end
 -- since Rata Die date)
 local function local_dt(obj)
     return builtin.tnt_dt_from_rdn(local_rd(local_secs(obj)))
+end
+
+local function datetime_cmp(lhs, rhs)
+    if is_string(lhs) then
+        lhs = datetime_parse_full(lhs, nil, 0, true)
+    end
+    if is_string(rhs) then
+        rhs = datetime_parse_full(rhs, nil, 0, true)
+    end
+    if not is_datetime(lhs) or not is_datetime(rhs) then
+        error('incompatible types for datetime comparison', 3)
+    end
+    local sdiff = lhs.epoch - rhs.epoch
+    return sdiff ~= 0 and sdiff or (lhs.nsec - rhs.nsec)
+end
+
+local function datetime_eq(lhs, rhs)
+    return datetime_cmp(lhs, rhs) == 0
+end
+
+local function datetime_lt(lhs, rhs)
+    return datetime_cmp(lhs, rhs) < 0
+end
+
+local function datetime_le(lhs, rhs)
+    return datetime_cmp(lhs, rhs) <= 0
 end
 
 --[[
@@ -852,13 +881,14 @@ end
        date [T] time [ ] time_zone
     Returns constructed datetime object and length of accepted string.
 ]]
-local function datetime_parse_full(str, tzname, offset, parse_fully)
+function datetime_parse_full(str, tzname, offset, is_raising)
     check_str(str, 'datetime.parse()')
+    assert(type(is_raising) == 'boolean')
     local date = ffi.new(datetime_t)
     local len = builtin.tnt_datetime_parse_full(date, str, #str, tzname, offset)
     if len > 0 then
-        if parse_fully and len ~= #str then
-            goto couldnot_parse_fully
+        if is_raising and len ~= #str then
+            goto could_not_parse_fully
         else
             return date, tonumber(len)
         end
@@ -868,7 +898,7 @@ local function datetime_parse_full(str, tzname, offset, parse_fully)
         error(("could not parse '%s' - ambiguous timezone"):format(str))
     end
     -- len <= 0 or partially parsed
-::couldnot_parse_fully::
+::could_not_parse_fully::
     error(("could not parse '%s'"):format(str))
 end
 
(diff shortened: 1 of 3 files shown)


sergepetrenko (changes requested):
https://github.com/tarantool/tarantool/pull/8324#discussion_r1108677369

src/lua/datetime.lua
```diff
@@ -915,6 +897,36 @@ local function datetime_parse_from(str, obj)
...
 
+local function datetime_cmp(lhs, rhs, is_raising)
+    if is_string(lhs) then
+        lhs = datetime_parse_full(lhs, '', 0, not is_raising)
+    end
+    if is_string(rhs) then
+        rhs = datetime_parse_full(rhs, '', 0, not is_raising)
+    end
+    if not is_datetime(lhs) or not is_datetime(rhs) then
+        if is_raising then
```
sergepetrenko, line 908: Pasting the same comment here so it's easier to discuss:

> equality operators (== and ~=) are silently compare with strings
and if we could parse then we compare datetime values,
if we could not parse then silently return false;

But why? For example, decimal reports an error even in equality operators:
```
tarantool> dec.new(1) == 'abracadabra'
---
- error: '[string "return dec.new(1) == ''abracadabra''"]:1: incorrect value to convert
    to decimal as 2 argument'
...
```

tsafin: As later @sergepetrenko has discovered, we actually have inconsistency between `uuid` and `decimal`, one of which is not raising errors on invalid strings at comparison, but another is raising. 

```
tarantool> uuid.fromstr('744c5225-a962-40d2-a301-d401c43789eb') == 'x744c5225-a962-40d2-a301-d401c43789eb'
---
- false
...

tarantool> decimal.new(123) == 'x123'
---
- error: '[string "return decimal.new(123) == ''x123''"]:1: incorrect value to convert
    to decimal as 2 argument'
```

So this is up to @Mons to decide what is more idiomatic approach here?

tsafin: Have changed `==` and `~=` operators to be raising if there is incorrect datetime literal used on either side (it considered incorrect if it's not fully parsable)
```diff
--- a/src/lua/datetime.lua
+++ b/src/lua/datetime.lua
@@ -835,21 +835,24 @@ end
        date [T] time [ ] time_zone
     Returns constructed datetime object and length of accepted string.
 ]]
-local function datetime_parse_full(str, tzname, offset, is_quiet)
+local function datetime_parse_full(str, tzname, offset, parse_fully)
     check_str(str, 'datetime.parse()')
     local date = ffi.new(datetime_t)
     local len = builtin.tnt_datetime_parse_full(date, str, #str, tzname, offset)
     if len > 0 then
-        return date, tonumber(len)
-    elseif is_quiet then
-        return nil
+        if parse_fully and len ~= #str then
+            goto couldnot_parse_fully
+        else
+            return date, tonumber(len)
+        end
     elseif len == -builtin.TZ_NYI then
         error(("could not parse '%s' - nyi timezone"):format(str))
     elseif len == -builtin.TZ_AMBIGUOUS then
         error(("could not parse '%s' - ambiguous timezone"):format(str))
-    else -- len <= 0
-        error(("could not parse '%s'"):format(str))
     end
+    -- len <= 0 or partially parsed
+::couldnot_parse_fully::
+    error(("could not parse '%s'"):format(str))
 end
 
 --[[
@@ -899,10 +902,10 @@ end
 
 local function datetime_cmp(lhs, rhs, is_raising)
     if is_string(lhs) then
-        lhs = datetime_parse_full(lhs, '', 0, not is_raising)
+        lhs = datetime_parse_full(lhs, nil, 0, true)
     end
     if is_string(rhs) then
-        rhs = datetime_parse_full(rhs, '', 0, not is_raising)
+        rhs = datetime_parse_full(rhs, nil, 0, true)
     end
     if not is_datetime(lhs) or not is_datetime(rhs) then
         if is_raising then
diff --git a/test/app-tap/datetime.test.lua b/test/app-tap/datetime.test.lua
index 59ba07a6e..fd4ac8b75 100755
--- a/test/app-tap/datetime.test.lua
+++ b/test/app-tap/datetime.test.lua
@@ -194,11 +194,11 @@ test:test("Default date creation and comparison", function(test)
     test:is(ts1 > '1970-01-01', false, "not ts1 > 1970-01-1")
     test:is(ts1 >= '1970-01-01', true, "ts1 >= 1970-01-1")
 
-    -- equality comparisons are quiet for invalid strings
-    test:isnt(ts1 == 'x1970-01-01', "not ts1 == x1970-01-1")
-    test:isnt(ts1 ~= 'x1970-01-01', "not ts1 ~= x1970-01-1")
-
-    -- but other comparison operations are quite vocal
+    -- all comparisons operations are raising errors for invalid strings
+    assert_raises(test, couldnt_parse('x1970-01-01'),
+                  function() return ts1 == 'x1970-01-01' end)
+    assert_raises(test, couldnt_parse('x1970-01-01'),
+                  function() return ts1 ~= 'x1970-01-01' end)
     assert_raises(test, couldnt_parse('x1970-01-01'),
                   function() return ts1 < 'x1970-01-01' end)
     assert_raises(test, couldnt_parse('x1970-01-01'),
```

src/lua/datetime.lua
```diff
@@ -915,6 +900,36 @@ local function datetime_parse_from(str, obj)
     end
 end
 
+local function datetime_cmp(lhs, rhs, is_raising)
```
sergepetrenko, line 903: There's another `datetime_parse_full` call on line 897, where you forgot to pass `false` as 4th parameter. Please do so.

src/lua/datetime.lua
```diff
@@ -915,6 +900,36 @@ local function datetime_parse_from(str, obj)
...
+        else
+            return nil
+        end
+    end
+    local sdiff = lhs.epoch - rhs.epoch
+    return sdiff ~= 0 and sdiff or (lhs.nsec - rhs.nsec)
+end
+
+local function datetime_eq(lhs, rhs)
+    return datetime_cmp(lhs, rhs, false) == 0
```
sergepetrenko, line 922: Since `datetime_eq` raises when comparing a datetime to a malformed string, it should raise when compairing with other incompatible types.

That is, you don't need `is_raising` parameter of `datetime_cmp` anymore, because it should always be raising.

sergepetrenko: Please ask the products whether we need a compat option for this change.

sergepetrenko (changes requested):
Thanks for the fixes!

A couple more comments from my side.

src/lua/datetime.lua
ligurio: Fixed:

```diff
@@ -901,7 +902,7 @@ local function datetime_parse_from(str, obj)
 
     if not fmt or fmt == '' or fmt == 'iso8601' or fmt == 'rfc3339' then
         -- Effect of .tz overrides .tzoffset
-        return datetime_parse_full(str, tzname, offset)
+        return datetime_parse_full(str, tzname, offset, false)
     else
         return datetime_parse_format(str, fmt)
     end
```

PR #8325, opened 2023-02-15
compat: use tarantool.io URLs for compat options

Totktonada:
We plan to provide compat options information via the website. It was requested by Kirill Yukhin instead of the wiki pages (and I don't mind it).

### DIFF
--- a/src/lua/compat.lua
+++ b/src/lua/compat.lua
@@ -20,7 +20,7 @@ json.encode() result. The old and the new behavior produce a result, which
 is compatible ith the JSON specification. However most of other JSON encoders
 don't escape the forward slash, so the new behavior is considered more safe.
 
-https://github.com/tarantool/tarantool/wiki/compat%3Ajson_escape_forward_slash
+https://tarantool.io/compat/json_escape_forward_slash
 ]]
 
 local YAML_PRETTY_MULTILINE_BRIEF = [[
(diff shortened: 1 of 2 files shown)


src/lua/compat.lua
```diff
@@ -90,7 +90,7 @@ local options_order = { }
 
 local help = [[
 Tarantool compatibility module.
-To get help, see the Tarantool manual at https://github.com/tarantool/tarantool/wiki/compat .
+To get help, see the Tarantool manual at https://tarantool.io/compat
```
Totktonada, line 93: @NickVolynkin Highlighting your attention on this page just in case.

Totktonada, line 93: NB: Adjust this PR if #8283 lands first.

Totktonada, line 93: NB: Adjust this PR if #8338 lands first.

lastoCHka42 (approved):
Website with redirects deployed

src/lua/compat.lua
Totktonada: It works, closing.

Totktonada: Commented in PR #8283.

Totktonada: Commented in PR #8338.

PR #8327, opened 2023-02-15
compat: make compat a top level module

Totktonada:
Now it is accessible using `require('compat')` instead of `require('tarantool').compat`.

It follows our usual way to expose built-in modules.

It was not done initially due to doubts about projects, which have its own `myproject/compat.lua` and have modified `package.loaded` to obtain the project's compat as `require('compat')`.

It is not recommended and should be avoided. Project's modules should be imported using `require('myproject.<...>')`. Otherwise it may clash with a future built-in module or any external one.

Follows up #7000

### DIFF
--- a/changelogs/unreleased/gh-3012-yaml-prettier-multiline-output.md
+++ b/changelogs/unreleased/gh-3012-yaml-prettier-multiline-output.md
@@ -1,5 +1,5 @@
 ## feature/yaml
 
-* Now `yaml.encode` can encode multiline strings in literal-scalar style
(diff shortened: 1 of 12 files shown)


igormunkin (approved):
@Totktonada, thanks for the patch! Though I'm still against this particular change, I see no flaws in it and no reasons to block it. Hence, LGETM.

Totktonada:
Well, okay, I asked several persons within the team and most are neural, some are slightly positive or negative. I think I can bet that it's nice to have and proceed.

PR #8334, opened 2023-02-17
[alternative implementation] luajit: proxy -j and -b flags

### LINKED ISSUE #5541: Introduce -j and -b flags to be proxied to LuaJIT environment

fckxorg:
There are two flags in the LuaJIT useful for debugging and runtime configuration purposes: `-j` and `-b`. However, if you want to check the same Lua code from the Tarantool, you will need to make some adjustments in the script itself, as those flags are not present in the Tarantool's CLI.

This patch introduces those flags to the Tarantool, so debugging is much more convenient now. Flags are working the same as they do in the LuaJIT.

Closes #5541

NO_DOC=LuaJIT proxy
NO_CHANGELOG=LuaJIT proxy


src/main.cc
```diff
@@ -704,7 +722,7 @@ main(int argc, char **argv)
 		if (fd >= 0)
 			close(fd);
 		printf("Can't open script %s: %s\n",
-		       argv[1], tt_strerror(save_errno));
+			argv[1], tt_strerror(save_errno));
```
Totktonada, line 725: Nit: Unneeded change.

test/box-py/args.result
```diff
@@ -152,3 +156,28 @@ arg[2] => 2
 arg[3] => 3
 arg[4] => --help
 
+-b
+Save Tarantool bytecode: tarantool -b[options] input output
+  -l        Only list bytecode.
+  -s        Strip debug info (default).
+  -g        Keep debug info.
+  -n name   Set module name (default: auto-detect from input name).
+  -t type   Set output file type (default: auto-detect from output name).
```
Totktonada, line 165: It is work in the form `-b -t <...>`? Will not `getopt()` complain about an unknown option?

src/lua/dobytecode.lua
```diff
@@ -0,0 +1,13 @@
+local M = {}
+
+M.dobytecode = function(...)
+    local arg = {...}
+    if string.sub(arg[1], 3, 3) ~= "" then
+        arg[1] = '-' .. string.sub(arg[1], 3, #arg[1])
+    else
+        table.remove(arg, 1)
+    end
```
Totktonada, line 9: I would request a comment (in the code) here.

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,53 @@
...
+local function runcmdopt(func, opt)
+    if opt ~= nil and string.sub(opt, 1, 1) ~= "" then
+        local args = {}
+        while true do
+            local next_ind = string.find(opt, ',')
+            if next_ind == nil then break
+            elseif next_ind == 1 then table.insert(args, nil)
+            else
+                table.insert(args, string.sub(opt, 1, next_ind))
+                print(args[#args])
```
Totktonada, line 12: Forgotten debug print?

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,53 @@
...
+end
+
+local function loadjitmodule(module_name)
+    local full_name = 'jit.' .. module_name
+    local st, module = pcall(require, full_name)
+    if not st then module = {} end
+
+    local entry_point = module['start']
+    if entry_point == nil then
+        print('unknown luaJIT command or jit.* modules not installed')
```
Totktonada, line 30: Shouldn't we say it to stderr?

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,53 @@
...
+    return entry_point
+end
+
+M.dojitcmd = function(cmd)
+    local optind = string.find(cmd, '=')
+    local opt = nil
+    if optind ~= nil then
+        opt = string.sub(cmd, optind + 1, #cmd)
+        cmd = string.sub(cmd, 1, optind - 1)
+    end
```
Totktonada, line 41: Looks like `local cmd, opt = cmd:split('=', 1)`.

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,53 @@
...
+        while true do
+            local next_ind = string.find(opt, ',')
+            if next_ind == nil then break
+            elseif next_ind == 1 then table.insert(args, nil)
+            else
+                table.insert(args, string.sub(opt, 1, next_ind))
+                print(args[#args])
+                opt = string.sub(opt, next_ind + 1, #opt)
+            end
+        end
```
Totktonada, line 15: Is it `local args = opt:split(',')`?

src/main.cc
```diff
@@ -671,29 +674,44 @@ main(int argc, char **argv)
 			return 0;
 		case 'i':
 			/* Force interactive mode */
-			interactive = true;
+			opt_mask |= O_INTERACTIVE;
 			break;
+		case 'b':
+			opt_mask |= O_BYTECODE;
+			lj_arg = true;
+			optind--;
```
Totktonada, line 682: I would ask a comment in the code here.

src/main.cc
```diff
@@ -671,29 +674,44 @@ main(int argc, char **argv)
...
+				optv[optc++] = "-e";
 			optv[optc++] = optarg;
 			break;
 		default:
 			/* "invalid option" is printed by getopt */
 			return EX_USAGE;
 		}
+
+		if (lj_arg)
+			break;
```
Totktonada, line 705: I would ask a comment in the code here.

src/main.cc
```diff
@@ -671,29 +674,44 @@ main(int argc, char **argv)
...
 	}
 
 	/* Shift arguments */
 	argc = 1 + (argc - optind);
 	for (int i = 1; i < argc; i++)
 		argv[i] = argv[optind + i - 1];
 
-	if (argc > 1 && strcmp(argv[1], "-") && access(argv[1], R_OK) != 0) {
+	if (!(opt_mask & O_BYTECODE) && argc > 1 &&
+	    strcmp(argv[1], "-") && access(argv[1], R_OK) != 0) {
```
Totktonada, line 714: Hm. Why we shouldn't check for a file if `-b` is passed?

test/box-py/args.result
```diff
@@ -152,3 +156,28 @@ arg[2] => 2
 arg[3] => 3
 arg[4] => --help
 
+-b
+Save Tarantool bytecode: tarantool -b[options] input output
```
tsafin, line 160: Very useful help message for `-b` option! Thanks!

(It was very hard to figure out list of `-j` or `-b` suboptions supported, as `luajit` usage help was not very helpful.)

tsafin: Oh, I was too much optimistic here, and apparently that was from script itself, and we still do not add any useful information in the tarantool usage screen. :(

tsafin (approved):
I've rebased my #7929 on top of this branch, and played around `-b` options I care about. And they passed and work as expected.

Thanks to @Totktonada for suggestion to convert former LuaC code to a plain Lua - it's now much, much easier to read and understand!

LGTM

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,39 @@
+local M = {}
+
+local function runcmdopt(func, opt)
+    if opt ~= nil and string.sub(opt, 1, 1) ~= "" then
+        local args = opt:split(',')
+        for i = 1, #args do
+            if args[i] == "" then args[i] = nil end
+        end
+        return pcall(func, unpack(args))
```
Totktonada, line 9:
```lua
tarantool> unpack({1, nil, 3})
---
- 1
...
```

`unpack()` has no guarantees regarding holed arrays. In such a case it is better to use `unpack(args, 1, table.maxn(args))`.

BTW, what this `nil` assignment logic is doing? Shouldn't it be tested?

src/lua/dojitcmd.lua
```diff
@@ -0,0 +1,39 @@
+local M = {}
+
+local function runcmdopt(func, opt)
+    if opt ~= nil and string.sub(opt, 1, 1) ~= "" then
```
Totktonada, line 4: I would ask a comment here as well (similarly to `dobytecode()`).

src/main.cc
Totktonada: This information is hard to deduce from this code hunk itself, so I would ask to comment it. So a future developer will know, whether (s)he can touch this code and what to test if this is touched.

src/lua/dojitcmd.lua
tsafin: Yeah, agreed that these assignments to nil looks confusing.

Why do not simple pass to `return pcall(func, unpack(args))` without that massaging of `args`?

tsafin (approved):
Although I do want to see answers to Sasha' questions, but I'm ok on my side.

test/box-py/args.result
Totktonada: TBH, I don't get what and where do you want to see.

src/main.cc
```diff
@@ -794,14 +818,16 @@ main(int argc, char **argv)
...
 		 * Load user init script.  The script should have access
 		 * to Tarantool Lua API (box.cfg, box.fiber, etc...) that
 		 * is why script must run only after the server was fully
 		 * initialized.
 		 */
-		if (tarantool_lua_run_script(script, interactive, optc, optv,
-					     main_argc, main_argv) != 0)
+		if (tarantool_lua_run_script(script, opt_mask,
+					     optc, optv, main_argc, main_argv)
+		    != 0)
```
Totktonada, line 830: Nit: Formatting here is a bit trange.

Totktonada (approved):
I glanced over the patch and it looks okay for me in general.

Since Timur performed the manual testing I'll pass it over.

src/lua/dojitcmd.lua
Totktonada: Hm. Things a bit strange here for me, but okay, the original question is irrelevant now and I'll resolve this discussion.

src/main.cc
Totktonada: Looks solved.

Issue #7936, opened 2022-11-14
Fix binary path finding in Lua tests


Lord-KA:
Currently, in many tests tarantool binary is found as `arg[-1]`, while it is not always correct according to [Lua Reference](https://www.lua.org/manual/5.1/manual.html#6). In case of `$ lua -la b.lua t1 t2` binary path would be at `arg[-2]`. We currently implement correct path lookup in [LuaJIT](https://github.com/tarantool/luajit/blob/tarantool/test/lua-Harness-tests/test_assertion.lua#L14-L20), thus we need to add a similar helper and fix all existing tests.

Totktonada:
```lua
-- a.lua 
if arg == nil then
    print('a', 'arg == nil')
    return
end
print('a', -3, tostring(arg[-3]))
print('a', -2, tostring(arg[-2]))
print('a', -1, tostring(arg[-1]))
print('a', 0, tostring(arg[0]))
print('a', 1, tostring(arg[1]))
print('a', 2, tostring(arg[2]))
print('a', 3, tostring(arg[3]))
```

```lua
-- b.lua
if arg == nil then
    print('b', 'arg == nil')
    return
end
print('b', -3, tostring(arg[-3]))
print('b', -2, tostring(arg[-2]))
print('b', -1, tostring(arg[-1]))
print('b', 0, tostring(arg[0]))
print('b', 1, tostring(arg[1]))
print('b', 2, tostring(arg[2]))
print('b', 3, tostring(arg[3]))
```

```lua
$ lua5.1 -la b.lua
a	arg == nil
b	-3	nil
b	-2	lua5.1
b	-1	-la
b	0	b.lua
b	1	nil
b	2	nil
b	3	nil
```

```lua
$ tarantool -la b.lua
a	-3	nil
a	-2	nil
a	-1	/usr/bin/tarantool
a	0	b.lua
a	1	nil
a	2	nil
a	3	nil
b	-3	nil
b	-2	nil
b	-1	/usr/bin/tarantool
b	0	b.lua
b	1	nil
b	2	nil
b	3	nil
```

Tarantool is not affected by the problem. `arg[-1]` is always the executable.

Lord-KA:
I confirm, adding jit.on/off() doesn't make any difference.

Totktonada:
NB: #244, 3229ad16c3e8c3653c60e386f228705e9fd31c22, 248998d6c35caaedaa8e665037a2a34c84a9ecf7.

[...]

Issue #3136, opened 2018-02-11
custom package.loaders[2] not working


aleclarson:
[Moonscript][1] sets `package.loaders[2]` when you call `require "moonscript"`.

I've tested this in Lua 5.1 REPL with success, but using Tarantool REPL gives me:
`module 'xxx' not found`

Is this an issue with Tarantool or LuaJIT?

[1]: http://moonscript.org/

aleclarson:
Worth noting that this returns true, yet no module found:

```lua
moon = require 'moonscript'
package.loaders[2] == moon.moon_loader
```

aleclarson:
Okay, I found the problem.

`package.path` in Lua 5.1 and Lua 5.3 contains `./?.lua`, and Tarantool does not.

So here's the workaround:
```lua
package.path = './?.lua;' .. package.path
require 'moonscript'
```

Now I'm wondering how modules relative to `fio.cwd()` are even working in Tarantool...

Totktonada:
Loading from a current working directory is implemented in tarantool as a separate loader (which is `package.loaders[2]`). Not as the `package.path` loader (which, unlike Lua 5.1/LuaJIT, is `package.loaders[6]`, not `package.loaders[2]`):

https://github.com/tarantool/tarantool/blob/d4a4a74ee6264f85086de2d537b11dc69c1316fb/src/lua/init.lua#L193-L200

I don't know why it is implemented this way. Maybe it worth to remove those loaders and fill `package.path` / `package.cpath` instead. At least it'll make tarantool's behavior more obvious for one, who looks at `package.path` / `package.cpath` to understand what is going on. It'll require some tricks around `package.setsearchroot()` implementation[^1], but that seems to be solvable.

I tracked it down to cb3d321121aee3cafbea23f493c62e18dc6edacb (#2767) and still don't see a reason why don't add all needed paths into `package.path` / `package.cpath`. To handle cwd change? Don't know.

[^1]: Find and replace paths in `package.path` / `package.cpath` at `package.setsearchroot()` call?

PR #8335, opened 2023-02-17
lua: search modules next to the main script

### LINKED ISSUE #8182: Set `package.searchroot` by default from the main init.lua
Almost all Tarantool applications use `package.setsearchroot` function to extend the set of predefined paths for module lookup.

It is proposed to make the search root initialization at the platform startup to reduce the boilerplate for init.lua. Furthermore, as a result of this change, one can load his preload/override (for more info see #7774) modules from .rocks directory.

Totktonada:
It eliminates a need to call `package.setsearchroot()` in the main script to find modules in the application's directory.

Such layout is typical for cartridge based applications. It it also used, when different versions of modules are in use in different applications and installing the modules into the system is not an option.

Let's assume that there is an application in the directory /path/to/myapp with modules inside:

```
+ /path/to/myapp/
  +- .rocks/share/tarantool/
     +- foo.lua
  +- init.lua
  +- myapp/
     +- bar.lua
```

Now the following command automatically adds the path to the main script directory to search paths for modules:

```sh
$ tarantool /path/to/app/init.lua
```

So `require('foo')` and `require('myapp.bar')` works without any additional work.

Follows up #7774
Fixes #8182

### DIFF
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -250,6 +250,62 @@ table.insert(package.loaders, 5, gen_loader_func(search_rocks_lib, load_lib))
 -- package.cpath  7
 -- croot          8
 
+-- Search for modules next to the main script.
+--
+-- If the script is not provided (script == nil) or provided as
+-- stdin (script == '-'), there is no script directory, so nothing
+-- to do here.
+local script = minifio.script()
+if script ~= nil and script ~= '-' then
+    -- It is important to obtain the directory at initialization,
+    -- before any cwd change may occur. The script path may be
+    -- passed as relative to the current directory.
+    local script_dir = minifio.dirname(minifio.abspath(script))
+    local function script_dir_fn()
+        return script_dir
+    end
+
+    -- Search non-recursively, only next to the script.
+    local search_app_lua = gen_search_func(script_dir_fn, LUA_TEMPLATES)
+    local search_app_lib = gen_search_func(script_dir_fn, LIB_TEMPLATES)
+    local search_app_rocks_lua = gen_search_func(script_dir_fn,
+        ROCKS_LUA_TEMPLATES)
+    local search_app_rocks_lib = gen_search_func(script_dir_fn,
+        ROCKS_LIB_TEMPLATES)
+
+    -- Mix the script directory loaders into corresponding
+    -- searchroot based loaders. It allows to avoid changing
+    -- ordinals of the loaders and also makes the override
+    -- loader search here.
+    --
+    -- We can just add more paths to package.path/package.cpath,
+    -- but:
+    --
+    -- * Search for override modules is implemented as a loader.
+    -- * Search inside .rocks in implemented as a loaders.
+    --
+    -- And it is simpler to wrap this logic rather than repeat.
+    -- It is possible (and maybe even desirable) to reimplement
+    -- all the loaders logic as paths generation, but we should
+    -- do that for all the logic at once.
+    package.loaders[2] = chain_loaders({
+        package.loaders[2],
+        gen_loader_func(search_app_lua, load_lua),
+    })
+    package.loaders[3] = chain_loaders({
+        package.loaders[3],
+        gen_loader_func(search_app_lib, load_lib),
+    })
+    package.loaders[4] = chain_loaders({
+        package.loaders[4],
+        gen_loader_func(search_app_rocks_lua, load_lua),
+    })
+    package.loaders[5] = chain_loaders({
+        package.loaders[5],
+        gen_loader_func(search_app_rocks_lib, load_lib),
+    })
+end
+
 local function getenv_boolean(varname, default)
     local envvar = os.getenv(varname)
 
--- a/test/app-luatest/module_search_test.lua
+++ b/test/app-luatest/module_search_test.lua
@@ -1,0 +1,94 @@
+local fio = require('fio')
+
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+
+local g = t.group()
+
+local MODULE_SCRIPT_TEMPLATE = [[
+print(require('json').encode({
+    ['script'] = '<script>',
+}))
+return {whoami = '<module_name>'}
+]]
+
+-- Print a result of the require call.
+local MAIN_SCRIPT_TEMPLATE = [[
+print(require('json').encode({
+    ['script'] = '<script>',
+    ['<module_name>'] = require('<module_name>'),
+}))
+]]
+
+g.before_all(function(g)
+    treegen.init(g)
+    treegen.add_template(g, '^main%.lua$', MAIN_SCRIPT_TEMPLATE)
+    treegen.add_template(g, '^.*%.lua$', MODULE_SCRIPT_TEMPLATE)
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+local function expected_output(module_relpath, module_name)
+    local res = {
+        {
+            ['script'] = module_relpath,
+        },
+        {
+            ['script'] = 'main.lua',
+            [module_name] = {
+                whoami = module_name,
+            },
+        },
+    }
+
+    return {
+        exit_code = 0,
+        stdout = res,
+    }
+end
+
+-- Create an 'application directory' with a main script and a
+-- module. Run the main script and ensure that the module is
+-- successfully required.
+for _, case in ipairs({
+    -- Application's 'foo' module.
+    {'foo.lua', 'foo'},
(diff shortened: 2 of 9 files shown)


Buristan (commented):
Hi, Alexander!
Thanks for the patch-set!
I'll proceed with review per-patch.

**[PATCH 1/2] lua: add minifio.script() to get main script path**
LGTM.

src/lua/loaders.lua
```diff
@@ -250,6 +250,47 @@ table.insert(package.loaders, 5, gen_loader_func(search_rocks_lib, load_lib))
...
+    -- Search non-recursively, only next to the script.
+    local search_app_lua = gen_search_func(script_dir_fn, LUA_TEMPLATES)
+    local search_app_lib = gen_search_func(script_dir_fn, LIB_TEMPLATES)
+    local search_app_rocks_lua = gen_search_func(script_dir_fn,
+        ROCKS_LUA_TEMPLATES)
+    local search_app_rocks_lib = gen_search_func(script_dir_fn,
+        ROCKS_LIB_TEMPLATES)
+
+    -- Mix the script directory loaders into corresponding
+    -- searchroot based loaders. It allows to don't change
```
Buristan, line 273: Minor: I suggest rephrasing it like the following:
`It allows to avoid changing ordinals of the loaders ...`
The original comment is hard for me to read:(.

test/app-luatest/module_search_test.lua
```diff
@@ -0,0 +1,94 @@
...
+    {'override/socket.lua', 'socket'},
+    {'.rocks/share/tarantool/override/socket.lua', 'socket'},
+}) do
+    local module_relpath = case[1]
+    local module_name = case[2]
+
+    local module_slug = module_relpath
+        :gsub('^%.rocks/share/tarantool/', 'rocks/'):gsub('/', '_'):sub(1, -5)
+
+    g['test_' .. module_slug] = function(g)
```
Buristan, line 78: Side note: Why "slug"?:)

src/lua/loaders.lua
```diff
@@ -250,6 +250,47 @@ table.insert(package.loaders, 5, gen_loader_func(search_rocks_lib, load_lib))
 -- package.cpath  7
 -- croot          8
 
+-- Search for modules next to the main script.
+local script = minifio.script()
+if script ~= nil and script ~= '-' then
```
Buristan, line 255: Side note: This part gave me some interesting ideas to explore:

```sh
$ cat $PWD/-
print("test")
$ luajit -e 'print("-e")' -- -
-e
test
$ lua -e 'print("-e")' -- -
-e
test
$ src/tarantool -e 'print("-e")' -- -
-e
^C^C^C^C^C^C^Z
[1]+  Stopped                 src/tarantool -e 'print("-e")' -- -
```

`--`: stops handling options in Lua/LuaJIT (see [Lua 5.1 Reference Manual](https://www.lua.org/manual/5.1/manual.html#6))
As far as it is not implemented via Tarantool we can ignore this case, for now.

Buristan (approved):
**[PATCH 2/2] lua: search modules next to the main script**
LGTM, just a few reflections below.

test/app-luatest/module_search_test.lua
Totktonada: TBH, I don't know from where it comes.

https://www.npmjs.com/package/slugify
https://pypi.org/project/python-slugify/
https://github.com/sindresorhus/slugify

src/lua/loaders.lua
Totktonada: No problems, changed.

test/app-luatest/module_search_test.lua
Buristan: Interesting, thanks!:)

igormunkin (approved):
**[PATCH 1/2] lua: add minifio.script() to get main script path**

@Totktonada, thanks for the patch! LGTM, but I'm disappointed with the way CLI arguments are handled by Tarantool.

```
$ cat t.lua
print(arg[0])
$ ./src/tarantool t.lua 
t.lua
$ ./src/tarantool -e 'print(arg[0])'
./src/tarantool
$ luajit t.lua
t.lua
$ luajit -e 'print(arg[0])'
nil
```

I have no idea, why Tarantool handles arguments this way; LuaJIT approach allows to solve the original issue without `minifio.script` helper (hence this patch is not needed).

Anyway, no changes are required within this patch.

Totktonada:
In turn, I dislike some aspects of how LuaJIT forms `arg` (see #7936). Anyway, I think that it is good to have an internal helper that is independent from our position on how `arg` should be interpreted in different contexts. (Maybe it worth to cover `minifio.script()` with a test, hm. I'll add a task to my wishlist.)

igormunkin (approved):
**[PATCH 2/2] lua: search modules next to the main script**

@Totktonada, thanks for the patch! LGTM except some nits, I've left below.

Typo: s/It it/It is/.

Minor: there is not a single word regarding the rationale for such a complex change instead of simply patching `package.path`/`package.cpath` and `LUA_PATH`/`LUA_CPATH`. I suggest to left the comment both in the commit message and in `src/lua/loaders` nearby the code patching the default loaders.

test/app-luatest/module_search_test.lua
```diff
@@ -0,0 +1,94 @@
...
+    {'.rocks/share/tarantool/foo.lua', 'foo'},
+    {'.rocks/share/tarantool/foo/init.lua', 'foo'},
+    {'.rocks/share/tarantool/app/foo.lua', 'app.foo'},
+    {'.rocks/share/tarantool/app/foo/init.lua', 'app.foo'},
+    -- Application's socket override module.
+    --
+    -- See also override_test.lua.
+    {'override/socket.lua', 'socket'},
+    {'.rocks/share/tarantool/override/socket.lua', 'socket'},
+}) do
```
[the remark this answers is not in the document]
Totktonada: A kind of my pattern: wrap everything related to a test case into a table. Sometimes it has named fields, sometimes just two items. I would leave it as present.

src/lua/loaders.lua
Totktonada: No problems.

```diff
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index 2d7cb69b0..794f74eb5 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -251,6 +251,10 @@ table.insert(package.loaders, 5, gen_loader_func(search_rocks_lib, load_lib))
 -- croot          8
 
 -- Search for modules next to the main script.
+--
+-- If the script is not provided (script == nil) or provided as
+-- stdin (script == '-'), there is no script directory, so nothing
+-- to do here.
 local script = minifio.script()
 if script ~= nil and script ~= '-' then
     -- It is important to obtain the directory at initialization,
```

Totktonada:
> Minor: there is not a single word regarding the rationale for such a complex change instead of simply patching `package.path`/`package.cpath` and `LUA_PATH`/`LUA_CPATH`. I suggest to left the comment both in the commit message and in `src/lua/loaders` nearby the code patching the default loaders.

In fact, I would rather implement all the loaders as package.{path,cpath} generators, but while we have them as loaders, the new logic is easier to add as loader wrappers. (See also a discussion in #3136).

I have added a comment into the code.

<details>
<summary>The change (collapsed).</summary>

```diff
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index 23d294cee..794f74eb5 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -277,6 +277,17 @@ if script ~= nil and script ~= '-' then
     -- searchroot based loaders. It allows to avoid changing
     -- ordinals of the loaders and also makes the override
     -- loader search here.
+    --
+    -- We can just add more paths to package.path/package.cpath,
+    -- but:
+    --
+    -- * Search for override modules is implemented as a loader.
+    -- * Search inside .rocks in implemented as a loaders.
+    --
+    -- And it is simpler to wrap this logic rather than repeat.
+    -- It is possible (and maybe even desirable) to reimplement
+    -- all the loaders logic as paths generation, but we should
+    -- do that for all the logic at once.
     package.loaders[2] = chain_loaders({
         package.loaders[2],
         gen_loader_func(search_app_lua, load_lua),
```
</details>

Added a remark into the commit message:

> Regarding the implementation. There are two possible ways to add more searching logic: add more entries into `package.path` and `package.cpath` or add/wrap a loader. I'm going the second way, because we already have a logic around .rocks and override modules implemented as loaders. It is easier to follow.
>
> However, I would reimplement all those loaders as paths generators for `package.path` and `package.cpath` in a future. See also #3136.

PR #8343, opened 2023-02-17
box: handle `region_alloc` failure in `tuple_field_map_create_plain`

locker:
Let's use `xregion_alloc` instead of `region_alloc`, because memory allocations from fiber region shouldn't normally fail, see #3534.

Closes tarantool/security#97


sergos (approved):
Please, add NO_DOC NO_TEST NO_CHANGELOG

locker:
Cherry-picked to 2.10.

Issue #6786, opened 2022-01-14
Iteration by a functional index is not stable


Totktonada:
Known affected tarantool versions:

2.4.1-57-g339ad6289
2.10.0-beta2-5-gdc19be406
2.10.0-beta2-40-g5cd399b77

The problem can be reproduced in the debug build:

> cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_BACKTRACE=ON -DENABLE_DIST=ON -DENABLE_FEEDBACK_DAEMON=OFF -DENABLE_BUNDLED_LIBCURL=OFF && make -j

Reproducer:

```lua
box.cfg()

local findex_code = [[function(tuple)
    return {tuple[2]}
end]]
box.schema.func.create('findex_func', {
    body = findex_code,
    is_deterministic = true,
    is_sandboxed = true,
})

local space = box.schema.create_space('s', {
    engine = 'memtx',
    format = {
        {
            name = 'id',
            type = 'number',
        },
        {
            name = 'data',
            type = 'string',
        },
    }
})

space:create_index('primary', {
    type = 'TREE',
    parts = {
        {
            field = 1,
        }
    }
})
space:create_index('functional_index', {
    type = 'TREE',
    parts = {
        {
            field = 1,
            type = 'string',
        },
    },
    func = 'findex_func',
})

space:insert({1, '12'})
space:insert({2, '21'})

for _, tuple in space.index.functional_index:pairs() do
    space:delete({tuple[1]})
end
```

(The reproducer is extracted from expirationd 1.1.1-44-g838c2d1, from the `test_memtx_tree_functional_index` test case.)

Result (those particular line numbers are from 2.10.0-beta2-40-g5cd399b77):

> tarantool: /home/alex/p/tarantool-meta/r/t-6/src/box/tuple_compare.cc:1394: int func_index_compare(tuple*, hint_t, tuple*, hint_t, key_def*) [with bool is_nullable = false; hint_t = long unsigned int]: Assertion `mp_typeof(*key_b) == MP_ARRAY' failed.
> Aborted (core dumped)

C and Lua backtraces from the coredump:

```c
(gdb) bt
#0  0x00007f3c941581cc in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f3c9410c982 in raise () from /lib64/libc.so.6
#2  0x00007f3c940f7449 in abort () from /lib64/libc.so.6
[...]

PR #8344, opened 2023-02-17
memtx: fix size calculation in functional index chunks

alyapunov:
When function of a functional index is called, the result is stored in memtx allocator chunks among data tuples. By design, memtx allocator requires size of allocation for deallocation, so the size has to be stored along with the data (actually right before it). By a mistake, when being deleted, the size of data was retrieved slightly wrong, giving the value of 4 bytes less. Due to the allocator specific design the size error leads to a rare crashes when the size of functional index function result was about 160 bytes (157..160 bytes with default config). It seems that sizes about 320 etc are affected to.

Fix it by correct size evaluation of functional index chunks.

Hotfix of #6786


locker (approved):

test/box-luatest/gh-6786_func_index_iterator_stable_test.lua
```diff
@@ -57,3 +57,54 @@ g.test_func_index_iterator_stable = function()
         t.assert_equals(s.index.f:select(), {})
     end)
 end
+
+g.before_test('test_func_index_allocator', function()
```
locker, line 61: I think we should commit the test to the master and backport it to 2.8 and 2.10.

alyapunov: done.

PR #8348, opened 2023-02-17
lua: allow to override a module only for application's code

Totktonada:
Use case: an override module is designed to replace a built-in module only when the module is required from an application, not from the platform itself.
    
Now it is possible:
    
```lua
-- override/fio.lua
local loaders = require('internal.loaders')

if loaders.initializing then
    loaders.no_package_loaded.fio = true
    return loaders.builtin.fio
end

return {
    whoami = 'override.fio',
}
```

Follows up #7774

### DIFF
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -967,6 +967,27 @@ tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 	tarantool_L = L;
 }
 
+int
+tarantool_lua_postinit(struct lua_State *L)
+{
+	/*
+	 * loaders.initializing = nil
+	 *
+	 * The loaders module set the `initializing` field to
+	 * `true` at first load (during tarantool_lua_init()).
+	 * Now it is time to set it to `nil` to state that all
+	 * the built-in modules are loaded.
+	 */
+	lua_getfield(L, LUA_GLOBALSINDEX, "require");
+	lua_pushstring(L, "internal.loaders");
+	lua_call(L, 1, 1);
+	lua_pushnil(L);
+	lua_setfield(L, -2, "initializing");
+	/* Pop internal.loaders table. */
+	lua_pop(L, 1);
+	return 0;
+}
+
 char *history = NULL;
 
 struct slab_cache *
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -373,6 +373,27 @@ rawset(package, "search", search)
 rawset(package, "searchroot", searchroot)
 rawset(package, "setsearchroot", setsearchroot)
 
+local no_package_loaded = {}
+local raw_require = require
+
+-- Allow an override module to refuse caching in package.loaded.
+--
+-- It may be useful to return a built-in module into the platform,
+-- but the override module into application's code.
+_G.require = function(modname)
+    if no_package_loaded[modname] then
+        no_package_loaded[modname] = nil
+        package.loaded[modname] = nil
+    end
+    -- NB: This call is a tail call and it is important to some
+    -- extent. At least luajit's test suites expect certain error
+    -- messages from `require()` calls and from the `-l` option.
+    --
+    -- A non-tail call changes the filename at the beginning of
+    -- the error message.
+    return raw_require(modname)
+end
+
 return {
     ROCKS_LIB_PATH = ROCKS_LIB_PATH,
     ROCKS_LUA_PATH = ROCKS_LUA_PATH,
@@ -386,4 +407,15 @@ return {
     -- It is `true` during tarantool initialization, but once all
     -- the built-in modules are ready, will be set to `nil`.
     initializing = true,
+    raw_require = raw_require,
+    -- Add a module name here to ignore package.loaded[modname] at
+    -- next require(). The flag is dropped at the require() call.
+    --
+    -- This is intended to be used from inside a module to refuse
+    -- caching in package.loaded. Any other usage has unspecified
+    -- behavior (just manipulate package.loaded directly if you
+    -- need it).
+    --
+    -- Usage: loaders.no_package_loaded[modname] = true
+    no_package_loaded = no_package_loaded,
 }
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -383,4 +383,7 @@ return {
     override_builtin_disable = function()
         override_loader_is_enabled = false
     end,
+    -- It is `true` during tarantool initialization, but once all
+    -- the built-in modules are ready, will be set to `nil`.
+    initializing = true,
 }
--- a/test/app-luatest/override_misc_test.lua
+++ b/test/app-luatest/override_misc_test.lua
@@ -1,0 +1,67 @@
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+
+local g = t.group()
+
+local OVERRIDE_SCRIPT_TEMPLATE = [[
+local loaders = require('internal.loaders')
+
+return {
+    whoami = 'override.<module_name>',
+    initializing = loaders.initializing,
+}
+]]
+
+-- Print a result of the require call.
+local MAIN_SCRIPT_TEMPLATE = [[
+local json = require('json')
+local loaders = require('internal.loaders')
+
+print(json.encode({
+    ['<module_name>'] = require('<module_name>'),
+    ['<script>'] = {
+        whoami = '<script>',
+        initializing = loaders.initializing,
+    }
+}))
+]]
+
+g.before_all(function(g)
+    treegen.init(g)
+    treegen.add_template(g, '^override/.*%.lua$', OVERRIDE_SCRIPT_TEMPLATE)
+    treegen.add_template(g, '^main%.lua$', MAIN_SCRIPT_TEMPLATE)
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+local function expected_output(module_name)
+    local res = {
+        {
+            [module_name] = {
+                whoami = ('override.%s'):format(module_name),
+                initializing = true,
+            },
+            ['main.lua'] = {
+                whoami = 'main.lua',
+                -- initializing is nil
+            }
+        }
+    }
+
+    return {
+        exit_code = 0,
+        stdout = res,
+    }
+end
+
+g.test_initializing = function(g)
+    local scripts = {'override/socket.lua', 'main.lua'}
+    local replacements = {module_name = 'socket'}
+    local dir = treegen.prepare_directory(g, scripts, replacements)
+    local res = justrun.tarantool(dir, {}, {'main.lua'})
+    local exp = expected_output('socket')
+    t.assert_equals(res, exp)
+end
--- a/test/app-luatest/override_misc_test.lua
+++ b/test/app-luatest/override_misc_test.lua
@@ -65,3 +65,26 @@ g.test_initializing = function(g)
     local exp = expected_output('socket')
     t.assert_equals(res, exp)
 end
+
+g.test_no_package_loaded = function()
+    local LOAD_LIMIT = 3
+    local LOAD_ATTEMPTS = LOAD_LIMIT + 1
+
+    local load_cnt = 0
+    local foo = {whoami = 'foo'}
+    package.preload.foo = function()
+        local loaders = require('internal.loaders')
+        load_cnt = load_cnt + 1
+        if load_cnt < LOAD_LIMIT then
+            loaders.no_package_loaded.foo = true
+        end
+        return foo
+    end
+
+    for _ = 1, LOAD_ATTEMPTS do
+        require('foo')
+    end
+
+    package.preload.foo = nil
+    t.assert_equals(load_cnt, LOAD_LIMIT)
+end
--- a/src/lua/init.h
+++ b/src/lua/init.h
@@ -53,6 +53,12 @@ void
 tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 		   char **argv);
 
+/**
+ * A code that runs after loading of all built-in modules.
+ */
+int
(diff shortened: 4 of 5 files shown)


test/app-luatest/override_misc_test.lua
```diff
@@ -0,0 +1,54 @@
...
+            initializing = true,
+        }
+    }
+
+    return {
+        exit_code = 0,
+        stdout = res,
+    }
+end
+
```
Buristan, line 46: I suggest also to add some base test, that `initializing` is not true, when we run this code.

Buristan (commented):
Hi, Alexander!
Thanks for the patch-set!
I'll proceed review per patch.

**[PATCH 1/2] lua: add internal modules initialization status**
Generally LGTM, just some minor suggestions about comments and one test-case.

> document ability of loading a replacement module itself, but I

Typo: s/ability/the ability/

test/app-luatest/override_misc_test.lua
```diff
@@ -52,3 +52,21 @@ g.test_initializing = function(g)
     local exp = expected_output('socket')
     t.assert_equals(res, exp)
 end
+
+g.test_no_package_loaded = function(g)
```
Buristan, line 56: I suggest to make test-case a little bit more interesting (or add another one):
After successful require, add our module to `no_package_loaded` and require it again to be sure that it will be removed from `package.loaded`.

test/app-luatest/override_misc_test.lua
```diff
@@ -52,3 +52,21 @@ g.test_initializing = function(g)
...
+    local load_cnt = 0
+    local foo = {whoami = 'foo'}
+    package.preload.foo = function()
+        local loaders = require('internal.loaders')
+        load_cnt = load_cnt + 1
+        if load_cnt < 3 then
```
Buristan, line 62: I suggest to use `LOAD_LIMIT` constant here and in `assert_equals()` below.

test/app-luatest/override_misc_test.lua
```diff
@@ -52,3 +52,21 @@ g.test_initializing = function(g)
...
+            loaders.no_package_loaded.foo = true
+        end
+        return foo
+    end
+    for i = 1, 10 do
```
Buristan, line 67: `i` is unused, see luacheck warnings.

test/app-luatest/override_misc_test.lua
Buristan, line 56: `g` is unused, see luacheck warnings.

src/lua/loaders.lua
```diff
@@ -317,6 +317,22 @@ rawset(package, "search", search)
...
 
+local no_package_loaded = {}
+local raw_require = require
+
+-- Allow an override module to refuse writing to package.loaded.
+--
+-- It may be useful to return a built-in module into the platform,
+-- but the override module into application's code.
+_G.require = function(modname)
+    local mod = raw_require(modname)
```
Buristan, line 328: This line will add to any error message about `require` its location (`builtin/internal.loaders.lua:328`) -- this is why <lua-Harness-tests/241-standalone.t> is failing:
```
tarantool> string.match("LuajitError: builtin/internal.loaders.lua:328: module 'no_lib' not found:", "^[^:]+: module 'no_lib' not found:")
---
- null
tarantool> string.match("LuajitError: module 'no_lib' not found:", "^[^:]+: module 'no_lib' not found:")
---
- 'LuajitError: module ''no_lib'' not found:'
```

We can simply set `require` to the `raw_require` in <luajit-test-init.lua> to avoid this error, if we think that this line will not bother our users.

test/app-luatest/override_misc_test.lua
Buristan: Also, can we use `LOAD_LIMIT + 1` here to be more consistent?

Buristan (changes requested):
**[PATCH 2/2] lua: allow to refuse caching in package.loaded**
Hi, Alexander!
Thanks for the patch!
Please, consider my comments below.

src/lua/init.c
```diff
@@ -961,6 +961,19 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
...
+int
+tarantool_lua_postinit(struct lua_State *L)
+{
+	/* loaders.initializing = nil */
+	lua_getfield(L, LUA_GLOBALSINDEX, "require");
+	lua_pushstring(L, "internal.loaders");
+	lua_call(L, 1, 1);
+	lua_pushnil(L);
+	lua_setfield(L, -2, "initializing");
+	lua_pop(L, 1);
```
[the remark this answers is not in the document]
Totktonada: No problem, added.

src/lua/init.c
```diff
@@ -961,6 +961,19 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
 	tarantool_L = L;
 }
 
+int
+tarantool_lua_postinit(struct lua_State *L)
+{
+	/* loaders.initializing = nil */
+	lua_getfield(L, LUA_GLOBALSINDEX, "require");
```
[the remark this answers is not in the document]
Totktonada: It is not different from other internal `require()` calls, so I'm not sure that I understand your desire. I have added a bit more context here:

```diff
diff --git a/src/lua/init.c b/src/lua/init.c
index d7fc2caee..5f2ef9a62 100644
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -968,12 +968,20 @@ tarantool_lua_init(const char *tarantool_bin, int argc, char **argv)
 int
 tarantool_lua_postinit(struct lua_State *L)
 {
-       /* loaders.initializing = nil */
+       /*
+        * loaders.initializing = nil
+        *
+        * The loaders module set the `initializing` field to
+        * `true` at first load (during tarantool_lua_init()).
+        * Now it is time to set it to `nil` to state that all
+        * the built-in modules are loaded.
+        */
        lua_getfield(L, LUA_GLOBALSINDEX, "require");
        lua_pushstring(L, "internal.loaders");
        lua_call(L, 1, 1);
```

Is it what you asked for?

test/app-luatest/override_misc_test.lua
Totktonada: Kinda this?

```diff
diff --git a/test/app-luatest/override_misc_test.lua b/test/app-luatest/override_misc_test.lua
index 47cf317a2..0287810b0 100644
--- a/test/app-luatest/override_misc_test.lua
+++ b/test/app-luatest/override_misc_test.lua
@@ -16,8 +16,15 @@ return {
 -- Print a result of the require call.
 local MAIN_SCRIPT_TEMPLATE = [[
 local json = require('json')
+local loaders = require('internal.loaders')
 
-print(json.encode(require('<module_name>')))
+print(json.encode({
+    ['<module_name>'] = require('<module_name>'),
+    ['<script>'] = {
+        whoami = '<script>',
+        initializing = loaders.initializing,
+    }
+}))
 ]]
 
 g.before_all(function(g)
@@ -33,8 +40,14 @@ end)
 local function expected_output(module_name)
     local res = {
         {
-            whoami = ('override.%s'):format(module_name),
-            initializing = true,
+            [module_name] = {
+                whoami = ('override.%s'):format(module_name),
+                initializing = true,
+            },
+            ['main.lua'] = {
+                whoami = 'main.lua',
+                -- initializing is nil
+            }
         }
     }
```

Added.

test/app-luatest/override_misc_test.lua
Totktonada: AFAIU you meant adding into `no_package_loaded` from outside of the module. It is not supposed to be used this way. In fact, it is not needed if you anyway have full control: add or remove anything from `package.loaded`.

This field is only needed to control `require()` behavior from inside the module. I don't know, whether it is good to hold a behavior that is not supposed to be used anyhow.

test/app-luatest/override_misc_test.lua
Buristan: Yep, it's OK.

test/app-luatest/override_misc_test.lua
Totktonada: Okay, done. Diff for this and several close comments:

```diff
diff --git a/test/app-luatest/override_misc_test.lua b/test/app-luatest/override_misc_test.lua
index f4814a2d8..96fa750c1 100644
--- a/test/app-luatest/override_misc_test.lua
+++ b/test/app-luatest/override_misc_test.lua
@@ -66,20 +66,25 @@ g.test_initializing = function(g)
     t.assert_equals(res, exp)
 end
 
-g.test_no_package_loaded = function(g)
+g.test_no_package_loaded = function()
+    local LOAD_LIMIT = 3
+    local LOAD_ATTEMPTS = LOAD_LIMIT + 1
+
     local load_cnt = 0
     local foo = {whoami = 'foo'}
     package.preload.foo = function()
         local loaders = require('internal.loaders')
         load_cnt = load_cnt + 1
-        if load_cnt < 3 then
+        if load_cnt < LOAD_LIMIT then
             loaders.no_package_loaded.foo = true
         end
         return foo
     end
-    for i = 1, 10 do
+
+    for _ = 1, LOAD_ATTEMPTS do
         require('foo')
     end
+
     package.preload.foo = nil
-    t.assert_equals(load_cnt, 3)
+    t.assert_equals(load_cnt, LOAD_LIMIT)
 end
```

test/app-luatest/override_misc_test.lua
Buristan: OK, I agree, that it may be stayed "undefined" (mean no guarantees) for users.

test/app-luatest/override_misc_test.lua
Totktonada: No problem. Added `LOAD_ATTEMPTS` as `LOAD_LIMIT + 1`.

test/app-luatest/override_misc_test.lua
Totktonada: Sorry for this mess. I was in a hurry a bit, but this doesn't excuses me. Fixed.

src/lua/loaders.lua
Totktonada: Sorry for this mess with failed CI. I saw it yesterday but had no time to fix.

I'll spin here for a while (found other LuaJIT test fails, at least when I attempt to fix it with `pcall()` + `error(<...>, 0)` on tarantool's side).

test/app-luatest/override_misc_test.lua
Buristan: It's OK, we are not robots (at least most of us):).

test/app-luatest/override_misc_test.lua
Buristan: Nice, thanks!

src/lua/loaders.lua
Totktonada: I changed the order of operations in the require wrapper: first drop `package.loaded`, next call `raw_require()`. This way the call is a tail call and the error messages remain the same.

src/lua/loaders.lua
```diff
@@ -317,6 +317,27 @@ rawset(package, "search", search)
...
+_G.require = function(modname)
+    if no_package_loaded[modname] then
+        no_package_loaded[modname] = nil
+        package.loaded[modname] = nil
+    end
+    -- NB: This call is a tail call and it is important to some
+    -- extent. At least luajit's test suites expect certain error
+    -- messages from `require()` calls and from the `-l` option.
+    --
+    -- A non-tail call changes the filename at beginning of the
```
Buristan, line 336: Typo: s/at beginning/at the beginning/

Buristan (approved):
**[PATCH 2/2] lua: allow to refuse caching in package.loaded**
Hi, Alexander!
Thanks for the fixes!
LGTM, with a single nit regarding to the comment below.

src/lua/loaders.lua
Buristan: Yes, as you mentioned it changes the semantics a bit, but still OK for our usage.

src/lua/loaders.lua
Totktonada: Fixed.

Totktonada: (Resolved as trivial.)

igormunkin (approved):
**[PATCH 1/2] lua: add internal modules initialization status**

src/lua/loaders.lua
```diff
@@ -330,4 +351,15 @@ return {
     -- It is `true` during tarantool initialization, but once all
     -- the built-in modules are ready, will be set to `nil`.
     initializing = true,
+    raw_require = raw_require,
```
igormunkin, line 354: Hm, this made a bit inconsistent our structure with other modules overriding Lua stdlib functions. IDK, maybe all this should be moved to a separate `'internal.require'` module. Thoughts?

igormunkin (approved):
**[PATCH 2/2] lua: allow to refuse caching in package.loaded**

@Totktonada, thanks for the patch! LGTM, but please consider my comment regarding the modules structure.

src/lua/loaders.lua
Totktonada: `require()` is also wrapped in `src/lua/init.lua`. Sure, it worth to do. But this refactoring requires a bit more effort[^1], so I would pass it over for now. (This API is anyway internal and we can move this field to another internal module. I exposed `raw_require` just in case -- it should be accessible from somewhere IMHO, when we wrap it.)

[^1]: I have mentioned it in 432745263024641f45a0f0866a3c2691ef4b8a8b:

      > The init.lua file contains initialization code from different domains and it worth to split it further. I'll not do that in this series.

igormunkin: OK, that works for me. Feel free to add this to your TODO list or file an issue.

Totktonada: [Added](https://www.notion.so/tarantool/Move-require-wrappers-into-src-lua-require-lua-1bc8ae64107a420faeb7c56a5ff0d5cf).

PR #8349, opened 2023-02-18
[2.8] Hot fix Tarantool and LuaJIT

kyukhin:
This PR incorporates two hotfixes:
- protecting call to lj_gc_step on trace exit
- size calculation in functional index chunks


ylobankov:
It looks like all workflow failures are due to some infrastructure issues. The `2.8` branch is quite old, and our infrastructure has changed since the last `2.8.4` tag. IMHO, these failures can be ignored, and you can proceed with your patch.

kyukhin:
It seems like failing CI jobs are not relevant to the patches but to the changed CI infrastructure.

PR #8351, opened 2023-02-19
lua: panic on suspicious built-in modules manipulations

Totktonada:
The built-in module registration process performs several assertions:

- luaT_newmodule() checks for attempts to register an already registered module.
- luaT_setmodule() does the same, but it allows the same value be 'registered' several times.

Attempt to register different values as the same built-in module is definitely an error. Once a module is registered, it may be required and captured. If its value is changed afterwards, different parts of tarantool's code will capture different values for the same built-in module. It is very unlikely that such effect may be intended.

Anyway, tarantool's code doesn't do this. However, now, after implementing the override modules feature, it is possible to influence tarantool's initialization from a user provided code.

It means that we can't assume the situations described above as impossible ones. A developer of an override module should receive an error if the code of the module leads to such a situation.

`assert()` works only in Debug build, but most of user's code is tested on RelWithDebInfo builds. So incorrect override module may be implemented and distributed. It is undesirable.

Let's replace `assert()` with explicit `panic()`.

----

Adjusted testing helpers:

* `treegen` now allows to just write given script into given directory without using templates. Sometimes it is more convenient.
* `justrun` can return pure stdout (without JSONL decoding) and can return stderr if requested.

`nojson` option in `justrun` may look unneded for given test cases, but it provides better diagnostics if something going wrong and one of the test cases fails (say, if tarantool runs successfully instead of going to panic).

----

Follows up #7774

### DIFF
--- a/test/app-luatest/override_panic_test.lua
+++ b/test/app-luatest/override_panic_test.lua
@@ -1,0 +1,74 @@
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+
+local g = t.group()
+
+g.before_all(function(g)
+    treegen.init(g)
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+-- Trigger luaT_newmodule() panic.
+--
+-- fio is required internally before box initialization, so we can
+-- register fake box module from its override module. Then, when
+-- box starts to register itself, luaT_newmodule() finds it as
+-- already registered.
+--
+-- luaT_newmodule() must panic in the case.
+g.test_newmodule_panic = function(g)
+    local dir = treegen.prepare_directory(g, {}, {})
+    treegen.write_script(dir, 'override/fio.lua', [[
+        local loaders = require('internal.loaders')
+        loaders.builtin.box = {}
+        return loaders.builtin.fio
+    ]])
+    treegen.write_script(dir, 'main.lua', '')
+    local opts = {nojson = true, stderr = true}
+    local res = justrun.tarantool(dir, {}, {'main.lua'}, opts)
+    t.assert_equals(res, {
+        exit_code = 1,
+        stderr = 'luaT_newmodule(box): the module is already registered',
+    })
+end
+
+-- Trigger luaT_setmodule() panic.
+--
+-- src/lua/fio.c registers `fio` module.
+--
+-- Then src/lua/fio.lua is loaded. It is structured like so:
+--
+--  | local fio = require('fio') -- from fio.c
+--  |
+--  | function fio.foo(<...>)
+--  |     <...>
+--  | end
+--  |
+--  | return fio
+--
+-- fio.lua's return value is passed to luaT_setmodule() and it
+-- must be the same as one that is already registered by fio.c.
+--
+-- We can change what the require call returns (and so what
+-- fio.lua returns) using an override module.
+--
+-- luaT_setmodule() must panic if it meets attempt to register
+-- different values as the same built-in module.
+g.test_setmodule_panic = function(g)
+    local dir = treegen.prepare_directory(g, {}, {})
+    treegen.write_script(dir, 'override/fio.lua', [[
+        return {}
+    ]])
+    treegen.write_script(dir, 'main.lua', '')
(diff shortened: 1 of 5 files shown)


test/justrun.lua
```diff
@@ -40,23 +105,35 @@ function justrun.tarantool(dir, env, args)
     end
 
     local exit_code = ph:wait().exit_code
+    local stderr = join_stderr_fiber(stderr_fiber)
     ph:close()
 
     -- If an error occurs, discard the output and return only the
-    -- exit code.
+    -- exit code. However return stderr.
```
Buristan, line 112: Typo: s/However/However,/

test/treegen.lua
```diff
@@ -44,22 +44,31 @@ local function find_template(g, script)
...
+    fio.mktree(scriptdir_abspath)
 
+    log.info(('Writing a script: %s'):format(script_abspath))
+    local fh = fio.open(script_abspath, flags, mode)
+    fh:write(body)
     fh:close()
 end
 
+-- Generate a script that follows a template and write it at given
+-- script file path in given directory.
```
Buristan, line 64: Typo: s/in given directory/in the given directory/

Buristan (approved):
Hi, Alexander!
Thanks for the patch!
LGTM, just a few nits regarding the comments and the commit message.

> definitely an error. Once a module is registered it may be required and

Typo: s/Once a module is registered/Once a module is registered,/

> Anyway, tarantool's code don't do this. However, now, after implementing

Typo: s/don't/doesn't/

> * `justrun` can return pure stdout (without JSONL decoding) and can
  return stdout if requested.

I suppose you mean "and can return stderr if requested".

test/treegen.lua
Totktonada: Fixed (and all similar typos around). Thank you!

test/justrun.lua
Totktonada: Ouch. Fixed.

test/treegen.lua
```diff
@@ -44,22 +44,31 @@ local function find_template(g, script)
...
     local script_abspath = fio.pathjoin(dir, script)
     local flags = {'O_CREAT', 'O_WRONLY', 'O_TRUNC'}
     local mode = tonumber('644', 8)
-    local fh = fio.open(script_abspath, flags, mode)
-    local template = find_template(g, script)
-    local replacements = fun.chain({script = script}, replacements):tomap()
 
-    log.info(('Writing a script: %s'):format(script_abspath))
-    fh:write(template:gsub('<(.-)>', replacements))
+    local scriptdir_abspath = fio.dirname(script_abspath)
```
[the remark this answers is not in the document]
Totktonada: > Does it equal to `dir` argument?

It doesn't for, say, `override/fio.lua` relative script path.

> IOW, why do you need an absolute path here?

To don't change current directory forth and back. Also working with absolute paths is sometimes more convenient (less error prone).

test/app-luatest/override_panic_test.lua
```diff
@@ -0,0 +1,74 @@
...
+    local dir = treegen.prepare_directory(g, {}, {})
+    treegen.write_script(dir, 'override/fio.lua', [[
+        return {}
+    ]])
+    treegen.write_script(dir, 'main.lua', '')
+    local opts = {nojson = true, stderr = true}
+    local res = justrun.tarantool(dir, {}, {'main.lua'}, opts)
+    t.assert_equals(res, {
+        exit_code = 1,
+        stderr = 'luaT_setmodule(fio): the module is already registered ' ..
```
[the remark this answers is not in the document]
Totktonada: Hmm.

1. It is split exactly as in the source code. There is some consistency here (unimportant).
2. It is split at point, where the `luaT_newmodule` messsage above ends. There is some consistency as well (as well unimportant).
3. Carrying a half-empty line feels inaccurate for me.

I understand the point from (AFAIR) kernel's style guide regarding splitting error/log messages, but it is not followed even in the kernel in fact.

I would prefer to leave it as is.

test/treegen.lua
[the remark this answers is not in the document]
Totktonada: (Looks answered, so I'll resolve it.)

Issue #7356, opened 2022-07-04
Populate index with keydef


Mons:
[PM link](https://www.notion.so/tarantool/Populate-index-with-keydef-0f9ffacefc7144ed8750567935ef18ba)

**The problem**

Currently, extracting key part from a tuple is not so simple.
Also, there is a handy keydef module.

Currently user have to create keydef for the index
```lua
local keydef = require('tuple.keydef')
local ix = box.space[ ... ].index[ ... ]
local kd = pcall(keydef.new, ix.parts)
```
And then use keydef to extract key from tuple
```lua
local key = kd:extract_key( tuple )
```

(And also keydef is not working with rtree)

**The proposition**

I propose to extend index object with it's keydef and then use it like
```
local key = box.space[ ... ].index[ ... ].keydef:extract_key( tuple )
```

olegrok:
Duplicate of https://github.com/tarantool/tarantool/issues/4955

Totktonada:
There are two primary applications of key_def objects: comparing tuples (usually for ordering) and extracting a key from a tuple (usually for pagination). Different key_def objects are used for those goals if the index is non-unique:

```lua
-- Sort.
--
-- Input: tuples.
local index = box.space.s.index.secondary
local kd = key_def.new(index.parts)
if not index.unique then
    local primary_index = box.space.s.index[0]
    kd = key_def:merge(key_def.new(primary_index.parts))
end
table.sort(tuples, function(a, b)
    return kd:compare(a, b) < 0
end)
-- Output: tuples.
```

```lua
-- Extract a key.
--
-- Input: tuple.
local index = box.space.s.index.secondary
local kd = key_def.new(index.parts)
local key = kd:extract_key(tuple)
-- Output: key.
```

My doubt is that it may be not so obvious that the key_def object provided by a non-unique index will not impose the same ordering as the index itself.

OTOH, it is applicable to the manually constructed `key_def.new(index.parts)` as well. So my doubt is not an objection.

Gumix:
> I propose to extend index object with it's keydef and then use it like
> local key = box.space[ ... ].index[ ... ].keydef:extract_key( tuple )

So the index object will contain 2 similar fields:
* `parts` — a table without methods (current implementation);
* `key_def` — the new cdata field, which behaves like the object created by
  ```lua
  require('key_def').new(box.space.T.index.I.parts)
  ```

E.g.:
```lua
tarantool> box.space.T.index.I
---
- parts:
  - type: string
[...]

PR #8356, opened 2023-02-19
box: populate `index_object.parts` with `key_def` module methods

Gumix:
**box: populate index_object.parts with key_def module methods**

This patch adds 4 methods of the `key_def` module instance to the
`index_object.parts`, see the docbot request for details. The rest
methods (`new()` and `totable()`) are not applicable here, because the
instance is already created and pushed to the stack as a table.

@TarantoolBot document
Title: `index_object.parts` methods
Product: Tarantool
Since: 3.0
Root document: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/parts/

`index_object.parts` has the following methods: `extract_key()`, `compare()`, `compare_with_key()`, `merge()`.
For their description and usage examples, refer to [Module key_def](https://www.tarantool.io/en/doc/latest/reference/reference_lua/key_def/#lua-object.key_def.key_def_object).

Root document: https://www.tarantool.io/en/doc/latest/reference/reference_lua/key_def/

`index_object.parts` can be used as a key_def module instance for calling the corresponding methods.

Example:

```lua
box.schema.space.create('T')
i = box.space.T:create_index('I', {parts={3, 'string', 1, 'unsigned'}})
box.space.T:insert{1, 99.5, 'X', nil, 99.5}
i.parts:extract_key(box.space.T:get({'X', 1}))
```

The code above is equivalent to:

```lua
key_def = require('key_def')
box.schema.space.create('T')
i = box.space.T:create_index('I', {parts={3, 'string', 1, 'unsigned'}})
box.space.T:insert{1, 99.5, 'X', nil, 99.5}
k = key_def.new(i.parts)
k:extract_key(box.space.T:get({'X', 1}))
```

Closes https://github.com/tarantool/tarantool/issues/7356

### DIFF
--- a/src/box/lua/key_def.c
+++ b/src/box/lua/key_def.c
@@ -40,8 +40,39 @@
 
 static uint32_t CTID_STRUCT_KEY_DEF_REF = 0;
 
+/**
+ * Free a key_def from a Lua code.
+ */
+static int
+lbox_key_def_gc(struct lua_State *L)
+{
+	struct key_def *key_def = luaT_is_key_def(L, 1);
+	assert(key_def != NULL);
+	key_def_delete(key_def);
+	return 0;
+}
+
+/**
+ * Push key_def as a cdata object to a Lua stack. This function takes ownership
+ * of key_def, and sets finalizer lbox_key_def_gc for it.
+ */
+static void
+luaT_push_key_def_nodup(struct lua_State *L, const struct key_def *key_def)
+{
+	void *ptr = luaL_pushcdata(L, CTID_STRUCT_KEY_DEF_REF);
+	*(const struct key_def **)ptr = key_def;
+	lua_pushcfunction(L, lbox_key_def_gc);
+	luaL_setcdatagc(L, -2);
+}
+
 void
 luaT_push_key_def(struct lua_State *L, const struct key_def *key_def)
+{
+	luaT_push_key_def_nodup(L, key_def_dup(key_def));
+}
+
+void
+luaT_push_key_def_parts(struct lua_State *L, const struct key_def *key_def)
 {
 	lua_createtable(L, key_def->part_count, 0);
 	for (uint32_t i = 0; i < key_def->part_count; ++i) {
@@ -242,7 +273,7 @@ luaT_key_def_check_tuple(struct lua_State *L, struct key_def *key_def, int idx)
 }
 
 struct key_def *
-luaT_check_key_def(struct lua_State *L, int idx)
+luaT_is_key_def(struct lua_State *L, int idx)
 {
 	if (lua_type(L, idx) != LUA_TCDATA)
 		return NULL;
@@ -254,33 +285,16 @@ luaT_check_key_def(struct lua_State *L, int idx)
 	return *key_def_ptr;
 }
 
-/**
- * Free a key_def from a Lua code.
- */
-static int
-lbox_key_def_gc(struct lua_State *L)
+int
+luaT_key_def_extract_key(struct lua_State *L, int idx)
 {
-	struct key_def *key_def = luaT_check_key_def(L, 1);
+	struct key_def *key_def = luaT_is_key_def(L, idx);
 	assert(key_def != NULL);
-	key_def_delete(key_def);
-	return 0;
-}
 
-/**
- * Extract key from tuple by given key definition and return
- * tuple representing this key.
- * Push the new key tuple as cdata to a LUA stack on success.
- * Raise error otherwise.
- */
-static int
-lbox_key_def_extract_key(struct lua_State *L)
-{
-	struct key_def *key_def;
-	if (lua_gettop(L) != 2 || (key_def = luaT_check_key_def(L, 1)) == NULL)
-		return luaL_error(L, "Usage: key_def:extract_key(tuple)");
-
-	struct tuple *tuple;
-	if ((tuple = luaT_key_def_check_tuple(L, key_def, 2)) == NULL)
+	if (key_def->is_multikey)
+		return luaL_error(L, "multikey path is unsupported");
+	struct tuple *tuple = luaT_key_def_check_tuple(L, key_def, -1);
+	if (tuple == NULL)
 		return luaT_error(L);
 
 	struct region *region = &fiber()->gc;
@@ -300,24 +314,14 @@ lbox_key_def_extract_key(struct lua_State *L)
 	return 1;
 }
 
-/**
- * Compare tuples using the key definition.
- * Push 0  if key_fields(tuple_a) == key_fields(tuple_b)
- *      <0 if key_fields(tuple_a) < key_fields(tuple_b)
- *      >0 if key_fields(tuple_a) > key_fields(tuple_b)
- * integer to a LUA stack on success.
- * Raise error otherwise.
- */
-static int
-lbox_key_def_compare(struct lua_State *L)
+int
+luaT_key_def_compare(struct lua_State *L, int idx)
 {
-	struct key_def *key_def;
-	if (lua_gettop(L) != 3 ||
-	    (key_def = luaT_check_key_def(L, 1)) == NULL) {
-		return luaL_error(L, "Usage: key_def:"
-				     "compare(tuple_a, tuple_b)");
-	}
+	struct key_def *key_def = luaT_is_key_def(L, idx);
+	assert(key_def != NULL);
 
+	if (key_def->is_multikey)
+		return luaL_error(L, "multikey path is unsupported");
 	if (key_def->tuple_compare == NULL) {
 		enum field_type type = key_def_incomparable_type(key_def);
 		assert(type != field_type_MAX);
@@ -326,10 +330,11 @@ lbox_key_def_compare(struct lua_State *L)
 		return luaT_error(L);
 	}
 
-	struct tuple *tuple_a, *tuple_b;
-	if ((tuple_a = luaT_key_def_check_tuple(L, key_def, 2)) == NULL)
+	struct tuple *tuple_a = luaT_key_def_check_tuple(L, key_def, -2);
+	if (tuple_a == NULL)
 		return luaT_error(L);
-	if ((tuple_b = luaT_key_def_check_tuple(L, key_def, 3)) == NULL) {
+	struct tuple *tuple_b = luaT_key_def_check_tuple(L, key_def, -1);
+	if (tuple_b == NULL) {
 		tuple_unref(tuple_a);
 		return luaT_error(L);
 	}
@@ -341,24 +346,14 @@ lbox_key_def_compare(struct lua_State *L)
 	return 1;
 }
 
-/**
- * Compare tuple with key using the key definition.
- * Push 0  if key_fields(tuple) == parts(key)
- *      <0 if key_fields(tuple) < parts(key)
- *      >0 if key_fields(tuple) > parts(key)
- * integer to a LUA stack on success.
- * Raise error otherwise.
- */
-static int
-lbox_key_def_compare_with_key(struct lua_State *L)
+int
+luaT_key_def_compare_with_key(struct lua_State *L, int idx)
 {
-	struct key_def *key_def;
-	if (lua_gettop(L) != 3 ||
-	    (key_def = luaT_check_key_def(L, 1)) == NULL) {
-		return luaL_error(L, "Usage: key_def:"
-				     "compare_with_key(tuple, key)");
-	}
+	struct key_def *key_def = luaT_is_key_def(L, idx);
+	assert(key_def != NULL);
 
+	if (key_def->is_multikey)
+		return luaL_error(L, "multikey path is unsupported");
 	if (key_def->tuple_compare_with_key == NULL) {
 		enum field_type type = key_def_incomparable_type(key_def);
 		assert(type != field_type_MAX);
@@ -367,13 +362,13 @@ lbox_key_def_compare_with_key(struct lua_State *L)
 		return luaT_error(L);
 	}
 
-	struct tuple *tuple = luaT_key_def_check_tuple(L, key_def, 2);
+	struct tuple *tuple = luaT_key_def_check_tuple(L, key_def, -2);
 	if (tuple == NULL)
 		return luaT_error(L);
 
 	struct region *region = &fiber()->gc;
 	size_t region_svp = region_used(region);
-	const char *key = luaT_tuple_encode(L, 3, NULL);
+	const char *key = luaT_tuple_encode(L, -1, NULL);
 	if (key == NULL || box_key_def_validate_key(key_def, key, NULL) != 0) {
 		region_truncate(region, region_svp);
 		tuple_unref(tuple);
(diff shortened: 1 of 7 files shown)


src/box/lua/key_def.c
```diff
@@ -273,14 +273,14 @@ lbox_key_def_gc(struct lua_State *L)
  * Raise error otherwise.
  */
 static int
-lbox_key_def_extract_key(struct lua_State *L)
+lbox_key_def_extract_key(struct lua_State *L, int idx, const char *obj_name)
```
Totktonada, line 239: It is not strictly followed across the code, but AFAIR there is a convention that `lbox_` prefixed functions adhere the [lua_CFunction](https://www.lua.org/manual/5.1/manual.html#lua_CFunction) prototype (so they can be passed to [lua_pushcfunction](https://www.lua.org/manual/5.1/manual.html#lua_pushcfunction) and exposed to the Lua land). Other functions that need the Lua state are usually prefixed with `luaT_` and they can accept more than just `struct lua_State *`.

I don't like `box` in this `lbox_`, to be honest, but such formalism is better than lack of any style rules or mixing different ones.

@igormunkin, can you, please, approve or deny this statement?

src/box/lua/key_def.c
```diff
@@ -397,13 +415,14 @@ lbox_key_def_compare_with_key(struct lua_State *L)
...
-	if (lua_gettop(L) != 2 ||
-	    (key_def_a = luaT_check_key_def(L, 1)) == NULL ||
-	    (key_def_b = luaT_check_key_def(L, 2)) == NULL)
-		return luaL_error(L, "Usage: key_def:merge(second_key_def)");
+	struct key_def *key_def_a = luaT_check_key_def(L, idx);
+	struct key_def *key_def_b = luaT_check_key_def(L, 2);
+	if (lua_gettop(L) != 2 || key_def_a == NULL || key_def_b == NULL) {
+		return luaL_error(L, "Usage: %s:merge(second_key_def)",
+				  obj_name);
+	}
```
Totktonada, line 407: It looks lopsided:

1. `key_def_a` is at `idx`, OK.
2. `key_def_b` is at `2`, hm.
3. The stack top is strictly `2`.

So it looks like the function is to be called only from Lua (I mean, wrapped and registered with lua_pushcfunction()), otherwise the stack top condition is strange. And so it is not clear why `key_def_a` is not always at slot `1`.

Totktonada: The question about the stack top checks is applicable to other functions as well.

src/box/lua/space.cc
```diff
@@ -453,7 +453,6 @@ lbox_fillspace(struct lua_State *L, struct space *space, int i)
 
 		lua_pushstring(L, "parts");
 		luaT_push_key_def(L, index_def->key_def);
-
```
Totktonada, line 456: Nit: It is the only change in the file :)

src/box/lua/key_def.c
Gumix: Now all these methods can be called in two ways:

1.
```lua
key_def = require('key_def')
k = key_def.new(idx.parts)
k:merge(second_key_def)
```
Here self is key_def_a (slot 1), key_def_b is at 2.

2.
```lua
idx.parts:merge(second_key_def)
```
Here unused self is at slot 1, key_def_b is at 2, key_def_a is in the upvalue.

src/box/lua/space.cc
Gumix: Yes, that empty line was a bit misleading.

Totktonada: Hm. OK.

src/box/lua/key_def.c
Totktonada: If we'll go with this implementation, it should be commented somehow in the code, because it is not obvious at reading. It is a part of the function contract in fact.

Gumix: Added a comment.

olegrok:
```lua
box.space.test:create_index('mk', {parts = {{type = 'string', path = '[*]', field = 2}}})
box.space.test:replace({1, {"a", "b"}})
tuple = box.space.test.index.mk:select()[1]

tarantool> box.space.test.index[0].parts:extract_key(tuple) -- Ok
---
- [1]
...

tarantool> box.space.test.index[1].parts:extract_key(tuple) -- Not Ok
Assertion failed: (!key_def->is_multikey), function tuple_validate_key_parts, file tuple_extract_key.cc, line 513.
[1]    10902 abort      ./src/tarantool
```

Totktonada:
While thinking on it I tried to create the captures in Lua to reduce a complexity of the code. It seems to work.

<details>
<summary>Patch (collapsed).</summary>

```diff
diff --git a/src/box/lua/key_def.c b/src/box/lua/key_def.c
index e5fc370e49..6245bba005 100644
--- a/src/box/lua/key_def.c
+++ b/src/box/lua/key_def.c
@@ -40,6 +40,9 @@
 
 static uint32_t CTID_STRUCT_KEY_DEF_REF = 0;
 
+static int
+lbox_key_def_gc(struct lua_State *L);
+
 void
 luaT_push_key_def(struct lua_State *L, const struct key_def *key_def)
 {
@@ -75,6 +78,35 @@ luaT_push_key_def(struct lua_State *L, const struct key_def *key_def)
 		}
 		lua_rawseti(L, -2, i + 1);
 	}
+
+	luaL_loadstring(L, "do"
+			"    local key_def = ..."
+			"    return {"
+			"        __index = {"
+			"            extract_key = function(_self, ...)"
+			"                return key_def:extract_key(...)"
+			"            end,"
+			"            compare = function(_self, ...)"
+			"                return key_def:compare(...)"
+			"            end,"
+			"            compare_with_key = function(_self, ...)"
+			"                return key_def:compare_with_key(...)"
+			"            end,"
+			"            merge = function(_self, ...)"
+			"                return key_def:merge(...)"
+			"            end,"
+			"        }"
+			"    }"
+			"end");
+
+	*(struct key_def **) luaL_pushcdata(L,
+				CTID_STRUCT_KEY_DEF_REF) = key_def_dup(key_def);
+	lua_pushcfunction(L, lbox_key_def_gc);
+	luaL_setcdatagc(L, -2);
+
+	lua_call(L, 1, 1);
+
+	lua_setmetatable(L, -2);
 }
 
 /**
```
</details>

NB: We can do some optimizations here: at least do `luaL_loadstring()` once, save a reference in the Lua registry and reuse the function.

NB: `key_def_dup()` is because keydef's are only dopying and not refcounted. We have no control over the original keydef lifetime, so we should copy it and free when unneeded.

This code passes the test except error messages:

<details>
<summary>Patch on the test (collapsed).</summary>

```diff
--- a/test/box-luatest/gh_7356_index_parts_methods_test.lua	2023-02-28 19:50:56.320584381 +0300
+++ b/test/box-luatest/gh_7356_index_parts_methods_test.lua	2023-02-28 19:51:04.320584448 +0300
@@ -29,10 +29,10 @@
         t.assert_equals(k, {'X', 1})
 
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:extract_key(tuple)',
+            'Usage: key_def:extract_key(tuple)',
             sk.parts.extract_key)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:extract_key(tuple)',
+            'Usage: key_def:extract_key(tuple)',
             sk.parts.extract_key, sk.parts)
         t.assert_error_msg_content_equals(
             'A tuple or a table expected, got number',
@@ -60,10 +60,10 @@
         t.assert_equals(i.parts.compare(nil, tuple_c, tuple_a), 1)
 
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:compare(tuple_a, tuple_b)',
+            'Usage: key_def:compare(tuple_a, tuple_b)',
             i.parts.compare)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:compare(tuple_a, tuple_b)',
+            'Usage: key_def:compare(tuple_a, tuple_b)',
             i.parts.compare, i.parts, tuple_a)
         t.assert_error_msg_content_equals(
             'A tuple or a table expected, got cdata',
@@ -89,13 +89,13 @@
         t.assert_equals(i.parts:compare_with_key(tuple, {'X', 2}), -1)
 
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:compare_with_key(tuple, key)',
+            'Usage: key_def:compare_with_key(tuple, key)',
             i.parts.compare_with_key)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:compare_with_key(tuple, key)',
+            'Usage: key_def:compare_with_key(tuple, key)',
             i.parts.compare_with_key, box.NULL)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:compare_with_key(tuple, key)',
+            'Usage: key_def:compare_with_key(tuple, key)',
             i.parts.compare_with_key, box.NULL, {0, nil, ''})
         t.assert_error_msg_content_equals(
             'Supplied key type of part 1 does not match index part type: ' ..
@@ -120,13 +120,13 @@
         t.assert_equals(i.parts:merge(kd2):totable(), exp)
 
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:merge(second_key_def)',
+            'Usage: key_def:merge(second_key_def)',
             i.parts.merge)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:merge(second_key_def)',
+            'Usage: key_def:merge(second_key_def)',
             i.parts.merge, i.parts)
         t.assert_error_msg_content_equals(
-            'Usage: index.parts:merge(second_key_def)',
+            'Usage: key_def:merge(second_key_def)',
             i.parts.merge, i.parts, kd2, box.NULL)
     end)
 end
```
</details>

It is not proposal to do it this way, just some stuff to think around.

Another possible way is to just add a field `_key_def` and use it in the functions instead of a capture. We'll need to hide it from serialization using `__serialize` (just like we do in src/lua/init.lua). And it is not so clean as captures. But it may be also acceptable.

src/box/lua/key_def.c
igormunkin: I don't know when `lbox_` prefix should be used. What about `luaT_` prefix: IIRC `T` is for Tarantool specifics, but these functions can have lua_CFunction signature or any other.

Gumix: Renamed

src/box/lua/space.cc
```diff
@@ -279,6 +279,120 @@ lbox_push_space_foreign_key(struct lua_State *L, struct space *space, int i)
...
+/**
+ * index.parts:merge(second_key_def)
+ * Stack: [1] unused; [2] second_key_def.
+ * key_def is passed in the upvalue.
+ */
+static int
+lbox_key_def_merge(struct lua_State *L)
+{
+	struct key_def *key_def_a = luaT_check_key_def(L, lua_upvalueindex(1));
+	struct key_def *key_def_b = luaT_check_key_def(L, 2);
```
Totktonada, line 343: Shouldn't it accept another index parts here?

test/box-luatest/gh_7356_index_parts_methods_test.lua
```diff
@@ -0,0 +1,148 @@
+local t = require('luatest')
```
Totktonada, line 1: I would add one `:alter()` case to verify that the methods are actually updated. Say, we can add a field into existing index.

src/box/lua/key_def.c
Totktonada: Irrelevant now. Closing.

src/box/lua/space.cc
```diff
@@ -279,6 +279,120 @@ lbox_push_space_foreign_key(struct lua_State *L, struct space *space, int i)
...
+	lua_newtable(L);
+	/* __index */
+	lua_createtable(L, 1, 0);
+	int idx_index = lua_gettop(L);
+
+	/* Push 4 references to cdata onto the stack, one for each closure. */
+	void *ptr = luaL_pushcdata(L, CTID_STRUCT_KEY_DEF_REF);
+	*(struct key_def **)ptr = key_def_dup(key_def);
+	lua_pushcfunction(L, lbox_key_def_gc);
+	luaL_setcdatagc(L, -2);
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/lua/space.cc
```diff
@@ -279,6 +279,120 @@ lbox_push_space_foreign_key(struct lua_State *L, struct space *space, int i)
...
+static void
+luaT_add_index_parts_methods(struct lua_State *L, const struct key_def *key_def)
+{
+	luaL_cdef(L, "struct key_def;");
+	uint32_t CTID_STRUCT_KEY_DEF_REF = luaL_ctypeid(L, "struct key_def&");
+
+	/* Metatable. */
+	lua_newtable(L);
+	/* __index */
+	lua_createtable(L, 1, 0);
```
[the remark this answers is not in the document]
Gumix: I have no idea why I wrote (1, 0). Changed to `lua_newtable()`.

src/box/lua/space.cc
```diff
@@ -279,6 +279,120 @@ lbox_push_space_foreign_key(struct lua_State *L, struct space *space, int i)
 	lua_setfield(L, i, "foreign_key");
 }
 
+/**
+ * index.parts:extract_key(tuple)
+ * Stack: [1] unused; [2] tuple.
+ * key_def is passed in the upvalue.
+ */
+static int
+lbox_key_def_extract_key(struct lua_State *L)
```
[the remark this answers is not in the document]
Gumix: Renamed.

src/box/lua/space.cc
```diff
@@ -279,6 +279,120 @@ lbox_push_space_foreign_key(struct lua_State *L, struct space *space, int i)
...
+static int
+lbox_key_def_compare(struct lua_State *L)
+{
+	struct key_def *key_def = luaT_check_key_def(L, lua_upvalueindex(1));
+	if (lua_gettop(L) != 3 || key_def == NULL) {
+		return luaL_error(L, "Usage: index.parts:compare("
+				     "tuple_a, tuple_b)");
+	}
+	if (key_def->is_multikey)
+		return luaL_error(L, "multikey path is unsupported");
```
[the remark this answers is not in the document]
Gumix: Moved the check into `luaT_key_def_extract_key()` and others.

src/box/lua/space.cc
Gumix: OK, done.

test/box-luatest/gh_7356_index_parts_methods_test.lua
Gumix: Done.

src/box/lua/key_def.h
```diff
@@ -56,6 +63,71 @@ luaT_push_key_def(struct lua_State *L, const struct key_def *key_def);
 struct key_def *
 luaT_check_key_def(struct lua_State *L, int idx);
 
+/**
+ * Free a key_def from a Lua code.
+ */
+int
+lbox_key_def_gc(struct lua_State *L);
```
Totktonada, line 70: AFAIU we can make it `static` again (and drop the declaration from the header).

Gumix: Yes, done.

p7nov (approved):

Totktonada (approved):

src/box/lua/key_def.h
```diff
@@ -56,6 +63,65 @@ luaT_push_key_def(struct lua_State *L, const struct key_def *key_def);
...
+ * Compare tuples using the key definition.
+ * Push 0  if key_fields(tuple_a) == key_fields(tuple_b)
+ *      <0 if key_fields(tuple_a) < key_fields(tuple_b)
+ *      >0 if key_fields(tuple_a) > key_fields(tuple_b)
+ * integer to a LUA stack on success.
+ * Raise error otherwise.
+ */
+int
+luaT_key_def_compare(struct lua_State *L, struct key_def *key_def, int idx_a,
+		     int idx_b);
```
locker, line 86: The signautre of this and other functions added by this patch look weird, not Lua-style, because some arguments are passed as C pointers while others are passed as indexes in Lua stack. Wouldn't it be better to pass the `key_def` via the Lua stack as well? Also, having that many indexes is confusing IMO. I'd change the signature of this function to

```C
int
luaT_key_def_compare(struct lua_State *L, int idx);
```

and make it compare the top two values at the Lua stack with the `key_def` located at `idx`. This would be consistent with `lua_settable`, for example. If the values that needs to be compared are located not at the top of the Lua stack, one can push them to the top with `lua_pushvalue`.

src/box/lua/key_def.h
```diff
@@ -38,6 +38,13 @@ extern "C" {
 struct lua_State;
 struct key_def;
 
+/**
+ * Push key_def as a cdata object to a Lua stack, and set finalizer function
+ * lbox_key_def_gc for it.
+ */
+void
+luaT_push_key_def(struct lua_State *L, const struct key_def *key_def);
```
locker, line 46: It may be unclear that, despite the `const` specifier, this function actually takes the ownership of the given `key_def` object. This makes it quite different from other similar functions like `luaT_pusherror` or `luaT_pushtuple`. Maybe, better call `key_def_dup` on the argument under the hood?

src/box/lua/key_def.h
Totktonada: NB: `lua_settable()` pop two elements from a *top of the stack*. I don't remember any Lua/C API function that has access to `idx+1` or `idx-1` element (relative to the passed `idx`).

src/box/lua/key_def.h
Totktonada: This point looks valid for me too. I overlooked it.

Totktonada: OTOH we use it internally to push just created key_def object (and it seems logical to have such a function). Hmm.

src/box/lua/key_def.h
Gumix: > and make it compare the top two values at the Lua stack with the key_def located at idx.

This is exactly what the initial patch looked like: https://github.com/tarantool/tarantool/pull/8356#discussion_r1111338354

src/box/lua/key_def.h
Gumix: If it's clearer, I can put
```c
    *(struct key_def **) luaL_pushcdata(L, CTID_STRUCT_KEY_DEF_REF) = new_key_def;
    lua_pushcfunction(L, lbox_key_def_gc);
    luaL_setcdatagc(L, -2);
```
back to `luaT_key_def_merge()` and `lbox_key_def_new()`, instead of calling `luaT_push_key_def()` there,
and use `luaT_push_key_def()` with embedded `key_def_dup()` solely in `luaT_add_index_parts_methods()`.

src/box/lua/key_def.h
locker: Discussed f2f. Agreed to pass arguments in the Lua way, as suggested in the first comment of this thread.

Also, agreed that we need to rename `luaT_check_key_def` to `luaT_is_key_def`, because `check` functions usually throw on error.

src/box/lua/key_def.h
locker: Agreed to duplicate the `key_def` argument in `luaT_push_key_def` and introduce the static function `luaT_key_def_nodup` that would take ownership of `key_def` for use in the module implementation.

src/box/lua/key_def.h
Gumix: Done

src/box/lua/key_def.h
Gumix: Done

locker (approved):

PR #8357, opened 2023-02-20
SQL refactoring

ImeevMA:
Minor refactoring of the SQL code.

### DIFF
--- a/src/box/sql/prepare.c
+++ b/src/box/sql/prepare.c
@@ -41,7 +41,7 @@
 
 int
 sql_stmt_compile(const char *zSql, int nBytes, struct Vdbe *pReprepare,
-		 sql_stmt **ppStmt, const char **pzTail)
+		 struct Vdbe **ppStmt, const char **pzTail)
 {
 	int rc = 0;	/* Result code */
 	Parse sParse;		/* Parsing context */
@@ -147,7 +147,7 @@ sql_stmt_compile(const char *zSql, int nBytes, struct Vdbe *pReprepare,
 		sqlVdbeFinalize(sParse.pVdbe);
 		assert(!(*ppStmt));
 	} else {
-		*ppStmt = (sql_stmt *) sParse.pVdbe;
+		*ppStmt = sParse.pVdbe;
 	}
 
 	/* Delete any TriggerPrg structures allocated while parsing this statement. */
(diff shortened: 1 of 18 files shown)


Gerold103 (approved):
Thanks for the patch!

src/box/sql/prepare.c
```diff
@@ -163,26 +163,23 @@ sql_stmt_compile(const char *zSql, int nBytes, struct Vdbe *pReprepare,
 	return rc;
 }
 
-/*
- * Rerun the compilation of a statement after a schema change.
- */
 int
-sqlReprepare(Vdbe * p)
+sqlReprepare(struct Vdbe *v)
```
locker, line 167: Why do you rename `p` to `v`? It bloats the diff for no reason. The commit message says nothing about this change.

src/box/sql/vdbe.c
```diff
@@ -4460,7 +4457,6 @@ default: {          /* This is really OP_Noop and OP_Explain */
 
 	/* This is the only way out of this procedure. */
 vdbe_return:
-	p->aCounter[SQL_STMTSTATUS_VM_STEP] += (int)nVmStep;
```
locker, line 4463: Looks unrelated.

src/box/sql/prepare.c
ImeevMA: Should I add this to the description? In this patch, I changed the use of the `struct Vdbe`, so I decided to rename variables in some places from `p` (it seems to be short for `pointer`) to the more specialized `v` (short for `vdbe`).

src/box/sql/vdbe.c
ImeevMA: This field is used only in the `sql_stmt_status()` function, which I decided to remove instead of refactoring, since it is not used anywhere.

locker: It'd be better to drop `sql_stmt_status` in a separate commit then, because this change is unrelated to dropping the `sql_stmt` alias.

src/box/sql/prepare.c
locker: I don't think you should change local variable names, because `p` sounds fine to me as a local variable of type `Vdbe`. Moreover, even after this patch is applied, it's still used (e.g. in `sqlVdbeExec`). That is, this change doesn't make the code look any better - it only bloats the diff and makes it more difficult to review.

ImeevMA: Fixed. Moved all dropped functions in the new commit.

Also, I will try to stick to the "one change = one commit" rule in the future, even with refactoring.

src/box/sql/vdbe.c
ImeevMA: Fixed. Moved all dropped functions in the new commit.

Also, I will try to stick to the "one change = one commit" rule in the future, even with refactoring.

locker (approved):

PR #8358, opened 2023-02-20
net_box: don't use integer implementation defined

nshy:
When calculating `rows_index` in `netbox_decode_execute` etc we use implementation defined behavior of compiler. `i < mapsize` and this is fine according to standard as unsigned arithmetic is well defined on overflows. But then we cast the result to int and this is implementation defined as the result can not be represented by int.

Closes https://github.com/tarantool/security/issues/98

NO_DOC=minor
NO_TEST=minor
NO_CHANGELOG=minor

### DIFF
--- a/src/box/lua/net_box.c
+++ b/src/box/lua/net_box.c
@@ -1695,16 +1695,16 @@ netbox_decode_execute(struct lua_State *L, const char **data,
 				netbox_decode_data(L, data,
 						   tuple_format_runtime);
 			}
-			rows_index = i - map_size;
+			rows_index = lua_gettop(L);
 			break;
 		case IPROTO_METADATA:
 			netbox_decode_metadata(L, data);
-			meta_index = i - map_size;
+			meta_index = lua_gettop(L);
 			break;
 		default:
 			assert(key == IPROTO_SQL_INFO);
 			netbox_decode_sql_info(L, data);
-			info_index = i - map_size;
+			info_index = lua_gettop(L);
 			break;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/box/lua/net_box.c
```diff
@@ -1681,9 +1681,11 @@ netbox_decode_execute(struct lua_State *L, const char **data,
 	(void)data_end;
 	(void)format;
 	assert(mp_typeof(**data) == MP_MAP);
-	uint32_t map_size = mp_decode_map(data);
+	uint32_t map_size_u = mp_decode_map(data);
+	assert(map_size_u <= INT_MAX);
+	int map_size = map_size_u;
 	int rows_index = 0, meta_index = 0, info_index = 0;
-	for (uint32_t i = 0; i < map_size; ++i) {
+	for (int i = 0; i < map_size; ++i) {
```
locker, line 1688: This patch doesn't eliminate overflow. I think we can get rid of subtraction altogether with something like this:

```diff
diff --git a/src/box/lua/net_box.c b/src/box/lua/net_box.c
index 5075894fe5ef..efb25138ac24 100644
--- a/src/box/lua/net_box.c
+++ b/src/box/lua/net_box.c
@@ -1699,12 +1699,12 @@ netbox_decode_execute(struct lua_State *L, const char **data,
 			break;
 		case IPROTO_METADATA:
 			netbox_decode_metadata(L, data);
-			meta_index = i - map_size;
+			meta_index = lua_gettop(L);
 			break;
 		default:
 			assert(key == IPROTO_SQL_INFO);
 			netbox_decode_sql_info(L, data);
-			info_index = i - map_size;
+			info_index = lua_gettop(L);
 			break;
 		}
 	}
@@ -1712,9 +1712,9 @@ netbox_decode_execute(struct lua_State *L, const char **data,
 		assert(meta_index != 0);
 		assert(rows_index != 0);
 		lua_createtable(L, 0, 2);
-		lua_pushvalue(L, meta_index - 1);
+		lua_pushvalue(L, meta_index);
 		lua_setfield(L, -2, "metadata");
-		lua_pushvalue(L, rows_index - 1);
+		lua_pushvalue(L, rows_index);
 		lua_setfield(L, -2, "rows");
 	} else {
 		assert(meta_index == 0);
```

nshy: I guess patch fixes the implementation defined issue yet I like the approach with positive stack indexes.

locker (approved):

locker:

PR #8364, opened 2023-02-21
test: fix result-file for box-py/args.test.py

ImeevMA:
After the introduction of the 3.0.0-entrypoint, the test result-file must be updated.


igormunkin (approved):
@ImeevMA, thanks for the patch! LGTM, but I'm not quite familiar with our release procedure, so @kyukhin, please proceed with the patch.

kyukhin (approved):

PR #8366, opened 2023-02-21
sql: check column name in index definition

### LINKED ISSUE #8365: Assertion if function is given instead of column name in index definition
Error in debug only:
```
tarantool> box.execute([[create table t2(i int, a int, primary key(i, a), unique(abs(i), a));]])
tarantool: ./src/box/sql/resolve.c:670: resolveExprStep: Assertion `!func->def->is_deterministic || (pNC->ncFlags & NC_IdxExpr) == 0' failed.
Aborted (core dumped)
```

ImeevMA:
After this patch, the column name will be checked for being an ID before
resolution, not after.

Closes https://github.com/tarantool/tarantool/issues/8365

NO_DOC=bugfix in debug
NO_CHANGELOG=bugfix in debug

### DIFF
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -554,67 +554,24 @@ field_def_create_for_pk(struct Parse *parser, struct field_def *field,
  * A table can have at most one primary key.  If the table already has
  * a primary key (and this is the second primary key) then create an
  * error.
- *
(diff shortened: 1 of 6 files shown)


Gerold103 (approved):
Thanks for the patch!

locker (approved):

PR #8377, opened 2023-02-27
box: make `space_foreach()` strictly ordered

### LINKED ISSUE #7954: Sort spaces in snapshot by ID
Now a snapshot is generated in the following was: firstly all spaces with id < 512, ordered by ID, then all other spaces in unspecified order.

Let's specify the order and place all spaces in order of their IDs.

Gumix:
This also changes the order in which spaces are stored in `.snap` files.

Currently `space_foreach()` iterates over spaces in the following order:
1. Spaces with id in the range [256..510], ordered by id;
2. All other (non-system) spaces in unspecified order.

This patch changes the order to:
1. Spaces with id in the range [256..511], ordered by id;
2. All other (non-system) spaces, ordered by id.

Strict ordered of non-system spaces is required by users who iterate over `.snap` files with `xlog.pairs()`.

Closes #7954

NO_DOC=[[1]](https://www.tarantool.io/en/doc/latest/dev_guide/internals/file_formats/#the-snapshot-file-format) already contains: "the .snap file’s records are ordered by space id"

### DIFF
--- a/changelogs/unreleased/gh-7954-sort-spaces-in-snapshot-by-id.md
+++ b/changelogs/unreleased/gh-7954-sort-spaces-in-snapshot-by-id.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Now non-system spaces in a snapshot file are sorted by their id. All
+  non-system spaces are stored after system spaces, as before (gh-7954).
--- a/src/box/space.c
+++ b/src/box/space.c
@@ -431,6 +431,67 @@ space_delete(struct space *space)
 	space->vtab->destroy(space);
 }
 
+/**
+ * Call a visitor function for spaces with id in the range [id_min..id_max].
+ */
+static int
+space_foreach_helper(uint32_t id_min, uint32_t id_max, struct index *pk,
+		     int (*func)(struct space *sp, void *udata), void *udata)
+{
+	char key[6];
+	assert(mp_sizeof_uint(id_min) <= sizeof(key));
+	mp_encode_uint(key, id_min);
+	struct iterator *it = index_create_iterator(pk, ITER_GE, key, 1);
+	if (it == NULL)
+		return -1;
+	int rc;
+	struct tuple *tuple;
+	while ((rc = iterator_next(it, &tuple)) == 0 && tuple != NULL) {
+		uint32_t id;
+		if (tuple_field_u32(tuple, BOX_SPACE_FIELD_ID, &id) != 0)
+			continue;
+		if (id > id_max)
+			break;
+		struct space *space = space_cache_find(id);
+		if (space == NULL)
+			break;
+		rc = func(space, udata);
+		if (rc != 0)
+			break;
+	}
+	iterator_delete(it);
+	if (rc != 0)
+		return -1;
+	return 0;
+}
+
+int
+space_foreach(int (*func)(struct space *sp, void *udata), void *udata)
+{
+	struct space *space = space_by_id(BOX_SPACE_ID);
+	assert(space != NULL);
+	struct index *pk = space_index(space, 0);
+	assert(pk != NULL);
+	/*
+	 * Iterate over system spaces.
+	 */
+	int rc = space_foreach_helper(BOX_SYSTEM_ID_MIN, BOX_SYSTEM_ID_MAX,
+				      pk, func, udata);
+	if (rc != 0)
+		return -1;
+	/*
+	 * Iterate over non-system spaces.
+	 */
+	rc = space_foreach_helper(0, BOX_SYSTEM_ID_MIN - 1, pk, func, udata);
+	if (rc != 0)
+		return -1;
+	rc = space_foreach_helper(BOX_SYSTEM_ID_MAX + 1, BOX_SPACE_MAX,
+				  pk, func, udata);
+	if (rc != 0)
+		return -1;
+	return 0;
+}
+
 void
 space_dump_def(const struct space *space, struct rlist *key_list)
 {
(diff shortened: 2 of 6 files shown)


src/box/space.c
```diff
@@ -431,6 +431,70 @@ space_delete(struct space *space)
...
 
+int
+space_foreach(int (*func)(struct space *sp, void *udata), void *udata)
+{
+	struct space *space = space_by_id(BOX_SPACE_ID);
+	if (space == NULL)
+		return -1;
+	struct index *pk = space_index(space, 0);
+	if (pk == NULL)
+		return -1;
```
locker, line 442: Neither `space_by_id` nor `space_index` sets diag on failure. You should use `space_cache_find` and `index_find` if you want to fail `space_foreach` if `box.space._space` doesn't exist. However, `box.space._space` must exist, since we insert it into the cache even if there's no such record in the snap file so I think it's better to change this to an assertion.

src/box/space_cache.c
```diff
@@ -71,63 +71,6 @@ space_by_name(const char *name)
 	return (struct space *)mh_strnptr_node(spaces_by_name, space)->val;
 }
 
-/**
- * Visit all spaces and apply 'func'.
- */
-int
-space_foreach(int (*func)(struct space *sp, void *udata), void *udata)
```
locker, line 78: Why did you move this function? Looks like it's more about the space cache than about a single space.

src/box/space.c
```diff
@@ -431,6 +431,70 @@ space_delete(struct space *space)
...
+	/*
+	 * Step 2: iterate over non-system spaces.
+	 * The second traverse is required, because spaces with ids in the range
+	 * [0..BOX_SYSTEM_ID_MIN] are actually non-system.
+	 */
+	mp_encode_uint(key, 0);
+	it = index_create_iterator(pk, ITER_GE, key, 1);
+	if (it == NULL)
+		return -1;
+	while ((rc = iterator_next(it, &tuple)) == 0 && tuple != NULL) {
```
locker, line 479: Nit: It'd probably look better (no code duplication) if we had a helper function that would iterate over a space range:

```C
int
space_foreach(...)
{
    space_foreach_helper(0, BOX_SYSTEM_ID_MIN, cb, arg);
    space_foreach_helper(BOX_SYSTEM_ID_MAX + 1, BOX_SPACE_MAX, cb, arg);
}
```

test/box-luatest/gh_7954_sort_spaces_in_snap_by_id_test.lua
```diff
@@ -0,0 +1,46 @@
...
+
+-- Check that spaces in the snapshot file are strictly ordered.
+g.test_spaces_in_snap = function(cg)
+    local fio = require('fio')
+    local xlog = require('xlog')
+
+    -- Create some spaces, fill them with data, and write a snapshot.
+    local snap_name = cg.server:exec(function()
+        local spaces = {}
+        for i, id in ipairs{739, 761, 863, 881, 907, 937, 1009, 1033, 1039} do
```
locker, line 22: This test doesn't check that spaces with id < 256 are stored after system spaces.

changelogs/unreleased/gh-7954-sort-spaces-in-snapshot-by-id.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
+
+* Now all spaces in a snapshot file are sorted by their id (gh-7954).
```
locker, line 3: The doc doesn't say that spaces with id < 256 are stored after system spaces. Also, we should probably update the doc to mention that spaces are guaranteed to be sorted by id only in Tarantool 3.0 or later.

locker (changes requested):

src/box/space_cache.c
Gumix: My thoughts were as follows: `space_foreach()` is more about spaces in general rather than space cache. So after I removed direct access to `struct mh_i32ptr_t *spaces`, it can be placed anywhere.

src/box/space.c
Gumix: Changed to `space_foreach_helper` (3 times).

test/box-luatest/gh_7954_sort_spaces_in_snap_by_id_test.lua
Gumix: Added spaces with id < 256

changelogs/unreleased/gh-7954-sort-spaces-in-snapshot-by-id.md
Gumix: Changed the changelog and TarantoolBot request.

test/box-luatest/gh_7954_sort_spaces_in_snap_by_id_test.lua
```diff
@@ -0,0 +1,54 @@
...
+    -- 1. System spaces (id > 256 && id < 511)
+    -- 2. Spaces 167, 193, 239
+    -- 3. Spaces 739, 761, 863, 881, 907, 937
+    -- For each subgroup id >= prev_id
+    local prev_id = 0
+    local snap_path = fio.pathjoin(cg.server.workdir, snap_name)
+    for _, row in xlog.pairs(snap_path) do
+        if row.HEADER.type == "INSERT" then
+            local id = row.BODY.space_id
+            if id == 167 then
```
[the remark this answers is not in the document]
Gumix: Fixed

locker (approved):

PR #8378, opened 2023-02-27
httpc: added a decoders to a http response

### LINKED ISSUE #8363: http client: unable to decode "application/json; charset=utf-8" body
Tarantool version: 2.11.0-rc1.

```lua
tarantool> httpc = require('http.client').new()
tarantool> response = httpc:get('https://jsonplaceholder.typicode.com/todos/1')
tarantool> response:decode()
---
- error: 'builtin/http.client.lua:301: attempt to index field ''decoders'' (a nil
    value)'
...
```

https://github.com/tarantool/tarantool/blob/a989cd41bfc9e01a06227d1da65fdc0d76e66511/src/lua/httpc.lua#L301

ligurio:
Patch fixes a bug when body in response couldn't be decoded:

```
tarantool> httpc = require('http.client').new()
tarantool> response = httpc:get('https://jsonplaceholder.typicode.com/todos/1')
tarantool> response:decode()

---
- error: 'builtin/http.client.lua:301: attempt to index field ''decoders'' (a nil
    value)'
...
```

Reported-by: Alexander Turenko <alexander.turenko@tarantool.org>

Fixes #8363

NO_DOC=bugfix

### DIFF
--- a/src/lua/httpc.lua
+++ b/src/lua/httpc.lua
@@ -57,6 +57,21 @@ local decoders = {
     ['application/msgpack'] = function(body, _content_type) return msgpack.decode(body) end,
 }
 
+-- Extract all fields from a table except ones that start from
+-- the underscore.
+-- Copy-pasted from src/lua/init.lua.
+--
+-- Useful for __serialize.
+local function filter_out_private_fields(t)
+    local res = {}
+    for k, v in pairs(t) do
+        if not k:startswith('_') then
+            res[k] = v
+        end
+    end
+    return res
+end
+
 --
 --  <http> - create a new curl instance.
 --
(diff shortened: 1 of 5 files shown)


src/lua/httpc.lua
```diff
@@ -585,6 +585,7 @@ curl_mt = {
 
             if not opts.chunked then
                 resp.url = url_with_params
+                resp.decoders = decoders
```
Totktonada, line 588: Maybe it would be better to start the field from an underscore (so it'll be excluded from the autocompletion) and add `__serialize` that skips fields that start from the underscore. Or capture `decoders` in a locally created function and pass it to `decode_body` in a separate argument.

Totktonada: Now the field is named `_decoders` and it is excluded from the serialization.

igormunkin:
The patchset is also backported to 2.11 (2.11.0-rc1-8-g5dcf3114f) in scope of the following patches:
* 5dcf311 (httpc: add a test with decoding using a custom body decoder)
* cf39611 (httpc: add a decoders to a http response)

PR #8379, opened 2023-02-28
"config" bootstrap strategy

### LINKED ISSUE #7999: box.cfg.bootstrap_strategy = "config"
Continuation of #5272.
expand `box.cfg{bootstrap_strategy=...}` with option "config" as described [here](https://www.notion.so/tarantool/replication_connect_quorum-rework-RCQ-a749e1e6b8934f3388b2176ab52c4300).

sergepetrenko:
This patchset introduces `bootstrap_strategy = "config"` setting, the second one of the 3 desigined [here](https://www.notion.so/tarantool/replication_connect_quorum-rework-RCQ-a749e1e6b8934f3388b2176ab52c4300)

Closes #7999


sergepetrenko:
Note, checkpatch raises a false-positive error:
```
Commit 216e591ff432 ("replication: add bootstrap_strategy "config"")
--------------------------------------------------------------------
ERROR:EMBEDDED_FUNCTION_NAME: Prefer using '"%s...", __func__' to using 'bootstrap', this function's name, in a string
#226: FILE: src/box/box.cc:4433:
+					  "failed to connect to the bootstrap "

total: 1 errors, 502 lines checked
Commit 216e591ff432 ("replication: add bootstrap_strategy "config"") has style problems, please review.
```

src/lib/uri/uri.h
```diff
@@ -154,6 +154,14 @@ uri_escape(const char *src, size_t src_size, char *dst,
 size_t
 uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus);
 
+/** Determine if uris refer to the same host:service or unix socket path. */
+bool
+uri_equals(const struct uri *a, const struct uri *b);
+
+/** Check if a uri isn't empty. */
+bool
+uri_is_set(const struct uri *uri);
```
grafin, line 163: Why not `uri_is_nil`? That would be more like `tt_uuid_is_nil` which serves the same purpose.

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +179,210 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+        box_cfg = {
+            bootstrap_strategy = 'config',
+            bootstrap_leader = uuid1,
+            instance_uuid = uuid1,
+            replication = nil,
+        },
+    }
+end)
+
+g_config.after_test('test_uuid', function(cg)
```
grafin, line 225: Can this be moved after the test itself? Right now it is kinda awkward to read `before_test`->`after_test`->`test`. Same for other tests as well, won't mention it there.

grafin (changes requested):
Thanks for the patchset! I didn't finish reviewing the tests (will do a bit later).

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +179,210 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+        },
+    }
+end)
+
+g_config_success.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+g_config_success.test_correct_bootstrap_leader = function(cg)
+    cg.replica_set:start{}
```
grafin, line 381: Maybe we should add a test that server can bootstrap in `config` mode even if it can only connect to the bootstrap leader, while having many (eg 3 or 5) servers listed in `replication`?

src/box/replication.cc
```diff
@@ -892,8 +997,11 @@ applier_on_connect_f(struct trigger *trigger, void *event)
 
 void
 replicaset_connect(struct applier **appliers, int count,
-		   bool connect_quorum, bool keep_connect)
+		   bool connect_quorum, bool keep_connect,
+		   struct replica **bootstrap_leader)
 {
+	if (bootstrap_leader != NULL)
+		*bootstrap_leader = NULL;
```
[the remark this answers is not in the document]
sergepetrenko: It's only there for the case when `count == 0` and we might return an uninitialized `bootstrap_leader`. So I moved the check there.

<details><summary>Diff in first commit:</summary>

```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index d126b1fbe..a7d782eb0 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -966,11 +966,11 @@ replicaset_connect(struct applier **appliers, int count,
                   bool connect_quorum, bool keep_connect,
                   struct replica **bootstrap_leader)
 {
-       if (bootstrap_leader != NULL)
-               *bootstrap_leader = NULL;
        if (count == 0) {
                /* Cleanup the replica set. */
                replicaset_update(appliers, 0, false);
+               if (bootstrap_leader != NULL)
+                       *bootstrap_leader = NULL;
                return;
        }
 
```
</details>

I also decided to set `bootstrap_leader` to `NULL` explicitly when it can't be found in the last commit.
<details><summary>Diff in last commit:</summary>

```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index d740c03bc..11af1d094 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -1119,15 +1119,26 @@ replicaset_connect(struct applier **appliers, int count,
        } catch (Exception *e) {
                goto error;
        }
-       if (bootstrap_leader != NULL) {
-               if (state.bootstrap_leader_applier != NULL) {
+
+       if (bootstrap_leader == NULL)
+               return;
+
+       switch (bootstrap_strategy) {
+       case BOOTSTRAP_STRATEGY_LEGACY:
+       case BOOTSTRAP_STRATEGY_AUTO:
+               *bootstrap_leader = replicaset_find_join_master();
+               break;
+       case BOOTSTRAP_STRATEGY_CONFIG:
+               if (state.bootstrap_leader_applier == NULL) {
+                       *bootstrap_leader = NULL;
+               } else {
                        struct applier *leader = state.bootstrap_leader_applier;
                        *bootstrap_leader = replica_by_uuid(&leader->uuid);
                        assert(bootstrap_leader != NULL);
-               } else if (bootstrap_strategy == BOOTSTRAP_STRATEGY_AUTO ||
-                          bootstrap_strategy == BOOTSTRAP_STRATEGY_LEGACY) {
-                       *bootstrap_leader = replicaset_find_join_master();
                }
+               break;
+       default:
+               unreachable();
        }
        return;
 error:

```
</details>

src/box/replication.cc
```diff
@@ -1006,6 +1119,16 @@ replicaset_connect(struct applier **appliers, int count,
 	} catch (Exception *e) {
 		goto error;
 	}
+	if (bootstrap_leader != NULL) {
+		if (state.bootstrap_leader_applier != NULL) {
+			struct applier *leader = state.bootstrap_leader_applier;
+			*bootstrap_leader = replica_by_uuid(&leader->uuid);
+			assert(bootstrap_leader != NULL);
```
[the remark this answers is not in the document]
sergepetrenko: Yep, if a bootstrap leader applier was found, this means a `struct replica` for this applier was created as well.

src/box/replication.h
```diff
@@ -569,10 +575,12 @@ replicaset_add_anon(const struct tt_uuid *replica_uuid);
  *                       appliers have successfully connected.
  * \param keep_connect   if this flag is set do not force a reconnect if the
  *                       old connection to the replica is fine.
+ * \param bootstrap_leader the node chosen as bootstrap leader.
```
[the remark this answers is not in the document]
sergepetrenko: Sure:
```diff
diff --git a/src/box/replication.h b/src/box/replication.h
index da1124538..3019fdc6f 100644
--- a/src/box/replication.h
+++ b/src/box/replication.h
@@ -562,7 +562,8 @@ replicaset_add_anon(const struct tt_uuid *replica_uuid);
  *                       appliers have successfully connected.
  * \param keep_connect   if this flag is set do not force a reconnect if the
  *                       old connection to the replica is fine.
- * \param bootstrap_leader the node chosen as bootstrap leader.
+ * \param bootstrap_leader the node chosen as bootstrap leader or NULL, if
+ *                         bootstrap leader wasn't found.
  */
 void
 replicaset_connect(struct applier **appliers, int count,
```

src/box/replication.h
```diff
@@ -569,10 +575,12 @@ replicaset_add_anon(const struct tt_uuid *replica_uuid);
...
  */
 void
 replicaset_connect(struct applier **appliers, int count,
-		   bool connect_quorum, bool keep_connect);
+		   bool connect_quorum, bool keep_connect,
+		   struct replica **bootstrap_leader);
```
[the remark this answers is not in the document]
sergepetrenko: I understand your point, but here I also use `bootstrap_leader` as an input parameter. When `bootstrap_leader != 0`, `replicaset_connect` exits as soon, as configured bootstrap leader is reached.
But this doesn't happen when `bootstrap_leader == 0`.

src/box/box.cc
```diff
@@ -1268,16 +1298,49 @@ box_check_replication_anon(void)
...
+			 "the option takes no effect when bootstrap strategy "
+			 "is not 'config'");
+		return -1;
+	} else if (source == NULL) {
+		diag_set(ClientError, ER_CFG, "bootstrap_leader",
+			 "the option can't be empty when bootstrap strategy "
+			 "is 'config'");
+		return -1;
+	}
+	if (box_check_uri(uri, "bootstrap_leader") == 0)
```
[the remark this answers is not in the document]
sergepetrenko: Yes, I didn't think this part through in the RFC. And now the RFC is locked already.
I think the approach from the patch is better.

src/box/box.cc
```diff
@@ -1293,6 +1310,39 @@ box_check_replicaset_uuid(struct tt_uuid *uuid)
...
+	enum bootstrap_strategy strategy = box_check_bootstrap_strategy();
+	if (strategy != BOOTSTRAP_STRATEGY_CONFIG) {
+		if (source == NULL) {
+			/* Nothing to do. */
+			return 0;
+		}
+		diag_set(ClientError, ER_CFG, "bootstrap_leader",
+			 "the option takes no effect when bootstrap strategy "
+			 "is not 'config'");
+		return -1;
```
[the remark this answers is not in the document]
sergepetrenko: IMO ignoring wrong options does more bad than good. The user might not notice a misconfiguration then.

I suppose ignoring misconfigured options is only needed when we want to preserve backward compatibility. For example, bootstrap_strategy is `auto` by default, and this means `replication_connect_quorum` isn't used, but when `replication_connect_quorum` is set, `bootstrap_strategy` is coerced to `legacy`.

Here, in contrary, both options `bootstrap_strategy = 'config'` and `bootstrap_leader = ...` appear together, so no one can have a bad config left from some old instance. And if backward compatibility isn't hurt, I think we should enforce correct option usage.

src/lib/uri/uri.h
sergepetrenko: Makes sense, done:

```diff
diff --git a/src/lib/uri/uri.c b/src/lib/uri/uri.c
index f64b1055e..7a162c266 100644
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -484,7 +484,7 @@ uri_equals(const struct uri *a, const struct uri *b)
 }
 
 bool
-uri_is_set(const struct uri *uri)
+uri_is_nil(const struct uri *uri)
 {
-       return memcmp(uri, &uri_nil, sizeof(*uri)) != 0;
+       return memcmp(uri, &uri_nil, sizeof(*uri)) == 0;
 }
diff --git a/src/lib/uri/uri.h b/src/lib/uri/uri.h
index 846fc4ec8..b068592b6 100644
--- a/src/lib/uri/uri.h
+++ b/src/lib/uri/uri.h
@@ -158,9 +158,9 @@ uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus);
 bool
 uri_equals(const struct uri *a, const struct uri *b);
 
-/** Check if a uri isn't empty. */
+/** Check if a uri is empty. */
 bool
-uri_is_set(const struct uri *uri);
+uri_is_nil(const struct uri *uri);
 
 #if defined(__cplusplus)
 } /* extern "C" */
```

and in the last commit:
```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index 11af1d094..582a97f9b 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -964,7 +964,7 @@ is_applier_cfg_bootstrap_leader(const struct applier *applier)
                return tt_uuid_compare(&applier->uuid,
                                       &cfg_bootstrap_leader_uuid) == 0;
        }
-       if (uri_is_set(&cfg_bootstrap_leader_uri))
+       if (!uri_is_nil(&cfg_bootstrap_leader_uri))
                return uri_equals(&applier->uri, &cfg_bootstrap_leader_uri);
        return false;
 }
```

test/replication-luatest/bootstrap_strategy_test.lua
sergepetrenko: Sure. Done, won't post the diff since it's trivial.

test/replication-luatest/bootstrap_strategy_test.lua
sergepetrenko: Good idea, done:
```diff
diff --git a/test/replication-luatest/bootstrap_strategy_test.lua b/test/replication-luatest/bootstrap_strategy_test.lua
index b1b95a7ca..a542b23f4 100644
--- a/test/replication-luatest/bootstrap_strategy_test.lua
+++ b/test/replication-luatest/bootstrap_strategy_test.lua
@@ -8,6 +8,9 @@ local g_auto = t.group('gh-5272-bootstrap-strategy-auto')
 local uuid1 = '11111111-1111-1111-1111-111111111111'
 local uuid2 = '22222222-2222-2222-2222-222222222222'
 local uuid3 = '33333333-3333-3333-3333-333333333333'
+local uuida = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa'
+local uuidb = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb'
+local uuidc = 'cccccccc-cccc-cccc-cccc-cccccccccccc'
 
 local function server_is_ready(server)
     local ok, _ = pcall(function()
@@ -222,10 +225,6 @@ g_config.before_test('test_uuid', function(cg)
```
<skipped some moving of `after_test`>
```diff
@@ -327,23 +326,34 @@ g_config.test_single_leader = function(cg)
     end)
 end
 
+g_config.after_test('test_single_leader', function(cg)
+    cg.server1:stop()
+end)
+
 local g_config_success = t.group('gh-7999-bootstrap-strategy-config-success', {
                        {leader = 'server3'},
                        {leader = uuid3},
                    })
 
-g_config_success.before_test('test_correct_bootstrap_leader', function(cg)
-    local leader = cg.params.leader
+g_config_success.before_each(function(cg)
+    cg.leader = cg.params.leader
     -- cg.params can't have "/" for some reason, so recreate the path here.
-    if string.match(leader, 'server3') then
-        leader = server.build_listen_uri(leader)
+    if string.match(cg.leader, 'server3') then
+        cg.leader = server.build_listen_uri(cg.leader)
     end
+end)
+
+g_config_success.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+g_config_success.before_test('test_correct_bootstrap_leader', function(cg)
     cg.replica_set = replica_set:new{}
     cg.server1 = cg.replica_set:build_and_add_server{
         alias = 'server1',
         box_cfg = {
             bootstrap_strategy = 'config',
-            bootstrap_leader = leader,
+            bootstrap_leader = cg.leader,
             instance_uuid = uuid1,
             replication = {
                 server.build_listen_uri('server1'),
@@ -356,33 +366,65 @@ g_config_success.before_test('test_correct_bootstrap_leader', function(cg)
     cg.server2 = cg.replica_set:build_and_add_server{
         alias = 'server2',
         box_cfg = {
-            replicaset_uuid = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa',
+            replicaset_uuid = uuida,
             instance_uuid = uuid2,
         }
     }
     cg.server3 = cg.replica_set:build_and_add_server{
         alias = 'server3',
         box_cfg = {
-            replicaset_uuid = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb',
+            replicaset_uuid = uuidb,
             instance_uuid = uuid3,
             listen = {
                 server.build_listen_uri('server3'),
-                server.build_listen_uri('server3-alias')
             },
         },
     }
 end)
 
-g_config_success.after_each(function(cg)
-    cg.replica_set:drop()
-end)
-
 g_config_success.test_correct_bootstrap_leader = function(cg)
     cg.replica_set:start{}
-    t.helpers.retrying({}, cg.server1.exec, cg.server1, function()
+    t.helpers.retrying({}, cg.server1.exec, cg.server1, function(uuid)
         local t = require('luatest')
-        t.assert_equals(box.info.cluster.uuid,
-                        'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb',
+        t.assert_equals(box.info.cluster.uuid, uuid,
                         'Server bootstrapped from correct leader')
-    end)
+    end, {uuidb})
+end
+
+g_config_success.before_test('test_wait_only_for_leader', function(cg)
+    cg.replica_set = replica_set:new{}
+    cg.server1 = cg.replica_set:build_and_add_server{
+        alias = 'server1',
+        box_cfg = {
+            bootstrap_strategy = 'config',
+            bootstrap_leader = cg.leader,
+            replication = {
+                server.build_listen_uri('server1'),
+                server.build_listen_uri('unreachable_2'),
+                server.build_listen_uri('server3'),
+                server.build_listen_uri('unreachable_4'),
+            },
+            replication_connect_timeout = 1000,
+            replication_timeout = 0.1,
+        },
+    }
+    cg.server3 = cg.replica_set:build_and_add_server{
+        alias = 'server3',
+        box_cfg = {
+            replicaset_uuid = uuidb,
+            instance_uuid = uuid3,
+            listen = {
+                server.build_listen_uri('server3'),
+            },
+        },
+    }
+end)
+
+g_config_success.test_wait_only_for_leader = function(cg)
+    cg.replica_set:start{}
+    t.helpers.retrying({}, cg.server1.exec, cg.server1, function(uuid)
+        local t = require('luatest')
+        t.assert_equals(box.info.cluster.uuid, uuid,
+                        'Server boots as soon as sees the leader')
+    end, {uuidb})
 end

```

src/box/replication.cc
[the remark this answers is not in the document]
grafin: Thanks for explanation! Makes sense now.

grafin (approved):
Thanks for the fixes! LGTM.

src/box/replication.h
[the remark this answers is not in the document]
grafin: Thanks for the explanation

src/lib/uri/uri.c
```diff
@@ -456,3 +459,32 @@ uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus)
...
+	 */
+	return fields_are_equal(a->host, b->host) &&
+	       fields_are_equal(a->path, b->path) &&
+	       fields_are_equal(a->service, b->service);
+}
+
+bool
+uri_is_nil(const struct uri *uri)
+{
+	return memcmp(uri, &uri_nil, sizeof(*uri)) == 0;
```
Gerold103, line 489: This won't work for a uri which has all members 0, but non-zero padding between the members. Maybe better check just `host` member? Without it the uri has no purpose anyway.

src/lib/uri/uri.h
```diff
@@ -154,6 +154,14 @@ uri_escape(const char *src, size_t src_size, char *dst,
 size_t
 uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus);
 
+/** Determine if uris refer to the same host:service or unix socket path. */
+bool
+uri_equals(const struct uri *a, const struct uri *b);
+
+/** Check if a uri is empty. */
+bool
+uri_is_nil(const struct uri *uri);
```
Gerold103, line 163: You might need new tests in `test/unit/uri.c`.

src/box/box.cc
```diff
@@ -1006,6 +1006,21 @@ bad_uri:;
 	return -1;
 }
 
+/** Check validity of a uri passed in configuration option. */
+static int
+box_check_uri(struct uri *uri, const char *option_name)
+{
+	const char *source = cfg_gets(option_name);
+	if (source == NULL)
+		return 0;
```
Gerold103, line 1015: Here `uri` remains potentially uninitialized if the caller didn't do it. Maybe better nullify it before return.

src/box/box.cc
```diff
@@ -1293,6 +1310,39 @@ box_check_replicaset_uuid(struct tt_uuid *uuid)
...
+		return -1;
+	} else if (source == NULL) {
+		diag_set(ClientError, ER_CFG, "bootstrap_leader",
+			 "the option can't be empty when bootstrap strategy "
+			 "is 'config'");
+		return -1;
+	}
+	if (box_check_uri(uri, "bootstrap_leader") == 0)
+		return 0;
+	/* Not a uri. Try uuid then. */
```
Gerold103, line 1338: The error wasn't critical, but the counter in `box.stat` is incremented. Maybe better find a way not to set a diag.

src/box/replication.cc
```diff
@@ -930,13 +942,33 @@ struct replicaset_connect_state {
 	int failed;
 	/** Signaled when an applier connects or stops. */
 	struct fiber_cond wakeup;
+	/** An applier which connected to the bootstrap leader. */
+	struct applier *bootstrap_leader_applier;
```
Gerold103, line 946: I would recommend to reconsider if you really need this as a member. Maybe better find the leader applier by a simple fullscan of all the appliers when needed. Then you won't have to care about updating this member. IMO it is not needed often enough to keep it cached and complicate this struct.

src/box/replication.cc
```diff
@@ -930,13 +942,33 @@ struct replicaset_connect_state {
...
 };
 
 struct applier_on_connect {
 	struct trigger base;
 	struct replicaset_connect_state *state;
 };
 
+struct tt_uuid cfg_bootstrap_leader_uuid;
+struct uri cfg_bootstrap_leader_uri;
```
Gerold103, line 955: Don't we keep all the globals in the beginning of `replication.cc`?

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +181,249 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+end
+
+g_config.after_test('test_single_leader', function(cg)
+    cg.server1:stop()
+end)
+
+local g_config_success = t.group('gh-7999-bootstrap-strategy-config-success', {
+                       {leader = 'server3'},
+                       {leader = uuid3},
+                   })
```
Gerold103, line 335: Something is weird with the indentation here.

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +181,249 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+        alias = 'server2',
+        box_cfg = {
+            replicaset_uuid = uuida,
+            instance_uuid = uuid2,
+        }
+    }
+    cg.server3 = cg.replica_set:build_and_add_server{
+        alias = 'server3',
+        box_cfg = {
+            replicaset_uuid = uuidb,
```
Gerold103, line 375: Why are there 2 different replicaset UUIDs in replicas of the same `cg.replica_set`?

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +181,249 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+    cg.replica_set:start()
+    cg.server1:exec(function()
+        local t = require('luatest')
+        t.assert_equals(box.info.status, 'running', 'server is working')
+    end)
+end
+
+g_config.after_test('test_single_leader', function(cg)
+    cg.server1:stop()
+end)
```
Gerold103, line 330: Up to you. And to others too. I just want to emphasize how extremely complicated it is to read test cases when they are scattered across 3 separate functions. `before_test` and `after_test` are cancer.

test/replication-luatest/bootstrap_strategy_test.lua
```diff
@@ -179,3 +181,249 @@ g_auto.test_sync_waits_for_all_connected = function(cg)
...
+        box_cfg = {
+            replicaset_uuid = uuidb,
+            instance_uuid = uuid3,
+            listen = {
+                server.build_listen_uri('server3'),
+            },
+        },
+    }
+end)
+
```
Gerold103, line 421: Up to you. I would recommend to cover invalid URIs and UUIDs. Currently all the tests either don't pass them or pass valid values.

Gerold103 (commented):
Thanks for the patch!

src/box/box.cc
```diff
@@ -1641,7 +1641,8 @@ box_set_election_fencing_mode(void)
  * don't start appliers.
```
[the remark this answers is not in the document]
sergepetrenko: Ok. It's not that important to remember the bootstrap leader from the moment of connect, indeed.
I dropped that commit.

src/box/replication.cc
```diff
@@ -849,6 +849,77 @@ replicaset_update(struct applier **appliers, int count, bool keep_connect)
 	replicaset_set_sync_quorum();
 }
 
+/**
+ * Find a node in the replicaset on which the instance can try to register to
+ * join the replicaset.
+ */
+struct replica *
+replicaset_find_join_master(void)
```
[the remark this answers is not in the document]
sergepetrenko: That commit is dropped.

src/lib/uri/uri.c
sergepetrenko: > Maybe better check just host member? Without it the uri has no purpose anyway.

Not really, for example:
```
tarantool> require('uri').parse(3301)
---
- service: '3301'
...

```

But checking for `host`, `service` and `path` should work just fine. Done:
```diff
diff --git a/src/lib/uri/uri.c b/src/lib/uri/uri.c
index 7a162c266..a67c2ff64 100644
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -486,5 +486,5 @@ uri_equals(const struct uri *a, const struct uri *b)
 bool
 uri_is_nil(const struct uri *uri)
 {
-       return memcmp(uri, &uri_nil, sizeof(*uri)) == 0;
+       return uri->host == NULL && uri->path == NULL && uri->service == NULL;
 }
```

src/box/replication.cc
sergepetrenko: Ok, moved the definitions closer to the beginning of the file:
```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index 6487e785d..2310430c1 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -58,6 +58,9 @@ bool replication_skip_conflict = false;
 bool replication_anon = false;
 int replication_threads = 1;
 
+struct tt_uuid cfg_bootstrap_leader_uuid;
+struct uri cfg_bootstrap_leader_uri;
+
 struct replicaset replicaset;
 
 struct rlist replicaset_on_quorum_gain =
@@ -880,9 +883,6 @@ struct applier_on_connect {
        struct replicaset_connect_state *state;
 };
 
-struct tt_uuid cfg_bootstrap_leader_uuid;
-struct uri cfg_bootstrap_leader_uri;
-
 /** Check if the applier is connected to the configured bootstrap leader. */
 static bool
 is_applier_cfg_bootstrap_leader(const struct applier *applier)
```

src/box/box.cc
sergepetrenko: Good point. Done.

src/box/box.cc
sergepetrenko: Yes, I missed that, thanks!
What do you think of this approach?

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index f64b7c2ab..4e1c3932a 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -984,7 +984,7 @@ box_check_election_fencing_mode(void)
 
 /** A helper to check validity of a single uri. */
 static int
-check_uri(const struct uri *uri, const char *option_name)
+check_uri(const struct uri *uri, const char *option_name, bool set_diag)
 {
 	const char *auth_type = uri_param(uri, "auth_type", 0);
 	const char *errmsg;
@@ -999,28 +999,32 @@ check_uri(const struct uri *uri, const char *option_name)
 	}
 	return 0;
 bad_uri:;
-	char *uristr = tt_static_buf();
-	uri_format(uristr, TT_STATIC_BUF_LEN, uri, false);
-	diag_set(ClientError, ER_CFG, option_name,
-		 tt_sprintf("bad URI '%s': %s", uristr, errmsg));
+	if (set_diag) {
+		char *uristr = tt_static_buf();
+		uri_format(uristr, TT_STATIC_BUF_LEN, uri, false);
+		diag_set(ClientError, ER_CFG, option_name,
+			 tt_sprintf("bad URI '%s': %s", uristr, errmsg));
+	}
 	return -1;
 }
 
 /** Check validity of a uri passed in configuration option. */
 static int
-box_check_uri(struct uri *uri, const char *option_name)
+box_check_uri(struct uri *uri, const char *option_name, bool set_diag)
 {
 	const char *source = cfg_gets(option_name);
 	if (source == NULL) {
 		uri_create(uri, NULL);
 		return 0;
 	}
-	if (uri_create(uri, source) != 0) {
+	if (uri_create(uri, source) == 0)
+		return check_uri(uri, option_name, set_diag);
+	if (set_diag) {
 		diag_set(ClientError, ER_CFG, option_name,
 			 tt_sprintf("bad URI '%s': expected host:service or "
 				    "/unix.socket", source));
 	}
-	return check_uri(uri, option_name);
+	return -1;
 }
 
 static int
@@ -1035,7 +1039,7 @@ box_check_uri_set(const char *option_name)
 	int rc = 0;
 	for (int i = 0; i < uri_set.uri_count; i++) {
 		const struct uri *uri = &uri_set.uris[i];
-		if (check_uri(uri, option_name) != 0) {
+		if (check_uri(uri, option_name, true) != 0) {
 			rc = -1;
 			break;
 		}
@@ -1256,19 +1260,22 @@ box_check_replication_sync_timeout(void)
 }
 
 static inline int
-box_check_uuid(struct tt_uuid *uuid, const char *name)
+box_check_uuid(struct tt_uuid *uuid, const char *name, bool set_diag)
 {
 	*uuid = uuid_nil;
 	const char *uuid_str = cfg_gets(name);
 	if (uuid_str == NULL)
 		return 0;
 	if (tt_uuid_from_string(uuid_str, uuid) != 0) {
-		diag_set(ClientError, ER_CFG, name, uuid_str);
+		if (set_diag)
+			diag_set(ClientError, ER_CFG, name, uuid_str);
 		return -1;
 	}
 	if (tt_uuid_is_nil(uuid)) {
-		diag_set(ClientError, ER_CFG, name,
-			 tt_sprintf("nil UUID is reserved"));
+		if (set_diag) {
+			diag_set(ClientError, ER_CFG, name,
+				 tt_sprintf("nil UUID is reserved"));
+		}
 		return -1;
 	}
 	return 0;
@@ -1303,13 +1310,13 @@ box_check_replication_anon(void)
 static int
 box_check_instance_uuid(struct tt_uuid *uuid)
 {
-	return box_check_uuid(uuid, "instance_uuid");
+	return box_check_uuid(uuid, "instance_uuid", true);
 }
 
 static int
 box_check_replicaset_uuid(struct tt_uuid *uuid)
 {
-	return box_check_uuid(uuid, "replicaset_uuid");
+	return box_check_uuid(uuid, "replicaset_uuid", true);
 }
 
 /** Check bootstrap_leader option validity. */
@@ -1335,10 +1342,10 @@ box_check_bootstrap_leader(struct uri *uri, struct tt_uuid *uuid)
 			 "is 'config'");
 		return -1;
 	}
-	if (box_check_uri(uri, "bootstrap_leader") == 0)
+	if (box_check_uri(uri, "bootstrap_leader", false) == 0)
 		return 0;
 	/* Not a uri. Try uuid then. */
-	if (box_check_uuid(uuid, "bootstrap_leader") == 0)
+	if (box_check_uuid(uuid, "bootstrap_leader", false) == 0)
 		return 0;
 	diag_set(ClientError, ER_CFG, "bootstrap_leader",
 		 "the value must be either a uri or a uuid");
```
</details>

Gerold103: Looks ugly. But I can't see a better option.

sergepetrenko: Another variant would be to make `box_check_uri` and `box_check_uuid` to return error strings like this:
`box_check_uuid(struct tt_uuid *uuid, const char *name, char **errmsg)` and not set diag at all.
The caller of `box_check_uuid/uri` can then decide whether to use the error message anyhow (like set diag) or ignore it.
I'm not sure it's any better than the variant above though.

src/lib/uri/uri.c
sergepetrenko: After giving it some thought, I'd like to return to checking for a fully empty uri.
The reasons are: the name uri_is_nil() implies that the uri is really unset / empty. But if we check for host / path / service, some strange uris (like "#fragment", "?q1=v1") and such will still be considered nil. Which's not fine IMO.

Besides, 'uri_create(NULL)` does a memset to 0, so such a uri won't have a non-zero padding.

src/lib/uri/uri.h
sergepetrenko: Sure:

```diff
diff --git a/src/lib/uri/uri.c b/src/lib/uri/uri.c
index 4e0792bc5..07551cc53 100644
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -483,5 +483,10 @@ uri_equals(const struct uri *a, const struct uri *b)
 bool
 uri_is_nil(const struct uri *uri)
 {
+	/*
+	 * Check only these 3 fields, because without them a uri doesn't make
+	 * sense. But technically this will give some false positives. For
+	 * example, for uris with non-empty fragment or query.
+	 */
 	return uri->host == NULL && uri->path == NULL && uri->service == NULL;
 }
diff --git a/test/unit/uri.c b/test/unit/uri.c
index 34961d6c6..97386d85f 100644
--- a/test/unit/uri.c
+++ b/test/unit/uri.c
@@ -54,7 +54,7 @@ empty_uri_create(struct uri *uri)
 static void
 empty_uri_check(struct uri *uri, const char *msg)
 {
-	plan(11);
+	plan(12);
 	is(uri->scheme, NULL, "%s scheme", msg);
 	is(uri->login, NULL, "%s login", msg);
 	is(uri->password, NULL, "%s password", msg);
@@ -66,6 +66,7 @@ empty_uri_check(struct uri *uri, const char *msg)
 	is(uri->host_hint, 0, "%s hint", msg);
 	is(uri->param_count, 0, "%s param count", msg);
 	is(uri->params, NULL, "%s params", msg);
+	ok(uri_is_nil(uri), "%s is_nil()", msg);
 	check_plan();
 }
 
@@ -137,6 +138,53 @@ test_move_empty(void)
 	footer();
 }
 
+struct uri_equal_expected {
+	/** Source string for the first uri. */
+	const char *src_a;
+	/** Source string for the second uri. */
+	const char *src_b;
+	/** Expected comparison result. */
+	bool is_equal;
+};
+
+static void
+test_equals(void)
+{
+	struct uri_equal_expected test_pairs[] = {
+		{NULL, NULL, true},
+		{"localhost", "localhost", true},
+		{"user@localhost", "localhost", true},
+		{"user:pass@localhost", "localhost", true},
+		{"user:pass@localhost", "user@localhost", true},
+		{"localhost:3301", "localhost:3302", false},
+		{"host_a", "host_b", false},
+		{"scheme://localhost", "localhost", true},
+		{"scheme1://host:port", "scheme2://host:port", true},
+		{"localhost/path/to/file", "localhost", false},
+		{"/path/to/file", "/path/to/file", true},
+		{"/path/to/file", "localhost/path/to/file", false},
+		{"unix/path/to/file", "/path/to/file", false},
+		{"unix/:/path/to/file", "/path/to/file", true},
+	};
+	header();
+	plan(3 * lengthof(test_pairs));
+	for (unsigned i = 0; i < lengthof(test_pairs); i++) {
+		struct uri uri_a, uri_b;
+		const char *src_a = test_pairs[i].src_a;
+		const char *src_b = test_pairs[i].src_b;
+		bool is_equal = test_pairs[i].is_equal;
+		ok(uri_create(&uri_a, src_a) == 0, "uri_create(%s)",
+		   src_a ? src_a : "NULL");
+		ok(uri_create(&uri_b, src_b) == 0, "uri_create(%s)",
+		   src_b ? src_b : "NULL");
+		is(uri_equals(&uri_a, &uri_b), is_equal, "%s %s equal to %s",
+		   src_a ? src_a : "NULL", is_equal ? "is" : "isn't",
+		   src_b ? src_b : "NULL");
+	}
+	check_plan();
+	footer();
+}
+
 struct uri_param_expected {
 	/** URI parameter name */
 	const char *name;
@@ -780,11 +828,12 @@ test_unescape_special_cases(void)
 int
 main(void)
 {
-	plan(10);
+	plan(11);
 	test_copy_sample();
 	test_copy_empty();
 	test_move_sample();
 	test_move_empty();
+	test_equals();
 	test_string_uri_with_query_params_parse();
 	test_string_uri_set_with_query_params_parse();
 	test_invalid_string_uri_set();
```
</details>

test/replication-luatest/bootstrap_strategy_test.lua
sergepetrenko: Indeed, done:
```diff
diff --git a/test/replication-luatest/bootstrap_strategy_test.lua b/test/replication-luatest/bootstrap_strategy_test.lua
index cb894a5fd..ca33c766c 100644
--- a/test/replication-luatest/bootstrap_strategy_test.lua
+++ b/test/replication-luatest/bootstrap_strategy_test.lua
@@ -436,3 +436,22 @@ g_config_success.test_wait_only_for_leader = function(cg)
                         'Server boots as soon as sees the leader')
     end, {uuidb})
 end
+
+local g_config_fail = t.group('gh-7999-bootstrap-strategy-config-fail')
+
+g_config_fail.test_bad_uri_or_uuid = function()
+    local bad_config = {
+        {}, -- empty table.
+        {'a'}, -- non-empty table.
+        {3301},
+        'abracadabra', -- neither a URI or a UUID.
+        'z2345678-1234-1234-1234-12345678', -- not a UUID.
+    }
+    local errmsg = "Incorrect value for option 'bootstrap_leader':"
+    for _, leader in pairs(bad_config) do
+        t.assert_error_msg_contains(errmsg, box.cfg, {
+            bootstrap_strategy = 'config',
+            bootstrap_leader = leader
+        })
+    end
+end 
```

src/lib/uri/uri.h
```diff
@@ -154,6 +154,14 @@ uri_escape(const char *src, size_t src_size, char *dst,
 size_t
 uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus);
 
+/** Determine if uris refer to the same host:service or unix socket path. */
+bool
+uri_equals(const struct uri *a, const struct uri *b);
```
Gerold103, line 159: @locker, what is the most canonic rule? `_equals()` or `_is_equal()`? I see both versions.

@sergepetrenko - the commit message mentions `uris_are_equal()` which is probably outdated. In the current version I see `uri_equals()`. Besides, it doesn't look like uris are equal. You compare only the host and port. So probably it should be called `_addr_is_equal()` or alike.

src/box/replication.cc
```diff
@@ -1279,6 +1319,40 @@ replicaset_find_join_master(void)
 	return leader;
 }
 
+struct replica *
+replicaset_find_join_master_cfg(void)
```
Gerold103, line 1323: `replicaset_find_join_master_auto()` and `replicaset_find_join_master_cfg()` can be static.

src/box/replication.cc
```diff
@@ -971,6 +1009,8 @@ replicaset_connect(struct applier **appliers, int count,
 			break;
 		if (count - state.failed < replicaset_connect_quorum(count))
 			break;
+		if (state.ready)
+			break;
```
Gerold103, line 1013: Do you really need this flag? It seems you could simply walk here all the appliers and check if any of them is a connected bootstrap config leader. Wouldn't take much more code, but would allow to avoid adding more flags which need to be remembered to be updated.

Gerold103 (commented):
Thanks for the fixes!

src/lib/uri/uri.h
locker: > @locker, what is the most canonic rule? `_equals()` or `_is_equal()`? I see both versions.

`_is_equal()`. I see the only place where we use `_equals()` - [`tt_bitset_index_expr_equals()`](https://github.com/tarantool/tarantool/blob/9fcafbd14b962d9f844a9d114518fa690753f276/src/lib/bitset/index.h#L211), but it's ancient code and it doesn't actually compare two objects of the same type.

sergepetrenko: Thanks!

src/box/replication.cc
sergepetrenko: It seems I don't, thanks! I was having trouble with this code especially with `bootstrap_strategy = 'supervised'`. I was trying to go the other way and put everything to this flag: https://github.com/sergepetrenko/tarantool/commit/c3b471f7e454049a377cd7d69e3d82dd9f153539. It didn't turn out too well, and your approach looks promising. Will try.

src/lib/uri/uri.h
sergepetrenko: Changed `uri_equals()` to `uri_addr_is_equal()`:

```diff
diff --git a/src/lib/uri/uri.c b/src/lib/uri/uri.c
index 07551cc53..7e9ac10f5 100644
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -469,7 +469,7 @@ fields_are_equal(const char *a, const char *b)
 }
 
 bool
-uri_equals(const struct uri *a, const struct uri *b)
+uri_addr_is_equal(const struct uri *a, const struct uri *b)
 {
 	/*
 	 * Either service or path will be NULL depending on whether this is a
diff --git a/src/lib/uri/uri.h b/src/lib/uri/uri.h
index b068592b6..78e453fed 100644
--- a/src/lib/uri/uri.h
+++ b/src/lib/uri/uri.h
@@ -156,7 +156,7 @@ uri_unescape(const char *src, size_t src_size, char *dst, bool decode_plus);
 
 /** Determine if uris refer to the same host:service or unix socket path. */
 bool
-uri_equals(const struct uri *a, const struct uri *b);
+uri_addr_is_equal(const struct uri *a, const struct uri *b);
 
 /** Check if a uri is empty. */
 bool
diff --git a/test/unit/uri.c b/test/unit/uri.c
index 97386d85f..e97166408 100644
--- a/test/unit/uri.c
+++ b/test/unit/uri.c
@@ -148,7 +148,7 @@ struct uri_equal_expected {
 };
 
 static void
-test_equals(void)
+test_addr_is_equal(void)
 {
 	struct uri_equal_expected test_pairs[] = {
 		{NULL, NULL, true},
@@ -177,9 +177,9 @@ test_equals(void)
 		   src_a ? src_a : "NULL");
 		ok(uri_create(&uri_b, src_b) == 0, "uri_create(%s)",
 		   src_b ? src_b : "NULL");
-		is(uri_equals(&uri_a, &uri_b), is_equal, "%s %s equal to %s",
-		   src_a ? src_a : "NULL", is_equal ? "is" : "isn't",
-		   src_b ? src_b : "NULL");
+		is(uri_addr_is_equal(&uri_a, &uri_b), is_equal,
+		   "%s %s equal to %s", src_a ? src_a : "NULL",
+		   is_equal ? "is" : "isn't", src_b ? src_b : "NULL");
 	}
 	check_plan();
 	footer();
@@ -833,7 +833,7 @@ main(void)
 	test_copy_empty();
 	test_move_sample();
 	test_move_empty();
-	test_equals();
+	test_addr_is_equal();
 	test_string_uri_with_query_params_parse();
 	test_string_uri_set_with_query_params_parse();
 	test_invalid_string_uri_set();
```
</details>

src/box/replication.cc
sergepetrenko: Thanks for noticing! Fixed.

src/box/replication.cc
sergepetrenko: I reworked this part, but can't share a diff, it's rather dirty. Please check out the new approach in the commit itself.
There's also one extra commit: `428b74932 replication: refactor replicaset_connect()`

Gerold103 (approved):
Thanks for the fixes!

PR #8387, opened 2023-03-01
module api: fix formatting in comments for Doxygen

### LINKED ISSUE #8194: Fix formatting in Doxygen comments
**Bug description**

Some doxygen comments has broken formatting:

- unnecessary round brackets around `\sa` and `@sa`
- unnecessary `<` and `>` around function's names (it prevents to make Doxygen links there)
- etc 

See Doxygen documentation rendered to HTML in https://tarantool.github.io/tarantool/api/html/module_8h.html

ligurio:
src: fix formatting comments for Doxygen
    
    - remove < and > around function names to make them resolvable
    - place @sa and \sa to the new lines and remove round parenthesis around
    - add missed dots
    - added comments for box_tuple_upsert, box_tuple_update, luaT_pusherror
      etc.
    - replace \example tags with \code and \endcode, because \example
      requires a file name with example, when \code allows to place a code
      inline
    - comments for macroses alignof and likely/unlikely looks a bit ugly,
      but I couldn't find a better way to resolve doxygen warning

Closes #8194

NO_CHANGELOG=fix doxygen
NO_DOC=fix doxygen
NO_TEST=fix doxygen

### DIFF
--- a/.github/workflows/publish-module-api-doc.yaml
+++ b/.github/workflows/publish-module-api-doc.yaml
@@ -1,10 +1,10 @@
-name: Publish module API documentation
+name: Build and publish module API documentation
 
 on:
   push:
     branches: [master]
   pull_request:
-    types: [opened, reopened, synchronize, labeled]
+    types: [opened, reopened, synchronize]
   workflow_dispatch:
 
 concurrency:
@@ -28,12 +28,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  publish-api-doc:
-    # Run on push to the branch 'master' or on pull request if the 'full-ci'
-    # label is set.
-    if: github.repository == 'tarantool/tarantool' &&
-        ( github.event_name != 'pull_request' ||
-          contains(github.event.pull_request.labels.*.name, 'full-ci') )
+  build-api-doc:
     runs-on: ubuntu-latest
     steps:
       - name: Prepare checkout
@@ -57,7 +52,8 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@v4.4.0
         with:
           folder: doc/
-        if: github.ref == 'refs/heads/master' &&
+        if: github.repository == 'tarantool/tarantool' &&
+            github.ref == 'refs/heads/master' &&
             github.event_name != 'pull_request'
 
       - name: Send VK Teams message on failure
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -626,7 +626,8 @@ API_EXPORT int
 box_session_push(const char *data, const char *data_end);
 
 /**
- * \return current session's unique monotonic identifier (\sa box.session.id)
+ * \return current session's unique monotonic identifier
+ * \sa box.session.id
  */
 API_EXPORT uint64_t
 box_session_id(void);
--- a/src/box/decimal.h
+++ b/src/box/decimal.h
@@ -247,7 +247,7 @@ box_decimal_to_uint64(const box_decimal_t *dec, uint64_t *num);
  * If @a scale if greater than current @a dec scale, do nothing.
  *
  * @param dec decimal number
- * @oaram scale target scale
+ * @param scale target scale
  * @return NULL if @a scale is out of supported range
  * @return @a dec (changed)
  */
@@ -269,7 +269,7 @@ box_decimal_round(box_decimal_t *dec, int scale);
  * @sa box_decimal_round
  *
  * @param dec decimal number
- * @oaram scale target scale
+ * @param scale target scale
  * @return NULL if @a scale is out of supported range
  * @return @a dec (changed)
  */
@@ -296,7 +296,7 @@ box_decimal_trim(box_decimal_t *dec);
  * @sa box_decimal_trim
  *
  * @param dec decimal number
- * @oaram scale target scale
+ * @param scale target scale
  * @return NULL if scale is out of supported range (less than zero
  *              or too big)
  * @return @a dec (changed)
@@ -353,7 +353,7 @@ box_decimal_minus(box_decimal_t *res, const box_decimal_t *dec);
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error (an overflow for example)
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_add(box_decimal_t *res, const box_decimal_t *lhs,
@@ -366,7 +366,7 @@ box_decimal_add(box_decimal_t *res, const box_decimal_t *lhs,
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error (an overflow for example)
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_sub(box_decimal_t *res, const box_decimal_t *lhs,
@@ -379,7 +379,7 @@ box_decimal_sub(box_decimal_t *res, const box_decimal_t *lhs,
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error (an overflow for example)
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_mul(box_decimal_t *res, const box_decimal_t *lhs,
@@ -392,7 +392,7 @@ box_decimal_mul(box_decimal_t *res, const box_decimal_t *lhs,
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_div(box_decimal_t *res, const box_decimal_t *lhs,
@@ -407,7 +407,7 @@ box_decimal_div(box_decimal_t *res, const box_decimal_t *lhs,
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_remainder(box_decimal_t *res, const box_decimal_t *lhs,
@@ -423,7 +423,7 @@ box_decimal_remainder(box_decimal_t *res, const box_decimal_t *lhs,
  * @param res where to hold the result
  * @param dec decimal operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_log10(box_decimal_t *res, const box_decimal_t *dec);
@@ -434,7 +434,7 @@ box_decimal_log10(box_decimal_t *res, const box_decimal_t *dec);
  * @param res where to hold the result
  * @param dec decimal operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_ln(box_decimal_t *res, const box_decimal_t *dec);
@@ -446,7 +446,7 @@ box_decimal_ln(box_decimal_t *res, const box_decimal_t *dec);
  * @param lhs left hand side operand, base
  * @param rhs right hand side operand, power
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_pow(box_decimal_t *res, const box_decimal_t *lhs,
@@ -458,7 +458,7 @@ box_decimal_pow(box_decimal_t *res, const box_decimal_t *lhs,
  * @param res where to hold the result
  * @param dec decimal operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_exp(box_decimal_t *res, const box_decimal_t *dec);
@@ -469,7 +469,7 @@ box_decimal_exp(box_decimal_t *res, const box_decimal_t *dec);
  * @param res where to hold the result
  * @param dec decimal operand
  * @return NULL on an error
- * @return decimal result (@res)
+ * @return decimal result (@a res)
  */
 API_EXPORT box_decimal_t *
 box_decimal_sqrt(box_decimal_t *res, const box_decimal_t *dec);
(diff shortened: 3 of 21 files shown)


Totktonada (approved):
Thank you for the patch!

I'm not a Doxygen expert and TBH some changes are like 'hm, okay, I guess this way it'll be rendered to HTML better' for me. However there is nothing I would be against.

src/lib/core/fiber.h
```diff
@@ -439,35 +439,39 @@ cord_slab_cache(void);
...
  * Typical usage is illustrated in the sketch below.
  *
+ * \code
  *  | size_t region_svp = box_region_used();
  *  | while (<...>) {
  *  |     char *buf = box_region_alloc(<...>);
  *  |     <...>
  *  | }
  *  | box_region_truncate(region_svp);
+ * \endcode
```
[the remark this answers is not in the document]
ligurio: Yes, we should.
I'll fix it.

image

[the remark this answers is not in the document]
ligurio: Fixed by tags `\code` and `\endcode`.

.github/workflows/publish-module-api-doc.yaml
```diff
@@ -1,4 +1,4 @@
-name: Publish module API documentation
+name: Build and publish module API documentation
```
Totktonada, line 1:
```yaml
on:
  push:
    branches: [master]
  pull_request:
    types: [opened, reopened, synchronize, labeled]
  workflow_dispatch:
```

`labeled` is redundant here: we don't want to re-run the workflow at changes in the pull request labels. It was relevant, when the `full-ci` logic was here, but it is removed now.

If omit `types`, [it'll be](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) `[opened, reopened, synchronize]` -- fits good for us.

I think the overall logic would be pretty similar to one of existing run-always workflows, say, `debug.yml`.

.github/workflows/publish-module-api-doc.yaml
```diff
@@ -28,12 +28,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  publish-api-doc:
-    # Run on push to the branch 'master' or on pull request if the 'full-ci'
-    # label is set.
-    if: github.repository == 'tarantool/tarantool' &&
```
Totktonada, line 34: `github.repository == 'tarantool/tarantool'` condition looks meaningful, at least for publishing.

src/box/decimal.h
```diff
@@ -353,7 +353,7 @@ box_decimal_minus(box_decimal_t *res, const box_decimal_t *dec);
  * @param lhs left hand side operand
  * @param rhs right hand side operand
  * @return NULL on an error (an overflow for example)
- * @return decimal result (@res)
```
Totktonada, line 356: It means that the return value *is* the first parameter, `res`. I would replace it with `(@a res)`. The same for the functions below.

src/box/decimal.h
```diff
@@ -515,19 +515,22 @@ box_decimal_mp_decode(box_decimal_t *dec, const char **data);
...
+ *  |
+ *  box_decimal_mp_decode() must be called for this position
+ * \endcode
  *
  * This function is suitable to finish decoding after calling
  * mp_decode_extl() (from the msgpuck library).
  *
  * @param dec where to store the decoded decimal
  * @param data pointer to a buffer with the msgpack data
+ * @param size size of a buffer a buffer with the msgpack data
```
Totktonada, line 533: The important detail here is that a caller should pass an exact size of the decimal data, not just size of the buffer. It is the size that `mp_decode_extl()` returns.

I would add this information into the description, because otherwise it is not clear how to call the function. For example:

```c
 * @param size size of the decimal data in the buffer; this value
 *             is returned by mp_decode_extl()
```

src/box/decimal.h
```diff
@@ -515,19 +515,22 @@ box_decimal_mp_decode(box_decimal_t *dec, const char **data);
...
- *  | box_decimal_mp_decode() must be called for this position
+ * \code
+ *  box_decimal_mp_decode_data() must be called for this position
+ *                                         |
+ *                                         v
+ *  <msgpack type> <size> <extension type> <data>
+ *  ^
+ *  |
+ *  box_decimal_mp_decode() must be called for this position
+ * \endcode
```
Totktonada, line 526: Nit: It seems the one whitespace indent doesn't affect anything, so I would eliminate it. Or it is by intention?

src/box/box.h
```diff
@@ -626,7 +626,7 @@ API_EXPORT int
 box_session_push(const char *data, const char *data_end);
 
 /**
- * \return current session's unique monotonic identifier (\sa box.session.id)
+ * \return current session's unique monotonic identifier \sa box.session.id
  */
```
Totktonada, line 630: It looks so, when rendered into HTML:

image

So, maybe, move `\sa <...>` into its own line in the comment too?

src/box/error.h
```diff
@@ -122,6 +122,8 @@ box_error_clear(void);
 /**
  * Set the last error.
  *
+ * \param file
+ * \param line
```
Totktonada, line 126: It suppresses the following Doxygen warning:

> <...>/src/module.h:2670: warning: The following parameters of box_error_set(const char *file, unsigned line, uint32_t code, const char *format,...) are not documented:
>   parameter 'file'
>   parameter 'line'

I wouldn't say that the parameters are documented now. My proposal:

```c
 * \param file file name, usually __FILE__ macro
 * \param line line number in the file, usually __LINE__ macro
```

src/box/index.h
```diff
@@ -243,6 +243,7 @@ box_index_count(uint32_t space_id, uint32_t index_id, int type,
  * @param tuple Tuple from which need to extract key.
  * @param space_id Space identifier.
  * @param index_id Index identifier.
+ * @param key_size
```
Totktonada, line 246: Same here, I think it is better to actually document parameters.

It is how it is documented in the `tuple_extract_key()` definition.

```c
 * @param key_size - here will be size of extracted key
```

First, I would mark the parameter as the output parameter: `@param[out]`. Next, I would mention that NULL can be passed here if the key size is not needed.

For example:

```c
 * @param[out] key_size where to store the size of the extracted
 *                      key; pass NULL here if it is not needed
```

src/box/key_def.h
```diff
@@ -310,24 +310,26 @@ enum {
...
+ *  * Verify that box_key_part_def_t has the same size when
+ *  * compiled within tarantool and within the module.
+ *  *
+ *  * It is important, because the module allocates an array of key
+ *  * parts and passes it to box_key_def_new_v2() tarantool
+ *  * function.
+ *  *(slash)
+ * static_assert(sizeof(box_key_part_def_t) == BOX_KEY_PART_DEF_T_SIZE,
+ *               "sizeof(box_key_part_def_t)");
+ * \endcode
```
Totktonada, line 332: Same here: is the one whitespace indent is on purpose?

src/box/key_def.h
```diff
@@ -310,24 +310,26 @@ enum {
 };
 
 /**
```
Totktonada, line 312: If we'll not add a documentation comment to the constant itself, it'll look so:

image

I propose the following:

```c
enum {
        /** The constant. */
        BOX_KEY_PART_DEF_T_SIZE = 64,
};
```

It looks so:

image

src/box/key_def.h
```diff
@@ -387,7 +389,7 @@ typedef union PACKED {
 		 *
 		 * => key: ["bar"]
 		 *
-		 * Note: When the path is given, <field_type>
+		 * Note: When the path is given, \a field_type
 		 * means type of the nested field.
```
Totktonada, line 393: The weird thing is that neither the fields on the structure, nor the comments to them are placed into the documentation. In fact, it means that one can't use the API based only on the doc. Need to look into module.h anyway.

Don't know, is there a way to instruct Doxygen to don't strip the documentation.

src/box/tuple.h
```diff
@@ -320,9 +320,27 @@ box_tuple_next(box_tuple_iterator_t *it);
...
 
+/**
+ * Update tuple data in a space.
+ * @param tuple  Tuple to update.
+ * @param expr
+ * @param expr_end
+ *
+ * @retval  NULL The tuple was not updated.
+ * @retval  box_tuple_t The tuple successfully updated in a space.
+ */
```
Totktonada, line 331: `box_update()` is about updating a tuple in a space. This function just update a tuple.

`expr` and `expr_end` description could be grabbed from `box_update()` or `xrow_update_read_ops()`.

Nits:

* I'm a markdown guy and I like when paragraphs are separated by an empty line. IOW, I would separate the brief description from the parameters description:

  ```c
  /**
   * Blah Blah.
   *
   * @param <...>
   * <...>
   */
  ```
* Why two whitespaces?

src/box/tuple.h
```diff
@@ -320,9 +320,27 @@ box_tuple_next(box_tuple_iterator_t *it);
...
 
+/**
+ * Upsert tuple data to a space.
+ * @param tuple  Tuple to upsert.
+ * @param expr
+ * @param expr_end
+ *
+ * @retval  NULL The tuple was not upserted.
+ * @retval  box_tuple_t The tuple successfully upserted to a space.
+ */
```
Totktonada, line 343: Same comments as for `box_tuple_update()`, but also I wouldn't say upsert/upserted. It is update or insert, right? We can just say so.

src/lib/core/coio.h
```diff
@@ -169,7 +169,7 @@ enum {
 /**
  * Wait until READ or WRITE event on socket (\a fd). Yields.
  * \param fd - non-blocking socket file description
- * \param events - requested events to wait.
+ * \param event - requested events to wait.
  * Combination of TNT_IO_READ | TNT_IO_WRITE bit flags.
  * \param timeout - timeout in seconds.
  * \retval 0 - timeout
```
Totktonada, line 175: It's a bit too picky, but on the 'just asked' right: shouldn't we strip the dashes?

image

Feel free to ignore.

src/lib/core/fiber.h
```diff
@@ -340,6 +340,7 @@ fiber_set_cancellable(bool yesno);
 
 /**
  * Set fiber to be joinable (false by default).
+ * \param fiber to be joinable
```
Totktonada, line 343: 'fiber to (un)set the joinable property' I would say. Because the second parameter can be `false`.

src/box/key_def.h
Totktonada: Mislooked it. Please, ignore.

src/lua/utils.h
```diff
@@ -388,9 +388,11 @@ luaL_toint64(struct lua_State *L, int idx);
 LUA_API int
 luaT_call(lua_State *L, int nargs, int nreturns);
 
-/*
+/**
  * Like luaL_dostring(), but in case of error sets fiber diag instead
  * of putting error on stack.
+ * @param L is a Lua state
+ * @param str
```
Totktonada, line 395: Nits:

* Please, separate parameters from the brief description with an empty line.
* Isn't `is` redundant? AFAIS, we usually omit it in `@param`.
* How about some `str` description? 'Lua code to load and run' or kinda.

src/box/box.h
```diff
@@ -626,7 +626,7 @@ API_EXPORT int
 box_session_push(const char *data, const char *data_end);
```
Totktonada, line 626: Nit: `src: <...>` in the commit message doesn't look as a component. I would say `module api` or `doc`.

Totktonada (approved):
Thank you for the update!

I'm approving with comments. It means that I think that I added all the value I can add here and leave you with it to make changes or discard comments.

I consider factual mistakes as 'must be fixed' and other comments as 'nice to fix while we're here'.

src/box/decimal.h
ligurio: replaced `@res` with `@a res` everywhere

src/box/decimal.h
ligurio: it was an attempt to visually highlight code block in a comment, removed indentation to make consistent with other comments with code blocks

src/box/error.h
ligurio: Updated, thanks

src/box/index.h
ligurio: Updated, thanks:

```diff
--- a/src/box/index.h
+++ b/src/box/index.h
@@ -243,7 +243,8 @@ box_index_count(uint32_t space_id, uint32_t index_id, int type,
  * @param tuple Tuple from which need to extract key.
  * @param space_id Space identifier.
  * @param index_id Index identifier.
- * @param key_size
+ * @param[out] key_size where to store the size of the extracted
+ *             key; pass NULL here if it is not needed
  * @retval not NULL Success
  * @retval     NULL Memory Allocation error
  */
```

src/box/key_def.h
ligurio: Looks much better! Updated, thanks.

src/lib/core/fiber.h
ligurio: Updated, thanks!

src/lua/utils.h
ligurio: > Please, separate parameters from the brief description with an empty line.

Done

> Isn't is redundant? AFAIS, we usually omit it in @param.

removed for consistency with other comments

> How about some str description? 'Lua code to load and run' or kinda.

```diff
@@ -391,8 +391,9 @@ luaT_call(lua_State *L, int nargs, int nreturns);
 /**
  * Like luaL_dostring(), but in case of error sets fiber diag instead
  * of putting error on stack.
- * @param L is a Lua state
- * @param str
+ *
+ * @param L Lua state
+ * @param str string with Lua code to load and run
  */
 int
 luaT_dostring(struct lua_State *L, const char *str);
```

src/box/box.h
ligurio: Updated

src/lib/core/coio.h
ligurio: removed dashes

src/box/tuple.h
ligurio: Updated, thanks

ligurio:
```diff
--- a/src/box/tuple.h
+++ b/src/box/tuple.h
@@ -334,12 +334,12 @@ box_tuple_update(box_tuple_t *tuple, const char *expr, const char *expr_end);
 
 /**
  * Upsert tuple data to a space.
- * @param tuple  Tuple to upsert.
+ * @param tuple  Tuple to update or insert.
  * @param expr
  * @param expr_end
  *
- * @retval  NULL The tuple was not upserted.
- * @retval  box_tuple_t The tuple successfully upserted to a space.
+ * @retval  NULL The tuple was not updated or inserted.
+ * @retval  box_tuple_t The tuple successfully updated or inserted to a space.
  */
 box_tuple_t *
 box_tuple_upsert(box_tuple_t *tuple, const char *expr, const char *expr_end);
```

src/box/tuple.h
ligurio: > I'm a markdown guy and I like when paragraphs are separated by an empty line. 

I would left as is to keep consistent with other comments, at least in this file

> Why two whitespaces?

removed

src/box/box.h
ligurio: placed to a separate line:

```diff
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -626,7 +626,8 @@ API_EXPORT int
 box_session_push(const char *data, const char *data_end);
 
 /**
- * \return current session's unique monotonic identifier \sa box.session.id
+ * \return current session's unique monotonic identifier
+ * \sa box.session.id
  */
 API_EXPORT uint64_t
 box_session_id(void);
```

ligurio:
@Totktonada  some comments left unanswered, I'll process them later

.github/workflows/publish-module-api-doc.yaml
ligurio: Updated:

```diff
--- a/.github/workflows/publish-module-api-doc.yaml
+++ b/.github/workflows/publish-module-api-doc.yaml
@@ -1,10 +1,10 @@
-name: Publish module API documentation
+name: Build and publish module API documentation
 
 on:
   push:
     branches: [master]
   pull_request:
-    types: [opened, reopened, synchronize, labeled]
+    types: [opened, reopened, synchronize]
   workflow_dispatch:
 
```

src/box/tuple.h
Totktonada: > > I'm a markdown guy and I like when paragraphs are separated by an empty line. 
>
> I would left as is to keep consistent with other comments, at least in this file

Fair enough.

src/box/key_def.h
Totktonada: Filed #8513.

PR #8388, opened 2023-03-01
msgpack: fix unsafe extension decoding

CuriousGeorgiy:
Closes tarantool/security#73

### DIFF
--- a/src/box/mp_error.cc
+++ b/src/box/mp_error.cc
@@ -514,16 +514,25 @@ error_unpack_unsafe(const char **data)
 	return err;
 }
 
+struct error *
+error_unpack(const char **data, uint32_t len)
+{
+	const char *end = *data + len;
+	const char *check = *data;
+	if (mp_check(&check, end) != 0 || check != end)
+		return NULL;
+	return error_unpack_unsafe(data);
+}
+
 int
 mp_validate_error(const char *data, uint32_t len)
 {
-	const char *end = data + len;
-	struct error *err = error_unpack_unsafe(&data);
+	struct error *err = error_unpack(&data, len);
 	if (err != NULL) {
 		/* A hack to delete the error. */
 		error_ref(err);
 		error_unref(err);
-		return data != end;
+		return 0;
 	} else {
 		return 1;
 	}
--- a/src/lib/core/decimal.c
+++ b/src/lib/core/decimal.c
@@ -479,11 +479,23 @@ static_assert(DECIMAL_DIGIT_CAPACITY % 2 == 1 ||
 decimal_t *
 decimal_unpack(const char **data, uint32_t len, decimal_t *dec)
 {
+	/*
+	 * MsgPack extensions have length greater or equal than 1 by
+	 * specification.
+	 */
+	assert(len > 0);
+
 	int32_t scale;
+	const char *end = *data + len;
 	const char *svp = *data;
-	if (mp_typeof(**data) == MP_UINT) {
+	enum mp_type type = mp_typeof(**data);
+	if (type == MP_UINT) {
+		if (mp_check_uint(*data, end) > 0)
+			return NULL;
 		scale = mp_decode_uint(data);
-	} else if (mp_typeof(**data) == MP_INT) {
+	} else if (type == MP_INT) {
+		if (mp_check_int(*data, end) > 0)
+			return NULL;
 		scale = mp_decode_int(data);
 	} else {
 		return NULL;
--- a/src/lib/core/mp_interval.c
+++ b/src/lib/core/mp_interval.c
@@ -99,13 +99,34 @@ interval_pack(char *data, const struct interval *itv)
 }
 
 struct interval *
-interval_unpack(const char **data, struct interval *itv)
+interval_unpack(const char **data, uint32_t len, struct interval *itv)
 {
+	/*
+	 * MsgPack extensions have length greater or equal than 1 by
+	 * specification.
+	 */
+	assert(len > 0);
+
+	const char *end = *data + len;
+	uint32_t count = mp_load_u8(data);
+	len -= sizeof(uint8_t);
+	if (count > 0 && len < 2)
+		return NULL;
+
 	memset(itv, 0, sizeof(*itv));
-	uint32_t len = mp_load_u8(data);
-	for (uint32_t i = 0; i < len; ++i) {
+	for (uint32_t i = 0; i < count; ++i) {
 		uint32_t field = mp_load_u8(data);
 		int32_t value;
+		enum mp_type type = mp_typeof(**data);
+		if (type == MP_UINT) {
+			if (mp_check_uint(*data, end) > 0)
+				return NULL;
+		} else if (type == MP_INT) {
+			if (mp_check_int(*data, end) > 0)
+				return NULL;
+		} else {
+			return NULL;
+		}
 		if (mp_read_int32(data, &value) != 0)
 			return NULL;
 		switch (field) {
@@ -142,6 +163,8 @@ interval_unpack(const char **data, struct interval *itv)
 			return NULL;
 		}
 	}
+	if (*data != end)
+		return NULL;
 	return itv;
 }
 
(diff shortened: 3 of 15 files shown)


src/lib/core/decimal.c
```diff
@@ -480,10 +480,16 @@ decimal_t *
 decimal_unpack(const char **data, uint32_t len, decimal_t *dec)
 {
 	int32_t scale;
+	const char *end = *data + len;
 	const char *svp = *data;
-	if (mp_typeof(**data) == MP_UINT) {
+	enum mp_type type = mp_typeof(**data);
```
locker, line 485: Shouldn't we check that `len` > 0 before dereferencing `*data`?

src/lib/core/mp_datetime.c
```diff
@@ -69,7 +70,7 @@ datetime_unpack(const char **data, uint32_t len, struct datetime *date)
 	if (len == 0)
 		goto return_validated;
 
-	if (len != SZ_TAIL) {
+	if (len != SZ_TAIL || *data + len != end) {
```
locker, line 73: `*data + len` always equals `end` AFAICS. Why do you need this check?

src/lib/core/mp_interval.c
```diff
@@ -99,13 +99,28 @@ interval_pack(char *data, const struct interval *itv)
 }
 
 struct interval *
-interval_unpack(const char **data, struct interval *itv)
+interval_unpack(const char **data, uint32_t len, struct interval *itv)
 {
+	const char *end = *data + len;
+	uint32_t count = mp_load_u8(data);
```
locker, line 105: AFAIU you should check that `len` is > 0 before loading `data`.

locker (changes requested):

src/lib/core/decimal.c
CuriousGeorgiy: No, MsgPack extensions have length greator equal than `1` by specification, and we check this prior to validating an extension.

src/lib/core/mp_interval.c
CuriousGeorgiy: Ditto.

src/box/mp_error.cc
```diff
@@ -466,7 +466,17 @@ error_to_mpstream(const struct error *error, struct mpstream *stream)
 }
 
 struct error *
-error_unpack_unsafe(const char **data)
+error_unpack_unsafe(const char **data, uint32_t len)
+{
+	const char *end = *data + len;
+	const char *check = *data;
+	if (mp_check(&check, end) != 0 || check != end)
+		return NULL;
```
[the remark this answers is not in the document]
CuriousGeorgiy: Because it performs decoding of unsafe MsgPack: the prefix refers to the source of the MsgPack.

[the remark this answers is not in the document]
locker: This looks confusing. We usually add `_safe` or `_unsafe` prefix to indicate that the function is safe or unsafe, not the arguments it operates on. Please rework the other way round: `error_unpack` should check the data and then call `error_unpack_unsafe`, not vice versa (BTW this is how it is now).

src/lib/core/decimal.c
locker: Then it'd be nice to add an assertion here.

src/lib/core/mp_datetime.c
CuriousGeorgiy: Indeed, this is redundant, looks like I forgot to drop this change — sorry for the inconvenience. Fixed.

CuriousGeorgiy:
@locker added a changelog, though I don't understand it's purpose with our current security issue workflow — the user won't be able to see the GitHub ticket related to the changelog anyways.

locker:
> @locker added a changelog, though I don't understand its purpose with our current security issue workflow — the user won't be able to see the GitHub ticket related to the changelog anyways.

I guess that's why @Totktonada suggested using GitHub security advisories, e.g. see https://github.com/tarantool/tarantool/security/advisories/GHSA-74jr-2fq7-vp42.

src/box/mp_error.cc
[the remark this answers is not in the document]
CuriousGeorgiy: > (BTW this is how it is now)

No, I just followed the pattern we have now:
https://github.com/tarantool/tarantool/blob/c561202d732650f263957d6c8ee348be307beeda/src/box/mp_error.h#L55-L72

src/lib/core/mp_interval.c
CuriousGeorgiy: Added an assertion.

src/box/mp_error.cc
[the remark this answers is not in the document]
CuriousGeorgiy: Renamed as you suggest.

locker (approved):

locker:
@CuriousGeorgiy Tests [fail](https://github.com/tarantool/tarantool/actions/runs/4345080573/jobs/7589292179) on some platforms. Please fix.

```
Duration of failed tests (seconds):
*   0.01 box/gh-6431-crash-while-encoding-error.test.lua              
*   0.68 app-tap/msgpackffi.test.lua                                  
*   0.90 box-tap/extended_error.test.lua
```

CuriousGeorgiy:
@locker I forgot to change `error_unpack_unsafe` to `error_unpack` in `exports`.

ImeevMA:
Sorry, I also plan to review some of these changes. Please wait a bit.

src/lib/core/mp_interval.c
```diff
@@ -99,13 +99,34 @@ interval_pack(char *data, const struct interval *itv)
...
+
+	const char *end = *data + len;
+	uint32_t count = mp_load_u8(data);
+	len -= sizeof(uint8_t);
+
 	memset(itv, 0, sizeof(*itv));
-	uint32_t len = mp_load_u8(data);
-	for (uint32_t i = 0; i < len; ++i) {
+	for (uint32_t i = 0; i < count; ++i) {
+		if (len < 2)
```
ImeevMA, line 116: Why are you checking `len` here? I mean it doesn't change anymore, so why not check it before the loop?

src/lib/core/mp_interval.c
```diff
@@ -99,13 +99,34 @@ interval_pack(char *data, const struct interval *itv)
...
+			return NULL;
 		uint32_t field = mp_load_u8(data);
 		int32_t value;
+		enum mp_type type = mp_typeof(**data);
+		if (type == MP_UINT) {
+			if (mp_check_uint(*data, end) > 0)
```
ImeevMA, line 122: The value is obtained using `mp_read_int32()` which checks the type. Do you really need extra validation for MP_INT and MP_UINT?

ImeevMA (changes requested):
Hi! Thank you for the patches! I have a couple of comments, please see them below.

src/lib/core/mp_interval.c
CuriousGeorgiy: > The value is obtained using mp_read_int32() which checks the type.

No it doesn't it returns `-1` if the MsgPack is not an integer https://github.com/tarantool/msgpuck/blob/28a7421cf7fa538a0180c79bd9c12ee0dd8c12eb/msgpuck.h#L2275-L2277

ImeevMA: What about this swith? I believe it checks type of the value: https://github.com/tarantool/msgpuck/blob/28a7421cf7fa538a0180c79bd9c12ee0dd8c12eb/msgpuck.h#L2253

CuriousGeorgiy: So, what does it do when the type is not an integer type?..

ImeevMA: I believe you can just return `itv` if `count` == 0. I think this looks better than an unnecessary check of two conditions.

src/lib/core/mp_interval.c
ImeevMA: I believe you can just return `itv` if `count` == 0. I think this looks better than an unnecessary check of two conditions.

src/lib/core/mp_interval.c
ImeevMA: I see. True, I was wrong.

src/lib/core/mp_interval.c
CuriousGeorgiy: No, we must make sure that all of the provided length is used.

locker:

PR #8389, opened 2023-03-01
test: add cleanup to digest_crc32_recording_test

ylobankov:
Add missing `server:drop()` to digest_crc32_recording_test.lua to avoid running tarantool processes after the test.


igormunkin (approved):
@ylobankov, thanks for noticing! LGTM as trivial.

PR #8390, opened 2023-03-02
test: fix gh_7974_force_recovery_bugs flaky test

locker:
The test fails with

```
 fail | 2023-03-01 15:54:30.550 [3724975] main/103/server_instance.lua F> can't initialize storage: unlink, called on fd 63,  aka unix/:(socket), peer of unix/:(socket): Address already in use
```

We fixed a similar issue in commit 3d3e9dea137e by using unique instance names. Let's do the same here.

Fixes commit b1095c1c3254
Follow-up #7974


ylobankov (approved):
LGTM

locker:

PR #8396, opened 2023-03-02
ci: eliminate 1.10 related logic, introduce 3.x

ylobankov:
- ci: not to push sources on `*-entrypoint` tag
- ci: eliminate 1.10 related logic for workflows 
- ci: introduce 3.x related logic for workflows

### DIFF
--- a/.github/workflows/source.yml
+++ b/.github/workflows/source.yml
@@ -2,8 +2,8 @@ name: source
 
 on:
   push:
-    tags:
-      - '**'
+    tags-ignore:
+      - '**-entrypoint'
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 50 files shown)


.github/workflows/source.yml
```diff
@@ -2,8 +2,8 @@ name: source
 
 on:
   push:
-    tags:
-      - '**'
+    tags-ignore:
+      - '**-entrypoint'
```
Totktonada, line 6: Shouldn't we do the same for packages? I checked `centos_8.yml` and it doesn't contain the exception.

Totktonada: I see now, it is skipped in `.github/actions/pack-and-deploy/action.yml`.

ylobankov: We have such an exception here https://github.com/tarantool/tarantool/blob/master/.github/actions/pack-and-deploy/action.yml#L24, but your variant looks better. There is no need to run testing on an `*-entrypoint` tag.

Totktonada: Anyway, this is not a problem of this patchset, so I'll resolve the comment.

Totktonada (approved):
I see no conceptual problems.

I have not performed any testing or consistency check of the changes, my review is purely visual.

PR #8288, opened 2023-02-09
Disable JIT engine on macOS platforms by default

igormunkin:
This PR contains the following changes:

1. Several Tarantool tests are adjusted (and even rewritten) to the Tarantool builds with JIT being turned off.
2. LuaJIT compiler is turned off by default on macOS and `LUAJIT_JIT_STATUS` CMake option is introduced
3. Testing routine is enabled back for macOS/M1 workflows in Tarantool CI.

Closes #8252

PR #8406, opened 2023-03-06
test: add rule to ignore Makefile

ImeevMA:
This patch adds a rule to ignore the Makefile on the path test/*/*.


igormunkin (approved):
@ImeevMA, my bad, thanks for noticing! LGTM, as trivial.

For the history: this patch follows up #8288.

igormunkin:
Also backported to 2.11 in scope of df062b6 (2.11.0-rc1-31-gdf062b6aa).

PR #8409, opened 2023-03-06
test: fix fail of gh_4669_applier_reconnect_test

Serpentian:
In the test we start replicas only with master in box.cfg.replication. We cannot use bootstrap_strategy = 'auto' mode, which is default, as it properly works only when all participants of the cluster are listed in replication parameter. Sometimes, when one replica connects to the master, the other one has already successfully joined, so the first replica sees in ballot, that it doesn't have all nodes from cluster in box.cfg.replication and fails to start.

Let's use 'legacy' boostrap strategy for now.

Closes tarantool/tarantool-qa#310

NO_DOC=test-fix
NO_CHANGELOG=test-fix

### DIFF
--- a/test/replication-luatest/gh_4669_applier_reconnect_test.lua
+++ b/test/replication-luatest/gh_4669_applier_reconnect_test.lua
@@ -8,6 +8,7 @@ g.before_each(function()
     g.cluster = cluster:new({})
     g.master = g.cluster:build_server({alias = 'master'})
     local box_cfg = {
+        bootstrap_strategy = 'legacy',
         replication = server.build_listen_uri('master'),
     }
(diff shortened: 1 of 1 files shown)


Serpentian:
As far as we don't have any other options to bootstrap a cluster with not all nodes listed `box.cfg.replication` (waiting for https://github.com/tarantool/tarantool/issues/7999 to be closed), let's use legacy mode.

We'll have to bootstrap cluster properly as soon as the ticket is closed in all tests, which use `legacy` mode.

sergepetrenko (approved):
Thanks for the fix!
LGTM.

sergepetrenko:
Cherry-picked to 2.11: 6bb2e9f7f0a7ae32650f35deea8ec4323167f483

PR #8410, opened 2023-03-06
Rework parsing of transaction and DDL queries

ImeevMA:
This patch-set reworks the parsing of transaction-related queries and DDL queries. These patches will help us separate the parser from the VDBE construction.

### DIFF
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -1705,23 +1705,20 @@ sql_code_drop_table(struct Parse *parse_context, struct space *space,
 void
 sql_drop_table(struct Parse *parse_context)
 {
-	struct drop_entity_def drop_def = parse_context->drop_table_def.base;
-	assert(drop_def.base.alter_action == ALTER_ACTION_DROP);
-	struct SrcList *table_name_list = drop_def.base.entity_name;
 	struct Vdbe *v = sqlGetVdbe(parse_context);
-	bool is_view = drop_def.base.entity_type == ENTITY_TYPE_VIEW;
-	assert(is_view || drop_def.base.entity_type == ENTITY_TYPE_TABLE);
+	bool is_view = parse_context->type == PARSE_TYPE_DROP_VIEW;
+	assert(is_view || parse_context->type == PARSE_TYPE_DROP_TABLE);
 	sqlVdbeCountChanges(v);
 	assert(!parse_context->is_aborted);
-	assert(table_name_list->nSrc == 1);
-	const char *space_name = table_name_list->a[0].zName;
+	assert(parse_context->src_list->nSrc == 1);
+	const char *space_name = parse_context->src_list->a[0].zName;
 	struct space *space = space_by_name(space_name);
 	if (space == NULL) {
-		if (!drop_def.if_exist) {
+		if (!parse_context->drop_object.if_exists) {
 			diag_set(ClientError, ER_NO_SUCH_SPACE, space_name);
 			parse_context->is_aborted = true;
 		}
-		goto exit_drop_table;
+		return;
 	}
 	/*
 	 * Ensure DROP TABLE is not used on a view,
@@ -1731,19 +1728,15 @@ sql_drop_table(struct Parse *parse_context)
 		diag_set(ClientError, ER_DROP_SPACE, space_name,
 			 "use DROP TABLE");
 		parse_context->is_aborted = true;
-		goto exit_drop_table;
+		return;
 	}
 	if (!is_view && space->def->opts.is_view) {
 		diag_set(ClientError, ER_DROP_SPACE, space_name,
 			 "use DROP VIEW");
 		parse_context->is_aborted = true;
-		goto exit_drop_table;
+		return;
 	}
 	sql_code_drop_table(parse_context, space, is_view);
-
- exit_drop_table:
-	if (table_name_list != parse_context->src_list)
-		sqlSrcListDelete(table_name_list);
 }
 
 /**
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -326,28 +326,26 @@ error:
 }
 
 void
-sql_create_column_start(struct Parse *parse)
+sql_create_column_start(struct Parse *parse, struct Token *name,
+			enum field_type type)
 {
-	struct create_column_def *create_column_def = &parse->create_column_def;
-	struct alter_entity_def *alter_entity_def =
-		&create_column_def->base.base;
-	assert(alter_entity_def->entity_type == ENTITY_TYPE_COLUMN);
-	assert(alter_entity_def->alter_action == ALTER_ACTION_CREATE);
 	struct space *space = parse->create_table_def.new_space;
 	bool is_alter = space == NULL;
 	if (is_alter) {
-		const char *space_name =
-			alter_entity_def->entity_name->a[0].zName;
+		const char *space_name = parse->src_list->a[0].zName;
 		space = space_by_name(space_name);
 		if (space == NULL) {
 			diag_set(ClientError, ER_NO_SUCH_SPACE, space_name);
-			goto tnt_error;
+			parse->is_aborted = true;
+			return;
 		}
 		space = sql_shallow_space_copy(parse, space);
-		if (space == NULL)
-			goto tnt_error;
+		if (space == NULL) {
+			parse->is_aborted = true;
+			return;
+		}
 	}
-	create_column_def->space = space;
+	parse->space = space;
 	struct space_def *def = space->def;
 	assert(def->opts.is_ephemeral);
 
@@ -355,16 +353,18 @@ sql_create_column_start(struct Parse *parse)
 	if ((int)def->field_count + 1 > SQL_MAX_COLUMN) {
 		diag_set(ClientError, ER_SQL_COLUMN_COUNT_MAX, def->name,
 			 def->field_count + 1, SQL_MAX_COLUMN);
-		goto tnt_error;
+		parse->is_aborted = true;
+		return;
 	}
 #endif
 
 	struct region *region = &parse->region;
-	struct Token *name = &create_column_def->base.name;
 	char *column_name =
 		sql_normalized_name_region_new(region, name->z, name->n);
-	if (column_name == NULL)
-		goto tnt_error;
+	if (column_name == NULL) {
+		parse->is_aborted = true;
+		return;
+	}
 
 	/*
 	 * Format can be set in Lua, then exact_field_count can be
@@ -385,14 +385,8 @@ sql_create_column_start(struct Parse *parse)
 	 */
 	column_def->nullable_action = ON_CONFLICT_ACTION_DEFAULT;
 	column_def->is_nullable = true;
-	column_def->type = create_column_def->type_def->type;
+	column_def->type = type;
 	def->field_count++;
-
-	sqlSrcListDelete(alter_entity_def->entity_name);
-	return;
-tnt_error:
-	parse->is_aborted = true;
-	sqlSrcListDelete(alter_entity_def->entity_name);
 }
 
 static void
@@ -401,7 +395,7 @@ vdbe_emit_create_constraints(struct Parse *parse, int reg_space_id);
 void
 sql_create_column_end(struct Parse *parse)
 {
-	struct space *space = parse->create_column_def.space;
+	struct space *space = parse->space;
 	assert(space != NULL);
 	struct space_def *def = space->def;
 	struct field_def *field = &def->fields[def->field_count - 1];
@@ -462,8 +456,8 @@ void
 sql_column_add_nullable_action(struct Parse *parser,
 			       enum on_conflict_action nullable_action)
 {
-	assert(parser->create_column_def.space != NULL);
-	struct space_def *def = parser->create_column_def.space->def;
+	assert(parser->space != NULL);
+	struct space_def *def = parser->space->def;
 	assert(def->field_count > 0);
 	struct field_def *field = &def->fields[def->field_count - 1];
 	if (field->nullable_action != ON_CONFLICT_ACTION_DEFAULT &&
@@ -496,7 +490,7 @@ sql_column_add_nullable_action(struct Parse *parser,
 void
 sqlAddDefaultValue(Parse * pParse, ExprSpan * pSpan)
 {
-	struct space *p = pParse->create_column_def.space;
+	struct space *p = pParse->space;
 	if (p != NULL) {
 		assert(p->def->opts.is_ephemeral);
 		struct space_def *def = p->def;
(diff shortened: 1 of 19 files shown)


Gerold103 (commented):
Thanks for the patch! I stopped at "sql: rework parsing of UNIQUE constraints". Will come back to continue. Unless you will decide to do any big changes.

src/box/sql/sqlInt.h
```diff
@@ -2062,6 +2062,10 @@ struct Parse {
 	 */
 	struct create_table_def create_table_def;
 	struct create_column_def create_column_def;
+	/** Parsed statement type. */
+	enum parse_type type;
+	/** Savepoint description for savepoint-related statements. */
+	struct sql_parse_savepoint savepoint;
```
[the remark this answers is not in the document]
ImeevMA: As far as I understand, when I refactor all statements in this way, I will be able to move all these fields into a new structure, something like `struct sql_parse`. In this case, `struct Parse` will only be used to build VDBE. I suggest to add all these fields to `struct Parse` until then.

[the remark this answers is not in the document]
ImeevMA: Values of `struct Token` use static memory. The `z` field usually points to some place in the parsed SQL statement.

src/box/sql/tokenize.c
```diff
@@ -474,6 +474,22 @@ new_xmalloc(size_t n)
 	return xmalloc(n);
 }
 
+/**
+ * Finish processing table properties for CREATE TABLE or ALTER TABLE ADD COLUMN
+ * statement.
```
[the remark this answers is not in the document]
ImeevMA: No, this function is only used for `CREATE TABLE` and `ALTER TABLE ADD COLUMN` statements.

test/sql-luatest/gh_8100_parser_separation_test.lua
```diff
@@ -0,0 +1,36 @@
...
+
+g.after_all(function()
+    g.server:stop()
+end)
+
+--
+-- Now VDBE for FOREIGN KEY creation is built after column creation, NULL and
+-- NOT NULL constraints, DEFAULT clause, COLLATE clause, PRIMARY KEY constraint,
+-- UNIQUE constraints, CHECK constraints. However, since we no longer have
+-- uniqueness for the constraint name, we can only check that the FOREIGN KEY is
```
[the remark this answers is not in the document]
ImeevMA: We can divide all named constraints into 4 categories:
1) Indexes (PRIMARY KEY, UNIQUE).
2) Field FK.
3) Tuple FK.
4) Field CHECK and  Tuple CHECK.

Within each category, it is forbidden to create constraints with the same name. There is no such restriction between categories.

There is also no such restriction formally for field CHECK constraint and tuple CHECK constraint, but they both create a function with the same name, which is not allowed.

For example, the following query works fine:
```
CREATE TABLE t (i INT CONSTRAINT a PRIMARY KEY constraint a REFERENCES t(i), constraint a CHECK(i < 20), constraint a
```

src/box/sql/tokenize.c
```diff
@@ -482,6 +482,11 @@ static int
 sql_finish_table_properties(struct Parse *parse)
 {
 	assert(!parse->is_aborted);
+	for (uint32_t i = 0; i < parse->check_list.n; ++i) {
+		sql_create_check_contraint(parse, &parse->check_list.a[i]);
+		if (parse->is_aborted)
+			return -1;
+	}
```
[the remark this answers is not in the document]
ImeevMA: No, this function adds a CHECK constraint defined in the `CREATE TABLE` and `ALTER TABLE ADD COLUMN` statements. The `ALTER TABLE ADD CONSTRAINT CHECK` statement is handled separately.

ImeevMA:
@Gerold103 thank you for the review! I plan to push a new version after the review is complete. However, I still answered some questions, because I think they are very important for a general understanding of my immediate goal.

src/box/sql/tokenize.c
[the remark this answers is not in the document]
Gerold103: Then this function name is not very good. It has a very generic name, but is called only for a couple of DDL statements, while not called for others. And it is not possible to tell for which it should be called except by reading the comment (which tend to outdate much faster than code).

src/box/sql/sqlInt.h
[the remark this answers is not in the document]
Gerold103: How about moving them there now? Introduce `struct sql_parse` right in this patchset and store it as a member in `struct Parse` then. Shouldn't complicate things? But would keep `struct Parse` "simple".

src/box/sql/parse_def.c
```diff
@@ -92,6 +92,15 @@ last_column_name(struct Parse *parse)
 	return &parse->create_column_def.base.name;
 }
 
+/** Return the name of the table from CREATE TABLE. */
+static const char *
+new_table_name(struct Parse *parse)
```
Gerold103, line 97: `const Parse`.

src/box/sql/parse_def.c
```diff
@@ -235,3 +244,52 @@ sql_parse_add_unique(struct Parse *parse, struct SrcList *table_name,
...
+primary_key_fill(struct Parse *parse, const struct Token *name,
+		 struct ExprList *cols)
+{
+	if (parse->primary_key.cols != NULL) {
+		const char *space_name;
+		if (parse->src_list == NULL)
+			space_name = new_table_name(parse);
+		else
+			space_name = parse->src_list->a[0].zName;
+		diag_set(ClientError, ER_CREATE_SPACE, space_name,
```
Gerold103, line 289: `new_table_name()` can return `NULL` and then this `diag_set` would either crash or print space name as "nil" or "null".

src/box/sql/parse_def.c
```diff
@@ -235,3 +244,52 @@ sql_parse_add_unique(struct Parse *parse, struct SrcList *table_name,
...
+			 "primary key has been already declared");
+		parse->is_aborted = true;
+		sql_expr_list_delete(cols);
+		sql_expr_list_delete(parse->primary_key.cols);
```
Gerold103, line 293: I suggest to either free this (parse->primary_key) in one place - when Parse is being destroyed. Or at least nullify this member here after you freed it. To ensure it won't be used after free.

test/sql/on-conflict.result
```diff
@@ -71,7 +71,8 @@ box.execute("CREATE TABLE t (id INTEGER PRIMARY KEY, a INTEGER CHECK (a > 5) ON
 box.execute("CREATE TABLE te17 (s1 INT NULL PRIMARY KEY NOT NULL);")
 ---
 - null
-- Primary index of space 'TE17' can not contain nullable parts
+- 'Failed to execute SQL statement: NULL declaration for column ''S1'' of table ''TE17''
+  has been already set to ''none'''
```
Gerold103, line 75: Something is wrong with this error message. What is "none" here?

src/box/sql/parse_def.c
```diff
@@ -306,3 +306,31 @@ sql_parse_create_index(struct Parse *parse, struct Token *table_name,
...
+
+/** Set the AUTOINCREMENT column name. */
+static void
+autoincrement_add(struct Parse *parse, struct Expr *column_name)
+{
+	if (parse->has_autoinc) {
+		diag_set(ClientError, ER_SQL_SYNTAX_WITH_POS, parse->line_count,
+			 parse->line_pos,
+			 "table must feature at most one AUTOINCREMENT field");
+		parse->is_aborted = true;
```
Gerold103, line 348: You need to delete `column_name` here.

src/box/sql/prepare.c
```diff
@@ -208,6 +208,8 @@ sql_parser_destroy(Parse *parser)
 		sql_xfree(parser->check_list.a);
 	if (parser->unique_list.n != 0)
 		sql_xfree(parser->unique_list.a);
+	if (parser->autoinc_name != NULL)
+		sql_expr_delete(parser->autoinc_name);
```
Gerold103, line 212: `sql_expr_delete()` already checks for `NULL`.

src/box/sql/build.c
```diff
@@ -558,7 +552,7 @@ field_def_create_for_pk(struct Parse *parser, struct field_def *field,
 void
 sqlAddPrimaryKey(struct Parse *pParse)
 {
-	struct space *space = pParse->create_column_def.space;
+	struct space *space = pParse->space;
```
Gerold103, line 550: I stand by what I said in another comment. I think this is wrong to merge such generic members back into Parse/sql_parse. This is already problematic to read, when I simply don't understand who uses `Parse.space`. Previously it was clear from the sub-struct names like `create_column_def`. IMO would be better to keep separate struct per statement type. And maybe even have some members duplicated in them, but by looking at each such struct I would immediately understand what data it needs.

As an example you can look at `xrow.h` and its structs `register_request`, `subscribe_request`, `subscribe_response`, `join_request` and others. They have a few duplicate members, but in each request it is crystal clear what it operates with.

Try to open `struct Parse` and tell which members are going to be used by `ALTER TABLE ADD COLUMN`. Among these: `autoname_i`, `triggered_space`, `column_list`, `foreign_key_list`, `check_list`, `unique_list`, `primary_key`, `create_index`, `autoinc_name`, `src_list`, `space`, `has_autoinc`, `autoinc_fieldno`? Personally, I have no idea ATM. You probably know because you work with this right now, but ask anybody else and they wouldn't be as confused as me. I know some of these members were here before this patchset, but now it feels we are loosing control again.

src/box/sql/build.c
```diff
@@ -558,7 +552,7 @@ field_def_create_for_pk(struct Parse *parser, struct field_def *field,
...
 	assert(space != NULL || pParse->src_list != NULL);
 	if (space == NULL)
```
Gerold103, line 552: Addition to `parse_type` - I see there is an existing `enum ast_type`. I think this one should be renamed to `sql_stmt_type/sql_ast_type` and used instead of `parse_type`.

src/box/sql/build.c
```diff
@@ -558,7 +552,7 @@ field_def_create_for_pk(struct Parse *parser, struct field_def *field,
...
 		return sql_create_index(pParse, &pParse->primary_key.name,
```
Gerold103, line 553: Addition to `struct Parse / struct sql_parse` and naming of all those new `sql_parse_...()` functions. The mega confusing part is that we somewhy use verb as a struct name. Verbs are for function names, not for struct names. Functions like `sql_parse_...` look like they want to "parse" something. Not like methods of an `sql_parse` object.

If you are going to introduce a new struct and you are preparing all those new functions `sql_parse_...` for it, I highly recommend to rethink the naming. Use `sql_ast` or `sql_query` or `sql_stree` (syntax tree) or `sql_ptree` (parser tree) or alike.

Gerold103 (commented):
I stopped at `sql: rework parsing of ADD COLUMN statement`. Will continue later.

src/box/sql/parse_def.c
```diff
@@ -334,3 +360,48 @@ sql_parse_table_autoincrement(struct Parse *parse, struct Expr *column_name)
...
+		action_str = on_conflict_action_strs[c->action];
+	} else if ((on_conflict != ON_CONFLICT_ACTION_ABORT ||
+		    action != ON_CONFLICT_ACTION_NONE) &&
+		   action != on_conflict) {
+		action_str = on_conflict_action_strs[ON_CONFLICT_ACTION_NONE];
+	} else {
+		c->action = action;
+		c->is_action_set = true;
+		return;
+	}
```
Gerold103, line 392: In `struct Parse` I see this:
```C
	/** Description of created FOREIGN KEY constraints. */
	struct sql_parse_column_list column_list;
	/** Description of created columns. */
	struct sql_parse_foreign_key_list foreign_key_list;
```
It looks like the comments are swapped.

src/box/sql/parse_def.h
```diff
@@ -235,6 +235,16 @@ struct sql_parse_column_list {
...
 
+/** Description of the table being created. */
+struct sql_parse_table {
+	/** Table name. */
+	struct Token name;
+	/** Table engine name. */
+	struct Token engine_name;
+	/** IF NOT EXISTS flag. */
+	bool if_not_exists;
+};
```
Gerold103, line 262: Further on the point about struct-per-statement - I would expect this struct to have column list, index list, fk list, etc. Not store that in the generic `struct Parse`. I know it wasn't like that even before, but it would be very good to do so.

src/box/sql/tokenize.c
```diff
@@ -581,6 +581,16 @@ sql_finish_parsing(struct Parse *parse)
...
 		break;
+	case PARSE_TYPE_CREATE_TRIGGER:
+		parse->initiateTTrans = true;
+		sql_trigger_begin(parse);
+		if (parse->is_aborted)
+			return;
+		sql_trigger_finish(parse, parse->create_trigger.step,
+				   &parse->create_trigger.all);
+		if (parse->is_aborted)
+			return;
```
Gerold103, line 592: Ditto. Same with `PARSE_TYPE_RENAME_TABLE`. Check other places too.

src/box/sql/parse_def.h
```diff
@@ -655,4 +583,9 @@ void
 sql_parse_drop_table(struct Parse *parse, struct SrcList *table_name,
 		     bool if_exists);
 
+/** Save parsed DROP TRIGGER statement. */
+void
+sql_parse_drop_trigger(struct Parse *parse, struct SrcList *table_name,
+		       bool if_exists);
```
Gerold103, line 589: And where is trigger name?

src/box/sql/trigger.c
```diff
@@ -321,18 +321,13 @@ vdbe_code_drop_trigger(struct Parse *parser, const char *trigger_name,
...
-	bool no_err = drop_def->if_exist;
+	bool no_err = parser->drop_object.if_exists;
 
 	struct Vdbe *v = sqlGetVdbe(parser);
 	sqlVdbeCountChanges(v);
 
-	assert(name->nSrc == 1);
-	const char *trigger_name = name->a[0].zName;
+	assert(parser->src_list->nSrc == 1);
+	const char *trigger_name = parser->src_list->a[0].zName;
```
Gerold103, line 330: Oh no, so `src_list` can be a list of tables or a trigger name? +1 point against merging those request structs and for having one struct per statement with full set of needed members with proper member names.

Gerold103 (commented):
Thanks for the patch! I finished the first review now. While the intention to finally split parsing and code generation is very inspiring, I think it needs more thought about how to isolate the statements from each other. Not merge their members into one place.

Also the I advice to think more about naming (`Parse`, `sql_parse` - verb is a bad name for a struct), about statement types (`parse_type` and `ast_type` are basically the same, I would suggest `sql_ast_type` or `sql_stmt_type`), about function names (specific prefix for statement start which asserts `Parse->type == NONE` and sets a proper type, and another special prefix for changing the started statement like add column or an index to "create table", add "check" to "add column", etc - all these functions should have assertions about statement type, some could be used for multiple types and then should assert the type to be equal to either of ok-types).

ImeevMA:
This PR seems to be unnecessary anymore, since it was reworked in #12989

PR #8414, opened 2023-03-06
relay: send heartbeats on top of replication stream

### LINKED ISSUE #7515: A syncing node may prevent others from deposing a hung leader
Now we enable Raft before a recovered node finishes syncing with others.
This means that such a node may vote for anyone, even some node whose vclock is not the biggest.
But the most important problem is that such a node broadcasts its "leader_is_seen" message to everyone. So if a leader hangs while some node is syncing with it, this node will tell everyone the leader is alive and will prevent the cluster from choosing a new not hung leader. The issue will only resolve when the node finishes syncing and sees that the leader is actually hung.

Another possible solution would be not to track who's leader until sync is done.

sergepetrenko:
There was a problem with the leader's relay continuing to ping the remote followers even when the leader's tx thread is hung. This tricked the followers into thinking the leader is alive and well, even though it couldn't serve any new requests.

The problem was partially fixed by commit 56571d83172f ("raft: make followers notice leader hang"): that commit made relay thread stop sending heartbeats in case tx thread is unresponsive.

Up to now we didn't differentiate between heartbeats and data rows: the receipt of both was considered a sign the master is alive. So if some replicas are not up to date with the master, they will continue thinking it's alive until they are fully synced and notice there are no more heartbeats from it.

In order to fix this, stop treating all data as heartbeats and start sending heartbeats on top of an active replication stream.

Add a feature bit for heartbeats in an active replication stream, so that replicas know how to treat the absence of heartbeats: either normal operation (for old versions) or tx thread hang (for new versions).

Closes #7515

@TarantoolBot document
Title: New iproto protocol feature: heartbeats_in_replication_stream

IPROTO_FEATURE_HEARTBEATS_IN_REPLICATION_STREAM = 5 is a feature bit set
when the master sends heartbeats not only in absence of other data to
send, but in between of data rows dispatch as well. This feature is used
by replicas to determine when to expect heartbeats from the master.

The version of the protocol (returned via the IPROTO_VERSION key) is now 5.

The user may set the corresponding feature -
`heartbeats_in_replication_stream` in the required_protocol_features
parameter of `net.box` `connect()` and `new()` methods.

### DIFF
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -171,8 +171,10 @@ struct relay {
 	 * confirmation from the replica.
 	 */
 	struct stailq pending_gc;
-	/** Time when last row was sent to peer. */
+	/** Time when last row was sent to the peer. */
 	double last_row_time;
+	/** Time when last heartbeat was sent to the peer. */
+	double last_heartbeat_time;
 	/** Time of last communication with the tx thread. */
 	double tx_seen_time;
 	/**
@@ -345,7 +347,8 @@ relay_start(struct relay *relay, struct iostream *io, uint64_t sync,
 	relay->need_new_vclock_sync = false;
 	relay->is_sending_tx = false;
 	relay->last_row_time = ev_monotonic_now(loop());
-	relay->tx_seen_time = ev_monotonic_now(loop());
+	relay->tx_seen_time = relay->last_row_time;
+	relay->last_heartbeat_time = relay->last_row_time;
 }
 
 void
@@ -785,8 +788,16 @@ relay_send_heartbeat(struct relay *relay)
 		++relay->last_sent_ack.vclock_sync;
 		RegionGuard region_guard(&fiber()->gc);
 		xrow_encode_relay_heartbeat(&row, &relay->last_sent_ack);
-		row.tm = ev_now(loop());
+		/*
+		 * Do not encode timestamp if this heartbeat is sent in between
+		 * data rows so as to not affect replica's upstream lag.
+		 */
+		if (relay->last_row_time > relay->last_heartbeat_time)
+			row.tm = 0;
+		else
+			row.tm = ev_now(loop());
 		row.replica_id = instance_id;
+		relay->last_heartbeat_time = ev_monotonic_now(loop());
 		relay_send(relay, &row);
 		relay->need_new_vclock_sync = false;
 	} catch (Exception *e) {
@@ -815,7 +826,7 @@ relay_send_heartbeat_on_timeout(struct relay *relay)
 	 * switchover.
 	 */
 	if (!relay->need_new_vclock_sync &&
-	    (now - relay->last_row_time <= replication_timeout ||
+	    (now - relay->last_heartbeat_time <= replication_timeout ||
 	    now - relay->tx_seen_time >= replication_disconnect_timeout()))
 		return;
 	relay_send_heartbeat(relay);
@@ -1198,6 +1209,9 @@ static void
 relay_send_row(struct xstream *stream, struct xrow_header *packet)
 {
 	struct relay *relay = container_of(stream, struct relay, stream);
+	/* Do not send heartbeats during a final join. */
+	if (relay->replica != NULL)
+		relay_send_heartbeat_on_timeout(relay);
 	if (packet->group_id == GROUP_LOCAL) {
 		/*
 		 * We do not relay replica-local rows to other
@@ -1265,8 +1279,5 @@ relay_send_row(struct xstream *stream, struct xrow_header *packet)
 		}
 		relay_send(relay, packet);
 		relay->is_sending_tx = !packet->is_commit;
-		if (!relay->is_sending_tx) {
-			relay_send_heartbeat_on_timeout(relay);
-		}
 	}
 }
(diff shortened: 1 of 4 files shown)


sergepetrenko:
@Gerold103, @grafin,
I'm not sure about adding a new iproto feature for heartbeats in the replication stream.
On the one hand, iproto features were meant to prevent us from trying to guess what features a server supports by looking at its version.
On the other hand, all iproto features added to this day were about user-visible behaviour, while this feature is about replication protocol, which is invisible to the user.

grafin (changes requested):
Thanks for the patch! No new comments from me. But I agree that we shouldn't use iproto features here. Mainly because it's not a feature that can be present or absent with the same peer version, it's not a feature at all.

test/replication-luatest/gh_7515_sync_node_sees_leader_hang_test.lua
```diff
@@ -0,0 +1,99 @@
...
+            box_cfg = box_cfg,
+        }
+    end
+    cg.replica_set:start()
+    cg.unique_filename = server.build_listen_uri('unique_filename')
+    fio.unlink(cg.unique_filename)
+end)
+
+g.after_each(function(cg)
+    --cg.replica_set:drop()
```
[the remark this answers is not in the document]
sergepetrenko: My bad, fixed.
Thanks for noticing!

sergepetrenko:
@grafin, @Gerold103,
My main concern with using peer version instead of iproto features request was that a user would have trouble connecting 2.11 versions before the fix (like 2.11.0-rc1 or rc2) to the newer versions. For example, during an upgrade new tarantool version would think that the master running on old version is hung.

But this seems a minor problem. It only affects Raft, and there's pre-vote, so as long as at least one extra old node remains (and sees the old leader), the new node won't disrupt the old leader.

Switched to using peer version.
Thanks for your answers!

sergepetrenko:
Here are the main diff parts:
<details>

```diff
diff --git a/src/box/applier.cc b/src/box/applier.cc
index b6e7ef39e..27795e536 100644
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1836,14 +1836,16 @@ applier_process_batch(struct cmsg *base)
        struct applier_data_msg *msg = (struct applier_data_msg *)base;
        struct applier *applier = msg->base.applier;
        struct applier_tx *tx;
-       bool has_heartbeats_in_stream = iproto_features_test(
-               &applier->features,
-               IPROTO_FEATURE_HEARTBEATS_IN_REPLICATION_STREAM);
        stailq_foreach_entry(tx, &msg->txs, next) {
                struct applier_tx_row *last_txr =
                        stailq_last_entry(&tx->rows, struct applier_tx_row,
                                          next);
-               if (!has_heartbeats_in_stream) {
+               /*
+                * Tarantool before version 2.11.0 doesn't send heartbeats when
+                * there is data to be sent. Instead each row is treated as
+                * heartbeat.
+                */
+               if (applier->version_id < version_id(2, 11, 0)) {
                        raft_process_heartbeat(box_raft(),
                                               applier->instance_id);
                }
diff --git a/src/box/relay.cc b/src/box/relay.cc
index 3d881b157..a7ae3d6d0 100644
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -792,10 +792,10 @@ relay_send_heartbeat(struct relay *relay)
                 * Do not encode timestamp if this heartbeat is sent in between
                 * data rows so as to not affect replica's upstream lag.
                 */
-               if (relay->last_row_time >= relay->last_heartbeat_time)
-                       row.tm = ev_now(loop());
-               else
+               if (relay->last_row_time > relay->last_heartbeat_time)
                        row.tm = 0;
+               else
+                       row.tm = ev_now(loop());
                row.replica_id = instance_id;
                relay->last_heartbeat_time = ev_monotonic_now(loop());
                relay_send(relay, &row);
```
</summary>

grafin (approved):
After discussing it yesterday over a call, I'm not so sure about iproto features anymore) Anyway, thanks for the fixes. LGTM. But I think we should consider adding some server-server features (ability to negotiate server-server communication not only on peers version).

src/box/relay.cc
```diff
@@ -785,8 +788,16 @@ relay_send_heartbeat(struct relay *relay)
...
 		xrow_encode_relay_heartbeat(&row, &relay->last_sent_ack);
-		row.tm = ev_now(loop());
+		/*
+		 * Do not encode timestamp if this heartbeat is sent in between
+		 * data rows so as to not affect replica's upstream lag.
+		 */
+		if (relay->last_row_time > relay->last_heartbeat_time)
+			row.tm = 0;
+		else
+			row.tm = ev_now(loop());
```
Gerold103, line 798: Maybe not encode it for heartbeats ever? Why is it even needed for non-data rows?

Gerold103 (approved):
Thanks for the fixes!

src/box/relay.cc
sergepetrenko: It's needed to notify the replica that it's up to date (synced).
If the master has no data to send, it sends a heartbeat with a non-zero row.tm, and upon receipt replica updates 'upstream.lag' and possibly transitions from 'sync' to 'follow'.

Besides, when there is no data,  upstream lag measures the network lag between the two nodes (if we ignore the fact that their clocks might diverge).

I'd rather keep it as is for now. Even though having 2 usages for 'upstream lag' is strange (the first usage is "fow far behind we are, measured in seconds" and the other is "network lag")

sergepetrenko:
Cherry-picked to 2.11: 3a4cbd50e53d24fd1e9f9797b6eb6af8bf912b90

PR #8416, opened 2023-03-07
sql: incomparable types in ORDER BY and GROUP BY

### LINKED ISSUE #6668: Problems with arrays
I have Ubuntu 20.04.
I have Tarantool 2.10, pulled from source today.
This version has a new feature for arrays, see issue #4762.
There are 6 problems.

1
I execute these requests:
``` lua
box.execute([[DROP TABLE n;]])
box.execute([[CREATE TABLE n (s1 INT PRIMARY KEY, s2 ARRAY, s3 ARRAY) ;]])
box.execute([[INSERT INTO n VALUES (1,[1, 'a', ['abc', 321], 1.5], NULL);]])
box.execute([[SELECT DISTINCT s2 FROM n;]])
```
Result = Segmentation fault.

A segmentation fault also occurs with SELECT COUNT(DISTINCT s2) FROM n;

2
I execute these requests:
``` lua
box.execute([[DROP TABLE n;]])
box.execute([[CREATE TABLE n (s1 INT PRIMARY KEY, s2 ARRAY, s3 ARRAY) ;]])
box.execute([[INSERT INTO n VALUES (1,[1, 'a', ['abc', 321], 1.5], NULL);]])
box.execute([[INSERT INTO n VALUES (2,NULL,LUA('return {1,2}'));]])
box.execute([[INSERT INTO n VALUES (3,[1, 'a', ['abc', 321], 1.5], NULL);]])
box.execute([[SELECT s2 FROM n GROUP BY s2;]])
```
Result = Assertion.
tarantool: /home/pgulutzan/tarantool-master/src/box/sql/mem.c:2367: mem_cmp_scalar: Assertion `0' failed.

3
I execute these requests:
``` lua
box.execute([[DROP TABLE n;]])
box.execute([[CREATE TABLE n (s1 INT PRIMARY KEY, s2 ARRAY) ;]])
box.execute([[INSERT INTO n VALUES (1,[1, 'a', ['abc', 321], 1.5]);]])
box.execute([[INSERT INTO n VALUES (3,[1, 'a', ['abc', 321], 1.5]);]])
box.execute([[SELECT s1,s2 FROM n ORDER BY s2, s1;]])
```
Result:
```
  - [3, [1, 'a', ['abc', 321], 1.5]]
  - [1, [1, 'a', ['abc', 321], 1.5]]
```
I believe this is wrong. The values in s2 are the same, therefore
the second ORDER BY column should have been used, so I think that
the correct result would be:
```
  - [1, [1, 'a', ['abc', 321], 1.5]]
  - [3, [1, 'a', ['abc', 321], 1.5]]
```

4
I execute these requests:
``` lua
box.execute([[DROP TABLE n;]])
box.execute([[CREATE TABLE n (s1 INT PRIMARY KEY, s2 ARRAY, s3 ARRAY);]])
box.execute([[INSERT INTO n VALUES (1,[1, 'a', ['abc', 321], 1.5], NULL);]])
box.execute([[ALTER TABLE n ADD COLUMN s4 INT;]])
```
Result =
Tuple field count 3 does not match space field count 4
Okay, this isn't specifically an ARRAY problem (sorry), but wouldn't it be
less confusing if the message was "ALTER is illegal because rows exist"?

5
I execute this request with Tarantool version 2.8:
``` lua
box.execute([[SELECT CAST("parts" AS VARBINARY) FROM "_vindex" limit 1;]])
```
Result = No error.
I execute this request with Tarantool version 2.10:
``` lua
box.execute([[SELECT CAST("parts" AS VARBINARY) FROM "_vindex" limit 1;]])
```
Result = Error = Type mismatch: can not convert array([[0, "string"]]) to varbinary'
I call this a "regression".

6
These perhaps could be described as limitations, so I am only asking
for a clear statement that the behaviour is deliberate.
a: There is no resemblance between this and the standard-SQL ARRAY data type.
b: Neither LENGTH() nor CARDINALITY() will work.
c: There is no plan to support associative arrays (maps).
d: Array columns cannot be used with any built-in function except CAST().
   That includes the previously-discussed functions LENGTH() and CARDINALITY().
   And CAST from any other type, except ANY, is illegal.
e: CREATE INDEX is is illegal. So is UNIQUE and PRIMARY KEY.
   And INDEXED BY is illegal with an RTREE index made via Lua.
g: Array columns cannot be used with = > >= < <= comparisons, only IS [NOT] NULL works.

Suggested fix: Cancel this feature, revert to the superior behaviour before issue #4762 was implemented.

ImeevMA:
This patch prohibits the use of ARRAY, MAP and INTERVAL in ORDER BY. In addition, GROUP BY now also checks the types of the arguments when building the VDBE.

Closes #6668

NO_DOC=bugfix

### DIFF
--- a/src/box/sql/resolve.c
+++ b/src/box/sql/resolve.c
@@ -981,6 +981,18 @@ sqlResolveOrderGroupBy(Parse * pParse,	/* Parsing context.  Leave error messages
 			resolveAlias(pEList, pItem->u.x.iOrderByCol - 1,
 				     pItem->pExpr, zType, 0);
 		}
+		/*
+		 * Currently, we cannot disallow ANY here, as values of ANY type
+		 * may be resolved later.
+		 */
+		enum field_type type = sql_expr_type(pItem->pExpr);
+		if (!field_type1_contains_type2(FIELD_TYPE_SCALAR, type) &&
+		    type != FIELD_TYPE_ANY) {
+			diag_set(ClientError, ER_SQL_TYPE_MISMATCH,
+				 field_type_strs[type], "comparable type");
+			pParse->is_aborted = true;
+			return -1;
+		}
 	}
 	return 0;
 }
--- a/changelogs/unreleased/gh-6668-ordering-incomparable-types.md
+++ b/changelogs/unreleased/gh-6668-ordering-incomparable-types.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* ARRAY, MAP, and INTERVAL values can no longer be used as ORDER BY
+  arguments (gh-6668).
--- a/test/sql-luatest/gh_6668_problems_with_map_array_test.lua
+++ b/test/sql-luatest/gh_6668_problems_with_map_array_test.lua
@@ -43,8 +43,7 @@ g.test_group_by = function()
         box.execute([[CREATE TABLE ta(i INT PRIMARY KEY, a ARRAY);]])
         box.execute([[INSERT INTO ta VALUES(1, [1]), (2, [2]);]])
         local sql = [[SELECT a FROM ta GROUP BY a;]]
-        local res = [[Type mismatch: can not convert array([2]) to ]]..
-                    [[comparable type]]
+        local res = [[Type mismatch: can not convert array to comparable type]]
         local _, err = box.execute(sql)
         t.assert_equals(err.message, res)
         box.execute([[DROP TABLE ta;]])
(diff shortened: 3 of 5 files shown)


src/box/sql/resolve.c
```diff
@@ -981,6 +981,14 @@ sqlResolveOrderGroupBy(Parse * pParse,	/* Parsing context.  Leave error messages
 			resolveAlias(pEList, pItem->u.x.iOrderByCol - 1,
 				     pItem->pExpr, zType, 0);
 		}
+		enum field_type type = sql_expr_type(pItem->pExpr);
+		if (type == FIELD_TYPE_ARRAY || type == FIELD_TYPE_MAP ||
+		    type == FIELD_TYPE_INTERVAL) {
```
Gerold103, line 986: Maybe you should better use `field_type1_contains_type2(FIELD_TYPE_SCALAR, type)`. To check if the type just isn't scalar. The result will be the same, but you won't hardcode specific types.

Gerold103 (commented):
Thanks for the patch!

src/box/sql/resolve.c
ImeevMA: Thank you for the suggestion! Fixed. However, we can't disallow ANY here for the time being, as it's possible that values of type ANY might be resolved later. Added a check for this.

Gerold103 (approved):
Thanks for the fix!

locker (approved):

locker:
Cherry-picked to 2.11.

PR #8417, opened 2023-03-07
sql: support collation for ANY

### LINKED ISSUE #8070: Collations cannot be used with ANY values
Collations can currently be used with SCALAR values, but cannot be used with ANY values.
```
tarantool> box.execute([[SELECT CAST('a' AS SCALAR) COLLATE "unicode_ci";]])
---
- metadata:
  - name: COLUMN_1
    type: scalar
  rows:
  - ['a']
...

tarantool> box.execute([[SELECT CAST('a' AS ANY) COLLATE "unicode_ci";]])
---
- null
- COLLATE clause can't be used with non-string arguments
...
```

Also, it might be worth to allow collations with bound variables and SQL_EXPR function arguments, since types of those will only be defined just before the VDBE is run.

ImeevMA:
This patch makes SQL to support collations for the ANY type.

Closes #8070

NO_DOC=ANY already supports collations in BOX.

### DIFF
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -265,7 +265,7 @@ check_collate_arg(struct Parse *parse, struct Expr *expr)
 	while (left->op == TK_COLLATE)
 		left = left->pLeft;
 	enum field_type type = sql_expr_type(left);
-	if (type != FIELD_TYPE_STRING && type != FIELD_TYPE_SCALAR) {
+	if (!field_type1_contains_type2(type, FIELD_TYPE_STRING)) {
 		diag_set(ClientError, ER_SQL_PARSER_GENERIC,
 			 "COLLATE clause can't be used with non-string "
 			 "arguments");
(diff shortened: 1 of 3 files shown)


src/box/sql/expr.c
```diff
@@ -265,7 +265,8 @@ check_collate_arg(struct Parse *parse, struct Expr *expr)
 	while (left->op == TK_COLLATE)
 		left = left->pLeft;
 	enum field_type type = sql_expr_type(left);
-	if (type != FIELD_TYPE_STRING && type != FIELD_TYPE_SCALAR) {
+	if (type != FIELD_TYPE_STRING && type != FIELD_TYPE_SCALAR &&
+	    type != FIELD_TYPE_ANY) {
```
Gerold103, line 269: Can also use `field_type1_contains_type2(type, FIELD_TYPE_STRING)`. Then you automatically include all types which can have string values in them.

Gerold103 (commented):
Thanks for the patch!

src/box/sql/expr.c
ImeevMA: Thank you for the review! Fixed.

Gerold103 (approved):
Thanks for the fix!

locker (approved):

locker:
Cherry-picked to 2.11.

PR #8419, opened 2023-03-07
space_upgrade: use old format for non-upgraded read view tuples - CE part

locker:
If a read view is created while a space upgrade is in progress, the space upgrade function should be applied to all tuples retrieved from the space read view. A tuple passed to the space upgrade function should have proper field names: tuples that had been upgraded by the time when the read view was created should have new field names while tuples that hadn't been upgraded should have old field names. Currently, all tuples use new field names. This works fine as long as fields aren't renamed (e.g. a new field is appended), but if a field is renamed (e.g. a new field is inserted in the middle), the space upgrade function will see invalid field names and, as a result, may not work as expected.

This PR is intended to fix this issue.

Needed for https://github.com/tarantool/tarantool-ee/issues/236

EE part: https://github.com/tarantool/tarantool-ee/pull/407

### DIFF
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -255,7 +255,7 @@ endif()
 if(ENABLE_SPACE_UPGRADE)
     list(APPEND box_sources ${SPACE_UPGRADE_SOURCES})
 else()
-    list(APPEND box_sources space_upgrade.c)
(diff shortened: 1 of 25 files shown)


alyapunov (approved):

locker:
Cherry-picked to 2.11.

PR #8422, opened 2023-03-09
Fix calling class method before it is initialized 

drewdzzz:
There are two cases when class method is used to calculate an argument for base
class constructor when it is not built yet. Fortunately, they does not
use class fields - let's make them static to use them before class
initialization legitimately.

Closes tarantool/security#96

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -1451,7 +1451,11 @@ RebuildIndex::~RebuildIndex()
  */
 class RebuildFuncIndex: public RebuildIndex
 {
-	struct index_def *
+	/*
+	 * This method is used before the class is built (is used to calculate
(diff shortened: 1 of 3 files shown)


CuriousGeorgiy (approved):

sergepetrenko (changes requested):
Thanks for the patch!

Nit:
Should be `Closes` instead of `Closes of` in the second commit.
Everything else looks good.

sergepetrenko (approved):
Thanks for the fix!

sergepetrenko:
Cherry-picked to 2.11: 0d0108810fc744b63479b3e79c51590ed2894196..65654729a5f85b478229b22f8c05ede6e2e17d9d

PR #8425, opened 2023-03-10
sql: properly check fieldno in sql.c

### LINKED ISSUE #8418: SQL SELECT from _space could lead to assertion
Reproducer:
```
box.execute([[SELECT "owner", "id" FROM "_space" WHERE "owner" = 1;]])
```

Result:
```
tarantool> box.execute([[SELECT "owner", "id" FROM "_space" WHERE "owner" = 1;]])
tarantool: ./src/box/tuple_format.h:305: tuple_format_field_by_path: Assertion `fieldno < tuple_format_field_count(format)' failed.
Aborted (core dumped)
```

ImeevMA:
Tuples that have been inserted into system spaces after the _space definition has been inserted into _space have the same format as the space into which they were inserted. However, previously inserted tuples have an incomplete tuple format with parts missing. One piece that is missing information is the JSON token, which is used to determine the number of fields.

The tarantoolsqlIdxKeyCompare() function contains the correct checks for the case when the number of fields from the format is equal to or less than the fieldno field, but uses tuple_format_field() before this check, resulting in an assertion.

This patch forces tarantoolsqlIdxKeyCompare() to call tuple_format_field() only after checking that fieldno is less than field_count.

Closes #8418

NO_DOC=Bugfix in debug.

### DIFF
--- a/changelogs/unreleased/gh-8418-asserton-on-select-from-_space.md
+++ b/changelogs/unreleased/gh-8418-asserton-on-select-from-_space.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* Fixed an assertion when selecting tuples with incomplete internal
+  format (gh-8418).
--- a/src/box/sql.c
+++ b/src/box/sql.c
@@ -794,9 +794,11 @@ tarantoolsqlIdxKeyCompare(struct BtCursor *cursor,
 		}
 
 		if (fieldno != next_fieldno) {
-			struct tuple_field *field =
-				tuple_format_field(format, fieldno);
-			if (fieldno >= field_count ||
+			struct tuple_field *field = NULL;
+			if (fieldno < field_count)
+				field = tuple_format_field(format, fieldno);
+
+			if (field == NULL ||
 			    field->offset_slot == TUPLE_OFFSET_SLOT_NIL) {
 				/* Outdated field_map. */
 				uint32_t j = 0;
--- a/test/sql-luatest/gh_8418_select_lead_to_assertion_test.lua
+++ b/test/sql-luatest/gh_8418_select_lead_to_assertion_test.lua
@@ -1,0 +1,20 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function()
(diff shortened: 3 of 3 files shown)


Gerold103 (commented):
Thanks for the patch! It would be nice the explain in the commit message why is it done, not what is done. How is it possible that a named field number was beyond the tuple format?

Gerold103 (approved):
Thanks for the fix!

src/box/sql.c
```diff
@@ -794,8 +794,10 @@ tarantoolsqlIdxKeyCompare(struct BtCursor *cursor,
 		}
 
 		if (fieldno != next_fieldno) {
-			struct tuple_field *field =
-				tuple_format_field(format, fieldno);
+			struct tuple_field *field = NULL;
+			if (fieldno < field_count)
+				field = tuple_format_field(format, fieldno);
+
 			if (fieldno >= field_count ||
```
locker, line 801:
```suggestion
			if (field == NULL ||
```

locker (approved):

changelogs/unreleased/gh-8418-asserton-on-select-from-_space.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/sql
+
+* Fixed assertion when selecting from '_space' using 'owner' field in
+  WHERE (gh-8418).
```
p7nov, line 4:
```suggestion
* Fixed an assertion when selecting from `_space` using the `owner` field in
  `WHERE` (gh-8418).
```

p7nov (approved):

changelogs/unreleased/gh-8418-asserton-on-select-from-_space.md
ImeevMA: Thank you for the review! I changed changelog a bit.

p7nov (approved):

src/box/sql.c
ImeevMA: Thank you for the review and suggestion! Fixed.

locker (approved):

locker:

Issue #8069, opened 2022-12-16
Backporting LuaJIT upstream commits part 3


Buristan:
This ticket is an umbrella for all commits we planning to back-port in the next Q.
This list includes commits excluded from #7230 and may be updated later with some new.

GC64:
* [x] [7e662e4f](https://github.com/LuaJIT/LuaJIT/commit/7e662e4f) x64/LJ_GC64: Fix emit_rma(). (**on review**, @Buristan)

Handle exeptions on traces:
* [x] [a32aeadc](https://github.com/LuaJIT/LuaJIT/commit/a32aeadc) Handle on-trace OOM errors from helper functions. (**on review**, @fckxorg)
* [x] [e9577376](https://github.com/LuaJIT/LuaJIT/commit/e9577376) Disable unreliable assertion for external frame unwinding. (**on review**, @fckxorg)
* [x] [be251d91](https://github.com/LuaJIT/LuaJIT/commit/be251d91) OSX: Disable unreliable assertion for external frame unwinding. (**on review**, @fckxorg)
* [x] [bf51d353](https://github.com/LuaJIT/LuaJIT/commit/bf51d353) Fix IR_RENAME snapshot number. Follow-up fix for a32aeadc. (**on review**, @fckxorg)

ARM64:
* [x] [b33e3f2d](https://github.com/LuaJIT/LuaJIT/commit/b33e3f2d) ARM64: Avoid side-effects of constant rematerialization.
* [x] [5bf0da3d](https://github.com/LuaJIT/LuaJIT/commit/5bf0da3d) ARM64: Fix {AHUV}LOAD specialized to nil/false/true.
* [x] [b4b2dce9](https://github.com/LuaJIT/LuaJIT/commit/b4b2dce9) ARM64: Fix pcall() error case.

general + x86/x64:
* [x] [03208c81](https://github.com/LuaJIT/LuaJIT/commit/03208c81) Fix math.min()/math.max() inconsistencies.
* [x] [522d2073](https://github.com/LuaJIT/LuaJIT/commit/522d2073) x86/x64: Fix loop realignment.
* [x] [fc63c938](https://github.com/LuaJIT/LuaJIT/commit/fc63c938) Fix os.date() for wider libc strftime() compatibility.
* [x] [430d9f8f](https://github.com/LuaJIT/LuaJIT/commit/430d9f8f) Fix string.format("%c", 0). 
[...]

Issue #7230, opened 2022-06-02
Backporting LuaJIT upstream commits part 2


Buristan:
This ticket is an umbrella for all commits we planning to back-port in the next Q.
This list includes commits excluded from #6548 and may be updated later with some new.

Unwinding on M1 and prerequisites:
* [x] [e1319361](https://github.com/LuaJIT/LuaJIT/commit/e1319361) Cleanup and enable external unwinding for more platforms.
* [x] [d4a554d6](https://github.com/LuaJIT/LuaJIT/commit/d4a554d6) OSX: Fix build by hardcoding external frame unwinding.
* [x] [27ee3bcd](https://github.com/LuaJIT/LuaJIT/commit/27ee3bcd) OSX/ARM64: Disable external unwinding for now.
* [x] [421c4c79](https://github.com/LuaJIT/LuaJIT/commit/421c4c79) ARM64: Reorder interpreter stack frame and fix unwinding.
* [x] [78350a25](https://github.com/LuaJIT/LuaJIT/commit/78350a25) OSX/ARM64: Disable unwind info.
* [x] [b9d52396](https://github.com/LuaJIT/LuaJIT/commit/b9d52396) BSD: Fix build with BSD grep.
* [x] [66563bda](https://github.com/LuaJIT/LuaJIT/commit/66563bda) Fix build with busybox grep.
* [x] [bfcb74ed](https://github.com/LuaJIT/LuaJIT/commit/bfcb74ed) OSX/ARM64: Fix external unwinding. (#6096)

general + x86/x64:
* [x] [b4ed3219](https://github.com/LuaJIT/LuaJIT/commit/b4ed3219) LJ_GC64: Fix ir_khash for non-string GCobj.
* [x] [05fbdf56](https://github.com/LuaJIT/LuaJIT/commit/05fbdf56) x64/LJ_GC64: Fix type-check-only variant of SLOAD.
* [x] [b5b20191](https://github.com/LuaJIT/LuaJIT/commit/b5b20191) Save trace recorder state around VM event call. (#6782)
* [x] [ca8d3257](https://github.com/LuaJIT/LuaJIT/commit/ca8d3257) Ensure correct stack top for OOM error message.
* [x] [0913b16e](https://github.com/LuaJIT/LuaJIT/commit/0913b16e) Fix handling of errors during snapshot restore.
* [x] [de8e88b0](https://github.com/LuaJIT/LuaJIT/commit/de8e88b0) Call error function on rethrow after trace exit.
[...]

PR #8426, opened 2023-03-10
luajit: bump new version

### LINKED ISSUE #6163: math.min/math.max JIT misbehaviour
Originally observed via [CI](https://github.com/tarantool/tarantool/runs/2859823675#step:5:9154).

When one runs the script it produces different results if `math.min()`/`math.max()` is compiled:
```lua
local pcall = pcall
local print = print
local tostring = tostring

local curr_test = 0
function ok (test, name)
    curr_test = curr_test + 1
    local out = ''
    if not test then
        out = "not "
    end
    out = out .. "ok " .. tostring(curr_test)
    if name then
        out = out .. " - " .. name
    end
    print(out)
end

function like (got, pattern, name)
    local pass = tostring(got):match(pattern)
    ok(pass, name)
    if not pass then
        diag("                  " .. tostring(got))
        diag("    doesn't match '" .. tostring(pattern) .. "'")
    end
end

function error_like (code, pattern, name)
    local r, msg = pcall(code)
    if r then
        ok(false, name)
        diag("    unexpected success")
        diag("    expected: " .. tostring(pattern))
    else
        like(msg, pattern, name)
    end
end

function diag (msg)
    print("# " .. msg)
end

for i= 1,1e3 do
    error_like(function () math.min() end,
               "^[^:]+:%d+: bad argument #1 to 'min' %(.- expected",
               "function min 0")
end
```

The output:
```
ok 1 - function min 0
ok 2 - function min 0
ok 3 - function min 0
...
not ok 769 - function min 0
#     unexpected success
#     expected: ^[^:]+:%d+: bad argument #1 to 'min' %(.- expected
ok 770 - function min 0
...
ok 1000 - function min 0
```

The problem occurs when recording trace is aborted.
The issue is fixed in vanilla LuaJIT.
Necessary to simplify test case and backport corresponding patches.

igormunkin:
* ARM64: Avoid side-effects of constant rematerialization.
* ARM64: Fix {AHUV}LOAD specialized to nil/false/true.
* ARM64: Fix pcall() error case.
* Fix math.min()/math.max() inconsistencies.
* test: add test case for math.modf

Closes #6163
Part of #8069
Follows up #7230

NO_DOC=LuaJIT submodule bump
NO_TEST=LuaJIT submodule bump

### DIFF
--- a/changelogs/unreleased/gh-8069-luajit-fixes.md
+++ b/changelogs/unreleased/gh-8069-luajit-fixes.md
@@ -1,0 +1,10 @@
+## bugfix/luajit
+
+Backported patches from vanilla LuaJIT trunk (gh-8069). In the scope of this
+activity, the following issues have been resolved:
+
+* Fixed successful `math.min/math.max` call with no args (gh-6163).
+* Fixed inconsistencies in `math.min/math.max` calls with NaN arg (gh-6163).
+* Fixed `pcall()` call without arguments on arm64.
+* Fixed assembling of IR_{AHUV}LOAD specialized to boolean for aarch64.
+* Fixed constant rematerialization on arm64.


Buristan (approved):
Hi, Igor!
Thanks for the patch!
LGTM!

ylobankov (approved):
LGTM

Issue #6293, opened 2021-08-05
Implement statistic for iproto streams and remove in from `errinj`


EvgenyMekhanik:
1. At this moment, statistics on the number of streams and messages in streams are collected only for test purposes and stored in `errinj` to check that there is no memory leak in streams. Discuss what statistics are needed for streams, add it and remove statistics for streams from errinj.
2. Also at this moment we use errinj to collect iproto statistic per thread for test purpose. We store in errinj thread
number, which determine for what thread we want to get statistic. May be in this issue discuss and resolve it in other way.

unera:
Let's add the key `STREAMS` into `box.stat.net`:

```yaml
STREAMS:
   threads:
   - active: 12  # count of active streams
     requests:
        total: 123
        rps: 2    # last 5 second
   - active: 0
     requests:
         total: 0
         rps: 0
```

PR #8430, opened 2023-03-13
test: fix flaky box/gh-6293-implement-new-net-stat

locker:
The test checks that the number of IPROTO requests handled by a test server is reported correctly in statistics. Since a net.box connection sends a few "service" requests (e.g. to fetch schema), the test excludes them from the total count. The problem is this doesn't always work with service requests sent to enable graceful shutdown.

To enable graceful shutdown a client sends an IPROTO_WATCH request. The server replies to the client with IPROTO_EVENT. Upon receiving the event, the client sends another IPROTO_WATCH request to ack it. The whole procedure is fully asynchronous, which means it may finish after we start processing user requests over the connection.

To correctly account service requests, let's disable this feature.

Closes tarantool/tarantool-qa#269

### DIFF
--- a/test/box/gh-6293-implement-new-net-stat.result
+++ b/test/box/gh-6293-implement-new-net-stat.result
@@ -142,7 +142,14 @@ test_run:switch("default")
 server_addr = test_run:cmd("eval test 'return box.cfg.listen'")[1]
 ---
 ...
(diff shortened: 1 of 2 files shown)


nshy (approved):

locker:

PR #8435, opened 2023-03-13
box: fix 64-bit schema version for message pack.

darthunix:
0b876b7600 introduced `uint64_t` schema version to deal with the possible 32-bit counter overflow problem. But for some reason message pack still serialized 64-bit schema version as 32-bit one. Current commit fixes the issue.


locker (approved):

locker:
Cherry-picked to 2.11.

Issue #6385, opened 2021-08-24
mvcc: crash with multikey indexes


olegrok:
Tarantool 2.8.2

```
tarantool> box.space.storage_InfoCard
---
- engine: memtx
  before_replace: 'function: 0x010afd4000'
  field_count: 0
  is_sync: false
  ck_constraint: []
  on_replace: 'function: 0x010afd3ec0'
  temporary: false
  index:
    0: &0
      unique: true
      parts:
      - type: string
        is_nullable: false
        fieldno: 1
      hint: true
      id: 0
      type: TREE
      space_id: 542
      name: name
    1: &1
      unique: false
      parts:
      - type: string
        is_nullable: true
        path: '[*]'
        fieldno: 5
      - type: string
        is_nullable: false
        fieldno: 1
      hint: true
      id: 1
      type: TREE
      space_id: 542
      name: metainfo
    2: &2
      unique: false
      parts:
      - type: unsigned
        is_nullable: false
        fieldno: 3
      hint: true
      id: 2
      type: TREE
      space_id: 542
      name: bucket_id
    bucket_id: *2
    metainfo: *1
    name: *0
  is_local: false
  enabled: true
  name: storage_InfoCard
  id: 542
...

tarantool> box.space.storage_InfoCard:replace({'abc', 123, 123, {true, true}, {'abc', 'abc'}})
Segmentation fault
  code: SEGV_MAPERR
  addr: 0x20f003f0e
  context: 0x108dff9d8
  siginfo: 0x108dff970
Current time: 1629827331
Please file a bug at http://github.com/tarantool/tarantool/issues
Attempting backtrace... Note: since the server has already crashed, 
this may fail as well
#0  0x10629ddb2 in crash_signal_cb+92
#1  0x7fff2055ed7d in _sigtramp+1d
#2  0x1069cccb1 in tuple_hash_field+21
#3  0x1069cbe1a in _Z19tuple_hash_slowpathILb1ELb1EEjP5tupleP7key_def+9a
#4  0x10615e43a in check_hole+4a
#5  0x10615974e in memtx_tx_history_add_insert_stmt+2fe
#6  0x106164fb7 in memtx_space_execute_replace+a7
#7  0x1061b9376 in space_execute_dml+86
#8  0x1061e17d0 in box_process_rw+a0
#9  0x1061e7868 in box_replace+48
#10 0x106270182 in lbox_replace+62
#11 0x1062cf823 in lj_BC_FUNCC+44
[...]

PR #8438, opened 2023-03-14
memtx: refactor transaction management of space replace

CuriousGeorgiy:
Needed for #6385

See also #8465 PR implementing #6385.

### DIFF
--- a/src/box/memtx_space.c
+++ b/src/box/memtx_space.c
@@ -91,6 +91,38 @@ memtx_space_update_bsize(struct space *space, struct tuple *old_tuple,
 	memtx_space->bsize += new_bsize - old_bsize;
 }
 
+int
+memtx_index_replace(struct txn *txn, struct index *index,
+		    struct tuple *old_tuple, struct tuple *new_tuple,
+		    enum dup_replace_mode mode, struct tuple **result,
+		    struct tuple **successor)
+{
+	if (txn == NULL) {
+		if (index_replace(index, old_tuple, new_tuple, mode, result,
+				  successor) != 0)
+			return -1;
+		return 0;
+	}
+	memtx_tx_story_gc();
+	if (new_tuple == NULL)
+		return memtx_tx_history_add_delete(index, old_tuple, result);
+	if (index_replace(index, NULL, new_tuple, DUP_REPLACE_OR_INSERT, result,
+			  successor) != 0)
+		return -1;
+	if (memtx_tx_history_add_insert(index, old_tuple, new_tuple, mode,
+					result, *successor) != 0) {
+		struct tuple *unused;
+		if (index_replace(index, new_tuple, *result, DUP_INSERT,
+				  &unused, &unused) != 0) {
+			diag_log();
+			unreachable();
+			panic("failed to rollback change");
+		}
+		return -1;
+	}
+	return 0;
+}
+
 /**
  * A version of space_replace for a space which has
  * no indexes (is not yet fully built).
@@ -151,8 +183,8 @@ memtx_space_replace_primary_key(struct space *space, struct tuple *old_tuple,
 				struct tuple **result)
 {
 	struct tuple *successor;
-	if (index_replace(space->index[0], old_tuple,
-			  new_tuple, mode, &old_tuple, &successor) != 0)
+	if (memtx_index_replace(NULL, space->index[0], old_tuple, new_tuple,
+				mode, &old_tuple, &successor) != 0)
 		return -1;
 	memtx_space_update_bsize(space, old_tuple, new_tuple);
 	if (new_tuple != NULL)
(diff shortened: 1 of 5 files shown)


CuriousGeorgiy:
@locker could you please review this patch from the `memtx_tx_history_add_{insert, delete}` layer of abstraction and above?

src/box/memtx_space.c
```diff
@@ -301,13 +297,13 @@ memtx_space_replace_all_keys(struct space *space, struct tuple *old_tuple,
 	for (i++; i < space->index_count; i++) {
 		struct tuple *unused;
 		struct index *index = space->index[i];
-		if (index_replace(index, old_tuple, new_tuple,
+		if (index_replace(index, needs_txm, old_tuple, new_tuple,
```
locker, line 300: An unnamed boolean argument looks terrible from the code readability point of view. It looks like this is the only place where you can possibly pass `true`. I don't understand why you can't call `index_replace_with_txm` here directly.

src/box/index.cc
```diff
@@ -806,6 +806,20 @@ index_delete(struct index *index)
...
+index_replace_with_txm(struct index *index, struct tuple *old_tuple,
+		       struct tuple *new_tuple, enum dup_replace_mode mode,
+		       struct tuple **result, struct tuple **successor)
+{
+	memtx_tx_story_gc();
+	if (new_tuple == NULL)
+		return memtx_tx_history_add_delete(index, old_tuple, result);
+	return index->vtab->replace(index, NULL, new_tuple,
+				    DUP_REPLACE_OR_INSERT, result, successor) ||
+		memtx_tx_history_add_insert(index, old_tuple, new_tuple, mode,
```
locker, line 819: Nit: Hard to read. Better write explicitly:

```C
if (index->vtab->replace(...) != 0)
        return memtx_tx_history_add_insert(...);
return 0;
```

CuriousGeorgiy: > Hard to read.

I believe it's rather unusual than hard. I know I could go for the conventional option you suggest, but I like the style I suggest, because it looks like a pipeline notation (i.e., perform a sequence of actions propagating the result further and aborting the pipeline if any of the actions happen to fail).

CuriousGeorgiy: Nevermind, I decided to pull out the index state rollback in case of transaction manager notification failure, so I now use @locker's approach.

src/box/memtx_space.c
CuriousGeorgiy: Good point, didn’t think of it: applied your suggestion.

src/box/index.cc
```diff
@@ -806,6 +806,31 @@ index_delete(struct index *index)
...
 
+int
+index_replace_with_txm(struct index *index, struct tuple *old_tuple,
+		       struct tuple *new_tuple, enum dup_replace_mode mode,
+		       struct tuple **result, struct tuple **successor)
+{
+	memtx_tx_story_gc();
+	if (new_tuple == NULL)
+		return memtx_tx_history_add_delete(index, old_tuple, result);
+	if (index->vtab->replace(index, NULL, new_tuple, DUP_REPLACE_OR_INSERT,
```
locker, line 817: Nit: Should use `index_replace()`.

src/box/index.cc
```diff
@@ -806,6 +806,31 @@ index_delete(struct index *index)
...
+				 result, successor) != 0)
+		return -1;
+	if (memtx_tx_history_add_insert(index, old_tuple, new_tuple, mode,
```
locker, line 820: Looks like this function is memtx-specific. I'd make it local to `memtx_space.cc`.

src/box/memtx_space.c
```diff
@@ -281,33 +281,33 @@ memtx_space_replace_all_keys(struct space *space, struct tuple *old_tuple,
...
+	bool needs_txm = memtx_tx_manager_use_mvcc_engine &&
+			 !space->def->opts.is_ephemeral;
 	/*
 	 * If old_tuple is not NULL, the index has to
 	 * find and delete it, or return an error.
 	 */
+	int (*index_replace_impl)(struct index *, struct tuple *,
+				  struct tuple *, enum dup_replace_mode,
+				  struct tuple **, struct tuple **) =
+		needs_txm ? index_replace_with_txm : index_replace;
```
locker, line 293: Dealing with a function pointer looks ugly. I'd make `index_replace_with_txm` decide whether it actually needs to use txm or not:

```C
static int
memtx_index_replace(...)
{
        if (!memtx_tx_manager_use_mvcc_engine)
                return index_replace(...);
        ...
}
```

src/box/index.cc
CuriousGeorgiy: Actually, AFAIC, the whole index manipulation is memtx-specific — vinyl goes through it's own path:
https://github.com/tarantool/tarantool/blob/921a07177521cfdea7c59ec84a78b19be7ef44ff/src/box/vinyl.c#L4653-L4668

> I'd make it local to `memtx_space.cc`.

I don't understand what you mean, could you please elaborate?

locker: `src/box/index.cc` shouldn't depend on memtx. I'd move this function to `src/box/memtx_space.cc`, because it's the only file where it's used.

CuriousGeorgiy: Refactored.

src/box/memtx_space.c
CuriousGeorgiy: Refactored.

Issue #8110, opened 2023-01-08
clang-15: build fails with 'Unable to get thread stack'


Totktonada:
The build works with clang 14:

```sh
CC=clang-14 CXX=clang++-14 cmake . \
    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
    -DENABLE_BACKTRACE=ON \
    -DENABLE_DIST=ON \
    -DENABLE_FEEDBACK_DAEMON=OFF \
    -DENABLE_BUNDLED_LIBCURL=OFF && \
    make -j
```

With clang 14 it shows the following cmake check results around pthread:

```
-- Looking for pthread_np.h
-- Looking for pthread_np.h - not found
-- Performing Test HAVE_PTHREAD_SETNAME_NP
-- Performing Test HAVE_PTHREAD_SETNAME_NP - Success
-- Performing Test HAVE_PTHREAD_SETNAME_NP_1
-- Performing Test HAVE_PTHREAD_SETNAME_NP_1 - Failed
-- Performing Test HAVE_PTHREAD_SET_NAME_NP
-- Performing Test HAVE_PTHREAD_SET_NAME_NP - Failed
-- Performing Test HAVE_PTHREAD_GETATTR_NP
-- Performing Test HAVE_PTHREAD_GETATTR_NP - Success
-- Performing Test HAVE_PTHREAD_STACKSEG_NP
-- Performing Test HAVE_PTHREAD_STACKSEG_NP - Failed
-- Performing Test HAVE_PTHREAD_ATTR_GET_NP
-- Performing Test HAVE_PTHREAD_ATTR_GET_NP - Failed
-- Performing Test HAVE_PTHREAD_GET_STACKSIZE_NP
-- Performing Test HAVE_PTHREAD_GET_STACKSIZE_NP - Failed
-- Performing Test HAVE_PTHREAD_GET_STACKADDR_NP
-- Performing Test HAVE_PTHREAD_GET_STACKADDR_NP - Failed
```

However the same on clang 15:

```sh
CC=clang-15 CXX=clang++-15 cmake . \
    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
    -DENABLE_BACKTRACE=ON \
    -DENABLE_DIST=ON \
    -DENABLE_FEEDBACK_DAEMON=OFF \
    -DENABLE_BUNDLED_LIBCURL=OFF && \
    make -j
```

Gives 'all failed' for the pthread related checks.

```
[...]

PR #8439, opened 2023-03-14
build: drop unused variable detected by clang 15

Totktonada:
```c
<...>/src/box/sql/vdbe.c:378:11: error: variable 'nVmStep' set but not
    used [-Werror,-Wunused-but-set-variable]
        unsigned nVmStep = 0;      /* Number of virtual machine steps */
                 ^
```

The usage of the variable was removed in commit dbad19ef0aae ("sql: drop unused functions").

See also #8110.


ImeevMA (approved):
Thank you for the patch!

PR #8447, opened 2023-03-16
net.box: fix crash when remote space field has unknown type

### LINKED ISSUE #4632: netbox: error when client and server have different sets of supported field types
When a client connects to the server using netbox, an error may
occur when the client and server have different sets of field
types.

Reproducer (Tarantool version in brackets):
1) Server (1.10):
```
box.cfg{listen=3302}
box.schema.user.grant('guest','read, write, execute', 'universe')
format = {}
format[1] = {type = 'unsigned', name = 'i'}
format[2] = {type = 'any', name = 'a'}
box.schema.space.create('s', {format=format})
box.space.s:create_index('i')
```
2) Client (2.3):
```
remote = require('net.box')
cn = remote.connect(3302)
decimal = require('decimal')
cn.space.s:insert({1, decimal.new('1.123')})
```
3) Server (1.10):

```
box.space.s:select()
```

Result:
```
tarantool> box.space.s:select()
---
- - 'builtin/msgpackffi.lua:546: assertion failed!'
...
```

Another reproducer:
1) Server (2.3)
```
box.cfg{listen=3302}
box.schema.user.grant('guest','read, write, execute', 'universe')
format = {}
format[1] = {type = 'unsigned', name = 'i'}
format[2] = {type = 'decimal', name = 'a'}
box.schema.space.create('s', {format=format})
box.space.s:create_index('i')
```

2) Client (1.10):
```
remote = require('net.box')
cn = remote.connect(3302)
cn.space.s:select()
```

Result:
```
tarantool> cn.space.s:select()
---
- - 'builtin/msgpackffi.lua:546: assertion failed!'
...
```

The problem is more serious if the client is 2.2 and the server has more field types than the client - since in 2.2 we create tuple_format for each server space (#2978). In this regard, if the client detects any new field type in server spaces , we will get assert in dev build or segfault in release build.

locker:
This commit fixes the following assertion failure that happens on a client in case a remote schema contains an unknown field type:

```
src/box/lua/misc.cc:395: int lbox_tuple_format_new(lua_State*): Assertion `fields[i].type != field_type_MAX' failed.
```

To fix the bug we remove the code that tries to set field types from `box.internal.new_tuple_format`. Actually, the format is used solely for providing field names so types are ignored anyway.

Closes #4632

### DIFF
--- a/changelogs/unreleased/gh-4632-net-box-unknown-field-type.md
+++ b/changelogs/unreleased/gh-4632-net-box-unknown-field-type.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* Fixed a net.box client crash when a remote space has a field type that isn't
+  supported by the client version (gh-4632).
--- a/src/box/lua/misc.cc
+++ b/src/box/lua/misc.cc
@@ -381,21 +381,9 @@ lbox_tuple_format_new(struct lua_State *L)
 	}
 	for (uint32_t i = 0; i < count; ++i) {
 		size_t len;
-
 		fields[i] = field_def_default;
-
 		lua_pushinteger(L, i + 1);
 		lua_gettable(L, 1);
-
-		lua_pushstring(L, "type");
-		lua_gettable(L, -2);
-		if (! lua_isnil(L, -1)) {
-			const char *type_name = lua_tolstring(L, -1, &len);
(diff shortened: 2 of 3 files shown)


alyapunov (approved):

locker:
Cherry-picked to 2.11.

PR #8450, opened 2023-03-16
Cleanup issues found with static analysis

sergepetrenko:
commit d526306e463e4cbd05a666a4b86f477c40c2f53a
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Mar 16 17:14:48 2023 +0300

refactoring: fix casting double to uint64_t in thread_sleep()

In order to avoid casting a too big double to uint64_t, let's return
without sleeping when given a negative or too large `sec` value.

While we are at it, let's reduce TIMEOUT_INFINITY to 50 years (365 *
86400 * 50), because this value is smaller than INT32_MAX (and 100 years
is greater). This way we may safely cast seconds to `time_t` a couple of
lines below (let's assume time_t can hold at least INT32_MAX).

Closes tarantool/security#114

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring

commit 7e425acf75349967758c281f6bf33b83768db281
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Mar 16 16:49:58 2023 +0300

refactoring: use size_t for strlen() result in fdprintf()

`strlen()` returns size_t, hence the variable holding the result should
be of type size_t as well.

Now  makes sense to compare `size_t len` with `ssize_t res` only for
positive `res`, so fix this as well.

Closes tarantool/security#113

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring

commit baba43d27b0d18fb2d2f8ffbc28d33bde053155e
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Mar 16 15:34:29 2023 +0300

refactoring: get rid of unsafe cast of unsigned to signed

Casting a uint64_t greater than INT64_MAX to int64_t is
implementation-defined behaviour, according to the C standard. Let's
avoid that.

In both cases fixed `len` is uint32_t and `ibuf_used(in)` returns a
size_t (aka uint64_t on every platform that we care about).
Hence the result of the subtraction is uint64_t and better use
it directly. Besides, `coio_breadn_timeout` also takes a size_t.

While I'm at it, let's actually change `len` to be uint64_t:
`mp_decode_uint()` returns that anyway.

Closes tarantool/security#108
Closes tarantool/security#109

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring


sergepetrenko:
I have a dilemma about the last commit (about thread_sleep())

Actually, we should also check that `req.tv_sec = ns / 1000000000;` does not overflow `time_t` which is the type of `tv_sec`.
Unfortunately, there are no specifications, on what a `time_t` is. So even if we add a check that `ns / 1000000000` is less than `INT_MAX`, or maybe `INT32_MAX`, this doesn't prove anything.

I think the assumption that time_t is at least as big as int32_t holds on all the platforms we care about.
So my initial idea was to make sure that TIMEOUT_INFINITY is less than INT32_MAX.  Then the one check about `sec < TIMEOUT_INFINITY` would suffice. I had to lower TIMEOUT_INFINITY to 50 years then (it's 100 years now). But for some unclear reason this broke `box-luatest/gh_6150_memtx_tx_memory_monitoring_test.lua` so I reverted that change.

Long story short, I don't know if it's worth checking that `ns / 1000000000 < INT32_MAX` separately or not. That extra check looks rather ugly.

src/lib/core/util.c
```diff
@@ -191,7 +192,7 @@ fdprintf(int fd, const char *format, ...)
 		res = write(fd, start, len);
 		if (res > 0)
 			total += res;
-		if (res != len)
+		if (res < 0 || (size_t)res != len)
```
locker, line 195: This function isn't used. You're fixing dead code here...

`total` is `ssize_t` so this function is screwed if `strlen` turns out to be that big.

I don't understand how this could possibly lead to a (security) bug. If arguments are insanely big, we'll write garbage to the file - so what?

Please drop this patch and mark the SVACE warning as "Won't fix" or "False positive".

src/lib/core/util.c
```diff
@@ -521,8 +521,10 @@ double_compare_nint64(double lhs, int64_t rhs, int k)
 void
 thread_sleep(double sec)
 {
+	assert(sec >= 0 && sec <= TIMEOUT_INFINITY);
+	if (sec < 0 || sec > TIMEOUT_INFINITY)
+		return;
```
locker, line 526: This is an internal function and we always pass sane values to it. But even if we passed a wrong value, I don't see why that would be a security issue.

AFAIU you're effectively swapping one assertion with an equivalent one. I don't see any point in it.

Please drop this patch and mark the SVACE issue as "False positive".

locker (changes requested):
Please start subject line with a subsystem name, e.g. ("xrow: fix ...").

src/lib/core/util.c
sergepetrenko: > This function isn't used. You're fixing dead code here...

Yep, missed that, thanks!
Let's remove it altogether then.

src/lib/core/util.c
sergepetrenko: Ok, makes sense to close it.

locker (approved):

locker:

Issue #8098, opened 2022-12-27
Introduce a way to get description of existing space


ImeevMA:
At present, it is not possible to obtain a complete description of an existing space. Some of the information can be retrieved by means such as selecting from `_space` and `_index` or by using various pragmas. However, some pragmas no longer work properly, such as `foreign_key_list`. Thus, we need a way to get complete information about the existing space. One suggested way is to use the new `SHOW CREATE TABLE` operator.

PR #8451, opened 2023-03-16
sql: introduce SHOW CREATE TABLE statement

ImeevMA:
This patch introduces the SHOW CREATE TABLE statement. This statement can be used to obtain a description of a space in the form of a corresponding CREATE TABLE statement.

Part of #8098

@TarantoolBot document
Title: SHOW CREATE TABLE statement

Statement can be used to obtain a description of a space in the form of a corresponding CREATE TABLE statement.

There is two syntax for SHOW CREATE TABLE statement: 1) Get description of single space:
```
SHOW CREATE TABLE table_name;
```

2) Get descriptions of all available non-system spaces:
```
SHOW CREATE TABLE;
```

### DIFF
--- a/changelogs/unreleased/gh-8098-show-create-table.md
+++ b/changelogs/unreleased/gh-8098-show-create-table.md
@@ -1,0 +1,3 @@
+## feature/sql
+
+* Introduced the `SHOW CREATE TABLE` statement (gh-8098).
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -3395,3 +3395,69 @@ sql_setting_set(struct Parse *parse_context, struct Token *name,
 	sqlVdbeAddOp4(vdbe, OP_SetSession, target, 0, 0, key, P4_DYNAMIC);
 	return;
 }
+
+/** Emit VDBE instructions for "SHOW CREATE TABLE table_name;" statement. */
+void
+sql_emit_show_create_table_one(struct Parse *parse, struct Token *name)
+{
+	struct Vdbe *v = sqlGetVdbe(parse);
+	char *space_name = sql_name_from_token(name);
+	sqlVdbeSetNumCols(v, 2);
+	vdbe_metadata_set_col_name(v, 0, "STATEMENTS");
+	vdbe_metadata_set_col_type(v, 0, "array");
+	vdbe_metadata_set_col_name(v, 1, "ERRORS");
+	vdbe_metadata_set_col_type(v, 1, "array");
+
+	int cursor = parse->nTab++;
+	int space_reg = ++parse->nMem;
+	int name_reg = ++parse->nMem;
+	sqlVdbeAddOp2(v, OP_OpenSpace, space_reg, BOX_VSPACE_ID);
+	sqlVdbeAddOp3(v, OP_IteratorOpen, cursor, 2, space_reg);
+	sqlVdbeChangeP5(v, OPFLAG_SEEKEQ);
+	sqlVdbeAddOp4(v, OP_String8, 0, name_reg, 0, space_name, P4_DYNAMIC);
+	int addr1 = sqlVdbeAddOp4Int(v, OP_SeekGE, cursor, 0, name_reg, 1);
+	int addr2 = sqlVdbeAddOp4Int(v, OP_IdxGT, cursor, 0, name_reg, 1);
+	int space_id_reg = ++parse->nMem;
+	sqlVdbeAddOp3(v, OP_Column, cursor, BOX_SPACE_FIELD_ID, space_id_reg);
+	int result_reg = ++parse->nMem;
+	++parse->nMem;
+	sqlVdbeAddOp2(v, OP_ShowCreateTable, space_id_reg, result_reg);
+	sqlVdbeAddOp2(v, OP_ResultRow, result_reg, 2);
+	int addr3 = sqlVdbeAddOp0(v, OP_Goto);
+	sqlVdbeJumpHere(v, addr1);
+	sqlVdbeJumpHere(v, addr2);
+
+	char *err = sqlMPrintf(tnt_errcode_desc(ER_NO_SUCH_SPACE), space_name);
+	sqlVdbeAddOp4(v, OP_SetDiag, ER_NO_SUCH_SPACE, 0, 0, err, P4_DYNAMIC);
+	sqlVdbeAddOp2(v, OP_Halt, -1, ON_CONFLICT_ACTION_ABORT);
+	sqlVdbeJumpHere(v, addr3);
+}
+
+/** Emit VDBE instructions for "SHOW CREATE TABLE;" statement. */
+void
+sql_emit_show_create_table_all(struct Parse *parse)
+{
+	struct Vdbe *v = sqlGetVdbe(parse);
+	sqlVdbeSetNumCols(v, 2);
+	vdbe_metadata_set_col_name(v, 0, "STATEMENTS");
+	vdbe_metadata_set_col_type(v, 0, "array");
+	vdbe_metadata_set_col_name(v, 1, "ERRORS");
+	vdbe_metadata_set_col_type(v, 1, "array");
+
+	int cursor = parse->nTab++;
+	int space_reg = ++parse->nMem;
+	int key_reg = ++parse->nMem;
+	sqlVdbeAddOp2(v, OP_OpenSpace, space_reg, BOX_VSPACE_ID);
+	sqlVdbeAddOp3(v, OP_IteratorOpen, cursor, 0, space_reg);
+	sqlVdbeAddOp2(v, OP_Integer, BOX_SYSTEM_ID_MAX, key_reg);
+	int addr1 = sqlVdbeAddOp4Int(v, OP_SeekGT, cursor, 0, key_reg, 1);
+	int space_id_reg = ++parse->nMem;
+	int addr2 = sqlVdbeAddOp3(v, OP_Column, cursor, BOX_SPACE_FIELD_ID,
+				  space_id_reg);
+	int result_reg = ++parse->nMem;
+	++parse->nMem;
+	sqlVdbeAddOp2(v, OP_ShowCreateTable, space_id_reg, result_reg);
+	sqlVdbeAddOp2(v, OP_ResultRow, result_reg, 2);
+	sqlVdbeAddOp2(v, OP_Next, cursor, addr2);
+	sqlVdbeJumpHere(v, addr1);
+}
--- a/src/box/sql/show.c
+++ b/src/box/sql/show.c
@@ -1,0 +1,496 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include <ctype.h>
+
+#include "sqlInt.h"
+#include "mem.h"
+#include "box/schema.h"
+#include "box/sequence.h"
+#include "box/coll_id_cache.h"
+#include "box/tuple_constraint_def.h"
+
+/** An objected used to accumulate a statement. */
+struct sql_desc {
+	/** Accumulate the string representation of the statement. */
+	struct StrAccum acc;
+	/** MEM where the array of compiled statements should be inserted. */
+	struct Mem *ret;
+	/** MEM where the array of compiled errors should be inserted. */
+	struct Mem *err;
+	/** Array of compiled but not encoded statements. */
+	char **statements;
+	/** Array of compiled but not encoded errors. */
+	char **errors;
+	/** Number of compiled statements. */
+	uint32_t statement_count;
+	/** Number of compiled errors. */
+	uint32_t error_count;
+};
+
+/** Initialize the object used to accumulate a statement. */
+static void
+sql_desc_initialize(struct sql_desc *desc, struct Mem *ret, struct Mem *err)
+{
+	sqlStrAccumInit(&desc->acc, NULL, 0, SQL_MAX_LENGTH);
+	desc->statements = NULL;
+	desc->statement_count = 0;
+	desc->errors = NULL;
+	desc->error_count = 0;
+	desc->ret = ret;
+	desc->err = err;
+}
+
+/** Append a new string to the object used to accumulate a statement. */
+CFORMAT(printf, 2, 3) static void
+sql_desc_append(struct sql_desc *desc, const char *fmt, ...)
+{
+	va_list ap;
+	va_start(ap, fmt);
+	sqlVXPrintf(&desc->acc, fmt, ap);
+	va_end(ap);
+}
+
+/** Append a name to the object used to accumulate a statement. */
+static void
+sql_desc_append_name(struct sql_desc *desc, const char *name)
+{
+	char *escaped = sql_escaped_name_new(name);
+	assert(escaped[0] == '"' && escaped[strlen(escaped) - 1] == '"');
+	char *normalized = sql_normalized_name_new(name, strlen(name));
+	if (isalpha(name[0]) && strlen(escaped) == strlen(name) + 2 &&
+	    strcmp(normalized, name) == 0)
+		sqlXPrintf(&desc->acc, "%s", normalized);
+	else
+		sqlXPrintf(&desc->acc, "%s", escaped);
+	sql_xfree(normalized);
+	sql_xfree(escaped);
+}
+
+/** Append a new error to the object used to accumulate a statement. */
+static void
+sql_desc_error(struct sql_desc *desc, const char *type, const char *name,
+	       const char *error)
+{
+	char *str = sqlMPrintf("Problem with %s '%s': %s.", type, name, error);
+	uint32_t id = desc->error_count;
+	++desc->error_count;
+	uint32_t size = desc->error_count * sizeof(desc->errors);
+	desc->errors = sql_xrealloc(desc->errors, size);
+	desc->errors[id] = str;
+}
+
+/** Complete the statement and add it to the array of compiled statements. */
+static void
+sql_desc_finish_statement(struct sql_desc *desc)
+{
+	char *str = sqlStrAccumFinish(&desc->acc);
+	sqlStrAccumInit(&desc->acc, NULL, 0, SQL_MAX_LENGTH);
+	uint32_t id = desc->statement_count;
+	++desc->statement_count;
+	uint32_t size = desc->statement_count * sizeof(desc->statements);
+	desc->statements = sql_xrealloc(desc->statements, size);
+	desc->statements[id] = str;
+}
+
+/** Finalize a described statement. */
+static void
+sql_desc_finalize(struct sql_desc *desc)
+{
+	sqlStrAccumReset(&desc->acc);
+	if (desc->error_count > 0) {
+		uint32_t size = mp_sizeof_array(desc->error_count);
+		for (uint32_t i = 0; i < desc->error_count; ++i)
+			size += mp_sizeof_str(strlen(desc->errors[i]));
+		char *buf = sql_xmalloc(size);
+		char *end = mp_encode_array(buf, desc->error_count);
+		for (uint32_t i = 0; i < desc->error_count; ++i) {
+			end = mp_encode_str0(end, desc->errors[i]);
+			sql_xfree(desc->errors[i]);
+		}
+		sql_xfree(desc->errors);
+		assert(end - buf == size);
+		mem_set_array_allocated(desc->err, buf, size);
+	} else {
+		mem_set_null(desc->err);
+	}
+
+	uint32_t size = mp_sizeof_array(desc->statement_count);
+	for (uint32_t i = 0; i < desc->statement_count; ++i)
+		size += mp_sizeof_str(strlen(desc->statements[i]));
+	char *buf = sql_xmalloc(size);
+	char *end = mp_encode_array(buf, desc->statement_count);
+	for (uint32_t i = 0; i < desc->statement_count; ++i) {
+		end = mp_encode_str0(end, desc->statements[i]);
+		sql_xfree(desc->statements[i]);
+	}
+	sql_xfree(desc->statements);
+	assert(end - buf == size);
+	mem_set_array_allocated(desc->ret, buf, size);
+}
+
+/** Add a field foreign key constraint to the statement description. */
+static void
+sql_describe_field_foreign_key(struct sql_desc *desc,
+			       const struct tuple_constraint_def *cdef)
+{
+	assert(cdef->type == CONSTR_FKEY && cdef->fkey.field_mapping_size == 0);
+	const struct space *foreign_space = space_by_id(cdef->fkey.space_id);
+	assert(foreign_space != NULL);
+	const struct tuple_constraint_field_id *field = &cdef->fkey.field;
+	if (field->name_len == 0 &&
+	    field->id >= foreign_space->def->field_count) {
+		const char *err = "foreign field is unnamed";
+		sql_desc_error(desc, "foreign key", cdef->name, err);
+		return;
+	}
+
+	const char *field_name = field->name_len > 0 ? field->name :
+				 foreign_space->def->fields[field->id].name;
+	sql_desc_append(desc, " CONSTRAINT ");
+	sql_desc_append_name(desc, cdef->name);
+	sql_desc_append(desc, " REFERENCES ");
+	sql_desc_append_name(desc, foreign_space->def->name);
+	sql_desc_append(desc, "(");
+	sql_desc_append_name(desc, field_name);
+	sql_desc_append(desc, ")");
+}
+
+/** Add a tuple foreign key constraint to the statement description. */
+static void
+sql_describe_tuple_foreign_key(struct sql_desc *desc,
+			       const struct space_def *def, int i)
+{
+	struct tuple_constraint_def *cdef = &def->opts.constraint_def[i];
+	assert(cdef->type == CONSTR_FKEY && cdef->fkey.field_mapping_size > 0);
+	const struct space *foreign_space = space_by_id(cdef->fkey.space_id);
+	assert(foreign_space != NULL);
+	bool is_error = false;
+	for (uint32_t i = 0; i < cdef->fkey.field_mapping_size; ++i) {
+		const struct tuple_constraint_field_id *field =
+			&cdef->fkey.field_mapping[i].local_field;
+		if (field->name_len == 0 &&
+		    field->id >= foreign_space->def->field_count) {
+			sql_desc_error(desc, "foreign key", cdef->name,
+				       "local field is unnamed");
+			is_error = true;
+		}
+		field = &cdef->fkey.field_mapping[i].foreign_field;
+		if (field->name_len == 0 &&
+		    field->id >= foreign_space->def->field_count) {
+			sql_desc_error(desc, "foreign key", cdef->name,
+				       "foreign field is unnamed");
+			is_error = true;
+		}
+	}
+	if (is_error)
+		return;
+
+	assert(def->field_count != 0);
+	sql_desc_append(desc, ",\nCONSTRAINT ");
+	sql_desc_append_name(desc, cdef->name);
+	sql_desc_append(desc, " FOREIGN KEY(");
+	for (uint32_t i = 0; i < cdef->fkey.field_mapping_size; ++i) {
+		const struct tuple_constraint_field_id *field =
+			&cdef->fkey.field_mapping[i].local_field;
+		const char *field_name = field->name_len != 0 ? field->name :
+					 def->fields[field->id].name;
+		if (i > 0)
+			sql_desc_append(desc, ", ");
+		sql_desc_append_name(desc, field_name);
+	}
+
+	sql_desc_append(desc, ") REFERENCES ");
+	sql_desc_append_name(desc, foreign_space->def->name);
+	sql_desc_append(desc, "(");
+	for (uint32_t i = 0; i < cdef->fkey.field_mapping_size; ++i) {
+		const struct tuple_constraint_field_id *field =
+			&cdef->fkey.field_mapping[i].foreign_field;
+		assert(field->name_len != 0 || field->id < def->field_count);
+		const char *field_name = field->name_len != 0 ? field->name :
+					 def->fields[field->id].name;
+		if (i > 0)
+			sql_desc_append(desc, ", ");
+		sql_desc_append_name(desc, field_name);
+	}
+	sql_desc_append(desc, ")");
+}
+
+/** Add a field check constraint to the statement description. */
+static void
+sql_describe_field_check(struct sql_desc *desc, const char *field_name,
+			 const struct tuple_constraint_def *cdef)
+{
+	assert(cdef->type == CONSTR_FUNC);
+	const struct func *func = func_by_id(cdef->func.id);
+	if (func->def->language != FUNC_LANGUAGE_SQL_EXPR) {
+		sql_desc_error(desc, "check constraint", cdef->name,
+			       "wrong constraint expression");
+		return;
+	}
+	if (!func_sql_expr_has_single_arg(func, field_name)) {
+		sql_desc_error(desc, "check constraint", cdef->name,
+			       "wrong field name in constraint expression");
+		return;
+	}
+
+	sql_desc_append(desc, " CONSTRAINT ");
+	sql_desc_append_name(desc, cdef->name);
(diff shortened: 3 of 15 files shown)


src/box/sql/show.c
```diff
@@ -0,0 +1,415 @@
...
+		sql_err_append(desc, "Foreign key", cdef->name);
+		return;
+	}
+	sql_desc_append(desc, "", cdef->name);
+	const char *field_name = field->name_len != 0 ? field->name :
+				 foreign_space->def->fields[field->id].name;
+	sql_desc_append(desc, " CONSTRAINT \"%s\" REFERENCES \"%s\"(\"%s\")",
+			cdef->name, foreign_space->def->name, field_name);
+}
+
```
Gerold103, line 142: I would say this behaviour is quite unexpected:
```
tarantool> box.execute('SHOW CREATE TABLE "404"')
---
- metadata:
  - name: CREATE TABLE statement
    type: string
  rows: []
...
```
```
tarantool> s = box.schema.create_space('test')
---
...

tarantool> box.execute('SHOW CREATE TABLE "test"')
---
- metadata:
  - name: CREATE TABLE statement
    type: string
  rows:
  - ['/* Space "test" cannot be defined in this statement. */']
...
```
Why aren't these normal errors? Users have no sane way to see if the statement has failed. They would need to check in their code for no rows and for `/*` in the beginning of the statement?

src/box/sql/build.c
```diff
@@ -3389,3 +3389,55 @@ sql_setting_set(struct Parse *parse_context, struct Token *name,
...
+	sqlVdbeAddOp2(v, OP_Integer, BOX_SYSTEM_ID_MAX, key_reg);
+	int addr1 = sqlVdbeAddOp4Int(v, OP_SeekGT, cursor, 0, key_reg, 1);
+	int space_id_reg = ++parse->nMem;
+	int addr2 = sqlVdbeAddOp3(v, OP_Column, cursor, BOX_SPACE_FIELD_ID,
+				  space_id_reg);
+	int result_reg = ++parse->nMem;
+	sqlVdbeAddOp2(v, OP_ShowTable, space_id_reg, result_reg);
+	sqlVdbeAddOp2(v, OP_ResultRow, result_reg, 1);
+	sqlVdbeAddOp2(v, OP_Next, cursor, addr2);
+	sqlVdbeJumpHere(v, addr1);
```
Gerold103, line 3442: Why do you need so many additional opcodes if you anyway added `OP_ShowTable`? Couldn't space name simply be its direct argument?

Gerold103 (changes requested):
Thanks for the patch!

src/box/sql/show.c
```diff
@@ -0,0 +1,415 @@
...
+	assert(foreign_space != NULL);
+	const struct tuple_constraint_field_id *field = &cdef->fkey.field;
+	if (field->name_len > 0) {
+		uint32_t id;
+		uint32_t hash = field_name_hash(field->name, field->name_len);
+		if (tuple_fieldno_by_name(foreign_space->def->dict, field->name,
+					  field->name_len, hash, &id) != 0) {
+			sql_err_append(desc, "Foreign key", cdef->name);
+			return;
+		}
```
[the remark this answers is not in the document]
ImeevMA: This case is checked in the test on line 154. This is when the outer space does not have a field referenced by the local space. This situation is possible in Lua, because the core FK checks the availability of a field at run time, not at creation.

src/box/sql/show.c
```diff
@@ -0,0 +1,415 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
```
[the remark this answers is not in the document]
ImeevMA: I add only one `\n`, I think the second one added when the result is shown in console:
```
tarantool> box.execute([[create table t(i int primary key, a int);]])
---
- row_count: 1
...

tarantool> print(box.execute([[show create table t;]]).rows[1][1])
CREATE TABLE "T"(
"I" INTEGER NOT NULL,
"A" INTEGER,
CONSTRAINT "pk_unnamed_T_1" PRIMARY KEY("I"));
---
...
tarantool> box.execute([[show create table t;]]).rows[1][1]
---
- 'CREATE TABLE "T"(

  "I" INTEGER NOT NULL,

  "A" INTEGER,

  CONSTRAINT "pk_unnamed_T_1" PRIMARY KEY("I"));'
...
```

src/box/sql/show.c
ImeevMA: I don't think the statement failed in the second case. As for the first one, since we didn't find anything, it seems logical to me that we return empty result. Either way, let's ask @unera.

src/box/sql/build.c
Gerold103: Ah, I see, it is called for each tuple during iteration. Not once per vdbe exec.

src/box/sql/show.c
Gerold103: But this is not a `"CREATE TABLE"` statement. So the request couldn't return what was expected. That is an error. You even literally call this message in your code `/** Buffer for errors. */`.

changelogs/unreleased/gh-8098-show-create-table.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/sql
+
+* Now field type ANY supports collation in SQL (gh-8070).
```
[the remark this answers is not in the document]
ImeevMA: True, fixed!

src/box/sql/build.c
```diff
@@ -3389,3 +3389,55 @@ sql_setting_set(struct Parse *parse_context, struct Token *name,
...
 	return;
 }
+
+void
+sql_emit_show_create_table_one(struct Parse *parse, struct Token *name)
+{
+	struct Vdbe *v = sqlGetVdbe(parse);
+	char *space_name = sql_name_from_token(name);
+	sqlVdbeSetNumCols(v, 1);
+	vdbe_metadata_set_col_name(v, 0, "CREATE TABLE statement");
```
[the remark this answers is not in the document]
ImeevMA: Thanks, replaced by "statement".

src/box/sql/show.c
```diff
@@ -0,0 +1,415 @@
...
+		desc->size *= 2;
+	desc->buf = sql_xrealloc(desc->buf, desc->size);
+	buf = &desc->buf[desc->len];
+	size = desc->size - desc->len;
+	va_start(ap, fmt);
+	rc = vsnprintf(buf, size, fmt, ap);
+	va_end(ap);
+end:
+	desc->len += rc;
+	assert(rc >= 0 && desc->len < desc->size);
```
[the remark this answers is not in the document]
ImeevMA: True, reworked using `struct StrAccum` and `sqlStrAccum*()` functions.

src/box/sql/show.c
```diff
@@ -0,0 +1,415 @@
+/*
```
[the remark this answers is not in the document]
ImeevMA: Thanks for noticing and suggesting! Fixed.

src/box/sql/show.c
ImeevMA: Replaced error messages by actual errors. However, this actually affects `SHOW CREATE TABLE;` statement, since one wrong space and no information will be given. Maybe we allow error messages for all cases except `Space 'space_name' does not exist`?

Replaced error messages with real errors. However, this actually affects the `SHOW CREATE TABLE;` statement, as one wrong space and no proper result will be provided. Maybe we allow error messages for all cases except `Space 'space_name' does not exist`?

unera: I agree, unknown name should throw an error.

unera: update: we should generate an error only if `SHOW CREATE TABLE t` links to unknown (or inaccessible) `t`.

The operator can contain some comments:
- about indexes, that can not be created by SQL
- about any the other problems

In an extreme case, this statement will only contain comments

src/box/sql/show.c
```diff
@@ -0,0 +1,425 @@
...
+static void
+sql_desc_initialize(struct sql_desc *desc)
+{
+	sqlStrAccumInit(&desc->acc, NULL, 0, SQL_MAX_LENGTH);
+	sqlStrAccumInit(&desc->err, NULL, 0, SQL_MAX_LENGTH);
+}
+
+/** Append a new string to the object used to accumulate a statement. */
+static void
+sql_desc_append(struct sql_desc *desc, const char *fmt, ...)
```
Gerold103, line 35: To be on the safe side I suggest to use here `CFORMAT` attribute (from `trivia/util.h`). It will validate the format string and args at compile time.

src/box/sql/show.c
```diff
@@ -0,0 +1,425 @@
...
+static void
+sql_describe_field(struct sql_desc *desc, const struct field_def *field)
+{
+	char type_str[TYPE_STR_SIZE];
+	const char *type_str_original = field_type_strs[field->type];
+	size_t type_str_len = strlen(type_str_original);
+	assert(type_str_len < TYPE_STR_SIZE);
+	for (size_t i = 0; i < type_str_len; ++i)
+		type_str[i] = toupper(type_str_original[i]);
+	type_str[type_str_len] = '\0';
```
Gerold103, line 230: We already have `strtolower()` / `strtolowerdup()` in `trivia/util.h`. Maybe worth adding `strtoupper()` / `strtoupperdup()`.

src/box/sql/show.c
```diff
@@ -0,0 +1,425 @@
...
+		    sdef->max != INT64_MAX || sdef->cache != 0 || sdef->cycle ||
+		    strcmp(sdef->name, space->def->name) != 0)
+			sql_err_append(desc, "Sequence", sdef->name);
+		else
+			has_sequence = true;
+	}
+
+	char *pk_name = sql_escaped_name_new(pk->def->name);
+	sql_desc_append(desc, ",\nCONSTRAINT %s PRIMARY KEY(", pk_name);
+	sql_xfree(pk_name);
```
Gerold103, line 280: There are quite many of these triplets - name new + append + xfree. I would suggest to introduce `sql_desc_append_escape(desc, name)` which would the the same. It would force you to write sometimes not much less code, but you won't have to keep track of all those xfree:
```C
sql_desc_append(desc, ",\nCONSTRAINT ");
sql_desc_append_escape(desc, pk->def->name);
sql_desc_append(desc, "PRIMARY KEY(");
```
In some places it will turn into just 1 or 2 lines. Up to you.

src/box/sql/show.c
```diff
@@ -0,0 +1,425 @@
...
+		if (pk->def->key_def->parts[i].coll_id != coll_id) {
+			diag_set(ClientError, ER_UNSUPPORTED, "SQL",
+				 "indexes with part collation not equal to "
+				 "corresponding field collation");
+			return NULL;
+		}
+	}
+
+	if (strcmp(space->engine->name, "memtx") != 0 &&
+	    strcmp(space->engine->name, "vinyl") != 0) {
```
Gerold103, line 389: `space_is_memtx()` and `space_is_vinyl()` from `space.h`.

test/sql-luatest/show_create_table_test.lua
```diff
@@ -0,0 +1,456 @@
...
+        local res1 = [[CREATE TABLE "T1"("I" INTEGER NOT NULL,"A" INTEGER,]]..
+                     [[CONSTRAINT "pk_unnamed_T1_1" PRIMARY KEY("I"));]]
+        local res2 = [[CREATE TABLE "T2"("I" INTEGER NOT NULL,"A" INTEGER,]]..
+                     [[CONSTRAINT "pk_unnamed_T2_1" PRIMARY KEY("I"));]]
+        local res3 = [[CREATE TABLE "T3"("I" INTEGER NOT NULL,"A" INTEGER,]]..
+                     [[CONSTRAINT "pk_unnamed_T3_1" PRIMARY KEY("I"));]]
+        t.assert_equals(#ret.rows, 3)
+        t.assert_equals(ret.rows[1][1]:gsub('\n', ''), res1)
+        t.assert_equals(ret.rows[2][1]:gsub('\n', ''), res2)
+        t.assert_equals(ret.rows[3][1]:gsub('\n', ''), res3)
```
Gerold103, line 62: You might want to consider this simplification (you do gsub ~35 times in this file right now):

```Diff
@@ -14,14 +14,26 @@ end)
 
 g.test_show_create_table_one = function()
     g.server:exec(function()
-        local _, err = box.execute([[SHOW CREATE TABLE t;]])
+        rawset(_G, 'test_exec', function(...)
+            local res = box.execute()
+            t.assert(res ~= nil)
+            for i = 1, #res.rows do
+                local field = res.rows[i][1]
+                if field ~= nil then
+                    res.rows[i]:update({{'=', 1, field:gsub('\n', '')}})
+                end
+            end
+            return res
+        end)
+
+        local _, err =  box.execute([[SHOW CREATE TABLE t;]])
         t.assert_equals(err.message, [[Space 'T' does not exist]])
 
         box.execute([[CREATE TABLE t(i INT PRIMARY KEY, a INT);]])
-        local ret = box.execute([[SHOW CREATE TABLE t;]])
+        local ret = _G.test_exec([[SHOW CREATE TABLE t;]])
         local res = [[CREATE TABLE "T"("I" INTEGER NOT NULL,"A" INTEGER,]]..
                     [[CONSTRAINT "pk_unnamed_T_1" PRIMARY KEY("I"));]]
-        t.assert_equals(ret.rows[1][1]:gsub('\n', ''), res)
+        t.assert_equals(ret.rows[1][1], res)
         box.execute([[DROP TABLE t;]])
 
         local sql = [[CREATE TABLE "T"("I" INTEGER NOT NULL,]]..
@@ -35,8 +47,8 @@ g.test_show_create_table_one = function()
                     [[CONSTRAINT "C5" FOREIGN KEY("I", "A") ]]..
                     [[REFERENCES "T"("A", "C")) WITH ENGINE = 'vinyl';]]
         box.execute(sql)
-        ret = box.execute([[SHOW CREATE TABLE t;]])
-        t.assert_equals(ret.rows[1][1]:gsub('\n', ''), sql)
+        ret = _G.test_exec([[SHOW CREATE TABLE t;]])
+        t.assert_equals(ret.rows[1][1], sql)
         box.execute([[DROP TABLE t;]])
     end)
 end
@@ -49,7 +61,7 @@ g.test_show_create_table_all = function()
         box.execute([[CREATE TABLE t1(i INT PRIMARY KEY, a INT);]])
         box.execute([[CREATE TABLE t2(i INT PRIMARY KEY, a INT);]])
         box.execute([[CREATE TABLE t3(i INT PRIMARY KEY, a INT);]])
-        local ret = box.execute([[SHOW CREATE TABLE;]])
+        local ret = _G.test_exec([[SHOW CREATE TABLE;]])
         local res1 = [[CREATE TABLE "T1"("I" INTEGER NOT NULL,"A" INTEGER,]]..
                      [[CONSTRAINT "pk_unnamed_T1_1" PRIMARY KEY("I"));]]
         local res2 = [[CREATE TABLE "T2"("I" INTEGER NOT NULL,"A" INTEGER,]]..
@@ -57,9 +69,9 @@ g.test_show_create_table_all = function()
         local res3 = [[CREATE TABLE "T3"("I" INTEGER NOT NULL,"A" INTEGER,]]..
                      [[CONSTRAINT "pk_unnamed_T3_1" PRIMARY KEY("I"));]]
         t.assert_equals(#ret.rows, 3)
-        t.assert_equals(ret.rows[1][1]:gsub('\n', ''), res1)
-        t.assert_equals(ret.rows[2][1]:gsub('\n', ''), res2)
-        t.assert_equals(ret.rows[3][1]:gsub('\n', ''), res3)
+        t.assert_equals(ret.rows[1][1], res1)
+        t.assert_equals(ret.rows[2][1], res2)
+        t.assert_equals(ret.rows[3][1], res3)
```
and etc similarly.

test/sql-luatest/show_create_table_test.lua
```diff
@@ -0,0 +1,456 @@
...
+        box.space.b:create_index('i', {parts = {{'i', 'integer'}}})
+        local ret = box.execute([[SHOW CREATE TABLE "b";]])
+        local res = [[CREATE TABLE "b"("i" INTEGER NOT NULL ]]..
+                    [[CONSTRAINT "a" REFERENCES "a"("i"),]]..
+                    [[CONSTRAINT "i" PRIMARY KEY("i"));]]
+        t.assert_equals(ret.rows[1][1]:gsub('\n', ''), res)
+
+        res = [[CREATE TABLE "b"("i" INTEGER NOT NULL,]]..
+              [[CONSTRAINT "i" PRIMARY KEY("i"));]]..
+              [[/* Foreign key "a" cannot be defined in this statement. */]]
```
Gerold103, line 166: AFAIU, the purpose of `SHOW CREATE TABLE` usually is to export the schema into an SQL script and then re-apply it anew in some other place. And I still see problems with it in the current API:

* If you silently ignore non-printable parts like unknown foreign keys, then you generate a valid but just wrong SQL statement. People will print it into a script, apply, won't see any errors, and then stuff will explode in their tests. Not at the time of the SQL script generation. This is sad. It makes the feature hardly usable as is, I think.

* If a space couldn't be serialized into SQL, as I can see, you still return it as an error-comment. This is fixed for the individual `SHOW CREATE TABLE <name>`, but not for showing all tables. Again, the scenario about schema export will break here. People will need to write boilerplate code around this to check for "statements" which start from `/*` and thus aren't really statements.

I understand the problems though. You want to return the error somehow. So I would suggest to either fail if anything couldn't be serialized, or just skip such spaces from the result (sad, no full schema, but the returned spaces will be at least full and valid), or return 2 fields: `statement` and `error` or `message`. `statement` is always fully serialized space-SQL or `nil`. `error` / `message` is either `nil` or the message. Then people at least will be able to check for individual errors normally, instead of doing `res.rows[i]:starts('/*')` which is just extremely weird. Note, that even this hack wouldn't help against the code block which I commented. When you appended the error somewhere in the end.

You summoned your product-pokemon last time? ;) Now lets do a battle - I summon mine! @Mons, ATTACK! (please)

Gerold103 (commented):
Thanks for the fixes!

src/box/sql/show.c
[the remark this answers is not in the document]
Gerold103: But you are not **doing** the create-table here, are you? You just **generate** the statement. You don't execute it. So what is the problem anyway? You have the name - you can use it. You check it now in 3 places (which I would recommend to extract into a function if you intend to keep them even after this comment), but never use the resulting fieldno. Because the field name was enough. Why care in this `SHOW` statement if the name matches with any fieldno if you don't use it? If this field will be missing at the time of the resulting create-table execution (if it will be executed at all), then just let it fail there.

I can expand my reasoning - *"why try to guess what CREATE TABLE fails at, if you are not doing it here?"*. You only need to serialize the definition precisely. Fail only when it is not possible (like field type mismatch in index and space). This concerns not only FK fieldnos. There might be other restrictions which CREATE TABLE has and you do not account them here. For instance, we have a max column count limit - `SQL_MAX_COLUMN`. You don't check it here, but the resulting `box.execute(res.rows[1][1])` will fail if you serialized a non-SQL space with more columns.

Up to you. Those checks just seem as harmless as useless.

test/sql-luatest/show_create_table_test.lua
unera: > If you silently ignore non-printable parts like unknown foreign keys, then you generate a valid but just wrong SQL statement.

some things is not provided by `CREATE TABLE` operator by design. For example non-unique indexes, etc.

So the pattern will never work in ALL user scenarios. I think it is a normal situation.
Worse if a user asks the list and receive an error because one item of the list contains non-printable index.

Mons: I mostly agree with @Gerold103. Explicit error in such a case is much better than silent hidden rake of wrong SQL.
Nevertheless I'd like to have partially dumped statement for such "undumpable" spaces, rather than none at all.
So, I'd expect explicit error in case of some things, that could not be dumped. And additional modifier, that will dump a space anyway, with errors shown as comments (like `show create table <identifier> ignore erorrs`)

ImeevMA: I discussed with @unera and we decided to introduce 4 versions of `SHOW CREATE TABLE`:
1) `SHOW CREATE TABLE <table_name>;` - return a description of the space. On any problem, it gives an error;
2) `SHOW CREATE TABLE <table_name> INCLUDING ERRORS;` - return a description of the space. If the table does not exist, an error is thrown, any other problems are included as comments to `CREATE TABLE`;
3) `SHOW CREATE TABLE;` - return descriptions of all non-system available spaces that can be described with `SHOW CREATE TABLE <table_name>;`. Spaces that cannot be described with `SHOW CREATE TABLE <table_name>;` are ignored;
4) `SHOW CREATE TABLE INCLUDING ERRORS;` -  return descriptions of all non-system available spaces. Any problems are included as comments to `CREATE TABLE`.

There is one more important change: if the INCLUDING ERRORS clause is used, the description will be returned even if the description is not working `CREATE TABLE`. For example, for a space created with `box.schema.space.create('a')`, the description would be:
```
CREATE TABLE "a";
/* Space "a" cannot be defined in this statement: PRIMARY KEY missing */
/* Space "a" cannot be defined in this statement: SQL does not support space without format */
```

unera: >
```sql
CREATE TABLE "a";
/* Space "a" cannot be defined in this statement: PRIMARY KEY missing */
/* Space "a" cannot be defined in this statement: SQL does not support space without format */
```

Could we place comments into brackets?

```sql
CREATE TABLE "a" (
/* Space "a" cannot be defined in this statement: PRIMARY KEY missing */
/* Space "a" cannot be defined in this statement: SQL does not support space without format */
);
```

unera: If a table is created, how it can throw an error 'Wrong engine'?

Anyway, we can place the comment inside brackets (As the first or last comment).

ImeevMA: For example:
```
SHOW CREATE TABLE "_space";
```

Gerold103: Why so complicated? Did you all read my comment fully? Citing, I wrote:

"... *or return 2 fields: `statement` and `error` or `message`. `statement` is always a fully serialized space-SQL or `nil`. `error` / `message` is either `nil` or an error message. Then people at least will be able to check for individual errors normally* ...".

What is wrong with this? You get complete error messages + you don't fail the whole `SHOW CREATE TABLE` because of just one space + you don't have incorrect SQL statements. And you don't need to introduce multiple versions of `SHOW` with weird syntax.

Mons: > Why so complicated?

I think, because in case of minor errors I'd like to receive my SQL statement with some comments, not nil and error message and then write this statement by hands. There are often no ability to fix an issue to make dump successful.

src/box/sql/show.c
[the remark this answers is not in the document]
ImeevMA: Removed this check.

src/box/sql/show.c
ImeevMA: Thanks, added.

src/box/sql/show.c
ImeevMA: Thanks! Done.

src/box/sql/util.c
```diff
@@ -1244,3 +1244,25 @@ sqlVListNameToNum(VList * pIn, const char *zName, int nName)
...
+
+char *
+sql_escaped_name_new(const char *name)
+{
+	size_t len = strlen(name);
+	size_t count = 0;
+	for (size_t i = 0; i < len; ++i) {
+		if (name[i] == '"')
+			++count;
+	}
```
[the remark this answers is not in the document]
ImeevMA: Done. If the first character is a letter, the name is equal to its normalized version, and the escaped version is exactly 2 bytes longer, then I use the normalized name.

test/sql-luatest/show_create_table_test.lua
ImeevMA: Instead, I decided to add '\n' in expected result. However, I still added two helper functions.

test/sql-luatest/show_create_table_test.lua
ImeevMA: I'm not sure if we should return both a not-NULL result and an error in the same answer. Also, there is a use case similar to this case in the downgrade feature.

Gerold103 (commented):
Thanks for the fixes!

src/box/sql/show.c
```diff
@@ -0,0 +1,494 @@
...
+	}
+	if (desc->err.nChar > 0) {
+		char *err = sqlStrAccumFinish(&desc->err);
+		sqlXPrintf(&desc->acc, "%s", err);
+		sql_xfree(err);
+	}
+
+	const char *engine = space_is_vinyl(desc->space) ?
+			     " WITH ENGINE = 'vinyl'" : "";
+	sqlXPrintf(&desc->acc, ")%s;", engine);
```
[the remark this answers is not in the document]
ImeevMA: Thank you for noticing! Fixed.

src/box/sql/show.c
```diff
@@ -0,0 +1,494 @@
...
+	bool is_error = false;
+	const struct func *func = func_by_id(cdef->func.id);
+	if (func->def->language != FUNC_LANGUAGE_SQL_EXPR) {
+		sql_desc_error(desc, "check constraint", cdef->name,
+			       "wrong constraint expression");
+		is_error = true;
+	} else if (!func_sql_expr_has_single_arg(func, field_name)) {
+		sql_desc_error(desc, "check constraint", cdef->name,
+			       "wrong field name in constraint expression");
+		is_error = true;
```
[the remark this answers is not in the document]
ImeevMA: No, this is actually different. In ANSI query
```
CREATE TABLE t(i INT PRIMARY KEY, a INT CHECK(i > 10));
```
is the same as
```
CREATE TABLE t(i INT PRIMARY KEY, a INT, CHECK(i > 10));
```
since all column constraints are just another syntax for table constraints. This is not so for us. For example, we can set constraint "i > 10" in Lua for field `a` and it will check field `a`, since no names are checked in core field constraints. However, if we allow `SHOW CREATE TABLE` for such cases we receive
```
CREATE TABLE t(i INT PRIMARY KEY, a INT CHECK(i > 10));
```
which could not be executed from out standpoint and will have another result for ANSI standpoint.

If only we were able to avoid so-called "field constraints", we wouldn't have such problems. But we seek for speed here, it seems.

src/box/sql/show.c
```diff
@@ -0,0 +1,494 @@
...
+	if (index->def->type != TREE) {
+		const char *err = "unsupported index type";
+		sql_desc_error(desc, "index", index->def->name, err);
+		is_error = true;
+	}
+	if (!index->def->opts.is_unique) {
+		const char *err = "non-unique index";
+		sql_desc_error(desc, "index", index->def->name, err);
+		is_error = true;
+	}
```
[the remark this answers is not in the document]
ImeevMA: Originally this statement was inspired by `SHOW CREATE TABLE` from MySQL. This statement in MySQL actually ignores indexes that cannot be described in `CREATE TABLE`.

There is also `.schema` and `.schema table_name` which return all `CREATE TABLE` and `CREATE INDEX` for all spaces and spaces with the given name, respectively.

I suggested doing something that works like `.schema table_name`, but `SHOW CREATE TABLE` seems more appealing to PM.

src/box/sql/show.c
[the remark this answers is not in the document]
Gerold103: So like I said, it can't be executed anyway. But you still care to validate it. Which means this is exactly like the situation with fieldno vs name was, and exactly like the tuple constraints which currently are not validated and won't be executed either (I gave you the example above). I don't see why keep this particular validation, if others are not present or don't help anyway, just like this one.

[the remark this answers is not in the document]
ImeevMA: I thought we decided that if the result was returned by `SHOW CREATE TABLE tablename;`, then that means the result should be executable. If I skip the check, this rule may be violated. Not sure if this is correct.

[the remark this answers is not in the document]
Gerold103: This is already violated. I gave you an example above. Can copy-paste it again:
```Lua
box.cfg{}
body = 'A > 100 AND A < 200'
box.schema.func.create('check', {is_deterministic = true, body = body, language = 'SQL_EXPR'})
s = box.schema.create_space('test', {
    format = {{'ID', 'unsigned'}},
    constraint = {check = 'check'}})
_ = s:create_index('pk')
box.execute('SHOW CREATE TABLE "test"')
```
```
---
- metadata:
  - name: statement
    type: string
  rows:
  - ['CREATE TABLE "test"(

      ID UNSIGNED NOT NULL,

      CONSTRAINT "pk" PRIMARY KEY(ID),

      CONSTRAINT "check" CHECK(A > 100 AND A < 200));']
...
```
The returned SQL statement can't be executed. It will throw an error. That means the `SHOW CREATE TABLE` returned a bad result.

Gerold103 (approved):
Thanks for the fixes! I still see problems here which make the feature barely usable except manually in an interactive console (as an alternative to looking directly at `_space` content). But I don't want to argue, so ack.

src/box/sql/show.c
```diff
@@ -0,0 +1,497 @@
...
+		sqlXPrintf(&desc->acc, "%s", err);
+		sql_xfree(err);
+	}
+
+	if (space_is_memtx(desc->space))
+		sqlXPrintf(&desc->acc, ")\nWITH ENGINE = 'memtx';");
+	else if (space_is_vinyl(desc->space))
+		sqlXPrintf(&desc->acc, ")\nWITH ENGINE = 'vinyl';");
+	else
+		sqlXPrintf(&desc->acc, ");");
```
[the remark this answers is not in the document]
ImeevMA: No, this is for the case when an error "wrong space engine" is printed. We still have to end statement with `);`.

src/box/sql/show.c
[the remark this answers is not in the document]
ImeevMA: I will fill a bug. I am not sure which of these are more correct.

Gerold103 (approved):
Thanks for the patch! One comment, then need a second review.

changelogs/unreleased/gh-8098-show-create-table.md
```diff
@@ -0,0 +1,3 @@
+## feature/sql
+
+* Statement SHOW CREATE TABLE is introduced (gh-8098).
```
p7nov, line 3:
```suggestion
* Introduced the `SHOW CREATE TABLE` statement (gh-8098).
```

p7nov (approved):

src/box/sql/show.c
```diff
@@ -0,0 +1,499 @@
...
+	uint32_t id = desc->statement_count;
+	++desc->statement_count;
+	uint32_t size = desc->statement_count * sizeof(desc->statements);
+	desc->statements = sql_xrealloc(desc->statements, size);
+	desc->statements[id] = str;
+}
+
+/** Finalize a described statement. */
+static void
+sql_desc_finalize(struct sql_desc *desc)
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! Fixed. This is not necessary as `acc` is only populated after all the checks have passed, but there is no proper way to check that `acc` is empty, so adding `sqlStrAccumReset()` seems to be the only correct way to make sure nothing is leaked.

changelogs/unreleased/gh-8098-show-create-table.md
ImeevMA: Thanks! Fixed.

p7nov (approved):

locker (approved):

locker:
@ImeevMA There are compilation errors on some platforms. Please fix.

ImeevMA:
@locker fixed!
<details>
<summary>Diff</summary>

```DIff
diff --git a/src/box/sql/show.c b/src/box/sql/show.c
index 4d2d0a4a4..668c2f1a1 100644
--- a/src/box/sql/show.c
+++ b/src/box/sql/show.c
@@ -7,14 +7,10 @@
 
 #include "sqlInt.h"
 #include "mem.h"
-#include "schema.h"
-#include "sequence.h"
-#include "coll_id_cache.h"
-#include "tuple_constraint_def.h"
-
-enum {
-       TYPE_STR_SIZE = 16,
-};
+#include "box/schema.h"
+#include "box/sequence.h"
+#include "box/coll_id_cache.h"
+#include "box/tuple_constraint_def.h"
 
 /** An objected used to accumulate a statement. */
 struct sql_desc {
```

</details>

ImeevMA:
@locker sorry, I found out that I wrote `Part of` instead of `Closes`. Will fix.

PR #8454, opened 2023-03-17
box: fix wrong validation of region_alloc() result

ImeevMA:
Prior to this patch, the return value of region_alloc() in lbox_tuple_format_new() was not checked. This patch fixes this by replacing region_alloc() with xregion_alloc().

Closes tarantool/security#116

NO_DOC=bugfix
NO_TEST=hard to reproduce the bug

### DIFF
--- a/src/box/lua/misc.cc
+++ b/src/box/lua/misc.cc
@@ -373,12 +373,10 @@ lbox_tuple_format_new(struct lua_State *L)
 	size_t size;
 	struct region *region = &fiber()->gc;
 	size_t region_svp = region_used(region);
-	struct field_def *fields = region_alloc_array(region, typeof(fields[0]),
-						      count, &size);
-	if (fields == NULL) {
-		diag_set(OutOfMemory, size, "region_alloc_array", "fields");
-		return luaT_error(L);
-	}
+	struct field_def *fields =
+		(struct field_def *)xregion_alloc_array(region,
+							struct field_def, count,
+							&size);
 	for (uint32_t i = 0; i < count; ++i) {
 		size_t len;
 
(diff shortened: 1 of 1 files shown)


src/box/lua/misc.cc
```diff
@@ -400,13 +400,7 @@ lbox_tuple_format_new(struct lua_State *L)
...
 		assert(! lua_isnil(L, -1));
 		const char *name = lua_tolstring(L, -1, &len);
-		fields[i].name = (char *)region_alloc(region, len + 1);
-		if (fields == NULL) {
-			diag_set(OutOfMemory, size, "region_alloc",
-				 "fields[i].name");
-			region_truncate(region, region_svp);
-			return luaT_error(L);
-		}
+		fields[i].name = (char *)xregion_alloc(region, len + 1);
```
locker, line 403: Please also replace `region_alloc_array` with `xregion_alloc_array` at line 376 while you are at it.

locker (changes requested):

src/box/lua/misc.cc
ImeevMA: Fixed. However, I added an explicit cast for the return value of xregion_alloc_array() because otherwise a compilation error occurs. I think it's because it's a cc file. There were no errors for region_alloc_array().

changelogs/unreleased/ghs-116-possible-segfault-in-misc.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* Fixed possible segmentation fault due to unvalidated return value of
+  region_alloc() (ghs-116).
```
locker, line 4: Let's drop the changelog? The bug shouldn't normally happen. And replacing segfault with panic doesn't look like a fix, actually. So this would only clutter the changelog.

locker (approved):

locker:

PR #8455, opened 2023-03-17
[2.10] sql: add return value check for sql_key_info_new()

ImeevMA:
Prior to this patch, the return value of sql_key_info_new() in multiSelectOrderBy() was not checked. This patch fixes that.

Closes tarantool/security#117

NO_DOC=bugfix
NO_TEST=hard to reproduce the bug

### DIFF
--- a/src/box/sql/select.c
+++ b/src/box/sql/select.c
@@ -3664,11 +3664,13 @@ multiSelectOrderBy(Parse * pParse,	/* Parsing context */
 		pParse->nMem += expr_count + 1;
 		sqlVdbeAddOp2(v, OP_Bool, 0, regPrev);
 		key_info_dup = sql_key_info_new(db, expr_count);
-		if (key_info_dup != NULL) {
-			for (int i = 0; i < expr_count; i++) {
-				key_info_dup->parts[i].coll_id =
-					multi_select_coll_seq(pParse, p, i);
-			}
+		if (key_info_dup == NULL) {
+			pParse->is_aborted = true;
+			return 1;
+		}
+		for (int i = 0; i < expr_count; i++) {
+			key_info_dup->parts[i].coll_id =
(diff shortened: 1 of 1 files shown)


changelogs/unreleased/ghs-117-possible-segfault-in-multiSelectOrderBy.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
```
locker, line 1: Looks more like bugfix/sql.

locker (changes requested):
Why only for 2.10? Please explain this in the commit message.

changelogs/unreleased/ghs-117-possible-segfault-in-multiSelectOrderBy.md
ImeevMA: Fixed.

ImeevMA:
@locker why in commit-message? In 2.11 and later we have patch [sql: refactor memory allocation system](https://github.com/tarantool/tarantool/commit/91fd360ca80ead5f7009d95e7084fcace2d65b94), which fixes all similar problems. For me it doesn't look like something that should be written in commit-message.

changelogs/unreleased/ghs-117-possible-segfault-in-multiSelectOrderBy.md
locker: Come to think of it, the changelog looks useless, see https://github.com/tarantool/tarantool/pull/8454/commits/08b7bc9daffa6547af8e383df61c58168c182912#r1139924387. Let's just drop it?

ImeevMA: Done.

locker (approved):

PR #8456, opened 2023-03-17
 sql: fix wrong int to dec conversion

### LINKED ISSUE #8460: Wrong int to decimal conversion in arithmetic operations
In arithmetic operations, an integer greater than INT64_MAX or less than 0 is incorrectly converted to a decimal number, 

Reproducer:
```
tarantool> box.execute([[SELECT 0.0 + 18446744073709551615;]])
---
- metadata:
  - name: COLUMN_1
    type: decimal
  rows:
  - [-1.0]
...

tarantool> box.execute([[SELECT 0.0 + -1;]])
---
- metadata:
  - name: COLUMN_1
    type: decimal
  rows:
  - [-9223372036854775808.0]
...
```

ImeevMA:
This patch fixes incorrect conversion of an integer greater than INT64_MAX or less than 0 to decimal during SQL arithmetic operations.

Closes https://github.com/tarantool/tarantool/issues/8460

### DIFF
--- a/src/box/sql/mem.c
+++ b/src/box/sql/mem.c
@@ -1855,11 +1855,11 @@ mem_get_dec(const struct Mem *mem, decimal_t *d)
 		return 0;
 	}
 	if (mem->type == MEM_TYPE_INT) {
-		decimal_from_int64(d, mem->u.r);
+		decimal_from_int64(d, mem->u.i);
 		return 0;
 	}
 	if (mem->type == MEM_TYPE_UINT) {
-		decimal_from_int64(d, mem->u.u);
+		decimal_from_uint64(d, mem->u.u);
 		return 0;
 	}
 	if (mem->type == MEM_TYPE_DEC) {
--- a/changelogs/unreleased/gh-8460-wrong-int-to-dec-convertion-in-arithmetic.md
+++ b/changelogs/unreleased/gh-8460-wrong-int-to-dec-convertion-in-arithmetic.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* Fixed incorrect conversion of an integer greater than INT64_MAX or
+  less than 0 to decimal during SQL arithmetic operations (gh-8460).
--- a/test/sql-luatest/gh_8460_wrong_int_to_dec_test.lua
+++ b/test/sql-luatest/gh_8460_wrong_int_to_dec_test.lua
@@ -1,0 +1,28 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function()
+    g.server = server:new({alias = 'master'})
+    g.server:start()
+end)
+
+g.after_all(function()
+    g.server:stop()
+end)
+
+g.test_wrong_big_uint_to_dec_convertion = function()
+    g.server:exec(function()
+        local ret = box.execute([[SELECT 1.0 + 18446744073709551615;]])
+        local res = require('decimal').new('18446744073709551616')
+        t.assert_equals(ret.rows[1][1], res)
+    end)
+end
+
+g.test_wrong_neg_int_to_dec_convertion = function()
+    g.server:exec(function()
+        local ret = box.execute([[SELECT 1.0 + -1;]])
+        t.assert_equals(ret.rows[1][1], require('decimal').new(0))
+    end)
+end


src/box/sql.c
```diff
@@ -1415,16 +1415,15 @@ vdbe_field_ref_create(struct vdbe_field_ref *ref, uint32_t capacity)
...
 	assert(space != NULL);
 	assert(idx != NULL);
 	assert(idx->def->space_id == space->def->id);
-	ssize_t tuple_count = index_size(idx);
-	ssize_t space_size = space_bsize(space);
-	ssize_t avg_tuple_size = tuple_count != 0 ?
-				 (space_size / tuple_count) : 0;
+	size_t tuple_count = index_size(idx);
+	size_t space_size = space_bsize(space);
+	size_t avg_tuple_size = tuple_count != 0 ? space_size / tuple_count : 0;
```
locker, line 1426: This would probably result in another SVACE error because `index_size` and `space_bsize` return a value of type `ssize_t`. What are you trying to accomplish here? What possible problems would this patch solve apart from suppressing SVACE? If it can't lead to a bug, then just mark the issue as "Won't fix" or "False positive".

src/box/sql/build.c
```diff
@@ -145,13 +145,14 @@ sql_space_column_is_in_pk(struct space *space, uint32_t column)
...
 {
-	ssize_t len = strlen(zName);
+	size_t len = strlen(zName);
 	if (len > BOX_NAME_MAX) {
 		diag_set(ClientError, ER_IDENTIFIER,
 			 tt_cstr(zName, BOX_INVALID_NAME_MAX));
 		pParse->is_aborted = true;
 		return -1;
 	}
+	assert(len < (size_t)INT32_MAX);
```
locker, line 155: If `zName` is expected to fit in int, you can just mark the SVACE issue as false-positive. This particular change doesn't make the code any better, because you still pass `size_t` for `int` to `identifier_check`.

src/box/sql/mem.c
```diff
@@ -719,8 +719,7 @@ static inline int
 uint_to_dec(struct Mem *mem)
 {
 	assert(mem->type == MEM_TYPE_UINT);
-	int64_t u = mem->u.u;
-	decimal_from_uint64(&mem->u.d, u);
+	decimal_from_uint64(&mem->u.d, mem->u.u);
```
locker, line 722: AFAIR we agreed that `uint64_t -> int64_t -> uint64_t` conversions are safe on any sane arch and so we aren't fixing them (see the SVACE chat).

src/box/sql/util.c
```diff
@@ -1047,7 +1047,7 @@ sql_div_int(int64_t lhs, bool is_lhs_neg, int64_t rhs, bool is_rhs_neg,
 		if (u_res > (uint64_t) INT64_MAX + 1)
 			return -1;
 		*is_res_neg = u_res != 0;
-		*res = -u_res;
+		*res = u_res == (uint64_t)INT64_MAX + 1 ? INT64_MIN : -u_res;
```
locker, line 1050: We agreed (in the SVACE chat) that all sane architectures use binary compliment (AFAIR somebody mentioned that this is even required by the C++ standard). So there's no need to fix this issue.

src/box/sql/mem.c
```diff
@@ -1854,11 +1854,11 @@ mem_get_dec(const struct Mem *mem, decimal_t *d)
 		return 0;
 	}
 	if (mem->type == MEM_TYPE_INT) {
-		decimal_from_int64(d, mem->u.r);
+		decimal_from_int64(d, mem->u.i);
 		return 0;
 	}
 	if (mem->type == MEM_TYPE_UINT) {
-		decimal_from_int64(d, mem->u.u);
+		decimal_from_uint64(d, mem->u.u);
```
locker, line 1861: This doesn't look like a security issue. It's just a bug. I think we should move the ticket to the open-source repository.

locker (changes requested):

src/box/sql.c
ImeevMA: I thought that both return size_t values, but I was wrong - `index_size()` returns a value of type 'ssize_t`. However, this is not the case for `space_bsize()`. But I agree, I will mark this issue as "Won't fix".

src/box/sql/mem.c
ImeevMA: Filled an issue.

src/box/sql/build.c
ImeevMA: Relevant SVACE issue is closed as "Won't fix".

src/box/sql/util.c
ImeevMA: Relevant SVACE issue is closed as "Won't fix".

locker (approved):

locker:
Cherry-picked to 2.11.

PR #8458, opened 2023-03-17
Bump embedded checks

DifferentialOrange:
Merge after #8203 , https://github.com/tarantool/checks/pull/54 and checks module new tag. Be careful since after https://github.com/tarantool/checks/pull/54 merge integration CI will start to fail since override would start to fail, so we need to merge this as fast as possible (or make several consequent bump PRs).


DifferentialOrange:
Closed in favor of #8475

PR #8462, opened 2023-03-17
box: fix big number encoding by msgpackffi

### LINKED ISSUE #6119: Error in comparison too small DOUBLE number with DECIMAL numbers in index.
Tarantool 2.9.0-59-gc5e185474

Too small (less than 1e-38) DOUBLE numbers truncated to 0 when compared to DECIMAL during construction of index.

Reproducer:
```
tarantool> s = box.schema.space.create('a', {format = {{'u', 'number'}}})
---
...

tarantool> _ = s:create_index('i')
---
...

tarantool> s:insert({require('decimal').new(0)})
---
- [0]
...

tarantool> s:insert({1e-39})
---
- error: Duplicate key exists in unique index "i" in space "a" with old tuple - [0]
    and new tuple - [1e-39]
...
```

alyapunov:
Due to a typo some big numbers were coded as MP_(U)INT.

Since msgpackffi is used in selectffi, which is used for memtx, that could lead to strange select results with big number keys.

Closes #6119

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-6119-fix-number-mp-encode.md
+++ b/changelogs/unreleased/gh-6119-fix-number-mp-encode.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed a bug when big numbers were incorrectly encoded by msgpackffi,
+  that could lead to wrong select results with big number keys (gh-6119).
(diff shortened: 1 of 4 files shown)


locker (approved):

locker:
Cherry-picked to 2.11.

PR #8464, opened 2023-03-18
httpc: fix signed integer overflow

Totktonada:
The `lua_add_key_u64()` function pushes an `uint64_t` value using `lua_pushinteger()`, which accepts `int64_t` argument. A value >= 2^63 will be interpreted as a negative value on all architectures we're support. However, technically it is implementation defined behavior (see n1256, 6.3.1.3.3).

It is not a problem, in fact, because the function is used only to report `http_client:stat()` statistics and because values beyond 2^63-1 are unreachable in practice.

OTOH, it is easy to eliminate the undefined behavior by replacing `lua_pushinteger()` with our own helper function, which accepts `uint64_t`: `luaL_pushuint64()`.

The values above 10^14 - 1 are now pushed as `cdata<uint64_t>`. Lower values are pushed as `number` just like before the commit.

Reported-in: https://github.com/tarantool/security/issues/103


Buristan (approved):
Hi, Alexander!
Thanks for the patch!

> will be interpreted as a negative value on all architectures we're
support.

Typo: s/we're support/we're supporting/

Totktonada:
> Typo: s/we're support/we're supporting/

Fixed the typo.

PR #8476, opened 2023-03-21
ci: back up artifacts to S3

lastoCHka42:
**ci: add s3-upload-artifact action**

Add a local GitHub action to save test artifacts to S3.
Print a warning if the provided artifact path doesn't exist.

Part of https://github.com/tarantool/multivac/issues/116

---

**ci: save test artifacts to S3**

Add the step to all workflows with test-run tests to save
test artifacts to S3 for providing artifact link from Multivac 
Grafana dashboard.

Resolves https://github.com/tarantool/multivac/issues/116

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -67,6 +67,8 @@ jobs:
           fuzz-seconds: 600
           dry-run: false
           sanitizer: ${{ matrix.sanitizer }}
+      - uses: actions/checkout@v3
+        if: failure()
       - name: Send VK Teams message on failure
         if: failure()
         uses: ./.github/actions/report-job-status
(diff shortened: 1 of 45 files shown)


.github/workflows/fuzzing.yml
```diff
@@ -79,3 +79,11 @@ jobs:
           name: ${{ matrix.sanitizer }}-artifacts
           retention-days: 21
           path: ./out/artifacts
+      - name: Sync artifacts
```
ylobankov, line 82: For now, we don't need to do that for this job.

lastoCHka42: fixed

ylobankov (approved):

PR #8483, opened 2023-03-22
SQL small refactorings

### LINKED ISSUE #8493: SQL small refactorings
Make two small refactorings in SQL source code:
- rename argument `new` to something another due to conflict with C++ reserved keyword
- replace `sql_value` to `struct Mem` everywhere

ligurio:
Closes #8493

### DIFF
--- a/src/box/port.c
+++ b/src/box/port.c
@@ -295,7 +295,7 @@ port_c_get_msgpack(struct port *base, uint32_t *size)
 extern void
 port_c_dump_lua(struct port *port, struct lua_State *L, bool is_flat);
 
-extern struct sql_value *
+extern struct Mem *
 port_c_get_vdbemem(struct port *base, uint32_t *size);
 
 const struct port_vtab port_c_vtab = {
--- a/src/box/port.h
+++ b/src/box/port.h
@@ -89,12 +89,10 @@ static_assert(sizeof(struct port_lua) <= sizeof(struct port),
 void
 port_lua_create(struct port *port, struct lua_State *L);
 
-struct sql_value;
-
 /** Port implementation used with vdbe memory variables. */
 struct port_vdbemem {
     const struct port_vtab *vtab;
-    struct sql_value *mem;
+    struct Mem *mem;
     uint32_t mem_count;
 };
 
@@ -103,7 +101,7 @@ static_assert(sizeof(struct port_vdbemem) <= sizeof(struct port),
 
 /** Initialize a port to dump data in sql vdbe memory. */
 void
-port_vdbemem_create(struct port *base, struct sql_value *mem,
+port_vdbemem_create(struct port *base, struct Mem *mem,
 		    uint32_t mem_count);
 
 struct port_c_entry {
--- a/src/box/sql/mem.c
+++ b/src/box/sql/mem.c
@@ -294,6 +294,15 @@ mem_destroy(struct Mem *mem)
 }
 
 void
+mem_delete(struct Mem *v)
+{
+	if (v == NULL)
+		return;
+	mem_destroy(v);
+	sql_xfree(v);
+}
+
+void
 mem_set_null(struct Mem *mem)
 {
 	mem_clear(mem);
@@ -2956,18 +2965,6 @@ sqlVdbeMemClearAndResize(Mem * pMem, int szNew)
 	return 0;
 }
 
-/*
- * Free an sql_value object
- */
-void
-sqlValueFree(sql_value * v)
-{
-	if (!v)
-		return;
-	mem_destroy(v);
-	sql_xfree(v);
-}
-
 void
 releaseMemArray(Mem * p, int N)
 {
@@ -3476,7 +3473,7 @@ error:
 static const struct port_vtab port_vdbemem_vtab;
 
 void
-port_vdbemem_create(struct port *base, struct sql_value *mem,
+port_vdbemem_create(struct port *base, struct Mem *mem,
 		    uint32_t mem_count)
 {
 	struct port_vdbemem *port = (struct port_vdbemem *) base;
@@ -3485,7 +3482,7 @@ port_vdbemem_create(struct port *base, struct sql_value *mem,
 	port->mem_count = mem_count;
 }
 
-static struct sql_value *
+static struct Mem *
 port_vdbemem_get_vdbemem(struct port *base, uint32_t *mem_count)
 {
 	struct port_vdbemem *port = (struct port_vdbemem *) base;
@@ -3504,7 +3501,7 @@ static const struct port_vtab port_vdbemem_vtab = {
 	.destroy = NULL,
 };
 
-struct sql_value *
+struct Mem *
 port_lua_get_vdbemem(struct port *base, uint32_t *size)
 {
 	struct port_lua *port = (struct port_lua *) base;
@@ -3624,7 +3621,7 @@ port_lua_get_vdbemem(struct port *base, uint32_t *size)
 			goto error;
 		}
 	}
-	return (struct sql_value *)val;
+	return val;
 error:
 	for (int i = 0; i < argc; i++)
 		mem_destroy(&val[i]);
@@ -3632,7 +3629,7 @@ error:
 	return NULL;
 }
 
-struct sql_value *
+struct Mem *
 port_c_get_vdbemem(struct port *base, uint32_t *size)
 {
 	struct port_c *port = (struct port_c *)base;
@@ -3773,7 +3770,7 @@ port_c_get_vdbemem(struct port *base, uint32_t *size)
 		}
 		i++;
 	}
-	return (struct sql_value *) val;
+	return val;
 error:
 	for (int i = 0; i < port->size; i++)
 		mem_destroy(&val[i]);
--- a/src/box/sql/sqlInt.h
+++ b/src/box/sql/sqlInt.h
@@ -3555,7 +3555,7 @@ sql_trigger_delete_step(struct Token *table_name, struct Expr *where);
  * @param parser  Parse context.
  * @param trigger List of triggers on table.
  * @param changes_list Changes list for any UPDATE OF triggers.
- * @param new  1 for new.* ref mask, 0 for old.* ref mask.
+ * @param is_new true for new.* ref mask, false for old.* ref mask.
  * @param tr_tm Mask of TRIGGER_BEFORE|TRIGGER_AFTER.
  * @param space The space to code triggers from.
  * @param orconf Default ON CONFLICT policy for trigger steps.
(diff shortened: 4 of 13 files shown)


src/box/sql/mem.c
```diff
@@ -3624,15 +3624,15 @@ port_lua_get_vdbemem(struct port *base, uint32_t *size)
 			goto error;
 		}
 	}
-	return (struct sql_value *)val;
+	return (struct Mem *)val;
```
ImeevMA, line 3624: `val` is already of type `struct Mem *`, so you don't need a cast here.

src/box/sql/sqlInt.h
```diff
@@ -220,7 +220,6 @@ typedef sql_uint64 sql_uint64;
 
 typedef struct sql_context sql_context;
 typedef struct sql sql;
-typedef struct Mem sql_value;
```
ImeevMA, line 223: Please add to the commit message that you are deleting both `struct sql_value` and `sql_value`.

src/box/sql/vdbe.c
```diff
@@ -1198,7 +1198,7 @@ case OP_FunctionByName: {
 
 	struct region *region = &fiber()->gc;
 	size_t region_svp = region_used(region);
-	port_vdbemem_create(&args, (struct sql_value *)argv, argc);
+	port_vdbemem_create(&args, (struct Mem *)argv, argc);
```
ImeevMA, line 1201: `argv` is already of type `struct Mem *`, so you don't need a cast here.

src/lib/core/CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ set(core_sources
     tweaks.c
 )
 
+# Required for sqlInt.h.
+include_directories(${PROJECT_BINARY_DIR}/src/box/sql)
```
ImeevMA, line 51: Why do you need this?

src/lib/core/port.h
```diff
@@ -33,6 +33,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include "box/sql/sqlInt.h"
```
ImeevMA, line 36: Please drop `#include "box/sql/sqlInt.h"` and add `struct Mem;` just after `struct port;` a little further down.

src/lib/core/sio.c
```diff
@@ -53,7 +53,7 @@ static_assert(SMALL_STATIC_SIZE > NI_MAXHOST + NI_MAXSERV,
 static int
 sio_socketname_to_buffer(int fd, char *buf, int size)
 {
-	int n = 0;
+	int n = 0;(void)n;(void)n;
```
ImeevMA, line 56: Why it is here?

src/lib/core/sio.c
```diff
@@ -257,7 +257,7 @@ sio_read(int fd, void *buf, size_t count)
 		 */
 		if (errno == ECONNRESET) {
 			errno = 0;
-			n = 0;
+			n = 0;(void)n;(void)n;
```
ImeevMA, line 260: This too.

test/fuzz/sql_fuzzer/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+add_executable(sql_fuzzer
```
ImeevMA, line 1: Why are all the changes in this file needed in this patch?

src/box/sql/where.c
```diff
@@ -1233,9 +1233,9 @@ whereRangeSkipScanEst(Parse * pParse,		/* Parsing & code generating context */
 	int rc = 0;
 	enum field_type type = p->key_def->parts[nEq].type;
 
-	sql_value *p1 = 0;	/* Value extracted from pLower */
-	sql_value *p2 = 0;	/* Value extracted from pUpper */
-	sql_value *pVal = 0;	/* Value extracted from record */
+	struct Mem *p1 = 0;	/* Value extracted from pLower */
```
ImeevMA, line 1236: Please use
```
/* Value extracted from pLower. */
struct Mem *p1 = NULL;
```
 instead of
```sql_value *p1 = 0;	/* Value extracted from pLower */```

The same for p2 and pVal.

File: src/box/port.h [hunk 10]
```diff
@@ -89,12 +89,10 @@ static_assert(sizeof(struct port_lua) <= sizeof(struct port),
 void
 port_lua_create(struct port *port, struct lua_State *L);
 
-struct sql_value;
-
 /** Port implementation used with vdbe memory variables. */
 struct port_vdbemem {
     const struct port_vtab *vtab;
-    struct sql_value *mem;
+    struct Mem *mem;
```
ImeevMA, line 95: Why `box/port.h` knows about `struct Mem`?

src/box/sql/mem.c
```diff
@@ -3773,7 +3773,7 @@ port_c_get_vdbemem(struct port *base, uint32_t *size)
 		}
 		i++;
 	}
-	return (struct sql_value *) val;
+	return (struct Mem *) val;
```
ImeevMA, line 3773: `val` is already of type `struct Mem *`, so you don't need a cast here.

src/box/sql/sqlInt.h
```diff
@@ -3564,7 +3564,7 @@ sql_trigger_delete_step(struct Token *table_name, struct Expr *where);
  */
 uint64_t
 sql_trigger_colmask(Parse *parser, struct sql_trigger *trigger,
-		    ExprList *changes_list, int new, int tr_tm,
+		    ExprList *changes_list, int ref_mask, int tr_tm,
```
ImeevMA, line 3566: Wouldn't it be better to replace `int new` with `bool is_new` or `bool is_old`? This way we can be sure that it only has two values, and we can choose an index for `prg->column_mask` depending on this argument value.

ImeevMA (changes requested):
Hi! Thank you for the patches! Please fix checkpatch warnings. Also, I have some comments, see them below.

src/box/sql/sqlInt.h
ligurio: Updated commit message.

src/box/sql/mem.c
```diff
@@ -293,6 +293,15 @@ mem_destroy(struct Mem *mem)
 	mem->z = NULL;
 }
 
+void
+mem_delete(struct Mem * v)
+{
+	if (!v)
```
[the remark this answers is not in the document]
ligurio: Updated:

```diff
--- a/src/box/sql/mem.c
+++ b/src/box/sql/mem.c
@@ -296,7 +296,7 @@ mem_destroy(struct Mem *mem)
 void
 mem_delete(struct Mem *v)
 {
-       if (!v)
+       if (v == NULL)
                return;
        mem_destroy(v);
        sql_xfree(v);
```

src/box/sql/mem.c
ligurio: Updated:

```diff
@@ -3621,7 +3621,7 @@ port_lua_get_vdbemem(struct port *base, uint32_t *size)
                        goto error;
                }
        }
-       return (struct Mem *)val;
+       return val;
 error:
        for (int i = 0; i < argc; i++)
                mem_destroy(&val[i]);

```

src/lib/core/sio.c
ligurio: already removed, this is a changes made by build scripts and were added  accidentally to commit

test/fuzz/sql_fuzzer/CMakeLists.txt
ligurio: already removed

src/lib/core/port.h
ligurio: Added.

src/box/sql/sqlInt.h
ligurio: Yes, `is_new` or `is_old` would be better and clear. According to comment above, 1 is passed for new mask and 0 for old mask, so I chose variable name `is_new`.
I agree, converting argument to boolean would be good too. Did it.
I'm not sure is it worth to convert passed o and 1 in callers to boolean explicitly because this conversion will be done implicitly and no sense to touch code there. Let me know your opinion.

src/lib/core/CMakeLists.txt
ligurio: it was needed for including `stdInt.h`, removed

src/box/sql/where.c
```diff
@@ -1233,9 +1233,12 @@ whereRangeSkipScanEst(Parse * pParse,		/* Parsing & code generating context */
 	int rc = 0;
 	enum field_type type = p->key_def->parts[nEq].type;
 
-	sql_value *p1 = 0;	/* Value extracted from pLower */
-	sql_value *p2 = 0;	/* Value extracted from pUpper */
-	sql_value *pVal = 0;	/* Value extracted from record */
+	/* Value extracted from pLower */
+	struct Mem *p1 = 0;
```
ImeevMA, line 1237: Please use `NULL` instead of `0` to initialize `p1`, `p2` and `pVal`.

ImeevMA (approved):
Thanks for fixes! I have a couple more cmall comments below.

src/box/sql/mem.c
ImeevMA: I don't see this change.

src/box/sql/sqlInt.h
ImeevMA: I think implicit casting `true` to 1 and `false` to 0 is fine for C.

src/box/sql/mem.c
ligurio: Yes, but for `port_c_get_vdbemem` I've removed casting too.

src/box/sql/mem.c
```diff
@@ -3773,7 +3770,7 @@ port_c_get_vdbemem(struct port *base, uint32_t *size)
 		}
 		i++;
 	}
-	return (struct sql_value *) val;
+	return (struct Mem *)val;
```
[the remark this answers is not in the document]
ligurio: Fixed, thanks!

locker (changes requested):

src/box/port.c
```diff
@@ -295,7 +295,7 @@ port_c_get_msgpack(struct port *base, uint32_t *size)
 extern void
 port_c_dump_lua(struct port *port, struct lua_State *L, bool is_flat);
 
-extern struct sql_value *
+extern struct Mem *
```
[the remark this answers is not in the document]
ImeevMA: I would suggest doing it in a different patch-set. `struct Mem` was added at the same time sqlite sources were added: https://github.com/tarantool/tarantool/blob/d0a9dfed5e31213aec5859015d9869e235ba6d44/src/box/sql/vdbeInt.h#L187

Please fill an issue, I will do it later.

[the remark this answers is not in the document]
locker: Using the name `Mem` in the core sources is confusing, because it's unclear what this "Mem" relates to. Besides, this name doesn't meet  the Tarantool naming convention. Judging by the comment to `struct Mem`, I think that this type should in fact be called `sql_value`. I see no reason in renaming `sql_value` to `Mem` in the core sources just to rename it back later. I think it's better to leave the typedef in place for now to be used outside `src/box/sql`.

ligurio:
Closes #8493

kyukhin (approved):

PR #8487, opened 2023-03-23
Fix issues from static analyzers and fuzzer

ImeevMA:
This patch fixes some issues found by static analyzers and the fuzzer. Most issues are false-positives, but some refactoring is needed to avoid warnings in the future. Some memory leaks and possible bugs have also been found and fixed. However, there are no tests, no docs, no changelog, as our tests current tests can't reproduce or catch any of found bugs.

Closes tarantool/security#120

### DIFF
--- a/changelogs/unreleased/ghs-120-use-after-free.md
+++ b/changelogs/unreleased/ghs-120-use-after-free.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* Fixed a Use-After-Free vulnerability in the `WITH RECURSIVE` clause
+  (ghs-119).
--- a/src/box/sql.c
+++ b/src/box/sql.c
@@ -304,8 +304,6 @@ sql_space_info_new_from_space_def(const struct space_def *def)
 {
 	uint32_t field_count = def->field_count + 1;
 	struct sql_space_info *info = sql_space_info_new(field_count, 0);
-	if (info == NULL)
-		return NULL;
 	for (uint32_t i = 0; i < def->field_count; ++i) {
 		info->types[i] = def->fields[i].type;
 		info->coll_ids[i] = def->fields[i].coll_id;
@@ -322,8 +320,6 @@ sql_space_info_new_from_index_def(const struct index_def *def, bool has_rowid)
 	if (has_rowid)
 		++field_count;
 	struct sql_space_info *info = sql_space_info_new(field_count, 0);
-	if (info == NULL)
-		return NULL;
 	for (uint32_t i = 0; i < def->key_def->part_count; ++i) {
 		info->types[i] = def->key_def->parts[i].type;
 		info->coll_ids[i] = def->key_def->parts[i].coll_id;
@@ -1681,10 +1677,7 @@ sql_check_create(const char *name, uint32_t space_id, uint32_t func_id,
 		 uint32_t fieldno, bool is_field_ck)
 {
 	const struct space *space = space_by_id(space_id);
-	if (space == NULL) {
-		diag_set(ClientError, ER_NO_SUCH_SPACE, space_name(space));
-		return -1;
-	}
+	assert(space != NULL);
 	struct tuple_constraint_def *cdefs;
 	uint32_t count;
 	const char *path;
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -2662,8 +2662,6 @@ sqlCodeSubselect(Parse * pParse,	/* Parsing context */
 			int reg_eph = ++pParse->nMem;
 			struct sql_space_info *info =
 				sql_space_info_new(nVal, 0);
-			if (info == NULL)
-				return 0;
 			sqlVdbeAddOp4(v, OP_OpenTEphemeral, reg_eph, 0, 0,
 				      (char *)info, P4_DYNAMIC);
 			sqlVdbeAddOp3(v, OP_IteratorOpen, pExpr->iTable, 0,
@@ -4046,10 +4044,6 @@ sqlExprCodeTarget(Parse * pParse, Expr * pExpr, int target)
 			if (func->def->language == FUNC_LANGUAGE_SQL_BUILTIN) {
 				struct sql_context *ctx =
 					sql_context_new(func, coll);
-				if (ctx == NULL) {
-					pParse->is_aborted = true;
-					return -1;
-				}
 				sqlVdbeAddOp4(v, OP_BuiltinFunction, nFarg,
 					      r1, target, (char *)ctx,
 					      P4_FUNCCTX);
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -3180,13 +3180,10 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
 	const char *sign = is_neg ? "-" : "";
 	if (z[0] == '0' && (z[1] == 'x' || z[1] == 'X')) {
 		errno = 0;
-		if (is_neg) {
+		if (is_neg)
 			value = strtoll(z, NULL, 16);
-		} else {
+		else
 			value = strtoull(z, NULL, 16);
-			if (value > INT64_MAX)
-				goto int_overflow;
-		}
 		if (errno != 0) {
 			diag_set(ClientError, ER_HEX_LITERAL_MAX, sign, z,
 				 strlen(z) - 2, 16);
@@ -3198,7 +3195,6 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
 		bool unused;
 		if (sql_atoi64(z, &value, &unused, len) != 0 ||
 		    (is_neg && (uint64_t) value > (uint64_t) INT64_MAX + 1)) {
-int_overflow:
 			diag_set(ClientError, ER_INT_LITERAL_MAX, sign, z);
 			parse->is_aborted = true;
 			return;
--- a/src/box/sql/select.c
+++ b/src/box/sql/select.c
@@ -128,8 +128,6 @@ sql_space_info_new_from_expr_list(struct Parse *parser, struct ExprList *list,
 	if (has_rowid)
 		++n;
 	struct sql_space_info *info = sql_space_info_new(n, 0);
-	if (info == NULL)
-		return NULL;
 	for (int i = 0; i < list->nExpr; ++i) {
 		bool b;
 		struct coll *coll;
@@ -159,8 +157,6 @@ sql_space_info_new_from_order_by(struct Parse *parser, struct Select *select,
 {
 	int n = order_by->nExpr + 2;
 	struct sql_space_info *info = sql_space_info_new(n, 0);
-	if (info == NULL)
-		return NULL;
 	for (int i = 0; i < order_by->nExpr; ++i) {
 		struct Expr *expr = order_by->a[i].pExpr;
 		enum field_type type = sql_expr_type(expr);
@@ -216,8 +212,6 @@ sql_space_info_new_for_sorting(struct Parse *parser, struct ExprList *order_by,
 
 	struct sql_space_info *info =
 		sql_space_info_new(field_count, part_count);
-	if (info == NULL)
-		return NULL;
 	int k;
 	for (k = 0; k < order_by->nExpr - start; ++k) {
 		bool b;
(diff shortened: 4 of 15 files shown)


src/box/sql/expr.c
```diff
@@ -113,7 +113,7 @@ sql_expr_type(struct Expr *pExpr)
 		assert(pExpr->pRight != NULL && pExpr->pLeft != NULL);
 		enum field_type lhs_type = sql_expr_type(pExpr->pLeft);
 		enum field_type rhs_type = sql_expr_type(pExpr->pRight);
-		return sql_type_result(rhs_type, lhs_type);
+		return sql_type_result(lhs_type, rhs_type);
```
Gerold103, line 116: `sql: fix order of arguments for sql_type_result` - what kind of weird warning is that? If a function's argument order doesn't matter, what is wrong with writing them in any order?

Can it just be ignored? Same with the `sql_xfree(NULL)` commit.

Gerold103 (commented):
Thanks for the patch! Just one comment.

src/box/sql/expr.c
ImeevMA: It is possible, but it will appear every time we use `sql_xfree(NULL)` or `sql_type_result(rhs, lhs)`, which is rather inconvenient. At the moment we have two static analyzers that have identified both problems in each case. For `sql_xfree()` the problem sounds like `Null point dereference`, which is unnerving to unprepared observers. I think it is easier to fix the problems.

Gerold103 (approved):
Thanks for the fixes!

src/box/sql/expr.c
locker, line 116: There's absolutely no point in fixing this weird warning. Please drop this patch.

src/box/sql/expr.c
```diff
@@ -3176,13 +3176,10 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
...
 		errno = 0;
-		if (is_neg) {
+		if (is_neg)
 			value = strtoll(z, NULL, 16);
-		} else {
+		else
 			value = strtoull(z, NULL, 16);
-			if (value > INT64_MAX)
-				goto int_overflow;
-		}
```
locker, line 3186: This `else` block makes no sense now. You should use `strtoll` in both cases otherwise you may get an undetected overflow.

src/box/sql/func.c
```diff
@@ -2374,6 +2374,10 @@ func_sql_expr_call(struct func *func, struct port *args, struct port *ret)
 	struct vdbe_field_ref *ref;
 	size_t size = sizeof(ref->slots[0]) * count + sizeof(*ref);
 	ref = region_aligned_alloc(region, size, alignof(*ref));
+	if (ref == NULL) {
+		diag_set(OutOfMemory, size, "region_aligned_alloc", "ref");
+		goto error;
+	}
```
locker, line 2380: Please use `xregion_aligned_alloc`. I'd prefer if you prepared one patch (in a separate PR) that replaces all region allocations in the SQL code with `x*` versions in one go rather than fixing each place individually.

src/box/field_def.c
```diff
@@ -118,6 +118,7 @@ const char *field_type_strs[] = {
 	/* [FIELD_TYPE_INTERVAL] = */ "interval",
 	/* [FIELD_TYPE_ARRAY]    = */ "array",
 	/* [FIELD_TYPE_MAP]      = */ "map",
+	/* [field_type_MAX]      = */ "unknown",
```
locker, line 121: Please don't do that. `field_type_MAX` should never be converted to a string. You basically introduce dead code here.

src/box/sql/select.c
```diff
@@ -2131,11 +2131,11 @@ sqlColumnsFromExprList(Parse * parse, ExprList * expr_list,
 			}
 			zName = sqlMPrintf("%.*z_%u", nName, zName, ++cnt);
 		}
+		assert(zName != NULL);
```
locker: The `while` loop above checks `zName` of `NULL` in the loop condition, which implies that `zName` can in fact be `NULL` here, so this won't silence static analyzers.

src/box/sql/select.c
```diff
@@ -374,6 +374,7 @@ sqlSelectNew(Parse * pParse,	/* Parsing context */
 		 Expr * pOffset)	/* OFFSET value.  NULL means no offset */
 {
 	Select standin;
+	memset(&standin, 0, sizeof(standin));
```
locker, line 377: If this isn't necessary, we shouldn't do that just to silence static analyzers. Please drop this patch.

src/box/sql/select.c
```diff
@@ -2102,55 +2093,45 @@ sqlColumnsFromExprList(Parse * parse, ExprList * expr_list,
...
-			if (nName > 0) {
-				int j;
-				for (j = nName - 1;
-				     j > 0 && sqlIsdigit(zName[j]); j--);
-				if (zName[j] == '_')
-					nName = j;
-			}
-			zName = sqlMPrintf("%.*z_%u", nName, zName, ++cnt);
-		}
-		assert(zName != NULL);
```
locker: And here you remove the assertion added in commit `sql: drop check of zName in sqlColumnsFromExprList`... Please squash the two patches.

locker (commented):
A perfectly worthless PR...

src/box/sql/expr.c
ImeevMA: I disagree. At the moment, we can work with all INTEGER values supported by Tarantool:
```Lua
tarantool> box.execute([[select 0xFFFFFFFFFFFFFFFF;]])
---
- metadata:
  - name: COLUMN_1
    type: integer
  rows:
  - [18446744073709551615]
...

tarantool> box.execute([[select -0x7FFFFFFFFFFFFFFF;]])
---
- metadata:
  - name: COLUMN_1
    type: integer
  rows:
  - [-9223372036854775807]
...
```

Let's assume that I appy this patch:
```Diff
diff --git a/src/box/sql/expr.c b/src/box/sql/expr.c
index 78a4c4541..d4f757878 100644
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -3179,10 +3179,7 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
        const char *sign = is_neg ? "-" : "";
        if (z[0] == '0' && (z[1] == 'x' || z[1] == 'X')) {
                errno = 0;
-               if (is_neg)
-                       value = strtoll(z, NULL, 16);
-               else
-                       value = strtoull(z, NULL, 16);
+               value = strtoll(z, NULL, 16);
                if (errno != 0) {
                        diag_set(ClientError, ER_HEX_LITERAL_MAX, sign, z,
                                 strlen(z) - 2, 16);
```

Then we will not be able to work with INTEGER values greater than INT64_MAX:
```Lua
tarantool> box.execute([[select 0xFFFFFFFFFFFFFFFF;]])
---
- null
- Hex literal 0xFFFFFFFFFFFFFFFF length 16 exceeds the supported limit (16)
...
```

src/box/sql/select.c
```diff
@@ -2608,9 +2610,6 @@ generateWithRecursiveQuery(Parse * pParse,	/* Parsing context */
 		VdbeComment((v, "Distinct table"));
 	}
 
-	/* Detach the ORDER BY clause from the compound SELECT */
-	p->pOrderBy = 0;
-
```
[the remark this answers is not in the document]
ImeevMA: On line 2557, `pOrderBy` becomes `p->pOrderBy`. If we jump to `end_of_recursive_query` from line 2598 (in the old code), then we remove `p->pOrderBy` (which is still `pOrderBy`) and then set `p->pOrderBy` to `pOrderBy`, which has already been released. So use-after-free is technically possible. However, this will not happen because this function is used in the queries similar to following:
```
box.execute([[WITH RECURSIVE cnt(x, y) AS (VALUES(1, 'asd' collate "binary") UNION ALL SELECT x+1, '1' collate "asd" FROM cnt WHERE x<2 ORDER BY 1) SELECT x, y FROM cnt;]])
```

The first COLLATE clause will be checked before the function is called, and the second (which may result in a possible use-after-free) will be ignored so no reproduce is possible.

src/box/sql/func.c
ImeevMA: Do we actually have `xregion_aligned_alloc()` Should I add this function?

src/box/sql/select.c
ImeevMA: Dropped.

src/box/sql/select.c
ImeevMA: Done.

src/box/sql/expr.c
ImeevMA: Reworked the patch. Arguments now have more generic names.

src/box/field_def.c
ImeevMA: Understood. Dropped this patch.

src/box/sql/select.c
ImeevMA: Squashed this patch.

src/box/sql.c
```diff
@@ -322,8 +320,6 @@ sql_space_info_new_from_index_def(const struct index_def *def, bool has_rowid)
 	if (has_rowid)
 		++field_count;
 	struct sql_space_info *info = sql_space_info_new(field_count, 0);
-	if (info == NULL)
-		return NULL;
```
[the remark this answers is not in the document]
ImeevMA: Thanks for noticing. I have added two new patches. I understand that I should have done this a long time ago, I can only say that I missed some functions.

ImeevMA:
@locker I do not think that this PR is completely worthless. At least, I will have fewer cases when I or someone else will have to spend time marking another static analyzer error as a false positive.

I also want to apologize for not noticing until tonight that you had already reviewed this PR. Until then, I thought you were postponing this review. Sorry.

src/box/sql/select.c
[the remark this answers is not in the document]
ImeevMA: @locker sorry, I was wrong. Here is a reproducer:
```
box.execute([[WITH RECURSIVE cnt(x, y) AS (VALUES(1, 'asd' collate "binary") UNION ALL SELECT x+1, '1' FROM cnt WHERE x<2 ORDER BY 1 collate "asd") SELECT x, y FROM cnt;]])
```

src/box/sql/expr.c
locker: This particular SVACE warning makes no sense (as many others, actually). It just checks that function arguments aren't swapped by mistake by looking at the argument names. Say, we have function `foo(int x, int y)`, then if you call it as `foo(y, x)` somewhere in the code, it'll trigger this warning while `foo(x, y)` will pass the check as well as `foo(a, b)`.

It's pointless to suppress this warning by swapping or renaming arguments, because there's nothing that prevents us from writing the code like that in future: tests will pass either way; `checkpatch` won't complain; a reviewer usually won't insist on a particular argument order. We have marked literally all `WRONG_ARGUMENTS_ORDER` errors as FP (and we have more than 30 of them). There's absolutely no point in going back there and fixing them. It'd be a waste of time.

ImeevMA: I dropped the patch as both reviewers think it's unnecessary.

src/box/sql/select.c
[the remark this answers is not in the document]
ImeevMA: Added.

ImeevMA:
@locker I decided to combine some patches anyway, so now there are only 3 questionable patches in this PR instead of 10+.

changelogs/unreleased/ghs-120-use-after-free.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/sql
+
+* Fixed use-after-free in WITH RECURSIVE (ghs-119).
```
andreyaksenov, line 3: My suggestion:

```
* Fixed a Use-After-Free vulnerability in the ``WITH RECURSIVE`` clause (ghs-119).
```

andreyaksenov (approved):

changelogs/unreleased/ghs-120-use-after-free.md
locker: Applied, thanks.

locker (approved):

PR #8491, opened 2023-03-23
test/fuzz: fix datetime_strptime fuzzing test

### LINKED ISSUE #8490: Fix datetime_strptime_fuzzer
**Bug description**

test datetime_strptime is "bloked" according to fuzzing introspector [^1]:

image

[^1]: https://storage.googleapis.com/oss-fuzz-introspector/tarantool/inspector-report/20230320/fuzz_report.html#High-level-conclusions

ligurio:
Function `datetime_strptime` decodes string with datetime according to specified format, it accepts a datetime struct, buffer with datetime and string with format in arguments. Fuzzing test used static string "iso8601" as a format and it blocked fuzzing test to cover functions used by datetime_strptime under the hood. Fuzz introspector shows that code coveraged by a test is quite low.

Patch updates the test to make it more effective: buffer with datetime and format string are generated using FDP (Fuzzing Data Provider).

Test file extension was changed to .cc, it is needed for building by C++ compiler.

Functions `datetime_strptime` and `tnt_strptime` uses asserts, these asserts breaks fuzzing and I replaces these asserts to if..then.

1. https://storage.googleapis.com/oss-fuzz-introspector/tarantool/

Fixes #8490

NO_CHANGELOG=fuzzing test
NO_DOC=fuzzing test
NO_TEST=fuzzing test


src/lib/core/datetime.c
```diff
@@ -163,9 +163,12 @@ tm_to_datetime(struct tnt_tm *tm, struct datetime *date)
 size_t
 datetime_strptime(struct datetime *date, const char *buf, const char *fmt)
 {
-	assert(date != NULL);
```
tsafin, line 166: Agreed with @ligurio that we still prefer to have some assertions enabled here in (average) Debug mode, but to have those assertions disabled if fuzzing mode is activated. At the moment it looks like we want to introduce some fuzzing-friendly `*_ASSERT` macro.

test/fuzz/datetime_strptime_fuzzer.cc
```diff
@@ -0,0 +1,31 @@
+extern "C"
```
tsafin, line 1: Usually this should not be necessary, and it's up-to header itself to guard his definitions with something like:
```
#ifdef __cplusplus
extern "C" {
#endif
```

I wonder, which of those headers was unprotected and led to compile errors?

src/lib/core/datetime.c
tsafin: OTOH, we do need to protect `datetime_strptime` against `NULL`s coming from outer world, and, even, with some sort of assertions left here, we still do need to have checks here for the `Release`/`RelWithDebInfo` compile modes.

src/lib/core/datetime.c
```diff
@@ -163,9 +163,12 @@ tm_to_datetime(struct tnt_tm *tm, struct datetime *date)
...
-	assert(fmt != NULL);
-	assert(buf != NULL);
+	if ((date == NULL) ||
+	    (fmt == NULL) ||
+	    (buf == NULL)) {
+		return -1;
```
tsafin, line 169: And we return 0, not -1 (as it's length of parsed prefix) if there are some premature errors.

test/fuzz/datetime_strptime_fuzzer.cc
ligurio: already removed

src/lib/core/datetime.c
ligurio: replaced with modified version of assert

src/lib/core/datetime.c
ligurio: Updated, these asserts were reverted.

ligurio:
Merge blocked by https://github.com/tarantool/tarantool/issues/8502

tsafin (approved):
Even more LGTM

PR #8503, opened 2023-03-27
Rework struct Mem

ImeevMA:
This patch-set reworks `struct Mem` and renames it to `struct sql_mem`.

Closes #6051


ImeevMA:
This PR is too old already. It seems it would be easier to rewrite the PR than to rebase it.

PR #8514, opened 2023-03-28
workflow: add stale PR last update setter

fckxorg:
This patch adds a Python script, that uses GitHub's GraphQL API to provide automatic last update date for PR items on any specified GitHub ProjectsV2 board.

Since there is no versatile enough  way to automate a GitHub ProjectV2 by its own means, that is done via standard mechanism of on schedule workflow, which runs the script every hour.

NO_DOC=workflow
NO_TEST=workflow
NO_CHANGELOG=workflow

### DIFF
--- a/.github/workflows/lango-stale-reviews.yml
+++ b/.github/workflows/lango-stale-reviews.yml
@@ -1,0 +1,23 @@
+name: lango-stale-reviews
+on:
+  schedule:
+    - cron:  '* */1 * * *'    # At every hour.
+
+jobs:
+  update-last-activity-dates:
+    env:
+      LANGO_REVIEW_BOARD_TOKEN: ${{ github.token }}
+      ORGANIZATION: Tarantool
+      PROJECT_ID: 83
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python3
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+          cache: pip
+      - name: Install dependencies
+        run: pip install requests
+      - name: Update dates
+        run: python tools/stale_pr_date_setter.py
--- a/tools/stale_pr_date_setter.py
+++ b/tools/stale_pr_date_setter.py
@@ -1,0 +1,273 @@
+#!/usr/bin/env python3
+"""
+Tarantool Lango team PR last activity date setter.
+
+This script makes use of GitHub GraphQL API to check the activity
+on pull requests listed on the specified board and update the
+last activity date correspondingly.
+
+Parameters are passed via environment variables.
+Available parameters:
+    - LANGO_REVIEW_BOARD_TOKEN: GitHub token with read/write rights
+    for the specified project and read rights for PRs in that project.
+    - ORGANIZATION: Name of the organization that project belongs to.
+    - PROJECT_ID: ID of the GitHub project.
+
+NB: This script is suitable only for GitHub ProjectV2.
+"""
+import json
+import logging
+import os
+import sys
+from datetime import datetime, timezone
+from enum import Enum
+
+import requests
+
+
+class Exitcode(Enum):
+    ERR_PROJECT = 1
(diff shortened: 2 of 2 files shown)


.github/workflows/lango-stale-reviews.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python3
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests
```
Totktonada, line 22: How long this step? I guess we can use caching here: https://github.com/actions/setup-python#caching-packages-dependencies (and I would skip updating pip itself, because I see no reason).

.github/scripts/stale_pr_date_setter.py
```diff
@@ -0,0 +1,272 @@
+"""
```
Totktonada, line 1: Feel free to use the `tools` directory. We collect here some scripts from time to time.

.github/workflows/lango-stale-reviews.yml
Totktonada: Nice, thanks!

tools/stale_pr_date_setter.py
```diff
@@ -0,0 +1,272 @@
...
+                    fieldValueByName(name: "%s")
+                    {
+                        ... on ProjectV2ItemFieldDateValue
+                        {
+                            date
+                        }
+                    }
+                    content
+                    {
+                        ...on PullRequest
```
Totktonada, line 111: Checkpatch complains:

```
ERROR:TYPO_SPELLING: 'PullRequest' may be misspelled - perhaps 'Pull request'?
#166: FILE: tools/stale_pr_date_setter.py:111:
+                        ...on PullRequest
                              ^^^^^^^^^^^
```

Don't know what is better to do here, just noted.

tools/stale_pr_date_setter.py
```diff
@@ -0,0 +1,272 @@
+"""
```
Totktonada, line 1: Nit: I would add a shebang.

Totktonada (approved):
I didn't looked for correctness, but rather for readability and ability for other to make modifications if necessary. The script and the automation looks okay for me from this point of view.

tools/stale_pr_date_setter.py
Totktonada: I think it can be ignored.

ylobankov (approved):
LGTM. But let's fix https://github.com/tarantool/checkpatch/issues/56 first. Otherwise, all PRs will fail on the `checkpatch` job.

.github/workflows/lango-stale-reviews.yml
```diff
@@ -0,0 +1,25 @@
...
+      ORGANIZATION: Tarantool
+      PROJECT_ID: 83
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python3
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+          cache: 'pip'
```
ylobankov, line 19: Quotes can be omitted.

Totktonada: BTW, quotes around `3.9` are desirable, because it is easy to forget about them for, say, `3.10` and get `3.1`.

tools/stale_pr_date_setter.py
locker: Fixed.

locker:
> LGTM. But let's fix [tarantool/checkpatch#56](https://github.com/tarantool/checkpatch/issues/56) first. Otherwise, all PRs will fail on the `checkpatch` job.

I fixed the issue.

Anyway, I think it's okay to push PRs with failing checkpatch because the checkpatch workflow doesn't run on commits to branches.

PR #8518, opened 2023-03-30
replication: recovery mixed transacrtions

### LINKED ISSUE #7932: Recovery: force recovery should try to recover mixed transactions from WAL
Transaction boundaries in WAL were introduced in 2.1.2 (872d6f1), but replication applier started to apply transaction atomically only in 2.2.1(e94ba2e), so within this range it is possible that transactions are mixed in WAL. There was no problem with them until 2.8.1 because recovering process was still working row-by-row, but in 2.8.1 (9311113) it became transactional as well so starting from 2.8.1 tarantool fails to handle mixed transactions from WAL.

Recently this problem arose [here](https://jira.vk.team/browse/TNT-656)

I find it difficult to provide 100% reproducer here, but at least you need:

1. Setup full mesh cluster of master replicas (the more the better) of any version from the range [2.1.2, 2.2.0]
2. Write various multi-component transactions simultaneously on different replicas
3. Upgrade one instance to 2.8.1 or later (w/o snapshots). If mixed transactions have been produced in WAL at step 2, upgraded instance will fail to recover from WAL 

Mixed transactions is obviously the one of unexpected scenarios in modern tarantool, so we should not handle it in normal mode, and we take care of it only in force recovery mode. But current 'force recovery' treats all the rows of the unexpected transaction as unexpected and thus just skips the entire transaction row-by-row. However it seems we have enough information in this case to handle it more data-friendly, so I'd suggest to improve force recovery to avoid data loss, or at least try to avoid

yanshtunder:
See the docbot request for details.

Closes #7932

@TarantoolBot document
Title: correct recovery of mixed transactions

Transaction boundaries in WAL were introduced in 2.1.2 (`872d6f1`), but replication applier started to apply transaction atomically only in 2.2.1(`e94ba2e`), so within this range it is possible that transactions are mixed in WAL. There was no problem with them until 2.8.1 because recovering process was still working row-by-row, but in 2.8.1 (`9311113`) it became transactional as well so starting from 2.8.1 tarantool fails to handle mixed transactions from WAL.

Mixed transactions is obviously the one of unexpected scenarios in modern tarantool, so we should not handle it in normal mode, and we take care of it only in force recovery mode. But current 'force recovery' treats all the rows of the unexpected transaction as unexpected and thus just skips the entire transaction row-by-row. However we have enough information in this case to handle it more data-friendly.


src/box/box.cc
```diff
@@ -793,6 +798,80 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+			alloc_row = item;
+			free(alloc_row);
+			alloc_row = NULL;
+		}
+		/* clear the cell of array */
+		assert(rlist_empty(&nodes_rows[id]));
+		assert(nodes_rows[id].prev != NULL);
+		nodes_rows[id].prev = NULL;
+		assert(nodes_rows[id].next != NULL);
+		nodes_rows[id].next = NULL;
```
sergepetrenko, line 869: There's no need to clear the list extensively. Having an empty rlist (which has `rlist_empty(...)` == true) is totally fine. It's ready to accept new members.

src/box/box.cc
```diff
@@ -793,6 +798,80 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		tnt_raise(OutOfMemory, sizeof(*alloc_row), "malloc",
+			  "alloc_row");
+	memcpy(alloc_row, row, sizeof(*row));
+
+	char *alloc_body = (char *)malloc(row->body[0].iov_len);
+	if (alloc_body == NULL)
+		tnt_raise(OutOfMemory, sizeof(*alloc_body), "malloc",
+			  "alloc_body");
+	memcpy(alloc_body, row->body[0].iov_base, row->body[0].iov_len);
+	alloc_row->body[0].iov_base = alloc_body;
```
sergepetrenko, line 842: This seems to work just fine, so now that we've tested the hypothesis please start allocating the rows on a mempool.

Speaking of row bodies, you may use `lsregion`. For example, check applier, it had to solve exactly the same problem we have here.
Here's an example of alloc:
https://github.com/tarantool/tarantool/blob/9fcafbd14b962d9f844a9d114518fa690753f276/src/box/applier.cc#L1988-L1996

And here's an example of free:
https://github.com/tarantool/tarantool/blob/9fcafbd14b962d9f844a9d114518fa690753f276/src/box/applier.cc#L1883

The only difference is that you will need to track the oldest lsregion_id you still need to keep and free everything up to that id, not up to the id of the recently completed row.

src/box/box.cc
```diff
@@ -793,6 +798,80 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+			if (rc == 0) {
+				rc = wal_stream_apply_dml_row(stream, item);
+			}
+			/* deallocating the memory */
+			alloc_body = (char *)item->body[0].iov_base;
+			free(alloc_body);
+			alloc_body = NULL;
+			alloc_row = item;
+			free(alloc_row);
+			alloc_row = NULL;
```
sergepetrenko, line 862: It seems you don't have to set `alloc_row` and `alloc_body` back to `NULL`.

src/box/box.cc
```diff
@@ -793,6 +798,80 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+ * to store the rows for correct further recovery. For example, with
+ * mixed transactions. The following function saves the newly arrived
+ * row to the rlist, which is stored in array is nodes_rows in cell is
+ * replica_id-1. As soon as a row arrives with the is_commit flag set,
+ * the corresponding transaction tries to apply. If an error occurs the
+ * transaction is rolled back. And we continue to work with the next
+ * transaction.
+ */
+static int
+wal_stream_apply_mixed_dml_row(struct wal_stream *stream,
```
sergepetrenko, line 814: You should also check that once recovery is finished there are no unfinished transactions left in your `nodes_rows` array, like simple recovery does it:
https://github.com/tarantool/tarantool/blob/9fcafbd14b962d9f844a9d114518fa690753f276/src/box/box.cc#L4526-L4533

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
```diff
@@ -0,0 +1,5 @@
+## feature/replication
+
+* This patch implements correct recovery of mixed transactions. To do this,
+  use `force_recovery` = true. If you wish to revert to the old behavior,
+  you must not set the `force_recovery` (gh-7932).
```
sergepetrenko, line 5: Please also check what happens when one node feeds the other such a broken xlog.
It seems a similar mechanic should be added to applier.

sergepetrenko (changes requested):
Hi! Thanks for the patch, not bad!
Please see my comments below.

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
andreyaksenov, line 5: My suggestion regarding wording:
```
* Implemented correct recovery of mixed transactions. To do this,
  set `box.cfg.force_recovery` to `true`. If you need to revert to the old behavior,
  don't set the `force_recovery` option (gh-7932).
```

andreyaksenov (approved):

src/box/box.cc
```diff
@@ -793,6 +798,80 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
 	return -1;
 }
 
+struct rlist nodes_rows[VCLOCK_MAX];
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/xrow.h
```diff
@@ -116,6 +116,8 @@ struct xrow_header {
 	/* See `IPROTO_SCHEMA_VERSION`. */
 	uint64_t schema_version;
 	struct iovec body[XROW_BODY_IOVMAX];
+	/* List of rows to stacked in the save_rows array. */
+	struct rlist in_row_list;
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
yanshtunder: Done. Added test.

src/box/box.cc
yanshtunder: I create a list if `nodes_rows[id].prev == NULL` and if `nodes_rows[id].next == NULL`. Therefore, checking that the list is empty is not enough. I don’t see the point of using  `RLIST_HEAD(nodes_rows[id]) ` macro, because it will complicate the logic for an array.

src/box/box.cc
yanshtunder: Done

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
yanshtunder: Done

src/box/box.cc
yanshtunder: I'm not sure that I work correctly with `lsr_id`.

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
yanshtunder: To replication do the following. Recover the broken xlog node by setting `force_recovery` to `true`. Then do replica rebootstrap.

xuniq (approved):

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
sergepetrenko: Please, mention that in the changelog and docbot request with a proper explanation.
That one is too short and lacks context.
If I were the user, I'd have questions. To replication from which node to which node? Which replica should be rebootstrapped? The just recovered one or some other?

src/box/box.cc
```diff
@@ -793,6 +799,172 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		struct wal_row *item, *next_row;
+		rlist_foreach_entry_safe(item, &nodes_rows[id],
+					 in_row_list, next_row) {
+			rlist_del_entry(item, in_row_list);
+			if (rc == 0) {
+				rc = wal_stream_apply_dml_row(stream,
+				     (struct xrow_header *)item);
+			}
+			/* deallocating the memory */
+			lsregion_gc(&wal_row_lsr, item->lsr_id);
```
sergepetrenko, line 916: This is not right.
Imagine you have an xlog
1.`replica 1 lsn 1 begin`
2.`replica2 lsn 1 begin`
3.`replica1 lsn2 commit`
4.`replica2 lsn2 commit`.
The first row will receive lsr_id 1, the second one lsr_id 2 and so on.

By the time you read row 3, which is commit one, you call `lsregion_gc` on each row of transaction. It will be `lsregion_gc(..., lsr_id = 1)` and `lsregion_gc(..., lsr_id = 3)`. So you'll free row 2 by accident. And it is still needed.

You need to track the smallest lsr_id still needed, and call lsregion_gc() on that lsr_id, not on the lsr_id of the applied row.

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
```diff
@@ -0,0 +1,163 @@
...
+
+g.test_good_xlog_with_force_recovery = function(cg)
+    cg.server:exec(function()
+        box.schema.space.create("test", {id = 512})
+        box.space.test:create_index("primary")
+    end)
+
+    t.assert_equals(cg.server:exec(function()
+        return box.space.test.id
+    end), 512)
```
sergepetrenko, line 30: You do not need to create the space in the tests, because all your test xlogs already contain space creation.

sergepetrenko (changes requested):
Thanks for the fixes!

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
yanshtunder: Done

src/box/box.cc
```diff
@@ -793,6 +799,172 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+ * transactions.
+ */
+static void
+wal_stream_save_body(struct wal_row *row)
+{
+	static int64_t id = 0;
+	assert(row->row.bodycnt <= 1);
+	if (row->row.bodycnt == 1) {
+		size_t len = row->row.body[0].iov_len;
+		char *new_base = (char *)lsregion_alloc(&wal_row_lsr, len,
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -153,6 +154,11 @@ static bool is_storage_initialized = false;
 static bool is_ro = true;
 static fiber_cond ro_cond;
 
+/*
+ * The flag is responsible for the force_recovery mode.
+ */
+static bool is_force_recovery = false;
```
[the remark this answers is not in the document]
yanshtunder: Ok

src/box/box.cc
```diff
@@ -793,6 +799,172 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+}
+
+/*
+ * Callback to stash row upon receipt, used to recover mixed transactions.
+ */
+static struct wal_row *
+wal_stream_save_row(struct xrow_header *row)
+{
+	struct wal_row *new_row =
+	(struct wal_row *)mempool_alloc(&wal_row_pool);
```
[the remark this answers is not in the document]
yanshtunder: Changed to `xlsregion_alloc`

src/box/box.cc
```diff
@@ -793,6 +799,172 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	memcpy(new_row, (struct wal_row *)row, sizeof(struct wal_row));
+	return new_row;
+}
+
+/*
+ * Callback to stash row bodies upon receipt, used to recover mixed
+ * transactions.
+ */
+static void
+wal_stream_save_body(struct wal_row *row)
```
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/box.cc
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/box.cc
```diff
@@ -793,6 +799,172 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+			}
+			/* deallocating the memory */
+			lsregion_gc(&wal_row_lsr, item->lsr_id);
+			mempool_free(&wal_row_pool, item);
+		}
+		/* clear the cell of array */
+		assert(rlist_empty(&nodes_rows[id]));
+		assert(nodes_rows[id].prev && nodes_rows[id].next);
+		nodes_rows[id].prev = NULL;
+		nodes_rows[id].next = NULL;
```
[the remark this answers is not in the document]
yanshtunder: Wanted to return `nodes_rows[id]` to the original state.  In the original state `nodes_rows[id].prev` and `nodes_rows[id].next` had NULL. I made the changes as you requested.

sergepetrenko (changes requested):
Thanks for the fixes!

We're on the right track here but there's still some things to improve.

src/box/box.cc
sergepetrenko: Good job!

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: Thanks!
In `applier.cc` `alloc_row` and `save_body` are separated because `alloc_row` creates a placeholder for a row to be read into by `coio_read_xrow`. And only after doing that you can actually save the body.

In our case now there is no need to separate the functions because you already receive a ready row and need to copy it completely at once.

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: Nice!

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
```diff
@@ -0,0 +1,14 @@
...
+* Implemented correct recovery of mixed transactions. To do this, set
+  `box.cfg.force_recovery` to `true`. If you need to revert to the old
+  behavior, don't set the `force_recovery` option.
+
+* Note:
+  Let there be two nodes (`node#1` and `node#2`). And let the data be
+  replicated from `node#1` to `node#2`. Suppose that at some point in time,
+  `node#1` is restoring data from an xlog containing mixed transactions. To
+   replicate data from `node#1` to `node#2`, do the following:
+    1. Disable `node#2`
```
sergepetrenko, line 12: What do you mean by "disable"?
If I remember correctly, we need to rebootstrap the node. Stop it, remove all the data files from it, and make it rejoin to master.

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+ * Callback to stash row and row bodies upon receipt, used to recover
+ * mixed transactions.
+ */
+static struct wal_row *
+wal_stream_save_row(struct wal_stream *stream, struct xrow_header *row)
+{
+	static int64_t lsr_id = 0;
+	struct lsregion *lsr = &stream->lsr;
+	struct wal_row *new_row = (struct wal_row *)xlsregion_alloc(lsr,
+				   sizeof(struct wal_row), ++lsr_id);
```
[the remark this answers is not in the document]
yanshtunder: ??

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
 	return -1;
 }
 
+/* Storage, adding rows to a rlist to separate mixed transactions. */
+struct wal_row {
+	/* Base class. */
+	struct xrow_header row;
+	/* List of rows to stacked in the nodes_rows array. */
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+			rlist_foreach_entry_safe(item, &nodes_rows[i],
+						 in_row_list, next_row) {
+				rlist_del_entry(item, in_row_list);
+			}
+			assert(rlist_empty(&nodes_rows[i]));
+		}
+	}
+	int64_t lsr_id = stream->base.row_count;
+	/* deallocating the memory */
+	deallocate_wal_row(stream, lsr_id);
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -843,6 +995,21 @@ wal_stream_create(struct wal_stream *ctx)
...
+	ctx->tsn = 0;
+	ctx->first_row_lsn = 0;
+	ctx->has_yield = false;
+	ctx->has_global_row = false;
+}
+
+static void
+wal_stream_destroy(struct wal_stream *ctx)
+{
+	clear_unfinished_txn(ctx);
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+ */
+static int
+wal_stream_apply_mixed_dml_row(struct wal_stream *stream,
+			       struct xrow_header *row)
+{
+	int rc = 0;
+	uint32_t id = row->replica_id;
+	struct rlist *nodes_rows = stream->nodes_rows;
+
+	/* When a new row arrives, it has the same address. Because the
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		return 0;
+	} else {
+		int64_t lsr_id = find_min_lsr_id(stream, save_row);
+		struct wal_row *item, *next_row;
+		rlist_foreach_entry_safe(item, &nodes_rows[id],
+					 in_row_list, next_row) {
+			rlist_del_entry(item, in_row_list);
+			if (rc == 0) {
+				rc = wal_stream_apply_dml_row
+				(stream, (struct xrow_header *)item);
```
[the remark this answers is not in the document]
yanshtunder: ?

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+
+	/* When a new row arrives, it has the same address. Because the
+	 * row is on the stack. Therefore, it needs to allocate memory on
+	 * the heap. We do with row->body[0].iov_base analogically.
+	 */
+	struct wal_row *save_row = wal_stream_save_row(stream, row);
+	rlist_add_tail_entry(&nodes_rows[id], save_row, in_row_list);
+
+	if (!row->is_commit) {
+		return 0;
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -4661,6 +4828,7 @@ local_recovery(const struct tt_uuid *instance_uuid,
 		diag_set(XlogError, "found a not finished transaction "
 			 "in the log");
 		wal_stream_abort(&wal_stream);
+		wal_stream_destroy(&wal_stream);
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: The coding style is wrong in this hunk. You can't put the opening bracket after a function on a separate line. Besides, the arguments should go below the opening bracket. Like I showed in other example.

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: Your argument alignment is wrong, please fix it like in my example.
Checkpatch doesn't catch this because it sees an opening bracket of `(struct wal_row *)` above, I suppose.

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		new_row->row.body[0].iov_base = new_base;
+	}
+	return new_row;
+}
+
+/*
+ * Deallocating memory for the wal_row
+ */
+static void
+deallocate_wal_row(struct wal_stream *stream, int64_t lsr_id)
```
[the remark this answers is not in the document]
yanshtunder: - [x]  `deletelocate_wal_row` renamed to `wal_stream_gc()`
- [ ]  remove `lsr_id` from arguments

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+{
+	struct lsregion *lsr = &stream->lsr;
+	lsregion_gc(lsr, lsr_id);
+}
+
+/*
+ * Find the min lsr_id that is still needed.
+ */
+static int64_t
+find_min_lsr_id(struct wal_stream *stream, struct wal_row *row)
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+{
+	int64_t min_lsr_id = row->lsr_id;
+	struct rlist *nodes_rows = stream->nodes_rows;
+	struct wal_row *item;
+	for (uint32_t i = 0; i < VCLOCK_MAX && i != row->row.replica_id;
+	     i++) {
+		if (!rlist_empty(&nodes_rows[i])) {
+			item = rlist_first_entry(&nodes_rows[i], wal_row,
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -790,6 +804,141 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		}
+	}
+	int64_t lsr_id = stream->base.row_count;
+	/* deallocating the memory */
+	deallocate_wal_row(stream, lsr_id);
+}
+
 /**
  * Yield once in a while, but not too often, mostly to allow signal handling to
  * take place.
```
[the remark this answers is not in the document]
yanshtunder: There is a situation where `has = false` and `unfinished = true`. This does not reflect the old behavior of the `wal_stream_has_tx()` and `wal_stream_try_yield()` functions.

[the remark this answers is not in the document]
yanshtunder: It was decided to rename the function: `wal_stream_has_tx() -> wal_stream_has_tx_in_progress()`. And to add a new function: `wal_stream_has_unfinished_tx()`.

src/box/box.cc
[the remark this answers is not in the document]
yanshtunder: Arguments are placed below the opening bracket.

```C
rlist_foreach_entry_safe(item, &nodes_rows[id],
                         in_row_list, next_row) {
```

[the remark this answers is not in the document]
yanshtunder: > The coding style is wrong in this hunk. You can't put the opening bracket after a function on a separate line.

`../checkpatch/checkpatch.pl --color=always --git HEAD`
total: 0 errors, 409 lines checked

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
yanshtunder: Done

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: > ../checkpatch/checkpatch.pl --color=always --git HEAD
total: 0 errors, 409 lines checked

Yep, checkpatch doesn't catch everything. It's good most of the times, but not always.

src/box/box.cc
```diff
@@ -4839,6 +5005,7 @@ local_recovery(const struct vclock *checkpoint_vclock)
 	wal_stream_create(&wal_stream);
 	auto stream_guard = make_scoped_guard([&]{
 		wal_stream_abort(&wal_stream);
+		wal_stream_destroy(&wal_stream);
```
sergepetrenko, line 5008: another `wal_stream_destroy` is needed where the guard is disabled (`stream_guard.is_active = false` line).

sergepetrenko (changes requested):
Thanks for the fixes, good job!
Only a couple of comments left.

src/box/box.cc
```diff
@@ -797,14 +827,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	/*
+	 * When a new row arrives, it has the same address. Because the
+	 * row is on the stack. Therefore, it needs to allocate memory on
+	 * the heap. We do with row->body[0].iov_base analogically.
+	 */
+	struct wal_row *save_row = wal_stream_save_row(stream, row);
+	rlist_add_tail_entry(&nodes_rows[id], save_row, in_row_list);
+
+	if (!row->is_commit) {
+		return rc;
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/box.cc
```diff
@@ -797,14 +827,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	 * iterate through the nodes_rows array of lists and see which of
+	 * the first `row` for each `replica_id` has the lowest lsr_id.
+	 * This lsr_id is still needed, but anything less than this is not.
+	 * So can free everything up to this lsr_id.
+	 */
+	for (uint32_t i = 0; i < VCLOCK_MAX; i++) {
+		if (!rlist_empty(&nodes_rows[i])) {
+			item = rlist_first_entry(&nodes_rows[i], wal_row,
+						 in_row_list);
+			if (item->lsr_id < min_lsr_id)
```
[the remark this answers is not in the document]
yanshtunder: Thanks for correction. Dumb mistake.

src/box/box.cc
```diff
@@ -797,14 +827,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	return new_row;
+}
+
+/*
+ * Find the min lsr_id that is still needed.
+ */
+static int64_t
+find_min_lsr_id(struct wal_stream *stream)
+{
+	int64_t min_lsr_id = stream->base.row_count;
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/box.cc
```diff
@@ -641,18 +646,43 @@ wal_stream_abort(struct wal_stream *stream)
...
-wal_stream_has_tx(const struct wal_stream *stream)
+wal_stream_has_tx_in_progress(const struct wal_stream *stream)
 {
 	bool has = stream->tsn != 0;
 	assert(has == (in_txn() != NULL));
 	return has;
 }
 
+/**
+ * The wrapper exists only for the debug purposes, to ensure tsn being non-0 is
```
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/box.cc
yanshtunder: Done.

src/box/box.cc
```diff
@@ -641,18 +646,39 @@ wal_stream_abort(struct wal_stream *stream)
...
+	bool unfinished = false;
+	const struct rlist *nodes_rows = stream->nodes_rows;
+	for (int i = 0; i < VCLOCK_MAX; i++) {
+		if (!rlist_empty((struct rlist *)&nodes_rows[i])) {
+			unfinished = true;
+			break;
+		}
+	}
+
+	return has || unfinished;
```
Gerold103, line 674: When you see `has` as `true`, you don't need to walk the rlists:
```Diff
@@ -660,18 +660,14 @@ wal_stream_has_tx_in_progress(const struct wal_stream *stream)
 static bool
 wal_stream_has_unfinished_tx(const struct wal_stream *stream)
 {
-	bool has = wal_stream_has_tx_in_progress(stream);
-
-	bool unfinished = false;
+	if (wal_stream_has_tx_in_progress(stream))
+		return true;
 	const struct rlist *nodes_rows = stream->nodes_rows;
 	for (int i = 0; i < VCLOCK_MAX; i++) {
-		if (!rlist_empty((struct rlist *)&nodes_rows[i])) {
-			unfinished = true;
-			break;
-		}
+		if (!rlist_empty((struct rlist *)&nodes_rows[i]))
+			return true;
 	}
-
-	return has || unfinished;
+	return false;
 }
```

src/box/box.cc
```diff
@@ -641,18 +646,39 @@ wal_stream_abort(struct wal_stream *stream)
...
+/**
+ * The function checks that the tail of the transaction from the journal
+ * has been read.
+ */
+static bool
+wal_stream_has_unfinished_tx(const struct wal_stream *stream)
+{
+	bool has = wal_stream_has_tx_in_progress(stream);
+
+	bool unfinished = false;
```
Gerold103, line 665: There needs to be a comment explaining why "in progress" != "unfinished".

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
 	return -1;
 }
 
+/* Storage, adding rows to a rlist to separate mixed transactions. */
+struct wal_row {
```
Gerold103, line 827: I couldn't parse the comment. Could you please rephrase?

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	/* Base class. */
```
Gerold103, line 828: Out-of-function comments should be opened with `/**`.

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+static struct wal_row *
+wal_stream_save_row(struct wal_stream *stream, struct xrow_header *row)
+{
+	static int64_t lsr_id = 0;
+	struct lsregion *lsr = &stream->lsr;
+	struct wal_row *new_row =
+		(struct wal_row *)xlsregion_alloc(lsr, sizeof(struct wal_row),
+						  ++lsr_id);
+	new_row->lsr_id = lsr_id;
+	memcpy(new_row, row, sizeof(struct xrow_header));
```
Gerold103, line 849: You can just do `new_row->row = *row`. Doing `memcpy()` over different struct types should be better avoided. It will break if you add any member before `wal_row::row`.

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	/* A growing identifier to track lsregion allocations. */
+	int64_t lsr_id;
+};
+
+/*
+ * Callback to stash row and row bodies upon receipt, used to recover
+ * mixed transactions.
+ */
+static struct wal_row *
+wal_stream_save_row(struct wal_stream *stream, struct xrow_header *row)
```
Gerold103, line 841: `row` can be made `const`.

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+		new_row->row.body[0].iov_base = new_base;
+	}
+	return new_row;
+}
+
+/*
+ * Find the min lsr_id that is still needed.
+ */
+static int64_t
+find_min_lsr_id(struct wal_stream *stream)
```
Gerold103, line 866: This seems to be a method of `wal_stream`, which means it needs to start with `wal_stream_` prefix. Like `wal_stream_find_min_lsr_id()`.

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	int64_t lsr_id = find_min_lsr_id(stream);
+	lsregion_gc(lsr, lsr_id);
+}
+
+/*
+ * When restoring the log is read row-by-row. However, it is necessary
+ * to store the rows for correct further recovery. For example, with
+ * mixed transactions. The following function saves the newly arrived
+ * row to the rlist, which is stored in array is nodes_rows in cell is
+ * replica_id. As soon as a row arrives with the is_commit flag set, the
```
Gerold103, line 908: Sorry, I couldn't parse this sentence. Can you please rephrase?

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+{
+	uint32_t id = row->replica_id;
+	struct rlist *nodes_rows = stream->nodes_rows;
+
+	/*
+	 * When a new row arrives, it has the same address. Because the
+	 * row is on the stack. Therefore, it needs to allocate memory on
+	 * the heap. We do with row->body[0].iov_base analogically.
+	 */
+	struct wal_row *save_row = wal_stream_save_row(stream, row);
```
Gerold103, line 925: This comment seems to be out of place. You are not doing anything with `row` members in here. Only inside `wal_stream_save_row()`. I would just drop it.

src/box/box.cc
```diff
@@ -797,14 +823,136 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+	rlist_add_tail_entry(&nodes_rows[id], save_row, in_row_list);
+
+	if (!row->is_commit) {
+		return 0;
+	} else {
```
Gerold103, line 930: This `else` branch can be dropped to reduce the indentation. You did `return` in the main branch, which means `else` is not needed.

src/box/box.cc
```diff
@@ -850,12 +1001,23 @@ wal_stream_create(struct wal_stream *ctx)
...
 	ctx->first_row_lsn = 0;
 	ctx->has_yield = false;
 	ctx->has_global_row = false;
 }
 
+static void
+wal_stream_destroy(struct wal_stream *ctx)
+{
+	lsregion_destroy(&ctx->lsr);
+	memset(ctx, 0, sizeof(*ctx));
```
Gerold103, line 1018: We usually use `TRASH(ctx)`.

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
```diff
@@ -0,0 +1,127 @@
...
+    }
+    cg.cluster:start()
+end)
+
+g.after_each(function(cg)
+    cg.cluster:drop()
+end)
+
+g.test_good_xlog_with_force_recovery = function(cg)
+    --Delete all *.xlogs on the server
```
Gerold103, line 23: Please, add a whitespace after `--`. Same other places.

Gerold103 (changes requested):
Thanks for the patch! I only have some nits. Also I didn't look at the tests too deeply. Will do that later when we fix the other comments.

src/box/box.cc
yanshtunder: This looks better. Fixed.

src/box/box.cc
yanshtunder: Done.

src/box/box.cc
yanshtunder: Fixed.

src/box/box.cc
yanshtunder: Fixed.

src/box/box.cc
yanshtunder: Done.

src/box/box.cc
yanshtunder: Ok.

src/box/box.cc
yanshtunder: Yes, indeed.

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
yanshtunder: Ok.

src/box/box.cc
yanshtunder: Ок.

src/box/box.cc
yanshtunder: Fixed.

src/box/box.cc
yanshtunder: Rephrased.

src/box/box.cc
yanshtunder: Done.

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
```diff
@@ -0,0 +1,13 @@
...
+  `box.cfg.force_recovery` to `true`. If you need to revert to the old
+  behavior, don't set the `force_recovery` option.
+
+* Let there be two nodes (`node#1` and `node#2`). And let the data be
+  replicated from `node#1` to `node#2`. Suppose that at some point in time,
+  `node#1` is restoring data from an xlog containing mixed transactions. To
+   replicate data from `node#1` to `node#2`, do the following:
+    1. Stop `node#2` and delete all xlog files from it
+    2. Restart `node#1` by setting `force_recovery` to `true`
+    3. Make `node#2` rejoin to `node#1` again (gh-7932).
```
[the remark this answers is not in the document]
yanshtunder: I threw out it of the changelog. I left this recipe in the commit message.

changelogs/unreleased/gh-7932-recovery-mixed-transactions.md
```diff
@@ -0,0 +1,13 @@
+## feature/replication
```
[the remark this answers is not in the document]
yanshtunder: 1) Fixed.
2) Yes, this is a recipe.

src/box/box.cc
Gerold103: From the comment I still couldn't understand it. Is "unfinished" tx the one, which has all the data, but simply isn't committed yet? Or what?

src/box/box.cc
Gerold103: Nope.

sergepetrenko:
Just thought about it:
please don't forget to check that transactions with both local and global spaces recover fine with your patch:

as far as I understand the results will be wrong now:
for example, you have a transaction with both local and global spaces, so some of its rows are marked with replica id 0 (the local ones) and some are marked with replica id 1 (the global ones).

With a transaction [begin 0:1, 1:1, 0:2, 1:2 commit] your code will accumulate [begin 0:1, 0:2] for 0-th replica id, and [1:1, 1:2 commit] for 1st replica id and will try to apply [1:1, 1:2 commit]. This will fail, because it's not a correct transaction. It lacks a "begin". Moreover, the "transaction" [begin 0:1, 0:2] lacks a "commit", so will never be applied.

sergepetrenko (changes requested):
Just thought about it:
please don't forget to check that transactions with both local and global spaces recover fine with your patch:

as far as I understand the results will be wrong now:
for example, you have a transaction with both local and global spaces, so some of its rows are marked with replica id 0 (the local ones) and some are marked with replica id 1 (the global ones).

With a transaction [begin 0:1, 1:1, 0:2, 1:2 commit] your code will accumulate [begin 0:1, 0:2] for 0-th replica id, and [1:1, 1:2 commit] for 1st replica id and will try to apply [1:1, 1:2 commit]. This will fail, because it's not a correct transaction. It lacks a "begin". Moreover, the "transaction" [begin 0:1, 0:2] lacks a "commit", so will never be applied.

src/box/box.cc
yanshtunder: Done.

Gerold103 (changes requested):
Please, re-request the review when the problem described above is fixed.

yanshtunder:
> Просто подумал об этом: пожалуйста, не забудьте проверить, что транзакции с локальными и глобальными пространствами нормально восстанавливаются с вашим патчем:
> 
> насколько я понимаю, теперь результаты будут неверными: например, у вас есть транзакция с локальным и глобальным пространством, поэтому некоторые ее строки помечены идентификатором реплики 0 (локальные), а некоторые - идентификатором реплики 1. (глобальные).
> 
> С транзакцией [begin 0:1, 1:1, 0:2, 1:2 commit] ваш код будет накапливать [begin 0:1, 0:2] для 0-го идентификатора реплики и [1:1, 1 :2 commit] для 1-го идентификатора реплики и попытается применить [1:1, 1:2 commit]. Это не удастся, потому что это неправильная транзакция. Не хватает «начала». Более того, в «транзакции» [начало 0:1, 0:2] отсутствует «фиксация», поэтому она никогда не будет применена.

All done

src/box/box.cc
yanshtunder: Fixed.

src/box/box.cc
```diff
@@ -797,14 +823,139 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
...
+			       struct xrow_header *row)
+{
+	/*
+	 * A local transaction should not be written to nodes_rows with
+	 * id equal to 0. It should be written to nodes_rows with the id
+	 * of the node from which we are recovering. Since a local
+	 * transaction can only exist on the node from which we are
+	 * recovering.
+	 */
+	uint32_t id = row->replica_id ? row->replica_id : instance_id;
```
Gerold103, line 926: What is this is an anonymous replica? Its `instance_id` is zero then. And it can write to replica-local spaces, hence creating mixed transactions.

Gerold103 (commented):
Thanks for the fixes! One question is left.

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
```diff
@@ -0,0 +1,167 @@
...
+        return box.space.glob:select()
+    end), {{1, 1}, {2, 1}, {3, 1}})
+    t.assert_equals(cg.server:exec(function()
+        return box.space.loc:select()
+    end), {{1, 0}, {2, 0}, {3, 0}})
+
+    cg.server:stop()
+    local xlog = fio.pathjoin(cg.server.workdir,
+                              '00000000000000000012.xlog')
+    fio.unlink(xlog)
```
sergepetrenko, line 48: Why do you need to unlink the xlog here?

sergepetrenko (approved):
Thanks for the fixes! All looks good except one question regarding the test.

src/box/box.cc
yanshtunder: I create new issue #8746

test/replication-luatest/gh_7932_recovery_mixed_transactions_test.lua
yanshtunder: Can be left.

Gerold103 (approved):
Thanks for the fixes! I don't see any problems atm. (Except that if some instance ID will have rows not ending with `is_commit` in the beginning of WAL, then lsregion won't be garbage collected ever and the entire WAL will be saved in it, but it is unlikely to happen (probably).)

yanshtunder:
box.cc
```diff
static struct wal_row *
wal_stream_save_row(struct wal_stream *stream,
                    const struct xrow_header *row)
 {
        static int64_t lsr_id = 0;
        struct lsregion *lsr = &stream->lsr;
-       struct wal_row *new_row = (struct wal_row *)xlsregion_alloc(lsr,
-                                  sizeof(struct wal_row), ++lsr_id);
+       struct wal_row *new_row = xlsregion_alloc_object(lsr, ++lsr_id,
+                                                        typeof(*new_row));
        new_row->lsr_id = lsr_id;
```
util.h
``` diff
 #define xlsregion_alloc(p, size, id) \
        xalloc_impl((size), lsregion_alloc, (p), (size), (id))
-
+#define xlsregion_aligned_alloc(p, size, align, id) \
+       xalloc_impl((size), lsregion_aligned_alloc, (p), (size), (align), (id))
+#define xlsregion_alloc_object(lsregion, id, T) ({                             \
+       (T *)xlsregion_aligned_alloc((lsregion), sizeof(T), alignof(T), (id));  \
+})
 #define xregion_alloc_object(region, T) ({                                     \
        (T *)xregion_aligned_alloc((region), sizeof(T), alignof(T));            \
 })

**Comment by sergepetrenko** [comment]
Applied the following diff: changed xlogs for testing to grant the 'guest' user a 'super' role instead of 'replication':

<details><summary>binary diff</summary>

```diff
diff --git a/test/replication-luatest/gh_7932_data/bad_xlog/00000000000000000000.xlog b/test/replication-luatest/gh_7932_data/bad_xlog/00000000000000000000.xlog
index 3d28648ff..fabf5569d 100644
--- a/test/replication-luatest/gh_7932_data/bad_xlog/00000000000000000000.xlog
+++ b/test/replication-luatest/gh_7932_data/bad_xlog/00000000000000000000.xlog
@@ -4,10 +4,10 @@
 00000030: 3161 3638 3565 3633 2d32 3463 642d 3466  1a685e63-24cd-4f
 00000040: 6335 2d39 3533 322d 6266 3835 6636 3439  c5-9532-bf85f649
 00000050: 6530 6162 0a56 436c 6f63 6b3a 207b 7d0a  e0ab.VClock: {}.
-00000060: 0ad5 ba0b ab21 00ce 4c7b 80d4 a700 0000  .....!..L{......
+00000060: 0ad5 ba0b ab21 00ce a020 d331 a700 0000  .....!... .1....
 00000070: 0000 0000 8400 0302 0103 0104 cb41 d901  .............A..
 00000080: 6d58 09b4 2182 10cd 0138 2195 0100 a472  mX..!....8!....r
-00000090: 6f6c 6503 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
+00000090: 6f6c 651f 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
 000000a0: a700 0000 0000 0000 8400 0402 0103 0204  ................
 000000b0: cb41 d901 6d5b 73fb fa84 10cd 0110 1501  .A..m[s.........
 000000c0: 2091 a66d 6178 5f69 6421 9193 a12b 0201   ..max_id!...+..
diff --git a/test/replication-luatest/gh_7932_data/good_xlog/00000000000000000000.xlog b/test/replication-luatest/gh_7932_data/good_xlog/00000000000000000000.xlog
index 30f6f5bee..ef93c5c4c 100644
--- a/test/replication-luatest/gh_7932_data/good_xlog/00000000000000000000.xlog
+++ b/test/replication-luatest/gh_7932_data/good_xlog/00000000000000000000.xlog
@@ -4,10 +4,10 @@
 00000030: 3161 3638 3565 3633 2d32 3463 642d 3466  1a685e63-24cd-4f
 00000040: 6335 2d39 3533 322d 6266 3835 6636 3439  c5-9532-bf85f649
 00000050: 6530 6162 0a56 436c 6f63 6b3a 207b 7d0a  e0ab.VClock: {}.
-00000060: 0ad5 ba0b ab21 00ce 4c7b 80d4 a700 0000  .....!..L{......
+00000060: 0ad5 ba0b ab21 00ce a020 d331 a700 0000  .....!... .1....
 00000070: 0000 0000 8400 0302 0103 0104 cb41 d901  .............A..
 00000080: 6d58 09b4 2182 10cd 0138 2195 0100 a472  mX..!....8!....r
-00000090: 6f6c 6503 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
+00000090: 6f6c 651f 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
 000000a0: a700 0000 0000 0000 8400 0402 0103 0204  ................
 000000b0: cb41 d901 6d5b 73fb fa84 10cd 0110 1501  .A..m[s.........
 000000c0: 2091 a66d 6178 5f69 6421 9193 a12b 0201   ..max_id!...+..
diff --git a/test/replication-luatest/gh_7932_data/not_finished_xlog/00000000000000000000.xlog b/test/replication-luatest/gh_7932_data/not_finished_xlog/00000000000000000000.xlog
index 7a36674d6..6bc9c2fdb 100644
--- a/test/replication-luatest/gh_7932_data/not_finished_xlog/00000000000000000000.xlog
+++ b/test/replication-luatest/gh_7932_data/not_finished_xlog/00000000000000000000.xlog
@@ -4,10 +4,10 @@
 00000030: 3161 3638 3565 3633 2d32 3463 642d 3466  1a685e63-24cd-4f
 00000040: 6335 2d39 3533 322d 6266 3835 6636 3439  c5-9532-bf85f649
 00000050: 6530 6162 0a56 436c 6f63 6b3a 207b 7d0a  e0ab.VClock: {}.
-00000060: 0ad5 ba0b ab21 00ce 4c7b 80d4 a700 0000  .....!..L{......
+00000060: 0ad5 ba0b ab21 00ce a020 d331 a700 0000  .....!... .1....
 00000070: 0000 0000 8400 0302 0103 0104 cb41 d901  .............A..
 00000080: 6d58 09b4 2182 10cd 0138 2195 0100 a472  mX..!....8!....r
-00000090: 6f6c 6503 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
+00000090: 6f6c 651f 04d5 ba0b ab28 00ce d74b 9ee1  ole......(...K..
 000000a0: a700 0000 0000 0000 8400 0402 0103 0204  ................
 000000b0: cb41 d901 6d5b 73fb fa84 10cd 0110 1501  .A..m[s.........
 000000c0: 2091 a66d 6178 5f69 6421 9193 a12b 0201   ..max_id!...+..
```
</details>

PR #8542, opened 2023-04-06
lua: add malloc info

### LINKED ISSUE #7311: Add malloc statistics
Tarantool heavily uses malloc. Sometimes memory allocated with malloc exceeds memory used by the runtime arena and even the memtx engine. Users often complain that Tarantool leaks memory while it just uses it for internal objects allocated with malloc. We need to export malloc statistics somewhere (`box.malloc.info`, `box.info.malloc`, `box.info.memory.malloc`, `box.runtime.info.malloc`?).

There are two ways how we could gather malloc statistics:
1. Wrap the `malloc` call in a function that accounts the amount of memory allocated with `malloc` in per-thread counters.
   Pros: platform-independent. Cons: doesn't account malloc fragmentation.
2. Use `malloc_info` with `open_mstream` glibc functions, then decode the stats using xml decoder and report them as a Lua table.
   Pros: excessive malloc stats. Cons: isn't available on all platforms; the output is subject to change.

We probably should implement both.

locker:
This commit adds two new Lua functions:
 - `box.malloc.info()` is a public function that returns the total memory usage of internal data structures allocated with [`malloc()`][1]. This information is extracted from the [`malloc_info()`][2] output.
 - `box.malloc.internal.info()` is an internal function that returns a parsed XML document returned by `malloc_info()` as is. It isn't documented. We may need it to get the detailed information about the system memory allocator.

Both functions are available only on Linux and only if ASAN is disabled because otherwise `malloc_info()` is unavailable. On other platforms `box.malloc.info()` reports zero memory usage while `internal.info()` returns an empty table.

Both functions may be used before `box.cfg()` is called.

`box.malloc.info()` returns a table with two fields:

```
tarantool> box.malloc.info()
---
- size: 2498560
  used: 1835726
...
```

The `used` value is the total amount of memory in bytes allocated by Tarantool for internal data structures with `malloc()`.

The `size` value is the total amount of memory in bytes allocated from the system by the `malloc()` allocator. It may not be less than `used`.

Since `malloc_info()` returns an XML document, we also have to add an internal XML parser.

Closes #7311

[1]: https://man7.org/linux/man-pages/man3/malloc.3.html
[2]: https://man7.org/linux/man-pages/man3/malloc_info.3.html

### DIFF
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -32,12 +32,22 @@
 
 #include "box/lua/slab.h"
 #include "lua/utils.h"
+#include "lua/xml.h"
 #include "lua/serializer.h" /* luaL_setmaphint */
 
 #include <lua.h>
 #include <lauxlib.h>
 #include <lualib.h>
 #include <lj_obj.h> /* internals: lua in box.runtime.info() */
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#if defined(HAVE_MALLOC_INFO)
+# include <malloc.h>
+#endif /* defined(HAVE_MALLOC_INFO) */
 
 #include "small/small.h"
 #include "small/quota.h"
@@ -265,6 +275,144 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
 	return 0;
 }
 
+/*
+ * Decodes and returns the XML document returned by malloc_info() as is.
+ *
+ * This is an internal function that isn't supposed to be called by users,
+ * but it may be useful for developers.
+ *
+ * Returns an empty table if malloc_info() isn't supported by the system.
+ * Raises a Lua error if it fails to retrieve or parse malloc_info() output.
+ */
+static int
+lbox_malloc_internal_info(struct lua_State *L)
+{
+#if defined(HAVE_MALLOC_INFO)
+	char *buf = NULL;
+	size_t buf_size = 0;
+	FILE *fp = open_memstream(&buf, &buf_size);
+	if (fp == NULL) {
+		diag_set(SystemError, "failed to open memory stream");
+		return luaT_error(L);
+	}
+	if (malloc_info(0, fp) != 0) {
+		diag_set(SystemError, "failed to get malloc info");
+		fclose(fp);
+		free(buf);
+		return luaT_error(L);
+	}
+	fclose(fp);
+	lua_pushlstring(L, buf, buf_size);
+	free(buf);
+	return luaT_xml_decode(L);
+#else /* !defined(HAVE_MALLOC_INFO) */
+	lua_newtable(L);
+	return 1;
+#endif /* defined(HAVE_MALLOC_INFO) */
+}
+
+/*
+ * Returns the malloc memory usage information in a table
+ *
+ *   {
+ *     size = <total allocated>,
+ *     used = <actually used>,
+ *   }
+ *
+ * (all numbers are in bytes).
+ *
+ * The information is retrieved with malloc_info(). If it isn't supported by
+ * the system or its format is unknown, {size = 0, used = 0} is returned.
+ *
+ * This function never raises.
+ */
+static int
+lbox_malloc_info(struct lua_State *L)
+{
+	int version = 0;
+	uint64_t total = 0;
+	uint64_t available = 0;
+	lua_pushcfunction(L, lbox_malloc_internal_info);
+	if (luaT_call(L, 0, 1) != 0)
+		goto out;
+	/*
+	 * The XML document name returned by malloc_info is expected to be
+	 * "malloc" so the document content should be in malloc[1].
+	 */
+	assert(lua_istable(L, -1));
+	lua_getfield(L, -1, "malloc");
+	if (!lua_istable(L, -1))
+		goto out;
+	lua_rawgeti(L, -1, 1);
+	if (!lua_istable(L, -1))
+		goto out;
+	/*
+	 * First, check the malloc_info version. The only known version is 1.
+	 * It's pointless to proceed if the version is different.
+	 */
+	lua_getfield(L, -1, "version");
+	version = lua_tonumber(L, -1);
+	lua_pop(L, 1);
+	if (version != 1)
+		goto out;
+	/*
+	 * Extract the size of used memory. Even though the document version is
+	 * valid, we still need to be careful accessing it.
+	 */
+	assert(lua_istable(L, -1));
+	lua_getfield(L, -1, "system");
+	if (lua_istable(L, -1)) {
+		lua_pushnil(L);
+		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
+			if (!lua_istable(L, -1))
+				continue; /* skip XML attributes */
+			lua_getfield(L, -1, "type");
+			const char *key = lua_tostring(L, -1);
+			lua_pop(L, 1);
+			lua_getfield(L, -1, "size");
+			uint64_t size = luaL_touint64(L, -1);
+			lua_pop(L, 1);
+			if (key == NULL) {
+				continue;
+			} else if (strcmp(key, "current") == 0) {
+				total += size;
+			}
+		}
+	}
+	lua_pop(L, 1);
+	lua_getfield(L, -1, "total");
+	if (lua_istable(L, -1)) {
+		lua_pushnil(L);
+		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
+			if (!lua_istable(L, -1))
+				continue; /* skip XML attributes */
+			lua_getfield(L, -1, "type");
+			const char *key = lua_tostring(L, -1);
+			lua_pop(L, 1);
+			lua_getfield(L, -1, "size");
+			uint64_t size = luaL_touint64(L, -1);
+			lua_pop(L, 1);
+			if (key == NULL) {
+				continue;
+			} else if (strcmp(key, "mmap") == 0) {
+				total += size;
+			} else if (strcmp(key, "fast") == 0 ||
+				   strcmp(key, "rest") == 0) {
+				available += size;
+			}
+		}
+	}
+	lua_pop(L, 1);
+out:
+	/* Return memory usage information. */
+	lua_newtable(L);
+	luaL_pushuint64(L, total);
+	lua_setfield(L, -2, "size");
+	luaL_pushuint64(L, total - available);
+	lua_setfield(L, -2, "used");
+	return 1;
+}
+
 /** Initialize box.slab package. */
 void
 box_lua_slab_init(struct lua_State *L)
@@ -295,6 +443,23 @@ box_lua_slab_init(struct lua_State *L)
 	lua_settable(L, -3);
 
 	lua_settable(L, -3); /* box.runtime */
+
+	lua_pushstring(L, "malloc");
+	lua_newtable(L);
+
+	lua_pushstring(L, "info");
+	lua_pushcfunction(L, lbox_malloc_info);
+	lua_settable(L, -3);
+
+	lua_pushstring(L, "internal");
+	lua_newtable(L);
+
+	lua_pushstring(L, "info");
+	lua_pushcfunction(L, lbox_malloc_internal_info);
+	lua_settable(L, -3);
+
+	lua_settable(L, -3); /* box.malloc.internal */
+	lua_settable(L, -3); /* box.malloc */
 
 	lua_pop(L, 1); /* box. */
 }
--- a/src/lua/xml.c
+++ b/src/lua/xml.c
@@ -1,0 +1,324 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include "lua/xml.h"
+
+#include <assert.h>
+#include <ctype.h>
+#include <lua.h>
+#include <lauxlib.h>
+#include <stddef.h>
+#include <string.h>
+
+#include "lua/utils.h"
+
+int
+luaT_xml_decode(struct lua_State *L)
+{
+	/* The input string must be at the top of the stack. */
+	int top = lua_gettop(L);
+	if (top < 1 || lua_type(L, top) != LUA_TSTRING)
+		return luaL_error(L, "expected string");
+	const char *s = lua_tostring(L, top);
+	assert(s != NULL);
+	/* Push the document table. */
+	lua_newtable(L);
+	enum {
+		ELEM,
+		TAG,
+		START_TAG,
+		END_TAG,
+		SPACE_AFTER_END_TAG,
+		ATTR,
+		ATTR_NAME,
+		ATTR_VALUE_SEP,
+		ATTR_VALUE_BEGIN,
+		ATTR_VALUE,
+		ATTR_VALUE_END,
+		ELEM_END,
+		DOC_END,
+	} state = ELEM;
+	const char *tag_name = NULL;
+	const char *attr_name = NULL;
+	const char *attr_value = NULL;
+	/*
+	 * Position of the current character in the input string recorded
+	 * as line and column numbers. Used for error reporting.
+	 */
+	int line = 1;
+	int column = 1;
+#define ERROR(msg) \
+	luaL_error(L, "XML decode error at line %d, column %d: %s", \
+		   line, column, (msg))
+	while (*s != '\0') {
+		/*
+		 * State machine:
+		 *  - 'break' skips the current character by leaving
+		 *    the switch-case.
+		 *  - 'continue' rechecks the state without skipping
+		 *    the current character.
+		 */
+		switch (state) {
+		/*
+		 * Expected an element. Currently, only tags are supported
+		 * (<section> or </section>) while values enclosed in tags are
+		 * considered invalid. Skip optional spaces and '<', then
+		 * transition to TAG.
+		 *
+		 * TODO: Handle values, such as <section>value</section>.
+		 */
+		case ELEM:
+			if (isspace(*s))
+				break; /* skip spaces */
+			if (*s != '<')
+				return ERROR("invalid token");
+			state = TAG;
+			break; /* skip '<' */
+		/*
+		 * If the current character is '/', skip it and transition to
+		 * END_TAG, otherwise transition to START_TAG.
+		 */
+		case TAG:
+			if (*s != '/') {
+				tag_name = s;
+				state = START_TAG;
+				continue;
+			}
+			if (lua_gettop(L) == top + 1) {
+				/*
+				 * The stack only contains the resulting
+				 * document table, which means there's no
+				 * start tag matching this end tag.
+				 */
+				return ERROR("invalid token");
+			}
+			tag_name = s + 1;
+			state = END_TAG;
+			break; /* skip '/' */
+		/*
+		 * Skip the start tag name, create a table for the tag
+		 * attributes and child elements, then transition to ATTR.
+		 */
+		case START_TAG: {
+			assert(tag_name != NULL);
+			/* TODO: Check names according to XML standard. */
+			if (isalnum(*s))
+				break; /* skip the tag name */
+			if ((!isspace(*s) && *s != '/' && *s != '>') ||
+			    s == tag_name)
+				return ERROR("invalid token");
+			/* Stack: parent table, ... */
+			assert(lua_gettop(L) >= top + 1);
+			assert(lua_istable(L, -1));
+			lua_pushlstring(L, tag_name, s - tag_name);
+			/*
+			 * Check that there's no attributes with the same name
+			 * in the parent table. A table is accepted because
+			 * there may be more than one element with the same
+			 * name (elements are stored in an array).
+			 */
+			lua_pushvalue(L, -1); /* tag name */
+			lua_rawget(L, -3);
+			int type = lua_type(L, -1);
+			if (type != LUA_TNIL && type != LUA_TTABLE)
+				return ERROR("duplicate name");
+			/*
+			 * Create an array entry in the parent table under this
+			 * name if it doesn't exist.
+			 */
+			if (type == LUA_TNIL) {
+				lua_pop(L, 1);
+				lua_newtable(L);
+				lua_pushvalue(L, -2); /* tag name */
+				lua_pushvalue(L, -2); /* new array */
+				lua_rawset(L, -5);
+			}
+			assert(lua_istable(L, -1));
+			int len = lua_objlen(L, -1);
+			/*
+			 * Create a new attribute table for this element and
+			 * append it to the array. Then pop the array leaving
+			 * the attribute table and the new tag name at the top
+			 * of the stack.
+			 */
+			lua_newtable(L);
+			lua_pushvalue(L, -1); /* attribute table */
+			lua_rawseti(L, -3, len + 1);
+			lua_replace(L, -2);
+			state = ATTR;
+			continue;
+		}
+		/*
+		 * Skip the end tag name, check it against the corresponding
+		 * start tag name, and transition to SPACE_AFTER_END_TAG.
+		 */
+		case END_TAG: {
+			assert(tag_name != NULL);
+			if (isalnum(*s))
+				break; /* skip the tag name */
+			if ((!isspace(*s) && *s != '>') ||
+			    s == tag_name)
+				return ERROR("invalid token");
+			/* Stack: attr table, tag name, parent table, ... */
+			assert(lua_gettop(L) >= top + 3);
+			/* Check that the start and end tag names match. */
+			size_t len;
+			const char *expected = lua_tolstring(L, -2, &len);
+			assert(expected != NULL);
+			if ((size_t)(s - tag_name) != len ||
+			    strncmp(tag_name, expected, len) != 0)
+				return ERROR("mismatched tag");
+			state = SPACE_AFTER_END_TAG;
+			continue;
+		}
+		/*
+		 * Skip optional spaces after the end tag name and transition
+		 * to ELEM_END.
+		 */
+		case SPACE_AFTER_END_TAG:
+			if (isspace(*s))
+				break; /* skip spaces */
+			state = ELEM_END;
+			continue;
+		/*
+		 * Skip optional spaces then transition to ELEM, ELEM_END, or
+		 * ATTR_NAME, depending on the current character.
+		 */
+		case ATTR:
+			if (isspace(*s))
+				break; /* skip spaces */
+			if (*s == '/') {
+				state = ELEM_END;
+				break; /* skip '/' */
+			} else if (*s == '>') {
+				state = ELEM;
+				break; /* skip '>' */
+			}
+			attr_name = s;
+			state = ATTR_NAME;
+			continue;
+		/*
+		 * Skip the attribute name and transition to ATTR_VALUE_SEP.
+		 */
+		case ATTR_NAME:
+			assert(attr_name != NULL);
+			/* TODO: Check names according to XML standard. */
+			if (isalnum(*s))
+				break; /* skip the attribute name */
+			if ((!isspace(*s) && *s != '=') ||
+			    s == attr_name)
+				return ERROR("invalid token");
+			/* Stack: attr table, tag name, parent table, ... */
+			assert(lua_gettop(L) >= top + 3);
+			lua_pushlstring(L, attr_name, s - attr_name);
+			/* Check that there's no duplicate attributes. */
+			lua_pushvalue(L, -1); /* attribute name */
+			lua_rawget(L, -3);
+			if (!lua_isnil(L, -1))
+				return ERROR("duplicate name");
+			lua_pop(L, 1);
+			state = ATTR_VALUE_SEP;
+			continue;
+		/*
+		 * Skip optional spaces and '=' separating the attribute value
+		 * from the name, then transition to ATTR_VALUE_BEGIN.
+		 */
+		case ATTR_VALUE_SEP:
+			if (isspace(*s))
+				break; /* skip spaces */
+			if (*s != '=')
+				return ERROR("invalid token");
+			state = ATTR_VALUE_BEGIN;
+			break; /* skip '=' */
+		/*
+		 * Skip optional spaces and '"' preceding the attribute value,
+		 * then transition to ATTR_VALUE.
+		 */
+		case ATTR_VALUE_BEGIN:
+			if (isspace(*s))
+				break; /* skip spaces */
+			if (*s != '"')
+				return ERROR("invalid token");
+			attr_value = s + 1;
+			state = ATTR_VALUE;
+			break; /* skip '"' */
+		/*
+		 * Skip until and including '"' following the attribute value,
+		 * insert the new attribute to the attribute table, then
+		 * transition to ATTR_VALUE_END.
+		 */
+		case ATTR_VALUE:
+			assert(attr_value != NULL);
+			/* TODO: Handle escaped quotation marks. */
+			if (*s != '"')
+				break; /* skip until '"' */
+			/*
+			 * Stack: attr name, attr table, tag name,
+			 *        parent table, ...
+			 */
+			assert(lua_gettop(L) >= top + 4);
+			/* Insert the new attribute into the table. */
+			lua_pushlstring(L, attr_value, s - attr_value);
+			lua_rawset(L, -3);
+			state = ATTR_VALUE_END;
+			break; /* skip '"' */
+		/*
+		 * Check that the attribute value is followed by a valid token,
+		 * then transition to ATTR.
+		 */
+		case ATTR_VALUE_END:
+			if (!isspace(*s) && *s != '/' && *s != '>')
+				return ERROR("invalid token");
+			state = ATTR;
+			continue;
+		/*
+		 * Skip '>' terminating the current element, pop the attribute
+		 * table and the tag name created for the element from the
+		 * stack, then transition to ELEM or DOC_END, depending on
+		 * whether we expect more elements.
+		 */
+		case ELEM_END:
+			if (*s != '>')
+				return ERROR("invalid token");
+			/* Stack: attr table, tag name, parent table, ... */
+			assert(lua_gettop(L) >= top + 3);
+			lua_pop(L, 2);
+			state = lua_gettop(L) == top + 1 ? DOC_END : ELEM;
+			break; /* skip '>' */
+		/*
+		 * End of document. No input except for spaces is expected in
+		 * this state.
+		 */
+		case DOC_END:
+			if (isspace(*s))
+				break; /* skip spaces */
+			return ERROR("junk after document");
+		}
+		if (*s == '\n') {
+			line++;
+			column = 1;
+		} else {
+			column++;
+		}
+		s++;
+	}
+	if (state != DOC_END)
+		return ERROR("truncated document");
+#undef ERROR
+	/* Replace the input string with the document table. */
+	lua_replace(L, -2);
+	return 1;
+}
+
+void
+tarantool_lua_xml_init(struct lua_State *L)
+{
+	const struct luaL_Reg module_funcs[] = {
+		{"decode", luaT_xml_decode},
+		{NULL, NULL},
+	};
+	luaT_newmodule(L, "internal.xml", module_funcs);
+	lua_pop(L, 1);
+}
--- a/test/app-luatest/gh_7311_malloc_info_test.lua
+++ b/test/app-luatest/gh_7311_malloc_info_test.lua
@@ -1,0 +1,115 @@
+local tarantool = require('tarantool')
+
+local ffi = require('ffi')
+ffi.cdef([[
+    void *malloc(size_t size);
+    void free(void *p);
+]])
+local malloc = ffi.C.malloc
+local free = ffi.C.free
+
+local t = require('luatest')
+local g = t.group()
+
+local ALLOC_SIZE = 10 * 1000 * 1000
+local MARGIN = ALLOC_SIZE * 0.05
+
+local function is_supported()
+    -- malloc_info() is a GNU extension available only on Linux.
+    if jit.os ~= 'Linux' then
+        return false
+    end
+    -- If ASAN is enabled, malloc_info() exists but it is not implemented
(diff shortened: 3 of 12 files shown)


locker:
Setting `full-ci` to check that this patch works fine on systems that don't support `malloc_info()` (Mac OS, FreeBSD).

p7nov (approved):

src/lua/xml.c
```diff
@@ -0,0 +1,318 @@
...
+#include <ctype.h>
+#include <lua.h>
+#include <lauxlib.h>
+#include <stddef.h>
+#include <string.h>
+
+#include "lua/utils.h"
+
+int
+luaT_xml_decode(struct lua_State *L)
```
olegrok, line 18: Can we onboard luarapidxml (https://github.com/tarantool/luarapidxml) module?

We have json/yaml serializers for now. If we will have built-in xml parser it could be useful in some cases.

locker: I was aware of the luarapidxml module, but I decided not to use it, because:
1. Onboarding the rapidxml library to parse XML documents returned by `malloc_info()` looks like an overkill. The size of `src/lua/xml.c` added by this PR is less than the Lua wrapper around the rapidxml library alone.
2. The output format of the `luarapidxml` is too verbose for my goal. I don't want `box.malloc.internal.info()` to have `tag` and `attr` keys for each entry. Compare:
   <details><summary>internal.xml</summary>
   
   ```
   tarantool> box.malloc.internal.info()
   ---
   - malloc:
     - heap:
       - nr: '0'
         system:
         - type: current
           size: '2523136'
         - type: max
           size: '2523136'
         sizes:
         - size:
           - count: '2'
             from: '33'
             to: '33'
             total: '66'
           - count: '5'
             from: '49'
             to: '49'
             total: '245'
           - count: '1'
             from: '193'
             to: '193'
             total: '193'
           - count: '1'
             from: '289'
             to: '289'
             total: '289'
           - count: '1'
             from: '18081'
             to: '18081'
             total: '18081'
           - count: '1'
             from: '50865'
             to: '50865'
             total: '50865'
           unsorted:
           - count: '1'
             from: '4337'
             to: '4337'
             total: '4337'
         total:
         - type: fast
           count: '0'
           size: '0'
         - type: rest
           count: '13'
           size: '594236'
         aspace:
         - type: total
           size: '2523136'
         - type: mprotect
           size: '2523136'
       version: '1'
       system:
       - type: current
         size: '2523136'
       - type: max
         size: '2523136'
       total:
       - type: fast
         count: '0'
         size: '0'
       - type: rest
         count: '13'
         size: '594236'
       - type: mmap
         count: '0'
         size: '0'
       aspace:
       - type: total
         size: '2523136'
       - type: mprotect
         size: '2523136'
   ...
   ```
   
   </details>
   
   (inspired by https://jsonformatter.org/xml-to-yaml)
   
   <details><summary>rapidxml</summary>
   
   ```
   tarantool> box.malloc.internal.info()
   ---
   - 1:
       1:
         1:
           attr:
             count: '5'
             from: '49'
             to: '49'
             total: '245'
           tag: size
         2:
           attr:
             count: '1'
             from: '193'
             to: '193'
             total: '193'
           tag: size
         3:
           attr:
             count: '1'
             from: '289'
             to: '289'
             total: '289'
           tag: size
         4:
           attr:
             count: '1'
             from: '18081'
             to: '18081'
             total: '18081'
           tag: size
         5:
           attr:
             count: '1'
             from: '50865'
             to: '50865'
             total: '50865'
           tag: size
         6:
           attr:
             count: '1'
             from: '3345'
             to: '3345'
             total: '3345'
           tag: unsorted
         tag: sizes
       2:
         attr:
           type: fast
           count: '0'
           size: '0'
         tag: total
       3:
         attr:
           type: rest
           count: '11'
           size: '593178'
         tag: total
       4:
         attr:
           type: current
           size: '2523136'
         tag: system
       5:
         attr:
           type: max
           size: '2523136'
         tag: system
       6:
         attr:
           type: total
           size: '2523136'
         tag: aspace
       7:
         attr:
           type: mprotect
           size: '2523136'
         tag: aspace
       attr:
         nr: '0'
       tag: heap
     2:
       attr:
         type: fast
         count: '0'
         size: '0'
       tag: total
     3:
       attr:
         type: rest
         count: '11'
         size: '593178'
       tag: total
     4:
       attr:
         type: mmap
         count: '0'
         size: '0'
       tag: total
     5:
       attr:
         type: current
         size: '2523136'
       tag: system
     6:
       attr:
         type: max
         size: '2523136'
       tag: system
     7:
       attr:
         type: total
         size: '2523136'
       tag: aspace
     8:
       attr:
         type: mprotect
         size: '2523136'
       tag: aspace
     attr:
       version: '1'
     tag: malloc
   ...
   ```
   
   </details>

Initially, I planned to parse `malloc_info()` output right in `box/src/lua/slab.cc`, but decided to factor this code out into a separate internal Lua module so that I can test it with `luatest`.

If you need to have an XML parser onboard and have a reason for that, please file a ticket. But even if we decide to add a public XML parser, I'm not sure we'll actually choose `luarapidxml`. There undoubtedly will be an investigation about which XML library suits best for our purposes. Since we needed `malloc_info()` yesterday, I think it's better to add ad hoc internal XML parser for now, just for `malloc_info()` sake. Later on we can either improve it, or replace it with an external library.

olegrok: Ok. Thanks for explanation.

src/lua/xml.c
```diff
@@ -0,0 +1,318 @@
...
+		 * END_TAG, otherwise transition to START_TAG.
+		 */
+		case TAG:
+			if (*s != '/') {
+				tag_name = s;
+				state = START_TAG;
+				continue;
+			}
+			if (lua_gettop(L) == top + 1)
+				return ERROR("invalid token");
```
Totktonada, line 90: TBH, I don't get this condition.

src/lua/xml.c
```diff
@@ -0,0 +1,318 @@
...
+			if (isalnum(*s))
+				break; /* skip the tag name */
+			if ((!isspace(*s) && *s != '/' && *s != '>') ||
+			    s == tag_name)
+				return ERROR("invalid token");
+			/* Stack: parent table, ... */
+			assert(lua_gettop(L) >= top + 1);
+			assert(lua_istable(L, -1));
+			/* TODO: Check the tag name. */
+			lua_pushlstring(L, tag_name, s - tag_name);
```
Totktonada, line 109: The TODO is to check beyond the `isalnum()` and emptiness checks above? If so, what check did you suppose here?

Totktonada: The same question about `/* TODO: Check the attribute name. */` below.

src/box/lua/slab.cc
```diff
@@ -265,6 +275,126 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
...
+{
+	lbox_malloc_internal_info(L);
+	/*
+	 * The XML document name returned by malloc_info is expected to be
+	 * "malloc" so the document content should be in malloc[1].
+	 */
+	assert(lua_istable(L, -1));
+	lua_getfield(L, -1, "malloc");
+	if (lua_istable(L, -1))
+		lua_rawgeti(L, -1, 1);
```
Totktonada, line 325: AFAIU from the comment, you expect `{malloc = {<..our element..>}}`. Why there is no error otherwise?

Totktonada: Or, well, other variant of an error reporting (I'm not sure which one is the best for a statistics reporting endpoint).

src/box/lua/slab.cc
```diff
@@ -265,6 +275,126 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
...
+	 * It's pointless to proceed if the version is different.
+	 */
+	int version = 0;
+	if (lua_istable(L, -1)) {
+		lua_getfield(L, -1, "version");
+		version = lua_tonumber(L, -1);
+		lua_pop(L, 1);
+	}
+	if (version != 1)
+		return luaL_error(L, "unknown malloc info format");
```
Totktonada, line 337: Is it okay to raise an error from a statistics endpoint? I mean, if we can't collect it, shouldn't we just report some nullish output?

src/box/lua/slab.cc
```diff
@@ -296,5 +426,26 @@ box_lua_slab_init(struct lua_State *L)
...
 	lua_settable(L, -3); /* box.runtime */
 
+	lua_pushstring(L, "malloc");
+	lua_newtable(L);
+
+#if defined(HAVE_MALLOC_INFO)
+	lua_pushstring(L, "info");
+	lua_pushcfunction(L, lbox_malloc_info);
+	lua_settable(L, -3);
+#endif /* defined(HAVE_MALLOC_INFO) */
```
Totktonada, line 436: Shouldn't we replace it with a stab instead? I would expect that statistics, logging, stringification do its best, but don't raise errors.


I'm not sure here and highlighted to think together.

src/box/lua/slab.cc
```diff
@@ -265,6 +275,126 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
...
+	 */
+	uint64_t total = 0;
+	uint64_t available = 0;
+	assert(lua_istable(L, -1));
+	lua_getfield(L, -1, "system");
+	if (lua_istable(L, -1)) {
+		lua_pushnil(L);
+		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
+			if (!lua_istable(L, -1))
+				continue;
```
Totktonada, line 350: This condition is not clear at glance. Isn't a table supposed by the format that the internal XML parser returns? I would acquire some brief notes.

test/app-luatest/gh_7311_malloc_info_test.lua
```diff
@@ -0,0 +1,116 @@
...
+end
+
+local function skip_if_unsupported()
+    t.skip_if(not is_supported(), 'malloc info is not supported')
+end
+
+local function check_malloc_info(info)
+    t.assert_ge(info.used, 0)
+    t.assert_ge(info.size, info.used)
+    t.assert_le(info.size, info.used + 2 * ALLOC_SIZE)
```
Totktonada, line 33: Can you clarify this check?

Totktonada (approved):
I guess that Vladimir asked my review mainly for APIs and I'm comfortable with them in general. LGTM so.

Asked several questions. The main one is whether we should report `{used = 0, size = 0}` when the statistics is not available or should raise an error/drop the function from the module.

locker:
> The main one is whether we should report `{used = 0, size = 0}` when the statistics is not available or should raise an error/drop the function from the module.

Discussed with @alyapunov. Agreed that it's best to return `{used = 0, size = 0}` if `malloc_info()` isn't supported because this way it's easier to use in Lua (less checks). Zero memory usage clearly indicates that something's wrong. For more information one can check `box.internal.malloc.info()`. In case of overridden malloc (e.g. by ASAN) it'll return a table filled with zeros. In case `malloc_info()` isn't supported by the system, it'll return an empty table. I'll rework this patch accordingly.

src/lua/xml.c
locker: XML has naming rules, which are more complicated than that. I updated the comments to make it clear.

```diff
diff --git a/src/lua/xml.c b/src/lua/xml.c
index fa4ad4035ff2..0eeba3500223 100644
--- a/src/lua/xml.c
+++ b/src/lua/xml.c
@@ -97,6 +97,7 @@ luaT_xml_decode(struct lua_State *L)
 		 */
 		case START_TAG: {
 			assert(tag_name != NULL);
+			/* TODO: Check names according to XML standard. */
 			if (isalnum(*s))
 				break; /* skip the tag name */
 			if ((!isspace(*s) && *s != '/' && *s != '>') ||
@@ -105,7 +106,6 @@ luaT_xml_decode(struct lua_State *L)
 			/* Stack: parent table, ... */
 			assert(lua_gettop(L) >= top + 1);
 			assert(lua_istable(L, -1));
-			/* TODO: Check the tag name. */
 			lua_pushlstring(L, tag_name, s - tag_name);
 			/*
 			 * Check that there's no attributes with the same name
@@ -198,6 +198,7 @@ luaT_xml_decode(struct lua_State *L)
 		 */
 		case ATTR_NAME:
 			assert(attr_name != NULL);
+			/* TODO: Check names according to XML standard. */
 			if (isalnum(*s))
 				break; /* skip the attribute name */
 			if ((!isspace(*s) && *s != '=') ||
@@ -205,7 +206,6 @@ luaT_xml_decode(struct lua_State *L)
 				return ERROR("invalid token");
 			/* Stack: attr table, tag name, parent table, ... */
 			assert(lua_gettop(L) >= top + 3);
-			/* TODO: Check the attribute name. */
 			lua_pushlstring(L, attr_name, s - attr_name);
 			/* Check that there's no duplicate attributes. */
 			lua_pushvalue(L, -1); /* attribute name */
diff --git a/src/lua/xml.h b/src/lua/xml.h
index d74ddba618cc..5e7ba706afa1 100644
--- a/src/lua/xml.h
+++ b/src/lua/xml.h
@@ -46,7 +46,10 @@ struct lua_State;
  * Limitations:
  *  - Element values, such as <section>value</section>, aren't supported.
  *  - Escape sequences in attribute values aren't supported.
- *  - Tag and attribute names aren't checked.
+ *  - Tag and attribute names aren't checked according to the XML standard.
+ *    The parser allows only digits and letters while in XML a name may also
+ *    contain dots, dashes, and underscores, and must start with a letter or
+ *    an underscore.
  */
 int
 luaT_xml_decode(struct lua_State *L);
```

</details>

src/lua/xml.c
locker: Added a comment:

```diff
diff --git a/src/lua/xml.c b/src/lua/xml.c
index 0eeba3500223..bc5a516b585a 100644
--- a/src/lua/xml.c
+++ b/src/lua/xml.c
@@ -86,8 +86,14 @@ luaT_xml_decode(struct lua_State *L)
 				state = START_TAG;
 				continue;
 			}
-			if (lua_gettop(L) == top + 1)
+			if (lua_gettop(L) == top + 1) {
+				/*
+				 * The stack only contains the resulting
+				 * document table, which means there's no
+				 * start tag matching this end tag.
+				 */
 				return ERROR("invalid token");
+			}
 			tag_name = s + 1;
 			state = END_TAG;
 			break; /* skip '/' */
```

</details>

Totktonada:
> > The main one is whether we should report `{used = 0, size = 0}` when the statistics is not available or should raise an error/drop the function from the module.
> 
> Discussed with @alyapunov. Agreed that it's best to return `{used = 0, size = 0}` if `malloc_info()` isn't supported because this way it's easier to use in Lua (less checks). Zero memory usage clearly indicates that something's wrong. For more information one can check `box.internal.malloc.info()`. In case of overridden malloc (e.g. by ASAN) it'll return a table filled with zeros. In case `malloc_info()` isn't supported by the system, it'll return an empty table. I'll rework this patch accordingly.

You said: zeros indicate that something is wrong. Overriden malloc is indicated this way. But lack of support on given OS isn't. I don't see a reason for the second way to indicate a problem. How about zeros on any problem?

locker:
> You said: zeros indicate that something is wrong. Overriden malloc is indicated this way. But lack of support on given OS isn't. I don't see a reason for the second way to indicate a problem. How about zeros on any problem?

There are two functions: `box.malloc.info()` (documented) and `box.malloc.internal.info()` (undocumented).

`box.malloc.internal.info()` returns decoded output of `malloc_info()` if the function is available. If `malloc` is overridden, `malloc_info()` is still available, but it may be not implemented properly by the allocator (at least, ASAN just returns all zeros). If `malloc_info()` is unavailable, this function will return an empty Lua table (because a table returned by `malloc_info()` is too complicated to forge).

`box.malloc.info()` will be implemented on all platforms in the same way: it will call `box.malloc.internal.info()` and try to sum memory usage reported by it. If any fields are absent or the `malloc_info()` version is unknown, it'll assume they are zeros, which means that even in case `malloc_info()` if not implemented by the system, it'll still return `{size=0, used=0}`.

**TLDR** The *documented* function `box.malloc.info()` will always return a table `{size=<number>, used=<number>}`, no matter whether `malloc_info()` is available or not. If it can't get memory usage info, it'll return 0 for both counters. A developer can figure out what's wrong by looking at `box.malloc.internal.info()` output.

Totktonada:
> **TLDR** The _documented_ function `box.malloc.info()` will always return a table `{size=<number>, used=<number>}`, no matter whether `malloc_info()` is available or not. If it can't get memory usage info, it'll return 0 for both counters. A developer can figure out what's wrong by looking at `box.malloc.internal.info()` output.

Nice! Sorry, I misread some parts of your initial message about the internal function as if it would be about the public one.

src/box/lua/slab.cc
locker: If the `malloc` table is absent, then the version won't be set and we'll fail below. I rewritten to make the check more obvious. Also, instead of an error we now return `{size = 0, used = 0}` as agreed.

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index fc833516f9b4..0006a59f02d1 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -310,10 +317,16 @@ lbox_malloc_internal_info(struct lua_State *L)
  *   }
  *
  * (all numbers are in bytes).
+ *
+ * The information is retrieved with malloc_info(). If it isn't supported by
+ * the system or its format is unknown, {size = 0, used = 0} is returned.
  */
 static int
 lbox_malloc_info(struct lua_State *L)
 {
+	int version = 0;
+	uint64_t total = 0;
+	uint64_t available = 0;
 	lbox_malloc_internal_info(L);
 	/*
 	 * The XML document name returned by malloc_info is expected to be
@@ -321,33 +334,31 @@ lbox_malloc_info(struct lua_State *L)
 	 */
 	assert(lua_istable(L, -1));
 	lua_getfield(L, -1, "malloc");
-	if (lua_istable(L, -1))
-		lua_rawgeti(L, -1, 1);
+	if (!lua_istable(L, -1))
+		goto out;
+	lua_rawgeti(L, -1, 1);
+	if (!lua_istable(L, -1))
+		goto out;
 	/*
 	 * First, check the malloc_info version. The only known version is 1.
 	 * It's pointless to proceed if the version is different.
 	 */
-	int version = 0;
-	if (lua_istable(L, -1)) {
-		lua_getfield(L, -1, "version");
-		version = lua_tonumber(L, -1);
-		lua_pop(L, 1);
-	}
+	lua_getfield(L, -1, "version");
+	version = lua_tonumber(L, -1);
+	lua_pop(L, 1);
 	if (version != 1)
-		return luaL_error(L, "unknown malloc info format");
+		goto out;
 	/*
 	 * Extract the size of used memory. Even though the document version is
 	 * valid, we still need to be careful accessing it.
 	 */
-	uint64_t total = 0;
-	uint64_t available = 0;
 	assert(lua_istable(L, -1));
 	lua_getfield(L, -1, "system");
 	if (lua_istable(L, -1)) {
@@ -385,6 +396,7 @@ lbox_malloc_info(struct lua_State *L)
 		}
 	}
 	lua_pop(L, 1);
+out:
 	/* Return memory usage information. */
 	lua_newtable(L);
 	luaL_pushuint64(L, total);
```

</details>

src/box/lua/slab.cc
locker: As agreed above, now `box.malloc.info()` always exists and returns `{used = 0, size = 0}` if `malloc_info()` is unavailable or its format is unknown. Full diff below.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 669d4760aed5..9b93bb6939af 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,11 +160,7 @@ check_function_exists(setproctitle HAVE_SETPROCTITLE)
 check_function_exists(setprogname HAVE_SETPROGNAME)
 check_function_exists(getprogname HAVE_GETPROGNAME)
 
-# If ASAN is enabled, malloc_info() exists but it is not implemented
-# (all counters in the returned document are set to zeros).
-if(NOT ENABLE_ASAN)
-    check_symbol_exists(malloc_info malloc.h HAVE_MALLOC_INFO)
-endif()
+check_symbol_exists(malloc_info malloc.h HAVE_MALLOC_INFO)
 
 #
 # Enable 'make tags' target.
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index fc833516f9b4..0006a59f02d1 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -275,15 +275,18 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
 	return 0;
 }
 
-#if defined(HAVE_MALLOC_INFO)
 /*
  * Decodes and returns the XML document returned by malloc_info() as is.
+ *
  * This is an internal function that isn't supposed to be called by users,
  * but it may be useful for developers.
+ *
+ * Returns an empty table if malloc_info() isn't supported by the system.
  */
 static int
 lbox_malloc_internal_info(struct lua_State *L)
 {
+#if defined(HAVE_MALLOC_INFO)
 	char *buf;
 	size_t buf_size;
 	FILE *fp = open_memstream(&buf, &buf_size);
@@ -299,6 +302,10 @@ lbox_malloc_internal_info(struct lua_State *L)
 	lua_pushlstring(L, buf, buf_size);
 	free(buf);
 	return luaT_xml_decode(L);
+#else /* !defined(HAVE_MALLOC_INFO) */
+	lua_newtable(L);
+	return 1;
+#endif /* defined(HAVE_MALLOC_INFO) */
 }
 
 /*
@@ -310,10 +317,16 @@ lbox_malloc_internal_info(struct lua_State *L)
  *   }
  *
  * (all numbers are in bytes).
+ *
+ * The information is retrieved with malloc_info(). If it isn't supported by
+ * the system or its format is unknown, {size = 0, used = 0} is returned.
  */
 static int
 lbox_malloc_info(struct lua_State *L)
 {
+	int version = 0;
+	uint64_t total = 0;
+	uint64_t available = 0;
 	lbox_malloc_internal_info(L);
 	/*
 	 * The XML document name returned by malloc_info is expected to be
@@ -321,33 +334,31 @@ lbox_malloc_info(struct lua_State *L)
 	 */
 	assert(lua_istable(L, -1));
 	lua_getfield(L, -1, "malloc");
-	if (lua_istable(L, -1))
-		lua_rawgeti(L, -1, 1);
+	if (!lua_istable(L, -1))
+		goto out;
+	lua_rawgeti(L, -1, 1);
+	if (!lua_istable(L, -1))
+		goto out;
 	/*
 	 * First, check the malloc_info version. The only known version is 1.
 	 * It's pointless to proceed if the version is different.
 	 */
-	int version = 0;
-	if (lua_istable(L, -1)) {
-		lua_getfield(L, -1, "version");
-		version = lua_tonumber(L, -1);
-		lua_pop(L, 1);
-	}
+	lua_getfield(L, -1, "version");
+	version = lua_tonumber(L, -1);
+	lua_pop(L, 1);
 	if (version != 1)
-		return luaL_error(L, "unknown malloc info format");
+		goto out;
 	/*
 	 * Extract the size of used memory. Even though the document version is
 	 * valid, we still need to be careful accessing it.
 	 */
-	uint64_t total = 0;
-	uint64_t available = 0;
 	assert(lua_istable(L, -1));
 	lua_getfield(L, -1, "system");
 	if (lua_istable(L, -1)) {
 		lua_pushnil(L);
 		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
 			if (!lua_istable(L, -1))
-				continue;
+				continue; /* skip XML attributes */
 			lua_getfield(L, -1, "type");
 			const char *key = lua_tostring(L, -1);
 			lua_pop(L, 1);
@@ -367,7 +378,7 @@ lbox_malloc_info(struct lua_State *L)
 		lua_pushnil(L);
 		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
 			if (!lua_istable(L, -1))
-				continue;
+				continue; /* skip XML attributes */
 			lua_getfield(L, -1, "type");
 			const char *key = lua_tostring(L, -1);
 			lua_pop(L, 1);
@@ -385,6 +396,7 @@ lbox_malloc_info(struct lua_State *L)
 		}
 	}
 	lua_pop(L, 1);
+out:
 	/* Return memory usage information. */
 	lua_newtable(L);
 	luaL_pushuint64(L, total);
@@ -393,7 +405,6 @@ lbox_malloc_info(struct lua_State *L)
 	lua_setfield(L, -2, "used");
 	return 1;
 }
-#endif /* defined(HAVE_MALLOC_INFO) */
 
 /** Initialize box.slab package. */
 void
@@ -429,20 +440,16 @@ box_lua_slab_init(struct lua_State *L)
 	lua_pushstring(L, "malloc");
 	lua_newtable(L);
 
-#if defined(HAVE_MALLOC_INFO)
 	lua_pushstring(L, "info");
 	lua_pushcfunction(L, lbox_malloc_info);
 	lua_settable(L, -3);
-#endif /* defined(HAVE_MALLOC_INFO) */
 
 	lua_pushstring(L, "internal");
 	lua_newtable(L);
 
-#if defined(HAVE_MALLOC_INFO)
 	lua_pushstring(L, "info");
 	lua_pushcfunction(L, lbox_malloc_internal_info);
 	lua_settable(L, -3);
-#endif /* defined(HAVE_MALLOC_INFO) */
 
 	lua_settable(L, -3); /* box.malloc.internal */
 	lua_settable(L, -3); /* box.malloc */
diff --git a/test/app-luatest/gh_7311_malloc_info_test.lua b/test/app-luatest/gh_7311_malloc_info_test.lua
index eea3c0c63ab6..847bc6cca67d 100644
--- a/test/app-luatest/gh_7311_malloc_info_test.lua
+++ b/test/app-luatest/gh_7311_malloc_info_test.lua
@@ -15,8 +15,16 @@ local ALLOC_SIZE = 10 * 1000 * 1000
 local MARGIN = ALLOC_SIZE * 0.05
 
 local function is_supported()
-    return jit.os == 'Linux' and
-           not tarantool.build.flags:match('-fsanitize=[%a,]*address')
+    -- malloc_info() is a GNU extension available only on Linux.
+    if jit.os ~= 'Linux' then
+        return false
+    end
+    -- If ASAN is enabled, malloc_info() exists but it is not implemented
+    -- (all counters in the returned document are set to zeros).
+    if tarantool.build.flags:match('-fsanitize=[%a,]*address') then
+        return false
+    end
+    return true
 end
 
 local function skip_if_supported()
@@ -30,38 +38,27 @@ end
 local function check_malloc_info(info)
     t.assert_ge(info.used, 0)
     t.assert_ge(info.size, info.used)
+    -- It's totally up to the malloc implementation whether to release system
+    -- memory immediately on free() or keep it for future allocations so we can
+    -- only check that the total system memory usage doesn't exceed the actual
+    -- memory usage by more than the size of memory allocated in this test.
+    -- Currently, there are two test cases that allocate ALLOC_SIZE each.
     t.assert_le(info.size, info.used + 2 * ALLOC_SIZE)
 end
 
 g.test_unsupported = function()
     skip_if_supported()
 
-    t.assert_equals(box.malloc, {internal = {}})
+    t.assert_equals(box.malloc.info(), {size = 0, used = 0})
 end
 
 g.test_malloc_info = function()
-    skip_if_unsupported()
-
     t.assert_type(box.malloc, 'table')
     t.assert_type(box.malloc.info, 'function')
+    t.assert_type(box.malloc.info(), 'table')
     t.assert_type(box.malloc.internal, 'table')
     t.assert_type(box.malloc.internal.info, 'function')
-
-    local info = box.malloc.info()
-    t.assert_type(info, 'table')
-    t.assert_type(info.used, 'number')
-    t.assert_type(info.size, 'number')
-    info.used = nil
-    info.size = nil
-    t.assert_equals(info, {})
-
-    info = box.malloc.internal.info()
-    t.assert_type(info, 'table')
-    t.assert_type(info.malloc, 'table')
-    t.assert_type(info.malloc[1], 'table')
-    t.assert_equals(info.malloc[1].version, '1')
-    info.malloc = nil
-    t.assert_equals(info, {})
+    t.assert_type(box.malloc.internal.info(), 'table')
 end
 
 g.test_malloc_small = function()
```

</details>

src/box/lua/slab.cc
locker: Both XML attributes and sub-elements are stored at the same level. This condition skips XML attributes. I added a comment clarifying this.

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index fc833516f9b4..0006a59f02d1 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -345,33 +358,31 @@ lbox_malloc_info(struct lua_State *L)
 		lua_pushnil(L);
 		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
 			if (!lua_istable(L, -1))
-				continue;
+				continue; /* skip XML attributes */
 			lua_getfield(L, -1, "type");
 			const char *key = lua_tostring(L, -1);
 			lua_pop(L, 1);
@@ -367,7 +378,7 @@ lbox_malloc_info(struct lua_State *L)
 		lua_pushnil(L);
 		for (; lua_next(L, -2) != 0; lua_pop(L, 1)) {
 			if (!lua_istable(L, -1))
-				continue;
+				continue; /* skip XML attributes */
 			lua_getfield(L, -1, "type");
 			const char *key = lua_tostring(L, -1);
 			lua_pop(L, 1);
```

</details>

test/app-luatest/gh_7311_malloc_info_test.lua
locker: Done:

```diff
diff --git a/test/app-luatest/gh_7311_malloc_info_test.lua b/test/app-luatest/gh_7311_malloc_info_test.lua
index eea3c0c63ab6..847bc6cca67d 100644
--- a/test/app-luatest/gh_7311_malloc_info_test.lua
+++ b/test/app-luatest/gh_7311_malloc_info_test.lua
@@ -30,38 +38,27 @@ end
 local function check_malloc_info(info)
     t.assert_ge(info.used, 0)
     t.assert_ge(info.size, info.used)
+    -- It's totally up to the malloc implementation whether to release system
+    -- memory immediately on free() or keep it for future allocations so we can
+    -- only check that the total system memory usage doesn't exceed the actual
+    -- memory usage by more than the size of memory allocated in this test.
+    -- Currently, there are two test cases that allocate ALLOC_SIZE each.
     t.assert_le(info.size, info.used + 2 * ALLOC_SIZE)
 end
 
```

src/box/lua/slab.cc
locker: Now we report zero memory usage in this case, see https://github.com/tarantool/tarantool/pull/8542#discussion_r1161535442.

test/app-luatest/gh_7311_malloc_info_test.lua
locker: After f2f discussion changed to:

```diff
diff --git a/test/app-luatest/gh_7311_malloc_info_test.lua b/test/app-luatest/gh_7311_malloc_info_test.lua
index 3ff33f93b35c..896b70aa8c2d 100644
--- a/test/app-luatest/gh_7311_malloc_info_test.lua
+++ b/test/app-luatest/gh_7311_malloc_info_test.lua
@@ -38,12 +38,14 @@ end
 local function check_malloc_info(info)
     t.assert_ge(info.used, 0)
     t.assert_ge(info.size, info.used)
-    -- It's totally up to the malloc implementation whether to release system
-    -- memory immediately on free() or keep it for future allocations so we can
-    -- only check that the total system memory usage doesn't exceed the actual
-    -- memory usage by more than the size of memory allocated in this test.
-    -- Currently, there are two test cases that allocate ALLOC_SIZE each.
-    t.assert_le(info.size, info.used + 2 * ALLOC_SIZE)
+    -- It's totally up to the malloc implementation whether to release memory
+    -- to the system on free() immediately or keep it for future allocations.
+    -- Still, it's reasonable to assume that a repetitive allocation and
+    -- freeing of an object of the same size won't result in growing total
+    -- memory usage infinitely. So we check that the system memory usage never
+    -- exceeds the allocated memory usage by more than the test allocation size
+    -- plus 10% overhead for internal housekeeping and fragmentation.
+    t.assert_le(info.size, info.used + 1.1 * ALLOC_SIZE)
 end
 
 g.test_malloc_info = function()
```

src/box/lua/slab.cc
```diff
@@ -265,6 +275,137 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
...
+	char *buf;
+	size_t buf_size;
+	FILE *fp = open_memstream(&buf, &buf_size);
+	if (fp == NULL) {
+		diag_set(SystemError, "failed to open memory stream");
+		return luaT_error(L);
+	}
+	if (malloc_info(0, fp) != 0) {
+		diag_set(SystemError, "failed to get malloc info");
+		return luaT_error(L);
```
alyapunov, line 299: You should `fclose` and `free` before returning.

test/app-luatest/gh_7311_malloc_info_test.lua
Totktonada: This explanation works for me, thanks!

src/box/lua/slab.cc
```diff
@@ -265,6 +275,137 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
...
+ * The information is retrieved with malloc_info(). If it isn't supported by
+ * the system or its format is unknown, {size = 0, used = 0} is returned.
+ */
+static int
+lbox_malloc_info(struct lua_State *L)
+{
+	int version = 0;
+	uint64_t total = 0;
+	uint64_t available = 0;
+	lbox_malloc_internal_info(L);
```
Totktonada, line 330: Shouldn't it be `pcall`'ed?

locker: `box.malloc.internal.info()` may fail because of:
1. `open_memstream`. The only reason I can think of why it may fail is OOM, but in this case tarantool is likely to panic anyway (because we use `xmalloc`).
2. `malloc_info`. According to the manual, this function may fail only on invalid arguments.
3. `luaT_xml_decode`. If `malloc_info` returns an invalid XML document, this will raise an error, but this means that something is seriously broken.

Neither of the three cases should normally happen. Of course, `malloc_info` may be overridden by a third-party allocator, but if it's broken it's better to raise an error instead of silencing this. I added a comment:

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index 0006a59f02d1..63c7f4c78e2c 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -327,6 +327,12 @@ lbox_malloc_info(struct lua_State *L)
 	int version = 0;
 	uint64_t total = 0;
 	uint64_t available = 0;
+	/*
+	 * box.malloc.internal.info() may raise only if something is seriously
+	 * wrong, for example, we have a bug in the XML parser or malloc_info
+	 * is broken. Either way, it's better to draw the attention by throwing
+	 * an error to the user.
+	 */
 	lbox_malloc_internal_info(L);
 	/*
 	 * The XML document name returned by malloc_info is expected to be
```

src/box/lua/slab.cc
locker: Sure. Fixed.

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index 63c7f4c78e2c..9d9e5e92e1c5 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -287,8 +287,8 @@ static int
 lbox_malloc_internal_info(struct lua_State *L)
 {
 #if defined(HAVE_MALLOC_INFO)
-	char *buf;
-	size_t buf_size;
+	char *buf = NULL;
+	size_t buf_size = 0;
 	FILE *fp = open_memstream(&buf, &buf_size);
 	if (fp == NULL) {
 		diag_set(SystemError, "failed to open memory stream");
@@ -296,6 +296,8 @@ lbox_malloc_internal_info(struct lua_State *L)
 	}
 	if (malloc_info(0, fp) != 0) {
 		diag_set(SystemError, "failed to get malloc info");
+		fclose(fp);
+		free(buf);
 		return luaT_error(L);
 	}
 	fclose(fp);
```

src/box/lua/slab.cc
Totktonada: Better for whom? We already discussed that a user unlikely expect statistics and logging to break a business logic. Your proposal was to report zeros from the public function and report all the details from the internal function. I would follow it consistently, not only for some kinds of an unexpected result.

If it really matters to show the errors from the public function, how about adding them to a field in the result (and keep `size` and `used` as zeros)?

```lua
tarantool> box.malloc.info()
- used: 0
  size: 0
  error: "slab.cc:298: failed to get malloc info"
```

locker: Fair enough. Let's call `box.malloc.internal.info()` with `pcall` then. As for adding the `error` field to the stats table, I don't think it's necessary, because a user application collecting stats would silently ignore it anyway.

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index 9d9e5e92e1c5..2e0ba9e680ef 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -282,6 +282,7 @@ lbox_slab_check(MAYBE_UNUSED struct lua_State *L)
  * but it may be useful for developers.
  *
  * Returns an empty table if malloc_info() isn't supported by the system.
+ * Raises a Lua error if it fails to retrieve or parse malloc_info() output.
  */
 static int
 lbox_malloc_internal_info(struct lua_State *L)
@@ -322,6 +323,8 @@ lbox_malloc_internal_info(struct lua_State *L)
  *
  * The information is retrieved with malloc_info(). If it isn't supported by
  * the system or its format is unknown, {size = 0, used = 0} is returned.
+ *
+ * This function never raises.
  */
 static int
 lbox_malloc_info(struct lua_State *L)
@@ -329,13 +332,9 @@ lbox_malloc_info(struct lua_State *L)
 	int version = 0;
 	uint64_t total = 0;
 	uint64_t available = 0;
-	/*
-	 * box.malloc.internal.info() may raise only if something is seriously
-	 * wrong, for example, we have a bug in the XML parser or malloc_info
-	 * is broken. Either way, it's better to draw the attention by throwing
-	 * an error to the user.
-	 */
-	lbox_malloc_internal_info(L);
+	lua_pushcfunction(L, lbox_malloc_internal_info);
+	if (luaT_call(L, 0, 1) != 0)
+		goto out;
 	/*
 	 * The XML document name returned by malloc_info is expected to be
 	 * "malloc" so the document content should be in malloc[1].
```

alyapunov (approved):

PR #8550, opened 2023-04-11
box: allow setting box.cfg values via env vars

### LINKED ISSUE #8051: `log_modules` option can not be set thru env var on box.cfg{}

### LINKED ISSUE #8494: `metrics` option can not be set thru env var on box.cfg{}
See also https://github.com/tarantool/tarantool/issues/8051

After https://github.com/tarantool/tarantool/issues/7725, the new option `metrics` were introduced. `metrics` is a key-value `table` with fields that could be `string`, array `table` with strings and key-value `table` with `string` values:
```lua
metrics = {
    include = 'all',
    exclude = {'vinyl'},
    labels = { alias = 'mystorage' }
```
(See also https://www.tarantool.io/en/doc/latest/book/monitoring/api_reference/#lua-function.metrics.cfg).

For now, only array table values are supported through environment, see TT_LISTEN (https://github.com/tarantool/doc/issues/2066) and current implementation: https://github.com/tarantool/tarantool/blob/e987427f6befb00cdbcaffacffd74f5275ab2fce/src/box/lua/load_cfg.lua#L1142-L1148

Possible solutions were discussed in https://github.com/tarantool/tarantool/issues/8051 . It is possible that there would be a solution to satisfy both tickets, or it could be two different solutions.

Lord-KA:
All box.cfg options values used to be plain values or arrays. Now some options contain key-value or nested tables.

This patch allows all such options to be set through environment variables too.

Closes #8494
Closes #8051

@TarantoolBot document
Title: Set box.cfg options via env vars

This patch implements a way to set tables as box.cfg options value in two different ways:
* as plain key-value table:
```
bash> export TT_LOG_MODULES=aaa=info,bbb=error && ./src/tarantool
Tarantool 2.10.0-beta1-1977-g2970bd57a
tarantool> box.cfg{}
...
2023-04-11 07:22:10.951 [219020] main/103/interactive/box.load_cfg I> set 'log_modules' configuration option to {"aaa":"info","bbb":"error"}
---
...

tarantool> box.cfg.log_modules.aaa
---
- info
...

tarantool> box.cfg.log_modules.bbb
---
- error
...
```

* as a table in json encoding (important: don't forget to put env var value into single quotes):
```
bash> export TT_METRICS='{"labels":{"alias":"mystorage"},"include":"all","exclude":["vinyl"]}' && ./src/tarantool
Tarantool 2.10.0-beta1-1977-g2970bd57a
tarantool> box.cfg{}
...
2023-04-11 07:26:03.635 [219288] main/103/interactive/box.load_cfg I> set 'metrics' configuration option to {"exclude":["vinyl"],"include":"all","labels":{"alias":"mystorage"}}
---
...

tarantool> box.cfg.metrics.include
---
- all
...
```


changelogs/unreleased/gh-8051-set-box-cfg-thru-env.md
```diff
@@ -0,0 +1,4 @@
+## feature/box/cfg
```
Totktonada, line 1: > box: allow setting box.cfg values via env vars

> Title: Set box.cfg options via env vars

The titles are confusing. This ability was implemented in 1b330121ec7dafcc0fc6ab63279e0dc52f9f947c. Next, several mappings (non-array tables) were added as box.cfg option. The feature was not supported only for them.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,72 @@
+local popen = require('popen')
```
Totktonada, line 1: What doesn't work with `luatest.server` + `:exec()`?

changelogs/unreleased/gh-8051-set-box-cfg-thru-env.md
Lord-KA: Rephrased it to this:
`box: allow to set box.cfg table values via env var`
It is not perfect, but I don't see how to fit all specifics into 50 symbols.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Done.

andreyaksenov (approved):

src/box/lua/load_cfg.lua
```diff
@@ -1149,11 +1149,22 @@ local function get_option_from_env(option)
...
         local res = {}
         for i, v in ipairs(raw_value:split(',')) do
-            res[i] = tonumber(v) or v
+            local eq = v:find('=')
+            if (eq == nil) then
+                res[i] = tonumber(v) or v
+            else
+                local lhs = string.sub(v, 1, eq - 1)
+                local rhs = string.sub(v, eq + 1)
+                res[lhs] = tonumber(rhs) or rhs
```
Totktonada, line 1166: `TT_XXX=bar=baz,foo` would be accepted as `{bar = 'baz', [2] = 'foo'}`, but I think it shouldn't.

So we can split `raw_value:find('=')` and `raw_value:find(',')` cases.

```lua
elseif param_type:find('table') and raw_value:find('=') then
    <...>
    for _, v in ipairs(raw_value:split(',')) do
        local eq = v:find('=')
        if eq == nil then
            error('XXX')
        end
        <...>
    end
    <...>
elseif param_type:find('table') and raw_value:find(',') then
    <...>
```

Totktonada (approved):
LGTM.

I left a proposal for a bit more tight validation, but that's minor.

src/box/lua/load_cfg.lua
```diff
@@ -1149,11 +1149,22 @@ local function get_option_from_env(option)
...
+                                     raw_value:startswith('[')) then
+        return json.decode(raw_value)
+    elseif param_type:find('table') and (raw_value:find(',') or
+                                         raw_value:find('=')) then
         assert(not param_type:find('boolean'))
         local res = {}
         for i, v in ipairs(raw_value:split(',')) do
-            res[i] = tonumber(v) or v
+            local eq = v:find('=')
+            if (eq == nil) then
```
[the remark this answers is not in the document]
Lord-KA: Fixed.

src/box/lua/load_cfg.lua
Lord-KA: Done:

```diff
+            local eq = v:find('=')
+            if eq == nil then
+                error(err_msg_fmt:format(env_var_name, option,
+                                         'in `key=value` or `value` format'))
+            end
+            local lhs = string.sub(v, 1, eq - 1)
+            local rhs = string.sub(v, eq + 1)
+            res[lhs] = tonumber(rhs) or rhs
```

src/box/lua/load_cfg.lua
```diff
@@ -1149,7 +1149,24 @@ local function get_option_from_env(option)
...
+    if param_type:find('table') and (raw_value:startswith('{') or
+                                     raw_value:startswith('[')) then
+        return json.decode(raw_value)
+    elseif param_type:find('table') and raw_value:find('=') then
+        assert(not param_type:find('boolean'))
+        local res = {}
+        for _, v in ipairs(raw_value:split(',')) do
+            local eq = v:find('=')
+            if eq == nil then
+                error(err_msg_fmt:format(env_var_name, option,
```
alyapunov, line 1161: There must be a test case for this error.

src/box/lua/load_cfg.lua
```diff
@@ -1149,7 +1149,24 @@ local function get_option_from_env(option)
...
+                                         'in `key=value` or `value` format'))
+            end
+            local lhs = string.sub(v, 1, eq - 1)
```
alyapunov, line 1164: It seems that there must be an error if `lhs` is an empty string. And there must be a test case that checks it.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,23 @@
...
+    local env = {["TT_METRICS"] = '{"labels":{"alias":"gh_8051"},' ..
+                                  '"include":"all","exclude":["vinyl"]}',
+                 ["TT_LOG_MODULES"] = 'aaa=info,bbb=error'}
+
+    g.server = server:new{alias='default', env=env}
+    g.server:start()
+
+    g.server:exec(server_test_nested_table)
+
+    server:stop()
```
alyapunov, line 22: If the test above fails, the server will not be stopped, which is not graceful. We usually use 'before test' and 'after test' helpers, they are readable and safe.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,23 @@
...
+local g = t.group()
+
+local function server_test_nested_table()
+    t.assert_equals(box.cfg.metrics.labels.alias, 'gh_8051')
+
+    t.assert_equals(box.cfg.log_modules, {['aaa'] = 'info', ['bbb'] = 'error'})
+end
+
+g.test_nested_table = function()
+    local env = {["TT_METRICS"] = '{"labels":{"alias":"gh_8051"},' ..
```
alyapunov, line 13: Json options that starts with '[' must also be tested. I mean if I remove `or raw_value:startswith('[')` from the code above, all tests will still pass. All functionality must be tested.

alyapunov (changes requested):

alyapunov (changes requested):
Thanks for the patch! see 4 comments below.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Good point, thx. Fixed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: I am not able to create a sane check on json encoding that starts with `'['`, because for now we don't have many `box.cfg{}` options that take in keyless tables (arrays). I can't use `TT_LISTEN` as `luatest.server` ignores it. It seems like this feature uses all the same lines as one with ``{`` and we don't need to test json, so maybe ignore such test? Otherwise, please help me find an option that can obtain keeless tables as value.

src/box/lua/load_cfg.lua
Lord-KA: Done.

src/box/lua/load_cfg.lua
Lord-KA: It is quite hard to test error thrown from `box.cfg{}` in `luatest.server`, as it is being run during `server:start()` and not inside `server:exec()`. Generally, we can test it via `popen()`, but it would be quite ugly. You can see it in the latest test version.

alyapunov: Perhaps it would be ugly, but without test it would be more ugly. If a code is not tested at all - it can simply be dropped during refactoring. To be sure I discussed it with colleagues. The conclusion is that there must be a test.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
alyapunov: At least replication takes arrays. But it won't be simple to start a cluster. In any case `popen` will help with TT_LISTEN. We should cover with tests all the code we write.

alyapunov (changes requested):

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Totktonada: I guess it can be tested with `luatest.server` using an own instance file (`command = <...>`).

The default instance file is just convenient feature introduced less than two years ago (61bf298542420e6cd1496ffe35f1ec5dd84f5e04). Before that the developers have to write own instance file per test (but sometimes they were shared between tests). Of course, nobody is obligated to use the default instance file, which sets `listen` on its own (for convenience in the most cases).

src/box/lua/load_cfg.lua
Totktonada: How about `treegen`+`justrun`? See the example: https://github.com/tarantool/tarantool/blob/55295f5f1be769cb88ceb262d8817d2c9e34c32c/test/app-luatest/override_panic_test.lua.

Lord-KA: Added `popen` helper, it seems somewhat clean this way.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Implemented the test via `popen` helper.

changelogs/unreleased/gh-8051-set-box-cfg-thru-env.md
alyapunov, line 1: It seems that you added 00000000000000000000.snap by mistake.

alyapunov (changes requested):
Almost there!

changelogs/unreleased/gh-8051-set-box-cfg-thru-env.md
Lord-KA: Oh, fixed.

alyapunov (approved):

ligurio:
Added an optional reviewer Gosha Moiseev.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
+local server = require('luatest.server')
+local popen = require('popen')
+local clock = require('clock')
+
+local t = require('luatest')
+local g = t.group()
+
+g.after_all = function()
```
Totktonada, line 8: I guess it should be `after_each`, because the tests reassign `g.server`.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+local TARANTOOL_PATH = arg[-1]
+
+local function popen_run(command)
+    local cmd = TARANTOOL_PATH .. ' -i 2>&1'
+    local ph = popen.new({cmd}, {
+        shell = true,
+        setsid = true,
+        group_signal = true,
+        stdout = popen.opts.PIPE,
+        stderr = popen.opts.PIPE,
```
Totktonada, line 23: We neither write, nor read `stderr`, so I would set it to `popen.opts.DEVNULL`.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+        stdin = popen.opts.PIPE,
+    })
+    t.assert(ph, 'process is not up')
+
+    ph:write(command)
+
+    local output = '';
+    local time_quota = 2.0
+    local start_time = clock.monotonic();
```
Totktonada, line 32: Nit: Redundant semicolons.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+    while output:find('message') == nil
+        and clock.monotonic() - start_time < time_quota do
+        local data = ph:read({timeout = 1.0})
```
Totktonada, line 35: I'm not a fan of tight timeouts -- it is often lead to flakiness in some environments (small machine, many parallel jobs). We can leave it as is, but I would set 20/10 seconds instead of 2/1.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+        if data ~= nil then
+            output = output .. data
+        end
```
Totktonada, line 38: EOF is not handled?

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+    g.server = server:new{alias='json_table_curly_bracket', env=env}
+    g.server:start()
+
+    t.assert_equals(g.server:get_box_cfg().metrics.labels.alias, 'gh_8051')
+end
+
+g.test_json_table_square_bracket = function()
+    local res = popen_run([=[
+        os.setenv('TT_LISTEN', '["localhost:9999"]')
+        box.cfg{}
```
Totktonada, line 58: Here we can get 'address already in use'. Use a Unix socket or binding to zero port.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,93 @@
...
+
+    g.server = server:new{alias='plain_table', env=env}
+    g.server:start()
+
+    t.assert_equals(g.server:get_box_cfg().log_modules,
+                    {['aaa'] = 'info', ['bbb'] = 'error'})
+end
+
+g.test_format_error = function()
+
```
Totktonada, line 76: Nit: An empty line at start of a code block looks odd (here and in the next test case).

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Fixed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Removed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Fixed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Set time_quota to 10.0, it seems to be the standard for other popen tests.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Fixed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Done.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Fixed.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: What do you mean?

Totktonada: > Return:
>
> <...> empty string if EOF

https://www.tarantool.io/en/doc/latest/reference/reference_lua/popen/#popen-read

This case is not handled here AFAIS, so we'll spin in the loop till the time quota will be reached. However there is no sense to do so when we got EOF.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
```diff
@@ -0,0 +1,91 @@
...
+        stdin = popen.opts.PIPE,
+    })
+    t.assert(ph, 'process is not up')
+
+    ph:write(command)
+
+    local output = ''
+    local time_quota = 10.0
+    local start_time = clock.monotonic()
+    while output:find('message') == nil
```
Totktonada, line 33: Hmm. I don't get the `:find('message')` condition. Can you explain?

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: Done:
```
      while clock.monotonic() - start_time < time_quota do
          local chunk = ph:read({timeout = 1.0})
          if chunk == '' or chunk == nil then                                                                       
              -- EOF or error 
              break                                                                                                 
          end                                                                                                       
          output = output .. chunk                                                                                  
      end                                                                                                           
```

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Lord-KA: You are right, dropped it.

test/box-luatest/gh_8051_set_box_cfg_thru_env_test.lua
Totktonada: I would raise an error in case of a `:read()` error, but just return may work for this given test. Closing the discussion.

PR #8552, opened 2023-04-11
box: apply dynamic cfg even if option value is unchanged

### LINKED ISSUE #8551: `box.cfg` doesn't try to re-apply `listen` or `replication` options if the option value is the same
Reproduced with Tarantool 3.0.0-entrypoint-162-g22d7636b521f. Tarantool 2.10 and 2.11 are affected as well.

`box.cfg.listen` and `box.cfg.replication` aren't re-applied if the configuration value is the same.

```
tarantool> box.cfg{listen = 3301}
2023-04-11 15:14:41.011 [929508] main/103/interactive I> tx_binary: bound to 0.0.0.0:3301
2023-04-11 15:14:41.012 [929508] main/103/interactive/box.load_cfg I> set 'listen' configuration option to 3301
---
...

tarantool> box.cfg{listen = 3301}
---
...
```

Workaround: pass the same uri in a different way, for example, as a string:

```
tarantool> box.cfg{listen = 3301}
---
...

tarantool> box.cfg{listen = '3301'}
2023-04-11 15:17:18.168 [929508] main/103/interactive I> tx_binary: stopped
2023-04-11 15:17:18.168 [929508] main/103/interactive I> tx_binary: bound to 0.0.0.0:3301
2023-04-11 15:17:18.169 [929508] main/103/interactive/box.load_cfg I> set 'listen' configuration option to "3301"
---
...
```

This is confusing and sometimes may complicate operation. For example, if there's a uri parameter that points to a file (e.g. an SSL certificate), then after updating the file one can't just call `box.cfg` with the same uri to reload the file (https://github.com/tarantool/tarantool-ee/issues/432). Another example: to restart replication one has to reset the `replication` table first by calling `box.cfg{replication = {}}`.

We should apply all options passed to `box.cfg` unconditionally, even if the value is unchanged. It should be up to the underlying callback to decide whether it should do anything or not. For example, the replication configuration callback already doesn't restart replication for appliers that are up and running, see commit https://github.com/tarantool/tarantool/commit/5994892c2f2c34823356efe5d8560e31e9ad7c42, so this extra check in `load_cfg.lua` is pointless.

locker:
If a configuration option value passed to `box.cfg` is the same as the old one, the option handler isn't called. As a result, one can't restart the server IPROTO connection or replication by passing the same URI to `box.cfg`. This is inconvenient, for example:
 - To update an SSL key file without renaming it, one has to first pass an empty URI to `box.cfg.listen` and then reset it back.
 - To restart broken replication after fixing it manually (e.g. deleting a conflicting record on the replica), one has to first set `box.cfg.replication` to `{}` and then set it back instead of just calling `box.cfg{replication = box.cfg.replication}`.

There's no reason to skip an option update if the value is unchanged. It should be up to the option configuration callback.

The only callback that may actually fail on an attempt to set the same value is `box.cfg.memtx_memory`. This happens because the value is rounded up to a multiple of the quota unit size so we just need to fix the check accordingly.

Note that the `box.cfg.replication` configuration callback won't restart replication if it's up and running even if the URI list is rearranged thanks to commit 5994892c2f2c ("replication: fix replica disconnect upon reconfiguration"). This commit doesn't break this behavior - it just removes the `load_cfg.lua` code that would skip option update if the old and new option values are equal in Lua.

Needed for tarantool/tarantool-ee#432
Closes #8551

### DIFF
--- a/changelogs/unreleased/gh-8551-box-cfg-reload-fix.md
+++ b/changelogs/unreleased/gh-8551-box-cfg-reload-fix.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+ * `box.cfg.listen` and `box.cfg.replication` can now be reconfigured even if
+   the URI is unchanged (gh-8551).
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -910,11 +910,12 @@ local function reload_cfg(oldcfg, cfg)
     local module_keys = {}
     -- iterate over original table because prepare_cfg() may store NILs
     for key in pairs(cfg) do
-        if not compare_cfg(oldcfg[key], newcfg[key]) then
-            local name = option2module_name[key]
-            if name == nil then
+        local name = option2module_name[key]
+        if name == nil then
+            if not compare_cfg(oldcfg[key], newcfg[key]) then
                 box.error(box.error.RELOAD_CFG, key);
             end
+        else
             if module_keys[name] == nil then
                 module_keys[name] = {}
             end
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -1557,7 +1557,8 @@ memtx_engine_set_snap_io_rate_limit(struct memtx_engine *memtx, double limit)
 int
 memtx_engine_set_memory(struct memtx_engine *memtx, size_t size)
 {
-	if (size < quota_total(&memtx->quota)) {
+	if (DIV_ROUND_UP(size, QUOTA_UNIT_SIZE) <
+	    quota_total(&memtx->quota) / QUOTA_UNIT_SIZE) {
 		diag_set(ClientError, ER_CFG, "memtx_memory",
 			 "cannot decrease memory size at runtime");
 		return -1;
--- a/test/box-luatest/gh_8551_box_cfg_reload_test.lua
+++ b/test/box-luatest/gh_8551_box_cfg_reload_test.lua
@@ -1,0 +1,71 @@
+local server = require('luatest.server')
+local t = require('luatest')
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.master = server:new({alias = 'gh_8551_master'})
+    cg.master:start()
+    cg.master:exec(function()
+        box.schema.create_space('test')
+        box.space.test:create_index('primary')
+    end)
+    cg.replica = server:new({
+        alias = 'gh_8551_replica',
+        box_cfg = {
+            replication = server.build_listen_uri('gh_8551_master')
+        },
+    })
+    cg.replica:start()
+end)
+
+g.after_all(function(cg)
+    cg.master:drop()
+end)
+
+g.test_listen = function(cg)
+    cg.master:exec(function()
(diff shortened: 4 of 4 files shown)


Mons:
Don't do this, It's breaking behavior.

locker:
> Don't do this, It's breaking behavior.

What does it break? All tests pass. The replication configuration callback won't restart replication if it's up and running thanks to commit https://github.com/tarantool/tarantool/commit/5994892c2f2c34823356efe5d8560e31e9ad7c42.

Totktonada:
Even if it doesn't break things it sets the new mechanics and requires extra attention from callback authors to keep the old behavior. I have a proposal: https://github.com/tarantool/tarantool/issues/8551#issuecomment-1504204003.

changelogs/unreleased/gh-8551-box-cfg-reload-fix.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+ * `box.cfg.listen` and `box.cfg.replication` are now reconfigured even if
+   the URI is unchanged (gh-8551).
```
p7nov, line 4:
```suggestion
 * `box.cfg.listen` and `box.cfg.replication` can now be reconfigured
   without changing the URI (gh-8551).
```

p7nov (approved):
Please see my suggestion on the changelog wording.

unera:
what exactly can be broken?

Mons:
> what exactly can be broken?

Discussed in issue.
Confusing wording in description lead to misunderstanding.

If no active replication is restarted in case when nothing changed, then everything is ok.

changelogs/unreleased/gh-8551-box-cfg-reload-fix.md
locker: Applied the suggestion. Thanks.

nshy (approved):

kostja:
The client has a right to expect that if the network is OK and the hardware is running fine its queries will not fail. There is a whole class of applications for which it is good enough. Now you're saying you'll be dropping connections for no reason whatsoever, just because somebody reset an unrelated parameter.

A lot of applications are not idempotent. With interactive transactions it's even worse.

locker:
> The client has a right to expect that if the network is OK and the hardware is running fine its queries will not fail. There is a whole class of applications for which it is good enough. Now you're saying you'll be dropping connections for no reason whatsoever, just because somebody reset an unrelated parameter.

I didn't say that.

No active connections would be dropped on reconfiguration after this change, because neither `box.cfg.replication` nor `box.cfg.listen` callbacks drop active connections. With `box.cfg.listen` it's always been this way while `box.cfg.replication` was patched to behave this way in commit https://github.com/tarantool/tarantool/commit/5994892c2f2c34823356efe5d8560e31e9ad7c42.

Still, after this change reconfiguration of `box.cfg.replication` would wait for sync is complete, which may be an issue. At least, this breaks vshard tests.

locker:
https://github.com/tarantool/tarantool/issues/8551#issuecomment-1511633900

PR #8556, opened 2023-04-12
test: fix flaky gh_8121_memtx_mvcc_stream_test

Serpentian:
The test checks that replication stream doesn't get transaction conflict errors. However, currently it fails with the following error: "attempt to index field 's' (a nil value)", where 's' is a space, created at master during before_all phase.

The problem is the fact, that the test doesn't wait for replica to be synchronized with master before dropping the replication source. So, sometimes replica doesn't have any space 's' at the beginning of the test body exection.

Let's add waiting for vclocks synchronization before reconfiguring box.cfg.replication on replica.

Closes tarantool/tarantool-qa#311

NO_CHANGELOG=test
NO_DOC=test

### DIFF
--- a/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
+++ b/test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
@@ -29,10 +29,7 @@ g.before_all(function(cg)
         s:create_index('pk')
         s:insert{0}
     end, {cg.params.engine})
-    t.helpers.retrying({}, function()
-        cg.replica:assert_follows_upstream(cg.master:get_instance_id())
-    end)
-    cg.master:wait_for_downstream_to(cg.replica)
+    cg.replica:wait_for_vclock_of(cg.master)
     cg.replica:exec(function()
         box.cfg{replication = ""}
     end)


sergepetrenko:
@Serpentian. Yep, I always understood that comment one way, but I see now that it may be interpreted vaguely.
```lua
--- Wait until all own data is replicated and confirmed by the given server.
--
-- @tab server Server's object.
function Server:wait_for_downstream_to(server)
```

Still, I suppose "own" data refers to `self` object, not to `server`. Because the data is "replicated and confirmed by the given server". The "replicated and confirmed by" part wouldn't be there if the function did what you suggest. There might be no replication from replica back to master at all.

Besides there are only two uses of this function, and it seems in both cases authors used the function just like `wait_for_vclock_of`.
```
➜  sptnt git:(master) grep -rl 'wait_for_downstream_to' ./test                             
./test/replication-luatest/gh_7253_election_long_wal_write_test.lua
./test/replication-luatest/gh_8121_memtx_mvcc_replication_stream_test.lua
```

Serpentian:
Let's ask @ylobankov, as he've written that function. I don't really like the idea of waiting for vclock in `wait_for_downstream`.

sergepetrenko:
The function was actually added here: 544955109140e5dd5b956178fe47fd5ed3da2952 by @Gerold103. Let's ask him what was on his mind.

Serpentian:
Eventually we decided to change the behavior of the `wait_for_downstream_to` function, so the issue will be resolved with merging of https://github.com/tarantool/luatest/pull/301

PR #8559, opened 2023-04-12
[2.10] sql: fix integer overflow in built-in functions

ImeevMA:
This patch replaces the type for some int and uint32_t values with size_t to avoid problems with integer overflow.

Closes tarantool/security#119

NO_DOC=bugfix

(cherry picked from commit 60a187d7e113fb1c0768774d46215792cc0cef7e)


locker (approved):

locker:
@ImeevMA the fuzzing jobs fail for some reason. @ylobankov says it might be caused by this PR. Please take a look.

PR #8563, opened 2023-04-12
gdb: nice list walking

### LINKED ISSUE #8524: GDB tooling: nice and comfortable list walking
Currently the suggested way to walk the list is (mentioned in 'help tt-list'):

```
# Walk the engines (repeat the last command)
(gdb) set $i=0
(gdb) tt-list -f '$index==$i' -post '$i++' -- engines
```
But:
1. It looks ugly and unnatural
2. It iterates the requested list from the beginning for every step

Currently there is single command that handle lists -- `tt-list` and the method above was just a quick hack to get the desired feature using functional that is implemented already (however `-post` option was added specifically to make this "walk" possible).

The nature of `tt-list` command is 'walk all steps **internally** and display the **all the collected entries at once**'. The nature of walking is somewhat opposite as it requires to perform **single** step and display **single** entry. That's why emulating the one over another looks unnatural. 

Thus it is suggested to implement a new command (say, `tt-list-walk`) that has 2 modes:
1. Setup mode (with arguments identical to the arguments of `tt-list` command)
Specify the list and possibly additional "walk" configuration (like direction, predicate, entry fields filtering)
2. Single step mode (w/o arguments)
Do single step of walk and display the next (single) entry that satisfies the condition configured at setup mode.

```
# Walking sequence would look like (repeat the last command)
(gdb) tt-list-walk main_cord.alive
(gdb) tt-list-walk

# If only interested in the 'name' field of fibers with 'fid' greater than 110
(gdb) tt-list-walk -f name -p '$entry->fid > 110' -- main_cord.alive
(gdb) tt-list-walk
```

elhimov:
Introduced command 'tt-list-walk'

Changes in 'tt-list' options:
- dropped options -pre/post (previously used to implement walking
  functionality; no longer needed)
- option -filter was renamed to -predicate (so it can be specified with
  just -p while previously at least -fil need to be used to avoid
  ambiguity with -fields option)

Closes #8524

NO_DOC=gdb extension
NO_TEST=gdb extension
NO_CHANGELOG=gdb extension

### DIFF
--- a/tools/tarantool-gdb.py
+++ b/tools/tarantool-gdb.py
@@ -14,6 +14,7 @@ import re
 import sys
 
 if sys.version_info[0] == 2:
+    map = itertools.imap
     filter = itertools.ifilter
     zip = itertools.izip
 elif sys.version_info[0] == 3:
@@ -1578,6 +1579,7 @@ class ListLut(object):
         container = cls._containers_map.get(container_info.container_type.tag)
         return container.get(container_info.offset) if container is not None else None
 
+
 class Rlist(object):
     gdb_type = gdb.lookup_type('rlist')
     item_gdb_type = gdb_type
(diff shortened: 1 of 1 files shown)


tools/tarantool-gdb.py
```diff
@@ -2354,3 +2481,67 @@ def invoke(self, arg, from_tty):
             TtListPrinter.reset_config()
 
 TtListSelect()
+
+
+class TtListWalk(gdb.Command):
+    """
+tt-list-walk
+Display list entry LIST_EXP refers to or the entire list if LIST_EXP refers
+to the list head.
```
locker, line 2490: This command has an exactly the same description as `tt-list`. It's unclear to me, as a user, which one I should pick. Please elaborate on that in the command description: why we need two commands to walk over a list; which one is preferable in a particular scenario.

locker (approved):

Issue #4479, opened 2019-09-04
Consider support of space / index names in the binary protocol


Totktonada:
Currently tarantool adds an incremental schema id to responses, allows to set a client's schema id for requests and gives an error for a request with a non actual schema id. This allows connectors to determine the situation when a request was made against an old schema, fetch the new one, map space / index names to ids and then retry the request. This however complicates a connector code, especially if it support asynchronous requests: the many situations need to be handled, say, a schema can be updated during a schema fetch.

For ones who interesting: we had discussed some corner cases in https://github.com/tarantool/tarantool-java/pull/200

It seems logical to eliminate this complexity and allow to send requests with index / space names. We already allow to use field names for update/upsert operations (it is the part of #1261) in the binary protocol and they are part of a schema (AFAIU).

The main point here is that we have many connectors and an ability to make them easier is something valuable: at least it saves developers from stepping into possible mistakes with handling corner cases, performing synchronization and retrying of requests.

Are there any pitfalls in the proposed feature?

rybakit:
What will happen to the schema id once this is implemented?

Totktonada:
In the proposal I assumed that it will be supported anyway.

rybakit:
Apart from the reducing complexity, I think that this feature might be useful for short-living scripts that do not persist information across runs (this is especially true for PHP with its stateless execution model). So they don't waste resources by fetching schema to just make one or two requests and die.

CuriousGeorgiy:
Superseeded by #8146.

Issue #6996, opened 2022-04-05
Add more checks to tools/gen-release-notes


locker:
- Check that only known section names are used to avoid section naming mess:
  https://github.com/tarantool/tarantool/blob/a303b53beb1cd752816a7f1991235dab8a091569/changelogs/unreleased/gh-4861-net-box-return-raw.md?plain=1#L1

  https://github.com/tarantool/tarantool/blob/a303b53beb1cd752816a7f1991235dab8a091569/changelogs/unreleased/gh-6144-incorrect-timeout-error.md?plain=1#L1
- Check for indentation:
  - Bad alignment
    https://github.com/tarantool/tarantool/blob/a303b53beb1cd752816a7f1991235dab8a091569/changelogs/unreleased/gh-4415-introduce-decimal.md?plain=1#L1-L4
  - Extra space before *
    https://github.com/tarantool/tarantool/blob/a303b53beb1cd752816a7f1991235dab8a091569/changelogs/unreleased/add-check-for-user-input-of-the-number-of-iproto-threads.md?plain=1#L1-L3
- Check for long lines:
  https://github.com/tarantool/tarantool/blob/a303b53beb1cd752816a7f1991235dab8a091569/changelogs/unreleased/gh-5799-add-info-to-fiber-object.md?plain=1#L3

PR #8573, opened 2023-04-15
box: support space and index names in IPROTO requests

### LINKED ISSUE #8146: Accept names in IPROTO requests
Now in IPROTO requests a user has to specify ID of space (and ID of index for selects).
Let's add an alternative that would allow to specify name of the space (and name of index).
We could use the space key(s) in IPROTO body type or create new - we should investigate what is better,

CuriousGeorgiy:
Closes #8146


changelogs/unreleased/gh-8416-space-and-index-names-in-iproto-requests.md
```diff
@@ -0,0 +1,10 @@
+## feature/box
+
+* Added support for accepting IPROTO requests with space or index name instead 
+  of identifier (gh-8146).
```
Totktonada, line 4: BTW, it also closes #4479, I guess.

CuriousGeorgiy: I forgot about #4479, should I close #8146 as duplicate?

Totktonada: I guess that the fresh issue is tracked in some epics/boards and I would just close both by one patch to not do any mess in the management tools.

CuriousGeorgiy: Closed #4479 as duplicate.

changelogs/unreleased/gh-8146-space-and-index-names-in-iproto-requests.md
```diff
@@ -0,0 +1,10 @@
+## feature/box
+
+* Added support for accepting IPROTO requests with space or index name instead
```
andreyaksenov, line 3:
```
with space -> 
with a space
```

changelogs/unreleased/gh-8146-space-and-index-names-in-iproto-requests.md
```diff
@@ -0,0 +1,10 @@
...
+  of identifier (gh-8146).
+
+## feature/net.box
+
+* Added support for sending CRUD requests from connection objects with disabled
+  schema fetching by manually specifying space or index name or identifier
```
andreyaksenov, line 9:
```
specifying space or index name ->
specifying a space or index name
```

andreyaksenov (approved):
A couple of minor suggestions:

changelogs/unreleased/gh-8146-space-and-index-names-in-iproto-requests.md
CuriousGeorgiy: I guess, I should also change `instead of identifier -> instead of an identifier`. Fixed.

src/box/iproto.cc
```diff
@@ -2138,6 +2138,39 @@ tx_process_rollback(struct cmsg *m)
...
+ * identifier is not present in the request and we fail to get an identifier
+ * from the space name (or the space name is missing), we must return with an
+ * error, setting a missing request field diagnostic, just as it was done in
+ * `xrow_decode_dml` before space names in IPROTO requests were introduced.
+ */
+static int
+tx_resolve_space_and_index_name(struct request *dml)
+{
+	if (dml->space_id == 0 && dml->space_name != NULL) {
+		uint32_t space_id = box_space_id_by_name(dml->space_name,
```
locker, line 2154: Why do you not use `space_cache_find` here? It's more lightweight and should be faster than the public API function. In general, in the core code we should prefer internal function over public API functions.

src/box/iproto.cc
```diff
@@ -2138,6 +2138,39 @@ tx_process_rollback(struct cmsg *m)
...
+							 dml->space_name_len);
+		if (space_id != BOX_ID_NIL)
+			dml->space_id = space_id;
+	}
+	if (dml->space_id == 0) {
+		diag_set(ClientError, ER_MISSING_REQUEST_FIELD,
+			 iproto_key_name(IPROTO_SPACE_ID));
```
locker, line 2161: If `space_name` is present, but there's no such space, we should raise `ER_NO_SUCH_SPACE` error. Same comment for the `index_name`. Please fix and add a test.

Also, technically, if `space_id` is present but set to 0, we should return `ER_NO_SUCH_SPACE` error too instead of `ER_MISSING_REQUEST_FIELD`.

src/box/iproto.cc
```diff
@@ -2138,6 +2138,39 @@ tx_process_rollback(struct cmsg *m)
...
+		return -1;
+	}
+	if (dml->index_id == 0 && dml->index_name != NULL) {
```
locker, line 2164: Nit: If the user sets `index_id` to 0, which is a valid id, then it'll be ignored in case `index_name` is present. This isn't critical, but still contradicts the comment to `IPROTO_INDEX_NAME`.

Also, we don't need to resolve the index name for insert and replace requests. Again, not critical, but doesn't look good.

src/box/iproto_constants.h
```diff
@@ -179,6 +179,17 @@ enum iproto_key {
 	 * authentication method.
 	 */
 	IPROTO_AUTH_TYPE = 0x5b,
+	/**
+	 * Space name used instead of identifier (IPROTO_SPACE_ID) in CRUD
```
locker, line 183: Nit: Nowhere in the code we use the term "CRUD". We call these requests "DML". Using two terms for the same thing may look confusing. Better continue using "DML".

src/box/iproto_constants.h
```diff
@@ -179,6 +179,17 @@ enum iproto_key {
...
+	 * requests. Ignored when identifier is present.
+	 */
+	IPROTO_SPACE_NAME = 0x5c,
+	/**
+	 * Index name used instead of identifier (IPROTO_INDEX_ID) in
+	 * IPROTO_SELECT and IPROTO_DELETE requests. Ignored when identifier is
```
locker, line 189: Nit: It also may be used in update requests. Same note to the comment to `IPROTO_FEATURE_SPACE_AND_INDEX_NAMES`.

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
...
+                                            box.internal.new_tuple_format()}
+                        space.index = setmetatable({space = space,
+                                                    _remote = remote}, {
+                            __index = function(self, key)
+                                return setmetatable({id = key, name = key,
+                                                     space = self.space,
+                                                     _remote = self._remote},
+                                                    self._remote._index_mt)
+                            end})
+                        return setmetatable(space, self._remote._space_mt)
```
locker, line 409: AFAIR we agreed to use weak tables to avoid creation of a space and index objects each time they are accessed.

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
                     })
                 end
             end
+            if not opts.fetch_schema and features.space_and_index_names then
+                remote.space = setmetatable({_remote = remote}, {
+                    __index = function(self, key)
+                        local space = {id = key, name = key,
```
locker, line 397: `id` is supposed to be a number while `name` should be a string. I don't think we should set both to the given key. We should set either `id` or `name` depending on the key type. The same comment for index id/name.

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
...
+                                       _remote = self._remote,
+                                       _format_cdata =
+                                            box.internal.new_tuple_format()}
+                        space.index = setmetatable({space = space,
+                                                    _remote = remote}, {
```
locker, line 402: Why do you use `remote` here, but `self._remote` in the same function two lines above?

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
                     })
                 end
             end
+            if not opts.fetch_schema and features.space_and_index_names then
+                remote.space = setmetatable({_remote = remote}, {
+                    __index = function(self, key)
+                        local space = {id = key, name = key,
+                                       _remote = self._remote,
+                                       _format_cdata =
```
locker, line 399: `_format_cdata` is never used.

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
...
+                                            box.internal.new_tuple_format()}
+                        space.index = setmetatable({space = space,
```
locker, line 401: There shouldn't be such thing as `space.index.space` unless there's an index named "space".

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,24 @@ local function new_sm(uri, opts)
                     })
                 end
             end
+            if not opts.fetch_schema and features.space_and_index_names then
+                remote.space = setmetatable({_remote = remote}, {
```
locker, line 395: Theoretically, there may be a space named `_remote`, in which case it'll be masked by the private field. We could avoid that by binding `remote` to the `__index` function. However, this would require creation of a new `__index` function per each connection (which you already do). I'm not sure it's okay. We should ask @Totktonada.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
```diff
@@ -0,0 +1,86 @@
+local netbox = require('net.box')
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new({alias = 'dflt'})
```
locker, line 8: Nit: `alias` is optional - you may skip it, in which case it'll be set to "server".

src/box/lua/net_box.lua
CuriousGeorgiy: It is used in requests (see space and index object metatabels).

locker: Missed that, sorry.

src/box/lua/net_box.lua
CuriousGeorgiy: AFAIC, spaces with ’_’ prefix are reserved to be system spaces, so that cannot happen.

src/box/lua/net_box.lua
CuriousGeorgiy: I don't understand how you want to use weak tables here. I thought we wanted to make the `c.space` and `c.space.s.index` tables have weak values, so that the connection objects have weak references, but it doesn't help, since their metamethods are closures that capture connection objects.

src/box/lua/net_box.lua
locker: Theoretically, we can create a system space with this name (`_remote`) either in core or in some system module (e.g. vshard). We could fix this by using captured (bound) values in `__index`.

src/box/lua/net_box.lua
CuriousGeorgiy: Looks like the problem was caused by the typo I made that you mention two comments bellow (I used `remote` instead of `self._remote` and it created an upvalue). Now setting `__mode = 'v'` for `c.space` and `c.space.s.index` works as expected (except for the latter, the space field will always get garbage collected, which I don't think we want to happen).

> to avoid creation of a space and index objects each time they are accessed.

I still don't understand how you want to use weak tables here with this goal, could you please elaborate a bit or redirect me to some example?

src/box/lua/net_box.lua
CuriousGeorgiy: Then AFAIC we won't be able to make the connection a weak reference (see discussion above).

src/box/iproto.cc
CuriousGeorgiy: `space_cache_find` has an inconvenient interface: it only support NULL-terminated strings. Guess I will have to refactor it to allow arbitrary strings :slightly_frowning_face:

changelogs/unreleased/gh-8146-space-and-index-names-in-iproto-requests.md
```diff
@@ -0,0 +1,10 @@
+## feature/box
+
+* Added support for accepting IPROTO requests with a space or index name instead
+  of an identifier (gh-8146).
+
+## feature/net.box
+
+* Added support for sending CRUD requests from connection objects with disabled
+  schema fetching by manually specifying a space or index name or an identifier
+  (gh-8146).
```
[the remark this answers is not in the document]
CuriousGeorgiy: > I think it's better to merge these two changelog entries in one, because it's a single IPROTO feature. The built-in net.box connector must be up-to-date with the current IPROTO implementation anyway.

Done.

> I'd also put the entry under the `feature/core` category.

But IPROTO is part of box.

src/box/iproto.cc
CuriousGeorgiy: Fixed, now I use `space_by_name`.

src/box/iproto_constants.h
CuriousGeorgiy: Actually DQL (select, get) is not DML, but in terms of our codebase they are, which is confusing. Fixed.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
CuriousGeorgiy: Thanks, I didn't know about this! Fixed.

src/box/lua/net_box.lua
CuriousGeorgiy: Typo, fixed.

src/box/iproto_constants.h
CuriousGeorgiy: Thanks, I missed it, when writing the design document! Fixed.

src/box/iproto_constants.c
```diff
@@ -179,15 +181,15 @@ const char *iproto_type_strs[] =
...
+	bit(LIMIT) | bit(KEY),                                 /* SELECT */
+	bit(TUPLE),                                            /* INSERT */
+	bit(TUPLE),                                            /* REPLACE */
+	bit(KEY) | bit(TUPLE),                                 /* UPDATE */
+	bit(KEY),                                              /* DELETE */
 	0,                                                     /* CALL_16 */
 	0,                                                     /* AUTH */
 	0,                                                     /* EVAL */
-	bit(SPACE_ID) | bit(OPS) | bit(TUPLE),                 /* UPSERT */
+	bit(OPS) | bit(TUPLE),                                 /* UPSERT */
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed, as we discussed with @locker and @alyapunov: mark the `SPACE_ID` key present when encountering `SPACE_NAME`.

src/box/lua/net_box.lua
CuriousGeorgiy: Fixed.

src/box/lua/net_box.lua
CuriousGeorgiy: Fixed, now `space` and `index` table work as registers.

src/box/iproto.cc
CuriousGeorgiy: Fixed, added test cases for this.

src/box/iproto.cc
CuriousGeorgiy: > Nit: If the user sets index_id to 0, which is a valid id, then it'll be ignored in case index_name is present. This isn't critical, but still contradicts the comment to IPROTO_INDEX_NAME.

Fixed: now name is preferred over idenitifer. Added test cases for this and updated the design document.

> Also, we don't need to resolve the index name for insert and replace requests. Again, not critical, but doesn't look good.

Fixed, now name is ignored for these requests.

src/box/lua/net_box.lua
CuriousGeorgiy: Discussed with @locker and @alyapunov: the space and object tables shall be register tables (i.e., cache the values they are indexed with) with strong references — all necessary context stored as upvalues.

src/box/lua/net_box.lua
CuriousGeorgiy: Fixed.

src/box/lua/net_box.lua
CuriousGeorgiy: Actually, `nil` format is ok:
https://github.com/tarantool/tarantool/blob/a27d958d060ae2725ed8df6c1a4cc2a51d776233/src/box/lua/net_box.c#L2255-L2258

Dropped this field.

changelogs/unreleased/gh-8146-space-and-index-names-in-iproto-requests.md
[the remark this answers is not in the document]
locker: > But IPROTO is part of box.

I think that the box is a part of the core. Actually, there's no section "box" in the known section list:

https://github.com/tarantool/tarantool/blob/a27d958d060ae2725ed8df6c1a4cc2a51d776233/tools/gen-release-notes#L14-L17

Unfortunately, we don't have an automated tool to check the section name nor do we have a clear guideline about how to choose a section name so we use "box" now and then and nobody cares and you can continue using "box" if you like. (I filed a ticket to enable section name checking some time ago - see #6996.)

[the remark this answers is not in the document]
CuriousGeorgiy: AFAIC, core is stuff like fibers and other primitives build on — `box` is database stuff.

src/box/xrow.c
```diff
@@ -959,6 +959,15 @@ xrow_decode_dml(struct xrow_header *row, struct request *request,
 			request->after_tuple = value;
 			request->after_tuple_end = data;
 			break;
+		case IPROTO_SPACE_NAME:
+			key_map &= ~iproto_key_bit(IPROTO_SPACE_ID);
```
locker, line 963: This function is also used for decoding replication and recovery requests, where space names aren't accepted. I suggest to use the following approach insrtead:

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 22ed07ddd626..e93a876abf9b 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1661,6 +1661,9 @@ iproto_msg_decode(struct iproto_msg *msg, struct cmsg_hop **route)
 		assert(type < sizeof(iproto_thread->dml_route) /
 			      sizeof(*iproto_thread->dml_route));
 		*route = iproto_thread->dml_route[type];
+		uint64_t key_map = dml_request_key_map(xrow->type);
+		/* space_id is optional - one may use space_name instead */
+		key_map |= 1ULL << IPROTO_SPACE_NAME;
 		if (xrow_decode_dml(&msg->header, &msg->dml,
 				    dml_request_key_map(type)))
 			return -1;
diff --git a/src/box/xrow.c b/src/box/xrow.c
index 0e71abb30468..ce72b0fbcb0d 100644
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -960,7 +960,6 @@ error:
 			request->after_tuple_end = data;
 			break;
 		case IPROTO_SPACE_NAME:
-			key_map &= ~iproto_key_bit(IPROTO_SPACE_ID);
 			request->space_name =
 				mp_decode_str(&value, &request->space_name_len);
 			break;
@@ -973,6 +972,11 @@ error:
 		}
 	}
 done:
+	if ((key_map & (1ULL << IPROTO_SPACE_NAME)) != 0 &&
+	    request->space_name != NULL) {
+		/* The caller can accept space_name instead of space_id. */
+		key_map &= ~(1ULL << IPROTO_SPACE_ID);
+	}
 	if (key_map) {
 		enum iproto_key key = (enum iproto_key) bit_ctz_u64(key_map);
 		xrow_on_decode_err(row, ER_MISSING_REQUEST_FIELD,
```

</details>

src/box/iproto.cc
```diff
@@ -2149,6 +2188,8 @@ tx_process1(struct cmsg *m)
 	struct obuf_svp svp;
 	struct obuf *out;
 	tx_inject_delay();
+	if (tx_resolve_space_and_index_name(&msg->dml) != 0)
+		goto error;
```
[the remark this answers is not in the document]
CuriousGeorgiy: This is impossible, since I have a bunch of test cases for non-existing index of space names (I bet you have seen them).

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
```diff
@@ -0,0 +1,211 @@
...
+end
+
+local function inject_insert_or_replace(c, request)
+        local header = msgpack.encode({
+        [box.iproto.key.REQUEST_TYPE] = box.iproto.type[request],
+        [box.iproto.key.SYNC] = c:_next_sync(),
+    })
+    local body = msgpack.encode({
+        [box.iproto.key.SPACE_NAME] = 's',
+        [box.iproto.key.INDEX_NAME] = 'nonexistent',
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/iproto.cc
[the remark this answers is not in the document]
locker: [image](https://coveralls.io/builds/59231406/source?filename=src%2Fbox%2Fiproto.cc)

You tests only check `tx_process_select()`. There's no test for `tx_process1()`.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
```diff
@@ -0,0 +1,211 @@
...
+    local i = s.index.i
+    t.assert_equals(s.index, {i = i})
+    t.assert_equals(i, {name = 'i', space = s})
+
+    local s512 = c.space[512]
+    t.assert_equals(c.space, {[512] = s512, s = s})
+    t.assert_equals(s512, {id = 512, index = {}})
+    local i1 = s.index[1]
+    t.assert_equals(s.index, {[1] = i1, i = i})
+    t.assert_equals(i1, {id = 1, space = s})
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,40 @@ local function new_sm(uri, opts)
...
+                        if type(key) == 'number' then
+                            id_or_name = 'id'
+                        elseif type(key) == 'string' then
+                            id_or_name = 'name'
+                        else
+                            return nil
+                        end
+                        local space = {[id_or_name] = key}
+                        space.index = setmetatable({}, {
+                            __index = function(self, key)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Okay, changed to `space_key` and `index_key`: `id` is associated with a numerical value, besides we have separate `name` and `id` fields.

src/box/space_cache.h
```diff
@@ -89,13 +89,19 @@ space_cache_destroy(void);
 struct space *
 space_by_id(uint32_t id);
 
+/**
+ * `space_by_name` for NULL-terminated names.
+ */
+struct space *
+space_by_name0(const char *name);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Indeed, thanks for your suggestion, applied it.

src/box/xrow.c
locker: I didn't notice that we clear the bit for `IPROTO_SPACE_ID` above. Fixed the diff accordingly.

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index a2ac55332a67..755f2c9b9166 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1666,8 +1666,9 @@ iproto_msg_decode(struct iproto_msg *msg, struct cmsg_hop **route)
 		assert(type < sizeof(iproto_thread->dml_route) /
 			      sizeof(*iproto_thread->dml_route));
 		*route = iproto_thread->dml_route[type];
-		if (xrow_decode_dml(&msg->header, &msg->dml,
-				    dml_request_key_map(type)))
+		uint64_t key_map = dml_request_key_map(type) |
+				   iproto_key_bit(IPROTO_SPACE_NAME);
+		if (xrow_decode_dml(&msg->header, &msg->dml, key_map) != 0)
 			return -1;
 		/*
 		 * In contrast to replication requests, for a client request
diff --git a/src/box/xrow.c b/src/box/xrow.c
index 0e71abb30468..d1443254adad 100644
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -890,7 +890,8 @@ error:
 		xrow_on_decode_err(row, ER_INVALID_MSGPACK, "packet body");
 		return -1;
 	}
-
+	bool space_name_is_accepted =
+		(key_map & iproto_key_bit(IPROTO_SPACE_NAME)) != 0;
 	uint32_t size = mp_decode_map(&data);
 	for (uint32_t i = 0; i < size; i++) {
 		if (mp_typeof(*data) != MP_UINT) {
@@ -973,7 +974,11 @@ error:
 		}
 	}
 done:
-	if (key_map) {
+	if (space_name_is_accepted && request->space_name != NULL) {
+		/* The caller can accept space_name instead of space_id. */
+		key_map &= ~iproto_key_bit(IPROTO_SPACE_ID);
+	}
+	if (key_map != 0) {
 		enum iproto_key key = (enum iproto_key) bit_ctz_u64(key_map);
 		xrow_on_decode_err(row, ER_MISSING_REQUEST_FIELD,
 				   iproto_key_name(key));
```

</details>

src/box/iproto.cc
```diff
@@ -2138,6 +2138,45 @@ tx_process_rollback(struct cmsg *m)
...
+		}
+		dml->space_id = space->def->id;
+	}
+	if ((dml->type == IPROTO_SELECT || dml->type == IPROTO_UPDATE ||
+	     dml->type == IPROTO_DELETE) && dml->index_name != NULL) {
+		if (space == NULL)
+			space = space_cache_find(dml->space_id);
+		const char *idx_name =
+			tt_cstr(dml->index_name, dml->index_name_len);
+		struct index *idx = space_index_by_name(space, idx_name);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Applied your suggestion.

src/box/iproto.cc
```diff
@@ -2138,6 +2138,45 @@ tx_process_rollback(struct cmsg *m)
...
+			diag_set(ClientError, ER_NO_SUCH_SPACE,
+				 tt_cstr(dml->space_name, dml->space_name_len));
+			return -1;
+		}
+		dml->space_id = space->def->id;
+	}
+	if ((dml->type == IPROTO_SELECT || dml->type == IPROTO_UPDATE ||
+	     dml->type == IPROTO_DELETE) && dml->index_name != NULL) {
+		if (space == NULL)
+			space = space_cache_find(dml->space_id);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed and added a test case.

src/box/iproto.cc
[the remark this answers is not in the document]
CuriousGeorgiy: Sorry, missed that.

src/box/lua/net_box.lua
```diff
@@ -1201,6 +1242,10 @@ index_metatable = function(remote)
         if opts and opts.buffer then
             error("index:count() doesn't support `buffer` argument")
         end
+        if self.space.name == nil or self.name == nil then
+            error("index:count() is available only when space and index " ..
+                  "names are set")
+        end
```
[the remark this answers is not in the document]
locker: @alyapunov insists that this should be fixed now, in a separate patch. Please also open an issue (unless we don't have one already) because this is a user-visible improvement (`call` will start accepting numeric indexes).

andreyaksenov (approved):

src/box/lua/schema.lua
```diff
@@ -1858,7 +1858,7 @@ local port_c = ffi.cast('struct port_c *', port)
 
 -- Helper function to check space:method() usage
 local function check_space_arg(space, method)
-    if type(space) ~= 'table' or space.id == nil then
+    if type(space) ~= 'table' or (space.id == nil and space.name == nil) then
```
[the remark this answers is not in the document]
CuriousGeorgiy: Applied you suggestion.

src/box/iproto.cc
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed, added test cases for this.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
```diff
@@ -0,0 +1,211 @@
...
+    i = s.index[1]
+    t.assert_equals(s:replace{1, 2}, {1, 2})
+    t.assert_equals(i:select{2}, {{1, 2}})
+    t.assert_equals(s:update({0}, {{'=', 2, 3}}), {0, 3})
+    t.assert_equals(i:update({3}, {{'=', 2, 0}}), {0, 0})
+    t.assert_equals(s:upsert({2, 4}, {{'=', 2, 5}}), nil)
+    t.assert_equals(s:upsert({2, 4}, {{'=', 2, 5}}), nil)
+    t.assert_equals(i:min(), {0, 0})
+    t.assert_equals(i:max(), {2, 5})
+    t.assert_equals(i:get{2}, {1, 2})
```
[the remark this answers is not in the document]
CuriousGeorgiy: Parameterized the test over connection object type: skipped introspection tests for streams.

src/box/xrow.c
CuriousGeorgiy: Fixed. Added a test for recovery. @locker says we don't need a test for replication.

CuriousGeorgiy:
@locker checkpatch linter errors seem to FP.

src/box/lua/net_box.lua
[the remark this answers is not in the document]
CuriousGeorgiy: Opened #8604 and added a patch that closes it. Patched `net.box` to use this feature.

locker:
> @locker checkpatch linter errors seem to FP.

They're not FP - tabs are not allowed in Lua files.

```
RROR:SPACE_BEFORE_TAB: please, no space before tabs
#1047: FILE: test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua:257:
+-- ^Iif (space_id == 512) {$

ERROR:TABSTOP: please, use spaces instead of tabs
#1047: FILE: test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua:257:
+-- ^Iif (space_id == 512) {$
```

src/box/lua/net_box.lua
```diff
@@ -390,6 +391,49 @@ local function new_sm(uri, opts)
...
+                                return nil
+                            end
+                            id_or_name = 'name'
+                        else
+                            return nil
+                        end
+                        local space = {[id_or_name] = space_key,
+                                       _id_or_name = space_key}
+                        space.index = setmetatable({}, {
+                            __index = function(self, idx_key)
```
locker, line 411: Nit: `self` is unused. Should be replaced with `_`.

src/box/lua/net_box.lua
```diff
@@ -1201,8 +1249,20 @@ index_metatable = function(remote)
...
+            space_qualifier = "[" .. self.space.id .. "]"
+        end
+        local idx_qualifier
+        if self.name then
+            idx_qualifier = "." .. self.name
+        else
+            idx_qualifier = "[" .. self.id .. "]"
+        end
+        local code = string.format('box.space%s.index%s:count',
+                                   space_qualifier, idx_qualifier)
```
locker, line 1265: Nit: "qualifier" doesn't seem to be a right word choice for this variable IMO. Anyway, using both string formatting and concatenation looks ugly. Let's use only concatenation since we need it anyway:

```Lua
        local code = 'box.space' .. (self.space.name ~= nil and
                                     '.' .. self.space.name or
                                     '[' .. self.space.id .. ']') ..
                     'index' .. (self.name ~= nil and
                                 '.' .. self.name or '[' .. self.id .. ']') ..
                     ':count'
```

src/box/lua/schema.lua
locker, line 1861: Nit: Should be `space._id_or_name == nil`? Same comment to `check_index_arg`.

src/box/iproto.cc
```diff
@@ -2138,6 +2138,47 @@ tx_process_rollback(struct cmsg *m)
...
 }
 
+/*
+ * In case the request does not contain a space or identifier but contains a
+ * corresponding name, tries to resolve the name.
+ * A space identifier is mandatory for a DML request, so in case the space
+ * identifier is not present in the request and we fail to get an identifier
+ * from the space name, we must return with an
+ * error, setting a missing request field diagnostic, just as it was done in
+ * `xrow_decode_dml` before space names in IPROTO requests were introduced.
```
locker, line 2148: Nit: This comment seems outdated - the code below doesn't set "missing request field diagnostic".

src/box/xrow.h
```diff
@@ -228,12 +236,33 @@ request_str(const struct request *request);
...
+static inline int
 xrow_decode_dml(struct xrow_header *xrow, struct request *request,
-		uint64_t key_map);
+		uint64_t key_map)
+{
+	return xrow_decode_dml_internal(xrow, request, key_map, false);
+}
+
+/**
+ * Decode DML from client request (Lua, C or IPROTO).
```
locker, line 258: Nit: AFAIU this function is only used by Lua clients.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
```diff
@@ -0,0 +1,296 @@
...
+end
+
+-- Checks that space and index name resolution works correctly.
+g.test_space_and_index_name_resolution = function(cg)
+    local c = netbox:connect(cg.server.net_box_uri, {fetch_schema = false})
+    if cg.params.connection == 'stream' then
+        c = c:new_stream()
+    end
+
+    t.assert_equals(inject_select(c, cg.s1_id, 's2'), {{2}})
```
locker, line 223: AFAICS `_inject` doesn't set stream id. You have to do it manually.

src/box/lua/call.c
```diff
@@ -92,61 +94,239 @@ get_call_serializer(void)
...
+struct lexer {
+	/* Source string. */
+	const char *src;
+	/* Length of string. */
+	size_t src_len;
+	/* Current lexer offset in bytes. */
+	size_t offset;
+	/* Current lexer offset in symbols. */
+	size_t symbol_count;
+};
```
locker, line 107: An external state seems like an overkill. Can you do the whole parsing in `box_lua_find` without polluting the file namespace?

Also, we don't need to handle unicode symbols here - at least we didn't handle them before this patch.

locker (changes requested):

test/box-luatest/gh_8604_support_square_brackets_in_procedure_resolution_for_lua_calls_test.lua
```diff
@@ -0,0 +1,210 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group(nil, t.helpers.matrix{proc = {
+    'a.b.c',
```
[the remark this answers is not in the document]
CuriousGeorgiy: It is already tested [here](https://github.com/tarantool/tarantool/blob/1d6043fac53bebd539ee5882f6073767eea753bd/test/box/function1.test.lua#L101-L129).

src/box/lua/call.c
CuriousGeorgiy: > Also, we don't need to handle unicode symbols here - at least we didn't handle them before this patch.

What makes you think so? Please study [this](https://github.com/tarantool/tarantool/blob/1d6043fac53bebd539ee5882f6073767eea753bd/test/box/function1.test.lua#L101-L129) test and the related GitHub tickets.

src/box/xrow.h
CuriousGeorgiy: And what function do IPROTO clients use?

src/box/lua/schema.lua
CuriousGeorgiy: Box spaces don't have this field.

src/box/lua/call.c
CuriousGeorgiy: > An external state seems like an overkill. Can you do the whole parsing in box_lua_find without polluting the file namespace?

It doesn't seem like an overkill, inlining everything looks incomprehensible.

src/box/iproto.cc
CuriousGeorgiy: Dropped this comment.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
CuriousGeorgiy: AFAIC, stream identifier is only needed for transactions.

src/box/lua/net_box.lua
CuriousGeorgiy: Applied your suggestion, thanks!

src/box/lua/net_box.lua
CuriousGeorgiy: Fixed.

src/box/lua/net_box.lua
```diff
@@ -1201,8 +1249,12 @@ index_metatable = function(remote)
...
             error("index:count() doesn't support `buffer` argument")
         end
-        local code = string.format('box.space.%s.index.%s:count',
-                                   self.space.name, self.name)
+        local code = 'box.space' .. (self.space.name ~= nil and
+                                     '.' .. self.space.name or
+                                     '[' .. self.space.id .. ']') ..
+                     '.index' .. (self.name ~= nil and
+                                 '.' .. self.name or '[' .. self.id .. ']') ..
+                     ':count'
```
olegrok, line 1257: I don't know that we should do here...

```
tarantool> c.space.abc.index.bcd:count()
---
- error: Procedure 'box.space.abc.index.bcd:count' is not defined
...

tarantool> c.space['is it possible'].index.bcd:count()
---
- error: Procedure 'box.space.is it possible.index.bcd:count' is not defined
...
```

In case if space/index contains spaces we anyway fail... (It's wierd but it's possible).
And probably error message talks about function... Not about space/index existance

CuriousGeorgiy: AFAIC, you cannot use non-printable characters in Tarantool identifiers.

olegrok:
```
tarantool> box.schema.space.create('you can')
---
- engine: memtx
  before_replace: 'function: 0x0102e5bf10'
  field_count: 0
  is_sync: false
  on_replace: 'function: 0x0102e5bed8'
  temporary: false
  index: []
  is_local: false
  enabled: false
  name: you can
  id: 517
- created
...

tarantool> box.space['you can']:create_index('do it')
---
- unique: true
  parts:
  - type: unsigned
    is_nullable: false
    fieldno: 1
  hint: true
  id: 0
  type: TREE
  space_id: 517
  name: do it
...

tarantool> box.space['you can'].index['do it']:select()
2023-04-29 11:10:00.462 [39623] main/103/interactive C> Potentially long select from space 'you can' (517)
 stack traceback:
	builtin/box/schema.lua:2497: in function 'log_long_select'
	builtin/box/schema.lua:2515: in function <builtin/box/schema.lua:2500>
	[C]: in function 'pcall'
	builtin/box/console.lua:427: in function 'eval'
	builtin/box/console.lua:819: in function 'repl'
	builtin/box/console.lua:870: in function <builtin/box/console.lua:858>
---
- []
...
```

CuriousGeorgiy:
```console
tarantool> box.schema.space.create(' s\n')
---
- error: 'Invalid identifier '' s

    '' (expected printable symbols only or it is too long)'
...
```

I thought all whitespace symbols are considered non-printable.

Okay, well, I guess we indefinitely using square brackets could solve this.

CuriousGeorgiy: Fixed, added a test case for this.

src/box/xrow.h
CuriousGeorgiy: Actually, I guess this function is only used by IPROTO clients (perhaps you meant the same). Renamed the function to `xrow_decode_dml_iproto` and changed the comment accordingly.

test/box-luatest/gh_8146_space_and_index_name_in_iproto_requests_test.lua
locker: No, streams can be used without transactions. Anyway, this test doesn't make sense to run with and without streams unless `inject_select` sets `stream_id`.

src/box/lua/call.c
locker: Up to @alyapunov.

src/box/lua/call.c
```diff
@@ -92,61 +95,239 @@ get_call_serializer(void)
...
 	}
+	return -1;
+}
 
+/*
+ * Parse digit sequence into integer until non-digit is met. Lexer stops right
+ * after the last digit.
+ */
+static int
+lexer_parse_integer(struct lexer *lexer, int *num)
```
olegrok, line 196: You need to support negative values

```
tarantool> _G.map = {[-1] = print, [0] = print, [1] = print}
---
...

tarantool> c:call('map[0]', {1, 2, 3})
1	2	3
---
...

tarantool> c:call('map[1]', {1, 2, 3})
1	2	3
---
...

tarantool> c:call('map[-1]', {1, 2, 3})
---
- error: Procedure 'map[-1]' is not defined
...
```

CuriousGeorgiy: I think it is possible to think of a lot of exotic cases, but AFAIC, we only want to support the main ones. Could you please confirm this is really needed from @unera or @alyapunov?

src/box/lua/call.c
alyapunov: I propose alternative solution: https://github.com/tarantool/tarantool/pull/8660
I refactored and commented `box_lua_find` and simply add bracket support. The function is just few lines longer than the original one, but I hope it became more simple and understandable. It also naturally support negative and fractional numbers in brackets. I also stole your test cases but changed the test that it runs much faster.

locker:
Will review after #8660 is merged and this PR is rebased.

CuriousGeorgiy:
@locker #8660 was merged, rebased onto it.

src/box/lua/call.c
CuriousGeorgiy: The implementation provided in #8660 by @alyapunov supports negative indices (merged).

src/box/lua/call.c
CuriousGeorgiy: #8660 was merged.

alyapunov (approved):

CuriousGeorgiy:
@locker @alyapunov looks like there is another problem with `IPROTO_SPACE_NAME`, it causes a bit shift overflow here:
https://github.com/tarantool/tarantool/blob/4507c59de6af427a4a8f7847fe6c1c31d3e7eacc/src/box/xrow.c#L922-L928

I added a `key < 64` check before the bit flipping.

PR #8580, opened 2023-04-19
sql: use xregion_*() functions

### LINKED ISSUE #8522: `xregion_alloc_object()` and `xregion_alloc_array()` are quite hard to use
These are "never failing" versions of `region_alloc_object()` and `region_alloc_array()`.

* The latter take an out-argument `size`. It is needed only to return size of the allocation on an error, to use it in `diag_set()`. If an alloc function never fails, it doesn't need any `size` out-args.

* These functions return `void` but the originals return the pointer of the needed type. It allowed to avoid casts in C++. The `x`-versions of those should behave the same.

ImeevMA:
This patch replaces `region_*()` functions with `xregion_*()` functions.

Closes #8522.


src/box/bind.c
```diff
@@ -175,12 +175,8 @@ sql_bind_list_decode(const char *data, struct sql_bind **out_bind)
...
 	uint32_t used = region_used(region);
 	size_t size;
-	struct sql_bind *bind = region_alloc_array(region, typeof(bind[0]),
-						   bind_count, &size);
-	if (bind == NULL) {
-		diag_set(OutOfMemory, size, "region_alloc_array", "bind");
-		return -1;
-	}
+	struct sql_bind *bind = xregion_alloc_array(region, typeof(bind[0]),
+						    bind_count, &size);
```
Gerold103, line 179: Would you mind doing this simple one first? https://github.com/tarantool/tarantool/issues/8522. Otherwise all those places would have to be updated again after that ticket.

src/box/sql/select.c
```diff
@@ -2049,7 +2045,7 @@ generate_column_metadata(struct Parse *pParse, struct SrcList *pTabList,
  * Return 0 on success.  If a memory allocation error occurs,
  * store NULL in *paCol and 0 in *pnCol and return -1.
  */
-int
+void
```
Gerold103, line 2048: The comment is outdated now.

Gerold103 (commented):
Thanks for the patch!

src/box/sql/select.c
ImeevMA: Fixed in another PR.

src/box/bind.c
ImeevMA: Thank you for the suggestion! Done

Gerold103 (approved):
Thanks for the fixes!

src/trivia/util.h
```diff
@@ -128,17 +128,21 @@ strnindex(const char *const *haystack, const char *needle, uint32_t len,
...
 #define xibuf_alloc(p, size)	xalloc_impl((size), ibuf_alloc, (p), (size))
 #define xibuf_reserve(p, size)	xalloc_impl((size), ibuf_reserve, (p), (size))
 #define xruntime_memory_alloc(size) \
 	xalloc_impl((size), runtime_memory_alloc, (size))
 
+#define xregion_alloc_object(region, T) ({					\
+	(T *)xregion_aligned_alloc((region), sizeof(T), alignof(T));		\
+})
+#define xregion_alloc_array(region, T, count) ({				\
+	(T *)xregion_aligned_alloc((region), sizeof(T) * (count), alignof(T));\
```
locker, line 143: Nit: `\` not aligned.

src/box/lua/execute.c
```diff
@@ -411,12 +401,8 @@ lua_sql_bind_decode(struct lua_State *L, struct sql_bind *bind, int idx, int i)
 			return -1;
 		}
 		bind->bytes = region_used(region) - used;
-		bind->s = region_join(region, bind->bytes);
-		if (bind->s != NULL)
-			break;
-		region_truncate(region, used);
-		diag_set(OutOfMemory, bind->bytes, "region_join", "bind->s");
-		return -1;
+		bind->s = xregion_join(region, bind->bytes);
```
locker, line 404: There's `OutOfMemory` above set when `region_serve_cb` or `region_alloc_cb` fails. Can you remove that too in the scope of this work? Same comment to other functions using `mpstream`, for example `sql_encode_table`.

src/box/sql/build.c
```diff
@@ -290,39 +274,25 @@ sql_shallow_space_copy(struct Parse *parse, struct space *space)
...
-	ret->index = region_alloc_array(&parse->region, typeof(struct index *),
-					ret->index_count, &size);
-	if (ret->index == NULL) {
-		diag_set(OutOfMemory, size, "region_alloc_array", "ret->index");
-		goto error;
-	}
+	ret->index = xregion_alloc_array(&parse->region, typeof(struct index *),
+					 space->index_count);
+	size_t size = sizeof(struct index) * space->index_count;
 	memcpy(ret->index, space->index, size);
```
locker, line 282: Nit: I wouldn't introduce the local variable `size` because it's used only once in `memcpy` - I'd pass the size directly to `memcpy` instead. Same comment to the `fields_size`.

src/trivia/util.h
ImeevMA: Thanks, fixed.

src/box/sql/build.c
ImeevMA: Fixed.

locker (approved):

locker:
ASAN workflow [fails](https://github.com/tarantool/tarantool/actions/runs/4787799009/jobs/8513527261?pr=8580). PTAL.

ImeevMA:
Hi, @locker! I hope you don't mind taking another look at this PR. I found two issues in the SQL code that were caught by ASAN and fixed them, this is the diff:
```Diff
diff --git a/src/box/sql/build.c b/src/box/sql/build.c
index 3d414caf9..061545111 100644
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -279,7 +279,7 @@ sql_shallow_space_copy(struct Parse *parse, struct space *space)
        ret->index = xregion_alloc_array(&parse->region, typeof(struct index *),
                                         space->index_count);
        memcpy(ret->index, space->index,
-              sizeof(struct index) * space->index_count);
+              sizeof(struct index *) * space->index_count);
        memcpy(ret->def, space->def, sizeof(struct space_def));
        ret->def->opts.is_temporary = true;
        ret->def->opts.is_ephemeral = true;
diff --git a/src/box/sql/mem.c b/src/box/sql/mem.c
index f75095344..aa46d551f 100644
--- a/src/box/sql/mem.c
+++ b/src/box/sql/mem.c
@@ -3365,7 +3365,7 @@ static struct Mem *
 vdbemem_alloc_on_region(uint32_t count)
 {
-       struct Mem *ret = xregion_alloc_array(region, typeof(*ret), count);
+       struct Mem *ret = xregion_alloc_array(region, struct Mem, count);
        memset(ret, 0, count * sizeof(*ret));
        for (uint32_t i = 0; i < count; i++) {
                mem_create(&ret[i]);
```

The first is quite obvious. For the second, I found that the `region_aligned_alloc()` function received `1 * count` as the size of the `struct Mem` and `1` as the alignment. I don't know exactly why this is, but I think it's because of a conflict between the name `struct Mem` and some other structure with the same name.

PR #8582, opened 2023-04-19
Always reload URIs when listen/replication is reconfigured

locker:
Currently, `box.cfg` doesn't call an option handler if the option value is unchanged (in Lua). This means that

```Lua
box.cfg{listen = box.cfg.listen}
box.cfg{replication = box.cfg.replication}
```

are no-ops. Since paths to SSL certificate files are stored in URI parameters, this means that one can't reload SSL certificate files by resetting the configuration to the same values, which is inconvenient.

We tried to unconditionally call the option handler even if the old and new option values are the same, see #8552, but this broke some integration tests. In particular, it broke vshard tests, as explained in https://github.com/tarantool/tarantool/issues/8551#issuecomment-1505503984, so we refused from this approach.

In this PR, apart from removing the Lua `box.cfg` check, we also patch the `box.cfg.replication` and `box.cfg.listen` option handlers to safely reload URI parameters by recreating IO stream contexts without disrupting existing connections or establishing new ones in case the URIs are the same.

Needed for tarantool/tarantool-ee#432

### DIFF
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -450,7 +450,7 @@ replicaset_on_health_change(void)
 		trigger_run(&replicaset_on_quorum_loss, NULL);
 }
 
-void
+static void
 replica_set_applier(struct replica *replica, struct applier *applier)
 {
 	assert(replica->applier == NULL);
@@ -482,7 +482,7 @@ replica_update_applier_health(struct replica *replica)
 	replicaset_on_health_change();
 }
 
-void
+static void
 replica_clear_applier(struct replica *replica)
 {
 	assert(replica->applier != NULL);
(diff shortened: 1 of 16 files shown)


src/box/replication.cc
```diff
@@ -1045,9 +1060,8 @@ replicaset_connect(struct applier **appliers, int count,
 		/* Timeout or connection failure. */
 		if (state.connected < replicaset_connect_quorum(count) &&
 		    connect_quorum) {
-			diag_set(ClientError, ER_CFG, "replication",
-				 "failed to connect to one or more replicas");
-			goto error;
+			tnt_raise(ClientError, ER_CFG, "replication",
```
alyapunov, line 1067: Can't catch why you don't trigger_clear(&triggers[i].base);

locker: It isn't necessary because we call `applier_delete` in the guard destructor, which destroys all triggers.

locker:
We found a bug in `box_sync_replication`: the `apppliers_guard` never deleted appliers on failure. I also forgot to disable the guard in `replication_connect` on success but the tests passed because of this bug. Fixed in a separate patch. The accumulated diff is below.

```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index d6686bb5a61a..eee635ceefef 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -991,7 +991,7 @@ replicaset_connect(const struct uri_set *uris,
 	}
 	int count = 0;
 	struct applier *appliers[VCLOCK_MAX] = {};
-	auto appliers_guard = make_scoped_guard([=]{
+	auto appliers_guard = make_scoped_guard([&]{
 		for (int i = 0; i < count; i++) {
 			struct applier *applier = appliers[i];
 			applier_stop(applier);
@@ -1086,7 +1086,7 @@ replicaset_connect(const struct uri_set *uris,
 
 	/* Now all the appliers are connected, update the replica set. */
 	replicaset_update(appliers, count, keep_connect);
-
+	appliers_guard.is_active = false;
 	uri_set_destroy(&replication_uris);
 	uri_set_copy(&replication_uris, uris);
 }
```

</details>

locker:
Added a comment to `iproto_listen` explaining why we update `iproto_uris` before applying configuration.

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 35175ee18c00..1ec8b738fa6a 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -3433,6 +3433,12 @@ iproto_listen(const struct uri_set *uri_set)
 		iproto_send_restart_msg();
 		return 0;
 	}
+	/*
+	 * Note that we set iproto_uris before trying to bind so even if
+	 * we fail, iproto_uris will still contain the new configuration.
+	 * It's okay because box.cfg.listen is reverted on failure at
+	 * the box.cfg level.
+	 */
 	uri_set_destroy(&iproto_uris);
 	uri_set_copy(&iproto_uris, uri_set);
 	iproto_send_stop_msg();
```

</details>

alyapunov (approved):

Issue #5485, opened 2020-11-02
AST generation from SQL frontend


sergos:
As part of the https://github.com/tarantool/tarantool/issues/5484 we need to implement abstract tree generation from the existent SQL parser. 
The parser should preserve capabilities to generate the VDBE input in its current form with zero performance impact.

PR #8592, opened 2023-04-21
sql: rework parsing of RENAME and DROP statements

ImeevMA:
This patch-set reworks RENAME and DROP statements. There are six of them:
```
ALTER TABLE table_name RENAME TO table_new_name;
ALTER TABLE table_name DROP CONSTRAINT constraint_name;
DROP INDEX index_name ON table_name;
DROP TRIGGER trigger_name;
DROP VIEW view_name;
DROP TABLE table_name;
```

Prior to this patch, VDBE for these statements were generated while the statement was in the process of being parsed. After this patch, VDBE for all these statements will be generated only after the parsing has successfully completed.

Part of https://github.com/tarantool/tarantool/issues/5485

### DIFF
--- a/src/box/sql/alter.c
+++ b/src/box/sql/alter.c
@@ -38,37 +38,32 @@
 #include "box/schema.h"
 
 void
-sql_alter_table_rename(struct Parse *parse)
+sql_alter_table_rename(struct Parse *parse, struct Token *old_name,
+		       struct Token *new_name)
 {
-	struct rename_entity_def *rename_def = &parse->rename_entity_def;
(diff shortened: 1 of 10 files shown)


Gerold103 (approved):
Thanks for the patch! See one nit.

locker:
@ImeevMA I don't feel I can review SQL patches anymore, sorry. Please assign to someone else for review/merge.

ImeevMA:
This PR seems to be unnecessary anymore, since it was reworked in #12989

PR #8596, opened 2023-04-24
app: fix possible integer overflow in net_box.c

ImeevMA:
The netbox_transport_send_and_recv() function takes a part of the data buffer size as a parameter, and it is possible that the resulting buffer size could become larger than SIZE_MAX, resulting in an integer overflow and a segmentation fault.

Closes #tarantool/security#121

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/ghs-121-too-big-buffer-size.md
+++ b/changelogs/unreleased/ghs-121-too-big-buffer-size.md
@@ -1,0 +1,3 @@
+## bugfix/core
+
+* Fixed an integer overflow issue in `net.box` (ghs-121).
--- a/src/box/lua/net_box.c
+++ b/src/box/lua/net_box.c
@@ -1161,8 +1161,14 @@ netbox_transport_send_and_recv(struct netbox_transport *transport,
 		} else {
 			const char *bufpos = transport->recv_buf.rpos;
 			const char *rpos = bufpos;
-			size_t len = mp_decode_uint(&rpos);
-			required = (rpos - bufpos) + len;
+			uint64_t len = mp_decode_uint(&rpos);
+			size_t size = rpos - bufpos;
+			if (len > SIZE_MAX - size) {
+				box_error_raise(ER_NO_CONNECTION,
+						"Response size too large");
+				return -1;
+			}
+			required = size + len;
 			if (data_len >= required) {
 				const char *body_end = rpos + len;
 				transport->recv_buf.rpos = (char *)body_end;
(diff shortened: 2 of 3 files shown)


src/box/lua/net_box.c
```diff
@@ -1162,7 +1162,13 @@ netbox_transport_send_and_recv(struct netbox_transport *transport,
 			const char *bufpos = transport->recv_buf.rpos;
 			const char *rpos = bufpos;
 			size_t len = mp_decode_uint(&rpos);
-			required = (rpos - bufpos) + len;
+			size_t size = rpos - bufpos;
+			if (size > SIZE_MAX - len) {
+				box_error_raise(ER_NO_CONNECTION,
+						"Data buffer size too large");
```
locker, line 1168: Nit:
 - It's not "data buffer size" that is too large. It's "response size".
 - Let's decode the length into `uint64_t` because `mp_decode_uint` returns `uint64_t`.
 - Let's put `len` to the left to emphasize that we check `len`, not `size`.

```C
uint64_t len = mp_decode_uint(&rpos);
size_t size = rpos - bufpos;
if (len > SIZE_MAX - size) {
        box_error_raise(ER_NO_CONNECTION, "Response size is too large");
        return -1;
}
```

changelogs/unreleased/ghs-121-too-big-buffer-size.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/app
+
+* Fixed integer overflow issue net.box (ghs-121).
```
locker, line 3: - **in** net.box.
- We never used the "app" section in changelogs. I'm not sure what section we should use for net.box, but I'd use "bugfix/core".

locker (changes requested):

src/box/lua/net_box.c
ImeevMA: Fixed.

changelogs/unreleased/ghs-121-too-big-buffer-size.md
ImeevMA: Fixed.

locker (approved):

locker:
Waiting for review from the doc team.

changelogs/unreleased/ghs-121-too-big-buffer-size.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/core
+
+* Fixed integer overflow issue in net.box (ghs-121).
```
andreyaksenov, line 3: A couple of minor suggestions:
```
* Fixed an integer overflow issue in `net.box` (ghs-121).
```

andreyaksenov (approved):

changelogs/unreleased/ghs-121-too-big-buffer-size.md
ImeevMA: Fixed.

PR #8607, opened 2023-04-27
static-build: enable compiler optimizations for dependencies

### LINKED ISSUE #8606: Static build doesn't enable compiler optimization for dependencies
It looks like compiler optimization isn't enabled for external projects when Tarantool is built with statically linked dependencies:

1. Build static binary:
   ```
   cmake -S static-build -B static-build/build/release -DCMAKE_TARANTOOL_ARGS=-DCMAKE_BUILD_TYPE=RelWithDebInfo
   make -C static-build/build/release -j16
   ```
2. Grep for compiler optimization flags:
   ```
   cd static-build/build/release/
   find *-prefix -type f -iregex '.*-build.*\(mk\|make\).*' | xargs grep -l '^[^#].*\-O[0-9]' | sed -e 's/\([a-z]*-prefix\).*/\1/p' | sort -u
   ```
   Output:
   ```
   openssl-prefix
   tarantool-prefix
   ```
   Other projects seem to have no optimization enabled:
   ```
   ls -1d *-prefix
   ```
   Output:
   ```
   iconv-prefix
   icu-prefix
   ncurses-prefix
   openssl-prefix
   readline-prefix
   tarantool-prefix
   zlib-prefix
   ```

This may cause performance degradation issues, like the one reported in https://github.com/tarantool/tarantool-ee/issues/440.

This happens because an autoconf `configure` script doesn't set compiler optimization flags if `CFLAGS` is set explicitly, even if `--enable-release` is specified. Here's an extract from the `configure` script generated for the icu project (`icu-prefix/src/icu/source/configure`):

```
# Ensure that if CXXFLAGS/CFLAGS were not set when calling configure, set it correctly based on (enable/disable) debug or release option
# The release mode use is the default one for autoconf
if test "$GCC" = yes; then
    if test "$CFLAGS" = ""; then
        if test "$ENABLE_DEBUG" = 1; then
            CFLAGS=-g
        fi
        if test "$ENABLE_RELEASE" = 1; then
            CFLAGS="$CFLAGS -O2"
        fi
    fi
    if test "$CXXFLAGS" = ""; then
        if test "$ENABLE_DEBUG" = 1; then
            CXXFLAGS=-g
        fi
        if test "$ENABLE_RELEASE" = 1; then
            CXXFLAGS="$CXXFLAGS -O2"
        fi
    fi
fi
```

We started to set `CFLAGS` since commit e6abe1c91ae36127bedcd3b9a5d5d4705b802c91 which first appeared in 2.11 and was also cherry-picked to 2.10. The first affected versions are 2.11.0-entrypoint-488-ge6abe1c91ae3 and 2.10.2-18-gce4b08ebb877.

locker:
An autoconf-generated `configure` script doesn't enable compiler optimization flags if `CFLAGS` / `CXXFLAGS` options are set explicitly. We started setting `CFLAGS` / `CXXFLAGS` in commit e6abe1c91ae3 ("cmake: add extra security compiler options"). As a result, users started experiencing performance degradation issues, like the one described in tarantool/tarantool-ee#440.

Let's set `-O2` in `CFLAGS` / `CXXFLAGS` explicitly to fix that.

Closes #8606
Needed for tarantool/tarantool-ee#440


static-build/cmake/AddDependencyProjects.cmake
```diff
@@ -33,14 +33,20 @@ set(BACKUP_STORAGE https://distrib.hb.bizmrg.com)
...
 # CFLAGS from environment, so we should not do it for LDFLAGS too.
-set(DEPENDENCY_CFLAGS "")
-set(DEPENDENCY_CXXFLAGS "")
+#
+# Note: A configure script generated by autoconf doesn't enable compiler
+# optimizations if CFLAGS / CXXFLAGS is set so we have to enable them
+# explicitly, see https://github.com/tarantool/tarantool/issues/8606
+#
+set(DEPENDENCY_CFLAGS "-O2")
+set(DEPENDENCY_CXXFLAGS "-O2")
```
olegrok, line 42: Should we also add "-g" for debug build?

changelogs/unreleased/gh-8606-static-build-enable-optimization.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/build
+
+* Enabled compiler optimizations for static build dependencies, which were
+  erroneously disabled in 2.10.3 (gh-8606).
```
xuniq, line 4:
```suggestion
  erroneously disabled in version 2.10.3 (gh-8606).
```
Suggestion:

xuniq (approved):

static-build/cmake/AddDependencyProjects.cmake
locker: There's no such option in Tarantool static-build CMake config as "debug build". Also, we never added `-g` to dependency projects. I don't think that we need to have debug info for them because we don't expect bugs in dependencies.

changelogs/unreleased/gh-8606-static-build-enable-optimization.md
locker: Applied. Thanks.

nshy (approved):

static-build/cmake/AddDependencyProjects.cmake
olegrok: > we don't expect bugs in dependencies

On the one hand you are right but on the other hand it's still possible and in case if it happens one day it could be quite painful.

Well, let's stay as is.

olegrok (approved):

locker:

Issue #8099, opened 2022-12-27
Drop ephemeral spaces


ImeevMA:
Ephemeral spaces have been introduced to help us integrate SQL into Tarantool. However, this increases the complexity of spaces and introduces many new constraints on SQL and BOX, such as the number of space IDs or the uncounted amount of memory used by spaces. To remove these restrictions, it is proposed to replace ephemeral spaces with something that should not be considered a space.

deffulsFHSS:
Hello, sorry for writing it here.
I got a problem with ephemeral space.
I am not able to perform an sql request from nodejs tarantool-driver. db.sql('select "user_id" from "users" where <cond1> and <cond2>;")

It says that my user has no read access to ephemeral space. Dont understand why because it has read, write, execute for universe.

PR #8617, opened 2023-04-28
sql: refactor PSEUDO and SORTER cursors

ImeevMA:
This patch-set refactors the PSEUDO and SORTER cursors.

The first patch make the SORTER cursor to prepare the data before passing it directly to the OP_Column, instead of passing the data to the OP_Column using the PSEUDO cursor and preparing the data in the first OP_Column for each row.

The second patch resets the nullRow field for the PSEUDO cursor and causes the PSEUDO cursor to prepare the data before passing it to the OP_Column instead of preparing the data in the first OP_Column for each row.

The logical continuation of these patches is the actual preparation of data for the TARANTOOL cursor before passing it to OP_Column and the removal of the cacheStatus field from the cursors. However, at the moment there are some problems with this. The main reasons are that a TARANTOOL cursor may not contain data and that the opcode OP_RowData can be used with a TARANTOOL cursor instead of OP_Column for optimization purposes. In case of passing to OP_RowData we do not need to prepare data.

This patch-set is a preparation for adding new cursors for new structures that will replace ephemeral tables.

Needed for #8099

### DIFF
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -3630,7 +3630,7 @@ sqlExprCodeTarget(Parse * pParse, Expr * pExpr, int target)
 				return pCol->iMem;
 			} else if (pAggInfo->useSortingIdx) {
 				sqlVdbeAddOp3(v, OP_Column,
-						  pAggInfo->sortingIdxPTab,
+					      pAggInfo->sortingIdx,
 						  pCol->iSorterColumn, target);
 				return target;
 			}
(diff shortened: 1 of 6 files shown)


Gerold103 (approved):
Thanks for the patch! Frankly, I don't understand most of it. But I don't see any obvious problems.

locker:
@ImeevMA I don't feel I can review SQL patches anymore, sorry. Please assign to someone else for review/merge.

ImeevMA:
It appears this pull request is already out of date. It's probably better to rework it entirely, if necessary, rather than update it.

PR #8622, opened 2023-05-03
lua: introduce module triggers

drewdzzz:
New module allows users to set arbitrary handlers on arbitrary events.
All the existing tarantool triggers will be moved to this module later.

**NB**: despite the fact that it was written that `set` method returns old trigger in the design document, I decided to do another way - `set` returns new trigger. Rationale: firstly, for the sake of consistency with `space:insert` which returns new tuple. Secondly, only with such API user can check if the trigger was set (we will introduce `trigger.on_change` trigger which will be able to throw an error to disallow to set the trigger, and ability of such check will be helpful then).

Closes https://github.com/tarantool/tarantool/issues/8656

### DIFF
--- a/changelogs/unreleased/gh-8656-trigger-registry.md
+++ b/changelogs/unreleased/gh-8656-trigger-registry.md
@@ -1,0 +1,4 @@
+## feature/lua
+
+* Introduced the new Lua module `trigger`. This module enables managing and
+  calling triggers stored in trigger registry (gh-8656).
--- a/src/box/lua/trigger.c
+++ b/src/box/lua/trigger.c
@@ -1,0 +1,255 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#include "box/lua/func_adapter.h"
+#include "core/event.h"
+#include <diag.h>
+#include <fiber.h>
+#include "lua/utils.h"
+
+/**
+ * Sets a trigger with passed name to the passed event.
+ * The first argument is event name, the second one is trigger name, the third
+ * one is new trigger handler - it can be a Lua function or another callable
+ * object. If there is an already registered trigger with such name in the
+ * event, it is replaced with new trigger.
+ * Returns new trigger handler (which was passed as the third argument).
+ */
+static int
+luaT_trigger_set(struct lua_State *L)
+{
+	if (lua_gettop(L) != 3)
+		luaL_error(L, "Usage: trigger.set(event, trigger, function)");
+	const char *event_name = luaL_checkstring(L, 1);
+	const char *trigger_name = luaL_checkstring(L, 2);
+	if (!luaL_iscallable(L, 3))
+		luaL_typerror(L, 3, "callable");
+	/*
+	 * The following code is written in assumption no error will be thrown.
+	 */
+	struct event *event = event_get(event_name, true);
+	assert(event != NULL);
+	struct func_adapter *func = func_adapter_lua_create(L, 3);
+	event_reset_trigger(event, trigger_name, func);
+	/* The new handler is still at the top of the stack. */
+	return 1;
+}
+
+/**
+ * Deletes a trigger with passed name from passed event.
+ * The first argument is event name, the second one is trigger name.
+ * Returns deleted trigger handler.
+ */
+static int
+luaT_trigger_del(struct lua_State *L)
+{
+	if (lua_gettop(L) != 2)
+		luaL_error(L, "Usage: trigger.del(event, trigger)");
+	const char *event_name = luaL_checkstring(L, 1);
+	const char *trigger_name = luaL_checkstring(L, 2);
+	struct event *event = event_get(event_name, false);
+	if (event == NULL)
+		return 0;
+	struct func_adapter *old = event_find_trigger(event, trigger_name);
+	if (old == NULL)
+		return 0;
+	func_adapter_lua_get_func(old, L);
+	event_reset_trigger(event, trigger_name, NULL);
+	return 1;
+}
+
+/**
+ * Calls all the triggers registered on passed event with variable number of
+ * arguments. Execution is stopped by a first exception.
+ * First argument must be a string - all the other arguments will be passed
+ * to the triggers without any processing or copying.
+ * Returns no values on success. If one of the triggers threw an error, it is
+ * raised again.
+ */
+static int
+luaT_trigger_call(struct lua_State *L)
+{
+	if (lua_gettop(L) < 1)
+		luaL_error(L, "Usage: trigger.call(event, [args...])");
+	const char *event_name = luaL_checkstring(L, 1);
+	struct event *event = event_get(event_name, false);
+	if (event == NULL)
+		return 0;
+	int top = lua_gettop(L);
+	int narg = top - 1;
+	struct event_trigger_iterator it;
+	event_trigger_iterator_create(&it, event);
+	struct func_adapter *trigger = NULL;
+	const char *name = NULL;
+	int rc = 0;
+	while (rc == 0 && event_trigger_iterator_next(&it, &trigger, &name)) {
+		func_adapter_lua_get_func(trigger, L);
+		for (int i = top - narg + 1; i <= top; ++i)
+			lua_pushvalue(L, i);
+		rc = luaT_call(L, narg, 0);
+	}
+	event_trigger_iterator_destroy(&it);
+	if (rc != 0)
+		return luaT_error(L);
+	return 0;
+}
+
+/**
+ * Sets an array of arrays [trigger_name, trigger_handler] by event->name key
+ * in a pre-created table. Never sets an empty array.
+ */
+static bool
+trigger_info_push_event(struct event *event, void *arg)
+{
+	lua_State *L = (lua_State *)arg;
+	struct event_trigger_iterator it;
+	struct func_adapter *trigger = NULL;
+	const char *name = NULL;
+	int idx = 0;
+	lua_createtable(L, 0, 0);
+	event_trigger_iterator_create(&it, event);
+	while (event_trigger_iterator_next(&it, &trigger, &name)) {
+		idx++;
+		lua_createtable(L, 2, 0);
+		lua_pushstring(L, name);
+		lua_rawseti(L, -2, 1);
+		func_adapter_lua_get_func(trigger, L);
+		lua_rawseti(L, -2, 2);
+		lua_rawseti(L, -2, idx);
+	}
+	event_trigger_iterator_destroy(&it);
+	lua_setfield(L, -2, event->name);
+	return true;
+}
+
+/**
+ * Pushes a key-value table, where the key is the event name and value is an
+ * array of triggers, represented by two-element [trigger_name, trigger_handler]
+ * arrays, registered on this event, in the order in which they will be called.
+ * If an event name is passed, a table contains only one key which is passed
+ * argument, if there is an event with such a name, or returned table is empty,
+ * if the event does not exist.
+ */
+static int
+luaT_trigger_info(struct lua_State *L)
+{
+	if (lua_gettop(L) > 1)
+		luaL_error(L, "Usage: trigger.info([event])");
+	if (lua_gettop(L) == 0) {
+		lua_createtable(L, 0, 0);
+		bool ok = event_foreach(trigger_info_push_event, L);
+		assert(ok);
+		(void)ok;
+	} else {
+		const char *event_name = luaL_checkstring(L, 1);
+		struct event *event = event_get(event_name, false);
+		if (event == NULL || !event_has_triggers(event)) {
+			lua_createtable(L, 0, 0);
+			return 1;
+		}
+		lua_createtable(L, 0, 1);
+		trigger_info_push_event(event, L);
+	}
+	return 1;
+}
+
+static const char *event_trigger_iterator_typename =
+	"trigger.event_trigger_iterator";
+
+/**
+ * Gets event_trigger_iterator from Lua stack with type check.
+ */
+static inline struct event_trigger_iterator *
+luaT_check_event_trigger_iterator(struct lua_State *L, int idx)
+{
+	return luaL_checkudata(L, idx, event_trigger_iterator_typename);
+}
+
+/**
+ * Takes an iterator step.
+ */
+static int
+luaT_trigger_iterator_next(struct lua_State *L)
+{
+	struct event_trigger_iterator *it =
+		luaT_check_event_trigger_iterator(L, 1);
+	struct func_adapter *trigger = NULL;
+	const char *name = NULL;
+	if (event_trigger_iterator_next(it, &trigger, &name)) {
+		lua_pushstring(L, name);
+		func_adapter_lua_get_func(trigger, L);
+		return 2;
+	}
+	return 0;
+}
+
+/**
+ * Takes an iterator step of exhausted iterator.
+ */
+static int
+luaT_trigger_iterator_next_exhausted(struct lua_State *L)
+{
+	(void)L;
+	return 0;
+}
+
+/**
+ * Destroys an iterator.
+ */
+static int
+luaT_trigger_iterator_gc(struct lua_State *L)
+{
+	struct event_trigger_iterator *it =
+		luaT_check_event_trigger_iterator(L, 1);
+	event_trigger_iterator_destroy(it);
+	TRASH(it);
+	return 0;
+}
+
+/**
+ * Creates iterator over triggers of event with passed name.
+ * The iterator yields a pair [trigger_name, trigger_handler].
+ * Return next method of iterator and iterator itself.
+ */
+static int
+luaT_trigger_pairs(struct lua_State *L)
+{
+	if (lua_gettop(L) != 1)
+		luaL_error(L, "Usage: trigger.pairs(event)");
+	const char *event_name = luaL_checkstring(L, 1);
+	struct event *event = event_get(event_name, false);
+	if (event == NULL) {
+		lua_pushcfunction(L, luaT_trigger_iterator_next_exhausted);
+		return 1;
+	}
+	lua_pushcfunction(L, luaT_trigger_iterator_next);
+	struct event_trigger_iterator *it = lua_newuserdata(L, sizeof(*it));
+	event_trigger_iterator_create(it, event);
+	luaL_getmetatable(L, event_trigger_iterator_typename);
+	lua_setmetatable(L, -2);
+	return 2;
+}
+
+void
+box_lua_trigger_init(struct lua_State *L)
+{
+	const struct luaL_Reg module_funcs[] = {
+		{"set", luaT_trigger_set},
+		{"del", luaT_trigger_del},
+		{"call", luaT_trigger_call},
+		{"info", luaT_trigger_info},
+		{"pairs", luaT_trigger_pairs},
+		{NULL, NULL}
+	};
+	luaT_newmodule(L, "trigger", module_funcs);
+	lua_pop(L, 1);
+	const struct luaL_Reg trigger_iterator_methods[] = {
+		{"__gc", luaT_trigger_iterator_gc},
+		{NULL, NULL}
+	};
+	luaL_register_type(L, event_trigger_iterator_typename,
+			   trigger_iterator_methods);
+}
--- a/src/lib/core/event.c
+++ b/src/lib/core/event.c
@@ -1,0 +1,337 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include "event.h"
+
+#include <assert.h>
+
+#include "assoc.h"
+#include "diag.h"
+#include "trivia/util.h"
+#include "func_adapter.h"
+
+/** Registry of all events: name -> event. */
+static struct mh_strnptr_t *event_registry;
+
+/**
+ * A named node of the list, containing func_adapter. Every event_trigger is
+ * associated with an event. Since event_triggers have completely different
+ * call interface, they are not inherited from core triggers.
+ */
+struct event_trigger {
+	/** A link in a list of all registered event triggers. */
+	struct rlist link;
+	/** Trigger function. */
+	struct func_adapter *func;
+	/** Backlink to event. Needed for reference count of event. */
+	struct event *event;
+	/** Unique name of the trigger. */
+	char *name;
+	/** Trigger reference counter. */
+	uint32_t ref_count;
+	/** This flag is set when the trigger is deleted. */
+	bool is_deleted;
+};
+
+/**
+ * Creates event_trigger from func_adapter and name and binds it to passed
+ * event. Created trigger owns the adapter and will destroy it in its
+ * destructor. Passed name must be zero terminated and will be copied.
+ * Passed event must not be NULL and will be referenced by created trigger.
+ * Note that the function does not increment trigger_count field of the event.
+ */
+static struct event_trigger *
+event_trigger_new(struct func_adapter *func, struct event *event,
+		  const char *name)
+{
+	assert(name != NULL);
+	assert(event != NULL);
+	size_t name_len = strlen(name);
+	struct event_trigger *trigger =
+		xmalloc(sizeof(*trigger) + name_len + 1);
+	trigger->func = func;
+	trigger->name = (char *)(trigger + 1);
+	strlcpy(trigger->name, name, name_len + 1);
+	trigger->ref_count = 0;
+	trigger->is_deleted = false;
+	trigger->event = event;
+	event_ref(event);
+	rlist_create(&trigger->link);
+	return trigger;
+}
+
+/**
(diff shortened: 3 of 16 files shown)


changelogs/unreleased/gh-8656-trigger-registry.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Added the new module `trigger` to Lua. This module allows to manage and
+  call triggers stored in trigger registry (gh-8656).
```
p7nov, line 4:
```suggestion
* Introduced the new Lua module `trigger`. This module enables managing and
  calling triggers stored in trigger registry (gh-8656).
```

p7nov (approved):

src/lib/core/event.h
```diff
@@ -0,0 +1,268 @@
...
+	 */
+	bool is_deleted;
+};
+
+/**
+ * Creates trigger from func_adapter and name. Created trigger owns the adapter
+ * and will destroy it in its destructor. Passed name is copied.
+ */
+struct event_trigger *
+event_trigger_create(struct func_adapter *func, const char *name);
```
alyapunov, line 54: Should be `event_trigger_new` (please check for more)

src/lib/core/event.h
```diff
@@ -0,0 +1,268 @@
...
+		assert(rlist_empty(&event->triggers));
+		event_destroy(event);
+	}
+}
+
+/**
+ * Checks if the event has no triggers.
+ */
+static inline bool
+event_empty(struct event *event)
```
alyapunov, line 134: Should be `is_empty`.

src/lib/core/event.h
```diff
@@ -0,0 +1,268 @@
...
+ * Initializes iterator.
+ */
+void
+event_iterator_create(struct event_iterator *it, struct event *event);
+
+/**
+ * Returns the next event_trigger from the underlying list of triggers of event.
+ */
+void
+event_iterator_next(struct event_iterator *it, struct event_trigger **trg);
```
alyapunov, line 223: You should return `struct event_trigger *`

src/lib/core/event.c
```diff
@@ -0,0 +1,266 @@
...
+
+void
+event_iterator_next(struct event_iterator *it, struct event_trigger **trg)
+{
+	if (it->curr == &it->event->triggers) {
+		*trg = NULL;
+		return;
+	}
+	struct rlist *old = it->curr;
+	/*
```
alyapunov, line 170: Nit: please make one-line comment.

src/lib/core/event.c
```diff
@@ -0,0 +1,266 @@
...
+	 * We need to skip all the deleted triggers.
+	 */
+	do {
+		if (it->curr == NULL) {
```
alyapunov, line 174: Nit: perhaps it would be possible to avoid this if.

test/app-tap/trigger.test.lua
```diff
@@ -6,12 +6,11 @@ box.cfg{
     log = "tarantool.log"
 }
 
-local trigger = require('internal.trigger')
```
alyapunov, line 9: It seems to belong to one of previous commits.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,272 @@
...
+#include "core/event.h"
+#include <diag.h>
+#include <fiber.h>
+#include "lua/utils.h"
+
+/**
+ * Sets a trigger with passed name to the passed event.
+ * The first argument is event name, the second one is trigger name, the third
+ * one is trigger function.
+ * Returns new trigger handler.
```
alyapunov, line 17: We should specify what 'trigger handler' is.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,272 @@
...
+ * A callback for event_foreach function that calls a trigger.
+ * Since the only supported language is Lua, all the handlers are Lua objects.
+ */
+static int
+trigger_call_cb(struct event_trigger *trigger, void *arg)
+{
+	struct lua_State *L = (struct lua_State *)arg;
+	int top_svp = lua_gettop(L);
+	func_adapter_lua_get_func(trigger->func, L);
+	/*
```
alyapunov, line 73: There should be `narg` argument.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,272 @@
...
+	 * Lua stack.
+	 */
+	lua_State *L;
+};
+
+/**
+ * Appends an array [trigger_name, trigger_handler] to a pre-created array.
+ */
+static int
+luaT_trigger_info_push_trigger(struct event_trigger *trigger, void *arg)
```
alyapunov, line 122: Please remove luaT prefix.

alyapunov (changes requested):

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Done. Left it only for public functions and methods.

test/app-tap/trigger.test.lua
drewdzzz: Fixed.

drewdzzz:
Please note that I rewrote `event_foreach` function - now it uses `event_iterator`.

src/lib/core/event.h
```diff
@@ -0,0 +1,251 @@
...
+ * Creates trigger from func_adapter and name. Created trigger owns the adapter
+ * and will destroy it in its destructor. Passed name should be zero terminated
+ * and will be copied.
+ */
+struct event_trigger *
+event_trigger_new(struct func_adapter *func, const char *name);
+
+/**
+ * Destroys an event_trigger. Underlying func_adapter is destroyed,
+ * event_trigger is deallocated.
```
alyapunov, line 47: Nit: I would mention that this function is for internal use only and that `event_trigger_unref` must be used.

alyapunov (changes requested):
Almost perfect, just a pair of comments.

src/lib/core/event.h
```diff
@@ -0,0 +1,251 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#include "func_adapter.h"
```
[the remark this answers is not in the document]
alyapunov: Fixed by myself.

src/lua/init.c
```diff
@@ -134,6 +134,7 @@ extern char minifio_lua[],
 	error_lua[],
 	argparse_lua[],
 	iconv_lua[],
+	trigger_list_lua[],
```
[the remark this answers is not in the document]
alyapunov: Moved by myself.

src/lib/core/event.h
alyapunov: Added the comment by myself.

alyapunov (approved):

ylobankov (approved):
Tests LGTM

alyapunov (approved):

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+	if (lua_gettop(L) < 3 || !luaL_iscallable(L, 3))
+		luaL_typerror(L, 3, "callable");
+	/*
+	 * The following code is written in assumption no error will be thrown.
+	 */
+	struct event *event = event_get(event_name, true);
+	assert(event != NULL);
+	struct func_adapter *func = func_adapter_lua_create(L, 3);
+	event_reset_trigger(event, trigger_name, func);
+	func_adapter_lua_get_func(func, L);
```
locker, line 35: Nit: This line doesn't seem to be necessary because the function is already at the top of the stack.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+	struct event *event = event_get(event_name, false);
+	if (event == NULL)
+		return 0;
+	const char *trigger_name = luaL_checkstring(L, 2);
+	struct func_adapter *old = event_find_trigger(event, trigger_name);
+	if (old != NULL)
+		func_adapter_lua_get_func(old, L);
+	else
+		lua_pushnil(L);
+	event_reset_trigger(event, trigger_name, NULL);
```
locker, line 57: Nit: There's no point in calling `event_reset_trigger` if the trigger doesn't exist, i.e. it looks like this code could be simplified:

```C
struct func_adapter *func = event_find_trigger(event, trigger_name);
if (func == NULL)
        return 0;
func_adapter_lua_get_func(func, L);
event_reset_trigger(event, trigger_name, NULL);
return 1;
```

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+	event_trigger_iterator_create(&it, event);
+	struct func_adapter *trg = NULL;
+	const char *name = NULL;
+	int rc = 0;
+	while (rc == 0 && event_trigger_iterator_next(&it, &trg, &name)) {
+		lua_settop(L, top);
+		func_adapter_lua_get_func(trg, L);
+		for (int i = top - narg + 1; i <= top; ++i)
+			lua_pushvalue(L, i);
+		rc = luaT_call(L, narg, LUA_MULTRET);
```
locker, line 88: You don't use the result so you could just set the `nresults` argument to 0 instead of `LUA_MULTRET`. Then you wouldn't need to call `lua_settop` at each iteration.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+		idx++;
+		lua_createtable(L, 2, 0);
+		lua_pushstring(L, name);
+		lua_rawseti(L, -2, 1);
+		func_adapter_lua_get_func(trg, L);
+		lua_rawseti(L, -2, 2);
+		lua_rawseti(L, -2, idx);
+	}
+	lua_setfield(L, -2, event->name);
+	return true;
```
locker, line 120: `event_trigger_iterator_destroy` seems to be missing.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+			lua_createtable(L, 0, 0);
+			return 1;
+		}
+		lua_createtable(L, 0, 1);
+		trigger_info_push_event(event, L);
+	}
+	return 1;
+}
+
+const char *event_trigger_iterator_typename = "trigger.event_trigger_iterator";
```
locker, line 152: Should be static.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Lua-iterator over triggers of the event.
+ */
+struct lua_event_trigger_iterator {
+	/** An iterator over triggers of the event. */
+	struct event_trigger_iterator it;
+};
```
locker, line 160: Why do you need this structure? Can't you use `event_trigger_iterator` directly?

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+ * The iterator yields a pair [trigger_name, trigger_handler].
+ * Return next method of iterator and iterator itself.
+ */
+static int
+luaT_trigger_pairs(struct lua_State *L)
+{
+	const char *event_name = luaL_checkstring(L, 1);
+	struct event *event = event_get(event_name, false);
+	if (event == NULL)
+		return 0;
```
locker, line 213: According to the coveralls report, this line isn't covered by any tests. Please add a test for this case.

locker (changes requested):

src/box/lua/trigger.c
drewdzzz: Added the check of pairs on empty event.

src/lib/core/event.h
```diff
@@ -0,0 +1,160 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#include <assert.h>
+#include <stdbool.h>
+
+#include "func_adapter.h"
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/lib/core/event.c
```diff
@@ -0,0 +1,336 @@
...
+
+	/*
+	 * If the only thing that holds the event is its trigger list,
+	 * the reference counter will reach zero when the list will be cleared
+	 * and the destructor will be called again. Let's reference the event
+	 * in order to prevent such situation.
+	 */
+	event_ref(event);
+	struct event_trigger *trigger, *tmp;
+	rlist_foreach_entry_safe(trigger, &event->triggers, link, tmp) {
```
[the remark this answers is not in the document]
drewdzzz: Moved triggers deletion to `event_free`.

src/lib/core/event.c
```diff
@@ -0,0 +1,336 @@
...
+
+/**
+ * Finds internal event_trigger entity by name in the event.
+ * All arguments must not be NULL.
+ */
+static struct event_trigger *
+event_find_trigger_internal(struct event *event, const char *name)
+{
+	if (rlist_empty(&event->triggers))
+		return NULL;
```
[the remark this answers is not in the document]
drewdzzz: I thought that `rlist_foreach_empty` can lead to UB with an empty list because we will have a pointer to non-allocated memory in initialize statement of the `for` loop. But we call this function with empty lists everywhere, so I removed this check.

src/lib/core/event.c
```diff
@@ -0,0 +1,336 @@
...
+{
+	assert(trigger != NULL);
+	assert(name != NULL);
+
+	*trigger = NULL;
+	*name = NULL;
+	/* Iterator is exhausted - return. */
+	if (it->curr == NULL)
+		return false;
+	struct event_trigger *trg;
```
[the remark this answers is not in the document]
drewdzzz: Renamed all `trg` variables to `trigger`.

src/lib/core/event.c
```diff
@@ -0,0 +1,336 @@
...
+	event_registry = mh_strnptr_new();
+}
+
+void
+event_free(void)
+{
+	assert(event_registry != NULL);
+	struct mh_strnptr_t *h = event_registry;
+	if (h == NULL)
+		return;
```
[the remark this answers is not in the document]
drewdzzz: Removed the check.

test/unit/event.c
```diff
@@ -0,0 +1,389 @@
...
+	test_event_trigger_iterator_stability();
+	return check_plan();
+}
+
+int
+main(void)
+{
+	srand(time(NULL));
+	memory_init();
+	fiber_init(fiber_c_invoke);
```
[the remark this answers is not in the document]
drewdzzz: Removed them.

src/box/lua/trigger.c
drewdzzz: Actually, no. I didn't check number of arguments, so if I call `trigger.set('event', 'trigger', handler, 'redundant')`, `'redundant'` will be at the top. So I added the check to all 4 public methods of the module, covered it with the test and removed this redundant `func_adapter_lua_get_func`.

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Added `static` qualifier.

src/box/lua/trigger.c
drewdzzz: Removed this structure.

src/lib/core/event.c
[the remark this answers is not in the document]
drewdzzz: Also, added the unit test to cover the case of deletion non-empty event registry.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+		{"call", luaT_trigger_call},
+		{"info", luaT_trigger_info},
+		{"pairs", luaT_trigger_pairs},
+		{NULL, NULL}
+	};
+	luaT_newmodule(L, "trigger", module_funcs);
+	lua_pop(L, 1);
+	const struct luaL_Reg trigger_iterator_methods[] = {
+		{"__gc", luaT_trigger_iterator_gc},
+		{"next", luaT_trigger_iterator_next},
```
locker, line 237: `next` is never used. Let's remove this export.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+		{ NULL, NULL }
```
locker, line 238: Nit: Let's remove the spaces for consistency with other entries in this and the previous array: `{NULL, NULL}`

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+ * one is new trigger handler - it can be a Lua function or another callable
+ * object. If there is an already registered trigger with such name in the
+ * event, it is replaced with new trigger.
+ * Returns new trigger handler (which was passed as the third argument).
+ */
+static int
+luaT_trigger_set(struct lua_State *L)
+{
+	if (lua_gettop(L) != 3)
+		luaL_error(L, "expected 3 arguments");
```
locker, line 25: This is a bad error message. Please replace with something like `Usage: trigger.set(event, trigger, function)` to hint the user about what they should do. Same comment for other similar error messages.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+ * arguments. Execution is stopped by a first exception.
+ * First argument must be a string - all the other arguments will be passed
+ * to the triggers without any processing or copying.
+ * Returns no values on success. If one of the triggers threw an error, it is
+ * raised again.
+ */
+static int
+luaT_trigger_call(struct lua_State *L)
+{
+	const char *event_name = luaL_checkstring(L, 1);
```
locker, line 75: It'd be nice to have the argument count check here as well.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+static const char *event_trigger_iterator_typename =
+	"trigger.event_trigger_iterator";
+
+/**
+ * Gets lua_event_trigger_iterator from Lua stack with type check.
+ */
+static inline struct event_trigger_iterator *
+lua_check_event_trigger_iterator(struct lua_State *L, int idx)
```
locker, line 164: Nit: There's no `lua_event_trigger_iterator` structure anymore so the comment is wrong. Also, better use `luaT_` prefix for consistency with other functions in this file.

src/box/lua/trigger.c
```diff
@@ -0,0 +1,242 @@
...
+ */
+static int
+luaT_trigger_pairs(struct lua_State *L)
+{
+	if (lua_gettop(L) != 1)
+		luaL_error(L, "expected 1 argument");
+	const char *event_name = luaL_checkstring(L, 1);
+	struct event *event = event_get(event_name, false);
+	if (event == NULL)
+		return 0;
```
locker, line 213: With this code iteration over triggers for non-existing event will fail with:

```
tarantool> for _, i in trigger.pairs('foobar') do end
---
- error: '[string "for _, i in trigger.pairs(''foobar'') do end"]:1: attempt to call
    a nil value'
...
```

It looks unexpected. Given that events are transient (removed when there's no triggers), I think we should return an empty iterator here so that `pairs` works properly in any case.

test/app-luatest/trigger_module_test.lua
```diff
@@ -0,0 +1,498 @@
...
+            errmsg, trigger.del, valid_name, invalid_name)
+        t.assert_error_msg_contains(
+            errmsg, trigger.del, invalid_name, valid_name)
+        t.assert_error_msg_contains(errmsg, trigger.call, invalid_name)
+        t.assert_error_msg_contains(errmsg, trigger.pairs, invalid_name)
+        t.assert_error_msg_contains(errmsg, trigger.info, invalid_name)
+    end
+    errmsg = "expected .* argument.*"
+    t.assert_error_msg_matches(errmsg, trigger.set, 'event', 'trigger',
+        handler, 'redundant')
```
locker, line 383: Please check all error messages properly with `assert_error_msg_equals` or `assert_error_msg_content_equals`, not with `assert_error_msg_matches` or `assert_error_msg_contains`.

test/app-luatest/trigger_module_test.lua
```diff
@@ -0,0 +1,498 @@
...
+end
+
+-- Check that pairs is successfully stopped after all triggers registered
+-- on the event are dropped
+g.test_clear_event_during_pairs = function()
+    local trigger = require('trigger')
+
+    local event = 'event'
+    local trigger_num = 100
+    local break_point = math.random(5, 95)
```
locker, line 316: Random tests are difficult to debug because it isn't easy to figure out which seed caused a failure. Can you rewrite the tests in a deterministic way?

test/app-luatest/trigger_module_test.lua
```diff
@@ -0,0 +1,498 @@
...
+-- Check that a trigger deleted during pairs does not break anything and
+-- isn't yielded by pairs after its deletion.
+g.test_del_during_pairs = function()
+    local trigger = require('trigger')
+
+    local event = 'event'
+    local trigger_num = 100
+    local break_point = math.random(5, 95)
+
+    t.assert_equals(trigger.pairs(event), nil)
```
locker, line 267: The test fails here if called repeatedly:

```
$ luatest -v test/app-luatest/trigger_module_test.lua -r 2
Tarantool version is 3.0.0-alpha1-136-ga1596bdc3b74
Started on Wed Aug 23 11:48:30 2023
    app-luatest.trigger_module.test_simple ... (0.000s) Ok
    app-luatest.trigger_module.test_order_and_shared_argument ... (0.000s) Ok
    app-luatest.trigger_module.test_shared_handler ... (0.000s) Ok
    app-luatest.trigger_module.test_exceptions ... (0.000s) Ok
    app-luatest.trigger_module.test_any_callable_trigger ... (0.000s) Ok
    app-luatest.trigger_module.test_trigger_clear_from_trigger ... (0.000s) Ok
    app-luatest.trigger_module.test_pairs ... (0.000s) Ok
    app-luatest.trigger_module.test_del_during_pairs ... (0.001s) fail
...ntool/tarantool/test/app-luatest/trigger_module_test.lua:267: expected: nil, actual: function: 0x2d368310
    app-luatest.trigger_module.test_clear_event_during_pairs ... (0.001s) Ok
    app-luatest.trigger_module.test_invalid_args ... (0.001s) Ok
    app-luatest.trigger_module.test_nil_args ... (0.000s) Ok
    app-luatest.trigger_module.test_info ... (0.001s) Ok
    app-luatest.trigger_module.test_info_no_empty_events ... (0.000s) Ok
=========================================================

Failed tests:
-------------

1) app-luatest.trigger_module.test_del_during_pairs
...ntool/tarantool/test/app-luatest/trigger_module_test.lua:267: expected: nil, actual: function: 0x2d368310
stack traceback:
        ...ntool/tarantool/test/app-luatest/trigger_module_test.lua:267: in function 'app-luatest.trigger_module.test_del_during_pairs'
        ...
        [C]: in function 'xpcall'

Ran 13 tests in 0.006 seconds, 12 succeeded, 1 failed

=========================================================
Failed tests:

app-luatest.trigger_module.test_del_during_pairs
```

It also fails with `--shuffle all` occasionally. Please fix.

locker (changes requested):

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Done.

src/box/lua/trigger.c
drewdzzz: Fixed the comment and changed the prefix.

src/box/lua/trigger.c
drewdzzz: I added a new function that returns nothing - it is returned if an event does not exist.

test/app-luatest/trigger_module_test.lua
drewdzzz: Done.

test/app-luatest/trigger_module_test.lua
drewdzzz: 100 triggers is overkill so I reduced this number to 10, but I did not removed random form the tests.

test/app-luatest/trigger_module_test.lua
drewdzzz: I think that the iterator didn't get garbage collected, so it referenced the event and it was not deleted. That is why a new iterator was created from a non-NULL event. Now I use a unique name for the test to avoid such a situation. And this line was removed - now I check that pairs still work correctly but do not make iterations.

locker (approved):

locker:
vshard and go-tarantool integration tests fail. I tried to restart them a few times but it didn't help. PTAL.

drewdzzz:
`vshard` uses `inernal.trigger` module for its triggers, so 00e0e79b8ab76982efa0f28dc18870e3c3b6a8a0 broke integration. The commit isn't important, so I removed it from the patch.

PR #8623, opened 2023-05-03
box: do not allow to drop system spaces

### LINKED ISSUE #5279: I can drop one system space
I have Ubuntu 18.04. I have Tarantool 2.5, pulled from source today.

If I try to drop a memtx system space, I get an error
Can't drop the primary key in a system space
But I can drop a local blackhole system space which has no primary key:
box.execute([[DROP TABLE "_vinyl_deferred_delete";]])
It works. The table is dropped.
I think this is an oversight, since I imagine that the intent was
that it would be illegal to drop any system space.

mkostoevr:
The patch adds a new check to the _space on_replace trigger failing on attempt to drop a system table.

Closes #5279

NO_DOC=bugfix

**PROBLEMS**
1. The check duplicates the one in the _index on_replace trigger: [src/box/alter.cc](https://github.com/tarantool/tarantool/blob/93b3bba9bf907f9199f31d4071b0e3a5616a243a/src/box/alter.cc#L2335). But if we remove it then we gonna have to fix a lot of tests checking system space drop, because removal of an entry from the _space space happens only after index drop: [src/box/lua/schema.lua](https://github.com/tarantool/tarantool/blob/1e02e0506f89f01d386fdaaed01ca04c54943de5/src/box/lua/schema.lua#L911), so the removal of the check invalidates the system space we attempted to drop making the lua code remove indexes of the space.

**OTHER POSSIBLE SOLUTIONS**
1. Add a dummy index to blackhole spaces so their primary index drop will trigger the existing check.

### DIFF
--- a/changelogs/unreleased/gh-5279-system-space-removal.md
+++ b/changelogs/unreleased/gh-5279-system-space-removal.md
@@ -1,0 +1,3 @@
+## bugfix/box
+
+- Fixed the ability to drop `_vinyl_deferred_delete` system space (gh-5279).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -2099,6 +2099,13 @@ on_replace_dd_space(struct trigger * /* trigger */, void *event)
 		/* Check whether old_space is used somewhere. */
 		if (space_check_pinned(old_space) != 0)
 			return -1;
+		/* One can't just remove a system space. */
+		if (space_is_system(old_space)) {
+			diag_set(ClientError, ER_DROP_SPACE,
+				 space_name(old_space),
+				 "the space is a system space");
+			return -1;
+		}
 		/**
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/gh-5279-system-space-removal.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/box
+
+- Fix ability to drop `_vinyl_deferred_delete` system space.
```
xuniq, line 3:
```suggestion
- Fixed the ability to drop `_vinyl_deferred_delete` system space.
```

xuniq (approved):

sergepetrenko (approved):
Thanks! LGTM.

alyapunov:
Cherry-picked to 2.11.

Issue #8147, opened 2023-01-12
Provide names in IPROTO responses


alyapunov:
Now if a user wants to know selected tuples' format, they have to select spaces format manually and maintain its relevance using SCHEMA_VERSION mechanism in IPROTO. Note that this approach is durable only for direct IPROTO DML, such as replaces or selects, while in general doesn't work for IPROTO calls.

Let's provide an optional field(s) in response that describes format of tuple in the response. We should discuss the details - what exactly must be placed in that field(s) and how optional it should be.

olegrok:
After implementing this feature please make sure that it won't lead to performance degradation. (Try to test several projects/modules).

Once we faced an issue because netbox.self returned cdata tuple in contrast of usual netbox that automatically converted response to table. https://github.com/tarantool/tdg2/issues/1835#issuecomment-1490602427

PR #8631, opened 2023-05-07
build: decouple box library from Tarantool binary

CuriousGeorgiy:
Needed for #8147


CuriousGeorgiy:
@locker checkpatch linter errors look FP, ignoring them.

src/box/box.cc
```diff
@@ -240,6 +248,18 @@ static char box_feedback_host[BOX_FEEDBACK_HOST_MAX];
...
+{
+	return ev_monotonic_now(loop()) - (ev_tstamp)tarantool_start_time;
+}
+
+sigint_cb_t __attribute__((weak))
+set_sigint_cb(sigint_cb_t new_sigint_cb)
+{
+	return new_sigint_cb;
+}
+
```
locker, line 262: I don't understand why these changes are needed and how they're connected with #8147.

test/unit/core_test_utils.c
```diff
@@ -32,6 +32,7 @@
 /**
  * Mock function to resolve circular dependencies in unit tests
  */
-void cord_on_yield(void)
+void __attribute__((weak))
+cord_on_yield(void)
```
locker, line 36: This doesn't look right. I'd expect the core version of `cord_on_yield` to be defined as weak. OTOH I'm not sure about possible implications of such a change so I'd avoid weak linkage altogether if possible.

Anyway, it's unclear what exactly you're trying to fix here. Is there a build failure on some platform?

src/box/box.h
```diff
@@ -91,12 +91,34 @@ extern double txn_timeout_default;
...
+sigint_cb_t __attribute__((weak))
+set_sigint_cb(sigint_cb_t new_sigint_cb);
+
 /*
  * Initialize box library
  * @throws C++ exception
  */
 void
-box_init(void);
+box_init(const char *tarantool_bin);
```
alyapunov, line 121: There's must be a comment about new argument.

alyapunov (changes requested):
I don't like current code about sigint, but actually I don't like the proposed fix too.
Is it possible to make behavior of tarantool and unit/perf test the same? It would be great to avoid weak symbols and tricky builds.

CuriousGeorgiy:
> Is it possible to make behavior of tarantool and unit/perf test the same?

Is this a rhetorical question? If we want independent libraries but one library depends on a symbol from another one, the only way is using weak functions or linking mock symbols.

PR #8640, opened 2023-05-10
sql: fix invalid negation.

kobrineli:
This patch fixes the check for invalid negation, which was found by [sydr-fuzz](https://github.com/ispras/oss-sydr-fuzz) in `box/sql/expr.c:3202`.

The error occurred because the value of `value` var can be equal to `INT64_MIN` (which is equal to `(uint64_t) INT64_MAX + 1`), and it won't be handled by corresponding check. To fix this it is enough just to change the check to `>=` instead of `>` to parse this value as well.

How to reproduce:

1. Build and run docker container from [here](https://github.com/ispras/oss-sydr-fuzz/tree/master/projects/tarantool):
 
        sudo docker build -t oss-sydr-fuzz-tarantool
        sudo docker run --rm --privileged -v `pwd`:/fuzz -it oss-sydr-fuzz-tarantool /bin/bash

2. Run target `/sql_fuzzer` on this input
[tarantool-neg.txt](https://github.com/tarantool/tarantool/files/11444190/tarantool-neg.txt)

        /sql_fuzzer tarantool-neg.txt

3. You will see the following output:

        tarantool/src/box/sql/expr.c:3202:11: runtime error: negation of -9223372036854775808 cannot be represented in type 'int64_t' (aka 'long'); cast to an unsigned type to negate this value to itself
        SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tarantool/src/box/sql/expr.c:3202:11 in

NO_DOC=refactoring
NO_CHANGELOG=refactoring
NO_TEST=refactoring

### DIFF
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -3198,7 +3198,12 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
 			return;
 		}
 	}
-	if (is_neg)
+
+	/*
+	 * We don't need to negate INT64_MIN value because it's negation is
+	 * equal to it.
+	 */
+	if (is_neg && value != INT64_MIN)
 		value = -value;
 	sqlVdbeAddOp4Dup8(v, OP_Int64, 0, mem, 0, (u8 *) &value,
 			  is_neg ? P4_INT64 : P4_UINT64);


SweetVishnya (approved):

igormunkin:
@kobrineli, thanks for the patch! I would like to mention our [contributing guides](https://github.com/tarantool/tarantool/blob/master/CONTRIBUTING.md) and relevant points your patch doesn't follow:
1. CI is red, please check what's going wrong. If the failure is not related to your changes, please mention it explicitly in PR, so maintainer is informed.
2. The commit message doesn't follow the [guidelines](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message).
3. Test case for your change is missing.

kobrineli:
@igormunkin 

Hi! I've changed the commit and PR descriptions according to guides. The changelog is not presented, because there is no issue which is closed by the fix.

Also, I don't know how to add test correctly and to which part, because the error was found by fuzzing `sql_fuzzer` fuzz target and the attached input is the input file for this target. What should I do with it?

ImeevMA (changes requested):

src/box/sql/expr.c
```diff
@@ -3192,7 +3192,7 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
 		size_t len = strlen(z);
 		bool unused;
 		if (sql_atoi64(z, &value, &unused, len) != 0 ||
-		    (is_neg && (uint64_t) value > (uint64_t) INT64_MAX + 1)) {
+		    (is_neg && (uint64_t)value >= (uint64_t)INT64_MAX + 1)) {
```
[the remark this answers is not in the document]
kobrineli: @ImeevMA, the error is not in conversions. The error is lower in `expr.c:3202`

        if (is_neg)
            value = -value;

Here, if the `value` equals to INT64_MIN, `-value` will be INT64_MIN as well (due to the ranges of integer numbers).
Is this correct, that the negation of value will be equal to the same value? I suppose no.

So the patch aimed to prevent the case, when `is_neg` is `true`, and `value` is INT64_MIN, which cannot be negated correctly.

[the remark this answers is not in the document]
ImeevMA: Here is actually just a bad code. Here we give value `9223372036854775808` to function `sql_atoi64()` (since `-` was processed before; we got into `expr_code_int()` from `case TK_UMINUS:`) which returns unsigned value `9223372036854775808`. Next we should negate the value, however, since we use `int64_t` instead of `uint64_t`, value is already `-9223372036854775808`.

The proper solution here would be to use `uint64_t value;` instead of `int64_t value;` and use `strtoull()` everywhere after which check the value no more than `INT64_MAX + 1`.

src/box/sql/expr.c
```diff
@@ -3192,13 +3192,17 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
 		size_t len = strlen(z);
 		bool unused;
 		if (sql_atoi64(z, &value, &unused, len) != 0 ||
-		    (is_neg && (uint64_t) value > (uint64_t) INT64_MAX + 1)) {
+		    (is_neg && (uint64_t)value > (uint64_t)INT64_MAX + 1)) {
```
ImeevMA, line 3195: Since this patch only fixes the mentioned issue, I don't see the need for this refactoring.

src/box/sql/expr.c
```diff
@@ -3192,13 +3192,17 @@ expr_code_int(struct Parse *parse, struct Expr *expr, bool is_neg,
...
 			diag_set(ClientError, ER_INT_LITERAL_MAX, sign, z);
 			parse->is_aborted = true;
 			return;
 		}
 	}
-	if (is_neg)
+
+	/* We don't need to negate INT64_MIN value because it's
+	   negation is equal to it.
+	*/
```
ImeevMA, line 3204: Please move `*/` to the previous line. You can also make the first line a little longer, up to 80 characters.

ImeevMA (commented):
Hi! Thank you for the fixes! I have a couple of small comments, see them below. Also, the line length for the commit message (excluding the header) is 72 characters, so you can make the lines a bit longer.

ImeevMA (approved):
Thanks for the fixes!

ligurio:
@kobrineli still failed checkpath:

```
ERROR:BLOCK_COMMENT_STYLE: Block comments use a trailing */ on a separate line
#28: FILE: src/box/sql/expr.c:3203:
+	 * equal to it. */
```

src/box/sql/expr.c
ImeevMA: Sorry, this suggestion seems to be wrong.

Totktonada: Multiline comments are started from the `/*` (or `/**`) line.

Correct:

```
/*
 * Foo.
 * 
 * Bar. Baz. Fiz.
 */
```

Wrong:

```
/* Foo.
 * 
 * Bar. Baz. Fiz.
 */
```

https://lkml.org/lkml/2016/7/8/625

kobrineli: @Totktonada 
Fixed

igormunkin:
> @igormunkin Hi! Will the PR be merged?

@kobrineli, thanks for your efforts and patience! Дело сделано.

PR #8645, opened 2023-05-11
iproto: send `IPROTO_WATCH` sync number in `IPROTO_EVENT` packet

### LINKED ISSUE #8393: WATCH doesn't maintain SYNC
Hi, there!

We realised a subscribe mechanism - IPROTO_WATCH.

Its flow is: `client[watch] -> server[event] -> client[watch] ->... client[unwatch]`

Most clients use SYNC as a way to connect server[packet] with a client (fiber) that waits for server[packet].

But the protocol extension breaks the scheme. Yes, a client don't have to use a fiber for WATCH-request, but it can.

So, it would be nice if EVENT message contains SYNC=[latest.request.client.sync]

|dir|→|←|→|←|
|-|-|-|-|-|
|**flow**|WATCH|EVENT|WATCH|EVENT|
|**SYNC**|123|123|345|345|

locker:
Initially the sync number sent by a client in an `IPROTO_WATCH` request was ignored and `IPROTO_EVENT` packet didn't have a sync number. There were complaints about it from users so we consider this to be a bug. Now the server sends the same sync number in an `IPROTO_EVENT` packet as the one sent by the client in the last corresponding `IPROTO_WATCH` request.

We don't use this functionality in net.box (sync number is always 0 for all watch/event packets), but other clients may actually use it.

Closes #8393

### DIFF
--- a/changelogs/unreleased/gh-8393-iproto-watch-sync.md
+++ b/changelogs/unreleased/gh-8393-iproto-watch-sync.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* An `IPROTO_EVENT` packet now has the same sync number as the last
+  corresponding `IPROTO_WATCH` request (gh-8393).
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -2360,8 +2360,9 @@ tx_process_id(struct iproto_connection *con, const struct id_request *id)
 	con->session->meta.features = id->features;
 }
 
+/** Callback passed to session_watch. */
 static void
-iproto_session_notify(struct session *session,
+iproto_session_notify(struct session *session, uint64_t sync,
 		      const char *key, size_t key_len,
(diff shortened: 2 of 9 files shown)


p7nov (approved):

alyapunov (approved):

locker:

p7nov:
Do we need a doc update? Perhaps, packet structures here: [iproto > Events and subscriptions](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/events/).

locker:
> Do we need a doc update? Perhaps, packet structures here: [iproto > Events and subscriptions](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/events/).

Yes. I forgot that the doc bot doesn't work anymore. Filed the ticket manually: https://github.com/tarantool/doc/issues/3458

PR #8651, opened 2023-05-12
box: add hostname to box.info

### LINKED ISSUE #8605: Add box.info.hostname
The hostname is a useful piece of information to supplement various instance state reports.

It looks convenient to have the hostname information together with other instance related details -- in `box.info`.

Note 1: Should be accessible as `box.info.hostname` and as `box.info().hostname` (for iproto calls) -- like other `box.info` fields.

Note 2: As regarding very long hostnames -- I would report `nil` (meaning 'cannot determine'), not a trimmed hostname. `gethostname()` on Linux reports `ENAMETOOLONG` in the case, but POSIX.1 says nothing about errors. Need to check macOS and FreeBSD behavior in this regard.

Note 3: The hostname may changes, so we should obtain it when requested, not cache at startup.

<details>
<summary>moonlibs/kit</summary>

https://github.com/moonlibs/kit/blob/7b46d4a8aba14944f1c924f16938dc2c8200bd8b/kit/1.lua#L50-L68
</details>

<details>
<summary>My snippet</summary>

```lua
-- utils/hostname.lua
local ffi = require('ffi')
local errno = require('errno')

-- Ignore errors to make the code reloadable.
pcall(ffi.cdef, 'int gethostname(char *name, size_t len);')

local function get()
    local buf_size = 256
    local buf = ffi.new('char[?]', buf_size)
    local rc = ffi.C.gethostname(buf, buf_size)
    if rc ~= 0 then
        error(('gethostname: %s'):format(errno.strerror()), 0)
    end
    return ffi.string(buf)
end

return {
    get = get,
}
```
</details>

Lord-KA:
Hostname is a useful piece of information in state reports. So it was decided to add it to box.info.

Hostname is obtained on requested and is not cached.

Closes #8605

@TarantoolBot document
Title: Add hostname to box.info

This patch adds hostname to box.info, it can be useful e.g. to supplement various instance state reports. It is not cached and is requested on each call.

### DIFF
--- a/changelogs/unreleased/gh-8605-add-gethostname.md
+++ b/changelogs/unreleased/gh-8605-add-gethostname.md
@@ -1,0 +1,3 @@
+## feature/box
+
+* Added a new `box.info` parameter `hostname` (gh-8605).
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -31,6 +31,7 @@
 #include "lua/info.h"
 
 #include <ctype.h> /* tolower() */
+#include <unistd.h>
 
 #include <lua.h>
 #include <lauxlib.h>
@@ -68,6 +69,18 @@
  */
 static bool box_info_cluster_new_meaning = true;
 TWEAK_BOOL(box_info_cluster_new_meaning);
+
+/**
+ * Known upper limits for a hostname (without a zero-terminating
+ * byte):
+ *
+ * sysconf(_SC_HOST_NAME_MAX) == 64 on Linux.
+ * sysconf(_SC_HOST_NAME_MAX) == 255 on macOS.
+ * sysconf(_SC_HOST_NAME_MAX) == 255 on BSD.
+ *
+ * The constant value is used to simplify the code.
+ */
+enum { TT_HOST_NAME_MAX = 255 };
 
 static inline void
 lbox_push_replication_error_message(struct lua_State *L, struct error *e,
@@ -701,6 +714,22 @@ lbox_schema_version(struct lua_State *L)
 	return 1;
 }
 
+/** gethostname() Lua interface inside box.info. */
+static int
+lbox_info_hostname(struct lua_State *L)
+{
+	char buffer[TT_HOST_NAME_MAX + 1];
+	int rc = gethostname(buffer, sizeof(buffer));
+	if (rc != 0) {
+		say_warn_ratelimited("failed to get hostname: %s",
+				     tt_strerror(errno));
+		lua_pushnil(L);
+		return 1;
+	}
+	lua_pushstring(L, buffer);
+	return 1;
+}
+
 static const struct luaL_Reg lbox_info_dynamic_meta[] = {
 	{"id", lbox_info_id},
 	{"uuid", lbox_info_uuid},
@@ -725,6 +754,7 @@ static const struct luaL_Reg lbox_info_dynamic_meta[] = {
 	{"election", lbox_info_election},
 	{"synchro", lbox_info_synchro},
 	{"schema_version", lbox_schema_version},
+	{"hostname", lbox_info_hostname},
 	{NULL, NULL}
 };
 
--- a/test/box-luatest/gh_8605_add_box_info_hostname_test.lua
+++ b/test/box-luatest/gh_8605_add_box_info_hostname_test.lua
@@ -1,0 +1,21 @@
+local popen = require('popen')
+local t = require('luatest')
+
+local g = t.group()
+
+g.test_hostname_result = function()
+    local ph = popen.shell("hostname", 'r')
+
+    local exp_status = {
+        state = popen.state.EXITED,
+        exit_code = 0,
+    }
+    local status = ph:wait()
+    t.skip_if(status.exit_code == 127, "no hostname on host")
+    t.assert_equals(status, exp_status, 'verify process status')
+
+    local hostname = ph:read():rstrip()
+    t.assert_equals(hostname, box.info.hostname, 'verify hostname')
+
+    ph:close()
+end
--- a/test/box/info.test.lua
+++ b/test/box/info.test.lua
@@ -17,6 +17,7 @@ box.info.status
 string.len(box.info.uptime) > 0
 string.match(box.info.uptime, '^[1-9][0-9]*$') ~= nil
 box.info.replicaset.uuid == box.space._schema:get{'replicaset_uuid'}[2]
+type(box.info.hostname) == 'string'
 t = {}
 for k, _ in pairs(box.info()) do table.insert(t, k) end
 table.sort(t)
--- a/test/box/info.result
+++ b/test/box/info.result
@@ -63,6 +63,10 @@ box.info.replicaset.uuid == box.space._schema:get{'replicaset_uuid'}[2]
 ---
 - true
 ...
+type(box.info.hostname) == 'string'
+---
+- true
+...
 t = {}
 ---
 ...
@@ -77,6 +81,7 @@ t
 - - cluster
   - election
   - gc
+  - hostname
   - id
   - listen
   - lsn


changelogs/unreleased/gh-8605-add-gethostname.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
+
+* Added hostname to box.info, it is obtained on request (gh-8605).
```
p7nov, line 3:
```suggestion
* Added a new `box.info` parameter `hostname` (gh-8605).
```

Though I don't know if "It is obtained on request" is really important for a brief changelog entry. You can leave it if it is. Just make is a full separate sentence.

p7nov (approved):

src/box/lua/info.c
```diff
@@ -650,6 +651,21 @@ lbox_schema_version(struct lua_State *L)
 	return 1;
 }
 
+/* gethostname() Lua interface inside box.info. */
+static int
+lbox_info_hostname(struct lua_State *L)
+{
+	char buffer[_SC_HOST_NAME_MAX + 1];
```
Totktonada, line 658: It is not a value of max hostname length, it is ID of the corresponding parameter in sysconf.

```sh
$ cc --std=c89 -Wall -Wextra -x c <(echo -e '#include <stdio.h>\n#include<unistd.h>\n\nint main() { printf("_SC_HOST_NAME_MAX: %d\\n", _SC_HOST_NAME_MAX); return 0; }') && ./a.out; rm a.out

_SC_HOST_NAME_MAX: 180
```

/usr/include/bits/confname.h:

```c
/* Values for the NAME argument to `pathconf' and `fpathconf'.  */
enum
  {
    <...>
    _SC_HOST_NAME_MAX,
#define _SC_HOST_NAME_MAX               _SC_HOST_NAME_MAX
```

changelogs/unreleased/gh-8605-add-gethostname.md
Lord-KA: Agreed.

src/box/lua/info.c
Lord-KA: As expected, the mac has 255 default:
```
sysconf(_SC_HOST_NAME_MAX): 255
```

Lord-KA: Same for FreeBSD:
```
sysconf(_SC_HOST_NAME_MAX): 255
```

changelogs/unreleased/gh-8605-add-gethostname.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
```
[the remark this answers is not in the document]
Lord-KA: Oh, I didn't recheck my source. Ok, removed it.

src/box/lua/info.c
Lord-KA: Added `enum` constant.

src/box/lua/info.c
```diff
@@ -58,6 +59,18 @@
...
+ * Known upper limits for a hostname (without a zero-terminating
+ * byte):
+ *
+ * sysconf(_SC_HOST_NAME_MAX) == 64 on Linux.
+ * sysconf(_SC_HOST_NAME_MAX) == 255 on macOS.
+ * sysconf(_SC_HOST_NAME_MAX) == 255 on BSD.
+ *
+ * The constant value is used to simplify the code.
+ */
+enum { TT_HOST_NAME_MAX = 255 };
```
ligurio, line 72: I would use a macro, not a constant number, because:

> Host names are limited to MAXHOSTNAMELEN (from <sys/param.h>) characters, currently 256. This includes the terminating NUL character. Note that the corresponding POSIX definition HOST_NAME_MAX in <limits.h> does [not](https://man.openbsd.org/sethostname#not) include the terminating NUL character. [^1]

[^1]: https://man.openbsd.org/sethostname

test/box/info.test.lua
```diff
@@ -17,6 +17,8 @@ box.info.status
 string.len(box.info.uptime) > 0
 string.match(box.info.uptime, '^[1-9][0-9]*$') ~= nil
 box.info.cluster.uuid == box.space._schema:get{'cluster'}[2]
+type(box.info.hostname) == 'string'
```
ligurio, line 20: After this `os.execute("hostname") ~= ''` to make sure that real hostname is not an empty string.

ligurio (commented):
See my comments.

test/box/info.test.lua
Lord-KA: It seems like `os.execute()` doesn't work this way:
```
tarantool> os.execute('hostname') == 'gkashkin-vps'
gkashkin-vps
---
- false
...

tarantool> os.execute('hostname')
gkashkin-vps
---
- 0
...

tarantool> os.execute('hostname') == 0
gkashkin-vps
---
- true
...
```

Lord-KA: (And I check that it is not empty string the line below.)

src/box/lua/info.c
Totktonada: There is no very clear definition in POSIX: is the `HOST_NAME_MAX` a constant/macro or a sysconf value? Why it is in the curly brackets? What header should define it?

Why `limits.h` on Linux actually defines the constant only if `__USE_POSIX` is defined?

Why `MAXHOSTNAMELEN` is defined on Linux (in `sys/param.h`) as `HOST_NAME_MAX` (without +1?), while macOS man page statement regarding `MAXHOSTNAMELEN` meaning is similar to what you quoted from OpenBSD (+1 is needed).

As for me, there are too many differences and nuances across OSes in those constants to play with them. We tried it with `LOG_MAKEPRI` (#8269) and lost.

`sysconf()` looks as a good way to obtain the value from the system, but we don't want complicate this code with dynamic allocations.

I bet 256 bytes buffer is good enough for the at least next 10 years.

src/box/lua/info.c
```diff
@@ -650,6 +663,22 @@ lbox_schema_version(struct lua_State *L)
 	return 1;
 }
 
+/* gethostname() Lua interface inside box.info. */
```
Totktonada, line 666: Nit: function/struct/field is usually described using a documentation-style comment: `/** <...> */`.

src/box/lua/info.c
```diff
@@ -650,6 +663,22 @@ lbox_schema_version(struct lua_State *L)
...
+static int
+lbox_info_hostname(struct lua_State *L)
+{
+	char buffer[TT_HOST_NAME_MAX + 1];
+	int status = gethostname(buffer, TT_HOST_NAME_MAX + 1);
```
Totktonada, line 671: Nit: Usually named `rc` across our codebase.

Totktonada (approved):

src/box/lua/info.c
Lord-KA: I would prefer to remove this comment entirely, as the nearby functions don't have one and the code is self-explanatory, but checkpatch is opposed to that. Otherwise, added extra `*`.

src/box/lua/info.c
ligurio: I suspect you address these question to me, not to Gleb.
I believe hardcoded constants are bad. In discussion with Gleb I suggest to use something like this:

```
#ifndef HOST_NAME_MAX
#define HOST_NAME_MAX 256
#endif
```

> sysconf() looks as a good way to obtain the value from the system, but we don't want complicate this code with dynamic allocations.

Probably it is good enough, but `sysconf` is not called in the patch.

src/box/lua/info.c
olegrok, line 671:
```suggestion
	int status = gethostname(buffer, sizeof(buffer));
```

nit: probably it looks a bit better.

src/box/lua/info.c
Totktonada: // Sorry for the long read. I started from a short note, but then found that not everything is cleanly explained. A bit vague topic.

TL;DR: I don't mind your proposal, but don't see benefits.

----

> I suspect you address these question to me, not to Gleb.

The questions are to the commitee behind POSIX and to libc implementations. They're here to illustrate difficulties with those constants.

Gleb asked me to summarize my opinion on using the hardcoded constant, so I did the attempt to illustrate my doubts about limits.h and sys/param.h constants.

> I believe hardcoded constants are bad.

As for me, the always-the-same constant is less error-prone comparing to the depend-on-os-and-build-flags constant.

At worst, when the world will be changed by future implemetations of future standards, ancient tarantool versions will not able to show insanely long host names. I don't feel it as a problem.

The behavior for the hardcoded constant can be described this way: till version X.Y.Z only hostnames up to 256 bytes long were shown.

At the opposite: say, we change a build host for packages and it appears somehow that `__USE_POSIX` becomes enabled and it may change tarantool's behavior, because the constant has nothing in common with the run host, which may have more recent `gethostname()` implementation with its own limits.

The behavior for the non-obviously-constructed constant... is hard to describe. It may be a problem for one who tries to reproduce a problem around.

Not a big problem, though. I wouldn't mind the proposal, just don't see benefits.

----

Regarding `sysconf()`.

We discussed this option with Gleb and agreed that it would add an unnecessary complication to the code. VLA is commonly considered unsafe (stack overflow -> denial of the service). Maybe sysconf's value is perfectly limited, but maybe there are some attack vectors. I'm not sure, so I would avoid it.

So, `sysconf()` means a dynamic allocation of a temporary buffer and, I guess, fiber region fits good here. A bit more code -- allocation and releasing the memory. Not so much, but...

I would like to avoid bikeshedding discussions around the best allocator: shouldn't we use cord_ibuf? Or, maybe, just use xmalloc, but move the allocation to the initialization code?

I mean, the value is limited in the real world, at least by [RFC 1025](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1) (section 2.3.1):

> There are also some restrictions on the length. Labels must be 63 characters or less.

macOS developers think that it should be 255. Okay, let's just pick up this limit.

It is anyway a constant, so the dynamic allocation here feels unnecessary.

---

BTW, there is no yield between the `gethostname()` and `lua_pushstring()` and all calls are from the TX threads. We can use `static` for the buffer.

I'll add it as a separate comment.

src/box/lua/info.c
```diff
@@ -650,6 +663,22 @@ lbox_schema_version(struct lua_State *L)
 	return 1;
 }
 
+/* gethostname() Lua interface inside box.info. */
+static int
+lbox_info_hostname(struct lua_State *L)
+{
+	char buffer[TT_HOST_NAME_MAX + 1];
```
Totktonada, line 670: There is no yield between the `gethostname()` and `lua_pushstring()` and all calls are from the TX threads. We can use `static` for the buffer.

src/box/lua/info.c
Totktonada: > added extra `*`.

Forgot to push?

src/box/lua/info.c
olegrok: There was similar issue when somebody tried to use static buffers in GC callback.

E.g. https://github.com/tarantool/tarantool/commit/c302360f0e27b6c1c30eb5787557bc5949db1336#diff-01cf72a8b25f913884f0b9d67ec342db4828a0d27490241f36e969db766f2f48R96

test/box/info.test.lua
ligurio: Sorry, I mean `io.popen` or `popen.shell`. Execute `hostname` via popen, obtain value via stdout and compare it with value obtained using `box.info.hostname`.

src/box/lua/info.c
Totktonada: Ouch, yes. If a GC callback invokes `box.info.hostname`, we rewrite the buffer before `lua_pushstring()` catches it.

It would be a strange GC callback, but possible. Okay, let's allocate it on the stack on each call.

src/box/lua/info.c
ligurio: Discussed with  @Totktonada verbally and decided to left it as is.

Totktonada: > BTW, there is no yield between the `gethostname()` and `lua_pushstring()` and all calls are from the TX threads. We can use static for the buffer.

Discussed and, as Oleg correctly said, it is wrong due to possible GC callback invocation.

test/box/info.test.lua
Lord-KA: Sure, will do.

src/box/lua/info.c
Lord-KA: Renamed to `rc`.

Lord-KA: Good idea, thanks.

test/box/info.test.lua
Lord-KA: Added [another test](https://github.com/tarantool/tarantool/blob/ce73fa2f5697f2017b45730434b10bbf34d11aa3/test/box-luatest/gh_8605_add_box_info_hostname_test.lua) for that.

ligurio: Thanks!

ligurio (approved):
LGTM

ligurio (approved):
LGTM

test/box-luatest/gh_8605_add_box_info_hostname_test.lua
```diff
@@ -0,0 +1,18 @@
+local popen = require('popen')
+local t = require('luatest')
+
+local g = t.group()
+
+g.test_hostname_result = function()
+    local ph = popen.shell("hostname", 'r')
```
Totktonada, line 7: Nit: `ph:close()` is desirable. It is collected by GC anyway, but the GC delay at freeing a tightly limited resource (such as a file descriptor) may cause unexpected effects under certain workloads. There may be a peak, where it hits the file descriptor limit.

Unlikely we'll be hit by this problem there, just a suggestion regarding the good practice.

changelogs/unreleased/gh-8605-add-gethostname.md
Totktonada, line 1: I'm ready to push, but there is a conflict. Please, rebase on a fresh master and set the full-ci label.

test/box-luatest/gh_8605_add_box_info_hostname_test.lua
Lord-KA: My bad, missed `close()`.

Lord-KA: Fixed.

changelogs/unreleased/gh-8605-add-gethostname.md
Lord-KA: Done.

test/box/info.test.lua
Lord-KA: Well, there is a problem with hostname on [redos](https://github.com/tarantool/tarantool/actions/runs/5030912165/jobs/9023648604?pr=8651):
```
[036] sh: hostname: command not found
[036] #   expected: {exit_code = 0, state = "exited"}
[036] #   actual: {exit_code = 127, state = "exited"}
```

Totktonada: I suspect there is no hostname command there. Let's skip the test if it is so (either by the exit code or use 'type/command/which hostname' beforehand).

changelogs/unreleased/gh-8605-add-gethostname.md
Totktonada: Waiting for the test fix.

test/box/info.test.lua
Totktonada: The exit code 127 is [standartized](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02).

Lord-KA: Added `skip_if()` to the test.

changelogs/unreleased/gh-8605-add-gethostname.md
Lord-KA: @Totktonada, the test is fixed, PR ready for merge.

test/box/info.test.lua
Totktonada: No `:close()` at skipping the test, but this is negligible due to popen object's GC hook.

PR #8658, opened 2023-05-15
Allow multiple updates to the same field

### LINKED ISSUE #8226: Enable several update operations of the same field
In 8fc00a9c4bdd95bfa920c4757e711a3368eeb195 several operations of the same field (in the same update) were prohibited:
```lua
tarantool> s:update(1, {{'&', 2, 0x0F}, {'|', 2, 0xC0}})
---
- error: 'Field 2 UPDATE error: double update of the same field'
...
```
It seem to be not the most common case but sometimes it's natural to make several operations, like in the example above: drop a group of bits while setting some others.

Of course it must be done not ruining the original optimization that disabled this feature.

nshy:
Note that there is a distinct PR https://github.com/tarantool/tarantool/pull/8619 for "salad: make rope_erase erase more then a single element" patch. It will reviewed there.

Closes #8226

Lint failure is FP for typedef:
```
ERROR: Function name and return value type should be placed on different lines
#210: FILE: src/lib/salad/rope.h:661:
+typedef void (*rope_visit_f)(const rope_data_t data, size_t size, void *cb_arg);

total: 1 errors, 293 lines checked
```


src/lib/salad/rope.h
```diff
@@ -440,74 +456,92 @@ rope_extract(struct rope *rope, rope_size_t offset)
...
+ * - deleted elements reside inside the node. In that case split
  *   the existing node at offset, and insert the tail.
  *
  * The implementation is a copycat of rope_insert(). If you're
  * trying to understand the code, it's recommended to start from
  * rope_insert().
+ *
+ * Return:
+ *  0 - success
+ * -1 - OOM
```
Gerold103, line 480: Comments like these are not helpful and tend to outdate very quickly. Lets just drop them.

src/lib/salad/rope.h
```diff
@@ -440,74 +456,92 @@ rope_extract(struct rope *rope, rope_size_t offset)
...
- * When deleting a single element, 3 cases are possible:
- * - offset falls at a node with a single element. In this case we
- *   perform a normal AVL tree delete.
- * - offset falls at the end or the beginning of an existing node
- *   with leaf_size > 1. In that case we trim the existing node
- *   and return.
- * - offset falls inside an existing node. In that case we split
+ * When deleting, 3 cases are possible:
+ * - offset is at the beginning of the node and number of deleted elements
+ *   is more then or equal to the number elements in the node. In this
```
Gerold103, line 467: You seem to misuse `then` where `than` is needed. Please, check all the usages and fix, in the commit message too.

src/box/xrow_update.c
```diff
@@ -169,13 +169,8 @@ xrow_update_read_ops(struct xrow_update *update, const char *expr,
```
Gerold103, line 169: In the second commit you updated test-run submodule. I suppose it was an accident?

src/box/xrow_update_field.c
```diff
@@ -503,10 +504,17 @@ xrow_update_op_do_splice(struct xrow_update_op *op, const char *old)
...
+	if (str_len > 0)
+		xrow_update_string_rope_insert(str, 0, old, str_len);
+	if (arg->cut_length > 0)
+		xrow_update_string_rope_erase(str, arg->offset,
+					      arg->cut_length);
+	if (arg->paste_length > 0)
+		xrow_update_string_rope_insert(str, arg->offset, arg->paste,
+					       arg->paste_length);
+	op->new_field_len =
+		mp_sizeof_str(xrow_update_string_rope_size(str));
```
Gerold103: Doing that even when the operation is just one looks like a big overkill. Same for traversing the rope with a callback ("virtual" function call on the number of nodes). I suppose it would show some noticeable perf penalty for splice operations. But I will leave that to the core team to decide if this is fine.

src/box/xrow_update_field.c
```diff
@@ -584,29 +592,38 @@ xrow_update_op_store_bit(struct xrow_update_op *op,
...
 
+/**
+ * Copy string rope leaf data to the output buffer.
+ *
+ * Arguments:
+ *  data - leaf data pointer
+ *  size - leaf data size
+ *  cb_arg - pointer to the output buffer pointer
+ *
+ *  Buffer pointer updated to point after the copied data.
```
Gerold103, line 640: If you want to describe the args, then use the doxygen format, please. Same in all the other new code.

src/box/xrow_update_field.h
```diff
@@ -128,10 +170,8 @@ struct xrow_update_arg_splice {
 	/** New content length. */
 	uint32_t paste_length;
```
Gerold103, line 188: This empty line now seems excess. The is just one member after it, not a group of 2.

Gerold103 (commented):
Thanks for the patch! I will return with more comments later.

src/box/xrow_update_array.c
```diff
@@ -448,10 +448,11 @@ xrow_update_op_do_array_##op_type(struct xrow_update_op *op,			\
 	}									\
 	if (item->field.type != XUPDATE_NOP)					\
 		return xrow_update_err_double(op);				\
-	if (xrow_update_op_do_##op_type(op, item->field.data) != 0)		\
+	const char *data = item->field.data;					\
+	xrow_update_mp_read_scalar(&data, &item->field.scalar);			\
```
Gerold103: Is that correct that this commit (`update: keep result of scalar operation in tree itself`) makes the code parse the scalar operation argument multiple times (AFAIK, it was parsed only one time before).

src/box/xrow_update_field.c
```diff
@@ -296,102 +324,135 @@ xrow_update_read_arg_splice(struct xrow_update_op *op, const char **expr,
 
 /* {{{ do_op helpers. */
 
+/*
```
Gerold103, line 327: Lets use `/**` opener for out-of-function comments. Same in the other places.

src/box/xrow_update_field.c
```diff
@@ -296,102 +324,135 @@ xrow_update_read_arg_splice(struct xrow_update_op *op, const char **expr,
...
-			assert(int96_is_neg_int64(&arg.int96));
-			return int96_extract_neg_int64(&arg.int96);
+			assert(int96_is_neg_int64(&scalar->int96));
+			return int96_extract_neg_int64(&scalar->int96);
 		}
+	case XUPDATE_TYPE_DECIMAL:
+	case XUPDATE_TYPE_STR:
+	case XUPDATE_TYPE_NONE:
+	default:
+		assert(false);
```
Gerold103, line 349: I suggest to use `unreachable()`. Same in the other places.

src/box/xrow_update_field.h
```diff
@@ -99,6 +99,38 @@ struct xrow_update_field;
...
+/**
+ * Scalar type.
+ */
+enum xrow_update_scalar_type {
+	XUPDATE_TYPE_NONE = 0, /** UNINITIALIZED */
+	XUPDATE_TYPE_DECIMAL = 1, /** decimal_t */
+	XUPDATE_TYPE_DOUBLE = 2, /** double */
+	XUPDATE_TYPE_FLOAT = 3, /** float */
+	XUPDATE_TYPE_INT = 4, /** struct int96_num */
+	XUPDATE_TYPE_STR = 5, /** xrow_update_string_rope */
```
Gerold103, line 111: We don't put comments on the same line as code (you might see violations in the old code, but it is legacy), lets move them above each of the constants. Or just drop them, they are obvious anyway.

src/box/xrow_update_field.h
```diff
@@ -435,11 +437,10 @@ struct xrow_update_field {
 		 * intersected with any another update operation.
 		 */
 		struct {
-			/**
-			 * Bar update is a single operation
-			 * always, no children, by definition.
-			 */
+			/** Bar operation. */
 			struct xrow_update_op *op;
+			/** Bar result calar in case of scalar operation. */
```
Gerold103, line 458: `calar` -> `scalar`.

Gerold103 (commented):
Next comments. I will continue the review a bit later.

test/box/update.result
```diff
@@ -1687,27 +1687,19 @@ ops = {{'=', '[4][4][1]', 1}, {'=', '[4][4][1]', 2}}
...
-- error: 'Field ''[4][4][5][4]'' UPDATE error: double update of the same field'
-...
-t:upsert(ops)
----
-- [1, {}, [], [1, 2, 3, [4, 5, 6, 7, [8000, 9, [10, 11, 12], 13013, 14], 15, 16, [
-        17, 18, 19]], 20, {'b': 22, 'a': 21, 'c': {'d': 23, 'e': 24}}, 25]]
-...
+-- ops = {{'+', '[4][4][5][4]', 13000}, {'=', '[4][4][5][1]', 8000}, {'+', '[4][4][5][4]', 1}}
+-- t:update(ops)
+-- t:upsert(ops)
```
Gerold103: Lets uncomment this and update the results.

test/box/update.result
```diff
@@ -1821,14 +1813,8 @@ t:update({{'=', '[4]', {1}}, {'+', '[4][1]', 1}})
...
-t:update({{'=', '[4][6]', {a = 33}}, {'+', '[4][6].aa', 1}})
----
-- error: 'Field ''[4][6].aa'' UPDATE error: double update of the same field'
-...
-t:update({{'=', '[4][4]', {3}}, {'+', '[4][4][1]', 1}})
----
-- error: 'Field ''[4][4][1]'' UPDATE error: double update of the same field'
-...
+-- t:update({{'=', '[4][6]', {a = 33}}, {'+', '[4][6].aa', 1}})
+-- t:update({{'=', '[4][4]', {3}}, {'+', '[4][4][1]', 1}})
```
Gerold103: Ditto.

test/engine/update.result
```diff
@@ -1219,6 +1222,159 @@ s = box.schema.create_space('test', {engine = engine})
...
+-- Double update of a field inside a map.
+s:replace{1, {a = 1, b = 3}}
+---
+- [1, {'a': 1, 'b': 3}]
+...
+s:update(1, {{'+', '[2].b', 2}, {'+', '[2].a', 2}, {'+', '[2].a', 3}})
+---
+- [1, {'b': 5, 'a': 6}]
+...
+-- Fail to apply scalar op to non scalar/nop field in array
```
Gerold103, line 1361: `array` -> `map`.

src/box/xrow_update_field.h
```diff
@@ -263,6 +263,28 @@ struct xrow_update_op {
 	bool is_for_root;
 };
 
+/**
+ * Return weather operation is scalar (all other then insert/delete/set).
```
Gerold103, line 267: `weather` -> `wether`.

src/box/xrow_update_field.h
```diff
@@ -263,6 +263,28 @@ struct xrow_update_op {
...
+ */
+static inline bool
+xrow_update_op_is_scalar(struct xrow_update_op *op)
```
Gerold103, line 270: Lets make the argument point at `const`.

src/box/xrow_update_field.h
```diff
@@ -263,6 +263,28 @@ struct xrow_update_op {
...
+	case '+':
+	case '-':
+	case '&':
+	case '|':
+	case '^':
+	case ':':
+		return true;
+	case '=':
+	case '#':
+	case '!':
```
Gerold103, line 282: These are not needed - they can be included into the `default`.

src/box/xrow_update_route.c
```diff
@@ -260,15 +260,16 @@ xrow_update_route_branch(struct xrow_update_field *field,
...
 		/*
-		 * Yeah, bar length can become 0 here, but it is
-		 * ok as long as it is a scalar operation (not '!'
-		 * and not '#'). When a bar is scalar, it operates
-		 * on one concrete field and works even if its
-		 * path len is 0. Talking of '#' and '!' - they
-		 * are handled by array and map 'branchers'
-		 * internally, below. They reapply such
-		 * operations.
+		 * Non scalar leaf bar (bar length is 0) will be
```
Gerold103, line 270: Typo? Non scalar leaf bar is when length is > 0, not 0.

test/box/update.result
```diff
@@ -1695,11 +1695,20 @@ t:upsert(ops)
 - [1, {}, [], [1, 2, 3, [2, 5, 6, 7, [8, 9, [10, 11, 12], 13, 14], 15, 16, [17, 18,
         19]], 20, {'b': 22, 'a': 21, 'c': {'d': 23, 'e': 24}}, 25]]
 ...
--- First two operations produce zero length bar update in field
--- [4][4][5][4]. The third operation should raise an error.
--- ops = {{'+', '[4][4][5][4]', 13000}, {'=', '[4][4][5][1]', 8000}, {'+', '[4][4][5][4]', 1}}
--- t:update(ops)
--- t:upsert(ops)
+-- First two operations produce zero length bar update in field [4][4][5][4].
```
Gerold103, line 1698: Ok, I see now why you commented out the tests in the prev patch. It is not a problem to change the same test multiple times during the patchset if it is justified. Lets not comment anything out.

changelogs/unreleased/gh-8226-multiple-same-field-update.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
+
+* Allow multiple update for the same field (gh-8226).
```
Gerold103, line 3: Lets at least mention that you mean tuple fields, not just some other fields. Also you need to write a docbot request and go through the commit messages to fix typos, add ticket references.

Gerold103 (changes requested):
Good stuff, I don't have any big comments.

src/lib/salad/rope.h
nshy: You mean the implementation part? I would keep the first paragraph and the return block - it is good to have comment what function do in general.

Implementation part is also useful IMHO. We can move it inside function.

src/box/xrow_update_array.c
nshy: Sorry, I don't see how it is more then a single parse. Please, explain)

src/box/xrow_update_array.c
```diff
@@ -446,11 +446,15 @@ xrow_update_op_do_array_##op_type(struct xrow_update_op *op,			\
...
-	const char *data = item->field.data;					\
-	xrow_update_mp_read_scalar(&data, &item->field.scalar);			\
-	if (xrow_update_op_do_##op_type(op, &item->field.scalar) != 0)		\
+	struct xrow_update_scalar *scalar = &item->field.scalar;		\
+	struct xrow_update_scalar na = { XUPDATE_TYPE_NONE };			\
+	if (item->field.type == XUPDATE_NOP) {					\
+		const char *data = item->field.data;				\
+		xrow_update_mp_read_scalar(&data, &item->field.scalar);		\
+	} else if (item->field.type != XUPDATE_SCALAR) {			\
+		scalar = &na;							\
```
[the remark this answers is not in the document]
nshy: So then we call `xrow_update_op_do_##op_type`. Every such function checks it's type and print error. We only follow this branch `item->field.type != XUPDATE_SCALAR && item->field.type != XUPDATE_NOP` when we try to apply scalar operation to non-scalar field so error message will be appropriate.

src/box/xrow_update_field.h
nshy: I'd like instead to put `unreachable()` to default so we have all operation explicitly in the switch.

src/box/xrow_update_route.c
nshy: Looks correct.

Here `child.bar.path_len` is referenced as length. This length can be 0 for scalar and non-scalar leafs. The branch above handles scalar so this comment is for non-scalar leaf. (I rephrased the previous comment as scalar leaf is now handled explicitly.) 

May be I'd better put () right after leaf word.

test/box/update.result
nshy: AFAIR at least some of the tests leads to assertion. I'll then comment only such tests.

src/lib/salad/rope.h
nshy: Fixed.

src/box/xrow_update.c
nshy: Fixed.

src/box/xrow_update_field.h
nshy: Done.

test/box/update.test.lua
```diff
@@ -629,7 +629,7 @@ t:upsert(ops)
 -- gh-8216: attempt to make an update inside a just created array field could
 -- crash.
 --
-t:update({{'=', '[3]', {a = 33}}, {'+', '[3].a', 1}})
+t:update({{'=', '[2]', {a = 33}}, {'+', '[2].a', 1}})
```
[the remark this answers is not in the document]
nshy: The thing is if I keep this test unchanged then after the patch I'll get:
```
---
- error: 'Tuple field 3 (g) type does not match one required by operation: expected
    array, got map'
...
```
It is not from update itself. Update will produce `{a=34}` in the second field of the tuple. But tuple format states second field should be an array. So test become a bit strange.

src/box/xrow_update_field.c
nshy: Discussed with @alyapunov Decided we should try to avoid performance degradation. We probably can do the simplest thing - use old structure until there is a second update - in that case convert it to rope.

test/box/update.test.lua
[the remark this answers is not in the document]
Gerold103: It might be that this test in its new form won't cover the crash fix done in the same commit where the test was introduced. That is why I would propose to keep it as is even if it fails weirdly. It just shouldn't crash. It is not expected to succeed AFAIR.

src/box/xrow_update_array.c
Gerold103: You added here `xrow_update_mp_read_scalar()` which parses `item->field.data`. This code can be done by multiple update operations having the same path, right? Which means that every operation reaching this place will read the scalar field again. Wouldn't it?

src/box/xrow_update_array.c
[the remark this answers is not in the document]
Gerold103: Ok, I see now, thanks. It is worth adding a short comment about that here.

changelogs/unreleased/gh-8226-multiple-same-field-update.md
p7nov, line 3:
```suggestion
* Allowed multiple updates to the same tuple field in a single call (gh-8226).
```

p7nov (approved):

src/lib/salad/rope.h
nshy: I guess I'll keep then. The thing is the API do not check operation validity (assertions at the beginning of `rope_erase`) so the only possible error is OOM. Looks like it good to know it. In particular it is used in the next patch in assertions for `xrow_update_rope_*` return value.

Regularly I don't specify error reasons it the comment too.

src/box/xrow_update_array.c
nshy: It is not quite so. After this patch the second update will still hit `item->field.type != XUPDATE_NOP` raising `xrow_update_err_double` error. And after the next patch which touches the same place the parse will be done only on the first time when `item->field.type == XUPDATE_NOP`. After that type is changed to `XUPDATE_SCALAR` and we only do the update operation itself without parsing.

src/box/xrow_update_field.c
nshy: After optimization degradation is gone:
A = 1.57
D = 0.029

test/box/update.test.lua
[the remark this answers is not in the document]
nshy: Fixed.

src/box/xrow_update_field.c
nshy: Fixed.

src/box/xrow_update_field.h
nshy: Fixed to separate line for the comment.

src/box/xrow_update_field.h
nshy: Fixed.

src/box/xrow_update_field.h
```diff
@@ -790,32 +791,100 @@ OP_DECL_GENERIC(splice)
...
+ *  this_node - link to the field format (NULL allowed)
+ *  out - output buffer position
+ *  ont_end - output buffer end
+ *
+ * xrow_update_scalar_type <-> MsgPack correspondence:
+ *  XUPDATE_TYPE_INT       - MP_UINT/MP_INT
+ *  XUPDATE_TYPE_DOUBLE    - MP_DOUBLE
+ *  XUPDATE_TYPE_FLOAT     - MP_FLOAT
+ *  XUPDATE_TYPE_STR       - MP_STR
+ *  XUPDATE_TYPE_DECIMAL   - MP_EXT+MP_DECIMAL
```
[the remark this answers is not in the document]
nshy: Dropped.

src/box/xrow_update_array.c
```diff
@@ -446,11 +446,15 @@ xrow_update_op_do_array_##op_type(struct xrow_update_op *op,			\
 		op->is_token_consumed = true;					\
 		return xrow_update_op_do_field_##op_type(op, &item->field);	\
 	}									\
-	if (item->field.type != XUPDATE_NOP)					\
-		return xrow_update_err_double(op);				\
-	const char *data = item->field.data;					\
-	xrow_update_mp_read_scalar(&data, &item->field.scalar);			\
-	if (xrow_update_op_do_##op_type(op, &item->field.scalar) != 0)		\
+	struct xrow_update_scalar *scalar = &item->field.scalar;		\
+	struct xrow_update_scalar na = { XUPDATE_TYPE_NONE };			\
```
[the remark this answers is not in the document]
nshy: Done.

src/box/xrow_update_array.c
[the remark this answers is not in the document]
nshy: Done.
`/* Just stub for non scalar field. op_do_ * will fail on it. */`

src/box/xrow_update_field.h
nshy: Fixed.

src/box/xrow_update_field.h
nshy: Done.

src/box/xrow_update_route.c
nshy: Changed to `(path length is 0)`.

changelogs/unreleased/gh-8226-multiple-same-field-update.md
nshy: Done.

test/engine/update.result
nshy: Fixed.

test/box/update.result
nshy: Adjusted patch set a bit to avoid temporary assertions and as a result avoid commenting tests out.

test/box/update.result
nshy: Done. Add a comment for temporary result:
`Currenly result is not correct. It is temporary and will be fixed later.`

test/box/update.result
nshy: Moved the below hunk to a later patch so now I don't need commenting these tests.
```diff
-               if (old_token.type == JSON_TOKEN_END) {
-                       /*
-                        * The bar path ended. It means the new operation either
-                        * is trying to update exactly the same field, or go
-                        * even deeper.
-                        */
-                       xrow_update_err_double(new_op);
-                       return NULL;
-               }
```
In the previous version of the patch set I deleted hunk too early -  it works until the patch where it deleted now.

test/engine/update.result
```diff
@@ -1210,3 +1210,195 @@ t:update({{'=', 'field3', 3}})
 s:drop()
 ---
 ...
+--
+-- gh-8226: allow several updates of the same field
+--
+s = box.schema.create_space('test', {engine = engine})
```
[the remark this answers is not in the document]
nshy: Moved tests to `box/update.test.lua`.

Not sure I need to add tests testing first field or field in the middle. New related code seems to be orthogonal to position in the array.

May be we can change tests to update field in the middle to check that last field in not corrupted/overwritten somehow? Not sure I need to test first/middle/last element for every case

src/lib/salad/rope.h
Gerold103: You still need to convert it to doxygen format.

Gerold103 (commented):
In the first commit message you still have the `*TODO* Add ticket reference.` stub.

test/engine/update.result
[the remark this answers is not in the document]
Gerold103: > Not sure I need to add tests testing first field or field in the middle. New related code seems to be orthogonal to position in the array.

I wouldn't say so. There is at least one commit where you touch the tail update data in a struct, which in your tests is almost always empty, it seems.

> May be we can change tests to update field in the middle to check that last field in not corrupted/overwritten somehow?

You could reuse the big complex tuples in `box/update.test.lua` (lines 297, 514) created specifically for that. The tests would remain the same, but would operate on a more complex tuple.

nshy:
I still plan to merge the first commit in a different PR. It's review is not even started though).

src/box/xrow_update_field.c
```diff
@@ -296,102 +334,133 @@ xrow_update_read_arg_splice(struct xrow_update_op *op, const char **expr,
...
+	case XUPDATE_TYPE_DECIMAL:
+	case XUPDATE_TYPE_STR:
+	case XUPDATE_TYPE_NONE:
+	default:
+		unreachable();
 	}
+	return 0;
 }
 
+/*
```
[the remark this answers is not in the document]
nshy: Fixed.

test/engine/update.result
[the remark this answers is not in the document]
nshy: > I wouldn't say so. There is at least one commit where you touch the tail update data in a struct, which in your tests is almost always empty, it seems.

If you are referencing commit `update: make '=' NOP in update tree instead of scalar` - this should be covered by existing tests.

> You could reuse the big complex tuples in box/update.test.lua (lines 297, 514) created specifically for that. The tests would remain the same, but would operate on a more complex tuple.

I would say large tuple will make it harder to follow testing case. In the above places it justified as JSON path is tested.

If I use a complex tuple should I in every test case pick up a different place? This seems a bit a different test flavor, more like a random test. If it is a desired then I'm OK to change but it looks a bit of work)

I guess it is not justified currently that there is a field before updated and there is no field after it. It is a remnant of the first version where I used `space:update()`. May be I should just add an after field?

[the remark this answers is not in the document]
Gerold103: Adding an after field would be good. And also keep the existing tests updating the last field as well. For such complicated code there can't be too many tests.

> every test case pick up a different place?

If you want. I would just recommend testing non-last and last field updates.

> but it looks a bit of work)

This isn't much more complicated than your current tests really. You just have to check results one time in the .result file.

Anyway, I will leave that to the next reviewer.

Gerold103 (approved):
Thanks for the fixes!

nshy:
Thanx for the review!

test/engine/update.result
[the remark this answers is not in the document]
nshy: Added last field for tuples in test. Typical tuple now looks like this `t = box.tuple.new({7, {a = 1}, 11}) `.

PR #8662, opened 2023-05-16
sql: assert in xferOptimization()

### LINKED ISSUE #8661: Assertion when executing INSERT INTO ... SELECT FROM
Reproducer:
```
Tarantool 3.0.0-entrypoint-228-gb3fb883b5
tarantool> box.cfg{log_level=0}
---
...

tarantool> box.execute([[CREATE TABLE t(i INT PRIMARY KEY);]])
---
- row_count: 1
...

tarantool> _ = box.schema.space.create('A', {format = {{'i', 'integer'}}})
---
...

tarantool> box.execute([[INSERT INTO t SELECT * FROM a;]])
tarantool: ./src/box/sql/build.c:840: vdbe_emit_open_cursor: Assertion `idx != NULL' failed.
```

Reported by @nshy.

ImeevMA:
This patch fixes problems with INSERT INTO ... SELECT FROM optimization. These problems appeared after 6b8acd8, where the check became redundant, but was not updated. Two problems arose:
1) an assertion or segmentation fault when optimization was used and the source space does not have an index;
2) optimization can be used even if the indexes are incompatible.

The second problem does not result in changes that are noticeable to the user, so there is no test.

Closes #8661

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-8661-assertion-in-xferOptimization.md
+++ b/changelogs/unreleased/gh-8661-assertion-in-xferOptimization.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* Fixed a possible assertion or segmentation fault when optimizing
+  `INSERT INTO ... SELECT FROM` (gh-8661).
--- a/src/box/sql/insert.c
+++ b/src/box/sql/insert.c
@@ -1081,6 +1081,9 @@ xferOptimization(Parse * pParse,	/* Parser context */
 	/* Src may not be a view. */
 	if (src->def->opts.is_view)
 		return 0;
+	/* Src should have at least one index. */
+	if (src->index_count < 1)
+		return 0;
 	/* Number of columns must be the same in src and dst. */
 	if (dest->def->field_count != src->def->field_count)
 		return 0;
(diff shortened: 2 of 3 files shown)


Gerold103 (approved):
Thanks for the patch!

changelogs/unreleased/gh-8661-assertion-in-xferOptimization.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/sql
+
+* Fixed a possible assertion or segmentation fault when optimizing
+  INSERT INTO ... SELECT FROM (gh-8661).
```
p7nov, line 4:
```suggestion
  `INSERT INTO ... SELECT FROM` (gh-8661).
```

p7nov (approved):

changelogs/unreleased/gh-8661-assertion-in-xferOptimization.md
ImeevMA: Thanks, fixed.

igormunkin:
Also backported to 2.11 and 2.10 in scope of ab61f77 and 4ea1fb2 respectively.

PR #8671, opened 2023-05-18
ci: fix lango stale PRs workflow

fckxorg:
This patch fixes the crontab expression in the
Lango team stale PRs workflow, so it runs not every minute, but every hour.

Also, python version is updated to 3.11, since the script is reliant onto `datetime.fromisoformat' changes, that were introduced in that version.

NO_DOC=Workflow fix
NO_TEST=Workflow fix
NO_CHANGELOG=Workflow fix


igormunkin (approved):
@fckxorg, thanks for the patch! LGTM as trivial.

Issue #8701, opened 2023-05-24
Regenerate `test/box-luatest/gh_6794_data`


locker:
The [directory](https://github.com/tarantool/tarantool/tree/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/gh_6794_data) contains snap and xlog files that are used to start a test instance. The files use an old schema (the schema was upgraded after they were generated). In the scope of #7149 we disable any DDL with an old schema version. The problem is luatest needs to grant permissions (which is a DDL operation) to run tests (it does it in the test server instance source file [here](https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/server_instance.lua#L47)). As a result, the [test](https://github.com/tarantool/tarantool/blob/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/gh_6794_recover_nonmatching_xlogs_test.lua) fails while trying to start a test server instance.

To fix this issue, we need to regenerate the data files with `box.schema.user.grant('guest', 'super')`. Unfortunately, there's no script that could be used for this (like [this one](https://github.com/tarantool/tarantool/blob/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/upgrade/2.10.4/gen.lua), for example) so they have to be regenerated manually.

It'd be nice if in the scope of this ticket we'd also add a script that would regenerate the data files automatically.

locker:
Reassigning to the test author @sergepetrenko.

Please regenerate the contents of [`test/box-luatest/gh_6794_data`](https://github.com/tarantool/tarantool/tree/97c2c9a4fb58b058e7dedd420fce11a6f1ade21f/test/box-luatest/gh_6794_data) while granting super permissions to the guest user, like it was done here:

https://github.com/tarantool/tarantool/blob/97c2c9a4fb58b058e7dedd420fce11a6f1ade21f/test/box-luatest/upgrade/2.10.4/gen.lua#L3-L5

Then enable the test by removing the following line:

https://github.com/tarantool/tarantool/blob/97c2c9a4fb58b058e7dedd420fce11a6f1ade21f/test/box-luatest/gh_6794_recover_nonmatching_xlogs_test.lua#L18

Optional but very desirable: add a script that would generate the data files automatically.

locker:
[...]

PR #8705, opened 2023-05-25
box: disable DDL with old schema

### LINKED ISSUE #7149: Forbid DDL before `box.schema.upgrade` is called
We already track which schema version is needed for correct Tarantool operation.
The user sees the following message after upgrading the database:
```
2022-05-18 12:54:07.790 [9379] main/103/interactive load_cfg.lua:857 W> Your schema version is 1.10.2 while Tarantool 2.11.0-entrypoint-49-g9ce8798d1 requires a more recent schema version. Please, consider using box.schema.upgrade().
```

Having said that, Tarantool still allows the user to run ddl on a cluster with non-upgraded schema.
This leads to various unpredictable errrors (because the ddl code assumes the schema is already upgraded).

Example (trying to create a space on 2.11 Tarantool with 1.10 schema):
```
tarantool> box.schema.space.create('test')
---
- error: 'Tuple field 1 (id) type does not match one required by operation: expected
    unsigned, got nil'
...
```
Example (trying to drop a space on 2.8 Tarantool with 1.10 schema):
```
- error: 'builtin/box/schema.lua:515: attempt to index field ''space_id'' (a nil value)'
```

First of all, DDL on an old schema might return an unclear error. Like in the example above.
Secondly, DDL might complete without errors but produce erroneous results.
Lastly, when DDL fails on a non-upgraded schema, it might leave the schema in an incorrect state. DDL is not transactional currently (#600), and if the error is generated on some intermediate DDL step, the schema will be left in a partially changed state. And all further attempts to perform DDL will fail.

Let's forbid DDL altogether unless the user has the most recent schema version.

locker:
## Description

Executing DDL operations with an old (not upgraded) system schema is dangerous and might result in unexpected breakages. So we decided to explicitly forbid all DDL operations with an old system schema until `box.schema.upgrade()` is called.  Note, one can still call `box.schema` functions with an old schema provided they do nothing, for example, if an object is created with the `if_not_exists` flag and the object with same id already exists:

```lua
box.schema.create_space('test', {if_not_exists = true})
```

Otherwise an attempt to create a space with an old schema will raise an error like shown below:

```yaml
tarantool> box.schema.space.create('test')
---
- error: Your schema version is 1.6.8 while Tarantool
    3.0.0-entrypoint-262-g3eaba1cef686 requires a more recent
    schema version. Please, consider using box.schema.upgrade().
...
```

## Implementation details

We disable DDL by patching the existing on_replace_dd_system_space trigger callback installed for each system space so that now it raises an error in case the current schema version is less than the most recent one known to this build. Since to perform a schema upgrade we need to execute DDL, we suppress the error for the fiber that is currently running a schema upgrade. To achieve that, the upgrade script calls box_schema_upgrade_begin and box_schema_upgrade_end before starting and after completing a schema upgrade. The functions keep track of the fiber that is currently running a schema upgrade so that we can allow all DDL operations for it. We also allow DDL during recovery so that we can replay DDL statements written to the WAL.

Since there may be a bug in the `box.schema.upgrade` implementation, we export `box.internal.run_schema_upgrade`, which runs the given function as a schema upgrade script (allowing DDL). The user may use this function to recover after a schema upgrade failure.

## Note about the tests

A test server instance started by luatest grants permissions to the guest user so that luatest can execute commands on it. It means that if a test uses a generated snap file committed to the repository for a test server instance, it will fail because granting permissions is a DDL operation. To prevent this, we have to regenerate snap files so that they contain all required permissions. This works because a test server instance grants permissions with the `if_not_exists` flag.

The problem is that it isn't easy to regenerate the snap files for the following tests because there's no generator script:
 - `test/box-luatest/gh_6794_recover_nonmatching_xlogs_test.lua`
 - `test/box-luatest/gh_7974_force_recovery_bugs_test.lua`

So we temporarily disable these tests and file tickets to fix them, see #8701 #8702.

Other notes:
 - We drop `test/box-luatest/upgrade/2.9.1` and make the test using it use `test/box-luatest/upgrade/2.10.0` instead. We do this because 2.9.1 was never released and the earliest Tarantool version using the 2.9.1 schema version is 2.10.0. This shouldn't affect the test anyhow.
 - We drop the part of the `user_auth_history_last_modified_upgrade` test that checks that creating users/roles with an old schema works fine because this is forbidden now.
 - We wrap the code that creates a space with an old schema in the downgrade test in `box.internal.run_schema_upgrade`. Even though it's unsupported now, we still need to check that space creation works after a downgrade.

Closes #7149


alyapunov (approved):
LGTM,
Nit: `upgrde` -> `upgrade` in the commit message.

changelogs/unreleased/gh-7149-forbid-ddl-until-box-schema-upgrade.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+- Disabled DDL operations with an old system schema. Now you have to run
+  `box.schema.upgrade()` before you can execute any DDL operations (gh-7149).
```
Totktonada, line 4: It is a breaking change.

locker: You're right - my bad. Fix is here: https://github.com/tarantool/tarantool/pull/8708

PR #8709, opened 2023-05-26
test: ban direct calling of box.cfg()

ochaplashkin:
Direct call and configuration of the runner instance is prohibited. Now if you need to test something with specific configuration use a server instance please (see luatest.Server module).

In-scope-of tarantool/luatest#245

NO_DOC=ban calling box.cfg
NO_TEST=ban calling box.cfg
NO_CHANGELOG=ban calling box.cfg

### DIFF
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -1,0 +1,20 @@
+-- There are situations when the Tarantool instance is configured
+-- for testing. For example:
+--
+--     g.test_foo = function()
+--         box.cfg{params}
+--         ...
+--         t.assert_equals(foo, bar)
+--
+--     g.test_bar = function()
+--         t.assert_equals(foo, bar)
+--
+-- We expect that the test environment will not be configured
+-- in the `test_bar` test, but this will not always be the case
+-- (depends on the order of execution of tests).
+-- We prohibit a direct call to `box.cfg` by overriding the
+-- `__call` meta method. Motivating issue tarantool/luatest#245.
+box.cfg = setmetatable({}, {__call = function()
+    error('A direct call to `box.cfg` is forbidden' ..
+          ' (you are changing the test runner instance)', 0)
+end})
--- a/test/metrics-luatest/helper.lua
+++ b/test/metrics-luatest/helper.lua
@@ -109,4 +109,12 @@ local function workaround_requires(path)
 end
 
 workaround_requires('test.utils')
-workaround_requires('test.psutils_linux_test_payload')
+
+require('test.utils')
+local test_root = fio.dirname(
+    fio.dirname(fio.abspath(
+        package.search('third_party.metrics.test.utils')))) -- luacheck: ignore
+
+package.loaded['test.utils'].LUA_PATH = os.getenv('LUA_PATH') ..
+    test_root .. '/?.lua;' ..
+    test_root .. '/?/init.lua;'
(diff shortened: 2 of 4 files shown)


test/helper.lua
```diff
@@ -0,0 +1,19 @@
...
+--     g.test_foo = function()
+--         box.cfg{params}
+--         ...
+--         t.assert_equals(foo, bar)
+--
+--     g.test_bar = function()
+--         t.assert_equals(foo, bar)
+--
+-- We expect that the test environment will not be configured
+-- in the `test_bar`, but this will not always be the case
```
ylobankov, line 13: ... in the `test_bar` test ...

test/helper.lua
```diff
@@ -0,0 +1,19 @@
...
+-- (depends on the order of execution of tests).
+-- We prohibit a direct call to `box.cfg` by overriding the
+-- `__call` meta method.
```
ylobankov, line 16: Please mention here the corresponding ticket as well.

test/helper.lua
```diff
@@ -0,0 +1,19 @@
...
+box.cfg = setmetatable({}, {__call = function()
+    error('A direct call to `box.cfg` is not recommended' ..
```
ylobankov, line 18: We implicitly fail the test if `box.cfg` is called, so `is not recommended` sounds strange. It is rather `is forbidden`.

test/metrics-luatest/helper.lua
```diff
@@ -109,4 +109,12 @@ local function workaround_requires(path)
 end
 
 workaround_requires('test.utils')
-workaround_requires('test.psutils_linux_test_payload')
+
+require('test.utils')
```
ylobankov, line 113: What is this line for?

test/helper.lua
```diff
@@ -0,0 +1,19 @@
...
+        ' (you are changing the test runner instance)', 0) end})
```
ylobankov, line 19: Please indent this line on the same level as the line above.

test/metrics-luatest/helper.lua
ochaplashkin: @DifferentialOrange Could you tell us the details about it? Thanks!

test/helper.lua
ochaplashkin: Fixed.

test/helper.lua
ochaplashkin: Yep, added.

test/metrics-luatest/helper.lua
DifferentialOrange: We need to set LUA_PATH for test external server. Before the patch, LUA_PATH was needed for tarantool/metrics tests (to find metrics itself and luatest) and wasn't needed for tarantool/tarantool tests (since metrics is embedded and path for luatest is set with test-run). After this patch, some tests were migrated for using external server. Some of these migrated tests now need `test.utils` and `test.psutils_linux_test_payload` (`third_party/metrics/test/utils.lua` and `third_party/metrics/test/psutils_linux_test_payload`). Before this patch, we needed to provide  `test.utils` and `test.psutils_linux_test_payload` only to main luatest instance, which was workarounded by `workaround_requires` helper. Now we also need to provide these packages to external server. We do it here through setting LUA_PATH. You can set LUA_PATH for metrics tests through `test.utils` package `LUA_PATH` field. So this code provides `LUA_PATH` required for updates tests to be able to work in tarantool/tarantool.

Issue #8613, opened 2023-04-28
Rework Tarantool CLI options and default behaviour


ligurio:
* [ ] Now Tarantool binary enters interactive mode starting Lua REPL by default. In scope of this issue Tarantool should stop running Lua interpreter and yield the help message.
* [ ] Introduce `--name` (`-n`) option to specify instance name to be started. The option can be omitted in case `TT_INSTANCE_NAME` environment variable is set.
* [ ] Introduce `--config` (`-c`) to specify the path to the config file. If the option is not set the value of `TT_CONFIG` environment variable is considered. If none of the above is specified, the default path is used.
* [ ] The help message should be adjusted to the changes been made.

Issue #5040, opened 2020-06-02
tarantool -e always enters interactive mode when stdin is a tty


un-def:
**Tarantool version:** 2.4.1-52-g4cd784999

**OS version:** Ubuntu 20.04 LTS

**Bug description:**

The comment of https://github.com/tarantool/tarantool/commit/a305d08ed9a097ddfe39faa3b7b1c4b5c4cc25c3 states that tarantool is “fully compatible with lua5.1” it terms of CLI. Actually, it's not quite true.

There is a difference in how `-e` parameter works. In PUC-Rio Lua and vanilla LuaJIT, if `-i` parameter is not passed, then the interpreter quits after execution of the expression regardless of whether or not stdin isatty.

In tarantool, the interpreter **always** enters interactive mode if stdin isatty regardless of the presence of `-i` parameter. The comment https://github.com/tarantool/tarantool/blob/03790ac5510648d1d9648bb2281857a7992d0593/src/main.cc#L752 says that the `-i` parameter actually **forces** interactive mode, not enables it (at least I could guess it from the comment).

The help message is misleading, it describes how it works in PUC-Rio Lua and vanilla LuaJIT:

```
  -e EXPR			execute string 'EXPR'
  -i				enter interactive mode after executing 'SCRIPT'
```

Actually, it is nearly a copy–paste of the original help message of original Lua REPL.

As I understand, the only way to enforce tarantool REPL to not to enter interactive mode is to disconnect its stdin from tty. For example, it works: `tarantool -e EXPR < /dev/null` or `echo | tarantool -e EXPR`.

The currect behavior breaks luarocks (vanilla luarocks, not luarocks that is shipped with tarantool) on Debian-based systems. luarocks [tries to find lua executable in PATH](https://github.com/luarocks/luarocks/blob/v3.3.1/src/luarocks/cmd.lua#L175) (and tarantool is linked to `/usr/bin/lua` via alternatives mechanism), [executes found binary](https://github.com/luarocks/luarocks/blob/v3.3.1/src/luarocks/util.lua#L439) with `-e` parameter and hangs on forever because tarantool does not exit.

**Steps to reproduce:**

* Execute in a terminal emulator: `tarantool -e 'print(_VERSION)'`.

**Expected behavior:** tarantool writes to stdout and exits.

**Actual behavior**:   tarantool writes to stdout and enters interactive mode.

akudiyar:
Faced the same issue with Tarantool 2.3

[...]

PR #8711, opened 2023-05-29
Rework Tarantool CLI options and default behaviour

igormunkin:
These are two patches changing the default Tarantool CLI behaviour and introducing new CLI flags required for centralized configuration.

The first patch disables Lua REPL by default. Previously Tarantool binary entered interactive mode. As a result of the patch Tarantool binary yields the message with usage and doesn't run Lua interpreter.

The second patch introduces new CLI options for conf module:
* `--name` (`-n`) to specify instance name to be started. The option can be omitted in case `TT_INSTANCE_NAME` environment variable is set.
* `--config` (`-c`) to specify the patch to the config file. If the option is not set the value of `TT_CONFIG` environment variable is considered. If none of the above is specified, the default path is used.

---

@Mons, could you please share the desired help message for Tarantool 3.0?


src/main.cc
```diff
@@ -766,6 +773,16 @@ main(int argc, char **argv)
 	if (strlen(tarantool_path) < strlen(tarantool_bin))
 		panic("executable path is trimmed");
 
+	if (script == NULL && !(opt_mask & O_INTERACTIVE)) {
```
ligurio: Should we test new behavior?

changelogs/unreleased/gh-8613-new-cli-options.md
```diff
@@ -0,0 +1,3 @@
+## feature/cli
```
ligurio, line 1: Typo in commit message: "patch" -> "path"

changelogs/unreleased/gh-8613-new-cli-options.md
```diff
@@ -0,0 +1,3 @@
...
+
+* Introduced new command-line options --name and --config (gh-8613).
```
ligurio, line 3: Should we test new options?

ligurio (commented):
Igor, thanks for the patches! See my comments.

changelogs/unreleased/gh-8613-new-cli-options.md
andreyaksenov, line 3:
```suggestion
* Introduced new command-line options `--name` and `--config` (gh-8613).
```

changelogs/unreleased/gh-8613-stop-running-repl-by-default.md
```diff
@@ -0,0 +1,3 @@
+## feature/cli
+
+* Stop running Lua REPL by default and yield the message with usage (gh-8613).
```
andreyaksenov, line 3: It would be convenient to have the information in the initial ticket on how to achieve the old behavior (use `--repl` option?).

src/main.cc
igormunkin: I believe this is tested via [this test](https://github.com/tarantool/tarantool/blob/fb040252a51d590a3e9c3a0afe6a2560b069aaec/test/app-tap/gh-5040-inter-mode-isatty.test.lua#L35-L38) case, that is updated respectively in the patch, isn't it? All conditions are met: no args (hence no script provided) and interactive mode is disabled. Otherwise do you suggest to add a separate test to check that Tarantool yields usage message for this case?

changelogs/unreleased/gh-8613-new-cli-options.md
igormunkin: Fixed, thanks!

changelogs/unreleased/gh-8613-stop-running-repl-by-default.md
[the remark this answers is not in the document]
igormunkin: Fixed, thanks!

igormunkin: Added this line:
> To run Lua REPL just set `-i` flag.

changelogs/unreleased/gh-8613-new-cli-options.md
igormunkin: Fixed, thanks!

src/main.cc
ligurio: > I believe this is tested via [this test](https://github.com/tarantool/tarantool/blob/fb040252a51d590a3e9c3a0afe6a2560b069aaec/test/app-tap/gh-5040-inter-mode-isatty.test.lua#L35-L38) case, that is updated respectively in the patch, isn't it? All conditions are met: no args (hence no script provided) and interactive mode is disabled.

You are right. But I don't like that testing of default behavior is a part of regression test for #5040. It is not obvious that we test Tarantool's default behavior in an updated testcase. Moreover, a patch for an existed test doesn't check the error message.

> Otherwise do you suggest to add a separate test to check that Tarantool yields usage message for this case?

I would add a separate test, but I will not insist.

src/main.cc
```diff
@@ -834,8 +880,8 @@ main(int argc, char **argv)
 		 * is why script must run only after the server was fully
 		 * initialized.
 		 */
-		if (tarantool_lua_run_script(script, opt_mask, optc, optv,
-					     main_argc, main_argv) != 0)
+		if (tarantool_lua_run_script(script, &instance, opt_mask, optc,
```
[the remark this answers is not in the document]
igormunkin: Thanks for testing! I'm working on the tests for the new options (as you've asked above), so this one definitely goes to the test cases.

src/main.cc
igormunkin: Added the separate file with the tests for the new options.

src/main.cc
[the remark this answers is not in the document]
igormunkin: Also added tests for these cases.

test/box-py/args.result
```diff
@@ -1,41 +1,81 @@
...
-  -v, --version			print program version and exit
-  -e EXPR			execute string 'EXPR'
-  -l NAME			require library 'NAME'
-  -j cmd			perform LuaJIT control command
-  -b ...			save or list bytecode
-  -d				activate debugging session for 'SCRIPT'
-  -i				enter interactive mode after executing 'SCRIPT'
-  --				stop handling options
-  -				execute stdin and stop handling options
+Tarantool Server 3.<minor>.<patch>-<suffix>
```
ligurio, line 2: Why "Tarantool Server" and not "Tarantool"? Is it a requirement from a product team?

ligurio, line 2: I would filter major version too, otherwise test will break after bumping major version.

ligurio (approved):
LGTM

changelogs/unreleased/gh-8613-new-cli-options.md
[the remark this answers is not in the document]
igormunkin: Ouch, I've lost this change while rebasing, thanks!

test/box-py/args.result
igormunkin: It was originally proposed by @Mons, but we've finally decided to omit this to make in consistent with the interactive console prompt.

igormunkin: Well... Honestly, I don't want to touch this file a lot within this patchset, so I prefer to follow the existing practice regarding version match. Besides, I strongly believe such decision should be made by @ylobankov or his pals.

src/main.cc
```diff
@@ -619,30 +619,32 @@ print_version(void)
...
-	fputs("  -\t\t\t\texecute stdin and stop handling options", stream);
-	fputs("", stream);
-	fputs("Please visit project home page at https://tarantool.org", stream);
-	fputs("to see online documentation, submit bugs or contribute a patch.",
-	      stream);
+	static const char help_msg[] = "Tarantool Server %s\n\n"
+		"Connect to an instance:\n\n"
+		"  tt connect <uri>\n\n"
+		"Execute Lua script with bundled LuaJIT:\n\n"
+		"  %s old-style-script.lua\n\n"
```
Totktonada, line 628: `[ARGS]` is missed, so `--` below looks like it is redundant.

src/lua/init.h
```diff
@@ -43,6 +43,14 @@ extern bool start_loop;
 
 extern struct lua_State *tarantool_L;
 
+/* Struct with the information for instance configuration. */
+struct instance_state {
+	/* Instance name to be started. */
+	const char *name;
+	/* Patch to configuration file. */
+	const char *config;
```
Totktonada, line 51: Typo: patch -> path.

src/main.cc
```diff
@@ -699,9 +701,16 @@ main(int argc, char **argv)
...
 			optind--;
 			break;
+		case 'e':
+			/*
+			 * XXX: If no script is specified, then
+			 * the oneliner passed via -e flag is
+			 * considered as a script.
+			 */
+			script = optarg;
+			FALLTHROUGH;
```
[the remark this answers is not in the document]
Totktonada: My proposal: introduce an `O_EXECUTE` flag for `opt_mask` instead of using `script` for it.

test/app-tap/gh-5040-inter-mode-isatty.test.lua
```diff
@@ -35,7 +35,7 @@ local testcases = {
         cmd_args = '',
         stdin = '/dev/null',
         interactive = false,
-        empty_output = true
+        empty_output = false
```
Totktonada, line 38: I suspect that `echo 'print(42)' | tarantool` is unintentionally affected. When `isatty(0) == 0` the script should be read from stdin. I guess this behavior is not supposed to be changed.

test/box-py/args.result
igormunkin: Fixed.

src/main.cc
```diff
@@ -766,6 +810,18 @@ main(int argc, char **argv)
 	if (strlen(tarantool_path) < strlen(tarantool_bin))
 		panic("executable path is trimmed");
 
+	if (script == NULL && instance.name == NULL &&
+	    !(opt_mask & O_INTERACTIVE)) {
+		static const char misuse_msg[] = "Invalid usage:"
+			" please either provide a Lua script name"
+			" or specify an instance name to be started"
+			" or set -i CLI flag to spawn Lua REPL\n\n"
```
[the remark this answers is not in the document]
igormunkin: Fixed, thanks!

test/app-luatest/gh_8613_centralized_config_options_test.lua
```diff
@@ -0,0 +1,137 @@
...
+    {
+        args = imerge(),
+        env = merge(args.name.env, args.config.env),
+    },
+}
+
+g.test_tarantool_run = function()
+    for _, case in pairs(cases) do
+        local result = justrun('.', case.env, case.args, {nojson = true})
+        t.assert_equals(result.exit_code, 0, 'Plain run is successfull')
```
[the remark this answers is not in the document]
igormunkin: Fixxxed, thanks! ;)

src/lua/init.h
igormunkin: Fixed, thanks!

test/app-tap/gh-5040-inter-mode-isatty.test.lua
igormunkin: Discussed in group. Here is the decision:
1. `echo 'print(42)' | tarantool` yields usage message
2. `echo 'print(42)' | tarantool -` yields `42` (i.e. executes Lua script)
3. The change should be mentioned in a ChangeLog entry.

src/main.cc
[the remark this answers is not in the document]
igormunkin: Implemented your proposal.

src/main.cc
igormunkin: Added, thanks!

test/app-tap/gh-5040-inter-mode-isatty.test.lua
Totktonada: Shouldn't we fix the test case by changing the `cmd_args` to `' -'` (and keep `empty_output = true`) to don't change the meaning of the test case?

src/main.cc
Totktonada: Hm. There is the following scenario:

```
%s old-style-script.lua [ARGS]
```

It has no `[OPTIONS]`, but, it seems, it should.

And there is the separate block:

```
Usage:
%s [OPTIONS] [SCRIPT [ARGS]]
```

Which duplicates the old-style-script scenario, but describes the CLI arguments correctly.

I think we should remove the 'Usage:' block and incorporate `[OPTIONS]` into the old-style-script block.

src/main.cc
```diff
@@ -766,6 +810,18 @@ main(int argc, char **argv)
 	if (strlen(tarantool_path) < strlen(tarantool_bin))
 		panic("executable path is trimmed");
 
+	if (script == NULL && instance.name == NULL &&
+	    !(opt_mask & O_INTERACTIVE)) {
+		static const char misuse_msg[] = "Invalid usage:"
+			" please either provide a Lua script name"
+			" or specify an instance name to be started"
```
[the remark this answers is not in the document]
Totktonada: Not exactly. The config file can be omitted, but when another condition is met: there is a source that feeds a cluster config with the given instance present in the topology. Tarantool Enterprise Edition has etcd source that can be configuured w/o a config file using environemnt variables. A custom cluster config source can be implemented by users.

We're not going to implement the default system-wide config for now.

Anyway, everything said above is handled by the config module code and I think that it closes the question.

test/app-luatest/gh_8613_new_cli_behaviour_test.lua
```diff
@@ -0,0 +1,106 @@
...
+local prompt = {
+    output = 'tarantool>',
+    pattern = false,
+}
+
+local script = string.dump(function() print(_VERSION) end)
+
+local rundir
+g.before_each(function()
+    treegen.init(g)
```
Totktonada, line 27: There should be a `treegen.cleanup()` call somewhere (`g.after_all()` or `g.after_each()`). Otherwise the temporary directories will stay there.

test/app-luatest/gh_8613_new_cli_behaviour_test.lua
```diff
@@ -0,0 +1,106 @@
...
+    t.assert_equals(result.exit_code, 1,
+                    'Running Tarantool w/o CLI arguments fails')
+    t.assert_str_contains(result.stderr, misuse.output, misuse.pattern,
+                          'Running Tarantool w/o CLI arguments yields usage')
+end
+
+g.test_tarantool_stdin_invalid_redirect = function()
+    local result = simplerun({'</dev/null'})
+    t.assert_equals(result.exit_code, 1,
+                    'Redirect to /dev/null w/o dash in CLI args')
```
Totktonada, line 47: to -> from?

src/main.cc
```diff
@@ -774,6 +778,16 @@ main(int argc, char **argv)
 	if (strlen(tarantool_path) < strlen(tarantool_bin))
 		panic("executable path is trimmed");
 
+	if (script == NULL && (opt_mask & (O_INTERACTIVE | O_EXECUTE)) == 0) {
+		static const char misuse_msg[] = "Invalid usage:"
+			" please either provide a Lua script name"
+			" or set -i CLI flag to spawn Lua REPL.\n\n"
+			;
```
Totktonada: Nit: I would format it like the help message below: with newlines instead of a single long line.

changelogs/unreleased/gh-8613-new-cli-options.md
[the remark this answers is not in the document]
Totktonada: It seems resolved and I'm going to resolve the thread.

src/main.cc
Totktonada: Looks resolved. Resolving the thread so.

changelogs/unreleased/gh-8613-new-cli-options.md
Totktonada: It seems solved. Resolving the thread.

Totktonada (approved):

changelogs/unreleased/gh-8613-stop-running-repl-by-default.md
```diff
@@ -0,0 +1,7 @@
+## feature/cli
+
+* **[Behavior change]** Disabled starting the Lua REPL by default when running
+  Tarantool. Now, Tarantool yields the message that shows the command usage.
+  To run Lua REPL just set `-i` flag. To pass Lua script contents via stdin,
+  use dash (-) as a SCRIPT name. For more info see help message by running
+  `tarantool -h` (gh-8613).
```
andreyaksenov, line 7:
```suggestion
  To run the Lua REPL, set the `-i` flag. To pass a Lua script contents via `stdin`,
  use dash (`-`) as the script name. For more information, see a help message by running
  `tarantool -h` (gh-8613).
```

andreyaksenov (approved):

src/main.cc
igormunkin: OK, no problem. Fixed.

test/app-luatest/gh_8613_new_cli_behaviour_test.lua
igormunkin: OK, if it makes the message clearer. Fixed.

src/main.cc
igormunkin: Added `[OPTIONS]` for "old style", since some options definitely can be used in this case. All other changes will be made in scope of another PR.

changelogs/unreleased/gh-8613-stop-running-repl-by-default.md
igormunkin: Fixed, thanks!

PR #8713, opened 2023-05-30
test: regenerate snapshots for #7974

### LINKED ISSUE #8702: Regenerate `test/box-luatest/gh_7974_data`
The [directory](https://github.com/tarantool/tarantool/tree/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/gh_7974_data) contains snap and xlog files that are used to start a test instance. The files use an old schema (the schema was upgraded after they were generated). In the scope of #7149 we disable any DDL with an old schema version. The problem is luatest needs to grant permissions (which is a DDL operation) to run tests (it does it in the test server instance source file [here](https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/server_instance.lua#L47)). As a result, the [test](https://github.com/tarantool/tarantool/blob/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/gh_7974_force_recovery_bugs_test.lua) fails while trying to start a test server instance.

To fix this issue, we need to regenerate the data files with `box.schema.user.grant('guest', 'super')`. Unfortunately, there's no script that could be used for this (like [this one](https://github.com/tarantool/tarantool/blob/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead/test/box-luatest/upgrade/2.10.4/gen.lua), for example) so they have to be regenerated manually.

It'd be nice if in the scope of this ticket we'd also add a script that would regenerate the data files automatically.

CuriousGeorgiy:
Closes #8702


locker:
Superseded by #8722.

PR #8715, opened 2023-05-31
box: allow to set *_uuid options to NULL

### LINKED ISSUE #8714: Reapplying the instance_uuid and replicaset_uuid parameters with a box.NULL field results in an error.
Prior to https://github.com/tarantool/tarantool/commit/9b2b3e58792fb1d5b079b110d50e7c5803803a4d, reapplying `instance_uuid` and `replicaset_uuid` did not throw an error if the corresponding values were `NULL`:
```
Tarantool 3.0.0-entrypoint-195-g2644add4e
tarantool> box.cfg{log_level = 0}
---
...

tarantool> box.cfg.replicaset_uuid
---
- null
...

tarantool> box.cfg.instance_uuid
---
- null
...

tarantool> box.cfg{replicaset_uuid = nil}
---
...

tarantool> box.cfg{instance_uuid = nil}
---
...

tarantool> box.cfg{replicaset_uuid = box.NULL}
---
...

tarantool> box.cfg{instance_uuid = box.NULL}
---
...
```

However this has changed after the patch:
```
Tarantool 3.0.0-entrypoint-272-g2afde5b1d
tarantool> box.cfg{log_level = 0}
---
...

tarantool> box.cfg.replicaset_uuid
---
- null
...

tarantool> box.cfg.instance_uuid
---
- null
...

tarantool> box.cfg{replicaset_uuid = nil}
---
...

tarantool> box.cfg{instance_uuid = nil}
---
...

tarantool> box.cfg{replicaset_uuid = box.NULL}
---
- error: Can't set option 'replicaset_uuid' dynamically
...

tarantool> box.cfg{instance_uuid = box.NULL}
---
- error: Can't set option 'instance_uuid' dynamically
...
```

The new behavior is incorrect and should be fixed.

ImeevMA:
This patch allows to set replicaset_uuid and instance_uuid to box.NULL. This fixes the issue described in #8714, however it introduces another change in behavior - we can now set these parameters to NULL even if they weren't NULL before. However, since we still cannot set a different uuid after setting the parameters to NULL, and we can still set the old uuid for them, this behavior is considered acceptable.

Closes #8714

NO_DOC=bugfix
NO_CHANGELOG=the bug was not released

### DIFF
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -446,13 +446,15 @@ local log_cfg_option = {
 
 
 local function check_instance_uuid()
-    if box.cfg.instance_uuid ~= box.info.uuid then
+    if box.cfg.instance_uuid ~= box.NULL and
+       box.cfg.instance_uuid ~= box.info.uuid then
         box.error(box.error.RELOAD_CFG, 'instance_uuid')
     end
 end
 
 local function check_replicaset_uuid()
-    if box.cfg.replicaset_uuid ~= box.info.replicaset.uuid then
+    if box.cfg.replicaset_uuid ~= box.NULL and
+       box.cfg.replicaset_uuid ~= box.info.replicaset.uuid then
         box.error(box.error.RELOAD_CFG, 'replicaset_uuid')
     end
 end
--- a/test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
+++ b/test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
@@ -1,0 +1,78 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_test('test_set_null_to_xxx_uuid', function()
+    g.server = server:new({
+        alias = 'master',
+        box_cfg = {
+            instance_uuid = box.NULL,
+            replicaset_uuid = box.NULL,
+        },
+    })
+    g.server:start()
+end)
+
+g.after_test('test_set_null_to_xxx_uuid', function()
+    g.server:stop()
+end)
+
+g.test_set_null_to_xxx_uuid = function()
+    g.server:exec(function()
+        local ok = pcall(box.cfg, {instance_uuid = box.NULL})
+        t.assert(ok)
+        ok = pcall(box.cfg, {replicaset_uuid = box.NULL})
+        t.assert(ok)
+    end)
+end
+
+g.before_test('test_set_uuid_to_xxx_uuid', function()
+    g.server = server:new({
+        alias = 'master',
+        box_cfg = {
+            instance_uuid = '11111111-1111-1111-1111-111111111111',
+            replicaset_uuid = '11111111-1111-1111-1111-111111111111',
+        },
+    })
+    g.server:start()
+end)
+
+g.after_test('test_set_uuid_to_xxx_uuid', function()
+    g.server:stop()
+end)
+
+g.test_set_uuid_to_xxx_uuid = function()
+    g.server:exec(function()
+        local old_id = '11111111-1111-1111-1111-111111111111'
+        local new_id = '11111111-1111-1111-1111-111111111112'
+        local err_ins = "Can't set option 'instance_uuid' dynamically"
+        local err_rpl = "Can't set option 'replicaset_uuid' dynamically"
+
+        -- The same uuid can be set if *_uuid is set.
+        local ok = pcall(box.cfg, {instance_uuid = old_id})
+        t.assert(ok)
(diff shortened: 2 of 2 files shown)


test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
```diff
@@ -0,0 +1,82 @@
...
+    g.server:start()
+    g.server:exec(function()
+        local ok, err = pcall(box.cfg, {instance_uuid = box.NULL})
+        t.assert(ok)
+        t.assert_equals(err, nil)
+        ok, err = pcall(box.cfg, {replicaset_uuid = box.NULL})
+        t.assert(ok)
+        t.assert_equals(err, nil)
+    end)
+    g.server:stop()
```
locker, line 23: If the test fails, the server won't be stopped properly. You should use `before_test` to start the server and `after_test` to stop it, because `after_test` runs even on test failure.

test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
```diff
@@ -0,0 +1,82 @@
...
+        t.assert(ok)
+        t.assert_equals(err, nil)
+        ok, err = pcall(box.cfg, {replicaset_uuid = old_uuid})
+        t.assert(ok)
+        t.assert_equals(err, nil)
+
+        -- Another uuid cannot be set if *_uuid is set.
+        ok, err = pcall(box.cfg, {instance_uuid = new_uuid})
+        t.assert(not ok)
+        t.assert_equals(err.message, err_ins)
```
locker, line 52: Better use `t.assert_error_message_equals`:

```Lua
t.assert_error_message_equals(err_ins, box.cfg, {instance_uuid = new_uuid})
```

test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
ImeevMA: Thank you for the review! Fixed.

test/box-luatest/gh_8714_set_null_to_xxx_uuid_test.lua
ImeevMA: Fixed.

locker (approved):

src/box/lua/load_cfg.lua
```diff
@@ -446,13 +446,15 @@ local log_cfg_option = {
 
 
 local function check_instance_uuid()
-    if box.cfg.instance_uuid ~= box.info.uuid then
+    if box.cfg.instance_uuid ~= box.NULL and
+       box.cfg.instance_uuid ~= box.info.uuid then
         box.error(box.error.RELOAD_CFG, 'instance_uuid')
     end
 end
```
Totktonada, line 453: Shouldn't we handle `nil` and `box.NULL` differently?

* `nil` -> pass over
* `box.NULL` -> check that `box.cfg.instance_uuid == nil`
* anything else -> compare with `box.info.uuid` (BTW, why not `box.cfg.instance_uuid`?)

ImeevMA: We don't have the old `box.cfg.instance_uuid` value because it was overwritten by the new one when `box.cfg{instance_uuid = ...}` was executed. I believe this is normal for the options described in `dynamic_cfg` . Because of this, we have no way to compare the old value with the new one.

locker: - The option callback isn't invoked if `box.cfg.instance_uuid` is `nil`.
- It'd be complicated because the `box.cfg` option callback currently doesn't have access to the old value. I don't think that fixing it is worthwhile. Semantically, resetting `instance_uuid` to `box.NULL` after it was set to a non-nil value looks correct, which was mentioned in the commit log.
- If the new value stored in `box.cfg.instance_uuid` isn't nil, we compare it with the current value, which is stored in `box.info.uuid`.

src/box/lua/load_cfg.lua
```diff
@@ -446,13 +446,15 @@ local log_cfg_option = {
 
 
 local function check_instance_uuid()
-    if box.cfg.instance_uuid ~= box.info.uuid then
+    if box.cfg.instance_uuid ~= box.NULL and
```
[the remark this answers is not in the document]
locker: The option callback isn't invoked in `box.cfg.instance_uuid` is `nil`. The only way to set `box.cfg.instance_uuid` to `nil` is using `box.NULL` so the comparison looks proper.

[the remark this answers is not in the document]
Totktonada: This is minor and since the previous comment is resolved, I'll close this too.

Totktonada (approved):

PR #8578, opened 2023-04-17
util: introduce new helper to generate Lua translation table

drewdzzz:
Draft.

Issue #8443, opened 2023-03-15
Export IPROTO constants to Lua automatically


locker:
To export IPROTO constants to Lua, we define a translation table in C manually:

https://github.com/tarantool/tarantool/blob/0ead015b252748b8493370bbfae08a64bc293e3b/src/box/lua/iproto.c#L85-L87

When a new IPROTO constant is added, the translation table has to be updated:

https://github.com/tarantool/tarantool/blob/0ead015b252748b8493370bbfae08a64bc293e3b/src/box/iproto_constants.h#L65-L70

This is inconvenient and error-prone. Instead we should generate the translation table automatically when the project is built.

It'd be nice to have a generic mechanism for that because there are other places where we need to export C enumeration constants to Lua, for example `box.error`.

```C
LUA_EXPORT(namespace = "box.iproto.key", strip_prefix = "IPROTO_")
enum iproto_key {
        IPROTO_REQUEST_TYPE = 0x00,
        IPROTO_SYNC = 0x01,
```

PR #8717, opened 2023-05-31
Refactor the code that generates names for IPROTO constants

locker:
This PR finishes the work started by #8578 that introduced automatic IPROTO constants exporting to Lua:
- Strips prefixes from the generated names because we strip them anyway when we export the IPROTO constants to Lua.
- Adds name-generating macros for Vinyl constants defined in `iproto_constants.h`.
- Drops `struct iproto_constant` in favor of string arrays because we need them anyway for looking up a name by code.
- Makes `net.box` use `box.iproto.feature` instead of defining feature names manually.

Follow-up #8443


locker:
Cherry-picked to 2.11.

Issue #8516, opened 2023-03-29
Backporting LuaJIT upstream commits part 4


igormunkin:
This ticket is an umbrella for all commits we planning to back-port in the next Q.
This list includes commits excluded from #8069 and may be updated later with some new.

GC64:
* [x] [dd0f09f9](https://github.com/LuaJIT/LuaJIT/commit/dd0f09f9) Another fix for lua_yield() from C hook.
* [x] [dd5032ed](https://github.com/LuaJIT/LuaJIT/commit/dd5032ed) Fix lua_yield() from C hook.
* [x] [850f8c59](https://github.com/LuaJIT/LuaJIT/commit/850f8c59) LJ_GC64: Make ASMREF_L references 64 bit.

general + x86/x64:
* [x] [7dbf0b05](https://github.com/LuaJIT/LuaJIT/commit/7dbf0b05) Fix saved bytecode encapsulated in ELF objects.
* [x] [d4c0c6e1](https://github.com/LuaJIT/LuaJIT/commit/d4c0c6e1) Fix TNEW load forwarding with instable types.
* [x] [96fc114a](https://github.com/LuaJIT/LuaJIT/commit/96fc114a) Fix canonicalization of +-0.0 keys for IR_NEWREF.
* [x] [2801500a](https://github.com/LuaJIT/LuaJIT/commit/2801500a) Fix use-def analysis for BC_VARG.
* [x] [0e53a314](https://github.com/LuaJIT/LuaJIT/commit/0e53a314) Fix use-def analysis for vararg functions.
* [x] [646148e7](https://github.com/LuaJIT/LuaJIT/commit/646148e7) Fix memory probing allocator to check for valid end address, too.
* [x] [03a7ebca](https://github.com/LuaJIT/LuaJIT/commit/03a7ebca) x64: Fix 64 bit shift code generation.

PR #8720, opened 2023-05-31
luajit: bump new version

mkokryashkin:
Fix memory probing allocator to check for valid end address, too.

Part of #8516

NO_DOC=LuaJIT bump
NO_TEST=LuaJIT bump


igormunkin:
The changeset is merged to master in scope of #8843, backported to 2.11 and 2.10 in scope of #8844 and #8845 respectively.

PR #8721, opened 2023-05-31
test: fix flaky feedback_daemon_metrics test

drewdzzz:
The patch fixes two mistakes that were made while writing the test. Firstly, we should replace metrics module with an empty table before every test case - otherwise, metrics can be collected before the test. Secondly, all attempts to check if a required amount of metrics was collected is pointless, even with margin - some environments are incredibly slow. So let's check if some metrics were collected, and check if there are not too many of them.

Also, the patch increases some timeouts to minimize the probability of fail due to slow environment.

Follows up #8192
Closes tarantool/tarantool-qa#312

### DIFF
--- a/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
+++ b/test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
@@ -28,6 +28,8 @@ end)
 
 g.before_each(function()
     g.server:exec(function()
+        -- Replace module metrics with empty table to avoid metrics collection.
+        package.loaded["metrics"] = {}
         -- Daemon is reloaded when cfg option is updated.
         box.cfg{
             feedback_send_metrics = true,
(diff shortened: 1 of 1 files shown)


test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
```diff
@@ -28,6 +28,8 @@ end)
 
 g.before_each(function()
     g.server:exec(function()
+        -- Replace module metrics with empty table to avoid metrics collection.
+        package.loaded["metrics"] = {}
```
CuriousGeorgiy, line 32: Why not simply `nil`?

CuriousGeorgiy: But you don't call `require` here. And if it's required internally by `box.cfg`, then I don't understand how it works with an empty table:
https://github.com/tarantool/tarantool/blob/606e50c4d56bb76714dc41339102b32230fd751e/src/box/lua/load_cfg.lua#L529

CuriousGeorgiy: I tried to change it to `nil` and run the test — everything works fine.

CuriousGeorgiy (approved):

test/box-luatest/gh_8192_feedback_daemon_metrics_test.lua
```diff
@@ -54,24 +56,22 @@ g.test_metrics = function()
...
-        -- ASAN and GCOV builds looks very affected and macOS too.
-        --
-        -- In fact, this check verifies that *some* samples are collected. It
+        -- This check verifies that *some* samples are collected. It
         -- doesn't attempt to ensure correct timings, which is nearly
         -- impossible in the general case, when many processes performs its own
         -- workloads in parallel.
-        t.assert_almost_equals(#fb.metrics, 30, 25)
+        t.assert_gt(#fb.metrics, 0)
+        t.assert_lt(#fb.metrics, 60)
```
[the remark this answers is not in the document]
drewdzzz: I wrote about it in the commit message, but agree, it's better to repeat it here. Done.

mkostoevr (approved):

locker (approved):

locker:
Cherry-picked to 2.11.

PR #8722, opened 2023-05-31
test: write bad snap with errinj in `gh_7974_force_recovery_bugs_test`

### LINKED ISSUE #8702: Regenerate `test/box-luatest/gh_7974_data`

locker:
Corrupted snap files that are used in the test were generated manually using a now old Tarantool version that has an outdated system schema. In the scope of #7149 DDL was forbidden until the system schema is upgraded. The problem is luatest tries to grant super privileges to the guest user (which is a DDL operation) after starting a test instance unless they are already granted. Since the snap files don't store the required privileges, luatest fails.

To fix this issue, let's generate corrupted snap files right in the test using error injection.

Closes #8702


test/box-luatest/gh_7974_force_recovery_bugs_test.lua
```diff
@@ -5,249 +5,247 @@ local t = require('luatest')
 local g = t.group()
 
 g.before_all(function()
-    t.skip('gh-8702')
+    t.tarantool.skip_if_not_debug()
```
CuriousGeorgiy, line 8: So now we can’t be sure force recovery works in release?

locker: https://github.com/tarantool/tarantool/issues/8723

alyapunov (approved):

Issue #8724, opened 2023-05-31
Declarative server and cluster configuration


Totktonada:
Tarantool has two main operation modes:

* A script interpreter with a handy set of modules.
* Interactive REPL for debugging pleasure (it is default -- enabled when no command-line parameters are passed).

And, if we add tarantoolctl/tt tooling here:

* Start tarantool application as a service (server/instance/daemon).
* Connect to a remote instance with interactive REPL.

The main scenario is to start a service. And we want to support it natively.

A service usually has a declarative configuration: ini file, json or yaml config file.

Since Tarantool 1.6 our configuration way is to find APIs you need and write a script to:

* configure log and pid files
* configure data directories
* setup a database schema
* setup users/passwords/privileges
* bind binary protocol and console sockets
* configure the application
* ...

This approach is complicated for a user, increases amount of a boilerplace code, requires more guides how to cook it in the right way.

Tarantool 3.0 is 1.5\*2.0 -- all the new features plus a declarative configuration.

Most of Tarantool based applications are not a single tarantool process, but a replicaset or a cluster of several replicasets. Sometimes even a set of such clusters.

We should look at the configuration as on a cluster configuration, not configs for individual instances. So, the config will define:

* global options
* per-group (cluster) options
* per-replicaset options
* per-instance options
* group/replicaset/instance topology

What is supposed to be changed in Tarantool 3.0.

* The new 'start as a service' mode, which is enabled with the `--name` (instance name) and `--config` (configuration file) options.
* The new default mode: show available startup modes. `tarantool -i` still provides REPL.
* The new built-in `config` module to acquire and reload the configuration.

```[tasklist]
* [ ] https://github.com/tarantool/tarantool/issues/8613
* [ ] https://github.com/tarantool/tarantool/issues/8725
* [ ] https://github.com/tarantool/tarantool/issues/8778
* [ ] https://github.com/tarantool/tarantool/issues/8789
* [ ] https://github.com/tarantool/tarantool/issues/8810
```

Totktonada:
All the subtasks are finished.

PR #8726, opened 2023-06-01
config: add schema-aware data processing module

### LINKED ISSUE #8725: config: schema-aware hierarchical data processing
See #8724 for the umbrella task description.

The new configuration format is hierarchical. It allows to keep related options together and make the configuration simpler to read.

The configuration data format is defined by a schema, at least for validation purposes. And, when we have a schema, it is convenient to define several other data manipulation functions as schema methods.

The API is briefly described below.

<details>
<summary>API</summary>

Schema node constructors:

* Define a scalar.

  ```lua
  schema.scalar({
      type = 'string', -- number, integer, boolean
      <..annotations..>
  })
  ```
* Define a record (an object with certain field names and field value types).

  ```lia
  schema.record({
      foo = <schema node>,
  }, {
      <..annotations..>
  })
  ```
* Define a map (an object with arbitrary key names, but strict about keys and values types).

  ```lua
  schema.map({
      key = <schema node>,
      value = <schema node>,
      <..annotations..>
  })
  ```
* Define an array.

  ```lua
  schema.array({
    items = <..schema node..>,
    <..annotations..>
  ```

A several supplementary schema node constructors may be defined, for example `schema.enum()` and `schema.set()`.

* ```lua
  -- Accepts 'foo', 'bar' or 'baz'.
  schema.enum({
    'foo',
    'bar',
    'baz',
  }, {
      <..annotations..>
  })
  ```
* ```lua
  -- Accepts ['foo'], ['foo', 'bar'] and so on.
  schema.set({
    'foo',
    'bar',
    'baz',
  }, {
      <..annotations..>
  })
  ```
* (I also have `schema.union_of_records()` in a PoC, but I hope I'll replace it with a more general `schema.union()` schema node type.)

The schema object constructor wraps a schema node and adds schema name and user-provided methods.

```lua
schema.new('myschema', <schema node>)
```

The schema object has the following methods:

* Traversing the schema.

  ```lua
  <schema object>:pairs() -> luafun iterator
  ```

* Validate data against the schema.

  ```lua
  <schema object>:validate(data)
  ```
* Filter data based on the schema annotations.

  ```lua
  <schema object>:filter(data, f) -> luafun iterator
  ```
* Map data based on the schema annotations.

  ```lua
  <schema object>:map(data, f, f_ctx) -> new_data
  ```
* Apply default values.

  ```lua
  <schema object>:apply_default(data) -> new_data
  ```
* Get/set a nested value.

  ```lua
  <schema object>:get(data, path)
  <schema object>:set(data, path, value)
  ```
* Merge two values.

  ```lua
  <schema object>:merge(a, b)
  ```

The following annotations are interpreted by the module itself:

* type (string)
* validate (function)
* allowed_values (table)
* default (any)
* apply_default_if (function)

Other annotations are ignored by the module and may be used for arbitrary purposes.

It looks logical to have parsing of data from environment variables here too. Convert `MAYVAR=3301` to a Lua number, parse `TT_REPLICATION_PEERS=localhost:3301,localhost:3302,localhost:3303` to an array and so on.

```lua
schema.fromenv('MYVAR', os.getenv('MYVAR'), <schema>)
```
</details>

This module is planned as utilitary code for config purposes, at least within this task. No new public APIs will be added.

Part of #8724

Totktonada:
The module offers a schema definition format and provides a bunch of functions to validate and manipulate hierarchical data.

See the API description in #8725.

The module is part of the new declarative configuration initiative, see #8724.

This pull request doesn't add any public API: the module is for internal use and its API can be changed later.

Fixes #8725

### DIFF
--- a/src/box/lua/config/utils/schema.lua
+++ b/src/box/lua/config/utils/schema.lua
@@ -342,6 +342,23 @@ end
 
 -- }}} Schema node constructors: scalar, record, map, array
 
+-- {{{ Derived schema node type constructors: enum
+
+-- Shortcut for a string scalar with given allowed values.
+local function enum(allowed_values, annotations)
+    local scalar_def = {
+        type = 'string',
+        allowed_values = allowed_values,
+    }
+    for k, v in pairs(annotations or {}) do
+        assert(k ~= 'type' and k ~= 'allowed_values')
+        scalar_def[k] = v
+    end
+    return scalar(scalar_def)
+end
+
+-- }}} Derived schema node type constructors: enum
+
 -- {{{ <schema object>:validate()
 
 -- Verify that the given table adheres array requirements.
@@ -1550,6 +1567,13 @@ return {
     record = record,
     map = map,
     array = array,
+
+    -- Constructors for 'derived types'.
+    --
+    -- It produces a scalar, record, map or array, but annotates
+    -- it in some specific way to, say, impose extra constraint
+    -- rules at validation.
+    enum = enum,
 
     -- Schema object constructor.
     new = new,
--- a/src/box/lua/config/utils/schema.lua
+++ b/src/box/lua/config/utils/schema.lua
@@ -613,6 +613,114 @@ end
 
 -- }}} <schema object>:get()
 
+-- {{{ <schema object>:set()
+
+local function set_usage()
+    error('Usage: schema:set(data: <as defined by the schema>, ' ..
+        'path: string/table, rhs: <as defined by the schema>)', 0)
+end
+
+local function set_impl(schema, data, rhs, ctx)
+    -- The journey is finished. Validate and return the new value.
+    if #ctx.journey == 0 then
+        -- Call validate_impl() directly to don't construct a
+        -- schema object.
+        validate_impl(schema, rhs, ctx)
+        return rhs
+    end
+
+    local requested_field = ctx.journey[1]
+    assert(requested_field ~= nil)
+
+    if is_scalar(schema) then
+        walkthrough_error(ctx, 'Attempt to index a scalar value of type %s ' ..
+            'by field %q', schema.type, requested_field)
+    elseif schema.type == 'record' then
+        walkthrough_enter(ctx, requested_field)
+        local field_def = schema.fields[requested_field]
+        if field_def == nil then
+            walkthrough_error(ctx, 'No such field in the schema')
+        end
+
+        walkthrough_assert_table(ctx, schema, data)
+        local field_value = data[requested_field] or {}
+        table.remove(ctx.journey, 1)
+        data[requested_field] = set_impl(field_def, field_value, rhs, ctx)
+        return data
+    elseif schema.type == 'map' then
+        walkthrough_enter(ctx, requested_field)
+        local field_def = schema.value
+
+        walkthrough_assert_table(ctx, schema, data)
+        local field_value = data[requested_field] or {}
+        table.remove(ctx.journey, 1)
+        data[requested_field] = set_impl(field_def, field_value, rhs, ctx)
+        return data
+    elseif schema.type == 'array' then
+        -- TODO: Support 'foo[1]' and `{'foo', 1}` paths. See the
+        -- normalize_path() function.
+        walkthrough_error(ctx, 'Indexing an array is not supported yet')
+    else
+        assert(false)
+    end
+end
+
+-- Set the given `rhs` value at the given path in the `data`.
+--
+-- Important: `data` is assumed as already validated against the
+-- given schema, but `rhs` is validated by the method before the
+-- assignment.
+--
+-- The function checks the path against the schema: it doesn't
+-- allow to use a non-existing field or index a scalar value.
+--
+-- The path is either array-like table or a string in the dot
+-- notation.
+--
+-- local data = {}
+-- schema:set(data, 'foo.bar', 42)
+-- print(data.foo.bar) -- 42
+--
+-- local data = {}
+-- schema:set(data, {'foo', 'bar'}, 42)
+-- print(data.foo.bar) -- 42
+--
+-- Nuances:
+--
+-- * A root node (pointed by the empty path) can't be set using
+--   this method.
+-- * Array indexing is not supported yet.
+-- * A scalar of the 'any' type can't be indexed, even when it is
+--   a table. It is OK to set the whole value of the 'any'
+--   type.
+function methods.set(self, data, path, rhs)
+    local schema = rawget(self, 'schema')
+
+    -- Detect `data` and `path` misordering for the most common
+    -- scenario: a record schema and a string path.
+    if schema.type == 'record' and type(data) ~= 'table' then
+        return set_usage()
+    end
+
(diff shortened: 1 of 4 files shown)


test/config-luatest/schema_test.lua
```diff
@@ -0,0 +1,271 @@
...
+    -- No args, no key, no value.
+    --
+    -- Ignore error messages. They are just 'assertion failed at
+    -- line X', purely for the schema creator.
+    t.assert_equals((pcall(schema.map)), false)
+    t.assert_equals((pcall(schema.map, {})), false)
+    t.assert_equals((pcall(schema.map, {value = scalar_1})), false)
+    t.assert_equals((pcall(schema.map, {key = scalar_1})), false)
+
+    -- Bad case: 'type' annotatoin is forbidden.
```
igormunkin, line 170: Typo: s/annotatoin/annotation/.

test/config-luatest/schema_test.lua
```diff
@@ -0,0 +1,271 @@
...
+    -- Simple bad cases.
+    --
+    -- No args, no items.
+    --
+    -- Ignore error messages. They are just 'assertion failed at
+    -- line X', purely for the schema creator.
+    t.assert_equals((pcall(schema.array)), false)
+    t.assert_equals((pcall(schema.array, {})), false)
+
+    -- Bad case: 'type' annotatoin is forbidden.
```
igormunkin, line 216: Typo: s/annotatoin/annotation/.

igormunkin (approved):
__[PATCH 1/14] config/schema: add schema and schema nodes ctors__

src/box/lua/config/utils/schema.lua
```diff
@@ -5,33 +5,156 @@ local schema_mt = {}
...
+    validate_noexc = function(data)
+        return validate_type_noexc(data, 'string')
+    end,
 }
 
 scalars.number = {
     type = 'number',
+    validate_noexc = function(data)
+        -- TODO: Should we accept cdata<int64_t> and
+        -- cdata<uint64_t> here?
```
igormunkin, line 174: Side note: I would rather leave this for `integer` validator.

igormunkin (approved):
__[PATCH 2/14] config/schema: add <schema object>:validate()__

igormunkin (approved):
__[PATCH 3/15] config/schema: support allowed_values annotation__

igormunkin (approved):
__[PATCH 4/15] config/schema: support validate annotation__

igormunkin (approved):
__[PATCH 5/14] config/schema: add <schema object>:get()__

src/box/lua/config/utils/schema.lua
```diff
@@ -613,6 +613,112 @@ end
...
+    if schema.type == 'record' and type(data) ~= 'table' then
+        return set_usage()
+    end
+
+    if path ~= nil then
+        path = normalize_path(path, set_usage)
+    end
+
+    if path == nil or next(path) == nil then
+        error('schema:set: empty path', 0)
```
igormunkin, line 848: Does it mean, the root of the schema can't be set via this method? AFAIU, the reason is the way `set_impl` is implemented.

I see this behaviour is tested below, but it would be nice to know the reason.

igormunkin (approved):
__[PATCH 6/14] config/schema: add <schema object>:set()__

src/box/lua/config/utils/schema.lua
```diff
@@ -719,6 +721,146 @@ end
 
 -- }}} <schema object>:set()
 
+-- {{{ <schema object>:filter()
+
+local filter_impl
+filter_impl = function(schema, data, f, ctx)
```
igormunkin, line 865: Minor: Why do you need such a complex definition here, but not for `get_impl`/`set_impl`?

igormunkin (approved):
__[PATCH 7/14] config/schema: add <schema object>:filter()__

igormunkin (approved):
__[PATCH 8/15] config/schema: add <schema object>:filter()__

igormunkin (approved):
__[PATCH 10/15] config/schema: add <schema object>:merge()__

igormunkin (approved):
__[PATCH 11/15] config/schema: add <schema object>:pairs()__

@Totktonada, thanks for the patch! LGTM, but it would be nice to clarify `pairs` semantics.

igormunkin (approved):
__[PATCH 13/15] config/schema: schema.enum()__

igormunkin (approved):
__[PATCH 14/15] config/schema: schema.set()__

igormunkin (approved):
__[PATCH 15/15] config/schema: add string-and-number scalar type__

test/config-luatest/schema_test.lua
Totktonada: Sent https://github.com/tarantool/checkpatch/pull/61

Totktonada: Fixed.

test/config-luatest/schema_test.lua
Totktonada: Fixed.

src/box/lua/config/utils/schema.lua
Totktonada: There are two possible APIs for this method:

```lua
local res = s:set(res, path, rhs)
```

```lua
s:set(res, path, rhs)
```

The former allows to set the root, but requires to remember about the assignment of the result back to `res` if `path` may point to the root. Each of the APIs has its cons and pros.

The latter doesn't allow to set the root. At least, not always -- the root should be an existing table. And not easily -- we should handle this case separately by removing all the fields and set all the new fields.

My choice is the latter API w/o ability to set the root node and the choice was more or less arbitrary. It appears to be good enough for my goals, so I left it implemented this way.

Added a bullet into the 'nuances' section of the method description to make it explicit:

```diff
diff --git a/src/box/lua/config/utils/schema.lua b/src/box/lua/config/utils/schema.lua
index 9b253d49e8..02f5e31aba 100644
--- a/src/box/lua/config/utils/schema.lua
+++ b/src/box/lua/config/utils/schema.lua
@@ -687,6 +687,8 @@ end
 --
 -- Nuances:
 --
+-- * A root node (pointed by the empty path) can't be set using
+--   this method.
 -- * Array indexing is not supported yet.
 -- * A scalar of the 'any' type can't be indexed, even when it is
 --   a table. It is OK to set the whole value of the 'any'
```

I left it without the motivation description, because, in fact, there is no partucular motivation and it should be OK to change it if we'll feel the need.

src/box/lua/config/utils/schema.lua
Totktonada: Thanks for catching it! All the recursive functions were defined this was in my early PoC, then I changed it to just `local function <...>_impl(<...>) <...> end`. But forgot to change this function.

Fixed.

igormunkin (approved):
__[PATCH 9/15] config/schema: add <schema object>:apply_default()__

igormunkin (approved):
__[PATCH 12/15] config/schema: schema.fromenv()__

PR #7010, opened 2022-04-08
tx: set session and user for commit/rollback triggers

locker:
Commit/rollback triggers are run asynchronously, upon receiving the write status from WAL. We can't run them in the original fiber that submitted the WAL request, because it would open a time window between writing a transaction to WAL and committing it in tx, which could lead to violating the cascading rolback principles. As a result, commit/rollback triggers run with admin privileges.

Let's fix this issue by temporarily setting session and credentials to the original fiber's for running commit/rollback triggers.

Closes #7005

PR #8736, opened 2023-06-04
limbo: fix commit/rollback failures with triggers

### LINKED ISSUE #8505: ER_CURSOR_NO_TRANSACTION in on_commit trigger for syncro space
**Bug description**

```
main/103/app.lua init.c:376 E> ER_CURSOR_NO_TRANSACTION: The transaction the cursor belongs to has ended
main/103/app.lua F> transaction commit trigger failed
```

 Tarantool 2.11.0-rc2-0-g2ae0c94a2 Linux-x86_64-RelWithDebInfo

**Steps to reproduce**

```lua

box.cfg{}
s1 = box.schema.space.create('s1', { if_not_exists = true, is_sync = true})
s1:create_index('primary', {if_not_exists = true, unique = true, parts = {{field = 1, type = 'unsigned'}}})
box.ctl.promote()

f_commit = function(row_pairs) row_pairs() end
on_replace_f = function(old,new) box.on_commit(f_commit) end

s1:on_replace(on_replace_f)
s1:insert({1})


```

Serpentian:
Currently any transaction on synchronous space fails to complete with the ER_CURSOR_NO_TRANSACTION error, when on_rollback/on_commit triggers are set. This is caused due to the fact, that some rollback/commit triggers require in_txn fiber variable to be set but it's not done when a transaction is completed from the limbo.

Let's assign transaction to the fiber when we complete transaction from the limbo. Moreover, let's add assertions, which check whether in_txn() is set, when on_rollback/on_commit triggers are run.

Closes #8505

NO_DOC=bugfix


Serpentian:
Actually this is a draft fix, and I see reasons why you may not like it (I'm not quite sure, where to put setting this `in_txn` variable, and I'm not confident that we don't need to refactor setting it everywhere). But the issue blocks me from testing vshard with synchronous spaces and it's really significant, so it'd be great if we fix it.

The problem here, that the logic of setting and removing `in_txn` variable eludes me. It seems logical enough to set `in_txn` with `txn_begin` and remove it after `txn_complete_success/fail` but why do we always remove it before `journal_write` (or similar function), like that:

<details>
<summary>Removing <code>in_txn</code> before write</summary>

https://github.com/tarantool/tarantool/blob/d187b14247d071b44f02a2f1adac7f1daac137bc/src/box/txn.c#L1124-L1126
https://github.com/tarantool/tarantool/blob/d187b14247d071b44f02a2f1adac7f1daac137bc/src/box/txn.c#L1079-L1084

</details>

It would be great if someone enlightened me.

----

Alternative solutions, I considered:

1. Before calling `txn_limbo_ack` in `txn.c` and `relay.c`

The example of the diff can be found below. I don't like that solution, as we get ugly code (both `txn_limbo_ack` and `txn_limbo_wait_complete` must have `in_txn` set), moreover, it must be set not only in `txn.c` but also in other places (e.g. in `relay.c` which calls `tx_status_update` and there might be others, which I didn't catch)

<details>
<summary> Before calling <code>txn_limbo_ack</code></summary>

```diff
diff --git a/src/box/txn.c b/src/box/txn.c
index 8a93e64239f..959e980be11 100644
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1131,6 +1131,15 @@ txn_commit(struct txn *txn)
 	if (txn_has_flag(txn, TXN_WAIT_SYNC)) {
 		struct txn_limbo_entry *limbo_entry = txn->limbo_entry;
 		assert(limbo_entry->lsn > 0);
+		/*
+		 * Some commit/rollback triggers require for in_txn fiber
+		 * variable to be set so restore it for the time triggers
+		 * are in progress. txn_limbo_ack() may try to invoke
+		 * txn_complete_success(), which runs on_commit triggers.
+		 * txn_limbo_wait_complete - txn_complete_fail with on_rollback
+		 * triggers.
+		 */
+		fiber_set_txn(fiber(), txn);
 		/*
 		 * XXX: ACK should be done on WAL write too. But it can make
 		 * another WAL write. Can't be done until it works
@@ -1142,6 +1151,7 @@ txn_commit(struct txn *txn)
 		}
 		if (txn_limbo_wait_complete(&txn_limbo, limbo_entry) < 0)
 			goto rollback;
+		fiber_set_txn(fiber(), NULL);
 	}
 	assert(txn_has_flag(txn, TXN_IS_DONE));
 	assert(txn->signature >= 0);
@@ -1161,6 +1171,7 @@ txn_commit(struct txn *txn)
 		fiber_set_txn(fiber(), txn);
 		txn_rollback(txn);
 	} else {
+		fiber_set_txn(fiber(), NULL);
 		assert(in_txn() == NULL);
 	}
 	txn_free(txn);
```

</details>

2. Inside `txn_complete_success/fail`

It's good, that we stay inside the `txn.c` file, only in which this `in_txn` variable is set now, but it'll be really difficult to debug such code in the future.

<details>
<summary>Inside <code>txn_complete_success/fail</code></summary>

```diff
diff --git a/src/box/txn.c b/src/box/txn.c
index 8a93e6423..8e49acf14 100644
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -764,6 +764,13 @@ txn_complete_fail(struct txn *txn)
 	if (txn->engine != NULL)
 		engine_rollback(txn->engine, txn);
 	if (txn_has_flag(txn, TXN_HAS_TRIGGERS)) {
+		/*
+		 * Some rollback triggers require for in_txn fiber
+		 * variable to be set so restore it for the time triggers
+		 * are in progress.
+		 */
+		bool was_in_txn = in_txn() != NULL;
+		fiber_set_txn(fiber(), txn);
 		if (trigger_run(&txn->on_rollback, txn) != 0) {
 			diag_log();
 			panic("transaction rollback trigger failed");
@@ -772,6 +779,8 @@ txn_complete_fail(struct txn *txn)
 		trigger_destroy(&txn->on_rollback);
 		/* Commit won't happen after rollback. */
 		trigger_destroy(&txn->on_commit);
+		if (!was_in_txn)
+			fiber_set_txn(fiber(), NULL);
 	}
 	txn_free_or_wakeup(txn);
 	rmean_collect(rmean_box, IPROTO_ROLLBACK, 1);
@@ -787,6 +796,13 @@ txn_complete_success(struct txn *txn)
 	if (txn->engine != NULL)
 		engine_commit(txn->engine, txn);
 	if (txn_has_flag(txn, TXN_HAS_TRIGGERS)) {
+		/*
+		* Some commit triggers require for in_txn fiber
+		* variable to be set so restore it for the time triggers
+		* are in progress.
+		*/
+		bool was_in_txn = in_txn() != NULL;
+		fiber_set_txn(fiber(), txn);
 		/*
 		 * Commit triggers must be run in the same order they were added
 		 * so that a trigger sees the changes done by previous triggers
@@ -800,6 +816,8 @@ txn_complete_success(struct txn *txn)
 		trigger_destroy(&txn->on_commit);
 		/* Rollback won't happen after commit. */
 		trigger_destroy(&txn->on_rollback);
+		if (!was_in_txn)
+			fiber_set_txn(fiber(), NULL);
 	}
 	txn_free_or_wakeup(txn);
 	rmean_collect(rmean_box, IPROTO_COMMIT, 1);

```
</details>

----

I like the current solution the most, but I'm not sure, that we should not refactor it in general, setting `in_txn` is already seems inconsistent, this change exacerbates the situation in my opinion. What do you think?

changelogs/unreleased/gh-8505-synchro-triggers-fail.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Fixed a bug because of which transactions on synchronous spaces failed with
+  the ER_CURSOR_NO_TRANSACTION error, when on_commit/on_rollback triggers were
+  set (gh-8505).
```
andreyaksenov, line 5:
```suggestion
* Fixed a bug causing the `ER_CURSOR_NO_TRANSACTION` failure for transactions
  on synchronous spaces when the `on_commit/on_rollback` triggers are
  set (gh-8505).
```

andreyaksenov (approved):

src/box/txn.c
```diff
@@ -787,6 +788,7 @@ txn_complete_success(struct txn *txn)
 	if (txn->engine != NULL)
 		engine_commit(txn->engine, txn);
 	if (txn_has_flag(txn, TXN_HAS_TRIGGERS)) {
+		assert(in_txn() != NULL);
```
locker, line 791: Shouldn't it be `assert(in_txn() == txn)`?

> Currently any transaction on synchronous space fails to complete with
the ER_CURSOR_NO_TRANSACTION error, when on_rollback/on_commit triggers
are set. This is caused due to the fact, that some rollback/commit
triggers require in_txn fiber variable to be set but it's not done
when a transaction is completed from the limbo.

Please elaborate. What triggers? How exactly does this error occur?

AFAICS a commit/rollback trigger callback is passed the `txn` as an argument so why don't you patch the triggers that use `in_txn()` so that they use the passed `txn` instead?

src/box/txn_limbo.c
```diff
@@ -72,6 +72,21 @@ txn_limbo_is_ro(struct txn_limbo *limbo)
...
 }
 
+void
+txn_limbo_complete(struct txn *txn, bool is_success)
+{
+	/*
+	 * Some rollback/commit triggers require the in_txn fiber
+	 * variable to be set.
+	 */
+	fiber_set_txn(fiber(), txn);
```
locker, line 82: The code seem to be missing `assert(in_txn() == NULL)`.

Serpentian: Agree. It indeed looks like, that `in_txn` cannot be set before confirming everything in limbo. It's removed before writing to journal (because we cannot yield while `in_txn()` is set?) and is returned only in case we need to run triggers.

src/box/txn.c
locker: > > Please elaborate. What triggers? How exactly does this error occur?
> 
> Everything is already described in the issue. It's `lbox_txn_pairs` (and many other callbacks) function, which opens an iterator over txn's satements. It has nothing but `txn_id` on the Lua stack. The only way it can get `txn` from is fiber. I see no reason to rewrite all triggers, moreover, I'm not sure, we can push txn to stack at all.

OK. Please describe it in the commit message as well.

> > Shouldn't it be assert(in_txn() == txn)?
> 
> It should not, we must be sure, that txn is set, when triggers are run. A lot of functions in `txn.c` set `in_txn` just to be sure, that `in_txn() != NULL`.

How can `in_txn()` be different from `txn` here? If so, it looks like a bug because this would mean that `lbox_txn_pairs()` would run on a different transaction. Also, this assertion should probably be unconditional (i.e. run even if `TXN_HASH_TRIGGERS` is unset).

src/box/txn_limbo.c
sergepetrenko, line 82: While you're at it, let's please fix the effective user/session for sync space on_commit trigger as well (like it was done in #7010 for async spaces).

Otherwise I get this with a sync space and 2 nodes (`replication_synchro_quorum= 2`):
```
tarantool> print(box.session.effective_user())
guest
---
...

tarantool> box.begin() box.space.sync:replace{1} box.on_commit(function() print(box.session.effective_user()) end) box.commit()
admin
---
...

tarantool> print(box.session.effective_user())
guest
---
...


```

sergepetrenko (changes requested):
Thanks for the patch! Please see my comments below.

src/box/txn.c
Serpentian: > How can in_txn() be different from txn here?

I misread your message and thought you meant `assert(in_txn() == NULL)`. I agree, that we should add `assert(in_txn() == txn)`.

> Please describe it in the commit message as well.

Extended description in commit message

src/box/txn_limbo.c
Serpentian: Fixed

src/box/txn.c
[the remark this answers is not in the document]
Serpentian: Fixed

src/box/txn_limbo.c
Serpentian: Ok. Fixed

changelogs/unreleased/gh-8742-session-on-triggers-with-sync-tx.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+* Fixed effective session and user not propagated to `box.on_commit` and
```
andreyaksenov, line 3:
```suggestion
* Fixed a bug causing the effective session and user are not propagated to `box.on_commit` and
```

andreyaksenov (approved):

sergepetrenko (approved):
Nicely done!

test/box-luatest/gh_8505_synchro_triggers_test.lua
```diff
@@ -0,0 +1,65 @@
+local t = require('luatest')
```
Gerold103, line 1: This test file belongs to `replication-luatest` suite.

test/box-luatest/gh_8505_synchro_triggers_test.lua
```diff
@@ -0,0 +1,65 @@
...
+        box.begin()
+        box.on_commit(function(iter) iter() end)
+        box.space.test:upsert({1}, {{'=', 1, 1}})
+        box.commit()
+    end)
+end
+
+g.test_on_rollback_trigger = function(g)
+    -- Force ACK gathering to fail and cause rollback
+    g.box_cfg.replication_synchro_timeout = 1e-9
```
Gerold103, line 50: When I run the test like this `python test-run.py $(yes 8505 | head -n 30)`, I get some failures:
```
[006] Test failed! Output from reject file /tmp/t/rejects/box-luatest/gh_8505_synchro_triggers.reject:
[006] Tarantool version is 3.0.0-entrypoint-290-g1bb7177b1
[006] TAP version 13
[006] 1..2
[006] # Started on Wed Jun  7 22:35:04 2023
[006] # Starting group: gh-8505-synchro-triggers
[006] ok     1	gh-8505-synchro-triggers.test_on_commit_trigger
[006] not ok 2	gh-8505-synchro-triggers.test_on_rollback_trigger
[006] #   ...ntool/test/box-luatest/gh_8505_synchro_triggers_test.lua:59: attempt to index local 'err' (a nil value)
[006] #   stack traceback:
[006] #   	...ntool/test/box-luatest/gh_8505_synchro_triggers_test.lua:54: in function 'gh-8505-synchro-triggers.test_on_rollback_trigger'
[006] #   	...
[006] #   	[C]: in function 'xpcall'
[006] # Ran 2 tests in 2.468 seconds, 1 succeeded, 1 errored
```

Gerold103 (commented):
Thanks for the patch!

test/box-luatest/gh_8505_synchro_triggers_test.lua
Serpentian: Done

test/box-luatest/gh_8505_synchro_triggers_test.lua
Serpentian: For some reason, transactions with tiny `replication_synchro_timeout` sometimes manage to complete. This doesn't happen, when connection is explicitly broken with proxy.  I'll investigate this behavior, for this patch, I suppose, we can stay with proxies

Gerold103: Still fails:
```
[015] replication-luatest/gh_8505_synchro_triggers_t>                 [ fail ]
[015] Test failed! Output from reject file /tmp/t/rejects/replication-luatest/gh_8505_synchro_triggers.reject:
[015] Tarantool version is 3.0.0-entrypoint-306-gde0731bd3
[015] TAP version 13
[015] 1..3
[015] # Started on Wed Jun 14 22:20:55 2023
[015] # Starting group: gh-8505-synchro-triggers
[015] ok     1	gh-8505-synchro-triggers.test_on_commit_trigger
[015] ok     2	gh-8505-synchro-triggers.test_on_rollback_trigger
[015] not ok 3	gh-8505-synchro-triggers.test_session_on_commit
[015] #   Quorum collection for a synchronous transaction is timed out
[015] #   stack traceback:
[015] #   	...st/replication-luatest/gh_8505_synchro_triggers_test.lua:82: in function 'gh-8505-synchro-triggers.test_session_on_commit'
[015] #   	...
[015] #   	[C]: in function 'xpcall'
[015] # Ran 3 tests in 2.496 seconds, 2 succeeded, 1 errored
```

Gerold103 (commented):
Thanks for the fixes! The test still fails though.

test/box-luatest/gh_8505_synchro_triggers_test.lua
Serpentian: Managed to repoduce it. Should be stable from now on

Gerold103 (approved):
Thanks for the fixes!

sergepetrenko:
Cherry-picked to 2.11: a7a51aef84ff22276d594fcdc65aeddb07ef5211 .. f4d308564f16ada7c8adebd63866512a181dcc0b

PR #8737, opened 2023-06-05
luajit: bump new version

fckxorg:
sysprof: fix crash during FFUNC stream

Closes #8594

NO_DOC=LuaJIT bump
NO_TEST=LuaJIT bump

Issue #7472, opened 2022-07-26
Limit the size of the test artefacts to avoid runner degradation


igormunkin:
There is an issue with `sysprof` tests being run by `coverage` workflow. It seems like the sampling interval is too small, so SIGPROF is fired too soon and "mutator doesn't go too far", since the platform also is instrumented with coverage (i.e. has another load vector). As a result, the huge profile is dumped on the system volume, leading to runner degradation that requires manual maintenance before returning to process the queued jobs. Unfortunately, I have no environment to reproduce the issue. I also have no idea how to dig the root cause (I don't even know whether the issue relates to `sysprof` itself). All in all there are no problems with other workflows and we need to fix CI degradation at first.

As a temporary solution it was decided to relax profiler pressure by increasing profiling interval (the new default is chosen as 0.1 seconds). As a more robust solution, it was decided to fix this issue within our infrastructure. Here are the steps to be made:
* [x] Make all tests (those related to sysprof in particular) dump all results to a directory specified by environment variable (i.e. `LUAJIT_TEST_VARDIR`)
* [x] Adjust actions/environment GitHub Action to set this environment variable
* [x] Limit the partition being set via `LUAJIT_TEST_VARDIR`

When all aforementioned steps are completed, the workflow is expected to fail when `sysprof` writes a huge profile on the disk, but the runner's system volume is not affected, hence, it stays alive, doesn't need manual maintenance and ready to process the queued jobs.

PR #8738, opened 2023-06-06
ci: use workaround for LuaJIT profiling tests in all workflows

ylobankov:
After this patch, the workaround for LuaJIT profiling tests to avoid runner's shutdown due to no space left on the disk is going to be used in all relevant workflows. Previously, we applied it only for the coverage test, but we noticed that other jobs may be affected as well after tarantool/tarantool#8737 is merged.

Follows up tarantool/tarantool#7472

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -86,10 +86,10 @@ runs:
         git clean -xffd
         git submodule foreach --recursive 'git clean -xffd'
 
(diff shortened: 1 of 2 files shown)


igormunkin (approved):
@ylobankov, I can say nothing regarding the second patch in the PR, but the first one related to LuaJIT LGTM. Thanks!

ylobankov:
Cherry picked to release/2.11 and release/2.10.

PR #8741, opened 2023-06-06
Introduce API to read a state value without subscribing (watch once)

### LINKED ISSUE #6493: API to read event data without making a subscription
[Design document](https://www.notion.so/tarantool/Watch-once-0143eee47a514fdd91a3232a156d81ef)

The ticket #6257 defines subscriptions. The other ticket #6260 defines built-in events available out of the box. One of the features of the subs is that they work without any grants. At least the built-in ones.

It might be useful to have an API to get data by a subscription key without making an actual subscription. It could be used by clients who are not able or don't want to process async events. They might want to poll the data. With the help of such API they wouldn't need to implement stored functions to get stuff like election state, some cfg options, and whatever will be exposed along.

It is possible to emulate it by sending subscribe + unsubscribe requests right after each other, but that means more network data, and more load on the server to create and drop the subscription even though it wasn't really needed.

locker:
This PR introduces the `box.watch_once()` function, `IPROTO_WATCH_ONCE` request type, and `conn:watch_once()` net.box connection method to read the value currently associated with a notification key locally and remotely. See the commit log for more information.

Besides that, it also:
- converts the watcher tests from difftest to luatest for easier extensibility;
- refactors the `netbox_method` enum so that it's exported to the net.box Lua code automatically.

Closes #6493


andreyaksenov (approved):

src/box/watcher.h
```diff
@@ -259,6 +259,25 @@ box_broadcast(const char *key, size_t key_len,
...
+ * @param[out] end End of the key data
+ * @retval Key data
+ *
+ * The function never fails.
+ *
+ * If there's no data attached to the given notification key (box_broadcast()
+ * has never been called for this key), the function returns NULL.
+ *
+ * Note that the data returned by this function may be updated by a concurrent
+ * call to box_broadcast() so the caller must copy it if it intends to yield.
```
Gerold103, line 276: JFYI: it makes the function not safe against Lua GC. While `luamp_decode()` is working in `lbox_watch_once()`, it might invoke `__gc` handlers in Lua which could do `box.broadcast()` for the same key. I am fine with keeping it as is though. Anyway our protection agains Lua GC is full of holes.

Gerold103 (approved):
Thanks for the patch!

src/box/watcher.h
locker: You're right, I missed that. This issue should be pretty easy to fix though: we just need to copy the data to an area allocated with `malloc` before calling `luamp_decode`. However, I don't see a way to reproduce this issue i.e. the fix would be untested. So I prefer to leave it as is and fix it only if someone actually complains. Maybe, we should just disable Lua GC while in Lua C and call it only on return to Lua. This would fix most of these potential issues.

Gerold103: The test would be simple - you just need to repeat what is done in `test/app-tap/gh-5632-6050-6259-gc-buf-reuse.test.lua`. I am fine with leaving it as is too.

PR #8749, opened 2023-06-08
ci: fix test-sdk job in submodule_update.yml

ylobankov:
Fix the `test-sdk` job in the `submodule_update.yml` workflow according to the new name policy for supported branches (2.10 -> release/2.10, 2.11 -> release/2.11).


locker (approved):
Please prepare the same patch for Tarantool EE - this workflow fails there as well.

Issue #6470, opened 2021-09-27
ddl: rollback of space:drop() does not revert state of Lua object


Korablev77:
Note that it has nothing to do with MVCC, bug is reproduced in regular memtx mode:
```
s = box.schema.space.create('test')
_ = s:create_index('primary')
box.begin() s:drop() box.rollback()
s
...
  index: []
  name: test
  id: 512
...

s:select()
---
- error: 'No index #0 is defined in space ''test'''
...

```
Meanwhile `box.space.test:select()` is fine:

```
tarantool> box.space.test:select()
---
- []
...
```

locker:
Here's what happens:
1. `box.begin()`.
2. `s:drop()` drops the space index, which triggers update of the Lua state - `box.space.test`. Since `s` is a reference to the `box.space.test` table, the change is reflected to it as well. This is okay, because in the transaction the index doesn't exist anymore and should be inaccessible.
3. `s:drop()` then drops the space, which deletes a reference to `box.space.test` from `box.space`. Again, this is okay, because the space doesn't exist in the transaction anymore.
4. `box.rollback()` restores the Lua state. Since `box.space.test` doesn't exist (was deleted at the previous step), it will create a new table in `box.space`, which will be different from `s` and hence `s` stops working.

We could cook up a quick fix to this issue by temporarily storing the deleted `box.space.test` table somewhere in an internal table until transaction ends and restoring it on rollback. However, the proper way to fix this issue is to introduce multi-version schema (#600), which would make it possible to present different views of the data dictionary in different transactions. Multi-version schema is orthogonal to the quick fix. Since the issue is non-critical and none of our customers ever complained about this behavior, let's close this ticket in favor of #600.

PR #6228, opened 2021-07-19
txm: disallow yields after finishing alter operation

Korablev77:
Subj.

Also see https://github.com/tarantool/tarantool/pull/6202#issuecomment-876524195

Closes #5998
Closes #6140

Issue #600, opened 2014-10-22
create/alter space/index in transaction


unera:
I want to create space and its indexes in one transaction (up/down)

for example my project is grown, so I want to upgrade database.
I write two scripts: `up.lua` and `down.lua`.

If `up.lua` raises an error it have to rollback all its works.
`down.lua` have to have the same behaviour

``` console
tarantool> box.begin()

---
...

tarantool> box.schema.create_space('test')

---
- error: Space _space does not support multi-statement transactions
...
```

farwayer:
One more problem is `before_replace`/`on_replace` executes in transaction context. So It is not possible to create space inside handler :disappointed:

alyapunov:
^^ closed by misclick and reopened.

CuriousGeorgiy:
@alyapunov @mkostoevr

The [_Online schema evolution is (almost) free for snapshot databases_](https://www.vldb.org/pvldb/vol16/p140-hu.pdf) paper by Tianzheng Wang et al. may be a good reference.

PR #8751, opened 2023-06-08
box: make box.schema functions transactional when possible

### LINKED ISSUE #4348: Make all box.schema functions transactional
Use `box.begin()` and `box.commit()` in all DDL functions exported to Lua, like `box.schema.space.create()` or `box.schema.user.grant()`.

Follow-up #4083

mkostoevr:
Wraps DDL functions into `begin`/`commit` block if no transaction is active.

Now disable of functional index causes index rebuild which does not allow the `box.schema.space.drop` function to be atomic. Since effectively the disabled functional index rebuild is a no-op, the "rebuild" step was omitted and effective changes applied instead.

Closes #4348

NO_DOC=bugfix


changelogs/unreleased/gh-4348-transactional-ddl.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* All the DDL functions from `box.schema` module are wrapped into transaction
+  to avoid database inconsistency on failed operation (gh-4348).
```
xuniq, line 4:
```suggestion
* All DDL functions from the `box.schema` module are now wrapped into a transaction
  to avoid database inconsistency on failed operations (gh-4348).
```

xuniq (approved):

ochaplashkin (approved):
LGTM :rocket:

src/box/lua/schema.lua
```diff
@@ -3677,30 +3677,101 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
     local uid = role_resolve(user_name)
     if uid == nil then
         box.error(box.error.NO_SUCH_ROLE, user_name)
     end
     role_check_grant_revoke_of_sys_priv(...)
     return revoke(uid, user_name, ...)
 end
-box.schema.role.info = function(role_name)
+
+local function box_schema_role_info(role_name)
```
alyapunov, line 3698: It should be `box_schema_role_info_do` to make it clear that it is a helper function now. The same is for all functions above.

src/box/lua/schema.lua
```diff
@@ -3677,30 +3677,101 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
     if rid == nil then
         box.error(box.error.NO_SUCH_ROLE, role_name)
     end
     return info(rid)
 end
 
+-- Wrap a function into transaction if none is active
+local function transactional_wrapper(func)
+    return function(...)
+        local atomic
```
alyapunov, line 3709: We usually call `is_something` for flags, there should be `is_atomic`.

src/box/lua/schema.lua
```diff
@@ -3677,30 +3677,101 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
+
+        if not box.is_in_txn() then
+            atomic = true
+        else
+            atomic = false
+        end
+
+        if atomic then box.begin() end
```
alyapunov, line 3717: Actually there's `box.atomic` function that does exactly what you need, and I believe it is perfectly optimized.

alyapunov (changes requested):

src/box/lua/schema.lua
mkostoevr: Reused it.

src/box/lua/schema.lua
mkostoevr: The change has been dropped from the commit.

src/box/lua/schema.lua
mkostoevr: Renamed the non-transactional functions.

src/box/lua/schema.lua
```diff
@@ -3593,30 +3593,84 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
+        -- No reason to start a transaction of one is active already.
+        if box.is_in_txn() then
+            return func(...)
+        else
+            return box.atomic(func, ...)
+        end
+    end
+end
+
+box.schema.space.create = atomic_wrapper(box_schema_space_create_do)
```
locker, line 3634: Having this list of functions distanced from the implementation makes the code difficult to follow. Why not drop the temporary functions and use the atomic wrapper right at function definition?

```diff
diff --git a/src/box/lua/schema.lua b/src/box/lua/schema.lua
index 87f931ca4353..8fbec7d76765 100644
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -717,7 +717,7 @@ local function check_space_type(space_type)
 end
 
 box.schema.space = {}
-box.schema.space.create = function(name, options)
+box.schema.space.create = atomic_wrapper(function(name, options)
     check_param(name, 'name', 'string')
     local options_template = {
         if_not_exists = 'boolean',
@@ -791,7 +791,7 @@ box.schema.space.create = function(name, options)
 
     feedback_save_event('create_space')
     return box.space[id], "created"
-end
+end)
 
 -- space format - the metadata about space fields
 function box.schema.space.format(id, format)
```

src/box/lua/schema.lua
```diff
@@ -3593,30 +3593,84 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
+box.schema.space.format = atomic_wrapper(box_schema_space_format_do)
+box.schema.space.upgrade = box_schema_space_upgrade_do
+box.schema.space.drop = atomic_wrapper(box_schema_space_drop_do)
+box.schema.space.rename = atomic_wrapper(box_schema_space_rename_do)
```
locker, line 3638: `rename` and `format` and probably some other functions don't need the atomic wrapper because they do exactly one statement. Why did you decide to wrap them?

src/box/lua/schema.lua
```diff
@@ -3593,30 +3593,84 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
+box.schema.space.alter = atomic_wrapper(box_schema_space_alter_do)
+
+box.schema.index.create = atomic_wrapper(box_schema_index_create_do)
+box.schema.index.drop = atomic_wrapper(box_schema_index_drop_do)
+box.schema.index.rename = atomic_wrapper(box_schema_index_rename_do)
+box.schema.index.alter = box_schema_index_alter_do
```
locker, line 3644: Why is this function not wrapped in atomic? AFAIU it may perform a few statements.

src/box/lua/schema.lua
```diff
@@ -3593,30 +3593,84 @@ local function role_check_grant_revoke_of_sys_priv(priv)
...
+box.schema.user.disable = atomic_wrapper(box_schema_user_disable_do)
+box.schema.user.drop = atomic_wrapper(box_schema_user_drop_do)
+box.schema.user.info = atomic_wrapper(box_schema_user_info_do)
+
+box.schema.role.exists = atomic_wrapper(box_schema_role_exists_do)
+box.schema.role.create = atomic_wrapper(box_schema_role_create_do)
+box.schema.role.drop = atomic_wrapper(box_schema_role_drop_do)
+box.schema.role.grant = atomic_wrapper(box_schema_role_grant_do)
+box.schema.role.revoke = atomic_wrapper(box_schema_role_revoke_do)
+box.schema.role.info = atomic_wrapper(box_schema_role_info_do)
```
locker, line 3670: `info` and `exists` methods are read-only so they don't have any side-effects hence they don't need the atomic wrapper, do they?

test/box-luatest/gh_4348_transactional_ddl_test.lua
```diff
@@ -0,0 +1,281 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group('gh-4348-transactional-ddl-test', {
+    {mvcc = false, engine = 'memtx'},
+    {mvcc = false, engine = 'vinyl'},
+    {mvcc = true, engine = 'memtx'},
+    {mvcc = true, engine = 'vinyl'},
```
locker, line 8: Nit: There's `t.helpers.matrix` helper that can simplify the test group definition a little bit:

```Lua
t.helpers.matrix({mvcc = {true, false}, engine = {'vinyl', 'memtx'}})
```

Anyway, I'm not sure that we actually need to run the tests for Vinyl - the code you patch is engine-independent.

test/box-luatest/gh_4348_transactional_ddl_test.lua
```diff
@@ -0,0 +1,281 @@
...
+        for i = 0, 1001  do -- More than 1k to ensure yielding.
+            fs1:insert({i + 1})
+            fs2:insert({i + 2})
+            fs3:insert({i + 3})
+            s:insert({i, i + 1, i + 2, i + 3})
+        end
+
+        ------------------------------------------------------------------------
+        -- box.schema.space.format ---------------------------------------------
+        ------------------------------------------------------------------------
```
locker, line 112: It'd be better to split the test into several test cases instead of using comments.

test/box-luatest/gh_4348_transactional_ddl_test.lua
```diff
@@ -0,0 +1,281 @@
...
+
+        -- Non-yielding format change.
+        value = {{name = 'id', type = 'scalar'}}
+        box.schema.space.format(s.id, value)
+        t.assert_equals(box.schema.space.format(s.id), value)
```
locker, line 117: The successful work of all the DDL functions is already checked by other tests. I don't see much point in adding more tests for it. What we do need to test in the scope of this issue is that failing in the middle of a composite DDL operation (such as `space.drop`) doesn't leave any traces.

test/box/errinj.result
```diff
@@ -415,6 +415,10 @@ s_withdata:drop()
 ---
 - error: Failed to write to disk
 ...
+-- FIXME? The space is recreated, so the s_withdata object is not valid anymore.
+s_withdata = box.space.withdata
```
locker, line 419: I think that technically this should be possible to fix some day (not now). Let's file a ticket for this issue and reference it here: `FIXME(gh-XXXX): The space is recreated...`.

locker (changes requested):

src/box/lua/schema.lua
mkostoevr: This makes the code a bit less accurate to me, but that's a matter of taste.

src/box/lua/schema.lua
mkostoevr: Just followed the order to "Make [all](https://github.com/tarantool/tarantool/issues/4348) box.schema functions transactional". It's not required for some of them.

src/box/lua/schema.lua
mkostoevr: See the [answer](https://github.com/tarantool/tarantool/pull/8751#discussion_r1317401750) to the previous comment.

src/box/lua/schema.lua
locker: With a separate code block wrapping all DDL functions in `atomic_wrapper` one by one, one can't understand whether a particular function is atomic by just looking at its definition - they'd have to browse to a different place in the code, which isn't really convenient. The `atomic_wrapper` looks like a function decorator to me so better use it at a function definition IMO.

It'd be a different story if we wrapped *all* functions in this wrapper *automatically* but AFAIU we can't do that because not all functions need the wrapping.

src/box/lua/schema.lua
locker: The issue description says:

> Use `box.begin()` and `box.commit()` in all DDL functions exported to Lua, like `box.schema.space.create()` or `box.schema.user.grant()`.

`exists` and `info` aren't DDL functions.

test/box/errinj.result
mkostoevr: Done. The ticket is there already: #6470.

src/box/lua/schema.lua
mkostoevr: It's not possible to make it atomic for now. The function can perform index rebuild/format check, which [are](https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9#diff-4783b9090bf7018ce3825d26935a097236f7caf5d87dc2fb0e7fa4e1d3c3f86bR1041) [forced](https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9#diff-4783b9090bf7018ce3825d26935a097236f7caf5d87dc2fb0e7fa4e1d3c3f86bR893) to be the first statements in a transaction in order to [prevent](https://github.com/tarantool/tarantool/pull/6228#issuecomment-891620195) other transactions from seeing partial schema changes.

At the same time it can [delete](https://github.com/tarantool/tarantool/blob/52884400993d1bca1c490e3a739f5528ce06ed3b/src/box/lua/schema.lua#L1366) an item from the `_space_sequence` table, which will be the first transaction statement. So the index rebuild/format check will fail. And we can't drop the sequence after altering the index because the alter can change the type of the field we have the existing sequence for, so the attached to the space sequence will become [invalid](https://github.com/tarantool/tarantool/blob/52884400993d1bca1c490e3a739f5528ce06ed3b/src/box/alter.cc#L164) for its key def.

So the function in guaranteed to fail in case of specific input if it's wrapped into a transaction. The only way to fix this I see is to implement the [transactional DDL](https://github.com/tarantool/tarantool/issues/600).

locker: Why do we have to delete an item from the `_space_sequence` table *before* updating an item in the `_index` space? Why can't we do that after we're done building the index?

src/box/lua/schema.lua
mkostoevr: Dropped atomicity of simple single statement and read-only functions.

test/box/errinj.result
mkostoevr: Created and linked to the #9120.

test/box-luatest/gh_4348_transactional_ddl_test.lua
mkostoevr: Just wanted to cover as much use cases of the transactioned DDL functions as possible, because the changes are very invasive and could break something (like some DDLs may be impossible to be performed in a transaction in specific conditions). So I tried a bunch of combinations that could lead to something unexpected.

Some untested cases had been covered according to the [coverity report](https://github.com/tarantool/tarantool/pull/8751#issuecomment-1582514568), but I doubt the suite is enough for this feature.

src/box/lua/schema.lua
mkostoevr: Fair enough, fixed.

It turns out there's only 9 functions to be wrapped, other are either read-only, singlewritestament or non-DDL.

test/box-luatest/gh_4348_transactional_ddl_test.lua
mkostoevr: Splitted.

test/box-luatest/gh_4348_transactional_ddl_test.lua
mkostoevr: Checked the atomicity of each transactioned function.

test/box-luatest/gh_4348_transactional_ddl_test.lua
mkostoevr: Yes, it seems like, made the code engine-independent.

src/box/lua/schema.lua
mkostoevr: Dropped the sequence check from the `_index:on_replace` trigger and moved the sequence unassignment after the index alter.

This has a caveat though: uncareful manual altering the index can silently put the space into an inconsistent condition: the next insert into the space can either fail or silently succeed whereas it's not supposed operation for the space. Example:

```lua
box.cfg{}
pk = s:create_index('pk', {parts = {{1, 'unsigned'}}, sequence = true})

-- Manual altering the index without dropping the sequence.
box.space._index:update({s.id, pk.id}, {{'=', 6, {{0, 'string'}}}})

-- This will fail: passed integer (from sequence) in string field.
s:insert({box.NULL})

-- Manual altering the index without dropping the sequence again.
box.space._index:update({s.id, pk.id}, {{'=', 6, {{0, 'scalar'}}}})

-- This will succeed: passed integer (from sequence) in scalar field.
-- Though it shouldn't have been done at all: one should not pass
-- NULL in the field which is indexed by the primary key.
s:insert({box.NULL})
```

src/box/lua/schema.lua
```diff
@@ -1587,21 +1571,19 @@ box.schema.index.create = function(space_id, name, options)
     if index_opts.func ~= nil and type(index_opts.func) == 'string' then
         index_opts.func = func_id_by_name(index_opts.func)
     end
-    local sequence_proxy = space_sequence_alter_prepare(format, parts, options,
-                                                        space_id, iid,
-                                                        space.name, name)
     _index:insert{space_id, iid, name, options.type, index_opts, parts}
-    space_sequence_alter_commit(sequence_proxy)
+    space_sequence_alter(format, parts, options, space_id,
+                         iid, space.name, name)
```
locker, line 1576: Let's do all the sequence-related checks *before* altering the index definition. Rationale: index rebuild may take quite a lot of time; it'd be nice to fail early if sequence parameters are invalid.

locker (changes requested):

src/box/lua/schema.lua
mkostoevr: That was a serious overlook, fixed.

locker (approved):

locker:
expirationd test fails. PTAL.

PR #8758, opened 2023-06-09
httpc: fix content-length with a stream

### LINKED ISSUE #8744: The HTTP client shouldn't make the 'chunked' request when the 'Content-Length' header is set
* OS: MacOS
* Architecture: arm
* Version: 2.11

The code below with the `chunked` option set to `true` should make the regular POST request (without `Transfer-Encoding`) if the `Content-Length` header is set:
```
local http_client = require('http.client').new()

local data = 'any data'
local io = http_client:post('http://127.0.0.1/anything', nil, {chunked = true, headers = {
        ['Content-Length'] = tostring(string.len(data))
    }})
io:write(data:sub(1, 3))
io:write(data:sub(4))
io:finish()
```

Instead, this request sets both `Content-Length` and `Transfer-Encoding` headers. See the attached screenshot.

image

oleg-jukovec:
The description of CURLOPT_UPLOAD [1] is confusing:

    If you use PUT to an HTTP 1.1 server, you can upload data without
    knowing the size before starting the transfer if you use chunked
    encoding. You enable this by adding a header like
    "Transfer-Encoding: chunked" with CURLOPT_HTTPHEADER.

In fact, libcurl adds this header itself. Actually we need to avoid
adding this header by libcurl with CURLOPT_INFILESIZE [2].

The CURLOPT_UPLOAD documentation has been updated [3].

1. https://github.com/curl/curl/blob/555bacd6d522bcca497573765056354f4d5d7b33/docs/libcurl/opts/CURLOPT_UPLOAD.3
2. https://curl.se/libcurl/c/httpput.html
3. https://github.com/curl/curl/pull/11300

Closes https://github.com/tarantool/tarantool/issues/8744

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
+++ b/changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
@@ -1,0 +1,4 @@
+## bugfix/lua/http client
+
+* Fixed the `Transfer-Encoding: chunked` setting being enabled even if
+  the `Content-Length` header exists for stream requests (gh-8744).
--- a/src/httpc.h
+++ b/src/httpc.h
@@ -133,11 +133,11 @@ struct httpc_request {
 	 */
 	bool set_keep_alive_header;
 	/**
-	 * True when Transfer-Encoding: chunked must be
-	 * set before execution automatically if chunked
-	 * io enabled.
+	 * It's the number of bytes of data in the body of the
+	 * request. The value is equal to "Content-Length"
+	 * header or -1 otherwise.
 	 */
-	bool set_chunked_header;
+	long content_length;
 	/** True when chunked io is enabled. */
 	bool io;
 	/**
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -164,6 +164,7 @@ httpc_request_new(struct httpc_env *env, const char *method,
 	req->env = env;
 	req->set_connection_header = true;
 	req->set_keep_alive_header = true;
+	req->content_length = -1;
 
 	region_create(&req->resp_headers, &cord()->slabc);
 	region_create(&req->recv, &cord()->slabc);
@@ -255,10 +256,19 @@ httpc_set_header(struct httpc_request *req, const char *fmt, ...)
 		 strlen(HTTP_CONNECTION_HEADER)) == 0)
 		req->set_connection_header = false;
 	else if (strncasecmp(header, HTTP_CONTENT_LENGTH_HEADER,
-			     strlen(HTTP_CONTENT_LENGTH_HEADER)) == 0)
-		req->set_chunked_header = false;
-	else if (strncasecmp(header, HTTP_KEEP_ALIVE_HEADER,
-			     strlen(HTTP_KEEP_ALIVE_HEADER)) == 0)
+			     strlen(HTTP_CONTENT_LENGTH_HEADER)) == 0) {
+		const char *len = header + strlen(HTTP_CONTENT_LENGTH_HEADER);
+		char *end = NULL;
+		long length = strtol(len, &end, 10);
+		if (*end != '\0' || length < 0) {
+			diag_set(IllegalParams,
+				 "Content-Length header value "
+				 "must be a non-negative integer");
+			return -1;
+		}
+		req->content_length = length;
+	} else if (strncasecmp(header, HTTP_KEEP_ALIVE_HEADER,
+			       strlen(HTTP_KEEP_ALIVE_HEADER)) == 0)
 		req->set_keep_alive_header = false;
 
 	struct curl_slist *l = curl_slist_append(req->headers, header);
@@ -287,7 +297,6 @@ httpc_set_body(struct httpc_request *req, const char *body, size_t size)
 
 	if (httpc_set_header(req, "Content-Length: %zu", size) != 0)
 		return -1;
-	req->set_chunked_header = false;
 
 	return 0;
 }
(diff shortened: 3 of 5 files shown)


changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/lua/http client
+
+* Fixed "Transfer-Encoding: chunked" setting even if "Content-Length" header
+  exists for stream requests.
```
Totktonada, line 4: Please, add an issue number (check other changelog entries for the format).

src/httpc.h
```diff
@@ -133,11 +133,11 @@ struct httpc_request {
 	 */
 	bool set_keep_alive_header;
 	/**
-	 * True when Transfer-Encoding: chunked must be
-	 * set before execution automatically if chunked
-	 * io enabled.
+	 * It's the number of bytes of data in the body of the
+	 * request. The value is equal to "Content-Length"
+	 * header or 0 otherwise.
 	 */
```
Totktonada, line 139: To think: a zero length request is a valid request. Should we always use the chunked encoding for it?

Totktonada: Moreover: we use a signed type here (and `curl_off_t` is signed as well), so there is no problem to use -1 as the special value.

src/httpc.h
```diff
@@ -133,11 +133,11 @@ struct httpc_request {
...
-	bool set_chunked_header;
+	long content_length;
```
Totktonada, line 140: To thinkL `long` is 8 bytes on amd64 and aarch64, so everything should be OK there on the platforms we're support (at least for Linux environments). However, it looks a bit inconsistent to use `long` with `CURLOPT_INFILESIZE_LARGE`, which AFAIU is introduced after `CURLOPT_INFILESIZE` just to accept a type that is possibly wider than `long`.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/lua/http client
```
Totktonada, line 1: How about reporting the problem against the libcurl's documentation for `CURLOPT_UPLOAD`?

src/httpc.h
oleg-jukovec: You are right:

https://www.rfc-editor.org/rfc/rfc9110#field.content-length

src/httpc.h
oleg-jukovec: https://github.com/georgi/curl/blob/master/lib/README.curl_off_t

curl_off_t is a data type provided by the external libcurl include headers. It is the type meant to be used for the curl_easy_setopt() options that end with LARGE. The type is 64bit large on most modern platforms.

oleg-jukovec: https://github.com/curl/curl/blob/3f784980cbbea61881a30e61c86f7c32caac361a/lib/setopt.c#L1371-L1390

src/httpc.h
oleg-jukovec: Updated:

1. `-1` by default for `req->content_length`.
2. Allow non-negative value for the `Content-Length` header.

src/httpc.h
oleg-jukovec: Thank you, updated `long` + `CURLOPT_INFILESIZE`.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
oleg-jukovec: Done:

`* Fixed "Transfer-Encoding: chunked" setting even if "Content-Length" header
  exists for stream requests (gh-8744).`

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
oleg-jukovec: Merged: https://github.com/curl/curl/pull/11300

src/httpc.h
Totktonada: OK, thanks! This change obsoletes current commit/PR description a bit (it mentions `CURLOPT_INFILESIZE_LARGE`).

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
Totktonada: Thanks! It obsoletes the commit/PR description a bit: I think we should highlight that the quoted option description is updated now. Otherwise it may be confusing.

src/httpc.h
oleg-jukovec: I'm sorry, updated.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
oleg-jukovec: Done.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
```diff
@@ -0,0 +1,4 @@
...
+
+* Fixed "Transfer-Encoding: chunked" setting even if "Content-Length" header
+  exists for stream requests (gh-8744).
```
andreyaksenov, line 4:
```suggestion
* Fixed the `Transfer-Encoding: chunked` setting even if the `Content-Length` header
  exists for stream requests (gh-8744).
```

andreyaksenov (approved):

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
oleg-jukovec: Thank you, updated.

Totktonada (approved):
I see no problems visually neither in the code, nor in the test cases.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/lua/http client
+
+* Fixed the `Transfer-Encoding: chunked` setting even if the `Content-Length`
+  header exists for stream requests (gh-8744).
```
Lord-KA, line 4: Nit: I think we should specify what is being fixed about Transfer-Encoding setting (I got quite confused even after reading the issue and curl doc). Maybe like this:
> Fixed the `Transfer-Encoding: chunked` setting being enabled even if the `Content-Length` header exists for stream requests (gh-8744).

Lord-KA (approved):
Thanks for the patch! See a minor issue below, otherwise LGTM.

changelogs/unreleased/gh-8744-fix-chunked-with-content-length.md
oleg-jukovec: Thank you, updated.

Totktonada:
Pushed to `master` in `3.0.0-entrypoint-375-ge2e5ef4b4` (will go to 3.0.0). Cherry-picked to `release/2.11` in `2.11.1-entrypoint-145-gcf57867b3` (will go to 2.11.1).

PR #8759, opened 2023-06-10
test: fix null pointer dereference in `coio::test_getaddrinfo`

godzie44:
in test/unit/coio.cc if rc at coio.cc:95 equals to 0 (no errors occured) then test crashes with SIGSEGV at this line:
```c
const char *errmsg = diag_get()->last->errmsg;
```

I ran into this by accident while running tests on a machine with an unusual network setup. I understand that this is not the expected branch code execution, but i think it good to fix it anyway.


alyapunov (approved):

alyapunov:
Thanks for the patch!

ochaplashkin (approved):

Issue #8657, opened 2023-05-15
Use universal triggers registry for tarantool triggers


drewdzzz:
Since we will have a new trigger mechanism, let's use it for tarantool triggers. This mechanism will allow to set triggers on entities from `box` (such as `space`) before space is created, or even before `box.cfg{}`. However, it will produce a breaking change - after a space will be dropped, its triggers won't. Another breaking change is that all the space triggers (including `before_replace` trigger) will be disabled on recovery.

We will rewrite all the tarantool triggers instead of transactional ones (`box.on_commit/rollback`), triggers from `swim` and `net.box` - they are not compatible with the ideology of the trigger registry and we will deal with them later. We will use address of a handler as the name of a trigger for the registry.

PR #8768, opened 2023-06-14
core/lua: introduce func_adapter for Lua

drewdzzz:
We are going to introduce universal triggers registry and use it for
tarantool triggers (e.g. space.on_replace). However, when trigger is set
to the registry, we do not know which arguments will be passed there
from C. That is why current implementation of triggers in tarantool is not
suitable for use in universal registry.

We decided to introduce func_adapter interface which allows to call
functions from any programming language. Now we support only one
language, which is Lua, so for the most part this interface is needed
to invert dependencies.

The patch introduces func_adapter interface and its implementation for
Lua functions.

Part of https://github.com/tarantool/tarantool/issues/8657

### DIFF
--- a/src/box/lua/func_adapter.c
+++ b/src/box/lua/func_adapter.c
@@ -1,0 +1,257 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include "box/lua/func_adapter.h"
+#include "box/lua/tuple.h"
+#include "box/tuple.h"
+#include "core/func_adapter.h"
+#include "lua/utils.h"
+
+/**
+ * Context for func_adapter_lua.
+ */
+struct func_adapter_lua_ctx {
+	/**
+	 * Lua state which stores arguments and is used to call Lua function.
+	 */
+	struct lua_State *L;
+	/**
+	 * A reference to Lua state in Lua registry.
+	 */
+	int coro_ref;
+	/**
+	 * Saved top of L.
+	 */
+	int top_svp;
+	/**
+	 * Index of an element that will be popped next. It is used not to
+	 * remove elements from the middle of the Lua stack.
+	 */
+	int idx;
+};
+
+static_assert(sizeof(struct func_adapter_lua_ctx) <= sizeof(struct func_adapter_ctx),
+	      "sizeof(func_adapter_lua_ctx) must be <= "
+	      "sizeof(func_adapter_ctx)");
+
+/**
+ * Specialization of func_adapter for Lua functions and other callable objects.
+ */
+struct func_adapter_lua {
+	/**
+	 * Virtual table.
+	 */
+	const struct func_adapter_vtab *vtab;
+	/**
+	 * Reference to the function in Lua registry.
+	 */
+	int func_ref;
+};
+
+/**
+ * Creates (or gets cached in fiber) Lua state and saves its top.
+ */
+static void
+func_adapter_lua_begin(struct func_adapter *base,
+		       struct func_adapter_ctx *base_ctx)
+{
+	struct func_adapter_lua *func = (struct func_adapter_lua *)base;
+	struct func_adapter_lua_ctx *ctx =
+		(struct func_adapter_lua_ctx *)base_ctx;
+	if (fiber()->storage.lua.stack == NULL) {
+		ctx->L = luaT_newthread(tarantool_L);
+		if (ctx->L == NULL)
+			panic("Cannot create Lua thread");
+		ctx->coro_ref = luaL_ref(tarantool_L, LUA_REGISTRYINDEX);
+	} else {
+		ctx->L = fiber()->storage.lua.stack;
+		ctx->coro_ref = LUA_REFNIL;
+	}
+	ctx->idx = 0;
+	ctx->top_svp = lua_gettop(ctx->L);
+	lua_rawgeti(ctx->L, LUA_REGISTRYINDEX, func->func_ref);
+}
+
+/**
+ * Releases Lua state.
+ */
+static void
+func_adapter_lua_end(struct func_adapter_ctx *base)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	lua_settop(ctx->L, ctx->top_svp);
+	luaL_unref(tarantool_L, LUA_REGISTRYINDEX, ctx->coro_ref);
+	ctx->coro_ref = LUA_REFNIL;
+	ctx->L = NULL;
+}
+
+/**
+ * Call the function with arguments that were passed before.
+ */
+static int
+func_adapter_lua_call(struct func_adapter_ctx *base_ctx)
+{
+	struct func_adapter_lua_ctx *ctx =
+		(struct func_adapter_lua_ctx *)base_ctx;
+	int nargs = lua_gettop(ctx->L) - ctx->top_svp - 1;
+	if (luaT_call(ctx->L, nargs, LUA_MULTRET) != 0)
+		return -1;
+	ctx->idx = ctx->top_svp + 1;
+	return 0;
+}
+
+static void
+func_adapter_lua_push_double(struct func_adapter_ctx *base, double val)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	lua_pushnumber(ctx->L, val);
+}
+
+static void
+func_adapter_lua_push_str(struct func_adapter_ctx *base,
+			  const char *str, size_t len)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	lua_pushlstring(ctx->L, str, len);
+}
+
+static void
+func_adapter_lua_push_tuple(struct func_adapter_ctx *base, struct tuple *tuple)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	luaT_pushtuple(ctx->L, tuple);
+}
+
+static void
+func_adapter_lua_push_null(struct func_adapter_ctx *base)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	lua_pushnil(ctx->L);
+}
+
+/**
+ * Checks if the next value is a Lua number. Cdata numeric types and decimal are
+ * not supported.
+ */
+static bool
+func_adapter_lua_is_double(struct func_adapter_ctx *base)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	return lua_gettop(ctx->L) >= ctx->idx &&
+	       lua_type(ctx->L, ctx->idx) == LUA_TNUMBER;
+}
+
+static void
+func_adapter_lua_pop_double(struct func_adapter_ctx *base, double *out)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	*out = lua_tonumber(ctx->L, ctx->idx++);
+}
+
+static bool
+func_adapter_lua_is_str(struct func_adapter_ctx *base)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	return lua_gettop(ctx->L) >= ctx->idx &&
+	       lua_type(ctx->L, ctx->idx) == LUA_TSTRING;
+}
+
+static void
+func_adapter_lua_pop_str(struct func_adapter_ctx *base, const char **str,
+			 size_t *len)
+{
+	struct func_adapter_lua_ctx *ctx = (struct func_adapter_lua_ctx *)base;
+	*str = lua_tolstring(ctx->L, ctx->idx++, len);
+}
+
+/**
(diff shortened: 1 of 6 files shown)


src/box/lua/func_adapter.h
```diff
@@ -0,0 +1,26 @@
...
+
+struct lua_State;
+struct lua_func_adapter;
+
+/**
+ * Creates lua_func_adapter, func_ref is a reference to the function in Lua
+ * registry. Never returns NULL.
+ */
+struct func_adapter *
+lua_func_adapter_create(int func_ref);
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/func_adapter.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+#include "core/func_adapter.h"
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+struct lua_State;
+struct lua_func_adapter;
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,343 @@
...
+	if (fiber()->storage.lua.stack == NULL) {
+		ctx->L = luaT_newthread(tarantool_L);
+		if (ctx->L == NULL)
+			panic("Cannot create Lua thread");
+		ctx->coro_ref = luaL_ref(tarantool_L, LUA_REGISTRYINDEX);
+	} else {
+		ctx->L = fiber()->storage.lua.stack;
+		ctx->coro_ref = LUA_REFNIL;
+	}
+	ctx->top_svp = lua_gettop(ctx->L);
```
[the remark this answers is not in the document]
drewdzzz: Initialize it to 0 now.

src/lib/core/func_adapter.h
```diff
@@ -0,0 +1,309 @@
...
+static inline void
+func_adapter_begin(struct func_adapter *func, struct func_adapter_ctx *ctx)
+{
+	func->vtab->begin(ctx);
+}
+
+static inline void
+func_adapter_end(struct func_adapter *func, struct func_adapter_ctx *ctx)
+{
+	func->vtab->end(ctx);
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,343 @@
...
+static int
+lua_func_adapter_call(struct func_adapter *base,
+		      struct func_adapter_ctx *base_ctx)
+{
+	struct lua_func_adapter *func = (struct lua_func_adapter *)base;
+	struct lua_func_adapter_ctx *ctx =
+		(struct lua_func_adapter_ctx *)base_ctx;
+	int nargs = lua_gettop(ctx->L) - ctx->top_svp;
+	lua_rawgeti(ctx->L, LUA_REGISTRYINDEX, func->func_ref);
+	lua_insert(ctx->L, ctx->top_svp + 1);
```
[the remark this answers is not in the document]
drewdzzz: Done.
After this change I pass `func_adapter` to `begin` instead of `call`.

src/lib/core/func_adapter.h
```diff
@@ -0,0 +1,309 @@
...
+	void (*pop_string)(struct func_adapter_ctx *ctx, const char **str,
+			   size_t *len);
+	/**
+	 * Pops null.
+	 */
+	void (*pop_null)(struct func_adapter_ctx *ctx);
+	/**
+	 * Check if the next returned value is a tuple.
+	 */
+	bool (*is_tuple)(struct func_adapter_ctx *ctx);
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,343 @@
...
+{
+	struct lua_func_adapter_ctx *ctx = (struct lua_func_adapter_ctx *)base;
+	assert(lua_isstring(ctx->L, ctx->idx));
+	*str = lua_tolstring(ctx->L, ctx->idx++, len);
+}
+
+/**
+ * Pops tuple. Returned tuple is referenced by the function and caller must
+ * unreference it.
+ */
```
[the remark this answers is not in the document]
drewdzzz: Done.

test/unit/lua_func_adapter.c
```diff
@@ -0,0 +1,398 @@
...
+test_iterator_next(struct func_adapter *func, struct func_adapter_ctx *ctx,
+		   void *state)
+{
+	struct test_iterator_state *test_state =
+		(struct test_iterator_state *)state;
+
+	for (int i = 0; i < 3; ++i) {
+		if (test_state->current > test_state->limit)
+			break;
+		func_adapter_push_integer(func, ctx, test_state->current++);
```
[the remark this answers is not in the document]
drewdzzz: There is a plan to migrate all the triggers, including transactional ones, to the new registry. But transactional triggers will be migrated later (in the scope of follow ups), so I removed iterators from the patch - let's design them when we need it.

src/lib/core/func_adapter.h
```diff
@@ -0,0 +1,309 @@
...
+	 */
+	void (*push_number)(struct func_adapter_ctx *ctx, double number);
+	/**
+	 * Pushes integer argument.
+	 */
+	void (*push_integer)(struct func_adapter_ctx *ctx, int64_t integer);
+	/**
+	 * Pushes string argument.
+	 */
+	void (*push_lstring)(struct func_adapter_ctx *ctx, const char *str,
```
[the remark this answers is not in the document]
drewdzzz: I renamed this virtual method to `push_string`. But what's about wrappers, we will need both of them - should I leave `func_adapter_push_lstring` wrapper as it is, or come up with a better name? (Maybe, `push_cstring` for zero terminated strings and `push_string` for strings with explicitly passed length?)
As an alternative, we can always pass length (explicitly pass `strlen(str)` for zero-terminated ones).

src/lib/core/func_adapter.h
```diff
@@ -0,0 +1,284 @@
...
+func_adapter_push_lstring(struct func_adapter *func,
+			  struct func_adapter_ctx *ctx, const char *str,
+			  size_t len)
+{
+	func->vtab->push_string(ctx, str, len);
+}
+
+static inline void
+func_adapter_push_string(struct func_adapter *func,
+			 struct func_adapter_ctx *ctx, const char *str)
```
locker, line 193: Better call it `push_str0`. Also, there's no point to stick to Lua API naming conventions. Better name the methods `push_str`, `push_str0`, `push_int`, `push_double`. It'd be more consistent with other parts of the Tarantool internals.

src/lib/core/func_adapter.h
```diff
@@ -0,0 +1,284 @@
...
+	bool (*is_tuple)(struct func_adapter_ctx *ctx);
+	/**
+	 * Pops tuple. Returned tuple is referenced by the function and caller
+	 * must unreference it. Never returns NULL.
+	 */
+	void (*pop_tuple)(struct func_adapter_ctx *ctx, struct tuple **tuple);
+	/**
+	 * Check if the next returned value is a number.
+	 */
+	bool (*is_number)(struct func_adapter_ctx *ctx);
```
locker, line 89: Does it return true if the value is of type `int64_t`? If yes, what would `pop_number` return in this case? Can it result in loss of precision? (e.g. 2^63-1 can't be stored in double precisely). Similar question to `is_integer`: does it return true if the value has a fractional part?

Please describe the behavior in the comment.

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,280 @@
...
+};
+
+static_assert(sizeof(struct lua_func_adapter_ctx) <= sizeof(struct func_adapter_ctx),
+	      "sizeof(lua_func_adapter_ctx) must be <= "
+	      "sizeof(func_adapter_ctx)");
+
+/**
+ * Specialization of func_adapter for Lua functions.
+ */
+struct lua_func_adapter {
```
locker, line 42: `lua_` is the Lua API namespace. We shouldn't use it for Tarantool internals. Probably better call this `func_adapter_lua` (similarly to `port_lua`, `func_lua`).

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,280 @@
...
+	struct lua_func_adapter_ctx *ctx = (struct lua_func_adapter_ctx *)base;
+	assert(lua_isnumber(ctx->L, ctx->idx));
+	*out = lua_tointeger(ctx->L, ctx->idx++);
+}
+
+static bool
+lua_func_adapter_is_string(struct func_adapter_ctx *base)
+{
+	struct lua_func_adapter_ctx *ctx = (struct lua_func_adapter_ctx *)base;
+	return lua_gettop(ctx->L) >= ctx->idx && lua_isstring(ctx->L, ctx->idx);
```
locker, line 175: `lua_isstring` returns true for numbers, which is probably not what we want.

src/box/lua/func_adapter.c
```diff
@@ -0,0 +1,280 @@
...
+{
+	struct lua_func_adapter_ctx *ctx = (struct lua_func_adapter_ctx *)base;
+	assert(lua_isnumber(ctx->L, ctx->idx));
+	*out = lua_tonumber(ctx->L, ctx->idx++);
+}
+
+static bool
+lua_func_adapter_is_integer(struct func_adapter_ctx *base)
+{
+	/* There is no integer type in Lua. */
```
locker, line 159: There's int64 type in Tarantool. Also, we could return true if the Lua number can be converted to `int64_t` and false otherwise.

test/unit/lua_func_adapter.c
```diff
@@ -0,0 +1,267 @@
...
+	func_adapter_end(func, &ctx);
+	func_adapter_destroy(func);
+
+	double number_expected = 3 * 5 * 7 * 11;
+	int64_t integer_expected = 3 + 5 + 7 + 11;
+	ok(number_eq(number_expected, number_retval),
+	   "Returned number must be as expected");
+	is(integer_retval, integer_expected,
+	   "Returned integer must be as expected");
+
```
locker, line 78: You should reset the Lua stack after each test (`lua_settop(L, 0)`) otherwise they might affect each other.

test/unit/lua_func_adapter.c
```diff
@@ -0,0 +1,267 @@
...
+static bool
+number_eq(double a, double b)
+{
+	return fabs(a - b) < EPS;
+}
+
+#undef EPS
+
+static struct lua_State *L;
+static struct func_adapter_ctx ctx;
```
locker, line 31: Nit: There's no need for `ctx` to be global. Please define it on stack where it's used.

test/unit/lua_func_adapter.c
```diff
@@ -0,0 +1,267 @@
...
+
+int
+main(void)
+{
+	memory_init();
+	fiber_init(fiber_c_invoke);
+	tuple_init(NULL);
+
+	L = luaT_newteststate();
+	tarantool_L = L;
```
locker, line 252: Nit: There's no need to make `L` global if you set `tarantool_L` to it anyway. Please use `tarantool_L` instead of `L`.

src/lib/core/func_adapter.h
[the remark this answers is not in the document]
drewdzzz: Thanks, named it so.

src/lib/core/func_adapter.h
drewdzzz: Done (`push_int` was removed after f2f discussion).

src/box/lua/func_adapter.c
drewdzzz: Done. Note that the unit test is still called `lua_func_adapter.c` in order to group all unit tests for Lua.

src/box/lua/func_adapter.c
drewdzzz: Thanks, changed it to `lua_type` here.

src/box/lua/func_adapter.c
drewdzzz: Discussed f2f, decided not to imlement this functionality.

test/unit/lua_func_adapter.c
drewdzzz: Done.

test/unit/lua_func_adapter.c
drewdzzz: Done.

test/unit/lua_func_adapter.c
drewdzzz: Done.

src/lib/core/func_adapter.h
drewdzzz: Left such a comment:
```c
/**
  * Checks if the next returned value is a number that can be represented
  * by double without loss of precision.
  */
bool (*is_double)(struct func_adapter_ctx *ctx);
```

drewdzzz:
1. Since we pass a Lua object directly, I added a test case that checks if a callable table can be wrapped by a `func_adapter`.
2. I moved typecheck assertions to virtual method wrappers, an example:
```c
static inline void
func_adapter_pop_double(struct func_adapter *func,
			struct func_adapter_ctx *ctx, double *out)
{
	assert(func_adapter_is_double(func, ctx));
	func->vtab->pop_double(ctx, out);
}
```

locker (approved):

alyapunov (approved):

PR #8774, opened 2023-06-15
build: replace string with JOIN option to string_join helper in CMake

### LINKED ISSUE #5881: Add corresponding check with cmake_minimum_required for ASan build
If one tries to build Tarantool with ASan using old CMake version, he sees the following error:
```bash
$ CC=clang CXX=clang++ cmake . -DENABLE_WERROR=ON -DENABLE_ASAN=ON -DENABLE_UB_SANITIZER=ON -DENABLE_FUZZER=ON
...
CMake Error at cmake/compiler.cmake:284 (string):
  string does not recognize sub-command JOIN
  src/CMakeLists.txt:4 (enable_tnt_compile_flags)
```

CMake version:
```
$ cmake --version
cmake version 3.1.3
``` 

According to [this](https://cmake.org/cmake/help/latest/command/string.html#join) section of CMake documentation `string(JOIN, ...)` was introduced in CMake version 3.10.

CuriousGeorgiy:
Closes #5881

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -145,3 +145,21 @@ function(list_add_prefix
     endforeach()
     set(${list_out} ${result} PARENT_SCOPE)
 endfunction()
+
+# string() has JOIN option but only since 3.12.
+function(string_join
+    glue
+    string_out
+)
+    set(result "")
+    set(is_first_item TRUE)
+    foreach(item IN LISTS ARGN)
+        if(is_first_item)
+            set(result "${item}")
+            set(is_first_item FALSE)
+        else()
+            set(result "${result}${glue}${item}")
+        endif()
+    endforeach()
+    set(${string_out} ${result} PARENT_SCOPE)
+endfunction()
--- a/cmake/compiler.cmake
+++ b/cmake/compiler.cmake
@@ -1,3 +1,5 @@
+include(cmake/utils.cmake)
+
 #
 # Check if the same compile family is used for both C and CXX
 #
@@ -234,7 +236,7 @@ macro(enable_tnt_compile_flags)
         if (NOT CMAKE_COMPILER_IS_CLANG)
             message(FATAL_ERROR "Undefined behaviour sanitizer only available for clang")
         else()
-            string(JOIN "," SANITIZE_FLAGS
+            string_join("," SANITIZE_FLAGS
                 alignment bool bounds builtin enum float-cast-overflow
                 float-divide-by-zero function integer-divide-by-zero return
                 shift unreachable vla-bound


cmake/utils.cmake
```diff
@@ -145,3 +145,15 @@ function(list_add_prefix
...
+function(string_join
+    glue
+    string_out
+)
+    set(result "")
+    foreach(item IN LISTS ARGN)
+        string(CONCAT result "${i}${glue}")
+    endforeach()
+    set(${list_out} ${result} PARENT_SCOPE)
+endfunction()
```
locker, line 159: This function simply doesn't work...

locker (changes requested):

cmake/utils.cmake
CuriousGeorgiy: Sorry, I relied on building with `-DENABLE_UB_SANITIZER`, and it didn't give any errors, so I didn't double-check.

Fixed.

cmake/utils.cmake
```diff
@@ -145,3 +145,16 @@ function(list_add_prefix
...
+
+# string() has JOIN option but only since 3.12.
+function(string_join
+    glue
+    string_out
+)
+    set(result "")
+    foreach(item IN LISTS ARGN)
+        set(result "${result}${item}${glue}")
+        message(STATUS "result=${result}")
```
locker, line 157: The `message` is obviously redundant.

In contrast to the builtin function, this one adds a dangling glue to the end of the result string.

locker (changes requested):

cmake/utils.cmake
CuriousGeorgiy: Fixed.

cmake/utils.cmake
```diff
@@ -145,3 +145,18 @@ function(list_add_prefix
...
+# string() has JOIN option but only since 3.12.
+function(string_join
+    glue
+    string_out
+)
+    set(result "")
+    foreach(item IN LISTS ARGN)
+        set(result "${result}${item}${glue}")
+    endforeach()
+    string(LENGTH ${result} len)
```
locker, line 158: `${result}` must be enclosed in quotes otherwise the function wouldn't work if the item list is empty.

Anyway, adding the terminating comma just to strip it on the next line looks ugly. Let's not add it in the first place:

```cmake
function(string_join
    glue
    string_out
)
    set(result "")
    set(is_first_item TRUE)
    foreach(item IN LISTS ARGN)
        if(is_first_item)
            set(result "${item}")
            set(is_first_item FALSE)
        else()
            set(result "${result}${glue}${item}")
        endif()
    endforeach()
    set(${string_out} ${result} PARENT_SCOPE)
endfunction()
```

locker (changes requested):

cmake/utils.cmake
CuriousGeorgiy: This doesn't look perfect either, but I guess it's better, applied your suggestion.

locker (approved):

Issue #7640, opened 2022-09-01
LuaJIT: `ffi.C.dlsym()` doesn't work on FreeBSD with PIC


Gumix:
**Bug description**

`ffi.C.dlsym()` doesn't work when Tarantool is compiled with Position-Independent Code enabled.

* OS: FreeBSD
* OS Version: 12 and 13
* Architecture: amd64

```
Tarantool 2.11.0-entrypoint-426-g08070d45c
Target: FreeBSD-amd64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=FALSE
Compiler: /usr/bin/cc /usr/bin/c++
C_FLAGS:-fPIC -fexceptions -funwind-tables -fno-common -fopenmp -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression
CXX_FLAGS:-fPIC -fexceptions -funwind-tables -fno-common -fopenmp -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression
```

**Steps to reproduce**

Build Tarantool with the following flags:
```bash
cmake .. \
-DCMAKE_BUILD_TYPE=RelWithDebInfo \
-DCMAKE_C_FLAGS="-fPIC" \
-DCMAKE_CXX_FLAGS="-fPIC" \
-DCMAKE_EXE_LINKER_FLAGS="-pie"
```

Reproducer script:

```lua
local ffi = require('ffi')

ffi.cdef([[
    void *dlsym(void *handle, const char *symbol);
    char *dlerror(void);
]])

for i = 1, 64 do
    local RTLD_DEFAULT = ffi.cast("void *", 0LL)
    local sym = ffi.C.dlsym(RTLD_DEFAULT, 'box_select')
    local err = ffi.C.dlerror()
    err = err == ffi.NULL and 'no error' or ffi.string(err)
    print(i .. '\t' .. tostring(sym) .. '\t' .. err)
end
```

How to run: just type `tarantool test.lua`.

**Actual behavior**

```
1	cdata<void *>: 0x012d48f0	no error
2	cdata<void *>: 0x012d48f0	no error
[...]
46	cdata<void *>: 0x012d48f0	no error
47	cdata<void *>: 0x012d48f0	no error
[...]

PR #8784, opened 2023-06-16
Remove dependencies on libgomp and enable hardening on AArch64 and FreeBSD

Gumix:


cmake/hardening.cmake
```diff
@@ -1,9 +1,7 @@
 # Depends on os.cmake and profile.cmake modules.
-# Uses `ENABLE_FUZZER` option and `TARGET_OS_FREEBSD` variable.
+# Uses `ENABLE_FUZZER` option and `TARGET_OS_DARWIN` variable.
 
-# LuaJIT in FreeBSD doesn't work with PIC (gh-7640),
```
Totktonada, line 4: Isn't #7640 still relevant?

Gumix: https://github.com/tarantool/tarantool/pull/8784/commits/376b0b093105fe0cd6232a34d1e2d6b92980c26c

**build: do not disable hardening on FreeBSD**
> It was disabled because dlsym tests failed when PIC is on (https://github.com/tarantool/tarantool/issues/7640).
> However, later it turned out that the issue is not related to PIC, and
> LuaJIT was turned off for such tests by commit https://github.com/tarantool/tarantool/commit/67e79b15f03fb6b7149a8f8a6f35932d785a314c ("test: turn
> LuaJIT off in tests with dlsym"). Now it's safe to turn on PIC for FreeBSD.

andreyaksenov (approved):

cmake/hardening.cmake
```diff
@@ -1,9 +1,8 @@
...
 # ligomp.a for AArch64 CentOS is compiled without PIC support.
-if (ENABLE_FUZZER OR TARGET_OS_FREEBSD OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64")
+if (ENABLE_FUZZER OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64")
     set(ENABLE_HARDENING_DEFAULT FALSE)
 else()
     set(ENABLE_HARDENING_DEFAULT TRUE)
```
nshy, line 7: Nit: May be we should add "Follow-up" reference to that hardening ticket.

nshy (approved):

cmake/hardening.cmake
Totktonada: Thanks for the explanation!

locker (approved):

locker:
Cherry-picked the second commit (the one about enabling hardening on FreeBSD) to 2.11 and 2.10. The first and the third commit weren't cherry-picked because Tarantool 2.11 and 2.10 still depend on libgomp.

PR #8786, opened 2023-06-19
yaml: don't encode unprintable strings as binary blobs

### LINKED ISSUE #8756: Encoding of unprintable strings in YAML works unpredictably
Reproduced with 3.0.0-entrypoint-300-g83158695aae5.
2.11 and 2.10 are affected as well.

In some cases unprintable characters are escaped but in other cases the whole string is converted to binary data:

```
tarantool> '\x11'
---
- "\x11"
...

tarantool> '\xff'
---
- !!binary /w==
...
```

We're going to introduce a new type for binary data, see #1629. The type will always be encoded to YAML with the binary tag (even if it contains a printable string). A YAML field with the binary tag will be decoded to a binary Lua object, not to a plain string. To avoid confusion, we must always encode plain strings in YAML as strings with non-printable characters escaped and never implicitly convert them to binary data.

locker:
Historically, we encode strings that contain invalid or non-printable utf-8 sequences in YAML as binary base64 blobs. We do that because of limitations/bugs of the YAML encoder, which refuses to encode invalid utf-8 strings. To work around this issue, we introduced the helper `utf8_check_printable`, which is basically a copy of `yaml_check_utf8`, and treat strings for which it fails as binary data (`MP_BIN`).

This commit updates the YAML submodule to the version where all known issues with encoding invalid/unprintable utf-8 strings are fixed and removes special treatment of such strings (drops `utf8_check_printable`). Now unprintable or invalid utf-8 sequences are emitted as code points, e.g. `\xFF` or `\uFFFF`. This change is a pre-requisite for introducing the new varbinary type to Lua. Without it plain strings would be implicitly converted to varbinary after decoding/encoding them in YAML,
which would be confusing.

Closes #8756
Related PR https://github.com/tarantool/libyaml/pull/2


alyapunov (approved):

changelogs/unreleased/fselect-multiline-output.md
```diff
@@ -0,0 +1,7 @@
+## feature/box
+
+* Changed the output of the `fselect` index method from a table of strings to
+  a multi-line string and dropped the `print` and `use_nbsp` options. With the
+  `yaml_pretty_multiline` compat option enabled by default, multi-line strings
+  now look good in the console so there's no need to return a table of strings
```
andreyaksenov, line 6:
```suggestion
  now look good in the console, so there's no need to return a table of strings
```

changelogs/unreleased/gh-8756-yaml-unprintable-utf8-encoding-fix.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Eliminated implicit conversion of unprintable utf-8 strings to binary blobs
+  when encoded in YAML. Now unprintable characters are encoded as escaped utf-8
+  code points, for example `\x80` or `\u200B` (gh-8756).
```
andreyaksenov, line 5:
```suggestion
  code points, for example, `\x80` or `\u200B` (gh-8756).
```

andreyaksenov (approved):

changelogs/unreleased/fselect-multiline-output.md
locker: Fixed, thanks!

changelogs/unreleased/gh-8756-yaml-unprintable-utf8-encoding-fix.md
locker: Fixed, thanks!

PR #8787, opened 2023-06-19
ci: extend default tests run with osx wokflows

ylobankov:
It was decided to include the `osx_debug.yml` and `osx_release.yml` workflows to the default tests run (without the `full-ci` label). Now we can get test results for macOS faster and without an extra load on CI.


igormunkin (approved):
@ylobankov, thanks the changes! LGTM.

PR #8808, opened 2023-06-23
Introduce the varbinary type to Lua

### LINKED ISSUE #1629: Insert binary data with Lua
[PM link](https://www.notion.so/tarantool/Support-varbinary-in-Lua-91ca22a16b5e4f10b3cd2a6c3eb8dc23)
[Design document](https://www.notion.so/tarantool/varbinary-in-Lua-a0ce453dcf5a46e3bc421bf80d4cc276)

In MessagePack there is a distinction between "the str format family" and "the bin format family".
https://github.com/msgpack/msgpack/blob/master/spec.md
It seems to me that there is no way to use the bin format family with Lua.
 If I enter '\x41' or [[A]] I get MessagePack string, and if I try to insert cdata, or pass cata to msgpack.encode I get "error: unsupported Lua type 'cdata'".
I brought this matter up with Konstantin Osipov and in his reply email he wrote:
"If you can't figure out how to do it, please file a bug that there is no way to insert binary data from Lua, and maybe some developer fixes it."

locker:
The new module `varbinary` was introduced. The module implements the following functions:
 - `varbinary.new` - constructs a varbinary object from a plain string or cdata pointer and size (to be used with the `buffer` module).
 - `varbinary.is` - returns true if the argument is a varbinary object.

```Lua
local bin = varbinary.new('data')
assert(varbinary.is(bin))
assert(not varbinary.is('data'))
```

Like a plain string, a varbinary object stores arbitrary data. Unlike a plain string, it's encoded as a binary blob by the built-in encoders that support the binary type (MsgPack, YAML). (Actually, encoding binary blobs with the proper type is the main goal of the new type.)

```
tarantool> '\xFF\xFE'
---
- "\xFF\xFE"
...

tarantool> varbinary.new('\xFF\xFE')
---
- !!binary //4=
...

tarantool> msgpack.encode('\xFF\xFE')
---
- "\xA2\xFF\xFE"
...

tarantool> msgpack.encode(varbinary.new('\xFF\xFE'))
---
- "\xC4\x02\xFF\xFE"
...
```

Note, the JSON format doesn't support the binary type so a varbinary object is still encoded as a plain string:

```
tarantool> json.encode('\xFF\xFE')
---
- "\"\xFF\xFE\""
...

tarantool> json.encode(varbinary.new('\xFF\xFE'))
---
- "\"\xFF\xFE\""
...
```

The built-in decoders now decode binary data fields (fields with the 'binary' tag in YAML; the `MP_BIN` type in MsgPack) to a varbinary object by default:

```
tarantool> varbinary.is(msgpack.decode('\xC4\x02\xFF\xFE'))
---
- true
...

tarantool> varbinary.is(yaml.decode('!!binary //4='))
---
- true
...
```

This also implies that the data stored in the database under the 'varbinary' field type is now returned to Lua not as a plain string,
but as a varbinary object. It's possible to revert to the old behavior by toggling the new compat option `binary_data_decoding` because this change may break backward compatibility:

```
tarantool> compat.binary_data_decoding = 'old'
---
...

tarantool> varbinary.is(msgpack.decode('\xC4\x02\xFF\xFE'))
---
- false
...

tarantool> varbinary.is(yaml.decode('!!binary //4='))
---
- false
...
```

A varbinary object implements the following meta-methods:
- `__len` - returns the length of the binary data, in bytes.
- `__tostring` - returns the data in a plain string.
- `__eq` - returns true if the varbinary object contains the same data as another varbinary object or a string.

```Lua
local bin = varbinary.new('foo')
assert(#bin == 3)
assert(tostring(bin) == 'foo')
assert(bin == 'foo')
assert(bin ~= 'bar')
assert(bin == varbinary.new('foo'))
assert(bin ~= varbinary.new('bar'))
```

There are no string manipulation methods, like `string.sub` or `string.match`. If you need to match a substring in a varbinary object, you have to convert it to a string first.

For more details, see the [design document][1].

Closes #1629

[1]: https://www.notion.so/tarantool/varbinary-in-Lua-a0ce453dcf5a46e3bc421bf80d4cc276


alyapunov (approved):

p7nov (approved):

locker:
The go-tarantool integration test fails because it sends the UUID argument of `queue.identify()` as a binary blob, not a plain string, while this function expects a string. This works fine without this PR because Tarantool decodes `MP_BIN` to a plain string. This PR changes this behavior: `MP_BIN` is now decoded to a varbinary object.

Fix for go-tarantool: https://github.com/tarantool/go-tarantool/pull/313

PR #8811, opened 2023-06-26
config: replicaset building enhancements

### LINKED ISSUE #8810: config: improve replicaset building experience
The basic flow that is implemented in #8789 supports automatic building of peers URI list for a replicaset from instances' `iproto.advertise` and `iproto.listen` options. However, there are a few possible enhancement.

* Verify that `iproto.advertise` is a valid URI and it is suitable to connect to an instance: the host is not `0.0.0.0` or `::`, the port is not zero.
* Verify that `iproto.listen` is a valid list of URIs.
* Handle multiple `iproto.listen` URIs in the code that forms peer URIs.
* Filter out `iproto.listen` URIs that are not suitable to connect in the code that forms peer URIs.
* Support `user@`. `user:pass@` and `user@host:port` syntax for `iproto.advertise`, which means to get host/port from `iproto.listen` (if the advertise option has no one) and get a password from the `credentials` section (if it is not set in the advertise option).
* Split `iproto.advertise` to several parameters:

  ```yaml
  iproto:
    advertise:
      client: <string>
      peer: <string>
      sharding: <string> # uses peer if not set
  ```
* Implement several failover modes: no failover, manual failover, election failover (RAFT). Set appropriate ro/rw database modes for each failover mode.

Part of #8724

Totktonada:
This pull request accumulates implementation of several techdebt tasks and new suggestions around autobuilding of replication upstream URIs from peer URIs, replicaset startup process and several basic failover modes. See #8810 for the general description and commit messages for details.

Closes #8810.


ImeevMA (approved):
Hi! Thank you for the patch-set! I'm not a replication expert, so I just followed the commit descriptions and looked at the code, and with that in mind, I think this PR is LGTM.

PR #8815, opened 2023-06-27
test: fix typo in serializer test

ArtDu:
ref: https://stackoverflow.com/questions/1848700/biggest-integer-that-can-be-stored-in-a-double

### DIFF
--- a/test/app-tap/lua/serializer_test.lua
+++ b/test/app-tap/lua/serializer_test.lua
@@ -75,7 +75,7 @@ local function test_unsigned(test, s)
     rt(test, s, 4294967297LL, "number")
     rt(test, s, 4294967297ULL, "number")
 
-    -- 1e52 - maximum int that can be stored to double without losing precision
+    -- 2^53 - maximum int that can be stored to double without losing precision
     -- decrease double capacity to fit .14g output format
     rt(test, s, 99999999999999, "number")
     rt(test, s, 99999999999999LL, "number")
@@ -131,7 +131,7 @@ local function test_signed(test, s)
     rt(test, s, -2147483648, 'number')
     rt(test, s, -2147483648LL, 'number')
 
-    -- 1e52 - maximum int that can be stored to double without losing precision
+    -- 2^53 - maximum int that can be stored to double without losing precision
     -- decrease double capacity to fit .14g output format
     rt(test, s, -99999999999999, "number")
     rt(test, s, -99999999999999LL, "number")


test/app-tap/lua/serializer_test.lua
```diff
@@ -75,7 +75,7 @@ local function test_unsigned(test, s)
     rt(test, s, 4294967297LL, "number")
     rt(test, s, 4294967297ULL, "number")
 
-    -- 1e52 - maximum int that can be stored to double without losing precision
+    -- 2^53 - maximum int that can be stored to double without losing precision
     -- decrease double capacity to fit .14g output format
```
Totktonada, line 79: I failed to find a link between the comment and the test cases until I looked over the history: 093828c6e44f997914253d6281ae0237b8cbb4de. Shouldn't we rewrite the comment to bring more context to a reader?

ArtDu: You're right, we need to expand it. I also understood the meaning of the comment after reading the history

ArtDu: We can add description from here:
https://github.com/tarantool/tarantool/blob/a19d1b73ba03a37f41061af68a1893dbd6c7c974/src/lua/utils.c#L455-L460

ArtDu, line 79:
```suggestion
    -- 1e14-1 - maximum integer that doesn't lose precision on tostring() conversion.
    -- Lua uses sprintf("%.14g") to format its numbers, see gh-1279.
```

test/app-tap/lua/serializer_test.lua
```diff
@@ -131,7 +131,7 @@ local function test_signed(test, s)
     rt(test, s, -2147483648, 'number')
     rt(test, s, -2147483648LL, 'number')
 
-    -- 1e52 - maximum int that can be stored to double without losing precision
+    -- 2^53 - maximum int that can be stored to double without losing precision
     -- decrease double capacity to fit .14g output format
```
ArtDu, line 135:
```suggestion
    -- -1e14+1 - minimum integer that doesn't lose precision on tostring() conversion.
    -- Lua uses sprintf("%.14g") to format its numbers, see gh-1279.
```

test/app-tap/lua/serializer_test.lua
Totktonada: I would start from the goal of the cases and their constraints. Say,

> Verify that a marshaling over the serializer gives the same value for the given cases. IOW, s.decode(s.encode(x)) == x and the Lua type remains the same.
>
> This property is true for integral numbers within the [-10^14 + 1; 10^14 - 1] range. Decoding of an integral number beyond this range gives a cdata number.
>
> The reason of this decoding behavior is a peculiar of Lua's tostring() function. It interprets all number values out of the given range as floats and prints them in the scientific notation with possible precision loss. (It uses sprintf("%.14g") under the hood.)
>
> For a better user experience all the numbers beyond the range are decoded to cdata<int64\> or cdata<uint64\>. tostring() always shows cdata numbers without precision loss.

Totktonada: BTW, if you're interested I investigated this topic some time ago and I have a related picture.

```lua
--       +-------- <number64> is always encoded as integer ----------+
--       |                                                           |
--       |     +-- encode <number> as floating point number ----+    |
--       |     |                                                |    |
--       |     |   +---- encode <number> as integer ----+       |    |
--       |     |   |  (when it has no fractional part)  |       |    |
--       |   <-+   |                                    |       +->  |
-- -------------------------------------------------------------------------
-- ... -2^63 ... -10^14 + 1 ......................... 10^14 - 1 ... 2^64 ...
-- -------------------------------------------------------------------------
-- <-+   |         |                                    |            |   +->
--   |   |         +--- decode integer into <number> ---+            |   |
--   |   |                                                           |   |
--   |   +------------ decode integer into <number64> ---------------+   |
--   |                                                                   |
--   +----------------- error on decoding an integer --------------------+
```

It is from my own serializer, but it reflects behavior of the built-in ones (at least it was the one of the goals).

You can find more related (and unrelated) details in comments in the code and the test.

* https://github.com/tarantool/conf-legacy/blob/master/conf/client/etcd/scalar_serializer.lua
* https://github.com/tarantool/conf-legacy/blob/master/test/client/etcd/scalar_serializer_test.lua

Sadly, it is part of a closed source project. I think I should create a gist for sharing in discussions like this one.

ArtDu: Updated suggestions below

Totktonada: Since these words are from me, I would prefer to leave the review to someone else. However, before you'll ask for a review, please ensure that the following conditions are met.

1. The code to review is in your patchset, not somewhere else (suggestion, gist, etc).
2. CI is green (now it fails on the style check).
3. The code follows our code style (the suggestion contains very long lines).
4. The commit message follows our convention (see the lint CI job, it is red now).

Please, follow our [contributing guide](https://github.com/tarantool/tarantool/blob/master/CONTRIBUTING.md) or, if you lost your interest, close the pull request.

PR #8822, opened 2023-06-29
core, box: fix cord buffer leaks on error conditions in Lua APIs

### LINKED ISSUE #8677: Memory leak in msgpackffi.encode on encode error
Tarantool 3.0.0-entrypoint-217-g00a9e59927
Target: Linux-x86_64-Debug

Repro:
```lua
local bad = setmetatable({}, {__serialize = 1})
local mp = require('msgpackffi')
for i = 1,10000 do
    pcall(mp.encode, {string.rep('x', 1024 * 1024), bad})
end
```
This program is killed by the OOM killer in about 10 seconds or so.

The issue is `cord_ibuf_drop` is not called on cleanup if error is thrown. However allocated cord ibufs will be released on next fiber yield/stop.

The issue introduced in 2.8.0-152-gade45685f when cord ibuf was introduced.

Let's inspect all other cord ibuf usages in scope of this issue. I also suggest to add unconditional assertion to `cord_buf_on_yield`/`cord_buf_on_stop`. I guess it is program mistake if cord buf it not released. It will also help to catch similar bugs in the future.

Similar to #7939

CuriousGeorgiy:
Closes #8677

### DIFF
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -1967,12 +1967,19 @@ local function normalize_position(index, pos, ret_pos, ret_pos_end)
         ret_pos_end[0] = ret_pos[0] + #pos
     else
         local tuple_ibuf = cord_ibuf_take()
-        local tuple, tuple_end = tuple_encode(tuple_ibuf, pos)
-        local nok = builtin.box_index_tuple_position(index.space_id, index.id,
-                                                     tuple, tuple_end,
-                                                     ret_pos, ret_pos_end) ~= 0
+        box.error.clear()
+        local ok, res = pcall(function()
+            local tuple, tuple_end = tuple_encode(tuple_ibuf, pos)
+            return builtin.box_index_tuple_position(index.space_id, index.id,
+                                                    tuple, tuple_end,
+                                                    ret_pos, ret_pos_end) == 0
+        end)
         cord_ibuf_put(tuple_ibuf)
-        if nok then
+        ok = ok and res
+        if not ok then
+            if box.error.last() == nil then
+                error(res)
+            end
             box.error()
         end
     end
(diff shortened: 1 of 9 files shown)


src/box/lua/schema.lua
```diff
@@ -1967,12 +1967,18 @@ local function normalize_position(index, pos, ret_pos, ret_pos_end)
...
-        local tuple, tuple_end = tuple_encode(tuple_ibuf, pos)
-        local nok = builtin.box_index_tuple_position(index.space_id, index.id,
-                                                     tuple, tuple_end,
-                                                     ret_pos, ret_pos_end) ~= 0
+        local ok, res = pcall(function()
+            local tuple, tuple_end = tuple_encode(tuple_ibuf, pos)
+            return builtin.box_index_tuple_position(index.space_id, index.id,
+                                                    tuple, tuple_end,
+                                                    ret_pos, ret_pos_end) == 0
+        end)
```
locker, line 1975: In the [original discussion](https://lists.tarantool.org/tarantool-patches/06d4c252-3f41-d91b-6943-ec4cca520a99@tarantool.org/) (which is mentioned in a [comment](https://github.com/tarantool/tarantool/issues/8677#issuecomment-1556740005) to this issue), @Gerold103 claimed that not using `pcall` was an intentional design decision to avoid possible performance degradation. Have you run any performance tests to disprove his statement?

p7nov (approved):

src/box/lua/schema.lua
CuriousGeorgiy: <details>
<summary>I did a simple benchmark:</summary>

```lua

clock = require('clock')
log = require('log')

box.cfg{}

s:create_index('pk')
s:insert{0}

N = 100
total_time = 0
for _ = 1, N do
    total_time = total_time + clock.bench(function() box.space.s:get{0} end)[1]
end

log.info("Average time: %g ms", total_time / N * 1e6)

os.exit()
```
</details>

Results:

| Branch         | Time, ms        |
| -------------- | --------------- |
| Master         | 4.79248         |
| Patch          | 5.18757         |

Issue #6128, opened 2021-06-06
Background start of Tarantool fails with LuajitError: cannot read stdin


ochaton:
**Bug description**

* OS: MacOS
* Tarantool 2.8.1-39-g77c0af501

## Reproducer
```bash
TT_PID_FILE=tarantool.pid TT_LOG=tarantool.log TT_BACKGROUND=true TT_LISTEN=3000 ~/tarantool/2.8.1/bin/tarantool -e 'box.cfg{}'

cat tarantool.log
```

```
2021-06-06 10:31:39.173 [96924] main/103/interactive C> Tarantool 2.8.1-39-g77c0af501
2021-06-06 10:31:39.174 [96924] main/103/interactive C> log level 5
2021-06-06 10:31:39.174 [96924] main/103/interactive I> wal/engine cleanup is paused
2021-06-06 10:31:39.175 [96924] main/103/interactive I> mapping 268435456 bytes for memtx tuple arena...
2021-06-06 10:31:39.175 [96924] main/103/interactive I> Actual slab_alloc_factor calculated on the basis of desired slab_alloc_factor = 1.044274
2021-06-06 10:31:39.175 [96924] main/103/interactive I> mapping 134217728 bytes for vinyl tuple arena...
2021-06-06 10:31:39.183 [96924] main/103/interactive I> instance uuid dc94ea40-792d-4a98-aa00-8a18981ca183
2021-06-06 10:31:39.185 [96924] main/103/interactive I> : bound to [::]:3000
2021-06-06 10:31:39.185 [96924] main/103/interactive I> initializing an empty data directory
2021-06-06 10:31:39.215 [96924] main/103/interactive I> assigned id 1 to replica dc94ea40-792d-4a98-aa00-8a18981ca183
2021-06-06 10:31:39.215 [96924] main/103/interactive I> cluster uuid f34a5b7e-1e50-4b9e-9d93-03b4097f04eb
2021-06-06 10:31:39.217 [96924] snapshot/101/main I> saving snapshot `./00000000000000000000.snap.inprogress'
2021-06-06 10:31:39.218 [96924] snapshot/101/main I> done
2021-06-06 10:31:39.219 [96924] main/103/interactive I> ready to accept requests
2021-06-06 10:31:39.219 [96924] main/104/gc I> wal/engine cleanup is resumed
2021-06-06 10:31:39.219 [96924] main/103/interactive I> set 'log_level' configuration option to 5
[...]

PR #8823, opened 2023-06-29
test: fix for gh_6128 test to avoid flakiness

Lord-KA:
Followup to #6128

NO_DOC=test fix
NO_CHANGELOG=test fix

### DIFF
--- a/test/app-luatest/gh_6128_background_mode_test.lua
+++ b/test/app-luatest/gh_6128_background_mode_test.lua
@@ -12,10 +12,16 @@ local function tarantool_path(arg)
     return arg[index + 1]
 end
 
+-- Check presence of string 'msg' in file 'file'.
+-- Returns true when string is found and false otherwise.
+-- Function is not as smart as grep_log in a luatest (luatest/server.lua)
+-- and reads the whole log file every time, but this log file has a small
+-- size so it is ok.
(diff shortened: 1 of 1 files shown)


test/app-luatest/gh_6128_background_mode_test.lua
```diff
@@ -60,7 +60,7 @@ g.after_test("test_background_mode_box_cfg", function(cg)
 end)
 
 g.test_background_mode_box_cfg = function(cg)
-    t.helpers.retrying({timeout = 2, delay = 0.01}, function()
+    t.helpers.retrying({timeout = 5, delay = 0.5}, function()
```
ligurio, line 63: Nit: I would add a comment where say that increased delay is required here.

ligurio (approved):
Gleb, thanks for fixing flakiness! I have no objections, so LGTM.
However, I've added a minor comment, see inline.

Lord-KA:
@ligurio, thanks for the alternative fix! Your solution solves the underneath problem, so lets stick with it. I can't set a gh approve because this is my PR, but anyway LGTM.

igormunkin:
Also backported to 2.11 and 2.10 in scope of 6381fc211 and 6a2c73f81 respectively.

Issue #7327, opened 2022-06-29
Teach ASAN to detect leaks in custom Tarantool allocators


locker:
Beside the libc malloc, Tarantool implements and uses other allocators, e.g. runtime region and memory pools, memtx arena. Currently, ASAN ignores such allocations and doesn't report leaks. One way to fix this is simply use malloc instead of our internal allocators if ASAN is enabled. Maybe, ASAN allows to mark alloc/free functions so that it can enable leak detection for custom allocators?

locker:
https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning

CuriousGeorgiy:
@ligurio started working on this: rougly, the idea is to annotate the same places as for valgrind.

PR #8826, opened 2023-06-30
misc: get rid of small _xc functions

nshy:
Small library currently depends on Tarantool core thru 'exception.h'. This is not the way to go. Let's drop this dependency and instead of moving _xc functions to Tarantool repo we can just stop using them. Our current policy is to panic on OOM in case of runtime allocation.

Part of #7327

### DIFF
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1043,7 +1043,7 @@ iproto_connection_input_buffer(struct iproto_connection *con)
 	 * (in only has unparsed content).
 	 */
 	if (ibuf_used(old_ibuf) == con->parse_size) {
-		ibuf_reserve_xc(old_ibuf, to_read);
+		xibuf_reserve(old_ibuf, to_read);
 		return old_ibuf;
 	}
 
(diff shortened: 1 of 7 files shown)


locker (approved):

src/box/iproto.cc
```diff
@@ -1043,7 +1043,7 @@ iproto_connection_input_buffer(struct iproto_connection *con)
 	 * (in only has unparsed content).
 	 */
 	if (ibuf_used(old_ibuf) == con->parse_size) {
-		ibuf_reserve_xc(old_ibuf, to_read);
+		xibuf_reserve(old_ibuf, to_read);
```
locker, line 1046: I'm not sure it's safe to assume that allocation from IPROTO buffers never fail - technically the buffers may have quota set. OTOH if we set the quota, it'll be difficult to test all the fail paths so it maybe much better to account memory consumed by buffers in one place, at the IPROTO level, not at the buffer level. Anyway, this change looks incomplete: most of the alloc paths in the IPROTO code still handle OOM. I'd probably replace `_xc` version with explicit `diag_raise` in this patch and (maybe) remove all OOM checks from IPROTO later, in a separate commit with a proper justification. Up to @alyapunov.

alyapunov: I think that use of x* versions on allocator functions should not be goal. In some cases we can think and decide that it's OK to panic in some subsystem, but there must be a practical reason for that. But we should not replace all malloc with xmalloc, for example, having the only goal to remove malloc.
I think that generally it's OK to use xregion_alloc for fiber and txn region; in all other cases we should make an identical change - replace xc version to simple, check result and raise in case of error.

alyapunov (approved):

Issue #8391, opened 2023-03-02
Bump zstd submodule to pre-1.5.5


ligurio:
See https://github.com/facebook/zstd/pull/3517

It is relevant only for levels 13..22, which are unlikely to be used with Tarantool (because they are too slow).
We use level 3, see https://github.com/tarantool/tarantool/blob/6a3e31333d3a5a10f44da6375e8bae6b1c7f88d8/src/box/xlog.c#L1161-L1162
Anyway, it's better update zstd to a newer version until final Tarantool release.

PR #8830, opened 2023-07-01
box: switch to normal streaming compression API of ZSTD

### LINKED ISSUE #8537: Bump zstd to v1.5.5
> This is a quick fix release. The primary focus is to correct a rare corruption bug in high compression mode. ... While the probability might be very small, corruption issues are nonetheless very serious, so an update to this version is highly recommended, especially if you employ high compression modes (levels 16+).

Full changelog: https://github.com/facebook/zstd/releases/tag/v1.5.5

CuriousGeorgiy:
Closes #8537
Follows up #8391

Note: tarantool/tarantool-ee does not use the streaming compression API of zstd, so we don't need to patch it.

### DIFF
--- a/changelogs/unreleased/gh-8537-bump-zstd.md
+++ b/changelogs/unreleased/gh-8537-bump-zstd.md
@@ -1,0 +1,3 @@
+## feature/build
+
+* Updated `zstd` to version 1.5.5 (gh-8537).
--- a/src/box/xlog.c
+++ b/src/box/xlog.c
@@ -817,6 +817,14 @@ xlog_init(struct xlog *xlog, const struct xlog_opts *opts)
 				 "failed to create context");
 			return -1;
 		}
+		size_t rc = ZSTD_CCtx_setParameter(xlog->zctx,
+						   ZSTD_c_compressionLevel, 3);
+		if (ZSTD_isError(rc)) {
+			ZSTD_freeCCtx(xlog->zctx);
+			diag_set(ClientError, ER_COMPRESSION,
+				 ZSTD_getErrorName(rc));
+			return -1;
+		}
 	}
 	return 0;
 }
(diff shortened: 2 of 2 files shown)


p7nov (approved):

changelogs/unreleased/gh-8537-bump-zstd.md
```diff
@@ -0,0 +1,3 @@
+## feature/build
+
+*  The `zstd` version was updated to 1.5.5 (gh-8537).
```
p7nov, line 3:
```suggestion
*  Updated `zstd` to version 1.5.5 (gh-8537).
```

CuriousGeorgiy:
tarantool/tarantool-ee does not use the streaming API of zstd, so we don't need to patch it.

src/box/xlog.c
```diff
@@ -1158,43 +1158,73 @@ xlog_tx_write_zstd(struct xlog *log)
...
-					 iov->iov_len - offset);
-		if (ZSTD_isError(zsize)) {
-			diag_set(ClientError, ER_COMPRESSION,
-				 ZSTD_getErrorName(zsize));
-			goto error;
+		ZSTD_EndDirective mode =
+			iov == log->obuf.iov + log->obuf.pos ||
+			(iov + 1)->iov_len == 0 ? ZSTD_e_end : ZSTD_e_continue;
+		while (true) {
+			if (zobuf.pos == zobuf.size) {
```
drewdzzz, line 1187: > output->pos and input->pos will be updated. They are guaranteed to remain below their respective limit.

After reading docs, I'm not sure that the output buffer is guaranteed to reach its limit. Also, [here](https://github.com/facebook/zstd/blob/dev/examples/streaming_compression.c) authors rely on return value and position of input buffer, not position of output buffer. I think we should rewrite this logic.

drewdzzz (changes requested):

src/box/xlog.c
```diff
@@ -1158,43 +1158,73 @@ xlog_tx_write_zstd(struct xlog *log)
...
 	uint32_t crc32c = 0;
 	struct iovec *iov;
-	/* 3 is compression level. */
-	ZSTD_compressBegin(log->zctx, 3);
+	size_t rc =
+		ZSTD_CCtx_setParameter(log->zctx, ZSTD_c_compressionLevel, 3);
+	if (ZSTD_isError(rc)) {
+		diag_set(ClientError, ER_COMPRESSION, ZSTD_getErrorName(rc));
+		goto error;
+	}
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/xlog.c
CuriousGeorgiy: > ZSTD_e_continue is guaranteed to make some forward progress when called, but doesn't guarantee maximal forward progress. [^1]

> The function nonetheless guarantees forward progress : it will return only after it reads or write at least 1+ byte. [^2]

[^1]: https://github.com/facebook/zstd/blob/118200f7b95deaf38b3368cb445a564f187da1a2/lib/zstd.h#L706-L707

[^2]: https://github.com/facebook/zstd/blob/118200f7b95deaf38b3368cb445a564f187da1a2/lib/zstd.h#L770

src/box/xlog.c
```diff
@@ -1158,43 +1158,73 @@ xlog_tx_write_zstd(struct xlog *log)
...
+					 */
+					obuf_alloc(&log->zbuf, zobuf.size);
+					/* Update crc32c */
+					crc32c = crc32_calc(crc32c, zobuf.dst,
+							    zobuf.size);
+				}
+				size_t remaining = zibuf.size - zibuf.pos;
+				zobuf.size =
+					MIN(ZSTD_CStreamOutSize(),
+					    ZSTD_compressBound(remaining));
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/xlog.c
CuriousGeorgiy: Removed this `if`.

src/box/xlog.c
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/xlog.c
```diff
@@ -1158,43 +1166,68 @@ xlog_tx_write_zstd(struct xlog *log)
...
+			if (zobuf.dst != NULL) {
+				/*
+				 * Advance output buffer to the end of
+				 * compressed data.
+				 */
+				obuf_alloc(&log->zbuf, zobuf.pos);
+				/* Update crc32c */
+				crc32c = crc32_calc(crc32c, zobuf.dst,
+						    zobuf.pos);
+			}
```
drewdzzz, line 1198: AFAICS, `dst` is `NULL` only at the first iteration, so let's move `obuf_alloc` to the end of the loop to get rid of this `if` clause.

src/box/xlog.c
```diff
@@ -1158,43 +1166,68 @@ xlog_tx_write_zstd(struct xlog *log)
...
+			size_t remaining = zibuf.size - zibuf.pos;
+			/*
+			 * `ZSTD_compressBound` fails, if  provided
+			 * size > `ZSTD_MAX_INPUT_SIZE`.
+			 */
+			zobuf.size = MIN(ZSTD_MAX_INPUT_SIZE,
+					 MAX(64, /* `remaining == 0` */
```
drewdzzz, line 1205: Nit: it would be nice to elaborate on this constant more (I would move it to a separate constant like `min_buf_size` and write that we are concerned about progress guarantee, so picked a constant which is equal to `compressBound(0)` in the current version of `zstd`).

drewdzzz (changes requested):

src/box/xlog.c
CuriousGeorgiy: Done.

src/box/xlog.c
CuriousGeorgiy: Done.

drewdzzz (approved):

Issue #8789, opened 2023-06-20
Startup flow with reading an external configuration


Totktonada:
See #8724 for the umbrella task description.

This issue accumulates all the tasks we need to solve #8724 (after #8613, #8725, #8778) and offer an MVP of the new startup flow.

The general structure is the following:

* Instance/cluster schemas (implemented in #8778).
* Instance/cluster configuration sources -- has `.sync()` and `.get()` functions.
* Config data object -- accumulates configuration for the given instance, for its peers and provide convenient accessors (such as `:get()` with applied defaults).
* Configuration appliers -- handlers for the given configuration options. For example, `box_cfg` applier, which prepares a table to call `box.cfg(<table>)`. Or `console` applier, which starts/stops an interactive console on a unix domain socket.
* The main logic with the following entry points:
  * startup -- sync sources, validate, call appliers; if something goes wrong -- exit.
  * reload -- sync sources, validate, call appliers; if something goes wrong -- report to logs and alerts.
* An accessor for the instance configuration: `config:get()`.
* Diagnostics/infomational function: `config:info()`. It should contain alerts if any.

Part of #8724

PR #8833, opened 2023-07-03
config: rework the way the meta is populated

ImeevMA:
This patch reworks the way the meta is populated. This is done to incrementally populate the metadata, instead of set the metadata at the end of reading data from the source. This allows to get the correct meta in cases where getting data from the source failed.

Follow-up #8789

### DIFF
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -80,13 +80,18 @@ function methods._alert(self, alert)
     table.insert(self._alerts, alert)
 end
 
+function methods._meta(self, source_name, key, value)
+    local data = self._metadata[source_name] or {}
+    data[key] = value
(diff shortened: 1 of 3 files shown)


src/box/lua/config/init.lua
```diff
@@ -80,13 +80,18 @@ function methods._alert(self, alert)
     table.insert(self._alerts, alert)
 end
 
+function methods._meta(self, source, key, value)
```
Totktonada, line 83: Nit: If it is a source name, let's name it `source_name`. We use `source` as a source object just below.

Totktonada (approved):

src/box/lua/config/init.lua
ImeevMA: Thanks, fixed!

PR #8834, opened 2023-07-03
Fix flaky test app-luatest/gh_6128_background_mode_test.lua

ligurio:

PR #8836, opened 2023-07-03
[master] test: fix flakiness in gh_6128_background_mode_test (again)

ligurio:
**NOTE**: PR 2.10 - https://github.com/tarantool/tarantool/pull/8834

Previous attempt to fix flakiness in commit 6a2c73f8 ("test: fix flakiness in gh_6128_background_mode_test") used a constant buffer size in check_err_msg function. Tarantool 2.10 has a bit bigger log before a desired message that other versions of Tarantool and it leads to a this resulted in a truncated message ("entering the even" instead of "entering the event loop"). Patch replaces check_err_msg() implementation to grep_log used in luatest, it reads the whole log.

Also patch renames check_err_msg to check_msg, because "entering the event loop" is not an error message.

Follows up #6128

NO_CHANGELOG=fix test
NO_DOC=fix test

### DIFF
--- a/test/app-luatest/gh_6128_background_mode_test.lua
+++ b/test/app-luatest/gh_6128_background_mode_test.lua
@@ -14,16 +14,13 @@ end
 
 -- Check presence of string 'msg' in file 'file'.
 -- Returns true when string is found and false otherwise.
--- Function is not as smart as grep_log in a luatest (luatest/server.lua)
--- and reads the whole log file every time, but this log file has a small
--- size so it is ok.
+-- Function uses grep_log() defined in a luatest (luatest/server.lua).
+-- grep_log is a class method, so it is required to create a Server instance.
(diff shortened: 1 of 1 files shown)


ylobankov (approved):

Lord-KA (approved):
Thanks for the patch! We've discussed in dm, that `luatest.Server` can't be used instead of `popen`, as it automatically initializes `box.cfg{}`, while for the test purposes it should be done from the shell (with `-e`). Besides that, as the dummy server in `check_msg` is not started, we [don't need to kill it](https://github.com/tarantool/luatest#test-helpers), thus it is ok too. Overall, LGTM.

Nit: maybe comment on the issues above in the code.

igormunkin:
Also backported to 2.11 and 2.10 in scope of 9d5cd29d9 and 2a36fa7aa respectively.

PR #8771, opened 2023-06-14
Add static build packages

foxzi:
This patch adds the facility to make DEB and RPM packages with a statically
compiled Tarantool binary inside. The build is performed in a Docker
container, using PackPack docker image (centos-7) and CPack.

The packpack/packpack:centos-7 image has all the necessary dependencies
for building Tarantool and quite old glibc 2.17 which theoretically
provides compatibility of the created packages with any distro where
glibc >= 2.17.

The build can be run with the command below:

    $ VERSION=3.0.0 ./static-build/make_packages.sh

In the `static_build` directory, there will be the following packages:

    tarantool_3.0.0-1_amd64.deb
    tarantool-dev_3.0.0-1_amd64.deb
    tarantool-3.0.0-1.x86_64.rpm
    tarantool-devel-3.0.0-1.x86_64.rpm

`tarantool_3.0.0-1_amd64.deb`, `tarantool-3.0.0-1.x86_64.rpm` are
packages with the Tarantool server binary inside.

`tarantool-dev_3.0.0-1_amd64.deb`, `tarantool-devel-3.0.0-1.x86_64.rpm`
are packages with the Tarantool server development files inside.

PR #8840, opened 2023-07-03
ci/cd: drop distro-specific packaging workflows

ylobankov:
Tarantool 3.0.0 is not going to be distributed as a set of packages for
specific distros. Instead, it is planned to have a few `deb` and `rpm`
packages with a statically compiled Tarantool binary inside, which will
work on almost any distro. Also, we drop building packages with GC32.
Now only GC64 packages will be shipped.

Static build packaging was added at https://github.com/tarantool/tarantool/pull/8771.


changelogs/unreleased/drop-distro-based-packages.md
```diff
@@ -0,0 +1,3 @@
+## feature/build
+
+* Drop building distro-based `deb` and `rpm` packages.
```
p7nov, line 3:
```suggestion
* Dropped building system-specific `deb` and `rpm` Tarantool packages.
```

changelogs/unreleased/gc64-packages-only.md
```diff
@@ -0,0 +1,3 @@
+## feature/build
+
+* Switch to GC64 packages only.
```
p7nov, line 3:
```suggestion
* Dropped building Tarantool packages with GC32. Now only GC64 packages are shipped.
```

p7nov (approved):
Please see my suggestions on the changelog.

changelogs/unreleased/drop-distro-based-packages.md
ylobankov: Fixed.

changelogs/unreleased/gc64-packages-only.md
ylobankov: Fixed.

PR #8841, opened 2023-07-04
Fix teardown in gh_6128_background_mode_test.lua

ligurio:
Test uses a popen module that starts tarantool process in background mode. Tarantool process started in background mode forks a new process and closes a parent, after that popen loses a PID of the started process and `ph:kill()` and `ph:terminate()` doesn't work anymore. It leads to non-terminated tarantool processes after running the test.

Patch fixes that by running `kill` using os.execute with a PID of tarantool process written to a pid file.

Follows up #6128

NO_CHANGELOG=fix test
NO_DOC=fix test


test/app-luatest/gh_6128_background_mode_test.lua
```diff
@@ -53,12 +53,11 @@ pid_file='%s', background=true, work_dir='%s', log='%s',
     t.helpers.retrying({timeout = 2, delay = 0.01}, function(path)
```
Lord-KA, line 53: Nit:
> after that popen losts a PID of the started process

`%s/losts/loses`

Lord-KA (approved):
As all of the previous ones, the patch is pretty straightforward. I don't see anything else to fix here.

ylobankov (approved):

test/app-luatest/gh_6128_background_mode_test.lua
ligurio: Fixed, thanks.

igormunkin:
Also backported to 2.11 and 2.10 in scope of 5602474 and 102d7ad respectively.

PR #8847, opened 2023-07-05
sql: remove one row limit in EXISTS subquery

### LINKED ISSUE #8676: EXISTS fails in case its subquery returns several values
**Bug description**

Executing `SELECT` with `EXISTS` condition operator fails in case subquery returns several values.

* OS: Linux
* OS Version: Ubuntu 20.04.5 LTS
* Architecture: amd64

Tarantool version: 2.10.5-18-g080d07d71

**Steps to reproduce**

1. Run tarantool instance
2. Execute `box.cfg{}`
3. Execute `box.execute([[create table t (id integer primary key);]])`
4. Execute `box.execute([[select * from t where exists (select * from (values (1), (2)));]])` <- on empty table `t` it runs successfully, returning no rows
5. Execute `box.execute([[insert into t values (1);]])`
6. Execute `box.execute([[select * from t where exists (select * from (values (1), (2)));]])` <- on non-empty table `t` it fails with error `'Failed to execute SQL statement: Expression subquery returned more than 1 row'`

**Actual behavior**

SELECT from non-empty table with additional EXISTS (which subquery returns two values) condition fails.

**Expected behavior**

SELECT from non-empty table with additional EXISTS (which subquery returns two values) condition should return all the rows from the table (the result should be the same as if subquery was returning just one value: executing query with `values (1)` instead of `values (1), (2)` produce no errors).

ImeevMA:
According to ANSI, EXISTS is a predicate that tests a given subquery and returns true if it returns more than 0 rows, false otherwise. However, before this patch, EXISTS worked correctly only if there were exactly 0 or 1 rows, and in all other cases it gave an error. This patch makes EXITS work properly.

Closes #8676

@TarantoolBot document
Title: EXISTS predicate

The EXISTS predicate in SQL is a predicate that returns true if the given SQL subquery returns at least one row, and false otherwise.

Example:
```
SELECT EXISTS(SELECT * FROM (VALUES(1), (2)));
```

### DIFF
--- a/changelogs/unreleased/gh-8676-fix-exists-predicate.md
+++ b/changelogs/unreleased/gh-8676-fix-exists-predicate.md
@@ -1,0 +1,4 @@
+## bugfix/sql
+
+* The `EXISTS` predicate no longer requires `LIMIT 1` to work correctly if more
+  than one row is returned in the subselect (gh-8676).
--- a/src/box/sql/expr.c
+++ b/src/box/sql/expr.c
@@ -2774,9 +2774,6 @@ sqlCodeSubselect(Parse * pParse,	/* Parsing context */
 			 *
 			 * If this is an EXISTS, write an integer 0 (not exists) or 1 (exists)
 			 * into a register and return that register number.
-			 *
-			 * In both cases, the query is augmented with "LIMIT 1".  Any
-			 * preexisting limit is discarded in place of the new LIMIT 1.
 			 */
 			Select *pSel;	/* SELECT statement to encode */
 			SelectDest dest;	/* How to deal with SELECT result */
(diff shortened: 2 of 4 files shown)


p7nov (approved):
Minor changelog formatting issue.

changelogs/unreleased/gh-8676-fix-exists-predicate.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/sql
+
+* The `EXIST` predicate now returns true if the subquery returns more than 0
+  rows, and false otherwise (gh-8676).
```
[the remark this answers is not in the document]
ImeevMA: Thank you for the review! Fixed.

src/box/sql/expr.c
```diff
@@ -2801,7 +2801,8 @@ sqlCodeSubselect(Parse * pParse,	/* Parsing context */
 							    &sqlIntTokens[1]);
 				ExprSetProperty(pSel->pLimit, EP_System);
 			}
-			pSel->selFlags |= SF_SingleRow;
+			if (pExpr->op == TK_SELECT)
+				pSel->selFlags |= SF_SingleRow;
```
Gerold103, line 2805: In the comments above I see `the query is augmented with "LIMIT 1".  Any preexisting limit is discarded in place of the new LIMIT 1.` Doesn't it mean the code internally should only return 0 or 1 rows? If it fails on > 1 rows, it looks like the limit got lost somewhere and we actually perform the whole query even when <= 1 rows was enough.

Gerold103 (commented):
Thanks for the patch!

src/box/sql/expr.c
ImeevMA: Thank you for the review! And sorry for such a long answer.

I believe this comment is untrue since https://github.com/tarantool/tarantool/commit/2a720d114dbd604a4f40a1d7b659a4abece2e646. In this commit, the implicit `LIMIT 1` was replaced with an error if more than one row was returned. I believe the fix was for subqueries, and `EXISTS` was changed along the way. I changed EXISTS to work the old way and corrected the comment accordingly. Also, I dropped doc-request since this is just a bug.

p7nov (approved):

ochaplashkin (approved):

test/sql-luatest/gh_8676_fix_exists_test.lua
```diff
@@ -0,0 +1,20 @@
+local server = require('luatest.server')
```
Gerold103, line 1: Please, rename the test file to something like `gh_8676_exists_in_multiselect_test.lua` or something alike. `fix` suffix is redundant - all regression tests are about a "fix".

Gerold103 (approved):
Thanks for the fixes!

test/sql-luatest/gh_8676_fix_exists_test.lua
ImeevMA: Thanks, fixed.

igormunkin (approved):
@ImeevMA, thanks for the patch! LGTM.

PR #8856, opened 2023-07-06
config: fix error message during reload

ImeevMA:
Prior to this patch, if a cluster configuration was not present or an instance was not found in the configuration during a reload, the error would begin with 'Startup failure'. Now it starts with 'Reload failure' in these cases.

Part of https://github.com/tarantool/tarantool/issues/8862

### DIFF
--- a/changelogs/unreleased/gh-8861-cluster-config-error-message.md
+++ b/changelogs/unreleased/gh-8861-cluster-config-error-message.md
@@ -1,0 +1,4 @@
+## bugfix/config
+
+* Fixed an error message if the cluster configuration was not provided or the
+  instance was not found in the cluster configuration during reload (gh-8862).
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -192,8 +192,11 @@ function methods._collect(self, opts)
     end
 
     if next(cconfig) == nil then
+        local is_reload = self._status == 'reload_in_progress'
+        local action = is_reload and 'Reload' or 'Startup'
+        local source_info_str = table.concat(source_info, '\n')
         error(dedent([[
-            Startup failure.
+            %s failure.
 
             No cluster config received from the given configuration sources.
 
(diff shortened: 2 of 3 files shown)


src/box/lua/config/init.lua
```diff
@@ -212,8 +212,11 @@ function methods._collect(self, opts)
     end
 
     if cluster_config:find_instance(cconfig, self._instance_name) == nil then
+        local is_reload = self._status == 'reload_in_progress'
+        local action = is_reload and 'Reload' or 'Startup'
+        local source_info_str = table.concat(source_info, '\n')
         error(dedent([[
-            Startup failure.
+            %s failure.
```
Totktonada, line 219: I guess the same is applicable for the error at line 196.

ImeevMA: Thank you for the review! I did not think that this error can be reproduced, but apparently I was mistaken. Fixed.

test/config-luatest/config_test.lua
```diff
@@ -281,3 +281,97 @@ g.test_config_option = function()
     t.assert_equals(res.exit_code, 0)
     t.assert_equals(res.stdout, table.concat({16, 100000000, 0}, "\n"))
 end
+
+g.test_no_cluster_config_failure = function()
```
Totktonada, line 285: We agreed to try to re-write this test using test.luatest_helpers.server instance to ensure that the instance is operational after failed reload and to better structurize the testing code.

ImeevMA: Done

test/config-luatest/reload_test.lua
```diff
@@ -0,0 +1,80 @@
...
+              roles:
+              - super
+        iproto:
+          listen: unix/:./{{ instance_name }}.iproto
+        groups:
+          group-001:
+            replicasets:
+              replicaset-001:
+                instances:
+                  instance-001: []
```
Totktonada, line 36: Nit: Technically it is not quite correct. This field accepts a record, not an array. It works, because an empty array and an empty map are represented in the same way in Lua.

src/box/lua/config/init.lua
```diff
@@ -192,8 +192,11 @@ function methods._collect(self, opts)
     end
```
Totktonada, line 192: Nit: I think it worth to write a changelog entry, because now we have a baseline: 3.0.0-alpha1.

Totktonada (approved):
Thanks! LGTM.

test/config-luatest/reload_test.lua
ImeevMA: Thanks, fixed

src/box/lua/config/init.lua
ImeevMA: Done

changelogs/unreleased/gh-8861-cluster-config-error-message.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Fixed error message if cluster configuration was not provided or instance was
+  not found in cluster configuration during reload (gh-8862).
```
andreyaksenov, line 4:
```suggestion
* Fixed an error message if the cluster configuration was not provided or the instance was
  not found in the cluster configuration during reload (gh-8862).
```

andreyaksenov (approved):

changelogs/unreleased/gh-8861-cluster-config-error-message.md
Totktonada: Applied. Thanks!

Issue #8033, opened 2022-12-09
Update compat tests to use `server:exec()`


Lord-KA:
Now [tarantool.compat](https://github.com/tarantool/tarantool/issues/7000) tests change compat configuration of the tarantool instance they are running on, same for options tests ([1](https://github.com/tarantool/tarantool/pull/6910), [2](https://github.com/tarantool/tarantool/pull/7250), [3](https://github.com/tarantool/tarantool/pull/7825)). As our tests are being run in multiple threads, these test can be flaky (e.g. option set to 'new' when 'old' is tested). We should run such tests on another tarantool instance to mitigate the problem. Maybe we should even introduce a test tool for compat options, as there will be many of those.

PR #8858, opened 2023-07-06
compat: add is_new and is_old to options

### LINKED ISSUE #8807: Comparing of compat option to old/new works unexpectedly
```
tarantool> require('compat').yaml_pretty_multiline == 'old'
---
- false
...

tarantool> require('compat').yaml_pretty_multiline == 'new'
---
- false
...

tarantool> require('compat').yaml_pretty_multiline == 'default'
---
- false
...

tarantool> require('compat').yaml_pretty_multiline.current == 'old'
---
- false
...

tarantool> require('compat').yaml_pretty_multiline.current == 'new'
---
- false
...

tarantool> require('compat').yaml_pretty_multiline.current == 'default'
---
- true
...
```

That is, the correct way to figure out whether the option is set to 'new' is:

```Lua
if compat.yaml_pretty_multiline.current == 'new' or
        (compat.yaml_pretty_multiline.current == 'default' and
         compat.yaml_pretty_multiline.default == 'new') then
    ...
end
```

This is confusing and inconvenient. It should be possible to compare the compat option value to 'old' and 'new' directly.

Lord-KA:
It used to be somewhat complicated to check the effective value of a compat option, because `<option_name>.current` could contain 'default' state.
This patch introduces helper functions that take care of that.

The following alternatives were considered:
* `compat.<option_name>.effective` - it is excessive in the presence if `current` and `default`, and is visible in serialization
* `compat.<option_name>.get()` - while it is a function, it does only half of the work required, user still has to compare result to 'new'

Closes #8807

@TarantoolBot document
Title: Add `:is_new/old()` helpers to tarantool.compat options

`compat.<option_name>.current` can be 'new', 'old' or 'default', thus when it is default there must be an additional check if `compat.<option_name>.default` is 'new'. It is handier to have a helper to deal with that instead of complicated `if`:
* check if effective value is 'new' before the patch: 
   ```lua
   if compat.<option_name>.current == 'new'
       or (compat.<option_name>.current == 'default' and
             compat.<option_name>.default == 'new') then 
      ...
  end 
   ```
* after the patch: 
   ```lua
   if compat.<option_name>:is_new() then
      ...
   end 
   ```

Please update [tutorial on using compat], maybe add an example to [Listing options details].

[tutorial on using compat]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/compat/compat_tutorial/
[Listing options details]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/compat/compat_tutorial/#listing-options-details

### DIFF
--- a/changelogs/unreleased/gh-8807-compat-is-new-old.md
+++ b/changelogs/unreleased/gh-8807-compat-is-new-old.md
@@ -1,0 +1,4 @@
+## feature/lua
+
+* Added the `:is_new()` and `:is_old()` helpers to `tarantool.compat`
+  options to simplify effective value checks (gh-8807).
--- a/src/lua/compat.lua
+++ b/src/lua/compat.lua
@@ -446,6 +446,28 @@ function compat_mt.__index(_, key)
         result.current = 'old'
     end
 
+    -- Whether the effective value of the option is `new`.
+    function result.is_new(option)
+        if type(option) ~= 'table' then
+            error('usage: compat.<option_name>:is_new()')
+        end
+        if option.current == 'new' then
+            return true
+        end
+        if option.current == 'old' then
+            return false
+        end
+        return option.default == 'new'
+    end
+
+    -- Whether the effective value of the option is `old`.
+    function result.is_old(option)
+        if type(option) ~= 'table' then
+            error(('usage: compat.%s:is_old()'):format(key))
+        end
+        return not option:is_new()
+    end
+
     return result
 end
 
--- a/test/app-luatest/gh_7000_compat_module_test.lua
+++ b/test/app-luatest/gh_7000_compat_module_test.lua
@@ -216,6 +216,26 @@ g.test_help = function()
     t.assert(compat.help())
 end
 
+g.test_is_new_is_old = function()
+    for _, option_def in pairs(definitions) do
+        local name = option_def.name
+        t.assert(compat[name])
+
+        compat[name] = 'new'
+        t.assert(compat[name]:is_new())
+        t.assert_not(compat[name]:is_old())
+
+        compat[name] = 'old'
+        t.assert_not(compat[name]:is_new())
+        t.assert(compat[name]:is_old())
+
+        compat[name] = 'default'
+        local is_new = option_def.default == 'new'
+        t.assert_equals(compat[name]:is_new(), is_new)
+        t.assert_equals(compat[name]:is_old(), not is_new)
+    end
+end
+
 g.test_hot_reload = function()
     local hot_reload_option_def_1 = {
         name = 'hot_reload_option',


test/app-luatest/gh_7000_compat_module_test.lua
```diff
@@ -216,6 +216,26 @@ g.test_help = function()
     t.assert(compat.help())
 end
 
+g.test_is_new_is_old = function()
+    for _, option_def in pairs(definitions) do
+        local name = option_def.name
+        t.assert(compat[name])
+
+        compat[name] = 'new'
```
Totktonada, line 224: I think we shouldn't change compat options in the main testing code, see #8033.

changelogs/unreleased/gh-8807-compat-is-new-old.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Introduced `:is_new()` and `:is_old()` helpers to tarantool.compat options
+  for easier effective value checks (gh-8807).
```
andreyaksenov, line 4:
```suggestion
* Added the `:is_new()` and `:is_old()` helpers to `compat` options
  to simplify effective value checks (gh-8807).
```

andreyaksenov (approved):

test/app-luatest/gh_7000_compat_module_test.lua
Lord-KA: Similar to `test_new_index()`, I switch only dummy options defined in `definitions`, so it is safe.

Totktonada (approved):
No objections from my side. LGTM.

src/lua/compat.lua
```diff
@@ -446,6 +446,28 @@ function compat_mt.__index(_, key)
...
+        if option.current == 'old' then
+            return false
+        end
+        return option.default == 'new'
+    end
+
+    -- Whether the effective value of the option is `old`.
+    function result.is_old(option)
+        if type(option) ~= 'table' then
+            error('usage: compat.<option_name>:is_old()')
```
ligurio, line 466: Nit: we could pass an option name to `is_old` and use option name in a message `error('usage: yaml_pretty_multiline:is_old()')`. I suppose it is more user-friendly.

I'm not insist, feel free to ignore.

ligurio (approved):
LGTM, however, see a minor comment.

src/lua/compat.lua
Lord-KA: I really like the idea! Applied.
```diff
-            error('usage: compat.<option_name>:is_old()')
+            error(('usage: compat.%s:is_old()'):format(key))
```

igormunkin:
Also backported to 2.11 in scope of 380956f.

src/lua/compat.lua
igormunkin: \*Sigh\*

This spot is fixed, but another one in `is_new` is not. Glad, this is not critical thing in this PR.

@Lord-KA, please next time **resolve** all the discussions with their owners and **do not assign** me, until all of them are resolved.

@ligurio, I guess, it's worth mentioning all the spots either via commenting the particular lines or by mentioning everything in the one comment.

PR #8860, opened 2023-07-06
config: introduce remaining vinyl options

ImeevMA:
This patch introduces all remaining vinyl options that have not been introduced before.

Part of https://github.com/tarantool/tarantool/issues/8861


Totktonada (approved):
Thanks you for the patch! I see no problems. LGTM.

Issue #8862, opened 2023-07-06
config: perform feedback inspired fixes


Totktonada:
We've implemented startup/reload flow based on a declarative configuration in the scope of #8724 (released as 3.0.0-alpha1). It is, of course, just first demonstrational version and we're working on a good release-to-market version.

We've found some problems during development, testing, demonstrations and discusions within the team.

I'll add such (mostly small/minor) feedback here to solve it before a release candidate. The task list below is to be updated (I have ~20 different notes to process.)

Larger task are to be filed separately.

```[tasklist]
- [x] `config:reload()` may report an error that contains 'Startup failure' message (https://github.com/tarantool/tarantool/pull/8856)
- [x] A startup failure message contains an obsolete configuration example (`database.rw` instead of `database.mode`) (https://github.com/tarantool/tarantool/pull/8918)
- [x] test: enable parallelization for the `config-luatest` test suite (https://github.com/tarantool/tarantool/pull/8863)
- [x] Handle `TT_*_DEFAULT` environment variables, which don't rewrite file/etcd configuration sources (https://github.com/tarantool/tarantool/pull/8917)
- [x] Make startup configuration available with config:get() from application script (https://github.com/tarantool/tarantool/pull/8929)
- [x] List environment variables interpreted by tarantool (https://github.com/tarantool/tarantool/pull/8933)
- [x] Don't persist group_name as cluster_name (https://github.com/tarantool/tarantool/pull/9004)
- [x] Allow to extend Tarantool Community Edition config functionality with user-defined logic using a non-public API (https://github.com/tarantool/tarantool/pull/8981)
- [x] Support nested directories in wal.dir, snapshot.dir and so on (https://github.com/tarantool/tarantool/pull/9005)
- [x] Change default directories to layout similar to `tt`'s (https://github.com/tarantool/tarantool/pull/9009)
- [x] Create parent directories for process.work_dir, console.socket, log.file (https://github.com/tarantool/tarantool/pull/9009)
- [x] Calculate directory paths relative to process.work_dir (https://github.com/tarantool/tarantool/pull/9009)
- [x] Advertise `tt` as the tool for starting tarantool instance in `tarantool --help` (https://github.com/tarantool/tarantool/pull/9288)
- [x] Validate instance/replicaset/group names, extracted from #9148 (https://github.com/tarantool/tarantool/pull/9292)
[...]

PR #8863, opened 2023-07-07
test: run tests in config test suite in parallel

Totktonada:
All the tests are independent and nothing prevents running them in parallel. The option to enable the parallel running was just forgotten.

Part of #8862


Lord-KA (approved):
Thanks for the patch! The change is trivial, LGTM.

Issue #8659, opened 2023-05-15
Provide old triggers API with name


drewdzzz:
After the universal trigger registry will be implemented, the triggers that will be moved there will be nameable. Let's add trigger name to old API for all trigger - it will allow to take advantage of nameable triggers for ones that won't be moved to the registry and for users who doesn't want to rewrite trigger setting code and uses old triggers API. I suggest the following APIs:
1. `space:on_replace([new_trigger, [old_trigger, [trigger_name]]])` - we will add a name as a third argument - users with lots of code will need just to add the 3rd arguments to all trigger setters. If the name is not passed, the behavior isn't changed. If `trigger_name` is passed, `old_trigger` argument is ignored, and `new_trigger` is set by `trigger_name` (or deleted, if it is `nil`).
2. Old API is a bit ugly and will be deprecated later, so we introduce another API - `space:on_replace{name = trigger_name, func = handler_or_nil}` - this API will be introduce for all triggers in tarantool and will not be deprecated.

PR #8869, opened 2023-07-11
box: move session triggers to trigger registry

drewdzzz:
The patch populates submodule session with an event for each public
trigger - they will be used for user-defined triggers. Old triggers are
not removed beacause they are used for internal purposes - audit from EE
uses them, for instance. Also, the commit introduces a test for all
triggers that use old API. It checks only session triggers now - all the
other triggers will be included there after they will be moved to the
trigger registry.

In order to have an opportunity to move user-defined triggers to trigger
registry, the commit introduces helper luaT_event_reset_trigger that
supports two different ways to set triggers - key-value API and positional
API. The first one is a new API that allows to set or delete trigger only
by name. The second API is implemented for the sake of backward
compatibility - it mimics behavior of function lbox_trigger_reset, which
was used to modify triggers, but sets the triggers to event from trigger
registry instead of separate trigger list. Also, positional API allows
to pass a name as the third argument - in this case the function will
set or delete trigger by passed name - the first argument is a new
trigger (or nil, if the function is called to delete a trigger), the
second argument will be ignored in this case.

Part of https://github.com/tarantool/tarantool/issues/8657

### DIFF
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -40,6 +40,7 @@
 #include <base64.h>
 
 #include "version.h"
+#include "event.h"
 #include "fiber.h"
 #include "fiber_cond.h"
 #include "cbus.h"
@@ -1777,10 +1778,8 @@ tx_process_disconnect(struct cmsg *m)
 		 * closed, its push() method is replaced with a stub.
 		 */
 		con->tx.is_push_pending = false;
-		if (! rlist_empty(&session_on_disconnect)) {
-			tx_fiber_init(con->session, 0);
-			session_run_on_disconnect_triggers(con->session);
-		}
+		tx_fiber_init(con->session, 0);
+		session_run_on_disconnect_triggers(con->session);
 	}
 }
 
@@ -2725,8 +2724,7 @@ tx_process_connect(struct cmsg *m)
 	greeting_encode(greeting, tarantool_version_id(), &uuid,
 			con->salt, IPROTO_SALT_SIZE);
 	xobuf_dup(out, greeting, IPROTO_GREETING_SIZE);
-	if (!rlist_empty(&session_on_connect) &&
-	    session_run_on_connect_triggers(con->session) != 0)
+	if (session_run_on_connect_triggers(con->session) != 0)
 		goto error;
 	iproto_wpos_create(&msg->wpos, out);
 	return;
--- a/src/box/lua/trigger.c
+++ b/src/box/lua/trigger.c
@@ -9,6 +9,7 @@
 #include <diag.h>
 #include <fiber.h>
 #include "lua/utils.h"
+#include "tt_static.h"
 
 /**
  * Sets a trigger with passed name to the passed event.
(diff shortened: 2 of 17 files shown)


alyapunov (changes requested):

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
 	luaL_register_type(L, event_trigger_iterator_typename,
 			   trigger_iterator_methods);
 }
+
```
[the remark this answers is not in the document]
drewdzzz: Fixed all the tips discussed f2f.

ochaplashkin (approved):
LGTM :rocket:

alyapunov (approved):

locker:
Please resolve the issue with the audit EE subsystem before merging this PR.

src/box/lua/trigger.h
```diff
@@ -9,12 +9,55 @@
...
 
+/**
+ * Creates a Lua trigger, replaces an existing one, or deletes a trigger.
+ *
+ * The function accepts a Lua stack. Values starting from index bottom are
+ * considered as the function arguments.
+ *
+ * The function supports two API versions.
+ *
+ * The first version - key-value arguments. The function is called with one Lua
```
locker, line 28: I don't understand why we need to extend the legacy API. AFAIU we're planning to drop it sooner or later so we need this solely as a backward compatibility. Also, I think that we should print a warning on an attempt to use it. I'll talk to PM to clarify these points.

drewdzzz: #8659 is about it. We decided that there are users who wants to use named triggers, but does not want to completely rewrite lots of code that sets triggers - they'll just need to add a third argument. And key-value API won't be deprecated.

What's about deprecation message, I was going to add it as a separate patch after we will deal with all triggers (`net.box`, transactional ones, etc.).

src/box/iproto.cc
```diff
@@ -1777,7 +1778,8 @@ tx_process_disconnect(struct cmsg *m)
 		 * closed, its push() method is replaced with a stub.
 		 */
 		con->tx.is_push_pending = false;
-		if (! rlist_empty(&session_on_disconnect)) {
+		if (event_has_triggers(session_on_disconnect_event) ||
+		    !rlist_empty(&session_on_disconnect)) {
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/schema.h
```diff
@@ -138,10 +138,15 @@ sequence_cache_delete(uint32_t id);
...
- * Triggers fired after access denied error is created.
+ * Internal triggers fired after access denied error is created.
  */
 extern struct rlist on_access_denied;
 
+/**
+ * User-definded triggers fired after access denied error is created.
+ */
+extern struct event *on_access_denied_event;
+
```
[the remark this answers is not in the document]
drewdzzz: Moved them to `box/error.h` along with `on_access_denied_ctx`.

src/box/session.c
```diff
@@ -321,8 +327,13 @@ session_remove_stmt_id(struct session *session, uint32_t stmt_id)
...
 struct credentials admin_credentials;
 
+/**
+ * Runs on_connect or on_disconnect triggers. Argument triggers is for internal
+ * triggers and argument event is for user-defined ones.
+ */
 static int
-session_run_triggers(struct session *session, struct rlist *triggers)
+session_run_connect_triggers(struct session *session, struct rlist *triggers,
+			     struct event *event)
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
...
+ * Checks positional arguments for luaT_event_reset_trigger.
+ * Throws an error if the format is not suitable.
+ */
+static void
+luaT_event_reset_trigger_check_positional_input(struct lua_State *L, int bottom)
+{
+	assert(lua_checkstack(L, bottom + 2));
+	/* Push optional arguments. */
+	while (lua_gettop(L) < bottom + 2)
+		lua_pushnil(L);
```
locker, line 269: `lua_settop(L, bottom + 2)`

And `lua_checkstack` doesn't seem to be necessary.

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
...
+	const int top = bottom + 2;
+	if (!lua_isnil(L, top) && !luaL_isnull(L, top))
+		return luaT_event_reset_trigger_by_name(L, event, top, bottom);
+	/*
+	 * Name is not passed - old API support.
+	 * 1. If triggers are not passed, return table of triggers.
+	 * 2. If new_trigger is passed and old_trigger is not - set
+	 * new_trigger using string representation of function as name.
+	 * 3. If old_trigger is passed and new_trigger is not - delete
+	 * trigger with by function string representation as a name.
```
locker, line 338: Nit: "with" is redundant.

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
...
+	 * 4. If both triggers are provided - replace old trigger with
+	 * new one if they have the same name, delete old trigger and
+	 * insert new one at the beginning of the trigger list otherwise.
```
locker, line 341: It's not the same behavior as the legacy one. AFAIU, if both old and new functions are specified, the old function should be replaced with the new one in the trigger list.

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
...
+		}
+		event_trigger_iterator_destroy(&it);
+		return 1;
+	}
+	int ret_count = 0;
+	const char *old_name = NULL;
+	const char *new_name = NULL;
+	struct func_adapter *new_trg = NULL;
+	if (luaL_iscallable(L, bottom + 1)) {
+		old_name = luaT_tolstring(L, bottom + 1, NULL);
```
locker, line 363: It's not entirely correct to use `tostring` on a callable object because it may have the `__tostring` metamethod overwritten, in which case it'd return the same value even for different objects. You should use `lua_topointer` here.

src/box/lua/trigger.c
```diff
@@ -253,3 +253,146 @@ box_lua_trigger_init(struct lua_State *L)
...
+	 * anything else is error.
+	 */
+	bool ok = true;
+	/* Name must be a string if it is passed. */
+	ok = ok && (lua_isnil(L, bottom + 2) || luaL_isnull(L, bottom + 2) ||
+		    lua_type(L, bottom + 2) == LUA_TSTRING);
+	ok = ok && (lua_isnil(L, bottom + 1) || luaL_isnull(L, bottom + 1) ||
+		    luaL_iscallable(L, bottom + 1));
+	ok = ok && (lua_isnil(L, bottom) || luaL_isnull(L, bottom) ||
+		    luaL_iscallable(L, bottom));
```
locker, line 284: AFAIU the old API only supports plain functions. You don't need to extend it to support callback objects.

src/box/lua/trigger.h
locker: Talked to @unera. He said that we *will* drop the old API completely in 4.0, including the extension introduced in this PR. He also said that he only asked to add the third argument to `trigger.reset` and not the table argument (point 2 in #8659).

Talked to @Totktonada. He said that he doesn't remember discussion of the table argument to `trigger.reset` (point 2 in #8659). Since we're planning to drop it, he says he wouldn't add it.

Something seems to be broken in the feature design process... Since we already implemented it, let's leave it as is.

src/box/lua/trigger.c
drewdzzz: The event API supports only set/delete by name, so we would have to extend the API to be able to replace a trigger with another name. And, since we are going to deprecate this interface, we decided that it's not worth it.

src/box/lua/trigger.c
drewdzzz: Use `lua_settop` now, removed `lua_checkstack`.

src/box/lua/trigger.c
drewdzzz: Fixed

src/box/lua/trigger.c
drewdzzz: Good point, thanks. Use handler address now.

src/box/lua/trigger.c
drewdzzz: Mentioned in commit message that the new helper support any callable object, linked the commit to the issue.

locker (approved):

locker:
The fuzzing CI workflow fails. PTAL.

drewdzzz:
Sql buzzer uses box module, but event was not initialized there. Did it in a separate commit.

Issue #8825, opened 2023-06-29
Backporting LuaJIT upstream commits part 5


Buristan:
This ticket is an umbrella for all commits we planning to back-port in the next Q.
This list includes commits excluded from #8516 and may be updated later with some new.

general + x86/x64:
* [x] [93a65d3c](https://github.com/LuaJIT/LuaJIT/commit/93a65d3c) Fix BC_UCLO insertion for returns.
* [x] [56c04acc](https://github.com/LuaJIT/LuaJIT/commit/56c04acc) FFI: Fix recording of union initialization.
* [x] [377a8488](https://github.com/LuaJIT/LuaJIT/commit/377a8488) Fix binary number literal parsing.
* [x] [c50232eb](https://github.com/LuaJIT/LuaJIT/commit/c50232eb) Always exit after machine code page protection change fails.
* [x] [c7db8255](https://github.com/LuaJIT/LuaJIT/commit/c7db8255) Fix TDUP load forwarding after table rehash.
* [x] [9f452bbe](https://github.com/LuaJIT/LuaJIT/commit/9f452bbe) Fix handling of instable types in TNEW/TDUP load forwarding.
* [x] [2d8300c1](https://github.com/LuaJIT/LuaJIT/commit/2d8300c1) Fix frame for on-trace out-of-memory error.
* [x] [d5bbf9cd](https://github.com/LuaJIT/LuaJIT/commit/d5bbf9cd) Fix frame for more types of on-trace error messages.
* [x] [0cc5fdfb](https://github.com/LuaJIT/LuaJIT/commit/0cc5fdfb) Fix maxslots when recording BC_TSETM.
* [x] [0ef51b49](https://github.com/LuaJIT/LuaJIT/commit/0ef51b49) Handle table unsinking in the presence of IRFL_TAB_NOMM.
* [x] [94ada596](https://github.com/LuaJIT/LuaJIT/commit/94ada596) Fix maxslots when recording BC_VARG.
* [x] [a01cba9d](https://github.com/LuaJIT/LuaJIT/commit/a01cba9d) Fix maxslots when recording BC_VARG, part 2.
* [x] [abb27c77](https://github.com/LuaJIT/LuaJIT/commit/abb27c77) Fix maxslots when recording BC_VARG, part 3.
* [x] [caf7cbc5](https://github.com/LuaJIT/LuaJIT/commit/caf7cbc5) Fix predict_next() in parser.
[...]

PR #8871, opened 2023-07-12
LuaJIT: bump new version

fckxorg:
Mark CONV as non-weak, to prevent elimination of its side-effect.

Part of #8825

NO_DOC=LuaJIT bump
NO_TEST=LuaJIT bump


igormunkin:
The changeset is merged to master in scope of #9388, backported to 2.11 and 2.10 in scope of #9389 and #9390 respectively.

Issue #4263, opened 2019-05-31
DDL for temporary and local spaces should not be replicated


R-omk:
Tarantool version:
`2.2.0-343-gd07e3ac2f`

When I perform truncate on  _temporary_ (and _is_local_) spaces, it is performed on all replicas.

```
engine = 'memtx',
is_local = true,
temporary = true,
```

GeorgyKirichenko:
As truncate is used as fast shortcut for "table delete all" it should not be replicated across a cluster. Any other operation such as create/modify/drop should continue their working as before.

PR #8879, opened 2023-07-14
Allow to truncate temporary and local spaces in the read-only mode

### LINKED ISSUE #5616: Cannot truncate a temporary or local space on a ro instance
When I have a ro instance, I may freely insert data to temporary and local spaces:
```lua
tarantool> box.info.ro
---
- true
...


tarantool> box.space.loc:insert{1}
---
- [1]
...

tarantool> box.space.temp:insert{1}
---
- [1]
...
```
However, If I try to truncate any of the spaces, I get `ER_READONLY` error.
```lua
tarantool> box.space.loc:truncate{}
2020-12-10 16:57:17.361 [50808] main/103/interactive box.cc:179 E> ER_READONLY: Can't modify data because this instance is in read-only mode.
---
- error: Can't modify data because this instance is in read-only mode.
...

tarantool> box.space.temp:truncate()
2020-12-10 16:57:23.343 [50808] main/103/interactive box.cc:179 E> ER_READONLY: Can't modify data because this instance is in read-only mode.
---
- error: Can't modify data because this instance is in read-only mode.
...
```

locker:
DML operations on temporary and local spaces can be performed even if the instance is in the read-only mode, but DDL operations (such as `alter`) are forbidden in this case[^1]. Technically, `truncate` is a DDL operation so initially it was forbidden as well. However, it should be safe to perform this operation on a temporary or local space because logically it only modifies the data stored in the space (like DML) and it isn't replicated (see #4263). So starting from Tarantool 2.11.1 we allow users to truncate temporary spaces in the read-only mode.

To achieve that, we bypass the read-only check for the `_truncate` system space in `box_process1()` and perform it in the `on_replace` system trigger instead, when we know which space is truncated.

Closes #5616

[^1]: https://www.tarantool.io/en/doc/latest/concepts/replication/repl_architecture/#replication-local

### DIFF
--- a/changelogs/unreleased/gh-5616-temp-space-truncate-ro.md
+++ b/changelogs/unreleased/gh-5616-temp-space-truncate-ro.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Allowed truncation of temporary and local spaces in the read-only mode
+  (gh-5616).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -2624,6 +2624,10 @@ on_replace_dd_truncate(struct trigger * /* trigger */, void *event)
 		/* Space drop - nothing to do. */
 		return 0;
 	}
+	if (recovery_state == INITIAL_RECOVERY) {
+		/* Space creation during initial recovery - nothing to do. */
+		return 0;
+	}
 
 	uint32_t space_id;
 	if (tuple_field_u32(new_tuple, BOX_TRUNCATE_FIELD_SPACE_ID, &space_id) != 0)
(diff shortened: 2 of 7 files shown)


andreyaksenov (approved):

kostja (approved):

alyapunov (approved):

locker:
Cherry-picked to 2.11.

a1div0:
For old version lua code fix:
``` lua
local function local_space_truncate(space)
    local qty = 0
    while true do
        local tuple_list = space:select(nil, {fullscan=true, limit=1})
        if #tuple_list == 0 then
            break
        end
        space:delete(tuple_list[1].id)
        qty = qty + 1
    end
    return qty
end
```

It works

PR #8882, opened 2023-07-14
datetime: allow boundary values for interval.new

### LINKED ISSUE #8878: Inconsistent interval boudaries check
`datetime.interval` allowed range should include boundaries, yet it doesn't.

```lua
tarantool> datetime.interval.new{year=11759221}
---
- error: 'builtin/box/console.lua:415: value 11759221 of year is out of allowed range
    [-11759221, 11759221]'
...
```

Square brackets imply boundaries inclusion, yet the boundaries are excluded.

https://github.com/tarantool/tarantool/blob/ff57f990f359f6d7866c1947174d8ba0e97b1ea6/src/lua/datetime.lua#L297-L301

At the same time, datetime range check

https://github.com/tarantool/tarantool/blob/ff57f990f359f6d7866c1947174d8ba0e97b1ea6/src/lua/datetime.lua#L263

and C interval range check

https://github.com/tarantool/tarantool/blob/ff57f990f359f6d7866c1947174d8ba0e97b1ea6/src/lib/core/datetime.c#L695

include boundaries, as expected.

DifferentialOrange:
Before this patch, one couldn't create new datetime interval with boundary value from Lua. At the same time, it was possible to create such interval from Lua through addition and subtraction. C range verification allow to create boundary value intervals, error message also implies that they should be allowed. (See #8878 for more info.)

Closes #8878

NO_DOC=small bug fix

### DIFF
--- a/changelogs/unreleased/gh-8878-interval-boundaries-fix.md
+++ b/changelogs/unreleased/gh-8878-interval-boundaries-fix.md
@@ -1,0 +1,4 @@
+## bugfix/datetime
+
+* Fixed a bug raising a false positive error when creating new intervals with
+  range boundary values (gh-8878).
--- a/test/app-tap/datetime.test.lua
+++ b/test/app-tap/datetime.test.lua
@@ -4,9 +4,12 @@ local tap = require('tap')
 local test = tap.test('errno')
 local date = require('datetime')
 local ffi = require('ffi')
+local json = require('json')
 local TZ = date.TZ
 
 test:plan(39)
+
+local INT_MAX = 2147483647
 
 -- minimum supported date - -5879610-06-22
 local MIN_DATE_YEAR = -5879610
@@ -23,6 +26,9 @@ local MAX_DAY_RANGE = MAX_YEAR_RANGE * AVERAGE_DAYS_YEAR
 local MAX_HOUR_RANGE = MAX_DAY_RANGE * 24
 local MAX_MIN_RANGE = MAX_HOUR_RANGE * 60
 local MAX_SEC_RANGE = MAX_DAY_RANGE * SECS_PER_DAY
+local MAX_NSEC_RANGE = INT_MAX
+local MAX_USEC_RANGE = math.floor(MAX_NSEC_RANGE / 1e3)
+local MAX_MSEC_RANGE = math.floor(MAX_NSEC_RANGE / 1e6)
 
 local incompat_types = 'incompatible types for datetime comparison'
 local only_integer_ts = 'only integer values allowed in timestamp'..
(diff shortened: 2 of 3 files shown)


andreyaksenov (approved):

changelogs/unreleased/gh-8878-interval-boundaries-fix.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/datetime
+
+* Fixed a bug (raising false positive error) when creating new intervals with
+  range boundary values (gh-8878).
```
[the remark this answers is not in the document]
DifferentialOrange: Fixed, thanks

test/app-tap/datetime.test.lua
```diff
@@ -1316,6 +1319,29 @@ test:test("Time intervals creation - range checks", function(test)
...
+        {{ min = -MAX_MIN_RANGE }, '-6184879877160 minutes'},
+        {{ min =  MAX_MIN_RANGE }, '+6184879877160 minutes'},
+        {{ sec = -MAX_SEC_RANGE }, '-371092792629600 seconds'},
+        {{ sec =  MAX_SEC_RANGE }, '+371092792629600 seconds'},
+        {{ nsec = -MAX_NSEC_RANGE }, '+0 seconds, -2147483647 nanoseconds'},
+        {{ nsec =  MAX_NSEC_RANGE }, '+0 seconds, 2147483647 nanoseconds'},
+    }
+    for _, row in pairs(boundary_values) do
+        local args, str = unpack(row)
+        test:is(tostring(inew(args)), str)
```
Lord-KA, line 1343: Nit:
`%s/str/expected`
`%s/args/value`
The current phrasing is a bit less intuitive, but feel free to ignore.

Lord-KA: To check that this is actually a boundary value, we can expect error for `+-1` values (with `assert_raises()`). I think it would be nice.

Lord-KA (approved):
@DifferentialOrange, thanks for the patch!
Overall, I don't see any major issues, so LGTM, but see two comments below.

test/app-tap/datetime.test.lua
DifferentialOrange: Thanks, fixed both

test/app-tap/datetime.test.lua
```diff
@@ -1316,6 +1319,54 @@ test:test("Time intervals creation - range checks", function(test)
...
+        {{ day = -math.floor(MAX_DAY_RANGE) }, '-4295055470 days'},
+        {{ day =  math.floor(MAX_DAY_RANGE) }, '+4295055470 days'},
+        {{ hour = -MAX_HOUR_RANGE }, '-103081331286 hours'},
+        {{ hour =  MAX_HOUR_RANGE }, '+103081331286 hours'},
+        {{ min = -MAX_MIN_RANGE }, '-6184879877160 minutes'},
+        {{ min =  MAX_MIN_RANGE }, '+6184879877160 minutes'},
+        {{ sec = -MAX_SEC_RANGE }, '-371092792629600 seconds'},
+        {{ sec =  MAX_SEC_RANGE }, '+371092792629600 seconds'},
+        {{ nsec = -MAX_NSEC_RANGE }, '+0 seconds, -2147483647 nanoseconds'},
+        {{ nsec =  MAX_NSEC_RANGE }, '+0 seconds, 2147483647 nanoseconds'},
```
ligurio, line 1339: Boundary values in messages are calculated manually and hardcoded now. I would build these message in runtime:
```
'-11759221 years' -> ('-%d'):format(MAX_YEAR_RANGE)
'+11759221 years' -> ('%d'):format(MAX_YEAR_RANGE) 
...
```
In latter case these messages will depend on real boundary values and automatically will be updated when we will update boundary values.

ligurio (approved):
I've no objections, so LGTM.
However, I left a minor comment, please take a look.

test/app-tap/datetime.test.lua
Lord-KA: Actually, same goes for +-1 test, it seems quite easy to automate and make it more compact as well.

DifferentialOrange: Reworked tests to be the most compact and automized

test/app-tap/datetime.test.lua
```diff
@@ -1316,6 +1351,27 @@ test:test("Time intervals creation - range checks", function(test)
...
+
+    local range_boundary = {
+        year = MAX_YEAR_RANGE,
+        month = MAX_MONTH_RANGE,
+        week = MAX_WEEK_RANGE,
+        day =  MAX_DAY_RANGE,
+        hour = MAX_HOUR_RANGE,
+        min = MAX_MIN_RANGE,
+        sec = MAX_SEC_RANGE,
+        nsec = MAX_NSEC_RANGE,
```
ligurio, line 1363: Missed `MAX_USEC_RANGE` and `MAX_MSEC_RANGE`.
Both could be used in intervals:

```
tarantool>  datetime.interval.new{msec=2147}
---
- error: 'builtin/box/console.lua:415: value 2147 of msec is out of allowed range
    [-2147, 2147]'
...

tarantool>  datetime.interval.new{usec=2147}
---
- +0 seconds, 2147000 nanoseconds
...

tarantool> 
```

test/app-tap/datetime.test.lua
ligurio: Thanks!

test/app-tap/datetime.test.lua
DifferentialOrange: Added

igormunkin:
Also backported to 2.11 and 2.10 in scope of 230aba5 and 3af2b56 respectively.

PR #8888, opened 2023-07-17
msgpack: fix decoding intervals with int64

### LINKED ISSUE #8887: Fail to decode interval with value bigger than int32
It is allowed to create intervals with values bigger than int32 for minutes and seconds.

https://github.com/tarantool/tarantool/blob/ff57f990f359f6d7866c1947174d8ba0e97b1ea6/src/lua/datetime.lua#L144-L145

Since `double`s are used to store values, interval itself can process such big values.

https://github.com/tarantool/tarantool/blob/054526ac8472d005ae7468774f06290597278a1d/src/lib/core/datetime.h#L109-L128

Msgpack encoding also succeeds:

```lua
tarantool> msgpack.encode(datetime.interval.new{min=6184879877159}):hex()
---
- c70d060205cf000005a007916c270801
...
```

Yet msgpack decoding fails:

```lua
tarantool> msgpack.decode(msgpack.encode(datetime.interval.new{min=6184879877159}))
---
- error: 'msgpack.decode: invalid MsgPack'
...
```

The reason is that int32 is used to extract the value

https://github.com/tarantool/tarantool/blob/054526ac8472d005ae7468774f06290597278a1d/src/lib/core/mp_interval.c#L130

while allowed values are not really restricted to int32. (For example, msgpack above has uint64 value for minutes.)

DifferentialOrange:
It is possible for interval to have days, hours, minutes and seconds bigger than INT_MAX (or less than INT_MIN). Before this patch, msgpack decoding had failed to parse intervals with msgpack int64 and uint64. int64_t should be enough to store any value allowed for datetime intervals.

Closes #8887

NO_DOC=small bug fix

### DIFF
--- a/test/unit/interval.c
+++ b/test/unit/interval.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include <assert.h>
+#include <limits.h>
 
 #include "string.h"
 #include "datetime.h"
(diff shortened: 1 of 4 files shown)


changelogs/unreleased/gh-8887-fix-interval-int64
```diff
@@ -0,0 +1,4 @@
+## bugfix/msgpack
+
+* Fixed decoding datetime intervals with fields bigger than possible int32
+  values (gh-8887).
```
andreyaksenov, line 4:
```suggestion
* Fixed decoding datetime intervals with fields larger than possible int32
  values (gh-8887).
```

andreyaksenov (approved):

changelogs/unreleased/gh-8887-fix-interval-int64
DifferentialOrange: Fixed

test/unit/interval.c
```diff
@@ -96,11 +96,62 @@ test_interval_encode_decode(void)
...
+	interval_mp_recode(&itv, &result);
+	ok(is_interval_equal(&itv, &result), "Intervals are equal.");
+
+	itv.nsec = -2147483647;
+	interval_mp_recode(&itv, &result);
+	ok(is_interval_equal(&itv, &result), "Intervals are equal.");
+
+	itv.nsec = 2147483647;
+	interval_mp_recode(&itv, &result);
+	ok(is_interval_equal(&itv, &result), "Intervals are equal.");
```
ligurio, line 146: I would not use magic number and use constants defined in `limits.h` [^1][^2] instead.

[^1]: https://en.cppreference.com/w/c/types/limits
[^2]: https://learn.microsoft.com/ru-ru/cpp/c-language/cpp-integer-limits

test/unit/interval.c
```diff
@@ -96,11 +96,62 @@ test_interval_encode_decode(void)
 	is(result.adjust, DT_EXCESS, "Adjust value is right");
 }
 
+static void
+test_interval_encode_decode_limits(void)
```
ligurio, line 100: Tests for msgpack function is good, but users will work with intervals in Lua, so I would add a test for `msgpack.encode`/`msgpack.decode` for interval objects with max numbers.

ligurio (approved):
Thanks for the patch, Georgy! See my minor comments related to testing.

test/unit/interval.c
DifferentialOrange: Updated to test on INT_MAX + 1, INT_MIN - 1, where allowed.

test/unit/interval.c
DifferentialOrange: Added pure Lua msgpack encode/decode test cases

test/unit/interval.c
ligurio: Thanks!

test/unit/interval.c
ligurio: Thanks!

ligurio (approved):

alyapunov (approved):

igormunkin:
Also backported to 2.11 and 2.10 in scope of f98b977 and 26acd38 respectively.

PR #8891, opened 2023-07-17
test: reset readline configuration for justrun too

nshy:
This fixes gh_8613_new_cli_behaviour_test run with my custom .inputrc
(I use vi-cmd-mode-string/vi-ins-mode-string).

We already reset readline configuration in interactive_tarantool.lua.

Follows up ground works done for #7774

### DIFF
--- a/test/justrun.lua
+++ b/test/justrun.lua
@@ -72,6 +72,10 @@ function justrun.tarantool(dir, env, args, opts)
     local opts = opts or {}
     assert(type(opts) == 'table')
 
+    -- Prevent system/user inputrc configuration file from
+    -- influencing testing code.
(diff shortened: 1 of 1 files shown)


test/justrun.lua
```diff
@@ -72,6 +72,7 @@ function justrun.tarantool(dir, env, args, opts)
     local opts = opts or {}
     assert(type(opts) == 'table')
 
+    env['INPUTRC'] = '/dev/null'
```
Totktonada, line 75: Nit: I would move it into its own code section (with an empty line before and after) and leave some explanatory comment.

nshy: Done.

locker (approved):

Issue #8861, opened 2023-07-06
config: add remaining box.cfg options


Totktonada:
We've created an initial configuration schema in the scope of #8724 (released as 3.0.0-alpha1).

However, there are configuration option that are supported on the low level (by `box.cfg()`), but not present in the declarative configuration yet.

Let's add the following sections and options into the configuration.

```[tasklist]
- [ ] audit (EE)
- [x] security (EE)
- [x] feedback
- [x] flightrec (EE)
- [x] metrics
- [x] vinyl
- [x] `box.cfg.bootstrap_leader` into the `replication` section
- [x] `box.cfg.memtx_sort_threads` into the `memtx` section
- [x] Add a test that verifies that no `box.cfg()` options are missed
```

The sections marked as EE are to be accepted only on Tarantool Enterprise Edition.

PR #8895, opened 2023-07-19
config: introduce feedback options

ImeevMA:
This patch introduces all feedback options.

Part of #8861

### DIFF
--- a/changelogs/unreleased/gh-8861-feedback-options.md
+++ b/changelogs/unreleased/gh-8861-feedback-options.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* All feedback options are now supported (gh-8861).
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -232,6 +232,17 @@ local function advertise_peer_uri_validate(data, w)
     return uri
 end
 
+local function feedback_apply_default_if(_data, _w)
+    return box.internal.feedback_daemon ~= nil
+end
+
+local function feedback_validate(data, w)
+    if data == nil or box.internal.feedback_daemon ~= nil then
+        return
+    end
+    w.error('Tarantool is built without feedback reports sending support')
+end
+
 return schema.new('instance_config', schema.record({
     config = schema.record({
         version = schema.enum({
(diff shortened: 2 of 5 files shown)


andreyaksenov (approved):

src/box/lua/config/instance_config.lua
```diff
@@ -232,6 +232,13 @@ local function advertise_peer_uri_validate(data, w)
     return uri
 end
 
+local function feedback_options(data)
+    if box.internal.feedback_daemon == nil then
+        return nil
+    end
+    return data
+end
```
Totktonada, line 240: I think that the schema shouldn't be different on different tarantool build types. How about handling options that aren't supported by the given build in the `validate` annotation?

changelogs/unreleased/gh-8861-feedback-options.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
```
Totktonada, line 1: Please, rebase after PR #8877 and remove the relevant options in the `test_box_cfg_coverage` test case from the ignore list.

src/box/lua/config/instance_config.lua
ImeevMA: Fixed.

changelogs/unreleased/gh-8861-feedback-options.md
ImeevMA: Done.

src/box/lua/config/instance_config.lua
```diff
@@ -1179,6 +1183,57 @@ return schema.new('instance_config', schema.record({
...
+            box_cfg = 'feedback_send_metrics',
+            default = true,
+            apply_default_if = feedback_apply_default_if,
+        }),
+        enabled = schema.scalar({
+            type = 'boolean',
+            box_cfg = 'feedback_enabled',
+            default = true,
+            apply_default_if = feedback_apply_default_if,
+        }),
```
Totktonada, line 1216: Nit: I would place it first.

src/box/lua/config/instance_config.lua
```diff
@@ -1179,6 +1183,57 @@ return schema.new('instance_config', schema.record({
...
+            box_cfg = 'feedback_interval',
+            default = 3600,
+            apply_default_if = feedback_apply_default_if,
+        }),
+        metrics_limit = schema.scalar({
+            type = 'number',
+            box_cfg = 'feedback_metrics_limit',
+            default = 1024 * 1024,
+            apply_default_if = feedback_apply_default_if,
+        }),
```
Totktonada, line 1228: This value is in bytes, so the fractional part of the number has no sense. I would mark it as `type = 'integer'`.

src/box/lua/config/instance_config.lua
```diff
@@ -1179,6 +1183,57 @@ return schema.new('instance_config', schema.record({
...
+    }, {
+        validate = function(data, w)
+            if data == nil or box.internal.feedback_daemon ~= nil then
+                return
+            end
+            w.error('Feedback is disabled')
+        end,
+    }),
```
Totktonada, line 1236: AFAIU, it discards the following config on tarantool that has the feedback daemon build-time disabled.

```yaml
feedback: {}
```

However, I see nothing criminal in the config: neither of the feedback daemon options are set.

I propose to sink the validation function to individual parameters that should be discarded on the given instance with the given build type.

src/box/lua/config/instance_config.lua
```diff
@@ -1179,6 +1183,57 @@ return schema.new('instance_config', schema.record({
...
+            box_cfg = 'feedback_metrics_limit',
+            default = 1024 * 1024,
+            apply_default_if = feedback_apply_default_if,
+        }),
+    }, {
+        validate = function(data, w)
+            if data == nil or box.internal.feedback_daemon ~= nil then
+                return
+            end
+            w.error('Feedback is disabled')
```
Totktonada, line 1234: I would make it clear that the feedback sending is disabled at compile time, not just using `feedback.enabled`.

For example,

```lua
w.error('Tarantool is built without feedback reports sending support')
```

src/box/lua/config/instance_config.lua
Totktonada: BTW, I would allow the following config as well.

```yaml
feedback:
  enabled: false
```

The interesting question is what to do with the following one?

```yaml
feedback:
  enabled: false
  host: 'https://example.org'
```

Sending the feedback is not requested by this configuration, so we can accept it.

For example, strip all the feedback parameters before passing to `box.cfg()` in `applier/box_cfg.lua` if `feedback.enabled` is `false`.

OTOH, this is non-typical build type (we don't deliver such builds on our own) and it looks as not worthful give the effort.

So, I think that this comment can be ignored (but the one above looks simple and I would implement it).

Totktonada (approved):

src/box/lua/config/instance_config.lua
ImeevMA: Done

src/box/lua/config/instance_config.lua
ImeevMA: Done

src/box/lua/config/instance_config.lua
ImeevMA: Done

src/box/lua/config/instance_config.lua
ImeevMA: Moved validate to individual parameters. However, I think it's better to disable any values for the feedback parameters, since this logic looks simpler and clearer.

PR #8896, opened 2023-07-19
lua: replace `api_check()` with `assert()`

Buristan:
`api_check()` is LuaJIT internal assertion. To prevent inconsistency during internal assertion changing (for example during backporting) use glibc's `assert()` instead.

NO_DOC=internal
NO_TEST=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -67,7 +67,7 @@ index2adr(lua_State *L, int idx)
 		TValue *o = L->base + (idx - 1);
 		return o < L->top ? o : niltv(L);
 	} else if (idx > LUA_REGISTRYINDEX) {
-		api_check(L, idx != 0 && -idx <= L->top - L->base);
+		assert(idx != 0 && -idx <= L->top - L->base);
 		return L->top + idx;
 	} else if (idx == LUA_GLOBALSINDEX) {
 		TValue *o = &G(L)->tmptv;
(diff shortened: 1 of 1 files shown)


fckxorg (changes requested):
Please update your commit message to the one below (grammar and punctuation fixed):

```
`api_check()` is the LuaJIT internal assertion. To prevent inconsistency during internal assertion changes (for example, during backporting), use glibc's `assert()` instead.
```

Buristan:
Hi, Maxim!
Thanks for the review!
Updated the commit message as you suggested.

fckxorg (approved):

igormunkin (approved):
@Buristan, thanks for the patch! LGTM, as trivial.

However, I would like to share my thoughts regarding the change. I'd rather redefine `api_check` in src/lua/utils.c instead of replacing it with plain `assert`. So later, you can just remove the define when the new assertions are backported.

All in all, I see no strong pros for my proposal and have no objections regarding your approach, so feel free to ignore.

Buristan:
Hi, Igor!
Thanks for the review!

`assert()` is used <src/lua/utils.c>-wide, instead some Lua-specific macros. So, to be consistent I prefer to use plain `assert()`.

igormunkin:
Also backported to 2.11 in scope of cb131f6.

PR #8897, opened 2023-07-19
config: introduce flightrec options

ImeevMA:
This patch introduces all flightrec options.

Part of #8861

### DIFF
--- a/changelogs/unreleased/gh-8861-flightrec-options.md
+++ b/changelogs/unreleased/gh-8861-flightrec-options.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* All flight recorder options are now supported (gh-8861).
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -1247,6 +1247,54 @@ return schema.new('instance_config', schema.record({
             validate = feedback_validate,
         }),
     }),
+    flightrec = schema.record({
+        enabled = enterprise_edition(schema.scalar({
+            type = 'boolean',
+            box_cfg = 'flightrec_enabled',
+            default = false,
+        })),
+        logs_size = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_logs_size',
+            default = 10485760,
+        })),
+        logs_max_msg_size = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_logs_max_msg_size',
+            default = 4096,
+        })),
+        logs_log_level = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_logs_log_level',
+            default = 6,
+            allowed_values = {0, 1, 2, 3, 4, 5, 6, 7},
+        })),
+        metrics_interval = enterprise_edition(schema.scalar({
+            type = 'number',
+            box_cfg = 'flightrec_metrics_interval',
+            default = 1.0,
+        })),
+        metrics_period = enterprise_edition(schema.scalar({
+            type = 'number',
+            box_cfg = 'flightrec_metrics_period',
+            default = 60 * 3,
+        })),
+        requests_size = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_requests_size',
+            default = 10485760,
+        })),
+        requests_max_req_size = enterprise_edition(schema.scalar({
+            type = 'integer',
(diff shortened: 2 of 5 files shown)


changelogs/unreleased/gh-8861-flightrec-options.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* All flightrec options are now supported (gh-8861).
```
xuniq, line 3:
```suggestion
* All flight recorder options are now supported (gh-8861).
```

xuniq (approved):

changelogs/unreleased/gh-8861-flightrec-options.md
ImeevMA: Thanks, fixed!

test/config-luatest/cluster_config_schema_test.lua
```diff
@@ -100,6 +100,13 @@ g.test_cluster_config = function()
     t.assert_equals(result, expected)
 end
 
+local function check_enterprise(data_ee, data_ce)
+    if require('tarantool').package == 'Tarantool Enterprise' then
+        return data_ee
+    end
+    return data_ce
+end
```
Lord-KA, line 108: It feels somewhat less verbose than `<if> and true or false` because of the naming, maybe just add a variable `is_tarantool_ee` or something similar? Or just think of the better name, but I can't think of one.

src/box/lua/config/instance_config.lua
```diff
@@ -1179,6 +1179,53 @@ return schema.new('instance_config', schema.record({
...
+        logs_size = enterprise_edition(schema.scalar({
+            type = 'number',
+            box_cfg = 'flightrec_logs_size',
+            default = 10485760,
+        })),
+        logs_max_msg_size = enterprise_edition(schema.scalar({
+            type = 'number',
+            box_cfg = 'flightrec_logs_max_msg_size',
+            default = 4096,
+        })),
```
Totktonada, line 1197: If it is a size, I guess that it should be `integer`, not `number`. Please, re-check other options.

changelogs/unreleased/gh-8861-flightrec-options.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
```
Totktonada, line 1: Needs rebase after PR #8877 and update of the ignore list in the corresponding test case.

test/config-luatest/cluster_config_schema_test.lua
ImeevMA: Replaced by ternary operator. Also, introduced `is_enterprise` variable, so I suggest to use it in #8883.

src/box/lua/config/instance_config.lua
ImeevMA: Changed types of `logs_size`, `logs_max_msg_size`, `logs_log_level`, `requests_size`, `requests_max_req_size`, `requests_max_res_size` to `integer`.

changelogs/unreleased/gh-8861-flightrec-options.md
ImeevMA: Done.

test/config-luatest/cluster_config_schema_test.lua
Lord-KA: Thanks, will do.

ImeevMA:
Changed `flightrec_enabled` default to `false`, as it should be. It was `true` before.

src/box/lua/config/instance_config.lua
```diff
@@ -1241,6 +1241,53 @@ return schema.new('instance_config', schema.record({
...
+        logs_max_msg_size = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_logs_max_msg_size',
+            default = 4096,
+        })),
+        logs_log_level = enterprise_edition(schema.scalar({
+            type = 'integer',
+            box_cfg = 'flightrec_logs_log_level',
+            default = 6,
+        })),
```
Totktonada, line 1264: Nit: Shouldn't we list all possible values as in the `log.level` option? AFAIS from the code (`flightrec_check_cfg`), `flightrec_logs_log_level` doesn't support string values and it is OK to don't add them here too. But I would restrict values to 0-7.

Totktonada (approved):

src/box/lua/config/instance_config.lua
ImeevMA: Thanks, fixed.

PR #8898, opened 2023-07-19
config: introduce memtx.sort_threads option

ImeevMA:
Part of #8861

PR #8903, opened 2023-07-20
config: introduce bootstrap_leader option

ImeevMA:
Part of #8861

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -352,6 +352,9 @@ local function apply(config)
     box_cfg.replicaset_name = names.replicaset_name
     box_cfg.instance_name = names.instance_name
 
+    -- Set bootstrap_leader option.
+    box_cfg.bootstrap_leader = configdata:bootstrap_leader()
+
     -- The startup process may need a special handling and differs
     -- from the configuration reloading process.
     local is_startup = type(box.cfg) == 'function'
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -79,6 +79,10 @@ function methods.is_leader(self)
     return self._leader == self._instance_name
 end
 
+function methods.bootstrap_leader(self)
+    return self._bootstrap_leader
+end
+
 local mt = {
     __index = methods,
 }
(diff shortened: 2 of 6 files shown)


xuniq (approved):

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -351,6 +351,7 @@ local function apply(config)
     box_cfg.cluster_name = names.group_name
     box_cfg.replicaset_name = names.replicaset_name
     box_cfg.instance_name = names.instance_name
+    box_cfg.bootstrap_leader = configdata:bootstrap_leader()
```
Totktonada, line 354: Nit: it is fused with the block regarding names that at least makes the comment to the block incomplete:

> -- Add instance, replicaset and group (cluster) names.

I would move it into its own block.

test/config-luatest/config_test.lua
```diff
@@ -343,3 +343,124 @@ g.test_remaining_vinyl_options = function()
...
+              - super
+
+        iproto:
+          listen: unix/:./{{ instance_name }}.iproto
+
+        groups:
+          group-001:
+            replicasets:
+              replicaset-001:
+                bootstrap_leader: 'instance-002'
```
Totktonada, line 363: Nit: The error message depends on the order of the checks in the code, because this configuration is incorrect in two ways: no `bootstrap_strategy`, non-existing `bootstrap_leader`. I would change the leader to `instance-001`.

src/box/lua/config/configdata.lua
```diff
@@ -191,6 +195,25 @@ local function new(iconfig, cconfig, instance_name)
...
+        'replication.bootstrap_strategy')
+    local bootstrap_leader = found.replicaset.bootstrap_leader
+    if bootstrap_strategy ~= 'config' then
+        if bootstrap_leader ~= nil then
+            error('The "bootstrap_leader" option cannot be set if '..
+                  '"bootstrap_strategy" is not "config"', 0)
+        end
+    elseif bootstrap_leader == nil then
+        error('Incorrect value for option "bootstrap_leader": the option '..
+              'can\'t be empty when bootstrap strategy is "config"', 0)
```
Totktonada, line 208: Nit: Shouldn't we add instance/replicaset/group names into these error messages like it is done for other error messages above and below?

Totktonada (approved):

src/box/lua/config/applier/box_cfg.lua
ImeevMA: Fixed.

test/config-luatest/config_test.lua
ImeevMA: Fixed.

ImeevMA:
I suggest to continue with this PR after #8898 will be merged.

Issue #8473, opened 2023-03-20
Implement UBSan support in LuaJIT


ligurio:
```
[132/242] Generating lj_bcdef.h

18037FAILED: third_party/luajit/src/lj_bcdef.h /src/tarantool/build/third_party/luajit/src/lj_bcdef.h 

18038cd /src/tarantool/build/third_party/luajit/src && /src/tarantool/build/third_party/luajit/src/host/buildvm -m bcdef -o lj_bcdef.h /src/tarantool/third_party/luajit/src/lib_base.c /src/tarantool/third_party/luajit/src/lib_math.c /src/tarantool/third_party/luajit/src/lib_bit.c /src/tarantool/third_party/luajit/src/lib_string.c /src/tarantool/third_party/luajit/src/lib_table.c /src/tarantool/third_party/luajit/src/lib_io.c /src/tarantool/third_party/luajit/src/lib_os.c /src/tarantool/third_party/luajit/src/lib_package.c /src/tarantool/third_party/luajit/src/lib_debug.c /src/tarantool/third_party/luajit/src/lib_jit.c /src/tarantool/third_party/luajit/src/lib_ffi.c /src/tarantool/third_party/luajit/src/lib_misc.c

18039/src/tarantool/third_party/luajit/src/vm_x86.dasc:633:59: runtime error: member access within null pointer of type 'lua_State' (aka 'struct lua_State')

18040SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /src/tarantool/third_party/luajit/src/vm_x86.dasc:633:59 in 

18041[133/242] Building C object src/lib/msgpuck/CMakeFiles/msgpuck.dir/msgpuck.c.o

18042clang-15: warning: the object size sanitizer has no effect at -O0, but is explicitly enabled: -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unsigned-integer-overflow,unreachable,vla-bound,vptr [-Winvalid-command-line-argument]

18043ninja: build stopped: subcommand failed.

180442023-03-20 15:42:57,865 - root - ERROR - Building fuzzers failed.
```

### How-to reproduce:

- Comment out condition for `ENABLE_UB_SANITIZER` in `test/fuzz/CMakeLists.txt`
- Follow steps described in https://google.github.io/oss-fuzz/advanced-topics/reproducing/

[...]

PR #8904, opened 2023-07-21
Add fuzzing tests for IPROTO decoders and decoding decimals

ligurio:
~~Blocked by: #8948, #8921, #8473.~~


0x501D (approved):

sergepetrenko (approved):

ligurio:
(rebased to master and force-pushed)

ylobankov:
Now the `fuzzing (address)` job takes 30 mins instead of 15. IMHO, It is a significant increase. We should think a little what we can do with that.

ylobankov (approved):

Issue #8650, opened 2023-05-12
Please add possiblity to set is_sync=true in box.begin to enforce transaction to be synchronous


ochaton:
Now we have is_sync-spaces, but most general use of Tarantool is still asynchronous spaces. Some operations with asynchronous spaces better be synchronous (ex. `queue:put{ ... }` or insert of some new data).

Only developer knows when to make this specific transaction synchronous. Now we only can `:put{}` with `wait_lsn` or `box.atomic` with update of garbage synchronous space, that looks odd.

It would be better, if we could write `box.begin{is_sync=true} ... box.commit()` and `box.sync_atomic` when we want to make this specific transaction to be synchronous. Of course, this means that replicaset must be configured in full-mesh topology, and instance must be the synchronous queue owner.

PR #8907, opened 2023-07-24
box: add is_sync option

### LINKED ISSUE #7932: Recovery: force recovery should try to recover mixed transactions from WAL

yanshtunder:
Added the new is_sync parameter to `box.begin()`, `box.commit()`, and `box.atomic()`. To make the transaction synchronous, set the `is_sync` option to `true`. If to set any value other than `true/nil`, for example `is_sync = "some string"`, then an error will be thrown. Examples:
```Lua
-- Sync transactions
box.atomic({is_sync = true}, function() ... end)

box.begin({is_sync = true}) ... box.commit({is_sync = true})

box.begin({is_sync = true}) ... box.commit()

box.begin() ... box.commit({is_sync = true})

-- Async transactions
box.atomic(function() ... end)

box.begin() ... box.commit()

```

Closes https://github.com/tarantool/tarantool/issues/8650

NO_DOC=internal

### DIFF
--- a/changelogs/unreleased/gh-8650-add-is_sync-option.md
+++ b/changelogs/unreleased/gh-8650-add-is_sync-option.md
@@ -1,0 +1,6 @@
+## feature/box
+
+* Added a new `is_sync` parameter to `box.atomic()`. To make the transaction
+synchronous, set the `is_sync` option to `true`. Setting `is_sync = false` is
+prohibited. If any value other than true/nil is set, for example
+`is_sync = "some string"`, then an error will be thrown (gh-8650).
--- a/extra/exports
+++ b/extra/exports
@@ -146,6 +146,7 @@ box_txn_begin
 box_txn_commit
 box_txn_id
 box_txn_isolation
+box_txn_make_sync
 box_txn_rollback
 box_txn_rollback_to_savepoint
 box_txn_savepoint
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -377,10 +377,12 @@ struct iproto_msg
 		struct auth_request auth;
 		/** Features request. */
 		struct id_request id;
-		/* SQL request, if this is the EXECUTE/PREPARE request. */
+		/** SQL request, if this is the EXECUTE/PREPARE request. */
 		struct sql_request sql;
-		/* BEGIN request */
+		/** BEGIN request */
 		struct begin_request begin;
+		/** COMMIT request */
+		struct commit_request commit;
 		/** In case of iproto parse error, saved diagnostics. */
 		struct diag diag;
 	};
@@ -1653,6 +1655,8 @@ iproto_msg_decode(struct iproto_msg *msg, struct cmsg_hop **route)
 		return 0;
 	case IPROTO_COMMIT:
 		*route = iproto_thread->commit_route;
+		if (xrow_decode_commit(&msg->header, &msg->commit) != 0)
+			return -1;
 		return 0;
 	case IPROTO_ROLLBACK:
 		*route = iproto_thread->rollback_route;
@@ -2048,6 +2052,7 @@ tx_process_begin(struct cmsg *m)
 	struct obuf *out;
 	struct obuf_svp header;
 	uint32_t txn_isolation = msg->begin.txn_isolation;
+	bool is_sync = msg->begin.is_sync;
 
 	if (tx_check_msg(msg) != 0)
 		goto error;
@@ -2068,6 +2073,9 @@ tx_process_begin(struct cmsg *m)
 		(void)rc;
 		goto error;
 	}
+	if (is_sync)
+		box_txn_make_sync();
+
 	out = msg->connection->tx.p_obuf;
 	header = obuf_create_svp(out);
 	iproto_reply_ok(out, msg->header.sync, ::schema_version);
@@ -2087,9 +2095,13 @@ tx_process_commit(struct cmsg *m)
 	struct iproto_msg *msg = tx_accept_msg(m);
 	struct obuf *out;
 	struct obuf_svp header;
+	bool is_sync = msg->commit.is_sync;
 
 	if (tx_check_msg(msg) != 0)
 		goto error;
+
+	if (is_sync)
+		box_txn_make_sync();
 
 	if (box_txn_commit() != 0)
 		goto error;
--- a/src/box/iproto_constants.h
+++ b/src/box/iproto_constants.h
@@ -214,7 +214,11 @@ extern const char *iproto_flag_bit_strs[];
 	 * Mapping of format identifier to format clause consisting of field
 	 * names and field types.
 	 */								\
-	_(TUPLE_FORMATS, 0x60, MP_MAP)
+	_(TUPLE_FORMATS, 0x60, MP_MAP)					\
+	/**
+	 * Flag indicating whether the transaction is synchronous.
+	 */								\
+	 _(IS_SYNC, 0x61, MP_BOOL)
 
 #define IPROTO_KEY_MEMBER(s, v, ...) IPROTO_ ## s = v,
 
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -417,6 +417,40 @@ static const char *lua_sources[] = {
 static int
 lbox_commit(lua_State *L)
 {
+	int top = lua_gettop(L);
+	if (top >= 1) {
+		/* box.commit(nil) */
+		if (lua_isnil(L, 1))
+			goto box_commit;
+
+		if (lua_type(L, 1) != LUA_TTABLE) {
+			diag_set(IllegalParams, "Illegal parameters, options "
+				 "should be a table");
+			return luaT_error(L);
+		}
+		/* Get is_sync */
+		lua_getfield(L, 1, "is_sync");
+		/* box.commit({is_sync = nil}) */
+		if (lua_isnil(L, -1))
+			goto box_commit;
+
+		if (!lua_isboolean(L, -1)) {
+			diag_set(IllegalParams, "Illegal parameters, is_sync "
+				 "must be a boolean");
+			return luaT_error(L);
+		}
+		bool is_sync = lua_toboolean(L, lua_gettop(L));
+		lua_pop(L, 1);
+
+		if (!is_sync) {
+			diag_set(IllegalParams, "Illegal parameters, is_sync "
+				 "can only be true");
+			return luaT_error(L);
+		}
+		box_txn_make_sync();
+	}
+
+box_commit:
 	if (box_txn_commit() != 0)
 		return luaT_error(L);
 	return 0;
--- a/src/box/lua/net_box.c
+++ b/src/box/lua/net_box.c
@@ -1361,19 +1361,6 @@ netbox_encode_unprepare(lua_State *L, int idx,
 	return netbox_encode_prepare(L, idx, ctx);
 }
 
-static inline int
-netbox_encode_commit_or_rollback(lua_State *L, enum iproto_type type, int idx,
-				 struct mpstream *stream, uint64_t sync,
-				 uint64_t stream_id)
-{
-	(void)L;
-	(void) idx;
-	assert(type == IPROTO_COMMIT || type == IPROTO_ROLLBACK);
-	size_t svp = netbox_begin_encode(stream, sync, type, stream_id);
-	netbox_end_encode(stream, svp);
-	return 0;
-}
-
 /**
  * Encode an `IPROTO_BEGIN` request and write it to the provided MsgPack stream.
  */
@@ -1385,9 +1372,11 @@ netbox_encode_begin(struct lua_State *L, int idx,
 					 ctx->stream_id);
 	bool has_timeout = !lua_isnoneornil(L, idx);
 	bool has_txn_isolation = !lua_isnoneornil(L, idx + 1);
-	if (has_timeout || has_txn_isolation) {
+	bool has_is_sync = !lua_isnoneornil(L, idx + 2);
+	if (has_timeout || has_txn_isolation || has_is_sync) {
 		uint32_t map_size = (has_timeout ? 1 : 0) +
-				    (has_txn_isolation ? 1 : 0);
+				    (has_txn_isolation ? 1 : 0) +
+				    (has_is_sync ? 1 : 0);
 		mpstream_encode_map(ctx->stream, map_size);
 	}
 	if (has_timeout) {
@@ -1402,6 +1391,13 @@ netbox_encode_begin(struct lua_State *L, int idx,
 		mpstream_encode_uint(ctx->stream, IPROTO_TXN_ISOLATION);
 		mpstream_encode_uint(ctx->stream, txn_isolation);
 	}
+	if (has_is_sync) {
+		if (lua_type(L, idx + 2) == LUA_TBOOLEAN) {
+			bool is_sync = lua_toboolean(L, idx + 2);
+			mpstream_encode_uint(ctx->stream, IPROTO_IS_SYNC);
+			mpstream_encode_bool(ctx->stream, is_sync);
+		}
+	}
 	netbox_end_encode(ctx->stream, svp);
 	return 0;
 }
@@ -1410,18 +1406,32 @@ static int
 netbox_encode_commit(struct lua_State *L, int idx,
 		     struct netbox_method_encode_ctx *ctx)
 {
-	return netbox_encode_commit_or_rollback(L, IPROTO_COMMIT, idx,
-						ctx->stream, ctx->sync,
-						ctx->stream_id);
+	size_t svp = netbox_begin_encode(ctx->stream, ctx->sync, IPROTO_COMMIT,
+					 ctx->stream_id);
+	bool has_is_sync = !lua_isnoneornil(L, idx);
+	if (has_is_sync) {
+		uint32_t map_size = 1;
+		mpstream_encode_map(ctx->stream, map_size);
+		if (lua_type(L, idx) == LUA_TBOOLEAN) {
+			bool is_sync = lua_toboolean(L, idx);
+			mpstream_encode_uint(ctx->stream, IPROTO_IS_SYNC);
+			mpstream_encode_bool(ctx->stream, is_sync);
+		}
+	}
+	netbox_end_encode(ctx->stream, svp);
+	return 0;
 }
 
 static int
 netbox_encode_rollback(struct lua_State *L, int idx,
 		       struct netbox_method_encode_ctx *ctx)
 {
-	return netbox_encode_commit_or_rollback(L, IPROTO_ROLLBACK, idx,
-						ctx->stream, ctx->sync,
-						ctx->stream_id);
+	(void)L;
+	(void)idx;
+	size_t svp = netbox_begin_encode(ctx->stream, ctx->sync,
+					 IPROTO_ROLLBACK, ctx->stream_id);
+	netbox_end_encode(ctx->stream, svp);
+	return 0;
 }
 
 /**
--- a/src/box/lua/net_box.lua
+++ b/src/box/lua/net_box.lua
@@ -523,6 +523,7 @@ local function stream_begin(stream, txn_opts, netbox_opts)
     check_param_table(netbox_opts, REQUEST_OPTION_TYPES)
     local timeout
     local txn_isolation
+    local is_sync
     if txn_opts then
         if type(txn_opts) ~= 'table' then
             error("txn_opts should be a table")
@@ -536,19 +537,94 @@ local function stream_begin(stream, txn_opts, netbox_opts)
             txn_isolation =
                 box.internal.normalize_txn_isolation_level(txn_isolation)
         end
+        is_sync = txn_opts.is_sync
+        if is_sync ~= nil and type(is_sync) ~= "boolean" then
+            error("is_sync must be a boolean")
+        end
+        if is_sync == false then
+            error("is_sync can only be true")
+        end
     end
     local res = stream:_request('BEGIN', netbox_opts, nil,
-                                stream._stream_id, timeout, txn_isolation)
+                                stream._stream_id, timeout, txn_isolation,
+                                is_sync)
     if netbox_opts and netbox_opts.is_async then
         return res
     end
 end
 
-local function stream_commit(stream, opts)
+-- A function that correctly orders commit options. This is done so that
+-- txn_opts is the second parameter (as in begin), and opts is the third
+-- parameter. This is done for backward compatibility.
+local function order_opts_commit(opts_1, opts_2)
+    if opts_1 == nil and opts_2 == nil then
+        return opts_1, opts_2
+    end
+    if type(opts_1) ~= 'table' and type(opts_2) ~= 'table' then
+        box.error(box.error.ILLEGAL_PARAMS, "options should be a table")
+    end
+
+    local new_opts_1 = opts_1
+    local new_opts_2 = opts_2
+
+    if opts_1 ~= nil and type(opts_1) == 'table' then
+        for i, _ in pairs(opts_1) do
+            for k, _ in pairs(REQUEST_OPTION_TYPES) do
+                if i == k then
+                    new_opts_1, new_opts_2 = opts_2, opts_1
+                    break
+                end
+            end
+        end
+    end
+
+    if opts_2 ~= nil and type(opts_2) == 'table' then
+        for i, _ in pairs(opts_2) do
+            for k, _ in pairs(REQUEST_OPTION_TYPES) do
+                if i == k then
+                    new_opts_1, new_opts_2 = opts_1, opts_2
+                    break
+                end
+            end
+        end
+    end
+    return new_opts_1, new_opts_2
+end
+
+local function stream_commit(stream, txn_opts, opts)
     check_remote_arg(stream, 'commit')
-    check_param_table(opts, REQUEST_OPTION_TYPES)
-    local res = stream:_request('COMMIT', opts, nil, stream._stream_id)
-    if opts and opts.is_async then
+    local new_txn_opts, new_opts = order_opts_commit(txn_opts, opts)
+    check_param_table(new_opts, REQUEST_OPTION_TYPES)
+
+    -- The options are in the wrong order or mixed up. Example:
+    -- stream:commit({opts})                => ok
+    -- stream:commit({txn_opts})            => ok
+    -- stream:commit({txn_opts}, {opts})    => ok
+    -- stream:commit({opts}, {txn_opts})    => error
+    -- stream:commit({txn_opts, opts})      => error
+    -- stream:commit({opts, txn_opts})      => error
+    -- The opts is netbox options. The txn_opts is transaction options.
+    if new_txn_opts and new_opts and new_txn_opts ~= txn_opts then
+        box.error(box.error.ILLEGAL_PARAMS, "options are either in the " ..
+                  "wrong order or mixed up")
+    end
+
+    local is_sync
+    if new_txn_opts then
+        if type(new_txn_opts) ~= 'table' then
+            box.error(box.error.ILLEGAL_PARAMS, "options should be a table")
+        end
+        is_sync = new_txn_opts.is_sync
+        if type(is_sync) ~= "boolean" and is_sync ~= nil then
+            box.error(box.error.ILLEGAL_PARAMS, "is_sync must be a boolean")
+        end
+        if is_sync == false then
+            box.error(box.error.ILLEGAL_PARAMS, "is_sync can only be true")
+        end
+    end
+    local res = stream:_request('COMMIT', new_opts, nil, stream._stream_id,
+                                is_sync)
+    if new_opts and new_opts.is_async then
         return res
     end
 end
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -85,6 +85,8 @@ ffi.cdef[[
     int
     box_txn_set_timeout(double timeout);
     void
+    box_txn_make_sync();
+    void
     memtx_tx_story_gc_step();
     int
     box_sequence_current(uint32_t seq_id, int64_t *result);
@@ -297,16 +299,27 @@ local begin_options = {
         return true
     end,
     txn_isolation = normalize_txn_isolation_level,
+    is_sync = function(is_sync)
+        if type(is_sync) ~= "boolean" then
+            box.error(box.error.ILLEGAL_PARAMS, "is_sync must be a boolean")
+        end
+        if is_sync == false then
+            box.error(box.error.ILLEGAL_PARAMS, "is_sync can only be true")
+        end
+        return true
+    end,
 }
 
 box.begin = function(options)
     local timeout
     local txn_isolation
+    local is_sync
+    check_param_table(options, begin_options)
     if options then
-        check_param_table(options, begin_options)
         timeout = options.timeout
         txn_isolation = options.txn_isolation and
                         normalize_txn_isolation_level(options.txn_isolation)
+        is_sync = options.is_sync
     end
     if builtin.box_txn_begin() == -1 then
         box.error()
@@ -318,6 +331,9 @@ box.begin = function(options)
        internal.txn_set_isolation(txn_isolation) ~= 0 then
         box.rollback()
         box.error()
+    end
+    if is_sync then
+        builtin.box_txn_make_sync()
     end
 end
 
@@ -358,9 +374,9 @@ local function atomic_tail(status, ...)
     if not status then
         box.rollback()
         error((...), 2)
-     end
-     box.commit()
-     return ...
+    end
+    box.commit()
+    return ...
 end
 
 box.atomic = function(arg0, arg1, ...)
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -850,7 +850,8 @@ txn_journal_entry_new(struct txn *txn)
 
 	struct xrow_header **remote_row = req->rows;
 	struct xrow_header **local_row = req->rows + txn->n_applier_rows;
-	bool is_sync = false;
+	bool is_sync = txn_has_flag(txn, TXN_WAIT_ACK);
+
 	/*
 	 * A transaction which consists of NOPs solely should pass through the
 	 * limbo without waiting. Even when the limbo is not empty. This is
@@ -892,6 +893,15 @@ txn_journal_entry_new(struct txn *txn)
 	 */
 	if (!txn_has_flag(txn, TXN_FORCE_ASYNC) && !is_fully_nop) {
 		if (is_sync) {
+			/*
+			 * Can't commit a fully local transaction synchronously.
+			 */
+			if (txn_is_fully_local(txn)) {
+				diag_set(IllegalParams,
+					 "An attempt to commit a fully local "
+					 "transaction synchronously");
+				return NULL;
+			}
 			txn_set_flags(txn, TXN_WAIT_SYNC | TXN_WAIT_ACK);
 		} else if (!txn_limbo_is_empty(&txn_limbo)) {
 			/*
@@ -1273,6 +1283,20 @@ box_txn_set_timeout(double timeout)
 	}
 	txn_set_timeout(txn, timeout);
 	return 0;
+}
+
+void
+box_txn_make_sync(void)
+{
+	struct txn *txn = in_txn();
+	/*
+	 * Do nothing if transaction is not started,
+	 * it's the same as BEGIN + COMMIT.
+	 */
+	if (!txn)
+		return;
+
+	txn_set_flags(txn, TXN_WAIT_ACK);
 }
 
 /** Wait for a linearization point for a transaction. */
--- a/src/box/txn.h
+++ b/src/box/txn.h
@@ -1042,6 +1042,11 @@ box_txn_set_timeout(double timeout);
 API_EXPORT int
 box_txn_set_isolation(uint32_t level);
 
+/**
+ * Make the transaction synchronous.
+ */
+API_EXPORT void
+box_txn_make_sync(void);
 /** \endcond public */
 
 typedef struct txn_savepoint box_txn_savepoint_t;
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -1755,6 +1755,7 @@ xrow_decode_begin(const struct xrow_header *row, struct begin_request *request)
 	if (row->type != IPROTO_BEGIN)
 		goto bad_msgpack;
 	memset(request, 0, sizeof(*request));
+	request->is_sync = false;
 
 	/** Request without extra options. */
 	if (row->bodycnt == 0)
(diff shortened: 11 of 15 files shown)


src/box/lua/schema.lua
```diff
@@ -460,13 +476,37 @@ box.savepoint = function()
...
+    if options then
+        check_param_table(options, commit_options)
+        is_sync = options.is_sync
+    end
+    if is_sync then
+        assert(builtin.box_txn_set_sync(is_sync) == 0)
+    end
+    if builtin.box_txn_commit() == -1 then
+        box.error()
+    end
```
olegrok, line 500: You can't call box_txn_commit via FFI since it yields

Read about details here https://github.com/tarantool/tarantool/commit/34234427a3cc1f181ef6583c36ed7e4193bafde7

igormunkin: Well, this is not quite true.

The main problem of calling yielding function via FFI machinery is "conditional" yield (one can find an example in [test/app-luatest/lib/libyield.c](https://github.com/tarantool/tarantool/blob/master/test/app-luatest/lib/libyield.c)). Hence the trace can be compiled with no yield on in scope of the function called via FFI, but when yield happens, there is no other way, than call panic.

However, there is no problems with calling a function that yields unconditionally (i.e. `fiber_yield` is called each time the caller is called via FFI), but in that case it has less sense, since trace compilation is aborted and all FFI-related "JIT benefits" are gone.

I glanced on the patch and see no reason moving existing interfaces to FFI (I also found neither benchmarks, nor rationale in commit message/comment for such change).

Considering everything written above, I suggest to not use FFI for you changes, or provide a strong reasoning for this.

ochaton: afaik, box_txn_commit() will not yield if wal_mode is none. so this execution flow is conditionally yielding

changelogs/unreleased/gh-8650-add-is_sync-option.md
```diff
@@ -0,0 +1,10 @@
+## feature/box
+
+* Added a new parameter `is_sync` to `box.begin()` and `box.commit()`.
+Thanks to this parameter can make the transaction synchronous. To do this
+set `is_sync = true`. `box.commit()` cannot change `is_sync` to `false`
+if the transaction was opened as synchronous or has writes to synchro
+spaces.
+
+* Also, now for these purposes can use `box.atomic()` with the `is_sync`
+parameter (gh-8650).
```
xuniq, line 10:
```suggestion
* Added the new `is_sync` parameter to `box.begin()`, `box.commit()`, and `box.atomic()`.
To make the transaction synchronous, set the `is_sync` option to `true`. `box.commit()` can't change `is_sync` to `false`
if the transaction was opened as synchronous or has writes to synchronous
spaces (gh-8650).
```

xuniq (approved):

src/box/lua/schema.lua
yanshtunder: Done

changelogs/unreleased/gh-8650-add-is_sync-option.md
yanshtunder: Fixed.

lastoCHka42 (approved):

src/box/lua/schema.lua
```diff
@@ -403,16 +405,25 @@ local begin_options = {
         return true
     end,
     txn_isolation = normalize_txn_isolation_level,
+    is_sync = function(is_sync)
```
sergepetrenko, line 408: Please also add `IS_SYNC` flag to the protocol:
one might start a transaction over the net by issuing a `IPROTO_BEGIN` request. You need to add `IS_SYNC` to `IPROTO_BEGIN` options.

See `xrow_decode_begin()`, `tx_process_begin()` for reference.

sergepetrenko (changes requested):
Thanks for the patch!

src/box/txn.c
```diff
@@ -850,10 +851,26 @@ txn_journal_entry_new(struct txn *txn)
...
+		}
+
 		if (stmt->row->type != IPROTO_NOP) {
 			is_fully_nop = false;
 			is_sync = is_sync || (stmt->space != NULL &&
-					      space_is_sync(stmt->space));
+					      space_is_sync(stmt->space)) ||
+					      txn->is_sync;
+
+			txn->is_sync = false;
```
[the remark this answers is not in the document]
yanshtunder: `txn->is_sync` needs to be cleared as `box_txn_set_sync()` sets `txn->is_sync = true`. It is done this way because `box.commit({is_sync = false})` cannot change the transaction. Since if the transaction was opened as synchronous (or there were synchronous spaces), it should remain so. Yes, indeed, it is not necessary to clear `txn->is_sync` on every row.

src/box/txn.c
```diff
@@ -850,10 +851,26 @@ txn_journal_entry_new(struct txn *txn)
...
+		    IPROTO_FLAG_WAIT_SYNC | IPROTO_FLAG_WAIT_ACK)) {
+			is_sync = true;
+		}
+
 		if (stmt->row->type != IPROTO_NOP) {
 			is_fully_nop = false;
 			is_sync = is_sync || (stmt->space != NULL &&
-					      space_is_sync(stmt->space));
+					      space_is_sync(stmt->space)) ||
+					      txn->is_sync;
```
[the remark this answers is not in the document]
yanshtunder: Fixed

src/box/txn.h
```diff
@@ -1024,6 +1026,14 @@ box_txn_set_timeout(double timeout);
 API_EXPORT int
 box_txn_set_isolation(uint32_t level);
 
+/**
+ * Set @a sync for a transaction.
```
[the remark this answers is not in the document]
yanshtunder: Done

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,155 @@
...
+    cg.cluster = cluster:new({})
+
+    local box_cfg = {
+        replication = {
+            server.build_listen_uri('master', cg.cluster.id)
+        },
+        replication_synchro_timeout = 0.2,
+        replication_synchro_quorum = 2
+    }
+    cg.master = cg.cluster:build_server({alias = 'master',
```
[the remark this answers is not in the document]
yanshtunder: Fixed

src/box/lua/init.c
```diff
@@ -388,6 +388,17 @@ static const char *lua_sources[] = {
...
+	bool is_sync = false;
+	if (top >= 1 || lua_istable(L, 1)) {
+		lua_getfield(L, 1, "is_sync"); /* Get is_sync. */
+		if (!lua_isboolean(L, -1))
+			return luaL_error(L, "is_sync must be a boolean");
+		is_sync = lua_toboolean(L, lua_gettop(L));
+		lua_pop(L, 1);
+	}
+
+	box_txn_set_sync(is_sync);
```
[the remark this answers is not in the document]
yanshtunder: Can be done as you suggest.

changelogs/unreleased/gh-8650-add-is_sync-option.md
```diff
@@ -0,0 +1,7 @@
+## feature/box
```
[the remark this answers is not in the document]
yanshtunder: What have we agreed:

-  We leave `is_sync` in all three places (begin, commit, atomic) with the only `true` option.
    - We throw an error in the face of the user when he specifies `is_sync = false` or any other value like `is_sync = "some string"`.

-  For users, we document only `atomic`. Documentation for the guru about `begin` and `commit` will be described only in the commit message.

- `is_sync`  should also appear in the `iproto` protocol (`iproto_begin` and `iproto_commit`).

src/box/lua/schema.lua
yanshtunder: **There are several problems here.**
1. At the encoding stage, you need to understand whether `is_sync = false` explicitly or not (`is_sync = nil`). I have to do so:
```C
// iproto_constants.h

_(IS_SYNC_BOOL, 0x48, MP_BOOL) \
_(IS_SYNC_NIL, 0x49, MP_NIL)   \
```
to be able to separate `bool` from `nil` for `is_sync` option. It doesn't look very good. How can this be done better without kludge?

src/box/lua/net_box.c
```diff
@@ -1294,10 +1294,26 @@ netbox_encode_commit_or_rollback(lua_State *L, enum iproto_type type, int idx,
 				 struct mpstream *stream, uint64_t sync,
 				 uint64_t stream_id)
 {
-	(void)L;
-	(void) idx;
 	assert(type == IPROTO_COMMIT || type == IPROTO_ROLLBACK);
 	size_t svp = netbox_begin_encode(stream, sync, type, stream_id);
+	if (type == IPROTO_COMMIT) {
+		bool has_is_sync = !lua_isnone(L, idx);
```
sergepetrenko, line 1300: `lua_isnoneornil` should work just fine.

src/box/lua/net_box.c
```diff
@@ -1309,9 +1325,11 @@ netbox_encode_begin(struct lua_State *L, int idx, struct mpstream *stream,
 	size_t svp = netbox_begin_encode(stream, sync, IPROTO_BEGIN, stream_id);
 	bool has_timeout = !lua_isnoneornil(L, idx);
 	bool has_txn_isolation = !lua_isnoneornil(L, idx + 1);
-	if (has_timeout || has_txn_isolation) {
+	bool has_is_sync = !lua_isnone(L, idx + 2);
```
sergepetrenko, line 1328: `lua_isnoneornil` should work just fine

src/box/lua/net_box.c
```diff
@@ -1294,10 +1294,26 @@ netbox_encode_commit_or_rollback(lua_State *L, enum iproto_type type, int idx,
...
 				 uint64_t stream_id)
 {
-	(void)L;
-	(void) idx;
 	assert(type == IPROTO_COMMIT || type == IPROTO_ROLLBACK);
 	size_t svp = netbox_begin_encode(stream, sync, type, stream_id);
+	if (type == IPROTO_COMMIT) {
+		bool has_is_sync = !lua_isnone(L, idx);
+		if (has_is_sync) {
+			uint32_t map_size = (has_is_sync ? 1 : 0);
```
sergepetrenko, line 1302: `map_size` is always 1 here.

src/box/lua/net_box.lua
```diff
@@ -536,18 +537,40 @@ local function stream_begin(stream, txn_opts, netbox_opts)
...
-                                stream._stream_id, timeout, txn_isolation)
+                                stream._stream_id, timeout, txn_isolation,
+                                is_sync)
     if netbox_opts and netbox_opts.is_async then
         return res
     end
 end
 
-local function stream_commit(stream, opts)
+local function stream_commit(stream, txn_opts, opts)
```
sergepetrenko, line 556: This changes the public user interface.

For example, [here](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#overview) it's stated, that `Most net.box methods accept the last {options} argument`, and this is true for `stream:commit`.
But [stream.commit() doc](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#lua-function.stream.commit) says `stream:commit()` has no arguments. So the user expects the first argument to be `{opts}`. But this is not true after your patch.

We need to document this change somehow and better make the change backward compatible. Users might already be using streams, we can't break them.

For example, you might try to understand if the second option is `opts` or `txn_opts`:
if the second option contains `is_sync` - it is txn_opts, it can't contain `is_async`,  and `opts` should be found in the 3rd argument.
If the second option contains `is_async` -- it is opts, it can't contain `is_sync`, and there should be no 3rd argument. The user can't pass `txn_opts` then.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,259 @@
...
+    t.assert_error_msg_content_equals(err_msg, function()
+        inject_iproto_begin_or_commit(s, 'COMMIT', false)
+    end)
+
+    err_msg = "Invalid MsgPack - request body"
+    t.assert_error_msg_content_equals(err_msg, function()
+        inject_iproto_begin_or_commit(s, 'COMMIT', 'foo')
+    end)
+
+    -- is_sync is set incorrectly
```
sergepetrenko, line 221: Correctly in this case

sergepetrenko (changes requested):
Thanks for the fixes!

Good job on writing a comprehensive test!
We're on the right track here but still a lot has to be fixed.

src/box/txn.c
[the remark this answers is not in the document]
sergepetrenko: > txn->is_sync needs to be cleared as box_txn_set_sync() sets txn->is_sync = true. Yes, indeed, it is not necessary to clear txn->is_sync after every row.

Why can't you leave it `true`?

src/box/iproto.cc
```diff
@@ -2039,6 +2044,8 @@ tx_process_begin(struct cmsg *m)
 		(void)rc;
 		goto error;
 	}
+	say_info("begin is_sync: %s", is_sync ? "TRUE" : "FALSE");
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/txn.h
```diff
@@ -506,6 +506,8 @@ struct txn {
 	bool is_schema_changed;
 	/** Timeout for transaction, or TIMEOUT_INFINITY if not set. */
 	double timeout;
+	/** True in case box.begin() set to true. */
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/xrow.c
```diff
@@ -46,6 +46,7 @@
 #include "iproto_features.h"
 #include "mpstream/mpstream.h"
 #include "errinj.h"
+#include "txn.h"
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/txn.c
```diff
@@ -1283,6 +1297,20 @@ box_txn_set_timeout(double timeout)
 	return 0;
 }
 
+void
+box_txn_set_sync(bool is_sync)
```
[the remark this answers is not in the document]
yanshtunder: Yes, it really is better.

src/box/txn.c
```diff
@@ -869,6 +880,9 @@ txn_journal_entry_new(struct txn *txn)
 
 		req->approx_len += xrow_approx_len(stmt->row);
 	}
+
+	txn->is_sync = false;
```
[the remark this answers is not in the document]
yanshtunder: Looks like you're right. Done.

src/box/lua/schema.lua
```diff
@@ -296,16 +298,28 @@ local begin_options = {
...
+        if type(is_sync) ~= "boolean" then
+            box.error(box.error.ILLEGAL_PARAMS,
+                      "is_sync must be a boolean")
+        end
+        if is_sync == false then
+            box.error(box.error.PROC_LUA, "is_sync can only be true")
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/lua/net_box.lua
```diff
@@ -536,18 +537,40 @@ local function stream_begin(stream, txn_opts, netbox_opts)
             txn_isolation =
                 box.internal.normalize_txn_isolation_level(txn_isolation)
         end
+        is_sync = txn_opts.is_sync
+        if type(is_sync) ~= "boolean" and is_sync ~= nil then
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/lua/net_box.lua
```diff
@@ -536,18 +537,40 @@ local function stream_begin(stream, txn_opts, netbox_opts)
...
+            error("is_sync must be a boolean")
+        end
+        if is_sync == false then
```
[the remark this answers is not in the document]
yanshtunder: I did it by analogy with timeout. I left it as is.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
yanshtunder: Fixed.

src/box/lua/init.c
```diff
@@ -399,6 +399,23 @@ static const char *lua_sources[] = {
 static int
 lbox_commit(lua_State *L)
 {
+	int top = lua_gettop(L);
+	if (top >= 1 || lua_istable(L, 1)) {
+		lua_getfield(L, 1, "is_sync"); /* Get is_sync. */
```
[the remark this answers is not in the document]
yanshtunder: Done

src/box/xrow.c
```diff
@@ -1758,6 +1759,17 @@ xrow_decode_begin(const struct xrow_header *row, struct begin_request *request)
...
 			break;
+		case IPROTO_IS_SYNC_BOOL:
+			if (mp_decode_bool(&d)) {
+				request->is_sync = true;
+				break;
+			} else {
+				goto bad_is_sync;
+			}
+			break;
+		case IPROTO_IS_SYNC_NIL:
```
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/xrow.c
```diff
@@ -1758,6 +1759,17 @@ xrow_decode_begin(const struct xrow_header *row, struct begin_request *request)
 		case IPROTO_TXN_ISOLATION:
 			request->txn_isolation = mp_decode_uint(&d);
 			break;
+		case IPROTO_IS_SYNC_BOOL:
+			if (mp_decode_bool(&d)) {
+				request->is_sync = true;
+				break;
+			} else {
+				goto bad_is_sync;
```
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/xrow.c
```diff
@@ -1768,6 +1780,60 @@ xrow_decode_begin(const struct xrow_header *row, struct begin_request *request)
...
+			}
+			break;
+		case IPROTO_IS_SYNC_NIL:
+			request->is_sync = false;
+			break;
```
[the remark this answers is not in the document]
yanshtunder: Done.

src/box/lua/net_box.c
yanshtunder: Done.

src/box/lua/net_box.c
yanshtunder: Done.

src/box/lua/net_box.c
yanshtunder: Fixed.

src/box/txn.c
```diff
@@ -856,6 +857,16 @@ txn_journal_entry_new(struct txn *txn)
...
 			continue;
 
+		/*
+		 * Look at the flags stmt->row->flags. If the transaction
+		 * is synchronous, then set is_sync = true. This should
+		 * only be done on replicas. The master sets these flags
+		 * and independently decides whether the transaction is
+		 * synchronous or not.
+		 */
+		if ((stmt->row->flags & IPROTO_FLAG_WAIT_ACK) != 0)
```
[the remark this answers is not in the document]
yanshtunder: This is true. In a loop we go through the list until the last element. At the end **item = 0**, (0 is cast to type of item). If we do as you suggest: 
```C
    ....
}

if ((stmt->row->flags & IPROTO_FLAG_WAIT_ACK) != 0)
    is_sync = true;
```
we will get a segfault.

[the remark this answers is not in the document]
yanshtunder: Maybe I don't understand something.

src/box/lua/net_box.lua
yanshtunder: Done.

src/box/lua/net_box.lua
```diff
@@ -536,19 +537,84 @@ local function stream_begin(stream, txn_opts, netbox_opts)
...
 
-local function stream_commit(stream, opts)
+-- A function that correctly orders commit options. This is done so that
+-- txn_opts is the second parameter (as in begin), and opts is the third
+-- parameter. This is done for backward compatibility.
+local function order_opts_commit(opts_1, opts_2)
+    if opts_1 == nil and opts_2 == nil then
+        return opts_1, opts_2
+    end
+    if type(opts_1) ~= 'table' and type(opts_2) ~= 'table' then
```
sergepetrenko, line 563: You missed some of the checks:
```

tarantool> s:commit(1, {})
---
- error: 'builtin/box/net_box.lua:607: attempt to index local ''new_txn_opts'' (a
    number value)'
...

tarantool> s:commit({}, 1)
---
- error: Illegal parameters, options should be a table
...


```

src/box/lua/net_box.lua
```diff
@@ -536,19 +537,84 @@ local function stream_begin(stream, txn_opts, netbox_opts)
...
+    return new_opts_1, new_opts_2
+end
+
+local function stream_commit(stream, txn_opts, opts)
     check_remote_arg(stream, 'commit')
-    check_param_table(opts, REQUEST_OPTION_TYPES)
-    local res = stream:_request('COMMIT', opts, nil, stream._stream_id)
-    if opts and opts.is_async then
+    --check_param_table(opts, REQUEST_OPTION_TYPES)
+    local new_txn_opts, new_opts = order_opts_commit(txn_opts, opts)
```
sergepetrenko, line 597: nit: you may write
`txn_opts, opts = order_opts_commit(txn_opts, opts)`

src/box/lua/schema.lua
```diff
@@ -296,16 +298,28 @@ local begin_options = {
...
+        return true
+    end,
 }
 
 box.begin = function(options)
     local timeout
     local txn_isolation
-    if options then
+    local is_sync
+    if options or options == false then
```
sergepetrenko, line 317: Let's do `options ~= nil`

src/box/txn.c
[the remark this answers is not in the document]
sergepetrenko: > This is true. In a loop we go through the list until the last element. At the end **item = 0**, (0 is cast to type of item). If we do as you suggest:
> 
> ```c
> stailq_foreach_entry(stmt, &txn->stmts, next) {
>     ....
> }
> 
> if ((stmt->row->flags & IPROTO_FLAG_WAIT_ACK) != 0)
>     is_sync = true;
> ```
> 
> we will get a segfault.

Outside of the loop you may do
`stmt = stailq_last_entry(&txn->stmts, typeof(*stmt), next)` and then do the check.

test/box-luatest/gh_6530_net_box_opts_check_test.lua
```diff
@@ -144,7 +144,7 @@ g.test_netbox_all_methods_opts = function()
         stream:begin({}, 123)
     end)
     t.assert_error_msg_contains("options should be a table", function()
-        stream:commit(123)
+        stream:commit({}, 123)
```
sergepetrenko, line 147: Since you made `stream:commit` accept old order of parameters, this test shouldn't require changing to work correctly.

src/box/iproto_constants.h
```diff
@@ -164,6 +164,7 @@ extern const char *iproto_flag_bit_strs[];
 	 */								\
 	_(SQL_INFO, 0x42, MP_MAP)					\
 	_(STMT_ID, 0x43, MP_UINT)					\
+	_(IS_SYNC, 0x49, MP_BOOL)					\
```
sergepetrenko, line 167: Let's better leave the `0x40-0x4f` range for all the SQL keys.
You may make IS_SYNC = 0x60, for example.

sergepetrenko (changes requested):
Thanks for the fixes!
Looking better already. Just some comments remaining.

src/box/lua/net_box.lua
yanshtunder: Done.

src/box/lua/net_box.lua
yanshtunder: These variables will be needed below when comparing `new_txn_opts` and `txn_opts`.

```Lua
if new_txn_opts and new_opts and new_txn_opts ~= txn_opts then
```

src/box/lua/schema.lua
yanshtunder: Done.

src/box/txn.c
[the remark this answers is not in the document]
yanshtunder: Done.

test/box-luatest/gh_6530_net_box_opts_check_test.lua
yanshtunder: Fixed.

src/box/iproto_constants.h
yanshtunder: Fixed.

src/box/txn.c
```diff
@@ -836,7 +837,7 @@ txn_journal_entry_new(struct txn *txn)
 
 	struct xrow_header **remote_row = req->rows;
 	struct xrow_header **local_row = req->rows + txn->n_applier_rows;
-	bool is_sync = false;
+	bool is_sync = txn->is_sync;
```
sergepetrenko, line 840: It seems you shouldn't allow to make fully-local transactions synchronous:
```
tarantool> box.begin({is_sync=true}) box.space.loc:replace{1} box.commit()
Assertion failed: (lsn > prev_lsn), function vclock_follow, file vclock.c, line 46.
```
`loc` is a local space in this example.

Let's better throw an error when trying to commit a fully local transaction with `is_sync = true` .

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,319 @@
...
+
+g.before_each(function(cg)
+    cg.cluster = cluster:new({})
+    local box_cfg = {
+        memtx_use_mvcc_engine = true,
+        replication = {
+            server.build_listen_uri('master', cg.cluster.id),
+            server.build_listen_uri('replica', cg.cluster.id)
+        },
+        replication_synchro_timeout = 0.2,
```
sergepetrenko, line 17: I suggest you leave the default value for `replication_synchro_timeout` here.

When you're testing a rollback you may re-set `replicaiton_synchro_timeout` to a tiny value like `0.01`.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,319 @@
...
+        box.begin({is_sync = true})
+            box.space.test:insert{4, 'is_sync = true'}
+        box.commit({is_sync = true})
+    end)
+
+    t.assert_error_msg_content_equals('Illegal parameters, is_sync ' ..
+                                      'can only be true',
+    function()
+        cg.master:exec(function()
+            box.begin({is_sync = false})
```
sergepetrenko, line 57: this function always fails on the first `box.begin({is_sync = false})` without testing the other variants.

You have to call 4 different functions with all possible erroneous argument combinations.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,319 @@
...
+
+    t.assert_error_msg_content_equals('Illegal parameters, is_sync ' ..
+                                      'can only be true',
+    function()
+        cg.master:exec(function()
+            box.atomic({is_sync = false}, function()
+                box.space.test:insert{5, 'is_sync = error'}
+                box.space.test:insert{6, 'is_sync = error'}
+            end)
+            box.atomic({is_sync = "foo"}, function()
```
sergepetrenko, line 134: Same as above, you should break these calls into two functions. Otherwise only the first one is tested.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
```diff
@@ -0,0 +1,319 @@
...
+    s:commit()
+
+    -- Sync transaction
+    t.assert_error_msg_content_equals('Quorum collection for a synchronous ' ..
+                                      'transaction is timed out',
+    function()
+        s:begin({is_sync = true})
+        s.space.test:insert({5, 'is_sync = true'})
+        s:commit()
+        s:begin()
```
sergepetrenko, line 250: Same as above, the first begin-commit always fails due to a timeout, the second begin-commit is not tested.

sergepetrenko (changes requested):
Thanks for the fixes!

Mostly good now, but some minor fixes are needed here and there.

src/box/lua/net_box.lua
sergepetrenko: Ok, I see.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
yanshtunder: Fixed

test/box-luatest/gh_8650_add_is_sync_option_test.lua
yanshtunder: Fixed

test/box-luatest/gh_8650_add_is_sync_option_test.lua
yanshtunder: Fixed

src/box/txn.c
yanshtunder: Done.

test/box-luatest/gh_8650_add_is_sync_option_test.lua
sergepetrenko: Let's also set `replication_synchro_timeout` to a tiny value before testing rollbacks on master. So that the test doesn't run for 30 seconds.

Like this:
```diff
diff --git a/test/box-luatest/gh_8650_add_is_sync_option_test.lua b/test/box-luatest/gh_8650_add_is_>
index 2693a2bdf5..969a63b304 100644
--- a/test/box-luatest/gh_8650_add_is_sync_option_test.lua
+++ b/test/box-luatest/gh_8650_add_is_sync_option_test.lua
@@ -107,9 +107,12 @@ g.test_box_begin_commit_is_sync = function(cg)
     err_msg = "Quorum collection for a synchronous transaction is timed out"
     t.assert_error_msg_content_equals(err_msg, function()
         cg.master:exec(function()
+            local timeout = box.cfg.replication_synchro_timeout
+            box.cfg{replication_synchro_timeout = 0.01}
             box.begin({is_sync = true})
                 box.space.test:insert{10, 'is_sync = true'}
             box.commit()
+            box.cfg{replication_synchro_timeout=timeout}
         end)
     end)
 
```

sergepetrenko (approved):
Thanks for the fixes!
Looks good. Just don't forget to address my last comment:
https://github.com/tarantool/tarantool/pull/8907#discussion_r1356522369

changelogs/unreleased/gh-8650-add-is_sync-option.md
```diff
@@ -0,0 +1,6 @@
...
```
Gerold103, line 1: The docbot only mentions `box.atomic`, but you added it to `box.begin` and `box.commit` too. Please, document all three. Same for the netbox changes.

src/box/lua/schema.lua
```diff
@@ -296,16 +298,28 @@ local begin_options = {
...
+        return true
+    end,
 }
 
 box.begin = function(options)
     local timeout
     local txn_isolation
-    if options then
+    local is_sync
+    if options ~= nil or options == false then
```
Gerold103, line 317: Why check `options == false`? Do you mean the code can be called as `box.begin(false)`? Is it a leftover from some previous version of the patch?

src/box/txn.c
```diff
@@ -885,6 +896,19 @@ txn_journal_entry_new(struct txn *txn)
...
+	 * is synchronous, then set is_sync = true. This should
+	 * only be done on replicas. The master sets these flags
+	 * and independently decides whether the transaction is
+	 * synchronous or not. All txn meta flags are set only for
+	 * the last txn row.
+	 */
+	stmt = stailq_last_entry(&txn->stmts, typeof(*stmt), next);
+	if ((stmt->row->flags & IPROTO_FLAG_WAIT_ACK) != 0)
+		is_sync = true;
+
```
Gerold103, line 911: This number of new `if` and a new `bool` flag in the txn object don't look good at all given how hot the code is. Why couldn't you set `TXN_WAIT_ACK` flag in `make_sync()`? It doesn't seem that you need a new boolean member in `struct txn`. It is only confusing now, that we have the "sync"-property spread across 2 members (flags and the new bool).

Gerold103 (changes requested):
Thanks for the patch!

changelogs/unreleased/gh-8650-add-is_sync-option.md
yanshtunder: ### **What have we agreed:**

-  We leave `is_sync` in all three places (begin, commit, atomic) with the only `true` option.
    - We throw an error to the user when he specifies `is_sync != true` or any other value like `is_sync = "some string"`.

-  For users, we document only `atomic`. Documentation for the guru about `begin` and `commit` will be described only in the commit message.

- `is_sync`  should also appear in the `iproto` protocol (`iproto_begin` and `iproto_commit`).

src/box/lua/schema.lua
yanshtunder: Ok, I removed that from the check condition. But if the user makes a mistake and writes:
```
tarantool> box.begin(true) box.space.test:insert{5, 'is_sync = nil'} box.commit()
---
- error: Illegal parameters, options should be a table
...

tarantool> box.begin(false) box.space.test:insert{6, 'is_sync = nil'} box.commit()
---
...
```
Would like to be able to warn the user.

src/box/txn.c
```diff
@@ -850,7 +850,17 @@ txn_journal_entry_new(struct txn *txn)
...
+	/*
+	 * Can't commit a fully local transaction synchronously.
+	 */
+	if (is_sync && txn_is_fully_local(txn)) {
+		diag_set(IllegalParams, "An attempt to commit a fully local "
+			 "transaction synchronously");
+		return NULL;
+	}
+
```
Gerold103, line 863: This better be moved into the existing `if` below, inside `if (!txn_has_flag(txn, TXN_FORCE_ASYNC) && !is_fully_nop) {` branch. Then we won't have this additional check of `is_sync` for regular transactions.

src/box/txn.c
```diff
@@ -885,6 +895,19 @@ txn_journal_entry_new(struct txn *txn)
...
+	 * is synchronous, then set is_sync = true. This should
+	 * only be done on replicas. The master sets these flags
+	 * and independently decides whether the transaction is
+	 * synchronous or not. All txn meta flags are set only for
+	 * the last txn row.
+	 */
+	stmt = stailq_last_entry(&txn->stmts, typeof(*stmt), next);
+	if ((stmt->row->flags & IPROTO_FLAG_WAIT_ACK) != 0)
+		is_sync = true;
+
```
Gerold103, line 910: AFAIU, this is needed only for remote transaction coming from the applier. Then it has to be moved into the applier code. To `apply_plain_tx()` and wherever else it starts transactions. Same for the recovery stage for transactions obtained from the xlogs. Like I said, this code (`txn_journal_entry_new()`) is very hot, it has to have as few `if`s as possible.

Gerold103 (changes requested):
Thanks for the fixes! Just 2 small comments are left.

src/box/lua/schema.lua
Gerold103: Then you would also have to check for all other types like strings, numbers, etc. I like the new version better. Where you've started using `check_param_table()`.

src/box/txn.c
yanshtunder: Moved to applier.

extra/exports
```diff
@@ -151,6 +151,7 @@ box_txn_rollback_to_savepoint
 box_txn_savepoint
 box_txn_set_isolation
 box_txn_set_timeout
+box_txn_make_sync
```
Gerold103, line 154: Please, keep the file sorted. It is explained in the beginning of it.

src/box/txn.c
```diff
@@ -885,13 +886,23 @@ txn_journal_entry_new(struct txn *txn)
 
 		req->approx_len += xrow_approx_len(stmt->row);
 	}
+
```
Gerold103, line 889: Unnecessary diff.

Gerold103 (approved):
Thanks for the fixes! Please, fix the remaining two nits and proceed to the next reviewer.

extra/exports
yanshtunder: Fixed.

src/box/txn.c
yanshtunder: Fixed.

src/box/lua/net_box.lua
```diff
@@ -536,19 +537,87 @@ local function stream_begin(stream, txn_opts, netbox_opts)
...
-    check_param_table(opts, REQUEST_OPTION_TYPES)
-    local res = stream:_request('COMMIT', opts, nil, stream._stream_id)
-    if opts and opts.is_async then
+    local new_txn_opts, new_opts = order_opts_commit(txn_opts, opts)
+    check_param_table(new_opts, REQUEST_OPTION_TYPES)
+
+    -- The options are in the wrong order. Example:
+    -- stream:commit({opts}, {txn_opts})
+    if new_txn_opts and new_opts and new_txn_opts ~= txn_opts then
+        error("Incorrect use of options")
```
sergos, line 602: I can't see the reason for inconsistency in this function: you use plain 'error' here, then 'box.error' later, then plain 'error' again. Should it be the same for all?

sergos (changes requested):
Thanks for the patch!
Please, update the commit message: 
"Added the new is_sync" -> "Added a new is_sync"
"If to set any value other than true/nil" -> "If  any value other than true/nil is set"

Also, there's some inconsistency in error handling, please follow up.

changelogs/unreleased/gh-8650-add-is_sync-option.md
```diff
@@ -0,0 +1,6 @@
...
+
+* Added the new `is_sync` parameter to `box.atomic()`. To make the
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

changelogs/unreleased/gh-8650-add-is_sync-option.md
```diff
@@ -0,0 +1,6 @@
...
+transaction synchronous, set the `is_sync` option to `true`. Setting
+`is_sync = false` is prohibited. If to set any value other than true/nil,
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/lua/net_box.lua
yanshtunder: Fixed.

sergos (approved):

PR #8911, opened 2023-07-25
git: add diff=cpp attribute for C/C++ source files

locker:
This improves diff hunk name detection. Needed for checkpatch to correctly detect if patched code belongs to a function.


alyapunov (approved):

locker:

PR #8914, opened 2023-07-25
perf: initial version of 1M operations test

sergos:
The test can be used for regression testing. It is advisable to tune the machine: check the NUMA configuration, fix the pstate or similar CPU autotune. Although, running dozen times gives more-less stable result for the peak performance, that should be enough for regression identification.

### DIFF
--- a/perf/lua/1mops_write.lua
+++ b/perf/lua/1mops_write.lua
@@ -1,0 +1,335 @@
+#!/bin/env tarantool
+
+local err, popen = pcall(require,'popen')
+if err ~= true then
+    print(err)
+    popen = {}
+end
+local clock = require('clock')
+local fiber = require('fiber')
+local math = require('math')
+local json = require('json')
+local fio = require('fio')
+
+local params = require('internal.argparse').parse(arg, {
+    {'engine', 'string'},
+    {'fibers', 'number'},
+    {'help', 'boolean'},
+    {'h', 'boolean'},
+    {'index', 'string'},
+    {'nodes', 'number'},
+    {'nohint', 'boolean'},
+    {'ops', 'number'},
+    {'sync', 'boolean'},
+    {'transaction', 'number'},
+    {'wal_mode', 'string'},
+    {'warmup', 'number'},
+})
+
+local test_dir = fio.tempdir()
+
+local function exit(res, details)
+    if (details ~= nil) then
+        print(details)
+    end
+    if test_dir ~= nil then
+        fio.rmdir(test_dir)
+        test_dir = nil
+    end
+    os.exit(res)
+end
+
+if params.help or params.h then
+    print([[
+
+ Usage: taskset 0xef tarantool 1mops_write.lua [options]
+
+ Options can be used with '--', followed by the value if it's not
+ a boolean option. The options list with default values:
+
+   engine <string, 'memtx'>   - select engine type
+   fibers <number, 50>        - number of fibers to run simultaneously
+   help (same as -h)          - print this message
+   index <string, 'HASH'>     - select index type
+   nodes <number, 1>          - number of nodes to test one-way replication
+   nohint <boolean>           - to turn the index hints off
+   ops <number, 10000000>     - total amount of replaces to be performed
+   sync <boolean>             - to turn the synchro replication on
+   transaction <number, 150>  - number of replaces in one transaction
+   wal_mode <string, 'write'> - WAL synchronization mode
+   warmup <number, 10>        - percent of ops to skip before measurement
+
+ Being run without options tests the HASH index of memtx engine by
+ running of 50 fibers that replace 150 integers using each as a primary
+ key. Refer to the 'fiber_load' function below to change the workload.
+]])
+    exit(0)
+end
+
+-- turn true to test the qsync
+local test_qsync = params.qsync or false
+
+-- number of operations performed by test
+local num_ops = params.ops or 10000000
+
+-- number of fibers
+local num_fibers = params.fibers or 50
+
+-- number of operations per transaction
+local ops_per_txn = params.transaction or 150
+
+-- number of nodes - master and replicas
+local nodes = params.nodes or 1
+
+-- engine to use
+local engine = params.engine or 'memtx'
+assert(engine == 'memtx' or engine == 'vinyl')
+
+-- WAL mode
+local wal_mode = params.wal_mode or 'write'
+assert(wal_mode == 'write'
+       or wal_mode == 'none'
+       or wal_mode == 'fsync')
+
+-- type of index to use
+local index_config = {}
+index_config.type = params.index or 'HASH'
+index_config.type = string.upper(index_config.type)
+
+local hints_msg = ''
+if params.index == 'TREE' then
+    hints_msg = ' with hints'
+    if params.nohint then
+        index_config.hint = false
+        hints_msg = ' without hints'
+    end
+end
+
+-- warm up part, not counted in average
+local warmup_thr = params.warmup or 10
+warmup_thr = warmup_thr > 100 and 100 or warmup_thr
+warmup_thr = warmup_thr < 0 and 0 or warmup_thr
+-- END OF TUNABLE OPTIONS
+
+-- transactions per fiber
+local trans_per_fiber = num_ops/ops_per_txn/num_fibers
+
+-- by default no output from replicas are received
+-- redirect it into master's one breaks terminal
+local std_redirect = {
+    stdin='devnull',
+    stdout='devnull',
+    stderr='devnull'
+}
+
+print(string.format([[
+# making %d REPLACE operations,
+# %d operations per txn,
+# using %d fibers,
+# in a replicaset of %d nodes,
+# using %s index type%s
+# with WAL mode %s
+# ]],
+      num_ops, ops_per_txn, num_fibers, nodes,
+      index_config.type, hints_msg, wal_mode))
+
+box.cfg{
+    log_level = 0,
+    listen = 0,
+    read_only = false,
+    memtx_memory = 2*1024*1024*1024,
+    work_dir = test_dir,
+    wal_mode = wal_mode,
+}
+
+box.schema.user.create('replicator', {password = 'password'})
+box.schema.user.grant('replicator', 'replication')
+
+-- number of nodes, storage for popen handles
+local nodes_ph = {}
+
+-- run replicas (if needed)
+if (nodes > 1) then
+    if (nodes ~= nil and nodes < 32 and nodes > 0) then
+        if (type(popen) ~= 'table' or popen.new == nil) then
+            exit(1, "can't run replication w/o popen")
+        end
+        print('# starting ' .. nodes - 1 .. ' replicas')
+        for i = 3302, 3300+nodes do
+            -- subdir for replica's data
+            os.execute('mkdir ' .. i)
+            -- command line for replica to start
+            local cmd = {
+                arg[-1],
+                '-e',
+                string.format([[
+                box.cfg {
+                    read_only = false,
+                    log_level = 5,
+                    replication = {'replicator:password@%s'},
+                    work_dir = '%s',
+                    replication_connect_quorum=1
+                }]], box.info.listen, i)
+            }
+            local res, err = popen.new(cmd, std_redirect)
+            if (res) then
+                nodes_ph[i] = res
+            else
+                exit(1, 'error running replica: ' .. err)
+            end
+        end
+
+        -- wait for all replicas to connect
+        while #box.info.replication < nodes do
+            print('# replication', #box.info.replication)
+            fiber.sleep(0.1)
+        end
+    else
+        exit(1, 'Incorrect number of nodes \"' .. arg[1] .. '\" must be 1..31')
+    end
+end
+
+local space
+local done = false
+local err
+
+if (test_qsync) then
+    box.cfg{replication_synchro_quorum = nodes}
+    print('# promoting')
+    box.ctl.promote()
+    print('# done')
+    space, err = box.schema.create_space('test',
+                                         {engine = engine, is_sync = true})
+else
+    space, err = box.schema.create_space('test',
+                                         {engine = engine})
+end
+if space == false then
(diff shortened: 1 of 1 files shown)


perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+    popen = {}
+end
+local clock = require('clock')
+local fiber = require('fiber')
+local math = require('math')
+local json = require('json')
+
+-- working dir is cleared/created for every run, set test_dir
+-- during the test a subdir will be created for each replica
+local function getarg(s, f)
```
igormunkin, line 30: It would be nice to tweak the benchmark setup via environment variables instead of the CLI options to run it within container.

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+
+print(string.format([[
+making %d operations,
+%d operations per txn usnig %d fibers,
+replicaset of %d nodes]],
+      num_ops, ops_per_txn, num_fibers, nodes))
+
+-- cleanup the test dir - handy for restart
+os.execute('rm -rf ' .. test_dir)
+os.execute('mkdir ' .. test_dir)
```
igormunkin, line 91: I believe you can simply use [os.tmpname](https://www.lua.org/manual/5.1/manual.html#pdf-os.tmpname) here, instead of such kind of cleanup?

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+
+-- stop timer for master
+local len = {
+    clock.time()-timer_begin[1],
+    clock.proc()-timer_begin[2]
+}
+
+ops_done = box.info.lsn - ops_done
+
+print(string.format('master done %d ops in time: %f, cpu: %f',
```
igormunkin, line 257: It would be perfect to have an option for TSV/CSV output to process the result via Perl script or load it into Excel to calculate some statistics.

igormunkin (changes requested):
@sergos, thanks for finally moving this to the trunk! I added myself to the reviewers list, since I use this benchmark a lot (though it's [mine fork](https://gist.github.com/igormunkin/e956396d8fb3348d3ebbad60588af297)).

See my major comments for the first iteration of review below.

perf/lua/1mops_write.lua
sergos: Hmm... I suppose running it in any form of virtual environment will hurt the stability of the test.
Nevertheless - I can handle it, perhaps along with rewrite the arguments parsing through the internal.argparse.

perf/lua/1mops_write.lua
sergos: Then it will pile up the temporary files from sequential runs? At least for those crashed, where I can't handle all type of exits or make any on_shutdown() operable. Removing the work_dir right after box_cfg() will work only until xlog rotation, where new file should be open in this folder.

igormunkin: What are the critical points where the benchmark can crash?

sergos: Well, beyond other issues, the very beginning of the test
```{lua}
[s.ostanevich@dev1 lua]$ ../../src/tarantool -i
tarantool> box.cfg{work_dir='asdf'}
can't chdir to `asdf': No such file or directory
[s.ostanevich@dev1 lua]$ echo $?
1
[...]
2023-07-26 13:27:47.084 [52922] main/103/interactive F> failed to preallocate 4398046511104 bytes: Cannot allocate memory, check option 'memtx_memory' in box.cfg(..)
[s.ostanevich@dev1 lua]$ echo $?
1
```
Literally - any problem in box.cfg{} can lead to an immediate termination.

sergos: Unfortunately, the `os.tmpname()` creates a temporary file, so as `io.tmpfile()`. Apparently, there's no safe interface to provide a unique name that will be available and unique over time between its creation and using it.
The only w/a I can envision is to create the temp file, then extend its name with a suffix and create a temp dir along with it, cleaning up both upon completion. Upvote if it is feasible.

sergos: Nevermind, will follow @ligurio proposal of `fio.tmpdir()`

perf/lua/1mops_write.lua
sergos: Followed the rules of [compare.lua](https://github.com/tarantool/tarantool/blob/master/perf/lua/compare.lua) - see the comment below

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
+-- Usage: taskset 0xf0 tarantool relay-1mops.lua [options]
```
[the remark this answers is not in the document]
sergos: done

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+--   nodes <number, 1> - number of nodes to test one-way replication
+--   nohint <boolean> - to turn the hints off
+--   ops <number, 10000000> - total amount of replaces to be performed
+--   sync <boolean> - to turn the synchro replication on
+--   transaction <number, 150> - number of replaces in one transaction
+--   warmup <number, 10> - percent of ops to skip before measurement
+--
+-- Being run without options tests the HASH index of memtx engine by
+-- running of 50 fibers that replace 150 integers using each as a primary
+-- key. Refer to the 'fiber_load' function below to change the workload.
```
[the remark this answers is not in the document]
sergos: done

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+        if arg[i] == '--'..s or arg[i] == '-'..s then
+            if f == nil then return true
+            else return f(arg[i+1]) or error('wrong option -' .. s)
+            end
+        end
+    end
+    return false
+end
+
+local test_dir = './relay-test.data'
```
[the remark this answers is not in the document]
sergos: done

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+    print('error creating space', json.encode(err))
+    os.exit(1)
+end
+local res, err = pcall(space.create_index, space, 'pk', index_config)
+if (res ~= true) then
+    print('error creating index: ', json.encode(err))
+    os.exit(1)
+end
+
+-- THE load fiber
```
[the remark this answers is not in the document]
sergos: an intentional emphasis

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+    clock.time()-timer_begin[1],
+    clock.proc()-timer_begin[2]
+}
+
+ops_done = box.info.lsn - ops_done
+
+print(string.format('master done %d ops in time: %f, cpu: %f',
+                    ops_done, len[1], len[2]))
+print('master average speed', math.floor(ops_done / len[1]), 'ops/sec')
+print('master peak speed', math.floor(max_rps), 'ops/sec')
```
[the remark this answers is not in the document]
sergos: done, we also agreed with @locker on the `#` symbol as a comment in the tests output to let `compare.lua` to ignore them

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,294 @@
...
+-- a boolean option. The options list with default values:
+--   engine <string, 'memtx'> - select engine type
+--   fibers <number, 50> - number of fibers to run simultaneously
+--   index <string, 'HASH'> - select index type
+--   nodes <number, 1> - number of nodes to test one-way replication
+--   nohint <boolean> - to turn the hints off
+--   ops <number, 10000000> - total amount of replaces to be performed
+--   sync <boolean> - to turn the synchro replication on
+--   transaction <number, 150> - number of replaces in one transaction
+--   warmup <number, 10> - percent of ops to skip before measurement
```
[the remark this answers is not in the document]
sergos: done, using the comment prefix (see above)

perf/lua/1mops_write.lua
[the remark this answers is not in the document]
ligurio: Nice! Thanks!

perf/lua/1mops_write.lua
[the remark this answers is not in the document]
ligurio: Nice!

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,323 @@
+local err, popen = pcall(require,'popen')
```
ligurio, line 1: Currently it is not possible to run a script without specifying interpreter:
```
$ ./perf/lua/1mops_write.lua
-bash: ./perf/lua/1mops_write.lua: Permission denied    
```

I would add a shebang:

```diff
--- a/perf/lua/1mops_write.lua
+++ b/perf/lua/1mops_write.lua
@@ -1,3 +1,5 @@
+#!/bin/env tarantool
+
 local err, popen = pcall(require,'popen')
 if err ~= true then
     print(err)
```

and give exec permissions:

```
$ chmod +x perf/lua/1mops_write.lua 
$ ./perf/lua/1mops_write.lua 
...
```

ligurio (approved):
LGTM (after answering to a comment about shebang)

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+   ops <number, 10000000>    - total amount of replaces to be performed
+   sync <boolean>            - to turn the synchro replication on
+   transaction <number, 150> - number of replaces in one transaction
+   warmup <number, 10>       - percent of ops to skip before measurement
+
+ Being run without options tests the HASH index of memtx engine by
+ running of 50 fibers that replace 150 integers using each as a primary
+ key. Refer to the 'fiber_load' function below to change the workload.
+]])
+    os.exit(0)
```
igormunkin, line 52: Nit: I'd rather use `exit` helper here either, however, this is totally minor, so feel free to ignore.

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+            start = start + 1
+        end
+        box.commit()
+    end
+end
+
+-- fiber storage to join them
+local fibers_storage = {}
+
+collectgarbage()
```
igormunkin, line 223: Nit: It also would be nice to have an option whether run in a "dirty"/"clean" environment or even with GC disabled.

igormunkin (approved):
@sergos, thanks for the changes! Honestly, I see no reason to block this PR from my side. However, I left a couple of nits regarding the benchmark usage: you can fix it now, or file a ticket to improve the bench later, or even resolve it as irrelevant (if you consider anything you find below).

All in all, the changes LGETM, thanks again for moving them into the trunk!

ligurio:
Probably it is worth fixing name and email:

> commit cb4580cf09d225e6956d8d60f7092b7c885e53a1 (sergos/sergos/1mops_perf_test, sergos/1mops_perf_test)
> Author: s.ostanevich <s.ostanevich@dev1.tarantool.i>
> Date:   Tue Jul 25 18:45:01 2023 +0300
>
>    perf: initial version of 1M operations test

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+local std_redirect = {
+    stdin='devnull',
+    stdout='devnull',
+    stderr='devnull'
+}
+
+print(string.format([[
+# making %d REPLACE operations,
+# %d operations per txn,
+# usnig %d fibers,
```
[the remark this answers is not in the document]
sergos: Done.

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+# in a replicaset of %d nodes,
+# using %s index type%s
+# ]],
+      num_ops, ops_per_txn, num_fibers, nodes, index_config.type, hints_msg))
+
+box.cfg{
+    log_level = 0,
+    listen = 0,
+    read_only = false,
+    memtx_memory = 2*1024*1024*1024,
```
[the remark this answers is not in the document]
sergos: Actually I have a limit to the number of raws in the table, it is a 1M. Should I minimize/automate the amount of memory is questionable, better make yet another CLI option with a default?

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+    work_dir=test_dir,
+    wal_mode='write',
```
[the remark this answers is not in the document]
sergos: Done.

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+box.schema.user.grant('replicator', 'replication')
+
+-- number of nodes, storage for popen handles
+local nodes_ph = {}
+
+-- run replicas (if needed)
+if (nodes > 1) then
+    if (nodes ~= nil and nodes < 32 and nodes > 0) then
+        if (type(popen) ~= 'table' or popen.new == nil) then
+            exit(1, 'can\'t run replication w/o popen')
```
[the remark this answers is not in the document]
sergos: The reason is delivery(?) team claimed that Lua should have `'` to allow transparent `"` from external calls, like `-e "local s = 'hello' print(s)"`
Anyways, here it goes.

perf/lua/1mops_write.lua
```diff
@@ -0,0 +1,325 @@
...
+        end
+        print('# starting ' .. nodes - 1 .. ' replicas')
+        for i = 3302, 3300+nodes do
+            -- subdir for replica's data
+            os.execute('mkdir ' .. i)
```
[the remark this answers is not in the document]
sergos: The replica folders are inside the temporary folder, no need to generate more of them.

perf/lua/1mops_write.lua
sergos: Done.

perf/lua/1mops_write.lua
sergos: I believe it should be eliminated as a class of possible fluctuation - these GC triggers were introduced to stabilize the results, still they failed. I would like to keep them as is, proceeding with further search of the fluctuation source.
Overloading the test with big number of options is no good IMHO.

PR #8915, opened 2023-07-26
Implement key part sort order.

### LINKED ISSUE #5529: Introduce descending indexes
It is possible to set sort order in index definition:
`i = s:create_index('test', {type='tree', parts={{1, 'uint', sort_order='desc'}}})`.
However, it affects nothing:
```
s = box.schema.create_space('t')
i = s:create_index('test', {type='tree', parts={{1, 'uint', sort_order='desc'}}})
s:insert({1})
s:insert({3})
s:insert({2})
i:select()
---
- - [1]
  - [2]
  - [3]
```
As one can see, results are still appear in ascending order. So the proposal is to make `sort_order` work as it is supposed to do.

Needed for #2596

mkostoevr:
## Overview

This patch implements the previously introduced (but non-functional) `sort_order` key part parameter and exposes corresponding API to Lua's `key_def` API and C API.

The `key_def` is made aware if its index is unordered (the new `is_unordered` field is introduced), the `key_def_new` function signature has been modified: new `flags` parameter has been introduced, the `for_func_index` flag has been moved to it. Also `is_unordered` flag has been introduced.

Added semantically correct iterators for indexes with descending parts (these are aliases actually):
```lua
box.index.FORWARD_INCLUSIVE = box.index.GE
box.index.FORWARD_EXCLUSIVE = box.index.GT
box.index.REVERSE_INCLUSIVE = box.index.LE
box.index.REVERSE_EXCLUSIVE = box.index.LT
```

In order to prevent code duplciation tuple_compare.cc was refactored: field comparsion code, which was highly copy-pasted around the file is moved to a separated templated function.

In order to assure the changes didn't break anything, new tests has been introduced. The tests cover all branches of all currently existing comparators. To simplify the tests the key_def.c test was switched to C++.

Closes #5529.

## Documentation

Sort order specifies the way indexes iterate over tuples with different fields in the same part. It can be either ascending (which is the case by default) and descending.

Tuples with different ascending parts are sorted in indexes from lesser to greater, whereas tuples with different descending parts are sorted in the opposte order: from greater to lesser.

Given example:

```lua
box.cfg{}

s = box.schema.create_space('tester')
pk = s:create_index('pk', {parts = {
  {1, 'unsigned', sort_order = 'desc'},
  {2, 'unsigned', sort_order = 'asc'},
  {3, 'unsigned', sort_order = 'desc'},
}})

s:insert({1, 1, 1})
s:insert({1, 1, 2})
s:insert({1, 2, 1})
s:insert({1, 2, 2})
s:insert({2, 1, 1})
s:insert({2, 1, 2})
s:insert({2, 2, 1})
s:insert({2, 2, 2})
s:insert({3, 1, 1})
s:insert({3, 1, 2})
s:insert({3, 2, 1})
s:insert({3, 2, 2})
```

In this case field 1 and 3 are descending, whereas field 2 is ascending. So `s:select()` will return this result:

```yaml
---
- [3, 1, 2]
- [3, 1, 1]
- [3, 2, 2]
- [3, 2, 1]
- [2, 1, 2]
- [2, 1, 1]
- [2, 2, 2]
- [2, 2, 1]
- [1, 1, 2]
- [1, 1, 1]
- [1, 2, 2]
- [1, 2, 1]
...
```

Beware, that when using other sort order than 'asc' for any field 'LE', 'LT', 'GE' and 'GT' iterator lose their meaning and specify 'opposite including', 'opposite excluding', 'direct including' and 'direct excluding' iteration direction respectively. Given example above, `s:select({2}, {iterator = 'GT'})` will return this:

```yaml
---
- [1, 1, 2]
- [1, 1, 1]
- [1, 2, 2]
- [1, 2, 1]
...
```

And `s:select({1}, {iterator = 'LT'})` will give us:

```yaml
---
- [2, 2, 1]
- [2, 2, 2]
- [2, 1, 1]
- [2, 1, 2]
- [3, 2, 1]
- [3, 2, 2]
- [3, 1, 1]
- [3, 1, 2]
...
```

In order to be more clear alternative iterator aliases can be used: `FORWARD_INCLUSIVE`, `FORWARD_EXCLUSIVE`, `REVERSE_INCLUSIVE`, `REVERSE_EXCLUSIVE`:

```
> s:select({1}, {iterator = 'REVERSE_EXCLUSIVE'})
---
- [2, 2, 1]
- [2, 2, 2]
- [2, 1, 1]
- [2, 1, 2]
- [3, 2, 1]
- [3, 2, 2]
- [3, 1, 1]
- [3, 1, 2]
...
```

### DIFF
--- a/changelogs/unreleased/gh-5529-key-part-sort-order.md
+++ b/changelogs/unreleased/gh-5529-key-part-sort-order.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Now it is possible to specify the sort order for each part of an index
+  (gh-5529).
--- a/src/box/key_def.h
+++ b/src/box/key_def.h
@@ -64,7 +64,12 @@ struct key_part_def {
 	bool is_nullable;
 	/** Action to perform if NULL constraint failed. */
 	enum on_conflict_action nullable_action;
-	/** Part sort order. */
+	/**
+	 * Part sort order. key_part_def can have "undef" sort order, but when
+	 * key_part is created, the "undef" sort order is translated to "asc",
+	 * so, unlike key_part_def, only "asc" and "desc" values are possible
+	 * in the key_part.
+	 */
 	enum sort_order sort_order;
 	/**
 	 * JSON path to indexed data, relative to the field number,
@@ -212,6 +217,8 @@ struct key_def {
 	bool is_multikey;
 	/** True if it is a functional index key definition. */
 	bool for_func_index;
+	/** True if it is unordered index key definition. */
+	bool is_unordered;
 	/**
 	 * True, if some key parts can be absent in a tuple. These
 	 * fields assumed to be MP_NIL.
@@ -651,13 +658,18 @@ key_def_sizeof(uint32_t part_count, uint32_t path_pool_size)
 	       path_pool_size;
 }
 
+enum key_def_new_flags {
+	KEY_DEF_FOR_FUNC_INDEX = 1 << 0,
+	KEY_DEF_UNORDERED = 1 << 1,
+};
+
 /**
  * Allocate a new key_def with the given part count
  * and initialize its parts.
  */
 struct key_def *
 key_def_new(const struct key_part_def *parts, uint32_t part_count,
-	    bool for_func_index);
+	    unsigned flags);
 
 /**
  * Dump part definitions of the given key def.
@@ -812,6 +824,22 @@ key_def_has_collation(const struct key_def *key_def)
 {
 	for (uint32_t part_id = 0; part_id < key_def->part_count; part_id++) {
 		if (key_def->parts[part_id].coll != NULL)
+			return true;
+	}
+	return false;
+}
+
+/**
+ * Return true if any part of @a key_def is descending.
+ * @param key_def key_def
+ * @retval true if the key_def has descending parts
+ * @retval false otherwise
+ */
+static inline bool
+key_def_has_desc_parts(struct key_def *key_def)
+{
+	for (uint32_t part_id = 0; part_id < key_def->part_count; part_id++) {
+		if (key_def->parts[part_id].sort_order == SORT_ORDER_DESC)
 			return true;
 	}
 	return false;
--- a/src/box/key_def.h
+++ b/src/box/key_def.h
@@ -314,6 +314,7 @@ typedef struct key_def box_key_def_t;
 enum {
 	BOX_KEY_PART_DEF_IS_NULLABLE = 1 << 0,
 	BOX_KEY_PART_DEF_EXCLUDE_NULL = 1 << 1,
+	BOX_KEY_PART_DEF_SORT_ORDER_DESC = 1 << 2,
 };
 
 /**
@@ -437,10 +438,11 @@ box_key_def_new(uint32_t *fields, uint32_t *types, uint32_t part_count);
  *
  * Default flag values are the following:
  *
- *  | Flag                          | Default value |
- *  | ----------------------------- | ------------- |
- *  | BOX_KEY_PART_DEF_IS_NULLABLE  | 0 (unset)     |
- *  | BOX_KEY_PART_DEF_EXCLUDE_NULL | 0 (unset)     |
+ *  | Flag                             | Default value |
+ *  | -------------------------------- | ------------- |
+ *  | BOX_KEY_PART_DEF_IS_NULLABLE     | 0 (unset)     |
+ *  | BOX_KEY_PART_DEF_EXCLUDE_NULL    | 0 (unset)     |
+ *  | BOX_KEY_PART_DEF_SORT_ORDER_DESC | 0 (unset)     |
  *
  * Default values of fields and flags are permitted to be changed
  * in future tarantool versions. However we should be VERY
--- a/src/box/lua/key_def.c
+++ b/src/box/lua/key_def.c
@@ -85,6 +85,11 @@ luaT_push_key_def_parts(struct lua_State *L, const struct key_def *key_def)
 		lua_pushnumber(L, part->fieldno + TUPLE_INDEX_BASE);
 		lua_setfield(L, -2, "fieldno");
 
+		if (!key_def->is_unordered) {
+			lua_pushstring(L, sort_order_strs[part->sort_order]);
+			lua_setfield(L, -2, "sort_order");
+		}
+
 		if (part->path != NULL) {
 			lua_pushlstring(L, part->path, part->path_len);
 			lua_setfield(L, -2, "path");
@@ -227,6 +232,22 @@ luaT_key_def_set_part(struct lua_State *L, struct key_part_def *part,
 			return -1;
 		}
 		part->coll_id = coll_id->id;
+	}
+	lua_pop(L, 1);
+
+	/* Set part->sort_order. */
+	lua_pushstring(L, "sort_order");
+	lua_gettable(L, -2);
+	if (!lua_isnil(L, -1)) {
+		size_t sort_order_len;
+		const char *sort_order = lua_tolstring(L, -1, &sort_order_len);
+		part->sort_order = strnindex(sort_order_strs, sort_order,
+					     sort_order_len, sort_order_MAX);
+		if (part->sort_order == sort_order_MAX) {
+			diag_set(IllegalParams, "Unknown sort order: \"%s\"",
+				 sort_order);
+			return -1;
+		}
 	}
 	lua_pop(L, 1);
 
@@ -486,6 +507,7 @@ lbox_key_def_new(struct lua_State *L)
 				  "{fieldno = fieldno, type = type"
 				  "[, is_nullable = <boolean>]"
 				  "[, exclude_null = <boolean>]"
+				  "[, sort_order = <string>]"
 				  "[, path = <string>]"
 				  "[, collation_id = <number>]"
 				  "[, collation = <string>]}, ...}");
--- a/src/box/lua/key_def.c
+++ b/src/box/lua/key_def.c
@@ -517,7 +517,7 @@ lbox_key_def_new(struct lua_State *L)
 		lua_pop(L, 1);
 	}
 
-	struct key_def *key_def = key_def_new(parts, part_count, false);
+	struct key_def *key_def = key_def_new(parts, part_count, 0);
 	region_truncate(region, region_svp);
 	if (key_def == NULL)
 		return luaT_error(L);
--- a/src/box/sql.c
+++ b/src/box/sql.c
@@ -393,7 +393,7 @@ sql_ephemeral_space_new(const struct sql_space_info *info)
 		parts[i].coll_id = info->coll_ids[j];
 	}
 
-	struct key_def *key_def = key_def_new(parts, part_count, false);
+	struct key_def *key_def = key_def_new(parts, part_count, 0);
 	if (key_def == NULL)
 		return NULL;
 
--- a/src/box/tuple_compare.cc
+++ b/src/box/tuple_compare.cc
@@ -573,6 +573,25 @@ tuple_compare_field_with_type(const char *field_a, enum mp_type a_type,
 }
 
 /*
+ * Reverse the compare result if the key part sort order is descending.
+ */
+static inline int
+key_part_compare_result(struct key_part *part, int result)
+{
+	return result * (part->sort_order != SORT_ORDER_DESC ? 1 : -1);
+}
+
+/*
+ * Reverse the hint if the key part sort order is descending.
+ */
+static hint_t
+key_part_hint(struct key_part *part, hint_t hint)
+{
+	/* HINT_MAX - HINT_NONE underflows to HINT_NONE. */
+	return part->sort_order != SORT_ORDER_DESC ? hint : HINT_MAX - hint;
+}
+
+/*
  * Implements the field comparison logic. If the key we use is not nullable
  * then a simple call to tuple_compare_field is used.
  *
@@ -606,7 +625,7 @@ key_part_compare_fields(struct key_part *part, const char *field_a,
 	if (!is_nullable) {
 		rc = tuple_compare_field(field_a, field_b,
 					 part->type, part->coll);
-		return rc;
+		return key_part_compare_result(part, rc);
 	}
 	enum mp_type a_type = (a_is_optional && field_a == NULL) ?
 			      MP_NIL : mp_typeof(*field_a);
@@ -623,7 +642,7 @@ key_part_compare_fields(struct key_part *part, const char *field_a,
 						   field_b, b_type,
 						   part->type, part->coll);
 	}
-	return rc;
+	return key_part_compare_result(part, rc);
 }
 
 template<bool is_nullable, bool has_optional_parts, bool has_json_paths,
@@ -1943,26 +1962,27 @@ field_hint(const char *field, struct coll *coll)
 	return HINT_NONE;
 }
 
-template <enum field_type type, bool is_nullable>
+template<enum field_type type, bool is_nullable>
 static hint_t
 key_hint(const char *key, uint32_t part_count, struct key_def *key_def)
 {
 	assert(!key_def->is_multikey);
 	if (part_count == 0)
 		return HINT_NONE;
-	return field_hint<type, is_nullable>(key, key_def->parts->coll);
+	hint_t h = field_hint<type, is_nullable>(key, key_def->parts->coll);
+	return key_part_hint(key_def->parts, h);
 }
 
-template <enum field_type type, bool is_nullable>
+template<enum field_type type, bool is_nullable>
 static hint_t
 tuple_hint(struct tuple *tuple, struct key_def *key_def)
 {
 	assert(!key_def->is_multikey);
 	const char *field = tuple_field_by_part(tuple, key_def->parts,
 						MULTIKEY_NONE);
-	if (is_nullable && field == NULL)
-		return hint_nil();
-	return field_hint<type, is_nullable>(field, key_def->parts->coll);
+	hint_t h = is_nullable && field == NULL ? hint_nil() :
+		   field_hint<type, is_nullable>(field, key_def->parts->coll);
+	return key_part_hint(key_def->parts, h);
 }
 
 static hint_t
@@ -1986,7 +2006,7 @@ key_hint_stub(const char *key, uint32_t part_count, struct key_def *key_def)
 }
 
 static hint_t
-key_hint_stub(struct tuple *tuple, struct key_def *key_def)
+tuple_hint_stub(struct tuple *tuple, struct key_def *key_def)
 {
 	(void) tuple;
 	(void) key_def;
@@ -2017,7 +2037,7 @@ key_def_set_hint_func(struct key_def *def)
 {
 	if (def->is_multikey || def->for_func_index) {
 		def->key_hint = key_hint_stub;
-		def->tuple_hint = key_hint_stub;
+		def->tuple_hint = tuple_hint_stub;
 		return;
 	}
 	switch (def->parts->type) {
@@ -2071,6 +2091,7 @@ key_def_set_compare_func_fast(struct key_def *def)
 	assert(!def->has_optional_parts);
 	assert(!def->has_json_paths);
 	assert(!key_def_has_collation(def));
+	assert(!key_def_has_desc_parts(def));
 
 	tuple_compare_t cmp = NULL;
 	tuple_compare_with_key_t cmp_wk = NULL;
@@ -2172,8 +2193,8 @@ key_def_set_compare_func(struct key_def *def)
 			key_def_set_compare_func_for_func_index<true>(def);
 		else
 			key_def_set_compare_func_for_func_index<false>(def);
-	} else if (!key_def_has_collation(def) &&
-	    !def->is_nullable && !def->has_json_paths) {
+	} else if (!def->is_nullable && !def->has_json_paths &&
+	    !key_def_has_collation(def) && !key_def_has_desc_parts(def)) {
 		key_def_set_compare_func_fast(def);
 	} else if (!def->has_json_paths) {
 		if (def->is_nullable && def->has_optional_parts) {
--- a/src/box/tuple_compare.cc
+++ b/src/box/tuple_compare.cc
@@ -572,6 +572,60 @@ tuple_compare_field_with_type(const char *field_a, enum mp_type a_type,
 	}
 }
 
+/*
+ * Implements the field comparison logic. If the key we use is not nullable
+ * then a simple call to tuple_compare_field is used.
+ *
+ * Otherwise one of \p field_a and \p field_b can be NIL: either it's encoded
+ * as MP_NIL or if the field is absent (corresponding field pointer equals to
+ * NULL). In this case we perform comparison so that NIL is lesser than any
+ * value but two NILs are equal.
+ *
+ * The template parameters a_is_optional and b_is_optional specify if the
+ * corresponding field arguments can be absent (equal to NULL).
+ *
+ * If \p was_null_met is not NULL, sets the boolean pointed by it to true if
+ * any of \p field_a and \p field_b is absent or NIL. Othervice the pointed
+ * value is not modified.
+ *
+ * @param part the key part we compare
+ * @param field_a the field to compare
+ * @param field_b the field to compare against
+ * @param was_null_met pointer to the value to set to true if a NIL is met,
+ *                     can be set to NULL if the information isn't required
+ * @retval 0  if field_a == field_b
+ * @retval <0 if field_a < field_b
+ * @retval >0 if field_a > field_b
+ */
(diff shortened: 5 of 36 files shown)


changelogs/unreleased/gh-5529-key-part-sort-order.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
+
+* It's possible to specify the sort order of each index part now (gh-5529).
```
xuniq, line 3:
```suggestion
* Now it is possible to specify the sort order for each part of the index (gh-5529).
```

xuniq (approved):

ochaplashkin (approved):
LGTM :rocket:

test/app-tap/module_api.c
```diff
@@ -1711,6 +1715,83 @@ test_key_def_merge(struct lua_State *L)
...
+		{{1, 4, "unsigned", NULL, NULL}}, /* clash */
+		{{2, 0, "unsigned", NULL, NULL}},
+	};
+	box_key_part_def_t exp_28[] = {
+		{{3, 0, "unsigned", NULL, NULL}},
+		{{1, 0, "unsigned", NULL, NULL}}, /* coalesced */
+		{{2, 0, "unsigned", NULL, NULL}},
+	};
+	key_def_check_merge(a_28, lengthof(a_28), b_28, lengthof(b_28),
+			    exp_28, lengthof(exp_28));
```
Totktonada, line 1737: There is a comment (written by me) with attempt to formalize invariants of the `box_key_def_merge()` function. Quoted below.

> For any given kd1 and kd2, kd3 = merge(kd1, kd2) should impose the same order of tuples as if they would be ordered by kd1, but all kd1-equal tuples would be ordered by kd2.

(It is in the beginning of the `test_key_def_merge()` function.)

It seems, this comment is incorrect now. I think I know how to fix it:

> <...> but all kd1-equal tuples **that are not kd2-equal** would be ordered by kd2.

If you have a better description of merge's invariants or better know reasons behind them, I would appreciate if you enhance my description. At now, it is based on observations. It was written without good understanding of internals (such as comparators, validators and key extractors), but, as I see, you actively work on this code and likely know more than me.

----

Your interpretation of the test cases is a reference to the case 3, which is technically applicable to cases 28-29, but IMHO it doesn't really describes why the invariants are hold.

I think that if you describe why the fields are coalesced in these test cases in your words, it would be much better.

test/app-tap/module_api.c
```diff
@@ -1711,6 +1715,83 @@ test_key_def_merge(struct lua_State *L)
...
+		{{2, 0, "unsigned", NULL,      NULL}},
+	};
+	box_key_part_def_t exp_30[] = {
+		{{3, 0, "unsigned", NULL,      NULL}},
+		{{1, 0, "unsigned", "unicode", NULL}}, /* from <a> */
+		{{1, 4, "unsigned", "binary",  NULL}}, /* from <b> */
+		{{2, 0, "unsigned", NULL,      NULL}},
+	};
+	key_def_check_merge(a_30, lengthof(a_30), b_30, lengthof(b_30),
+			    exp_30, lengthof(exp_30));
```
Totktonada, line 1793: What is the idea behind this test case?

I mean, a test case is usually written with idea to verify some group of similar situations: they are either handled by the same code (if-branch) or they're similar logically.

It is not clear for me, what is the group that is assumed here.

I suppose that the idea is to verify that the logic that is described in the interpretation of the case 8 works for key parts with different sort order. Is it correct?

Or there is some logic that is specific for situations, when there are collations and the sort orders are different?

Feel free to ignore this comment. It is mostly for my understanding.

test/app-tap/module_api.c
```diff
@@ -536,7 +536,9 @@ key_part_padding_offset(void)
 static uint32_t
```
Totktonada, line 536: Shouldn't we verify something about `box_tuple_compare()` and `box_tuple_compare_with_key()`? AFAIU, their logic depends on the flag.

src/box/lua/key_def.c
```diff
@@ -221,6 +226,24 @@ luaT_key_def_set_part(struct lua_State *L, struct key_part_def *part,
...
+	lua_pushstring(L, "sort_order");
+	lua_gettable(L, -2);
+	if (!lua_isnil(L, -1)) {
+		size_t sort_order_len;
+		const char *sort_order = lua_tolstring(L, -1, &sort_order_len);
+		part->sort_order = strnindex(sort_order_strs, sort_order,
+					     sort_order_len, sort_order_MAX);
+		assert(part->sort_order != SORT_ORDER_UNDEF);
+		if (part->sort_order == sort_order_MAX) {
+			diag_set(IllegalParams, "unknown sort order");
```
Totktonada, line 239: Nit: It is good to show a content of a wrong parameter in the error message, just like a compiler shows a line in the code that has a problem. It is not essential, but reduces a time needed to understand what is the problem.

See error messages for `type` and `collation` fields above for example.

test/box-luatest/gh_5529_sort_order_test.lua
```diff
@@ -178,3 +188,237 @@ g1.test_sequential = function(cg)
...
+            -- The key_def.field_is_indexed_by_part should be filled earlier.
+            t.assert(key_def.field_is_indexed_by_part ~= nil)
+            t.assert(is_number(fieldno))
+            t.assert(fieldno > 0)
+            return key_def.field_is_indexed_by_part[fieldno]
+        end
+
+        -- Generates the values that are possible for this field according to
+        -- the key_def given. The values are returned in an array and sorted
+        -- in ascending order.
```
Totktonada, line 231: > <...> sorted in ascending order.

As I see, they're sorted according to the sort order in the given key part. Is it correct?

Totktonada (approved):
I have no objections regarding key_def's public APIs (C and Lua).

test/box-luatest/gh_5529_sort_order_test.lua
mkostoevr: Yes, fixed the comment.

test/app-tap/module_api.c
mkostoevr: Yes, the case is technically identical to the case 17, it just assures that the merged key part preserves its sort order.

Updated the case comment.

test/app-tap/module_api.c
mkostoevr: Updated the test comment and slightly expanded your description, but it seems to me it's still correct: even if the equal-by-kd1 tuples are also equal-by-kd2, the tuples inside groups of kd-1-equal tuples are ordered by the kd2, even though the order is not changed.

alyapunov (approved):

src/box/tuple_compare.cc
```diff
@@ -572,6 +572,60 @@ tuple_compare_field_with_type(const char *field_a, enum mp_type a_type,
 	}
 }
 
+/*
+ * Implements the field comparison logic. If the key we use is not nullable
+ * then a simple call to tuple_compare_field is used.
+ *
+ * Othervice one of \p field_a and \p field_b can be NIL: either it's encoded
```
locker, line 597: Typo: otherwice -> otherwise. Also, there's a typo in the commit log: siplify -> simplify

src/box/key_def.h
```diff
@@ -47,9 +47,9 @@ extern "C" {
 extern const char *sort_order_strs[];
 
 enum sort_order {
-	SORT_ORDER_ASC = 0,
-	SORT_ORDER_DESC,
-	SORT_ORDER_UNDEF,
+	SORT_ORDER_UNDEF = -1,  /* There's no strings for negative values. */
```
locker, line 50: Why do you need this constant (`SORT_ORDER_UNDEF`)? Let's drop ti as well.

src/box/sql.c
```diff
@@ -283,7 +283,7 @@ sql_space_info_new(uint32_t field_count, uint32_t part_count)
 	info->parts = part_count == 0 ? NULL :
 		      (uint32_t *)((char *)info->coll_ids + colls_size);
 	info->sort_orders = part_count == 0 ? NULL :
-			    (uint32_t *)((char *)info->parts + parts_size);
+			    (int32_t *)((char *)info->parts + parts_size);
```
locker, line 286: Nit: Should be `(enum sort_order *)`?

test/box-luatest/gh_5529_sort_order_test.lua
```diff
@@ -0,0 +1,180 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g1 = t.group('sort_order_test', {
+    {engine = 'memtx', is_unique = true},
+    {engine = 'memtx', is_unique = false},
+    {engine = 'vinyl', is_unique = true},
+    {engine = 'vinyl', is_unique = false}
+})
```
locker, line 9: The test should be moved to `engine-luatest` since it's multi-engine.

Also, please add more test cases:
- Errors raised when the sort order is unsupported.
- Snapshot and recovery.
- Reversing the sort order while altering a space.

test/vinyl/ddl.result
```diff
@@ -571,6 +571,7 @@ box.space.test.index.pk
 - unique: true
   parts:
   - type: unsigned
+    sort_order: asc
```
locker, line 574: This is going to break the ddl integration test, see #8994. @Gumix should be working on the fix. Please talk to him to ensure that his fix will cover your case as well.

src/box/key_def.h
```diff
@@ -307,6 +307,9 @@ typedef struct key_def box_key_def_t;
 enum {
 	BOX_KEY_PART_DEF_IS_NULLABLE = 1 << 0,
 	BOX_KEY_PART_DEF_EXCLUDE_NULL = 1 << 1,
+	BOX_KEY_PART_DEF_SORT_ORDER = 1 << 2,
+	BOX_KEY_PART_DEF_SORT_ORDER_ASC = 0,
```
locker, line 311: I don't understand why we need this empty flag. Why not add one flag `BOX_KEY_PART_DEF_IS_REVERSE`? Come to think, is there any point in having a enum in `key_part` for describing the sort order? Wouldn't a boolean flag (`is_reverse`) suffice? Can there be any other sort order except "ascending" and "descending"?

src/box/lua/key_def.c
```diff
@@ -85,6 +85,11 @@ luaT_push_key_def_parts(struct lua_State *L, const struct key_def *key_def)
 		lua_pushnumber(L, part->fieldno + TUPLE_INDEX_BASE);
 		lua_setfield(L, -2, "fieldno");
 
+		assert(part->sort_order == SORT_ORDER_ASC ||
+		       part->sort_order == SORT_ORDER_DESC);
+		lua_pushstring(L, sort_order_strs[part->sort_order]);
+		lua_setfield(L, -2, "sort_order");
```
locker, line 91: The sort order shouldn't be reported as "asc" for a hash index.

```
tarantool> _ = box.schema.create_space('test')
---
...

tarantool> box.space.test:create_index('pk', {type = 'hash'})
---
- unique: true
  parts:
  - type: unsigned
    sort_order: asc
    is_nullable: false
    fieldno: 1
  id: 0
  space_id: 512
  type: HASH
  name: pk
...
```

test/app-tap/module_api.c
Totktonada: Thank you!

Two nits regarding the new test:

* I would check results for `< 0`, `> 0` instead of equivalence to `-1`, `1`.
* I would call `box_tuple_unref()` at the end.

src/box/lua/key_def.c
mkostoevr: Done.

src/box/tuple_compare.cc
```diff
@@ -572,6 +572,24 @@ tuple_compare_field_with_type(const char *field_a, enum mp_type a_type,
...
+	return result * (part->sort_order == SORT_ORDER_ASC ? 1 : -1);
+}
+
+/*
+ * Reverse the hint if the key part sort order is descending.
+ */
+static hint_t
+key_part_hint(struct key_part *part, hint_t hint)
+{
+	return part->sort_order == SORT_ORDER_ASC ? hint : HINT_NONE - hint;
```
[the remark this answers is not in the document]
mkostoevr: Fixed.

test/box-luatest/gh_5529_sort_order_test.lua
mkostoevr: Done.

src/box/sql.c
mkostoevr: The change has been dropped from the patchset.

src/box/key_def.h
mkostoevr: Discussed offile: it's secided to keep it.

src/box/tuple_compare.cc
mkostoevr: Fixed.

src/box/key_def.h
Totktonada: The flag is de-facto boolean, but it is represented in the structure as part of the `flags` field to consume one bit instead of one byte. This economy is due to the frozen size of the public structure and a wish to have the ability to add flags/fields for a long time without breaking the binary layout.

Regarding the name of the flag: it looks natural to name it with words similar to ones from the Lua API, hence 'sort order'.

I agree that the idea with two additional constants for ascending and descending orders is controversial. One should be OR'ed, another should be AND'ed, we anyway need to know the values and the abstraction is leaking.

How about one constant for the mask: BOX_KEY_PART_DEF_SORT_ORDER_DESC? It is quite long, sure, but that's mostly due to namespacing.

BOX_KEY_PART_DEF_IS_DESCENDING and BOX_KEY_PART_DEF_IS_DESC are OK too.

All the options have pros and cons, so I'll accept any of them.

test/vinyl/ddl.result
mkostoevr: The change has been dropped from the patchset.

src/box/key_def.h
mkostoevr: Dropped the `BOX_KEY_PART_DEF_SORT_ORDER_ASC` and called the mask `BOX_KEY_PART_DEF_SORT_ORDER_DESC`.

test/app-tap/module_api.c
mkostoevr: Thanks, I forgot about the `box_tuple_unref`.

Done.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1307 @@ test_tuple_validate_key_parts_raw(void)
...
+	key_def_delete(key_def);
+
+	free((void *)funcname);
+
+	footer();
+	check_plan();
+}
+
+static void
+test_tuple_compare_with_key_slowpath(bool is_nullable, bool has_optional_parts)
```
drewdzzz, line 816: What's about json path and multikey?
The same tip about tuple_compare test.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1307 @@ test_tuple_validate_key_parts_raw(void)
...
+
+	/* Type is number to prevent from using precompiled comparators. */
+	struct key_def *key_def = test_key_def_new(
+		"[{%s%u%s%s%s%b}]", "field", 1, "type", "number",
+		"is_nullable", is_nullable_and_has_optional_parts);
+
+	/* Update has_optional_parts if the last parts can't be nil. */
+	size_t min_field_count = tuple_format_min_field_count(&key_def, 1,
+							      NULL, 0);
+	key_def_update_optionality(key_def, min_field_count);
```
drewdzzz, line 735: Nit: I would check with `fail_unless` function that `key_def` has `has_optional_parts` attribute (or does not, if it is not expected). Just to be sure. This goes for all tests with `has_optional_parts` argument.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1307 @@ test_tuple_validate_key_parts_raw(void)
...
+
+static void
+test_tuple_compare_with_key_slowpath(bool is_nullable, bool has_optional_parts)
+{
+	size_t p = 14 + (is_nullable ? 10 : 0) + (has_optional_parts ? 40 : 0);
+	plan(p);
+	header();
+
+	/* has_optional_parts is only valid if is_nullable. */
+	assert(!has_optional_parts || is_nullable);
```
drewdzzz, line 823: Nit: We usually use `fail_if` and `fail_unless` functions to check invariants in unit tests.

drewdzzz (commented):
The test is generally OK, but it's really large (3-6 my screens per test case), and all the tests are generic. I think we could make it much more compact - we could factor out tuple generation and comparation in a separate function. This function will accept a pointer to `key_def` object, then it will inspect it (for `is_nullable`, `is_unique` and `has_optional_parts` properties) and generate `tuples_eq` and `tuples_gt` vectors for them.  For simplicity, every test can use `key_def`s of the same size (4, for example) except for singlepart tests. Then, tuple generation logic will be located at one place, and all the tests will be compact - key_def generation and call of a generic function.

test/unit/key_def.cc
mkostoevr: The idea of the test is to cover each parameter that affects the comparison logic, the `has_json_paths` and `is_multikey` parameters don't do it (except for not using the compare hint in case of multikey JSON index).

BTW, maybe it will be good to compare with and without hints and except the same result.

drewdzzz: 1. You should write the same in commit message.
2. Yep, it would be nice. Hint comparation logic is trivial, but let's do it for greater reliability.

test/unit/key_def.cc
mkostoevr: Added the checks.

test/unit/key_def.cc
mkostoevr: Replaced the asserts with `fail_unless`.

test/unit/key_def.cc
mkostoevr: Updated the commit message and added hint vs nohint comparison check.

Since the key extraction and the key hint validation is off-topic, I suggest to implement it in another PR if required.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1330 @@ test_tuple_validate_key_parts_raw(void)
...
+	key_def_delete(key_def);
+
+	free((void *)funcname);
+
+	footer();
+	check_plan();
+}
+
+static void
+test_tuple_compare_with_key(struct tuple *tuple_a, struct tuple *tuple_b,
```
drewdzzz, line 713: Nit: it would be nice to call such functions like `test_check_tuple_compare_with_key` to easily distinguish a test case from the function that checks a function.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1330 @@ test_tuple_validate_key_parts_raw(void)
...
+
+	struct key_def *cmp_def = key_def_merge(key_def, pk_def);
+
+	if (is_unique) {
+		/*
+		 * It's assumed that PK and SK index different parts. So we
+		 * cover cmp_def->unique_part_count < cmp_def->part_count
+		 * branch of the slowpath comparator (its last loop).
+		 */
+		fail_unless(cmp_def->unique_part_count > key_def->part_count);
```
drewdzzz, line 1058: It seems to me that this check should be outside the `if` closure.

test/unit/key_def.cc
```diff
@@ -457,16 +460,1330 @@ test_tuple_validate_key_parts_raw(void)
...
+			tuples_eq.push_back(
+				test_tuple_new("[%u%u%u%u]", 0, 0, 0, 0));
+		} else {
+			/*
+			 * FIXME: tuple_compare_slowpath has a logic I don't
+			 * quite understand. If the tuples are equal by SK and
+			 * we have no nils met, we should skip the PK comparison
+			 * and conclude the tuples are equal, but the comparator
+			 * has this `!is_nullable` condition making it compare
+			 * all parts of the key (including PK).
```
drewdzzz, line 1112:
```c
if (!is_nullable || !was_null_met)
	return 0;
```

If `NULL` was not met, we must compare SK only. But if the index is not nullable, `was_null_met` is definitely `false`, so this check of template parameter is done for the sake of optimization. I don't see any problem in this logic.

drewdzzz: Wow, I see the problem in the beginning of the comparator:
```c
if (is_nullable)
	end = part + key_def->unique_part_count;
else
	end = part + key_def->part_count;
```
It's really weird. Did you try to always use unique_part_count?

test/unit/key_def.cc
```diff
@@ -457,16 +460,1330 @@ test_tuple_validate_key_parts_raw(void)
...
+	struct key_def *cmp_def = key_def_merge(key_def, pk_def);
+
+	/* See index_def_new(). */
+	if (is_unique) {
+		/*
+		 * It's assumed that PK and SK index different parts. So we
+		 * cover cmp_def->unique_part_count < cmp_def->part_count
+		 * branch of the sequential comparator (its last loop).
+		 */
+		fail_unless(cmp_def->unique_part_count > key_def->part_count);
```
drewdzzz, line 1483: Ditto.

test/unit/key_def.cc
mkostoevr: Yes, it makes the vinyl MVCC fell bad in the `vinyl/iterator.test.lua` test.

Haven't digged into that yet.

test/unit/key_def.cc
mkostoevr: Moved.

test/unit/key_def.cc
mkostoevr: Moved.

test/unit/key_def.cc
mkostoevr: Renamed.

test/unit/key_def.cc
mkostoevr: Just realized it's not a vinyl problem, but a memtx one: it has two tuples with the same value of unique second field in the output.

drewdzzz (approved):
Key def tests LGTM.

mkostoevr:
The ddl integration fixes are in the tarantool/ddl#115.

test/unit/key_def.cc
```diff
@@ -457,16 +460,875 @@ test_tuple_validate_key_parts_raw(void)
...
+key_def_field_is_indexed(const struct key_def *key_def, uint32_t fieldno)
+{
+	for (uint32_t i = 0; i < key_def->part_count; i++)
+		if (key_def->parts[i].fieldno == fieldno)
+			return true;
+	return false;
+}
+
+static void
+test_create_common_cases(std::vector<struct tuple *> &tuples_eq,
```
drewdzzz, line 488: I'd call it like `test_generate_common_cases`. Also, it would be nice to cover it with a comment - this function inspects passed `key_def` and generates tuples to test it, but it does not generate tuples for `key_def`'s of unique indexes because of inconsistent behavior and so on.

test/unit/key_def.cc
```diff
@@ -457,16 +460,875 @@ test_tuple_validate_key_parts_raw(void)
...
+	const bool last_2_are_nullable =
+		field_2_is_nullable && field_3_is_nullable;
+
+	const bool field_0_is_indexed = key_def_field_is_indexed(key_def, 0);
+	const bool field_1_is_indexed = key_def_field_is_indexed(key_def, 1);
+	const bool field_2_is_indexed = key_def_field_is_indexed(key_def, 2);
+	const bool field_3_is_indexed = key_def_field_is_indexed(key_def, 3);
+
+	const bool last_2_are_indexed =
+		field_1_is_indexed && field_2_is_indexed;
```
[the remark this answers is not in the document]
mkostoevr: Dropped the variable as unused.

test/unit/key_def.cc
mkostoevr: Renamed and commented.

test/vinyl/ddl.result
mkostoevr: Fix is merged with in tarantool/ddl@7ee51b5543bed7f25ac3a314368c238401df4013.

test/unit/key_def.cc
```diff
@@ -457,16 +460,892 @@ test_tuple_validate_key_parts_raw(void)
...
+	fail_unless(cmp_def->is_nullable == is_nullable);
+	fail_unless(!cmp_def->has_optional_parts);
+
+	const char *funcname = (const char *)xstrdup(tt_sprintf(
+		"tuple_compare_sequential<%s, false, key_def is unique>",
+		is_nullable ? "true" : "false"));
+
+	/* Regular cases are tested in test_tuple_compare_sequential. */
+	std::vector<struct tuple *> tuples_eq;
+	std::vector<struct tuple *> tuples_gt;
```
drewdzzz, line 1320: Why don't you generate common test cases here?

mkostoevr: The original problem was a key def difference, but it's not a problem after `test_generate_common_cases` refactoring.

Added the common cases testing.

drewdzzz (approved):
I have no objections regarding `unit/key_def.cc`.

Issue #8848, opened 2023-07-05
ASAN: heap-use-after-free in applier set_next_tx_row


nshy:
```console
Tarantool 3.0.0-entrypoint-256-g088b32f38
type 'help' for interactive help
```

Issue on ASAN run of replication tests with ASAN friendly small allocators.

These allocators are not yet in upsteam. It is WIP in scope of https://github.com/tarantool/tarantool/issues/7327. Temporarily available in https://github.com/nshy/tarantool/tree/small-asan. One need to disable WA in applier to get the repro.

```log
=================================================================
==179984==ERROR: AddressSanitizer: heap-use-after-free on address 0x6080000152e8 at pc 0x555ad074d65d bp 0x7f3d2932c370 sp 0x7f3d2932c368
READ of size 8 at 0x6080000152e8 thread T2
    #0 0x555ad074d65c in stailq_add_tail(stailq*, stailq_entry*) /home/shiny/dev/tarantool/src/lib/salad/stailq.h:124:10
    #1 0x555ad074d65c in set_next_tx_row(stailq*, applier_tx_row*, long) /home/shiny/dev/tarantool/src/box/applier.cc:1163:2
    #2 0x555ad074d65c in applier_read_tx(applier*, stailq*, applier_read_ctx const*, double) /home/shiny/dev/tarantool/src/box/applier.cc:1185:9
    #3 0x555ad0757e18 in applier_thread_reader_f(__va_list_tag*) /home/shiny/dev/tarantool/src/box/applier.cc:2034:4
    #4 0x555ad030d108 in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*) /home/shiny/dev/tarantool/src/lib/core/fiber.h:1234:10
    #5 0x555ad0a8af6f in fiber_loop /home/shiny/dev/tarantool/src/lib/core/fiber.c:1007:18
    #6 0x555ad140e47c in coro_init /home/shiny/dev/tarantool/third_party/coro/coro.c:108:3

0x6080000152e8 is located 72 bytes inside of 87-byte region [0x6080000152a0,0x6080000152f7)
freed by thread T2 here:
    #0 0x555ad02bdd92 in __interceptor_free.part.0 asan_malloc_linux.cpp.o
    #1 0x555ad13970b9 in lsregion_gc /home/shiny/dev/tarantool/src/lib/small/small/lsregion_malloc.c:79:3
[...]

PR #8916, opened 2023-07-26
applier: fix use after free

sergepetrenko:
Applier thread uses lsregion to allocate the messages for tx thread. The
messages are freed upon return to the applier thread using a
corresponding lsr_id.

Due to a typo, one of the lsregion allocations was made with a postfix
increment of lsr_id instead of the prefix one. Essentially, part of a
new message was allocated with an old lsr_id, and might be freed early
by a return of a previous message.

Fix this.

Closes https://github.com/tarantool/tarantool/issues/8848

NO_DOC=bugfix
NO_TEST=covered by asan in https://github.com/tarantool/tarantool/pull/8901
NO_CHANGELOG=bugfix

### DIFF
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -2023,7 +2023,7 @@ applier_thread_reader_f(va_list ap)
 				 TIMEOUT_INFINITY :
 				 replication_disconnect_timeout();
 		struct applier_tx *tx;
-		tx = lsregion_alloc_object(lsr, applier->thread.lsr_id++,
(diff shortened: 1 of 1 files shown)


Serpentian (approved):
LGTM. Thanks for the patch!

Gerold103 (approved):
Thanks for the patch!

sergepetrenko:
Cherry-picked to 2.11: 0c4293d75659d4ee7de987d4f44ef97508cafc96

PR #8917, opened 2023-07-26
config: add low priority env source

Totktonada:
The usual environment configuration source is useful for parametrized run:

```sh
TT_MEMTX_MEMORY=<...> tarantool --name <...> --config <...>
```

However, sometimes a user may need to set a default value, which doesn't rewrite one that is provided in the configuration. Now, it is possible to do using the environment variables with the `_DEFAULT` suffix.

```sh
TT_MEMTX_MEMORY_DEFAULT=<...> tarantool --name <...> --config <...>
```

This feature may be especially useful for wrappers that run tarantool internally with some default paths for data directories, socket files, pid file. We likely will use it in the `tt` tool.

Part of #8862

----

Tarantool EE part: https://github.com/tarantool/tarantool-ee/pull/504

### DIFF
--- a/changelogs/unreleased/config-low-priority-env-source.md
+++ b/changelogs/unreleased/config-low-priority-env-source.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Added a low priority environment configuration source, which looks into
+  `TT_*_DEFAULT` variables. It is useful to declare defaults (gh-8862).
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -110,10 +110,10 @@ function methods._initialize(self)
     -- priority. The menthal rule here is the following: values
     -- closer to the process are preferred: env first, then file,
     -- then etcd (if available).
-    self:_register_source(require('internal.config.source.env'))
+    self:_register_source(require('internal.config.source.env').new())
 
     if self._config_file ~= nil then
-        self:_register_source(require('internal.config.source.file'))
+        self:_register_source(require('internal.config.source.file').new())
     end
 
     self:_register_applier(require('internal.config.applier.mkdir'))
@@ -148,14 +148,14 @@ function methods._collect(self, opts)
         -- argument. The 'config' section of the config may
         -- contain a configuration needed for a source.
         if sync_source == source.name or sync_source == 'all' then
-            source.sync(self, iconfig)
+            source:sync(self, iconfig)
         end
 
         -- Validate configurations gathered from the sources.
         if source.type == 'instance' then
-            instance_config:validate(source.get())
+            instance_config:validate(source:get())
         elseif source.type == 'cluster' then
-            cluster_config:validate(source.get())
+            cluster_config:validate(source:get())
         else
             assert(false)
         end
@@ -173,12 +173,12 @@ function methods._collect(self, opts)
         -- accumulator is passed as the second.
         local source_iconfig
         if source.type == 'cluster' then
-            local source_cconfig = source.get()
+            local source_cconfig = source:get()
             cconfig = cluster_config:merge(source_cconfig, cconfig)
             source_iconfig = cluster_config:instantiate(cconfig,
                 self._instance_name)
         elseif source.type == 'instance' then
-            source_iconfig = source.get()
+            source_iconfig = source:get()
         else
             assert(false)
         end
@@ -186,7 +186,7 @@ function methods._collect(self, opts)
 
         -- If a source returns an empty table, mark it as ones
         -- that provide no data.
-        local has_data = next(source.get()) ~= nil
+        local has_data = next(source:get()) ~= nil
         table.insert(source_info, ('* %q [type: %s]%s'):format(source.name,
             source.type, has_data and '' or ' (no data)'))
     end
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -104,12 +104,14 @@ end
 
 function methods._initialize(self)
     -- The sources are synchronized in the order of registration:
-    -- env, file, etcd (the latter is present in Tarantool EE).
+    -- env, file, etcd (present in Tarantool EE), env for
+    -- defaults.
     --
     -- The configuration values from the first source has highest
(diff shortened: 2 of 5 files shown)


src/box/lua/config/source/env.lua
```diff
@@ -20,20 +26,23 @@ local function sync(_config_module, _iconfig)
...
+    return self.values
 end
 
-local function get()
-    return values
+local function new()
+    return setmetatable({
+        name = 'env',
+        type = 'instance',
+        values = {},
```
Totktonada, line 57: I would name it `_values` to highlight that this is considered as a private field and to encourage the outer code to call `:get()`. The same for the file source.

test/config-luatest/sources_test.lua
```diff
@@ -69,6 +71,104 @@ g.test_source_env = function()
...
+        env = {
+            TT_PROCESS_TITLE_DEFAULT = 'from env default',
+            TT_LOG_FILE = 'from env',
+            TT_LOG_PIPE = 'from env',
+            TT_LOG_PIPE_DEFAULT = 'from env default',
+            TT_LOG_SYSLOG_IDENTITY_DEFAULT = 'from env default',
+        },
+    }
+    g.server = server:new(fun.chain(opts, {alias = 'instance-001'}):tomap())
+    g.server:start()
```
Totktonada, line 166: Needs `g.server:stop()` in an after_test/after_each hook to stop the tarantool process if the test case fails.

src/box/lua/config/source/env.lua
Totktonada: Done for the both sources. Done for the etcd source in the EE part.

test/config-luatest/sources_test.lua
Totktonada: Added the `after_each()` hook.

src/box/lua/config/source/env.lua
```diff
@@ -35,11 +43,20 @@ function methods.get(self)
...
+local function new(opts)
+    local opts = opts or {}
+    local env_var_suffix = opts.env_var_suffix
+
+    if env_var_suffix ~= nil then
+        env_var_suffix = '_' .. env_var_suffix:upper()
+    end
+
     return setmetatable({
         name = 'env',
```
Totktonada, line 55: I would like to make the names of the standard configuration sources unique. Let's add the suffix into the name.

Totktonada: Fixed this way:

```diff
diff --git a/src/box/lua/config/source/env.lua b/src/box/lua/config/source/env.lua
index 0adbf52732..735fb17aa2 100644
--- a/src/box/lua/config/source/env.lua
+++ b/src/box/lua/config/source/env.lua
@@ -47,12 +47,15 @@ local function new(opts)
     local opts = opts or {}
     local env_var_suffix = opts.env_var_suffix
 
+    local name = 'env'
+
     if env_var_suffix ~= nil then
+        name = ('%s (%s)'):format(name, env_var_suffix)
         env_var_suffix = '_' .. env_var_suffix:upper()
     end
 
     return setmetatable({
-        name = 'env',
+        name = name,
         type = 'instance',
 
         _values = {},
```

xuniq (approved):

changelogs/unreleased/config-low-priority-env-source.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Added a low priority environment configuration source, which looks into
+  `TT_*_DEFAULT` variables. It is useful to declare defaults (gh-8862).
```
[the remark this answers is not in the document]
Totktonada: Applied. Thanks!

ylobankov (approved):
Tests LGTM.

src/box/lua/config/init.lua
```diff
@@ -128,6 +130,10 @@ function methods._initialize(self)
     if extras ~= nil then
         extras.initialize(self)
     end
+
+    self:_register_source(require('internal.config.source.env').new({
```
ImeevMA, line 134: For me, it looks more correct to have a module for each source.

test/config-luatest/sources_test.lua
```diff
@@ -1,8 +1,11 @@
 local t = require('luatest')
+local fun = require('fun')
 local json = require('json')
+local yaml = require('yaml')
 local treegen = require('test.treegen')
 local justrun = require('test.justrun')
 local source_file = require('internal.config.source.file').new()
+local server = require('test.luatest_helpers.server')
```
ImeevMA, line 8: I would prefer to move the tests that require Tarantool to run into test files that already have such tests, or create a new test file. This file was created for the purpose of testing "source" modules only.

ImeevMA (approved):
Hi! Thank you for the patches. I have a couple of non-blocking comments. You can ignore them.

src/box/lua/config/init.lua
Totktonada: The idea is to write the logic of collecting environment variable values once: `<schema>:pairs()` call and the postprocessing. It reduces a chance to change one source and forget to change another one.

There is another option: implement a kind of 'base class' with this logic and two small modules. I considered it and it seems a bit too complex for me.

Is there other option to don't repeat the env vars collection logic?

TBH, I think that using an object table as a storage for internal values is the good practice (as opposite to using module's upvalues). At least, it allows introspection: we can look inside the internal state of the object.

I appreciate that you marked it as non-blocking comment, but I'm open for suggestions/discussion here.

test/config-luatest/sources_test.lua
Totktonada: My idea was to group per topic: source, applier, cli options and so on -- not instruments we're using to implement the test.

I'm OK to move the test case into another file, but how to name it? `sources_execute_test.lua`? Don't know.

The existing test cases in this file also run tarantool processes, but using another helper: `justrun`, not `server`. I can rewrite the new test case in the same way. I guess it is simple enough to be implemented in either way.

Personally I just use what looks simpler in particular case. In one of the next PRs I even wrap server/justrun into a set of more high-level helpers and it feels natural to use them in one test file.

Issue #8718, opened 2023-05-31
`app-luattest/http_client_test.lua` fails after Lua JIT update


locker:
Only the master (3.0) branch seems to be affected. I couldn't reproduce the issue on 2.11 or 2.10.

After commit https://github.com/tarantool/tarantool/commit/cde911d0e0a65a17d39b5ac60bae2bc2fe63bead was pushed, `app-luatest/http_client_test.lua` started to fail on CI with the following errors:

```
    http_client.sock_family:"AF_INET".test_encode_body_method_post_default_content_type_number ... (5.705s) fail
...arantool/tarantool/test/app-luatest/http_client_test.lua:55: connection is ok
expected: 200, actual: 595
    http_client.sock_family:"AF_INET".test_encode_body_method_post_default_content_type_boolean ... (5.198s) fail
...arantool/tarantool/test/app-luatest/http_client_test.lua:55: connection is ok
expected: 200, actual: 595
    http_client.sock_family:"AF_INET".test_body_decoders ... (5.127s) fail
...arantool/tarantool/test/app-luatest/http_client_test.lua:55: connection is ok
expected: 200, actual: 595
    http_client.sock_family:"AF_INET".test_default_body_decoders ... (5.096s) fail
...arantool/tarantool/test/app-luatest/http_client_test.lua:55: connection is ok
expected: 200, actual: 595
```

The issue can be reproduced locally with (it fails quite often but not always so you might need to run it a few times):

```
luatest -v test/app-luatest/http_client_test.lua -r 10
```

Disabling Lua JIT in the test helps.

<details><summary>Patch disabling Lua JIT in the test</summary>

```diff
diff --git a/test/app-luatest/http_client_test.lua b/test/app-luatest/http_client_test.lua
index b45cc9cc21f5..86f8a102a925 100644
--- a/test/app-luatest/http_client_test.lua
+++ b/test/app-luatest/http_client_test.lua
@@ -5,6 +5,9 @@ local uri = require('uri')
 local os = require('os')
 local t = require('luatest')
 
+jit.off()
+jit.flush()
+
 local g = t.group('http_client', {
     {sock_family = 'AF_INET'},
     {sock_family = 'AF_UNIX'},
```

</details>

**NOTE:** We disabled Lua JIT in the test in commit https://github.com/tarantool/tarantool/commit/53c94bc79ecfc0451bc3ffa57ffb88e3217671ff. Please roll it back as soon as the bug is fixed.

[...]

PR #8920, opened 2023-07-27
test: enable JIT back in app-luatest/http_client_test

igormunkin:
This patch reverts the temporary fix introduced in commit 53c94bc79ecfc0451bc3ffa57ffb88e3217671ff ("test: disable Lua JIT in app-luatest/http_client_test") since the issues with invalid traces generation for vararg functions are resolved, so JIT machinery can be enabled back then.

Follows up #8718
Relates to #8516

NO_DOC=test
NO_CHANGELOG=test

### DIFF
--- a/test/app-luatest/http_client_test.lua
+++ b/test/app-luatest/http_client_test.lua
@@ -5,10 +5,6 @@ local uri = require('uri')
 local os = require('os')
 local t = require('luatest')
 
--- FIXME(gh-8718): The test fails if Lua JIT is enabled.
-jit.off()
-jit.flush()
-
 local g = t.group('http_client', {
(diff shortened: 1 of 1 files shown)


Buristan:
Hi, Igor!
Thanks for the patch!
It LGTM, the error by checkpatch.pl is well-known as we discussed offline (still I expect, that it should be fixed somehow:)).

Buristan (approved):

igormunkin:
@Buristan, thanks for your review! I've fixed the bug in [checkpatch.pl](https://github.com/tarantool/checkpatch/pull/64), so everything is fine now.

ylobankov (approved):

PR #8922, opened 2023-07-27
box: fix invalid memory access in tuple_compare_with_key_sequential

### LINKED ISSUE #8899: Possible invalid memory access in tuple_compare_with_key_sequential with double key part.
**Bug description**

Since the `number` type has been implemented, the constraint about equality of lengths of equal tuples has been discontinued. But one of comparators uses the length equality to perform its computations: [tuple_compare_with_key_sequential](https://github.com/tarantool/tarantool/blob/51af059c10ffa451aaaa59053014f11cc64f8d5c/src/box/tuple_compare.cc#L954).

The comment I linked states that: "Key's and tuple's first field_count fields are equal, and their bsize too". This is not always true now: if we compare double and integer zero, for example, they can be equal but they have different sizes if they're put into the db differently.

If the tuple has unt8(0) and the key has double(0), then we advance the key pointer to the first byte of the double, which is unsupposed msgpack value starting point.

If the tuple has double(0) and the key has unt8(0), then we have a possible out-of-key access.

In either case the further behaviour is undefined.

**Reproduce script**

This script triggers the bug if we try to compare ffil-casted and non-casted doubles.

```
#!/usr/bin/env tarantool
box.cfg{}

function compare_bug()
    -- Just a pretty-printer, not important.
    local function pr(...)
        local objects = {...}
        local object = objects[#objects]
        table.remove(objects, #objects)
        for i = 1, #objects do
            io.write(objects[i])
        end
        print(require('yaml').encode(object))
    end

    -- Just a double-cast wrapper, not important.
    local function double(n)
        return require('ffi').cast('double', n)
    end

    local s = box.schema.space.create('test', {engine = 'memtx'})
    local pk = s:create_index('pk')
    local sk = s:create_index('sk', {parts = {{1, 'unsigned'}, {2, 'number', is_nullable = true}, {3, 'number', is_nullable = true}}})

    -- Remove the `double` cast and select will fail.
    s:insert({1, double(2)})
    pr(sk:select({1, double(2), box.NULL}, {iterator = 'EQ'}))
end

compare_bug()

os.exit()
```

How to run:

Just type `tarantool test.lua`.

mkostoevr:
Since number type was introduced we can not assume if tuples are equal by comparison then their sizes are equal too. So the place the assumption is used is fixed.

Closes #8899

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
+++ b/changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed the invalid memory access in a corner case of a specialized comparison
+  function (gh-8899).
--- a/src/box/tuple_compare.cc
+++ b/src/box/tuple_compare.cc
@@ -941,8 +941,9 @@ tuple_compare_with_key_sequential(struct tuple *tuple, hint_t tuple_hint,
 		cmp_part_count = part_count;
 	}
 	bool unused;
-	rc = key_compare_parts<is_nullable>(tuple_key, key, cmp_part_count,
-					    key_def, &unused);
+	rc = key_compare_and_skip_parts<is_nullable>(&tuple_key, &key,
+						     cmp_part_count,
+						     key_def, &unused);
 	if (!has_optional_parts || rc != 0)
 		return rc;
 	/*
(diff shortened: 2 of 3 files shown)


xuniq (approved):

alyapunov (approved):

ochaplashkin (approved):
LGTM :rocket:

alyapunov:
Back-ported to 2.11.

alyapunov:
Failed to backport to 2.10 since the patch depends on ca832f278855554f2741e54a997cece9bb7e2197.

PR #8929, opened 2023-07-28
config: allow to call config:get() from app script

Totktonada:
It is convenient to access configuration using `config:get()` from the application script (`app.file` or `app.module`).

However, before this commit, it was not possible, because the configuration was not considered as applied before the application script is loaded.

Now, the script loading is moved into the post-apply phase.

The resulting sequence of steps on startup/reload is the following.

* collect configuration information (from all the sources)
* \<if the previous step is failed, set `check_errors` status and break>
* apply the configuration (call all the appliers)
* \<if the previous step is failed, set `check_errors` status and break>
* \<set the new successful status: `ready` or `check_warnings`>
* call post-apply hooks (including application script loading)
* \<if the previous step is failed, set `check_errors` status and break>
* \<set the new successful status: `ready` or `check_warnings`>

Part of #8862

----

See also a state diagram in https://github.com/tarantool/doc/issues/3544.

### DIFF
--- a/changelogs/unreleased/config-get-configuration-from-app.md
+++ b/changelogs/unreleased/config-get-configuration-from-app.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* It is now possible to access configuration from the application script using
+  the `config:get()` method (gh-8862).
--- a/test/config-luatest/app_test.lua
+++ b/test/config-luatest/app_test.lua
@@ -1,0 +1,110 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+-- Start a script that is pointed by app.file or app.module.
+--
+-- TODO: Verify that the script has access to config:get() values.
+g.test_startup_success = function(g)
+    local script = [[
+        -- TODO: Verify that the script has access to config:get()
+        -- values.
+        -- local config = require('config')
+        -- assert(config:get('app.cfg.foo') == 42)
+
+        _G.foo = 42
+    ]]
+    local verify = function()
+        local config = require('config')
+        t.assert_equals(_G.foo, 42)
+        t.assert_equals(config:get('app.cfg.foo'), 42)
+    end
+
+    helpers.success_case(g, {
+        script = script,
+        options = {
+            ['app.file'] = 'main.lua',
+            ['app.cfg'] = {foo = 42},
+        },
+        verify = verify,
+    })
+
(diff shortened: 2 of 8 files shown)


changelogs/unreleased/config-get-configuration-from-app.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Allow to access a configuration from an application script using
+  `config:get()` (gh-8862).
```
xuniq, line 4:
```suggestion
* It is now possible to access configuration from the application script using the
  `config:get()` function (gh-8862).
```

xuniq (approved):

changelogs/unreleased/config-get-configuration-from-app.md
Totktonada: Applied with replacement of `function` to `method`. Thanks!

test/config-luatest/app_test.lua
```diff
@@ -0,0 +1,110 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+-- Start a script that is pointed by app.file or app.module.
+--
+-- TODO: Verify that the script has access to config:get() values.
+g.test_startup_success = function(g)
```
ImeevMA, line 9: I think it's a little odd that the test is added here, but logically it should be added in the next commit.

test/config-luatest/app_test.lua
```diff
@@ -0,0 +1,110 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
```
ImeevMA, line 4: Adding new helper functions for these tests seems redundant to me.

ImeevMA (approved):
Thank you for the patches! I have couple of non-blocking comments, feel free to ignore them. In general, I think init.lua is getting pretty complex.

ylobankov (approved):
Tests LGTM

test/config-luatest/app_test.lua
Totktonada: Are you about all the new cases? If so, they work without the second patch. And, I think, they're useful without it: it verifies the logic of the `app` section with running of a new instance with a config and without using internals.

Or about these two commented lines (and TODO comments)? If so, I'm OK to remove them and add in the next commit (instead of uncommenting).

PR #8943, opened 2023-08-03
config: add metrics section to schema

Lord-KA:
This patch introduces all metrics configuration.

Part of #8861

NO_DOC=tarantool/doc#3544 links the most actual schema,
       no need to update the issue.

### DIFF
--- a/changelogs/unreleased/gh-8861-metrics-options.md
+++ b/changelogs/unreleased/gh-8861-metrics-options.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* All metrics configuration options from `box.cfg{}` are now supported
+  in the YAML config (gh-8861).
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -355,6 +355,17 @@ local function apply(config)
     -- Set bootstrap_leader option.
     box_cfg.bootstrap_leader = configdata:bootstrap_leader()
 
+    -- Set metrics option.
+    local include = configdata:get('metrics.include', {use_default = true})
+    local exclude = configdata:get('metrics.exclude', {use_default = true})
+    local labels = configdata:get('metrics.labels', {use_default = true})
+
+    box_cfg.metrics = {
+        include = include or { 'all' },
+        exclude = exclude or { },
+        labels = labels or { alias = names.instance_name },
+    }
+
     -- The startup process may need a special handling and differs
     -- from the configuration reloading process.
     local is_startup = type(box.cfg) == 'function'
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -1357,6 +1357,52 @@ return schema.new('instance_config', schema.record({
             box_cfg = 'password_history_length',
         })),
     }),
+    metrics = schema.record({
+        -- Metrics doesn't have box_cfg annotation, because currently nested
+        -- options and maps/arrays defaults are not supported.
+        include = schema.set({
+            'all',
+            'network',
+            'operations',
+            'system',
+            'replicas',
+            'info',
+            'slab',
+            'runtime',
+            'memory',
+            'spaces',
+            'fibers',
+            'cpu',
+            'vinyl',
+            'memtx',
+            'luajit',
+            'clock',
+            'event_loop',
+        }),
+        exclude = schema.set({
+            'all',
+            'network',
+            'operations',
+            'system',
+            'replicas',
+            'info',
+            'slab',
+            'runtime',
+            'memory',
+            'spaces',
+            'fibers',
+            'cpu',
+            'vinyl',
+            'memtx',
+            'luajit',
+            'clock',
+            'event_loop',
+        }),
+        labels = schema.map({
+            key = schema.scalar({type = 'string'}),
+            value = schema.scalar({type = 'string'}),
+        }),
+    }),
 }, {
     -- This kind of validation cannot be implemented as the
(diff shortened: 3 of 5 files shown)


test/config-luatest/instance_config_schema_test.lua
```diff
@@ -1037,7 +1037,7 @@ g.test_box_cfg_coverage = function()
         audit_format = true,
         audit_filter = true,
 
-        -- TODO: Will be added in the scope of gh-8861.
+        -- For now defaults are managed by options applier.
         metrics = true,
```
Totktonada, line 1041: The problem is not only in defaults (otherwise we would just add it into `ignore_default`). The option is expressed as a record in the declarative configuration, so it needs a bit special handling in the test. See how `wal.ext` is handled.

src/box/lua/config/instance_config.lua
```diff
@@ -1357,6 +1357,50 @@ return schema.new('instance_config', schema.record({
             box_cfg = 'password_history_length',
         })),
     }),
+    metrics = schema.record({
```
Totktonada, line 1360: AFAIR, I usually leave comments, where we can't handle box.cfg options using box_cfg annotation. Please, look at other examples in the instance config schema and mark this option accordingly.

src/box/lua/config/applier/metrics.lua
```diff
@@ -0,0 +1,30 @@
...
+    labels = labels or default_labels
+
+    box.cfg.metrics.include = include
+    log.verbose('metrics.apply: include applied')
+
+    box.cfg.metrics.exclude = exclude
+    log.verbose('metrics.apply: exclude applied')
+
+    box.cfg.metrics.labels = labels
+    log.verbose('metrics.apply: labels applied')
```
Totktonada, line 24: Even if it works, it is definitely not a proper way to set a `box.cfg()` parameter. Let's pass all the parameters into `box.cfg()` in the `box_cfg` applier.

test/config-luatest/instance_config_schema_test.lua
Totktonada: Gleb points me that there are no `box_cfg` annotations in the `metrics` schema node, so my suggestion regarding handling it like `wal.ext` doesn't work. OK. Let's just move the `metrics = true` in `ignore` close to `log`, where the reason of ignoring is pretty same.

Lord-KA: Done.

src/box/lua/config/instance_config.lua
Lord-KA: Done.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -355,6 +355,17 @@ local function apply(config)
...
 
+    -- Set metrics option.
+    local include = configdata:get('metrics.include', {use_default = true})
+    local exclude = configdata:get('metrics.exclude', {use_default = true})
+    local labels = configdata:get('metrics.labels', {use_default = true})
+
+    box_cfg.metrics = {
+        include = include or { 'all' },
+        exclude = exclude or { },
+        labels = labels or { alias = configdata._instance_name },
```
Totktonada, line 366: Nit: It is better to use `names.instance_name` (`names` is defined above as configdata:names()`). I'm trying to keep some entities as 'more internal' and some as 'more public'.

Totktonada (approved):
Thank you! The patches look visually good.

I'll add Georgy to also look here.

DifferentialOrange (approved):

src/box/lua/config/applier/box_cfg.lua
Lord-KA: Ok

Lord-KA: Done.

test/config-luatest/config_test.lua
```diff
@@ -687,3 +687,81 @@ g.test_security_options = function()
...
+        chdir = dir,
+    }
+    g.server = server:new(opts)
+    g.server:start()
+    g.server:exec(function()
+        t.assert_equals(box.cfg.metrics.include, {'all'})
+        t.assert_equals(box.cfg.metrics.exclude, { })
+        t.assert_equals(box.cfg.metrics.labels, {alias = 'instance-001'})
+    end)
+    g.server:stop()
```
ylobankov, line 725: It looks like [this](https://github.com/tarantool/tarantool/pull/8943/commits/1fe36b80f565db5b76cc42da632bfa1bf60b6a0f#diff-29dffb3fd90865518abc81db0984c1c101cf7be1fa03e45668002c9453ba087cR19-R23) does the work.

ylobankov (approved):
Tests LGTM with one nit.

changelogs/unreleased/gh-8861-metrics-options.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* All metrics `box.cfg{}` options are now supported by config (gh-8861).
```
p7nov, line 3:
```suggestion
* All metrics  configuration options from `box.cfg{}` are now supported
  in the YAML config (gh-8861).
```

p7nov (approved):

changelogs/unreleased/gh-8861-metrics-options.md
Lord-KA: Done.

test/config-luatest/config_test.lua
Lord-KA: Oh, right. Its from the previous iteration of the test. Removed `:stop()`.

PR #8944, opened 2023-08-03
lua/socket: introduce `socket.from_fd`, `socket:detach`, `socket.socketpair`

### LINKED ISSUE #8927: Add Lua wrapper for socketpair
The `socketpair` system call creates an unnamed local socket pair. It can be useful to establish a communication channel between related processes (a process started with `popen` and its parent). We need a Lua wrapper for this system call.

```Lua
local s1, s2 = socket.socketpair()
```

We also need helpers to work with raw file descriptors so that we can pass them around to the child.

```Lua
local s1 = socket.connect('localhost', 3301)

-- Transfer the socket fd from s1 to s2.
local s2 = s2:from_fd(s1:fd())
s1:detach()
```

locker:
This PR adds two socket module functions and one socket object method:

- `socket.from_fd(fd)`: constructs a socket object from the file descriptor number. Returns the new socket object on success. Never fails. Note, the function doesn't perform any checks on the given file descriptor so technically it's possible to pass a closed file descriptor or a file descriptor that refers to a file, in which case the new socket object methods may not work as expected.

- `socket:detach()`: like `socket:close()` but doesn't close the socket file descriptor, only switches the socket object to the closed state. Returns nothing. If the socket was already detached or closed, raises an exception.

  Along with `socket.from_fd`, this method may be used for transferring file descriptor ownership from one socket to another:

  ```Lua
  local socket = require('socket')
  local s1 = socket('AF_INET', 'SOCK_STREAM', 'tcp')
  local s2 = socket.from_fd(s1:fd())
  s1:detach()
  ```

- `socket.socketpair(domain, type, proto)`: a wrapper around the [`socketpair`][1] system call. Returns two socket objects representing the two ends of the new socket pair on success. On failure, returns nil and sets [`errno`][2].

  Example:

  ```Lua
  local errno = require('errno')
  local socket = require('socket')
  local s1, s1 = socketpair('AF_UNIX', 'SOCK_STREAM', 0)
  if not s1 then
      error('socketpair: ' .. errno.strerror())
  end
  s1:send('foo')
  assert(s2:recv() == 'foo')
  s1:close()
  s2:close()
  ```

Closes #8927

[1]: https://man7.org/linux/man-pages/man2/socketpair.2.html
[2]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/errno/

### DIFF
--- a/changelogs/unreleased/gh-8927-socketpair.md
+++ b/changelogs/unreleased/gh-8927-socketpair.md
@@ -1,0 +1,4 @@
+## feature/lua/socket
+
+* Introduced new socket functions `socket.socketpair`, `socket.from_fd`, and
+  `socket:detach` (gh-8927).
--- a/src/lua/socket.lua
+++ b/src/lua/socket.lua
@@ -105,17 +105,27 @@ end
 
 local gc_socket_sentinel = ffi.new(gc_socket_t, { fd = -1 })
 
+local function do_detach(socket)
+    -- .fd is const to prevent tampering
+    ffi.copy(socket._gc_socket, gc_socket_sentinel, ffi.sizeof(gc_socket_t))
+end
+
 local function socket_close(socket)
     local fd = check_socket(socket)
     socket._errno = nil
     local r = ffi.C.coio_close(fd)
-    -- .fd is const to prevent tampering
-    ffi.copy(socket._gc_socket, gc_socket_sentinel, ffi.sizeof(gc_socket_t))
+    do_detach(socket)
     if r ~= 0 then
         socket._errno = boxerrno()
         return false
     end
     return true
+end
+
+local function socket_detach(socket)
+    check_socket(socket)
+    socket._errno = nil
+    do_detach(socket)
 end
 
 local soname_mt = {
@@ -1256,6 +1266,7 @@ end
 socket_mt   = {
     __index = {
         close = socket_close;
+        detach = socket_detach;
         errno = socket_errno;
         error = socket_error;
         sysconnect = socket_sysconnect;
--- a/src/lua/socket.lua
+++ b/src/lua/socket.lua
@@ -952,6 +952,22 @@ local function socket_new(domain, stype, proto)
     end
 end
 
+local function socket_from_fd(fd)
+    if type(fd) ~= 'number' then
+        error('fd must be a number')
+    end
+    -- Try to determine the socket type. Ignore errors.
+    local itype = 0
+    local level = internal.SOL_SOCKET
+    local opt = internal.SO_OPT[level].SO_TYPE.iname
+    local value = ffi.new('int[1]')
+    local len = ffi.new('size_t[1]', ffi.sizeof('int'))
+    if ffi.C.getsockopt(fd, level, opt, value, len) == 0 and len[0] == 4 then
+        itype = tonumber(value[0])
+    end
+    return make_socket(fd, itype)
+end
+
 local function getaddrinfo(host, port, timeout, opts)
     if type(timeout) == 'table' and opts == nil then
         opts = timeout
@@ -1594,6 +1610,7 @@ end
 --------------------------------------------------------------------------------
 
 return setmetatable({
+    from_fd = socket_from_fd;
     getaddrinfo = getaddrinfo,
     tcp_connect = tcp_connect,
     tcp_server = tcp_server,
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-8927-socketpair.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua/socket
+
+* Introduced the new socket functions: `socket.socketpair`, `socket.from_fd`,
```
p7nov, line 3:
```suggestion
* Introduced new socket functions `socket.socketpair`, `socket.from_fd`,
```

p7nov (approved):

changelogs/unreleased/gh-8927-socketpair.md
locker: Fixed. Thanks!

src/lua/socket.lua
```diff
@@ -952,6 +952,22 @@ local function socket_new(domain, stype, proto)
     end
 end
 
+local function socket_from_fd(fd)
```
Totktonada, line 993: Shouldn't we check that the socket is non-blocking? Otherwise the socket object methods (such as `:read()`) may work in an unexpected way.

And since we're owning the file descriptor now, I guess it is legal to set the non-blocking mode here.

test/app-luatest/socket_test.lua
```diff
@@ -35,3 +35,16 @@ g.test_from_fd = function()
...
     t.assert_not(s:close())
 end
+
+g.test_detach = function()
+    local s1 = socket('AF_INET', 'SOCK_STREAM', 'tcp')
+    t.assert(s1)
+    local s2 = socket.from_fd(s1:fd())
+    t.assert(s2)
+    t.assert_error_msg_content_equals('Usage: socket:method()', s1.detach)
+    t.assert_is(s1:detach(), nil)
```
Totktonada, line 46: Shouldn't we loss the `s1` object, call `collectgarbage()` and ensure that `s2` remains valid?

changelogs/unreleased/gh-8927-socketpair.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua/socket
```
Totktonada, line 1: Typo in the commit message/PR description:

```lua
local socket = require('socket')
local s1, s1 = socketpair('AF_UNIX', 'SOCK_STREAM', 0)
```

`socketpair` -> `socket.socketpair`.

Totktonada (approved):

src/lua/socket.lua
locker: The function doesn't perform any checks on the given fd. One may pass an invalid fd or fd that represents a file, in which case the `read` method may fail or block. It is by [design](https://www.notion.so/tarantool/IPROTO-over-socketpair-fb5a30dc42a348689655bf75248d9020?pvs=4#f5320789c81e466eb2632073ed47bb7c).

As for switching the fd to the non-blocking, it would be redundant in most cases because the fd returned by `socket:fd()` should already be in the non-blocking mode, no matter if the socket was created with `socket()` or `socket.socketpair()`. Note that the non-blocking mode is inherited by the child process after fork.

test/app-luatest/socket_test.lua
locker: Makes sense. Added a test case:

```diff
diff --git a/test/app-luatest/socket_test.lua b/test/app-luatest/socket_test.lua
index 701b5f6bc389..08dbc40ecbcc 100644
--- a/test/app-luatest/socket_test.lua
+++ b/test/app-luatest/socket_test.lua
@@ -48,6 +48,16 @@ g.test_detach = function()
     t.assert_error_msg_content_equals(errmsg, s1.detach, s1)
     t.assert_error_msg_content_equals(errmsg, s1.close, s1)
     t.assert(s2:close())
+
+    s1 = socket('AF_UNIX', 'SOCK_STREAM', 0)
+    t.assert(s1)
+    s2 = socket.from_fd(s1:fd())
+    t.assert(s2)
+    t.assert_is(s1:detach())
+    s1 = nil
+    collectgarbage('collect')
+    t.assert(s2:readable())
+    t.assert(s2:close())
 end
 
 g.test_socketpair = function()
```

ochaplashkin (approved):
LGTM :rocket:

src/lua/socket.lua
Totktonada: OK, just highlighted this point, because `socket()` performs this action. Maybe worth to highlight it in the documentation.

Anyway, I'm OK with the answer and closing the discussion.

PR #8945, opened 2023-08-03
Run perftests as a part of regression suite

ligurio:


perf/lua/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+add_custom_target(test-lua-perf
+                  DEPENDS 1mops_write_perftest
+                          box_select_perftest
+                          uri_escape_unescape_perftest
+                  COMMENT "Running Lua performance tests"
+)
+
+set(TARANTOOL_BIN ${PROJECT_BINARY_DIR}/src/tarantool)
```
igormunkin, line 8: You can use genexpr here: `$<TARGET_FILE:tarantool>`. See more info [here](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#target-dependent-expressions).

perf/lua/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+
+add_custom_target(1mops_write_perftest
+                  COMMAND ${TARANTOOL_BIN} 1mops_write.lua
+                  DEPENDS tarantool
+                  COMMENT Running 1mops_write_perftest
+                  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/perf/lua
```
igormunkin, line 14: Nit: this can be a variable, since you use it widely here.

perf/lua/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+add_custom_target(test-lua-perf
+                  DEPENDS 1mops_write_perftest
+                          box_select_perftest
+                          uri_escape_unescape_perftest
+                  COMMENT "Running Lua performance tests"
+)
+
+set(TARANTOOL_BIN ${PROJECT_BINARY_DIR}/src/tarantool)
+
+add_custom_target(1mops_write_perftest
```
igormunkin, line 10: Too much copy-pasting IMHO. I suggest to create a CMake macro/function to ease further benchmarks additions.

igormunkin (changes requested):
@ligurio, thanks for the patch! Please, consider my comments below. Furthermore, please add a common target for all benchmarks in scope of this commit.

perf/lua/CMakeLists.txt
ligurio: Fixed, thanks!

perf/lua/CMakeLists.txt
ligurio: Ignored, because I've added a function that generates targets with tests.

perf/lua/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+set(TARANTOOL_BIN $<TARGET_FILE:tarantool>)
+set(RUN_PERF_LUA_TEST "")
+
+function(create_perf_lua_test)
+  cmake_parse_arguments(
```
igormunkin, line 5: Minor: I'd rather `cmake_parse_arguments` has way too complex signature, so I recommend to use as close as you can to this [snippet](https://github.com/tarantool/luajit/blob/tarantool/master/cmake/MakeSourceList.cmake#L24-L37). Anyway, this is quite minor nit, so feel free to ignore, if you have any strong arguments for your approach.

perf/lua/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    ""
+    "NAME"
+    ""
+    ${ARGN}
+  )
+  message(STATUS "Creating Lua performance test ${PERF_NAME}_perftest")
+  add_custom_target(${PERF_NAME}_perftest
+                    COMMAND ${TARANTOOL_BIN} ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua
+                    COMMENT Running ${PERF_NAME}_perftest
+                    DEPENDS tarantool
```
igormunkin, line 16: Minor: It's worth adding Lua sources to `DEPENDS` section either.

igormunkin (approved):
**[PATCH 1/4] perf: add targets for running Lua performance tests**

@ligurio, thanks for the patch! Basically LGTM, but please consider my nits below.

perf/CMakeLists.txt
```diff
@@ -5,6 +5,11 @@ add_subdirectory(lua)
 find_package(benchmark QUIET)
 if (NOT ${benchmark_FOUND})
     message(AUTHOR_WARNING "Google Benchmark library was not found")
+    set(MSG "Target test-c-perf is dummy, Google Benchmark library was not found")
```
igormunkin, line 8: Minor: I don't think this variable is required here. IMHO it only makes the whole part less readable. Feel free to ignore.

perf/CMakeLists.txt
```diff
@@ -13,6 +18,25 @@ include_directories(${PROJECT_SOURCE_DIR}/src/box)
 include_directories(${PROJECT_SOURCE_DIR}/third_party)
 include_directories(${EXTRA_CORE_INCLUDE_DIRS})
 
+set(RUN_PERF_C_TEST "")
+
+function(create_perf_test_target)
```
igormunkin, line 27: [Ditto.](https://github.com/tarantool/tarantool/pull/8945#discussion_r1341417873)

perf/CMakeLists.txt
```diff
@@ -30,8 +53,17 @@ create_perf_test(PREFIX tuple
                  SOURCES tuple.cc ${PROJECT_SOURCE_DIR}/test/unit/box_test_utils.c
                  LIBRARIES core box tuple benchmark::benchmark
 )
+create_perf_test_target(PREFIX tuple)
```
igormunkin, line 60: Well, it's not the `PREFIX` but rather `TARGET`, isn't it? If so, should the name of the parameter be adjusted?

perf/CMakeLists.txt
```diff
@@ -30,8 +53,17 @@ create_perf_test(PREFIX tuple
...
 create_perf_test(PREFIX light
                  SOURCES light.cc ${PROJECT_SOURCE_DIR}/test/unit/box_test_utils.c
                  LIBRARIES small benchmark::benchmark
 )
+create_perf_test_target(PREFIX light)
+
+create_perf_test_target(PREFIX small)
+
+add_custom_target(test-c-perf
+                  DEPENDS ${RUN_PERF_C_TEST}
```
igormunkin, line 71: Minor: I forgot to mention it for the previous commit, but this is valid for both (i.e. Lua and C) patches: I'd rather use `RUN_PERF_<LANG>_TESTS_LIST` to emphasize the CMake variable type. Or use plural for `*_TESTS` at least...

perf/CMakeLists.txt
```diff
@@ -5,6 +5,11 @@ add_subdirectory(lua)
 find_package(benchmark QUIET)
 if (NOT ${benchmark_FOUND})
     message(AUTHOR_WARNING "Google Benchmark library was not found")
+    set(MSG "Target test-c-perf is dummy, Google Benchmark library was not found")
+    add_custom_target(test-c-perf
+                      COMMAND ${CMAKE_COMMAND} -E cmake_echo_color --red ${MSG}
```
igormunkin, line 10: Side note: Nicely done! Should be used widely in LuaJIT CMake machinery. Thanks!

igormunkin (approved):
**[PATCH 2/4] perf: add targets for running C performance tests**

@ligurio, thanks for the patch! LGTM like the previous one, but please consider my comments here too.

igormunkin (approved):
**[PATCH 3/4] cmake: build performance tests only with release build**

@ligurio, thanks for the patch! Did you think about introducing new CMake configuration option for building and running perf tests? If you consider this out of the scope of this PR, please mention it explicitly in this commit message. I would also use the default `WARNING` here, but have no strong arguments against the current approach. All in all, LGTM, but I'm looking forward to your reply regarding the separate CMake option.

igormunkin (approved):
**[PATCH 4/4] ci: run performance tests**

@ligurio, thanks for the patch! LGTM.

perf/CMakeLists.txt
ligurio: String with message is used below two times, so variable is requied.

perf/CMakeLists.txt
ligurio: Renamed.

perf/CMakeLists.txt
ligurio: renamed in both patches

perf/lua/CMakeLists.txt
ligurio: Updated:

```diff
--- a/perf/lua/CMakeLists.txt
+++ b/perf/lua/CMakeLists.txt
@@ -10,10 +10,11 @@ function(create_perf_lua_test)
     ${ARGN}
   )
   message(STATUS "Creating Lua performance test ${PERF_NAME}_perftest")
+  set(TEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua)
   add_custom_target(${PERF_NAME}_perftest
-                    COMMAND ${TARANTOOL_BIN} ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua
+                    COMMAND ${TARANTOOL_BIN} ${TEST_PATH}
                     COMMENT Running ${PERF_NAME}_perftest
-                    DEPENDS tarantool
+                    DEPENDS tarantool ${TEST_PATH}
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   )
   set(RUN_PERF_LUA_TESTS_LIST ${RUN_PERF_LUA_TESTS_LIST} ${PERF_NAME}_perftest PARENT_SCOPE)
```

perf/lua/CMakeLists.txt
ligurio: I have no objections, this looks more clear.

```diff
diff --git a/perf/CMakeLists.txt b/perf/CMakeLists.txt
index 63108c04d8..4c6a619fd0 100644
--- a/perf/CMakeLists.txt
+++ b/perf/CMakeLists.txt
@@ -25,13 +25,18 @@ include_directories(${EXTRA_CORE_INCLUDE_DIRS})
 set(RUN_PERF_C_TESTS_LIST "")
 
 function(create_perf_test_target)
-  cmake_parse_arguments(
-    PERF
-    ""
-    "TARGET"
-    ""
-    ${ARGN}
-  )
+  set(prefix PERF)
+  set(noValues)
+  set(singleValues TARGET)
+  set(multiValues)
+
+  # FIXME: if we update to CMake >= 3.5, can remove this line.
+  include(CMakeParseArguments)
+  cmake_parse_arguments(${prefix}
+                        "${noValues}"
+                        "${singleValues}"
+                        "${multiValues}"
+                        ${ARGN})
   message(STATUS "Creating C performance test ${PERF_TARGET}_perftest")
   add_custom_target(${PERF_TARGET}_perftest
                     COMMAND "$<TARGET_FILE:${PERF_TARGET}.perftest>"
@@ -42,13 +47,18 @@ function(create_perf_test_target)
 endfunction()
 
 function(create_perf_test)
-  cmake_parse_arguments(
-    PERF
-    ""
-    "NAME"
-    "SOURCES;LIBRARIES"
-    ${ARGN}
-  )
+  set(prefix PERF)
+  set(noValues)
+  set(singleValues NAME)
+  set(multiValues "SOURCES;LIBRARIES")
+
+  # FIXME: if we update to CMake >= 3.5, can remove this line.
+  include(CMakeParseArguments)
+  cmake_parse_arguments(${prefix}
+                        "${noValues}"
+                        "${singleValues}"
+                        "${multiValues}"
+                        ${ARGN})
   add_executable(${PERF_TARGET}.perftest ${PERF_SOURCES})
   target_link_libraries(${PERF_TARGET}.perftest PUBLIC ${PERF_LIBRARIES})
 endfunction()
diff --git a/perf/lua/CMakeLists.txt b/perf/lua/CMakeLists.txt
index 908b293efd..e33aaeaea6 100644
--- a/perf/lua/CMakeLists.txt
+++ b/perf/lua/CMakeLists.txt
@@ -2,13 +2,19 @@ set(TARANTOOL_BIN $<TARGET_FILE:tarantool>)
 set(RUN_PERF_LUA_TESTS_LIST "")
 
 function(create_perf_lua_test)
-  cmake_parse_arguments(
-    PERF
-    ""
-    "NAME"
-    ""
-    ${ARGN}
-  )
+  set(prefix PERF)
+  set(noValues)
+  set(singleValues NAME)
+  set(multiValues)
+
+  # FIXME: if we update to CMake >= 3.5, can remove this line.
+  include(CMakeParseArguments)
+  cmake_parse_arguments(${prefix}
+                        "${noValues}"
+                        "${singleValues}"
+                        "${multiValues}"
+                        ${ARGN})
+
   message(STATUS "Creating Lua performance test ${PERF_NAME}_perftest")
   set(TEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua)
   add_custom_target(${PERF_NAME}_perftest
```

perf/CMakeLists.txt
ligurio: Updated.

ligurio:
> [PATCH 3/4] cmake: build performance tests only with release build

> Did you think about introducing new CMake configuration option for building and  running perf tests? If you consider this out of the scope of this PR, please mention it explicitly in this commit message. I would also use the default WARNING here, but have no strong arguments against the current approach.

Added a note to commit:

```
commit 75e64808afcec70bc5c864bc1cbdc05d4cad7274
Date:   Mon Aug 14 19:42:17 2023 +0300

    cmake: build performance tests only with release build
    
    Note that targets for running performance tests are generated only when
    CMAKE_BUILD_TYPE is equal to Release or RelWithDebug. Additionally, C++
    performance tests require Google Benchmark library. Using non-debug
    build and having installed Google Benchmark library is rare case, so I
    suppose we don't need to introduce CMake option for performance testing.
```

Issue #7309, opened 2022-06-27
space:truncate() ignores foreign keys


ImeevMA:
Reproducer:
```
a = box.schema.space.create('a', {format = {{'a', 'integer'}}})
_ = a:create_index('i')
b = box.schema.space.create('b', {format = {{'a', 'integer'}, {'b', 'integer', foreign_key = {asd = {field = 'a', space = 'a'}}}}})
_ = b:create_index('i')
a:insert({1, 1})
b:insert({2, 1})
a:truncate()
b:insert({3, 1})
a:select()
b:select()
```

Result:
```
tarantool> b:insert({3, 1})
---
- error: 'Foreign key constraint ''asd'' failed for field ''2 (b)'': foreign tuple
    was not found'
...

tarantool> a:select()
2022-06-27 10:34:40.819 [69589] main/103/interactive C> Potentially long select from space 'a' (514)
 stack traceback:
	builtin/box/schema.lua:2382: in function 'check_select_safety'
	builtin/box/schema.lua:2396: in function <builtin/box/schema.lua:2386>
	[C]: in function 'pcall'
	builtin/box/console.lua:404: in function 'eval'
	builtin/box/console.lua:710: in function 'repl'
	builtin/box/console.lua:759: in function <builtin/box/console.lua:747>
---
- []
...

tarantool> b:select()
2022-06-27 10:34:40.826 [69589] main/103/interactive C> Potentially long select from space 'b' (515)
 stack traceback:
	builtin/box/schema.lua:2382: in function 'check_select_safety'
	builtin/box/schema.lua:2396: in function <builtin/box/schema.lua:2386>
	[C]: in function 'pcall'
	builtin/box/console.lua:404: in function 'eval'
	builtin/box/console.lua:710: in function 'repl'
	builtin/box/console.lua:759: in function <builtin/box/console.lua:747>
---
- - [2, 1]
...
```

PR #8947, opened 2023-08-03
box: loose truncate check in case of foreign key

### LINKED ISSUE #8946: Can't truncate a space that another empty space refers to via foreign key
In #7309 truncation of spaces that another space refers using foreign key was disabled.
However, if the referring space is empty - there's no harm in truncation of referred space.
We must allow truncation in this case.
That a severe bug since there's no way to truncate such a space at all.
<details>
<summary>Reproduce</summary>

```lua
box.schema.space.create('client', {if_not_exists = true}) 
  
box.space.client:format({ 
                          {name = 'customer_id', type = 'string', is_nullable = false}, 
                          {name = 'esia_id', type = 'string', is_nullable = false},                           
                         }) 
  
box.space.client:create_index('pk_client_customer_id', {parts = {{field='customer_id', collation='unicode'}}, type='tree', unique=true, if_not_exists = true}) 
box.space.client:create_index('idx_client_esia_id',    {parts = {{field='esia_id', collation='unicode'}},     type='tree', unique=false, if_not_exists = true}) 
       
 box.schema.space.create('client_phones', {if_not_exists = true, foreign_key = {space = 'client', field ={customer_id='customer_id'}} }) 
  
 box.space.client_phones:format({ 
                                {name = 'phone', type = 'string', is_nullable = false}, 
                                {name = 'customer_id'}, 
                                }) 
  
box.space.client_phones:create_index('idx_client_phones_phone', {parts = {{field='phone', collation='unicode'}}, type='tree', unique=true, if_not_exists = true}) 
  
  
box.space.client:insert{'01','esia-01'} 
box.space.client:insert{'02','esia-02'} 
  
box.space.client_phones:insert{'9121234','01'} 
box.space.client_phones:insert{'3222222','02'} 
  
box.space.client_phones:truncate() 
box.space.client:truncate() 
```
</details>

alyapunov:
In #7309 a truncation of a space that was referenced by foreign key from some other space was prohibited.

It appeared that this solution is too bothering since a user can't truncate a space even if he truncated referring space before that.

Fix it by allowing space truncate if referring spaces are empty.

Note that by design space truncate is implemented quite similar to space drop. Both delete all indexes, from secondary to primary. Since this patch allows deletion of the primary index (which is the action that actually deletes all data from the space), this patch changes the result of space drop too: the space remains alive with no indexes, while before this patch it remained alive with no secondary indexes but with present primary. In both cases the behaviour is quite strange and must be fixed in #4348. To make tests pass I had to perform drop in box.atomic manually.

Closes #8946

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-8946-foreign-key-disables-truncate.md
+++ b/changelogs/unreleased/gh-8946-foreign-key-disables-truncate.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when a space that is referenced by a foreign key could not
+  be truncated even if the referring space was empty (gh-8946).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -40,6 +40,7 @@
 #include "coll_id_def.h"
 #include "txn.h"
 #include "tuple.h"
+#include "tuple_constraint.h"
 #include "fiber.h" /* for gc_pool */
 #include "scoped_guard.h"
 #include <base64.h>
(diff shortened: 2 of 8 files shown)


changelogs/unreleased/gh-8946-foreign-key-disables-truncate.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when a space that is referenced by foreign key could not
+  be truncated even if referring space is empty (gh-8946).
```
p7nov, line 4:
```suggestion
* Fixed a bug when a space that is referenced by a foreign key could not
  be truncated even if the referring space was empty (gh-8946).
```

p7nov (approved):
Minor grammar fixes in the changelog.

locker (approved):

ylobankov (approved):

alyapunov:

PR #8953, opened 2023-08-04
lua/popen: introduce `inherit_fds` option for `popen.new`

### LINKED ISSUE #8926: Allow to inherit fd on popen
`popen.new` either leaves all parent file descriptors open in the child or closes them all. There is no way to inherit a specific subset of parent file descriptors. This would be useful to establish a connection between the parent and child processes via a socket created with the `socketpair` system call.

```Lua
popen.new({...}, {inherit_fds = {fd1, fd2, ...}})
```

locker:
The new option takes an array of file descriptor numbers that should be left open in the child process if the `close_fds` flag is set. If the `close_fds` flag isn't set, the option has no effect because in this case none of the parent's file descriptors are closed anyway.

The option may be useful for establishing a communication channel between a Tarantool instance and a process spawned by it with `popen.new`, for example, using `socket.socketpair`.

Closes #8926

### DIFF
--- a/src/lua/popen.c
+++ b/src/lua/popen.c
@@ -302,6 +302,28 @@ luaL_tolstring_strict(struct lua_State *L, int idx, size_t *len_ptr)
 }
 
 /**
+ * Extract an integer number from the Lua stack.
+ *
+ * Return true for an integer number and store its value in @a value.
+ *
+ * Unlike lua_tointeger() it accepts only an integer number and
+ * does not accept a string.
+ */
+static bool
+luaL_tointeger_strict(struct lua_State *L, int idx, int *value)
+{
+	if (lua_type(L, idx) != LUA_TNUMBER)
+		return false;
+	double num = lua_tonumber(L, idx);
+	if (num < INT_MIN || num > INT_MAX)
+		return false;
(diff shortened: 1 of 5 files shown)


src/lua/popen.c
```diff
@@ -1250,8 +1203,6 @@ luaT_popen_parse_mode(struct lua_State *L, int idx)
  *                getpgrp() fails in the parent process.
  * - SystemError: (temporary restriction) the parent process
  *                has closed stdin, stdout or stderr.
- * - OutOfMemory: unable to allocate the handle or a temporary
- *                buffer.
```
Totktonada, line 1254: We should also update the documentation: https://www.tarantool.io/en/doc/latest/reference/reference_lua/popen/#popen-new

The same for `:read()`.

Totktonada (approved):
Thank you for the patches!

I see no problems visually.

p7nov (approved):

src/lua/popen.c
locker: Right. Added a doc bot request to the relevant commit.

ylobankov (approved):
Test LGTM

PR #8955, opened 2023-08-04
election: fix box.ctl.demote() nop in off-mode

### LINKED ISSUE #6860: box.ctl.demote() and synchro queue owner interaction is inconsistent.
**Bug description**

box.ctl.demote() and synchro queue owner interaction is inconsistent.

Running box.ctl.demote() with election_mode = 'off' sets synchro queue owner to NIL.
Running box.ctl.demote() with election_mode = 'manual' leaves synchro queue owner as is.

**Steps to reproduce**

```lua
box.cfg{election_mode = 'manual'}
box.ctl.promote()
box.ctl.demote()
box.cfg{election_mode = 'off'}
box.schema.create_space('test'):create_index('pk')
box.space.test:replace({1})
box.ctl.demote()
box.space.test:replace({1})
```

This leads to not so obvious behaviour:
On leader node with election_mode = 'manual' run box.ctl.demote(), set box.cfg{replication_mode='off'} now this node (assuming no one else took synchro queue from it) is RW (for synchronous transactions too). Further calls to box.ctl.demote() will not do anything, because term was bumped during previous demote and now this node doesn't consider itself leader.

**Actual behavior**

Node is RW, it can't be demoted.

**Expected behavior**

I don't really know.

Gerold103:
box.ctl.demote() used not to do anything with election_mode='off' if the synchro queue didn't belong to the caller in the same term as the election state.

The reason could be that if the synchro queue term is "outdated", there is no guarantee that some other instance doesn't own it in the latest term right now.

The "problem" is that this could be workarounded easily by just calling promote + demote together.

There isn't much sense in fixing it for the off-mode because the only reasons off-mode exists are 1) for people who don't use synchro at all, 2) who did use it and want to stop. Hence they need demote just to disown the queue.

The patch "legalizes" the mentioned workaround by allowing to perform demote in off-mode even if the synchro queue term is old.

Closes #6860

NO_DOC=bugfix

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -2958,13 +2958,33 @@ box_demote(void)
 	if (!is_box_configured)
 		return 0;
 
-	/* Currently active leader is the only one who can issue a DEMOTE. */
-	bool is_leader = txn_limbo_replica_term(&txn_limbo, instance_id) ==
-			 box_raft()->term && txn_limbo.owner_id == instance_id;
-	if (box_election_mode != ELECTION_MODE_OFF)
-		is_leader = is_leader && box_raft()->state == RAFT_STATE_LEADER;
-	if (!is_leader)
-		return 0;
+	const struct raft *raft = box_raft();
+	if (box_election_mode == ELECTION_MODE_OFF) {
+		assert(raft->state == RAFT_STATE_FOLLOWER);
+		if (raft->leader != REPLICA_ID_NIL) {
+			diag_set(ClientError, ER_NOT_LEADER, raft->leader);
+			return -1;
+		}
+		if (txn_limbo.owner_id == REPLICA_ID_NIL)
+			return 0;
+		/*
+		 * If the limbo term is up to date with Raft, then it might have
+		 * a valid owner right now. Demotion would disrupt it. In this
+		 * case the user has to explicitly overthrow the old owner with
+		 * local promote(), or call demote() on the actual owner.
+		 */
+		if (txn_limbo.promote_greatest_term == raft->term &&
+		    txn_limbo.owner_id != instance_id)
+			return 0;
+	} else {
+		if (txn_limbo_replica_term(&txn_limbo, instance_id) !=
+		    raft->term)
+			return 0;
+		if (txn_limbo.owner_id != instance_id)
+			return 0;
+		if (raft->state != RAFT_STATE_LEADER)
+			return 0;
+	}
 	if (box_trigger_elections() != 0)
 		return -1;
 	if (box_election_mode != ELECTION_MODE_OFF)
--- a/test/replication-luatest/gh_6860_election_off_demote_test.lua
+++ b/test/replication-luatest/gh_6860_election_off_demote_test.lua
@@ -1,0 +1,196 @@
+local t = require('luatest')
+local server = require('luatest.server')
+local replicaset = require('luatest.replica_set')
+
+local g = t.group('gh-6860')
+
+local function replicaset_create(g)
+    g.replicaset = replicaset:new({})
+    local box_cfg = {
+        replication_timeout = 0.1,
+        replication_synchro_quorum = 2,
+        replication_synchro_timeout = 1000,
+        replication = {
+            server.build_listen_uri('server1', g.replicaset.id),
+            server.build_listen_uri('server2', g.replicaset.id),
+        },
+    }
+    g.server1 = g.replicaset:build_and_add_server({
+        alias = 'server1', box_cfg = box_cfg
+    })
+    -- For stability. To guarantee server-1 is first, server-2 is second.
+    box_cfg.read_only = true
+    g.server2 = g.replicaset:build_and_add_server({
+        alias = 'server2', box_cfg = box_cfg
+    })
+    g.replicaset:start()
+    g.server2:update_box_cfg{read_only = false}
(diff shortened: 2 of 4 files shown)


src/box/box.cc
```diff
@@ -2958,13 +2958,24 @@ box_demote(void)
...
-	if (!is_leader)
-		return 0;
+	const struct raft *raft = box_raft();
+	if (box_election_mode == ELECTION_MODE_OFF) {
+		assert(raft->state == RAFT_STATE_FOLLOWER);
+		if (raft->leader != REPLICA_ID_NIL) {
+			diag_set(ClientError, ER_NOT_LEADER, raft->leader);
+			return -1;
+		}
+		if (txn_limbo.owner_id == REPLICA_ID_NIL)
```
sergepetrenko, line 2968: In the discussion [here](https://github.com/tarantool/tarantool/issues/6860#issuecomment-1663432137) I meant to skip the term check but not the owner check.

I suppose you removed the owner check because the actual limbo owner might be dead. But do we really want to allow any node to demote the possibly running queue owner?

Maybe leave the crutch with 'promote + demote' for the case when the queue owner is dead?

sergepetrenko (commented):
Thanks for the patch!

src/box/box.cc
Gerold103: The problem is that if you were an owner in an old term, you have no guarantees in the new term. And we don't even track the limbo owner. We only track the election leader. So we can't tell if the queue is owned by anybody else in the newest term.

You said *"Not checking the owner means you may "steal" the synchro queue from the current owner."*, but it still can happen with checking the owner. Because the fact that you were an owner in term N doesn't tell anything about the owner in the term N+1.

Do I miss anything?

sergepetrenko: Oh, I think I understood. Yes, indeed, there is no point in checking limbo owner for an old term (when synchro term is N, and election term is N + 1).

My point was that when the limbo term is up to date (equal to election term), it's better to check the limbo owner. Otherwise the running owner may be disrupted by anyone who issues a DEMOTE. And issuing demote on all replicaset nodes simultaneously will make all of them write a DEMOTE to WAL and thus lead to split-brain.

When the limbo term is outdated, it's fine to not check the owner. Even if there is another owner in term N + 1, our demote from term N will lead to split-brain and won't disrupt the owner in term N + 1.

Gerold103: > What do you think?

Agreed, done.

sergepetrenko (approved):
Thanks for the fixes!
LGTM.

sergepetrenko:
@Gerold103, the test hangs occasionally. Please, take a look:
```
[029] replication-luatest/gh_6860_election_off_demot>                 [ fail ]
[029] Test failed! Output from reject file /tmp/t/rejects/replication-luatest/gh_6860_election_off_demote.reject:
[029] Tarantool version is 3.0.0-alpha1-48-g5c271cbd1
[029] TAP version 13
[029] 1..4
[029] # Started on Thu Aug 17 11:07:18 2023
[029] # Starting group: gh-6860
[029] ok     1	gh-6860.test_election_off_demote_self_no_leader
[029] not ok 2	gh-6860.test_election_off_demote_other_no_leader
[029] #   ...gey/Source/sptnt/test-run/lib/luatest/luatest/server.lua:344: Timed out to wait for "synchro queue term" condition for server (alias: server1, workdir: server1-IiT-fhoCY90I, pid: 123472) within 60s
[029] #   stack traceback:
[029] #   	...replication-luatest/gh_6860_election_off_demote_test.lua:126: in function 'gh-6860.test_election_off_demote_other_no_leader'
[029] #   	...
[029] #   	[C]: in function 'xpcall'
[029] ok     3	gh-6860.test_election_off_demote_other_same_term
[029] ok     4	gh-6860.test_election_off_demote_other_leader
[029] # Ran 4 tests in 61.000 seconds, 3 succeeded, 1 errored
[029] 
[029] [test-run server "luatest_server"] The log file /tmp/t/029_replication-luatest/gh_6860_election_off_demote_test.log has zero size

```

Gerold103:
@sergepetrenko I've run the test a few thousand times but couldn't reproduce. Can you please double-check if you maybe have some local changes done?

sergepetrenko:
@Gerold103, hm, yes, I can't reproduce the failure anymore. Sorry for the false alarm.

grafin (approved):
Thanks for the patch!

andreyaksenov (approved):

ylobankov (approved):
Tests LGTM

test/replication-luatest/gh_6860_election_off_demote_test.lua
```diff
@@ -0,0 +1,196 @@
...
+local replicaset = require('luatest.replica_set')
+
+local g = t.group('gh-6860')
+
+local function replicaset_create(g)
+    g.replicaset = replicaset:new({})
+    local box_cfg = {
+        replication_timeout = 0.1,
+        replication_synchro_quorum = 2,
+        replication_synchro_timeout = 1000,
```
ochaplashkin, line 12: too much value, isn't it?

Gerold103: Why? I intentionally don't want to trigger any timeouts.

Issue #8967, opened 2023-08-09
config: upgrade credentials


Lord-KA:
Right now, credentials support is very primitive, it allows to create users and roles from config, assign them privileges (`read`, `write`, etc) for `universe` (meaning all objects), and "inherit" roles. For the full [access control](https://www.tarantool.io/en/doc/latest/book/admin/access_control/) functionality the following should be implemented:

```[tasklist]
- [x] Synchronization of privileges and role "inheritance" between current box configuration and config (#8968)
- [x] Remove hashes from schema (#8997)
- [x] Password sync for users (#8997)
- [x] Add test that users/roles are not deleted when removed from config (https://github.com/tarantool/tarantool/pull/9088)
- [x] Set privileges in a "lazy" way using triggers (e.g. on space/function/sequence rename or creation) (https://github.com/tarantool/tarantool/pull/9151)
- [x] Update privileges when instance goes from Read Only state to Read Write (https://github.com/tarantool/tarantool/pull/9119)
- [x] Add alerts when privileges can't be set for some reason (e.g. instance is in Read Only state) (https://github.com/tarantool/tarantool/pull/9208)
- [ ] ~Add example config for credentials~ Postponed
- [x] Add support for `lua_eval`, `lua_call`, `sql` object types. (https://github.com/tarantool/tarantool/pull/9231)
```

Totktonada:
All the code tasks are done. I'll track the example enhancements separately. Closing so.

PR #8968, opened 2023-08-09
config: rework credentials to support priv sync

Lord-KA:
Before this patch, credentials applier used to just grant all privileges and permissions with {if_not_exists = true}. It didn't allow removing a permission, nor setting only new permissions.

Now credentials applier converts box configuration and desired config to an intermediate representation, calculates diff for them and only after that applies the diff.

Part of #8967

NO_DOC=yet

### DIFF
--- a/changelogs/unreleased/gh-8861-config-privilege-sync.md
+++ b/changelogs/unreleased/gh-8861-config-privilege-sync.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Improved the credentials applier: now it supports two-way synchronization
+  of roles and privileges for both users and roles (gh-8861).
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -1,30 +1,328 @@
 local log = require('internal.config.utils.log')
 
-local function grant_privileges(name, privileges, role_or_user, grant_f)
-    for _, privilege in ipairs(privileges or {}) do
-        log.verbose('credentials.apply: grant %s to %s %s (if not exists)',
-            privilege, role_or_user, name)
-        for _, permission in ipairs(privilege.permissions or {}) do
-            local opts = {if_not_exists = true}
-            if privilege.universe then
-                grant_f(name, permission, 'universe', nil, opts)
-            end
-            -- TODO: It is not possible to grant a permission for
-            -- a non-existing object. It blocks ability to set it
-            -- from a config. Disabled for now.
-            --[[
-            for _, space in ipairs(privilege.spaces or {}) do
-                grant_f(name, permission, 'space', space, opts)
-            end
-            for _, func in ipairs(privilege.functions or {}) do
-                grant_f(name, permission, 'function', func, opts)
-            end
-            for _, seq in ipairs(privilege.sequences or {}) do
-                grant_f(name, permission, 'sequence', seq, opts)
-            end
-            ]]--
+--[[
+This intermediate representation is a formatted set of
+all permissions for a user or role. It is required to
+standardize diff functions. All the validation is done
+by config or box.info(), so neither the format nor the
+helper function don't perform it. Below you can find two
+converters to this representation, from box format and
+config schema format, accordingly `privileges_{box,config}_convert()`.
+
+[obj_type][obj_name] = {
+    read = true,
+    write = true,
+    ...
+}
+
+obj_types:
+ - 'user'
+ - 'role'
+ - 'space'
+ - 'function'
+ - 'sequence'
+ - 'universe'
+
+obj_names:
+ - mostly user defined strings, provided by config or box
+ - special value '', when there is no obj_name, e.g. for
+   'universe' obj_type or for granting permission for all
+   objects of a type.
+
+privs:
+ - read
+ - write
+ - execute
+   - lua_eval
+   - lua_call
+   - sql
+ - session
+ - usage
+ - create
+ - drop
+ - alter
+ - reference
+ - trigger
+ - insert
+ - update
+ - delete
+
+Examples:
+- - box.schema.user.grant('myuser', 'execute', 'function', 'myfunc')
+  - ['function']['myfunc']['execute'] = true
+  - grant execute of myfunc
+
+- - box.schema.user.grant('myuser', 'execute', 'function')
+  - ['function']['']['execute'] = true
+  - grant execute of all registered functions
+
+- - box.schema.user.grant('myuser', 'read', 'universe')
+  - ['universe']['']['read'] = true
+  - grant read to universe
+
+- - box.schema.user.grant('myuser', 'execute', 'role', 'super')
+  - ['role']['super']['execute'] = true
+  - equivalent to granting a role to myuser
+
+]]--
+
+local function privileges_from_box(privileges)
+    privileges = privileges or {}
+    assert(type(privileges) == 'table')
+
+    local res = {
+        ['user'] = {},
+        ['role'] = {},
+        ['space'] = {},
+        ['function'] = {},
+        ['sequence'] = {},
+        ['universe'] = {},
+    }
+
+    for _, priv in ipairs(privileges) do
+        local perms, obj_type, obj_name = unpack(priv)
+        obj_name = obj_name or ''
+
+        res[obj_type][obj_name] = res[obj_type][obj_name] or {}
+
+        for _, perm in ipairs(perms:split(',')) do
+            res[obj_type][obj_name][perm] = true
         end
     end
+
+    return res
+end
+
+-- Note that 'all' is considered a special value, meaning all objects of
+-- obj_type will be granted this permission. Don't use this function if it
+-- may occur in any other meaning, e.g. user defined name.
+--
+-- Note: `obj_names` can be either an array with objects names or a string
+--       with a single one. It could also be `nil`, meaning "do nothing".
+local function privileges_add_perm(obj_type, obj_names, perm, intermediate)
+    if obj_names == nil then
+        return
+    end
+    if type(obj_names) == 'string' then
+        obj_names = {obj_names}
+    end
+
+    for _, obj_name  in ipairs(obj_names) do
+        if obj_name == 'all' then
+            -- '' is a special value, meaning all objects of this obj_type.
+            obj_name = ''
+        end
+        intermediate[obj_type][obj_name] =
+            intermediate[obj_type][obj_name] or {}
+        intermediate[obj_type][obj_name][perm] = true
+    end
+end
+
+local function privileges_from_config(config_data)
+    local privileges = config_data.privileges or {}
+    assert(type(privileges) == 'table')
+
+    local intermediate = {
+        ['user'] = {},
+        ['role'] = {},
+        ['space'] = {},
+        ['function'] = {},
+        ['sequence'] = {},
+        ['universe'] = {},
+    }
+
+    for _, priv in ipairs(privileges) do
+        for _, perm in ipairs(priv.permissions) do
+            if priv.universe then
+                privileges_add_perm('universe', 'all', perm, intermediate)
+            end
+            privileges_add_perm('space', priv.spaces, perm, intermediate)
+            privileges_add_perm('function', priv.functions, perm, intermediate)
+            privileges_add_perm('sequence', priv.sequences, perm, intermediate)
+        end
+    end
+
+    local roles = config_data.roles or {}
+
+    for _, role_name in ipairs(roles) do
+        -- Unlike spaces, functions and sequences, role is allowed to be
+        -- named 'all', so `privileges_add_perm()` isn't used.
+        intermediate['role'][role_name] = intermediate['role'][role_name] or {}
+        intermediate['role'][role_name]['execute'] = true
+    end
+
+    return intermediate
+end
+
+-- Intermediate representation is basically a set, so this function subtracts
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/gh-8861-config-privilege-sync.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Improved credentials applier, now it supports a two-way synchronisation
+  of roles and privileges for both users and roles (gh-8861).
```
p7nov, line 4:
```suggestion
* Improved the credentials applier: it now supports two-way synchronization
  of roles and privileges for both users and roles (gh-8861).
```

p7nov (approved):

changelogs/unreleased/gh-8861-config-privilege-sync.md
Lord-KA: Done.

test/interactive_tarantool.lua
```diff
@@ -224,6 +224,14 @@ function mt.close(self)
     self.ph:close()
 end
 
+function mt.repl(child, command, expected)
+    child:execute_command(command)
+    local response = child:read_response()
+    if (expected ~= nil) then
+        t.assert_equals(response, expected)
+    end
+end
```
Totktonada, line 233: Nits:

1. `child` -> `self` (for consistency with other methods).
2. Extra parentheses in the if-condition (please, add it into your self-check list).

Lord-KA: Fixed. I also renamed `repl()` to `roundtrip()` as you suggested and added a comment.

Totktonada: Please, reflect the method name change in the commit message.

The `response` may be a table, so your initial idea to use `t.assert_equals()` makes sense. I would return it back.

Lord-KA: My bad, missed that. Fixed.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+
+        for _, perm in ipairs(perms:split(',')) do
+            res[obj_type][obj_name][perm] = true
+        end
+    end
+
+    return res
+end
+
+local function privileges_config_convert(config_data)
```
Totktonada, line 96: Regarding names. The usual convention is either `<component>_<action>` or just `<action>`. The `<action>` is usually a verb or a verb + a noun group. There are also constructors like `<component>_from_<something>`. We can use this naming scheme.

```lua
-- Create a privileges definition represented in the intermediate
-- representation described at beginning of the source file from
-- `box.schema.{user,role}.info(name)` data.
local function priv_ir_from_box(data)
    <...>
end

-- Create a privileges definition represented in the intermediate
-- representation described at beginning of the source file from
-- the declarative config data.
local function priv_ir_from_config(data)
    <...>
end
```

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+ - 'space'
+ - 'function'
+ - 'sequence'
+ - 'universe'
+
+obj_names:
+ - mostly user defined strings, provided by config or box
+ - special value '', when there is no obj_name, e.g. for
+   'universe' obj_type or for granting permission for all
+   objects of a type.
```
Totktonada, line 30: Raw idea (feel free to decline): we can use `universe` as the only available name of the object in the `universe` object type. It is similar how `nil` is the only object of the `nil` type.

The empty string fits here too.

src/box/lua/config/applier/credentials.lua
```diff
@@ -57,15 +345,13 @@ local function create_roles(role_map)
...
     end
 
     -- Assign underlying roles.
     for role_name, role_def in pairs(role_map or {}) do
         if role_def ~= nil then
-            assign_roles_to_role(role_name, role_def.roles)
+            sync_privileges(role_name, role_def, 'role')
         end
     end
 end
```
Totktonada, line 357: The comment is outdated here and in the block just above.

Totktonada (approved):
I like the general approach and feels the code understandable.

Please, consider a few comments below and we'll go ahead with it.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+    for _, priv in ipairs(privileges) do
+        for _, perm in ipairs(priv.permissions) do
+            if priv.universe then
+                intermediate['universe'][''] =
+                        intermediate['universe'][''] or {}
+                intermediate['universe'][''][perm] = true
+            else
+                add_permission(intermediate, perm, priv.spaces, 'space')
+                add_permission(intermediate, perm, priv.functions, 'function')
+                add_permission(intermediate, perm, priv.sequences, 'sequence')
```
Totktonada, line 134: Can't permissions on `universe` be set together with other privileges? I would handle them all, without guessing that if `universe` is present, we shouldn't process others.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+            add_permission('universe', '', 'read', res)
+
+        elseif name == 'super' then
+            add_permission('universe', '', 'read', res)
+            add_permission('universe', '', 'write', res)
+            add_permission('universe', '', 'execute', res)
+            -- TODO add when #8906 is merged:
+            -- add_permission('universe', '', 'lua_eval', res)
+            -- add_permission('universe', '', 'lua_call', res)
+            -- add_permission('universe', '', 'sql', res)
```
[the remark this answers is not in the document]
Lord-KA: Done, same for `'admin'` case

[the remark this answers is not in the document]
Lord-KA: Ok, it was caught by my `add_defaults()` test, neither `'super'` nor `'admin'` don't have this new permissions in box, because they are already covered by `'execute'`.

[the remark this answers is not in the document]
Lord-KA: I suggest we remove them here too.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+    }
+
+    local function add_permission(intermediate, perm, obj_def, obj_type)
+        obj_def = obj_def or {}
+        -- XXX maybe rename 'all' to something better
+        if obj_def[1] == 'all' then
+            intermediate[obj_type][''] = intermediate[obj_type][''] or {}
+            intermediate[obj_type][''][perm] = true
+            return
+        end
```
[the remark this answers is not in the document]
Lord-KA: That's a good idea, thanks! Applied. Now `'all'` is just another object name, that could be added with others. In your example that would mean, granting `'execute'` to functions `'box.info'`, `'box.stat'`, `'metrics.getmetrics'` and all functions from `_func`.

src/box/lua/config/applier/credentials.lua
Lord-KA: I don't want to have e.g. `'functions', 'all'` and `'universe'` being stored differently, so no.

src/box/lua/config/applier/credentials.lua
Lord-KA: Fixed.

src/box/lua/config/applier/credentials.lua
Lord-KA: Inside schema it doesn't feel very natural to have both `'universe'` and some objects in one privilege. But as far as I can recall, there is no verification and I don't think that applier should just ignore those objects. So yea, lets add them all here.

Lord-KA: Generally, I think, that the same scenario as you mentioned above can be applied here too:
```yaml
credentials:
  users:
    myuser:
      privileges:
      - permissions: [execute]
        functions:
        - some.func
        # Temporary grant all functions for testing purposes.
        universe: true
```

So, verification can be left as is too.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+
+        for _, obj_name  in ipairs(obj_def) do
+            intermediate[obj_type][obj_name] =
+                intermediate[obj_type][obj_name] or {}
+            intermediate[obj_type][obj_name][perm] = true
```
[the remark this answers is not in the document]
Lord-KA: I've managed to create a single `privileges_add_perm()` function out of two local `add_permission()` in `config_convert()` and `add_defaults()`. Note that I left similar code for role assignment with an accommodating comment:
```lua
-- Unlike spaces, functions and sequences, role is allowed to be
-- named 'all', so `privileges_add_perm()` isn't used.
```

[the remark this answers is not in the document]
Lord-KA: And please, lets have all naming discussions in the according thread below.

test/config-luatest/credentials_applier_test.lua
```diff
@@ -0,0 +1,471 @@
...
+        {'role', 'super'},
+        {'role', '<newly_created>'},
+    }
+
+    for _, case in ipairs(cases) do
+        local role_or_user, name = unpack(case)
+
+        local child = it.new()
+        -- TODO check if it is done in temp dir
+        -- maybe look into config-luatest helpers
```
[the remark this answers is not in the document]
Lord-KA: `fio.cwd()` actually returns current dir, so I passed `work_dir=...` to `box.cfg{}`.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+-- Find all privileges that current is lacking compared to target.
+-- Return them as a list in the following format:
+-- - obj_type: my_type
+--   obj_name: my_name
+--   privs:
+--    - read
+--    - write
+--    - ...
+--
+local function privileges_lack(current, target)
```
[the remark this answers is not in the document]
Lord-KA: We've discussed naming verbally and decided to rename `privileges_lack(current, target)` to `privileges_subtract(target, current)`. I've updated it and surrounding comments.

src/box/lua/config/applier/credentials.lua
Lord-KA: We've discussed naming verbally and came to the following agreement:
1) `privileges_` prefix for all helper functions working with intermediate representation (besides `sync_privileges()` because it is part of applier logic more than ir).
2) `privileges_box_convert()` -> `privileges_from_box`
3) `privileges_config_convert()` -> `privileges_from_config`
4) `privileges_lack(current, target)` -> `privileges_subtract(target, current)`

test/config-luatest/credentials_applier_test.lua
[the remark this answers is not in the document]
Lord-KA: Fixed.

src/box/lua/config/applier/credentials.lua
```diff
@@ -1,30 +1,326 @@
...
+    -- applier is to not to revoke them. For reasoning consider the following
+    -- scenarios:
+    --  * Admin doesn't change the defaults, then they are set by box
+    --    on startup and don't need to be updated
+    --  * Admin revokes some of the defaults, then config shouldn't restore
+    --    them without explicit command (adding them to config file)
+    -- Thus, config_privileges should be enriched with defaults only after
+    -- all of the grants are calculated.
+    config_privileges = privileges_add_defaults(name, role_or_user,
+                                                config_privileges)
```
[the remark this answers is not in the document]
Lord-KA: Thanks! I like your version, so applied.

src/box/lua/config/applier/credentials.lua
Lord-KA: (The helper functions working with intermediate representation don't need to know the difference between object types (e.g. `'function'`, `'space'`, `'universe'`), when handling `'all'` case (`''` in ir))

src/box/lua/config/applier/credentials.lua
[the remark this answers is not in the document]
Totktonada: OK.

src/box/lua/config/applier/credentials.lua
Totktonada: OK, got it.

Totktonada (approved):

test/config-luatest/credentials_applier_test.lua
```diff
@@ -0,0 +1,480 @@
...
+            obj_name = 'myseq1',
+            privs = {'read', 'write'},
+        },
+    }
+
+    t.assert_items_equals(internal.privileges_subtract(target, current), lack)
+end
+
+g.test_privileges_add_defaults = function(g)
+    local cases = {
```
ylobankov, line 372: I don't insist, but it looks like it can be parametrized by luatest facilities.

ylobankov (approved):

test/config-luatest/credentials_applier_test.lua
Lord-KA: I would prefer to leave it as is. There is no need for extra parallelization and the test would look a bit more complex than it needs to be.

Totktonada: I agree that we have a trade-off between output beauty (plus ability to run test cases separately) and readability of the testing code. I usually don't insist on parametrization using luatest facilities for small test cases.

I'm going to accept the patch as is.

PR #8971, opened 2023-08-09
Silence coverity in instance names code

Gerold103:
Coverity complained about a couple of places. The patch silences them.

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -2336,8 +2336,14 @@ box_set_instance_name(void)
 	strlcpy(old_cfg_name, cfg_instance_name, NODE_NAME_SIZE_MAX);
 	auto guard = make_scoped_guard([&]{
 		strlcpy(cfg_instance_name, old_cfg_name, NODE_NAME_SIZE_MAX);
-		box_restart_replication();
-		replicaset_follow();
+		try {
+			box_restart_replication();
+			replicaset_follow();
+		} catch (Exception *exc) {
+			exc->log();
+		} catch (...) {
+			panic("Unknown exception on instance name set failure");
+		}
 	});
 	strlcpy(cfg_instance_name, name, NODE_NAME_SIZE_MAX);
 	/* Nil means the config doesn't care, allows to use any name. */
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4606,6 +4606,7 @@ on_replace_dd_cluster(struct trigger *trigger, void *event)
 			return on_replace_dd_cluster_update(old_def, new_def);
 		return on_replace_dd_cluster_insert(new_def);
 	}
+	assert(old_def != NULL);
 	return on_replace_dd_cluster_delete(old_def);
 }
 


src/box/box.cc
```diff
@@ -2336,8 +2336,14 @@ box_set_instance_name(void)
 	strlcpy(old_cfg_name, cfg_instance_name, NODE_NAME_SIZE_MAX);
 	auto guard = make_scoped_guard([&]{
 		strlcpy(cfg_instance_name, old_cfg_name, NODE_NAME_SIZE_MAX);
-		box_restart_replication();
-		replicaset_follow();
+		try {
+			box_restart_replication();
+			replicaset_follow();
```
sergepetrenko, line 2341: It seems only `box_restart_replication()` may throw an error, no?
Did you put both function calls inside the try catch on purpose?

Gerold103: No reason, I just put the whole code block into try-catch. Do you want to move the follow outside?

sergepetrenko: > No reason, I just put the whole code block into try-catch. Do you want to move the follow outside?

No, it's fine. It would look rather ugly outside.

> Perhaps we should consider adding noexcept to our functions to catch these things at compile time, not at coverity-check time. WDYT?

It's a good idea, but I tried experimenting with it a bit and it seems the compiler isn't that good at checking this (I tried this on apple clang 14.0.0). For example, the only way to generate a compile warning / error was explicit `throw` in the function body:
```diff
@@ -2334,10 +2335,11 @@ box_set_instance_name(void)
                return;
        char old_cfg_name[NODE_NAME_SIZE_MAX];
        strlcpy(old_cfg_name, cfg_instance_name, NODE_NAME_SIZE_MAX);
-       auto guard = make_scoped_guard([&]{
+       auto guard = make_scoped_guard([&]() noexcept {
                strlcpy(cfg_instance_name, old_cfg_name, NODE_NAME_SIZE_MAX);
                box_restart_replication();
                replicaset_follow();
+               throw 1;
        });
```
Even adding a `noexcept(false)` to `box_restart_replication()` and throwing explicitly right in the function body isn't detected. This compiles without warnings:
```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 2a8c57028..7218668b9 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -1973,11 +1973,12 @@ box_sync_replication(bool do_quorum, bool do_reuse)
 }
 
 static inline void
-box_restart_replication(void)
+box_restart_replication(void) noexcept(false)
 {
        const bool do_quorum = true;
        const bool do_reuse = false;
        box_sync_replication(do_quorum, do_reuse);
+       throw 1;
 }
 
 static inline void
@@ -2334,7 +2335,7 @@ box_set_instance_name(void)
                return;
        char old_cfg_name[NODE_NAME_SIZE_MAX];
        strlcpy(old_cfg_name, cfg_instance_name, NODE_NAME_SIZE_MAX);
-       auto guard = make_scoped_guard([&]{
+       auto guard = make_scoped_guard([&]() noexcept {
                strlcpy(cfg_instance_name, old_cfg_name, NODE_NAME_SIZE_MAX);
                box_restart_replication();
                replicaset_follow();
```

Gerold103: Then I propose we pretend `noexcept` doesn't exist. Anyway C++ will be purged at some point. Any other comments to the PR?

sergepetrenko: No. Everything looks fine. Thanks!

sergepetrenko (approved):

PR #8973, opened 2023-08-10
box: forbid same names for check constraints and foreign keys

### LINKED ISSUE #7650: Forbid same names for check constraints and foreign keys
Now a user can create a set of check constraints and foreign keys for each field and entire tuple. By design for each field/tuple the check constraints must have unique names, the same is true for foreign keys. But now it is possible to create a check constraint and a foreign key with the same name.

This approach is good, but introduces difficulties in SQL implementation. On the other hand in real life it is hard to imagine a user that wants to create both check and foreign constraints for a field, especially with the same names. As for entire tuple it seems less obvious, but still it seems that there will be no harm if we forbid same names.

So, let's forbid same names for check constraints and foreign keys.

Gumix:
It is possible to create a set of check constraints and foreign keys for each field and for entire tuple. By design for each field/tuple the check constraints must have unique names, the same is true for foreign keys. But it is possible to create a check constraint and a foreign key with the same name. This introduces difficulties in SQL implementation. On the other hand in real life it is hard to imagine a user that wants to create both check and foreign constraints for a field, especially with the same names. As for entire tuple it seems less obvious, but still it seems that there will be no harm if we forbid same names.

Closes #7650

### DIFF
--- a/changelogs/unreleased/gh-7650-forbid-same-names-for-constraints-and-fkeys.md
+++ b/changelogs/unreleased/gh-7650-forbid-same-names-for-constraints-and-fkeys.md
@@ -1,0 +1,4 @@
+## feature/core
+
+* It is now forbidden for a check constraint and a foreign key of each field
+  (or entire tuple) to have the same name (gh-7650).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -428,6 +428,12 @@ space_opts_decode(struct space_opts *opts, const char *map,
 			 diag_last_error(diag_get())->errmsg);
 		return -1;
 	}
+	if (tuple_constraint_def_array_check(opts->constraint_def,
+					     opts->constraint_count) != 0) {
+		diag_set(ClientError, ER_WRONG_SPACE_OPTIONS,
+			 diag_last_error(diag_get())->errmsg);
+		return -1;
+	}
 	return 0;
 }
 
--- a/src/box/tuple_constraint_def.c
+++ b/src/box/tuple_constraint_def.c
@@ -13,6 +13,7 @@
 #include "salad/grp_alloc.h"
 #include "small/region.h"
 #include "msgpuck.h"
+#include "tt_static.h"
 
 #include <PMurHash.h>
 
@@ -563,4 +564,23 @@ tuple_constraint_def_array_dup_raw(const struct tuple_constraint_def *defs,
 	/* If we did it correctly then there is no more space for strings. */
 	assert(grp_alloc_size(&all) == 0);
 	return res;
+}
+
+int
+tuple_constraint_def_array_check(const struct tuple_constraint_def *defs,
+				 size_t count)
+{
+	for (uint32_t i = 0; i < count; i++) {
+		const struct tuple_constraint_def *c1 = &defs[i];
+		for (uint32_t j = i + 1; j < count; j++) {
+			const struct tuple_constraint_def *c2 = &defs[j];
+			if (strcmp(c1->name, c2->name) == 0) {
+				diag_set(IllegalParams, tt_sprintf(
+					"duplicate constraint name '%s'",
+					c1->name));
+				return -1;
+			}
+		}
+	}
+	return 0;
 }
--- a/src/box/tuple_constraint_def.h
+++ b/src/box/tuple_constraint_def.h
@@ -185,6 +185,17 @@ tuple_constraint_def_array_dup_raw(const struct tuple_constraint_def *defs,
 				   size_t count, size_t object_size,
 				   size_t additional_size);
 
+/**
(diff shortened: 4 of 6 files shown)


changelogs/unreleased/gh-7650-forbid-same-names-for-constraints-and-fkeys.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* Now it is forbidden for a check constraint and a foreign key of each field
+  (or entire tuple) to have the same name (gh-7650).
```
p7nov, line 4:
```suggestion
* It is now forbidden for a check constraint and a foreign key of each field
  (or entire tuple) to have the same name (gh-7650).
```

p7nov (approved):
A minor changelog improvement.

src/box/tuple_constraint_def.c
```diff
@@ -564,3 +564,20 @@ tuple_constraint_def_array_dup_raw(const struct tuple_constraint_def *defs,
...
 	return res;
 }
+
+bool
+tuple_constraint_def_check_names(const struct tuple_constraint_def *defs,
+				 size_t count)
+{
+	if (count < 2)
+		return true;
+	for (uint32_t i = 0; i < count - 1; i++) {
```
locker, line 574: Nit:

```suggestion
	for (uint32_t i = 0; i < count; i++) {
```

src/box/tuple_constraint_def.h
```diff
@@ -185,6 +185,14 @@ tuple_constraint_def_array_dup_raw(const struct tuple_constraint_def *defs,
 				   size_t count, size_t object_size,
 				   size_t additional_size);
 
+/**
+ * Check that all constraints in the `defs` array have unique names.
+ * Return true if the check passed, false otherwise.
+ */
+bool
+tuple_constraint_def_check_names(const struct tuple_constraint_def *defs,
+				 size_t count);
```
locker, line 194: This function should be named `tuple_constraint_def_array_check` because it checks an array. I'd omit the `_names` suffix for the sake of extensibility. Also, it should return 0 or -1 and set diag that would include the duplicate constraint name. You can use `IllegalParams` error and convert it to the right error code like it's done here:

https://github.com/tarantool/tarantool/blob/78c8cab84197f4a55298d20b7707bd6436c71552/src/box/field_def.c#L349-L350

src/box/alter.cc
```diff
@@ -534,6 +534,12 @@ space_def_new_from_tuple(struct tuple *tuple, uint32_t errcode,
 			 "local space can't be synchronous");
 		return NULL;
 	}
+	if (!tuple_constraint_def_check_names(opts.constraint_def,
+					      opts.constraint_count)) {
+		diag_set(ClientError, ER_WRONG_SPACE_OPTIONS, "check "
+			 "constraints and foreign keys must have unique names");
+		return NULL;
+	}
```
locker, line 542: Better put it in `space_opts_decode`, I think, because that's where we create constraints.

locker (changes requested):

ochaplashkin (approved):
LGTM :rocket:

changelogs/unreleased/gh-7650-forbid-same-names-for-constraints-and-fkeys.md
Gumix: Done.

src/box/tuple_constraint_def.c
Gumix: Done.

src/box/tuple_constraint_def.h
Gumix: Done.

src/box/alter.cc
Gumix: Done.

changelogs/unreleased/gh-7650-forbid-same-names-for-constraints-and-fkeys.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* It is now forbidden for a check constraint and a foreign key of each field
+  (or entire tuple) to have the same name (gh-7650).
```
locker, line 4: I'm not sure this change is 100% safe because there may be installations that use duplicate names for some reason. It may be worth disabling the checks during recovery or introduce a `compat` module option for it. Up to @alyapunov.

locker (approved):
Looks good code-wise but I have my reservations about backward compatibility (see the comment below). Need to discuss with @alyapunov.

Gumix:
New decision:
- https://github.com/tarantool/tarantool/issues/7650#issuecomment-1710048271

PR #8977, opened 2023-08-11
core: fix `ASAN_START_SWITCH_FIBER()` usage

Gumix:
The `__sanitizer_start_switch_fiber()` function takes a pointer as the first argument to store the current fake stack if there is one (it is necessary when stack-use-after-return detection is enabled). When leaving a fiber definitely, `NULL` must be passed so that the fake stack is destroyed.

Before this patch, `NULL` was passed for dead fibers, however this is wrong for dead fibers that are recycled and resumed. In such cases ASAN destroys the fake stack, and the fiber crashes trying to use it in `fiber_yield()` upon return from `coro_transfer()`.

Closes tarantool/tarantool-qa#321

### DIFF
--- a/changelogs/unreleased/ghqa-321-segfault-with-clang-16-asan.md
+++ b/changelogs/unreleased/ghqa-321-segfault-with-clang-16-asan.md
@@ -1,0 +1,5 @@
+## bugfix/core
+
+* Fixed a crash that could happen when Tarantool is compiled by `clang`
+  version 15 and above with enabled AddressSanitizer
+  (tarantool/tarantool-qa#321).
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -143,21 +143,27 @@ static int (*fiber_invoke)(fiber_func f, va_list ap);
 #if ENABLE_ASAN
 #include <sanitizer/asan_interface.h>
 
-#define ASAN_START_SWITCH_FIBER(var_name, will_switch_back, bottom, size) \
-	void *var_name = NULL; \
-	__sanitizer_start_switch_fiber((will_switch_back) ? &var_name : NULL, \
+#define ASAN_START_SWITCH_FIBER(fake_stack_save, will_switch_back, bottom,     \
+				size)					       \
+	/*								       \
+	 * When leaving a fiber definitely, NULL must be passed as the first   \
+	 * argument so that the fake stack is destroyed.		       \
+	 */								       \
+	void *fake_stack_save = NULL;					       \
+	__sanitizer_start_switch_fiber((will_switch_back) ? &fake_stack_save   \
+							  : NULL,	       \
                                        (bottom), (size))
 #if ASAN_INTERFACE_OLD
-#define ASAN_FINISH_SWITCH_FIBER(var_name) \
-	__sanitizer_finish_switch_fiber(var_name);
+#define ASAN_FINISH_SWITCH_FIBER(fake_stack_save) \
+	__sanitizer_finish_switch_fiber(fake_stack_save)
 #else
-#define ASAN_FINISH_SWITCH_FIBER(var_name) \
-	__sanitizer_finish_switch_fiber(var_name, 0, 0);
+#define ASAN_FINISH_SWITCH_FIBER(fake_stack_save) \
+	__sanitizer_finish_switch_fiber(fake_stack_save, NULL, NULL)
 #endif
 
 #else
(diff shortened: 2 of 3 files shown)


p7nov (approved):

src/lib/core/fiber.c
```diff
@@ -1030,14 +1052,22 @@ fiber_loop(MAYBE_UNUSED void *data)
 		fiber_on_stop(fiber);
 		/* reset pending wakeups */
 		rlist_del(&fiber->state);
+		bool is_reusable = (fiber->flags & FIBER_CUSTOM_STACK) == 0;
```
locker, line 1055: You have two checks if a fiber is reusable - here and in `fiber_join`. Let's add an inline helper function `fiber_is_reusable` with a proper comment so that we don't have to patch them both if we need to do some other kind of non-reusable fibers. The `is_reusable` argument of `fiber_recycle` looks redundant - we could use the helper function there as well.

changelogs/unreleased/ghqa-321-segfault-with-clang-16-asan.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+* Fixed a crash that could happen when Tarantool is compiled by `clang`
+  version 16 with enabled AddressSanitizer (tarantool/tarantool-qa#321).
```
locker, line 4: It fails with clang 15 and above.

src/lib/core/fiber.c
```diff
@@ -143,21 +143,27 @@ static int (*fiber_invoke)(fiber_func f, va_list ap);
...
 #if ASAN_INTERFACE_OLD
-#define ASAN_FINISH_SWITCH_FIBER(var_name) \
-	__sanitizer_finish_switch_fiber(var_name);
+#define ASAN_FINISH_SWITCH_FIBER(fake_stack_save) \
+	__sanitizer_finish_switch_fiber(fake_stack_save)
 #else
-#define ASAN_FINISH_SWITCH_FIBER(var_name) \
-	__sanitizer_finish_switch_fiber(var_name, 0, 0);
+#define ASAN_FINISH_SWITCH_FIBER(fake_stack_save) \
+	__sanitizer_finish_switch_fiber(fake_stack_save, 0, 0)
```
locker, line 161: Nit: Should `NULL, NULL`, not `0, 0`.

locker (changes requested):

src/lib/core/fiber.h
```diff
@@ -269,6 +269,8 @@ fiber_new_ex(const char *name, const struct fiber_attr *fiber_attr, fiber_func f
 
 /**
  * Return control to another fiber and wait until it'll be woken.
+ * Note: this is not a cancellation point (see `fiber_testcancel()`), but it is
+ * considered a good practice to call `fiber_testcancel()` after each yield.
```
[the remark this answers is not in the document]
Gumix: Done.

src/lib/core/fiber.c
Gumix: Done.

changelogs/unreleased/ghqa-321-segfault-with-clang-16-asan.md
Gumix: Changed to "clang version 15 and above".

src/lib/core/fiber.c
Gumix: Done.

src/lib/core/fiber.c
```diff
@@ -1329,13 +1365,10 @@ fiber_new_ex(const char *name, const struct fiber_attr *fiber_attr,
...
 
-	/* Now we can not reuse fiber if custom attribute was set */
-	if (!(fiber_attr->flags & FIBER_CUSTOM_STACK) &&
-	    !rlist_empty(&cord->dead)) {
-		fiber = rlist_first_entry(&cord->dead,
-					  struct fiber, link);
+	if (fiber_is_reusable(fiber_attr->flags) && !rlist_empty(&cord->dead)) {
+		fiber = rlist_first_entry(&cord->dead, struct fiber, link);
 		rlist_move_entry(&cord->alive, fiber, link);
-		assert((fiber->flags | FIBER_IS_DEAD) != 0);
```
Gumix, line 1338: Fixed a bug in this assertion.

locker (approved):

locker:

PR #8979, opened 2023-08-13
box: fix invalid memory access in tuple_compare_with_key_sequential (2.10 special)

mkostoevr:
1. Since number type was introduced we can not assume if tuples are equal by comparison then their sizes are equal too. So the place the assumption is used is fixed. These changes has been taken from [#8922](https://github.com/tarantool/tarantool/pull/8922).
2. Partially applied the ca832f278855554f2741e54a997cece9bb7e2197 commit. Effectively it just introduced the `key_compare_and_skip_parts` function that is required for [#8922](https://github.com/tarantool/tarantool/pull/8922) and created the key_def test. Changes to the commit in details:
   
   src/box/tuple_compare.cc:
   - Omitted the `func_index_compare_with_key` update.

   test/unit/key_def.c:
   - Omitted four tests that checked the `func_index_compare_with_key` update.

### DIFF
--- a/changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
+++ b/changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed an invalid memory access in a corner case of a specialized comparison
+  function (gh-8899).
--- a/src/box/tuple_compare.cc
+++ b/src/box/tuple_compare.cc
@@ -929,8 +929,9 @@ tuple_compare_with_key_sequential(struct tuple *tuple, hint_t tuple_hint,
 		assert(field_count >= part_count);
 		cmp_part_count = part_count;
 	}
-	rc = key_compare_parts<is_nullable>(tuple_key, key, cmp_part_count,
-					    key_def);
+	rc = key_compare_and_skip_parts<is_nullable>(&tuple_key, &key,
+						     cmp_part_count,
+						     key_def);
 	if (!has_optional_parts || rc != 0)
 		return rc;
 	/*
@@ -938,11 +939,6 @@ tuple_compare_with_key_sequential(struct tuple *tuple, hint_t tuple_hint,
 	 * corresponding key fields to be equal to NULL.
 	 */
 	if (field_count < part_count) {
-		/*
-		 * Key's and tuple's first field_count fields are
-		 * equal, and their bsize too.
-		 */
-		key += tuple_bsize(tuple) - mp_sizeof_array(field_count);
 		for (uint32_t i = field_count; i < part_count;
 		     ++i, mp_next(&key)) {
 			if (mp_typeof(*key) != MP_NIL)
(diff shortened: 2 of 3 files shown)


test/unit/key_def.c
```diff
@@ -0,0 +1,267 @@
...
+	/*
+	 * Model def is a copy of cmp_def, but not for_func_index, and hence
+	 * it has general implementation of tuple_compare_with_key method.
+	 */
+	struct key_def *model_def = test_key_def_new(
+		"[{%s%u%s%s}{%s%u%s%s}{%s%u%s%s}{%s%u%s%s}]",
+		"field", 0, "type", "unsigned",
+		"field", 1, "type", "string",
+		"field", 3, "type", "unsigned",
+		"field", 4, "type", "string");
```
drewdzzz, line 186: The test still contains some logic of primary key comparison check. I think it's better not to export it at all.

drewdzzz (changes requested):

test/unit/key_def.c
mkostoevr: Agreed. Replaced the test with comparator tests from #8915, they fully cover the `func_index_compare_with_key` and other comparators.

drewdzzz: Wow, it's quite a large test suite. Let's just export my commit without tests :)

drewdzzz: The test suit will be reviewed in the scope of #8915, and the commit is trivial - just factoring out some logic. And it will be covered by other tests (`key_compare_parts` wrapper is already covered, `key_compare_and_skip_parts` will be covered in the next commit).

changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* Fixed the invalid memory access in a corner case of a specialized comparison
+  function (gh-8899).
```
p7nov, line 4:
```suggestion
* Fixed an invalid memory access in a corner case of a specific comparison
  function (gh-8899).
```

Or maybe mention the function name explicitly, if it is visible to users.

p7nov (approved):

drewdzzz (approved):
LGTM

alyapunov (approved):

alyapunov:
LGTM
Please take into account comment from doc team.

ylobankov (approved):

changelogs/unreleased/gh-8899-compare-with-key-slowpath-invalid-memory-access.md
mkostoevr: Hi, fixed the article. Can you please verify the new changelog?

The comparison function is specialized on non-sequential keys and keys with JSON paths.

This is why the function is called "specialized" - it only works with the keys satisfying its requirements.

alyapunov: I bet everything is fine here.
Actually this is backport to 2.10, so we have already passed doc check when pushed #8922.

PR #8984, opened 2023-08-14
Introduce `box.session.new`

### LINKED ISSUE #8801: Create server IPROTO connection over socket fd
This is required to start IPROTO connection over a socket created with the `socketpair` system call.

```Lua
-- Starts server IPROTO connection over the given fd.
box.session.new({type = 'binary', fd = socket_fd})

-- Same but authenticate the client session as 'admin'.
box.session.new({type = 'binary', fd = socket_fd, user = 'admin'})
```

locker:
The new function creates a new session given a table of options:

 - `type`: string, optional. Default: "binary". [Type][session-type] of the new session. Currently, only "binary" is supported, which creates a new IPROTO session.

 - `fd`: number, mandatory. File descriptor number (fd) to be used for the new connection input-output. The fd must refer to a [socket] and be switched to the [non-blocking mode][socket-nonblock] but this isn't enforced, i.e. the user may pass an invalid fd, in which case the connection won't work as expected.

 - `user`: string, optional. Default: "guest". Name of the user to authenticate the new session as. Note, this doesn't prevent the other end to perform explicit [authentication].

 - `storage`: table, optional. Default: empty table. Initial value of the [session-local storage][session-storage].

On success, `box.session.new` takes ownership of the `fd` and returns nothing. On failure, an error is raised.

Possible errors:
 - Invalid option value type.
 - `fd` isn't specified or has an invalid value.
 - `box.cfg` wasn't called.
 - `user` doesn't exist.

Example:

The code below creates a TCP server that accepts all incoming IPROTO connections on port 3301, authenticates them as 'admin' and sets the session-local storage to `{foo = 'bar'}`.

```lua
box.cfg()
require('socket').tcp_server('localhost', 3301, function(s)
    box.session.new({
        type = 'binary',
        fd = s:fd(),
        user = 'admin',
        storage = {foo = 'bar'},
    })
    s:detach()
end)
```

Notes:
- `box.cfg` must be called before using `box.session.new` to start IPROTO threads. Setting [`box.cfg.listen`][box-cfg-listen] isn't required though.
- The socket object must be detached after passing its fd to `box.session.new`, otherwise the fd would be closed on Lua garbage collection.

Closes #8801

[authentication]: https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/authentication/
[box-cfg-listen]: https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-basic-listen
[session-storage]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_session/storage/
[session-type]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_session/type/
[socket]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/
[socket-nonblock]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-nonblock

### DIFF
--- a/changelogs/unreleased/gh-8801-iproto-session-from-fd.md
+++ b/changelogs/unreleased/gh-8801-iproto-session-from-fd.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Introduced the new function `box.session.new` for creating a new IPROTO
+  session from a socket file descriptor number (gh-8801).
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -358,6 +358,12 @@ struct iproto_msg
 			};
 			/** Peer address size. */
 			socklen_t addrlen;
+			/**
+			 * Session to use for the new connection.
+			 * Optional. If omitted, a new session object
+			 * will be created in the TX thread.
+			 */
+			struct session *session;
 		} connect;
 		/** Box request, if this is a DML */
 		struct request dml;
@@ -2760,7 +2766,12 @@ tx_process_connect(struct cmsg *m)
 	struct iproto_msg *msg = (struct iproto_msg *) m;
 	struct iproto_connection *con = msg->connection;
 	struct obuf *out = msg->connection->tx.p_obuf;
-	con->session = session_new(SESSION_TYPE_BINARY);
+	if (msg->connect.session != NULL) {
+		con->session = msg->connect.session;
+		session_set_type(con->session, SESSION_TYPE_BINARY);
+	} else {
+		con->session = session_new(SESSION_TYPE_BINARY);
+	}
 	con->session->meta.connection = con;
 	session_set_peer_addr(con->session, &msg->connect.addr,
 			      msg->connect.addrlen);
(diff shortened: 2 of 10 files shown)


changelogs/unreleased/gh-8801-iproto-session-from-fd.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Introduced new function `box.session.new` for creating a new IPROTO session
+  from a socket file descriptor number (gh-8801).
```
p7nov, line 4:
```suggestion
* Introduced the new function `box.session.new` for creating a new IPROTO
  session from a socket file descriptor number (gh-8801).
```

p7nov (approved):

changelogs/unreleased/gh-8801-iproto-session-from-fd.md
locker: Fixed. Thanks!

alyapunov (approved):

ylobankov (approved):
Tests LGTM

PR #8988, opened 2023-08-15
Introduce `net.box.from_fd`

### LINKED ISSUE #8928: Create net.box connection from socket fd
This will be useful for establishing an IPROTO connection between a process started with `popen` and its parent over a socket created with the `socketpair` system call.

```Lua
local conn = require('net.box').from_fd(socket_fd)
```

locker:
The function takes a file descriptor number (fd) as its first argument and a table of connection options as its second, optional argument. It creates and returns a new connection object.

The fd should point to a socket and be switched to the non-blocking mode, but this isn't enforced. If the fd is invalid, the connection may not work as expected.

The functions takes the same options as [`connect`][net-box-connect].

The function takes the ownership of the fd, i.e. the fd must not be used or closed after this function successfully returns.

Example:

The code below connects to a Tarantool instance at port 3301 using the [`socket`][socket] module, then wraps the socket fd in a `net.box` connection object.

```Lua
local net = require('net.box')
local socket = require('socket')

-- Connect a socket then wrap it in a net.box connection object.
local s = socket.tcp_connect('localhost', 3301)
assert(s ~= nil)
local conn = net.from_fd(s:fd(), {fetch_schema = false})
s:detach()

conn:call('box.schema.user.info')
conn:close()
```

Closes #8928

[net-box-connect]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#net-box-connect
[socket]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/


p7nov (approved):

alyapunov (approved):

ylobankov (approved):
Tests LGTM

PR #9000, opened 2023-08-18
config: introduce initial vshard support

### LINKED ISSUE #9007: Add an ability to start vshard though config
This issue proposes adding the ability to call a configured vshard via the `config` module.

ImeevMA:
This path-set introduces initial vshard support in config.

Closes #9007


src/box/lua/config/instance_config.lua
```diff
@@ -1513,6 +1513,139 @@ return schema.new('instance_config', schema.record({
...
+        lock = schema.scalar({
+            type = 'boolean',
+            validate = function(_data, w)
+                local scope = w.schema.scope
+                if scope ~= nil and scope ~= 'global' and scope ~= 'replicaset'
+                   and scope ~= 'instance' then
+                    w.error('sharding.lock should be a defined in '..
+                            'replicaset, group or global scope')
+                end
+            end,
```
Totktonada, line 1533: I suspect that this scope validation (and all the similar functions below) doesn't work. There are several reasons.

1. `scope` annotation is present only in the top level record. Please, consider using of `w.computed.annotations.scope`.
2. `data` is not checked against `nil`/`box.NULL`.
3. The `scope` variable check itself strange. The error message suggests that we should allow `global`, `group` and `replicaset`, but actually we allow `global`, `replicaset` and `instance` (and not `group`).
4. I would also note that I don't understand what the 'cluster vshard options' means. I would understand it if the option would be only allowed in the `global` scope.

There is a tricky point in the scope validation. It seems, we should allow `scope == nil` always, because otherwise a cluster config instantiation into instance config would produce invalid instance config. However, it doesn't mean that we can't forbid something in `scope == 'instance'`.

src/box/lua/config/instance_config.lua
```diff
@@ -1513,6 +1513,139 @@ return schema.new('instance_config', schema.record({
             value = schema.scalar({type = 'string'}),
         }),
     }),
+    sharding = schema.record({
+        -- Instance vshard options.
+        roles = schema.set({
+            'router',
+            'storage',
+        }),
```
Totktonada, line 1521: Shouln't we allow it on the replicaset level, but not on the instance level?

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -241,6 +241,8 @@ local function apply(config)
     box_cfg.cluster_name = names.group_name
     box_cfg.replicaset_name = names.replicaset_name
     box_cfg.instance_name = names.instance_name
+    box_cfg.instance_uuid = configdata._instance_uuid
+    box_cfg.replicaset_uuid = configdata._replicaset_uuid
```
Totktonada, line 245: I prefer to use underscored fields from `configdata` only inside `configdata` itself. Can we add `configdata:uuids()` simarly to `configdata:names()` to even just add the UUIDs into the `:names()` result?

src/box/lua/config/init.lua
```diff
@@ -150,6 +150,7 @@ function methods._initialize(self)
     self:_register_applier(require('internal.config.applier.console'))
     self:_register_applier(require('internal.config.applier.fiber'))
     self:_register_applier(require('internal.config.applier.app'))
+    self:_register_applier(require('internal.config.applier.sharding'))
```
Totktonada, line 153: We should configure sharding first and only them call the user provided code (application).

(I know, the `app` applier actually performs all its logic in the `post_apply()` hook, but I would anyway place `app` at the last position.)

src/box/lua/config/configdata.lua
```diff
@@ -65,6 +66,116 @@ function methods.names(self)
     }
 end
 
+local function new_uuid(str)
+    local sha = digest.sha1_hex(str)
+    return sha:sub(1,8)..'-'..sha:sub(9,12)..'-'..sha:sub(13,16)..'-'..
+           '00'..sha:sub(17,18)..'-'..sha:sub(19,30)
+end
```
Totktonada, line 73: I found the name a bit confusing. It is not just some new uuid. It is a kind of a hash function and it is fully determined by the given source string.

How about something like `sha1_as_uuid()`? Or, maybe, `uuid_from_name()`?

src/box/lua/config/applier/sharding.lua
```diff
@@ -0,0 +1,71 @@
...
+        return
+    end
+    if _G.vshard == nil then
+        _G.vshard = require('vshard')
+    end
+    for _, role in pairs(roles) do
+        local cfg = configdata:sharding()
+        cfg.listen = box.cfg.listen
+        cfg.read_only = box.cfg.read_only
+        cfg.replication = box.cfg.replication
```
Totktonada, line 17: I think that this workaround should be commented, ideally with a link to vshard's issue.

src/box/lua/config/applier/sharding.lua
```diff
@@ -0,0 +1,71 @@
...
+local function wait_ro_rw(configdata)
+    while true do
+        if box.info.ro then
+            box.ctl.wait_rw()
+        else
+            box.ctl.wait_ro()
+        end
+        vshard_cfg(configdata)
+    end
+end
```
Totktonada, line 60: If `vshard_cfg()` raises an error, the fiber exited. Is it intentional?

src/box/lua/config/applier/sharding.lua
```diff
@@ -0,0 +1,71 @@
...
+
+local function apply(config)
+    local configdata = config._configdata
+    vshard_cfg(configdata)
+    fiber.create(wait_ro_rw, configdata)
+end
```
Totktonada, line 66: So, we get many fibers that're trying to apply its own config version in some unspecified order. I guess that we should get actual config from a background always-existing fiber. Maybe we also need a check, whether given config version is applied in the given ro/rw state to eliminate races.

changelogs/unreleased/gh-8861-vshard-options.md
```diff
@@ -1,3 +1,4 @@
 ## feature/config
 
 * Most of vshard options are now added in the config (gh-8861).
+* Added initial support of vshard (gh-8861).
```
Totktonada, line 4: > 4) The router can automatically find all masters, but once all masters
> are found, any changes to the masters will be ignored until
> vshard.router.cfg() is called manually.

TBH, I don't get why it is so (maybe we talked around, but I forgot). Is it some vshard problem? Is it tracked in an issue?

andreyaksenov (approved):

src/box/lua/config/applier/box_cfg.lua
ImeevMA: Added to names().

src/box/lua/config/applier/sharding.lua
ImeevMA: Added a comment.

src/box/lua/config/applier/sharding.lua
ImeevMA: No, but I am not sure what should be do in this case. Should we raise an error in Tarantool?

src/box/lua/config/applier/sharding.lua
ImeevMA: Should we do this as a separate issue?

src/box/lua/config/init.lua
ImeevMA: True, fixed.

src/box/lua/config/configdata.lua
ImeevMA: Renamed to `uuid_from_name()`.

changelogs/unreleased/gh-8861-vshard-options.md
ImeevMA: I am not sure if this actually should work. And isn't this part of one of master auto-discovering issues?

ImeevMA (commented):
Thank you for the review!

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -1,121 +1,11 @@
-local urilib = require('uri')
 local fio = require('fio')
```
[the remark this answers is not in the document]
Totktonada: Please, reflect it in the PR description too.

src/box/lua/config/instance_config.lua
Totktonada: I don't insist on prohibiting right now, but I would mark the options as related to a replicaset (even if they're currently accepted for an instance).

changelogs/unreleased/gh-8861-vshard-options.md
Totktonada: In fact, I just want to understand what is going on. My thought was that everything should be fine with master discovery on routers. Now I'm not sure and I afraid that we can find more problems if we look inside.

src/box/lua/config/applier/box_cfg.lua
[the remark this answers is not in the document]
ImeevMA: Added in the PR description.

src/box/lua/config/instance_config.lua
```diff
@@ -232,6 +232,121 @@ local function advertise_peer_uri_validate(data, w)
     return uri
 end
 
+-- Accept a comma separated list of URIs and return the first one
+-- that is suitable to create a client socket (not just to listen
+-- on the server as, say, 0.0.0.0:3301 or localhost:0).
+--
+-- See the uri_is_suitable_to_connect() method in the instance
+-- schema object for details.
+local function find_suitable_uri_to_connect(self, uris)
```
Totktonada, line 241: It seems, `self` is unused and can be dropped (`luacheck` likely complains here).

src/box/lua/config/instance_config.lua
Totktonada: As I see, it is placed into the replicaset options now without actual validation that it is not set in the instance scope. It is OK for now. Thanks!

src/box/lua/config/applier/sharding.lua
Totktonada: Now it writes a message to logs and continue to wait for ro/rw changes. It looks as an appropriate solution. Thanks!

I would consider issuing an alert in a future.

src/box/lua/config/applier/sharding.lua
Totktonada: It looks solved now. Thanks!

I'm not sure, whether we should check the fiber for being in the `dead` state. Let's think on it later.

src/box/lua/config/configdata.lua
```diff
@@ -62,9 +63,120 @@ function methods.names(self)
...
+            local replicaset_uuid
+            local replicaset_cfg = {}
+            for instance_name, _ in pairs(value.instances) do
+                local vars = {instance_name = instance_name}
+                local iconfig = cluster_config:instantiate(self._cconfig,
+                                                           instance_name)
+                iconfig = instance_config:apply_default(iconfig)
+                iconfig = iconfig_apply_vars(iconfig, vars)
+                if lock == nil then
+                    lock = instance_config:get(iconfig, 'sharding.zone')
```
Totktonada, line 130: Typo? It seems that `sharding.zone` should be `sharding.lock`.

src/box/lua/config/configdata.lua
```diff
@@ -119,6 +214,19 @@ local function new(iconfig, cconfig, instance_name)
...
+    local instance_uuid = instance_config:get(iconfig_def,
+        'database.instance_uuid')
+    if instance_config:get(iconfig_def, 'sharding') ~= nil then
+        if replicaset_uuid == nil then
+            replicaset_uuid = uuid_from_name(found.replicaset_name)
+        end
+        if instance_uuid == nil then
+            instance_uuid = uuid_from_name(instance_name)
+        end
+    end
```
Totktonada, line 228: I think that we shouldn't assign UUIDs depending on a presence of the `sharding` options. One may want to start a bunch of instances and later assign them sharding roles.

src/box/lua/config/configdata.lua
```diff
@@ -132,10 +240,21 @@ local function new(iconfig, cconfig, instance_name)
...
+        local peer_instance_uuid = instance_config:get(peer_iconfig,
+                                                       'database.instance_uuid')
+        if peer_instance_uuid == nil then
+            peer_instance_uuid = uuid_from_name(peer_name)
+        end
         peers[peer_name] = {
             iconfig = peer_iconfig,
             iconfig_def = peer_iconfig_def,
+            replicaset_uuid = peer_replicaset_uuid,
+            instance_uuid = peer_instance_uuid
```
Totktonada, line 257: Are these values used anywhere?

src/box/lua/config/configdata.lua
```diff
@@ -62,9 +63,120 @@ function methods.names(self)
...
+    return sha:sub(1,8)..'-'..sha:sub(9,12)..'-'..sha:sub(13,16)..'-'..
+           '00'..sha:sub(17,18)..'-'..sha:sub(19,30)
+end
+
+local function instance_sharding(iconfig, instance_name)
+    local roles = instance_config:get(iconfig, 'sharding.roles')
+    if roles == nil then
+        return nil
+    end
+    assert(type(roles) == 'table' and #roles > 0)
```
Totktonada, line 82: Nit: I wouldn't forbid the empty list of roles. I propose to interpret it similarly to `roles == nil`.

test/config-luatest/vshard_test.lua
```diff
@@ -0,0 +1,365 @@
...
+    g.server_2:wait_until_ready()
+    g.server_3:wait_until_ready()
+    g.server_4:wait_until_ready()
+    g.server_5:wait_until_ready()
+
+    -- Check that cluster was created.
+    local info = g.server_1:eval('return box.info')
+    t.assert_equals(info.name, 'instance-001')
+    t.assert_equals(info.replicaset.name, 'replicaset-001')
+    t.assert_equals(info.cluster.name, 'group-001')
```
Totktonada, line 92: It'll fail after rebase, see 07e6060bcfde9d379342d50acf9ea40722437ee4.

src/box/lua/config/instance_config.lua
ImeevMA: Thanks, dropped.

src/box/lua/config/configdata.lua
ImeevMA: Done.

src/box/lua/config/configdata.lua
ImeevMA: No, dropped.

src/box/lua/config/configdata.lua
ImeevMA: Fixed.

src/box/lua/config/instance_config.lua
```diff
@@ -1518,6 +1518,186 @@ return schema.new('instance_config', schema.record({
...
+                            'scope')
+                end
+            end,
+        }),
+        -- TODO: Add validate.
+        roles = schema.set({
+            'router',
+            'storage',
+        }),
+        -- Global vshard options.
```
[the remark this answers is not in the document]
ImeevMA: For now I added auto-adding 'guest' to formed URI, if there were no username.

test/config-luatest/vshard_test.lua
ImeevMA: Thanks, fixed.

src/box/lua/config/instance_config.lua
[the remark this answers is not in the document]
Totktonada: Ok for now. Let's return here back later.

changelogs/unreleased/gh-8861-vshard-options.md
Totktonada: Let's elaborate it later.

Totktonada (approved):

test/config-luatest/vshard_test.lua
```diff
@@ -0,0 +1,360 @@
...
+        box.schema.func.create('get')
+        box.schema.role.grant('public', 'execute', 'function', 'get')
+        local format = {{'id', 'unsigned'}, {'bucket_id', 'unsigned'}}
+        a = box.schema.space.create('a', {format = format})
+        a:create_index('id', {parts = {'id'}})
+        a:create_index('bucket_id', {parts = {'bucket_id'}, unique = false})
+    ]]
+    g.server_1:eval(exec)
+    g.server_3:eval(exec)
+    fiber.sleep(1)
```
ylobankov, line 268: Looks like a potentially flaky test. Why 1 sec, but not 2, 3, 4, etc. It seems we should wrap below asserts into `retrying` helper.

test/config-luatest/vshard_test.lua
```diff
@@ -0,0 +1,360 @@
...
+    t.assert_equals(g.server_4:eval([[return box.space.a:select()]]), {})
+
+    exec = [[
+        vshard.router.bootstrap()
+        vshard.router.call(1, 'write', 'put', {{id = 1, bucket_id = 1}})
+        vshard.router.call(800, 'write', 'put', {{id = 800, bucket_id = 800}})
+    ]]
+    g.server_5:eval(exec)
+    -- Wait for sharding execution.
+    fiber.sleep(1)
```
ylobankov, line 279: ditto

test/config-luatest/vshard_test.lua
```diff
@@ -0,0 +1,360 @@
...
+    res = g.server_2:eval([[return box.space.a:select()]])
+    t.assert_equals(res, {{800, 800}})
+    res = g.server_4:eval([[return box.space.a:select()]])
+    t.assert_equals(res, {{1, 1}})
+
+    -- Check that vshard cfg is reformed when master is changed.
+    g.server_1:eval([[box.cfg{read_only = true}]])
+    fiber.sleep(1)
```
ylobankov, line 287: ditto

test/config-luatest/vshard_test.lua
```diff
@@ -0,0 +1,360 @@
...
+            weight = 1,
+        },
+    }
+    res = g.server_1:eval(exec)
+    t.assert_equals(res.sharding, exp)
+    res = g.server_2:eval(exec)
+    t.assert_equals(res.sharding, exp)
+
+    g.server_2:eval([[box.cfg{read_only = false}]])
+    fiber.sleep(1)
```
ylobankov, line 325: ditto

test/config-luatest/vshard_test.lua
Totktonada: I've updated the patch following your suggestion. Thanks!

test/config-luatest/vshard_test.lua
Totktonada: Updated with `t.helpers.retrying()`.

test/config-luatest/vshard_test.lua
Totktonada: Updated with `t.helpers.retrying()`.

ylobankov (approved):
Tests LGTM

PR #9003, opened 2023-08-18
test: fix fiber stack overflow test not overflowing

Gumix:
`test/unit/guard.cc` calls `stack_break_f()` recursively until the stack overflows and a signal is fired, however it relies on undefined behavior when compares pointers to local variables. Fixed by comparing `__builtin_frame_address()` instead.

One of the examples of this UB is when ASAN allocates local variables on fake stacks, in that case the test completes without the stack overflow.

Also this patch disables ASAN for `stack_break_f()` to keep the array on the fiber stack (see the corresponding comment) and marks it as volatile to avoid optimizing it away by the compiler.

Closes tarantool/tarantool-qa#323

### DIFF
--- a/test/unit/guard.cc
+++ b/test/unit/guard.cc
@@ -11,20 +11,28 @@ sigsegf_handler(int signo)
 	exit(0);
 }
 
-static int __attribute__((noinline))
-stack_break_f(char *ptr)
+/*
+ * ASAN is disabled for this function, because for the stack-use-after-return
+ * detection it could allocate the `block' on a fake stack, rather than placing
+ * it on a fiber stack. In that case, a lot more recursive calls will be
+ * required to overflow the stack.
+ */
+static NOINLINE NO_SANITIZE_ADDRESS int
+stack_break_f(char *frame_zero)
 {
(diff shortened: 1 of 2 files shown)


locker (approved):

test/unit/guard.cc
```diff
@@ -11,20 +11,29 @@ sigsegf_handler(int signo)
...
-static int __attribute__((noinline))
-stack_break_f(char *ptr)
+/*
+ * ASAN is disabled for this function, because for the stack-use-after-return
+ * detection it could allocate the `block' on a fake stack, rather than placing
+ * it on a fiber stack. In that case, a lot more recursive calls will be
+ * required to overflow the stack.
+ */
+static NOINLINE int
+__attribute__((no_sanitize("address")))
```
locker, line 21: `default_gcc_centos_7` workflow fails because of this. I guess we should use this attribute only if ASAN is enabled. It's worth adding a new macro for this to `trivia/util.h`.

Gumix: Fixed.

locker:

PR #9005, opened 2023-08-21
config: create parent directories in mkdir applier

Totktonada:
Before this patch an attempt to set, say, `wal.dir` option to a non-existent directory `foo` succeeds, while the same with `foo/bar` directory fails.

Part of #8862

### DIFF
--- a/test/config-luatest/mkdir_test.lua
+++ b/test/config-luatest/mkdir_test.lua
@@ -1,0 +1,24 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+g.test_nested_dirs = function(g)
+    local verify = function()
+        local fio = require('fio')
+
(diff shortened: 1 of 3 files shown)


andreyaksenov (approved):

ImeevMA (approved):

ylobankov (approved):

test/config-luatest/mkdir_test.lua
```diff
@@ -0,0 +1,24 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+g.test_startup_success = function(g)
```
Totktonada, line 6: The test case name is accidentally copy-pasted from the `app_test.lua`. Let's adjust.

Totktonada: Renamed to `test_nested_dirs`.

PR #9011, opened 2023-08-22
 config: introduce example for sharding

ImeevMA:
Follow-up https://github.com/tarantool/tarantool/issues/9007

### DIFF
--- a/doc/examples/config/sharding.yaml
+++ b/doc/examples/config/sharding.yaml
@@ -1,0 +1,76 @@
+# See the vshard documentation for proper usage of vshard. The basic scenario
+# with this configuration is described below.
+#
+# Before using sharding, you need to create a spaces that will contain data and
+# define functions to work with these spaces on the storage masters, for
+# example:
+#
+# format = {{'id', 'unsigned'}, {'bucket_id', 'unsigned'}}
+# s = box.schema.space.create('test', {format = format})
+# s:create_index('id', {parts = {'id'}})
+# s:create_index('bucket_id', {parts = {'bucket_id'}, unique = false})
+
+# function get(id) return box.space.test:get(id) end
+# box.schema.func.create('get')
+# box.schema.role.grant('public', 'execute', 'function', 'get')
+
+# function put(id, b_id) box.space.test:insert({id, b_id}) return true end
+# box.schema.func.create('put')
+# box.schema.role.grant('public', 'execute', 'function', 'put')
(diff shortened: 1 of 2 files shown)


doc/examples/config/sharding.yaml
```diff
@@ -0,0 +1,39 @@
+credentials:
+  users:
+    guest:
+      roles: [super]
+
+iproto:
+  listen: 'unix/:./{{ instance_name }}.iproto'
+  advertise:
+    sharding: 'guest@'
```
Totktonada, line 9: I think that we should leave a comment what does it actually mean.

Currently, it is not just advertise URI, but also an instruction for all master storage nodes to grant necessary privileges for this user. This is important details.

Totktonada: BTW, I would add a separate user for this purpose. And it would be nice to set a password for it in the credentials section.

What doubts me: will not credentials applier revoke permissions granted by vshard at `config:reload()`?

doc/examples/config/sharding.yaml
```diff
@@ -0,0 +1,39 @@
...
+groups:
+  storages:
+    sharding:
+      roles: [storage]
+    replicasets:
+      storages-001:
+        instances:
+          storage-001:
+            database:
+              mode: rw
```
Totktonada, line 23: Can we use `failover.mode: manual` in this example? I think this is better mode, because it gives more safety checks -- allows only single master. I think that the `off` mode with expicit ro/rw flags should be used if multi-master is needed.

doc/examples/config/sharding.yaml
```diff
@@ -0,0 +1,39 @@
...
+  routers:
+    replicasets:
+      routers-001:
+        sharding:
+          roles: [router]
+        instances:
+          router-001:
+            database:
+              mode: rw
```
Totktonada, line 39: It is not necessary for an instance that is the only one in its replicaset. It'll be rw by default.

andreyaksenov (approved):

doc/examples/config/sharding.yaml
ImeevMA: Added a new user. Also, grants will be revoked after reload().

Totktonada (approved):
This minimal example is surely helps to understand where to start with configuring sharding. Thanks!

ylobankov (approved):

PR #9012, opened 2023-08-22
sql: refactor update_view_references a bit

alyapunov:
The function update_view_references is called when an SQL view
is created or dropped. The goal of this function is to modify
(increment or decrement) view_ref_count member of spaces that
the view references.

There were a several issues that deserves to be refactored:
* By design in case of error it left the job partially done, so
  some space references were modified while some other - not.
  Although there was no bug since special steps were made in case
  of error, this pattern is inconvenient and should be avoided.
* In case of error the failing space name was returned via special
  argument which is not flexible and even requires allocation.
* Another argument - suppress_error - has actually never
  suppressed any error because the only case when an error could
  occur is creation of a view, which used suppress_error = false.
* Fail of that function was not actually covered with tests.

So the last commit:
* Makes the function to do all or nothing.
* Forces the function to set diag by itself in case of error.
* Removes suppress_error argument while adding several asserts.\
* Adds a small test that fulfills coverage.

On a way to success this patch also refactors SrcList structure for SQL.

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -1865,8 +1865,6 @@ update_view_references(struct Select *select, int update_value,
 {
 	assert(update_value == 1 || update_value == -1);
 	struct SrcList *list = sql_select_expand_from_tables(select);
-	if (list == NULL)
-		return -1;
 	int from_tables_count = sql_src_list_entry_count(list);
 	for (int i = 0; i < from_tables_count; ++i) {
 		const char *space_name = sql_src_list_entry_name(list, i);
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -1853,19 +1853,15 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
  *
  * @param select Tables from this select to be updated.
  * @param update_value +1 on view creation, -1 on drop.
- * @param suppress_error If true, silently skip nonexistent
- *                       spaces from 'FROM' clause.
- * @param[out] not_found_space Name of a disappeared space.
- * @retval 0 on success, -1 if suppress_error is false and space
- *         from 'FROM' clause doesn't exist.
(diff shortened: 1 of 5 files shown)


src/box/alter.cc
```diff
@@ -1830,56 +1830,83 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
...
+	 * Skip iteration.
+	 */
+	if (sql_select_constains_cte(select, space_name))
+		return 0;
+	struct space *space = space_by_name0(space_name);
+	if (space == NULL) {
+		if (suppress_error)
+			return 0;
+		diag_set(ClientError, ER_NO_SUCH_SPACE, space_name);
+		return -1;
```
locker, line 1860: According to the coveralls report, this never happens. Let's drop this code and inline this function.

src/box/alter.cc
```diff
@@ -1830,56 +1830,83 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
...
-update_view_references(struct Select *select, int update_value,
-		       bool suppress_error, const char **not_found_space)
+update_view_references(struct Select *select,
+		       int update_value, bool suppress_error)
 {
 	assert(update_value == 1 || update_value == -1);
 	struct SrcList *list = sql_select_expand_from_tables(select);
 	if (list == NULL)
-		return -1;
+		panic("Out of memory while parsing sql view request");
```
locker, line 1884: `sql_select_expand_from_tables` should never fail with OOM since it uses `xmalloc`. Let's update its comment and code where required and drop this check.

locker (changes requested):

src/box/alter.cc
alyapunov: OK, added two prior commits that refactors `sql_select_expand_from_tables` and SQL SrcList.

locker (approved):

PR #9029, opened 2023-08-24
config: introduce audit options

ImeevMA:
This patch introduces all audit options.

Part of #8861

NO_DOC=Was already described before.

### DIFF
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -1694,7 +1694,114 @@ return schema.new('instance_config', schema.record({
                 end
             end,
         }),
-    })
+    }),
+    audit_log = enterprise_edition(schema.record({
+        -- The same as the destination for the logger, audit logger destination
+        -- is handled separately in the box_cfg applier, so there are no
+        -- explicit box_cfg and box_cfg_nondynamic annotations.
+        --
+        -- The reason is that there is no direct-no-transform
+        -- mapping from, say, `audit_log.file` to `box_cfg.audit_log`.
+        -- The applier should add the `file:` prefix.
+        to = enterprise_edition(schema.enum({
+            'devnull',
+            'file',
+            'pipe',
+            'syslog',
+        }, {
+            default = 'devnull',
+        })),
+        file = enterprise_edition(schema.scalar({
+            type = 'string',
+            -- The mk_parent_dir annotation is not present here,
+            -- because otherwise the directory would be created
+            -- unconditionally. Instead, mkdir applier creates it
+            -- if audit_log.to is 'file'.
+            default = 'var/log/{{ instance_name }}/audit.log',
+        })),
+        pipe = enterprise_edition(schema.scalar({
+            type = 'string',
+            default = box.NULL,
+        })),
+        syslog = schema.record({
+            identity = enterprise_edition(schema.scalar({
+                type = 'string',
+                default = 'tarantool',
+            })),
+            facility = enterprise_edition(schema.scalar({
+                type = 'string',
+                default = 'local7',
+            })),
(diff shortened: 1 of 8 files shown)


andreyaksenov (approved):

src/box/lua/config/instance_config.lua
```diff
@@ -358,6 +360,17 @@ local function feedback_validate(data, w)
...
 
+local function audit_apply_default_if(_data, _w)
+    return has_audit
+end
+
+local function audit_validate(data, w)
+    if data == nil or has_audit then
+        return
+    end
+    w.error('Module audit is not available')
```
Totktonada, line 371: The module is part of Tarantool EE. I would raise the same error as for other Enterprise Edition options. Moreover, with the `enterprise_edition = true` annotation the options will be listed in `--help-env-list` as available only in the Enterprise Edition. It should reduce a possible confusion.

src/box/lua/config/instance_config.lua
```diff
@@ -1694,7 +1707,108 @@ return schema.new('instance_config', schema.record({
...
+            "none",
+            "all",
+            "audit",
+            "auth",
+            "priv",
+            "ddl",
+            "dml",
+            "data_operations",
+            "compatibility",
+        }),
```
Totktonada, line 1810: 1. Can we write a test that all the events and events group are covered?
2. I would like to have some explicit separation of groups and events.

Totktonada: 3. This option should have a validate in a future, so it would be good to add a TODO comment similar to `sharding.roles` one.

ochaplashkin (approved):
LGTM :rocket:

src/box/lua/config/instance_config.lua
ImeevMA: Thank you for the review! Done.

src/box/lua/config/instance_config.lua
ImeevMA: 1. Done.
2. Added a comment. Should I make different options?
3. Done.

src/box/lua/config/instance_config.lua
```diff
@@ -1694,7 +1694,98 @@ return schema.new('instance_config', schema.record({
                 end
             end,
         }),
-    })
+    }),
+    audit_log = enterprise_edition(schema.record({
+        to = enterprise_edition(schema.enum({
```
Totktonada, line 1699: Nit: I would comment `audit_log.to` regarding lack of `box_cfg` and `box_cfg_nondefault` annotations like it is done for `log.to`.

src/box/lua/config/instance_config.lua
```diff
@@ -1694,7 +1694,98 @@ return schema.new('instance_config', schema.record({
...
+            'file',
+            'pipe',
+            'syslog',
+        }, {
+            default = 'devnull',
+        })),
+        file = enterprise_edition(schema.scalar({
+            type = 'string',
+            default = 'var/log/{{ instance_name }}/audit.log',
+        })),
```
Totktonada, line 1710: It should be supported by the mkdir applier, a comment should be added here. See `log.file` for example.

Totktonada: We should actually create a directory in applier/mkdir.lua. I guess it'll be just copy-paste of a code block that serves `log.file`.

src/box/lua/config/instance_config.lua
Totktonada: 1. I meant that it should be good to check a consistency between available `box.cfg.audit_filter` options and instance_config's ones.
2. OK. No, that's is exactly what I asked.
3. OK.

Thanks!

Totktonada (approved):

PR #9032, opened 2023-08-24
Fix transaction boundaries in WAL and replication

### LINKED ISSUE #8746: Сannot recover from an anonymous replica after writing to a local space due to an on_replace trigger for a global space.
If at an anonymous replica due to an on_replace trigger on the global space write is made  to the local space, then recovering will be impossible. Run the tarantool like below:

```LUA
------------
-- Master
------------

-- Step 1
box.cfg{
    listen = 3301,
    replication = {3301, 3302},
}

-- Step 3   
box.schema.space.create('loc', {is_local = true})
box.space.loc:create_index('pk')
box.schema.space.create('glob')
box.space.glob:create_index('pk')

-- Step 5
box.space.glob:insert{1}


---------------------
-- Anonymous replica
---------------------

-- Step 2
box.cfg{
    listen = 3302,
    replication = {3301, 3302},
    replication_anon = true,
    read_only = true,
}

-- Step 4
box.space.glob:on_replace(function()
    box.space.loc:replace{42}
end)
```
Now try to recover from the anonymous replica. If you recover without `force_recovery`, you will get error:
```
2023-06-07 10:53:32.787 [8125] main/103/interactive I> recover from `./00000000000000000004.xlog'
2023-06-07 10:53:32.788 [8125] main/103/interactive box.cc:822 E> error at request: {type: 'INSERT', replica_id: 1, lsn: 6, space_id: 513, index_id: 0, tuple: [42]}
2023-06-07 10:53:32.788 [8125] main/103/interactive box.cc:765 E> XlogError: found a first global row in a transaction with LSN/TSN mismatch
2023-06-07 10:53:32.788 [8125] main/103/interactive F> can't initialize storage: found a first global row in a transaction with LSN/TSN mismatch
```
 If you recover with `force_recovery = true`, you will get error:
```
2023-06-07 12:32:19.031 [16600] main/103/interactive I> recover from `./00000000000000000004.xlog'
2023-06-07 12:32:19.031 [16600] main/103/interactive box.cc:822 E> error at request: {type: 'INSERT', replica_id: 1, lsn: 6, space_id: 513, index_id: 0, tuple: [42]}
2023-06-07 12:32:19.031 [16600] main/103/interactive recovery.cc:292 E> skipping row {1: 6}
2023-06-07 12:32:19.031 [16600] main/103/interactive box.cc:765 E> XlogError: found a first global row in a transaction with LSN/TSN mismatch
tarantool: ./src/box/recovery.cc:280: void recover_xlog(recovery*, xstream*, const vclock*): Assertion `row.replica_id != 0 || row.group_id == GROUP_LOCAL' failed.
```

### LINKED ISSUE #8958: Tarantool 2.11.0: ER_PROTOCOL: Transaction id must be equal to LSN of the first row in the transaction.
**Bug description**

I caught misterious bug in asynchronous replication Tarantool 2.11.0:
`ER_PROTOCOL: Transaction id must be equal to LSN of the first row in the transaction.`

* OS: MacOS M1, Macos M1 (docker alpine)
* Architecture: arm, amd64?

```
Tarantool 2.11.0-0-g247a9a4
Target: Darwin-arm64-Release
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/opt/homebrew/Cellar/tarantool/2.11.0_1 -DENABLE_BACKTRACE=TRUE
Compiler: /usr/bin/clang /usr/bin/clang++
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -fopenmp -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tmp/tarantool-20230615-7295-5x7z7g/tarantool-2.11.0=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -Wno-cast-function-type
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -fopenmp -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tmp/tarantool-20230615-7295-5x7z7g/tarantool-2.11.0=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type
```

**Steps to reproduce**

It is quite hard to ease reproducer. Complete reproducer with steps is described in this repo/branch:
https://github.com/ochaton/hunting/tree/mmm-er-protocol

In general we setup 3 nodes in full-mesh master-master replication and create 2 spaces `bus` (ordinary space) and `inbox` (local space). We setup `bus:on_replace` trigger which filters tuples inside applier and inserts some tuples into local `inbox` space.

We have background fiber which atomically removes tuples from local space `inbox` and inserts tuples into ordinary `bus`.

**Actual behavior**

After sending single message via this mechanism 3rd node (not participating in logical communication) stops replication from replicaset. But other 2 nodes continue playing with each other.

Bug is reproduced when we mix local spaces with ordinary spaces in full-mesh master-master replication. Bug reproduced only with 3 instances or greater.

**Expected behavior**

Replication should continue running, because we do not violate any constraint in asynchronous replication

@sergepetrenko @sergos Please take a look

sergepetrenko:
commit 04aef9a9e0b9e6e3550e9413a0022cf3956f48d6
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Aug 24 11:28:39 2023 +0300

    box: fix force recovery for transactions with local rows
    
    Force recovery first tries to collect all rows of a transaction into a
    single list, and only then applies those rows.
    
    The problem was that it collected rows based on the row replica_id. For
    local rows replica_id is set to 0, but actually such rows can be part
    of a transaction coming from any instance.
    
    Fix recovery of such rows
    
    Follow-up #8746
    Follow-up #7932
    
    NO_DOC=bugfix
    NO_CHANGELOG=the broken behaviour couldn't be seen due to bug #8746

commit ce9444bfee4ccc08c6629d064a9a28393cccfaf5
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Wed Aug 23 14:10:59 2023 +0300

    box: get rid of dummy NOPs after transactions ending with local rows
    
    In order to preserve transaction boundaries over replication, Tarantool
    writes a global NOP row after the last transaction row, if this row
    happens to be local. This is done to make sure that the is_commit flag,
    which is set only in the last transaction row, reaches the replica. This
    wouldn't happen if the last row was local.
    
    This workaround works fine for transactions completely authored by one
    instance: when both global and local rows come from operations of a
    single master.
    
    However, it's possible to append local rows to a remote master's
    transaction on a replica. For example, one can use on_replace triggers
    to write to replica's local space on each new transaction coming from
    master.
    
    In this case essentially a global NOP entry is added at the end of a
    remote master's transaction. This leads to several problems.
    
    First of all, this bumps replica's LSN, which is counter-intuitive,
    given that the replica might even be read-only. Besides, in a star
    topology this leads to master being unable to connect to the replica
    later on due to their vclocks becoming incompatible.
    
    Secondly, even if replication channel between master and replica is
    bidirectional, it creates a new row which should be replicated from
    replica to master, but at the same time is the last row of the master's
    transaction. Once master receives this row, it breaks its connection to
    replica due to transaction boundary violation (the last row of the
    transaction is received without its beginning).
    
    Adding a NOP row became extraneous since the previous commit, which made
    relay find transaction boundaries by itself.
    
    Closes #8958
    
    NO_DOC=bugfix

commit 4f16ab8f080ebe827f6eabbff37d026403e47203
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Sep 7 17:47:44 2023 +0300

    relay: send rows transactionally
    
    Some time ago we started writing transaction boundaries to WAL and
    respecting them in the replication stream: replicas wait for a full
    transaction receipt before applying it.
    
    However, during all these changes relay remained transaction-agnostic:
    it simply read single rows from WAL and sent them over to the receiver.
    
    This lead to a handful of ugly crutches: for example, tsn is not always
    equal to the lsn of the first global row of the transaction: if the
    first low is local, tsn is deduced from the first global row of the
    transaction.
    
    Also a dummy NOP was appended to the end of a transaction ending by a
    local row, so that is_commit flag wasn't lost by the replication.
    
    Let's make relay read a full transaction, filter out all the unnecessary
    rows, set the transaction boundaries accordingly and then send the
    transaction at once.
    
    Since in relay a single fiber sends data to the remote peer, there is no
    chance for a heartbeat to get in between rows of a single transaction:
    they're all sent at once. Hence the deletion of a corresponding guard
    `relay->is_sending_tx`.
    
    Prerequisite #8958
    
    NO_DOC=internal change
    NO_CHANGELOG=internal change
    NO_TEST=covered by existing tests

commit 5756f956bcda75b561ca5897d0c1f65a398527c5
Author: Serge Petrenko <sergepetrenko@tarantool.org>
Date:   Thu Jul 27 14:36:56 2023 +0300

    wal: fix transaction boundaries for replicated transactions
    
    Transaction boundaries were not updated correctly for transactions in
    which local space writes were made from a replication trigger. Existing
    transaction boundaries and row flags from the master were written as is
    on the replica. Actually, the replica should recalculate transaction
    boundaries and even WAIT_SYNC/WAIT_ACK flags.
    
    Transaction boundaries should be recalculated when a replica appends a
    local write at the end of the master's transaction, and
    WAIT_SYNC/WAIT_ACK should be overwritten when nopifying synchronous
    transactions coming from an old term.
    
    The latter fix has uncovered the bug in skipping outdated synchronous
    transactions: if one replica replaces a transaction from an old term
    with NOPs and then passes that transaction to the other replica, the
    other replica raises a split brain error. It believes the NOPs are an
    async transaction form an old term. This worked before the fix, because
    the rows were written with the original WAIT_ACK = true bit. Now this
    is fixed properly: we allow fully NOP async tranasctions from the old
    term.
    
    Closes #8746
    
    NO_DOC=bugfix
    NO_CHANGELOG=covered by the next commit

Closes #8746
Closes #8958

### DIFF
--- a/changelogs/unreleased/gh-8958-remove-transaction-nop.md
+++ b/changelogs/unreleased/gh-8958-remove-transaction-nop.md
@@ -1,0 +1,6 @@
+## bugfix/replication
+
+* Fixed replicas writing corrupted xlogs when appending data to a local space
+  from an `on_replace` or `before_replace` trigger on a global replicated space.
+  Such xlogs were unrecoverable and caused other nodes to break replication with
+  the replica (gh-8746, gh-8958).
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1536,15 +1536,21 @@ applier_synchro_filter_tx(struct stailq *rows)
 	 */
 	struct xrow_header *last_row =
 		&stailq_last_entry(rows, struct applier_tx_row, next)->row;
+	struct applier_tx_row *item;
 	if (!last_row->wait_sync) {
-		if (iproto_type_is_dml(last_row->type) &&
-		    txn_limbo.owner_id != REPLICA_ID_NIL) {
+		if (!iproto_type_is_dml(last_row->type) ||
+		    txn_limbo.owner_id == REPLICA_ID_NIL) {
+			return;
+		}
+		stailq_foreach_entry(item, rows, next) {
+			row = &item->row;
+			if (row->type == IPROTO_NOP)
+				continue;
 			tnt_raise(ClientError, ER_SPLIT_BRAIN,
 				  "got an async transaction from an old term");
 		}
 		return;
 	}
-	struct applier_tx_row *item;
 	stailq_foreach_entry(item, rows, next) {
 		row = &item->row;
 		row->type = IPROTO_NOP;
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -37,6 +37,7 @@
 #include "cbus.h"
 #include "errinj.h"
 #include "fiber.h"
+#include "memory.h"
 #include "say.h"
 
 #include "coio.h"
@@ -122,6 +123,14 @@ struct relay {
 	struct recovery *r;
 	/** Xstream argument to recovery */
 	struct xstream stream;
+	/** A region used to save rows when collecting transactions. */
+	struct lsregion lsregion;
+	/** A monotonically growing identifier for lsregion allocations. */
+	int64_t lsr_id;
+	/** The tsn of the currently read transaction. */
+	int64_t read_tsn;
+	/** A list of rows making up the currently read transaction. */
+	struct rlist current_tx;
 	/** Vclock to stop playing xlogs */
 	struct vclock stop_vclock;
 	/** Remote replica */
@@ -189,13 +198,6 @@ struct relay {
 	enum relay_state state;
 	/** Whether relay should speed up the next heartbeat dispatch. */
 	bool need_new_vclock_sync;
-	/**
-	 * Whether relay is in process of sending a tx to the remote peer. In
-	 * this case cannot interrupt the transaction stream with any heartbeats
-	 * or other requests.
-	 */
-	bool is_sending_tx;
-
 	struct {
 		/* Align to prevent false-sharing with tx thread */
 		alignas(CACHELINE_SIZE)
@@ -267,8 +269,10 @@ static void
 relay_send(struct relay *relay, struct xrow_header *packet);
 static void
 relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row);
+
+/** Process a single row from the WAL stream. */
 static void
-relay_send_row(struct xstream *stream, struct xrow_header *row);
+relay_process_row(struct xstream *stream, struct xrow_header *row);
 
 struct relay *
 relay_new(struct replica *replica)
@@ -343,10 +347,11 @@ relay_start(struct relay *relay, struct iostream *io, uint64_t sync,
 	relay->state = RELAY_FOLLOW;
 	relay->sent_raft_term = sent_raft_term;
 	relay->need_new_vclock_sync = false;
-	relay->is_sending_tx = false;
 	relay->last_row_time = ev_monotonic_now(loop());
 	relay->tx_seen_time = relay->last_row_time;
 	relay->last_heartbeat_time = relay->last_row_time;
+	/* Never send rows for REPLICA_ID_NIL to anyone */
+	relay->id_filter = 1 << REPLICA_ID_NIL;
 }
 
 void
@@ -377,6 +382,7 @@ relay_exit(struct relay *relay)
 	 */
 	recovery_delete(relay->r);
 	relay->r = NULL;
+	lsregion_destroy(&relay->lsregion);
 }
 
 static void
@@ -431,6 +437,17 @@ relay_set_cord_name(int fd)
 		snprintf(name, sizeof(name), "relay/<unknown>");
 	}
 	cord_set_name(name);
+}
+
+static void
+relay_cord_init(struct relay *relay)
+{
+	coio_enable();
+	relay_set_cord_name(relay->io->fd);
+	lsregion_create(&relay->lsregion, &runtime);
+	relay->lsr_id = 0;
+	relay->read_tsn = 0;
+	rlist_create(&relay->current_tx);
 }
 
 void
@@ -514,8 +531,7 @@ relay_final_join_f(va_list ap)
 	struct relay *relay = va_arg(ap, struct relay *);
 	auto guard = make_scoped_guard([=] { relay_exit(relay); });
 
-	coio_enable();
-	relay_set_cord_name(relay->io->fd);
+	relay_cord_init(relay);
 
 	/* Send all WALs until stop_vclock */
 	assert(relay->stream.write != NULL);
@@ -538,7 +554,8 @@ relay_final_join(struct replica *replica, struct iostream *io, uint64_t sync,
 	struct relay *relay = replica->relay;
 	assert(relay->state != RELAY_FOLLOW);
 
-	relay_start(relay, io, sync, relay_send_row, relay_yield, UINT64_MAX);
+	relay_start(relay, io, sync, relay_process_row, relay_yield,
+		    UINT64_MAX);
 	auto relay_guard = make_scoped_guard([=] {
 		relay_stop(relay);
 	});
@@ -783,8 +800,6 @@ relay_reader_f(va_list ap)
 static inline void
 relay_send_heartbeat(struct relay *relay)
 {
-	if (relay->is_sending_tx)
-		return;
 	struct xrow_header row;
 	try {
 		++relay->last_sent_ack.vclock_sync;
@@ -965,8 +980,7 @@ relay_subscribe_f(va_list ap)
 {
 	struct relay *relay = va_arg(ap, struct relay *);
 
-	coio_enable();
-	relay_set_cord_name(relay->io->fd);
+	relay_cord_init(relay);
 
 	cbus_endpoint_create(&relay->tx_endpoint,
 			     tt_sprintf("relay_tx_%p", relay),
@@ -1104,7 +1118,7 @@ relay_subscribe(struct replica *replica, struct iostream *io, uint64_t sync,
 
 	if (replica_version_id < version_id(2, 6, 0) || replica->anon)
 		sent_raft_term = UINT64_MAX;
-	relay_start(relay, io, sync, relay_send_row,
+	relay_start(relay, io, sync, relay_process_row,
 		    relay_yield_and_send_heartbeat, sent_raft_term);
 	replica_on_relay_follow(replica);
 	auto relay_guard = make_scoped_guard([=] {
@@ -1121,7 +1135,7 @@ relay_subscribe(struct replica *replica, struct iostream *io, uint64_t sync,
 	vclock_copy(&relay->tx.vclock, replica_clock);
 	relay->version_id = replica_version_id;
 
-	relay->id_filter = replica_id_filter;
+	relay->id_filter |= replica_id_filter;
 
 	int rc = cord_costart(&relay->cord, "subscribe",
 			      relay_subscribe_f, relay);
@@ -1210,39 +1224,14 @@ relay_push_raft(struct relay *relay, const struct raft_request *req)
 	relay_push_raft_msg(relay);
 }
 
-/** Send a single row to the client. */
-static void
-relay_send_row(struct xstream *stream, struct xrow_header *packet)
+/** Check if a row should be sent to a remote replica. */
+static bool
+relay_filter_row(struct relay *relay, struct xrow_header *packet)
 {
-	struct relay *relay = container_of(stream, struct relay, stream);
 	assert(cord() == &relay->cord);
 	assert(fiber()->f == relay_subscribe_f ||
 	       fiber()->f == relay_final_join_f);
 	bool is_subscribe = fiber()->f == relay_subscribe_f;
-	/* Do not send heartbeats during a final join. */
-	if (is_subscribe)
-		relay_send_heartbeat_on_timeout(relay);
-	if (packet->group_id == GROUP_LOCAL) {
-		/*
-		 * We do not relay replica-local rows to other
-		 * instances, since we started signing them with
-		 * a zero instance id. However, if replica-local
-		 * rows, signed with a non-zero id are present in
-		 * our WAL, we still need to relay them as NOPs in
-		 * order to correctly promote the vclock on the
-		 * replica.
-		 */
-		if (packet->replica_id == REPLICA_ID_NIL)
-			return;
-		packet->type = IPROTO_NOP;
-		packet->group_id = GROUP_DEFAULT;
-		packet->bodycnt = 0;
-	}
-	assert(iproto_type_is_dml(packet->type) ||
-	       iproto_type_is_synchro_request(packet->type));
-	/* Check if the rows from the instance are filtered. */
-	if ((1 << packet->replica_id & relay->id_filter) != 0)
-		return;
 	/*
 	 * We're feeding a WAL, thus responding to FINAL JOIN or SUBSCRIBE
 	 * request. If this is FINAL JOIN (i.e. relay->replica is NULL),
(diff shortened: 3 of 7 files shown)


test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua
```diff
@@ -0,0 +1,124 @@
...
+    cg.replica_set = replica_set:new{}
+end)
+
+g.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+local function prepare(cg)
+    cg.replica_set:start()
+    cg.master:exec(function()
```
Serpentian, line 17: The test is flaky. Orphan state:
```
[001] replication-luatest/gh_8746_replica_transactio>                 [ fail ]
[001] Test failed! Output from reject file /tmp/t/rejects/replication-luatest/gh_8746_replica_transaction_boundary.reject:
[001] Tarantool version is 3.0.0-alpha2-38-g5c6d83159
[001] TAP version 13
[001] 1..3
[001] # Started on Thu Aug 31 15:49:21 2023
[001] # Starting group: gh-8746-transaction-boundaries
[001] ok     1	gh-8746-transaction-boundaries.test_replica_recovery
[001] not ok 2	gh-8746-transaction-boundaries.test_replication_to_master
[001] #   Can't modify data on a read-only instance - it is an orphan
[001] #   stack traceback:
[001] #   	...on-luatest/gh_8746_replica_transaction_boundary_test.lua:17: in function 'prepare'
[001] #   	...on-luatest/gh_8746_replica_transaction_boundary_test.lua:77: in function 'gh-8746-transaction-boundaries.test_replication_to_master'
[001] #   	...
[001] #   artifacts:
[001] #   master -> /tmp/t/001_replication-luatest/artifacts/rs-7LOVjZBt2xAa/master-szUBIBOHrp_4
[001] #   replica -> /tmp/t/001_replication-luatest/artifacts/rs-7LOVjZBt2xAa/replica-MKQXDNdzjZy5
[001] ok     3	gh-8746-transaction-boundaries.test_replica_force_recovery
[001] # Ran 3 tests in 3.281 seconds, 2 succeeded, 1 errored
```

p7nov (approved):

test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua
sergepetrenko: Thanks for noticing!

Doesn't flake on my machine, so I can't check, but hope this helps:
```diff
diff --git a/test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua b/test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua
index d0a2f73dfb..4fac79bb51 100644
--- a/test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua
+++ b/test/replication-luatest/gh_8746_replica_transaction_boundary_test.lua
@@ -15,6 +15,7 @@ end)
 local function prepare(cg)
     cg.replica_set:start()
     cg.master:exec(function()
+        box.ctl.wait_rw()
         box.schema.space.create('test')
         box.space.test:create_index('pk')
         box.schema.space.create('loc', {is_local = true})
```

ochaplashkin (approved):
LGTM :rocket:

src/box/wal.c
```diff
@@ -971,30 +972,16 @@ wal_assign_lsn(struct vclock *vclock_diff, struct vclock *base,
 			 * only capable of writing to local and
 			 * temporary spaces.
 			 */
-			if ((*row)->group_id != GROUP_LOCAL)
+			if ((*row)->group_id != GROUP_LOCAL) {
 				(*row)->replica_id = instance_id;
+				last_glob_row = row;
```
Serpentian, line 977: This is the question of code readability, it's up to you, to fix it or not. I don't like, that we duplicate assigning `last_glob_row` in two mutually exclusive `if` code blocks. We can do it below, alongside with `first_glob_row`.

```diff
diff --git a/src/box/wal.c b/src/box/wal.c
index f28d6211e..dd65e34ed 100644
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -974,14 +974,12 @@ wal_assign_lsn(struct vclock *vclock_diff, struct vclock *base,
 			 */
 			if ((*row)->group_id != GROUP_LOCAL) {
 				(*row)->replica_id = instance_id;
-				last_glob_row = row;
 			}
 
 			(*row)->lsn = vclock_inc(vclock_diff, (*row)->replica_id) +
 				      vclock_get(base, (*row)->replica_id);
 		} else {
 			int64_t diff = (*row)->lsn - vclock_get(base, (*row)->replica_id);
-			last_glob_row = row;
 			if (diff <= vclock_get(vclock_diff,
 					       (*row)->replica_id)) {
 				int64_t confirmed_lsn =
@@ -999,16 +997,18 @@ wal_assign_lsn(struct vclock *vclock_diff, struct vclock *base,
 			/* Reset row flags taken from the remote instance. */
 			(*row)->flags = 0;
 		}
+
 		/*
 		 * Use lsn of the first global row as
 		 * transaction id.
 		 */
-		if ((*row)->group_id != GROUP_LOCAL && tsn == 0) {
-			tsn = (*row)->lsn;
-			/*
-			 * Remember the tail being processed.
-			 */
-			first_glob_row = row;
+		if ((*row)->group_id != GROUP_LOCAL) {
+			last_glob_row = row;
+			if (tsn == 0) {
+				tsn = (*row)->lsn;
+				/** Remember the tail being processed. */
+				first_glob_row = row;
+			}
 		}
 		(*row)->tsn = tsn == 0 ? (*start)->lsn : tsn;
 	}

```

</details>

Serpentian (approved):
Thank you for the patch. Looks great. Only one minor comment from me

src/box/applier.cc
```diff
@@ -1536,15 +1536,21 @@ applier_synchro_filter_tx(struct stailq *rows)
 	 */
```
Gerold103, line 1536: Can you apply this to the first commit?
```Diff
@@ -999,15 +999,16 @@ wal_assign_lsn(struct vclock *vclock_diff, struct vclock *base,
 		 * Use lsn of the first global row as
 		 * transaction id.
 		 */
-		if ((*row)->group_id != GROUP_LOCAL && tsn == 0) {
+		if (tsn == 0 && (*row)->group_id != GROUP_LOCAL) {
 			tsn = (*row)->lsn;
 			/*
 			 * Remember the tail being processed.
 			 */
 			first_glob_row = row;
 		}
-		(*row)->tsn = tsn == 0 ? (*start)->lsn : tsn;
 	}
+	if (tsn == 0)
+		tsn = (*start)->lsn;
 
 	/*
```
Now you set `row->tsn` 2 times for each row. The diff suggests to only find the target value in the first loop, fallback to `start->tsn` in a branch, and then set all in the second loop (which is already present below this diff). It should help to avoid 1 write for each row.

Also I swapped the condition parts - it is more likely to fail on `tsn == 0`, should help to avoid the second check most of the times.

changelogs/unreleased/gh-8958-remove-transaction-nop.md
```diff
@@ -0,0 +1,6 @@
+## bugfix/replication
```
Gerold103, line 1: Speaking of the second commit. I just had an idea now. Have you considered hacking the transaction flags inside the relay? So it would read the whole txn into memory, find the last global row, and send it with `is_commit`. So no need to store this flag on disk. We simply stop writing NOPs, and the relay will find the last global row itself. Would it work?

Gerold103 (commented):
Thanks for the patch!

src/box/applier.cc
sergepetrenko: > Now you set row->tsn 2 times for each row. The diff suggests to only find the target value in the first loop, fallback to start->tsn in a branch, and then set all in the second loop (which is already present below this diff). It should help to avoid 1 write for each row.

I don't understand. Why?

In the first loop (`for (struct xrow_header **row = start; row <= end; row++) {`) I set a valid tsn for all rows after the first glob row. And all the local rows before that get `start->tsn`.

In the second loop (`for (struct xrow_header **row = start; row < first_glob_row; row++)`) I only set tsns to the first couple of local rows, which came before first glob row. Not to all rows.

So if you hae a transaction [l, l, ... l, g, ...] where `l` stands for a local row and `g` stands for a global row, the rows up to the first `g` will receive a `(*start)->tsn`, while all the rows after the first `g` will receive a valid tsn. The second loop will rewrite tsns of the rows up to the first `g`,  but not of all the rows.

If I understood you correctly, you want to assign all tsns in the second loop. But then I'll always have to go through two loops. Even if tsn is known right in the first transaction row. And even if there are no global rows in the transaction.

With the current approach both cases above require a single loop (the first one), and the second one is skipped. Meaning there is only one tsn assignment per row. The only case when we write tsn twice happens only for the first local rows of a transaction, up to the first global row.

> Also I swapped the condition parts - it is more likely to fail on tsn == 0, should help to avoid the second check most of the times.

Applied:
```diff
diff --git a/src/box/wal.c b/src/box/wal.c
index ac30de209a..13ed7a0184 100644
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -999,7 +999,7 @@ wal_assign_lsn(struct vclock *vclock_diff, struct vclock *base,
                 * Use lsn of the first global row as
                 * transaction id.
                 */
-               if ((*row)->group_id != GROUP_LOCAL && tsn == 0) {
+               if (tsn == 0 && (*row)->group_id != GROUP_LOCAL) {
                        tsn = (*row)->lsn;
                        /*
                         * Remember the tail being processed.
```

changelogs/unreleased/gh-8958-remove-transaction-nop.md
sergepetrenko: Good idea, I didn't think of it, actually. Sounds much better than the current approach. Although there will be an added cost of finding the last global row after reading the transaction into memory.

It seems we'll have to read full transactions in relay for the sake of replication filters anyway. To preserve transaction boundaries in filtered streams. So I think it's worth choosing the same approach here. Will try.

src/box/wal.c
sergepetrenko: It seems I'll have to dump this commit and instead make relay transactional.

src/box/applier.cc
Gerold103: Sorry, I made a mistake, all is much simpler. I just didn't notice that the 2 loops have different end condition.

sergepetrenko:
To be sure that such relay rewriting won't hurt replication performance, I conducted some tests with [1mops_write](https://github.com/tarantool/tarantool/blob/master/perf/lua/1mops_write.lua) test.

I ran Tarantool 3.0.0-alpha3-24-g74f723bb26 (labeled master in csv) and the same version with this patch applied on top (labeled relay in csv). I also tried various `--fibers` and `--transaction` settings and collected the data for 2 and 3 nodes (1 and 2 replicas).

Here are the results in csv:

src/box/relay.cc
```diff
@@ -37,6 +37,7 @@
 #include "cbus.h"
```
Serpentian, line 37: Typo in the commit msg:

> if the first ~~low~~ row is local,

Serpentian (approved):
Thank you for your patch. It'll be much easier to integrate replication filters here. I have no objections

src/box/relay.cc
sergepetrenko: Thanks for noticing!
Fixed.

src/box/relay.cc
```diff
@@ -1255,9 +1238,104 @@ relay_send_row(struct xstream *stream, struct xrow_header *packet)
...
+
+/** Save a single transaction row for the future use. */
+static void
+relay_save_row(struct relay *relay, struct rlist *tx,
+	       struct xrow_header *packet)
+{
+	struct relay_row *tx_row = lsregion_alloc_object(&relay->lsregion,
+							 ++relay->lsr_id,
+							 struct relay_row);
+	if (tx_row == NULL) {
```
Gerold103, line 1310: Maybe use `xlsregion_alloc()`? Same in the other place.

src/box/relay.cc
```diff
@@ -1266,27 +1344,43 @@ relay_send_row(struct xstream *stream, struct xrow_header *packet)
...
+	rlist_create(tx);
+	lsregion_gc(&relay->lsregion, relay->lsr_id);
+}
+
+static void
+relay_process_row(struct xstream *stream, struct xrow_header *packet)
+{
+	struct relay *relay = container_of(stream, struct relay, stream);
+	static __thread int64_t read_tsn = 0;
+	static __thread struct rlist current_tx;
```
Gerold103, line 1359: I would suggest to avoid global variables where possible, and move them into `struct relay`.

Gerold103 (approved):
Thanks for the fixes!

sergepetrenko:
Cherry-picked to 2.11: c8594fbd80 .. 4643a26a2d

sergepetrenko:
Cherry-picked to 2.10: 14e1d055a1 .. 4753208cf7

Didn't cherry-pick the last commit, it fixes a system nonexistent in 2.10
Cherry-picked an additional auxiliary commit (14e1d055a1)

PR #9033, opened 2023-08-24
test: allow to omit message in ok/is/isnt unit test helpers

locker:
The main goal of this PR is to make the message argument of ok/is/isnt unit test helpers optional because it's really annoying to specify it after each check.

Along the way, it also does some cleanup:
 - Removes duplicated code from `unit.h`.
 - Adds missing trailing semicolons after ok/is/isnt.

### DIFF
--- a/test/unit/unit.h
+++ b/test/unit/unit.h
@@ -70,7 +70,9 @@ extern "C" {
 */
 
 /* private function, use ok(...) instead */
-int _ok(int condition, const char *fmt, ...);
+void
+_ok(int condition, const char *expr, const char *file, int line,
+    const char *fmt, ...);
 
 /* private function, use note(...) or diag(...) instead */
 void _space(FILE *stream);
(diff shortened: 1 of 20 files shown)


test/unit/unit.h
```diff
@@ -90,41 +90,42 @@ _plan(int count, bool tap);
...
-	if (!res) {					\
-		_space(stderr);			\
-		fprintf(stderr, "#   Failed test '");	\
-		fprintf(stderr, fmt, ##args);		\
-		fprintf(stderr, "'\n");			\
-		_space(stderr);			\
-		fprintf(stderr, "#   in %s at line %d\n", __FILE__, __LINE__); \
-	}						\
-}
+#define _ok0(condition) _ok1(condition, "line %d", __LINE__)
```
alyapunov, line 93: As I said, I would call `_ok1(condition, "%s", #condition)`, but actually for ok case it does not matter.

alyapunov (approved):

test/unit/unit.h
alyapunov: After a short discussion I was convinced that line number is always more informative.

alyapunov (approved):

nshy (approved):
LGTM

ochaplashkin (approved):
LGTM :rocket:

locker:

PR #9034, opened 2023-08-24
test: fix flaky gh_7515_sync_node_sees_leader_hang

Serpentian:
The test checks, whether the instance notices leader hang during sync. For that it stops server2 and generates some data on the master (server1). The problem is, when connection to server3 flakes, server1 resigns and becomes read only due to fencing enabled.

Let's disable fencing for this test, as we don't want automatic leader resigning here.

Closes tarantool/tarantool-qa#325

NO_CHANGELOG=test
NO_DOC=test


sergepetrenko (approved):
Thanks for the fix!

LGTM.

ylobankov (approved):

sergepetrenko:
Cherry-picked to 2.11: d74dd308918a43b795f09ded28e228e72a2fe5d6

PR #9042, opened 2023-08-25
Add @Totktonada as codeowner for config

Totktonada:
I want to receive review request notifications about the changes and let people know who is responsible to accept them.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,3 +5,7 @@
 /test/       @tarantool/devx
 /.test.mk    @tarantool/devx
 /test-run/   @tarantool/devx
+# Alexander Turenko wants to track all the activities around the
+# declarative configuration.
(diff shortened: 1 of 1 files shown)


igormunkin (approved):
I personally have no objections (and even glad we're moving to splitting the sources by "maintainers"). However, it looks better to introduce `@config` group to be notified for the changes, and hence manage the working group there. Anyway, I leave the final decision on everything written above to @kyukhin.

kyukhin (approved):

Totktonada:
@igormunkin @tarantool/config would mean all the developers that're involved into config's development. So, it should be @tarantool/config-maintainers with just me in the group :)

I don't mind, but I see no much difference from assigning me directly. I'll do, if Kirill decides that it is a more proper way.

ylobankov (approved):

Issue #7273, opened 2022-06-16
datetime: unclear error message on new()


olegrok:
```lua
Tarantool 2.11.0-entrypoint-124-g17287ed88
type 'help' for interactive help
tarantool> dt = require('datetime')
---
...

tarantool> dt.new({timestamp = 1LL})
---
- error: 'builtin/datetime.lua:587: bad argument #1 to ''math_modf'' (number expected,
    got cdata)'
...
```

PR #9062, opened 2023-08-29
lua: provide whether ASAN build in tarantool.build.asan

nshy:
We already use this info in one of the test and going to use it more.

Let's also move initializing tarantool package before initializing other packages implemented in Lua. This way we will have access to build info in those packages. In particularly build.asan flag is going to be used in buffer.lua in scope of #7327.

Part of #7327

### DIFF
--- a/changelogs/unreleased/add-asan-flag-to-tarantool-build.md
+++ b/changelogs/unreleased/add-asan-flag-to-tarantool-build.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Added a new flag `tarantool.build.asan` that shows whether build
+  flag `ENABLE_ASAN` is set.
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -795,15 +795,7 @@ luaopen_tarantool(lua_State *L)
 	lua_pushstring(L, tarantool_version());
 	lua_setfield(L, LUA_GLOBALSINDEX, "_TARANTOOL");
 
-	/*
-	 * Get tarantool module.
-	 *
-	 * src/lua/init.lua is already registered as tarantool
-	 * module, so we `require` it here, not create.
-	 */
-	lua_getfield(L, LUA_GLOBALSINDEX, "require");
-	lua_pushstring(L, "tarantool");
-	lua_call(L, 1, 1);
+	luaT_newmodule(L, "tarantool", NULL);
 
 	/* package */
 	lua_pushstring(L, tarantool_package());
@@ -998,6 +990,7 @@ tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 #if defined(ENABLE_BACKTRACE)
 	backtrace_lua_init();
 #endif /* defined(ENABLE_BACKTRACE) */
+	luaopen_tarantool(L);
 	for (const char **s = lua_modules; *s; s += 2) {
 		const char *modname = *s;
 		const char *modsrc = *(s + 1);
@@ -1018,9 +1011,6 @@ tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 		builtin_modcache_put(modname, modsrc);
 	}
 	lua_pop(L, 1); /* _PRELOAD */
-
-	luaopen_tarantool(L);
-
 #ifdef NDEBUG
 	/* Unload strict after boot in release mode */
 	if (luaL_dostring(L, "require('strict').off()") != 0)
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -851,6 +851,15 @@ luaopen_tarantool(lua_State *L)
 #endif
 	lua_settable(L, -3);
 
+	/* build.asan */
+	lua_pushstring(L, "asan");
+#ifdef ENABLE_ASAN
+	lua_pushboolean(L, true);
+#else
+	lua_pushboolean(L, false);
+#endif
+	lua_settable(L, -3);
+
 	lua_settable(L, -3);    /* box.info.build */
 
 	/* debug */
--- a/src/lua/init.lua
+++ b/src/lua/init.lua
@@ -1,8 +1,8 @@
-
 -- init.lua -- internal file
 
 local ffi = require('ffi')
 local loaders = require('internal.loaders')
+local tarantool = require('tarantool')
 
 ffi.cdef[[
 struct method_info;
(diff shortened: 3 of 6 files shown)


locker (approved):

src/lua/init.c
```diff
@@ -851,6 +851,13 @@ luaopen_tarantool(lua_State *L)
 #endif
 	lua_settable(L, -3);
 
+#ifdef ENABLE_ASAN
+	/* build.asan */
+	lua_pushstring(L, "asan");
+	lua_pushboolean(L, true);
+	lua_settable(L, -3);
+#endif
```
Totktonada, line 859: Is it a good idea to push explicit `false` if ASAN is disabled?

This way we can distinguish 'unknown' situation (on an old tarantol version) from known-to-disabled on the newer version. Also, it highlight that this value exists in the table for a developer.

Totktonada (approved):

src/lua/init.c
```diff
@@ -851,6 +851,13 @@ luaopen_tarantool(lua_State *L)
 #endif
 	lua_settable(L, -3);
 
+#ifdef ENABLE_ASAN
```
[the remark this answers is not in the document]
nshy: Done.

nshy:
I've also moved `luaopen_tarantool` before initialization of Lua implemented packages.

src/lua/utils.c
```diff
@@ -422,7 +422,8 @@ luaT_newmodule(struct lua_State *L, const char *modname,
 	lua_newtable(L);
 
 	/* Fill the module table with functions. */
-	luaL_setfuncs(L, funcs, 0);
+	if (funcs != NULL)
+		luaL_setfuncs(L, funcs, 0);
```
locker, line 426: AFAIU you should use `luaT_setmodule` instead of patching `luaT_newmodule`.

src/lua/init.lua
```diff
@@ -211,12 +211,18 @@ local mt = {
...
     pid = pid,
     _internal = {
         strip_cwd_from_path = strip_cwd_from_path,
         module_name_from_filename = module_name_from_filename,
         run_preload = run_preload,
     },
-}, mt)
+}
+-- tarantool module is already registered by src/lua/init.c
+local tarantool = require('tarantool')
```
locker, line 224: AFAIU, by our convention, the internal module should be named `tarantool.lib`.

src/lua/init.c
```diff
@@ -851,6 +843,15 @@ luaopen_tarantool(lua_State *L)
...
 
+	/* build.asan */
+	lua_pushstring(L, "asan");
+#ifdef ENABLE_ASAN
+	lua_pushboolean(L, true);
+#else
+	lua_pushboolean(L, false);
+#endif
+	lua_settable(L, -3);
+
```
locker, line 854: Since this change is documented, it should have a changelog entry.

src/lua/init.c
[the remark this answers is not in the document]
Totktonada: As Vladimir highlighted, there is no changelog. It should be present, because the change is user-visible.

src/lua/init.c
```diff
@@ -795,15 +795,7 @@ luaopen_tarantool(lua_State *L)
...
-	/*
-	 * Get tarantool module.
-	 *
-	 * src/lua/init.lua is already registered as tarantool
-	 * module, so we `require` it here, not create.
-	 */
-	lua_getfield(L, LUA_GLOBALSINDEX, "require");
-	lua_pushstring(L, "tarantool");
-	lua_call(L, 1, 1);
+	luaT_newmodule(L, "tarantool", NULL);
```
Totktonada, line 798: I have reviewed a different patch on the first iteration. It had no modules loading tweaks. I propose to do it in a separate patch.

> In particularly build.asan flag is going to be used in buffer.lua in scope of #7273.
>
> Part of #7327

As I see, buffer.lua is loaded after init.lua, so I don't see why the order of loading should be changed.

There are other changes around the `tarantool` module and I also see no reason behind it. Why it works for `build.target`, but not for `build.asan`?

Maybe it worth to discuss the patch f2f.

BTW, the issue number looks wrong (7273 vs 7327).

src/lua/utils.c
Totktonada: We use `luaT_newmodule()` for a new module created from C. I think it is OK to use it here as well -- the case is pretty same.

However, it is different in that we don't want to register any function. This is the first such case, however I don't see anything wrong here. Just a module without functions.

src/lua/init.lua
Totktonada: THere are two ways to express a module that is partially implemented in C and partially implemented in Lua.

1. `foo.lib` (C) + `foo` (Lua). In this case `foo` exports needed `foo.lib` functions directly or using a wrapper function. Example: `socket`.
2. `foo` (C) + `foo` (Lua). In this case `foo` MUST return the same module table, but it can add it more functions. Example: `fio`.

Both ways looks OK for me.

We discussed it f2f with Nikolay. He prefers the second way, because the functions implemented in C are exposed directly, without any wrapper. This reason also seems OK to me.

src/lua/init.c
nshy: Done.

src/lua/init.lua
nshy: Moved `require('tarantool')` to the top of the file as disscussed f2f. Also used `tarantool_lua` instead of `extension`.

src/lua/init.c
nshy: Fixed wrong PR reference. Also as disscussed f2f split to two commits.

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* A new flag `tarantool.build.asan` is added showing whether build
+  flag `ENABLE_ASAN` is set or not.
```
p7nov, line 4:
```suggestion
* Added a  new flag `tarantool.build.asan` that shows whether build
  flag `ENABLE_ASAN` is set.
```

p7nov (approved):

locker (approved):

ylobankov (approved):

Totktonada (approved):

locker:
@nshy Please fix the changelog.

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
nshy: Fixed.

PR #9064, opened 2023-08-29
trigger: introduce tarantool.trigger.on_change triggers

### LINKED ISSUE #8664: Introduce `tarantool.trigger.on_change` trigger
Sometimes it is useful to understand if any trigger has been registered on the event. For example, when argument preparation is a heavy operation, it would be very beneficial to skip such a useless and resource-consuming action if event does not actually contain any trigger. The solution is `tarantool.trigger.on_change` - a handler accepts the only argument - `event_name`. If an error is thrown, the change is rolled back - it will allow to limit the number of triggers registered on the event.

drewdzzz:
New event named 'tarantool.trigger.on_change' is called when any event is modified (trigger.set or trigger.del). All the handlers are called with one argument - name of the modified event. Returned value of each handler is ignored. Handlers are fired after the event is changed (the event contains inserted trigger, if any, and does not contain deleted one, if any). All thrown errors are logged with error level and do not stop execution of triggers.

Closes #8664

### DIFF
--- a/test/app-luatest/trigger_module_test.lua
+++ b/test/app-luatest/trigger_module_test.lua
@@ -1,3 +1,4 @@
+local server = require('luatest.server')
 local t = require('luatest')
 
 local g = t.group()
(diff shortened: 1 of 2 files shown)


test/app-luatest/trigger_module_test.lua
```diff
@@ -1,7 +1,20 @@
+local server = require('luatest.server')
 local t = require('luatest')
 
 local g = t.group()
 
+g.after_each(function()
+    local trigger = require('trigger')
```
ochaplashkin, line 7: You are using tarantool which is used as a **test runner** (this may affect other tests during a competitive launch).

Maybe we'll switch to a separate server instance?

drewdzzz: Thank you, moved the tests to a separate server in the first commit.

alyapunov (approved):

alyapunov (approved):

PR #9069, opened 2023-08-30
wal: add stubs for wal_retention_period

Serpentian:
This PR adds a new configuration option wal_retention_period and
function stubs for it. It's needed to avoid rebootstrap on anonymous
replicas, as Tarantool doesn't save xlog for them.

  * wal_retention_period. Type: double. Default: 0. Unit: seconds.

  Description: When xlog file is closed it won't be immediately
  deleted by garbage collector if wal_retention_period is set to
  non-zero value. wal_retention_delay determines how many seconds
  after closing this file should be saved. During instance restart
  the timeout is set to 'wal_retention_period - time' passed since
  file modification.

The option is dynamic. Its value is stored and used in C code, so we
define configuration callbacks in EE: cfg_set_wal_retention_period.

Needed for https://github.com/tarantool/tarantool-ee/issues/513

NO_DOC=EE
NO_CHANGELOG=EE

### DIFF
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -178,7 +178,12 @@ endif()
 add_library(tuple STATIC ${tuple_sources})
 target_link_libraries(tuple json box_error core ${MSGPUCK_LIBRARIES} misc bit coll)
 
-add_library(xlog STATIC xlog.c)
+set(xlog_sources xlog.c)
+if(ENABLE_RETENTION_PERIOD)
+    list(APPEND xlog_sources ${RETENTION_PERIOD_SOURCES})
+endif()
+
+add_library(xlog STATIC ${xlog_sources})
 target_link_libraries(xlog core box_error crc32 ${ZSTD_LIBRARIES})
 
 set(box_sources
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -1654,6 +1654,29 @@ box_check_wal_max_size(int64_t wal_max_size)
 	return wal_max_size;
 }
 
+/** Validate that wal_retention_period is >= 0. */
+static double
+box_check_wal_retention_period()
+{
+	double value = cfg_getd("wal_retention_period");
+	if (value < 0) {
+		diag_set(ClientError, ER_CFG, "wal_retention_period",
+			 "the value must be >= 0");
+		return -1;
+	}
+	return value;
+}
+
+/** Validate wal_retention_period and raise error, if needed. */
+static double
+box_check_wal_retention_period_xc()
+{
+	double value = box_check_wal_retention_period();
+	if (value < 0)
+		diag_raise();
+	return value;
+}
+
 static ssize_t
 box_check_memory_quota(const char *quota_name)
 {
@@ -1912,6 +1935,8 @@ box_check_config(void)
 	if (box_check_wal_queue_max_size() < 0)
 		diag_raise();
 	if (box_check_wal_cleanup_delay() < 0)
+		diag_raise();
+	if (box_check_wal_retention_period() < 0)
 		diag_raise();
 	if (box_check_memory_quota("memtx_memory") < 0)
 		diag_raise();
@@ -3154,6 +3179,16 @@ box_set_wal_cleanup_delay(void)
 	if (box_is_anon())
 		delay = 0;
 	gc_set_wal_cleanup_delay(delay);
+	return 0;
+}
+
+int
+box_set_wal_retention_period(void)
+{
+	double delay = box_check_wal_retention_period();
+	if (delay < 0)
+		return -1;
+	wal_set_retention_period(delay);
 	return 0;
 }
 
@@ -5860,8 +5895,10 @@ box_storage_init(void)
 	int64_t wal_max_size = box_check_wal_max_size(
 		cfg_geti64("wal_max_size"));
 	enum wal_mode wal_mode = box_check_wal_mode(cfg_gets("wal_mode"));
+	double wal_retention_period = box_check_wal_retention_period_xc();
 	if (wal_init(wal_mode, cfg_gets("wal_dir"), wal_max_size,
-		     &INSTANCE_UUID, on_wal_garbage_collection,
+		     wal_retention_period, &INSTANCE_UUID,
+		     on_wal_garbage_collection,
 		     on_wal_checkpoint_threshold) != 0) {
 		diag_raise();
 	}
--- a/src/box/gc.c
+++ b/src/box/gc.c
@@ -226,6 +226,15 @@ gc_run_cleanup(void)
 		consumer = gc_tree_next(&gc.consumers, consumer);
 	}
 
+	/*
+	 * Acquire minimum vclock of a file, which is protected from garbage
+	 * collection by wal_retention_period option.
+	 */
+	struct vclock retention_vclock;
+	wal_get_retention_vclock(&retention_vclock);
+	if (vclock_is_set(&retention_vclock))
+		vclock_min(&min_vclock, &retention_vclock);
+
 	if (vclock_sum(&min_vclock) > vclock_sum(&gc.vclock)) {
 		vclock_copy(&gc.vclock, &min_vclock);
 		run_wal_gc = true;
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -77,6 +77,12 @@ local function ifdef_security(value)
     end
 end
 
+local function ifdef_wal_retention_period(value)
+    if private.cfg_set_wal_retention_period ~= nil then
+        return value
+    end
+end
+
 -- all available options
 local default_cfg = {
     listen              = nil,
@@ -151,6 +157,7 @@ local default_cfg = {
     wal_dir_rescan_delay= 2,
     wal_queue_max_size  = 16 * 1024 * 1024,
     wal_cleanup_delay   = 4 * 3600,
+    wal_retention_period = ifdef_wal_retention_period(0),
     wal_ext             = ifdef_wal_ext(nil),
     force_recovery      = false,
     replication         = nil,
@@ -349,6 +356,7 @@ local template_cfg = {
     wal_max_size        = 'number',
     wal_dir_rescan_delay= 'number',
     wal_cleanup_delay   = 'number',
+    wal_retention_period = ifdef_wal_retention_period('number'),
     wal_ext             = ifdef_wal_ext('table'),
     force_recovery      = 'boolean',
     replication         = 'string, number, table',
@@ -497,6 +505,7 @@ local dynamic_cfg = {
     -- do nothing, affects new replicas, which query this value on start
     wal_dir_rescan_delay    = nop,
     wal_cleanup_delay       = private.cfg_set_wal_cleanup_delay,
+    wal_retention_period    = private.cfg_set_wal_retention_period,
     custom_proc_title       = function()
         require('title').update(box.cfg.custom_proc_title)
     end,
@@ -693,6 +702,7 @@ local dynamic_cfg_skip_at_load = {
     disable_guest           = ifdef_security(true),
     secure_erasing          = ifdef_security(true),
     password_lifetime_days  = ifdef_security(true),
+    wal_retention_period    = ifdef_wal_retention_period(true),
 }
 
 -- Options that are not part of dynamic_cfg_modules and applied individually
--- a/src/box/retention_period.h
+++ b/src/box/retention_period.h
@@ -1,0 +1,68 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#include "trivia/config.h"
+
+#if defined(ENABLE_RETENTION_PERIOD)
+#include "retention_period_impl.h"
+#else /* !defined(ENABLE_RETENTION_PERIOD) */
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+#include "vclock/vclock.h"
+#include "xlog.h"
+
+/**
+ * Allocate vclock structure. In EE additional memory is allocated,
+ * where expiration time is saved.
+ */
+static inline struct vclock*
+retention_vclock_new(void)
+{
+	return (struct vclock *)xmalloc(sizeof(struct vclock));
+}
+
+/**
+ * Set expiration time of the @retention_vclock to now + @period.
+ */
+static inline void
+retention_vclock_set(struct vclock *retention_vclock, double period)
+{
+	(void)retention_vclock;
+	(void)period;
+}
+
+/**
+ * Update expiration time of all files. New period must be saved inside xdir.
+ */
+static inline void
+retention_index_update(struct xdir *xdir, double old_period)
+{
+	(void)xdir;
+	(void)old_period;
+}
+
+/**
+ * Return vclock of the oldest file, which is protected from garbage collection.
+ * Vclock is cleared, if none of the files are protected. Vclock must be
+ * non-nil.
+ */
+static inline void
+retention_index_get(vclockset_t *index, struct vclock *vclock)
+{
+	(void)index;
+	assert(vclock != NULL);
+	vclock_clear(vclock);
+}
+
+#if defined(__cplusplus)
+} /* extern "C" */
+#endif /* defined(__cplusplus) */
+
+#endif /* !defined(ENABLE_RETENTION_PERIOD) */
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -418,6 +418,7 @@ tx_notify_checkpoint(struct cmsg *msg)
 static void
 wal_writer_create(struct wal_writer *writer, enum wal_mode wal_mode,
 		  const char *wal_dirname, int64_t wal_max_size,
+		  double wal_retention_period,
 		  const struct tt_uuid *instance_uuid,
 		  wal_on_garbage_collection_f on_garbage_collection,
 		  wal_on_checkpoint_threshold_f on_checkpoint_threshold)
@@ -434,6 +435,11 @@ wal_writer_create(struct wal_writer *writer, enum wal_mode wal_mode,
 	struct xlog_opts opts = xlog_opts_default;
 	opts.sync_is_async = true;
 	xdir_create(&writer->wal_dir, wal_dirname, XLOG, instance_uuid, &opts);
+	/*
+	 * wal_retention_period must be set before gc is woken up.
+	 * Otherwise files which must be preserved can be deleted.
+	 */
+	xdir_set_retention_period(&writer->wal_dir, wal_retention_period);
 	xlog_clear(&writer->current_wal);
 	if (wal_mode == WAL_FSYNC)
 		writer->wal_dir.open_wflags |= O_SYNC;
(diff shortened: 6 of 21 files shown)


test/config-luatest/instance_config_schema_test.lua
```diff
@@ -43,7 +43,7 @@ local function validate_fields(config, record)
                 validate_fields(v1, v.items)
             end
         end
-        if v.type ~= 'record' or not v.enterprise_edition or is_enterprise then
+        if not v.enterprise_edition or is_enterprise then
```
Totktonada, line 46: The removed condition actually looks as a mistake: we should check that the keys are the same disregarding of the type of the value. I see that it is needed in `test_wal` below, since the `wal` section receives a first non-record enterprise-only option.

However, I would like to re-check it with Mergen.

@ImeevMA Do you confirm that this change is valid?

Totktonada (approved):
No objections from my side regarding the config part. Thank you for taking care of it!

test/config-luatest/instance_config_schema_test.lua
ImeevMA: It looks like it was most likely a mistake. I seem to have added this condition thinking `enterprise_edition` could only be `true` for `record`.

ochaplashkin (approved):
LGTM :rocket:

src/box/lua/load_cfg.lua
```diff
@@ -148,6 +154,7 @@ local default_cfg = {
     wal_dir_rescan_delay= 2,
     wal_queue_max_size  = 16 * 1024 * 1024,
     wal_cleanup_delay   = 4 * 3600,
+    wal_retention_delay = ifdef_wal_retention_delay(0),
```
locker, line 157: I find the option name rather unfortunate. In English it means "delay of retention" but actually it denotes the retention time interval so I'd call it `wal_retention_time` or `wal_retention_interval` or `wal_retention_period`. It might be wroth consulting the doc team regarding the option name.

src/box/wal_retention_delay.h
```diff
@@ -0,0 +1,55 @@
...
+			   struct cpipe *tx_prio_pipe)
+{
+	(void)wal_dir;
+	(void)current_wal;
+	(void)tx_prio_pipe;
+}
+
+/** Cleanup wal_retention_delay module. */
+static inline void
+wal_retention_delay_destroy(void)
```
locker, line 33: Nit: Subsystem constructor and destructor should be named `X_init` and `X_free`, e.g. `wal_init` and `wal_free`.

src/box/lua/load_cfg.lua
Serpentian: @tarantool/doc, what do you think of `wal_retention_delay` name for an option, which  postpones the deletion of wal files for a number of seconds?

grafin (approved):

src/box/lua/load_cfg.lua
```diff
@@ -497,6 +505,7 @@ local dynamic_cfg = {
     -- do nothing, affects new replicas, which query this value on start
     wal_dir_rescan_delay    = nop,
     wal_cleanup_delay       = private.cfg_set_wal_cleanup_delay,
+    wal_retention_delay     = private.cfg_set_wal_retention_delay,
```
sergepetrenko, line 508: Should also be in `dynamic_cfg_skip_at_load`, because you unconditionally pass the delay to wal_init.

Serpentian:
Skipped checkpatch warnings. There must be malloc, not xmalloc, as NULL is processed differently depending on the place. Comments on struct is not needed, it's pretty obvious

src/box/lua/load_cfg.lua
Serpentian: Done, thank you, didn't know about that.

src/box/xlog.h
```diff
@@ -159,6 +160,17 @@ struct xdir {
...
+	 * Number of seconds to delay garbage collection of files
+	 * after they are closed.
+	 */
+	double retention_delay;
+	/*
+	 * A list of files, protected from garbage collection. Always
+	 * ordered in non-decreasing order of vclocks and expire time.
+	 * New entries are added to the tail.
+	 */
+	struct rlist retention_index;
```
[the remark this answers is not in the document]
Serpentian: Done, though it looks like a crutch. But it significantly simplifies the implementation, so I'm ok with it.

src/box/box.cc
```diff
@@ -5848,8 +5859,11 @@ box_storage_init(void)
 	int64_t wal_max_size = box_check_wal_max_size(
 		cfg_geti64("wal_max_size"));
 	enum wal_mode wal_mode = box_check_wal_mode(cfg_gets("wal_mode"));
+	double wal_retention_delay = box_check_wal_retention_delay(
```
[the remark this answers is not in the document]
Serpentian: Tried to be consistent with the surrounding code:

https://github.com/tarantool/tarantool/blob/e844cd09b4a3b35395173f292d88f3755360435d/src/box/box.cc#L5848-L5850

Though, I like doing it your way more, so it's done

src/box/xlog.h
```diff
@@ -215,6 +250,12 @@ static inline bool
 xdir_has_garbage(struct xdir *dir, int64_t signature)
 {
 	struct vclock *vclock = vclockset_first(&dir->index);
+
+	struct vclock retention_vclock;
+	xdir_retention_vclock(dir, &retention_vclock);
+	if (vclock_is_set(&retention_vclock))
+		vclock_min_ignore0(vclock, &retention_vclock);
+
 	return vclock != NULL && vclock_sum(vclock) < signature;
```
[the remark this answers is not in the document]
Serpentian: Agree, we should unconditionally check signature of `retention_vclock` if it's set, as it's always >= `vclockset_first`. Also agree, that `ignore0` should not be used. Fixed

src/box/gc.c
```diff
@@ -213,6 +213,11 @@ gc_run_cleanup(void)
 		consumer = gc_tree_next(&gc.consumers, consumer);
 	}
 
+	struct vclock retention_vclock;
+	wal_get_retention_vclock(&retention_vclock);
+	if (vclock_is_set(&retention_vclock))
+		vclock_min_ignore0(&min_vclock, &retention_vclock);
```
sergepetrenko, line 219: `retention_vclock` has a non-zero 0-th component, so `min_vclock` should account for it.
We ignore 0-th vclock component only for gc consumers.

src/box/retention_delay.h
```diff
@@ -0,0 +1,85 @@
...
+#include "vclock/vclock.h"
+#include "xlog.h"
+
+/**
+ * Allocate vclock structure. In EE additional memory is allocated,
+ * where expiration time is saved. Note, that the returned value
+ * must be processed, as xmalloc is not used.
+ */
+static inline struct vclock*
+retention_vclock_create(void)
```
sergepetrenko, line 43: Nit: this should be called `retention_vclock_new()`.

See https://www.tarantool.io/en/doc/latest/dev_guide/c_style_guide/#chapter-4-naming
```
For function naming we have a convention is to use:

    new/delete for functions which allocate + initialize and destroy + deallocate an object,
    create/destroy for functions which initialize/destroy an object but do not handle memory management,
    init/free for functions which initialize/destroy libraries and subsystems.

```

sergepetrenko (changes requested):
Thanks for working on this!
Looks much cleaner now. I only have two comments left.

src/box/xlog.c
```diff
@@ -420,10 +422,11 @@ xdir_index_file(struct xdir *dir, int64_t signature)
...
 
 	/*
-	 * Append the clock describing the file to the
-	 * directory index.
+	 * Append the clock describing the file to the directory index.
+	 * If retention feature is available (EE), then additional memory is
+	 * allocated. Time, when xlog protection should stop is saved there.
 	 */
-	struct vclock *vclock = (struct vclock *) malloc(sizeof(*vclock));
+	struct vclock *vclock = retention_vclock_create();
```
sergepetrenko, line 429: It seems you may actually use `xmalloc` in `retention_vclock_create()`.
Looks like all the code which checks `xdir_scan` errors (it calls `xdir_index_file`), makes tarantool exit with an error.
So I see nothing bad in panicking in case of malloc failure.

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: I see, thanks!

src/box/xlog.h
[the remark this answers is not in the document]
sergepetrenko: Looks nice (although I haven't seen the EE part yet)!

src/box/lua/load_cfg.lua
sergepetrenko: @andreyaksenov

sergepetrenko:
`box_check_wal_retention_delay()` should also be called from `box_check_config()`. Otherwise tarantool exits, when I pass an out of bounds value in EE:
```
tarantool> box.cfg{readahead=-1}
---
- error: 'Incorrect value for option ''readahead'': specified value is out of bounds'
...

tarantool>
```
compare to this:
```
tarantool> box.cfg{wal_retention_delay=-1}
<stripped>
2023-12-13 12:24:19.433 [62856] main/103/interactive box.cc:1663 E> ER_CFG: Incorrect value for option 'wal_retention_delay': the value must be >= 0
2023-12-13 12:24:19.433 [62856] main/103/interactive F> can't initialize storage: Incorrect value for option 'wal_retention_delay': the value must be >= 0
```

src/box/lua/load_cfg.lua
andreyaksenov: I vote for `wal_retention_period`. Also found similar names in other docs.

src/box/gc.c
Serpentian: Sure, let's use `vclock_compare` instead, as there's no `vclock_min` method.

src/box/retention_delay.h
Serpentian: Sure, fixed

src/box/xlog.c
Serpentian: It was `xlog_cursor_close`, which was called, when `vclock` allocation failed in `xdir_index_file`. Will it be closed correctly in case of panic?

Anyway, I've used `xmalloc` for now.

src/box/gc.c
sergepetrenko: > Sure, let's use vclock_compare instead, as there's no vclock_min method.

Why don't you add one?

It seems with `vclock_compare` you may miss the case when vclocks are incomparable.

For example:
`retention_vclock: {1: 1, 2: 2}`
`min_vclock: {1:2, 2: 1}`,
vclocks are incomparable, so `min_vclock` remains as is, even though it should become `{1: 1, 2:1}`.

src/box/retention_period.h
```diff
@@ -0,0 +1,83 @@
...
+static inline void
+retention_index_update(struct xdir *xdir, double old_period)
+{
+	(void)xdir;
+	(void)old_period;
+}
+
+/**
+ * Return vclock of the oldest file, which is protected from garbage collection.
+ * Vclock is cleared, if none of a files are protected. Vclock must be non-nil.
```
sergepetrenko, line 69: typo: `none of the files`

src/box/wal.c
```diff
@@ -782,6 +791,58 @@ wal_set_queue_max_size(int64_t size)
...
+{
+	struct wal_writer *writer = &wal_writer_singleton;
+	if (writer->wal_dir.retention_period == 0) {
+		vclock_clear(vclock);
+		return;
+	}
+	struct wal_vclock_msg msg;
+	cbus_call(&writer->wal_pipe, &writer->tx_prio_pipe,
+		  &msg.base, wal_get_retention_vclock_f);
+	if (vclock != NULL)
```
sergepetrenko, line 842: Nit: you may assert vclock != NULL in the beginning, or return immediately if vclock == NULL.
Besides you already `vclock_clear`'d the vclock without checking above.

sergepetrenko (changes requested):
Thanks for the fixes!
A couple of minor comments left.

src/box/xlog.c
sergepetrenko: > Will it be closed correctly in case of panic?

Yes. Besides, there are other places where we panic, and none of the open files are closed in these cases.

src/box/CMakeLists.txt
```diff
@@ -177,7 +177,12 @@ endif()
 add_library(tuple STATIC ${tuple_sources})
```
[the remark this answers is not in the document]
Serpentian: Now it's consistent with EE and doesn't mention `retention_delay`

src/box/gc.c
Serpentian: > It seems with vclock_compare you may miss the case when vclocks are incomparable.

Now I see, thank you! Introduced `min_vclock`

src/box/wal.c
Serpentian: Indeed, fixed

src/box/gc.c
```diff
@@ -213,6 +213,15 @@ gc_run_cleanup(void)
 		consumer = gc_tree_next(&gc.consumers, consumer);
 	}
 
+	/*
+	 * Acquire minimym vclock of a file, which is protected from garbage
```
sergepetrenko, line 217: typo: minimum

src/lib/vclock/vclock.h
```diff
@@ -457,6 +468,11 @@ vclockset_match(vclockset_t *set, const struct vclock *key)
 	return vclockset_first(set);
 }
 
+#define vclockset_foreach(set, vclock) \
+	for ((vclock) = vclockset_first(set); \
+	     (vclock) != NULL; \
+	     (vclock) = vclockset_next((set), (vclock)))
+
```
sergepetrenko, line 475: Please, introduce `vclockset_foreach` and `vclock_min/max` as separate commits.

src/lib/vclock/vclock.h
```diff
@@ -374,23 +374,27 @@ vclock_lex_compare(const struct vclock *a, const struct vclock *b)
...
- * vclock[0] into account.
+ * maximum (when @sign is +1) between vclocks @a a and @a b.
  *
  * @param[in,out] a Vclock containing the minimum components.
  * @param b Vclock to compare with.
  * @param sign Which extremum to find:
  *             -1 for minimum,
  *             +1 for maximum.
+ * @param ignore_zero Whether to order by 0-th component or not.
+ *        The 0 component is ignored on anonymous replicas when
```
sergepetrenko, line 385: I'd say the 0th vclock component is ignored in replication generally. No only on anonymous replicas.
Every replica has its own 0th vclock component and uses it for local space writes

src/box/xlog.h
```diff
@@ -215,7 +241,16 @@ static inline bool
...
 	struct vclock *vclock = vclockset_first(&dir->index);
-	return vclock != NULL && vclock_sum(vclock) < signature;
+	if (vclock == NULL)
+		return false;
+
+	struct vclock retention_vclock;
+	xdir_get_retention_vclock(dir, &retention_vclock);
+	if (vclock_is_set(&retention_vclock))
+		/** retention_vclock is >= vclock. */
+		return vclock_sum(&retention_vclock) < signature;
```
sergepetrenko, line 251: Wait. When `xdir_has_garbage` returns true, it means that we may delete something.

But `vclock_sum(&retention_vclock) < signature` means that `signature` is protected by retention vclock.

Shouldn't it be
```c
if (vclock_is_set(&retention_vclock)) {
    return vclock_sum(&retention_vclock) >= signature && vclock_sum(vclock) < signature;
}
return vclock_sum(vclock) < signature;
```
?

sergepetrenko (approved):
Thanks for your service!

Good job with all the fixes. I've left a couple of comments, and I think that's all.

src/box/xlog.h
Serpentian: You're right, it won't work. `xdir_has_garbage` always returns false, when retention is used. However, your approach is not fully correct too. Even though it will work now, when `xdir_has_garbage` is used only in one place, inside `wal_fallocate` (`signature` is always retention), it won't in general. 

---

When there's `retention_vclock` set, we may have the following situations:

1. Situation 1. `vclock_sum(vclock)` is 10, `vclock_sum(retention_vclock)` is 20. `signature` is 21.
```
index                     10    20    30    ...
retention                       20    30    ...
```

Your return value will be `false`, as retention sum is < than signature, but we have garbage, which is 10. My return value is correct.

2. Situation 2. The same, as in the 1st one, but `signature` is 19. Your return value correctly returns true, mine - false.
3. Sitation 3.  `vclock_sum(vclock)` is 10, `vclock_sum(retention_vclock)` is 10. `signature` is 11.
```
index                     10    20    30    ...
retention                 10    20    30    ...
```
Mine returns true, which is wrong, no garbage. Your - false, which is correct. 

----

I suppose, we must introduce more general solution, which works in all cases. If retention is used, we must find a vclock, which is less, than a signature and which is not protected by gc. If retention is set and it's not the first in the index, then it's enough to check, that the first index is less than a signature. If retention is used, but retention == to the first index, then all files are protected. Please, check implementation.

I also added unit tests (in EE part), as I made such a critical error in the function. Thank you for noticing!

src/lib/vclock/vclock.h
Serpentian: Done, also added tests

src/lib/vclock/vclock.h
Serpentian: Fixed

locker (approved):

src/box/xlog.h
sergepetrenko: Thanks for paying attention! I also made a mistake.

The current version looks good to me, but I suggest applying this small diff. It doesn't change anything now, because `retention` can't be smaller or incomparable with `vclock`, but still looks safer.
```diff
diff --git a/src/box/xlog.h b/src/box/xlog.h
index 5310ac52c0..b60e30d335 100644
--- a/src/box/xlog.h
+++ b/src/box/xlog.h
@@ -246,8 +246,10 @@ xdir_has_garbage(struct xdir *dir, int64_t signature)
 
        struct vclock retention;
        xdir_get_retention_vclock(dir, &retention);
-       if (!vclock_is_set(&retention) || vclock_compare(&retention, vclock) != 0)
+       if (!vclock_is_set(&retention) ||
+           vclock_compare(vclock, &retention) < 0) {
                return vclock_sum(vclock) < signature;
+       }
 
        /** All files are protected from gc. */
        return false;
```

src/box/retention_period.h
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2023 VK Company Limited. All Rights Reserved.
```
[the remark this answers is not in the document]
Serpentian: My bad, fixed

src/box/xlog.h
Serpentian: Sure, agree. Fixed

src/box/xlog.h
```diff
@@ -215,7 +241,17 @@ static inline bool
...
 {
 	struct vclock *vclock = vclockset_first(&dir->index);
-	return vclock != NULL && vclock_sum(vclock) < signature;
+	if (vclock == NULL)
+		return false;
+
+	struct vclock retention;
+	xdir_get_retention_vclock(dir, &retention);
+	if (!vclock_is_set(&retention) ||
+	    vclock_compare(&retention, vclock) < 0)
```
sergepetrenko, line 250: I don't get why shouldn't it be `vclock_compare(vclock, &retention) < 0`?

Serpentian: This is an error, the test was also broken: I didn't initialize `expiration_time` of vclocks, which were before the retention. So, `retention_index_get` always returned uninitialized vclock, check was always true. Last `false` was never returned. 

Fixed this and the test in EE

sergepetrenko: ¡Gracias!

PR #9073, opened 2023-08-30
Lua fuzzer fixes (return and break statements)

rtolkacheva:
Revised:

1. The serializer no longer generates `return` and `break` statements without a corresponding function or loop, so they are now skipped.
2. The serializer was generating `else if` instead of `elseif`. Due to this typo, several errors occurred as a result of unclosed `if` statements.

I generated a set of 50,000 protobuf struct samples and tested the fuzzer on this set. Here are my results:
  | Before | After | Diff (% of errors of the kind)
-- | -- | -- | --
TOTAL | 27775 | 18294 | -34.14%
attempt to call | 5897 | 6963 | 18.08%
attempt to index | 4735 | 5790 | 22.28%
'for' initial value must be a | 1825 | 2820 | 54.52%
attempt to perform arithmetic on | 1636 | 1214 | -25.79%
attempt to compare | 524 | 502 | -4.20%
cannot use '...' outside a vararg function near '...' | 311 | 272 | -12.54%
attempt to get length of | 273 | 236 | -13.55%
table index is | 146 | 145 | -0.68%
attempt to concatenate | 177 | 109 | -38.42%
ambiguous syntax | 99 | 100 | 1.01%
'for' limit must be a | 68 | 82 | 20.59%
unexpected symbol near | 67 | 57 | -14.93%
'for' step must be a | 4 | 4 | 0.00%
no loop to break | 11150 | 0 | -100.00%
'end' expected | 863 | 0 | -100.00%

I got these results:
1. Due to the skipped `break` statements, 11,150 errors (40% of all errors) disappeared.
2. Due to the skipped `return` statements, 599 errors (2% of all errors) disappeared. However, it's worth noting that these 599 errors were not caused by the absence of the `return` statement itself, but rather by errors in the return expressions.
3. Due to fixed `elseif` statements 863 errors (3% of all errors) disappeared.
4. New errors appeared in 3131 samples (11% of all errors, 25% of disappeared errors).

### DIFF
--- a/test/fuzz/luaL_loadbuffer/serializer.cc
+++ b/test/fuzz/luaL_loadbuffer/serializer.cc
@@ -611,7 +611,7 @@ PROTO_TOSTRING(IfStatement, statement)
 
 NESTED_PROTO_TOSTRING(ElseIfBlock, elseifblock, IfStatement)
 {
-	std::string elseifblock_str = "else if ";
+	std::string elseifblock_str = "elseif ";
 	elseifblock_str += ExpressionToString(elseifblock.condition());
 	elseifblock_str += " then\n\t";
 	elseifblock_str += BlockToString(elseifblock.block());
--- a/test/fuzz/luaL_loadbuffer/serializer.cc
+++ b/test/fuzz/luaL_loadbuffer/serializer.cc
@@ -8,6 +8,8 @@
 #include <stack>
 #include <string>
 
+#include <trivia/util.h>
+
 using namespace lua_grammar;
 
 #define PROTO_TOSTRING(TYPE, VAR_NAME) \
@@ -185,15 +187,100 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
 }
 
 /**
+ * Class that registers and provides context during code
+ * generation.
+ * Used to generate correct Lua code.
+ */
+class Context {
+public:
+	enum class BlockType {
+		kReturnable,
+		kBreakable,
+	};
+
+	void step_in(BlockType type)
+	{
+		block_stack_.push(type);
+		if (type == BlockType::kReturnable) {
+			++returnable_counter_;
+		}
+	}
+
+	void step_out()
+	{
+		assert(!block_stack_.empty());
+		if (block_stack_.top() == BlockType::kReturnable) {
+			assert(returnable_counter_ > 0);
+			--returnable_counter_;
+		}
+		block_stack_.pop();
+	}
+
+	std::string get_next_block_setup()
+	{
+		std::size_t id = GetCounterIdProvider().next();
+		std::string counter_name = GetCounterName(id);
+
+		return GetCondition(counter_name, get_exit_statement_()) +
+		       GetCounterIncrement(counter_name);
+	}
+
+	bool break_is_possible()
+	{
+		return !block_stack_.empty() &&
+		       block_stack_.top() == BlockType::kBreakable;
+	}
+
+	bool return_is_possible()
+	{
+		return returnable_counter_ > 0;
+	}
+
+private:
+
+	std::string get_exit_statement_()
+	{
+		assert(!block_stack_.empty());
+		switch (block_stack_.top()) {
+		case BlockType::kBreakable:
+			return "break";
+		case BlockType::kReturnable:
+			return "return";
+		}
+		unreachable();
+	}
+
+	std::stack<BlockType> block_stack_;
+	/*
+	 * The returnable block can be exited with return from
+	 * the breakable block within it, but the breakable block
+	 * cannot be exited with break from the returnable block within
+	 * it.
+	 * Valid code:
+	 * `function foo() while true do return end end`
+	 * Erroneous code:
+	 * `while true do function foo() break end end`
+	 * This counter is used to check if `return` is possible.
+	 */
+	uint64_t returnable_counter_ = 0;
+};
+
+Context&
+GetContext()
+{
+	static Context context;
+	return context;
+}
+
+/**
  * Block may be placed not only in a cycle, so specially for cycles
  * there is a function that will add a break condition and a
  * counter increment.
  */
 std::string
-BlockToStringCycleProtected(const Block &block, const std::string &counter_name)
+BlockToStringCycleProtected(const Block &block)
 {
-	std::string retval = GetCondition(counter_name, "break");
-	retval += GetCounterIncrement(counter_name);
+	std::string retval = GetContext().get_next_block_setup();
 	retval += ChunkToString(block.chunk());
 	return retval;
 }
@@ -204,11 +291,10 @@ BlockToStringCycleProtected(const Block &block, const std::string &counter_name)
  * BlockToStringCycleProtected().
  */
 std::string
-DoBlockToStringCycleProtected(const DoBlock &block,
-			      const std::string &counter_name)
+DoBlockToStringCycleProtected(const DoBlock &block)
 {
 	std::string retval = "do\n";
-	retval += BlockToStringCycleProtected(block.block(), counter_name);
+	retval += BlockToStringCycleProtected(block.block());
 	retval += "end\n";
 	return retval;
 }
@@ -219,8 +305,7 @@ DoBlockToStringCycleProtected(const DoBlock &block,
  * increment.
  */
 std::string
-FuncBodyToStringReqProtected(const FuncBody &body,
-			     const std::string &counter_name)
+FuncBodyToStringReqProtected(const FuncBody &body)
 {
 	std::string body_str = "( ";
 	if (body.has_parlist()) {
@@ -228,8 +313,7 @@ FuncBodyToStringReqProtected(const FuncBody &body,
 	}
 	body_str += " )\n\t";
 
-	body_str += GetCondition(counter_name, "return");
-	body_str += GetCounterIncrement(counter_name);
+	body_str += GetContext().get_next_block_setup();
 
 	body_str += BlockToString(body.block());
 	body_str += "end\n";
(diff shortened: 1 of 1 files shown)


test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -184,16 +184,98 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
...
+ * Used to generate correct Lua code.
+ */
+class Context {
+public:
+	enum BlockType {
+		Returnable,
+		Breakable,
+	};
+
+	void step_in(BlockType type)
```
ligurio, line 199: Don't mix snake-case and camel-case names styles. Here and below.

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -184,16 +184,98 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
...
+		block_stack_.pop();
+	}
+
+	std::string get_next_block_setup()
+	{
+		std::size_t id = GetCounterIdProvider().next();
+		std::string counter_name = GetCounterName(id);
+
+		return GetCondition(counter_name, get_exit_statement_()) +
+			GetCounterIncrement(counter_name);
```
ligurio, line 223: bad indendation

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -607,7 +607,7 @@ PROTO_TOSTRING(IfStatement, statement)
```
ligurio, line 607: It is worth to explain *why* you changed `else if` to `elseif`.
I propose to tell something like: from the beginning serializer uses keywords `else if` in conditions, however in Lua [^1] these keywords used without whitespace.

I would add this to commit message too:
> fixed elseif solved 863 (2%) errors

[^1]: https://www.lua.org/manual/5.1/manual.html#2.1

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -235,15 +235,16 @@ class Context {
 	}
```
ligurio, line 235: commit message: it is worth to explain why do you add these checks.
You say this in PR description, I would add stats to commit message too:
> Skipped break without cycle solved 11150 (22%) errors, skipped return without function solved 599 (1%) errors

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -184,16 +184,98 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
 	return retval;
 }
```
ligurio, line 185: commit message: 
please add a wording about introduced context and say that this commit is needed for two next commits.

ligurio (changes requested):
Rimma, thank you for patches!
See my comments inline.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Here are some conventions regarding naming styles in C++ to improve readability:
1. Variables should be in snake_case;
2. Type names should be in PascalCase;
3. Consts (even ones in enum) should be like `kConstValue`.

It's quite different how function should be named: STL uses snake_case, [Qt uses](https://wiki.qt.io/Qt_Coding_Style) camelCase, [Google](https://google.github.io/styleguide/cppguide.html#Function_Names) and [Yandex](https://github.com/yandex/CMICOT/blob/master/CPP_STYLE_GUIDE.md#variables) use PascalCase.

In my previous PR we agreed to use PascalCase for non-member functions and snake_case for member functions, so in this PR I use these naming style.

Also, I made a mistake by not naming the enum values as constants, but I've now fixed this error.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Fixed.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Fixed.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Fixed.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Fixed.

test/fuzz/luaL_loadbuffer/serializer.cc
ligurio: Okay, thanks!

test/fuzz/luaL_loadbuffer/serializer.cc
ligurio: Thanks!

test/fuzz/luaL_loadbuffer/serializer.cc
ligurio: Thanks!

ligurio (approved):
LGTM

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -184,16 +184,99 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
...
+
+	std::string get_exit_statement_()
+	{
+		assert(!block_stack_.empty());
+		switch (block_stack_.top()) {
+		case BlockType::kBreakable:
+			return "break";
+		case BlockType::kReturnable:
+			return "return";
+		}
```
Buristan, line 247: `default` case is missing.

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -184,16 +184,99 @@ GetCondition(const std::string &counter_name, const std::string &then_block)
...
+	}
+
+	std::stack<BlockType> block_stack_;
+	/*
+	 * Returnable block can be exited with return from
```
Buristan, line 253: Typo: s/Returnable/The returnable/
Typo: s/Breakable/The breakable/

Buristan (commented):
Hi, Rimma!
Thanks for the patch set!
It's really great improvement for our fuzzing.

I'll procced with the review per patch below.

[PATCH 1/3] test/fuzz: introduce class Context

LGTM, with several nits below.

> Context created to manage context of Lua program.

I suggest to reformulate it as the following:
```The context object is created to manage the context of the Lua progam.```

File: test/fuzz/luaL_loadbuffer/serializer.cc [hunk 8]
```diff
@@ -245,6 +245,7 @@ class Context {
 		case BlockType::kReturnable:
 			return "return";
 		}
+		return "";
```
Buristan, line 248: Ok, I found missing default case:).
Please, add this as the `default` case in the previous commit.

test/fuzz/luaL_loadbuffer/serializer.cc
```diff
@@ -384,20 +385,26 @@ PROTO_TOSTRING(LastStatement, laststat)
 	using LastStatType = LastStatement::LastOneofCase;
 	switch (laststat.last_oneof_case()) {
 	case LastStatType::kExplist:
-		laststat_str = ReturnOptionalExpressionListToString(
-			laststat.explist());
+		if (GetContext().return_is_possible()) {
```
Buristan, line 388: We may add this condition as the part of `ReturnOptionalExpressionListToString()` instead. Here and below.
Feel free to ignore.

Buristan, line 388: Minor: brackets are excess. Here and below.

Buristan (commented):
[PATCH 2/3] test/fuzz: add return and break checks

Generally LGTM, after fixing my comments below.

Buristan (approved):
[PATCH 3/3] test/fuzz: fixed ElseIfBlock serialization

LGTM.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: It is a switch on an enumerated value, so a default case was skipped by choice. The compiler will produce a warning if any values are not handled.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: This return was added by a mistake.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: As in my previous PR, I want to keep brackets.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: An integer can be converted to any enum in C++, so if you want to protect this switch statement, I would prefer to add unreachable() from[ src/trivia/util.h](https://github.com/tarantool/tarantool/blob/06be32fbbdb884e407a923615606a3c0db599453/src/trivia/util.h#L260-L264) to keep compiler warnings for the future.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Moved to `ReturnOptionalExpressionListToString()`.

test/fuzz/luaL_loadbuffer/serializer.cc
rtolkacheva: Fixed.

test/fuzz/luaL_loadbuffer/serializer.cc
Buristan: We may add `unreachable()` *and* `assert(0)` to the `default` statement to be sure that this will be covered in the future.

rtolkacheva: If some new enumerators will be added, and this switch won't be expanded, this mistake will be noticed only in a runtime. So I want  `unreachable()` / `assert(0)` after the switch statement.

Buristan: OK, I get it now. Thanks for the explanation. LGTM.

PR #9083, opened 2023-09-01
tools: add a tool to check downgrade versions list

### LINKED ISSUE #8319: Provide tool to check that box.schema.downgrade_versions() knows all released Tarantool versions
`box.schema.downgrade()` downgrades only to versions listed in `box.schema.downgrade_versions()`. Thus on Tarantool release we need to check that we don't forget to add to the list version being released and all the versions released in downstream branches since last release. (Of course we need to add appropriate downgrade actions too.) It will be convenient to have to tool that makes such a check.

nshy:
We need to make sure the downgrade versions list in `src/box/lua/upgrade.lua` has all the released versions in downstream branches. As well as the version being released itself. Here we add a tool to make this check.

Usage:

- checkout the tag being released (should be annotated tag)
- run ./tools/check-downgrade-versions.sh

If the downgrade versions list is not correct then the difference between the current list and correct one will be printed.

Close #8319

### DIFF
--- a/src/box/lua/upgrade.lua
+++ b/src/box/lua/upgrade.lua
@@ -2095,6 +2095,7 @@ end
 
 -- List of all Tarantool releases we can downgrade to.
 local downgrade_versions = {
+    -- DOWNGRADE VERSIONS START
     "2.8.2",
     "2.8.3",
     "2.8.4",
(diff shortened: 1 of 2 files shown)


locker (changes requested):

nshy:
Superseded by #10135

PR #9084, opened 2023-09-01
static-build: return `CMAKE_FIND_USE_CMAKE_SYSTEM_PATH=FALSE` back

### LINKED ISSUE #6998: libunwind submodule follow-up fixes

Gumix:
This option controls the default behavior of the following commands for whether or not to search paths provided by platform-specific cmake variables:
 - find_program()
 - find_library()
 - find_file()
 - find_path()
 - find_package()

By default this variable is not set, which is equivalent to it having a value of TRUE.

`-DCMAKE_FIND_USE_CMAKE_SYSTEM_PATH=FALSE` was initially added by commit 800e5ed617f7 ("build: refactor static build process"), but it was later removed by commit 7dc9fe44a5ae ("libunwind: add new submodule").
It seems that we can return it back.

Closes #6998


static-build/cmake/AddTarantoolProject.cmake
```diff
@@ -16,6 +16,8 @@ ExternalProject_Add(tarantool
         -DCMAKE_INSTALL_LOCALSTATEDIR=<INSTALL_DIR>/var
         -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
         -DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}
+        # Ensure that no external dependencies are dynamically linked.
+        -DCMAKE_FIND_USE_CMAKE_SYSTEM_PATH=FALSE
```
Totktonada, line 20: AFAIR, @CuriousGeorgiy had some specific reason to eliminate the option: something goes wrong with it. Sadly, I don't remember the exact reason. It worth to re-check it I guess.

CuriousGeorgiy: Just try to run static build on macOS, and the problem should probably manifestate itself.

CuriousGeorgiy: As far as I vaguely remember, we have a bug in the static build setup in that we link against the `.dylib` instead of the `.tlb` and this doesn’t work for system libraries like libunwind.

This didn’t manifestate itself before I refactored the libunwind library setup, but when it did, I just used this workaround for some reason.

CuriousGeorgiy: https://discourse.cmake.org/t/newer-versions-of-macos-require-linking-against-tbd-files-and-not-old-system-paths-to-dylib/6871

Totktonada: But we link libunwind statically in the static build, correct? How .dylib/.tbd are involved?

CuriousGeorgiy: > But we link libunwind statically in the static build

libunwind on macOS is a system library (taken from the [LLVM project](https://github.com/llvm/llvm-project/tree/main/libunwind)), it has nothing to do with the [libunwind library](https://github.com/libunwind/libunwind) we use on Linux/FreeBSD.

CuriousGeorgiy: As far as I see right now from the comment above (can't check this right now), the problem is that we need to link against the `System` umbrella framework instead of the `libunwind` library, and we can't find the `System` umbrella framework if we set `CMAKE_FIND_USE_CMAKE_SYSTEM_PATH=FALSE`.

Gumix: Yes, reproduced on `osx_static_cmake`:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIBUNWIND_INCLUDE_DIR (ADVANCED)
```
I will close this PR.

Issue #9078, opened 2023-09-01
Introduce roles in config


ImeevMA:
- [x] Introduce base support for roles https://github.com/tarantool/tarantool/pull/9087)
- [x] Implement `dependencies` for roles (https://github.com/tarantool/tarantool/pull/9338).
- [ ] ~Implement a way to account the necessary privileges for roles.~ Postponed.

Totktonada:
The main scope is done. I propose to track the permissions enhancements separately. Closing so.

PR #9087, opened 2023-09-03
config: introduce roles

ImeevMA:
This patch introduces initial support for roles. Dependencies are not
currently supported for roles.

Part of https://github.com/tarantool/tarantool/issues/9078

@TarantoolBot document
Title: Roles

Two new options have been added: "roles" and "roles_cfg". The first one
is an array and the second one is a map. Each of these can be defined
per instance, replica set, group, and globally. As with almost all other
options, with the exception of those defined as 'map', the 'roles'
option for the lower scope will replace the roles for the higher scope.
Value roles_cfg however defined as "map", so it will be merged.

The "roles" option defines the roles for each instance. A role is a
program that runs when a configuration is loaded or reloaded. If a role
is defined more than once on an instance, it will still only be run
once. Three functions must be defined in the role: validate(), apply()
and stop(). Each of these functions should throw an error if it occurs.

The "roles_cfg" option specifies the configuration for each role. In
this option, the role name is the key and the role configuration is the
value.

On each run, all roles will be loaded (if necessary) in the order in
which they were specified; the configuration for each role will then be
validated using the corresponding validate() function in the same order;
and then they will all be run with apply() function in the same order.
If some roles have been removed from the instance, they will be stopped
in reverse order using the stop() function.

Example of a role structure:
```Lua
local M = {}

-- Validates configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the validation fail.
function M.validate(cfg)
    -- <...>
end

-- Applies the given configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the given configuration can't be applied.
function M.apply(cfg)
    -- <...>
end

-- Stops the role.
--
-- Called on configuration reload if the role was enabled before
-- and removed now from the list of roles of the given instance.
--
-- Should cancel all background fibers and clean up hold
-- resources.
--
-- Must raise an error if this action can't be performed.
function M.stop()
    -- <...>
end

return M
```

### DIFF
--- a/changelogs/unreleased/gh-9078-roles.md
+++ b/changelogs/unreleased/gh-9078-roles.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Introduced the initial support for roles - programs that run when
+  a configuration is loaded or reloaded (gh-9078).
--- a/src/box/lua/config/applier/roles.lua
+++ b/src/box/lua/config/applier/roles.lua
@@ -1,0 +1,87 @@
+local log = require('internal.config.utils.log')
+
+local last_loaded = {}
+local last_loaded_names_ordered = {}
+
+local function stop_roles(roles_to_skip)
+    for id = #last_loaded_names_ordered, 1, -1 do
+        local role_name = last_loaded_names_ordered[id]
+        if roles_to_skip == nil or roles_to_skip[role_name] == nil then
+            log.verbose('roles.apply: stop role ' .. role_name)
+            local ok, err = pcall(last_loaded[role_name].stop)
+            if not ok then
+                error(('Error stopping role %s: %s'):format(role_name, err), 0)
+            end
+        end
+    end
+end
+
+local function apply(config)
+    local configdata = config._configdata
+    local role_names = configdata:get('roles', {use_default = true})
+    if role_names == nil or next(role_names) == nil then
+        stop_roles()
+        return
+    end
+
+    -- Remove duplicates.
+    local roles = {}
+    local roles_ordered = {}
+    for _, role_name in pairs(role_names) do
+        if roles[role_name] == nil then
+            table.insert(roles_ordered, role_name)
+        end
+        roles[role_name] = true
+    end
+
+    -- Stop removed roles.
+    stop_roles(roles)
+
+    -- Run roles.
+    local roles_cfg = configdata:get('roles_cfg', {use_default = true}) or {}
+    local loaded = {}
+    local loaded_names_ordered = {}
+
+    -- Load roles.
+    for _, role_name in ipairs(roles_ordered) do
+        local role = last_loaded[role_name]
+        if not role then
+            log.verbose('roles.apply: load role ' .. role_name)
+            role = require(role_name)
+            local funcs = {'validate', 'apply', 'stop'}
+            for _, func_name in pairs(funcs) do
+                if type(role[func_name]) ~= 'function' then
+                    local err = 'Role %s does not contain function %s'
+                    error(err:format(role_name, func_name), 0)
+                end
+            end
+        end
+        loaded[role_name] = role
+        table.insert(loaded_names_ordered, role_name)
+    end
+
+    -- Validate configs for all roles.
+    for _, role_name in ipairs(roles_ordered) do
+        local ok, err = pcall(loaded[role_name].validate, roles_cfg[role_name])
+        if not ok then
+            error(('Wrong config for role %s: %s'):format(role_name, err), 0)
+        end
+    end
+
+    -- Apply configs for all roles.
+    for _, role_name in ipairs(roles_ordered) do
+        log.verbose('roles.apply: apply config for role ' .. role_name)
+        local ok, err = pcall(loaded[role_name].apply, roles_cfg[role_name])
+        if not ok then
+            error(('Error applying role %s: %s'):format(role_name, err), 0)
+        end
+    end
+
+    last_loaded = loaded
+    last_loaded_names_ordered = loaded_names_ordered
+end
+
+return {
+    name = 'roles',
+    apply = apply,
+}
--- a/test/config-luatest/helpers.lua
+++ b/test/config-luatest/helpers.lua
@@ -88,11 +88,18 @@ local simple_config = {
 
 local function prepare_case(g, opts)
     local dir = opts.dir
+    local roles = opts.roles
     local script = opts.script
     local options = opts.options
 
     if dir == nil then
         dir = treegen.prepare_directory(g, {}, {})
+    end
+
+    if roles ~= nil and next(roles) ~= nil then
+        for name, body in pairs(roles) do
+            treegen.write_script(dir, name .. '.lua', body)
+        end
     end
 
     if script ~= nil then
@@ -135,6 +142,10 @@ end
 -- Start a server with the given script and the given
 -- configuration, run a verification function on it.
 --
+-- * opts.roles
+--
+--   Role codes for writing into corresponding files.
+--
 -- * opts.script
 --
 --   Code write into the main.lua file.
@@ -164,6 +175,7 @@ end
 -- Start tarantool process with the given script/config and check
 -- the error.
 --
+-- * opts.roles
 -- * opts.script
 -- * opts.options
 --
@@ -185,11 +197,16 @@ end
 -- Start a server, write a new script/config, reload, run a
 -- verification function.
 --
+-- * opts.roles
 -- * opts.script
 -- * opts.options
 -- * opts.verify
 --
 --   Same as in success_case().
+--
+-- * opts.roles_2
+--
+--   A new list of roles to prepare before config:reload().
 --
 -- * opts.script_2
 --
@@ -205,6 +222,7 @@ end
 --
 --   Verify test invariants after config:reload().
 local function reload_success_case(g, opts)
+    local roles_2 = opts.roles_2
     local script_2 = opts.script_2
     local options = assert(opts.options)
     local verify_2 = assert(opts.verify_2)
@@ -214,6 +232,7 @@ local function reload_success_case(g, opts)
 
     prepare_case(g, {
         dir = prepared.dir,
+        roles = roles_2,
         script = script_2,
         options = options_2,
     })
@@ -227,31 +246,44 @@ end
 -- Start a server, write a new script/config, reload, run a
 -- verification function.
 --
+-- * opts.roles
 -- * opts.script
 -- * opts.options
 -- * opts.verify
 --
 --   Same as in success_case().
 --
+-- * opts.roles_2
+--
+--   A new list of roles to prepare before config:reload().
+--
 -- * opts.script_2
 --
 --   A new script to write into the main.lua file before
 --   config:reload().
 --
+-- * opts.options_2
+--
+--   A new config to use for the config:reload(). It is optional,
+--   if not provided opts.options is used instead.
+--
 -- * opts.exp_err
 --
 --   An error that config:reload() must raise.
 local function reload_failure_case(g, opts)
-    local script_2 = assert(opts.script_2)
+    local script_2 = opts.script_2
+    local roles_2 = opts.roles_2
     local options = assert(opts.options)
+    local options_2 = opts.options_2 or options
     local exp_err = assert(opts.exp_err)
 
     local prepared = success_case(g, opts)
 
     prepare_case(g, {
         dir = prepared.dir,
+        roles = roles_2,
         script = script_2,
-        options = options,
+        options = options_2,
     })
     t.assert_error_msg_equals(exp_err, g.server.exec, g.server, function()
         local config = require('config')
--- a/test/config-luatest/roles_test.lua
+++ b/test/config-luatest/roles_test.lua
@@ -1,0 +1,355 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+-- Make sure the role is properly loaded.
+g.test_single_role_success = function(g)
+    local one = [[
+        local function apply(cfg)
+            _G.bar = cfg
+        end
+
+        _G.foo = 42
+        _G.bar = nil
+
+        return {
+            validate = function() end,
+            apply = apply,
+            stop = function() end,
+        }
+    ]]
+    local verify = function()
+        local config = require('config')
+        t.assert_equals(_G.foo, 42)
+        local roles_cfg = config:get('roles_cfg')
+        t.assert_equals(roles_cfg['one'], 12345)
+        t.assert_equals(roles_cfg['one'], _G.bar)
+    end
+
+    helpers.success_case(g, {
+        roles = {one = one},
+        options = {
+            ['roles_cfg'] = {one = 12345},
+            ['roles'] = {'one'}
+        },
+        verify = verify,
+    })
+end
+
+-- Make sure the role is loaded only once during each run.
+g.test_role_repeat_success = function(g)
+    local one = [[
+        local function apply(cfg)
+            _G.foo = _G.foo * cfg
+        end
+
+        _G.foo = 42
+
+        return {
+            validate = function() end,
+            apply = apply,
+            stop = function() end,
+        }
+    ]]
+    local verify = function()
+        local config = require('config')
+        local roles_cfg = config:get('roles_cfg')
+        t.assert_equals(roles_cfg['one'], 3)
+        t.assert_equals(_G.foo, 42 * 3)
+    end
+
+    helpers.success_case(g, {
+        roles = {one = one},
+        options = {
+            ['roles_cfg'] = {one = 3},
+            ['roles'] = {'one', 'one', 'one', 'one', 'one', 'one', 'one'}
+        },
+        verify = verify,
+    })
+end
+
+-- Make sure all roles are loaded correctly and in correct order.
+g.test_multiple_role_success = function(g)
+    local one = [[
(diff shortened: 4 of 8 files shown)


changelogs/unreleased/gh-9078-roles.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* Added initial support of roles (gh-9078).
```
p7nov, line 3:
```suggestion
* Introduced the initial support for roles – programs that run when
  a configuration is loaded or reloaded (gh-9078).```

**Review by p7nov — APPROVED** [review]

File: changelogs/unreleased/gh-9078-roles.md (hunk 1 above)
Comment (Totktonada) [reply changelogs/unreleased/gh-9078-roles.md:3]: I think that we shouldn't use non-ascii characters in texts for commit messages, issues, pulls requests, releases. I'm about 'en dash' here.

File: changelogs/unreleased/gh-9078-roles.md [hunk 2]
```diff
@@ -0,0 +1,3 @@
+## feature/config
```
Totktonada, line 1: I think it's time to start writting documentation requests, at least for new features.

As for me, it would also make the API of the user defined code more clear.

I would like to discuss an error convention here. `cartridge` uses `nil, err` everywhere, but I doubt that we should inherit it. At least, we should consider a raising error as a legal error reporting way. OTOH, support of `nil, err` and/or `false, err` may be ocnvenient for porting cartridge role code to the new framework. I don't mind it.

Anyway, let's write down the API description somewhere (preferably in the documentation request).

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,75 @@
+local log = require('internal.config.utils.log')
+
+local last_loaded = {}
+
+local function apply(_config)
+    log.verbose('roles.apply: do nothing')
+end
+
+local function post_apply(config)
```
Totktonada, line 9: `app` is loaded in `post_apply`, because we want the app code working in the `ready` config state with configuration accessible from `config:get()`.

Is the motivation for using `post_apply` the same for role's code? Should we consider a config as successfully applied before we're run role's code?

I have no objections, but I want to understand all the hidden details.

test/config-luatest/helpers.lua
```diff
@@ -185,31 +197,43 @@ end
...
+--
 -- * opts.verify_2
 --
 --   Verify test invariants after config:reload().
 local function reload_success_case(g, opts)
+    assert(opts.options)
+    local roles_2 = opts.roles_2
     local script_2 = opts.script_2
-    local options = assert(opts.options)
+    local options_2 = opts.options_2
```
Totktonada, line 227: It is also added in #9088. I guess that I'll push it first. So, this comment is a reminder to rebase afterwards.

Totktonada: PR #9088 lands to `master`.

changelogs/unreleased/gh-9078-roles.md
ImeevMA: Fixed. Used `-` instead of `–`.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,75 @@
...
+    -- Run roles.
+    local roles_cfg = configdata:get('roles_cfg', {use_default = true}) or {}
+    local loaded = {}
+    for _, role_name in pairs(roles) do
+        local role_cfg = roles_cfg[role_name]
+        -- All roles must be described in roles_cfg, even if the role does not
+        -- require a cfg.
+        if role_cfg == nil then
+            error('No config for role ' .. role_name, 0)
+        end
```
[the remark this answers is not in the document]
ImeevMA: Fixed.

changelogs/unreleased/gh-9078-roles.md
ImeevMA: Moved part of commit-message to doc-request. As for error convention, haven't thought about it yet.

src/box/lua/config/applier/roles.lua
ImeevMA: I thought maybe the roles should also have access to the config. However, they have their own config, so I removed post_apply.

test/config-luatest/helpers.lua
ImeevMA: Rebased.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
+local log = require('internal.config.utils.log')
+
+local last_loaded = {}
+
+local function stop_roles(roles)
+    for role_name, role in pairs(last_loaded) do
+        if roles == nil or not roles[role_name] then
+            log.verbose('roles.post_apply: stop role ' .. role_name)
```
Totktonada, line 8: Here and below: post_apply -> apply.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
+local log = require('internal.config.utils.log')
+
+local last_loaded = {}
+
+local function stop_roles(roles)
```
Totktonada, line 5: It is not obvious that the `roles` arguments means roles to skip, not to stop.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
...
+    local roles_cfg = configdata:get('roles_cfg', {use_default = true}) or {}
+    local loaded = {}
+    for role_name in pairs(roles) do
+        local role = last_loaded[role_name]
+        if not role then
+            role = require(role_name)
+            local methods = {'init', 'validate_config', 'apply_config', 'stop'}
+            for _, method_name in pairs(methods) do
+                if type(role[method_name]) ~= 'function' then
+                    local err = 'Role %s does not contain method %s'
```
Totktonada, line 38: Nit: a method assumes that it receives its own object as a first argument. I would call them functions.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
...
+                end
+            end
+            log.verbose('roles.post_apply: initialize role ' .. role_name)
+            role.init()
+        end
+        loaded[role_name] = role
+        local role_cfg = roles_cfg[role_name]
+        if not role.validate_config(role_cfg) then
+            error('Wrong config for role ' .. role_name, 0)
+        end
```
Totktonada, line 49: It seems, if the function returns `nil, err`, the error will be ignored. However, it worth to include the error message into our own error message. (Don't forget about `tostring()` or, better, use `('<...>'):format()`.)

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
...
+        log.verbose('roles.post_apply: apply config for role ' .. role_name)
+        role.apply_config(role_cfg)
```
Totktonada, line 51: Are the contracts for `validate_config()` and `apply_config()` different? The first is checked for return value, the second doesn't. However, by construction, should have ability to report an error (either by raising it or by returning).

changelogs/unreleased/gh-9078-roles.md
Totktonada: I think that we should at least write down the current calling convention, including an error convention. An author of a role should know what he/she should use to report a success or a failure.

test/config-luatest/roles_test.lua
```diff
@@ -0,0 +1,443 @@
...
+        options = {
+            ['roles_cfg'] = {one = 1},
+            ['roles'] = {'one'}
+        },
+        exp_err = 'Role one does not contain method stop'
+    })
+end
+
+-- Verify that the error is raised by config:reload() and the same
+-- error appears in the alerts.
```
Totktonada, line 410: I appreciate your work on the test cases. Thank you!

I would add cases, when `apply_config` fails and when `stop` fails.

src/box/lua/config/applier/roles.lua
Totktonada, line 49: The idea of separation of the `validate` and the `apply` phases is that if some `validateconfig` fails we don't go to call `applyconfig` functions. So, we should call all the `validate_config` functions first and only then (if they're succeeded) call `apply_config` ones.

(Maybe it worth to call all the `init` functions before all the `validate_config` ones too.)

changelogs/unreleased/gh-9078-roles.md
Totktonada: My personal opinion how the API of the role should look, at least for now.

```lua
local M = {}

-- Validates configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the validation fail.
function M.validate(cfg)
    -- <...>
end

-- Applies the given configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the given configuration can't be appied.
function M.apply(cfg)
    -- <...>
end

-- Stops the role.
--
-- Called on configuration reload if the role was enabled before
-- and removed now from the list of roles of the given instance.
--
-- Should cancel all background fibers and clean up hold
-- resources.
--
-- Must raise an error if this action can't be performed.
function M.stop()
    -- <...>
end

return M
```

Key points:

* No `init()`, because I see no difference from just placing initialization code into the module scope.
* `validate` and `apply` without `_config` suffix, because I see no much sense in the suffix.
* All the function raise an error at failure. I looked at several cartridge roles (for vshard, crud, metrics, expirationd) and all ones, where `validate_config` is present, raise an error from `validate_config` instead of returning `nil, err` or `false, err`.

The key idea is to start from a clean API and add cartridge compatibility only where it is actually needed.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,62 @@
...
+    if role_names == nil or next(role_names) == nil then
+        stop_roles()
+        return
+    end
+
+    -- Remove duplicates.
+    local roles = {}
+    for _, role_name in pairs(role_names) do
+        roles[role_name] = true
+    end
```
Totktonada, line 26: I looked on [cartridge's approach](https://github.com/tarantool/cartridge/blob/986e8065ae83747b184fa7655f1f6e415645887b/cartridge/roles.lua#L48) to define the order of loading of the given roles and I liked it.

Previously I said that it seems OK to load roles in arbitrary order, while there are no dependencies between them. I changed my mind. I think that the order should be strictly defined to prevent subtle problems from appear.

My proposal for the resorting stage is the following.

1. Collect all the role names and start to process them one-by-one.
2. `require` the given current role R.
3. If it has no dependencies and was not visited by the algorithm before, add it into resulting list `RS`.
4. For each dependency in the given order check that it is present in the initial list of the roles. If it is not so, give an error. If the check passed, process the dependency role name using this algorith recursively.

Or, the same, as a code.

```lua
#!/usr/bin/env tarantool

local fun = require('fun')
local json = require('json')

-- bar <- foo <- baz
local function requirex(x)
    if x == 'foo' then
        return {
            dependencies = {'baz'},
        }
    elseif x == 'bar' then
        return {
            dependencies = {'foo'},
        }
    elseif x == 'baz' then
        return {}
    else
        assert(false)
    end
end

local function resort_roles(src)
    local resorted = {}

    -- Needed to detect roles that are not listed in the
    -- instance's role list.
    local all_ht = fun.iter(src):map(function(x) return x, true end):tomap()

    -- Needed to detect circular dependencies.
    local to_be_added = {}

    -- To skip already added roles.
    local added_ht = {}

    local function add_role(role_name)
        if added_ht[role_name] then
            return
        end

        local role = requirex(role_name)
        to_be_added[role_name] = true

        for _, dep in ipairs(role.dependencies or {}) do
            -- Detect a role that is not in the list of instance's
            -- roles.
            if not all_ht[dep] then
                error(('Role %q requires role %q, but the latter is not in ' ..
                    'the list of roles of the instance'):format(role_name, dep),
                    0)
            end

            -- Detect a circular depdendency.
            if to_be_added[dep] and role_name == dep then
                error(('Circular dependency: role %q depends on itself'):format(
                    role_name), 0)
            end
            if to_be_added[dep] and role_name ~= dep then
                error(('Circular dependency: roles %q and %q depend on each ' ..
                    'other'):format(role_name, dep), 0)
            end

            -- Go into the recursion: add the dependency.
            add_role(dep)
        end

        -- Add the role itself and track it in the temporary
        -- hashtables.
        to_be_added[role_name] = nil
        added_ht[role_name] = true
        table.insert(resorted, role_name)
    end

    -- Keep the order, where the dependency tree doesn't obligate
    -- us to change it.
    for _, role_name in ipairs(src) do
        add_role(role_name)
    end

    return resorted
end

print(json.encode(resort_roles({'foo', 'bar', 'baz', 'foo'})))
-- ["bar", "foo", "bar"]
```

While we have no dependencies, I think, it is OK to guarantee the order of `require`/`validate`/`apply` is the same as written in the config (except excluding duplicates).

Totktonada: And, while we're here, I would (like cartridge) define some specific order between `apply` and `stop`.

For example, we can perform all stops before all starts. The stops should also be performed in some guaranteed order. I propose the reverse of the starting order (reverse of `resort_order`'s result).

As another example, cartridge forms a list of all known roles (from previous and from the next configuration) and perform start/stop in the order of the list (depending on whether the role is present in the new config). I think it is error-prone to stop dependencies before a role that depends on them, but maybe there is something useful in the overall approach.

Anyway, while we have no dependencies, we still should define the order between starts and stops. I propose to do all the stops in the reversed order and then all starts in the direct order.

changelogs/unreleased/gh-9078-roles.md
ImeevMA: Rewrote doc-request.

src/box/lua/config/applier/roles.lua
ImeevMA: Fixed.

src/box/lua/config/applier/roles.lua
ImeevMA: Fixed.

src/box/lua/config/applier/roles.lua
ImeevMA: Fixed.

src/box/lua/config/applier/roles.lua
ImeevMA: Done.

src/box/lua/config/applier/roles.lua
ImeevMA: Now they should throw an error if it happens.

test/config-luatest/roles_test.lua
ImeevMA: Done.

src/box/lua/config/applier/roles.lua
ImeevMA: Now there is no init(), but now there are three phases: loading unloaded roles; validate() for all roles; apply() for all roles.

src/box/lua/config/applier/roles.lua
ImeevMA: Since there is no dependency, I used a slightly simpler approach. In addition, roles are now loaded, validated, and applied in the same order in which they were given. Stopped in reversed order.

changelogs/unreleased/gh-9078-roles.md
Totktonada: Thank you!

I have a few small comments regarding the texts.

1. Please, update the pull request description.
2. > In this regard, there is no strict order in which roles are performed.

   Seems obsoleted.
3. The documentation request doesn't mention arguments of role defined functions anywhere. (How about including a role example in the form of the snippet above?)
4. > On each run, the role calls validate() and then apply().

   The functions are called, but not by the role.
5. > Roles can be defined for each instance.

   May be interpreted as 'only on the instance scope', I would add 'for a replicaset, a group or for the whole cluster'.
6. There is no description of the new instance/cluster schema keys: `roles` and `roles_cfg`.

src/box/lua/config/applier/roles.lua
```diff
@@ -0,0 +1,87 @@
...
+    -- Stop removed roles.
+    stop_roles(roles)
+
+    -- Run roles.
+    local roles_cfg = configdata:get('roles_cfg', {use_default = true}) or {}
+    local loaded = {}
+    local loaded_names_ordered = {}
+
+    -- Load roles.
+    for _, role_name in pairs(roles_ordered) do
```
Totktonada, line 46: Here and below: `ipairs()` should be used instead `pairs()` to guarantee the order.

Totktonada (approved):
Thank you for the patch!

I'm OK in general, but, please consider a few remaining small comments above. (Let's add the `full-ci` label right after the fix.)

src/box/lua/config/applier/roles.lua
ImeevMA: Fixed.

changelogs/unreleased/gh-9078-roles.md
ImeevMA: Thanks, fixed.

PR #9088, opened 2023-09-03
test/config: verify that user/role is not deleted

Lord-KA:
When the configuration changes and the instance is reloaded with it,
some roles or users may be removed from the config. In such case,
it would be destructive to delete/disable them on the instance, so
this test checks that all users and roles removed in config stay
on the instance.
    
Part of #8967

### DIFF
--- a/test/config-luatest/credentials_applier_test.lua
+++ b/test/config-luatest/credentials_applier_test.lua
@@ -2,8 +2,9 @@ local json = require('json')
 local it = require('test.interactive_tarantool')
 local t = require('luatest')
 local treegen = require('test.treegen')
+local helpers = require('test.config-luatest.helpers')
 
-local g = t.group()
+local g = helpers.group()
 
 local internal = require('internal.config.applier.credentials')._internal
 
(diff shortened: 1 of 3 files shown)


test/config-luatest/helpers.lua
```diff
@@ -196,20 +196,26 @@ end
 --   A new script to write into the main.lua file before
 --   config:reload().
 --
+-- * opts.config_2
+--
+--   A new config to use for the config:reload(). It is optional,
+--   if not provided opts.config is used instead.
```
Totktonada, line 202: Typo: `opts.config_2` -> `opts.options_2`, `opts.config` -> `opts.options`.

test/config-luatest/credentials_applier_test.lua
```diff
@@ -560,3 +561,60 @@ g.test_set_password = function(g)
...
+                        roles = { 'super' }
+                    },
+                }
+            }
+        },
+        verify_2 = function()
+            local ok, err = pcall(box.schema.user.info, 'myuser')
+            t.assert(ok, err)
+            ok, err = pcall(box.schema.role.info, 'myrole')
+            t.assert(ok, err)
```
Totktonada, line 617: It looks worthful to verify that permissions that are granted on the first config apply are still there.

Totktonada (approved):
Looks OK except a few minor points. I'm ready to push after you'll resolve them.

test/config-luatest/helpers.lua
```diff
@@ -196,20 +196,26 @@ end
...
+--
 -- * opts.verify_2
 --
 --   Verify test invariants after config:reload().
 local function reload_success_case(g, opts)
     local script_2 = opts.script_2
     local options = assert(opts.options)
     local verify_2 = assert(opts.verify_2)
+    local options_2 = opts.options_2 or options
```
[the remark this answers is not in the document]
Totktonada: Regarding 'we're asking for all other parameters to be explicitly defined': I just started with more strict validation and then relaxed it where it is convenient for writting the testing code (for example, f1a09989e7cb483ada64384a3e1842712ae1e4f5).

As for me, it is OK to have optional parameters here if it doesn't lead to some silently incorrect configuration.

It seems natual to pick up old options if newer ones are not provided. Say, if we want to update Lua scripts and reload. Or if we want to update SSL keys (that reside in separate files). Or just ensure that the configuration is idempotent: startup and reload give the same resulting state.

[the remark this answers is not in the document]
Lord-KA: I agree with @Totktonada, lets leave it as is.

test/config-luatest/credentials_applier_test.lua
Lord-KA: Done. I've also moved verification function to a local var for convenience.

ochaplashkin (approved):
LGTM :rocket:

PR #9091, opened 2023-09-04
config: revoke privs for default users and roles

Lord-KA:
All user-defined users and roles are not being removed and their privileges are not being revoked when this user or role is removed from config. This is done to prevent extreme repercussions of misconfiguration, e.g. empty config is provided to cluster and it breaks up.

Default users and roles are not supposed to be changed, so this rule does not apply to them. Now all of non-default privileges will be revoked if such user or role is removed from config.

Default users:
* guest
* admin

Default roles:
* super
* public
* replication

Part of #8967

### DIFF
--- a/changelogs/unreleased/gh-8967-creds-restore-defaults-for-default-user.md
+++ b/changelogs/unreleased/gh-8967-creds-restore-defaults-for-default-user.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Non-default privileges are now revoked from default users and roles
+  when they are removed from the config (gh-8967).
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -493,6 +493,35 @@ local function apply(config)
         return
     end
 
+    -- Tarantool has the following roles and users present by default on every
+    -- instance:
+    --
+    -- Default roles:
+    -- * super
+    -- * public
+    -- * replication
+    --
+    -- Default users:
+    -- * guest
+    -- * admin
+    --
+    -- These roles and users have according privileges pre-granted by design.
+    -- Credentials applier adds such privileges with `priviliges_add_default()`
+    -- when syncing. So, for the excessive (non-default) privs to be removed,
+    -- these roles and users must be present inside configuration at least in a
+    -- form of an empty table. Otherwise, the privileges will be left unchanged,
+    -- similar to all used-defined roles and users.
+
+    credentials.roles = credentials.roles or {}
+    credentials.roles['super'] = credentials.roles['super'] or {}
+    credentials.roles['public'] = credentials.roles['public'] or {}
+    credentials.roles['replication'] = credentials.roles['replication'] or {}
+
+    credentials.users = credentials.users or {}
+    credentials.users['guest'] = credentials.users['guest'] or {}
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-8967-creds-restore-defaults-for-default-user.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Implemented revoking of non-default privileges for default
+  users and roles (gh-8967).
```
p7nov, line 4:
```suggestion
* Non-default privileges are now revoked from default users and roles
  when they are removed from the configuration (gh-8967).
```

p7nov (approved):

src/box/lua/config/applier/credentials.lua
```diff
@@ -493,6 +493,28 @@ local function apply(config)
...
+    -- for roles and users that are present by default on every instance:
+    -- - - super
+    --   - public
+    --   - replication
+    -- - - guest
+    --   - admin
+    --
+    -- It is done to revoke all non-default privileges for this roles and
+    -- users if they are not present in config. Otherwise, the privileges
+    -- will be synced as usual.
```
Totktonada, line 506: Nits:

1. 'Should be enrihed with empty config' -- I would add somewhere a note that it is later updated with defaults in `privileges_add_defaults()`.
2. 'Otherwise, the privileges will be synced as usual' -- does it mean 'as for a user/role that doesn't exist by default'? I didn't get it initially. I propose to say 'Otherwise, the privileges will be left without changes. It is exactly how it works for all user defined users and roles', or something similar.

Totktonada (approved):

changelogs/unreleased/gh-8967-creds-restore-defaults-for-default-user.md
Lord-KA: Applied.

changelogs/unreleased/gh-8967-creds-restore-defaults-for-default-user.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
```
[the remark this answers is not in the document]
Lord-KA: Done:
>     NO_DOC=documentation request will be filed manually for the whole credentials

src/box/lua/config/applier/credentials.lua
Lord-KA: Rephrased the whole comment:
```lua
      -- Tarantool has the following roles and users present by default on every
      -- instance:
      --
      -- Default roles:
      -- * super
      -- * public
      -- * replication
      --
      -- Default users:
      -- * guest
      -- * admin                                                                                                                       
      --                                                                                                                               
      -- These roles and users have according privileges pre-granted by design.                                                        
      -- Credentials applier adds such privileges with `priviliges_add_default()`                                                      
      -- when syncing. So, for the excessive (non-default) privs to be removed,                                                        
      -- these roles and users must be present inside configuration at least in a                                                      
      -- form of an empty table. Otherwise, the privileges will be left unchanged,                                                     
      -- similar to all used-defined roles and users.                                                                                  
```

Totktonada (approved):

PR #9113, opened 2023-09-07
box: drop debug log on tuple new/delete

nshy:
They are rather noisy. Also delete debug log on arena creation. These two make sense only with each other.

Part of #7327


src/box/vy_stmt.c
```diff
@@ -100,7 +100,6 @@ vy_tuple_delete(struct tuple_format *format, struct tuple *tuple)
 {
 	struct vy_stmt_env *env = format->engine;
 	size_t size = tuple_size(tuple);
-	say_debug("%s(%p)", __func__, tuple);
```
locker, line 103: Missed one more in `vy_stmt_new`.

locker (changes requested):

src/box/vy_stmt.c
nshy: Fixed.

locker (approved):

locker:
Tests fail. Pls fix.

ylobankov (approved):

locker (approved):

PR #9119, opened 2023-09-09
config: postpone creds applier when in RO till RW

Lord-KA:
Before this patch credentials applier used to just skip if Tarantool was in Read Only mode. Now it starts a fiber that waits for instance to be switched to Read Write mode and then applies in the background.

Part of #8967

NO_DOC=documentation request will be filed manually for the whole
       credentials

### DIFF
--- a/changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
+++ b/changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
@@ -1,0 +1,5 @@
+## feature/config
+
+* On read-only instances, Tarantool now synchronizes credentials
+  in the background when switching to read-write mode instead of
+  skipping the synchronization (gh-8967).
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -1,5 +1,6 @@
 local log = require('internal.config.utils.log')
 local digest = require('digest')
+local fiber = require('fiber')
 
 --[[
 This intermediate representation is a formatted set of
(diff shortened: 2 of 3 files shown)


src/box/lua/config/applier/credentials.lua
```diff
@@ -526,6 +513,37 @@ local function apply(config)
...
+
+    -- If Tarantool is already in Read Write mode, credentials are still
+    -- applied in the fiber to avoid possible concurrency issues, but the
+    -- main thread is blocked by the `join()`.
+    if not box.info.ro then
+        local ok, err = fiber_wait_rw:join()
+        if not ok then
+            error(err)
+        end
+    end
```
Totktonada, line 544: I guess we can't mark a fiber as joinable and refuse to join it. All there fibers will wait infinitely for a join call.

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
```diff
@@ -0,0 +1,5 @@
+## feature/config
+
+* If Tarantool is in Read Only mode, credentials are now synchronised
+  in the background when switched to Read Write mode, instead of being
+  skipped (gh-8967).
```
p7nov, line 5:
```suggestion
* On read-only instances, Tarantool now synchronizes credentials
  in the background when switching to read-write mode instead of
  skipping the synchronization (gh-8967).
```

p7nov (approved):

src/box/lua/config/applier/credentials.lua
Lord-KA: You are right, let's move `srt_joinable()` to just above the `join()`.

src/box/lua/config/applier/credentials.lua
```diff
@@ -526,6 +513,37 @@ local function apply(config)
...
+
+    -- Note that only one fiber exists at a time.
+    if fiber_wait_rw == nil or fiber_wait_rw:status() == 'dead' then
+        fiber_wait_rw = fiber.new(wait_rw, config)
+        fiber_wait_rw:set_joinable(true)
+    end
+
+    -- If Tarantool is already in Read Write mode, credentials are still
+    -- applied in the fiber to avoid possible concurrency issues, but the
+    -- main thread is blocked by the `join()`.
```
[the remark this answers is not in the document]
Lord-KA: Now, with `set_joinable()` inside if clause, the comment is legit. We block the main thread when Tarantool is already in RW.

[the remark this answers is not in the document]
Totktonada: The fiber that processes the configuration applying is blocked, that's right. I guess that it is a kind of the main fiber in context of startup. However, if config:reload() is invoked, it is the fiber that processes the call. If an update comes from etcd, this code will be executed in a fiber that watches for etcd updates.

But! I want to highlight that TX **thread** is not blocked. Other fibers can work within the thread, while this fiber waits for `:join()`.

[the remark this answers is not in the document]
Lord-KA: Oh, right. Reworded:
```
      -- If Tarantool is already in Read Write mode, credentials are still
      -- applied in the fiber to avoid possible concurrency issues, but the
      -- main applier fiber is blocked by the `join()`.
```

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
Lord-KA: Applied.

src/box/lua/config/applier/credentials.lua
Lord-KA: Done.

Totktonada (approved):

PR #9129, opened 2023-09-12
cbus: introduce lcpipe - light cpipe

### LINKED ISSUE #9128: Communication between third party thread and TX thread mechanism
Currently tarantool have not builtin mechanism to comunicate between any third party thread and TX thread. This mechanism will be really useful for implementing stored procedures on c/rust languages.

There is a xtm library for this communication, but it has some cons:
* it is not native for tarantool, and require vendor c++ lib in the project (this may be not trivial for non c/c++ projects)
* it has a complex API
* eventfd is a linux only syscall (looks like there is a bad performance on mac)
* as assume later, I suggest using a cbus, for performance it may look better, cause conceptually have the number of sync barriers is O(number of threads) and not O(number of messages)

I solved that i suggest - reuse tarantool cbus and export cbus interfaces. Currently there is one problem with using cbus as is - it communicates coord's not threads. The solution to this problem is a new type of CPipe - LCPipe (light c pipe) which is a means of communication between cord thread (TX thread in most cases) and the third-party stream.

godzie44:
Introduced a new type of cbus pipe - lcpipe. The current pipe in the cbus - cpipe, has a number of limitations, first of all - the cpipe cannot be used from the 3rd party threads, cpipe only works as a channel between two cords. That why lcpipe is needed. Its main responsibility - create channel between any thread and tarantool cord.

Internally lcpipe is a cpipe, but:
- on flush triggers removed, cause triggers use thread-local mem-pool, this is not possible on a third party thread
- producer event loop removed, cause there is no libev event loop in third party thread

Also, lcpipe interface is exported to the outside world.

Close #9128

NO_DOC=core feature


ochaplashkin (approved):
LGTM :rocket:

xuniq (approved):

unera:
I think we should apply the patch, but it have to be reviewed by @alyapunov

src/lib/core/cbus.c
```diff
@@ -177,6 +187,145 @@ cpipe_destroy(struct cpipe *pipe)
...
+
+	return pipe;
+}
+
+/**
+ * Flush all staged messages into consumer output and wake up it after.
+ */
+static void
+lcpipe_flush(struct lcpipe *pipe)
+{
```
Gerold103, line 210: This function is a line to line copy-paste of `cpipe_flush_cb`.

src/lib/core/cbus.c
```diff
@@ -177,6 +187,145 @@ cpipe_destroy(struct cpipe *pipe)
...
+
+void
+lcpipe_flush_input(struct lcpipe *pipe)
+{
+	/** Flush may be called with no input. */
+	if (pipe->n_input > 0) {
+		lcpipe_flush(pipe);
+	}
+}
+
```
Gerold103, line 254: `lcpipe_flush` already checks `n_input` and doesn't do anything when it is zero. This check is not needed. Which makes `lcpipe_flush_input` == `lcpipe_flush`, which means you can rename `lcpipe_flush` to `lcpipe_flush_input` and delete this code block.

src/lib/core/cbus.c
```diff
@@ -177,6 +187,145 @@ cpipe_destroy(struct cpipe *pipe)
...
+static inline void
+lcpipe_insert_input(struct lcpipe *pipe, struct cmsg *msg)
+{
+	assert(msg->hop->pipe == NULL);
+	stailq_add_tail_entry(&pipe->input, msg, fifo);
+	pipe->n_input++;
+	if (pipe->n_input >= pipe->max_input) {
+		lcpipe_flush(pipe);
+	}
+}
```
Gerold103, line 267: This code is identical to `cpipe_push_input` with the only difference that you call here `lcpipe_flush` and the former calls `ev_invoke(pipe->producer, &pipe->flush_input, EV_CUSTOM);`. But if you look at what `ev_invoke` does, you will find out, that it calls `cpipe_flush_cb` here which is the same as your `lcpipe_flush`.

src/lib/core/cbus.c
```diff
@@ -177,6 +187,145 @@ cpipe_destroy(struct cpipe *pipe)
...
+
+void
+lcpipe_push(struct lcpipe *pipe, struct cmsg *msg)
+{
+	lcpipe_insert_input(pipe, msg);
+	assert(pipe->n_input < pipe->max_input);
+}
```
Gerold103, line 274: You don't really need `lcpipe_insert_input` if `lcpipe_push` does the same. Just rename `lcpipe_insert_input` -> `lcpipe_push` and drop this code block.

src/lib/core/cbus.c
```diff
@@ -177,6 +187,145 @@ cpipe_destroy(struct cpipe *pipe)
...
+lcpipe_push_now(struct lcpipe *pipe, struct cmsg *msg)
+{
+	lcpipe_insert_input(pipe, msg);
+	assert(pipe->n_input < pipe->max_input);
+	lcpipe_flush(pipe);
+}
+
+void
+lcpipe_delete(struct lcpipe *pipe)
+{
```
Gerold103, line 286: This looks identical to `cpipe_destroy`.

src/lib/core/cbus.h
```diff
@@ -466,6 +488,66 @@ cbus_unpair(struct cpipe *dest_pipe, struct cpipe *src_pipe,
...
+	 * latency, while still keeping the bus mutex cold enough).
+	 */
+	int max_input;
+	/**
+	 * The cbus endpoint at the destination cord to handle
+	 * flushed messages.
+	 */
+	struct cbus_endpoint *endpoint;
+};
+
```
Gerold103, line 514: Given that this structure itself and all its methods are a complete copy-paste of some parts of `cpipe`, I would suggest to find a way to reuse the code.

One suggestion could be that you do the C-inheritance. So this struct must be added into `struct cpipe` as `base` member instead of those duplicated members here. Then you make the relevant `cpipe` methods call the `lcpipe` methods on its `&base` member pointer.

If my guess about `ev_invoke` behaviour above is correct, then this should be a simple change which will eliminate a lot of duplicated code.

src/lib/core/cbus.h
```diff
@@ -466,6 +488,66 @@ cbus_unpair(struct cpipe *dest_pipe, struct cpipe *src_pipe,
...
+ **/
+void
+lcpipe_delete(struct lcpipe *pipe);
+
+/**
+ * Flush all staged messages into the pipe and eventually to the
+ * consumer.
+ */
+void
+lcpipe_flush_input(struct lcpipe *pipe);
```
Gerold103, line 535: This is questionable whether we should expose those lcpipe and cpipe methods directly with these names instead of using `tnt_` or `box_` or some other prefix common for Tarantool exports. But I will leave that to the later core team reviewers who so gently gave me the place of the first reviewer. At first I suggest to eliminate the duplicated code.

changelogs/unreleased/cbus-lcpipe.md
```diff
@@ -0,0 +1,4 @@
+## feature/build
```
Gerold103, line 1: `on_flush` triggers seem unused in all the non-test code. You can delete them from `cpipe` too.

Gerold103 (changes requested):
Thanks for the patch! I have already seen this before and was hoping the core team would have a look at the code which they own, but we can also do some things right away before those guys join us.

extra/exports
```diff
@@ -30,10 +30,10 @@ box_decimal_is_neg
 box_decimal_ln
 box_decimal_log10
 box_decimal_minus
-box_decimal_mp_sizeof
-box_decimal_mp_encode
 box_decimal_mp_decode
 box_decimal_mp_decode_data
+box_decimal_mp_encode
+box_decimal_mp_sizeof
```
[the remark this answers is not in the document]
godzie44: done

src/lib/core/cbus.c
godzie44: it's true, im move all duplicated code in `move_messages` function

src/lib/core/cbus.c
godzie44: thanks, done

src/lib/core/cbus.c
godzie44: There is a fundamental difference between cpipe_push_input and lcpipe_insert_input: the second can be called in any context, but the first can only be called in cord.
There are some differences in the code:
- cpipe_push_input checks for the presence of a loop (we need cord for this) in the producer side: ```assert(loop() == pipe->producer);```
- lcpipe_insert_input calls lcpipe_flush for which only lcpipe itself is needed, it cannot call cpipe_flush_cb because it requires ev_loop and ev_async which are not available outside the cord context

src/lib/core/cbus.c
godzie44: thanks, done

src/lib/core/cbus.c
godzie44: Yes it looks similar, but lcpipe_delete:
- calls `free` because lcpipe is always allocated to the heap
- cpipe_destroy requires that the producer to work inside a cord (lcpipe_delete can be called from any thread)
- cpipe_destroy call trigger_destroy (lcpipe doesnt support triggers)

src/lib/core/cbus.h
godzie44: Done, but I couldn’t ifx code duplication this way (except in cbus.h), because wherever cpipe is used, both its base part and other fields are required

src/lib/core/cbus.h
godzie44: Yes, you are right, maybe we should make this in a separate PR? Looks like this not about lcpipe, but about just cpipe refactoring

changelogs/unreleased/cbus-lcpipe.md
godzie44: Yes, you are right, maybe we should make this in a separate PR? Looks like this not about lcpipe, but about just cpipe refactoring

extra/exports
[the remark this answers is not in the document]
Gerold103: Nope, there is still just one commit.

src/lib/core/cbus.h
Gerold103: The main idea was exactly to fix the duplication in cbus.c. Please, read a bit about C struct inheritance and check some examples maybe in how `struct port`. Applied to cbus code this is what I've meant:

```Diff
diff --git a/src/lib/core/cbus.c b/src/lib/core/cbus.c
index 2ada0eef9..0cc9b17e2 100644
--- a/src/lib/core/cbus.c
+++ b/src/lib/core/cbus.c
@@ -139,36 +139,33 @@ cbus_endpoint_poison_f(struct cmsg *msg)
 	free(msg);
 }
 
-void
-cpipe_destroy(struct cpipe *pipe)
+static void
+lcpipe_destroy(struct lcpipe *pipe)
 {
 	/*
 	 * The thread should not be canceled while mutex is locked.
 	 * And everything else must be protected for consistency.
-	*/
+	 */
 	int old_cancel_state;
 	tt_pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old_cancel_state);
 
-	ev_async_stop(pipe->producer, &pipe->flush_input);
-
 	static const struct cmsg_hop route[1] = {
 		{cbus_endpoint_poison_f, NULL}
 	};
-	trigger_destroy(&pipe->on_flush);
 
-	struct cbus_endpoint *endpoint = pipe->base.endpoint;
-	struct cmsg_poison *poison = malloc(sizeof(struct cmsg_poison));
+	struct cbus_endpoint *endpoint = pipe->endpoint;
+	struct cmsg_poison *poison = xmalloc(sizeof(*poison));
 	cmsg_init(&poison->msg, route);
-	poison->endpoint = pipe->base.endpoint;
+	poison->endpoint = pipe->endpoint;
 	/*
-	 * Avoid the general purpose cpipe_push_input() since
+	 * Avoid the general purpose lcpipe_push() since
 	 * we want to control the way the poison message is
 	 * delivered.
 	 */
 	tt_pthread_mutex_lock(&endpoint->mutex);
 	/* Flush input */
-	stailq_concat(&endpoint->output, &pipe->base.input);
-	pipe->base.n_input = 0;
+	stailq_concat(&endpoint->output, &pipe->input);
+	pipe->n_input = 0;
 	/* Add the pipe shutdown message as the last one. */
 	stailq_add_tail_entry(&endpoint->output, poison, msg.fifo);
 	/* Count statistics */
@@ -183,7 +180,21 @@ cpipe_destroy(struct cpipe *pipe)
 	tt_pthread_mutex_unlock(&endpoint->mutex);
 
 	tt_pthread_setcancelstate(old_cancel_state, NULL);
+}
+
+void
+lcpipe_delete(struct lcpipe *pipe)
+{
+	lcpipe_destroy(pipe);
+	free(pipe);
+}
 
+void
+cpipe_destroy(struct cpipe *pipe)
+{
+	ev_async_stop(pipe->producer, &pipe->flush_input);
+	trigger_destroy(&pipe->on_flush);
+	lcpipe_destroy(&pipe->base);
 	TRASH(pipe);
 }
 
@@ -275,50 +286,7 @@ lcpipe_push_now(struct lcpipe *pipe, struct cmsg *msg)
 	lcpipe_flush_input(pipe);
 }
 
-void
-lcpipe_delete(struct lcpipe *pipe)
-{
-	/*
-	 * The thread should not be canceled while mutex is locked.
-	 * And everything else must be protected for consistency.
-	 */
-	int old_cancel_state;
-	tt_pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old_cancel_state);
-
-	static const struct cmsg_hop route[1] = {
-		{cbus_endpoint_poison_f, NULL}
-	};
-
-	struct cbus_endpoint *endpoint = pipe->base.endpoint;
-	struct cmsg_poison *poison = xmalloc(sizeof(*poison));
-	cmsg_init(&poison->msg, route);
-	poison->endpoint = pipe->base.endpoint;
-	/*
-	 * Avoid the general purpose lcpipe_push() since
-	 * we want to control the way the poison message is
-	 * delivered.
-	 */
-	tt_pthread_mutex_lock(&endpoint->mutex);
-	/* Flush input */
-	stailq_concat(&endpoint->output, &pipe->base.input);
-	pipe->base.n_input = 0;
-	/* Add the pipe shutdown message as the last one. */
-	stailq_add_tail_entry(&endpoint->output, poison, msg.fifo);
-	/* Count statistics */
-	rmean_collect(cbus.stats, CBUS_STAT_EVENTS, 1);
-	/*
-	 * Keep the lock for the duration of ev_async_send():
-	 * this will avoid a race condition between
-	 * ev_async_send() and execution of the poison
-	 * message, after which the endpoint may disappear.
-	 */
-	ev_async_send(endpoint->consumer, &endpoint->async);
-	tt_pthread_mutex_unlock(&endpoint->mutex);
-
-	tt_pthread_setcancelstate(old_cancel_state, NULL);
 
-	free(pipe);
-}
 
 static void
 cbus_create(struct cbus *bus)
diff --git a/src/lib/core/cbus.h b/src/lib/core/cbus.h
index c4e48d850..cdb20c8df 100644
--- a/src/lib/core/cbus.h
+++ b/src/lib/core/cbus.h
@@ -110,9 +110,12 @@ void
 cmsg_deliver(struct cmsg *msg);
 
 /**
- * A common part of `cpipe` and `lcpipe`.
+ * A  uni-directional FIFO queue from any thread to cord.
+ * It is a light version of the cpipe, since some features are not needed for
+ * 3td party threads or they cannot be implemented.
+ *
  */
-struct pipe_base {
+struct lcpipe {
 	/** Staging area for pushed messages */
 	struct stailq input;
 	/** Counters are useful for finer-grained scheduling. */
@@ -133,7 +136,7 @@ struct pipe_base {
 /** A  uni-directional FIFO queue from one cord to another. */
 struct cpipe {
 	/** Base part of a pipe. */
-	struct pipe_base base;
+	struct lcpipe base;
 	/**
 	 * Rather than flushing input into the pipe
 	 * whenever a single message or a batch is
@@ -496,17 +499,6 @@ cbus_unpair(struct cpipe *dest_pipe, struct cpipe *src_pipe,
 	    void (*unpair_cb)(void *), void *unpair_arg,
 	    void (*process_cb)(struct cbus_endpoint *));
 
-/**
- * A  uni-directional FIFO queue from any thread to cord.
- * It is a light version of the cpipe, since some features are not needed for
- * 3td party threads or they cannot be implemented.
- *
- */
-struct lcpipe {
-	/** Base part of a pipe. */
-	struct pipe_base base;
-};
-
 /**
  * Create and initialize a pipe and connect it to the consumer.
  * The call returns only when the consumer, identified by consumer name, has
```
</details>

Please, rework the code in this way. lcpipe is the base, and cpipe reuses it similar to how I show it in the diff above.

src/lib/core/cbus.c
Gerold103: Yes, you need to do a few extra steps. But they are all easy to do. I've shown you diff in my other comment. Please, apply it and do the same for the other methods. When cpipe is doing something on top, you can just do it on top. The methods obviously won't be 100%. But the cpipe-specific code looks like it can be done before or after the work on its `base` member.

changelogs/unreleased/cbus-lcpipe.md
Gerold103: I would simply make it in a separate commit in the same PR.

src/lib/core/cbus.c
Gerold103: > cpipe_push_input checks for the presence of a loop (we need cord for this) in the producer side: assert(loop() == pipe->producer);

And what is the problem with that? The idea is not to make cpipe and lcpipe use exactly the same methods but reuse the main part of the code. This assertion you can simply leave at the cpipe level:
```C
void
lcpipe_push(struct lcpipe *pipe, struct cmsg *msg)
{
    assert(msg->hop->pipe == NULL);
    stailq_add_tail_entry(&pipe->base.input, msg, fifo);
    pipe->base.n_input++;
    if (pipe->base.n_input >= pipe->base.max_input) {
        lcpipe_flush_input(pipe);
    }
}

static inline void
cpipe_push_input(struct cpipe *pipe, struct cmsg *msg)
{
    assert(loop() == pipe->producer);
    lcpipe_push(&pipe->base, msg);
}
```

As for the triggers - I noted in another commit, that these on-flush triggers are unused. Delete them as a preparatory commit, and you will be able to reuse the code like I did above.

> it cannot call cpipe_flush_cb because it requires ev_loop and ev_async which are not available outside the cord context

Where does it require them? If you mean `loop()`, then do like I've shown in the code above - keep it at the cpipe level. If you mean `ev_invoke()`, then I explained it in my first comment in this thread, that it isn't needed even in the cpipe. Fix it as another preparatory commit, and you will be able to reuse the code effortlessly.

sergepetrenko:
Thanks for your work on this and for the feature suggestion!
We have implemented a slightly different idea based on your approach in #10799

PR #9133, opened 2023-09-13
third_party: update libcurl from 7.87.0 to 8.3.0

### LINKED ISSUE #9086: http.client does not finish current streams after a connection lost
**Bug description**

I am trying to use the `http.client` library to communicate with [etcd](https://github.com/etcd-io/etcd) HTTP stream API. Everything functions properly until the connection with `etcd` is unexpectedly lost. In this case, all current streams hang until the request timeout ends.

In fact, this is a bug in `curl` (see https://github.com/curl/curl/issues/11769).

**Tarantool version**

tarantool version is `2.11.0-0-g247a9a418`.

**Operating system**

Ubuntu 22.04.2 LTS.
Linux 6.2.0-26-generic x86-64.

**Steps to reproduce**

1. Run etcd (for installation please check [this guide](https://github.com/etcd-io/etcd/releases)).
2. Execute in tarantool:
```
require("http.client").new():post("http://localhost:2379/v3/watch", nil, {chunked=true, keepalive_interval = 1, keepalive_idle = 1, verbose = true}):read(1, 10000)
```
3. Close etcd server: ```killall -SIGKILL $(pidof etcd)```
4. See that the reproducer hangs.

**Actual behavior**

tarantool instance hangs (until the timeout of the read ends).

**Expected behavior**

I expected that the request finishes (with an error) due to a lost connection. So the reproducer will finish too.

ligurio:
The patch updates curl module to the version 8.3.0 [1] and updates a CMake module for building curl library.

Changes in CMake module:

- Option `CURL_STATICLIB` is gone and replaced with `BUILD_STATIC_LIBS`.
- Option `CURL_USE_GNUTLS` was added and disabled by default.
- NSS library support was removed and option `CURL_USE_NSS` has been removed as well.
- Option `CMAKE_UNITY_BUILD` was added and disabled by default.
- Option `CURL_DISABLE_FORM_API` was added and disabled by default. It is in fact depends on `CURL_DISABLE_MIME`, but anyway disabled explicitly.

Changelog: https://curl.se/changes.html#8_3_0

1. https://github.com/curl/curl/releases/tag/curl-8_3_0

NO_DOC=libcurl submodule bump
NO_TEST=libcurl submodule bump

Fixes #9086
Follows up tarantool/etcd-client#62


Totktonada (approved):
No objections from my side. Thank you!

changelogs/unreleased/bump-libcurl-to-8.3.0.md
```diff
@@ -0,0 +1,3 @@
+## feature/build
+
+* Updated libcurl to version 8.3.0 (gh-9086).
```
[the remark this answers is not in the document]
ligurio: > I would make the description more for a user: fixed a streaming connection stuck if etcd is stopped unexpectedly, for example. (And I would qualify this change as a bugfix.)

I like idea. However, it is worth to say about bump too, so I've added changelog with bug and left changelog with bump.

changelogs/unreleased/bump-libcurl-to-8.3.0.md
```diff
@@ -0,0 +1,3 @@
+## feature/build
+
+* Updated libcurl to version 8.3.0.
```
xuniq, line 3:
```suggestion
* Updated libcurl to version 8.3.0 (gh-9086).
```

xuniq (approved):

kyukhin (approved):

changelogs/unreleased/bump-libcurl-to-8.3.0.md
ligurio: This bug is specified in the second changelog. Bump was done without an issue.

Issue #5878, opened 2021-03-11
Implement ASAN support in LuaJIT


igormunkin:
Despite the fact [the necessary changes are committed to Tarantool LuaJIT mirror](https://github.com/tarantool/tarantool/issues/1765#issuecomment-246313302), *SAN support is done in Tarantool repo, but it's not [fully implemented in LuaJIT fork](https://github.com/tarantool/tarantool/blob/07c83aab5c066ca75c149112b331b4dbb81b3f38/cmake/luajit.cmake#L60-L64): only some places are wrapped with `LUAJIT_USE_ASAN` ifdef here and there. However, as a result of #4862 we faced some [failures in CI](https://github.com/tarantool/tarantool/runs/1999839396) reporting suspicious leaks in LuaJIT runtime.

As a result of the brief investigation here are the following inputs:
* There are [some default flags](https://github.com/tarantool/luajit/blob/c66e815769dc2d39a7758ef5f7bd25720549afaa/src/Makefile.original#L35-L39) in LuaJIT original build system, e.g. `CCOPT` containing `-fomit-frame-pointer`.
* Before #4862 is resolved, all LuaJIT build flags are managed within Tarantool CMake machinery, so `CCOPT` flag [was tweaked](https://github.com/tarantool/tarantool/blob/8484cabc40d97433a70c77f0ff3051c21cae6961/cmake/luajit.cmake#L255).
* After porting LuaJIT to CMake in scope of #4862, `-fomit-frame-pointer` [is enabled by default](https://github.com/tarantool/luajit/blob/b8b0c418f946b9515de06ecbc18c49c079aa8f46/CMakeLists.txt#L91-L96) in LuaJIT and cannot be disabled outside from LuaJIT build machinery.
* Running both ASAN and LSAN with `fast_unwind_on_malloc=0` doesn't make CI green.

Hence, as a follow up for #4862 ASAN support ought to be introduced within LuaJIT build system. Until the issue isn't resolved, all *SAN warnings reported against LuaJIT runtime are suppressed, since they might be related to build peculiarities (i.e. considered false positives).

Issue #3071, opened 2018-01-23
Trigger garbage collector before exit to prevent valgrind's false positives


Totktonada:
At least it worth to check whether this helps us with valgrind.

Totktonada:
Tried like so:

cmake . -DSQL_MAINTAINER_MODE=1 -DCMAKE_BUILD_TYPE=Debug -DENABLE_VALGRIND=ON && make -j
valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes -v --suppressions=./third_party/luajit/src/lj.supp ./src/tarantool < <(echo "print('hello'); collectgarbage()")

With and without collectgarbage() call. Got 654 errors with collectgarbage and 710 errors w/o. A lot luajit errors still reported, the overall picture does not changed significantly. We need either investigate luajit futher and deeper, or skip this effort for now and close the issue.

igormunkin:
I believe the root cause is that [`tarantool_lua_free`](https://github.com/tarantool/tarantool/blob/master/src/lua/init.c#L750-L785) routine [**is not called**](https://github.com/tarantool/tarantool/blob/master/src/main.cc#L536-L547) due to some reasons (unclear to me). I remember @Buristan also has some complains regarding this.

Buristan:
Yes, It is disabled via tarantool/tarantool@a718d469100a2992707c16cd251a56e5e5159be2. To me it looks like a black hole attracting memory leaks. Also, I see no tests failures (at least locally) if enable back `tarantool_lua_free()` routine. I assume, that we may **carefully** enable part of those "atexit handlers". But enabling the only one `tarantool_lua_free()` is already the good start.

PR #9142, opened 2023-09-14
ci: fix CMake options in a build with enabled ASAN

ligurio:
**Blocked by** https://github.com/tarantool/tarantool/issues/3071

As a result of the commit in LuaJIT submodule tarantool/luajit@a86e376 ("build: introduce LUAJIT_USE_ASAN option") ASan and LSan support has been finally added to LuaJIT runtime.

Furthermore, the internal LuaJIT memory allocator is not instrumented yet, so to find any memory faults it's worth building LuaJIT with system provided memory allocator (i.e. enable LUAJIT_USE_SYSMALLOC option). Using LUAJIT_USE_SYSMALLOC requires enabling LUAJIT_ENABLE_GC64.

The patch enables both CMake options in a build with enabled ASAN. Without path the following message is appears:

CMake Warning at third_party/luajit/CMakeLists.txt:278 (message):
  Unfortunately, internal LuaJIT memory allocator is not instrumented yet, so
  to find any memory errors it's better to build LuaJIT with system provided
  memory allocator (i.e.  run CMake configuration phase with
  -DLUAJIT_USE_SYSMALLOC=ON).

Follows up #5878

NO_CHANGELOG=ci
NO_DOC=ci
NO_TEST=ci

### DIFF
--- a/.test.mk
+++ b/.test.mk
@@ -81,6 +81,8 @@ test-release: build run-luajit-test run-test
 # Experiments once again confirm the notorious quote that "640 Kb
 # ought to be enough for anybody".
 test-release-asan: CMAKE_PARAMS = -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+                                  -DLUAJIT_USE_SYSMALLOC=ON \
+                                  -DLUAJIT_ENABLE_GC64=ON \
                                   -DENABLE_WERROR=ON \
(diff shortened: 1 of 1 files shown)


ligurio:
Blocked by https://github.com/tarantool/tarantool/issues/3071.

Issue #4799, opened 2020-03-08
C stored function works different when called locally and remotely


Gerold103:
Example. Run this from tarantool sources root folder:
```
box.cfg{listen = 3313}
box.schema.user.grant('guest', 'super')
build_path = '/Users/gerold/Work/Repositories/tarantool'
package.cpath = build_path..'/test/box/?.so;'..build_path..'/test/box/?.dylib;'..package.cpath

box.schema.func.create('function1.multireturn', {language = "C", exports = {'LUA'}})

c = require('net.box').connect(box.cfg.listen)
c:call('function1.multireturn')
---
- [[1], [1]]
...
box.func['function1.multireturn']:call()
---
- [1]
- [1]
...
```
Result of the function depends on from where it was called. When I call a Lua function, it returns the same result regardless of how I call it.
I don't know what a behaviour is correct - return an array in both cases, or multireturn. Seems like C functions always returned an array, because there was no other way to call them except via netbox, until `box.func` was introduced. I don't know. But behaviour should be at least the same.

Korablev77:
It does not concern only C stored functions - Lua ones show the same bevaiour. Setup:
```
box.cfg{listen = 3301}
nb = require('net.box')
cn = nb:connect(3301)
box.execute('create table t(id int primary key)')
box.space.T:insert({1})
box.space.T:insert({2})
box.space.T:insert({3})
function foo() return 1, 2, 3 end
function select() return box.space.T:select() end
function replace() return box.space.T:replace({3}) end
box.schema.user.grant('guest', 'super')
```
Local calls:
```
tarantool> foo()
---
- 1
- 2
- 3
...

tarantool> select()
---
- - [1]
  - [2]
  - [3]
...

tarantool> replace()
---
- [3]
...
```
Remote calls:
```
tarantool> cn:call('select')
---
- [[1], [2], [3]]
...

tarantool> cn:call('replace')
---
- [3]
...

tarantool> cn:call('foo')
---
- 1
- 2
- 3
...
tarantool> cn.space.T:select()
---
- - [1]
  - [2]
  - [3]
...

```
As one can see, result for `select()` different for local and remote calls. However, there's special (legacy) `call16`, which emulates local behaviour:
```
tarantool> cn:call_16('select')
---
- - [1]
  - [2]
  - [3]
...
```
I have no idea why current call replaced call_16, but it looks like we shouldn't touch it...

Gerold103:
[...]

PR #9143, opened 2023-09-14
Fix inconsistent multireturn values for stored C functions

Buristan:
The first commit introduces the `c_func_iproto_multireturn` option (when set to new, the results of the stored C function aren't wrapped in an additional msgpack array, see below). By default, it is set to 'old', to avoid conflicts.  Also, this makes the first patch backportable to 2.11 as is (except conflict with compatibility options in compat.lua, see #9144). The second commit switches the behaviour to 'new' and fixes some diff tests.

As described in the commit message to the first patch, the SQL mechanism depends on the `port_dump_msgpack()` too, so just using the introduced flag in `port_c_dump_msgpack()` isn't good. As an alternative option, we can introduce 2 new fields (say `encode_call_results16`, `encode_call_results`) in `port_vtab` to call them and use the flag only there. But I'm not sure that adding 2 new fields to the structure only for this is good either.

As far as there are only two possible fields in the changelog header entry (`feature/*` or `bugfix/*`), I prefer the second one instead of `compatibility/*`.

I provide the doc-request below for your convenience in reading:

---

Title: Document `c_func_iproto_multireturn` compat option

Please create a documentation page for the new compat option:
https://tarantool.io/compat/c_func_iproto_multireturn

In the new behaviour, the multiresults returned by a stored C function
via iproto aren't wrapped in the additional msgpack array (old).

```
tarantool> compat.c_func_iproto_multireturn = 'old'
---
...

tarantool> net_box.connect(box.cfg.listen):call('myclib.cfunc')
---
- [true, -1]
...

tarantool> compat.c_func_iproto_multireturn = 'new'
---
...

tarantool> net_box.connect(box.cfg.listen):call('myclib.cfunc')
---
- true
- -1
...

```

The new behaviour is consistent with the local call of the function
via `box.func`:

```
tarantool> box.func['myclib.cfunc']:call()
---
- true
- -1
...

```

Assume you have a stored C function that returns values like the
following:

```c
char *position = mp_encode_bool(buffer, true);
box_return_mp(ctx, buffer, position);
/* ... */
position = mp_encode_int(buffer, -1);
box_return_mp(ctx, buffer, position);
```

If you want to preserve the format of the returned array for your C
functions, when the `c_func_iproto_multireturn` option is set to "new",
you should add the additional wrapping, like the following:

```c
char *position = mp_encode_array(buffer_with_results, n_results);
position = mp_encode_bool(position, true);
/* ... */
position = mp_encode_int(position, -1);
box_return_mp(ctx, buffer_with_results, position);
```

The amount of `box_return_mp()` calls indicates the number of values to
be returned.

Also, you should update its usage via `box.func` if there is any.

### DIFF
--- a/changelogs/unreleased/gh-4799-consistent-c-functions-calls.md
+++ b/changelogs/unreleased/gh-4799-consistent-c-functions-calls.md
@@ -1,0 +1,5 @@
+## bugfix/box
+
+* **[Breaking change]** Added a `c_func_iproto_multireturn` option to the
+  `compat` module. The new behavior drops an additional array that wraps
+  multiple results returned via iproto (gh-4799).
--- a/changelogs/unreleased/gh-8576-set-compat-options-to-new.md
+++ b/changelogs/unreleased/gh-8576-set-compat-options-to-new.md
@@ -7,5 +7,6 @@
     * `fiber_channel_close_mode`
     * `fiber_slice_default`
     * `box_cfg_replication_sync_timeout`
+    * `c_func_iproto_multireturn`
 
   More information on the new behavior can be found on the [Module compat](https://www.tarantool.io/en/doc/latest/reference/reference_lua/compat/) page.
--- a/src/box/port.c
+++ b/src/box/port.c
@@ -231,7 +231,7 @@ port_c_dump_msgpack(struct port *base, struct obuf *out)
 /**
  * If set, don't wrap results into an additional array on encode.
  */
-static bool c_func_iproto_multireturn = false;
+static bool c_func_iproto_multireturn = true;
 TWEAK_BOOL(c_func_iproto_multireturn);
 
 int
--- a/src/box/port.c
+++ b/src/box/port.c
@@ -37,6 +37,7 @@
 #include <fiber.h>
 #include "errinj.h"
 #include "mpstream/mpstream.h"
+#include "tweaks.h"
 
 /**
  * The pool is used by port_c to allocate entries and to store
@@ -204,7 +205,7 @@ port_c_add_str(struct port *base, const char *str, uint32_t len)
 }
 
 static int
-port_c_dump_msgpack_16(struct port *base, struct obuf *out)
+port_c_dump_msgpack(struct port *base, struct obuf *out)
 {
 	struct port_c *port = (struct port_c *)base;
 	struct port_c_entry *pe;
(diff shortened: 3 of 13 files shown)


ochaplashkin (approved):
LGTM :rocket:

src/box/iproto.cc
```diff
@@ -2322,7 +2322,8 @@ tx_process_call(struct cmsg *m)
 	out = msg->connection->tx.p_obuf;
 	iproto_prepare_select(out, &svp);
 
-	if (msg->header.type == IPROTO_CALL_16)
+	if (msg->header.type == IPROTO_CALL_16 ||
+	    port_c_unwrap_multireturn(&port))
```
Totktonada, line 2326: It seems, we can get rid of iproto.cc changes by adjusting `port_sql_dump_msgpack` function. It is a bit messy, but this mess will be at the lower level.

I see no test fails regarding box.session.push/box_session_push if I change port_c_dump_msgpack directly. Please, re-check this point.

Buristan: Yes. Updated the branch. Also, I renamed functions, as we discussed offline, and added a note about the amount of `box_return_mp()` in the doc request.

`box.session.push()` uses the Lua port, so these changes are unrelated.

changelogs/unreleased/gh-4799-consistent-c-functions-calls.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/box
+
+* **[Breaking change]** Added a `c_func_iproto_multireturn` option to
+  tarantool.compat. The new behaviour drops an additional array that wraps
+  multiple results returned via iproto (gh-4799).
```
xuniq, line 5:
```suggestion
* **[Breaking change]** Added a `c_func_iproto_multireturn` option to the
  `compat` module. The new behavior drops an additional array that wraps
  multiple results returned via iproto (gh-4799).
```

xuniq (approved):

Totktonada (approved):
I'm OK with the implementation if the approach with changing the default behavior for existing code (including built *.so/*.dylib) is the best that we can do.

I would like some more gentle upgrade path, but I don't see any variant that would solve the main problem: consistency between iproto call and box.func call (plus consistency with iproto/box.func calls of a Lua code).

If anyone has ideas here, I would appreciate them.

Gerold103 (approved):
Thanks for the patch! All looks good.

PR #9147, opened 2023-09-14
ci: convert shallow submodules to complete ones 

ylobankov:
This patch is intended to resolve the following issue that appears while
calling `cmake` for submodules from time to time:

    fatal: No tags can describe '<sha>'.
    Try --always, or create some tags.

This happens because the `actions/checkout` GitHub action fetches the
repository with the `--unshallow` option, but it doesn't do the same for
submodules. So this leads to the error above while using `git describe`
for submodules.

This is fixed by converting a shallow submodule to a complete one.

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -72,6 +72,13 @@ runs:
         # 2. Build package on tag 2.4.8. It will produce package with
         #    name 2.4.8.x.
         git fetch --tags -f
+        # Found that actions/checkout fetches the repository with the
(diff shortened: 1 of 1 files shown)


ochaplashkin (approved):
LGTM :rocket:

Totktonada:
For ones who is interested, the original problem is about LuaJIT's `SetVersion` CMake function, which fails on a shallow clone. Reported by @locker 2023-09-13 in the developer experience team chat.

image

Totktonada:
The issue in `actions/checkout`: https://github.com/actions/checkout/issues/435.

PR #9151, opened 2023-09-15
config: allow postponing object grants till creation

Lord-KA:
Before this patch privileges for space, function and sequence alone
were disabled. Only permission for universe could be granted. It was
done because there wasn't a way to grant or revoke a privilege when
subjecting object hasn't been created yet.
    
This patch allows grants and revokes for objects that will be created
after the applier. These actions will be postponed and executed via
on_commit trigger after the object is registered in the according system
space of box.space._space/_func/_sequence.
    
Now permission can be granted to space/function/sequence from within the
config.
    
Part of #8967
    
NO_DOC=documentation request will be filed manually for the whole credentials

### DIFF
--- a/changelogs/unreleased/gh-8967-set-creds-in-lazy-way.md
+++ b/changelogs/unreleased/gh-8967-set-creds-in-lazy-way.md
@@ -1,0 +1,5 @@
+## feature/config
+
+* Now the permissions for space, function, and sequence can be granted
+  via the config. If the object has not been created yet, the grant is
+  postponed and executed after the creation of the object (gh-8967).
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -2,6 +2,11 @@ local log = require('internal.config.utils.log')
 local digest = require('digest')
 local fiber = require('fiber')
 
+-- Var is set with the first apply() call.
+local config
+
+-- {{{ Sync helpers
+
 --[[
 This intermediate representation is a formatted set of
 all permissions for a user or role. It is required to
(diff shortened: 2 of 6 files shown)


test/config-luatest/helpers.lua
```diff
@@ -226,6 +226,7 @@ local function reload_success_case(g, opts)
     local script_2 = opts.script_2
     local options = assert(opts.options)
     local verify_2 = assert(opts.verify_2)
+    local verify_args_2 = opts.verify_args_2
```
Totktonada, line 229: Nit: While we're here it would be good to add `verify_args` and `verify_args_2` into the list of options in the comment for the function:

```diff
 -- * opts.roles
 -- * opts.script
 -- * opts.options
 -- * opts.verify
+-- * opts.verify_args
 --
 --   Same as in success_case().
@@
 -- * opts.verify_2
 --
 -- Verify test invariants after config:reload().
+--
+-- * opts.verify_args_2
+--
+-- Arguments for the verify_2 function.
```

test/config-luatest/credentials_applier_test.lua
```diff
@@ -783,3 +783,255 @@ g.test_sync_ro_rw = function(g)
...
+
+            local exp_perm_1 = iconfig.credentials.users.myuser1
+            exp_perm_1 = internal.privileges_from_config(exp_perm_1)
+            exp_perm_1 = internal.privileges_add_defaults('myuser1', 'user',
+                                                          exp_perm_1)
+
+            local exp_perm_2 = iconfig.credentials.users.myuser2
+            exp_perm_2 = internal.privileges_from_config(exp_perm_2)
+            exp_perm_2 = internal.privileges_add_defaults('myuser2', 'user',
+                                                          exp_perm_2)
```
Totktonada, line 910: Here we validate tarantool's behavior that depends on the given `privileges_*` functions using these functions. Comparing to a simple-but-large test with hardcoded expected values this approach may miss some mistakes.

Let's assume that `privileges_from_config` should return some privilege, but it doesn't. It means that tarantool will not grant the given privilege. However, the test will also think that the privilege shouldn't be granted.

I'm in-doubt-but-ok regarding using `privileges_from_box` in test oracles. However, using `privileges_from_config` and `privileges_from_defaults` seems dangerous for me.

test/config-luatest/credentials_applier_test.lua
```diff
@@ -783,3 +783,255 @@ g.test_sync_ro_rw = function(g)
...
+                            },
+                        },
+                    },
+                },
+            },
+        },
+    }
+
+    -- Create myspace1, check, rename to myspace2, check, create myspace1,
+    -- check.
```
Totktonada, line 945: TBH, I doesn't spot the key point of the test. I'll try to ask several questions around in order to understand it better.

There is a configuration reloading before the second creating of myspace1. Is it important for the test case? If so, it should be mentioned in the quoted list of actions. If not, `success_case` can be used. Is it correct?

What is the key difference of this test case from the next test case (which uses `success_case`)?

Maybe it worth to mention possible negative scenario: what exactly shouldn't go in a wrong way. For example, write not just 'check', but 'check that <...>'.

test/config-luatest/credentials_applier_test.lua
```diff
@@ -783,3 +783,255 @@ g.test_sync_ro_rw = function(g)
...
+            local exp_perm = iconfig.credentials.users.myuser1
+            exp_perm = internal.privileges_from_config(exp_perm)
+            exp_perm = internal.privileges_add_defaults('myuser1', 'user',
+                                                        exp_perm)
+            t.assert_equals(perm, exp_perm)
+        end,
+        verify_args_2 = {iconfig},
+    })
+
+    -- Create myspace2, check, rename to myspace1, check.
```
Totktonada, line 991: I propose a bit more detailed wording.

* Declare myspace1 privs in the config.
* Create myspace2.
* Check that myspace2 has no privs declared for myspace1[^1].
* Rename myspace2 to myspace1.
* Check that myspace1 has privs as declared in the config.
* Check that myspace2 has no privs declared for myspace1[^1].

[^1]: Whether this check is needed -- there is no such object at all?

Totktonada, line 991: I'm interested in effect of the following sequence of steps.

* Declare *different* myspace1 and myspace2 privs in the config.
* Create myspace1 and myspace2.
* Check that the privs for myspace1 and myspace2 in sync with the config.
* Rename myspace1 to myspace_tmp.
* Rename myspace2 to myspace1.
* Rename myspace_tmp to myspace2.
* Check that the privs for myspace1 and myspace2 in sync with the config.

I'm reading the patch starting from tests, so at the moment I don't know, whether it is important to add such a test case. Please, consider whether it is needed.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,6 +277,101 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+       end
+       actions_on_creation[obj_type][obj_name] = nil
+       ::skip::
+    end
+    log.debug('on_commit_trigger finished')
+end
+
+-- Check that event is space/function/sequence creation and set
+-- on_commit_trigger for the transaction.
+local function on_replace_trigger(_, new_space, obj_type, request_type)
```
Totktonada, line 346: I guess that we don't want to attempt to perform DML in the read-only mode, so we should skip all the actions for 'applier' [session type](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_session/type/). It is a usual way to write on_replace/before_replace triggers.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,6 +277,101 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+--     privileges
+--   }, {
+--     ...
+--   },
+-- }
+local actions_on_creation = {
+    ['space'] = {},
+    ['function'] = {},
+    ['sequence'] = {},
+}
```
Totktonada, line 295: I see two possible ways to interpret a sequence of configs: snapshots or deltas. You can feel this distinction familiar by the [What is Git?](https://git-scm.com/book/en/v2/Getting-Started-What-is-Git%3F) section of Git Book.

While one representation is generally equivalent to another, there is a nuance. Let's consider an example of configuration snapshots: C1 -> C2 -> C3 and corresponding deltas: D1 (? -> C1), D2 (C1 -> C2), D3 (C2 -> C3).

Say, we have a function that grant/revoke needed privileges. It looks negligible, whether it accepts a snapshot (C3) or a delta (D3). The nuance here is that a delta is only valid, when it is applied on top of a right previous snapshot: applying D3 gives C3 if we're in C2, but may give some different C3' otherwise.

Back to our task. There are several points, where I'm in doubts about using deltas.

First, we're already operating by snapshots on startup/reload and on ro -> rw database mode change. The intermediate representation and all the functions around use it. Now, we have another delta-based intermediate representation and a parallel set of functions around it.

It seems, it reduces our ability to reuse the code and adds unnecessary entities.

I'm also in doubt about synchronization. If we would reuse the same code, we would reuse the synchronization code too.

Second, the delta is calculated using some current database schema state, while it is applied on a possibly different state. So, we can attempt to apply some changes twice (`if_not_exists` helps here) and, what is worse, we can miss some of the changes.

It seems that continue using snapshot-based approach should make the code simpler and more durable. If you considered this approach, please, share reasons to don't use it. Otherwise, please, pay some time to investigate this implementation variant.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,6 +277,101 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+            obj_type == nil then
+        return
+    end
+
+    -- Check that there actions postponed for the object.
+    local obj_name = new_space.name
+    local obj_info = actions_on_creation[obj_type][obj_name] or {}
+    if not next(obj_info) then
+        return
+    end
```
Totktonada, line 362: If we rename a space, we should revoke all the privileges that we given by the provious configuration. IOW, we should finish in the state that is equivalent to one after `:reload()` with the new config.

Totktonada: If we reuse the config -> box synchronization code on any trigger wake up, this problem should gone.

test/config-luatest/credentials_applier_test.lua
Lord-KA: > There is a configuration reloading before the second creating of myspace1. Is it important for the test case? If so, it should be mentioned in the quoted list of actions. If not, success_case can be used. Is it correct?

Yes, it is important. This is the scenario I told you about previously. We create an object1, it gets its privs, we rename it to object2 and it tries to apply colliding set of privs or just twice, depending on implementation. In my implementation it checks whether the list of actions is cleared on reload and if the `if_not_exists` inside trigger action is `true`.

Lord-KA: > What is the key difference of this test case from the next test case (which uses success_case)?

The next test case is straightforward rename, not a corner case. I felt that it is not well covered by the previous one.

Lord-KA: Added `reload` to the comment.

src/box/lua/config/applier/credentials.lua
Lord-KA: > If we rename a space

Generally speaking, we don't rename spaces. And when an object gets changed by users hand by `box.*` command, it isn't our responsibility to change configuration afterwards. We explicitly agreed that all the privileges granted after the applier are not to be revoked. I am not sure that in this case it is a good idea either.
Another question is, what do we do if the renaming happens twice? Do we sync privileges each time or just once?

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
```diff
@@ -0,0 +1,5 @@
+## feature/config
+
+* On read-only instances, Tarantool now synchronizes credentials
+  in the background when switching to read-write mode instead of
+  skipping the synchronization (gh-8967).
```
Totktonada, line 5: 711756d727456bddd39ee52945dd5ae27a1a5684 is in `master`. It worth to rebase.

test/config-luatest/credentials_applier_test.lua
Totktonada: So, in brief, this scenario attempts to trigger 'already exists' error on a privilege granting on `:reload()`. OK, got it.

IMHO, it is hard to deduce from the code itself, so worth to clarify in a comment.

I'm not insisting about test cases, as usual, but you can use my feeling to guess how another developer will read it.

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
```diff
@@ -0,0 +1,5 @@
+## feature/config
+
+* On read-only instances, Tarantool now synchronizes credentials
+  in the background when switching to read-write mode instead of
```
xuniq, line 4:
```suggestion
  in the background when switching to read-write mode, instead of
```

xuniq (approved):

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
Lord-KA: The changelog has already been merged.

test/config-luatest/helpers.lua
Lord-KA: Applied.

changelogs/unreleased/gh-8967-set-creds-in-lazy-way.md
```diff
@@ -0,0 +1,5 @@
+## feature/config
+
+* Now permissions for space, function and sequence can be granted via
+  the config. If the object has not been created yet, the grant will
+  be postponed and executed after the creation (gh-8967).
```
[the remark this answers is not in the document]
Lord-KA: Applied. Thanks!

changelogs/unreleased/gh-8967-creds-postope-sync-when-RO-till-RW.md
Lord-KA: Done.

src/box/lua/config/applier/credentials.lua
Lord-KA: Done:
```diff
+      if box.session.type() == 'applier' then
+          -- The event is caused by replication, so the instance is in
+          -- RO mode. Thus, skip all applier actions.
+          return                                                                          
+      end

File: test/config-luatest/credentials_applier_test.lua (hunk 3 above)
Comment (Lord-KA) [reply test/config-luatest/credentials_applier_test.lua:945]: Updated the comments with info about what they test instead of how they test it.

File: test/config-luatest/credentials_applier_test.lua (hunk 4 above)
Comment (Lord-KA) [reply test/config-luatest/credentials_applier_test.lua:991]: > Updated the comments with info about what they test instead of how they test it.

File: src/box/lua/config/applier/credentials.lua (hunk 6 above)
Comment (Lord-KA) [reply src/box/lua/config/applier/credentials.lua]: Implemented per-object sync inside the trigger.

File: src/box/lua/config/applier/credentials.lua (hunk 7 above)
Comment (Lord-KA) [reply src/box/lua/config/applier/credentials.lua]: We've discussed it verbally and agreed to revoke excessive privileges on rename. It is done with the brand-new per-object sync that is implemented in the `TO_SQUASH` commit.

File: test/config-luatest/credentials_applier_test.lua (hunk 2 above)
Comment (Lord-KA) [reply test/config-luatest/credentials_applier_test.lua:921]: Ok, will do.

Comment (Lord-KA) [reply test/config-luatest/credentials_applier_test.lua:921]: Added additional manual checks for both tests.

File: test/config-luatest/credentials_applier_test.lua (hunk 4 above)
Comment (Lord-KA) [reply test/config-luatest/credentials_applier_test.lua]: Sigh, we discussed it verbally once again and decided to have the sync performed (per object) on every object creation or rename. So now the test can be implemented as you suggest, so I added it.

File: src/box/lua/config/applier/credentials.lua [hunk 11]
```diff
@@ -277,6 +282,190 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+
+        -- Check if the request is for obj creation or obj rename.
+        if old_obj == nil or old_obj.name ~= new_obj.name then
+            local obj_name = new_obj.name
+
+            -- Check that the object is present inside config.
+            if on_creation_flags[obj_type][obj_name] then
+                local credentials = get_credentials(config)
+                sync_privileges(credentials, {name = obj_name, type = obj_type})
+            end
```
Totktonada, line 381: Gleb highlighted that DQL/DML/DDL is forbidden in on_commit triggers (see a49c8c02a2f6b7620db12988ec7ee0369076d4d4). So, it seems, we're obligated to implement it in some other way. (This comment is to track the problem.)

Lord-KA: Now the sync is performed inside a dedicated fiber. The `on_commit` trigger just schedules a sync for the object via `sync_tasks` fiber channel.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,85 +282,291 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+-- Iterate through requests and sync when the according flag is set.
+local function on_commit_trigger(iterator)
+    log.debug('credentials.apply: on_commit_trigger() started')
+    for _, old_obj, new_obj, space_id in iterator() do
+        if box.space[space_id] == nil then
+            goto skip
+        end
+
+        local obj_type = box.space[space_id].name
+        obj_type = translate_obj_type(obj_type)
```
Totktonada, line 369: AFAIU it is undesirable to access space objects in an on_commit trigger. So, I would map `space_id` directly to an object type. Let's use `box.schema.SPACE_ID`, `box.schema.SEQUENCE_ID` and `box.schema.FUNC_ID` constants.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,85 +282,291 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
-        log.debug('credentials.apply: ' .. role_or_user ..
-                  '.grant(%q, %q, %q, %q)', name, privs, obj_type, obj_name)
-        box.schema[role_or_user].grant(name, privs, obj_type, obj_name)
+-- The table contains flags for objects that should be synced in the trigger:
+-- [obj_type][obj_name] = true/nil
+local on_creation_flags = {
+    ['space'] = {},
+    ['function'] = {},
+    ['sequence'] = {},
+}
```
Totktonada, line 343: I think that it would be helpful to highlight that it is a mapping from a current config. It worth to reflect it in the name and it would be good to mention that it is always in sync with a latest config in a comment.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,85 +282,291 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+        return 'sequence'
     end
-    local revoke_f = function(name, privs, obj_type, obj_name)
-        privs = table.concat(privs, ',')
-        log.debug('credentials.apply: ' .. role_or_user ..
-                  '.revoke(%q, %q, %q, %q)', name, privs, obj_type, obj_name)
-        box.schema[role_or_user].revoke(name, privs, obj_type, obj_name)
+end
+
+local sync_privileges
```
Totktonada, line 358: The function is not used till its definition, so we can just use `local function sync_privileges(<...>) <...> end` as before. IOW, there is no need to use a forward declaration.

src/box/lua/config/applier/credentials.lua
```diff
@@ -277,85 +282,291 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+                privileges_action_f('revoke', role_or_user, name,
+                                    to_revoke.privs, to_revoke.obj_type,
+                                    to_revoke.obj_name)
+            end
+        end
     end
 
+    -- The logic inside the sync() function, i.e. privileges dump from
+    -- box, diff calculation and grants/revokes should happen inside one
+    -- transaction. It is the case when called from set_credentials(). When
```
Totktonada, line 538: There is no `set_credentials()` function now. It seems, the comment needs an update.

src/box/lua/config/applier/credentials.lua
Lord-KA: Thanks, missed that. Removed.

src/box/lua/config/applier/credentials.lua
Lord-KA: Updated:
```lua
  -- This is a map of all objects that are mentioned in the config.
  -- It is always in sync with the latest reload and has the following format:
  -- [obj_type][obj_name] = true/nil                                                                     
  local target_object_map = {                                                                         
      ['space'] = {},                                                                                    
      ['function'] = {},                                                                                 
      ['sequence'] = {},                                                                                 
  }                                                                                                      
```

src/box/lua/config/applier/credentials.lua
Lord-KA: Good catch! I removed `box.space[space_id].name `, added `if-elseif` checks inside `on_commit` trigger and inlined old `translate_obj_type` in `on_replace_trigger()`.

Totktonada (approved):
Thank you for your work!

Wow, I felt that this functionality shouldn't be so large.

I see no flaws in the implementation visually, while I'm able to understand what is going on in the code. I also can't propose more scenarios to test. I think it means that the patchset is good.

PR #9160, opened 2023-09-19
config: add `security.auth_retries` option

locker:
The new option is backed by `box.cfg.auth_retries`. It is available only in Enterprise Edition builds.
For the implementation see https://github.com/tarantool/tarantool-ee/pull/560.

Needed for tarantool/tarantool-ee#541


Totktonada (approved):
Looks OK on a brief glance.

But I'm not sure how it is better to pass config schema updates to the documentation team. https://github.com/tarantool/doc/issues/3544 links the last schema and, maybe, it actually doesn't require a separate documentation request.

I think we'll sync regarding the schemas once again with the doc team before a release candidate.

PR #9168, opened 2023-09-21
fix curL

fckxorg:


igormunkin:
The changeset is merged to master in scope of #9300, backported to 2.11 and 2.10 in scope of #9301 and #9302 respectively.

Issue #8978, opened 2023-08-11
Allow to set instance/replicaset/cluster names on recovery at first `box.cfg`


Gerold103:
The names are supposed to be set on instance bootstrap and never change. However this is problematic when it comes to setting them first time on already existing instances which were upgraded to >= 3.0.

Right now those instances have to be started without names, then the names have to be set in a non-first `box.cfg{}`.

Apparently it is not enough and people want to set them first time on first `box.cfg{}` on recovery.

There is no an easy way to do that. The problems are on read-only instances:
- They can't set replicaset and cluster names. So such `box.cfg{}` with new names would fair or hang, no other options.
- Instance name can't be set because the instance have to wait for the master to register the new name for this instance. The waiting might contradict with options like `replication_sync_timeout`.

Besides, if there are multiple read-write instances, it is not clear what happens if they all set replicaset and cluster names.

The solution approved by PMs is this: if `force_recovery` is set and at least one stored name doesn't match the name in the config, then try to fix them. `box.cfg{}` will hang until the instance becomes writable. The writable instance persists new replicaset and cluster names, and own instance name. The read-only instances have to wait until the writeable one sets the new names. For the instance name they will have to retry their `IPROTO_REGISTER` request until they get the new name. The timeout for this waiting could be something like `replication_timeout x 4` or `replication_connect_timeout` or something. That isn't too important, up to the implementation.

This should only work if there was no old name at all (i.e. it was `NULL`). Otherwise the behaviour remains like now.

PR #9187, opened 2023-09-27
config: set names automatically on schema upgrade

Serpentian:
box: setting names is NoOp before schema 3.0.0

Currently it's impossible to recover from the snapshot, which doesn't
have names set in it. This makes upgrading to 3.0.0 difficult and
requires starting Tarantool manually in interactive mode, invoking
box.cfg without names, upgrading schema and setting names manually.

We need to make Tarantool set names automatically, when it's configured
via yaml file. Even with invoking box.cfg two times (without names and
with them) it's impossible to set names before the time, schema is
upgraded. But we consider schema auto-upgrade unacceptable and expect
from users to do it, when they're ready.

So, let's allow setting names in box.cfg on the first configuration
and recover from snaps, which don't have names set. As we cannot apply
such names before schema upgrade, setting names on schema version below
3.0.0 has no effect.

Needed for https://github.com/tarantool/tarantool/issues/8978

NO_DOC=https://github.com/tarantool/doc/issues/3661

---

box: move mkversion into a separate module

This commit moves all code, related to working with versions and which
was used in box/lua/upgrade.lua, into a separate module and exports it
to Lua API as 'internal.version'

This is needed, as in the following commit we start to broadcast the
dd_version_id variable, which shows the version of schema, stored in
_schema['version']. This module allows us to write a trigger for
config module and avoid code duplication.

Needed for https://github.com/tarantool/tarantool/issues/8978

NO_DOC=internal
NO_TEST=<already tested>
NO_CHANGELOG=internal

---

box: broadcast dd_version_id in box.schema

This commit starts to broadcast dd_version_id, which will be used
to write a trigger, which will be run on schema upgrade. We cannot
use on_replace/on_commit triggers for that, as it's either forbidden
or not safe to yield in them. Moreover, on_replace trigger forbids
making DDL changes in it.

Needed for https://github.com/tarantool/tarantool/issues/8978

NO_TEST=<in the following commit>

@TarantoolBot document
Title: broadcast dd_version_id in box.schema event

One new field have been added to box.schema system event: dd_version_id,
which is represented as bit mask. Use 'internal.mkversion' to get more
convenient way of version representation:

```
tarantool> mkversion = require('internal.mkversion')
---
...

tarantool> version = mkversion.from_id(dd_version_id)
---
...

tarantool> version
---
- patch: 0
  minor: 0
  id: 196608
  major: 3
...
```
---

config: automatically set names on schema upgrade

This commit adds a trigger, which sets names via box.cfg, when
dd_version_id, seen from box.schema event, is changed to the
version >= 3.0.0 from version < 3.0.0.

On master instances we set replicaset_name and instance_name.
On replicas - only instance_name, they'll automatically register
themselves in _cluster space by making IPROTO_REGISTER to master.
This behavior allows live upgrade to Tarantool 3.0.0.

Closes https://github.com/tarantool/tarantool/issues/8978

NO_DOC=https://github.com/tarantool/doc/issues/3661


test/config-luatest/upgrade_test.lua
```diff
@@ -0,0 +1,108 @@
+local t = require('luatest')
+local server = require('test.luatest_helpers.server')
+local treegen = require('test.treegen')
+
+local g = t.group('config_upgrade_test')
+
+g.before_all(function()
```
Serpentian, line 7: It's messy, as config tests for some reason doesn't work inside `workdir` if `wal.dir` and `snapshot.dir` is not set. So, manual tweaks are needed to copy xlogs to the same directory, where instances work, and to make luatest's netbox work.

src/box/box.cc
```diff
@@ -1506,6 +1509,22 @@ box_check_node_name(char *out, const char *cfg_name, bool set_diag)
...
+	 */
+	if (dd_version_id != 0 && dd_version_id < MIN_NAMES_VERSION_ID) {
+		say_warn("Setting %s was skipped. Please, consider upgrading "
+			 "schema. If you're using yaml configuration names "
+			 "will be automatically set after schema upgrade",
+			 cfg_name);
+		*out = 0;
+		return 0;
+	}
+
```
Serpentian, line 1527: We should probably also allow skip MISMATCH errors when in snaps there's nil value and force_recovery is set. This will allow to use yaml config for restoring instance, which failed to set names in box.watch trigger.

This doesn't contradict with NoOp rule, if names are set on 2.11, they don't work correctly, even though setting them is allowed without this commit. It's impossible to set them on 3.0, as they're supposed to be already set (saved in `cfg_instance_name` and so on).

changelogs/unreleased/config-automatic-names.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* If Tarantool is congired via yaml file or etcd, then
+  instance/replicaset/cluster names will be automatically set on schema upgrade.
```
andreyaksenov, line 4:
```suggestion
* If Tarantool is congired via a YAML file or etcd, then
  instance/replicaset/cluster names are automatically set on schema upgrade.
```

changelogs/unreleased/dd-version-id-broadcast.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* dd_version_id, which shows the current version of schema, is now broadcasted
+  via box.schema event.
```
andreyaksenov, line 4:
```suggestion
* `dd_version_id` showing the current schema version is now broadcasted
  via the `box.schema` event.
```

andreyaksenov (approved):

src/box/box.cc
Serpentian: Instead of NoOp we may absolutely forbid specifying names, when schema version is < 3.0.0. The problem, is that we'll recover snaps, which can take the significant amount of time, and only after that we'll say user, that it's forbidden

Serpentian: The option I like more - doesn't change the way, names works now at all. Do everything in `box/lua/config`

Serpentian:
Closed, as we rejected this approach

PR #9189, opened 2023-09-28
Add failover coordinator stubs

Totktonada:
This patchset prepares declarative configuration collecting code to be used in the failover coordinator service, adds `--failover` option, adds failover module stubs.

The failover agent is to be implemented as part of Tarantool Enterprise Edition.

Part of https://github.com/tarantool/tarantool-ee/issues/564

### DIFF
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -1102,6 +1102,7 @@ run_script_f(va_list ap)
 	bool bytecode = opt_mask & O_BYTECODE;
 	bool debugging = opt_mask & O_DEBUGGING;
 	bool help_env_list = opt_mask & O_HELP_ENV_LIST;
+	bool failover = opt_mask & O_FAILOVER;
 	/*
 	 * An error is returned via an external diag. A caller
 	 * can't use fiber_join(), because the script can call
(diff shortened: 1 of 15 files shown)


src/lua/init.c
```diff
@@ -1102,6 +1102,7 @@ run_script_f(va_list ap)
 	bool bytecode = opt_mask & O_BYTECODE;
 	bool debugging = opt_mask & O_DEBUGGING;
 	bool help_env_list = opt_mask & O_HELP_ENV_LIST;
+	bool failover = opt_mask & O_FAILOVER;
```
ImeevMA, line 1105: Error in commit message:
> The failover **cooordinator** code is part of Tarantool Enterprise Edition.

How about squash this and the next patch? Also, we can add test for calling --failover in CE here.

src/main.cc
```diff
@@ -656,6 +656,7 @@ print_help(FILE *stream, const char *program)
 		program, program, program, program);
 }
 
+#if !defined(ENABLE_FAILOVER)
```
ImeevMA, line 659: This looks a little strange. What I mean is that we only use this function if `ENABLE_FAILOVER` is 1, and this function exists only in this case. How about inlining this function or removing the `static` from it along with `#if !defined(ENABLE_FAILOVER)`?

ImeevMA (approved):
Hi! Thank you for the patch. I have couple of small comments.

src/main.cc
Totktonada: OK. Added the `fprintf()` call directly into the switch-case.

src/lua/init.c
Totktonada: Fixed the typo.

Squashed the commits.

Added a simple test.

```diff
diff --git a/test/config-luatest/cli_test.lua b/test/config-luatest/cli_test.lua
index 25e311a300..116bdadbd3 100644
--- a/test/config-luatest/cli_test.lua
+++ b/test/config-luatest/cli_test.lua
@@ -131,3 +131,11 @@ g.test_force_recovery = function()
     t.assert_equals(res.exit_code, 0)
     t.assert_equals(res.stdout, 'false')
 end
+
+g.test_failover = function()
+    local opts = {nojson = true, stderr = true}
+    local res = justrun.tarantool('.', {}, {'--failover'}, opts)
+    t.assert_not_equals(res.exit_code, 0)
+    t.assert_equals(res.stderr, '--failover CLI option is available only in ' ..
+        'Tarantool Enterprise Edition')
+end
```

ochaplashkin (approved):

ylobankov (approved):

PR #9208, opened 2023-10-02
config: unify logs and alerts inside creds applier

Lord-KA:
With this patch the following log/alert structure is introduced:
* all informational messages (e.g. some privs granted) use `log.verbose()`
* all warnings about postponed apply use `config:_alert{type = 'warn'}`
* all apply errors are reported with `config:_alert{type = 'error'}`
    
Debug logs are removed.

Part of #8967


Lord-KA:
It was decided, that in case of error, the state machine of config must be set to `check_errors`. It will be implemented in this PR.

ochaplashkin (approved):

Issue #9182, opened 2023-09-27
Forbid downgrading from schema version, which is more, then version of Tarantool


Serpentian:
**Bug description**

The problem here is the fact, that we allow to downgrade from from schema version, which is higher then current Tarantool version. Even though it's impossible to properly downgrade in such case. E.g. in "Steps to reproduce", I downgrade from schema version 3.0.0 to 2.11.0 on Tarantool 2.11.0. It's impossible to check, whether the names are set, e.g. So, all schema changes of Tarantool 3.0 persist, but `box.space._schema` says, that it's 2.11.0 now.

**Steps to reproduce**

1. Run Tarantool 3.0.0:

```
tarantool> box.cfg{replicaset_name = 'replicaset', instance_name = 'instance'}
2023-09-27 05:01:21.079 [4845] main/103/interactive I> Tarantool 3.0.0-alpha3-22-g3240201a2 Linux-x86_64-Debug
<a bunch of logs>
2023-09-27 05:01:21.152 [4845] main/103/interactive/box.load_cfg I> set 'instance_name' configuration option to "instance"
2023-09-27 05:01:21.152 [4845] main/103/interactive/box.load_cfg I> set 'replicaset_name' configuration option to "replicaset"
---
...

tarantool> box.space._schema:select()
---
- - ['replicaset_name', 'replicaset']
  - ['replicaset_uuid', '3945cbf5-7cc2-4e0a-9807-4726aeee2653']
  - ['version', 3, 0, 0]
...
```

2. Run Tarantool 2.11.0 with the same snaps:

```
tarantool> box.cfg{}
2023-09-27 05:03:18.437 [5063] main/103/interactive I> Tarantool 2.11.0-0-g247a9a418 Linux-x86_64-RelWithDebInfo
<a bunch of logs>
---
...

tarantool> box.schema.downgrade('2.11.0')
2023-09-27 05:03:30.843 [5063] main/103/interactive/box.upgrade I> set schema version to 2.11.0
---
...

tarantool> box.space._schema:select()
---
- - ['replicaset_name', 'replicaset']
  - ['replicaset_uuid', '3945cbf5-7cc2-4e0a-9807-4726aeee2653']
  - ['version', 2, 11, 0]
...
```

**Expected behavior**

This behavior should be either forbidden or at least warning message should be written during downgrading.

PR #9209, opened 2023-10-02
downgrade: fix test to downgrade on available version

nshy:
It is not convenient that `test_downgrade_from_more_recent_version` breaks if we create tag for new version and do not add next version to the downgrade versions list. If the version is released we should add it to the list anyway but it is not matter of this test.

Follow up #9182


locker (approved):

ochaplashkin (approved):

locker:
Cherry-picked to 2.11.

PR #9210, opened 2023-10-02
ci add Astra Linux for 2.11

foxzi:


ylobankov:
It is not planned to add support for Astra Linux at this moment.

PR #9225, opened 2023-10-05
ci: add debug_asan_clang workflow

nshy:
Similarly to release_asan_clang but to test debug build. It is also run only under `asan-ci` and `full-ci` labels.

Fiber stack size is 2 times bigger than in the release workflow for luajit tests to pass. Note that this factor is a wild guess.

Part of #7327

PR #9226, opened 2023-10-05
Prepare for debug asan workflow

nshy:
With these changes we can turn on debug asan workflow.

Part of #7327


nshy:
The commits are move to PR https://github.com/tarantool/tarantool/pull/9225

PR #9230, opened 2023-10-06
sql: assign collation to indexes in CREATE TABLE

### LINKED ISSUE #9229: Assertion on create table with multifield indexes and collations
Reproducer:
```
tarantool> box.execute([[CREATE TABLE t0(i INT, a INT, PRIMARY KEY(i, a), b STRING COLLATE "unicode_ci";]])
tarantool: ./src/box/sql/build.c:751: sqlAddCollateType: Assertion `idx->def->key_def->part_count == 1' failed.
Aborted (core dumped)
```
Note, that assertion is only happen on debug build and does not lead to segfault.

Moreover, collation should be added to index, but it is not always so.
How it should be:
```
tarantool> box.execute([[CREATE TABLE t(i INT PRIMARY KEY, a STRING UNIQUE COLLATE "unicode_ci");]])
---
- row_count: 1
...

tarantool> box.space.T.index[1].parts[1].collation
---
- unicode_ci
...
```

However, this doesn't always work:
```
tarantool> box.execute([[CREATE TABLE t(i INT PRIMARY KEY, b INT, a STRING UNIQUE COLLATE "unicode_ci");]])
---
- row_count: 1
...

tarantool> box.space.T.index[1].parts[1]
---
- fieldno: 3
  sort_order: asc
  type: string
  exclude_null: false
  is_nullable: true
...
```

As you can see, no collation is specified.

ImeevMA:
Before this patch, if an index was created due to a column's UNIQUE constraint or a column's PRIMARY KEY constraint before adding a collation, and if the column's fieldno was not equal to the index's position in space->index, the collation would not be assigned to the index.

Also, this patch fixes an assertion in debug build for the case when an index with more that one field was created befo a collation was added.

Closes #9229

### DIFF
--- a/changelogs/unreleased/gh-9229-no-collation-for-indexes.md
+++ b/changelogs/unreleased/gh-9229-no-collation-for-indexes.md
@@ -1,0 +1,6 @@
+## bugfix/sql
+
+* Fixed assertion in a debug build when a collation was added after an index
+  with more than one field (gh-9229).
+* Fixed a bug that in some cases would not assign collations to an index created
+  during `CREATE TABLE` (gh-9229).
--- a/src/box/sql/build.c
+++ b/src/box/sql/build.c
@@ -737,26 +737,32 @@ sql_create_check_contraint(struct Parse *parser, bool is_field_ck)
 void
 sqlAddCollateType(Parse * pParse, Token * pToken)
 {
+	char *coll_name = sql_name_from_token(pToken);
+	uint32_t coll_id;
+	struct coll *coll = sql_get_coll_seq(pParse, coll_name, &coll_id);
+	sql_xfree(coll_name);
+	if (coll == NULL)
+		return;
+
 	struct space *space = pParse->create_column_def.space;
 	assert(space != NULL);
-	uint32_t i = space->def->field_count - 1;
-	char *coll_name = sql_name_from_token(pToken);
-	uint32_t *coll_id = &space->def->fields[i].coll_id;
-	if (sql_get_coll_seq(pParse, coll_name, coll_id) != NULL) {
-		/* If the column is declared as "<name> PRIMARY KEY COLLATE <type>",
-		 * then an index may have been created on this column before the
-		 * collation type was added. Correct this if it is the case.
-		 */
-		for (uint32_t i = 0; i < space->index_count; ++i) {
-			struct index *idx = space->index[i];
-			assert(idx->def->key_def->part_count == 1);
-			if (idx->def->key_def->parts[0].fieldno == i) {
-				coll_id = &idx->def->key_def->parts[0].coll_id;
-				(void)sql_column_collation(space->def, i, coll_id);
-			}
-		}
+	uint32_t fieldno = space->def->field_count - 1;
+	space->def->fields[fieldno].coll_id = coll_id;
+
+	/*
+	 * Indexes that include the current field may have already been created
+	 * before the collation was added. These indexes can be created using
+	 * ... field_name STRING PRIMARY KEY COLLATE ...
+	 * or
+	 * ... field_name STRING UNIQUE COLLATE ...
+	 * In this case, we should add a collation to them.
+	 */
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-9229-no-collation-for-indexes.md
```diff
@@ -0,0 +1,6 @@
+## bugfix/sql
+
+* Fixed assertion in debug build when collation was added after an index with
+  more than one field (gh-9229).
+* Fixed a bug that in some cases would not assign collations to an index created
+  during CREATE TABLE (gh-9229).
```
andreyaksenov, line 6:
```suggestion
* Fixed assertion in a debug build when a collation was added after an index with
  more than one field (gh-9229).
* Fixed a bug that in some cases would not assign collations to an index created
  during `CREATE TABLE` (gh-9229).
```

andreyaksenov (approved):

ochaplashkin (approved):

changelogs/unreleased/gh-9229-no-collation-for-indexes.md
ImeevMA: Thanks, fixed.

Gerold103 (approved):
Thanks for the patch!

src/box/sql/build.c
```diff
@@ -736,26 +736,32 @@ sql_create_check_contraint(struct Parse *parser, bool is_field_ck)
 void
 sqlAddCollateType(Parse * pParse, Token * pToken)
 {
+	char *coll_name = sql_name_from_token(pToken);
```
igormunkin, line 739: Can we use `sql_tt_name_from_token` here? I see the main blocker within `sql_get_coll_seq` that checks `coll_name` against `NULL`. Is this the only reason for not using the static buffer? I believe, we can omit this change in scope of this PR, but it would be nice to make a research/refactoring for such places in SQL code.

src/box/sql/build.c
```diff
@@ -736,26 +736,32 @@ sql_create_check_contraint(struct Parse *parser, bool is_field_ck)
...
+	 * ... field_name STRING PRIMARY KEY COLLATE ...
+	 * or
+	 * ... field_name STRING UNIQUE COLLATE ...
+	 * In this case, we should add a collation to them.
+	 */
+	for (uint32_t iid = 0; iid < space->index_count; ++iid) {
+		struct index *idx = space->index[iid];
+		if (idx->def->key_def->part_count == 1 &&
+		    idx->def->key_def->parts[0].fieldno == fieldno)
+			idx->def->key_def->parts[0].coll_id = coll_id;
```
igormunkin, line 763: Side note: Could you please explain, why `sql_column_collation` is not used as it was before? I see there is a special flow handling collations for ephemeral spaces, but the collation is fetched from the field info.

igormunkin (approved):
@ImeevMA, thanks for the patch! LGTM with a single nit and a couple of questions below.

---

There is a typo in commit message.

> Also, this patch fixes an assertion in debug build for the case when an
> index with more that one field was created befo a collation was added.

Typo: s/befo/before/.

src/box/sql/build.c
ImeevMA: We cannot use sql_tt_name_from_token() because static buffer is limited and it is possible that name length will be more than static buffer size. It does not matter for an errors and comments, but it is important in all other places. Though, there will be added sql_parser_name(), which will allow to use region from parser for temporary allocations, but not in this PR.

ImeevMA:
@igormunkin thank you for the review! Fixed a typo. Answered a questions.

src/box/sql/build.c
igormunkin: Got it, thanks for clarification!

src/box/sql/build.c
igormunkin: Alright. Thanks for explanation!

igormunkin:
Also backported to 2.11 and 2.10 in scope of b215f12 and 719b8d5 respectively.

PR #9238, opened 2023-10-09
fiber: basic api exports

### LINKED ISSUE #9237: Basic ffi api for fiber properties
Currently there's no way to access some basic properties of fibers via ffi api. For instance there's no way to get/set fiber name, get fiber id, get fiber csw, etc. without going through lua api, which is very expensive for such use cases.

Ffi api for this would be ideal

Related #6372

gmoshkin:
Closes #9237

Add exports for fiber_set_name_n, fiber_get_name, fiber_get_id, fiber_get_csw & fiber_find.

@TarantoolBot document
Title: add basic fiber api to ffi exports.

5 basic functions can now be used via ffi api, which were previously only accessible via lua api: fiber_set_name_n, fiber_get_name, fiber_get_id, fiber_get_csw & fiber_find.

### DIFF
--- a/changelogs/unreleased/fiber-basic-api.md
+++ b/changelogs/unreleased/fiber-basic-api.md
@@ -1,0 +1,6 @@
+## feature/fiber
+
+* Exported `fiber_set_name_n`, `fiber_name`, `fiber_id`, `fiber_csw` and
+  `fiber_find` into the public C API and usable via FFI as well.
+* Make `fiber_set_joinable`, `fiber_set_ctx` and `fiber_get_ctx`
+  treat the NULL argument as the current fiber.
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -530,12 +530,16 @@ fiber_make_ready(struct fiber *f)
 void
 fiber_set_ctx(struct fiber *f, void *f_arg)
 {
+	if (f == NULL)
+		f = fiber();
 	f->f_arg = f_arg;
 }
 
 void *
 fiber_get_ctx(struct fiber *f)
 {
+	if (f == NULL)
+		f = fiber();
 	return f->f_arg;
 }
 
@@ -593,6 +597,8 @@ fiber_is_cancelled(void)
 void
 fiber_set_joinable(struct fiber *fiber, bool yesno)
 {
+	if (fiber == NULL)
+		fiber = fiber();
 	if (yesno == true)
 		fiber->flags |= FIBER_IS_JOINABLE;
 	else
@@ -1077,9 +1083,11 @@ fiber_loop(MAYBE_UNUSED void *data)
 }
 
 void
-fiber_set_name(struct fiber *fiber, const char *name)
+fiber_set_name_n(struct fiber *fiber, const char *name, uint32_t len)
 {
-	size_t size = strlen(name) + 1;
+	if (fiber == NULL)
+		fiber = fiber();
+	size_t size = len + 1;
 	if (size <= FIBER_NAME_INLINE) {
 		if (fiber->name != fiber->inline_name) {
 			free(fiber->name);
@@ -1100,6 +1108,30 @@ fiber_set_name(struct fiber *fiber, const char *name)
 	--size;
 	memcpy(fiber->name, name, size);
 	fiber->name[size] = 0;
+}
+
+const char *
+fiber_name(const struct fiber *fiber)
+{
+	if (fiber == NULL)
+		fiber = fiber();
+	return fiber->name;
+}
+
+uint64_t
+fiber_id(const struct fiber *fiber)
+{
+	if (fiber == NULL)
+		fiber = fiber();
+	return fiber->fid;
+}
+
+uint64_t
+fiber_csw(const struct fiber *fiber)
+{
+	if (fiber == NULL)
+		fiber = fiber();
+	return fiber->csw;
 }
 
 static inline void *
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -294,6 +294,7 @@ fiber_start(struct fiber *callee, ...);
  * fiber_start which means no yields.
  *
  * \param f     fiber to set the context for
+ *              if NULL, the current fiber is used
  * \param f_arg context for the fiber function
  */
 API_EXPORT void
@@ -302,6 +303,7 @@ fiber_set_ctx(struct fiber *f, void *f_arg);
 /**
  * Get the context for the fiber which was set via the fiber_set_ctx
  * function. Can be used to avoid calling fiber_start which means no yields.
+ * If \a f is NULL, the current fiber is used.
  *
  * \retval      context for the fiber function set by fiber_set_ctx function
  *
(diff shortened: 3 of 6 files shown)


test/app-tap/module_api.c
```diff
@@ -3296,6 +3296,49 @@ test_box_iproto_override_reset(struct lua_State *L)
 
 /* }}} Helpers for `box_iproto_override` Lua/C API test cases */
 
+static int
+fiber_basic_api_func(va_list va)
+{
+	(void)va;
+	const char *name = "openheimer";
```
ligurio, line 3303: typo: oppenheimer

changelogs/unreleased/fiber-basic-api.md
```diff
@@ -0,0 +1,4 @@
+## feature/fiber
+
+* Exported `fiber_set_name_n`, `fiber_get_name`, `fiber_get_id`,
+  `fiber_get_csw` & `fiber_find` ffi api functions.
```
andreyaksenov, line 4:
```suggestion
* Exported the `fiber_set_name_n`, `fiber_get_name`, `fiber_get_id`,
  `fiber_get_csw`, and `fiber_find` FFI API functions.
```

test/app-tap/module_api.c
gmoshkin: done

changelogs/unreleased/fiber-basic-api.md
gmoshkin: done

ochaplashkin (approved):

src/lib/core/fiber.c
```diff
@@ -1077,9 +1077,11 @@ fiber_loop(MAYBE_UNUSED void *data)
 }
 
 void
-fiber_set_name(struct fiber *fiber, const char *name)
+fiber_set_name_n(struct fiber *fiber, const char *name, uint32_t len)
 {
-	size_t size = strlen(name) + 1;
+	if (fiber == NULL)
+		fiber = fiber();
```
Totktonada, line 1083: I wouldn't do it for some of the functions, it makes the API inconsistent. All the other public API `fiber_*()` functions that accept `struct fiber *` doesn't assume that `NULL` means the current fiber.

src/lib/core/fiber.c
```diff
@@ -1102,6 +1104,30 @@ fiber_set_name(struct fiber *fiber, const char *name)
 	fiber->name[size] = 0;
 }
 
+const char *
+fiber_get_name(struct fiber *fiber)
```
Totktonada, line 1108: AFAIU, `fiber_set_name_n()` is introduced to allow a zero byte inside the name. If it is so, than the `fiber_get_name()` function should accept `uint32_t *len` output parameter. Or don't accept the length parameter in both.

BTW, there is `fiber_name()` named without `get`. I would expose it into the public API under the name without `get`. AFAIR, tarantool's code generally avoid `get` for accessors.

Totktonada: The same for `fiber_get_id()` vs `fiber_id()` and `fiber_get_csw()` vs `fiber_csw()`.

changelogs/unreleased/fiber-basic-api.md
```diff
@@ -0,0 +1,4 @@
+## feature/fiber
+
+* Exported `fiber_set_name_n`, `fiber_get_name`, `fiber_get_id`,
+  `fiber_get_csw`, and `fiber_find` FFI API functions.
```
Totktonada, line 4: I would say 'exported into the public C API and usable with LiaJIT FFI as well'.

Totktonada (approved):
I'm Ok in general, but, please see several style/naming/API comments above.

I asked a second reviewer (@Gerold103) to look into the patch after the fixes.

src/lib/core/fiber.c
gmoshkin: I would really appreciate, if they did. Making a `fiber_self` ffi call each time is both less efficient (impossible to inline) and less ergonomic for the programmer.

What do you think about adding the same functionality to other fiber_* functions?

src/lib/core/fiber.c
gmoshkin: 1.) Actually it's introduced to avoid extra allocations, because for example rust's strings aren't nul-terminated by default. fiber_set_name adds the nul byte anyways and doesn't store the length, so, firstly, there's no need to explicitly return the length, and secondly it would mean doing extra work for computing this length.

2.) I decided to introduce another function in addition to fiber_name, because fiber_name is currently `static inline` and if I export it to ffi, it would no longer be auto inlinable, which may affect the performance. Altough I guess I could just rename the `static inline` version instead

changelogs/unreleased/fiber-basic-api.md
gmoshkin: done

src/lib/core/fiber.h
```diff
@@ -429,6 +430,49 @@ fiber_clock64(void);
...
+API_EXPORT void
+fiber_set_name_n(struct fiber *fiber, const char *name, uint32_t len);
+
+/**
+ * Get fiber name.
+ * @param fiber Target fiber, if it's NULL the current fiber is used.
+ * @return pointer to a nul-terminated string.
+ */
+API_EXPORT const char *
+fiber_get_name(struct fiber *fiber);
```
Gerold103, line 448: I suggest to make all getters take `const struct fiber *` instead of a mutable pointer.

Gerold103 (changes requested):
Thanks for the patch!

src/lib/core/fiber.c
Gerold103: `fiber_name()` is used just in logs, in exceptions, and in Lua C. Generally, if you need a fiber name, then you are not writing a hot performant code block. I suggest to uninline `fiber_name()`. It will change nothing when it comes to perf.

gmoshkin: Removed `_get` from getter functions

src/lib/core/fiber.h
gmoshkin: Done

changelogs/unreleased/fiber-basic-api.md
```diff
@@ -0,0 +1,7 @@
+## feature/fiber
+
+* Exported `fiber_set_name_n`, `fiber_get_name`, `fiber_get_id`,
+  `fiber_get_csw` and `fiber_find` into the public C API
+  and usable via FFI as well.
+* Make `fiber_set_joinable`, `fiber_set_ctx` and `fiber_get_ctx`
+  treat the NULL argument as the current fiber.
```
Gerold103, line 7: The changelog has to be updated with the new export names since the previous review.

Gerold103 (approved):
Thanks for the fixes! Please, update the changelog. Then ok.

changelogs/unreleased/fiber-basic-api.md
gmoshkin: done

Gerold103 (approved):
Thanks for the fixes!

PR #9250, opened 2023-10-11
config: update password hashes after auth_type change

Lord-KA:
User password is stored in a system space is a form of hash when 'chap-sha1' auth type is set, and in a form of hash with salt when 'pap-sha256' is set.

Now, if a user is set inside config, and the current auth type is different from the type the users password is stored in, the password hash will be regenerated.

Part of #8967

NO_DOC=documentation request will be filed manually for the whole
       credentials

### DIFF
--- a/changelogs/unreleased/gh-8967-creds-consider-auth-type.md
+++ b/changelogs/unreleased/gh-8967-creds-consider-auth-type.md
@@ -1,0 +1,5 @@
+## feature/config
+
+* Now a password hash (and salt) will be regenerated for users managed
+  in the configuration file if `security.auth_type` differs from a user's
+  `auth_type` (gh-8967).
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/gh-8967-creds-consider-auth-type.md
```diff
@@ -0,0 +1,5 @@
+## feature/config
+
+* Now the password hash (and salt) will be regenerated for users managed
+  by config, when `security.auth_type` is different from the `auth_type`
+  for the user (gh-8967).
```
andreyaksenov, line 5:
```suggestion
* Now a password hash (and salt) will be regenerated for users managed
  in a configuration file if `security.auth_type` differs from a user's `auth_type`
  (gh-8967).
```

andreyaksenov (approved):

Totktonada (approved):
I see no flaws. Thank you!

changelogs/unreleased/gh-8967-creds-consider-auth-type.md
Lord-KA: Applied.

PR #9258, opened 2023-10-13
test: fixup the 1mops perf test

sergos:
The option --sync was ignored due to a misprint.
The temporary dir was never removed, overflowing the /tmp by default.

NO_TEST=test update
NO_DOC=test update
NO_CHANGELOG=test update

### DIFF
--- a/perf/lua/1mops_write.lua
+++ b/perf/lua/1mops_write.lua
@@ -28,12 +28,25 @@ local params = require('internal.argparse').parse(arg, {
 
 local test_dir = fio.tempdir()
 
+local function rmtree(s)
+    if (fio.path.is_file(s) or fio.path.is_link(s)) then
+        fio.unlink(s)
+        return
+    end
+    if fio.path.is_dir(s) then
+        for _,i in pairs(fio.listdir(s)) do
+            rmtree(s..'/'..i)
+        end
+        fio.rmdir(s)
+    end
+end
+
 local function exit(res, details)
     if (details ~= nil) then
         print(details)
     end
     if test_dir ~= nil then
-        fio.rmdir(test_dir)
+        rmtree(test_dir)
         test_dir = nil
     end
     os.exit(res)
@@ -66,8 +79,8 @@ if params.help or params.h then
     exit(0)
 end
 
--- turn true to test the qsync
-local test_qsync = params.qsync or false
+-- turn true to test the synchronous replication
+local test_sync = params.sync or false
 
 -- number of operations performed by test
 local num_ops = params.ops or 10000000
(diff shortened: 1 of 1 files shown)


perf/lua/1mops_write.lua
```diff
@@ -67,7 +80,7 @@ if params.help or params.h then
 end
 
 -- turn true to test the qsync
-local test_qsync = params.qsync or false
+local test_qsync = params.sync or false
```
ligurio, line 83: Looks like a mess: in comment `qsync` and variable is `test_qsync`, but `params.sync`. Should we rename `qsync` to `sync` everywhere?

perf/lua/1mops_write.lua
```diff
@@ -28,12 +28,25 @@ local params = require('internal.argparse').parse(arg, {
 
 local test_dir = fio.tempdir()
 
+local function rmtree(s)
```
ligurio, line 31: Please add a check for root directory (`/`). It would be very-very bad if we will do `rm -rf /` on user's machine.

ligurio (commented):
Thanks for the patch!

perf/lua/1mops_write.lua
sergos: donw

perf/lua/1mops_write.lua
sergos: It's a local function, never exposed to the user. The only use of the function is to remove the contents of the `tmp_dir` generated by the `fio.tempdir()`. Hence, no check is necessary.

perf/lua/1mops_write.lua
ligurio: Thanks!

perf/lua/1mops_write.lua
ligurio: Right, thanks!

ligurio (approved):
LGTM

igormunkin (approved):
@sergos, thanks for the patch! I'd rather implement the changeset via two separate patches, but I do not insist on it, hence LGTM.

perf/lua/1mops_write.lua
```diff
@@ -28,12 +28,25 @@ local params = require('internal.argparse').parse(arg, {
...
+    end
+end
+
 local function exit(res, details)
     if (details ~= nil) then
         print(details)
     end
     if test_dir ~= nil then
-        fio.rmdir(test_dir)
+        rmtree(test_dir)
```
sergepetrenko, line 49: There's `fio.rmtree()` and it seems to be doing the same thing: https://www.tarantool.io/en/doc/latest/reference/reference_lua/fio/#lua-function.fio.rmtree
Can you use it here?

sergepetrenko: Ok, I see now. Thanks!

Issue #9112, opened 2023-09-07
Introduce variations of DROP CONSTRAINT


ImeevMA:
Since names for different types of constraints do not have the same namespace, let's introduce
```ALTER TABLE table_name DROP CONSTRAINT FOREIGN KEY tuple_foreign_key_constraint_name;```
```ALTER TABLE table_name DROP CONSTRAINT UNIQUE tuple_check_constraint_name;```
```ALTER TABLE table_name DROP CONSTRAINT CHECK tuple_unique_constraint_name;```
```ALTER TABLE table_name DROP CONSTRAINT FOREIGN KEY field_name.field_foreign_key_constraint_name;```
```ALTER TABLE table_name DROP CONSTRAINT UNIQUE field_name.field_check_constraint_name;```
```ALTER TABLE table_name DROP CONSTRAINT CHECK field_name.field_unique_constraint_name;```

Also, maybe we should drop original `ALTER TABLE DROP CONSTRAINT`.

PR #9259, opened 2023-10-13
box: drop unused SQL constraint-related code

ImeevMA:
This patch removes unused code that was previously used to check name uniqueness for SQL constraints. Since SQL now uses the BOX constraint mechanism, this code becomes unnecessary.

Follow-up #9112


Gerold103 (approved):
Thanks for the patch!

alyapunov (approved):

PR #9267, opened 2023-10-16
console: forward original URI to net.box when connecting over IPROTO

locker:
Tarantool supports two console protocols: text and binary. The binary protocol is implemented with IPROTO EVAL request so the console module reuses the net.box module to establish and maintain a binary connection. Currently, instead of passing the original URI specified by the user to net.box.connect as is, the console module parses the URI and passes the host and port. As a result, extra information that may be specified in URI parameters is lost. This prevents the user from connecting to the binary console using the SSL transport because to use the SSL transport the user must specify transport=ssl URI parameter.

Needed for tarantool/tarantool-ee#567


nshy (approved):

locker:
Cherry-picked to 2.11.

PR #9269, opened 2023-10-17
replication: improve name conflict error message

### LINKED ISSUE #8692: Confusing cluster/replicaset/instance name mismatch error
3.0.0-entrypoint-259-g13159230c

Start an instance and stop it:

```sh
$ tarantool
tarantool> box.cfg({cluster_name = 'foo'})
<...>
tarantool> Ctrl+D
```

Start it again with another `cluster_name`:

```sh
$ tarantool
tarantool> box.cfg({cluster_name = 'bar'})
<...>
<...> box.cc:4572 E> ER_CLUSTER_NAME_MISMATCH: Cluster name mismatch: expected bar, got foo
<...> F> can't initialize storage: Cluster name mismatch: expected bar, got foo
```

'Expected' and 'got' values are mixed up.

The same for `replicaset_name` and `instance_name`.

sergos:
The error message was not quite clear in case instance has any name defined and config provides a different one.

Closes #8692

NO_CHANGELOG=no previous release with this error message
NO_DOC=not needed since no changes to the public API

### DIFF
--- a/src/box/errcode.h
+++ b/src/box/errcode.h
@@ -319,10 +319,10 @@ struct errcode_record {
 	/*264 */_(ER_NIL_UUID,			"Nil UUID is reserved and can't be used in replication") \
 	/*265 */_(ER_WRONG_FUNCTION_OPTIONS,	"Wrong function options: %s") \
 	/*266 */_(ER_MISSING_SYSTEM_SPACES,	"Snapshot has no system spaces") \
-	/*267 */_(ER_CLUSTER_NAME_MISMATCH,	"Cluster name mismatch: expected %s, got %s") \
-	/*268 */_(ER_REPLICASET_NAME_MISMATCH,	"Replicaset name mismatch: expected %s, got %s") \
+	/*267 */_(ER_CLUSTER_NAME_MISMATCH,	"Cluster name mismatch: name '%s' provided in config confilcts with the instance one '%s'") \
+	/*268 */_(ER_REPLICASET_NAME_MISMATCH,	"Replicaset name mismatch: name '%s' provided in config confilcts with the instance one '%s'") \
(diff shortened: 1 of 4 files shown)


src/box/errcode.h
```diff
@@ -319,10 +319,10 @@ struct errcode_record {
 	/*264 */_(ER_NIL_UUID,			"Nil UUID is reserved and can't be used in replication") \
 	/*265 */_(ER_WRONG_FUNCTION_OPTIONS,	"Wrong function options: %s") \
 	/*266 */_(ER_MISSING_SYSTEM_SPACES,	"Snapshot has no system spaces") \
-	/*267 */_(ER_CLUSTER_NAME_MISMATCH,	"Cluster name mismatch: expected %s, got %s") \
-	/*268 */_(ER_REPLICASET_NAME_MISMATCH,	"Replicaset name mismatch: expected %s, got %s") \
+	/*267 */_(ER_CLUSTER_NAME_MISMATCH,	"Cluster name mismatch: name '%s' provided in config confilcts with the instance one '%s'") \
```
Totktonada, line 322: > NO_CHANGELOG=internal change
> NO_DOC=internal change

I agree that it doesn't require a documentation request, but this is a user visible change, not internal one.

So, I would say that the patch doesn't change the public API (an error message is not a part of the API) and so shouldn't be documented.

Regarding changelog: there was no release with these error messages, so there is no sense to mention the intermediate change in the release notes. The same point here: I agree to don't add the changelog entry, but the reason is different.

Totktonada (approved):

Gerold103 (approved):
Thanks for the patch!

ochaplashkin (approved):

src/box/errcode.h
sergos: updated

PR #9274, opened 2023-10-18
box: fail safely if tuple limit in hash index is reached

### LINKED ISSUE #3594: Fail safely if tuple limit is reached
If tuple limit (which is around 2B right now) is reached, Tarantool segfaults like this:
```
0  0x00000000004787e3 in matras_view_get_no_check (v=0x227b688, id=4294967295, m=0x227b688) at /var/tmp/portage/dev-db/tarantool-1.6.9-r9999/work/tarantool-1.6.9.80/src/lib/small/small/matras.h:353
#1  matras_get (id=4294967295, m=0x227b688) at /var/tmp/portage/dev-db/tarantool-1.6.9-r9999/work/tarantool-1.6.9.80/src/lib/small/small/matras.h:372
#2  light_index_insert (value=<optimized out>, hash=1314228454, ht=0x227b670) at /var/tmp/portage/dev-db/tarantool-1.6.9-r9999/work/tarantool-1.6.9.80/src/lib/salad/light.h:786
```

Need to reports a error and exit normally.

Gumix:
Closes #3594


p7nov (approved):

changelogs/unreleased/gh-3594-fail-safely-if-tuple-limit-is-reached.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/core
+
+* Fixed a potential data corruption when the number of tuples in a hash index
```
locker, line 3: Don't we have the same problem with the tree index?

src/box/memtx_hash.cc
```diff
@@ -371,16 +371,14 @@ memtx_hash_index_replace(struct index *base, struct tuple *old_tuple,
...
 			light_index_delete(hash_table, pos);
 			pos = light_index_end;
 		});
 
 		if (pos == light_index_end) {
-			diag_set(OutOfMemory,
-				 (ssize_t)light_index_count(hash_table),
-				 "hash_table", "key");
+			diag_set(ClientError, ER_HASH_INDEX_REPLACE_FAILED,
+				 (unsigned)light_index_count(hash_table));
```
locker, line 381: This doesn't necessarily happen because of the matras limit. It also happens when we reach the memtx arena limit. So I think that the error type should remain `OutOfMemory`. Also, changing the error type would be inconsistent with the BPS tree implementation where we [raise](https://github.com/tarantool/tarantool/blob/0ea1ba877b8e62e0acb558b55623d93bf58a0e1f/src/box/memtx_tree.cc#L1135) `OutOfMemory`.

I don't see much point in including the index size into the error message because the user can check it out with `index.count` when they investigate. Anyway, if we decide to add it, it's better to include this information as a [payload](https://github.com/tarantool/tarantool/blob/0ea1ba877b8e62e0acb558b55623d93bf58a0e1f/src/lib/core/diag.h#L146), and this should be done consistently for all memtx indexes.

changelogs/unreleased/gh-3594-fail-safely-if-tuple-limit-is-reached.md
Gumix: Yes, but the limit for the tree index is way bigger:
- https://github.com/tarantool/tarantool/issues/3594#issuecomment-576360867

src/box/memtx_hash.cc
Gumix: OK, but what number of bytes to return with `OutOfMemory` here?

locker: `MEMTX_EXTENT_SIZE`, like in the tree index.

Gumix: Done.

Gumix:
Divided into 2 commits: submodule bump + changelog.

locker (approved):

locker:
Cherry-picked to 2.11.

PR #9288, opened 2023-10-23
main: rewrite help message

Totktonada:
This commit aims several goals.

First, we're going to recommend `tt` as a tool to run and manage tarantool instances. The tool is going to replace `tarantoolctl` and it seems valuable to sync recipes across different parts of the documentation (including this help message).

Second, the action options and modifier options are separated now. At least one action option should be specified. I hope, it is more obvious from the help message now.

Third, single line descriptions are replaced with a bit more extended ones to give a user basic context to understand an influence of an option.

Updated the man page accordingly: synchronized the options, removed `tarantoolctl`.

Part of #8862


src/main.cc
```diff
@@ -620,41 +620,117 @@ print_version(void)
...
-		"Please visit project home page at https://tarantool.org\n"
-		"to see online documentation, submit bugs or contribute a patch.\n"
-		;
-	fprintf(stream, help_msg, tarantool_version(),
-		program, program, program, program);
+	static const char help_msg[] = "Tarantool %s\n"
+		"\n"
+		"----------------------------------------------------------\n"
+		"\n"
+		"The recommended way to work with Tarantool is to use tt\n"
```
igormunkin, line 629: Minor: It's worth to add the link to `tt` documentation in case more info is needed.

src/main.cc
```diff
@@ -620,41 +620,117 @@ print_version(void)
...
+		"\n"
+		"    The file may be pointed as `-` to read the script from\n"
+		"    stdin.\n"
+		"\n"
+		"-e <..code..>\n"
+		"\n"
+		"    Run a Lua code.\n"
+		"\n"
+		"    This option can be used together with other action\n"
+		"    options or on its own.\n"
```
igormunkin, line 688: Minor: It's worth to mention, that `-e` can be also used as a *modifier*. Feel free to ignore.

igormunkin (approved):
@Totktonada, thanks for the patch! Generally, LGTM: I have no objections regarding splitting the option into actions/modifiers groups, hope it'll help.

I also decided to drop a reminder right here about updating the corresponding man page. And please consider a couple of nits below either.

src/main.cc
Totktonada: I'll clarify it a bit.

```diff
diff --git a/src/main.cc b/src/main.cc
index ae612baca1..33818950c7 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -684,8 +684,8 @@ print_help(FILE *stream)
                "\n"
                "    Run a Lua code.\n"
                "\n"
-               "    This option can be used together with other action\n"
-               "    options or on its own.\n"
+               "    It can be used on its own (as an action option) or\n"
+               "    together with another action option (as a modifier).\n"
                "\n"
                "-i\n"
                "\n"
```

src/main.cc
Totktonada: It looks as a good idea on the first glance, but if I'll do it, it looks strange that we link `tt`'s documentation directly, but not `tarantool`'s documentation: CLI and configuration options, deploy guides and so on. https://tarantool.io contains all the topics and we link it on the bottom of the help message.

I think that the actual problem is that `tt`'s documentation is burried too deep into the website structure and we should place it to a more visible place on the website.

I would leave the help message it as is for now, because I don't know how to make the links informative and consistent without changes on the website.

src/main.cc
Totktonada: Done.

src/main.cc
```diff
@@ -620,41 +620,117 @@ print_version(void)
...
+		"----------------------------------------------------------\n"
+		"\n"
+		"At the lower level there are the following action options.\n"
+		"\n"
+		"-h\n"
+		"--help\n"
+		"\n"
+		"    Print this help message.\n"
+		"\n"
+		"-v\n"
```
Totktonada, line 657: Forgot about the `-V` variant of the option.

Totktonada: Fixed.

Totktonada:
@igormunkin I've synchronized all the options in the the man page with the help message.

src/main.cc
Totktonada: I changed my mind after discussions with Igor and the documentation team. Added the link in PR #9377.

PR #9290, opened 2023-10-23
log: make `log.cfg{modules=...}` work as `box.cfg{log_modules=...}`

### LINKED ISSUE #7962: `log.cfg{modules = ...}` isn't equivalent to `box.cfg{log_modules = ...}`
#7672 allows to specify a log level for each module individually, however there is some inconsistency:
* `box.cfg{log_level_modules={...}}` overrides the existing list of modules in `log.cfg`, while
* `log.cfg{level_modules={...}}` appends the specified modules to the existing list.

Also it's not possible to remove a single module from the config by setting it to `box.NULL`.

[`gh_3211_per_module_log_level_test.lua` contains the corresponding test-cases under `TODO`](https://github.com/tarantool/tarantool/blob/4557608841c672dfe5c0899277ab3d41cd4d74da/test/box-luatest/gh_3211_per_module_log_level_test.lua#L111).

@nshy:
> I would also check that resetting works if we first call `log.cfg` and then `box.cfg` (first `box.cfg` call, reset here smth from `level_modules`). I bet it will not work now. Unfortunately it is not easy to test with `server:exec` as it calls `box.cfg` on start.

The fix of this issue requires thorough changes in [merge_cfg()](https://github.com/tarantool/tarantool/blob/f2f40ff711a30b3a9e16ef965c99eb68b368f125/src/box/lua/load_cfg.lua#L645), etc.

locker:
Configuring log modules work differently with `log.cfg` and `box.cfg`: `box.cfg{log_modules=...}` overwrites the current config completely while `log.cfg{modules=...}` overwrites the currently config only for the specified modules. Let's fix this inconsistency by making `log.cfg` behave exactly as `box.cfg`.

Closes #7962

### DIFF
--- a/changelogs/unreleased/gh-7962-log-cfg-modules-fix.md
+++ b/changelogs/unreleased/gh-7962-log-cfg-modules-fix.md
@@ -1,0 +1,6 @@
+## bugfix/core
+
+* Fixed the behavior of `log.cfg{modules = ...}`. Now, instead of merging the
+  new log modules configuration with the old one, it completely overwrites the
+  current configuration, which is consistent with `box.cfg{log_modules = ...}`
+  (gh-7962).
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -812,11 +812,10 @@ local function check_cfg_option_type(template, name, value)
     end
 end
 
-local function prepare_cfg(cfg, default_cfg, template_cfg, modify_cfg)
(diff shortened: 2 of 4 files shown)


nshy (approved):

p7nov (approved):

ylobankov (approved):

locker:
Cherry-picked to 2.11.

PR #9297, opened 2023-10-24
box: fix memory leak when dropping temporary spaces

### LINKED ISSUE #9296: Memory leak when dropping fully-temporary spaces
**Bug description**

When dropping fully-temporary spaces the memory is leaking, because on_commit triggers aren't run for NOOP txn statements.

To reproduce do the following in the console:
```lua
box.cfg()
require'compat'.fiber_slice_default = 'old'
for i=1,70000 do box.schema.space.create('s', { type = 'temporary' }):drop() end
```

You will get this error message:
```yaml
---
- error: 'Tuple format limit reached: 65536'
...
```

This means that the number of live tuple format instances has overflown the capacity. This doesn't happen, if you instead create & drop spaces with `type = 'data-temporary'`, which means that this is a leak.

This is happening because the actual structs are supposed to be freed in the `txn_stmt::on_commit` trigger, but it will only be invoked if it's first transferred into `txn::on_commit`, which only happens when `txn_journal_entry_new` is run. But because dropping a temporary space is considered to be a NOOP (no wal entry is needed), this function is never run after `txn_commit_noop` results in a early exit.

gmoshkin:
Before this fix we used to leak memory when dropping fully-temporary
spaces. The problem was that the memory is only freed in the on_commit
callback of the transaction, but dropping a temporary space is a NOP
(meta-data is not persisted), and the on_commit triggers used to not be
invoked for NOP transactions.

Closes https://github.com/tarantool/tarantool/issues/9296

NO_DOC=bug fix

### DIFF
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -762,6 +762,11 @@ txn_complete_success(struct txn *txn)
 	assert(!txn_has_flag(txn, TXN_WAIT_SYNC));
 	assert(txn->signature >= 0);
 	assert(in_txn() == txn);
+#ifndef NDEBUG
+	struct txn_stmt *stmt;
+	stailq_foreach_entry(stmt, &txn->stmts, next)
+		assert(!stmt->has_triggers || rlist_empty(&stmt->on_commit));
+#endif
 	txn->status = TXN_COMMITTED;
 	if (txn->engine != NULL)
 		engine_commit(txn->engine, txn);
@@ -1018,6 +1023,19 @@ static bool
 txn_commit_nop(struct txn *txn)
 {
 	if (txn->n_new_rows + txn->n_applier_rows == 0) {
+		/*
+		 * A transcation without any new rows may still have
+		 * on_commit triggers, for example if it's a ddl operation
+		 * for a fully-temporary space.
+		 */
+		struct txn_stmt *stmt;
+		stailq_foreach_entry(stmt, &txn->stmts, next) {
+			if (stmt->has_triggers) {
+				txn_init_triggers(txn);
+				rlist_splice(&txn->on_commit, &stmt->on_commit);
+			}
+		}
+
 		txn->signature = TXN_SIGNATURE_NOP;
 		txn_complete_success(txn);
 		fiber_set_txn(fiber(), NULL);
--- a/test/box-luatest/fully-temporary_spaces_test.lua
+++ b/test/box-luatest/fully-temporary_spaces_test.lua
@@ -520,3 +520,40 @@ g.test_meta_data_not_persisted = function()
 
     check_wal_and_snap({})
 end
+
+-- check memory is not leaked when dropping temporary spaces
+g.test_no_memory_leak_after_drop = function()
+    g.server:exec(function()
+        -- Create a `weak_refs` table, with a weak reference to `object`
+        local object = {'object'}
+        local weak_refs = setmetatable({object}, {__mode = 'v'})
+
+        local s = box.schema.space.create('temp', { type = 'temporary' })
+
+        -- Create a strong reference to the same `object` which belongs to
+        -- the on_replace trigger of the temporary space.
+        s:on_replace((function()
+            local strong_ref = object
+            return function() print(strong_ref) end
+        end)())
+
+        -- Delete the original strong reference to `object`
+        object = nil
+
+        -- Trigger garbage collection (needs to be called twice to collect
+        -- memory freed by __gc methods invoked on the first call).
+        collectgarbage()
+        collectgarbage()
+
+        -- on_replace trigger is still live, so `object` still exists.
+        t.assert(weak_refs[1])
+
+        -- space.drop should free the on_replace triggers
+        s:drop()
+        collectgarbage()
+        collectgarbage()
+
+        -- on_replace trigger was destroyed and so was `object`
+        t.assert(not weak_refs[1])
+    end)
+end
--- a/changelogs/unreleased/gh-9296-fix-memory-leak-when-dropping-fully-temporary-spaces.md
+++ b/changelogs/unreleased/gh-9296-fix-memory-leak-when-dropping-fully-temporary-spaces.md
@@ -1,0 +1,3 @@
+## bugfix/box
+
+* Fixed a memory leak when dropping fully-temporary spaces (gh-9296).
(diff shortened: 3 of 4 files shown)


ochaplashkin (approved):

drewdzzz (changes requested):
Thank you for the patch!

I have no objections regardless the idea of your solution, but there are couple details to think about while we are here. 

Since the problem is that `txn_complete_success` expects all triggers from `stmt->on_commit` to be moved to `txn->on_commit`, let's describe this invariant in the function description. Also, let's add a check that iterates over all the statements and checks that its list of `on_commit` triggers is empty, but only in Debug mode.

Also, please, write an appropriate commit message. The issue does a good job of describing the problem and its details, but you need to do the same in the commit message, but in less detail - sometimes it's not very handy to go to an issue, for example when you're blaming the code locally. The commit message should also reflect the motivation behind the chosen solution - you could iterate over all statements right in `txn_complete_success`, it would be a more "general" solution, but rather expensive because it is hot path.

And one last thing. I'm sorry that no one has reviewed this patch in a long time, but now you need to write changelog since the feature was released.

src/box/txn.c
```diff
@@ -963,6 +963,14 @@ static bool
 txn_commit_nop(struct txn *txn)
 {
 	if (txn->n_new_rows + txn->n_applier_rows == 0) {
+		struct txn_stmt *stmt;
```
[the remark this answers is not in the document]
gmoshkin: Done

xuniq (approved):

gmoshkin:
>  Also, let's add a check that iterates over all the statements and checks that its list of `on_commit` triggers is empty, but only in Debug mode.

I tried that, the problem is during bootstrap `on_commit` structs aren't properly initialized and instead are overwritten with `0x505050...`, so it's not safe to call `rlist_empty` on them.

Am I supposed to go refactor the bootstrap code? Or can I just drop the debug assertion part?

drewdzzz:
Struct `txn_stmt` has `txn->has_triggers` flag - it is set iff its triggers lists are initialized.

gmoshkin:
> Struct `txn_stmt` has `has_triggers` flag - it is set iff its trigger lists are initialized.

Yes, thanks, I figured it out

src/box/txn.c
```diff
@@ -755,13 +755,25 @@ txn_complete_fail(struct txn *txn)
 	rmean_collect(rmean_box, IPROTO_ROLLBACK, 1);
 }
 
+/**
+ * Commit and finalize the transaction.
+ *
+ * This function will run any on_commit triggers associated with @a txn and only
+ * them, so if any substatements of this transaction have on_commit triggers,
```
drewdzzz, line 762: We write descriptions next to function declarations, not definitions - please, move the description there.
Also, `substatements` -> `statements`.

drewdzzz (changes requested):
Thanks for your fixes, I've got one last nitpick.

Also, please, write an appropriate PR description (copy-pasting commit message would be OK).

src/box/txn.c
gmoshkin: done

drewdzzz (approved):
LGTM, thank you for the patch!

test/box-luatest/fully-temporary_spaces_test.lua
```diff
@@ -520,3 +520,40 @@ g.test_meta_data_not_persisted = function()
...
+            local strong_ref = object
+            return function() print(strong_ref) end
+        end)())
+
+        -- Delete the original strong reference to `object`
+        object = nil
+
+        -- Trigger garbage collection (needs to be called twice to collect
+        -- memory freed by __gc methods invoked on the first call).
+        collectgarbage()
```
sergepetrenko, line 545: Nit: the comment and an extra call to `collectgarbage()` are only needed in the second case, when you drop the space. The "double collectgarbage()" thing doesn't apply to normal lua objects.

sergepetrenko (changes requested):

test/box-luatest/fully-temporary_spaces_test.lua
gmoshkin: can something bad happen because of the additional `collectgarbage()` call?

sergepetrenko: No, but it's confusing. And the comment describes a situation which's not possible in this place. The `object` doesn't have a `__gc` metamethod (and can't possibly have one AFAICS). 

The comment and double `collectgarbage` both belong to the second place. Altough even there they seem unnecessary. The test continues working 100% the same when I remove the extra `collectgarbage()` calls:
```diff
diff --git a/test/box-luatest/fully-temporary_spaces_test.lua b/test/box-luatest/fully-temporary_spaces_test.lua
index 783fd42f2f..e7f75bfc71 100644
--- a/test/box-luatest/fully-temporary_spaces_test.lua
+++ b/test/box-luatest/fully-temporary_spaces_test.lua
@@ -543,7 +543,6 @@ g.test_no_memory_leak_after_drop = function()
         -- Trigger garbage collection (needs to be called twice to collect
         -- memory freed by __gc methods invoked on the first call).
         collectgarbage()
-        collectgarbage()
 
         -- on_replace trigger is still live, so `object` still exists.
         t.assert(weak_refs[1])
@@ -551,7 +550,6 @@ g.test_no_memory_leak_after_drop = function()
         -- space.drop should free the on_replace triggers
         s:drop()
         collectgarbage()
-        collectgarbage()
 
         -- on_replace trigger was destroyed and so was `object`
         t.assert(not weak_refs[1])
```

gmoshkin: The point of the double `collectgarbage` call is that in Lua it is often not enough to just call it once because after the first call the memory may become unused but not yet freed (for the reason explained in the comment). And because the test is checking for memory leaks, it's trying to do a best effort at freeing that memory.

I agree that currently these calls are redundant and removing them doesn't change anything, but I do believe it is possible that at some point in the future something may change and these calls will become necessary. So if there's nothing bad that can happen because of the double collectgarbage calls, I'd rather leave them in the test.

sergepetrenko: Ok, let's keep it your way.

sergepetrenko (approved):
Thanks for the fix!

sergepetrenko:
Backport:
3.0 https://github.com/tarantool/tarantool/pull/9941

PR #9311, opened 2023-10-27
fiber: make the concurrent fiber_join safer

### LINKED ISSUE #7562: Concurrent C API `fiber_join()` is unsafe
**Bug description**

If two or more fibers are yielding in `fiber_join()`, one of them will eventually join and recycle the fiber, while the rest will crash on accessing the recycled fiber's struct, or hang, or totally foreign fiber will be joined. There is no simple solution for this. [More info is here](https://github.com/tarantool/tarantool/pull/7513#discussion_r942985411) and in #7537. The same issue in Lua is addressed in #7531, this ticket is about C API.

**Steps to reproduce**

1. Apply this patch:
```diff
diff --git a/test/unit/fiber.cc b/test/unit/fiber.cc
index 09ef30cf1..19e4debac 100644
--- a/test/unit/fiber.cc
+++ b/test/unit/fiber.cc
@@ -129,6 +129,14 @@ waker_f(va_list ap)
    return 0;
 }

+static int
+joiner_f(va_list ap)
+{
+   struct fiber *fiber = (struct fiber *)fiber()->f_arg;
+   fiber_join(fiber);
+   return 0;
+}
+
 static void
 fiber_join_test()
 {
@@ -182,6 +190,18 @@ fiber_join_test()
    fiber_cancel(fiber);
    fiber_join(fiber);

+   /*
+    * Try to join a fiber by 2 concurrent joiners.
+    */
+   struct fiber *joinee = fiber_new_xc("joinee", noop_f);
+   struct fiber *joiner = fiber_new_xc("joiner", joiner_f);
+   joiner->f_arg = joinee;
+   fiber_set_joinable(joinee, true);
+   fiber_wakeup(joinee);
+   fiber_wakeup(joiner);
+   /* This fiber_join will crash. */
+   fiber_join(joinee);
+
    footer();
 }
```
2. Build Tarantool
3. Run `./test-run.py --builddir ../build unit/fiber.test`

**Expected behavior**

By way of example, `man pthread_join` says:
> If multiple threads simultaneously try to join with the same thread, the results are undefined.

mkostoevr:
Prior to this patch a bunch of illegal conditions were possible:
1. The joinability of a fiber could be changed while the fiber is being joined by someone. This could lead to double recycling: the first one happened on the fiber finish, and the second one in the fiber join.
2. The joinability of a dead joinable fiber could be altered, this led to inability jo join the dead fiber and free its resources.
3. A running fiber could be joined concurrently by two or more fibers, so the fiber could be recycled more than once (once per each concurrent join).
4. A dead joined fiber could be made joinable and joined once more leading to a double recycle.

Fixed these issues by adding a new FIBER_IS_JOINED flag and using it where appropriate: now the `fiber_join` detects the double join and joinability change requests are ignored in some cases.

It's still possible that a fiber join is performed on a struct which has been recycled and, if the new fiber is joinable too, this can't be detected. The current fiber API does not allow to fix this, so this is to be the user's responsibility, they should be warned about the fact the double join to the same fiber is illegal.

Closes #7562

@TarantoolBot document
Title: `fiber_join` and `fiber_set_joinable` behave differently now.

`fiber_join` now detects a join of a joined fiber in case if the fiber struct wasn't reused. If the double join is detected the function raises an error.

`fiber_set_joinable` only changes the fiber joinability if it's alive and is not yet joined, othervice it does nothing. This prevents some amount of error conditions that could happen when using the API in an unexpected way:
- Making a dead joinable fiber non-joinable could lead to a memory leak: one can't join the fiber anymore.
- Making an alive joined fiber non-joinable would lead to the double free: once on the fiber function finish, and secondly in the active fiber join finish.
- Making a dead joined fiber joinable allowed to join the fiber once again leading to a double free.

Any given by the API `struct fiber` should only be joined once. If a fiber is joined after the first join on it has finished the behavior is undefined: it can either be a runtime error or an incidental join to a totally foreign fiber.

### DIFF
--- a/changelogs/unreleased/gh-7562-fiber_join.md
+++ b/changelogs/unreleased/gh-7562-fiber_join.md
@@ -1,0 +1,6 @@
+## bugfix/core
+
+* Fixed a possible inconsistent state entering if fibers are joined incorrectly.
+  Now the `fiber_set_joinable` function panics if the fiber is dead or joined
+  already. The `fiber_join` and `fiber_join_timeout` functions now panic on a
+  double join if it is possible to detect it (gh-7562).
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -664,6 +664,56 @@ fiber_set_joinable(struct fiber *fiber, bool yesno)
 {
 	if (fiber == NULL)
 		fiber = fiber();
+
+	/*
+	 * The C fiber API does not allow to report any error to the caller. At
+	 * the same time using the function in some conditions is unsafe. So in
+	 * case if such a condition is met the function panics.
+	 *
+	 * Here're the conditions the function may be called in:
+	 * 1. The fiber is not joinable and hasn't finished yet. We're good to
+	 *    change the fiber's joinability.
+	 * 2. The fiber is not joinable and is finished. That means the fiber
+	 *    is recycled already, so this is a use after free. This case is
+	 *    impossible from the Lua world, but can be done with C API, so it
+	 *    is to be prohibited.
+	 * 3. The fiber is joinable, not finished and not joined. We're good to
+	 *    change its joinability.
+	 * 4. The fiber is joinable, not finished, but joined already. Making
+	 *    the fiber non-joinable will lead to a double free: one in the
+	 *    fiber_loop, the second one in the active fiber_join_timeout.
+	 *    Making it joinable is a sign of attempt to join it later, so we
+	 *    can expect the second join in the future - detect it now, while
+	 *    we can do it, because in the future the fiber struct may be
+	 *    reused and it will be much more difficult to find the root cause
+	 *    of the double join (or we won't be able to detect it at all).
+	 * 5. The fiber is joinable, finished and not joined. The fiber should
+	 *    be joined in order to be recycled, so making it non-joinable can
+	 *    lead to a fiber leak and to be prohibited. The point about making
+	 *    it joinable from the statement #4 is applied here too.
+	 * 6. The fiber is joinable, finished and joined. In order to avoid
+	 *    introducing a new fiber flag, a joined fiber becomes non-joinable
+	 *    on death (see the fiber_join_timeout implementation), so this
+	 *    condition transforms into #2. And it's a use after free too.
+	 *
+	 * Simplifying:
+	 *   OK:
+	 *     1: !is_joinable && !is_dead
+	 *     3: is_joinable && !is_dead && !is_joined
+	 *   NOT OK:
+	 *     2: !is_joinable && is_dead
+	 *     4: is_joinable && !is_dead && is_joined
+	 *     5: is_joinable && is_dead && !is_joined
+	 *     6. is_joinable && is_dead && is_joined
+	 *
+	 *  Othervice, OK: !is_dead && (!is_joinable || !is_joined)
+	 *  So NOT OK: is_dead || (is_joinable && is_joined)
+	 *  So NOT OK: is_dead || is_joined, because joined implies joinable.
+	 */
+	if ((fiber->flags & FIBER_IS_DEAD) != 0 ||
+	    (fiber->flags & FIBER_JOIN_BEEN_INVOKED) != 0)
+		panic("%s on a dead or joined fiber detected", __func__);
+
 	if (yesno == true)
 		fiber->flags |= FIBER_IS_JOINABLE;
 	else
@@ -732,6 +782,12 @@ fiber_join_timeout(struct fiber *fiber, double timeout)
 	if ((fiber->flags & FIBER_IS_JOINABLE) == 0)
 		panic("the fiber is not joinable");
 
+	if ((fiber->flags & FIBER_JOIN_BEEN_INVOKED) != 0)
+		panic("join of a joined fiber detected");
+
+	/* Prohibit joining the fiber and changing its joinability. */
+	fiber->flags |= FIBER_JOIN_BEEN_INVOKED;
+
 	if (!fiber_is_dead(fiber)) {
 		double deadline = fiber_clock() + timeout;
 		while (!fiber_wait_on_deadline(fiber, deadline) &&
@@ -752,6 +808,11 @@ fiber_join_timeout(struct fiber *fiber, double timeout)
 	}
 	assert((fiber->flags & FIBER_IS_RUNNING) == 0);
 	assert((fiber->flags & FIBER_IS_JOINABLE) != 0);
+	/*
+	 * This line is here just to make the following statement true: if
+	 * a fiber is dead and is not joinable, that means it's recycled.
+	 * So we don't have to introduce a new FIBER_IS_RECYCLED flag.
+	 */
 	fiber->flags &= ~FIBER_IS_JOINABLE;
 
 	/* Move exception to the caller */
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -170,6 +170,11 @@ enum {
 	 * This flag idicates, if the fiber can be killed from the Lua world.
 	 */
 	FIBER_IS_SYSTEM         = 1 << 8,
+	/**
+	 * Someone has joined the fiber already. So this fiber can't be joined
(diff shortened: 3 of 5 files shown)


ochaplashkin (approved):

mkostoevr:
@ochaplashkin sorry for the approval drop, I had to update the documentation in the commit message.

changelogs/unreleased/gh-7562-fiber_join.md
```diff
@@ -0,0 +1,6 @@
+## bugfix/core
+
+* Fixed a possible inconsistent state entering if the `fiber_join` function is
+  used in a wrong way: now the `fiber_set_joinable` function only works if the
+  fiber is alive and is not joined yet and the `fiber_join` diagnoses a double
+  join as an error if it's possible to detect it (gh-7562).
```
xuniq, line 6:
```suggestion
* Fixed a possible inconsistent state entering if the `fiber_join` function was
  used incorrectly. Now the `fiber_set_joinable` function only works if the
  fiber is alive and is not joined yet. The `fiber_join` function now treats a double join as an error if it is possible to detect it (gh-7562).
```

xuniq (approved):

src/lib/core/fiber.c
```diff
@@ -732,6 +777,14 @@ fiber_join_timeout(struct fiber *fiber, double timeout)
 	if ((fiber->flags & FIBER_IS_JOINABLE) == 0)
 		panic("the fiber is not joinable");
 
+	if ((fiber->flags & FIBER_IS_JOINED) != 0) {
+		diag_set(JoinJoinedFiber);
+		return -1;
```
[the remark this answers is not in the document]
mkostoevr: I see, the `fiber_join` user does not expect any error except the one raised by the joined fiber.

At the same time panicing here may break the system for few users who has the actual use after free in their code, this may and may not be the lesser evil.

Gerold103 (commented):
I think this patch is not needed. We should panic if the user tries to do any use-after-free-like tricks. It is not safe to continue as it may lead to all sorts of even worse bugs than just crashes.

If we somewhy still want to allow that, I will leave the review to someone else. For instance, to @locker.

src/lib/core/fiber.c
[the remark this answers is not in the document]
mkostoevr: @Gerold103 thanks for the review, made it panic on a detected double join.

src/lib/core/fiber.c
```diff
@@ -664,6 +664,51 @@ fiber_set_joinable(struct fiber *fiber, bool yesno)
...
+	 *     5: is_joinable && is_dead && !is_joined
+	 *     6. is_joinable && is_dead && is_joined
+	 *
+	 *  Othervice, allow if: !is_dead && (!is_joinable || !is_joined)
+	 *  So ignore if: is_dead || (is_joinable && is_joined)
+	 *  So ignore if: is_dead || is_joined, because joined implies joinable.
+	 */
+	if ((fiber->flags & FIBER_IS_DEAD) != 0 ||
+	    (fiber->flags & FIBER_IS_JOINED) != 0)
+		return;
```
Gerold103, line 710: This is also use-after-free. If a fiber is already joined or dead, it means it is "deleted" and going to be reused soon. Shouldn't be accessed by any users in this state. I would make it also a panic. If user's code reaches this place, then something is very very wrong in that code. It is like trying to make `free(memory)` nop if you see that the memory is already freed. That won't make the code correct anyway, and such code needs to be fixed.

Gerold103 (commented):
Hm, that was unusual - I did expect arguing 😁. Anyway, there is still one use-after-free case left.

src/lib/core/fiber.c
mkostoevr: I see, the incorrect code should be corrected, and the most effective way to do it is to kill the app when it's executed. I'm just being conservative and trying to create as few problems as possible since Tarantool can be used in mission-critical systems (not sure about this though).

It looks to me it's a non-zero probability that someone calls the `set_joinable` by accident and it will never actually join the dead or joined fiber. If we let it live in this case, nothing bad is going to happen.

So I would rather print a big warning here. Or, maybe, panic, but introduce a compat option that makes it warn instead? Not sure about the second option, it can lead to at least one production failure.

Gerold103: I totally agree about non-critical errors. But this is straight one of the most dangerous mistakes one can make, which can lead at best to an ungraceful crash, or at worst the app will keep running and the code will mess with totally unrelated fibers, or even corrupt memory (use-after-free can lead to anything).

The warnings or any other logs nobody reads until something breaks. Speaking of mission-critical systems - they are in more danger if they access invalid memory or totally random fibers, than if they would notice and fix the problem right away.

I can approve only the panic here, it is too dangerous to make this behaviour "supported" in any way. The compat option I don't mind, but this sounds bizarre - the option would be called `"allow_use_after_free"` :D?

mkostoevr: Thought over it a bit, it seems making the `fiber_set_joinable` ignore the invalid state may lead to much worse consequences than I expected.

Updated the comment and made it panic in this condition.

mkostoevr: @Gerold103, one last question about the patch. The two possible conditions aren't UAFs, but they can lead to ones sometimes:
1. Attempt to make a joined but not dead fiber joinable.
   a. Making such a fiber non-joinable will lead to a double recycle so it's OK to panic.
   b. Making it joinable is effectively a no-nop, but this can signify an intention to join the fiber later, which will be a UAF. Also this is a logical mistake: there's no reason to make the same fiber joinable twice. So I made it panic, is this OK?
2. Attempt to make a joinable finished and not-yet-joined fiber joinable.
   a. Making such a fiber non-joinable can lead to a fiber leak, so it's OK to panic.
   b. Making it joinable is the same as for 1.b.

What do you think about 1.b and 2.b?

Gerold103: I agree with both 1 and 2. I would only recommend to rename `FIBER_IS_JOINED` to `FIBER_IS_JOINING`. Otherwise it looks like it was actually joined (which means it stopped and was deleted), but in fact it is not yet joined.

Or otherwise you could delete `FIBER_IS_JOINED/FIBER_IS_JOINING` and simply unset `FIBER_IS_JOINABLE` before you start waiting inside `fiber_join_timeout()`. Then the effect will be the same but we will have less flags.

mkostoevr: Thanks for the review, renamed the flag to `FIBER_JOIN_BEEN_INVOKED`, so we don't have to reset it on the join finish semantically ("is joining" impies the process is going to finish at some point, which will require to clear the flag on the join finish to maintain the semantic). Effectively clearing the flag does not affect the tarantool behavior.

The second option will not allow to distinguish between the fiber which is initially non-joinable and the one made non-joinable by a join (this is required for preventing a double join by using the `fiber_set_joinable` while the joined fiber is still alive).

Gerold103: > The second option will not allow to distinguish between the fiber which is initially non-joinable and the one made non-joinable by a join

Do you need to distinguish it anywhere?

Gerold103 (approved):
Thanks for the fix!

src/lib/core/fiber.c
mkostoevr: > Do you need to distinguish it anywhere?

In the `fiber_set_joinable` function. If the `fiber_join` simply unsets the "joinable" flag I see no way to prevent the user from seting it back using the `fiber_set_joinable` if the fiber is alive. So we can have the situation below:

1. `fiber_new(joinee);`
2. `fiber_set_joinable(joinee, true), fiber_join(joinee);` - someone makes the fiber joinable and joins it, the join makes the fiber is non-joinable, but...
3. `fiber_set_joinable(joinee, true), fiber_join(joinee);` - someone else makes the fiber joinable and joins it again. Neither the `fiber_set_joinable` not the `fiber_join` can detect it because we don't know if the fiber is made non-joinable by a join or it's non-joinable initially if no "is_joined" flag exist.

locker (approved):

PR #9316, opened 2023-10-27
config: fix hanging alert about missing name

Serpentian:
At the last itaration of review it was decided to throw an alerts, when no UUID was passed to config and name is not set in _cluster.

This leads to alerts, thrown during replicaset bootstrap. However, _cluster:on_replace trigger wasn't updated for that change, it asssumed that when insert of a new replica is done, no alerts was thrown.

Let's fix the behavior of on_replace trigger so that it deletes alerts as soon as replica joined with name.

Follow-up #8978


Totktonada (approved):
Here we drop the alert when the new instance is joined. The alert appears, because the instance was added into the config beforehand. OK, got it. LGTM.

ylobankov (approved):

PR #9317, opened 2023-10-27
lua: add helpers for encoding/decoding IPROTO packets

### LINKED ISSUE #8054: Add helpers for encoding and decoding IPROTO requests
In the scope of #7716, we added the ability to override specific IPROTO requests on a configured Tarantool sever and send IPROTO responses over an IPROTO connection established by the server. Unfortunately, there's still no convenient helpers available that would allow to implement the IPROTO protocol in Lua on top of raw sockets, without configuring box. It could be useful for application developers that want to emulate IPROTO for some reason as well as in tests. Currently, we have to generate IPROTO requests manually, which is tedious:

https://github.com/tarantool/tarantool/blob/2574ff1a82e5910c0334c11c1c429a45d9f399b4/test/box-luatest/ghs_16_user_enumeration_test.lua#L18-L43

What we need is:
- Encoding and decoding of a greeting message.
- Generating authentication data.
- Handling of request/response fix header.

This is all done by net.box under the hood. We could probably just make the code public.

locker:
The following new constants and functions were added to the `box.iproto` namespace:

 - `GREETING_SIZE`: Size of a Tarantool greeting message

 - `GREETING_PROTOCOL_LEN_MAX`: Max length of a protocol string that can be encoded in a Tarantool greeting message.

 - `GREETING_SALT_LEN_MAX`: Max length of a salt string that can be encoded in a Tarantool greeting message.

 - `box.iproto.encode_greeting({version = x, uuid = x, salt = x})`: Encodes a Tarantool greeting message. Takes a table. Returns a string. Raises on error. The protocol is set to "Binary" (IPROTO).

 - `box.iproto.decode_greeting(string)`: Decodes a Tarantool greeting message. Takes a string. Returns a table with the following fields: `protocol`, `version`, `uuid`, `salt`. Raises on error. The input string must be exactly `GREETING_SIZE` bytes long.

 - `box.iproto.encode_packet(header[, body])`: Encodes an IPROTO packet. Takes a packet header and optionally a body given as a table or a string. A table argument will be encoded in MsgPack. A string argument will be copied as is (it's supposed to contain valid MsgPack but it isn't enforced). Returns a string. Raises on error.

 - `box.iproto.decode_packet(string[, pos])`: Decodes an IPROTO packet. Takes a string containing one or more encoded IPROTO packets and optionally a position in the string to start decoding from. If the position is omitted, the function will start decoding from the beginning of the string. On success returns the decoded packet header, body, and the position in the string where decoding stopped. Both header and body are returned as `msgpack.object`. The body may be absent (set to nil). If the input is truncated, returns nil and the min number of bytes required to decode the packet. On failure, raises an error.

Closes #8054

### DIFF
--- a/changelogs/unreleased/gh-8054-lua-iproto-encoder.md
+++ b/changelogs/unreleased/gh-8054-lua-iproto-encoder.md
@@ -1,0 +1,3 @@
+## feature/lua
+
+* Introduced helpers for encoding and decoding IPROTO packets in Lua (gh-8054).
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -79,7 +79,6 @@
 #include "mpstream/mpstream.h"
 
 enum {
-	IPROTO_SALT_SIZE = 32,
 	IPROTO_PACKET_SIZE_MAX = 2UL * 1024 * 1024 * 1024,
 };
 
--- a/src/box/iproto_constants.h
+++ b/src/box/iproto_constants.h
@@ -43,8 +43,10 @@ extern "C" {
 enum {
 	/** Maximal iproto package body length (2GiB) */
 	IPROTO_BODY_LEN_MAX = 2147483648UL,
-	/* Maximal length of text handshake (greeting) */
+	/** Size of iproto greeting message. */
 	IPROTO_GREETING_SIZE = 128,
+	/** Size of salt sent in iproto greeting message. */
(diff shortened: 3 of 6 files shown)


ochaplashkin (approved):

xuniq (approved):

Totktonada (approved):
Thank you for the patch! I have no objections.

PR #9324, opened 2023-11-01
test: fix number to string conversion in tuple format map test

CuriousGeorgiy:
This patch fixes build warnings issued by GCC 12.


Gumix (approved):
Thanks for the patch!

ochaplashkin (approved):

locker (approved):

PR #9328, opened 2023-11-02
config: support {{ replicaset_name }} and {{ group_name }}

Totktonada:
The declarative configuration supports `{{ instance_name }}` template variable from the beginning.

Now, two more ones were added: `{{ replicaset_name }}` and `{{ group_name }}`.

All the template variables are set according to the instance/replicaset/group topology from the given configuration.

Example:

```yaml
console:
  socket: '{{ group_name }}/{{ replicaset_name }}/{{ instance_name }}.admin'

groups:
  g:
    replicasets:
      r:
        instances:
          i: {}
```

The admin console will be at `g/r/i.admin` for instance `i`.

Part of #8862

### DIFF
--- a/changelogs/unreleased/config-replicaset-group-name-templates.md
+++ b/changelogs/unreleased/config-replicaset-group-name-templates.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Added support for the `{{ replicaset_name }}` and `{{ group_name }}`
+  templates in addition to the existing `{{ instance_name }}` (gh-8862).
--- a/test/config-luatest/vars_test.lua
+++ b/test/config-luatest/vars_test.lua
@@ -1,0 +1,307 @@
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+local helpers = require('test.config-luatest.helpers')
+
+local g = helpers.group()
+
+-- Write the given function into a Lua file and execute it.
+--
+-- Ensure that the exit code is zero and there is no stdout/stderr
+-- output.
+--
+-- Use require('myluatest') instead of require('luatest') inside
+-- the test case to get better diagnostics at failure.
+local function run_as_script(f)
+    return function(g)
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/config-replicaset-group-name-templates.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Added support of `{{ replicaset_name }}` and `{{ group_name }}` templates in
+  addition to existing `{{ instance_name }}` (gh-8862).
```
xuniq, line 4:
```suggestion
* Added support for the `{{ replicaset_name }}` and `{{ group_name }}` templates in
  addition to the existing `{{ instance_name }}` (gh-8862).
```

xuniq (approved):

changelogs/unreleased/config-replicaset-group-name-templates.md
Totktonada: Fixed.

test/config-luatest/vars_test.lua
```diff
@@ -0,0 +1,307 @@
...
+                        },
+                    },
+                },
+            },
+        },
+    }
+    local iconfig = cluster_config:instantiate(cconfig, instance_name)
+    local configdata = configdata_lib.new(iconfig, cconfig, instance_name)
+
+    local function exp_uri(group_name, replicaset_name, instance_name)
```
ImeevMA, line 154: This function is a one-time use, so maybe it's easier to use a local variable: `local exp = table.concat(...)`?

ImeevMA (approved):
Hi! Thank you for the patch! I have one minor comment.

test/config-luatest/vars_test.lua
Totktonada: My idea is to make the test cases constructed in a similar way (so it is easier to spot actual differences).

PR #9331, opened 2023-11-03
cmake: allow to bundle static dependencies in main project

### LINKED ISSUE #9242: Add facility to build tarantool statically via a cmake option

In order to build tarantool statically with downloading and building required libraries from the cloud, I must provide the `./static-build/` directory as the source directory for `cmake`.

```console
cmake -S static-build -B static-build -DCMAKE_TARANTOOL_ARGS="-DCMAKE_BUILD_TYPE=RelWithDebInfo;-DENABLE_WERROR=ON"
```

However, we have the `-DBUILD_STATIC=ON` option, but it does a little different job. This option forces tarantool to build statically, but all libraries except `openssl` are embedded in the binary from the system. In order to embed `openssl` in the binary as well, we need to provide the `-DOPENSSL_USE_STATIC_LIBS=TRUE` option to `cmake`.

```console
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_WERROR=ON -DBUILD_STATIC=ON -DOPENSSL_USE_STATIC_LIBS=TRUE
```

So it would be nice to have an option similar to `-DBUILD_STATIC=ON`, but allowing to build tarantool statically with downloading and building required libraries from the cloud.

**Additional context**

This feature is also needed to have a facility to build source packages via `cpack`.

**Definition of Done**

When I do something like this:

```console
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_WERROR=ON -DBUILD_STATIC_FOO_BAR=ON .
make -j $(nproc)
```

I would like to get statically built tarantool with embedded libraries from the cloud.

locker:
The `BUILD_STATIC` cmake config option forces the build system link Tarantool binary statically with its dependencies. It expects that all static libraries on which Tarantool binary depends are available at build time.

We don't use this option directly to create static binaries because it would produce different results on different build systems. Instead, we use the separate cmake config located in the static-build directory, which fetches all Tarantool dependencies from a predefined location before building a static binary.

Having a separate cmake config is inconvenient. Let's enable bundling of static binary dependencies right in the main project cmake config, like we bundle, for example, libcurl. To achieve that, the new build option was introduced `BUILD_STATIC_WITH_BUNDLED_LIBS`. It implies `BUILD_STATIC` and also fetches and builds all required dependencies, like the static-build cmake config used to. The latter doesn't do it anymore; from now on, it just sets `BUILD_STATIC_WITH_BUNDLED_LIBS` when building Tarantool. We can't remove the static-build cmake config yet because there are quire a few CI workflows depending on it.

Note that, just like `BUILD_STATIC`, `BUILD_STATIC_WITH_BUNDLED_LIBS` doesn't imply `OPENSSL_USE_STATIC_LIBS` so the latter should be set explicitly if one wants to use the static openssl library. However, setting `OPENSSL_USE_STATIC_LIBS` with `BUILD_STATIC_WITH_BUNDLED_LIBS` will force the build system use bundled static openssl library.

Closes #9242

**EE part**: https://github.com/tarantool/tarantool-ee/pull/588

#### Next steps
- ~Move Lua tests from `/static-build/test` to `/test/app-tap`. They should work fine for all builds.~ Done in #9337, tarantool/tarantool-ee#589.
- Drop `/static-build/cmake/CheckDependencies.cmake`. We should probably make it a target in the main project config.
- Update CI workflows so as not to use `/static-build/CMakeLists.txt` and drop `/static-build`.


Totktonada:
> There's no need to use `/static-build/CMakeLists.txt` anymore, but it wasn't deleted because it's still used by CI. We will switch the CI to the main cmake config in the future and drop the `/static-build` directory completely.

I'm a bit worrying that it ends up with two build ways: one is used, another is tested. Can we finish the story? The issue originates from the @tarantool/devx team, so we can involve them.

locker:
> > There's no need to use `/static-build/CMakeLists.txt` anymore, but it wasn't deleted because it's still used by CI. We will switch the CI to the main cmake config in the future and drop the `/static-build` directory completely.
> 
> I'm a bit worrying that it ends up with two build ways: one is used, another is tested. Can we finish the story? The issue originates from the @tarantool/devx team, so we can involve them.

It was a draft PR. I reworked it so that the `BUILD_STATIC` option of `/CMakeLists.txt` and `/static-build/CMakeLists.txt` work exactly like before. That is, the former uses the static libraries installed on the system while the latter fetches and builds all dependencies. However, `/static-build/CMakeLists.txt` is now a thin wrapper around `/CMakeLists.txt` - it doesn't fetch dependencies by itself but instead sets the new option `BUILD_STATIC_WITH_BUNDLED_LIBS` that forces the main project to do that.

The plan is to drop `/static-build` completely but unfortunately it isn't that simple because it is a part of the packaging CI workflow. It uses cpack under the hood, see `/static-build/cmake/MakePackages.cmake`, and it should be reworked first. I mentioned it in the PR description. It's better to do in following PRs.

ylobankov (approved):

Issue #4467, opened 2019-08-30
Case insensitive access to Lua spaces from SQL


kostja:
If SQL identifier is unquoted, it is uppercased before store/use. If it is quoted, it is left intact. This is ANSI, but this creates a lot of confusion - first of all, because it is not easy to use, since people don't like their names being mangled by the database and forcefully converted to uppercase, second, because there are Lua spaces, which do not have uppercase representation.

Tarantool should preserve the original case supplied by the user. To ensure the name can be looked up with case-insensitive search, it should add a unique case-insensitive accent-sensitive index for each metadata object which can have a name (spaces, views, field names, functions, triggers). This will allow using Lua spaces and SQL tables in both frontends without mangling the original case: Lua frontend will use case-sensitive search, while SQL frontend will use the case-insensitive one.
By making the case-insensitive index unique we will make sure there is no ambiguity.

For example, if I do `create table foo (a int primary key)`, I should be able to use it from lua using `box.space.foo`.

If I do `box.scheam.space.create('foo')`, `select * from foo` should work fine. 

For quoted SQL identifiers we can continue using the case-sensitive index.
Note: some bizarre use case may actually require that the database upper-cases their object name before store (i.e. strict ansi compatibility). We can keep the uppercasing as an option for these, box.cfg.strict={bunch of flags} is a possible name of the switch.

Totktonada:
Why not just avoid doing any implicit case transformations as all modern programming languages do?

kostja:
This would break a lot of SQL code.

Totktonada:
Since we don't compatible with any SQL dialect of popular databases, manual actions (maybe, a lot of them) will anyway needed to migrate from other database to tarantool. I guess Tarantool/SQL will be mostly used for new projects rather then for adopting old ones. While your objection is true, I think existing code will be 'broken' anyway in many and many ways.

Gerold103:
> Not supporting case insensitive lookup would break a lot of SQL code.
So how does it require to store a second column? Just use case-insensitive comparison when lookup a column name in field array. If it is quoted, lookup like now. If it is not, then use strcasecmp.

pgulutzan:
As admitted in the issue description: the current behaviour is compatible with the standard. I hope that this suggested behaviour change will be rejected.

kostja:
@pgulutzan how exactly is case insensitive access to spaces created from Lua in SQL violates SQL standard? 
This is obviously outside the SQL standard domain.
[...]

PR #9333, opened 2023-11-07
sql: fix segmentation fault when JOIN uses USING

ImeevMA:
Before this patch, nameInUsingClause() expected old_col to be non-NULL, but this may not be the case. This patch fixes this feature accordingly.

Follow-up #4467

NO_DOC=bugfix
NO_CHANGELOG=feature was not released yet


igormunkin (approved):
@ImeevMA, thanks for the patch! LGTM.

ylobankov (approved):

Issue #8793, opened 2023-06-21
Reworking "default" functionality in SQL


ImeevMA:
Follow-up of https://github.com/tarantool/tarantool/issues/8157 and https://github.com/tarantool/tarantool/issues/8609. Once these issues are resolved, we need to redesign the SQL "default" functionality so that it use the newly created core "default" functionality.

PR #9334, opened 2023-11-07
sql: remove field sql_default_value

ImeevMA:
This patch removes the sql_default_value field from the struct field_def and the sql_default_value_expr field from the struct tuple_field as they are no longer needed.

Follow-up #8793

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring


ochaplashkin (approved):

Gerold103 (approved):
Thanks for the patch!

alyapunov (approved):

PR #9335, opened 2023-11-07
luajit: bump new version

Buristan:
* x64: Properly fix __call metamethod return dispatch.

NO_DOC=LuaJIT submodule bump
NO_TEST=LuaJIT submodule bump


igormunkin:
The changeset is merged to master in scope of #9388, backported to 2.11 and 2.10 in scope of #9389 and #9390 respectively.

PR #9338, opened 2023-11-07
config: add dependencies for roles

ImeevMA:
This patch adds dependencies support for roles.

Part of #9078

@TarantoolBot document
Title: dependencies for roles

Roles can now have dependencies. This means that the verify() and apply() methods will be executed for these roles, taking into account the dependencies. Dependencies should be written in the "dependencies" field of the array type. Note, the roles will be loaded (not applied!) in the same order in which they were specified, i.e. not taking dependencies into account.

Example:

Dependencies of role A: B, C
Dependencies of role B: D
No other role has dependencies.

Order in which roles were given: [E, C, A, B, D, G]
They will be loaded in the same order: [E, C, A, B, D, G]
The order, in which functions verify() and apply() will be executed: [E, C, D, B, A, G].

### DIFF
--- a/changelogs/unreleased/gh-9078-role-dependencies.md
+++ b/changelogs/unreleased/gh-9078-role-dependencies.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* Introduced dependencies for roles (gh-9078).
--- a/src/box/lua/config/applier/roles.lua
+++ b/src/box/lua/config/applier/roles.lua
@@ -1,19 +1,109 @@
 local log = require('internal.config.utils.log')
 
 local last_loaded = {}
-local last_loaded_names_ordered = {}
+local last_roles_ordered = {}
 
 local function stop_roles(roles_to_skip)
-    for id = #last_loaded_names_ordered, 1, -1 do
-        local role_name = last_loaded_names_ordered[id]
+    local roles_to_stop = {}
+    for id = #last_roles_ordered, 1, -1 do
+        local role_name = last_roles_ordered[id]
         if roles_to_skip == nil or roles_to_skip[role_name] == nil then
-            log.verbose('roles.apply: stop role ' .. role_name)
-            local ok, err = pcall(last_loaded[role_name].stop)
-            if not ok then
-                error(('Error stopping role %s: %s'):format(role_name, err), 0)
-            end
+            table.insert(roles_to_stop, role_name)
         end
     end
+    if #roles_to_stop == 0 then
+        return
+    end
+    local deps = {}
+    for role_name in pairs(roles_to_skip or {}) do
+        local role = last_loaded[role_name] or {}
+        -- There is no need to check transitive dependencies for roles_to_skip,
+        -- because they were already checked when roles were started, i.e. if
+        -- role A depends on role B, which depends on role C, and we stop
+        -- role C, then we will get the error that role B depends on role C.
+        for _, dep in pairs(role.dependencies or {}) do
+            deps[dep] = deps[dep] or {}
+            table.insert(deps[dep], role_name)
+        end
+    end
+    for _, role_name in ipairs(roles_to_stop) do
+        if deps[role_name] ~= nil then
+            local err
+            if #deps[role_name] == 1 then
+                err =('role %q depends on it'):format(deps[role_name][1])
+            else
+                local names = {}
+                for _, v in ipairs(deps[role_name]) do
+                    table.insert(names, ("%q"):format(v))
+                end
+                local names_str = table.concat(names, ', ')
+                err = ('roles %s depend on it'):format(names_str)
+            end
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-9078-role-dependencies.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* Dependencies for roles is introduced (gh-9078).
```
xuniq, line 3:
```suggestion
* Introduced dependencies for roles (gh-9078).
```

xuniq (approved):

changelogs/unreleased/gh-9078-role-dependencies.md
ImeevMA: Thanks, fixed!

src/box/lua/config/applier/roles.lua
```diff
@@ -57,10 +108,18 @@ local function apply(config)
...
-        table.insert(loaded_names_ordered, role_name)
+        if role.dependencies ~= nil and type(role.dependencies) ~= 'table' then
+            local err = 'Wrong type of role %s dependencies'
+            error(err:format(role_name), 0)
+        end
+        roles_dependencies[role_name] = {dependencies = role.dependencies or {}}
     end
 
+    -- Re-sorting of roles taking into account dependencies between them.
+    roles_ordered = resort_roles(roles_ordered, roles_dependencies)
```
Totktonada, line 119: Nit: Can't we just pass `loaded` instead of `roles_dependencies`? (`ipairs()` in `add_role` should be updated with `or {}` in this case.)

test/config-luatest/roles_test.lua
```diff
@@ -353,3 +353,164 @@ g.test_role_reload_error = function(g)
...
+            stop = function() end,
+        }
+    ]]
+
+    helpers.failure_case(g, {
+        roles = {one = one},
+        options = {
+            ['roles'] = {'one'}
+        },
+        exp_err = 'Wrong type of role one dependencies'
```
Totktonada, line 447: Nit: I propose to try to make the error more clear.

> Role "one" has field "dependencies" of type string, array-like table or nil expected

test/config-luatest/roles_test.lua
```diff
@@ -353,3 +353,164 @@ g.test_role_reload_error = function(g)
...
+        }
+    ]]
+
+    helpers.failure_case(g, {
+        roles = {one = one},
+        options = {
+            ['roles'] = {'one'}
+        },
+        exp_err = 'Role one requires role two, but the latter is not in the ' ..
+                  'list of roles of the instance'
```
Totktonada, line 467: Nit: I propose to use `%q` instead of `%s` for role names placeholders: this way the error message should be more clear. (The same about error messages regarding circular dependencies.)

changelogs/unreleased/gh-9078-role-dependencies.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
```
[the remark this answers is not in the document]
ImeevMA: Fixed.

src/box/lua/config/applier/roles.lua
```diff
@@ -1,11 +1,11 @@
 local log = require('internal.config.utils.log')
 
 local last_loaded = {}
-local last_loaded_names_ordered = {}
+local last_roles_ordered = {}
 
 local function stop_roles(roles_to_skip)
```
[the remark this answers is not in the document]
ImeevMA: True, fixed.

src/box/lua/config/applier/roles.lua
ImeevMA: Done.

test/config-luatest/roles_test.lua
ImeevMA: Done.

test/config-luatest/roles_test.lua
ImeevMA: Done.

src/box/lua/config/applier/roles.lua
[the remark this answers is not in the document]
Totktonada: Thanks! It looks correct. I have two nits.

* It is not obvious why we don't need to collect all the transient dependencies from `roles_to_skip` and why just the first level of dependencies is enough. I would appreciate a comment in the code.
* Is it hard to add information about roles depending on the given to-be-stopped role into the error message? I think it is useful to fix the misconfiguration.

Totktonada (approved):

ylobankov (approved):

src/box/lua/config/applier/roles.lua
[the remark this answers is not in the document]
ImeevMA: Added a comment about transitive dependencies and added names of roles which depend on the removed role in the error message.

PR #9341, opened 2023-11-07
Fix cleanup in http.client

### LINKED ISSUE #9283: libcurl update from 8.3.0 to 8.4.0 broke etcd client integration tests
Commit https://github.com/tarantool/tarantool/commit/0ea1ba877b8e62e0acb558b55623d93bf58a0e1f (merged in #9270) broke the following tests in the tarantool-ee repository:
- `etcd-client/etcd_client_general_integration_test.lua`
- `etcd-client/etcd_client_tls_integration_test.lua`

The full log is [attached](https://github.com/tarantool/tarantool/files/13054484/0_etcd-client.v3.4.25.txt).

ligurio:
Fixes #9283


src/curl.c
```diff
@@ -180,7 +188,7 @@ curl_multi_sock_cb(CURL *easy, curl_socket_t fd, int what, void *envp,
 	struct curl_env *env = (struct curl_env *) envp;
 	struct ev_io *watcher = (struct ev_io *) watcherp;
 
-	if (what == CURL_POLL_REMOVE) {
+	if (what == CURL_POLL_REMOVE || env->cleanup == 1) {
```
olegrok, line 191: Is it possible to run several chunked requests?

I guess if we have two such requests behaviour will be following:
- gc sets cleanup - true
- the first finished request removes mempool
- application crashes because the second request is in fly

An approach with fair refcounts should be better. At first it considers the case above. At second you don't need to handle cleanup == 1 case in different places since it makes code more complex and unclear

src/curl.c
```diff
@@ -75,8 +75,13 @@ curl_multi_process(CURLM *multi, curl_socket_t sockfd, int events)
...
 	int msgs_left;
 	while ((msg = curl_multi_info_read(multi, &msgs_left))) {
-		if (msg->msg != CURLMSG_DONE)
+		/* See https://curl.se/libcurl/c/curl_multi_info_read.html */
+		if (msg && (msg->msg != CURLMSG_DONE)) {
+			CURL *e = msg->easy_handle;
+			curl_multi_remove_handle(multi, msg->easy_handle);
+			curl_easy_cleanup(e);
 			continue;
+		}
```
oleg-jukovec, line 87: From the commit message:

> Documentation for this function [1] describes
how messages with status `CURLMSG_DONE` must be handled: one need call
`curl_multi_remove_handle(multi_handle, e)` and then
`curl_easy_cleanup(e)`.

But the  msg->msg != CURLMSG_DONE. This is not a problem now because so the code is not executed:

> When msg is CURLMSG_DONE, the message identifies a transfer that is done, and then result contains the return code for the easy handle that just completed.

> At this point, there are no other msg types defined.

src/httpc.c
```diff
@@ -145,9 +145,12 @@ httpc_env_destroy(struct httpc_env *ctx)
...
 
-	curl_env_destroy(&ctx->curl_env);
-
-	mempool_destroy(&ctx->req_pool);
+	if (mempool_count(&ctx->req_pool) == 0) {
+		mempool_destroy(&ctx->req_pool);
+		curl_env_destroy(&ctx->curl_env);
+	} else {
+		ctx->curl_env.cleanup = 1;
+	}
```
oleg-jukovec, line 153: I don't see a path of execution when the `curl_env_destroy(&ctx->curl_env);` is called with `ctx->curl_env.cleanup == 1`. There could be a memory leak here.

src/curl.c
ligurio: Right, commit has been removed.

src/httpc.c
ligurio: Patch has been significantly updated.

changelogs/unreleased/gh-9283-fix-crash-on-gc-collect.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/http
+
+* Fixed crash on garbage collection httpc objects (gh-9283).
```
andreyaksenov, line 3:
```suggestion
* Fixed a crash on garbage collection of httpc objects (gh-9283).
```

andreyaksenov (approved):

changelogs/unreleased/gh-9283-fix-crash-on-gc-collect.md
ligurio: Fixed, thanks!

ochaplashkin (approved):

oleg-jukovec (approved):
Thank you for the patch!

test/app-luatest/http_client_test.lua
```diff
@@ -1352,3 +1352,14 @@ g.test_http_client_io_post_parts = function(cg)
     t.assert_equals(io.status, 200, 'io')
     t.assert_equals(io.reason, 'Ok', '200 - Ok')
 end
+
+g.test_gh_9346_httpc_io_cleanup = function(cg)
+    local c = require('http.client').new()
+    local r = c:get(cg.url, {chunked = true}) -- luacheck: no unused
```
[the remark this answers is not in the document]
olegrok: E.g.

```lua
url = 'https://api.open-meteo.com/v1/forecast?latitude=52.52&longitude=13.41&current=temperature_2m,wind_speed_10m&hourly=temperature_2m,relative_humidity_2m,wind_speed_10m'

c = require('http.client').new()


r1 = c:get(url, {chunked = true})
r2 = c:get(url, {chunked = true})

r1:read(20)
r2:read(20)

r1 = nil
collectgarbage()
collectgarbage()

r2 = nil

collectgarbage()
collectgarbage()

r3 = c:get(url, {chunked = true})
r4 = c:get(url, {chunked = true})

c = nil
collectgarbage()
collectgarbage()

r3:read(20) -- fails here
r3 = nil
collectgarbage()
collectgarbage()

data = r4:read(20)
print(data)

r4 = nil
collectgarbage()
collectgarbage()

os.exit()
```

Before this patch script fails with:
```
IllegalParams: io: request must be io
```

Currently it works fine.

olegrok (commented):
Thanks for your changes. See my comments below.

src/curl.c
```diff
@@ -180,28 +180,28 @@ curl_multi_sock_cb(CURL *easy, curl_socket_t fd, int what, void *envp,
 {
```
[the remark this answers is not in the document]
ligurio: Added.

cmake/BuildLibCURL.cmake
```diff
@@ -185,11 +185,14 @@ macro(curl_build)
     # catch the symbols may require extra work.
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_ZSTD=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DLIBCURL_OUTPUT_NAME=libcurl")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_CURLDEBUG=OFF")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_DEBUG=OFF")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_CURLDEBUG=${TARANTOOL_DEBUG}")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_DEBUG=${TARANTOOL_DEBUG}")
```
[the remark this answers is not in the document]
ligurio:
```diff
--- a/cmake/BuildLibCURL.cmake
+++ b/cmake/BuildLibCURL.cmake
@@ -185,8 +185,6 @@ macro(curl_build)
     # catch the symbols may require extra work.
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_ZSTD=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DLIBCURL_OUTPUT_NAME=libcurl")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_CURLDEBUG=${TARANTOOL_DEBUG}")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_DEBUG=${TARANTOOL_DEBUG}")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DUSE_MSH3=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DENABLE_WEBSOCKETS=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCMAKE_UNITY_BUILD=OFF")
```

test/app-luatest/http_client_test.lua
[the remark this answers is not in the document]
ligurio: Fixed and separate test was added. I've reduced your initial reproducer, this could be reproduced with a single request as well.

src/curl.c
```diff
@@ -59,6 +59,8 @@ curl_multi_process(CURLM *multi, curl_socket_t sockfd, int events)
 	 * more data to send off before it is "satisfied".
 	 */
 	do {
+		if (!multi)
+			return;
```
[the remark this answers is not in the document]
olegrok: > Moved to the top of the function.

Somewhy I don't see this change. We don't need to check this value for each iteration. It's enough to check it only once

src/curl.c
```diff
@@ -180,28 +180,28 @@ curl_multi_sock_cb(CURL *easy, curl_socket_t fd, int what, void *envp,
...
 		}
-		ev_io_init(watcher, curl_sock_cb, fd, 0);
-		watcher->data = env;
+		ev_io_init(env->watcher, curl_sock_cb, fd, 0);
+		env->watcher->data = env;
 		++env->stat.sockets_added;
-		curl_multi_assign(env->multi, fd, watcher);
+		curl_multi_assign(env->multi, fd, env->watcher);
 		say_debug("curl %p: add fd=%d", env, fd);
 	}
```
olegrok, line 219: nit: isn't it better to keep this code above as it was and only here assign `env->watcher = watcher`?

src/curl.c
```diff
@@ -275,7 +276,7 @@ void
 curl_env_finish(struct curl_env *env)
 {
 	assert(env);
-	if (env->multi != NULL) {
```
olegrok: nit: I don't know that our codestyle says about it but IMP explicit comparison was better.

test/app-luatest/http_client_test.lua
```diff
@@ -1352,3 +1353,36 @@ g.test_http_client_io_post_parts = function(cg)
...
+                       collectgarbage()
+                       collectgarbage()
+
+                       r:read(1)]]
+    local ph = popen.new({ tarantool_exe, '-e', lua_code:format(cg.url) },
+                         { stderr = popen.opts.PIPE })
+    t.assert(ph)
+
+    -- Wait for the process startup.
+    grep_stderr_or_fail(ph, "IllegalParams: io: request must be io")
```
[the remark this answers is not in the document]
ligurio: > It's quite strange to add a test for "wrong" behaviour.

Okay, I can squash commits. For me, it is not strange because in this way I show old behavior and with fix change an assent. With squashed commits, all things will not be so obvious. 

> It's a bit ugly... Is it possible to implement it without popen in commit that fixes an issue?

No, with a repro inside a luatest testcase bug #9346 is triggered.

src/curl.c
ligurio: Already fixed, thanks.

src/curl.c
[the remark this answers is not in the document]
ligurio: > It doesn't set env->multi to NULL. It just deallocates a memory. After that pointer still points to some memory (that you doesn't own anymore).

1. `curl_multi_cleanup`, https://github.com/curl/curl/blob/ca6bafce95365fadd906a533bb16e47b638c5c2a/lib/multi.c#L2748C11-L2748C29
2. `curl_multi_remove_handle`, https://github.com/curl/curl/blob/ca6bafce95365fadd906a533bb16e47b638c5c2a/lib/multi.c#L930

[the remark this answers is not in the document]
olegrok: My question is not about "it's broken". My question is do we really need this condition and that will happen in case if it will be removed?

cmake/BuildNghttp2.cmake
```diff
@@ -12,7 +12,7 @@ macro(nghttp2_build)
     endif()
 
     set(NGHTTP2_CMAKE_FLAGS "-DENABLE_WERROR:BOOL=ON")
-    list(APPEND NGHTTP2_CMAKE_FLAGS "-DENABLE_DEBUG:BOOL=OFF")
+    list(APPEND NGHTTP2_CMAKE_FLAGS "-DENABLE_DEBUG:BOOL=${TARANTOOL_DEBUG}")
```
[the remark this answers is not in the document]
olegrok: > we could use CMAKE_VERBOSE_MAKEFILE [1](https://github.com/tarantool/tarantool/pull/9341#user-content-fn-1-e3fa3e62e5769095c76be7d31979a7ff) for this purpose.

I guess it's about make build logs. You can use `VERBOSE=1 make` to achieve that.

src/curl.c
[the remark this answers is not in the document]
ligurio: > My question is do we really need this condition and that will happen in case if it will be removed?

Initially, it was added as a protection for a case when env is destroyed and we have an active ev callback with http request. I've removed condition and checked on etcd-client tests, seems this condition is not required anymore.

src/curl.c
```diff
@@ -185,10 +199,7 @@ curl_multi_sock_cb(CURL *easy, curl_socket_t fd, int what, void *envp,
 
 	if (what == CURL_POLL_REMOVE) {
 		say_debug("curl %p: remove fd=%d", env, fd);
-		assert(env->watcher != NULL);
```
olegrok: Old logic expected that watcher always exists. But after your change absence of watcher will be ignored. Is it expected?

test/app-luatest/http_client_test.lua
```diff
@@ -1352,3 +1353,47 @@ g.test_http_client_io_post_parts = function(cg)
...
+    local tarantool_exe = arg[-1]
+    local lua_code = [[local url = '%s'
+                       local c = require('http.client').new()
+                       local r = c:get(url, {chunked = true})
+
+                       c = nil
+                       collectgarbage()
+                       collectgarbage()
+
+                       r:read(1)]]
```
olegrok, line 1379: It's better to check result of this read because error message can be renamed. You test will check nothing.

ligurio: Updated:

```diff
diff --git a/test/app-luatest/http_client_test.lua b/test/app-luatest/http_client_test.lua
index 66f8d52c12..e684296d25 100644
--- a/test/app-luatest/http_client_test.lua
+++ b/test/app-luatest/http_client_test.lua
@@ -1354,19 +1354,11 @@ g.test_http_client_io_post_parts = function(cg)
     t.assert_equals(io.reason, 'Ok', '200 - Ok')
 end
 
--- Read from ph:stderr and fail if it doesn't contain a `pattern'.
-local function grep_stderr_or_fail(ph, pattern)
-    local output = ''
-    t.helpers.retrying({}, function()
-        local chunk = ph:read({timeout = 0.05, stderr = true})
-        if chunk ~= nil then
-           output = output .. chunk
-        end
-        t.assert_not_str_contains(output, pattern)
-    end)
-end
-
 g.test_http_client_gc_finalizer_gh_9453 = function(cg)
+    -- Nothing specific for a socket family here,
+    -- but sock_family == AF_UNIX will require a bit more additional code,
+    -- so skipped.
+    t.skip_if(cg.params.sock_family == 'AF_UNIX')
     local tarantool_exe = arg[-1]
     local lua_code = [[local url = '%s'
                        local c = require('http.client').new()
@@ -1376,13 +1368,17 @@ g.test_http_client_gc_finalizer_gh_9453 = function(cg)
                        collectgarbage()
                        collectgarbage()
 
-                       r:read(1)]]
+                       local ok, res = pcall(r.read, r, 1)
+                       assert(ok == true)
+                       assert(res ~= nil)
+                       ]]
     local ph = popen.new({ tarantool_exe, '-e', lua_code:format(cg.url) },
                          { stderr = popen.opts.PIPE })
     t.assert(ph)
 
-    -- Wait for the process startup.
-    grep_stderr_or_fail(ph, "IllegalParams: io: request must be io")
+    local stderr = ph:read({timeout = 0.05, stderr = true})
+    t.assert_equals(stderr, nil)
+    t.assert_covers(ph:wait(), {exit_code = 0})
 
     ph:close()
 end
```

src/curl.c
ligurio: Seems it was lost. Now, function's body is exactly the same as a condition's body.

cmake/BuildNghttp2.cmake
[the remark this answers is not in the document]
olegrok: I read this. I just expressed my opinion.

olegrok (approved):
Thanks for the fix!

andreyaksenov (approved):

Totktonada:
I was added into reviewers by a mistake (a temporary patch touches a file that has me as a code owner). So, I'm leaving from the reviewers.

If some my participation is needed, please, let me know.

src/lua/httpc.c
```diff
@@ -607,7 +615,11 @@ luaT_httpc_io_finish(lua_State *L)
 static int
 luaT_httpc_io_cleanup(lua_State *L)
 {
-	httpc_io_destroy(luaT_httpc_checkio(L));
+	struct httpc_io *io = luaT_httpc_checkio(L);
+	struct httpc_env *env = io->req->env;
+	httpc_io_destroy(io);
+	if (env->req_count == 0)
+		luaL_unref(L, LUA_REGISTRYINDEX, env->lua_ref);
```
olegrok, line 622: To be honest it's not clear why did you add one lua_ref but two lua_unref. There is a risk that double free (in Lua world) will happen.

ligurio: > To be honest it's not clear why did you add one lua_ref but two lua_unref. 

First case: there is a http client object and zero http chunked requests. We need to call `luaL_unref` in http client obj GC finalizer (`luaT_httpc_cleanup`), otherwise `env->lua_ref` will not be unreferenced.

Second case: there is a http client with a number of chunked http requests. We need to call  `luaL_unref` in the latest http chunked request GC finalizer (`luaT_httpc_io_cleanup`).

> There is a risk that double free (in Lua world) will happen.

Right, it is not good.

I propose another way to link httpc and http req Lua objects:

```diff
diff --git a/src/httpc.h b/src/httpc.h
index c78dbf6e90..bd9f498d9a 100644
--- a/src/httpc.h
+++ b/src/httpc.h
@@ -70,8 +70,6 @@ struct httpc_env {
 	struct mempool req_pool;
 	/** Statistics */
 	struct httpc_stat stat;
-	/** Lua reference. */
-	int lua_ref;
 };
 
 /**
diff --git a/src/lua/httpc.c b/src/lua/httpc.c
index 1679e8c7d7..743442a314 100644
--- a/src/lua/httpc.c
+++ b/src/lua/httpc.c
@@ -495,9 +495,6 @@ luaT_httpc_new(lua_State *L)
 	if (httpc_env_create(ctx, max_conns, max_total_conns) != 0)
 		return luaT_error(L);
 
-	lua_pushvalue(L, -1); /* Popped by luaL_ref(). */
-	ctx->lua_ref = luaL_ref(L, LUA_REGISTRYINDEX);
-
 	luaL_getmetatable(L, DRIVER_LUA_UDATA_NAME);
 	lua_setmetatable(L, -2);
 
@@ -511,7 +508,6 @@ luaT_httpc_cleanup(lua_State *L)
 	httpc_env_finish(env);
 	if (env->req_count == 0) {
 		httpc_env_destroy(env);
-		luaL_unref(L, LUA_REGISTRYINDEX, env->lua_ref);
 	}
 
 	/** remove all methods operating on ctx */
@@ -616,10 +612,7 @@ static int
 luaT_httpc_io_cleanup(lua_State *L)
 {
 	struct httpc_io *io = luaT_httpc_checkio(L);
-	struct httpc_env *env = io->req->env;
 	httpc_io_destroy(io);
-	if (env->req_count == 0)
-		luaL_unref(L, LUA_REGISTRYINDEX, env->lua_ref);
 
 	/** remove all methods operating on io */
 	lua_newtable(L);
diff --git a/src/lua/httpc.lua b/src/lua/httpc.lua
index 3fc5fe4ee6..32f0917d6e 100644
--- a/src/lua/httpc.lua
+++ b/src/lua/httpc.lua
@@ -616,6 +616,7 @@ curl_mt = {
                     },
                     _internal = {
                         io = resp._internal.io,
+                        httpc = self,
                     },
                 })
             end
```

src/httpc.c
```diff
@@ -208,12 +217,18 @@ httpc_request_new(struct httpc_env *env, const char *method,
...
+	++env->req_count;
+
 	return req;
 }
 
 void
 httpc_request_delete(struct httpc_request *req)
 {
+	struct httpc_env *env = req->env;
+	++env->req_count;
```
olegrok, line 229: Why `++`, not `--`?

ligurio: Fixed, thanks!

test/app-luatest/http_client_test.lua
```diff
@@ -60,20 +61,22 @@ end
 
 g.before_each(function(cg)
     local sock_family = cg.params.sock_family
+    package.loaded['http.client'] = nil
```
olegrok, line 64: It's quite strange solution. Just try to use `require('http.client').new()` to create isolated client.

src/curl.c
```diff
@@ -262,8 +262,20 @@ void
 curl_env_destroy(struct curl_env *env)
 {
 	assert(env);
-	if (env->multi != NULL)
+	if (env->multi) {
```
olegrok, line 265: Is it possible to avoid changing this line. Explicit comparison was better than implicit.

test/app-luatest/http_client_test.lua
```diff
@@ -1375,3 +1375,49 @@ g.test_http_client_io_post_parts = function(cg)
     t.assert_equals(io.status, 200, 'io')
     t.assert_equals(io.reason, 'Ok', '200 - Ok')
 end
+
+-- GH-9453: GC has collected HTTP client object
```
olegrok, line 1379: Probably we need to use past tense here

src/curl.c
ligurio: Reverted:

```diff
--- a/src/curl.c
+++ b/src/curl.c
@@ -262,7 +262,7 @@ void
 curl_env_finish(struct curl_env *env)
 {
        assert(env);
-       if (env->multi) {
+       if (env->multi != NULL) {
                CURL **list = curl_multi_get_handles(env->multi);
                if (list) {
                        for (int i = 0; list[i]; i++) {
```

test/app-luatest/http_client_test.lua
ligurio: Yes, it is a bit strange and unusual. However, there are at least two reasons to use it:
- we can achieve better isolation when module fully loaded 
- `http.client` has two interfaces:
  - with default client and access to table fields:
```lua
local httpc = require('http.client')
httpc.get('https://luajit.org')
```
  - with custom client
```lua
local httpc = require('http.client').new()
httpc.get('https://luajit.org')
```
and the second client's method cannot be accessed via dot:
```lua
tarantool> httpc = require('http.client').new()
---
...

tarantool> httpc.get('https://ya.ru/').status
---
- error: 'builtin/http.client.lua:628: Use client:get(...) instead of client.get(...):'
...

tarantool> 
```
so with `require('http.client').new()` most part of existed testcases will not work

So I left this as is.

olegrok: > we can achieve better isolation when module fully loaded

I don't understand that you mean when say "better". Standalone http client has enough isolation

> http.client has two interfaces: 

It's actually the same things. `httpc.get` is just a wrapper around some `static_client` that is also created via `httpc.new()`. So there is no any sense to even test it this static client deeply because it's just a test of this wrapper.

test/app-luatest/http_client_test.lua
ligurio: Changed to:

```
-- GH-9453: GC had collected HTTP client object
-- when HTTP IO object is alive.
```

olegrok (approved):

PR #9347, opened 2023-11-08
sql: update parser to exclude out of bounds access

sergos:
SVACE found a problem in parser that can cause an out of bounds access. The test is made via assert() and works for debug build. Fixed with early exit for the same condition as assert.

NO_DOC=no user-visible change
NO_CHANGELOG=code quality
NO_TEST=no change to actual behavior

### DIFF
--- a/extra/lempar.c
+++ b/extra/lempar.c
@@ -451,8 +451,8 @@ static unsigned int yy_find_shift_action(
   int i;
   int stateno = pParser->yytos->stateno;
  
-  if( stateno>=YY_MIN_REDUCE ) return stateno;
-  assert( stateno <= YY_SHIFT_COUNT );
+  assert(stateno>=YY_MIN_REDUCE || stateno<=YY_SHIFT_COUNT);
+  if( stateno>YY_SHIFT_COUNT ) return stateno;
   do{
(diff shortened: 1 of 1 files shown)


extra/lempar.c
```diff
@@ -451,8 +451,7 @@ static unsigned int yy_find_shift_action(
   int i;
   int stateno = pParser->yytos->stateno;
  
-  if( stateno>=YY_MIN_REDUCE ) return stateno;
-  assert( stateno <= YY_SHIFT_COUNT );
+  if( stateno>YY_SHIFT_COUNT ) return stateno;
```
ImeevMA, line 454: How about adding `assert(stateno >= YY_MIN_REDUCE || stateno <= YY_SHIFT_COUNT);` to the previous line?

ImeevMA (approved):
Hi! Thank you for the patch! I have just one small comment. The comment is about adding another assertion to replace the one removed here.

extra/lempar.c
sergos: done.

sergos:
The reason of OOB access is not possible because the table of `yy_action` contains only valid values according the YACC-like grammar provided to the lemon parser generator.

PR #9348, opened 2023-11-09
box: decrement `BOX_SPACE_MAX`

### LINKED ISSUE #9118: BOX_ID_NIL is a valid space id
**Bug description**

There're 2 constants in tarantool `BOX_ID_NIL` & `BOX_SPACE_MAX` which have the same value of `0x7fff_ffff`. This would make sense if BOX_SPACE_MAX was interpreted as the first invalid id, but it's not actually.

Here you can see that `>` is used
https://github.com/tarantool/tarantool/blob/70e423e92fc00df2ffe385f31dae9ea8e1cc1732/src/box/alter.cc#L489-L492
which implies BOX_SPACE_MAX is the last VALID space id.

However BOX_ID_NIL will be returned from for example box_space_id_by_name as a signifier of an invalid space id.
https://github.com/tarantool/tarantool/blob/70e423e92fc00df2ffe385f31dae9ea8e1cc1732/src/box/box.cc#L3511-L3515

**Steps to reproduce**

Here's an example which shows the problem. We create a space with id `0x7fff_ffff` and use box_space_id_by_name which returns the id of our existing space when passed a name of a nonexistent space:
```c
#include <stdint.h>
#include <assert.h>

uint32_t box_space_id_by_name(const char *, uint32_t);

int repro() {
	char s1_name[] = "no such space";
	uint32_t s1_id = box_space_id_by_name(s1_name, sizeof(s1_name) - 1);

	char s2_name[] = "space with legal id";
	uint32_t s2_id = box_space_id_by_name(s2_name, sizeof(s2_name) - 1);

	assert(s1_id != s2_id);

	return 0;
}
```
```lua
box.cfg()
box.schema.space.create('space with legal id', { id = 0x7fffffff })
box.schema.func.create('repro', { language = 'C' })
box.func.repro:call() -- this crashes even though it shouldn't
```

**Expected behaviour**

I guess 0x7fffffff shouldn't be a valid space id

locker:
Currently, `BOX_SPACE_MAX` equals `BOX_ID_NIL`, which is used as an error indicator in the box C API. As a result, if `box_space_id_by_name()` returns `BOX_ID_NIL`, it's impossible to figure out whether there's no space with the give name or the space exists and has the id equal to `BOX_SPACE_MAX`.

Let's decrement `BOX_SPACE_MAX` to fix this issue. Since this may break recovery, let's also introduce the new compatibility module option `box_space_max` to allow the user revert to the old behavior.

Closes #9118

### DIFF
--- a/src/box/schema_def.h
+++ b/src/box/schema_def.h
@@ -39,7 +39,6 @@ extern "C" {
 
 enum {
 	BOX_ENGINE_MAX = 3, /* + 1 to the actual number of engines */
-	BOX_SPACE_MAX = INT32_MAX,
 	BOX_FUNCTION_MAX = 32000,
 	BOX_INDEX_MAX = 128,
 	BOX_NAME_MAX = 65000,
@@ -130,6 +129,9 @@ enum {
 	BOX_ID_NIL = 2147483647
 };
 /** \endcond public */
+
+/** Max possible space id. */
+extern uint64_t BOX_SPACE_MAX;
 
 /** _space fields. */
 enum {
--- a/test/unit/lua_tweaks.c
+++ b/test/unit/lua_tweaks.c
@@ -2,12 +2,14 @@
 #include "lualib.h"
 #include <lauxlib.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <string.h>
 
 #include "core/tweaks.h"
 #include "diag.h"
 #include "fiber.h"
 #include "lua/error.h"
+#include "lua/msgpack.h"
 #include "lua/tweaks.h"
 #include "lua/utils.h"
 #include "memory.h"
@@ -22,8 +24,11 @@ static struct lua_State *L;
 static bool bool_var = true;
 TWEAK_BOOL(bool_var);
 
-static int int_var = 42;
+static int64_t int_var = 42;
 TWEAK_INT(int_var);
+
+static uint64_t uint_var = 123;
+TWEAK_UINT(uint_var);
 
 static double double_var = 3.14;
 TWEAK_DOUBLE(double_var);
@@ -45,7 +50,7 @@ TWEAK_ENUM(my_enum, enum_var);
 static void
 test_index(void)
 {
-	plan(10);
+	plan(12);
 	header();
 	lua_settop(L, 0);
 	is(luaT_dostring(L, "return tweaks.no_such_var"), 0, "no_such_var");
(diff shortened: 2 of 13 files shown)


andreyaksenov (approved):

ochaplashkin (approved):

test/unit/lua_tweaks.c
```diff
@@ -82,12 +90,6 @@ test_newindex(void)
 	   strcmp(diag_last_error(diag_get())->errmsg,
 		  "Value must be boolean, number, or string") == 0,
 	   "check error");
-	is(luaT_dostring(L, "tweaks.bool_var = 999999999999999999LLU"), -1,
-	   "invalid value - big int");
-	ok(!diag_is_empty(diag_get()) &&
-	   strcmp(diag_last_error(diag_get())->errmsg,
-		  "Value must be boolean, number, or string") == 0,
-	   "check error");
```
Totktonada, line 90: What is wrong with this test case?

src/box/schema_def.h
```diff
@@ -131,6 +130,9 @@ enum {
 };
 /** \endcond public */
 
+/** Max possible space id. */
+extern uint64_t BOX_SPACE_MAX;
```
Totktonada, line 134: 2^31-1 and 2^31-2 are in the `int32_t` (and `uint32_t`) range. Why the tweak type is `uint64_t`?

Is the idea to allow to relax the limit in a future if we'll rewrite all the related APIs? I guess, if we do, the tweak type is the simplest part.

OTOH, I see no problems in such implementation and I think that 64 bit tweaks may be useful in a future. So, if you want, let's leave it as is.

Totktonada (approved):
Thank you for the patch!

test/unit/lua_tweaks.c
locker: This test case covers error conditions checked by `luaT_tweaks_newindex`. Before this change `int64_t` was unsupported so `luaT_tweaks_newindex` raised an error. Now, it doesn't raise an error in this case. The tweak handler (`tweak_set_bool`) will still raise an error (a different one though) but this test case isn't about it. The latter is tested in the `test_bool_var` test case. So I removed this particular check because it isn't relevant to `luaT_tweaks_newindex` anymore.

src/box/schema_def.h
locker: True, we could use `uint32_t` for `BOX_SPACE_MAX` but `TWEAK_UINT` requires `uint64_t`. I could introduce a new helper `TWEAK_UINT32` that would work for `uint32_t` but I don't want to because wherever we need `uint32_t` `uint64_t` will do as well.

PR #9359, opened 2023-11-10
replication: fix extraneous split-brain alerting

### LINKED ISSUE #9138: CONFIRM without quorum leads to split-brain error
## My case:

There is a cartridge cluster with a storage replicaset with three nodes.

All spaces are in sync mode.

Stateful failover with etcd is enabled.

Election mode is manual. Function `box.ctl.promote()` is called inside `apply_config` if the node is the master.

During high load, the cpu ran out and a master in the replicaset stuck. The failover decided to switch the master. The new master also got stuck. And there were several such switches. At some point, the error "Split-Brain discovered: got a request from a foreign synchro queue owner" appeared in the replicaset.

## Possible cause:

A node while it was a master wrote a transaction in WAL and began obtaining quorum. At that moment master switching occurred, but the node managed to obtain quorum and write CONFIRM in its WAL. After writing, the node sent the WAL entry to other nodes in replicaset in which the limbo owner had already been changed.

<details> 
  <summary>screenshots</summary>
  Errors in Cartridge UI  

  
  
  storage-9-1  

  Pasted image 20230913132200
  
  storage-9-2  

  Pasted image 20230913132209
  
  storage-9-3  

  Pasted image 20230913132221
</details>

Last WAL entries
```
---
HEADER:
  lsn: 85936
  replica_id: 3
  type: CONFIRM
  timestamp: 1694197746.2396
BODY:
- null
- 3
- 85930
---
HEADER:
  lsn: 85937
  replica_id: 3
  type: CONFIRM
  timestamp: 1694197747.5981
BODY:
- null
- 3
- 85935
...
```

* OS: RockyLinux
* OS Version: 8.5
* Architecture: amd64

Tarantool 2.11.0

sergepetrenko:
Current split-brain detector implementation raises an error each time a CONFIRM or ROLLBACK entry is received from the previous synchronous transaction queue owner. It is assumed that the new queue owner must have witnessed all the previous CONFIRMS. Besides, according to Raft, ROLLBACK should never happen.

Actually there is a case when a CONFIRM from an old term is legal: it's possible that during leader transition old leader writes a CONFIRM for the same transaction that is confirmed by the new leader's PROMOTE. If PROMOTE and CONFIRM lsns match there is nothing bad about such situation.

Symmetrically, when an old leader issues a ROLLBACK with the lsn right after the new leader's PROMOTE lsn, it is not a split-brain.

Allow such cases by tracking the last confirmed lsn for each synchronous transaction queue owner and silently nopifying CONFIRMs with an lsn less than the one recorded and ROLLBACKs with lsn greater than that.

Closes #9138

NO_DOC=bugfix


andreyaksenov (approved):

ochaplashkin (approved):

sergepetrenko:
@Gerold103 @Serpentian I just understood I missed persisting the `confirmed_vclock`. If a node restarts, after making a snapshot, it won't have any info about `confirmed_lsn` for all the replica ids but the current limbo owner.

Please don't review this yet. I'll come back with an improved patch.

sergepetrenko:
@Gerold103, @Serpentian, all ready now. PTAL.

Serpentian (commented):
Great patch! Feel free to resolve the nit msg without changes.

src/box/xrow.h
```diff
@@ -355,9 +354,14 @@ struct synchro_request {
...
 	 * The new term the instance issuing this request is in. Only used for
-	 * PROMOTE request.
+	 * PROMOTE and DEMOTE requests.
 	 */
 	uint64_t term;
+	/**
+	 * Confirmed lsns of all the previous limbo owners. Only used for
+	 * PROMOTE and DEMOTE requests.
+	 */
+	struct vclock *confirmed_vclock;
```
Gerold103, line 364: Having the single `synchro_request` for all request types was fine, but since now it starts looking as ugly as the old huge subscribe request was looking with all its fields being optional and it being used for all requests at once. I would suggest to split the request into separate `synchro_ownership_request` (for `PROMOTE` and `DEMOTE`) and `synchro_txn_request` (for `CONFIRM`/`ROLLBACK`). The latter wouldn't have `confirmed_vclock` field. Some or all the other fields could be duplicated, yes, but it would look simpler.

src/box/txn_limbo.c
```diff
@@ -586,19 +599,15 @@ txn_limbo_write_promote(struct txn_limbo *limbo, int64_t lsn, uint64_t term)
...
+	limbo->confirmed_lsn = vclock_get(&limbo->confirmed_vclock,
+					  replica_id);
 	box_update_ro_summary();
-	/*
-	 * Only nullify confirmed_lsn when the new value is unknown. I.e. when
-	 * prev_id != replica_id.
-	 */
-	if (replica_id != prev_id)
-		limbo->confirmed_lsn = 0;
 }
```
Gerold103, line 611: It seems you could extract confirmed vclock persistency into a separate commit. And the filtering update would be in its own small commit.

Gerold103 (commented):
Thanks for the patch!

src/box/xrow.h
```diff
@@ -355,9 +354,14 @@ struct synchro_request {
 	int64_t lsn;
 	/**
 	 * The new term the instance issuing this request is in. Only used for
-	 * PROMOTE request.
+	 * PROMOTE and DEMOTE requests.
 	 */
 	uint64_t term;
+	/**
```
Serpentian, line 360: DocBot request: [doc](https://www.tarantool.io/en/doc/latest/dev_guide/internals/iproto/replication/#iproto-raft-promote)

src/box/xrow.c
```diff
@@ -1310,6 +1324,13 @@ xrow_decode_synchro(const struct xrow_header *row, struct synchro_request *req)
 		case IPROTO_TERM:
 			req->term = mp_decode_uint(&d);
 			break;
+		case IPROTO_VCLOCK:
```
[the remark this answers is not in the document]
sergepetrenko: AFAICS we memset the request to all 0s at the beginning of both `xrow_decode_raft` and `xrow_decode_synchro`, so I don't see much difference here.
https://github.com/tarantool/tarantool/blob/18eaeb737dd595cf5d008e8c70e7b353b4331a74/src/box/xrow.c#L1287

src/box/txn_limbo.c
sergepetrenko: Good idea, done.

src/box/xrow.c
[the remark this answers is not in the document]
Serpentian: You're right, POSIX [guarantees](https://www.austingroupbugs.net/view.php?id=940#c2696), that this'll be interpreted as NULL, my bad.

src/box/txn_limbo.c
```diff
@@ -372,7 +376,7 @@ synchro_request_write(const struct synchro_request *req)
 	 * This is a synchronous commit so we can
 	 * allocate everything on a stack.
 	 */
-	char body[XROW_SYNCHRO_BODY_LEN_MAX];
+	char body[XROW_BODY_LEN_MAX];
```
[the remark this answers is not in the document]
sergepetrenko: Sure. Done

src/box/xrow.h
sergepetrenko: I've thought about this for some time and I think it would turn out quite ugly.

First of all, I'd like to keep the existing synchro request processing code intact. This means, the ownership request should be inherited from the base synchro request. So that I can pass either of them by pointer and let the processing code (`txn_limbo_req_prepare`/`txn_limbo_process`/`txn_limbo_req_commit`) decide how to interpret the input. This particular part would work ok.

Now, if I inherit the synchro ownership requrest from the base request, I want to have a single encoder/decoder for their common part. To avoid a case when we have to change two separate decoders when the base request is changed somehow.

I tried to implement these parts, and they look satisfactory. Not perfect, but ok. The main problem is that I either end up with going over the same request body twice (can be seen in patch below) or end up with two almost identical decoders/encoders (what I tried to avoid). Anyway, I attached the code below and IMO it already is questionable. And there are problems besides encoding/decoding which we haven't talked about yet.

There's one more possible approach: create a proxy struct holding all the possible synchro request field pointers, like you did with `struct replication_request`. But then we end up with the same decoder which accepts every possible field and every field is optional. I think this is what you wanted to avoid. So, it doesn't solve the problem, but adds complications of splitting request into two, just like the other approaches.

<details><summary>Here's what I've tried</summary>

```diff
diff --git a/src/box/xrow.c b/src/box/xrow.c
index 78ee40450..49fa04fc7 100644
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -1237,37 +1237,94 @@ xrow_encode_id(struct xrow_header *row)
 	row->body[0].iov_len = size;
 }
 
+uint32_t
+xrow_encode_synchro_body(char **pos, const struct synchro_request *req)
+{
+	/* Skip 1 byte for the yet unknown map size. */
+	++*pos;
+
+	uint32_t map_size = 0;
+	*pos = mp_encode_uint(*pos, IPROTO_REPLICA_ID);
+	*pos = mp_encode_uint(*pos, req->replica_id);
+	map_size++;
+	*pos = mp_encode_uint(*pos, IPROTO_LSN);
+	*pos = mp_encode_uint(*pos, req->lsn);
+	map_size++;
+	return map_size;
+}
+
 void
-xrow_encode_synchro(struct xrow_header *row, char *body,
-		    const struct synchro_request *req)
+xrow_encode_synchro(struct xrow_header *row, char *body, const struct synchro_request *req)
 {
 	assert(iproto_type_is_synchro_request(req->type));
 
 	char *pos = body;
+	uint32_t map_size = xrow_encode_synchro_body(&pos, req);
 
-	pos = mp_encode_map(pos,
-			    iproto_type_is_promote_request(req->type) ? 3 : 2);
+	mp_encode_map(body, map_size);
 
-	pos = mp_encode_uint(pos, IPROTO_REPLICA_ID);
-	pos = mp_encode_uint(pos, req->replica_id);
+	memset(row, 0, sizeof(*row));
+	row->type = req->type;
+	row->body[0].iov_base = body;
+	row->body[0].iov_len = pos - body;
+	row->bodycnt = 1;
+}
 
-	pos = mp_encode_uint(pos, IPROTO_LSN);
-	pos = mp_encode_uint(pos, req->lsn);
+void
+xrow_encode_synchro_ownership(struct xrow_header *row, char *body,
+			      const struct synchro_ownership_request *req)
+{
+	assert(iproto_type_is_promote_request(req->base.type));
+	char *pos = body;
+	uint32_t map_size = xrow_encode_synchro_body(&pos, &req->base);
 
-	if (iproto_type_is_promote_request(req->type)) {
-		pos = mp_encode_uint(pos, IPROTO_TERM);
-		pos = mp_encode_uint(pos, req->term);
-	}
+	pos = mp_encode_uint(pos, IPROTO_TERM);
+	pos = mp_encode_uint(pos, req->term);
+	map_size += 1;
 
-	assert(pos - body < XROW_BODY_LEN_MAX);
+	mp_encode_map(body, map_size);
 
 	memset(row, 0, sizeof(*row));
-	row->type = req->type;
 	row->body[0].iov_base = body;
 	row->body[0].iov_len = pos - body;
 	row->bodycnt = 1;
 }
 
+int
+xrow_decode_synchro_body(const char *data, struct synchro_request *req)
+{
+	if (mp_typeof(*data) != MP_MAP) {
+		return -1;
+	}
+
+	memset(req, 0, sizeof(*req));
+	uint32_t map_size = mp_decode_map(&data);
+	for (uint32_t i = 0; i < map_size; i++) {
+		enum mp_type type = mp_typeof(*data);
+		if (type != MP_UINT) {
+			mp_next(&data);
+			mp_next(&data);
+			continue;
+		}
+		uint8_t key = mp_decode_uint(&data);
+		if (key < iproto_key_MAX &&
+		    iproto_key_type[key] != mp_typeof(*data)) {
+			return -1;
+		}
+		switch (key) {
+		case IPROTO_REPLICA_ID:
+			req->replica_id = mp_decode_uint(&data);
+			break;
+		case IPROTO_LSN:
+			req->lsn = mp_decode_uint(&data);
+			break;
+		default:
+			mp_next(&data);
+		}
+	}
+	return 0;
+}
+
 int
 xrow_decode_synchro(const struct xrow_header *row, struct synchro_request *req)
 {
@@ -1275,49 +1332,63 @@ xrow_decode_synchro(const struct xrow_header *row, struct synchro_request *req)
 		diag_set(ClientError, ER_INVALID_MSGPACK, "request body");
 		return -1;
 	}
-
 	assert(row->bodycnt == 1);
 
-	const char *d = (const char *)row->body[0].iov_base;
-	if (mp_typeof(*d) != MP_MAP) {
+	memset(req, 0, sizeof(*req));
+
+	if (xrow_decode_synchro_body(row->body[0].iov_base, req) != 0) {
 		xrow_on_decode_err(row, ER_INVALID_MSGPACK, "request body");
 		return -1;
 	}
 
+	req->type = row->type;
+	req->origin_id = row->replica_id;
+	return 0;
+}
+
+int
+xrow_decode_synchro_ownership(const struct xrow_header *row,
+			      struct synchro_ownership_request *req)
+{
+	if (row->bodycnt == 0) {
+		diag_set(ClientError, ER_INVALID_MSGPACK, "request body");
+		return -1;
+	}
+	assert(row->bodycnt == 1);
+
 	memset(req, 0, sizeof(*req));
-	uint32_t map_size = mp_decode_map(&d);
+
+	const char *data = row->body[0].iov_base;
+	if (xrow_decode_synchro_body(data, &req->base) != 0) {
+		xrow_on_decode_err(row, ER_INVALID_MSGPACK, "request body");
+		return -1;
+	}
+	uint32_t map_size = mp_decode_map(&data);
 	for (uint32_t i = 0; i < map_size; i++) {
-		enum mp_type type = mp_typeof(*d);
+		enum mp_type type = mp_typeof(*data);
 		if (type != MP_UINT) {
-			mp_next(&d);
-			mp_next(&d);
+			mp_next(&data);
+			mp_next(&data);
 			continue;
 		}
-		uint8_t key = mp_decode_uint(&d);
+		uint8_t key = mp_decode_uint(&data);
 		if (key < iproto_key_MAX &&
-		    iproto_key_type[key] != mp_typeof(*d)) {
+		    iproto_key_type[key] != mp_typeof(*data)) {
 			xrow_on_decode_err(row, ER_INVALID_MSGPACK,
 					   "request body");
 			return -1;
 		}
 		switch (key) {
-		case IPROTO_REPLICA_ID:
-			req->replica_id = mp_decode_uint(&d);
-			break;
-		case IPROTO_LSN:
-			req->lsn = mp_decode_uint(&d);
-			break;
 		case IPROTO_TERM:
-			req->term = mp_decode_uint(&d);
+			req->term = mp_decode_uint(&data);
 			break;
 		default:
-			mp_next(&d);
+			mp_next(&data);
 		}
 	}
 
-	req->type = row->type;
-	req->origin_id = row->replica_id;
-
+	req->base.type = row->type;
+	req->base.origin_id = row->replica_id;
 	return 0;
 }
 
diff --git a/src/box/xrow.h b/src/box/xrow.h
index fd5c4d0fc..88c3cf572 100644
--- a/src/box/xrow.h
+++ b/src/box/xrow.h
@@ -352,6 +352,12 @@ struct synchro_request {
 	 * In case of PROMOTE it means CONFIRM(lsn) + ROLLBACK(lsn+1)
 	 */
 	int64_t lsn;
+};
+
+/** Synchronous queue ownership transfer request. PROMOTE or DEMOTE. */
+struct synchro_ownership_request {
+	/** Base synchro request. */
+	struct synchro_request base;
 	/**
 	 * The new term the instance issuing this request is in. Only used for
 	 * PROMOTE request.
```

</details>

Another complication is that I have to teach `synchro_request_write` to look at the request type and interpret it either as a plain synchro_request or as a ownership_request and coll the corresponding encoder.

To sum up, IMO this doesn't simplify anything. Although I think I can make the current patch version slightly better. Applied the following diff:
```diff
diff --git a/src/box/xrow.c b/src/box/xrow.c
index 0d67c2555..89e1b62df 100644
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -1245,32 +1245,32 @@ xrow_encode_synchro(struct xrow_header *row, char *body,
 
        char *pos = body;
 
-       int map_size = 2;
-       if (iproto_type_is_promote_request(req->type)) {
-               /* Term. */
-               map_size += 1;
-               /* Vclock. */
-               if (req->confirmed_vclock != NULL)
-                       map_size += 1;
-       }
-       pos = mp_encode_map(pos, map_size);
+       /* Skip one byte for the map. */
+       pos++;
+       uint32_t map_size = 0;
 
        pos = mp_encode_uint(pos, IPROTO_REPLICA_ID);
        pos = mp_encode_uint(pos, req->replica_id);
+       map_size++;
 
        pos = mp_encode_uint(pos, IPROTO_LSN);
        pos = mp_encode_uint(pos, req->lsn);
+       map_size++;
 
-       if (iproto_type_is_promote_request(req->type)) {
+       if (req->term != 0) {
                pos = mp_encode_uint(pos, IPROTO_TERM);
                pos = mp_encode_uint(pos, req->term);
-               if (req->confirmed_vclock != NULL) {
-                       pos = mp_encode_uint(pos, IPROTO_VCLOCK);
-                       pos = mp_encode_vclock_ignore0(pos,
-                                                      req->confirmed_vclock);
-               }
+               map_size++;
        }
 
+       if (req->confirmed_vclock != NULL) {
+               pos = mp_encode_uint(pos, IPROTO_VCLOCK);
+               pos = mp_encode_vclock_ignore0(pos, req->confirmed_vclock);
+               map_size++;
+       }
+
+       mp_encode_map(body, map_size);
+
        assert(pos - body < XROW_BODY_LEN_MAX);
 
        memset(row, 0, sizeof(*row));
```

test/replication-luatest/gh_5295_split_brain_test.lua
```diff
@@ -140,26 +149,80 @@ g.test_async_old_term = function(cg)
...
+    reconnect_and_check_no_split_brain(cg)
+end
+
+-- A conflicting sync transaction rollback from an obsolete term means a
 -- split-brain.
-g.test_rollback_old_term = function(cg)
+g.test_bad_rollback_old_term = function(cg)
+    t.tarantool.skip_if_not_debug()
+    -- Delay rollback on the old leader, so that the transaction is included
+    -- into new leader's PROMOTE, but the old leader writes a ROLLBACK for it.
```
Gerold103, line 192: `included into leader's promote` -> `rolled back by leader's PROMOTE`.

Gerold103 (approved):
Thanks for the patch, cool stuff!

src/box/xrow.h
Gerold103: I see, thanks for the try. My idea was a bit different though - I wanted us to intentionally duplicate some code. We would have two synchro request types, yes, and they would mostly match. But this is fine as long as their encode and decode could be done via single place, like `struct replication_request` or some other base request, like you said above.

It means that this

> First of all, I'd like to keep the existing synchro request processing code intact. This means, the ownership request should be inherited from the base synchro request.

wasn't needed. Yes, there would be some duplication, but there also would be clear isolation. It would be easier to read when you see a request and you know that it corresponds to one specific type of requests. `CONFIRM` and `PROMOTE` now are diverging, so reading this new request became a little bit harder.

Anyway, I am fine with the current solution for now as well. But I will raise this topic again if we would need to diverge those request types even further from each other.

Serpentian (approved):
Looks great!

src/box/xrow.h
sergepetrenko: > Yes, there would be some duplication, but there also would be clear isolation. It would be easier to read when you see a request and you know that it corresponds to one specific type of requests. CONFIRM and PROMOTE now are diverging, so reading this new request became a little bit harder.

Thanks for the explanation, I got your point better now.

test/replication-luatest/gh_5295_split_brain_test.lua
sergepetrenko: No, it's committed by the new leader, but rolled back by the old leader.

src/box/xrow.h
sergepetrenko: Resolving since you've approved the PR.

test/replication-luatest/gh_5295_split_brain_test.lua
sergepetrenko: Resolving, since you've approved the PR.

ylobankov (approved):

sergepetrenko:
Cherry-picked to 2.11: 77853befbd7feafcfc7c6997b8826f9bd8d8e688 .. 718aeb14b4b594ba074102a3bd3b685e698213b4

PR #9368, opened 2023-11-14
box: allow `lua_call` priv owners to call registered Lua functions

### LINKED ISSUE #9363: Allow to call a function with `lua_call` privilege disregarding of presence of the `_func` entry
Quote from the 'Context' section of the [Execute privilege for unregistered Lua functions](https://www.notion.so/tarantool/Execute-privilege-for-unregistered-Lua-functions-3191e0cf5c314349b6fff2eb6891b012) RFC document:

```lua
function my_func() end
box.schema.user.create('alice')

box.schema.user.grant('alice', 'execute', 'lua_call')
-- Now, alice can execute any Lua function accessible via _G except built-ins,
-- in particular my_func.

box.schema.func.create('my_func')
-- Now, alice can't execute my_func because it's registered so lua_call doesn't
-- work for it.
```

In my understanding, if `lua_call` is granted, it should be enough to call the function. Lack or presence of the `_func` entry shouldn't affect it.

Motivation: it simplifies upgrading from `execute, function` to `execute, lua_call` privileges and feels more natural per se.

There is a nuance. Everything is OK if the `_func` entry points to a Lua function by its name in `_G` (globals). It is to be defined what is better to do if this is a persistent function and if this is a C function.

locker:
Currently, the `lua_call` privilege simply doesn't work for registered functions. This patch fixes this issue: now, it grants access to any registered function that is written in Lua unless it's a built-in function, such as `dostring`, or a persistent function.

Note, this patch renames `access_check_call` to `access_check_lua_call` to avoid confusion because this function is now global. It also renames `access_check_eval` to `access_check_lua_eval` for consistency.

Closes #9363

### DIFF
--- a/src/box/call.c
+++ b/src/box/call.c
@@ -160,9 +160,8 @@ box_run_on_call(enum iproto_type type, const char *expr, int expr_len,
 	trigger_run(&box_on_call, &ctx);
 }
 
-/** Checks if the current user may execute any global Lua function. */
-static int
-access_check_call(const char *name, uint32_t name_len)
+int
+access_check_lua_call(const char *name, uint32_t name_len)
 {
 	struct credentials *cr = effective_user();
 	user_access_t access = PRIV_X | PRIV_U;
@@ -182,7 +181,7 @@ access_check_call(const char *name, uint32_t name_len)
 
 /** Checks if the current user may execute an arbitrary Lua expression. */
 static int
-access_check_eval(void)
+access_check_lua_eval(void)
 {
 	struct credentials *cr = effective_user();
 	user_access_t access = PRIV_X | PRIV_U;
@@ -229,7 +228,7 @@ box_process_call(struct call_request *request, struct port *port)
 			goto cleanup;
 		}
 	} else {
-		if (access_check_call(name, name_len) != 0) {
+		if (access_check_lua_call(name, name_len) != 0) {
 			rc = -1;
 			goto cleanup;
 		}
@@ -249,7 +248,7 @@ box_process_eval(struct call_request *request, struct port *port)
 {
 	rmean_collect(rmean_box, IPROTO_EVAL, 1);
 	/* Check permissions */
-	if (access_check_eval() != 0)
+	if (access_check_lua_eval() != 0)
 		return -1;
 	struct mp_box_ctx ctx;
 	if (mp_box_ctx_create(&ctx, NULL, request->tuple_formats) != 0)
--- a/src/box/func.c
+++ b/src/box/func.c
@@ -31,6 +31,7 @@
 #include "func.h"
 #include "fiber.h"
 #include "assoc.h"
+#include "call.h"
 #include "lua/call.h"
 #include "diag.h"
 #include "port.h"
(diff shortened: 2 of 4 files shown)


src/box/call.c
```diff
@@ -160,9 +160,8 @@ box_run_on_call(enum iproto_type type, const char *expr, int expr_len,
 	trigger_run(&box_on_call, &ctx);
```
Totktonada, line 160: Nit: Can we say 'available outside of its translation unit' (or kinda) instead of 'public'? It confuses me initially, because module API functions are also called 'public'.

> Note, this patch renames access_check_call to access_check_lua_call to
> avoid confusion because this function is now public.

src/box/func.c
```diff
@@ -567,6 +568,9 @@ func_access_check(struct func *func)
 	if ((func_access & PRIV_U) != 0 ||
 	    (func->def->uid != credentials->uid &&
 	     func_access & ~func->access[credentials->auth_token].effective)) {
+		if (func->def->language == FUNC_LANGUAGE_LUA)
+			return access_check_lua_call(func->def->name,
+						     func->def->name_len);
```
Totktonada, line 573: Shouldn't we check, whether the function is persistent?

src/box/call.c
locker: Sure. Replaced "public" with "global".

src/box/func.c
locker: After an offline discussion, agreed that `lua_call` shouldn't work for persistent Lua functions.

```diff
diff --git a/src/box/func.c b/src/box/func.c
index c52cdac81006..0dfea46254f0 100644
--- a/src/box/func.c
+++ b/src/box/func.c
@@ -568,7 +568,8 @@ func_access_check(struct func *func)
 	if ((func_access & PRIV_U) != 0 ||
 	    (func->def->uid != credentials->uid &&
 	     func_access & ~func->access[credentials->auth_token].effective)) {
-		if (func->def->language == FUNC_LANGUAGE_LUA)
+		if (func->def->language == FUNC_LANGUAGE_LUA &&
+		    func->def->body == NULL)
 			return access_check_lua_call(func->def->name,
 						     func->def->name_len);
 		/* Access violation, report error. */
diff --git a/test/box-luatest/gh_8803_exec_priv_test.lua b/test/box-luatest/gh_8803_exec_priv_test.lua
index 5aa3248650dc..fbb9a6067465 100644
--- a/test/box-luatest/gh_8803_exec_priv_test.lua
+++ b/test/box-luatest/gh_8803_exec_priv_test.lua
@@ -147,16 +147,16 @@ g.after_test('test_lua_call_func', function(cg)
 end)
 
 -- Checks that execute privilege granted on lua_call grants access to
--- Lua functions registered in _func except for built-ins.
+-- Lua functions registered in _func except for built-ins and persistent
+-- functions.
 g.test_lua_call_func = function(cg)
     local c = cg.conn
     local errfmt = "Execute access to function '%s' is denied for user 'test'"
     cg.grant('test', 'execute', 'lua_call')
-    for _, func in ipairs({'c_func', 'dostring'}) do
+    for _, func in ipairs({'c_func', 'stored_lua_func', 'dostring'}) do
         t.assert_error_msg_equals(errfmt:format(func), c.call, c, func)
     end
     t.assert(c:call('lua_func'))
-    t.assert(c:call('stored_lua_func'))
 end
 
 -- Checks that execute privilege granted on lua_call does not grant access
```

</details>

Totktonada (approved):
I'm not a permission professional, but I see no flaws. Thanks!

ylobankov (approved):

Issue #9340, opened 2023-11-07
`on_rollback` function parameter is empty for some types of rollbacks


Gumix:
**Bug description**

Currently there are 5 main reasons why a transaction can be aborted:

1. Explicit `box.rollback()`;
2. WAL write failure;
3. A conflict;
4. Fiber yield (without MVCC);
5. Timeout.

However `on_rollback` triggers are invoked with correct arguments only for 1-3, in other cases the trigger-function parameter is empty.

**Steps to reproduce**

```lua
local fiber = require('fiber')
local s = box.schema.space.create('test')
s:create_index('pk')

function on_rollback(iterator)
    print('on_rollback:')
    for num, old_tuple, new_tuple, space_id in iterator() do
        print(string.format('%d. space_id = %d, old_tuple = %s, new_tuple = %s',
                            num, space_id, tostring(old_tuple), tostring(new_tuple)))
    end
end

print('box.begin()')
box.begin()
  s:insert{1}
  s:insert{2}
  box.on_rollback(on_rollback)
  fiber.yield()
  pcall(s.insert, s, {3})
  pcall(s.insert, s, {4})
print('box.commit()')
pcall(box.commit)
print('the end')
```

**Actual behavior**

```
box.begin()
box.commit()
on_rollback:
the end
```

**Expected behavior**

```
box.begin()
box.commit()
on_rollback:
1. space_id = 512, old_tuple = nil, new_tuple = [2]
2. space_id = 512, old_tuple = nil, new_tuple = [1]
the end
```

unera:
Hi,

a user must access all changed tuples in the trigger in all cases.

unera:
If a tuple was passed to on_replace, before_commit, etc have to be passed to on_rollback.

Issue #7810, opened 2022-10-14
Why doesn't txn rollback invoke `on_replace` triggers for affected spaces?


Gerold103:
Rollback reverts all changes in reversed order. Makes `space:replace(old_tuple)` basically. Shouldn't these rollbacks also trigger `on_replace` triggers?

The question is mostly about savepoints problem. Assume I did a savepoint and then made a few changes. Then I rollback to the savepoint. From this moment the made changes won't appear in neither `on_commit` nor `on_rollback` triggers. For these statements I was unable to track whether they were committed or rolled back after all.

I faced this problem when trying to implement extra-secure control over `_bucket` content in vshard. I wanted to reject all invalid updates in `on_replace` + do some work like if a bucket turned from `active` to `sending` then I set `rw_lock = true`. If the transaction is rolled back (so the state goes back from `sending` to `active`), then I want to revert `rw_lock` back to `false`.

With the current API I would have been able to do it via `on_replace` + `on_rollback` if not for the savepoints. If I rollback this particular statement via a savepoint, then it is untrackable.

kostja:
Rollback can't fail. And shouldn't yield.

kostja:
Perhaps there could be on_savepoint_rollback.

PR #9370, opened 2023-11-14
txn: fix `on_rollback` trigger arguments for transactions aborted by yield or timeout

Gumix:
#### txn: pass `txn_stmt` instead of `txn` to `on_commit`/`on_rollback`

Currently `on_rollback` triggers are called on rollback of the whole transaction. To make it possible to invoke them on rollback to a savepoint, we need to pass a statement at which the savepoint was created.

#### txn: fix `on_rollback` trigger args for txns aborted by yield/timeout

Currently, if a transaction is aborted by a fiber yield or by a timeout, `txn_rollback_to_svp()` is called to rollback all statements of the txn. After that the transaction is completely aborted on attempt to commit it. If an `on_rollback` trigger is set, it is called from `txn_complete_fail()`, however it receives an empty iterator, because at this point the statements are already destroyed. This patch invokes `on_rollback` triggers directly from `txn_rollback_to_svp()` for abort-by-yield, and postpones the rollback for abort-by-timeout.

#### txn: run `on_rollback` triggers on rollback to savepoint

Currently, if a statement is rolled back during rollback to a savepoint, it does not appear in neither `on_commit` nor `on_rollback` triggers. Fix this by running `on_rollback` triggers during the rollback to the savepoint.

Closes https://github.com/tarantool/tarantool/issues/7810
Closes https://github.com/tarantool/tarantool/issues/9340

### DIFF
--- a/changelogs/unreleased/gh-9340-fix-on_rollback-function-args.md
+++ b/changelogs/unreleased/gh-9340-fix-on_rollback-function-args.md
@@ -1,0 +1,5 @@
+## bugfix/core
+
+* Fixed a bug when `on_rollback` trigger functions were invoked with an empty
+  iterator argument if a transaction was aborted by a fiber yield or by a
+  timeout (gh-9340).
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -579,9 +579,9 @@ lbox_txn_iterator_next(struct lua_State *L)
 }
 
 /**
- * Open an iterator over the transaction statements. This is a C
- * closure and 1 upvalue should be available - id of the
- * transaction to iterate over.
+ * Open an iterator over the transaction statements. This is a C closure and
+ * 2 upvalues should be available: first is an id of the transaction to iterate
+ * over, second is the first statement of the iteration.
  * It returns 3 values which can be used in Lua 'for': iterator
  * generator function, unused nil and the zero key.
  */
@@ -590,13 +590,14 @@ lbox_txn_pairs(struct lua_State *L)
 {
 	int64_t txn_id = luaL_toint64(L, lua_upvalueindex(1));
 	struct txn *txn = in_txn();
+	struct txn_stmt *stmt;
+	stmt = (struct txn_stmt *)lua_topointer(L, lua_upvalueindex(2));
 	if (txn == NULL || txn->id != txn_id) {
 		diag_set(ClientError, ER_CURSOR_NO_TRANSACTION);
 		return luaT_error(L);
 	}
 	luaL_pushint64(L, txn_id);
-	lua_pushlightuserdata(L, stailq_first_entry(&txn->stmts,
-						    struct txn_stmt, next));
+	lua_pushlightuserdata(L, stmt);
 	lua_pushcclosure(L, lbox_txn_iterator_next, 2);
 	lua_pushnil(L);
 	lua_pushinteger(L, 0);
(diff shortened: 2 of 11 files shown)


changelogs/unreleased/gh-9340-fix-on_rollback-function-args.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Fixed a bug when `on_rollback` trigger functions were invoked with an empty
+  iterator argument, if a transaction was aborted by a fiber yield or by a
+  timeout (gh-9340).
```
andreyaksenov, line 5:
```suggestion
* Fixed a bug when `on_rollback` trigger functions were invoked with an empty
  iterator argument if a transaction was aborted by a fiber yield or by a
  timeout (gh-9340).
```

andreyaksenov (approved):

ochaplashkin (approved):

src/box/lua/init.c
```diff
@@ -545,14 +545,14 @@ lbox_txn_iterator_next(struct lua_State *L)
...
-		return luaT_error(L);
-	}
 	struct txn_stmt *stmt =
 		(struct txn_stmt *) lua_topointer(L, lua_upvalueindex(2));
 	if (stmt == NULL)
 		return 0;
+	if (txn == NULL || txn->id != txn_id) {
+		diag_set(ClientError, ER_CURSOR_NO_TRANSACTION);
+		return luaT_error(L);
+	}
```
locker, line 555: Why did you move this check? Looks like an unrelated change.

src/box/lua/init.c
```diff
@@ -590,29 +590,33 @@ lbox_txn_pairs(struct lua_State *L)
...
  */
 static int
 lbox_push_txn(struct lua_State *L, void *event)
 {
-	struct txn *txn = (struct txn *) event;
-	luaL_pushint64(L, txn->id);
-	lua_pushcclosure(L, lbox_txn_pairs, 1);
+	struct txn_stmt *stmt = (struct txn_stmt *)event;
+	/* stmt is NULL for empty transactions. */
+	int64_t txn_id = stmt != NULL ? stmt->txn->id : -1;
```
locker, line 616: Since `stmt` can be `NULL`, wouldn't it be better to use `in_txn()` here and in other triggers instead of `stmt->txn`?

changelogs/unreleased/gh-9340-fix-on_rollback-function-args.md
Gumix: Fixed.

src/box/lua/init.c
Gumix: Left over from previous version of the patch, fixed.

src/box/lua/init.c
Gumix: Done.

Gumix:
Also changed an `if` to `txn_on_timeout` to optimize memory usage.

src/box/txn.c
```diff
@@ -1497,10 +1523,16 @@ txn_on_yield(struct trigger *trigger, void *event)
...
+		txn_rollback_to_svp(txn, NULL, true);
 		txn->status = TXN_ABORTED;
 		txn_set_flags(txn, TXN_IS_ABORTED_BY_YIELD);
 		say_warn("Transaction has been aborted by a fiber yield");
+		if (txn_has_flag(txn, TXN_HAS_TRIGGERS)) {
+			/* Can't rollback more than once. */
+			trigger_destroy(&txn->on_rollback);
+			/* Commit won't happen after rollback. */
+			trigger_destroy(&txn->on_commit);
+		}
```
Gerold103, line 1535: From that moment you have the destroyed triggers here but the flag `TXN_HAS_TRIGGERS` is still set. Either the triggers need to be re-created empty, or the flag needs to be unset.

Gerold103 (approved):
Thanks for the patch!

src/box/txn.c
locker: AFAIU after `trigger_destroy` is called, the trigger list becomes empty, but yeah, it's probably better to reinitialize it explicitly.

Issue #4349, opened 2019-07-15
Test transactional DDL more thoroughly


locker:
A test plan by @kostja:

1. A transaction which swaps two spaces by name.
```
name = foo.name;
foo.name = bar.name
bar.name = name
```

2. A transaction which creates a space, formats it to have an integer column, changes the format to 'any', changes the values to 'string', changes the format to 'string'

3. A transaction which creates a space, a user, and grants all privileges on this space to the user. Then sudoes to the user and tries to use the space.

4. A transaction, which revokes all privileges from a user and drops the space.

5. A transaction which creates a space and secondary indexes

6. A transaction which creates a space with a sequence.

7. Transactional drop a) empty b) nonempty space with all its belongings: secondary indexes, check constraints, sequences.

5. A transaction which creates a space and sets space triggers, as well as on_rollback trigger, then inserts a bunch of data and then a) commits b) rolls back.

6. An error injection crash test which starts 10 fibers which perform a random combination of transactional DDL in a loop, then, after 0.1 second ER_WAL_IO is enabled, fibers are joined, ER_WAL_IO is disabled. The DDL is using objects which names clash,  so that it does something useful, e.g:

```Lua
   function name(object_type)
        return object_type_name ..  math.random(1, NUM_NAMES)
    end

    function ddl(object_type, object_name) 
        -- this function actually checks if the name exists 
        --  at the moment and only returns valid statements
        if object_type is user or role then
            return random("grant", "revoke", "create", "drop")
        end
        if object_type is table then 
            return random("create ", "format", "rename", "add
            index", "drop index", "rename index", "change index
            keys", "change unique -> non_unique", "add
            constraint", "drop constraint", "drop")
        end


    function gen_test()
        for each object type:
            eval = "box.begin()"
            eval = eval .. ddl(object_type, name(object_type))
            eval = eval .. "box.commit()"
        end
        return eval
    end
    function fiber()
        while true do
            eval(gen_test)
        end
    end

    for i in 1.. NUM_FIBERS do
        fibers[i] = fiber.create(fiber)
    end
    fiber.sleep(0.1)
    errinj.enable(er_wal_io)
    for i in 1.. NUM_FIBERS do
        fiber.join(fibers[i])
    end
    (cleanup all objects matching the name..$i pattern)
```

Follow-up #4083

dannote:
@kostja I doubt it's okay to put an unreproducible fuzzy test lasting for a magic constant time along with other casual unit tests.

[...]

PR #9371, opened 2023-11-15
test: add the extended transactional DDL tests

mkostoevr:
Added more tests for various scenarios of the transactional DLL usage:
1. A DDL transaction which creates a space, formats it to have an integer column, fills it with some data, changes the format to 'any', changes the data values to 'string', and changes the format to 'string'.
2. A DDL transaction that creates a space, user, grants privileges to the user to use the new space and accesses the space on behalf of the new user.
3. Creation of a space and its secondary indexes in one transaction.
4. Creation of a space with a sequence attached in one transaction.
5. A DDL transaction which sets the on_rollback trigger and creates a space, the space triggers, inserts a bunch of data and then commits or rolls back.
6. Transactional drop of a space with various indexes, field and tuple constraints and a sequence attached. The formers are also created in a single transaction.

Part of #4349

NO_DOC=test
NO_CHANGELOG=test

### DIFF
--- a/test/box-luatest/gh_4349_transactional_ddl_test.lua
+++ b/test/box-luatest/gh_4349_transactional_ddl_test.lua
@@ -1,0 +1,257 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g1 = t.group('generic')
+local g2 = t.group('two_way', {{flag = true}, {flag = false}})
+
+for _, g in pairs({g1, g2}) do
+    g.before_all(function(cg)
+        cg.server = server:new()
+        cg.server:start()
+    end)
+
+    g.after_all(function(cg)
+        cg.server:drop()
+    end)
+
+    g.after_each(function(cg)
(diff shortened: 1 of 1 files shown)


ochaplashkin (approved):

ylobankov (approved):

test/box-luatest/gh_4349_transactional_ddl_test.lua
```diff
@@ -0,0 +1,263 @@
...
+        end)
+
+        local s = box.schema.space.create('s')
+
+        for _, trigger in pairs(s:on_replace()) do
+            -- Drop the old on_replace triggers. This is required because the
+            -- triggers are bound to space IDs, and they're not deleted on the
+            -- space drop.
+            s:on_replace(nil, trigger)
+        end
```
locker, line 159: Better use the new trigger API - see the module `trigger`.

locker (approved):

PR #9391, opened 2023-11-22
cmake: use legacy layout for static-build bundled libs

### LINKED ISSUE #9242: Add facility to build tarantool statically via a cmake option

locker:
SVACE stopped working after commit 98b38e89ba054 changed the bundled libs directory layout. To fix this, let's introduce the new cmake option `BUNDLED_LIBS_INSTALL_DIR` and set it in `static-build/CMakeLists.txt` to the legacy location. Also, let's use the legacy directories for each external project's `PREFIX`, `SOURCE_DIR`, `BINARY_DIR`, and `STAMP_DIR`.

Follow-up #9242

EE part: https://github.com/tarantool/tarantool-ee/pull/595


cmake/BuildICONV.cmake
```diff
@@ -1,6 +1,6 @@
 set(ICONV_VERSION 1.17)
 set(ICONV_HASH d718cd5a59438be666d1575855be72c3)
-set(ICONV_INSTALL_DIR ${PROJECT_BINARY_DIR}/build/iconv)
+set(ICONV_INSTALL_DIR ${BUNDLED_LIBS_INSTALL_DIR}/iconv-prefix)
```
ylobankov, line 3: Shouldn't we do the same for openssl?

locker: Missed it. Fixed. Thanks.

ylobankov (approved):

locker:

PR #9408, opened 2023-11-24
box.ctl: add security.iproto_lockdown option

Lord-KA:
The new option is backed by `box.ctl.iproto_lockdown`. It is available only in Enterprise Edition builds.

Needed for tarantool/tarantool-ee#585

NO_DOC=will be added to Enterprise Edition
NO_CHANGELOG=see NO_DOC

### DIFF
--- a/src/box/lua/ctl.c
+++ b/src/box/lua/ctl.c
@@ -45,6 +45,7 @@
 #include "box/engine.h"
 #include "box/memtx_engine.h"
 #include "box/raft.h"
+#include "box/security.h"
 
 #include "core/event.h"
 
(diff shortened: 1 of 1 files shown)


Totktonada:
As far as I understand, the lockdown mode is a result of a execution of an application or a role initialization logic or (in rare cases) execution of a request. It is not to be configured, it just *occurs*.

We need an admin interface to forcely exit from the lockdown mode. However, it doesn't look as a configuration interface (such as `box.cfg` or `config`). It looks more like a command interface (such as `box.ctl`).

@Lord-KA @igormunkin Please, consider this idea of the API.

Lord-KA:
We decided to move the option to `box.ctl`, as it is more of an internal feature than a user-configurable one.

ochaplashkin (approved):

igormunkin (approved):
@Lord-KA, thanks for the patch! LGTM as trivial.

PR #9410, opened 2023-11-24
config: introduce options for storage, add source file to the build

askalt:
This patchs adds config options for tarantool config storage, as fhis feature will be implemented in EE.
Also, adds the source file to the build.

Part of https://github.com/tarantool/tarantool-ee/issues/593

### DIFF
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -64,6 +64,7 @@ lua_source(lua_sources lua/config/utils/tabulate.lua      config_utils_tabulate_
 if (ENABLE_CONFIG_EXTRAS)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/etcd.lua config_source_etcd_lua)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/storage/init.lua config_storage_init_lua)
+    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/storage.lua config_source_storage_lua)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/extras.lua config_extras_lua)
 endif()
 # }}} config
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -484,6 +484,83 @@ return schema.new('instance_config', schema.record({
                 end
             end,
         })),
+        storage = enterprise_edition(schema.record({
+            prefix = schema.scalar({
+                type = 'string',
+                validate = function(data, w)
+                    if not data:startswith('/') then
+                        w.error(('config.storage.prefix should be ' ..
+                            'a path alike value, got %q'):format(data))
+                    end
+                end,
+            }),
+            endpoints = schema.array({
+                items = schema.record({
+                    uri = schema.scalar({
+                        type = 'string',
+                    }),
+                    login = schema.scalar({
+                        type = 'string',
+                    }),
+                    password = schema.scalar({
+                        type = 'string',
+                    }),
+                    params = schema.record({
+                        transport = schema.enum({
+                            'plain',
+                            'ssl',
+                        }),
+                        ssl_key_file = enterprise_edition(schema.scalar({
+                            type = 'string',
+                        })),
+                        ssl_cert_file = enterprise_edition(schema.scalar({
+                            type = 'string',
+                        })),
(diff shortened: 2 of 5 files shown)


src/box/lua/config/instance_config.lua
```diff
@@ -454,6 +454,66 @@ return schema.new('instance_config', schema.record({
...
+                ciphers = schema.scalar({
+                    type = 'string',
+                }),
+                password = schema.scalar({
+                    type = 'string',
+                }),
+                password_file = schema.scalar({
+                    type = 'string',
+                })
+            }),
```
Totktonada, line 506: Several points we discussed verbally with @askalt.

1. The idea was to attempt to make the options similar to `config.etcd`, but it seems that `net.box` options are too different. So, there is not consistency with `config.etcd` anyway and it worth to try to make the options consistent with `iproto.listen` and `iproto.advertise.{peer,sharding}`.
2. It is OK to split the tasks and implement the storage without TLS options support first and add them in a separate PR.
3. Timeout options (`connect_timeout` in this case) should accept a floating point number. The type in the schema should be changed from `integer` to `number`.
4. It is desirable to have meaningful defaults for timeout options, not infinity.

src/box/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ lua_source(lua_sources lua/config/utils/tabulate.lua      config_utils_tabulate_
 
 if (ENABLE_CONFIG_EXTRAS)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/etcd.lua config_source_etcd_lua)
+    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/tarantool.lua config_source_tarantool_lua)
```
Totktonada, line 63: There is no word about the new internal EE module in the commit message/pull request description. I think that it can be extracted into its own commit (within the PR). The PR title/description should reflect this change somehow.

src/box/lua/config/instance_config.lua
oleg-jukovec: I suggest to rename `connect_timeout` to `timeout` and use it as a connect timeout and a request timeout to make things easier.

askalt: Updated.

src/box/CMakeLists.txt
askalt: Updated.

src/box/lua/config/instance_config.lua
```diff
@@ -484,6 +484,77 @@ return schema.new('instance_config', schema.record({
...
+                end,
+            }),
+            timeout = schema.scalar({
+                type = 'number',
+                default = 3,
+            }),
+            reconnect_after = schema.scalar({
+                type = 'number',
+                default = 3,
+            })
```
Totktonada, line 547: It seems, if `config.storage` is missed in config and `:apply_default()` is called, `:validate()` doesn't pass on the resulting data. See a comment for `config.etcd`.

It seems, we're lucky that it is not called in this order on our testing scenarios.

We can either move defaults to the source (and leave them here as comments, just like in `config.etcd.http.request.timeout`. Or keep the defaults in the schema and change the `validate` annotation logic.

I would prefer the second. Let's just ensure that either `prefix` and `endpoints` are set both or unset both. It matches the source enabling criteria in the EE part: if `endpoints` is `nil` the source is disabled.

src/box/lua/config/instance_config.lua
```diff
@@ -484,6 +484,77 @@ return schema.new('instance_config', schema.record({
                 end
             end,
         })),
+        storage = enterprise_edition(schema.record({
+            prefix = schema.scalar({
+                type = 'string',
+                validate = function(data, w)
+                    if not data:startswith('/') then
+                        w.error(('config.tarantool.prefix should be ' ..
+                            'a path alike value, got %q'):format(data))
```
Totktonada, line 493: Nit: config.tarantool.prefix -> config.storage.prefix.

Totktonada (approved):

src/box/lua/config/instance_config.lua
askalt: Changed.

src/box/lua/config/instance_config.lua
askalt: Thank you! Checks added

PR #9412, opened 2023-11-27
replication: make yield more often when recovering from long WAL

### LINKED ISSUE #9094: A replica may timeout while master is scanning xlogs
**Bug description**

If the master has a long enough xlog, it's possible that it reads it for more than `replication_disconnect_timeout` without sending heartbeats. In this case replica will drop the connection by a timeout and retry connecting, entering an infinite reconnect loop.
This issue was already found and fixed before: ee6de025f7ff4a6fd8e60efafcc78bf613f1686d.

But now it manifests itself again. Commit 56571d83172f made relay stop sending heartbeats, when tx is unresponsive, but that shouldn't happen when WALs are scanned. There is no way to check if tx is responsive while WAL is scanned, and after replication_disconnect_timeout of reading WALs heartbeats are stopped.

We shouldn't stop sending heartbeats during WAL scan, probably.

yanshtunder:
As soon as replica subscribes, relay starts scanning the first requested WAL file to find the position from which to start replication. If a replica connects and needs data from the end of a really long WAL, recovery will read up to the needed position without yields, making relay disconnect by replication_disconnect_timeout. This issue was already found and fixed before: ee6de02. To fix this problem, make sure that when recovering, a yield occurs after each WAL row. In this case, the recovery time from WAL will increase by approximately 50 percent.

Closes #9094

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
+++ b/changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
@@ -1,0 +1,5 @@
+# bugfix/replication
+
+* Now at replication no stop sending heartbeats during WAL scan. This prevents
+  relay from timing out when a freshly subscribed replica needs rows from the
+  end of a long `.xlog` (gh-9094).
--- a/test/small
+++ b/test/small
@@ -1,1 +1,1 @@
-../src/lib/small/test/
+/home/yan/Projects/Mail/Tarantool/tarantool/test/../src/lib/small/test/
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -218,6 +218,7 @@ applier_thread_writer_f(va_list ap)
 			ERROR_INJECT(ERRINJ_APPLIER_SLOW_ACK, {
 				fiber_sleep(0.01);
 			});
+			//say_info("Sent ACK");
 			/*
 			 * Even if new ACK is requested during the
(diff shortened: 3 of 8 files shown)


changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
```diff
@@ -0,0 +1,5 @@
+# bugfix/replication
+
+* Now at replication no stop sending heartbeats during WAL scan. This prevents
+  relay from timing out when a freshly subscribed replica needs rows from the
+  end of a long `.xlog` (gh-9094).
```
p7nov, line 5:
```suggestion
* Now Tarantool sends heartbeats during WAL scan when starting replication
  to a freshly subscribed replica. This prevents relay from timing out when
  the replica needs rows from the end of a long `.xlog` (gh-9094).
```

p7nov (approved):

test/small
```diff
@@ -1 +1 @@
-../src/lib/small/test/
\ No newline at end of file
+/home/yan/Projects/Mail/Tarantool/tarantool/test/../src/lib/small/test/
```
ochaplashkin, line 1: It seems that this is not the path that we want to see in the end.

ochaplashkin (changes requested):

PR #9413, opened 2023-11-27
lua: prevent serialization of error for ucdata

Buristan:
The alternative solution is to use just `lua_call()` instead, since the error should be caught somewhere in the upper frames.
```diff
diff --git a/src/lua/serializer.c b/src/lua/serializer.c
index 1018f9572..56400e7f3 100644
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -274,7 +274,11 @@ lua_field_inspect_ucdata(struct lua_State *L, struct luaL_serializer *cfg,
                 }
                 /* copy object itself */
                 lua_pushvalue(L, idx);
-                lua_pcall(L, 1, 1, 0);
+                /*
+                 * The possible error should be catched in some
+                 * upper frame.
+                 */
+                lua_call(L, 1, 1);
                 /* replace obj with the unpacked value */
                 lua_replace(L, idx);
                 if (luaL_tofield(L, cfg, idx, field) < 0)
```

The results of the benchmark below for both ways are quite similar, so I prefer the most common way (the one in the patch):
```sh
time src/tarantool -e "                                                                
local yml = require'yaml'
local ud_mt = {__serialize = function()
  return ''
end}
ud_mt.__index = ud_mt
local ud = newproxy(true)
debug.setmetatable(ud, ud_mt)
collectgarbage('stop')
for _ = 1, 3e6 do
    yml.encode(ud)
end
"  
```

|`lua_pcall()` | `lua_call()` |
|--|--|
| $4.896s\ (\sigma=0.037$) | $4.879s\ (\sigma=0.034$) |

### DIFF
--- a/changelogs/unreleased/gh-9396-ucdata-serialize-error.md
+++ b/changelogs/unreleased/gh-9396-ucdata-serialize-error.md
@@ -1,0 +1,4 @@
+## bugfix/lua
+
+* An error from a serializer function for cdata and userdata is not ignored now
+  (gh-9396).
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -274,7 +274,10 @@ lua_field_inspect_ucdata(struct lua_State *L, struct luaL_serializer *cfg,
 		}
 		/* copy object itself */
 		lua_pushvalue(L, idx);
-		lua_pcall(L, 1, 1, 0);
+		if (lua_pcall(L, 1, 1, 0) != LUA_OK) {
+			diag_set(LuajitError, lua_tostring(L, -1));
+			return -1;
+		}
 		/* replace obj with the unpacked value */
 		lua_replace(L, idx);
 		if (luaL_tofield(L, cfg, idx, field) < 0)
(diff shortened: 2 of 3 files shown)


p7nov (approved):

ochaplashkin (approved):

igormunkin (approved):
@Buristan, thanks for the patch! LGTM.

Lord-KA (commented):
Thanks for the patch. I don't see any issues with the change, LGTM.

Nit: @Buristan, I am just curious, why did you leave the `XXX` in the test comments?

Buristan:
@Lord-KA, 
AFAIK, `XXX` is just a legacy keyword coming [from Sun code convention](https://www.oracle.com/java/technologies/javase/codeconventions-programmingpractices.html):
> 10.5.4 Special Comments
Use `XXX` in a comment to flag something that is bogus but works. Use `FIXME` to flag something that is bogus and broken.

Lord-KA:
Personally, I don't see anything wrong with using `yaml.encode()` in this reproducer, but sure.

igormunkin:
Also backported to 2.11 in scope of 074fe0b. To apply the changes for 2.10, @Buristan will create a separate PR with the corresponding updates.

PR #9415, opened 2023-11-27
main: don't break graceful shutdown on init script exit

### LINKED ISSUE #9411: Graceful shutdown exit before shutdown timeout expiration
Tarantool version is 3.0.0-beta1-11-g3ee68d8ba.

Repro script:

```lua

box.ctl.on_shutdown(function()
    fiber.sleep(0.2)
    print('shutdown callback finished')
end, nil)
box.ctl.set_on_shutdown_timeout(1)

fiber.create(function()
    os.exit(0)
end)

fiber.sleep(0.1)
```

The script is expected to print `shutdown callback finished` but it does not.

nshy:
Graceful shutdown is done in a special fiber which is started for example on `SIGTERM`. So it can run concurrently with fiber executing Tarantool init script. On init fiber exit we break event loop to pass control back to the Tarantool initialization code. But we fail to run event loop a bit more to finish graceful shutdown.

The test is a bit contrived. A more real world case is when Tarantool is termintated during lingering `box.cfg()`.

Close #9411


locker (approved):

ochaplashkin (approved):

p7nov (approved):

locker:
Cherry-picked to 2.11.

PR #9417, opened 2023-11-27
box: add integrity check option

Lord-KA:
The new module can be enable only in the Enterprise Edition builds via `--integrity-check` cli option.

Needed for tarantool/tarantool-ee#585

NO_DOC=will be added to Enterprise Edition
NO_CHANGELOG=see NO_DOC
NO_TEST=see NO_DOC


ochaplashkin (approved):

Totktonada (approved):
I've no full picture of the activity about integrity checks, but given the code that is proposed here, I've no objections.

test/luatest_helpers/server.lua
```diff
@@ -105,7 +105,10 @@ Server.constructor_checks = fun.chain(Server.constructor_checks, {
 function Server:initialize()
     if self.config_file ~= nil then
         self.command = arg[-1]
-        self.args = {'--name', self.alias}
+
+        self.args = self.args or {}
+        table.insert(self.args, '--name')
+        table.insert(self.args, self.alias)
```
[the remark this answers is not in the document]
Lord-KA: Thanks for the suggestion! Fixed.

src/box/lua/config/source/file.lua
```diff
@@ -69,6 +69,18 @@ function methods.sync(self, config_module, _iconfig)
     assert(config_module._config_file ~= nil)
 
     local data = universal_read(config_module._config_file, 'config file')
+
+    do
```
igormunkin, line 73: Minor: Could you clarify why `do ... end` is used here?

src/box/lua/integrity.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#include "trivia/config.h"
+
+#if defined(ENABLE_INTEGRITY)
```
igormunkin, line 10: Minor: Why not just use `#ifdef` here?

src/main.cc
```diff
@@ -813,6 +818,18 @@ main(int argc, char **argv)
 		case 'F':
 			box_is_force_recovery = true;
 			break;
+		case 'I':
+#if ENABLE_INTEGRITY
+			opt_mask |= O_INTEGRITY;
+			instance.hashes = (const char *)xstrdup(optarg);
```
igormunkin, line 824: There is no need in `xstrdup` here: it is used for `name`/`config` to make it consistent with `getenv_safe`, hence `instance.*` fields are released below. So, I suggest to just copy the pointer, but if the contents are required to be copied, please release the copied content.

src/trivia/config.h.cmake
```diff
@@ -290,6 +290,7 @@
 #cmakedefine ENABLE_ETCD_CLIENT 1
 #cmakedefine ENABLE_CONFIG_EXTRAS 1
 #cmakedefine ENABLE_FAILOVER 1
+#cmakedefine ENABLE_INTEGRITY 1
```
igormunkin, line 293: Minor: You can use `#cmakedefine01 ENABLE_INTEGRITY` to use `#if` instead of `#ifdef`. Feel free to ignore.

igormunkin (approved):
**[PATCH 2/2] box: add integrity check option**

@Lord-KA, thanks for the patch! LGTM, but please consider my comments below.

igormunkin (approved):
**[PATCH 1/2] test: allow user-defined args in luatest_helpers.server**

@Lord-KA, thanks for the patch! LGTM.

src/box/lua/config/source/file.lua
Lord-KA: I felt, that this code is alien to the place it was inserted, but now it seems like an overreaction, so removed.

src/box/lua/integrity.h
Lord-KA: I wanted to use the same style as for the other EE modules, for more details of this style see https://github.com/tarantool/tarantool/pull/8621.

src/main.cc
Lord-KA: Right, fixed that.

src/trivia/config.h.cmake
Lord-KA: Ignored for consistency reasons.

src/box/lua/config/source/file.lua
igormunkin: Well, honestly such splitting into blocks is widely used in Lua world, so I was just asking for your reasons... It was much better before, but if you feel it was not, I'll not insist.

src/box/lua/integrity.h
igormunkin: OK, thanks for clarification!

PR #9418, opened 2023-11-28
config: allow to configure an anonymous replica

### LINKED ISSUE #9432: config: support anonymous replicas
There is a couple of problems we need to resolve regarding anonymous replicas in the declarative configuration code.

* An anonymous replica shouldn't attempt to persist its name in `_cluster`.
* An anonymous replica shouldn't be used as an upstream for non-anonymous replica.
* (I would also don't use anonymous replicas as upstreams for other anonymous replicas by default.)
* An anonymous replica shouldn't be chosen as a boostrap leader in `replication.failover: supervised` mode.
* Verify that an anonymous replica is not configured in read-write mode (using `database.mode` or `<repicaset>.leader` options).
* Verify that an anonymous replica is not configured with `replication.election_mode != off`.
* An anonymous replica should be able to be bootstrapped from a replicaset, where all the instances are in read-only mode (because it doesn't need a write to `_cluster` to join the replicaset).

Totktonada:
`replication.anon: true` option is now working.

Mostly, it is the only thing that a user needs to know about this changeset.

However, technically speaking, several related configuration checks are introduced here and dependent defaults were adjusted to make usage of anonymous replicas smoother.

The new configuration constrainsts are the following.
    
* A replicaset must contain at least one non-anonymous instance.
* An anonymous replica can't be configured as writable instance using `database.mode` or `<replicaset>.leader` options.
* An anonymous replica can't be configured with `replication.election_mode` equals to `candidate`, `voter` or `manual` (only `off` is allowed).

A few more nuances about anonymous replicas:

* Anonymous replicas are filtered out from default upstream list.
* A `replication.failover: election` replicaset can contain anonymous replicas, but `replication.election_mode` defaults to `off` for them (unlike non-anonymous instances, where the default is `candidate`).
* `replication.failover: supervised` skips anonymous replicas, when choosing a bootstrap leader.
* A anonymous replica can joined a replicaset, which has all the instances in read-only mode (unlike a non-anonymous instance).

Please, check out commits of the patchset for more detailed description of the constrains and nuances.

Fixes #9432

### DIFF
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -5,9 +5,6 @@
 
   There are caveats that are not resolved yet:
 
-  * An attempt to configure a replicaset where all instances are anonymous
-    replicas should lead to an error on config validation, before configuration
-    applying.
   * An attempt to configure an anonymous replica in read-write mode (using
     `database.mode` or `<replicaset>.leader`) should lead to an error on config
     validation, before configuration applying.
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -5,8 +5,6 @@
 
   There are caveats that are not resolved yet:
 
-  * An anonymous replica shouldn't be chosen as a bootstrap leader in
-    `replication.failover: supervised` mode.
   * An attempt to configure a replicaset where all instances are anonymous
     replicas should lead to an error on config validation, before configuration
     applying.
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -1,0 +1,25 @@
+## bugfix/config
+
+* Effectively supported `replication.anon` option, which was broken in several
+  ways before (gh-9432).
+
+  There are caveats that are not resolved yet:
+
+  * An anonymous replica shouldn't be used as an upstream for non-anonymous
+    replica.
+  * An anonymous replica possibly shouldn't be used as an upstream for other
+    anonymous replicas by default.
+  * An anonymous replica shouldn't be chosen as a bootstrap leader in
+    `replication.failover: supervised` mode.
+  * An attempt to configure a replicaset where all instances are anonymous
+    replicas should lead to an error on config validation, before configuration
+    applying.
+  * An attempt to configure an anonymous replica in read-write mode (using
+    `database.mode` or `<replicaset>.leader`) should lead to an error on config
+    validation, before configuration applying.
+  * An attempt to configure an anonymous replica with
+    `replication.election_mode` != `off` should lead to an error on config
+    validation, before configuration applying.
+  * An anonymous replica can't be bootstrapped from a replicaset, where all the
+    instance are in read-only mode, however there are no technical problems
+    there (just too tight validation).
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -2,9 +2,3 @@
 
 * Effectively supported `replication.anon` option, which was broken in several
   ways before (gh-9432).
-
-  There are caveats that are not resolved yet:
-
-  * An anonymous replica can't be bootstrapped from a replicaset, where all the
-    instance are in read-only mode, however there are no technical problems
-    there (just too tight validation).
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -5,10 +5,6 @@
 
   There are caveats that are not resolved yet:
 
-  * An anonymous replica shouldn't be used as an upstream for non-anonymous
-    replica.
-  * An anonymous replica possibly shouldn't be used as an upstream for other
-    anonymous replicas by default.
   * An anonymous replica shouldn't be chosen as a bootstrap leader in
     `replication.failover: supervised` mode.
   * An attempt to configure a replicaset where all instances are anonymous
--- a/changelogs/unreleased/config-anonymous-replica.md
+++ b/changelogs/unreleased/config-anonymous-replica.md
@@ -5,12 +5,6 @@
 
   There are caveats that are not resolved yet:
 
-  * An attempt to configure an anonymous replica in read-write mode (using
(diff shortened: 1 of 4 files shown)


src/box/lua/config/configdata.lua
```diff
@@ -486,6 +486,82 @@ local function new(iconfig, cconfig, instance_name)
...
+    -- The default 'off' is not applied, because the explicit
+    -- 'candidate' value is set in the replicaset scope. It can be
+    -- fixed like so:
+    --
+    -- <...>
+    --       i-004:
+    --         replication:
+    --           anon: true
+    --           election_mode: off # !!
+    if failover ~= 'election' then
```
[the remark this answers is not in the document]
Totktonada: I agree. Done in PR #9433.

test/config-luatest/replicaset_test.lua
```diff
@@ -0,0 +1,712 @@
...
+
+-- {{{ General purpose helpers
+
+local function toline(s)
+    return s:gsub('\n', ''):gsub(' +', ' '):strip()
+end
+
+-- }}} General purpose helpers
+
+-- {{{ Config builder
```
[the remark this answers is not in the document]
Totktonada: Thanks!

Now, we have `test.luatest_helpers.server`, which is compatible extension of `luatest.server` and should go into luatest itself sooner or later.

However, the replicaset helper is different. It has almost nothing common with `luatest.replica_set` and I would not attempt to make it look like `luatest.replica_set`. So, I'm in doubt regarding `test/luatest_helpers` directory.

However, the utils itself are helpful. I decided to place `cbuilder` and `replicaset` into `test/config-luatest`.

Done in PR #9433.

Please, note that PR #9433 has only part of the replicaset helper. I'll add all the other functions within PR #9418 (this PR) at the next update.

src/box/lua/config/configdata.lua
```diff
@@ -530,7 +606,109 @@ local function new(iconfig, cconfig, instance_name)
...
+    -- in different ways, so we need specific checks for each of
+    -- them in regard of an anonymous replica.
+    --
+    -- These checks don't verify the whole cluster config, only
+    -- the given replicaset.
+
+    -- failover: <any>
+    --
+    -- A replicaset can't consist of only anonymous replicas.
+    assert(#peer_names > 0)
```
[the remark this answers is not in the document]
Totktonada: Sure.

Also, moved it before `supervised` mode logic, which chooses a bootstrap leader, because it assumes that there is at least one non-anonymous instance in the given replicaset.

test/config-luatest/replicaset_test.lua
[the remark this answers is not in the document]
Totktonada: Now, PR #9433 has all the new testing helpers (it appears that I need to test it more intensively). Let's discuss them in PR #9433. Closing the discussion.

src/box/lua/config/configdata.lua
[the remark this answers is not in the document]
Totktonada: Closing the discussion, because it is now relevant to PR #9433 and not relevant to this PR.

changelogs/unreleased/config-anonymous-replica.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/config
+
+* Support `replication.anon` option (gh-9432).
```
p7nov, line 3:
```suggestion
* Introduced the `replication.anon` configuration option (gh-9432).
```

p7nov (approved):

changelogs/unreleased/config-anonymous-replica.md
Totktonada: In fact, the option was there, but an attempt to use it would reveal existing problems.

Totktonada: Moreover, 'introduced' is suggested for significant features in our changelog writing guide. I wouldn't qualify it so.

Totktonada: We discussed it f2f and agreed on the following wording.

> * Effectively supported `replication.anon` option, which was broken in several ways before (gh-xxxx).

Totktonada: Updated. Thanks!

changelogs/unreleased/config-anonymous-replica.md
```diff
@@ -2,9 +2,3 @@
 
 * Effectively supported `replication.anon` option, which was broken in several
   ways before (gh-9432).
-
-  There are caveats that are not resolved yet:
```
ImeevMA: I find it a little strange to add this list to the changelog and then remove it at the end of the PR. Since the items in the list don't look large enough to have their own changelog entry, it would be more natural to have them in the description of the PR and/or in the commit message of the first commit. Although it's just a thought, feel free to ignore it.

ImeevMA (approved):
Thanks for the PR! I don't see any problem with PR, except that it seems a little strange to add a list to the changelog and then remove it. But I do not think that this is a bit deal.

changelogs/unreleased/config-anonymous-replica.md
Totktonada: I would like to make the resulting changelog a kind of 'anonymous replicas are now supported' (because it is likely what a user of 3.0.0-beta1 actually need to know). I can just add it in the last commit -- as I see we usually proceed this way -- but my idea was that a bit more verbose variant shouldn't hurt.

Totktonada: This list is also present in the documentation request (last commit) and in the pull request description (formulated as constrainsts and nuances).

src/box/lua/config/configdata.lua
```diff
@@ -445,6 +445,37 @@ local function validate_failover(found, peers, failover, leader)
     end
 end
 
+-- Verify that the given replicaset contains at least one
```
Serpentian: Nit: commit message typo: "must be in read-only mode"

Serpentian (approved):
Thank you for the patch! Looks great. Feel free to skip nits

src/box/lua/config/configdata.lua
Totktonada: Thanks! Fixed.

ylobankov (approved):

PR #9427, opened 2023-11-28
box: fix the comparison in the functions cache

### LINKED ISSUE #9426: Can't create a function in very specific conditions.
**Bug description**

If some conditions are met, the dropped function is not removed from the `func_by_name` table. The reason is that the `func_cache_delete` compares the function bucket ID in the `func_by_name` table with the `mh_end` of the `funcs` table.

The issue is detected in the scope of the #9371.

**Steps to reproduce**

```lua
box.cfg{}

-- funcs: bucket #0, n_dirty = 3.
-- funcs_by_name: bucket #0, n_dirty = 3.
box.schema.func.create('tortoise')

-- funcs: bucket #1, n_dirty = 4
-- funcs_by_name: collision with bucket #0, goes to bucket #2, n_dirty = 4.
box.schema.func.create('dog')

-- funcs: bucket #0 is emptied, n_dirty = 3.
-- funcs_by_name: bucket #0 is emptied, but n_dirty is not decremented, because
-- the dog had collision with it (so it's dirty), so n_dirty = 4.
--
-- Important: now funcs->n_dirty is one less than funcs_by_name->n_dirty.
box.schema.func.drop('tortoise')

-- Next simply fill more buckets, but do not collide with the bucket #0 in the
-- funcs_by_name. So in the n_dirty modification will be the same for both maps.
box.schema.func.create('mouse')
box.schema.func.create('chicken')
box.schema.func.create('horse')

-- Here we have an important moment. Before the next line funcs->n_dirty = 6,
-- and funcs_by_name->n_dirty = 7. This causes the funcs_by_name table to start
-- the resizing process once the "rat" function is registered: from now some of
-- table entries are moved over the buckets.
box.schema.func.create('rat')

-- Here we have the following situation:
--
--   funcs_by_name->n_buckets = 23 
--   funcs->n_buckets = 11
--   funcs_by_name["horse"] = 11
--
-- So now the funcs_by_name["horse"] is equal to the mh_end(funcs). This makes
-- the func_cache_delete think the horse does not exist in the funcs_by_name
-- table because its bucket ID is compared with the wrong mh_end(), so it does
-- not delete it from the table.
box.schema.func.drop('horse')

-- So this fires an assertion: the horse exists in the funcs_by_name already.
box.schema.func.create('horse')
```

**Actual behavior**

```
tarantool: ./src/box/func_cache.c:48: func_cache_insert: Assertion `func_by_name(func->def->name, strlen(func->def->name)) == NULL' failed.
Aborted (core dumped)
```

**Expected behavior**

No assertion is raised.

mkostoevr:
Before this patch it could happen that deletion of a function from the function cache didn't delete it from the funcs_by_name map. The reason is that the check if the function exists in the map performs the comparison of the search result with the `end` backet id of the wrong hash table.

The situation in which this could happen is the following:
1. Insertion of a new function into the cache triggers resize of thefuncs_by_value map, but the size of the funcs map remains the same.
2. Then user deletes a function. This removes the function from the funcs map. Then we check if the function exists in thefuncs_by_value map. The function exists there, but it so happens that it's bucket id equals to the funcs map bucket count, so the incorrect check if the function exists in the funcs_by_value map states that the function does not exist there, so it's not dropped from the map.
3. Now we have the following result: the function is referenced inthe funcs_by_value map, but not in funcs map. This triggers the assertion failure on any attempt to insert a new function with the same name.

Closes #9426

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-9426-func-cache.md
+++ b/changelogs/unreleased/gh-9426-func-cache.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed a very rare bug when a function deletion could cause an inconsistent
+  state of the database (gh-9426).
--- a/src/box/func_cache.c
+++ b/src/box/func_cache.c
@@ -66,7 +66,7 @@ func_cache_delete(uint32_t fid)
 	mh_i32ptr_del(funcs, k, NULL);
 	k = mh_strnptr_find_str(funcs_by_name, func->def->name,
 				strlen(func->def->name));
-	if (k != mh_end(funcs))
+	if (k != mh_end(funcs_by_name))
 		mh_strnptr_del(funcs_by_name, k, NULL);
 }
 
--- a/test/box-luatest/gh_9426_func_cache_test.lua
+++ b/test/box-luatest/gh_9426_func_cache_test.lua
@@ -1,0 +1,29 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new()
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.test_func_cache = function(cg)
+    cg.server:exec(function()
+        -- See the explanation in the #9426.
+        box.schema.func.create('tortoise')
+        box.schema.func.create('dog')
+        box.schema.func.drop('tortoise')
+        box.schema.func.create('mouse')
+        box.schema.func.create('chicken')
+        box.schema.func.create('horse')
+        box.schema.func.create('rat')
+        box.schema.func.drop('horse')
+        box.schema.func.create('horse')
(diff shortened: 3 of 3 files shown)


changelogs/unreleased/gh-9426-func-cache.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* Fixed a very rare bug when a function drop entered an inconsistent state of
+  the database (gh-9426).
```
p7nov, line 4:
```suggestion
* Fixed a very rare bug when a function deletion could cause
  an inconsistent state of the database (gh-9426).
```

p7nov (approved):

ochaplashkin (approved):

locker (approved):

locker:
Cherry-picked to 2.11.

PR #9437, opened 2023-12-01
Config names tests stabilization

Serpentian:
Closes tarantool/tarantool-qa#328
Closes tarantool/tarantool-qa#329


src/box/lua/config/applier/box_cfg.lua
```diff
@@ -158,20 +158,21 @@ local function missing_names_is_empty(missing_names, replicaset_name)
            table.equals(missing_names._peers, {})
 end
 
-local function names_on_rw_transit()
+local function names_try_set()
```
Totktonada, line 161: > NO_CHANGELOG=bugfix

The code with the bug was released as 3.0.0-beta1, so it worth to write a changelog entry.

Serpentian: Done

p7nov (approved):

PR #9439, opened 2023-12-04
uri: add login and password fields

### LINKED ISSUE #9435: Add `login` and `password` fields for the URI table representation
Currently, the `login` and `password` in URI can be set using `'login:password@host:port?params'`, `{'login:password@host : port', params = {...}}` or `{uri = 'login:password@host:port', 'params = {...}}`. However, since some options in our `config` module contain `login` and `password` as separate fields, it would be great to support `login` and `password` fields in the URI table representation. It is suggested to add two more variations: `{'host : port', login = '...', password = '...', params = {...}}` and `{uri = 'host : port', login = '...', password = '...', 'params = {...}}`.

ImeevMA:
Closes #9435

@TarantoolBot document
Title: `login` and `password` fields in URI table

Now `uri.parse()` can parse a table containing the `login` and `password` fields. Values from these fields take precedence over values obtained from the string URI. For example, login and password of `{uri = 'one:two:localhost:3301, login = 'alpha', password = 'omega'}` will be `alpha` and `omega` respectively. If the `login` field is set and the `password` field is not set, the password is set to `nil`. If the "password" field is set, the "login" field must be present.

### DIFF
--- a/changelogs/unreleased/gh-9435-login-password-fields-in-uri.md
+++ b/changelogs/unreleased/gh-9435-login-password-fields-in-uri.md
@@ -1,0 +1,3 @@
+## feature/uri
+
+* Introduced the `login` and `password` fields in `uri.parse()` (gh-9435).
--- a/src/lib/uri/uri.c
+++ b/src/lib/uri/uri.c
@@ -198,6 +198,17 @@ uri_create_params(struct uri *uri, const char *query)
 }
 
 void
+uri_set_credentials(struct uri *uri, const char *login, const char *password)
+{
+	assert(login != NULL);
+	free(uri->login);
+	uri->login = xstrdup(login);
+
+	free(uri->password);
+	uri->password = password == NULL ? NULL : xstrdup(password);
+}
+
+void
 uri_copy(struct uri *dst, const struct uri *src)
 {
 	dst->scheme = XSTRDUP(src->scheme);
--- a/src/lib/uri/uri.h
+++ b/src/lib/uri/uri.h
@@ -64,6 +64,10 @@ uri_add_param(struct uri *uri, const char *name, const char *value);
 void
 uri_remove_param(struct uri *uri, const char *name);
 
+/** Set @a login and @a password in the @a uri. */
+void
+uri_set_credentials(struct uri *uri, const char *login, const char *password);
+
 /**
  * Copy constructor for @a dst URI. Copies all fiels from
  * @a src URI to @dst URI.
--- a/src/lua/uri.c
+++ b/src/lua/uri.c
@@ -161,12 +161,42 @@ uri_create_from_lua_table(struct lua_State *L, int idx, struct uri *uri)
 	lua_pop(L, 1);
 	if (rc != 0)
 		return rc;
+
+	lua_getfield(L, idx, "login");
+	if (lua_type(L, -1) == LUA_TSTRING) {
+		lua_getfield(L, idx, "password");
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			uri_set_credentials(uri, lua_tostring(L, -2),
+					    lua_tostring(L, -1));
+		} else if (lua_type(L, -1) == LUA_TNIL) {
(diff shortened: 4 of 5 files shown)


src/lib/uri/uri.c
```diff
@@ -197,6 +197,28 @@ uri_create_params(struct uri *uri, const char *query)
 	free(copy);
 }
 
+void
+uri_add_login(struct uri *uri, const char *login)
+{
+	assert(login != NULL);
+	size_t size = strlen(login) + 1;
+	uri->login = xrealloc(uri->login, size);
+	memcpy(uri->login, login, size);
```
locker, line 206: Nit: Please use `xstrdup`.

src/lib/uri/uri.h
```diff
@@ -64,6 +64,14 @@ uri_add_param(struct uri *uri, const char *name, const char *value);
 void
 uri_remove_param(struct uri *uri, const char *name);
 
+/** Set @a login in the @a uri. */
+void
+uri_add_login(struct uri *uri, const char *login);
+
+/** Set @a password in the @a uri. */
+void
+uri_add_password(struct uri *uri, const char *password);
```
locker, line 73: Probably, better call them `uri_set_XXX`, not `uri_add_XXX` because you can't set more than one login/password.

Also, since setting password without a login is pointless, it may be better to introduce one function to set them both:

```C
void
uri_set_credentials(struct uri *uri, const char *login, const char *password);
```

src/lua/uri.c
```diff
@@ -161,12 +161,53 @@ uri_create_from_lua_table(struct lua_State *L, int idx, struct uri *uri)
 	lua_pop(L, 1);
 	if (rc != 0)
 		return rc;
+
+	bool has_login = false;
+	lua_pushstring(L, "login");
+	lua_rawget(L, idx);
```
locker, line 167: Nit: You can use `lua_getfield` helper here.

src/lua/uri.c
```diff
@@ -161,12 +161,53 @@ uri_create_from_lua_table(struct lua_State *L, int idx, struct uri *uri)
...
+	if (lua_type(L, -1) != LUA_TNIL) {
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			has_login = true;
+			uri_add_login(uri, lua_tostring(L, -1));
+		} else {
+			diag_set(IllegalParams, "Invalid URI table: expected "
+				 "type for login is string");
+			rc = -1;
+		}
+	}
```
locker, line 177: Nit: You don't need an extra nesting level here:

```C
if (lua_type(L, -1) == LUA_TSTRING) {
        // ok
} else if (lua_type(L, -1) != LUA_TNIL) {
        // error
}
```

locker (changes requested):

src/lib/uri/uri.c
ImeevMA: Fixed.

src/lua/uri.c
ImeevMA: Thanks, used.

locker (approved):

ochaplashkin (approved):

changelogs/unreleased/gh-9435-login-password-fields-in-uri.md
```diff
@@ -0,0 +1,3 @@
+## feature/uri
+
+* Introduced fields `login` and `password` in `uri.parse()` (gh-9435).
```
andreyaksenov, line 3:
```suggestion
* Introduced the `login` and `password` fields in `uri.parse()` (gh-9435).
```

andreyaksenov (approved):

changelogs/unreleased/gh-9435-login-password-fields-in-uri.md
ImeevMA: Thanks, fixed.

ylobankov (approved):

PR #9440, opened 2023-12-04
config: support master auto-discovery in vshard

ImeevMA:
This patch adds support for automatic master discovery in vshard. There is no longer a need to reapply the vshard storage configuration every time an instance becomes a master or ceases to be a master.

Automatic master discovery also solves the problem with the rebalancer. Previously, the rebalancer could not work correctly if the masters of some replicasets were unknown, and since the vshard config generated by the config module did not contain information about all the masters, the rebalancer was disabled. The rebalancer can now perform master discovery
on its own, which is why it is enabled.

Before this patch, the config module was based on the "test: use dofile() for configs instead of require" commit in vshard. At a minimum, commit "storage: fix assertion error in conn_manager" is now required.

Part of #8862

### DIFF
--- a/src/box/lua/config/applier/sharding.lua
+++ b/src/box/lua/config/applier/sharding.lua
@@ -1,10 +1,7 @@
 local log = require('internal.config.utils.log')
-local fiber = require('fiber')
 _G.vshard = nil
 
-local fiber_wait_ro_rw
-
-local function vshard_cfg(config)
+local function apply(config)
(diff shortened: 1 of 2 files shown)


Totktonada (approved):
Thank you!

It looks much more clean now!

Totktonada:
NB: I've added the information from the commit message into the pull request description for convenience.

PR #9444, opened 2023-12-04
relay: check return value of xrow_decode_synchro

### LINKED ISSUE #9265: Unchecked return value of `xrow_decode_synchro` in `relay_filter_row`.
**Bug description**
[Coverity report link](https://scan7.scan.coverity.com/reports.htm#v39692/p13437/fileInstanceId=151003889&defectInstanceId=19060508&mergedDefectId=1569145)

sergepetrenko:
Relay sometimes decodes the PROMOTE packets to be sent in order to conditionally delay their despatch. It was believed that own WALs can't be corrupted and hence there is no point in checking that decoding succeeds.

Let's be more strict here and check the decoding result before proceeding.

Closes #9265

NO_DOC=bugfix
NO_TEST=hard to test
NO_CHANGELOG=not user-visible

### DIFF
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -1280,7 +1280,8 @@ relay_filter_row(struct relay *relay, struct xrow_header *packet)
 	 */
 	if (iproto_type_is_promote_request(packet->type)) {
 		struct synchro_request req;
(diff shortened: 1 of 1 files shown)


Gerold103 (approved):
Thanks for the patch!

Issue #8423, opened 2023-03-09
Get rid of `pthread_cancel()`


Gumix:
Motivation:
1. In general, thread cancellation is not a good idea, it is better to do a graceful shutdown.
2. `pthread_cancel()` doesn't work well in macOS:
https://github.com/tarantool/tarantool/issues/8420#issuecomment-1458803803
See also https://github.com/tarantool/tarantool-qa/issues/32#issuecomment-735321024

Currently `pthread_cancel()` is used by:
```
src/box/applier.cc:2106:        cord_cancel_and_join(&thread->cord);
src/box/iproto.cc:3562:         cord_cancel_and_join(&iproto_threads[i].net_cord);
src/box/memtx_engine.cc:836:    cord_cancel_and_join(&ckpt->cord);
src/box/memtx_engine.cc:848:    cord_cancel_and_join(replica_join_cord);
src/box/relay.cc:356:           cord_cancel_and_join(&relay->cord);
src/box/vy_run.c:169:           cord_cancel_and_join(&reader->cord);
src/box/vy_scheduler.c:359:     cord_cancel_and_join(&worker->cord);
```

PR #9446, opened 2023-12-05
fiber: make cord_cojoin cancellable

nshy:
We may need to cancel fiber that waits for cord to finish. For this purpose let's cancel fiber started by `cord_costart` inside the cord.

Note that there is a race between stopping `cancel_event` in cord and triggering it using `ev_async_send` in joining thread. AFAIU it is safe.

We also need to fix stopping wal cord to address stack-use-after-return issue shown below. Is arises because we did not stop async which resides in wal endpoint and endpoint resides on stack. Later when we stop the introduced `cancel_event` we access not stopped async which at this moment gone out of scope.

```
==3224698==ERROR: AddressSanitizer: stack-use-after-return on address 0x7f654b3b0170 at pc 0x555a2817c282 bp 0x7f654ca55b30 sp 0x7f654ca55b28
WRITE of size 4 at 0x7f654b3b0170 thread T3
    #0 0x555a2817c281 in ev_async_stop /home/shiny/dev/tarantool/third_party/libev/ev.c:5492:37
    #1 0x555a27827738 in cord_thread_func /home/shiny/dev/tarantool/src/lib/core/fiber.c:1990:2
    #2 0x7f65574aa9ea in start_thread /usr/src/debug/glibc/glibc/nptl/pthread_create.c:444:8
    #3 0x7f655752e7cb in clone3 /usr/src/debug/glibc/glibc/misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

Part of #8423

### DIFF
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -589,6 +589,7 @@ wal_free(void)
 	struct wal_writer *writer = &wal_writer_singleton;
 
 	cbus_stop_loop(&writer->wal_pipe);
+	cpipe_destroy(&writer->wal_pipe);
 
 	if (cord_join(&writer->cord)) {
 		/* We can't recover from this in any reasonable way. */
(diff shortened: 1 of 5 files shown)


nshy:
Unfortunately due to changes to stopping wal thread the patch depends on stopping relay threads to pass the tests.

Issue #6597, opened 2021-11-09
Update luarocks to 3.9.2


Totktonada:
Needed to autoreplace `git://` with `git+https://`, see https://github.com/tarantool/tarantool/issues/6587#issuecomment-963988433.

Totktonada:
See also #6223.

DifferentialOrange:
See also https://github.com/tarantool/cartridge/issues/1834

PR #9448, opened 2023-12-05
Proposed changes to changelog for 2.11.2 release.

sergos:
Updates to make minor release feature-less. 

NO_DOC=no features
NO_TEST=no code added
NO_CHANGELOG=it is the changelog

### DIFF
--- a/changelogs/unreleased/add-asan-flag-to-tarantool-build.md
+++ b/changelogs/unreleased/add-asan-flag-to-tarantool-build.md
@@ -1,4 +1,5 @@
-## feature/box
+## testing
 
 * Added a new flag `tarantool.build.asan` that shows whether build
-  flag `ENABLE_ASAN` is set.
+  flag `ENABLE_ASAN` is set. The flag is intended to ease tests
+  backporting from the master branch.
--- a/changelogs/unreleased/bump-libcurl-to-8.3.0.md
+++ b/changelogs/unreleased/bump-libcurl-to-8.3.0.md
@@ -1,3 +1,0 @@
-## feature/build
-
-* Updated libcurl to version 8.3.0.
--- a/changelogs/unreleased/gh-6597-update-luarocks.md
+++ b/changelogs/unreleased/gh-6597-update-luarocks.md
@@ -1,3 +1,3 @@
-## feature/misc
+## bugfix/misc
 
-* Updated luarocks version to 3.9.2 (gh-6597).
+* Updated luarocks version to 3.9.2 to address rockspec security issues (gh-6587).
--- a/changelogs/unreleased/gh-7593-console-debugger.md
+++ b/changelogs/unreleased/gh-7593-console-debugger.md
@@ -1,3 +1,3 @@
-## feature/debugger
+## tools/debugger
 
 * Introduced readline support to tarantool debugger (gh-7738).
--- a/changelogs/unreleased/ghe-523-max-config-param-length.md
+++ b/changelogs/unreleased/ghe-523-max-config-param-length.md
@@ -1,3 +1,3 @@
-## feature/box
+## bugfix/box
 
-* The maximum length of `box.cfg{}` string parameters is now 512 instead of 256.
+* The maximum length of `box.cfg{}` string parameters is now 512 to support a lengthy args, such as list of audit events (ghe-523).
--- a/changelogs/unreleased/gh-9086-fix-streaming-connection-stuck.md
+++ b/changelogs/unreleased/gh-9086-fix-streaming-connection-stuck.md
@@ -1,4 +1,4 @@
-## bugfix/http
+## bugfix/lua/http client
 
 * Fixed a streaming connection stuck if etcd is stopped
   unexpectedly (gh-9086).


changelogs/unreleased/bump-libcurl-to-8.3.0.md
```diff
@@ -1,3 +1,3 @@
-## feature/build
+## bugfix/build
 
-* Updated libcurl to version 8.3.0.
```
Totktonada, line 3: Let's look into the words.

The previous changelog entry states the following: we updated a build time dependency. Feature? Not exactly, rather a change in an existing functionality.

The new entry states that we fixed a bug (`## bugfix`) in the build machinery (`/build`). So, there was a build failure? Or incorrect executable file? Seems, no. There is no bug in the build machinery.

Next, what is the bug? libcurl-7.87.0 is not a bug, libcurl 8.3.0 is not a fix. If we want to tell about a bugfix, we should include some problem statement. In fact, the issue states a problem in the built-in http client and I think that we should tell about it here.

How exactly the problem is fixed (by a change in the build machinery) is on the deeper level.

changelogs/unreleased/gh-6597-update-luarocks.md
```diff
@@ -1,3 +1,3 @@
-## feature/misc
+## bugfix/misc
 
 * Updated luarocks version to 3.9.2 (gh-6597).
```
Totktonada, line 3: Same here: it was backported to 2.11 in order to fix some problem and we should have a problem statement in the changelog entry for a bugfix.

changelogs/unreleased/gh-7593-console-debugger.md
```diff
@@ -1,3 +1,3 @@
-## feature/debugger
+## bugfix/debugger
 
-* Introduced readline support to tarantool debugger (gh-7738).
```
Totktonada, line 3: It doesn't look as a bugfix. Just feature that was added to 2.11 for some reason. Don't know why.

I remember some discussion that we want to allow tooling features in bugfix releases, but I don't remember any result of the discussion or proposals to update the release policy.

I think that the patch shouldn't be there according to our policy. @igormunkin, please, reveal details if you have them.

changelogs/unreleased/ghe-523-max-config-param-length.md
```diff
@@ -1,3 +1,3 @@
-## feature/box
+## bugfix/box
 
-* The maximum length of `box.cfg{}` string parameters is now 512 instead of 256.
```
Totktonada, line 3: It looks almost OK. I can understand that there was a problem with some valid configuration due to the old limit. However, it would be better to say it explicitly.

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
```diff
@@ -1,4 +0,0 @@
-## feature/box
-
-* Added a new flag `tarantool.build.asan` that shows whether build
-  flag `ENABLE_ASAN` is set.
```
Totktonada, line 4: I agree that it is a small feature. However, it is a feature: here we add something into the public API.

AFAIR, we need it for testing purposes. The obvious workaround would be to move the flag to some internal API, however it would complicate backporting of new test cases.

Can we drop the flag from the executable, but add a workaround into test/helper.lua? This file is loaded at luatest startup and can be used to patch the same `tarantool.build` table.

OTOH, I understand that the feature is small and safe, so we can also consider some alternatives toward the policy change. I would note that we need some separate release notes category for such changes. 'Functionality added or changed' shouldn't appear in a bugfix release.

Let's discuss?

sergos: I would question it is a feature. Reading the documentation:

> one can answer some questions about how the tarantool server was built, such as “what flags were used”, or “what was the version of the compiler”

There's no promises regarding fields in has or has not, neither any reasonable "public" inference from the facts it contains. E.g. what should you - as a client - take from the cmake options used while tarantool was built? Will you make a choice in favor of '-Wpedantic'? It says "some questions" so it has no limitation in my opinion. 

Totally different story are `pid()` and `uptime()`, which are not only pinned by the doc, but advised for use:

> Additionally one can see the uptime and the server version and the process id. Those information items can also be accessed with [box.info()](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/#box-introspection-box-info) but use of the tarantool module is recommended.

So for this particular case I believe it doesn't represent a feature that we have to find a way to introduce in a patch revision. Not saying it can't be a real problem per se: sometimes we need some feature to be backported. 
I propose to introduce a **'waiver'** (in terms of our policy) status for such cases and discuss it every time it appears. It will designate a feature/change in functionality we intentionally introduce with understanding we violate the policy (waiving it).

changelogs/unreleased/gh-7593-console-debugger.md
sergos: Well, the debugger CLI has an interference with Tarantool's one and this can be considered as a bug. Making proper support of readline in debugger (or vice versa) means a fix. But I would like to hear @igormunkin too.

changelogs/unreleased/gh-6597-update-luarocks.md
sergos: fixed

changelogs/unreleased/ghe-523-max-config-param-length.md
sergos: fixed

changelogs/unreleased/gh-7593-console-debugger.md
igormunkin: "Ah shit, here we go again." (c)

> It doesn't look as a bugfix. Just feature that was added to 2.11 for some reason. Don't know why.

We discussed this with @kyukhin offline. We both were OK with it.

> I remember some discussion that we want to allow tooling features in bugfix releases, but I don't remember any result of the discussion or proposals to update the release policy.

I also remember us talking regarding the features added for our toolchain (profilers, debugger, gdb extensions, etc). Unfortunately, I don't remember the final decision, but only the fact nobody was against such relaxation of our release policy.

> I think that the patch shouldn't be there according to our policy.

OK, then, feel free to revert it. In fact it means the 2.11 release is totally freezed forever and can be barely used if we talk about our raw tooling.

> Well, the debugger CLI has an interference with Tarantool's one and this can be considered as a bug. Making proper support of readline in debugger (or vice versa) means a fix.

AFAIR, it was covered with the corresponding stub, so it can hardly be considered as a bug.

---

All in all, I have no strong opinion regarding this particular patch. But I still believe that tooling features has to be maintained separately from the *core-related* ones. Otherwise, I see little sense in toolchain enhancements and development. If anybody think the patch shouldn't be landed to this release, feel free to revert it.

sergos: Ok, I see another tag in changelog - **'tooling'** - that means we update/upgrade/change/revert the programmer's tooling in our platform on our own discretion. @Totktonada do you consider this as a way to go?

changelogs/unreleased/bump-libcurl-to-8.3.0.md
sergos: fixed

Totktonada: I would write `## bugfix/lua/http client`.

changelogs/unreleased/gh-7593-console-debugger.md
Totktonada: We (me, @igormunkin, @sergos) agreed to relax the release policy regarding metrics and developers tools later. It is on @igormunkin.

We agreed to add a changelog category next to `feature` and `bugfix`: `tools`.

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
Totktonada: Let's move it to its own category `testing`. We need to add some words that this is to ease backporting of tests from master.

changelogs/unreleased/bump-libcurl-to-8.3.0.md
sergos: It appears we have another entry in our list:

git show 83ddadb46ef110948960209cb9197b85372109fb
```
diff --git a/changelogs/unreleased/bump-libcurl-to-8.3.0.md b/changelogs/unreleased/bump-libcurl-to-8.3.0.md
new file mode 100644
index 0000000000..ee7d9e3b45
--- /dev/null
+++ b/changelogs/unreleased/bump-libcurl-to-8.3.0.md
@@ -0,0 +1,3 @@
+## feature/build
+
+* Updated libcurl to version 8.3.0.
diff --git a/changelogs/unreleased/gh-9086-fix-streaming-connection-stuck.md b/changelogs/unreleased/gh-9086-fix-streaming-connection-stuck.md
new file mode 100644
index 0000000000..34a10bb183
--- /dev/null
+++ b/changelogs/unreleased/gh-9086-fix-streaming-connection-stuck.md
@@ -0,0 +1,4 @@
+## bugfix/http
+
+* Fixed a streaming connection stuck if etcd is stopped
+  unexpectedly (gh-9086).
```

I believe we have to merge them in one, since they're about one fix - perhaps change the `bugfix/http` for `bugfix/lua/http client` according to your request.

changelogs/unreleased/gh-7593-console-debugger.md
sergos: fixed

tools/gen-release-notes
```diff
@@ -14,7 +14,7 @@ import subprocess
 # Ensure certain order for known sections.
 SECTIONS_TOP = ['core', 'memtx', 'vinyl', 'replication', 'swim',
                 'raft', 'luajit', 'lua', 'sql']
-SECTIONS_BOTTOM = ['build', 'testing', 'misc']
+SECTIONS_BOTTOM = ['build', 'testing', 'tools', 'misc']
```
Totktonada, line 17: Nit: It is for direct subsections of `feature` and `bugfix`. Since we're going to move `testing` and `tools` into top level sections, I would rather remove `testing` from this list.

NB: Shouldn't we push the script changes via `master`?

sergos:
I've split this in two - first part is about change wording for bug fixes, it is done - @Totktonada please have a look and we can merge.
The second one needs changes re new categories from master to be backported from https://github.com/tarantool/tarantool/commit/4d6937ef69a55b72ce1c8194cf9c42b913752a73

tools/gen-release-notes
sergos: https://github.com/tarantool/tarantool/commit/4d6937ef69a55b72ce1c8194cf9c42b913752a73

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
sergos: Sure, will do after categories backport.

Totktonada: I'll left the comment floating here open till the update.

Totktonada (approved):
OK for me (taking into account one remaining comment with an agreement what to do).

Thank you for the patience!

changelogs/unreleased/add-asan-flag-to-tarantool-build.md
sergos: done, have a look in https://github.com/tarantool/tarantool/pull/9448/commits/eb0c06411ab507f83a6dae6d9a5ec16026610d2f

Totktonada:
Thanks for the update! OK for me.

PR #9457, opened 2023-12-07
Iproto and relay graceful shutdown

nshy:
Iproto and relay part of #8423

**update**: We have to leave `cord_cancel_and_join` for iproto for now. So iproto part is not complete.

### DIFF
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -203,6 +203,15 @@ struct iproto_thread {
 	struct evio_service binary;
 	/** Requests count currently pending in stream queue. */
 	size_t requests_in_stream_queue;
+	/** List of all connections. */
+	struct rlist connections;
+	/** Number of connections that pending drop. */
+	size_t drop_pending_connection_count;
+	/**
+	 * Message used to notify TX thread that all connections marked
+	 * to de dropped are dropped.
+	 */
+	struct cmsg drop_finished_msg;
 	/**
 	 * The following fields are used exclusively by the tx thread.
 	 * Align them to prevent false-sharing.
@@ -215,6 +224,11 @@ struct iproto_thread {
 		struct rmean *rmean;
 	} tx;
 };
+
+/** Condition for drop finished. */
+static struct fiber_cond drop_finished_cond;
+/** Count of iproto threads that are not finished connections drop yet. */
+static size_t drop_pending_thread_count;
 
 /**
  * IPROTO listen URIs. Set by box.cfg.listen.
@@ -437,12 +451,10 @@ struct iproto_msg
 	struct stailq_entry in_stream;
 	/** Stream that owns this message, or NULL. */
 	struct iproto_stream *stream;
-	/**
-	 * True if message processing in tx thread is started. The flag is
-	 * used to prevent double accounting of message in case of iproto
-	 * override fallback to original handler.
-	 */
-	bool accepted;
+	/** Link in connection->tx.inprogress. */
+	struct rlist in_inprogress;
+	/** TX thread fiber that processing this message. */
+	struct fiber *fiber;
 };
 
 /**
@@ -722,6 +734,8 @@ struct iproto_connection
 		 * return.
 		 */
 		bool is_push_pending;
+		/** List of inprogress messages. */
+		struct rlist inprogress;
 	} tx;
 	/** Authentication salt. */
 	char salt[IPROTO_SALT_SIZE];
@@ -732,6 +746,15 @@ struct iproto_connection
 	 * IO is handled by relay code.
 	 */
 	bool is_in_replication;
+	/** Link in iproto_thread->connections. */
+	struct rlist in_connections;
+	/** Set if connection is being dropped. */
+	bool is_drop_pending;
+	/**
+	 * Messaged sent to TX to cancel all inprogress requests of the
+	 * connection.
+	 */
+	struct cmsg cancel_msg;
 };
 
 /** Returns a string suitable for logging. */
@@ -819,7 +842,7 @@ iproto_msg_new(struct iproto_connection *con)
 	msg->close_connection = false;
 	msg->connection = con;
 	msg->stream = NULL;
-	msg->accepted = false;
+	msg->fiber = NULL;
 	rmean_collect(con->iproto_thread->rmean, IPROTO_REQUESTS, 1);
 	return msg;
 }
@@ -1476,7 +1499,10 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	con->long_poll_count = 0;
 	con->session = NULL;
 	con->is_in_replication = false;
+	con->is_drop_pending = false;
 	rlist_create(&con->in_stop_list);
+	rlist_create(&con->tx.inprogress);
+	rlist_add_entry(&iproto_thread->connections, con, in_connections);
 	/* It may be very awkward to allocate at close. */
 	cmsg_init(&con->destroy_msg, con->iproto_thread->destroy_route);
 	cmsg_init(&con->disconnect_msg, con->iproto_thread->disconnect_route);
@@ -1485,6 +1511,27 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	con->tx.is_push_sent = false;
 	rmean_collect(iproto_thread->rmean, IPROTO_CONNECTIONS, 1);
 	return con;
+}
+
+/** Notify that connections drop is finished. */
+static void
+tx_process_drop_finished(struct cmsg *m)
+{
+	(void)m;
+	assert(drop_pending_thread_count > 0);
+	if (--drop_pending_thread_count == 0)
+		fiber_cond_signal(&drop_finished_cond);
+}
+
+/** Send message to TX thread to notify that connections drop is finished. */
+static void
+iproto_send_drop_finished(struct iproto_thread *iproto_thread)
+{
+	static const struct cmsg_hop drop_finished_route[1] =
+					{{ tx_process_drop_finished, NULL }};
+
+	cmsg_init(&iproto_thread->drop_finished_msg, drop_finished_route);
+	cpipe_push(&iproto_thread->tx_pipe, &iproto_thread->drop_finished_msg);
 }
 
 /** Recycle a connection. */
@@ -1506,6 +1553,14 @@ iproto_connection_delete(struct iproto_connection *con)
 
 	assert(mh_size(con->streams) == 0);
 	mh_i64ptr_delete(con->streams);
+	rlist_del(&con->in_connections);
+	if (con->is_drop_pending) {
+		struct iproto_thread *iproto_thread = con->iproto_thread;
+
+		assert(iproto_thread->drop_pending_connection_count > 0);
+		if (--iproto_thread->drop_pending_connection_count == 0)
+			iproto_send_drop_finished(iproto_thread);
+	}
 	mempool_free(&con->iproto_thread->iproto_connection_pool, con);
 }
 
@@ -1781,6 +1836,17 @@ net_finish_rollback_on_disconnect(struct cmsg *m)
 		iproto_connection_try_to_start_destroy(con);
 }
 
+/** Cancel all inprogress requests of the connection. */
+static void
+tx_process_cancel_inprogress(struct cmsg *m)
+{
+	struct iproto_connection *con =
+		container_of(m, struct iproto_connection, cancel_msg);
+	struct iproto_msg *msg;
+	rlist_foreach_entry(msg, &con->tx.inprogress, in_inprogress)
+		fiber_cancel(msg->fiber);
+}
+
 static void
 tx_process_disconnect(struct cmsg *m)
 {
@@ -1951,13 +2017,15 @@ static inline struct iproto_msg *
 tx_accept_msg(struct cmsg *m)
 {
 	struct iproto_msg *msg = (struct iproto_msg *) m;
-	if (msg->accepted)
+	if (msg->fiber != NULL)
 		return msg;
-	msg->accepted = true;
 	tx_accept_wpos(msg->connection, &msg->wpos);
 	tx_fiber_init(msg->connection->session, msg->header.sync);
 	tx_prepare_transaction_for_request(msg);
 	msg->connection->iproto_thread->tx.requests_in_progress++;
+	rlist_add_entry(&msg->connection->tx.inprogress, msg,
+			in_inprogress);
+	msg->fiber = fiber();
 	rmean_collect(msg->connection->iproto_thread->tx.rmean,
 		      REQUESTS_IN_PROGRESS, 1);
 	flightrec_write_request(msg->reqstart, msg->len);
@@ -2009,6 +2077,8 @@ tx_end_msg(struct iproto_msg *msg, struct obuf_svp *svp)
 		msg->stream->txn = txn_detach();
 	}
 	msg->connection->iproto_thread->tx.requests_in_progress--;
+	rlist_del(&msg->in_inprogress);
+	msg->fiber = NULL;
 	struct obuf *out = msg->connection->tx.p_obuf;
 	if (msg->connection->tx.p_obuf->used != svp->used)
 		/* Log response to the flight recorder. */
@@ -2621,6 +2691,8 @@ tx_process_replication(struct cmsg *m)
 		  * In case of a timeout the error could come after a partially
 		  * written row. Do not push it on top.
 		  */
+	} catch (FiberIsCancelled *e) {
+		/* Do not write into connection on connection drop. */
 	} catch (Exception *e) {
 		iproto_write_error(io, e, ::schema_version, msg->header.sync);
 	}
@@ -2788,6 +2860,11 @@ net_end_join(struct cmsg *m)
 
 	assert(! ev_is_active(&con->input));
 	con->is_in_replication = false;
+
+	if (con->is_drop_pending) {
+		iproto_connection_close(con);
+		return;
+	}
 	/*
 	 * Enqueue any messages if they are in the readahead
 	 * queue. Will simply start input otherwise.
@@ -3205,6 +3282,7 @@ iproto_thread_init(struct iproto_thread *iproto_thread)
 	rlist_create(&iproto_thread->stopped_connections);
 	iproto_thread->tx.requests_in_progress = 0;
 	iproto_thread->requests_in_stream_queue = 0;
+	rlist_create(&iproto_thread->connections);
 }
 
 /** Initialize the iproto subsystem and start network io thread */
@@ -3226,6 +3304,7 @@ iproto_init(int threads_count)
 	evio_service_create(loop(), &tx_binary, "tx_binary", NULL, NULL);
 	iproto_threads = (struct iproto_thread *)
 		xcalloc(threads_count, sizeof(struct iproto_thread));
+	fiber_cond_create(&drop_finished_cond);
 
 	for (int i = 0; i < threads_count; i++, iproto_threads_count++) {
 		struct iproto_thread *iproto_thread = &iproto_threads[i];
@@ -3283,6 +3362,10 @@ enum iproto_cfg_op {
 	 * Command code to create a new IPROTO session.
 	 */
 	IPROTO_CFG_SESSION_NEW,
+	/**
+	 * Command code to drop all current connections.
+	 */
+	IPROTO_CFG_DROP_CONNECTIONS,
 };
 
 /**
@@ -3397,6 +3480,33 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
 		iproto_thread_accept(iproto_thread, io, addr, addrlen, session);
 		break;
 	}
+	case IPROTO_CFG_DROP_CONNECTIONS: {
+		struct iproto_connection *con;
+		static const struct cmsg_hop cancel_route[1] =
+				{{ tx_process_cancel_inprogress, NULL }};
+		rlist_foreach_entry(con, &iproto_thread->connections,
+				    in_connections) {
+			/*
+			 * Replication IO is done outside iproto so we
+			 * cannot close them as usual. Anyway we cancel
+			 * replication fibers as well and close connection
+			 * after replication is breaked.
+			 */
+			if (!con->is_in_replication &&
+			    con->state == IPROTO_CONNECTION_ALIVE)
+				iproto_connection_close(con);
+			con->is_drop_pending = true;
+			iproto_thread->drop_pending_connection_count++;
+			if (con->state != IPROTO_CONNECTION_DESTROYED) {
+				cmsg_init(&con->cancel_msg, cancel_route);
+				cpipe_push(&iproto_thread->tx_pipe,
+					   &con->cancel_msg);
+			}
+		}
+		if (iproto_thread->drop_pending_connection_count == 0)
+			iproto_send_drop_finished(iproto_thread);
+		break;
+	}
 	default:
 		unreachable();
 	}
@@ -3458,6 +3568,24 @@ iproto_send_start_msg(void)
 	iproto_cfg_msg_create(&cfg_msg, IPROTO_CFG_START);
 	for (int i = 0; i < iproto_threads_count; i++)
 		iproto_do_cfg(&iproto_threads[i], &cfg_msg);
+}
+
+void
+iproto_drop_connections(void)
+{
+	static struct latch latch = LATCH_INITIALIZER(latch);
+	latch_lock(&latch);
+	drop_pending_thread_count = iproto_threads_count;
+	for (int i = 0; i < iproto_threads_count; i++) {
+		struct iproto_cfg_msg *cfg_msg =
+			(struct iproto_cfg_msg *)xmalloc(sizeof(*cfg_msg));
+		iproto_cfg_msg_create(cfg_msg, IPROTO_CFG_DROP_CONNECTIONS);
+		iproto_do_cfg_async(&iproto_threads[i], cfg_msg);
+	}
+
+	while (drop_pending_thread_count != 0)
+		fiber_cond_wait(&drop_finished_cond);
+	latch_unlock(&latch);
 }
 
 /** Send IPROTO_CFG_RESTART to all threads. */
@@ -3683,6 +3811,7 @@ iproto_free(void)
 		iproto_req_handler_delete(handler);
 	}
 	mh_i32ptr_delete(tx_req_handlers);
+	fiber_cond_destroy(&drop_finished_cond);
 
 	/*
 	 * Here we close sockets and unlink all unix socket paths.
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -454,13 +454,10 @@ iproto_resume(struct iproto_thread *iproto_thread);
 /**
  * Prepares IPROTO message: decodes the message header, checks the message's
  * stream identifier, and set's the message's cbus route.
- * If the message contains a replication request, sets the stop input flag,
- * which means the connection's buffers needs to be detached from the event
- * loop.
  */
 static void
-iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend,
-		   bool *stop_input);
+iproto_msg_prepare(struct iproto_msg *msg, const char **pos,
+		   const char *reqend);
 
 enum rmean_net_name {
 	IPROTO_SENT,
@@ -730,6 +727,11 @@ struct iproto_connection
 	char salt[IPROTO_SALT_SIZE];
 	/** Iproto connection thread */
 	struct iproto_thread *iproto_thread;
+	/**
+	 * The connection is processing replication command so that
+	 * IO is handled by relay code.
+	 */
+	bool is_in_replication;
 };
 
 /** Returns a string suitable for logging. */
@@ -1149,9 +1151,8 @@ iproto_enqueue_batch(struct iproto_connection *con, struct ibuf *in)
 {
 	assert(rlist_empty(&con->in_stop_list));
 	int n_requests = 0;
-	bool stop_input = false;
 	const char *errmsg;
-	while (con->parse_size != 0 && !stop_input) {
+	while (con->parse_size != 0 && !con->is_in_replication) {
 		if (iproto_check_msg_max(con->iproto_thread)) {
 			iproto_connection_stop_msg_max_limit(con);
 			cpipe_flush_input(&con->iproto_thread->tx_pipe);
@@ -1187,7 +1188,7 @@ err_msgpack:
 		msg->len = reqend - reqstart; /* total request length */
 		con->input_msg_count[msg->p_ibuf == &con->ibuf[1]]++;
 
-		iproto_msg_prepare(msg, &pos, reqend, &stop_input);
+		iproto_msg_prepare(msg, &pos, reqend);
 		if (iproto_msg_start_processing_in_stream(msg)) {
 			cpipe_push_input(&con->iproto_thread->tx_pipe, &msg->base);
 			n_requests++;
@@ -1198,7 +1199,7 @@ err_msgpack:
 		assert(con->parse_size >= (size_t) (reqend - reqstart));
 		con->parse_size -= reqend - reqstart;
 	}
-	if (stop_input) {
+	if (con->is_in_replication) {
 		/**
 		 * Don't mess with the file descriptor
 		 * while join is running. ev_io_stop()
@@ -1474,6 +1475,7 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	con->can_write = true;
 	con->long_poll_count = 0;
 	con->session = NULL;
+	con->is_in_replication = false;
 	rlist_create(&con->in_stop_list);
 	/* It may be very awkward to allocate at close. */
 	cmsg_init(&con->destroy_msg, con->iproto_thread->destroy_route);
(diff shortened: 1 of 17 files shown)


ochaplashkin (approved):

src/box/iproto.cc
```diff
@@ -3461,6 +3567,43 @@ iproto_send_start_msg(void)
 		iproto_do_cfg(&iproto_threads[i], &cfg_msg);
 }
 
+void
+iproto_shutdown(void)
```
locker, line 3571: We need a separate function to drop all existing connections for https://github.com/tarantool/tarantool-ee/issues/585. Let's implement a function that would do just that, without deinitializing iproto. Since we stop accepting new connections in `iproto_on_shutdown_f`, this should be enough to shutdown iproto gracefully - we'd just need to stop and join all iproto threads in `iproto_free`.

src/box/iproto.cc
```diff
@@ -2789,6 +2857,12 @@ net_end_join(struct cmsg *m)
 	iproto_msg_delete(msg);
 
 	assert(! ev_is_active(&con->input));
+	con->is_in_replication = false;
+
+	if (iproto_shutdown_started) {
```
locker, line 2862: This global variable is updated from the tx thread so accessing it from the iproto thread without synchronization looks unsafe. Also, I think we should maintain this flag per connection so that we can drop existing connections without shutting down iproto for the sake of https://github.com/tarantool/tarantool-ee/issues/585.

src/box/iproto.cc
```diff
@@ -2639,28 +2701,35 @@ static void
...
+	struct iproto_req_handler *handler;
+	const char *header;
+	const char *header_end;
+	const char *body = "\x80"; /* Empty MsgPack map encoding. */
+	const char *body_end = body + 1;
+
+	if (iproto_shutdown_started) {
+		diag_set(FiberIsCancelled);
+		goto error;
+	}
```
locker, line 2715: I don't understand why you need to handle this case explicitly.

src/box/iproto.cc
```diff
@@ -2624,7 +2684,9 @@ tx_process_replication(struct cmsg *m)
 		  * written row. Do not push it on top.
 		  */
 	} catch (Exception *e) {
-		iproto_write_error(io, e, ::schema_version, msg->header.sync);
+		if (!iproto_shutdown_started)
+			iproto_write_error(io, e, ::schema_version,
```
locker, line 2688: Better catch the proper exception type here and skip logging for it, like we do for socket and timeout errors.

src/box/iproto.cc
```diff
@@ -1485,6 +1506,37 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
...
+{
+	struct iproto_thread *iproto_thread =
+		container_of(m, struct iproto_thread, shutdown_msg);
+	fiber_cond_signal(&iproto_thread->tx.shutdown_cond);
+}
+
+/** Check whether iproto thread shutdown is finished. */
+static void
+iproto_check_shutdown_finished(struct iproto_thread *iproto_thread)
+{
```
locker, line 1521: Bad name: seems like this function just checks something and returns bool while in fact it initiates shutdown.

src/box/iproto.cc
```diff
@@ -730,6 +745,13 @@ struct iproto_connection
 	char salt[IPROTO_SALT_SIZE];
 	/** Iproto connection thread */
 	struct iproto_thread *iproto_thread;
+	/** Link in iproto_thread->connections. */
+	struct rlist in_connections;
+	/**
+	 * The connection is processing replication command so that
+	 * IO is handled by relay code.
+	 */
+	bool is_in_replication;
```
locker, line 754: I don't like that we try to stop replication connections the same way as other iproto connections. I think we should introduce a separate function that would stop relays and call it instead of forcefully cancelling the fibers handling replication connections. Rationale: relays may not have proper cancellation points; e.g. a relay fiber may be uninterruptedly waiting for some other thread, which should be stopped first.

src/lib/core/fiber.c
```diff
@@ -2110,6 +2128,8 @@ cord_cojoin(struct cord *cord)
 		 */
 		do {
 			assert(cord->id != 0);
+			if (fiber_is_cancelled())
+				ev_async_send(cord->loop, &cord->cancel_event);
```
locker, line 2132: This looks dubious. I don't think it's a good idea to stop a thread upon cancelling a fiber trying to join it. IMO it looks unexpected and may obscure the code. We need to discuss that.

src/box/vy_scheduler.c
```diff
@@ -356,7 +356,13 @@ vy_worker_pool_stop(struct vy_worker_pool *pool)
...
 	for (int i = 0; i < pool->size; i++) {
 		struct vy_worker *worker = &pool->workers[i];
-		cord_cancel_and_join(&worker->cord);
+		cbus_stop_loop(&worker->worker_pipe);
+		cpipe_destroy(&worker->worker_pipe);
+	}
+	for (int i = 0; i < pool->size; i++) {
+		struct vy_worker *worker = &pool->workers[i];
+		if (cord_join(&worker->cord) != 0)
+			panic_syserror("iproto cord join failed");
```
locker, line 365: You're effectively reverting e463128e4bd40 here. Looks like no using `cbus_stop_loop` was a deliberate decision.

src/box/engine.c
```diff
@@ -228,7 +228,13 @@ engine_prepare_join(struct engine_join_ctx *ctx)
...
 engine_join(struct engine_join_ctx *ctx, struct xstream *stream)
 {
-	ERROR_INJECT_YIELD(ERRINJ_ENGINE_JOIN_DELAY);
+	ERROR_INJECT_WHILE(ERRINJ_ENGINE_JOIN_DELAY, {
+		fiber_sleep(0.001);
+		if (fiber_is_cancelled()) {
+			diag_set(FiberIsCancelled);
+			return -1;
+		}
+	});
```
locker, line 237: Do we really need to care about this? Tests shouldn't leave the error injection set at exit. Did you encounter any problems because of this injection?

src/box/relay.cc
```diff
@@ -468,6 +451,7 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 	/* Freeze a read view in engines. */
 	struct engine_join_ctx ctx;
 	engine_prepare_join_xc(&ctx);
+	say_info("readview done");
```
locker, line 454: Looks like a debug print.

src/box/relay.cc
```diff
@@ -533,6 +517,7 @@ relay_final_join_f(va_list ap)
 	auto guard = make_scoped_guard([=] { relay_exit(relay); });
 
 	relay_cord_init(relay);
+	say_info("final join started");
```
locker, line 520: Another debug print.

src/box/vinyl.c
```diff
@@ -3102,6 +3102,10 @@ vinyl_engine_join(struct engine *engine, void *arg, struct xstream *stream)
 			if (vy_join_send_tuple(stream, it->lsm->space_id,
 					       entry.stmt) != 0)
 				return -1;
+			if (fiber_is_cancelled()) {
+				diag_set(FiberIsCancelled);
+				return -1;
+			}
```
locker, line 3108: This looks redundant - `vy_join_send_tuple` calls `relay_send_initial_join_row`, which is a cancellation point.

src/box/memtx_engine.cc
```diff
@@ -1327,6 +1311,11 @@ memtx_join_f(va_list ap)
 		while (true) {
 			RegionGuard region_guard(&fiber()->gc);
 			struct read_view_tuple result;
+			if (fiber_is_cancelled()) {
+				diag_set(FiberIsCancelled);
+				rc = -1;
+				break;
+			}
```
locker, line 1318: Again, looks redundant because `memtx_join_send_tuple` calls `relay_initial_join` via `xstream_write`, which is a cancellation point.

src/box/iproto.cc
```diff
@@ -3660,19 +3803,8 @@ iproto_req_handler_delete(struct iproto_req_handler *handler)
...
-		 * failing to bind in case it tries to bind before socket
-		 * is closed by OS.
-		 */
-		evio_service_detach(&iproto_threads[i].binary);
-		rmean_delete(iproto_threads[i].rmean);
-		rmean_delete(iproto_threads[i].tx.rmean);
-		slab_cache_destroy(&iproto_threads[i].net_slabc);
-	}
+	for (int i = 0; i < iproto_threads_count; i++)
+		iproto_thread_destroy(&iproto_threads[i]);
```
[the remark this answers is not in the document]
nshy: Discussed f2f and decided to don't change `cord_cancel_and_join` here in scope of this PR.

src/box/iproto.cc
```diff
@@ -3461,6 +3567,43 @@ iproto_send_start_msg(void)
...
+		struct iproto_msg *msg;
+		rlist_foreach_entry(msg, &iproto_threads[i].tx.inprogress,
+				    in_inprogress)
+			fiber_cancel(msg->fiber);
+		/*
+		 * We should not yield in between so we won't cancel processing
+		 * required for closing connections.
+		 */
+		assert(csw == fiber()->csw);
+		(void)csw;
```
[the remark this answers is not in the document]
nshy: The way we cancel TX fibers processing requests is changed. We don't need this check anymore.

src/box/iproto.cc
nshy: Done.

```c
/**
 * Drop all current connections. That is stop IO and cancel all inprogress
 * requests. Return when the requests are finished and connection is freed.
 * Concurrent calls are serialized.
 */
void
iproto_drop_connections(void);
```

src/box/iproto.cc
nshy: Replaced with per connection flag.

src/box/iproto.cc
nshy: Other functions processing requests use `tx_check_msg` which has this check. This handler does not use the function so we have to add it explicitly. Anyway we changed the way we cancel requests which reside in queues and do not need this check anymore.

src/box/iproto.cc
nshy: Changed to the function which only sends message.
```c
/** Send message to TX thread to notify that connections drop is finished. */
static void
iproto_send_drop_finished(struct iproto_thread *iproto_thread)
```

src/box/vy_scheduler.c
nshy: Discussed f2f and decided to drop this change from the PR. The reason this and similar change got in the PR is to fix ASAN issues in other place. We decided to temporarily add ASAN suppression for that place.

src/box/relay.cc
nshy: This print is used in `replication_luatest/shutdown_test.lua`. But we decided it should be written another way and this version of test is just dropped. So the log is dropped too.

src/box/relay.cc
nshy: This print is used in `replication_luatest/shutdown_test.lua`. But we decided it should be written another way and this version of test is just dropped. So the log is dropped too.

src/box/vinyl.c
nshy: It is not yet a cancellation point. But after discussion f2f we decided to make `coio_write_timeout` a cancellation point. So this change is dropped.

src/box/memtx_engine.cc
nshy: It is not yet a cancellation point. But after discussion f2f we decided to make `coio_write_timeout` a cancellation point. So this change is dropped.

src/lib/core/fiber.c
nshy: Discussed f2f and decided to keep it. Good for default and we can add option to change default behavior in the future.

src/box/iproto.cc
nshy: Discussed f2f and decided keep this way. See the next comment.

nshy:
Do not merge as the first commit will be merged in the scope of PR https://github.com/tarantool/tarantool/pull/9465

src/box/iproto.cc
nshy: Changed to:

```diff
@@ -2621,6 +2701,8 @@ tx_process_replication(struct cmsg *m)
                  * In case of a timeout the error could come after a partially
                  * written row. Do not push it on top.
                  */
+       } catch (FiberIsCancelled *e) {
+               /* Do not write into connection on connection drop. */
        } catch (Exception *e) {
                iproto_write_error(io, e, ::schema_version, msg->header.sync);
        }
```

src/lib/core/coio.c
```diff
@@ -370,6 +370,10 @@ coio_write_timeout(struct iostream *io, const void *buf, size_t sz,
 {
 	ssize_t towrite = sz;
 	ev_tstamp start, delay;
+	if (fiber_is_cancelled()) {
+		diag_set(FiberIsCancelled);
+		return -1;
+	}
```
locker, line 376: Let's also add this to `coio_writev_timeout` and `coio_read_ahead_timeout`.

src/box/iproto.cc
```diff
@@ -213,6 +225,13 @@ struct iproto_thread {
 		size_t requests_in_progress;
 		/** Iproto thread stat collected in tx thread. */
 		struct rmean *rmean;
+		/**
+		 * Connections drop is finished. That is all connections marked
+		 * to be dropped are dropped.
+		 */
+		bool drop_finished;
+		/** Condition for drop_finished. */
+		fiber_cond drop_cond;
```
locker, line 234: Nit: `struct fiber_cond`. Also, better call it `drop_finished_cond`.

src/box/iproto.cc
```diff
@@ -203,6 +203,18 @@ struct iproto_thread {
 	struct evio_service binary;
 	/** Requests count currently pending in stream queue. */
 	size_t requests_in_stream_queue;
+	/** List of all connections. */
+	struct rlist connections;
+	/**
+	 * Number of connections that marked to be dropped but not
+	 * yet dropped.
+	 */
+	size_t drop_count;
```
locker, line 212: Nit: Better call `drop_pending_count`.

src/box/iproto.cc
```diff
@@ -732,6 +757,15 @@ struct iproto_connection
 	 * IO is handled by relay code.
 	 */
 	bool is_in_replication;
+	/** Link in iproto_thread->connections. */
+	struct rlist in_connections;
+	/** Set if connection to be dropped. */
+	bool drop;
```
locker, line 763: Nit: Better call `is_drop_pending`.

src/box/iproto.cc
```diff
@@ -985,8 +1019,6 @@ iproto_connection_close(struct iproto_connection *con)
 		con->state = IPROTO_CONNECTION_CLOSED;
 	} else if (con->state == IPROTO_CONNECTION_PENDING_DESTROY) {
 		iproto_connection_try_to_start_destroy(con);
-	} else {
-		assert(con->state == IPROTO_CONNECTION_CLOSED);
```
locker, line 987: Why?

src/box/iproto.cc
```diff
@@ -1781,6 +1845,17 @@ net_finish_rollback_on_disconnect(struct cmsg *m)
 		iproto_connection_try_to_start_destroy(con);
 }
 
+/** Cancel all inprogress requests of the connection. */
+static void
+tx_process_cancel(struct cmsg *m)
```
locker, line 1850: Nit: "cancel inprogress" may be a better name.

src/box/iproto.cc
```diff
@@ -1958,6 +2033,9 @@ tx_accept_msg(struct cmsg *m)
 	tx_fiber_init(msg->connection->session, msg->header.sync);
 	tx_prepare_transaction_for_request(msg);
 	msg->connection->iproto_thread->tx.requests_in_progress++;
+	rlist_add_entry(&msg->connection->tx.inprogress, msg,
+			in_inprogress);
+	msg->fiber = fiber();
```
locker, line 2038: Now, we could probably drop the `accepted` flag and use `msg->fiber != NULL` instead.

src/box/iproto.cc
```diff
@@ -3397,6 +3490,26 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
...
+				    in_connections) {
+			if (!con->is_in_replication)
+				iproto_connection_close(con);
+			con->drop = true;
+			iproto_thread->drop_count++;
+			if (con->state != IPROTO_CONNECTION_DESTROYED) {
+				cmsg_init(&con->cancel_msg, cancel_route);
+				cpipe_push(&iproto_thread->tx_pipe,
+					   &con->cancel_msg);
+			}
```
locker, line 3507: No need to set the `drop` flag for already destroyed connections.

src/box/iproto.cc
```diff
@@ -3397,6 +3490,26 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
...
 		break;
 	}
+	case IPROTO_CFG_DROP_CONNECTIONS: {
+		struct iproto_connection *con;
+		static const struct cmsg_hop cancel_route[1] =
+						{{ tx_process_cancel, NULL }};
+		rlist_foreach_entry(con, &iproto_thread->connections,
+				    in_connections) {
+			if (!con->is_in_replication)
+				iproto_connection_close(con);
```
locker, line 3500: Would be nice to have a comment somewhere explaining why we skip replication connections.

src/box/iproto.cc
```diff
@@ -3460,6 +3573,26 @@ iproto_send_start_msg(void)
...
+		struct iproto_cfg_msg *cfg_msg =
+			(struct iproto_cfg_msg *)xmalloc(sizeof(*cfg_msg));
+		iproto_cfg_msg_create(cfg_msg, IPROTO_CFG_DROP_CONNECTIONS);
+		iproto_do_cfg_async(&iproto_threads[i], cfg_msg);
+	}
+
+	for (int i = 0; i < iproto_threads_count; i++) {
+		while (!iproto_threads[i].tx.drop_finished)
+			fiber_cond_wait(&iproto_threads[i].tx.drop_cond);
+		iproto_threads[i].tx.drop_finished = false;
```
locker, line 3591: It looks like we don't need to have these flag and condition variable per thread.

locker (changes requested):

src/lib/core/coio.c
nshy: Done. Also moved hunk in `coio_write_timeout`.

<details>
<summary>diff</summary>

```diff
--- a/src/lib/core/coio.c
+++ b/src/lib/core/coio.c
@@ -270,6 +270,10 @@ coio_read_ahead_timeout(struct iostream *io, void *buf, size_t s>
                        size_t bufsiz, ev_tstamp timeout)
 {
        assert(sz <= bufsiz);
+       if (fiber_is_cancelled()) {
+               diag_set(FiberIsCancelled);
+               return -1;
+       }
        ev_tstamp start, delay;
        coio_timeout_init(&start, &delay, timeout);
        ssize_t to_read = (ssize_t) sz;
@@ -368,6 +372,10 @@ ssize_t
 coio_write_timeout(struct iostream *io, const void *buf, size_t sz,
                   ev_tstamp timeout)
 {
+       if (fiber_is_cancelled()) {
+               diag_set(FiberIsCancelled);
+               return -1;
+       }
        ssize_t towrite = sz;
        ev_tstamp start, delay;
        coio_timeout_init(&start, &delay, timeout);
@@ -422,6 +430,10 @@ ssize_t
 coio_writev_timeout(struct iostream *io, struct iovec *iov, int iovcnt,
                    size_t size_hint, ev_tstamp timeout)
 {
+       if (fiber_is_cancelled()) {
+               diag_set(FiberIsCancelled);
+               return -1;
+       }
        size_t total = 0;
        size_t iov_len = 0;
        struct iovec *end = iov + iovcnt;
```
</details>

src/box/iproto.cc
nshy: Done. Also moved flag and condition to the global level.

<details>
<summary>diff</summary>

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 7bf81d242..28320b355 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -225,16 +225,14 @@ struct iproto_thread {
 		size_t requests_in_progress;
 		/** Iproto thread stat collected in tx thread. */
 		struct rmean *rmean;
-		/**
-		 * Connections drop is finished. That is all connections marked
-		 * to be dropped are dropped.
-		 */
-		bool drop_finished;
-		/** Condition for drop_finished. */
-		fiber_cond drop_cond;
 	} tx;
 };
 
+/** Condition for drop finished. */
+static struct fiber_cond drop_finished_cond;
+/** Count of iproto threads that are not finished connections drop yet. */
+static size_t drop_pending_count;
+
 /**
  * IPROTO listen URIs. Set by box.cfg.listen.
  */
@@ -1526,10 +1524,10 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 static void
 tx_process_drop_finished(struct cmsg *m)
 {
-	struct iproto_thread *iproto_thread =
-		container_of(m, struct iproto_thread, drop_finished_msg);
-	iproto_thread->tx.drop_finished = true;
-	fiber_cond_signal(&iproto_thread->tx.drop_cond);
+	(void)m;
+	assert(drop_pending_count > 0);
+	if (--drop_pending_count == 0)
+		fiber_cond_signal(&drop_finished_cond);
 }
 
 /** Send message to TX thread to notify that connections drop is finished. */
@@ -3293,7 +3291,6 @@ iproto_thread_init(struct iproto_thread *iproto_thread)
 	iproto_thread->tx.requests_in_progress = 0;
 	iproto_thread->requests_in_stream_queue = 0;
 	rlist_create(&iproto_thread->connections);
-	fiber_cond_create(&iproto_thread->tx.drop_cond);
 }
 
 /** Initialize the iproto subsystem and start network io thread */
@@ -3315,6 +3312,7 @@ iproto_init(int threads_count)
 	evio_service_create(loop(), &tx_binary, "tx_binary", NULL, NULL);
 	iproto_threads = (struct iproto_thread *)
 		xcalloc(threads_count, sizeof(struct iproto_thread));
+	fiber_cond_create(&drop_finished_cond);
 
 	for (int i = 0; i < threads_count; i++, iproto_threads_count++) {
 		struct iproto_thread *iproto_thread = &iproto_threads[i];
@@ -3578,6 +3576,7 @@ iproto_drop_connections(void)
 {
 	static struct latch latch = LATCH_INITIALIZER(latch);
 	latch_lock(&latch);
+	drop_pending_count = iproto_threads_count;
 	for (int i = 0; i < iproto_threads_count; i++) {
 		struct iproto_cfg_msg *cfg_msg =
 			(struct iproto_cfg_msg *)xmalloc(sizeof(*cfg_msg));
@@ -3586,9 +3585,8 @@ iproto_drop_connections(void)
 	}
 
 	for (int i = 0; i < iproto_threads_count; i++) {
-		while (!iproto_threads[i].tx.drop_finished)
-			fiber_cond_wait(&iproto_threads[i].tx.drop_cond);
-		iproto_threads[i].tx.drop_finished = false;
+		while (drop_pending_count != 0)
+			fiber_cond_wait(&drop_finished_cond);
 	}
 	latch_unlock(&latch);
 }
@@ -3816,6 +3814,7 @@ iproto_free(void)
 		iproto_req_handler_delete(handler);
 	}
 	mh_i32ptr_delete(tx_req_handlers);
+	fiber_cond_destroy(&drop_finished_cond);
 
 	/*
 	 * Here we close sockets and unlink all unix socket paths.

```

</details>

src/box/iproto.cc
nshy: Renamed to `drop_pending_connection_count` as discussed f2f.

<details>
<summary>diff</summary>

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 8df0d36f2..7b99ece99 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -205,11 +205,8 @@ struct iproto_thread {
 	size_t requests_in_stream_queue;
 	/** List of all connections. */
 	struct rlist connections;
-	/**
-	 * Number of connections that marked to be dropped but not
-	 * yet dropped.
-	 */
-	size_t drop_count;
+	/** Number of connections that pending drop. */
+	size_t drop_pending_connection_count;
 	/**
 	 * Message used to notify TX thread that all connections marked
 	 * to de dropped are dropped.
@@ -1564,8 +1561,8 @@ iproto_connection_delete(struct iproto_connection *con)
 	if (con->drop) {
 		struct iproto_thread *iproto_thread = con->iproto_thread;
 
-		assert(iproto_thread->drop_count > 0);
-		if (--iproto_thread->drop_count == 0)
+		assert(iproto_thread->drop_pending_connection_count > 0);
+		if (--iproto_thread->drop_pending_connection_count == 0)
 			iproto_send_drop_finished(iproto_thread);
 	}
 	mempool_free(&con->iproto_thread->iproto_connection_pool, con);
@@ -3497,14 +3494,14 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
 			if (!con->is_in_replication)
 				iproto_connection_close(con);
 			con->drop = true;
-			iproto_thread->drop_count++;
+			iproto_thread->drop_pending_connection_count++;
 			if (con->state != IPROTO_CONNECTION_DESTROYED) {
 				cmsg_init(&con->cancel_msg, cancel_route);
 				cpipe_push(&iproto_thread->tx_pipe,
 					   &con->cancel_msg);
 			}
 		}
-		if (iproto_thread->drop_count == 0)
+		if (iproto_thread->drop_pending_connection_count == 0)
 			iproto_send_drop_finished(iproto_thread);
 		break;
 	}

```

</details>

src/box/iproto.cc
nshy: Done.

<details>
<summary>diff</summary>

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 7b99ece99..a75010bcb 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -754,8 +754,8 @@ struct iproto_connection
 	bool is_in_replication;
 	/** Link in iproto_thread->connections. */
 	struct rlist in_connections;
-	/** Set if connection to be dropped. */
-	bool drop;
+	/** Set if connection is being dropped. */
+	bool is_drop_pending;
 	/**
 	 * Messaged sent to TX to cancel all inprogress requests of the
 	 * connection.
@@ -1503,7 +1503,7 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	con->long_poll_count = 0;
 	con->session = NULL;
 	con->is_in_replication = false;
-	con->drop = false;
+	con->is_drop_pending = false;
 	rlist_create(&con->in_stop_list);
 	rlist_create(&con->tx.inprogress);
 	rlist_add_entry(&iproto_thread->connections, con, in_connections);
@@ -1558,7 +1558,7 @@ iproto_connection_delete(struct iproto_connection *con)
 	assert(mh_size(con->streams) == 0);
 	mh_i64ptr_delete(con->streams);
 	rlist_del(&con->in_connections);
-	if (con->drop) {
+	if (con->is_drop_pending) {
 		struct iproto_thread *iproto_thread = con->iproto_thread;
 
 		assert(iproto_thread->drop_pending_connection_count > 0);
@@ -2866,7 +2866,7 @@ net_end_join(struct cmsg *m)
 	assert(! ev_is_active(&con->input));
 	con->is_in_replication = false;
 
-	if (con->drop) {
+	if (con->is_drop_pending) {
 		iproto_connection_close(con);
 		return;
 	}
@@ -3493,7 +3493,7 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
 				    in_connections) {
 			if (!con->is_in_replication)
 				iproto_connection_close(con);
-			con->drop = true;
+			con->is_drop_pending = true;
 			iproto_thread->drop_pending_connection_count++;
 			if (con->state != IPROTO_CONNECTION_DESTROYED) {
 				cmsg_init(&con->cancel_msg, cancel_route);

```

</details>

src/box/iproto.cc
nshy: On connection drop we call `iproto_connection_close` for connection in any state. Let's call only for alive.

```diff
@@ -3491,7 +3493,8 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
                                                {{ tx_process_cancel, NULL }};
                rlist_foreach_entry(con, &iproto_thread->connections,
                                    in_connections) {
-                       if (!con->is_in_replication)
+                       if (!con->is_in_replication &&
+                           con->state == IPROTO_CONNECTION_ALIVE)
                                iproto_connection_close(con);
                        con->is_drop_pending = true;
                        iproto_thread->drop_pending_connection_count++;
```

src/box/iproto.cc
nshy: Done.

```diff
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1844,7 +1844,7 @@ net_finish_rollback_on_disconnect(struct cmsg *m)

 /** Cancel all inprogress requests of the connection. */
 static void
-tx_process_cancel(struct cmsg *m)
+tx_process_cancel_inprogress(struct cmsg *m)
 {
        struct iproto_connection *con =
                container_of(m, struct iproto_connection, cancel_msg);
@@ -3490,7 +3490,7 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
        case IPROTO_CFG_DROP_CONNECTIONS: {
                struct iproto_connection *con;
                static const struct cmsg_hop cancel_route[1] =
-                                               {{ tx_process_cancel, NULL }};
+                               {{ tx_process_cancel_inprogress, NULL }};
                rlist_foreach_entry(con, &iproto_thread->connections,
                                    in_connections) {
                        if (!con->is_in_replication &&
```

src/box/iproto.cc
nshy: Done.

<details>
<summary>diff</summary>

```diff
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -451,12 +451,6 @@ struct iproto_msg
        struct stailq_entry in_stream;
        /** Stream that owns this message, or NULL. */
        struct iproto_stream *stream;
-       /**
-        * True if message processing in tx thread is started. The flag is
-        * used to prevent double accounting of message in case of iproto
-        * override fallback to original handler.
-        */
-       bool accepted;
        /** Link in connection->tx.inprogress. */
        struct rlist in_inprogress;
        /** TX thread fiber that processing this message. */
@@ -848,7 +842,7 @@ iproto_msg_new(struct iproto_connection *con)
        msg->close_connection = false;
        msg->connection = con;
        msg->stream = NULL;
-       msg->accepted = false;
+       msg->fiber = NULL;
        rmean_collect(con->iproto_thread->rmean, IPROTO_REQUESTS, 1);
        return msg;
 }
@@ -2023,9 +2017,8 @@ static inline struct iproto_msg *
 tx_accept_msg(struct cmsg *m)
 {
        struct iproto_msg *msg = (struct iproto_msg *) m;
-       if (msg->accepted)
+       if (msg->fiber != NULL)
                return msg;
-       msg->accepted = true;
        tx_accept_wpos(msg->connection, &msg->wpos);
        tx_fiber_init(msg->connection->session, msg->header.sync);
        tx_prepare_transaction_for_request(msg);
```

</details>

src/box/iproto.cc
nshy: We want to wait when all connections are freed on shutdown.

src/box/iproto.cc
nshy: Done.

```diff
@@ -3486,6 +3486,12 @@ iproto_do_cfg_f(struct cbus_call_msg *m)
                                {{ tx_process_cancel_inprogress, NULL }};
                rlist_foreach_entry(con, &iproto_thread->connections,
                                    in_connections) {
+                       /*
+                        * Replication IO is done outside iproto so we
+                        * cannot close them as usual. Anyway we cancel
+                        * replication fibers as well and close connection
+                        * after replication is breaked.
+                        */
                        if (!con->is_in_replication &&
                            con->state == IPROTO_CONNECTION_ALIVE)
                                iproto_connection_close(con);
```

src/box/iproto.cc
nshy: Fixed. Patch is in the one of the previous comments.

src/box/iproto.cc
```diff
@@ -3460,6 +3570,26 @@ iproto_send_start_msg(void)
...
+	latch_lock(&latch);
+	drop_pending_thread_count = iproto_threads_count;
+	for (int i = 0; i < iproto_threads_count; i++) {
+		struct iproto_cfg_msg *cfg_msg =
+			(struct iproto_cfg_msg *)xmalloc(sizeof(*cfg_msg));
+		iproto_cfg_msg_create(cfg_msg, IPROTO_CFG_DROP_CONNECTIONS);
+		iproto_do_cfg_async(&iproto_threads[i], cfg_msg);
+	}
+
+	for (int i = 0; i < iproto_threads_count; i++) {
```
locker, line 3586: This loop is obviously redundant now.

locker (changes requested):

locker (approved):

Lord-KA:
FIY: I've tested the `iproto_drop_connections()` handle and it works just as required in scope of my task (tarantool/tarantool-ee#585).

src/box/iproto.cc
nshy: Fixed)

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index 0d0a0864c..cf36bbc2b 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -3583,10 +3583,8 @@ iproto_drop_connections(void)
                iproto_do_cfg_async(&iproto_threads[i], cfg_msg);
        }

-       for (int i = 0; i < iproto_threads_count; i++) {
-               while (drop_pending_thread_count != 0)
-                       fiber_cond_wait(&drop_finished_cond);
-       }
+       while (drop_pending_thread_count != 0)
+               fiber_cond_wait(&drop_finished_cond);
        latch_unlock(&latch);
 }
```

Issue #7283, opened 2022-06-19
SQL: prepared join fails to execute


darthunix:
**Bug description**
Intro. When we prepare a statement with parameters in the result columns (for example `box.prepare('select ?')`) Tarantool has no information about the type of the output column and set it to default [boolean](https://github.com/tarantool/tarantool/blob/b1953b59a94f75261b74fdcaadc92d4ceba6f84b/src/box/sql/expr.c#L1358). Then, on the execution phase, the type would be [recalculated](https://github.com/tarantool/tarantool/blob/b1953b59a94f75261b74fdcaadc92d4ceba6f84b/src/box/sql/vdbeapi.c#L414) during the parameter [binding](https://github.com/tarantool/tarantool/blob/b1953b59a94f75261b74fdcaadc92d4ceba6f84b/src/box/execute.c#L599).

Tarantool expects that there is no other way for parameter to appear in the result tuple other than exactly be mentioned it in the final projection. But it is incorrect - we can easily propagate  parameter from the inner part of the join (for example, `box.prepare([[select COLUMN_1 from t1 join (values (?)) as t2 on true]])`). In this case column `COLUMN_1` in the final projection is not a parameter, but a "reference" to it and its type depends on the parameter from the inner part of the join. But as Tarantool recalculates only binded parameters in the result projection, it doesn't change the default `boolean` metadata type of the `COLUMN_1` and the query fails on [comparison](https://github.com/tarantool/tarantool/blob/2a83f9e53fc20191e8e6bf660218289a25494a34/src/box/tuple_format.c#L1042) with the actual type of the parameter.

**Steps to reproduce**
```
box.execute([[create table t1(a int primary key)]], {})
q = box.prepare([[select COLUMN_1 from t1 join (values (?)) as t2 on true]])
q:execute({42})
---
- null
- 'Tuple field 1 (_COLUMN_0) type does not match one required by operation: expected
  boolean, got unsigned'
...
```

**Solution**
It seems that `Vdbe` structure doesn't keep any information about `COLUMN_1` referencing the parameter from the inner part of the join. Considering this fact I can suggest two solutions:
1. Simply change the default type from `boolean` to `any` for parameters (something like this)
   ```
    diff --git a/src/box/sql/expr.c b/src/box/sql/expr.c
    index 67c4cdd85..3fde451d3 100644
    --- a/src/box/sql/expr.c
    +++ b/src/box/sql/expr.c
[...]

Issue #6363, opened 2021-08-19
Find a way to determine type of result of COALESCE() during parsing.


ImeevMA:
Coalesce returns first not-NULL argument. For now I am not sure how we should calculate type of result of this function. As I see, we have a few options:
1) Always return SCALAR.
2) Determine type of result using this approach:
 a) if all arguments are NULLs and this can be determined during parsing, the type of the result is SCALAR;
 b) if all arguments have the same field type, the type of the result is this field type;
 c) if all arguments of numeric types, and at least two arguments have different types, the type of the result is NUMBER;
 d) otherwise the type of the result is SCALAR.
In this approach in case we can determine that value is NULL during parsing, it will be ignored during calculation of the type of the result.

PR #9462, opened 2023-12-11
Correct type cast for COALESCE and IFNULL

darthunix:
Originally, the problem was inspired by the error in SQL query:
```
picodata> box.execute([[select abs(coalesce(1, 0))]])
---
- null
- 'Failed to execute SQL statement: wrong arguments for function ABS()'
...
```
The root of the problem is that COALESCE always returns scalar type for any input
parameters (even for arrays or maps). It also allows to mix
different types in its arguments, that is not good for the
query type checking in SQL. Moreover, ABS operator didn't work with
COALESCE as a child, because COALESCE returns scalar while
ABS allowed only numeric types for its children. It was also found out that IFNULL has the same
problems (in fact it is a shorter version of COALESCE operator with exactly two parameters).
So,  it was decided to implement more explicit and strict type checks for these operators.

Later on (while writing tests), it was found out that we still can't perform
```
select abs(coalesce(?, ?))
```
because parameters are always of the boolean type. So, I have also fixed
https://github.com/tarantool/tarantool/issues/7283 to switch to the any type for parameters.


test/sql-luatest/coalesce_test.lua
```diff
@@ -0,0 +1,78 @@
...
+g.before_all(function()
+    g.server = server:new({alias = 'master'})
+    g.server:start()
+end)
+
+g.after_all(function()
+    g.server:stop()
+end)
+
+g.test_coalesce_1 = function()
```
ochaplashkin, line 15: name of the test does not tell us very transparently what is going on.

`test_something_1`, `test_something_2` is a bad practice.
Simple template:

`test_<what>_<when>_<additional condition>`

ochaplashkin (changes requested):

test/sql-luatest/gh_7283_result_parameter_type_test.lua 
```diff
@@ -0,0 +1,28 @@
...
+    g.server:start()
+end)
+
+g.after_all(function()
+    g.server:stop()
+end)
+
+g.test_result_parameter_type_1 = function()
+    g.server:exec(function()
+        local t = require('luatest')
```
ochaplashkin, line 18: `Luatest` is already available on your server instance.
Feel free to use it `t.assert(...)` 👍

ochaplashkin (approved):

changelogs/unreleased/gh-7283-result-parameter-type.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/sql
+
+Fixes a failing inner join query, projecting parameters from the inner
+table to the result projection. As a side effect the default column
+metadata was changed from `boolean` to `any` type.
```
xuniq, line 5:
```suggestion
Fixed a failing inner join query, projecting parameters from the inner
table to the result projection. As a side effect, changed the default type of column
metadata from `boolean` to `any`.
```

xuniq (approved):

darthunix:
@ImeevMA hi, I have found out that current PR is still on review. Are there any problems with it?

ImeevMA:
I remember the last time I reviewed a PR, I was planning to get the product manager's decision on some issue. I don’t remember exactly what it was, but at that time I couldn’t get a definite answer. I'll reviewed it again today.

ImeevMA (changes requested):
Hi, @darthunix! Thank you for the patch-set. Why would we want to forbid mix different types in `COALESCE` arguments? Also, your patches don't actually prohibit mixing different types, because the following can still be done:
```Lua
tarantool> box.execute([[SELECT COALESCE(1, 'a');]])
---
- metadata:
  - name: COLUMN_1
    type: any
  rows:
  - [1]
...
tarantool> box.execute([[SELECT COALESCE(1, 1.5, 2e0);]])
---
- metadata:
  - name: COLUMN_1
    type: number
  rows:
  - [1]
...

```

I think there is something strange with type `ANY`, please see my comments below.

In general, one of the reasons I am not too fond on this change is that I think we should rework the result types calculation system for built-in SQL functions. I've already had an [unsuccessful] attempt to do this.

I'm not sure that forbidding mixed types is the right thing to do. There is an issue related to this: https://github.com/tarantool/tarantool/issues/6363. The issue was filled when there was no ANY support, so the issue description should now be expanded a bit. I think with a little work your patches could solve this issue.

PR #9464, opened 2023-12-11
integration testing

fckxorg:
Currently, if there is a need to test submodule integration with Tarantool and its integration, it is required to create a PR. That is inconvenient, so this patch introduces the option to run the same jobs that are triggered by the `full-ci` label as reusable workflows with the desired submodule revision. This allows for integration testing of submodules within their designated repositories.

### DIFF
--- a/.github/actions/bump-submodule/action.yml
+++ b/.github/actions/bump-submodule/action.yml
@@ -1,0 +1,36 @@
+name: 'Bump submodule'
+description: 'Bump specified submodule from third_party'
+
+inputs:
+  submodule:
+    description: Name of submodule to bump.
+    required: true
+    type: string
+  revision:
+    description: Git revision from submodule repository
+    required: true
+    type: string
+
+runs:
+  using: 'composite'
+  # XXX: Environment variables to suppress git config warning.
+  # The commit is required to avoid resetting dirty
+  # changes in submodule.
+  env:
+    GIT_COMMITTER_NAME: "integration testing workflow"
+    GIT_COMMITTER_EMAIL: "noreply@tarantool.org"
+    GIT_AUTHOR_NAME: "integration testing workflow"
+    GIT_AUTHOR_EMAIL: "noreply@tarantool.org"
+  steps:
+    run: |
+      pushd third_party/${{ inputs.submodule }}
+      git fetch origin
+      git checkout ${{ inputs.revision }}
+      popd
+      # XXX: --allow-empty is required to rerun flaky tests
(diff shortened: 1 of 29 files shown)


.github/actions/bump-submodule/action.yml
```diff
@@ -0,0 +1,35 @@
...
+      git fetch origin
+      git checkout ${{ inputs.revision }}
+      popd
+      # XXX: export the following environment variables to
+      # suppress git config warning. The commit is required to
+      # avoid resetting dirty changes in submodule.
+      export GIT_COMMITTER_NAME="integration testing workflow"
+      export GIT_COMMITTER_EMAIL="noreply@tarantool.org"
+      export GIT_AUTHOR_NAME="integration testing workflow"
+      export GIT_AUTHOR_EMAIL="noreply@tarantool.org"
```
ochaplashkin, line 28: It looks like it can be moved to another section.

```yml
# as a global variable
env:
  MY_VAR: foo


jobs:
  my_job:
    env:
      # as a local for all steps
      MY_VAR: foo
    steps:
      - name: ...
        run: ...
        env:
          # as a local variable
          MY_VAR: foo
```
Using variables:
```yml
run: echo "$MY_VAR"
```

[Documentation](https://docs.github.com/en/actions/learn-github-actions/variables)

ochaplashkin:
Do not forget about commit message:

```
...

NO_DOC: <your_reason>
NO_TEST: <your_reason>
NO_CHANGELOG: <your_reason>
```

.github/actions/bump-submodule/action.yml
fckxorg: Fixed, thanks!

ochaplashkin (approved):
LGTM :rocket:

.github/workflows/static_build.yml
```diff
@@ -10,6 +10,106 @@ on:
...
+      submodule:
+        description: Name of submodule to bump.
+        required: true
+        type: string
+      revision:
+        description: Git revision from submodule repository
+        required: true
+        type: string
+
+concurrency:
```
igormunkin, line 24: Looks like miscopypasting.

igormunkin (approved):
@fckxorg, thanks for the patch! LGTM, but `static_build` workflow looks bad (consider my comment below).

.github/workflows/static_build.yml
fckxorg: Fixed, thanks!

PR #9465, opened 2023-12-11
replication: don't rollback qsync limbo wait on fiber cancel

### LINKED ISSUE #9480: Fiber cancellation triggers a synchronous transaction rollback
**Bug description**
Fiber cancellation leads to the synchronous transaction rollback.
According to Raft, only the new leader may roll back some of the ongoing transactions, the current leader should never rollback the transactions it has started.

Issue #7486 states the same problem from the different angle: leader shouldn't roll back the transactions by timeout, it shouldn't roll them back due to fiber cancellation. The transaction should stay in limbo.

Otherwise the user might trigger a split-brain by simply cancelling a fiber which hangs waiting for quorum on the old leader.

**Reproducer**
```lua
-- Start 1 server in console.
box.cfg{replication_synchro_quorum = 2, replication_synchro_timeout=1000, memtx_use_mvcc_engine = 'off'}
box.ctl.promote()
box.schema.space.create('test', {is_sync=true}):create_index('pk')
fiber = require("fiber")
f = fiber.new(function() box.space.test:insert{1} end)
f:set_joinable(true)
f:cancel()
```
Now observe
```lua
f:join()
---
- false
- Quorum collection for a synchronous transaction is timed out
...

```
and
```lua
tarantool> box.space.test:fselect{}
---
- |-
  +-----+
  |col1 |
  +-----+
  +-----+
...

```

nshy:
During iproto graceful shutdown which is WIP we cancel all iproto request in progress. This causes election_qsync_stress test failure.

We shutdown master on waiting transaction confirmation from quorum (which is never exist in this test). Currently on shutdown we rollback transaction in this state. So that when previous master is restarted after electing new master we don't expect the rollback on previous master.

Let's keep the transaction in limbo if fiber is cancelled as our direction is to do only quorum rollbacks.

Part of #8423
Closes #9480

### DIFF
--- a/changelogs/unreleased/gh-9480-dont-rollback-on-quorum-wait-cancel.md
+++ b/changelogs/unreleased/gh-9480-dont-rollback-on-quorum-wait-cancel.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Now transactions are not rolled back if the transaction fiber is
+  cancelled when waiting for quorum from replicas (gh-9480).
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1105,8 +1105,15 @@ txn_commit(struct txn *txn)
 			txn_limbo_ack(&txn_limbo, txn_limbo.owner_id,
 				      limbo_entry->lsn);
 		}
-		if (txn_limbo_wait_complete(&txn_limbo, limbo_entry) < 0)
-			goto rollback;
+		int rc = txn_limbo_wait_complete(&txn_limbo, limbo_entry);
+		if (rc < 0) {
+			if (fiber_is_cancelled()) {
+				txn->fiber = NULL;
+				return -1;
+			} else {
+				goto rollback;
+			}
+		}
 	}
 	assert(txn_has_flag(txn, TXN_IS_DONE));
 	assert(txn->signature >= 0);
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -283,6 +283,8 @@ txn_limbo_wait_complete(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
 		int rc = fiber_cond_wait_timeout(&limbo->wait_cond, timeout);
 		if (txn_limbo_entry_is_complete(entry))
 			goto complete;
+		if (rc != 0 && fiber_is_cancelled())
+			return -1;
 		if (txn_limbo_is_frozen(limbo))
 			goto wait;
 		if (rc != 0)
--- a/src/box/txn_limbo.h
+++ b/src/box/txn_limbo.h
@@ -343,6 +343,7 @@ txn_limbo_ack(struct txn_limbo *limbo, uint32_t replica_id, int64_t lsn);
  * entry is either committed or rolled back.
  * If timeout is reached before acks are collected, the tx is
  * rolled back as well as all the txs in the limbo following it.
+ * If fiber is cancelled before acks are collected, the tx is left in limbo.
  * Returns -1 when rollback was performed and tx has to be freed.
  *          0 when tx processing can go on.
  */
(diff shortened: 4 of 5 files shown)


src/box/txn_limbo.c
```diff
@@ -285,8 +285,11 @@ txn_limbo_wait_complete(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
 			goto complete;
 		if (txn_limbo_is_frozen(limbo))
 			goto wait;
-		if (rc != 0)
+		if (rc != 0) {
+			if (fiber_is_cancelled())
```
sergepetrenko, line 289: The check for `fiber_is_cancelled` should go before the `txn_limbo_is_frozen()` check now. Otherwise a cancelled fiber will continue running until the limbo is unfrozen.

src/box/txn.c
```diff
@@ -1105,8 +1105,15 @@ txn_commit(struct txn *txn)
 			txn_limbo_ack(&txn_limbo, txn_limbo.owner_id,
```
sergepetrenko, line 1105: > NO_TEST=issue is covered with iproto graceful shutdown

I suggest you add a separate test for this issue. It should be quite easy to write one: you only need one server, set quorum to 2 on it, then create a synchronous transaction in a separate fiber and cancel that fiber. If the change is still seen (without mvcc enabled), the test passes.

> NO_CHANGELOG

But this is a user-visible change: now fiber cancellation doesn't lead to synchronous transaction rollback now. AFAIU this is more consistent with async transactions now. They can't be rolled back by a cancellation after a WAL write is started.

sergepetrenko (changes requested):
Thanks for the patch! Please see my 2 comments below.

src/box/txn_limbo.c
nshy: Done. Changed to:

```diff
@@ -283,6 +283,8 @@ txn_limbo_wait_complete(struct txn_limbo *limbo, struct txn_limbo>
                int rc = fiber_cond_wait_timeout(&limbo->wait_cond, timeout);
                if (txn_limbo_entry_is_complete(entry))
                        goto complete;
+               if (rc != 0 && fiber_is_cancelled())
+                       return -1;
                if (txn_limbo_is_frozen(limbo))
                        goto wait;
                if (rc != 0)
```

src/box/txn.c
nshy: Added a test and a changelog:

```md
## bugfix/replication

* Fixed to not rollback a transaction waiting for quorum from replicas
  on transaction fiber cancel (gh-9480).
```

sergepetrenko: Thanks!

sergepetrenko (approved):
Thanks for the quick fixes!
LGTM.

changelogs/unreleased/gh-9480-dont-rollback-on-quorum-wait-cancel.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/replication
+
+* Fixed to not rollback a transaction waiting for quorum from replicas
+  on transaction fiber cancel (gh-9480).
```
p7nov, line 4:
```suggestion
* Now transactions are not rolled back if the transaction fiber is
  cancelled when waiting for quorum from replicas (gh-9480).
```

p7nov (approved):

ochaplashkin (approved):

changelogs/unreleased/gh-9480-dont-rollback-on-quorum-wait-cancel.md
nshy: Done.

sergepetrenko:
Backported to 2.11: https://github.com/tarantool/tarantool/pull/9482

PR #9473, opened 2023-12-12
iproto: override request handlers using event triggers

### LINKED ISSUE #8138: Allow IPROTO request handler overriding before `box.cfg` with `listen`
People can only override IPROTO request handlers after `box.cfg` with `listen`: it means some requests might arrive after listening is started, but before the override happened, hence there is no way to override a request type guaranteed for all its 100% usages — a few requests might slip before the override is complete.

We can fix this issue by allowing to override IPROTO request handlers before `box.cfg` and by setting the corresponding handlers in IPROTO threads prior to `listen`.

The best way is to use brand new trigger registry.

Follow-up #7901

Gumix:
Closes #8138

### DIFF
--- a/changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
+++ b/changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
@@ -1,0 +1,10 @@
+## feature/core
+
+* Now it is possible to set IPROTO request handler callbacks by using the new
+  Lua module `trigger`, for example:
+  ```lua
+  local trigger = require('trigger')
+  trigger.set('box.iproto.override.select', 'my_select', my_select_handler)
+  ```
+  The method works before the initial `box.cfg{}` call. Also, the method allows
+  setting multiple handlers for a single request type (gh-8138).
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -2925,6 +2925,25 @@ iproto_req_handlers_set_event(struct iproto_req_handlers *handlers,
 }
 
 /**
+ * Deletes an event, which is set in `handlers' by request type id (if
+ * `is_by_id'), or by request type name.
+ */
+static void
+iproto_req_handlers_del_event(struct iproto_req_handlers *handlers,
+			      bool is_by_id)
+{
+	assert(handlers != NULL);
+
+	if (is_by_id) {
+		event_unref(handlers->event_by_id);
+		handlers->event_by_id = NULL;
+	} else {
+		event_unref(handlers->event_by_name);
+		handlers->event_by_name = NULL;
+	}
+}
+
+/**
  * Returns `true' if there is at least one handler in `handlers'.
  */
 static bool
@@ -3747,6 +3766,39 @@ iproto_override_finish(struct iproto_req_handlers *handlers, uint32_t req_type,
 	}
 }
 
+/**
+ * Trigger which is fired on any change in the event registry.
+ */
+static int
+trigger_on_change_iproto_notify(struct trigger *trigger, void *arg)
+{
+	(void)trigger;
+	uint32_t type;
+	bool is_by_id;
+	struct event *event = (struct event *)arg;
+	if (!get_iproto_type_from_event_name(event->name, &type, &is_by_id))
+		return 0;
+
+	struct iproto_req_handlers *handlers;
+	handlers = mh_req_handlers_get(type);
+	bool is_set = iproto_req_handler_is_set(handlers);
+
+	if (event_has_triggers(event)) {
+		if (handlers == NULL) {
+			handlers = iproto_req_handlers_new();
+			mh_req_handlers_put(type, handlers);
+		}
+		iproto_req_handlers_set_event(handlers, event, is_by_id);
+	} else {
+		iproto_req_handlers_del_event(handlers, is_by_id);
+	}
+
+	iproto_override_finish(handlers, type, is_set);
+	return 0;
+}
+
+TRIGGER(trigger_on_change, trigger_on_change_iproto_notify);
+
 /** Initialize the iproto subsystem and start network io thread */
 void
 iproto_init(int threads_count)
@@ -3797,6 +3849,7 @@ iproto_init(int threads_count)
 
 	session_vtab_registry[SESSION_TYPE_BINARY] = iproto_session_vtab;
 
+	event_on_change(&trigger_on_change);
 	if (box_on_shutdown(NULL, iproto_on_shutdown_f, NULL) != 0)
 		panic("failed to set iproto shutdown trigger");
 }
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -33,6 +33,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <fcntl.h>
+#include <ctype.h>
 
 #include <msgpuck.h>
 #include <small/ibuf.h>
@@ -41,6 +42,7 @@
 
 #include "version.h"
 #include "event.h"
+#include "func_adapter.h"
 #include "fiber.h"
 #include "fiber_cond.h"
 #include "cbus.h"
@@ -296,9 +298,21 @@ static int iproto_msg_max = IPROTO_MSG_MAX_MIN;
 /**
  * Request handlers meta information. The IPROTO request of each type can be
  * overridden by the following types of handlers (listed in priority order):
- *  1. C handler, set by `iproto_override()'.
+ *  1. Lua handlers, set in the event registry by request type id;
+ *  2. Lua handlers, set in the event registry by request type name;
+ *  3. C handler, set by `iproto_override()'.
  */
 struct iproto_req_handlers {
+	/**
+	 * Triggers from the event registry, set by request type id.
+	 * NULL if no such triggers.
+	 */
+	struct event *event_by_id;
+	/**
+	 * Triggers from the event registry, set by request type name.
+	 * NULL if no such triggers.
+	 */
+	struct event *event_by_name;
 	/**
 	 * C request handler.
 	 */
@@ -2829,6 +2843,10 @@ iproto_req_handlers_new(void)
 static void
 iproto_req_handlers_delete(struct iproto_req_handlers *handlers)
 {
+	if (handlers->event_by_id != NULL)
+		event_unref(handlers->event_by_id);
+	if (handlers->event_by_name != NULL)
+		event_unref(handlers->event_by_name);
 	if (handlers->c.destroy != NULL)
 		handlers->c.destroy(handlers->c.ctx);
 	TRASH(handlers);
@@ -2879,6 +2897,34 @@ mh_req_handlers_del(uint32_t req_type)
 }
 
 /**
+ * Replaces an event in `handlers' by the new `event'. If `is_by_id', the
+ * handler is set by request type id, otherwise it is set by request type name.
+ */
+static void
+iproto_req_handlers_set_event(struct iproto_req_handlers *handlers,
+			      struct event *event, bool is_by_id)
+{
+	assert(handlers != NULL);
+	assert(event != NULL);
+
+	if (is_by_id) {
+		if (handlers->event_by_id == NULL) {
+			event_ref(event);
+			handlers->event_by_id = event;
+		} else {
+			assert(handlers->event_by_id == event);
+		}
+	} else {
+		if (handlers->event_by_name == NULL) {
+			event_ref(event);
+			handlers->event_by_name = event;
+		} else {
+			assert(handlers->event_by_name == event);
+		}
+	}
+}
+
+/**
  * Returns `true' if there is at least one handler in `handlers'.
  */
 static bool
@@ -2887,7 +2933,79 @@ iproto_req_handler_is_set(struct iproto_req_handlers *handlers)
 	if (handlers == NULL)
 		return false;
 
-	return handlers->c.cb != NULL;
+	return handlers->event_by_id != NULL ||
+	       handlers->event_by_name != NULL ||
+	       handlers->c.cb != NULL;
+}
+
+/**
+ * Returns `enum iproto_type' if `name' is a valid IPROTO type name or equals
+ * "unknown". Otherwise, iproto_type_MAX is returned. The name is expected to
+ * be in lowercase.
+ */
+static enum iproto_type
+get_iproto_type_by_name(const char *name)
+{
+	for (uint32_t i = 0; i < iproto_type_MAX; i++) {
+		const char *type_name = iproto_type_name_lower(i);
+		if (type_name != NULL && strcmp(type_name, name) == 0)
+			return (enum iproto_type)i;
+	}
+	if (strcmp(name, "unknown") == 0)
+		return IPROTO_UNKNOWN;
+	return iproto_type_MAX;
+}
+
+/**
+ * Runs triggers registered for the `event'.
+ * The given header and body the IPROTO packet are passed as trigger args.
+ * Returns IPROTO_HANDLER_OK if some trigger successfully handled the request,
+ * IPROTO_HANDLER_FALLBACK if no triggers handled the request, or
+ * IPROTO_HANDLER_ERROR on failure.
+ */
+static enum iproto_handler_status
+tx_run_override_triggers(struct event *event, const char *header,
+			 const char *header_end, const char *body,
+			 const char *body_end)
+{
+	enum iproto_handler_status rc = IPROTO_HANDLER_FALLBACK;
+	const char *name = NULL;
+	struct func_adapter *trigger = NULL;
+	struct func_adapter_ctx ctx;
+	struct event_trigger_iterator it;
+	event_trigger_iterator_create(&it, event);
+
+	while (event_trigger_iterator_next(&it, &trigger, &name)) {
+		struct mp_ctx mp_ctx_header, mp_ctx_body;
+		mp_ctx_create_default(&mp_ctx_header, iproto_key_translation);
+		mp_ctx_create_default(&mp_ctx_body, iproto_key_translation);
+
+		func_adapter_begin(trigger, &ctx);
+		func_adapter_push_msgpack_with_ctx(trigger, &ctx, header,
+						   header_end, &mp_ctx_header);
+		func_adapter_push_msgpack_with_ctx(trigger, &ctx, body,
+						   body_end, &mp_ctx_body);
+		if (func_adapter_call(trigger, &ctx) == 0) {
+			if (func_adapter_is_bool(trigger, &ctx)) {
+				bool ok = false;
+				func_adapter_pop_bool(trigger, &ctx, &ok);
+				if (ok)
+					rc = IPROTO_HANDLER_OK;
+			} else {
+				diag_set(ClientError, ER_PROC_LUA,
+					 "Invalid Lua IPROTO handler return "
+					 "type: expected boolean");
+				rc = IPROTO_HANDLER_ERROR;
+			}
+		} else {
+			rc = IPROTO_HANDLER_ERROR;
+		}
+		func_adapter_end(trigger, &ctx);
+		if (rc != IPROTO_HANDLER_FALLBACK)
+			break;
+	}
+	event_trigger_iterator_destroy(&it);
+	return rc;
 }
 
 /**
@@ -2921,10 +3039,29 @@ tx_process_override(struct cmsg *m)
 	if (handlers == NULL)
 		handlers = mh_req_handlers_get(IPROTO_UNKNOWN);
 	assert(handlers != NULL);
-	enum iproto_handler_status rc;
+	enum iproto_handler_status rc = IPROTO_HANDLER_FALLBACK;
 
-	rc = handlers->c.cb(header, header_end, body, body_end,
-			    handlers->c.ctx);
+	/*
+	 * Run handlers from the event registry, set by request type id.
+	 */
+	if (handlers->event_by_id != NULL) {
+		rc = tx_run_override_triggers(handlers->event_by_id, header,
+					      header_end, body, body_end);
+	}
+	/*
+	 * Run handlers from the event registry, set by request type name.
+	 */
+	if (rc == IPROTO_HANDLER_FALLBACK && handlers->event_by_name != NULL) {
+		rc = tx_run_override_triggers(handlers->event_by_name, header,
+					      header_end, body, body_end);
+	}
+	/*
+	 * Run C handlers.
+	 */
+	if (rc == IPROTO_HANDLER_FALLBACK && handlers->c.cb != NULL) {
+		rc = handlers->c.cb(header, header_end, body, body_end,
+				    handlers->c.ctx);
+	}
 
 	struct cmsg_hop *route = NULL;
 	switch (rc) {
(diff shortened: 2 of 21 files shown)


src/box/iproto.cc
```diff
@@ -3291,6 +3425,81 @@ iproto_thread_init(struct iproto_thread *iproto_thread)
...
+	const char *req_name = name + prefix_len;
+	uint32_t type = iproto_type_MAX;
+
+	if (*req_name == '[') {
+		char *end;
+		req_name++;
+		*is_by_id = true;
+		type = strtol(req_name, &end, 10);
+		if (req_name == end)
+			return iproto_type_MAX;
```
drewdzzz, line 3478: `box.iproto.override[select]` or `box.iproto.override[15].header` are invalid names which could very possibly come here - they should be logged too.
More strictly, let's log change on every event, that belongs to `box.iproto.override` namespace but does not override any request handler.

src/box/iproto.cc
```diff
@@ -2815,9 +2815,16 @@ tx_run_override_triggers(struct event *event, const char *header,
...
+		struct mp_ctx mp_ctx_body;
+		mp_ctx_create_default(&mp_ctx_body, iproto_key_translation);
+
 		func_adapter_begin(trigger, &ctx);
-		func_adapter_push_msgpack(trigger, &ctx, header, header_end);
-		func_adapter_push_msgpack(trigger, &ctx, body, body_end);
+		func_adapter_push_msgpack(trigger, &ctx, header, header_end,
+					  &mp_ctx_header);
+		func_adapter_push_msgpack(trigger, &ctx, body, body_end,
+					  &mp_ctx_body);
```
drewdzzz, line 2827: Lua hashes are used for translation, while `func_adapter` was designed to call arbitrary functions.
Now `func_adapter` is used only for Lua, but we should be ready for new `func_adapter` implementations.
Probably, it will be a good solution to add a used hash function to `struct mp_ctx` in the future - then other languages will use `lua_hash`, because with high probability only Lua has pre-calculated hash on its values.
Let's discuss it @alyapunov before merging this PR.

test/unit/lua_func_adapter.c
```diff
@@ -270,7 +270,7 @@ test_msgpack(void)
 	struct func_adapter *func = func_adapter_lua_create(tarantool_L, idx);
 	struct func_adapter_ctx ctx;
 	func_adapter_begin(func, &ctx);
-	func_adapter_push_msgpack(func, &ctx, mp_buf, mp);
+	func_adapter_push_msgpack(func, &ctx, mp_buf, mp, NULL);
```
drewdzzz, line 273: Let's factor `func_adapter` update out to a separate commit and add a unit test for translation.

src/box/space.c
```diff
@@ -813,12 +813,12 @@ space_run_replace_triggers(struct event *event, struct txn *txn,
...
 			assert(row != NULL && row->header != NULL);
 			func_adapter_push_msgpack(trigger, &ctx, row->header,
-						  row->header_end);
+						  row->header_end, NULL);
 			assert(row->bodycnt == 1);
 			const char *body = row->body[0].iov_base;
 			const char *body_end = body + row->body[0].iov_len;
 			func_adapter_push_msgpack(trigger, &ctx, body,
-						  body_end);
+						  body_end, NULL);
```
drewdzzz, line 821: We could add translation to the space recovery triggers (as a separate commit).

src/box/lua/iproto.lua
```diff
@@ -0,0 +1,28 @@
...
+-- Passing nil as the callback resets the corresponding request handler.
+box.iproto.override = function(request_type, callback)
+    local trigger = require('trigger')
+
+    if request_type == nil then
+        box.error(box.error.PROC_LUA,
+                  'Usage: box.iproto.override(request_type, callback)')
+    end
+    if type(request_type) ~= 'number' then
+        box.error(box.error.PROC_LUA, 'expected uint64_t as 1 argument')
```
drewdzzz, line 12: There is no `uint64_t` in Lua.

src/box/lua/iproto.lua
```diff
@@ -0,0 +1,28 @@
...
+    end
+    if callback ~= nil and type(callback) ~= 'function' then
+        box.error(box.error.PROC_LUA,
+                  ("bad argument #2 to 'override' (function expected, " ..
+                   "got %s)"):format(type(callback)))
+    end
+
+    local event_name = ('box.iproto.override[%d]'):format(request_type)
+    local trigger_name = '__tarantool__internal__iproto__override__trigger__'
```
drewdzzz, line 21: I think it's OK to set a handler with `box.iproto.override` and then delete it with trigger registry.
So let's call a trigger with more readable name (`box.iproto.override.lua.user_handler`, for example).

drewdzzz (changes requested):

ochaplashkin (approved):

changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
```diff
@@ -0,0 +1,10 @@
+## feature/core
+
+* Now it is possible to set IPROTO request handler callbacks by using the new
+  Lua module `trigger`, e.g.:
```
xuniq, line 4:
```suggestion
  Lua module `trigger`, for example:
```

changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
```diff
@@ -0,0 +1,10 @@
...
+  ```lua
+  local trigger = require('trigger')
+  trigger.set('box.iproto.override.select', 'my_select', my_select_handler)
+  ```
+  This new method works before the initial `box.cfg{}` call, also it allows to
+  set multiple handlers for a single request type (gh-8138).
```
xuniq, line 10:
```suggestion
  The method works before the initial `box.cfg{}` call. Also, the method allows settinng multiple handlers for a single request type (gh-8138).
```

xuniq (approved):

src/box/iproto.cc
```diff
@@ -2660,8 +2728,23 @@ tx_process_override(struct cmsg *m)
 		body_end = body + msg->header.body[0].iov_len;
 	}
 
+	/*
+	 * If we took the `override_route', there must exist either request
+	 * type-specific or unknown request type handler. Their availability
+	 * is checked by the IPROTO thread.
+	 */
```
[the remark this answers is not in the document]
Gumix: Yes. If I'm not wrong, `iproto_thread->req_handlers` and `tx_req_handlers` are always in sync, because `iproto_do_cfg()` is implemented via synchronous call over cbus. I'd also ask @CuriousGeorgiy to verify this.

src/box/iproto.cc
```diff
@@ -2631,22 +2638,83 @@ tx_process_replication(struct cmsg *m)
...
 /**
- * Process a request using the overridden handler (or the unknown request
- * handler as a last resort). We assume that the IPROTO thread checked the
- * handler availability.
+ * Allocates a new `iproto_req_handlers' for the given `req_type' and inserts it
+ * into the `tx_req_handlers' table. There must be no previous entries in the
+ * table for this key.
+ */
+static struct iproto_req_handlers *
+iproto_req_handlers_init(uint32_t req_type)
```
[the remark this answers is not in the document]
Gumix: Fixed.

src/box/iproto.cc
```diff
@@ -2631,22 +2638,83 @@ tx_process_replication(struct cmsg *m)
...
+}
+
+/**
+ * Destroys all handlers and deallocates the `handlers' structure.
+ * If `req_type' is a valid request type, the handlers are also deleted from the
+ * `tx_req_handlers' table.
  */
 static void
-tx_process_override(struct cmsg *m)
+iproto_req_handlers_free(struct iproto_req_handlers *handlers,
```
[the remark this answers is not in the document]
Gumix: Fixed.

src/box/iproto.cc
```diff
@@ -2631,22 +2638,83 @@ tx_process_replication(struct cmsg *m)
...
+		return NULL;
 	struct mh_i32ptr_node_t *node = mh_i32ptr_node(tx_req_handlers, k);
-	struct iproto_req_handler *handler =
-		(struct iproto_req_handler *)node->val;
+	return (struct iproto_req_handlers *)node->val;
+}
+
+/**
+ * Returns `true' if a handler in `handlers' is set.
+ */
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/iproto.cc
```diff
@@ -41,6 +41,7 @@
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/iproto.cc
Gumix: Done.

test/unit/lua_func_adapter.c
Gumix: Done.

src/box/lua/iproto.lua
Gumix: Fixed.

src/box/lua/iproto.lua
Gumix: Done.

changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
Gumix: Done.

changelogs/unreleased/gh-8138-iproto-override-via-event-triggers.md
Gumix: Done.

src/box/iproto.cc
```diff
@@ -3778,39 +3849,30 @@ iproto_override(uint32_t req_type, iproto_handler_t cb,
...
+	} else if (handlers != NULL) {
+		handlers->c.cb = NULL;
+		handlers->c.destroy = NULL;
+		handlers->c.ctx = NULL;
 	}
-	iproto_cfg_override(req_type, true);
+
+	bool new_is_set = iproto_req_handler_is_set(handlers);
+	if (!new_is_set && handlers != NULL)
+		iproto_req_handlers_delete(handlers, req_type);
```
CuriousGeorgiy, line 4170: Looks like a bug: you can delete a handler from TX only after notifying IPROTO. Otherwise there can be inflight requests that arrive at the override route only to see there is no request handler set (which is unexpected).

src/box/iproto.cc
```diff
@@ -3519,6 +3538,48 @@ iproto_override_event_init(struct event *event, void *arg)
...
+		if (handlers == NULL)
+			handlers = iproto_req_handlers_new(type);
+		iproto_req_handlers_set_event(handlers, event, is_by_id);
+	} else {
+		iproto_req_handlers_del_event(handlers, is_by_id);
+	}
+
+	bool new_is_set = iproto_req_handler_is_set(handlers);
+	if (!new_is_set)
+		iproto_req_handlers_delete(handlers, type);
```
CuriousGeorgiy, line 3574: Looks like a bug: you can delete a handler from TX only after notifying IPROTO. Otherwise there can be inflight requests that arrive at the override route only to see there is no request handler set (which is unexpected).

src/box/lua/iproto.lua
```diff
@@ -0,0 +1,30 @@
...
+    end
+    if type(request_type) ~= 'number' then
+        box.error(box.error.PROC_LUA,
+                  ("bad argument #1 to 'override' (number expected, " ..
+                   "got %s)"):format(type(callback)))
+    end
+    if callback ~= nil and type(callback) ~= 'function' then
+        box.error(box.error.PROC_LUA,
+                  ("bad argument #2 to 'override' (function expected, " ..
+                   "got %s)"):format(type(callback)))
```
CuriousGeorgiy, line 19: I am not an expert on the trigger subsystem, but can't it also be a callable table now? @drewdzzz

src/box/lua/iproto.lua
```diff
@@ -0,0 +1,30 @@
...
+    local trigger = require('trigger')
+
+    if request_type == nil then
+        box.error(box.error.PROC_LUA,
+                  'Usage: box.iproto.override(request_type, callback)')
+    end
+    if type(request_type) ~= 'number' then
+        box.error(box.error.PROC_LUA,
+                  ("bad argument #1 to 'override' (number expected, " ..
+                   "got %s)"):format(type(callback)))
```
CuriousGeorgiy, line 14: Can't it also be a string now?

src/box/lua/iproto.lua
```diff
@@ -0,0 +1,30 @@
...
+    end
+    if callback ~= nil and type(callback) ~= 'function' then
+        box.error(box.error.PROC_LUA,
+                  ("bad argument #2 to 'override' (function expected, " ..
+                   "got %s)"):format(type(callback)))
+    end
+
+    local event_name = ('box.iproto.override[%d]'):format(request_type)
+    local trigger_name = 'box.iproto.override.lua.user_handler'
```
CuriousGeorgiy, line 23: What if a user incidentally has a trigger with the same name?

CuriousGeorgiy (changes requested):

src/box/iproto.cc
```diff
@@ -275,15 +275,22 @@ unsigned iproto_readahead = 16320;
...
+ *  1. Lua or C handler, set by `iproto_override()'.
  */
-struct iproto_req_handler {
-	/** Request handler callback. */
-	iproto_handler_t cb;
-	/** Request handler destructor, can be NULL. */
-	iproto_handler_destroy_t destroy;
-	/** Context passed to request handler callback and destructor. */
-	void *ctx;
+struct iproto_req_handlers {
```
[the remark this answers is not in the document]
Gumix: I don't like `registry`, because it is a single entry in the hash table, not a registry.

src/box/iproto.cc
CuriousGeorgiy: BTW, please also refactor the common logic out into a separate helper.

src/box/space.c
Gumix: Done.

src/box/lua/func_adapter.c
```diff
@@ -141,10 +141,10 @@ func_adapter_lua_push_null(struct func_adapter_ctx *base)
 
 static void
 func_adapter_lua_push_msgpack(struct func_adapter_ctx *base, const char *data,
-			      const char *data_end)
+			      const char *data_end, void *mp_ctx)
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/space.c
```diff
@@ -813,12 +813,12 @@ space_run_replace_triggers(struct event *event, struct txn *txn,
 			struct xrow_header *row = stmt->row;
 			assert(row != NULL && row->header != NULL);
 			func_adapter_push_msgpack(trigger, &ctx, row->header,
-						  row->header_end);
+						  row->header_end, NULL);
```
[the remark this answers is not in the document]
Gumix: Done.

src/lib/core/func_adapter.h
```diff
@@ -81,7 +81,7 @@ struct func_adapter_vtab {
 	 * Pushes MsgPack argument.
 	 */
 	void (*push_msgpack)(struct func_adapter_ctx *ctx, const char *data,
-			     const char *data_end);
+			     const char *data_end, void *mp_ctx);
```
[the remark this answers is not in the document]
Gumix: Done.

src/lib/core/func_adapter.h
```diff
@@ -226,9 +226,9 @@ func_adapter_push_null(struct func_adapter *func, struct func_adapter_ctx *ctx)
 static inline void
 func_adapter_push_msgpack(struct func_adapter *func,
 			  struct func_adapter_ctx *ctx,
-			  const char *data, const char *data_end)
+			  const char *data, const char *data_end, void *mp_ctx)
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/iproto.cc
```diff
@@ -275,15 +275,22 @@ unsigned iproto_readahead = 16320;
 static int iproto_msg_max = IPROTO_MSG_MAX_MIN;
 
 /**
- * Request handler meta information.
+ * Request handlers meta information. The IPROTO request of each type can be
+ * overridden by the following types of handlers (listed in priority order):
+ *  1. Lua or C handler, set by `iproto_override()'.
```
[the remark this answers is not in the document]
Gumix: Done.

src/box/iproto.cc
```diff
@@ -2631,22 +2638,83 @@ tx_process_replication(struct cmsg *m)
...
-	struct iproto_msg *msg = tx_accept_msg(m);
-	mh_int_t k = mh_i32ptr_find(tx_req_handlers, msg->header.type, NULL);
-	if (k == mh_end(tx_req_handlers)) {
-		k = mh_i32ptr_find(tx_req_handlers, IPROTO_UNKNOWN, NULL);
+	if (handlers->c.destroy != NULL)
+		handlers->c.destroy(handlers->c.ctx);
+	TRASH(handlers);
+	free(handlers);
+
+	if (req_type != iproto_type_MAX) {
```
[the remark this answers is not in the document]
Gumix: Split into `iproto_req_handlers_delete` and `mh_req_handlers_del`.
Same for `iproto_req_handlers_new` and `mh_req_handlers_put`.

src/box/iproto.cc
```diff
@@ -2704,7 +2746,88 @@ iproto_req_handler_is_set(struct iproto_req_handlers *handlers)
...
+		const char *type_name = iproto_type_name(i);
+		if (type_name == NULL)
+			continue;
+		char *type_name_lower = strtolowerdup(type_name);
+		if (strcmp(type_name_lower, name) == 0) {
+			type = (enum iproto_type)i;
+			free(type_name_lower);
+			break;
+		}
+		free(type_name_lower);
```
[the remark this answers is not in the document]
Gumix: > We will now have 2 places where we do this lowercase conversion (see also src/box/lua/iproto.c).

In `src/box/lua/iproto.c` the `iproto_key_strs` is lowered, not `iproto_type_strs`.

> How about creating an array of lowercase string constants one time during iproto initialization?

Done.

> That way you also use strindex from src/lib/core/utils.c here.

`strindex()` ignores the case of the characters, but I need case-sensitive comparison here, because from the trigger-registry point of view the events "box.iproto.override.ping" and "box.iproto.override.PING" are totally unique. For this reason only lowercase names are allowed in `get_iproto_type_by_name()` to avoid any further inconsistency.

src/box/iproto.cc
Gumix: Thanks, fixed.

src/box/iproto.cc
Gumix: Thanks, fixed.

> BTW, please also refactor the common logic out into a separate helper.

I tried, but couldn't make the code more beautiful than it is now.

src/box/lua/iproto.lua
Gumix: AFAIK, it's hard to support callable tables here. Let's allow only functions here. Callable objects can be set via `trigger.set()`.

src/box/lua/iproto.lua
Gumix: Done.

src/box/lua/iproto.lua
Gumix: The new trigger will override the old one. That's why I called it "__tarantool__internal__iproto__override__trigger__" in the first PR iteration.

src/box/iproto.cc
```diff
@@ -2701,22 +2722,211 @@ tx_process_replication(struct cmsg *m)
...
+	enum iproto_handler_status rc = IPROTO_HANDLER_FALLBACK;
+	const char *name = NULL;
+	struct func_adapter *trigger = NULL;
+	struct func_adapter_ctx ctx;
+	struct event_trigger_iterator it;
+	event_trigger_iterator_create(&it, event);
+
+	while (event_trigger_iterator_next(&it, &trigger, &name)) {
+		struct mp_ctx mp_ctx_header, mp_ctx_body;
+		mp_ctx_create_default(&mp_ctx_header, iproto_key_translation);
```
Gumix, line 2891: I'm not sure about this part. Is it necessary to call `mp_ctx_create_default()` on every iteration?

src/box/lua/iproto.lua
drewdzzz: We cannot check if an object is callable from Lua (for example, AFAIK, there is no way to check if `cdata` is callable from Lua). But if we really want to accept callable objects, we could call `trigger.set` with `pcall` or rewrite this function in C. But as for me, it's OK to accept only functions here.

src/box/lua/iproto.lua
drewdzzz: In the triggers documentation (which doesn't exist yet) we will write that `tarantool` and `box` root namespaces are reserved for our purposes.

src/box/iproto.cc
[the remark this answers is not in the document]
drewdzzz: So, if `iproto` -> `tx` messages are asynchronous, there is a chance a message was sent, but the handlers were deleted. Hence, this comment and an assertion below are not fair.

src/box/iproto.cc
CuriousGeorgiy: I mean these two snippets are very similar, I suggest refactoring them into a separate function:

https://github.com/Gumix/tarantool/blob/f5953480a14118dd06efb8f5668b398afc5f4637/src/box/iproto.cc#L3653-L3660

https://github.com/Gumix/tarantool/blob/f5953480a14118dd06efb8f5668b398afc5f4637/src/box/iproto.cc#L4322-L4329

Feel free to ignore.

src/box/iproto.cc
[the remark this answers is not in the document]
drewdzzz: Discussed f2f, it's OK.

CuriousGeorgiy (approved):

CuriousGeorgiy (approved):

src/box/iproto.cc
drewdzzz: Also, can we use one `mp_ctx` for both packets?

src/box/iproto.cc
```diff
@@ -3377,6 +3510,94 @@ iproto_thread_init(struct iproto_thread *iproto_thread)
...
+				 */
+				if (*endptr != ']' || endptr[1] != 0)
+					type = iproto_type_MAX;
+			}
+		}
+	}
+
+	if (type == iproto_type_MAX) {
+		say_crit("The event `%s' is in IPROTO override namespace, but "
+			 "`%s' is not a valid request type", name, req_name);
```
drewdzzz, line 3576: It's not correct to use `req_name` here since it was advanced during the name parsing. Let's use `name + prefix_name`.

src/box/iproto.cc
```diff
@@ -2774,6 +2792,30 @@ mh_req_handlers_del(uint32_t req_type)
...
+iproto_req_handlers_set_event(struct iproto_req_handlers *handlers,
+			      struct event *event, bool is_by_id)
+{
+	assert(handlers != NULL);
+	assert(event != NULL);
+
+	event_ref(event);
+
+	if (is_by_id) {
+		if (handlers->event_by_id != NULL)
```
drewdzzz, line 2809: Why do you need to replace event with another one?

drewdzzz (commented):
Please address my comments.
Also:
1. Trigger documentation is already written, so you can write doc for this feature.
2. Shouldn't we mark `box.iproto.override` change as breaking?

src/box/iproto.cc
CuriousGeorgiy: Yes.

src/box/iproto.cc
Gumix: Moved to the new function `iproto_override_finish()`.

src/box/iproto.cc
Gumix: Actually it doesn't work, because `func_adapter_push_msgpack_with_ctx` calls `mp_ctx_move`.

src/box/space.c
```diff
@@ -887,15 +889,21 @@ space_run_replace_triggers(struct event *event, struct txn *txn,
 				       iproto_type_name(stmt->type));
 		/* Pass xrow header and body to recovery triggers. */
 		if (stmt->space->run_recovery_triggers) {
+			struct mp_ctx mp_ctx_header, mp_ctx_body;
+			mp_ctx_create_default(&mp_ctx_header,
+					      iproto_key_translation);
+			mp_ctx_create_default(&mp_ctx_body,
+					      iproto_key_translation);
```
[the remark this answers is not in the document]
Gumix: It doesn't work, because `func_adapter_push_msgpack_with_ctx` calls `mp_ctx_move`.

src/box/iproto.cc
Gumix: Changed to:
```diff
@@ -3650,11 +3654,13 @@ get_iproto_type_from_event_name(const char *name, bool *is_by_id)
                return iproto_type_MAX;

        const char *req_name = name + prefix_len;
+       const char *req_name_err = req_name;
        uint32_t type = iproto_type_MAX;

        if (*req_name == '.') {
                /* Skip the dot. */
                req_name++;
+               req_name_err = req_name;
                type = get_iproto_type_by_name(req_name);
        } else if (*req_name == '[') {
                /* Skip open bracket. */
@@ -3674,11 +3680,14 @@ get_iproto_type_from_event_name(const char *name, bool *is_by_id)
                                        type = iproto_type_MAX;
                        }
                }
+       } else {
+               return iproto_type_MAX;
        }

        if (type == iproto_type_MAX) {
                say_crit("The event `%s' is in IPROTO override namespace, but "
-                        "`%s' is not a valid request type", name, req_name);
+                        "`%s' is not a valid request type",
+                        name, req_name_err);
                return iproto_type_MAX;
        }
        bool is_unsupported_request = type == IPROTO_JOIN ||
```

Gumix:
> Trigger documentation is already written, so you can write doc for this feature.

Added TarantoolBot request.

> Shouldn't we mark box.iproto.override change as breaking?

Added:
```diff
diff --git a/changelogs/unreleased/gh-8138-switch-box-iproto-override-to-triggers.md b/changelogs/unreleased/gh-8138-switch-box-iproto-override-to-triggers.md
new file mode 100644
index 000000000..8c1fc1d1e
--- /dev/null
+++ b/changelogs/unreleased/gh-8138-switch-box-iproto-override-to-triggers.md
@@ -0,0 +1,5 @@
+## feature/core
+
+* `box.iproto.override()` was switched to the universal trigger registry. As a
+  side effect, now this function does not raise an error when a wrong request
+  type is passed. All such errors are logged with `CRITICAL` level (gh-8138).
```

src/box/iproto.cc
```diff
@@ -3482,6 +3619,122 @@ iproto_thread_init(struct iproto_thread *iproto_thread)
...
+ * `is_by_id' set to True if the request is overridden by id, False if by name.
+ */
+static uint32_t
+get_iproto_type_from_event_name(const char *name, bool *is_by_id)
+{
+	*is_by_id = false;
+	const char *prefix = "box.iproto.override";
+	const size_t prefix_len = strlen(prefix);
+	if (strncmp(name, prefix, prefix_len) != 0)
+		return iproto_type_MAX;
```
drewdzzz, line 3654: I'm not sure it's a good decision because in future we can have a request type with value 129 (current `iproto_type_MAX` value).
I think we shouldn't have any reserved values here - let's return bool indicating if event name was successfully parsed and return value with an output argument.
Also, let's test this case if we can.

drewdzzz (commented):
Sorry for such an "iterative" review.
Generally, it's OK, just one last detail in the comment above.

src/box/iproto.cc
Gumix: Done.

> Also, let's test this case if we can.

Added a testcase with hardcoded value 129.

drewdzzz (approved):

Issue #8204, opened 2023-01-24
Optimize select with offset option


alyapunov:
[PM link](https://www.notion.so/tarantool/offset-n-count-key-log-n-4aba908b48674123892474957420c85f)

Now the cost of `select({key}, offset)` in TREE index with N tuples is `O(log(N) + offset)`. That's a big problem for those users who use huge offsets. Actually it seems that it can simply optimized to O(log(N)) with a tiny cost - just store the total size of subtree in every node of bps tree.

On this way we could also optimize cost of `count{key}` to O(count).

R-omk:
It may also optimize  count between A,B , like redis ZCOUNT command.

but I would prefer it to be an index option or even other type of index

alyapunov:
> It may also optimize count between A,B , like redis ZCOUNT command.
> 
> but I would prefer it to be an index option or even other type of index

1. Tarantool doesn't have methods neither for selecting tuples between A and B nor for counting them, while this issue does not introduce new API, just optimization of it. But if we implement the proposal then you will able to get the desired count with `count(A, {iterator='ge'}) - count(B, {iterator='gt'})`.
2. Why would you prefer an option? An option (or new index) is an additional pain for support. On the other hand the counter is veery cheap in terms of memory consumption and rather cheap in terms of calculation. The only problem I see that in case of presence of a read view the copy-on-write will be used more aggressively.

R-omk:
1. if the count function becomes optimized that should work good.

2. I'm just worried about performance due to the presence of an unpopular feature. To avoid the "bottleneck" effect. But if you think that there are few negative effects, then ok.

viciious:
This would be a neat feature indeed!

PR #9478, opened 2023-12-13
bps: add support for logarithmic offset

mkostoevr:
bps: add 2-way support for logarithmic offsets

The current tree does not allow to find offset of an element or create
an iterator to an element based on its offset. This patch is meant to
fix this by expanding the data structure with additional information
and introducing methods using it: subtree cardinalities.

A subtree cardinality is the amount of elements in it. For example,
cardinality of a leaf block is count of elements in it (effectively
it equals to leaf.header.size), cardinality of an inner block is the
sum of cardinalities of its chlidren.

The patch includes two chosable ways to store this information:
`BPS_INNER_CARD` and `BPS_INNER_CHILD_CARDS`.

The first implementation sores block cardinality in each inner block.
This implementation has minimal memory overhead (it just introduces
a new 64-bit field in `struct bps_inner`), but calculation of offsets
is not that fast, since in order to find an offset of a particular
child of an inner node we have to look into each of its children
prior to the looking one.

The second one sores an array of children cardinalities in inner
blocks. The memory overhead of this implementation is visible since
it significantly decreases the children capacity of inner blocks. The
max count in inner block is decreased from 42 to 25 for tree of 8-byte
elements with 512-byte blocks and from 25 to 18 for tree of 16-byte
elements with 512-byte blocks. Offset calcluations are faster though.

It's possible (though impractical) to enable both solutions, the tree
will use the best ways to perform offset-based tasks, but will have to
maintain both children cardinality array and inner own cardinalities.

Along with the theoretical support this patch introduces a bunch of
functions using it:
- `iterator_at(t, offset)`: gives an iterator to an element of a tree
  or tree view by its offset;
- `find_get_offset(t, key, offset_ptr)`: the same as `find` but also
  provides to the user the offset of the found element in the output
  parameter;
- `[lower|upper]_bound[_elem]_get_offset(t, key, exact, offset_ptr)`:
  the same as upper/lower bound functions but provide to the user the
  offset to the found position (end of the tree included).
- `insert_get_offset(t, new_elem, replaced, offset_ptr)`: the same as
  `insert`, but also provides the offset to the inserted element.
- `delete_get_offset(t, elem, offset_ptr)`: same as `delete`, but also
  returns offset to the deleted element prior to the deletion in the
  output parameter.

Another new function introduced is bps_tree_view_debug_check(t). This
function is similar to the bps_tree_debug_check(t), but is applicable
to tree views. It's used to adopt default tree view tests to the new
tree variations.

Each new implementation is tested by old tree tests (these now support
several tree variations selected with a C definition, the definitions
are specified in the test/unit/CMakeLists.txt).

Since creation of new tests with .result files is prohibited by the
checkpatch, the new flavor tests from the test/unit/bps_tree.cc are
made TAP-compatible. Given their output is not checked, the printing
test from the suite had been disabled for them.

New offset API-related test introduced (this one tests both of two
tree variations - BPS_INNER_CARD and BPS_INNER_CHILD_CARDS).

Part of #8204

NO_DOC=internal
NO_CHANGELOG=internal


ochaplashkin (approved):

alyapunov (changes requested):
Great!
I think we should squash all three commits into two:
1. refactoring of debug functions
2. the improvement and its test
And please consider rename `power` -> `cardinality`. If it is required to make this name shorter, we can use `cardin` or even `card`

src/lib/salad/bps_tree.h
```diff
@@ -314,6 +314,14 @@
  * #define BPS_BLOCK_LINEAR_SEARCH
  */
 
+/**
+ * A switch to make the tree store the cardinality of each of its
+ * child blocks in an array. A block cardinality is the amount of
+ * elements in its underlying leaf blocks (or a sum of its child
+ * cardinalities if the children are inner).
+ * #define BPS_BLOCK_CHILD_CARDS_ARRAY
```
locker, line 322: Abundance of macros makes the code ugly. Let's measure the degradation introduced by this change, both in CPU and memory consumption. If it proves to be insignificant, better enable the feature unconditionally.

src/lib/salad/bps_tree.h
```diff
@@ -1111,6 +1127,62 @@ bps_tree_debug_check_internal_functions(bool assertme);
...
+#define BPS_TREE_CARD_UP_INNER(...) bps_tree_card_up_inner_impl(__VA_ARGS__)
+#else
+#define BPS_TREE_CARD_UP_LEAF(...)
+#define BPS_TREE_CARD_UP_INNER(...)
+#endif
+
+/**
+ * The config-dependent block information.
+ */
+struct bps_block_info {
```
locker, line 1178: Passing this struct around looks ugly. Please consider storing own cardinality in each block (along with the array of child cardinalities). The increase in memory consumption should be negligible but then you wouldn't need to pass the cardinality around and introduce the `block_info` struct.

test/unit/bps_tree_generic.cc
```diff
@@ -4,6 +4,10 @@
 #include <inttypes.h>
 #include <time.h>
 
+#ifdef BPS_BLOCK_CHILD_CARDS_ARRAY
+#define UNIT_TAP_COMPATIBLE 1
+#endif
```
locker, line 9: Looks like a good change, even if without the rest of the patch. Please consider patching the tests in a separate commit.

test/unit/bps_tree_with_child_cards.cc
```diff
@@ -0,0 +1,14 @@
...
+ * Some branches lead to dummy rebalancing (moving data of zero size) when
+ * the block size is small because of integer arithmetic on reballancing.
+ * This raises assertions, because some data movement routines are designed
+ * to only move non-zero amount of data. Let's make the block size greater
+ * to prevent this.
+ */
+#define SMALL_BLOCK_SIZE 256
+#define COMPARE_WITH_SPTREE_CHECK_BRANCHES_ELEM_LIMIT 2048
+#define TEST_COUNT_IN_LEAF 30
+#define TEST_COUNT_IN_INNER 12
```
locker, line 13: Let's introduce the new public API function that would return the number of BPS tree elements less than the specified key to test the feature. It can be useful anyway.

src/lib/salad/bps_tree.h
mkostoevr: There're three ways to store cardinalities implemented:
1. Using an Array of Children Cardinalities (AoCC), which was the first approach implemented. We store chlidren cardinalities in `child_cards` arrays of each inner node.
2. Using Inner Block Cardinalities (IBC), which makes the tree store own cardinalities in inner blocks.
3. A hybrid approach using both `child_cards` array and inner block own cardinality, technically the one suggested in the comment below.

So here're benchmarking results.

<details>

<summary>Raw BPS tree benchmarks (perf/bps_tree.cc)...</summary>

| Benchmark               | Base   | AoCC                          | IBC                               | Both                          |
| ----------------------- | ------ | ----------------------------- | --------------------------------- | ----------------------------- |
| find_rand/1M            | 288ns  | 306ns ($\color{red}+6.25％$)   | 290ns ($\color{RedOrange}+0.69％$) | 306ns ($\color{red}+6.25％$)   |
| find_rand/10M           | 528ns  | 562ns ($\color{red}+6.44％$)   | 511ns ($\color{green}-3.22％$)     | 527ns ($\color{green}-0.19％$) |
| find_rand/100M          | 747ns  | 766ns ($\color{red}+2.54％$)   | 748ns ($\color{RedOrange}+0.13％$) | 764ns ($\color{red}+2.28％$)   |
| find_rand/500M          | 978ns  | 1011ns ($\color{red}+3.37％$)  | 982ns ($\color{RedOrange}+0.41％$) | 1023ns ($\color{red}+4.60％$)  |
| delete_insert_rand/1M   | 569ns  | 711ns ($\color{red}+24.96％$)  | 657ns ($\color{red}+15.47％$)      | 704ns ($\color{red}+23.73％$)  |
| delete_insert_rand/10M  | 887ns  | 1017ns ($\color{red}+14.66％$) | 1006ns ($\color{red}+13.42％$)     | 1012ns ($\color{red}+14.09％$) |
| delete_insert_rand/100M | 1176ns | 1360ns ($\color{red}+15.65％$) | 1303ns ($\color{red}+10.80％$)     | 1355ns ($\color{red}+15.22％$) |
| delete_insert_rand/500M | 1439ns | 1686ns ($\color{red}+17.16％$) | 1593ns ($\color{red}+10.70％$)     | 1689ns ($\color{red}+17.37％$) |
| insert_rand/10M         | 490ns  | 537ns ($\color{red}+9.59％$)   | 523ns ($\color{red}+6.73％$)       | 532ns ($\color{red}+8.57％$)   |
| insert_rand/100M        | 771ns  | 854ns ($\color{red}+10.77％$)  | 812ns ($\color{red}+5.32％$)       | 849ns ($\color{red}+10.12％$)  |
| insert_rand/500M        | 1045ns | 1141ns ($\color{red}+9.19％$)  | 1098ns ($\color{red}+5.07％$)      | 1138ns ($\color{red}+8.90％$)  |
| delete_rand/10M         | 555ns  | 610ns ($\color{red}+9.91％$)   | 592ns ($\color{red}+6.67％$)       | 605ns ($\color{red}+9.01％$)   |
| delete_rand/100M        | 819ns  | 898ns ($\color{red}+9.65％$)   | 862ns ($\color{red}+5.25％$)       | 887ns ($\color{red}+8.30％$)   |
| delete_rand/500M        | 1061ns | 1166ns ($\color{red}+9.90％$)  | 1119ns ($\color{red}+5.47％$)      | 1180ns ($\color{red}+11.22％$) |

</details>

<details>

<summary>Measuring RPS using raw sockets, batch size is 1000...</summary>

| Benchmark | Base    | AoCC                                | IBC                             | Both                            |
| --------- | ------- | ----------------------------------- | ------------------------------- | ------------------------------- |
| insert    | 457krps | 440krps ($\color{red}-3.63％$)       | 448krps ($\color{red}-2.02％$)   | 427krps ($\color{red}-6.59％$)   |
| replace   | 451krps | 451krps ($\color{RedOrange}-0.02％$) | 451krps ($\color{green}+0.05％$) | 451krps ($\color{green}+0.16％$) |
| select    | 760krps | 723krps ($\color{red}-4.83％$)       | 737krps ($\color{red}-3.01％$)   | 736krps ($\color{red}-3.22％$)   |
| delete    | 424krps | 415krps ($\color{red}-2.08％$)       | 415krps ($\color{red}-2.00％$)   | 403krps ($\color{red}-4.99％$)   |

</details>

The memory overhead of the IBC solution is negligible. The overhead of the AoCC is a bit more complex, but practically I've got about 2%.

<details>

<summary>Measuring memory consumption (512B block, 16K extent size)</summary>

| Tree configuration | Base       | AoCC                             | IBC                                   |
| ------------------ | ---------- | -------------------------------- | ------------------------------------- |
| 8-byte elem x1M    | 11024KB    | 11264KB ($\color{red}+2.18％$)    | 11040KB ($\color{RedOrange}+0.15％$)   |
| 8-byte elem x10M   | 109904KB   | 112384KB ($\color{red}+2.26％$)   | 110016KB ($\color{RedOrange}+0.10％$)  |
| 8-byte elem x100M  | 1098720KB  | 1123520KB ($\color{red}+2.26％$)  | 1099856KB ($\color{RedOrange}+0.10％$) |
| 8-byte elem x500M  | 5493440KB  | 5617440KB ($\color{red}+2.26％$)  | 5499168KB ($\color{RedOrange}+0.10％$) |
| 16-byte elem x1M   | 22032KB    | 22480KB ($\color{red}+2.03％$)    | 22032KB (-)                           |
| 16-byte elem x10M  | 220048KB   | 224496KB ($\color{red}+2.02％$)   | 220048KB (-)                          |
| 16-byte elem x100M | 2200192KB  | 2244720KB ($\color{red}+2.02％$)  | 2200192KB (-)                         |
| 16-byte elem x500M | 11000768KB | 11223456KB ($\color{red}+2.02％$) | 11000768KB (-)                        |

<details>

I would keep this feature optional but if it'll be decided to apply it to default tree too, it seems the least affecting the current tree option is IBC (only storing own cardinalities in inner blocks).

src/lib/salad/bps_tree.h
mkostoevr: The thing is, we won't need the children cardinality array if we introduce the inner block own cardinality. It will in fact have a better performance. See the benchmarks above.

test/unit/bps_tree_generic.cc
mkostoevr: Moved this into a separated commit.

test/unit/bps_tree_with_child_cards.cc
mkostoevr: Introduced `find_get_offset` and `[lower|upper]_bound[_elem]_get_offset` families of functions.

src/lib/salad/bps_tree.h
mkostoevr: Discussed offline: it's decided to keep both options but use the default tree with no changes for regular tree index to prevent the regressions.

test/unit/CMakeLists.txt
```diff
@@ -136,9 +136,17 @@ create_unit_test(PREFIX interval
...
+                 SOURCES bps_tree_default.cc
+                 LIBRARIES unit small misc
+)
+create_unit_test(PREFIX bps_tree_inner_card
+                 SOURCES bps_tree_inner_card.cc
+                 LIBRARIES unit small misc
+)
+create_unit_test(PREFIX bps_tree_inner_child_cards
+                 SOURCES bps_tree_inner_child_cards.cc
+                 LIBRARIES unit small misc
```
locker, line 149: What about `bps_tree_iterator` and `bps_tree_view` tests? It looks like the view API is untested.

perf/bps_tree.cc
```diff
@@ -4,13 +4,19 @@
 #include <inttypes.h>
 #include <time.h>
 #include <random>
+#include <memory>
 
 #include <benchmark/benchmark.h>
 #include <trivia/util.h>
 
+/* A config to enable height-based and first/last/inc/dec benchmarks. */
+#define ENABLE_EXTRA_BENCHMARKS 0
```
locker, line 13: I'd rather you reduced the data set size instead of disabling some tests. Even better, make the data size dynamically configurable and set it to a smaller value in the `perf_micro` CI workflow.

alyapunov (approved):

perf/bps_tree.cc
mkostoevr: Reduced the data set since the perf pipeline setup is not the primary goal in this patch.

test/unit/CMakeLists.txt
mkostoevr: Made these tests test the new trees too. All the ols BPS tree tests support the tree flavor specification via define. The defines are specified in the CMakeFile.

perf/bps_tree.cc
```diff
@@ -494,22 +581,26 @@ test_insert_rand(benchmark::State &state, size_t count)
 	test_insert<tree>(state, count, RandomKey(count));
 }
 
-generate_benchmarks_size_iterations(insert_rand, 10000000);
+generate_benchmarks_size_iterations(insert_rand, 1000000);
```
locker, line 584: Please mention in the commit message that you reduce the data set size or, even better, do it in a separate patch with a proper explanation.

mkostoevr: Created a separated commit for that.

test/unit/bps_tree.cc
```diff
@@ -164,6 +196,52 @@ extent_free(void *ctx, void *extent)
...
+static int64_t *
+arr_seq_shuffled(size_t size)
+{
+	int64_t *arr = arr_seq(size);
+	shuffle(arr, size);
+	return arr;
+}
+
+static void
+debug_check(test &tree)
```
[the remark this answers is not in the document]
mkostoevr: Make this and helper functions from bps_tree_view.c and bps_tree_offset_api.cc macros.

test/unit/bps_tree.cc
```diff
@@ -1,16 +1,47 @@
...
-#include <inttypes.h>
-#include <time.h>
+#include <cstdint>
+#include <cstdio>
+#include <cinttypes>
+#include <ctime>
+
+#ifndef TEST_DEFAULT
+#define UNIT_TAP_COMPATIBLE 1
+#endif
```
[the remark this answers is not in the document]
mkostoevr: Done. The test that checked the tree print can't be universally checked, so it just checks if the function does not segfault now.

Issue #9145, opened 2023-09-14
Backporting LuaJIT upstream commits part 6


Buristan:
This ticket is an umbrella for all commits we planning to back-port in the next Q.
This list includes commits excluded from #8825 and may be updated later with some new.

GC64:
* [ ] [ff1e72ac](https://github.com/LuaJIT/LuaJIT/commit/ff1e72ac) LJ_GC64: Always snapshot functions for non-base frames. (**in progress**, @ligurio, depends on "Improve assertions." [8ae5170c](https://github.com/LuaJIT/LuaJIT/commit/8ae5170c))
* [ ] [91bc6b8a](https://github.com/LuaJIT/LuaJIT/commit/91bc6b8a) LJ_GC64: Fix HREFK optimization. (**in progress** @ligurio)

general + x86/x64:
* [ ] [1d75cd4d](https://github.com/LuaJIT/LuaJIT/commit/1d75cd4d) Avoid assertion in case of stack overflow from stitched trace.
* [ ] [1c279127](https://github.com/LuaJIT/LuaJIT/commit/1c279127) Print errors from __gc finalizers instead of rethrowing them.
* [ ] [224129a8](https://github.com/LuaJIT/LuaJIT/commit/224129a8) Fix last commit.
* [ ] [8135de2a](https://github.com/LuaJIT/LuaJIT/commit/8135de2a) Improve error reporting on stack overflow.
* [ ] [62903bac](https://github.com/LuaJIT/LuaJIT/commit/62903bac) Fix C file generation in jit.bcsave.
* [ ] [384d6d56](https://github.com/LuaJIT/LuaJIT/commit/384d6d56) Fix Clang build.
* [ ] [75ee3a61](https://github.com/LuaJIT/LuaJIT/commit/75ee3a61) Prevent compile of __concat with tailcall to fast function.
* [ ] [ae533e3a](https://github.com/LuaJIT/LuaJIT/commit/ae533e3a) FFI: Fix dangling reference to CType.
* [ ] [50d6883e](https://github.com/LuaJIT/LuaJIT/commit/50d6883e) Abstract out on-demand loading of FFI library.
* [ ] [3f9389ed](https://github.com/LuaJIT/LuaJIT/commit/3f9389ed) Throw any errors before stack changes in trace stitching.
* [ ] [44bd7437](https://github.com/LuaJIT/LuaJIT/commit/44bd7437) Fix jit.dump() output for IR_CONV.
[...]

PR #9484, opened 2023-12-14
Avoid assertion in case of stack overflow from stitched trace

fckxorg:
Avoid assertion in case of stack overflow from stitched trace.

Part of #9145

NO_DOC=LJ
NO_TEST=LJ


igormunkin:
The changeset is merged to master in scope of #9638, backported to 3.0 and 2.11 in scope of #9639 and #9640 respectively.

PR #9492, opened 2023-12-15
config: credential sharding role

ImeevMA:
This patch-set introduces new credential role - credential role sharding.

Part of #8862


src/box/lua/config/applier/sharding.lua
```diff
@@ -21,6 +21,9 @@ local function apply(config)
     end
     local cfg = configdata:sharding()
     if is_storage then
+        local vexports = require('vshard.storage.exports')
+        local exports = vexports.compile(vexports.log[#vexports.log])
+        vexports.deploy_funcs(exports)
```
Totktonada, line 26: Shouldn't we do it only in RW and wait for RW on replicas?

src/box/lua/config/applier/credentials.lua
```diff
@@ -296,6 +296,62 @@ local function privileges_add_defaults(name, role_or_user, intermediate)
...
+--
+-- Create a credential sharding role if it is assigned to any user or role and
+-- has not been declared in 'credentials.roles'.
+--
+-- Replicasets with the sharding storage role must be assigned the credentials
+-- sharding role. In these replicasets, the role will be assigned all necessary
+-- privileges.
+--
+-- For replicasets that do not have the sharding storage role assigned, the
+-- credentials sharding role will not have any privileges.
```
Totktonada, line 308: Nit: We discussed it as a workaround solution for problems around ER_SCHEMA_NEEDS_UPGRADE. I think it should be clarified here. So, we'll have this information if we'll go to refactor config-vshard interface.

src/box/lua/config/configdata.lua
```diff
@@ -102,6 +102,19 @@ local function instance_sharding(iconfig, instance_name)
...
+    local has_sharding = false
+    local user_roles = instance_config:get(iconfig, {'credentials', 'users',
+                                                     u.login, 'roles'}) or {}
+    for _, cred_role in pairs(user_roles) do
+        has_sharding = has_sharding or cred_role == 'sharding'
+    end
+    if not has_sharding then
+        local err = "storage user %q should have %q role"
+        error(err:format(u.login, 'sharding'), 0)
+    end
```
Totktonada, line 116: Shouldn't we allow to set a user in `iproto.advertise.sharding` that is not present in the `credentials` section at all? I mean, it may be present in snapshots already and have all the needed privileges.

xuniq (approved):

src/box/lua/config/applier/credentials.lua
ImeevMA: Expanded the comment.

src/box/lua/config/configdata.lua
ImeevMA: True. Fixed.

ochaplashkin (approved):

src/box/lua/config/applier/credentials.lua
Totktonada: Nice, thanks!

src/box/lua/config/applier/sharding.lua
```diff
@@ -21,6 +21,17 @@ local function apply(config)
...
+            local vexports = require('vshard.storage.exports')
+            local exports = vexports.compile(vexports.log[#vexports.log])
+            vexports.deploy_funcs(exports)
+        end
+        box.watch('box.status', function(_, status)
+            -- It's ok, if deploy_funcs() will be triggered several times.
+            if status.is_ro == false then
+                deploy_funcs()
+            end
+        end)
```
Totktonada, line 34: 1. A new watcher will start at each reload.
2. Shouldn't we re-check that the instance has the `storage` sharding role on each RO -> RW transition (and stop the watcher if we're not storage anymore)?

src/box/lua/config/configdata.lua
```diff
@@ -210,7 +210,7 @@ function methods.sharding(self)
     local cfg = {
         sharding = sharding,
         box_cfg_mode = 'manual',
-        schema_management_mode = 'auto',
+        schema_management_mode = 'manual_access',
```
Totktonada, line 213: Nit: I would appreciate a few words about a reson behind using this option here (in the code).

Totktonada (approved):
Thanks! It is generally OK for me. I left a few proposals -- they're up to you.

src/box/lua/config/applier/sharding.lua
ImeevMA: 1. True, fixed.
2. For now I'm only checking if the instance has the storage sharding role on startup and reload. The watcher now stops if the storage sharding role has been removed. The storage sharding role shouldn't be lost on RO->RW transition, or is there a way?

src/box/lua/config/configdata.lua
ImeevMA: Added a comment.

src/box/lua/config/applier/sharding.lua
Totktonada: Thanks!

> The storage sharding role shouldn't be lost on RO->RW transition, or is there a way?

I see no such a way.

src/box/lua/config/configdata.lua
Totktonada: I meant a bit different thing. Here you said that this mode allows to manage everything manually. However, the most interesting (and non-obvious) fact is why we need it. AFAIU, there are two reasons:

* To sync privilige management with applier/credentials.lua.
* To deploy the functions on RO->RW transition -- it is important if all instances of a replicaset are started in RO (say, in case of `replication.failover: election`).

Feel free to ignore. I just want some record about reasons of our decisions.

ImeevMA: Thanks for the suggestion! I fixed the comment and rebased on the master.

Totktonada: Thank you for the update!

PR #9493, opened 2023-12-16
config: don't show box.cfg table in debug logs

Totktonada:
It may contain passwords in box.cfg.replication option and the changed options are printed to logs anyway. No much sense to keep printing the full table.


ImeevMA (approved):

Totktonada:
No need to run extra CI jobs: here we change a log message on the debug level and it is not checked in any test. I'll pass it over.

PR #9494, opened 2023-12-16
config: better diagnostic for a corrupted snapshot

Totktonada:
If a given snapshot has no instance or replicaset UUID, give a meaningful diagnostic and fail.

Part of #8862

### DIFF
--- a/changelogs/unreleased/config-better-corrupted-snap-diagnostic.md
+++ b/changelogs/unreleased/config-better-corrupted-snap-diagnostic.md
@@ -1,0 +1,4 @@
+## bugfix/config
+
+* Improved the diagnostic given for a corrupted snapshot file without an
+  instance or replica set UUID (gh-8862).
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -6,6 +6,7 @@
 local fun = require('fun')
 local urilib = require('uri')
 local digest = require('digest')
+local uuid = require('uuid')
 local instance_config = require('internal.config.instance_config')
 local cluster_config = require('internal.config.cluster_config')
 local snapshot = require('internal.config.utils.snapshot')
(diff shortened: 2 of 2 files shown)


changelogs/unreleased/config-better-corrupted-snap-diagnostic.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Give a better diagnostic for a corrupted snapshot file: if it has no instance
+  or replicaset UUID (gh-8862).
```
xuniq, line 4:
```suggestion
* Improved the diagnostic given for a corrupted snapshot file without an instance or replica set UUID (gh-8862).
```

xuniq (approved):

src/box/lua/config/configdata.lua
```diff
@@ -316,8 +319,16 @@ local mt = {
 local function validate_names(saved_names, config_names)
     -- Snapshot always has replicaset uuid and
     -- at least one peer in _cluster space.
-    assert(saved_names.replicaset_uuid)
-    assert(saved_names.instance_uuid)
+    if saved_names.replicaset_uuid == nil then
+        error(('Snapshot file %s has no "replicaset_uuid" key in _cluster ' ..
+            'system space. The snapshot is likely corrupted.'):format(
+            saved_names.snap_path), 0)
```
Serpentian, line 325: Maybe `snapshot.get_path()`. I see no reason to save `snap_path` inside names, it's not a name and it's not needed anywhere, but this message, which is hard to reproduce.

Serpentian (approved):
Looks great. Just one minor comment

src/box/lua/config/configdata.lua
Serpentian: Ohh, I completely forgot, that it requires `iconfig` for that. The alternative is to add this checks to `snapshot.get_names()` or `configdata.find_saved_names`. It's actually an error while getting snapshot names, not really a validation error.

It's completely up to you, I'm ok with the diff too

Totktonada: I've considered this variant, but than decided that there is some sense in keeping all the uuids/names related validation in one place. TBH, I don't see much difference.

PR #9495, opened 2023-12-17
main: hide --failover CLI option

Totktonada:
The failover agent is not ready at the moment. Let's hide the CLI option from the help message, but still handle it: it is useful to continue the development of the agent.

See tarantool/tarantool-ee#564


ImeevMA (approved):

ochaplashkin (approved):

Totktonada:
Extra CI jobs are not needed here: we just change the help message and the test that verifies it is present in the reduced set of CI jobs. I'm going to pass over full-ci and push.

PR #9496, opened 2023-12-17
config: support relative path in --config

Totktonada:
Before this change the following conditions lead to an error on `config:reload()`.
    
* `--config <...>` CLI option is passed with a relative config path
* `process.work_dir` config option is set to a non-null value

It is fixed by calculating of an absolute config path on startup.

Part of #8862

### DIFF
--- a/changelogs/unreleased/config-relative-config-path.md
+++ b/changelogs/unreleased/config-relative-config-path.md
@@ -1,0 +1,5 @@
+## bugfix/config
+
+* Added support for the relative path in the `--config <...>` option. Before
+  this change, `config:reload()` failed if the `process.work_dir` option was
+  set to a non-null value (gh-8862).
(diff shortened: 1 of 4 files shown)


changelogs/unreleased/config-relative-config-path.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/config
+
+* Support relative path in `--config <...>` option. Before this change
+  `config:reload()` fails in this case if `process.work_dir` option is set to a
+  non-null value (gh-8862).
```
xuniq, line 5:
```suggestion
* Added support for the relative path in the `--config <...>` option. Before this change,
  `config:reload()` failed  if the `process.work_dir` option was set to a
  non-null value (gh-8862).
```

xuniq (approved):

ochaplashkin (approved):

ImeevMA (approved):
Thanks for the patch-set!

changelogs/unreleased/config-relative-config-path.md
Totktonada: Thanks! Updated.

PR #9504, opened 2023-12-18
config: fix automatic names apply before schema upgrade

### LINKED ISSUE #9502: config: ER_SCHEMA_NEEDS_UPGRADE when reloading configuration on a node with schema version < 3.0.0
**Bug description**

Commit 90ebc380cb25c99437b5a4210b5d4b2a351027ca introduced the logic to try and apply names right on configuration reload, if the node can do that. It seems it missed checking that the node has the most recent schema version, hence the error `ER_SCHEMA_NEEDS_UPGRADE` is thrown when config tries to set instance names.

The error also may happen during initial configuration, when configuring a single node. For some reason the config module calls `apply(config)` for a second time right after the initial configuration. This is considered a reload, and the node tries to set the missing names, like in case with `config:reload()`, and fails.

**Steps to reproduce**

Apply the following patch:
```patch
diff --git a/test/config-luatest/names_upgrade_test.lua b/test/config-luatest/names_upgrade_test.lua
index 212d74dee..83225168b 100644
--- a/test/config-luatest/names_upgrade_test.lua
+++ b/test/config-luatest/names_upgrade_test.lua
@@ -117,6 +117,7 @@ g.test_upgrade = function()
 
     g.instance_1:exec(function()
         box.ctl.wait_rw()
+        require('config'):reload()
         box.schema.upgrade()
     end)
 
```
Then run the `config-luatest/names_upgrade_test`, and observe the failure:
```
[001] instance-001 | 2023-12-18 15:50:01.702 [49329] main/123/main/tarantool.config log.lua:130 E> Your schema version is 2.11.0 while Tarantool 3.0.0-beta1-71-g79a0eb919 requires a more recent schema version. Please, consider using box.schema.upgrade().
[001] not ok 1	set-names-automatically-on-upgrade.test_upgrade
[001] #   Your schema version is 2.11.0 while Tarantool 3.0.0-beta1-71-g79a0eb919 requires a more recent schema version. Please, consider using box.schema.upgrade().
[001] #   stack traceback:
[001] #   	.../Source/sptnt/test/config-luatest/names_upgrade_test.lua:118: in function 'set-names-automatically-on-upgrade.test_upgrade'
[001] #   	...

```

sergepetrenko:
Each config reload on a node with old schema (less than 3.0.0) lead to the error ER_SCHEMA_NEEDS_UPGRADE. This happened because the relevant code didn't check for schema version before trying to apply the names.

Fix this.

Closes #9502

NO_DOC=bugfix
NO_CHANGELOG=fix in a not yet released behavior

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -283,7 +283,9 @@ local function names_apply(config, missing_names, schema_version)
     names_alert_missing(config, missing_names)
     -- Don't wait for box.status to change, we may be already rw, set names
     -- on reload, if it's possible and needed.
-    names_try_set_missing()
+    if schema_version and schema_version >= mkversion.get_latest() then
+        names_try_set_missing()
+    end
 
     if names_state.is_configured then
         -- All triggers are already configured, nothing to do, but wait.
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -210,10 +210,10 @@ local function names_schema_upgrade_on_replace(old, new)
         return
     end
 
-    local version_3 = mkversion(3, 0, 0)
+    local latest_version = mkversion.get_latest()
     local old_version = mkversion.from_tuple(old)
     local new_version = mkversion.from_tuple(new)
-    if old_version < version_3 and new_version >= version_3 then
+    if old_version < latest_version and new_version >= latest_version then
         -- We cannot do it inside on_replace trigger, as the version
         -- is not considered set yet and we may try to set names, when
         -- schema is not yet updated, which will fail as DDL is
@@ -295,7 +295,7 @@ local function names_apply(config, missing_names, schema_version)
 
     -- Wait for rw state. If schema version is nil, bootstrap is
     -- going to be done.
-    if schema_version and schema_version < mkversion(3, 0, 0) then
+    if schema_version and schema_version < mkversion.get_latest() then
         box.space._schema:on_replace(names_schema_upgrade_on_replace)
         names_state.is_upgrade_wait = true
     else
--- a/extra/exports
+++ b/extra/exports
@@ -92,6 +92,7 @@ box_latch_lock
 box_latch_new
 box_latch_trylock
 box_latch_unlock
+box_latest_dd_version_id
 box_on_shutdown
 box_read_ffi_disable
 box_read_ffi_is_disabled
--- a/src/box/lua/mkversion.lua
+++ b/src/box/lua/mkversion.lua
@@ -1,7 +1,11 @@
 local bit = require('bit')
 local ffi = require('ffi')
 
-ffi.cdef([[ uint32_t box_dd_version_id(void); ]])
+ffi.cdef([[
+    uint32_t box_dd_version_id(void);
+    uint32_t box_latest_dd_version_id(void);
+]])
+
 local builtin = ffi.C
 
 local mkversion_mt = {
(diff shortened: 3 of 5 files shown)


src/box/lua/config/applier/box_cfg.lua
```diff
@@ -283,7 +283,9 @@ local function names_apply(config, missing_names, schema_version)
     names_alert_missing(config, missing_names)
     -- Don't wait for box.status to change, we may be already rw, set names
     -- on reload, if it's possible and needed.
-    names_try_set_missing()
+    if schema_version and schema_version >= mkversion(3, 0, 0) then
+        names_try_set_missing()
+    end
```
Totktonada, line 288: Will this check fail on tarantool 3.1?

Shoudn't we check that the schema is the last one, not just >= 3.0.0?

Serpentian: Agree, I propose to export `box_latest_dd_version_id` the same way it's done with `box_dd_version_id` and add this to `mkversion.lua` file. This way we'll get latest schema version

sergepetrenko: Good catch, will do.

sergepetrenko: Samee stands for https://github.com/tarantool/tarantool/blob/26acba832c4e33668b6eef52051a7d1f4a30ee2e/src/box/lua/config/applier/box_cfg.lua#L298-L300
and
https://github.com/tarantool/tarantool/blob/26acba832c4e33668b6eef52051a7d1f4a30ee2e/src/box/lua/config/applier/box_cfg.lua#L213-L223
right?

Serpentian: Right, we're not supposed to set names before the time, the schema is latest. We cannot do it at all. Missed that during initial implementation

sergepetrenko:
@ImeevMA, can you check that this patch fixes your problem?

src/box/lua/config/applier/box_cfg.lua
Totktonada: OTOH, it works while the latest schema is 3.0.0 and we can rework it after the 3.0.0 release.

sergepetrenko: > OTOH, it works while the latest schema is 3.0.0 and we can rework it after the 3.0.0 release.

It was quite easy to fix so I did this as well.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -283,7 +283,9 @@ local function names_apply(config, missing_names, schema_version)
     names_alert_missing(config, missing_names)
     -- Don't wait for box.status to change, we may be already rw, set names
     -- on reload, if it's possible and needed.
-    names_try_set_missing()
+    if schema_version and schema_version >= mkversion.get_latest() then
```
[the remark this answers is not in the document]
sergepetrenko: Actually, it can. The user may run `box.schema.upgrade()` before upgrading all the instances to the latest tarantool version. Most probably something will break, if the upgrade is non-trivial. But there is a possibility that it'll succeed.

Besides, the user may simply do a replace to space `_schema` with a fake huge version. Again, something will break in this case, but we shouldn't assert that the version is never greater than `latest_version`.

TBH I think we're lacking a knob to prohibit DDL, when schema version is greater than the last known to this tarantool build. Similarly to how we prohibit DDL when schema version is too old.

Long story short, I'd keep this check as is. This will be consistent with other DDL, which wouldn't error if real schema version is greater than the last_version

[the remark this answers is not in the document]
Serpentian: I'm ok with that, thank you for the explanation

Totktonada (approved):
I see no problems. Thanks for the investatigation and the patchset!

src/box/lua/config/applier/box_cfg.lua
Totktonada: Thank you!

ylobankov (approved):

PR #9511, opened 2023-12-19
Rework Lua module searchers and loaders

igormunkin:
This patchset implements the refactoring of Tarantool `package.loaders` required for Tarantool integrity protection (needed for tarantool/tarantool-ee#585).

**Per-patch review is highly recommended (though GitHub works badly in this mode).**

Since Tarantool follows [Lua 5.1 semantics](https://www.lua.org/manual/5.1/manual.html), the set of routines in package.loaders is used to tweak Lua modules loading. E.g. overriding
machinery has been implemented on top of the `package.loaders`. Unfortunately, `package.loaders` are not such flexible as it's needed and this mechanism is redesigned in [Lua 5.2](https://www.lua.org/manual/5.2/manual.html).

Starting from Lua 5.2 package.loaders has been superseded by `package.searchers`. This is quite natural change, since nobody wants to tweak the "loading" rules, but rather the "searching" process. As a result of this redesign evolution up to the [Lua 5.4](https://www.lua.org/manual/5.4/manual.html), users are able to obtain the "source" of the module being loaded (i.e. file name or specific sentinel for preload packages).

---

This first patch introduces Tarantool internal table with package.searchers and reimplements package.preload and tarantool.builtin loaders using the machinery described above:
* The new searcher function, respecting Lua 5.4 semantics, is implemented for each loader.
* Each searcher is wrapped with the function to implement "dummy" loader respecting Lua 5.1 semantics.

---

Originally, Lua uses only constant path patterns in `package.path` and `package.cpath`. Tarantool enhances user experience here and provides more relaxed rules for module searching: some paths are calculated at the moment the module is being searched and are relative to the current working directory. Unfortunately it makes "searchers" more complex, since all the paths can't be calculated when Tarantool starts up.

Furthermore, there is the particular priority while loading Lua modules in Tarantool. Originally, Lua tries to load the particular Lua script using the hints from `package.path` and *only if* no Lua module is found, Lua tries to load the corresponding Lua C library using the hints from `package.cpath`. Tarantool loaders are prioritized by the locations at first and only then by the module implementation. This also makes "searchers" structure more complex than it could be.

Anyway, "file" loaders are split into the three layers:
1. "Pathogen": the function generated by `<gen_path_builder>`, building the hints for searchers.
2. "Searcher": the function generated by `<gen_file_searcher>` helper, using the particular file loading method (`<load_lua>` or `<load_lib>`) and the given pathogen to obtain the hints with path patterns to be used in `<package.searchpath>` routine.
3. "Loader": the function generated by `<gen_file_loader>` helper, following Lua 5.1 loaders semantics and using the particular searcher.

All the generated searchers are stored in the <package.searchers> analogue introduced in the first patch of this refactoring series.

---

Considering the way "app" paths injection is implemented in Tarantool loaders to save the loading priority, the chained loaders machinery introduced in the commit 7e9051c44905dbe5c2396568ee252e09bf1963da ("lua: don't set built-in modules to package.loaded") is also implemented for the new searching mechanism. The approach is the same as the one used for chaining loaders. As a result, "app" paths searchers are injected into the "cwd" paths searchers the same way it has been done in the original loaders context.

The third patch also introduces numerical indices into <package.searchers> analogue to provide the convenient searchers <-> loaders mapping used in the final patch of this series.

--- 

Within the fourth patch the overriding mechanism introduced in the commit ec47c389fd5e5e64a1a8d1d33dac108c19b325c0 ("lua: allow to override a built-in module") is moved to the "searching" phase of the new loading machinery. Furthermore, `package.path`/`package.cpath` searchers are also added to the indexed list made for the mapping.

---

Within the fifth patch the mechanism for enabling/disabling loaders introduced in the commit 87b4da31e1ac2c14e9779054b28601862652d706 ("lua: enable/disable override loader") is moved to the "searching" phase of the new loading machinery.

---

Though croot loader is not the popular way to work with Lua C libraries, some modules are organised using this mechanism, so they have to be verified via Tarantool integrity protection system, hence croot loader has to be reimplemented to the new searching machinery.
    
The corresponding croot searcher is implemented in this patch and added to the indexed part of the <package.searchers> table. 
    
It's worth to mention two peculiarities related to the croot loader and searcher:
* Since Tarantool uses original croot loader from LuaJIT, it ignores all paths except ones specified in `<package.cpath>`. However, there might be some croot-like modules in the Tarantool-specific cpaths (cwd or application root). It looks like croot machinery is not widely used, if this bug bothers nobody, but it's still need to be fixed.
* Similar situation relates to the overriding mechanism, that was recently introduced: there is no "prefixed" croot loader and all libraries using croot loading machinery are ignored for the overridden paths as a result.

---

Though all Lua 5.1 loaders machinery will be superseded by Lua 5.4 searchers machinery, Tarantool still should follow Lua 5.1 semantics, so the "legacy" loader is required. This patch implements the common helper `<gen_legacy_loader>` to wrap all loaders and make the new machinery conform the Lua 5.1 semantics.
    
Moreover, some since all tweaks and helpers are implemented for searching machinery, `<chain_loaders>` is not required anymore.

---

Finally, to make it possible to instrument the searchers with Tarantool integrity protection, two steps are left.
* Implement mapping from searchers table to `<package.loaders>` using `<gen_legacy_loader>` helper.
* Obtain the searcher from the table directly, so the runtime instrumentation take the effect.

As a result `<gen_legacy_loader>` is reimplemented to use the table with searchers as an upvalue instead of the particular searcher routiine, and `<package.loaders>` table is filled by the function generated with the updated helper.

---
---

The first patch of the series fixes error handling between module "searching" and "loading" processes. Originally, if module is not found, the error message is returned to <require> and is accumulated to the "not found" message. If no suitable module is found, the error is raised by `<require>`. In case, when the desired module is found, <require> tries to load it. If any error occurs while loading the module, the "loader" yields the corresponding error. The latter part was broken since the commit 4e27b0e61747c5c6a8e7b9e53b18fa78f24ae61a ("app: let user define search root for packages"), so if the error occurs while loading the module, Tarantool proceed with searching the module in other locations.
    
This bug was found via [lua-Harness suite](https://fperrad.frama.io/lua-Harness/) used in tarantool/luajit testing routine, so the test added in the patch partially duplicates the particular test case in lua-Harness.

### DIFF
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -58,36 +58,63 @@ local function traverse_path(path)
     return paths
 end
 
--- Generate a search function, which performs searching through
--- templates setup in options.
+-- Generate a path builder function, which yields the set of the
+-- paths to be searched through for the desired module.
 --
--- @param path_fn function which returns a base path for the
---     resulting template
+-- @param basepath_fn function which returns a base path for the
+--     resulting path set
 -- @param templates table with lua search templates
 -- @param need_traverse bool flag which tells search function to
 --     build multiple paths by expanding base path up to the
 --     root ('/')
--- @return a searcher function which builds a path template and
---     calls package.searchpath
-local function gen_search_func(path_fn, templates, need_traverse)
-    assert(type(path_fn) == 'function', 'path_fn must be a function')
+-- @return a path builder function which builds and yields the
+--     resulting path set
+local function gen_path_builder(basepath_fn, templates, need_traverse)
+    assert(type(basepath_fn) == 'function', 'basepath_fn must be a function')
     assert(type(templates) == 'table', 'templates must be a table')
 
-    return function(name)
-        local path = path_fn() or '.'
-        local paths = need_traverse and traverse_path(path) or { path }
+    return function()
+        local base = basepath_fn() or '.'
+        local dirs = need_traverse and traverse_path(base) or { base }
 
         local searchpaths = {}
 
-        for _, path in ipairs(paths) do
+        for _, dir in ipairs(dirs) do
             for _, template in pairs(templates) do
-                table.insert(searchpaths, minifio.pathjoin(path, template))
+                table.insert(searchpaths, minifio.pathjoin(dir, template))
             end
         end
 
-        local searchpath = table.concat(searchpaths, ';')
+        return table.concat(searchpaths, ';')
+    end
+end
 
-        return package.searchpath(name, searchpath)
+local path = {
+    package = function() return package.path end,
+    cwd = {
+        dot = gen_path_builder(searchroot, LUA_TEMPLATES),
+        rocks = gen_path_builder(searchroot, ROCKS_LUA_TEMPLATES, true),
+    }
+}
+
+local cpath = {
+    package = function() return package.cpath end,
+    cwd = {
+        dot = gen_path_builder(searchroot, LIB_TEMPLATES),
+        rocks = gen_path_builder(searchroot, ROCKS_LIB_TEMPLATES, true),
+    }
+}
+
+local function gen_file_searcher(loader, pathogen)
+    assert(type(loader) == 'function', '<loader> must be a function')
+    assert(type(pathogen) == 'function', '<pathogen> must be a function')
+
+    return function(name)
+        local path, errmsg = package.searchpath(name, pathogen())
+        if not path then
+            return errmsg
+        end
+        return loader, path
     end
 end
 
@@ -118,45 +145,58 @@ rawset(searchers, 'builtin', function(name)
     return ("\n\tno field loaders.builtin['%s']"):format(name)
 end)
 
+rawset(searchers, 'path.package', gen_file_searcher(load_lua, path.package))
+rawset(searchers, 'cpath.package', gen_file_searcher(load_lib, cpath.package))
+rawset(searchers, 'path.cwd.dot', gen_file_searcher(load_lua, path.cwd.dot))
+rawset(searchers, 'cpath.cwd.dot', gen_file_searcher(load_lib, cpath.cwd.dot))
+rawset(searchers, 'path.cwd.rocks', gen_file_searcher(load_lua, path.cwd.rocks))
+rawset(searchers, 'cpath.cwd.rocks', gen_file_searcher(load_lib, cpath.cwd.rocks))
+
 -- Compose a loader function from options.
 --
--- @param search_fn function will be used to search a file from
+-- @param searcher function will be used to search a file from
 --     path template
--- @param load_fn function will be used to load a file, found by
---     search function
 -- @return function a loader, which first search for the file and
 --     then loads it
-local function gen_loader_func(search_fn, load_fn)
-    assert(type(search_fn) == 'function', 'search_fn must be defined')
-    assert(type(load_fn) == 'function', 'load_fn must be defined')
+local function gen_file_loader(searcher)
+    assert(type(searcher) == 'function', '<searcher> must be defined')
 
     return function(name)
         if not name then
             return "empty name of module"
         end
-        local file, err = search_fn(name)
-        if not file then
-            return err
+        local loader, data = searcher(name)
+        -- XXX: <loader> (i.e. the first return value) contains
+        -- error message in this case. Just propagate it to the
+        -- <require> frame...
+        if not data then
+           return loader
         end
-        local loaded, err = load_fn(file, name)
+        -- XXX: ... Otherwise, this is a valid module loader.
+        -- Load the given <data> and return the result.
+        local loaded, err = loader(data, name)
         local message = ("error loading module '%s' from file '%s':\n\t%s")
-                        :format(name, file, err)
+                        :format(name, data, err)
         return assert(loaded, message)
     end
 end
 
-local search_lua = gen_search_func(searchroot, LUA_TEMPLATES)
-local search_lib = gen_search_func(searchroot, LIB_TEMPLATES)
-local search_rocks_lua = gen_search_func(searchroot, ROCKS_LUA_TEMPLATES, true)
-local search_rocks_lib = gen_search_func(searchroot, ROCKS_LIB_TEMPLATES, true)
-
+-- XXX: Though <package.search> looks like a handy public
+-- function, it's only the helper used for searching Tarantool
+-- C modules. Hence, despite it uses the same searchers as loaders
+-- for Lua modules do, it follows its own semantics ignoring
+-- preload, override, builtin modules, and application specific
+-- locations.
+-- FIXME: Lua paths look really odd in the list, since the
+-- function is used to load only C modules. However, these paths
+-- are left to avoid breaking change.
 local search_funcs = {
-    search_lua,
-    search_lib,
-    search_rocks_lua,
-    search_rocks_lib,
-    function(name) return package.searchpath(name, package.path) end,
-    function(name) return package.searchpath(name, package.cpath) end,
+    searchers['path.cwd.dot'],
+    searchers['cpath.cwd.dot'],
+    searchers['path.cwd.rocks'],
+    searchers['cpath.cwd.rocks'],
+    searchers['path.package'],
+    searchers['cpath.package'],
 }
 
 local function search(name)
@@ -164,8 +204,8 @@ local function search(name)
         return "empty name of module"
     end
     for _, searcher in ipairs(search_funcs) do
-        local file = searcher(name)
-        if file ~= nil then
+        local loader, file = searcher(name)
+        if type(loader) == 'function' then
             return file
         end
     end
@@ -239,10 +279,10 @@ local function chain_loaders(subloaders)
 end
 
 -- loader_preload 1
-table.insert(package.loaders, 2, gen_loader_func(search_lua, load_lua))
-table.insert(package.loaders, 3, gen_loader_func(search_lib, load_lib))
-table.insert(package.loaders, 4, gen_loader_func(search_rocks_lua, load_lua))
-table.insert(package.loaders, 5, gen_loader_func(search_rocks_lib, load_lib))
+table.insert(package.loaders, 2, gen_file_loader(searchers['path.cwd.dot']))
+table.insert(package.loaders, 3, gen_file_loader(searchers['cpath.cwd.dot']))
+table.insert(package.loaders, 4, gen_file_loader(searchers['path.cwd.rocks']))
+table.insert(package.loaders, 5, gen_file_loader(searchers['cpath.cwd.rocks']))
 -- package.path   6
 -- package.cpath  7
 -- croot          8
@@ -263,12 +303,19 @@ if script ~= nil and script ~= '-' then
     end
 
     -- Search non-recursively, only next to the script.
-    local search_app_lua = gen_search_func(script_dir_fn, LUA_TEMPLATES)
-    local search_app_lib = gen_search_func(script_dir_fn, LIB_TEMPLATES)
-    local search_app_rocks_lua = gen_search_func(script_dir_fn,
-        ROCKS_LUA_TEMPLATES)
-    local search_app_rocks_lib = gen_search_func(script_dir_fn,
-        ROCKS_LIB_TEMPLATES)
+    path.app = {
+        dot = gen_path_builder(script_dir_fn, LUA_TEMPLATES),
+        rocks = gen_path_builder(script_dir_fn, ROCKS_LUA_TEMPLATES),
+    }
+    cpath.app = {
+        dot = gen_path_builder(script_dir_fn, LIB_TEMPLATES),
+        rocks = gen_path_builder(script_dir_fn, ROCKS_LIB_TEMPLATES)
+    }
+
+    rawset(searchers, 'path.app.dot', gen_file_searcher(load_lua, path.app.dot))
+    rawset(searchers, 'cpath.app.dot', gen_file_searcher(load_lib, cpath.app.dot))
+    rawset(searchers, 'path.app.rocks', gen_file_searcher(load_lua, path.app.rocks))
+    rawset(searchers, 'cpath.app.rocks', gen_file_searcher(load_lib, cpath.app.rocks))
 
     -- Mix the script directory loaders into corresponding
     -- searchroot based loaders. It allows to avoid changing
@@ -287,19 +334,19 @@ if script ~= nil and script ~= '-' then
     -- do that for all the logic at once.
     package.loaders[2] = chain_loaders({
         package.loaders[2],
-        gen_loader_func(search_app_lua, load_lua),
+        gen_file_loader(searchers['path.app.dot']),
     })
     package.loaders[3] = chain_loaders({
         package.loaders[3],
-        gen_loader_func(search_app_lib, load_lib),
+        gen_file_loader(searchers['cpath.app.dot']),
     })
     package.loaders[4] = chain_loaders({
         package.loaders[4],
-        gen_loader_func(search_app_rocks_lua, load_lua),
+        gen_file_loader(searchers['path.app.rocks']),
     })
     package.loaders[5] = chain_loaders({
         package.loaders[5],
-        gen_loader_func(search_app_rocks_lib, load_lib),
+        gen_file_loader(searchers['cpath.app.rocks']),
     })
 end
 
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -152,6 +152,13 @@ rawset(searchers, 'cpath.cwd.dot', gen_file_searcher(load_lib, cpath.cwd.dot))
 rawset(searchers, 'path.cwd.rocks', gen_file_searcher(load_lua, path.cwd.rocks))
 rawset(searchers, 'cpath.cwd.rocks', gen_file_searcher(load_lib, cpath.cwd.rocks))
 
+-- Indices in <searchers> table must represent the order of the
+-- functions in <package.loaders> below.
+rawset(searchers, 2, searchers['path.cwd.dot'])
+rawset(searchers, 3, searchers['cpath.cwd.dot'])
+rawset(searchers, 4, searchers['path.cwd.rocks'])
+rawset(searchers, 5, searchers['cpath.cwd.rocks'])
+
 -- Compose a loader function from options.
 --
 -- @param searcher function will be used to search a file from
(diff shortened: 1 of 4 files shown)


ochaplashkin (approved):

Buristan (changes requested):
Hi, Igor!
Thanks for the patchset!
I'll proceed with the review per-patch.

**[PATCH 1/9] lua: fix loader behaviour in case of syntax error**

Thanks for the patch!
Please consider my comments below.

> processes. Originally, if module is not found, the error message is

Typo: s/module/a module/

> returned to <require> and is accumulated to the "not found" message.

Nit: "accumulated" isn't clear to me, do you mean "appended"?

> Tarantool proceed with searching the module in other locations.

Typo: s/proceed/proceeds/

> This bug was found via lua-Harness suite[1] used in tarantool/luajit

Typo: s/via lua-Harness/via the lua-Harness/
Typo: s/suite[1]/suite [1]/
Typo: s<in tarantool/luajit><in the tarantool/luajit>

> NO_CHANGELOG=bugfix

Should we mention this in the changelog as bugfix then?

test/app-luatest/fix_lua_loader_error_handling_test.lua
```diff
@@ -0,0 +1,24 @@
...
+    luadir = treegen.prepare_directory(g, {('%s.lua'):format(module)})
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+g.test_loader_error_handling = function()
+    package.path = ('%s/?.lua'):format(luadir)
+    local errlike = ("error loading module '%s' from file '%s/%s%%.lua':.+")
```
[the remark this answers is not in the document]
Buristan: PEBKAC, didn't notice the single quotes inside string.

src/lua/loaders.lua
```diff
@@ -119,6 +91,33 @@ local function gen_search_func(path_fn, templates, need_traverse)
...
 
+-- Lua table with "package searchers".
+local searchers = debug.getregistry()._PACKAGE_SEARCHERS
+
+rawset(searchers, 'preload', function(name)
+    assert(type(package.preload) == 'table',
+           "'package.preload' must be a table")
+    local loader = package.preload[name]
+    if loader ~= nil then
+        return loader, ':preload:'
```
Buristan: I've seen several usages of the keyword `preload` (`:preload:`) or `builtin` (`:builtin:`) here. Maybe, it is better to use some constants for them (like `BUILTIN_SEARHER_NAME`)?
Feel free to ignore.

Buristan (approved):
**[PATCH 2/9] lua: implement searchers for preload and builtin**

Thanks for the patch!
LGTM, with a few insignificant nits below.

> Since Tarantool follows Lua 5.1[1] semantics, the set of routines in

Typo: s/Lua 5.1[1]/Lua 5.1 [1]/
Here and below with other links.

> Unfortunately, package.loaders are not such flexible as it's needed and

Typo: s/not such flexible as/not as flexible as/
Typo: s/it's needed/they should be/

> package.searchers. This is quite natural change, since nobody wants to

Typo: s/quite natural/quite a natural/

Buristan (approved):
**[PATCH 4/9] lua: implement chained searchers**

I see no flaws, so patch LGTM, except a single nit regarding to the commit message and a comment below.

> This patch proceeds the refactoring of Tarantool package.loaders

Typo: s/proceeds/proceeds with/

Buristan (approved):
**[PATCH 6/9] lua: implement conditional searchers**

LGTM, with two nits regarding the commit message.

> This patch proceeds the refactoring of Tarantool package.loaders

Typo: s/proceeds/proceeds with/

> Within this patch the mechanism for enabling/disabling loaders

Typo: s/Within this patch/Within this patch,/

Buristan (approved):
**[PATCH 7/9] lua: implement croot searcher**

LGTM, with a minor nits regarding the commit message and the comments below.

> This patch proceeds the refactoring of Tarantool package.loaders

Typo: s/proceeds/proceeds with/

> verified via Tarantool integrity protection system, hence croot loader

Typo: s/hence/hence,/

> * Since Tarantool uses original croot loader from LuaJIT, it ignores all

Typo: s/original croot/the original croot/

>  if this bug bothers nobody, but it's still need to be fixed.

Typo: s/it's still need/it still needs/

> * Similar situation relates to the overriding mechanism, that was

Typo: s/Similar/A similar/
Typo: s/mechanism,/mechanism/

Buristan (approved):
**[PATCH 8/9] lua: implement common legacy loader generator**

LGTM, with a minor comments regarding the commit message.

> This patch proceeds the refactoring of Tarantool package.loaders

Typo: s/proceeds/proceeds with/

> Moreover, some since all tweaks and helpers are implemented for

Typo: s/some//

Buristan (approved):
**[PATCH 9/9] lua: implement searchers-to-loaders mapping**

LGTM, with a minor comments regarding the commit message.

> instrumentation take the effect.

Typo: s/take/takes/

> As a result <gen_legacy_loader> is reimplemented to use the table with

Typo: s/As a result/As a result,/

test/app-luatest/fix_lua_loader_error_handling_test.lua
```diff
@@ -0,0 +1,24 @@
...
+    treegen.add_template(g, ('^%s%%.lua$'):format(module), [[?syntax error?]])
+    luadir = treegen.prepare_directory(g, {('%s.lua'):format(module)})
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+g.test_loader_error_handling = function()
+    package.path = ('%s/?.lua'):format(luadir)
```
Totktonada, line 20: It affects other test files if several tests are run using `luatest` directly (w/o test-run).

src/lua/loaders.lua
```diff
@@ -351,6 +350,23 @@ local override_loader = conditional_loader(chain_loaders({
     prefix_loader('override', package.loaders[7]),
 }), override_loader_onoff)
 
+
+local function dummy_loader(searcher, sentinel)
+    return function(name)
+        local loader, data = searcher(name)
```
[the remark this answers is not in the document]
igormunkin: This part was reimplemented in **[PATCH 8/9] lua: implement common legacy loader generator**. Ignoring.

src/lua/loaders.lua
```diff
@@ -119,6 +91,33 @@ local function gen_search_func(path_fn, templates, need_traverse)
     end
 end
 
+-- Lua table with "package searchers".
+local searchers = debug.getregistry()._PACKAGE_SEARCHERS
```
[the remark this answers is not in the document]
igormunkin: Agree here, thanks! Fixed.

```diff --git a/src/lua/init.c b/src/lua/init.c
index 14ab09fd53..30bcf828d9 100644
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -915,7 +915,7 @@ tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 
        /* Create a table for "package.searchers" analogue. */
        lua_newtable(L);
-       lua_setfield(L, LUA_REGISTRYINDEX, "_PACKAGE_SEARCHERS");
+       lua_setfield(L, LUA_REGISTRYINDEX, "_TARANTOOL_PACKAGE_SEARCHERS");
 
        /*
         * Create a table for storing loaded built-in modules.
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index f4f939c68f..44599a960d 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -148,7 +148,7 @@ local function yield_builtin_loader(loader, sentinel)
 end
 
 -- Lua table with "package searchers".
-local searchers = debug.getregistry()._PACKAGE_SEARCHERS
+local searchers = debug.getregistry()._TARANTOOL_PACKAGE_SEARCHERS
 
 rawset(searchers, 'preload', function(name)
     assert(type(package.preload) == 'table',```

File: src/lua/loaders.lua (hunk 2 above)
Comment (igormunkin) [reply src/lua/loaders.lua]: This seems to be resolved in scope of **[PATCH 8/9] lua: implement common legacy loader generator** either, doesn't it?

File: src/lua/loaders.lua [hunk 6]
```diff
@@ -152,6 +152,11 @@ rawset(searchers, 'cpath.cwd.dot', gen_file_searcher(load_lib, cpath.cwd.dot))
 rawset(searchers, 'path.cwd.rocks', gen_file_searcher(load_lua, path.cwd.rocks))
 rawset(searchers, 'cpath.cwd.rocks', gen_file_searcher(load_lib, cpath.cwd.rocks))
 
+rawset(searchers, 2, searchers['path.cwd.dot'])
```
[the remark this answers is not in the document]
igormunkin: Fixed.

```diff
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index 95cf5d8570..eaec6a7f6d 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -152,6 +152,8 @@ rawset(searchers, 'cpath.cwd.dot', gen_file_searcher(load_lib, cpath.cwd.dot))
 rawset(searchers, 'path.cwd.rocks', gen_file_searcher(load_lua, path.cwd.rocks))
 rawset(searchers, 'cpath.cwd.rocks', gen_file_searcher(load_lib, cpath.cwd.rocks))
 
+-- Indices in <searchers> table must represent the order of the
+-- functions in <package.loaders> below.
 rawset(searchers, 2, searchers['path.cwd.dot'])
 rawset(searchers, 3, searchers['cpath.cwd.dot'])
 rawset(searchers, 4, searchers['path.cwd.rocks'])```

File: test/app-luatest/fix_lua_loader_error_handling_test.lua [hunk 7]
```diff
@@ -0,0 +1,24 @@
+local t = require('luatest')
+local treegen = require('test.treegen')
+
+local g = t.group()
+
+local luadir
+local module = 'syntax'
```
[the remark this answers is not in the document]
igormunkin: I guess `s/module/name/gc` is fine, thanks!

test/app-luatest/fix_lua_loader_error_handling_test.lua
```diff
@@ -0,0 +1,24 @@
+local t = require('luatest')
```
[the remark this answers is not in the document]
igormunkin: Oops, my bad. I've finally found out the reason why the test checks nothing. The fix is below. Furthermore, I've fixed all other your comments within this changeset.

```diff
diff --git a/test/app-luatest/fix_lua_loader_error_handling_test.lua b/test/app-luatest/fix_lua_loader_error_handling_test.lua
index 9aa552f6ea..f42f15abd3 100644
--- a/test/app-luatest/fix_lua_loader_error_handling_test.lua
+++ b/test/app-luatest/fix_lua_loader_error_handling_test.lua
@@ -1,24 +1,38 @@
 local t = require('luatest')
 local treegen = require('test.treegen')
+local fio = require('fio')
 
 local g = t.group()
 
+-- XXX: I would rather tweak <package.path>, since it's the
+-- recommended way to manage the hints for the loaders. However,
+-- the loader provided by LuaJIT that uses <package.path> works
+-- fine (i.e. throws the expected error), unlike the ones added
+-- for Tarantool. Hence, the easiest way to check Tarantool
+-- loaders is switching the current working directory for the
+-- test. Preserve the original working directory to restore it at
+-- the end of the test.
+local cwd = fio.cwd()
+
 local luadir
-local module = 'syntax'
+local name = 'syntax'
+local errlike = [[error loading module '<NAME>' from file '<DIR>/<NAME>.lua':
+       <DIR>/<NAME>.lua:1: unexpected symbol near '?']]
+
 
 g.before_all(function()
     treegen.init(g)
-    treegen.add_template(g, ('^%s%%.lua$'):format(module), [[?syntax error?]])
-    luadir = treegen.prepare_directory(g, {('%s.lua'):format(module)})
+    treegen.add_template(g, ('^%s%%.lua$'):format(name), [[?syntax error?]])
+    luadir = treegen.prepare_directory(g, {('%s.lua'):format(name)})
+    fio.chdir(luadir)
 end)
 
 g.after_all(function(g)
     treegen.clean(g)
+    fio.chdir(cwd)
 end)
 
 g.test_loader_error_handling = function()
-    package.path = ('%s/?.lua'):format(luadir)
-    local errlike = ("error loading module '%s' from file '%s/%s%%.lua':.+")
-                   :format(module, luadir, module)
-    t.assert_error_msg_matches(errlike, require, module)
+    local errmsg = errlike:gsub('%<(%w+)>', { NAME = name, DIR = luadir })
+    t.assert_error_msg_equals(errmsg, require, name)
 end
```

test/app-luatest/fix_lua_loader_error_handling_test.lua
```diff
@@ -0,0 +1,24 @@
...
+end)
+
+g.after_all(function(g)
+    treegen.clean(g)
+end)
+
+g.test_loader_error_handling = function()
+    package.path = ('%s/?.lua'):format(luadir)
+    local errlike = ("error loading module '%s' from file '%s/%s%%.lua':.+")
+                   :format(module, luadir, module)
```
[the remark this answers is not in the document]
igormunkin: I've reimplemented this part (see [above](https://github.com/tarantool/tarantool/pull/9511/files/a47393cacce1a56d22acdfe9dd19eead42e7cb55#r1464024558)).

src/lua/loaders.lua
```diff
@@ -58,36 +58,63 @@ local function traverse_path(path)
...
+    }
+}
+
+local cpath = {
+    package = function() return package.cpath end,
+    cwd = {
+        dot = gen_path_builder(searchroot, LIB_TEMPLATES),
+        rocks = gen_path_builder(searchroot, ROCKS_LIB_TEMPLATES, true),
+    }
+}
```
Totktonada, line 106: On the first glance I read it as `path` and `cpath` objects and only then understood that it is a way to group the data.

I would also highlight that we have:

* Module level variables with quite common names `path` and `cpath` that contain functions, not paths.
* Functions of the same kind on different nesting levels: `path.package` and `path.cwd.dot`.

The alternative would be, for example, the following.

```lua
local path_builders = {
    ['path.package'] = <...>,
    ['path.cwd.dot'] = <...>,
    ['path.cwd.rocks'] = <...>,
    <...>
}
```

Feel free to ignore. This is a matter of taste.

BTW, `{path,cpath}.package` goes after `{path,cpath}.cwd.{dot,rocks}` in the searchers/loaders order. It would be good to keep this order in these definitions too.

test/app-luatest/fix_lua_loader_error_handling_test.lua
Totktonada: Thanks! Personally, I would prefer more isolated runs (using `luatest.server`, for example), but I think that this solution is acceptable too.

src/lua/loaders.lua
```diff
@@ -86,152 +58,268 @@ local function traverse_path(path)
...
-    search_rocks_lib,
-    function(name) return package.searchpath(name, package.path) end,
-    function(name) return package.searchpath(name, package.cpath) end,
+    searchers['path.cwd.dot'],
+    searchers['cpath.cwd.dot'],
+    searchers['path.cwd.rocks'],
+    searchers['cpath.cwd.rocks'],
+    searchers['path.package'],
+    searchers['cpath.package'],
 }
```
Totktonada, line 249: A side note: maybe it worth to add a comment about inconsistency with the searchers/loaders machinery like one you've added for `searchers['croot.cwd']` (in fact, it is what I had to add in my patchset about override loaders).

Feel free to ignore.

src/lua/loaders.lua
```diff
@@ -331,43 +419,56 @@ local function getenv_boolean(varname, default)
...
+-- XXX: The default LuaJIT loaders are:
+-- [1]: preload
+-- [2]: package.path
+-- [3]: package.cpath
+-- [4]: croot
+--
+-- Tarantool loaders just maps the <searchers> table.
+for index, _ in ipairs(searchers) do
+    rawset(package.loaders, index, gen_legacy_loader(searchers, index))
+end
```
Totktonada, line 471: Maybe it worth to highlight that tarantool historically has different loader ordinals and they're preserved here (IOW, the default LuaJIT loader ordinals are NOT preserved).

Feel free to ignore.

Totktonada (approved):
OK for the patchset.

Thank you for the decent work!

src/lua/loaders.lua
Buristan: Yes, so feel free to ignore.

src/lua/loaders.lua
igormunkin: OK, added.
```diff
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index bdc1bd4d72..c175a1b843 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -465,7 +465,15 @@ rawset(searchers, 1, chain_searchers({
 -- [3]: package.cpath
 -- [4]: croot
 --
--- Tarantool loaders just maps the <searchers> table.
+-- Tarantool loaders just maps the <searchers> table:
+-- [1]: preload -> override -> builtin
+-- [2]: path.cwd.dot -> [path.app.dot]
+-- [3]: cpath.cwd.dot -> [cpath.app.dot]
+-- [4]: path.cwd.rocks -> [path.app.rocks]
+-- [5]: cpath.cwd.rocks -> [cpath.app.rocks]
+-- [6]: package.path
+-- [7]: package.cpath
+-- [8]: croot
 for index, _ in ipairs(searchers) do
     rawset(package.loaders, index, gen_legacy_loader(searchers, index))
 end
```

src/lua/loaders.lua
igormunkin: I agree this is the matter of taste, so I prefer to left this intact for now. Ignoring.

src/lua/loaders.lua
igormunkin: Added a comment. As for me, croot seems to be completely out of the context, since there are no Tarantool modules using this type of distribution.
```diff
diff --git a/src/lua/loaders.lua b/src/lua/loaders.lua
index c175a1b843..e9e50eea48 100644
--- a/src/lua/loaders.lua
+++ b/src/lua/loaders.lua
@@ -239,6 +239,15 @@ local function gen_legacy_loader(searchers, index)
     end
 end
 
+-- XXX: Though <package.search> looks like a handy public
+-- function, it's only the helper used for searching Tarantool
+-- C modules. Hence, despite it uses the same searchers as loaders
+-- for Lua modules do, it follows its own semantics ignoring
+-- preload, override, builtin modules, and application specific
+-- locations.
+-- FIXME: Lua paths look really odd in the list, since the
+-- function is used to load only C modules. However, these paths
+-- are left to avoid breaking change.
 local search_funcs = {
     searchers['path.cwd.dot'],
     searchers['cpath.cwd.dot'],
     ```

File: changelogs/unreleased/fix-lua-loader-error-hendling.md [hunk 13]
```diff
@@ -0,0 +1,4 @@
+## bugfix/lua
+
+* Fixed the misbehaviour with Lua error propagation when the error is raised
+  while loading the module via the Tarantool-specific loaders.
```
p7nov, line 4:
```suggestion
* Fixed the incorrect propagation of Lua errors raised while loading
  the module via the Tarantool-specific loaders.
```

p7nov (approved):

changelogs/unreleased/fix-lua-loader-error-hendling.md
igormunkin: Thanks for the suggestion! There was also a typo in the file name.

```diff
diff --git a/changelogs/unreleased/fix-lua-loader-error-handling.md b/changelogs/unreleased/fix-lua-loader-error-handling.md
new file mode 100644
index 0000000000..518f7268a7
--- /dev/null
+++ b/changelogs/unreleased/fix-lua-loader-error-handling.md
@@ -0,0 +1,4 @@
+## bugfix/lua
+
+* Fixed the incorrect propagation of Lua errors raised while loading
+  the module via the Tarantool-specific loaders.
diff --git a/changelogs/unreleased/fix-lua-loader-error-hendling.md b/changelogs/unreleased/fix-lua-loader-error-hendling.md
deleted file mode 100644
index 9ad041050c..0000000000
--- a/changelogs/unreleased/fix-lua-loader-error-hendling.md
+++ /dev/null
@@ -1,4 +0,0 @@
-## bugfix/lua
-
-* Fixed the misbehaviour with Lua error propagation when the error is raised
-  while loading the module via the Tarantool-specific loaders.
```

igormunkin:
The patchset is backported to release/3.0 in scope of a24ebea..1ddbfc2.

PR #9385, opened 2023-11-20
config: support TLS in iproto.listen and iproto.advertise.{peer, sharding}

ImeevMA:
This patch-set introduces new format for `iproto.listen` and `iproto.advertise.{peer, sharding}` and introduces support for TLS in these options.

Part of #8862

PR #9514, opened 2023-12-19
config: introduce vshard named identification

Serpentian:
Vshard now supports names instead of UUIDs in configuration. Let's use them and finally drop generating UUIDs in config.

From now on we can recover instances without explicitly passing UUIDs to configuration.

### DIFF
--- a/test/config-luatest/basic_test.lua
+++ b/test/config-luatest/basic_test.lua
@@ -6,6 +6,7 @@ local treegen = require('test.treegen')
 local justrun = require('test.justrun')
 local server = require('test.luatest_helpers.server')
 local helpers = require('test.config-luatest.helpers')
+local cbuilder = require('test.config-luatest.cbuilder')
 
 local g = helpers.group()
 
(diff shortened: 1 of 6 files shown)


test/config-luatest/basic_test.lua
```diff
@@ -569,3 +570,48 @@ g.test_loader_paths = function(g)
...
+    cfg_file = treegen.write_script(dir, 'config.yaml', yaml.encode(cfg))
+    g.server.config_file = cfg_file
+    -- Recovery process.
+    g.server:restart()
+    g.server:exec(function()
+        t.assert_equals(box.info.status, 'running')
+    end)
+
+    -- Clean everything.
+    g.server:drop()
```
Serpentian, line 615: It's strange, that all other tests doesn't clean `g` after themselves

Serpentian: Also, I don't really know, where to put that test. It's pretty basic)

Totktonada: > It's strange, that all other tests here doesn't clean `g` after themselves 

It is in the `after_each` hook, which is set by the `helpers.group()` call.

Such cleaning shouldn't be performed by test cases itself, because if the test case fails, it passes over the clean up. The hook doesn't.

Totktonada: > Also, I don't really know, where to put that test. It's pretty basic)

This `basic_test.lua` test is already pretty heavy and I'm a bit nervious about test-run's test timeout.

Personally I just pick up some 'main topic' of a change and if there is no such test, create it. Here we verify a step from the upgrade scenario, so `names_upgrade_test.lua` fits good.

Serpentian: I see, it's in `config-luatest/helpers.lua`. Will remove cleaning

Totktonada (approved):
Thanks!

I see no flaws. OK from my side.

Totktonada:
PR #9385 goes first and it conflicts a bit with this one (by `vshard_test.lua`).

ImeevMA (approved):
Thanks for the patch!

ochaplashkin (approved):

ylobankov (approved):

PR #9515, opened 2023-12-20
config: read cluster config parts from env/file

### LINKED ISSUE #9506: config: grab passwords from env/file
We should give a user ability to hide passwords from config.

@R-omk proposed a structure like the following.

```yaml
config:
  context:
    myvar_1:
      from: env
      env: MYVAR_1
    myvar_2:
      from: file
      file: path/to/file # relative to process.work_dir

credentials:
  users:
    myuser_1:
      password: '{{ context.myvar_1 }}'
    myuser_2:
      password: '{{ context.myvar_2 }}'
```

Totktonada:
Example:

```yaml
config:
  context:
    replicator_password:
      from: file
      file: secrets/replicator_password.txt
      rstrip: true
    client_password:
      from: file
      file: secrets/client_password.txt
      rstrip: true

credentials:
  users:
    replicator:
      password: '{{ context.replicator_password }}'
      roles: [replication]
    client:
      password: '{{ context.client_password }}'
      roles: [super]
```

The new `config.context` section allows a user to define its own
variables, whose content resides in an environment variable or a file.

The primary usage is to extract passwords from the configuration.

All the variables are defined in `config.context` section as keys. Each
key is a record with the following fields.

* `from`: either `env` or `file`
* `file`: a file path
* `env`: an environment variable name
* `rstrip`: whether to strip whitespace characters from the end of the
   data (it strips newlines too)

A relative file path is interpreted as relative to `process.work_dir`.

`from` is mandatory, `file` is mandatory when `from` is `file`, `env` is
mandatory when `from` is `env`, `rstrip` is optional.

If there is no given file or no given environment variable, an error of
a configuration applying is reported.

Fixes #9506


Totktonada:
It doesn't work with fifo files created from bash's `<(command)`: first read gives the content, but the second one gives an empty string. At the same time we can read the configuration twice on startup (c80b121544f60740fd2605dc2e15f86c84480c59), which assumes reading of all the user defined variables. And, for example, passwords in the `credentials` section are applied only on the second pass. So, they all are empty.

There were no goal to support fifo files, but I tested it out of curiosity.

It is possible to cache fifo files content. I'm NOT going to do it for now.

ochaplashkin (approved):

changelogs/unreleased/config-read-values-from-env-file.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Added ability to read a cluster config value from an environment variable or
+  a file (gh-9506).
```
xuniq, line 4:
```suggestion
* Added the ability to read a cluster config value from an environment variable or
  a file (gh-9506).
```

xuniq (approved):

ImeevMA (approved):
Thanks for the patch-set!

changelogs/unreleased/config-read-values-from-env-file.md
Totktonada: Thanks! Updated.

PR #9518, opened 2023-12-20
config: fix examples

ImeevMA:
This patch-set fixes iproto.listen and iproto.advertise.* in configs from doc/examples and adds tests that validate these configs.


Totktonada (approved):
Looks OK for me. Thank you!

ylobankov (approved):

PR #9520, opened 2023-12-20
build: call abort instead of _exit if ASAN found an error

ligurio:
Now LeakSanitizer silently prints a report about found error and testing is continue. Due to this found errors could be ignored in CI. The patch enables option `abort_on_error` in `ASAN_OPTIONS` [1] and after that Address Sanitizer and LeakSanitizer will call `abort()` instead of `_exit()` after printing the error report.

1. https://github.com/google/sanitizers/wiki/SanitizerCommonFlags

NO_CHANGELOG=build
NO_DOC=build
NO_TEST=build

### DIFF
--- a/test/config-luatest/log_wrapper_test.lua
+++ b/test/config-luatest/log_wrapper_test.lua
@@ -26,6 +26,7 @@ g.test_jsonify_table = function(g)
         exit_code = 0,
         stdout = '',
         stderr = 'foo: {"bar":"baz"}',
+        env = os.environ(),
     })
 end
 
--- a/.test.mk
+++ b/.test.mk
@@ -99,7 +99,8 @@ TEST_RUN_ENV_ASAN = ASAN=ON \
                     ASAN_OPTIONS=heap_profile=0:unmap_shadow_on_exit=1:$\
                                  detect_invalid_pointer_pairs=1:symbolize=1:$\
                                  detect_leaks=1:dump_instruction_bytes=1:$\
-                                 print_suppressions=0
+                                 print_suppressions=0:$\
+                                 abort_on_error=true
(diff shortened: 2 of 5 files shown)


test/config-luatest/log_wrapper_test.lua
```diff
@@ -26,6 +26,7 @@ g.test_jsonify_table = function(g)
         exit_code = 0,
         stdout = '',
         stderr = 'foo: {"bar":"baz"}',
+        env = os.environ(),
```
Totktonada, line 29: AFAIR, it shouldn't affect anything. Can you tell some details, how it changes the environment for the child process?

ligurio: No.

Totktonada: So, let's drop these changes.

ligurio: No.

ligurio: PR is not ready actually, don't waste your time on it. I'll let you know when it will be ready if you want review these changes. I suspect GH automatically requested review from you because you are a code owner (it is a test for config). Just ignore for now.

Totktonada: It was not marked as a draft, when I received a email with the review request.

It is OK to be in a middle of an investigation -- just say that you're there.

If you can confirm that this change actually has some effect, it is also useful information per se. It breaks my expectations and, so, interesting.

The tone of your responses saddens me, to be honest.

(BTW, refined-github allows to create a draft pull request without opening a usual PR first -- it adds a separate button on the pull request submission form.)

PR #9521, opened 2023-12-20
tools: remove tarantoolctl utility

### LINKED ISSUE #9443: Consider eliminating `tarantoolctl` stuff before 3.0.0 release
As we all know, the `tarantoolctl` script is going to be replaced by using the new [tt](https://github.com/tarantool/tt) utility. Due to this fact, it would be good to eliminate all stuff related to `tarantoolctl` before release 3.0.0.

0x501D:
tarantoolctl has been removed. Systemd, sysvinit and logrotate scripts based on it were also removed. All this functionality is covered by the tt utility.

Closes #9443


changelogs/unreleased/gh-9443-remove-tarantoolctl.md
```diff
@@ -0,0 +1,5 @@
+## feature/tools
+
+* `tarantoolctl` has been removed. Systemd, sysvinit and logrotate scripts based
```
locker, line 3: We probably need to add the breaking change marker.

locker (approved):
Please check that https://github.com/tarantool/tarantool-ee works after this change.

changelogs/unreleased/gh-9443-remove-tarantoolctl.md
0x501D: Added.

Totktonada: The main part of this change (remove the tool from the rpm/deb packages) was done in the 3.0 release.

It seems undesirable to include breaking changes after a major release. OTOH, this change technically breaking for users who build tarantool from sources.

Possible options:

* Discard or postpone till 4.0.
* Drop the 'breaking change' marker and include into 3.1 and 3.0.
* Include into 3.1 with the 'breaking change' marker.

I think that the latter is worse, because tarantool violates semver quite often in the past and we need to repair our reputation in this regard.

If we want to include the change into 3.0/3.1 we should consider it as non-breaking and so we should mark it as non-breaking. If we consider it as breaking, we shouldn't include it into 3.0/3.1.

Let's discuss?

LeonidVas: I don't understand why this is a "breaking change"?
`tarantoolctl` is not delivered to users and is not designed to work with tarantool 3.0. This PR is just а dead code cleanup.

0x501D: Agree. Just remove the "breaking change" label?

0x501D:
> Please check that [tarantool/tarantool-ee](https://github.com/tarantool/tarantool-ee) works after this change.

Seems to be OK: https://github.com/tarantool/tarantool-ee/pull/661

p7nov (approved):

LeonidVas (approved):

changelogs/unreleased/gh-9443-remove-tarantoolctl.md
0x501D: Changelog was updated.

ylobankov (approved):
LGTM

PR #9526, opened 2023-12-20
test: cover cluster upgrade with vshard

Serpentian:
This commit adds a test, which checks, that we're able to upgrade several vshard clusters without downtime. Previously it was impossible due to strict name validation on connect.

Let's test, that now everything is ok.

NO_DOC=test
NO_CHANGELOG=test


test/config-luatest/vshard_upgrade_test.lua
```diff
@@ -0,0 +1,222 @@
...
+
+local g = t.group()
+
+local has_vshard = pcall(require, 'vshard')
+
+--
+-- Test, that Tarantool upgrade with vshard happens without downtime:
+--
+--     1. Recover cluster from 2.11 xlogs with config.
+--     3. Schema upgrade.
```
Totktonada, line 16: 2. Vshard works before schema upgrade.

test/config-luatest/vshard_upgrade_test.lua
```diff
@@ -0,0 +1,222 @@
...
+    g.instance_3:eval(upgrade_exec)
+
+    -- Check, that vshard works before all names are set.
+    check_vshard(g, 2, 2001)
+    -- Wait for names to be applied.
+    local rs_prefix = 'replicaset-00'
+    g.instance_1:exec(wait_for_names, {g.instance_1.alias, rs_prefix .. '1'})
+    g.instance_2:exec(wait_for_names, {g.instance_2.alias, rs_prefix .. '1'})
+    g.instance_3:exec(wait_for_names, {g.instance_3.alias, rs_prefix .. '2'})
+    g.instance_4:exec(wait_for_names, {g.instance_4.alias, rs_prefix .. '2'})
```
Totktonada, line 218: > -- Check, that vshard works before all names are set.
> check_vshard(g, 2, 2001)

We can't say whether the check occurs before and after names are set. Especially due to retrying in the check.

Don't know, whether we should do anything in this regard.

test/config-luatest/vshard_upgrade_test.lua
```diff
@@ -0,0 +1,222 @@
...
+    -- Check, that vshard works before schema upgrade.
+    check_vshard(g, 1, 2000)
+
+    -- Upgrade schema.
+    local upgrade_exec = [[
+        box.ctl.wait_rw()
+        box.schema.upgrade()
+    ]]
+    g.instance_1:eval(upgrade_exec)
+    g.instance_3:eval(upgrade_exec)
```
Totktonada, line 209: We can also do a check with one upgraded and one non-upgraded replicaset.

Totktonada:
NB: It covers integration of tarantool's config module with vshard and mainly aims to verify correctness of changes in #9514 and https://github.com/tarantool/vshard/pull/458.

test/config-luatest/vshard_upgrade_test.lua
Serpentian: Yeah, this check is "flaky": it can occur before any names were applied, when some names were already applied, or even after all names were applied. We don't have errinjections in config module yet, and I didn't want to introduce one for the test.

But I want to be sure, that vshard works in all these cases, so I'd rather not remove it. The test is stable, so I suppose, it works in all cases

Totktonada: That's a way to go. I would just clarify it in a comment in the code.

test/config-luatest/vshard_upgrade_test.lua
Serpentian: Fixed the comment

ochaplashkin (approved):

test/config-luatest/vshard_upgrade_test.lua
```diff
@@ -0,0 +1,239 @@
...
+    local upgrade_exec = [[
+        box.ctl.wait_rw()
+        box.schema.upgrade()
+    ]]
+    g.instance_1:eval(upgrade_exec)
+    local rs_prefix = 'replicaset-00'
+    g.instance_1:exec(wait_for_names, {g.instance_1.alias, rs_prefix .. '1'})
+    g.instance_2:exec(wait_for_names, {g.instance_2.alias, rs_prefix .. '1'})
+    -- Check, that vshard works, when one shard has names, other - doesn't.
+    check_vshard(g, 2, 2001)
```
Totktonada, line 229: We're not verifying the intermediate state: after the upgrade, but before the names apply. We discussed it f2f with @Serpentian and agreed that it is worth to do.

Totktonada (approved):
Thank you!

I'm OK in general. Letf one comment by our today f2f discussion.

Serpentian:
The test is stabilized with https://github.com/tarantool/vshard/pull/462. Frankly speaking, I have no idea, how `_call is undefined` error during master search is related with the fixes in vshard, but there was really a bug there and now the test always works for me.

Serpentian:
After fixing vshard the test is 100% stable on my machine

Totktonada (approved):

test/config-luatest/vshard_upgrade_test.lua
Totktonada: It is done now.

PR #9527, opened 2023-12-20
config: update sharding.yaml

ImeevMA:
This patch updates example for vshard usage in config module.

NO_DOC=example update
NO_TEST=example update
NO_CHANGELOG=example update


Totktonada (approved):
No objections from my side. Thanks!

PR #9528, opened 2023-12-20
config: add missing field in test

ImeevMA:
This patch adds missing field in instance_config_schema_test.lua.

Follow-up #9506

NO_DOC=fix for test
NO_CHANGELOG=fix for test


Totktonada (approved):

Totktonada:
I'm going to push it without an extra review: the patch is trivial.

PR #9532, opened 2023-12-20
config: TLS options support -- follow up

Totktonada:
I accidentally discarded some changes made by @ImeevMA in PR #9385. The previous version of the patchset goes to the main branch as result. I'm very sorry.

This commit returns the code of the last version of the patchset.

The changes are described in https://github.com/tarantool/tarantool/pull/9385#discussion_r1430935585. Quoted here:

> 1. Reduce code duplication in the validation logic. It also
>    eliminates remaining (dead) code that interprets user@ and
>    user:pass@ syntax. And give better error messages in come cases
>    (comma separates URIs for example).
> 2. Forbid {} as an URI to simplify logic in instance_uri.
> 3. Refresh the comments.

Also added copying of an original URI table before a modification to prevent influence of the original URIs from config. See the second commit.

Part of #8862

### DIFF
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -70,7 +70,7 @@ function methods.names(self)
     }
 end
 
-local function instance_sharding(iconfig)
+local function instance_sharding(iconfig, instance_name)
     local roles = instance_config:get(iconfig, 'sharding.roles')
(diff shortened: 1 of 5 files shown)


ImeevMA (approved):
Thank you for the patch-set!

PR #9533, opened 2023-12-21
config: check vshard version

ImeevMA:
This patch sets requirements for vshard version that is supported by config module.

Part of #8862

NO_TEST=internal

@TarantoolBot document
Title: config: supported vshard version

The vshard version supported by the config module cannot be less than 0.1.25. Note, that vshard will only be loaded if any of instances have sharding storage role or sharding router role.

### DIFF
--- a/changelogs/unreleased/gh-8862-limit-vshard-version.md
+++ b/changelogs/unreleased/gh-8862-limit-vshard-version.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* The minimum supported `vshard` module version is now 0.1.25 (gh-8862).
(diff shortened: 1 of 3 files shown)


doc/examples/config/single.yaml
```diff
@@ -1,12 +1,15 @@
 credentials:
   users:
     guest:
-      roles: [super]
+      roles: [sharding]
```
Totktonada, line 4: It seems, these are forgotten experimental changes.

Totktonada (approved):

andreyaksenov (approved):

doc/examples/config/single.yaml
ImeevMA: True, fixed.

PR #9534, opened 2023-12-21
config: drop alert about credentials and RO mode

Totktonada:
Before this patch a replica adds an alert regarding inability to write new credentials from config on startup. In most cases these new credentials are applied on a master and the warning becomes obsolete. The problem is that it is not removed at this point. A manual `config:reload()` is needed to flush it.

It makes the warning more confusing than helpful. Let's remove it for now.

We can return the alert back in a future, but we should do it together with a logic that drops the alert if the needed data is received from a master.

Part of #8862

### DIFF
--- a/changelogs/unreleased/config-credentials-and-ro-alert.md
+++ b/changelogs/unreleased/config-credentials-and-ro-alert.md
@@ -1,0 +1,5 @@
+## bugfix/config
+
+* Removed a warning in `config:info().alerts` that appears on startup on a
+  replica if there are configured credentials to be written on the master
+  (gh-8862).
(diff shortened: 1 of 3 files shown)


changelogs/unreleased/config-credentials-and-ro-alert.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/config
+
+* Removed a warning in `config:info().alerts` that appears on startup on a
+  replica if there are configured credentials to be written on master
```
andreyaksenov, line 4:
```suggestion
  replica if there are configured credentials to be written on the master
```

andreyaksenov (approved):

changelogs/unreleased/config-credentials-and-ro-alert.md
Totktonada: Thanks! Fixed.

ImeevMA (approved):
Thank you for the patch!

PR #9537, opened 2023-12-21
config: start console before first box.cfg() call

Totktonada:
This commit changes the order of configuration appliers and move `console` before `box_cfg`. The `console` applier is adjusted to correctly interpret the configured socket file path before the first box.cfg().
    
The main reason to change the order is to allow a user to call `box.ctl.make_bootstrap_leader()` for a replicaset that starts in the `replication.boostrap_strategy: supervised` mode.

Part of #8862

### DIFF
--- a/changelogs/unreleased/config-start-console-before-box-cfg.md
+++ b/changelogs/unreleased/config-start-console-before-box-cfg.md
@@ -1,0 +1,5 @@
+## bugfix/config
+
+* Start the interactive console before the first `box.cfg()` call. This allows
+  you to issue the `box.ctl.make_bootstrap_leader()` command to a replica set
+  that uses the `supervised` bootstrap strategy (gh-8862).
--- a/src/box/lua/config/applier/console.lua
+++ b/src/box/lua/config/applier/console.lua
@@ -1,6 +1,7 @@
 local errno = require('errno')
 local console = require('console')
 local log = require('internal.config.utils.log')
+local instance_config = require('internal.config.instance_config')
 
 local function socket_file_to_listen_uri(file)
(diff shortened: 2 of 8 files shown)


ImeevMA (approved):
Thank you for the patch-set!

changelogs/unreleased/config-start-console-before-box-cfg.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/config
+
+* Start the interactive console before the first `box.cfg()` call. It allows to
+  issue the `box.ctl.make_bootstrap_leader()` command to a replicaset that uses
+  the `supervised` bootstrap strategy (gh-8862).
```
andreyaksenov, line 5:
```suggestion
* Start the interactive console before the first `box.cfg()` call. This allows you to
  issue the `box.ctl.make_bootstrap_leader()` command to a replica set that uses
  the `supervised` bootstrap strategy (gh-8862).
```

andreyaksenov (approved):

changelogs/unreleased/config-start-console-before-box-cfg.md
Totktonada: Fixed. Thanks!

Totktonada:
Added commit `config: use a relative path when possible` to eliminate a problem with a very long current working directory path. It allows the console to listen on a Unix file path in such a case. It is also required for our the Mac OS testing, which has a quite long `TMPDIR` variable value: testing instances are started inside it.

Also made instance names in the new tests shorter to create a client socket in a test in the situation with long `TMPDIR`.

ylobankov (approved):

PR #9542, opened 2023-12-22
Bump curl 8.5.0 and fix checkpatch module

ligurio:
Passed CI in Tarantool EE: https://github.com/tarantool/tarantool-ee/pull/616

### DIFF
--- a/changelogs/unreleased/bump-libcurl-to-8.5.0.md
+++ b/changelogs/unreleased/bump-libcurl-to-8.5.0.md
@@ -1,0 +1,3 @@
+## bugfix/build
+
+* Updated libcurl to version 8.5.0.
--- a/cmake/BuildLibCURL.cmake
+++ b/cmake/BuildLibCURL.cmake
@@ -179,7 +179,6 @@ macro(curl_build)
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_GSSAPI=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_LIBSSH=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_LIBPSL=OFF")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_OPENLDAP=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_WERROR=OFF")
     # CURL_ZSTD adds zstd encoding/decoding support. Tuning libcurl's build to
     # catch the symbols may require extra work.
--- a/cmake/BuildLibCURL.cmake
+++ b/cmake/BuildLibCURL.cmake
@@ -89,7 +89,11 @@ macro(curl_build)
     # Additionaly disable some more protocols.
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_SMB=ON")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_GOPHER=ON")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_CRYPTO_AUTH=ON")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_BASIC_AUTH=ON")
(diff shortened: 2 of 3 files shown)


andreyaksenov (approved):

changelogs/unreleased/bump-libcurl-to-8.5.0.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/build
+
+* Updated libcurl to version 8.5.0.
```
olegrok, line 3: Before release it's better to change existing note https://github.com/tarantool/tarantool/blob/master/changelogs/unreleased/bump-libcurl-to-8.4.0.md

To avoid:
* Updated libcurl to version 8.3.0.
* Updated libcurl to version 8.4.0.
* Updated libcurl to version 8.5.0.

In scope of single release

ligurio: Seems it will a first bump of libcurl in 3.1, so this comment is irrelevant now. I got you and will leave only the latest bump entry in changelog in a future.

cmake/BuildLibCURL.cmake
```diff
@@ -179,7 +179,6 @@ macro(curl_build)
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_GSSAPI=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_LIBSSH=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_LIBPSL=OFF")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_OPENLDAP=OFF")
```
olegrok, line 182: typo in commit message: relects -> reflects

cmake/BuildLibCURL.cmake
```diff
@@ -89,7 +89,11 @@ macro(curl_build)
     # Additionaly disable some more protocols.
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_SMB=ON")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_GOPHER=ON")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_CRYPTO_AUTH=ON")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_BASIC_AUTH=ON")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_BEARER_AUTH=ON")
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_DIGEST_AUTH=ON")
```
olegrok, line 94: Is it possible to enable this options? They could be useful and shouldn't require any dependencies as kerberos and aws

cmake/BuildLibCURL.cmake
```diff
@@ -206,6 +206,15 @@ macro(curl_build)
     # libcurl-d.a. We use this name below.
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCMAKE_DEBUG_POSTFIX=")
 
+    # Disables 'install' targets.
+    list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_DISABLE_INSTALL=OFF")
```
olegrok, line 210: DISABLE_INSTALL=OFF means ENABLE=ON. Was it expected?

cmake/BuildLibCURL.cmake
ligurio: Fixed, thanks!

cmake/BuildLibCURL.cmake
ligurio: I would not to do this here because it is a part of update CMake module after curl version bump. It could be done as a separate commit with described motivation.

cmake/BuildLibCURL.cmake
ligurio: Yes. The same value as in Curl's option, we use install target in CMake module [^1]

[^1]: https://github.com/tarantool/tarantool/blob/03c184d4da9be9e1431762989833a6762a704cfb/cmake/BuildLibCURL.cmake#L219

olegrok (approved):

Issue #585, opened 2014-10-14
uri.parse doesn't decode escaped characters


pgulutzan:
I can create a user with a password that is blank,
or contains ':' or '@' or '?' or ' ' etc.
Such a password won't be accepted later.
That's no good because a common recommendation for
passwords is: use special characters.

How to repeat:
box.cfg{listen=3301}
box.schema.user.create('Z3',{password='?'})
console=require('console')
console.connect('Z3:?@localhost:3301')

kostja:
You must escape such characters, but the problem is that our URI parser doesn't unescape escaped characters.

kostja:
Roman:

```tarantool> uri.parse('%20a@h')
---
- host: h
  login: '%20a'
...
```
as you can see the parser recognizes escaped characters but doesn't unescape them.
This needs to be fixed.

zloidemon:
Also, if you have an url which contains '{}"' and etc, you will have a parse error. For " that I converted symbols to %22 and etc

tmr-g:
> Roman:
> 
> ```
> tarantool> uri.parse('%20a@h')
> ---
> - host: h
>   login: '%20a'
> ...
> ```
> 
> as you can see the parser recognizes escaped characters but doesn't unescape them. This needs to be fixed.

It looks that `uri.parse` and `uri.format` must do the things coherently. So turning on auto unescaping in `parse` requires auto escaping in `format`.

This also mentioned in https://github.com/tarantool/tarantool/issues/7116
> parse(format(uri)) == uri

tmr-g:
The point where the escaping in `uri.format` is needed is peer URI in `box.info.replication`.

The test `box_info_uri_escape_test.lua` doesn't pass when escape is needed for now.

```lua
-- tags: parallel
local t = require('luatest')
local cbuilder = require('luatest.cbuilder')
local cluster = require('luatest.cluster')
local uri = require('uri')

local g = t.group(nil, {
    {custom_login = 'replicAtor'}, -- escape not needed
    {custom_login = 'replic@tor'}, -- escape needed
})

g.test_box_info_repl_peer_escaped = function(cg)
    local custom_login = cg.params.custom_login

    local cfg = cbuilder:new(nil)
        :set_global_option('credentials.users.replicator', nil)
        :set_global_option('credentials.users.'..custom_login, {password = 'secret', roles = {'replication'}})
        :set_global_option('iproto.advertise.peer.login', custom_login)
        :add_instance('i1', {})
[...]

PR #9543, opened 2023-12-22
test: add luatest.server helpers for integrity module

Lord-KA:
This patch introduces new helpers for luatest.server to support `box.ctl.iproto_lockdown()` in tests. The lockdown drops all connections including the current one, so the helper reconnects with the server.

Part of #585

NO_DOC=test helper upgrade
NO_CHANGELOG=see NO_DOC
NO_TEST=see NO_DOC

### DIFF
--- a/test/luatest_helpers/server.lua
+++ b/test/luatest_helpers/server.lua
@@ -190,4 +190,28 @@ function Server:start(opts)
     getmetatable(getmetatable(self)).start(self, opts)
 end
 
+-- Eval a string similar to `Server:eval()` and reconnect to the server
+-- in case the connection has been dropped.
+function Server:eval_reconnect(code, ...)
+    if self.net_box == nil then
+        error('net_box is not connected', 2)
+    end
+    local ok, err = pcall(self.net_box.eval, self.net_box, code, ...)
+    if not ok and tostring(err):find('Peer closed') then
+        self:wait_until_ready()
+    elseif not ok then
+        error(err)
+    end
+    return err
+end
+
+-- Prepare for iproto_lockdown caused by integrity check failure.
+function Server:prepare_integrity_lockdown()
+    -- Drop connection to the server ahead of time, before it is dropped
+    -- uncontrollingly in test-body. If the iproto_lockdown is already on,
+    -- it won't drop connections until turned off and on again.
+
+    self:eval_reconnect("box.ctl.iproto_lockdown(true)")
+end
+
 return Server


ochaplashkin (approved):

test/luatest_helpers/server.lua
```diff
@@ -190,4 +190,28 @@ function Server:start(opts)
...
+    if self.net_box == nil then
+        error('net_box is not connected', 2)
+    end
+    local ok, err = pcall(self.net_box.eval, self.net_box, code, ...)
+    if not ok and tostring(err):find('Peer closed') then
+        self:wait_until_ready()
+    elseif not ok then
+        error(err)
+    end
+    return err
```
Totktonada, line 205: It looks a bit strange: the error is either raised or returned depending of whether the reconnect was occured.

test/luatest_helpers/server.lua
```diff
@@ -190,4 +190,28 @@ function Server:start(opts)
     getmetatable(getmetatable(self)).start(self, opts)
 end
 
+-- Eval a string similar to `Server:eval()` and reconnect to the server
+-- in case the connection has been dropped.
+function Server:eval_reconnect(code, ...)
```
Totktonada, line 195: Is it really a general purpose function?

A function for retrying a request with an on demand reconnection seems meaningful (but requires request_timeout, time_budget, retry_count and so on). But the reconnect without retrying looks quite specific.

Maybe it is better to include it into a test.

test/luatest_helpers/server.lua
```diff
@@ -190,4 +190,28 @@ function Server:start(opts)
...
+end
+
+-- Prepare for iproto_lockdown caused by integrity check failure.
+function Server:prepare_integrity_lockdown()
+    -- Drop connection to the server ahead of time, before it is dropped
+    -- uncontrollingly in test-body. If the iproto_lockdown is already on,
+    -- it won't drop connections until turned off and on again.
+
+    self:eval_reconnect("box.ctl.iproto_lockdown(true)")
+end
```
Totktonada, line 215: It is related to one specific feature. How about including it into a test or into a test helper inside the given test suite?

test/luatest_helpers/server.lua
Lord-KA: Right, renamed `err` to `res`.

Lord-KA:
The helpers are moved to EE, so superseded by PR https://github.com/tarantool/tarantool-ee/pull/610.

PR #9544, opened 2023-12-24
box: support TT_* uri env vars with query params

### LINKED ISSUE #9539: TT_LISTEN and TT_REPLICATION are parsed incorrectly if a query parameter is used
The [documentation](https://www.tarantool.io/en/doc/latest/enterprise/security/#traffic-encryption) provides the following example.

```sh
export TT_LISTEN="localhost:3301?transport=ssl&ssl_cert_file=/path_to_cert_file&ssl_key_file=/path_to_key_file"
```

It doesn't work and both Community Edition and Enterprise Edition are affected.

The reason is that it contains `=` and the parsing code in `load_cfg.lua` interprets it as a key-value separator in a map.

The following example should work on CE, but it doesn't.

```sh
export TT_LISTEN=localhost:3301?transport=plain
```

Tarantool doesn't listen on the given address.

Totktonada:
`TT_LISTEN` and `TT_REPLICATION` environment variables were interpreted by `box.cfg()` in a confusing way if query parameters with values are present. For example, `localhost:3301?transport=plain` was interpreted as the following map: `{['localhost:3301?transport'] = 'plain'}`. Later, `box.cfg()` looks into this map for known URI fields like `login`, `password`, `uri`, `host`, `service` and so on. It found nothing and doesn't start a listening socket.

The reason of such a behaviour is that the environment value is interpreted as a mapping in the `key=value,key=value` format, because there is `=` in it.

The patch changes this behavior for an `key=value,key=value` environment variable that contains `?` in a key: now such a value is not interpreted as a mapping.

Note: Everything said above is also applicable to the so called multilisten case: when several URIs are defined in the environment variable. The following URI list is interpreted correctly now.

```sh
export TT_LISTEN=localhost:3301?transport=plain,localhost:3302?transport=plain
```

Note 2: Examples are given with the `plain` transport, which is default, but the query parameters are the way to define TLS options. They're supported in Tarantool Enterprise Edition, see https://www.tarantool.io/en/doc/latest/enterprise/security/#traffic-encryption.

Fixes #9539

### DIFF
--- a/changelogs/unreleased/gh-9539-box-cfg-env-vars-with-query-params.md
+++ b/changelogs/unreleased/gh-9539-box-cfg-env-vars-with-query-params.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Now `box.cfg()` correctly interprets the `TT_LISTEN` and `TT_REPLICATION`
+  environment variables with query parameters (gh-9539).
--- a/src/box/lua/load_cfg.lua
+++ b/src/box/lua/load_cfg.lua
@@ -1209,9 +1209,12 @@ local function get_option_from_env(option)
     if param_type:find('table') and (raw_value:startswith('{') or
                                      raw_value:startswith('[')) then
         return json.decode(raw_value)
-    elseif param_type:find('table') and raw_value:find('=') then
+    end
+
+    if param_type:find('table') and raw_value:find('=') then
         assert(not param_type:find('boolean'))
         local res = {}
+        local contains_uri = false
         for _, v in ipairs(raw_value:split(',')) do
             local eq = v:find('=')
             if eq == nil then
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-9539-box-cfg-env-vars-with-query-params.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* Now `box.cfg()` correctly interprets `TT_LISTEN` and `TT_REPLICATION`
```
andreyaksenov, line 3:
```suggestion
* Now `box.cfg()` correctly interprets the `TT_LISTEN` and `TT_REPLICATION`
```

andreyaksenov (approved):

src/box/lua/load_cfg.lua
```diff
@@ -1209,6 +1222,28 @@ local function get_option_from_env(option)
     if param_type:find('table') and (raw_value:startswith('{') or
                                      raw_value:startswith('[')) then
         return json.decode(raw_value)
+    elseif param_type:find('uri_list') then
```
locker, line 1225: Instead of introducing the new pseudo-type `uri_list`, what about *not* interpreting a string like `k1=v1,k2=v2,...` as a map in case `ki` contains `?`? I doubt anyone would use `?`.

ochaplashkin (approved):

changelogs/unreleased/gh-9539-box-cfg-env-vars-with-query-params.md
Totktonada: Thanks! Done.

src/box/lua/load_cfg.lua
Totktonada: Good idea! I've implemented it. Please, take a look.

locker (approved):

Totktonada:
I'll backport it to 2.11 and 2.10 later.

Totktonada:
Backport to 2.11: PR #9564.

Backport of the test to 2.10: PR #9565 (this branch is not affected by the problem, the fix is not needed).

PR #9546, opened 2023-12-25
config: rework replication.peers

### LINKED ISSUE #9541: Update `replications.peers` in config module
All URIs in the config module have been redesigned except `replication.peers`. We can update the URI representation of the option, but it actually makes more sense to use instance names instead of URIs.

ImeevMA:
After this patch, replication.peers will accept a list of instance names instead of a list of URIs.

Closes #9541

@TarantoolBot document
Title: replication.peers config option

The replication.peers option accepts a list of instance names that will be set as upstreams. If the option is not nil, it cannot be empty. For an instance, the option can only contain the names of the instances of the replicaset to which the instance belongs.

Currently, if the option does not contain all instances of a replicaset, except maybe one instance, the instance `bootstrap_strategy` option must be set to `config` and the instance replicaset `bootstrap_leader` option must be one of the instances from the `replication.peers`.

### DIFF
--- a/changelogs/unreleased/gh-9541-change-replication-peers.md
+++ b/changelogs/unreleased/gh-9541-change-replication-peers.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* The `replication.peers` option now accepts a list of instance names (gh-9541).
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -5,23 +5,10 @@ local snapshot = require('internal.config.utils.snapshot')
 local schedule_task = fiber._internal.schedule_task
 local mkversion = require('internal.mkversion')
 
-local function peer_uris(configdata)
-    local peers = configdata:peers()
-    if #peers <= 1 then
-        return nil
-    end
-
-    local names = configdata:names()
-    local err_msg_prefix = ('replication.peers construction for instance %q ' ..
-        'of replicaset %q of group %q'):format(names.instance_name,
-        names.replicaset_name, names.group_name)
-
-    -- Is there a peer in our replicaset with an URI suitable to
-    -- connect (except ourself)?
-    local has_upstream = false
-
-    local uris = {}
-    for _, peer_name in ipairs(peers) do
+local function auto_upstreams(configdata)
+    local all_peers = configdata:peers()
+    local peers = {}
+    for _, peer_name in ipairs(all_peers) do
         local iconfig_def = configdata._peers[peer_name].iconfig_def
         local is_anon = instance_config:get(iconfig_def, 'replication.anon')
         -- Don't use anonymous replicas as upstreams.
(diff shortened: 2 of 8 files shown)


ochaplashkin (approved):

changelogs/unreleased/gh-9541-change-replication-peers.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* The `replication.peers` option now accepts list of instance names (gh-9541).
```
andreyaksenov, line 3:
```suggestion
* The `replication.peers` option now accepts a list of instance names (gh-9541).
```

andreyaksenov (approved):

changelogs/unreleased/gh-9541-change-replication-peers.md
ImeevMA: Thanks, fixed!

ImeevMA:
@Totktonada  I pushed some changes:
```diff
diff --git a/changelogs/unreleased/gh-9541-change-replication-peers.md b/changelogs/unreleased/gh-9541-change-replication-peers.md
index 437aa65d1..1d7c0cced 100644
--- a/changelogs/unreleased/gh-9541-change-replication-peers.md
+++ b/changelogs/unreleased/gh-9541-change-replication-peers.md
@@ -1,3 +1,3 @@
 ## feature/config
 
-* The `replication.peers` option now accepts list of instance names (gh-9541).
+* The `replication.peers` option now accepts a list of instance names (gh-9541).
diff --git a/test/config-luatest/config_test.lua b/test/config-luatest/config_test.lua
index 58e2add97..199d93137 100644
--- a/test/config-luatest/config_test.lua
+++ b/test/config-luatest/config_test.lua
@@ -1452,6 +1452,8 @@ g.test_peers_success = function()
     g.server_2:wait_until_ready()
     g.server_3:wait_until_ready()
 
-    local replication = g.server_1:eval('return box.cfg.replication')
-    t.assert_equals(replication, {{uri = "unix/:./instance-001.iproto"}})
+    local exp = {{uri = "unix/:./instance-001.iproto"}}
+    t.assert_equals(g.server_1:eval('return box.cfg.replication'), exp)
+    t.assert_equals(g.server_2:eval('return box.cfg.replication'), exp)
+    t.assert_equals(g.server_3:eval('return box.cfg.replication'), exp)
 end
```

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -32,24 +19,50 @@ local function peer_uris(configdata)
...
+                'iproto.listen URI suitable to create a client socket',
+                err_msg_prefix, peer_name)
         end
+        table.insert(uris, uri)
     end
 
-    if not has_upstream then
+    if next(uris) == nil then
         error(('%s: no suitable peer URIs found'):format(err_msg_prefix), 0)
     end
```
Totktonada, line 67: I need to think about this check a bit more.

The logic behind the former check is the following: it is OK while any node has at least one upstream (except itself). The check is not performed for a single-node replicaset. It is also not performed for a user provided list of URIs.

It has the following problem.

* It doesn't allow to bootstrap a replicaset with one non-anonymous instance and all the other instances as anonymous replicas.

Next, if we're going to use instance names instead of URIs, we can verify a user-provided list of instances.

The 'has at least one upstream' check is insufficient on a replicaset bootstrap (it is still possible to bootstrap two separate replicasets) and, at the same time, may be too tight during a maintenance of an existing replicaset (we may want to temporary stop replication to the given instance).

src/box/lua/config/instance_config.lua
```diff
@@ -1404,13 +1404,24 @@ return schema.new('instance_config', schema.record({
         }, {
             default = 'off',
         }),
-        -- XXX: needs more validation
         peers = schema.array({
```
Totktonada, line 1407: We're out of 3.0 feature window now, so we shouldn't change `replication.peers` format/logic. I propose to add `replication.upstream` (or `replication.upstreams`?) that accepts a list of instance names.

src/box/lua/config/instance_config.lua
```diff
@@ -1404,13 +1404,24 @@ return schema.new('instance_config', schema.record({
...
             items = schema.scalar({
                 type = 'string',
             }),
-            box_cfg = 'replication',
             default = box.NULL,
+            validate = function(data, w)
+                if #data == 0 then
+                    w.error('replication.peers can only be nil or non-empty')
+                end
```
Totktonada, line 1415: I think it is OK to temporary disable replication to the given instance during some maintenance task.

ImeevMA:
We decided to not rework `replication.peers` for now.

Issue #9547, opened 2023-12-25
Builtin http client returns nothing on connection to etcd via TLS


ligurio:
**Bug description**

builtin http client returns nothing on connection to etcd via TLS

* OS: Linux
* OS Version: 22.04 x86_64
* Architecture: amd64

**Steps to reproduce**

Run etcd-client tests with Tarantool CE/EE with curl 8.4.0.

**Actual behavior**

```
3) integration.general.tls:true.test_watch_works_on_disconnect                                                                          
...ee/src/lua/etcd-client/test/integration/general_test.lua:1260: expected: a value evaluating to true, actual: nil                     
stack traceback:                                                                                                                        
        ...ee/src/lua/etcd-client/test/integration/general_test.lua:1260: in function 'integration.general.tls:true.test_watch_works_on_
disconnect'                                                                                                                             
        ...                                                                                                                             
        [C]: in function 'xpcall'                                                                                                       
Captured stdout:  
```

**Expected behavior**

Passed tests in etcd-client suite.

- Full log of runing the whole testsuite:  [log.txt](https://github.com/tarantool/tarantool/files/13765622/log.txt)
- Disabled TLS (passed), ` integration.general.tls:false.test_watchwait`: [tls_disable.txt](https://github.com/tarantool/tarantool/files/13765659/tls_disable.txt)
- Enabled TLS (failed), ` integration.general.tls:true.test_watchwait`: [tls_enable.txt](https://github.com/tarantool/tarantool/files/13765662/tls_enable.txt)

ligurio:
Running `integration.general.tls:true.test_watchwait` with patch below and logging in `debug` verbosity level

```diff
diff --git a/test/integration/general_test.lua b/test/integration/general_test.lua
index 60d794d..941d74e 100644
--- a/test/integration/general_test.lua
+++ b/test/integration/general_test.lua
@@ -1,7 +1,9 @@
 local etcd_utils = require('etcd-client.utils')
 local etcd_txn = require('etcd-client.txn')
 local etcd_client = require('etcd-client')
+local log = require('log')
 local t = require('luatest')
+local luatest_helpers = require('luatest.helpers')
 local test_utils = require('test.utils')
 
 local pgroup = t.group(nil, {
@@ -1184,13 +1186,18 @@ pgroup.test_watchwait = function(g)
     t.skip_if(not test_utils.is_httpc_stream_supported(),
         'Tarantool does not support streams yet')
 
+    log.cfg{ level='debug' }
[...]

PR #9551, opened 2023-12-25
[3.0] httpc: use http 1.1 by default

ligurio:
Commit 0ea1ba877b8e ("third_party: update libcurl from 8.3.0 to 8.4.0") bumps libcurl version and bringed a regression: http client returns nothing using connection with HTTP/2 protocol. The problem could be reproduced by etcd-client, see description in issue [1]. The problem in http client is due to commit "h2: testcase and fix for pausing h2 streams" [2] in libcurl. With reverted commit in libcurl the problem is gone.

To work around the issue, commit sets HTTP 1.1 as default protocol in http client.

Needed for #9547

1. https://github.com/tarantool/tarantool/issues/9547
2. https://github.com/curl/curl/commit/6b9a591bf7d82031f463373706d7de1cba0adee6

NO_CHANGELOG=workaround
NO_DOC=workaround
NO_TEST=etcd client's tests in Tarantool EE

### DIFF
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -214,6 +214,8 @@ httpc_request_new(struct httpc_env *env, const char *method,
 			 curl_easy_header_cb);
 	curl_easy_setopt(req->curl_request.easy, CURLOPT_NOPROGRESS, 1L);
 	curl_easy_setopt(req->curl_request.easy, CURLOPT_NOSIGNAL, 1L);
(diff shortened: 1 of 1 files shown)


olegrok (approved):
Thanks for the fix. I guess it can be accepted as temporary solution (currently root cause of this issue is not clear).

sergos (approved):

PR #9560, opened 2023-12-27
Fckxorg/ci integration

fckxorg:


fckxorg:
All workflows from integration.yml are red for now, but that's because the reusable_build.yml is being taken from master. There are a few important changes that were made in that workflow. Everything should be just fine after the merge.

igormunkin (approved):
**[PATCH 1/2] ci: disable the reusable run for some workflows**

@fckxorg, thanks for the patch! LGTM as trivial.

igormunkin (approved):
**[PATCH 2/2] ci: fix step parameters for reusable runs**

@fckxorg, thanks for the patch! The changeset is quite clear if one reads the commit message, so LGTM.

However, I still have a couple of questions:
1. Is it possible the workflow concurrency group pattern will clash again, if the particular workflow is copy-pasted to another repo? If so, it looks worthing to append repo name to the concurrency group name.
2. In case the particular workflow is backported to the other long-term branches, the hardcoded `master` reference should be updated. Moreover, `release` argument has no effect in tarantool/luajit repository workflows. So, the question is "what's next"?

ylobankov (approved):

PR #9561, opened 2023-12-27
ci: rm `test-sdk` job from submodule_update.yml

ylobankov:
The approach to build SDK since 3.0 release series is changed and will be under development for some period of time. So the `test-sdk` job is just irrelevant now and expected to fail. Let's remove the job until it is clear how to implement the integration testing in the scope of the new approach for SDK build.


ochaplashkin (approved):

ylobankov:
Cherry picked to 3.0.

PR #9567, opened 2023-12-28
replication: recovering from long WAL

### LINKED ISSUE #9094: A replica may timeout while master is scanning xlogs

yanshtunder:
As soon as replica subscribes, relay starts scanning the first requested WAL file to find the position from which to start replication. If a replica connects and needs data from the end of a really long WAL, recovery will read up to the needed position without yields, making relay disconnect by `replication_disconnect_timeout`. This issue was already found and fixed before: ee6de02. To fix this problem, make sure that when recovering, a yield occurs after `N=32` rows in WAL. In this case, the recovery time from WAL will increase by approximately 40 percent.

Closes #9094

NO_DOC=bugfix


changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
```diff
@@ -0,0 +1,5 @@
+# bugfix/replication
+
+* Now Tarantool sends heartbeats during WAL scan when starting replication
+  to a freshly subscribed replica. This prevents relay from timing out when
+  the replica needs rows from the end of a long `.xlog` (gh-9094).
```
andreyaksenov, line 5:
```suggestion
* Now Tarantool sends heartbeats during WAL scanning when starting replication
  to a freshly subscribed replica. This prevents the relay from timing out when
  the replica needs rows from the end of a long `.xlog` file (gh-9094).
```

andreyaksenov (approved):

changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
yanshtunder: Done

ochaplashkin (approved):

src/box/wal.h
```diff
@@ -68,7 +68,7 @@ enum {
 	 * applied from WAL. It allows not to block the event
 	 * loop for the whole recovery stage.
 	 */
-	WAL_ROWS_PER_YIELD = 1 << 15,
+	WAL_ROWS_PER_YIELD = 1 << 5,
```
Gerold103, line 71: Why?

src/box/relay.cc
```diff
@@ -987,12 +1044,21 @@ relay_subscribe_f(va_list ap)
 	if (!relay->replica->anon)
 		trigger_add(&relay->r->on_close_log, &on_close_log);
 
+	/*
+	 * When recovering from a long WAL, fiber_yield() will occur after
+	 * N rows and the current fiber will be called.
+	 */
+	char name[FIBER_NAME_MAX];
+	snprintf(name, sizeof(name), "%s:%s", fiber()->name, "uptime");
```
Gerold103, line 1052: Second `%s` can be just replaced with `uptime`.

src/box/relay.cc
```diff
@@ -166,11 +166,11 @@ struct relay {
...
 	 * as term bumps from TX with PROMOTE rows from WAL.
 	 */
-	struct cbus_endpoint tx_endpoint;
+	struct cbus_endpoint tx_endpoint[2];
 	/** A pipe from 'relay' thread to 'tx' */
-	struct cpipe tx_pipe;
+	struct cpipe tx_pipe[2];
 	/** A pipe from 'tx' thread to 'relay' */
-	struct cpipe relay_pipe;
+	struct cpipe relay_pipe[2];
```
Gerold103, line 173: Why do you need a second endpoint for that? I can't see the underlying problem why isn't the first endpoint being handled properly. I see that `relay_yield_and_send_heartbeat()` should be called, so what is the bug? Is the problem that we do not get notifications from the TX thread because we do not call `cbus_process()`? It doesn't look like you need a new endpoint then.

For example, we could create a separate fiber which runs `cbus_loop(tx_endpoint)` and remove `cbus_process()` from all other places. Would it work?

Adding a new endpoint will only make the code more complex due to more things to take care of and will eventually introduce races when we push different stuff into different endpoints and won't have any delivery order guarantees.

Gerold103 (commented):
Thanks for the patch!

src/box/relay.cc
```diff
@@ -581,7 +585,9 @@ relay_status_update(struct cmsg *msg)
 	struct relay_status_msg *status = (struct relay_status_msg *)msg;
 	struct relay *relay = status->relay;
 	relay->tx_seen_time = ev_monotonic_now(loop());
-	relay_check_status_needs_update(relay);
+	if (relay->tx.is_uptime) {
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

src/box/relay.cc
yanshtunder: Fixed.

src/box/relay.cc
yanshtunder: The first solution that came to mind.

> Adding a new endpoint will only make the code more complex due to more things to take care of and will eventually introduce races ...

Yes, that's true.

yanshtunder: Done

src/box/wal.h
yanshtunder: `WAL_ROWS_PER_YIELD = 1 << 5`  leave only for the test by entering a new errinj.

changelogs/unreleased/gh-9094-recovery-from-long-WAL.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/replication
```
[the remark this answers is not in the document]
yanshtunder: Fixed.

PR #9569, opened 2023-12-28
Update httpc default client interface and add test isolation

ligurio:


src/lua/httpc.lua
```diff
@@ -743,17 +743,30 @@ local this_module = {
...
-this_module.curl = http_default.curl
+local this_module_mt = {
+    __index = function(self, key, _value)
+        if this_module_raw[key] ~= nil then
+            return self[key]
+        end
+        local httpc_method = http_default[key]
+        if http_default ~= nil and
+           type(httpc_method) == 'function' then
+            return function(self, ...)
```
olegrok, line 759: Function creation for each request is really bad approach: it's slow and creates additional GC pressure. 

I don't see anything bad in previous interfaces. There is no any inconsistency - e.g. for spaces/indexes we require to use space:get() and return an error for single dot.

I didn't see anything who suffer from this

ligurio: >  There is no any inconsistency - e.g. for spaces/indexes we require to use space:get() and return an error for single dot.

If space/indexes implemented the same way, it doesn't mean that there is no inconsistency.
Inconsistency is in difference of using http methods for default and non-default client:

```lua
local http = require('http.client')
http.get('https://luajit.org/') -- This is fine.
http:get('https://luajit.org/') -- Error.

local httpc = http.new()
httpc.get('https://luajit.org/') -- Error.
httpc:get('https://luajit.org/') -- This is fine.
```

olegrok: > These arguments are valid. Don't use http methods via : in default client if you bother.

As a client how should I understand that one approach is more slow than another?
We don't need to add some unclear interfaces and makes our customers read source code to choose the best implementation.

Totktonada: The first argument check can be performed in the function that is generated in `http_default_wrap`. The same logic without per-request function generation.

ligurio: Summary of private conversations:
- there is an overhead for supporting two kinds of http clients: default http client [^1] in module and clients created via `new`
- default http client has been added from the beginning in `http.client` module import [^2]
- default http client was not officially documented [^3] but it is described at least in a cookbook [^4] and in source code of Cartridge (`cartridge/etcd2.lua`, `cartridge/feedback.lua`, `test/integration/etcd2_client_test.lua`) so we cannot remove it

1. https://github.com/tarantool/tarantool/blob/57c1f625ffb280aea950030962cff104b3f3a47d/src/lua/httpc.lua#L749-L757
2. https://github.com/tarantool/tarantool/commit/7e62ac79b9cfcaf7937041365d36a0ccde7a93a2
3. https://www.tarantool.io/en/doc/latest/reference/reference_lua/http/
4. https://www.tarantool.io/en/doc/latest/how-to/app/cookbook/#http-get-lua

ligurio:
Oleg, thanks for review! But I'll close PR as useless for now.

PR #9575, opened 2023-12-29
Replication graceful shutdown

nshy:
Part of https://github.com/tarantool/tarantool/issues/8423 related to replication.

### DIFF
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -675,7 +675,7 @@ applier_connect(struct applier *applier)
 	applier->version_id = greeting.version_id;
 
 	/* Don't display previous error messages in box.info.replication */
-	diag_clear(&fiber()->diag);
+	diag_clear(&applier->diag);
 
 	/*
 	 * Send an IPROTO_ID request if it's supported by the master.
@@ -2529,6 +2529,12 @@ applier_disconnect(struct applier *applier, enum applier_state state)
 	ibuf_reinit(&applier->ibuf);
 }
 
+static void
+applier_set_last_error(struct applier *applier, struct error *e)
+{
+	diag_set_error(&applier->diag, e);
+}
+
 static int
 applier_f(va_list ap)
 {
@@ -2594,9 +2600,10 @@ applier_f(va_list ap)
 			    tt_uuid_is_equal(&applier->uuid, &INSTANCE_UUID)) {
 				/* Connection to itself, stop applier */
 				applier_disconnect(applier, APPLIER_OFF);
-				return 0;
+				break;
 			} else if (e->errcode() == ER_LOADING) {
 				/* Autobootstrap */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_LOADING);
 				goto reconnect;
@@ -2609,6 +2616,7 @@ applier_f(va_list ap)
 				 * until they receive _cluster record of this
 				 * instance. From some third node, for example.
 				 */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_LOADING);
 				goto reconnect;
@@ -2618,6 +2626,7 @@ applier_f(va_list ap)
 				 * Join failure due to synchronous
 				 * transaction rollback.
 				 */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_LOADING);
 				goto reconnect;
@@ -2626,12 +2635,14 @@ applier_f(va_list ap)
 				   e->errcode() == ER_NO_SUCH_USER ||
 				   e->errcode() == ER_CREDS_MISMATCH) {
 				/* Invalid configuration */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_LOADING);
 				goto reconnect;
 			} else if (e->errcode() == ER_SYSTEM ||
 				   e->errcode() == ER_SSL) {
 				/* System error from master instance. */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_DISCONNECTED);
 				goto reconnect;
@@ -2643,15 +2654,17 @@ applier_f(va_list ap)
 				 * will be replaced by a normal tarantool node
 				 * sooner or later.
 				 */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier,
 						   APPLIER_DISCONNECTED);
 				goto reconnect;
 			} else {
 				/* Unrecoverable errors */
+				applier_set_last_error(applier, e);
 				applier_log_error(applier, e);
 				applier_disconnect(applier, APPLIER_STOPPED);
-				return -1;
+				break;
 			}
 		} catch (XlogGapError *e) {
 			/*
@@ -2676,33 +2689,38 @@ applier_f(va_list ap)
 			 * and in the meantime try to get newer changes from
 			 * node1.
 			 */
+			applier_set_last_error(applier, e);
 			applier_log_error(applier, e);
 			applier_disconnect(applier, APPLIER_LOADING);
 			goto reconnect;
 		} catch (FiberIsCancelled *e) {
 			if (!diag_is_empty(&applier->diag)) {
-				diag_move(&applier->diag, &fiber()->diag);
 				applier_disconnect(applier, APPLIER_STOPPED);
-				break;
+			} else {
+				applier_set_last_error(applier, e);
+				applier_disconnect(applier, APPLIER_OFF);
 			}
-			applier_disconnect(applier, APPLIER_OFF);
 			break;
 		} catch (SocketError *e) {
+			applier_set_last_error(applier, e);
 			applier_log_error(applier, e);
 			applier_disconnect(applier, APPLIER_DISCONNECTED);
 			goto reconnect;
 		} catch (SystemError *e) {
+			applier_set_last_error(applier, e);
 			applier_log_error(applier, e);
 			applier_disconnect(applier, APPLIER_DISCONNECTED);
 			goto reconnect;
 		} catch (SSLError *e) {
+			applier_set_last_error(applier, e);
 			applier_log_error(applier, e);
 			applier_disconnect(applier, APPLIER_DISCONNECTED);
 			goto reconnect;
 		} catch (Exception *e) {
+			applier_set_last_error(applier, e);
 			applier_log_error(applier, e);
 			applier_disconnect(applier, APPLIER_STOPPED);
-			return -1;
+			break;
 		}
 		/* Put fiber_sleep() out of catch block.
 		 *
@@ -2874,8 +2892,9 @@ applier_wait_for_state(struct applier_on_state *trigger, double timeout)
 		assert(applier->state == APPLIER_OFF ||
 		       applier->state == APPLIER_STOPPED);
 		/* Re-throw the original error */
-		assert(!diag_is_empty(&applier->fiber->diag));
-		diag_move(&applier->fiber->diag, &fiber()->diag);
+		assert(!diag_is_empty(&applier->diag));
+		diag_set_error(&fiber()->diag,
+			       diag_last_error(&applier->diag));
 		return -1;
 	}
 	return 0;
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -2124,7 +2124,9 @@ applier_thread_f(va_list ap)
 
 	cbus_loop(&thread->endpoint);
 
-	unreachable();
+	cbus_endpoint_destroy(&thread->endpoint, cbus_process);
+	cpipe_destroy(&thread->tx_pipe);
+	return 0;
 }
 
 /** Initialize and start the applier thread. */
@@ -2171,7 +2173,10 @@ applier_free(void)
 {
 	for (int i = 0; i < replication_threads; i++) {
 		struct applier_thread *thread = applier_threads[i];
-		cord_cancel_and_join(&thread->cord);
+		cbus_stop_loop(&thread->thread_pipe);
+		cpipe_destroy(&thread->thread_pipe);
+		if (cord_join(&thread->cord) != 0)
+			panic_syserror("applier cord join failed");
 		free(thread);
 	}
 	free(applier_threads);
@@ -2554,9 +2559,14 @@ applier_f(va_list ap)
 	bool was_anon = true;
 
 	/* Re-connect loop */
-	while (!fiber_is_cancelled()) {
+	while (true) {
 		FiberGCChecker gc_check;
 		try {
+			/*
+			 * Test cancel after reconnection sleep at the end
+			 * of the loop here to use catch for FiberIsCancelled.
+			 */
+			fiber_testcancel();
 			applier_connect(applier);
 			if (tt_uuid_is_nil(&REPLICASET_UUID)) {
 				/*
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -62,6 +62,18 @@ double replication_sync_timeout = 300.0; /* seconds */
 bool replication_skip_conflict = false;
 int replication_threads = 1;
 
+/**
+ * Fiber executing replicaset_connect. NULL if the function
+ * is not being executed.
+ */
+static struct fiber *replication_connect_fiber;
+
+/** Condition that replicaset_connect finished execution. */
+static struct fiber_cond replication_connect_cond;
+
+/** If set then replication shutdown is started. */
+static bool replication_is_shutting_down;
+
 bool cfg_replication_anon = true;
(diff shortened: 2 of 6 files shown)


src/box/replication.cc
```diff
@@ -219,21 +222,44 @@ replication_init(int num_threads)
...
 	/* The local instance is always part of the quorum. */
 	replicaset.healthy_count = 1;
 
 	applier_init();
 }
 
+void
+replication_shutdown(void)
+{
+	replication_is_shutdown = true;
```
locker, line 236: `replication_is_shutting_down` would probably be a better name. Also, it'd be consistent with the names used in iproto.

src/box/replication.cc
```diff
@@ -1160,6 +1197,8 @@ replicaset_connect(const struct uri_set *uris,
 	/* Now all the appliers are connected, update the replica set. */
 	replicaset_update(appliers, count, keep_connect);
 	appliers_guard.is_active = false;
+	replication_fiber = NULL;
```
locker, line 1200: Bad name: it's not a replication fiber; I think it's better call it after the function, i.e. `replicaset_connect_fiber`.

src/box/replication.cc
```diff
@@ -1160,6 +1197,8 @@ replicaset_connect(const struct uri_set *uris,
...
+	fiber_cond_signal(&replication_connect_cond);
```
locker, line 1201: Probably, better add another guard that wouldn't be disarmed on success to signal this condition variable.

src/box/replication.cc
```diff
@@ -1126,8 +1160,11 @@ replicaset_connect(const struct uri_set *uris,
 	while (!replicaset_is_connected(&state, appliers, count,
 					connect_quorum)) {
 		double wait_start = ev_monotonic_now(loop());
-		if (fiber_cond_wait_timeout(&state.wakeup, timeout) != 0)
+		if (fiber_cond_wait_timeout(&state.wakeup, timeout) != 0) {
+			if (fiber_is_cancelled())
+				diag_raise();
```
locker, line 1165: You could use `fiber_testcancel` helper.

src/box/replication.cc
```diff
@@ -1258,6 +1297,10 @@ void
 replicaset_sync(void)
 {
 	int quorum = replicaset_sync_quorum();
+	double deadline;
+
+	if (replication_is_shutdown)
+		goto shutdown;
```
locker, line 1303: I don't think this is necessary. AFAIU if replication is shutting down, the the loop condition below would make it break immediately anyway. For the same reason, I don't think we need to set `replication_fiber` in this function.

src/box/applier.cc
```diff
@@ -2842,7 +2844,8 @@ applier_on_state_f(struct trigger *trigger, void *event)
 	/* Wake up waiter */
 	fiber_cond_signal(&on_state->wakeup);
 
-	applier_pause(applier);
+	if (applier->state == on_state->desired_state)
+		applier_pause(applier);
```
locker, line 2848: Why not always call the trigger from the applier fiber then, even in case the fiber is stopped with `applier_stop`?

locker (changes requested):

src/box/applier.cc
```diff
@@ -2169,9 +2171,14 @@ applier_init(void)
...
 {
+	for (int i = 0; i < replication_threads; i++) {
+		cbus_stop_loop(&applier_threads[i]->thread_pipe);
+		cpipe_destroy(&applier_threads[i]->thread_pipe);
+	}
 	for (int i = 0; i < replication_threads; i++) {
 		struct applier_thread *thread = applier_threads[i];
-		cord_cancel_and_join(&thread->cord);
+		if (cord_join(&thread->cord) != 0)
+			panic_syserror("applier cord join failed");
```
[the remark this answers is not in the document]
nshy: I wanted to make it work a bit faster. But as `iproto_free` does not have such an optimization then let's remove it here.

```diff
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -2172,11 +2172,11 @@ void
 applier_free(void)
 {
        for (int i = 0; i < replication_threads; i++) {
-               cbus_stop_loop(&applier_threads[i]->thread_pipe);
-               cpipe_destroy(&applier_threads[i]->thread_pipe);
        }
        for (int i = 0; i < replication_threads; i++) {
                struct applier_thread *thread = applier_threads[i];
+               cbus_stop_loop(&thread->thread_pipe);
+               cpipe_destroy(&thread->thread_pipe);
                if (cord_join(&thread->cord) != 0)
                        panic_syserror("applier cord join failed");
                free(thread);
```

src/box/replication.cc
```diff
@@ -61,6 +61,9 @@ double replication_synchro_timeout = 5.0; /* seconds */
 double replication_sync_timeout = 300.0; /* seconds */
 bool replication_skip_conflict = false;
 int replication_threads = 1;
+struct fiber_cond replication_connect_cond;
+struct fiber *replication_fiber;
+bool replication_is_shutdown;
```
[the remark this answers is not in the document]
nshy: Fixed. 

```diff
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -63,7 +63,8 @@ bool replication_skip_conflict = false;
 int replication_threads = 1;
 struct fiber_cond replication_connect_cond;
 struct fiber *replication_fiber;
-bool replication_is_shutdown;
+/** If set then replication shutdown is started. */
+static bool replication_is_shutting_down;

 bool cfg_replication_anon = true;
 struct tt_uuid cfg_bootstrap_leader_uuid;
@@ -233,7 +234,7 @@ replication_init(int num_threads)
 void
 replication_shutdown(void)
 {
-       replication_is_shutdown = true;
+       replication_is_shutting_down = true;
        if (replication_fiber != NULL)
                fiber_cancel(replication_fiber);
        while (replication_fiber != NULL)
@@ -1065,7 +1066,7 @@ void
 replicaset_connect(const struct uri_set *uris,
                   bool connect_quorum, bool keep_connect)
 {
-       if (replication_is_shutdown)
+       if (replication_is_shutting_down)
                tnt_raise(ClientError, ER_SHUTDOWN);

        if (uris->uri_count == 0) {
```

src/box/replication.cc
```diff
@@ -219,21 +222,44 @@ replication_init(int num_threads)
...
 	diag_destroy(&replicaset.applier.diag);
 	trigger_destroy(&replicaset.on_ack);
 	trigger_destroy(&replicaset.on_relay_thread_start);
-
+	fiber_cond_destroy(&replication_connect_cond);
+	fiber_cond_destroy(&replicaset.applier.cond);
+	latch_destroy(&replicaset.applier.order_latch);
 	applier_free();
+
+	memset(&replicaset, 0, sizeof(replicaset));
```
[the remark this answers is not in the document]
nshy: It was intented as trashing. Fixed.

```diff
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -260,7 +260,7 @@ replication_free(void)
        latch_destroy(&replicaset.applier.order_latch);
        applier_free();

-       memset(&replicaset, 0, sizeof(replicaset));
+       TRASH(&replicaset);
 }
```

src/box/replication.cc
```diff
@@ -685,7 +711,8 @@ replica_on_applier_state_f(struct trigger *trigger, void *event)
 		 * to the same master, or the applier fiber
 		 * has been cancelled. Assume synced.
 		 */
-		replica_on_applier_sync(replica);
+		if (replica->applier_sync_state == APPLIER_CONNECTED)
+			replica_on_applier_sync(replica);
```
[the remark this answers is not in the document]
nshy: I guess we'd better to drop assertion in `replica_on_applier_sync`. The comment above the function is
still valid. But looks like the assertion that was added in [`replication: stop syncing if quorum cannot be formed`](https://github.com/tarantool/tarantool/commit/042c07dcc72099a36b4fdd6549504ebb3e596cc2#diff-032d66327afbb91d05d49b131da2727018ae6456857ecac57acfef636a30ee2dR221) does not account that we sync on appiler fiber cancel which can occur on probably any `replica->applier_sync_state`.

```diff
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -544,7 +544,6 @@ replica_on_applier_sync(struct replica *replica)
 {
        if (replica->applier_sync_state == APPLIER_SYNC)
                return;
-       assert(replica->applier_sync_state == APPLIER_CONNECTED);

        replica->applier_sync_state = APPLIER_SYNC;
        replicaset.applier.synced++;
@@ -712,8 +711,7 @@ replica_on_applier_state_f(struct trigger *trigger, void *event)
                 * to the same master, or the applier fiber
                 * has been cancelled. Assume synced.
                 */
-               if (replica->applier_sync_state == APPLIER_CONNECTED)
-                       replica_on_applier_sync(replica);
+               replica_on_applier_sync(replica);
                break;
        case APPLIER_STOPPED:
                /* Unrecoverable error. */
```

src/box/replication.cc
nshy: Done.

src/box/replication.cc
nshy: Fixed.

<details>
<summary>diff</summary>

```diff
diff --git a/src/box/replication.cc b/src/box/replication.cc
index 55120ffd3..73b6fe561 100644
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -61,8 +61,13 @@ double replication_synchro_timeout = 5.0; /* seconds */
 double replication_sync_timeout = 300.0; /* seconds */
 bool replication_skip_conflict = false;
 int replication_threads = 1;
+/**
+ * Fiber executing replicaset_connect. NULL if the function
+ * is not being executed.
+ */
+struct fiber *replication_connect_fiber;
+
+/** Condition that replicaset_connect finished execution. */
 struct fiber_cond replication_connect_cond;
-struct fiber *replication_fiber;
+
 /** If set then replication shutdown is started. */
 static bool replication_is_shutting_down;
 
@@ -235,9 +240,9 @@ void
 replication_shutdown(void)
 {
 	replication_is_shutting_down = true;
-	if (replication_fiber != NULL)
-		fiber_cancel(replication_fiber);
-	while (replication_fiber != NULL)
+	if (replication_connect_fiber != NULL)
+		fiber_cancel(replication_connect_fiber);
+	while (replication_connect_fiber != NULL)
 		fiber_cond_wait(&replication_connect_cond);
 
 	struct replica *replica;
@@ -1079,8 +1084,8 @@ replicaset_connect(const struct uri_set *uris,
 		tnt_raise(ClientError, ER_CFG, "replication",
 			  "too many replicas");
 	}
-	assert(replication_fiber == NULL);
-	replication_fiber = fiber();
+	assert(replication_connect_fiber == NULL);
+	replication_connect_fiber = fiber();
 	int count = 0;
 	struct applier *appliers[VCLOCK_MAX] = {};
 	auto appliers_guard = make_scoped_guard([&]{
@@ -1089,7 +1094,7 @@ replicaset_connect(const struct uri_set *uris,
 			applier_stop(applier);
 			applier_delete(applier);
 		}
-		replication_fiber = NULL;
+		replication_connect_fiber = NULL;
 		fiber_cond_signal(&replication_connect_cond);
 	});
 	for (; count < uris->uri_count; count++)
@@ -1196,7 +1201,7 @@ replicaset_connect(const struct uri_set *uris,
 	/* Now all the appliers are connected, update the replica set. */
 	replicaset_update(appliers, count, keep_connect);
 	appliers_guard.is_active = false;
-	replication_fiber = NULL;
+	replication_connect_fiber = NULL;
 	fiber_cond_signal(&replication_connect_cond);
 	uri_set_destroy(&replication_uris);
 	uri_set_copy(&replication_uris, uris);
```

</details>

src/box/replication.cc
nshy: Done.

```diff
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -1089,6 +1089,10 @@ replicaset_connect(const struct uri_set *uris,
        }
        assert(replication_connect_fiber == NULL);
        replication_connect_fiber = fiber();
+       auto connect_fiber_guard = make_scoped_guard([&]{
+               replication_connect_fiber = NULL;
+               fiber_cond_signal(&replication_connect_cond);
+       });
        int count = 0;
        struct applier *appliers[VCLOCK_MAX] = {};
        auto appliers_guard = make_scoped_guard([&]{
@@ -1097,8 +1101,6 @@ replicaset_connect(const struct uri_set *uris,
                        applier_stop(applier);
                        applier_delete(applier);
                }
-               replication_connect_fiber = NULL;
-               fiber_cond_signal(&replication_connect_cond);
        });
        for (; count < uris->uri_count; count++)
                appliers[count] = applier_new(&uris->uris[count]);
@@ -1204,8 +1206,6 @@ replicaset_connect(const struct uri_set *uris,
        /* Now all the appliers are connected, update the replica set. */
        replicaset_update(appliers, count, keep_connect);
        appliers_guard.is_active = false;
-       replication_connect_fiber = NULL;
-       fiber_cond_signal(&replication_connect_cond);
        uri_set_destroy(&replication_uris);
        uri_set_copy(&replication_uris, uris);
 }
```

src/box/replication.cc
nshy: Done.

```diff
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -1170,8 +1170,7 @@ replicaset_connect(const struct uri_set *uris,
                                        connect_quorum)) {
                double wait_start = ev_monotonic_now(loop());
                if (fiber_cond_wait_timeout(&state.wakeup, timeout) != 0) {
-                       if (fiber_is_cancelled())
-                               diag_raise();
+                       fiber_testcancel();
                        break;
                }
                timeout -= ev_monotonic_now(loop()) - wait_start;
```

src/box/replication.cc
nshy: Dropped changes in this function. Looks like this function will work as is because on stop applier is accounted for as sync. Also this function is not called before `replicaset_connect` which will fail after shutdown.

src/box/replication.cc
```diff
@@ -517,7 +552,6 @@ replica_on_applier_sync(struct replica *replica)
 {
 	if (replica->applier_sync_state == APPLIER_SYNC)
 		return;
-	assert(replica->applier_sync_state == APPLIER_CONNECTED);
```
locker, line 520: Instead of removing the assertion, better extend it with all possible conditions and add an explanatory comment.

src/box/applier.cc
```diff
@@ -2553,9 +2558,10 @@ applier_f(va_list ap)
 	bool was_anon = true;
 
 	/* Re-connect loop */
-	while (!fiber_is_cancelled()) {
+	while (true) {
 		FiberGCChecker gc_check;
 		try {
+			fiber_testcancel();
```
locker, line 2564: Do we really need this extra check? AFAIU `applier_connect()` already checks if the fiber is cancelled implicitly, via coio.

src/box/applier.cc
```diff
@@ -2840,7 +2859,8 @@ applier_on_state_f(struct trigger *trigger, void *event)
 	/* Wake up waiter */
 	fiber_cond_signal(&on_state->wakeup);
 
-	applier_pause(applier);
+	if (applier->state == on_state->desired_state)
+		applier_pause(applier);
```
locker, line 2863: Maybe, better make `applier_pause` and `applier_resume` no-op in case the fiber is in the 'stopped' or 'off' state? It'd make the code look more robust.

src/box/applier.cc
```diff
@@ -2890,6 +2911,7 @@ void
 applier_resume_to_state(struct applier *applier, enum applier_state state,
 			double timeout)
 {
+	assert(state != APPLIER_STOPPED && state != APPLIER_OFF);
```
locker, line 2914: What's wrong with these states? AFAIU this function should work just fine for them too.

locker (changes requested):

src/box/replication.cc
nshy: Discussed f2f and decided to leave as is because shutdown can happen at any applier state.

src/box/applier.cc
nshy: Discussed f2f and decided to leave as is because it is a good style to have fiber cancel test right after sleep. Also added a comment:

```diff
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -2562,6 +2562,10 @@ applier_f(va_list ap)
        while (true) {
                FiberGCChecker gc_check;
                try {
+                       /*
+                        * Test cancel after reconnection sleep at the end
+                        * of the loop here to use catch for FiberIsCancelled.
+                        */
                        fiber_testcancel();
                        applier_connect(applier);
                        if (tt_uuid_is_nil(&REPLICASET_UUID)) {
```

src/box/applier.cc
nshy: Discussed f2f and decided to drop this change as it not required for fixing the issue.

src/box/applier.cc
nshy: Dropped. This is remnant of the previous version of the patch where diag was stealed in `applier_wait_for_state`.

locker (approved):

PR #9585, opened 2024-01-10
box: remove static_assert() from field_def.h

ImeevMA:
This patch removes the static_assert(). This static_assert was necessary because after commit 037e2e449be8 ("sql: clean-up affinity from SQL source code") and before commit 078bcf00437f ("sql: remove implicit cast from comparison opcodes"), 4 bits of the p5 field of the struct VdbeOp were used to store the field type in VDBE comparison opcodes. After the commit 078bcf00437f ("sql: remove implicit cast from comparison opcodes"), these opcodes no longer need the field type, so this static_assert() is now unneeded.

### DIFF
--- a/src/box/field_def.h
+++ b/src/box/field_def.h
@@ -89,22 +89,6 @@ enum on_conflict_action {
 
 /** \endcond public */
 
-enum {
-	/**
-	 * This mask allows to store in VdbeOp.p5 operand of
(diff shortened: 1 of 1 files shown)


Gumix (approved):
> 4 bytes of the p5 field of the struct VdbeOp were used to store the field type in VDBE comparison opcodes

4 bits

locker (approved):

PR #9592, opened 2024-01-12
box: introduce fixed-size numeric field types

### LINKED ISSUE #9548: Introduce fixed-size numeric types
Tarantool has `unsigned` and `integer` types which can store 64-bit integer values. We don't have fixed-size integer types like `uint8_t` or `int16_t` because the MsgPack format used internally by Tarantool efficiently packs integer values, e.g. 129 would be encoded as a 2-byte sequence `\xcc\x81`, not as a 8-byte value.

However, sometimes it may be useful to have fixed-size numeric types. For example, one may want to store them unencoded in an array for effective scanning (like a column store). For this purpose, let's introduce the following new types to Tarantool:
 - `uint8`: integer in range [0 .. 255]
 - `int8`: integer in range [-128 .. 127]
 - `uint16`: integer in range [0 .. 65,535]
 - `int16`: integer in range [-32,768 .. 32,767]
 - `uint32`: integer in range [0 .. 4,294,967,295]
 - `int32`: integer in range [-2,147,483,648 .. 2,147,483,647]
 - `uint64`: integer in range [0 .. 18,446,744,073,709,551,615]
 - `int64`: integer in range [-9,223,372,036,854,775,808 .. 9,223,372,036,854,775,807]
 - `float32`: 32-bit floating point number
 - `float64`: 64-bit floating point number

The types shouldn't affect how data is stored in Tarantool's memtx or vinyl engines but they should enforce the limits when a tuple is inserted or updated.

Note, introduction of the new field types may require updating the SQL subsystem.

Gumix:
Closes #9548


andreyaksenov (approved):

src/box/field_def.c
```diff
@@ -155,6 +242,19 @@ field_type1_contains_type2(enum field_type type1, enum field_type type2)
...
+	 */
+	bool is_fixed_int = type1 == FIELD_TYPE_INT8 ||
+			    type1 == FIELD_TYPE_UINT8 ||
+			    type1 == FIELD_TYPE_INT16 ||
+			    type1 == FIELD_TYPE_UINT16 ||
+			    type1 == FIELD_TYPE_INT32 ||
+			    type1 == FIELD_TYPE_UINT32 ||
+			    type1 == FIELD_TYPE_INT64;
+	if (is_fixed_int)
+		return false;
```
locker, line 257: It should be possible to convert `uint16` to `uint32` or `uint64` without checking values. Same for signed integer types and, probably, even `float32` -> `float64`.

src/box/field_def.c
```diff
@@ -233,6 +333,12 @@ field_type_by_name(const char *name, size_t len)
 					       field_type_MAX);
 	if (field_type != field_type_MAX)
 		return field_type;
+	/*
+	 * `uint64' is missed in `field_type_strs', because it is an alias for
+	 * `unsigned'.
+	 */
+	if (len == strlen("uint64") && strncasecmp(name, "uint64", len) == 0)
+		return FIELD_TYPE_UINT64;
```
locker, line 341: Why make `uint64` an alias? Can't we have it as a separate type but allow conversion to/from `unsigned` without checking values?

src/box/tuple_format.c
```diff
@@ -1024,6 +1024,79 @@ tuple_format_required_fields_validate(struct tuple_format *format,
...
+	const char *min_str, *max_str, *val_str;
+
+	switch (field->type) {
+	case FIELD_TYPE_INT8:
+	case FIELD_TYPE_INT16:
+	case FIELD_TYPE_INT32:
+	case FIELD_TYPE_INT64: {
+		int64_t min = field_type_min_value[field->type];
+		int64_t max = field_type_max_value[field->type];
+		if (mp_type == MP_INT) {
```
locker, line 1052: I don't quite like this switch-case and the min/max constants defined for *all* types. The code would be more generalized if we used callbacks for checking values. However, it could affect performance. Need to discuss with @alyapunov.

locker (changes requested):

ochaplashkin (approved):

src/box/field_def.c
Gumix: Done (I tried to avoid duplicate entries in `enum field_type`).

src/box/tuple_format.c
Gumix: Discuss f2f and decided not to use callbacks.

src/box/field_def.c
```diff
@@ -164,7 +324,16 @@ field_type1_contains_type2(enum field_type type1, enum field_type type2)
...
+
+	bool type1_contains_type2 = (mp_type1 & mp_type2) == mp_type2;
+	if (!type1_contains_type2)
+		return false;
+
+	int64_t min1 = field_type_min_value[type1];
+	int64_t min2 = field_type_min_value[type2];
+	uint64_t max1 = field_type_max_value[type1];
+	uint64_t max2 = field_type_max_value[type2];
+	return min1 <= min2 && max1 >= max2;
```
locker, line 336: IMO it's better not to set the limits for non-fixed types and instead check the `is_fixed` flag here:

```C
if (field_type_is_fixed_signed[type2]) {
	return !field_type_is_fixed_signed[type1] ||
		(field_type_min_value[type1] <= field_type_min_value[type2] &&
		 field_type_max_value[type1] >= field_type_max_value[type2]);
}
if (field_type_is_fixed_unsigned[type2]) {
	return (!field_type_is_fixed_signed[type1] &&
		!field_type_is_fixed_unsigned[type1]) ||
		field_type_max_value[type1] >= field_type_max_value[type2];
}
```

Gumix: This condition doesn't work for:
```
    not ok 41 - int8 can store values of unsigned
    not ok 42 - uint8 can store values of unsigned
    not ok 43 - int16 can store values of unsigned
    not ok 44 - uint16 can store values of unsigned
    not ok 45 - int32 can store values of unsigned
    not ok 46 - uint32 can store values of unsigned
    not ok 47 - int64 can store values of unsigned
    not ok 141 - int8 can store values of integer
    not ok 143 - int16 can store values of integer
    not ok 145 - int32 can store values of integer
    not ok 147 - int64 can store values of integer
```
I've fixed it.

locker (approved):

PR #9598, opened 2024-01-16
config: fix missing and stuck alerts

### LINKED ISSUE #9574: config: delayed privilege grant/revoke doesn't clear an alert
Let's consider the following simple config.

```yaml
credentials:
  users:
    guest:
      roles: [super]
      privileges:
      - permissions: [read, write]
        spaces: [s]

iproto:
  listen:
  - uri: 'unix/:./var/run/{{ instance_name }}/tarantool.iproto'

app:
  file: app.lua

groups:
  group-001:
    replicasets:
      replicaset-001:
        instances:
          instance-001: {}
```

And run it with the following application (`app.lua`).

```lua
box.once('app', function()
    box.schema.space.create('s')
    box.space.s:create_index('pk')
end)
```

Next, start the instance.

```sh
./src/tarantool --name instance-001 --config config.yaml
```

And connect to the admin console.

```sh
tt connect ./var/run/instance-001/tarantool.control
```

Check the space and the privileges.

```yaml
# ./var/run/instance-001/tarantool.control> box.schema.user.info('guest')
---
- - - execute
    - role
    - public
  - - execute
    - role
    - super
  - - read,write
    - space
    - s
  - - session,usage
    - universe
    - 
...
```

`read, write` on space `s` is granted, everything is OK.

But,

```yaml
# ./var/run/instance-001/tarantool.control> require('config'):info()
---
- status: check_warnings
  meta: []
  alerts:
  - type: warn
    message: 'credentials.apply: space "s" hasn''t been created yet, ''box.schema.user.grant("guest",
      "read,write", "space", "s")'' will be applied later'
    timestamp: 2023-12-29T02:44:11.898258+0300
...
```

### LINKED ISSUE #9586: config: all except one delayed privilege alerts are missing
Let's consider the following config.

```yaml
credentials:
  users:
    guest:
      roles: [super]
      privileges:
      - permissions: [read, write]
        spaces: [s, t]
      - permissions: [execute]
        functions: [f, g]

iproto:
  listen:
  - uri: 'unix/:./var/run/{{ instance_name }}/tarantool.iproto'

groups:
  group-001:
    replicasets:
      replicaset-001:
        instances:
          instance-001: {}
```

Here we have privileges for spaces `s`, `t` and functions `f`, `g` to be granted at their creation.

However, the warning about delayed privilege granting is issued only for `f`.

```sh
$ tarantool --name instance-001 --config config.yaml
```

```sh
$ tt connect ./var/run/instance-001/tarantool.control
```

```yaml
# tarantool> require('config'):info()
# tarantool> config:info()
---
- status: check_warnings
  meta: []
  alerts:
  - type: warn
    message: 'credentials.apply: function "f" hasn''t been created yet, ''box.schema.user.grant("guest",
      "execute", "function", "f")'' will be applied later'
    timestamp: 2024-01-10T15:39:25.478785+0300
...
```

Looks like a regression from fa97cc0c12e9f9c051e7455c71d8be264a50d554.

Totktonada:
There are cases, when an alert is not reported, when it should (#9586). There are cases, when an alert is present, but its reason is already gone (#9574). These problems are fixed in this patchset. See details in the commit messages.

Fixes #9574
Fixes #9586


changelogs/unreleased/config-missing-alerts.md
```diff
@@ -0,0 +1,7 @@
+## bugfix/config
+
+* Fixed a bug when only one alert of a particular type is reported in
+  `config:info().alerts` instead of several ones. Examples of such alerts are
+  privilege grant delay due to a lack of a space/function/sequence and skipping
+  of non-dynamic `box.cfg()` option apply on the configuration reloading
```
andreyaksenov, line 6:
```suggestion
  of a non-dynamic `box.cfg()` option applies on the configuration reloading
```

changelogs/unreleased/config-stuck-privilege-granting-alert.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* An alert regarding delayed privilege granting is now cleared, when the
```
andreyaksenov, line 3:
```suggestion
* An alert regarding delayed privilege granting is now cleared when the
```

andreyaksenov (approved):

changelogs/unreleased/config-stuck-privilege-granting-alert.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
```
Lord-KA, line 1: Nit:
A few tipos?

> It is OK to have privileges for a space/function/sequence that **does not exist**

> successful privilege grant calls `config._aboard:drop()`.

src/box/lua/config/utils/aboard.lua
```diff
@@ -14,6 +14,11 @@ local function aboard_set(self, key, alert)
     self._alerts[key] = alert
 end
 
+local function aboard_add(self, alert)
```
Serpentian, line 17: Nit: it doesn't look good to add the new function, which basically does the same as `aboard_set`. Maybe we can pass key to `aboard_set` as a second option? If it's not passed, then we'll use `_next_key` automatically. Up to you, not a principal issue for me

Serpentian (approved):
Thank you for the patch! Looks great, just one minor comment

src/box/lua/config/applier/credentials.lua
```diff
@@ -525,17 +525,93 @@ end
...
+    -- What to do if we found an inconsistency that leads to one
+    -- of the three situations above?
+    --
+    -- The simplest solution is to issue several alerts for each
+    -- object: one for the 'read' permission, one for 'write'
+    -- permission and so on.
+    --
+    -- More complicated solution is to perform a search for
+    -- relevant alerts on successful privilege granting to modify
+    -- or drop a right alert.
```
Lord-KA, line 606: I don't see, how this approach is more complicated, the search in `_aboard:drop()` is still performed. I would say it is just less durable.

src/box/lua/config/applier/credentials.lua
```diff
@@ -525,17 +525,93 @@ end
...
+    --            spaces: [s]
+    --
+    --    credentials:
+    --      users:
+    --        myuser:
+    --          privileges:
+    --          - permissions: [read]
+    --            spaces: [s]
+    --          - permissions: [write]
+    --            spaces: [s]
```
Lord-KA, line 588: Nit:

Maybe skip `credentials: users: myuser:`, the comment is long enough and we don't really need a full config to understand what is going on. Feel free to ignore.

src/box/lua/config/applier/credentials.lua
```diff
@@ -525,17 +525,93 @@ end
...
+    -- (for example 'read' and 'write'). The warning will be
+    -- issued with the "box.schema.user.grant(<...>, 'read,write',
+    -- <...>)" key.
+    --
+    -- Now, let's assume that the space is created. We should drop
+    -- the warning and, so, we should generate exactly same key to
+    -- pass to the :drop() aboard method.
+    --
+    -- 1. Is it possible that the key to drop contains just 'read',
+    --    not 'read,write'? Seems unlikely.
```
Lord-KA, line 551: This comment is somewhat confusing to me. I feel that it should describe the guarantees that we rely on to provide the exact same warning message to the `aboard`. Instead it goes over some corner cases and solves them with a varying degree of confidence ("Seems, unlikely", "Seems, no").

The sync (that does the grants and revokes) is called in two scenarios:
1) on reload, when config is changed and all of such warnings are dropped anyway
2) on object appearance, when the config remains the same

The second one might be an issue, because an object can be renamed to a configurable one. At this point it might already have the permissions that are to be granted. Than, `privileges_action_f()` isn't called, because the diff is empty and no warnings are dropped.

Lord-KA (commented):
Thanks for the patchset, see my 8 comments below.

src/box/lua/config/applier/credentials.lua
Lord-KA: E.g. the config includes the following:

```yaml
privileges:
- permissions: [read, write]
  spaces: [myspace1]
```

1) there is no `myspace1`
2) `myspace2` is created and `myuser` is granted `read,write` for it
3) `myspace2` is renamed to `myspace1`

This way, the applier does not grant/revoke a thing and the configuration is still in sync.

Totktonada: Looks valid. Thanks! I'll think around.

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,95 @@
...
+    local status = 'ready'
+
+    for _, alert in pairs(self._alerts) do
+        assert(alert.type == 'error' or alert.type == 'warn')
+        if alert.type == 'error' then
+            return 'check_errors'
+        end
+
+        status = 'check_warnings'
+    end
```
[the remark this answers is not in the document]
Totktonada: It changes a meaning of the statuses. An instance reports `startup_in_progress` or `reload_in_progress` while it is in a transient state. The `ready`, `check_warnings` or `check_errors` status is set only after applying a configuration.

`aboard_status()` is called at end of the configuration applying. `config:info()` returns the saved value.

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,95 @@
...
+        self._on_drop(self, key)
+    end
+end
+
+local function aboard_clean(self)
+    self._alerts = {}
+end
+
+local function aboard_alerts(self)
+    -- Don't return alert keys.
```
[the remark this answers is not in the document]
Totktonada: Keys are not messages. These are the separate values.

ochaplashkin (approved):

src/box/lua/config/applier/credentials.lua
Totktonada: It is a hash table lookup now (using the `key`). In the described scenario we need to perform a traverse and match the keys against a condition. Or allow to mark several alerts with one key/label and search using this label. I'll expand the comment.

src/box/lua/config/applier/credentials.lua
Totktonada: I would keep it with this redundancy to bring a proper context to an occasional reader.

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,95 @@
...
+    self._alerts[key] = nil
+    if self._on_drop ~= nil then
+        self._on_drop(self, key)
+    end
+end
+
+local function aboard_clean(self)
+    self._alerts = {}
+end
+
```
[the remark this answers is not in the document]
Totktonada: Interesting point.

There are several thoughts around.

* The `_on_drop` callback receives a key of the alert that is dropped, so it is supposed to be called for each dropped alert. Calling it for all the alerts on `:clean()` seems a bit expensive.
* It adds extra transitions between states. For example, instead of `check_errors` -> `reload_in_progress` we would do `check_errors` -> `ready` -> `reload_in_progress`.
* The alert board is not the single source of truth for the configuration status, it is one of the sources. The warnings/errors presence is used as a supplementary information to decide about the next status on some transitions, but it has no effect on a transition to, say, `reload_in_progress`.

I would add a separate `on_clean` callback if we actually need to react on this action. However, we call `:clean()` right from the code that manages the status, so there is no need now.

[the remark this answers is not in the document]
Lord-KA: I am just worried, that there are a lot of status changes happening in different places completely independently:
1) in `config/init.lua`
2) in `_aboard:status()`
3) in `_on_drop()`

Right now, it is quite difficult to understand status state machine from the code as is, and adding another inconstancy (`:clean()` doesn't call the call-back) isn't helping.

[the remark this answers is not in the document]
Lord-KA: Actually, is there a different `_on_drop()` function expected? Maybe we can just inline it and simplify everything.

src/box/lua/config/applier/credentials.lua
Lord-KA: Ok.

src/box/lua/config/applier/credentials.lua
Lord-KA: Oh, now I get it.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Lord-KA: Let's discuss it all together in the [thread above](https://github.com/tarantool/tarantool/pull/9598#discussion_r1457093321).

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Totktonada: > Right now, it is quite difficult to understand status state machine from the code as is

I agree. OTOH, I don't think that changing the status from `aboard:clean()` would change it.

There are two ways how the status is changed:

* Directly from the `config/init.lua` code.
* As a reaction on an external event using the `on_drop()` callback (which is written in `config/init.lua` code too).

At least, all the changes goes over the same code file.

> Actually, is there a different `_on_drop()` function expected? Maybe we can just inline it and simplify everything.

Inline to where?

If we inline it to aboard.lua, it depends on the config -- it is a circular dependency. And it will manage config's status.

If we inline it to `credentials.lua`, it will manage config's status instead of config itself.

I think that the `on_drop` callback variant looks the best.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Totktonada: OTOH, if it you mean aboard's status (not config's status) or some `has_warning`/`has_error` flags, it has sense.

I'll experiment around and update it this way if the code will not look too complicated for me.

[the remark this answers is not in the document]
Lord-KA: Well, yes, this crossed my mind, maybe having two state machines would help. But that needs to be looked into.

Totktonada:
I've marked this PR as draft to work on it a bit more.

I'll answer to the discussions, but some changes are only in my local copy at the moment. I'll sync everything a bit later and will notify the participants by a message here.

changelogs/unreleased/config-stuck-privilege-granting-alert.md
Totktonada: Thanks! Applied.

changelogs/unreleased/config-missing-alerts.md
Totktonada: Thanks! Applied.

changelogs/unreleased/config-stuck-privilege-granting-alert.md
Totktonada: Thanks! Fixed.

src/box/lua/config/utils/aboard.lua
Totktonada: I agree, the methods are doing very similar things and could be merged into one method.

I've placed it into the `opts` table, so it'll be more obvious what is the string at the caller side.

```lua
aboard:set({type = 'warn', message = '<...>'}, {key = '<...>'})
```

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,95 @@
...
+local function aboard_set(self, key, alert)
+    assert(alert.type == 'error' or alert.type == 'warn')
+    if alert.type == 'error' then
+        log.error(alert.message)
+    else
+        log.warn(alert.message)
+    end
+    alert.timestamp = datetime.now()
+    self._alerts[key] = alert
```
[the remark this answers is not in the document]
Totktonada: I've added description for the methods and it seems that it should be enough to follow the logic of the module.

```diff
diff --git a/src/box/lua/config/utils/aboard.lua b/src/box/lua/config/utils/aboard.lua
index 5f6d88b0c4..a910468506 100644
--- a/src/box/lua/config/utils/aboard.lua
+++ b/src/box/lua/config/utils/aboard.lua
@@ -3,6 +3,28 @@
 local datetime = require('datetime')
 local log = require('internal.config.utils.log')
 
+-- Set an alert.
+--
+-- The `key` is an unique indentifier of the alert. Attempt to set
+-- a new alert with the same key replaces the existing alert.
+--
+-- The `key` can be used to drop the alert using :drop().
+--
+-- The `alert` is a table of the following structure.
+--
+-- {
+--     type = 'warn' or 'error',
+--     message = <string>,
+-- }
+--
+-- More fields may be placed into the table and all these fields
+-- are returned from `:alerts()`.
+--
+-- The alert board takes an ownership of the `alert` table and
+-- modifies it.
+--
+-- A timestamp is saved on the `:set()` method call to show it in
+-- the `:alerts()` result.
 local function aboard_set(self, key, alert)
     assert(alert.type == 'error' or alert.type == 'warn')
     if alert.type == 'error' then
@@ -14,21 +36,42 @@ local function aboard_set(self, key, alert)
     self._alerts[key] = alert
 end
 
+-- Get an alert or return nil.
 local function aboard_get(self, key)
     return self._alerts[key]
 end
 
+-- Drop an alert pointed by the given `key`.
+--
+-- The function is no-op if the alert doesn't exist.
+--
+-- The `on_drop` callback (see the aboard.new() function) is
+-- called if the alert had exist.
 local function aboard_drop(self, key)
+    if self._alerts[key] == nil then
+        return
+    end
     self._alerts[key] = nil
     if self._on_drop ~= nil then
         self._on_drop(self, key)
     end
 end
 
+-- Drop all the alerts.
+--
+-- The `on_drop` callback is NOT called.
 local function aboard_clean(self)
     self._alerts = {}
 end
 
+-- Serialize the alerts to show them to a user.
+--
+-- The alerts are sorted by a time of the :set() method call and
+-- returned as an array-like table.
+--
+-- The keys passed to the :set() method are NOT shown in the
+-- result of this method: they're considered as internal
+-- identifiers.
 local function aboard_alerts(self)
     -- Don't return alert keys.
     local alerts = {}
@@ -62,6 +105,7 @@ local function aboard_status(self)
     return status
 end
 
+-- Return `true` if there are no alerts, `false` otherwise.
 local function aboard_is_empty(self)
     return next(self._alerts) == nil
 end
@@ -78,6 +122,16 @@ local mt = {
     },
 }
 
+-- Create a new alert board.
+--
+-- The `on_drop` callback is called on the :drop() method call.
+-- It is not called on the :clean() call.
+--
+-- The callback should have the following prototype.
+--
+-- | local function on_drop_cb(aboard, key)
+-- |     <...>
+-- | end
 local function new(opts)
     local on_drop
     if opts ~= nil then
```

As you can see from the diff I also skipped the `on_drop` callback invocation if the alert doesn't exist before the call. It doesn't affect any user visible behavior, but eliminates unnecessary actions in the code.

The diff is applied to the first patch of the series, where the `aboard` module is introduced.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Totktonada: I've experimented around, see below.

```diff
diff --git a/src/box/lua/config/utils/aboard.lua b/src/box/lua/config/utils/aboard.lua
index 480f4c98d2..9b572a93cb 100644
--- a/src/box/lua/config/utils/aboard.lua
+++ b/src/box/lua/config/utils/aboard.lua
@@ -28,6 +28,7 @@ local log = require('internal.config.utils.log')
 local function aboard_set(self, key, alert)
     assert(alert.type == 'error' or alert.type == 'warn')
     if alert.type == 'error' then
+        self._error_count = self._error_count + 1
         log.error(alert.message)
     else
         log.warn(alert.message)
@@ -48,9 +49,13 @@ end
 -- The `on_drop` callback (see the aboard.new() function) is
 -- called if the alert had exist.
 local function aboard_drop(self, key)
-    if self._alerts[key] == nil then
+    local alert = self._alerts[key]
+    if alert == nil then
         return
     end
+    if alert.type == 'error' then
+        self._error_count = self._error_count - 1
+    end
     self._alerts[key] = nil
     if self._on_drop ~= nil then
         self._on_drop(self, key)
@@ -61,6 +66,7 @@ end
 --
 -- The `on_drop` callback is NOT called.
 local function aboard_clean(self)
+    self._error_count = 0
     self._alerts = {}
 end
 
@@ -91,18 +97,13 @@ end
 -- * check_warnings
 -- * check_errors
 local function aboard_status(self)
-    local status = 'ready'
-
-    for _, alert in pairs(self._alerts) do
-        assert(alert.type == 'error' or alert.type == 'warn')
-        if alert.type == 'error' then
-            return 'check_errors'
-        end
-
-        status = 'check_warnings'
+    if self._error_count > 0 then
+        return 'check_errors'
+    elseif self:is_empty() then
+        return 'ready'
+    else
+        return 'check_warnings'
     end
-
-    return status
 end
 
 -- Return `true` if there are no alerts, `false` otherwise.
@@ -139,6 +140,7 @@ local function new(opts)
     end
 
     return setmetatable({
+        _error_count = 0,
         _alerts = {},
         _on_drop = on_drop,
     }, mt)
```

TBH, I don't like to tweak the counter in several methods. It looks as an unnecessary complication.

I expect that amount of alerts is small enough to just traverse them on an event.

I'll leave it as is now. Let's discuss.

src/box/lua/config/applier/credentials.lua
Totktonada: I've expanded it a bit.

```diff
diff --git a/src/box/lua/config/applier/credentials.lua b/src/box/lua/config/applier/credentials.lua
index 62f92e2781..332444bbdc 100644
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -601,9 +601,10 @@ local privileges_action_f = function(grant_or_revoke, role_or_user, name, privs,
     -- object: one for the 'read' permission, one for 'write'
     -- permission and so on.
     --
-    -- More complicated solution is to perform a search for
+    -- More complicated solution is to perform a traverse for
     -- relevant alerts on successful privilege granting to modify
-    -- or drop a right alert.
+    -- or drop a right alert. In this case we don't know an exact
+    -- list of privileges to form a key to drop the alert.
     local key = ('box.schema.%s.%s(%q, %q, %q, %q)'):format(
         role_or_user, grant_or_revoke, name, privs, obj_type, obj_name)
```

However, this comment will likely gone during the work on the problem with an object renaming.

src/box/lua/config/applier/credentials.lua
Totktonada: I've implemented another solution that covers this scenario. See details in the description of the last commit.

In brief, a simple way with discarding an alert after granting of a privilege doesn't work here. We need to traverse alerts on an upper level and compare them to an actual state of privileges in the database.

I've tried to implement the solution in a more or less simple way. It means several scans over the alerts. I think that it shouldn't be a problem: the cost is anyway small comparing to a write of the rename operation to WAL. And it shouldn't be a frequent operation.

src/box/lua/config/applier/credentials.lua
Totktonada: This comment is removed in the new implementation (see [below](https://github.com/tarantool/tarantool/pull/9598#discussion_r1457157778)).

Totktonada:
I've addressed all the comments (or answered to them). I've changed a solution for #9574 (see details [here](https://github.com/tarantool/tarantool/pull/9598#discussion_r1457157778)), so I'm going to re-request reviews.

src/box/lua/config/applier/credentials.lua
Lord-KA: The solution looks valid, but quite complicated. It seems like you are trying to store and update independent diff between current configuration and desired one (after a sync, we don't consider in-progress situation). As the alerts don't have to be exactly up to date at each moment, maybe it is fine to calculate this diff once in a while? We can mark all grant warning in a separate subtype, drop them all together and add new base on the diff (using  `privileges_subtract()`).

Lord-KA: The solution looks valid, but quite complicated. It seems like you are trying to store and update independent diff between current configuration and desired one (after a sync, we don't consider in-progress situation). As the alerts don't have to be exactly up to date at each moment, maybe it is fine to calculate this diff once in a while? We can mark all grant warning in a separate subtype, drop them all together and add new base on the diff (using  `privileges_subtract()`).

Lord-KA: This can also simplify the status state machine implementation. The state would change only on this diff calculation.

ochaplashkin (approved):

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,178 @@
...
+
+-- Return `true` if there are no alerts, `false` otherwise.
+local function aboard_is_empty(self)
+    return next(self._alerts) == nil
+end
+
+-- Traverse over alerts with their internal identifiers (keys).
+local function aboard_pairs(self)
+    return pairs(self._alerts)
+end
```
Lord-KA, line 138: `aboard_pairs()` is introduced, but never used in the patchset.

src/box/lua/config/applier/credentials.lua
Lord-KA: I looked into it and came up with a draft diff:
```diff
diff --git a/src/box/lua/config/applier/credentials.lua b/src/box/lua/config/applier/credentials.lua
index c2fb2cba8..553049a88 100644
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -654,6 +654,8 @@ local function sync_privileges(credentials, obj_to_sync)
                     obj_to_sync.type, obj_to_sync.name)
     end
 
+    aboard_drop_alerts({trait = 'postponed_grant'})
+
     -- The privileges synchronization between A and B is performed in 3 steps:
     -- 1. Grant all privileges that are present in B,
     --    but not present in A (`grant(B - A)`).
@@ -703,7 +705,22 @@ local function sync_privileges(credentials, obj_to_sync)
             end
         end
 
-        sync_alerts(role_or_user, name)
+        -- Update box_privileges to check which grants have been postponed and
+        -- issue a warning for each one of them.
+        box_privileges = box.schema[role_or_user].info(name)
+        box_privileges = privileges_from_box(box_privileges)
+        local postponed_grants = privileges_subtract(config_privileges, box_privileges)
+
+        for _, grant in ipairs(postponed_grants) do
+            local alert = {}
+            local msg = "credentials.apply: %s %q hasn't been created yet, " ..
+                        "'box.schema.%s.%s(%q, %q, %q, %q)' will be applied later"
+            alert.type = 'warn'
+            alert._trait = 'postponed_grant'
+            alert.message = msg:format(grant.obj_type, grant.obj_name, role_or_user,
+                                       'grant', name, grant.privs, grant.obj_type,
+                                       grant.obj_name)
+            config._aboard:set(alert)
+        end
     end
 
     -- Note that the logic inside the sync() function, i.e. privileges dump
```

It looks quite simple in terms of implementation, all we really need to do it to drop all "postponed_grant" issues on a sync and issue new warnings after sync iteration (per user).

Lord-KA: On the other hand, it scares me, how much work the program has to do on each small sync with possibly just a few grants/revokes.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Lord-KA: We've discussed it more privately and agreed that `status` is required only after the loading, so it is ok to scan the alerts afterwards.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Totktonada: > The only thing that bothers me now is that I still can't be sure from this comment/code below, what fields are always present in the `self._alerts`.

AFAIU, if I add a description for the `_alert` field, it solves the problem? It is done.

```diff
diff --git a/src/box/lua/config/utils/aboard.lua b/src/box/lua/config/utils/aboard.lua
index 58bdc8a0ee..6cfbb000e9 100644
--- a/src/box/lua/config/utils/aboard.lua
+++ b/src/box/lua/config/utils/aboard.lua
@@ -167,6 +167,22 @@ local function new(opts)
     end
 
     return setmetatable({
+        -- _alerts is a table of the following structure.
+        --
+        -- {
+        --     [key1] = {
+        --         type = 'warn' or 'error',
+        --         message = <string>,
+        --         timestamp = <datetime object>,
+        --         <..arbitrary fields..>
+        --     },
+        --     [key2] = {
+        --         <...>
+        --     },
+        --     <...>
+        -- }
+        --
+        -- The key can be a string or a number.
         _alerts = {},
         _next_key = 1,
         _on_drop = on_drop,

```

> <...> maybe verify whole argument structure (add `assert(type(alert.message) == "string"`) and <...>

I'm not sure that I should forbid table/cdata/userdata with `__tostring` and/or `__serialize` as a message, so I would leave it unchecked for now.

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,149 @@
...
+        log.warn(alert.message)
+    end
+    alert.timestamp = datetime.now()
+    self._alerts[key] = alert
+end
+
+-- Return an alert pointed by the given `key` or nil.
+local function aboard_get(self, key)
+    return self._alerts[key]
+end
```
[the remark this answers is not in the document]
Totktonada: I would keep it for stylistic purposes: it is easier to switch from one description to another, when all of them are present. Otherwise, I need to switch my brain into the Lua interpreter mode :)

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,149 @@
...
+
+-- Drop an alert pointed by the given `key`.
+--
+-- The function is no-op if the alert doesn't exist.
+--
+-- The `on_drop` callback (see the aboard.new() function) is
+-- called if the alert had exist.
```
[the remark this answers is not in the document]
Totktonada: Thanks! Fixed.

src/box/lua/config/utils/aboard.lua
```diff
@@ -0,0 +1,149 @@
...
+end
+
+-- Serialize the alerts to show them to a user.
+--
+-- The alerts are sorted by a time of the :set() method call and
+-- returned as an array-like table.
+--
+-- The keys passed to the :set() method are NOT shown in the
+-- result of this method: they're considered as internal
+-- identifiers.
```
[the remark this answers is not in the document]
Totktonada: No problems, dropped.

changelogs/unreleased/config-stuck-privilege-granting-alert.md
```diff
@@ -0,0 +1,4 @@
...
+
```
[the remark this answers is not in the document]
Totktonada: Done.

> 1. The `aboard` module now ignores underscored fields of an alert on its serialization to allow a caller to store a machine-readable information in them.

src/box/lua/config/utils/aboard.lua
Totktonada: It is called as `config._aboard:pairs()` in [sync_alerts](https://github.com/tarantool/tarantool/commit/6ae9a0a55419ef2f95705a24a0064d201cb836a1#diff-b71368783dfc0d7f6c89c69d145494c61e9e5432f97e6895beadf1767da139a7R612).

src/box/lua/config/applier/credentials.lua
Totktonada: A brief summary of our offline discussion with Gleb: I'm going to implement this proposal. I also plan to post here a summary of cons and pros I see.

src/box/lua/config/utils/aboard.lua
Lord-KA: Oh, missed that, thank you.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Lord-KA: > AFAIU, if I add a description for the _alert field, it solves the problem? It is done.

Thanks, that's great.

> I'm not sure that I should forbid table/cdata/userdata with __tostring and/or __serialize as a message, so I would leave it unchecked for now.

I didn't thing about other types of messages. While I still don't really expect anything else but string here, I agree with your explanation. The only "but" is the comment message, it states message type only as `<string>`. I suggest replacing it with something like `<string> or <serializable>`.

src/box/lua/config/applier/credentials.lua
Totktonada: I've pushed the new implementation.

src/box/lua/config/utils/aboard.lua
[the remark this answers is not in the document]
Lord-KA: > I don't want to mark non-string values as a proper input for the message field. I would better add the assert. I'll do if you insist.

Personally I don't really see a reason for such in-place serialization, and I am not a big fan of a "gray" API. On the other hand, this isn't a blocker for me, so do as you wish.

changelogs/unreleased/config-stuck-privilege-granting-alert.md
```diff
@@ -0,0 +1,4 @@
...
+* An alert regarding delayed privilege granting is now cleared when the
```
Lord-KA, line 3: Nit: 
> privilege will be granted, when the object will be created.

privilege will be granted, when the object *is* created [1]

[1]: https://jakubmarian.com/will-after-when-in-english/

Lord-KA: Nit:

> dropping alerts *at* this point has no effect.

Lord-KA: Nit:

> The third one is implemented here with idea of code simplicity.

src/box/lua/config/utils/aboard.lua
```diff
@@ -67,6 +68,30 @@ local function aboard_drop(self, key)
...
+--
+-- | local function filter_f(key, alert)
+-- |     return <boolean>
+-- | end
+--
+-- The filter function is called on each alert. An alert is
+-- dropped if the function returns `true`.
+--
+-- The `on_drop` callback (see the aboard.new() function) is
+-- called for each of the dropped alerts.
```
Lord-KA, line 81: Nit:
> called for each of the dropped *alert*.

Lord-KA (approved):
Thanks for the patchset! Everything looks great, besides a few grammar-nazi nits above. LGTM

src/box/lua/config/applier/credentials.lua
Lord-KA: Thank you!

src/box/lua/config/applier/credentials.lua
```diff
@@ -613,6 +619,50 @@ local function sync_privileges(credentials, obj_to_sync)
...
+        -- reload from a remote config storage in Tarantool
+        -- Enterprise Edition).
+        --
+        -- Important: this code deliberately ignores the
+        -- `obj_to_sync` filter. It should be in-sync with the
+        -- :drop_if() call above.
+        for _, grant in ipairs(missed_grants) do
+            local alert = {
+                type = 'warn',
+                _trait = 'missed_privilege',
```
Serpentian, line 656: Do we really need this `_trait`? Can't we just use lua regex to the alert message? The better option would be to use alert keys, if they start with some value, we can delete that in `drop_if`.

It just seems to me, that all droppable alerts should have keys.

src/box/lua/config/utils/aboard.lua
Totktonada: Fixed.

changelogs/unreleased/config-stuck-privilege-granting-alert.md
Totktonada: All done.

src/box/lua/config/applier/credentials.lua
Totktonada: > Can't we just use lua regex to the alert message?

Message are for humans. It is acceptable to use them programmatically in a test, but not in an application logic. Neither inside tarantool, nor in a user code.

So, we need some machine-readable information. We need to drop alerts according to some properties and either encode them into keys somehow (and parse then) or add a field. The latter fits better for a machine processing (no need to encode and parse).

> It just seems to me, that all droppable alerts should have keys.

The keys serve the point access pattern, but they don't fit good to a bulk access. I wouldn't try to follow the 'droppable == have caller defined keys' rule.

Well, ideally we have an index based on the needed properties and find necessary alerts faster. However, I don't expect such amount of alerts that costs this complexity.

ImeevMA (approved):

andreyaksenov (approved):

Totktonada:
Pushed to `master`, it is future `3.1.0`.

Backported to `release/3.0` (future 3.0.2) in PR #9645.

Issue #3225, opened 2018-03-07
Bump compaction priority of an index/range if read amplification is high


kostja:
Test scenario (Mail.Ru visapi):
- create an index which is 35GB in size
- delete all data in an index with expirationd
- expirationd keeps using a lot of cpu while iterating through garbae and major compaction never happens, while the space is empty.

This happens because we never feed back read amplification cost into the compaction priority of a range. This is an omission, please fix.

locker:
We actually ran into this problem while investigating #3603: running the random insertion test attached to the above-mentioned issue on a space with two indexes resulted in about 15M records in the primary index and more than 250M records in the secondary index after several hours of run time while the number of unique keys was 10M, i.e. space amplification for the secondary index exceeded 25, which is huge even for an LSM-tree based design. Probably, we need to take into account the number of DELETE statements and force compaction if there are too many of them.

locker:
So, if we didn't have DELETE statements, neither explicit nor implicit (i.e. generated for secondary indexes on tuple update), we wouldn't need to do anything at all, because in this case both read and write and space amplification are limited by the target LSM tree shape. So this issue is about taking into account DELETE statements while calculating compaction priority.

If we didn't have implicit DELETEs, i.e. if there was no secondary indexes, write and space amplification would still be limited by the LSM tree shape, because the number of unique statements in a primary index can be estimated as the number of tuples at the last (largest) LSM tree level. Read amplification could be greater though, because there might be a group of deleted statements requested frequently and bloom filters don't reflect such requests. To handle this, we can't simply trigger compaction when the portion of DELETE statements is high - it may be not enough, because the number of deleted statements may be low while the number of requests for them may be high. We should keep track of such garbage requests and trigger compaction when there are too many of them (probably need to maintain a histogram for that).

[...]

Issue #5430, opened 2020-10-16
test: flaky replication/election_qsync.test.lua test


avtikhon:
Tarantool version:
Tarantool 2.6.0-208-ga20a04cba
Target: FreeBSD-amd64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=OFF
Compiler: /usr/bin/cc /usr/bin/c++
C_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression -Werror
CXX_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Werror

OS version:
FreeBSD 12

Bug description:
https://gitlab.com/tarantool/tarantool/-/jobs/795916093#L5385

[artifacts.zip](https://github.com/tarantool/tarantool/files/5393978/artifacts.zip)

# results file checksum: afaa5d0f392c8de5420a05b268d04741
```
[013] --- replication/election_qsync.result	Fri Oct 16 19:25:53 2020
[013] +++ replication/election_qsync.reject	Fri May  8 08:21:42 2020
[013] @@ -127,14 +127,15 @@
[013]   | ...
[013]  _ = box.space.test:replace{2}
[013]   | ---
[013] + | - error: Found uncommitted sync transactions from other instance with id 2
[013]   | ...
[013]  box.space.test:select{}
[013]   | ---
[013]   | - - [1]
[013] - |   - [2]
[013]   | ...
[013]  box.space.test:drop()
[013]   | ---
[013] + | - error: Found uncommitted sync transactions from other instance with id 2
[013]   | ...
[013]  
[013]  test_run:cmd('delete server replica')
[013] 
```

Steps to reproduce:

Optional (but very desirable):
* coredump
* backtrace
* netstat

avtikhon:
Closed as duplicated https://github.com/tarantool/tarantool/issues/5395

Gerold103:
```
[001] --- replication/election_qsync.result     Fri May  8 08:56:08 2020
[...]

Issue #3637, opened 2018-08-22
Replication tests fail sporadically with "User 'cluster' is not found"


locker:
Reproduced by [Travis CI](https://travis-ci.org/tarantool/tarantool/jobs/419163281) on 1.10:
```
[011] replication/ddl.test.lua                        memtx           [ pass ]
[011] replication/ddl.test.lua                        vinyl           
[011] 
[011] [Instance "ddl4" returns with non-zero exit code: 1]
[011] 
[011] Last 15 lines of Tarantool Log file [Instance "ddl4"][/tarantool/test/var/011_replication/ddl4.log]:
[011] 2018-08-22 13:32:24.465 [10448] main/101/ddl4 C> log level 5
[011] 2018-08-22 13:32:24.465 [10448] main/101/ddl4 I> mapping 268435456 bytes for memtx tuple arena...
[011] 2018-08-22 13:32:24.465 [10448] main/101/ddl4 I> mapping 134217728 bytes for vinyl tuple arena...
[011] 2018-08-22 13:32:24.468 [10448] iproto/101/main I> binary: bound to unix/:/tarantool/test/var/011_replication/autobootstrap4.sock
[011] 2018-08-22 13:32:24.468 [10448] main/108/applier/cluster@unix/:/tarantoo I> remote master is 1.10.1 at unix/:/tarantool/test/var/011_replication/autobootstrap2.sock
[011] 2018-08-22 13:32:24.470 [10448] main/109/applier/cluster@unix/:/tarantoo I> remote master is 1.10.1 at unix/:/tarantool/test/var/011_replication/autobootstrap3.sock
[011] 2018-08-22 13:32:24.470 [10448] main/107/applier/cluster@unix/:/tarantoo I> remote master is 1.10.1 at unix/:/tarantool/test/var/011_replication/autobootstrap1.sock
[011] 2018-08-22 13:32:24.470 [10448] main/110/applier/cluster@unix/:/tarantoo I> remote master is 1.10.1 at unix/:/tarantool/test/var/011_replication/autobootstrap4.sock
[011] 2018-08-22 13:32:24.470 [10448] main/107/applier/cluster@unix/:/tarantoo I> failed to authenticate
[011] 2018-08-22 13:32:24.470 [10448] main/107/applier/cluster@unix/:/tarantoo xrow.c:891 E> ER_LOADING: Instance bootstrap hasn't finished yet
[...]

PR #9604, opened 2024-01-19
box: finish client fibers on shutdown

nshy:
In the process of graceful shutdown it is convenient to first finish all client (non system) fibers. Otherwise we should be ready for any subsystem to handle request from client fiber during or after subsystem shutdown. This would make code more complex.

We first cancel client fibers and then wait for their finishing. The fiber may not respond to cancel and hang which cause shutdown hang but this is the approach we choose for iproto shutdown already.

Note that as a result of this approach application will panic if it is shutdown during execution of initialization script (in particular if this script is doing box.cfg).

There are changes in application/test to adopt to client fibers shutdown:

- make code cancellable (only to pass existing tests, we did not investigate all the possible places that should be made such).

- make console stop sending echo to client before client fibers shutdown. Otherwise as console server fiber is client one we will send message that fiber is cancelled on shutdown which breaks a lot of existing tests. This approach is on par with iproto shutdown.

- some tests (7743, replication-luatest/shutdown, replication/anon, replication/force_recovery etc etc) test shutdown during execution of init script. Now panic is expected so change them accordingly.

- some tests (8530, 5430, errinj_vylog) use injection that block client fiber finishing. In that tests we don't need graceful shutdown so let's just kill tarantool instead.

- we change test in vinyl/errinj for `gh-3225`. We don't really need to check when vinyl reader is blocked as it executes small tasks (we assume reading syscall will not hang). Also change test for vinyl dump shutdown by slowing dump down instead of blocking it entirely. This is required to finish in time client fibers in the test.

- other similar changes

Also we can drop code from replication shutdown which is required to handle client requests during/after shutdown.

Part of ##8423


src/box/box.cc
```diff
@@ -5926,6 +5930,7 @@ box_storage_shutdown()
 	if (!is_storage_initialized)
 		return;
 	iproto_shutdown();
+	fiber_shutdown();
```
locker, line 5933: Shouldn't we call it even if `is_storage_initialized` is unset? That is, shouldn't this be moved to `box_shutdown`?

src/box/iproto.cc
```diff
@@ -2130,6 +2130,7 @@ tx_accept_msg(struct cmsg *m)
 	rlist_add_entry(&msg->connection->tx.inprogress, msg,
 			in_inprogress);
 	msg->fiber = fiber();
+	fiber_set_system(fiber(), false);
```
locker, line 2133: We shouldn't reset the flag here for all fibers just in case. It should be set before we start system work and unset right after system work is done. You just want to treat idle pool fibers as non-cancellable, better set/clear the flag in the fiber pool with a proper comment.

src/box/lua/console.lua
```diff
@@ -985,6 +986,9 @@ local function client_handler(client, _peer)
     state:print(string.format("%-63s\n%-63s\n",
         "Tarantool ".. version.." (Lua console)",
         "type 'help' for interactive help"))
+    box.ctl.on_shutdown(function()
+        state.running = false
```
locker, line 990: Don't you need to wake up the console fiber here?

src/lib/core/fiber.c
```diff
@@ -2339,3 +2350,18 @@ fiber_lua_state(struct fiber *f)
...
+	assert(cord()->shutdown_fiber == NULL);
+	struct fiber *fiber;
+	rlist_foreach_entry(fiber, &cord()->alive, link) {
+		if (!(fiber->flags & FIBER_IS_SYSTEM))
+			fiber_cancel(fiber);
+	}
+	cord()->shutdown_fiber = fiber();
+	while (cord()->client_fiber_count != 0)
+		fiber_yield();
+	cord()->shutdown_fiber = NULL;
```
locker, line 2366: It may be more appropriate to use a condition variable here. Anyway, what happens, if a client fiber starts another client fiber after it's cancelled?

test/box-luatest/gh_7743_term_initial_cfg_snap_test.lua
```diff
@@ -20,7 +20,6 @@ g.test_sigterm_during_initial_snapshot = function()
             -- uses usleep() which is a pthread cancellation point.
             TARANTOOL_RUN_BEFORE_BOX_CFG = [[
                 box.ctl.set_on_shutdown_timeout(1000)
-                box.error.injection.set('ERRINJ_MAIN_MAKE_FILE_ON_RETURN', true)
```
locker, line 23: Why? Does it stop working?

test/replication/gh-3637-misc-error-on-replica-auth-fail.result
```diff
@@ -49,6 +49,11 @@ vclock[0] = nil
 _ = test_run:wait_vclock('replica_auth', vclock)
 ---
 ...
+-- If we stop prematurely replica will not finish its start and
+-- will panic.
+_ = test_run:wait_log('replica_auth', 'entering the event loop')
```
locker, line 54: Wouldn't it be better to wait until `box.ctl.is_recovery_finished()` returns true?

test/replication/gh-5430-cluster-mvcc.result
```diff
@@ -136,7 +136,7 @@ test_run:cmd('delete server replica1')
  | ---
  | - true
  | ...
-test_run:cmd('stop server master')
+test_run:cmd('stop server master with signal=KILL')
```
locker, line 139: Why?

test/vinyl/errinj_vylog.test.lua
```diff
@@ -198,7 +198,7 @@ _ = fiber.create(function() s:create_index('sk', {parts = {2, 'unsigned'}}) end)
 fiber.sleep(0.01)
 
 -- Should ignore the incomplete index on recovery.
-test_run:cmd('restart server default')
+test_run:cmd('restart server default with signal=KILL')
```
locker, line 201: Why?

src/box/box.cc
nshy: Added a comment as discussed f2f.

```diff
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -5930,6 +5930,10 @@ box_storage_shutdown()
        if (!is_storage_initialized)
                return;
        iproto_shutdown();
+       /*
+        * Finish client fibers after iproto_shutdown otherwise new fibers
+        * can be started through new iproto requests.
+        */
        fiber_shutdown();
        replication_shutdown();
 }
```

src/box/iproto.cc
nshy: Done. Now instead cbus message callback should set system flag for the fiber. Fibers inside pool are considered system.

<details>
<summary>diff</summary>

```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index a640661e9..a51810116 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -2130,7 +2130,6 @@ tx_accept_msg(struct cmsg *m)
 	rlist_add_entry(&msg->connection->tx.inprogress, msg,
 			in_inprogress);
 	msg->fiber = fiber();
-	fiber_set_system(fiber(), false);
 	rmean_collect(msg->connection->iproto_thread->tx.rmean,
 		      REQUESTS_IN_PROGRESS, 1);
 	flightrec_write_request(msg->reqstart, msg->len);
@@ -2184,7 +2183,6 @@ tx_end_msg(struct iproto_msg *msg, struct obuf_svp *svp)
 	msg->connection->iproto_thread->tx.requests_in_progress--;
 	rlist_del(&msg->in_inprogress);
 	msg->fiber = NULL;
-	fiber_set_system(fiber(), true);
 	struct obuf *out = msg->connection->tx.p_obuf;
 	if (msg->connection->tx.p_obuf->used != svp->used)
 		/* Log response to the flight recorder. */
diff --git a/src/box/vy_scheduler.c b/src/box/vy_scheduler.c
index 208b6695d..680ab935c 100644
--- a/src/box/vy_scheduler.c
+++ b/src/box/vy_scheduler.c
@@ -890,6 +890,7 @@ vy_deferred_delete_batch_process_f(struct cmsg *cmsg)
 	struct vy_deferred_delete_batch *batch = container_of(cmsg,
 				struct vy_deferred_delete_batch, cmsg);
 	struct vy_task *task = batch->task;
+	fiber_set_system(fiber(), true);
 	/*
 	 * Wait for memory quota if necessary before starting to
 	 * process the batch (we can't yield between statements).
diff --git a/src/lib/core/fiber_pool.c b/src/lib/core/fiber_pool.c
index e15532516..e24f4f461 100644
--- a/src/lib/core/fiber_pool.c
+++ b/src/lib/core/fiber_pool.c
@@ -61,7 +61,9 @@ fiber_pool_f(va_list ap)
 			f->caller->flags |= FIBER_IS_READY;
 			assert(f->caller->caller == &cord->sched);
 		}
+		fiber_set_system(fiber(), false);
 		cmsg_deliver(msg);
+		fiber_set_system(fiber(), true);
 		fiber_check_gc();
 		/*
 		 * Normally fibers die after their function
@@ -131,7 +133,7 @@ fiber_pool_cb(ev_loop *loop, struct ev_watcher *watcher, int events)
 			f = rlist_shift_entry(&pool->idle, struct fiber, state);
 			fiber_call(f);
 		} else if (pool->size < pool->max_size) {
-			f = fiber_new_system(cord_name(cord()), fiber_pool_f);
+			f = fiber_new(cord_name(cord()), fiber_pool_f);
 			if (f == NULL) {
 				diag_log();
 				break;

```

</details>

src/lib/core/fiber.c
nshy: Discussed f2f and decided to keep it like this (rationale: we don't want to bring `fiber_cond.h` dependence here).

test/replication/gh-5430-cluster-mvcc.result
nshy: Hmm, this was added by mistake. In other places of this patch I used `KILL` because of injection blocking graceful shutdown. Here is different, injection is reset before stopping the master.

Yet this test prints the below message from time to time.
```
[001] * [QA Notice]
[001] *
[001] * The server 'gh-5430-mvcc-master' does not stop during 5 seconds after the 15 (SIGTERM) signal.
[001] * Info: process 488354 [R (running); tarantool gh-5430-mvcc-master.lua <running>]
[001] * Sending SIGKILL...
```

As test is finished successfully even in the presence of the above message let's remove this change and investigate the issue in the scope of ticket https://github.com/tarantool/tarantool/issues/9617

test/vinyl/errinj_vylog.test.lua
nshy: Added a comment here in the other places where we start using `KILL` signal.

<details>
<summary>diff</summary>

```diff
diff --git a/test/box-luatest/gh_8530_alter_space_snapshot_test.lua b/test/box-luatest/gh_8530_alter_space_snapshot_test.lua
index efe47814f..1e9eb0004 100644
--- a/test/box-luatest/gh_8530_alter_space_snapshot_test.lua
+++ b/test/box-luatest/gh_8530_alter_space_snapshot_test.lua
@@ -49,6 +49,8 @@ g.test_build_index = function(cg)
         box.snapshot()
         t.assert_equals(f:status(), 'suspended')
     end)
+    -- Use KILL because server will hang on shutdown due to injection.
+    -- We don't need graceful shutdown for the test anyway.
     cg.server.process:kill('KILL')
     cg.server:restart()
     cg.server:exec(function()
@@ -70,6 +72,8 @@ g.test_change_format = function(cg)
         box.snapshot()
         t.assert_equals(f:status(), 'suspended')
     end)
+    -- Use KILL because server will hang on shutdown due to injection.
+    -- We don't need graceful shutdown for the test anyway.
     cg.server.process:kill('KILL')
     cg.server:restart()
     cg.server:exec(function()
diff --git a/test/vinyl/errinj_vylog.result b/test/vinyl/errinj_vylog.result
index 90fc56d92..6ac76b2c8 100644
--- a/test/vinyl/errinj_vylog.result
+++ b/test/vinyl/errinj_vylog.result
@@ -399,6 +399,8 @@ fiber.sleep(0.01)
 ---
 ...
 -- Should ignore the incomplete index on recovery.
+-- Use KILL because server will hang on shutdown due to injection.
+-- We don't need graceful shutdown for the test anyway.
 test_run:cmd('restart server default with signal=KILL')
 s = box.space.test
 ---
diff --git a/test/vinyl/errinj_vylog.test.lua b/test/vinyl/errinj_vylog.test.lua
index 72513d923..54a69c659 100644
--- a/test/vinyl/errinj_vylog.test.lua
+++ b/test/vinyl/errinj_vylog.test.lua
@@ -198,6 +198,8 @@ _ = fiber.create(function() s:create_index('sk', {parts = {2, 'unsigned'}}) end)
 fiber.sleep(0.01)
 
 -- Should ignore the incomplete index on recovery.
+-- Use KILL because server will hang on shutdown due to injection.
+-- We don't need graceful shutdown for the test anyway.
 test_run:cmd('restart server default with signal=KILL')
 
 s = box.space.test

```

</details>

test/box-luatest/gh_7743_term_initial_cfg_snap_test.lua
nshy: Yeah, in this test we terminate server during execution of init script. Before this patch server exits with 0 exit code but after the patch server will panic due to `FiberIsCancelled` exception during init script execution.

Dropped `ERRINJ_MAIN_MAKE_FILE_ON_RETURN` as discussed:

<details>
<summary>diff</summary>

```diff
diff --git a/src/lib/core/errinj.h b/src/lib/core/errinj.h
index 89d81a606..12e5412fd 100644
--- a/src/lib/core/errinj.h
+++ b/src/lib/core/errinj.h
@@ -104,7 +104,6 @@ struct errinj {
 	_(ERRINJ_IPROTO_TX_DELAY, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_WRITE_ERROR_DELAY, ERRINJ_BOOL, {.bparam = false})\
 	_(ERRINJ_LOG_ROTATE, ERRINJ_BOOL, {.bparam = false}) \
-	_(ERRINJ_MAIN_MAKE_FILE_ON_RETURN, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_MEMTX_DELAY_GC, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_NETBOX_DISABLE_ID, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_NETBOX_FLIP_FEATURE, ERRINJ_INT, {.iparam = -1}) \
diff --git a/src/main.cc b/src/main.cc
index 42f3a68a2..2bdf67593 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -1109,14 +1109,5 @@ main(int argc, char **argv)
 	free((void *)instance.name);
 	free((void *)instance.config);
 	tarantool_free();
-	ERROR_INJECT(ERRINJ_MAIN_MAKE_FILE_ON_RETURN, do {
-		int fd = open("tt_exit_file.txt.inprogress",
-			      O_WRONLY | O_CREAT | O_TRUNC, -1);
-		if (fd < 0)
-			break;
-		dprintf(fd, "ExitCode: %d\n", exit_code);
-		close(fd);
-		rename("tt_exit_file.txt.inprogress", "tt_exit_file.txt");
-	} while (false));
 	return exit_code;
 }
diff --git a/test/box/errinj.result b/test/box/errinj.result
index 8c4c8150f..979031600 100644
--- a/test/box/errinj.result
+++ b/test/box/errinj.result
@@ -77,7 +77,6 @@ evals
   - ERRINJ_IPROTO_TX_DELAY: false
   - ERRINJ_IPROTO_WRITE_ERROR_DELAY: false
   - ERRINJ_LOG_ROTATE: false
-  - ERRINJ_MAIN_MAKE_FILE_ON_RETURN: false
   - ERRINJ_MEMTX_DELAY_GC: false
   - ERRINJ_NETBOX_DISABLE_ID: false
   - ERRINJ_NETBOX_FLIP_FEATURE: -1

```

</details>

src/box/lua/console.lua
nshy: Change this hunk to:
```diff
@@ -985,7 +986,14 @@ local function client_handler(client, _peer)
     state:print(string.format("%-63s\n%-63s\n",
         "Tarantool ".. version.." (Lua console)",
         "type 'help' for interactive help"))
+    local on_shutdown = function()
+        state.running = false
+        fiber.cancel(state.fiber)
+    end
+    state.fiber = fiber.self()
+    box.ctl.on_shutdown(on_shutdown)
     repl(state)
+    box.ctl.on_shutdown(nil, on_shutdown)
     session_internal.run_on_disconnect()
 end
```

test/replication/gh-5430-cluster-mvcc.result
nshy: Turned out the issue with stopping master server was due to 2 factors. First is strange #9617 execution itself (so the ticket stays) which causes starting a lot of consoles on master during test. Second I forgot to cleanup `on_shutdown` trigger on console exit.

nshy:
I also added this hunk to fix memleak with injections.

```diff
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -1028,7 +1028,7 @@ checkpoint_f(va_list ap)
                return -1;
        }

-       struct mh_i32_t *temp_space_ids = mh_i32_new();
+       struct mh_i32_t *temp_space_ids;

        say_info("saving snapshot `%s'", snap->filename);
        ERROR_INJECT_WHILE(ERRINJ_SNAP_WRITE_DELAY, {
@@ -1040,6 +1040,7 @@ checkpoint_f(va_list ap)
        });
        ERROR_INJECT(ERRINJ_SNAP_SKIP_ALL_ROWS, goto done);
        struct space_read_view *space_rv;
+       temp_space_ids = mh_i32_new();
        read_view_foreach_space(space_rv, &ckpt->rv) {
                FiberGCChecker gc_check;
                bool skip = false;
```

test/replication/gh-3637-misc-error-on-replica-auth-fail.result
nshy: It is not clear can `box.cfg` yield or not after `box.ctl.is_recovery_finished()` is true. So change to next:

```diff
--- a/test/replication/gh-3637-misc-error-on-replica-auth-fail.test.lua
+++ b/test/replication/gh-3637-misc-error-on-replica-auth-fail.test.lua
@@ -24,6 +24,11 @@ vclock = test_run:get_vclock('default')
 vclock[0] = nil
 _ = test_run:wait_vclock('replica_auth', vclock)

+-- Wait server init script finish or server will panic on stop.
+test_run:switch('replica_auth')
+test_run:wait_cond(function() return _G.startup_finished == true end)
+
+test_run:switch('default')
--- a/test/replication/replica_auth.lua
+++ b/test/replication/replica_auth.lua
@@ -4,9 +4,12 @@ local USER_PASS = arg[1]
 local TIMEOUT = arg[2] and tonumber(arg[2]) or 0.1

 require('console').listen(os.getenv('ADMIN'))
+_G.startup_finished = false

 box.cfg({
     listen = os.getenv("LISTEN"),
     replication = USER_PASS .. "@" .. os.getenv("MASTER"),
     replication_timeout = TIMEOUT,
 })
+
+_G.startup_finished = true
```

src/box/lua/console.lua
```diff
@@ -985,7 +986,14 @@ local function client_handler(client, _peer)
     state:print(string.format("%-63s\n%-63s\n",
         "Tarantool ".. version.." (Lua console)",
         "type 'help' for interactive help"))
+    local on_shutdown = function()
+        state.running = false
+        fiber.cancel(state.fiber)
```
locker, line 991: Cancelling the console fiber will raise an exception in it so setting `self.running` here to false and checking it in `function repl()` looks pointless. Also, it `repl` fails with an exception, then `box.ctl.on_shutdown` trigger won't be reset.

src/box/replication.cc
```diff
@@ -1342,6 +1313,7 @@ replicaset_sync(void)
 		say_info("replica set sync complete");
 		box_set_orphan(false);
 	}
+	fiber_testcancel();
```
locker, line 1316: Should be checked after `fiber_cond_wait_deadline`? Anyway, do we need to raise an exception in this case while we could just return as we do in case we failed to form a quorum?

src/box/vy_scheduler.c
```diff
@@ -734,6 +734,10 @@ vy_scheduler_wait_checkpoint(struct vy_scheduler *scheduler)
 			return -1;
 		}
 		fiber_cond_wait(&scheduler->dump_cond);
+		if (fiber_is_cancelled()) {
+			diag_set(FiberIsCancelled);
+			return -1;
```
locker, line 739: Better log an error in this case as we do for other errors.

src/lib/core/fiber.h
```diff
@@ -852,6 +852,10 @@ struct cord {
 	struct fiber *main_fiber;
 	/** An event triggered to cancel cord main fiber. */
 	ev_async cancel_event;
+	/** Number of alive client (non system) fibers. */
+	unsigned int client_fiber_count;
```
locker, line 856: Nit: Please don't use `unsigned` for counters. Use `int`.

src/lib/core/fiber.h
```diff
@@ -1243,6 +1247,28 @@ fiber_check_gc(void);
...
 
+/** Change whether fiber is system or not. */
+static inline void
+fiber_set_system(struct fiber *f, bool yesno)
+{
+	if (yesno) {
+		if (!(f->flags & FIBER_IS_SYSTEM)) {
+			f->flags |= FIBER_IS_SYSTEM;
+			assert(cord()->client_fiber_count > 0);
+			cord()->client_fiber_count--;
```
locker, line 1258: Shouldn't we wake up the shutdown fiber in this case?

src/lib/core/fiber_pool.c
```diff
@@ -61,7 +61,9 @@ fiber_pool_f(va_list ap)
 			f->caller->flags |= FIBER_IS_READY;
 			assert(f->caller->caller == &cord->sched);
 		}
+		fiber_set_system(fiber(), false);
```
locker, line 64: This means that all new fibers created by a pool should be system, but I don't see where they are marked so.

Also, there should be a comment somewhere explaining why pool fibers are marked system.

locker (changes requested):

src/box/lua/console.lua
nshy: Changed to close socket to peep instead as dicussed f2f.

New version of console patch:

```diff
--- a/src/box/lua/console.lua
+++ b/src/box/lua/console.lua
@@ -985,7 +985,15 @@ local function client_handler(client, _peer)
     state:print(string.format("%-63s\n%-63s\n",
         "Tarantool ".. version.." (Lua console)",
         "type 'help' for interactive help"))
+    local on_shutdown = function()
+        -- Fiber is going to be cancelled on shutdown. Do not report
+        -- cancel induced error to the peer.
+        client:close();
+    end
+    state.fiber = fiber.self()
+    box.ctl.on_shutdown(on_shutdown)
     repl(state)
+    box.ctl.on_shutdown(nil, on_shutdown)
     session_internal.run_on_disconnect()
 end
```

src/box/vy_scheduler.c
nshy: Done.

```diff
--- a/src/box/vy_scheduler.c
+++ b/src/box/vy_scheduler.c
@@ -730,17 +730,20 @@ vy_scheduler_wait_checkpoint(struct vy_scheduler *scheduler)
                        /* A dump error occurred, abort checkpoint. */
                        struct error *e = diag_last_error(&scheduler->diag);
                        diag_set_error(diag_get(), e);
-                       say_error("vinyl checkpoint failed: %s", e->errmsg);
-                       return -1;
+                       goto error;
                }
                fiber_cond_wait(&scheduler->dump_cond);
                if (fiber_is_cancelled()) {
                        diag_set(FiberIsCancelled);
-                       return -1;
+                       goto error;
                }
        }
        say_info("vinyl checkpoint completed");
        return 0;
+error:
+       say_error("vinyl checkpoint failed: %s",
+                 diag_last_error(diag_get())->errmsg);
+       return -1;
 }
```

src/lib/core/fiber.h
nshy: Fixed.

<details>
<summary>diff</summary>

```diff
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -853,7 +853,7 @@ struct cord {
        /** An event triggered to cancel cord main fiber. */
        ev_async cancel_event;
        /** Number of alive client (non system) fibers. */
-       unsigned int client_fiber_count;
+       int client_fiber_count;
        /** Fiber calling fiber_shutdown. NULL if there is no such. */
        struct fiber *shutdown_fiber;
 };
diff --git a/test/unit/fiber.cc b/test/unit/fiber.cc
index 4a774420e..82785006e 100644
--- a/test/unit/fiber.cc
+++ b/test/unit/fiber.cc
@@ -586,7 +586,7 @@ fiber_test_client_fiber_count(void)
 {
        header();

-       unsigned int count = cord()->client_fiber_count;
+       int count = cord()->client_fiber_count;

        struct fiber *fiber1 = fiber_new("fiber1", wait_cancel_f);
        fail_unless(fiber1 != NULL);
@@ -634,7 +634,7 @@ fiber_test_set_system(void)

        struct fiber *fiber1 = fiber_new("fiber1", wait_cancel_f);
        fail_unless(fiber1 != NULL);
-       unsigned int count = cord()->client_fiber_count;
+       int count = cord()->client_fiber_count;

        fiber_set_system(fiber1, true);
        fail_unless(--count == cord()->client_fiber_count);
```

</details>

src/lib/core/fiber.h
nshy: Fixed.

<details>
<summary>diff</summary>

```diff
diff --git a/src/lib/core/fiber.c b/src/lib/core/fiber.c
index 5c4aed7fe..26bc4459e 100644
--- a/src/lib/core/fiber.c
+++ b/src/lib/core/fiber.c
@@ -2351,6 +2351,26 @@ fiber_lua_state(struct fiber *f)
 	return f->storage.lua.stack;
 }
 
+void
+fiber_set_system(struct fiber *f, bool yesno)
+{
+	if (yesno) {
+		if (!(f->flags & FIBER_IS_SYSTEM)) {
+			f->flags |= FIBER_IS_SYSTEM;
+			assert(cord()->client_fiber_count > 0);
+			cord()->client_fiber_count--;
+			if (cord()->shutdown_fiber != NULL &&
+			    cord()->client_fiber_count == 0)
+				fiber_wakeup(cord()->shutdown_fiber);
+		}
+	} else {
+		if (f->flags & FIBER_IS_SYSTEM) {
+			f->flags &= ~FIBER_IS_SYSTEM;
+			cord()->client_fiber_count++;
+		}
+	}
+}
+
 void
 fiber_shutdown(void)
 {
diff --git a/src/lib/core/fiber.h b/src/lib/core/fiber.h
index f356bc45d..e58553bdd 100644
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -1248,22 +1248,8 @@ struct lua_State *
 fiber_lua_state(struct fiber *f);
 
 /** Change whether fiber is system or not. */
-static inline void
-fiber_set_system(struct fiber *f, bool yesno)
-{
-	if (yesno) {
-		if (!(f->flags & FIBER_IS_SYSTEM)) {
-			f->flags |= FIBER_IS_SYSTEM;
-			assert(cord()->client_fiber_count > 0);
-			cord()->client_fiber_count--;
-		}
-	} else {
-		if (f->flags & FIBER_IS_SYSTEM) {
-			f->flags &= ~FIBER_IS_SYSTEM;
-			cord()->client_fiber_count++;
-		}
-	}
-}
+void
+fiber_set_system(struct fiber *f, bool yesno);
 
 /** Cancel all client (non system) fibers and wait until they finished. */
 void
diff --git a/test/unit/fiber.cc b/test/unit/fiber.cc
index 82785006e..7caab3a8e 100644
--- a/test/unit/fiber.cc
+++ b/test/unit/fiber.cc
@@ -682,6 +682,17 @@ fiber_test_set_system(void)
 	footer();
 }
 
+static int
+hang_on_cancel_f(va_list ap)
+{
+	while (!fiber_is_cancelled())
+		fiber_yield();
+	fiber_set_system(fiber(), true);
+	while (true)
+		fiber_yield();
+	return 0;
+}
+
 static void
 fiber_test_shutdown(void)
 {
@@ -692,10 +703,13 @@ fiber_test_shutdown(void)
 	fiber_set_joinable(fiber1, true);
 	struct fiber *fiber2 = fiber_new_system("fiber2", wait_cancel_f);
 	fail_unless(fiber2 != NULL);
+	struct fiber *fiber3 = fiber_new("fiber3", hang_on_cancel_f);
+	fail_unless(fiber3 != NULL);
 
 	fiber_shutdown();
 	fail_unless((fiber1->flags & FIBER_IS_DEAD) != 0);
 	fail_unless((fiber2->flags & FIBER_IS_DEAD) == 0);
+	fail_unless((fiber3->flags & FIBER_IS_DEAD) == 0);
 
 	fiber_join(fiber1);
 ```

</details>

File: src/lib/core/fiber_pool.c (hunk 14 above)
Comment (nshy) [reply src/lib/core/fiber_pool.c:64]: Let's make intention explicit thru use of `fiber_new_system` for creating pool fiber. Also add a comment.

```diff
--- a/src/lib/core/fiber_pool.c
+++ b/src/lib/core/fiber_pool.c
@@ -133,7 +133,17 @@ fiber_pool_cb(ev_loop *loop, struct ev_watcher *watcher, int eve>
                        f = rlist_shift_entry(&pool->idle, struct fiber, state);
                        fiber_call(f);
                } else if (pool->size < pool->max_size) {
-                       f = fiber_new(cord_name(cord()), fiber_pool_f);
+                       /*
+                        * We don't want fibers to be cancellable by client
+                        * while they are in the pool. However system flag is
+                        * reset during proccessing message from pool endpoint
+                        * so that fiber is made cancellable back.
+                        *
+                        * If some message processing should not be cancellable
+                        * by client then it can just set system flag during
+                        * it's execution.
+                        */
+                       f = fiber_new_system(cord_name(cord()), fiber_pool_f);
                        if (f == NULL) {
                                diag_log();
                                break;
```

ochaplashkin (approved):

locker (approved):

PR #9606, opened 2024-01-19
raft: leader steps off on `ER_WAL_IO`

### LINKED ISSUE #9399: Raft leader should step off when encounters ER_WAL_IO
**Bug description**

Raft leader continues to send heartbeats even if it can't write anything, so it is seen by others as alive. As a result, once the leader encounters a disk error, it can't write anything, but the elections do not start.

**Steps to reproduce**

1. Start 3 nodes with `election_mode = 'candidate'`, grant necessary replication privileges
2. On the first chosen leader, issue `box.error.injection.set("ERRINJ_WAL_WRITE", true)`
3. Try to write something, for example:
```
tarantool> box.space._schema:replace{'a'}
2023-11-24 11:01:34.565 [13307] wal/101/main xlog.c:1276 E> ER_INJECTION: Error injection 'xlog write injection'
2023-11-24 11:01:34.565 [13307] main/103/interactive txn.c:1069 E> ER_WAL_IO: Failed to write to disk
---
- error: Failed to write to disk
...

```
4. Check `box.info.election` and `box.info.replication` on the leader. Nothing changes:
```
tarantool> box.info.election
---
- leader_idle: 0
  leader_name: null
  state: leader
  vote: 1
  term: 2
  leader: 1
...

---
- 1:
    id: 1
    uuid: 676834e6-1c9d-426d-92a5-419fa8087424
    lsn: 4
    name: null
  2:
    id: 2
    uuid: 6a8132e2-1ffa-4818-92f4-6caad586709c
    lsn: 0
    upstream:
      status: follow
      idle: 0.78926000000001
      peer: '3303'
      lag: 0.00019121170043945
    name: null
    downstream:
      status: follow
      idle: 0.81881099999964
      vclock: {1: 4}
      lag: 0.00066828727722168
  3:
    id: 3
    uuid: e7dece36-6ae9-46b3-9967-bf3afedfe950
    lsn: 0
    upstream:
      status: follow
      idle: 0.75874699999986
      peer: '3302'
      lag: 0.00017905235290527
    name: null
    downstream:
      status: follow
      idle: 0.7589109999999
      vclock: {1: 4}
      lag: 0
...

```

Reported as part of #9383

Astronomax:
Prior to this patch, the raft leader continued to send heartbeats even if it can't write anything, so it is seen by others as alive. As a result, once the leader encounters a disk error, it can't write anything, but the elections do not start. Now the leader resigns on the first encounter with `ER_WAL_IO` write error.

Closes #9399

@TarantoolBot document
Title: raft: leader resigns on the first encounter with `ER_WAL_IO`

* Now the leader resigns on the first encounter with `ER_WAL_IO` write error (the leader broadcasts this).


ochaplashkin (approved):

Astronomax:
It is worth noting that in the `test_cluster_elect_new_leader test`, the leader does not have time to broadcast its state before it falls, so the remaining nodes understand that the leader is no longer valid by timeout. And in `test_leader_continues_work_after_one_write_error`, where only one write fails, the leader actually tells others that he is no longer a leader.

src/lib/raft/raft.c
```diff
@@ -1154,11 +1154,7 @@ raft_start_candidate(struct raft *raft)
 	}
 }
 
-/**
- * Make the instance stop taking part in new elections and demote if it was a
- * leader.
- */
-static void
+void
```
sergepetrenko, line 1157: You can keep the function static

src/box/wal.c
```diff
@@ -1185,6 +1186,7 @@ wal_write_to_disk(struct cmsg *msg)
 		assert(err_code == JOURNAL_ENTRY_ERR_UNKNOWN);
 	}
 	wal_notify_watchers(writer, WAL_EVENT_WRITE);
+
```
sergepetrenko, line 1189: Unnecessary change

sergepetrenko (changes requested):
Thanks for working on this!

locker (approved):

changelogs/unreleased/gh-9399-raft-leader-steps-off-on-ER_WAL_IO.md
```diff
@@ -0,0 +1,7 @@
+## feature/raft
+
+* Now the leader resigns on the first encounter with `ER_WAL_IO` write error.
+  Immediately after this, the node tries to write a new term. If this attempt
```
sergepetrenko, line 4: The just resigned node doesn't attempt to write a new term. It only does so if others bumped the term after its resign. Let's not mention this.

Same for the docbot request.

src/box/raft.c
```diff
@@ -301,6 +302,21 @@ box_raft_fence(void)
 	raft_resign(raft);
 }
 
+/**
+ * Resign RAFT leadership and freeze limbo.
```
sergepetrenko, line 306: Please describe here why this function is different from `box_raft_fence` above.
For example, this function ignores `election_fencing` setting and doesn't prevent the node from starting new elections.

sergepetrenko (approved):
Thanks for the fixes!

changelogs/unreleased/gh-9632-fix-box-watch-hangs.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/core
+
+* Fixed a bug where the tarantool could hang due to `box.watch` (gh-9632).
```
xuniq, line 3:
```suggestion
* Fixed a bug when Tarantool could hang due to `box.watch` (gh-9632).
```

changelogs/unreleased/gh-9399-raft-leader-steps-off-on-ER_WAL_IO.md
```diff
@@ -0,0 +1,4 @@
+## feature/raft
+
+* Now the leader resigns on the first encounter with `ER_WAL_IO` write
```
xuniq, line 3:
```suggestion
* Now the leader resigns on the first encounter with the `ER_WAL_IO` write
```

xuniq (approved):

changelogs/unreleased/gh-9632-fix-box-watch-hangs.md
Astronomax: Done.

changelogs/unreleased/gh-9399-raft-leader-steps-off-on-ER_WAL_IO.md
Astronomax: Done.

changelogs/unreleased/gh-9399-raft-leader-steps-off-on-ER_WAL_IO.md
Astronomax: Done.

src/box/raft.c
Astronomax: Done.

src/box/watcher.c
```diff
@@ -352,7 +352,7 @@ watchable_worker_f(va_list ap)
 		fiber_check_gc();
 		if (!watchable_run(watchable)) {
 			/* No more watchers to run, wait... */
-			fiber_sleep(TIMEOUT_INFINITY);
+			fiber_yield();
```
Gerold103, line 355: I don't understand what happens here. Both `fiber_sleep` and `fiber_yield` can be interrupted with `fiber_wakeup`. The only difference is that the sleep will wakeup on its own when the time expires which in case of this timeout would take quite a long time. Why is this change needed?

src/box/raft.c
```diff
@@ -301,6 +302,24 @@ box_raft_fence(void)
...
+ * round of elections.
+ */
+static void
+box_raft_leader_step_off(void)
+{
+	struct raft *raft = box_raft();
+	if (!raft->is_enabled || raft->state != RAFT_STATE_LEADER)
+		return;
+
+	txn_limbo_fence(&txn_limbo);
```
Gerold103, line 318: Who is going to unfence it later? Please, leave a comment.

src/lib/raft/raft.h
```diff
@@ -315,6 +315,10 @@ raft_process_heartbeat(struct raft *raft, uint32_t source);
 void
 raft_cfg_is_enabled(struct raft *raft, bool is_enabled);
 
+/** Make the instance a candidate. */
+void
+raft_start_candidate(struct raft *raft);
```
Gerold103, line 320: This isn't right. This is an internal Raft method. Raft behaviour must be manipulated either via the config `raft_cfg_...` methods or via the Rafts high level methods affecting its state temporarily. For example,  "restore", "resign", "promote" - they are high-level methods making sense for the Raft algorithm. The "start candidate" looks odd, because 1) there is no "stop candidate" (there is, but it is still internal), 2) you are calling in your code essentially stop + start when you do `raft_resign` + `raft_start_candidate`.

The combination of this pair of calls I can't grasp right away. What exactly are you trying to achieve by this? Does the comment on `box_raft_leader_step_off` mean, that you will start new term after the death timeout passes since the latest WAL error?

If these couple of calls do what I understood, then it also seems you can do it as `raft_resign` + `raft_restore`, can you? The latter call will sync your cfg with the actual ability to be a candidate. Which will basically make it a candidate again in this case. And then you don't need to export any new methods from raft.

src/box/raft.c
```diff
@@ -657,6 +683,12 @@ box_raft_init(void)
 		       NULL, NULL);
 	trigger_create(&box_raft_on_quorum_loss, box_raft_on_quorum_change_f,
 		       NULL, NULL);
+	struct watcher *watcher = xmalloc(sizeof(*watcher));
+	const char *key = "box.wal_error";
+	size_t key_len = strlen(key);
+	box_register_watcher(key, key_len, box_raft_on_wal_error_f,
+			     (watcher_destroy_f)free, 0,
+			     watcher);
```
Gerold103, line 691: Why do you need a watcher for this and not some sort of journal's internal `struct trigger` member? Watchers are public. Why is this event needed in public?

If there is a justification, then the WAL error event introduction should be in a separate commit with its own test, and the new event needs to be documented (AFAIK, we do document public events, do we?).

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
```diff
@@ -0,0 +1,92 @@
...
+
+-- Сheck that after one unsuccessful attempt to write the leader does not fails,
+-- but a new election round starts
+g.test_leader_continues_work_after_one_write_error = function(cg)
+    local term = cg.master:get_election_term()
+    cg.master:exec(function()
+        t.assert_equals(box.info.election.state, 'leader')
+        box.error.injection.set('ERRINJ_WAL_IO', true)
+        -- Only one write attempt fails
+        box.error.injection.set('ERRINJ_WAL_IO_COUNTDOWN', 0)
```
Gerold103, line 84: Why do you need a countdown error for this? Can you simply unset `ERRINJ_WAL_IO` manually back to false after the first error?

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
```diff
@@ -0,0 +1,92 @@
...
+    cg.cluster:drop()
+end)
+
+g.test_cluster_elect_new_leader = function(cg)
+    cg.replica:exec(function()
+        box.cfg{election_mode='candidate'}
+    end)
+    cg.master:exec(function()
+        t.assert_equals(box.info.election.state, 'leader')
+        box.error.injection.set('ERRINJ_WAL_IO', true)
```
Gerold103, line 64: The errinj must be unset back to false after the test case. Otherwise if any other test case would run after this one on the same instance, it would fail. In this case right now it is not so, but better not leave any errinj dangling anywhere. Code sometimes gets moved or copy-pasted and the bug could actually migrate to a place where it would manifest.

Gerold103 (changes requested):
Thanks for the patch!

src/box/watcher.c
Astronomax: There was a rather strange problem when using `fiber_sleep`. [I just added a call to `box_register_watcher`, after which the tarantool entered an infinite loop.](https://github.com/tarantool/tarantool/issues/9632)
`box_register_watcher` calls `watchable_wakeup_worker` which creates a worker fiber and immediately calls `fiber_wakeup` on it. However, the scheduler wakes up only after the first measurement of the number of events in `tarantool_lua_run_script`.
https://github.com/tarantool/tarantool/blob/ebb170cb8cf2b9c4634bcf0178665909f578c335/src/main.cc#L1058-L1065
After this, the `watchable->worker` wakes up and calls `fiber_sleep` which increases the number of events (`fiber_sleep` -> `fiber_yield_timeout` -> `ev_timer_start` -> `ev_start` -> `ev_ref`). As a result, we get that the number of events has increased after calling `tarantool_lua_run_script`.
https://github.com/tarantool/tarantool/blob/ebb170cb8cf2b9c4634bcf0178665909f578c335/src/main.cc#L1073-L1074
In summary, answering the question what is the difference between `fiber_yield` and `fiber_sleep` - the main difference in our context is that `fiber_sleep` generates a new event, while `fiber_yeild` does not.

src/box/raft.c
Astronomax: The WAL error event was already entered in https://github.com/tarantool/tarantool/pull/9422 and was also documented.
Of course, it would be possible to use a trigger here, however it seems that using wather has more advantages than disadvantages.
disadvantages of watchers compared to triggers:
* delivering messages in a separate fiber
* serialize data before delivering
However, the `watchable_node` "box.wal_error" already exists. That is, for every event there, the `watchable->worker` fiber already wakes up regardless of this patch. As for data serialization, in our case there is only one `uint64_t`. Moreover, data serialization occurs regardless of whether there are subscribed wathcers on watchable node. That is, adding a subscriber does not increase the number of serializations. I do not use the data itself in the handler, that is, I do not deserialize it.

advantages:
* We don't make the code too difficult to read. It will look strange if we call the trigger everywhere we call `broadcast_fmt("wal.error", ...`.

But I don’t know how important it is for us to comply with the accepted convention that we do not use watchers in system code. Is this a strict agreement?))

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
Astronomax: I may not have time to unset `ERRINJ_WAL_IO` manually back to false before the node tries to write a term and crashes. 
https://github.com/tarantool/tarantool/blob/ebb170cb8cf2b9c4634bcf0178665909f578c335/src/box/raft.c#L438-L442
And I have no way to stop the node before it tries to write. `ERROR_INJECT(ERRINJ_WAL_IO, ...` is located earlier than `ERROR_INJECT_SLEEP(ERRINJ_WAL_DELAY)`.
I don't know why `ERROR_INJECT(ERRINJ_WAL_IO, ...` is located in `wal_write_async`. It might be worth moving `ERROR_INJECT(ERRINJ_WAL_IO, ...` to `wal_write_to_disk`.
Do something like there:
```patch
 	ERROR_INJECT_SLEEP(ERRINJ_WAL_DELAY);

	ERROR_INJECT_COUNTDOWN(ERRINJ_WAL_DELAY_COUNTDOWN, {
		struct errinj *e = errinj(ERRINJ_WAL_DELAY, ERRINJ_BOOL);
		e->bparam = true;
		ERROR_INJECT_SLEEP(ERRINJ_WAL_DELAY);
	});

+	ERROR_INJECT(ERRINJ_WAL_IO, {
+		err_code = JOURNAL_ENTRY_ERR_IO;
+		goto done;
+	}

	if (writer->is_in_rollback) {
		/* We're rolling back a failed write. */
		err_code = JOURNAL_ENTRY_ERR_CASCADE;
		goto done;
	}

	/* Xlog is only rotated between queue processing  */
	if (wal_opt_rotate(writer) != 0) {
		err_code = JOURNAL_ENTRY_ERR_IO;
		goto done;
	}
```

changelogs/unreleased/gh-9399-raft-leader-steps-off-on-ER_WAL_IO.md
```diff
@@ -0,0 +1,4 @@
+## feature/raft
+
+* Now the leader resigns on the first encounter with `ER_WAL_IO` the
```
sergepetrenko, line 3:
```suggestion
* Now the leader resigns on the first encounter with the `ER_WAL_IO`
```

Astronomax: Done.

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
Astronomax: The master falls in this test, it is called panic. Therefore, there is no way to unset it back to false manually. 
I can do `box.error.injection.set('ERRINJ_WAL_IO_COUNTDOWN', 1)` and then `ERRINJ_WAL_IO` will automatically be set to false before the instance falls. But is this necessary if the instance cannot be reused anyway?

src/box/raft.c
Astronomax: Done.

src/lib/raft/raft.h
Astronomax: Ok, I understand the concept, it's convenient. Thanks! Done.

src/box/watcher.c
Gerold103: That seems like a bug in the wakeup code or any other code which relies on this difference. Please, file a separate ticket with this bug extracted into a simple example (AFAIU, you create a fiber which calls sleep and then you do wakeup on it and it doesn't work?).

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
Gerold103: So the instance crashes and the test passes? That doesn't sound right in terms of how luatest is handling it. Who knows where else our instances are crashing in the tests and we don't see it. Please, file a ticket in luatest repo if it isn't already there.

Gerold103 (approved):
Thanks for the patch! After fixing the comments above you can resolve them yourself and merge the patch. (Unless some of those comments would appear untrivial - then we shall have another look.)

src/box/watcher.c
Astronomax: This ticket has already been created #9632. There are several examples there on how to reproduce this.

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
Astronomax: I asked the developers of luatest. They said it was normal. luatest only detects segfaults, it ignores panics (non-zero return code). [luatest doesn't detect crash at exit](https://github.com/tarantool/luatest/issues/252) This is done so that you can test the panic call on an instance, as I do in this test.

test/replication-luatest/gh_9399_raft_leader_steps_off_on_ER_WAL_IO_test.lua
Astronomax: Done.

sergepetrenko:
Backport to 3.0: https://github.com/tarantool/tarantool/pull/9702

PR #9614, opened 2024-01-23
ci: fix fail on artifact upload to S3 for freebsd

ylobankov:
This patch fixes the following issue:

    Error: Unable to retrieve job ID by provided job name

The `${{ github.job }} (${{ join(matrix.*, ', ') }})` expression was evaluated to `freebsd (Object)` instead of proper job name. It looks like the ChristopherHX/github-act-runner agent has a bug somewhere. Now the job name is made manually to bypass the issue.


ochaplashkin (approved):

ylobankov:
Cherry picked to 3.0, 2.11, 2.10.

PR #9619, opened 2024-01-25
test: fix fix_lua_loader_error_handling on macOS

Buristan:
This patch is a follow-up for the commit
b99a35412d0c65dd84209872654fec9aaf2f1d02 ("lua: fix loader behaviour in case of syntax error").

This test uses full matching of the error message and assumes that the temporary directory is short enough, so the full patch is less than 60 symbols (size of `lua_Debug.short_src`). But for macOS, the temporary directory has a longer path, so the error message is shrunk and isn't equal to the expected string.

This patch replaces full matching with matching of the last part of the error message.

Relates to tarantool/tarantool-ee#585

NO_DOC=test fix
NO_CHANGELOG=test fix

---
I tested it on my macOS locally -- no more failures:).

### DIFF
--- a/test/app-luatest/fix_lua_loader_error_handling_test.lua
+++ b/test/app-luatest/fix_lua_loader_error_handling_test.lua
@@ -14,22 +14,16 @@ local g = t.group()
 -- the end of the test.
 local cwd = fio.cwd()
 
-local luadir
 local name = 'syntax'
-
--- XXX: Unfortunately, multiline literals do not interpolate
--- escape sequences, so just concatenate two lines of the error
--- message to build the final template.
-local errlike = table.concat({
-    "error loading module '<NAME>' from file '<DIR>/<NAME>.lua':",
-    "<DIR>/<NAME>.lua:1: unexpected symbol near '?'",
-}, "\n\t")
-
+-- XXX: luatest uses only `string.find()` with 4th argument
+-- (`plain`) unconditionally set to `true` for error matching, so
+-- there is no way to use any pattern here.
+local errtail = ("/%s.lua:1: unexpected symbol near '?'"):format(name)
 
 g.before_all(function()
     treegen.init(g)
     treegen.add_template(g, ('^%s%%.lua$'):format(name), [[?syntax error?]])
-    luadir = treegen.prepare_directory(g, {('%s.lua'):format(name)})
+    local luadir = treegen.prepare_directory(g, {('%s.lua'):format(name)})
     fio.chdir(luadir)
 end)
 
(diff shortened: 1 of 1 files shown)


Totktonada (approved):

test/app-luatest/fix_lua_loader_error_handling_test.lua
```diff
@@ -14,22 +14,16 @@ local g = t.group()
...
--- message to build the final template.
-local errlike = table.concat({
-    "error loading module '<NAME>' from file '<DIR>/<NAME>.lua':",
-    "<DIR>/<NAME>.lua:1: unexpected symbol near '?'",
-}, "\n\t")
-
+-- XXX: luatest uses only `string.find()` with 4th argument
+-- (`plain`) unconditionally set to `true` for error matching, so
+-- there is no way to use any pattern here.
+local errlike = ("/%s.lua:1: unexpected symbol near '?'"):format(name)
```
igormunkin, line 21: Since it's not pattern or format string anymore, so the variable is `errmsg`, isn't?

igormunkin (approved):
@Buristan, thanks for the fix! LGTM with a single nit below.

test/app-luatest/fix_lua_loader_error_handling_test.lua
Buristan: Replaced with `errtail`. Branch is force-pushed, see the iterative patch below.
```diff
diff --git a/test/app-luatest/fix_lua_loader_error_handling_test.lua b/test/app-luatest/fix_lua_loader_error_handling_test.lua
index 45d188616..6cf251121 100644
--- a/test/app-luatest/fix_lua_loader_error_handling_test.lua
+++ b/test/app-luatest/fix_lua_loader_error_handling_test.lua
@@ -18,7 +18,7 @@ local name = 'syntax'
 -- XXX: luatest uses only `string.find()` with 4th argument
 -- (`plain`) unconditionally set to `true` for error matching, so
 -- there is no way to use any pattern here.
-local errlike = ("/%s.lua:1: unexpected symbol near '?'"):format(name)
+local errtail = ("/%s.lua:1: unexpected symbol near '?'"):format(name)
 
 g.before_all(function()
     treegen.init(g)
@@ -33,5 +33,5 @@ g.after_all(function(g)
 end)
 
 g.test_loader_error_handling = function()
-    t.assert_error_msg_contains(errlike, require, name)
+    t.assert_error_msg_contains(errtail, require, name)
 end
```

ylobankov (approved):
LGTM

igormunkin:
The changeset is backported to release/3.0 in scope of c0acf75.

PR #9620, opened 2024-01-25
Generate changelog for 3.0.1

sergos:
Also, remove unreleased/ entries.

NO_DOC=changelog
NO_TEST=changelog
NO_CHANGELOG=changelog

### DIFF
--- a/changelogs/3.0.1.md
+++ b/changelogs/3.0.1.md
@@ -1,0 +1,48 @@
+# 3.0.1
+
+Date: 2024-01-25
+Tag: 3.0.1
+
+## Overview
+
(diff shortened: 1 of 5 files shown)


changelogs/3.0.1.md
```diff
@@ -0,0 +1,48 @@
+# 3.0.1
```
Totktonada, line 1: We should strip changelog entries that reside in the separate files, when compose them into release notes. See https://github.com/tarantool/tarantool/blob/master/doc/changelogs.md#generate-changelog-file

sergos: Almost broken with "strip" in relation to files removal :)
Fixed.

ylobankov (approved):
LGTM

PR #9622, opened 2024-01-26
trigger: introduce persistent triggers

### LINKED ISSUE #8663: Persistent triggers
When the universal triggers registry will be implements, it won't be difficult to implement persistent triggers. The design is as follows: `box.schema.func.create` function is populated with new option `trigger`, which if of the string type. When a tuple in space `_func` with non-NULL field `trigger` is being replaced, it is added (or deleted, if the field is dropped) to the trigger registry - `trigger` field is an event name and name of the function is the name of a new trigger.

drewdzzz:
After this patch, functions with non-empty option `trigger` will be
inserted into the trigger registry. The option specifies event names in
which the function will be inserted as a trigger.

When a function is created, it is set as a trigger without any checks -
it can replace an existing one. When it is deleted, it can delete a
trigger it hasn't set - for example, it can happen when user manually
replaces a persistent trigger. So, it is recommended to have different
names (or even separate namespaces) for persistent and usual triggers.

Note that both persistent and non-persistent funcs can be used as
triggers. Also, when the triggers are called, access rights are
not checked.

**NB**: the first commit should be backported to 3.0

Closes #8663


changelogs/unreleased/gh-8663-persistent-triggers.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Added a `trigger` option to `box.schema.func.create`. It allows to create
```
andreyaksenov, line 3:
```suggestion
* Added a `trigger` option to `box.schema.func.create`. This option allows creating
```

andreyaksenov (approved):

src/box/lua/upgrade.lua
```diff
@@ -1942,10 +1958,38 @@ local function drop_instance_names(issue_handler)
...
+            end
+        end
+        return
+    end
+    log.info('drop trigger from _func')
+    local ops = {{'#', '[7][20]', 1}}
+    _space:update({_func.id}, ops)
+    _space:update({_vfunc.id}, ops)
+    for _, v in _func:pairs() do
+        _func:update(v[1], {{'#', 20, 1}})
```
locker, line 1984: AFAIU we shouldn't drop triggers on downgrade. We should refuse to downgrade if there are triggers set.

src/box/lua/trigger.c
```diff
@@ -226,7 +346,9 @@ luaT_trigger_pairs(struct lua_State *L)
 		lua_pushcfunction(L, luaT_trigger_iterator_next_exhausted);
 		return 1;
 	}
-	lua_pushcfunction(L, luaT_trigger_iterator_next);
+	size_t *epoch = lua_newuserdata(L, sizeof(size_t));
```
locker, line 349: I don't understand why we need to pass the "epoch" as a Lua object. I think we should create a weak reference object for functions, like we did for index, see [`index_weak_ref.h`](https://github.com/tarantool/tarantool/blob/58f5ffe43bffa0901d8d577bbb4d523c4e7e682a/src/box/index_weak_ref.h).

changelogs/unreleased/gh-8663-persistent-triggers.md
locker, line 1: Please add some code examples to the doc bot request illustrating how the new API is supposed to be used.

locker (changes requested):

src/box/func_adapter.c
```diff
@@ -0,0 +1,81 @@
...
+#include "box/port.h"
+#include "box/func.h"
+#include "box/func_cache.h"
+
+#include "core/func_adapter.h"
+
+/**
+ * Specialization of func_adapter for persistent functions.
+ */
+struct func_adapter_func {
```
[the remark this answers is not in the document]
drewdzzz: Discussed f2f, left it as it was.

src/box/alter.cc
```diff
@@ -3259,6 +3259,29 @@ func_def_new_from_tuple(struct tuple *tuple)
...
+			tuple_field_with_type(tuple, BOX_FUNC_FIELD_TRIGGER,
+					      MP_ARRAY);
+		if (triggers == NULL)
+			return NULL;
+		const char *triggers_begin = triggers;
+		uint32_t trigger_count = mp_decode_array(&triggers);
+		for (uint32_t i = 0; i < trigger_count; i++) {
+			enum mp_type actual_type = mp_typeof(*triggers);
+			if (actual_type != MP_STR) {
+				diag_set(ClientError, ER_FIELD_TYPE,
```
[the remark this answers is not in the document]
drewdzzz: Fixed. Along the way, fixed the error in `param_list` check (I actually took this check from there).

src/box/func_def.c
```diff
@@ -143,6 +144,35 @@ func_def_cmp(const struct func_def *def1, const struct func_def *def2)
...
 	if (def1->comment != NULL && strcmp(def1->comment, def2->comment) != 0)
 		return strcmp(def1->comment, def2->comment);
+	/*
+	 * The field is set only if underlying array is not empty. So an empty
+	 * array is equivalent to the field absence (NULL) here.
+	 */
+	if ((def1->triggers != NULL) != (def2->triggers != NULL))
+		return def1->triggers - def2->triggers;
+	/* See the function description for equality definition. */
+	if (def1->triggers != NULL) {
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/func_def.c
```diff
@@ -161,6 +191,7 @@ func_def_dup(const struct func_def *def)
 	copy->aggregate = def->aggregate;
 	copy->exports.all = def->exports.all;
 	copy->opts = def->opts;
+	copy->triggers = def->triggers;
```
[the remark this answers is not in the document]
drewdzzz: It was a pointer to the field of tuple stored in `_func`. I reworked this part, so now I copy the triggers here.

src/box/lua/upgrade.lua
drewdzzz: Done, also added a test for this case.

src/box/lua/trigger.c
drewdzzz: Discussed f2f, this part was reworked.

changelogs/unreleased/gh-8663-persistent-triggers.md
drewdzzz: Added two examples: single-node and cluster scenarios.
Also, I added a test for a replicated persistent trigger.

src/box/lua/upgrade.lua
```diff
@@ -1394,9 +1394,25 @@ local function add_instance_names()
...
+    end
+    local ops = {{'=', '[7][20]', {name = 'trigger', type = 'array'}}}
+    _space:update({_func.id}, ops)
+    _space:update({_vfunc.id}, ops)
+end
+
 local function upgrade_to_3_0_0()
     store_replicaset_uuid_in_new_way()
     add_instance_names()
+    add_trigger_to_func()
```
locker, line 1415: 3.0.0 has already been released. You can't update the past. You must create an upgrade handler for 3.1.0.

src/box/lua/upgrade.lua
```diff
@@ -1942,10 +1958,41 @@ local function drop_instance_names(issue_handler)
...
+    local msg_suffix = 'has trigger option. It is supported from version 3.0.0'
+    if #_func:format() == 19 then
+        return
+    end
+    if issue_handler.dry_run then
+        for _, v in _func:pairs() do
+            -- Non-empty trigger option cannot be dropped because
+            -- func does not support alter.
+            if #v > 19 and not table.equals(v[20], {}) then
+                issue_handler('Func %s %s', v.name, msg_suffix)
```
locker, line 1974: This looks ugly. Better write:

```lua
local fmt = 'Function %s is registered as event trigger. It is supported starting from version 3.1.0"
issue_handler(fmt, v.name)

File: src/box/lua/upgrade.lua [hunk 10]
```diff
@@ -1942,10 +1958,41 @@ local function drop_instance_names(issue_handler)
...
+            end
+        end
+        return
+    end
+    log.info('drop trigger from _func')
+    local ops = {{'#', '[7][20]', 1}}
+    _space:update({_func.id}, ops)
+    _space:update({_vfunc.id}, ops)
+    for _, v in _func:pairs() do
+        if #v > 19 and not table.equals(v[20], {}) then
```
locker, line 1984: Having this check in two loops looks ugly. Better check the dry run flag in one loop:

```lua
for ... do
    if ... then
        issue_handler(...)
    end
    if not dry_run then
        ...
    end
end
```

src/box/lua/func_adapter.c
locker, line 1: > ```
> The patch makes a small breaking change - `trigger.del` returns boolean
> (true if a trigger was actually deleted, false otherwise) instead of
> handler. Rationale: actually, lifetime of func_adapter is independent
> from Lua, so when you are deleting a trigger, it's fair to consider that
> the handler will be destroyed as well. So, it's just weird to return
> a reference to object from its destructor, so let's return flag instead.
> ```

Looks like this refactoring was done in a separate patch.

src/box/lua/trigger.c
```diff
@@ -74,20 +135,62 @@ luaT_trigger_call(struct lua_State *L)
...
+	 * the underlying one, and it's cheaper to leave fiber storage stack
+	 * to triggers - they won't have to create their own stacks then.
+	 *
+	 * If we have no args, port won't be created, so we don't need to
+	 * create a stack.
+	 */
+	struct lua_State *args_L = L;
+	int args_coro_ref = LUA_REFNIL;
+	if (nargs > 0) {
+		if (fiber()->storage.lua.stack == L) {
```
locker, line 152: Why can they be different?

src/box/lua/trigger.c
```diff
@@ -74,20 +135,62 @@ luaT_trigger_call(struct lua_State *L)
 	struct event *event = event_get(event_name, false);
 	if (event == NULL)
 		return 0;
-	int top = lua_gettop(L);
-	int narg = top - 1;
+	int nargs = lua_gettop(L) - 1;
+
+	/*
+	 * If the function is called on Lua-stack from fiber storage, create
+	 * new thread because we cannot dump port_lua to the same stack as
```
locker, line 142: Can you make `port_lua` work with the same stack?

src/box/lua/trigger.c
```diff
@@ -74,20 +135,62 @@ luaT_trigger_call(struct lua_State *L)
...
-		func_adapter_lua_get_func(trigger, L);
-		for (int i = top - narg + 1; i <= top; ++i)
-			lua_pushvalue(L, i);
-		rc = luaT_call(L, narg, 0);
+		/*
+		 * Push arguments before each trigger - they will be popped when
+		 * the trigger is invoked. If nargs is 0, loop won't be started.
+		 */
+		for (int i = 1; i <= nargs; i++)
+			lua_pushvalue(args_L, i);
```
locker, line 186: Looks like a hack. Better create the port before each call.

src/box/lua/trigger.c
```diff
@@ -11,6 +11,67 @@
...
+	struct port args, ret;
+	port_lua_create(&args, L);
+	bool fiber_storage_is_used = fiber()->storage.lua.stack == L;
+	/*
+	 * Prevent using the same Lua-stack in trigger because port_lua
+	 * doesn't allow to dump it to the same stack as the underlying one.
+	 */
+	if (fiber_storage_is_used)
+		fiber()->storage.lua.stack = NULL;
+	int rc = func_adapter_call(trigger, &args, &ret);
```
locker, line 46: Looks like code duplication with `luaT_trigger_call`.

test/box-luatest/gh_8663_persistent_triggers_test.lua
```diff
@@ -0,0 +1,306 @@
...
+g.test_recovery = function()
+    g.server:exec(function()
+        local trigger = require('trigger')
+        local max_i = 5
+        local event_name = 'persistent_event'
+        trigger.set(event_name, 'not_truly_persistent' .. tostring(max_i + 1),
+                    function() end)
+        for i = max_i, 1, -1 do
+            box.schema.func.create('truly_persistent' .. tostring(i), {
+                language = 'lua', body = 'function() end', trigger = event_name
```
locker, line 28: So it's not a trigger name - it's an event name, actually. Maybe, better call it `on_event`, not `trigger`?

src/box/alter.cc
```diff
@@ -3387,6 +3419,19 @@ on_replace_dd_func(struct trigger * /* trigger */, void *event)
...
+			txn_alter_trigger_new(on_drop_func_commit, old_func);
+		struct trigger *on_rollback =
+			txn_alter_trigger_new(on_drop_func_rollback, old_func);
+		if (on_commit == NULL || on_rollback == NULL)
+			return -1;
+		txn_stmt_on_commit(stmt, on_commit);
+		txn_stmt_on_rollback(stmt, on_rollback);
+
+		/* Triggers pin the function - drop them before pin check. */
+		func_alter_triggers(old_func, false);
```
locker, line 3433: Why do we need to pin the function by triggers if the triggers are deleted automatically when the function is dropped?

locker (changes requested):

src/box/lua/trigger.c
locker: Decided to move this check to `func_adapter_lua_call`: `func_adapter_lua_call` will check if the port storing arguments is a Lua port and if it is and points to the current fiber Lua stack, create a new stack for it.

src/box/lua/trigger.c
locker: Decided to move this to `port_lua_dump`, i.e. make it possible to call this function multiple times on the same port.

src/box/lua/trigger.c
```diff
@@ -52,14 +52,9 @@ luaT_trigger_del(struct lua_State *L)
...
-		return 0;
-	struct func_adapter *old = event_find_trigger(event, trigger_name);
-	if (old == NULL)
-		return 0;
-	func_adapter_lua_get_func(old, L);
-	event_reset_trigger(event, trigger_name, NULL);
-	return 1;
+	if (event != NULL)
+		event_reset_trigger(event, trigger_name, NULL);
+	return 0;
```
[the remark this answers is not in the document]
drewdzzz: Decided to leave it without return value.

src/box/CMakeLists.txt
```diff
@@ -237,6 +237,7 @@ set(box_sources
     space_def.c
     sequence.c
     func.c
+    func_adapter.c
```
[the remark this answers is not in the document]
drewdzzz: Sorry, forgot that we discussed it. Done.

src/box/func_adapter.c
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Virtual table.
+	 */
+	const struct func_adapter_vtab *vtab;
+	/**
+	 * Reference to the function itself. Must be in func_cache. Is pinned.
+	 */
+	struct func *func;
```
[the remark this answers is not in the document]
drewdzzz: Dropped the field.

src/box/func_adapter.c
```diff
@@ -0,0 +1,81 @@
...
+#include "core/func_adapter.h"
+
+/**
+ * Specialization of func_adapter for persistent functions.
+ */
+struct func_adapter_func {
+	/**
+	 * Virtual table.
+	 */
+	const struct func_adapter_vtab *vtab;
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/func_adapter.h
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Copyright 2010-2023, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "box/func_cache.h"
```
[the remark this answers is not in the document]
drewdzzz: Outdated: the file was removed.

src/box/alter.cc
```diff
@@ -3147,9 +3147,31 @@ func_def_new_from_tuple(struct tuple *tuple)
...
+						 MP_ARRAY);
+		if (triggers == NULL)
+			return NULL;
+		const char *triggers_cursor = triggers;
+		uint32_t trigger_count = mp_decode_array(&triggers_cursor);
+		for (uint32_t i = 0; i < trigger_count; i++) {
+			enum mp_type actual_type = mp_typeof(*triggers_cursor);
+			if (actual_type != MP_STR) {
+				diag_set(ClientError, ER_CREATE_FUNCTION,
+					 name, "each trigger must be a string");
```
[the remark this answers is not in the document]
drewdzzz: That's better, thanks. Done.

src/box/alter.cc
```diff
@@ -3228,10 +3250,9 @@ func_def_new_from_tuple(struct tuple *tuple)
 		for (uint32_t i = 0; i < argc; i++) {
 			enum mp_type actual_type = mp_typeof(*param_list);
 			if (actual_type != MP_STR) {
-				diag_set(ClientError, ER_FIELD_TYPE,
-					 int2str(BOX_FUNC_FIELD_PARAM_LIST + 1),
-					 mp_type_strs[MP_STR],
-					 mp_type_strs[actual_type]);
+				diag_set(ClientError, ER_CREATE_FUNCTION,
+					 def->name,
+					 "each param must be a string");
```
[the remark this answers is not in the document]
drewdzzz: Thanks, done.

src/box/func_def.h
```diff
@@ -104,6 +104,11 @@ struct func_def {
 	char *body;
 	/** User-defined comment for a function. */
 	char *comment;
+	/**
+	 * MsgPack array of strings - names of events in trigger registry.
+	 * If it is set, underlying array is not empty. Is not zero-terminated.
```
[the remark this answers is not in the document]
drewdzzz: Agree, dropped this sentence.

src/box/func_def.h
```diff
@@ -145,18 +150,23 @@ struct func_def {
...
 
 /** Frees a function definition object. */
 void
 func_def_delete(struct func_def *def);
 
-/** Compare two given function definitions. */
+/**
+ * Compare two given function definitions.
+ * Two triggers fields (which are pointers to a MsgPack array) are compared
+ * byte wise.
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/upgrade.lua
drewdzzz: Sorry, I forgot about the new release. I created new handlers for upgrade and downgrade.

src/box/lua/upgrade.lua
drewdzzz: Done.

src/box/lua/upgrade.lua
drewdzzz: Done.

src/box/lua/func_adapter.c
drewdzzz: Dropped this part.

src/box/lua/trigger.c
drewdzzz: This part was reworked, so the comment is outdated.

src/box/lua/trigger.c
drewdzzz: Done.
By the way, now one empty on_replace trigger slows Tarantool down by 27%, but without the optimization slowdown is about 42%.

src/box/lua/trigger.c
drewdzzz: Done as a separate commit.

src/box/lua/trigger.c
drewdzzz: After rework these function are not similar.

test/box-luatest/gh_8663_persistent_triggers_test.lua
drewdzzz: We also operate with term `event` in `box.watch`, and there we have keys, which are strings as well, so `on_event` will be confisung.
The idea of this name is that it refers to the module `trigger`.

src/box/alter.cc
drewdzzz: It is needed to forbid deleting triggers that are being fired. The triggers are deleted from the event, which is actually unreferencing, not a deletion. So if the trigger is used at the moment, it won't be deleted, and the function will be still pinned - then the alter will fail.

src/box/lua/upgrade.lua
```diff
@@ -1394,9 +1394,25 @@ local function add_instance_names()
     box.space._space:update({box.schema.CLUSTER_ID}, {{'=', 7, format}})
 end
 
+local function add_trigger_to_func()
+    log.info('add trigger to _func')
+    local _space = box.space[box.schema.SPACE_ID]
+    local _func = box.space[box.schema.FUNC_ID]
+    local _vfunc = box.space[box.schema.VFUNC_ID]
+    for _, v in _func:pairs() do
+        if #v == 19 then
```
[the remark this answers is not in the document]
drewdzzz: Just found a place with such check in `upgrade.lua` and made it similar. Actually, I don't need it, so it was dropped.

locker (changes requested):

src/box/alter.cc
locker: So if a trigger runs very often, it'll be quite hard to delete it?

locker:
@nshy Please review the upgrade/downgrade part.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
         t.assert_equals(#box.space._cluster:format(), 2)
     end)
 end
+
+-----------------------------
+-- Check downgrade from 3.1.0
+-----------------------------
+
+g.test_downgrade_func_trigger = function(cg)
```
nshy, line 981: Nit. Please fix indentation in this function.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
...
+        local trigger = require('trigger')
+        local _func = box.space._func
+        local prev_version = helper.prev_version(helper.app_version('3.0.0'))
+        local lua_code = 'function(a, b) return a + b end'
+        t.assert_equals(trigger.info('test'), {})
+        box.schema.func.create('trigger_func',
+            {body = lua_code, trigger = 'test'})
+        local errmsg = 'Function trigger_func is registered as ' ..
+            'event trigger. It is supported starting from version 3.1.0'
+        t.assert_equals(box.schema.downgrade_issues(prev_version), {errmsg})
```
nshy, line 993: Let's 
- check that we set correctly from version in downgrade. That is there is no with issues downgrading to `helper.app_version('3.1.0')` when persistent trigger is present (see `test_downgrade_schema_max_id` for example).
- Nit. check that `downgrade_issues` works with correctly with several issues.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
...
+        -- Check if cannot actually downgrade
+        t.assert_error_msg_content_equals(errmsg, box.schema.downgrade,
+            prev_version)
```
nshy, line 996: Let's at least test for `errmsg`. `prev_version` can be from issue of a different kind.

Also let's test that we don't drop 20-th column by accident (as under certain conditions we drop it).

I would also check that `non_trigger_function` is not updated in this case if present.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
...
+        box.schema.func.drop('trigger_func')
+        box.schema.func.create('non_trigger_func',
+            {body = lua_code, trigger = {}})
+        t.assert_equals(box.schema.downgrade_issues(prev_version), {})
+        t.assert_equals(#_func:format(), 20)
+        t.assert_equals(_func:format()[20].name, 'trigger')
```
nshy, line 1002: For consistency I would do similar checks above after `downgrade_issues`.

Let's also add comments for every part of the test for consistency. Like `Check if cannot actually downgrade`. Now it is kind of outstanding.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
...
+        -- 2 for idempotence.
+        for _ = 1, 2 do
+            box.schema.downgrade(prev_version)
+            t.assert_equals(#_func:format(), 19)
+            t.assert_equals(trigger.info('test'), {})
+            local func_tuple = _func.index.name:get('non_trigger_func')
+            t.assert_equals(#func_tuple, 19)
+        end
```
nshy, line 1010: I guess `trigger.info` test here and above is not the point of this test.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,40 @@ g.test_downgrade_global_names = function(cg)
...
+        t.assert_error_msg_content_equals(errmsg, box.schema.downgrade,
+            prev_version)
+        box.schema.func.drop('trigger_func')
+        box.schema.func.create('non_trigger_func',
+            {body = lua_code, trigger = {}})
+        t.assert_equals(box.schema.downgrade_issues(prev_version), {})
+        t.assert_equals(#_func:format(), 20)
+        t.assert_equals(_func:format()[20].name, 'trigger')
+        -- 2 for idempotence.
+        for _ = 1, 2 do
```
nshy, line 1004: Nice usage of loop for idempotence check!

nshy (changes requested):

src/box/lua/upgrade.lua
```diff
@@ -2137,7 +2137,9 @@ end
 box.schema.downgrade_issues = function(version)
     return downgrade_impl(version, true)
 end
-box.internal.bootstrap = bootstrap;
+box.internal.bootstrap = function()
+    run_upgrade(bootstrap)
+end
```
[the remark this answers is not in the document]
drewdzzz: At the moment when `clear_system_formats` is called, the triggers are enabled, so the [check](https://github.com/tarantool/tarantool/blob/60fdffbb6c2850e6d4a8436cec6162a595d863d1/src/box/schema.cc#L201) fails.

src/box/lua/call.c
```diff
@@ -566,7 +566,9 @@ port_lua_do_dump_with_ctx(struct port *base, struct mpstream *stream,
 	const int size = lua_gettop(L) - port->bottom + 1;
 	lua_rawgeti(L, LUA_REGISTRYINDEX, execute_lua_refs[handler]);
 	assert(lua_isfunction(L, -1) && lua_iscfunction(L, -1));
-	lua_insert(L, port->bottom);
+	/* Duplicate values. */
+	for (int i = 0; i < size; i++)
+		lua_pushvalue(L, port->bottom + i);
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/lua/call.c
```diff
@@ -622,6 +624,9 @@ port_lua_dump_lua(struct port *base, struct lua_State *L,
 	if (mode == PORT_DUMP_LUA_MODE_FLAT) {
 		struct port_lua *port = (struct port_lua *)base;
 		uint32_t size = lua_gettop(port->L) - port->bottom + 1;
+		/* Duplicate values. */
```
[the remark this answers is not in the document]
drewdzzz: Done.

src/box/iproto.cc
```diff
@@ -42,7 +42,7 @@
 
 #include "version.h"
 #include "event.h"
-#include "func_adapter.h"
+#include "core/func_adapter.h"
```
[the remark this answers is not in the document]
drewdzzz: Fixed.

src/box/lua/upgrade.lua
```diff
@@ -1948,6 +1970,39 @@ local function downgrade_from_3_0_0(issue_handler)
...
+--------------------------------------------------------------------------------
+-- Tarantool 3.1.0
+--------------------------------------------------------------------------------
+
+local function drop_trigger_from_func(issue_handler)
+    local _space = box.space[box.schema.SPACE_ID]
+    local _func = box.space[box.schema.FUNC_ID]
+    local _vfunc = box.space[box.schema.VFUNC_ID]
+    local fmt = 'Function %s is registered as event trigger. ' ..
+        'It is supported starting from version 3.1.0'
```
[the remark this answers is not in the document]
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

src/box/alter.cc
drewdzzz: Yes, but "safety" is guaranteed then.
Also, if a trigger doesn't yield, the problem won't appear.

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
drewdzzz: Thanks :)

test/box-luatest/downgrade_test.lua
drewdzzz: Agree, dropped it.

test/box-luatest/downgrade_test.lua
drewdzzz: I'm not sure I understand what you're talking about, but I revised the indentation in this function.

locker (approved):

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,93 @@ g.test_downgrade_global_names = function(cg)
...
+            end
+        end
+
+        local lua_code = 'function(a, b) return a + b end'
+        local event_name = 'persistent_trigger_event'
+        local func_names = {'trigger_func1', 'trigger_func2'}
+        local func_no_trigger_name = 'trigger_test_func_no_trigger'
+        for _, func_name in pairs(func_names) do
+            box.schema.func.create(func_name,
+                {body = lua_code, trigger = event_name})
```
nshy, line 1019: Nit. Better usual indentation.

```lua
            box.schema.func.create(func_name,
                                   {body = lua_code, trigger = event_name})
```

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,93 @@ g.test_downgrade_global_names = function(cg)
...
+        box.schema.downgrade(app_version)
+        check_space_func(true)
+        check_func(func_names, event_name, true)
+        check_func(func_no_trigger_name, nil, true)
+
+        -- Check if downgrade_issues are collected correctly and downgrade
+        -- actually doesn't happen
+        local fmt = 'Function %s is registered as ' ..
+                    'event trigger. It is supported starting from version 3.1.0'
+        local expected_issues = {}
```
nshy, line 1036: Nit. I think test are more readable if they have explicit expectations. Here I would make issues array by hand.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,93 @@ g.test_downgrade_global_names = function(cg)
...
+                local func_tuple_len = 19
+                local trigger_field = nil
+                if has_trigger then
+                    func_tuple_len = 20
+                    trigger_field = {event_name}
+                end
+                t.assert_equals(func_tuple.trigger, trigger_field)
+                t.assert_equals(#func_tuple, func_tuple_len)
+            end
+        end
```
nshy, line 1011: Nit. I think test are more readable if they have explicit expectations.

We can have `before_downgrade` check and `after_downgrade` check as in other cases. Both have single flag argument whether we have trigger functions or not. No loops, branch only on argument.

nshy (approved):

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
```diff
@@ -973,3 +973,93 @@ g.test_downgrade_global_names = function(cg)
...
+        local fmt = 'Function %s is registered as ' ..
+                    'event trigger. It is supported starting from version 3.1.0'
+        local expected_issues = {}
+        for _, func_name in pairs(func_names) do
+            local errmsg = string.format(fmt, func_name)
+            table.insert(expected_issues, errmsg)
+        end
+        local prev_version = helper.prev_version('3.1.0')
+        t.assert_equals(box.schema.downgrade_issues(prev_version),
+            expected_issues)
```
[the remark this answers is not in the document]
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

test/box-luatest/downgrade_test.lua
drewdzzz: Done.

PR #9623, opened 2024-01-29
config: call extras.post_apply() if reload failed

ImeevMA:
This patch changes when extras.post_apply() is executed. Previously, it would only run if a _post_apply() was successful and all previous reload steps completed properly. It will now be executed, even if some reload steps failed.

Note that if an error appears during startup, Tarantool will stop and extras.post_apply() will not be executed.

Also note that if the startup or restart is successful, there will be no change in behavior due to this patch.

Needed for https://github.com/tarantool/tarantool-ee/issues/643

### DIFF
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -374,10 +374,6 @@ function methods._post_apply(self)
             applier.post_apply(self)
         end
     end
-
-    if extras ~= nil then
-        extras.post_apply(self)
-    end
 end
 
 -- Set proper status depending on received alerts.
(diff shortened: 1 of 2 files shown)


ochaplashkin (approved):

Totktonada (approved):
I see no problems. Thanks!

I would like to engage @askalt into the review to think about an effect of this change on his code.

askalt (approved):
Thank you for the patch! lgtm

Totktonada:
Pushed to `master`, upcoming 3.1.0 release. Backported to `release/3.0`, upcoming 3.0.2 release in PR #9653.

PR #9636, opened 2024-01-31
Once again fix transaction boundaries in replication for txs ending with a local row

### LINKED ISSUE #9491: ER_PROTOCOL in relay when replicating transactions with last local row
**Bug description**

Replication is broken on replica reconnect if master writes transactions where last row is local while the replica doesn't follow.

Tarantool version:

**Steps to reproduce**

Run two tarantool instances iteractively in clean directories.
```lua
-- Instance 1.
-- Step 1.
box.cfg{listen=3301}
box.schema.space.create('glob'):create_index('pk')
box.schema.space.create('loc', {is_local=true}):create_index('pk')

-- Step 4.
box.begin() box.space.glob:replace{1} box.space.loc:replace{1} box.commit()
box.space.glob:replace{2}

-- Instance 2.
-- Step 2.
box.cfg{replication=3301}
-- Step 3.
box.cfg{replication=""}

-- Step 5.
box.cfg{replication=3301}
```
After step 5 replication won't start. Logs on both instances will contain the following error:
```
ER_PROTOCOL: Found a new transaction with previous one not yet committed
```
And `box.info.replication` on replica will show
```
---
- 1:
    id: 1
    uuid: 5b636832-905c-4ed3-98cc-5d17448a3ccf
    lsn: 6
    upstream:
      peer: '3301'
      lag: 0.00042891502380371
      status: stopped
      idle: 686.26692
      message: Found a new transaction with previous one not yet committed
  2:
    id: 2
    uuid: fc8d7386-bd1b-43e0-98db-4fd2edadb91d
    lsn: 0
...
```

sergepetrenko:
We've had numerous problems with transaction boundaries in replication.
They were mostly caused by various cases when either the beginning or
end of the transaction happened to be a local row. Local rows are not
replicated, so the peer saw "corrupted" transactions with either no
beginning or no end flag, even though the transaction contents were
fine.

The problem with starting a transaction with a local row was solved in
commit f41d1ddd5faf ("wal: fix tx boundaries"), and that fix seems to
continue working fine to this day.

The problem with ending transactions with a local row was first fixed
in commit 25382617b957 ("replication: append NOP as the last tx row"),
however there were problems with this approach: when a user tried to
write to local spaces on a replica from a replication trigger, it made
it impossible to ever start replicating from replica back to master.

Another fix was proposed: in commit f96782b53214 ("relay: send rows
transactionally") we made relay read a full transaction into memory and
then send it all at once mangling with transanction start and end flags
when necessary.

After that the NOPs were removed in commit f5e52b2c2cf5 ("box: get rid
of dummy NOPs after transactions ending with local rows"), since relay
became capable of fixing transaction boundaries itself.

Turns out the assumption that relay always sees a full transaction and
may correctly set transaction boundaries is wrong: when a replica
reconnects to master we set its starting vclock[0] to the one master has
at the moment of reconnect, so when recovery reads local rows with lsns
less than vclock[0] it silently skips them without showing them to
relay. When such skipped rows contain the is_commit flag for a currently
sent transaction we get the same problem as described before.

Let's make recovery track whether it has pushed any transaction rows to
relay or not, and if yes, recover rows with is_commit flag regardless of
whether the rows were already applied. To prevent recovering the same
data twice, recovery replaces such row contents with NOPs. Basically the
row is "recovered" only for the sake of showing its is_commit flag to
relay. Relay will skip the row anyway, since it remains local.

Follow-up #8958
Closes #9491

### DIFF
--- a/changelogs/unreleased/gh-9491-last-local-row-tx-boundary.md
+++ b/changelogs/unreleased/gh-9491-last-local-row-tx-boundary.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a bug when replication broke with `ER_PROTOCOL` when transactions ended
+  with a local space operation (gh-9491).
--- a/src/box/recovery.cc
+++ b/src/box/recovery.cc
@@ -268,9 +268,6 @@ recover_xlog(struct recovery *r, struct xstream *stream,
 		if (stop_vclock != NULL &&
 		    r->vclock.signature >= stop_vclock->signature)
 			return;
-		int64_t current_lsn = vclock_get(&r->vclock, row.replica_id);
-		if (row.lsn <= current_lsn)
-			continue; /* already applied, skip */
 
 		/*
 		 * All rows in xlog files have an assigned replica
@@ -278,6 +275,9 @@ recover_xlog(struct recovery *r, struct xstream *stream,
 		 * are signed with a zero replica id.
 		 */
 		assert(row.replica_id != 0 || row.group_id == GROUP_LOCAL);
+		int64_t current_lsn = vclock_get(&r->vclock, row.replica_id);
+		if (row.lsn <= current_lsn)
+			continue; /* already applied, skip */
 		/*
 		 * We can promote the vclock either before or
 		 * after xstream_write(): it only makes any impact
--- a/src/box/recovery.cc
+++ b/src/box/recovery.cc
@@ -248,6 +248,7 @@ recover_xlog(struct recovery *r, struct xstream *stream,
 	     const struct vclock *stop_vclock)
 {
 	struct xrow_header row;
+	bool is_sending_tx = false;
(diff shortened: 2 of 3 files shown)


ochaplashkin (approved):

Serpentian (approved):
Thank you for the patch! It fixes the problem, though I can't parse the meaning of `is_sending_tx` flag and how it works.

sergepetrenko:
@Serpentian `is_sending_tx` indicates that recovery has started recovering a transaction, but hasn't yet read its `is_commit`. It means we need to "recover" the row with `is_commit` even if it is local and is already "recovered" judging by recovery vclock[0]

Gerold103 (approved):
Thanks for the patch!

xuniq (approved):

sergepetrenko:
Backports:
3.0: #9668 
2.11: #9667 
2.10: #9669

PR #9648, opened 2024-02-05
config: allow to set credentials in roles

### LINKED ISSUE #9643: config: allow to give priveleges from app roles
Sometimes it's needed to create the required user or role and grant it the privileges to work correctly with an app role.

The plan is to adjust config's privileges in a way that it revokes only these privileges that were granted by the config itself.

ImeevMA:
Closes #9643

@TarantoolBot document
Title: Credentials in roles

Credentials can now be set from roles. To do this, the new `credentials` field could be used. Credentials can be either a Lua table or in yaml format.

Example:
```
local credentials = {
    roles = {
        myrole = {
            privileges = {{
                permissions = {'read'},
                universe = true,
            }}
        },
    },
}

return {
    credentials = credentials,
    validate = function() end,
    apply = function() end,
    stop = function() end,
}
```

Another example:
```
local credentials = [[
    users:
      myuser:
        privileges:
        - permissions:
          - execute
          lua_eval: true
]]

return {
    credentials = credentials,
    validate = function() end,
    apply = function() end,
    stop = function() end,
}
```

Credentials from roles have lower priority than credentials from other sources.

The further a role is in the `roles` field of the config, the higher its priority. I.e. the second role's credentials have more priority than the first role's credentials, but are inferior to the third role's credentials.


ochaplashkin (approved):

Totktonada:
NB: We're in discussion/experimenting how to better implement this functionality.

There are config-level approaches, where a user code (a role) gives some configuration part and we automatically apply it (on some specific stage or pass to all the appliers again).

Here we have static/dynamic privileges approaches: whether `<role>.credentials` is a ready-to-apply table or a function that should have access to the role/full configuration. Early feedback from the @tarantool/ecosystem team suggests that a dynamic credentials configuration is wishful.

Another point to consider is whether the runtime privileges are shown from `config:get()`. It is not clear, whether we should do it.

There are solutions of the credentials/privileges synchronization machinery level, where runtime privilege information is passed to it not as part of the configuration, but as additional data.

xuniq (approved):

ImeevMA:
Decided to use another design.

PR #9654, opened 2024-02-06
key_def: ignore json path in `key_def_find_by_fieldno`

Gumix:
This function returns a `key_def` part by a field number. However, currently it returns `NULL` for parts that contain a JSON path to indexed data. Fix it.

Needed for tarantool/tarantool-ee#671

### DIFF
--- a/src/box/key_def.h
+++ b/src/box/key_def.h
@@ -730,7 +730,9 @@ key_def_decode_parts(struct key_part_def *parts, uint32_t part_count,
 		     uint32_t field_count, struct region *region);
 
 /**
- * Returns the part in index_def->parts for the specified fieldno.
+ * Returns the first part in index_def->parts for the specified fieldno.
+ * The part is returned regardless of whether it is indexed by JSON path or not.
+ *
  * If fieldno is not in index_def->parts returns NULL.
  */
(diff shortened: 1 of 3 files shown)


ochaplashkin (approved):

src/box/key_def.h
```diff
@@ -731,6 +731,8 @@ key_def_decode_parts(struct key_part_def *parts, uint32_t part_count,
 
 /**
  * Returns the part in index_def->parts for the specified fieldno.
```
drewdzzz, line 733: Nit: if we don't reject json path here, we should write something like "returns the first part in `index_def->parts` for the specified fieldno" since we can have several parts indexing one field with different json paths.

drewdzzz (approved):

src/box/key_def.h
Gumix: Done.

Gumix:
This patch fixes a segfault in EE, so I guess it should be backported to 3.0 and 2.11.

sergepetrenko (approved):
Thanks for the patch! LGTM.

sergepetrenko:
Backports:
3.0: #9674
2.11: #9673

PR #9666, opened 2024-02-08
[backport 2.10] Once again fix transaction boundaries in replication for txs ending with a local row

sergepetrenko:


sergepetrenko:
Pardon, I missclicked.

PR #9675, opened 2024-02-10
relay: fix stack use after return

### LINKED ISSUE #9505: relay: stack-use-after-return with txn_isolation = 'linearizable'
## Version

Tarantool 3.0.0-beta1-91-g1262e99dbd Linux-x86_64-Debug

## Build

```sh
$ CC=clang-16 CXX=clang++-16 cmake . -DCMAKE_BUILD_TYPE=Debug -DFIBER_STACK_SIZE=640Kb -DENABLE_WERROR=ON -DENABLE_ASAN=ON -DENABLE_UB_SANITIZER=ON -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION=OFF -DENABLE_FUZZER=ON -DTEST_BUILD=ON
```

## Script

```lua
-- t.lua
#!/usr/bin/env tarantool

local log = require('log')
local fio = require('fio')

local instance_name = arg[1]
assert(instance_name == 'i1' or instance_name == 'i2' or instance_name == 'i3')
fio.mktree(instance_name)
box.cfg({
    wal_dir = instance_name,
    memtx_dir = instance_name,
    listen = ('unix/:./%s.iproto'):format(instance_name),
    replication = {
        ('unix/:./%s.iproto'):format('i1'),
        ('unix/:./%s.iproto'):format('i2'),
        ('unix/:./%s.iproto'):format('i3'),
    },
    memtx_use_mvcc_engine = true,
    election_mode = 'candidate',
})
box.schema.user.grant('guest', 'super', nil, nil, {if_not_exists = true})

local cfg = {
    status_check_interval = 0.1,
}
local monitor_status_fiber
local current_status

local function monitor_status_step()
    -- Try to reach someone from quorum.
    local ok, err = pcall(box.begin, {
        txn_isolation = 'linearizable',
        timeout       = box.cfg.replication_synchro_timeout,
    })
    if ok then
        box.commit()
        current_status = 'connected'
    else
        log.info('status monitor: %s', err)
        current_status = 'disconnected'
    end
end

local function monitor_status_loop()
    fiber.self():name('monitor_status')
    while true do
        local next_step_time = fiber.time() + cfg.status_check_interval
        monitor_status_step()
        local time_to_sleep = next_step_time - fiber.time()
        if time_to_sleep > 0 then
            fiber.sleep(time_to_sleep)
        end
    end
end

local function init_schema()
    if not box.info.ro then
        box.once('init_schema', function()
            box.schema.create_space('s', {
                format = {{
                    type = 'string',
                    name = 'x',
                }, {
                    type = 'string',
                    name = 'y',
                }, {
                    type = 'unsigned',
                    name = 'z',
                }},
                is_sync = true,
            })
            box.space.s:create_index('primary', {
                unique = true,
                parts = {'x'},
            })
            box.schema.create_space('t', {
                format = {{
                    type = 'string',
                    name = 'x',
                }, {
                    type = 'scalar',
                    name = 'y',
                }},
                is_sync = true,
            })
            box.space.t:create_index('primary', {
                unique = true,
                parts = {'x'},
            })
        end)
    end
end

local function start()
    init_schema()

    if monitor_status_fiber == nil or
       monitor_status_fiber:status() == 'dead' then
        monitor_status_fiber = fiber.new(monitor_status_loop)
    end
end

start()
```

```sh
-- run.sh
#!/bin/sh

export ASAN=ON
export LSAN_OPTIONS=suppressions=$(realpath asan/lsan.supp)
export ASAN_OPTIONS=heap_profile=0:unmap_shadow_on_exit=1:detect_invalid_pointer_pairs=1:symbolize=1:detect_leaks=1:dump_instruction_bytes=1
export PATH=$(realpath src):$PATH

#rm -r i1 i2 i3 i1.iproto i2.iproto i3.iproto
./t.lua i1 &
./t.lua i2 &
./t.lua i3 &

wait
```

## Run

```sh
$ ./run.sh
Ctrl+C
$ ./run.sh
Ctrl+C
$ ./run.sh
Ctrl+C
$ ./run.sh
Ctrl+C
<...>
```

The problem is reproduced from time to time. I have no stable reproducer.

## Result

```c
==16681==ERROR: AddressSanitizer: stack-use-after-return on address 0x7f80b2817c38 at pc 0x55fafe3c4a2e bp 0x7f80cd0120d0 sp 0x7f80cd0120c8
WRITE of size 8 at 0x7f80b2817c38 thread T0
2023-12-19 03:30:42.564 [16680] relay/unix/:(socket)/101/main I> recover from `i2/00000000000000000026.xlog'
    #0 0x55fafe3c4a2d in relay_trigger_vclock_sync /home/alex/p/tarantool-meta/tarantool/src/box/relay.cc:918:15
    #1 0x55fafe252bf3 in relay_get_sync_on_start(trigger*, void*) /home/alex/p/tarantool-meta/tarantool/src/box/box.cc:2564:6
    #2 0x55fafece5bc6 in trigger_run_list(rlist*, void*) /home/alex/p/tarantool-meta/tarantool/src/lib/core/trigger.cc:100:7
    #3 0x55fafece4146 in trigger_run /home/alex/p/tarantool-meta/tarantool/src/lib/core/trigger.cc:133:9
    #4 0x55fafe3d26cd in relay_thread_on_start(void*) /home/alex/p/tarantool-meta/tarantool/src/box/relay.cc:864:2
    #5 0x55fafec0b57b in cbus_pair_perform /home/alex/p/tarantool-meta/tarantool/src/lib/core/cbus.c:487:3
    #6 0x55fafec07ac5 in cmsg_deliver /home/alex/p/tarantool-meta/tarantool/src/lib/core/cbus.c:373:2
    #7 0x55fafec16ac4 in fiber_pool_f /home/alex/p/tarantool-meta/tarantool/src/lib/core/fiber_pool.c:64:3
    #8 0x55fafd8dbfa8 in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*) /home/alex/p/tarantool-meta/tarantool/src/lib/core/fiber.h:1297:10
    #9 0x55fafebc7fbb in fiber_loop /home/alex/p/tarantool-meta/tarantool/src/lib/core/fiber.c:1160:18
    #10 0x55faffbde955 in coro_init /home/alex/p/tarantool-meta/tarantool/third_party/coro/coro.c:108:3

Address 0x7f80b2817c38 is a wild pointer inside of access range of size 0x000000000008.
SUMMARY: AddressSanitizer: stack-use-after-return /home/alex/p/tarantool-meta/tarantool/src/box/relay.cc:918:15 in relay_trigger_vclock_sync
Shadow bytes around the buggy address:
  0x7f80b2817980: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817a00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817a80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817b00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817b80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
=>0x7f80b2817c00: f5 f5 f5 f5 f5 f5 f5[f5]f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817c80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817d00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817d80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817e00: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
  0x7f80b2817e80: f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5 f5
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==16681==ABORTING
```

Astronomax:
Fixed a bug when it was possible to use the
box_collect_confirmed_vclock stack after return.

Closes #9505

NO_DOC=bugfix
NO_TEST=asan

### DIFF
--- a/changelogs/unreleased/gh-9505-relay-stack-use-after-return.md
+++ b/changelogs/unreleased/gh-9505-relay-stack-use-after-return.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed an issue when it was possible to use the
+  `box_collect_confirmed_vclock` stack after return (gh-9505).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -2540,6 +2540,15 @@ box_wait_quorum(uint32_t lead_id, int64_t target_lsn, int quorum,
 	return 0;
 }
 
+/**
+ * The pool is used by box_cc to allocate sync_trigger_data that is used in
+ * box_collect_confirmed_vclock and relay_get_sync_on_start. We allocate
+ * sync_trigger_data dynamically because these functions are running in
+ * different fibers. The lifetime of sync_trigger_data is not limited by the
+ * execution time of box_collect_confirmed_vclock.
+ */
+static struct mempool sync_trigger_data_pool;
+
 /** A structure holding trigger data to collect syncs. */
 struct sync_trigger_data {
 	/** Syncs to wait for. */
@@ -2558,7 +2567,29 @@ struct sync_trigger_data {
 	int count;
 	/** Whether the request is timed out. */
 	bool is_timed_out;
+	/** Count of fibers that are using data. */
+	int ref_count;
 };
+
+/** Let others know we need data. */
+void
+sync_trigger_data_ref(struct sync_trigger_data *data)
+{
+	++data->ref_count;
+}
+
+/**
+ * Let others know that we no longer need the data.
+ * If no one else needs the data, free it.
+ */
+void
+sync_trigger_data_unref(struct sync_trigger_data *data)
+{
+	--data->ref_count;
+	assert(data->ref_count >= 0);
+	if (data->ref_count == 0)
+		mempool_free(&sync_trigger_data_pool, data);
+}
 
 /**
  * A trigger executed on each ack to collect up to date remote node vclocks.
@@ -2605,12 +2636,15 @@ relay_get_sync_on_start(struct trigger *trigger, void *event)
 	/* Already accounted. */
 	if (bit_test(&data->collected_vclock_map, id))
 		return 0;
+
+	sync_trigger_data_ref(data);
 	if (relay_trigger_vclock_sync(relay, &data->vclock_syncs[id],
 				      data->deadline) != 0) {
 		diag_clear(diag_get());
 		data->is_timed_out = true;
 		fiber_wakeup(data->waiter);
 	}
+	sync_trigger_data_unref(data);
 	return 0;
 }
 
@@ -2632,54 +2666,60 @@ box_collect_confirmed_vclock(struct vclock *confirmed_vclock, double deadline)
 	if (vclock_count <= 1)
 		return 0;
 
-	struct sync_trigger_data data = {
-		.vclock_syncs = {0},
-		.collected_vclock_map = 0,
-		.waiter = fiber(),
-		.vclock = confirmed_vclock,
-		.deadline = deadline,
-		.count = vclock_count,
-		.is_timed_out = false,
-	};
-	bit_set(&data.collected_vclock_map, instance_id);
+	struct sync_trigger_data *data = (sync_trigger_data *)
+		xmempool_alloc(&sync_trigger_data_pool);
+	memset(data->vclock_syncs, 0, sizeof(data->vclock_syncs));
+	data->collected_vclock_map = 0;
+	data->waiter = fiber();
+	data->vclock = confirmed_vclock;
+	data->deadline = deadline;
+	data->count = vclock_count;
+	data->is_timed_out = false;
+	data->ref_count = 0;
+
+	sync_trigger_data_ref(data);
+	bit_set(&data->collected_vclock_map, instance_id);
 	struct trigger on_relay_thread_start;
-	trigger_create(&on_relay_thread_start, relay_get_sync_on_start, &data,
+	trigger_create(&on_relay_thread_start, relay_get_sync_on_start, data,
 		       NULL);
 	trigger_add(&replicaset.on_relay_thread_start, &on_relay_thread_start);
 	struct trigger on_ack;
-	trigger_create(&on_ack, check_vclock_sync_on_ack, &data, NULL);
+	trigger_create(&on_ack, check_vclock_sync_on_ack, data, NULL);
 	trigger_add(&replicaset.on_ack, &on_ack);
+
+	auto guard = make_scoped_guard([&] {
+		trigger_clear(&on_ack);
+		trigger_clear(&on_relay_thread_start);
+		sync_trigger_data_unref(data);
+	});
+
 	replicaset_foreach(replica) {
 		if (relay_get_state(replica->relay) != RELAY_FOLLOW ||
 		    replica->anon) {
 			continue;
 		}
 		/* Might be already filled by on_relay_thread_start trigger. */
-		if (data.vclock_syncs[replica->id] != 0)
+		if (data->vclock_syncs[replica->id] != 0)
 			continue;
 		if (relay_trigger_vclock_sync(replica->relay,
-					      &data.vclock_syncs[replica->id],
+					      &data->vclock_syncs[replica->id],
 					      deadline) != 0) {
 			/* Timed out. */
-			trigger_clear(&on_ack);
-			trigger_clear(&on_relay_thread_start);
 			return -1;
 		}
 	}
 
-	while (bit_count_u32(data.collected_vclock_map) < vclock_count &&
-	       !data.is_timed_out && !fiber_is_cancelled()) {
+	while (bit_count_u32(data->collected_vclock_map) < vclock_count &&
+	       !data->is_timed_out && !fiber_is_cancelled()) {
 		if (fiber_yield_deadline(deadline))
 			break;
 	}
 
-	trigger_clear(&on_ack);
-	trigger_clear(&on_relay_thread_start);
 	if (fiber_is_cancelled()) {
 		diag_set(FiberIsCancelled);
 		return -1;
 	}
-	if (bit_count_u32(data.collected_vclock_map) < vclock_count) {
+	if (bit_count_u32(data->collected_vclock_map) < vclock_count) {
 		diag_set(TimedOut);
 		return -1;
 	}
@@ -5984,6 +6024,8 @@ box_init(void)
 	 */
 	builtin_events_init();
 	crash_callback = box_crash_callback;
+	mempool_create(&sync_trigger_data_pool, &cord()->slabc,
+		       sizeof(struct sync_trigger_data));
 }
 
 /** Shutdown box storage i.e. stop parts that need TX loop running. */
@@ -6036,6 +6078,7 @@ box_free(void)
 	txn_event_trigger_free();
 	tuple_free();
 	iproto_constants_free();
+	mempool_destroy(&sync_trigger_data_pool);
 	/* schema_module_free(); */
 	/* session_free(); */
 	/* user_cache_free(); */


changelogs/unreleased/gh-9505-relay-stack-use-after-return.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Fixed an issue when it was possible to use the
+  box_collect_confirmed_vclock stack after return (gh-9505).
```
xuniq, line 4:
```suggestion
  `box_collect_confirmed_vclock` stack after return (gh-9505).
```

xuniq (approved):

src/box/box.cc
```diff
@@ -2514,6 +2514,8 @@ struct sync_trigger_data {
 	int count;
 	/** Whether the request is timed out. */
```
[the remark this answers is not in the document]
Serpentian: Let's please also extend the commit message with description, how this bug can be reproduced

src/box/box.cc
```diff
@@ -2617,8 +2636,8 @@ box_collect_confirmed_vclock(struct vclock *confirmed_vclock, double deadline)
 					      &data.vclock_syncs[replica->id],
 					      deadline) != 0) {
 			/* Timed out. */
-			trigger_clear(&on_ack);
-			trigger_clear(&on_relay_thread_start);
+			while (data.data_access_count > 0)
```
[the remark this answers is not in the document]
Serpentian: It looks like the better solution will be introducing new variable inside relay, which will be reset to false in scope_guard of `box_collect_confirmed_vclock`. We'll check this variable in `relay_trigger_vclock_sync` before accessing `vclock_sync` pointer. @sergepetrenko what do you think?

Serpentian (changes requested):
Thank you for the patch! Let's consult with @sergepetrenko about possible solutions to the problem.

src/box/box.cc
```diff
@@ -2514,6 +2514,8 @@ struct sync_trigger_data {
 	int count;
```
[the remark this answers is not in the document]
Astronomax: I reproduced this bug on Ubuntu. About once for 5-10 attempts. After this patch, it is no longer reprodu cable.

src/box/box.cc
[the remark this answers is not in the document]
sergepetrenko: I don't see how we could set this variable.
It's not clear which exact relay is running the `on_relay_thread_start` trigger. And if you try to set the variable in all the relays, it'll lead to a case when other concurrent calls to `box_collect_confirmed_vclock` fail, because the relays beleive they shouldn't set `vclock_sync`.

I think it's better to start allocating the trigger data dynamically and probably reference counting it.

So that you don't have to wait `while (data.data_access_count > 0)` but instead free the data only when `data_access_count` becomes 0. No matter where this happens, right in `box_collect_confirmed_vclock` or `relay_trigger_vclock_sync`.

So basically `relay_trigger_vclock_sync` will always succed setting the `vclock_sync` field, because it will reference the trigger data and it won't be deleted. But as soon as it sets the `vclock_sync`, it will decrement the reference count and might actually delete the data right away.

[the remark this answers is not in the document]
Astronomax: If we have entered this loop, it means that there are currently running `relay_get_sync_on_start`. Here we are waiting for them to finish and wake up our fiber. However, `relay_get_sync_on_start` accepts deadline as one of the arguments. Therefore, everyone we are waiting for here ends when the deadline arrives. So I don't see there any problems.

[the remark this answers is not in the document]
Serpentian: > And if you try to set the variable in all the relays, it'll lead to a case when other concurrent calls to box_collect_confirmed_vclock fail, because the relays beleive they shouldn't set vclock_sync.

Yeah, sounds reasonable, not great.

> But as soon as it sets the vclock_sync, it will decrement the reference count and might actually delete the data right away.

I like the idea of allocating `sync_trigger_data` dynamically. AFAICS, freeing cannot be done in `box_collect_confirmed_vclock`, as it may exit earlier, then trigger is executed. Should be implemented inside `relay_trigger_vclock_sync`, when counter is 0, we free the memory.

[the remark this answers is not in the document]
sergepetrenko: I suggest switching to refcounts and dynamic allocation.

Waiting until some other code stops using our stack-allocated data doesn't look too good.

src/box/box.cc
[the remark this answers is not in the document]
Serpentian: > Is it possible to write the test for the bug? Asan doesn't seem to be a reason for the test to be skipped)

Eventually, it's. `luatest` doesn't check returned value of the server and succeedes, even if asan error happens: https://github.com/tarantool/luatest/issues/349.

It seems, we can skip testing for this issue

src/box/box.cc
```diff
@@ -2540,6 +2540,12 @@ box_wait_quorum(uint32_t lead_id, int64_t target_lsn, int quorum,
 	return 0;
 }
 
+/**
+ * The pool is used by box_cc to allocate sync_trigger_data
+ * that is used in box_collect_confirmed_vclock.
```
Serpentian, line 2545: Please, describe why we need dynamic allocation. We use it in `box_collect_confirm_vclock` and in `on_relay_thread_start` trigger. They're executed in different fibers, so we cannot free data in `box_collect_confirm_vclock`

src/box/box.cc
```diff
@@ -2540,6 +2540,12 @@ box_wait_quorum(uint32_t lead_id, int64_t target_lsn, int quorum,
 	return 0;
 }
 
+/**
```
Serpentian, line 2543: Please, extend the commit message, describe the bug and the solution. It's not convenient to go to the ticket, when you use `git log`

src/box/box.cc
```diff
@@ -2605,6 +2613,17 @@ relay_get_sync_on_start(struct trigger *trigger, void *event)
 	/* Already accounted. */
 	if (bit_test(&data->collected_vclock_map, id))
 		return 0;
+
+	/*
+	 * Let box_collect_confirmed_vclock know that the data is needed.
+	 * Otherwise, it may free the data on return while this fiber
+	 * yielding inside relay_trigger_vclock_sync.
+	 */
+	++data->ref_count;
```
Serpentian, line 2622: Now you're incrementing/decrementing counter in too places: `box_collect_confirmed_vclock` and `relay_get_sync_on_start`. I'd rather put everything inside `relay_trigger_vclock_sync`. This way it'll be all in one place, you increment before yielding, and decrement, free memory after yielding.

Serpentian (changes requested):
Looks great. Just minor comments

src/box/box.cc
Astronomax: Done.

src/box/box.cc
Astronomax: Done.

src/box/box.cc
Astronomax: Done.

src/box/box.cc
Serpentian: I suppose we misunderstood each other. It's still in `relay_get_sync_on_start` and `box_collect_confirmed_vclock`, not in `relay_trigger_vclock_sync`. Moreover, now there's initialization with 1 in `data->ref_count`, which is not easy to parse.

Serpentian: My bad, we cannot do that, as `data` is used after exiting from `rellay_trigger_vclock_sync` in `box_collect_confirm_vclock`

src/box/box.cc
```diff
@@ -2632,54 +2649,70 @@ box_collect_confirmed_vclock(struct vclock *confirmed_vclock, double deadline)
...
+	trigger_create(&on_ack, check_vclock_sync_on_ack, data, NULL);
 	trigger_add(&replicaset.on_ack, &on_ack);
+
+	auto guard = make_scoped_guard([&] {
+		trigger_clear(&on_ack);
+		trigger_clear(&on_relay_thread_start);
+		/*
+		 * box_collect_confirmed_vclock no longer uses data.
+		 * If no one else is using the data, free it.
+		 */
```
Serpentian, line 2684: Nit: I'd not use prefix decrement here, so we can add asertion, that `ref_count` after decrement is >= 0. Just to be sure)

```c
--data->ref_count;
assert(data->ref_count >= 0);
if (data->ref_count == 0)
    ...
```

Same in other place

Serpentian (approved):
Thank you for working on this! Last nit

changelogs/unreleased/gh-9505-relay-stack-use-after-return.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
```
sergepetrenko, line 1: Why the `/config` subsystem?

It should be `bugfix/replication` or probably `bugfix/core`

src/box/box.cc
```diff
@@ -2558,6 +2567,8 @@ struct sync_trigger_data {
 	int count;
 	/** Whether the request is timed out. */
 	bool is_timed_out;
+	/** Count of fibers that are using data. **/
```
sergepetrenko, line 2570: nit: the closing  `*/` should have one `*`.

src/box/box.cc
```diff
@@ -2605,6 +2616,12 @@ relay_get_sync_on_start(struct trigger *trigger, void *event)
 	/* Already accounted. */
 	if (bit_test(&data->collected_vclock_map, id))
 		return 0;
+
+	++data->ref_count;
```
sergepetrenko, line 2620: Let's introduce helpers `sync_trigger_data_ref` and `sync_trigger_data_unref`
which would increment and decrement the `ref_count`, and free the data, if needed

src/box/box.cc
```diff
@@ -2632,54 +2649,70 @@ box_collect_confirmed_vclock(struct vclock *confirmed_vclock, double deadline)
...
+	memset(data->vclock_syncs, 0, sizeof(data->vclock_syncs));
+	data->collected_vclock_map = 0;
+	data->waiter = fiber();
+	data->vclock = confirmed_vclock;
+	data->deadline = deadline;
+	data->count = vclock_count;
+	data->is_timed_out = false;
+	/*
+	 * Let relay_trigger_vclock_sync know that the data is needed in
+	 * box_collect_confirmed_vclock. Otherwise, relay_trigger_vclock_sync
```
sergepetrenko, line 2663: Nit: `relay_get_sync_on_start` frees the data, not `relay_trigger_vclock_sync`.

src/box/box.cc
```diff
@@ -2605,6 +2616,12 @@ relay_get_sync_on_start(struct trigger *trigger, void *event)
 	/* Already accounted. */
 	if (bit_test(&data->collected_vclock_map, id))
 		return 0;
+
+	++data->ref_count;
+	auto guard = make_scoped_guard([&] {
+		if (--data->ref_count == 0)
+			mempool_free(&sync_trigger_data_pool, data);
```
sergepetrenko, line 2623: There's a single exit from `relay_get_sync_on_start`, by `return 0` below. So it's easier to call `sync_trigger_data_unref()` there one time than introduce a guard.

sergepetrenko (changes requested):
Thanks for the patch!

Looks good mostly, but needs some polishing. Please find my comments below.

src/box/box.cc
```diff
@@ -2632,54 +2666,60 @@ box_collect_confirmed_vclock(struct vclock *confirmed_vclock, double deadline)
...
-		.collected_vclock_map = 0,
-		.waiter = fiber(),
-		.vclock = confirmed_vclock,
-		.deadline = deadline,
-		.count = vclock_count,
-		.is_timed_out = false,
-	};
-	bit_set(&data.collected_vclock_map, instance_id);
+	struct sync_trigger_data *data = (sync_trigger_data *)
+		mempool_alloc(&sync_trigger_data_pool);
```
sergepetrenko, line 2670: Let's use `xmempool_alloc` here.

src/box/box.cc
```diff
@@ -2558,8 +2567,30 @@ struct sync_trigger_data {
...
 
+/** Let others know we need data. */
+void
+sync_trigger_data_ref(struct sync_trigger_data *data)
+{
+	++data->ref_count;
+}
+
+/*
+ * Let others know that we no longer need the data.
```
sergepetrenko, line 2582: Nit: comments describing structure members and function declarations should start with `/**`.

sergepetrenko:
Backports:
2.11 https://github.com/tarantool/tarantool/pull/9821
3.0 https://github.com/tarantool/tarantool/pull/9822

PR #9681, opened 2024-02-13
box/lua: allow to set custom error payload fields

### LINKED ISSUE #9104: Allow to set error payload
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#578816e27a1f480497d5a750d89412c7)

Setting a payload field will only be possible on error construction. The payload fields will be passed as key-value pairs to the error constructor. For example:

```lua
box.error{
   type = "MyAppError",
   message = "Something broke!",
   foo = 42,        -- payload
   bar = "abc",     -- payload
   baz = {1, 2, 3}, -- payload
}
```

One may pass any object serializable to MsgPack as a payload field value (string, number, array, map, uuid, etc). A payload field name can be arbitrary except it should not be the same as any of the built-in error fields: `type`, `base_type`, `custom_type`, `code`, `errno`, `message`, `prev`, `trace`. Some of the built-in fields can be set in exactly the same way as payload fields and will be taken into account: `type`, `code`, `message`, `prev`. Other payload fields, for example, `errno`, will be silently masked by the built-in fields set by the error object constructor.

A payload field may be accessed with the `.` operator:

```lua
local e = box.error.new{foo = 42, bar = "abc"}
e.foo -- 42
e.bar -- "abc"
```

Gumix:
Now it is possible to add a custom payload to the error on error construction. The payload is passed as key-value pairs, where `key` is a string, and `value` is any Lua object. The key name can be arbitrary except it should not be the same as any of the built-in error field name: reason, code, type, base_type, custom_type, errno, message, prev, trace.

E.g.:
```
box.error.new({ reason = string[, code = number, type = string, key1 = value1, ...] }])
```
Or:
```
box.error({ reason = string[, code = number, type = string, key1 = value1, ...] }])
```

Closes #9104

### DIFF
--- a/test/box-luatest/error_subsystem_improvements_test.lua
+++ b/test/box-luatest/error_subsystem_improvements_test.lua
@@ -1,0 +1,81 @@
+local t = require('luatest')
+local g = t.group()
+
+-- Test custom error payload (gh-9104).
+g.test_custom_payload = function()
+    local uuid = require('uuid')
+    local decimal = require('decimal')
+    local datetime = require('datetime')
+    local varbinary = require('varbinary')
+
+    local e = box.error.new({
+        reason = "Something happened",
+        -- Non-string key is ignored.
+        [12] = 34,
+        -- Custom payload fields.
+        foo = 42,
+        bar = "abc",
+    })
+    -- Check that payload fields can be accessed with the dot operator.
+    t.assert_equals(e.foo, 42)
+    t.assert_equals(e.bar, "abc")
+    -- Ignore the `trace' field, which contains trimmed filename.
+    local unpacked = e:unpack()
+    unpacked.trace = nil
+    t.assert_equals(unpacked, {
+        base_type = "ClientError",
+        type = "ClientError",
+        message = "Something happened",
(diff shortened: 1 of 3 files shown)


nshy (approved):

ochaplashkin (approved):

xuniq (approved):

sergepetrenko (approved):
Thanks! LGTM.

test/box-luatest/error_subsystem_improvements_test.lua
```diff
@@ -0,0 +1,81 @@
...
+        code = 0,
+        foo = 42,
+        bar = "abc",
+    })
+
+    -- Test `box.error()' with payload fields of various types.
+    local my_uuid = uuid.new()
+    local _, e = pcall(box.error, {
+        type = "MyAppError",
+        reason = "Keyboard not found, press F1 to continue",
```
sergepetrenko, line 38: 🤣

PR #9687, opened 2024-02-14
lua: add RSA-PSS verify OSSL wrapper

Lord-KA:
This patch adds Lua wrapper for OpenSSL RSA-PSS signature verification. It is required by integrity module to validate hashes.json file.

Part of tarantool/tarantool-ee#679

NO_DOC=yet
NO_CHANGELOG=yet
NO_TEST=yet

### DIFF
--- a/extra/exports
+++ b/extra/exports
@@ -193,6 +193,7 @@ crypto_stream_delete
 crypto_stream_new
 crypto_X509_get_default_cert_dir_env
 crypto_X509_get_default_cert_file_env
+crypto_RSA_PSS_verify
 csv_create
 csv_destroy
 csv_escape_field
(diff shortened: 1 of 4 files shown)


Lord-KA:
For now it fails because on static builds we use OpenSSL 1.* and the patch utilizes API from OpenSSL 3.*.
Blocked by ossl bump: https://github.com/tarantool/tarantool-ee/issues/176.

Lord-KA:
Well, there is an issue with the dynamic builds too, as they are build on runners with obsolete OpenSSL 1.1.1. Maybe it's time to drop all support for ossl 1 and move the runners to 3.*. Note that this should be accommodated by removing all duplicating with ossl 1 API.

PR #9695, opened 2024-02-15
lua: add `message` argument to table constructor of `box.error.new`

### LINKED ISSUE #9102: Allow to set error message via message argument
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#c984e766372743c99a4a7be79a5783f6)

Currently, the error message is specified in the `reason` argument of `box.error`. This is confusing, because the message is accessed and printed by [`unpack`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/error_object/#lua-function.error_object.unpack) as `message`. Let’s allow users to pass the error message in the `message` argument. We won’t drop the `reason` argument so as not to break potential users. For example, the code below

```lua
box.error{type = 'MyError', message = 'Error!'}
```

will be equivalent to

```lua
box.error{type = 'MyError', reason = 'Error!'}
```

We’ll also allow to omit the `message` key by treating the first map entry as an error message if set:

```lua
box.error{'Error!', type = 'MyError'}
```

CuriousGeorgiy:
This patch add a `message` option to the table constructor of `box.error.new`.

Closes #9102

### DIFF
--- a/src/box/lua/error.cc
+++ b/src/box/lua/error.cc
@@ -79,6 +79,31 @@ luaT_error_add_payload(lua_State *L, struct error *error)
 }
 
 /**
+ * In case the error is constructed from a table, retrieves the reason.
+ *
+ * @param L Lua state
+ * @param index table index on Lua stack
+ *
+ * @return reason
+ * @retval "" failed to retrieve reason
+ */
+static const char *
+error_create_table_case_get_reason(lua_State *L, int index)
+{
+	lua_rawgeti(L, index, 1);
+	const char *reason = lua_tostring(L, -1);
+	if (reason != NULL)
+		return reason;
+	lua_getfield(L, index, "message");
+	reason = lua_tostring(L, -1);
+	if (reason != NULL)
+		return reason;
+	lua_getfield(L, index, "reason");
+	reason = lua_tostring(L, -1);
+	return reason != NULL ? reason : "";
+}
+
+/**
  * Parse Lua arguments (they can come as single table or as
  * separate members) and construct struct error with given values.
  *
@@ -146,10 +171,7 @@ luaT_error_create(lua_State *L, int top_base)
 		lua_getfield(L, top_base, "code");
 		if (!lua_isnil(L, -1))
 			code = lua_tonumber(L, -1);
-		lua_getfield(L, top_base, "reason");
-		reason = lua_tostring(L, -1);
-		if (reason == NULL)
-			reason = "";
+		reason = error_create_table_case_get_reason(L, top_base);
 		lua_getfield(L, top_base, "type");
 		if (!lua_isnil(L, -1))
 			custom_type = lua_tostring(L, -1);
--- a/test/box-luatest/error_subsystem_improvements_test.lua
+++ b/test/box-luatest/error_subsystem_improvements_test.lua
@@ -79,3 +79,15 @@ g.test_custom_payload = function()
         custom_type = "MyAppError",
     })
 end
+
+-- Test the `message` argument to the table constructor of `box.error.new`
+-- (gh-9102).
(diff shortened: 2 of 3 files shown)


andreyaksenov (approved):

Gumix (approved):
> The error message setting has the following priority:
> first table constructor entry < `message` < `reason`.

I guess you meant:
> The error message setting has the following priority:
> first table constructor entry > `message` > `reason`.

src/box/lua/error.cc
```diff
@@ -146,8 +170,7 @@ luaT_error_create(lua_State *L, int top_base)
 		lua_getfield(L, top_base, "code");
 		if (!lua_isnil(L, -1))
 			code = lua_tonumber(L, -1);
-		lua_getfield(L, top_base, "reason");
-		reason = lua_tostring(L, -1);
+		reason = error_create_table_case_get_reason(L, top_base);
 		if (reason == NULL)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

CuriousGeorgiy:
> > The error message setting has the following priority:
> > first table constructor entry < `message` < `reason`.
> 
> I guess you meant:
> 
> > The error message setting has the following priority:
> > first table constructor entry > `message` > `reason`.

Fixed.

src/box/lua/error.cc
```diff
@@ -78,6 +78,31 @@ luaT_error_add_payload(lua_State *L, struct error *error)
...
 }
 
+/**
+ * In case the error is constructed from a table, retrieves the reason.
+ *
+ * @param L Lua state
+ * @param index table index on Lua stack
+ *
+ * @return reason
+ * @retval NULL failed to retrieve reason
```
nshy, line 88: NULL is never returned.

test/box-luatest/error_subsystem_improvements_test.lua
```diff
@@ -79,3 +79,14 @@ g.test_custom_payload = function()
...
+
+-- Test the `message` argument to the table constructor of `box.error.new`
+-- (gh-9102).
+g.test_error_new_message_arg = function()
+    local msg = 'message'
+    t.assert_equals(box.error.new{msg}.message, msg)
+    t.assert_equals(box.error.new{message = msg}.message, msg)
+    t.assert_equals(box.error.new{reason = msg}.message, msg)
+    t.assert_equals(box.error.new{msg, message = 'other'}.message, msg)
+    t.assert_equals(box.error.new{message = msg, reason = 'other'}.message, msg)
```
nshy, line 91: Looks like it is worth to test [1] precedence othe `reason` field.

Also it does not look safe as stated in issue. It would be safe if `reason` has the highest precedence. But may be I'm just too paranoid.

nshy (changes requested):

test/box-luatest/error_subsystem_improvements_test.lua
CuriousGeorgiy: > Looks like it is worth to test [1] precedence othe `reason` field.

Fixed.

> Also it does not look safe as stated in issue. It would be safe if reason has the highest precedence. But may be I'm just too paranoid.

I think that if the user passed a `message` field or an unnamed field prior to 3.1 (which would previously be simply ignored), it is him to blame.

src/box/lua/error.cc
CuriousGeorgiy: Fixed.

nshy (approved):

ochaplashkin (approved):

ylobankov (approved):

PR #9704, opened 2024-02-16
vinyl: check if fiber is cancelled after tasks completion

nshy:
`sql-tap/intpkey.test` start to flak under load after the commit fe769b0a0325 ("vinyl: add graceful shutdown"). The issue is vy_scheduler_complete_tasks() may yield. So on shutdown `vinyl.scheduler` fiber may be cancelled during this yield and then we go to sleep waiting for new tasks forever.  Now vinyl shutdown hangs.

Part of #8423

### DIFF
--- a/src/box/vy_scheduler.c
+++ b/src/box/vy_scheduler.c
@@ -2066,13 +2066,19 @@ vy_scheduler_f(va_list va)
 {
 	struct vy_scheduler *scheduler = va_arg(va, struct vy_scheduler *);
 
-	while (!fiber_is_cancelled()) {
+	while (true) {
 		struct vy_task *task;
 		int tasks_done, tasks_failed;
 
 		fiber_check_gc();
 		vy_scheduler_complete_tasks(scheduler, &tasks_done,
 					    &tasks_failed);
+		/*
+		 * Tasks completion may yield thus check fiber cancel
+		 * status before any wait.
+		 */
+		if (fiber_is_cancelled())
+			break;
 		/*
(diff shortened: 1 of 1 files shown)


src/box/vy_scheduler.c
```diff
@@ -2073,6 +2073,12 @@ vy_scheduler_f(va_list va)
 		fiber_check_gc();
 		vy_scheduler_complete_tasks(scheduler, &tasks_done,
 					    &tasks_failed);
+		/*
+		 * Tasks completion may yield thus check fiber cancel
+		 * status before any wait.
+		 */
+		if (fiber_is_cancelled())
+			break;
```
locker, line 2081: Let's remove the check in the `while` condition then. It doesn't look like we need it anymore.

locker (changes requested):

src/box/vy_scheduler.c
nshy: Done.

```diff
diff --git a/src/box/vy_scheduler.c b/src/box/vy_scheduler.c
index 29374c656..c28bd1324 100644
--- a/src/box/vy_scheduler.c
+++ b/src/box/vy_scheduler.c
@@ -2066,7 +2066,7 @@ vy_scheduler_f(va_list va)
 {
        struct vy_scheduler *scheduler = va_arg(va, struct vy_scheduler *);

-       while (!fiber_is_cancelled()) {
+       while (true) {
                struct vy_task *task;
                int tasks_done, tasks_failed;
```

PR #9707, opened 2024-02-16
ci: fix integration concurrency group

fckxorg:
In commit 3daf2399dd5b ("ci: fix step parameters for reusable runs") the integration workflow was made reusable, but concurrency group pattern modification that was done for other workflows made reusable in the same patch was forgotten. This patch fixes the mentioned issue.

NO_DOC=CI
NO_TEST=CI
NO_CHANGELOG=CI


igormunkin (approved):
@fckxorg, thanks for the fix! LGTM.

ylobankov (approved):

PR #9714, opened 2024-02-19
box/lua: implement inheritance of error payload fields

### LINKED ISSUE #9106: Allow to access cause payload from effect
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#c080fe2ac28b46c8b0eda7234a8852ce)

Suppose an error has a cause with some payload fields, for example:

```lua
local e1 = box.error.new{'Error A', foo = 'bar'} -- cause
local e2 = box.error.new{'Error B', prev = e1}   -- effect
```

Then to access the cause payload fields given the effect one has to access the cause first:

```lua
e2.prev.foo -- 'bar'
```

This may not always be convenient so the suggestion is to allow to access cause payload fields directly:

```lua
e2.foo -- 'bar'
```

While looking for a payload field with a given name, we’ll always stop at the topmost (closest to the effect) field. If there’s a field with the same name deeper in the stack it’ll be masked:

```lua
local e1 = box.error.new{'Error A', a = 1, b = 2}
local e2 = box.error.new{'Error B', b = 3, c = 4, prev = e1}
local e3 = box.error.new{'Error C', c = 5, prev = e2}
e3.a -- 1 (inherited from e1)
e3.b -- 3 (inherited from e2)
e3.c -- 5 (own payload field)
```

Gumix:
Suppose an error has a cause with some payload fields, for example:
```lua
  local e1 = box.error.new{'e1', foo = 'bar'} -- cause
  local e2 = box.error.new{'e2', prev = e1}   -- effect
```

Now it is possible to access cause payload fields via `e2` directly:
```lua
  e2.foo -- 'bar'
```

While looking for a payload field with a given name, we always stop at the topmost (closest to the effect) field. If there's a field with the same name deeper in the stack it is masked.

Closes #9106


changelogs/unreleased/gh-9106-error-payload-inheritance.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Now it is possible to access a payload field of error's cause directly
```
andreyaksenov, line 3:
```suggestion
* Now it is possible to access a payload field of an error's cause directly
```

andreyaksenov (approved):

ochaplashkin (approved):

src/lua/error.lua
```diff
@@ -178,11 +178,19 @@ local function error_index(err, key)
     if getter ~= nil then
         return getter(err)
     end
-    local f = ffi.C.error_find_field(err, key)
-    if f ~= nil then
-        return mp_decode(f._data)
+    local method = error_methods[key]
```
CuriousGeorgiy, line 181: Why did you move this to the top? This doesn’t look like a backwards-compatible change.

CuriousGeorgiy (changes requested):

src/lua/error.lua
Gumix: I don't know… Moved it back.

CuriousGeorgiy: While we are here, could you please add a test to maintain this masking priority?

Gumix: Done.

CuriousGeorgiy (approved):

src/lua/error.lua
nshy: Given we now allow custom payload fields in eb2c6a4a4de8a958f07e5d4efc4453ded1c3026d it makes sense NOT to mask fields like `unpack` etc. @locker is that correct?

The change can done in a different commit I guess.

locker: I think that error object methods shouldn't be masked by payload fields (i.e. they should be checked first in `__index`) because the user can still access masked payload fields through `unpack` while accessing masked methods would be problematic.

Gumix: > error object methods shouldn't be masked by payload fields

Done.

CuriousGeorgiy: Can’t we break a client that used the C error API to set payloads and relied on the masking behavior?

PR #9724, opened 2024-02-22
Iproto drop connection fixes for 3.0

nshy:


andreyaksenov (approved):

locker (approved):

nshy:
I reworked the backport. Actually we cannot remove `iproto_shutdown()` as we now stop relay cords using it. Let's keep it. Looks safe because we have panic on iproto request finishing timeout, so we should not introduce hangs in 3.0. Also we don't need in future backport panic fixes to 3.0 I guess.

locker (approved):

ochaplashkin (approved):

PR #9725, opened 2024-02-22
build/lua: add TEST_BUILD define/tarantool.build flag

nshy:
In the commit 22d507d5e968 ("iproto: don't hang on uncancellable iproto request") we used `TEST_BUILD` define which is absent, we only have a CMake build option with such name. Let's add a define too.

While at it let's also show this flag in the `tarantool.build` table.

Follow-up #8423

### DIFF
--- a/changelogs/unreleased/add-test-build-flag-to-tarantool-build.md
+++ b/changelogs/unreleased/add-test-build-flag-to-tarantool-build.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Added a new flag `tarantool.build.test_build` that shows whether a build
+  flag `TEST_BUILD` is set.
--- a/src/trivia/config.h.cmake
+++ b/src/trivia/config.h.cmake
@@ -299,6 +299,7 @@
 #cmakedefine BUILD_STATIC 1
 #cmakedefine EMBED_LUAZLIB 1
 #cmakedefine EMBED_LUAZIP 1
+#cmakedefine TEST_BUILD 1
 
 /*
  * vim: syntax=c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -818,7 +818,7 @@ elseif(IS_DIRECTORY .git AND GIT)
     )
 endif()
 
-option(TEST_BUILD "Abort if memory leak is found." OFF)
+option(TEST_BUILD "Use defaults suited for tests" OFF)
 set(ABORT_ON_LEAK_DEFAULT ${TEST_BUILD})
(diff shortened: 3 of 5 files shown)


locker (approved):

changelogs/unreleased/add-test-build-flag-to-tarantool-build.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Added a new flag `tarantool.build.test_build` that shows whether build
+  flag `TEST_BUILD` is set.
```
andreyaksenov, line 4:
```suggestion
* Added a new flag `tarantool.build.test_build` that shows whether a build
  flag `TEST_BUILD` is set.
```

andreyaksenov (approved):

src/trivia/config.h.cmake
```diff
@@ -299,6 +299,7 @@
 #cmakedefine BUILD_STATIC 1
 #cmakedefine EMBED_LUAZLIB 1
 #cmakedefine EMBED_LUAZIP 1
+#cmakedefine TEST_BUILD 1
```
Totktonada, line 302: Nits:

1. Shouldn't we update the description of the option in `CMakeLists.txt:821`? It only mentions a memory leak now.
2. Shouldn't we describe what behavior is affected by the option in the documentation request?

ochaplashkin (approved):

changelogs/unreleased/add-test-build-flag-to-tarantool-build.md
nshy: Done.

src/trivia/config.h.cmake
nshy: 1. Fixed. The description was mistaken from the beginning.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1b357f38c..bd9082973 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -818,7 +818,7 @@ elseif(IS_DIRECTORY .git AND GIT)
     )
 endif()

-option(TEST_BUILD "Abort if memory leak is found." OFF)
+option(TEST_BUILD "Use defaults suited for tests" OFF)
 set(ABORT_ON_LEAK_DEFAULT ${TEST_BUILD})
 option(ABORT_ON_LEAK "Abort if memory leak is found." ${ABORT_ON_LEAK_DEFAULT})
```

2. Looks like we don't document build options.

PR #9727, opened 2024-02-22
core: do not limit the length of an error message by 512 bytes

### LINKED ISSUE #4975: Errors in init scripts are stripped
Tarantool version: 2.5.0-16-g9f2819785 (master)

Bug description:

Tarantool strips output if init script raises an error:

```console
$ tarantool test.lua
LuajitError: test.lua:1: module 'unknown' not found:
        no field package.preload['unknown']
        no file '/tmp/tmp.amMqKJqgsu/unknown.lua'
        no file '/tmp/tmp.amMqKJqgsu/unknown/init.lua'
        no file '/tmp/tmp.amMqKJqgsu/unknown.so'
        no file '/tmp/tmp.amMqKJqgsu/.rocks/share/tarantool/unknown.lua'
        no file '/tmp/tmp.amMqKJqgsu/.rocks/share/tarantool/unknown/init.lua'
        no file '/tmp/.rocks/share/tarantool/unknown.lua'
        no file '/tmp/.rocks/share/tarantool/unknown/init.lua'
        no file '/.rocks/share/tarantool/unknown.lua'
        no file '/.r
```

Steps to reproduce:

```bash
echo 'require("unknown")' > test.lua
tarantool ./test.lua
```

Gumix:
Now the error message is allocated by `malloc` if it doesn't fit into the static buffer.

Closes #4975


kostja:
What a wonderful idea, run out of memory when running out of memory.

src/lib/core/diag.c
```diff
@@ -158,17 +171,43 @@ error_format_msg(struct error *e, const char *format, ...)
...
+	va_copy(ap2, ap1);
+	int static_buf_size = sizeof(e->errmsg_buf);
+	int curr_len = strlen(e->errmsg);
+	int len = vsnprintf(NULL, 0, format, ap1);
+	int new_size = curr_len + len + 1;
+	va_end(ap1);
+
+	bool is_malloced = e->errmsg != e->errmsg_buf;
+	if (is_malloced || new_size > static_buf_size) {
+		char *new_buf = xmalloc(new_size);
```
CuriousGeorgiy, line 185: @kostja has a good point: we'd better log the error rather than just panic and lose it completely.

kostja: The original static buffer was there to make sure we can squeak something meaningful even if running out of memory and can't allocate the error message.

Gumix: It's unlikely that `OutOfMemory` error message will be longer than 512 bytes, so it will be printed to the static buffer.

CuriousGeorgiy: > It's unlikely that `OutOfMemory` error message will be longer than 512 bytes, so it will be printed to the static buffer.

The point is that we can run out of memory by coincidence.

test/box/error.test.lua
```diff
@@ -275,3 +275,7 @@ box.schema.func.drop('runtimeerror')
 --
 err = box.error.new('TestType', 'Message arg1: %s. Message arg2: %u', '1', 2)
 err.message
+
+-- gh-4975: errors in init scripts are stripped
+err = box.error.new('TestType', string.rep('a', 10000))
+string.len(err.message)
```
Totktonada, line 281: Shouldn't we also test that the error message is not actually stripped if an init script raises an error? I usually use `treegen` and `justrun` testing helpers for such tests.

andreyaksenov (approved):

ochaplashkin (approved):

src/lib/core/diag.c
```diff
@@ -158,17 +171,43 @@ error_format_msg(struct error *e, const char *format, ...)
...
+	va_list ap_copy;
+	va_copy(ap_copy, ap);
+	error_msg_destroy(e);
+	int static_buf_size = sizeof(e->errmsg_buf);
+	int len = vsnprintf(e->errmsg, static_buf_size, format, ap);
+	bool is_truncated = len >= static_buf_size;
+	if (is_truncated) {
+		e->errmsg = xmalloc(len + 1);
+		int new_len = vsnprintf(e->errmsg, len + 1, format, ap_copy);
+		VERIFY(new_len == len);
```
nshy, line 210: It does not look good that we have append and allocate buffer without exponential growth.

Also `error_msg_destroy` does not look good here. With amortized buffer we can try to reuse it.

Also as as Kostya mentioned the reason we have this buffer is not performance, may be we can reevaluate design and drop the preallocated buffer.

src/lib/core/diag.c
nshy: IMHO the chances that we can allocate error but then fail to allocate message buffer is low.

Anyway OOM is not happened in practice on Linux. See #https://github.com/tarantool/tarantool/issues/3534

nshy (changes requested):

src/lib/core/diag.c
CuriousGeorgiy: Okay, I guess if we already have tarantool/tarantool@5a031fb6, then we should follow the current policy.

src/lib/core/diag.h
```diff
@@ -97,8 +97,12 @@ struct error {
 	unsigned line;
 	/* Source file name. */
 	char file[DIAG_FILENAME_MAX];
-	/* Error description. */
-	char errmsg[DIAG_ERRMSG_MAX];
+	/*
```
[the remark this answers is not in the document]
Gumix: Fixed.

test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
```diff
@@ -0,0 +1,22 @@
...
+Require a non-existent module with a long enough name to get an error message,
+which is longer than 1000 characters, like:
+
+LuajitError: module 'aaa<...>aaa' not found:
+    no file '/usr/local/lib/lua/5.1/aaa<...>aaa.so'
+    no file '/usr/local/lib/lua/5.1/override/aaa<...>aaa.so'
+    <...>
+]]
+g.test_long_init_error = function(cg)
+    local dir = fio.tempdir()
```
Totktonada, line 17: This temporary directory is never removed (and it is why I usually use `test.treegen` for such cases).

However, we can run the process in the current directory (pass `'.'` instead of `dir` to `justrun.tarantool()`).

test/box/error.test.lua
Totktonada: Thanks!

src/lib/core/diag.c
Gumix: > It does not look good that we have append and allocate buffer without exponential growth.

There is no sense in optimizing `error_append_msg()` for long messages. It is used in a few places, which likely will be switched to error payloads.

> Also error_msg_destroy does not look good here. With amortized buffer we can try to reuse it.

Removed `error_msg_destroy()`.

> Also as as Kostya mentioned the reason we have this buffer is not performance, may be we can reevaluate design and drop the preallocated buffer.

Let's keep the preallocated buffer.

test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
Gumix:
```diff
--- a/test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
+++ b/test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
@@ -14,9 +14,8 @@ LuajitError: module 'aaa<...>aaa' not found:
     <...>
 ]]
 g.test_long_init_error = function(cg)
-    local dir = fio.tempdir()
     local module_name = string.rep('a', 500)
     local args = {'-l', module_name, '-e', '""'}
-    local res = justrun.tarantool(dir, {}, args, {stderr = true, nojson = true})
+    local res = justrun.tarantool('.', {}, args, {stderr = true, nojson = true})
     t.assert_gt(string.len(res.stderr), 1000, 'error is longer than 1000 chars')
 end
```

src/lib/core/diag.c
```diff
@@ -158,17 +168,45 @@ error_format_msg(struct error *e, const char *format, ...)
...
+	va_list ap1, ap2;
+	va_start(ap1, format);
+	va_copy(ap2, ap1);
+	int static_buf_size = sizeof(e->errmsg_buf);
+	int curr_len = strlen(e->errmsg);
+	int len = vsnprintf(NULL, 0, format, ap1);
+	int new_size = curr_len + len + 1;
+	va_end(ap1);
+
+	if (error_msg_is_malloced(e) || new_size > static_buf_size) {
```
nshy, line 180: We don't need the first condition really.

Gumix: Thanks, fixed:

```diff
diff --git a/src/lib/core/diag.c b/src/lib/core/diag.c
index 228ca462f..123aa47d0 100644
--- a/src/lib/core/diag.c
+++ b/src/lib/core/diag.c
@@ -177,7 +177,7 @@ error_append_msg(struct error *e, const char *format, ...)
        int new_size = curr_len + len + 1;
        va_end(ap1);

-       if (error_msg_is_malloced(e) || new_size > static_buf_size) {
+       if (new_size > static_buf_size) {
                char *new_buf = xmalloc(new_size);
                memcpy(new_buf, e->errmsg, curr_len + 1);
                if (error_msg_is_malloced(e))
```

src/lib/core/diag.c
CuriousGeorgiy: /cc @sergepetrenko

test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
```diff
@@ -0,0 +1,20 @@
...
+LuajitError: module 'aaa<...>aaa' not found:
+    no file '/usr/local/lib/lua/5.1/aaa<...>aaa.so'
+    no file '/usr/local/lib/lua/5.1/override/aaa<...>aaa.so'
+    <...>
+]]
+g.test_long_init_error = function()
+    local module_name = string.rep('a', 500)
+    local args = {'-l', module_name, '-e', '""'}
+    local res = justrun.tarantool('.', {}, args, {stderr = true, nojson = true})
+    t.assert_gt(string.len(res.stderr), 1000, 'error is longer than 1000 chars')
```
CuriousGeorgiy, line 19: Nit: maybe at least match the error (`LuajitError: module 'aaa<...>aaa' not found`) with `assert_str_matches` for sanity?

src/lib/core/diag.c
```diff
@@ -158,17 +168,45 @@ error_format_msg(struct error *e, const char *format, ...)
...
+	va_copy(ap2, ap1);
+	int static_buf_size = sizeof(e->errmsg_buf);
+	int curr_len = strlen(e->errmsg);
+	int len = vsnprintf(NULL, 0, format, ap1);
+	int new_size = curr_len + len + 1;
+	va_end(ap1);
+
+	if (new_size > static_buf_size) {
+		char *new_buf = xmalloc(new_size);
+		memcpy(new_buf, e->errmsg, curr_len + 1);
```
CuriousGeorgiy, line 182: Nit: why `curr_len + 1`?

CuriousGeorgiy (approved):

src/lib/core/diag.c
sergepetrenko: > Okay, I guess if we already have https://github.com/tarantool/tarantool/commit/5a031fb6, then we should follow the current policy.

Sounds good.

src/lib/core/diag.c
Gumix: I feel a little uncomfortable leaving non-terminated strings in memory, even for a while.

test/app-luatest/gh_4975_errors_in_init_scripts_are_stripped_test.lua
Gumix: Done.

sergepetrenko:
Backports:
2.11 https://github.com/tarantool/tarantool/pull/9770
3.0 https://github.com/tarantool/tarantool/pull/9771

PR #9730, opened 2024-02-25
config: add ordered dictionary (internal module)

Totktonada:
In development of the configuration module we met several cases, when an order of appearance of some items is as important as a constant time access using a key.

This commit adds a collection that serves such a need. It is to be used internally in the src/box/lua/config code.

The module is inspired by Python's collections.OrderedDict. See the description in the module code for details.

### DIFF
--- a/src/box/lua/config/utils/odict.lua
+++ b/src/box/lua/config/utils/odict.lua
@@ -1,0 +1,213 @@
+-- Ordered dictionary.
+--
+-- The collection tracks the order in which the items are added
+-- and provides odict.pairs() function to get them in this order.
+--
+-- The ordered dictionary is a usual Lua table with a specific
+-- metatable. All the table operations are applicable.
+--
+-- It is similar to Python's collections.OrderedDict.
+--
+-- Example:
+--
+--  | local od = odict.new()
+--  |
+--  | od.a = 1
+--  | od.b = 2
+--  | od.c = 3
+--  |
+--  | print('od.a', od.a) -- 1
+--  | print('od.b', od.b) -- 2
+--  | print('od.c', od.c) -- 3
+--  |
+--  | for k, v in odict.pairs(od) do
+--  |     print(k, v)
+--  | end
+--  | -- print: a, 1
+--  | -- print: b, 2
+--  | -- print: c, 3
+--
+-- If an element is deleted and added again, it is added to the
+-- end.
+--
+--  | local od = odict.new()
+--  |
+--  | od.a = 1
+--  | od.b = 2
+--  | od.c = 3
+--  |
+--  | od.b = nil
+--  | od.b = 4
+--  |
+--  | for k, v in odict.pairs(od) do
+--  |     print(k, v)
+--  | end
+--  | -- print: a, 1
+--  | -- print: c, 3
+--  | -- print: b, 4
+--
+-- If an element is changed (without prior deletion), it remains
+-- on the same position.
+--
+-- Beware: console shows the fields as unordered. The same for the
+-- serialization into JSON/YAML/MessagePack formats. It should be
+-- solved after gh-9747.
+
+local REINDEX_THRESHOLD = 1000
+
+local registry = setmetatable({}, {__mode = 'k'})
+
+-- Generator function for odict.pairs().
+--
+-- Returns (key, value).
+local function gen(od, prev_key)
+    local ctx = registry[od]
+
+    -- The previous key is nil only on the first call of the
+    -- generator function.
+    local id = prev_key == nil and 0 or ctx.key2id[prev_key]
+    -- NB: This assert doesn't catch all the kinds of changes
+    -- during an iteration. It rather verifies a precondition
+    -- for the following cycle.
+    assert(id ~= nil, 'ordered dictionary is changed during iteration')
+
+    while id <= ctx.max_id do
+        id = id + 1
+        local key = ctx.id2key[id]
+        -- id2key may have holes for IDs <= max_id in place of
+        -- deleted items.
+        --
+        -- id2key may contain a stalled entry, because __newindex
+        -- is not called on assignment of an existing field,
+        -- including assignment to nil.
+        if key ~= nil and od[key] ~= nil then
+            return key, od[key]
+        end
+    end
+end
+
+-- Returns a Lua iterator triplet (gen, param, state).
+--
+-- Usage:
+--
+--  | for k, v in odict.pairs(od) do
+--  |     <...>
+--  | end
+--
+-- The iterator stability guarantees are the same as for usual
+-- pairs().
+--
+-- Quote from the Lua 5.1 reference manual:
+--
+-- > The behavior of `next` is undefined if, during the traversal,
+-- > you assign any value to a non-existent field in the table.
+-- > You may however modify existing fields. In particular, you
+-- > may clear existing fields.
+local function _pairs(od)
+    return gen, od, nil
+end
+
+-- Delete the key from the id<->key mappings.
+local function release(ctx, key)
+    local id = ctx.key2id[key]
+    if id ~= nil then
+        ctx.key2id[key] = nil
+        ctx.id2key[id] = nil
+
+        -- No need to grow the IDs if the same key is assigned and
+        -- deleted in a loop.
+        --
+        -- for <...> do
+        --     od.x = 'x'
+        --     od.x = nil
+        -- end
+        if id == ctx.max_id then
+            ctx.max_id = ctx.max_id - 1
+        end
+    end
+end
+
+-- Track the new key in the id<->key mappings.
+local function track(ctx, key)
+    local id = ctx.max_id + 1
+    ctx.id2key[id] = key
+    ctx.key2id[key] = id
+    ctx.max_id = id
+end
+
+-- Renew the id<->key mappings to eliminate holes.
+local function reindex(od, ctx)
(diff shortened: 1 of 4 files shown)


Totktonada:
`checkpatch.pl` complains about using `od` as a variable name in Lua (it states that it is misspelled 'of'). Filed https://github.com/tarantool/checkpatch/issues/70.

src/box/lua/config/utils/odict.lua
```diff
@@ -0,0 +1,209 @@
...
+-- > The behavior of `next` is undefined if, during the traversal,
+-- > you assign any value to a non-existent field in the table.
+-- > You may however modify existing fields. In particular, you
+-- > may clear existing fields.
+local function _pairs(od)
+    return gen, od, nil
+end
+
+-- Delete the key from the id<->key mappings.
+local function trash(ctx, key)
```
Lord-KA, line 107: I would prefer a different function name here, because "to trash something" has a different meaning. Maybe: `remove_key(ctx, key)` or `drop_key(ctx, key)`

src/box/lua/config/utils/odict.lua
```diff
@@ -0,0 +1,209 @@
...
+        ctx.key2id[key] = nil
+        ctx.id2key[id] = nil
+
+        -- No need to grow the IDs if the same key is assigned and
+        -- deleted in a loop.
+        --
+        -- for <...> do
+        --     od.x = 'x'
+        --     od.x = nil
+        -- end
```
Lord-KA, line 119: Do you mean that we should have sequential ids?

src/box/lua/config/utils/odict.lua
```diff
@@ -0,0 +1,209 @@
...
+    end
+end
+
+-- Track the new key in the id<->key mappings.
+local function track(ctx, key)
+    local id = ctx.max_id + 1
+    ctx.id2key[id] = key
+    ctx.key2id[key] = id
+    ctx.max_id = id
+end
```
Lord-KA, line 132: Nit:
This seems more like a `new_index()`/`add_key()` function.
Actually, it would be nice to have `drop_key()` and `add_key()` instead of `trash()` and `track()`.

src/box/lua/config/utils/odict.lua
```diff
@@ -0,0 +1,209 @@
...
+    local old_max_id = ctx.max_id
+    ctx.max_id = 0
+    for id = 1, old_max_id do
+        local key = ctx.id2key[id]
+        -- Drop the given key-id pair from the key<->id mappings.
+        if key ~= nil then
+            trash(ctx, key)
+        end
+        -- Add the new key-id pair into the key<->id mappings.
+        if key ~= nil and od[key] ~= nil then
```
Lord-KA, line 151: Nit:
I may be mistaken, but it seems like we don't need `key ~= nil` check, do we?

Lord-KA (approved):
Thanks for the patch. Overall looks good, see my 5 suggestions below.

src/box/lua/config/utils/odict.lua
```diff
@@ -0,0 +1,209 @@
...
+--  |
+--  | od.b = nil
+--  | od.b = 4
+--  |
+--  | for k, v in odict.pairs(od) do
+--  |     print(k, v)
+--  | end
+--  | -- print: a, 1
+--  | -- print: c, 3
+--  | -- print: b, 4
```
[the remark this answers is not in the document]
Totktonada: Thanks for noticing! I've filed #9747. I'll also mention this detail in the module header comment.

I'm against implicit transformation from a mapping to an array. I would rather leave the ordered serialization unimplemented (but keep the shape of the collection as is) and implement such serialization mechanism later (without changing the shape).

src/box/lua/config/utils/odict.lua
Totktonada: Internal IDs? Nope.

This code may be removed without any visible behavior change except more frequent reindexing in cases like the quoted one.

src/box/lua/config/utils/odict.lua
Totktonada: Yep. We can use `nil` as a table index key and will always get `nil` as a result (but we can't assign a `nil` key). However, I like this condition to be written explicitly, because:

1. There is the `key ~= nil` condition above and it is easier to spot that both `trash()` and `track()` are skipped in case of `key == nil`.
2. Using `nil` as an index key for a table is a trick (not known by any Lua user) and it anyway should be commented. An explicit check looks better for me than an implicit one + comment.

src/box/lua/config/utils/odict.lua
Totktonada: I've started from `set` and `drop` functions, but than found that I need functions to control tracking of the keys separately from assignments into the key-value table.

The idea of the `trash` and `track` names are that they have no association with the key-value table and dedicated only to the internal tracking structures. IOW, they have no `drop`/`delete`/`remove` and `set`/`add`/`append` words. Also, `trash` reminds the `TRASH` macro from `src/trivia/util.h`.

I like these names, but I would hear thoughts from other reviewers before a decision to keep or change them.

src/box/lua/config/utils/odict.lua
Totktonada: Answered [above](https://github.com/tarantool/tarantool/pull/9730#discussion_r1505606234).

src/box/lua/config/utils/odict.lua
Lord-KA: If you prefer so, sure.

ImeevMA (approved):
Thank you for the patch! LGTM.

src/box/lua/config/utils/odict.lua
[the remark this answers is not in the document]
Totktonada: While we're in a discussion, I've added a note.

```diff
diff --git a/src/box/lua/config/utils/odict.lua b/src/box/lua/config/utils/odict.lua
index 925e2d1f43..dce3e9ad52 100644
--- a/src/box/lua/config/utils/odict.lua
+++ b/src/box/lua/config/utils/odict.lua
@@ -48,6 +48,10 @@
 --
 -- If an element is changed (without prior deletion), it remains
 -- on the same position.
 --
+-- Beware: console shows the fields as unordered. The same for the
+-- serialization into JSON/YAML/MessagePack formats. It should be
+-- solved after gh-9747.
 
 local REINDEX_THRESHOLD = 1000
```

src/box/lua/config/utils/odict.lua
Totktonada: Renamed.

```diff
diff --git a/src/box/lua/config/utils/odict.lua b/src/box/lua/config/utils/odict.lua
index dce3e9ad52..ffefb00ec0 100644
--- a/src/box/lua/config/utils/odict.lua
+++ b/src/box/lua/config/utils/odict.lua
@@ -108,7 +108,7 @@ local function _pairs(od)
 end
 
 -- Delete the key from the id<->key mappings.
-local function trash(ctx, key)
+local function release(ctx, key)
     local id = ctx.key2id[key]
     if id ~= nil then
         ctx.key2id[key] = nil
@@ -149,7 +149,7 @@ local function reindex(od, ctx)
         local key = ctx.id2key[id]
         -- Drop the given key-id pair from the key<->id mappings.
         if key ~= nil then
-            trash(ctx, key)
+            release(ctx, key)
         end
         -- Add the new key-id pair into the key<->id mappings.
         if key ~= nil and od[key] ~= nil then
@@ -168,7 +168,7 @@ local mt = {
 
         -- If a field with this key had existed, we should update
         -- the key<->id mappings to let gen() know the new order.
-        trash(ctx, k)
+        release(ctx, k)
 
         -- If a non-existing field is assigned to nil, we have
         -- nothing to do.
```

src/box/lua/config/utils/odict.lua
Totktonada: Updated, see the linked discussion thread.

Totktonada:
> `checkpatch.pl` complains about using `od` as a variable name in Lua (it states that it is misspelled 'of'). Filed https://github.com/tarantool/checkpatch/issues/70.

Fixed in PR #9752. Rebased this patch on top of it.

src/box/lua/config/utils/odict.lua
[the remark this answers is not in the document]
igormunkin: Well, the idea is quite nice, but I agree with @Totktonada's argument from our offline discussion: though there is already the similar prisedanie with `box.cfg`, I doubt we have to manually patch `pairs` builtin for this. The reason this hack is implemented for `box.cfg` relates to this object nature: it's the singleton being often observed by humans. But do we need implement this hack for such a basic thing? I don't think so.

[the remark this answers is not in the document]
igormunkin: The same argument works for `__serialize`, `compat` table and odict object.

ochaplashkin (approved):

src/box/lua/config/utils/odict.lua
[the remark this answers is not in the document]
Totktonada: @Lord-KA Can I resolve this discussion leaving the 'beware' comment in the code and the issue to return here back later?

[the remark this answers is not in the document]
Lord-KA: I am okay with solving it separately in https://github.com/tarantool/tarantool/issues/9747.

PR #9733, opened 2024-02-26
cmake: add libresolv to static dependency list

ylobankov:
If we run a static build with ASAN enabled via Clang 16, the build will fail unless `libresolv` is in the white list of static dependencies. It looks like it is a peculiarity of Clang 16 and higher.

Fixes https://github.com/tarantool/tarantool/issues/9740

### DIFF
--- a/cmake/CheckDependencies.cmake
+++ b/cmake/CheckDependencies.cmake
@@ -30,6 +30,10 @@ elseif(UNIX)
         libsvace
         libstdc++
     )
+    # See for details https://github.com/tarantool/tarantool/issues/9740
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND ENABLE_ASAN)
+        set(ALLOWLIST ${ALLOWLIST} libresolv)
+    endif()
 else()
     message(FATAL_ERROR "Unknown platform")
 endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -427,6 +427,8 @@ if(BUILD_STATIC AND OPENSSL_USE_STATIC_LIBS)
         TARGET tarantool POST_BUILD
         COMMAND ${CMAKE_COMMAND}
             -DFILE=${PROJECT_BINARY_DIR}/src/tarantool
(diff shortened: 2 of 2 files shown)


cmake/CheckDependencies.cmake
```diff
@@ -29,6 +29,7 @@ elseif(UNIX)
         libpthread
         libsvace
         libstdc++
+        libresolv
```
locker, line 32: This shouldn't happen because we [use](https://github.com/tarantool/tarantool-ee/blob/5ff219582c92c5742c1204900c0ccb15973d68b4/cmake/BuildLDAP.cmake#L80-L94) static libresolv on Linux.

I tried building Tarantool EE statically with ASAN enabled with the following options:

```
CC=clang CXX=clang++ cmake -S . -B build/debug-asan-static -DENABLE_ASAN=ON -DENABLE_UB_SANITIZER=ON -DTEST_BUILD=ON -DBUILD_STATIC_WITH_BUNDLED_LIBS=ON
```

The resulting binary didn't have libresol as a dependency:

```
$ ldd build/debug-asan-static/tarantool/src/tarantool
        linux-vdso.so.1 (0x00007ffcf779e000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f99e2b19000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f99e2800000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f99e92cd000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f99e2400000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f99e930a000)
```

locker: This looks like a bug. We need to figure out why this happens and whether we can eliminate this dependency.

ylobankov: Filed https://github.com/tarantool/tarantool/issues/9740.

locker (approved):

PR #9734, opened 2024-02-27
Cover some implementations of `port` interface with tests

drewdzzz:
The patch introduces new `unit/port.cc` test that covers almost all port implementations and almost all port methods. This test was intended as a testing point for future port changes.
    
Implementations `port_sql` and `port_vdbemem` and method  `dump_vdbemem` are not tested because I don't know much about the sql subsystem. For the same reason, I didn't tested `port_dump_msgpack_with_context`. Method `dump_msgpack_16` is not actually supported, but it is still used for compatibility with very old connectors. The problem is I found a bug in `port_c_dump_msgpack_16` while writing the test, and we don't want to fix it - probably, we will get rid of this method in future. Method `dump_plain` is not tested because it requires functions from `console.lua` file - it is easier to test `console.push` from Lua then loading the Lua file in unit test.

Along the way, the patch fixes some minor problems detected by the test.

### DIFF
--- a/src/box/lua/misc.cc
+++ b/src/box/lua/misc.cc
@@ -200,8 +200,6 @@ extern "C" void
 port_c_dump_lua(struct port *base, struct lua_State *L,
 		enum port_dump_lua_mode mode)
 {
-	assert(mode == PORT_DUMP_LUA_MODE_FLAT ||
-	       mode == PORT_DUMP_LUA_MODE_TABLE);
 	struct port_c *port = (struct port_c *)base;
 	if (mode == PORT_DUMP_LUA_MODE_MP_OBJECT) {
 		port_dump_lua_mp_object_mode_slow(base, L, &fiber()->gc,
--- a/test/unit/port.cc
+++ b/test/unit/port.cc
@@ -1,0 +1,788 @@
+#include "lua_test_utils.h"
+
+#include "box/lua/call.h"
+#include "box/lua/tuple.h"
+#include "box/port.h"
+#include "box/session.h"
+#include "box/tuple.h"
+#include "box/user.h"
+#include "core/event.h"
+#include "core/port.h"
+#include "lua/init.h"
+#include "lua/minifio.h"
+#include "lua/msgpack.h"
+#include "lua/utils.h"
+#include "memory.h"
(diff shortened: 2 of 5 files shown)


test/unit/port.cc
```diff
@@ -0,0 +1,779 @@
...
+
+/**
+ * Unpacks Lua table which is on top of passed Lua stack.
+ * The table is popped, its contents are pushed to the Lua stack.
+ */
+static void
+lua_table_unpack(struct lua_State *L)
+{
+	const char *text = "return function(t) return unpack(t) end";
+	int rc = luaT_dostring(L, text);
```
CuriousGeorgiy, line 100: Nit: you could use `lua_getglobal(L, "unpack")`.

test/unit/port.cc
```diff
@@ -0,0 +1,779 @@
...
+		"for k, _ in pairs(b) do "
+		"    if type(a[k]) == 'nil' then "
+		"        return false "
+		"    end "
+		"end "
+		"return true "
+		"end "
+		"return table_eq";
+	int rc = luaT_dostring(L, text);
+	fail_if(rc != 0);
```
CuriousGeorgiy, line 83: Nit: you could cache the function definition in `main` once.

CuriousGeorgiy (changes requested):

ochaplashkin (approved):

src/box/lua/misc.cc
```diff
@@ -200,8 +200,6 @@ extern "C" void
 port_c_dump_lua(struct port *base, struct lua_State *L,
 		enum port_dump_lua_mode mode)
 {
-	assert(mode == PORT_DUMP_LUA_MODE_FLAT ||
```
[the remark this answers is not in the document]
drewdzzz: Done (used tuple_constraint)

test/unit/port.cc
drewdzzz: Done

test/unit/port.cc
drewdzzz: Done

CuriousGeorgiy (approved):

PR #9736, opened 2024-02-27
lyaml: fix alias serialization

### LINKED ISSUE #8310: YAML serializer doesn't detect references between table and nested table with `__serialize` meta method
PR https://github.com/tarantool/tarantool/pull/8245 fixes alias detection in a table returned by the `__serialize` meta method. However, if a table returned by the `__serialize` meta method references fields from an outer table, they aren't identified as aliases:

```
tarantool> t = {}
---
...

tarantool> {a = t, b = setmetatable({}, {__serialize = function() return {t, t} end})}
---
- a: []
  b:
  - &0 []
  - *0
...
```

To fix this issue, we need to add yet another phase to the YAML serializer:
1. Run all `__serialize` methods recursively. **This is the new phase.**
2. Build a reference table.
3. Encode the resulting object in YAML.

While performing the new phase, the YAML serialize also need to maintain the table mapping all objects to their serialization so that we don't call `__serialize` multiple times for different references of the same object.

### LINKED ISSUE #8321: YAML serializer doesn't detect aliases to userdata objects with `__serialize` metamethod
We fixed detection of references to tables with `__serialize` metamethod in #8245, but the fix doesn't work for userdata/cdata objects

```
tarantool> f = require('fiber').self()
---
...

tarantool> type(f)
---
- userdata
...

tarantool> {f, f}
---
- - status: running
    name: interactive
    id: 103
  - status: running
    name: interactive
    id: 103
...
```

This happens, because we don't keep track of userdata/cdata objects in the Lua YAML serializer:

https://github.com/tarantool/tarantool/blob/5ed111293a75e8cfef74c5707a32218098b2f93a/third_party/lua-yaml/lyaml.cc#L786-L789

We should fix that and add a test case.

Related issues: #8240 #8310

### LINKED ISSUE #8350: yaml: unnecessary anchors
Tarantool version: 2.11.0-entrypoint-1037-gbf8b76a4d.

```yaml
Tarantool 2.11.0-entrypoint-1037-gbf8b76a4d
tarantool> require('tarantool')
---
- build: &0
    target: Linux-x86_64-Debug
    options: cmake . -DCMAKE_INSTALL_PREFIX=/usr -DENABLE_BACKTRACE=TRUE
    linking: dynamic
    mod_format: so
    flags: -march=native -O2 -ggdb -fexceptions -funwind-tables -fasynchronous-unwind-tables
      -fno-common -fopenmp -msse2 -mavx -Wformat -Wformat-security -Werror=format-security
      -fstack-protector-strong -fPIC -fmacro-prefix-map=/var/tmp/portage/dev-db/tarantool-9999/work/tarantool-9999=.
      -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type
      -Werror
    compiler: GNU-12.2.1
  package: Tarantool
  debug: &1
    getsources: 'function: 0x40fcd508'
  uptime: 'function: 0x408364d8'
  version: 2.11.0-entrypoint-1037-gbf8b76a4d
  pid: 'function: 0x408364f8'
...
```

`build` and `debug` are anchored, but not present second time in the output.

nshy:
The #8350 was introduced by the commit b42302f52ada ("lua-yaml: enable aliasing for objects returned by __serialize") so the patch is effectively reversed.

The idea is to call all object `__serialize()` methods recursively before finding references. In the process we need to keep an eye on reference structure of course.

Closes #8350
Closes #8321
Closes #8310


changelogs/unreleased/fix-yaml-alias-serialization.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/lua
+
+* Fixed yaml alias serialization (gh-8350, gh-8321, gh-8310).
```
andreyaksenov, line 3:
```suggestion
* Fixed serialization of YAML aliases (gh-8350, gh-8321, gh-8310).
```

andreyaksenov (approved):

ochaplashkin (approved):

third_party/lua-yaml/lyaml.cc
```diff
@@ -819,6 +815,113 @@ static void find_references(struct lua_yaml_dumper *dumper) {
...
+
+      lua_pushnil(L);
+      while (lua_next(L, -2)) {
+         /* copy key for iteration */
+         lua_pushvalue(L, -2);
+         lua_insert(L, -3);
+         materialize(dumper, -1, reftable_index);
+         materialize(dumper, -2, reftable_index);
+         lua_settable(L, materialized_index);
+      }
```
Totktonada, line 900: Do you consider just creating a reference table (original object -> serialized object mapping) and use it in `find_references()` and `dump_node()` to replace the original table with the serialized one of the fly?

This way we reduce amount of needed temporary GC tracked objects.

Also, it allows the `dump_node()` code to traverse over the original table (if it has no `__serialize`), so we don't need to update so much tests here and in external projects with the same testing approaches.

I've played with it a bit. The patch is [here](https://github.com/Totktonada/tarantool/commit/cd91395d473a2674d6c1f9e69a3b4657db51a92d). It is a little dirty, but at least it seems that this approach is possible to implement and it doesn't look much more complicated.

third_party/lua-yaml/lyaml.cc
```diff
@@ -819,6 +815,113 @@ static void find_references(struct lua_yaml_dumper *dumper) {
...
+ * keeps the reference structure of the original table. Also the reference
+ * table is checked for luaL_checkfield result so that reference structure
+ * brought by __serialize is kept too. Reference table is at `reftable_index`
+ * on stack.
+ *
+ * Also tables of materialized object keep the hints of __serialize as the
+ * hints affects later serialization.
+ */
+static void
+materialize(struct lua_yaml_dumper *dumper, int obj_index, int reftable_index)
```
Totktonada, line 856: Just raw idea: wouldn't it be simpler to write the traversal code using Lua (place to `src/box/serializer.lua`) and call it from the C code using `lua_call()`?

I suspect that there may be problems accessing `__serialize` metamethod for cdata, but I didn't look deeper.

Feel free to close this discussion without any actions -- this is just thought around.

Totktonada (approved):
Thank you for the decent work!

I've several topics to discuss together. Please, consider them in the discussion threads below.

nshy:
Closed as superseded by PR https://github.com/tarantool/tarantool/pull/9777

PR #9743, opened 2024-02-28
trigger: log error in triggers that does not throw an error

### LINKED ISSUE #9309: `box.ctl.on_schema_init` doesn't print error msg
If  `on_schema_init` fails, no error message is printed to log, which is very difficult to debug.

Current behavior:
```
tarantool> box.ctl.on_schema_init(function() log.warn('ON_SCHEMA_INIT') error('ERROR') log.warn('NOT PRINTED') end)
---
- 'function: 0x401c67c0'
...

tarantool> box.cfg{}
2023-10-26 13:31:40.561 [1821664] main/103/interactive I> Tarantool 3.0.0-alpha3-142-g39afee4e1 Linux-x86_64-Debug
2023-10-26 13:31:40.561 [1821664] main/103/interactive I> log level 5 (INFO)
2023-10-26 13:31:40.561 [1821664] main/103/interactive I> wal/engine cleanup is paused
2023-10-26 13:31:40.564 [1821664] main/103/interactive I> mapping 268435456 bytes for memtx tuple arena...
2023-10-26 13:31:40.564 [1821664] main/103/interactive I> Actual slab_alloc_factor calculated on the basis of desired slab_alloc_factor = 1.044274
2023-10-26 13:31:40.564 [1821664] main/103/interactive I> mapping 134217728 bytes for vinyl tuple arena...
2023-10-26 13:31:40.570 [1821664] main/103/interactive/tarantool [string "return box.ctl.on_schema_init(function() log...."]:1 W> ON_SCHEMA_INIT
2023-10-26 13:31:40.585 [1821664] main/103/interactive I> update replication_synchro_quorum = 1
<...>
```

Expected behavior: error message is printed to log.

drewdzzz:
Now on_access_denied and on_schema_init triggers do not throw error, so it just sinks into oblivion, which is not good because a thrown error indicates that something went wrong and database administrator should be definitely notified about it.

The commit makes Tarantool log errors thrown from these triggers with error level. Also, the commit adds a test checking if all triggers are either raise an error or log it.

Closes #9309

### DIFF
--- a/changelogs/unreleased/gh-9309-errors-in-triggers.md
+++ b/changelogs/unreleased/gh-9309-errors-in-triggers.md
@@ -1,0 +1,5 @@
+## bugfix/box
+
+* Now all triggers either have a direct impact on the execution flow (for
+  example, closing a connection or throwing an error) or print a message to
+  the error log when they throw an error (gh-9309).
--- a/src/box/error.cc
+++ b/src/box/error.cc
@@ -274,22 +274,26 @@ struct event *on_access_denied_event;
  * Runs on access denied triggers. Does not run triggers from the event if it
  * is not initialized.
  */
-static int
+static void
 run_on_access_denied_triggers(const char *access_type, const char *object_type,
 			      const char *object_name)
 {
-	struct on_access_denied_ctx trigger_ctx =
-		{access_type, object_type, object_name};
-	if (trigger_run(&on_access_denied, &trigger_ctx) != 0)
-		return -1;
-
-	if (on_access_denied_event == NULL)
-		return 0;
+	int rc = 0;
 	const char *name = NULL;
 	struct func_adapter *trigger = NULL;
 	struct func_adapter_ctx ctx;
 	struct event_trigger_iterator it;
-	int rc = 0;
+	struct on_access_denied_ctx trigger_ctx =
+		{access_type, object_type, object_name};
+
+	if (trigger_run(&on_access_denied, &trigger_ctx) != 0) {
+		rc = -1;
+		goto out;
+	}
+
+	if (on_access_denied_event == NULL)
+		goto out;
+
 	event_trigger_iterator_create(&it, on_access_denied_event);
 	while (rc == 0 && event_trigger_iterator_next(&it, &trigger, &name)) {
 		func_adapter_begin(trigger, &ctx);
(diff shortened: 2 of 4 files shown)


ochaplashkin (approved):

changelogs/unreleased/gh-9309-errors-in-triggers.md
```diff
@@ -0,0 +1,5 @@
+## feature/box
+
+* Now all triggers are either have a direct impact on the execution flow
+  (e.g. close a connection or throw an error) or print a message to error
+  log when they throw an error (gh-9309).
```
andreyaksenov, line 5:
```suggestion
* Now all triggers either have a direct impact on the execution flow
  (for example, closing a connection or throwing an error) or print a message to the error
  log when they throw an error (gh-9309).
```

Serpentian (approved):
Thank you for the patch and testing all of the triggers! Looks great!

CuriousGeorgiy (commented):
I don't see this how this patch scales to addition of new triggers. One can easily forget to add a trigger to this test. 

Is it not possible to test the trigger registry abstractly? I am not familiar with the trigger registry functionality so I am just making a shot in the dark.

drewdzzz:
@CuriousGeorgiy
Trigger registry is just a key-value storage (`String` -> `List<NamedTrigger>`), so I have no idea how to test the triggers abstractly. Moreover, some triggers (`net.box` ones, for example) do not use trigger registry. And every Tarantool triggers has its own scenario to be fired (I mean, `space:on_replace` is fired on replace, `net.box.conn:on_connect` is fired when it is connected and so on).

The only thing I came up is kind of hack:
We can patch `luatest` - `Server.new` will have new option `allow_arbitrary_triggers` or something like it. If you need arbitrary triggers (not used by Tarantool) - you need to pass this flag. If the flag is not set, we will check in `tarantool.trigger.on_change` trigger if the event is in the list of Tarantool events. If it is not - we will panic. Hence, when you creating a new trigger, you will place it in the trigger registry, so the test will fail with error "Unknown trigger, please add it to the list of Tarantool triggers". And in this test we could check that all the triggers from the list were checked.

It sounds "wild", but I can't imagine anything better.

CuriousGeorgiy (approved):
Please open a ticket in Tarantool or Luatest to track the issue we discussed with the solution you propose.

drewdzzz:
Done: #9799

sergepetrenko:
Backport 3.0: https://github.com/tarantool/tarantool/pull/9805

PR #9745, opened 2024-02-28
Hide redundant fields of box.error.unpack()

### LINKED ISSUE #9101: Hide useless error fields from unpack output
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#3f1a17a9673f408eaba2cdef28890886)

Some error fields only clutter the [`unpack`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/error_object/#lua-function.error_object.unpack) output and bear no useful information:

- `custom_type` - not set for Tarantool errors; same as `type` for user-defined errors.
- `base_type` - same as `type` for Tarantool errors; `CustomError` for user-defined errors.
- `code` - only makes sense if it’s set; 0 if unset.

Suggestion:

- Hide `custom_type` and `base_type` from `unpack` output.
- Hide `code` from `unpack` output if it’s 0.

Before:

```yaml
tarantool> box.error.new{type = 'MyError', reason = 'foobar'}:unpack()
---
- code: 0
  base_type: CustomError
  type: MyError
  custom_type: MyError
  message: foobar
  trace:
  - file: '[string "return box.error.new{type = ''MyError'', reason..."]'
    line: 1
...
```

After:

```yaml
tarantool> box.error.new{type = 'MyError', reason = 'foobar'}:unpack()
---
- type: MyError
  message: foobar
  trace:
  - file: '[string "return box.error.new{type = ''MyError'', reason..."]'
    line: 1
...
```

The fields may still be accessed via the `.` operator of an error object, just like before. In case of `code` it means that if the code is 0, `e.code` will return 0, not `nil`.

Comment from Vladislav Shpilevoy re `base_type`:

> `base_type` was exposed to allow people get the real error type in case a user is trying to mimic `ClientError` or `SystemError` or something else internal using the custom type field.
> 

However, since we only hide `base_type` from the `unpack` output, not delete it completely, one can still figure out if an error is a true system error by accessing `base_type` via the `.` operator so removing it from the `unpack` output seems to be acceptable.

Since this may potentially break existing users, we’ll introduce the new [`compat`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/compat/) module option `box_error_unpack_type_and_code` to revert to the old behavior.

nshy:
Closes #9101

### DIFF
--- a/changelogs/unreleased/gh-9101-hide-error-unpack-redundant-fields.md
+++ b/changelogs/unreleased/gh-9101-hide-error-unpack-redundant-fields.md
@@ -1,0 +1,5 @@
+## feature/box
+
+* Hide redundant fields from `box.error.unpack()` if
+  the `box_error_unpack_type_and_code` compat option is set to 'new'.
+  The default behaviour is 'old' (gh-9101).
--- a/src/lua/error.lua
+++ b/src/lua/error.lua
@@ -2,6 +2,7 @@
 
 local ffi = require('ffi')
 local msgpack = require('msgpack')
+local compat = require('compat')
 
 local mp_decode = msgpack.decode_unchecked
 
@@ -142,6 +143,14 @@ local function error_unpack(err)
     for i = 0, payload._count - 1 do
         local f = fields[i]
         result[ffi.string(f._name)] = mp_decode(f._data)
+    end
+    -- Hide redundant fields from unpack output (gh-9101).
+    if compat.box_error_unpack_type_and_code:is_new() then
+        result.custom_type = nil
+        result.base_type = nil
+        if result.code == 0 then
+            result.code = nil
+        end
     end
     return result
 end
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2351,6 +2351,12 @@ return schema.new('instance_config', schema.record({
         }, {
             default = 'new',
         }),
+        box_error_unpack_type_and_code = schema.enum({
+            'old',
+            'new',
+        }, {
+            default = 'old',
+        }),
     }),
 }, {
     -- This kind of validation cannot be implemented as the
--- a/src/lua/compat.lua
+++ b/src/lua/compat.lua
@@ -111,6 +111,14 @@ used as an error indicator in the box C API.
 https://tarantool.io/compat/box_space_max
 ]]
 
+local BOX_ERROR_UNPACK_TYPE_AND_CODE_BRIEF = [[
+Whether to show redundant fields in box.error.unpack(). The new behaviour
+is not to show 'base_type' and 'custom_type' fields. 'code' field is also not
+shown if it is 0. Note that 'base_type' is still accessible for error object.
+
+https://tarantool.io/compat/box_error_unpack_type_and_code
+]]
+
 -- Returns an action callback that toggles a tweak.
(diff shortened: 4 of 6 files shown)


Gumix (approved):

src/lua/error.lua
```diff
@@ -143,6 +144,14 @@ local function error_unpack(err)
         local f = fields[i]
         result[ffi.string(f._name)] = mp_decode(f._data)
     end
+    -- Hide redundant fields from unpack output (gh-9101).
+    if compat.box_error_unpack_type_and_code.current == 'new' then
```
Totktonada, line 148: I would use `:is_new()`, because `.curent` may be `'default'`. See PR #8858.

src/lua/error.lua
```diff
@@ -143,6 +144,14 @@ local function error_unpack(err)
...
+        result.custom_type = nil
+        result.base_type = nil
+        if result.code == 0 then
+            result.code = nil
+        end
```
Totktonada, line 153: AFAIR, the idea is to make it more concise in a console output/in a log message. IOW, it should affect serialization.

Am I understand right that it will become true after PR #9716?

Totktonada (approved):
No objections from me.

changelogs/unreleased/gh-9101-hide-error-unpack-redundant-fields.md
```diff
@@ -0,0 +1,5 @@
+## feature/box
+
+* Hide redundant fields from `box.error.unpack()` if
+  `box_error_unpack_type_and_code` compat option is set to 'new'. Default is
+  'old' behaviour. (gh-9101).
```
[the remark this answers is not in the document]
nshy: Fixed.

src/lua/error.lua
nshy: Fixed.

```diff
diff --git a/src/lua/error.lua b/src/lua/error.lua
index 01efbb96d..90e18a282 100644
--- a/src/lua/error.lua
+++ b/src/lua/error.lua
@@ -145,7 +145,7 @@ local function error_unpack(err)
         result[ffi.string(f._name)] = mp_decode(f._data)
     end
     -- Hide redundant fields from unpack output (gh-9101).
-    if compat.box_error_unpack_type_and_code.current == 'new' then
+    if compat.box_error_unpack_type_and_code:is_new() then
         result.custom_type = nil
         result.base_type = nil
         if result.code == 0 then
```

src/lua/error.lua
nshy: No, this one is independent. Instead the error serialization done in #9716 depends on this one as it uses `error:unpack()`.

ochaplashkin (approved):

src/lua/error.lua
Totktonada: That's what I meant. Thanks for clarification!

changelogs/unreleased/gh-9101-hide-error-unpack-redundant-fields.md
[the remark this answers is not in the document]
sergepetrenko: @nshy you might've forgotten to push your fixes.

src/lua/error.lua
sergepetrenko: This one is missing as well

PR #9773, opened 2024-03-05
config: set config status after "post_apply" phase

ImeevMA:
Previously, the config status was set after the "apply" phase, and then set again after the "post_apply" phase. This can lead to a situation where the status becomes "ready" after the "apply" phase and changes to "check_errors" or "check_warnings" after the "post_apply" phase. Because of this, it was possible that config:reload() may be called from a role or an app even if though the config loading is not finished.

Note that due to this change, the "startup_in_progress" and "reload_in_progress" statuses are now expanded to include the "post_apply" phase.

Also note that the status still changes before the "extras.post_apply" stage.

Needed for https://github.com/tarantool/tarantool-ee/issues/643

### DIFF
--- a/changelogs/unreleased/ghe-643-set-status-after-post-apply.md
+++ b/changelogs/unreleased/ghe-643-set-status-after-post-apply.md
@@ -1,0 +1,3 @@
+## bugfix/config
+
+* The config status is now set after the `post_apply` phase (ghe-643).
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -407,7 +407,6 @@ function methods._startup(self, instance_name, config_file)
     else
         self:_apply_on_startup({phase = 2})
     end
-    self:_set_status_based_on_alerts()
 
     self:_post_apply()
     self:_set_status_based_on_alerts()
@@ -449,7 +448,6 @@ function methods._reload_noexc(self, opts)
     local ok, err = pcall(function(opts)
         self:_store(self:_collect(opts))
         self:_apply()
-        self:_set_status_based_on_alerts()
         self:_post_apply()
     end, opts)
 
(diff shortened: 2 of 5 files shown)


ochaplashkin (approved):

changelogs/unreleased/ghe-643-set-status-after-post-apply.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
```
Totktonada, line 1: Nit: It looks more like a bugfix, isn't it?

Nit 2: AFAIR, our motivation to make this change is to use the config status to prevent reentering into the post_apply phase from the EE code. I think it is important to make it clear.

changelogs/unreleased/ghe-643-set-status-after-post-apply.md
```diff
@@ -0,0 +1,3 @@
...
+
+* The config status is now only set after the `post_apply` phase (ghe-643).
```
Totktonada, line 3: Are we need this change in `release/3.0`?

Totktonada (approved):

changelogs/unreleased/ghe-643-set-status-after-post-apply.md
p7nov:
```suggestion
* The config status is now set after the `post_apply` phase (ghe-643).
```

p7nov (approved):

changelogs/unreleased/ghe-643-set-status-after-post-apply.md
ImeevMA: > Are we need this change in `release/3.0`?

Since this is a bugfix, it makes sense to add this path to `release/3.0`.

@p7nov thanks for the suggestion. Applied.

Totktonada:
Backported to `release/3.0` (future `3.0.2`) in PR #9789.

PR #9776, opened 2024-03-05
config: introduce sharding.weight

### LINKED ISSUE #9775: Allow to set weight in sharding
The [vshard](https://github.com/tarantool/vshard) module has the weight replicaset-level option. We need to add it to the config options.

ImeevMA:
Closes #9775

@TarantoolBot document
Title: The `sharding.weight` option

The `sharding.weight` option can be set to facilitate rebalancing. The larger the weight of a replicaset, the more data it can store. This is a relative value, so a replicaset with a value N times larger than a value of the another replicaset can store N times more data. The default value is 1.

### DIFF
--- a/changelogs/unreleased/gh-9775-option-weight.md
+++ b/changelogs/unreleased/gh-9775-option-weight.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Introduced the `sharding.weight` configuration option, which reflects
+  the relative amount of data that a replicaset can store (gh-9775).
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -147,6 +147,10 @@ function methods.sharding(self)
     for group_name, group in pairs(self._cconfig.groups) do
         for replicaset_name, value in pairs(group.replicasets) do
             local lock
+            local weight
+            -- For replicaset-level options, we need to get them from the any
+            -- instance of the replicaset.
+            local is_rs_options_set = false
             local replicaset_uuid
             local replicaset_cfg = {}
             local is_rebalancer = nil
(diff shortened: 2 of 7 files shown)


src/box/lua/config/configdata.lua
```diff
@@ -147,6 +147,10 @@ function methods.sharding(self)
     for group_name, group in pairs(self._cconfig.groups) do
         for replicaset_name, value in pairs(group.replicasets) do
             local lock
+            local weight
+            -- For replicaset-level options, we need to get them from the first
+            -- instance of the replicaset.
+            local is_first = true
```
Totktonada, line 153: Nit: It sounds for me like we're obligated to acquire the values from the first instance (not any other), while the idea here is that we can acquire it from any instance of the replicaset (so, the check is needed only to skip unnecessary actions).

It confused me on the first reading: I had thoughts how the first instance is different and what is the order in which we traverse them here.

ImeevMA: Replaced by `is_rs_options_set`. Also replaced `first` in the comment by `any`.

Totktonada (approved):

changelogs/unreleased/gh-9775-option-weight.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* Introduced the `sharding.weight` option (gh-9775).
```
p7nov, line 3:
```suggestion
* Introduced the `sharding.weight` configuration option, which reflects
  the relative amount of data that a replicaset can store (gh-9775).
```

p7nov (approved):

changelogs/unreleased/gh-9775-option-weight.md
ImeevMA: Thanks, fixed!

ochaplashkin (approved):

PR #9783, opened 2024-03-07
config: expand config:info()

### LINKED ISSUE #9780: Show meta of active configuration
The meta is currently displayed for the last configuration load, even if was the failed reload. It would be helpful to show the meta for the active configuration as well.

ImeevMA:
Closes #9780

@TarantoolBot document
Title: Expansion of `config:info()`

Before this patch, the `meta` field of the `config:info()` function contained the meta data of the last load, even if it was unsuccessful reload. This can be inconvenient because if the reload fails, we expect that the last configuration that was successfully applied is used, but its metadata was lost.

Now `config:info()` can take one argument — the version of the information that should be returned by the function. If no arguments are specified or the argument is the string `v1`, the return value has the same structure as before. If the argument is the string `v2`, the `meta` field in the return value has fields - `active` (meta for the last successfully applied configuration) and `last` (meta of the last loaded configuration). Currently only these values are accepted.

### DIFF
--- a/changelogs/unreleased/gh-9780-show-active-meta.md
+++ b/changelogs/unreleased/gh-9780-show-active-meta.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* `config:info()` can now display source metadata from the last successful
+  load along with source metadata from the very last load attempt (gh-9780).
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -410,6 +410,9 @@ function methods._startup(self, instance_name, config_file)
 
     self:_post_apply()
     self:_set_status_based_on_alerts()
+    if self._status == 'ready' or self._status == 'check_warnings' then
+        self._active_metadata = self._metadata
+    end
     if extras ~= nil then
         extras.post_apply(self)
     end
@@ -457,6 +460,9 @@ function methods._reload_noexc(self, opts)
     end
 
     self:_set_status_based_on_alerts()
+    if self._status == 'ready' or self._status == 'check_warnings' then
+        self._active_metadata = self._metadata
+    end
     if extras ~= nil then
         extras.post_apply(self)
     end
@@ -472,13 +478,36 @@ function methods.reload(self)
     end
 end
 
-function methods.info(self)
+function methods.info(self, version)
     selfcheck(self, 'info')
-    return {
-        alerts = self._aboard:alerts(),
+    version = version == nil and 'v1' or version
+    local info = {}
+    local alerts = self._aboard:alerts()
+    info['v1'] = {
+        alerts = alerts,
         meta = self._metadata,
         status = self._status,
     }
+    info['v2'] = {
+        alerts = alerts,
+        meta = {
+            active = self._active_metadata,
+            last = self._metadata,
+        },
+        status = self._status,
+    }
+    if info[version] ~= nil then
+        return info[version]
+    end
+
+    local supported_versions = {}
+    for key in pairs(info) do
+        table.insert(supported_versions, key)
+    end
+    table.sort(supported_versions)
+
+    error(('config:info() expects %s or nil as an argument, got %q'):format(
+        table.concat(supported_versions, ', '), version), 0)
 end
 
 -- The object is a singleton. The constructor should be called
--- a/test/config-luatest/config_test.lua
+++ b/test/config-luatest/config_test.lua
@@ -1433,3 +1433,238 @@ g.test_post_apply_status = function(cg)
         verify_2 = verify_2,
     })
 end
+
+-- Check that config:info('v2') works properly.
+g.test_info_new_version = function(g)
+    local dir = treegen.prepare_directory(g, {}, {})
+
+    local config = [[
+        credentials:
+          users:
+            guest:
+              roles:
+              - super
+
+        iproto:
+          listen:
+            - uri: unix/:./{{ instance_name }}.iproto
+
+        groups:
+          group-001:
+            replicasets:
+              replicaset-001:
+                instances:
+                  instance-001: {}
+    ]]
+    treegen.write_script(dir, 'config.yaml', config)
+
+    -- New source that sets meta.
+    local script = string.dump(function()
+        local yaml = require('yaml')
+        local file = require('internal.config.utils.file')
+
+        local methods = {}
+        local mt = {
+            __index = methods,
+        }
+
+        function methods.sync(self, config)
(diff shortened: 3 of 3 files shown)


src/box/lua/config/init.lua
```diff
@@ -450,6 +453,9 @@ function methods._reload_noexc(self, opts)
         self:_store(self:_collect(opts))
         self:_apply()
         self:_set_status_based_on_alerts()
+        if self._status == 'ready' or self._status == 'check_warnings' then
+            self._active_metadata = self._metadata
+        end
```
Totktonada, line 458: I think we should do it after the post_apply phase (after roles and the app execution, but before extras.post_apply). This way if a role/app execution failed we don't lie about active configuration.

In fact, this situation means some intermediate active configuration, but I would expect that `active` will show the latest one that was fully and successfuly applied.

Please, hold this property with a test: a fail in role/app applying should lead to different `active` and `last` metadata.

changelogs/unreleased/gh-9780-show-active-meta.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* The `config:info()` functionality has been expanded (gh-9780).
```
Totktonada, line 3: Nit: As for me, a changelog entry assumes that we point a functionality itself, not just the fact that it is added. For example, 'added active configuration metainformation (usually a revision in a remote source)'.

src/box/lua/config/init.lua
```diff
@@ -474,13 +480,26 @@ function methods.reload(self)
...
+        return {
+            alerts = self._aboard:alerts(),
+            meta = {
+                active = self._active_metadata,
+                last = self._metadata,
+            },
+            status = self._status,
+        }
+    end
+    error(('Expected "v2" or nil as info version, got %q'):format(version), 0)
```
Totktonada, line 502: Nit: The error message is a bit inaccurate. I mean the following.

1. It doesn't mention 'v1'.
2. It requires updating with every version change (and it is easy to overlook).

I would update it this way:

```lua
local function set2string(s)
    local list = {}
    for k, _ in pairs(s) do
        table.insert(list)
    end 
    table.sort(list)
    return ('"%s"'):format(table.concat(list, '", "'))
end

function methods.info(self, version)
    selfcheck(self, 'info')
    version = version or 'v1'
    local supported_versions = { 
        v1 = true,
        v2 = true,
    }   
    if supported_versions[version] == nil then
        error(('config:info() expects nil or %s as an argument, got %q'):format(
            set2string(supported_versions), version), 0)
    end 
    <...>
end
```

Or don't mention versions at all (but this information looks useful).

Or just hardcode `nil, "v1" or "v2"` and try to remember to update it next time.

test/config-luatest/config_test.lua
```diff
@@ -1393,3 +1393,180 @@ g.test_non_dynamic_alert = function(g)
...
+        }
+
+        function methods.sync(self, config)
+            local data = file.universal_read('myconf.yaml', 'my config file')
+            local ok, res = pcall(yaml.decode, data)
+            if not ok then
+                error('Unable to parse my config file')
+            end
+            self._values = res
+            config:_meta('myconf', 'revision', self._revision)
```
Totktonada, line 1438: Nit: It is not important for testing, but if we take it as example of a configuration source with metainformation, I would use the same `:_meta()` key as the source name.

Totktonada (approved):
The patch looks OK for me except a few small comments. Please, take them into account.

src/box/lua/config/init.lua
ImeevMA: I am not sure about this. I mean, this way the `active` meta in roles and app will be from the last successful load, but `configdata` will be from the current load.

ochaplashkin (approved):

src/box/lua/config/init.lua
Totktonada: We do the same for statuses.

There is an alternative: we can set the active metadata after `apply`, but revert it back if an error occurs in `post_apply`.

However, I think that it would be too complicated, while it is not necessary that it was expected by a user. I would prefer a simple solution.

Since we provide both `last` and `active` I don't think that it worth to do the tricks like we do for `config:get()`.

ImeevMA: Agree. Fixed.

changelogs/unreleased/gh-9780-show-active-meta.md
ImeevMA: Fixed.

src/box/lua/config/init.lua
ImeevMA: Fixed, rewrote the code.

test/config-luatest/config_test.lua
ImeevMA: True, fixed.

src/box/lua/config/init.lua
Totktonada: Sorry for nitpicking, but I would highlight two tiny points.

1. `:alerts()` is called twice. It has a cycle inside and it seems easy to avoid it.
2. `pairs(info)` order is arbitrary, so I would add `table.sort(supported_versions)`.

test/config-luatest/config_test.lua
Totktonada: Sorry for nitpicking, but the source still named as `myfile` (in the `name` field), while placed into `mysource.lua` and reported as `mysource` in the metainformation.

src/box/lua/config/init.lua
ImeevMA: Thanks, fixed.

test/config-luatest/config_test.lua
ImeevMA: Sorry, fixed.

filonenko-mikhail (approved):

changelogs/unreleased/gh-9780-show-active-meta.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* The `config:info()` can now display source metadata from the last
+  successful load along with source metadata from the last load.
```
p7nov, line 4:
```suggestion

* `config:info()` can now display source metadata from the last successful
  load along with source metadata from the very last load attempt.
```

p7nov (approved):

changelogs/unreleased/gh-9780-show-active-meta.md
ImeevMA: Thank you! Applied.

PR #9791, opened 2024-03-11
luajit: bump new version

Buristan:
* Check frame size limit before returning to a lower frame.

NO_DOC=LuaJIT submodule bump
NO_TEST=LuaJIT submodule bump


Buristan:
The changeset is merged to master in scope of #9832, backported to 3.0 and 2.11 in scope of #9833 and #9834 respectively.

Issue #9049, opened 2023-08-28
`box.schema.downgrade('2.8.4')` doesn't work with replication


sergepetrenko:
Tarantool version: 3.0, 2.11.1-entrypoint-257-gb0324b5aa

Simply run 2 Tarantool instances
**How to reproduce:**
```lua
-- Instance 1
-- Step 1
box.cfg{listen=3301, replication={3301,3302}}
-- Step 3
box.schema.user.grant('guest', 'replication')
-- Step 4
box.schema.downgrade('2.8.4')

-- Instance 2
-- Step 2
box.cfg{listen = 3302, replication = {3301, 3302}, read_only=true}
-- Step 5
box.info.replication
```
After performing `box.schema.downgrade('2.8.4')` on the first node, the second one will report a replication error in the log:
```
2023-08-28 16:44:52.075 [65838] main/110/applier/3301 applier.cc:310 E> error applying row: {type: 'DELETE', replica_id: 1, lsn: 12, space_id: 288, index_id: 0, key: [341, 0]}
2023-08-28 16:44:52.075 [65838] main/110/applier/3301 I> can't read row
2023-08-28 16:44:52.075 [65838] main/110/applier/3301 alter.cc:2459 E> ER_LAST_DROP: Can't drop the primary key in a system space, space '_vspace_sequence'
```
Re-setting `box.cfg.replication` doesn't help, the issue stays the same:
```
tarantool> box.info.replication
---
- 1:
    id: 1
    uuid: ab3b4d9d-f075-485a-9bab-a966f5105a7c
    lsn: 11
    upstream:
      peer: '3301'
      lag: 281.11064696312
      status: stopped
      idle: 4.3063350000011
      message: Can't drop the primary key in a system space, space '_vspace_sequence'
    downstream:
      status: follow
      idle: 0.58552500000042
      vclock: {1: 145}
      lag: 0
  2:
    id: 2
    uuid: 1ffcc3cd-5e99-4d23-b1d0-35a4716dd956
    lsn: 0
...

```

PR #9794, opened 2024-03-12
box: fix DDL check for recovery

ImeevMA:
Commit https://github.com/tarantool/tarantool/commit/71de4b2c921777a4150bac36c8d9dea8bbb082c9 ("box: fix schema downgrade replication") introduces check that decides when DDL is disabled if current schema is not equal to required schema. This patch adds a condition for DDL operations to be enabled during recovery, so new bootstrap snapshot can be properly generated.

Follow-up https://github.com/tarantool/tarantool/issues/9049

Co-authored-by: Vladimir Davydov <vdavydov@tarantool.org>

### DIFF
--- a/src/box/schema.cc
+++ b/src/box/schema.cc
@@ -32,6 +32,7 @@
 #include "sequence.h"
 #include "assoc.h"
 #include "alter.h"
+#include "box.h"
 #include "scoped_guard.h"
 #include "user.h"
 #include "vclock/vclock.h"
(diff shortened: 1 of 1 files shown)


locker (approved):

src/box/schema.cc
locker, line 1: Bad commit message:
- Please reference a ticket that introduced the issue.
- The issue happens only during bootstrap, not recovery.
- Maybe, it's worth updating the comment.

locker:
Cherry-picked to 3.0.

PR #9801, opened 2024-03-13
test: fix `gh_8937_recover_wrong_hint_options_test.lua` snapshots

Gumix:
The test fails when Tarantool with a renewed schema version recovers from a `gh_8937_data/` snapshot,
because `luatest/server_instance.lua` executes:
```lua
box.schema.user.grant('guest', 'super', nil, nil, {if_not_exists = true})
```
while insert into system spaces is forbidden until schema is upgraded.
Fix it by adding `box.schema.user.grant()` to the snapshot.


nshy (approved):
LGTM.

ochaplashkin (approved):

PR #9804, opened 2024-03-13
box: add support for origin in _priv

ImeevMA:
This patch adds a new option '_origin' to the 'box.schema.user.grant()', 'box.schema.role.grant()', 'box.schema.user.revoke()' and 'box.schema.role.revoke()' functions. This is a private option that allows to differentiate privileges based on their origin. If this option is not provided an empty string is used by default.

Privileges that were granted using 'grant()' can be revoked using 'revoke()' with the same '_origin'.

The motivation for this change is to avoid actually revoking a privilege granted by two actors if one actor revokes it.

Needed for https://github.com/tarantool/tarantool/issues/9643

### DIFF
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -16,6 +16,11 @@ local check_param = utils.check_param
 local check_param_table = utils.check_param_table
 local update_param_table = utils.update_param_table
 
+local DEFAULT_ORIGIN = ''
+-- We use the field ID instead of the field name so we don't need to upgrade
+-- the schema to use _origins option.
+local PRIV_OPTS_FIELD_ID = 6
+
 local function setmap(table)
     return setmetatable(table, { __serialize = 'map' })
 end
@@ -3427,6 +3432,56 @@ box.schema.user.exists = function(name)
     end
 end
 
+-- Return expanded origins from the given tuple.
+local function origins_from_tuple(tuple)
+    if tuple == nil then
+        return {[DEFAULT_ORIGIN] = 0}
+    end
+    if tuple[PRIV_OPTS_FIELD_ID] == nil or
+       tuple[PRIV_OPTS_FIELD_ID].origins == nil then
+        return {[DEFAULT_ORIGIN] = tuple.privilege}
+    end
+    return tuple[PRIV_OPTS_FIELD_ID].origins
+end
+
+-- Return total privileges from the given origins.
+local function privilege_from_origins(origins)
+    local new_privilege = 0
+    for _, privilege in pairs(origins) do
+        new_privilege = bit.bor(new_privilege, privilege)
+    end
+    return new_privilege
+end
+
+-- Return resulting opts from the given origins.
+local function opts_from_origins(origins)
+    local normalized_origins = {}
+    for name, privilege in pairs(origins) do
+        if privilege ~= 0 then
+            normalized_origins[name] = privilege
+        end
+    end
+    assert(next(normalized_origins) ~= nil)
+    -- If only default origin present, we do not need opts.
+    if normalized_origins[DEFAULT_ORIGIN] ~= nil and
+       next(normalized_origins, next(normalized_origins)) == nil then
+        return nil
+    end
+    return {origins = normalized_origins}
+end
+
+local function grant_error(name, object_name, object_type, privilege)
+    if object_type == 'role' and object_name ~= '' and
+       privilege == 'execute' then
+        box.error(box.error.ROLE_GRANTED, name, object_name)
+    end
+    if not is_singleton_object_type(object_type) then
+        object_name = string.format(" '%s'", object_name)
+    end
+    box.error(box.error.PRIV_GRANTED, name, privilege, object_type,
+              object_name)
+end
+
 local function grant(uid, name, privilege, object_type,
                      object_name, options)
     -- From user point of view, role is the same thing
@@ -3453,35 +3508,77 @@ local function grant(uid, name, privilege, object_type,
     else
         options.grantor = user_or_role_resolve(options.grantor)
     end
+    if options._origin == nil then
+        options._origin = DEFAULT_ORIGIN
+    elseif type(options._origin) ~= 'string' then
+        box.error(box.error.ILLEGAL_PARAMS, "options parameter '_origin' " ..
+                  "should be of type 'string'")
+    end
     local _priv = box.space[box.schema.PRIV_ID]
     local _vpriv = box.space[box.schema.VPRIV_ID]
     -- add the granted privilege to the current set
-    local tuple = _vpriv:get{uid, object_type, oid}
-    local old_privilege
-    if tuple ~= nil then
-        old_privilege = tuple.privilege
-    else
-        old_privilege = 0
-    end
-    privilege_hex = bit.bor(privilege_hex, old_privilege)
-    privilege_hex = tonumber(ffi.cast('uint32_t', privilege_hex))
+    local origins = origins_from_tuple(_vpriv:get({uid, object_type, oid}))
+
     -- do not execute a replace if it does not change anything
     -- XXX bug if we decide to add a grant option: new grantor
     -- replaces the old one, old grantor is lost
-    if privilege_hex ~= old_privilege then
-        _priv:replace{options.grantor, uid, object_type, oid, privilege_hex}
-    elseif not options.if_not_exists then
-            if object_type == 'role' and object_name ~= '' and
-               privilege == 'execute' then
-                box.error(box.error.ROLE_GRANTED, name, object_name)
-            else
-                if not is_singleton_object_type(object_type) then
-                    object_name = string.format(" '%s'", object_name)
-                end
-                box.error(box.error.PRIV_GRANTED, name, privilege,
-                          object_type, object_name)
-            end
+    local old_privilege = origins[options._origin] or 0
+    if bit.band(bit.bnot(old_privilege), privilege_hex) ~= 0 then
+        -- Update given privileges by origin.
+        origins[options._origin] = bit.bor(old_privilege, privilege_hex)
+    else
+        -- No new privileges can be given.
+        if options.if_not_exists then
+            return
+        end
+        grant_error(name, object_name, object_type, privilege)
     end
+
+    local new_privilege = privilege_from_origins(origins)
+    local opts = opts_from_origins(origins)
+    _priv:replace({options.grantor, uid, object_type, oid, new_privilege, opts})
+end
+
+local function revoke_error(name, object_name, object_type, privilege, origin,
+                            tuple)
+    local total_privileges = tuple == nil and 0 or tuple.privilege
+    local new_privilege = privilege_check(privilege, object_type)
+    local prev_origin
+    local reason
+    local code
+    if bit.band(new_privilege, total_privileges) == 0 then
+        prev_origin = nil
+    else
+        prev_origin = origin == '' and 'default' or origin
+    end
+
+    if object_type == 'role' and object_name ~= '' and
+       privilege == 'execute' then
+        local msg
+        code = box.error.ROLE_NOT_GRANTED
+        if prev_origin == nil then
+            msg = "User '%s' does not have role '%s'"
+        else
+            msg = "User '%s' does not have role '%s' provided by %s origin"
+        end
+        reason = msg:format(name, object_name, prev_origin)
+    else
+        local msg
+        code = box.error.PRIV_NOT_GRANTED
+        if prev_origin == nil then
+            msg = "User '%s' does not have %s access on %s '%s'"
+        else
+            msg = "User '%s' does not have %s access on %s '%s' provided by " ..
+                  "%s origin"
+        end
+        reason = msg:format(name, privilege, object_type, object_name,
+                            prev_origin)
+    end
+    box.error({
+        code = code,
+        reason = reason,
+        prev_origin = prev_origin,
+    })
 end
 
 local function revoke(uid, name, privilege, object_type, object_name, options)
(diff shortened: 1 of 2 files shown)


src/box/lua/schema.lua
```diff
@@ -3453,35 +3453,64 @@ local function grant(uid, name, privilege, object_type,
...
+        end
+        if object_type == 'role' and object_name ~= '' and
+           privilege == 'execute' then
+            box.error(box.error.ROLE_GRANTED, name, object_name)
+        end
+        if not is_singleton_object_type(object_type) then
+            object_name = string.format(" '%s'", object_name)
+        end
+        box.error(box.error.PRIV_GRANTED, name, privilege, object_type,
+                  object_name)
```
Totktonada, line 3496: The error message now can be confusing. Let's look on your example from the documentation request.

```lua
box.schema.role.create('a')
box.schema.role.grant('a', 'read, write', 'space', '_space', {origin = 'one'})
box.schema.role.grant('a', 'read', 'space', '_space', {origin = 'two'})
box.schema.role.revoke('a', 'write', 'space', '_space', {origin = 'two'})
-- error: User 'a' does not have write access on space '_space'
```

The role 'a' has the `write` privilege on the space `_space`. It has no `write` privilege on this space granted by `two`.

I would also appreciate if this information will be presented in `box.schema.{user,role}.info()`. Otherwise it may be hard to debug what is going on. However, I'm OK if the `.info()` adjustment will be implemented in a separate patchset.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
```diff
@@ -0,0 +1,6 @@
+## feature/box
```
Totktonada, line 1: It is not easy to deduce a motivation behind this feature from the documentation request. You're saying the following.

> This option allows to differentiate privileges based on their origin. Moreover, if a privilege was granted using `box.schema.*.grant()` with an origin, it can only be revoked using `box.schema.*.revoke()` with the same origin. Also, now the same privilege with different origin can be applied.

Our motivation is to don't revoke a privilege granted by two actors if one actor revokes it. This simple idea can be expressed in one sentence, but it is the most important thing to say about the feature, I think.

src/box/lua/schema.lua
```diff
@@ -3500,12 +3529,29 @@ local function revoke(uid, name, privilege, object_type, object_name, options)
     end
     local privilege_hex = privilege_check(privilege, object_type)
     options = options or {}
+    if options.origin == nil then
+        options.origin = ''
+    elseif type(options.origin) ~= 'string' or #options.origin == 0 then
```
Totktonada, line 3534: I think it is OK to allow an explicit `origin = ''`. A user may see it in the `_priv` space or in the `.info()` endpoint. If would be logical to expect that it is accepted.

Totktonada: BTW, I would use `DEFAULT_ORIGIN = ''` constant or something of this kind across the code.

src/box/lua/schema.lua
```diff
@@ -3500,12 +3529,29 @@ local function revoke(uid, name, privilege, object_type, object_name, options)
...
     local oid = object_resolve(object_type, object_name)
     local _priv = box.space[box.schema.PRIV_ID]
     local _vpriv = box.space[box.schema.VPRIV_ID]
     local tuple = _vpriv:get{uid, object_type, oid}
     -- system privileges of admin and guest can't be revoked
+
+    local origins
     if tuple == nil then
+        origins = {[''] = 0}
+    elseif tuple[6] == nil or tuple[6].origins == nil then
```
Totktonada, line 3547: I vote for using a constant like `PRIV_OPTS_FIELD_ID = 6` constant or simething of this kind.

I also think that it is important to leave a comment that a field id is used instead of a field name deliberately: we don't want to require a database schema upgrade to use the `origin` option.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
Totktonada, line 1: Nit: The commit message and the pull request title has the `config: ` prefix, while the change itself doesn't touch or affect any `config` module code.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
```diff
@@ -0,0 +1,6 @@
...
+
+* Introfuced new field `origin` to `option` argument of
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
ImeevMA: For now I left a simple commit-message. I'll rewrite the doc-request in a commit about `info()`.

src/box/lua/schema.lua
ImeevMA: Thanks, fixed.

src/box/lua/schema.lua
ImeevMA: Thanks. fixed.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
ImeevMA: Changed to `box:`.

src/box/lua/schema.lua
ImeevMA: It was decided to not make this feature public, so it will not be added to `info()`.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
ImeevMA: It was decided to not make this feature public, so doc-request in not needed.

src/box/lua/schema.lua
Totktonada: > I also think that it is important to leave a comment that a field id is used instead of a field name deliberately: we don't want to require a database schema upgrade to use the `origin` option.

How about this?

src/box/lua/schema.lua
Totktonada: I still think that the error may be very confusing: there is a privilege, it works, but an attempt to revoke it says that there is no such a privilege.

We discussed it with Mergen in a chat and we agreed on the following solution.

1. A check against `tuple.prigileve` is performed first. It issues an error with the old message if there is no needed privilege.
2. Then, a check against a given `tuple[6].origins[opts._origin]` is performed. If there is no needed privilege, an error says that it is missed *in the given origin*.

changelogs/unreleased/gh-9643-origin-in-priv-options.md
Totktonada: > motivation is to don't revoke a privilege granted by two actors if one actor revokes it

I think that it is important point and it makes things more clear, at least for me.

BTW, we can say it a bit better:

> motivation is to don't *actually* revoke a privilege granted by two actors if one actor revokes it

ImeevMA: Fixed.

src/box/lua/schema.lua
ImeevMA: Fixed.

src/box/lua/schema.lua
```diff
@@ -3500,42 +3563,63 @@ local function revoke(uid, name, privilege, object_type, object_name, options)
...
+
+    local origins
     if tuple == nil then
+        origins = {[DEFAULT_ORIGIN] = 0}
+    elseif tuple[PRIV_OPTS_FIELD_ID] == nil or
+           tuple[PRIV_OPTS_FIELD_ID].origins == nil then
+        origins = {[DEFAULT_ORIGIN] = tuple.privilege}
+    else
+        origins = tuple[PRIV_OPTS_FIELD_ID].origins
+    end
```
Totktonada, line 3586: We have such a code as in `grant` as well as in `revoke`. Can we extract it into a function like `origins_from_tuple` or `privilege_origins_from_tuple`?

I would also leave a comment that the function returns the origins in a full form (as a contrary to the normalized form stored in the _priv space) to ease further processing.

src/box/lua/schema.lua
```diff
@@ -3453,35 +3470,81 @@ local function grant(uid, name, privilege, object_type,
...
+        new_privilege = bit.bor(new_privilege, origin_privilege)
+        if origin_privilege == 0 then
+            origins[origin_name] = nil
+        end
     end
+    local res = {options.grantor, uid, object_type, oid, new_privilege}
+    -- If only default origin present, we do not need opts.
+    assert(next(origins) ~= nil)
+    if origins[DEFAULT_ORIGIN] == nil or
+       next(origins, next(origins)) ~= nil then
```
Totktonada, line 3521: We have such a code as in `grant` as well as in `revoke`. Can we extract it into its own functions?

For example, `privilege_from_origins(origins)` and `normalize_origins(origins)`. The latter should strip origins with zero privileges and should return `nil` if origins shouldn't be stored (equals to `{[''] = new_privilege}` or just has only `''` key).

src/box/lua/schema.lua
Totktonada: The new error messages were added. Thanks!

Totktonada (approved):

src/box/lua/schema.lua
ImeevMA: Done.

src/box/lua/schema.lua
ImeevMA: Done. Introduced two new functions: `privilege_from_origins()` and `opts_from_origins()`.

src/box/lua/schema.lua
Totktonada: Thanks!

src/box/lua/schema.lua
```diff
@@ -16,6 +16,11 @@ local check_param = utils.check_param
 local check_param_table = utils.check_param_table
 local update_param_table = utils.update_param_table
 
+local DEFAULT_ORIGIN = ''
+-- We use the field ID instead of the field name so we don't need to upgrade
+-- the schema to use _origins option.
+local PRIV_OPTS_FIELD_ID = 6
+
```
locker, line 23: We use a strict schema for system tables so I think we do need upgrade. The upgrade function should set the format for the new field and add an empty map to each existing privilege. Please don't forget to implement and test downgrade as well.

src/box/errcode.h
```diff
@@ -395,6 +395,8 @@ struct errcode_record {
 	_(ER_IN_ANOTHER_PROMOTE, 278,		"box.ctl.promote() is already running") \
 	_(ER_SHUTDOWN, 279,			"Server is shutting down") \
 	_(ER_FIELD_VALUE_OUT_OF_RANGE, 280,	"The value of field %s exceeds the supported range for type '%s': expected [%s..%s], got %s") \
+	_(ER_PRIV_NOT_GRANTED_BY_ORIGIN, 281,	"%s '%s' does not have %s access on %s '%s' provided by %s origin") \
+	_(ER_ROLE_NOT_GRANTED_BY_ORIGIN, 282,	"%s '%s' does not have role '%s' provided by %s origin") \
```
locker, line 399: Please use payload (see `TEST_ERROR_CODES` and/or ask @nshy for an example).

src/box/lua/schema.lua
```diff
@@ -3427,6 +3432,55 @@ box.schema.user.exists = function(name)
...
+        new_privilege = bit.bor(new_privilege, privilege)
+    end
+    return new_privilege
+end
+
+-- Return retulting opts from the given origins.
+local function opts_from_origins(origins)
+    for name, privilege in pairs(origins) do
+        if privilege == 0 then
+            origins[name] = nil
```
locker, line 3460: Better avoid modifying function arguments - this is confusing and error-prone.

src/box/lua/schema.lua
```diff
@@ -3453,35 +3507,58 @@ local function grant(uid, name, privilege, object_type,
...
+        end
+        box.error(box.error.ROLE_NOT_GRANTED_BY_ORIGIN, grantee_type, name,
+                  object_name, origin_name)
+    end
+    if bit.band(new_privilege, total_privileges) == 0 then
+        box.error(box.error.PRIV_NOT_GRANTED, name, privilege, object_type,
+                  object_name)
     end
+    box.error(box.error.PRIV_NOT_GRANTED_BY_ORIGIN, grantee_type, name,
+              privilege, object_type, object_name, origin_name)
```
locker, line 3561: It may be better to add a new payload parameter to the existing errors `PRIV_NOT_GRANTED` and `ROLE_NOT_GRANTED` instead of introducing the new error codes.

test/box-luatest/gh_9643_priv_origins_test.lua
```diff
@@ -0,0 +1,182 @@
...
+        }}
+        revoke('myrole', 'somerole')
+        t.assert_equals(_priv:get({rid, 'role', tid}), exp)
+
+        ok, err = pcall(revoke, 'myrole', 'somerole')
+        exp_err = "Role 'myrole' does not have role 'somerole' provided by " ..
+                  "default origin"
+        t.assert(not ok)
+        t.assert_equals(err.message, exp_err)
+        t.assert_equals(_priv:get({rid, 'role', tid}), exp)
```
locker, line 180: Please also test that `if_exists` and `if_not_exists` options work with origins.

test/box-luatest/gh_9643_priv_origins_test.lua
```diff
@@ -0,0 +1,182 @@
...
+        -- Check grant with origins.
+        exp[5] = box.priv.X
+        exp[6] = nil
+        grant('myrole', 'somerole', nil, nil)
+        t.assert_equals(_priv:get({rid, 'role', tid}), exp)
+
+        exp[6] = {origins = {
+            [''] = box.priv.X,
+            two = box.priv.X,
+        }}
```
locker, line 152: Better write the test so that it doesn't break if we add new priv options.

locker (changes requested):

src/box/lua/schema.lua
ImeevMA: Ok, I will do in in the separate PR.

Totktonada: I think that the new logic shouldn't depend on the upgrade to simplify things for a user.

However, the upgrade is a good thing to assign a name for the 6th `_priv` field for better introspection.

src/box/errcode.h
Totktonada: AFAIR, it will not be shown (almost) anywhere while the `box_error_serialize_verbose` compat option remains `old`. But IMHO the old message is just incorrect in the case. It states that there are no privilege, when it is present (just granted by another origin).

src/box/lua/schema.lua
ImeevMA: For now I suggest to leave upgrade for another PR, if we decide to execute it.

src/box/errcode.h
ImeevMA: In the new version we use a special `box.error()` syntax that allows to use a known error code but a different error message.

src/box/lua/schema.lua
ImeevMA: Thanks, fixed.

test/box-luatest/gh_9643_priv_origins_test.lua
ImeevMA: Thanks, added a few tests.

test/box-luatest/gh_9643_priv_origins_test.lua
ImeevMA: True, fixed.

locker (approved):

ylobankov (approved):

Totktonada:
Backported to `release/3.0` (future `3.0.2`) in PR #9840.

ImeevMA:
@Totktonada @locker created a new [issue](https://github.com/tarantool/tarantool/issues/9841) about the field name of the new field in `_priv`.

PR #9808, opened 2024-03-13
config: access configuration of other cluster members

### LINKED ISSUE #9680: Allow to retrieve cluster config with `config` module
The `config` module allows to get the current instance config. However, it's impossible to get the config of the cluster this instance is a part of. This would be really useful for implementing application sharding: an instance representing a router needs to know the list of all replica-sets and instances representing data shards so that the application developer wouldn't need to specify all the shard uris in the role config. Please extend the `config` module with this functionality. The `apply` method of the role should be able to fetch the actual (applied) configuration of the whole cluster.

Totktonada:
## `config:instances()`

List all instances of the cluster.

Returns a table of the following format.

```lua
{
    [<instance_name>] = {
        instance_name = <...>,
        replicaset_name = <...>,
        group_name = <...>,
    },
    <...>
}
```

If an action should be performed for each instance of the given cluster, it can be written this way:

```lua
for instance_name in pairs(config:instances()) do
    action(instance_name)
end
```

If replicaset/group names matter, the instances may be filtered this way:

```lua
-- Perform an action for all instances of the given replicaset.
for instance_name, def in pairs(config:instances()) do
    if def.replicaset_name == box.info.replicaset.name then
        action(instance_name)
    end
end
```

## `config:get(<...>, {instance = '<...>'})`

The new `instance` option of the `config:get()` method allows to retrieve a configuration value of another instance from the given cluster.

Example:

```lua
-- Collect all the enabled roles within the cluster.
local enabled_roles = {}
for instance_name in pairs(config:instances()) do
    local roles = config:get('roles', {instance = instance_name})
    for _, role in ipairs(roles) do
        enabled_roles[role] = true
    end
end
```

Note: There is a difference between a missing `instance` option and the `instance` option that is equal to the given instance name. The former returns an instance configuration taking into account instance configuration sources (environment variables). The latter takes into account only cluster configuration, so the environment variables are ignored.

Fixes #9680

### DIFF
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -11,16 +11,16 @@ local cluster_config = require('internal.config.cluster_config')
 local snapshot = require('internal.config.utils.snapshot')
 
 local function choose_iconfig(self, opts)
-    if opts ~= nil and opts.peer ~= nil then
-        local peers = self._peers
-        local peer = peers[opts.peer]
-        if peer == nil then
-            error(('Unknown peer %q'):format(opts.peer), 0)
+    if opts ~= nil and opts.instance ~= nil then
+        local instances = self._instances
+        local instance = instances[opts.instance]
+        if instance == nil then
+            error(('Unknown instance %q'):format(opts.instance), 0)
         end
         if opts ~= nil and opts.use_default then
-            return peer.iconfig_def
+            return instance.iconfig_def
         end
-        return peer.iconfig
+        return instance.iconfig
     end
 
     if opts ~= nil and opts.use_default then
@@ -36,7 +36,7 @@ local methods = {}
 --
 -- opts:
 --     use_default: boolean
---     peer: string
+--     instance: string
 function methods.get(self, path, opts)
     local data = choose_iconfig(self, opts)
     return instance_config:get(data, path)
@@ -46,7 +46,7 @@ end
 --
 -- opts:
 --     use_default: boolean
---     peer: string
+--     instance: string
 function methods.filter(self, f, opts)
     local data = choose_iconfig(self, opts)
     return instance_config:filter(data, f)
@@ -70,8 +70,14 @@ function methods.names(self)
     }
 end
 
-local function instance_sharding(iconfig, instance_name)
-    local roles = instance_config:get(iconfig, 'sharding.roles')
+-- Generate a part of a vshard configuration that relates to
+-- a particular instance.
+--
+-- opts:
+--     use_default: boolean
+--     instance: string
+function methods._instance_sharding(self, opts)
+    local roles = self:get('sharding.roles', opts)
     if roles == nil or #roles == 0 then
         return nil
     end
@@ -83,11 +89,12 @@ local function instance_sharding(iconfig, instance_name)
     if not is_storage then
         return nil
     end
-    local zone = instance_config:get(iconfig, 'sharding.zone')
-    local uri = instance_config:instance_uri(iconfig, 'sharding')
+    local zone = self:get('sharding.zone', opts)
+    local uri = instance_config:instance_uri(choose_iconfig(self, opts),
+        'sharding')
     if uri == nil then
         local err = 'No suitable URI provided for instance %q'
-        error(err:format(instance_name), 0)
+        error(err:format(opts.instance), 0)
     end
     --
     -- Currently, vshard does not accept URI without a username. So if we got a
@@ -100,9 +107,9 @@ local function instance_sharding(iconfig, instance_name)
         u.login = 'guest'
         uri = urilib.format(u, true)
     end
-    local uuid = instance_config:get(iconfig, 'database.instance_uuid')
+    local uuid = self:get('database.instance_uuid', opts)
 
-    local user = instance_config:get(iconfig, {'credentials', 'users', u.login})
+    local user = self:get({'credentials', 'users', u.login}, opts)
     --
     -- If the user is not described in the credentials, this may mean that the
     -- user already exists and may have all the necessary privileges. If not, an
@@ -121,7 +128,7 @@ local function instance_sharding(iconfig, instance_name)
             end
             for _, role_name in pairs(roles) do
                 local path = {'credentials', 'roles', role_name, 'roles'}
-                if check_sharding_role(instance_config:get(iconfig, path)) then
+                if check_sharding_role(self:get(path, opts)) then
                     return true
                 end
             end
@@ -144,7 +151,7 @@ end
 function methods.sharding(self)
     local sharding = {}
     local rebalancers = {}
-    for group_name, group in pairs(self._cconfig.groups) do
+    for _, group in pairs(self._cconfig.groups) do
         for replicaset_name, value in pairs(group.replicasets) do
             local lock
             local weight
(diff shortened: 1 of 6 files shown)


src/box/lua/config/init.lua
```diff
@@ -420,13 +430,22 @@ function methods._print_env_list(self)
     io.stdout:write(env_source:_env_list())
 end
 
-function methods.get(self, path)
+-- Note: There is a difference between missing opts.instance and
+-- opts.instance that is equal to the given instance name. The
+-- former returns an instance configuration taking into account
+-- instance configuration sources (environment variables). The
+-- latter takes into account only cluster configuration, so the
+-- environment variables are ignored.
```
Totktonada, line 438: @locker Please, let me know, whether this nuance fits good into your scenarios.

I feel it as a natural behavior, but it is possible to compare the given instance name against the name of the running instance and return the same value as if `opts.instance` is omitted.

src/box/lua/config/init.lua
```diff
@@ -481,6 +500,55 @@ function methods.info(self)
...
+-- Returns a table of the following format.
+--
+-- {
+--     [<instance_name>] = {
+--         instance_name = <...>,
+--         replicaset_name = <...>,
+--         group_name = <...>,
+--     },
+--     <...>
+-- }
```
Totktonada, line 529: @locker I have changed the shape of the return value a bit comparing to the workaround I've provided in #9680. The keys are instance names now.

It feels natural, because the order is anyway unspecified. It also allows easier access to the instance name.

Before (however, it works now too):

```lua
for _, def in pairs(config:instances())
    action(def.instance_name)
end
```

Now:

```lua
for instance_name in pairs(config:instances())
    action(instance_name)
end
```

If you're using `ipairs()` for traversing over `config:instance()`, it doesn't work after the change.

What do you think about this change?

locker: This change looks good to me (although I do use `ipairs`).

src/box/lua/config/init.lua
locker: AFAIU, when specifying `opts.instance` different from self, we can't possibly get env. Making this behavior (ignoring env) for every value of `opts.instance` (including self) looks sane. Including env if `opts.instance` is omitted is probably fine although having a separate explicit option for that could be more straightforward.

src/box/lua/config/configdata.lua
```diff
@@ -364,6 +424,18 @@ function methods.missing_names(self)
...
 
+-- Cluster configuration.
+--
+-- It is given as is after merging from all the configuration
+-- sources. Default values are NOT applied. Variables are NOT
+-- substituted.
+--
+-- Use :get() to receive an instance config for a particular
+-- instance with applied defaults and substituted variables.
+function methods.cconfig(self)
```
locker, line 435: Do we really need to introduce this method now, when we have `get` and `instances`? If we do, I'd name it differently. Maybe, `raw_config` to emphasize that the config is returned as is (without applying defaults and substituting variables)?

Still, I'd probably avoid introducing this method now because it looks controversial and isn't really needed AFAIU.

src/box/lua/config/init.lua
Totktonada: Thank you for the feedback!

I've filed #9824 about the explicit option that controls, whether to take env into account. I'm not sure, how the API should look, so I want to discuss/postpone it.

src/box/lua/config/configdata.lua
Totktonada: It gives me a confidence that if I've missed some usage scenario, a user is able to acquire a raw cluster configuration and traverse it on its own.

However, this feature was not asked explicitly and I've no any usage scenario that wouldn't be covered by `:instances()` + `:get()`.

I agree to hide it, at least for now.

----

I would leave some words regarding possible representations of the cluster configuration.

Substituting variables is not possible for a cluster configuration in a general case. A replicaset level option may have `{{ instance_name }}`. The substitution has sense only in context of a particular instance.

Applying defaults on the whole cluster configuration would break ability to calculate an effective option for a particular instance. If an option is set only on the replicaset level, applying defaults to the instance level would give a higher priority to the default.

OTOH, we can apply defaults only on the global level. I'm not sure that I like it, but I don't see how it may break anything.

----

Regarding the name of the method: I like such concise names. 
`iconfig` and `cconfig` are used internally across the code, they look like a one word. Maybe it is not fair and I should use two words instead of constructing a domain specific pseudo-word. I don't know how to deduce a best general rule.

Maybe it should be `:get(nil, {cluster = true})`.

Totktonada:
@ImeevMA Thank you for the review and the questions!

I would answer them in the reverse order.

Q:

> Also, I find the result of `config:instances()` a little strange. I would expect either an array with instance names or a map with the full instance configurations.

<details>
<summary>A:</summary>

A list of instance names would be the simplest shape of the result for `config:instances()`.

```lua
-- returns [foo, bar, baz]
config:instances()
```

However, it gives too little information. A consumer of a cluster configuration likely needs to group information about instances per replicaset. At least, I do (in failover) and Vladimir does (in aeon).

Taking into account the assumption that just instance names are not very useful, we should mark each of the instances with its replicaset name.

```lua
-- returns [
--     {
--         instance_name = 'foo',
--         replicaset_name = 'rs',
--     },
--     <...>
-- ]
config:instances()
```

We can group it per replicaset.

```lua
-- Alternarive shape.
--
-- returns {
--     rs = [foo, bar, baz],
--     <...>
-- ]
config:replicasets()
```

Also, since we have the groups as part of the taxonomy, it seems meaningful to add them too.

```lua
-- returns [
--     {
--         instance_name = 'foo',
--         replicaset_name = 'rs',
--         group_name = 'g',
--     },
--     <...>
-- ]
config:instances()
```

```lua
-- Alternarive shape.
--
-- returns {
--     rs = [
--         {
--             instance_name = 'foo',
--             group_name = 'g',
--         },
--         <...>
--     ],
--     <...>
-- ]
config:replicasets()

-- Alternarive shape.
--
-- returns {
--     g = {
--         rs = [foo, bar, baz],
--         <...>
--     },
--     <...>
-- ]
config:groups()
```

When we're looking into the alternative shape taking the groups into account, it doesn't look so concise. It needs two or three nested loops to traverse and it doesn't add much to the pure cluster configuration (which needs three nested loops to traverse).

We need something that would be easy to traverse.

So, we have basically a list of instance definitions with instance/replicaset/group names.

I've defined it as a list initially, but than I [found](https://github.com/tarantool/tarantool/pull/9808#discussion_r1523966063) that saying 'it has indices, but they have no sense' looks less obvious than saying 'the keys are instance names, we've no any order here'.

Finally, we have the following structure that describes a full topology and, at the same time, can be traversed using one loop.

```lua
-- returns {
--     foo = {
--         instance_name = 'foo',
--         replicaset_name = 'rs',
--         group_name = 'g',
--     }
--     <...>
-- }
config:instances()
```
</details>

Q:

> <...> what about adding `group` and `replicaset` to the instance config? What I mean is that each instance has defined `group_name` and `replicaset_name`, but we cannot get them using `config:get()`.

<details>
<summary>A:</summary>

All the topology information is available from `config:instances()`. Each instance definition table has the `replicaset_name` and `group_name` fields.

So, we have one method to acquire a topology and another to acquire an instance configuration.

Let's consider what I can do using your proposal. I can acquire a replicaset name and a group name using `config:get()`.

```lua
config:get('replicaset_name', {instance = 'foo'})
config:get('group_name', {instance = 'foo'})
```

But I can do it now using `config:instances()`.

```lua
config:instances().foo.replicaset_name
config:instances().foo.group_name
```

The separation between the topology and the instance configuration feels natural for me and I see no reason to mix these things together in the `config:get()` method.

However, I'm a bit worrying that we can't acquire instance/replicaset/group names of our current instance before the instance name is persisted and appears in `box.info.name`. OTOH, nobody complained and I consider that it is either not needed in practice or `box.info.name` appears early enough for practical scenarios.
</details>

xuniq (approved):

ochaplashkin (approved):

Issue #9108, opened 2023-09-07
Replace ClientError format arguments with payload


locker:
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#f9d5980310a143a6b0ffff5e6bed24c5)

For most [`ClientError`](https://github.com/tarantool/tarantool/blob/1d6043fac53bebd539ee5882f6073767eea753bd/src/box/error.h#L200) errors we include extra error information in the error message. For this purpose we define special error message formats in [`src/box/errcode.h`](https://github.com/tarantool/tarantool/blob/master/src/box/errcode.h). For example, [`ER_ALTER_SPACE`](https://github.com/tarantool/tarantool/blob/00a9e59927399c91158aa2bf9698c4bfa6e11322/src/box/errcode.h#L67) is defined in the core Tarantool code as

```c
/* 12 */_(ER_ALTER_SPACE,		"Can't modify space '%s': %s") \
```

and raised using

```c
diag_set(ClientError, ER_ALTER_SPACE, space_name(old_space),
				 "replication group is immutable");
```

Although this is usually good enough for users, errors composed this way can’t be parsed automatically. Also, it isn’t always possible to include all the information in the message.

The suggestion is to replace all `ClientError` error message formats with plain constant string messages and store all extra information in error payload. We’d thus define `ER_ALTER_SPACE` as

```c
/* 12 */_(ER_ALTER_SPACE,		"Can't modify space") \
```

and raise it using

```c
struct error *e = diag_set(ClientError, ER_ALTER_SPACE);
error_set_str(e, "space", space_name(old_space));
error_set_str(e, "reason", "replication group is immutable");
```

Note:

- This is just an illustration. We’ll probably set the reason using stacked diagnostic instead.
- Setting payload may look verbose but we’re planning to add helpers for that, see #9109.

Since a user may create an error object in Lua with a system error code passing format arguments to [`box.error`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/error/), this means that this change will have user-visible effects, for example if an error was created with the following Lua command:

```lua
box.error.new(box.error.UNSUPPORTED, 'foo', 'bar')
```

it’d have “foo does not support bar” error message before this change and “Feature is unsupported” after this change (message format arguments would get lost). The user will have to add extra error information as payload fields, for example:

```lua
[...]

PR #9810, opened 2024-03-14
sql: rework OP_SetDiag opcode

ImeevMA:
Prior to this patch, the OP_SetDiag opcode accepted an error code and error description, which already included the description from the error code. But this makes #9108 more difficult, so now the accepted error description does not include the description from the error code.

Note that this change is possible because all accepted error codes use exactly one argument.

Additionally, this patch removes p4 from the OP_Halt opcode in one place, since OP_Halt only uses p1 and p2 and does not use p4. Moreover, the address specified in p4 may not be available by the time OP_Halt is executed.

Needed for #9108

### DIFF
--- a/src/box/sql/select.c
+++ b/src/box/sql/select.c
@@ -2276,9 +2276,8 @@ computeLimitRegisters(Parse * pParse, Select * p, int iBreak)
 		sqlVdbeAddOp2(v, OP_Integer, 0, r1);
 		sqlVdbeAddOp3(v, OP_Ge, r1, positive_limit_label, iLimit);
 		/* Otherwise return an error and stop */
-		const char *err = tt_sprintf(tnt_errcode_desc(ER_SQL_EXECUTE),
-					     "Only positive integers are "\
-					     "allowed in the LIMIT clause");
+		const char *err = "Only positive integers are allowed in the "
+				  "LIMIT clause";
 		sqlVdbeResolveLabel(v, halt_label);
 		sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, err,
 			      P4_STATIC);
@@ -2308,10 +2307,8 @@ computeLimitRegisters(Parse * pParse, Select * p, int iBreak)
 				sqlVdbeAddOp2(v, OP_Integer, 1, r1);
 				int no_err = sqlVdbeMakeLabel(v);
 				sqlVdbeAddOp3(v, OP_Eq, iLimit, no_err, r1);
-				err = tnt_errcode_desc(ER_SQL_EXECUTE);
-				err = tt_sprintf(err, "Expression subquery "\
-						 "could be limited only "\
-						 "with 1");
+				err = "Expression subquery could be limited "
+				      "only with 1";
 				sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0,
 					      0, err, P4_STATIC);
 				sqlVdbeAddOp1(v, OP_Halt, -1);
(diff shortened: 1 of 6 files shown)


nshy (approved):
Do the ground job for #9108. No objections.

src/box/sql/select.c
```diff
@@ -2276,12 +2276,11 @@ computeLimitRegisters(Parse * pParse, Select * p, int iBreak)
...
-		const char *err = tt_sprintf(tnt_errcode_desc(ER_SQL_EXECUTE),
-					     "Only positive integers are "\
-					     "allowed in the LIMIT clause");
+		const char *err = "Only positive integers are allowed in the "
+				  "LIMIT clause";
 		sqlVdbeResolveLabel(v, halt_label);
-		sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, err,
-			      P4_STATIC);
+		sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0,
+			      sql_xstrdup(err), P4_DYNAMIC);
```
Gerold103, line 2283: Given that this string is constant, you can pass the pointer here directly and keep the memory `P4_STATIC`.

src/box/sql/select.c
```diff
@@ -2308,12 +2307,10 @@ computeLimitRegisters(Parse * pParse, Select * p, int iBreak)
...
 				sqlVdbeAddOp3(v, OP_Eq, iLimit, no_err, r1);
-				err = tnt_errcode_desc(ER_SQL_EXECUTE);
-				err = tt_sprintf(err, "Expression subquery "\
-						 "could be limited only "\
-						 "with 1");
+				err = "Expression subquery could be limited "
+				      "only with 1";
 				sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0,
-					      0, err, P4_STATIC);
+					      0, sql_xstrdup(err), P4_DYNAMIC);
```
Gerold103, line 2313: Same.

src/box/sql/select.c
```diff
@@ -2335,12 +2332,11 @@ computeLimitRegisters(Parse * pParse, Select * p, int iBreak)
...
-			err = tt_sprintf(tnt_errcode_desc(ER_SQL_EXECUTE),
-					 "Only positive integers are allowed "\
-					 "in the OFFSET clause");
+			err = "Only positive integers are allowed in the "
+			      "OFFSET clause";
 			sqlVdbeResolveLabel(v, offset_error_label);
-			sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, err,
-				      P4_STATIC);
+			sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0,
+				      sql_xstrdup(err), P4_DYNAMIC);
```
Gerold103, line 2339: Same.

src/box/sql/select.c
```diff
@@ -5506,10 +5502,9 @@ vdbe_code_raise_on_multiple_rows(struct Parse *parser, int limit_reg, int end_ma
 	int r1 = sqlGetTempReg(parser);
 	sqlVdbeAddOp2(v, OP_Integer, 0, r1);
 	sqlVdbeAddOp3(v, OP_Ne, r1, end_mark, limit_reg);
-	const char *error = tt_sprintf(tnt_errcode_desc(ER_SQL_EXECUTE),
-				       "Expression subquery returned more "\
-				       "than 1 row");
-	sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, error, P4_STATIC);
+	const char *error = "Expression subquery returned more than 1 row";
+	sqlVdbeAddOp4(v, OP_SetDiag, ER_SQL_EXECUTE, 0, 0, sql_xstrdup(error),
+		      P4_DYNAMIC);
```
Gerold103, line 5507: As well.

Gerold103 (changes requested):
Thanks for the patch!

src/box/sql/select.c
ImeevMA: Thank you for the review! Fixed.

src/box/sql/select.c
ImeevMA: Fixed.

src/box/sql/select.c
ImeevMA: Fixed.

src/box/sql/select.c
ImeevMA: Fixed.

Gerold103 (approved):
Thanks for the fixes!

PR #9813, opened 2024-03-14
config: introduce labels to schema

### LINKED ISSUE #9809: config: add labels to schema
Sometimes it is useful for end user or application to consider tags or labels set for the instance. In a [private discussion](https://www.notion.so/tarantool/b7a2c5a9dc0c4ef6a8de923f5c560ff3) it was decided to implement labels as a simple `schema.map[string]string` similar to cartridge:

```yaml
# New 3.1 top level field.
labels:
  zone: msk
  color: fuchsia
```

Note that `schema.map` are merged as sets:
```yaml
groups:
  g-1:
    replicasets:
      r-1:
        labels:
          foo: 'true'
          bar: 'true'
      instances:
        i-1:
          labels: # result -> [{bar: 'true', baz: 'true', foo: 'false'}]
            baz: 'true'
            foo: 'false'
```

Lord-KA:
Closes #9809

@TarantoolBot document
Title: New labels in config schema

The new labels are basically maps with string keys and values, that are merged down to instance level. It means that if a replicaset and an instance inside it have label with same key and different value, the actual value is the one instance provided, e.g.

```yaml
groups:
  group-001:
    replicasets:
      replicaset-001:
        labels:
          foo: 'true'
          bar: 'true'
        instances:
          instance-001:
            labels:
              baz: 'true'
              foo: 'false'
```

Results in:

```
./instance-001.iproto> require('config'):get('labels')
---
- baz: 'true'
  foo: 'false'
  bar: 'true'
...
```

### DIFF
--- a/changelogs/unreleased/gh-9809-config-add-labels.md
+++ b/changelogs/unreleased/gh-9809-config-add-labels.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Introduced the `labels` configuration option. The labels are maps with
+  string keys that are merged to instance level (gh-9809).
--- a/test/config-luatest/labels_test.lua
+++ b/test/config-luatest/labels_test.lua
@@ -1,0 +1,38 @@
+local t = require('luatest')
+local cbuilder = require('test.config-luatest.cbuilder')
+local cluster = require('test.config-luatest.cluster')
+
+local g = t.group()
+
+g.before_all(cluster.init)
+g.after_each(cluster.drop)
+g.after_all(cluster.clean)
+
+-- Verify that instance labels are merged correctly.
+g.test_labels = function(g)
+    local config = cbuilder.new()
+        :set_replicaset_option('labels', {
+            foo = 'true',
+            bar = 'true',
+        })
+        :add_instance('instance-001', {
+            labels = {
+                baz = 'true',
+                foo = 'false',
+            },
+        })
+        :config()
+
+    local cluster = cluster.new(g, config)
+    cluster:start()
+
+    cluster['instance-001']:exec(function()
+        local config = require('config')
+
+        t.assert_equals(config:get('labels'), {
+            foo = 'false',
+            bar = 'true',
+            baz = 'true',
+        })
+    end)
+end
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2378,6 +2378,15 @@ return schema.new('instance_config', schema.record({
             default = 'old',
         }),
     }),
+    -- Instance labels.
+    labels = schema.map({
+        key = schema.scalar({
+            type = 'string',
+        }),
+        value = schema.scalar({
+            type = 'string',
+        }),
+    }),
 }, {
     -- This kind of validation cannot be implemented as the
     -- 'validate' annotation of a particular schema node. There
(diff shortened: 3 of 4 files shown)


test/config-luatest/labels_test.lua
```diff
@@ -0,0 +1,35 @@
+local t = require('luatest')
+local helpers = require('test.config-luatest.helpers')
+local cbuilder = require('test.config-luatest.cbuilder')
+local cluster = require('test.config-luatest.cluster')
+
+local g = helpers.group()
```
Totktonada, line 6: Nit: It performs all the job regarding stopping tarantool processes and removing temporary files/directories. However, it leaves `g.cluster` in an inconsistent state.

For example, an attempt to add another test case that uses `cluster.new()` would fail. (At least I guess that it is so based on the source code.)

Please, use the usual boilerplate.

https://github.com/tarantool/tarantool/blob/5040fba9cf1da942371721e36e81c7372699600c/test/config-luatest/cluster.lua#L3-L12

`helpers.group()` fits good, when functions from this helper module are in use.

Totktonada (approved):

test/config-luatest/labels_test.lua
Lord-KA: Thanks for the catch. Fixed.

Totktonada: It calls the treegen's `init`/`clean` functions twice. It is harmless, but I would simplify it a bit. In fact, we don't need the `test/config-luatest/helpers.lua` usage at all.

```diff
diff --git a/test/config-luatest/labels_test.lua b/test/config-luatest/labels_test.lua
index 0db0235c27..5da961f2cc 100644
--- a/test/config-luatest/labels_test.lua
+++ b/test/config-luatest/labels_test.lua
@@ -1,9 +1,8 @@
 local t = require('luatest')
-local helpers = require('test.config-luatest.helpers')
 local cbuilder = require('test.config-luatest.cbuilder')
 local cluster = require('test.config-luatest.cluster')
 
-local g = helpers.group()
+local g = t.group()
 
 g.before_all(cluster.init)
 g.after_each(cluster.drop)
```

I'm going to just fix it and don't play a ping-pong.

I want to eliminate a need in this boilerplate during integration of the helpers into `luatest`. When `treegen` and `cluster` will be part of `luatest`, these modules should automatically add appropriate hooks.

Totktonada: Force-pushed the branch.

ochaplashkin (approved):

changelogs/unreleased/gh-9809-config-add-labels.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Introduced labels as configuration option, they are basically string to
+  string maps that are merged together (gh-9809).
```
xuniq, line 4:
```suggestion
* Introduced the `labels` configuration option. The labels are maps with string keys that are merged to instance level. (gh-9809).
```

xuniq (approved):

changelogs/unreleased/gh-9809-config-add-labels.md
Lord-KA: Applied, thanks.

PR #9839, opened 2024-03-20
Implement net.replicaset base API

### LINKED ISSUE #9823: Add net.replicaset to tarantool 3.0
We decided to move the project from Aeon to Tarantool.

See https://github.com/tarantool/aeon/pull/15

### LINKED ISSUE #9876: Generate message for box.error{code = ..} automatically
Now there are several way to generate box errors:
```
tarantool> box.error(box.error.NIL_UUID) -- number positional argument is a ClientError
---
- error: Nil UUID is reserved and can't be used in replication
...
tarantool> box.error('MyCustomError', 'Bad thing happened') -- string positional argument is a custom error
---
- error: Bad thing happened
...
tarantool> box.error{code = box.error.NIL_UUID} -- no named argument 'type' - is a ClientError
---
- error: 
...
tarantool> box.error{type = 'MyCustomError', reason = 'Bad thing happened'} -- named argument 'type' - is a custom error
---
- error: Bad thing happened
...
```
Note that if ClientError is created using positional argument (the code) - the reason is substituted from internal error list. But when we use named arguments (table with arguments) - the reason an message are left empty. So using named arguments in box.error for ClientError implies knowing of error message, which at least inconvenient (if at least possible).

I propose to use error template (which defined in box core) for reason in ClientError when only code is known. Actually we plan to avoid argument substitution to this template, so the template would be exactly what is needed.

alyapunov:
That's a MVP so only part of the design document is implemented.
Only `call_leader` and `watch_leader` methods are implemented.
No additional options of connection are implemented (except those which are accepted by net.box.connect).
Having that the implementation is in internal namespace.

On a way here fast fixed #9876

Closes #9823


Gerold103 (changes requested):
Thanks for the patch! I will come back later for the other commit.

unera:
> Only call_leader and watch_leader methods are implemented.

Please, implement watch_all, too. Looks like that only the method will be useful.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+                for _, search_instance in pairs(replicaset.instances) do
+                    if search_instance.is_rw then
+                        replicaset.leader_instance = search_instance
+                    end
+                end
+            end
+        end
+        if prev_rw_count == 1 then
+            replicaset.leader_instance = nil
+        end
```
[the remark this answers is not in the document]
alyapunov: If I understood correctly in your code you will not need `prev_rw_count` but will have to add check and clear `replicaset.leader_instance` in both code branches. I don't think that temporary `prev_rw_count` is so bad that should be avoided.
In any case I rewrote this part of code in some kind of third way.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+    end
+end
+
+-- Internal watcher that examines statuses of replicas.
+local function box_status_watcher(replicaset, instance, _key, status)
+    local is_rw = not status.is_ro
+    if (replicaset.wait_status_count == 1 and instance.wait_status) or
+            (replicaset.wait_status_count ~= 0 and is_rw) then
+        replicaset.wait_status_cond:broadcast()
+    end
```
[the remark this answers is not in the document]
alyapunov: great, fixed/

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+end
+
+-- Call a function on replicaset leader if the leader is known.
+-- Args and opts are exactly the same as in net.box.call(..).
+-- Note that known leader can become not leader while the call is inprogress,
+-- so a certain error (like ER_READONLY) may happen.
+-- NO_WRITEABLE and MORE_THAN_ONE_WRITEABLE may be thrown.
+function replicaset_methods:call_leader(func, args, opts)
+    check_replicaset_arg(self, 'call_leader')
+    self:wait_rw(opts and opts.timeout or nil)
```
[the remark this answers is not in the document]
alyapunov: Ok, added timeout countdown.
BTW my first [version](https://github.com/tarantool/aeon/commit/47c8317bc8353308be003d9893a0df91092b67b0#diff-a408549ef4a3678eff703b2582e8f6475a9138d55bb8b0ac02447e72ab61f268R27) was logically similar to your `netbox_wait_connected` (I mean the sequence precheck -> gettime -> loop -> wait -> check -> gettime), but it didn't passed a [review](https://github.com/tarantool/aeon/pull/15#discussion_r1524574091).

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+    end
+end
+
+-- Wait for status notification from at least one RW instance or entire
+-- replicaset if none of the instances is RW.
+-- Return true if at least one RW was found, false otherwise.
+-- Opts can be a table with members:
+--  timeout: limits the wait.
+function replicaset_methods:wait_rw(opts)
+    local timeout = opts and opts.timeout and opts.timeout or TIMEOUT_INFINITY
```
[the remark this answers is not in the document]
alyapunov: fixed.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+        for k, v in pairs(cfg) do
+            if k ~= 'instances' and k ~= 'name' then
+                cumulative_cfg[k] = v
+            end
+        end
+        for k, v in pairs(cfg_instance) do
+            if k ~= 'endpoint' then
+                cumulative_cfg[k] = v
+            end
+        end
```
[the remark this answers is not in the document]
alyapunov: Added `reconnect_timeout` explicitly, didn't implement all other (as not expected in MVP).
Actually I see no reason in rename reconnect_after/reconnect_timeout and I don't see any harm in bypassing any current or future net.box options. User/pass, `fetch_schema` - why not. But the difference does not cost arguing.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,266 @@
...
+        cumulative_cfg.fetch_schema = false
+        cumulative_cfg.wait_connected = false
+        local conn = net_box.connect(endpoint, cumulative_cfg)
+        local instance = {
+            instance_name = instance_name,
+            endpoint = endpoint,
```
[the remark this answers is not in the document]
alyapunov: Removed along with cfg.

test/box-luatest/gh_9823_net_replicaset_test.lua
```diff
@@ -0,0 +1,161 @@
...
+        if server.alias:startswith('storage') then
+            server:exec(function()
+                box.cfg{read_only = true}
+            end)
+        end
+    end)
+
+    -- Check that the replicaset does not work.
+    cluster.router_1:exec(function()
+        local t = require('luatest')
```
Gerold103, line 57: From what I know, you don't need to do it in execs anymore. It is there by default with the same name as in the test file.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,376 @@
...
+local function rs_error(err_code, replicaset_name, payload)
+    local error_info = error_list[err_code]
+    assert(error_info)
+    local args = payload and table.copy(payload) or {}
+    args.type = ERROR_TYPE
+    args.code = err_code
+    args.name = error_info.name
+    args.message = error_info.message
+    args.replicaset = replicaset_name
+    box.error(args)
```
locker, line 42: What's the point in introducing the new type of errors. Why not reuse the existing type `ClientError` used for all other Tarantool errors? We can add the new error codes to `src/box/errcode.h`.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,376 @@
...
+        if status == 'rw' then
+            replicaset.leader_instance = instance
+        else
+            for _, search_instance in pairs(replicaset.instances) do
+                if search_instance.status == 'rw' then
+                    replicaset.leader_instance = search_instance
+                end
+            end
+        end
+        replicaset_rebind_watchers(replicaset)
```
locker, line 236: A good thing about watchers is that it's guaranteed that a watcher callback is never executed in parallel with itself for the same key. This simplifies programming callbacks because one doesn't have to bother about synchronization.

Unregistering a net.box watcher doesn't stop the currently running callback. As a result, during leader switching it may occur that the same watcher callback is running in two fibers - for the old leader and for the new one.

src/box/lua/net_replicaset.lua
unera: Yes, I think that it was my mistake (to use dedicated error type) if the module is a Core module, it should use Core error type.

src/box/lua/net_replicaset.lua
unera: Here we can use a condvar (for example)

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,376 @@
...
+}
+
+-- Map of error codes by error name.
+local error_codes = {}
+for code, err in ipairs(error_list) do
+    error_codes[err.name] = code
+end
+
+-- Raise a net.replicaset specific error.
+local function rs_error(err_code, replicaset_name, payload)
```
unera, line 33: Now we patched box.error (added `level` option), so please review the code to provide the functional (add argument `level`, and set `level` to default value (if a user doesn't use the argument), to point error place)

alyapunov (commented):
Added a commit that fixes box.error.

src/box/lua/net_replicaset.lua
alyapunov: Remade using ClientError.

alyapunov:
added `fetch_schema == false` and repushed.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,355 @@
...
+local replicaset_methods = {}
+
+local replicaset_mt = {
+    __index = replicaset_methods,
+}
+
+-- Check that the arguments table comply table template.
+-- Template key-value pairs show possible options and their required types -
+-- a string concatenated from several types with reasonable delimiter.
+local function check_options(tbl, tmpl, name)
```
unera, line 24: Кажется, эти ошибки стоит сделать box.error, поскольку их получает клиент.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,355 @@
...
+
+-- Call a function on replicaset leader if the leader is known.
+-- Args and opts are exactly the same as in net.box.call(..).
+-- Note that known leader can become not leader while the call is inprogress,
+-- so a certain error (like ER_READONLY) may happen.
+-- NO_WRITEABLE and MORE_THAN_ONE_WRITEABLE may be thrown.
+function replicaset_methods:call_leader(func, args, opts)
+    replicaset_check(self, 'call_leader')
+    opts = opts or {}
+    check_options(opts, call_opts_template, 'call options')
```
unera, line 85: этот метод бросает исключение.
соответственно, хорошо бы удалённый узел получал нормальную ошибку, а не error (см коммент выше)

src/box/lua/net_replicaset.lua
alyapunov: Мы обычно так не делаем. Например в том же net.box, если перередать неправильные аргументы в conn:call - будет просто error. Логика в том, что такая ошибка никогда не должна прилетать пользователю, это косяк на этапе разработки который должен быть вскрыт на раннем этапе при помощи этой ошибки.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,268 @@
...
+-- The leader is the only known RW replica.
+-- Note1. This method does not wait for all peers to connect and send status.
+-- For example it will successfully return if one replica is known as RW while
+-- two others have not send status event. This is done for availability of
+-- replicaset connection: if one or more replica instance are down it should
+-- be possible to proceed requests with replicaset remains.
+-- Note2. Due to asynchronous nature of replicaset connection it is possible
+-- that the known leader is not actually a leader or perhaps not even RW.
+-- Returns remaining timeout.
+function replicaset_methods:_wait_leader(timeout)
```
Gerold103, line 59: Btw, if a method is fully internal (never even used outside of this Lua file), then you can just declare it as a function and it doesn't have to be in the methods table. Then you would have to pass the `self` argument explicitly, but the function wouldn't be visible outside at all.

Up to you. Just a note.

src/box/lua/net_replicaset.lua
```diff
@@ -0,0 +1,268 @@
...
+function replicaset_methods:close()
+    replicaset_check(self, 'close')
+    for _, instance in pairs(self.instances) do
+        instance.conn:close()
+    end
+end
+
+-- Callback that called if the status of an instance is (possibly) changed.
+-- Called from watcher ('ro' or 'rw') or from on_disconnect ('unknown').
+local function instance_status_change(replicaset, instance, status)
```
Gerold103, line 198: I would recommend naming triggers with `on` suffix or prefix. Up to you.

src/box/lua/net_replicaset.lua
```diff
@@ -102,6 +102,89 @@ function replicaset_methods:call_leader(func, args, opts)
...
+    end
+    watcher.replicaset.leader_watchers[watcher] = nil
+end
+
+-- Get closure with `key`, `value` arguments that calls user watcher function
+-- with extended signature (with the third `instance_name` argument).
+-- In addition the closure uses guard preventing concurrent execution of func.
+local function wrap_watcher_function(watcher, instance)
+    local instance_name = instance.instance_name
+    return function(key, value)
```
Gerold103, line 134: Dangerous ground here. If we will ever reference `instance` in the callback, we would get another cyclic ref  and break GC. Lets add a test that adding a watch doesn't prevent GC when all public refs are gone.

src/box/lua/net_replicaset.lua
alyapunov: Oh, there were too many battles about how object's methods should be implemented.. I would prefer not to change it again.

src/box/lua/net_replicaset.lua
unera: даже наше "обычно" не совсем обычно.

например в net.box есть места где валидация параметров бросает lua error, а есть места где бросается ILLEGAL_PARAM

так что в net.replicaset давайте делать единообразно -- все ошибки box.error, а не error

unera: box.error.ILLEGAL_PARAMS в net_box.lua встречается 14 раз
простой error - 192 раза

в общем у нас бардак - и та практика и эта

с учётом того, что мы выдаём новую подсистему, чтобы привести всё в порядок - предлагаю новые проекты писать в парадигме KV-ошибок

test/box-luatest/gh_9823_net_replicaset_test.lua
```diff
@@ -180,3 +180,238 @@ g.test_net_replicaset_basics = function(cg)
...
+                    box.broadcast('key1', 1)
+                else
+                    box.broadcast('key1', 2)
+                end
+            end)
+        end
+    end)
+
+    -- Subscribe and check that event was received only from leader.
+    local leader_name = cluster.router:exec(function()
```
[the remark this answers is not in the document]
alyapunov: I tried to do this way, spend several hours debugging and found out that the test cluster is started in a random temporary folder and I found no public API to retrieve its name. There's `cluster._dir` but it seems to be private and a subject to change.
Since the cluster instances are started with an unix socket listen port, they are cannot be accessed from the main test process.

test/box-luatest/gh_9823_net_replicaset_test.lua
```diff
@@ -180,3 +180,238 @@ g.test_net_replicaset_basics = function(cg)
...
+                require('fiber').sleep(0.1)
+            end
+        end
+        test_data.long_watcher = rs:watch_leader('long', long_func)
+
+        t.helpers.retrying({timeout = 2, delay = 0.1}, function()
+            t.assert_equals(test_data.key1_value, 1)
+        end)
+        t.assert_equals(test_data.key2_value, 0)
+        t.helpers.retrying({timeout = 2, delay = 0.1}, function()
```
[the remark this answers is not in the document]
alyapunov: Set 10.

src/box/lua/net_replicaset.lua
Gerold103: I am fine with whatever.

Gerold103 (commented):
I was asked to have another look today. I can't see any other comments besides the ones which are unresolved above.

src/box/lua/net_replicaset.lua
alyapunov: Checked, considering that while the watcher is not unregistered the replicaset must live.

src/box/lua/net_replicaset.lua
alyapunov: Ok, made it `ILLEGAL_PARAMS`.

test/box-luatest/gh_9823_net_replicaset_test.lua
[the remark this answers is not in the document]
alyapunov: I mean now the test uses static relative unix socket paths. Perhaps we could collect connect_cfg from test.config-luatest.cluster. Is it worth it?

unera (approved):
lgtm

test/box-luatest/gh_9823_net_replicaset_test.lua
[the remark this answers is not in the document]
alyapunov: Ok, remade in this way.

test/box-luatest/gh_9823_net_replicaset_test.lua
```diff
@@ -0,0 +1,218 @@
...
+
+    -- Check that unknown replicaset does not work.
+    cluster.router:exec(function()
+        local net_replicaset = require('internal.net.replicaset')
+        local err_msg = 'The replicaset was not found by its name'
+        local succ, err = pcall(net_replicaset.connect, 'moon')
+        t.assert_equals(succ, false)
+        t.assert_equals(err.message, err_msg)
+        t.assert_equals(err.replicaset, 'moon')
+    end)
```
Gerold103, line 79: It seems you can do it right in the test instance and drop router now completely. This is the only place where router still used. Is there any reason this test case couldn't be done in the main process?

Gerold103 (commented):
Perfect, just one last thing to clarify.

test/box-luatest/gh_9823_net_replicaset_test.lua
alyapunov: Actually there's another place in test_net_replicaset_basics. These tests check connect by replicaset name, and that requires cluster config. The router has has cluster config while the main instance does not, so connection by name have to be tested on router instance.

Gerold103: Sad that the config is created in the main process, but somehow can't be used in it.

Gerold103 (approved):
Thanks for the patch!

ochaplashkin (approved):

src/box/lua/net_replicaset.lua
alyapunov: переделал, см. выше, теперь там box.error.

PR #9846, opened 2024-03-21
config: allow empty configuration file

### LINKED ISSUE #9845: config: nil error on empty configuration file
When providing empty (in terms of YAML) file as configuration to Tarantool, it returns non-verbose error:
```
$ ./src/tarantool --name instance-001 --config single.yaml 
LuajitError: [cluster_config] Unexpected data type for a record: "nil"
```

Lord-KA:
Before this patch, Tarantool, when started with empty configuration file, used to fail with a non verbose error:
```
$ ./src/tarantool -n instance-001 -c single.yaml
LuajitError: [cluster_config] Unexpected data type for a record: "nil"
```

Now Tarantool instance can be set up with configuration from remote
configuration sources available in Tarantool Enterprise Edition and
run with empty config file. When there is no configuration provided
for the instance an according error is thrown.

Closes #9845

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-9845-nil-error-on-empty-config.md
+++ b/changelogs/unreleased/gh-9845-nil-error-on-empty-config.md
@@ -1,0 +1,5 @@
+## bugfix/config
+
+* Fixed a non-verbose error on an empty configuration file.
+  Now Tarantool can successfully start up with an empty configuration
+  file using data from other configuration sources (gh-9845).
--- a/src/box/lua/config/source/file.lua
+++ b/src/box/lua/config/source/file.lua
@@ -26,6 +26,12 @@ function methods.sync(self, config_module, _iconfig)
             config_module._config_file, res))
     end
 
+    -- YAML returns `nil` or `box.NULL` on empty file,
+    -- while config sources should be {} if empty.
+    if res == nil then
+        res = {}
+    end
+
     self._values = res
 end
 
--- a/test/config-luatest/sources_test.lua
+++ b/test/config-luatest/sources_test.lua
@@ -176,3 +176,24 @@ g.test_sources_priority = function(g)
         t.assert_equals(config:get('log.syslog.identity'), 'from env default')
     end)
 end
+
+g.test_empty_sources = function()
+    local dir = treegen.prepare_directory(g, {}, {})
+    local test_cases = {
+        "",
+        "--- null\n...\n",
+    }
+
+    for _, case in ipairs(test_cases) do
+        treegen.write_script(dir, 'single.yaml', case)
+
+        local exp = "No cluster config received from the given " ..
+                    "configuration sources."
+
+        local args = {'--name', 'instance-001', '--config', 'single.yaml'}
+        local res = justrun.tarantool(dir, {}, args, {stderr = true})
+
+        t.assert_equals(res.exit_code, 1)
+        t.assert_str_contains(res.stderr, exp)
+    end
+end


ochaplashkin (approved):

changelogs/unreleased/gh-9845-nil-error-on-empty-config.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/config
+
+* Fixed non-verbose error on empty config file. Now Tarantool can
+  successfully startup with empty config file using data from other
+  configuration sources (gh-9845).
```
xuniq, line 5:
```suggestion
* Fixed a non-verbose error on an empty configuration file. Now Tarantool
  successfully starts up with an empty configuration file using data from other
  configuration sources (gh-9845).
```

xuniq (approved):

changelogs/unreleased/gh-9845-nil-error-on-empty-config.md
Lord-KA: Thanks, applied.

src/box/lua/config/source/file.lua
```diff
@@ -26,6 +26,10 @@ function methods.sync(self, config_module, _iconfig)
             config_module._config_file, res))
     end
 
+    -- YAML returns `nil` on empty file, while config sources
+    -- should be {} if empty.
+    res = res or {}
```
Totktonada, line 31: Let's consider the following possible situations.

1. No config file.
2. An empty config file.
3. A config file with an empty YAML document (`---\n...\n`).
4. A config file with YAML's representation of nil/null value (`null` or `--- null\n...\n`).

Before this patch we handle 1) correctly, but other leads to a non-descriptive error. Now, 2) and 3) are handled correctly too.

However, 4) gives `box.NULL` from `yaml.decode()`, so `or {}` doesn't work.

```lua
tarantool> yaml.decode(yaml.encode(nil)) or {}
---
- null
...
```

I propose to handle `box.NULL` here in the same way as `nil`:

```lua
if res == nil then
    res = {}
end
```

This way we cover 4) in the same way as 1), 2) and 3).

Totktonada (approved):
Thank you for the fix!

It looks OK for me.

I have a small proposal here. Please, consider my comment above.

src/box/lua/config/source/file.lua
Lord-KA: Actually, it seems case 3 isn't working properly as well, YAML treats it as an empty string and returns such, so the check should be:
```lua
if res == nil or res == box.NULL or res == ""
    res = {}
end
```

Totktonada: Wow, I forgot about this. Found my own commit 4095e3050a82223b138a71f1723ffb320f6ec142, where I've commented this behavior as tarantool specific one.

Personally I think that we shouldn't fix yaml decoding in configuration sources. But it would be good to interpret `nil` and `box.NULL` values from decoding in the same way.

BTW,

```lua
if res == nil or res == box.NULL then
    <...>
end

-- ..is the same as..

if res == nil then
    <...>
end
```

So, finally, I would leave it as follows.

```lua
if res == nil then
    res = {}
end
```

Lord-KA: Okay, fixed.

Totktonada:
Backported to `release/3.0` (future `3.0.2`) in PR #9860.

PR #9850, opened 2024-03-21
error: allow to specify where to get error location

### LINKED ISSUE #9792: Allow to set trace level in `box.error` and `box.error.new`
[`box.error`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/error/) always uses the current level to set the trace:

https://github.com/tarantool/tarantool/blob/567cba8ae7fb1c398d39e7c247466c2c8bd216a1/src/box/lua/error.cc#L193

As a result, the trace becomes useless if `box.error` or `box.error.new` is called from a helper function used to construct errors, as it will always point to the helper function itself. For example,

```lua
-- Helper function that raises an error of type 'MyError' with the given message.
local function my_error(msg)
    box.error{type = 'MyError', message = msg}
end

-- Some function that may raise an error.
local function check_answer(v)
    if v ~= 42 then
        my_error('Bad answer!')
    end
end

local ok, err = pcall(check_answer, 1)
err.trace -- points to my_error, not to my_func, as it should
```

The vanilla function [`error`](https://www.lua.org/pil/8.5.html) has the level argument which can be used to force using a trace from an upper level. We should introduce a similar functionality for `box.error` and `box.error.new`. For example,

```Lua
box.error{type = 'MyError', message = 'Error!', level = 3}
-- or
box.error({type = 'MyError', message = 'Error!'}, 3)
```

locker:
Now, when used with a table argument, `box.error` and `box.error.new` also accept an optional second argument called `level`. It has the same meaning as the `level` argument of the built-in Lua function `error`: it specifies how to get the error location. With level 1 (the default), the error location is where `box.error` / `box.error.new` was called. Level 2 points the error to where the function that called `box.error` / `box.error.new` was called; and so on. Passing level 0 avoids addition of location information to the error.

Example of using `level` with `box.error`:

```lua
local json = require('json')

local function inner(level)
    box.error({message = 'test'}, level)    -- line:4
end

local function outer(level)
    inner(level)                            -- line:8
end

local ok, err
ok, err = pcall(outer)
print(json.encode(err.trace))               -- prints line:4
ok, err = pcall(outer, 1)
print(json.encode(err.trace))               -- prints line:4
ok, err = pcall(outer, 2)
print(json.encode(err.trace))               -- prints line:8
ok, err = pcall(outer, 0)
print(json.encode(err.trace))               -- prints empty table
```

Example of using `level` with `box.error.new`:

```lua
local json = require('json')

local function inner(level)
    local err = box.error.new({message = 'test'}, level)    -- line:4
    return err
end

local function outer(level)
    local err = inner(level)                                -- line:9
    return err
end

local err
err = outer()
print(json.encode(err.trace))               -- prints line:4
err = outer(1)
print(json.encode(err.trace))               -- prints line:4
err = outer(2)
print(json.encode(err.trace))               -- prints line:9
ok, err = pcall(outer, 0)
print(json.encode(err.trace))               -- prints empty table
```

It is also possible to specify `level` when using `box.error` to re-raise an error created earlier with `box.error.new`, for example:

```lua
local json = require('json')

local err0 = box.error.new{message = 'test'}    -- line:3

local function raise(err, level)
    box.error(err, level)                       -- line:6
end

ok, err = pcall(raise, err0)
print(json.encode(err.trace))               -- prints line:3

ok, err = pcall(raise, err0, 1)
print(json.encode(err.trace))               -- prints line:6

ok, err = pcall(raise, err0, 0)
print(json.encode(err.trace))               -- prints empty table
```

Closes #9792

### DIFF
--- a/changelogs/unreleased/gh-9792-error-level.md
+++ b/changelogs/unreleased/gh-9792-error-level.md
@@ -1,0 +1,4 @@
+## feature/lua
+
+* Added the `level` argument to `box.error` and `box.error.new` to specify
+  the stack frame used for setting the error location (gh-9792).
--- a/test/app-luatest/gh_9792_error_level_test.lua
+++ b/test/app-luatest/gh_9792_error_level_test.lua
@@ -1,0 +1,186 @@
+local t = require('luatest')
+
+local g = t.group()
+
+local TEST_FILE = debug.getinfo(1, 'S').short_src
+
+g.test_invalid_args = function()
+    local arg, errmsg
+
+    arg = {type = 'foo', message = 'bar'}
+
+    -- Raising a new error with a bad level argument.
+    errmsg = 'box.error(): bad arguments'
+    t.assert_error_msg_equals(errmsg, box.error, arg, {})
+    t.assert_error_msg_equals(errmsg, box.error, arg, 'x')
+    t.assert_error_msg_equals(errmsg, box.error, arg, '1')
+    t.assert_error_msg_equals(errmsg, box.error, arg, 1, 2)
+
+    -- Creating a new error with a bad level argument.
+    errmsg = 'box.error.new(): bad arguments'
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, {})
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, 'x')
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, '1')
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, 1, 2)
+
+    arg = box.error.new(arg)
+
+    -- Raising an existing error with a bad level argument.
+    errmsg = 'box.error(): bad arguments'
+    t.assert_error_msg_equals(errmsg, box.error, arg, {})
+    t.assert_error_msg_equals(errmsg, box.error, arg, 'x')
+    t.assert_error_msg_equals(errmsg, box.error, arg, '1')
+    t.assert_error_msg_equals(errmsg, box.error, arg, 1, 2)
+
+    -- Sic: box.error.new() doesn't take an error object.
+    errmsg = 'box.error.new(): bad arguments'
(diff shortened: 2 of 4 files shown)


nshy (approved):

test/app-luatest/gh_9792_error_level_test.lua
```diff
@@ -0,0 +1,186 @@
...
+    t.assert_error_msg_equals(errmsg, box.error, arg, 1, 2)
+
+    -- Sic: box.error.new() doesn't take an error object.
+    errmsg = 'box.error.new(): bad arguments'
+    t.assert_error_msg_equals(errmsg, box.error.new, arg)
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, 1)
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, {})
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, 'x')
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, '1')
+    t.assert_error_msg_equals(errmsg, box.error.new, arg, 1, 2)
```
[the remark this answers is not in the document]
locker: I just want to test that passing illegal values raises a proper error. It shouldn't hurt.

changelogs/unreleased/gh-9792-error-level.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Added a `level` argument to `box.error` and `box.error.new` for specifying
+  the stack frame used for setting the error location (gh-9792).
```
xuniq, line 4:
```suggestion
* Added the `level` argument to `box.error` and `box.error.new` to specify
  the stack frame used for setting the error location (gh-9792).
```

xuniq (approved):

ylobankov (approved):

changelogs/unreleased/gh-9792-error-level.md
locker: Fixed, thanks!

Issue #9217, opened 2023-10-04
sysprof and memprof parsers lack error handling


mkokryashkin:
**Bug description**
* OS: Linux
* OS Version: any
* Architecture: x86_64

Both the sysprof and memprof parsers lack thorough error handling. It results in raw lua errors for all kinds of situations: incorrect file passed, file being corrupted, empty, etc. Below are some examples of the errors mentioned:

```
$ ./build/src/luajit -ts sysprof.bin
$ file sysprof.bin
sysprof.bin: empty
$ ./build/src/luajit -ts sysprof.bin
./build/src/luajit: ./tools/utils/symtab.lua:107: attempt to concatenate local 'magic' (a nil value)
```

```
$  ./build/src/luajit -ts mee
./build/src/luajit: ./tools/utils/bufread.lua:122: fopen, errno: 2
```

**Expected behavior**

Parsers should be refactored so all possible errors are reported in a concise and user-friendly way. Maybe it would be nice to provide some possible solutions along with errors. No raw Lua errors must be thrown.

igormunkin:
Looks like part of #5994.

PR #9853, opened 2024-03-21
build: add CMake configuration for evread module

mkokryashkin:
This patch adds necessary cmake configurations for the <evread.lua> module, so it can be used later to implement human-readable error reporting in profile parsers.

Part of #9217

NO_DOC=LJ
NO_TEST=LJ
NO_CHANGELOG=build


Buristan (approved):
Hi, Maxim!
Thanks for the patch!
LGTM, except a general nit to all changes.
Please sort all new entries alphabetically.

mkokryashkin:
> Hi, Maxim! Thanks for the patch! LGTM, except a general nit to all changes. Please sort all new entries alphabetically.

Thanks for the comment! Fixed for places where order is insignificant.

ligurio (commented):
I would replace `LJ` with something more clear, for example "covered by LuaJIT tests":

```
NO_DOC=LJ
NO_TEST=LJ
```

src/lua/init.c
```diff
@@ -229,6 +230,7 @@ static const char *lua_modules[] = {
 	"utils.avl", utils_avl_lua,
 	"utils.bufread", utils_bufread_lua,
 	"utils.symtab", utils_symtab_lua,
+	"utils.evread", utils_evread_lua,
```
ligurio, line 233: Why not alphabetically sorted?

ligurio (approved):
Maxim, thanks for the patch! LGTM with minor comments.

src/lua/init.c
mkokryashkin: Because the symtab module is used by evread, hence it should go first. In other places order has no such meaning.

mkokryashkin:
> I would replace `LJ` with something more clear, for example "covered by LuaJIT tests":
> 
> ```
> NO_DOC=LJ
> NO_TEST=LJ
> ```

Fixed, thanks

src/lua/init.c
ligurio: Thanks!

ligurio (approved):

ylobankov (approved):

PR #9857, opened 2024-03-22
console: make __autocomplete retrival for cdata exception-safe

### LINKED ISSUE #9828: Tarantool crashes when pressing `Tab`
Reproducer: start Tarantool interactively and type `box.`:
```
mergen@mega:~/work/ee/tarantool/src$ ./tarantool -i
Tarantool 3.1.0-entrypoint-160-g40bd0eb1a
tarantool> box.
```

Then press `Tab`. The result is:
```
mergen@mega:~/work/ee/tarantool/src$ ./tarantool -i
Tarantool 3.1.0-entrypoint-160-g40bd0eb1a
ER_UNCONFIGURED: Please call box.cfg{} first
tarantool> box.mergen@mega:~/work/ee/tarantool/src$ 
```

CuriousGeorgiy:
Fixes a regression introduced in cf644abd, which added exception-unsafe code for retrieval of `__autocomplete` metamethod of cdata objects.

I wasn't able to test the orginal reported bug, since the luatest server starts with `box` configured.

Closes #9828

### DIFF
--- a/src/box/lua/console.c
+++ b/src/box/lua/console.c
@@ -288,6 +288,7 @@ static int saved_point = 0;
 
 static int console_hide_prompt_ref = LUA_NOREF;
 static int console_show_prompt_ref = LUA_NOREF;
+static int gettable_ref = LUA_NOREF;
 
 /**
  * Don't attempt to hide/show prompt in certain readline states.
@@ -983,6 +984,13 @@ tarantool_lua_console_init(struct lua_State *L)
 	console_hide_prompt_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 	lua_pushcfunction(L, lbox_console_show_prompt);
 	console_show_prompt_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+
+	/*
+	 * Compile a Lua wrapper for table indexation. For sanity, verify that
+	 * it compiles successfully.
+	 */
+	VERIFY(luaT_dostring(L, "return function(t, k) return t[k] end") == 0);
+	gettable_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


src/box/lua/console.c
```diff
@@ -1116,8 +1120,14 @@ lua_rl_getcompletion(lua_State *L)
...
 		 * metatable.
 		 */
+		lua_rawgeti(L, LUA_REGISTRYINDEX, gettable_ref);
+		lua_pushvalue(L, -3);
 		lua_pushstring(L, "__autocomplete");
-		lua_gettable(L, -3);
+		if (lua_pcall(L, 2, 1, 0) != 0) {
+			/* pcall returns an error to the stack */
+			lua_pushnil(L);
+			lua_replace(L, -2);
```
nshy, line 1129: Nit: looks like easier to pop error and then push nil.

src/box/lua/console.c
```diff
@@ -1116,8 +1120,14 @@ lua_rl_getcompletion(lua_State *L)
...
+		}
```
nshy, line 1130: I guess we'd better `return 0` as in case of `__autocomple` function error for consitency. We don't need to fallback to pass the test.

nshy (changes requested):

src/box/lua/console.c
CuriousGeorgiy: Okay, let's do it this way.

src/box/lua/console.c
CuriousGeorgiy: I thought that this variant of stack manipulation is slightly more efficient.

nshy (approved):

src/box/lua/console.c
```diff
@@ -983,6 +984,9 @@ tarantool_lua_console_init(struct lua_State *L)
 	console_hide_prompt_ref = luaL_ref(L, LUA_REGISTRYINDEX);
 	lua_pushcfunction(L, lbox_console_show_prompt);
 	console_show_prompt_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+
+	VERIFY(luaT_dostring(L, "return function(t, k) return t[k] end") == 0);
```
Gumix, line 988: I'd drop a comment here to explain what does it verify and why.

test/app-luatest/gh_6305_autocomplete_metamethod_test.lua
```diff
@@ -58,6 +58,12 @@ g.test__autocomplete = function()
                                                    'table11.second',
                                                    'table11.auto',
                                                    })
+    -- __index function can throw an error - should be Ok
+   setmetatable(tab, {__index = function() error('test error') end})
```
Gumix, line 62: Nit: Bad alignment.

Gumix (approved):

src/box/lua/console.c
CuriousGeorgiy: It's just a sanity check that the compilation is successful, not really meaningful. Anyways, dropped a comment here.

PR #9859, opened 2024-03-22
test: increase process termination waiting timeout

Totktonada:
This commit increases a time to wait of the process termination. It may take longer than 5 seconds, when tarantool is built with an address sanitizer. The address sanitizer generates a report at the process termination and it is not always a fast thing.

### DIFF
--- a/test/box-luatest/gh_7974_force_recovery_bugs_test.lua
+++ b/test/box-luatest/gh_7974_force_recovery_bugs_test.lua
@@ -27,7 +27,11 @@ g.test_unknown_request_type = function(cg)
     local s = cg.server
     s:start{wait_until_ready = false}
     local log = fio.pathjoin(s.workdir, s.alias .. '.log')
-    t.helpers.retrying({}, function()
+    -- NB: The address sanitizer may take some time to generate
+    -- its report and all this time the process is alive. It takes
+    -- more than the default retrying() timeout (5 seconds) in
+    -- some circumstances.
+    t.helpers.retrying({timeout = 60}, function()
         t.assert_not_equals(s:grep_log("can't initialize storage: " ..
                                        "Unknown request type 777", nil,
                                        {filename = log}), nil)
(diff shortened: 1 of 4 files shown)


Totktonada:
Assigned authors of the tests as reviewers.

@Gerold103, @CuriousGeorgiy, please, take a look. It seems, I need this patch to don't retry CI in tarantool-ee till my death and proceed with pending patches.

I've verified manually that the process termination in the given testing environment[^1] and on given snapshot/xlogs takes around 6 seconds if other tests are executing in parallel on the debug+asan build. This time drops to ~2 seconds if there is no background workload.

I suspect that we'll need to return back to these tests after https://github.com/tarantool/luatest/pull/357 will be pulled into tarantool's testing, because the ASAN report is not empty. However, I consider it as a separate potential problem. I didn't investigate it at all.

[^1]: A docker container on a virtual machine from the testing workers pool.

CuriousGeorgiy (approved):

Gerold103 (approved):
Thanks for the patch! I think `retrying()` timeout is in general too small. I said it when luatest was introduced. It is too small even for simper workloads. On my machine it was exploding when I was running some particularly heavy tests in too many parallel runs. In vshard I had to make it 50 seconds default because of that, but I observed this problem in core as well. Maybe just bump the default timeout in luatest itself.

My point is that if a test is stuck, then it will hang more than minutes until killed. And if a test isn't stuck, then it will run in less than a minute for sure. But it can sometimes take seconds if the machine is loaded. Timeouts like 5 seconds don't make much sense and lead to flakiness.

Totktonada:
@Gerold103 I completely agree that the default `retrying()` timeout is bad.

I don't know, how breaking would be an update of this default timeout right in luatest. Maybe it is acceptable.

Maybe it would be good to introduce some `.luatestrc` file and start to spread good defaults on a per-project basis.

ylobankov (approved):

PR #9870, opened 2024-03-26
[backport 3.0] test: increase process termination waiting timeout

Totktonada:
*(This is a backport of PR #9859 to `release/3.0`, future `3.0.2` release.)*

---

This commit increases a time to wait of the process termination. It may take longer than 5 seconds, when tarantool is built with an address sanitizer. The address sanitizer generates a report at the process termination and it is not always a fast thing.


Totktonada:
This PR has no changes comparing to #9859, which is approved by the authors of the tests and the infrastructure team. I'll proceed with the backport.

PR #9879, opened 2024-03-27
build: include header "unit.h" after c++ headers

drewdzzz:
Header "unit.h" contains `ok` and `is` macros used to check test cases. The problem is such simple names can be used in C++ STL library headers (it's OK because such short names can be hidden in a namespace), so when including, for example, header "vector" after "unit.h", build can fail because function declaration or definition in the C++ header will turn into a macro invocation. I faced this problem building Tarantool on MacOS with SDK of 14.4 version.

### DIFF
--- a/test/unit/rtree_multidim.cc
+++ b/test/unit/rtree_multidim.cc
@@ -2,12 +2,14 @@
 #include <time.h>
 #include <algorithm>
 
-#include "unit.h"
 #include "salad/rtree.h"
 #include "../../src/lib/salad/rtree.h"
 
 #include <vector>
 #include <set>
+
+#include "unit.h"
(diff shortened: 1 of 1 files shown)


sergepetrenko (approved):
LGTM. Please, reassign me once you get an approval from devX

ochaplashkin (approved):

sergepetrenko:
@drewdzzz, please also add the 'full-ci' lable after assigning the PR on me for merging.

sergepetrenko:
Backports:
2.11 https://github.com/tarantool/tarantool/pull/9886
3.0 https://github.com/tarantool/tarantool/pull/9887

PR #9890, opened 2024-03-28
httpc: HTTP version option for HTTP client

### LINKED ISSUE #9806: http client: support non-TLS HTTP 2 connection without upgrade from HTTP 1.1
I'm using the built-in HTTP client to issue a request to etcd. The client uses HTTP 1.1 for connections without TLS (can upgrade to HTTP 2) and HTTP 2 for connections with TLS.

However, etcd interprets HTTP 1.1 requests using GRPC-JSON gateway. It doesn't detect that I'm sending a protocol buffer message, not JSON.

So, it is impossible to issue a protocol buffer request to etcd that serves non-TLS connections using the built-in HTTP client. (As for me, etcd without TLS is useful for testing matters, because it is easy to setup/startup.)

OTOH, libcurl has the [CURLOPT_HTTP_VERSION](https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html) option that controls, whether to use HTTP 1.1 or HTTP 2.

I propose to add a Lua API that allows to set this option to `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`.

This is how I understood the values of the libcurl's option.

| Option                              | HTTP version (if `https://`) | HTTP version (if `http://`) |
| ----------------------------------- | ---------------------------- | --------------------------- |
| CURL_HTTP_VERSION_1_0               | HTTP 1.0                     | HTTP 1.0                    |
| CURL_HTTP_VERSION_1_1               | HTTP 1.1                     | HTTP 1.1                    |
| CURL_HTTP_VERSION_2                 | HTTP 2                       | HTTP 2 [using upgrade][1]   |
| CURL_HTTP_VERSION_2TLS (default)    | HTTP 2, then fallback to HTTP 1.1 | HTTP 1.1, then upgrade to HTTP 2 |
| CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE | HTTP 2                       | HTTP 2                      |

[1]: https://curl.se/docs/manpage.html#--http2

As far as I can judge I need an ability to use the HTTP client with `CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE`. I've tried to set it manually in the source code and it works for me.

I'm not sure how it is better to express these values in the Lua API. At least, it would not be bad, if we reflect it directly.

```lua
local http_client = require('http.client').new()

http_client:post(uri, data, {
    http_version = '2-prior-knowledge',
})
```

Let's name the values this way:

* `'1.0'`
* `'1.1'`
* `'2'`
* `'2-tls'`
* `'2-prior-knowledge'`

And raise an error if the given value is unknown.

Col-Waltz:
The `http_version` option allows to set version of http in functions like
`request()`, `post()`, `get()` and etc using this curl option:
https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
The http_version takes one of the strings listed below or a nil value
and sets the appropriate protocol version.

List of supported values:
1) nil - works as '2-tls' by default.
2) '1.1' - sets HTTP/1.1 version.
3) '2' - sets HTTP/2 version that falls back to HTTP/1.1 if cannot be
   negotiated with the server.
4) '2-tls' - works as '2' for HTTPS and as '1.1' for plain text HTTP.
5) '2-prior-knowledge' sets HTTP/2 version without HTTP/1.1 at back.
   It requires prior knowledge that the server supports HTTP/2.

Closes #9806

### DIFF
--- a/changelogs/unreleased/gh-9806-http-set-version.md
+++ b/changelogs/unreleased/gh-9806-http-set-version.md
@@ -1,0 +1,4 @@
+## feature/lua/http client
+
+* Added the `http_version` option to the `request()`, `post()`,
+  `get()` and other request functions from the `http.client` module (gh-9806).
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -464,6 +464,29 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
 #endif
 }
 
+int
+httpc_set_http_version(struct httpc_request *req, const char *version)
+{
+	if (strcmp(version, "1.1") == 0) {
+		curl_easy_setopt(req->curl_request.easy, CURLOPT_HTTP_VERSION,
+				 (long)CURL_HTTP_VERSION_1_1);
+	} else if (strcmp(version, "2") == 0) {
+		curl_easy_setopt(req->curl_request.easy, CURLOPT_HTTP_VERSION,
+				 (long)CURL_HTTP_VERSION_2);
+	} else if (strcmp(version, "2-tls") == 0) {
+		curl_easy_setopt(req->curl_request.easy, CURLOPT_HTTP_VERSION,
+				 (long)CURL_HTTP_VERSION_2TLS);
+	} else if (strcmp(version, "2-prior-knowledge") == 0) {
+		curl_easy_setopt(req->curl_request.easy, CURLOPT_HTTP_VERSION,
+				 (long)CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE);
+	} else {
+		diag_set(IllegalParams, "Not a supported http version: %s",
+			 version);
+		return -1;
+	}
+	return 0;
+}
+
 /**
  * The callback to call after a CURL request is completed.
  */
--- a/src/httpc.h
+++ b/src/httpc.h
@@ -440,6 +440,16 @@ void
 httpc_set_accept_encoding(struct httpc_request *req, const char *encoding);
 
 /**
+ * Specify http protocol version for request
+ * @param req request
+ * @param version - http version to use for this request
+ * Version can be 1.1, 2, 2-tls or 2-prior-knowledge
+ * @see https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
+ */
+int
+httpc_set_http_version(struct httpc_request *req, const char *version);
+
+/**
  * Enable a chunked io interface for the request. It allows to
  * send and receive data via chunks.
  *
--- a/src/lua/httpc.lua
+++ b/src/lua/httpc.lua
@@ -525,6 +525,8 @@ end
 --
 --      accept_encoding - enables automatic decompression of HTTP responses;
 --
+--      http_version - describes an http version;
+--
 --      chunked - enables chunked io interface;
 --
 --      params - a table with query parameters;
--- a/test/app-luatest/http_client_http_version_test.lua
+++ b/test/app-luatest/http_client_http_version_test.lua
@@ -1,0 +1,94 @@
+-- This tests are using treegen to catch information about protocol version
+-- by catching stderr. Treegen creates and tracks temporary directories, where
+-- scripts with testing workloads reside.
+
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+local socket = require('socket')
+
+local wrong_version_group = t.group()
+
+local http_version_group = t.group('versions', {
+    { proto = 'http', version = nil },
+    { proto = 'http', version = '1.1' },
+    { proto = 'http', version = '2' },
+    { proto = 'http', version = '2-tls' },
+    { proto = 'http', version = '2-prior-knowledge' },
+    { proto = 'https', version = nil },
+    { proto = 'https', version = '1.1' },
+    { proto = 'https', version = '2' },
+    { proto = 'https', version = '2-tls' },
+    { proto = 'https', version = '2-prior-knowledge' },
+})
+
+local expected = {
+    ['nil'] = {
+        http = 'GET / HTTP/1.1',
+        https = 'ALPN: curl offers h2,http/1.1',
+    },
+    ['1.1'] = {
+        http = 'GET / HTTP/1.1',
+        https = 'ALPN: curl offers http/1.1',
+    },
+    ['2'] = {
+        http = 'Connection: Upgrade, HTTP2-Settings',
+        https = 'ALPN: curl offers h2,http/1.1',
+    },
+    ['2-tls'] = {
+        http = 'GET / HTTP/1.1',
+        https = 'ALPN: curl offers h2,http/1.1',
+    },
+    ['2-prior-knowledge'] = {
+        http = '[HTTP/2] [1] OPENED stream',
+        https = 'ALPN: curl offers h2,http/1.1',
+    },
+}
+
+http_version_group.before_all(treegen.init)
+http_version_group.after_all(treegen.clean)
+
+http_version_group.before_each(function(g)
+    g.server = socket.tcp_server('127.0.0.1', 0, function(_s) end)
+end)
+
+http_version_group.after_each(function(g)
+    g.server:close()
+end)
+
+local http_request_script = string.dump(function()
+    local http_client = require('http.client').new()
+    local proto = os.getenv('PROTO')
+    local port = os.getenv('HTTP_SERVER_PORT')
+    local uri = proto .. '://127.0.0.1:' .. port
+    local ok = pcall(http_client.get, http_client, uri, {
+        verbose = true,
+        http_version = os.getenv('HTTP_VERSION'),
+    })
+    -- The server accepts the connection and closes it immediately so
+    -- the http request is expected to fail.
+    assert(not ok)
+end)
+
+http_version_group.test_http_version = function(g)
+    local dir = treegen.prepare_directory(g, {}, {})
+    local script_name = 'http_request.lua'
+    treegen.write_script(dir, script_name, http_request_script)
+    local args = {script_name}
+    local opts = {nojson = true, stderr = true}
+    local env = {HTTP_SERVER_PORT = g.server:name().port}
+    env.PROTO = g.params.proto
+    env.HTTP_VERSION = g.params.version
+    local res = justrun.tarantool(dir, env, args, opts)
+    local exp = expected[tostring(g.params.version)][g.params.proto]
+    t.assert_str_contains(res.stderr, exp)
+end
+
+wrong_version_group.test_request_wrong_http_version = function()
+    local http_client = require('http.client').new()
+    local exp_err = 'Not a supported http version: Not an http version'
+    local uri = 'http://127.0.0.1:8080'
+    t.assert_error_msg_equals(exp_err, function()
+        http_client:get(uri, {http_version = 'Not an http version'})
+    end)
+end
--- a/src/lua/httpc.c
+++ b/src/lua/httpc.c
@@ -364,6 +364,15 @@ luaT_httpc_request(lua_State *L)
 		httpc_set_accept_encoding(req, lua_tostring(L, -1));
 	lua_pop(L, 1);
 
+	lua_getfield(L, 5, "http_version");
+	if (!lua_isnil(L, -1)) {
+		if (httpc_set_http_version(req, lua_tostring(L, -1))) {
+			httpc_request_delete(req);
+			return luaT_error(L);
+		}
+	}
+	lua_pop(L, 1);
+
 	bool chunked = false;
 	lua_getfield(L, 5, "chunked");
 	if (!lua_isnil(L, -1) && lua_isboolean(L, -1))


changelogs/unreleased/gh-9806-http-set-version.md
```diff
@@ -0,0 +1,3 @@
+## feature/lua/http client
```
Totktonada, line 1: Please, add a documentation request into the commit message (see examples in other commits).

changelogs/unreleased/gh-9806-http-set-version.md
```diff
@@ -0,0 +1,3 @@
...
+
+* Added an option to set http version.
```
Totktonada, line 3: Please, be certain: 'added the XXX option to the YYY function' would be better. Also, please, add an issue number (see other changelog entries for examples).

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
 #endif
 }
 
+int
+httpc_set_http_version(struct httpc_request *req, const char *version)
+{
+	if (strcmp(version, "1.0") == 0){
```
Totktonada, line 470: Nit: This code formatting is unusual for the project (`){` should be `) {`). Please, check the code around for examples.

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2, version);
+        }
+	if (strcmp(version, "2-tls") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2TLS, version);
+        }
+	if (strcmp(version, "2-prior-knowledge") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE, version);
```
Totktonada, line 488: Nit: Space/tab mix.

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+        }
+	else{
+		diag_set(IllegalParams, "Not a supported http version");
+		return -1;
+	}
```
Totktonada, line 493: AFAIU, this else-branch is called on any value that is not equal to `2-prior-knowledge`. For a valid one too: for example, for `1.1`.

src/lua/httpc.lua
```diff
@@ -529,6 +529,8 @@ end
 --
 --      params - a table with query parameters;
 --
+--      http_version - describes a http version;
```
Totktonada, line 532: Nit: 'a http' -> 'an http'.

https://www.pristineword.com/grammar-an-http/

changelogs/unreleased/gh-9806-http-set-version.md
Totktonada, line 1: Regarding the commit message: `Needed for #xxxx` should be `Fixes #xxxx` or `Closes #xxxx`. The problem described in the issue is solved, right?

https://github.blog/2013-01-22-closing-issues-via-commit-messages/

src/lua/httpc.lua
Totktonada, line 532: Nit: `accept_encoding` looks as the latest added request option except one that enables a specific behavior like `chunked`.

I would place this comment right after the `accept_encoding` so.

Also, I would place the `lua_getfield(L, 5, "http_version");` block after the `lua_getfield(L, 5, "accept_encoding");` block, the `httpc_set_http_version` function declaration after the `httpc_set_accept_encoding` function declaration.

Such consistency helps with a code readability.

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+	if (strcmp(version, "2-tls") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2TLS, version);
+        }
+	if (strcmp(version, "2-prior-knowledge") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE, version);
+        }
+	else{
+		diag_set(IllegalParams, "Not a supported http version");
```
Totktonada, line 491: I would add the value into the error. It helps wth debugging.

Example:

https://github.com/tarantool/tarantool/blob/655c144e7e0a556fe9a8f45c2a9e900d351623cb/src/box/lua/key_def.c#L183-L184

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
 #endif
 }
 
+int
+httpc_set_http_version(struct httpc_request *req, const char *version)
+{
+	if (strcmp(version, "1.0") == 0){
+		curl_easy_setopt(req->curl_request.easy, 
+			CURL_HTTP_VERSION_1_0, version);
```
Totktonada, line 472: The option is `CURLOPT_HTTP_VERSION`, its value is `CURL_VERSION_1_0`.

```c
curl_easy_setopt(req->curl_request.easy, CURLOPT_HTTP_VERSION,
		 CURL_VERSION_1_0);
```

src/httpc.c
Totktonada: IOW, we need just do the if-else if-else if-else chain, not separate if statements.

changelogs/unreleased/gh-9806-http-set-version.md
Totktonada, line 1: Please, add a subsystem prefix into the commit message subject. For example, `httpc:`. The same for the pull request title.

src/lua/httpc.lua
ligurio, line 532: This option is not tested at all.

I propose to check at least a version that http client is enforce in request:
Something like this:

1st console:
```
[0] ~$ netcat -l 127.0.0.1 8080
HEAD / HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: curl/7.81.0
Accept: */*
...
```

2nd console:
```
[0] ~$ curl -I 127.0.0.1:8080
```

This will not test support of different protocol versions, but will test that protocol version enforced by client is correct.

We can implement TCP echo server using builtin `socket` module in Tarantool, https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-tcp-server

src/httpc.c
Col-Waltz: Fixed

src/httpc.c
Col-Waltz: Fixed

src/httpc.c
Col-Waltz: Fixed by using else-if chain

src/lua/httpc.lua
Col-Waltz: Fixed

Col-Waltz: Fixed

src/httpc.c
Col-Waltz: Fixed with "Not a supported http version: %s", version

src/httpc.c
Col-Waltz: Fixed with 
'''C
curl_easy_setopt(req->curl_request.easy,
        CURLOPT_HTTP_VERSION, (long)CURL_HTTP_VERSION_1_0); 
'''
according to https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Documentation request added

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

Col-Waltz: Fixed

Totktonada: Thanks!

Minor nits:

* I would mention that the option also appears in all the methods specialized to a request type: `:get()`, `:post()` and so on.
* I would lease a link to the curl's option.

changelogs/unreleased/gh-9806-http-set-version.md
Totktonada: I'll quote and comment.

> * Added the http-version option to the
>   `require()` function from `http` module (gh-9806)

1. `require()` -> `request()` (and `post()`, `get()`, etc)
2. `http-version` -> `http_version`
3. Unusual formatting: the line break after ~40 characters, not close to 80 (use `gqap` if you're using Vim).
4. Add a period a the end.

changelogs/unreleased/gh-9806-http-set-version.md
Totktonada: And the pull request title, please.

src/httpc.c
```diff
@@ -464,6 +464,36 @@ httpc_set_accept_encoding(struct httpc_request *req, const char *encoding)
...
+                        CURL_HTTP_VERSION_2, version);
+        }
+	if (strcmp(version, "2-tls") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2TLS, version);
+        }
+	if (strcmp(version, "2-prior-knowledge") == 0){
+                curl_easy_setopt(req->curl_request.easy,
+                        CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE, version);
+        }
```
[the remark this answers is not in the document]
Totktonada: It was my proposal: just provide what libcurl provides.

I see no a real use case for HTTP/1.0.

We can eliminate it, I don't mind. Or lease as it, I don't mind too.

@ligurio Do you propose to remove it?

src/lua/httpc.lua
Totktonada: Nice idea.

After your message I've thought on it a bit more and remembered that there is the `verbose = true` option, which writes request details on stderr. We can run tarantool using the `test.justrun` helper with a script that runs an HTTP request. Then, analyze the stderr output in a test case.

Maybe this variant doesn't require an alive listening socket at all. At least, it worth to try.

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed. Added a link

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

changelogs/unreleased/gh-9806-http-set-version.md
Totktonada: 5. `http` module. (gh-xxxx) -> `http.client` module (gh-xxxx).

(The module name and the period position.)

Sorry, overlooked it on the first glance.

src/lua/httpc.lua
Totktonada: I've experimented a bit with it: we still need a TCP server, but it is enough to have a simple one, which just accepts and closes a connection.

It looks like we can differentiate all the needed situations using the verbose output:

* `http://`: HTTP 1.1 vs HTTP/1.1 + Upgrade to HTTP/2 vs HTTP/2
* `https://`: HTTP/1.1 vs HTTP/2

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

Totktonada: Sorry for nitpicking, but, please, adjust the period position as well.

```diff
 * Added the `http_version` option to the `request()` (and `post()`,
-  `get()`, etc) functions from `http.client` module. (gh-9806)
+  `get()`, etc) functions from `http.client` module (gh-9806).
```

(Also, I think that it should be 'the `http.client` module', with the definite article.)

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,100 @@
...
+    treegen.write_script(dir, script_name, http_request_script)
+
+    local args = {script_name}
+    local opts = {nojson = true, stderr = true}
+    local env = {HTTP_SERVER_PORT = g.server:name().port}
+    env.PROTO = 'http'
+    env.HTTP_VERSION = '3.0'
+    local res = justrun.tarantool(dir, env, args, opts)
+    t.assert_equals(res.exit_code, 0)
+end
```
Totktonada, line 100: We don't need to catch stderr output from the `verbose` option here, so it is not necessary to start the HTTP client in a separate process.

I would do something like the following.

```lua
local http_client = require('http.client').new()
local exp_err = <...>
t.assert_error_msg_equals(exp_err, function()
    http_client:get(<...>, {http_version = <...>})
end)
```

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,100 @@
...
+g.before_all(treegen.init)
+g.after_all(treegen.clean)
+
+g.before_each(function(g)
+   g.server = socket.tcp_server('127.0.0.1', 0, function(_s) end)
+end)
+
+g.after_each(function(g)
+   g.server:close()
+end)
```
Totktonada, line 17: Nit: three spaces indent (instead of 4).

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,100 @@
...
+
+local http_request_script = string.dump(function()
+    local http_client = require('http.client').new()
+    local proto = os.getenv('PROTO')
+    local port = os.getenv('HTTP_SERVER_PORT')
+    local uri = proto .. '://127.0.0.1:' .. port
+    pcall(http_client.get, http_client, uri, {
+            verbose = true,
+            http_version = os.getenv('HTTP_VERSION'),
```
Totktonada, line 26: Nit: 8 spaces indent (instead of 4).

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed)

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: Fixed

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: Fixed, but luacheck didn't say anything

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: Fixed

src/lua/httpc.lua
Totktonada: Ilya added the tests. Sergey, do you want to glance on them?

changelogs/unreleased/gh-9806-http-set-version.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua/http client
+
+* Added the `http_version` option to the `request()` (and `post()`,
+  `get()`, etc) functions from the `http.client` module (gh-9806).
```
andreyaksenov, line 4:
```suggestion
* Added the `http_version` option to the `request()`, `post()`,
  `get()`, and other request functions from the `http.client` module (gh-9806).
```

andreyaksenov (approved):

Totktonada (approved):

src/httpc.c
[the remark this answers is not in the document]
ligurio: > @ligurio Do you propose to remove it?

Right, it is better to remove it.

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,95 @@
...
+    env.HTTP_VERSION = '2'
+    local res = justrun.tarantool(dir, env, args, opts)
+    t.assert_str_contains(res.stderr, 'ALPN: curl offers h2,http/1.1')
+    env.HTTP_VERSION = '2-tls'
+    local res = justrun.tarantool(dir, env, args, opts)
+    t.assert_str_contains(res.stderr, 'ALPN: curl offers h2,http/1.1')
+    env.HTTP_VERSION = '2-prior-knowledge'
+    local res = justrun.tarantool(dir, env, args, opts)
+    t.assert_str_contains(res.stderr, 'ALPN: curl offers h2,http/1.1')
+end
```
ligurio, line 86: Please rewrite testcases using luatest parameterization, see example here - https://github.com/tarantool/tarantool/blob/58f5ffe43bffa0901d8d577bbb4d523c4e7e682a/test/app-luatest/uri_unit_test.lua#L11C7-L33

About parameterization in luatest doc: https://github.com/tarantool/luatest/?tab=readme-ov-file#parametrization

the main idea - to split test data and test code

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,95 @@
...
+g.after_each(function(g)
+    g.server:close()
+end)
+
+local http_request_script = string.dump(function()
+    local http_client = require('http.client').new()
+    local proto = os.getenv('PROTO')
+    local port = os.getenv('HTTP_SERVER_PORT')
+    local uri = proto .. '://127.0.0.1:' .. port
+    pcall(http_client.get, http_client, uri, {
```
ligurio, line 24: You didn't check `pcall` ret code. Why?

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,95 @@
...
+        verbose = true,
+        http_version = os.getenv('HTTP_VERSION'),
+    })
+end)
+
+g.test_http_version = function()
+    local dir = treegen.prepare_directory(g, {}, {})
+    local script_name = 'http_request.lua'
+    treegen.write_script(dir, script_name, http_request_script)
```
ligurio, line 33: To be honestly, I don't get why do you need `treegen` here.
After removing `treegen` env variables will be gone too:

```lua
 local http_req = string.dump(function(port, version)
    local http_client = require('http.client').new()
    local uri = 'http://127.0.0.1:' .. port
    pcall(http_client.get, http_client, uri, {
        verbose = true,
        http_version = version,
    })
end)

assert(loadstring(http_req))('35249', '1.0')
```

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,95 @@
...
+
+    local args = {script_name}
+    local opts = {nojson = true, stderr = true}
+    local env = {HTTP_SERVER_PORT = g.server:name().port}
+    env.PROTO = 'http'
```
ligurio, line 38: What is a point to parameterize PROTO if it is always constant?

src/lua/httpc.lua
ligurio: added comments regarding tests

test/app-luatest/gh_9806_http_client_version_test.lua
Totktonada: This request will definitely fail, because our `socket.tcp_server` server just closes a connection after accepting it. I agree that it worth to comment it here.

test/app-luatest/gh_9806_http_client_version_test.lua
Totktonada: The main problem is to catch `stderr`.

We can perform the usual `dup/dup2/pipe` catching from the same process. However, luatest performs something like this on its own and I don't want to clash here. In theory, the double wrapping should work, but it seems dangerous for me.

So, we need a separate process. It means some `popen` (we use `justrun` wrapper here). That means that we need to run `tarantool some_script.lua`.

We can place this script into the source code or generate in runtime. I prefer the second way, because it allows to keep all the code that relates to a test case in the same file.

Since we generate the file in runtime, we need a temporary directory to store it. We can create it using `fio.tempdir()` and remove using `fio.rmtree()` in an `after_all` hook to don't leave garbage after each run of the testing.

In fact, it is what `treegen` is doing. It creates and tracks temporary directories and provides some helpers for creating scripts in runtime.

test/app-luatest/gh_9806_http_client_version_test.lua
Totktonada: It is `http` or `https`.

changelogs/unreleased/gh-9806-http-set-version.md
Col-Waltz: Fixed

src/httpc.c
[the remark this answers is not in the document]
Col-Waltz: Removed HTTP/1.0 from options and tests

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: Rewrited testcases using luatest parametrization. Met some troubles with `group()` function from `luatest`. If `res` string contains symbol '/' test-run.py can't run this test and gets an error in `group()` function. Asked Oleg about this issue and waiting for an answer tomorrow. 
Could you check the rest of the testcases? In comments I added strings that I want to use for checking. If current strings are enough I can leave them and delete comments.

test/app-luatest/gh_9806_http_client_version_test.lua
```diff
@@ -0,0 +1,73 @@
...
+    { req = 'https', version = nil, res = 'ALPN: curl offers h2,http' },
+    --'ALPN: curl offers http/1.1'
+    { req = 'https', version = '1.1', res = 'ALPN: curl offers http' },
+    --'ALPN: curl offers h2,http/1.1'
+    { req = 'https', version = '2', res = 'ALPN: curl offers h2,http' },
+    --'ALPN: curl offers h2,http/1.1'
+    { req = 'https', version = '2-tls', res = 'ALPN: curl offers h2,http' },
+    --'ALPN: curl offers h2,http/1.1'
+    { req = 'https', version = '2-prior-knowledge',
+      res = 'ALPN: curl offers h2,http' },
```
ligurio, line 28: Please add whitespace after dashes and dots at the end of comments.

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: Added a check and a comment

test/app-luatest/gh_9806_http_client_version_test.lua
Col-Waltz: These comments were temporary, I've already deleted them

src/httpc.h
```diff
@@ -439,6 +439,16 @@ httpc_set_follow_location(struct httpc_request *req, long follow);
 void
 httpc_set_accept_encoding(struct httpc_request *req, const char *encoding);
 
+/**
+ * Specify http protocol version for request
+ * @param req request
+ * @param version - http version to use for this request
+ * Version can be 1.0, 1.1, 2, 2-tls or 2-prior-knowledge
```
ligurio, line 446: we decided to remove http version 1.0, please update comment

test/app-luatest/http_client_http_version_test.lua
```diff
@@ -0,0 +1,94 @@
+-- This tests are using treegen to catch information about protocol version
+-- by catching stderr. Treegen creates and tracks temporary directories, where
+-- scripts with testing workloads reside.
+
+local t = require('luatest')
+local treegen = require('test.treegen')
+local justrun = require('test.justrun')
+local socket = require('socket')
+
+local wrong_version_group = t.group('wrong version')
```
ligurio, line 10: You don't need a parametrization group for a single testcase.

test/app-luatest/http_client_http_version_test.lua
```diff
@@ -0,0 +1,94 @@
...
+    local exp = expected[tostring(g.params.version)][g.params.proto]
+    t.assert_str_contains(res.stderr, exp)
+end
+
+wrong_version_group.test_request_wrong_http_version = function()
+    local http_client = require('http.client').new()
+    local exp_err = 'Not a supported http version: 3.0'
+    local uri = 'http://127.0.0.1:8080'
+    t.assert_error_msg_equals(exp_err, function()
+        http_client:get(uri, {http_version = '3.0'})
```
ligurio, line 92: I would replace '3.0' to something that will never be a real http version. For example: 'not existent http version' or similar. HTTP 3.0 is a real http version, but currently it is not enabled in tarantool's http client. Perhaps, someday, we will enable it and this testcase will be broken.

test/app-luatest/gh_9806_http_client_version_test.lua
ligurio: Seems mostly fine, but left minor comments

src/httpc.h
Col-Waltz: Fixed

test/app-luatest/http_client_http_version_test.lua
Col-Waltz: Replaced with 'Not an http version'

test/app-luatest/http_client_http_version_test.lua
```diff
@@ -0,0 +1,94 @@
...
+    local exp = expected[tostring(g.params.version)][g.params.proto]
+    t.assert_str_contains(res.stderr, exp)
+end
+
+wrong_version_group.test_request_wrong_http_version = function()
+    local http_client = require('http.client').new()
+    local exp_err = 'Not a supported http version: 3.0'
+    local uri = 'http://127.0.0.1:8080'
+    t.assert_error_msg_equals(exp_err, function()
+        http_client:get(uri, {http_version = 'Not an http version'})
```
ligurio, line 92: Not: `an` -> `a`

ligurio (approved):
Ilya, thanks for the patch! LGTM

test/app-luatest/http_client_http_version_test.lua
Col-Waltz: I've had same fix from a -> an according to this paper https://www.pristineword.com/grammar-an-http/

ligurio: Okay.

PR #9892, opened 2024-03-29
Introduce RPC for roles

### LINKED ISSUE #9842: Introduce RPC for roles
Currently, it is quite problematic to execute a function provided by a role on one instance from another instance. We need to provide a way to do this.

ImeevMA:
This patch-set introduces RPC for roles from the same config.

Closes #9842

### DIFF
--- a/changelogs/unreleased/gh-9842-rpc-for-roles.md
+++ b/changelogs/unreleased/gh-9842-rpc-for-roles.md
@@ -4,3 +4,5 @@
   `connect()` function of this module (gh-9842).
 * Introduced the `filter()` function in the `experimental.connpool`
   module (gh-9842).
+* Introduced the `call()` function in the `experimental.connpool`
+  module (gh-9842).
--- a/changelogs/unreleased/gh-9842-rpc-for-roles.md
+++ b/changelogs/unreleased/gh-9842-rpc-for-roles.md
@@ -1,0 +1,4 @@
+## feature/experimental
+
+* Introduced the `experimental.connpool` module and the
+  `connect()` function of this module (gh-9842).
--- a/changelogs/unreleased/gh-9842-rpc-for-roles.md
+++ b/changelogs/unreleased/gh-9842-rpc-for-roles.md
@@ -2,3 +2,5 @@
 
 * Introduced the `experimental.connpool` module and the
   `connect()` function of this module (gh-9842).
+* Introduced the `filter()` function in the `experimental.connpool`
+  module (gh-9842).
--- a/src/box/lua/experimental/connpool.lua
+++ b/src/box/lua/experimental/connpool.lua
@@ -55,6 +55,64 @@ local function connect(instance_name, opts)
     return conn
 end
 
+local function is_roles_match(expected_roles, present_roles)
+    if expected_roles == nil or next(expected_roles) == nil then
+        return true
+    end
+    if present_roles == nil or next(present_roles) == nil then
+        return false
+    end
+
+    local roles = {}
+    for _, present_role_name in pairs(present_roles) do
+        roles[present_role_name] = true
+    end
+    for _, expected_role_name in pairs(expected_roles) do
+        if roles[expected_role_name] == nil then
+            return false
+        end
+    end
+    return true
+end
+
+local function is_labels_match(expected_labels, present_labels)
+    if expected_labels == nil or next(expected_labels) == nil then
+        return true
+    end
+    if present_labels == nil or next(present_labels) == nil then
+        return false
+    end
+
+    for label, value in pairs(expected_labels) do
+        if present_labels[label] ~= value then
+            return false
+        end
+    end
+    return true
+end
+
+local function is_candidate_match(instance_name, opts)
+    assert(opts ~= nil and type(opts) == 'table')
+    local get_opts = {instance = instance_name}
+    return is_roles_match(opts.roles, config:get('roles', get_opts)) and
+           is_labels_match(opts.labels, config:get('labels', get_opts))
+end
+
+local function filter(opts)
+    checks({
+        labels = '?table',
+        roles = '?table',
+    })
+    local candidates = {}
+    for instance_name in pairs(config:instances()) do
+        if is_candidate_match(instance_name, opts or {}) then
+            table.insert(candidates, instance_name)
+        end
+    end
+    return candidates
+end
+
 return {
     connect = connect,
+    filter = filter,
 }
--- a/src/box/lua/experimental/connpool.lua
+++ b/src/box/lua/experimental/connpool.lua
@@ -112,7 +112,80 @@ local function filter(opts)
     return candidates
 end
 
+local function get_connection(all_candidates, prefer_local)
+    local candidates = table.copy(all_candidates)
+    if prefer_local ~= false then
+        local candidate_idx = nil
+        for n, candidate in ipairs(candidates) do
+            if candidate == box.info.name then
+                candidate_idx = n
+                local conn = connect(box.info.name, {wait_connected = false})
+                if conn:wait_connected() then
+                    return conn
+                end
+                break
+            end
+        end
+        if candidate_idx ~= nil then
+            table.remove(candidates, candidate_idx)
+        end
+    end
+
+    while #candidates > 0 do
+        local n = math.random(#candidates)
+        local instance_name = table.remove(candidates, n)
+        local conn = connect(instance_name, {wait_connected = false})
+        if conn:wait_connected() then
+            return conn
+        end
+    end
+    return nil
+end
+
+local function call(func_name, args, opts)
+    checks('string', '?table', {
+        labels = '?table',
+        roles = '?table',
+        prefer_local = '?boolean',
+        -- The following options passed directly to net.box.call().
+        timeout = '?',
+        buffer = '?',
+        on_push = '?function',
+        on_push_ctx = '?',
+        is_async = '?boolean',
+    })
+    opts = opts or {}
+
+    local candidates_opts = {
+        labels = opts.labels,
+        roles = opts.roles,
+    }
+    local candidates = filter(candidates_opts)
+    if next(candidates) == nil then
+        local msg = "Couldn't execute function %s: no candidates are " ..
+                    "available with these conditions"
+        error(msg:format(func_name), 0)
+    end
+
+    local conn = get_connection(candidates, opts.prefer_local)
+    if conn == nil then
+        local msg = "Couldn't execute function %s: connection to " ..
+                    "candidates failed"
+        error(msg:format(func_name), 0)
+    end
+
+    local net_box_call_opts = {
+        timeout = opts.timeout,
+        buffer = opts.buffer,
+        on_push = opts.on_push,
+        on_push_ctx = opts.on_push_ctx,
+        is_async = opts.is_async,
+    }
+    return conn:call(func_name, args, net_box_call_opts)
+end
+
 return {
     connect = connect,
     filter = filter,
+    call = call,
 }
--- a/src/box/lua/experimental/connpool.lua
+++ b/src/box/lua/experimental/connpool.lua
@@ -1,0 +1,60 @@
+local config = require('config')
+local checks = require('checks')
+local netbox = require('net.box')
+
+local connections = {}
+
+local function is_connection_valid(conn, opts)
+    if conn == nil or conn.state == 'error' or conn.state == 'closed' then
+        return false
+    end
+    assert(type(opts) == 'table')
+    local conn_opts = conn.opts or {}
+    if opts.fetch_schema ~= false and conn_opts.fetch_schema == false then
+        return false
+    end
+    return true
+end
+
+local function connect(instance_name, opts)
+    checks('string', {
+        connect_timeout = '?number',
+        wait_connected = '?boolean',
+        fetch_schema = '?boolean',
+    })
+    opts = opts or {}
+
+    local conn = connections[instance_name]
+    if not is_connection_valid(conn, opts) then
+        local uri = config:instance_uri('peer', {instance = instance_name})
+        if uri == nil then
+            local err = 'No suitable URI provided for instance %q'
+            error(err:format(instance_name), 0)
+        end
+
+        local conn_opts = {
+            connect_timeout = opts.connect_timeout,
+            wait_connected = false,
+            fetch_schema = opts.fetch_schema,
+        }
+        local ok, res = pcall(netbox.connect, uri, conn_opts)
+        if not ok then
+            local msg = 'Unable to connect to instance %q: %s'
+            error(msg:format(instance_name, res.message), 0)
+        end
+        conn = res
+        connections[instance_name] = conn
+    end
+
+    -- If opts.wait_connected is not false we wait until the connection is
+    -- established or an error occurs (including a timeout error).
+    if opts.wait_connected ~= false and conn:wait_connected() == false then
+        local msg = 'Unable to connect to instance %q: connection timeout'
+        error(msg:format(instance_name), 0)
+    end
+    return conn
+end
+
+return {
+    connect = connect,
+}
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -170,16 +170,16 @@ extern char session_lua[],
 	config_utils_odict_lua[],
 	config_utils_schema_lua[],
 	config_utils_snapshot_lua[],
-	config_utils_tabulate_lua[]
+	config_utils_tabulate_lua[],
 #if ENABLE_CONFIG_EXTRAS
-	,
 	config_source_etcd_lua[],
 	config_storage_init_lua[],
 	config_source_storage_lua[],
-	config_extras_lua[]
+	config_extras_lua[],
 #endif
-	;
 	/* }}} config */
+
+	experimental_connpool_lua[];
 
 /**
  * List of box's built-in modules written using Lua.
@@ -454,6 +454,10 @@ static const char *lua_sources[] = {
 #endif
 
 	/* }}} config */
+
+	"experimental/connpool",
+	"experimental.connpool",
+	experimental_connpool_lua,
 
 	FAILOVER_LUA_MODULES
 
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -76,6 +76,8 @@ if (ENABLE_CONFIG_EXTRAS)
 endif()
 # }}} config
 
+lua_source(lua_sources lua/experimental/connpool.lua      experimental_connpool_lua)
+
 if (ENABLE_FAILOVER)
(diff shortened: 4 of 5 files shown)


src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,161 @@
...
+        local conn = connect_using_config(instance_name, candidate_opts)
+        if conn:wait_connected() then
+            return conn
+        end
+    end
+    error('Connection to candidate instances failed', 0)
+end
+
+local function call_role(role_name, func_name, args, opts)
+    checks('string', 'string', '?table', {
```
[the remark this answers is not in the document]
ImeevMA: I was planning to add them a little later as I find it a little weird to add them directly to the opts. But I have no problem adding another patch for them here.

src/box/lua/config/init.lua
```diff
@@ -578,6 +578,34 @@ function methods.instances(self)
...
+    end
+    if opts.show_password ~= nil and type(opts.show_password) ~= 'boolean' then
+        error(('Expected boolean, got %s'):format(opts.show_password), 0)
+    end
+
+    local uri_opts = {
+        instance = opts.instance,
+        use_default = true,
+    }
+    local uri = table.copy(self._configdata:_instance_uri(uri_type, uri_opts))
```
Totktonada, line 602: Shouldn't it be `_configdata_applied` for consistency with `config:get()`?

changelogs/unreleased/gh-9842-rpc-for-roles.md
```diff
@@ -0,0 +1,9 @@
+## feature/config
+
+* Introduced `instance_uri()` method for `config` module (gh-9842).
+* Introduced module `experimental.connpool` and function
+  `connect_using_config()` (gh-9842).
+* Introduced function `get_candidates()` from `experimental.connpool`
+  module (gh-9842).
+* Introduced function `call_role()` from `experimental.connpool`
+  module (gh-9842).
```
andreyaksenov, line 9:
```suggestion
* Introduced the `instance_uri()` method for the `config` module (gh-9842).
* Introduced the `experimental.connpool` module and the
  `connect_using_config()` function (gh-9842).
* Introduced the `get_candidates()` function in the `experimental.connpool`
  module (gh-9842).
* Introduced the `call_role()` function in the `experimental.connpool`
  module (gh-9842).
```

andreyaksenov (approved):

src/box/lua/init.c
```diff
@@ -169,15 +169,15 @@ extern char session_lua[],
...
 	config_source_etcd_lua[],
 	config_storage_init_lua[],
 	config_source_storage_lua[],
-	config_extras_lua[]
+	config_extras_lua[],
 #endif
-	;
+	experimental_connpool_lua[];
+
 	/* }}} config */
```
Totktonada, line 181: Nit: This new item should be outside of the `/* {{{ config */` `/* }}} config */` block.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,59 @@
...
+local function connect_using_config(instance_name, opts)
+    checks('string', {
+        wait_connected = '?boolean|number',
+        fetch_schema = '?boolean',
+    })
+    opts = opts or {}
+
+    local conn = connections[instance_name]
+    if conn == nil or conn.state == 'error' or conn.state == 'closed' or
+       conn.space == nil and opts.fetch_schema then
```
Totktonada, line 16: I'm not sure about this check. First, `opts.fetch_schema` may be `nil` (and it is assumed as `true`). So, it should be `opts.fetch_schema ~= false` at least.

Second, it is possible that a space information is in fly. I guess that it is possible that space information fetching may be slower than retrying of the calls to `connect_using_config`. It means that a new connection will be created each time.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,59 @@
...
+    -- wait_connected == true is not supported.
+    local wait_connected =
+        opts.wait_connected ~= true and opts.wait_connected or nil
+
+    assert(conn ~= nil)
+    local ok, res = pcall(conn.wait_connected, conn, wait_connected)
+    if not ok then
+        local msg = 'Unable to connect to instance %q: %s'
+        error(msg:format(instance_name, res.message), 0)
+    end
```
Totktonada, line 52: First, I think that if something is not supported, it should lead to an error.

Next, it is a bit hard to understand that the `wait_connected` variable contains a timeout for creating a connection. I would check the type and assign it to a variable named as `connect_timeout`.

BTW, the API may be more straightforward if we add it as a separate `connect_timeout` option into the API. It is completely up to you, just my though.

changelogs/unreleased/gh-9842-rpc-for-roles.md
```diff
@@ -1,3 +1,5 @@
 ## feature/config
 
 * Introduced `instance_uri()` method for `config` module (gh-9842).
+* Introduced module `experimental.connpool` and function
+  `connect_using_config()` (gh-9842).
```
Totktonada, line 5: The `config:` prefix in commit messages regarding the connection pool module seems irrelevant. The same for the `## feature/config` changelog category.

changelogs/unreleased/gh-9842-rpc-for-roles.md
```diff
@@ -1,3 +1,5 @@
 ## feature/config
```
Totktonada, line 1: > NO_DOC=experimental feature

We agreed to have some brief documenation for experimental features.

src/box/lua/experimental/connpool.lua
```diff
@@ -54,6 +54,45 @@ local function connect_using_config(instance_name, opts)
...
+    for _, label in pairs(labels) do
+        if instance_labels[label] == nil then
+            return false
+        end
+    end
+    return true
+end
+
+local function get_candidates(role_name, opts)
```
Totktonada, line 72: In a voice discussion with Mergen we agreed that filtering by a role should be one of the filters, not some kind of a primary filter in a separate argument.

So, we likely will be somethings like the following.

```lua
pool.get_candidates({
    role = 'myrole',
    labels = {zone = 'msk'},
})
```

Or the following.

```lua
pool.get_candidates({
    roles = {'myrole1', 'myrole2'},
    labels = {zone = 'msk'},
})
```

src/box/lua/experimental/connpool.lua
```diff
@@ -54,6 +54,45 @@ local function connect_using_config(instance_name, opts)
...
+        return true
+    end
+    if instance_labels == nil then
+        return false
+    end
+    for _, label in pairs(labels) do
+        if instance_labels[label] == nil then
+            return false
+        end
+    end
```
Totktonada, line 68: As I see, we check a presence of the label, but we don't verify that it has a value given by a caller.

It is easy to imagine that one have the `zone` label and wants to filter instances from a specific zone.

I propose to implement the `labels` filter with the following semantic.

The instance `I` is accepted by the `opts.labels` filter if `I.labels` is *covered* by `opts.labels`. It means that each key-value pair from `opts.labels` has the same key-value pair in `I.labels`.

If we will need a filter that verifies a presence of a label without taking into account its value, I would add it later with a name like `has_labels`.

```lua
pool.get_candidates({
    has_labels = {'mylabel1', 'mylabel2'},
})
```

src/box/lua/experimental/connpool.lua
```diff
@@ -54,6 +54,45 @@ local function connect_using_config(instance_name, opts)
...
+        local roles = iconfig.roles or {}
+        if is_labels_match(opts.labels, iconfig.labels) then
+            for _, role in pairs(roles) do
+                if role == role_name then
+                    table.insert(candidates, instance_name)
+                    break
+                end
+            end
+        end
+    end
```
Totktonada, line 91: I would structurize the code in a way that allows to decouple the filters. For example,

```lua
local function filter_by_role(instance_name)
    if opts.role == nil then
        return true
    end
    local roles = config:get('roles', {instance = instance_name})
    return table_has_key(roles, opts.role)
end

local function filter_by_labels(instance_name)
    if opts.labels == nil then
        return true
    end
    local labels = config:get('labels', {instance = instance_name})
    return table_is_covered_by(labels, opts.labels)
end

return fun.iter(config:instances())
    :filter(filter_by_role)
    :filter(filter_by_labels)
    :totable()
```

(Extracting the `table_{has/is}_*` helpers would also increase readability, IMHO.)

(I don't test this code, it is just illustration.)

I don't think that repeated `config:get()` call is a problem. If it is, it is better to implement some caching inside the call.

src/box/lua/experimental/connpool.lua
```diff
@@ -92,7 +92,70 @@ local function get_candidates(role_name, opts)
...
+        end
+        local msg = 'No instances with role %q are available'
+        error(msg:format(role_name), 0)
+    end
+
+    if opts.prefer_local == nil or opts.prefer_local == true then
+        local name = box.info.name
+        for n, candidate_name in pairs(candidates) do
+            if candidate_name == name then
+                table.remove(candidates, n)
```
Totktonada, line 119: We shoudn't modify a table during `pairs()`.

src/box/lua/experimental/connpool.lua
```diff
@@ -92,7 +92,70 @@ local function get_candidates(role_name, opts)
...
+    end
+
+    while #candidates > 0 do
+        local n = math.random(#candidates)
+        local instance_name = table.remove(candidates, n)
+        local candidate_opts = {wait_connected = false}
+        local conn = connect_using_config(instance_name, candidate_opts)
+        if conn:wait_connected() then
+            return conn
+        end
```
Totktonada, line 137: I would let it fail if it doesn't became connected within the given/default connect timeout. It sounds better than an infinite waiting.

src/box/lua/experimental/connpool.lua
Totktonada: BTW, this code would be simpler if we just check `candidates` using a helper like `table_has_key` (that I proposed in comments for one of the previous patches).

src/box/lua/experimental/connpool.lua
Totktonada: Or `conn:wait_connected()` is not infinite?

BTW, if it is infinite, there is no point in the `while` loop here.

src/box/lua/experimental/connpool.lua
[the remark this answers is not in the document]
ImeevMA: @olegrok I added these options.

changelogs/unreleased/gh-9842-rpc-for-roles.md
ImeevMA: Thanks, fixed!

src/box/lua/init.c
ImeevMA: Fixed.

src/box/lua/experimental/connpool.lua
ImeevMA: Changed the condition.

src/box/lua/experimental/connpool.lua
ImeevMA: After discussion we decided to accept only boolean for `wait_connected` and add another numeric option `connect_timeout`. These options are passed directly to `net.box.connect()`. These two options works about the same way as `wait_connected` option for `cartridge.pool.connect()`. There is couple of cases when the behavior is different:
1) If `wait_connected` option for `cartridge.pool.connect()` more than default `connect_timeout` of `net.box.connect()`, than default `connect_timeout` is actually used;
2) If `wait_connected` option for `cartridge.pool.connect()` less than default `connect_timeout` of `net.box.connect()`, it was actually possible that an error returned before the connection is successfully established. This actually seems to be expected behavior, as the error says `Connection not established (yet)`.

changelogs/unreleased/gh-9842-rpc-for-roles.md
ImeevMA: Changed to `experimental`.

src/box/lua/experimental/connpool.lua
ImeevMA: Reworked this part.

src/box/lua/experimental/connpool.lua
```diff
@@ -92,7 +92,70 @@ local function get_candidates(role_name, opts)
...
+        prefer_local = '?boolean',
+    })
+    opts = opts or {}
+
+    local conn = get_connection(role_name, {
+        prefer_local = opts.prefer_local,
+        labels = opts.labels,
+    })
+    assert(conn ~= nil)
+    return conn:call(('%s.%s'):format(role_name, func_name), args)
```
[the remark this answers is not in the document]
ImeevMA: Renamed the function to `call` and moved role to options.

src/box/lua/experimental/connpool.lua
ImeevMA: True, fixed. For now I did not implemented `table_has_key()`.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,64 @@
...
+        wait_connected = '?boolean',
+        fetch_schema = '?boolean',
+    })
+    opts = opts or {}
+
+    local conn = connections[instance_name]
+    if not is_connection_valid(conn, opts) then
+        local uri_opts = {
+            instance = instance_name,
+            show_password = true,
```
Totktonada, line 31: AFAIR, this `:instance_uri()` option is gone (was present in a draft patch version).

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,64 @@
...
+    -- established or an error occurs (including a timeout error).
+    if opts.wait_connected ~= false and conn:wait_connected() == false then
+        local msg = 'Unable to connect to instance %q: connection timeout'
+        error(msg:format(instance_name), 0)
+    end
+    return conn
+end
+
+return {
+    connect_using_config = connect_using_config,
```
Totktonada, line 194: It appears that all the functions in the module takes information from config. So the `_using_config` suffix now looks, hm, inconsistent. It should be either in all the functions or neither.

I propose to drop it and name the function `connect`.

And, while we're here, if we do something with `get_candidates`, all the functions will be one-word-named, which would look nice. For example, `connect`, `filter` and `call`.

Just thoughts. Feel free to ignore.

Totktonada (approved):
It looks OK for me at whole. Thank you for working on it!

There are some thoughts around. Please, consider comments above.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,64 @@
...
+
+local function is_connection_valid(conn, opts)
+    if conn == nil or conn.state == 'error' or conn.state == 'closed' then
+        return false
+    end
+    assert(type(opts) == 'table')
+    local conn_opts = conn.opts or {}
+    if opts.fetch_schema == true and conn_opts.fetch_schema == false then
+        return false
+    end
```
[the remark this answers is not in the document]
ImeevMA: Thanks, fixed.

src/box/lua/experimental/connpool.lua
ImeevMA: True, fixed.

src/box/lua/experimental/connpool.lua
```diff
@@ -116,7 +116,82 @@ local function get_candidates(opts)
...
+        local candidate_opts = {wait_connected = false}
+        local conn = connect_using_config(instance_name, candidate_opts)
+        if conn:wait_connected() then
+            return conn
+        end
+    end
+    return nil
+end
+
+local function call(func_name, args, opts)
```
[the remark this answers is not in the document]
ImeevMA: Added.

src/box/lua/experimental/connpool.lua
ImeevMA: Fixed.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,191 @@
+local config = require('config')
+local checks = require('checks')
+local netbox = require('net.box')
+
+local connections = {}
```
olegrok, line 5: Is there any mechanism to clean this table? It is possible to overflow it somehow during long runtime?

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,191 @@
...
+    local candidates = table.copy(all_candidates)
+    if prefer_local ~= false then
+        local candidate_idx = nil
+        for n, candidate in ipairs(candidates) do
+            if candidate == box.info.name then
+                candidate_idx = n
+                local conn = connect(box.info.name, {wait_connected = false})
+                if conn:wait_connected() then
+                    return conn
+                end
```
olegrok, line 125: Is not it the same as just connect with wait_connected = true?

Totktonada: I had the same question, but then found that if we'll do `connect()` without options, we need to pcall it to try the next candidate.

src/box/lua/experimental/connpool.lua
ImeevMA: At the moment the only way to clear the table is to restart Tarantool. As for overflow, we have standard Lua problems here. What I mean is that if you're using a GC64, you have a maximum of 2^26 keys, otherwise you'll likely hit the 2GB memory limit sooner.

src/box/lua/experimental/connpool.lua
ImeevMA: Also, `connect()` returns an already existing connection, if it exists.

src/box/lua/experimental/connpool.lua
olegrok: The main problem here that you can silently get overflow error and it will be very hard to understand exact reason.

olegrok (approved):

src/box/lua/experimental/connpool.lua
Totktonada: I don't think that we can realistically reach any Lua memory/Lua table size limit here. However, I agree that it would be good to delete items that are not in the config anymore. We can do it in `connect()`.

I wouldn't insist on it in the scope of this PR.

src/box/lua/experimental/connpool.lua
```diff
@@ -0,0 +1,60 @@
...
+
+local function connect(instance_name, opts)
+    checks('string', {
+        connect_timeout = '?number',
+        wait_connected = '?boolean',
+        fetch_schema = '?boolean',
+    })
+    opts = opts or {}
+
+    local conn = connections[instance_name]
```
Totktonada, line 27: Yet another thing that I would solve in a separate PR.

It seems, if instance's URI is changed, we don't re-create a connection.

ImeevMA: I created a new issue about this: https://github.com/tarantool/tarantool/issues/9910

Issue #9895, opened 2024-04-01
config: verify that replicaset contains an instance.


Lord-KA:
There is a quite common mistake, when people misplace a section in config to a different level, creating confusion.
For example, here there is an empty replicaset `sharding` instead of sharding option for all replicasets: 
```yaml
groups:
  g-001:
    replicasets:
      sharding:
        roles:
        - storage
      r-001:
        instances:
          i-001: []
```
All of the ways to misplace the "global" configuration:
```yaml
# global                                                                        
groups:          
  # unintentional "sharding" group                                                                                                                                    
  g-001:                                                                        
    # global for a group                                                        
    replicasets:
      # unintentional "sharding" replicaset                                                                
      r-001:                                                                    
        # global for a replicaset                                               
        instances:
          # unintentional "sharding" instance                                                              
          i-001: {                                                              
            # "global" for an instance                                          
          }                                                                     
```

We should check the following:
* every group contains a replicaset (so there is no empty "sharding" group)
* every replicaset contains an instance (so there is no empty "sharding" replicaset)

Note that we can't check instance this way, because it doesn't have the succeeding level.

PR #9896, opened 2024-04-01
config: verify replicaset to contain an instance

Lord-KA:
It is very easy to misplace a config option to a different lever, for
example create an empty replicaset `sharding` with storage role, instead
of configuring sharding option to `storage`:
```yaml
groups:
  g-001:
    replicasets:
      sharding:
        roles:
        - storage
      r-001:
        instances:
          i-001: {}
```

This patch adds verification that forbids creating a replicaset without
a single instance in it.

Closes https://github.com/tarantool/tarantool/issues/9895

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-9895-forbid-empty-group-or-replicaset.md
+++ b/changelogs/unreleased/gh-9895-forbid-empty-group-or-replicaset.md
@@ -1,0 +1,5 @@
+## bugfix/config
+
+* Added additional validation to a cluster's configuration.
+  Now it is forbidden to create an empty group or
+  replicaset (gh-9895).
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -731,11 +731,31 @@ local function validate_anon(found, peers, failover, leader)
     end
 end
 
+local function validate_misplacing(cconfig)
+    for group_name, group_cfg in pairs(cconfig.groups) do
+        if group_cfg.replicasets == nil or
+                next(group_cfg.replicasets) == nil then
+            error(('group %q should include at ' ..
+                   'least one replicaset.'):format(group_name), 0)
+        end
+
+        for replicaset_name, replicaset_cfg in pairs(group_cfg.replicasets) do
+            if replicaset_cfg.instances == nil or
+                    next(replicaset_cfg.instances) == nil then
+                error(('replicaset %q should include at ' ..
+                       'least one instance.'):format(replicaset_name), 0)
+            end
+        end
+    end
+end
+
 local function new(iconfig, cconfig, instance_name)
     -- Find myself in a cluster config, determine peers in the same
     -- replicaset.
     local found = cluster_config:find_instance(cconfig, instance_name)
     assert(found ~= nil)
+
+    validate_misplacing(cconfig)
 
     -- Precalculate configuration with applied defaults.
     local iconfig_def = instance_config:apply_default(iconfig)
--- a/test/config-luatest/cluster_config_test.lua
+++ b/test/config-luatest/cluster_config_test.lua
@@ -281,3 +281,33 @@ g.test_instance_uri_errors = function(g)
         end)
     end)
 end
+
+-- Attempt to pass an empty group and an empty replicaset.
+g.test_misplace_option = function(g)
+    local config = cbuilder.new()
+        :use_group('g-001')
+
+        :use_replicaset('r-001')
+        :add_instance('i-001', {})
+
+        :use_group('sharding')
+        :set_group_option('roles', {'storage'})
+        :config()
+
+    cluster.startup_error(g, config, "group \"sharding\" should " ..
+                                     "include at least one replicaset.")
+
+    local config = cbuilder.new()
+        :use_group('g-001')
+
+        :use_replicaset('r-001')
+        :add_instance('i-001', {})
+
+        :use_replicaset('sharding')
+        :set_replicaset_option('roles', {'storage'})
+
+        :config()
+
+    cluster.startup_error(g, config, "replicaset \"sharding\" should " ..
+                                     "include at least one instance.")
+end
--- a/test/config-luatest/cbuilder.lua
+++ b/test/config-luatest/cbuilder.lua
@@ -106,6 +106,17 @@ local function cbuilder_set_global_option(self, path, value)
     return self
 end
 
+-- Set an option for the selected group.
+local function cbuilder_set_group_option(self, path, value)
+    assert(type(path) == 'string')
+    path = fun.chain({
+        'groups', self._group,
+    }, path:split('.')):totable()
+
+    cluster_config:set(self._config, path, value)
+    return self
+end
+
 -- Set an option for the selected replicaset.
 local function cbuilder_set_replicaset_option(self, path, value)
     assert(type(path) == 'string')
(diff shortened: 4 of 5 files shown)


src/box/lua/config/cluster_config.lua
```diff
@@ -205,6 +205,15 @@ local nested_cluster_config = schema.new(schema_name, record_from_fields({
                         value = instance_config_with_scope('instance'),
                     }),
                 }),
+                validate = function(replicasets, w)
```
Totktonada, line 208: We discussed, whether it is possible that the same configuration mistake may appear in the upper scopes: empty `replicasets`, empty `groups`.

We also discussed that a partial config may miss `groups`. For example, a local file config may contain global options to connect to etcd or tarantool based configuration storage, while all the groups/replicasets/instances are in the remote config storage.

Another similar case is when a key in a remote storage contains just global options.

The main point is that a presence of `groups` is validated on the whole configuration, compiled from all the sources/keys, not before.

Whether we should apply the same logic for `replicasets` and `instances`?

For example, one may want to define options for a replicaset:

```yaml
groups:
  g-001:
    replicasets:
      r-001:
        replication:
          failover: election
        memtx:
          memory = <...>
        labels:
          size: large
```

But have `instances` defined in another key.

```yaml
groups:
  g-001:
    replicasets:
      r-001:
        instances:
          i-001: {}
          i-002: {}
          i-003: {}
```

Does it look meaningful?

changelogs/unreleased/gh-9895-forbid-empty-replicaset.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Added additional validation to cluster configuration.
+  Now it is forbidden to create an empty replicaset (gh-9895).
```
andreyaksenov, line 4:
```suggestion
* Added additional validation to a cluster's configuration.
  Now it is forbidden to create an empty replica set (gh-9895).
```

andreyaksenov (approved):

src/box/lua/config/cluster_config.lua
Lord-KA: We've discussed it more thoroughly and decided to forbid only empty replicasets and groups in merged config.

Lord-KA: Fixed.

changelogs/unreleased/gh-9895-forbid-empty-replicaset.md
Lord-KA: Applied the first line diff. I'd like to leave `replicaset` as a single word, similar to config schema.

src/box/lua/config/configdata.lua
```diff
@@ -731,12 +731,32 @@ local function validate_anon(found, peers, failover, leader)
...
 local function new(iconfig, cconfig, instance_name)
     -- Find myself in a cluster config, determine peers in the same
     -- replicaset.
     local found = cluster_config:find_instance(cconfig, instance_name)
     assert(found ~= nil)
 
+    validate_misplacing(cconfig)
+
     -- Precalculate configuration with applied defaults.
     local iconfig_def = instance_config:apply_default(iconfig)
```
Totktonada, line 762: `src/lib/small` submodule is updated. It seems, it is by a mistake.

src/box/lua/config/configdata.lua
```diff
@@ -731,12 +731,32 @@ local function validate_anon(found, peers, failover, leader)
     end
 end
 
+local function validate_misplacing(cconfig)
+    for group_name, group_cfg in pairs(cconfig.groups) do
+        if group_cfg.replicasets == nil or
+                next(group_cfg.replicasets) == nil then
+            error(('group %q should include at ' ..
+                   'least one replicaset.'):format(group_name))
```
Totktonada, line 739: Nit: Please, use `error(<...>, 0)` in config's code. It would be consistent with other errors from this subsystem.

All the other errors are reported this way (on startup):

```
LuajitError: You are doing X wrong, let's do Y
```

But if we miss the second `error()`'s argument it looks like so.

```
LuajitError: internal/config/somemodule.lua:237: You are doing X wrong, let's do Y
```

The report looks like a problem in tarantool's internal code, not like a problem in user's data. I don't want these reports look like something is wrong in tarantool itself.

The same idea is applicable to `config:info().alerts` if an error occurs on reload.

(I understand that file:line may be a useful piece of information for debugging, but I take this cost and trying to write nice error messages to make everything clean for a user.)

test/config-luatest/cluster_config_test.lua
```diff
@@ -281,3 +281,33 @@ g.test_instance_uri_errors = function(g)
...
+-- Attempt to pass an empty group and an empty replicaset.
+g.test_misplace_option = function(g)
+    local config = cbuilder.new()
+        :use_group('g-001')
+
+        :use_replicaset('r-001')
+        :add_instance('i-001', {})
+
+        :use_group('sharding')
+        :set_group_option('roles', {'super'})
```
Totktonada, line 294: Nit: The original report contains the `sharding: {roles: [storage]}` field in a wrong place. Not `super`, but `storage` -- a sharding role.

An attempt to use `super` as a sharding role would fail the validation even if it is placed on a correct level. So, the test case has two errors.

I would use `storage` in the test to make it more clear what is the actual error we aim to verify.

Feel free to ignore, just thought.

changelogs/unreleased/gh-9895-forbid-empty-group-or-replicaset.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/config
```
Totktonada, line 1: Typos from commit messages: `lever`, `replciaset`.

src/box/lua/config/cluster_config.lua
Totktonada: Thank you!

changelogs/unreleased/gh-9895-forbid-empty-replicaset.md
Totktonada: I use 'replicaset' as a single word too in comments in the code. I know that our documentation on the website uses 'replica set'. I remember that I changed my wording in changelogs to 'replica set' on a request several times.

However, I see both ways in existing release notes on GitHub: there is no consistency.

Can't suggest anything certain. I propose to don't spin here and go ahead.

@andreyaksenov I'm closing this discussion thread. I'm OK to discuss some common vocabulary our of scope of this PR.

Totktonada (approved):

src/box/lua/config/configdata.lua
Lord-KA: Fixed.

src/box/lua/config/configdata.lua
Lord-KA: That's a fair point, fixed.

test/config-luatest/cluster_config_test.lua
Lord-KA: > I would use storage in the test to make it more clear what is the actual error we aim to verify.

Well, we expect a specific error message, so it's not a big deal, but for consistency I will update all examples, including the test.

changelogs/unreleased/gh-9895-forbid-empty-group-or-replicaset.md
Lord-KA: Fixed.

PR #9902, opened 2024-04-02
config: introduce config:instance_uri()

ImeevMA:
This patch introduces the instance_uri() method for the config module. This method returns the URI of the specified instance from the config.

Needed for #9842

@TarantoolBot document
Title: instance_uri() method for config module

The configuration module's `instance_uri()` method returns the URI of the instance that is used to create a replicaset or sharding cluster. This method takes two arguments:
1) `peer` or null to return the URI that is used to create the
   replicaset, or `sharding` to return the URI that is used to create
   the sharding cluster;
2) `options`, which only support `options.instance`, to specify the
   instance whose URI should be returned; by default, the URI of the
   instance on which the function is executed is returned.

### DIFF
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -578,6 +578,28 @@ function methods.instances(self)
     return res
 end
 
+function methods.instance_uri(self, uri_type, opts)
+    selfcheck(self, 'instance_uri')
+    initcheck(self, 'instance_uri', 'cluster')
+    uri_type = uri_type == nil and 'peer' or uri_type
+    if uri_type ~= 'peer' and uri_type ~= 'sharding' then
+        error(('Expected "peer" or "sharding", got %q'):format(uri_type), 0)
+    end
+    opts = opts == nil and {} or opts
+    if type(opts) ~= 'table' then
+        error(('Expected table, got %s'):format(type(opts)), 0)
+    end
+    if opts.instance ~= nil and type(opts.instance) ~= 'string' then
+        error(('Expected string, got %s'):format(type(opts.instance)), 0)
+    end
+
+    local uri_opts = {
+        instance = opts.instance,
+        use_default = true,
+    }
+    return self._configdata_applied:_instance_uri(uri_type, uri_opts)
+end
+
 -- The object is a singleton. The constructor should be called
 -- only once.
 local function new()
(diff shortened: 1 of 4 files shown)


ImeevMA:
I submitted the first patch from https://github.com/tarantool/tarantool/pull/9892 to a separate PR.

andreyaksenov (approved):

src/box/lua/config/init.lua
```diff
@@ -578,6 +578,27 @@ function methods.instances(self)
     return res
 end
 
+function methods.instance_uri(self, uri_type, opts)
+    selfcheck(self, 'instance_uri')
```
Totktonada, line 582: An attempt to access this method before a startup is finished would lead to attempt to index `self._configdata_applied`, which is `nil`. I would call `initcheck()` to raise a proper error in the case.

src/box/lua/config/init.lua
```diff
@@ -578,6 +578,27 @@ function methods.instances(self)
...
+    uri_type = uri_type == nil and 'peer' or uri_type
+    if uri_type ~= 'peer' and uri_type ~= 'sharding' then
+        error(('Expected "peer" or "sharding", got %s'):format(uri_type), 0)
```
Totktonada, line 585: Nit: "peer" and "sharding" is in quotes, so I would use `%q` instead of `%s` for consistency.

src/box/lua/config/init.lua
```diff
@@ -578,6 +578,27 @@ function methods.instances(self)
...
+    end
+    opts = opts == nil and {} or opts
+    if type(opts) ~= 'table' then
+        error(('Expected table, got %s'):format(opts), 0)
+    end
+    if opts.instance ~= nil and type(opts.instance) ~= 'string' then
+        error(('Expected string, got %s'):format(opts.instance), 0)
+    end
```
Totktonada, line 593: I guess that `type(opts)` and `type(opts.instance)` should be in the `:format()` arguments.

Totktonada (approved):

src/box/lua/config/init.lua
ImeevMA: Thanks, fixed.

src/box/lua/config/init.lua
ImeevMA: Fixed.

src/box/lua/config/init.lua
ImeevMA: Fixed.

src/box/lua/config/init.lua
Totktonada: I don't think that it is crucial.

PR #9903, opened 2024-04-02
box: add error details for main box operations

### LINKED ISSUE #7223: Invalid key part count error is not descriptive enough

### LINKED ISSUE #9111: Add new ClientError codes for DML operations
[Link to the design document](https://www.notion.so/tarantool/Error-subsystem-improvements-90faa0a4714b4143abaf8bed2c10b2fc?pvs=4#0532e8fbb67a48e58fb0df4535807149)

Users often complain that errors thrown by `space.insert` or `space.select` functions are difficult to analyze because they lack information about the space for which the failed operation was executed. For example, #7223 was filed because an error thrown by some of Tarantool core DML operations looks like this:

```
Invalid key part count in an exact match (expected 1, got 0)
```

To make investigation easier, we’ll introduce a few new [`ClientError`](https://github.com/tarantool/tarantool/blob/1d6043fac53bebd539ee5882f6073767eea753bd/src/box/error.h#L200) codes, one for each DML operation: `ER_SELECT`, `ER_REPLACE`, `ER_INSERT`, `ER_UPDATE`, `ER_UPSERT`, `ER_DELETE`. They will be set by top-level DML operations ([`box_select`](https://github.com/tarantool/tarantool/blob/96706b7b7e860f23ee59913e347f9b5fd306b662/src/box/box.cc#L3244), [`box_process_rw`](https://github.com/tarantool/tarantool/blob/96706b7b7e860f23ee59913e347f9b5fd306b662/src/box/box.cc#L442)) with [`diag_add`](https://github.com/tarantool/tarantool/blob/1d6043fac53bebd539ee5882f6073767eea753bd/src/lib/core/diag.h#L457) so that the original error that caused the failure will be linked as a cause. All the new errors will store information about the space in the payload: `space_id`, `space_name`, `index_id`, `index_name`. They will also store `key`, `tuple`, and `ops`, if applicable.

nshy:
Add error details for basic box operations like insert, replace, select etc.

**DO NOT MERGE** until `luatest.assert_error_covers` make it into Tarantool (job in progress).  

Closes #9111
Closes #7223

@TarantoolBot document
Title: Provide more error details for basic box operations Product: Tarantool
Since: 3.1

For example for this setup:

```lua
  local test = box.schema.create_space('test')
  test:create_index('primary')
  test:create_index('second', {parts = {2, 'unsigned', 1, 'unsigned'}})
```

Before we get (note that compat option `box_error_serialize_verbose` is set to 'new' to show error details in console. It does not affect whether details are added or not):

```
tarantool> test.index.second:get{1}
---
- error:
    code: 19
    base_type: ClientError
    type: ClientError
    message: Invalid key part count in an exact match (expected 2, got 1)
    trace:
    - file: ./src/box/index.cc
      line: 149
...
```

And now we get (see 'operation', 'space', 'index' and 'key' fields):

```
tarantool> test.index.second:get{1}
---
- error:
    operation: get
    code: 19
    base_type: ClientError
    message: Invalid key part count in an exact match (expected 2, got 1)
    trace:
    - file: ./src/box/index.cc
      line: 149
    type: ClientError
    index: second
    space: test
    key: [1]
...
```

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3690,11 +3690,17 @@ box_index_id_by_name(uint32_t space_id, const char *name, uint32_t len)
 int
 box_process1(struct request *request, box_tuple_t **result)
 {
-	if (box_check_slice() != 0)
-		return -1;
 	struct space *space = space_cache_find(request->space_id);
+	const char *tuple = request->tuple;
+	const char *tuple_end = request->tuple_end;
+	const char *key = request->key;
+	const char *key_end = request->key_end;
+	uint32_t id;
+
 	if (space == NULL)
-		return -1;
+		goto error;
+	if (box_check_slice() != 0)
+		goto error;
 	/*
 	 * Allow to write to data-temporary and local spaces in the read-only
 	 * mode. To handle space truncation and/or ddl operations on temporary
@@ -3702,7 +3708,7 @@ box_process1(struct request *request, box_tuple_t **result)
 	 * _index system spaces till the on_replace trigger is called, when
 	 * we know which spaces are concerned.
 	 */
-	uint32_t id = space_id(space);
+	id = space_id(space);
 	if (is_ro_summary &&
 	    id != BOX_TRUNCATE_ID &&
 	    id != BOX_SPACE_ID &&
@@ -3710,7 +3716,7 @@ box_process1(struct request *request, box_tuple_t **result)
 	    !space_is_data_temporary(space) &&
 	    !space_is_local(space) &&
 	    box_check_writable() != 0)
-		return -1;
+		goto error;
 	if (space_is_memtx(space)) {
 		/*
 		 * Due to on_init_schema triggers set on system spaces,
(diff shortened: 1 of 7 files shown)


andreyaksenov (approved):

src/box/box.cc
```diff
@@ -3726,11 +3732,33 @@ box_process1(struct request *request, box_tuple_t **result)
...
+		error_set_str(e, "space", space->def->name);
+	if (request->index_id != 0) {
+		assert(request->key != NULL);
+		assert(request->key_end != NULL);
+		error_set_mp(e, "key", key, key_end - key);
+		if (space != NULL) {
+			struct index *index = space_index(space,
+							  request->index_id);
+			if (index != NULL)
+				error_set_str(e, "index", index->def->name);
```
locker, line 3758: We should set `space_id` and `index_id` as well.

src/box/box.cc
```diff
@@ -3726,11 +3732,33 @@ box_process1(struct request *request, box_tuple_t **result)
...
+		}
+	}
+	return -1;
```
locker, line 3761: You can and should set the operation right here - by looking at the request type.

src/box/index.cc
```diff
@@ -336,30 +351,45 @@ box_index_min(uint32_t space_id, uint32_t index_id, const char *key,
...
+	error_set_str(e, "operation", "min");
+	error_set_mp(e, "key", key_array, key_end - key_array);
+	/* space and index pointers can be invalid at this point. */
+	space = space_by_id(space_id);
+	if (space != NULL) {
+		error_set_str(e, "space", space->def->name);
+		index = space_index(space, index_id);
+		if (index != NULL)
+			error_set_str(e, "index", index->def->name);
+	}
```
locker, line 391: This snippet is repeated a few times. Maybe, factor it out to a helper function?

test/box-luatest/extra_assertions.lua
```diff
@@ -0,0 +1,11 @@
+local t = require('luatest')
+
+-- XXX: This is temporary. The assertion is present in latest Luatest and
+-- is making its way into Tarantool currently.
+t.assert_error_covers = function(expected, fn, ...)
```
locker, line 5: This helps looks trivial. Let's merge it into luatest now, before committing this patch.

test/box-luatest/extra_assertions.lua
```diff
@@ -0,0 +1,11 @@
...
+    local ok, actual = pcall(fn, ...)
+    t.assert_not(ok)
+    actual = actual:unpack()
```
locker, line 8: You should check that the error has the proper type before unpacking it.

test/box-luatest/extra_assertions.lua
```diff
@@ -0,0 +1,11 @@
...
+    t.assert_equals(type(actual), 'table')
```
locker, line 9: This assertion looks redundant. We know that if `actual` is a box error, `unpack` will return a table.

locker (changes requested):

PR #9907, opened 2024-04-03
space_upgrade: fix crashes on upgrading nullability

drewdzzz:
The patch fixes crashes on upgrading nullability of trailing fields. If we had nullable one, and we upgrade it to non-nullable, attribute `has_optional_part` of some indexes can be changed (even if no indexed fields are changed). So in order to correctly index both old and new tuples we should set new min_field_count value to the minimal min_field_count of old and new formats. Actual value will be set when space upgrade completes.

This is CE part of the patch. It must be merged after EE part is approved and all tests there are passed.
EE part: tarantool/tarantool-ee#748

Part of tarantool/tarantool-ee#698
Part of tarantool/tarantool-ee#750

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -1720,7 +1720,16 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
 {
 	struct space *old_space = alter->old_space;
 	bool is_min_field_count_changed;
-	if (old_space->format != NULL) {
+	if (old_space->def->opts.upgrade_def != NULL) {
+		/*
+		 * When space upgrade completes, we should unconditionally
+		 * update optionality because current min_field_count value
+		 * of a format respects min_field_count of format which space
+		 * had before upgrade. It was needed to correctly index both
+		 * old and new tuples.
+		 */
+		is_min_field_count_changed = true;
+	} else if (old_space->format != NULL) {
 		is_min_field_count_changed =
 			old_space->format->min_field_count !=
 			alter->new_min_field_count;
@@ -2355,6 +2364,19 @@ on_replace_dd_space(struct trigger * /* trigger */, void *event)
 						     old_space->index_count,
 						     def->fields,
 						     def->field_count);
+		/*
+		 * In the case of space upgrade both old and new tuples will
+		 * be indexed, so new_min_field_count should be the minimal
+		 * min_field_count of old and new formats.
+		 * Space upgrade does a replace in this space without
+		 * upgrade_def on completion, so actual min_field_count will
+		 * be set when space upgrade completes.
+		 */
+		if (def->opts.upgrade_def != NULL) {
+			alter->new_min_field_count = MIN(
+				alter->new_min_field_count,
+				old_space->format->min_field_count);
+		}
 		try {
 			(void) new CheckSpaceFormat(alter);
 			(void) new ModifySpace(alter, def);


src/box/alter.cc
```diff
@@ -1748,8 +1760,13 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
 		if (alter->pk_def == NULL || !index_depends_on_pk(old_index)) {
 			if (is_min_field_count_changed) {
 				new_def = index_def_dup(old_def);
-				index_def_update_optionality(new_def,
-							     min_field_count);
+				if (alter->is_space_upgrade) {
+					index_def_extend_optionality(
```
sergepetrenko, line 1764: Please drop a comment here describing when we "restore" the optionality.

sergepetrenko (approved):

src/box/alter.cc
drewdzzz: Done.

src/box/alter.cc
```diff
@@ -1108,8 +1108,20 @@ CheckSpaceFormat::prepare(struct alter_space *alter)
 	if (new_space->upgrade != NULL) {
 		/*
 		 * Tuples stored in the space will be checked against
-		 * the new format during space upgrade.
+		 * the new format during space upgrade. We should only
+		 * check if format of indexed fields wasn't updated.
```
locker, line 1124: I don't see any point in forbidding *compatible* space format updates (i.e. extending a space format). It should work both with space upgrade and without. This doesn't contradict the documentation.

src/box/alter.cc
```diff
@@ -1773,7 +1797,14 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
...
 					old_def->key_def, alter->pk_def);
-		index_def_update_optionality(new_def, min_field_count);
+		/*
+		 * Optionality is "extended" for the same reason as for indexes
+		 * not depending on primary key.
+		 */
+		if (alter->is_space_upgrade)
+			index_def_extend_optionality(new_def, min_field_count);
+		else
+			index_def_update_optionality(new_def, min_field_count);
```
locker, line 1807: I think there's a much easier way to achieve the same result:

```diff
diff --git a/src/box/alter.cc b/src/box/alter.cc
index 660cd1ef1c90..3753fc7bdcea 100644
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -2355,6 +2355,11 @@ on_replace_dd_space(struct trigger * /* trigger */, void *event)
 						     old_space->index_count,
 						     def->fields,
 						     def->field_count);
+		if (def->opts.upgrade_def != NULL) {
+			alter->new_min_field_count = MIN(
+				alter->new_min_field_count,
+				old_space->format->min_field_count);
+		}
 		try {
 			(void) new CheckSpaceFormat(alter);
 			(void) new ModifySpace(alter, def);
```

This will extend index optionality on space upgrade. Note that patching the EE repository isn't required because this code is also executed when space upgrade is completed.

locker (changes requested):

src/box/alter.cc
drewdzzz: Allowed compatible updates.

src/box/alter.cc
drewdzzz: Thanks, I stick with your solution now.

src/box/alter.cc
locker: I meant dropping the first patch altogether - incompatible changes of indexed fields won't work anyway because it isn't possible to create such tuple format.

src/box/alter.cc
```diff
@@ -1739,6 +1739,14 @@ alter_space_move_indexes(struct alter_space *alter, uint32_t begin,
...
 		is_min_field_count_changed = false;
 	}
+	/*
+	 * When space upgrade completes, we should unconditionally update
+	 * optionality because current min_field_count value of a format
+	 * respects min_field_count of format which space had before upgrade.
+	 * It was needed to correctly index both old and new tuples.
+	 */
+	if (alter->old_space->def->opts.upgrade_def != NULL)
+		is_min_field_count_changed = true;
```
locker, line 1749: Nit:
- You can use `old_space` instead of `alter->old_space` here.
- I'd put in one `if-else-if-else` block.

```C
if (old_space->def->opts.upgrade_def != NULL) {
	/* ... */
	is_min_field_count_changed = true;
} else if (old_space->format != NULL) {
	is_min_field_count_changed =
		old_space->format->min_field_count !=
		alter->new_min_field_count;
} else {
	is_min_field_count_changed = false;
}
```

src/box/alter.cc
drewdzzz: Dropped the first patch.

src/box/alter.cc
drewdzzz: Fixed.

locker (approved):

locker:
`app-luatest/http_client_test.lua` failulre in the `release_asan_clang` workflow is unrelated - the test is known to be flaky.

locker:

PR #9911, opened 2024-04-04
net.replicaset: fix a flaky test

alyapunov:
The test checks that net.replicaset connection switches to the new leader when the new leader is elected. Unfortunately the test expects that after successful call of box.ctl.promote() the old leader is guaranteed to be in RO state. Actually this is incorrect: the new leader can legally rely on the quorum with the third instance without even notifying the original leader about new election term started, so the old leader may remain in RW state. Since the replicaset connection updates states of instances asynchronously it also may think (for some time) that the old leader is the only RW instance in replicaset.

Fix it by including check of the new leader into retrying block of the test.

Hotfix #9823

NO_CHANGELOG=fix flaky test
NO_DOC=fix flaky test


Totktonada (approved):
The fix looks meaningful for me. Thanks!

unera (approved):
lgtm

PR #9922, opened 2024-04-09
relay: batch sent rows

### LINKED ISSUE #10161: relay: buffer sends to save on syscalls
Currently there is one write() per row in relay. This needs to be optimized to save on syscalls.

This is a reincarnation of an old ticket created by @kostja.

sergepetrenko:
relay: batch sent rows

Relay sends rows one by one, issuing one `write()` system call per each
sent row. This hurts performance quite a lot: in the 1mops_write test
[1] on my machine master node is able to reach ~ 1.9 million RPS while
replication performance peaks at ~ 0.5 million RPS.

As shown by a PoC patch, batching the rows allows to remove the
replication bottleneck almost completely. Even sending 2 rows per one
write() call already doubles the performance, and sending 64 rows per
one write() call allows the replica to catch up with master's speed [2].

This patch makes relay buffer rows until the total buffer size reaches a
certain threshold, and then flush everything with a single writev()
call.

The flush threshold is configurable via a new tweak -
xrow_buf_flush_size, which has a default value of 16 kb.

A size threshold is chosen over simple row count, because it's more
flexible for different row sizes: if the user has multi-kilobyte tuples,
batching hundreds of such tuples might overflow the socket's send buffer
size, resulting in extra wait until the send buffer becomes free again.
This extra time might've been used for enqueuing extra rows. The
problem would become even worse for hundred-kilobyte tuples and so on.

At the same time the user might have different workload types. One with
large tuples and other with relatively small ones, in which case
batching by size rather than by row count is more predictable
performance wise.

[1] https://github.com/tarantool/tarantool/blob/master/perf/lua/1mops_write.lua
[2] https://www.notion.so/tarantool/relay-248de4a9600f4c4d8d83e81cf1104926

Closes #10161

NO_DOC=not a documentable feature
NO_TEST=covered by existing tests

### DIFF
--- a/changelogs/unreleased/gh-10161-relay-speedup.md
+++ b/changelogs/unreleased/gh-10161-relay-speedup.md
@@ -1,0 +1,5 @@
+## bugfix/replication
+
+* Significantly improved replication performance by batching rows to be sent
+  before dispatch. Batch size in bytes may be controlled by a new tweak
+  `xrow_stream_flush_size` (default is 16 kb) (gh-10161).
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -112,6 +112,8 @@ struct relay_raft_msg {
 struct relay {
 	/** Replica connection */
 	struct iostream *io;
+	/** A stream to write rows to the remote peer. */
+	struct xrow_stream xrow_stream;
 	/** Request sync */
 	uint64_t sync;
 	/** Last ACK sent to the replica. */
@@ -373,6 +375,7 @@ relay_exit(struct relay *relay)
 	if (inj != NULL && inj->dparam > 0)
 		fiber_sleep(inj->dparam);
 
+	xrow_stream_destroy(&relay->xrow_stream);
 	/*
 	 * Destroy the recovery context. We MUST do it in
 	 * the relay thread, because it contains an xlog
@@ -442,6 +445,35 @@ relay_cord_init(struct relay *relay)
 	relay->lsr_id = 0;
 	relay->read_tsn = 0;
 	rlist_create(&relay->current_tx);
+	xrow_stream_create(&relay->xrow_stream);
+}
+
+/** Flush any relay stream contents to the remote peer immediately. */
+static inline int
+relay_flush(struct relay *relay)
+{
+	if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0)
+		return -1;
+#ifndef NDEBUG
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
+#endif
+	return 0;
+}
+
+/** Check if relay stream has enough data to flush and flush it. */
+static inline int
+relay_check_flush(struct relay *relay)
+{
+	if (xrow_stream_check_flush(&relay->xrow_stream, relay->io) < 0)
+		return -1;
+#ifndef NDEBUG
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
+#endif
+	return 0;
 }
 
 void
@@ -455,7 +487,9 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 
 	relay_start(relay, io, sync, relay_send_initial_join_row, relay_yield,
 		    UINT64_MAX);
+	xrow_stream_create(&relay->xrow_stream);
 	auto relay_guard = make_scoped_guard([=] {
+		xrow_stream_destroy(&relay->xrow_stream);
 		relay_stop(relay);
 		relay_delete(relay);
 	});
@@ -477,6 +511,8 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 
 	/* Send read view to the replica. */
 	engine_join_xc(&ctx, &relay->stream);
+	if (relay_flush(relay) < 0)
+		diag_raise();
 }
 
 int
@@ -491,6 +527,8 @@ relay_final_join_f(va_list ap)
 	assert(relay->stream.write != NULL);
 	recover_remaining_wals(relay->r, &relay->stream,
 			       &relay->stop_vclock, true);
+	if (relay_flush(relay) < 0)
+		diag_raise();
 	assert(vclock_compare(&relay->r->vclock, &relay->stop_vclock) == 0);
 	return 0;
 }
@@ -767,26 +805,25 @@ static inline void
 relay_send_heartbeat(struct relay *relay)
 {
 	struct xrow_header row;
-	try {
-		++relay->last_sent_ack.vclock_sync;
-		RegionGuard region_guard(&fiber()->gc);
-		xrow_encode_relay_heartbeat(&row, &relay->last_sent_ack);
-		/*
-		 * Do not encode timestamp if this heartbeat is sent in between
-		 * data rows so as to not affect replica's upstream lag.
-		 */
-		if (relay->last_row_time > relay->last_heartbeat_time)
-			row.tm = 0;
-		else
-			row.tm = ev_now(loop());
-		row.replica_id = instance_id;
-		relay->last_heartbeat_time = ev_monotonic_now(loop());
-		relay_send(relay, &row);
-		relay->need_new_vclock_sync = false;
-	} catch (Exception *e) {
-		relay_set_error(relay, e);
+	++relay->last_sent_ack.vclock_sync;
+	RegionGuard region_guard(&fiber()->gc);
+	xrow_encode_relay_heartbeat(&row, &relay->last_sent_ack);
+	/*
+	 * Do not encode timestamp if this heartbeat is sent in between
+	 * data rows so as to not affect replica's upstream lag.
+	 */
+	if (relay->last_row_time > relay->last_heartbeat_time)
+		row.tm = 0;
+	else
+		row.tm = ev_now(loop());
+	row.replica_id = instance_id;
+	relay->last_heartbeat_time = ev_monotonic_now(loop());
+	relay_send(relay, &row);
+	if (relay_flush(relay) < 0) {
+		relay_set_error(relay, diag_last_error(diag_get()));
 		fiber_cancel(fiber());
 	}
+	relay->need_new_vclock_sync = false;
 }
 
 /**
@@ -1022,6 +1059,10 @@ relay_subscribe_f(va_list ap)
 					 relay->last_row_time + timeout);
 		cbus_process(&relay->wal_endpoint);
 		relay_subscribe_update(relay);
+		if (relay_flush(relay) < 0) {
+			relay_set_error(relay, diag_last_error(diag_get()));
+			fiber_cancel(fiber());
+		}
 	}
 
 	/*
@@ -1099,13 +1140,10 @@ relay_send(struct relay *relay, struct xrow_header *packet)
 {
 	ERROR_INJECT_YIELD(ERRINJ_RELAY_SEND_DELAY);
 
+	struct xrow_stream *stream = &relay->xrow_stream;
 	packet->sync = relay->sync;
 	relay->last_row_time = ev_monotonic_now(loop());
-	coio_write_xrow(relay->io, packet);
-
-	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
-	if (inj != NULL && inj->dparam > 0)
-		fiber_sleep(inj->dparam);
+	xrow_stream_write(stream, packet);
 }
 
 static void
@@ -1118,6 +1156,8 @@ relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 	 */
 	if (row->group_id != GROUP_LOCAL)
 		relay_send(relay, row);
+	if (relay_check_flush(relay) < 0)
+		diag_raise();
 }
 
 /**
@@ -1128,16 +1168,16 @@ static void
 relay_raft_msg_push(struct cmsg *base)
 {
 	struct relay_raft_msg *msg = (struct relay_raft_msg *)base;
+	struct relay *relay = msg->relay;
 	struct xrow_header row;
 	RegionGuard region_guard(&fiber()->gc);
 	xrow_encode_raft(&row, &fiber()->gc, &msg->req);
-	try {
-		relay_send(msg->relay, &row);
-		msg->relay->sent_raft_term = msg->req.term;
-	} catch (Exception *e) {
-		relay_set_error(msg->relay, e);
+	relay_send(relay, &row);
+	if (relay_flush(relay) < 0) {
+		relay_set_error(relay, diag_last_error(diag_get()));
 		fiber_cancel(fiber());
 	}
+	relay->sent_raft_term = msg->req.term;
 }
 
 static void
@@ -1299,6 +1339,8 @@ relay_send_tx(struct relay *relay)
 		}
 		relay_send(relay, packet);
 	}
+	if (relay_check_flush(relay) < 0)
+		diag_raise();
 
 	rlist_create(&relay->current_tx);
 	lsregion_gc(&relay->lsregion, relay->lsr_id);
--- a/src/box/xrow_io.cc
+++ b/src/box/xrow_io.cc
@@ -57,8 +57,7 @@ coio_read_xrow(struct iostream *io, struct ibuf *in, struct xrow_header *row)
 	if (len > ibuf_used(in))
 		coio_breadn(io, in, len - ibuf_used(in));
 
-	xrow_header_decode_xc(row, (const char **) &in->rpos, in->rpos + len,
-			      true);
+	xrow_decode_xc(row, (const char **)&in->rpos, in->rpos + len, true);
 }
 
 void
@@ -88,8 +87,7 @@ coio_read_xrow_timeout_xc(struct iostream *io, struct ibuf *in,
 	if (len > ibuf_used(in))
 		coio_breadn_timeout(io, in, len - ibuf_used(in), delay);
 
-	xrow_header_decode_xc(row, (const char **) &in->rpos, in->rpos + len,
-			      true);
+	xrow_decode_xc(row, (const char **)&in->rpos, in->rpos + len, true);
 }
 
 
--- a/src/box/xrow_io.cc
+++ b/src/box/xrow_io.cc
@@ -28,12 +28,15 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  */
+#include <sys/uio.h>
 #include "xrow_io.h"
 #include "xrow.h"
 #include "coio.h"
 #include "coio_buf.h"
 #include "error.h"
 #include "msgpuck/msgpuck.h"
+#include "scoped_guard.h"
+#include "tweaks.h"
 
 void
 coio_read_xrow(struct iostream *io, struct ibuf *in, struct xrow_header *row)
@@ -102,3 +105,58 @@ coio_write_xrow(struct iostream *io, const struct xrow_header *row)
 		diag_raise();
 }
 
+uint64_t xrow_stream_flush_size = 16384;
+TWEAK_UINT(xrow_stream_flush_size);
+
+void
+xrow_stream_write(struct xrow_stream *stream, const struct xrow_header *row)
+{
+	/* Fixheader - encodes length of the packet. */
+	size_t fixheader_len = 5;
+	assert(fixheader_len == mp_sizeof_uint(UINT32_MAX));
+	size_t approx_len = fixheader_len + xrow_approx_len(row);
+	/* Reserve excess space to save on exact size calculation. */
+	char *data = (char *)xlsregion_reserve(&stream->lsregion, approx_len);
+	/* Leave space for the fixheader. */
+	char *d = data + fixheader_len;
+	d += xrow_header_encode(row, row->sync, d);
+	for (int i = 0; i < row->bodycnt; i++) {
+		size_t l = row->body[i].iov_len;
+		memcpy(d, row->body[i].iov_base, l);
+		d += l;
+	}
+	size_t data_len = d - data;
+	assert(data_len <= approx_len);
+	*data = 0xce; /* MP_UINT32 */
+	store_u32(data + 1, mp_bswap_u32(data_len - fixheader_len));
+	xlsregion_alloc(&stream->lsregion, data_len, ++stream->lsr_id);
+}
+
+int
+xrow_stream_flush(struct xrow_stream *stream, struct iostream *io)
+{
+#ifndef NDEBUG
+	assert(stream->owner == NULL);
+	stream->owner = fiber();
+	auto guard = make_scoped_guard([&] {
+		stream->owner = NULL;
+	});
+#endif
+	ssize_t to_flush = lsregion_used(&stream->lsregion);
+	/*
+	 * Might flush more than requested if data is added to the buffer
+	 * during the coio_writev yield.
+	 */
+	while (to_flush > 0) {
+		struct iovec iov[IOV_MAX];
+		int iovcnt = lengthof(iov);
+		int64_t gc_id = lsregion_to_iovec(&stream->lsregion, iov,
+						  &iovcnt, &stream->flush_pos);
+		ssize_t written = coio_writev(io, iov, iovcnt, 0);
+		if (written < 0)
+			return -1;
+		to_flush -= written;
+		lsregion_gc(&stream->lsregion, gc_id);
+	}
+	return 0;
+}
--- a/src/box/xrow_io.h
+++ b/src/box/xrow_io.h
@@ -30,6 +30,10 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  */
+
+#include "small/lsregion.h"
+#include "memory.h"
+
 #if defined(__cplusplus)
 extern "C" {
 #endif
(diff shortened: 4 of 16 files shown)


src/box/xrow_buf.h
```diff
@@ -0,0 +1,80 @@
...
+typedef int
+(*xrow_buf_flush_cb)(void *cb_data, struct iovec *iov, int iovcnt);
+
+/** A buffer for encoded rows. */
+struct xrow_buf {
+	/** A callback invoked each time xrow_buf is flushed. */
+	xrow_buf_flush_cb flush_cb;
+	/** Data for \a flush_cb. */
+	void *flush_cb_data;
+};
```
Gerold103, line 27: A callback which can implicitly on any push perform a flush looks like an overkill and makes relay code harder to understand what happens when exactly. I would suggest to keep `xrow_buf` a simple buffer. And make relay after each push check if the buffer needs to be flushed, and do the flush then.

Which in case of CE apparently will mean that the buffer can store only a single packet at once.

Gerold103 (commented):
Putting this into EE is literally the same as if we would place `sleep`s into random places of CE, and remove them in EE. This is extremely disappointing to see this happening with the product. Not even the commercialisation of it, but how small and paltry it is, that we even hide a simple optimization.

Given how trivial it is, picodata will just do it themselves if they didn't already, and will take more customers.

test/replication-luatest/gh_7515_sync_node_sees_leader_hang_test.lua
```diff
@@ -58,7 +58,7 @@ g.test_notice_leader_hang_during_sync = function(cg)
     -- restart. Generate transactions, to make sure heartbeats do not interfere
     -- with transactions, and make relay sleep after each sent row to make sure
     -- it takes server2 long enough to sync.
-    local old_term = cg.server1:exec(function()
+    local old_term, tweaks_default = cg.server1:exec(function()
```
Gerold103, line 61: Is the test changed for EE? That doesn't look good that EE CI depends on these hacks in CE tests. I suggest to find a way not to make CE tests so fragile. Is it possible for this "feature"?

src/box/relay.cc
```diff
@@ -422,11 +426,26 @@ relay_set_cord_name(int fd)
...
+static int
+relay_flush_cb(void *data, struct iovec *iov, int iovcnt)
+{
+	struct iostream *io = (struct iostream *)data;
+	if (coio_writev(io, iov, iovcnt, 0) < 0)
+		return -1;
+
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
```
locker, line 439: Why not leave it in `relay_buffer_row` (former `relay_send`)? Then we wouldn't need to update the tests AFAICS.

locker (approved):

src/box/relay.cc
```diff
@@ -1325,7 +1356,7 @@ relay_send_tx(struct relay *relay)
 			say_warn("injected broken lsn: %lld",
 				 (long long) packet->lsn);
 		}
-		relay_send(relay, packet);
+		relay_buffer_row(relay, packet);
```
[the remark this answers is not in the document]
sergepetrenko: Agree, the function remains `relay_send`.

src/box/relay.cc
sergepetrenko: It's not possible, unfortunately. We have to update the tests anyway.

Some tests (`replication/sync`, for example) rely on the fact that upstream lag will become tiny eventually, even if relay is slow to send rows. This wouldn't be possible if we started waiting **after** the `send`, but **before** the `flush`.

Speaking of altering the batch size in some tests, they rely on the fact that relay performs a sequence of sends with some given timeout in between. It's just easier to perform the needed amount of sends with a tiny batch size. So that I don't have to create ~16 kilobytes of data per each desired send.

sergepetrenko:
Good news, everyone!

We've reconsidered the decision to put this into EE. The patch will be put into CE. Thank you for your feedback and thank you for caring.

test/replication-luatest/gh_7515_sync_node_sees_leader_hang_test.lua
sergepetrenko: Now that's everything is implemented in CE, the test is changed for CE.

I have to tweak batch size, because the test requires sending multiple batches with timeouts in between (so that server sync takes some time), and I wanted to send as little data as possible.

src/box/relay_stream.h
```diff
@@ -0,0 +1,66 @@
...
+
+/** Written data size after which the stream should be flushed. */
+extern size_t relay_stream_flush_size;
+
+/**
+ * A structure encapsulating writes made by relay. Collects the rows into a
+ * buffer and flushes it to the network as soon as its size reaches a specific
+ * threshold.
+ */
+struct relay_stream {
```
locker, line 21: I don't think that "stream" is a proper name. It's more like a buffer.

src/box/relay_stream.c
```diff
@@ -0,0 +1,50 @@
...
+size_t relay_stream_flush_size = 16384;
+TWEAK_UINT(relay_stream_flush_size);
+
+void
+relay_stream_write(struct relay_stream *stream, const struct xrow_header *row)
+{
+	size_t used = region_used(&fiber()->gc);
+	int iovcnt;
+	struct iovec iov[XROW_IOVMAX];
+	xrow_to_iovec(row, iov, &iovcnt);
```
locker, line 15: Would it be possible to copy xrow directly to `lsregion`, bypassing `fiber->gc`?

src/box/relay_stream.h
Gerold103: It will become a stream as a part of #9968. Data will be written into it potentially with a yield, with no explicit flush required. From multiple places in relay. And one fiber will be handling the stream to actually forward the data into a socket (which will probably be inside this struct). It is called here "stream" so as we wouldn't need to rename it later.

src/box/relay_stream.c
```diff
@@ -0,0 +1,50 @@
...
+	 */
+	while (to_flush > stream->flush_pos.max_id) {
+		struct iovec iov[IOV_MAX];
+		int iovcnt = lengthof(iov);
+		lsregion_to_iovec(&stream->lsregion, iov, &iovcnt,
+				  &stream->flush_pos);
+		if (coio_writev(io, iov, iovcnt, 0) < 0)
+			return -1;
+		lsregion_gc(&stream->lsregion, stream->flush_pos.max_id);
+	}
```
Gerold103, line 45: I don't remember - we do have a guarantee, that only one fiber is here at a time, right? Otherwise 1) we get mixed garbage in the socket, 2) `lsregion_gc()` must use the position remembered before send. For (2) consider this: one  fiber writes to the buffer, takes iovec, moves `flush_pos`, and yields in `coio_writev()`, then another fiber does same. Then first fiber finishes and calls `lsregion_gc()` on the position updated by the second fiber. Hence deleting the data which is being sent right now in the second fiber.

src/box/relay_stream.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
```
Gerold103, line 1: Having it in a separate file made sense when it was EE plugin included into CE via `#include stream.cc` or something. But now perhaps it isn't so needed? And might even be reused someday for appliers? We coincidentally already have `box/xrow_io.h/c`. Do you think this struct would fit there?

I didn't think of this before, but as wrote above - I now think we might reuse it for appliers if eventually will make the appliers also do multiple tasks in different fibers. And that means the struct could be called `coio_funnel` or `coio_crater` - like "воронка": multiple fibers can write safely, and only one fiber is flushing into an `iostream`. The flush-size could be stored as a member and set once on the object creation.

What do you think of that?

src/box/relay_stream.h
```diff
@@ -0,0 +1,66 @@
...
+
+static inline void
+relay_stream_destroy(struct relay_stream *stream)
+{
+	lsregion_destroy(&stream->lsregion);
+}
+
+/** Test whether the stream is full and needs a flush. */
+static inline bool
+relay_stream_needs_flush(struct relay_stream *stream)
```
Gerold103, line 51: The stream here can be `const`.

Gerold103 (commented):
Thanks for the changes! I am very glad that this goes into CE, thanks for helping with that.

changelogs/unreleased/gh-10161-relay-speedup.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/replication
+
+* Significantly improved the replication performance by batching rows to be sent
+  before the dispatch. Batch size in bytes may be controlled by the new tweak
+  option `relay_stream_flush_size` (default 16 kb) (gh-10161).
```
lenkis, line 5:
```suggestion
* Significantly improved replication performance by batching rows to be sent
  before dispatch. Batch size in bytes may be controlled by a new configuration
  option `relay_stream_flush_size` (default is 16 kb) (gh-10161).
```

lenkis (approved):

src/box/relay_stream.c
sergepetrenko: Thanks for paying attention!

Yes, we do have such guarantee. There's a single fiber which reads WALs and sends them to the remote peer, sends occasional heartbeats and processes RAFT pushes. Besides, it always flushes a full transaction (see `relay_send_tx()`).
If this wasn't the case (either the tx was sent in parts or multiple fibers wrote to the socket), we would still need the check `relay->is_sending_tx`, which got deleted some time ago.

BTW, when implementing the relay rework in scope of #9968, you'll have to pay attention that a full transaction is added to the stream wthout yields for the same reasons as mentioned above.

src/box/relay_stream.h
sergepetrenko: I think it fits good in `xrow_io.h/.c` now, indeed. Moved the stream there.

Speaking of the name - I'd settle with a more general `xrow_stream`. `xrow_io_stream`, maybe.

TBH I don't see how this could be reused by appliers: they mostly read, not write. And the only things they write are occasional heartbeats.

src/box/relay_stream.h
sergepetrenko: > And one fiber will be handling the stream to actually forward the data into a socket (which will probably be inside this struct)

Yes, during some WIP iterations I tried to hide the `struct iostream` from relay into the `relay_stream`. So that it would have both `relay_stream_read` and `relay_stream_write` methods. And `relay_stream_read` would simply call `coio_read_xrow`. But I encountered some problems with needing to write to `iostream` from tx (some initial relay messages are sent from there), and then initializing the stream in relay. Although I think this may be bypassed. IIRC, `lsregion` doesn't have to be used in the same thread it was initialized in.

src/box/relay_stream.h
sergepetrenko: For now I settled with `struct xrow_stream`.

src/box/relay_stream.h
sergepetrenko: Thanks for noticing! Fixed.

src/box/relay_stream.c
sergepetrenko: Yes, it's possible. Thanks for the suggestion!

The only problem I encountered is that I have to know the exact header size before I encode it (otherwise there'll be some garbage padding between consequent lsregion allocations, which will spoil the whole idea of dumping a whole slab at once).

So I added the `xrow_header_sizeof()` method. Unfortunately it duplicates the `xrow_header_encode()` almost completely, but I don't think there is anything we can do about it. I also thought about using macros here, like:
```c
#define ACTION(type, val) data = mp_encode_##type(data, val);
/* OR*/
#define ACTION(type, val) total_size += mp_sizeof_##type(val);

/* and replace all mp_encode_type(data, val) with ACTION(type, val) inside xrow_header_encode() */
```
But it seems this would make `xrow_header_encode` even uglier, so for now I went with the duplication approach.

src/box/xrow_io.cc
```diff
@@ -102,3 +105,54 @@ coio_write_xrow(struct iostream *io, const struct xrow_header *row)
...
+	for (int i = 0; i < row->bodycnt; i++) {
+		len += row->body[i].iov_len;
+	}
+	char *data = (char *)xlsregion_alloc(&stream->lsregion,
+					     len + fixheader_len,
+					     ++stream->lsr_id);
+	*(data) = 0xce; /* MP_UINT32 */
+	store_u32(data + 1, mp_bswap_u32(len));
+	data = data + fixheader_len;
+	data += xrow_header_encode(row, row->sync, data);
```
Gerold103, line 127: Could you also make `xlsregion_reserve(XROW_HEADER_LEN_MAX)`, do `xrow_header_encode()`, and then `xlsregion_alloc(actually_encoded_size)`? Then you wouldn't need to do the relatively expensive size calculation before the encoding + wouldn't have gaps in the lsregion.

src/box/xrow_io.cc
```diff
@@ -102,3 +105,54 @@ coio_write_xrow(struct iostream *io, const struct xrow_header *row)
...
+	while (to_flush > 0) {
+		struct iovec iov[IOV_MAX];
+		int iovcnt = lengthof(iov);
+		int64_t gc_id = lsregion_to_iovec(&stream->lsregion, iov,
+						  &iovcnt, &stream->flush_pos);
+		ssize_t written = coio_writev(io, iov, iovcnt, 0);
+		if (written < 0)
+			return -1;
+		to_flush -= written;
+		lsregion_gc(&stream->lsregion, gc_id);
```
Gerold103, line 152: I know, not possible, at least yet, but imagine that 2 fibers enter this place. One remembers `to_flush`, second remembers `to_flush + offset` where `offset` is how much the second fiber added. Now they both will spin in `while(to_flush > 0)` trying to send first `to_flush` bytes twice, no? The second one will be stuck in the loop, because combined they want to send `to_flush * 2 + offset`, but after first fiber finishes sending, the lsregion only has left `offset` bytes.

src/box/xrow_io.cc
```diff
@@ -102,3 +105,54 @@ coio_write_xrow(struct iostream *io, const struct xrow_header *row)
...
+	}
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
+	return 0;
```
Gerold103, line 157: Can this errinj somehow stay in relay?

src/box/xrow_io.h
```diff
@@ -34,6 +34,9 @@
 extern "C" {
 #endif
 
+#include "small/lsregion.h"
+#include "memory.h"
+
 struct ibuf;
```
Gerold103, line 40: Might be not the best idea to have inclusions inside `extern "C"`. Can lead to all sorts of funny business with the linker.

Gerold103 (commented):
Thanks for the fixes! Clear stuff, good job 💪, I was surprised tbh that this batching wasn't done earlier.

src/lib/small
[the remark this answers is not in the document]
sergepetrenko: Done.

src/box/relay.cc
```diff
@@ -1152,6 +1162,10 @@ relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 	 */
 	if (row->group_id != GROUP_LOCAL)
 		relay_send(relay, row);
+	if (xrow_stream_needs_flush(&relay->xrow_stream) &&
+	    xrow_stream_flush(&relay->xrow_stream, relay->io) < 0) {
```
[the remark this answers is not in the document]
sergepetrenko: Done, thanks for the suggestion!

src/box/xrow_io.h
sergepetrenko: Fixed, thanks!

src/box/xrow_io.cc
sergepetrenko: Yes, indeed. It didn't come to my mind. Done, thanks!

src/box/xrow_io.cc
sergepetrenko: Yes, indeed, it's possible. But haven't we agreed that only a single fiber can perform the flush? Won't it be the case after the relay rework?

It seems this is not the only problem here if we want to allow multiple fibers inside `xrow_stream_flush`. Even if I track `to_flush` correctly, the following sequence of events is possible:
1. fiber1 enters `xrow_stream_flush()`, sees lsregion contents `"abc"`, and remembers `to_flush = 3`.
2. fiber1 performs `lsregion_to_iovec` + `coio_writev()`, which yields after writing `"ab"`. `lsregion_svp` also has `pos = 3`.
3. Someone adds `"d"` to the stream.
4. fiber2 enters `xrow_stream_flush()`, sees lsregion contents "abcd" and `lsregion_svp` `pos = 3`, so it remembers `to_flush = 1`.
5. fiber2 performs `lsregion_to_iovec + `coio_writev()`, which writes `"d"` successfully. So as a result the receiver sees `"abdc"` instead of `"abcd"`.

Is this possible?

I could add guards against this, like that:

```diff
diff --git a/src/box/xrow_io.cc b/src/box/xrow_io.cc
index 0dae40b60a..7f895786a2 100644
--- a/src/box/xrow_io.cc
+++ b/src/box/xrow_io.cc
@@ -36,6 +36,7 @@
 #include "errinj.h"
 #include "error.h"
 #include "msgpuck/msgpuck.h"
+#include "scoped_guard.h"
 #include "tweaks.h"
 
 void
@@ -135,6 +136,13 @@ xrow_stream_write(struct xrow_stream *stream, const struct xrow_header *row)
 int
 xrow_stream_flush(struct xrow_stream *stream, struct iostream *io)
 {
+#ifndef NDEBUG
+       assert(stream->owner == NULL);
+       stream->owner = fiber();
+       auto guard = make_scoped_guard([&] {
+               stream->owner = NULL;
+       });
+#endif
        ssize_t to_flush = lsregion_used(&stream->lsregion);
        /*
         * Might flush more than requested if data is added to the buffer
diff --git a/src/box/xrow_io.h b/src/box/xrow_io.h
index 48eb8b5bad..dfd31cbf06 100644
--- a/src/box/xrow_io.h
+++ b/src/box/xrow_io.h
@@ -68,6 +68,10 @@ struct xrow_stream {
        int64_t lsr_id;
        /** A savepoint used between flushes. */
        struct lsregion_svp flush_pos;
+#ifndef NDEBUG
+       /** A fiber which's currently using the stream. */
+       struct fiber *owner;
+#endif
 };
 
 /** Initialize the stream. */
@@ -82,6 +86,7 @@ xrow_stream_create(struct xrow_stream *stream)
 static inline void
 xrow_stream_destroy(struct xrow_stream *stream)
 {
+       assert(stream->owner == NULL);
        lsregion_destroy(&stream->lsregion);
 }
 
```

</details>

But I don't think it makes sense. We don't have such guards in coio or anywhere else. Haven't applied the change yet.

src/box/xrow_io.cc
sergepetrenko: Ok:

```diff
diff --git a/src/box/relay.cc b/src/box/relay.cc
index fa9d8e500a..405a9c437a 100644
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -442,6 +442,34 @@ relay_cord_init(struct relay *relay)
 	xrow_stream_create(&relay->xrow_stream);
 }
 
+/** Flush any relay stream contents to the remote peer immediately. */
+static inline int
+relay_flush(struct relay *relay)
+{
+	if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0)
+		return -1;
+#ifndef NDEBUG
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
+#endif
+	return 0;
+}
+
+/** Check if relay stream has enough data to flush and flush it. */
+static inline int
+relay_check_flush(struct relay *relay)
+{
+	if (xrow_stream_check_flush(&relay->xrow_stream, relay->io) < 0)
+		return -1;
+#ifndef NDEBUG
+	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
+	if (inj != NULL && inj->dparam > 0)
+		fiber_sleep(inj->dparam);
+#endif
+	return 0;
+}
+
 void
 relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 		   uint32_t replica_version_id)
@@ -518,7 +546,7 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 
 	/* Send read view to the replica. */
 	engine_join_xc(&ctx, &relay->stream);
-	if (xrow_stream_flush(&relay->xrow_stream, io) < 0)
+	if (relay_flush(relay) < 0)
 		diag_raise();
 }
 
@@ -534,7 +562,7 @@ relay_final_join_f(va_list ap)
 	assert(relay->stream.write != NULL);
 	recover_remaining_wals(relay->r, &relay->stream,
 			       &relay->stop_vclock, true);
-	if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0)
+	if (relay_flush(relay) < 0)
 		diag_raise();
 	assert(vclock_compare(&relay->r->vclock, &relay->stop_vclock) == 0);
 	return 0;
@@ -826,7 +854,7 @@ relay_send_heartbeat(struct relay *relay)
 	row.replica_id = instance_id;
 	relay->last_heartbeat_time = ev_monotonic_now(loop());
 	relay_send(relay, &row);
-	if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0) {
+	if (relay_flush(relay) < 0) {
 		relay_set_error(relay, diag_last_error(diag_get()));
 		fiber_cancel(fiber());
 	}
@@ -1066,7 +1094,7 @@ relay_subscribe_f(va_list ap)
 					 relay->last_row_time + timeout);
 		cbus_process(&relay->wal_endpoint);
 		relay_subscribe_update(relay);
-		if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0) {
+		if (relay_flush(relay) < 0) {
 			relay_set_error(relay, diag_last_error(diag_get()));
 			fiber_cancel(fiber());
 		}
@@ -1162,7 +1190,7 @@ relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 	 */
 	if (row->group_id != GROUP_LOCAL)
 		relay_send(relay, row);
-	if (xrow_stream_check_flush(&relay->xrow_stream, relay->io) < 0)
+	if (relay_check_flush(relay) < 0)
 		diag_raise();
 }
 
@@ -1179,7 +1207,7 @@ relay_raft_msg_push(struct cmsg *base)
 	RegionGuard region_guard(&fiber()->gc);
 	xrow_encode_raft(&row, &fiber()->gc, &msg->req);
 	relay_send(relay, &row);
-	if (xrow_stream_flush(&relay->xrow_stream, relay->io) < 0) {
+	if (relay_flush(relay) < 0) {
 		relay_set_error(relay, diag_last_error(diag_get()));
 		fiber_cancel(fiber());
 	}
@@ -1338,7 +1366,7 @@ relay_send_tx(struct relay *relay)
 		}
 		relay_send(relay, packet);
 	}
-	if (xrow_stream_check_flush(&relay->xrow_stream, relay->io) < 0)
+	if (relay_check_flush(relay) < 0)
 		diag_raise();
 
 	rlist_create(&relay->current_tx);
diff --git a/src/box/xrow_io.cc b/src/box/xrow_io.cc
index 0dae40b60a..696032e43b 100644
--- a/src/box/xrow_io.cc
+++ b/src/box/xrow_io.cc
@@ -33,7 +33,6 @@
 #include "xrow.h"
 #include "coio.h"
 #include "coio_buf.h"
-#include "errinj.h"
 #include "error.h"
 #include "msgpuck/msgpuck.h"
 #include "tweaks.h"
@@ -151,8 +150,5 @@ xrow_stream_flush(struct xrow_stream *stream, struct iostream *io)
 		to_flush -= written;
 		lsregion_gc(&stream->lsregion, gc_id);
 	}
-	struct errinj *inj = errinj(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE);
-	if (inj != NULL && inj->dparam > 0)
-		fiber_sleep(inj->dparam);
 	return 0;
 }

```

<details>

sergepetrenko:
Some quick perf test results.
Current master:
```shell
Tarantool 3.2.0-entrypoint-114-g7db4de75f8
Target: Linux-x86_64-RelWithDebInfo

## 150 operations per txn
./src/tarantool  perf/lua/1mops_write.lua --nodes 2
# 150 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1006808
# master done 8990700 ops in time: 5.759516, cpu: 13.514103
# master average speed	1561016	ops/sec
# master peak speed	1666003	ops/sec
# replicas done 8990700 ops in time: 34.837695, cpu: 42.816187
1mops_master 1561016 rps
1mops_replica 258073 rps

## 10 operations per txn
./src/tarantool  perf/lua/1mops_write.lua --nodes 2 --transaction 10
# 10 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1001138
# master done 8998870 ops in time: 7.005794, cpu: 18.728187
# master average speed	1284489	ops/sec
# master peak speed	1522003	ops/sec
# replicas done 8998870 ops in time: 35.623021, cpu: 47.583947
1mops_master 1284489 rps
1mops_replica 252613 rps

## 1 operation per txn
./src/tarantool  perf/lua/1mops_write.lua --nodes 2 --transaction 1
# 1 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1000172
# master done 8999836 ops in time: 22.332386, cpu: 57.682858
# master average speed	402994	ops/sec
# master peak speed	443098	ops/sec
# replicas done 8999836 ops in time: 37.330169, cpu: 72.829966
1mops_master 402994 rps
1mops_replica 241087 rps
```
Patched version:
```shell
Tarantool 3.2.0-entrypoint-118-g421e8433c8
Target: Linux-x86_64-RelWithDebInfo

## 150 operations per txn
./src/tarantool  perf/lua/1mops_write.lua --nodes 2
# 150 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1006808
# master done 8990700 ops in time: 5.639669, cpu: 10.398049
# master average speed	1594189	ops/sec
# master peak speed	1711011	ops/sec
# replicas done 8990700 ops in time: 5.639697, cpu: 10.398076
1mops_master 1594189 rps
1mops_replica 1594181 rps

## 10 operations per txn
tarantool  perf/lua/1mops_write.lua --nodes 2 --transaction 10
# 10 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1000358
# master done 8999650 ops in time: 6.291824, cpu: 15.241492
# master average speed	1430372	ops/sec
# master peak speed	1565742	ops/sec
# replicas done 8999650 ops in time: 6.291856, cpu: 15.241523
1mops_master 1430372 rps
1mops_replica 1430364 rps

## 1 operation per txn
./src/tarantool  perf/lua/1mops_write.lua --nodes 2 --transaction 1
# 1 operations per txn,
# using 50 fibers,
# in a replicaset of 2 nodes,
# using HASH index type
# with WAL mode write
# 
# starting 1 replicas
# replication	1
# Warmup... done, lsn: 	1000300
# master done 8999708 ops in time: 21.076192, cpu: 40.496671
# master average speed	427008	ops/sec
# master peak speed	447861	ops/sec
# replicas done 8999708 ops in time: 21.076220, cpu: 40.496697
1mops_master 427008 rps
1mops_replica 427007 rps
```

locker (approved):

Gerold103 (approved):
Thanks, cool stuff! Such a huge impact with such a small patch, huh?

src/box/xrow_io.cc
Gerold103: I would add the guard. It wouldn't hurt and would prevent questions like I had above, because the guarantee would be clearly in the code and would be covered by tests.

sergepetrenko:
> Such a huge impact with such a small patch, huh?

Yeah, the difference is crazy.

src/box/xrow_io.cc
sergepetrenko: Ok, I've applied the diff above.
Resolving, since the PR is approved.

ligurio (approved):
Patch **xrow: split header encoding from body and allocations**.

Thanks for the patch! I have no objections.

PR #9931, opened 2024-04-10
Introduce `mode` option for `call()` and `filter()` from `experimental.connpool` module

### LINKED ISSUE #9930: Add filter by RO/RW in experimental.filter() and experimental.call() functions
These filters are in great demand.

ImeevMA:
This patch adds a `mode` option to `experimental.call()` and `experimental.filter()`. This option allows to execute filter candidates for these functions depending on RO status of the instances.

Closes #9930

### DIFF
--- a/changelogs/unreleased/gh-9930-mode-option.md
+++ b/changelogs/unreleased/gh-9930-mode-option.md
@@ -1,0 +1,4 @@
+## feature/experimental
+
+* Introduced the `mode` option for the `filter()` and `call()` functions
+  in the `experimental.connpool` module (gh-9930).
--- a/src/box/lua/experimental/connpool.lua
+++ b/src/box/lua/experimental/connpool.lua
@@ -207,29 +207,60 @@ local function get_connection(opts)
         return nil, "no candidates are available with these conditions"
     end
 
-    if opts.prefer_local ~= false then
-        local candidate_idx = nil
-        for n, candidate in ipairs(candidates) do
-            if candidate == box.info.name then
-                candidate_idx = n
-                local conn = connect(box.info.name, {wait_connected = false})
-                if conn:wait_connected() then
-                    return conn
-                end
-                break
+    -- Initialize the weight of each candidate.
+    local weights = {}
+    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
+        candidates = connect_to_candidates(candidates)
+    end
+    for _, instance_name in pairs(candidates) do
+        weights[instance_name] = 0
+    end
+
+    -- Increase weights of candidates preferred by mode.
+    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
+        local mode = opts.mode == 'prefer_ro' and 'ro' or 'rw'
+        local weight_mode = 2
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-9930-mode-option.md
```diff
@@ -0,0 +1,4 @@
+## feature/experimental
+
+* Introduced the `mode` option for `filter()` and `call()` functions in
+  the `experimental.connpool` module (gh-9930).
```
andreyaksenov, line 4:
```suggestion
* Introduced the `mode` option for the `filter()` and `call()` functions in
  the `experimental.connpool` module (gh-9930).
```

andreyaksenov (approved):

changelogs/unreleased/gh-9930-mode-option.md
ImeevMA: Thanks, fixed!

Totktonada:
I would like to highlight constraints of the current implementation.

* If there are unavailable candidates, `connpool.filter` with the `mode = 'ro'/'rw'` filter tries to (re)connect to them. It adds RTT * number of unavailable candidates delay. I would like to re-use existing information that an instance is not available (considering that it is actual during some time or just always refreshing it in background). We also can skip unavailable candidates if there are suitable available ones (and start reconnects in background). I want `connpool.filter` to be on the order of 'processing tables in Lua' time complexity rather then on the order of network delay.

  It seems that this `connpool.filter` complexity affects `connpool.call` when `mode ~= nil`.
* `connpool.call` also waits for connections' (re)establishing for unavailable instances if `mode == nil` even if there are available candidates.
* I would like to have some kind of `mode = 'alive'` filter in `connpool.filter` and maybe it should be set by default. The reason is that I'm not interested in unavailable candidates if I'm going to perform some remote call. Such a (default) behaviour of `connpool.filter` would be consistent with how `connpool.call` behaves.
* ~I think that `prefer_local` should affect candidates order if `mode = 'prefer_ro'` or `mode = 'prefer_rw'` is set.~ It seems fixed by a recent update with filters' weight.

changelogs/unreleased/gh-9930-mode-option.md
```diff
@@ -0,0 +1,4 @@
+## feature/experimental
```
Totktonada, line 1: From the documentation request:

> Also, if `prefer_ro` or `prefer_rw` mode is requested, the
> `prefer_local` option is ignored. With all other modes, `prefer_local`
> works as expected.

AFAIU, it was solved.

Totktonada (approved):
I've objections regarding timings in presence of unavailable instances. I also have a proposal to filter out unavailable instances in `connpool.filter()` by default.

These points are summarized above with some details.

Taking it into account, I still want this PR be part of the 3.1.0 release. It allows to start using the `mode` filtering for application development. The timing constraints look as a blocker for production usage, but we can fix it in a bugfix release.

I'm OK with the proposed implementation as with a temporary one within an experimental module.

changelogs/unreleased/gh-9930-mode-option.md
ImeevMA: Thanks, fixed.

PR #9933, opened 2024-04-10
lua: bump metrics module

DifferentialOrange:
(This is a backport of PR https://github.com/tarantool/tarantool/pull/9901 to release/3.0, future 3.0.2 release.)

----

Bump metrics package submodule. There are 8 new commits, but only two of them affects Tarantool:
- Add election_leader_idle metric [1];
- test: run log capture tests on a separate server [2]; the other 6 affect documentation and CI/CD scripts of the original repo. The latter one is required to bump luatest in test-run [3], since otherwise metrics log capture test fails on a new version.

1. https://github.com/tarantool/metrics/commit/ba9726d9b0cdfb22aa56d852f7fdc7b0aa22756a
2. https://github.com/tarantool/metrics/commit/3370f856efd4172bf24916e66e3846eeb01550a8
3. https://github.com/tarantool/test-run/pull/426

NO_DOC=doc is a part of submodule
NO_TEST=repo runs submodule tests, no new test files are added

(cherry picked from commit 50f74250e2423223352ed2dd1907789629b4bfad)


changelogs/unreleased/bump-metrics-to-3370f85.md
```diff
@@ -0,0 +1,3 @@
+## feature/lua
+
+* Added a `tnt_election_leader_idle` metric to built-in `metrics`.
```
Totktonada, line 3: Technically speaking, we can't add features into a bugfix release.

However, I guess that your motivation to propose this update to `release/3.0` is about testing changes: something may broke if we don't push it here. Am I right? I would like if you reveal more details about a motivation *of the backport*.

Taking it into account, we can prepare a special `tarantool-3.0` branch in the metrics repository, push needed testing fixes to it and update the submodule in `release/3.0` on the head of this new branch.

OTOH, such a branching requires extra effort and attention, it costs. So, I would propose an alternative.

First, why we don't add features to bugfix releases? The usual answer is stability, but it is not the whole story. There are quite isolated features that unlikely degrade any existing functionality, but we still don't want to have them in a bugfix release.

The reason is that we want to make the branching model clear for a user. A bugfix release solves problems in existing functionality. A feature release brings new functionality. If we mix these kinds of changes in the release notes, the branching model looks like a mess.

The solution is to be fair in motivation of an update and broadcast it via changelog/release notes.

Let's compare.

```markdown
## Release 3.0.2

### Functionality added

* `metrics` submodule is updated to X to bring the new metric Y.
```

```markdown
## Release 3.0.2

### Testing

* `metrics` submodule is updated to X for compatibility with recent version of
  Y, which is used in our testing/CI.
```

The top level section `Testing` [was added](https://github.com/tarantool/tarantool/pull/9451) for this kind of changes.

You can [check out](https://github.com/tarantool/tarantool-ee/pull/725) my example of such a change.

DifferentialOrange: See updated changelog

nshy (approved):

Totktonada (approved):

PR #9936, opened 2024-04-11
txn: run statement `on_rollback` triggers before rolling back statement

### LINKED ISSUE #9893: Rollback of DDL on `_space` space is completely broken
**Steps to reproduce**

```lua

box.cfg{}

box.error.injection.set("ERRINJ_WAL_IO", true)
box.space._space:alter{is_sync = false}
box.error.injection.set("ERRINJ_WAL_IO", false)

require('console').start()
```

**Actual behavior**

```console
tarantool> box.space._space:select{}
---
- error: 'console: exception while formatting the output: "builtin/msgpackffi.lua:622:
    Unsupported extension type"'
...

tarantool> box.space._space:alter{is_sync = false}
---
- error: 'builtin/box/schema.lua:941: assertion failed!'
...

```

Under ASAN build, this reproducer crashes. While, I was modifying the source code to insert additional log messages I also experienced crashes.

**Expected behavior**

Everything works correctly.

CuriousGeorgiy:
This patch fixes the order of running a transaction statement's `on_rollback` triggers and rolling back the statement itself.

Closes #9893

### DIFF
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -341,11 +341,15 @@ txn_stmt_prepare_rollback_info(struct txn_stmt *stmt, struct tuple *old_tuple,
 }
 
 /*
- * Undo changes done by a statement and run the corresponding
- * rollback triggers.
+ * Run the statement's rollback triggers and undo changes done by the statement.
+ *
+ * Logically, we call triggers after running statements. These triggers can
+ * make significant changes (for instance, DDL triggers), so, for consistency,
+ * we should call the statement's `on_rollback` triggers before rolling back
+ * the corresponding statement.
  *
  * Note, a trigger set by a particular statement must be run right
- * after the statement is rolled back, because rollback triggers
+ * before the statement is rolled back, because rollback triggers
  * installed by DDL statements restore the schema cache, which is
  * necessary to roll back previous statements. For example, to roll
  * back a DML statement applied to a space whose index is dropped
@@ -355,12 +359,12 @@ txn_stmt_prepare_rollback_info(struct txn_stmt *stmt, struct tuple *old_tuple,
 static void
 txn_rollback_one_stmt(struct txn *txn, struct txn_stmt *stmt)
 {
-	if (txn->engine != NULL && stmt->space != NULL)
-		engine_rollback_statement(txn->engine, txn, stmt);
 	if (stmt->has_triggers && trigger_run(&stmt->on_rollback, stmt) != 0) {
 		diag_log();
 		panic("statement rollback trigger failed");
 	}
+	if (txn->engine != NULL && stmt->space != NULL)
+		engine_rollback_statement(txn->engine, txn, stmt);
 }
 
 /*
--- a/src/box/user.cc
+++ b/src/box/user.cc
@@ -41,6 +41,7 @@
 #include "sequence.h"
 #include "trivia/util.h"
 #include "tt_static.h"
+#include "txn.h"
 
 struct universe universe;
 static struct user users[BOX_USER_MAX];
@@ -401,10 +402,31 @@ user_set_effective_access(struct user *user)
 }
 
 /**
- * Reload user privileges and re-grant them.
+ * Reload a single user privilege from a privilege tuple.
  */
 static int
-user_reload_privs(struct user *user)
+user_reload_priv(struct user *user, struct tuple *tuple)
+{
+	struct priv_def priv;
+	if (priv_def_create_from_tuple(&priv, tuple) != 0)
+		return -1;
+	/**
+	 * Skip role grants, we're only
+	 * interested in real objects.
+	 */
+	if (priv.object_type != SC_ROLE || !(priv.access & PRIV_X))
+		user_grant_priv(user, &priv);
+	return 0;
+}
+
+/**
+ * Reload user privileges and re-grant them. If a rolled back statement is
+ * passed, the privileges are reloaded as part of the rollback process, so we
+ * must account for the changes made by the statement and use the old data
+ * (which will be the actual data after rollback).
+ */
+static int
+user_reload_privs(struct user *user, struct txn_stmt *rolled_back_stmt)
 {
 	if (user->is_dirty == false)
 		return 0;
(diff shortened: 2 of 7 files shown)


src/box/txn.c
```diff
@@ -340,27 +340,48 @@ txn_stmt_prepare_rollback_info(struct txn_stmt *stmt, struct tuple *old_tuple,
...
- * after the statement is rolled back, because rollback triggers
+ * before the statement is rolled back, because rollback triggers
  * installed by DDL statements restore the schema cache, which is
  * necessary to roll back previous statements. For example, to roll
  * back a DML statement applied to a space whose index is dropped
  * later in the same transaction, we must restore the dropped index
  * first.
+ *
+ * Note, the _priv space rollback trigger accesses the database from
+ * `user_reload_privs` (gh-xxx), so it needs to be run after rolling back the
```
CuriousGeorgiy, line 373: If you agree with this patch and that the `_priv` space rollback trigger works wrong, I will open a ticket.

drewdzzz (commented):
Do we need an opportunity to alter `_space` when Tarantool is running?
It seems that this case is really tricky, so I'd prefer to forbid `_space:alter`.

drewdzzz (commented):
I don't like that we need to add `priv_reload_info` to `txn_stmt_rollback_info`.
As I understand, we assume that space `_priv` is small, so we could save all tuples in `on_replace`, insert the new tuple, if any, and delete the old one, if any - then we could pass `priv_reload_info` as `on_rollback->data` to the trigger.

CuriousGeorgiy:
> As I understand, we assume that space `_priv` is small, so we could save all tuples in `on_replace`, insert the new tuple, if any, and delete the old one, if any - then we could pass `priv_reload_info` as `on_rollback->data` to the trigger.

I am already doing the same thing. The problem is we need to also let the trigger know what tuple to use for rolling back information. An alternative would be passing `priv_reload_info` to the trigger, but then we would need to dynamically allocate it.

drewdzzz:
You can allocate `priv_reload_info` with `region_alloc(&txn->region)` - it's cheap so it's not a problem.

CuriousGeorgiy:
@drewdzzz The solution you suggest requires manually patching the port contents to restore the old space state, i.e., removing the new tuple, if any, and restoring the old tuple, if any.

Ports are append-only structures, and doing this requires hacking around them and adding a lot of new one-time-use code to support this compared to the current solution.

I want a second opinion to opt for this.

drewdzzz:
> requires manually patching the port contents to restore the old space state, i.e., removing the new tuple, if any, and restoring the old tuple, if any.

You can compare a tuple to a changed one right before appending it to the port - you won't need hacking around then.

CuriousGeorgiy:
@drewdzzz Yeah, sorry, I missed this opportunity. I pushed your alternative solution in #10004, please take a look.

sergepetrenko (commented):
Thanks for the patch!

Please consider the following solution: instead of collecting all the space contents into port to iterate over them in `user_reload_privs` you may pass `old_tuple`, `new_tuple` pairs corresponding to the rolled back statement, and keep iteration over the space contents as is. Simply replace `new_tuple` with `old_tuple` when you encounter one during iteration.

This way the code changes will be minimal.

CuriousGeorgiy:
@drewdzzz @sergepetrenko I implemented the solution that @sergepetrenko suggested, please take a look.

drewdzzz (approved):
Thanks for the patch, LGTM, please address my comment.

src/box/user.cc
```diff
@@ -444,23 +463,38 @@ user_reload_privs(struct user *user)
...
+		bool in_rollback = stmt != NULL;
+		struct tuple *old_tuple = in_rollback ? stmt->old_tuple : NULL;
+		struct tuple *new_tuple = in_rollback ? stmt->new_tuple : NULL;
+		struct key_def *key_def = index->def->key_def;
+		int cmp_old_tuple = old_tuple != NULL ?
+			tuple_compare_with_key(old_tuple, HINT_NONE, key, 1,
+					       HINT_NONE, key_def) : 0;
+		int cmp_new_tuple = new_tuple != NULL ?
+			tuple_compare_with_key(new_tuple, HINT_NONE, key, 1,
+					       HINT_NONE, key_def) : 0;
```
drewdzzz, line 475: Nit: maybe we should do only one comparison in Release? I would put the second comparison to the scope of `#ifndef NDEBUG` or even get rid of it - it's quite reliable invariant that old and new tuple have the same primary key.

src/box/user.h
```diff
@@ -237,7 +237,7 @@ role_revoke(struct user *grantee, struct user *role);
  * role if this role.
  */
 int
-priv_grant(struct user *grantee, struct priv_def *priv);
+priv_grant(struct user *grantee, struct priv_def *priv, struct txn_stmt *stmt);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/user.cc
CuriousGeorgiy: Fixed: added an `ifdef` condition for the second comparison. I would leave the assertions for clarity of what's going on, since the code looks hacky.

drewdzzz: Well, now it looks weird - I think it's a bad idea to use `ifndef` right in the middle of an expression. Let's compare with key one of the tuples, which is not `NULL`, and  then compare two tuples with `tuple_compare` if they both are not NULL in the scope of `ifndef` with a comment like "just making sure that the tuples have the same primary key".

CuriousGeorgiy: Fixed as you suggested: removed the `ifndef` and moved the comparison under the assertion.

test/box-luatest/rollback_ddl_on__priv_space_test.lua
```diff
@@ -0,0 +1,51 @@
...
+        box.session.su('modify')
+        local modify_user = box.session.uid()
+
+        box.session.su('admin')
+
+        local grant_privs = box.space._priv:select{grant_user}
+        box.begin()
+        box.schema.user.grant('grant', 'execute', 'universe')
+        box.rollback()
+        t.assert_equals(box.space._priv:select{grant_user}, grant_privs)
```
sergepetrenko, line 35: The test passes even without your patch (I reverted the _priv commit while being on top of your branch).

It seems you should check that actual operations are not allowed (output from bad tarantool version: note that `box.schema.user.info()` shows correct output, while the actual access is allowed) :
```lua
tarantool> box.schema.user.create('user')
---
...
tarantool> box.session.su('user')
---
...
tarantool> box.space._space:fselect{}
---
- error: Read access to space '_space' is denied for user 'user'
...
tarantool> box.session.su('admin')
---
...
tarantool> box.begin() box.schema.user.grant('user', 'read', 'space', '_space') box.rollback()
---
...
-- Note that the rolled back grant isn't present in `user.info()`, but actually still applies.
tarantool> box.schema.user.info('user')
---
- - - execute
    - role
    - public
  - - session,usage
    - universe
    - 
  - - alter
    - user
    - user
...
tarantool> box.session.su('user')
---
...
tarantool> box.space._space:fselect{}
---
- |-
  +-----+-----+------------------------+-----------+-----------+------------------+-----------------------------------------------+
  | id  |owner|          name          |  engine   |field_count|      flags       |                    format                     |
  +-----+-----+------------------------+-----------+-----------+------------------+-----------------------------------------------+
  | 257 |  1  |"_vinyl_deferred_delete"|"blackhole"|     0     |  {"group_id":1}  |[{"name":"space_id","type":"unsigned"},{"name":|
    
    <truncated for simplicity>

  | 380 |  1  |  "_session_settings"   | "service" |     2     |{"temporary":true}|[{"name":"name","type":"string"},{"name":"value|
  +-----+-----+------------------------+-----------+-----------+------------------+-----------------------------------------------+
...
```

sergepetrenko (changes requested):
Thanks for the fixes!
I like this version much more. Please, handle the issues with the test and we'll be good to go.

test/box-luatest/rollback_ddl_on__priv_space_test.lua
CuriousGeorgiy: Fixed: tested the `grant`, `revoke`, `modify` of `read` and `write` privileges with operations on a `test` space.

sergepetrenko: Thanks!

sergepetrenko (approved):
Thanks for the fixes!
LGTM.

PR #9944, opened 2024-04-11
test: cover split-brain during promote

### LINKED ISSUE #9943: Cover with tests, that promote of the previous term is detected and causes Split Brain
Picodata makes a commit, which nopifies PROMOTEs and DEMOTEs in some cases: [link](https://git.picodata.io/picodata/tarantool/-/merge_requests/160/diffs?commit_id=350ebf48cd2b26105f0cf199452c37d95a0f72d7). However, none of a tarantool tests fails, which means that this is not covered with tests. We should fix that.

Serpentian:
This test checks, that when PROMOTE from the prevous term is encountered we immediately notice split-brain situation and break replication without corrupting data.

Closes #9943

NO_DOC=test
NO_CHANGELOG=test

### DIFF
--- a/test/replication-luatest/gh_5295_split_brain_test.lua
+++ b/test/replication-luatest/gh_5295_split_brain_test.lua
@@ -380,3 +380,90 @@ g_very_old_term.test_confirm = function( cg)
     cg.servers[1]:update_box_cfg({replication = cg.box_cfg.replication})
     cg.cluster:wait_for_fullmesh()
 end
+
+local function test_promote_split_brain(cg, with_data)
+    if with_data then
+        cg.servers[1]:exec(function()
+            box.schema.create_space('test', {is_sync = true}):create_index('pk')
+        end)
+        cg.servers[2]:wait_for_vclock_of(cg.servers[1])
+        cg.servers[3]:wait_for_vclock_of(cg.servers[1])
+    end
+    for i = 2, 3 do
+        cg.servers[i]:exec(function()
+            box.cfg{election_mode = 'manual'}
+        end)
+    end
+
+    -- Third server is network partitioned.
+    local new_replication = {
+        cg.servers[1].net_box_uri,
+        cg.servers[2].net_box_uri,
+    }
+
+    cg.servers[3]:exec(update_replication, {})
+    cg.servers[1]:exec(update_replication, new_replication)
+    cg.servers[2]:exec(update_replication, new_replication)
+
+    -- Bump term on the first server, so that it is more, than term
+    -- of the third server after its promotion.
+    cg.servers[2]:exec(write_promote)
+    cg.servers[2]:wait_for_election_leader()
+    local term = cg.servers[2]:get_synchro_queue_term()
+    cg.servers[1]:wait_for_synchro_queue_term(term)
+    cg.servers[1]:exec(write_promote)
+    if with_data then
+        cg.servers[1]:exec(function()
+            box.space.test:replace({1, 's1'})
+        end)
+    end
+    cg.servers[3]:exec(function()
+        -- Otherwise it won't be able to start the promotion.
+        box.cfg{replication_synchro_quorum = 1}
+    end)
+    cg.servers[3]:exec(write_promote)
+    t.assert(cg.servers[3]:get_synchro_queue_term() <
+             cg.servers[1]:get_synchro_queue_term())
+    if with_data then
+        cg.servers[3]:exec(function()
+            box.space.test:replace({1, 's3'})
+        end)
+    end
+
+    -- Split-Brain should be noticed, when the first server manages to
+    -- connect to the third one.
+    cg.servers[1]:exec(update_replication, {
+        cg.servers[1].net_box_uri,
+        cg.servers[2].net_box_uri,
+        cg.servers[3].net_box_uri,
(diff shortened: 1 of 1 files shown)


test/replication-luatest/gh_9943_promote_split_brain_test.lua
```diff
@@ -0,0 +1,124 @@
...
+
+    promote(g.rs.servers[2])
+    g.rs.servers[1]:exec(function(id)
+        -- Without this wait PROMOTE may be not executed and exit with TimedOut
+        -- error. Let's make sure, that server_1 known about new leader.
+        t.helpers.retrying({}, function()
+            t.assert_equals(box.info.election.leader, id)
+        end)
+    end, {g.rs.servers[2]:get_instance_id()})
+    promote(g.rs.servers[1])
```
Gerold103, line 76: Why did you need to promote 2 and then 1? Could you simply promote 1 right away? If this is to bump the term, then lets add a comment about that.

test/replication-luatest/gh_9943_promote_split_brain_test.lua
```diff
@@ -0,0 +1,124 @@
...
+    -- Split-Brain should be noticed.
+    set_replication(g.rs.servers[1], {
+        g.rs.servers[1].net_box_uri,
+        g.rs.servers[2].net_box_uri,
+        g.rs.servers[3].net_box_uri,
+    })
+
+    t.helpers.retrying({}, function()
+        t.assert_not_equals(g.rs.servers[1]:grep_log('ER_SPLIT_BRAIN'), nil)
+    end)
```
Gerold103, line 101: Can you instead get it from the `info.replication` from upstream/downstream status? Log greps is the last resort.

Gerold103 (commented):
Thanks for the patch!

test/replication-luatest/gh_9943_promote_split_brain_test.lua
Serpentian: > If this is to bump the term, then lets add a comment about that

it's, added comments

test/replication-luatest/gh_9943_promote_split_brain_test.lua
Serpentian: Sure, it's better to check `box.info.replication.upstream`

sergepetrenko (commented):
Thanks for the test!
Please consider adding this case to the `test/replication-luatest/gh_5295_split_brain_test.lua` test. It contains all the similar split-brain test cases. I was sure that the case with an old promote is also handled there, but apparently it's not.

Gerold103 (approved):
Thanks for the test!

Serpentian:
> Please consider adding this case to the test/replication-luatest/gh_5295_split_brain_test.lua test.

Moved the test there, no changes to the test were done

test/replication-luatest/gh_5295_split_brain_test.lua
```diff
@@ -380,3 +380,99 @@ g_very_old_term.test_confirm = function( cg)
...
+    cg.servers[3]:exec(update_replication, {})
+    cg.servers[1]:exec(update_replication, new_replication)
+    cg.servers[2]:exec(update_replication, new_replication)
+
+    -- Bump term on the first server, so that it is more, than term
+    -- of the third server after its promotion.
+    promote(cg.servers[2])
+    cg.servers[1]:exec(function(id)
+        -- Without this wait PROMOTE may be not executed and exit with TimedOut
+        -- error. Let's make sure, that server_1 known about new leader.
```
sergepetrenko, line 418: typo: known->knows

test/replication-luatest/gh_5295_split_brain_test.lua
```diff
@@ -380,3 +380,99 @@ g_very_old_term.test_confirm = function( cg)
...
+        t.helpers.retrying({}, function()
+            t.assert_equals(box.info.election.leader, id)
+        end)
+    end, {cg.servers[2]:get_instance_id()})
+    promote(cg.servers[1])
```
sergepetrenko, line 423: Can you reuse the `write_promote` function defined above?

test/replication-luatest/gh_9943_promote_split_brain_test.lua
sergepetrenko: Resolving since @Gerold103 approved the PR.

sergepetrenko (commented):
Thanks for the fixes!

test/replication-luatest/gh_5295_split_brain_test.lua
Serpentian: Dropped the comment. Since now `write_promote` is used, I sync severs by their limbo term, which is enough to stabilize the test

test/replication-luatest/gh_5295_split_brain_test.lua
Serpentian: Sure, done

sergepetrenko (approved):
Thanks for the test!
LGTM.

sergepetrenko:
Backports:
3.1 https://github.com/tarantool/tarantool/pull/10255
2.11 https://github.com/tarantool/tarantool/pull/10254

PR #9945, opened 2024-04-12
connpool: instances, replicasets and groups filter

ImeevMA:
Follow-up #9842

NO_CHANGELOG=the experimental module is not released yet

@TarantoolBot document
Title: The `instances`, `replicasets` and `groups` options

Three new options were introduced in `connpool.filter()` and `connpool.call()` functions: `instances`, `replicasets` and `groups`. 1) The `instances` option is a list of instance names to which the
   filtered instances should belong.
2) The `replicasets` option is a list of replicaset names to which the
   filtered instances should belong.
2) The `groups` option is a list of group names to which the filtered
   instances should belong.

### DIFF
--- a/src/box/lua/experimental/connpool.lua
+++ b/src/box/lua/experimental/connpool.lua
@@ -93,6 +93,42 @@ local function connect_to_candidates(candidates)
     return connected_candidates
 end
 
+local function is_group_match(expected_groups, present_group)
+    if expected_groups == nil or next(expected_groups) == nil then
+        return true
+    end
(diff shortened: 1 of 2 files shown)


ImeevMA:
It has been decided that this PR will not be pushed to 3.1.0.

Totktonada (approved):

Totktonada:
@sergos said that he postpones the pre-release freeze. So, I'm going to include this patch right into 3.1.0.

PR #9955, opened 2024-04-16
memtx: fix a bug with mvcc and exclude_null option

### LINKED ISSUE #9954: Crash with mvcc and exclude_null
Simple repro:
```lua
box.cfg({
  listen='localhost:3333',
  memtx_use_mvcc_engine=true,
})

space = box.schema.space.create('test', { if_not_exists = true })
space:format({
  { name = 'id';    type = 'unsigned';  },
  { name = 'data';  type = 'string';  },
  { name = 'cnt';      type = 'unsigned';  is_nullable=true },
})

space:create_index( 'pk', {
  parts = {'id'},
  if_not_exists = true,
})

space:create_index( 'second', {
  parts = {{'data'}, {'cnt', exclude_null=true}},
  unique = false,
  if_not_exists = true,
})

require('console').start()
```
Then call `box.space.test:insert({1, 'fdf'})`

<details>
<summary>If falls with the following backtrace</summary>

```
#1  0x55cc76b35385 in crash_collect+256
#2  0x55cc76b35d97 in crash_signal_cb+100
#3  0x7ffa72fd9420 in __restore_rt+0
#4  0x55cc768d1585 in tuple_has_flag+56
#5  0x55cc768d7e8d in memtx_tx_handle_gap_write+404
#6  0x55cc768d8960 in memtx_tx_history_add_insert_stmt+1106
#7  0x55cc768d8e92 in memtx_tx_history_add_stmt+287
#8  0x55cc768edb1b in memtx_space_replace_all_keys+361
#9  0x55cc768edd85 in memtx_space_replace_tuple+177
#10 0x55cc768edf44 in memtx_space_execute_replace+180
#11 0x55cc76975ad7 in space_execute_dml+613
#12 0x55cc769c1234 in box_process_rw+306
#13 0x55cc769cdace in box_process1+453
#14 0x55cc769ce5f8 in box_insert+181
#15 0x55cc76acff7f in lbox_insert+307
#16 0x55cc76b9d827 in lj_BC_FUNCC+70
#17 0x55cc76baadc6 in lua_pcall+945
#18 0x55cc76b13df9 in luaT_call+45
#19 0x55cc76b06f02 in lua_main+265
#20 0x55cc76b07acc in run_script_f+2993
#21 0x55cc7688303a in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*)+34
#22 0x55cc76b41919 in fiber_loop+210
#23 0x55cc76ee8f6d in coro_init+120
Aborted (core dumped)
```
</details>

alyapunov:
Before this patch MVCC engine expected that if index_replace sets `result` to NULL then index_replace sets `successor` to something (NULL or existing tuple, depending on index type). That looked fine because by contract `successor` is set when true insertion was happened.

Unfortunately it was not considered that case of `exclude_null` part in index the insertion can be silently skipped and thus `successor` can be not set. The latter access of it led to UB.

Fix it by explicit check of tuple_key_is_excluded and work on this case correctly.

Note that logically `index_replace` should return a flag whether the new tuple was filtered (excluded) by key_def. But on the other hand this flag is required only for mvcc while the function is already has lots of agruments and it's very cheap to determine this flag right from memtx_tx, so I decided to make the most simple patch.

Closes #9954

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-9954-mvcc-crash-with-exclude-null.md
+++ b/changelogs/unreleased/gh-9954-mvcc-crash-with-exclude-null.md
@@ -1,0 +1,3 @@
+## bugfix/core
+
+* Fixed a bug that resulted in a crash when both MVCC and index with the `exclude_null` part were used (gh-9954).
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -1227,7 +1227,10 @@ memtx_tx_story_unlink_top_common(struct memtx_story *story, uint32_t idx)
 		unreachable();
 		panic("failed to rebind story in index");
 	}
-	assert(story->tuple == removed);
+	assert(story->tuple == removed ||
+	       (removed == NULL && tuple_key_is_excluded(story->tuple,
+							 index->def->key_def,
+							 MULTIKEY_NONE)));
 	story->link[idx].in_index = NULL;
 	if (old_story != NULL)
 		old_story->link[idx].in_index = index;
@@ -1396,7 +1399,13 @@ memtx_tx_story_full_unlink_on_space_delete(struct memtx_story *story)
 					unreachable();
 					panic("failed to rollback change");
 				}
-				assert(story->tuple == removed);
+				struct key_def *key_def = index->def->key_def;
+				assert(story->tuple == removed ||
+				       (removed == NULL &&
+					tuple_key_is_excluded(story->tuple,
+							      key_def,
+							      MULTIKEY_NONE)));
+				(void)key_def;
 				link->in_index = NULL;
 				/*
 				 * All tuples in pk are referenced.
@@ -1471,7 +1480,13 @@ memtx_tx_story_full_unlink_story_gc_step(struct memtx_story *story)
 					unreachable();
 					panic("failed to rollback change");
 				}
-				assert(story->tuple == removed);
+				struct key_def *key_def = index->def->key_def;
+				assert(story->tuple == removed ||
+				       (removed == NULL &&
+					tuple_key_is_excluded(story->tuple,
+							      key_def,
+							      MULTIKEY_NONE)));
+				(void)key_def;
 				link->in_index = NULL;
 				/*
 				 * All tuples in pk are referenced.
@@ -1997,12 +2012,18 @@ memtx_tx_history_add_insert_stmt(struct txn_stmt *stmt,
 	/* Process replacement in indexes. */
 	struct tuple *directly_replaced[space->index_count];
 	struct tuple *direct_successor[space->index_count];
+	bool tuple_excluded[space->index_count];
 	uint32_t directly_replaced_count = 0;
 	for (uint32_t i = 0; i < space->index_count; i++) {
 		struct index *index = space->index[i];
 		struct tuple **replaced = &directly_replaced[i];
 		struct tuple **successor = &direct_successor[i];
-		if (index_replace(index, NULL, new_tuple,
+		*replaced = *successor = NULL;
+		tuple_excluded[i] = tuple_key_is_excluded(new_tuple,
+							  index->def->key_def,
+							  MULTIKEY_NONE);
+		if (!tuple_excluded[i] &&
+		    index_replace(index, NULL, new_tuple,
 				  DUP_REPLACE_OR_INSERT,
 				  replaced, successor) != 0)
 		{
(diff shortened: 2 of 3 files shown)


src/box/memtx_tx.c
```diff
@@ -1997,12 +1997,22 @@ memtx_tx_history_add_insert_stmt(struct txn_stmt *stmt,
...
+	memset(direct_successor, '#', sizeof(direct_successor));
+#endif
+	bool tuple_excluded[space->index_count];
 	uint32_t directly_replaced_count = 0;
 	for (uint32_t i = 0; i < space->index_count; i++) {
 		struct index *index = space->index[i];
 		struct tuple **replaced = &directly_replaced[i];
 		struct tuple **successor = &direct_successor[i];
-		if (index_replace(index, NULL, new_tuple,
+		*replaced = *successor = NULL;
```
CuriousGeorgiy, line 2010: Nit: didn't you already initialize these values under `NDEBUG` earlier?

src/box/memtx_tx.c
```diff
@@ -2037,13 +2047,15 @@ memtx_tx_history_add_insert_stmt(struct txn_stmt *stmt,
...
 		struct tuple *succ = direct_successor[i];
-		if (next == NULL) {
-			/* Collect conflicts. */
+		/* Collect conflicts. */
+		if (next == NULL && !tuple_excluded[i]) {
 			memtx_tx_handle_gap_write(space, add_story, succ, i);
 			memtx_tx_handle_point_hole_write(space, add_story, i);
+		}
+		/* Form chains. */
+		if (next == NULL) {
```
CuriousGeorgiy, line 2056: Why should we still link the story to this index?

alyapunov: I mean it would be sufficient to set newer_story and older_story to NULL, but in this case `memtx_tx_story_link_top` does what needed.

changelogs/unreleased/gh-9954-mvcc-crash-with-exclude-null.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/core
+
+* Fixed a bug that would lead to crash when both MVCC and index with `exclude_null` part were used (gh-9954).
```
xuniq, line 3:
```suggestion
* Fixed a bug that resulted in a crash when both MVCC and index with the `exclude_null` part were used (gh-9954).
```

xuniq (approved):

CuriousGeorgiy (approved):

src/box/memtx_tx.c
alyapunov: Discussed online, decided that since that link is already has NULLs (see `story_new`) there's no need to call link_top at all.

drewdzzz (changes requested):
My test:
```lua
g.test_mvcc_exclude_null_replace_rollback = function()
    g.server:exec(function()
        local txn_proxy = require("test.box.lua.txn_proxy")
        local tx1 = txn_proxy.new()
        local tx2 = txn_proxy.new()
        box.space.test:insert{1, 'fdf'}
        tx1:begin()
        tx1("box.space.test:replace{3, 'b'}")
        t.assert_equals(tx1:rollback(), '')
    end)
end
```

Result:
```
Assertion failed: (story->tuple == removed), function memtx_tx_story_full_unlink_story_gc_step, file memtx_tx.c, line 1474.
```

You still partially rely on the fact that `replace` always actually replaces something. Please revise all places calling `index_replace` and cover index with `exclude_null` with more test cases.

drewdzzz:
Another breaking test case:
```lua
g.test_mvcc_exclude_null_drop_space_during_txn = function()
    g.server:exec(function()
        local txn_proxy = require("test.box.lua.txn_proxy")
        local tx1 = txn_proxy.new()
        local tx2 = txn_proxy.new()
        box.space.test:insert{1, 'fdf'}
        tx1:begin()
        tx1("box.space.test:replace{3, 'b'}")
        box.space.test:drop()
        t.assert_equals(tx1:rollback(), '')
    end)
end
```

Result:
```
Assertion failed: (story->tuple == removed), function memtx_tx_story_unlink_top_common, file memtx_tx.c, line 1230
```

drewdzzz (approved):
Thanks for your fixes, LGTM!

alyapunov:
Backported to 3.0 too.

PR #9972, opened 2024-04-23
third_party: initial import of `arrow/abi.h`

Gumix:
Needed for tarantool/tarantool-ee#718


third_party/arrow/abi.h
locker, line 1: If we need this header only for EE, why not add it just there?

Gumix: There are no `third_party/` in EE.

locker: Let's add it there. Shouldn't be difficult. We just need to update `/CMakeLists.txt` in the tarantool-ee repo to add the new ABI header into include directory list.

Gumix: Done.

PR #9973, opened 2024-04-24
Add stubs for file FETCH_SNAPSHOT

Serpentian:
This PR adds a new way to fetch snapshot for anonymous replicas. Instead of read-view join, file snapshot fetching may be used. File snapshot fetching is available only in EE.

Needed for tarantool/tarantool-ee#741


src/box/relay_file_join.h
```diff
@@ -0,0 +1,46 @@
...
+ * for vinyl - .run files.
+ *
+ * @param io client connection
+ * @param stream relay's stream, to which data is sent
+ * @param sync sync from incoming JOIN request
+ * @param vclock[out] vclock of the snapshot
+ * @param replica_version_id peer's version
+ * @param cursor_engine code of the cursor engine
+ * @param cursor_vclock vclock of the snapshot to fetch from
+ * @param cursor_lsn start lsn in file, engine specific
```
locker, line 25: AFAIU the problem you're trying to solve here is avoiding rebootstrap of a CDC client if it fails while fetching a snaphot. One way to achieve that is sending a snapshot because we can seek a position in it. There's another way though: we could leave the join iterator lingering for a while after a failure (say for 1 minute) to let the recipient reconnect and resume from where it stopped. Did you consider it? This could be done with the regular join, without introducing "file join".

src/box/iproto_constants.h
```diff
@@ -218,7 +218,23 @@ extern const char *iproto_flag_bit_strs[];
...
 	 * Flag indicating whether the transaction is synchronous.
 	 */								\
-	 _(IS_SYNC, 0x61, MP_BOOL)
+	 _(IS_SYNC, 0x61, MP_BOOL)					\
+	/**
+	 * Cursor, used in IRPOTO_FETCH_SNAPSHOT, in order to fetch snapshot
+	 * from a file:
+	 *
+	 * IPROTO_CURSOR: {
+	 *      IRPOTO_ENGINE: number,
```
locker, line 227: Why do you need the engine? I don't think that it's that bad to re-scan memtx snapshot if the replica failed to read vinyl, but even this could be avoided by remembering the number of rows in the memtx snapshot, without requiring the caller to specify the engine.

src/box/iproto_constants.h
```diff
@@ -218,7 +218,23 @@ extern const char *iproto_flag_bit_strs[];
...
+	 *      IPROTO_VCLOCK: map,
+	 *      IPROTO_LSN: number,
```
locker, line 229: - I don't think you need a vclock to resume - signature would be enough.
- The name "lsn" is confusing - it's a row number in the snapshot and should be named appropriately.
- Why do you need to put these fields in a map? Why not store them at the top level?
- `IPROTO_CURSOR` is a too generic name - for example, it could be used for pagination.

src/box/relay.cc
```diff
@@ -472,6 +475,16 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 		relay_delete(relay);
 	});
 
+	if (unlikely(cursor_engine != NONE_ENGINE)) {
+		relay_initial_file_join(relay->io, &relay->stream, sync,
+					vclock, replica_version_id,
+					cursor_engine, cursor_vclock,
+					cursor_lsn);
```
locker, line 482: I don't like that there's the new "file join" code path. Basically, it's normal join with the resume position. Can you pass these arguments to the engine's join callback and let the engine handler resume join instead of restarting it?

src/box/box.cc
```diff
@@ -4313,12 +4313,11 @@ box_process_fetch_snapshot(struct iostream *io,
...
 
-	say_info("sending current read-view to replica at %s", sio_socketname(io->fd));
-
 	/* Send the snapshot data to the instance. */
 	struct vclock start_vclock;
-	relay_initial_join(io, header->sync, &start_vclock, req.version_id);
-	say_info("read-view sent.");
+	relay_initial_join(io, header->sync, &start_vclock, req.version_id,
+			   req.cursor_engine, &req.cursor_vclock,
+			   req.cursor_lsn);
```
locker, line 4320: The way you split the PR doesn't make it easy to review - quite the contrary: to figure out what's `cursor_vlock` and `cursor_lsn` I have to select a different commit.

src/box/vinyl.h
```diff
@@ -33,13 +33,72 @@
...
+#include <small/mempool.h>
+
+#include "engine.h"
+#include "vy_quota.h"
+#include "vy_lsm.h"
+#include "vy_cache.h"
+#include "vy_run.h"
+#include "vy_mem.h"
+#include "vy_scheduler.h"
+#include "vy_regulator.h"
```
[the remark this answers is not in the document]
Serpentian: Thank you! Done everything like that: vinyl obviously requires exposing `vy_env`, which we don't want, memtx is done like that for consistency with vinyl, relay - for even stronger code path divergence.

src/box/relay_file_join.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+void
+relay_initial_file_join(struct iostream *io, struct xstream *stream,
+			uint64_t sync, struct vclock *vclock,
+			uint32_t replica_version_id, uint64_t cursor_engine,
+			const struct vclock *cursor_vclock, int64_t cursor_lsn)
```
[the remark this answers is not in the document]
Serpentian: Sure, let's change the name to "checkpoint join". I like it better too.

src/box/relay_file_join.h
Serpentian: > There's another way though: we could leave the join iterator lingering for a while after a failure (say for 1 minute) to let the recipient reconnect and resume from where it stopped.

This requires another RFC, we don't have time for that already. File snapshot fetching is already described in approved: [link](https://www.notion.so/tarantool/File-snapshot-fetching-443746c172274086be63c942f7fbd36f). Moreover, file snapshot fetching strongly connected to the persistent GC feature, we're currently implementing. 

The major advantage of files is that they don't take place in RAM unlike read-views. If we start saving read views, we'll have to deal with the maximum memory, which may be consumed by read-views, in order to avoid memory overflow. Files are ok, IMHO

src/box/iproto_constants.h
Serpentian: I dropped engine, thank you for an idea!!! From now on we rewrite LSNs of the vinyl engine to create an ascending sequence of LSNs in order to make cleint think, that this is just one big snap file.

src/box/iproto_constants.h
Serpentian: > I don't think you need a vclock to resume - signature would be enough.

It'll be used in persistent GC. Moreover, we send VCLOCK during JOIN, it won't be great to expect client to modify the data and send it back to us.

> The name "lsn" is confusing - it's a row number in the snapshot and should be named appropriately.

Let's not create a new iproto constant just for that, we know, that LSN is a number of row in snap.

> Why do you need to put these fields in a map? Why not store them at the top level?

Because in the future there may be other VCLOCK and LSN required for the more general case, than checkpoint cursor. If we put them at the top level, then we'll have to create new iproto constants for the future features, if they'll use VCLOCK and LSN.

> IPROTO_CURSOR is a too generic name - for example, it could be used for pagination.

Sure, let's change it to `IPROTO_CHECKPOINT_CURSOR`

src/box/relay.cc
Serpentian: I'll discuss this with Sergey Petrenko. Please, don't resolve the comment

src/box/box.cc
Serpentian: Maybe I splitted it a little bit too much. Now everything related to relay is in one commit.

src/box/relay_file_join.h
Gerold103: Tbh, I like Vova's idea. I don't think we considered it? The fact that the RFC is approved already is true, and it will allow us to get back to the file approach fast, but I would still be interested how much simpler the "lingering" cursors would be.

Memory-wise I wouldn't worry. We can limit the max number of cursors being open if it ever becomes an issue. Right now it wasn't ever a problem and it already works this way when joining not from a file. Would a cursor die after one minute or right away won't matter if join itself takes an hour or more. If it takes less, then the dataset is small, and it is as well unlikely to cause memory size issues.

Serpentian: > I don't think we considered it?

We didn't. Though, we'll be able to do it with the same API, proposed in RFC. I see no reason to write a new one, it's a question of implementation.

> Would a cursor die after one minute or right away won't matter if join itself takes an hour or more.

We'll have to allow client to configure how much time read view lives. So, if some big value is set, it may become an issue. The question, whether we should deal with this right now or later. IMHO, two new box.cfg options for read views are needed - max_memory, taken by all read views, and retention period.

---

@sergepetrenko, what do you think?  Whether I should write a new version with read views? It indeed seems it'll be easier in terms of implementation and subsequent support, as current implementation of vinyl file join is way to difficult and I see no way to simplify it. However, the current implementation with files works and fully complies with the API from RFC.

Serpentian: With read views we will also be able to drop opts from `_gc_consumers` and this cringe `with_snap` option. I'm voting for spending some time now and implementing read view retention period, as I don't want to support vinyl file join)

src/box/relay.cc
Serpentian: > Basically, it's normal join with the resume position.

But it doesn't have same logic: instead of syncing wal we find needed vclock from gc checkpoints, we don't waiting for limbo confirmation, we don't need to prepare read-views. The only thing in common - sending meta stage. Moreover, we need the shared context in case of file join in order to continue from the same LSN in vinyl.

I can add cursor to `engine_join_ctx` structure and pass it `engine_join_ctx`, but it doesn't seem it'll be better, due to a huge number of IFs in `relay_initial_join`, `engine_prepare/complete_join`, `engine_join_xc`.

Let's see, what others will say. IMHO, the separate path for file join and separate `checkpoint_initial_join` is better. @Gerold103 @sergepetrenko

src/box/applier.cc
```diff
@@ -861,9 +861,8 @@ applier_fetch_snapshot(struct applier *applier)
 	struct iostream *io = &applier->io;
 	struct xrow_header row;
 
-	struct fetch_snapshot_request req = {
-		.version_id = tarantool_version_id(),
-	};
+	struct fetch_snapshot_request req;
+	req.version_id = tarantool_version_id();
```
Gerold103, line 865: Why change it?

src/box/relay.cc
```diff
@@ -516,6 +545,21 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
...
 	engine_join_xc(&ctx, &relay->stream);
+	say_info("read-view sent.");
+}
+
+void
+relay_initial_join(struct iostream *io, uint64_t sync,
+		   struct vclock *vclock, uint32_t replica_version_id,
+		   const struct vclock *cursor_vclock, int64_t cursor_lsn)
+{
+	if (unlikely(cursor_vclock != NULL && vclock_is_set(cursor_vclock))) {
```
Gerold103, line 556: Nitpicking, but lets drop `unlikely()`. That is certainly one of the coldest paths in the codebase where an `if` optimization gives no impact.

Gerold103 (commented):
Thanks for the patch!

Another corporate blow for the poor dying CE.

src/box/relay.cc
Gerold103: I agree with Vova, looks like you could even reduce some code duplication if you make it at the engine level. Huge number of `if`s - how many? Lets give this way a chance, maybe it won't look bad?

src/box/iproto_constants.h
locker: > > Why do you need to put these fields in a map? Why not store them at the top level?
> 
> Because in the future there may be other VCLOCK and LSN required for the more general case, than checkpoint cursor. If we put them at the top level, then we'll have to create new iproto constants for the future features, if they'll use VCLOCK and LSN.

Why not just call the new constants `IPROTO_CHECKPOINT_VCLOCK` and `IPROTO_CHECKPOINT_LSN`? Putting them in a map complicates parsing (with a flat structure, you'd just put the new constants in the switch-case in `xrow_decode_replication_request`).

src/box/xrow.c
```diff
@@ -2054,14 +2058,16 @@ struct replication_request {
 	struct tt_uuid *instance_uuid;
 	/** IPROTO_INSTANCE_NAME. */
 	char *instance_name;
-	/** IPROTO_VCLOCK. */
+	/** IPROTO_VCLOCK or IPROTO_CHECKPOINT_CURSOR.VCLOCK. */
 	struct vclock *vclock;
```
locker, line 2062: Nit: Better add a new field to the struct for `checkpoint_vclock`.

src/box/memtx_engine.cc
```diff
@@ -1358,6 +1358,31 @@ memtx_engine_complete_join(struct engine *engine, void *arg)
...
+			     int64_t *lsn, int64_t start_lsn,
+			     struct xstream *stream)
+{
+	(void)engine;
+	(void)vclock;
+	(void)lsn;
+	(void)start_lsn;
+	(void)stream;
+	diag_set(ClientError, ER_UNSUPPORTED, "Tarantool CE",
+		 "checkpoint snapshot fetching");
```
locker, line 1380: Nit: This function isn't going to be called in CE so you could use the generic stub:

```C
#define memtx_engine_checkpoint_join generic_engine_checkpoint_join
```

src/box/iproto_constants.h
Serpentian: > Why not just call the new constants IPROTO_CHECKPOINT_VCLOCK and IPROTO_CHECKPOINT_LSN?

Done

src/box/relay.cc
Serpentian: Done this. scoped guards looks way to ugly

src/box/applier.cc
```diff
@@ -861,9 +861,8 @@ applier_fetch_snapshot(struct applier *applier)
 	struct iostream *io = &applier->io;
```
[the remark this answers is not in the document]
Serpentian: Fixed the commit msg

src/box/applier.cc
Serpentian: In order not to initialize all rows explicitly. `checkpoint_vclock` and `checkpoint_lsn` remains unitialized

src/box/relay.cc
Serpentian: Dropped

src/box/xrow.c
Serpentian: Done

src/box/memtx_engine.cc
Serpentian: Done

src/box/memtx_engine.cc
```diff
@@ -1093,10 +1110,17 @@ checkpoint_f(va_list ap)
...
+	/*
+	 * There may be no user data (e.g. when only RAFT_PROMOTE is written),
+	 * write limbo and raft states right after system spaces, at the end
+	 * of the checkpoint.
+	 */
+	if (!is_synchro_written) {
+		if (checkpoint_write_raft(snap, &ckpt->raft) != 0)
+			goto fail;
+		if (checkpoint_write_synchro(snap, &ckpt->synchro_state) != 0)
+			goto fail;
```
locker, line 1122: Let's put these two function calls in a new helper function (`checkpoint_write_system_data` or something like that).

test/engine-luatest/checkpoint_test.lua
```diff
@@ -0,0 +1,53 @@
...
+    local lsn = g.server:exec(function()
+        box.space.test:insert{1, 'data'}
+        box.snapshot()
+        return box.info.vclock[1]
+    end)
+
+    -- Read the whole snap in memory, as otherwise
+    -- it's impossible to access data by index.
+    local data = {}
+    local snap_template = '0000000000000000000%d.snap'
```
locker, line 33: `%020d.snap`

src/box/engine.c
```diff
@@ -235,14 +235,16 @@ engine_prepare_join(struct engine_join_ctx *ctx)
...
+engine_join(struct engine_join_ctx *ctx, struct engine_checkpoint_cursor *cur,
+	    struct xstream *stream)
 {
 	ERROR_INJECT_YIELD(ERRINJ_ENGINE_JOIN_DELAY);
 
 	int i = 0;
 	struct engine *engine;
 	engine_foreach(engine) {
-		if (engine->vtab->join(engine, ctx->array[i], stream) != 0)
+		void *ctx_object = ctx->array != NULL ? ctx->array[i] : NULL;
```
locker, line 246: I don't understand this change: `ctx->array` must be allocated because `engine_prepare_join` must be called first.

src/box/engine.h
```diff
@@ -88,6 +88,20 @@ struct engine_memory_stat {
 	size_t tx;
 };
 
+/**
+ * Cursor used during checkpoint initial join.
+ */
+struct engine_checkpoint_cursor {
+	/** Flag, which indicates, whether checkpoint join should be done. */
+	bool is_checkpoint_join;
```
locker, line 96: - I think we could name the structure simply `checkpoint_cursor`.
- I don't think we really need a flag here - instead we could pass `NULL` for `cursor` to `engine_join`.

src/box/memtx_engine.cc
```diff
@@ -1354,10 +1354,38 @@ memtx_join_f(va_list ap)
...
 }
 
+#if defined(ENABLE_CHECKPOINT_JOIN)
+#include "memtx_checkpoint_join.c"
+#else /* !defined(ENABLE_CHECKPOINT_JOIN) */
+
+#define memtx_engine_checkpoint_join generic_engine_join
+
+int
+memtx_engine_recover_synchro(struct memtx_engine *memtx,
```
locker, line 1364: AFAICS this function isn't used anywhere in CE so we don't need a stub for it.

src/box/memtx_engine.cc
```diff
@@ -1354,10 +1354,38 @@ memtx_join_f(va_list ap)
...
+
 static int
-memtx_engine_join(struct engine *engine, void *arg, struct xstream *stream)
+memtx_engine_join(struct engine *engine, void *arg,
+		  struct engine_checkpoint_cursor *cur, struct xstream *stream)
 {
-	(void)engine;
+	if (cur != NULL && cur->is_checkpoint_join)
+		/* Join from files. */
+		return memtx_engine_checkpoint_join(engine, arg, cur, stream);
```
locker, line 1386: I'd expect `ER_UNSUPPORTED` error here in CE, not the generic join implementation, which always succeeds.

**UPDATE:** After looking at the third patch in the series, I see that you raise this error in the relay code. Splitting this change in two patches actually makes it hard to review.

src/box/box.cc
```diff
@@ -4312,12 +4312,27 @@ box_process_fetch_snapshot(struct iostream *io,
...
+		.lsn_counter = 0,
+	};
+	struct gc_checkpoint_ref gc;
+	auto gc_guard = make_scoped_guard([&] {
+		gc_unref_checkpoint(&gc);
+	});
+	gc_guard.is_active = false;
+	if (cursor.is_checkpoint_join) {
+		if (relay_prepare_checkpoint_for_join(&cursor, &gc) != 0)
+			diag_raise();
```
locker, line 4329: Referencing the last checkpoint is really trivial. Let's implement it  completely in CE, without using the stubs, and let the engines raise the `ER_UNSUPPORTED` error in CE.

src/box/iproto_constants.h
```diff
@@ -218,7 +218,22 @@ extern const char *iproto_flag_bit_strs[];
 	/**
 	 * Flag indicating whether the transaction is synchronous.
 	 */								\
-	 _(IS_SYNC, 0x61, MP_BOOL)
+	 _(IS_SYNC, 0x61, MP_BOOL)					\
+	 /**
+	  * Flag indicating whether checkpoint join should be done.
+	  */								\
+	 _(CHECKPOINT_JOIN, 0x62, MP_BOOL)				\
```
locker, line 225: I guess it should be named `IS_CHECKPOINT_JOIN` according to our naming conventions.

src/box/box.cc
```diff
@@ -4312,12 +4312,27 @@ box_process_fetch_snapshot(struct iostream *io,
...
+		gc_guard.is_active = true;
+	}
 	/* Send the snapshot data to the instance. */
 	struct vclock start_vclock;
-	relay_initial_join(io, header->sync, &start_vclock, req.version_id);
-	say_info("read-view sent.");
```
locker, line 4320: It's fine to leave this message here - a checkpoint is a read view.

src/box/relay.cc
```diff
@@ -453,33 +492,50 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
...
+	 * of the function. It works only during read-view join.
+	 */
 	auto join_guard = make_scoped_guard([&] {
 		engine_complete_join(&ctx);
 	});
+	if (cursor != NULL && cursor->is_checkpoint_join) {
+		join_guard.is_active = false;
+		/** Checkpoint initial join is going to be done. */
+		int rc = relay_prepare_checkpoint_initial_join(cursor, vclock,
+							       &raft_req, &req);
```
locker, line 511: Why not do it in the memtx engine callback?

src/box/relay.cc
```diff
@@ -453,33 +492,50 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
...
+		if (txn_limbo_wait_confirm(&txn_limbo) != 0)
+			diag_raise();
 
-	struct synchro_request req;
-	struct raft_request raft_req;
-	struct vclock limbo_vclock;
-	txn_limbo_checkpoint(&txn_limbo, &req, &limbo_vclock);
-	box_raft_checkpoint_local(&raft_req);
+		txn_limbo_checkpoint(&txn_limbo, &req, &limbo_vclock);
+		box_raft_checkpoint_local(&raft_req);
```
locker, line 537: Do we really need **not** to do it in case of checkpoint join? The code would be simpler if we did it unconditionally for both types of join.

locker (changes requested):

src/box/applier.cc
locker: Do we really need them to remain uninitialized? What's wrong with setting them to 0?

src/box/memtx_engine.cc
```diff
@@ -1036,6 +1037,22 @@ checkpoint_f(va_list ap)
 		});
 		if (skip)
 			continue;
+		/*
+		 * Raft and limbo states are written right after system spaces
+		 * but before user ones. This is needed to reduce the time,
+		 * needed to acquire states from the checkpoint, which is used
+		 * during checkpoint join.
```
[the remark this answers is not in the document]
Serpentian: `force_recovery` is not needed. The change is not breaking, old Tarantool version can recover from such snap without any problems. We should not introduce breaking changes in this patch, as it'll be backported to 2.11. That's the reason, why raft and limbo rows are after system spaces and not at the beginning of the snap.

src/box/memtx_engine.cc
Serpentian: But it's used in EE. It just seems not right to put such function in `relay_checkpoint_join.cc`. Morever, as it does pretty the same work as `memtx_engine_checkpoint_join`, I made a common part for them, which recovers snap in the separate thread. If I put it in `relay_checkpoint_join.cc` then it'll be a major code duplication. I don't like that a lot

src/box/relay.cc
Serpentian: From Vladislav:

> The problem is that you are dumping the latest state, while the checkpoint might be old and have an outdated state. For example, the checkpoint might be from term 10, while the latest term is 100. Then how will the subscribe work after such join, when the replica will be receiving stuff from "outdated" terms?

And I agree with him, we can **not** send the latest state we have. In such case replica will have term 100 with data from the 10th term. After that during subscribe it'll get term 11 and this will be lowering term from replica's point of view, which cannot happen at all. Let's **not** do that just for code simplicity, we must recover states from the snap, we're sending

src/box/relay.cc
Serpentian: Because it does much more, than callback, it recovers limbo and raft states, which are needed exclusively by relay and cannot be put inside `engine_join_ctx`, this is not a relay part. Vclock is also taken from checkpoint and not from the WAL. That's the reason, why initially everything was done as a separate function, logic differs too much

src/box/iproto_constants.h
Serpentian: I tried to be consistent with `FETCH_POSITION`, `REPLICA_ANON`. The only flag prefixed with `IS_` is `IS_SYNC`. Not really sure, that prefixing in iproto flags is out naming convention

src/box/engine.c
locker: I see that the problem is that you don't call `engine_prepare_join`. That's confusing. Can we call `engine_prepare_join` and `engine_complete_join` for both checkpoint and read view join?

src/box/relay.cc
Serpentian: > What do you think about moving the code sending the RAFT state to memtx_engine_join? The memtx engine is already aware about RAFT so it should be fine IMO. This would make the relay code handle checkpoint join the same way as read view join, without confusing conditions.

Let's do that

src/box/relay.cc
locker: OK, I understand now. My suggestion is to consider moving the code sending the RAFT state completely to memtx, see https://github.com/tarantool/tarantool/pull/9973/files#r1627409862.

BTW, we already have a function named `memtx_engine_recover_synchro` in `memtx_engine.cc`. The new function works only because C++ argument overloading, but we don't use that in Tarantool. Better rename it.

src/box/relay.cc
Serpentian: > What do you think about moving the code sending the RAFT state to memtx_engine_join? 

@locker, maybe `engine_prepare_join` and `engine_join` are the better places for that? IMHO, sending vclock and `META_JOIN` stage are the common parts for all engine and we should not put them right into memtx, even though we know, that memtx is send firstly during join. It's completely not memtx related.

locker: `engine_prepare_join` and `engine_join` (as well as other `engine_*` methods) are supposed to be thin wrappers around engine callbacks. I think that sending the RAFT state in the memtx engine callback is fine because memtx is special: it stores system spaces and system data. Sending the RAFT state from `memtx_engine_join` would only be consistent with `memtx_engine_checkpoint`, which stores the RAFT state in the snapshot.

src/box/applier.cc
Serpentian: Now they're intiialized

src/box/memtx_engine.cc
Serpentian: Done

test/engine-luatest/checkpoint_test.lua
Serpentian: Fixed

src/box/engine.c
Serpentian: > Can we call engine_prepare_join and engine_complete_join for both checkpoint and read view join?

Done

src/box/engine.h
Serpentian: Both points were fixed

src/box/memtx_engine.cc
Serpentian: > Splitting this change in two patches actually makes it hard to review.

Hope, it's easier to review now)

src/box/box.cc
Serpentian: I managed to properly split CE and EE part without moving "relay" part to CE

src/box/iproto_constants.h
Serpentian: Done this actually

src/box/box.cc
Serpentian: No problem, the msg untouched

src/box/engine.h
```diff
@@ -312,6 +314,17 @@ struct engine_read_view {
 	struct rlist link;
 };
 
+/**
+ * Argument for engine_prepare_join functions, shared between engines.
+ * Used to pass arguments to join context setup.
+ */
+struct engine_prepare_join_arg {
```
Serpentian, line 333: Maybe smb have a better idea to pass arguments? This way it's easy to extend

locker (changes requested):

src/box/memtx_engine.cc
```diff
@@ -1285,18 +1297,43 @@ memtx_engine_prepare_join(struct engine *engine, void **arg)
...
-		free(ctx);
-		return -1;
+	if (read_view_open(&ctx->rv, &rv_opts) != 0)
+		goto fail;
+	/*
+	 * Sync WAL to make sure that all changes visible from
+	 * the frozen read view are successfully committed and
+	 * obtain corresponding vclock.
+	 */
+	if (wal_sync(ctx->vclock) != 0)
```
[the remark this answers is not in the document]
Serpentian: Good catch! The same applies for `limbo_wait_confirm`. This makes us to move vclock, limbo and raft states preparation to `engine_join`

src/box/memtx_engine.cc
```diff
@@ -1285,18 +1297,43 @@ memtx_engine_prepare_join(struct engine *engine, void **arg)
 			 "malloc", "struct memtx_join_ctx");
 		return -1;
 	}
+	ctx->vclock = arg->vclock;
+	ctx->is_meta_send = arg->is_meta_send;
```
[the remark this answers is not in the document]
Serpentian: Now it's called `send_meta`. While we discussed refactoring with Sergey Petrenko, he said, that `replica_version_id` have no relation to `engine` part, it's replication one. And I agree with him, let's not do that and just find a better way

src/box/vinyl.c
```diff
@@ -3061,8 +3061,10 @@ vy_join_add_space(struct space *space, void *arg)
 }
 
 static int
-vinyl_engine_prepare_join(struct engine *engine, void **arg)
+vinyl_engine_prepare_join(struct engine *engine, void **ctx_arg,
+			  struct engine_prepare_join_arg *arg)
```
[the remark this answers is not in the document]
Serpentian: Done, good point

src/box/iproto_constants.h
```diff
@@ -218,7 +218,22 @@ extern const char *iproto_flag_bit_strs[];
...
+	  * Shows the signature of the checkpoint to read from.
+	  * Requires CHECKPOINT_JOIN to be true.
+	  */								\
+	 _(CHECKPOINT_VCLOCK, 0x63, MP_MAP)				\
+	 /**
+	  * Shows the lsn to start sending from. Server sends all rows
+	  * >= IPROTO_CHECKPOINT_LSN. Requires CHECKPOINT_JOIN to be
+	  * true and CHECKPOINT_VCLOCK to be set.
+	  */								\
+	 _(CHECKPOINT_LSN, 0x64, MP_UINT)				\
```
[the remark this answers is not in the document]
Serpentian: Added the feature

src/box/xrow.c
```diff
@@ -2274,6 +2280,35 @@ id_filter_decode_err:		xrow_on_decode_err(row, ER_INVALID_MSGPACK,
 				*req->id_filter |= 1 << val;
 			}
 			break;
+		case IPROTO_IS_CHECKPOINT_JOIN:
+			if (req->is_checkpoint_join == NULL)
+				goto skip;
+			if (mp_typeof(*d) != MP_BOOL) {
+				xrow_on_decode_err(row, ER_INVALID_MSGPACK,
+						   "invalid CHECKPOINT_JOIN");
```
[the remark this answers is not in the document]
Serpentian: Fixed

src/box/vinyl.c
```diff
@@ -3102,12 +3105,33 @@ vy_join_send_tuple(struct xstream *stream, uint32_t space_id,
...
+{
+	(void)engine;
+	(void)ctx;
+	(void)stream;
+	diag_set(ClientError, ER_UNSUPPORTED, "Tarantool CE",
+		 "checkpoint join");
+	return -1;
+}
+
+#endif /* !defined(ENABLE_CHECKPOINT_JOIN) */
```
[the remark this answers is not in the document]
Serpentian: My bad, fixed

src/box/iproto_features.h
```diff
@@ -77,6 +77,12 @@ extern "C" {
 	 * tuple formats are received in IPROTO_TUPLE_FORMATS field.
 	 */								\
 	_(CALL_ARG_TUPLE_EXTENSION, 9)					\
+	/**
+	 * Checkpoint (from files) FETCH_SNAPSHOT request support:
+	 * IPROTO_IS_CHECKPOINT_JOIN, IPROTO_CHECKPOINT_VCLOCK and
+	 * IRPOTO_CHECKPOINT_LSN in FETCH_SNAPSHOT.
+	 */								\
+	_(CHECKPOINT_FETCH_SNAPSHOT, 10)				\
```
locker, line 85: The name is kinda awkward. I think we'd better name it `FETCH_SNAPSHOT_CURSOR` to emphasize that it's an extension of the `IPROTO_FETCH_SNAPSHOT` request.

Also, I think we should rename `ENABLE_CHECKPOINT_JOIN` to `ENABLE_FETCH_SNAPSHOT_CURSOR` for consistency and because `IPROTO_JOIN` doesn't support the feature (initially, I thought it'd support it). Using `checkpoint_join` in engine callback names is fine because they should work both for `IPROTO_FETCH_SNAPSHOT` and `IPROTO_JOIN`.

locker (changes requested):

src/box/memtx_engine.cc
```diff
@@ -1416,8 +1476,16 @@ memtx_engine_join(struct engine *engine, void *arg, struct xstream *stream)
 	 * All these versions know of additional META stage of initial join.
 	 */
 	if (ctx->send_meta) {
-		txn_limbo_checkpoint(&txn_limbo, &synchro_req, &limbo_vclock);
-		box_raft_checkpoint_local(&raft_req);
+		if (ctx->cursor != NULL) {
+			int rc = memtx_engine_recover_synchro_states(
+				engine, ctx->vclock, &raft_req, &synchro_req);
+			if (rc != 0)
+				return -1;
```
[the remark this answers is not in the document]
Serpentian: > You wouldn't need to store the vclock in the join context then.

The problem here's that I have to pass it back, from engine to box. This one is `start_vclock`, which is used in `relay_final_join` to start from. So it seems we should pass it anyway through "engine-independent" part to properly initialize it in engines.

src/box/memtx_engine.h
```diff
@@ -188,6 +188,14 @@ struct memtx_engine {
 	int sort_threads;
 };
 
+/** Data for initialization of the join context. */
+struct memtx_prepare_data {
+	/** Vclock to respond with. */
+	struct vclock *vclock;
+	/** Whether sending JOIN_META stage is required. */
+	bool send_meta;
+};
```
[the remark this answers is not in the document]
Serpentian: But actually the only "engine-independent" part is `cursor`. `send_meta` and `vclock` are related only to memtx engine, vinyl doesn't need them at all.

src/box/memtx_engine.cc
```diff
@@ -1293,18 +1345,26 @@ memtx_engine_prepare_join(struct engine *engine, void **arg, void *arg_data)
 	struct memtx_prepare_data *data = (struct memtx_prepare_data *)arg_data;
 	ctx->vclock = data->vclock;
 	ctx->send_meta = data->send_meta;
+	ctx->cursor = data->cursor;
+	if (ctx->cursor != NULL) {
+		if (memtx_engine_prepare_checkpoint_join(ctx) != 0)
+			goto fail;
+		goto success;
+	}
```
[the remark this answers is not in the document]
Serpentian: You mean passing variables directly to `memtx_engine_prepare/complete_join` and not the whole `ctx`?

src/box/memtx_engine.h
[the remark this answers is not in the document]
locker: By "engine-independent" I mean that it can be set pr used outside the engine. Both `send_meta` and `vclock` can. It's okay that they aren't used by Vinyl.

src/box/memtx_engine.cc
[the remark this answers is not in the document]
locker: Yes, passing all arguments.

```C
static int
memtx_engine_prepare_join(struct engine *engine, struct engine_join_ctx *ctx)
{
        if (ctx->cursor != NULL)
                return memtx_engine_prepare_checkpoint_join(engine, ctx);
        ...
}

static int
memtx_engine_complete_join(struct engine *engine, struct engine_join_ctx *ctx)
{
        if (ctx->cursor != NULL)
                return memtx_engine_complete_checkpoint_join(engine, ctx);
        ...
}
```

src/box/memtx_engine.h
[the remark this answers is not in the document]
Serpentian: Indeed, misunderstood you, it's better this way. Moreover, we don't need to allocate contexts of the engines for checkpoint join. Fixed.

src/box/memtx_engine.cc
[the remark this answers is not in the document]
Serpentian: Good proposal, done

src/box/memtx_engine.cc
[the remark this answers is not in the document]
Serpentian: Hope, I understood you right. Now I don't "share context" and immediately call `<...>_engine_<...>_checkpoint_join` everywhere I can

src/box/vinyl.c
```diff
@@ -3132,6 +3162,9 @@ vinyl_engine_complete_join(struct engine *engine, void *arg)
 {
 	struct vy_env *env = vy_env(engine);
 	struct vy_join_ctx *ctx = arg;
+	if (ctx->cursor != NULL)
+		return;
```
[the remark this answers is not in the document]
Serpentian: Done

src/box/memtx_engine.cc
```diff
@@ -1264,11 +1264,63 @@ struct memtx_join_ctx {
...
+static int
+memtx_engine_checkpoint_join(struct engine *engine, struct memtx_join_ctx *ctx,
+			     struct xstream *stream)
+{
+	(void)engine;
+	(void)ctx;
+	(void)stream;
+	diag_set(ClientError, ER_UNSUPPORTED, "Tarantool CE",
+		 "checkpoint join");
+	return -1;
```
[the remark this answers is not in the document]
Serpentian: `complete` callbacks are always reachable, we get there if `prepare` fails

https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/engine.c#L225-L235

The only unreachable is CE `memtx/vinyl_engine_checkpoint_join`. Added

src/box/iproto_features.c
```diff
@@ -85,4 +85,6 @@ iproto_features_init(void)
 			    IPROTO_FEATURE_CALL_RET_TUPLE_EXTENSION);
 	iproto_features_set(&IPROTO_CURRENT_FEATURES,
 			    IPROTO_FEATURE_CALL_ARG_TUPLE_EXTENSION);
+	iproto_features_set(&IPROTO_CURRENT_FEATURES,
+			    IPROTO_FEATURE_CHECKPOINT_FETCH_SNAPSHOT);
```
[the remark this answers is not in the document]
Serpentian: Done. Also added the test for the feature in EE. `IPROTO_CURRENT_FEATURES` are used only for responding on `IPROTO_ID`. In order to exclude them from `box.iproto.feature/protocol_features` we should also exclude them from `iproto_feature_id`.

src/box/box.cc
```diff
@@ -4311,10 +4311,20 @@ box_process_fetch_snapshot(struct iostream *io,
 	}
 
 	say_info("sending current read-view to replica at %s", sio_socketname(io->fd));
```
locker, line 4313: Nit: I'd remove "current" from the log message and say simply "sending read-view to replica at".

src/box/engine.h
```diff
@@ -315,11 +316,27 @@ struct engine_read_view {
...
+
 struct engine_join_ctx {
 	/** Vclock to respond with. */
 	struct vclock *vclock;
 	/** Whether sending JOIN_META stage is required. */
 	bool send_meta;
+	/** Checkpoint join cursor. */
+	struct checkpoint_cursor *cursor;
+	/** Ref to protect checkpoint from gc during checkpoint join. */
+	struct gc_checkpoint_ref gc;
```
locker, line 339: I don't like that `engine_join_ctx::gc` is defined in CE but used only in EE. Let's store it in a context allocated in EE.

src/box/iproto_features.c
```diff
@@ -14,6 +14,9 @@
 
 const char *iproto_feature_id_strs[iproto_feature_id_MAX] = {
 	IPROTO_FEATURES(IPROTO_FEATURE_ID_STRS_MEMBER)
+#if defined(ENABLE_FETCH_SNAPSHOT_CHECKPOINT_CURSOR)
+	IPROTO_FEATURE_ID_STRS_MEMBER(FETCH_SNAPSHOT_CHECKPOINT_CURSOR, 10)
+#endif /* defined(ENABLE_FETCH_SNAPSHOT_CHECKPOINT_CURSOR) */
```
locker, line 19: IMO the name is awkward because "checkpoint" and "snapshot" are synonyms in Tarantool. I still think `ENABLE_FETCH_SNAPSHOT_CURSOR` is a better name but I'm okay with it if other reviewers think it's okay.

src/box/iproto_features.h
```diff
@@ -77,11 +77,19 @@ extern "C" {
 	 * tuple formats are received in IPROTO_TUPLE_FORMATS field.
 	 */								\
 	_(CALL_ARG_TUPLE_EXTENSION, 9)					\
+	/**
+	 *_(FETCH_SNAPSHOT_CHECKPOINT_CURSOR, 10) - only in EE:
+	 * IPROTO_IS_CHECKPOINT_JOIN, IPROTO_CHECKPOINT_VCLOCK and
+	 * IRPOTO_CHECKPOINT_LSN in FETCH_SNAPSHOT.
+	 */								\
```
locker, line 84: We do need to define the feature bit unconditionally and show it in `box.iproto.feature`. We just shouldn't send it in reply to `IPROTO_ID` or show it in `box.iproto.protocol_features`.

src/box/memtx_engine.cc
```diff
@@ -1265,6 +1265,84 @@ struct memtx_join_ctx {
 	struct xstream *stream;
 };
 
+static void
+send_join_meta(struct xstream *stream, const struct vclock *vclock,
+	       const bool send_meta, const struct raft_request *raft_req,
+	       const struct synchro_request *synchro_req)
```
locker, line 1271: This function was introduced in a previous commit and now you move it. Please don't do that. Instead, add it to the right place in the commit that introduced it.

src/box/memtx_engine.cc
```diff
@@ -1265,6 +1265,84 @@ struct memtx_join_ctx {
...
+#else /* !defined(ENABLE_FETCH_SNAPSHOT_CHECKPOINT_CURSOR) */
+
+static int
+memtx_engine_prepare_checkpoint_join(struct engine *engine,
+				     struct engine_join_ctx *ctx)
+{
+	(void)engine;
+	(void)ctx;
+	diag_set(ClientError, ER_UNSUPPORTED, "Tarantool CE",
+		 "checkpoint join");
```
locker, line 1321: Nit: Usually we say that "Community edition does not support <feature>", not "Tarantool CE":

https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/lua/schema.lua#L924

https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/lua/schema.lua#L4090

src/box/vinyl.c
```diff
@@ -3029,6 +3029,42 @@ struct vy_join_ctx {
...
+#else /* !defined(ENABLE_FETCH_SNAPSHOT_CHECKPOINT_CURSOR) */
+
+static int
+vinyl_engine_prepare_checkpoint_join(struct engine *engine,
+				     struct engine_join_ctx *ctx)
+{
+	(void)engine;
+	(void)ctx;
+	diag_set(ClientError, ER_UNSUPPORTED, "Tarantool CE",
+		 "checkpoint join");
```
locker, line 3043: Nit: This function is unreachable because `engine_prepare_join` fails on memtx.

src/box/memtx_engine.cc
```diff
@@ -1457,7 +1498,11 @@ memtx_engine_join(struct engine *engine, struct engine_join_ctx *arg,
 static void
 memtx_engine_complete_join(struct engine *engine, struct engine_join_ctx *arg)
 {
-	(void)engine;
+	if (arg->cursor != NULL) {
+		memtx_engine_complete_checkpoint_join(engine, arg);
+		return;
```
locker, line 1503: Nit: In C it's possible to write this in one line, like in `memtx_engine_prepare_join`:

```C
	if (arg->cursor != NULL)
		return memtx_engine_complete_checkpoint_join(engine, arg);
```

In `vinyl_engine_complete_join`, too.

test/box-luatest/gh_7894_export_iproto_constants_and_features_test.lua
```diff
@@ -3,6 +3,8 @@ local t = require('luatest')
 
 local g = t.group()
 
+local is_enterprise = require('tarantool').package == 'Tarantool Enterprise'
```
locker, line 6: Nit: Better use the luatest helper `t.tarantool.is_enterprise_package()`.

src/box/memtx_engine.cc
```diff
@@ -1360,12 +1361,79 @@ memtx_join_f(va_list ap)
...
+	struct xrow_header row;
+	/* Encoding replication request uses fiber()->gc region. */
+	RegionGuard region_guard(&fiber()->gc);
+	xrow_encode_vclock(&row, vclock);
+	xstream_write(stream, &row);
+
+	/*
+	 * Version is present starting with 2.7.3, 2.8.2, 2.9.1
+	 * All these versions know of additional META stage of initial join.
+	 */
```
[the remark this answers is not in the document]
Serpentian: Fixed. The comment is in `relay.cc`

src/box/memtx_engine.cc
```diff
@@ -1360,12 +1361,79 @@ memtx_join_f(va_list ap)
...
+		xstream_write(stream, &row);
+	}
+}
+
 static int
-memtx_engine_join(struct engine *engine, void *arg, struct xstream *stream)
+memtx_engine_join(struct engine *engine, struct engine_join_ctx *arg,
+		  struct xstream *stream)
 {
 	(void)engine;
```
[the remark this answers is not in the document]
Serpentian: Missed that. Fixed

src/box/box.cc
Serpentian: Removed

src/box/engine.h
Serpentian: Good idea. The code is cleaner now

src/box/iproto_features.c
Serpentian: I don't care that much about feature's naming. Let's use `ENABLE_FETCH_SNAPSHOT_CURSOR`

src/box/iproto_features.h
Serpentian: Done

src/box/memtx_engine.cc
Serpentian: My bad, fixed

src/box/memtx_engine.cc
Serpentian: Indeed, fixed

src/box/vinyl.c
Serpentian: Fixed

test/box-luatest/gh_7894_export_iproto_constants_and_features_test.lua
Serpentian: Done

locker (approved):

src/box/box.cc
```diff
@@ -4310,7 +4310,7 @@ box_process_fetch_snapshot(struct iostream *io,
 			  "wal_mode = 'none'");
 	}
 
-	say_info("sending current read-view to replica at %s", sio_socketname(io->fd));
+	say_info("sending read-view to replica at %s", sio_socketname(io->fd));
```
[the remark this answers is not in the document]
Serpentian: Fixed

src/box/lua/iproto.c
```diff
@@ -175,29 +175,48 @@ push_iproto_constants(struct lua_State *L)
...
-	lua_setfield(L, -2, "protocol_version");
-
-	for (int i = 0; i < 2; i++)
-		lua_newtable(L);
+	lua_newtable(L);
 	for (int i = 0; i < iproto_feature_id_MAX; i++) {
+#if !defined(ENABLE_FETCH_SNAPSHOT_CURSOR)
+		if (i == IPROTO_FEATURE_FETCH_SNAPSHOT_CURSOR)
+			continue;
+#endif /* !defined(ENABLE_FETCH_SNAPSHOT_CURSOR) */
```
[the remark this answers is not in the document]
Serpentian: Done. Please, note, that this also requires putting `iproto_features_init` in `box_init`. As `box.iproto` is initiialized in `box_lua_init` we must init features before, doing this during `box.cfg` won't work.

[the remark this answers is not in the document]
locker: That's fine, but please remove `iproto_features_init` from `iproto_init` then.

locker (approved):

src/box/xrow.c
```diff
@@ -2541,6 +2567,27 @@ xrow_encode_vclock_ignore0(struct xrow_header *row, const struct vclock *vclock)
...
+	data = mp_encode_map(data, map_size);
+	data = mp_encode_uint(data, IPROTO_VCLOCK);
+	data = mp_encode_vclock(data, vclock);
+	assert(data <= buf + size);
+	row->body[0].iov_base = buf;
+	row->body[0].iov_len = (data - buf);
+	row->bodycnt = 1;
+}
+
 int
```
Gerold103, line 2638: Please, reuse `replication_request` + `xrow_encode_replication_request`. You can extend `replication_request` with ignore0 flag, for example. Or make it have two optional fields - `struct vclock *vclock;` and `struct vclock *vclock_no0;`.

src/box/applier.cc
```diff
@@ -863,8 +863,14 @@ applier_fetch_snapshot(struct applier *applier)
...
 	struct xrow_header row;
 
+	struct vclock vclock;
+	vclock_create(&vclock);
 	struct fetch_snapshot_request req = {
 		.version_id = tarantool_version_id(),
+		/* Applier doesn't support checkpoint join. */
+		.is_checkpoint_join = false,
+		.checkpoint_vclock = vclock,
+		.checkpoint_lsn = 0,
```
Gerold103, line 873: Why do you specify empty vclock then? Is this field mandatory? Otherwise you can drop these initializations from here. The struct initialization in the form of `var = { .field1 = ..., .field2 = ... ...}` will nullify all non-specified fields.

src/box/box.cc
```diff
@@ -4310,11 +4310,21 @@ box_process_fetch_snapshot(struct iostream *io,
...
+	/* Used for checkpoint initial join. */
+	struct checkpoint_cursor cursor;
+	struct checkpoint_cursor *cursor_ptr = NULL;
+	if (req.is_checkpoint_join) {
+		cursor.vclock = &req.checkpoint_vclock;
+		cursor.start_lsn = req.checkpoint_lsn;
+		cursor.lsn_counter = 0;
+		cursor_ptr = &cursor;
+	}
```
Gerold103, line 4322: Please, either assign the struct via `var = {.field1 = ..., .field2 = .... }` or make a `memset(0)` before filling it. Otherwise we have a high risk to get corrupted data here when the struct will grow in the future.

src/box/xrow.c
```diff
@@ -2386,7 +2422,18 @@ xrow_decode_fetch_snapshot(const struct xrow_header *row,
...
+		.checkpoint_lsn = &req->checkpoint_lsn,
 	};
+	/*
+	 * IS_CHECKPOINT_JOIN, CHECKPOINT_VCLOCK and CHECKPOINT_LSN are optional
+	 * keys, they are used exclusively by CDC now, so we need to set
+	 * default values for vclock and lsn before decoding.
+	 */
+	req->is_checkpoint_join = false;
+	vclock_clear(&req->checkpoint_vclock);
+	req->checkpoint_lsn = 0;
```
Gerold103, line 2436: There is `memset(0)` above which does all the same. These lines don't seem necessary.

Gerold103 (commented):
Thanks for the fixes, looks very clean and understandable now!

src/box/applier.cc
Serpentian: > The struct initialization in the form of var = { .field1 = ..., .field2 = ... ...} will nullify all non-specified fields.

Our compilation flags forbid doing this:

```
/home/serpentian/Programming/tnt/tarantool-ee/tarantool/src/box/applier.cc: In function ‘void applier_fetch_snapshot(applier*)’:
/home/serpentian/Programming/tnt/tarantool-ee/tarantool/src/box/applier.cc:874:9: error: missing initializer for member ‘fetch_snapshot_request::checkpoint_vclock’ [-Werror=missing-field-initializers]
  874 |         };
      |         ^
```

Moreover, you and Vladimir were against unitialized fields in https://github.com/tarantool/tarantool/pull/9973#discussion_r1605324874.

I dropped `vclock_create` since it really is not mandatory, but the structure must be fully initialized, if we don't wanna use `req.checkpoint_lsn = ...`

src/box/xrow.c
Serpentian: `vclock_clear` does a lot more:

https://github.com/tarantool/tarantool/blob/7db4de75f8bd975066635f942ee49f33a2209353/src/lib/vclock/vclock.h#L128-L134

This -1 signature is what I need. `memset` won't do that for me. And I strongly rely on it in `memtx_engine_prepare_checkpoint_join`

So, let's leave everything as it is.

src/box/box.cc
Serpentian:
```
/home/serpentian/Programming/tnt/tarantool-ee/tarantool/src/box/box.cc: In function ‘void box_process_fetch_snapshot(iostream*, const xrow_header*)’:
/home/serpentian/Programming/tnt/tarantool-ee/tarantool/src/box/box.cc:4321:42: error: narrowing conversion of ‘req.fetch_snapshot_request::checkpoint_lsn’ from ‘uint64_t’ {aka ‘long unsigned int’} to ‘int64_t’ {aka ‘long int’} [-Werror=narrowing]
 4321 |                         .start_lsn = req.checkpoint_lsn,
      |                                      ~~~~^~~~~~~~~~~~~~
At global scope:
```

src/box/memtx_engine.cc
```diff
@@ -1093,10 +1118,15 @@ checkpoint_f(va_list ap)
...
-		goto fail;
+	/*
+	 * There may be no user data (e.g. when only RAFT_PROMOTE is written),
+	 * write limbo and raft states right after system spaces, at the end
+	 * of the checkpoint.
+	 */
+	if (!is_synchro_written)
+		if (checkpoint_write_system_data(snap, &ckpt->raft,
+						 &ckpt->synchro_state) != 0)
+			goto fail;
```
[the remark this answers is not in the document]
Serpentian: Fixed

src/box/memtx_engine.cc
```diff
@@ -1264,6 +1265,37 @@ struct memtx_join_ctx {
 	struct xstream *stream;
 };
 
+static void
+send_join_meta(struct xstream *stream, const struct vclock *vclock,
+	       const bool send_meta, const struct raft_request *raft_req,
+	       const struct synchro_request *synchro_req)
```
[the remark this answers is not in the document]
Serpentian: Splitted the function into `send_join_header` and `send_join_meta`

src/box/relay.cc
```diff
@@ -455,60 +455,16 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 
 	/* Freeze a read view in engines. */
 	struct engine_join_ctx ctx;
```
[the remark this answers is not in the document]
Serpentian: Done

src/box/xrow.c
Serpentian: Reused the `replication_request`. Introduced `vclock_ignore0` and `vclock` in order to be consistent with function naming

src/box/memtx_engine.cc
```diff
@@ -1264,6 +1265,37 @@ struct memtx_join_ctx {
...
+		xstream_write(stream, &row);
+
+		char body[XROW_BODY_LEN_MAX];
+		xrow_encode_synchro(&row, body, synchro_req);
+		row.replica_id = synchro_req->replica_id;
+		xstream_write(stream, &row);
+
+		/* Mark the end of the metadata stream. */
+		xrow_encode_type(&row, IPROTO_JOIN_SNAPSHOT);
+		xstream_write(stream, &row);
```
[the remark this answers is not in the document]
Gerold103: Then send it conditionally? My main point is that `IPROTO_JOIN_SNAPSHOT` is not the end of meta section. It is beginning of snapshot data. Which means it can't be in this function which is only for sending the meta.

src/box/box.cc
```diff
@@ -4310,11 +4310,22 @@ box_process_fetch_snapshot(struct iostream *io,
...
-	say_info("sending current read-view to replica at %s", sio_socketname(io->fd));
+	say_info("sending read-view to replica at %s", sio_socketname(io->fd));
+	/* Used for checkpoint initial join. */
+	struct checkpoint_cursor cursor;
+	struct checkpoint_cursor *cursor_ptr = NULL;
+	if (req.is_checkpoint_join) {
+		memset(&cursor, 0, sizeof(cursor));
+		cursor.vclock = &req.checkpoint_vclock;
+		cursor.start_lsn = req.checkpoint_lsn;
+		cursor.lsn_counter = 0;
```
Gerold103, line 4321: This isn't needed after you did `memset` above. It is already 0.

src/box/memtx_engine.cc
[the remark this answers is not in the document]
Serpentian: > Then send it conditionally?

I see, fixed

Gerold103 (approved):
Thanks for the fixes! Good job, we are lucky to have you on the team 💪.

src/box/xrow.c
Gerold103: But you surely don't need `is_checkpoint_join = false;` and `checkpoint_lsn = 0;`?

src/box/memtx_engine.cc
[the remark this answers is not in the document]
Gerold103: Thanks for this small detail 🙏!

src/box/xrow.c
Serpentian: I don't) Fixed

src/box/box.cc
Serpentian: `lsn_counter` initialization was dropped

src/box/xrow.c
```diff
@@ -90,8 +97,22 @@ mp_encode_vclock_ignore0(char *data, const struct vclock *vclock)
...
+
+static inline char *
+mp_encode_vclock(char *data, const struct vclock *vclock)
+{
+	data = mp_encode_map(data, vclock_size(vclock));
+	return mp_encode_vclock_impl(data, vclock, false);
+}
+
 static int
-mp_decode_vclock_ignore0(const char **data, struct vclock *vclock)
```
sergepetrenko, line 94: It seems you don't use `mp_decode_vlcock` in this patch. Only the `mp_decode_vclock_ignore0`.
Let's leave only that version. So that noone decodes the vclock with 0th component accidentally

src/box/applier.cc
```diff
@@ -863,8 +863,14 @@ applier_fetch_snapshot(struct applier *applier)
 	struct iostream *io = &applier->io;
```
sergepetrenko, line 863: Typo in the commit message:
```
 engine: introduce stubs for checkpoint FETCH_SNAPSHOT

This commit introduces engine jstubs that enable a new method
                              ^^^^^^
```

src/box/xrow.c
```diff
@@ -2400,7 +2436,16 @@ xrow_decode_fetch_snapshot(const struct xrow_header *row,
...
+		.is_checkpoint_join = &req->is_checkpoint_join,
+		.checkpoint_vclock = &req->checkpoint_vclock,
+		.checkpoint_lsn = &req->checkpoint_lsn,
 	};
+	/*
+	 * Vclock must be cleared, as it sets -1 signature, which cannot be
+	 * done by memset above. This is done in order to distinguish not
+	 * initialized vclock from the zero one.
+	 */
+	vclock_clear(&req->checkpoint_vclock);
```
sergepetrenko, line 2448: Nice solution!

Just as I was going to ask how do you distinguish these two cases.

sergepetrenko (commented):
Thanks for the patch!

Looks well-polished, good job!
I've left some minor comments.

src/box/memtx_engine.cc
```diff
@@ -1264,6 +1265,37 @@ struct memtx_join_ctx {
...
+send_join_meta(struct xstream *stream, const struct raft_request *raft_req,
+	       const struct synchro_request *synchro_req)
+{
+	struct xrow_header row;
+	/* Mark the beginning of the metadata stream. */
+	xrow_encode_type(&row, IPROTO_JOIN_META);
+	xstream_write(stream, &row);
+
+	/* Encoding raft request uses fiber()->gc region. */
+	RegionGuard region_guard(&fiber()->gc);
```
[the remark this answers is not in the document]
Serpentian: Done

src/box/xrow.c
Serpentian: I use it in subsequent commits in order to decode vclock inside `CURSOR`. There's also a comment about this in commit message

src/box/applier.cc
Serpentian: Fixed

src/box/memtx_engine.cc
```diff
@@ -1093,9 +1118,14 @@ checkpoint_f(va_list ap)
...
-		goto fail;
-	if (checkpoint_write_synchro(snap, &ckpt->synchro_state) != 0)
+	/*
+	 * There may be no user data (e.g. when only RAFT_PROMOTE is written),
+	 * write limbo and raft states right after system spaces, at the end
+	 * of the checkpoint.
+	 */
+	if (!is_synchro_written &&
+	    checkpoint_write_system_data(snap, &ckpt->raft,
+					 &ckpt->synchro_state) != 0)
```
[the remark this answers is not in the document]
Serpentian: That's a difficult question. I opened a PR https://github.com/tarantool/tarantool/pull/10251 for that case and propose to discuss it thoroughly there, let's not block this PR due to that, it seems as a follow up. We need a proper review iteration from Vladimir and Vladislav there and we need to discuss, do we want it at all

---

I agree, that we should probably consider such rows system ones and don't allow to force recover, when they're broken. Moreover, we should not recover, when they're missing. However, this breaks recovery from old versions of snaps a lot: e.g. 1.10 has no raft/limbo states in snap, 3.X won't be able to recover from such snaps and even force_recovery won't help, if we decide to consider them system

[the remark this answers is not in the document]
sergepetrenko: Thanks!

src/box/xrow.c
sergepetrenko: Yes, indeed. I missed that.

sergepetrenko (approved):

Issue #9563, opened 2023-12-27
net.box: recurrent connection liveness probe


Totktonada:
`net.box` supports publish-subscribe interaction pattern since 2.10.0 (PR #6510).

`net.box` also supports the `on_disconnect` trigger ([docs](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#net-box-on-disconnect)), so we can perform some actions if the connection is closed (either by server or by the client itself). For example, we may want to reconnect to another instance and subscribe to its events if the current one goes down.

The problem is that a TCP connection doesn't always advertize its death (articles on the topic: [1], [2]). A client need to do recurrent liveness probes to know an actual state of the connection.

Linux has `SO_KEEPALIVE` and `TCP_USER_TIMEOUT` (with a lot of nuances how to properly use them, especially in presence of outcoming data). OTOH, we can implement platform independent liveness probes by sending `IPROTO_PING`.

I vote for a platform-independent solution if it is not a way harder, because it allows to perform testing of a behavior of a module/application in the connection loss situation disregarding of a platform. (And I'm a bit scary about Linux API for the keepalives.)

[1]: https://holmeshe.me/network-essentials-setsockopt-SO_KEEPALIVE/
[2]: https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/

PR #9977, opened 2024-04-25
Introduce net.cluster

alyapunov:
It's a module that provides connection to all replicasert of the cluster. It maintains a list of connections (that actually are net.replicaset connections) and allows to subscribe to an event (aka watch) on entire cluster.

See module header for more detailed description.


src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+    return default_cluster:get_connections(opts)
+end
+
+-- watch_leaders for default cluster (which is created if needed).
+local function watch_leaders(key, func, opts)
+    if not default_cluster then
+        default_cluster = cluster_new(cluster_cfg_defaults)
+    end
+    return default_cluster:watch_leaders(key, func, opts)
+end
```
Totktonada, line 449: I have an idea (inspired by earlier discussions with @Mons) to make the default instace and a user created instace APIs the same.

I mean, if there are the following objects:

```lua
local net_cluster = require('net.cluster')
local my_net_cluster = require('net.cluster').new()
```

The usage is the following now.

```lua
net_cluster.watch_leaders(<...>) -- self is NOT passed
my_net_cluster:watch_leaders(<...>)
```

But I would like the following one.

```lua
net_cluster:watch_leaders(<...>) -- self is passed
my_net_cluster:watch_leaders(<...>)
```

I see two pros:

* The same API: we can implement a function that accepts any of these objects and just call the methods.
* The default instance has methods that explicitly accept a state in the `self` argument instead of implicit passing it via captured upvalues (that feels hacky).

A possible con is that it is inconsistent with some of existing modules: for example, `json`. OTOH, I use the proposed approach in new modules, where appropriate: for example, `config`.

I don't insist here in any way, but if you like the idea too, I see no reasons to discard it.

src/box/lua/init.c
```diff
@@ -227,6 +228,7 @@ static const char *lua_sources[] = {
 	"box/upgrade", NULL, upgrade_lua,
 	"box/net_box", "net.box", net_box_lua,
 	"box/net_replicaset", "internal.net.replicaset", net_replicaset_lua,
+	"box/net_cluster", "experimental.net.cluster", net_cluster_lua,
```
Totktonada, line 231: I think that a documentation request is desirable for experimental modules. This way engages users into testing and leaving a feedback.

I also think that an item in the changelog/release notes is good too as a little advertisement.

src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+        default_cluster = cluster_new(cluster_cfg_defaults)
+    end
+    return default_cluster:reload_config(config)
+end
+
+return {
+    new = cluster_new,
+
+    get = get,
+    get_connections = get_connections,
```
Totktonada, line 463: I'm in a doubt about these methods/functions.

First, I'm not sure about names. `get` looks like something to be frequently used, because of the short/common word name. `get` and `get_connections` feels non-uniform: just 'get' and 'get something' for a same-level methods.

Huh, choosing names is the hardest task in programming :)

Second, I'm not sure about exposing `net.replicaset` objects from the a-kind-of-public API, while the `net.replicaset` module is internal.

Also, if we have a more general connections-by-filters method, should we also provide a particular connections-by-replicaset-name method? (I assume that a filter by a replicaset name may be included into the list of the filters.)

If it worth such a shortcut, how about an approach that we use in the [testing helper](https://github.com/tarantool/tarantool/blob/2467e421cbfd42d0f58c0d27e4e397e730412e6c/test/config-luatest/cluster.lua#L18) for working with a cluster: get a replicaset/instance representation (connection) using the indexing operator? Sure, it technically forbids using of some instance/replicaset names, but it doesn't look as a problem in practice.

These are a bit too picky comments considering the experimental status of the module. I left them to share thoughts and I don't want them to be blockers.

We discussed it with @alyapunov and he said that he can replace these two methods with one `:call_leader(replicaset_name, <...>)` (or something of this kind). It seems it would solve my concerns.

src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+        -- No desired roles.
+        return false
+    end
+    for _, role in pairs(roles) do
+        if replicaset.roles[role] then
+            return true
+        end
+    end
+    return false
+end
```
Totktonada, line 61: In `experimantal.connpool` the `roles` [filter](https://github.com/tarantool/tarantool/blob/2467e421cbfd42d0f58c0d27e4e397e730412e6c/src/box/lua/experimental/connpool.lua#L132-L150) verifies that an instance has all the provided roles (AND semantic). Here we have the opposite: a replicaset should have any of the provided roles (OR semantic).

I don't know for sure what is better, but I would like to make the same named filter working consistently.

Personally, I think that the AND semantics is more intuitive here: look up an instance/a replicaset that match all the given criteria.

src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+            own_group_name = info.group_name
+            break
+        end
+    end
+
+    -- Find replicasets of own group.
+    for _, info in pairs(config:instances()) do
+        if info.group_name ~= own_group_name then
+            goto continue
+        end
```
Totktonada, line 184: I see no reason, why we should work only within our own group. Also, as I see on a brief glance, this constraint is not mentioned in comments (which hopefully will be transformed later into the documentation).

src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+        local rs = replicasets[info.replicaset_name] or {roles = {}}
+        for _, role in ipairs(config:get('roles',
+                {instance = info.instance_name})) do
+            rs.roles[role] = role
+        end
+        replicasets[info.replicaset_name] = rs
```
Totktonada, line 190: Here we assume that 'replicaset roles' are the roles that are enabled on **at least one** instance in the given replicaset (OR semantic). I don't know, what is better: AND or OR semantic. However, I think that this nuance should mentioned in the documentation.

src/box/lua/net_cluster.lua
```diff
@@ -0,0 +1,466 @@
...
+    end
+    return default_cluster:reload_config(config)
+end
+
+return {
+    new = cluster_new,
+
+    get = get,
+    get_connections = get_connections,
+    watch_leaders = watch_leaders,
```
Totktonada, line 464: In my understanding, the publish-subscribe pattern requires the ability to set handlers for the following situations:

* on connect
* on change
* on disconnect

I'm thinking on it in terms of a local represention/a local caching of a remote state. This local cache has the `not yet filled` state, has the `actual` state and has the `stalled/expired` state. We need the 'on disconnect' handler to implement the `stalled/expired` state.

The 'on connect' handler is coalesced with 'on change' in `net.box`, it is OK for my taste.

We can call 'on change' periodically even if the remote state is not changed and don't provide 'on disconnect'. However, `net.box` provides the `on_disconnect` trigger separately (and don't call 'on change' if there were no changes). Both ways are OK in my understanding.

Note re `net.box`: it may be used to actually track actuality of the local state after #9563.

Totktonada (commented):
I looked over the API: please, consider my comments above.

I don't look deeper into the implementation yet.

PR #9980, opened 2024-04-27
box: pass source tuple to `default_func` field option of `space:format`

### LINKED ISSUE #9825: Feature `space:format` `default_func` need tuple argument
A simple case is to fill in a default field at the time of recording, but the value of the field depends on other fields of the record:

```lua
box.schema.func.create('default_fingerprint', {
    language = 'Lua',
    body = 'function(_, tuple) return digest.md5(tuple[1] .. tuple[2]) end'
})

box.space.test:format({
    {name = 'id', type = 'string'},
    {name = 'value', type = 'string'},
    {name = 'fingerprint', type = 'string', default_func = 'default_fingerprint'}
})
```

**Definition of Done**

An additional argument is passed to the function that generates the default value for the field - the recordable tuple.

CuriousGeorgiy:
This patch adds a source tuple as a second argument to the `default_func` field option to `space:format`.

Closes #9825

### DIFF
--- a/changelogs/unreleased/gh-9825-space-format-default-func-tuple-arg.md
+++ b/changelogs/unreleased/gh-9825-space-format-default-func-tuple-arg.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* Added the source tuple as a second argument for the functional variant of the
+  default field values (gh-9825).
--- a/src/box/tuple_format.c
+++ b/src/box/tuple_format.c
@@ -34,6 +34,7 @@
 #include "json/json.h"
 #include "coll_id_cache.h"
 #include "trivia/util.h"
+#include "tuple.h"
 #include "tuple_builder.h"
 #include "tuple_constraint.h"
 #include "field_default_func.h"
@@ -1556,6 +1557,11 @@ tuple_format_apply_defaults(struct tuple_format *format, const char **data,
 	tuple_builder_new(&builder, &fiber()->gc);
 	size_t region_svp = region_used(&fiber()->gc);
 	bool is_tuple_changed = false;
+	struct tuple *source = NULL;
+	if (tuple_format_has_default_funcs(format)) {
+		source = tuple_new(tuple_format_runtime, *data, *data_end);
+		tuple_ref(source);
+	}
 	/*
 	 * Process fields that are present in both the format and the tuple.
 	 * Break prematurely when all defaults are applied.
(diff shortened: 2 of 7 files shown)


src/box/tuple_format.c
```diff
@@ -1556,6 +1557,9 @@ tuple_format_apply_defaults(struct tuple_format *format, const char **data,
 	tuple_builder_new(&builder, &fiber()->gc);
 	size_t region_svp = region_used(&fiber()->gc);
 	bool is_tuple_changed = false;
+	struct tuple *source =
+		tuple_new(tuple_format_runtime, *data, *data_end);
```
Gumix, line 1561: I'd consider the following optimization — build a tuple only if there is at least 1 field with default func.

changelogs/unreleased/gh-9825-space-format-default-func-tuple-arg.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Added the source tuple as a second argument to the `default_func` field
+  option of `space:format` (gh-9825).
```
Gumix, line 4: Sounds confusing (Added the source tuple to the field option of space:format), I'd rewrite it a bit more detailed.

src/box/tuple_format.c
CuriousGeorgiy: Fixed.

changelogs/unreleased/gh-9825-space-format-default-func-tuple-arg.md
CuriousGeorgiy: Done.

Gumix (approved):

andreyaksenov (approved):

src/box/tuple_format.c
```diff
@@ -1556,6 +1557,11 @@ tuple_format_apply_defaults(struct tuple_format *format, const char **data,
 	tuple_builder_new(&builder, &fiber()->gc);
 	size_t region_svp = region_used(&fiber()->gc);
 	bool is_tuple_changed = false;
+	struct tuple *source = NULL;
+	if (tuple_format_has_default_funcs(format)) {
+		source = tuple_new(tuple_format_runtime, *data, *data_end);
```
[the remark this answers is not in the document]
CuriousGeorgiy: I guess using a MsgPack object here would be better for performance, but AFAIU, we simply have a requirement that the user has to receive a tuple object (I will update the commit message if this is correct before we resolve this comment).

Am I right @Gumix?

Issue #9997, opened 2024-05-08
Support splice update operation for varbinary fields


alyapunov:
It would be nice to have that:
```lua
tarantool> varbinary = require('varbinary')
---
...

tarantool> s:replace{1, varbinary.new('abc')}
---
- [1, !!binary YWJj]
...

tarantool> s:update({1}, {{':', 2, 2, 1, varbinary.new('d')}})
---
- [1, !!binary YWRj]
...

tarantool> tostring(s:get{1}[2])
---
- adc
...
```

PR #9998, opened 2024-05-08
box: support varbinary in splice update operation

alyapunov:
Just in case splice is an update operation that can modify string
field, deleting a part of it and inserting something instead.
```
tarantool> box.tuple.new{'1234567'}:update{{':', 1, 2, 3, '!'}}
---
- ['1!567']
...
```
In the this example splice operation ':' takes field 1 of tuple,
takes position 2 in string, removes 3 symbols and inserts '!' to
that position.

Both deletion and insertion can naturally degrade to no-op.

Before this patch a splice operation required both field and
inserting argument to be strings.

This patch allows the field and the argument to be varbinary.
```
tarantool> varb = require('varbinary')
tarantool> t = box.tuple.new{varb.new('1234567')}
tarantool> t:update{{':', 1, 2, 3, varb.new('!')}}
---
- [!!binary MSE1Njc=]
...

```

That also allows to insert strings into varbinary fields and
insert varbinary data into string. The actual field type after
update such operation remains the same, so updated string will
be string, while updated varbinary will be varbinary.

Closes https://github.com/tarantool/tarantool/issues/9997

### DIFF
--- a/changelogs/unreleased/gh-9997-support-splice-for-varbinary.md
+++ b/changelogs/unreleased/gh-9997-support-splice-for-varbinary.md
@@ -1,0 +1,4 @@
+## feature/core
+
+* The splice operation (':') in update/upsert is now allowed for varbinary fields
+  and can take a varbinary argument for insertion (gh-9997).
--- a/src/box/xrow_update_field.c
+++ b/src/box/xrow_update_field.c
@@ -237,14 +237,25 @@ xrow_update_mp_read_scalar(const char **expr,
 		ret->type = XUPDATE_TYPE_FLOAT;
 		ret->flt = mp_decode_float(expr);
 		break;
-	case MP_STR: {
+	case MP_STR:
+	case MP_BIN:
+	{
 		ret->type = XUPDATE_TYPE_STR;
 		struct xrow_update_string *str = &ret->str;
-		str->array[0].data = mp_decode_str(expr, &str->array[0].length);
-		memset(&str->array[1], 0, sizeof(str->array[1]) * 2);
-		ret->str.type = XROW_UPDATE_STRING_SINGLE;
-		break;
+		if (mp_typeof(**expr) == MP_STR) {
+			str->array[0].data =
+				mp_decode_str(expr, &str->array[0].length);
+			str->store_type = XUPDATE_STORE_TYPE_STR;
+		} else {
+			assert(mp_typeof(**expr) == MP_BIN);
+			str->array[0].data =
+				mp_decode_bin(expr, &str->array[0].length);
+			str->store_type = XUPDATE_STORE_TYPE_BIN;
 		}
+		memset(&str->array[1], 0, sizeof(str->array[1]) * 2);
+		str->type = XROW_UPDATE_STRING_SINGLE;
+		break;
+	}
 	case MP_EXT:
 		len = mp_decode_extl(expr, &ext_type);
 		if (ext_type == MP_DECIMAL) {
@@ -258,6 +269,10 @@ xrow_update_mp_read_scalar(const char **expr,
 	}
 }
 
+/**
+ * Decode string or varbinary argument of update operation.
+ * Set error and return -1 if it's not a string or varbinary.
+ */
 static inline int
 xrow_update_mp_read_str(struct xrow_update_op *op, const char **expr,
 			uint32_t *len, const char **ret)
@@ -265,8 +280,11 @@ xrow_update_mp_read_str(struct xrow_update_op *op, const char **expr,
 	if (mp_typeof(**expr) == MP_STR) {
 		*ret = mp_decode_str(expr, len);
 		return 0;
+	} else if (mp_typeof(**expr) == MP_BIN) {
+		*ret = mp_decode_bin(expr, len);
+		return 0;
 	}
-	return xrow_update_err_arg_type(op, "a string");
+	return xrow_update_err_arg_type(op, "a string or varbinary");
 }
 
 /* }}} read_arg helpers. */
@@ -423,7 +441,16 @@ xrow_update_scalar_sizeof(const struct xrow_update_scalar *scalar)
 	case XUPDATE_TYPE_DECIMAL:
 		return mp_sizeof_decimal(&scalar->dec);
 	case XUPDATE_TYPE_STR:
-		return mp_sizeof_str(xrow_update_string_size(&scalar->str));
+	{
+		int data_length = xrow_update_string_size(&scalar->str);
+		if (scalar->str.store_type == XUPDATE_STORE_TYPE_STR) {
+			return mp_sizeof_str(data_length);
+		} else {
+			assert(scalar->str.store_type ==
+			       XUPDATE_STORE_TYPE_BIN);
+			return mp_sizeof_bin(data_length);
+		}
+	}
 	case XUPDATE_TYPE_NONE:
 	default:
 		unreachable();
@@ -568,7 +595,7 @@ xrow_update_op_do_splice(struct xrow_update_op *op,
 	struct xrow_update_arg_splice *arg = &op->arg.splice;
 
 	if (scalar->type != XUPDATE_TYPE_STR)
-		return xrow_update_err_arg_type(op, "a string");
+		return xrow_update_err_arg_type(op, "a string or varbinary");
 
 	struct xrow_update_string *str = &scalar->str;
 	int32_t str_len = xrow_update_string_size(str);
@@ -620,12 +647,12 @@ xrow_update_op_do_splice(struct xrow_update_op *op,
 /* {{{ store_op */
 
 /**
- * Store string rope representation as MsgPack.
+ * Store string or varbinary rope representation as MsgPack.
  *
  * @param str A string.
  * @param out A pointer to output buffer.
  *
- * @return Output buffer position after encoded string.
+ * @return Output buffer position after encoded string (or varbinary).
  */
 static char *
 xrow_update_store_string(const struct xrow_update_string *str, char *out);
@@ -695,7 +722,12 @@ xrow_update_string_store_leaf(const char *data, size_t size, void *cb_arg)
 static char *
 xrow_update_store_string(const struct xrow_update_string *str, char *out)
 {
-	out = mp_encode_strl(out, xrow_update_string_size(str));
+	if (str->store_type == XUPDATE_STORE_TYPE_STR) {
+		out = mp_encode_strl(out, xrow_update_string_size(str));
+	} else {
+		assert(str->store_type == XUPDATE_STORE_TYPE_BIN);
+		out = mp_encode_binl(out, xrow_update_string_size(str));
+	}
 	if (str->type < XROW_UPDATE_STRING_ROPE) {
 		for (size_t i = 0; i < lengthof(str->array); i++) {
 			memcpy(out, str->array[i].data, str->array[i].length);
--- a/test/box-luatest/gh_10032_update_upsert_splice_test.lua
+++ b/test/box-luatest/gh_10032_update_upsert_splice_test.lua
@@ -1,0 +1,79 @@
+local t = require('luatest')
+local server = require('luatest.server')
+local helper = require('test.box-luatest.update_upsert_splice_helper')
+
+-- Type data stored in tuple: string or varbinary.
+local types = {{type = 'str'}, {type = 'bin'}}
+local g = t.group('gh-10032-update-upsert-splice', types)
+
+g.before_all(function(cg)
+    cg.server = server:new()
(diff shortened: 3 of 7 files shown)


src/box/xrow_update_field.c
```diff
@@ -244,7 +245,15 @@ xrow_update_mp_read_scalar(const char **expr,
...
-		}
+	}
+	case MP_BIN: {
+		ret->type = XUPDATE_TYPE_BIN;
+		struct xrow_update_string *str = &ret->str;
+		str->array[0].data = mp_decode_bin(expr, &str->array[0].length);
+		memset(&str->array[1], 0, sizeof(str->array[1]) * 2);
+		ret->str.type = XROW_UPDATE_STRING_SINGLE;
+		break;
+	}
```
nshy, line 256: I guess we can reduce code duplication. For example here we can introduce `xrow_update_mp_read_str` which will call `mp_decode_bin` or ` mp_decode_str` depending on type.

Similarly I would not indroduce `xrow_update_mp_read_bin` and `xrow_update_store_varbinary`. They are too similar to string ones. Basically the only difference is in encoding/decoding length.

test/box-luatest/gh_9997_update_splice_test.lua
```diff
@@ -0,0 +1,273 @@
...
+            local res = {ok = ok, ret = ret}
+            local msg = get_test_case_message(test_case, res)
+            t.assert_equals(res, test_case.res, msg)
+            if ok then
+                t.assert_equals(field_type(ret[2]), data_type, msg)
+            end
+        end
+    end, {cg.params.type, tmpl, test_cases,
+          get_test_case_message_source, field_type_source})
+end
```
nshy, line 273: My overall impression is the test is too complicated to test the introduced changes. In includes fuzzy testing, helpers to improve report on failed test case as cases are run from a list, generating reference result instead of simply set it by hand. 

I guess making a fuzzy test for update/upsert is different task. Here we'd better make a simple functional test. Just a bunch of assertions is the best)

nshy (changes requested):

test/box-luatest/gh_10032_update_upsert_splice_test.lua
```diff
@@ -0,0 +1,274 @@
...
+        local s = box.schema.space.create('test', {engine = 'memtx'})
+        s:create_index('pk')
+
+        for _, test_case in ipairs(test_cases) do
+            local ops = test_case.ops
+            s:replace(tmpl)
+            local ok, ret = pcall(s.update, s, tmpl[1], ops)
+            ret = ok and ret or tostring(ret)
+            local res = {ok = ok, ret = ret}
+            local msg = get_test_case_message(test_case, res)
```
nshy, line 237: I guess it is easier to consider failure and success cases explicitly:

```lua
    local ok, ret = pcall(<>)
    local ops = test_case.ops
    t.assert_equals(test_case.ok, ok, ops)
    if ok then
        t.assert_equals(ret, test_case.ret, ops)
        t.assert_equals(field_type, data_type, ops)
    else
        t.assert_equals(tostring(ret), test_case.ret, ops)
    end
```

nshy (changes requested):

src/box/xrow_update_field.c
```diff
@@ -693,9 +728,13 @@ xrow_update_string_store_leaf(const char *data, size_t size, void *cb_arg)
 }
 
 static char *
-xrow_update_store_string(const struct xrow_update_string *str, char *out)
+xrow_update_store_strbin(const struct xrow_update_string *str, char *out,
+			 bool is_string)
```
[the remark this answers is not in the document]
alyapunov: So you propose to add something like `enum xrow_update_scalar_type type` to `struct xrow_update_string`, which is only used as member in `struct xrow_update_scalar`, which in turn already has `enum xrow_update_scalar_type type`. It seems to violate DRY principle, and I don't like it.
Perhaps we can create one scalar type `XUPDATE_TYPE_STRBIN` instead of `XUPDATE_TYPE_STR/XUPDATE_TYPE_BIN` pair, introduce new enum with `SUBTYPE_STR/SUBTYPE_BIN` values then rename `xrow_update_string` -> `xrow_update_strbin` and store the latter enum in it.

test/box-luatest/gh_9997_update_splice_test.lua
```diff
@@ -0,0 +1,35 @@
...
+    tuple = box.tuple.new{varbinary.new('1234567890')}
+    tuple = tuple:update{{':', 1, 4, 2, varbinary.new('abc')}}
+    t.assert(varbinary.is(tuple[1]))
+    t.assert_equals(tuple[1], '123abc67890')
+
+    -- Other types are still forbidden.
+    local mess = "Argument type in operation ':' on field 1 does not match " ..
+                 "field type: expected a string or varbinary"
+    tuple = box.tuple.new{1234567890}
+    t.assert_error_msg_content_equals(mess, tuple.update, tuple,
```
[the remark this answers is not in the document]
alyapunov: ok

test/box-luatest/gh_10032_update_upsert_splice_test.lua
```diff
@@ -0,0 +1,274 @@
...
+        return res
+    end
+    return {
+        ops = test_case.ops,
+        expect_res = prettify_case_res(test_case.res),
+        got_res = prettify_case_res(got_res),
+    }
+end
+]]
+local get_test_case_message = loadstring(get_test_case_message_source)()
```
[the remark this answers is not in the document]
alyapunov: If without this argument luatest prints error like that:
```
[001] not ok 1	gh-10032-update-upsert-splice.type:"bin".test_tuple_update_splice
[001] #   .../test/box-luatest/gh_10032_update_upsert_splice_test.lua:214: expected: {ok = true, ret = {100, bbbbmmmm2ffff34eeee!, 200}}
[001] #   actual: {ok = true, ret = [100, !!binary YmJiYm1tbW0yZmZmZjM0ZWVlZQ==, 200]}
[001] #   stack traceback:
[001] #   	.../test/box-luatest/gh_10032_update_upsert_splice_test.lua:214: in function 'gh-10032-update-upsert-splice.type:"bin".test_tuple_update_splice'
[001] #   	...
```
I think it's almost impossible to understand what's going wrong without test modification.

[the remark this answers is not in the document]
alyapunov: Oh, if we pass test_case.ops, perhaps it would be enough.

test/box-luatest/gh_10032_update_upsert_splice_test.lua
alyapunov: I don't see much difference, but ok.

test/box-luatest/gh_10032_update_upsert_splice_test.lua
```diff
@@ -0,0 +1,274 @@
...
+        if ok then
+            t.assert_equals(field_type(ret[2]), cg.params.type, msg)
+        end
+    end
+end
+
+-- Test lots of various variants of splice (':') of memtx space update.
+g.test_memtx_update_splice = function(cg)
+    local tmpl = generate_tuple_template(cg.params.type)
+    local test_cases = generate_test_cases(tmpl)
```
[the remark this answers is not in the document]
alyapunov: done.

test/box-luatest/gh_10032_update_upsert_splice_test.lua
```diff
@@ -0,0 +1,274 @@
...
+local field_type_source = [[
+return function(field)
+    local varbinary = require('varbinary')
+    if type(field) == 'string' then
+        return 'str'
+    elseif varbinary.is(field) then
+        return 'bin'
+    end
+end
+]]
```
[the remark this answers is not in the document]
alyapunov: Good idea, thanks.

changelogs/unreleased/gh-9997-support-splice-for-varbinary.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* Now splice operation (':') in update/upsert is allowed for varbinary fields
+  and can take varbinary argument for insertion (gh-9997).
```
xuniq, line 4:
```suggestion
* The splice operation (':') in update/upsert is now allowed for varbinary fields
  and can take a varbinary argument for insertion (gh-9997).
```

xuniq (approved):

src/box/xrow_update_field.c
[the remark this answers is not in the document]
nshy: I for the second variant too. May be we can start from just boolean flag instead of `XUPDATE_TYPE_STRBIN` or as you like.

test/box-luatest/gh_10032_update_upsert_splice_test.lua
```diff
@@ -0,0 +1,78 @@
+local t = require('luatest')
+local server = require('luatest.server')
+-- Type data stored in tuple: string or varbinary.
+local types = {{type = 'str'}, {type = 'bin'}}
```
nshy, line 4: Nit. Let's not mix imports and other variables.

nshy (changes requested):

src/box/xrow_update_field.c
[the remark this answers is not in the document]
alyapunov: I've already done it in this way, please review.

src/box/xrow_update_field.c
```diff
@@ -237,14 +237,25 @@ xrow_update_mp_read_scalar(const char **expr,
...
+			str->store_type = XUPDATE_STORE_TYPE_STR;
+		} else {
+			assert(mp_typeof(**expr) == MP_BIN);
+			str->array[0].data =
+				mp_decode_bin(expr, &str->array[0].length);
+			str->store_type = XUPDATE_STORE_TYPE_BIN;
+		}
 		memset(&str->array[1], 0, sizeof(str->array[1]) * 2);
-		ret->str.type = XROW_UPDATE_STRING_SINGLE;
+		str->type = XROW_UPDATE_STRING_SINGLE;
```
nshy, line 256: I guess it is worth rename `xrow_update_string_type` also.

src/box/xrow_update_field.c
```diff
@@ -620,15 +647,16 @@ xrow_update_op_do_splice(struct xrow_update_op *op,
 /* {{{ store_op */
 
 /**
- * Store string rope representation as MsgPack.
+ * Store string or varbinary rope representation as MsgPack.
  *
  * @param str A string.
  * @param out A pointer to output buffer.
+ * @param is_string True - encode as string, false - encode as varbinary.
```
nshy, line 654: Outdated change.

src/box/xrow_update_field.c
alyapunov: ok, done.

src/box/xrow_update_field.c
alyapunov: Thanks, removed.

test/box-luatest/gh_10032_update_upsert_splice_test.lua
alyapunov: Good practice, fixed.

src/box/xrow_update_field.c
nshy: Let's then rename `enum xrow_update_strbin_type` values too.

I don't know, if we consistently update all names then we need to rename so many constants, struct, comments and variables. `xrow_update_string_chunk`, `xrow_update_string_rope`, change string to `string (or varbinary)` in comments, rename `str` vars to `strbin` etc etc.

Anyway I guess either we should do it or keep string name everywhere. What do you think?

alyapunov: Actually I think that rename of `xrow_update_string_type` was excess. Let's rename it back?
The reason is that we actually load MP_BIN value to string, aka sequence of chars. From that point we work with thing as with string. That thing can consist of one or several sustrings etc. So we should name struct type and the corresponding member of scalar as *strbin to be sure we will not forget to store is as MP_BIN in the end. The all other names may remain string-like. As for local 'str' names, that does not matter much. It could be 's' or whatereber.

alyapunov: Discusses offline. Decided not to use 'strbin' at all.

nshy (approved):

CuriousGeorgiy (approved):
Looks great that you managed to reduce that patch to just changing the input and output points of the slice operation!

Regarding the slice operation testing: would be great to have fuzzing tests for `xrow_update_execute` or some lower level `xrow_update_*` functions just like we already have `test/fuzz/xrow_decode_*` tests (out of the scope of this patch series).

PR #10001, opened 2024-05-09
console: create per-session variables scope

### LINKED ISSUE #9985: console: don't spoil globals
When I write something like the following in the admin console...

```lua
localhost:3301> x = 1
```

...this `x` variable is written to globals. It is accessible from all the loaded modules and other console sessions.

Moreover, if I do something like...

```lua
localhost:3301> function foo() return 'something' end
```

...this function is also accessible from all the Lua code, all the other console sessions and it may be called via iproto (if permitted by configured privileges).

It is counter-intuitive and may even lead to unexpected name clashes.

I propose to assign variables to a scope bound to the console session and allow explicit assignment of globals using `_G.x = 1`.

Let's enable the new behavior under the new `compat` option: `console_session_scope_vars`. This way ensure that it doesn't break any scripts written on top of `tarantoolctl connect` / `tt connect`.

Totktonada:
It is counter-intuitive that all the non-local assignments in the console affect globals and may interfere with application's logic.

It is also counter-intuitive that the non-local assignments are shared between different console sessions.

Now, each console session has its own variables scope and the non-local assignments use it instead of globals.

Let's consider examples of the new behavior.

Example 1. A console session has a variable scope that is separate from globals.

```lua
console_1> _G.x = 1
console_1> x = 2

console_1> _G.x
---
- 1
...
console_1> x
---
- 2
...
```

Note: A global variable is still accessible using `_G` even if the same named session scope variable exists.

Example 2. A global variable is read if there is no session local variable.

```lua
console_1> _G.x = 1
console_1> x
---
- 1
...
```

Example 3. Different console sessions have separate variable scopes.

```lua
console_1> x = 1
console_2> x = 2

console_1> x
---
- 1
...
console_2> x
---
- 2
...
```

The new behavior is enabled using the `console_session_scope_vars` compat option. The option is `old` by default in Tarantool 3.X, `new` by default in 4.X. The `old` behavior is to be removed in 5.X.

Fixes #9985

### DIFF
--- a/changelogs/unreleased/console-session-scope-vars.md
+++ b/changelogs/unreleased/console-session-scope-vars.md
@@ -1,0 +1,5 @@
+## feature/console
+
+* Interactive console now performs non-local assignments to a per-session
+  variable scope if the `console_session_scope_vars` compat option is set to
+  `new` (gh-9985).
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2414,6 +2414,12 @@ return schema.new('instance_config', schema.record({
         }, {
             default = 'old',
         }),
+        console_session_scope_vars = schema.enum({
+            'old',
+            'new',
+        }, {
+            default = 'old',
+        }),
     }),
     -- Instance labels.
     labels = schema.map({
--- a/src/box/lua/console.lua
+++ b/src/box/lua/console.lua
@@ -19,6 +19,7 @@ ffi.cdef[[
 
 local internal = require('console.lib')
 local session_internal = box.internal.session
+local compat = require('compat')
 local fiber = require('fiber')
 local socket = require('socket')
 local log = require('log')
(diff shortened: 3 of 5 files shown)


changelogs/unreleased/console-session-scope-vars.md
```diff
@@ -0,0 +1,5 @@
+## feature/console
+
+* Interactive console now performs non-local assignments to a per-session
+  variable scope if `console_session_scope_vars` compat option is `new`
+  (gh-9985).
```
xuniq, line 5:
```suggestion
* Interactive console now performs non-local assignments to a per-session
  variable scope if the `console_session_scope_vars` compat option is set to `new`
  (gh-9985).
```

xuniq (approved):

changelogs/unreleased/console-session-scope-vars.md
Totktonada: Thanks! Fixed.

Buristan (approved):
Hi, Alexander!
Thanks for the patch!
LGTM!

PR #10003, opened 2024-05-11
box: prevent demoted leader from being a candidate in the next elections

### LINKED ISSUE #9855: box.ctl.demote() does not guarantee demoting
`box.ctl.demote()` does not guarantee that the node will not become a leader in the new term.

For example, we have the cluster with 3 nodes and election failover.
```sh
$ tt status app
INSTANCE           STATUS      PID
app:election-1     RUNNING     3867795
app:election-2     RUNNING     3867796
app:election-3     RUNNING     3868129

$ echo "box.info.election" | tt connect app:election-2
---
- leader_idle: 0
  leader_name: election-2
  state: leader
  vote: 2
  term: 80
  leader: 2
...

$ echo "box.ctl.demote()" | tt connect app:election-2
---
...

$ echo "box.info.election" | tt connect app:election-2
---
- leader_idle: 0
  leader_name: election-2
  state: leader
  vote: 2
  term: 81
  leader: 2
...
```

CuriousGeorgiy:
This patch prevents the demoted leader from being a candidate in the next elections.

Closes #9855

### DIFF
--- a/changelogs/unreleased/gh-9855-box-ctl-demote-guarantee-demoting.md
+++ b/changelogs/unreleased/gh-9855-box-ctl-demote-guarantee-demoting.md
@@ -1,0 +1,5 @@
+## bugfix/replication
+
+* Fixed a bug that allowed the old leader in
+  `box.cfg{election_mode = 'candidate'` mode to get re-elected after resigning
+  himself through `box.ctl.demote` (gh-9855).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3080,24 +3080,7 @@ box_demote(void)
 		return 0;
 
 	const struct raft *raft = box_raft();
-	if (box_election_mode == ELECTION_MODE_OFF) {
-		assert(raft->state == RAFT_STATE_FOLLOWER);
-		if (raft->leader != REPLICA_ID_NIL) {
-			diag_set(ClientError, ER_NOT_LEADER, raft->leader);
-			return -1;
-		}
-		if (txn_limbo.owner_id == REPLICA_ID_NIL)
-			return 0;
-		/*
-		 * If the limbo term is up to date with Raft, then it might have
-		 * a valid owner right now. Demotion would disrupt it. In this
-		 * case the user has to explicitly overthrow the old owner with
-		 * local promote(), or call demote() on the actual owner.
-		 */
-		if (txn_limbo.promote_greatest_term == raft->term &&
-		    txn_limbo.owner_id != instance_id)
-			return 0;
-	} else {
+	if (box_election_mode != ELECTION_MODE_OFF) {
 		if (txn_limbo_replica_term(&txn_limbo, instance_id) !=
 		    raft->term)
 			return 0;
@@ -3105,11 +3088,27 @@ box_demote(void)
 			return 0;
 		if (raft->state != RAFT_STATE_LEADER)
 			return 0;
+		return box_trigger_elections();
 	}
+
+	assert(raft->state == RAFT_STATE_FOLLOWER);
+	if (raft->leader != REPLICA_ID_NIL) {
+		diag_set(ClientError, ER_NOT_LEADER, raft->leader);
+		return -1;
+	}
+	if (txn_limbo.owner_id == REPLICA_ID_NIL)
+		return 0;
+	/*
+	 * If the limbo term is up to date with Raft, then it might have
+	 * a valid owner right now. Demotion would disrupt it. In this
+	 * case the user has to explicitly overthrow the old owner with
+	 * local promote(), or call demote() on the actual owner.
+	 */
+	if (txn_limbo.promote_greatest_term == raft->term &&
+	    txn_limbo.owner_id != instance_id)
+		return 0;
 	if (box_trigger_elections() != 0)
 		return -1;
-	if (box_election_mode != ELECTION_MODE_OFF)
-		return 0;
 	if (box_try_wait_confirm(2 * replication_synchro_timeout) < 0)
 		return -1;
 	int64_t wait_lsn = box_wait_limbo_acked(replication_synchro_timeout);
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3086,9 +3086,8 @@ box_demote(void)
 			return 0;
 		if (txn_limbo.owner_id != instance_id)
 			return 0;
-		if (raft->state != RAFT_STATE_LEADER)
-			return 0;
-		return box_trigger_elections();
+		box_raft_leader_step_off();
+		return 0;
 	}
 
 	assert(raft->state == RAFT_STATE_FOLLOWER);
--- a/test/replication-luatest/gh_9855_box_ctl_demote_guarantee_demoting_test.lua
+++ b/test/replication-luatest/gh_9855_box_ctl_demote_guarantee_demoting_test.lua
@@ -1,0 +1,77 @@
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g = t.group()
+
+g.before_each(function(cg)
+    cg.replica_set = replica_set:new{}
+    local box_cfg = {
+        replication = {
+            server.build_listen_uri('server1', cg.replica_set.id),
+            server.build_listen_uri('server2', cg.replica_set.id),
+        },
+        election_mode = 'candidate',
+        replication_timeout = 0.1,
+        election_timeout = 0.5,
+    }
+    cg.server1 = cg.replica_set:build_and_add_server{
+        alias = 'server1',
+        box_cfg = box_cfg,
+    }
+    box_cfg.election_mode = 'voter'
+    cg.server2 = cg.replica_set:build_and_add_server{
+        alias = 'server2',
+        box_cfg = box_cfg,
+    }
+    cg.replica_set:start()
+    cg.replica_set:wait_for_fullmesh()
+end)
+
+g.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+-- Test that the demoted leader can still win in subsequent elections.
+local function test_subsequent_elections(demoted_leader, current_leader)
+    current_leader:exec(function()
+        box.ctl.demote()
+    end)
+    demoted_leader:wait_for_election_leader();
+    demoted_leader:exec(function()
+        t.assert_equals(box.info.election.vote, box.info.election.leader)
+    end)
+    current_leader:exec(function()
+        t.assert_equals(box.info.election.vote, box.info.election.leader)
+    end)
+end
+
+-- Test that the demoted leader cannot get elected after resigning himself
+-- through `box.ctl.demote`.
+g.test_demote_guarantee = function(cg)
+    cg.server1:wait_for_election_leader()
+    cg.server2:update_box_cfg({election_mode = 'candidate'})
+    cg.server1:exec(function()
+        box.ctl.demote()
+    end)
+    cg.server2:wait_for_election_leader();
+    cg.server1:exec(function()
+        t.assert_equals(box.info.election.vote, box.info.election.leader)
+    end)
+    test_subsequent_elections(cg.server1, cg.server2)
+end
+
+-- Test that the demoted leader waits for leader death timeout after resigning
+-- himself through `box.ctl.demote` and starts subsequent elections.
+g.test_demoted_leader_election_in_subsequent_elections = function(cg)
+    cg.server1:wait_for_election_leader()
+    local demoted_term = cg.server1:get_election_term()
+    cg.server1:exec(function()
+        box.ctl.demote()
+    end)
+    cg.server1:wait_for_election_leader()
+    cg.server1:exec(function(demoted_term)
+        t.assert_equals(box.info.election.term, demoted_term + 1)
+        t.assert_equals(box.info.election.vote, box.info.election.leader)
+    end, {demoted_term})
+end
--- a/src/box/raft.c
+++ b/src/box/raft.c
@@ -302,13 +302,7 @@ box_raft_fence(void)
 	raft_resign(raft);
 }
 
-/**
- * Resign RAFT leadership and freeze limbo regardless of
- * box_election_fencing_mode. It waits until the elections
- * begin. After the death-timeout expires, it starts a new
- * round of elections.
- */
-static void
+void
 box_raft_leader_step_off(void)
 {
 	struct raft *raft = box_raft();
--- a/src/box/raft.h
+++ b/src/box/raft.h
@@ -152,6 +152,15 @@ box_raft_set_election_fencing_mode(enum election_fencing_mode mode);
 void
 box_raft_election_fencing_pause(void);
 
+/**
+ * Resign RAFT leadership and freeze limbo regardless of
+ * box_election_fencing_mode. It waits until the elections
+ * begin. After the death-timeout expires, it starts a new
+ * round of elections.
+ */
+void
+box_raft_leader_step_off(void);
+
 void
 box_raft_init(void);
 
--- a/test/replication-luatest/gh_6033_box_promote_demote_test.lua
+++ b/test/replication-luatest/gh_6033_box_promote_demote_test.lua
@@ -108,6 +108,11 @@ end
 local function cluster_reinit(g)
     box_cfg_update(g.cluster.servers, g.box_cfg)
     wait_sync(g.cluster.servers)
+    for _, server in ipairs(g.cluster.servers) do
+        server:exec(function()
+            box.ctl.demote()
+        end)
+    end
 end
 
 local function cluster_stop(g)
--- a/test/replication-luatest/gh_6860_election_off_demote_test.lua
+++ b/test/replication-luatest/gh_6860_election_off_demote_test.lua
@@ -15,6 +15,7 @@ local function replicaset_create(g)
             server.build_listen_uri('server2', g.replicaset.id),
         },
     }
+    g.replication = box_cfg.replication
     g.server1 = g.replicaset:build_and_add_server({
         alias = 'server1', box_cfg = box_cfg
     })
@@ -111,9 +112,21 @@ g.test_election_off_demote_other_no_leader = function(g)
     end)
     -- Server-2 sees that the queue is owned by server-1.
     g.server2:wait_for_synchro_queue_term(g.server1:get_election_term())
+    g.server2:update_box_cfg{replication = ''}
     g.server1:exec(function()
         box.ctl.demote()
+        -- Bump the election term while leaving the synchro queue term intact.
+        local msg = "Not enough peers connected to start elections: 1 out " ..
+                     "of minimal required 2"
+        t.assert_error_msg_content_equals(msg, box.ctl.promote)
+        t.assert_equals(box.info.election.term, box.info.synchro.queue.term + 1)
         box.cfg{election_mode = 'off'}
+    end)
+    g.server2:update_box_cfg({replication = g.replication})
+    g.server2:exec(function()
+        t.helpers.retrying({timeout = 10, delay = 0.5}, function()
+            t.assert_not_equals(box.info.status, "orphan")
+        end)
     end)
     -- Server-2 sees that server-1 is no longer a leader. But the queue still
     -- belongs to the latter.


src/box/box.cc
```diff
@@ -3106,8 +3106,17 @@ box_demote(void)
 		if (raft->state != RAFT_STATE_LEADER)
 			return 0;
 	}
-	if (box_trigger_elections() != 0)
-		return -1;
+	bool is_cfg_candidate_saved = raft->is_cfg_candidate;
```
Serpentian, line 3109: The problem here, that if quorum is 1, the node won't become leader automatically.

```
box.cfg{election_mode = 'candidate'}
box.ctl.demote()
-- Node remains in "follower" state until box.ctl.promote is called.
```

Before this patch Tarantool automatically calls `raft_sm_become_leader` from `raft_worker_handle_io`. However, with this patch the node is not a candidate at all, so it doesn't transition to leader. Moreover, as timer is not started, new term won't happen at all. The node is in "follower" until `box.ctl.promote` is called. This is definitely not a correct behavior.

Serpentian: I also don't really like, that we make raft transit between candidate <-> not candidate state during demote a lot. This causes timer to be restarted. The solution with new internal flag in `raft` may be better. Up to you

Serpentian (changes requested):
Thank you for the patch. Left some comments

test/replication-luatest/gh_9855_box_ctl_demote_guarantee_demoting_test.lua
```diff
@@ -0,0 +1,96 @@
...
+    t.helpers.retrying({timeout = 10, delay = 0.1}, function()
+        cg.server2:exec(function()
+            box.ctl.promote()
+        end)
+    end)
+    cg.server1:exec(function()
+        box.error.injection.set('ERRINJ_RAFT_WAIT_TERM_PERSISTED_DELAY', false)
+    end)
+    cg.server2:wait_for_election_leader()
+    cg.server1:exec(function()
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed: now I wait for the synchronous queue term to get bumped.

src/box/box.cc
CuriousGeorgiy: Reworked the solution following your comments: added a `last_demoted_term` field to `struct raft`, and added a test case to ensure that that demoted leader can start the next elections when he is alone.

src/lib/raft/raft.c
```diff
@@ -729,10 +729,20 @@ raft_worker_handle_io(struct raft *raft)
...
 							  raft->vclock);
+			} else {
+				assert(raft->last_demoted_term ==
+				       raft->volatile_term);
+				/*
+				 * Resigned from the leader role in the current
+				 * term, hence, cannot become a candidate in the
+				 * current elections, so just wait.
+				 */
+				raft_sm_wait_election_end(raft);
```
Serpentian, line 745: Nit: we already do the same logic just one if above, in `raft->vote != 0`. IMHO, it'll be easier to read, if this chunk will also go there with `if (raft->vote != 0 || raft->last_demoted_term == raft->volatile_term)`

src/box/box.cc
```diff
@@ -3106,6 +3106,7 @@ box_demote(void)
 		if (raft->state != RAFT_STATE_LEADER)
 			return 0;
 	}
+	raft_update_last_demoted_term(raft);
```
Serpentian, line 3109: Nit: please, update the commit message, we don't stop the timer anymore.

Serpentian (approved):
Thank you for the patch. LGTM! No major objections

src/lib/raft/raft.c
CuriousGeorgiy: Fixed.

src/box/box.cc
CuriousGeorgiy: Fixed.

src/lib/raft/raft.c
```diff
@@ -723,13 +723,17 @@ raft_worker_handle_io(struct raft *raft)
...
-				 * Voted for some other node. Wait if it manages
-				 * to become a leader.
+				 * Voted for some other node or resigned from
+				 * leadership in the current turn (i.e., cannot
+				 * become a candidate in the current elections),
+				 * so just wait.
 				 */
 				raft_sm_wait_election_end(raft);
-			} else {
+			} else if (raft->last_demoted_term <
```
Serpentian, line 735: Nit: it seems safer to make "voting" a default path and just leave `else without condition. `last_demoted_term` should be in assert.

CuriousGeorgiy: Yeah, sorry I missed this after you suggest to merge the branches — fixed.

src/lib/raft/raft.c
```diff
@@ -723,13 +723,18 @@ raft_worker_handle_io(struct raft *raft)
...
 				else
 					raft_sm_become_candidate(raft);
-			} else if (raft->vote != 0) {
+			} else if (raft->vote != 0 || raft->last_demoted_term ==
+						      raft->volatile_term) {
 				/*
-				 * Voted for some other node. Wait if it manages
-				 * to become a leader.
+				 * Voted for some other node or resigned from
+				 * leadership in the current turn (i.e., cannot
```
sergepetrenko, line 730: typo: turn -> term?
Or maybe rephrase this somehow. "resigned from leadership in the just-ended term", for example.

src/box/box.cc
sergepetrenko, line 3109: TBH I don't really like that we introduce new concepts to the raft library. Maybe we can reuse some existing functions here? It seems, you may replace `box_trigger_elections()` with `box_raft_leader_step_off()`: this way the leader will drop its leadership, but will still watch out for other candidates, and will re-start elections if noone else is found.

Yes, this way the demoted node may become leader right in the next term, but this will only happen if there are no other candidates.

@Serpentian, what do you think of this?

sergepetrenko (changes requested):
Thanks for the patch!

src/box/box.cc
Serpentian: @sergepetrenko `box_raft_leader_step_off` indeed resigns from the leader state, but right after that it calls `raft_restore`, which schedules vote for itself, so we won't really fix the issue. Just tested this, it's still diff

Initially Georgiy had version with `raft_cfg_is_candidate_later`:

```c
	bool is_cfg_candidate_saved = raft->is_cfg_candidate;
	/* This node must not be a candidate in the current elections. */
	raft_cfg_is_candidate_later(raft, false);
	int rc = box_trigger_elections();
	/*
	 * Even if we fail here, we definitely started a new term, so it is
	 * preferable that we are not a candidate in the current term.
	 */
	raft_cfg_is_candidate_later(raft, is_cfg_candidate_saved);
	if (rc != 0)
		return rc;
```

It did prevent node from voting, but it also prevented it from starting the new term at all, which is not great. See [this comment](https://github.com/tarantool/tarantool/pull/10003#discussion_r1597647241). Maybe we can return to this implementation and modify it somehow. Do you like it better? If so, we need to figure out a way, to modify it in the way, that it properly starts new term after resigning without modifying `lib/raft`.

sergepetrenko: >  but right after that it calls raft_restore, which schedules vote for itself, so we won't really fix the issue.

But why? I don't see why this would happen by looking at the code: https://github.com/tarantool/tarantool/blob/39af9fbea6a9c13fe438b9546856416f7bdf1917/src/lib/raft/raft.c#L1139-L1154

Shouldn't we end up in `raft_sm_wait_leader_found` letting other candidates promote themselves?

My suggested diff is (note we skip `box_trigger_elections()` completely for `election_mode 'candidate'/'manual'`:
```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 447237f8da..c14613a4a0 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3105,11 +3105,11 @@ box_demote(void)
                        return 0;
                if (raft->state != RAFT_STATE_LEADER)
                        return 0;
+               box_raft_leader_step_off();
+               return 0;
        }
        if (box_trigger_elections() != 0)
                return -1;
-       if (box_election_mode != ELECTION_MODE_OFF)
-               return 0;
        if (box_try_wait_confirm(2 * replication_synchro_timeout) < 0)
                return -1;
        int64_t wait_lsn = box_wait_limbo_acked(replication_synchro_timeout);
diff --git a/src/box/raft.c b/src/box/raft.c
index b5b97efe00..23eed9419e 100644
--- a/src/box/raft.c
+++ b/src/box/raft.c
@@ -308,7 +308,7 @@ box_raft_fence(void)
  * begin. After the death-timeout expires, it starts a new
  * round of elections.
  */
-static void
+void
 box_raft_leader_step_off(void)
 {
        struct raft *raft = box_raft();
diff --git a/src/box/raft.h b/src/box/raft.h
index 41611d27fc..6d3df7a567 100644
--- a/src/box/raft.h
+++ b/src/box/raft.h
@@ -144,6 +144,9 @@ box_raft_wait_term_persisted(void);
 void
 box_raft_set_election_fencing_mode(enum election_fencing_mode mode);
 
+void
+box_raft_leader_step_off(void);
+
 /**
  * Pause fencing. Instance will not resign its leader role when it looses
  * quorum even if fencing is enabled. Useful to prevent fencing while adding
```

Serpentian: > But why?

Seems, I tested incorrect version. The diff, you propose indeed seems to work. But let's also refactor `box_demote` while we're here, I spent 10 minutes to figure out, what is going on there. Here's the proposed refactoring:

<details>

<summary>Diff</summary>

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 447237f8da..224a7de67f 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -3080,24 +3080,7 @@ box_demote(void)
 		return 0;
 
 	const struct raft *raft = box_raft();
-	if (box_election_mode == ELECTION_MODE_OFF) {
-		assert(raft->state == RAFT_STATE_FOLLOWER);
-		if (raft->leader != REPLICA_ID_NIL) {
-			diag_set(ClientError, ER_NOT_LEADER, raft->leader);
-			return -1;
-		}
-		if (txn_limbo.owner_id == REPLICA_ID_NIL)
-			return 0;
-		/*
-		 * If the limbo term is up to date with Raft, then it might have
-		 * a valid owner right now. Demotion would disrupt it. In this
-		 * case the user has to explicitly overthrow the old owner with
-		 * local promote(), or call demote() on the actual owner.
-		 */
-		if (txn_limbo.promote_greatest_term == raft->term &&
-		    txn_limbo.owner_id != instance_id)
-			return 0;
-	} else {
+	if (box_election_mode != ELECTION_MODE_OFF) {
 		if (txn_limbo_replica_term(&txn_limbo, instance_id) !=
 		    raft->term)
 			return 0;
@@ -3105,11 +3088,28 @@ box_demote(void)
 			return 0;
 		if (raft->state != RAFT_STATE_LEADER)
 			return 0;
+		box_raft_leader_step_off();
+		return 0;
 	}
-	if (box_trigger_elections() != 0)
+
+	assert(raft->state == RAFT_STATE_FOLLOWER);
+	if (raft->leader != REPLICA_ID_NIL) {
+		diag_set(ClientError, ER_NOT_LEADER, raft->leader);
 		return -1;
-	if (box_election_mode != ELECTION_MODE_OFF)
+	}
+	if (txn_limbo.owner_id == REPLICA_ID_NIL)
 		return 0;
+	/*
+	 * If the limbo term is up to date with Raft, then it might have
+	 * a valid owner right now. Demotion would disrupt it. In this
+	 * case the user has to explicitly overthrow the old owner with
+	 * local promote(), or call demote() on the actual owner.
+	 */
+	if (txn_limbo.promote_greatest_term == raft->term &&
+			txn_limbo.owner_id != instance_id)
+		return 0;
+	if (box_trigger_elections() != 0)
+		return -1;
 	if (box_try_wait_confirm(2 * replication_synchro_timeout) < 0)
 		return -1;
 	int64_t wait_lsn = box_wait_limbo_acked(replication_synchro_timeout);
diff --git a/src/box/raft.c b/src/box/raft.c
index b5b97efe00..23eed9419e 100644
--- a/src/box/raft.c
+++ b/src/box/raft.c
@@ -308,7 +308,7 @@ box_raft_fence(void)
  * begin. After the death-timeout expires, it starts a new
  * round of elections.
  */
-static void
+void
 box_raft_leader_step_off(void)
 {
 	struct raft *raft = box_raft();
diff --git a/src/box/raft.h b/src/box/raft.h
index 41611d27fc..3b43d6fa38 100644
--- a/src/box/raft.h
+++ b/src/box/raft.h
@@ -158,6 +158,9 @@ box_raft_init(void);
 void
 box_raft_free(void);
 
+void
+box_raft_leader_step_off(void);
+
 #if defined(__cplusplus)
 }
 #endif

```

</details>

sergepetrenko: Nice!

test/replication-luatest/gh_9855_box_ctl_demote_guarantee_demoting_test.lua
```diff
@@ -0,0 +1,109 @@
...
+
+g.before_each(function(cg)
+    cg.replica_set = replica_set:new{}
+    local box_cfg = {
+        replication = {
+            server.build_listen_uri('server1', cg.replica_set.id),
+            server.build_listen_uri('server2', cg.replica_set.id),
+        },
+        election_mode = 'candidate',
+        replication_timeout = 0.1,
```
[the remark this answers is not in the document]
CuriousGeorgiy: Thanks, will note this in future tests! Fixed.

src/lib/raft/raft.c
CuriousGeorgiy: Dropped this hunk

src/box/box.cc
CuriousGeorgiy: @Serpentian applied your refactoring as a separate patch, @sergepetrenko applied your solution, thanks for the feedback guys!

CuriousGeorgiy: It seems like the current solution creates a regression to 1afe2274: in manual election mode, the demoted leader would always bump the term (now it doesn't), which allowed other nodes to disown the synchronous queue when they issued `box.ctl.demote` with `box.cfg{election_mode = 'off'`:
https://github.com/tarantool/tarantool/blob/8ed6104987bf54d82a7a79cfa2726a57dd889894/src/box/box.cc#L3091-L3099

It breaks this test:
https://github.com/tarantool/tarantool/blob/8ed6104987bf54d82a7a79cfa2726a57dd889894/test/replication-luatest/gh_6860_election_off_demote_test.lua#L101-L131

sergepetrenko: > By the way, while we are here, why do we need both of these checks?
> 
> https://github.com/tarantool/tarantool/blob/ef3152cdc249045be8250c65d47b1a002e49467e/src/box/box.cc#L3104-L3107
> 
> Doesn't the leader always own the synchronous queue?

The desired state for a leader is to own the synchronous queue. But this doesn't happen instantly, just because a node is promoted leader.

For example, when just being elected, Raft leader starts the synchronous queue transition, but it happens asynchronously. See `box_raft_update_synchro_queue()`. So there's a point in time, when `election_state == 'leader'` but `synchro_queue_owner` is someone else.

Also, when a leader steps off or is being overthrown, its state is `follower`, but it still is the synchro queue owner.

sergepetrenko: > TBH, when I was looking for a solution, this snippet boggled me:
> 
> https://github.com/tarantool/tarantool/blob/39af9fbea6a9c13fe438b9546856416f7bdf1917/src/lib/raft/raft.c#L1139-L1154
> 
> It's not immediately obvious that the leader's `leader_witness_map` is always zero, and it's not noted in the comment to `leader_witness_map`.

I re-checked the code, and indeed it seems the leader doesn't keep track of `leader_witness_map`. And nither does it run the election timer (or leader death timer). Which seems quite logical.

But I don't get what's wrong with the snippet you provide. Could you elaborate, please? We know that the node can't be a current leader, because this is a `raft_start_candidate` call (meaning that a node now may become candidate. It doesn't mean it enters "raft candidate" state immediately):
https://github.com/tarantool/tarantool/blob/39af9fbea6a9c13fe438b9546856416f7bdf1917/src/lib/raft/raft.c#L1131-L1139

CuriousGeorgiy: > But I don't get what's wrong with the snippet you provide. Could you elaborate, please? 

Nothing wrong, just noting that it wasn't obvious for me that this branch would be taken by a just demoted leader, and I didn't take into account — would be nice to drop a comment near `leader_witness_map` that it's always zero for the leader.

changelogs/unreleased/gh-9855-box-ctl-demote-guarantee-demoting.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a bug that allowed the old leader to get re-elected after resigning
+  himself through `box.ctl.demote` (gh-9855).
```
sergepetrenko, line 4: let's clarify that this could happen only with `election_mode = 'candidate'`.

src/box/box.cc
sergepetrenko: > To fix this, I suggest to leave everything as is for manual election mode, i.e., call `box_trigger_elections` instead of `box_raft_leader_step_off`. It's obvious that #9855 doesn't occur for manual elections, however, the fact that the demoted leader bumps the Raft term is essential for allowing `box.ctl.demote` to work properly on nodes with disabled elections.
> 
> The problem above can technically still happen with a `voter` node in automatic elections, if nobody triggers new elections, and some `voter` node gets its elections disabled, but I guess it's not a legit case we should try to fix.

I think this is a case where you better rewrite the test than alter `demote()` behavior.
I like the idea that `demote()` behaves consistently with "elections enabled" (`election_mode='manual'/'candidate'). We better keep it this way. I mean, it always keeps the same term and it never writes an actual DEMOTE request.

The problem you mention (with one candidate and many voters) is the same as if all nodes were manual and the leader died. Noone could demote the queue without first promoting itself.

And if the leader is just fine, there's no point to 'demote' it in election_mode = 'manual'. You may first set its election_mode to off and only then issue a demote. Or, alternatively, you may demote it twice. Before and after turning off elections.

The ticket which got fixed by https://github.com/tarantool/tarantool/commit/1afe2274d9155fc19d228237d04d1a1ddfa17270 complained about another issue: even the former leader couldn't release synchro queue without first calling box.ctl.promote().

sergepetrenko (changes requested):
Thanks for the fixes!

src/box/box.cc
CuriousGeorgiy: I don't understand how to create a situation where the synchro queue term is behind the election term with the current patch (which is what the test relies on).

sergepetrenko: For example, you may increase synchro quorum and make a node try to promote. It won't succeed. But the term will be bumped. And the other node will be able to demote.

Although it seems the test becomes almost pointless after the demote rewrite.

changelogs/unreleased/gh-9855-box-ctl-demote-guarantee-demoting.md
CuriousGeorgiy: Fixed.

src/box/box.cc
sergepetrenko: > But the promote still succeeds (AFAIC, because the promote is currently not synchronous?). Or what am I doing wrong?

Oh, yes, I meant promote with `election_mode = 'manual'`. It would trigger elections which the node couldn't win due to huge quorum.
With `election_mode = 'off'` indeed the user is free to do whatever he wants to and may promote an instance without a quorum.

CuriousGeorgiy: I also thought about it, but I got the same result:

```console
tarantool> os.execute('rm -rf *.snap *.xlog *.vylog 512')
---
- 0
...

tarantool> box.cfg{replication_synchro_quorum = 4}
2024-05-27 12:58:50.063 [13625] main/104/interactive I> Tarantool 3.1.0-entrypoint-238-gb3284e10c0 Darwin-x86_64-Debug
2024-05-27 12:58:50.064 [13625] main/104/interactive I> log level 5 (INFO)
2024-05-27 12:58:50.064 [13625] main/104/interactive I> wal/engine cleanup is paused
2024-05-27 12:58:50.065 [13625] main/104/interactive I> mapping 268435456 bytes for memtx tuple arena...
2024-05-27 12:58:50.065 [13625] main/104/interactive I> Actual slab_alloc_factor calculated on the basis of desired slab_alloc_factor = 1.044274
2024-05-27 12:58:50.065 [13625] main/104/interactive I> mapping 134217728 bytes for vinyl tuple arena...
2024-05-27 12:58:50.077 [13625] main/104/interactive I> RAFT: fencing paused
2024-05-27 12:58:50.077 [13625] main/104/interactive I> instance uuid 24e89994-b751-4aae-99a0-78065d23ecb9
2024-05-27 12:58:50.078 [13625] main/104/interactive I> initializing an empty data directory
2024-05-27 12:58:50.133 [13625] main/104/interactive I> assigned id 1 to replica 24e89994-b751-4aae-99a0-78065d23ecb9
2024-05-27 12:58:50.134 [13625] main/104/interactive I> replicaset uuid 737f2a1d-99f9-4d0f-88af-18e460d83edc
2024-05-27 12:58:50.137 [13625] snapshot/101/main I> saving snapshot `./00000000000000000000.snap.inprogress'
2024-05-27 12:58:50.141 [13625] snapshot/101/main I> done
2024-05-27 12:58:50.143 [13625] main/104/interactive I> ready to accept requests
2024-05-27 12:58:50.143 [13625] main/105/gc I> wal/engine cleanup is resumed
2024-05-27 12:58:50.143 [13625] main/104/interactive/box.load_cfg I> set 'replication_synchro_quorum' configuration option to 4
2024-05-27 12:58:50.143 [13625] main/106/checkpoint_daemon I> scheduled next checkpoint for Mon May 27 13:59:34 2024
---
...

tarantool> box.cfg{election_mode = 'manual'}
2024-05-27 12:58:58.366 [13625] main/104/interactive I> RAFT: start state machine
2024-05-27 12:58:58.366 [13625] main/104/interactive/box.load_cfg I> set 'election_mode' configuration option to "manual"
---
...

tarantool> box.ctl.promote()
2024-05-27 12:59:04.798 [13625] main/104/interactive I> RAFT: bump term to 2, follow
2024-05-27 12:59:04.798 [13625] main/104/interactive I> RAFT: vote for 1, follow
2024-05-27 12:59:04.799 [13625] main/114/raft_worker I> RAFT: persisted state {term: 2, vote: 1}
2024-05-27 12:59:04.799 [13625] main/114/raft_worker I> RAFT: enter leader state with quorum 1
---
...

```

CuriousGeorgiy: Actually, it's enough to just stop one of the two servers. 

Fixed: stopped one of the servers to make box.ctl.promote fail and then restarted it.

sergepetrenko: Thanks for the fixes!

sergepetrenko (approved):
LGTM.

sergepetrenko:
@CuriousGeorgiy, some tests still fail in CI:
`replication-luatest/gh_6860_election_off_demot>                 [ fail ]`
https://github.com/tarantool/tarantool/actions/runs/9284403504/job/25546715760?pr=10003#step:8:12557

CuriousGeorgiy:
> riousGeorgiy, some tests still fail in CI:
> `replication-luatest/gh_6860_election_off_demot> [ fail ]`
> [tarantool/tarantool/actions/runs/9284403504/job/25546715760?pr=10003#step:8:12557](https://github.com/tarantool/tarantool/actions/runs/9284403504/job/25546715760?pr=10003#step:8:12557)

@sergepetrenko fixed: added a synchronization point to wait for the restarted server to synchronize with the leader.

sergepetrenko:
Backports:
3.1 https://github.com/tarantool/tarantool/pull/10067
2.11 https://github.com/tarantool/tarantool/pull/10068

PR #9999, opened 2024-05-08
test/interactive: a couple of enhancements

Totktonada:
This patchset proposes a couple of enhancements for the `test.interactive_tarantool` testing helper.

* `:roundtrip()` now able to verify a response against `nil` and returns the response.
* The hide/show prompt feature (#7169) is now disabled by default.
* The new `connect()` function makes it easier to work with a remote console.

See the commit messages for details.

Part of #9985

PR #10014, opened 2024-05-15
console: autorequire frequently used modules

### LINKED ISSUE #9986: console: autorequire modules suitable for administration and debugging
It is convenient to have some built-in modules accessible right after connection to the admin console (using `tt connect`) or after starting a debugging session (using `tarantool -i`).

These are ones from where an admin wants to obtain state information: `compat`, `fiber.info()`, `config:info()`, `config:get()` are such examples. `fun` is often used to filter/transform such information.

There are modules that implement a data type and they're used to write a value of the given type to the database: `uuid`, `datetime`, `decimal`, `varbinary`.

Also, some modules are often used in interactive experimenting: `clock.bench()`, `ffi.cast()` and serializers such as `json`, `yaml`, `msgpack`.

There are function that are useful for debugging: `fio.open(<...>):read()`, `log.info()`, `popen.shell()`.

The proposed list of modules:

* clock
* compat
* config
* datetime
* decimal
* ffi
* fiber
* fio
* fun
* json
* log
* msgpack
* popen
* uuid
* varbinary
* yaml

I propose to don't add these names to globals, but make them accessible only in the interactive console: a code in the Lua files still need to `require` the modules. It is easy to do after implementing console session scoped variables, see #9985.

I propose to autorequire the modules in the interactive console session under the same `console_session_scope_vars` compat option.

Totktonada:
There are built-in modules that are frequently used for administration or debugging purposes. It is convenient to have them accessible in the interactive console without extra actions.

They're accessible now without a manual `require` call if the `console_session_scope_vars` compat option is set to `new` (see also tarantool/doc#4191).

The list of the autorequired modules is below.

* clock
* compat
* config
* datetime
* decimal
* ffi
* fiber
* fio
* fun
* json
* log
* msgpack
* popen
* uuid
* varbinary
* yaml

See tarantool/tarantool#9986 for motivation behind this feature.

This list forms so called initial environment for an interactive console session. The default initial environment may be adjusted by an application, for example, to include application specific administrator functions.

Two public functions are added for this purpose: `console.initial_env()` and `console.set_initial_env(env)`.

Example 1 (keep autorequired modules, but add one more variable):

```lua
local console = require('console')

-- Add myapp_info function.
local initial_env = console.initial_env()
initial_env.myapp_info = function()
    <...>
end
```

Example 2 (replace the whole initial environment):

```lua
local console = require('console')

-- Add myapp_info function, discard the autorequired modules.
console.set_initial_env({
    myapp_info = function()
        <...>
    end,
})
```

The `console.set_initial_env()` call without an argument or with a `nil` argument drops the initial environment to its default.

A modification of the initial environment doesn't affect existing console sessions. It affects console sessions that are created after the modification.

Fixes #9986

### DIFF
--- a/changelogs/unreleased/console-autorequire.md
+++ b/changelogs/unreleased/console-autorequire.md
@@ -1,0 +1,5 @@
+## feature/console
+
+* Interactive console now automatically adds frequently used built-in modules
+  into the initial environment if the `console_session_scope_vars` compat
+  option is set to `new` (gh-9986).
--- a/src/box/lua/console.lua
+++ b/src/box/lua/console.lua
@@ -52,6 +52,8 @@ compat.add_option({
     action = function() end,
 })
 
+local M = {}
+
 --
 -- Default output handler set to YAML for backward
 -- compatibility reason.
@@ -174,7 +176,7 @@ local function parse_output(value)
     return nil, fmt, opts, local_eos
 end
 
-local function set_default_output(value)
+function M.set_default_output(value)
     if value == nil then
         error("Nil output value passed")
     end
@@ -186,7 +188,7 @@ local function set_default_output(value)
     default_output_format["opts"] = opts
 end
 
-local function get_default_output(...)
+function M.get_default_output(...)
     local args = ...
     if args ~= nil then
         error("Arguments provided while prohibited")
@@ -236,7 +238,7 @@ end
 --
 -- Set/get current console EOS value from
 -- currently active output format.
-local function console_eos(eos_value)
+function M.eos(eos_value)
     if not eos_value then
         return tostring(current_eos())
     end
@@ -268,7 +270,7 @@ end
 --
 -- Set delimiter
 --
-local function delimiter(delim)
+function M.delimiter(delim)
     local self = fiber.self().storage.console
     if self == nil then
         error("console.delimiter(): need existing console")
@@ -288,7 +290,7 @@ local function set_delimiter(_storage, value)
         return error('Can not install delimiter for net box sessions')
     end
     value = value or ''
-    return delimiter(value)
+    return M.delimiter(value)
 end
 
 local function set_language(storage, value)
@@ -475,7 +477,7 @@ local function local_eval(storage, line)
     return format(unpack(res, 1, res.n))
 end
 
-local function eval(line)
+function M.eval(line)
     return local_eval(box.session.storage, line)
 end
 
@@ -859,7 +861,7 @@ local function repl(self)
     fiber.self().storage.console = nil
 end
 
-local function on_start(foo)
+function M.on_start(foo)
     if foo == nil or type(foo) == 'function' then
         repl_mt.__index.on_start = foo
         return
@@ -867,7 +869,7 @@ local function on_start(foo)
     error('Wrong type of on_start hook: ' .. type(foo))
 end
 
-local function on_client_disconnect(foo)
+function M.on_client_disconnect(foo)
     if foo == nil or type(foo) == 'function' then
         repl_mt.__index.on_client_disconnect = foo
         return
@@ -878,7 +880,7 @@ end
 --
 --
 --
-local function ac(yes_no)
+function M.ac(yes_no)
     local self = fiber.self().storage.console
     if self == nil then
         error("console.ac(): need existing console")
@@ -890,7 +892,7 @@ end
 -- Start REPL on stdin
 --
 local started = false
-local function start()
+function M.start()
     if started then
         error("console is already started")
     end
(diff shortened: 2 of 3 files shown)


Buristan (approved):
Hi, Alexander!
Thanks for the patchset!
I'll proceed with the review per-patch now.

**[PATCH 1/2] console: define module table earlier**
LGTM, as trivial.

src/box/lua/console.lua
```diff
@@ -417,14 +420,68 @@ local function table_pack(...)
...
+-- command modifies the environment.
+--
+-- The initial environment is returned by console.initial_env().
+-- It is a table that may be modified by a user or completely
+-- replaced with a new table using console.set_initial_env().
+-- These calls don't affect existing console sessions, only ones
+-- that are created afterwards.
+function M.initial_env()
+    if initial_env == nil then
+        initial_env = {
```
Buristan, line 442: What do you think about adding `ffi` by default too?
It is not about the user experience but more about the developer experience. Sometimes it is useful to check some `ffi` triks in the console.

test/app-luatest/console_autorequire_test.lua
```diff
@@ -0,0 +1,171 @@
...
+    'uuid',
+    'varbinary',
+    'yaml',
+}
+
+local function assert_no_autorequire(it)
+    it:roundtrip("require('strict').off()")
+
+    for _, m in ipairs(autorequire_list) do
+        it:roundtrip(('type(%s)'):format(m), 'nil')
```
Buristan, line 30: Why do you use `type(modulename)` here instead of just `modulename`? #9999 allows you to use the `nil` check in `roundtrip`, doesn't it?

test/app-luatest/console_autorequire_test.lua
```diff
@@ -0,0 +1,171 @@
...
+    'log',
+    'msgpack',
+    'popen',
+    'uuid',
+    'varbinary',
+    'yaml',
+}
+
+local function assert_no_autorequire(it)
+    it:roundtrip("require('strict').off()")
```
Buristan, line 27: Does it mean that user should disable `strict` for the console each time by hands too?

src/box/lua/console.lua
```diff
@@ -417,14 +420,68 @@ local function table_pack(...)
...
+-- Set initial console environment.
+--
+-- Accepts a table or a nil value. If the value is nil, the
+-- initial environment is reset to a default value.
+--
+-- See console.initial_env() for details.
+function M.set_initial_env(env)
+    if type(env) ~= 'table' and type(env) ~= 'nil' then
+        error(('console.set_initial_env: expected table or nil, got %s'):format(
+            type(env)), 0)
```
Buristan, line 472: Why is the error level 0 and not 2?

Buristan (approved):
**[PATCH 2/2] console: autorequire frequently used modules**
Hi, Alexnader!
Thanks for the patch.
Generally, LGTM, but have some ignorable nits below.

src/box/lua/console.lua
Totktonada: No problems, added.

test/app-luatest/console_autorequire_test.lua
Totktonada: There is no idea behind it, I just implemented the positive and the negative checks similarly. Changed to `modulename`.

test/app-luatest/console_autorequire_test.lua
Totktonada: I need it to receive the same response for Debug and RelWithDebInfo builds in the test. I don't think that it is a problem for a user of the interactive console.

For example, on Debug build (the strict mode is enabled):

```lua
tarantool> unknown_var
---
- error: '[string "return unknown_var"]:1: variable ''unknown_var'' is not declared'
...
```

It seems OK for me.

src/box/lua/console.lua
Totktonada: That's my personal preference (we can't judge what level is a cause of the problem in the general case), but I don't mind to set it to 2. Changed.

test/app-luatest/console_autorequire_test.lua
Buristan: Got it, thanks for the clarifications!

changelogs/unreleased/console-autorequire.md
```diff
@@ -0,0 +1,5 @@
+## feature/console
+
+* Interactive console automatically adds a couple of frequently used
```
xuniq, line 3:
```suggestion
* Interactive console now automatically adds frequently used
```

xuniq (approved):

sergos (approved):
Thanks for the patch!
LGTM

changelogs/unreleased/console-autorequire.md
Totktonada: Thanks! Fixed.

Issue #5536, opened 2020-11-19
OOM on a joining/syncing replica


sergepetrenko:
### Tarantool version:
2.4.1+ (probably 2.2.1+, since parallel applier introduction)

### Bug description:
Replica fetches snapshot successfully, then its log is spammed with 
```
too long WAL write: X rows at LSN N: 187.076 sec
```
There may be literally millions of such messages: `25004156 messages suppressed`
And lag may be as high as 480 seconds in some cases.

After some time the connection with master is broken. Master's log shows `relay/<ipaddr>/101/main coio.cc:340 !> SystemError timed out: Connection timed out`.

Replica restarts the connection and the story starts all over again.

During this process replicas memory consumption continues to grow.

After some more retries one of the following happens:
* replica receives `ER_INVALID_MSGPACK` error (hooray, #4040 finally has a reproducer)
   Replication is stopped after this message. This lets replica process all the piled up
   messages in its WAL, its cpu usage drops to zero, but the consumed RAM doesn't decrease.
* replica is killed by OOM

#### The hard way
Give the master enough load so that while the replica is fetching its snapshot, it manages to write roughly the same amount of data to its `.xlog` files.

Now the replica will fetch the snapshot successfully (during this phase no WAL writes are involved, the replica just applies whatever the master sends it, so it doesn't fall back).
When the master finishes sending its snapshot, it already has a pile of transactions in its `.xlog` files, and it immediately starts sending them to the replica. Unlike the first stage, now replica tries to write the transactions to WAL. There is no limit for the speed relay or applier send/receive transactions at, so replica receives the transactions at RPS much higher than master does (the master is simply reading the transactions from the already written WAL, and does it much faster, than could write the transactions to WAL itself).

Replica's WAL is never in time to write all the incoming transactions, so replica has more and more open transactions and cbus messages between tx and WAL, which leads to increasing memory consumption and, eventually, OOM.

#### The easy way
1. Build a debug version of tarantool.
2. Start replica & master.
3. Create some load on master (not necessarily high-RPS).
4. Issue `box.error.injection.set("ERRINJ_WAL_DELAY", true)` on replica
5. Wait some time, issue `box.error.injection.set("ERRINJ_WAL_DELAY", false)` and repeat steps 4-5

I tried it with the following load on master:
```lua
i = 0
whie true
    box.space.test:insert{i}
    if i % 100000 == 0 then
         print("Processed "..i)
    end
    i = i + 1
end
```

[...]

PR #10015, opened 2024-05-16
wal: fix wal_queue_max_size assignment during initial box.cfg

### LINKED ISSUE #10013: `wal_queue_max_size` is set too late on initial `box.cfg`
**Bug description**

`wal_queue_max_size` is a dynamic cfg parameter, and it's applied by `load_cfg` only after `box_cfg_xc()` finishes.
In case `box_cfg_xc` includes syncing with a remote master, all the syncing wil happen with default `wal_queue_max_size` setting, 16 megabytes. Turns out, the default is too big for some applications when rows are tiny and wal_queue allows hundreds of thousands of rows to enter the queue in one event loop cycle. If this is the case, the same symptoms already discussed in #5536 appear, and the node is never in time to sync with the master.

Reproduced on Tarantool 2.8.4

sergepetrenko:
wal_queue_max_size took effect only after the initial box.cfg call, meaning that users with non-zero `replication_sync_timeout` still synced using the default 16 Mb queue size. In some cases the default was too big and the same issues described in #5536 arose.

Fix this.

Closes #10013

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-10013-wal-queue-max-size.md
+++ b/changelogs/unreleased/gh-10013-wal-queue-max-size.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* Fixed the `wal_queue_max_size` configuration option not being applied during
+  the initial configuration (gh-10013).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -5510,6 +5510,8 @@ box_cfg_xc(void)
 	box_set_replication_skip_conflict();
 	if (box_check_instance_name(cfg_instance_name) != 0)
 		diag_raise();
+	if (box_set_wal_queue_max_size() != 0)
+		diag_raise();
 	cfg_replication_anon = box_check_replication_anon();
 	box_broadcast_ballot();
 	/*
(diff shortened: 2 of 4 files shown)


changelogs/unreleased/gh-10013-wal-queue-max-size.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/box
+
+* Fixed `wal_queue_max_size` configuration option not being applied during
+  initial configuration (gh-10013).
```
xuniq, line 4:
```suggestion
* Fixed the `wal_queue_max_size` configuration option not being applied during the
  initial configuration (gh-10013).
```

xuniq (approved):

Serpentian (approved):
Thank you for the patch! No objections. Looks great!

Gerold103 (approved):
Thanks for the patch!

changelogs/unreleased/gh-10013-wal-queue-max-size.md
sergepetrenko: Thanks! Fixed.

sergepetrenko:
Backports:
3.1 https://github.com/tarantool/tarantool/pull/10035
2.11 https://github.com/tarantool/tarantool/pull/10034

PR #10016, opened 2024-05-16
vinyl: fix bug when tuple not committed to unique nullable index

### LINKED ISSUE #9769: Vinyl unique index by `is_nullable = true` field loses tuples
reproduced with `3.0.1-0-g31c2ddb`, `2.11.2-0-g1bac2d257b`

```lua

box.cfg({ log_level = "error" })

local space = box.schema.space.create("SPACE", {
    engine = "vinyl",
    format = {
        { name = "ID", type = "unsigned" },
        { name = "ANY", type = "scalar", is_nullable = true },
    },
    is_sync = false,
})

space:create_index("PRIMARY", {
    unique = true,
    type = "TREE",
    parts = { { field = "ID", type = "unsigned" } },
})

space:create_index("SECONDARY", {
    unique = false,
    type = "TREE",
    parts = { { field = "ANY", type = "scalar" } },
})

space:create_index("SECONDARY_UNIQUE", {
    unique = true,
    type = "TREE",
    parts = { { field = "ANY", type = "scalar" } },
})

local function show(...)
    print(require("json").encode({ ... }))
end

local function run(v1, v2)
    space:insert({ 1, v1 })

    box.atomic(function()
        space:replace({ 1, v2 })
        space:insert({ 2, v1 })
    end)

    print()
    show(space.index.PRIMARY:select({}, { fullscan = true }))
    show(space.index.SECONDARY:select({}, { fullscan = true }))
    show(space.index.SECONDARY_UNIQUE:select({}, { fullscan = true }))

    space:truncate()
end

local data = { 1, false, true, "" }

for _, vi in ipairs(data) do
    for _, vj in ipairs(data) do
        if vi ~= vj then
            run(vi, vj)
        end
    end
end

for _, v in ipairs(data) do
    run(v, box.NULL)
end

-- but

for _, v in ipairs(data) do
    run(box.NULL, v)
end

```

Out:
```yaml

[[[1,false],[2,1]]]
[[[1,false],[2,1]]]
[[[1,false]]]

[[[1,true],[2,1]]]
[[[1,true],[2,1]]]
[[[1,true]]]

[[[1,""],[2,1]]]
[[[2,1],[1,""]]]
[[[1,""]]]

[[[1,1],[2,false]]]
[[[2,false],[1,1]]]
[[[1,1]]]

[[[1,true],[2,false]]]
[[[2,false],[1,true]]]
[[[1,true]]]

[[[1,""],[2,false]]]
[[[2,false],[1,""]]]
[[[1,""]]]

[[[1,1],[2,true]]]
[[[2,true],[1,1]]]
[[[1,1]]]

[[[1,false],[2,true]]]
[[[1,false],[2,true]]]
[[[1,false]]]

[[[1,""],[2,true]]]
[[[2,true],[1,""]]]
[[[1,""]]]

[[[1,1],[2,""]]]
[[[1,1],[2,""]]]
[[[1,1]]]

[[[1,false],[2,""]]]
[[[1,false],[2,""]]]
[[[1,false]]]

[[[1,true],[2,""]]]
[[[1,true],[2,""]]]
[[[1,true]]]

[[[1,null],[2,1]]]
[[[1,null],[2,1]]]
[[[1,null]]]

[[[1,null],[2,false]]]
[[[1,null],[2,false]]]
[[[1,null]]]

[[[1,null],[2,true]]]
[[[1,null],[2,true]]]
[[[1,null]]]

[[[1,null],[2,""]]]
[[[1,null],[2,""]]]
[[[1,null]]]

# but

[[[1,1],[2,null]]]
[[[2,null],[1,1]]]
[[[2,null],[1,1]]]

[[[1,false],[2,null]]]
[[[2,null],[1,false]]]
[[[2,null],[1,false]]]

[[[1,true],[2,null]]]
[[[2,null],[1,true]]]
[[[2,null],[1,true]]]

[[[1,""],[2,null]]]
[[[2,null],[1,""]]]
[[[2,null],[1,""]]]

```

Expected: index (3 line) doesn't lose tuple

Works correctly when `is_nullable = false`, `engine = "memtx"` or with other transaction combinations:

```lua
space:insert({ 1, v1 })
space:replace({ 1, v2 })
space:insert({ 2, v1 })
```

```lua
box.atomic(function()
    space:insert({ 1, v1 })
    space:replace({ 1, v2 })
end)

space:insert({ 2, v1 })

```

locker:
A unique nullable key definition extended with primary key parts (`cmp_def`) assumes that two tuples are equal *without* comparing primary key fields if all secondary key fields are equal and not nulls, see `tuple_compare_slowpath()`. This is a hack required to ignore the uniqueness constraint for nulls in memtx. The memtx engine can't use the secondary key definition as is (`key_def`) for comparing tuples in the index tree, as it does for a non-nullable unique index, because this wouldn't allow insertion of any duplicates, including nulls. It couldn't use `cmp_def` without this hack, either, because then conflicting tuples with the same secondary key fields would always compare as not equal due to different primary key parts.

For Vinyl, this hack isn't required because it explicitly skips the uniqueness check if any of the indexed fields are nulls, see `vy_check_is_unique_secondary()`. Furthermore, this hack is harmful because Vinyl relies on the fact that two tuples compare as equal by `cmp_def` if and only if *all* key fields (both secondary and primary) are equal. For example, this is used in the transaction manager, which overwrites statements equal by `cmp_def`, see `vy_tx_set_entry()`.

Let's disable this hack by resetting `unique_part_count` in `cmp_def`.

Closes #9769

### DIFF
--- a/changelogs/unreleased/gh-9769-vy-unique-nullable-index-fix.md
+++ b/changelogs/unreleased/gh-9769-vy-unique-nullable-index-fix.md
@@ -1,0 +1,4 @@
+## bugfix/vinyl
+
+* Fixed a bug in the transaction manager when a tuple was not committed to
+  a unique nullable secondary index (gh-9769).
--- a/src/box/vy_lsm.c
+++ b/src/box/vy_lsm.c
@@ -142,6 +142,31 @@ vy_lsm_new(struct vy_lsm_env *lsm_env, struct vy_cache_env *cache_env,
 
 	struct key_def *key_def = key_def_dup(index_def->key_def);
 	struct key_def *cmp_def = key_def_dup(index_def->cmp_def);
+	/*
+	 * A unique nullable key definition extended with primary key parts
+	 * (cmp_def) assumes that two tuples are equal *without* comparing
+	 * primary key fields if all secondary key fields are equal and not
+	 * nulls, see tuple_compare_slowpath(). This is a hack required to
+	 * ignore the uniqueness constraint for nulls in memtx. The memtx
+	 * engine can't use the secondary key definition as is (key_def) for
+	 * comparing tuples in the index tree, as it does for a non-nullable
+	 * unique index, because this wouldn't allow insertion of any
+	 * duplicates, including nulls. It couldn't use cmp_def without this
+	 * hack, either, because then conflicting tuples with the same
+	 * secondary key fields would always compare as not equal due to
+	 * different primary key parts.
+	 *
+	 * For Vinyl, this hack isn't required because it explicitly skips
+	 * the uniqueness check if any of the indexed fields are nulls, see
+	 * vy_check_is_unique_secondary(). Furthermore, this hack is harmful
+	 * because Vinyl relies on the fact that two tuples compare as equal by
+	 * cmp_def if and only if *all* key fields (both secondary and primary)
+	 * are equal. For example, this is used in the transaction manager,
+	 * which overwrites statements equal by cmp_def, see vy_tx_set_entry().
+	 *
+	 * We disable this hack by resetting unique_part_count in cmp_def.
+	 */
+	cmp_def->unique_part_count = cmp_def->part_count;
 
 	lsm->cmp_def = cmp_def;
 	lsm->key_def = key_def;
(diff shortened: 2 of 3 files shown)


xuniq (approved):

nshy (approved):
LGTM.

locker:

PR #10023, opened 2024-05-17
vinyl: fix index name in duplicate key error message

### LINKED ISSUE #5975: Duplicate-key error message for vinyl says index name is null
I have Ubuntu 20.04
Tarantool version: Tarantool 2.7.1-186-g365e67c51   pulled from source today

I execute these requests:
```lua
box.execute([[CREATE TABLE j (s1 INT PRIMARY KEY, s2 STRING) WITH ENGINE='vinyl';]])
box.space.J:create_index('I3', {parts = {2, 'string'}})
box.space.J:create_index('I5', {parts = {2, 'string'}})
box.execute([[INSERT INTO j VALUES (7, 'a');]])
box.space.J.index.I3:drop()
box.execute([[INSERT INTO j VALUES (8, 'a');]])
```
Result: error message =
Duplicate key exists in unique index '(null)' in space 'J'
... There is a duplicate key but there is no unique index named '(null)'.
The expected behaviour is: error message =
Duplicate key exists in unique index 'I5' in space 'J'
.

locker:
The code setting `ER_TUPLE_FOUND` uses `index_name_by_id()` to find the index name, but it passes an index in the dense index map to it while the function expects an index in the sparse index map. Apparently, this doesn't work as expected after an index is removed from the middle of the index map. This bug was introduced by commit fc3834c01e9e6.

Instead of just fixing the index passed to `index_name_by_id()`, we do a bit of refactoring. We stop passing `index_name` and `space_name` to `vy_check_is_unique_*()` functions and instead get them right before raising `ER_TUPLE_FOUND`. Note, to get the space name, we need to call `space_by_id()` but it should be fine because (a) the space is very likely to be cached as the last accessed one and (b) this is an error path so it isn't performance critical. We also drop `index_name_by_id()` and extract the index name from the LSM tree object.

Closes #5975

### DIFF
--- a/changelogs/unreleased/gh-5975-vy-duplicate-key-error-fix.md
+++ b/changelogs/unreleased/gh-5975-vy-duplicate-key-error-fix.md
@@ -1,0 +1,3 @@
+## bugfix/vinyl
+
+* Fixed a bug when a duplicate key error referred to a wrong index (gh-5975).
--- a/src/box/space.c
+++ b/src/box/space.c
@@ -833,15 +833,6 @@ space_index_def(struct space *space, int n)
 	return space->index[n]->def;
 }
 
-const char *
-index_name_by_id(struct space *space, uint32_t id)
-{
-	struct index *index = space_index(space, id);
-	if (index != NULL)
-		return index->def->name;
-	return NULL;
-}
-
 /**
  * Pushes arguments for replace triggers (on_replace, before_replace)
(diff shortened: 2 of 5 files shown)


nshy (approved):

xuniq (approved):

locker:
The `integration/php-queue` test failure isn't caused by this fix.

locker:

PR #10028, opened 2024-05-17
box: introduce next and previous prefix iterators

### LINKED ISSUE #9994: new iterator for TREE index: `NP` (NextPrefix)
example table contains PRIMARY KEY

```yaml
- foo
- foo bar
- foo bar baz
- g
- g abc
- g cde
- g cde def
```

```lua
for _, tuple in space:pairs('foo', { iterator = 'GE' }) do
   print(tuple)
end

- foo
- foo bar
...
```

```lua
for _, tuple in space:pairs('foo', { iterator = 'NP' }) do
  print(tuple)
end

- g
- g abc

```

the iterator works like `GT` but for the first key it can receive a string that can be part of the field (prefix substring).
the first iteration must return value which prefix is more than noticed in key.

alyapunov:
Implement 'np' (next prefix) and 'pp' (previous prefix) iterators. They work only in memtx tree and in a nutshell searches for strings with greater ('np') or less ('pp') prefix of size as in given key, comparing with given key.

Closes #9994

@TarantoolBot document
Title:'np' and 'pp' (next/previous prefix) iterators

Now there are two more iterators avaialbe: 'np' (next prefix) and 'pp' (previous prefix). They work only in memtx tree. Also, if the last part of key is not a string, they degrade to 'gt' and 'lt' iterators.

These iterators introduce special comparison of the last part of key (if it is a string). In terms of lua, if s is the search part, and t is the corresponding tuple part, 'np' iterator searches for the first tuple with string.sub(t, 1, #s) > s, while 'pp' searches for the last tuple with string.sub(t, 1, #s) < s.

Comparison of all other parts of the key remains normal.

As usual, these iterators are available both in select and pairs, in index and space methods.

Similar to all other tree iterators, they change only initial search of selection. Once the first tuple found, the rest are selected sequentially in direct (for 'np') or reverse (for 'pp') order of the index.

For example:
```
tarantool> s:select{}
---
- - ['a']
  - ['aa']
  - ['ab']
  - ['b']
  - ['ba']
  - ['bb']
  - ['c']
  - ['ca']
  - ['cb']
...

tarantool> s:select({'b'}, {iterator = 'np'})
---
- - ['c']
  - ['ca']
  - ['cb']
...

tarantool> s:select({'b'}, {iterator = 'pp'})
---
- - ['ab']
  - ['aa']
  - ['a']
...
```

### DIFF
--- a/changelogs/unreleased/gh-9994-next-prefix-iterator.md
+++ b/changelogs/unreleased/gh-9994-next-prefix-iterator.md
@@ -1,0 +1,4 @@
+## feature/core
+
+* Implemented new iterators for the memtx tree index: 'np' (next prefix)
+  and 'pp' (previous prefix) (gh-9994).
--- a/test/box-luatest/gh_9994_next_prefix_iterator_test.lua
+++ b/test/box-luatest/gh_9994_next_prefix_iterator_test.lua
@@ -1,0 +1,405 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new()
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.after_each(function(cg)
+    cg.server:exec(function()
+        if box.space.test then
+            box.space.test:drop()
+        end
+        if box.func.func then
+            box.func.func:drop()
+        end
+    end)
+end)
+
+-- Check some cases when NP and PP iterators are unsupported..
+g.test_next_prefix_unsupported = function(cg)
+    cg.server:exec(function()
+        local s = box.schema.space.create('test')
+        s:create_index('pk', {type = 'hash', parts = {{1, 'string'}}})
+        local m = "Index 'pk' (HASH) of space 'test' (memtx) " ..
+                  "does not support requested iterator type"
+        t.assert_error_msg_content_equals(m, s.select, s, '', {iterator = 'np'})
+        t.assert_error_msg_content_equals(m, s.select, s, '', {iterator = 'pp'})
+
+        s:create_index('sk', {parts = {{1, 'str', collation = 'unicode_ci'}}})
+        local i = s.index.sk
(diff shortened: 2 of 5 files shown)


drewdzzz (approved):
LGTM

changelogs/unreleased/gh-9994-next-prefix-iterator.md
```diff
@@ -0,0 +1,4 @@
+## feature/core
+
+* Implemented new iterators for memtx tree index: 'np' (next prefix)
```
xuniq, line 3:
```suggestion
* Implemented new iterators for the memtx tree index: 'np' (next prefix)
```

xuniq (approved):

test/box-luatest/gh_9994_next_prefix_iterator_test.lua
```diff
@@ -0,0 +1,406 @@
...
+        s:replace{1, '/home/user/data1', 0}
+        s:replace{1, '/home/user/data2', 0}
+        s:replace{1, '/home/user/folder/', 0}
+        s:replace{1, '/home/user/folder/subfolder/data.txt', 0}
+        s:replace{1, '/home/user/folder1/subfolder/data.txt', 0}
+        s:replace{1, '/home/user/bin/tarantool', 1}
+        s:replace{1, '/home/user/bin/tarantool_ctl', 1}
+        s:replace{1, '/home/user/work/tarantool/src/main.cc', 0}
+        s:replace{1, '/home/user/work/tarantool/src/main.cc', 1}
+        s:replace{1, '/home/user/work/tarantool/src/main.cc', 0}
```
Gumix, line 352: This line looks redundant.

Gumix (approved):

changelogs/unreleased/gh-9994-next-prefix-iterator.md
alyapunov: Fixed, thanks.

alyapunov:
Both problems that lint was found are false-positive.

PR #10046, opened 2024-05-23
net.replicaset: fix a flaky test

alyapunov:
Fix several problems that led to unexpected results:
* Handle error thrown by box.clt.promote. Not very often but it could fail ruining the test. Fixed by retry.
* It's not so obvious but RO server after successful call of box.ctl.promote for some time remains RO. On the other hand both net.replicaset and the test rely box.info.ro to define the leader, for in their view the actual switch to the new leader happens some (perhaps different) time after call to box.ctl.promote. That asynchrony sometimes led to unexpected results. Fixed by waiting of RW status.
* The fix above (wait for RW of the new leader) introduced new problem: in case with sequential leader change and broadcast of new values, the new leader has good chances (but not 100%) to deliver old broadcasted value to the watcher. Fix it by splitting the sequence to the leader change with verification and broadcast with verification.

Follow-up #9823

NO_DOC=fix flaky test
NO_CHANGELOG=fix flaky test

### DIFF
--- a/test/box-luatest/gh_9823_net_replicaset_test.lua
+++ b/test/box-luatest/gh_9823_net_replicaset_test.lua
@@ -173,7 +173,8 @@ g.test_net_replicaset_basics = function(cg)
     -- Change the leader.
     local new_leader_name = get_another_storage_name(leader_name)
     cluster[new_leader_name]:exec(function()
-        box.ctl.promote()
+        t.helpers.retrying({timeout = 10, delay = 0.1}, box.ctl.promote)
+        box.ctl.wait_rw()
     end)
 
     -- Check rebind to the new leader.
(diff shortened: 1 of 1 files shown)


test/box-luatest/gh_9823_net_replicaset_test.lua
```diff
@@ -322,7 +325,10 @@ g.test_watch_leader = function(cg)
     -- Change the leader.
```
sergepetrenko, line 325: Found another flaky fail when testing locally:
```
[002] not ok 4T>box-luatest.gh_9823_net_replicaset.test_watch_leader                                                           
[002] #   ...e/sptnt/test/box-luatest/gh_9823_net_replicaset_test.lua:295: expected: 1, actual: 2
[002] #   stack traceback: 
```

sergepetrenko (commented):
Thanks for the patch!

test/box-luatest/gh_9823_net_replicaset_test.lua
alyapunov: Added another fix.

sergepetrenko: Thanks! I can't reproduce the failure anymore.

sergepetrenko (approved):
Thanks for the fix!

PR #10048, opened 2024-05-24
ci: temporary disable aarch64 packaging job

ylobankov:
We are experiencing some issues with aarch64 runners at this moment, so we have to disable aarch64 packaging job until the issue is resolved. Other aarch64 jobs can be disabled via GitHub UI, so let's do it there.


foxzi (approved):
OK

PR #10052, opened 2024-05-24
config: give better error if a role is not a table

### LINKED ISSUE #10049: Consider an invalid role module
If the role module (required with `roles:`) returns an object, which couldn't be indexed with `[]`, tarantool stops with an unexpected message (... lua:143: attempt to index local 'role' (a boolean value)). I think it would be nice to handle this case to improve user experience.

https://github.com/tarantool/tarantool/blob/ef3152cdc249045be8250c65d47b1a002e49467e/src/box/lua/config/applier/roles.lua#L140-L148

Totktonada:
An attempt to use a non-table module as a role now reports a more descriptive error:

> Unable to use module <...> as a role: expected table, got <...>

Fixes #10049

### DIFF
--- a/changelogs/unreleased/config-not-a-table-role-error-message.md
+++ b/changelogs/unreleased/config-not-a-table-role-error-message.md
@@ -1,0 +1,4 @@
+## bugfix/config
+
+* Improved an error message on an attempt to use a non-table module as a role
+  (gh-10049).
--- a/src/box/lua/config/applier/roles.lua
+++ b/src/box/lua/config/applier/roles.lua
@@ -138,6 +138,11 @@ local function post_apply(config)
         if not role then
             log.verbose('roles.post_apply: load role ' .. role_name)
             role = require(role_name)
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/config-not-a-table-role-error-message.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Improved an error message on attempt to use a non-table module as a role
```
andreyaksenov, line 3:
```suggestion
* Improved an error message on an attempt to use a non-table module as a role
```

andreyaksenov (approved):

changelogs/unreleased/config-not-a-table-role-error-message.md
Totktonada: Thanks! Fixed.

Totktonada:
Backported to `release/3.1` (future `3.1.1` release) in PR #10054.

PR #10060, opened 2024-05-28
[DRAFT]: Collect, aggregate and send perf statistics to InfluxDB

Buristan:
This patchset standardizes the output of all benchmarks in the perf directory to have similar output to [the Google Benchmark](https://github.com/google/benchmark). It aggregates all results in the <perf_output/perf_summary.txt> file in [InfluxDB line protocol](https://docs.influxdata.com/influxdb/v1/write_protocols/line_protocol_tutorial/) format and sends this file in a separate step to the corresponding InfluxDB bucket.

### DIFF
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -72,6 +72,19 @@ jobs:
           revision: ${{ inputs.revision }}
       - name: test
         run: make -f .test.mk test-perf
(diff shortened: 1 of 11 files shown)


Buristan:
Closed in favor of #10061, since secrets can't be used from any fork repository until code is merged to master.

Issue #7089, opened 2022-04-27
Race when accessing `replicaset.vclock` from applier thread


sergepetrenko:
Found while debugging hangs in tarantool/tarantool-qa#166

One of the nodes logs contains the following output:
```console
2022-04-26 17:32:07.767 [36183] relay/unix/:(socket)/102/main:reader V> Got a corrupted row:
2022-04-26 17:32:07.767 [36183] relay/unix/:(socket)/102/main:reader V> 00000000: 81 00 00 81 26 81 01 09 02 01
2022-04-26 17:32:07.767 [36183] main/117/applier/unix/:(socket) I> RAFT: message {term: 4, vote: 2, leader: 2, leader is seen: true, state: follower} from 3
2022-04-26 17:32:07.767 [36183] main/117/applier/unix/:(socket) I> RAFT: vote request is skipped - the leader is already known - 2
2022-04-26 17:32:07.767 [36183] relay/unix/:(socket)/101/main xrow.c:229 E> ER_INVALID_MSGPACK: Invalid MsgPack - packet body
```

The issue is with bad vclock encoding:
```markdown
| header: map of 1 element
|    | type: IPROTO_OK
|   / \  | body: map of 1 element
V  /   \ V
81 00 00 81 26 81 01 09 02 01
               \            /
                   vclock
```

The vclock is incorrect: `81` states that the vclock contains a single replica_id : lsn pair, but two follow: `01 09` and `02 01`.

With the following patch one of the nodes occasionally fails with an assertion in `xrow_encode_vclock` inside `applier_thread_writer_f`:
```diff
diff --git a/src/box/xrow.c b/src/box/xrow.c
index bbd697f11..2447fb8f2 100644
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -62,6 +62,7 @@ mp_sizeof_vclock_ignore0(const struct vclock *vclock)
 static inline char *
 mp_encode_vclock_ignore0(char *data, const struct vclock *vclock)
 {
+       int map_size = vclock_size_ignore0(vclock);
        data = mp_encode_map(data, vclock_size_ignore0(vclock));
        struct vclock_iterator it;
        vclock_iterator_init(&it, vclock);
@@ -72,6 +73,7 @@ mp_encode_vclock_ignore0(char *data, const struct vclock *vclock)
        for ( ; replica.id < VCLOCK_MAX; replica = vclock_iterator_next(&it)) {
                data = mp_encode_uint(data, replica.id);
[...]

PR #10061, opened 2024-05-29
Collect, aggregate and send perf statistics to InfluxDB

Buristan:
This patchset standardizes the output of all benchmarks in the perf directory to have similar output to [the Google Benchmark](https://github.com/google/benchmark). It aggregates all results in the <perf/output/summary.txt> file in [InfluxDB line protocol](https://docs.influxdata.com/influxdb/v1/write_protocols/line_protocol_tutorial/) format and sends this file in a separate step to the corresponding InfluxDB bucket.


perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
+-- File to aggregate the benchmark results from JSON files to the
```
Totktonada, line 1: How about a shebang (`#!/usr/bin/env tarantool`)?

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+
+local input_dir = params.input_dir
+assert(input_dir and fio.path.is_dir(input_dir),
+       'given input_dir is not a directory')
+
+local output = params.output
+local out_fh = assert(io.open(output, 'w'))
+
+local function exec(cmd)
+    return io.popen(cmd):read('*all'):gsub('^%s+', ''):gsub('%s+$', '')
```
Totktonada, line 35: Since it is definitely a tarantool specific script, we can use `:strip()` instead of `:gsub('^%s+', ''):gsub('%s+$', '')`.

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+        for _, field in ipairs(REPORTED_FIELDS) do
+            -- XXX: The stub for the `small` benchmark.
+            if bench[field] then
+                table.insert(field_set, ('%s=%s'):format(field, bench[field]))
+            end
+        end
+
+        local line = ('%s,%s %s %d\n'):format(
+            bench_name, full_tag_set, table.concat(field_set, ','), time
+        )
```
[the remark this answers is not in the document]
Totktonada: I see, I misinterpreted `time` with a value that is reported by a benchmark, but I guess that the main idea of my comment is clear.

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+    {'input_dir', 'string'},
+    {'output', 'string'},
+})
+
+local input_dir = params.input_dir
+assert(input_dir and fio.path.is_dir(input_dir),
+       'given input_dir is not a directory')
+
+local output = params.output
+local out_fh = assert(io.open(output, 'w'))
```
Totktonada, line 32: Since we anyway use `fio` in this script, I would use it for all the file operations. `assert(fio.open(output, {'O_WRONLY', 'O_CREAT', 'O_TRUNC'})` in this case.

`O_RDONLY` below in `read_all`.

perf/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 set(CMAKE_CXX_STANDARD 14)
 
+set(PERF_OUTPUT_DIR ${PROJECT_BINARY_DIR}/perf_output)
+file(MAKE_DIRECTORY ${PERF_OUTPUT_DIR})
```
Totktonada, line 4: How about `perf/output` instead of `perf_output` on the top level?

perf/lua/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ function(create_perf_lua_test)
                     COMMAND env
                       LUA_PATH="${LUA_PATH}"
                       ${TARANTOOL_BIN} ${TEST_PATH}
+                        --output="${PERF_OUTPUT_DIR}/${PERF_NAME}.json"
+                        --output_format=json
```
Totktonada, line 27: Raw idea: make the options verbatim follow the `gbench` ones. This way we can add more useful options from `gbench` step-by-step. And possibly end with a convenient benchmarking microframework with some compatibility with `gbench` for integrations of various kinds.

perf/lua/box_select.lua
```diff
@@ -1,27 +1,34 @@
 --
 -- The test measures run time of various space select operations.
 --
--- Output format:
--- <test-case> <run-time-nanoseconds>
+-- Output format (console):
+-- <test-case> <iterations-per-second>
```
Totktonada, line 5: I would ask @locker to glance, because he is the author of the benchmark.

perf/lua/box_select.lua
```diff
@@ -222,6 +229,15 @@ local TESTS = {
...
 }
 
+local function single_run(func)
+    local real_time_start = clock.time()
+    local cpu_time_start = clock.proc()
+    func()
+    local delta_real = clock.time() - real_time_start
+    local delta_cpu = clock.proc() - cpu_time_start
+    return delta_real, delta_cpu
+end
```
Totktonada, line 239: I see that a similar helper is added to several benchmarks. How about providing it directly from the `benchmark` module?

perf/lua/box_select.lua
```diff
@@ -1,27 +1,34 @@
...
 
 local clock = require('clock')
 local fiber = require('fiber')
 
 local params = require('internal.argparse').parse(arg, {
+    {'output', 'string'},
+    {'output_format', 'string'},
     {'pattern', 'string'},
     {'read_view', 'boolean'},
 })
```
Totktonada, line 25: Maybe wrap `argparse.parse()` into some `benchmark.parse_args()` function to add more options consistently to all the benchmarks?

perf/lua/uri_escape_unescape.lua
```diff
@@ -24,23 +31,37 @@ local decode_sample = uri.escape(encode_sample, escape_opts)
...
+    local real_delta = clock.time() - real_time
+    local cpu_delta = clock.proc() - cpu_time
+    bench:add_result(testname, real_delta, cpu_delta, CYCLES)
 end
-local encode_time = cycles / (clock.monotonic() - start)
 
-start = clock.monotonic()
-for _ = 1, cycles do
-    uri.unescape(decode_sample)
+for testname, test_func in pairs(tests) do
```
Totktonada, line 63: I'm a bit worrying about an unpredictable order of the subtests.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,63 @@
...
+    }
+    table.insert(bench.results, result)
+    return result
+end
+
+local function dump_results(bench)
+    dump_report(bench, format_report(bench))
+end
+
+function M.init(opts)
```
Totktonada, line 51: As for me, it creates an object rather than initializaes a module. So I would name it `new()`.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,63 @@
...
+    if output then
+        local fh = assert(io.open(output, 'w'))
+        fh:write(text)
+        fh:close()
+    else
+        io.stdout:write(text)
+    end
+end
+
+local function add_result(bench, name, real_time, cpu_time, items)
```
Totktonada, line 34: How about named options (`opts` table)? It is easy to misplace the real time and the CPU time for example.

.gitignore
```diff
@@ -42,6 +42,7 @@ doc/*/Makefile
 extra/Makefile
 extra/*/Makefile
 perf/Makefile
+perf/lua/Makefile
```
Totktonada, line 45: The PR title says `[DRAFT] <...>`, but you asked me for review. Aren't you forgot to change the title?

perf/CMakeLists.txt
```diff
@@ -5,6 +5,16 @@ file(MAKE_DIRECTORY ${PERF_OUTPUT_DIR})
 
 add_subdirectory(lua)
 
+set(TARANTOOL_BIN $<TARGET_FILE:tarantool>)
+add_custom_target(test-perf-aggregate
+                  DEPENDS test-c-perf test-lua-perf
+                  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+                  COMMAND ${TARANTOOL_BIN} ${CMAKE_CURRENT_SOURCE_DIR}/tools/aggregate.lua
+                    --output=${PERF_OUTPUT_DIR}/perf_summary.txt
```
Totktonada, line 13: I would use just `${PERF_OUTPUT_DIR}/summary.txt` or `${PERF_OUTPUT_DIR}/summary_influx.txt`.

Totktonada (approved):
Thank you for a decent work!

I'm happy that we have this activity in the team.

I've no objections regarding the proposed patches.

Following our tradition, I've sent a couple of nits :) Nothing that would block the patchset and some comments are just thoughts around. Feel free to close them without actions and proceed further.

.gitignore
Buristan: For sure! Thanks! Updated the title.

perf/bps_tree.cc
```diff
@@ -120,6 +120,7 @@
 	type##_##func##_size_##size(benchmark::State &state) \
 	{ \
 		test_##func<type>(state, size); \
+		state.SetItemsProcessed(size); \
```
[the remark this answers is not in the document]
mkostoevr: Thanks, I'm OK with these changes.

I like the activity too BTW.

perf/lua/CMakeLists.txt
```diff
@@ -18,7 +20,9 @@ function(create_perf_lua_test)
   message(STATUS "Creating Lua performance test ${PERF_NAME}_perftest")
   set(TEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua)
   add_custom_target(${PERF_NAME}_perftest
-                    COMMAND ${TARANTOOL_BIN} ${TEST_PATH}
+                    COMMAND env
```
ligurio, line 23: it is better to use builtin `env`:

```
COMMAND ${CMAKE_COMMAND} -E env
        LUA_PATH="${LUA_PATH}"
        ${TARANTOOL_BIN} ${TEST_PATH}
```

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,63 @@
...
+
+local M = {}
+
+local function format_report(bench)
+    local output_format = bench.output_format
+    local results = bench.results
+    local report = ''
+    if output_format == 'json' then
+        -- The output should have the same format as the Google
+        -- Benchmark.
```
ligurio, line 11: Google Benchmark supports three different output formats: `console`, `json`, and `csv` [^1]. I would say explicitly that Google Benchmark JSON output format. Probably link to Google Benchmark and a sample with output will be useful in comments.

[^1]: https://github.com/google/benchmark/blob/main/docs/user_guide.md

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,63 @@
+local json = require('json')
+
+local M = {}
+
+local function format_report(bench)
+    local output_format = bench.output_format
+    local results = bench.results
+    local report = ''
+    if output_format == 'json' then
```
ligurio, line 9: This will not work with `output_format` equal to `JSON`. I would use `string.lower(output_format)` to avoid this. The same everywhere where you compare `output_format` with constants.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,63 @@
...
+        -- The output should have the same format as the Google
+        -- Benchmark.
+        report = json.encode({benchmarks = results})
+    else
+        assert(output_format == 'console', 'unkonwn output format')
```
ligurio, line 14: Why not an `error()` here and below?

perf/lua/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 set(TARANTOOL_BIN $<TARGET_FILE:tarantool>)
```
ligurio, line 1: probably a reason of `NO_DOC`, `NO_CHANGELOG`, `NO_TEST` should be `infrastracture.` Feel free to ignore.

perf/lua/1mops_write.lua
```diff
@@ -20,12 +20,16 @@ local params = require('internal.argparse').parse(arg, {
     {'nodes', 'number'},
```
ligurio, line 20: commit message probably should be ` perf: standardize an output of 1mops_write benchmark `

perf/lua/uri_escape_unescape.lua
```diff
@@ -5,6 +5,13 @@ local uri = require("uri")
 local clock = require("clock")
```
ligurio, line 5: commit message to "perf: standardize an output of uri_escape_unescape benchmark "

perf/lua/uri_escape_unescape.lua
```diff
@@ -24,23 +31,37 @@ local decode_sample = uri.escape(encode_sample, escape_opts)
 -- iterations slower than next ones. To avoid such effect we can introduce
 -- "warmup" iterations, disable JIT at all or make JIT compilation
 -- aggressive from the beginning.
---   - hotloop=1 enables compilation in loops on first iteration.
---   - hotexit=1 enables faster compilation for root traces as well as side
---     traces.
+--   - hotloop=1 enables compilation in loops after the first iteration.
```
ligurio, line 34: it is not related to commit, I would move it to a separate commit

perf/lua/box_select.lua
```diff
@@ -1,27 +1,34 @@
 --
```
ligurio, line 1: commit message to "perf: standardize an output of box_select benchmark "

perf/lua/1mops_write.lua
```diff
@@ -20,12 +20,16 @@ local params = require('internal.argparse').parse(arg, {
...
     {'nohint', 'boolean'},
     {'ops', 'number'},
+    {'output_format', 'string'},
+    {'output', 'string'},
     {'sync', 'boolean'},
```
ligurio, line 25: I would ask @sergos to glance, because he is the author of the benchmark.

perf/lua/column_scan.lua
```diff
@@ -1,13 +1,15 @@
 --
```
ligurio, line 1: commit message to " perf: standardize an output of column_scan benchmark "

perf/CMakeLists.txt
```diff
@@ -5,6 +5,16 @@ file(MAKE_DIRECTORY ${PERF_OUTPUT_DIR})
```
ligurio, line 5: don't forget to update a dir name in commit message, see a comment from Alexander in a previous commit

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+local json = require('json')
+local fio = require('fio')
+
+local params = require('internal.argparse').parse(arg, {
+    {'input_dir', 'string'},
+    {'output', 'string'},
+})
+
+local input_dir = params.input_dir
+assert(input_dir and fio.path.is_dir(input_dir),
```
ligurio, line 28: why `assert` and not an `error`? here and below

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+local REPORTED_FIELDS = {
+    'cpu_time',
+    'items_per_second',
+    'iterations',
+    'real_time',
+}
+
+local time = os.time()
+
+for _, file in pairs(fio.listdir(input_dir)) do
```
ligurio, line 63: I don't get why do you need to traverse a dir and search JSON reports if we know these reports in advance. I would remove a loop here and leave a function that accepts a path to a filename of report. Call `aggregate.lua` in CMake and pass to it paths to all benchmark reports.

ligurio (commented):
Sergey, thanks for the patches! Please take a look on my comments.

Looking forward to merging your work to a master branch.

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,93 @@
...
+-- The reported tag set is a set of values that can be used for
+-- filtering data (i.e., branch or benchmark name).
+--
+-- The script accepts the following parameters:
+--
+-- <input_dir> -- the directory from which the .json files are
+--                taken.
+-- <output>    -- the filename where the results are saved.
+--
+-- [1]: https://docs.influxdata.com/influxdb/v1/write_protocols/line_protocol_tutorial/
```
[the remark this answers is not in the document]
Buristan: AFAICS, the second version is backward compatible with the first version (at least I don't see any contradictions with it inside this helper). So, I updated the link to the new one.

perf/tools/aggregate.lua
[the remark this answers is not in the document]
Buristan: Rearranged, as you suggested, thanks!

perf/tools/aggregate.lua
Buristan: Added, thanks!

perf/tools/aggregate.lua
Buristan: Fixed, thanks!

perf/tools/aggregate.lua
Buristan: Since I don't need to specify the error level in the script. Also, the additional branch looks not so pretty to me.

perf/tools/aggregate.lua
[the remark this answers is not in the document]
Buristan: I suppose that this is a not general tool to be used Tarantool-wide (for now), but it is OK only for performance testing.

perf/tools/aggregate.lua
Buristan: Traversing them in script is just a little bit more convenient than specifying all paths in the command -- it may lead to a huge command in the future (like we have now with `LUA_CPATH` for our LuaJIT tests). Also, it allows you to easily test the script in the console while updating it.
Ignoring for now.

perf/lua/CMakeLists.txt
Buristan: I've thought about it and prefer to set a different flags intentionally. The status of the `gbench` isn't very clear for me.
For example, google/benchmark#794 reports that CSV format is depricated and will be dropped soon, but still there is no progress. Also, I am not sure how we want to match `--benchmark_min_time`, `--benchmark_min_warmup_time`, `--benchmark_repetitions` with our Lua benchmarks and how they are useful. Plus, this `benchmark` prefix in flags looks obsolete to me.

But I'm OK to rename flags to match the `gbench` ones if you insist.

perf/lua/box_select.lua
Buristan: I'm not sure about its usage compared to the other benchmarks, since for this particular benchmark we want to just sum all timings, but this may not be suitable for others. So, I prefer to postpone this decision until the moment when we will have some common pattern for writing such benchmarks. Maybe this functionality should be implemented via the `--benchmark_repetitions` option.

perf/lua/benchmark.lua
Buristan: Fixed at the moment of initialization, thanks!

perf/lua/benchmark.lua
Buristan: Clarified the comment and added the link, thanks!

perf/lua/1mops_write.lua
Buristan: In that case, the commit title is more than 50 symbols, so I shorten it to this one.

perf/lua/uri_escape_unescape.lua
Buristan: In that case, the commit title is more than 50 symbols, so I shorten it to this one.

perf/lua/box_select.lua
Buristan: In that case, the commit title is more than 50 symbols, so I shorten it to this one.

perf/lua/column_scan.lua
Buristan: In that case, the commit title is more than 50 symbols, so I shorten it to this one.

perf/lua/CMakeLists.txt
Buristan: I've replaced with `perf test` as one that was used before.

perf/lua/uri_escape_unescape.lua
Buristan: Fixed, changed the tests structure and replaced `pairs()` with `ipairs()`.

perf/lua/benchmark.lua
Buristan: Sure, updated.

perf/lua/benchmark.lua
Buristan: Replaced with `new()`.

perf/lua/box_select.lua
Buristan: Added a wrapper, so the usage is the following:
```lua
local benchmark = require('benchmark')

local opts = benchmark.argparse(arg, {--[[custom options]]})
local bench = benchmark.new(opts)
```

perf/lua/uri_escape_unescape.lua
Buristan: Splitted to the separate commit, thanks!

perf/lua/CMakeLists.txt
Buristan: Fixed, thanks!

perf/CMakeLists.txt
Buristan: Replaced.

perf/CMakeLists.txt
Buristan: Replaced with `${PERF_OUTPUT_DIR}/summary.txt`.

perf/CMakeLists.txt
Buristan: Updated, thanks!

perf/tools/aggregate.lua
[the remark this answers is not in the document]
Totktonada: No problem. It was just an idea.

Buristan:
I've also noticed that outside pull_request we should use just `github.ref` and `github.sha`, when the job is running per commit to the `master` or `release/*` branch.
The last patch is updated:
```diff
diff --git a/fa6fc45df8 b/2d6aceb881
index fa6fc45df8..2d6aceb881 100644
--- a/fa6fc45df8
+++ b/2d6aceb881
@@ -53,8 +53,16 @@ jobs:
     timeout-minutes: 60

     env:
-        PERF_BRANCH: ${{ github.event.pull_request.head.ref }}
-        PERF_COMMIT: ${{ github.event.pull_request.head.sha }}
+        PERF_BRANCH: ${{ github.event.pull_request.head.ref || github.ref }}
+        PERF_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}

     steps:
       - name: Prepare checkout

```

perf/lua/CMakeLists.txt
Totktonada: OK, I've no objection. Just checked the idea with you.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,106 @@
...
+--
+-- These options are parsed by the module by default:
+-- -- output = 'string',
+-- -- output_format = 'string',
+-- local opts = benchmark.argparse(arg, {
+--     <..your options..>
+-- })
+-- benchmark.new(
+--     require('internal.argparse').parse(opts)
+-- )
```
Totktonada, line 14: `benchmark.argparse` and `argparse.parse` are called both in this example. It seems, it is by a mistake.

Totktonada, line 14: `benchmark.new` result is not saved to a variable.

Buristan: Fixed, thanks!

Buristan: Fixed, thanks!

perf/lua/benchmark.lua
ligurio: Thanks!

perf/lua/1mops_write.lua
```diff
@@ -11,7 +11,9 @@ local math = require('math')
 local json = require('json')
 local fio = require('fio')
 
-local params = require('internal.argparse').parse(arg, {
+local benchmark = require('benchmark')
```
sergos, line 14: Can we make it optional, so that standalone 1mops still can be used?

sergos (changes requested):

perf/lua/1mops_write.lua
Buristan: It can be run standalone, like the following:
```sh
LUA_PATH="perf/lua/?.lua;;" src/tarantool perf/lua/1mops_write.lua 
# 150 operations per txn,
# using 50 fibers,
# in a replicaset of 1 nodes,
# using HASH index type
# with WAL mode write
# 
# Warmup... done, lsn:  1006207
# master done 8991300 ops in time: 16.221424, cpu: 21.494186
# master average speed  554285  ops/sec
# master peak speed     664726  ops/sec
1mops_master 554285 rps
```
Do you have concerns about the `LUA_PATH` declaration?

locker (approved):
`perf/lua/column_scan.lua` LGTM

perf/lua/box_select.lua
```diff
@@ -1,27 +1,33 @@
...
 -- The test measures run time of various space select operations.
 --
--- Output format:
--- <test-case> <run-time-nanoseconds>
+-- Output format (console):
+-- <test-case> <iterations-per-second>
 --
 -- Options:
--- --pattern <string>  run only tests matching the pattern; it's possible
```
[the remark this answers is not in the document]
ligurio: seems fixed already :)

perf/lua/1mops_write.lua
Buristan: It is always required since we always have common parts for formatting (even in console mode) and timing aggregation (moreover, timing counting may be moved to this module later, when we will have a clear vision of the API).

perf/lua/column_scan.lua
```diff
@@ -1,13 +1,15 @@
...
+-- Output format (console):
+-- <test-case> <rows-per-second>
 --
 -- Options:
--- --engine <string>         space engine to use for the test
 -- --column_count <number>   number of columns in the test space
+-- --engine <string>         space engine to use for the test
 -- --row_count <number>      number of rows in the test space
+-- --output <string>         file to output
+-- --output_format <string>  console (default)|json
```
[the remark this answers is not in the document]
Buristan: I agree with you that "raw" or "plain" is better, but Google Benchmark uses "console" as an output option, so I use the same name for consistency.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,105 @@
...
+    local output_format = bench.output_format
+    local results = bench.results
+    local report = ''
+    if output_format == 'json' then
+        -- The output should have the same format as the Google
+        -- Benchmark JSON output format:
+        -- https://github.com/google/benchmark/blob/main/docs/user_guide.md
+        report = json.encode({benchmarks = results})
+    else
+        assert(output_format == 'console', 'unkonwn output format')
```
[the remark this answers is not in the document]
Buristan: Fixed, thanks!

.github/workflows/perf_micro.yml
```diff
@@ -72,6 +76,19 @@ jobs:
...
+        # in the output in case of an error.
+        # --fail: Exit with the 22 error code is status >= 400.
+        # -w: See the reason for the failure, if any.
+        run: >
+          curl --request POST
+          "${{ secrets.INFLUXDB_URL }}/api/v2/write?org=tarantool&bucket=perf-debug&precision=s"
+          -w "%{http_code}"
+          --fail --silent -o /dev/null
+          --header "Authorization: Token ${{ secrets.INFLUXDB_TOKEN_DEBUG }}"
+          --data-binary @./perf/output/summary.txt
```
ligurio, line 91: Running benchmarks is a time-consuming task, it would be a shame if `curl` will fail to upload results.
I would add options to automatically retry sending in case of fail.

```
curl --connect-timeout 5 \
    --max-time 10 \
    --retry 5 \
    --retry-delay 0 \
    --retry-max-time 40 \
    <URL>


--max-time 10     (how long each retry will wait)
--retry 5         (it will retry 5 times)
--retry-delay 0   (an exponential backoff algorithm)
--retry-max-time  (total time before it's considered failed)
```

perf/CMakeLists.txt
```diff
@@ -1,7 +1,20 @@
 set(CMAKE_CXX_STANDARD 14)
```
ligurio, line 1: Seems `perf/gh-7089-vclock-copy/` is not integrated to a build system and its results doesn't send to InfluxDB.
I propose to fix this.

perf/tools/aggregate.lua
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env tarantool
+
+-- File to aggregate the benchmark results from JSON files to the
+-- format parsable by the InfluxDB line protocol [1]:
```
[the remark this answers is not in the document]
Buristan: The tool is still working, but it prints the difference in RPS instead of seconds (due to the output change). I suppose that this is OK, since we can still analyze the difference.
```sh
LUA_PATH="perf/lua/?.lua;;" src/tarantool perf/tools/compare.lua /tmp/t1 /tmp/t1 /tmp/t2
       /tmp/t1          /tmp/t1          /tmp/t2
get_0  3100261  3100261 (+  0%)  2958616 (-  4%)
```
I've moved it to the tools directory to avoid confusion.

perf/CMakeLists.txt
Buristan: I've added the corresponding commit. I would ask @sergepetrenko to glance, because he is the author of the benchmark.

perf/lua/box_select.lua
```diff
@@ -1,27 +1,33 @@
...
 --
+-- --output <string>        file to output
+-- --output_format <string> console (default)|json
 
 local clock = require('clock')
 local fiber = require('fiber')
+local benchmark = require('benchmark')
 
-local params = require('internal.argparse').parse(arg, {
+local params = benchmark.argparse(arg, {
```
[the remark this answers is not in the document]
Buristan: Added usage to all benchmarks. Each benchmark can provide custom help to the `argparse` routine to be printed before the general usage (`help`, `output`, `output_format` options).

.github/workflows/perf_micro.yml
Buristan: I've added the following options. 
```
--retry 5
--retry-delay 5
--connect-timeout 120
```

perf/tools/aggregate.lua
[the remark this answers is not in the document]
ligurio: Okay, thanks!

perf/lua/box_select.lua
ligurio: Volodya gave LGTM, so this can be closed.

https://github.com/tarantool/tarantool/pull/10061#pullrequestreview-2098975704

perf/lua/box_select.lua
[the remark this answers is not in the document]
ligurio: Thanks!

perf/lua/uri_escape_unescape.lua
```diff
@@ -1,9 +1,19 @@
...
--- taskset -c 1 tarantool uri_escape_unescape.lua
-
 local uri = require("uri")
 local clock = require("clock")
 local t = require("tarantool")
+local benchmark = require("benchmark")
+
+local USAGE = [[
+
+ Usage: taskset -c 0x1 tarantool uri_escape_unescape.lua
```
ligurio, line 8: `Usage: taskset -c 0x1 tarantool uri_escape_unescape.lua` -> ` Usage: tarantool uri_escape_unescape.lua [options]`.

perf/lua/benchmark.lua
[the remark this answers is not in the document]
ligurio: Thanks!

.github/workflows/perf_micro.yml
ligurio: Thanks!

.github/workflows/perf_micro.yml
```diff
@@ -72,6 +76,25 @@ jobs:
           revision: ${{ inputs.revision }}
       - name: test
         run: make -f .test.mk test-perf
+      - name: Send statistics to InfluxDB
+        # TODO: For now, use the debug bucket for this PoC.
+        # --silent -o /dev/null: Prevent dumping any reply part
+        # in the output in case of an error.
+        # --fail: Exit with the 22 error code is status >= 400.
+        # -w: See the reason for the failure, if any.
```
ligurio, line 84: Long option is  `--write-out`.

perf/CMakeLists.txt
```diff
@@ -11,7 +24,7 @@ if (NOT ${benchmark_FOUND})
                       COMMENT ${MSG}
     )
     add_custom_target(test-perf
-                      DEPENDS test-c-perf test-lua-perf
```
ligurio, line 14: I propose to make independent targets for running tests and aggregating results. Discussed verbally and PoC is below:

```diff
diff --git a/perf/CMakeLists.txt b/perf/CMakeLists.txt
index 805a8eaf78..609ebfe608 100644
--- a/perf/CMakeLists.txt
+++ b/perf/CMakeLists.txt
@@ -5,9 +5,11 @@ file(MAKE_DIRECTORY ${PERF_OUTPUT_DIR})
 
 add_subdirectory(lua)
 
+set(BENCH_RESULTS "")
+
 set(TARANTOOL_BIN $<TARGET_FILE:tarantool>)
 add_custom_target(test-perf-aggregate
-                  DEPENDS test-c-perf test-lua-perf
+                  DEPENDS ${BENCH_RESULTS}
                   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
                   COMMAND ${TARANTOOL_BIN} ${CMAKE_CURRENT_SOURCE_DIR}/tools/aggregate.lua
                     --output=${PERF_OUTPUT_DIR}/summary.txt
@@ -24,7 +26,7 @@ if (NOT ${benchmark_FOUND})
                       COMMENT ${MSG}
     )
     add_custom_target(test-perf
-                      DEPENDS test-perf-aggregate
+                      DEPENDS test-c-perf test-lua-perf
                       COMMENT "Running performance tests"
     )
     return()
@@ -51,14 +53,17 @@ function(create_perf_test_target)
                         "${multiValues}"
                         ${ARGN})
   message(STATUS "Creating C performance test ${PERF_TARGET}_perftest")
+  set(BENCH_RESULT ${PERF_OUTPUT_DIR}/${PERF_TARGET}.json)
   add_custom_target(${PERF_TARGET}_perftest
                     COMMAND "$<TARGET_FILE:${PERF_TARGET}.perftest>"
                             "--benchmark_out_format=json"
-                            "--benchmark_out=${PERF_OUTPUT_DIR}/${PERF_TARGET}.json"
+                            "--benchmark_out=${BENCH_RESULT}"
                     DEPENDS ${PERF_TARGET}.perftest
+                    BYPRODUCTS ${BENCH_RESULT}
                     COMMENT Running ${PERF_TARGET}_perftest
   )
   set(RUN_PERF_C_TESTS_LIST ${RUN_PERF_C_TESTS_LIST} ${PERF_TARGET}_perftest PARENT_SCOPE)
+  set(BENCH_RESULTS "${BENCH_RESULTS} ${BENCH_RESULT}" PARENT_SCOPE)
 endfunction()
 
 function(create_perf_test)
@@ -110,6 +115,6 @@ add_custom_target(test-c-perf
 )
 
 add_custom_target(test-perf
-                  DEPENDS test-perf-aggregate
+                  DEPENDS test-c-perf test-lua-perf
                   COMMENT "Running performance tests"
 )
diff --git a/perf/lua/CMakeLists.txt b/perf/lua/CMakeLists.txt
index 55d523f92e..40000e5ddf 100644
--- a/perf/lua/CMakeLists.txt
+++ b/perf/lua/CMakeLists.txt
@@ -19,17 +19,20 @@ function(create_perf_lua_test)
 
   message(STATUS "Creating Lua performance test ${PERF_NAME}_perftest")
   set(TEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${PERF_NAME}.lua)
+  set(BENCH_RESULT "${PERF_OUTPUT_DIR}/${PERF_NAME}.json")
   add_custom_target(${PERF_NAME}_perftest
                     COMMAND ${CMAKE_COMMAND} -E env
                       LUA_PATH="${LUA_PATH}"
                       ${TARANTOOL_BIN} ${TEST_PATH}
-                        --output="${PERF_OUTPUT_DIR}/${PERF_NAME}.json"
+                        --output="${BENCH_RESULT}"
                         --output_format=json
                     COMMENT Running ${PERF_NAME}_perftest
                     DEPENDS tarantool ${TEST_PATH}
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+                    BYPRODUCTS ${BENCH_RESULT}
   )
   set(RUN_PERF_LUA_TESTS_LIST ${RUN_PERF_LUA_TESTS_LIST} ${PERF_NAME}_perftest PARENT_SCOPE)
+  set(BENCH_RESULTS "${BENCH_RESULTS} ${BENCH_RESULT}" PARENT_SCOPE)
 endfunction()
 
 create_perf_lua_test(NAME 1mops_write)

```

ligurio (approved):
Sergey, once again - thanks for the patches! I left  a number of minor new comments, but overall patchset is LGTM.

perf/lua/1mops_write.lua
```diff
@@ -11,7 +11,30 @@ local math = require('math')
...
+
+local USAGE = [[
```
ligurio, line 16: I propose to split usage for two parts: first part is autogenerated in `benchmark.lua` and contains short usage like `Usage: taskset 0xef tarantool 1mops_write.lua [options]` and a link to recommendations for benchmarking [^1], second part is test-specific, and it is a multiline comment with description of all supported options.


[^1]: https://github.com/tarantool/tarantool/wiki/Benchmarking

mkostoevr (approved):

.github/workflows/perf_micro.yml
```diff
@@ -72,6 +76,25 @@ jobs:
           revision: ${{ inputs.revision }}
       - name: test
         run: make -f .test.mk test-perf
+      - name: Send statistics to InfluxDB
+        # TODO: For now, use the debug bucket for this PoC.
```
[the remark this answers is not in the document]
Buristan: I prefer to use the test bucket until we stabilize bench results to sane values. For now, just use the test bucket for the statistics.

perf/lua/gh-7089-vclock-copy.lua
```diff
@@ -0,0 +1,203 @@
...
+    local dt_realtime = clock.time() - start_realtime
+    local dt_cputime = clock.proc() - start_cputime
+    return dt_realtime, dt_cputime, num_replaces / dt_realtime
+end
+
+local num_iters = 40
+
+-- Fiber count > 1 makes no sense for `wal_mode = 'none'`. There
+-- are no yields on replace when there are no WAL writes.
+local num_fibers = wal_mode == 'none' and 1 or num_iters
```
sergepetrenko, line 182: This was an accident that `num_iters` and `num_fibers` matched. I'd keep `num_fibers` equal to 100 for `wal_mode = 'write'`.

sergepetrenko (commented):
Thanks for working on this!
I have only one comment regarding the `gh-7089-vclock-copy` test.

perf/lua/1mops_write.lua
Buristan: I've added stubs for all places where it is necessary, so 1mops can still be run standalone.

perf/lua/benchmark.lua
```diff
@@ -0,0 +1,130 @@
...
+local function dump_results(bench)
+    dump_report(bench, format_report(bench))
+end
+
+local GENERAL_HELP = [[
+ The supported general options list:
+
+   help (same as -h) <boolean>       - print this message
+   output <string>                   - filename to dump the benchmark results
+   output_format <string, 'console'> - format (console, json) in which results
```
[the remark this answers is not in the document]
Buristan: 'console' here is the default value for this flag.

perf/lua/gh-7089-vclock-copy.lua
Buristan: Hi, Sergey!
Thanks for the review. Replaced back to 100.

.github/workflows/perf_micro.yml
Buristan: Thanks! Renamed to the long name of the option.

.github/workflows/perf_micro.yml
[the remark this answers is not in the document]
ligurio: Okay.

.github/workflows/perf_micro.yml
ligurio: Thanks!

sergepetrenko (approved):
Thanks!
The test `gh-7089-vclock-copy` LGTM.

perf/lua/1mops_write.lua
Buristan: I've moved the autogenerated part to the `benchmark.lua`. Also, I don't mention taskset in the generated part since it may not be the same for different benchmarks. Hence, I've added a link to our wiki with all the useful notes about the setup of the machine for benchmarking.

perf/lua/uri_escape_unescape.lua
Buristan: I moved it to the generated part without mentioning the custom taskset directly.

perf/CMakeLists.txt
Buristan: I've introduced two separated targets -- one depends for the results, and the second runs the command unconditionally.

perf/lua/uri_escape_unescape.lua
ligurio: Thanks!

perf/CMakeLists.txt
```diff
@@ -98,3 +123,15 @@ add_custom_target(test-perf
                   DEPENDS test-c-perf test-lua-perf
                   COMMENT "Running performance tests"
 )
+
+
+set(PERF_SUMMARY ${PERF_OUTPUT_DIR}/summary.txt)
+add_custom_target(test-perf-aggregate
+                  DEPENDS ${BENCH_RESULTS}
+                  BYPRODUCTS ${PERF_SUMMARY}
+                  COMMENT "Aggregate performance test results"
```
ligurio, line 132: I propose to print the path to a final report.

perf/CMakeLists.txt
ligurio: Thanks!

perf/CMakeLists.txt
Buristan: Good idea!
Added to the comment as you suggested.

lastoCHka42 (approved):
LGTM

perf/lua/1mops_write.lua
```diff
@@ -11,11 +11,38 @@ local math = require('math')
 local json = require('json')
 local fio = require('fio')
 
-local params = require('internal.argparse').parse(arg, {
+-- XXX: This benchmark should be able to work standalone without
+-- any provided libraries or set LUA_PATH. Add stubs for that
+-- case.
+local benchmark_is_required, benchmark = pcall(require, 'benchmark')
```
sergos, line 17: s/required/available/g

sergos (commented):
Just naming: required means it is needed, not present.
Otherwise LGTM.

perf/lua/1mops_write.lua
Buristan: Fixed, thanks!

sergos (approved):

PR #10072, opened 2024-05-30
Expose additional info regarding synchronous transaction queue state to `box.info.synchro.queue`

### LINKED ISSUE #9918: Expose additional info regarding synchronous transaction queue state to `box.info`
Add fields
`box.info.synchro.queue.age` - time the oldest entry currently present in the queue has spent waiting for quorum
`box.info.synchro.queue.confirm_lag` - time the latest successfully confirmed entry waited for quorum

CuriousGeorgiy:
This patch series exposes additional `age` and `confirm_lag` field to `box.info.synchro.queue`.

Closes #9918

### DIFF
--- a/changelogs/unreleased/gh-9918-synchro-queue-additional-info.md
+++ b/changelogs/unreleased/gh-9918-synchro-queue-additional-info.md
@@ -1,0 +1,6 @@
+## feature/replication
+
+* Added the `age` and `confirm_lag` fields to `box.info.synchro.queue`: the
+  former shows the time that the oldest entry currently present in the queue has
+  spent waiting for the quorum, while the latter shows the time that the latest
+  successfully confirmed entry waited for the quorum to gather (gh-9918).
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -740,6 +740,8 @@ lbox_info_synchro(struct lua_State *L)
 		lua_pushnumber(L, now - oldest_entry->insertion_time);
 	}
 	lua_setfield(L, -2, "age");
+	lua_pushnumber(L, queue->confirm_lag);
+	lua_setfield(L, -2, "confirm_lag");
 	lua_setfield(L, -2, "queue");
 
 	return 1;
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -731,6 +731,15 @@ lbox_info_synchro(struct lua_State *L)
 	lua_setfield(L, -2, "busy");
 	luaL_pushuint64(L, queue->promote_greatest_term);
 	lua_setfield(L, -2, "term");
+	if (queue->len == 0) {
+		lua_pushnumber(L, 0);
+	} else {
+		struct txn_limbo_entry *oldest_entry =
+			txn_limbo_first_entry(queue);
+		double now = fiber_clock();
+		lua_pushnumber(L, now - oldest_entry->insertion_time);
+	}
+	lua_setfield(L, -2, "age");
 	lua_setfield(L, -2, "queue");
 
 	return 1;
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -59,6 +59,7 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->frozen_reasons = 0;
 	limbo->is_frozen_until_promotion = true;
 	limbo->do_validate = false;
+	limbo->confirm_lag = 0;
 }
 
 static inline bool
@@ -497,6 +498,8 @@ txn_limbo_read_confirm(struct txn_limbo *limbo, int64_t lsn)
 			continue;
 		}
 		e->is_commit = true;
+		if (txn_has_flag(e->txn, TXN_WAIT_ACK))
+			limbo->confirm_lag = fiber_clock() - e->insertion_time;
 		e->txn->limbo_entry = NULL;
 		txn_limbo_remove(limbo, e);
 		txn_clear_flags(e->txn, TXN_WAIT_SYNC | TXN_WAIT_ACK);
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -170,6 +170,7 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
 	e->ack_count = 0;
 	e->is_commit = false;
 	e->is_rollback = false;
+	e->insertion_time = fiber_clock();
 	rlist_add_tail_entry(&limbo->queue, e, in_queue);
 	limbo->len++;
 	return e;
--- a/src/box/txn_limbo.h
+++ b/src/box/txn_limbo.h
@@ -69,6 +69,8 @@ struct txn_limbo_entry {
 	 */
 	bool is_commit;
 	bool is_rollback;
+	/** When this entry was added to the queue. */
+	double insertion_time;
 };
 
 static inline bool
--- a/src/box/txn_limbo.h
+++ b/src/box/txn_limbo.h
@@ -232,6 +232,11 @@ struct txn_limbo {
 	 * can't be any inconsistencies.
 	 */
 	bool do_validate;
+	/**
+	 * The time that the latest successfully confirmed entry waited for
+	 * quorum.
+	 */
+	double confirm_lag;
 };
 
 /**
--- a/test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
+++ b/test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
@@ -28,6 +28,8 @@ g.before_each(function(cg)
         box.ctl.promote()
         local s = box.schema.space.create('s', {is_sync = true})
         s:create_index('pk')
+        local as = box.schema.space.create('as', {is_sync = false})
+        as:create_index('pk')
     end)
     cg.replica_set:wait_for_fullmesh()
     cg.leader:wait_for_downstream_to(cg.replica)
@@ -37,7 +39,7 @@ g.after_each(function(cg)
     cg.replica_set:drop()
 end)
 
--- Test that new `age` field of `box.info.synchro.queue` works correctly.
+-- Test that the new `age` field of `box.info.synchro.queue` works correctly.
 g.test_age_field = function(cg)
     -- The synchronous queue is originally empty.
     cg.leader:exec(function()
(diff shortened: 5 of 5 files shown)


changelogs/unreleased/gh-9918-synchro-queue-additional-info.md
```diff
@@ -0,0 +1,6 @@
+## feature/replication
+
+* Added `age` and `confirm_lag` fields to `box.info.synchro.queue`: the former
+  shows the time that the oldest entry currently present in the queue has spent
+  waiting for quorum, while the latter shows the time that the latest
+  successfully confirmed entry waited for the quorum to gather (gh-9918).
```
andreyaksenov, line 6:
```suggestion
* Added the `age` and `confirm_lag` fields to `box.info.synchro.queue`: the former
  shows the time that the oldest entry currently present in the queue has spent
  waiting for the quorum, while the latter shows the time that the latest
  successfully confirmed entry waited for the quorum to gather (gh-9918).
```

andreyaksenov (approved):

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -0,0 +1,102 @@
...
+
+        local f = fiber.new(function() box.space.s:replace{0} end)
+        f:set_joinable(true)
+        return f:id(), fiber.clock()
+    end)
+    local wait_time = 1
+    fiber.sleep(wait_time)
+
+    cg.leader:exec(function(wait_time)
+        t.assert_almost_equals(box.info.synchro.queue.age, wait_time, 0.15)
```
Serpentian, line 64: I'm afraid we cannot test the time like that. The test is already very flaky and it depends on hardware speed a lot. I propose to sleep for 0.6 seconds and then check, that wait time is > 0.5. After that we may add a new entry to limbo, sleep for 0.6 seconds more and check, that age is more than 1.

src/box/lua/info.c
```diff
@@ -741,6 +741,8 @@ lbox_info_synchro(struct lua_State *L)
 		lua_pushnumber(L, now - oldest_entry->insertion_time);
```
Serpentian, line 741: Nit: in doc request you say, that the feature is from 3.1. But the version was already released and the feature will be only in 3.2. Or we're going to backport it to 3.1 (which I don't like)

src/box/txn_limbo.c
```diff
@@ -59,6 +59,7 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->frozen_reasons = 0;
 	limbo->is_frozen_until_promotion = true;
 	limbo->do_validate = false;
+	limbo->confirm_lag = 0;
```
Serpentian, line 62: Nit: probably the same logic here. Confirm lag is 0, when the last entry was immediately confirmed. Initial value should be probably -1, nil should be pushed to lua table.

We have to discuss the default values for `age` and `confirm_lag` together. @Gerold103 @CuriousGeorgiy what do you think?

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -0,0 +1,102 @@
...
+    -- The synchronous queue is originally empty.
+    cg.leader:exec(function()
+        t.assert_equals(box.info.synchro.queue.age, 0)
+    end)
+    cg.replica:exec(function()
+        t.assert_equals(box.info.synchro.queue.age, 0)
+    end)
+
+    -- Ensure that the synchronous transactions stay in the queue for a while.
+    cg.leader:update_box_cfg{replication_synchro_quorum = 3,
```
Serpentian, line 50: You don't revert box.cfg back (`replication_synchro_timeout`). Let's either put it in `before_all` trigger or properly revert the state of a leader at the end of a test.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -100,3 +100,72 @@ g.test_age_field = function(cg)
...
+    -- Allow the synchronous queue to advance.
+    cg.leader:update_box_cfg{replication_synchro_quorum = ''}
+    cg.leader:exec(function(fid)
+       t.assert(require('fiber').find(fid):join())
+    end, {fid})
+
+    -- The confirm lag must now be shown.
+    local wait_time = fiber.clock() - oldest_entry_ts
+    cg.leader:exec(function(wait_time)
+        t.assert_almost_equals(box.info.synchro.queue.confirm_lag, wait_time,
```
Serpentian, line 145: Maybe just >0?

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -100,3 +100,72 @@ g.test_age_field = function(cg)
...
+
+        local start = fiber.clock()
+        box.space.s:replace{0}
+        return fiber.clock() - start
+    end)
+    t.assert_lt(expected_lag * 10, wait_time)
+
+    -- The confirm lag must be updated by the smaller one.
+    cg.leader:exec(function(expected_lag)
+        t.assert_almost_equals(box.info.synchro.queue.confirm_lag, expected_lag,
```
Serpentian, line 164: Maybe just test, that it's smaller, then it was, when we had `quourum` 3?

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -100,3 +100,72 @@ g.test_age_field = function(cg)
...
+    cg.leader:exec(function()
+        t.assert_equals(box.info.synchro.queue.confirm_lag, 0)
+    end)
+    cg.replica:exec(function()
+        t.assert_equals(box.info.synchro.queue.confirm_lag, 0)
+    end)
+
+    -- Ensure that the synchronous transaction stays in the queue for a while.
+    cg.leader:update_box_cfg{replication_synchro_quorum = 3,
+                             replication_synchro_timeout = 60}
```
Serpentian, line 117: Same here

Serpentian (changes requested):
Thank you for the patch! Looks great! Just some minor comments but we have to decide, which default value we should use, and fix tests)

src/box/txn_limbo.c
CuriousGeorgiy: > Confirm lag is 0, when the last entry was immediately confirmed.

I don't think it will ever be exactly 0.

Overall, I think there is no better default value, than 0. It exists only when the server never did any synchronous replication. An alternative would be to not push the `confirm_lag` field, if we see it is exactly 0.

src/box/lua/info.c
```diff
@@ -731,6 +731,16 @@ lbox_info_synchro(struct lua_State *L)
 	lua_setfield(L, -2, "busy");
 	luaL_pushuint64(L, queue->promote_greatest_term);
 	lua_setfield(L, -2, "term");
+	if (queue->len == 0) {
+		lua_pushnumber(L, 0);
```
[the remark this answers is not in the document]
CuriousGeorgiy: See my comment below about `confirm_lag`.

> IMHO, 0 means the entry was just committed.

It will never be exactly 0, if there is an entry in the limbo. 

The user can also look at `box.synchro.queue.len`. But I think that if the user sees that `age` is zero, he should simply look at the `confirm_lag`.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -0,0 +1,102 @@
...
+    -- Add the first entry to the synchronous queue.
+    local fid1, oldest_entry_ts = cg.leader:exec(function()
+        local fiber = require('fiber')
+
+        local f = fiber.new(function() box.space.s:replace{0} end)
+        f:set_joinable(true)
+        return f:id(), fiber.clock()
+    end)
+    local wait_time = 1
+    fiber.sleep(wait_time)
```
Gerold103, line 61: One second is a very long time for such a trivial test. Please, make it something like 0.1. And Nikita is right about being hardly possible to check the time for comparison. Unfortunately, the time is completely unpredictable. You can only test reliably very basic things - that the age is > the wait time, and it is, say, < 2 minutes.

src/box/txn_limbo.h
```diff
@@ -232,6 +232,11 @@ struct txn_limbo {
 	 * can't be any inconsistencies.
 	 */
 	bool do_validate;
+	/**
+	 * The time that the latest successfully confirmed entry waited for
+	 * quorum.
+	 */
+	double confirm_lag;
```
Gerold103, line 239: Just successfully? Some entries are deleted due to a rollback (even when we delete limbo timeout rollback, the promotes still will be able to rollback txns). The function `txn_limbo_remove()` is used only by successful commit, but its name doesn't suggest that anyhow. I suggest to rename it or even better - place the update right into `txn_limbo_read_confirm()`.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -100,3 +100,72 @@ g.test_age_field = function(cg)
...
+                             replication_synchro_timeout = 60}
+    -- Add an entry to the synchronous queue.
+    local fid, oldest_entry_ts = cg.leader:exec(function()
+        local fiber = require('fiber')
+
+        local f = fiber.new(function() box.space.s:replace{0} end)
+        f:set_joinable(true)
+        return f:id(), fiber.clock()
+    end)
+    fiber.sleep(1)
```
Gerold103, line 126: Same as before about a too long wait time.

Gerold103 (changes requested):
Thanks for the patch!

src/box/lua/info.c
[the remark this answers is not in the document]
Gerold103: I agree with @CuriousGeorgiy here. Better show it as 0 than show it from time to time.

src/box/txn_limbo.c
Gerold103: Same as for the other comment - I think value 0 is fine. Speaking of "immediately confirmed" - it isn't really possible. Confirm would have to happen in the matter of < microsecond which isn't possible. But regardless of that, having a negative lag looks inconvenient - it is not a valid time duration value. People might not expect that.

src/box/lua/info.c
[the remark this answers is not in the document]
Serpentian: Ok, let it be 0. I don't mind

changelogs/unreleased/gh-9918-synchro-queue-additional-info.md
CuriousGeorgiy: Fixed.

src/box/lua/info.c
```diff
@@ -731,6 +731,16 @@ lbox_info_synchro(struct lua_State *L)
...
+	} else {
+		struct txn_limbo_entry *oldest_entry =
+			rlist_first_entry(&queue->queue, typeof(*oldest_entry),
+					  in_queue);
+		double now = ev_monotonic_now(loop());
```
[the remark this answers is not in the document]
CuriousGeorgiy: I tried to be consistent with the surrounding code which uses `ev_monotonic_now(loop())`.

Fixed.

src/box/lua/info.c
CuriousGeorgiy: Fixed.

src/box/lua/info.c
```diff
@@ -731,6 +731,16 @@ lbox_info_synchro(struct lua_State *L)
 	lua_setfield(L, -2, "busy");
 	luaL_pushuint64(L, queue->promote_greatest_term);
 	lua_setfield(L, -2, "term");
+	if (queue->len == 0) {
+		lua_pushnumber(L, 0);
+	} else {
+		struct txn_limbo_entry *oldest_entry =
+			rlist_first_entry(&queue->queue, typeof(*oldest_entry),
+					  in_queue);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed, thanks!

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: Fixed. Testing that the age > the wait time seems to also be unreliable, so I just test it is > 0.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: Fixed, move to `before_all`.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: This doesn't seem to be a reliable test, so I just dropped this hunk.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: Yes, fixed.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: Moved to `before_all`.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
CuriousGeorgiy: Testing that the age > the wait time seems to also be unreliable, so I just test it is > 0 and < 120.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -100,3 +100,72 @@ g.test_age_field = function(cg)
...
+        return fiber.clock() - start
+    end)
+    t.assert_lt(expected_lag * 10, wait_time)
+
+    -- The confirm lag must be updated by the smaller one.
+    cg.leader:exec(function(expected_lag)
+        t.assert_almost_equals(box.info.synchro.queue.confirm_lag, expected_lag,
+                               0.15)
+    end, {expected_lag})
+    cg.replica:exec(function(expected_lag)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Doesn't flake on my machine, hope it won't on yours after my fixes below.

src/box/txn_limbo.h
CuriousGeorgiy: > Just successfully?

Yes, I don't think rollbacks are of interest, and #9918 explicitly mentions successfully confirmed entries.

> I suggest to rename it or even better - place the update right into txn_limbo_read_confirm().

Then I will have to patch two places in `txn_limbo_read_confirm` that remove async and sync transactions (not sure it will be enough to update the confirm lag only for synchronous transactions). So let's rename `txn_limbo_remove` and `txn_limbo_pop` to `txn_limbo_pop_confirmed` and `txn_limbo_pop_rolled_back`.

src/box/txn_limbo.c
```diff
@@ -176,7 +176,8 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
 }
 
 static inline void
-txn_limbo_remove(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
+txn_limbo_pop_confirmed_entry(struct txn_limbo *limbo,
+			      struct txn_limbo_entry *entry)
 {
 	assert(!rlist_empty(&entry->in_queue));
 	assert(txn_limbo_first_entry(limbo) == entry);
```
Gerold103, line 185: Please, add `assert(entry->is_commit)`.

Gerold103 (commented):
Thanks for the fixes, good work, it is nice to see this statistics.

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
Gerold103: Hm. But how is that possible? If you waited for N seconds, you couldn't have committed it faster than that. You would need to sleep less than this time. Could you find the reason please? It could mean a bug.

CuriousGeorgiy: We can test this on the leader (changed the tests from `0` to `wait_time`, where applicable), but we cannot test this on the replica — since the limbo entry arrives on the replica with some delay, and the timekeeping there starts later.

src/box/txn_limbo.c
CuriousGeorgiy: Done.

Serpentian (approved):
Thank you for the patch. No more objections! @Gerold103 can you merge, if everything is all right?

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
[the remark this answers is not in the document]
Serpentian: Yes, the test is absolutely stable now

src/box/txn_limbo.c
```diff
@@ -176,16 +176,22 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
 }
 
 static inline void
-txn_limbo_remove(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
+txn_limbo_pop_confirmed_entry(struct txn_limbo *limbo,
+			      struct txn_limbo_entry *entry, bool is_async)
 {
 	assert(!rlist_empty(&entry->in_queue));
 	assert(txn_limbo_first_entry(limbo) == entry);
+	(void)is_async;
```
Serpentian, line 186: Do we really wanna add a function argument just to pass it to the assertion? I don't like that a lot

src/box/txn_limbo.c
Serpentian: Do we really wanna add a function argument just to pass it to the assertion? I don't like that a lot

Serpentian: Since entry doesn't always has `is_commit` flag, let's not add so sophisticated assertion.

Gerold103 (commented):
Thanks for the fixes! Hm, this assertion revealed a bug. See my comment.

src/box/txn_limbo.c
Gerold103: Yes, the new assertion looks indeed quite overlycomplex, and it reveals a bug. See my comment below.

src/box/txn_limbo.c
```diff
@@ -176,16 +176,22 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
...
+	assert((is_async && entry->txn->signature == TXN_SIGNATURE_UNKNOWN &&
+		entry->txn->status == TXN_PREPARED) || entry->is_commit);
+
```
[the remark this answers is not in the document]
CuriousGeorgiy: > just add the confirm_lag update in a single place, right after `e->is_commit = true;` in `txn_limbo_read_confirm()`.

Actually, we should also check that the transaction was synchronous here (i.e., `txn_has_flag(e->txn, TXN_WAIT_ACK)`). 

Added two checks: (i) for committed asynchronous transactions mixed with synchronous transactions in the limbo; (ii) for prepared asynchronous transactions mixed with synchronous transactions in the limbo.

Gerold103 (approved):
Thanks for addressing my comments! The patch looks better now 💪.

changelogs/unreleased/gh-9918-synchro-queue-additional-info.md
Gerold103: Resolving since the MR is approved.

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: Thanks!

PR #10074, opened 2024-05-30
box: feature `tuple:get_format` to get a format of a tuple

### LINKED ISSUE #10005: Feature `tuple:format` to get a format of a tuple

We need to be able to output tuples in `tt` with their names in a table format. Something like this:

| id  | age  | field  | 
|---|---|---|
| 1  | 1 | 'foo'  |
| 2 |  2 | 'bar'  |

The format could be set per tuple at now:
https://github.com/tarantool/tarantool/pull/8549/commits

I want to be able to get the format object from a tuple with Lua:

```Lua
local format = tuple:format()
```

It will allow me to get mapping field number - field name for the tuple from the format object.

`tuple:tomap()` returns a field name - value, a field number - value mapping. Unfortunately, this does not allow map values by a field name - value in the field number order.

We can add the option to `tuple:totable()`, it looks less obvious from the user point of view.

**Definition of Done**

It is possible to get mapping: field name - field value in the field number order for a tuple.

DerekBum:
This patch adds `tuple:get_format()` method to get a format of a tuple.

Closes #10005


oleg-jukovec:
We need RFC first.

PR #10075, opened 2024-05-30
connpool: move source code file

Totktonada:
The source files for built-in Lua modules are generally placed on the same level: in `src/lua` or in `src/box/lua`, disregarding whether they're public or internal.

The recently introduced `experimental.connpool` built-in module is placed in the `experimental` subdirectory.

This commit moves `src/box/lua/experimental/connpool.lua` to `src/box/lua/connpool.lua` to follow the existing file structure. Public, internal and experimental modules are all on the same level now.

The `connpool` module is still experimental and `require('experimental.connpool')` is needed to use it.

This commit doesn't change the code of the module.


ImeevMA (approved):
Thanks for the patch!

PR #10080, opened 2024-05-31
odict: fix ffi.new('void *') as a key

Totktonada:
The problem is found by @ochaton.


filonenko-mikhail:
the burning fire
the waterfall
the ffi.new('void*')

Totktonada:
Backported to `release/3.1` (future `3.1.1` release) in PR #10110.

PR #10091, opened 2024-06-04
box: update synchro quorum in on_commit trigger instead of on_replace

### LINKED ISSUE #10087: Update quorum in `on_commit` trigger instead of `on_replace` trigger.
For addition of new replicas, let's update the quorum in the `on_commit` trigger. This way we’ll be able to insert a node regardless of the current configuration. This contradicts with the Raft specification, which requires application of all configuration changes in the `on_replace` trigger (i.e., as soon as they are persisted in the WAL, without quorum confirmation).

CuriousGeorgiy:
This patch moves the synchronous replication quorum update from the `on_replace` trigger to the  `on_commit` trigger.

Closes #10087

### DIFF
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4719,6 +4719,11 @@ on_replace_dd_cluster_delete(const struct replica_def *old_def)
 	if (on_commit == NULL)
 		return -1;
 	txn_stmt_on_commit(stmt, on_commit);
+	on_commit = txn_alter_trigger_new(
+		on_replace_cluster_update_quorum, replica);
+	if (on_commit == NULL)
+		return -1;
+	txn_stmt_on_commit(stmt, on_commit);
 	return 0;
 }
 
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4424,6 +4424,15 @@ on_replace_cluster_clear_id(struct trigger *trigger, void * /* event */)
 	return 0;
 }
 
+/** Update the synchronous replication quorum. */
+static int
+on_replace_cluster_update_quorum(struct trigger * /* trigger */,
+				 void * /* event */)
+{
+	box_update_replication_synchro_quorum();
+	return 0;
+}
+
 /** Replica definition. */
 struct replica_def {
 	/** Instance ID. */
@@ -4639,6 +4648,15 @@ on_replace_dd_cluster_insert(const struct replica_def *new_def)
 			 tt_uuid_str(&replica->uuid));
 		return -1;
 	}
+	/*
+	 * Update the quorum only after commit. Otherwise the replica would have
+	 * to ack its own insertion.
+	 */
+	struct trigger *on_commit = txn_alter_trigger_new(
+		on_replace_cluster_update_quorum, NULL);
+	if (on_commit == NULL)
+		return -1;
+	txn_stmt_on_commit(stmt, on_commit);
 	struct trigger *on_rollback = txn_alter_trigger_new(
 		on_replace_cluster_clear_id, NULL);
 	if (on_rollback == NULL)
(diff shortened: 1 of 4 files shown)


andreyaksenov (approved):

src/box/alter.cc
```diff
@@ -4425,6 +4425,15 @@ on_replace_cluster_clear_id(struct trigger *trigger, void * /* event */)
 	return 0;
```
Serpentian, line 4425: Nit: commit msg typo: ~~the~~ during the join process

test/replication-luatest/gh_10087_update_quorum_in_on_commit_trigger_test.lua
```diff
@@ -0,0 +1,88 @@
+local t = require('luatest')
```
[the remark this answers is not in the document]
CuriousGeorgiy: What do you mean? I make the `_cluster` space synchronous in the tests — these tests don't pass on current master.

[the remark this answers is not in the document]
Serpentian: Oh, missed that.

src/box/alter.cc
CuriousGeorgiy: Fixed.

src/box/replication.cc
```diff
@@ -376,7 +376,6 @@ replica_set_id(struct replica *replica, uint32_t replica_id)
 	say_info("assigned id %d to replica %s",
 		 replica->id, tt_uuid_str(&replica->uuid));
 	replica->anon = false;
-	box_update_replication_synchro_quorum();
```
Gerold103, line 379: Please, do the same for `replica_clear_id()`. Otherwise it is confusing, why lowering the quorum via clearing the ID is allowed outside of DDL, but increasing the quorum in `set_id()` is not.

test/replication-luatest/gh_10087_update_quorum_in_on_commit_trigger_test.lua
```diff
@@ -0,0 +1,92 @@
...
+    cg.master:wait_for_downstream_to(cg.replica)
+end)
+
+g.after_test('test_three_member_cluster_insertion', function(cg)
+    cg.replica_set:drop()
+end)
+
+-- Test that synchronous insertion into 3-member cluster with 1 disabled node
+-- works properly.
+g.test_three_member_cluster_insertion = function(cg)
```
Gerold103, line 73: Could you please move the `after_test` after the test? Otherwise it really complicates the reading - I have to jump back above the test's code to see what it does in the end.

Gerold103 (changes requested):
Thanks for the patch! Nicely done, the quorum update by DDL instead of doing it in replication.cc looks more natural.

test/replication-luatest/gh_10087_update_quorum_in_on_commit_trigger_test.lua
Serpentian: And we can also make different test groups just for the sake of code readability. Tests seems to be really independent.

Serpentian (approved):
Thank you for the patch!

test/replication-luatest/gh_10087_update_quorum_in_on_commit_trigger_test.lua
CuriousGeorgiy: > Could you please move the `after_test` after the test?

> And we can also make different test groups just for the sake of code readability. Tests seems to be really independent.

Tests groups consisting of 1 tests sounds like overengineering, I don't think it will really improve the code readibility — let's leave it as is.

Gerold103 (approved):
Thanks for the fixes, good job.

sergepetrenko:
Backports:
2.11 https://github.com/tarantool/tarantool/pull/10229
3.1 https://github.com/tarantool/tarantool/pull/10230

PR #10114, opened 2024-06-07
box: apply deletion of replica from `_cluster` on the deleted replica

### LINKED ISSUE #10088: Apply deletion of replica from `_cluster` on deleted replica
For removal of replicas, let's allow the replica to apply its own deletion, after which it should not try to rejoin. Currently deletion fails with `ER_LOCAL_INSTANCE_ID_IS_READ_ONLY` , which forbids to change the replica identifier. After that the applier on the client side (i.e., the to be removed replica) is stopped due to that error. Instead, we will stop the applier from the `on_commit` trigger. However, if a replica is down, then it will still be impossible to remove it from a two node cluster without manual lowering of the `replication_synchro_quorum` option. There is no way to fix this automatically, so we need to specify in the documentation that in such a scenario the instance should not be turned off before it is dropped from the `_cluster` space when synchronous replication is enabled.

### LINKED ISSUE #10266: Asynchronous transactions from a replica deleted from the cluster arrive on the remaining cluster members
**Steps to reproduce**

```lua
-- Instance 1

box.cfg{
    listen = 3301,
    replication = {3301, 3302},
	read_only = false,
}

-- Step 1.
box.schema.space.create('t'):create_index('p')

-- Step 3.
box.space._cluster:delete{2}

-- Step 6.
assert(box.space.t:get{0} == nil)
```

```lua
-- Instance 2

box.cfg{
    listen = 3302,
    replication = {3301, 3302},
	-- For bootstrap leader certainty.
	read_only = true,
}
box.cfg{read_only = false}

-- Step 2.
assert(box.space.t ~= nil)

-- Step 4.
assert(box.info.replication[1].upstream.status == 'stopped')

-- Step 5.
box.space.t:replace{0}
```

**Actual behavior**

Assertion on step 6 does not fail.

**Expected behavior**

Assertion on step 6 must fail.

CuriousGeorgiy:
This patch series applies deletion of replica from `_cluster` on the deleted replica.

Closes #10088
Closes #10266

### DIFF
--- a/changelogs/unreleased/gh-10088-apply-deletion-of-replica-from_cluster-on-deleted-replica.md
+++ b/changelogs/unreleased/gh-10088-apply-deletion-of-replica-from_cluster-on-deleted-replica.md
@@ -1,0 +1,4 @@
+## feature/replication
+
+* A replica deleted from the `_cluster` space now applies its own deletion and
+  does not try to rejoin (gh-10088).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -61,6 +61,7 @@
 #include "authentication.h"
 #include "node_name.h"
 #include "core/func_adapter.h"
+#include "relay.h"
 
 /* {{{ Auxiliary functions and methods. */
 
@@ -4434,6 +4435,8 @@ on_replace_cluster_clear_id(struct trigger *trigger, void *event)
 				if (replica->applier != NULL)
 					applier_kill(replica->applier,
 						     diag_last_error(diag));
+				if (replica->relay != NULL)
+					relay_cancel(replica->relay);
 			}
 			diag_clear(diag);
 		}
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4418,9 +4418,27 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
 
 /** Unregister the replica affected by the change. */
 static int
-on_replace_cluster_clear_id(struct trigger *trigger, void * /* event */)
+on_replace_cluster_clear_id(struct trigger *trigger, void *event)
 {
-	replica_clear_id((struct replica *)trigger->data);
+	struct replica *replica = (struct replica *)trigger->data;
+	struct txn_stmt *stmt = (struct txn_stmt *)event;
+	(void)stmt;
+	if (replica->id == instance_id) {
+		assert(stmt->row->replica_id != 0 ||
+		       recovery_state != FINISHED_RECOVERY);
+		if (recovery_state == FINISHED_RECOVERY) {
+			diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
+				 (unsigned)instance_id);
+			struct diag *diag = diag_get();
+			replicaset_foreach(replica) {
+				if (replica->applier != NULL)
+					applier_kill(replica->applier,
+						     diag_last_error(diag));
+			}
+			diag_clear(diag);
+		}
+	}
+	replica_clear_id(replica);
 	return 0;
 }
 
@@ -4680,18 +4698,18 @@ on_replace_dd_cluster_insert(const struct replica_def *new_def)
 static int
 on_replace_dd_cluster_delete(const struct replica_def *old_def)
 {
+	struct txn_stmt *stmt = txn_current_stmt(in_txn());
 	/*
-	 * It's okay to delete the instance id while it is joining to a cluster
-	 * as long as the id is set by the time bootstrap is complete. Deletion
-	 * can be coming from the master when the latter tried to purge dead
-	 * replicas from _cluster.
+	 * It's okay to delete the instance id when the deletion is coming from
+	 * the master or doing recovery (i.e., after we already applied the
+	 * deletion from the master).
 	 */
-	if (!replicaset.is_joining && old_def->id == instance_id) {
+	if (old_def->id == instance_id && stmt->row->replica_id == 0 &&
+	    recovery_state == FINISHED_RECOVERY) {
 		diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
 			 (unsigned)old_def->id);
 		return -1;
 	}
-	struct txn_stmt *stmt = txn_current_stmt(in_txn());
 	struct replica *replica = replica_by_id(old_def->id);
 	if (replica == NULL) {
 		/*
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1788,12 +1788,8 @@ applier_on_rollback(struct trigger *trigger, void *event)
 	(void) event;
 	struct applier *applier = (struct applier *)trigger->data;
 	/* Setup a shared error. */
-	if (!diag_is_empty(&replicaset.applier.diag)) {
-		diag_set_error(&applier->diag,
-			       diag_last_error(&replicaset.applier.diag));
-	}
-	/* Stop the applier fiber. */
-	fiber_cancel(applier->fiber);
+	applier_kill(applier, diag_is_empty(&replicaset.applier.diag) ? NULL :
+			      diag_last_error(&replicaset.applier.diag));
 	return 0;
 }
 
@@ -2751,6 +2747,15 @@ applier_stop(struct applier *applier)
 	ERROR_INJECT_YIELD(ERRINJ_APPLIER_STOP_DELAY);
 	applier_set_state(applier, APPLIER_OFF);
 	applier->fiber = NULL;
+}
+
+void
+applier_kill(struct applier *applier, struct error *e)
+{
+	assert(applier->fiber != NULL);
+	if (e != NULL)
+		applier_set_last_error(applier, e);
+	fiber_cancel(applier->fiber);
 }
 
 struct applier *
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -1652,6 +1652,10 @@ applier_apply_tx(struct applier *applier, struct stailq *rows)
 	struct latch *latch = (replica ? &replica->order_latch :
 			       &replicaset.applier.order_latch);
 	latch_lock(latch);
+	if (fiber_is_cancelled()) {
+		rc = -1;
+		goto finish;
+	}
 	if (vclock_get(&replicaset.applier.vclock,
 		       last_row->replica_id) >= last_row->lsn) {
 		goto finish;
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4697,8 +4697,12 @@ box_process_subscribe(struct iostream *io, const struct xrow_header *header)
 	/* Check replica uuid */
 	struct replica *replica = replica_by_uuid(&req.instance_uuid);
 
-	if (!req.is_anon &&
-	    (replica == NULL || replica->id == REPLICA_ID_NIL)) {
+	if (req.is_anon && replica != NULL) {
+		if (replica->id != REPLICA_ID_NIL)
+			tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe "
+				  "an anonymous replica having an ID assigned");
+	} else if (!req.is_anon &&
+		   (replica == NULL || replica->id == REPLICA_ID_NIL)) {
 		/*
 		 * The instance is not anonymous, and is registered (at least it
 		 * claims so), but its ID is not delivered to the current
@@ -4712,10 +4716,6 @@ box_process_subscribe(struct iostream *io, const struct xrow_header *header)
 		 */
 		tnt_raise(ClientError, ER_TOO_EARLY_SUBSCRIBE,
 			  tt_uuid_str(&req.instance_uuid));
-	}
-	if (req.is_anon && replica != NULL && replica->id != REPLICA_ID_NIL) {
-		tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe an "
-			  "anonymous replica having an ID assigned");
 	}
 	/*
 	 * Replica name mismatch is not considered a critical error. It can
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4701,6 +4701,10 @@ box_process_subscribe(struct iostream *io, const struct xrow_header *header)
 		if (replica->id != REPLICA_ID_NIL)
 			tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe "
 				  "an anonymous replica having an ID assigned");
+		if (!replica->anon)
+			tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe "
+				  "a previously deleted non-anonymous replica "
+				  "as an anonymous replica");
 	} else if (!req.is_anon &&
 		   (replica == NULL || replica->id == REPLICA_ID_NIL)) {
 		/*
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -416,7 +416,6 @@ replica_clear_id(struct replica *replica)
 	--replicaset.registered_count;
 	gc_delay_unref();
 	if (replica->id == instance_id) {
-		assert(replicaset.is_joining);
 		instance_id = REPLICA_ID_NIL;
 		box_broadcast_id();
 	} else {
--- a/src/box/replication.cc
+++ b/src/box/replication.cc
@@ -665,12 +665,6 @@ replica_on_applier_state_f(struct trigger *trigger, void *event)
 			struct replica, on_applier_state);
 	replica_update_applier_health(replica);
 	switch (replica->applier->state) {
-	case APPLIER_WAIT_SNAPSHOT:
-		replicaset.is_joining = true;
-		break;
-	case APPLIER_JOINED:
-		replicaset.is_joining = false;
-		break;
 	case APPLIER_CONNECTED:
 		try {
 			if (tt_uuid_is_nil(&replica->uuid))
--- a/test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
+++ b/test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
@@ -1,0 +1,174 @@
+local fio = require('fio')
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g_two_member_cluster = t.group('two_member_cluster')
+local g_three_member_cluster = t.group('three_member_cluster')
+
+g_two_member_cluster.before_each(function(cg)
+    cg.master = server:new{alias = 'master', box_cfg = {
+        replication_synchro_timeout = 120,
+    }}
+    cg.master:start()
+    cg.replica = server:new{alias = 'replica', box_cfg = {
+        replication = cg.master.net_box_uri,
+        replication_synchro_timeout = 120,
+    }}
+    cg.replica:start()
+    -- Add the replica to the master's replicaset table: needed for testing
+    -- the behaviour of the deleted replica after recovery.
+    cg.master:update_box_cfg{replication = cg.replica.net_box_uri}
+    cg.master:exec(function()
+        t.helpers.retrying({timeout = 120}, function()
+            t.assert_not_equals(box.info.status, "orphan")
+        end)
+    end)
+    -- Make `_cluster` space synchronous.
+    cg.master:exec(function()
+        box.ctl.promote()
+        box.space._cluster:alter{is_sync = true}
+        box.schema.space.create('sync', {is_sync = true}):create_index('pk')
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+end)
+
+-- Test that synchronous deletion from 2-member cluster works properly.
+-- Also test the behaviour of the deleted replica after recovery.
+g_two_member_cluster.test_deletion = function(cg)
+    local replica_id = cg.replica:get_instance_id()
+    local log_file = cg.master:exec(function() return box.cfg.log end)
+    fio.truncate(log_file)
+    cg.master:exec(function(replica_id)
+        t.assert(box.space._cluster:delete{replica_id})
+    end, {replica_id})
+    cg.replica:wait_for_vclock_of(cg.master)
+    cg.replica:exec(function()
+        t.assert_equals(box.info.id, nil)
+    end)
+    t.assert_equals(cg.master:grep_log('joining replica'), nil)
+    -- Test the behaviour of the deleted replica after recovery.
+    cg.replica:restart(nil, {wait_until_ready = false})
+    local panic_msg = "'replication_anon' cfg didn't work"
+    local log = fio.pathjoin(cg.replica.workdir, cg.replica.alias..'.log')
+    -- By default, the deleted replica must panic after recovery, because it is
+    -- bootstrapped, it does not have an instance ID, and it is not configured
+    -- as anonymous.
+    t.helpers.retrying({timeout = 120}, function()
+        t.assert(cg.replica:grep_log(panic_msg, nil, {filename = log}))
+    end)
+    cg.replica:restart({box_cfg = {
+        replication = cg.master.net_box_uri,
+        replication_anon = true,
+        read_only = true,
+    }}, {wait_until_ready = false})
+    local err_msg = "Can't subscribe a previously deleted non%-anonymous " ..
+                    "replica as an anonymous replica"
+    -- If the deleted replica is configured as anonymous during recovery, the
+    -- master sees it as a non-anonymous replica in his replicaset table, and
+    -- rejects the subscription request.
+    t.helpers.retrying({timeout = 120}, function()
+        t.assert(cg.replica:grep_log(err_msg, nil, {filename = log}))
+    end)
+end
+
+g_two_member_cluster.after_each(function(cg)
+    cg.master:drop()
+    cg.replica:drop()
+end)
+
+g_three_member_cluster.before_each(function(cg)
+    cg.replica_set = replica_set:new{}
+    cg.master = cg.replica_set:build_and_add_server{alias = 'master',
+                                                    box_cfg = {
+        replication_synchro_timeout = 120,
+    }}
+    cg.master:start()
+    cg.replica =
+        cg.replica_set:build_and_add_server{alias = 'replica',
+                                            box_cfg = {
+        replication = {
+            cg.master.net_box_uri,
+            server.build_listen_uri('to_be_deleted', cg.replica_set.id),
+        },
+    }}
+    cg.replica_to_be_deleted =
+        cg.replica_set:build_and_add_server{alias = 'to_be_deleted',
+                                            box_cfg = {
+        replication = {
+            cg.master.net_box_uri,
+            server.build_listen_uri('replica', cg.replica_set.id),
+        },
+    }}
+    cg.replica_set:start()
+    cg.master:exec(function()
+        box.ctl.promote()
+        box.space._cluster:alter{is_sync = true}
+        box.schema.space.create('sync', {is_sync = true}):create_index('pk')
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+    cg.master:wait_for_downstream_to(cg.replica_to_be_deleted)
+end)
+
+-- Test that synchronous deletion from 3-member cluster with 1 disabled node and
+-- the deleted node alive works properly.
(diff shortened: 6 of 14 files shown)


lenkis (approved):

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
```diff
@@ -0,0 +1,126 @@
...
+
+g_one_member_cluster.before_all(function(cg)
+    cg.master = server:new{alias = 'master'}
+    cg.master:start()
+    cg.replica = server:new{alias = 'replica', box_cfg = {
+        replication = cg.master.net_box_uri,
+    }}
+    cg.replica:start()
+    -- Add the replica to the master's replicaset table: needed for testing
+    -- behaviour or deleted replica after recovery.
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
```diff
@@ -0,0 +1,126 @@
...
+    cg.master:exec(function()
+        box.ctl.promote()
+        box.space._cluster:alter{is_sync = true}
+    end)
+    cg.master:wait_for_downstream_to(cg.replica_to_be_disabled)
+    cg.master:wait_for_downstream_to(cg.replica_to_be_deleted)
+end)
+
+-- Test that synchronous deletion from 3-member cluster with 1 disabled node and
+-- (the deleted node alive) works properly.
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

changelogs/unreleased/gh-10088-apply-deletion-of-replica-from_cluster-on-deleted-replica.md
```diff
@@ -0,0 +1,4 @@
+## feature/replication
+
+* The deletion of a replica from the `_cluster` space is now applied by the
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

test/replication/prune.result
```diff
@@ -137,37 +137,28 @@ test_run:cmd('stop server replica1')
 ---
```
Gerold103, line 137: If your test covers this old one completely, we could delete it, I think. Instead of fixing.

Gerold103 (changes requested):
Thanks for the patch! Good planing, I like how you managed to split the feature into independent small PRs.

test/replication/prune.result
CuriousGeorgiy: Yeah, let's delete it.

src/box/replication.cc
```diff
@@ -418,7 +418,6 @@ replica_clear_id(struct replica *replica)
 	--replicaset.registered_count;
 	gc_delay_unref();
 	if (replica->id == instance_id) {
-		assert(replicaset.is_joining);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Deleted.

src/box/alter.cc
```diff
@@ -4418,9 +4418,24 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
...
+		assert(stmt->row->replica_id != 0 ||
+		       recovery_state != FINISHED_RECOVERY);
+		struct replica *origin = replica_by_id(stmt->row->replica_id);
+		if (recovery_state == FINISHED_RECOVERY && origin != NULL) {
+			assert(origin->applier != NULL &&
+			       origin->applier->fiber != NULL);
+			diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
+				 (unsigned)instance_id);
+			diag_move(diag_get(), &origin->applier->diag);
+			fiber_cancel(origin->applier->fiber);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Done.

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
```diff
@@ -0,0 +1,129 @@
...
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+end)
+
+-- Test that synchronous deletion from 2-member cluster works properly.
+-- Also test the behaviour of the deleted replica after recovery.
+g_one_member_cluster.test_deletion = function(cg)
+    local replica_id = cg.replica:exec(function()
+        return box.info.id
+    end)
```
Gerold103, line 38: There is method `cg.replica:get_instance_id()`.

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
```diff
@@ -0,0 +1,129 @@
...
+    cg.master:update_box_cfg{replication = cg.replica.net_box_uri}
+    cg.master:exec(function()
+        t.helpers.retrying({timeout = 120}, function()
+            t.assert_not_equals(box.info.status, "orphan")
+        end)
+    end)
+    -- Make `_cluster` space synchronous.
+    cg.master:exec(function()
+        box.ctl.promote()
+        box.space._cluster:alter{is_sync = true}
```
Gerold103, line 28: Are you sure `box.ctl.promote()` after completion guarantees the instance is writable? I think it might be still read-only for short time after that. Is the test stable if you run it 50 times in parallel 16 workers, for example?

Gerold103 (commented):
Thanks for the patch! Cool, the synchro system spaces are moving quick.

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
CuriousGeorgiy: The instances are read-write by default (I don't use elections), and AFAIC, the instance owns the limbo after `box.ctl.promote` returns, so there is a guarantee it can add entries to it, i.e., it is synchro-writable.

AFAIC, the test is stable.

test/replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua
CuriousGeorgiy: Fixed.

Gerold103 (approved):
Thanks for the fixes! All good now.

src/box/applier.cc
```diff
@@ -2775,6 +2771,15 @@ applier_stop(struct applier *applier)
 	applier->fiber = NULL;
 }
 
+void
+applier_kill(struct applier *applier, struct error *e)
+{
+	assert(applier->fiber != NULL);
+	if (e != NULL)
+		applier_set_last_error(applier, e);
+	fiber_cancel(applier->fiber);
```
Serpentian, line 2780: Shouldn't the function nullify `applier->fiber`? So that it's impossible to call `applier_kill` several times in a row.

src/box/alter.cc
```diff
@@ -4418,9 +4418,25 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
...
+{
+	struct replica *replica = (struct replica *)trigger->data;
+	struct txn_stmt *stmt = (struct txn_stmt *)event;
+	if (replica->id == instance_id) {
+		assert(stmt->row->replica_id != 0 ||
+		       recovery_state != FINISHED_RECOVERY);
+		struct replica *origin = replica_by_id(stmt->row->replica_id);
+		if (recovery_state == FINISHED_RECOVERY && origin != NULL) {
+			assert(origin->applier != NULL &&
+			       origin->applier->fiber != NULL);
```
Serpentian, line 4431: Nit: we already assert, that `fiber` is not `NULL` in `applier_kill`. Seems redundant

Serpentian (approved):
Thank you for the patch. No objections)

src/box/alter.cc
CuriousGeorgiy: Fixed.

sergepetrenko:
Sorry, didn't have time today. Will take a closer look and merge tomorrow.

src/box/alter.cc
```diff
@@ -4418,9 +4418,24 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
...
+	if (replica->id == instance_id) {
+		assert(stmt->row->replica_id != 0 ||
+		       recovery_state != FINISHED_RECOVERY);
+		struct replica *origin = replica_by_id(stmt->row->replica_id);
+		if (recovery_state == FINISHED_RECOVERY && origin != NULL) {
+			assert(origin->applier != NULL);
+			diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
+				 (unsigned)instance_id);
+			struct diag *diag = diag_get();
+			applier_kill(origin->applier, diag_last_error(diag));
```
sergepetrenko, line 4434: I think you should kill all appliers this instance has, not only the one which delivered the cluster deletion message.
Otherwise having 3 nodes in a full mesh and deleting one of them, you'll keep another connection alive, and the deleted node will continue to receive updates.

Moreover, I conducted some tests, and it seems the node which didn't participate in deletion continues to consider the deleted node as a normal one, so this also has to be fixed. Example:
```lua
-- Instance 1.
-- Step 1.
box.cfg{listen=3301, replication={3301,3302,3303}}
-- Step 2. For all the following steps here use the instance which boots first (most likely instance 1, but not 100%).
box.schema.user.grant("guest", "replication")

-- Step 3. Wait for all nodes to connect.
box.space._cluster:delete{2}

-- Instance 2.
-- Step 2.
box.cfg{listen=3302, replication={3301,3302,3303}}

-- Step 6. Try to insert smth.
box.space._schema:replace{"asdf"}

-- Step 8. Even synchro with quorum 3 succeeds.
box.ctl.promote()
box.cfg{replication_synchro_quorum=3}
box.schema.space.create('test', {is_sync=true}):create_index('pk')
box.space.test:insert{123}

-- Instance 3.
-- Step 3.
box.cfg{listen=3303, replication={3301,3302,3303}}

-- Step 4. Find the node which got deleted (in my  case it was instance 3).
box.info.id
---
- null
...
-- Step 5. Check replication: it's not fully broken.
box.info.replication
---
- 1:
    id: 1
    uuid: 3779b774-e64c-4c83-a23a-593aa9a4a0ae
    lsn: 4
    upstream:
      peer: '3301'
      lag: 0.00023484230041504
      status: stopped
      idle: 18.927021
      message: The local instance id 2 is read-only
    name: null
    downstream:
      status: follow
      idle: 0.31105300000036
      vclock: {1: 4}
      lag: 0
  3:
    id: 3
    uuid: d72536ad-39e7-4c81-9ea9-f6688f619ad3
    lsn: 0
    upstream:
      status: follow
      idle: 0.85617600000023
      peer: '3302'
      lag: 0.0002131462097168
    name: null
    downstream:
      status: follow
      idle: 0.86647700000049
      vclock: {1: 4}
      lag: 0
...

-- Step 7. See that replication continues to work on the deleted instance:
box.space._schema:fselect{}
```

Also after deletion of `instance 3` (which had id 2), `instance 2` still sees `instance 3` as a normal cluster member, even though the `_cluster` tells otherwise:
```
-- Instance 2.
2024-06-14 10:23:11.732 [39583] main/113/applier/3303 I> RAFT: message {term: 2, state: follower} from 2
2024-06-14 10:23:11.732 [39583] main/111/applier/3301 I> RAFT: message {term: 2, state: follower} from 1
tarantool> box.cfg{replication_synchro_quorum=3}
2024-06-14 10:23:21.482 [39583] main/104/interactive/box.load_cfg I> set 'replication_synchro_quorum' configuration option to 3
---
...

tarantool> box.space._cluster:fselect{}
---
- |-
  +-----+--------------------------------------+-----+
  | id  |                 uuid                 |name |
  +-----+--------------------------------------+-----+
  |  1  |"3779b774-e64c-4c83-a23a-593aa9a4a0ae"|     |
  |  3  |"d72536ad-39e7-4c81-9ea9-f6688f619ad3"|     |
  +-----+--------------------------------------+-----+
...

```

sergepetrenko (changes requested):
Hi! Thanks for the patch!

I've found some issues with how other nodes process the deletion. Please, take a closer look at this

src/box/alter.cc
CuriousGeorgiy: Thanks for the catch! 

> Otherwise having 3 nodes in a full mesh and deleting one of them, you'll keep another connection alive, and the deleted node will continue to receive updates.

AFAIC, the reason this didn't happen previously, is that the deleted replica consequently tried to apply the failing deletion from all other nodes, and hence all the appliers got stopped.

Since the deletion now succeeds, it is only applied from one node. So we indeed need to stop all the appliers explicitly. This change helps stop all of the appliers, effectively stopping replication, fixing the synchronous replication issue:

```diff
diff --git a/src/box/alter.cc b/src/box/alter.cc
index 82204347b7..b6c4db59cc 100644
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4422,16 +4422,19 @@ on_replace_cluster_clear_id(struct trigger *trigger, void *event)
 {
        struct replica *replica = (struct replica *)trigger->data;
        struct txn_stmt *stmt = (struct txn_stmt *)event;
+       (void)stmt;
        if (replica->id == instance_id) {
                assert(stmt->row->replica_id != 0 ||
                       recovery_state != FINISHED_RECOVERY);
-               struct replica *origin = replica_by_id(stmt->row->replica_id);
-               if (recovery_state == FINISHED_RECOVERY && origin != NULL) {
-                       assert(origin->applier != NULL);
+               if (recovery_state == FINISHED_RECOVERY) {
                        diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
                                 (unsigned)instance_id);
                        struct diag *diag = diag_get();
-                       applier_kill(origin->applier, diag_last_error(diag));
+                       replicaset_foreach(replica) {
+                               if (replica->applier != NULL)
+                                       applier_kill(replica->applier,
+                                                    diag_last_error(diag));
+                       }
                        diag_clear(diag);
                }
        }
```

> Also after deletion of instance 3 (which had id 2), instance 2 still sees instance 3 as a normal cluster member, even though the `_cluster` tells otherwise:

This is reproducible on the current master branch, so if this is an issue, I don't think we should be fixing in it the scope of this patch series.

sergepetrenko: > AFAIC, the reason this didn't happen previously, is that the deleted replica consequently tried to apply the failing deletion from all other nodes, and hence all the appliers got stopped.

Yes, indeed.

> This is reproducible on the current master branch, i.e., I didn't introduce this behavior in this patch series, so if this is an issue, I don't think we should be fixing it in the scope of this patch series.

Are you sure it's reproducible on the current master? I don't see how you could reproduce this on master if the deleted node voluntarily drops all connections to any remote peer which sends it a deletion.

The issue with your patch applied is that instance 2 already knows that instance 3 is dropped from cluster, but still counts in an ack from instance 3. Note, dropping connections from the instance 3 side won't help with this one, generally.

AFAIU the following sequence is possible:
1. Instance 1 applies the deletion of instance 3
2. Instance 2 applies the deletion of instance 3
3. Either of them sends the deletion to instance 3
4. Instance 3 starts applying a deletion, and applies a couple extra txns after it (they all go to WAL together, so the _cluster delete trigger isn't fired yet, and the connections are alive).
5. Instance 3 acks a couple of extra txns to instances 1 and 2. And instances 1/2 count instance's 3 vote/ack in. Although they shouldn't.

CuriousGeorgiy: > Are you sure it's reproducible on the current master? I don't see how you could reproduce this on master if the deleted node voluntarily drops all connections to any remote peer which sends it a deletion.

Under `instance 2 still sees instance 3 as a normal cluster member, even though the _cluster tells otherwise` I meant the Raft messages:

```console
2024-06-14 10:23:11.732 [39583] main/113/applier/3303 I> RAFT: message {term: 2, state: follower} from 2
```

And yes, you can also see them on current master branch. I can check again tomorrow. Didn’t check, whether this can affect elections. AFAIC, they come from the applier on the master node.

> The issue with your patch applied is that instance 2 already knows that instance 3 is dropped from cluster, but still counts in an ack from instance 3. Note, dropping connections from the instance 3 side won't help with this one, generally.

Okay, I will look into it.

CuriousGeorgiy: > The issue with your patch applied is that instance 2 already knows that instance 3 is dropped from cluster, but still counts in an ack from instance 3. Note, dropping connections from the instance 3 side won't help with this one, generally.

AFAIC, we have a fundamental problem here. We start sending a transaction through the relay before the `on_commit` triggers are run, so an ack from the replica can arrive before we run the `on_commit` triggers. This means that we cannot track whether an ack arrived from a deleted replica: if we try to track this from the `on_replace` trigger, we will also ignore the deletion ack from the deleted replica, making this patch useless.

Serpentian (approved):
Thank you for the patch. No objections

src/box/box.cc
```diff
@@ -4668,6 +4668,10 @@ box_process_subscribe(struct iostream *io, const struct xrow_header *header)
 		if (replica->id != REPLICA_ID_NIL)
 			tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe "
 				  "an anonymous replica having an ID assigned");
+		if (!replica->anon)
+			tnt_raise(ClientError, ER_PROTOCOL, "Can't subscribe "
+				  "a previously deleted non-anonymous replica "
+				  "as an anonymous replica");
```
sergepetrenko, line 4674: Are you sure this check is needed?

Why can't we allow to subscribe the replica as anonymous?

Besides, you need another check for `!req.is_anon`: I just managed to:
1. delete replica from _cluster
2. issue `box.cfg{replication=""}` and `box.cfg{replication="old_replication"}` on it
3. and insert it into the just-deleted slot.

Is this expected?

I think that the replica should be able to subscribe anonymously after the _cluster deletion.

cc @Gerold103, @Serpentian

src/box/alter.cc
sergepetrenko: Thanks for digging into this!
You're right, it's hard to fix this.

I think the correct behavior should be:
1. ack the transaction which contains replica deletion from `_cluster`.
2. Do not ack any further transactions.

Otherwise we might end up in a situation where master commits replica deletion (2 out of 3 nodes agree, but the replica to be deleted hasn't seen the transaction yet), then master is free to add another replica in place of the old one and perform more transactions. Our old replica might receive its own deletion from cluster together with a bunch of new transactions and ack them wrongly. This can lead to a split-brain potentially.

Maybe you can stop acking everything at the moment cluster deletion is received by the replica and send a single special ack just for the transaction containing the deletion from an on_commit trigger?

Another option is to track lsns up to which a particular node has the right to ACK.

I mean, you perform _cluster deletion of a particular replica in a transaction which ends at LSN 8. Then txn_limbo may only count acks for transactions with lsn <= 8.

The second option has a problem that you have to distinguish acks not by replica ids of their senders, but by their UUIDs.

I'm summoning @Gerold103 and @Serpentian here to discuss

CuriousGeorgiy: 3 weeks ago we discussed this with you and @Serpentian and came to a conclusion that this is not a problem, since the deleted replica is part of the cluster up the point where its deletion is confirmed.

src/box/box.cc
CuriousGeorgiy: We cannot subscribe the deleted replica as anonymous, since the master node sees it as non-anonymous in his replicaset table, and AFAIC, transitioning it to anonymous (instead of deleting it from the replicaset table and inserting it as a new anonymous replica) is not easy, since it will break a lot of invariants.

src/box/alter.cc
CuriousGeorgiy: > I think the correct behavior should be:
> 
> 1. ack the transaction which contains replica deletion from _cluster.
> 2. Do not ack any further transactions.

Mustn't we also ack the confirm request?

Serpentian: > Otherwise we might end up in a situation where master commits replica deletion (2 out of 3 nodes agree, but the replica to be deleted hasn't seen the transaction yet), then master is free to add another replica in place of the old one and perform more transactions. Our old replica might receive its own deletion from cluster together with a bunch of new transactions and ack them wrongly. This can lead to a split-brain potentially.

It doesn't seem we may cause split brain with 3 nodes, but what about 4? 

| Phase | Node 1  | Node 2      | Node 3      |Node 4      | Node 5      |
| ------------- | ------------- | ------------- |   ------------- | ------------- |------------- |
| a| 1234 | 1234 |1234 |1234 | - |
| b| 123 | 123 |123 | 1234 | - |
| c| 1235 | 1235 |123 | 1234 | - |
| d| 1235 | 1235 |123 | 1234 | 1235 |

In (a) Node 1 is a leader. In (b) we're deleting node 4 from the cluster. The deleted node doesn't ack its own deletion. In (c) new node 5 is joining to node 1. And this is the place, I don't like: in (c) node 3 may become a leader with vote from node 4. Even if PROMOTE is synchronous (and I hope, it'll be soon), node 3 will be able to commit such promote if it comes in a batch of transaction, where node 4 is deleted. Even though node 3 won't be able to commit any subsequent transactions, it's split brain.

It seems we indeed cannot continue with confirming a bunch of transactions on the node, we're going to delete. Promote shall not pass.

> Actually, I think for this case we can simply track on the master node that an ack arrived from a deleted replica using the on_commit trigger. I.e., mark a replica as deleted in the on_commit trigger and ignore acks from it further on.

Looks valid

src/box/box.cc
Serpentian: > Is this expected?

Also seems valid to me.

> since it will break a lot of invariants.

The question here why it is forbidden to turn a normal replica into an anonymous one. Maybe @sergepetrenko knows the motivation. Moreover, you won't catch the situation of resubscription if replica object is deleted.

src/box/alter.cc
```diff
@@ -4418,9 +4418,27 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
...
+	if (replica->id == instance_id) {
+		assert(stmt->row->replica_id != 0 ||
+		       recovery_state != FINISHED_RECOVERY);
+		if (recovery_state == FINISHED_RECOVERY) {
+			diag_set(ClientError, ER_LOCAL_INSTANCE_ID_IS_READ_ONLY,
+				 (unsigned)instance_id);
+			struct diag *diag = diag_get();
+			replicaset_foreach(replica) {
+				if (replica->applier != NULL)
+					applier_kill(replica->applier,
```
Serpentian, line 4435: Don't we also wanna stop the relay to the deleted replica? So that it becomes orphan and we delete it in `replica_clear_id`?

src/box/alter.cc
```diff
@@ -4418,9 +4418,27 @@ on_replace_dd_schema(struct trigger * /* trigger */, void *event)
...
+						     diag_last_error(diag));
+			}
+			diag_clear(diag);
+		}
+	}
```
Serpentian, line 4440: Don't we also wanna stop the relay on the over nodes to the deleted replica? So that it becomes orphan and we can delete it in `replica_clear_id`?

src/box/alter.cc
Gerold103: This is a very good discussion 💪. I agree with the above - we can't take ACKs from a node which has been deleted from `_cluster`. But I don't know if we need a flag for it. Whether we allow this replica to keep same connection or we kill its applier (and it reconnects right back anyway), in both cases `replica` object of that connection is going to have no `id`, i.e. it will be anon. And we already ignore anon acks.

A potentially very cool important point I saw here from @CuriousGeorgiy - he said *"a transaction ack from the replica can arrive before we run the transaction's on_commit triggers"*. That indeed has been the case since we have the relay, and it has shot our legs numerous times. Just look at the limbo - what kind of squats it is doing to workaround the same problem? We have to process LSN ACKs for txns which are in the limbo but have no LSN.

I think we might want to find a way to solve this problem once and forever. For example, let the relay not send ACKs into TX thread until the latter has at least same vclock or newer as the one represented by the confirmed txns. Then we would be able to safely rely on the code that `on_commit` comes first, replication comes second. It is intuitive and simplifies many places. Or there is another way. I created a ticket here: https://github.com/tarantool/tarantool/issues/10252.

sergepetrenko: > > I think the correct behavior should be:
> > 
> > 1. ack the transaction which contains replica deletion from _cluster.
> > 2. Do not ack any further transactions.
> 
> Mustn't we also ack the confirm request?

@CuriousGeorgiy, for what? To let the leader know we have applied the deletion from _cluster? I'm not sure. Please, elaborate on that.

src/box/alter.cc
sergepetrenko: I'd say we want to stop relays **on** the deleted node to the others.
Because basically it has become "anonymous", and we don't allow to replicate from anonymous nodes.

The relays on the other nodes to this replica will be stopped just because the replica stops its appliers.

src/box/alter.cc
sergepetrenko: > Actually, I think for this case we can simply track on the master node that an ack arrived from a deleted replica using the `on_commit` trigger. I.e., mark a replica as deleted in the `on_commit` trigger and ignore acks from it further on.

Yes, but to make it work we also need to apply @Gerold103's suggestion: make sure `on_commit` on master always happens before any acks for this transaction (#10252).

Long story short, here's the plan:
1. on_commit trigger on master node sets `replica->id = 0`
2. acks from all replicas are processed only after the corresponding on_commit trigger (simply means acks should be processed only up to current instance vclock, but not further #10252). So that after on_commit of replica deletion further acks from it are ignored.
3. on_commit trigger on replica: stops all appliers to non-anonymous(?) nodes, and stops all relays to non-anonymous nodes.

That's it. Also please make sure that as soon as `on_commit` is executed on master or on some other replicaset node, the replica can't resubscribe to this node, and shouldn't issue the rejoin.

src/box/box.cc
sergepetrenko: Sorry, I didn't explain my point good enough.

Basically what happened on steps 2 and 3, the deleted replica registered in the cluster once again.

Anyway, I thought about this some more and I don't see how we can forbid this.

This only happens if the user restarts replication manually, so this seems fine, indeed.

sergepetrenko: > The question here why it is forbidden to turn a normal replica into an anonymous one. Maybe @sergepetrenko knows the motivation. Moreover, you won't catch this situation of resubscription if replica object is deleted.

This simply was never needed.

src/box/alter.cc
Gerold103: Sounds right. I didn't think though that we would go and do #10252 right away, but if we can, then this is very good.

CuriousGeorgiy: > Yes, but to make it work we also need to apply @Gerold103's suggestion: make sure on_commit on master always happens before any acks for this transaction (https://github.com/tarantool/tarantool/issues/10252).

I had the following in mind: if the deleted replica is part of the deletion quorum, then the problem does not exist. If the replica is not part of the deletion quorum, then the `on_commit` trigger on the master node is executed earlier than the ack from the deleted replica is received, so we can simply rely on setting the node as anonymous from the master node in the `on_commit` trigger.

src/box/box.cc
CuriousGeorgiy: After oral discussion: this check is legit.

src/box/alter.cc
CuriousGeorgiy: Added relay stopping in the scope of #10266.

src/box/alter.cc
CuriousGeorgiy: After oral discussion: added a `fiber_is_cancelled` check the in applier code after `latch_lock` in `applier_apply_tx`.

AFAIC, the only thing we could test here is that the replica does not ack a transaction that arrives in one batch with a confirm, however it is quite sophisticated to set this up, so I suggest to rely on the `fiber_cancel` code mentally.

src/box/alter.cc
sergepetrenko: Resolving since the PR is approved.

src/box/alter.cc
sergepetrenko: Thanks! Looks polished, finally!

PR #10131, opened 2024-06-13
test/fuzz: speedup string serialization

ligurio:
- clamp before cleaning string because cleaning is not cheap (O(n), where max n is equal to kMaxStrLength)
- call cleaning for identifiers only, there is no sense to cleaning string literals
- replace symbols disallowed by Lua grammar in indentifier's names with '_'

The patch saves 16 sec on 145k samples (401 sec before the patch and 385 sec after the patch). It is actually not so much, but it is about 2.5 min per hour.

NO_CHANGELOG=testing
NO_DOC=testing


Buristan (approved):
Hi, Sergey!
Thanks for the patch!
LGTM!

PR #9869, opened 2024-03-25
replication: prohibit roll back due to `replication_synchro_timeout`

Astronomax:
To better match the canonical Raft design, this patch prohibits automatic transaction rollback due to `replication_synchro_timeout`. A new compat option has been added for this purpose. The compat option is named `box_cfg_replication_synchro_timeout` and is `'old'` by default. When set to 'new', the `replication_synchro_timeout` option has slightly different semantics. With this semantics, transactions are no longer rolled back at this timeout, `replication_synchro_timeout` is used only to wait confirmation in promote/demote and gc-checkpointing. If some transaction in limbo did not have time to commit within `replication_synchro_timeput`, the corresponding operation: promote/demote or gc-checkpointing can be aborted automatically (in this aspect, the behavior of the option is no different from what it was before). If 'old' is set, the option has the same semantics as before.

In order to be able to understand from the code what value the `box_cfg_replication_synchro_timeout` option is set to - 'old' or 'new', a special Boolean tweak `replication_synchro_timeout_enabled` was introduced.

Note that PROMOTE and DEMOTE can still rollback a transaction. Only the ability to rollback by timeout has been prohibited.

Since with the new behavior synchronous transactions can collect confirmations indefinitely, the queue can grow very large. You need to be able to limit the size of a synchronous queue. Therefore a new configuration option `replication_synchro_queue_max_size` has been added. This option puts a limit on a number of transactions in the master synchronous queue.

Once `txn_limbo.queue` size (acessible in terminal via `box.info.synchro.queue.size`) reaches the maximum value, `txn_limbo` blocks incoming requests until some of the transactions in the queue have a quorum of confirmations and there is free space.

The size limit isn't strict, i.e. if there's at least one free byte, the whole write request fits and no blocking is involved.

Closes #7486

PR #10133, opened 2024-06-14
Enable synchronous replication for systems spaces when the synchronous queue is claimed

### LINKED ISSUE #9723: Make all system spaces synchronous when synchronous replication is enabled
Mixing asynchronous and synchronous replication leads to split-brain when a new synchronous transaction queue owner doesn't have all the async transactions of the former owner. The problem is that system spaces remain asynchronous even when all the user-defined spaces are synchronous.
There is a rather ugly workaround, which requires user intervention:
```lua
box.space._space:alter{is_sync=true}
-- same for all other system spaces.
```
We should make system spaces synchronous by default when synchronous replication is enabled (when synchronous queue is claimed by someone) and test that all system spaces function as expected in this setup.

CuriousGeorgiy:
This patch series enables synchronous replication for systems spaces when the synchronous queue is claimed.

Closes #9723

### DIFF
--- a/changelogs/unreleased/gh-9723-sync-system-spaces.md
+++ b/changelogs/unreleased/gh-9723-sync-system-spaces.md
@@ -1,0 +1,10 @@
+## feature/replication
+
+* Made all of the system spaces synchronous by default when the synchronous
+  queue is claimed, i.e., `box.info.synchro.queue.owner ~= 0`. Added a
+  `box_consider_system_spaces_synchronous` backward compatibility option to
+  control this behavior. Added a new read-only `state` subtable for the Lua
+  space object returned from the `box.space` registry. This subtable has an
+  `is_sync` field that reflects the effective state of replication for system
+  spaces. For user spaces, this field will always mirror the `is_sync` option
+  set for the space (gh-9723).
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2423,6 +2423,12 @@ return schema.new('instance_config', schema.record({
         }, {
             default = 'old',
         }),
+        box_consider_system_spaces_synchronous = schema.enum({
+            'old',
+            'new',
+        }, {
+            default = 'old',
+        }),
     }),
     -- Instance labels.
     labels = schema.map({
--- a/src/box/schema_def.h
+++ b/src/box/schema_def.h
@@ -72,61 +72,64 @@ static_assert(BOX_INVALID_NAME_MAX <= BOX_NAME_MAX,
 
 /**
  * List of system space definitions in the following format:
- * (name, identifier).
+ * (name, identifier, is_sync),
+ * where is_sync determines whether synchronous replication is enabled for this
+ * system space when the synchronous queue is claimed. If is_sync is false, a
+ * reason must be supplied after the space definition comment.
  */
-#define SYSTEM_SPACES(_) /* (name, id) */ \
-	/** Space id of _vinyl_deferred_delete. */ \
-	_(VINYL_DEFERRED_DELETE, 257) \
+#define SYSTEM_SPACES(_) /* (name, id, is_sync) */ \
+	/** Space id of _vinyl_deferred_delete. Local space. */ \
+	_(VINYL_DEFERRED_DELETE, 257, false) \
 	/** Space id of _schema. */ \
-	_(SCHEMA, 272) \
+	_(SCHEMA, 272, true) \
 	/** Space id of _collation. */ \
-	_(COLLATION, 276) \
+	_(COLLATION, 276, true) \
 	/** Space id of _vcollation. */ \
-	_(VCOLLATION, 277) \
+	_(VCOLLATION, 277, true) \
 	/** Space id of _space. */ \
-	_(SPACE, 280) \
+	_(SPACE, 280, true) \
 	/** Space id of _vspace view. */ \
-	_(VSPACE, 281) \
+	_(VSPACE, 281, true) \
 	/** Space id of _sequence. */ \
-	_(SEQUENCE, 284) \
-	/** Space id of _sequence_data. */ \
-	_(SEQUENCE_DATA, 285) \
+	_(SEQUENCE, 284, true) \
+	/** Space id of _sequence_data. Synchronized by space operations. */ \
+	_(SEQUENCE_DATA, 285, false) \
 	/** Space id of _vsequence view. */ \
-	_(VSEQUENCE, 286) \
+	_(VSEQUENCE, 286, true) \
 	/** Space id of _index. */ \
-	_(INDEX, 288) \
+	_(INDEX, 288, true) \
 	/** Space id of _vindex view. */ \
-	_(VINDEX, 289) \
+	_(VINDEX, 289, true) \
 	/** Space id of _func. */ \
-	_(FUNC, 296) \
+	_(FUNC, 296, true) \
 	/** Space id of _vfunc view. */ \
-	_(VFUNC, 297) \
+	_(VFUNC, 297, true) \
 	/** Space id of _user. */ \
-	_(USER, 304) \
+	_(USER, 304, true) \
 	/** Space id of _vuser view. */ \
-	_(VUSER, 305) \
+	_(VUSER, 305, true) \
 	/** Space id of _priv. */ \
-	_(PRIV, 312) \
+	_(PRIV, 312, true) \
 	/** Space id of _vpriv view. */ \
-	_(VPRIV, 313) \
+	_(VPRIV, 313, true) \
 	/** Space id of _cluster. */ \
-	_(CLUSTER, 320) \
+	_(CLUSTER, 320, true) \
 	/** Space id of _trigger. */ \
-	_(TRIGGER, 328) \
+	_(TRIGGER, 328, true) \
 	/** Space id of _truncate. */ \
-	_(TRUNCATE, 330) \
+	_(TRUNCATE, 330, true) \
 	/** Space id of _space_sequence. */ \
-	_(SPACE_SEQUENCE, 340) \
+	_(SPACE_SEQUENCE, 340, true) \
 	/** Space id of _vspace_sequence. */ \
-	_(VSPACE_SEQUENCE, 341) \
+	_(VSPACE_SEQUENCE, 341, true) \
 	/** Space id of _fk_constraint. */ \
-	_(FK_CONSTRAINT, 356) \
+	_(FK_CONSTRAINT, 356, true) \
 	/** Space id of _ck_contraint. */ \
-	_(CK_CONSTRAINT, 364) \
+	_(CK_CONSTRAINT, 364, true) \
 	/** Space id of _func_index. */ \
-	_(FUNC_INDEX, 372) \
+	_(FUNC_INDEX, 372, true) \
 	/** Space id of _session_settings. */ \
-	_(SESSION_SETTINGS, 380) \
+	_(SESSION_SETTINGS, 380, true) \
 
 /** System space identifier definition. */
 #define SYSTEM_SPACE_MEMBER(name, id, ...) BOX_ ## name ## _ID = id,
--- a/src/box/schema_def.h
+++ b/src/box/schema_def.h
@@ -69,61 +69,72 @@ static_assert(BOX_INVALID_NAME_MAX <= BOX_NAME_MAX,
 	      "invalid name max is less than name max");
 
 /** \cond public */
+
+/**
+ * List of system space definitions in the following format:
+ * (name, identifier).
+ */
+#define SYSTEM_SPACES(_) /* (name, id) */ \
+	/** Space id of _vinyl_deferred_delete. */ \
+	_(VINYL_DEFERRED_DELETE, 257) \
+	/** Space id of _schema. */ \
+	_(SCHEMA, 272) \
+	/** Space id of _collation. */ \
+	_(COLLATION, 276) \
+	/** Space id of _vcollation. */ \
+	_(VCOLLATION, 277) \
+	/** Space id of _space. */ \
+	_(SPACE, 280) \
+	/** Space id of _vspace view. */ \
+	_(VSPACE, 281) \
+	/** Space id of _sequence. */ \
+	_(SEQUENCE, 284) \
+	/** Space id of _sequence_data. */ \
+	_(SEQUENCE_DATA, 285) \
+	/** Space id of _vsequence view. */ \
+	_(VSEQUENCE, 286) \
+	/** Space id of _index. */ \
+	_(INDEX, 288) \
+	/** Space id of _vindex view. */ \
+	_(VINDEX, 289) \
+	/** Space id of _func. */ \
+	_(FUNC, 296) \
+	/** Space id of _vfunc view. */ \
+	_(VFUNC, 297) \
+	/** Space id of _user. */ \
+	_(USER, 304) \
+	/** Space id of _vuser view. */ \
+	_(VUSER, 305) \
+	/** Space id of _priv. */ \
+	_(PRIV, 312) \
+	/** Space id of _vpriv view. */ \
+	_(VPRIV, 313) \
+	/** Space id of _cluster. */ \
+	_(CLUSTER, 320) \
+	/** Space id of _trigger. */ \
+	_(TRIGGER, 328) \
+	/** Space id of _truncate. */ \
+	_(TRUNCATE, 330) \
+	/** Space id of _space_sequence. */ \
+	_(SPACE_SEQUENCE, 340) \
+	/** Space id of _vspace_sequence. */ \
+	_(VSPACE_SEQUENCE, 341) \
+	/** Space id of _fk_constraint. */ \
+	_(FK_CONSTRAINT, 356) \
+	/** Space id of _ck_contraint. */ \
+	_(CK_CONSTRAINT, 364) \
+	/** Space id of _func_index. */ \
+	_(FUNC_INDEX, 372) \
+	/** Space id of _session_settings. */ \
+	_(SESSION_SETTINGS, 380) \
+
+/** System space identifier definition. */
+#define SYSTEM_SPACE_MEMBER(name, id, ...) BOX_ ## name ## _ID = id,
+
 enum {
 	/** Start of the reserved range of system spaces. */
 	BOX_SYSTEM_ID_MIN = 256,
-	/** Space if of _vinyl_deferred_delete. */
-	BOX_VINYL_DEFERRED_DELETE_ID = 257,
-	/** Space id of _schema. */
-	BOX_SCHEMA_ID = 272,
-	/** Space id of _collation. */
-	BOX_COLLATION_ID = 276,
-	/** Space id of _vcollation. */
-	BOX_VCOLLATION_ID = 277,
-	/** Space id of _space. */
-	BOX_SPACE_ID = 280,
-	/** Space id of _vspace view. */
-	BOX_VSPACE_ID = 281,
-	/** Space id of _sequence. */
-	BOX_SEQUENCE_ID = 284,
-	/** Space id of _sequence_data. */
-	BOX_SEQUENCE_DATA_ID = 285,
-	/** Space id of _vsequence view. */
-	BOX_VSEQUENCE_ID = 286,
-	/** Space id of _index. */
-	BOX_INDEX_ID = 288,
-	/** Space id of _vindex view. */
-	BOX_VINDEX_ID = 289,
-	/** Space id of _func. */
-	BOX_FUNC_ID = 296,
-	/** Space id of _vfunc view. */
-	BOX_VFUNC_ID = 297,
-	/** Space id of _user. */
-	BOX_USER_ID = 304,
-	/** Space id of _vuser view. */
-	BOX_VUSER_ID = 305,
-	/** Space id of _priv. */
-	BOX_PRIV_ID = 312,
-	/** Space id of _vpriv view. */
-	BOX_VPRIV_ID = 313,
-	/** Space id of _cluster. */
-	BOX_CLUSTER_ID = 320,
-	/** Space id of _trigger. */
-	BOX_TRIGGER_ID = 328,
-	/** Space id of _truncate. */
-	BOX_TRUNCATE_ID = 330,
-	/** Space id of _space_sequence. */
-	BOX_SPACE_SEQUENCE_ID = 340,
-	/** Space id of _vspace_sequence. */
-	BOX_VSPACE_SEQUENCE_ID = 341,
-	/** Space id of _fk_constraint. */
-	BOX_FK_CONSTRAINT_ID = 356,
-	/** Space id of _ck_contraint. */
-	BOX_CK_CONSTRAINT_ID = 364,
-	/** Space id of _func_index. */
-	BOX_FUNC_INDEX_ID = 372,
-	/** Space id of _session_settings. */
-	BOX_SESSION_SETTINGS_ID = 380,
+	SYSTEM_SPACES(SYSTEM_SPACE_MEMBER)
 	/** End of the reserved range of system spaces. */
 	BOX_SYSTEM_ID_MAX = 511,
 	BOX_ID_NIL = 2147483647
--- a/src/box/space.c
+++ b/src/box/space.c
@@ -59,6 +59,39 @@
 #include "lua/utils.h"
 #include "core/mp_ctx.h"
 #include "port.h"
+#include "tweaks.h"
+#include "txn_limbo.h"
+
+/**
+ * Controls whether to consider system spaces indefinitely synchronous when the
+ * synchronous queue is claimed.
+ */
+bool box_consider_system_spaces_synchronous;
+TWEAK_BOOL(box_consider_system_spaces_synchronous);
+
+#define SYNC_SYSTEM_SPACE_MEMBER(name, id, is_sync) \
+	is_sync ? BOX_ ## name ## _ID : BOX_ID_NIL,
+
+/*
+ * System spaces for which synchronous replication is enabled when the
+ * synchronous queue is claimed.
+ */
+static uint32_t sync_system_space_ids[] = {
+	SYSTEM_SPACES(SYNC_SYSTEM_SPACE_MEMBER)
+};
+
+static bool
+system_space_is_sync(struct space *space)
+{
+	for (size_t i = 0; i < lengthof(sync_system_space_ids); ++i) {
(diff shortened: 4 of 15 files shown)


src/box/lua/config/instance_config.lua
```diff
@@ -2420,6 +2420,12 @@ return schema.new('instance_config', schema.record({
         }, {
             default = 'old',
         }),
+        box_consider_system_spaces_synchronous = schema.enum({
+            'old',
+            'new',
+        }, {
+            default = 'old',
```
Serpentian, line 2427: Could you remind me, please, when do we set these compat options to `new` by default?

Serpentian (commented):
Thank you for the patch! Looks great. Left some minor comments

test/replication-luatest/gh_9723_sync_system_spaces_test.lua
```diff
@@ -0,0 +1,430 @@
+local compat = require('compat')
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g = t.group()
+
+local check_sync_system_spaces = [[
```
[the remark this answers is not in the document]
CuriousGeorgiy: To avoid code duplication: I need to call the function on both the master and the replica.

src/box/lua/config/instance_config.lua
CuriousGeorgiy: AFAIC, on major releases.

test/replication-luatest/gh_9723_sync_system_spaces_test.lua
```diff
@@ -0,0 +1,430 @@
...
+    fid = cg.server1:exec(function()
+        box.error.injection.set('ERRINJ_RAFT_WAIT_TERM_PERSISTED_DELAY', true)
+        local f = require('fiber').create(function()
+            box.ctl.demote()
+        end)
+        f:set_joinable(true)
+        return f:id()
+    end)
+    t.helpers.retrying({timeout = 60}, function()
+        t.assert(cg.server2:grep_log('RAFT: persisted state {term: 5}'))
```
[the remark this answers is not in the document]
CuriousGeorgiy: `box.info.election` shows the volatile term, while I need the persisted term, to make sure that the Raft message is processed earlier than the promote message.

src/box/space.h
```diff
@@ -284,6 +284,8 @@ struct space {
 	 * this object in sync. For more information see #9120.
 	 */
 	int lua_ref;
+	/** The effective state of synchronous replication for this space. */
+	bool is_sync;
```
sergepetrenko, line 288: Let's try to find another name for this variable.
In Lua it's good (`state.is_sync`), but here it's too easy to confuse with `space_def::is_sync`.

I can't think of a decent name, so maybe we should hide it into a substruct:
```c
struct {
    bool is_sync;
} state;
```

Then we'll have `space->def->is_sync` (the desired state set by the user) and `space->state.is_sync` (the actual state). It'll also mirror the Lua layout you came up with quite nicely.

src/box/txn_limbo.c
```diff
@@ -1086,6 +1091,84 @@ txn_limbo_filter_request(struct txn_limbo *limbo,
 	}
 }
 
+static uint32_t sync_system_space_ids[] = {
```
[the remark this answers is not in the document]
sergepetrenko: I think we should keep a separate blacklist of the spaces, which shouldn't be sync.
It's quite small as for now, and it's a nice place to keep all the comments on why a specific space is blacklisted.

Adding the "should be sync" field to the enum of all system spaces doesn't look right as for me.

src/box/txn_limbo.c
```diff
@@ -1086,6 +1091,84 @@ txn_limbo_filter_request(struct txn_limbo *limbo,
...
+		space->is_sync = space->def->opts.is_sync ||
+				 (box_consider_system_spaces_synchronous &&
+				  req_type == IPROTO_RAFT_PROMOTE);
+		assert(space->lua_ref != LUA_NOREF);
+		lua_rawgeti(tarantool_L, LUA_REGISTRYINDEX, space->lua_ref);
+		lua_getfield(tarantool_L, -1, "state");
+		lua_pushboolean(tarantool_L, space->is_sync);
+		lua_setfield(tarantool_L, -2, "is_sync");
+		lua_pop(tarantool_L, 2);
+	}
```
sergepetrenko, line 1169: I think the list of spaces (or rather a blacklist) and the code manipulating their is_sync should be in `space.c` (or `schema.c` or somewhere near there). This stuff is internal to the space subsystem.

You could have two methods:
`system_spaces_make_sync`
and
`system_spaces_make_async`.

Limbo would call them depending on the request being processed (`PROMOTE` or `DEMOTE`). And sometimes they would take no action, if `consider_system_spaces_synchronous` is `false`.

Besides, you could call these methods when the user changes `consider_system_spaces_synchronous` on the fly, without promoting or demoting anyone.

src/lua/compat.lua
```diff
@@ -127,6 +127,16 @@ more verbose representation is used.
 https://tarantool.io/compat/box_error_serialize_verbose
 ]]
 
+local BOX_CONSIDER_SYSTEM_SPACES_SYNCHRONOUS = [[
+Controls whether to consider system spaces indefinitely synchronous when the
```
sergepetrenko, line 131: nit: it's not clear what "indefinitely synchronous" means

src/box/txn_limbo.c
```diff
@@ -1086,6 +1091,84 @@ txn_limbo_filter_request(struct txn_limbo *limbo,
...
 
+static uint32_t sync_system_space_ids[] = {
+	/* BOX_VINYL_DEFERRED_DELETE_ID, local space*/
+	BOX_SCHEMA_ID,
+	BOX_COLLATION_ID,
+	BOX_VCOLLATION_ID,
+	BOX_SPACE_ID,
+	BOX_VSPACE_ID,
+	BOX_SEQUENCE_ID,
+	/* BOX_SEQUENCE_DATA_ID, synchronized by synchronous space operations */
```
sergepetrenko, line 1102: Please, mention these two exceptions in the docbot request with a small comment on why they are kept async.

sergepetrenko (changes requested):
Hi! Thanks for the patch! Good job.

changelogs/unreleased/gh-9723-sync-system-spaces.md
```diff
@@ -0,0 +1,9 @@
+## feature/replication
+
+* Made all of the system spaces synchronous by default when the synchronous
+  queue is claimed. A `box_consider_system_spaces_synchronous` backward
+  compatibility option was added to control this behaviour. A new read-only
+  `state` subtable for the Lua space object returned from the `box.space`
+  registry with an `is_sync` field was added to reflect the effective state of
+  replication for system spaces. For user spaces, this field will always mirror
+  the is_sync option set for the space (gh-9723).
```
lenkis, line 9: * Made all of the system spaces synchronous by default when the synchronous
  queue is claimed. Added a `box_consider_system_spaces_synchronous` backward
  compatibility option to control this behavior. Added a new read-only
  `state` subtable for the Lua space object returned from the `box.space`
  registry. This subtable has an `is_sync` field that reflects the effective
  state of replication for system spaces. For user spaces, this field will
  always mirror the `is_sync` option set for the space (gh-9723).

lenkis (approved):

Totktonada (approved):
No objections regarding the declarative config part.

changelogs/unreleased/gh-9723-sync-system-spaces.md
```diff
@@ -0,0 +1,9 @@
+## feature/replication
+
+* Made all of the system spaces synchronous by default when the synchronous
+  queue is claimed. A `box_consider_system_spaces_synchronous` backward
```
Totktonada, line 4: The issue, the documentation request and the changelog entry define a predicate when the system spaces become synchronous this way:

> when the synchronous queue is claimed

It is obscure from a user perspective. I think that the documentation request should shed some light here.

I asked @sergepetrenko and if I understood his words correctly, whether the system spaces are synchronous depends on the following properties:

* The compat option.
* Whether RAFT is in use.
* Whether `box.ctl.promote()` and `box.ctl.demote()` is used and which of them is the last.

And these properties are somehow related to whether it is possible that synchronous transactions are present in the replicaset history. I think that it is important to highlight this relation.

CuriousGeorgiy: I don't think it is obscure. The predicate is simply `box.info.synchro.queue.owner ~= 0`, and it is described in the documentation [^1]:

> `owner` (since version [2.10.0](https://www.tarantool.io/ru/doc/latest/release/2.10.0/)) – ID of the replica that owns the synchronous transaction queue. Once an owner instance appears, all other instances become read-only. If the owner field is 0, then every instance may be writable, but they can’t create any synchronous transactions. To claim or reclaim the queue, use [box.ctl.promote()](https://www.tarantool.io/ru/doc/latest/reference/reference_lua/box_ctl/promote/#box-ctl-promote) on the instance that you want to promote. With elections enabled, an instance runs box.ctl.promote() command automatically after winning the elections. To clear the ownership, call [box.ctl.demote()](https://www.tarantool.io/ru/doc/latest/reference/reference_lua/box_ctl/demote/#box-ctl-demote) on the synchronous queue owner.

[^1]: https://www.tarantool.io/ru/doc/latest/reference/reference_lua/box_info/synchro/

src/box/space.h
CuriousGeorgiy: Fixed.

src/lua/compat.lua
CuriousGeorgiy: Fixed.

src/box/txn_limbo.c
CuriousGeorgiy: Fixed.

changelogs/unreleased/gh-9723-sync-system-spaces.md
CuriousGeorgiy: Fixed.

src/box/txn_limbo.c
```diff
@@ -37,6 +37,11 @@
 #include "raft.h"
 #include "tt_static.h"
 #include "session.h"
+#include "space_cache.h"
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/txn_limbo.c
```diff
@@ -1086,6 +1091,84 @@ txn_limbo_filter_request(struct txn_limbo *limbo,
...
+	       req_type == IPROTO_RAFT_DEMOTE);
+	if (is_rollback)
+		req_type = req_type == IPROTO_RAFT_PROMOTE ?
+			   IPROTO_RAFT_DEMOTE : IPROTO_RAFT_PROMOTE;
+	if (recovery_state <= INITIAL_RECOVERY ||
+	    (!is_rollback &&
+	     ((req_type == IPROTO_RAFT_PROMOTE &&
+	       limbo->owner_id != REPLICA_ID_NIL) ||
+	      (req_type == IPROTO_RAFT_DEMOTE &&
+	       limbo->owner_id == REPLICA_ID_NIL))))
```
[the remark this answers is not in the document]
CuriousGeorgiy: > This code block is very difficult to parse, probably we need comments here and I'd refactor it, so that every OR condition is just a separate if -> return.

Fixed.

> Why don't we apply space transitions only when we recover from the snap file. They're needed during xlog recovery?

Actually, AFAIC, it's pointless to apply space transition during recovery, so let's just leave the condition as `recovery_state != FINISHED_RECOVERY`.

> AFAICS, the second check is PROMOTE from some instance to smth else, as promote cannot be done with zero `origin_id`. Nothing to do. Last one is DEMOTE from nil to nil, also nothing to do. Am I right?

Right.

src/box/txn_limbo.c
CuriousGeorgiy: > I think the list of spaces (or rather a blacklist) and the code manipulating their is_sync should be in space.c (or schema.c or somewhere near there). This stuff is internal to the space subsystem.

Moved to `box/schema.{h, cc}`.

> Besides, you could call these methods when the user changes consider_system_spaces_synchronous on the fly, without promoting or demoting anyone

How should I call them if do not know what state should I transition the system space to? 

> when the user changes consider_system_spaces_synchronous on the fly

I don't think this is a valid use case.

src/box/txn_limbo.c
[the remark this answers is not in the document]
Serpentian: I like X macro you propose. Let's see what @sergepetrenko will say

src/box/txn_limbo.c
Serpentian: > Moved to box/schema.{h, cc}.

@CuriousGeorgiy they're still in `txn_limbo`

CuriousGeorgiy: Sorry, I forgot to push the fixes.

src/lua/compat.lua
```diff
@@ -127,6 +127,16 @@ more verbose representation is used.
 https://tarantool.io/compat/box_error_serialize_verbose
 ]]
 
+local BOX_CONSIDER_SYSTEM_SPACES_SYNCHRONOUS = [[
```
Serpentian, line 130: Shouldn't compat option switching make all system spaces sync if there's already limbo owner? Current behavior is not good, IMHO:

```
tarantool> box.cfg{}
tarantool> box.ctl.promote()
tarantool> box.info.synchro.queue
---
- owner: 1
  confirm_lag: 0
  term: 2
  age: 0
  len: 0
  busy: false
...
tarantool> require('compat').box_consider_system_spaces_synchronous = 'new'
tarantool> box.space._schema.state
---
- is_sync: false
...

tarantool> box.ctl.promote()
tarantool> box.space._schema.state
---
- is_sync: false
...
tarantool> box.ctl.demote()
tarantool> box.ctl.promote()
tarantool> box.space._schema.state
---
- is_sync: true
...

```

Demote must be done in order to enable force sync. Are we ok with that?

Serpentian (commented):
Thank you for the patch! Looks great. Left some comments

src/lua/compat.lua
Serpentian: We can forbid changing it to 'new', when limbo owner is not nil. As we do with deprecation of `ROLLBACK`: https://github.com/tarantool/tarantool/pull/9869#issuecomment-2208711656

src/box/schema_def.h
```diff
@@ -69,61 +69,72 @@ static_assert(BOX_INVALID_NAME_MAX <= BOX_NAME_MAX,
...
+	_(BOX_FK_CONSTRAINT_ID, 356) \
+	/** Space id of _ck_contraint. */ \
+	_(BOX_CK_CONSTRAINT_ID, 364) \
+	/** Space id of _func_index. */ \
+	_(BOX_FUNC_INDEX_ID, 372) \
+	/** Space id of _session_settings. */ \
+	_(BOX_SESSION_SETTINGS_ID, 380) \
+
+/** System space identifier definition. */
+#define SYSTEM_SPACE_ID(name, id, ...) name = id,
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed, thanks!

src/box/space.c
```diff
@@ -59,6 +59,25 @@
...
+#include "txn_limbo.h"
+
+/**
+ * Controls whether to consider system spaces indefinitely synchronous when the
+ * synchronous queue is claimed.
+ */
+bool box_consider_system_spaces_synchronous;
+TWEAK_BOOL(box_consider_system_spaces_synchronous);
+
+#define SYNC_SYSTEM_SPACE_ID(name, id, is_sync) is_sync ? name : BOX_ID_NIL,
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/lua/compat.lua
Serpentian: @sergepetrenko

sergepetrenko: > I don't think this is a valid use case to support and maintain: the user should set the compat option at the instance start through configuration.

This means the user can only test the new behavior after restarting his service. No good. I think we should allow the user to change the option on the fly in both directions: "old" -> "new" and "new" -> "old". And it should take effect immediately. Not after limbo owner change.

> We can forbid changing it to 'new', when limbo owner is not nil. As we do with deprecation of ROLLBACK: https://github.com/tarantool/tarantool/pull/9869#issuecomment-2208711656

Hmm. Again, no. For example, when raft is enabled, the limbo owner is never nil. It only transitions form one owner to another without demotions. And as I said earlier, we wan't to let the user test the new behavior without restarting the service or performing extra disruptive steps (for example, calling box.ctl.demote() on the current raft leader will cause a leader change and unavailability for rw requests for the time of elections)

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: The macros turned out nice!

src/box/txn_limbo.h
```diff
@@ -474,6 +474,10 @@ txn_limbo_fence(struct txn_limbo *limbo);
 void
 txn_limbo_unfence(struct txn_limbo *limbo);
 
+/** Return whether limbo has an owner. */
+bool
+txn_limbo_has_owner(struct txn_limbo *limbo);
```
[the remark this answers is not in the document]
CuriousGeorgiy: This requires adding the `replication.h` header for the `REPLICA_ID_NIL` definition, while the `snap_quorum_delay.cc` test defines its own `instance_id`, which causes a compilation error:
https://github.com/tarantool/tarantool/blob/b0b32bff022caeffd64ca687baab908f25fa2fcb/test/unit/snap_quorum_delay.cc#L76-L82

However, according to 8c50e069:
> Also the commit disables snap_quorum_delay.test, which uses internal API of replication and txn modules assuming the journal is initialized somewhere inside. But now it is not, and it can't be fixed in a sane way inside the test. It will be deleted/rewritten later.

So I guess it safe to simply remove the `instance_id` definition from `snap_quorum_delay.cc`.

src/box/schema_def.h
```diff
@@ -72,61 +72,64 @@ static_assert(BOX_INVALID_NAME_MAX <= BOX_NAME_MAX,
...
+ * (name, identifier, is_sync),
+ * where is_sync determines whether synchronous replication is enabled for this
+ * system space when the synchronous queue is claimed. If is_sync is false, a
+ * reason must be supplied after the space definition comment.
  */
-#define SYSTEM_SPACES(_) /* (name, id) */\
-	/** Space if of _vinyl_deferred_delete. */ \
-	_(VINYL_DEFERRED_DELETE, 257) \
+#define SYSTEM_SPACES(_) /* (name, id, is_sync) */\
+	/** Space if of _vinyl_deferred_delete. Local space. */ \
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/lua/compat.lua
CuriousGeorgiy: Thank you for the feedback! 

Added support for transition of the compat option, and added test cases for it.

src/box/txn_limbo.h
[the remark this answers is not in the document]
sergepetrenko: Oh. So we have a test disabled for 4 years, huh? :)

src/box/txn_limbo.c
```diff
@@ -1086,6 +1086,36 @@ txn_limbo_filter_request(struct txn_limbo *limbo,
...
+			   IPROTO_RAFT_DEMOTE : IPROTO_RAFT_PROMOTE;
+	}
+	/* Synchronous replication is already enabled. */
+	if (!is_rollback && req_type == IPROTO_RAFT_PROMOTE &&
+	    limbo->owner_id != REPLICA_ID_NIL)
+		return;
+	/* Synchronous replication is already disabled. */
+	if (!is_rollback && req_type == IPROTO_RAFT_DEMOTE &&
+	    limbo->owner_id == REPLICA_ID_NIL)
+		return;
```
sergepetrenko, line 1115: This place seems incorrect.
Looks like in case you already have `limbo->owner_id != 0` set by some previous PROMOTE, and now another PROMOTE request is rolled back, you will make system spaces asynchronous. If I'm correct, please cover this with an extra test.

Let's take a look at all the 4 cases. Note that during a rollback `owner_id` is already in the 'old' state, because it's only changed inside `txn_limbo_req_commit()`.
1. Normal PROMOTE: if `owner_id != 0`, do nothing (because PROMOTE always sets a non-zero id, and we already have `owner_id != 0`)
2. Rollback of PROMOTE: if owner_id != 0, do nothing (because we are rolling back a PROMOTE, and during a corresponding `req_prepare` we did nothing. See point 1)
3. Normal DEMOTE: if `owner_id == 0`, do nothing (because DEMOTE always sets a zero id, and we already have `owner_id == 0`)
4. Rollback of DEMOTE: if `owner_id == 0`, do nothing (because we are rolling back a DEMOTE, and during a corresponding `req_prepare` we did nothing. See point 3)

Long story short, I'd rewrite this somewhat along these lines:
```diff
diff --git a/src/box/txn_limbo.c b/src/box/txn_limbo.c
index 61444ae84c..533fe08d0b 100644
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -1100,20 +1100,17 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
        uint16_t req_type = req->type;
        assert(req_type == IPROTO_RAFT_PROMOTE ||
               req_type == IPROTO_RAFT_DEMOTE);
-       if (is_rollback) {
-               /* Reverse the request type. */
-               req_type = req_type == IPROTO_RAFT_PROMOTE ?
-                          IPROTO_RAFT_DEMOTE : IPROTO_RAFT_PROMOTE;
-       }
+       bool is_promote = req_type == IPROTO_RAFT_PROMOTE;
        /* Synchronous replication is already enabled. */
-       if (!is_rollback && req_type == IPROTO_RAFT_PROMOTE &&
-           limbo->owner_id != REPLICA_ID_NIL)
+       if (is_promote && limbo->owner_id != REPLICA_ID_NIL)
                return;
        /* Synchronous replication is already disabled. */
-       if (!is_rollback && req_type == IPROTO_RAFT_DEMOTE &&
-           limbo->owner_id == REPLICA_ID_NIL)
+       if (!is_promote && limbo->owner_id == REPLICA_ID_NIL)
                return;
-       system_spaces_update_is_sync_state(req_type == IPROTO_RAFT_PROMOTE);
+       /* Flip operation types for a rollback. */
+       if (is_rollback)
+               is_promote = !is_promote;
+       system_spaces_update_is_sync_state(is_promote);
 }
 
 int
```

sergepetrenko (commented):
Thanks for the fixes!
I really like the effort you've put into the test. It's comprehensive and well stabilized.

src/box/txn_limbo.c
CuriousGeorgiy: Thanks for the catch!

> Note that during a rollback owner_id is already in the 'old' state, because it's only changed inside txn_limbo_req_commit().

Indeed, this is my misconception, I thought that the state is changed earlier.

> Long story short, I'd rewrite this somewhat along these lines:

Applied the changes you suggested. 

> Rollback of PROMOTE: if owner_id != 0, do nothing (because we are rolling back a PROMOTE, and during a corresponding req_prepare we did nothing. See point 1)

Added a test case for this.

> Rollback of DEMOTE: if owner_id == 0, do nothing (because we are rolling back a DEMOTE, and during a corresponding req_prepare we did nothing. See point 3)

Seems like this case is impossible:
https://github.com/tarantool/tarantool/blob/62c49367a2eab6d8008da2d7201ffa03e002c31e/src/box/box.cc#L3120-L3121

Added an assertion for it.

Serpentian (approved):
Looks polished. Thank you for the patch! No objections.

But I suppose, we should block the PR until https://github.com/tarantool/tarantool/pull/10114 is merged. Otherwise we'll break replicasets, which consist of 2 nodes, which is a common case. We should take our time here.

src/box/space.c
```diff
@@ -621,7 +641,10 @@ space_create(struct space *space, struct engine *engine,
 	rlist_create(&space->memtx_stories);
 	rlist_create(&space->alter_stmts);
 	space->lua_ref = LUA_NOREF;
-	space->state.is_sync = space->def->opts.is_sync;
+	space->state.is_sync = space->def->opts.is_sync ||
+			       (box_consider_system_spaces_synchronous &&
+				space_is_system(space) &&
+				txn_limbo_has_owner(&txn_limbo));
```
sergepetrenko, line 647: In case some async system space is added in the future or if we're upgrading from an old enough schema where some of the async system spaces didn't exist yet this code will make them synchronous.

CuriousGeorgiy: Thanks for the catch! Fixed.

Unfortunately, there was no synchronous replication available in 1.7.6, when the last asynchronous system space was added, so I suggest we just trust this code, since it is pretty simple.

src/box/space.c
```diff
@@ -621,6 +654,10 @@ space_create(struct space *space, struct engine *engine,
 	rlist_create(&space->memtx_stories);
 	rlist_create(&space->alter_stmts);
 	space->lua_ref = LUA_NOREF;
+	space->state.is_sync = space->def->opts.is_sync ||
+			       (box_consider_system_spaces_synchronous &&
+				system_space_is_sync(space) &&
```
sergepetrenko, line 659: nit: let's rewrite as
```c
space_is_system(space) &&
system_space_is_sync(space)
```

sergepetrenko (approved):
Thanks for working on this!

The patch turned out quite nice.
Please, fix the last nit and reassign me to push.

changelogs/unreleased/gh-9723-sync-system-spaces.md
sergepetrenko: I agree with @CuriousGeorgiy here. We have mentioned `box.info.synchro.queue.owner` on multiple occasions in the documentation as being a predicate for some mode of operation.

I also agree that the documentation should be clarified regarding the cases when `synchro.queue.owner` is not nil. I've opened a corresponding doc ticket: https://github.com/tarantool/doc/issues/4400

src/box/space.c
CuriousGeorgiy: Fixed, thanks!

changelogs/unreleased/gh-9723-sync-system-spaces.md
sergepetrenko: Resolving since the PR is approved.

PR #10135, opened 2024-06-14
ci: add workflow to check downgrade versions

### LINKED ISSUE #8319: Provide tool to check that box.schema.downgrade_versions() knows all released Tarantool versions

nshy:
Tarantool has hardcoded list of versions it can downgrade to. This list should consist of all the released versions less than Tarantool version. This workflow helps to make sure we update the list before release.

When run on the commit with release tag it check the list and fail if it misses some released version less than current. In this case we are supposed to update downgrade list (with required downgrade code) and update the release tag.

Closes #8319

### DIFF
--- a/.github/workflows/downgrade_versions.yml
+++ b/.github/workflows/downgrade_versions.yml
@@ -1,0 +1,58 @@
+name: downgrade_versions
+
+on:
+  push:
+    tags:
+      - '**'
+
+concurrency:
+  # Update of a developer branch cancels the previously scheduled workflow
+  # run for this branch. However, the 'master' branch, release branch, and
+  # tag workflow runs are never canceled.
+  #
+  # We use a trick here: define the concurrency group as 'workflow run ID' +
+  # 'workflow run attempt' because it is a unique combination for any run.
+  # So it effectively discards grouping.
+  #
+  # Important: we cannot use `github.sha` as a unique identifier because
+  # pushing a tag may cancel a run that works on a branch push event.
+  group: ${{ (
+    github.ref == 'refs/heads/master' ||
+    startsWith(github.ref, 'refs/heads/release/') ||
+    startsWith(github.ref, 'refs/tags/')) &&
+    format('{0}-{1}', github.run_id, github.run_attempt) ||
+    format('{0}-{1}', github.workflow, github.ref) }}
+  cancel-in-progress: true
+
+jobs:
+  downgrade_versions:
+    if: github.repository == 'tarantool/tarantool'
+
+    runs-on: ubuntu-20.04-self-hosted
+
+    container:
+      image: docker.io/tarantool/testing:ubuntu-jammy
+
+    steps:
+      - name: Prepare checkout
+        uses: tarantool/actions/prepare-checkout@master
+
+      - name: Sources checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          # If workflow is triggered by pushing a tag and we checkout
+          # using github.ref then checkout action overwrite tag with tag
+          # without annotation. This breaks check-downgrade-versions.sh
+          # script as it is triggered only on annotated tags. Let's
+          # checkout by sha in this case.
+          ref: ${{ github.sha }}
+
+      - name: Check downgrade versions
+        run: ./tools/check-downgrade-versions.sh
+
+      - name: Send VK Teams message on failure
+        if: failure()
+        uses: ./.github/actions/report-job-status
+        with:
+          bot-token: ${{ secrets.VKTEAMS_BOT_TOKEN }}
--- a/tools/check-downgrade-versions.sh
+++ b/tools/check-downgrade-versions.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+#
+# This script checks that we can downgrade to all the released versions
+# less then being released. For example when releasing version 2.11.4
+# we check that we can downgrade to 2.11.3, 2.11.2, ..., 2.10.8, 2.10.7, ...
+# 2.8.2.
+#
+# The versions we can downgrade to are those mentioned in downgrade_versions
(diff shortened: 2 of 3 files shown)


tools/check-downgrade-versions.sh
```diff
@@ -0,0 +1,79 @@
...
+
+set -eo pipefail
+
+error() {
+   echo "$@" 1>&2
+   exit 1
+}
+
+# Tags that should not be considered.
+tag_exceptions=`mktemp`
```
locker, line 25: Please remove all temporary files at exit (e.g. with `trap EXIT`).

Anyway, do you really need to use temporary files in this script? Can't you use arrays?

.github/workflows/tree.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Check entrypoint tag
         run: ./tools/check-entrypoint-tag.sh
 
+      - name: Check downgrade versions
+        run: ./tools/check-downgrade-versions.sh
+
```
locker, line 51: AFAIU this workflow won't run on pushing a tag.

tools/check-downgrade-versions.sh
```diff
@@ -0,0 +1,79 @@
...
+    # Cut part of $upgrade_file between $begin_mark and $end_mark
+    # and for every line strip everything except version.
+    awk "/$begin_mark/{flag=1; next}
+         /$end_mark/{flag=0} flag" $upgrade_file |
+         sed -E 's/.*"(.*)".*/\1/' > $actual_versions
+
+    cat $tag_exceptions >> $actual_versions
+    # Sort is usual order before using `comm`.
+    sort -o $expected_versions $expected_versions
+    sort -o $actual_versions $actual_versions
```
locker, line 69: Should be `sort -V`? Also, I don't think it's safe to write to the input file.

tools/check-downgrade-versions.sh
nshy: Done.

```diff
--- a/tools/check-downgrade-versions.sh
+++ b/tools/check-downgrade-versions.sh
@@ -21,6 +21,12 @@ error() {
    exit 1
 }

+cleanup() {
+  rm $tag_exceptions $expected_versions $actual_versions || true
+}
+
+trap "cleanup" EXIT
+
```

Discussed f2f usage of arrays. Looks like it will complicate the script.

tools/check-downgrade-versions.sh
nshy: We don't use `sort -V` because `comm` does not have a corresponding option. By the way fixed the comment to `Sort in usual ...`.

Although it is not stated in my distribution's `man sort` `sort -o` should work for inplace sort [according to POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sort.html):

```
-o  output
Specify the name of an output file to be used instead of the standard output. This file can be the same as one of the input files.
```

tools/check-downgrade-versions.sh
locker: You should use `rm -f`. Also, better add a helper function for creating a temporary file that would add it to the cleanup list.

.github/workflows/tree.yml
nshy: Fixed. We have do some lifting though to keep tag from changing on checkout action.

```diff
diff --git a/.github/workflows/tree.yml b/.github/workflows/tree.yml
index 74d15a77a..0d841a635 100644
--- a/.github/workflows/tree.yml
+++ b/.github/workflows/tree.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - 'master'
       - 'release/**'
+    tags:
+      - '**'

 concurrency:
   # Update of a developer branch cancels the previously scheduled workflow
@@ -42,6 +44,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          # If workflow is triggered by pushing a tag then and we checkout
+          # using github.ref then checkout action overwrite tag with tag
+          # without annotation. This breaks check-downgrade-versions.sh
+          # script as it is triggered only on annotated tags. Let's
+          # checkout by sha in this case.
+          ref: ${{ github.reftype == 'tag' && github.sha || github.ref }}

       - name: Check entrypoint tag
         run: ./tools/check-entrypoint-tag.sh
```

tools/check-downgrade-versions.sh
nshy: Added `-f` flag. Discussed the cleanup suggestion f2f - bash is not very suited for this.

nshy:
Moved checking downgrade versions to a new `downgrade_versions` workflow as discussed f2f.

.github/workflows/downgrade_versions.yml
```diff
@@ -0,0 +1,58 @@
+name: tree
```
locker, line 1: It's probably fine to group `entrypoint-tag` and `downgrade_versions` jobs in the same workflow `tree` but is it possible to spread one workflow among different files in GitHub? If it is possible and you decide to group the two jobs in one workflow, better rename `downgrade_versions` to `downgrade-version` for consistency with `entrypoint-tag`.

nshy: This is just copy-paste mistake. I'd better use `downgrade_versions`.

locker (approved):

locker:
@nshy No need to run `full-ci` on this one, apparently. Just solicit an approval from the devx team and reassign back to me for merge.

foxzi (approved):
OK

locker:

PR #8549, opened 2023-04-09
box: introduce formats for standalone tuples

CuriousGeorgiy:
Closes #4693

PR #10136, opened 2024-06-14
box: feature `tuple:format` to get a format of a tuple

### LINKED ISSUE #10005: Feature `tuple:format` to get a format of a tuple

DerekBum:
This patch adds `tuple:format()` method to get a format of a tuple.

Closes #10005

@TarantoolBot document
Title: New `format` method for `box.tuple`
Product: Tarantool
Since: 3.2

The `tuple:format` method returns a format of a tuple.

### DIFF
--- a/changelogs/unreleased/gh-10005-tuple-get-format.md
+++ b/changelogs/unreleased/gh-10005-tuple-get-format.md
@@ -1,0 +1,3 @@
+## feature/box
+
+* Added the `tuple:format` method to get a format of a tuple (gh-10005).
--- a/src/box/lua/tuple.c
+++ b/src/box/lua/tuple.c
@@ -757,6 +757,23 @@ lbox_tuple_info(lua_State *L)
 	return 1;
 }
 
+/**
+ * Push to Lua stack an array with the information about a format of a tuple.
+ * Elements of the array are maps
+ * {'name' = 'field_name', 'type' = 'field_type'}.
+ */
+static int
+lbox_tuple_get_format(lua_State *L)
+{
+	int argc = lua_gettop(L);
+	if (argc != 1)
+		luaL_error(L, "Usage: tuple:format()");
+
+	struct tuple *tuple = luaT_checktuple(L, 1);
+	struct tuple_format *format = tuple_format(tuple);
+	return box_tuple_format_serialize_impl(L, format);
+}
+
 static const struct luaL_Reg lbox_tuple_meta[] = {
 	{"__gc", lbox_tuple_gc},
 	{"tostring", lbox_tuple_to_string},
@@ -766,6 +783,7 @@ static const struct luaL_Reg lbox_tuple_meta[] = {
 	{"tuple_field_by_path", lbox_tuple_field_by_path},
 	{"new", lbox_tuple_new},
 	{"info", lbox_tuple_info},
+	{"tuple_get_format", lbox_tuple_get_format},
 	{NULL, NULL}
 };
 
--- a/src/box/lua/tuple.lua
+++ b/src/box/lua/tuple.lua
@@ -332,6 +332,7 @@ local methods = {
     ["update"]      = tuple_update;
     ["upsert"]      = tuple_upsert;
     ["bsize"]       = tuple_bsize;
+    ["format"]      = box.tuple.format;
     ["tomap"]       = internal.tuple.tuple_to_map;
     ["info"]        = internal.tuple.info;
 }
--- a/src/box/lua/tuple_format.c
+++ b/src/box/lua/tuple_format.c
@@ -106,17 +106,15 @@ lbox_tuple_format_tostring(struct lua_State *L)
 	return 1;
 }
 
-/*
- * Returns the format clause with which this tuple format was created.
- */
-static int
-lbox_tuple_format_serialize(struct lua_State *L)
+int
+box_tuple_format_serialize_impl(struct lua_State *L,
+				struct tuple_format *format)
 {
-	struct tuple_format *format = luaT_check_tuple_format(L, 1);
 	if (format->data == NULL) {
 		lua_createtable(L, 0, 0);
 		return 1;
 	}
+
 	const char *data = format->data;
 	luamp_decode(L, luaL_msgpack_default, &data);
 	luaL_findtable(L, LUA_GLOBALSINDEX, "box.internal.space", 1);
@@ -125,6 +123,16 @@ lbox_tuple_format_serialize(struct lua_State *L)
 	lua_pushvalue(L, -2);
 	lua_call(L, 1, 1);
 	return 1;
+}
+
+/*
+ * Returns the format clause with which this tuple format was created.
+ */
+static int
+lbox_tuple_format_serialize(struct lua_State *L)
+{
+	struct tuple_format *format = luaT_check_tuple_format(L, 1);
+	return box_tuple_format_serialize_impl(L, format);
 }
 
 /*
--- a/src/box/lua/tuple_format.h
+++ b/src/box/lua/tuple_format.h
@@ -28,6 +28,13 @@ luaT_check_tuple_format(struct lua_State *L, int narg);
 void
 box_lua_tuple_format_init(struct lua_State *L);
 
+/**
+ * Returns the clause for the given format.
+ */
+int
+box_tuple_format_serialize_impl(struct lua_State *L,
+				struct tuple_format *format);
+
 #if defined(__cplusplus)
 } /* extern "C" */
 #endif /* defined(__cplusplus) */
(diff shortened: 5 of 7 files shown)


test/box/tuple.result
```diff
@@ -364,6 +364,34 @@ t[nil]
 - null
 ...
 --------------------------------------------------------------------------------
+-- test tuple:format
+--------------------------------------------------------------------------------
+f = box.tuple.format.new({{'id', 'number'}, {'name', 'string'}})
+---
```
oleg-jukovec, line 370: It looks like it's better to test this feature using unit tests, see example:
https://github.com/tarantool/tarantool/pull/8549/

tests with the output have not been updated for a long time, there is unit-tests prefered at now.

test/box/tuple.result
```diff
@@ -364,6 +364,34 @@ t[nil]
...
+box.tuple.format(t)
+---
+- [{'name': 'id', 'type': 'number'}, {'name': 'name', 'type': 'string'}]
+...
+t = box.tuple.new({2, 'a', false})
+---
+...
+t:format()
+---
+- []
```
oleg-jukovec, line 388: Maybe it should return a nil in this case.

oleg-jukovec: But space:format() without a specified format returns an empty format object. So it could be an expected behavior.

src/box/lua/tuple.c
```diff
@@ -757,6 +757,30 @@ lbox_tuple_info(lua_State *L)
...
+		luaL_error(L, "Usage: tuple:format()");
+
+	struct tuple *tuple = luaT_checktuple(L, 1);
+	struct tuple_format *format = tuple_format(tuple);
+	tuple_format_ref(format);
+
+	struct tuple_format **format_ptr =
+		lua_newuserdata(L, sizeof(*format_ptr));
+
+	luaL_getmetatable(L, "box.tuple.format");
```
oleg-jukovec, line 778: We need to create a constant for the string.
```suggestion
	luaL_getmetatable(L, "box.tuple.format");
```

sergepetrenko (changes requested):
Thanks for working on this!

src/box/lua/tuple.c
```diff
@@ -757,6 +757,30 @@ lbox_tuple_info(lua_State *L)
 	return 1;
 }
 
+/**
+ * Push to Lua stack a table with the information about a format of a tuple.
+ * Elements of a table are structs {'name', 'type'}.
+ */
```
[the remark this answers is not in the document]
oleg-jukovec: So is the implementation should be similar to `lbox_tuple_to_map`:

https://github.com/tarantool/tarantool/blob/c5b3e5949b43131276b492e59dec5a41fa49573a/src/box/lua/tuple.c#L499-L555

But create a table-array with {name = 'field_name', type = 'field_type'} instead of table-map?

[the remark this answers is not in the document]
sergepetrenko: > So is the implementation should be similar to `lbox_tuple_to_map`:
> 
> https://github.com/tarantool/tarantool/blob/c5b3e5949b43131276b492e59dec5a41fa49573a/src/box/lua/tuple.c#L499-L555
> 
> But create a table-array with items `{name = 'field_name', type = 'field_type'}` instead of the table-map `{name/idx = value}`?

I think you can reuse the implementation of `tuple_format_serialize` here:
https://github.com/tarantool/tarantool/blob/c5b3e5949b43131276b492e59dec5a41fa49573a/src/box/lua/tuple_format.c#L112-L128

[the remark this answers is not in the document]
DerekBum: Or can I return the `format:totable()` for the format I got? Will it be ok?

test/box/tuple.result
DerekBum: Added new tests into `test/app-luatest` directory.

src/box/lua/tuple.c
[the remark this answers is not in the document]
DerekBum: Fixed, this function now pushes a table for real. I kept the current comment, is it ok?

src/box/lua/tuple.c
DerekBum: This line was removed, so no constants were added.

src/box/lua/tuple.c
```diff
@@ -757,6 +757,34 @@ lbox_tuple_info(lua_State *L)
...
+	struct tuple_format *format = tuple_format(tuple);
+	if (format->data == NULL) {
+		lua_createtable(L, 0, 0);
+		return 1;
+	}
+
+	const char *data = format->data;
+	luamp_decode(L, luaL_msgpack_default, &data);
+	luaL_findtable(L, LUA_GLOBALSINDEX, "box.internal.space", 1);
+	lua_getfield(L, -1, "denormalize_format");
```
oleg-jukovec, line 781: At first glance, these strings should be constants.

src/box/lua/tuple.c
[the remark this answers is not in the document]
oleg-jukovec: It still confuses me. Something like this will be more clear:
```
* Returns an array with the information about a format of a tuple.
* Elements of the array are structs {'name' = field_name, 'type' = field_type}.
```

test/box/tuple.result
oleg-jukovec: So we could remove the tests now.

src/box/lua/tuple.c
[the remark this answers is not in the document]
oleg-jukovec: It still confuses me. Something like this will be more clear:
```
* Returns an array with the information about a format of a tuple.
* Elements of the array are structs {'name' = field_name, 'type' = field_type}.
```

test/box/tuple.result
oleg-jukovec: So we could remove the tests now.

DerekBum: Ok, updated.

src/box/lua/tuple.c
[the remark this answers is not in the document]
DerekBum: Changed this comment.

src/box/lua/tuple.c
DerekBum: In this file these strings are used only once each. I don't see a necessity to wrap them into separate `char*` constants. Why do you think this change should be done?

oleg-jukovec: Twice already:
https://github.com/tarantool/tarantool/blob/c5b3e5949b43131276b492e59dec5a41fa49573a/src/box/lua/tuple_format.c#L122-L123

But with the `grep LUA_GLOBALSINDEX -r src` it seems like ok.

oleg-jukovec (approved):
Thank you for the patch!

sergepetrenko (changes requested):
Thanks for the quick fixes!
Two comments left and we'll be good to go.

test/app-luatest/gh_10005_tuple_get_format.lua
```diff
@@ -0,0 +1,50 @@
+local server = require('luatest.server')
```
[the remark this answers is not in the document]
DerekBum: Thanks, updated the test name.

src/box/lua/tuple.c
```diff
@@ -757,6 +757,35 @@ lbox_tuple_info(lua_State *L)
...
+	lua_remove(L, -2);
+	lua_pushvalue(L, -2);
+	lua_call(L, 1, 1);
+	return 1;
```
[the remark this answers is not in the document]
DerekBum: Thanks, updated.

sergepetrenko (approved):
Thanks! LGTM.

changelogs/unreleased/gh-10005-tuple-get-format.md
```diff
@@ -0,0 +1,3 @@
+## feature/box
+
+* Added `tuple:format` method to get a format of a tuple (gh-10005).
```
xuniq, line 3:
```suggestion
* Added the `tuple:format` method to get a format of a tuple (gh-10005).
```

xuniq (approved):

PR #10139, opened 2024-06-14
Implement logarithmic `index:count` in the memtx tree

### LINKED ISSUE #10140: Memtx's `index:count(..., {iterator = 'ALL'})` is inconsistent with `index:select`
**Bug description**

According to the documentation the `index:count(key, params)` return amount of tuples matched with given key and iterator. In other words:  `s:count(key, {iterator = it}) == #(s:select(key, {iterator = it}))`. This is not the case for `ALL` iterator in memtx:

```lua
box.cfg{}

s = box.schema.create_space('s')
s:create_index('pk')

s:insert({1})
s:insert({2})
s:insert({3})

s:select(2, {iterator = 'ALL'}) -- {2, 3}
s:count(2, {iterator = 'ALL'}) -- 3, but expected 2
```

The same code works as expected in vinyl.

May be fixed by #10139 or separately.

mkostoevr:
Prior to this patch the count request had been performed in the memtx
in a straightforward way: we created an iterator by a type and key and
simply called its `next` method until it's exhausted. That means the
operation had a linear complexity, which could lead to DoS situations.

Also the count operation with ALL iterator hadn't been recorded in the
MVCC previously and had an erroneous logic (see the https://github.com/tarantool/tarantool/issues/10140). This is
fixed too as a side effect.

This patch makes the memtx tree index use a version of BPS tree with
the cardinality information enabled and takes advantage of its offset
based API to implement the count operation using tree lookups.

Since the method does not read each counted tuple now, MVCC subsystem
would be unaware of it. In order to fix this, this patch introduces a
new entity in the memtx transaction manager to track such operations:
GAP_COUNT, and the corresponding `memtx_tx_track_count` function.

The entity (gap item) is a record that a concurrent transaction has
counted tuples matching some key and iterator type in an index.

If a transaction creates such an entity, any insertion or deletion
of a matching tuple in the index will be conflicted with it. This
works differently for inserts and deletes:

1. If a concurrent transaction inserts a new matching tuple, then
   its read_gaps list is modified like the counted transaction had
   read the exact key of the tuple and found nothing. This creates
   a conflict.

2. If a concurrent transaction deletes a matching tuple, then the
   transaction that counted the tuple is inserted into the tuple
   reader list: it pretends to have read the tuple prior to the
   deletion. This creates a conflict.

The existing stories of matching dirty tuples are handled differently.
Since its's not stored directly whether the story is a product of an
insertion or a replacement, any matching visible tuple is marked as
read by the counting transaction and any matching invisible one is
marked as gap read, so the conflicting scenarios remain the same for
old stories.

One thing to be noted is that using the cardinality config introduces
a performance regression against the current version of memtx, this is
to be mitigated in the scope of https://github.com/tarantool/tarantool/issues/10322.

Part of #8204
Closes #10140

NO_DOC=perf improvement
NO_CHANGELOG=TBD when fully implemented

### DIFF
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -772,32 +772,42 @@ prepare_start_prefix_iterator(struct memtx_tree_key_data<USE_HINT> *start_data,
 	return true;
 }
 
-template <bool USE_HINT>
-static int
-tree_iterator_start(struct iterator *iterator, struct tuple **ret)
+/**
+ * Creates an iterator based on the given key, after data and iterator type.
+ * Also updates @a start_data and iterator @a type as required.
+ *
+ * @param tree - the tree to lookup in;
+ * @param start_data - the key to lookup with, may be updated;
+ * @param after_data - the after key, can be empty if not required;
+ * @param type - the lookup iterator type, may be updated;
+ * @param region - the region to allocate a new @a start_data on if required.
+ * @param[out] iterator - the result of the lookup;
+ * @param[out] equals - true if the lookup gave the exact key match;
+ * @param[out] initial_elem - the element approached on initial lookup, stepped
+ *  over if the iterator is reverse, see the end of this function for more info.
+ *
+ * @retval true on success;
+ * @retval false if the iteration must be stopped without an error.
+ */
+template<bool USE_HINT>
+static bool
+memtx_tree_lookup(memtx_tree_t<USE_HINT> *tree,
+		  struct memtx_tree_key_data<USE_HINT> *start_data,
+		  struct memtx_tree_key_data<USE_HINT> after_data,
+		  enum iterator_type *type, struct region *region,
+		  memtx_tree_iterator_t<USE_HINT> *iterator,
+		  bool *equals,
+		  struct memtx_tree_data<USE_HINT> **initial_elem)
 {
-	struct region *region = &fiber()->gc;
-	RegionGuard region_guard(region);
-	*ret = NULL;
-	struct space *space;
-	struct index *index_base;
-	index_weak_ref_get_checked(&iterator->index_ref, &space, &index_base);
-	struct memtx_tree_index<USE_HINT> *index =
-		(struct memtx_tree_index<USE_HINT> *)index_base;
-	struct tree_iterator<USE_HINT> *it = get_tree_iterator<USE_HINT>(iterator);
-	iterator->next_internal = exhausted_iterator_next;
-	memtx_tree_t<USE_HINT> *tree = &index->tree;
-	struct txn *txn = in_txn();
-	struct key_def *cmp_def = index->base.def->cmp_def;
-	struct memtx_tree_key_data<USE_HINT> start_data =
-		it->after_data.key != NULL ? it->after_data : it->key_data;
-	enum iterator_type type = it->type;
-	if ((type == ITER_NP || type == ITER_PP) &&
-	    it->after_data.key == NULL) {
-		if (!prepare_start_prefix_iterator(&start_data, &type,
+	struct key_def *cmp_def = memtx_tree_cmp_def(tree);
+
+	if ((*type == ITER_NP || *type == ITER_PP) &&
+	    after_data.key == NULL) {
+		if (!prepare_start_prefix_iterator(start_data, type,
 						   cmp_def, region))
-			return 0;
+			return false;
 	}
+
 	/*
 	 * Since iteration with equality iterators returns first found tuple,
 	 * we need a special flag for EQ and REQ if we want to start iteration
@@ -806,20 +816,14 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 	 * As for range iterators with equality, we can simply change them
 	 * to their equivalents with inequality.
 	 */
-	bool skip_equal_tuple = it->after_data.key != NULL;
-	if (skip_equal_tuple && type != ITER_EQ && type != ITER_REQ)
-		type = iterator_type_is_reverse(type) ? ITER_LT : ITER_GT;
-	/*
-	 * The key is full - all parts a present. If key if full, EQ and REQ
-	 * queries can return no more than one tuple.
-	 */
-	bool key_is_full = start_data.part_count == cmp_def->part_count;
-	/* The flag will be change to true if found tuple equals to the key. */
-	bool equals = false;
-	assert(it->last.tuple == NULL);
-	if (start_data.key == NULL) {
-		assert(type == ITER_GE || type == ITER_LE);
-		if (iterator_type_is_reverse(type))
+	bool skip_equal_tuple = after_data.key != NULL;
+	if (skip_equal_tuple && *type != ITER_EQ && *type != ITER_REQ)
+		*type = iterator_type_is_reverse(*type) ? ITER_LT : ITER_GT;
+
+	/* Perform the initial lookup. */
+	if (start_data->key == NULL) {
+		assert(*type == ITER_GE || *type == ITER_LE);
+		if (iterator_type_is_reverse(*type))
 			/*
 			 * For all reverse iterators we will step back,
 			 * see the and explanation code below.
@@ -827,12 +831,13 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 			 * a back step from invalid iterator set its
 			 * position to the last element. Let's use that.
 			 */
-			invalidate_tree_iterator(&it->tree_iterator);
+			invalidate_tree_iterator(iterator);
 		else
-			it->tree_iterator = memtx_tree_first(tree);
+			*iterator = memtx_tree_first(tree);
+
 		/* If there is at least one tuple in the tree, it is
 		 * efficiently equals to the empty key. */
-		equals = memtx_tree_size(tree) != 0;
+		*equals = memtx_tree_size(tree) != 0;
 	} else {
 		/*
 		 * We use lower_bound on equality iterators instead of LE
@@ -842,34 +847,29 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 		 * So, lower_bound is used for EQ, GE and LT iterators,
 		 * upper_bound is used for REQ, GT, LE iterators.
 		 */
-		bool need_lower_bound = type == ITER_EQ || type == ITER_GE ||
-					type == ITER_LT;
+		bool need_lower_bound = *type == ITER_EQ || *type == ITER_GE ||
+					*type == ITER_LT;
 
 		/*
 		 * If we need to skip first tuple in EQ and REQ iterators,
 		 * let's just change lower_bound to upper_bound or vice-versa.
 		 */
-		if (skip_equal_tuple && (type == ITER_EQ || type == ITER_REQ))
+		if (skip_equal_tuple && (*type == ITER_EQ || *type == ITER_REQ))
 			need_lower_bound = !need_lower_bound;
+
 		if (need_lower_bound) {
-			it->tree_iterator =
-				memtx_tree_lower_bound(tree, &start_data,
-						       &equals);
+			*iterator = memtx_tree_lower_bound(tree, start_data,
+							   equals);
 		} else {
-			it->tree_iterator =
-				memtx_tree_upper_bound(tree, &start_data,
-						       &equals);
+			*iterator = memtx_tree_upper_bound(tree, start_data,
+							   equals);
 		}
 	}
 
-	/*
-	 * `it->tree_iterator` could potentially be positioned on successor of
-	 * key: we need to track gap based on it.
-	 */
-	struct memtx_tree_data<USE_HINT> *res =
-		memtx_tree_iterator_get_elem(tree, &it->tree_iterator);
-	struct tuple *successor = res == NULL ? NULL : res->tuple;
-	if (iterator_type_is_reverse(type)) {
+	/* Save the element we approached on the initial lookup. */
+	*initial_elem = memtx_tree_iterator_get_elem(tree, iterator);
+
+	if (iterator_type_is_reverse(*type))
 		/*
 		 * Because of limitations of tree search API we use
 		 * lower_bound for LT search and upper_bound for LE and
@@ -881,11 +881,53 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 		 * iterator_prev call will convert the iterator to the
 		 * last position in the tree, that's what we need.
 		 */
-		memtx_tree_iterator_prev(tree, &it->tree_iterator);
+		memtx_tree_iterator_prev(tree, iterator);
+	return true;
+}
+
+template<bool USE_HINT>
+static int
+tree_iterator_start(struct iterator *iterator, struct tuple **ret)
+{
+	struct region *region = &fiber()->gc;
+	RegionGuard region_guard(region);
+
+	*ret = NULL;
+	iterator->next_internal = exhausted_iterator_next;
+
+	struct tree_iterator<USE_HINT> *it =
+		get_tree_iterator<USE_HINT>(iterator);
+	assert(it->last.tuple == NULL);
+
+	struct space *space;
+	struct index *index_base;
+	index_weak_ref_get_checked(&iterator->index_ref, &space, &index_base);
+	struct memtx_tree_index<USE_HINT> *index =
+		(struct memtx_tree_index<USE_HINT> *)index_base;
+	memtx_tree_t<USE_HINT> *tree = &index->tree;
+	struct memtx_tree_key_data<USE_HINT> start_data =
+		it->after_data.key != NULL ? it->after_data : it->key_data;
+	enum iterator_type type = it->type;
+	/* The flag is true if the found tuple equals to the key. */
+	bool equals;
+	struct memtx_tree_data<USE_HINT> *initial_elem;
+	if (!memtx_tree_lookup(tree, &start_data, it->after_data,
+			       &type, region, &it->tree_iterator,
+			       &equals, &initial_elem))
+		return 0;
+
+	/*
+	 * The initial element could potentially be a successor of the key: we
+	 * need to track gap based on it.
+	 */
+	struct tuple *successor = initial_elem ? initial_elem->tuple : NULL;
+
+	struct memtx_tree_data<USE_HINT> *res = initial_elem;
+	if (iterator_type_is_reverse(type))
 		res = memtx_tree_iterator_get_elem(tree, &it->tree_iterator);
-	}
+
 	/* If we skip tuple, flag equals is not actual - need to refresh it. */
-	if (skip_equal_tuple && res != NULL &&
+	if (it->after_data.key != NULL && res != NULL &&
 	    (type == ITER_EQ || type == ITER_REQ)) {
 		equals = tuple_compare_with_key(res->tuple, res->hint,
 						it->key_data.key,
@@ -893,6 +935,8 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 						it->key_data.hint,
 						index->base.def->key_def) == 0;
 	}
+
+	struct txn *txn = in_txn();
 	/*
 	 * Equality iterators requires exact key match: if the result does not
 	 * equal to the key, iteration ends.
@@ -910,7 +954,14 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 		*ret = memtx_tx_tuple_clarify(txn, space, res->tuple,
 					      index_base, mk_index);
 	}
+
 /********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
+	/*
+	 * If the key is full then all parts present, so EQ and REQ iterators
+	 * can return no more than one tuple.
+	 */
+	struct key_def *cmp_def = index->base.def->cmp_def;
+	bool key_is_full = start_data.part_count == cmp_def->part_count;
 	if (key_is_full && !eq_match)
 		memtx_tx_track_point(txn, space, index_base, it->key_data.key);
 	/*
@@ -927,6 +978,7 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 				   MIN(start_data.part_count, key_part_count));
 	memtx_tx_story_gc();
 /*********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND END**********/
+
 	return res == NULL || !eq_match || *ret != NULL ? 0 :
 	       iterator->next_internal(iterator, ret);
 }
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -110,6 +110,7 @@ memtx_tree_data_is_equal(const struct memtx_tree_data_common *a,
 	return a->tuple == b->tuple;
 }
 
+#define BPS_INNER_CARD
 #define BPS_TREE_NAME memtx_tree
 #define BPS_TREE_BLOCK_SIZE (512)
 #define BPS_TREE_EXTENT_SIZE MEMTX_EXTENT_SIZE
@@ -782,9 +783,11 @@ prepare_start_prefix_iterator(struct memtx_tree_key_data<USE_HINT> *start_data,
  * @param type - the lookup iterator type, may be updated;
  * @param region - the region to allocate a new @a start_data on if required.
  * @param[out] iterator - the result of the lookup;
+ * @param[out] offset - the offset @a iterator points to;
  * @param[out] equals - true if the lookup gave the exact key match;
- * @param[out] initial_elem - the element approached on initial lookup, stepped
- *  over if the iterator is reverse, see the end of this function for more info.
+ * @param[out] initial_elem - the optional pointer to the element approached on
+ *  the initial lookup, stepped over if the iterator is reverse, see the end of
+ *  this function for more information.
  *
  * @retval true on success;
  * @retval false if the iteration must be stopped without an error.
@@ -796,7 +799,7 @@ memtx_tree_lookup(memtx_tree_t<USE_HINT> *tree,
 		  struct memtx_tree_key_data<USE_HINT> after_data,
 		  enum iterator_type *type, struct region *region,
 		  memtx_tree_iterator_t<USE_HINT> *iterator,
-		  bool *equals,
+		  size_t *offset, bool *equals,
 		  struct memtx_tree_data<USE_HINT> **initial_elem)
 {
 	struct key_def *cmp_def = memtx_tree_cmp_def(tree);
@@ -823,7 +826,7 @@ memtx_tree_lookup(memtx_tree_t<USE_HINT> *tree,
 	/* Perform the initial lookup. */
 	if (start_data->key == NULL) {
 		assert(*type == ITER_GE || *type == ITER_LE);
-		if (iterator_type_is_reverse(*type))
+		if (iterator_type_is_reverse(*type)) {
 			/*
 			 * For all reverse iterators we will step back,
 			 * see the and explanation code below.
@@ -832,8 +835,11 @@ memtx_tree_lookup(memtx_tree_t<USE_HINT> *tree,
 			 * position to the last element. Let's use that.
 			 */
 			invalidate_tree_iterator(iterator);
-		else
+			*offset = memtx_tree_size(tree);
+		} else {
 			*iterator = memtx_tree_first(tree);
+			*offset = 0;
+		}
 
 		/* If there is at least one tuple in the tree, it is
 		 * efficiently equals to the empty key. */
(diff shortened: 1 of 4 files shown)


src/box/memtx_tx.c
```diff
@@ -2032,6 +2143,18 @@ memtx_tx_history_add_insert_stmt(struct txn_stmt *stmt,
...
 		struct tuple *next = directly_replaced[i];
 		struct tuple *succ = direct_successor[i];
+		/*
+		 * The tuple was inserted as new according to the operating
+		 * transaction and it's not excluded from the index. It's
+		 * possible that the operation replaced a tuple in the index,
+		 * but, if so, the tuple is not visible to the inserter.
+		 */
+		if (old_tuple == NULL && !tuple_excluded[i])
+			memtx_tx_handle_count_conflicts(space, add_story, i);
```
mkostoevr, line 2179: It's not clear to me now why this required a separated condition, maybe other conflict detecting functions called below should be moved here too.

src/box/memtx_tree.cc
```diff
@@ -2235,19 +2330,21 @@ tree_read_view_create_iterator(struct index_read_view *base,
...
 /** Implementation of create_read_view index callback. */
-template <bool USE_HINT>
+template<bool FAST_OFFSET, bool USE_HINT>
 static struct index_read_view *
 memtx_tree_index_create_read_view(struct index *base)
 {
 	static const struct index_read_view_vtab vtab = {
-		.free = tree_read_view_free<USE_HINT>,
+		.free = tree_read_view_free<FAST_OFFSET, USE_HINT>,
 		.get_raw = tree_read_view_get_raw<USE_HINT>,
```
mkostoevr, line 2468: The `tree_read_view_get_raw` needs to consume the `FAST_OFFSET` template parameter too (it's required to implement the read view get in the Tarantool EE (for memtx tree read views).

src/box/lua/space.cc
```diff
@@ -566,8 +566,13 @@ lbox_fillspace(struct lua_State *L, struct space *space, int i)
...
 
-		lua_pushnil(L);
-		lua_setfield(L, -2, "fast_offset");
+		if (space_is_memtx(space) && index_def->type == TREE) {
+			lua_pushboolean(L, index_opts->fast_offset);
+			lua_setfield(L, -2, "fast_offset");
+		} else {
+			lua_pushnil(L);
+			lua_setfield(L, -2, "fast_offset");
+		}
```
locker, line 575: Nit: There's exactly the same condition a few lines above - where we report 'hint'. You could fold this code there.

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,91 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
...
+
+		/* An empty key matches to any tuple. */
+		return true;
+	}
+
+	struct key_def *def = index->def->key_def;
+	hint_t key_hint = def->key_hint(key, part_count, def);
+	hint_t tuple_hint = def->tuple_hint(tuple, def);
+	int cmp = def->tuple_compare_with_key(tuple, tuple_hint, key,
+					      part_count, key_hint, def);
```
locker, line 1815: Nit: Please use wrappers instead of calling `key_def` members directly.

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,91 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
...
+	rlist_foreach_entry_safe(item_base, &index->read_gaps,
+				 in_read_gaps, tmp) {
+		if (item_base->type != GAP_COUNT)
+			continue;
+		struct count_gap_item *item =
+			(struct count_gap_item *)item_base;
+
+		bool tuple_matches = memtx_tx_tuple_matches(
+			index, story->tuple, item->base.type,
+			item->base.key, item->base.part_count);
```
locker, line 1856: AFAIR the iterator key may be deleted after the iterator is closed and before the transaction completes. Can this result in use-after-free here?

mkostoevr: The space for the key is allocated in the base class object constructor (`memtx_tx_nearby_gap_item_new`), so it's safe to use it here. Actually, it's either stored on the transaction region or in the object itself.

src/box/lua/space.cc
```diff
@@ -566,6 +566,9 @@ lbox_fillspace(struct lua_State *L, struct space *space, int i)
 			lua_setfield(L, -2, "hint");
 		}
 
+		lua_pushnil(L);
+		lua_setfield(L, -2, "fast_offset");
```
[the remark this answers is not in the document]
mkostoevr: Squashed with introducing new tree for memtx index.

Whether the option should exist of we should enable the featrure unconditionally is yet undecided, see the [benchmaks](https://github.com/mkostoevr/notes/blob/master/perf0.md).

src/box/memtx_tree.cc
```diff
@@ -728,6 +728,50 @@ prepare_start_prefix_iterator(struct memtx_tree_key_data<USE_HINT> *start_data,
 	return true;
 }
 
+static int
+memtx_tree_lookup_canonicalize(struct index *base, struct key_def *cmp_def,
+			       enum iterator_type *out_type,
+			       const char **out_key, uint32_t part_count)
```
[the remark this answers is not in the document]

src/box/memtx_tree.cc
```diff
@@ -772,32 +772,55 @@ memtx_tree_lookup_canonicalize(struct index *base, struct key_def *cmp_def,
...
-template <bool USE_HINT>
-static int
-tree_iterator_start(struct iterator *iterator, struct tuple **ret)
+enum memtx_tree_lookup_op {
+	NOOP,
+	FIRST,
+	LOWER_BOUND,
+	UPPER_BOUND,
+	INVALIDATE,
+};
```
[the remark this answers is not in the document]
mkostoevr: Done, the common code looks much better now.

src/box/lua/space.cc
mkostoevr: Done.

src/box/memtx_tx.c
mkostoevr: Done.

src/box/memtx_tree.cc
```diff
@@ -772,32 +772,44 @@ prepare_start_prefix_iterator(struct memtx_tree_key_data<USE_HINT> *start_data,
...
+ *
+ * @param tree - the tree to lookup in;
+ * @param start_data - the key to lookup with;
+ * @param after_data - the after key, can be empty if not required;
+ * @param type - the lookup iterator type;
+ * @param region - the region to allocate a new @a start_data on if required.
+ * @param[out] iterator - the result of the primary lookup;
+ * @param[out] equals - true if the primary lookup gave the exact key match;
+ * @param[out] step_back - true if the result iterator is not valid and needs
+ *  to be adjusted using the memtx_tree_iterator_prev to reach the target.
```
locker, line 885: I don't like that the caller may need to "step back" after calling this function. AFAIU the function doesn't step back by itself for the caller to get the successor. IMO it's better to return the successor instead of the flag.

src/box/memtx_space.c
```diff
@@ -844,6 +858,13 @@ memtx_space_check_index_def(struct space *space, struct index_def *index_def)
 				 "BITSET can not be unique");
 			return -1;
 		}
+		if (index_def->opts.fast_offset) {
+			diag_set(ClientError, ER_MODIFY_INDEX,
+				 index_def->name, space_name(space),
+				 "BITSET index does not support "
+				 "logarithmic select with offset");
+			return -1;
+		}
```
locker, line 867: Instead of duplicating this check for each index type, let's just put it near the hint check below:

```C
	if (index_def->type != TREE && index_def->opts.fast_offset) {
		diag_set(ClientError, ER_MODIFY_INDEX, index_def->name,
			 space_name(space),
			 "fast_offset is only reasonable with memtx tree index");
		return -1;
	}
```

src/box/index_def.c
```diff
@@ -89,6 +90,7 @@ const struct opt_def index_opts_reg[] = {
 	OPT_DEF("func", OPT_UINT32, struct index_opts, func_id),
 	OPT_DEF_LEGACY("sql"),
 	OPT_DEF_CUSTOM("hint", index_opts_parse_hint),
+	OPT_DEF("fast_offset", OPT_BOOL, struct index_opts, fast_offset),
```
locker, line 93: If we decide to keep the flag, let's enable it by default, as it was suggested in the design document.

src/box/memtx_tree.cc
```diff
@@ -1220,16 +1258,122 @@ memtx_tree_index_random(struct index *base, uint32_t rnd, struct tuple **result)
...
+		start_data.set_hint(key_hint(key, part_count, cmp_def));
+	struct memtx_tree_key_data<USE_HINT> null_after_data = {};
+	memtx_tree_iterator_t<true, USE_HINT> begin;
+	size_t begin_offset;
+	bool unused;
+	bool step_back;
+	if (!memtx_tree_lookup<true, USE_HINT, true>(
+		tree, &start_data, null_after_data, &type, region,
+		&begin, &begin_offset, &unused, &step_back))
+		return 0;
```
locker, line 1306: Nit: Confusing indentation. I'd write:

```C
	if (!memtx_tree_lookup<true, USE_HINT, true>(
			tree, &start_data, null_after_data, &type, region,
			&begin, &begin_offset, &unused, &step_back))
		return 0;
```

src/box/memtx_tree.cc
```diff
@@ -1220,16 +1258,122 @@ memtx_tree_index_random(struct index *base, uint32_t rnd, struct tuple **result)
...
+
+	struct txn *txn = in_txn();
+	struct space *space = space_by_id(base->def->space_id);
+	size_t full_size = memtx_tree_size(tree);
+	size_t end_offset;
+
+	if (step_back)
+		begin_offset--; /* Unsigned overflow possible. */
+
+	if (begin_offset == size_t(-1)) {
```
locker, line 1316: Nit: `size_t(-1)` -> `SIZE_MAX`

src/box/memtx_tree.cc
mkostoevr: Done.

src/box/memtx_space.c
mkostoevr: Done.

src/box/memtx_tree.cc

test/box-luatest/gh_8204_fast_offset_test.lua
```diff
@@ -68,14 +102,844 @@ g.test_option = function()
...
+        for _, it in pairs(all_iterators) do
+            for _, key in pairs(test_keys) do
+                local expected_count = s_gold:count(key, {iterator = it})
+                local all_matching = s_all_keys:select(key, {iterator = it})
+                local matches = to_contains(all_matching)
+                for _, tuple in pairs(all_keys) do
+                    if not matches(tuple) then
+                        -- The tuple might not exist, no conflict anyways.
+                        count_do(tx1, tx2, it, key, expected_count,
+                                 'insert', tuple, success)
```
mkostoevr, line 906: Typo: "delete".

src/box/memtx_tx.c
```diff
@@ -2032,6 +2142,18 @@ memtx_tx_history_add_insert_stmt(struct txn_stmt *stmt,
...
 		struct tuple *next = directly_replaced[i];
 		struct tuple *succ = direct_successor[i];
+		/*
+		 * The tuple was inserted as new according to the operating
+		 * transaction and it's not excluded from the index. It's
+		 * possible that the operation replaced a tuple in the index,
+		 * but, if so, the tuple is not visible to the inserter.
+		 */
+		if (old_tuple == NULL && !tuple_excluded[i])
+			memtx_tx_handle_counted_write(space, add_story, i);
```
drewdzzz, line 2178: If write happened before count, it's not handled - you need to handle counted writes when they are prepared as well.
I already have a test case for you:
```lua
g_mvcc.test_review = function()
    g_mvcc.server:exec(function()
        -- The test space.
        local s = box.schema.space.create('test')
        s:create_index('pk', {fast_offset = true,
                              parts = {{1, 'unsigned'}, {2, 'unsigned'}}})
            
        -- Prepare proxies.
        local txn_proxy = require('test.box.lua.txn_proxy')
        local tx1 = txn_proxy.new()
        local tx2 = txn_proxy.new()

        tx1:begin()
        tx2:begin()
        tx1('box.space.test:replace{1, 0}')
        
        -- No prepared tuples - count must return 0
        local count = tx2('return box.space.test:count{1, 0}')[1]
        t.assert_equals(count, 0)

        -- Commit writer
        tx1:commit()

        -- Count again - the same value (zero) must be returned
        count = tx2('return box.space.test:count{1, 0}')[1]
        t.assert_equals(count, 0)
        tx2:commit()
    
        -- Check if insert actually happened
        t.assert_equals(box.space.test:select{}, {{1, 0}})
    end)
end
```

Now it fails on the second check of `count`.

src/box/memtx_tx.c
```diff
@@ -2017,16 +2017,42 @@ memtx_tx_handle_gap_write(struct space *space, struct memtx_story *story,
...
+
+		/*
+		 * The key+iterator pair can only match the new tuple, and its
+		 * predecessors, but not the successor's predecessors anymore.
+		 */
 		bool need_move = !need_split &&
 				 ((dir < 0 && cmp > 0) ||
 				  (cmp > 0 && item->type == ITER_EQ) ||
-				  (cmp == 0 && ((dir < 0 && is_full_key) ||
+				  (cmp == 0 && ((is_eq && is_full_key) ||
```
drewdzzz, line 2037: Doesn't it belong to the previous commit?

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,90 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
...
+		 * matching to the inserted or deleted tuple, it's a conflict.
+		 */
+		if (tuple_matches) {
+			if (is_insert) {
+				/*
+				 * Record like the counted transaction had read
+				 * by a key matching the tuple and got nothing
+				 * there. Now this insertion is conflicting.
+				 */
+				memtx_tx_track_story_gap(item_base->txn,
```
drewdzzz, line 1868: All tracking handlers would be wrapped into a warning. See rationale in one of my previous comments.

drewdzzz (changes requested):
Thanks for the patch!
I don't really like having this hanging commit introducing `gap_count_item`. It would be nice to have such structure:
1. One commit introduces all needed helpers in MVCC and rewrites linear count using them.
2. Another commit doesn't patch MVCC - it only introduces logarithmic count using already implemented and tested MVCC helpers.

By the way, now linear and logarithmic counts behave differently: linear conflicts with updates while logarithmic don't. But, IMHO, it doesn't really matter.

The only showstopper here is that linear count will be significantly slowed down in the cases when we have a lot of parallel transactions. AFAIU, it's more of a design problem with MVCC - it is designed for a small amount of parallel transactions. But, anyway, user considering `count` as a cheap operation can be unpleasantly surprised. And if we decide to always use `fast_offset` tree modification, we could rewrite linear count because it will be deleted anyway.

third_party/curl
mkostoevr, line 1: Unrelated and harmful change.

test/box-luatest/gh_8204_fast_offset_test.lua
```diff
@@ -68,14 +102,844 @@ g.test_option = function()
...
+        -- No conflict with itself (count empty by key & insert matching).
+        count_do(tx, tx, 'lt', {0, 1}, 0, 'insert', {0, 0}, success)
+        count_do(tx, tx, 'le', {0, 0}, 0, 'insert', {0, 0}, success)
+        count_do(tx, tx, 'req', {0, 0}, 0, 'insert', {0, 0}, success)
+        count_do(tx, tx, 'eq', {0, 0}, 0, 'insert', {0, 0}, success)
+        count_do(tx, tx, 'ge', {0, 0}, 0, 'insert', {0, 0}, success)
+        count_do(tx, tx, 'gt', {0, 0}, 0, 'insert', {0, 1}, success)
+
+        -- No conflict with itself (count empty by key & insert not matching).
+        count_do(tx, tx, 'lt', {0, 0}, 0, 'insert', {0, 0}, success)
```
mkostoevr, line 395: The insert silently fails, it must be a replace.

test/box-luatest/gh_8204_fast_offset_test.lua
```diff
@@ -0,0 +1,946 @@
...
+        -- Generate key lists.
+        for i = 0, 6 do
+            for j = 0, 6 do
+                if not to_exist(i) or not to_exist(j) then
+                    table.insert(unexisting_keys, {i, j})
+                else
+                    table.insert(existing_keys, {i, j})
+                end
+                table.insert(all_keys, {i, j})
+                table.insert(test_keys, {i, j})
```
mkostoevr, line 783: This is much more test keys than required, this slows down the test.

src/box/memtx_tx.h
```diff
@@ -318,6 +318,31 @@ memtx_tx_track_gap(struct txn *txn, struct space *space, struct index *index,
...
+void
+memtx_tx_track_count_slow(struct txn *txn, struct index *index,
+			  enum iterator_type type, const char *key,
+			  uint32_t part_count);
+
+/**
+ * Record in TX manager that a transaction @a txn have counted @a index from @a
+ * space by @a key and iterator @a type. This function must be used for queries
+ * that count tuples in indexes (for example, index:size or index:count).
+ */
```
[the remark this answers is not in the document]
mkostoevr: Done.

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,90 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
...
 }
 
+static bool
+memtx_tx_tuple_matches(struct index *index, struct tuple *tuple,
+		       enum iterator_type type, const char *key,
+		       uint32_t part_count)
+{
+	if (key == NULL) {
+		assert(part_count == 0);
+		assert(type == ITER_LE || type == ITER_GE);
```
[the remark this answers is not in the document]
mkostoevr: The `EQ/REQ` is transated to `GE/LE` if the part count is 0 in the [`create_iterator`](https://github.com/tarantool/tarantool/blob/4cac1677c88a0ee13944d8dc44b10813738bb445/src/box/memt

This meant to show to the reader which iterators are expected here. Please let me know if it's confusing.

src/box/memtx_tx.c
```diff
@@ -267,6 +274,17 @@ struct full_scan_gap_item {
...
+/**
+ * Derived class for count gap, @sa GAP_COUNT.
+ */
+struct count_gap_item {
+	/** Base class. */
+	struct nearby_gap_item base;
+};
+
+/* Count gap uses the same mempool as nearby gap. */
+static_assert(sizeof(struct count_gap_item) == sizeof(struct nearby_gap_item));
```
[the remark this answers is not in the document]
mkostoevr: Done.

src/box/memtx_tx.c
mkostoevr: Fixed.

src/box/memtx_tx.c
mkostoevr: No, these changes are introduced because the `count_fast` implementation (introduced in this commit) does not use the `memtx_tx_track_point` for readability reasons.

src/box/memtx_tx.c
mkostoevr: Done.

test/box-luatest/gh_8204_fast_offset_test.lua
```diff
@@ -0,0 +1,907 @@
...
+            local line = debug.getinfo(2, 'l').currentline
+
+            -- Remove past stories cause they cause unconditional conflicts,
+            -- whereas future statements only conflict with count if they
+            -- insert a new matching tuple or delete a counted one.
+            box.internal.memtx_tx_gc(100)
+
+            tx1:begin()
+            check(tx1, it, key, expected_count, file, line);
+            tx2(string.format('box.space.test:%s(%s)', op, to_lua_code(tuple)))
```
mkostoevr, line 322: One should check if the operation succeed (replace actually replaced, insert actually inserted, etc.).

src/box/index_def.c
mkostoevr: It's been decided to only keep one version of the tree (the one with cardinality info enabled).

src/box/memtx_tree.cc
```diff
@@ -782,21 +783,24 @@ prepare_start_prefix_iterator(struct memtx_tree_key_data<USE_HINT> *start_data,
...
- *  over if the iterator is reverse, see the end of this function for more info.
+ * @param[out] initial_elem - the optional pointer to the element approached on
+ *  the initial lookup, stepped over if the iterator is reverse, see the end of
+ *  this function for more information.
  *
  * @retval true on success;
  * @retval false if the iteration must be stopped without an error.
  */
-template<bool USE_HINT>
+template<bool USE_HINT, bool NEED_ITERATOR = true, bool NEED_OFFSET = false>
```
locker, line 796: It's okay to always return offset and iterator. It's cheap. If they aren't needed, they will be optimized out by the compiler. Please drop the template parameters.

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,90 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
...
+		assert(type == ITER_LE || type == ITER_GE);
+
+		/* An empty key matches to any tuple. */
+		return true;
+	}
+
+	struct key_def *def = index->def->key_def;
+	hint_t kh = key_hint(key, part_count, def);
+	hint_t th = tuple_hint(tuple, def);
+	int cmp = tuple_compare_with_key(tuple, th, key, part_count, kh, def);
```
locker, line 1814: Nit: There's no point in computing hints right before passing them to the comparator. You could just as well pass `HINT_NONE`.

locker (approved):

src/box/memtx_tx.c
mkostoevr: Thanks, calculating hints here does not seem reasonable.

src/box/memtx_tx.c
```diff
@@ -1769,6 +1795,88 @@ memtx_tx_handle_point_hole_write(struct space *space, struct memtx_story *story,
 	mh_point_holes_del(ht, pos, 0);
 }
 
+static bool
+memtx_tx_tuple_matches(struct index *index, struct tuple *tuple,
```
mkostoevr, line 1799: This needs the index just to extract a key def from it, let's consume it directly.

src/box/memtx_tree.cc
```diff
@@ -1156,9 +1165,104 @@ static ssize_t
...
+	if (!memtx_tree_lookup(tree, &start_data, null_after_data, &type,
+			       region, &unused1, &begin_offset, &unused2,
+			       &initial_elem))
+		return 0;
+
+	struct txn *txn = in_txn();
+	struct space *space = space_by_id(base->def->space_id);
+	size_t full_size = memtx_tree_size(tree);
+	size_t end_offset;
+
```
drewdzzz, line 1200: For the case when index is unique, iterator is `EQ` or `REQ` and full key is specified, we could simply call `memtx_tree_index_get_internal` as an optimization and see at its returned value. We will conflict with updates then, but it seems to me that advantages are more significant:
1. We will use "local" trackers instead of "global" ones (`track_point`/`track_read`/`inplace_gap` instead of `nearby_gap`/`count_gap`).
2. We won't need to patch `memtx_tx_handle_gap_write` (that extra logic for equality iterators with full key specified).

drewdzzz (approved):
Reviewed the last commit: LGTM, but left a suggestion.

src/box/memtx_tree.cc
mkostoevr: Made it use the `memtx_tx_track_point` function on missed full equal key lookup and removed the `memtx_tx_handle_gap_write` changes as not required anymore:

```diff
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -1198,6 +1198,15 @@ memtx_tree_index_count(struct index *base, enum iterator_type type,
        size_t full_size = memtx_tree_size(tree);
        size_t end_offset;
 
+       /* Fast path: not found equal with full key. */
+       if (start_data.part_count == cmp_def->part_count &&
+           !equals && (type == ITER_EQ || type == ITER_REQ)) {
+/********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
+               memtx_tx_track_point(txn, space, base, start_data.key);
+/*********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND END**********/
+               return 0; /* No tuple matching the full key. */
+       }
+
        /* Fast path: not found with reverse iterator. */
        if (begin_offset == size_t(-1)) {
                assert(iterator_type_is_reverse(type));
```

I don't like this invonsistency of conflicting cases for the `count` function, so the `GAP_COUNT` is still created, but if it will become an issue it can be fixed by simply moving the `COUNT` gap item into a specific element and introducing a little of additional logic based on it.

Please let me know if you're OK with it.

src/box/memtx_tx.c
```diff
@@ -1907,16 +2015,42 @@ memtx_tx_handle_gap_write(struct space *space, struct memtx_story *story,
...
+		/*
+		 * The inserted tuple matches the key and iterator. That means
+		 * we've inserted a tuple at the place that was read by some TX
+		 * that found nothing there. So the reader must be conflicted
+		 * with us once we commit, let's track that.
+		 */
 		bool need_track = need_split ||
-				  (is_full_key && cmp == 0 && is_e);
+				  (is_full_key && cmp == 0 && (is_e || is_eq));
+
```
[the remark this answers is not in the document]
mkostoevr: Removed these changes from the patch, please see the [comment below](https://github.com/tarantool/tarantool/pull/10139#discussion_r1721430485).

src/box/memtx_tree.cc
```diff
@@ -1156,9 +1165,115 @@ static ssize_t
...
+	memtx_tree_t<USE_HINT> *tree = &index->tree;
+	struct key_def *cmp_def = memtx_tree_cmp_def(&index->tree);
+	struct memtx_tree_key_data<USE_HINT> start_data;
+	start_data.key = key;
+	start_data.part_count = part_count;
+	if (USE_HINT)
+		start_data.set_hint(key_hint(key, part_count, cmp_def));
+	struct memtx_tree_key_data<USE_HINT> null_after_data = {};
+	memtx_tree_iterator_t<USE_HINT> unused1;
```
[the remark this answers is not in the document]
mkostoevr: Creating a BPS tree iterator is essentially free, so it's done for the sake of code reuseability (one can use the same lookup routine both to create a MemTX iterator and count by key).

src/box/memtx_tx.c
```diff
@@ -3246,6 +3396,73 @@ memtx_tx_track_gap_slow(struct txn *txn, struct space *space, struct index *inde
...
 
+/**
+ * Record in TX manager that a transaction @a txn have counted @a index of @a
+ * space by @a key and iterator @a type. This function must be used for queries
+ * that count tuples in indexes (for example, index:size or index:count).
+ *
+ * @return the amount of invisible tuples counted.
+ */
+uint32_t
+memtx_tx_track_count_slow(struct txn *txn, struct space *space,
```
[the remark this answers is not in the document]
mkostoevr: The main common thing is iteration over top stories, refactored it into a macro in a separated commit.

src/box/memtx_tx.c
```diff
@@ -267,6 +274,17 @@ struct full_scan_gap_item {
 	struct gap_item_base base;
 };
 
+/**
+ * Derived class for count gap, @sa GAP_COUNT.
+ */
+struct count_gap_item {
+	/** Base class. */
+	struct nearby_gap_item base;
+};
```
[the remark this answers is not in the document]
mkostoevr: Used a typedef. Checkpatch complained about it, so I used a hack to stop it: appened `__` to the struct name.

Maybe it's better to use a regular `define` here instead, not sure.

This structure is going to be overwritten in the select with offset implementation BTW.

src/box/memtx_tx.c
```diff
@@ -2812,6 +2932,21 @@ memtx_tx_tuple_clarify_impl(struct txn *txn, struct space *space,
...
+static struct tuple *
+memtx_tx_tuple_clarify_impl(struct txn *txn, struct space *space,
+			    struct tuple *tuple, struct index *index,
+			    uint32_t mk_index, bool is_prepared_ok)
+{
+	assert(tuple_has_flag(tuple, TUPLE_IS_DIRTY));
+	struct memtx_story *story = memtx_tx_story_get(tuple);
+	return memtx_tx_story_clarify_impl(txn, space, story, index,
+					   mk_index, is_prepared_ok);
+}
```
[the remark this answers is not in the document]
mkostoevr: Moved into a separated commit.

src/box/memtx_tree.cc
mkostoevr: It's created in the slow path (even for full key with EQ iterator).

src/box/memtx_tx.c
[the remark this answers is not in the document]
mkostoevr: Good idea, used a `#define` for now, since this is going to be a separated structure when the select with offset is implemented.

CuriousGeorgiy (approved):

src/box/memtx_tree.cc
drewdzzz: Thanks for optimization, looks nice!

src/box/memtx_tx.c
```diff
@@ -2870,25 +2870,30 @@ memtx_tx_tuple_clarify_slow(struct txn *txn, struct space *space,
 	return res;
 }
 
+/**
+ * Run @a code on @a space @a index's top stories.
+ */
+#define memtx_tx_foreach_top_story(space, index, story, code) do {	       \
```
drewdzzz, line 2992: Should be updated when #10456 is merged. Please, do not forget to write in the description of the macro that all the chains containing excluded tuples are skipped :)
I labeled this PR as `do not merge` so it won't be accidentally merged before my patch - feel free to drop the label after rebasing :)

PR #10141, opened 2024-06-17
sio: use `kern.ipc.somaxconn` for `listen()` on Mac

### LINKED ISSUE #8130: IProto or netbox rejects some connection requests? Isn't that cringe?!
I have the following test:
```Lua
--
-- Instance 1
--
-- Step 1
--
box.cfg{listen = 3313, net_msg_max = 5000}

--
-- Instance 2
--
-- Step 2
--
net = require('net.box')
conns = {}
for i = 1, 1000 do
    table.insert(conns, net.connect(3313, {wait_connected = false}))
end
bad_conn = nil
for _, c in pairs(conns) do
    c:wait_connected()
    if c.state ~= 'active' then
        bad_conn = c
        error('found bad conn')
    end
end
```
At `1.10.13-18-g588c65383` the second instance establishes all connections flawlessly. But at `2.10.0-344-g9e0da2a40` and at `2.11.0-entrypoint-854-gff7dc37fa` (latest master I had locally) some connections fail. If print any of them, can see this:
```
tarantool> bad_conn
---
- opts:
    wait_connected: false
  state: error
  error: 'connect, called on fd 231, aka 127.0.0.1:59083: Connection reset by peer'
  port: '3313'
...
```
Both client and server logs are empty. From debugger I can see those connections never make it to the server (`close()` breakpoint never hits), could be a netbox problem. But I didn't investigate too deep.

Gerold103:
listen() on Mac used to take SOMAXCONN as the backlog size. It is just 128, which is too small when connections are incoming too fast. They get rejected.

Increase of the queue size wasn't possible, because the limit was hardcoded. But now sio takes the runtime limit from kern.ipc.somaxconn sysctl setting.

One weird thing is that when set too high, it seems to have no effect, like if nothing was changed. Specifically, values above 32767 are not doing anything, even though stay visible in kern.ipc.somaxconn.

It seems listen() on Mac internally might be using 'short' or int16_t to store the queue size and it gets broken when anything above INT16_MAX is used. The code truncates the queue size to this value if the given one is too high.

Closes #8130

NO_DOC=bugfix
NO_TEST=requires root privileges for testing

### DIFF
--- a/changelogs/unreleased/gh-8130-enable-somaxconn-on-mac.md
+++ b/changelogs/unreleased/gh-8130-enable-somaxconn-on-mac.md
@@ -1,0 +1,5 @@
+## bugfix/core
+
+* Fixed the bug that on Mac the system setting `kern.ipc.somaxconn` was ignored
+  for listening sockets. Now it is used, but capped at 32367 due to how
+  `listen()` works on Mac (gh-8130).
--- a/src/lib/core/sio.c
+++ b/src/lib/core/sio.c
@@ -43,6 +43,10 @@
 #include "uri/uri.h"
 #include "errinj.h"
 
+#if TARGET_OS_DARWIN
+#include <sys/sysctl.h>
+#endif
+
 static_assert(SMALL_STATIC_SIZE > NI_MAXHOST + NI_MAXSERV,
 	      "static buffer should fit host name");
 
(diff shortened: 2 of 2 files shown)


changelogs/unreleased/gh-8130-enable-somaxconn-on-mac.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Fixed the bug that on Mac the system setting `kern.ipc.somaxconn` was ignored
+  for listening sockets. Now it is used, but capped at 32367 due to how
+  `listen()` works on Mac (gh-8130).
```
lenkis, line 5:
```suggestion
* Fixed the bug with the system setting `kern.ipc.somaxconn` ignored on Mac OS
  for listening sockets. Now it is used, but truncated at 32367 due to how
  `listen()` works on Mac (gh-8130).
```

lenkis (approved):

src/lib/core/sio.c
```diff
@@ -43,6 +43,10 @@
 #include "uri/uri.h"
 #include "errinj.h"
 
+#if TARGET_OS_DARWIN
```
Serpentian, line 46: Why does the test requires root privileges? ERRINJ, which replaces the value of `somaxconn` variable?

Serpentian (approved):
I cannot test the patch, since I have no Mac machine, but I have no objections about the code. However, I suppose, we can write the test, if we really want

src/lib/core/sio.c
Gerold103: Because increase of the systemwide limit only works under root. Having the errinj of such kind would only test the errinj itself, which isn't very handy. Testing the actual system limit really isn't possible automatically (in our test env).

changelogs/unreleased/gh-8130-enable-somaxconn-on-mac.md
Gerold103: Resolving since the PR is approved.

src/lib/core/sio.c
Gerold103: Resolving since the PR is approved.

PR #10151, opened 2024-06-21
ci: avoid sending perf stat to Influx from forks

Buristan:
This patch is the follow-up for the commit
49946a72b72884d4f84269b90e6a61 ("ci: send perf statistics to InfluxDB"). Since secrets are unavailable for fork repositories, the sending step fails due to a missed InfluxDB URL and token. This patch allows to run this step only for on push events or PRs from the main repository itself.

NO_DOC=CI
NO_CHANGELOG=CI
NO_TEST=CI


lastoCHka42 (approved):

lastoCHka42 (approved):
LGTM

mkokryashkin (approved):

PR #10175, opened 2024-06-26
config: expose experimental.config.utils.schema

### LINKED ISSUE #10117: Expose configuration validation/processing module (as experimental)
Tarantool offers a [declarative configuration way](https://github.com/tarantool/tarantool/issues/8724) since 3.0.0. The configuration has strictly defined schema, but there are sections that accept arbitrary values. These are `app.cfg.*` and `roles_cfg.*`.

An author of an application or a role defines how these values are validated and processed.

Validation and hierarchical data processing are boring tasks and it is counter-productive to implement it each time from scratch. I want to offer a tool to ease validation and processing of application/role configurations.

In fact, tarantool already has a [module](https://github.com/tarantool/tarantool/issues/8725) suitable for such purposes, but it is internal: `internal.config.utils.schema` ([source code](https://github.com/tarantool/tarantool/blob/9b817848c4e87665aa38133f97b86ff6148279a1/src/box/lua/config/utils/schema.lua)). It is used by tarantool itself for processing of the declarative configuration.

<details>
<summary>Expose as is or change something?</summary>

After a year of using of the module, I think that it is helpful for configuration processing tasks. There are things that I would implement a bit differently now. However, I prefer to expose the module as is now, because it is useful for app/role developers in its current state. Redesigning would significantly increase time needed to offer this functionality and also would raise compatibility concerns in regards to processing of the existing declarative configuration.
</details>

This task is to expose the module for users as an experimental one: `experimental.config.utils.schema`.

I have no plans to do backward-incompatible changes in it, but I would like to keep it in the experimental status for a year or so to listen a feedback from the community.

The documentation request is [here](https://github.com/tarantool/doc/issues/4279).

Totktonada:
The module is renamed from `internal.config.utils.schema` to `experimental.config.utils.schema` without changes.

It is useful for validation of configuration data in roles and applications.

Also, it provides a couple of methods that aim to simplify usual tasks around processing of hierarchical configuration data. For example,

* get/set a nested value
* apply defaults from the schema
* filter data based on annotations from the schema
* transform a hierarchical data using a function
* merge two hierarchical values
* parse environment variable according to its type in the schema

See https://github.com/tarantool/doc/issues/4279 for an in-depth description.

Fixes #10117

### DIFF
--- a/changelogs/unreleased/config-expose-config-utils-schema.md
+++ b/changelogs/unreleased/config-expose-config-utils-schema.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Expose the `experimental.config.utils.schema` built-in module to the public
+  API (gh-10117).
--- a/src/box/lua/config/cluster_config.lua
+++ b/src/box/lua/config/cluster_config.lua
@@ -1,4 +1,4 @@
-local schema = require('internal.config.utils.schema')
+local schema = require('experimental.config.utils.schema')
 local instance_config = require('internal.config.instance_config')
 local expression = require('internal.config.utils.expression')
 
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -1,4 +1,4 @@
-local schema = require('internal.config.utils.schema')
+local schema = require('experimental.config.utils.schema')
 local tarantool = require('tarantool')
 local compat = require('compat')
 local uuid = require('uuid')
--- a/src/box/lua/config/source/env.lua
+++ b/src/box/lua/config/source/env.lua
@@ -1,6 +1,6 @@
 local uri = require('uri')
 local fun = require('fun')
-local schema = require('internal.config.utils.schema')
+local schema = require('experimental.config.utils.schema')
 local tabulate = require('internal.config.utils.tabulate')
 local instance_config = require('internal.config.instance_config')
 
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -362,7 +362,7 @@ static const char *lua_sources[] = {
 	config_utils_aboard_lua,
 
 	"config/utils/schema",
-	"internal.config.utils.schema",
+	"experimental.config.utils.schema",
 	config_utils_schema_lua,
 
 	"config/utils/tabulate",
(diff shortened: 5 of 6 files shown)


changelogs/unreleased/config-expose-config-utils-schema.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Expose the `experimental.config.utils.schema` built-in module into the public
```
lenkis, line 3:
```suggestion
* Expose the `experimental.config.utils.schema` built-in module to the public
```

lenkis (approved):

changelogs/unreleased/config-expose-config-utils-schema.md
Totktonada: Fixed. Thanks!

mkokryashkin (approved):
As we've discussed offline, I went through the doc ticket with the detailed specification of the API under discussion without submerging into the topic too deeply. Among the strong points, I want to mention the resemblance of some other schema-building notations, like OpenAPI and lua-schema, in terms of terminology, typing, and general feel.

I am a bit concerned about higher-order functions like `map` and `filter` as their API seems clumpy. Moreover, I am not even sure whether they should be publicly available because they are not that easy to use with an API like that, and their use cases might be very limited (correct me if I am wrong).

Otherwise, the API is quite well thought out, so I have no strong objections.

Totktonada:
We have discussed it a bit with Maxim.

Maxim points that `:filter()` and `:map()` have complicated API, while they're likely needed in rare cases. He proposes to exclude them from the public API.

I think that a role/an application developer scenario may be the following:

1. A developer found that many options are processed similarly and that it is convenient to annotate them somehow in the schema and decouple the processing code from certain option list.
2. He/she found that the `:filter()` or the `:map()` functions are useful for this purpose and adopt a usage example from the documentation.
3. If the given algorithm is not suitable for the given task, he/she adopts code of these functions for the domain specific task.

I think that it is easier to start from some existing implementation even if it appears that a domain-specific one is necessary. However, I hope that the existing functions cover most of scenarios.

It seems that Maxim still have doubts, but he let me decide. I'm going to add these functions and hear users' feedback.

PR #10176, opened 2024-06-26
sio: fix error message displaing bind adress

### LINKED ISSUE #5925: Strange address in bind error message
When I call `box.cfg{listen = ...}` with an address to which it can't bind, I get an error message with an invalid address, not the one I specified.
```
box.cfg{listen = '8.8.8.8:3313'}
2021-03-19 22:26:59.285 [67636] iproto/101/main evio.c:396 E> binary: failed to bind on 8.8.8.8:3313: bind, called on fd 15, aka 0.0.0.0:0
```
This `aka 0.0.0.0:0` should be replaced with the actual address to which the bind was requested. It happens in `sio_bind()` function. It should not use `sio_socketname()` for the error message, but instead should convert the passed `struct sockaddr` to a string directly.

LevKats:
Now `sio_bind` function prints adress into error message directly instead of relying on `fd` used in `bind` that failed to execute.

`sio_bind` used `sio_socketname_to_buffer` for error message effectively attempting printing address bound to `fd` while there actually was an error in binding that address to that socket in the first place.

Fixes #5925

### DIFF
--- a/src/lib/core/sio.c
+++ b/src/lib/core/sio.c
@@ -55,38 +55,49 @@ static_assert(SMALL_STATIC_SIZE > NI_MAXHOST + NI_MAXSERV,
  * checks and all.
  */
 static int
-sio_socketname_to_buffer(int fd, char *buf, int size)
+sio_socketname_to_buffer(int fd,
+			 const struct sockaddr *base_addr, socklen_t addrlen,
+			 const struct sockaddr *peer_addr, socklen_t peerlen,
+			 char *buf, int size)
 {
 	int n = 0;
 	(void)n;
 	SNPRINT(n, snprintf, buf, size, "fd %d", fd);
 	if (fd < 0)
 		return 0;
-	struct sockaddr_storage addr;
-	socklen_t addrlen = sizeof(addr);
-	struct sockaddr *base_addr = (struct sockaddr *)&addr;
-	int rc = getsockname(fd, base_addr, &addrlen);
-	if (rc == 0) {
+	if (base_addr != NULL) {
 		SNPRINT(n, snprintf, buf, size, ", aka ");
 		SNPRINT(n, sio_addr_snprintf, buf, size, base_addr, addrlen);
 	}
-	addrlen = sizeof(addr);
-	rc = getpeername(fd, (struct sockaddr *) &addr, &addrlen);
-	if (rc == 0) {
+	if (peer_addr != NULL) {
 		SNPRINT(n, snprintf, buf, size, ", peer of ");
-		SNPRINT(n, sio_addr_snprintf, buf, size, base_addr, addrlen);
+		SNPRINT(n, sio_addr_snprintf, buf, size,
+			peer_addr, peerlen);
 	}
 	return 0;
 }
 
 const char *
-sio_socketname(int fd)
+sio_socketname_addr(int fd, const struct sockaddr *base_addr,
+		    socklen_t addrlen)
 {
 	/* Preserve errno */
 	int save_errno = errno;
 	int name_size = SERVICE_NAME_MAXLEN;
 	char *name = static_alloc(name_size);
-	int rc = sio_socketname_to_buffer(fd, name, name_size);
+
+	struct sockaddr_storage peer_addr_storage;
+	socklen_t peerlen = sizeof(peer_addr_storage);
+	struct sockaddr *peer_addr = (struct sockaddr *)&peer_addr_storage;
+	int rcp = getpeername(fd, peer_addr, &peerlen);
+	if (rcp != 0) {
+		peer_addr = NULL;
+	}
+
+	int rc = sio_socketname_to_buffer(fd,
+					  base_addr, addrlen,
+					  peer_addr, peerlen,
+					  name, name_size);
 	/*
 	 * Could fail only because of a bad format in snprintf, but it is not
 	 * bad, so should not fail.
@@ -96,6 +107,29 @@ sio_socketname(int fd)
 	/*
 	 * Restore the original errno, it might have been reset by
 	 * snprintf() or getsockname().
+	 */
+	errno = save_errno;
+	return name;
+}
+
+const char *
+sio_socketname(int fd)
+{
+	/* Preserve errno */
+	int save_errno = errno;
+
+	struct sockaddr_storage addr;
+	socklen_t addrlen = sizeof(addr);
+	struct sockaddr *base_addr = (struct sockaddr *)&addr;
+	int rcb = getsockname(fd, base_addr, &addrlen);
+	if (rcb != 0) {
+		base_addr = NULL;
+	}
+
+	const char *name = sio_socketname_addr(fd, base_addr, addrlen);
+	/*
+	 * Restore the original errno, it might have been reset by
+	 * getsockname().
 	 */
 	errno = save_errno;
 	return name;
@@ -245,7 +279,9 @@ sio_bind(int fd, const struct sockaddr *addr, socklen_t addrlen)
 {
 	int rc = bind(fd, addr, addrlen);
 	if (rc < 0)
-		diag_set(SocketError, sio_socketname(fd), "bind");
+		diag_set(SocketError,
+			 sio_socketname_addr(fd, addr, addrlen),
+			 "bind");
 	return rc;
 }
 
--- a/test/unit/sio.c
+++ b/test/unit/sio.c
@@ -90,6 +90,32 @@ check_uri_to_addr(void)
 }
 
 static void
+check_sio_socketname_addr(void)
+{
+	header();
+	plan(2);
+	struct sockaddr_storage base_storage;
+	struct sockaddr *base_addr = (struct sockaddr *)&base_storage;
+	socklen_t addrlen = sizeof(base_storage);
+
+	bool is_host_empty;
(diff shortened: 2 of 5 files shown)


src/lib/core/sio.c
```diff
@@ -240,12 +240,38 @@ sio_connect(int fd, const struct sockaddr *addr, socklen_t addrlen)
...
+}
+
 int
 sio_bind(int fd, const struct sockaddr *addr, socklen_t addrlen)
 {
 	int rc = bind(fd, addr, addrlen);
-	if (rc < 0)
-		diag_set(SocketError, sio_socketname(fd), "bind");
+	if (rc < 0) {
+		/* Preserve errno */
```
Serpentian, line 260: The problem here is that it's major code duplication of the `sio_socketname` function. Let's try to find a better way to implement this.

E.g. we may make `sio_socketname_to_buffer` to accept `sockaddr`, which it'll use for its `aka ...`. At the same time we should rename `sio_socketname` to smth like `sio_socketname_addr` and make it accept `sockaddr` too. This way we can use `sio_socketname_addr` in our `sio_bind` and directly pass `sockaddr` we want.

At the end we'll introduce `sio_socketname` with the same API it has now. This func will get `sockaddr` from `getsocketname` and call `sio_scoketname_addr` with `fd` and this addr.

@LevKats what do you think?

Serpentian (changes requested):
Thank you for the patch! Great work! But we gotta work on it a little bit more)

src/lib/core/sio.c
LevKats: Interesting. But doesn't these 2 cases require different handling of `fd < 0`? I mean in old API address isn't printed for negative `fd` because I guess it's considered useless.  In `sio_bind` address should be printed anyway as we don't need to report actual state of `fd` but an attempt to bind this specific address. It seems like there may be a problem with printing peers too here:
```
rc = getpeername(fd, (struct sockaddr *) &addr, &addrlen);
if (rc == 0) {
	SNPRINT(n, snprintf, buf, size, ", peer of ");
	SNPRINT(n, sio_addr_snprintf, buf, size, base_addr, addrlen);
}
```
If you had `fd` already bound to some address it would fail to bind new one but it would show old peers here which may be counterintuitive.

So we may have 2 intents while using some new API: to print actual socket state and our attempt to bind address that doesn't seem to have much in common.

Serpentian: > If you had fd already bound to some address it would fail to bind new one but it would show old peers here which may be counterintuitive.

You cannot have fd bound to some address, as this one is a newly created socket (in `evio_service_entry_bind_addr`):

https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/lib/core/evio.c#L241-L258

So, no problems with peers.

>  But doesn't these 2 cases require different handling of fd < 0? I mean in old API address isn't printed for negative fd because I guess it's considered useless. In sio_bind address should be printed anyway as we don't need to report actual state of fd but an attempt to bind this specific address.

`fd < 0` means that we failed to create socket:

https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/lib/core/sio.c#L172-L182

In such case we'll have errmsg: `binary: failed to bind on 8.8.8.8:3313: bind, called on fd -1`, which is also ok, I suppose

test/box-luatest/gh_5925_errormsg_bind_invalid_address_test.lua
```diff
@@ -0,0 +1,26 @@
...
+    g.server:drop()
+end)
+
+g.test_listen_address = function(g)
+    t.assert_error(function()
+        g.server:exec(function()
+            box.cfg{listen = '0.42.42.42:1234'}
+        end)
+    end)
+    t.assert_not_equals(
```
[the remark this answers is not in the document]
LevKats: Fixed

test/box-luatest/gh_5925_errormsg_bind_invalid_address_test.lua
```diff
@@ -0,0 +1,26 @@
...
+    g.server:start()
+end)
+
+g.after_all(function(g)
+    g.server:drop()
+end)
+
+g.test_listen_address = function(g)
+    t.assert_error(function()
+        g.server:exec(function()
```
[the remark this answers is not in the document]
LevKats: Didn't know it exists. Thanks. Fixed.

src/lib/core/sio.c
```diff
@@ -55,38 +55,39 @@ static_assert(SMALL_STATIC_SIZE > NI_MAXHOST + NI_MAXSERV,
...
-	struct sockaddr *base_addr = (struct sockaddr *)&addr;
-	int rc = getsockname(fd, base_addr, &addrlen);
-	if (rc == 0) {
+	if (base_addr != NULL) {
 		SNPRINT(n, snprintf, buf, size, ", aka ");
 		SNPRINT(n, sio_addr_snprintf, buf, size, base_addr, addrlen);
 	}
-	addrlen = sizeof(addr);
-	rc = getpeername(fd, (struct sockaddr *) &addr, &addrlen);
+	struct sockaddr_storage peer_addr_storage;
```
Serpentian, line 70: It seems to me, that this peer searching should not belong to this function. IMHO, `getpeername` should be moved out to the `sio_socketname_addr` so that `sio_socketname_to_buffer` would only dump info it gets and wouldn't try to find out anything about `fd`.

Serpentian (approved):
I don't have any major objections anymore. Let's see, what @Gerold103 will say

Gerold103 (commented):
Thanks for the patch! Good, finally the old bug is being fixed!

src/lib/core/sio.c
LevKats: Fixed

test/box-luatest/gh_5925_errormsg_bind_invalid_address_test.lua
```diff
@@ -0,0 +1,23 @@
...
+end)
+
+g.test_listen_address = function(g)
+    t.assert_error(function()
+        g.server:update_box_cfg({listen = '0.42.42.42:1234'})
+    end)
+    t.assert(
+        g.server:grep_log("aka 0.42.42.42:1234: " ..
+                          "Cannot assign requested address", 2048)
+    )
```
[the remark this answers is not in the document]
LevKats: Fixed. [check_sio_socketname_addr](https://github.com/LevKats/tarantool/blob/9b436fcd384c4ac0f3675519d42a6ae9f8e11a7f/test/unit/sio.c#L97) in test/unit/sio.c

src/lib/core/sio.c
```diff
@@ -245,7 +281,9 @@ sio_bind(int fd, const struct sockaddr *addr, socklen_t addrlen)
 {
 	int rc = bind(fd, addr, addrlen);
 	if (rc < 0)
-		diag_set(SocketError, sio_socketname(fd), "bind");
+		diag_set(SocketError,
+			 sio_socketname_addr(fd, addr, addrlen, NULL, 0),
```
Serpentian, line 285: Please, let's not pass NULL as peers. If we fail to bind to an address, e.g. due to already taken socket, we probably wanna see, which peer already taken that address. Let's move getting peer name to `sio_socketname_addr`

test/unit/sio.c
```diff
@@ -37,6 +37,10 @@
 #define UNIT_TAP_COMPATIBLE 1
 #include "unit.h"
 
+#ifndef is_str
+#define is_str(a, b, fmt, args...) ok(strcmp(a, b) == 0, fmt, ##args)
```
Serpentian, line 41: Since it's already the second place, where we need `is_str`, let's move it to the `unit.h`

test/unit/sio.c
```diff
@@ -89,6 +93,46 @@ check_uri_to_addr(void)
 	footer();
 }
 
+static int
```
Serpentian, line 96: Test function should be `void`

Serpentian (changes requested):
Thank you for the patch. The code looks way cleaner now. Almost there)

test/unit/sio.c
LevKats: Done

src/lib/core/sio.c
LevKats: Ok. Fixed

test/unit/sio.c
LevKats: Fixed

Serpentian (approved):
Thank you for the patch! Great work 💪! No more objections

Gerold103 (approved):
Thanks for working on this!

src/lib/core/sio.c
```diff
@@ -55,38 +55,49 @@ static_assert(SMALL_STATIC_SIZE > NI_MAXHOST + NI_MAXSERV,
  * checks and all.
```
sergepetrenko, line 55: Please fix the typo in the commit title:
displaing -> displaying

LevKats: Thanks :) Fixed.

sergepetrenko (approved):

sergepetrenko:
Backports
3.1 https://github.com/tarantool/tarantool/pull/10256
2.11 https://github.com/tarantool/tarantool/pull/10257

PR #10189, opened 2024-07-02
bps: introduce the bps_tree_max_mem_insert_batch function

mkostoevr:
The function returns the max amount of memory that may be requested by the tree for insertion of N consequent elements. It may be used to calculate the amount extents to reserve on batch insertion into the tree.

The batch insertion is not implemetned yet, so this function the memory is calculated for a sequence of raw inserts.

NO_DOC=internal
NO_CHANGELOG=internal


mkostoevr:
Not required anymore after tarantool/tarantool-ee#955 is merged.

Issue #3110, opened 2018-02-01
Implement an onboard script which prunes dead replica in a replica set


kostja:
It should look at box.info.replication, build a map of all alive replicas, and remove identifiers of the dead ones.

andrew-statsenko:
Really need this tool. Thanks !

OKriw:
https://www.freelists.org/post/tarantool-patches/PATCH-rfc-schema-add-possibility-to-find-and-throw-away-dead-replicas,5

kostja:
The patch is looking good but we need to finalize the api. Please check with Mons re specific module name, API, and put into a separate module in http://github.com/moonlibs org

PR #10192, opened 2024-07-03
gc: introduce WAL GC cosumers for anonymous replicas

### LINKED ISSUE #10755: Retain xlogs for anonymous replicas
WAL GC consumers are used for regular replicas to retain required xlogs, otherwise data needed by a replica that left behind too much would be deleted. Regarding anonymous replicas, the problem is not solved and such replica can easily receive a `XlogGapError` - let's use the consumers for anonymous replicas as well.

Also, since their number is unlimited and the new anonymous replicas can appear very often and disappear without any notification, we need some convenient ways to release their resources. Let's provide a new function that releases all resources occupied by a disconnected replica and let's release them automatically if we haven't heard from a replica for too long.

drewdzzz:
The patch introduces WAL GC state for anonymous replicas and a couple of ways to manage them.

Closes #10755 

EE part: https://github.com/tarantool/tarantool-ee/pull/831


src/box/lua/ctl.c
```diff
@@ -201,6 +223,7 @@ static const struct luaL_Reg lbox_ctl_lib[] = {
 	{"set_on_shutdown_timeout", lbox_ctl_set_on_shutdown_timeout},
 	{"iproto_lockdown", lbox_ctl_set_iproto_lockdown},
 	{"wal_sync", lbox_ctl_wal_sync},
+	{"deactivate_replica", lbox_ctl_deactivate_replica},
```
Totktonada, line 226: Shouldn't we also add an ability to list all of such replicas (with uuids and connected/disconnected status)? For example, add the list to `box.info` somewhere.

src/box/replication.cc
```diff
@@ -206,6 +210,71 @@ replicaset_set_sync_quorum(void)
...
+		replicaset_foreach(replica) {
+			if (!replica->anon || replica_has_connections(replica))
+				continue;
+			double last_seen = relay_last_row_time(replica->relay);
+			double unused = clock_monotonic() - last_seen;
+			double ttl = timeout - unused;
+			if (ttl <= 0) {
+				expired_replica = replica;
+				break;
+			}
```
Totktonada, line 241: If I don't miss something, `lowest_ttl` is `anon_gc_timeout` in the case, however it is possible that there is a replica with a lower TTL in the tail of the list.

Totktonada (approved):
LGTM for the YAML configuration part.

src/box/replication.cc
drewdzzz: That is why it is wrapped into a `while (true)` loop - it will stop only when there are no expired replicas left - then the whole list will be scanned and `lowest_ttl` will be correct.

src/box/lua/ctl.c
drewdzzz: Do you mean list of all deactivated replicas? Why you might need it?

AFAIU, it should work this way: monitoring observes `box.info.gc()` - if there is a stale consumer, database administrator connects, inspects the owner of the consumer and deactivates it if needed (or even removes the replica from the cluster). This scenario doesn't need such list and I can't come up with any other reasonable one. Or do I miss something?

By the way, all anonymous replicas are deleted on deactivation, so the list would contain only regular ones.

src/box/replication.cc
Totktonada: OK, got it.

src/box/lua/ctl.c
Totktonada: I mean a list of anonymous replicas with their UUIDs. IOW, what we can pass to the `box.ctl.deactivate_replica()` function.

Does `box.info.gc()` show UUID and connected/disconnected status of a replica?

src/box/lua/ctl.c
```diff
@@ -185,6 +187,26 @@ lbox_ctl_wal_sync(struct lua_State *L)
 	return 0;
 }
 
+static int
+lbox_ctl_deactivate_replica(struct lua_State *L)
```
Serpentian, line 191: I didn't review (I just read it) the RFC and may be missing smth. But why don't we want to remove the replica from the `_cluster` space since we introduce the function, which is called `deactivate_replica`. It seems, judging by its name, the function should do everything, not only deletion of the GC consumer.

We don't want user to touch system spaces at all, and removing replica from the replicaset requires that. Maybe it's time to fix that?

Moreover, I don't like, that it's possible to delete gc consumer for non anonymous replica and allow it to exist in `_cluster` space. It seems to me, that we agreed, that gc consumer must always exist for non-anon replica.

I propose to do that the default behavior: for anon - delete consumer, for non-anon - delete from `_cluster` + delete consumer. Maybe we can introduce the option to `box.ctl.deactivate_replica`, which will say, whether replica should remain in `_cluster`, but frankly speaking, I don't see a reason for such option to exist.

src/box/gc.c
```diff
@@ -913,11 +920,16 @@ gc_erase_consumer(const struct tt_uuid *uuid)
...
-	txn_stmt_on_commit(txn_current_stmt(in_txn()), on_commit);
+	if (in_txn() != NULL) {
+		/* Do actual work on commit. */
+		struct trigger *on_commit =
+			xregion_alloc_object(&in_txn()->region, struct trigger);
+		trigger_create(on_commit, gc_erase_consumer_on_commit, consumer,
+			       NULL);
+		txn_stmt_on_commit(txn_current_stmt(in_txn()), on_commit);
+	} else {
+		gc_erase_consumer_finalize(consumer);
```
Gerold103, line 931: That doesn't look right. The change is not transactional. There is time when the replica is already not in `_gc_consumers`, but in-memory its consumer still exists. Please, wrap `boxk()` into a txn when there isn't one existing already. Then you can leave the remaining code untouched. I.e. always erase on commit.

src/box/box.cc
```diff
@@ -1533,6 +1533,18 @@ box_check_replication_anon(void)
 	return anon;
```
Gerold103, line 1533: I wanted to comment a bit on the commit message statements:

> a fiber for each replica and libev timers

Why do you need libev timers for this? They are only needed really when you can't afford a fiber. If you have a fiber, you can do timers easier. Wait sleep for the given timeout and you have the timer. No?

> Creation of a fiber for each anonymous replica seems to be a meaningless overhead

Fibers cost nothing compared to overhead of a relay thread which we create for each replica. I don't think it would be noticeable at all, tbh. How much simpler would the code be with individual fibers? I mean, right now it doesn't look complex really.

src/box/replication.cc
```diff
@@ -206,6 +210,71 @@ replicaset_set_sync_quorum(void)
...
+		/*
+		 * Search for the first expired replica. Note that we cannot
+		 * just deactivate it and continue the iteration because a yield
+		 * will happen and the replicaset can be changed significantly.
+		 */
+		replicaset_foreach(replica) {
+			if (!replica->anon || replica_has_connections(replica))
+				continue;
+			double last_seen = relay_last_row_time(replica->relay);
+			double unused = clock_monotonic() - last_seen;
```
Gerold103, line 236: Relay's `last_row_time` is set as `ev_monotonic_now(loop())`. Lets use the same timer here, not `clock_monotonic()`. The difference is that the latter is an actual syscall, while the former isn't. Also libev timer can be a bit behind the real monotonic clock, because is collected only once per event loop iteration.

src/box/replication.cc
```diff
@@ -231,6 +300,12 @@ replication_init(int num_threads)
 	replicaset.healthy_count = 1;
 
 	applier_init();
+	replication_anon_gc_fiber =
```
Gerold103, line 303: Shouldn't you also shutdown this fiber in `replication_free()`?

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -227,3 +239,193 @@ g.test_fetch_snapshot_on_connected_replica = function(g)
...
+
+-- The case checks if the consumer is not expired when fetching snapshot
+-- for too long.
+g.test_expiration_long_fetch_snapshot = function(g)
+    t.tarantool.skip_if_not_debug()
+    local fiber = require('fiber')
+    local uuid = require('uuid').str()
+    g.server:exec(function() box.cfg{replication_anon_gc_timeout = 0.1} end)
+
+    g.server:exec(function()
```
Gerold103, line 303: How about merging those 2 execs into 1? They are to the same server in a row anyway.

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -227,3 +239,193 @@ g.test_fetch_snapshot_on_connected_replica = function(g)
...
+    vclock[0] = nil
+    write_subscribe(g.s, uuid, replicaset_uuid, vclock)
+    local h = socket_read(g.s)
+    t.assert_equals(h[key.REQUEST_TYPE], type.OK)
+
+    -- Keep connected to master
+    for _ = 1, 10 do
+        fiber.sleep(0.1)
+        write_ok(g.s, {[key.VCLOCK] = setmap({})})
+    end
```
Gerold103, line 384: Can we please make it a bit faster? Such a simple test takes more than a second now. Reduce the timeout like twice here, would it still cover what you need?

Gerold103 (commented):
Thanks, good stuff 💪!

src/box/lua/ctl.c
Gerold103: I agree. The name `deactivate_replica()` sounds quite generic. Like it must completely disable the replica, no matter what the replica is. Not just delete its `_gc_consumer`. Either the naming is confusing, or the actions must change to complete destruction of the replica, perhaps/

src/box/lua/ctl.c
drewdzzz: Anonymous replicas are listed in [`box.info.replication_anon()`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/replication_anon/) and the field `downstream.status` reflects if the replica is connected.

src/box/replication.cc
```diff
@@ -374,15 +374,6 @@ replica_set_id(struct replica *replica, uint32_t replica_id)
...
-	} else if (replica->anon) {
-		/*
-		 * Set replica gc on its transition from
-		 * anonymous to a normal one.
-		 */
-		assert(replica->gc == NULL);
-		replica->gc = gc_consumer_register(
-			instance_vclock, GC_CONSUMER_REPLICA,
-			&replica->uuid);
 	}
```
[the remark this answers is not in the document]
drewdzzz: The code is actually [uncovered](https://coveralls.io/builds/70710293/source?filename=src%2Fbox%2Freplication.cc#L382). After my patch the branch is started to be taken sometimes and it crashes Tarantool since we already create a consumer in `box_process_register`.
Probably, I should drop this snippet in a separate commit.

src/box/replication.cc
```diff
@@ -1362,18 +1353,13 @@ void
...
-		if (!replica->anon) {
-			gc_consumer_unregister(replica->gc);
-			replica->gc = NULL;
-		} else {
-			assert(replica->gc == NULL);
-		}
+	if (!replica->anon && replica->id == REPLICA_ID_NIL) {
+		gc_consumer_unregister(replica->gc);
+		replica->gc = NULL;
 	}
```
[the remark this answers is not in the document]
drewdzzz: > AFAIU, this function is supposed to be called when the replica is evicted from the cluster.

The function is called not only when a replica is evicted from the cluster - it is called on relay stop, IOW after each subscribe. So, if we remove GC consumer of anonymous replicas here, they will be pointless since being deleted on every disconnect.

> The fact that an anon replica being evicted from the cluster still retains the consumer

Anon replica does not participate in the cluster by definition, hence, cannot be evicted.

What actually this diff does is simply removing an assertion that anonymous replicas have no GC consumers. For non-anonymous replicas, the logic remains - we remove GC consumer in [`replica_clear_id`](https://github.com/tarantool/tarantool/blob/f69e2ae488b3620e31f1a599d8fb78a66917dbfd/src/box/replication.cc#L451) only if the replica's relay is disconnected, otherwise it is deleted here. Condition `!replica->anon && replica->id == REPLICA_ID_NIL` basically means "the replica was evicted from the cluster".

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -0,0 +1,229 @@
...
+
+-- Helper that checks if IPROTO_SUBSCRIBE correctly updates WAL GC state
+local test_check_subscribe = function(g, uuid)
+    local replicaset_uuid = g.server:eval("return box.info.replicaset.uuid")
+    local vclock = g.server:get_vclock()
+    vclock[0] = nil
+    write_subscribe(g.s, uuid, replicaset_uuid, vclock)
+    local h = socket_read(g.s)
+    t.assert_equals(h[key.REQUEST_TYPE], type.OK)
+
```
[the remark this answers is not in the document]
drewdzzz: These are two helper functions - one to check fetch snapshot isolated and the second one is to check subscribe isolated. Below the test cases themselves:

```lua
g.test_fetch_snapshot = function(g)
    local uuid = luuid.str()
    test_check_fetch_snapshot(g, uuid)
end

g.test_subscribe = function(g)
    local uuid = luuid.str()
    test_check_subscribe(g, uuid)
end

-- Typical scenario: fetch_snapshot -> subscribe
g.test_fetch_snapshot_subscribe = function(g)
    local uuid = luuid.str()
    test_check_fetch_snapshot(g, uuid)
    test_check_subscribe(g, uuid)
end
```

Note that I restart the server in `after_each`/`before_each` hooks, so here I have 3 isolated test cases.

src/box/box.cc
```diff
@@ -4448,7 +4469,7 @@ box_process_fetch_snapshot(struct iostream *io,
 	/* Send the snapshot data to the instance. */
 	struct vclock start_vclock;
 	relay_initial_join(io, header->sync, &start_vclock, req.version_id,
-			   cursor_ptr);
+			   cursor_ptr, replica);
```
locker, line 4541: I don't understand why we need to pass `replica` to `relay_initial_join`. We don't do it in `box_process_join` although the replica may exist there as well.

src/box/replication.cc
```diff
@@ -288,7 +288,7 @@ replica_check_id(uint32_t replica_id)
 static bool
 replica_is_orphan(struct replica *replica)
 {
-	return replica->id == REPLICA_ID_NIL &&
+	return replica->id == REPLICA_ID_NIL && replica->gc == NULL &&
```
locker, line 366: AFAIU any anonymous replica has a garbage collector state now so shouldn't it be `&& !replica->anon`? Anyway, the comment needs to be updated.

src/box/box.cc
```diff
@@ -4434,6 +4434,33 @@ box_process_fetch_snapshot(struct iostream *io,
...
+	 * Find checkpoint for checkpoint join. If replica didn't request
+	 * specific one, take the newest one. Initialize checkpoint cursor
+	 * with chosen checkpoint and use its vclock for WAL GC consumer.
+	 * If requested checkpoint is not found, raise an error.
+	 */
+	struct checkpoint_cursor cursor;
+	struct checkpoint_cursor *cursor_ptr = NULL;
+	struct gc_checkpoint *checkpoint = NULL;
+	const struct vclock *gc_vclock = instance_vclock;
+	if (req.is_checkpoint_join) {
```
locker, line 4472: This new branch and the one below should be covered in CE tests.

src/box/box.cc
```diff
@@ -4449,24 +4476,41 @@ box_process_fetch_snapshot(struct iostream *io,
...
+	});
+	gc_checkpoint_ref_guard.is_active = false;
+	if (checkpoint != NULL) {
+		if (replica == NULL) {
+			gc_ref_checkpoint(checkpoint, &checkpoint_ref,
+					  "checkpoint join");
+			gc_checkpoint_ref_guard.is_active = true;
+		} else {
+			replica->gc_checkpoint_ref =
+				xalloc_object(struct gc_checkpoint_ref);
```
locker, line 4529: Let's make `gc_ref_checkpoint` allocate an object for us, like `gc_consumer_register` does?

src/box/relay.cc
```diff
@@ -493,6 +493,9 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 	ctx.send_meta = replica_version_id > 0;
 	ctx.vclock = vclock;
 	ctx.cursor = cursor;
+	/* If cursor is passed, we should respond with its vclock. */
+	if (cursor != NULL)
+		ctx.vclock = cursor->vclock;
```
locker, line 498:
```C
ctx->vclock = cursor != NULL ? cursor->vclock : vclock;
```

src/box/lua/ctl.c
```diff
@@ -185,6 +187,26 @@ lbox_ctl_wal_sync(struct lua_State *L)
...
+{
+	const char *uuid_str = luaT_checkstring(L, 1);
+	struct tt_uuid uuid;
+	if (tt_uuid_from_string(uuid_str, &uuid) != 0)
+		luaT_error(L);
```
locker, line 196: `tt_uuid_from_string` doesn't set diag. Please use `IllegalParams` error and add a test for invalid arguments.

src/box/lua/ctl.c
```diff
@@ -185,6 +187,26 @@ lbox_ctl_wal_sync(struct lua_State *L)
...
+	struct replica *replica = replica_by_uuid(&uuid);
+	if (replica == NULL)
+		luaL_error(L, "Cannot deactivate replica: does not exist");
+	if (tt_uuid_is_equal(&uuid, &INSTANCE_UUID))
+		luaL_error(L, "Cannot deactivate replica: deactivation of "
+			   "self is not allowed");
+	if (relay_get_state(replica->relay) == RELAY_FOLLOW)
+		luaL_error(L, "Cannot deactivate replica: it is connected");
```
locker, line 204: Please use `diag_set` with a proper error code. Better move these checks to `replica_deactivate`.

src/box/box.cc
drewdzzz: Hm, actually we should do it there as well - it's my mistake. The point is we determine if a replica is connected by the state of its relay so we should replica's relay as often as possible so that another connection with the same UUID will be rejected.

After your comment I actually realised that this protection is like a sieve. If another replica will try to join with the same UUID when relay is not used (on `_cluster:insert`, for example), master will fail. Probably, we should do a better protection against multiple connections? For example, something like `RegionGuard` but for replicas that will be used in all four replication endpoints.

Maybe the problem with join is not so critical since only regular replicas use it so the protocol is kind of "closed". But what's about anonymous replicas, anyone can write its own and use the protocol in his way. For example, one can send a new fetch_snapshot request with the same UUID right after the previous one failed because of the network failure. However, Tarantool may notice the disconnection a bit later so it's actually a duplicate connection from Tarantool's point of view - and not rejected duplicate connection is likely to crash the Tarantool.

src/box/lua/ctl.c
drewdzzz: Added to the doc request:

> By the way, there is a function `box.info.replication_anon()` that can be used
to list all anonymous replicas so one can deactivate them with the newly
introduced function - we should mention it somewhere around.

src/box/box.cc
```diff
@@ -4434,6 +4434,27 @@ box_process_fetch_snapshot(struct iostream *io,
...
+				  "duplicate connection with the same replica "
+				  "UUID");
+		}
+		if (replica->gc != NULL)
+			gc_consumer_unregister(replica->gc);
+		replica->gc = gc_consumer_register(
+			instance_vclock, GC_CONSUMER_REPLICA, &replica->uuid);
+		/* Persist consumer before sending data. */
+		if (gc_consumer_persist(replica->gc) != 0)
+			diag_raise();
```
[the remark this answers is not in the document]
drewdzzz: Done, I created a new helper `box_connect_replica` that checks for a duplicate replica connection and handles WAL GC.

I wasn't sure where to put the new helper so I put it to `box.cc`.

src/box/box.cc
```diff
@@ -4434,6 +4434,27 @@ box_process_fetch_snapshot(struct iostream *io,
...
+		replica = replica_by_uuid(&req.instance_uuid);
+		if (replica == NULL)
+			replica = replicaset_add_anon(&req.instance_uuid);
+
+		/* Don't allow multiple relays for the same replica */
+		if (relay_get_state(replica->relay) == RELAY_FOLLOW) {
+			tnt_raise(ClientError, ER_CFG, "replication",
+				  "duplicate connection with the same replica "
+				  "UUID");
+		}
```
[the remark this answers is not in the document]
drewdzzz: Moved the logic to the new helper `box_connect_replica`. By the way, we somewhy didn't check for duplicate relay on JOIN 😄 

Also, I enhanced the protection against duplicate connection, see [the comment](https://github.com/tarantool/tarantool/pull/10192#discussion_r1834766495).

src/box/relay.cc
```diff
@@ -462,16 +462,25 @@ relay_check_flush(struct relay *relay)
...
+	if (replica != NULL)
+		replica_on_relay_follow(replica);
 	xrow_stream_create(&relay->xrow_stream);
 	auto relay_guard = make_scoped_guard([=] {
 		xrow_stream_destroy(&relay->xrow_stream);
 		relay_stop(relay);
-		relay_delete(relay);
+		if (replica != NULL)
+			replica_on_relay_stop(replica);
+		if (replica == NULL)
```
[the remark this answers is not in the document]
drewdzzz: Outdated - I don't pass replica to `relay_initial_join` anymore.

src/box/relay.h
```diff
@@ -134,7 +134,7 @@ relay_cancel(struct relay *relay);
 void
 relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 		   uint32_t replica_version_id,
-		   struct checkpoint_cursor *cursor);
+		   struct checkpoint_cursor *cursor, struct replica *replica);
```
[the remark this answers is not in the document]
drewdzzz: Outdated - I don't pass replica to relay_initial_join anymore.

src/box/replication.cc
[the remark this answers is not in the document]
drewdzzz: Done - the snippet is dropped in a separate commit.

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -0,0 +1,229 @@
...
+-- Hint that table must be encoded as MP_MAP
+local function setmap(map)
+    return setmetatable(map, {__serialize = 'map'})
+end
+
+--
+-- Helpers for IPROTO protocol communication via raw socket
+--
+
+local function socket_connect(server)
```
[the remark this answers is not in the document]
drewdzzz: Copy-pasted (and slightly modified) from checkpoint join test. But thanks anyway ❤️

src/box/gc.c
drewdzzz: Fixed - now I always delete the in-memory consumer in an on_commit trigger.

src/box/replication.cc
drewdzzz: Fiber works extremely rare, so I used `clock_monotonic` here intentionally - to be more precise at a price of system call. But it doesn't really matter for me - let's use libev clock for consistency.
Replaced with `ev_monotonic_now(loop())`.

src/box/replication.cc
drewdzzz: Oh... Thanks! Shutdown it in `replication_shutdown`.

test/replication-luatest/persistent_gc_anon_test.lua
drewdzzz: Yep, somehow missed this place. Fixed!

test/replication-luatest/persistent_gc_anon_test.lua
drewdzzz: Reduced all the timeouts (including `replication_anon_gc_timeout` and number of iterations.

src/box/box.cc
drewdzzz: 1. Removed the replica from `relay_initial_join` (reverted my change, actually).
2. Introduced `ReplicaConnectionGuard` for better protection against duplicate connections.

src/box/box.cc
drewdzzz: Thanks, done!

src/box/box.cc
drewdzzz: Yep, good idea, thanks. Done in a separate commit.

src/box/relay.cc
drewdzzz: Oh, thanks! Fixed.

src/box/lua/ctl.c
drewdzzz: Thanks! Fixed and added tests with invalid UUID.

src/box/lua/ctl.c
drewdzzz: > Please use diag_set with a proper error code.

Didn't find any matching error codes. Anyway, it's a Lua function for database administrator that will be actually used manually - does it worth adding a new error code?

src/box/box.cc
drewdzzz: > Why do you need libev timers for this?

That's how we decided to implement expiration in the design doc. Turned out that it's more complicated that a separate fiber.

> Fibers cost nothing compared to overhead of a relay thread.

Relay works only for connected replicas and expiration fibers are needed for all ones. But yep, the overhead still seems not critical.

> How much simpler would the code be with individual fibers?

As you already said, each fiber would simply do `sleep` + `gc_erase_consumer`. On the other hand, we would need to create such fibers on every replica disconnection - taking that into account, separate fibers seem to be even a bit more complicated. Added a note about it to the commit message.

src/box/replication.cc
drewdzzz: > Anyway, the comment needs to be updated.

Thanks, done.

> AFAIU any anonymous replica has a garbage collector state now so shouldn't it be && !replica->anon?

Why? Then anonymous replica will always be non-orphan and won't be deleted at all.

locker: Anonymous replicas should be deleted only on timeout, i.e. they should never be considered orphan, no?

drewdzzz: Fiber `replication_anon_gc` calls general-purpose `replica_deactivate` function that deletes the replica only if it is orphan (since this function can be called manually for regular replicas). We could do something like
```c
if (replica_is_orphan(replica) || replica->anon) {
    ...
}
```
but it looks weird, we always use delete-replica-if-orphan construction without additional conditions.

locker: How can an anonymous replica become orphan? Its gc state should be deleted but it can only happen in `replication_anon_gc` or do I miss somehing?

drewdzzz:
@locker @Gerold103
Thanks for your review! While addressing your comments, I recognized some flaws and the PR was significantly changed. Major updates:
1. Logic checking duplicate replica connection and re-creating WAL GC consumer of replica was moved to a separate helper `box_connect_replica`. For the sake of simplicity (and to solve another problem), now we register a replica as anonymous before registration. See commit `replication: provide anonymous replicas with WAL GC consumers` for details.
2. It turned out that our check of duplicate replica connection (`relay->state != RELAY_FOLLOW`) is poor - replica can be connected but not using a relay (for example, on initial join, or when persisting a WAL GC consumer). The commit `replication: provide anonymous replicas with WAL GC consumers` introduces a new guard that improves the protection. I tried to add the check in a separate commit, but it's too tightly coupled with anonymous replica's WAL GC commit, so I merged them.
3. I forgot to handle WAL GC consumers of anonymous replicas after restart. Added the logic, a test case and populated the commit message in `replication: add expiration for anonymous replicas`.

src/box/lua/ctl.c
drewdzzz: > I don't like, that it's possible to delete gc consumer for non anonymous replica and allow it to exist in _cluster space

[Here](https://github.com/tarantool/tarantool/pull/9960#discussion_r1611568006) and [here](https://github.com/tarantool/tarantool/pull/9960#discussion_r1611600335) are comments on motivation of leaving regular replicas without its WAL GC state. It doesn't really matter to me to allow or disallow removing WAL GC state only, but I would add such opportunity since there is a real demand from a real user.

> The name deactivate_replica() sounds quite generic. Like it must completely disable the replica, no matter what the replica is.

Yeah, I agree, we should think about a better name, as you suggested during the design doc review. Do you have any ideas? I could suggest:
* Names associated with WAL: `replica_release_xlogs`/`replica_release_wal`/`replica_clear_wal_gc`.
* Names associated with connection: something like `replica_back_off` - the replica leaves us, but not completely. All such names sound weird somewhy. Also, I came up with `replica_suspend`, but it sounds like "forbid to connect".

> we can introduce the option to `box.ctl.deactivate_replica`

I don't like that the same function can be used to "remove local WAL GC consumer" and to "completely drop the replica and its WAL GC state from all replicas".  These two operations are completely different and have very different scale of consequences.

src/box/box.cc
```diff
@@ -1529,6 +1529,18 @@ box_check_replication_anon(void)
 	return anon;
 }
 
+static double
+box_check_replication_anon_gc_timeout(void)
```
Serpentian, line 1533: Why did we decide to go with `replication_anon_gc_timeout` and not `wal_anon_gc_timeout`. It seems it has little to do with replication but it's definitely related to gc. And we already have `wal_retention_period`

src/box/replication.cc
```diff
@@ -206,6 +210,71 @@ replicaset_set_sync_quorum(void)
...
 
+/**
+ * Deactivates all expired anonymous replicas. They are expected to be
+ * deleted on deactivation as orphan.
+ * Returns lowest time-to-live among remaining anonymous replicas via
+ * `lowest_ttl`. If there are no anonymous replicas left or the function has
+ * failed to deactivate a replica, `replication_anon_gc_timeout` is returned.
+ */
+static void
+replication_anon_gc(double *lowest_ttl)
```
Serpentian, line 221: Why is this done in `replication.cc` and not in `gc_cleanup_fiber`. It seems the latter is the better place for this, we anyway don't need to remove consumers precisely after timeout and gc fiber's waken up every `gc_advance`, very often

src/box/lua/ctl.c
drewdzzz: > Let's not introduce new error code, we already have plenty of them

Is this a problem?

> nobody parses them in code

Where this conclusion comes from?

> We can use smth like `box.error.ILLEGAL_PARAMS`

We are going to move this checks to the internal `replica_deactivate` function. In this case, `ILLEGAL_PARAMS` doesn't seem appropriate.

Serpentian: > Where this conclusion comes from?

We discussed that with @Gerold103 and Vladislav Grubov in the scope of timeout option for synchronous transactions. It was said, that we have 300 error codes and nobody parses them. Just retry in case of any error and that's it

> We are going to move this checks to the internal replica_deactivate function.

Why? We already have assertion there

src/box/box.cc
```diff
@@ -4412,6 +4412,77 @@ box_process_auth(struct auth_request *request,
...
+		    struct replica **out)
+{
+	if (tt_uuid_is_nil(uuid)) {
+		*out = NULL;
+		return ReplicaConnectionGuard(NULL);
+	}
+
+	struct replica *replica = replica_by_uuid(uuid);
+	if (replica == NULL)
+		replica = replicaset_add_anon(uuid);
```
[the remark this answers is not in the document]
drewdzzz: Well-well, someone hasn't read the commit message 😄 

```console
Important to note that the commit changes the structure of REGISTER and
JOIN endpoints. Before the commit they used the following steps:
1. Retrieve the replica object and clear its WAL GC state, if any.
2. Register a WAL GC consumer but not put it to the replica object.
3. After successful registration, put the created consumer to the replica.
The problem here is if the registration fails, replica loses its
in-memory consumer, so we will never reach if from anonymous replica
garbage collector and its persistent WAL GC consumer will linger in the
system space. To solve this problem, let's register the new WAL GC consumer
and put it to the replica right after deleting the old one. If there is no
corresponding replica object in memory, let's create an anonymous one.
So actual behavior change is that any replica is actually registered as
an anonymous one before the registration. The good thing is we are going
to implement a garbage collector for disconnected anonymous replicas, so
failed to register replicas will be deleted eventually.
```

Also, the function's description (see pt.3):
```c++
/**
 * A helper for replication endpoints to handle a connecting replica.
 * 1. Checks if the replica doesn't have another connection.
 * 2. Creates an anonymous replica object if it doesn't exist.
 * 3. Handles WAL GC state of the replica. It's important that if the replica
 *    already has a WAL GC consumer, it is re-created and never will be deleted
 *    so the replica won't lose its consumer, even in the case of an error.
 * 4. Returns a guard that protects against duplicate replica connection.
 */
NODISCARD static ReplicaConnectionGuard
box_connect_replica(const struct tt_uuid *uuid, const struct vclock *gc_vclock,
		    struct replica **out);
```

src/box/box.cc
```diff
@@ -4690,19 +4744,8 @@ box_process_join(struct iostream *io, const struct xrow_header *header)
...
-		replica->gc = NULL;
-	}
-
-	/*
-	 * Register the replica as a WAL consumer so that
-	 * it can resume FINAL JOIN where INITIAL JOIN ends.
-	 */
-	struct gc_consumer *gc = gc_consumer_register(
-		instance_vclock, GC_CONSUMER_REPLICA, &req.instance_uuid);
-	auto gc_guard = make_scoped_guard([&] { gc_consumer_unregister(gc); });
```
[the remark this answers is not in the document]
drewdzzz: That's exactly what we want - see https://github.com/tarantool/tarantool/pull/10192#discussion_r1839723176

test/replication-luatest/join_test.lua
```diff
@@ -107,3 +107,38 @@ g.test_raft_state_during_meta_join = function(lg)
...
+    })
+    local replica2 = server:new({
+        alias = 'replica2',
+        box_cfg = box_cfg,
+    })
+    replica1:start({wait_until_ready = false})
+    replica2:start({wait_until_ready = false})
+    local msg = ('joining replica'):format(replica_uuid):gsub('%-', '%%-')
+    t.helpers.retrying({}, function()
+        t.assert(lg.master:grep_log(msg))
```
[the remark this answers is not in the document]
drewdzzz: As you can see, I not only introduce a new case here but also turn `before_all` hook into `before_each`. The reason of the failure you sent is that the cases are not isolated so old cluster topology breaks it.

> Moreover, grep_log doesn't checks, that there's exactly one expected log message in logs

I don't know how to check if a string appears in logs exactly once.

> Isn't it a duplicate of `g.test_fetch_snapshot_on_connected_replica`.

The idea is the same, but this case covers IPROTO_JOIN.

> We don't assert, that error on one of replicas happened, just that master starts to join replica.

The second replica infinitely tries to join but it can't because another replica with the same UUID is connected. Until the join succeeds, any attempt to do `replica2:exec` fails with "net box not connected" error.

src/box/replication.cc
drewdzzz: Conceptually, it's garbage collector for anonymous replicas themselves, not only their WAL GC consumers.

src/box/replication.cc
drewdzzz: Discussed f2f, got your point - now all anonymous replicas are not orphan.

src/box/lua/ctl.c
locker: > Let's not introduce new error code, we already have plenty of them and nobody parses them in code. We can use smth like `box.error.ILLEGAL_PARAMS` here, as connected replica is actually an illegal parameter

It isn't an illegal parameter. All function parameters are legal. The replica can't be deactivated because it's busy.



> We discussed that with @Gerold103 and Vladislav Grubov in the scope of timeout option for synchronous transactions.

Having different kinds of timeout errors seems to be an overkill, I agree. In this case I don't see an error we could reuse.

src/box/box.cc
Serpentian: > We decided to implement garbage collector for anonymous replicas, not WAL GC consumers. The name was changed correspondingly.

The name should not depend on internal implementation. `wal_retention_period` is a number of seconds to save xlogs after last use. `replication_anon_gc_timeout` does the same thing: number of seconds to save xlogs for anon replica after its disconnection. From the user point of view, it's non-obvious, why there're two almost the same options, which are named very differently.

drewdzzz: No, `replication_anon_gc_timeout` - timeout after which we delete an anonymous replica.

src/box/box.cc
```diff
@@ -4434,6 +4505,11 @@ box_process_fetch_snapshot(struct iostream *io,
 			  "wal_mode = 'none'");
 	}
```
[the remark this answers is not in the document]
drewdzzz: `box_connect_replica` is supposed to do replica-specific checks. But, actually, read access can be considered as a replica specific check, and there is no point to connect replica without WAL, so yeah, moved it to the helper.

But the way, had to update `replication-py/cluster.test` test during to check order changes.

src/box/lua/ctl.c
drewdzzz: Added a special error code `ER_REPLICA_DEACTIVATE` for the function. The error will be updated when we will choose a new name for `box.ctl.deactivate_replica`.

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -0,0 +1,259 @@
...
+        box.schema.space.create('test')
+        box.space.test:create_index('pk')
+        box.space.test:insert{0, 0}
+    end)
+    g.s = socket_connect(g.server)
+end)
+
+g.after_each(function(g)
+    g.s:close()
+    g.server:stop()
```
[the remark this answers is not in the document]
drewdzzz: Thanks, fixed.

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -0,0 +1,259 @@
...
+    read_fetch_snapshot_response(g)
+
+    g.server:exec(function()
+        t.assert_equals(box.info.gc().consumers, {})
+        t.assert_equals(box.space._gc_consumers:select{}, {})
+    end)
+end
+
+-- Helper that checks if IPROTO_FETCH_SNAPSHOT correctly updates WAL GC state
+local test_check_fetch_snapshot = function(g, uuid)
```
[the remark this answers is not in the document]
drewdzzz: Hm, don't find it misleading personally. But nevermind - made them `local function ...()`.

test/replication-luatest/persistent_gc_anon_test.lua
```diff
@@ -276,3 +288,215 @@ g.test_checkpoint_join = function(g)
...
+    read_fetch_snapshot_response(g)
+
+    -- Set WAL errinj and for failure message in logs
+    g.server:exec(function()
+        box.error.injection.set('ERRINJ_WAL_WRITE', true)
+    end)
+    t.helpers.retrying({}, function()
+        local msg = string.format(EXPIRATION_FAILURE_MSG_PATTERN,
+            escape_uuid(uuid))
+        t.assert(g.server:grep_log(msg))
```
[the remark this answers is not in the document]
drewdzzz: Oh, many thanks! I've actually checked it locally and it was stable, but increasing worker num from 24 to 50 revealed the problem.

src/box/lua/ctl.c
drewdzzz: > My suggestion is `drop_gc_consumer`

The function also drops checkpoint reference after checkpoint join.
Also, now we have the only type of WAL GC consumers - `"replica"`. However, in future some new types could be added (at least, consumers are designed in this way), so it would be better to reflect the the function manipulates only replicas.

drewdzzz: A couple of thoughts about the function:
1. Imagine you need to drop WAL GC state of a replica from the whole replicaset - in order to do this with the new function, you have to call it on all alive replicas and wait for dead ones to call it on them too. Do we have any mechanisms to call a function on all replicas eventually? If we have, will it safe to use them? (For example, the function can be called too late, after a replica appeared and disconnected again).
2. Is there any case when one needs to drop WAL GC state only on one replica?

src/box/relay.cc
```diff
@@ -482,8 +482,9 @@ relay_initial_join(struct iostream *io, uint64_t sync, struct vclock *vclock,
 	 * All these versions know of additional META stage of initial join.
 	 */
 	ctx.send_meta = replica_version_id > 0;
-	ctx.vclock = vclock;
 	ctx.cursor = cursor;
+	/* If cursor is passed, we should respond with its vclock. */
+	ctx.vclock = cursor != NULL ? cursor->vclock : vclock;
```
Gerold103, line 487: I see what. But why?

src/box/gc.c
```diff
@@ -903,14 +903,17 @@ gc_erase_consumer_on_commit(struct trigger *trigger, void *event)
...
 		return 0;
 	}
 
+	bool is_autocommit = in_txn() == NULL;
+	if (is_autocommit && txn_begin() == NULL)
+		return -1;
+
 	if (boxk(IPROTO_DELETE, BOX_GC_CONSUMERS_ID, "[%s]",
 		 tt_uuid_str(uuid)) != 0)
 		return -1;
```
Gerold103, line 920: Need to roll the autocommit txn back on failure. Otherwise the transaction will stay in the fiber and won't allow to start any new transaction.

src/box/gc.c
```diff
@@ -919,8 +922,22 @@ gc_erase_consumer(const struct tt_uuid *uuid)
...
+		 * back.
+		 */
+		txn_on_commit(in_txn(), on_commit);
+	} else {
+		/*
+		 * If external txn is used, the statement can be rolled back
+		 * before its commit, so we should set the trigger to the
+		 * current statement, not the txn itself.
+		 */
+		txn_stmt_on_commit(txn_current_stmt(in_txn()), on_commit);
```
Gerold103, line 939: Why not simply set it on the statement always?

src/box/replication.cc
```diff
@@ -43,6 +43,7 @@
 #include "relay.h"
 #include "sio.h"
 #include "tweaks.h"
+#include "clock.h"
```
Gerold103, line 46: No need for this header anymore.

src/box/replication.cc
```diff
@@ -206,6 +210,71 @@ replicaset_set_sync_quorum(void)
...
+	while (true) {
+		*lowest_ttl = timeout;
+		struct replica *expired_replica = NULL;
+		/*
+		 * Search for the first expired replica. Note that we cannot
+		 * just deactivate it and continue the iteration because a yield
+		 * will happen and the replicaset can be changed significantly.
+		 */
+		replicaset_foreach(replica) {
+			if (!replica->anon || replica_has_connections(replica))
```
Gerold103, line 233: I am getting lost in the connection states.
* `replica->is_connected`.
* `relay_get_state(replica->relay) == RELAY_FOLLOW`.
* `replica->applier != NULL`.

Interestingly, `replica_has_connections()` doesn't check `replica->is_connected`. Why?

Gerold103 (commented):
Absolutely titanic work for a long awaited feature.

src/box/replication.cc
Gerold103: Hm. I think some confusion is coming from the option name - `replication_anon_gc_timeout`. It makes us think like it is about GC state of the anon replicas, not about the anon replicas themselves. Perhaps we could remove the uncertainty by naming it `replication_anon_deactivation_timeout`?

src/box/box.cc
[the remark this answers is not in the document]
Gerold103: Funny how we fallback to C++ sometimes, accepting its superiority in cases like this, and how much do we need in C to mimic even just trivial destructors. But as soon as there is a suggestion to completely switch to C++, some people would cry from phantom pains of imaginary performance drops 🤔. @sergepetrenko, just the simplest C++ usage as "C with classes" would allow to drop so much code ...

src/box/replication.cc
```diff
@@ -487,8 +491,10 @@ bool
 replica_has_connections(const struct replica *replica)
 {
 	assert(replica->relay != NULL);
-	return relay_get_state(replica->relay) == RELAY_FOLLOW ||
-	       replica->applier != NULL;
+	/* Relay is expected to be active only for connected replicas. */
+	assert(relay_get_state(replica->relay) != RELAY_FOLLOW ||
+	       replica->is_connected);
```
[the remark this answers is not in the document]
drewdzzz: > Is it possible that `relay_state == FOLLOW` and not `replica->is_connected`? AFAIU, it is created for each relay-requiring replica even before the relay itself, and remains set until the relay's TCP connection is fully dead.

AFAICS, no, it's not possible. And that's why this assertion checks that relay has `RELAY_FOLLOW` state only if its replica is connected.

src/box/replication.cc
drewdzzz: > Interestingly, `replica_has_connections()` doesn't check `replica->is_connected`. Why?

Actually, it checks:
```c
bool
replica_has_connections(const struct replica *replica)
{
	assert(replica->relay != NULL);
	/* Relay is expected to be active only for connected replicas. */
	assert(relay_get_state(replica->relay) != RELAY_FOLLOW ||
	       replica->is_connected);
	return replica->is_connected || replica->applier != NULL;
}
```

src/box/replication.cc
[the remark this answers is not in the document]
drewdzzz: By the way, wrote about it in `is_connected` flag description.

src/box/replication.h
```diff
@@ -402,6 +402,8 @@ struct replica {
 	bool is_relay_healthy;
 	/** Whether there is an applier subscribed to this replica. */
 	bool is_applier_healthy;
+	/** Whether the replica is connected right now. */
+	bool is_connected;
```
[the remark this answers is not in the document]
drewdzzz: Renamed it to `has_incoming_connection` and left a comment with elaboration about its difference from `is_relay_healthy`.

src/box/relay.cc
drewdzzz: Elaborated here:
```c
/*
 * If cursor is passed, we should respond with its vclock because
 * it will actually be vclock of a last sent row.
 */
```

src/box/gc.c
drewdzzz: Because in auto commit mode we have no "current statement" - a statement that is being processed right now. In other cases we have it because we call the function only on `_cluster:replace`, so the replace in `_cluster` is actually a current statement.

But after your question I realised that it looks weird, so I introduced a new helper `txn_last_stmt` and use it here to always put the `on_commit` trigger to the actual `_gc_consumers:delete` statement.

src/box/replication.cc
drewdzzz: Thanks, dropped it.

src/box/replication.cc
drewdzzz: >  I think some confusion is coming from the option name - `replication_anon_gc_timeout`.

Yeah, since we call WAL GC as GC, any abbreviation of "garbage collector" makes us think about xlogs if it's somehow related to it :persevere:

I've asked Vlad Grubov and I'm waiting for his answer. Now I suggest an alternative name - `replication_anon_ttl`.

src/box/gc.c
drewdzzz: Oh, thanks!
Fixed and covered with a test case.

src/box/box.cc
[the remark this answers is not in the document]
drewdzzz: Have you ever seen how we [emulate](https://github.com/tarantool/tarantool/blob/master/src/lib/salad/bps_tree.h) C++ templates in B-tree? Brings me to tears faster than any Ernest Hemingway story.

src/box/replication.h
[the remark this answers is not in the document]
Gerold103: Nope, that would make it inconsistent with `is_applier_healthy`. The flags are probably ok to have, but then it is confusing that we also directly check the relay status with `relay_state == FOLLOW`. Perhaps it would be a good idea to drop those checks completely, and introduce a new `enum replica_relay_state` and `enum replica_applier_state` which are stored and updated only in `replica` object, and replace all its flags. For example, when joining, we have `REPLICA_RELAY_CONNECTED`, and when subscribed - `REPLICA_RELAY_STREAMING`, etc.

Not necessary to do in this patch though.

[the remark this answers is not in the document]
Gerold103: Another idea - why don't we drop `replica->relay` when it is unused? Right now replica always has a relay, whether it is started or not. But for applier it is different - we set it to `NULL` when the replica has no appliers.

I am wondering why not do the same with `replica->relay`. Make it `NULL` when no connection. And create it for each incoming connection. Then the flag could be replaced with `replica->relay != NULL`.

src/box/lua/ctl.c
Gerold103: I am joining here that the function's name right now is very confusing. Deactivation can mean many things. Does it delete from `_cluster`? Does it abort the connection on active replicas? When I first saw this function, my expectation was that it disconnects and deletes the replica completely, no matter its state.

I suggest to find a name which is harder to misunderstand. Like `box.ctl.dead_replica_gc()`, `box.ctl.prune_replica()`, `box.ctl.replica_gc()`, or alike. `gc` emphasizes both GC subsystem cleanup and actual deletion of the replica when it has no links anymore (for anon disconnected replicas it is so). `prune` emphasizes that it removes stale replica's state (alongside removing the replica itself if GC was its only state (having an "id" is a state too - this is why non-anon replicas aren't deleted but their state GC info still is dropped)). Maybe there are more names? Again, I find it particularly useful to ask customers what do they think of a name.

Gerold103:
@drewdzzz I have no new comments, good work! Also thanks to @Serpentian for his notes. Now we just need to finalize those small things.

src/box/replication.cc
drewdzzz: Renamed to `replication_anon_ttl`.

src/box/replication.h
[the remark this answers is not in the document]
drewdzzz: > I am wondering why not do the same with `replica->relay`. Make it NULL when no connection.

Good idea! However, some minor questions appear. For example, when `replica->relay == NULL`, we don't know if it is equivalent of `RELAY_STOPPED` or `RELAY_OFF` status (it affects `box.info.replication` entry). Seems like relay rework is a perfect place to rework relay states as well, so I'd leave the `has_incoming_connection` flag for now. @locker, what do you think about it?

src/box/lua/ctl.c
sergepetrenko: I like `box.ctl.replica_gc()` and `box.ctl.cleanup_replica()`.

`prune_replica` has strong connotations with removing the replica from _cluster as well. But maybe that's just for me because of an old ticket #3110 (prune dead replicas from a replica set), where we wanted to remove the replicas from _cluster.

I think we shouldn't mention WAL GC consumer in the function name, because we agreed not to mention it in `replication_anon_ttl`. That is, we do not "drop consumer", we "cleanup additional info regarding the replica". Maybe the replica itself as well if it is anonymous.

Serpentian: > That is, we do not "drop consumer", we "cleanup additional info regarding the replica"

We cleanup additional info, which is invisible to the user. After "that function" (whatever it's named) is called, replica is still in `box.info.replication`, it's still in `_cluster` space. The only difference from the user side is that now `box.info.gc().consumers` and `box.info.gc().checkpoints` don't mention that replica. So, at least for now, all this function does is removing all gc related stuff. And I suppose, it should mention GC subsystem.

Maybe smth like `replica_drop_gc_state` or just `replica_drop_gc`?

src/box/gc.h
```diff
@@ -347,18 +347,19 @@ void
...
  * @format... specifies a human-readable name that will be
  * used for listing the reference in box.info.gc().
+ *
+ * The function never fails - returned value is always not NULL.
  */
-CFORMAT(printf, 3, 4)
-void
-gc_ref_checkpoint(struct gc_checkpoint *checkpoint,
-		  struct gc_checkpoint_ref *ref, const char *format, ...);
+CFORMAT(printf, 2, 3)
```
Serpentian, line 367: Nit: `NODISCARD` in order to avoid memory leak

src/box/memtx_engine.cc
```diff
@@ -1341,36 +1341,18 @@ send_join_meta(struct xstream *stream, const struct raft_request *raft_req,
 #include "memtx_checkpoint_join.cc"
 #else /* !defined(ENABLE_FETCH_SNAPSHOT_CURSOR) */
 
-static int
-memtx_engine_prepare_checkpoint_join(struct engine *engine,
```
Serpentian, line 1345: Let's also drop `data` from the `engine_join_ctx` and allocation of that memory (in `engine_prepare_join`) . AFAIR, it was used only in `memtx_prepare_checkpoint_join`

https://github.com/tarantool/tarantool/blob/9a17476508ce0eb2562dfff1357281e25e274e15/src/box/engine.h#L361-L362

src/box/lua/ctl.c
Serpentian: > But it should work for anonymous replicas as well, in which case it'll drop the replica, no?

Yep, for anon replica the function drops it from `box.info.replication_anon`

sergepetrenko: > > But it should work for anonymous replicas as well, in which case it'll drop the replica, no?
> 
> Yep, for anon replica the function drops it from `box.info.replication_anon`

That's why I think the function name shouldn't mention wal gc

src/box/memtx_engine.cc
drewdzzz: Are you talking about [this commit](https://github.com/tarantool/tarantool/commit/72cc2b3ef7f2755ca7d611ee2f392f3311aad8a3#diff-42bfb43d7d5ce64ca78190e4c475ac8d6e17eddb90a44a43f7ea5cfe6e0f7754)? The member is only renamed there.

By the way, do we still need the above-mentioned refactoring, or should I revert it?

src/box/gc.h
drewdzzz: Actually, AFAICS, we use it only in vinyl. And I use it for `box_connect_replica` because it returns a class using RAII idiom so discarding it will break many things. What's about returning a pointer, we have a lot of such functions without this attribute, so this one will just look "peculiar".

src/box/memtx_engine.cc
drewdzzz: No, I meant the same [commit](https://github.com/tarantool/tarantool/commit/72cc2b3ef7f2755ca7d611ee2f392f3311aad8a3#diff-42bfb43d7d5ce64ca78190e4c475ac8d6e17eddb90a44a43f7ea5cfe6e0f7754) in which you renamed the member. It was preparatory for your checkpoint join PR, so I'm making sure if it's still needed after I moved checkpoint-referencing part to CE.

src/box/lua/ctl.c
Gerold103: The names proposed by @sergepetrenko are ok for me too. So I suggest Sergey chooses one name and we proceed.

src/box/replication.h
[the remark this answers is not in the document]
Gerold103: Why do we have to distinguish between `RELAY_STOPPED` and `RELAY_OFF`? The exact state could perhaps be moved out of relay like I suggested above about `enum replica_relay_state`. So the `replica` and `relay` would be more detached from each other.

Yeah, I agree it is perhaps better to do it later during the relay rework.

Gerold103 (approved):

src/box/memtx_engine.cc
Serpentian: Probably, we should not do that. We decided to move part of `relay_initial_join` to the `memtx_engine_join`, as it was ugly inside the relay. We'll have to export new functions from EE in order to send checkpoint `JOIN_META` from relay, which majorly differs from the `JOIN_META` from read view and we didn't like that. Now it's ok, and I'd not go any further

Serpentian (approved):

src/box/box.cc
```diff
@@ -4508,23 +4512,54 @@ box_process_fetch_snapshot(struct iostream *io,
...
+	auto replica_guard = box_connect_replica(&req.instance_uuid,
+						 gc_vclock, &replica);
+	/* Reference checkpoint in case of checkpoint join. */
+	struct gc_checkpoint_ref *checkpoint_ref = NULL;
+	auto gc_checkpoint_ref_guard = make_scoped_guard([&]() {
+		if (checkpoint_ref != NULL)
+			gc_unref_checkpoint(checkpoint_ref);
+	});
+	if (checkpoint != NULL) {
+		if (replica == NULL) {
```
locker, line 4576: How can `replica` be `NULL` here? AFAICS `box_connect_replica` creates an anonymous replica object if the replica isn't found by id.

src/box/replication.h
```diff
@@ -634,6 +628,12 @@ replicaset_add(uint32_t replica_id, const struct tt_uuid *instance_uuid);
 struct replica *
 replicaset_add_anon(const struct tt_uuid *replica_uuid);
 
+#if defined(__cplusplus)
+} /* extern "C" */
+
+int
+replica_check_id(uint32_t replica_id);
```
locker, line 645: I think we could move `replica_check_id` to the `extern "C"` section as well, while we are at it, - it doesn't throw.

src/box/replication.cc
```diff
@@ -206,6 +209,70 @@ replicaset_set_sync_quorum(void)
...
+				expired_replica = replica;
+				break;
+			}
+			*lowest_ttl = MIN(*lowest_ttl, ttl);
+		}
+		if (expired_replica == NULL)
+			break;
+		/* Save UUID since the replica will be deleted on success. */
+		struct tt_uuid uuid = expired_replica->uuid;
+		if (replica_deactivate(expired_replica) != 0) {
```
locker, line 246: Most of the checks performed by `replica_deactivate` don't make sense here. I think it's better to factor out the code doing actual deactivation into a separate static function (`replica_deactivate_impl`) and make `relica_deactivate` take uuid instead of a replica object to make the Lua wrapper thin.

src/box/replication.cc
```diff
@@ -206,6 +209,70 @@ replicaset_set_sync_quorum(void)
...
+}
+
+static int
+replication_anon_gc_fiber_f(va_list ap)
+{
+	(void)ap;
+	while (!fiber_is_cancelled()) {
+		fiber_check_gc();
+		double lowest_ttl = 0;
+		replication_anon_gc(&lowest_ttl);
```
locker, line 270: Since this function is called in a single place and already has a loop in it, I think it'd be better to inline it here, replacing `break` with `sleep` + `continue`.

changelogs/unreleased/box-ctl-deactivate-replica.md
```diff
@@ -0,0 +1,4 @@
+## feature/replication
+
+* Added `box.ctl.deactivate_replica` function that allows to release WAL GC
+  state of a disconnected replica.
```
lenkis, line 4:
```suggestion
* Added `box.ctl.deactivate_replica` function that allows to release the WAL GC
  state of a disconnected replica.
```

lenkis (approved):

src/box/box.cc
drewdzzz: INSTANCE_UUID is optional in fetch snapshot request - the replica is `NULL` when we have no UUID.

src/box/gc.c
```diff
@@ -903,24 +903,30 @@ gc_erase_consumer_on_commit(struct trigger *trigger, void *event)
...
 		return -1;
+	}
 
 	/* Do actual work on commit. */
 	struct trigger *on_commit =
 		xregion_alloc_object(&in_txn()->region, struct trigger);
 	trigger_create(on_commit, gc_erase_consumer_on_commit, consumer, NULL);
-	txn_stmt_on_commit(txn_current_stmt(in_txn()), on_commit);
-	return 0;
+	txn_stmt_on_commit(txn_last_stmt(in_txn()), on_commit);
```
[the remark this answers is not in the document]
drewdzzz: Added an explanation to the commit message:
```
In order to call `gc_erase_consumer` under the hood, the commit
populates it with autocommit mode so that it can be called without an
active transaction. However, underlying `txn_current_stmt` call is not
appropriate in this case - current statement is actually one that is
being executed: for example, we used to call this function in
`_cluster:on_replace` - in this case current statement is the replace in
`_cluster`. Since we don't have current statement in autocommit mode,
let's use last statement (the actual deletion in `_gc_consumers`) instead.
```

src/box/lua/ctl.c
```diff
@@ -45,6 +45,8 @@
 #include "box/engine.h"
 #include "box/memtx_engine.h"
 #include "box/raft.h"
+#include "box/relay.h"
```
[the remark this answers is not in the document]
drewdzzz: Fixed.

src/box/replication.cc
```diff
@@ -465,6 +465,41 @@ replica_clear_id(struct replica *replica)
...
+			 "deactivation of self is not allowed");
+		return -1;
+	}
+	if (replica->has_incoming_connection) {
+		diag_set(ClientError, ER_REPLICA_DEACTIVATE,
+			 tt_uuid_str(&replica->uuid),
+			 "it is connected");
+		return -1;
+	}
+	if (gc_erase_consumer(&replica->uuid) != 0)
```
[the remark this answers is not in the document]
drewdzzz: Good idea - done with a corresponding test case.

src/box/replication.h
drewdzzz: Done.

src/box/replication.cc
drewdzzz: Done.

changelogs/unreleased/wal-gc-of-anonymous-replicas.md
```diff
@@ -0,0 +1,5 @@
+## feature/replication
+
+* Now anonymous replicas retain required xlogs. Outdated anonymous replicas
+  and their WAL GC state are deleted automatically when being disconnected
+  for `box.cfg.replication_anon_ttl` seconds.
```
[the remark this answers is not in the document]
drewdzzz: Fixed.

Totktonada (approved):
Glanced over the cluster configuration part--no objections.

src/box/replication.cc
```diff
@@ -206,6 +207,61 @@ replicaset_set_sync_quorum(void)
...
+int
+replica_gc(const struct tt_uuid *uuid)
+{
+	if (in_txn() != NULL) {
+		diag_set(ClientError, ER_ACTIVE_TRANSACTION);
+		return -1;
+	}
+	if (tt_uuid_is_equal(uuid, &INSTANCE_UUID)) {
+		diag_set(ClientError, ER_REPLICA_GC, tt_uuid_str(uuid),
+			 "deactivation of self is not allowed");
```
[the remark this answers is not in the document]
drewdzzz: Sounds better, thanks! Also, now I start all the "details" messages with a capital letter since they are not the part of the main message.

By the way, updated `ER_REPLICA_GC` message: "Cannot cleanup replica's resources" -> "Cannot clean up replica's resources".

src/box/gc.c
```diff
@@ -1103,6 +1104,26 @@ gc_load_consumers(void)
...
+			 */
+			assert(!replica->anon);
+			continue;
+		}
+		/*
+		 * Create an anonymous replica object and register its
+		 * consumer so that it won't be lost.
+		 */
+		replica = replicaset_add_anon(&consumer->uuid);
+		replica->gc = gc_consumer_register(&consumer->vclock,
```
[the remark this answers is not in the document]
drewdzzz: Introduced a new function `replication_anon_gc_init` that is called after `gc_load_consumers`. By the way, added `replication_anon_gc_fiber != NULL` checks to `replication_shutdown` and `box_set_replication_anon_ttl` since the fiber is started later.

locker (approved):

PR #10197, opened 2024-07-04
fiber: phohibit fiber self join

### LINKED ISSUE #10196: hang when fiber join itself
The below script never finish.

```lua
fiber.self():set_joinable(true)
fiber.self():join()
os.exit()
```
Situation is similar when fiber join itself using C API.

nshy:
In this case join will just hang. Instead let's raise an error in case of Lua API and panic in case of C API.

Closes #10196

### DIFF
--- a/src/lib/core/fiber.h
+++ b/src/lib/core/fiber.h
@@ -370,6 +370,7 @@ fiber_set_joinable(struct fiber *fiber, bool yesno);
  *
  * @pre FIBER_IS_JOINABLE flag is set (panic if not).
  * @pre the fiber is not joined yet (panic if not).
+ * @pre the fiber is different from current (panic if not).
  *
  * \param f fiber to be woken up
  * \return fiber function ret code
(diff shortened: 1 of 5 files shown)


locker (approved):

src/lib/core/fiber.h
```diff
@@ -370,6 +370,7 @@ fiber_set_joinable(struct fiber *fiber, bool yesno);
  *
  * @pre FIBER_IS_JOINABLE flag is set (panic if not).
  * @pre the fiber is not joined yet (panic if not).
+ * @pre the fiber if different from current (panic if not).
```
lenkis, line 373:
```suggestion
 * @pre the fiber is different from current (panic if not).
```

lenkis (approved):

src/lib/core/fiber.h
nshy: Fixed.

locker:
Cherry-picked to 3.1. Cherry-picked to 2.11 along with 44401529a2253cbf00583a3680ae01f0cb74693c (#7562, #9311).

PR #10198, opened 2024-07-04
config: close old sockets after replacing socket

### LINKED ISSUE #9535: config: console leaves old listening sockets on reconfiguration
Say, we have a configuration w/o `console` section and then apply the following one.

```yaml
console:
  socket: foo.control
```

Tarantool listens on the new `foo.control`, but keeps listening on the default `var/{{ instance_name }}/run/tarantool.control`.

We should close the old listening sockets on the reconfiguration.

<details>
<summary>Details</summary>

We shouldn't re-create the listening socket if the path was not changed. And the check regarding the path change should take into account `chdir` on the first `box.cfg()` if `process.work_dir` is set.
</details>

mandesero:
Аfter replacing the `console.socket` in config tarantool keeps listening on the previous socket (by default `var/{{ instance_name }}/run/tarantool.control` or the previous socket set to `console.socket`). This patch closes the old listening sockets on the reconfiguration.

Closes #9535

NO_DOC=internal
NO_CHANGELOG=internal

### DIFF
--- a/changelogs/unreleased/gh-9535-close-listen-socket-after-replace.md
+++ b/changelogs/unreleased/gh-9535-close-listen-socket-after-replace.md
@@ -1,0 +1,4 @@
+## bugfix/config
+
+* Fixed a bug that causes Tarantool to continue listening on the previous socket
+  even after `console.socket` has changed (gh-9535).
--- a/src/box/lua/config/applier/console.lua
+++ b/src/box/lua/config/applier/console.lua
@@ -1,7 +1,11 @@
-local errno = require('errno')
 local console = require('console')
 local log = require('internal.config.utils.log')
 local instance_config = require('internal.config.instance_config')
+local fio = require('fio')
+
+-- Active socket and its location.
+local listening_socket = nil
+local listening_socket_abspath = nil
 
 local function socket_file_to_listen_uri(file)
     if file:startswith('/') or file:startswith('./') then
@@ -10,34 +14,47 @@ local function socket_file_to_listen_uri(file)
     return ('unix/:./%s'):format(file)
 end
 
+local function close_listening_socket()
+    if listening_socket ~= nil then
+        listening_socket:close()
+        listening_socket = nil
+        listening_socket_abspath = nil
+    end
+end
+
 local function apply(config)
     local configdata = config._configdata
     local enabled = configdata:get('console.enabled', {use_default = true})
     if not enabled then
         log.debug('console.apply: console is disabled by the ' ..
             'console.enabled option; skipping...')
+        -- If the console is disabled, we must close the active socket.
+        close_listening_socket()
         return
     end
 
-    local socket_file = configdata:get('console.socket', {use_default = true})
-    assert(socket_file ~= nil)
+    local new_socket_file = configdata:get('console.socket',
+        {use_default = true})
+    assert(new_socket_file ~= nil)
 
     -- The same socket file is pointed by different paths before
     -- and after first box.cfg() if the `process.work_dir` option
     -- is set.
     local iconfig_def = configdata._iconfig_def
-    socket_file = instance_config:prepare_file_path(iconfig_def, socket_file)
+    new_socket_file = instance_config:prepare_file_path(iconfig_def,
+        new_socket_file)
+    local new_socket_abspath = fio.abspath(new_socket_file)
 
-    local listen_uri = socket_file_to_listen_uri(socket_file)
-    log.debug('console.apply: %s', listen_uri)
+    -- If console.socket has changed we must close the listening socket.
+    if listening_socket_abspath ~= new_socket_abspath then
+        close_listening_socket()
+    end
 
-    -- Ignore 'Address already in use' error, because it is
-    -- natural effect of reloading the configuration. All others
-    -- errors are re-raised and lead to failure of apply of the
-    -- configuration.
-    local ok, res = pcall(console.listen, listen_uri)
-    if not ok and errno() ~= errno.EADDRINUSE then
-        error(res)
+    if listening_socket_abspath == nil then
+        local listen_uri = socket_file_to_listen_uri(new_socket_file)
+        log.debug('console.apply: %s', listen_uri)
+        listening_socket = console.listen(listen_uri)
+        listening_socket_abspath = new_socket_abspath
     end
 end
 
--- a/test/config-luatest/console_test.lua
+++ b/test/config-luatest/console_test.lua
@@ -58,11 +58,8 @@ local function assert_console_works(cluster, config)
     t.assert_equals(eval(s, 'return 123'), 123)
 end
 
-local function assert_console_closed(_replicaset, _config)
-    -- TODO (gh-9535): Enable it after a fix in the console
-    -- applier. Now it doesn't stop old console sockets.
-    --
-    -- t.assert_error(connect, cluster, 'instance-001', config, 0.1)
+local function assert_console_closed(cluster, config)
+    t.assert_error(connect, cluster, 'i-001', config, 0.1)
 end
 
 -- Start, reload, change, return back, disable.
@@ -133,4 +130,29 @@ g.test_parent_dir_in_socket_path = function(g)
 
     cluster:reload(config)
     assert_console_works(cluster, config)
+end
+
+-- Re-open console by setting enable to false and true
+g.test_reopen_console = function(g)
+    local config = cbuilder.new()
+        :add_instance('i-001', {})
+        :config()
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    assert_console_works(cluster, config)
+
+    local new_config = cbuilder.new(config)
+        :set_global_option('console.enabled', false)
+        :config()
+    cluster:reload(new_config)
+    assert_console_closed(cluster, config)
+    assert_console_closed(cluster, new_config)
+
+    local new_config_2 = cbuilder.new(new_config)
+        :set_global_option('console.enabled', true)
+        :config()
+    cluster:reload(new_config_2)
+    assert_console_works(cluster, config)
+    assert_console_works(cluster, new_config)
+    assert_console_works(cluster, new_config_2)
 end


src/box/lua/config/applier/console.lua
```diff
@@ -3,6 +3,9 @@ local console = require('console')
 local log = require('internal.config.utils.log')
 local instance_config = require('internal.config.instance_config')
 
+local old_socket_file = nil
+local old_socket = nil
```
mkokryashkin, line 7: Isn't just a socket enough?

src/box/lua/config/applier/console.lua
```diff
@@ -28,6 +31,18 @@ local function apply(config)
...
     socket_file = instance_config:prepare_file_path(iconfig_def, socket_file)
 
+    if old_socket_file == nil then
+        old_socket_file = socket_file
+    end
+
+    if old_socket ~= nil and socket_file ~= old_socket_file then
+        if type(old_socket.shutdown) == 'function' then
+            old_socket:shutdown()
+        end
```
mkokryashkin, line 41: `shutdown` and `close` are not interchangable. You should call `close` after shutdown anyway, since only `close` actually destroys the socket.

mkokryashkin (changes requested):
Let's paraphrase the commit message the following way:

After a socket change in the "console.socket" config property and subsequent config reload, Tarantool keeps listening on the old socket. This patch fixes this behavior by properly closing the connection on the old socket on config reload.

src/box/lua/config/applier/console.lua
mandesero: I tried to do something like this (src/box/lua/config/applier/console.lua:apply)
```.lua
 local ok, res = pcall(console.listen, listen_uri)
 if old_socket ~= nil and old_socket ~= res then
    if type(old_socket.close) == "function" then
        old_socket:close()
    end
end
old_socket = res
```
But for some reason there are problems with calling config:reload() again. And also the test test/config-luatest/console_test.lua is falling apart.

mkokryashkin: `socket.lua` specifies that socket has method `name`, which is implemented via `getsockname`. Try to use that

mkokryashkin (dismissed):

test/config-luatest/console_test.lua
```diff
@@ -58,11 +58,8 @@ local function assert_console_works(cluster, config)
     t.assert_equals(eval(s, 'return 123'), 123)
 end
 
-local function assert_console_closed(_replicaset, _config)
-    -- TODO (gh-9535): Enable it after a fix in the console
-    -- applier. Now it doesn't stop old console sockets.
-    --
-    -- t.assert_error(connect, cluster, 'instance-001', config, 0.1)
+local function assert_console_closed(_replicaset, config)
+    t.assert_error(connect, cluster, 'instance-001', config, 0.1)
```
Totktonada, line 62: It always passes, because the instance name is `instance-001` instead of `i-001`.

src/box/lua/config/applier/console.lua
```diff
@@ -31,11 +33,19 @@ local function apply(config)
     local listen_uri = socket_file_to_listen_uri(socket_file)
     log.debug('console.apply: %s', listen_uri)
 
+    if old_socket ~= nil and old_socket:name() ~= socket_file then
```
Totktonada, line 36: `<socket object>:name()` doesn't react on changing a current working directory.

```lua
local fio = require('fio')
local console = require('console')

local s = console.listen('unix/:./foo.control')
print(s:name())
fio.mkdir('bar')
box.cfg({work_dir = 'bar'})
print(s:name())
```

It prints:

```
unix/:./foo.control
unix/:./foo.control
```

I see the following scenarios, where it fails.

<details>
<summary>Existing socket is not closed, while it should</summary>

* **Action:** start an instance with `process.work_dir` = `w`.
* Default console socket is `<process.work_dir>/var/run/i-001/tarantool.control`.
* The `console.listen()` call is performed before changing CWD, so it uses the `w/var/run/i-001/tarantool.control` path.
* CWD is changed to `w`.
* **Action:** update config with `console.socket` = `w/var/run/{{ instance_name }}/tarantool.control`.
* **Action:** call `config:reload()`.
* It is expected that the old socket is closed and the new `w/w/<...>` one is created.
* `<existing socket>:name()` returns `unix/:./w/var/run/i-001/tarantool.control` that looks the same as the previous one, but it should be without `w/` taking into account the changed CWD.
* **Check:** the old socket is not closed, while is should be.
</details>

<details>
<summary>Existing socket is closed, while it shouldn't</summary>

* **Action:** start an instance.
* **Action:** update config with `console.socket` = `w/../var/run/{{ instance_name }}/tarantool.control`.
* **Action:** call `config:reload()`.
* It is expected that the socket is kept as is, without re-creation.
* Actually, the socket is closed, because the relative paths are different.
* **Check:** I guess that the effect of the socket re-creation leads to closing connections with currently connected clients.
</details>

Please, add test cases that would implement these scenarios.

I think that we should store an absolute path to perform the comparison.

src/box/lua/config/applier/console.lua
```diff
@@ -3,6 +3,8 @@ local console = require('console')
 local log = require('internal.config.utils.log')
 local instance_config = require('internal.config.instance_config')
 
+local old_socket = nil
```
Totktonada, line 6: Nit: It is not necessarily 'old' (that sounds like 'non-actual'). Just 'last', 'saved', 'previous', 'existing'.

src/box/lua/config/applier/console.lua
```diff
@@ -3,6 +3,8 @@ local console = require('console')
 local log = require('internal.config.utils.log')
```
Totktonada, line 3: I think that we should add a changelog entry, because the patch fixes a user visible behavior. The change is not internal: `NO_CHANGELOG=internal` is incorrect.

Also, the `NO_DOC` motivation can't be 'internal' too. I would just say that it is a bugfix.

src/box/lua/config/applier/console.lua
Totktonada, line 6: I guess it should be dropped to `nil` if the console is stopped. Please, add a test case that reveals a problem if we don't do that.

src/box/lua/config/applier/console.lua
```diff
@@ -31,11 +33,19 @@ local function apply(config)
...
+
     -- Ignore 'Address already in use' error, because it is
     -- natural effect of reloading the configuration. All others
     -- errors are re-raised and lead to failure of apply of the
     -- configuration.
     local ok, res = pcall(console.listen, listen_uri)
+    old_socket = res
     if not ok and errno() ~= errno.EADDRINUSE then
         error(res)
     end
```
Totktonada, line 51: Now we tracks the listening socket and there is no reason to perform `console.listen` if it already listens (and perform the `errno` check). We can just do the following (in pseudocode):

```lua
if socket_path_is_changed(<...>) then
    listening_socket:close()
    listening_socket = nil
end
if listening_socket == nil then
    listening_socket = console.listen(listen_uri)
end
```

src/box/lua/config/applier/console.lua
mandesero: work_dir option is non-dynamic and config:reload() doesn't change the work_dir path.
Is this a reason not to watch this case?

Totktonada: The console starts before the first box.cfg() call, which changes the working directory.

src/box/lua/config/applier/console.lua
mandesero: Does `enabled = false` mean that the console is stopped? If yes, then this case is covered by existing tests
(test/config-luatest/console_test.lua:test_basic).
```.lua
-- Disable the console.
local new_config_2 = cbuilder.new(config)
    :set_global_option('console.enabled', false)
    :config()
cluster:reload(new_config_2)
assert_console_closed(cluster, config)
assert_console_closed(cluster, new_config)
assert_console_closed(cluster, new_config_2)
```

src/box/lua/config/applier/console.lua
Totktonada: Please, leave comments open: I'll look over them at the next review iteration.

src/box/lua/config/applier/console.lua
Totktonada: Please, leave comments open: it is easier to look over them on the next review iteration.

test/config-luatest/console_test.lua
Totktonada: Please, leave comments open: usually a person, who opens the discussion, resolves it.

src/box/lua/config/applier/console.lua
Totktonada: It seems important to also verify that the console can be enabled back after disabling (without changing the socket path).

mandesero: test added

src/box/lua/config/applier/console.lua
mandesero: I don't really understand how to implement these scripts because `box.cfg()` is implicitly called when creating or starting a `cluster` in `test/config-luatest/console_test.lua`, as a result an attempt to change the working directory results in an error `work_dir option is non-dynamic`.

Totktonada: The config is applied in the following steps:

https://github.com/tarantool/tarantool/blob/cfd4bf4685de6ee2dced133519ce34f71e3bb71b/src/box/lua/config/init.lua#L146-L154

`console` applier is executed first, then `box_cfg`.

Let's consider the startup + reload scenario.

* A user starts tarantool with `--name` and `--config`
* `console.apply()` -- console starts to listen
* `box_cfg.apply()` -- CWD is changed
* A user reloads the configuration.
* `console.apply()` -- now it should compare a previous socket path that is relative to the previous CWD and the new (possibly unchanged) socket path that is relative to the new CWD
* `box_cfg.apply()`

The key point is that the first `console.apply()` is performed before changing the CWD, but the second one (triggered by `config:reload()`) after changing the CWD.

changelogs/unreleased/gh-9535-close-listen-socket-after-replace.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Fix a bug that when console.socket changes, tarantool listens on the new socket,
+  but keeps listening on the previous socket (gh-9535).
```
andreyaksenov, line 4:
```suggestion
* Fixed a bug that causes Tarantool to continue listening on the previous socket
  even after `console.socket` has changed (gh-9535).
```

andreyaksenov (approved):

src/box/lua/config/applier/console.lua
mandesero: Done

src/box/lua/config/applier/console.lua
mandesero: corrected

src/box/lua/config/applier/console.lua
mandesero: corrected

test/config-luatest/console_test.lua
mandesero: Done

src/box/lua/config/applier/console.lua
```diff
@@ -10,34 +14,49 @@ local function socket_file_to_listen_uri(file)
...
-    local ok, res = pcall(console.listen, listen_uri)
-    if not ok and errno() ~= errno.EADDRINUSE then
-        error(res)
+    new_socket_file = instance_config:prepare_file_path(
+        iconfig_def, new_socket_file)
+    local new_socket_file_path = fio.abspath(new_socket_file)
+
+    -- If console.socket has changed we must close the listening socket.
+    if listening_socket_file ~= new_socket_file_path and
+            listening_socket ~= nil then
```
Totktonada, line 51: The `listening_socket ~= nil` check is already in `close_listening_socket()`, it is redundant here.

src/box/lua/config/applier/console.lua
```diff
@@ -10,34 +14,49 @@ local function socket_file_to_listen_uri(file)
...
-    -- configuration.
-    local ok, res = pcall(console.listen, listen_uri)
-    if not ok and errno() ~= errno.EADDRINUSE then
-        error(res)
+    new_socket_file = instance_config:prepare_file_path(
+        iconfig_def, new_socket_file)
+    local new_socket_file_path = fio.abspath(new_socket_file)
+
+    -- If console.socket has changed we must close the listening socket.
+    if listening_socket_file ~= new_socket_file_path and
```
Totktonada, line 50: The naming is inconsistent. One `<...>_file` valiable holds a relative path, another holds an absolute path. I propose to use `_file` for a relative path and `_abspath` for the absolute one.

<details>
<summary>I proopse the following changes.</summary>

```diff
diff --git a/src/box/lua/config/applier/console.lua b/src/box/lua/config/applier/console.lua
index 7de5ab8f3a..c34d409280 100644
--- a/src/box/lua/config/applier/console.lua
+++ b/src/box/lua/config/applier/console.lua
@@ -3,9 +3,9 @@ local log = require('internal.config.utils.log')
 local instance_config = require('internal.config.instance_config')
 local fio = require('fio')
 
--- active socket and its location
+-- Active socket and its location.
 local listening_socket = nil
-local listening_socket_file = nil
+local listening_socket_abspath = nil
 
 local function socket_file_to_listen_uri(file)
     if file:startswith('/') or file:startswith('./') then
@@ -18,7 +18,7 @@ local function close_listening_socket()
     if listening_socket ~= nil then
         listening_socket:close()
         listening_socket = nil
-        listening_socket_file = nil
+        listening_socket_abspath = nil
     end
 end
 
@@ -33,30 +33,28 @@ local function apply(config)
         return
     end
 
-    local new_socket_file = configdata:get(
-        'console.socket', {use_default = true}
-    )
+    local new_socket_file = configdata:get('console.socket',
+        {use_default = true})
     assert(new_socket_file ~= nil)
 
     -- The same socket file is pointed by different paths before
     -- and after first box.cfg() if the `process.work_dir` option
     -- is set.
     local iconfig_def = configdata._iconfig_def
-    new_socket_file = instance_config:prepare_file_path(
-        iconfig_def, new_socket_file)
-    local new_socket_file_path = fio.abspath(new_socket_file)
+    new_socket_file = instance_config:prepare_file_path(iconfig_def,
+        new_socket_file)
+    local new_socket_abspath = fio.abspath(new_socket_file)
 
     -- If console.socket has changed we must close the listening socket.
-    if listening_socket_file ~= new_socket_file_path and
-            listening_socket ~= nil then
+    if listening_socket_abspath ~= new_socket_abspath then
         close_listening_socket()
     end
 
-    if listening_socket_file == nil then
+    if listening_socket_abspath == nil then
         local listen_uri = socket_file_to_listen_uri(new_socket_file)
         log.debug('console.apply: %s', listen_uri)
         listening_socket = console.listen(listen_uri)
-        listening_socket_file = new_socket_file_path
+        listening_socket_abspath = new_socket_abspath
     end
 end
 
```
</details>

Totktonada (approved):
Thanks!

The implementation looks valid and seems good for me.

I've left a few stylistic remarks, please, take a look.

src/box/lua/config/applier/console.lua
mandesero: extra check removed

src/box/lua/config/applier/console.lua
mandesero: Corrected

georgiy-belyanin:
Thanks for the patch. After quite a lot of remarks from other people LGTM.

Totktonada:
@mkokryashkin is on vacation and I guess he can't proceed with the review. I'm going to push this patch.

Totktonada:
Pushed to `master` (a future `3.2.0` release).

Backported to `release/3.1` (a future `3.1.2` release) in PR #10381.

Issue #9917, opened 2024-04-08
Speed up synchronous transaction queue processing


sergepetrenko:
Performance of synchronous replication degrades too much with the increase of parallel synchronous requests.
For example, on my machine with 6000 requests in parallel (`box.info.synchro.queue.len` reads `6000`) master was only able to pull about 8000 RPS, and messages like this were frequent in the log:
```
2024-04-08 17:23:11.047 [43663] main txn.c:830 W> too long WAL write: 1 rows at LSN 11794161: 0.520 sec
```

The issue manifested itself even with `replication_synchro_quorum = 1`, meaning it's not related to network delay, also the size of the quorum didn't influence the results too much. It seems the problem lies in the way synchronous transactions are processed in the queue.

Besides, when trying the same 6000 concurrent requests to replace something in an async space, the RPS was as high as 300k, meaning the issue isn't related to batch finalization of transactions.

Most likely the cause of degradation is the way our `txn_limbo_ack` traverses the whole transaction list. In the example above `txn_limbo_ack` is always called with lsn of the last of 6000 transactions, but it still traverses the whole list and assigns `ack_count` separately to each transaction. We might improve this: persist an array of lsn's of acks, once the ack_lsn is increased - find the point up to which everything should be committed via binary search, for example.

Astronomax:
I don't understand how an array of lsns of acks will help. If you mean `limbo->vclock`, which already exists, then I understand. But this is not enough, we need to use something like `std::deque` instead of `rlist` to store `txn_limbo_entry` because we need to be able to random access in O(1) time. In addition, `std::deque` has all the necessary advantages: push/pop back for O(1) time amortized and pointers to elements are not invalidated during push.

Astronomax:
@CuriousGeorgiy ring buffer looks like a good solution in this case. However, the implementation that already exists in tarantool given in `fifo.h` is not good enough for this case. The main problem is that push does not work in constant time. There is a magic constant `FIFO_WATERMARK`:
https://github.com/tarantool/tarantool/blob/9b63ced353eaf6c1224358c7d3655ca52f7d21fb/src/lib/salad/fifo.h#L35
There is also code here that moves the entire used block to the beginning if it turns out that more than `256` pops were produced after the last memmove on push operation.
https://github.com/tarantool/tarantool/blob/9b63ced353eaf6c1224358c7d3655ca52f7d21fb/src/lib/salad/fifo.h#L76-L85
[...]

PR #10209, opened 2024-07-05
limbo: speed up synchronous transaction queue processing

Astronomax:
This patch optimizes the process of collecting ACKs from replicas for synchronous transactions. Before this patch, collecting confirmations was slow in some cases. There was a possible situation where it was necessary to go through the entire queue again every time the next ACK was received from the replica. This was especially noticeable in the case of a large number of parallel synchronous requests.
Performance was measured on the `1mops_write` benchmark. 6000 fibers are executed simultaneously on the master. 
Each fiber generates one transaction and waits for its completion, then generates the next one. Thus, on average there are about 6000 transactions in limbo at a time.

```console
tarantool:~$ tarantool perf/lua/1mops_write.lua --nodes=31 --fibers=6000 --ops=1000000 --transaction=1 --warmup=10 --sync
# making 1000000 REPLACE operations,
# 1 operations per txn,
# using 6000 fibers,
...
```

<table border="1" cellspacing="0" cellpadding="8">
    <thead>
        <tr>
            <th>--nodes=</th>
            <th>Before</th>
            <th>After</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>1</td>
            <td>master average speed: 67527 ops/sec</td>
            <td>master average speed: 569607 ops/sec</td>
        </tr>
        <tr>
            <td rowspan="2">2</td>
            <td>master average speed: 10020 ops/sec</td>
            <td>master average speed: 185128 ops/sec</td>
        </tr>
        <tr>
            <td>1mops_replica_rps: 10020</td>
            <td>1mops_replica_rps: 185126</td>
        </tr>
        <tr>
            <td rowspan="2">3</td>
            <td>master average speed: 10725 ops/sec</td>
            <td>master average speed: 163334 ops/sec</td>
        </tr>
        <tr>
            <td>1mops_replica_rps: 10725</td>
            <td>1mops_replica_rps: 163332</td>
        </tr>
        <tr>
            <td rowspan="2">31</td>
            <td>master average speed: 7157 ops/sec</td>
            <td>master average speed: 15058 ops/sec</td>
        </tr>
        <tr>
            <td>1mops_replica_rps: 7157</td>
            <td>1mops_replica_rps: 15057</td>
        </tr>
    </tbody>
</table>

Closes https://github.com/tarantool/tarantool/issues/9917

NO_DOC=performance improvement
NO_TEST=performance improvement

### DIFF
--- a/changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
+++ b/changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
@@ -1,0 +1,3 @@
+## bugfix/limbo
+
+* Optimized synchronous transaction queue processing (gh-9917).
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -53,6 +53,7 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->promote_greatest_term = 0;
 	latch_create(&limbo->promote_latch);
 	limbo->confirmed_lsn = 0;
+	limbo->entry_to_confirm = NULL;
 	limbo->rollback_count = 0;
 	limbo->is_in_rollback = false;
 	limbo->svp_confirmed_lsn = -1;
@@ -162,13 +163,17 @@ txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
 	size_t size;
 	struct txn_limbo_entry *e = region_alloc_object(&txn->region,
 							typeof(*e), &size);
+	if (limbo->entry_to_confirm == NULL &&
+	    txn_has_flag(txn, TXN_WAIT_ACK)) {
+		limbo->entry_to_confirm = e;
+		limbo->ack_count = 0;
+	}
 	if (e == NULL) {
 		diag_set(OutOfMemory, size, "region_alloc_object", "e");
 		return NULL;
 	}
 	e->txn = txn;
 	e->lsn = -1;
-	e->ack_count = 0;
 	e->is_commit = false;
 	e->is_rollback = false;
 	e->insertion_time = fiber_clock();
@@ -202,6 +207,8 @@ void
 txn_limbo_abort(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
 {
 	entry->is_rollback = true;
+	if (entry == limbo->entry_to_confirm)
+		limbo->entry_to_confirm = NULL;
 	/*
 	 * The simple rule about rollback/commit order applies
 	 * here as well: commit always in the order of WAL write,
@@ -230,6 +237,9 @@ txn_limbo_assign_remote_lsn(struct txn_limbo *limbo,
 	entry->lsn = lsn;
 }
 
+static void
+txn_limbo_confirm(struct txn_limbo *limbo);
+
 void
 txn_limbo_assign_local_lsn(struct txn_limbo *limbo,
 			   struct txn_limbo_entry *entry, int64_t lsn)
@@ -240,19 +250,8 @@ txn_limbo_assign_local_lsn(struct txn_limbo *limbo,
 	assert(lsn > 0);
 
 	entry->lsn = lsn;
-	/*
-	 * The entry just got its LSN after a WAL write. It could
-	 * happen that this LSN was already ACKed by some
-	 * replicas. Update the ACK counter to take them into
-	 * account.
-	 */
-	struct vclock_iterator iter;
-	vclock_iterator_init(&iter, &limbo->vclock);
-	int ack_count = 0;
-	vclock_foreach(&iter, vc)
-		ack_count += vc.lsn >= lsn;
-	assert(ack_count >= entry->ack_count);
-	entry->ack_count = ack_count;
+	if (entry == limbo->entry_to_confirm)
+		limbo->ack_count = vclock_count_ge(&limbo->vclock, entry->lsn);
 }
 
 void
@@ -597,6 +596,7 @@ txn_limbo_read_promote(struct txn_limbo *limbo, uint32_t replica_id,
 	limbo->owner_id = replica_id;
 	limbo->confirmed_lsn = vclock_get(&limbo->confirmed_vclock,
 					  replica_id);
+	limbo->entry_to_confirm = NULL;
 	box_update_ro_summary();
 }
 
@@ -633,6 +633,56 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
 	return txn_limbo_read_promote(limbo, REPLICA_ID_NIL, lsn);
 }
 
+/**
+ * Check that some synchronous transactions have gathered quorum and
+ * write a confirmation entry of the last confirmed transaction.
+ */
+static void
+txn_limbo_confirm(struct txn_limbo *limbo)
+{
+	assert(limbo->owner_id == instance_id);
+	if (limbo->is_in_rollback)
+		return;
+	if (limbo->entry_to_confirm == NULL ||
+	    limbo->entry_to_confirm->lsn == -1)
+		return;
+	if (limbo->ack_count < replication_synchro_quorum)
+		return;
+	int32_t k = (int32_t)vclock_size(&limbo->vclock)
+		    - replication_synchro_quorum;
+	/**
+	 * limbo->ack_count >= replication_synchro_quorum =>
+	 * vclock_size(&limbo->vclock) >= replication_synchro_quorum
+	 */
+	assert(k >= 0);
+	int64_t confirm_lsn = vclock_nth_element(&limbo->vclock, k);
+	assert(confirm_lsn >= limbo->entry_to_confirm->lsn);
+	struct txn_limbo_entry *e = limbo->entry_to_confirm;
+	limbo->entry_to_confirm = NULL;
+	int64_t max_assigned_lsn = -1;
+	for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
+	       e = rlist_next_entry(e, in_queue)) {
+		if (!txn_has_flag(e->txn, TXN_WAIT_ACK))
+			continue;
+		if (e->lsn == -1 || e->lsn > confirm_lsn) {
+			limbo->entry_to_confirm = e;
+			/**
+			 * It may be that a quorum has been gathered, but
+			 * ack_count = 0. It's ok. CONFIRM will be written as
+			 * soon as the lsn is assigned to the transaction.
+			 */
+			limbo->ack_count = (e->lsn == -1) ? 0 :
+				vclock_count_ge(&limbo->vclock, e->lsn);
+			break;
+		} else {
+			max_assigned_lsn = e->lsn;
+		}
+	}
+	assert(max_assigned_lsn != -1);
+	txn_limbo_write_confirm(limbo, max_assigned_lsn);
+	txn_limbo_read_confirm(limbo, max_assigned_lsn);
+}
+
 void
 txn_limbo_ack(struct txn_limbo *limbo, uint32_t replica_id, int64_t lsn)
 {
@@ -655,9 +705,10 @@ txn_limbo_ack(struct txn_limbo *limbo, uint32_t replica_id, int64_t lsn)
 		return;
 	assert(limbo->owner_id != REPLICA_ID_NIL);
 	int64_t prev_lsn = vclock_get(&limbo->vclock, replica_id);
+	assert(lsn >= prev_lsn);
 	/*
 	 * One of the reasons why can happen - the remote instance is not
-	 * read-only and wrote something under its own insance_id. For qsync
+	 * read-only and wrote something under its own instance_id. For qsync
 	 * that most likely means that the remote instance decided to take over
 	 * the limbo ownership, and the current node is going to become a
 	 * replica very soon.
@@ -665,31 +716,15 @@ txn_limbo_ack(struct txn_limbo *limbo, uint32_t replica_id, int64_t lsn)
 	if (lsn == prev_lsn)
 		return;
 	vclock_follow(&limbo->vclock, replica_id, lsn);
-	struct txn_limbo_entry *e;
-	int64_t confirm_lsn = -1;
-	rlist_foreach_entry(e, &limbo->queue, in_queue) {
-		assert(e->ack_count <= VCLOCK_MAX);
-		if (e->lsn > lsn)
-			break;
-		/*
-		 * Sync transactions need to collect acks. Async
-		 * transactions are automatically committed right
-		 * after all the previous sync transactions are.
-		 */
-		if (!txn_has_flag(e->txn, TXN_WAIT_ACK)) {
-			continue;
-		} else if (e->lsn <= prev_lsn) {
-			continue;
-		} else if (++e->ack_count < replication_synchro_quorum) {
-			continue;
-		} else {
-			confirm_lsn = e->lsn;
-		}
+
+	if (limbo->entry_to_confirm != NULL &&
+	    limbo->entry_to_confirm->lsn != -1) {
+		if (limbo->entry_to_confirm->lsn <= prev_lsn ||
+		    lsn < limbo->entry_to_confirm->lsn)
+			return;
+		if (++limbo->ack_count >= replication_synchro_quorum)
+			txn_limbo_confirm(limbo);
 	}
-	if (confirm_lsn == -1 || confirm_lsn <= limbo->confirmed_lsn)
-		return;
-	txn_limbo_write_confirm(limbo, confirm_lsn);
-	txn_limbo_read_confirm(limbo, confirm_lsn);
 }
 
 /**
@@ -1262,23 +1297,9 @@ txn_limbo_on_parameters_change(struct txn_limbo *limbo)
 {
 	if (rlist_empty(&limbo->queue) || txn_limbo_is_frozen(limbo))
 		return;
-	struct txn_limbo_entry *e;
-	int64_t confirm_lsn = -1;
-	rlist_foreach_entry(e, &limbo->queue, in_queue) {
-		assert(e->ack_count <= VCLOCK_MAX);
-		if (!txn_has_flag(e->txn, TXN_WAIT_ACK)) {
-			continue;
-		} else if (e->ack_count < replication_synchro_quorum) {
-			continue;
-		} else {
-			confirm_lsn = e->lsn;
-			assert(confirm_lsn > 0);
-		}
-	}
-	if (confirm_lsn > limbo->confirmed_lsn && !limbo->is_in_rollback) {
-		txn_limbo_write_confirm(limbo, confirm_lsn);
-		txn_limbo_read_confirm(limbo, confirm_lsn);
-	}
+	/* The replication_synchro_quorum value may have changed. */
+	if (limbo->owner_id == instance_id)
+		txn_limbo_confirm(limbo);
 	/*
 	 * Wakeup all the others - timed out will rollback. Also
 	 * there can be non-transactional waiters, such as CONFIRM
--- a/src/box/txn_limbo.h
+++ b/src/box/txn_limbo.h
@@ -59,11 +59,6 @@ struct txn_limbo_entry {
 	 */
 	int64_t lsn;
 	/**
-	 * Number of ACKs. Or in other words - how many replicas
-	 * confirmed receipt of the transaction.
-	 */
-	int ack_count;
-	/**
 	 * Result flags. Only one of them can be true. But both
 	 * can be false if the transaction is still waiting for
 	 * its resolution.
@@ -169,6 +164,19 @@ struct txn_limbo {
 	 * illegal.
 	 */
 	int64_t confirmed_lsn;
+	/**
+	 * The first unconfirmed synchronous transaction in the current term.
+	 * Is NULL if there is no such transaction, or if the current instance
+	 * does not own limbo.
+	 */
+	struct txn_limbo_entry *entry_to_confirm;
+	/**
+	 * Number of ACKs of the first unconfirmed synchronous transaction
+	 * (entry_to_confirm->txn). Contains the actual value only for a
+	 * non-NULL entry_to_confirm with a local lsn assigned. Otherwise
+	 * it may contain any trash.
+	 */
+	int ack_count;
 	/**
 	 * Total number of performed rollbacks. It used as a guard
 	 * to do some actions assuming all limbo transactions will


src/box/txn.c
```diff
@@ -734,6 +734,8 @@ txn_complete_fail(struct txn *txn)
 	if (txn->limbo_entry != NULL) {
 		assert(txn_has_flag(txn, TXN_WAIT_SYNC));
 		txn_limbo_abort(&txn_limbo, txn->limbo_entry);
+		txn_limbo_update_confirmed_entry_after_rollback(
+			&txn_limbo, txn->limbo_entry->lsn);
```
Gerold103, line 738: Lets update the confirmed entry right inside of `txn_limbo_abort()`. It doesn't look right that the limbo's state gets broken after the abort and needs the explicit manual "repairing" via this new additional call.

src/box/txn_limbo.c
```diff
@@ -53,6 +55,9 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->promote_greatest_term = 0;
 	latch_create(&limbo->promote_latch);
 	limbo->confirmed_lsn = 0;
+	limbo->confirmed_entry =
+		rlist_entry(limbo->queue, struct txn_limbo_entry, in_queue);
+	limbo->confirmed_entry_is_valid = false;
```
Gerold103, line 60: Is this flag needed? Why can't the entry being `NULL` be counted as "invalid"?

src/box/txn_limbo.h
```diff
@@ -90,14 +85,15 @@ txn_limbo_entry_is_complete(const struct txn_limbo_entry *e)
...
+	struct txn_limbo_entry head_entry;
 	/**
 	 * Queue of limbo entries. Ordered by LSN. Some of the
 	 * entries in the end may not have an LSN yet (their local
 	 * WAL write is still in progress), but their order won't
 	 * change anyway. Because WAL write completions will give
 	 * them LSNs in the same order.
 	 */
-	struct rlist queue;
+	struct rlist *queue;
```
Gerold103, line 96: I would suggest to reconsider the storage approach here. Lets keep the queue as is. What is wrong with that? Why do you need the queue to be a pointer?

src/box/txn_limbo.c
```diff
@@ -268,6 +267,21 @@ txn_limbo_assign_lsn(struct txn_limbo *limbo, struct txn_limbo_entry *entry,
...
+txn_limbo_update_confirmed_entry_after_rollback(struct txn_limbo *limbo,
+						int64_t lsn)
+{
+	struct txn_limbo_entry *last_entry = rlist_last_entry(
+		limbo->queue, struct txn_limbo_entry, in_queue);
+	if (limbo->confirmed_entry == last_entry) {
+		limbo->confirmed_entry_is_valid = false;
+	} else if (lsn <= rlist_next_entry(limbo->confirmed_entry,
+					   in_queue)->lsn) {
+		txn_limbo.confirmed_entry = last_entry;
```
Gerold103, line 280: `txn_limbo` is the global object, you probably wanted to use `limbo->`, from the arguments of this function.

src/box/txn_limbo.c
```diff
@@ -444,14 +460,30 @@ txn_limbo_write_confirm(struct txn_limbo *limbo, int64_t lsn)
...
+void
+txn_limbo_update_confirmed_entry_after_confirm(struct txn_limbo *limbo,
+					       int64_t lsn)
+{
+	if (lsn >= limbo->confirmed_entry->lsn) {
+		limbo->confirmed_entry = &limbo->head_entry;
+		limbo->confirmed_entry_is_valid = false;
+	}
+}
+
```
Gerold103, line 477: Given this function is called in a single place of a single cpp file and is just 3 lines long, I suggest to simply inline it there.

src/box/txn_limbo.c
```diff
@@ -509,7 +541,9 @@ txn_limbo_read_confirm(struct txn_limbo *limbo, int64_t lsn)
 		 */
 		assert(e->txn->signature >= 0);
 		txn_limbo_complete(e->txn, true);
+		last_confirm_lsn = e->lsn;
 	}
+	txn_limbo_update_confirmed_entry_after_confirm(limbo, last_confirm_lsn);
```
Gerold103, line 546: Below is a check:
```C
	if (limbo->confirmed_lsn < lsn) {
		limbo->confirmed_lsn = lsn;
		vclock_follow(&limbo->confirmed_vclock, limbo->owner_id, lsn);
	}
```
Why do you update the confirmed entry even when `confirmed_lsn` didn't change?

src/box/txn_limbo.c
```diff
@@ -649,10 +685,66 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
...
+		limbo->confirmed_entry_is_valid = true;
+		limbo->ack_count = vclock_count_ge(&limbo->vclock, e->lsn);
+	}
+	if (limbo->ack_count < replication_synchro_quorum)
+		return;
+	e = rlist_next_entry(limbo->confirmed_entry, in_queue);
+	/* Make sure confirmed_entry is valid. It must be true. */
+	assert(!rlist_entry_is_head(e, limbo->queue, in_queue));
+	assert(txn_has_flag(e->txn, TXN_WAIT_ACK));
+	assert(e->lsn != -1);
```
Gerold103, line 723: The limbo can have txns not having an LSN yet and also being synchronous. It happens when their WAL write isn't finished yet, but they already received ACKs and a confirm. How does this assert survive that?

src/box/txn_limbo.c
```diff
@@ -671,41 +763,30 @@ txn_limbo_ack(struct txn_limbo *limbo, uint32_t replica_id, int64_t lsn)
...
+		assert(e->lsn != -1);
+		if (e->lsn <= prev_lsn || lsn < e->lsn)
+			return;
+		++limbo->ack_count;
 	}
-	if (confirm_lsn == -1 || confirm_lsn <= limbo->confirmed_lsn)
-		return;
-	txn_limbo_write_confirm(limbo, confirm_lsn);
-	txn_limbo_read_confirm(limbo, confirm_lsn);
+	txn_limbo_confirm(limbo);
```
Gerold103, line 789: Why do you call confirm even when `limbo->ack_count` didn't change?

src/lib/vclock/vclock.h
```diff
@@ -481,6 +481,18 @@ vclockset_match(vclockset_t *set, const struct vclock *key)
...
+ */
+int64_t
+vclock_nth_element(const struct vclock *vclock, uint32_t k);
+
+/**
+ * Count the number of components that are more than or equal to a given value.
+ */
+int
+vclock_count_ge(const struct vclock *vclock, int64_t lsn);
+
```
Gerold103, line 495: Please, add unit tests for these, in `test/unit/vclock.cc`.

changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/raft
+
+* Optimized transaction queue processing (gh-9917).
```
Gerold103, line 3: Lets say "synchronous transaction", not just "transaction".

Gerold103 (commented):
Good job, thanks for doing such a thorough investigation, the approach to the benchmarks ⚡️ was impressive.

Could you also please mention in the commit message the speed difference? I.e. the speed up is x2, x5, x10, or what? I see the old number "8k RPS", but what is the new one? And what was the bench? (Yeah, that is all in the ticket, but the commit message ideally must be a summary and be self-sufficient - it would help the future us a lot.)

src/box/txn_limbo.c
Astronomax: `confirmed_lsn` is not the same as `confirmed_entry->lsn`. 
There is only such a statement `confirmed_lsn >= confirmed_entry->lsn`. Perhaps `confirm_entry` is not the best name for it. May be it should be named as `confirmation_pointer` or something like.

`confirmed_entry_is_valid = true` means that the next transaction after `confirmed_entry` is synchronous and `ack_count` stores the correct value. Only in this case confirmed_entry is considered valid. But even if `confirmed_node` is not valid, it still stores a useful value. In this case, `confirmed_entry` stores an entry from which it will be possible to continue searching for the next synchronous transaction.

This was done this way only so as not to once again go through the entries that we had already checked. Let's imagine that `confirmed_entry` is literally the last confirmed synchronous transaction (`confirmed_lsn = confirmed_entry->lsn`). Then the following situation is possible:
1. limbo queue: [..., sync, async, async, async, async, ...] (all transactions after `confirmed_entry` are asynchronous)
`confirmed_entry` --^
Soon `CONFIRM` will be written to WAL and this synchronous transaction, along with all subsequent asynchronous transactions, will be deleted from limbo. But until this happens, they continue to hang in limbo.
2. We receive an ACK for the last transaction in the queue from one of the replicas in the replicaset.
3. We move on, starting with `confirmed_entry` in hopes of finding a confirmed synchronous transaction and updating `confirmed_lsn`. But we don’t find it.
https://github.com/Astronomax/tarantool/blob/6eee079a537a40c9721f6ca8a9fe616748f3b5c6/src/box/txn_limbo.c#L680-L684
4. But we don’t remember where we came, leave the `confirmed_entry` value the same.
5. Having received the same ACK from another replica, we will go through the same transactions again. And that's the problem.

There is no difference in performance on the `1mops_write` benchmark. But if you use synchronous and asynchronous replication together, this can be noticeable if there are noticeably more asynchronous transactions than synchronous ones.

src/box/txn_limbo.h
Astronomax: I decided to add a dummy entry `head_entry` which is the head of the list. This significantly simplifies the code and allows you to avoid several checks.
https://github.com/Astronomax/tarantool/blob/6eee079a537a40c9721f6ca8a9fe616748f3b5c6/src/box/txn_limbo.h#L88-L92
Therefore `queue` is forced to be a pointer.

src/box/txn_limbo.c
Astronomax: `confirmed_lsn` might not have changed, but it could be that `lsn == confirmed_entry`. In this case, `confirmed_entry` needs to be updated.
https://github.com/Astronomax/tarantool/blob/6eee079a537a40c9721f6ca8a9fe616748f3b5c6/src/box/txn_limbo.c#L531-L534

src/box/txn_limbo.c
Astronomax: This is the invariant we provide.
`confirmed_entry_is_valid == true` only when the next transaction after `confirmed_entry` is synchronous and has lsn (not -1) and `ack_count` stores the correct value.
The `confirmed_entry_is_valid` variable is written to `true` in a single place in the code and only if lsn != -1. https://github.com/Astronomax/tarantool/blob/6eee079a537a40c9721f6ca8a9fe616748f3b5c6/src/box/txn_limbo.c#L689
If at any point this is no longer true, `confirmed_entry_id_valid` is immediately set to `false`. This is processed in all places where the limbo is modified in any way.

src/box/txn_limbo.c
Astronomax: This does not mean that `ack_count` has not changed. If `confirmed_entry_is_valid == false`, this means that inside the `txn_limbo_confirm` call, we will try to make `confirmed_entry` valid again and calculate the correct `ack_count` value using `vclock_count_ge`. This ACK will be taken into account since before calling `txn_limbo_confirm` we executed `vclock_follow(&limbo->vclock, replica_id, lsn);` (https://github.com/Astronomax/tarantool/blob/6eee079a537a40c9721f6ca8a9fe616748f3b5c6/src/box/txn_limbo.c#L753C2-L753C49).

src/lib/vclock/vclock.h
Astronomax: Added.

changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
Astronomax: Fixed.

src/box/txn_limbo.c
Astronomax: Fixed.

src/box/txn_limbo.c
Astronomax: Fixed.

src/box/txn.c
Astronomax: Fixed.

src/box/txn_limbo.c
```diff
@@ -208,7 +212,16 @@ txn_limbo_abort(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
 	 * rollback in the reversed order. Rolled back transaction
 	 * is always the last.
 	 */
```
Gerold103, line 214: The comment above was for `txn_limbo_pop` a few lines below. Lets keep the comment and that code line together.

src/box/txn_limbo.c
```diff
@@ -208,7 +212,16 @@ txn_limbo_abort(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
...
+	if (entry == limbo->confirmed_entry) {
+		limbo->confirmed_entry = rlist_prev_entry(
+			limbo->confirmed_entry, in_queue);
+		limbo->confirmed_entry_is_valid = false;
+	} else if (entry == rlist_next_entry(limbo->confirmed_entry,
+					     in_queue)) {
+		limbo->confirmed_entry_is_valid = false;
+	}
```
Gerold103, line 222: - How could a confirmed entry be aborted? After a txn is confirmed (its lsn <= confirmed_lsn), it is 100% commited, hm?
- Why does it become invalid? A txn was aborted, ok. If somehow the aborted txn was the confirmed one, you make a step back and it is valid again - confirm of LSN L automatically means all LSNs < L are also confirmed. Otherwise if the aborted txn was after the confirmed LSN, there seem to be no reason why it would affect the confirmed entry. Hm?

src/box/txn_limbo.c
```diff
@@ -519,6 +528,10 @@ txn_limbo_read_confirm(struct txn_limbo *limbo, int64_t lsn)
 		limbo->confirmed_lsn = lsn;
 		vclock_follow(&limbo->confirmed_vclock, limbo->owner_id, lsn);
 	}
+	if (limbo->confirmed_entry->lsn <= lsn) {
+		limbo->confirmed_entry = &limbo->head_entry;
+		limbo->confirmed_entry_is_valid = false;
+	}
```
Gerold103, line 534: Here this `lsn` is always `== limbo->confirmed_lsn`. Your invariant was that `confirmed_lsn >= confirmed_entry->lsn`, which makes this condition always true. I added ` else {assert(false);}` to this `if` and the tests passed. Can we drop this `if` then?

src/box/txn_limbo.c
```diff
@@ -649,10 +662,66 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
 	return txn_limbo_read_promote(limbo, REPLICA_ID_NIL, lsn);
 }
 
+/**
+ * Check that some synchronous transactions have gathered quorum and
+ * write a confirmation entry of the last confirmed transaction.
+ */
+void
+txn_limbo_confirm(struct txn_limbo *limbo)
```
Gerold103, line 670: The function is never used outside of `txn_limbo.c` which means we can make it `static`. Lets do this.

changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/raft
```
Gerold103, line 1: Sorry, I just realized - in the commit message lets change the topic from `raft:` to `limbo:`. Raft didn't change here. It is located in `src/lib/raft` and in `src/box/raft.h+c`.

src/box/txn_limbo.c
```diff
@@ -649,10 +662,66 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
...
+		for (; !rlist_entry_is_head(e, limbo->queue, in_queue);
+		       e = rlist_next_entry(e, in_queue))
+			if (txn_has_flag(e->txn, TXN_WAIT_ACK))
+				break;
+		limbo->confirmed_entry = rlist_prev_entry(e, in_queue);
+		/*
+		 * There is no next transaction or the next transaction has no
+		 * lsn. Therefore, confirmed_entry cannot be considered valid.
+		 */
+		if (rlist_entry_is_head(e, limbo->queue, in_queue) || e->lsn == -1)
```
Gerold103, line 689: How can `e` be the head, if on the line 678 you did an early exit with that condition, and since then `e` was only moving forwards (towards the tail)? I added `assert(!rlist_entry_is_head(e, limbo->queue, in_queue));` and tests passed. Does it mean the `if` can be simplified?

test/unit/vclock.cc
```diff
@@ -444,17 +444,82 @@ test_minmax(void)
...
+#define test(s, n, expected) ({						\
+	struct vclock vc;						\
+	vclock_create(&vc);						\
+	vclock_from_string(&vc, (s));					\
+	is(vclock_nth_element(&vc, n), expected,			\
+		"vclock_nth_element \"%s\", %d => %d", s, n, expected);	\
+})
+
+int
+test_nth_element(void)
```
Gerold103, line 456: Please, make this and the other function `static`.

Gerold103 (changes requested):

src/box/txn_limbo.c
Gerold103: Ah, ok, I see, this clears many things out, thanks 🙏.

src/box/txn_limbo.h
Gerold103: From what I see, because of this check you had to introduce the flag, and check `!rlist_is_head` in many places. Seems like not really better than checking for `==/!= NULL`, hm? Would you please try to delete the dummy entry and respond here if you find any specifically ugly places?

src/box/txn_limbo.c
Gerold103: Right, thanks for the explanation.

src/box/txn_limbo.c
Gerold103: Yeah, missed that, thanks.

src/box/txn_limbo.c
Astronomax: I was wrong in that comment about the invariant (fixed). There should be an inverse inequality. `confirmed_lsn <= confirmed_entry->lsn`.
> Here this lsn is always == limbo->confirmed_lsn.

It seems that this is not true. Let's imagine that several fibers executed their transactions, then collected confirmations. At this point, WAL slowed down greatly and each of the fibers got stuck in `txn_limbo_write_confirm`. Then the confirmation for the first transaction was written to disk and called `txn_limbo_read_confirm` here lsn will be strictly less than `confirmed_entry->lsn`. I'll try to reproduce this later.
But I agree that this check can be simplified to `if (limbo->confirmed_entry->lsn == lsn)`.

src/box/txn_limbo.c
Astronomax: As I see `rlist`'s tail - his head.

src/box/txn_limbo.c
Astronomax: - I was wrong in that comment about the invariant (fixed). There should be an inverse inequality. `confirmed_lsn <= confirmed_entry->lsn`. Therefore, in the case where `confirmed_lsn < confirmed_entry->lsn` this may well happen.
- > If somehow the aborted txn was the confirmed one.

But this cannot be. Aborted txn can be `confirmed_entry`, but not a confirmed transaction. Yes, this is bad naming, as I already said. `confirmed_entry` can point to an asynchronous transaction located after the last **actually** confirmed synchronous transaction. Let's rename it something so no one gets confused.
> Otherwise if the aborted txn was after the confirmed LSN, there seem to be no reason why it would affect the confirmed entry.

But we have an invariant `confirmed_entry_is_valid == true` <=> the next transaction after `confirmed_entry` is synchronous and has lsn (not -1) and `ack_count` stores the correct value. Here it breaks down.

test/unit/vclock.cc
Astronomax: Fixed.

changelogs/unreleased/gh-9917-speed-up-transaction-queue-processing.md
Astronomax: Fixed.

src/box/txn_limbo.c
Astronomax: Fixed.

src/box/txn_limbo.c
Astronomax: Fixed.

src/box/txn_limbo.h
Astronomax: This avoided several checks due to the fact that `head_entry` has lsn = -1. But that was when the `txn_limbo_update_confirmed_entry_after_rollback` function was. Now this does not matter, there will be the same number of checks, you’re right.

andreyaksenov (approved):

src/box/txn_limbo.c
Astronomax: One more comment about this invariant `confirmed_lsn <= confirmed_entry->lsn`: 
I just found out that it is only valid if the following patch is applied:
```diff
diff --git a/src/box/txn_limbo.c b/src/box/txn_limbo.c
index 458b3027f..97f50886e 100644
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -708,13 +718,17 @@ txn_limbo_confirm(struct txn_limbo *limbo)
        assert(e->lsn != -1);
        int32_t k = (int32_t)vclock_size(&limbo->vclock)
                - replication_synchro_quorum;
-       int64_t confirm_lsn = (k < 0) ? 0 :
+       int64_t confirm_lsn1 = (k < 0) ? 0 :
                vclock_nth_element(&limbo->vclock, k);
-       assert(confirm_lsn != -1);
+       assert(confirm_lsn1 != -1);
+       int64_t confirm_lsn2 = -1;
        for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
-              e = rlist_next_entry(e, in_queue))
-               if (e->lsn == -1 || e->lsn > confirm_lsn)
+              e = rlist_next_entry(e, in_queue)) {
+               if (e->lsn == -1 || e->lsn > confirm_lsn1)
                        break;
+               if (txn_has_flag(e->txn, TXN_WAIT_ACK))
+                       confirm_lsn2 = e->lsn;
+       }
        limbo->confirmed_entry = rlist_prev_entry(e, in_queue);
        /*
         * We have not made sure that the transaction following confirmed_entry
@@ -722,15 +736,15 @@ txn_limbo_confirm(struct txn_limbo *limbo)
         * valid.
         */
        limbo->confirmed_entry_is_valid = false;
-       txn_limbo_write_confirm(limbo, confirm_lsn);
-       txn_limbo_read_confirm(limbo, confirm_lsn);
+       txn_limbo_write_confirm(limbo, confirm_lsn2);
+       txn_limbo_read_confirm(limbo, confirm_lsn2);
 }
 ```

File: src/box/txn_limbo.c (hunk 13 above)
Comment (Astronomax) [reply src/box/txn_limbo.c]: One more comment about this invariant `confirmed_lsn <= confirmed_entry->lsn`: 
I just found out that it is only valid if the following patch is applied:
```diff
diff --git a/src/box/txn_limbo.c b/src/box/txn_limbo.c
index 458b3027f..97f50886e 100644
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -708,13 +718,17 @@ txn_limbo_confirm(struct txn_limbo *limbo)
        assert(e->lsn != -1);
        int32_t k = (int32_t)vclock_size(&limbo->vclock)
                - replication_synchro_quorum;
-       int64_t confirm_lsn = (k < 0) ? 0 :
+       int64_t confirm_lsn1 = (k < 0) ? 0 :
                vclock_nth_element(&limbo->vclock, k);
-       assert(confirm_lsn != -1);
+       assert(confirm_lsn1 != -1);
+       int64_t confirm_lsn2 = -1;
        for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
-              e = rlist_next_entry(e, in_queue))
-               if (e->lsn == -1 || e->lsn > confirm_lsn)
+              e = rlist_next_entry(e, in_queue)) {
+               if (e->lsn == -1 || e->lsn > confirm_lsn1)
                        break;
+               if (txn_has_flag(e->txn, TXN_WAIT_ACK))
+                       confirm_lsn2 = e->lsn;
+       }
        limbo->confirmed_entry = rlist_prev_entry(e, in_queue);
        /*
         * We have not made sure that the transaction following confirmed_entry
@@ -722,15 +736,15 @@ txn_limbo_confirm(struct txn_limbo *limbo)
         * valid.
         */
        limbo->confirmed_entry_is_valid = false;
-       txn_limbo_write_confirm(limbo, confirm_lsn);
-       txn_limbo_read_confirm(limbo, confirm_lsn);
+       txn_limbo_write_confirm(limbo, confirm_lsn2);
+       txn_limbo_read_confirm(limbo, confirm_lsn2);
 }
 ```
I suggest not guaranteeing this invariant, but simply leaving this check as is.

File: src/box/txn_limbo.h (hunk 3 above)
Comment (Gerold103) [reply src/box/txn_limbo.h]: Lets then drop the dummy entry? (Sorry, didn't see the new patch version yet, just looking at comments!)

File: src/box/txn_limbo.c (hunk 12 above)
Comment (Gerold103) [reply src/box/txn_limbo.c:214]: Hm. Yeah, ok, the naming breaks my brain 😵‍💫. This invariant is very hard to keep in the head when it is not about confirmed entry itself, but about the next one. Have you considered what happens if you store the pointer at that next entry instead? It would be less cognitive load if the entry and its guarantees would be one. Not like entry is one, and the guarantees are for another entry related to the first one somehow.

Comment (Astronomax) [reply src/box/txn_limbo.c:214]: You can think of it this way:
That is, it’s just `confirmed_entry` is a transaction such that all previous ones are committed. You can probably call it `committed_entry` or `last_committed_entry` (by committed I mean that it can no longer be rolled back). This fully expresses the meaning of this variable.

**Review by Gerold103 — COMMENTED** [review]
Thanks for the fixes! Yep, that looks simpler, good stuff 🔥.

File: src/box/txn_limbo.c (hunk 13 above)
Comment (Astronomax) [reply src/box/txn_limbo.c]: I read my comments a month later and it took me a while to understand what I meant :sweat_smile:. Therefore, I decided to write in a little more detail why, for example, the invariant would break without this patch. For myself and perhaps for those who will read this patch. The thing is that a situation is possible when the master has not yet assigned the lsn to the transaction (lsn = -1), while the transaction has already been replicated to a quorum and an ACK is obtained. It sounds counterintuitive and doesn't immediately come to mind.
_See comment below_:
```c
	for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
	       e = rlist_next_entry(e, in_queue))
		if (e->lsn == -1 || e->lsn > confirm_lsn)
			break;
	limbo->confirmed_entry = rlist_prev_entry(e, in_queue);
	/** 
	 * rlist_prev_entry(e, in_queue)->lsn (limbo->confirmed_entry)
	 * may be < confirm_lsn here in case master has not yet assigned
	 * lsn to the transaction (lsn = -1)
	 */
	limbo->confirmed_entry_is_valid = false;
	txn_limbo_write_confirm(limbo, confirm_lsn);
	txn_limbo_read_confirm(limbo, confirm_lsn);
```

src/box/txn_limbo.c
```diff
@@ -53,6 +53,9 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->promote_greatest_term = 0;
 	latch_create(&limbo->promote_latch);
 	limbo->confirmed_lsn = 0;
+	limbo->confirmed_entry =
+		rlist_entry(&limbo->queue, struct txn_limbo_entry, in_queue);
```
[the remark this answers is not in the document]
Astronomax: This is not the same as cast `list<object*>*` to `object*`, because `rlist` is an intrusive list.
`rlist_entry` simply shifts the pointer to {offset of member} bytes back. `rlist` head is the same anchor as any that is inside a list element. So there's nothing stopping us from thinking that head is an anchor inside some dummy element. I never dereference it, so there's no problem.
```c
/**
 * return entry by list item
 */
#define rlist_entry(item, type, member) ({				\
	(type *)( (char *)(item) - offsetof(type,member));		\
})
```
https://github.com/tarantool/small/blob/90a365fdb65da3ffc7b233fe1777a5055fda4ce7/include/small/rlist.h#L275-L280

src/box/txn_limbo.c
```diff
@@ -649,6 +662,63 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
...
+	if (limbo->is_in_rollback)
+		return;
+	struct txn_limbo_entry *e;
+	if (!limbo->confirmed_entry_is_valid) {
+		/* Try to find the next valid confirmed_entry value. */
+		e = rlist_next_entry(limbo->confirmed_entry, in_queue);
+		if (rlist_entry_is_head(e, &limbo->queue, in_queue))
+			return;
+		for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
+		       e = rlist_next_entry(e, in_queue))
```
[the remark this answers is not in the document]
Astronomax: `rlist` always looks like this:
**head** <-> item 1 <-> item2 <-> ... <-> item n <-> **head**
(it's cycled in a ring)
So the iteration macro in `rlist.h` looks like this:
```c
/**
 * foreach through all list entries
 */
#define rlist_foreach_entry(item, head, member)				\
	for (item = rlist_first_entry((head), typeof(*item), member);	\
	     !rlist_entry_is_head((item), (head), member);		\
	     item = rlist_next_entry((item), member))
```
https://github.com/tarantool/small/blob/90a365fdb65da3ffc7b233fe1777a5055fda4ce7/include/small/rlist.h#L378-L384
Therefore, the list does not have any specially allocated tail. Even the head is no different from the item (The library user must remember where the head is. It is possible to pass the wrong head and iterate through the shifted list.).

src/box/txn_limbo.c
```diff
@@ -202,6 +204,14 @@ void
...
 {
 	entry->is_rollback = true;
+	if (entry == limbo->confirmed_entry) {
+		limbo->confirmed_entry = rlist_prev_entry(
+			limbo->confirmed_entry, in_queue);
+		limbo->confirmed_entry_is_valid = false;
+	} else if (entry == rlist_next_entry(limbo->confirmed_entry,
+					     in_queue)) {
+		limbo->confirmed_entry_is_valid = false;
+	}
```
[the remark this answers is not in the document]
Astronomax: I once again read all the places where I wrote about the invariant.
> `confirmed_entry_is_valid == true` only when the next transaction after `confirmed_entry` is synchronous and has lsn (not -1) and `ack_count` stores the correct value.
If at any point this is no longer true, `confirmed_entry_id_valid` is immediately set to `false`. This is processed in all places where the limbo is modified in any way.

> But we have an invariant `confirmed_entry_is_valid == true` <=> the next transaction after `confirmed_entry` is synchronous and has lsn (not -1) and `ack_count` stores the correct value. Here it breaks down.


By calling `txn_limbo_abort` we do not know whether `confirmed_entry_is_valid` is `true`. And we don't check this in any way. And adding this check does not make sense, because it will not simplify anything. 
But what is really true is that if `entry == limbo->confirmed_entry` then `limbo->confirmed_entry_is_valid == false`, so the following patch can be applied:
```diff
void
txn_limbo_abort(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
{
	entry->is_rollback = true;
	if (entry == limbo->confirmed_entry) {
		limbo->confirmed_entry = rlist_prev_entry(
			limbo->confirmed_entry, in_queue);
-		limbo->confirmed_entry_is_valid = false;
+		assert(!limbo->confirmed_entry_is_valid);
	} else if (entry == rlist_next_entry(limbo->confirmed_entry,
					     in_queue)) {
		limbo->confirmed_entry_is_valid = false;
	}
	txn_limbo_pop(limbo, entry);
}
```
Good point, thanks!

src/box/txn_limbo.c
```diff
@@ -649,6 +662,63 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
...
+txn_limbo_confirm(struct txn_limbo *limbo)
+{
+	if (limbo->is_in_rollback)
+		return;
+	struct txn_limbo_entry *e;
+	if (!limbo->confirmed_entry_is_valid) {
+		/* Try to find the next valid confirmed_entry value. */
+		e = rlist_next_entry(limbo->confirmed_entry, in_queue);
+		if (rlist_entry_is_head(e, &limbo->queue, in_queue))
+			return;
```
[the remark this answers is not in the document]
Astronomax: `rlist` always looks like this:
**head** <-> item 1 <-> item2 <-> ... <-> item n <-> **head**
So `confirmed_entry` may point to item n and in this case `confirmed_entry->next == e` will be true.
i.e. what I wrote here is equivalent to `confirmed_entry == rlist_last_entry(&limbo->queue, struct txn_limbo_entry, in_queue)`. I can rewrite this and work with `rlist` strictly through the API as if I don’t know what it looks like under the hood. I don't know what's better in terms of programming practice. Please advise.
It is also possible to rewrite loops according to this convention:
```c
for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);
      e = rlist_next_entry(e, in_queue))
```
to:
```c
for (; !rlist_entry_is_head(e, &limbo->queue, in_queue);) {
	//loop body
	if (e == rlist_last_entry(&limbo->queue, struct txn_limbo_entry, in_queue))
		break;
}
```
(Looks worse to me. You will have to create a separate variable for `rlist_last_entry(&limbo->queue, struct txn_limbo_entry, in_queue)`. It is unlikely that the compiler will be able to understand that this value does not change in the loop. It seems that `rlist_entry_is_head` costs a little less than `rlist_last_entry` (one less dereference)).

src/box/txn_limbo.c
```diff
@@ -649,6 +662,63 @@ txn_limbo_read_demote(struct txn_limbo *limbo, int64_t lsn)
...
+ */
+static void
+txn_limbo_confirm(struct txn_limbo *limbo)
+{
+	if (limbo->is_in_rollback)
+		return;
+	struct txn_limbo_entry *e;
+	if (!limbo->confirmed_entry_is_valid) {
+		/* Try to find the next valid confirmed_entry value. */
+		e = rlist_next_entry(limbo->confirmed_entry, in_queue);
```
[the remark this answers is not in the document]
Astronomax: I really like this idea, but there are some complications: `entry_to_confirm` does not always exist, unlike the previous one before it. And that's the only reason I did it this way.
Let's imagine a situation where there is the following limbo queue:
entry 1, entry2, ..., entry n - **all synchronous and all confirmed, there is nothing after entry n**.
Where will `entry_to_confirm` point? Okay, let it point to NULL. I can patch `txn_limbo_append`:
```diff
txn_limbo_append(struct txn_limbo *limbo, uint32_t id, struct txn *txn)
{
	//...
	struct txn_limbo_entry *e = region_alloc_object(&txn->region,
							typeof(*e), &size);
+	if(limbo->entry_to_confirm == NULL && txn_has_flag(txn, TXN_WAIT_ACK))
+		limbo->entry_to_confirm = e;
}
```
Hmm, maybe this can be done. Always, when I thought about this before, it seemed to me that we could come up with a use case where we would have to constantly run from the beginning along a large queue. But it seems not.


I'll try to rewrite it in this form and see if any pitfalls appear. It seems that even the rollback doesn’t really break anything.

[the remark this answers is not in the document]
Astronomax: It seems to have worked. I wanted to ensure that if we had to iterate in search of entry_to_confirm, so that we would not have to go through the traversed entries a second time (so that each entry was guaranteed to be visited no more than once). And in this solution this is also achieved.

src/box/txn_limbo.c
```diff
@@ -597,6 +600,7 @@ txn_limbo_read_promote(struct txn_limbo *limbo, uint32_t replica_id,
 	limbo->owner_id = replica_id;
 	limbo->confirmed_lsn = vclock_get(&limbo->confirmed_vclock,
 					  replica_id);
+	limbo->entry_to_confirm = NULL;
```
Gerold103, line 603: That looks a bit dangerous. Intuitively I would expect that after `txn_limbo_read_confirm()` the pointer is already updated. But here you update it afterwards manually, inside `txn_limbo_read_promote()`.

Currently it works, but if replicas would start collecting acks, they would have `entry_to_confirm` non-NULL and not updated when the replicas receive a CONFIRM.

Anyway, not principal to me if we fix it now.

src/box/txn_limbo.h
```diff
@@ -169,6 +164,18 @@ struct txn_limbo {
 	 * illegal.
 	 */
 	int64_t confirmed_lsn;
+	/**
+	 * The entry preceding the first unconfirmed synchronous transaction
+	 * or the last entry in the limbo if there are no unconfirmed
+	 * synchronous transactions yet.
+	 */
+	struct txn_limbo_entry *entry_to_confirm;
```
Gerold103, line 172: The comment seems to be describing the old behaviour. Now this points at the first unconfirmed txn itself, not at the preceding entry. The comment below also says `the transaction is the next one immediately after the entry_to_confirm`, but really `ack_count` is for this entry.

Gerold103 (approved):
@Astronomax , impressive 💪. Looks so much simpler now 🔥. Only a nit and a small comment-related remark are left. Proceed with the second review!

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: Thanks for the thorough investigation!

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: Yep, I see now, thanks for explaining!

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: Yes, rlists are a bit confusing when used as a circle.

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: A hack, but an interesting one for sure.

src/box/txn_limbo.h
Astronomax: Fixed.
```diff
diff --git a/src/box/txn_limbo.h b/src/box/txn_limbo.h
index bbf130bc4..6f434c712 100644
--- a/src/box/txn_limbo.h
+++ b/src/box/txn_limbo.h
@@ -165,15 +165,16 @@ struct txn_limbo {
         */
        int64_t confirmed_lsn;
        /**
-        * The entry preceding the first unconfirmed synchronous transaction
-        * or the last entry in the limbo if there are no unconfirmed
-        * synchronous transactions yet.
+        * The first unconfirmed synchronous transaction in the current term.
+        * Is NULL if there is no such transaction, or if the current instance
+        * does not own limbo.
         */
        struct txn_limbo_entry *entry_to_confirm;
        /**
-        * Number of ACKs of the first unconfirmed synchronous transaction in
-        * limbo (the transaction is the next one immediately after the
-        * entry_to_confirm).
+        * Number of ACKs of the first unconfirmed synchronous transaction
+        * (entry_to_confirm->txn). Contains the actual value only for a
+        * non-NULL entry_to_confirm with a local lsn assigned. Otherwise
+        * it may contain any trash.
         */
        int ack_count;
        /**
```

src/lib/vclock/vclock.c
```diff
@@ -175,4 +175,36 @@ vclockset_node_compare(const struct vclock *a, const struct vclock *b)
 	return res;
 }
 
+int64_t
+vclock_nth_element(const struct vclock *vclock, uint32_t n)
```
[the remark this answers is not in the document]
Astronomax: Fixed.

src/box/txn_limbo.c
Astronomax: Let's not fix this now.

src/lib/vclock/vclock.c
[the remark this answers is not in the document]
sergepetrenko: Thanks!

src/box/txn_limbo.c
```diff
@@ -240,19 +250,12 @@ txn_limbo_assign_local_lsn(struct txn_limbo *limbo,
...
-	int ack_count = 0;
-	vclock_foreach(&iter, vc)
-		ack_count += vc.lsn >= lsn;
-	assert(ack_count >= entry->ack_count);
-	entry->ack_count = ack_count;
+	if (entry == limbo->entry_to_confirm) {
+		limbo->ack_count = vclock_count_ge(
+			&limbo->vclock, entry->lsn);
+		if (limbo->ack_count >= replication_synchro_quorum)
+			txn_limbo_confirm(limbo);
```
sergepetrenko, line 257: I see the following segfaults when running the tests:
```
[060] master | #1  0x557dea5a48c5 in crash_collect+256
[060] master | #2  0x557dea5a5301 in crash_signal_cb+100
[060] master | #3  0x76d759e42520 in __sigaction+80
[060] master | #4  0x557dea5ae8d5 in cord_reset_slice+263
[060] master | #5  0x557dea5afdba in fiber_yield_impl+262
[060] master | #6  0x557dea5aff26 in fiber_yield+18
[060] master | #7  0x557dea465a77 in journal_write+40
[060] master | #8  0x557dea466117 in journal_write_row+188
[060] master | #9  0x557dea4107a5 in synchro_request_write+85
[060] master | #10 0x557dea4108fb in txn_limbo_write_synchro+190
[060] master | #11 0x557dea4109e8 in txn_limbo_write_confirm+214
[060] master | #12 0x557dea411552 in txn_limbo_confirm+622
[060] master | #13 0x557dea4100cd in txn_limbo_assign_local_lsn+337
[060] master | #14 0x557dea410113 in txn_limbo_assign_lsn+67
[060] master | #15 0x557dea40a3e2 in txn_on_journal_write+1616
[060] master | #16 0x557dea473ac4 in journal_async_complete+102
[060] master | #17 0x557dea474532 in tx_schedule_queue+54
[060] master | #18 0x557dea4746be in tx_complete_batch+133
[060] master | #19 0x557dea5bad9e in cmsg_deliver+52
[060] master | #20 0x557dea5bb72e in cbus_process+100
[060] master | #21 0x557dea431cf3 in tx_prio_cb(ev_loop*, ev_watcher*, int)+185
[060] master | #22 0x557dea979528 in ev_invoke_pending+205
[060] master | #23 0x557dea97a4ce in ev_run+1529

```

This must be because you started writing CONFIRM from the same fiber which processes a WAL write confirmation (sched).

Previously we explicitly didn't ACK such transactions right on their WAL write confirmation to avoid this. Take a look at txn.c, for example:
https://github.com/tarantool/tarantool/blob/ea6062964828015e5e1f376c6ec2f0b7829b40c2/src/box/txn.c#L1221-L1229
As you can see, instead of ack-ing the transaction from sched right on its WAL write, we ack it from a fiber which owns the transaction.

sergepetrenko (changes requested):
Thanks for working on this!

We really are on the finish line here, but I've found a crash while running the tests.

Please, handle it and rebase the patchset on the current master, and we'll be good to go.

src/box/txn_limbo.c
Gerold103: You can cooperate with @CuriousGeorgiy, he apparently might need to solve the same problem for async commit of sync txns.

CuriousGeorgiy: Working on this in #10626.

Astronomax:
```diff
void
txn_limbo_assign_local_lsn(struct txn_limbo *limbo,
			   struct txn_limbo_entry *entry, int64_t lsn)
{
	assert(limbo->owner_id != REPLICA_ID_NIL);
	assert(limbo->owner_id == instance_id);
	assert(entry->lsn == -1);
	assert(lsn > 0);

	entry->lsn = lsn;

-	if (entry == limbo->entry_to_confirm) {
-		limbo->ack_count = vclock_count_ge(
-			&limbo->vclock, entry->lsn);
-		if (limbo->ack_count >= replication_synchro_quorum)
-			txn_limbo_confirm(limbo);
-	}
+	if (entry == limbo->entry_to_confirm)
+		limbo->ack_count = vclock_count_ge(&limbo->vclock, entry->lsn);
}
```
I fixed it just like that. Since before this patch the behavior was essentially the same - `CONFIRM` will only be written if the limbo owner said ACK, even if enough replicas (>= `replication_synchro_quorum`) have confirmed this transaction.

sergepetrenko (approved):
Thanks for the patch!
LGTM.

sergepetrenko:
Backports:
3.2 #10648
2.11 #10649

PR #10212, opened 2024-07-08
Performance results stabilization

ligurio:


Buristan:
Please, create the similar PR from the main repository. See the rationale in #10151.

Buristan:
See the subsequent discussion in #10219.

PR #10213, opened 2024-07-08
misc: fix compilation errors under clang 18

funbringer:
This PR aims to fix a couple of clang's new pedantic warnings which break the project build.

### DIFF
--- a/src/trivia/util.h
+++ b/src/trivia/util.h
@@ -40,6 +40,7 @@
 #include <inttypes.h>
 #include <assert.h>
 #include <stdbool.h>
+#include <stddef.h>
 
 #if defined(__cplusplus)
 extern "C" {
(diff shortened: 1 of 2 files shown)


src/trivia/util.h
```diff
@@ -40,6 +40,7 @@
 #include <inttypes.h>
 #include <assert.h>
 #include <stdbool.h>
+#include <stddef.h>
```
funbringer, line 43: This is a tricky one. Clang's `stddef.h` provides its own `offsetof`

```c
#if !defined(offsetof) ||                                                      \
    (__has_feature(modules) && !__building_module(_Builtin_stddef))
#define offsetof(t, d) __builtin_offsetof(t, d)
#endif
```

which is a true constexpr unlike `((size_t) &((type *)0)->member)`.

This is what clang 18 has to say about luajit's asserts:

```
tarantool/third_party/luajit/src/lj_obj.h:488:18: note: cast that performs the conversions of a reinterpret_cast is not allowed in a constant expression
  488 | LJ_STATIC_ASSERT(offsetof(Node, val) == 0);
      | ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

locker:
Superseded by #10571.

PR #10218, opened 2024-07-10
lua-yaml: wrap large doubles in quotes

### LINKED ISSUE #10164: yaml: encoding-decoding transforms a string into a number
**System info**

Tarantool 3.1.0-entrypoint-135-gdeaa9599b

* OS: Linux
* OS Version: Gentoo
* Architecture: amd64

**Bug description**

The `yaml` built-in module encodes a string in such a way that decoding it using the same module gives a number.

I don't know whether we should adjust encoding or decoding.

**Steps to reproduce**

```lua
#!/usr/bin/env tarantool

local yaml = require('yaml')

local data = {hostname = '9e0260492063'}
local encoded = yaml.encode(data)
local decoded = yaml.decode(encoded)
print(decoded.hostname)
```

How to run:

Just type `tarantool test.lua`.

**Actual behavior**

```
$ tarantool test.lua
inf
```

**Expected behavior**

```
$ tarantool test.lua
9e0260492063
```

**Additional info**

Tarantool 1.10.15 quotes the string on encoding:

```yaml
Tarantool 1.10.15-0-ge14d0b4eab

tarantool> yaml = require('yaml')
tarantool> print(yaml.encode({hostname = '9e0260492063'}))
---
hostname: '9e0260492063'
...
```

So, this problem looks like a regression in encoding.

georgiy-belyanin:
Since tarantool/luajit@a16313f large exponent double strings are not considered convertable to number. It broke encoding lua objects to YAML because single quotes weren't considered neccessary for decoding.

This fix adds wrapping of every string containing double values equal to infinity into a single quotes.

Closes #10164

### DIFF
--- a/changelogs/unreleased/gh-10164-yaml-encoding-large-exps.md
+++ b/changelogs/unreleased/gh-10164-yaml-encoding-large-exps.md
@@ -1,0 +1,4 @@
+## bugfix/lua/yaml
+
+* Strings with large exponential values equal to infinity are now encoded as
+  strings instead of numbers (gh-10164).
--- a/third_party/lua-yaml/lyaml.cc
+++ b/third_party/lua-yaml/lyaml.cc
@@ -166,6 +166,54 @@ yaml_is_null(const char *str, size_t len)
    return false;
 }
 
+/**
+ * Verify whether a string represents a number literal in YAML.
+ *
+ * Non-standard:
+ *
+ * False-positives:
+ * - 'inf', 'nan' literals despite the case are parsed as numbers
+ *   (the standard specifies only 'inf', 'Inf', 'INF', 'nan',
+ *   'NaN', 'NAN').
+ * - 'infinity' (ignoring case) is considered a number.
+ * - Binary literals ('0b...') are considered numbers.
+ *
+ * Bugs:
+ * - Octal numbers are not supported.
+ *
+ * This function is used only in encoding for wrapping strings
+ * containing number literals in quotes to make YAML parser
+ * handle them as strings. It means false-positives will lead to
+ * extra quotation marks and are not dangerous at all.
+ *
+ * @param str Literal to check.
+ * @param len Length of @a str.
+ *
+ * @retval Whether @a str represents a number value.
+ */
+static inline bool
+yaml_is_number(const char *str, size_t len, struct lua_State *L)
+{
+   /*
+    * TODO: Should be implemented with the literal parser
+    * instead of using strtod() and lua_isnumber().
+    * Using parser will make it possible to remove the third
+    * argument.
+    */
+   if (len == 0)
+      return false;
+
+   if (lua_isnumber(L, -1))
+      return true;
+
+   char *endptr = NULL;
+   fpconv_strtod(str, &endptr);
+   if (endptr == str + len)
+      return true;
+
+   return false;
+}
+
 static void generate_error_message(struct lua_yaml_loader *loader) {
    char buf[256];
    luaL_Buffer b;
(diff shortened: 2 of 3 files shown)


p7nov (approved):

third_party/lua-yaml/lyaml.cc
```diff
@@ -682,8 +684,10 @@ static int dump_node(struct lua_yaml_dumper *dumper)
    case MP_STR:
       str = field.sval.data;
       len = field.sval.len;
+      dval = fpconv_strtod(str, &endptr);
+
       if (yaml_is_null(str, len) || yaml_is_bool(str, len, &unused) ||
-          lua_isnumber(dumper->L, -1)) {
+          lua_isnumber(dumper->L, -1) || (isinf(dval) && endptr == str + len)) {
```
Totktonada, line 690: I propose to wrap the check into the `yaml_is_number` function and call `fpconv_strtod` only when the `lua_isnumber` check fails.

It is OK to accept any value that is accepted by `fpconv_strtod`, not only a large one that is converted to `inf`. It would make the idea simpler, IMHO, and would require less explanation in comments.

I also propose to explain why and how this function works in comments. Key points:

* The idea is to catch all the string values that may be interpreted as a number by some decoder. It is OK to have a kind of false-positives: it means that these values are quoted on encoding that is harmless.
* Our decoder interprets unquoted values that are accepted by `strtoll`, `strtoull` and `strtod` as numbers. We must quote all such values in the encoder.
* `lua_isnumber` seems enough on the first glance, but it discards values with a large exponent, which are interpreted as `inf` in our decoder.
* `fpconv_strtod` seems enough on the first glance, but it doesn't accept binary `0b<...>` literals and since they're quoted by older tarantool versions it is undesirable to leave them unquoted in the newer ones.
* The best solution would be a format check without taking into account how large the significand and the exponent. (It can check first N characters for performance and simplicity, because false-positives are OK to some extent.) It needs to implement a parser that supports binary/dec/hex/(octal?) integral/floating point number literals in plain/scientific notations, so it is postponed.

Totktonada (approved):
I'm generally OK, but, please, consider some stylistic remarks.

third_party/lua-yaml/lyaml.cc
Totktonada, line 690: NB: I've wrote a simple tonumber/strtod testing code while thinking on this patch.

```lua
local ffi = require('ffi')
local tabulate = require('internal.config.utils.tabulate')

ffi.cdef('double strtod(const char *nptr, char **endptr);')

local function convert_lua(s)
    return tonumber(s)
end

local function convert_c(s)
    local nptr = ffi.cast('char *', s)
    local endptr = ffi.new('char *[1]')
    endptr[0] = nptr + s:len()
    local res = ffi.C.strtod(nptr, endptr)
    if endptr[0] ~= nptr + s:len() then
        return nil
    end
    return res
end

local function stringify(x)
    return x == nil and 'nil' or tostring(x)
end

local function run_case(s)
    return {s, stringify(convert_lua(s)), stringify(convert_c(s))}
end

local cases = {
    '1',
    '2e3',
    '0x2p1',
    '0x2p-1',
    '0x2p+1',
    '9e0260492063',
    'abc',
    '077',
    '0b11',
}

local res = {
    {'input', 'tonumber', 'strtod'},
}
for _, case in ipairs(cases) do
    table.insert(res, run_case(case))
end
print(tabulate.encode(res))
```

changelogs/unreleased/gh-10164-yaml-encoding-large-exps.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/lua-yaml
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied. Thank you.

third_party/lua-yaml/lyaml.cc
georgiy-belyanin: That'll be quite consistent. Added `yaml_is_number` (that works the same way yaml is decoded). Specified the difference from the YAML standard in the docs.

locker (approved):

PR #10219, opened 2024-07-10
 Performance results stabilization

ligurio:


perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+echo -n "Disable AMD SMT or Intel Hyperthreading "
+sysfs_path="/sys/devices/system/cpu/smt/active"
+if [ -f "$sysfs_path" ]; then
+  is_set=$(cat $sysfs_path)
+  [ "$is_set" = 1 ] && echo "SUCCESS" || echo "FAILED (hint: set 'nosmt' kernel parameter)"
+else
+  echo "SKIPPED"
+fi
+
+echo -n "Isolate CPUs for benchmarking "
```
Buristan, line 43: Minor: Maybe it is worth it to introduce the function that takes the path and hint message as arguments (for options `nosmt`, `isolcpus`, `irqaffinity`, `nohz_full`). Also, it will be nice to sort entries alphabetically.
Feel free to ignore.

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+sysfs_path="/sys/devices/system/cpu/isolated"
+if [ -f "$sysfs_path" ]; then
+  isolcpus=$(cat "$sysfs_path")
+  [ -n "$isolcpus" ] && echo "SUCCESS" || echo "FAILED (hint: set 'isolcpus' kernel parameter)"
```
Buristan, line 47: Minor: Since fail doesn't mean stop of the script, I suggest renaming `FAILED` to the `WARNING` or something like this. Similary, we may rename `SUCCESS` -> `CHECKED`. Here and below for checking routines.
Feel free to ignore.

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+else
+  echo "SKIPPED"
+fi
+
+echo -n "Offload interrupts from the isolated CPUs "
+sysfs_path="cat /proc/irq/default_smp_affinity"
```
Buristan, line 53: This is not a path, but the content of the file. Thus, the next line is incorrect.
See also [this run](https://github.com/tarantool/tarantool/actions/runs/9911380869/job/27383926994?pr=10219#step:7:44).

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+  nohz=$(cat "$sysfs_path")
+  [ -n "$nohz" ] && echo "SUCCESS" || echo "FAILED (hint: set 'nohz_full' kernel parameter)"
+else
+  echo "FAILED (hint: enable CONFIG_NO_HZ_FULL in a kernel config)"
+fi
+
+echo -n "Disable transparent_hugepage "
+sysfs_path="/sys/kernel/mm/transparent_hugepage/enabled"
+if [ -f "$sysfs_path" ]; then
+  sudo sh -c "echo never > $sysfs_path"
```
Buristan, line 73: Do we need `sudo` since the script is running as root?
Here and below.

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+  if [ -f "$sysfs_path" ]; then
+    sudo sh -c "echo 1 > $sysfs_path"
+    echo "SUCCESS"
+  else
+    echo "SKIPPED"
+  fi
+fi
+
+echo -n "Stabilize the CPU frequency "
+sys_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
```
Buristan, line 112: Typo: s/sys_path/sysfs_path/
See also [this run](https://github.com/tarantool/tarantool/actions/runs/9911380869/job/27383926994?pr=10219#step:7:72).

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+if [ -f "$sysfs_path" ]; then
+  sh -c 'echo performance | sudo tee $sysfs_path'
+  echo "SUCCESS"
+else
+  echo "SKIPPED"
+fi
+
+echo -n "Workload-specific: drop disk caches "
+proc_path="/proc/sys/vm/drop_caches"
```
Buristan, line 121: Typo: s/proc_path/sysfs_path/
See also [this run](https://github.com/tarantool/tarantool/actions/runs/9911380869/job/27383926994?pr=10219#step:7:78).

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+else
+  echo "SKIPPED"
+fi
+
+###
+# Prepare the OS environment.
+###
+
+echo -n "Workload-specific: Swappiness "
+sudo sh -c "sysctl -w vm.swappiness=10 2>/dev/null"
```
Buristan, line 134: Do we need this since it is workflow specific?
I suggest dropping this lines for now.

Buristan (changes requested):
Hi, Sergey!
Thanks for the patch set!
I'll proceed with a review per patch.

**[PATCH 1/4] perf: add a script for setting environment**

> "Benchmarking" article [0] in Tarantool's wiki contains a lot

Typo: s/a lot/a lot of/

> recommendations that helps to setup Linux operating system and

Typo: s/helps/help/
Typo: s/Linux/the Linux/

> This recommendations written in a plain text with examples of commands

Typo: s/This/These/
Typo: s/in a plain text/in plain text/

> setup test environment automatically before running benchmarks.

Typo: s/test environment/the test environment/

> There are many guides with benchmarking tips but unfortunately

Typo: s/tips but unfortunately/tips, but unfortunately,/

> The patch add a script for setting environment before running

Typo: s/add/adds/
Typo: s/environment/the environment/

> performance tests. All settings used in proposed script are

Typo: s/proposed script/the proposed script.

> described in the article [1]. Note, that uncertain settings were

Typo: s/[1]/[3]

Also, it looks like links [2] and [3] are switched.

perf/tools/setup_env.sh
ligurio: I like the idea, thanks.
Introduced variables with user messages:
```diff
--- a/perf/tools/setup_env.sh
+++ b/perf/tools/setup_env.sh
@@ -27,6 +27,10 @@ else
    exit 1
 fi
 
+FAILURE_MSG="WARNING"
+SUCCESS_MSG="CHECKED"
+SKIPPED_MSG="SKIPPED"
+
 ###
 # Kernel command line parameters.
 ###
```
and these variables used in all messages in a whole script.

perf/tools/setup_env.sh
ligurio: Right, fixed.
```diff
 echo -n "Offload interrupts from the isolated CPUs "
-sysfs_path="cat /proc/irq/default_smp_affinity"
+sysfs_path="/proc/irq/default_smp_affinity"
 if [ -f "$sysfs_path" ]; then
   smp_affinity=$(cat "$sysfs_path")
-  [ -n "$smp_affinity" ] && echo "SUCCESS" || echo "FAILED (hint: set 'irqaffinity' kernel parameter)"
+  [ -n "$smp_affinity" ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'irqaffinity' kernel parameter)"
 else
   echo "SKIPPED"
 fi
```

perf/tools/setup_env.sh
ligurio: Removed `sudo` everywhere.
I suspect check for `$(id -u) -ne 0` is enough.

perf/tools/setup_env.sh
ligurio: Fixed, thanks.

perf/tools/setup_env.sh
ligurio: Fixed, thanks!

perf/tools/setup_env.sh
ligurio: Okay, let's postpone it. Removed for now.

perf/tools/setup_env.sh
ligurio: Introduced two functions: setting to a value and check a value for non-zero:
```diff
diff --git a/perf/tools/setup_env.sh b/perf/tools/setup_env.sh
index b29fcad3dc..4be7f7bfa9 100755
--- a/perf/tools/setup_env.sh
+++ b/perf/tools/setup_env.sh
@@ -31,6 +31,33 @@ FAILURE_MSG="WARNING"
 SUCCESS_MSG="CHECKED"
 SKIPPED_MSG="SKIPPED"
 
+set_kernel_setting() {
+  desc_msg="$1"
+  file_path="$2"
+  value="$3"
+
+  echo -n "$desc_msg "
+  if [ -f "$file_path" ]; then
+    sh -c "echo $value > $file_path" && echo $SUCCESS_MSG || echo $FAILURE_MSG
+  else
+    echo $SKIPPED_MSG
+  fi
+}
+
+kernel_setting_is_nonzero() {
+  desc_msg="$1"
+  file_path="$2"
+  hint_msg="$3"
+
+  echo -n "$desc_msg "
+  if [ -f "$file_path" ]; then
+    value=$(cat "$file_path")
+    [ -n "$value" ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: $hint_msg)"
+  else
+    echo $SKIPPED_MSG
+  fi
+}
+
 ###
 # Kernel command line parameters.
 ###
@@ -41,91 +68,50 @@ if [ -f "$sysfs_path" ]; then
   is_set=$(cat $sysfs_path)
   [ "$is_set" = 1 ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'nosmt' kernel parameter)"
 else
-  echo "SKIPPED"
+  echo $SKIPPED_MSG
 fi
 
-echo -n "Isolate CPUs for benchmarking "
+desc_msg="Isolate CPUs for benchmarking"
 sysfs_path="/sys/devices/system/cpu/isolated"
-if [ -f "$sysfs_path" ]; then
-  isolcpus=$(cat "$sysfs_path")
-  [ -n "$isolcpus" ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'isolcpus' kernel parameter)"
-else
-  echo "SKIPPED"
-fi
+hint_msg="set 'isolcpus' kernel parameter"
+kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
 
-echo -n "Offload interrupts from the isolated CPUs "
+desc_msg="Offload interrupts from the isolated CPUs"
 sysfs_path="/proc/irq/default_smp_affinity"
-if [ -f "$sysfs_path" ]; then
-  smp_affinity=$(cat "$sysfs_path")
-  [ -n "$smp_affinity" ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'irqaffinity' kernel parameter)"
-else
-  echo "SKIPPED"
-fi
+hint_msg="set 'irqaffinity' kernel parameter"
+kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
 
-echo -n "Disable scheduling on single-task isolated CPUs "
+desc_msg="Disable scheduling on single-task isolated CPUs"
 sysfs_path="/sys/devices/system/cpu/nohz_full"
-if [ -f "$sysfs_path" ]; then
-  nohz=$(cat "$sysfs_path")
-  [ -n "$nohz" ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'nohz_full' kernel parameter)"
-else
-  echo "FAILED (hint: enable CONFIG_NO_HZ_FULL in a kernel config)"
-fi
+hint_msg="set 'nohz_full' kernel parameter"
+kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
 
-echo -n "Disable transparent_hugepage "
+desc_msg="Disable transparent huge pages"
 sysfs_path="/sys/kernel/mm/transparent_hugepage/enabled"
-if [ -f "$sysfs_path" ]; then
-  sh -c "echo never > $sysfs_path"
-  echo $SUCCESS_MSG
-else
-  echo $SKIPPED_MSG
-fi
+set_kernel_setting "$desc_msg" "$sysfs_path" "never"
 
-echo -n "Disable transparent_hugepage "
+desc_msg="Disable direct compaction of transparent huge pages "
 sysfs_path="/sys/kernel/mm/transparent_hugepage/defrag"
-if [ -f "$sysfs_path" ]; then
-  sh -c "echo never > $sysfs_path"
-  echo $SUCCESS_MSG
-else
-  echo $SKIPPED_MSG
-fi
+set_kernel_setting "$desc_msg" "$sysfs_path" "never"
 
 ###
 # System tuning.
 ###
 
-echo -n "Disable TurboBoost "
+desc_msg="Disable TurboBoost"
 if [ "$cpu_vendor" = "amd" ]; then
   sysfs_path="/sys/devices/system/cpu/cpufreq/boost"
-  if [ -f "$sysfs_path" ]; then
-    sh -c "echo 0 > $sysfs_path"
-    echo $SUCCESS_MSG
-  else
-    echo $SKIPPED_MSG
-  fi
+  value=0
 elif [ "$cpu_vendor" = "intel" ]; then
   sysfs_path="/sys/devices/system/cpu/intel_pstate/no_turbo"
-  if [ -f "$sysfs_path" ]; then
-    sh -c "echo 1 > $sysfs_path"
-    echo $SUCCESS_MSG
-  else
-    echo $SKIPPED_MSG
-  fi
+  value=1
 fi
+set_kernel_setting "$desc_msg" "$sysfs_path" "$value"
 
-echo -n "Stabilize the CPU frequency "
+desc_msg="Stabilize the CPU frequency"
 sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
-if [ -f "$sysfs_path" ]; then
-  sh -c 'echo performance | tee $sysfs_path'
-  echo $SUCCESS_MSG
-else
-  echo $SKIPPED_MSG
-fi
+set_kernel_setting "$desc_msg" "$sysfs_path" "performance"
 
-echo -n "Workload-specific: drop disk caches "
+desc_msg="Workload-specific: drop disk caches"
 proc_path="/proc/sys/vm/drop_caches"
-if [ -f "$proc_path" ]; then
-  sh -c "echo 1 > $proc_path"
-  echo $SUCCESS_MSG
-else
-  echo $SKIPPED_MSG
-fi
+set_kernel_setting "$desc_msg" "$proc_path" 1
```

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+fi
+
+echo -n "Isolate CPUs for benchmarking "
+sysfs_path="/sys/devices/system/cpu/isolated"
+if [ -f "$sysfs_path" ]; then
+  isolcpus=$(cat "$sysfs_path")
+  [ -n "$isolcpus" ] && echo "SUCCESS" || echo "FAILED (hint: set 'isolcpus' kernel parameter)"
+else
+  echo "SKIPPED"
+fi
```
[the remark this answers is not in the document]
ligurio: Left it as is, because we use `sudo taskset 0xef chrt 50` in workflow (see answer above).

perf/tools/setup_env.sh
Buristan: Maybe it is easier for reading to give arguments like the following (same for `set_kernel_setting()`):
```sh
kernel_setting_is_nonzero \
	"Isolate CPUs for benchmarking" \
	"/sys/devices/system/cpu/isolated" \
	"set 'isolcpus' kernel parameter"

kernel_setting_is_nonzero \
	"Offload interrupts from the isolated CPUs" \
	"/proc/irq/default_smp_affinity" \
	"set 'irqaffinity' kernel parameter"

kernel_setting_is_nonzero \
	"Disable scheduling on single-task isolated CPUs" \
	"/sys/devices/system/cpu/nohz_full" \
	"set 'nohz_full' kernel parameter"
```

What do you think?

perf/tools/setup_env.sh
Buristan: Nice, thanks!

perf/tools/setup_env.sh
```diff
@@ -41,91 +68,50 @@ if [ -f "$sysfs_path" ]; then
...
-  else
-    echo $SKIPPED_MSG
-  fi
+  value=1
 fi
+set_kernel_setting "$desc_msg" "$sysfs_path" "$value"
 
-echo -n "Stabilize the CPU frequency "
+desc_msg="Stabilize the CPU frequency"
 sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
```
Buristan, line 112: The `-f "$filename"` doesn't work correctly with wildcard symbol:

I've checked that locally:

```sh
desc_msg="TEST MASK:"
sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "TESTMASK:"

desc_msg="TEST RAW:"
sysfs_path="/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor"
kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "TESTRAW:"

# ...

TEST MASK: SKIPPED
TEST RAW: CHECKED
```

See [this run](https://github.com/tarantool/tarantool/actions/runs/9939553668/job/27454348529?pr=10219#step:7:112).

perf/tools/setup_env.sh
ligurio: Good idea, updated.

ligurio:
```diff
diff --git a/perf/tools/setup_env.sh b/perf/tools/setup_env.sh
index 4be7f7bfa9..792a228ea3 100755
--- a/perf/tools/setup_env.sh
+++ b/perf/tools/setup_env.sh
@@ -66,39 +66,41 @@ echo -n "Disable AMD SMT or Intel Hyperthreading "
 sysfs_path="/sys/devices/system/cpu/smt/active"
 if [ -f "$sysfs_path" ]; then
   is_set=$(cat $sysfs_path)
-  [ "$is_set" = 1 ] && echo $SUCCESS_MSG || echo "$FAILURE_MSG (hint: set 'nosmt' kernel parameter)"
+  msg="$FAILURE_MSG (hint: set 'nosmt' kernel parameter)"
+  [ "$is_set" = 1 ] && echo $SUCCESS_MSG || echo "$msg"
 else
   echo $SKIPPED_MSG
 fi
 
-desc_msg="Isolate CPUs for benchmarking"
-sysfs_path="/sys/devices/system/cpu/isolated"
-hint_msg="set 'isolcpus' kernel parameter"
-kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
+kernel_setting_is_nonzero \
+  "Isolate CPUs for benchmarking" \
+  "/sys/devices/system/cpu/isolated" \
+  "set 'isolcpus' kernel parameter"
 
-desc_msg="Offload interrupts from the isolated CPUs"
-sysfs_path="/proc/irq/default_smp_affinity"
-hint_msg="set 'irqaffinity' kernel parameter"
-kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
+kernel_setting_is_nonzero \
+  "Offload interrupts from the isolated CPUs" \
+  "/proc/irq/default_smp_affinity" \
+  "set 'irqaffinity' kernel parameter"
 
-desc_msg="Disable scheduling on single-task isolated CPUs"
-sysfs_path="/sys/devices/system/cpu/nohz_full"
-hint_msg="set 'nohz_full' kernel parameter"
-kernel_setting_is_nonzero "$desc_msg" "$sysfs_path" "$hint_msg"
+kernel_setting_is_nonzero \
+  "Disable scheduling on single-task isolated CPUs" \
+  "/sys/devices/system/cpu/nohz_full" \
+  "set 'nohz_full' kernel parameter"
 
-desc_msg="Disable transparent huge pages"
-sysfs_path="/sys/kernel/mm/transparent_hugepage/enabled"
-set_kernel_setting "$desc_msg" "$sysfs_path" "never"
+set_kernel_setting \
+  "Disable transparent huge pages" \
+  "/sys/kernel/mm/transparent_hugepage/enabled" \
+  "never"
 
-desc_msg="Disable direct compaction of transparent huge pages "
-sysfs_path="/sys/kernel/mm/transparent_hugepage/defrag"
-set_kernel_setting "$desc_msg" "$sysfs_path" "never"
+set_kernel_setting \
+  "Disable direct compaction of transparent huge pages" \
+  "/sys/kernel/mm/transparent_hugepage/defrag" \
+  "never"
 
 ###
 # System tuning.
 ###
 
-desc_msg="Disable TurboBoost"
 if [ "$cpu_vendor" = "amd" ]; then
   sysfs_path="/sys/devices/system/cpu/cpufreq/boost"
   value=0
@@ -106,12 +108,17 @@ elif [ "$cpu_vendor" = "intel" ]; then
   sysfs_path="/sys/devices/system/cpu/intel_pstate/no_turbo"
   value=1
 fi
-set_kernel_setting "$desc_msg" "$sysfs_path" "$value"
-
-desc_msg="Stabilize the CPU frequency"
-sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
-set_kernel_setting "$desc_msg" "$sysfs_path" "performance"
-
-desc_msg="Workload-specific: drop disk caches"
-proc_path="/proc/sys/vm/drop_caches"
-set_kernel_setting "$desc_msg" "$proc_path" 1
+set_kernel_setting \
+  "Disable TurboBoost" \
+  "$sysfs_path" \
+  "$value"
+
+set_kernel_setting \
+  "Stabilize the CPU frequency" \
+  "/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor" \
+  "performance"
+
+set_kernel_setting \
+  "Workload-specific: drop disk caches" \
+  "/proc/sys/vm/drop_caches" \
+  "1"
```

perf/tools/setup_env.sh
ligurio: Updated:

```diff
@@ -113,10 +114,14 @@ set_kernel_setting \
   "$sysfs_path" \
   "$value"
 
-set_kernel_setting \
-  "Stabilize the CPU frequency" \
-  "/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor" \
-  "performance"
+sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
+for sysfs_path_cpu in $sysfs_path; do
+  cpu_n=$(echo "$sysfs_path_cpu" | cut -d'/' -f6)
+  set_kernel_setting \
+    "Stabilize the CPU frequency $sysfs_path_cpu" \
+    "$sysfs_path_cpu" \
+    "performance"
+done
 
 set_kernel_setting \
   "Workload-specific: drop disk caches" \
```

perf/tools/setup_env.sh
ligurio: >  Also, it will be nice to sort entries alphabetically.

The order is the same as in benchmarking page. Do you want to have another order in the script?

Buristan: I don't insist, feel free to ignore.

perf/tools/setup_env.sh
Buristan: Locally, it works like a charm, but we still have troubles in CI, see [this run](https://github.com/tarantool/tarantool/actions/runs/9944968058/job/27474134158#step:7:103).

Buristan:
**[PATCH 4/5] perf/lua: print user-friendly error when module is not found**

Patch is generally LGTM, but the commit title is wider than 50 symbols.
I suggest the following title `perf/lua: protect require of column module`. Information about user-friendly error may be placed in the next paragraph.

Buristan:
**[PATCH 2/5] perf: introduce BENCH_CMD environment variable**

LGTM.

perf/tools/setup_env.sh
ligurio: Shell will not unfold mask when path doesn't exist:

On my personal machine (Ubuntu 22.04):

```
$ stat /sys/devices/system/cpu/cpu1/cpufreq/scaling_governor
  File: /sys/devices/system/cpu/cpu1/cpufreq/scaling_governor
  Size: 4096            Blocks: 0          IO Block: 4096   regular file
Device: 16h/22d Inode: 24057       Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2024-07-04 15:15:37.957732630 +0300
Modify: 2024-07-16 11:09:29.189046197 +0300
Change: 2024-07-16 11:09:29.189046197 +0300
 Birth: -
$ for cpu in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do echo $cpu; done
/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu1/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu2/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu3/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu4/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu5/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu6/cpufreq/scaling_governor
/sys/devices/system/cpu/cpu7/cpufreq/scaling_governor
$ 
```

On ghacts-tarantool-8-16-n5 (Ubuntu 20.04):

```
$ stat /sys/devices/system/cpu/cpu1/cpufreq/scaling_governor
stat: cannot stat '/sys/devices/system/cpu/cpu1/cpufreq/scaling_governor': No such file or directory
$ for cpu in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do echo $cpu; done
/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor
```

So I've updated this check in the script:

```diff
--- a/perf/tools/setup_env.sh
+++ b/perf/tools/setup_env.sh
@@ -119,11 +119,11 @@ set_kernel_setting \
   "$sysfs_path" \
   "$value"
 
-sysfs_path="/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor"
-for sysfs_path_cpu in $sysfs_path; do
-  cpu_n=$(echo "$sysfs_path_cpu" | cut -d'/' -f6)
+ncpu=$(getconf _NPROCESSORS_ONLN)
+for cpu_id in $(seq 0 1 $((ncpu-1))); do
+  sysfs_path_cpu="/sys/devices/system/cpu/cpu$cpu_id/cpufreq/scaling_governor"
   set_kernel_setting \
-    "Stabilize the frequency of CPU $cpu_n" \
+    "Stabilize the frequency of CPU $cpu_id" \
     "$sysfs_path_cpu" \
     "performance"
 done
```

However, on the aforementioned machine (ghacts-tarantool-8-16-n5) check is still skipped. The reason is missed SYSFS settings `scaling_governor` for CPUs. As I got it right, it is due to absence of CPU scaling driver [1]. I continue an investigation.

1. https://docs.kernel.org/admin-guide/pm/cpufreq.html

.github/workflows/perf_micro.yml
```diff
@@ -78,6 +78,17 @@ jobs:
...
+        env:
+          # The taskset alone will pin all the process threads
+          # into a single (random) isolated CPU, see
+          # https://bugzilla.kernel.org/show_bug.cgi?id=116701.
+          # The workaround is using realtime scheduler for the
+          # isolated task using chrt, e. g.:
+          # sudo taskset 0xef chrt 50.
+          # But this makes the process use non-standard, real-time
+          # round-robin scheduling mechanism.
+          BENCH_CMD: "sudo taskset 0xfe chrt 50"
```
Buristan, line 90: Do we need `sudo` here since we ran the previous step as root?

.github/workflows/perf_micro.yml
```diff
@@ -78,6 +78,17 @@ jobs:
...
+          VERBOSE: 1
```
Buristan, line 91: I suppose this should be gone after debugging.

Buristan (changes requested):
**[PATCH 3/5] ci: enable BENCH_CMD**

The patch itself is LGTM after fixing my comments below.
I suggest adding the more verbose commit message and adding a note about `chrt` usage to the commit message too.

perf/tools/setup_env.sh
ligurio: There is no cpufreq driver on ghacts-tarantool-8-16-n5:
```
buntu@ghacts-tarantool-8-16-n5:~$ cpufreq-info
cpufrequtils 008: cpufreq-info (C) Dominik Brodowski 2004-2009
Report errors and bugs to cpufreq@vger.kernel.org, please.
analyzing CPU 0:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 1:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 2:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 3:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 4:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 5:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 6:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
analyzing CPU 7:
  no or unknown cpufreq driver is active on this CPU
  maximum transition latency: 4294.55 ms.
```

.github/workflows/perf_micro.yml
ligurio: Right, I'll remove it.
To be honestly, I would always enable `VERBOSE=1` because it saves a time on debugging in CI.

ligurio: removed

perf/lua/benchmark.lua
```diff
@@ -119,13 +123,37 @@ function M.argparse(arg, argtable, custom_help)
...
 
+local function load_average()
+    local path = '/proc/loadavg'
+    local fh = assert(fio.open(path, {'O_RDONLY'}))
+    local loadavg_buf = fh:read(1024)
+    fh:close()
+    -- Format is '0.89 0.66 0.80 2/1576 1203510'.
+    local loadavg_re = '(%d+.%d+) (%d+.%d+) (%d+.%d+)'
+
+    return { loadavg_buf:gmatch(loadavg_re)() }
```
Buristan, line 134: Why do we need the additional variable `loadavg_re`? Also, I don't understand why do we use `gmatch()` instead of `match()`?

perf/lua/benchmark.lua
```diff
@@ -119,13 +123,37 @@ function M.argparse(arg, argtable, custom_help)
...
+end
+
+local function perf_context()
+    local ctx = {}
```
Buristan, line 138: I suppose that
```lua
return {
    date = os.date('%Y-%m-%d %H:%M:%S'),
    -- ...
}
```
Looks better. Feel free to ignore.

perf/lua/benchmark.lua
```diff
@@ -119,13 +123,37 @@ function M.argparse(arg, argtable, custom_help)
...
+    ctx.date = os.date('%Y-%m-%d %H:%M:%S')
```
Buristan, line 139: The format of the date isn't the same as for Google Benchmark:
```json
{
  "date": "2024-07-16T11:59:25+03:00",

```

vs
```json
{
  "date": "2024-07-16 11:51:59",
```

Buristan (changes requested):
**[PATCH 5/5] perf/lua: add context section to test output**

Generally LGTM, except several nits below.

.github/workflows/perf_micro.yml
Buristan: >  I would always enable VERBOSE=1 because it saves a time on debugging in CI.

I'm not really against it, since debugging CI issues always leads to trouble, and at the same time, it costs nothing.
But it should be done consistently for all jobs, not as the part of this patchset.

perf/tools/setup_env.sh
Buristan: Got it!
Thanks for the investigation!

.github/workflows/perf_micro.yml
ligurio: seems works fine without `sudo`, removed

perf/lua/benchmark.lua
ligurio: Updated:

```diff
--- a/perf/lua/benchmark.lua
+++ b/perf/lua/benchmark.lua
@@ -135,15 +135,14 @@ local function load_average()
 end
 
 local function perf_context()
-    local ctx = {}
-    ctx.date = os.date('%Y-%m-%d %H:%M:%S')
-    ctx.load_avg = load_average()
-    ctx.tarantool_version = tarantool.version
-    ctx.build_target = tarantool.build.target
-    ctx.build_flags = tarantool.build.flags
-    ctx.host_name = io.popen('hostname'):read()
-
-    return ctx
+    return {
+        build_flags = tarantool.build.flags,
+        build_target = tarantool.build.target,
+        date = os.date('%Y-%m-%d %H:%M:%S'),
+        host_name = io.popen('hostname'):read(),
+        load_avg = load_average(),
+        tarantool_version = tarantool.version,
+    }
 end
 
 function M.new(opts)
```

(keys sorted alphabetically!)

perf/lua/benchmark.lua
ligurio: Example of date is `"date": "2015/03/17-18:40:25"`, see [^1]. What is a source of your example?

[^1]: https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats

Buristan: The result of the local run. Do you have different values locally?

ligurio: Updated:

```diff
diff --git a/perf/lua/benchmark.lua b/perf/lua/benchmark.lua
index 7c04ae0867..e76d7bdb6a 100644
--- a/perf/lua/benchmark.lua
+++ b/perf/lua/benchmark.lua
@@ -35,6 +35,7 @@ local json = require('json')
 local fio = require('fio')
 local argparse = require('internal.argparse')
 local tarantool = require('tarantool')
+local datetime = require('datetime')
 
 local M = {}
 
@@ -134,11 +135,23 @@ local function load_average()
     return { string.match(loadavg_buf, loadavg_re) }
 end
 
+-- Google Benchmark reports contains a date in ISO 8061 format,
+-- example: 2013-07-01T17:55:13-07:00.
+-- See an implementation of `LocalDateTimeString()` in
+-- Google Benchmark library source code [1].
+-- The function return a ISO 8061 formatted timestamp.
+--
+-- 1. https://github.com/google/benchmark/blob/65668db27365d29ca4890cb2102e81acb6585b43/src/timers.cc#L209
+-- @return e.g. 2024-07-17T12:46:31+03:00
+local function iso_8061_timestamp()
+    return datetime.new(os.date('*t')):format('%FT%T.%f%z')
+end
+
 local function perf_context()
     return {
         build_flags = tarantool.build.flags,
         build_target = tarantool.build.target,
-        date = os.date('%Y-%m-%d %H:%M:%S'),
+        date = iso_8061_timestamp(),
         host_name = io.popen('hostname'):read(),
         load_avg = load_average(),
         tarantool_version = tarantool.version,
```

perf/tools/setup_env.sh
ligurio: ignored

perf/lua/benchmark.lua
Buristan: Thanks for the fix!

LGTM, after fixing the final nit:
In the created `datetime` object, the timezone isn't specified, so it returns `+0000` instead of `+03:00`. Also, the `:` delimiter is missed.

ligurio: > In the created datetime object, the timezone isn't specified, so it returns +0000 instead of +03:00

For compatibility with Google Benchmark reports and for simplicity, I would not set TZ:

```diff
@@ -144,7 +160,8 @@ end
 -- 1. https://github.com/google/benchmark/blob/65668db27365d29ca4890cb2102e81acb6585b43/src/timers.cc#L209
 -- @return e.g. 2024-07-17T12:46:31+03:00
 local function iso_8061_timestamp()
-    return datetime.new(os.date('*t')):format('%FT%T.%f%z')
+    local dt = os.date('*t')
+    return datetime.new(dt):format('%Y-%m-%dT%H:%M:%S') .. '+00:00'
 end
 
 local function perf_context()
```

```lua
tarantool> dt = os.date('*t')
---
...

tarantool> datetime.new(dt):format('%Y-%m-%dT%H:%M:%S') .. '+00:00'
---
- 2024-07-17T19:11:29+00:00
...
```

> Also, the : delimiter is missed.

Wikipedia says that both variants are ok [^1]:

> The [UTC offset](https://en.wikipedia.org/wiki/UTC_offset) is appended directly to the time instead of "Z" suffix above; other [nautical time](https://en.wikipedia.org/wiki/Nautical_time) zone letters are not used. The offset is applied to [UTC](https://en.wikipedia.org/wiki/UTC) to get the [civil time](https://en.wikipedia.org/wiki/Civil_time) in the designated [time zone](https://en.wikipedia.org/wiki/Time_zone) in the format '±[hh]:[mm]', '±[hh][mm]', or '±[hh]'.

Also see "Date and time — Representations for information interchange Part 1: Basic rules" (BS ISO 8601-1:2019) p. 28:

```
4.3.13 Timeshift
A timeshift, often used in the representation of local standard time against UTC, is represented as follows:

a) Basic, hours and minutes: [±][hour][min] or [“Z”]
EXAMPLE 1 ‘+0500’ or ‘Z’
b) Basic, hoursonly: [±][hour] or [“Z”]
EXAMPLE 2 ‘+05’ or ‘Z’
c) Extended, hours and minutes: [±][hour][“:’][min] or [“Z”]
EXAMPLE 3 ‘+05:00’ or ‘Z’

The UTC designator ["Z"] indicates that there is no time shift from UTC of day and is functionally equivalent to the expressions ‘+0000’ and ‘+00:00’. The time shift shall be expressed as positive(i.e. with the leading plus sign [“+”]) if it is ahead of or equal to UTC, and as negative(i.e. with the leading minus sign[“-”]) if it is behind UTC.
```

[^1]: https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

mkokryashkin (commented):
## perf/lua: protect require of column module:
- Not enough context, would be nice to see the reasoning for the patch. The module under discussion is present in the file tree, so it's not clear why this require may fail. Is that a preparation for backporting to older Tarantools?

## ci: enable BENCH_CMD / perf: introduce BENCH_CMD environment variable:
- It seems like it is more logical to join these patches together, as one
is incomplete without the other. Correct me if I am wrong.

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/sh
+
+# The script setting up a Linux operating system before running
+# Tarantool benchmarks, see [1].
+#
+# 1. https://github.com/tarantool/tarantool/wiki/Benchmarking
```
mkokryashkin, line 6: You are not setting up a Linux operating system, you are setting up its kernel parameters for testing.

ligurio:
>> perf/lua: protect require of column module:

> Not enough context, would be nice to see the reasoning for the patch. The module under discussion is present in the file tree, so it's not clear why this require may fail. Is that a preparation for backporting to older Tarantools?

test depends on Lua C module and test will fail with error that module is not found if that module is not built.
The patch makes this error mode user-friendly.

before patch:

```
LuajitError: column_scan.lua:56: module 'column_scan_module' not found:
	no field package.preload['column_scan_module']
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/override/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/sources/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/lib/tarantool/override/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/share/lua/5.1/override/column_scan_module.lua'
	no file '/home/sergeyb/.luarocks/share/lua/5.1/override/column_scan_module/init.lua'
	no file '/home/sergeyb/.luarocks/share/lua/override/column_scan_module.lua'
	no file '/home/sergeyb/.luarocks/share/lua/override/column_scan_module/init.lua'
	no file '/usr/local/share/tarantool/override/column_scan_module.lua'
	no file '/usr/local/share/tarantool/override/column_scan_module/init.lua'
	no file '/usr/share/tarantool/override/column_scan_module.lua'
	no file '/usr/share/tarantool/override/column_scan_module/init.lua'
	no file '/usr/local/share/lua/5.1/override/column_scan_module.lua'
	no file '/usr/local/share/lua/5.1/override/column_scan_module/init.lua'
	no file '/usr/share/lua/5.1/override/column_scan_module.lua'
	no file '/usr/share/lua/5.1/override/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/perf/lua/override/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/5.1/override/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/override/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/tarantool/override/column_scan_module.so'
	no file '/usr/local/lib/tarantool/override/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/tarantool/override/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/tarantool/override/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/lua/5.1/override/column_scan_module.so'
	no file '/usr/local/lib/lua/5.1/override/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/lua/5.1/override/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.1/override/column_scan_module.so'
	no field loaders.builtin['column_scan_module']
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/.rocks/share/tarantool/column_scan_module.lua'
	no file '/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/column_scan_module.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/share/tarantool/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/sources/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/.rocks/lib/tarantool/column_scan_module.so'
	no file '/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/.rocks/lib/tarantool/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/share/lua/5.1/column_scan_module.lua'
	no file '/home/sergeyb/.luarocks/share/lua/5.1/column_scan_module/init.lua'
	no file '/home/sergeyb/.luarocks/share/lua/column_scan_module.lua'
	no file '/home/sergeyb/.luarocks/share/lua/column_scan_module/init.lua'
	no file '/usr/local/share/tarantool/column_scan_module.lua'
	no file '/usr/local/share/tarantool/column_scan_module/init.lua'
	no file '/usr/share/tarantool/column_scan_module.lua'
	no file '/usr/share/tarantool/column_scan_module/init.lua'
	no file '/usr/local/share/lua/5.1/column_scan_module.lua'
	no file '/usr/local/share/lua/5.1/column_scan_module/init.lua'
	no file '/usr/share/lua/5.1/column_scan_module.lua'
	no file '/usr/share/lua/5.1/column_scan_module/init.lua'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/perf/lua/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/5.1/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/tarantool/column_scan_module.so'
	no file '/usr/local/lib/tarantool/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/tarantool/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/tarantool/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/lua/5.1/column_scan_module.so'
	no file '/usr/local/lib/lua/5.1/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/lua/5.1/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.1/column_scan_module.so'
	no file '/home/sergeyb/sources/MRG/tarantool/perf/lua/perf/lua/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/5.1/column_scan_module.so'
	no file '/home/sergeyb/.luarocks/lib/lua/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/tarantool/column_scan_module.so'
	no file '/usr/local/lib/tarantool/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/tarantool/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/tarantool/column_scan_module.so'
	no file '/usr/local/lib/x86_64-linux-gnu/lua/5.1/column_scan_module.so'
	no file '/usr/local/lib/lua/5.1/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/luarocks/lib/lua/5.1/column_scan_module.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.1/column_scan_module.so'
```

after the patch:

```
[0] ~/sources/MRG/tarantool/perf/lua$ ../../build/src/tarantool column_scan.lua 
Lua module "column_scan_module" is not found.
```

perf/tools/setup_env.sh
ligurio: The script is about setting OS environment, but yes, proposed settings are for the kernel only.

perf/tools/setup_env.sh
```diff
@@ -0,0 +1,134 @@
...
+  set_kernel_setting \
+    "Stabilize the frequency of CPU $cpu_id" \
+    "$sysfs_path_cpu" \
+    "performance"
+done
+
+set_kernel_setting \
+  "Workload-specific: drop disk caches" \
+  "/proc/sys/vm/drop_caches" \
+  "1"
```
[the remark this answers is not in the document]
ligurio: Magomed, you are right, it is better to drop cache right before running benchmark. I'll remove this from the script. Currently, used benchmarks are not about IO performance, so we can postpone workload-specific settings once we will need them.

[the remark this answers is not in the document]
ligurio:
```diff
--- a/perf/tools/setup_env.sh
+++ b/perf/tools/setup_env.sh
@@ -127,8 +127,3 @@ for cpu_id in $(seq 0 1 $((ncpu-1))); do
     "$sysfs_path_cpu" \
     "performance"
 done
-
-set_kernel_setting \
-  "Workload-specific: drop disk caches" \
-  "/proc/sys/vm/drop_caches" \
-  "1"
```

ligurio:
>> ci: enable BENCH_CMD / perf: introduce BENCH_CMD environment variable:

>    It seems like it is more logical to join these patches together, as one
>    is incomplete without the other. Correct me if I am wrong.

I don't think so. The patches are independent, `BENCH_CMD` can be used locally after the first commit.
I would keep these patches in a separate commits.

mkokryashkin (approved):

ylobankov (approved):

Issue #10143, opened 2024-06-18
UBsan: runtime error: member access within null pointer of type 'typeof (*trigger)' (aka 'struct trigger')


ligurio:
**Bug description**

* OS: Linux
* Architecture: amd64

**Steps to reproduce**

```
python /home/runner/work/oss-fuzz/oss-fuzz/infra/helper.py build_fuzzers tarantool --engine libfuzzer --sanitizer undefined --architecture x86_64
python /home/runner/work/oss-fuzz/oss-fuzz/infra/helper.py check_build tarantool --engine libfuzzer --sanitizer undefined --architecture x86_64
```

**Actual behavior**

```c++
Dictionary: 2 entries
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1337
INFO: Loaded 1 modules   (24025 inline 8-bit counters): 24025 [0x55af3f6acca8, 0x55af3f6b2a81), 
INFO: Loaded 1 PC tables (24025 PCs): 24025 [0x55af3f6b2a88,0x55af3f710818), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
/src/tarantool/src/lib/core/trigger.h:121:2: runtime error: member access within null pointer of type 'typeof (*trigger)' (aka 'struct trigger')
    #0 0x55af3f4af3ce in trigger_destroy /src/tarantool/src/lib/core/trigger.h:121:2
    #1 0x55af3f48fc45 in fiber_destroy /src/tarantool/src/lib/core/fiber.c:1643:2
    #2 0x55af3f48f9b7 in cord_destroy /src/tarantool/src/lib/core/fiber.c:1947:2
    #3 0x55af3f49ad16 in fiber_free /src/tarantool/src/lib/core/fiber.c:2262:2
    #4 0x55af3f43c14c in LLVMFuzzerTestOneInput /src/tarantool/test/fuzz/swim_proto_member_fuzzer.c:41:5
    #5 0x55af3f3801d0 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:13
[...]

PR #10222, opened 2024-07-10
ci: disable UBsan in a fuzzing workflow

ligurio:
Docker image has been updated on OSS Fuzz in scope of ticket [1]. UBSan in updated image revealed a member access within null pointer of type `'typeof (*trigger)' (aka 'struct trigger')`, see [2]. The patch disables temporarily UBsan in a fuzzing workflow until the fix of aforemenetioned problem.

1. https://github.com/google/oss-fuzz/pull/12085
2. https://github.com/tarantool/tarantool/issues/10143

Related to #10143

NO_CHANGELOG=ci
NO_DOC=ci
NO_TEST=ci

// CC: @LevKats

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sanitizer: [address, undefined]
+        sanitizer: [address]
 
     steps:
       - name: build fuzzers (${{ matrix.sanitizer }})


ylobankov (approved):

Buristan (commented):
Hi, Sergey!
Thanks for the patch!

May we instead disable the `null` check for UBSan when building Tarantool with `ENABLE_FUZZER`?

ligurio:
Fix for #10143 was merged.

Issue #9849, opened 2024-03-21
`SCHEMA_NEEDS_UPGRADE` error when Tarantool 3+ is loaded from old snapshot


ImeevMA:
The problem occurs when Tarantool is loaded using a configuration with a `credentials` section from a snapshot of an older version.

----

Remarks from @Totktonada:

The problem is mostly fixed by 8f43d0b3c90de8f512ab57c0039a252e79435ddb: the `SCHEMA_NEEDS_UPGRADE` error doesn't appear now, because we don't try to grant/revoke privileges if a schema is not upgraded yet.

However, we should also react on a successful schema upgrade and run the grant/revoke process without waiting for a new configuration.

----

More remarks from @Totktonada:

While granting a privilege to an existing user/role is skipped now (by 8f43d0b3c90de8f512ab57c0039a252e79435ddb), the declarative configuration performs other operations that are considered as DDL and forbidden on a non-upgraded instance (see #7149).

These operations are the following:

* Creating a user or a role.
* Setting or changing a password.

Tarantool `3.1.0` fails to perform these operations and stuck in the `startup_in_progress` status if it is started on a database snapshot with an old schema version.

Let's postpone these operation (as well as granting/revoking privileges) if the instance requires an upgrade (and forbids DDL).

PR #10223, opened 2024-07-10
config: add `lua.memory` option

### LINKED ISSUE #8881: New Lua memory limit option after GC64 enabled
Since enabling of GC64 Lua memory can go very high, leading to unexpected OOM across the whole server, instead of just pure OutOfMemory exception inside single instance. So, we need to implement explisit option to cap the memory with ability to change it.

Proposal: new option in section lua:
```yaml
lua:
    memory: 2147483648
```

Default value should be set to 2Gb (2147483648)

Like memtx.memory it could not be lowered, but it whould be good to have an ability to raise it dynamically.

georgiy-belyanin:
Sometimes we need to limit Lua maximum available memory. Add the config property `lua.memory` to specify the maximum byte amount used for lua allocations.

The limit is implemented with a custom allocation function. It's passed into a Lua state during instantiation. The function itself counts memory used by LuaJIT and prevents allocating more than specified in the config file.

If new memory limit is applied at runtime Tarantool waits for instance reload before applying a new option. 

Closes #8881

### DIFF
--- a/changelogs/unreleased/gh-9849-lua-memory-limit-config.md
+++ b/changelogs/unreleased/gh-9849-lua-memory-limit-config.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Added the `lua.memory` parameter to set up the maximum available RAM for
+  Lua scripts (gh-9849).
--- a/src/box/lua/config/applier/lua.lua
+++ b/src/box/lua/config/applier/lua.lua
@@ -1,0 +1,51 @@
+local alloc = require('internal.alloc')
+local log = require('internal.config.utils.log')
+
+-- After limiting the memory to some value we want to be sure
+-- there is unused memory not to fail right after setting the
+-- new limit.
+--
+-- This value represents the required amount of unused memory
+-- after applying the lua.memory parameter.
+--
+-- The value is equal to 16MB. It has been chosen arbitrarily
+-- as 1/16 of the minimum Lua memory limit.
+local REQUIRED_UNUSED_MEMORY_AFTER_APPLICATION = 16 * 1024 * 1024
+
+local function apply(config)
+    local configdata = config._configdata
+    local memory_limit = configdata:get('lua.memory', {use_default = true})
+    local old_memory_limit = alloc.getlimit()
+    local used_memory = alloc.used()
+
+    -- Nothing to do if the limit is unchanged.
+    if memory_limit == old_memory_limit then
+        return
+    end
+
+    -- Check there is enough unused space after applying
+    -- the new Lua memory limit.
+    --
+    -- Otherwise, the alert is set.
+    if memory_limit <
+       used_memory + REQUIRED_UNUSED_MEMORY_AFTER_APPLICATION then
+        local name = 'lua_memory_limit_too_small'
+        local warning = 'lua.apply: lua.memory will be applied ' ..
+                        'after restarting the instance since the ' ..
+                        'new limit is too close to the currently ' ..
+                        'allocated amount of memory'
+        config._aboard:set({type = 'warn', message = warning}, {key = name})
+        return
+    end
+
+    -- The call throws an error if the user tries to limit the
+    -- memory with values < 256MB.
+    alloc.setlimit(memory_limit)
+
+    log.verbose(('lua.apply: set memory limit to %d'):format(memory_limit))
+end
+
+return {
+    name = 'lua',
+    apply = apply,
+}
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -143,6 +143,7 @@ function methods._initialize(self)
         self:_register_source(require('internal.config.source.file').new())
     end
 
+    self:_register_applier(require('internal.config.applier.lua'))
     self:_register_applier(require('internal.config.applier.compat'))
     self:_register_applier(require('internal.config.applier.mkdir'))
     self:_register_applier(require('internal.config.applier.console'))
@@ -315,10 +316,11 @@ function methods._store(self, iconfig, cconfig, source_info)
     self._configdata = configdata.new(iconfig, cconfig, self._instance_name)
 end
 
--- Invoke compat, mkdir, console and box_cfg appliers at the first
--- phase. Invoke all the other ones at the second phase.
+-- Invoke lua, compat, mkdir, console and box_cfg appliers at the
+-- first phase. Invoke all the other ones at the second phase.
 function methods._apply_on_startup(self, opts)
     local first_phase_appliers = {
+        lua = true,
         compat = true,
         mkdir = true,
         console = true,
--- a/src/lua/alloc.c
+++ b/src/lua/alloc.c
@@ -1,0 +1,153 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include "lua/utils.h"
+
+#include <lauxlib.h> /* luaL_error */
+
+#define LUA_MEMORY_LIMIT_DEFAULT (2ULL * 1024 * 1024 * 1024)
+
+/**
+ * Default allocator function which is wrapped into a new one
+ * with the Lua memory limit checker.
+ */
+static lua_Alloc orig_alloc;
+/**
+ * Memory limit for LuaJIT.
+ */
+static size_t memory_limit = LUA_MEMORY_LIMIT_DEFAULT;
+/**
+ * Amount of memory used by LuaJIT.
+ */
+static size_t used;
+
+/**
+ * Lua custom memory allocation function. It extends the original
+ * one with a memory counter and a limit check.
+ */
+static void *
+alloc_with_limit(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	size_t new_used = used + nsize - osize;
+	if (new_used > memory_limit) {
+		/*
+		 * Returning NULL results in the "not enough
+		 * memory" error.
+		 */
+		return NULL;
+	}
+
+	void *result = orig_alloc(ud, ptr, osize, nsize);
+
+	if (result != NULL || nsize == 0)
+		used = new_used;
+
+	/*
+	 * Result may be NULL, in this case "not enough memory"
+	 * is raised.
+	 */
+	return result;
+}
+
+static int
+set_alloc_limit(size_t new_memory_limit)
+{
+	if (new_memory_limit < used) {
+		diag_set(LuajitError, "Cannot limit the Lua memory with values "
+			 "less than the currently allocated amount");
+		return -1;
+	}
+	memory_limit = new_memory_limit;
+	return 0;
+}
+
+void
+luaT_initalloc(struct lua_State *L)
+{
+	assert(L != NULL);
+
+	used = luaL_getgctotal(L);
+
+	void *orig_ud;
+	orig_alloc = lua_getallocf(L, &orig_ud);
+
+	assert(orig_alloc != NULL);
+
+	lua_setallocf(L, alloc_with_limit, orig_ud);
+}
+
+/**
+ * alloc.getlimit() -- get the allocator memory limit.
+ */
+static int
+lbox_alloc_getlimit(struct lua_State *L)
+{
+	lua_pushinteger(L, memory_limit);
+	return 1;
+}
+
+/**
+ * alloc.setlimit() -- set the allocator memory limit.
+ * Returns old memory limit.
+ */
+static int
+lbox_alloc_setlimit(struct lua_State *L)
+{
+	if (lua_gettop(L) < 1) {
+		diag_set(IllegalParams, "Usage: alloc.setlimit(amount)");
+		return luaT_error(L);
+	}
+
+	ssize_t amount = luaL_checkinteger(L, 1);
+
+	if (amount < 0) {
+		diag_set(IllegalParams, "Invalid memory limit: "
+			 "the value must be >= 0");
+		return luaT_error(L);
+	}
+
+	size_t old_memory_limit = memory_limit;
+
+	if (set_alloc_limit(amount) < 0)
+		return luaT_error(L);
+
+	lua_pushinteger(L, old_memory_limit);
+	return 1;
+}
+
+/**
+ * alloc.used() -- get the amount of the allocated memory.
+ */
+static int
+lbox_alloc_used(struct lua_State *L)
+{
+	lua_pushinteger(L, used);
+	return 1;
+}
+
+/**
+ * alloc.unused() -- get the amount of the unused memory.
+ */
+static int
+lbox_alloc_unused(struct lua_State *L)
+{
+	lua_pushinteger(L, memory_limit - used);
+	return 1;
+}
+
+void
+tarantool_lua_alloc_init(struct lua_State *L)
+{
+	static const struct luaL_Reg alloc_methods[] = {
+		{"setlimit", lbox_alloc_setlimit},
+		{"getlimit", lbox_alloc_getlimit},
+		{"used",     lbox_alloc_used},
+		{"unused",   lbox_alloc_unused},
+		{NULL, NULL}
+	};
+
+	luaT_newmodule(L, "internal.alloc", alloc_methods);
+	lua_pop(L, 1);
+}
--- a/src/lua/alloc.h
+++ b/src/lua/alloc.h
@@ -1,0 +1,29 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
+ */
+
+#pragma once
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+#include <stddef.h>
+
+#include <lua.h>
+
+/**
+ * Initializes custom allocator for the LuaJIT VM.
+ * The allocator supports memory limitation.
+ */
+void
+luaT_initalloc(struct lua_State *L);
+
+void
+tarantool_lua_alloc_init(struct lua_State *L);
+
+#if defined(__cplusplus)
+} /* extern "C" */
+#endif /* defined(__cplusplus) */
--- a/src/lua/init.c
+++ b/src/lua/init.c
@@ -49,6 +49,7 @@
 #include "coio.h"
 #include "core/backtrace.h"
 #include "core/tt_static.h"
+#include "lua/alloc.h"
 #include "lua/backtrace.h"
 #include "lua/fiber.h"
 #include "lua/fiber_cond.h"
@@ -733,10 +734,7 @@ void
 tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 		   char **argv)
 {
-	lua_State *L = luaL_newstate();
-	if (L == NULL) {
-		panic("failed to initialize Lua");
-	}
+	lua_State *L = luaT_newstate();
 	luaL_openlibs(L);
 #ifndef LUAJIT_JIT_STATUS
 	(void)luaJIT_setmode(L, 0, LUAJIT_MODE_ENGINE | LUAJIT_MODE_OFF);
@@ -784,6 +782,7 @@ tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 	lua_call(L, 0, 0);
 	lua_register(L, "tonumber64", lbox_tonumber64);
 
+	tarantool_lua_alloc_init(L);
 	tarantool_lua_tweaks_init(L);
 	tarantool_lua_uri_init(L);
 	tarantool_lua_utf8_init(L);
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -42,6 +42,7 @@
 #include "core/say.h"
 #include "core/tt_uuid.h"
 #include "lj_trace.h"
+#include "lua/alloc.h"
 #include "lua/serializer.h"
 #include "trivia/util.h"
 #include "vclock/vclock.h"
@@ -58,6 +59,18 @@ uint32_t CTID_VARBINARY;
 uint32_t CTID_UUID;
 uint32_t CTID_DATETIME = 0;
 uint32_t CTID_INTERVAL = 0;
+
+struct lua_State *
+luaT_newstate(void)
+{
+	struct lua_State *L = luaL_newstate();
+	if (L == NULL)
+		panic("failed to initialize Lua");
+
+	luaT_initalloc(L);
+
+	return L;
+}
 
 /** A copy of index2adr() from luajit/src/lj_api.c. */
 static TValue *
--- a/src/lua/utils.h
+++ b/src/lua/utils.h
@@ -82,6 +82,14 @@ extern uint32_t CTID_DATETIME;
 extern uint32_t CTID_INTERVAL;
 
 /**
+ * Creates a new Lua state with a custom allocator function.
+ * Guarantees the state is not NULL. It panics if lua_State can't
+ * be allocated.
+ */
+struct lua_State *
+luaT_newstate(void);
+
+/**
  * Pushes a new varbinary object with the given content to the Lua stack.
  */
 void
--- a/test/app-luatest/alloc_test.lua
+++ b/test/app-luatest/alloc_test.lua
@@ -1,0 +1,55 @@
+local alloc = require('internal.alloc')
+
+local t = require('luatest')
+local g = t.group()
+
+g.test_invalid_arg = function()
+    local errmsg
+    local function test(...)
+        t.assert_error_msg_equals(errmsg, alloc.setlimit, ...)
+    end
+
+    errmsg = 'Usage: alloc.setlimit(amount)'
+    test()
+
+    errmsg = 'bad argument #1 to \'?\' (number expected, got ' ..
+             'string)'
+    test('foo')
+
+    errmsg = 'Invalid memory limit: the value must be >= 0'
+    test(-1024)
+
+    errmsg = 'Cannot limit the Lua memory with values less ' ..
+             'than the currently allocated amount'
+    test(16)
+end
+
+g.test_basic = function()
+    local function test(amount)
+        local old_memory_limit = alloc.getlimit()
+        t.assert_equals(alloc.setlimit(amount), old_memory_limit)
+        t.assert_equals(alloc.getlimit(), amount)
+    end
+
+    test(256 * 1024 * 1024)
+    test(1024 * 1024 * 1024)
+    test(2 * 1024 * 1024 * 1024)
+    test(4 * 1024 * 1024 * 1024)
+    test(8 * 1024 * 1024 * 1024)
+end
+
+g.test_used_and_unused = function()
+    local function test(amount)
+        alloc.setlimit(amount)
+        t.assert_equals(collectgarbage('count'), alloc.used() / 1024)
+        t.assert_equals(amount / 1024 - collectgarbage('count'),
+                        alloc.unused() / 1024)
+        t.assert_equals(alloc.used() + alloc.unused(), alloc.getlimit())
+    end
+
+    test(256 * 1024 * 1024)
+    test(1024 * 1024 * 1024)
+    test(2 * 1024 * 1024 * 1024)
+    test(4 * 1024 * 1024 * 1024)
+    test(8 * 1024 * 1024 * 1024)
+end
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -187,6 +187,7 @@ set (server_sources
      ssl_cert_paths_discover.c
      systemd.c
      version.c
+     lua/alloc.c
      lua/digest.c
      lua/init.c
      lua/fiber.c
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -51,6 +51,7 @@ lua_source(lua_sources lua/config/applier/compat.lua       config_applier_compat
 lua_source(lua_sources lua/config/applier/console.lua      config_applier_console_lua)
 lua_source(lua_sources lua/config/applier/credentials.lua  config_applier_credentials_lua)
 lua_source(lua_sources lua/config/applier/fiber.lua        config_applier_fiber_lua)
+lua_source(lua_sources lua/config/applier/lua.lua          config_applier_lua_lua)
 lua_source(lua_sources lua/config/applier/mkdir.lua        config_applier_mkdir_lua)
 lua_source(lua_sources lua/config/applier/roles.lua        config_applier_roles_lua)
 lua_source(lua_sources lua/config/applier/sharding.lua     config_applier_sharding_lua)
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -742,6 +742,21 @@ return schema.new('instance_config', schema.record({
             default = 'var/run/{{ instance_name }}/tarantool.pid',
         }),
     }),
+    lua = schema.record({
+        -- Maximum allowed memory allocated by Lua.
+        -- The value can't be less than 256MB and can't be
+        -- changed without restarting the Tarantool instance.
+        memory = schema.scalar({
+            type = 'integer',
+            -- Default value: 2GB.
+            default = 2 * 1024 * 1024 * 1024,
+            validate = function(data, w)
+                if data < 256 * 1024 * 1024 then
+                    w.error('Memory limit should be >= 256MB')
+                end
+            end,
+        }),
+    }),
     console = schema.record({
         enabled = schema.scalar({
             type = 'boolean',
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -156,6 +156,7 @@ extern char session_lua[],
 	config_applier_console_lua[],
 	config_applier_credentials_lua[],
 	config_applier_fiber_lua[],
+	config_applier_lua_lua[],
 	config_applier_mkdir_lua[],
 	config_applier_roles_lua[],
 	config_applier_sharding_lua[],
@@ -438,6 +439,10 @@ static const char *lua_sources[] = {
 	"config/applier/fiber",
 	"internal.config.applier.fiber",
 	config_applier_fiber_lua,
+
+	"config/applier/lua",
+	"internal.config.applier.lua",
+	config_applier_lua_lua,
 
 	"config/applier/mkdir",
 	"internal.config.applier.mkdir",
--- a/test/config-luatest/cluster_config_schema_test.lua
+++ b/test/config-luatest/cluster_config_schema_test.lua
@@ -69,6 +69,7 @@ end
 local instance_config_fields = {
     'config',
     'process',
+    'lua',
     'console',
     'fiber',
     'log',
@@ -328,6 +329,7 @@ g.test_defaults = function()
             enabled = true,
             socket = 'var/run/{{ instance_name }}/tarantool.control',
         },
+        lua = {memory = 2147483648},
         memtx = {
             memory = 268435456,
             allocator = 'small',
--- a/test/config-luatest/lua_memory_test.lua
+++ b/test/config-luatest/lua_memory_test.lua
@@ -1,0 +1,204 @@
+local t = require('luatest')
+local cbuilder = require('luatest.cbuilder')
+local cluster = require('test.config-luatest.cluster')
+
+local g = t.group()
+
+g.before_all(cluster.init)
+g.after_each(cluster.drop)
+g.after_all(cluster.clean)
+
+g.test_basic = function(g)
+    -- Configuration limiting memory available to Lua with 256MB.
+    local lua_memory = 256 * 1024 * 1024
+    local config = cbuilder:new()
+        :set_global_option('lua.memory', lua_memory)
+        :add_instance('i-001', {})
+        :config()
+
+    local cluster = cluster.new(g, config)
+    cluster:start()
+
+    -- Check lua.memory has been successfully applied.
+    cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), lua_memory)
+        t.assert_equals(alloc.getlimit(), lua_memory)
+    end, {lua_memory})
+
+    -- Check there is a "not enough memory" error when Lua
+    -- tries to allocate more than configured.
+    cluster['i-001']:exec(function()
+        local alloc = require('internal.alloc')
+        local digest = require('digest')
+
+        local function overflow()
+            -- Try to allocate all of the Lua available memory
+            -- plus 1MB.
+            local size = alloc.unused() + 1024 * 1024
+            local _ = digest.urandom(size)
+        end
+
+        t.assert_error_msg_equals('not enough memory', overflow, {})
+    end)
+
+    -- Check it's still possible to allocate almost all of the
+    -- memory up to the limit.
+    cluster['i-001']:exec(function()
+        local alloc = require('internal.alloc')
+        local digest = require('digest')
+
+        -- Try to allocate all of the Lua available memory
+        -- minus 1MB.
+        local size = alloc.unused() - 1 * 1024 * 1024
+        local _ = digest.urandom(size)
+
+        -- Check there is <= 2MB left. 2MB is used instead
+        -- of 1MB to ensure the implementation details of Lua
+        -- memory management won't affect the comparison.
+        t.assert_le(alloc.unused(), 2 * 1024 * 1024)
+    end)
+end
+
+g.test_update = function(g)
+    -- Config with 512MB memory limit.
+    local lua_memory_0 = 512 * 1024 * 1024
+    local config_0 = cbuilder:new()
+        :set_global_option('lua.memory', lua_memory_0)
+        :add_instance('i-001', {})
+        :config()
+
+    local cluster = cluster.new(g, config_0)
+    cluster:start()
+
+    -- Config with 256MB memory limit.
+    local lua_memory_1 = 256 * 1024 * 1024
+    local config_1 = cbuilder:new(config_0)
+        :set_global_option('lua.memory', lua_memory_1)
+        :add_instance('i-001', {})
+        :config()
+
+    -- Load a config with a new lua.memory value.
+    cluster:reload(config_1)
+
+    -- lua.memory could be decreased dynamically if
+    -- the limit is not too close to the currently used
+    -- amount of memory.
+    -- Check there is no alerts and it's performed successfully.
+    cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), lua_memory)
+        t.assert_equals(alloc.getlimit(), lua_memory)
+
+        local alerts = config:info().alerts
+        t.assert_equals(alerts, {})
+    end, {lua_memory_1})
+
+    -- Config with 1GB Lua memory limit.
+    local lua_memory_2 = 1024 * 1024 * 1024
+    local config_2 = cbuilder:new(config_1)
+        :set_global_option('lua.memory', lua_memory_2)
+        :config()
+
+    -- Load a config with a new lua.memory value.
+    cluster:reload(config_2)
+
+    -- lua.memory could be increased dynamically.
+    -- Check there is no alerts and it's performed successfully.
+    cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), lua_memory)
+        t.assert_equals(alloc.getlimit(), lua_memory)
+
+        local alerts = config:info().alerts
+        t.assert_equals(alerts, {})
+    end, {lua_memory_2})
+
+    -- Allocate a huge chunk of memory and fetch the used
+    -- value (1/4 of the current memory limit).
+    local used = cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local digest = require('digest')
+
+        rawset(_G, 'stub', digest.urandom(lua_memory / 4))
+        return alloc.used()
+    end, {lua_memory_2})
+
+    local lua_memory_3 = used + 1024 * 1024
+    -- Config with currently used + 1MB memory limit.
+    local config_3 = cbuilder:new(config_2)
+        :set_global_option('lua.memory', lua_memory_3)
+        :config()
+    cluster:reload(config_3)
+
+    -- The lua.memory option can't be applied if after applying
+    -- there won't be enough unused space (currently it's 16MB).
+    --
+    -- Check applying doesn't change anything and creates an
+    -- alert on restart is needed to apply changes.
+    cluster['i-001']:exec(function(new_lua_memory, old_lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), new_lua_memory)
+        t.assert_equals(alloc.getlimit(), old_lua_memory)
+
+        local alerts = config:info().alerts
+        t.assert_equals(#alerts, 1)
+        t.assert_equals(alerts[1].type, 'warn')
+        local exp = 'lua.apply: lua.memory will be applied after ' ..
+                    'restarting the instance since the new limit is too ' ..
+                    'close to the currently allocated amount of memory'
+        t.assert_equals(alerts[1].message, exp)
+    end, {lua_memory_3, lua_memory_2})
+
+    -- Restart the cluster. Check the option has been
+    -- applied and there is no alerts.
+    cluster:stop()
+    cluster:start()
+
+    cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), lua_memory)
+        t.assert_equals(alloc.getlimit(), lua_memory)
+
+        local alerts = config:info().alerts
+        t.assert_equals(alerts, {})
+    end, {lua_memory_3})
+
+    cluster:reload(config_3)
+
+    -- Make sure loading configuration with the same lua.memory
+    -- value don't issue an alert.
+    cluster['i-001']:exec(function(lua_memory)
+        local alloc = require('internal.alloc')
+        local config = require('config')
+
+        t.assert_equals(config:get('lua.memory'), lua_memory)
+        t.assert_equals(alloc.getlimit(), lua_memory)
+
+        local alerts = config:info().alerts
+        t.assert_equals(alerts, {})
+    end, {lua_memory_3})
+
+end
+
+g.test_wrong_value = function(g)
+    local config = cbuilder:new()
+        :add_instance('i-001', {})
+        :config()
+
+    -- cbuilder automatically validates config.
+    -- Rewrite lua.memory option externally.
+    config.lua = {memory = 1024}
+
+    cluster.startup_error(g, config, 'Memory limit should be >= 256MB')
+end


Buristan (changes requested):
Hi, Georgiy!
Thanks for the patch!
Please, consider my comments below.

> Sometimes we need to limit lua maximum available memory. Add config

Typo: s/lua/Lua/
Typo: s/config/the config/

> property lua.memory to specify maximum byte amount used for lua

Typo: s/maximum/the maximum/
Typo: s/lua/Lua/

> into a lua state during instantiation. The function itself counts used RAM

Typo: s/lua/Lua/

Not RAM (since we don't count all memory), more likely "memory used by LuaJIT VM".

Width of the commit message line is more than 72 symbols ([see this section](https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message)).

Also, I am against such approach, see my comments below.

> and prevent allocating allocate more space than specified in the config

Typo: s/prevent/prevents/

Typo: s/allocate//

> Closes #10164

Typo: s/#10164/#8881/

src/lua/alloc.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include <stddef.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+/**
+ * Set maximum available memory for the lua allocator.
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied

test/box-luatest/gh_7389_memtx_memory_test.lua
```diff
@@ -1,14 +1,14 @@
...
-local g = t.group('memtx_memory setting test')
-
-g.before_all(function(g)
-    g.server = server:new{
-        alias = 'default',
-        box_cfg = {memtx_memory = 0}
-    }
-    g.server:start()
-end)
+ local t = require('luatest')
```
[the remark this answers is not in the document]
georgiy-belyanin: You're right. Fixed

src/main.cc
```diff
@@ -1018,6 +1024,7 @@ main(int argc, char **argv)
 	/*
 	 * Pass either a configuration file or a script file to
 	 * configure Lua loaders paths.
+
```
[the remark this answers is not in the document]
georgiy-belyanin: Done.

src/lua/alloc.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+int
+lua_alloc_set_memory(size_t memory);
```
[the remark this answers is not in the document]
georgiy-belyanin: Changed, moved the functions into `lua/utils.h`

src/lua/alloc.cc
```diff
@@ -0,0 +1,89 @@
...
+		      "allocate %zu bytes with a limit of %zu bytes", nsize,
+		      lua_memory);
+		return NULL;
+	}
+
+	if (nsize == 0) {
+		free(ptr);
+		return NULL;
+	} else {
+		void *result = xrealloc(ptr, nsize);
```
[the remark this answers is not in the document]
georgiy-belyanin: Changed everything to the described approach.

test/config-luatest/config_test.lua
```diff
@@ -236,6 +236,32 @@ g.test_config_general = function()
...
+    local config_file = treegen.write_script(dir, 'config.yaml', config)
+    local opts = {
+        config_file = config_file,
+        alias = 'instance-001',
+        chdir = dir,
+    }
+    g.server = server:new(opts)
+    g.server:start()
+    g.server:exec(function()
+        t.assert_equals(box.cfg.lua_memory, 1000000000)
```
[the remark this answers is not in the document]
georgiy-belyanin: Done.

src/box/lua/cfg.cc
```diff
@@ -298,6 +299,17 @@ lbox_cfg_set_worker_pool_threads(struct lua_State *L)
 	return 0;
 }
 
+static int
+lbox_cfg_set_lua_memory(struct lua_State *L)
+{
+	try {
+		lua_alloc_set_memory(cfg_geti("lua_memory"));
```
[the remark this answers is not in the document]
georgiy-belyanin: Done.

src/lua/alloc.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "lib/core/diag.h"
+#include "lib/core/exception.h"
+#include "lib/core/say.h"
+
+#include <stdlib.h>
+
+#define LUA_DEFAULT_MEMORY (2ll * 1024 * 1024 * 1024)
+#define LUA_MIN_MEMORY (256ll * 1024 * 1024)
+
+static size_t lua_memory = LUA_DEFAULT_MEMORY;
```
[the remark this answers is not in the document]
georgiy-belyanin: I think it's a lot clearer. Applied.

src/lua/alloc.cc
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void*
+lua_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	(void)ud;
+
+	lua_allocated_memory += nsize - osize;
+	if (lua_allocated_memory > lua_memory) {
+		panic("lua allocator reached memory limit when trying to "
```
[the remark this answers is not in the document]
georgiy-belyanin: Seems like now everything is ok.

src/lua/alloc.cc
```diff
@@ -0,0 +1,89 @@
...
+		return NULL;
+	}
+
+	if (nsize == 0) {
+		free(ptr);
+		return NULL;
+	} else {
+		void *result = xrealloc(ptr, nsize);
+		if (result == NULL) {
+			panic("lua allocator ran out of memory when trying to allocate "
```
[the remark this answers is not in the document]
georgiy-belyanin: Changed the implementation. It's no longer relatable.

test/box-luatest/lua_memory.lua
```diff
@@ -0,0 +1,30 @@
...
+
+        -- Changing available RAM for lua to 256 MB is ok.
+        box.cfg({lua_memory = 256 * 1024 * 1024})
+        t.assert_equals(box.cfg.lua_memory, 256 * 1024 * 1024)
+
+        -- We shouldn't be able to allocate less than 256MB.
+        t.assert_error_msg_equals(err, box.cfg,
+                                  {lua_memory = 200 * 1024 * 1024})
+    end)
+end
```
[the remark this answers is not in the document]
georgiy-belyanin: Added corresponding test.

src/lua/init.c
```diff
@@ -724,7 +726,7 @@ void
 tarantool_lua_init(const char *tarantool_bin, const char *script, int argc,
 		   char **argv)
 {
-	lua_State *L = luaL_newstate();
+	lua_State *L = lua_newstate(lua_alloc, NULL);
```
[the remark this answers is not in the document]
georgiy-belyanin: Now `tarantool_lua_init()` uses only `luaL_newstate()` and substitutes allocator using `lua_setallocf`

src/lua/init.c
```diff
@@ -49,6 +49,8 @@
 #include "coio.h"
 #include "core/backtrace.h"
 #include "core/tt_static.h"
+#include "cfg.h"
```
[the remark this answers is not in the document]
georgiy-belyanin: You're right. It's redundant.

test/box-luatest/lua_memory.lua
```diff
@@ -0,0 +1,30 @@
...
+    g.server:start()
+end)
+
+g.after_all(function(g)
+    g.server:drop()
+end)
+
+g.test_lua_memory = function(g)
+    g.server:exec(function()
+        local err = "cannot limit lua memory to less than 256MB: No such "..
```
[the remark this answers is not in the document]
georgiy-belyanin: Now single quotes should be used everywhere in the added tests.

test/box-luatest/lua_memory.lua
```diff
@@ -0,0 +1,30 @@
...
+                    "file or directory"
+
+        -- Changing available RAM for lua to 256 MB is ok.
+        box.cfg({lua_memory = 256 * 1024 * 1024})
+        t.assert_equals(box.cfg.lua_memory, 256 * 1024 * 1024)
+
+        -- We shouldn't be able to allocate less than 256MB.
+        t.assert_error_msg_equals(err, box.cfg,
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied this approach.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,42 @@
...
+    g.server:start()
+end)
+
+g.after_all(function(g)
+    g.server:drop()
+end)
+
+g.test_lua_memory_setting = function(g)
+    g.server:exec(function()
+        local MB = 1024 * 1024
```
Buristan, line 22: It is already declared as the upvalue above. This line can be dropped.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,42 @@
...
+        local LUA_MIN_MEMORY = 256 * MB
```
Buristan, line 23: Ditto.

test/config-luatest/memory_test.lua
```diff
@@ -0,0 +1,29 @@
+local yaml = require('yaml')
+local t = require('luatest')
+local treegen = require('test.treegen')
+local helpers = require('test.config-luatest.helpers')
+local justrun = require('test.justrun')
+
+local g = helpers.group()
+
+g.test_relative_config_path = function(g)
+    local dir = treegen.prepare_directory(g, {}, {})
```
Buristan, line 10: I suppose that running an additional instance for this test is overkill. We might just use `t.assert_error_msg_equals()` like you do for `test_lua_memory_setting()`.

Also, it may be done as a subtest in the <lua_memory_test.lua>, but I don't insist on it.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,42 @@
...
+        t.assert_equals(box.cfg.lua_memory, LUA_MIN_MEMORY + 1)
+
+        local err_decreasing = 'cannot decrease lua allocator available '..
+            'memory at runtime'
+        -- Decreasing shouldn't be allowed.
+        t.assert_error_msg_equals(err_decreasing, box.cfg,
+            {lua_memory = LUA_MIN_MEMORY})
+
+        local err_min = 'cannot limit lua memory with values less than 256MB'
+        -- Setting the limit to values less than should be disabled.
```
Buristan, line 38: Nit: Comment length is more than 66 symbols.

Buristan (changes requested):
Hi, Georgiy!
Thanks for the fixes!
Please don't forget to update the commit message according to the new changes and [the following comment](https://github.com/tarantool/tarantool/pull/10223#pullrequestreview-2182011371).

Also, please consider the bunch of nits and minor comments below.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: It won't work because `g.server:exec` seems not to be able to capture the outer variable.

src/lua/utils.c
```diff
@@ -59,6 +68,57 @@ uint32_t CTID_UUID;
...
+
+int
+luaT_setalloclimit(size_t memory)
+{
+	if (memory < LUA_MEMORY_LIMIT_MIN) {
+		diag_set(LuajitError, "cannot limit lua memory with values "
+			 "less than 256MB");
+		return -1;
+	}
+	if (memory < alloc.limit) {
```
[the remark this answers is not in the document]
georgiy-belyanin: It used to be the first approach implemented. But the issue says we shouldn't allow decreasing the memory limit. The same way as the `memtx_memory` config option works.

I'll apply the suggestion. Please, tell me if there is a reason for `memtx_memory` to be implemented in a different way

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Same as above. Probably I do something wrong but capturing the global variable:
```
test/box-luatest/lua_memory_test.lua:21: bad argument #2 to exec (excess upvalues: LUA_MIN_MEMORY) 
```

src/main.cc
```diff
@@ -444,6 +444,8 @@ load_cfg(void)
 		}
 	}
 
+	box_set_lua_memory();
```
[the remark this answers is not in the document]
georgiy-belyanin: `src/lua` subsystem doesn't use `src/box/` at all. It looks like breaking architectural dependency directions.
I'll bring it to `box_lua_init()`. Is it ok?

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Made it shorter.

src/lua/utils.h
```diff
@@ -81,6 +81,19 @@ extern uint32_t CTID_DATETIME;
 /** Type ID of struct interval. */
 extern uint32_t CTID_INTERVAL;
 
+/**
+ * Creates a new Lua state with a custom allocator function.
+ * Guarantees the state is not null. May panic if something has failed.
+ */
+struct lua_State*
```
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

src/box/box.cc
```diff
@@ -1972,6 +1972,17 @@ box_check_config(void)
 	box_check_memtx_sort_threads();
 }
 
+void
+box_set_lua_memory(void)
+{
+	int lua_memory = cfg_geti("lua_memory");
+	if (!lua_memory)
```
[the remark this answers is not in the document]
georgiy-belyanin: The situation is impossible. Removed the check.

src/lua/utils.c
```diff
@@ -59,6 +68,57 @@ uint32_t CTID_UUID;
...
+struct lua_State*
+luaT_newstate(void)
+{
+	struct lua_State *L = luaL_newstate();
+	if (L == NULL) {
+		panic("failed to initialize Lua");
+	}
+
+	void *orig_ud;
+	alloc.orig_alloc = lua_getallocf(L, &orig_ud);
```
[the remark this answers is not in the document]
georgiy-belyanin: Fixed it by initialising `alloc.allocated` with an allocated memory amount when creating a new state.

src/lua/utils.c
```diff
@@ -59,6 +68,57 @@ uint32_t CTID_UUID;
...
+		diag_set(OutOfMemory, nsize, "luaT_alloc", "lua objects");
+		diag_log();
+		return NULL;
+	}
+
+	void *result = alloc.orig_alloc(ud, ptr, osize, nsize);
+	return result;
+}
+
+struct lua_State*
```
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

src/lua/utils.c
```diff
@@ -59,6 +68,57 @@ uint32_t CTID_UUID;
 uint32_t CTID_DATETIME = 0;
 uint32_t CTID_INTERVAL = 0;
 
+/**
+ * Lua custom memory allocation function. It extends the original one with
+ * a memory counter and a limit check.
+ */
+static void*
```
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

src/lua/utils.c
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

src/lua/utils.h
```diff
@@ -81,6 +81,19 @@ extern uint32_t CTID_DATETIME;
 /** Type ID of struct interval. */
 extern uint32_t CTID_INTERVAL;
 
+/**
+ * Creates a new Lua state with a custom allocator function.
+ * Guarantees the state is not null. May panic if something has failed.
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied.

src/lua/utils.h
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Fixed.

test/config-luatest/memory_test.lua
georgiy-belyanin: When error occurs the Tarantool panics, as I found. `t.assert_error_msg_equals()` aren't able to handle it.

Probably it shouldn't work like this. I couldn't yet find a way to throw an error in allocator (either long jumps nor exceptions nor returning NULL from it works properly).

Buristan: > I couldn't yet find a way to throw an error in allocator

I suppose, in the case of a limit violation, the allocator function should just return `NULL`. In that case, the LuaJIT VM will raise the OOM error, as expected.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+	struct lua_State *L = luaL_newstate();
+	if (L == NULL) {
+		panic("failed to initialize Lua");
+	}
+
+	/*
+	 * lua_gc(L, LUA_GC_COUNT) returns the allocated memory in KBs
+	 * lua_gc(L, LUA_GC_COUNTB) returns the reminder
+	 */
+	alloc.allocated = lua_gc(L, LUA_GCCOUNT, 0) * 1024ll +
```
Buristan, line 102: Why do we need `ll` suffix here?

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,81 @@
...
+        local err_min = 'cannot limit lua memory with values less than 256MB'
+        -- Setting the limit less than 256MB is prohibited.
+        t.assert_error_msg_equals(err_min, box.cfg,
+            {lua_memory = LUA_MIN_MEMORY - 1})
+
+        box.cfg{lua_memory = LUA_EXTENDED_MEMORY}
+
+        local used = collectgarbage('count') * 1024
+        local free = LUA_EXTENDED_MEMORY - used
+        local a = string.rep('a', free / 3)
```
Buristan, line 47: It is better to use `_` for unused variable in Lua. See [this CI failure](https://github.com/tarantool/tarantool/actions/runs/9994276601/job/27623665785?pr=10223#step:6:999)

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,81 @@
...
+
+        --local err_decreasing = 'cannot decrease lua allocator available '..
```
Buristan, line 49: Looks like the remaining part of debugging.

src/lua/utils.h
```diff
@@ -81,6 +81,19 @@ extern uint32_t CTID_DATETIME;
 /** Type ID of struct interval. */
 extern uint32_t CTID_INTERVAL;
 
+/**
+ * Creates a new Lua state with a custom allocator function.
+ * Guarantees the state is not NULL. It panics if lua_State can't be allocated.
```
Buristan, line 86: Nit: Comment line length is more than 66 symbols.

Buristan (commented):
Hi, Georgiy!
Thanks for the fixes!
Please, consider some minor nits below.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+struct lua_State *
+luaT_newstate(void)
+{
+	struct lua_State *L = luaL_newstate();
+	if (L == NULL) {
+		panic("failed to initialize Lua");
+	}
+
+	/*
+	 * lua_gc(L, LUA_GC_COUNT) returns the allocated memory in KBs
```
Buristan, line 99: Missed dot at the end of the sentence.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+	 * lua_gc(L, LUA_GC_COUNTB) returns the reminder
```
Buristan, line 100: Missed dot at the end of the sentence.

src/lua/utils.c
georgiy-belyanin: `lua_gc` returns an `int` number. Without `ll` an implicit conversion into a `long long` isn't performed. An overflow is possible.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Applied.

src/lua/utils.h
georgiy-belyanin: Oh, it's related even to the docs. Fixed.

src/lua/utils.c
georgiy-belyanin: Fixed.

src/lua/utils.c
georgiy-belyanin: Fixed.

test/config-luatest/memory_test.lua
georgiy-belyanin: It's implemented like this right now. I wasn't able to capture the error with `t.assert_error_msg_equals()`

changelogs/unreleased/gh-9849-lua-memory-limit-config.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* Add `lua.memory` parameter for setting maximum available RAM for lua scripts.
```
andreyaksenov, line 3:
```suggestion
* Added the `lua.memory` parameter to set up the maximum available RAM for Lua scripts.
```

andreyaksenov (approved):

changelogs/unreleased/gh-9849-lua-memory-limit-config.md
georgiy-belyanin: Applied. Thank you.

src/main.cc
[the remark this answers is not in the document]
Totktonada: Raw thought: a `box.cfg` option was not requested. And the given option has no relation to the database at all. We can expose a Lua function to set the limit from some module in `src/lua`.

`tarantool` module is informational, not a good option. Create some `tarantool.vm` module? Don't know.

Any ideas?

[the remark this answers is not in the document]
Totktonada: There is an idea to place the Lua API to the `misc` module.

```yaml
# tarantool> misc
---
- getmetrics: 'function: builtin#203'
  sysprof:
    stop: 'function: builtin#205'
    report: 'function: builtin#206'
    start: 'function: builtin#204'
  memprof:
    start: 'function: builtin#207'
    stop: 'function: builtin#208'
...
```

It looks as a good place. However, the implementation of this module is in the luajit repository (our fork). If we're going to expose a function from it, it seems that the code should reside in the luajit fork too[^1].

I like the idea to have this machinery in the luajit codebase, because luajit's memory limit seems closer to luajit itself rather than tarantool. @Buristan, what do you think?

[^1]: Technically we can add a function into the module table from tarantool, but I don't like that the implementatoin of the module would spread across different files and even repositories in this case.

src/lua/utils.c
```diff
@@ -46,6 +46,16 @@
 #include "trivia/util.h"
 #include "vclock/vclock.h"
 
+#define LUA_MEMORY_LIMIT_MIN (256ll * 1024 * 1024)
+#define LUA_MEMORY_LIMIT_DEFAULT LUA_MEMORY_LIMIT_MIN
```
Buristan, line 50: It should be 2Gb instead.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+/**
+ * Lua custom memory allocation function. It extends the original one with
+ * a memory counter and a limit check.
+ */
+static void *
+luaT_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	alloc.allocated += nsize - osize;
+	if (alloc.allocated > alloc.limit) {
+		diag_set(OutOfMemory, nsize, "luaT_alloc", "lua objects");
```
Buristan, line 81: Typo: s/lua/Lua/

src/main.cc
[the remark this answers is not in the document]
Buristan: We discussed this with @Totktonada offline and came to the following conclusion:

Decalring this custom allocator function inside the LuaJIT fork is hardly maintainable (changes in the Tarantool allocator require changes in the LuaJIT fork that look inconsistent) and too specific to be a part of the API, so it should be done as an __internal__ module inside Tarantool.

So we should provide an internal module (say <src/lua/alloc.c>) that provides two Lua C functions -- to set the alloc function and restore the previous one. Also, it provides in the Lua namespace function `internal.alloc.setlimit()` which allows to easily set the limit of the Lua memory.

Buristan (changes requested):

test/config-luatest/memory_test.lua
Buristan: This is due to [not restored](https://github.com/tarantool/tarantool/pull/10223#discussion_r1704975669) `allocated` value in the case of the OOM.
After fixing the behaviour we are able to see last error via [`box.error.last()`](https://www.tarantool.io/ru/doc/2.11/reference/reference_lua/box_error/last/).

```sh
src/tarantool -e '
  local function bigalloc()
    -- On your branch before fixing the default value.
    ("x"):rep(1024 * 1024 * 256)
  end
  local r, e = pcall(bigalloc)
  print("err:",  box.error.last())
'
OutOfMemory: Failed to allocate 268435456 bytes in luaT_alloc for lua objects
err:    Failed to allocate 268435456 bytes in luaT_alloc for lua objects

```

src/lua/utils.c
georgiy-belyanin: Fixed.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+	void *orig_ud;
+	alloc.orig_alloc = lua_getallocf(L, &orig_ud);
+	assert(alloc.orig_alloc != NULL);
+	lua_setallocf(L, luaT_alloc, orig_ud);
+
+	return L;
+}
+
+int
+luaT_setalloclimit(size_t memory)
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied. It'll be more consistent.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+	struct lua_State *L = luaL_newstate();
+	if (L == NULL) {
+		panic("failed to initialize Lua");
+	}
+
+	/*
+	 * lua_gc(L, LUA_GC_COUNT) returns the allocated memory in KBs.
+	 * lua_gc(L, LUA_GC_COUNTB) returns the reminder.
+	 */
+	alloc.allocated = lua_gc(L, LUA_GCCOUNT, 0) * 1024ll +
```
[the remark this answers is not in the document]
georgiy-belyanin: Done! Thank you.

src/lua/utils.c
georgiy-belyanin: Now it's so. Thank you.

src/lua/utils.c
```diff
@@ -59,6 +69,64 @@ uint32_t CTID_UUID;
...
+ * a memory counter and a limit check.
+ */
+static void *
+luaT_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	alloc.allocated += nsize - osize;
+	if (alloc.allocated > alloc.limit) {
+		diag_set(OutOfMemory, nsize, "luaT_alloc", "lua objects");
+		diag_log();
+		return NULL;
```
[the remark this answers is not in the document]
georgiy-belyanin: Yep. Quite a stupid mistake. Thank you. Fixed

test/config-luatest/memory_test.lua
georgiy-belyanin: Oh, you're right. Fixed it. Thanks

src/lua/utils.c
```diff
@@ -46,6 +46,16 @@
 #include "trivia/util.h"
 #include "vclock/vclock.h"
 
+#define LUA_MEMORY_LIMIT_MIN (256ll * 1024 * 1024)
+#define LUA_MEMORY_LIMIT_DEFAULT LUA_MEMORY_LIMIT_MIN
+
+static struct {
+	lua_Alloc orig_alloc;
+
```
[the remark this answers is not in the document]
georgiy-belyanin: Now the variables are moved. Think the comment is no longer relatable.

src/main.cc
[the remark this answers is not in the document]
georgiy-belyanin: Applied the suggested approach.

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+void
+tarantool_lua_alloc_init(struct lua_State *L)
+{
+	static const struct luaL_Reg alloc_methods[] = {
+		{"setmemory", lbox_alloc_setmemory},
+		{NULL, NULL}
+	};
+
+	luaT_newmodule(L, "internal.alloc", alloc_methods);
+	lua_setfield(L, -2, "internal.alloc");
```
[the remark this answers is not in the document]
Buristan: Side note: you may debug the Lua C code with a `lj-stack` command which is a part of our [`luajit-gdb.py gdb extension for LuaJIT`](https://github.com/tarantool/luajit/blob/tarantool/master/src/luajit-gdb.py).

src/lua/alloc.h
```diff
@@ -0,0 +1,60 @@
...
+ * SUCH DAMAGE.
+ */
+
+#pragma once
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+#include <stdint.h>
```
Buristan, line 38: This include looks excess.

src/lua/alloc.h
```diff
@@ -0,0 +1,60 @@
...
+#include <lua.h>
+
+/**
+ * Limits the maximum available Lua memory.
+ */
+int
+luaT_setallocmemory(size_t memory);
+
+/**
+ * Initializes custom allocator for the state.
```
Buristan, line 49: Minor: "for the state" isn't very clear. "for the LuaJIT VM" is more logical to me.

src/main.cc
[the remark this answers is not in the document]
Buristan: In that case, why do we need to call `box_set_lua_memory()` here (I thought that it should be called in some other place, shouldn't it)?

test/app-luatest/alloc_test.lua
```diff
@@ -0,0 +1,32 @@
...
+    errmsg = 'Invalid memory amount: the value must be >= 0'
+    test(-1024)
+
+    errmsg = 'Cannot limit the Lua memory with values less than 256MB'
+    test(1024)
+    test(1024 * 1024)
+    test(256 * 1024 * 1024 - 1)
+end
+
+g.test_decode = function()
```
Buristan, line 24: Why is it named `decode`?

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,68 @@
+local server = require('luatest.server')
+local t = require('luatest')
+local g = t.group('lua_memory setting test')
+--local treegen = require('test.treegen')
```
Buristan, line 4: Looks like a part of the previous version.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,68 @@
...
+
+local MB = 1024 * 1024
+local LUA_MIN_MEMORY = 256 * MB
+
+g.before_each(function()
+    g.server = server:new{
```
Buristan, line 10: Please use `()` for function or methods calls. Here and below.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,68 @@
...
+        local LUA_MIN_MEMORY = 256 * MB
+        local LUA_EXTENDED_MEMORY = 600 * MB
+
+        t.assert_equals(box.cfg.lua_memory, LUA_MIN_MEMORY)
+
+        -- Increasing the limit should be ok.
+        box.cfg{lua_memory = LUA_MIN_MEMORY + 1}
+        t.assert_equals(box.cfg.lua_memory, LUA_MIN_MEMORY + 1)
+
+        local err_min = 'Cannot limit the Lua memory with values less than '..
```
Buristan, line 34: It is better to surround operators (`..` in that case) with spaces [1].
Here and below.
[1]: https://www.tarantool.io/en/doc/latest/contributing/lua_style_guide/

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,68 @@
...
+                        '256MB'
+        -- Setting the limit less than 256MB is prohibited.
+        t.assert_error_msg_equals(err_min, box.cfg,
+            {lua_memory = LUA_MIN_MEMORY - 1})
+
+        box.cfg{lua_memory = LUA_EXTENDED_MEMORY}
+
+        local used = collectgarbage('count') * 1024
+        local free = LUA_EXTENDED_MEMORY - used
+        local _ = string.rep('a', free / 3)
```
Buristan, line 44: Please add a comment that this string should be successfully allocated.

test/box-luatest/lua_memory_test.lua
```diff
@@ -0,0 +1,68 @@
...
+        local memory = 256 * 1024 * 1024
+
+        local function overflow()
+            local free = memory - (collectgarbage('count')) * 1024
+            local _ = string.rep('a', free / 2)
+        end
+
+        local err_min = 'not enough memory'
+        t.assert_error_msg_equals(err_min, overflow)
+    end)
```
Buristan, line 67: May be it is worth to check the `box.error.last()` too.

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+luaT_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	size_t new_allocated = allocated + nsize - osize;
+	if (new_allocated > memory) {
+		diag_set(OutOfMemory, nsize, "luaT_alloc", "Lua objects");
+		return NULL;
+	}
+	allocated = new_allocated;
+
+	return orig_alloc(ud, ptr, osize, nsize);
```
Buristan, line 65: I suppose that `orig_alloc` may return `NULL` too. So, we should check the returned pointer and update allocated only if it is not `NULL`.

Buristan (changes requested):
Hi, Georgiy!
Thanks for the fixes!
Please consider my comments below.

> into a Lua state during instantiation. The function itself counts memory

Typo: s/memory/the memory/

src/lua/alloc.h
georgiy-belyanin: It was supposed for including `size_t` type. Replaced with `stddef.h`

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+tarantool_lua_alloc_init(struct lua_State *L)
+{
+	static const struct luaL_Reg alloc_methods[] = {
+		{"setmemory", lbox_alloc_setmemory},
+		{NULL, NULL}
+	};
+
+	luaT_newmodule(L, "internal.alloc", alloc_methods);
+	lua_setfield(L, -2, "internal.alloc");
+	lua_pop(L, 1);
```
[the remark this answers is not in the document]
georgiy-belyanin: Yep, that was the thing I misunderstood when copying the code from the other place. Thanks. Fixed.

src/lua/alloc.c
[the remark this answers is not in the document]
georgiy-belyanin: Fixed. Thank you especially for the `lj-stack` debug solution.

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+ * The maximum allowed LuaJIT memory limit.
+ */
+static size_t memory = LUA_MEMORY_DEFAULT;
+/**
+ * Currently allocated memory amount by LuaJIT.
+ */
+static size_t allocated;
+
+/**
+ * Lua custom memory allocation function. It extends the original one with
```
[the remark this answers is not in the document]
georgiy-belyanin: Fixed.

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+ */
+#include "lua/utils.h"
+
+#include <lauxlib.h> /* luaL_error */
+
+#define LUA_MEMORY_MIN (256ll * 1024 * 1024)
+#define LUA_MEMORY_DEFAULT (2ll * 1024 * 1024 * 1024)
+
+/**
+ * The default allocator function which is wrapped into a new one.
```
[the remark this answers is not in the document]
georgiy-belyanin: Merged your and my approaches. Now it says `into a new one with the Lua memory limit checker`

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+#include "lua/utils.h"
+
+#include <lauxlib.h> /* luaL_error */
+
+#define LUA_MEMORY_MIN (256ll * 1024 * 1024)
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied.

src/lua/alloc.h
georgiy-belyanin: Applied.

test/app-luatest/alloc_test.lua
georgiy-belyanin: Oh, forget to change the copy/pasted code. Thanks!

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: My bad. Thank you. Fixed!

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Done.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: The comment was added.

src/lua/alloc.c
georgiy-belyanin: You're absolutely right, thank you for finding this point.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Done!

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Done.

src/lua/alloc.c
```diff
@@ -0,0 +1,135 @@
...
+	if (amount < 0) {
+		diag_set(IllegalParams, "Invalid memory amount: "
+			 "the value must be >= 0");
+		return luaT_error(L);
+	}
+
+	if (luaT_setallocmemory(amount) < 0)
+		return luaT_error(L);
+
+	lua_pushinteger(L, memory);
```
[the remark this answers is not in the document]
georgiy-belyanin: There is a test `app-luatest/alloc_test` that tests the return value of the function.
Actually, I don't understand the first suggestion. You mean we should push the value returned by the `luaT_setallocmemory`? Seems like everything is working ok rn.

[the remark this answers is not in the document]
Buristan: Oh, my bad!
This was counterintuitive for me. I expect that this function should return the __old__ value of the memory limit (like it is done in the `collectgarbage` API for GC settings).

@Totktonada, what do you think about returing of the old value instead?

test/box-luatest/lua_memory_test.lua
Buristan:
```diff
+        local err1 = 'not enough memory'
+        local err2 = 'Failed to allocate'
+        t.assert_error_msg_equals(err1, overflow)
+        t.assert_str_contains(tostring(box.error.last()), err2)
```

I suppose we can just drop local variables (`err1`, `err2`) here for better readability.

```lua
        t.assert_error_msg_equals('not enough memory', overflow)
        t.assert_str_contains(tostring(box.error.last()), 'Failed to allocate')
```
What do you think?

src/lua/alloc.c
Buristan: I'm not sure, that we should set the `diag` here. That may allow us to separate the error from our allocator and internal ones. @Totktonada, what do you think?

src/main.cc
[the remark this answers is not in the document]
Buristan: What I mean is that it is just an unexpected place for me (I suppose we should end managing with the `pid_file` first). OTOH, it should be done before `box_init_say()`, since it uses Lua memory.
I will ask @Totktonada for the final judgment here.

test/box-luatest/lua_memory_test.lua
Buristan: Nit: Do you mean `1/3`?

georgiy-belyanin: Actually, it allocates ~2/3 of the free memory. Last assignment allocates `free / 3 * 2` memory:
```
tarantool> _ = string.rep('1', 1000 * 1000 * 1000)
freed 6144
allocated 1073733632
allocated 1000000025
```
These log was got by adding printing allocations/frees with size >=4kb.

Some experiments made me think, that there is some kind of a cache for such large calls. The memory is allocated for `string.rep()` call result with the specified parameters and for the variable itself separately. It's easy to see because the later calls `any_other_var = string.rep('1', 1000 * 1000 * 1000)` won't cause any additional large allocations.

test/box-luatest/lua_memory_test.lua
georgiy-belyanin: Yes, you're right. The strings are short enough. Done!

test/box-luatest/lua_memory_test.lua
Buristan: I suppose this is due to the growth of the string buffer. Let's drop the mention of the specific memory amount to avoid confusion.

Side note: Later calls don't cause allocations since LuaJIT uses string interning, so there is no additional string created.

georgiy-belyanin: Dropped the mention of the strange `2/3` value. Thanks for the side note.

src/main.cc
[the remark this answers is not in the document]
Totktonada: I remember a result of our past discussions as follows:

* We create an internal Lua module (`internal.alloc`) with the allocator and functions to enable/disable it and get/set the memory limit.
* Tarantool enables the allocator for the main Lua state early at startup (and the default 2 GiB limit is set).
* We create a configuration option and a configuration applier (`src/box/lua/config/applier/lua.lua` or kinda) that allows to configure the limit (it uses the `internal.alloc` API directly).

This way we don't need to add any `box.cfg` option and it looks meaningful -- this option is not related to box (the database). We have fiber slice limit configuration implemented this way.

BTW, I propose to handle decreasing of the limit in the following way. If the limit is decreased, a warning is issued and the limit remains unchanged (until restart). It is how non-dynamic `box.cfg` options are handled.

https://github.com/tarantool/tarantool/blob/d33b14793bea58520139172599621daaaf66b947/src/box/lua/config/applier/box_cfg.lua#L550-L558

The motivation here is that sometimes it is convenient to setup needed options first and restart instances then.

src/lua/alloc.c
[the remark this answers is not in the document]
Totktonada: Returning the new value seems a bit confusing to me too.

I would return nothing or an old value.

I also think that we need some function to get the currently set limit.

src/main.cc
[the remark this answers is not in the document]
georgiy-belyanin: Thank you for the suggestion! Now I completely agree with you the option is not related to box at all. I am moving it to a new config applier.

src/lua/alloc.c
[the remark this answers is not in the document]
georgiy-belyanin: Thanks, now I am returning the old value and provide `internal.alloc.getlimit` call.

src/box/lua/config/init.lua
```diff
@@ -143,6 +143,7 @@ function methods._initialize(self)
         self:_register_source(require('internal.config.source.file').new())
     end
 
+    self:_register_applier(require('internal.config.applier.lua'))
```
Totktonada, line 146: It doesn't change things much, but I would add it to `first_phase_appliers` (see below) for consistency with all the other pre-box-cfg appliers.

src/box/lua/config/applier/lua.lua
```diff
@@ -0,0 +1,32 @@
...
+        -- memory with values < 256MB or with an amount less than
+        -- currently allocated.
+        alloc.setmemory(memory)
+        log.verbose(('lua.apply: set memory limit to %d'):format(memory))
+
+        is_lua_memory_applied = true
+    else
+        local warning = 'lua.apply: non-dynamic option lua.memory ' ..
+                        'will not be set until the instance is restarted'
+        config._aboard:set({type = 'warn', message = warning})
```
Totktonada, line 25: Technically this code is called once on reconfiguration and at most twice on startup (and the first call can't issue the alert). However, I wouldn't bother much about a code on the upper level and just be ready to several calls per (re)configuration process. IOW, I would add a key for this alert, so several alerts with this message can't be issued by construction.

src/lua/alloc.c
Totktonada: @georgiy-belyanin Any updates here?

src/lua/alloc.c
```diff
@@ -0,0 +1,171 @@
...
+	if (lua_gettop(L) < 1)
+		luaL_error(L, "Usage: alloc.setmemory(amount)");
+
+	long long amount = lua_tointeger(L, 1);
+
+	if (amount < 0) {
+		diag_set(IllegalParams, "Invalid memory amount: "
+			 "the value must be >= 0");
+		return luaT_error(L);
+	}
```
Totktonada, line 127: Seems like we should use either string errors or `IllegalParams` in both cases.

src/lua/alloc.c
```diff
@@ -0,0 +1,171 @@
...
+	lua_setallocf(L, luaT_alloc, orig_ud);
+}
+
+/**
+ * alloc.getmemory() -- get the allocator memory limit.
+ */
+static int
+lbox_alloc_getmemory(struct lua_State *L)
+{
+	lua_pushinteger(L, memory);
```
Totktonada, line 107: Should we use `luaL_pushuint64` or something like it? The `memory` variable may go beyond 2^31-1.

src/lua/alloc.c
```diff
@@ -0,0 +1,171 @@
...
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+#include "lua/utils.h"
+
+#include <lauxlib.h> /* luaL_error */
+
+#define LUA_MEMORY_DEFAULT (2LL * 1024 * 1024 * 1024)
```
Totktonada, line 35: Should be 2ULL (we use `size_t` for these values below, not `ssize_t`).

src/lua/alloc.c
```diff
@@ -0,0 +1,171 @@
...
+ * alloc.setmemory() -- set the allocator memory limit.
+ * Returns old memory limit.
+ */
+static int
+lbox_alloc_setmemory(struct lua_State *L)
+{
+	if (lua_gettop(L) < 1)
+		luaL_error(L, "Usage: alloc.setmemory(amount)");
+
+	long long amount = lua_tointeger(L, 1);
```
Totktonada, line 121: Should it be `size_t` too (unsigned)? Does `lua_tointeger` handles values beyond 2^31-1 correctly?

src/lua/alloc.c
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
```
Totktonada, line 2: Let's use `SPDX-License-Identifier` (grep for examples).

src/lua/init.c
```diff
@@ -49,6 +49,7 @@
 #include "coio.h"
 #include "core/backtrace.h"
 #include "core/tt_static.h"
+#include "lua/alloc.h"
```
Totktonada, line 52: Seems unused here.

Totktonada: Ah, it is `lua/alloc.h`. Sorry.

src/box/box.cc
```diff
@@ -35,6 +35,7 @@
 #include <sys/utsname.h>
 #include <spawn.h>
 
+#include "lua/alloc.h"
```
Totktonada, line 38: Seems unused here.

src/main.cc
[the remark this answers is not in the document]
Totktonada: Seems solved (raised a separate discussiong about the non-dynamism).

Totktonada:
@georgiy-belyanin Glanced briefly. I'm mostly comfortable with the patchset, but I have a few questions I would like to solve before proceed. Mainly the (non)-dynamism.

@Buristan PTAL on the updates if time permits.

src/lua/alloc.c
georgiy-belyanin: You're right. Applied.

src/box/box.cc
georgiy-belyanin: Thank you for noticing. Applied.

src/lua/alloc.c
georgiy-belyanin: You're right. Now using IllegalParams in both of the cases. Thank you.

src/lua/alloc.c
georgiy-belyanin: Done. Thank you for that idea.

src/lua/alloc.c
georgiy-belyanin: Seems like the following methods should be ok, I've stolen the approach from 
```lua
static int
luaT_httpc_io_read(lua_State *L)
{
	struct httpc_io *io = luaT_httpc_checkio(L);
	struct httpc_request *req = io->req;
	uint32_t ctypeid = 0;
	char *buf = *(char **)luaL_checkcdata(L, 2, &ctypeid);
	ssize_t len = luaL_checkinteger(L, 3);
	double timeout = luaL_checknumber(L, 4);

	if (len <= 0) {
		diag_set(IllegalParams, "io: payload length must be >= 0");
		return luaT_error(L);
	}

	if (timeout < 0) {
		diag_set(IllegalParams, "io: timeout must be >= 0");
		return luaT_error(L);
	}

	ssize_t res = httpc_request_io_read(req, buf, len, timeout);
	if (res < 0) {
		return luaT_error(L);
	}

	lua_pushinteger(L, res);
	return 1;
}
```
AFAIU, Lua uses 64-bit integers by default (though it can be compiled in 32-bit mode). I suppose there is no need to take it into account since either Lua is compiled in GC64 mode and the option is works OK or Lua memory limit can't be more than 2^31 - 1.

src/lua/alloc.c
georgiy-belyanin: I've tried to use `size_t` though it works quite strange. So I changed the value type to ssize_t.

You may see quite similar approach is mentioned in [one of the comments above](https://github.com/tarantool/tarantool/pull/10223#discussion_r1853557675).

Full diff of the method:
```diff
@@ -115,10 +90,12 @@ lbox_alloc_getmemory(struct lua_State *L)
 static int
 lbox_alloc_setmemory(struct lua_State *L)
 {
-       if (lua_gettop(L) < 1)
-               luaL_error(L, "Usage: alloc.setmemory(amount)");
+       if (lua_gettop(L) < 1) {
+               diag_set(IllegalParams, "Usage: alloc.setmemory(amount)");
+               return luaT_error(L);
+       }

-       long long amount = lua_tointeger(L, 1);
+       ssize_t amount = luaL_checkinteger(L, 1);
```

src/lua/alloc.c
Totktonada: Lua has no integers, just `number` (which is `double` under the hood).

`lua_pushinteger` accepts `lua_Integer`, which is `LUA_INTEGER` defined as `ptrdiff_t`. Hm. Seems like it should work with integers beyond 2^32-1.

OTOH, it likely has rounding problems for values above 2^53. It is ~8PiB. Hm. Maybe we can pass it over.

I would like to hear @Buristan's opinion.

src/box/lua/config/init.lua
georgiy-belyanin: Done. Thank you.

src/box/lua/config/applier/lua.lua
georgiy-belyanin: I've added the key for the alert. Now it's impossible for the alert to be duplicated. Thank you.

src/box/lua/config/applier/lua.lua
```diff
@@ -0,0 +1,32 @@
...
+    end
```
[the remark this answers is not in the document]
georgiy-belyanin: Oh, I misunderstood you. I though you meant the option shouldn't change during the runtime completely. 

I've changed the code in a such way it allows increasing but prohibits decreasing.

src/lua/alloc.c
georgiy-belyanin: I no longer do `diag_set` there. By default lua handles this as `not enough memory`. I suppose it's quite ok since this way the error is quite clear, no extra error handling logic is provided.

src/lua/alloc.c
Totktonada: > I've added the test on setting 8Gb memory limit

But, please, don't actually allocate so much memory in a test. We're testing on various hardware and sometimes these are quite small machines.

src/lua/alloc.c
Totktonada: I would do it in the header file too.

src/lua/alloc.c
georgiy-belyanin: It just tests it's possible to specify such limit, not actually allocate this much RAM.

Totktonada: OK. I don't bother about values beyond 2^63-1.

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+static int
+lbox_alloc_allocated(struct lua_State *L)
+{
+	lua_pushinteger(L, allocated);
+	return 1;
+}
+
+/**
+ * alloc.allocated() -- get the amount of free memory.
+ */
```
Totktonada, line 129: Typo: alloc.allocated() -> alloc.free().

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+}
+
+void
+tarantool_lua_alloc_init(struct lua_State *L)
+{
+	static const struct luaL_Reg alloc_methods[] = {
+		{"setmemory", lbox_alloc_setmemory},
+		{"getmemory", lbox_alloc_getmemory},
+		{"allocated", lbox_alloc_allocated},
+		{"free",      lbox_alloc_free},
```
Totktonada, line 144: If we use `ibuf`'s terminology, it is `used` and `unused`. The `free` word feels like a verb ('free memory').

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+	lua_pushinteger(L, memory - allocated);
+	return 1;
+}
+
+void
+tarantool_lua_alloc_init(struct lua_State *L)
+{
+	static const struct luaL_Reg alloc_methods[] = {
+		{"setmemory", lbox_alloc_setmemory},
+		{"getmemory", lbox_alloc_getmemory},
```
Totktonada, line 142: How about getlimit/setlimit? Seems more intuitive.

Or we can use word 'quota'.

src/lua/alloc.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <stddef.h>
+
+#include <lua.h>
+
+/**
+ * Limits the maximum available Lua memory.
+ */
+int
+luaT_setallocmemory(size_t memory);
```
Totktonada, line 46: It seems, it is not used anywhere outside of `alloc.c`.

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+
+	/*
+	 * Result may be NULL, in this case "not enough memory"
+	 * is raised.
+	 */
+	return result;
+}
+
+int
+luaT_setallocmemory(size_t amount)
```
Totktonada, line 52: Seems like `luaT` prefix is not needed: the function doesn't access a Lua state.

src/lua/alloc.c
Totktonada: Seems OK for me, thanks for the update!

src/box/lua/config/applier/lua.lua
```diff
@@ -0,0 +1,37 @@
...
+        alloc.setmemory(memory)
+
+        log.verbose(('lua.apply: set memory limit to %d'):format(memory))
+    -- If memory limit is decreased.
+    elseif memory < old_memory then
+        local name = 'lua_memory_decreased'
+        local warning = 'lua.apply: lua.memory decrease will not be applied ' ..
+                        'until the instance is restarted'
+
+        config._aboard:set({type = 'warn', message = warning}, {key = name})
```
Totktonada, line 26: I'm OK, but if the underlying API allows to decrease it to some extent, can we allow to do so from the config? In the similar way: if it is impossible, report an alert and wait for reloading.

I'm OK with the refusal to descrease it in runtime, just curious.

@georgiy-belyanin @Buristan What do you think?

Totktonada (approved):
I've no objections here. I wrote some remarks/ideas down, but they look optional. OK from me.

src/lua/alloc.c
Buristan: `alloc.setlimit()`/`alloc.getlimit()` sounds better for me too.

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Lua custom memory allocation function. It extends the original
+ * one with a memory counter and a limit check.
+ */
+static void *
+luaT_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	size_t new_allocated = allocated + nsize - osize;
+	if (new_allocated > memory) {
+		/* Returning NULL results with "not enough memory" error. */
```
Buristan, line 35: Nit: Comment line width is more than 66 symbols.
Typo: s/results with/results in the/

src/lua/alloc.c
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * The default allocator function which is wrapped into a new one
+ * with the Lua memory limit checker.
+ */
+static lua_Alloc orig_alloc;
+/**
+ * The maximum allowed LuaJIT memory limit.
+ */
+static size_t memory = LUA_MEMORY_DEFAULT;
```
Buristan, line 20: Minor: `memory_limit` sounds more clear to me (`memory` associates with a "currently allocated memory").
Feel free to ignore.

src/box/lua/config/applier/lua.lua
```diff
@@ -0,0 +1,37 @@
...
+    end
+    -- Nothing to do if memory limit is unchanged.
+
+
+    is_lua_memory_applied = true
```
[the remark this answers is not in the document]
Buristan: I think that maybe we should alow to decrease the memory limit, if it is not less than the currently allocated memory amount, since I see no drawbacks of this. Also, this alows to drop the special first-call case.

But I'm OK with your's or Georgiy's approach.

Buristan (approved):
Hi, Georgiy!
Thanks for the fixes!
I'll proceed with the review per-patch below.

---
**[PATCH 1/2] lua: add custom allocator with memory limit**

LGTM after fixing Alexander's comments, with a few nits regarding the commit message and comments.

> This patch adds support for custom Lua allocator supporting memory

Typo: s/custom/the custom/


> New internal Lua module `internal.alloc` could be used to access the

Typo: s/New/The new/

> checking currently allocated/free amount of memory.

Typo: s/currently/the currently/

Buristan (approved):
**[PATCH 2/2] config: add `lua.memory` option**

LGTM, with some notes below.

src/lua/alloc.c
georgiy-belyanin: Thank you for noticing the typo.

But for the line width: it's less than 66. Or one tab counts as 8 spaces? If so, I've fixed it.
And, actually, I don't understand why the limit should be 66. I see no such point in [https://www.tarantool.io/en/doc/latest/contributing/c_style_guide/](https://www.tarantool.io/en/doc/latest/contributing/c_style_guide/).

src/lua/alloc.c
georgiy-belyanin: Applied.

src/lua/alloc.c
georgiy-belyanin: Applied. Thank you.

src/lua/alloc.c
georgiy-belyanin: Now the method is called `alloc.unused()`. Applied.

src/lua/alloc.c
georgiy-belyanin: Oh, I am sorry. Applied

src/lua/alloc.c
Totktonada: 66 characters limit was enforced for comments in the past as optimal width for a prose text readability. But it was not about a significant columns, includes the indent. Don't know why. Later we agreed that 80 symbols is good too.

Personally, I continue to use 66, but it is not enforced anymore.

I guess Sergey refers to the old limit.

src/lua/alloc.c
georgiy-belyanin: Applied. You're right it's clearer.

src/lua/alloc.c
georgiy-belyanin: I've removed it.  Now it is `set_alloc_limit` .Thank you.

src/box/lua/config/applier/lua.lua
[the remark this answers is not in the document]
georgiy-belyanin: I've worked on this and I suggest to allow the limit to be decreased up to the current amount of the used Lua memory plus some extra unused space (I've chosen arbitrary value of 16MB) to make sure not to instantly throw and error after setting the limit. If there is already too much allocated memory the alert is issued.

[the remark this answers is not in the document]
Totktonada: Allowed limit values in `[used+16MiB; +inf]` look OK for me.

src/lua/alloc.c
```diff
@@ -0,0 +1,153 @@
...
+static int
+lbox_alloc_getlimit(struct lua_State *L)
+{
+	lua_pushinteger(L, memory_limit);
+	return 1;
+}
+
+/**
+ * alloc.set() -- set the allocator memory limit.
+ * Returns old memory limit.
```
Totktonada, line 93: Nit: alloc.set() -> alloc.setlimit()

src/box/lua/config/applier/lua.lua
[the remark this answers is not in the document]
Totktonada: I would possibly prefer to handle this safely memory amount on the `alloc.c` level, but it may require to report different error codes to the config's applier and complicate things so. Anyway, I'm OK with how it is done now, just thoughts.

src/lua/alloc.c
georgiy-belyanin: Applied. Thank you.

src/lua/alloc.c
Buristan: OK, maybe it is worth updating the Wiki then (I'm personally afraid to alter Vlad's manual :) ).

Issue #67, opened 2013-10-15
wal_mode = "async"


pmwkaa:
With the new architecture of WAL writer, it's way more feasible to implement asynchronous write mode: when clients don't wait for writes to disk. Some of the vendors even pride themselves in this mode (MemSQL), and generally this technique seems to be commonplace on the market (Redis).

If only for one reason, we need this mode to win in the last remaining benchmarks where we may not show off as well as we can. 

How to do it:
- from user perspective, extend iproto header with WAL_ASYNC flag, set to 0 by default 
- add a general wal_write="async" mode, which sets this flag to true for all write requests

When this flag is set, allocate struct wal_request on one of two rotating regions, not on fiber->gc,
and set it a different route. The route return function should decrement region reference counter,
and when the region has no references, it can be garbage collected become available for reuse.

mary3000:
1. It is decided that better to introduce an option to the commit itself (i.e. `box.commit({is_async=true})`) than to the whole wal. It is more explicit and will not change the behavior of implicit transactions.
2. As for now, the functionality will be introduced only for transactions in case of lua, and for iproto requests. That is, standalone requests in autocommit mode will not have this option.

PR #10225, opened 2024-07-10
Async commit

Gerold103:
See #67 and the commit messages.

### DIFF
--- a/src/box/errcode.h
+++ b/src/box/errcode.h
@@ -431,6 +431,7 @@ struct errcode_record {
 	 */ \
 	_(ER_READ_VIEW_BUSY, 285,		"The read view is busy") \
 	_(ER_READ_VIEW_CLOSED, 286,		"The read view is closed") \
+	_(ER_WAL_QUEUE_FULL, 287,		"The WAL queue is full") \
 	TEST_ERROR_CODES(_) /** This one should be last. */
 
 /*
--- a/src/box/journal.h
+++ b/src/box/journal.h
@@ -180,10 +180,6 @@ struct journal {
 	/** Asynchronous write */
 	int (*write_async)(struct journal *journal,
 			   struct journal_entry *entry);
-
-	/** Synchronous write */
-	int (*write)(struct journal *journal,
-		     struct journal_entry *entry);
 };
 
 /** Wake the journal queue up. */
@@ -263,20 +259,6 @@ journal_async_complete(struct journal_entry *entry)
  */
 extern struct journal *current_journal;
 
-/**
- * Write a single entry to the journal in synchronous way.
- *
- * @return 0 if write was processed by a backend or -1 in case of an error.
- */
-static inline int
-journal_write(struct journal_entry *entry)
-{
-	journal_queue_flush();
-	journal_queue_on_append(entry);
-
-	return current_journal->write(current_journal, entry);
-}
-
 /** Write a single row in a blocking way. */
 int
 journal_write_row(struct xrow_header *row);
@@ -293,6 +275,17 @@ journal_write_try_async(struct journal_entry *entry)
 	journal_queue_on_append(entry);
 
 	return current_journal->write_async(current_journal, entry);
+}
+
+/** Write a single entry to the journal in synchronous way. */
+static inline int
+journal_write(struct journal_entry *entry)
+{
+	if (journal_write_try_async(entry) != 0)
+		return -1;
+	while (!entry->is_complete)
+		fiber_yield();
+	return 0;
 }
 
 /**
@@ -325,18 +318,15 @@ journal_set(struct journal *new_journal)
 static inline void
 journal_create(struct journal *journal,
 	       int (*write_async)(struct journal *journal,
-				  struct journal_entry *entry),
-	       int (*write)(struct journal *journal,
-			    struct journal_entry *entry))
+				  struct journal_entry *entry))
 {
-	journal->write_async	= write_async;
-	journal->write		= write;
+	journal->write_async = write_async;
 }
 
 static inline bool
 journal_is_initialized(struct journal *journal)
 {
-	return journal->write != NULL;
+	return journal->write_async != NULL;
 }
 
 #if defined(__cplusplus)
--- a/src/box/journal.h
+++ b/src/box/journal.h
@@ -269,7 +269,7 @@ journal_write_row(struct xrow_header *row);
  * @return 0 if write was queued to a backend or -1 in case of an error.
  */
 static inline int
-journal_write_try_async(struct journal_entry *entry)
+journal_write_submit(struct journal_entry *entry)
 {
 	journal_queue_wait();
 	journal_queue_on_append(entry);
@@ -281,7 +281,7 @@ journal_write_try_async(struct journal_entry *entry)
 static inline int
 journal_write(struct journal_entry *entry)
 {
-	if (journal_write_try_async(entry) != 0)
+	if (journal_write_submit(entry) != 0)
 		return -1;
 	while (!entry->is_complete)
 		fiber_yield();
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -502,6 +502,8 @@ lbox_commit(lua_State *L)
 			/* Nop. */
 		} else if (strcmp(wait_mode_str, "submit") == 0) {
 			wait_mode = TXN_COMMIT_WAIT_MODE_SUBMIT;
+		} else if (strcmp(wait_mode_str, "none") == 0) {
+			wait_mode = TXN_COMMIT_WAIT_MODE_NONE;
 		} else {
 			diag_set(IllegalParams, "unknown 'wait' mode");
 			return luaT_error(L);
--- a/src/box/lua/init.c
+++ b/src/box/lua/init.c
@@ -471,37 +471,44 @@ static int
 lbox_commit(lua_State *L)
 {
 	int top = lua_gettop(L);
-	if (top >= 1) {
-		/* box.commit(nil) */
-		if (lua_isnil(L, 1))
-			goto box_commit;
-
-		if (lua_type(L, 1) != LUA_TTABLE) {
-			diag_set(IllegalParams, "options should be a table");
-			return luaT_error(L);
-		}
-		/* Get is_sync */
-		lua_getfield(L, 1, "is_sync");
-		/* box.commit({is_sync = nil}) */
-		if (lua_isnil(L, -1))
-			goto box_commit;
-
+	enum txn_commit_wait_mode wait_mode = TXN_COMMIT_WAIT_MODE_COMPLETE;
+	if (top == 0)
+		goto box_commit;
+	if (lua_isnil(L, 1))
+		goto box_commit;
+	if (lua_type(L, 1) != LUA_TTABLE) {
+		diag_set(IllegalParams, "options should be a table");
+		return luaT_error(L);
+	}
+	lua_getfield(L, 1, "is_sync");
+	if (!lua_isnil(L, -1)) {
 		if (!lua_isboolean(L, -1)) {
 			diag_set(IllegalParams, "is_sync must be a boolean");
 			return luaT_error(L);
 		}
 		bool is_sync = lua_toboolean(L, lua_gettop(L));
-		lua_pop(L, 1);
-
 		if (!is_sync) {
 			diag_set(IllegalParams, "is_sync can only be true");
 			return luaT_error(L);
 		}
 		box_txn_make_sync();
 	}
-
+	lua_getfield(L, 1, "wait");
+	if (!lua_isnil(L, -1)) {
+		const char *wait_mode_str = lua_tostring(L, -1);
+		if (wait_mode_str == NULL) {
+			/* Nop. */
+		} else if (strcmp(wait_mode_str, "complete") == 0) {
+			/* Nop. */
+		} else if (strcmp(wait_mode_str, "submit") == 0) {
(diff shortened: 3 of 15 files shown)


p7nov (approved):

src/box/txn.h
```diff
@@ -110,6 +110,11 @@ enum txn_flag {
 	 * Transaction has been rolled back so it cannot be continued.
 	 */
 	TXN_IS_ROLLED_BACK = 0x200,
+	/**
+	 * Transaction is not original. It is replayed from the journal or is
+	 * replicated from a master.
+	 */
+	TXN_IS_REPLAYED = 0x400,
```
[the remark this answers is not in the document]
Serpentian: I applied the following patch on top of your PR. Everything seems to work just fine:

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 636f40fa1c..9e1bd93ec6 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -786,7 +786,7 @@ wal_stream_apply_dml_row(struct wal_stream *stream, struct xrow_header *row)
 			say_error("couldn't begin a recovery transaction");
 			return -1;
 		}
-		txn_set_flags(txn, TXN_IS_REPLAYED);
+		// txn_set_flags(txn, TXN_IS_REPLAYED);
 	} else if (row->tsn != stream->tsn) {
 		diag_set(XlogError, "found a next transaction with the "
 			 "previous one not yet committed");
diff --git a/src/box/txn.c b/src/box/txn.c
index 9186834ac8..b4b19ed45f 100644
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1098,8 +1098,9 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 		 * logically strange to commit a synchronous transaction in an
 		 * async way, but the CONFIRM problem is the technical one.
 		 */
+		bool is_replayed = req->rows[0]->replica_id != 0;
 		if (txn_has_flag(txn, TXN_WAIT_ACK) &&
-		    !txn_has_flag(txn, TXN_IS_REPLAYED)) {
+		    !is_replayed) {
 			diag_set(ClientError, ER_UNSUPPORTED, "Tarantool",
 				 "Non-blocking commit of a synchronous txn");
 			goto rollback_abort;

```

src/box/txn.c
```diff
@@ -1069,65 +1069,41 @@ txn_add_limbo_entry(struct txn *txn, const struct journal_entry *req)
...
+			goto rollback_abort;
+		});
+		/*
+		 * The main reason why it is disabled is that the sync txn's
+		 * original instance wants to write CONFIRM. But it can't be
+		 * done asynchronously in the current code design. Also it is
+		 * logically strange to commit a synchronous transaction in an
+		 * async way, but the CONFIRM problem is the technical one.
+		 */
+		if (txn_has_flag(txn, TXN_WAIT_ACK) &&
```
Serpentian, line 1101: Maybe we can move this check to the beginning of the `txn_commit_impl`? I see no sense to prepare the tx if we'll anyway fail. Also the errinj was at the beginning, though it seems to work ok here too

src/box/txn.c
```diff
@@ -1103,6 +1104,11 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 				 "Non-blocking commit of a synchronous txn");
 			goto rollback_abort;
 		}
+		if (wait_mode == TXN_COMMIT_WAIT_MODE_NONE &&
+		    journal_queue_is_full()) {
+			diag_set(ClientError, ER_JOURNAL_FULL);
```
Serpentian, line 1109: Users don't know such concept as `JOURNAL`. There's not a single mention of it in documentation. Can we rename it to smth like `ER_WAL_QUEUE_FULL`?

Serpentian (changes requested):
Thank you for the patch. Looks great, left some comments)

src/box/txn.c
Gerold103: If I put this before prepare, then it would also be before `txn_journal_entry_new()`. This one fills `journal_entry` with necessary stuff, like `rows[0]->replica_id` and various flags. So if I move this before preparation, then I also have to create the journal entry before preparation. It seems worse, because prepare can fail during regular runtime if the txn was aborted by a conflict, for example, while this error right here only happens as a programming error in the application code, i.e. it mustn't happen at all, ideally.

The motivation was that we better optimize the more common runtime error path and leave the code error path in a somewhat suboptimal state.

src/box/lua/init.c
```diff
@@ -468,37 +468,44 @@ static int
 lbox_commit(lua_State *L)
```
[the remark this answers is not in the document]
Gerold103: Oh, how could I forget 😨. Done now, thanks for noticing.

src/box/journal.h
```diff
@@ -291,6 +273,17 @@ journal_write_try_async(struct journal_entry *entry)
 	return current_journal->write_async(current_journal, entry);
 }
 
+/** Write a single entry to the journal in synchronous way. */
+static inline int
+journal_write(struct journal_entry *entry)
+{
+	if (journal_write_try_async(entry) != 0)
```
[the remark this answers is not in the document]
Gerold103: Not really. Previously it would be probably more reactive, because any blocking commit would wakeup all the pending txns and then go in too. Now the txn goes to the end of the queue and just waits until the current WAL write is done.

The only theoretical change is that now the WAL batching might improve (blocking txns will also stand into the queue without rushing into the WAL thread), while reactivity might decrease.

src/box/journal.h
```diff
@@ -263,20 +259,6 @@ journal_async_complete(struct journal_entry *entry)
...
 
-/**
- * Write a single entry to the journal in synchronous way.
- *
- * @return 0 if write was processed by a backend or -1 in case of an error.
- */
-static inline int
-journal_write(struct journal_entry *entry)
-{
-	journal_queue_flush();
```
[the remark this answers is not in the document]
Gerold103: Done.

src/box/txn.h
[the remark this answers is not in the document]
Gerold103: Indeed, didn't think of that! Good solution, thanks, applied it.

The only nit I could think of is that relying on `replica_id == 0` as a flag whether the txn has been in any WALs is a bit shaky. If we ever start assigning it before sending txns to WAL, this assumption will break. But for I now it looks good, thanks again.

src/box/txn.c
Gerold103: Absolutely right, good that you noticed, I changed the error name and message.

src/box/lua/schema.lua
```diff
@@ -447,14 +464,16 @@ box.atomic = function(arg0, arg1, ...)
         if not utils.is_callable(arg1) then
             box.error(box.error.ILLEGAL_PARAMS, usage, 2)
         end
-        call_at(2, box.begin, arg0)
-        return atomic_tail(1, pcall(arg1, ...))
+        local options = arg0
+        check_param_table(options, atomic_options, 2)
+        call_at(2, box_begin_impl, options)
```
Serpentian, line 469: I didn't get, why don't we pass `level ` to `box_begin_impl` here

src/box/errcode.h
```diff
@@ -431,6 +431,7 @@ struct errcode_record {
 	 */ \
```
Serpentian, line 431: We decided, that from now on we specify since which version feature is present in DocBot request. It'll be `3.3.0`, i suppose? Doc team struggles a lot to figure this out.

Serpentian (approved):
Thank you for the patch! Great work!

src/box/txn.c
```diff
@@ -1069,65 +1069,41 @@ txn_add_limbo_entry(struct txn *txn, const struct journal_entry *req)
...
+			goto rollback_abort;
+		});
+		/*
+		 * The main reason why it is disabled is that the sync txn's
+		 * original instance wants to write CONFIRM. But it can't be
+		 * done asynchronously in the current code design. Also it is
+		 * logically strange to commit a synchronous transaction in an
+		 * async way, but the CONFIRM problem is the technical one.
+		 */
+		bool is_original = req->rows[0]->replica_id == 0;
```
sergepetrenko, line 1101: It seems `txn->n_new_rows > txn->n_local_rows` check would suit even better here.

src/box/txn.c
```diff
@@ -1193,6 +1177,18 @@ txn_commit(struct txn *txn)
 	return -1;
 }
 
+int
+txn_commit_submit(struct txn *txn)
```
sergepetrenko, line 1187: Do you think we should expose `box_txn_commit_submit` to the public C API as well?

src/box/lua/schema.lua
Gerold103: `call_at` takes level as the first argument. We can also pass it to `begin_impl`, but `call_at` overrides the level anyway.

src/box/errcode.h
Gerold103: Tbh, I don't know which version is the next one. Why not 3.2.0? I am a bit out of touch on the versions topic, sorry.

src/box/txn.c
Gerold103: Hm? Why? `txn_update_row_counts()` does the same check for replica ID. It seems `replica_id == 0` is the primary check from which others like row count can be derived.

src/box/txn.c
Gerold103: Nope. Only if somebody asks. If you think this is needed, then I can do it.

src/box/lua/schema.lua
Gerold103: Actually, thanks for the comment - I noticed that I don't need `call_at` here. I am passing the level directly to `box_begin_impl` now.

src/box/errcode.h
Serpentian: We wanted to release `3.2.0` recently, but it still has not been released and there's new `3.1.1`, so it looks like `3.2.0` is the next one indeed

src/box/txn.c
sergepetrenko: Ok, let's wait for people's request.

src/box/txn.c
Gerold103: > and IMO it's better to have a single place where we derive some conclusions out of replica id being zero

I totally agree. I would say even more - it would be better to have a function like `txn_is_replayed(txn)`. The problem here is that we don't have even just 2 places which use this "property" in the same way.
- `txn_add_limbo_entry()` uses `req->rows[0]->replica_id` as the origin ID. Transaction originator. It needs the specific ID of the txn author.
- `txn_commit_impl()` now uses `req->rows[0]->replica_id != 0` as a fact that the txn is replayed. It doesn't need to know the ID.
- `txn_update_row_counts()` works with statement, not txn. It looks at `stmt->row->replica_id == 0`. Again as a property whether it is replayed, but not the txn. Just this row.
- `txn_journal_entry_new()` uses `stmt->row->replica_id == 0` to see if the row is created locally or came remotely (replayed != remote).

As you can see, even if I introduce a function like `txn_is_replayed()`, it is going to be used in a single place. If I introduce a function `txn_get_origin_id()`, then will also be used in a single place, and will be lying (return 0) for the locally created but not yet committed txns, and won't tell if it is replayed or not (origin can be us, but can still be replayed during recovery).

Also it is clear, that *in case someday it changes* (the fact that `replica_id != 0` means a replayed txn), everything will break quite spectacularly.

> your check also relies on the fact that when a transaction is remote, local rows from before/on_replace triggers are never put first

Yes. We have that guarantee AFAIU. `txn_journal_entry_new()` builds the journal entry in this way. Non-local rows go first. Moreover, I treat it as a property of "replayed" transactions. Not only remote, but also replayed from WAL during recovery. Otherwise if that isn't true, the stuff above will also break and we will need to reconsider from scratch how do we determine origin of the transactions. Limbo would break for sure.

And you are absolutely correct about me mentioning shakiness of this construction. I would much more like if we would store author ID somewhere explicitly and not deduct it from the rows. But that would be a too serious change, to introduce any sort of transaction headers.

src/box/errcode.h
Gerold103: Ok, I mentioned 3.2.0 in the docbot request.

CuriousGeorgiy:
@Gerold103 could you please check whether this function needs an update?
https://github.com/tarantool/tarantool/blob/625afe418151de1015483ed24acaa57e1f85efb3/src/box/txn.h#L574-L590

It seems like we should also check if a transaction has status `TXN_PREPARED` so that the `on_rollback` trigger can forbid any database access.

src/box/txn.c
sergepetrenko: Thanks for taking the time to make this overview!

Ok, let's leave the check as is for now.

sergepetrenko (approved):

sergepetrenko:
@TarantoolBot document
Title: New `box.commit` and `box.atomic` option `wait`

Target: 3.2.0.

It can be passed like this: `box.commit({wait = ...})` or
`box.atomic({wait = ...}, ...)`. The possible values:
- `'complete'` - default behaviour, block the current fiber until
    the txn is fully finished.
- `'submit'` - send the txn to the journal, release the fiber.
    Completion happens asynchronously. Can block the fiber if the
    journal queue is full, until the txn can fit into the queue.
- `'none'` - send the txn to the journal if the journal queue has
    space, otherwise fail right away. Commit with this option
    never yields.

In case of a Lua error being thrown the txn is rolled back, like
with the default behaviour, regardless of its wait mode.

The journal max size can be found and changed in
`box.cfg.wal_queue_max_size`. It affects when the wait modes
`submit` blocks and `none` fails.

PR #10229, opened 2024-07-11
[backport 2.11] box: update synchro quorum in on_commit trigger instead of on_replace

### LINKED ISSUE #10087: Update quorum in `on_commit` trigger instead of `on_replace` trigger.

sergepetrenko:
This is a backport of #10091. ~~Note, that only the first commit was backported (https://github.com/tarantool/tarantool/commit/29d1c0fa6c76ff06d748d8916c5ffbbc9e4a55e9), because the second one (https://github.com/tarantool/tarantool/commit/9b63ced353eaf6c1224358c7d3655ca52f7d21fb) makes the situation worse without an additional backport of #10114~~ nevermind, backported commit https://github.com/tarantool/tarantool/commit/9b63ced353eaf6c1224358c7d3655ca52f7d21fb as well.

Currently, we update the synchronous replication quorum from the `on_replace` trigger of the `_cluster` space when registering a new replica. However, during the join process, the replica cannot ack its own insertion into the `_cluster` space. In the scope of #9723, we are going to enable synchronous replication for most of the system spaces, including the `_cluster` space. There are several problems with this:

1. Joining a replica to a 1-member cluster without manual changing of quorum won't work: it is impossible to commit the insertion into the `_cluster` space with only 1 node, since the quorum will equal to 2 right after the insertion.

2. Joining a replica to a 3-member cluster may fail: the quorum will become equal to 3 right after the insertion, the newly joined replica cannot ACK its own insertion into the `_cluster` space — if one out of original 3 nodes fails, then reconfiguration will fail.

Generally speaking, it will be impossible to join a new replica to the cluster, if a quorum, which includes the newly added replica (which cannot ACK), cannot be gathered.

To solve these problems, let's update the quorum in the `on_commit` trigger. This way we’ll be able to insert a node regardless of the current configuration. This somewhat contradicts with the Raft specification, which requires application of all configuration changes in the `on_replace` trigger (i.e., as soon as they are persisted in the WAL, without quorum confirmation), but still forbids several reconfigurations at the same time.

Closes #10087

NO_DOC=<no special documentation page devoted to cluster reconfiguration>

(cherry picked from commit 29d1c0fa6c76ff06d748d8916c5ffbbc9e4a55e9)

### DIFF
--- a/changelogs/unreleased/gh-10087-update-quorum-in-on_commit-trigger.md
+++ b/changelogs/unreleased/gh-10087-update-quorum-in-on_commit-trigger.md
@@ -1,0 +1,7 @@
+## bugfix/replication
+
+* Fixed the inability to add a new replica to the replicaset if the user has
+  manually made space `_cluster` synchronous. Now the synchronous replication
+  quorum is updated after the `_cluster` change is confirmed by a quorum rather
+  than immediately after persisting the configuration change in the WAL
+  (gh-10087).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4128,6 +4128,15 @@ on_replace_cluster_clear_id(struct trigger *trigger, void * /* event */)
 	return 0;
 }
 
+/** Update the synchronous replication quorum. */
+static int
+on_replace_cluster_update_quorum(struct trigger * /* trigger */,
+				 void * /* event */)
+{
+	box_update_replication_synchro_quorum();
+	return 0;
+}
+
 /**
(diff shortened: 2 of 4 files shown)


Gerold103 (approved):
💪

PR #10244, opened 2024-07-12
schema: error msg for array instead of map/record

### LINKED ISSUE #10241: schema: error message for an array instead of a map
If a map field is expected in a config field, and an array is received instead, it would be great to get a more detailed and informative error message, rather than a hard-to-understand error being thrown.

```config.yaml
credentials:
  users:
    me:
      passwd:
        plain: secret
      grant: [super]
groups:
  g:
    replicasets:
      r:
        instances:
          i:
            log:
              modules: [1,2,3]
```

```
$ tarantool --name i --config config.yaml
LuajitError: [cluster_config] groups.g.replicasets.r.instances.i.log.modules[1]: Unexpected data for scalar "string": Expected "string", got "number"
```

The goal is to provide a detailed, user-friendly error message for example
```
LuajitError: [cluster_config] groups.g.replicasets.r.instances.i.log.modules: Unexpected data: Expected "map", got "array": [1, 2, 3]
```

Corresponds to TNTP-239.

### LINKED ISSUE #10242: schema: error message for an array instead of a record
If a record field is expected in a config field, and an array is received instead, it would be great to get a more detailed and informative error message, rather than a hard-to-understand error being thrown.

```config.yaml
credentials:
  users:
    me:
	  passwd:
        plain: secret
      grant: [super]

groups: 
  g:
    replicasets:
      r:
        instances:
          i:
            iproto: [1,2,3]
```

```
$ tarantool --name i --config config.yaml
LuajitError: [cluster_config] groups.g.replicasets.r.instances.i.iproto: Unexpected field "1"
```

The goal is to provide a detailed, user-friendly error message for example
```
LuajitError: [cluster_config] groups.g.replicasets.r.instances.i.iproto: Unexpected data: Expected "record", got "array": [1, 2, 3]
```

This issue is a continuation of issue #10241

Corresponds to TNTP-240.

mandesero:
This patch addresses to issues where if a map/record field is
expected in a config field, and an array is received instead,
the error message thrown is not very detailed or user-friendly.

Now `{[1] = true}` (see test/config-luatest/schema_test.lua:
test_validate_map) is defined as an array rather than a map with
key 1, since this map representation corresponds to the array
representation `{true}`.
So it was replaced with `{[2] = true}`, which is not an array.

Closes #10241
Closes #10242

NO_DOC=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/box/lua/config/utils/schema.lua
+++ b/src/box/lua/config/utils/schema.lua
@@ -517,6 +517,14 @@ local function validate_by_node_function(schema, data, ctx)
     schema.validate(data, w)
 end
 
+-- Report an error if data is array but expected other type
+local function walkthrough_error_array(ctx, schema, data)
+    local status, _ = pcall(validate_table_is_array, data, ctx)
+    if not status or #data == 0 then return end
+    walkthrough_error(ctx, 'Unexpected data: Expected %q, got "array": %s',
+                schema.type, json.encode(data))
+end
+
 local function validate_impl(schema, data, ctx)
     if is_scalar(schema) then
         local scalar_def = scalars[schema.type]
@@ -529,6 +537,7 @@ local function validate_impl(schema, data, ctx)
         end
     elseif schema.type == 'record' then
         walkthrough_assert_table(ctx, schema, data)
+        walkthrough_error_array(ctx, schema, data)
 
         for field_name, field_def in pairs(schema.fields) do
             walkthrough_enter(ctx, field_name)
@@ -549,6 +558,7 @@ local function validate_impl(schema, data, ctx)
         end
     elseif schema.type == 'map' then
         walkthrough_assert_table(ctx, schema, data)
+        walkthrough_error_array(ctx, schema, data)
 
         for field_name, field_value in pairs(data) do
             walkthrough_enter(ctx, field_name)
@@ -1366,6 +1376,9 @@ local function merge_impl(schema, a, b, ctx)
         walkthrough_assert_table(ctx, schema, a)
         walkthrough_assert_table(ctx, schema, b)
 
+        walkthrough_error_array(ctx, schema, a)
+        walkthrough_error_array(ctx, schema, b)
+
         local res = {}
         for field_name, field_def in pairs(schema.fields) do
             walkthrough_enter(ctx, field_name)
@@ -1378,6 +1391,9 @@ local function merge_impl(schema, a, b, ctx)
     elseif schema.type == 'map' then
         walkthrough_assert_table(ctx, schema, a)
         walkthrough_assert_table(ctx, schema, b)
+
+        walkthrough_error_array(ctx, schema, a)
+        walkthrough_error_array(ctx, schema, b)
 
         local res = {}
         for field_name, a_field in pairs(a) do
--- a/test/config-luatest/schema_test.lua
+++ b/test/config-luatest/schema_test.lua
@@ -747,6 +747,12 @@ g.test_validate_record = function()
         '[myschema] foo: Unexpected data for scalar "string"',
         'Expected "string", got "boolean"',
     }, ': '))
+
+    -- Bad case: field is an array.
+    assert_validate_error(
+        s, {1, 2, 3},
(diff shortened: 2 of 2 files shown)


src/box/lua/config/utils/schema.lua
```diff
@@ -18,6 +18,65 @@ local schema_mt = {}
...
+        else
+            return -1
+        end
+    end
+    if max >= count * 2 then
+        return -1
+    end
+
+    return max
+end
```
mkokryashkin, line 76: We have a function in this exact file that is called `validate_table_is_array`, I believe you should find a way to reuse it or generalize the implementation.

src/box/lua/config/utils/schema.lua
```diff
@@ -18,6 +18,65 @@ local schema_mt = {}
...
+        else
+            result = result .. tostring(v)
+        end
+        if i < #array then
+            result = result .. ", "
+        end
+    end
+    result = result .. "]"
+    return result
+end
```
mkokryashkin, line 39: I suppose you should reuse the existing implementation of lua formatter. We have `internal.format_lua`, you can look into `src/box/lua/console.lua` for inspiration.

mkokryashkin (changes requested):

mkokryashkin (commented):
Shouldn't `merge_impl` checked too? There are assertions for tables and separate definitions for map and record, but no implementation for array.

test/config-luatest/schema_test.lua
```diff
@@ -747,6 +747,11 @@ g.test_validate_record = function()
         '[myschema] foo: Unexpected data for scalar "string"',
         'Expected "string", got "boolean"',
     }, ': '))
+
+    -- Bad case: field is an array.
+    assert_validate_error(s, {1, 2, 3}, table.concat({
+        '[myschema] Unexpected data: Expected "record", got "array": {1, 2, 3}',
+    }, ': '))
```
mkokryashkin, line 754: `table.concat` seems redundant here.

test/config-luatest/schema_test.lua
```diff
@@ -766,11 +771,19 @@ g.test_validate_map = function()
...
+    assert_validate_error(s, {[2] = true}, table.concat({
+        '[myschema] [2]: Unexpected data for scalar "string"',
         'Expected "string", got "number"',
     }, ': '))
 
+    -- Bad case: data is an array
+    assert_validate_error(s, {1, 2, 3}, table.concat({
+        '[myschema] Unexpected data: Expected "map", got "array": {1, 2, 3}',
+    }, ': '))
+
```
mkokryashkin, line 786: `table.concat` seems redundant here.

mkokryashkin (changes requested):

mkokryashkin (approved):

src/box/lua/config/utils/schema.lua
Totktonada: I vote for using the JSON encoder for prettifying data in an error/log message. We already do it here and there.

Also, in my opinion this Lua formatter is barely ready for production use: it was written once and we never touch it again. There is a couple of [known problems](https://github.com/tarantool/tarantool/issues?q=is%3Aopen+is%3Aissue+label%3A%22lua+output%22). (To be precise, @mandesero looked over them and reported that some of them are not reproducing anymore. However, this is likely a side effect of some other fixes -- there were no direct effect toward a good Lua formatter.)

I agree in your general point: we shouldn't write a formatter in place for a complex arbitrary data structure. Let's use something already implemented and widely used.

Totktonada:
We discussed it with @mandesero and he agreed with my proposal to steal the task.

My main point is that some record/map schemas accept array values. A record may have numeric field names and represent an array with a limited maximum length. A map may also have numeric keys.

Also, I wouldn't add extra validation to `:merge()`, because its contract assumes already validated data. It only performs some fast type checks to don't fail with 'attempt to index nil value'. Just to ease debugging.

I've proposed an alternative implementation in PR #10650.

@mandesero, @mkokryashkin I would appreciate if you glance on it.

Totktonada:
@mandesero I've incorporated your work into a reworked patch.

This pull request is superseded by #10650.

PR #10247, opened 2024-07-12
 small: bump new version with UBSan fixes 

LevKats:
This patch bumped small to the new version that does not trigger
UBSan with `*_entry*` macros and should support new oss-fuzz builder.

Fixes: https://github.com/tarantool/tarantool/issues/10143

NO_DOC=small submodule bump
NO_TEST=small submodule bump

### DIFF
--- a/src/trivia/util.h
+++ b/src/trivia/util.h
@@ -270,8 +270,12 @@ alloc_failure(const char *filename, int line, size_t size)
  * including padding if any.
  */
 #ifndef offsetof
-#define offsetof(type, member) ((size_t) &((type *)0)->member)
-#endif
+#  if __has_builtin(__builtin_offsetof)
+#    define offsetof(type, member) __builtin_offsetof(type, member)
+#  else
+#    define offsetof(type, member) ((size_t)&((type *)0)->member)
+#  endif
+#endif /* offsetof */
 
 /**
  * This macro is used to retrieve an enclosing structure from a pointer to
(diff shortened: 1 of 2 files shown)


src/lib/small
Serpentian, line 1: Please, specify a list of commits added in commit msg. Check out how we do this in other small bumps

```
    small: bump version
    
    New commits:
    
    * matras: increase max capacity from 2^31 to 2^32 blocks
    * matras: fix compilation on macOS
    
    Closes #9864
```

LevKats: Done, I guess

src/trivia/util.h
```diff
@@ -270,8 +270,12 @@ alloc_failure(const char *filename, int line, size_t size)
  * including padding if any.
  */
 #ifndef offsetof
-#define offsetof(type, member) ((size_t) &((type *)0)->member)
-#endif
+#  if defined(__has_builtin) && __has_builtin(__builtin_offsetof)
```
locker, line 273: `__has_builtin` is defined a few lines above so `defined(__has_builtin)` could be dropped:

https://github.com/tarantool/tarantool/blob/b0b32bff022caeffd64ca687baab908f25fa2fcb/src/trivia/util.h#L177-L185

locker (approved):

Serpentian (approved):

src/trivia/util.h
LevKats: Forgot about that. Thanks, I removed that check.

locker:

PR #10248, opened 2024-07-15
box: grant/revoke user right after upgrade

### LINKED ISSUE #9849: `SCHEMA_NEEDS_UPGRADE` error when Tarantool 3+ is loaded from old snapshot

georgiy-belyanin:
This PR:
- Uses `box.space._schema:on_replace()` for automatically granting/revoking user rights after upgrading the scheme without restarting.
- Adds error messages in case granting/revoking credentials broke.

Closes #9849

### DIFF
--- a/changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
+++ b/changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
@@ -1,0 +1,4 @@
+## bugfix/box
+
+* User rights are now automatically granted/revoked after upgrading
+  without restarting (gh-9849).
--- a/src/box/lua/config/applier/credentials.lua
+++ b/src/box/lua/config/applier/credentials.lua
@@ -3,6 +3,7 @@ local log = require('internal.config.utils.log')
 local loaders = require('internal.loaders')
 local digest = require('digest')
 local fiber = require('fiber')
+local mkversion = require('internal.mkversion')
 
 -- Var is set with the first apply() call.
 local config
@@ -516,8 +517,7 @@ local privileges_action_f = function(grant_or_revoke, role_or_user, name, privs,
     if err.code ~= box.error.NO_SUCH_SPACE and
             err.code ~= box.error.NO_SUCH_ROLE and
             err.code ~= box.error.NO_SUCH_FUNCTION and
-            err.code ~= box.error.NO_SUCH_SEQUENCE and
-            err.code ~= box.error.SCHEMA_NEEDS_UPGRADE then
+            err.code ~= box.error.NO_SUCH_SEQUENCE then
         err = ('credentials.apply: box.schema.%s.%s(%q, %q, %q, %q) failed: %s')
               :format(role_or_user, grant_or_revoke, name, privs, obj_type,
                       obj_name, err)
@@ -651,6 +651,52 @@ local function sync_privileges(credentials, obj_to_sync)
     config._aboard:drop_if(function(_key, alert)
         return alert._trait == 'missed_privilege_obsolete'
     end)
+end
+
+-- The flag indicating the box.space._schema:on_replace trigger
+-- is set. It's set when the config is applied if the schema
+-- version is not the latest. It's dropped when the schema was
+-- upgraded.
+local on_schema_replace_trigger_is_set = false
+
+-- The flag indicates if the schema is upgraded to the latest
+-- version. It's impossible to use the internal call
+-- `box.internal.schema_needs_upgrade()` because at the moment
+-- we are granting/revoking privileges its value isn't updated.
+local schema_is_upgraded = false
+
+-- The conditional is broadcasted when the schema was upgraded to
+-- the current tarantool version.
+local schema_is_upgraded_cond
+
+local function update_schema_upgraded_status(version)
+    local schema_version = version or mkversion.get()
+    schema_is_upgraded = schema_version == mkversion.get_latest()
+
+    if schema_is_upgraded then
+        assert(schema_is_upgraded_cond ~= nil)
+        schema_is_upgraded_cond:broadcast()
+    end
+end
+
+local function on_schema_replace_trigger(_, new)
+    assert(on_schema_replace_trigger_is_set)
+
+    if new == nil then
+        return
+    end
+
+    local latest_version = mkversion.get_latest()
+    local new_version = mkversion.from_tuple(new)
+
+    if new_version == latest_version then
+        box.on_commit(function()
+            update_schema_upgraded_status(new_version)
+        end)
+
+        on_schema_replace_trigger_is_set = false
+        box.space._schema:on_replace(nil, on_schema_replace_trigger)
+    end
 end
 
 -- }}} Main sync logic
@@ -809,6 +855,48 @@ end
 -- of main fiber in case of 'BLOCKING_FULL_SYNC'.
 local wait_sync
 
+local function sync_object(obj_to_sync)
+    if obj_to_sync.type == 'BLOCKING_FULL_SYNC' or
+        obj_to_sync.type == 'BACKGROUND_FULL_SYNC' then
+
+        if box.info.ro then
+            log.verbose('credentials: the database is in the read-only ' ..
+                'mode. Waiting for the read-write mode to set up the ' ..
+                'credentials given in the configuration. If another ' ..
+                'instance in the replicaset will write these new ' ..
+                'credentials, this instance will receive them and will ' ..
+                'skip further actions.')
+
+            box.ctl.wait_rw()
+        end
+
+        -- On config reload, drop the old list of registered objects.
+        target_object_map = {
+            ['role'] = {},
+            ['space'] = {},
+            ['function'] = {},
+            ['sequence'] = {},
+        }
+
+        local credentials = get_credentials(config)
+
+        register_objects(credentials.roles)
+        register_objects(credentials.users)
+
+        box.atomic(function()
+            create_roles(credentials.roles)
+            create_users(credentials.users)
+
+            sync_privileges(credentials)
+        end)
+
+    else
+        local credentials = get_credentials(config)
+
+        box.atomic(sync_privileges, credentials, obj_to_sync)
+    end
+end
+
 -- The worker gets synchronization commands from sync_tasks.
 -- If the instance is in RO mode, the worker will wait for it
 -- to be switched to RW.
@@ -819,49 +907,36 @@ local wait_sync
 -- {type = obj_type, name = obj_name} - perform a per-object sync, skip
 --                                      return message
 local function sync_credentials_worker()
+    fiber.self():name('sync_credentials', { truncate = true })
+
+    if not schema_is_upgraded then
+        local not_upgraded_alert_key = 'not_upgraded_schema'
+
+        config._aboard:set({
+            type = 'warn',
+            message = 'credentials: the database schema has an old version ' ..
+                      'and users/roles/privileges cannot be applied. '..
+                      'Consider executing box.schema.upgrade() to perform an '..
+                      'upgrade.'
+        }, {key = not_upgraded_alert_key})
+
+        schema_is_upgraded_cond:wait()
+        assert(schema_is_upgraded)
+
+        config._aboard:drop(not_upgraded_alert_key)
+    end
+
     while true do
         local obj_to_sync = sync_tasks:get()
 
-        if obj_to_sync.type == 'BLOCKING_FULL_SYNC' or
-                obj_to_sync.type == 'BACKGROUND_FULL_SYNC' then
+        local _, err = pcall(sync_object, obj_to_sync)
 
-            if box.info.ro then
-                log.verbose('credentials: the database is in the read-only ' ..
-                    'mode. Waiting for the read-write mode to set up the ' ..
-                    'credentials given in the configuration. If another ' ..
-                    'instance in the replicaset will write these new ' ..
-                    'credentials, this instance will receive them and will ' ..
-                    'skip further actions.')
-                box.ctl.wait_rw()
-            end
+        fiber.testcancel()
 
-            -- On config reload, drop the old list of registered objects.
-            target_object_map = {
-                ['role'] = {},
-                ['space'] = {},
-                ['function'] = {},
-                ['sequence'] = {},
-            }
-
-            local credentials = get_credentials(config)
-
-            register_objects(credentials.roles)
-            register_objects(credentials.users)
-
-            box.atomic(function()
-                create_roles(credentials.roles)
-                create_users(credentials.users)
-
-                sync_privileges(credentials)
-            end)
-
-            if obj_to_sync.type == 'BLOCKING_FULL_SYNC' then
-                wait_sync:put('Done')
-            end
-        else
-            local credentials = get_credentials(config)
-
-            box.atomic(sync_privileges, credentials, obj_to_sync)
+        if obj_to_sync.type == 'BLOCKING_FULL_SYNC' then
+            wait_sync:put(err or 'Done')
+        elseif err then
+            log.error(err)
         end
     end
 end
@@ -895,9 +970,20 @@ local function apply(config_module)
         wait_sync = fiber.channel()
     end
 
+    if not schema_is_upgraded_cond then
+        schema_is_upgraded_cond = fiber.cond()
+    end
+    update_schema_upgraded_status()
+
+    if not schema_is_upgraded and not on_schema_replace_trigger_is_set then
+        on_schema_replace_trigger_is_set = true
+        box.space._schema:on_replace(on_schema_replace_trigger)
+    end
+
     -- Set trigger on after space/function/sequence creation.
     if not triggers_are_set then
         triggers_are_set = true
+
         box.space._user:on_replace(on_replace_trigger)
         box.space._space:on_replace(on_replace_trigger)
         box.space._func:on_replace(on_replace_trigger)
@@ -916,11 +1002,14 @@ local function apply(config_module)
     -- If Tarantool is already in Read Write mode, credentials are still
     -- applied in the fiber to avoid possible concurrency issues, but the
     -- main applier fiber is blocked by `wait_sync`.
-    if not box.info.ro then
+    if not box.info.ro and schema_is_upgraded then
         -- Schedule a full sync with a result message on return.
         sync_tasks:put({type = 'BLOCKING_FULL_SYNC'})
 
-        wait_sync:get()
+        local sync_result = wait_sync:get()
+        if sync_result ~= 'Done' then
+            error(sync_result)
+        end
     else
         -- Schedule a full sync in the background. It will be executed
         -- when the instance switches to RW.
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
```diff
@@ -0,0 +1,6 @@
+## feature/box
+* Added `on_schema_upgrade_end` trigger (gh-9849).
```
andreyaksenov, line 2:
```suggestion
* Added the `on_schema_upgrade_end` trigger (gh-9849).
```

changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
```diff
@@ -0,0 +1,6 @@
...
+
+## bugfix/box
+* User rights are now granted/revoked after upgrading automatically without
+  restarting (gh-9849).
```
andreyaksenov, line 6:
```suggestion
* User rights are now automatically granted/revoked after upgrading without
  restarting (gh-9849).
```

andreyaksenov (approved):

src/box/lua/ctl.c
```diff
@@ -181,6 +188,7 @@ static const struct luaL_Reg lbox_ctl_lib[] = {
 	{"wait_rw", lbox_ctl_wait_rw},
 	{"on_shutdown", lbox_ctl_on_shutdown},
 	{"on_schema_init", lbox_ctl_on_schema_init},
+	{"on_schema_upgrade_end", lbox_ctl_on_schema_upgrade_end},
```
Totktonada, line 191: I've several doubts regarding this trigger.

* It was not asked in the public API and maybe we should prefer to implement it internally (in some internal namespace and without doc request), at least for now.
* In fact, we don't ever need a new trigger, we can just set an `on_replace` trigger on the `_schema` system space and look for tuples with the `'version'` value in the first field (plus check `box.internal.schema_needs_upgrade`).
* It seems, it re-enters into the Lua VM from the FFI call. AFAIK, it is improper FFI usage and may cause unpredicatable behavior.
* It seems, the trigger is called on downgrade too.
* It seems, it is called on a failed upgrade too.

I propose to look toward `box.space._schema:on_replace`.

changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
```diff
@@ -1,2 +1,6 @@
 ## feature/box
 * Added `on_schema_upgrade_end` trigger (gh-9849).
+
+## bugfix/box
+* User rights are now granted/revoked after upgrading automatically without
+  restarting (gh-9849).
```
Totktonada, line 6: Nit: The commit message has no 'fixes/closes #xxxx'.

Totktonada, line 6: Two headers within one changelog entry are not parsed correctly. Check out the `tools/gen-release-notes` output. You can run it locally or look on the `lint/release-notes` CI job.

Also, we usually separate the header from the body with a blank line (the tool understands it anyway, but it is for readability).

src/box/lua/config/applier/credentials.lua
```diff
@@ -653,6 +653,12 @@ local function sync_privileges(credentials, obj_to_sync)
     end)
 end
 
+box.ctl.on_schema_upgrade_end(function()
+    if not config then return end
+
+    sync_privileges(get_credentials(config))
+end)
```
Totktonada, line 660: It is a partial solution. We don't create users and roles, don't set passwords after an upgrade.

AFAIU, the right solution would be replacing the RW checks (`not box.info.ro`) with some 'DDL is enabled' checks (RW + not needs upgrade) and replacing RW waiting with 'DDL is enabled' waiting. We can use `on_replace` on `_schema` for the latter.

Also, ignoring of `box.error.SCHEMA_NEEDS_UPGRADE` is not needed with a full solution of the problem.

src/box/lua/ctl.c
georgiy-belyanin: Applied an approach with the `box.space._schema:on_replace` call.

changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
georgiy-belyanin: Thank you. Fixed it.

georgiy-belyanin: Added a blank line.
Removed the second entry because the trigger had been deleted.

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: Thank you. Now it's clearer how to fix it.
Applied the suggested way to fix the problem with `on_replace`.
Removed ignoring `SCHEMA_NEEDS_UPGRADE`

changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
georgiy-belyanin: Applied! Thank you!

changelogs/unreleased/gh-9849-grant-revoke-user-right-after-upgrade.md
Totktonada: Irrelevant now.

test/config-luatest/upgrade_test.lua
```diff
@@ -71,14 +71,13 @@ end)
...
-            table.insert(messages, alert.message)
-        end
-        local exp = 'box.schema.user.grant("guest", "read", "space", ' ..
-            '"_space") has failed because either the object has not been ' ..
-            'created yet, a database schema upgrade has not been performed, ' ..
-            'or the privilege write has failed (separate alert reported)'
-        t.assert_items_include(messages, {exp})
+
+        -- We will change 'guest' rights, so need to change user
+        box.session.su('admin')
```
Totktonada, line 76: `box.session.su` is not needed anymore (after changing the implementation to one that performs DDL from the background fiber).

test/config-luatest/upgrade_test.lua
```diff
@@ -71,14 +71,13 @@ end)
...
+        box.schema.upgrade()
+        -- After upgrade privileges should be granted automatically
+
+        local exp2 = { 'read', 'space', '_space' }
+        t.assert_items_include(box.schema.user.info('guest'), {exp2})
```
Totktonada, line 81: I would like if we check the following scenarios too:

* A user/a role is added.
* A password of a user is changed.

test/config-luatest/upgrade_test.lua
```diff
@@ -71,14 +71,13 @@ end)
...
     g.server:exec(function()
         t.assert_equals(box.space._schema:get("version"), {'version', 2, 11, 0})
-        local messages = {}
-        for _, alert in pairs(require("config"):info().alerts) do
-            table.insert(messages, alert.message)
-        end
-        local exp = 'box.schema.user.grant("guest", "read", "space", ' ..
-            '"_space") has failed because either the object has not been ' ..
-            'created yet, a database schema upgrade has not been performed, ' ..
-            'or the privilege write has failed (separate alert reported)'
```
Totktonada, line 81: Shouldn't we have some alert regarding this situation? At least some common, like 'Processing of the "credentials" section from config is postponed till database schema upgrade'.

src/box/lua/config/applier/credentials.lua
```diff
@@ -653,6 +653,33 @@ local function sync_privileges(credentials, obj_to_sync)
...
+    end
+
+    local latest_version = mkversion.get_latest()
+    local new_version = mkversion.from_tuple(new)
+
+    if new_version == latest_version then
+        box.on_commit(function()
+            update_schema_upgraded_status(new_version)
+        end)
+    end
```
Totktonada, line 680: Just curious: what is the behavior after `box.schema.downgrade()`?

I think that we can fix upgrading within this patch and postpone a work on downgrading if it requires an extra work, not just a couple of fixups.

I propose to experiment with downgrading a bit and dedice then.

src/box/lua/config/applier/credentials.lua
```diff
@@ -819,49 +896,21 @@ local wait_sync
...
+                wait_sync:put(err)
+            else
                 wait_sync:put('Done')
             end
-        else
-            local credentials = get_credentials(config)
-
-            box.atomic(sync_privileges, credentials, obj_to_sync)
+        elseif err then
+            error(err)
```
Totktonada, line 913: Should we really terminate the fiber in this case? Maybe it is better to log an error and pick up the next task?

Totktonada (approved):
I see no major flaws visually and it seems a good solution. @Serpentian, I would appreciate if you also glance on it.

@georgiy-belyanin I've left several remarks, please, take a look.

src/box/lua/config/applier/credentials.lua
```diff
@@ -809,6 +836,56 @@ end
...
+                'mode. Waiting for the read-write mode to set up the ' ..
+                'credentials given in the configuration. If another ' ..
+                'instance in the replicaset will write these new ' ..
+                'credentials, this instance will receive them and will ' ..
+                'skip further actions.')
+
+            box.ctl.wait_rw()
+        end
+
+        if not schema_is_upgraded then
```
Serpentian, line 854: Why do we need to always check it in a loop, on every new task? `schema_is_upgraded` can happen only after instance restart. It seems we should block `sync_credentials_worker` execution right after its start until the schema is upgraded.

Serpentian (commented):
Thank you for the patch! It's great we finally fix this issue! Check out my comments below)

test/config-luatest/upgrade_test.lua
georgiy-belyanin: Oh. I see. Thank you. Applied.

src/box/lua/config/applier/credentials.lua
```diff
@@ -819,49 +896,21 @@ local wait_sync
...
-                create_roles(credentials.roles)
-                create_users(credentials.users)
-
-                sync_privileges(credentials)
-            end)
-
-            if obj_to_sync.type == 'BLOCKING_FULL_SYNC' then
+        if obj_to_sync.type == 'BLOCKING_FULL_SYNC' then
+            if err then
+                wait_sync:put(err)
```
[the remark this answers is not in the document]
georgiy-belyanin: Seems a lot nicer. Thanks. Applied.

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: You are right. That's not the best way to deal with the worker thread. Fixed.

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: That'll be a lot cleaner. Thank you. Did it this way.

test/config-luatest/upgrade_test.lua
georgiy-belyanin: There used to be a verbose message about it. I made it a warning and added a test.

src/box/lua/config/applier/credentials.lua
```diff
@@ -809,6 +836,56 @@ end
...
+
+            box.ctl.wait_rw()
+        end
+
+        if not schema_is_upgraded then
+            log.verbose('credentials: the schema has an old version ' ..
+                'and user/roles can not be loaded. Consider executing ' ..
+                'box.schema.upgrade() to perform an upgrade.')
+
+            schema_is_upgraded_cond:wait()
```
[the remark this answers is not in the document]
georgiy-belyanin: Sounds good. Thank you. Added.

src/box/lua/config/applier/credentials.lua
```diff
@@ -653,6 +653,33 @@ local function sync_privileges(credentials, obj_to_sync)
     end)
 end
 
+local schema_is_upgraded = false
+local schema_is_upgraded_cond
+
+local function update_schema_upgraded_status(version)
+    local schema_version = version or mkversion.get()
+    schema_is_upgraded = schema_version == mkversion.get_latest()
```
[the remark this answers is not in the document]
georgiy-belyanin: Actually, no. Because `schema_needs_upgrade()` is still true, when the latest version schema is loaded and `on_replace` is fired. So I had to create a new variable. Probably this behaviour should be changed?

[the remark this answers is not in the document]
Serpentian: I suppose, it'll be way to difficult to change. I'm ok with new variable here

test/config-luatest/upgrade_test.lua
Totktonada: Nice! If it is not a big deal, please, also drop this warning when the upgrade occurs and extend the test with a check that the warning actually disappears.

It seems that we can do it here:

```lua
if not schema_is_upgraded then
    config._aboard:set({<...>})
    schema_is_upgraded_cond:wait()
    assert(schema_is_upgraded)
    -- <..here..>
end
```

test/config-luatest/upgrade_test.lua
georgiy-belyanin: Done all of the suggested! Thanks for the note.

test/config-luatest/upgrade_test.lua
georgiy-belyanin: Haven't noticed the fact the warnings had to be dropped. Thank you!

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: Right now after downgrading there is no new behaviour. 

Restarting the tarantool after downgrading results with the message similar to this:
```
2024-08-12 16:39:00.462 [16653] main/118/sync_credentials/tarantool.config log.lua:130 W> credentials: the schema has an old version and user/roles can not be loaded. Consider executing box.schema.upgrade() to perform an upgrade.
```

src/box/lua/config/applier/credentials.lua
```diff
@@ -819,49 +889,34 @@ local wait_sync
...
-            register_objects(credentials.users)
+        schema_is_upgraded_cond:wait()
+        assert(schema_is_upgraded)
 
-            box.atomic(function()
-                create_roles(credentials.roles)
-                create_users(credentials.users)
+        config._aboard:drop_if(function(_key, alert)
+            return alert._trait == 'not_upgraded_schema'
+        end)
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied. Thank you. Also I've added a test checking the alert dropping.

src/box/lua/config/applier/credentials.lua
Totktonada: I guess that the following scenario will result in failure:

* A normal (upgraded) instance is run.
* `box.schema.downgrade()` (plus `box.snapshot()`) is called.
* Something changed in the `credentials` section of the config: a new user is added, a role is added, a password is changed or a set of priviliges is changed.
* `config:reload()` is called.

I guess that it results in `check_errors` status in `box.info.config` and an error regarding non-upgraded schema in `box.info.config.alerts`.

I propose to elaborate a solution as part of a separate issue. I suggest to create the issue and mention it in the commit message here to make it clear what is solved by this commit and what doesn't.

src/box/lua/config/applier/credentials.lua
```diff
@@ -819,49 +889,35 @@ local wait_sync
...
-                ['space'] = {},
-                ['function'] = {},
-                ['sequence'] = {},
-            }
+        config._aboard:set({
+            type = 'warn',
+            message = 'credentials: the schema has an old version ' ..
+                      'and user/roles can not be loaded. Consider executing ' ..
+                      'box.schema.upgrade() to perform an upgrade.'
+        }, {key = not_upgraded_alert_key})
```
Totktonada, line 902: Nits regarding the message:

* I would say 'database schema', because just 'schema' seems too common. We have YAML configuration schema, for example.
* I wouldn't say 'loaded', because actually they can't be written into the database. 'Loading' sounds more like when something is read.
* 'user/roles' -> 'users/roles'?
* Should we say 'users/roles/privileges'? Technically speaking a privilege is a separate piece of data stored in the `_priv` system space. If a reader knows it, just 'users/roles' may not include privileges for he/she.

Totktonada:
@georgiy-belyanin I've glanced over the changes once again and they're OK for me in regards to the upgrading process. Please, finish the review with @Serpentian.

I suspect that a problem may appear after downgrading and reloading a configuration with a new content in the `credentials` section (without restart). The details are in [this](https://github.com/tarantool/tarantool/pull/10248#discussion_r1702086208) discussion thread.

However, I propose to glance on downgrading separately and verify/solve it out of this pull request.

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: Thank you for writing these concerns here. I'll create the issue right after I'll be able to reproduce and get these error in order to be sure how the problem looks like.

src/box/lua/config/applier/credentials.lua
```diff
@@ -895,13 +952,20 @@ local function apply(config_module)
...
+
     -- Set trigger on after space/function/sequence creation.
     if not triggers_are_set then
         triggers_are_set = true
+
         box.space._user:on_replace(on_replace_trigger)
         box.space._space:on_replace(on_replace_trigger)
         box.space._func:on_replace(on_replace_trigger)
         box.space._sequence:on_replace(on_replace_trigger)
+        box.space._schema:on_replace(credentials_upgrade_on_replace)
```
Serpentian, line 968: Nit: we don't need to set that trigger, if `schema_is_upgraded = true`

Serpentian (approved):
Thank you for the patch! Looks great

src/box/lua/config/applier/credentials.lua
georgiy-belyanin: Done! Added an extra variable to check if the trigger is set and set it only if the trigger is unset and the schema wasn't upgraded. Thank you!

src/box/lua/config/applier/credentials.lua
Totktonada: For the record: #10411.

Totktonada:
Pushed to `master` -- a future `3.2.0` release.

Backported to `release/3.1` (a future `3.1.2` release) in PR #10414.

PR #10251, opened 2024-07-15
Consider raft and limbo states system rows in snapshot

Serpentian:


Serpentian:
Related: https://github.com/tarantool/tarantool/issues/10307. Closing for now

PR #10287, opened 2024-07-22
connpool: make checking connections parallel

### LINKED ISSUE #10286: connpool: speedup the execution of the `filter()`
Connpool `filter()` and `call()` methods check connections to the existing instances. These checks are performed sequentially. Thus the maximum evaluation time is `number of instances * timeout` in case all of the instances are not reachable.

Evaluating the checks in parallel limits the time with a `timeout` bound. This improvement should be implemented.

Also tracked in TNTP-222

georgiy-belyanin:
Connpool `filter()` method used to be executed synchroniously in a single fiber. Thus the maximum evaluation time was `number of instances * timeout`. Make its execution parallel limiting the evaluation time with `timeout`.

Closes #10286

### DIFF
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -2,6 +2,7 @@ local fiber = require('fiber')
 local clock = require('clock')
 local config = require('config')
 local checks = require('checks')
+local fun = require('fun')
 local netbox = require('net.box')
 
 local WATCHER_DELAY = 0.1
@@ -21,7 +22,12 @@ local function is_connection_valid(conn, opts)
     return true
 end
 
+local connection_mode_update_cond = nil
 local function connect(instance_name, opts)
+    if not connection_mode_update_cond then
+        connection_mode_update_cond = fiber.cond()
+    end
+
     checks('string', {
         connect_timeout = '?number',
         wait_connected = '?boolean',
@@ -58,6 +64,7 @@ local function connect(instance_name, opts)
         conn.mode = mode
         local function watch_status(_key, value)
             conn._mode = value.is_ro and 'ro' or 'rw'
+            connection_mode_update_cond:broadcast()
         end
         conn:watch('box.status', watch_status)
     end
@@ -71,24 +78,49 @@ local function connect(instance_name, opts)
     return conn
 end
 
+local function is_candidate_connected(candidate)
+    local conn = connections[candidate]
+    return conn and conn.state == 'active' and conn:mode() ~= nil
+end
+
+-- Checks whether the candidate has responded with success or
+-- with an error.
+local function is_candidate_checked(candidate)
+    local conn = connections[candidate]
+
+    return not conn or
+           is_candidate_connected(candidate) or
+           conn.state == 'error' or
+           conn.state == 'closed'
+end
+
 local function connect_to_candidates(candidates)
     local delay = WATCHER_DELAY
-    local conn_opts = {connect_timeout = WATCHER_TIMEOUT}
-    local connected_candidates = {}
+    local connect_deadline = clock.monotonic() + WATCHER_TIMEOUT
+
     for _, instance_name in pairs(candidates) do
-        local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
-        local ok, conn = pcall(connect, instance_name, conn_opts)
-        -- If state is not 'active', conn:mode() cannot become not nil.
-        if ok and conn.state == 'active' then
-            while conn:mode() == nil and clock.monotonic() < time_connect_end do
-                fiber.sleep(delay)
-            end
-            if conn:mode() ~= nil then
-                table.insert(connected_candidates, instance_name)
-            else
-                conn:close()
-            end
+        pcall(connect, instance_name, {
+            wait_connected = false,
+            connect_timeout = WATCHER_TIMEOUT
+        })
+    end
+
+    assert(connection_mode_update_cond ~= nil)
+
+    local connected_candidates = {}
+    while clock.monotonic() < connect_deadline do
+        connected_candidates = fun.iter(candidates)
+            :filter(is_candidate_connected)
+            :totable()
+
+        local all_checked = fun.iter(candidates)
+            :all(is_candidate_checked)
+
+        if all_checked then
+            return connected_candidates
         end
+
+        connection_mode_update_cond:wait(delay)
     end
     return connected_candidates
 end
--- a/test/config-luatest/rpc_test.lua
+++ b/test/config-luatest/rpc_test.lua
@@ -2,6 +2,7 @@ local t = require('luatest')
 local fun = require('fun')
 local treegen = require('luatest.treegen')
 local server = require('luatest.server')
+local socket = require('socket')
 local helpers = require('test.config-luatest.helpers')
 
 local g = helpers.group()
@@ -13,6 +14,24 @@ end
 
 local function skip_if_no_vshard()
     t.skip_if(not has_vshard, 'Module "vshard-ee/vshard" is not available')
+end
+
+local function start_stub_servers(g, dir, instances)
+    g.stub_servers = g.stub_servers or {}
+    for _, instance in ipairs(instances) do
+        local uri = ('%s/%s.iproto'):format(dir, instance)
+        local s = socket.tcp_server('unix/', uri, function()
+            require('fiber').sleep(11000)
+        end)
+        t.assert(s)
+        g.stub_servers[instance] = s
+    end
+end
+
+local function stop_stub_servers(g)
+    for _, server in pairs(g.stub_servers) do
+        server:close()
+    end
 end
 
 g.test_connect = function(g)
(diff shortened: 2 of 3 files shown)


src/box/lua/connpool.lua
```diff
@@ -75,21 +75,34 @@ local function connect_to_candidates(candidates)
...
+    local connection_fibers = {}
+
     for _, instance_name in pairs(candidates) do
-        local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
-        local ok, conn = pcall(connect, instance_name, conn_opts)
-        -- If state is not 'active', conn:mode() cannot become not nil.
-        if ok and conn.state == 'active' then
-            while conn:mode() == nil and clock.monotonic() < time_connect_end do
-                fiber.sleep(delay)
+        local connection_fiber = fiber.new(function()
```
Totktonada, line 82: I wouldn't create a new fiber if the connection is alredy in a good state.

georgiy-belyanin: That's a good idea. Done.

src/box/lua/connpool.lua
```diff
@@ -75,21 +75,45 @@ local function connect_to_candidates(candidates)
     local delay = WATCHER_DELAY
     local conn_opts = {connect_timeout = WATCHER_TIMEOUT}
     local connected_candidates = {}
+
+    local connection_fibers = {}
+
     for _, instance_name in pairs(candidates) do
-        local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
         local ok, conn = pcall(connect, instance_name, conn_opts)
```
Totktonada, line 82: Hm. `net_box.connect()` plus `:wait_connected()` still called synchronously. This code parallelize the RO/RW mode waiting, but a network problem likely affect connecting too.

src/box/lua/connpool.lua
georgiy-belyanin: UPD: Now, not creating a new fibers at all.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you. Fixed. Now, it uses an approach with `on_connect` hook.

src/box/lua/connpool.lua
Totktonada: Irrelevant now, closing.

src/box/lua/connpool.lua
Totktonada: I like the direction of your idea. However, it should be implemented in a different way. Flaws in the current implementation:

1. Connections with `mode == nil` now added into the list of connected candidates.
2. `:wait_connected()` in a loop may wait too long and unblock the fiber after the deadline.

I propose to add a fiber cond to signal about the `mode ~= nil` event from `connect` and spin in a loop of `cond:wait(remaining_time)`. The exit conditions for the loop are either:

1. The time quota is exceeded.
2. All the waiting connections have `mode ~= nil`.

At least, it works in my head this evening. Need to check, of course :)

georgiy-belyanin: Yep. You're absolutely right. It should be implemented in a different way. Applied the discussed way.

Now there is a conditional variable that is triggered when some of connection modes are changed. We start connecting and wait for the moment when all of the instances are available until the timeout.

Actually, the perfect implementation would be like this (Lua-like pseudocode):
```lua
-- Creating connections:
function create_conn()
  -- ...
  conn = establish_conn()
  conn:on_update_mode(function() cond:broadcast() end)
  -- Currently the implementation lacks this.
  conn:on_error(function cond:broadcast() end)
  -- ...
end
-- Connecting to a candidates:
function connect_to_candidates()
  connect_to_all()
  while not the timeout reached do
    -- We may wait more than currently specified small delay.
    cond:wait(timeout - clock())
  end
end
```

The problem is in the absence of `on_error()` or `on_access()` connection triggers. Probably some of them should be implemented in order to fire the conditional variable when there is an error.

Totktonada: `:wait_state()` is in the [public API](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#conn-wait-state) and it returns on reaching a terminal state.

https://github.com/tarantool/tarantool/blob/90552e55e0921405c43ea086ae418a72c9f000e4/src/box/lua/net_box.lua#L1025-L1035

Not exactly what you're asking for, but may be useful.

georgiy-belyanin: Thank you. It seems useful, but not for that case. Because even using this the checks of the error will still be made once in some quite a small time interval. Is the current implementation considered ok?

test/config-luatest/rpc_test.lua
```diff
@@ -419,6 +419,91 @@ g.test_filter_mode = function(g)
...
+            instances:
+              instance-001:
+                database:
+                  mode: rw
+          replicaset-002:
+            instances:
+              instance-002:
+                iproto:
+                  listen:
+                    - uri: 1.0.0.0:8888
```
Totktonada, line 456: It doesn't lead to a long delay if the network interface is down. I would try to imitate it using `socket.tcp_server()`.

test/config-luatest/rpc_test.lua
```diff
@@ -419,6 +419,91 @@ g.test_filter_mode = function(g)
...
+        -- If the connection pool tries to connect in parallel, the
+        -- execution time is bounded with CONNECT_TIMEOUT plus some
+        -- small overhead.
+        --
+        -- Make sure we're trying to access the instances in parallel.
+        local opts = { mode = 'ro' }
+        local timestamp_before_filter = clock.monotonic()
+        t.assert_equals(connpool.filter(opts), {})
+        t.assert(clock.monotonic() - timestamp_before_filter <
+                 CONNECT_TIMEOUT * 3 / 2)
```
Totktonada, line 501: Nit: It is better to use `assert_lt`: it reports the both values in case of a failure.

test/config-luatest/rpc_test.lua
```diff
@@ -419,6 +419,91 @@ g.test_filter_mode = function(g)
...
+        local CONNECT_TIMEOUT = 10
+
+        -- If the connection pool tries to connect in parallel, the
+        -- execution time is bounded with CONNECT_TIMEOUT plus some
+        -- small overhead.
+        --
+        -- Make sure we're trying to access the instances in parallel.
+        local opts = { mode = 'ro' }
+        local timestamp_before_filter = clock.monotonic()
+        t.assert_equals(connpool.filter(opts), {})
```
Totktonada, line 499: I think it would be good to have a similar test for `call` too -- at least the task mentions that the same speedup effect should occur for `call` too :)

test/config-luatest/rpc_test.lua
```diff
@@ -419,6 +419,91 @@ g.test_filter_mode = function(g)
     g.server_4:exec(check)
 end
 
+g.test_filter_works_in_parallel = function(g)
```
Totktonada, line 422: AFAIU, the test case works ~10 seconds. That's generally bad (but understandable here). I think that it worth to move such long test cases into its own file to be sure that the default test timeout limit (110 seconds) is not reached even in a slow environment.

OTOH, other test cases are likely fast and unlikely there is a real problem with it. I'm in a little doubt, so just noted it. Feel free to ignore.

src/box/lua/connpool.lua
Totktonada: Sorry for the long delay! The implementation looks OK.

Maybe it worth to re-collect connected instances after exiting the loop by the timeout. OTOH, that's nigligible.

Totktonada (approved):
Thanks for working on it! Seems OK with a few remarks.

Sorry for the long delay.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Thank you. Made the test use it.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Done. Seems like everything works better now.

src/box/lua/connpool.lua
```diff
@@ -71,24 +78,49 @@ local function connect(instance_name, opts)
     return conn
 end
 
+local function is_candidate_connected(candidate)
+    local conn = connections[candidate]
+    return conn and conn.state == 'active' and conn:mode() ~= nil
+end
+
+-- Checks whether the candidate instance is not in some of the
+-- connecting/fetching state.
```
grafin, line 87: nit: for me this comment was not as clear as the function name and body itself.

src/box/lua/connpool.lua
```diff
@@ -71,24 +78,49 @@ local function connect(instance_name, opts)
...
+           is_candidate_connected(candidate) or
+           conn.state == 'error' or
+           conn.state == 'closed'
+end
+
 local function connect_to_candidates(candidates)
     local delay = WATCHER_DELAY
-    local conn_opts = {connect_timeout = WATCHER_TIMEOUT}
-    local connected_candidates = {}
+    local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
```
grafin, line 99: nit: i like a name like `deadline` better

test/config-luatest/rpc_test.lua
grafin: I like your idea with `connect_timeout` option. Can you create an issue for that, so we don't loose it.

grafin (approved):
Thanks for the patch! Just several nits from me, feel free to ignore.

test/config-luatest/rpc_test.lua
```diff
@@ -15,6 +16,24 @@ local function skip_if_no_vshard()
...
+            s:detach()
+        end)
+        t.assert(s)
+        g.stub_servers[instance] = s
+    end
+end
+
+local function stop_stub_servers(g)
+    for _, server in pairs(g.stub_servers) do
+        server:close()
```
georgiy-belyanin, line 33: Found out this may error if the socket is already closed. I should check if the socket is open before closing it.

test/config-luatest/rpc_test.lua
```diff
@@ -15,6 +16,24 @@ local function skip_if_no_vshard()
     t.skip_if(not has_vshard, 'Module "vshard-ee/vshard" is not available')
 end
 
+local function start_stub_servers(g, dir, instances)
+    g.stub_servers = g.stub_servers or {}
+    for _, instance in ipairs(instances) do
+        local uri = ('%s/%s.iproto'):format(dir, instance)
+        local s = socket.tcp_server('unix/', uri, function(s)
+            s:detach()
```
Totktonada, line 24: This way `:close()` in `stop_stub_servers` will not close the file descriptor, I guess. I think that we can just sleep the fiber here.

test/config-luatest/rpc_test.lua
```diff
@@ -419,6 +438,114 @@ g.test_filter_mode = function(g)
...
+        t.assert_lt(elapsed_time, CONNECT_TIMEOUT * 3 / 2)
+
+        -- call() also internally uses filter(). Ensure it tries
+        -- to access instances in parallel.
+        opts = { mode = 'prefer_ro' }
+        timestamp_before_filter = clock.monotonic()
+        t.assert_equals(connpool.call('f1', nil, opts), 'instance-001')
+        elapsed_time = clock.monotonic() - timestamp_before_filter
+        t.assert_gt(elapsed_time, CONNECT_TIMEOUT / 2)
+        t.assert_lt(elapsed_time, CONNECT_TIMEOUT * 3 / 2)
```
Totktonada, line 540: Should we separate the filter/call test cases (see the idea [here](https://github.com/tarantool/tarantool/pull/10438#discussion_r1774266075))?

src/box/lua/connpool.lua
georgiy-belyanin: Done. Thank you. For some reason I've chosen some strange name.

src/box/lua/connpool.lua
georgiy-belyanin: Fixed the comment to a bit more clearer (I suppose). Thanks.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Done. Thank you.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Fixed by removing the socket "detach".

src/box/lua/connpool.lua
Totktonada: Seems fixed now.

PR #10288, opened 2024-07-22
perf: introduce the box benchmarks

mkostoevr:
These are used to prevent regressions in the box and simplify its performance analysis:
- memtx_tree.lua
- memtx_tree_rv.lua

In order to check the benchmark results a new compare_ci.py tool is introduced: it uses the student's t-test in order to check if the difference between measurements is significant.

NO_DOC=perf test
NO_TEST=perf test
NO_CHANGELOG=perf test

### DIFF
--- a/perf/lua/CMakeLists.txt
+++ b/perf/lua/CMakeLists.txt
@@ -57,6 +57,12 @@ create_perf_lua_test(NAME uri_escape_unescape)
 
 include_directories(${MSGPUCK_INCLUDE_DIRS})
 
+build_module(benchmark_box_module benchmark_box_module.cc)
+target_link_libraries(benchmark_box_module msgpuck)
+create_perf_lua_test(NAME memtx_tree
+                     DEPENDS benchmark_box_module
+)
+
 build_module(column_scan_module column_scan_module.c)
 target_link_libraries(column_scan_module msgpuck)
 create_perf_lua_test(NAME column_scan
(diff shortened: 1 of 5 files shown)


perf/tools/compare_ci.py
mkostoevr, line 1: It would be nice to merge this with the existing `compare.lua`.

perf/lua/memtx_tree.lua
mkostoevr, line 1: The `memtx_tree_rv.lua` and this one duplicate a lot of code, would be nice to merge them.

perf/tools/compare_ci.py
ligurio: It is better to bring `compare.py` [^1] from Google Benchmark repository than reinvent  a wheel.
I've added it in https://github.com/tarantool/tarantool/pull/10365

[^1]: https://github.com/google/benchmark/tree/main/tools

perf/lua/CMakeLists.txt
```diff
@@ -57,6 +57,12 @@ create_perf_lua_test(NAME uri_escape_unescape)
 
 include_directories(${MSGPUCK_INCLUDE_DIRS})
 
+build_module(benchmark_box_module benchmark_box_module.cc)
+target_link_libraries(benchmark_box_module msgpuck)
+create_perf_lua_test(NAME memtx_tree
```
ligurio, line 62: Just to let you know:
It is desired to make sure the new benchmark is stable enough before merging it to the master branch.
`perf_micro` workflow executes benchmarks on regular runners if PR is based on a branch created in a fork repository, and executes on a special hardware machine if PR is based on a branch created in a main repository.
You can push the same branch to the main repo and create janitorial PR for review results in https://mon.tarantool.io/.

Issue #10089, opened 2024-06-03
relay: raft state is never sent during META_JOIN stage


Serpentian:
Relay never sends raft state during META_JOIN:

Relay encodes raft and writes it the to xstream:

https://github.com/tarantool/tarantool/blob/e0c9b6527f7897007028358adef963d542526d80/src/box/relay.cc#L500-L501

Encoding makes row LOCAL:

https://github.com/tarantool/tarantool/blob/e0c9b6527f7897007028358adef963d542526d80/src/box/xrow.c#L1390

Stream function `relay_send_initial_join_row`skips local rows:

https://github.com/tarantool/tarantool/blob/e0c9b6527f7897007028358adef963d542526d80/src/box/relay.cc#L1165-L1175

```
Tarantool 3.2.0-entrypoint-56-g05d03a1c58
Target: Linux-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/var/empty/local -DENABLE_BACKTRACE=TRUE
Compiler: GNU-12.3.0
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/serpentian/Programming/tnt/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror -g -ggdb -O0 
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/serpentian/Programming/tnt/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror -g -ggdb -O0 

```

**Steps to reproduce**

Apply the following patch:

```diff
diff --git a/src/box/applier.cc b/src/box/applier.cc
index 7369d40674..c92a2c4e55 100644
--- a/src/box/applier.cc
+++ b/src/box/applier.cc
@@ -804,6 +804,7 @@ applier_wait_snapshot(struct applier *applier)
 				if (txn_limbo_process(&txn_limbo, &req) != 0)
 					diag_raise();
 			} else if (iproto_type_is_raft_request(row.type)) {
[...]

PR #10293, opened 2024-07-22
Fix raft state never being sent

Serpentian:
Closes https://github.com/tarantool/tarantool/issues/10089
Closes https://github.com/tarantool/tarantool/issues/10727

### DIFF
--- a/changelogs/unreleased/gh-10089-raft-meta-join.md
+++ b/changelogs/unreleased/gh-10089-raft-meta-join.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a bug when the Raft state wasn't sent during the `META_JOIN` stage,
+  which could lead to split-brain (gh-10089).
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -1126,6 +1126,29 @@ relay_send(struct relay *relay, struct xrow_header *packet)
 	xrow_stream_write(stream, packet);
 }
 
+void
+relay_filter_raft(struct xrow_header *packet, uint32_t version)
+{
+	assert(iproto_type_is_raft_request(packet->type));
+	if (version > version_id(3, 2, 1) ||
+	    (version > version_id(2, 11, 4) && version < version_id(3, 0, 0)))
+		return;
+	/**
+	 * Until Tarantool 3.2.2 all raft requests were sent with GROUP_LOCAL
+	 * id. In order not to break the upgrade process, raft rows are still
+	 * sent as local to old replicas. This was also backported to 2.11.5.
+	 */
+	packet->group_id = GROUP_LOCAL;
+}
+
+static void
+relay_send_raft(struct relay *relay, struct xrow_header *packet)
+{
+	assert(iproto_type_is_raft_request(packet->type));
+	relay_filter_raft(packet, relay->version_id);
+	relay_send(relay, packet);
+}
+
 static void
 relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 {
@@ -1152,7 +1175,7 @@ relay_raft_msg_push(struct cmsg *base)
 	struct xrow_header row;
 	RegionGuard region_guard(&fiber()->gc);
 	xrow_encode_raft(&row, &fiber()->gc, &msg->req);
-	relay_send(relay, &row);
+	relay_send_raft(relay, &row);
 	if (relay_flush(relay) < 0) {
 		relay_set_error(relay, diag_last_error(diag_get()));
 		fiber_cancel(fiber());
--- a/src/box/relay.cc
+++ b/src/box/relay.cc
@@ -1153,6 +1153,8 @@ static void
 relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 {
 	struct relay *relay = container_of(stream, struct relay, stream);
+	if (iproto_type_is_raft_request(row->type))
+		return relay_send_raft(relay, row);
 	/*
 	 * Ignore replica local requests as we don't need to promote
 	 * vclock while sending a snapshot.
--- a/src/box/xrow.c
+++ b/src/box/xrow.c
@@ -1442,7 +1442,7 @@ xrow_encode_raft(struct xrow_header *row, struct region *region,
 	memset(row, 0, sizeof(*row));
 	row->type = IPROTO_RAFT;
 	row->body[0].iov_base = buf;
-	row->group_id = GROUP_LOCAL;
+	row->group_id = r->group_id;
 	row->bodycnt = 1;
 	const char *begin = buf;
 
@@ -1478,13 +1478,14 @@ xrow_decode_raft(const struct xrow_header *row, struct raft_request *r,
 {
 	if (row->type != IPROTO_RAFT)
 		goto bad_msgpack;
-	if (row->bodycnt != 1 || row->group_id != GROUP_LOCAL) {
+	if (row->bodycnt != 1) {
 		diag_set(ClientError, ER_INVALID_MSGPACK,
 			 "malformed raft request");
 		return -1;
 	}
 	memset(r, 0, sizeof(*r));
 
+	r->group_id = row->group_id;
 	const char *pos = row->body[0].iov_base;
 	uint32_t map_size = mp_decode_map(&pos);
 	for (uint32_t i = 0; i < map_size; ++i)
--- a/test/replication-luatest/gh_10727_no_local_rows_test.lua
+++ b/test/replication-luatest/gh_10727_no_local_rows_test.lua
@@ -1,0 +1,52 @@
+local t = require('luatest')
+local server = require('luatest.server')
+local proxy = require('luatest.replica_proxy')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.master = server:new{alias = 'master'}
+    cg.master:start()
+    cg.master:exec(function()
+        box.schema.space.create('test', {is_local = true}):create_index('pk')
+        box.space.test:insert{1, 'data'}
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.master:stop()
+end)
+
+local found_local_row = false
+local function process_replication_flow(conn, data)
+    local ok, h = pcall(box.iproto.decode_packet, data)
+    if ok and h.group_id == 1 then
+        found_local_row = true
+    end
+
+    conn:forward_to_client(data)
+end
+
+g.test_no_local_rows_are_sent = function(cg)
+    local proxy_uri = server.build_listen_uri('master_proxy')
+    local replica = server:new{alias = 'replica', box_cfg = {
+        replication = { proxy_uri },
+    }}
+    local server_proxy = proxy:new({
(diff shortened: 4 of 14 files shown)


changelogs/unreleased/gh-10089-raft-meta-join.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a bug when raft state wasn't sent during META_JOIN stage, which could
+  lead to split-brain (gh-10089).
```
andreyaksenov, line 4:
```suggestion
* Fixed a bug when the Raft state wasn't sent during the `META_JOIN` stage, which could
  lead to split-brain (gh-10089).
```

andreyaksenov (approved):

test/replication-luatest/join_test.lua
```diff
@@ -64,3 +64,46 @@ g.test_fetch_self_delete_during_final_join = function(lg)
         box.space._cluster:delete{replica_id}
     end, {replica_id})
 end
+
+--
+-- gh-10089: raft state is never sent during META_JOIN stage.
```
Gerold103, line 69: `is` -> `was`. To make the emphasis that now it is not so, and that is good.

Gerold103 (approved):
Good work on spotting and swiftly fixing the bug 💪.

src/box/xrow.c
```diff
@@ -1522,6 +1529,13 @@ xrow_decode_raft_local(const struct xrow_header *row, struct raft_request *r,
 	return xrow_decode_raft_impl(row, r, vclock, GROUP_LOCAL);
 }
 
+int
+xrow_decode_raft(const struct xrow_header *row, struct raft_request *r,
+		 struct vclock *vclock)
+{
+	return xrow_decode_raft_impl(row, r, vclock, GROUP_DEFAULT);
+}
```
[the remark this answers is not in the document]
Serpentian: > I would suggest to have just one xrow_decode_raft which can work with any group (which won't even check it). That would be a bit simpler.

I intentionally did that in order to fortify the code execution. I suppose, this way the code will be more strict and we'll think twice, whether the row, we're reading, is local or remote, since RAFT can be both. This way we won't repeat this mistake in the future.

I'd rather leave as it is. Let's see what @sergepetrenko think here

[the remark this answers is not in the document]
sergepetrenko: I think you could put the `group_id` into `raft_request`/`raft_msg` structs and make a single version of `xrow_encode/decode_raft` which handles the group id.

That's because we already have `raft_checkpoin_local` and `raft_checkpoint_remote`, which may set group id as well.

changelogs/unreleased/gh-10089-raft-meta-join.md
Serpentian: Fixed

test/replication-luatest/join_test.lua
Serpentian: Fixed

src/box/xrow.c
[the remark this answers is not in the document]
Serpentian: > I think you could put the group_id into raft_request/raft_msg structs and make a single version of xrow_encode/decode_raft which handles the group id.

Great idea! Done that. So much cleaner now

Serpentian:
@Gerold103, since the patch changed too majorly, I'm asking you to review it one more time)

src/lib/raft/raft.h
```diff
@@ -124,6 +124,12 @@ struct raft_msg {
 	 * Also is omitted when does not matter (when the message is for disk).
 	 */
 	const struct vclock *vclock;
+	/**
+	 * Whether the message is intended to be used locally or sent
+	 * to another replica. Used to figure out group_id of the xrow
+	 * (either GROUP_DEFAULT or GROUP_LOCAL).
+	 */
+	bool is_local;
```
Gerold103, line 132: Wait wait, lib/raft shouldn't care about this. It only produces messages and checkpoints. Group ID is a thing made up for our replication, unrelated to the basic Raft. Lets move this all into box/raft. It is built on top of lib/raft and can deal with all the Tarantool replication specifics.

I know it seems convenient now, but lib/raft doesn't use this flag, only sets it. Which means we leak the logic from an upper module into the lower module, and that would hit us eventually if more things come into lib/raft like this, in terms of code complexity and dependencies.

Gerold103 (commented):
Thanks for the fixes, it does look cleaner indeed! However we introduce into lib/raft some alien logic now 😰.

src/lib/raft/raft.h
Serpentian: Thank you for noticing! Didn't think about that. Now it's even simpler, but I had to drop the assertions from the `raft_process_recovery`, since now, when raft state is actually received, the message has `state` and may have `vclock`.

Gerold103 (approved):
Thanks for addressing all my comments 🙏!

src/box/xrow.c
```diff
@@ -1478,13 +1478,14 @@ xrow_decode_raft(const struct xrow_header *row, struct raft_request *r,
 {
 	if (row->type != IPROTO_RAFT)
 		goto bad_msgpack;
-	if (row->bodycnt != 1 || row->group_id != GROUP_LOCAL) {
+	if (row->bodycnt != 1) {
```
locker, line 1481: After this change, a replica running an older binary won't be able to subscribe to a master running a patched binary. This would break even patch-level upgrades.

src/box/relay.cc
```diff
@@ -1119,6 +1119,14 @@ static void
 relay_send(struct relay *relay, struct xrow_header *packet)
 {
 	ERROR_INJECT_YIELD(ERRINJ_RELAY_SEND_DELAY);
+	/**
```
Serpentian, line 1122: This is the least crutchy solution I managed to find. The solution for initial https://github.com/tarantool/tarantool/issues/10089 is easy, we can just patch `relay_send_initial_join_row` to make it send raft requests, but I don't want to send `GROUP_LOCAL` rows at all, this seems incorrect ideologically.

It would be perfect, if sending raft state from `box_process_subscribe` could be done via `relay_send`, but I didn't manage to make it work, seems like it's there for the reason.

Serpentian:
@Gerold103, since @locker found a major bug in the prev iteration, it was rewritten again, I need your review)

src/box/relay.cc
locker: The label `2.11` to #10089 suggests that it should be backported to 2.11 so we need to take it into account when checking `version_id`. Also, since the version check is used in two places, I'd put it in a helper function.

locker (approved):

test/replication-luatest/gh_10727_no_local_rows_test.lua
```diff
@@ -0,0 +1,53 @@
...
+end)
+
+g.after_all(function(cg)
+    cg.master:stop()
+end)
+
+local function process_replication_flow(conn, data)
+    local ok, h = pcall(box.iproto.decode_packet, data)
+    if ok and h.group_id == 1 then
+        -- This is the only way, I managed to find to stop the test, when
```
Gerold103, line 24: Nit: lets avoid pronouns in the comments like "I", "me", "my", etc. The code authorship is going to be not quite easily deductible after it is merged into master.

src/box/relay.cc
```diff
@@ -1119,6 +1119,14 @@ static void
...
+	 * Until Tarantool 3.2.1 all raft requests were sent with
+	 * GROUP_LOCAL id. In order not to break the upgrade process,
+	 * we still send local rows to old replicas.
+	 */
+	if (iproto_type_is_raft_request(packet->type) &&
+	    relay->version_id < version_id(3, 2, 1))
+		packet->group_id = GROUP_LOCAL;
```
Gerold103, line 1129: Hm. Having it in such a generic place feels quite crutchy, like you said. Perhaps extract Raft row sending into a new function called `relay_send_raft()` containing this check and invoke it explicitly in `relay_send_tx()` (when request type is Raft) and in `relay_raft_msg_push()` unconditionally?

Then we could at least isolate the hack.

test/replication-luatest/gh_10727_no_local_rows_test.lua
```diff
@@ -0,0 +1,53 @@
...
+        -- wrong result is found. Everything is pcalled there.
```
Gerold103, line 25: You could perhaps make this function a closure and wakeup the main test fiber when the condition is found. The test case would then terminate the proxy and do all the proper shutdown like a regular test.

Gerold103 (commented):
Thanks for fixing this, and shame on me that I didn't notice the compatibility problem 😰.

test/replication-luatest/gh_10727_no_local_rows_test.lua
Serpentian: Or we can use a global variable) Even simpler

src/box/relay.cc
Serpentian: WDYT of the current implementation? `relay_send_tx` never sends raft, all local rows are filtered out from it. You meant `relay_send_initial_join_row`, I suppose

test/replication-luatest/gh_10727_no_local_rows_test.lua
Serpentian: The comment doesn't exist anymore. I got you, for the future

src/box/relay.cc
Gerold103: Aah, yep, then in `relay_send_initial_join_row()`. The rest of the implementation is fine! I can't think of any other way.

Serpentian: @Gerold103, please, approve to PR, if we're good to go)

src/box/relay.cc
```diff
@@ -1155,6 +1155,8 @@ static void
 relay_send_initial_join_row(struct xstream *stream, struct xrow_header *row)
 {
 	struct relay *relay = container_of(stream, struct relay, stream);
+	if (iproto_type_is_raft_request(row->type))
+		return relay_send_raft(relay, row);
```
Gerold103, line 1159: Please, lets turn the `if (is raft request)` inside `relay_send_raft()` into an assertion. Otherwise it looks quite strange. When a request is not Raft, the function called "send raft" just ignores it. As a result, either the caller already knows it is a Raft request and uses this function, or has to check the type before calling the function. Either way the `if` inside of the function never fails. Which means it is an assertion. No?

Gerold103 (commented):
Sorry, just one last nit.

src/box/relay.cc
Serpentian: Agree, fixed

Gerold103 (approved):
Prime top 💪!

src/box/xrow.c
sergepetrenko: Resolving since the issue is addressed.

src/box/relay.cc
sergepetrenko: Resolving since the issue is addressed

sergepetrenko (approved):
Thanks for the patch!
LGTM.

PR #10297, opened 2024-07-23
refactor: use Lua C API instead of G(L)->gc.total

### LINKED ISSUE #10284: Fix usages of `G(L)->gc.total` in Tarantool sources
We use `G(L)->gc.total` instead of using the Lua C API in several places:
* https://github.com/tarantool/tarantool/blob/a3ef8fb60382e747eed2cb521ebb3329dd4058f7/src/box/lua/info.c#L478
* https://github.com/tarantool/tarantool/blob/a3ef8fb60382e747eed2cb521ebb3329dd4058f7/src/box/lua/slab.cc#L262

These usages can be replaced with the corresponding calls to [the `lua_gc()` function](https://www.lua.org/manual/5.1/manual.html#lua_gc):
```c
lua_gc(L, LUA_GCCOUNT, 0) * 1024 + lua_gc(L, LUA_GCCOUNTB, 0);
```

mandesero:
Replaced G(L)->gc.total calls to lua_gc().

To ensure better encapsulation, maintainability, and portability of
the code, it is necessary to replace direct access to the fields of
global structures with calls using the Lua C API.

Closes https://github.com/tarantool/tarantool/issues/10284

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring

### DIFF
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -475,7 +475,7 @@ lbox_info_memory_call(struct lua_State *L)
 	lua_settable(L, -3);
 
 	lua_pushstring(L, "lua");
-	lua_pushinteger(L, G(L)->gc.total);
+	lua_pushinteger(L, luaL_getgctotal(L));
 	lua_settable(L, -3);
 
 	return 1;
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -382,6 +382,12 @@ luaL_setcdatagc(struct lua_State *L, int idx)
 	lua_pop(L, 1);
 }
 
+size_t
+luaL_getgctotal(struct lua_State *L)
+{
+	return (lua_getgccount(L) * 1024ULL) + lua_gc(L, LUA_GCCOUNTB, 0);
+}
+
 /**
  * A helper to register a single type metatable.
  */
--- a/src/lua/utils.h
+++ b/src/lua/utils.h
@@ -219,6 +219,13 @@ LUA_API void
 luaL_setcdatagc(struct lua_State *L, int idx);
 
 /**
+ * @brief Return size of currently allocated memory.
+ * @param L Lua State
+ */
+size_t
+luaL_getgctotal(struct lua_State *L);
+
+/**
 * @brief Return CTypeID (FFI) of given СDATA type
 * @param L Lua State
 * @param ctypename С type name as string (e.g. "struct request" or "uint32_t")
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -259,7 +259,7 @@ lbox_runtime_info(struct lua_State *L)
 	 * Lua GC heap size
 	 */
 	lua_pushstring(L, "lua");
-	lua_pushinteger(L, G(L)->gc.total);
+	lua_pushinteger(L, luaL_getgctotal(L));
 	lua_settable(L, -3);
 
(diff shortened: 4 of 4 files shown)


src/box/lua/info.c
```diff
@@ -475,7 +475,8 @@ lbox_info_memory_call(struct lua_State *L)
 	lua_settable(L, -3);
 
 	lua_pushstring(L, "lua");
-	lua_pushinteger(L, G(L)->gc.total);
+	lua_pushinteger(
+		L, (lua_getgccount(L) << 10) + lua_gc(L, LUA_GCCOUNTB, 0));
```
ligurio, line 479: I would reformat like this:

```diff
--- a/src/box/lua/info.c
+++ b/src/box/lua/info.c
@@ -475,8 +475,8 @@ lbox_info_memory_call(struct lua_State *L)
        lua_settable(L, -3);
 
        lua_pushstring(L, "lua");
-       lua_pushinteger(
-               L, (lua_getgccount(L) << 10) + lua_gc(L, LUA_GCCOUNTB, 0));
+       int countb = (lua_getgccount(L) << 10) + lua_gc(L, LUA_GCCOUNTB, 0);
+       lua_pushinteger(L, countb);
        lua_settable(L, -3);
 
        return 1;
```

the same below

ligurio (commented):
Maksim, thanks for the patch!

1. Please fix full name in commit (`git commit --amend --author="Maksim Tyushev  <mandesero@gmail.com>"`) and set full name in Git settings for a future commits.
2. Replace a reason `refactor` to `refactoring`, because `refactor` is a verb, not a noun.
3. Describe in commit message a reason why we should replace `G(L)->gc.total` to `lua_gc`.

src/box/lua/info.c
Buristan: Also, I suggest declaring a new function: `luaT_getgccount(L)` in <src/lua/utils.c> and using it. It is also useful for #10223.

src/lua/utils.c
```diff
@@ -382,6 +382,12 @@ luaL_setcdatagc(struct lua_State *L, int idx)
 	lua_pop(L, 1);
 }
 
+int
+luaL_getgctotal(struct lua_State *L)
+{
+	return (lua_getgccount(L) << 10) + lua_gc(L, LUA_GCCOUNTB, 0);
```
Buristan, line 388: Minor: Maybe it is better to use `* 1024ULL` here, since the `lua_gc()` returns an `int` result that may overflow (theoretically for the GC64 build and 2 TB of Lua memory :)).
Also, I suppose that function should have the `size_t` returned value type, since it has the corresponding semantics.

src/lua/utils.h
```diff
@@ -218,6 +218,13 @@ luaL_checkcdata(struct lua_State *L, int idx, uint32_t *ctypeid);
 LUA_API void
 luaL_setcdatagc(struct lua_State *L, int idx);
 
+/**
+ * @brief Return size of currently allocated memory.
+ * @param L Lua State
+ */
+LUA_API int
```
Buristan, line 225: There is not a public function, so please don't use LUA_API here.

mandesero: Corrected

Buristan (approved):

ligurio (approved):
Maksim, thanks for the patch!

PR #10300, opened 2024-07-23
vinyl: skip bad vylog records in `force_recovery` mode

### LINKED ISSUE #10292: `ER_INVALID_VYLOG_FILE` error isn't ignored in `force_recovery` mode
We've encountered quite a few bugs when recovery failed because of a logically invalid vylog record: #10277, #10128, #5436, #4066, #3416. Bugs like those are quite severe because there's no easy for the user to recover data. We should let the user recover from a broken vylog by setting the `force_recovery` flag. At worst, they'll lose a chunk of data but it's still better than loosing all the data. BTW in all the bugs enumerated above, ignoring the invalid vylog records wouldn't result in any data loss - at worst, a garbage run file on disk would never be removed.

locker:
We've had a number of issues when Tarantool was permanently broken (unable to recover after restart) because of a bad vylog record. The `force_recovery` mode didn't help so the user would have no other choice but to rebootstrap. A funny thing is those bugs were usually caused by a race between the garbage collector and dump/compaction when a vylog record was written for a dropped index. The worst thing that could happen if we ignored such a bad record is an unused run file not deleted from disk. Apparently, this is better than a permanent recovery failure so let's support the `force_recovery` mode in vylog.

The tricky part here is handling checkpoint after restart. The problem is that to create a vylog checkpoint, we load the previous vylog file so we have to ignore errors if it was loaded in the `force_recovery` mode.

Closes #10292


locker:
Test failures are unrelated - they are caused by the recent test-run update:

```
Duration of failed tests (seconds):
*   9.39 replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua 
```

src/box/vy_log.c
```diff
@@ -1396,18 +1328,9 @@ void
 vy_log_write(const struct vy_log_record *record)
 {
 	say_verbose("write vylog record: %s", vy_log_record_str(record));
-
-	if (vy_log.tx_failed)
-		return;
-
 	assert(vy_log.tx != NULL);
 	struct vy_log_record *tx_record = vy_log_record_dup(&vy_log.tx->region,
```
mkostoevr, line 1352: I would be nice to have it named `vy_log_record_xdup` or something to show it never fails, but I couldn't find anything similar in the vinyl and `xvy_log_record_dup` would look ugly.

The same note about `vy_recovery_alloc_key_parts` (e. g. `vy_recovery_xalloc_key_parts`).

mkostoevr (approved):
LGTM, the best patches are the ones decreasing LoC.

andreyaksenov (approved):

src/box/vy_log.c
locker: We don't use the `x` prefix with non-failing functions that allocate various objects, for example, see `key_def_dup`. The ultimate goal is to make all `alloc` and `dup` functions non-failing (perhaps, for a few rare exceptions, like memtx tuple allocations) so the `x` prefix would be pointless. The only reason why we added the `x` prefix to generic allocation functions, like `mempool_alloc` or `region_alloc`, is that we have a lot of legacy code handling errors returned by those functions, and the `x` prefix allows for a gradual transition.

locker:
The failure of the `release_asan_clang` workflow is unrelated:

```
Duration of failed tests (seconds):
* 310.08 app-luatest/http_client_test.lua
```

locker:
Cherry-picked to 3.1 and 2.11.

PR #10302, opened 2024-07-23
box: filter out tmp and remote txs when aborting all txs for DDL

### LINKED ISSUE #9720: Temporary spaces abort all DDL even without MVCC
The reproducer requires 2 instances:
```Lua
--
-- Instance 1
--
-- Step 1
--
box.cfg{
    listen = 3301,
    replication = {3301, 3302},
}
box.once('grant guest', function() 
    box.schema.user.grant('guest', 'read,write,execute', 'universe')
end)

--
-- Step 3
--
format = {{name = 'id', type = 'unsigned'}}
space = box.schema.create_space('test', {format = format})
_ = space:create_index('pk')
do
    box.begin()
    for i = 1, 1000 do
        space:replace{i, i}
    end
    box.commit()
end

--
-- Step 5
--
table.insert(format, {name = 'val', type = 'unsigned'})
space:format(format)
```
and
```Lua
--
-- Instance 2
--
-- Step 2
--
box.cfg{
    listen = 3302,
    replication = {3301, 3302},
}
box.once('grant guest', function() 
    box.schema.user.grant('guest', 'read,write,execute', 'universe')
end)

--
-- Step 4
--
box.error.injection.set('ERRINJ_CHECK_FORMAT_DELAY', true)

--
-- Step 6
--
space = box.schema.space.create('TMP', { type = 'temporary' })
space:drop()
```

The result is that the step-6 will abort the long DDL txn in the applier, combing from the instance-1, even though those 2 DDL txns are not related anyhow. The temporary spaces don't even yield or store any actual data on disk.

This hits the users who are trying to use the temporary spaces as short living objects, creating and deleting them frequently.

The bug seems to be appearing since this commit: https://github.com/tarantool/tarantool/commit/8f4be3227635ef606581dda7e85f254a96347c0e.

CuriousGeorgiy:
This patch series filters out temporary and remote transactions when aborting all transactions for DDL, since they cannot conflict each other by definition.

Closes #9720


andreyaksenov (approved):

src/box/txn.h
```diff
@@ -885,6 +889,12 @@ txn_is_nop(const struct txn *txn)
 	return txn->n_new_rows + txn->n_applier_rows == 0;
 }
 
+static inline bool
+txn_is_remote(const struct txn *txn)
```
drewdzzz, line 893: Nit: maybe call it `txn_is_fully_remote`? (analogy with `txn_is_fully_local`).

src/box/memtx_tx.c
```diff
@@ -683,6 +683,17 @@ memtx_tx_acquire_ddl(struct txn *tx)
 	(void) txn_can_yield(tx, false);
 }
 
+/**
+ * Nop and remote transactions cannot possibly conflict each other by
+ * definition, so we can filter them out when aborting transactions for DDL.
+ */
+static bool
+memtx_tx_filter_nop_and_remote_txs(struct txn *txn1, struct txn *txn2)
```
drewdzzz, line 691: Actually, `nop` shouldn't conflict anything since it means no-op. But our `nop` is not no-op - it can be a temporary statement, and it's a bit confusing here. Let's, at least, call this function like `memtx_tx_filter_temporary_and_remote_txs` with elaboration in commentary so that it's clear that `nop` is not always a no-op.

drewdzzz: Also, cannot we use `txn_is_fully_local()` here instead of `txn_is_nop()`? Any temporary transaction is fully local, and it seems that any local transaction cannot conflict with purely remote one.

src/box/txn.c
```diff
@@ -653,6 +653,8 @@ txn_commit_stmt(struct txn *txn, struct request *request)
 		assert(stmt->row != NULL);
 		txn_update_row_counts(txn, stmt, 1);
 	}
+	if (stmt->space != NULL && space_is_data_temporary(stmt->space))
```
drewdzzz, line 656: Isn't it complement of the condition above? It could be just `else` clause.

drewdzzz (changes requested):

src/box/txn.h
```diff
@@ -525,6 +525,10 @@ struct txn {
 	 * Nesting level of space on_replace triggers for current txn.
 	 */
 	int space_on_replace_triggers_depth;
+	/**
+	 * Whether the transaction has temporary statements.
+	 */
+	bool has_temporary_stmts;
```
drewdzzz, line 531: Let's elaborate in description why we need the flag at all (rows of temporary statements are erased).

src/box/txn.h
CuriousGeorgiy: Okay, done.

src/box/txn.c
CuriousGeorgiy: No, the condition above has an `or`. We shouldn't set the flag if the statement doesn't have a space assigned to it.

CuriousGeorgiy: True, thanks!

src/box/memtx_tx.c
CuriousGeorgiy: > Any temporary transaction is fully local, and it seems that any local transaction cannot conflict with purely remote one.

A transaction on data-temporary spaces (for which the space metadata is replicated) can be conflicting, so we need `txn_is_nop` here.

src/box/txn.h
CuriousGeorgiy: Done.

src/box/memtx_tx.c
drewdzzz: Oh, thanks, I see. But `txn_is_nop` is not sufficient then. If a transaction did a replace in a data-temporary space, it is `nop` but it actually can conflict with remote one since it can do DDL on data-temporary spaces.

CuriousGeorgiy: No, AFAIC, it is enough. A transaction contaning DDL on data-temporary spaces would have `n_new_rows > 0`, since the DDL statement would get written to the WAL (see `txn_is_fully_local`).

drewdzzz: AFAICS, we [update](https://github.com/tarantool/tarantool/blob/e01c20edc31b65e020ac57e3f62a7427b6e27d53/src/box/txn.c#L650) row counts only for non-data-temporary statements.

CuriousGeorgiy: > So, we can have local DML transaction on data-temporary space, which is nop, and we can have transaction in applier doing DDL on this data-temporary space.

If we have a transaction in the applier doing DDL, it acquires DDL, so it is forbideen to yield — even with MVCC enabled:
https://github.com/tarantool/tarantool/blob/e01c20edc31b65e020ac57e3f62a7427b6e27d53/src/box/memtx_tx.c#L679-L684

Unless it is a space format change or a new index creation, for which yielding is safe by design.

drewdzzz: Consider the scenario:
```
TX1: BEGIN
TX1: DML on data-temporary space
APPLIER: DDL on data-temporary space
TX1: COMMIT
```
Here applier should abort TX1 since they are operating the same space, but it doesn't work in this way in your patch because TX1 is actually `nop` (it has no rows) and APPLIER is `fully_remote`.

CuriousGeorgiy: You are missing the point: the applier transaction cannot yield here, otherwise it will be aborted by yield.

src/box/txn.c
```diff
@@ -652,6 +652,8 @@ txn_commit_stmt(struct txn *txn, struct request *request)
 			goto fail;
 		assert(stmt->row != NULL);
 		txn_update_row_counts(txn, stmt, 1);
+	} else {
+		txn->has_temporary_stmts = true;
```
Gerold103, line 656: That unfortunately won't work well if the statement is rolled back by a savepoint-rollback. The flag would remain set but inaccurate. I would suggest to rather walk the statements in `txn_is_fully_remote()` to check if any of them are temporary. It is linear complexity, but is needed only for true DDL (i.e. not on temporary spaces), which can only be rare. Given that it aborts all concurrent txns (excluding applier and temporary spaces).

Also the flag isn't initialized now in `txn_begin`. Can as well become `true` randomly and start failing. And for flags anyway we use `enum txn_flag`. But I suggest not to add it at all really. It is deductible from the statements and is needed only on rollback (perf isn't important).

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
```diff
@@ -0,0 +1,166 @@
...
+    cg.server2:exec(function()
+        box.error.injection.set('ERRINJ_CHECK_FORMAT_DELAY', true)
+    end)
+    cg.server1:exec(function()
+        box.space.s:format{'i', 'unsigned'}
+    end)
+    cg.server2:exec(function()
+        box.schema.space.create('tmp', {type = 'temporary'})
+        box.error.injection.set('ERRINJ_CHECK_FORMAT_DELAY', false)
+    end)
```
Gerold103, line 61: Does it fail without the patch? It looks like the last exec is likely to be executed before the second one is replicated. So the format coming to `server2` won't even see `ERRINJ_CHECK_FORMAT_DELAY = true` and won't try to abort anything.

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
```diff
@@ -0,0 +1,166 @@
...
+    }
+    cg.replica_set:start()
+    cg.server1:exec(function()
+        local s = box.schema.create_space('s')
+        s:create_index('p')
+        box.begin()
+        for i = 1, 1000 do
+            s:replace{i, i}
+        end
+        box.commit()
```
Gerold103, line 34: Why do you need so many tuples in the space? Same below in `test_remote_ddl()`.

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
```diff
@@ -0,0 +1,166 @@
...
+    t.skip_if(not cg.params.memtx_use_mvcc)
+
+    cg.server2:exec(function()
+        local tmp = box.schema.space.create('tmp-replace', {type = 'temporary'})
+        tmp:create_index('p')
+
+        box.space.s:on_replace(function()
+            tmp:replace{0}
+            rawset(_G, 'executed_on_replace', true)
+            while true do require('fiber').yield() end
```
Gerold103, line 82: That is going to be a busy loop with almost 100% CPU usage. Lets better make it sleep(some big value) or `.stall()` (sleep until wakeup or cancel).

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
```diff
@@ -0,0 +1,166 @@
...
+    end)
+    cg.server2:exec(function()
+        t.helpers.retrying({timeout = 60}, function()
+            t.assert(_G.executed_on_replace)
+        end)
+        box.schema.space.create('tmp', {type = 'temporary'})
+    end)
+    local msg = "Transaction committing DDL (id=%d+) has aborted " ..
+                "another TX (id=%d+)"
+    cg.server2:grep_log(msg)
```
Gerold103, line 96: `grep_log` won't fail on its own, you need to check it with `t.assert`.

Gerold103 (commented):
Thanks, good patch 💪.

src/box/memtx_tx.c
CuriousGeorgiy: Sorry for my misconception, and thanks for the catch, fixed: added another flag to track data-temporary statements, and added a test case for this scenario.

src/box/txn.c
CuriousGeorgiy: > I would suggest to rather walk the statements in `txn_is_fully_remote()` to check if any of them are temporary.

I don't understand how you suggest to check if they are temporary. Temporary and nop statements are indistinguishable from this point in the code. Only if we add a `temporary` flag to `txn_stmt`.

> That unfortunately won't work well if the statement is rolled back by a savepoint-rollback. The flag would remain set but inaccurate.

Is this a common case? If not, I would suggest leaving it as is, since we are trying to fix a common source of pain for Picodata user workloads here.

> Also the flag isn't initialized now in `txn_begin`.

Fixed, thanks.

> And for flags anyway we use `enum txn_flag`.

Not always, for instance we have a `is_schema_changed` flag. I think this is a temporary solution, and it should also be just a separate flag.

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
CuriousGeorgiy: Because otherwise I won't get to the error injection:
https://github.com/tarantool/tarantool/blob/09bfdd473ffd23169dbe8d4e5d0510e8c70b11ec/src/box/memtx_space.c#L52-L65

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
CuriousGeorgiy: Yes, it fails. I see your point, but how do you suggest setting up a synchronization point here?

I tried to make the `_space` synchronous and wait for `box.info.synchro.queue.busy` on `server2`, but the replication connection starts to timeout then:
```console
2024-08-03 12:15:11.398 [42760] relay/unix/:(socket)/101/main coio.c:300 E> TimedOut: timed out
2024-08-03 12:15:11.398 [42760] relay/unix/:(socket)/101/main I> exiting the relay loop
```

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
CuriousGeorgiy: Thanks, fixed!

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
CuriousGeorgiy: `stall` is not public, AFAIC:
https://github.com/tarantool/tarantool/blob/09bfdd473ffd23169dbe8d4e5d0510e8c70b11ec/src/lua/fiber.lua#L75-L80

Thanks for the catch, changed it to `t.helpers.retrying`.

src/box/txn.c
CuriousGeorgiy: Fixed by applying your suggestion, thanks for the catch! Added test cases for savepoint-rollback.

src/box/txn.h
```diff
@@ -893,6 +893,38 @@ txn_is_nop(const struct txn *txn)
 	return txn->n_new_rows + txn->n_applier_rows == 0;
 }
 
+static inline bool
+txn_is_fully_temporary(struct txn *txn)
```
Gerold103, line 897: Nit: the txn can be `const` here, no? Same for `txn_is_fully_remote()`.

src/box/txn.h
```diff
@@ -893,6 +893,38 @@ txn_is_nop(const struct txn *txn)
 	return txn->n_new_rows + txn->n_applier_rows == 0;
 }
 
+static inline bool
```
Gerold103, line 896: Nit: the function is relatively big, I would say it doesn't have to be `inline`. Only increases the compilation time this way. Same for `txn_is_fully_remote()`.

src/box/txn.h
```diff
@@ -893,6 +893,38 @@ txn_is_nop(const struct txn *txn)
...
+txn_is_fully_temporary(struct txn *txn)
+{
+	if (!txn_is_nop(txn))
+		return false;
+	struct txn_stmt *stmt;
+	stailq_foreach_entry(stmt, &txn->stmts, next) {
+		if (stmt->space != NULL && !space_is_temporary(stmt->space) &&
+		    space_is_data_temporary(stmt->space))
```
Gerold103, line 904: Nit: hm, that looks weird. Perhaps it would be better to check here literally `space->opts.type == SPACE_TYPE_DATA_TEMPORARY`. Instead of these 2 checks.

src/box/txn.h
```diff
@@ -893,6 +893,38 @@ txn_is_nop(const struct txn *txn)
...
+{
+	if (txn->n_new_rows != 0)
+		return false;
+	struct txn_stmt *stmt;
+	stailq_foreach_entry(stmt, &txn->stmts, next) {
+		/*
+		 * Allow DDL on data-temporary spaces, since we allow only fully
+		 * temporary transactions to continue.
+		 */
+		if (stmt->row == NULL && stmt->space != NULL &&
```
Gerold103, line 921: Nit: the function is just a checker, looks strange that it mentions DDL. I guess this comments rather belongs to the place, where this function is used?

Gerold103 (approved):
Thanks for the patch, I like how all the corner cases are covered in the tests 💪.

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
Gerold103: Easy but not reliable way - just add here a `fiber.sleep(0.01)` to make it more likely to fail without the patch. Hard, but more correct way - introduce a errinj like `ERRINJ_WAL_DELAY_COUNTDOWN` or other `*_COUNTDOWN`s. An injection which gets set on Nth check. Such as `ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN`. And make it set `ERRINJ_CHECK_FORMAT_DELAY`. Then you can set countdown = 0, and wait until delay = true.

src/box/txn.h
CuriousGeorgiy: No, unfortunately, using the `stailq` API requires the original struct to be non-const:
```console
/Users/glebedev/Work/tarantool/src/box/txn.c:671:2: error: passing 'const struct stailq *' to parameter of type 'struct stailq *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]

```

src/box/txn.h
CuriousGeorgiy: Fixed.

src/box/txn.h
CuriousGeorgiy: This function determines the filtering condition. This comment explains why we take filter out only statements with `stmt->row == NULL`.

test/replication-luatest/gh_9720_tmp_spaces_ddl_abort_applier_txs_test.lua
CuriousGeorgiy: Thanks, added a `ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN` injection.

src/box/txn.h
drewdzzz: AFAIU, you expect that replace in a data-temporary space has `stmt->row == NULL` and DDL there has `stmt->row != NULL`. Well, any DDL is actually a replace in space `_space` which is not temporary at all, so it still would be allowed without this condition. Please, drop it.

drewdzzz (approved):
Thanks for the patch, LGTM!
Please, address my comment in the last Vlad's discussion before merging.

src/box/txn.h
CuriousGeorgiy: @drewdzzz Thanks, moved the `stmt->row == NULL` to an assert and moved the comment out of the loop.

src/box/memtx_space.c
```diff
@@ -1073,6 +1073,13 @@ memtx_space_check_format(struct space *space, struct tuple_format *format)
 
 		ERROR_INJECT_YIELD(ERRINJ_CHECK_FORMAT_DELAY);
 
+		ERROR_INJECT_COUNTDOWN(ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN, {
+			struct errinj *e =
+				errinj(ERRINJ_CHECK_FORMAT_DELAY, ERRINJ_BOOL);
+			e->bparam = true;
+			ERROR_INJECT_YIELD(ERRINJ_CHECK_FORMAT_DELAY);
+		});
```
Gerold103, line 1081: If you switch these 2 ERROR_INJECT calls, then you can drop the one that you have on line 1080.

src/box/vinyl.c
```diff
@@ -1150,6 +1150,12 @@ vinyl_space_check_format(struct space *space, struct tuple_format *format)
 		if (++loops % VY_YIELD_LOOPS == 0)
 			fiber_sleep(0);
 		ERROR_INJECT_YIELD(ERRINJ_CHECK_FORMAT_DELAY);
+		ERROR_INJECT_COUNTDOWN(ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN, {
+			struct errinj *e =
+				errinj(ERRINJ_CHECK_FORMAT_DELAY, ERRINJ_BOOL);
+			e->bparam = true;
+			ERROR_INJECT_YIELD(ERRINJ_CHECK_FORMAT_DELAY);
+		});
```
Gerold103, line 1158: Same here. Or do I miss why it wouldn't work?

Gerold103 (commented):
Thanks for the fixes!

src/box/memtx_space.c
CuriousGeorgiy: How will this work, if one wants to enable `ERRINJ_CHECK_FORMAT_DELAY` independently of `ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN`? For instance, if the latter is set to `-1` (default).

Also refer to `ERRINJ_WAL_DELAY{_COUNTDOWN}`?

src/box/vinyl.c
CuriousGeorgiy: `ERROR_INJECT_COUTNDOWN` executes the code only when it has the `0` value set. The default value is `-1`, so `ERRINJ_CHECK_FORMAT_DELAY` won’t work.

Or maybe I don’t understand what you mean by
> If you switch these 2 ERROR_INJECT calls

src/box/memtx_tx.c
Gerold103: Resolving since this is approved anyway.

src/box/memtx_space.c
Gerold103: I guess we will ignore this for now.

src/box/vinyl.c
Gerold103: I guess we will ignore this for now.

src/box/memtx_space.c
CuriousGeorgiy: Sorry, I forgot to push the fix.

src/box/vinyl.c
CuriousGeorgiy: Sorry, I forgot to push the fix.

Issue #9844, opened 2024-03-20
Protobuf encoder/decoder tasklist


Col-Waltz:
This is a tasklist for protobuf encoder/decoder module for gRPC protocol.

API for this module is described here https://www.notion.so/tarantool/RFC-Protocol-Buffers-encoder-decoder-17ff2ff599974f9a9d0f39d9ab0a5284?pvs=4

These tasks are arranged in which they will be done and may be separated in smaller one if needed. Feedback on the module development is welcome
```[tasklist]
### Tasks
- [ ] Four basic protobuf types encoder: https://github.com/tarantool/tarantool/pull/9488
- [ ] Repeated fields support for encoding
- [ ] Protobuf decoder
- [ ] .proto file parser
- [ ] Tarantool specific types (Timestamp, Userdata) supported
```

R-omk:
As part of the ongoing development of the protobuf ecosystem, I propose adding support for `optional` fields in proto3. Optional fields allow for more efficient data serialization and better backward compatibility, especially in use cases where field presence or absence is important but not always guaranteed.

example

```proto
syntax = "proto3";

message MessageFromJson {

    string k1 = 1;
    optional string k2 = 2;

}
```

fields that are not present in the serialized representation will have different representations in memory.

PR #9488, opened 2023-12-14
Introduce protobuf encoder module

Col-Waltz:
Introducing lua protobuf encoder
Encoder can create a new protocol and encode data according to it.
As a result a binary string is returned that can be transported
by wire or decoded back by another protobuf decoder.
The future versions will add support protocol creation from .proto
files and decode method for encoded data.

Part of #9844 

@TarantoolBot document
Title: Protobuf module
Product: Tarantool
Root document: -

Protobuf encoder API

Introducing protobuf encoder. All the Protocol Buffers wire types
are supported except group start/end, which are deprecated in proto3.
To encode data you need to create a protocol according to which
data will be encoded.

The two main components of the protocol are messages and enums.
To create them .message and .enum functions are used

Each message consists of name of the message and fields.
Each field has a name, a type and an id. Id must be unique
for all fields in one message. For example this is a message with
six fields:

```lua
protobuf.message('KeyValue', {
    key = {'bytes', 1},
    create_revision = {'int64', 2},
    mod_revision = {'int64', 3},
    version = {'int64', 4},
    value = {'bytes', 5},
    lease = {'int64', 6),
})
```

This implementation supports recursive definition for fields in
message. Depth of recursion is defined by input data because all
fields are optional by default. Example of recursive message:

```lua
protobuf.message('Node', {
    number = {'int64', 1},
    data = {'bytes', 2},
    next_node = {'Node', 3},
})
```

Each enum type consists of name of type and values.
Values must have a zero value to be set as default as in example:

```lua
protobuf.enum('EventType', {
    ['PUT'] = 0,
    ['DELETE'] = 1,
})
```

To create a protocol .protocol function is used. This function
supports forward declared types and nested messages so the tuple
can be set according to example:

```lua
schema = protobuf.protocol({
    protobuf.message(<...>),
    protobuf.message(<...>),
    protobuf.enum(<...>),
    protobuf.enum(<...>),
})
```

Output protocol can then be used for encoding entered data by the
method named encode. This method converts input data
according to chosen message definition from
protobuf protocol into protobuf wireformat. All fields in message
definition are optional so if some input data is missing
it simply will not be encoded. Input data can be
submitted using luatypes or using cdata (for example
entering int64) according to the example.

```lua
result = schema:encode(‘KeyValue’,
    {
        key = 'protocol',
        version = 2,
        lease = 5,
    }
)
```

Output result will be a binary string encoded according to the
protobuf standard and can be transmitted to another user.

PR #10303, opened 2024-07-23
lua: introduce protobuf decoder module

Col-Waltz:
Decoder can decode data according to existing protocol.
As a result a lua table with decoded values is returned.
This table can be used for encode method as an input
value.

Part of https://github.com/tarantool/tarantool/issues/9844

@TarantoolBot document
Title: Protobuf module decoder
Product: Tarantool
Root document: -

Protobuf decoder API

Introducing protobuf decoder. All the Protocol Buffers wire
types are supported except group start/end, which are
deprecated in proto3. To decode data you need to select a
protocol according to which data will be decoded.

Decoder takes a byte string which contains encoded protobuf
data and returns a lua table, which contains all existing in
binary values. Default values or not assigned field will be
absent in result table.

Here is an example of use:

```lua
result = protocol:decode('MessageName', binary_string)
```

If you need not assigned fields to have a value you can set
'set_default' flag in options, according to example:

```lua
result = protocol:decode('MessageName', binary_string,
    {set_default = true})
```

With `set_default` flag set the result table will have default
values for all fields. Scalar type fields or enums are assigned
with default values described here:
https://protobuf.dev/programming-guides/proto3/#default.
Default value for repeated field is an empty table and default
value for message is nil.

### DIFF
--- a/changelogs/unreleased/protobuf_decoder_module.md
+++ b/changelogs/unreleased/protobuf_decoder_module.md
@@ -1,0 +1,3 @@
+## feature/lua
+
+* Introduced a Lua implementation of the protobuf decoder (gh-9844).
--- a/src/lua/protobuf.lua
+++ b/src/lua/protobuf.lua
@@ -1,6 +1,7 @@
 local ffi = require('ffi')
 local wireformat = require('internal.protobuf.wireformat')
 local protocol_mt
+local ctx_mt
 -- These constants are used to define the boundaries of valid field ids.
 -- Described in more detail here:
 -- https://protobuf.dev/programming-guides/proto3/#assigning
@@ -32,14 +33,17 @@ local MAX_SINT32_LL = 2LL^31 - 1
 local MAX_SINT32_ULL = 2ULL^31 - 1
 local MAX_SINT64_ULL = 2ULL^63 - 1
 
-
 local int64_t = ffi.typeof('int64_t')
 local uint64_t = ffi.typeof('uint64_t')
 
 -- Forward declarations
 local encode
+local decode
+local decode_message
 local encode_field
+local decode_field
 local validate_scalar
+local validate_field_id
 
 local scalars = {}
 
@@ -52,7 +56,9 @@ scalars.float = {
     limits = {
         number = {-MAX_FLOAT, MAX_FLOAT},
     },
+    wire_type = wireformat.I32,
     encode = wireformat.encode_float,
+    decode = wireformat.decode_float,
 }
 
 scalars.fixed32 = {
@@ -64,7 +70,9 @@ scalars.fixed32 = {
         int64 = {0LL, MAX_UINT32_LL},
         uint64 = {nil, MAX_UINT32_ULL},
     },
+    wire_type = wireformat.I32,
     encode = wireformat.encode_fixed32,
+    decode = wireformat.decode_fixed32,
 }
 
 scalars.sfixed32 = {
@@ -76,14 +84,18 @@ scalars.sfixed32 = {
         int64 = {MIN_SINT32_LL, MAX_SINT32_LL},
         uint64 = {nil, MAX_SINT32_ULL},
     },
+    wire_type = wireformat.I32,
     encode = wireformat.encode_fixed32,
+    decode = wireformat.decode_sfixed32,
 }
 
 scalars.double = {
     accept_type = 'number',
     encode_as_packed = true,
     integral_only = false,
+    wire_type = wireformat.I64,
     encode = wireformat.encode_double,
+    decode = wireformat.decode_double,
 }
 
 scalars.fixed64 = {
@@ -94,7 +106,9 @@ scalars.fixed64 = {
         number = {0, MAX_UINT64},
         int64 = {0LL, nil},
     },
+    wire_type = wireformat.I64,
     encode = wireformat.encode_fixed64,
+    decode = wireformat.decode_fixed64,
 }
 
 scalars.sfixed64 = {
@@ -105,13 +119,17 @@ scalars.sfixed64 = {
         number = {MIN_INT64, MAX_INT64},
         uint64 = {nil, MAX_SINT64_ULL},
     },
+    wire_type = wireformat.I64,
     encode = wireformat.encode_fixed64,
+    decode = wireformat.decode_sfixed64,
 }
 
 scalars.string = {
     accept_type = 'string',
     encode_as_packed = false,
+    wire_type = wireformat.LEN,
     encode = wireformat.encode_len,
+    decode = wireformat.decode_len,
 }
 
 scalars.bytes = scalars.string
@@ -125,7 +143,9 @@ scalars.int32 = {
         int64 = {MIN_SINT32_LL, MAX_SINT32_LL},
         uint64 = {nil, MAX_SINT32_ULL},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_int,
+    decode = wireformat.decode_int,
 }
 
 scalars.sint32 = {
@@ -137,7 +157,9 @@ scalars.sint32 = {
         int64 = {MIN_SINT32_LL, MAX_SINT32_LL},
         uint64 = {nil, MAX_SINT32_ULL},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_sint,
+    decode = wireformat.decode_sint,
 }
 
 scalars.uint32 = {
@@ -149,7 +171,9 @@ scalars.uint32 = {
         int64 = {0LL, MAX_UINT32_LL},
         uint64 = {nil, MAX_UINT32_ULL},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_int,
+    decode = wireformat.decode_uint,
 }
 
 scalars.int64 = {
@@ -160,7 +184,9 @@ scalars.int64 = {
         number = {MIN_INT64, MAX_INT64},
         uint64 = {nil, MAX_SINT64_ULL},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_int,
+    decode = wireformat.decode_int,
 }
 
 scalars.sint64 = {
@@ -171,7 +197,9 @@ scalars.sint64 = {
         number = {MIN_INT64, MAX_INT64},
         uint64 = {nil, MAX_SINT64_ULL},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_sint,
+    decode = wireformat.decode_sint,
 }
 
 scalars.uint64 = {
@@ -182,13 +210,17 @@ scalars.uint64 = {
         number = {0, MAX_UINT64},
         int64 = {0LL, nil},
     },
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_int,
+    decode = wireformat.decode_uint,
 }
 
 scalars.bool = {
     accept_type = 'boolean',
     encode_as_packed = true,
+    wire_type = wireformat.VARINT,
     encode = wireformat.encode_int,
+    decode = wireformat.decode_bool,
 }
 
 -- }}} Scalar type definitions
@@ -217,16 +249,7 @@ local function message(message_name, message_def)
             error(('Id %d in field %q was already used'):format(field_id,
                 field_name))
         end
-        if field_id < MIN_FIELD_ID or field_id > MAX_FIELD_ID then
-            error(('Id %d in field %q is out of range [%d; %d]'):format(
-                field_id, field_name, MIN_FIELD_ID, MAX_FIELD_ID))
-        end
-        if field_id >= RESERVED_FIELD_ID_MIN and
-           field_id <= RESERVED_FIELD_ID_MAX then
-           error(('Id %d in field %q is in reserved ' ..
-               'id range [%d, %d]'):format(field_id, field_name,
-               RESERVED_FIELD_ID_MIN, RESERVED_FIELD_ID_MAX))
-        end
+        validate_field_id(field_id, field_name)
         local field_def = {
             type = field_type,
             name = field_name,
@@ -387,6 +410,22 @@ local function is_inf(value)
     return not is_nan(value) and is_nan(value - value)
 end
 
+local function add_path(ctx, next_step)
+    ctx.path = ctx.path .. '.' .. next_step
+end
+
+local function restore_path(ctx)
+    ctx.path = ctx.path:match("^(.-)%.[^%.]+$")
+end
+
+local function append(table1, table2)
+    table1 = table1 or {}
+    for _, value in ipairs(table2) do
+        table.insert(table1, value)
+    end
+    return table1
+end
+
 -- Checks input value assumed to be integer.
 --
 -- Checks 'number' type value to be integral and 'cdata' type value to be
@@ -410,6 +449,17 @@ local function remove_tag(value)
     until msb == 0
     return string.sub(value, tag_length + 1)
 end
+
+-- Function for handling errors on decode.
+--
+-- As input parameters function gets err_msg string which describes an error
+-- and format_args table which contains parameters for error message to format.
+-- local function handle_error(ctx, err_msg, ...)
+--    err_msg = err_msg:format(...)
+--    local pos_info = ('[%s] decoding error on byte %d: %s')
+--        :format(ctx.path, ctx.pos, err_msg)
+--    error(pos_info)
+--end
 
 -- }}} Global helpers
 
@@ -500,6 +550,19 @@ local function validate_type(field_def, value, exp_type)
     return
 end
 
+validate_field_id = function(field_id, field_name)
+    if field_id < MIN_FIELD_ID or field_id > MAX_FIELD_ID then
+        error(('Id %d in %q field is out of range [%d; %d]'):format(
+            field_id, field_name, MIN_FIELD_ID, MAX_FIELD_ID))
+    end
+    if field_id >= RESERVED_FIELD_ID_MIN and
+        field_id <= RESERVED_FIELD_ID_MAX then
+        error(('Id %d in %q field is in reserved ' ..
+            'id range [%d, %d]'):format(field_id, field_name,
+            RESERVED_FIELD_ID_MIN, RESERVED_FIELD_ID_MAX))
+    end
+end
+
 local function validate_range(field_def, value, range)
     local min = range ~= nil and range[1] or nil
     local max = range ~= nil and range[2] or nil
@@ -542,6 +605,29 @@ validate_scalar = function(field_def, value)
                 or 'uint64'
         end
         validate_range(field_def, value, scalar_def.limits[value_type])
+    end
+end
+
+local function validate_decoded_type(protocol, field_def, ctx, wire_type)
+    if wireformat.types[wire_type] == nil then
+        ctx:error('Decoding data contains incorrect protobuf type: %d',
+            wire_type)
+    end
+    local exp_wire_type
+    if scalars[field_def.type] ~= nil then
+        exp_wire_type = scalars[field_def.type]['wire_type']
+    elseif is_message(protocol, field_def) then
+        exp_wire_type = wireformat.LEN
+    else
+        exp_wire_type = wireformat.VARINT
+    end
+    if field_def.repeated and wire_type == wireformat.LEN.id then
+        return
+    end
+    if wire_type ~= exp_wire_type.id then
+        ctx:error('Value for field was encoded as %q and ' ..
+            'can`t be decoded as %q', wireformat.types[wire_type].name,
+             wireformat.types[exp_wire_type.id].name)
     end
 end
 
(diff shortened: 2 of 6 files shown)


changelogs/unreleased/protobuf_decoder_module.md
```diff
@@ -0,0 +1,3 @@
+## feature/lua
+
+* Introduced Lua implementation of protobuf decoder (gh-9844).
```
andreyaksenov, line 3:
```suggestion
* Introduced a Lua implementation of the protobuf decoder (gh-9844).
```

andreyaksenov (approved):

changelogs/unreleased/protobuf_decoder_module.md
Col-Waltz: Updated changelog according to comment.

R-omk:
Please make sure that for different schemes this protomessage  gives different decoding results

base64 proto mesasge:



```base64
CgJ2MRgF
```

----

schema and result  **with optional** fields
```proto
syntax = "proto3";

message MessageFromJson {

    string s = 1;
    optional    string s2 = 2;
    int32 i = 3;
}
```

```json
{
  "s": "v1",
  "i": 5
}
```

----

schema and result  **with NO optional** fields

```proto
syntax = "proto3";

message MessageFromJson {

    string s = 1;
    string s2 = 2;
    int32 i = 3;
}
```

```json
{
  "s": "v1",
  "s2": "",
  "i": 5
}
```

Totktonada:
@R-omk Thank you for the feedback!

The plan is to skip a support of the `optional` keyword in the initial version of the encoder and the decoder and look for a need.

Since you highlighted it, I guess, it matters in your scenario. I would like to solve it, but separately from this PR. Would you mind to file an issue?

Totktonada:
Looked on the data.

```yaml
tarantool> digest.base64_decode('CgJ2MRgF'):hex()
---
- 0a0276311805
...
```

```c
0a 02 7631   18 05
^^ ^^ ^^^^   ^^ ^^
|| ||  ||    || 5
|| || "v1"   VARINT (0), field 3
|| of length 2
LEN (2), field 1
```

It means that the field 2 (`s2`) is missed in the data. I agree that in the explicit presence (when the `optional` keyword is used) it shouldn't present in the result.

When there is no `optional` keyword, it depends on the `set_default` flag.

changelogs/unreleased/protobuf_decoder_module.md
Totktonada, line 1: Regarding the commit message and the pull request description (please, fix in both places where it is applicable).

> Introducing lua protobuf decoder.

I've missed it in PR #9488, but we usually add a subsystem to the commit subject. `lua:` fits good here.

Don't end the subject with a period.

> Title: Protobuf module

There is an issue with this title: https://github.com/tarantool/doc/issues/4379. Let's add the 'decoder' word to it.

> Default value for repeated field is an empty map

I guess that you meant an empty *table*.

src/lua/protobuf_wireformat.lua
```diff
@@ -97,6 +141,29 @@ local function encode_varint(value)
     return ffi.string(buf, size)
 end
 
+-- Decode input value as VARINT without a tag.
+--
+-- Input value type: string
```
Totktonada, line 146: It looks like the function accepts the decoding context, not a string. There is a string in `ctx.val`, but `val.pos` is also the input parameter.

Some functions says that the input value is `number` and it is even more confusing.

What is the idea behind this remark? Does it add useful information for a reader?

src/lua/protobuf_wireformat.lua
```diff
@@ -106,6 +173,21 @@ local function encode_tag(field_id, wire_type)
...
+    return bit.band(tag, 0x7), bit.rshift(tag, 3)
+end
+
+-- Return default value for input type.
+local function set_default(type)
+    assert(type ~= 'message')
+    if type == 'bool' then return BOOL_DEFAULT
+    elseif type == 'string' or type == 'bytes' then return STRING_DEFAULT
+    else return NUMERIC_DEFAULT
+    end
```
Totktonada, line 188: Unusual code formatting, let's use the following.

```lua
if <cond> then
    <statement>
elseif <cond> then
    <statement>
else
    <statement>
end
```

src/lua/protobuf_wireformat.lua
```diff
@@ -106,6 +173,21 @@ local function encode_tag(field_id, wire_type)
     return encode_varint(bit.bor(bit.lshift(field_id, 3), wire_type))
 end
 
+-- Decode a tag byte.
+local function decode_tag(ctx)
```
Totktonada, line 177: The module is splitted to the helpers and the public API sections. Seems like the second section contains the function that are exposed by the module.

The `decode_tag` and `set_default` functions are placed into the helpers section, but exposed by the module.

It feels confusing.

src/lua/protobuf.lua
```diff
@@ -500,6 +523,19 @@ local function validate_type(field_def, value, exp_type)
...
+        error(('Id %d is out of range [%d; %d]'):format(
+            field_id, MIN_FIELD_ID, MAX_FIELD_ID))
+    end
+    if field_id >= RESERVED_FIELD_ID_MIN and
+        field_id <= RESERVED_FIELD_ID_MAX then
+        error(('Id %d is in reserved ' ..
+            'id range [%d, %d]'):format(field_id,
+            RESERVED_FIELD_ID_MIN, RESERVED_FIELD_ID_MAX))
+    end
+end
```
Totktonada, line 537: A very similar code is in `protobuf.message()`. Can we reuse the function there?

src/lua/protobuf_wireformat.lua
```diff
@@ -106,6 +173,21 @@ local function encode_tag(field_id, wire_type)
...
+    local tag = decode_int(ctx)
+    return bit.band(tag, 0x7), bit.rshift(tag, 3)
+end
+
+-- Return default value for input type.
+local function set_default(type)
```
Totktonada, line 183: This function don't set a default to anywhere, it just returns it, so the name seems confusing. It could be `get_default`, I guess.

src/lua/protobuf_wireformat.lua
```diff
@@ -97,6 +141,29 @@ local function encode_varint(value)
...
+-- Input value type: string
+--
+-- This is a helper function to decode tag and data values.
+local function decode_varint(ctx)
+    local msb
+    local res = 0
+    local len = 0
+    repeat
+        local payload = ffi.cast('uint64_t', string.byte(ctx.val, ctx.pos))
+        if len >= 10 or payload == nil then
```
Totktonada, line 155: How `payload` can be `nil`?

Totktonada, line 155: I would define 10 as a constant and clarify it somehow.

src/lua/protobuf_wireformat.lua
```diff
@@ -97,6 +141,29 @@ local function encode_varint(value)
...
+--
+-- Input value type: string
+--
+-- This is a helper function to decode tag and data values.
+local function decode_varint(ctx)
+    local msb
+    local res = 0
+    local len = 0
+    repeat
+        local payload = ffi.cast('uint64_t', string.byte(ctx.val, ctx.pos))
```
Totktonada, line 154: It seems, here we can read data out of the buffer boundary. Let's add a boundary check. Consider introducing functions to read data from the current position of the decoding context: read one byte as a number, read 4 bytes as cdata<uint32_t> and so on -- anything that you really need. These functions should perform the boundary checks. (And also this way it is easier to test different buffer implemetations in a future to choose the most performant.)

src/lua/protobuf_wireformat.lua
```diff
@@ -97,6 +141,29 @@ local function encode_varint(value)
...
+        if len >= 10 or payload == nil then
+            error('Incorrect msb for varint to decode')
```
Totktonada, line 156: Imagine that you have a 100 KiB data and got this error on decoding. Your actions?

Sometimes things are even more complicated: you have no the message, just the error that the decoder reported.

The message now says not more than 'something is bad', at least for me. Can we at least say 'a varint exceeds 10 bytes is size' -- this is alredy something.

Next, we have the decoding context and we definitely can report the position, where the decoding fails. I suggest to consider introducing a function that raises decoding error with extending the given error message with extra infornation from the decoding context.

I guess we can even save a path to the current field (like `foo.bar.baz`) to the decoding context and report it in the error.

src/lua/protobuf.lua
```diff
@@ -657,9 +726,142 @@ end
...
+        if field_def.repeated and is_scalar(field_def) then
+                if type == WIRE_TYPE_LEN and field_def.type ~= 'string'
+                    and field_def.type ~= 'bytes' then
+                        decode_from_packed = true
+                end
+            end
+        ctx.ignore_repeated = false
+        ctx.packed = decode_from_packed
+            res[field_def.name] = decode_field(protocol, field_def, ctx)
+        end
```
Totktonada, line 839: Something is definitely bad with indent here.

src/lua/protobuf.lua
```diff
@@ -657,9 +726,142 @@ end
...
+        local value
+        while ctx.pos < start_pos + len do
+            ctx.ignore_repeated = true
+            value = decode_field(protocol, field_def, ctx)
+            table.insert(res, value)
+        end
+    else
+        repeat
+            local field_id
+        ctx.ignore_repeated = true
```
Totktonada, line 762: First, the indent is broken here.

Next, I'm in doubt about passing the flag using the context this way. The idea is set the flag to go to a different branch in the next `decode_field` call and only in it. Neither keep it in processing of nested fields, nor in processing of next fields. However, the flag is unset in some not a very obvious place in another function.

I think that things would be much simpler if we just pass it to the function as a separate argument, like it is done in `encode_field`. This way it doesn't interfere with recursion anyhow and works just for the next call -- exactly what we need here.

src/lua/protobuf.lua
```diff
@@ -32,13 +32,22 @@ local MAX_SINT32_LL = 2LL^31 - 1
 local MAX_SINT32_ULL = 2ULL^31 - 1
 local MAX_SINT64_ULL = 2ULL^63 - 1
 
+-- Acceptable protobuf types
+local WIRE_TYPE_VARINT = 0
+local WIRE_TYPE_I64 = 1
+local WIRE_TYPE_LEN = 2
+-- SGROUP (3) and EGROUP (4) are deprecated in proto3.
+local WIRE_TYPE_I32 = 5
```
Totktonada, line 40: Can we define there constants in the wireformat module? It seems as a more appropriate place. We can use it from this module anyway, but this way the numbers are incapsulated in the low-level module.

src/lua/protobuf.lua
```diff
@@ -545,6 +581,39 @@ validate_scalar = function(field_def, value)
     end
 end
 
+local function validate_decoded_type(protocol, field_def, wire_type)
+    local protobuf_wire_types = {[0] = 'varint', [1] = 'I64',
+        [2] = 'len', [5] = 'I32'}
```
Totktonada, line 586: Here we use the wireformat type IDs and define wireformat type names. Let's move it to the wireformat module as well.

TBH, while looking over the code it seems for me that there should be something like wireformat registry in the wirefomat module, similar to `scalars` here. So, we can do something like this:

```lua
wireformat.VARINT.id => 0
wireformat.VARINT.name => 'varint'
```

src/lua/protobuf.lua
```diff
@@ -545,6 +581,39 @@ validate_scalar = function(field_def, value)
...
+    local exp_type = field_def.type
+    local exp_wire_type = WIRE_TYPE_VARINT
+    if exp_type == 'string' or exp_type == 'bytes' then
+            exp_wire_type = WIRE_TYPE_LEN
+    elseif exp_type == 'double' or exp_type == 'fixed64' or
+        exp_type == 'sfixed64' then
+            exp_wire_type = WIRE_TYPE_I64
+    elseif exp_type == 'float' or exp_type == 'fixed32' or
+        exp_type == 'sfixed32' then
+            exp_wire_type = WIRE_TYPE_I32
```
Totktonada, line 603: Can we include this mapping into `scalars`?

So, we can do something like the following:

```lua
local scalar_def = scalars[field_def.type]
scalar_def.wire_type -- this is the expected wire type
```

src/lua/protobuf.lua
```diff
@@ -657,9 +726,142 @@ end
...
+        res = decode(protocol, field_def.type, msg_data, ctx.opts)
+    end
+    return res
+end
+
+decode = function(protocol, message_name, data, opts)
+    local opts = opts or {}
+    assert(type(opts) == 'table')
+    local res = {}
+    local ctx = {val = data, pos = 1, opts = opts}
```
Totktonada, line 804: Let's describe all the fields of this context somewhere.

I suggest to create a constructor function for the decoding context and set unneded fields to `nil` explicitly. This way we can comment them in the same way as we do for non-nil fields.

```lua
local function new_decode_context(data, decode_opts)
    return {
        -- The source data to decode.
        val = data,
        -- The next position in the data to read.
        pos = 1,
        -- Blah blah.
        foo = nil,
    }
end
```

src/lua/protobuf.lua
```diff
@@ -657,9 +726,142 @@ end
...
+            table.insert(res, value)
+        end
+    else
+        repeat
+            local field_id
+        ctx.ignore_repeated = true
+            local value = decode_field(protocol, field_def, ctx)
+            table.insert(res, value)
+            if ctx.pos > string.len(ctx.val) then
+                return res
```
Totktonada, line 766: Shouldn't we compare with some amount-of-elements in the message or size-in-bytes of the message? The global position check seems irrelevant at parsing of the particular nested message (except the buffer boundary check at reading a byte).

src/lua/protobuf_wireformat.lua
```diff
@@ -64,6 +82,32 @@ local function as_int64(value)
...
+
+-- Converts input unsigned cdata into lua number if it fits lua
+-- number integer representation without rounding range.
+local function cast_unsigned_type(value)
+    assert(type(value) == 'cdata')
+    if value < 2ULL^53 - 1 then
+        value = tonumber(value)
+    end
+    return value
+end
```
Totktonada, line 109: In the RFC we discussed that values of the `number` type are produced for the `[-10^14+1; 10^14-1]` range and `cdata<uint64_t>` or `cdata<int64_t>` for values outside (and uint64_t vs int64_t choose is based on the schema type).

Here we violate this agreement about the ranges from the RFC. The motivation was to be consistent with other tarantool functions like `msgpack.decode`.

I had a guess that we can achieve it by using some already existing function like `tonumber64`, but it seems it works this way only for a string input (not cdata).

I've looked into `src/lua/msgpack.ffi.lua` and it seems that it performs it manually (see the `decode_u64` and `decode_i64` functions). OK. Let's at least steal the constants from it.

src/lua/protobuf_wireformat.lua
```diff
@@ -152,6 +283,28 @@ local function encode_fixed32(field_id, value)
...
+
+-- Decode a signed integral value encoded as I32.
+--
+-- Input value type: string.
+--
+-- Used for Protocol Buffers types: sfixed32.
+local function decode_sfixed32(ctx)
+    local res = string.sub(ctx.val, ctx.pos, ctx.pos + 3)
+    ctx.pos = ctx.pos + 4
+    return tonumber(ffi.cast('int32_t', from_int32(res)))
```
Totktonada, line 305: [n1256.pdf](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf), 6.3.1.3 Signed and unsigned integers (page 55 as shown by my pdf reader, page 43 in on the paper):

> 6.3.1.3 Signed and unsigned integers
>
> 1. When a value with integer type is converted to another integer type other than _Bool, if the value can be represented by the new type, it is unchanged.
> 2. Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type
until the value is in the range of the new type.<sup>49)</sup>
> 3. Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.
>
> 49\) The rules describe arithmetic on the mathematical value, not the value of a given type of expression.

The bullet 2 is not applicable (the new type is signed). If the value is above 2^31-1, the result is implementation defined according to the bullet 3.

I would avoid casting from unsigned type to a signed type with intention to get possibly negative values.

I think that we should have two different functions that convert 4 bytes in a string to a number: for example, `extract_uint32` and `extract_int32`. Both should return `number` (if we're about the 32 bit variants).

changelogs/unreleased/protobuf_decoder_module.md
Col-Waltz: Fixed all these mistakes in both pr description and commit message.

src/lua/protobuf_wireformat.lua
Col-Waltz: I agree that it may be confusing. I will replace input value type with 'decoding context' where `ctx` passed as parameter.

src/lua/protobuf_wireformat.lua
```diff
@@ -106,6 +173,21 @@ local function encode_tag(field_id, wire_type)
     return encode_varint(bit.bor(bit.lshift(field_id, 3), wire_type))
 end
 
+-- Decode a tag byte.
+local function decode_tag(ctx)
+    local tag = decode_int(ctx)
+    return bit.band(tag, 0x7), bit.rshift(tag, 3)
+end
```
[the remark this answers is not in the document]
Col-Waltz: Added comment about return values.

src/lua/protobuf_wireformat.lua
Col-Waltz: I agree it will be better to move this functions in API functions. Also it allowed to get rid of forward declaration of `decode_int` function.

src/lua/protobuf.lua
Col-Waltz: We can use this function but there is one problem. If this error raises on encoding we know `field_name` from protocol and can use it in error message. On decoding we can't know `field_name` because `field_id` is incorrect. I adjusted error message for this situation, now it returnes `decoded` instead of name if an error occures while decoding. I suppose that I can add this logic, that will change error message, in upcoming `wrap_error` function.

src/lua/protobuf_wireformat.lua
Col-Waltz: This check was implemented earlier to raise an error when decoder meets the end of the line while decoding varint value. It seems like this check does not work as intended. Added a function to read known amount of bytes from decoded string with boundary check.

Col-Waltz: Added a constant `MAX_VARINT_LEN` and added a comment to it.

src/lua/protobuf_wireformat.lua
Col-Waltz: As mentioned above added a function `read_bytes` that read needed amout of bytes from decoding context with boundary check.

src/lua/protobuf_wireformat.lua
Col-Waltz: I want to work on specific function `wrap_error` which will proceed generation of error message based on the context from which it was called. But it's implementation isn't ready yet.

src/lua/protobuf.lua
Col-Waltz: Thanks, fixed indentation here.

src/lua/protobuf.lua
Col-Waltz: I saved previous implementation because I thought this movement will look strange, so I've restored it as it was before and fixed indentation.

src/lua/protobuf.lua
Col-Waltz: These constants were already defined in `protobuf_wireformat.lua` but as I found out later these constants may be contained in one structure similar to `scalars` from `protobuf.lua`

src/lua/protobuf.lua
Col-Waltz: This table is needed to correlate recieved integer numbers from decoding with protobuf types. I moved this table to wireformat module and made links to wireformat types from wireformat module.

src/lua/protobuf_wireformat.lua
Col-Waltz: Added constants and logic identical to `src/lua/msgpack.ffi.lua`.

src/lua/protobuf_wireformat.lua
Col-Waltz: Added two different functions to extract signed and unsigned integers from decoded string.

src/lua/protobuf_wireformat.lua
Totktonada: Seems solved now.

src/lua/protobuf_wireformat.lua
Totktonada: Seems solved.

src/lua/protobuf.lua
Totktonada: It seems OK to adjust the error message depending on whether there is a known field name.

src/lua/protobuf_wireformat.lua
Totktonada: I think we can reduce amount of allocated GC objects (substrings), but I'm OK to go with this simple solution for now. At least, we're doing a very similar thing in encoding.

src/lua/protobuf.lua
Totktonada: OK, but now it is called `form_context` that feels unusual. How about `new_decode_context`?

src/lua/protobuf.lua
Totktonada: It seems like we it is the end of a source data, not the end of a substring.

The approach with extracting a substring with the source data of a nested message feels confusing for me. This way we need to re-create the decoding context, and it also seems not like contexts usually work.

We can accept a source data not as just a string, but the string + its length. This way we eliminate extracting a substring for each nested message and eliminate a need to re-create the context.

The leght should be passed in arguments, together with data (we can't store it in context, a per-frame value is needed).

While a string extraction approach reduces amount of arguments we need to pass over the functions (no explicit length), I think it worth to consider my proposal. At least, it was a surprise for me that the decoding context is re-created and I misinterpreted this check.

src/lua/protobuf.lua
Col-Waltz: Renamed function with `new_decode_context`.

src/lua/protobuf.lua
Col-Waltz: In context of message decoding this seems strange for me. Let's imagine that in the middle of decoding we decode field_id and type and then we decide that the next one field is message type and decoded as `LEN`. It means that we can read known amount of bytes and it contains full embedded message. We add existing path to new context and start decoding embedded message and guarantee that we will not run out of embedded message and add extra fields in it.

If we will not create new decoding context we have to pass length of embedded message in `decode_message` function and watch until position runs out of this boundary. As I see it can lead to unexpected behaviour. For example we have a protocol describing this message:
```
protocol = proto.protocol({
    proto.message('test_message', {
        val1 = {'int32', 1},
        val2 = {'test_message', 2},
        val3 = {'int32', 20000},
    })
})
```
And somebody encoded this values:
```
protocol:encode('test_message', {val1 = 12, val2 = {val1 = 500, val3 = 1}, val3 = 25})
```

As a result we will get this binary string: `08 0c 12 07 80 e2 09 01 08 f4 03 80 e2 09 19` where
`08 0c` - val1 =12
`12 07 80 e2 09 01 08 f4 03` - val2, where `80 e2 09 01` - val3 = 1, `08 f4 03` - val1 = 500
`80 e2 09 19` - val3 = 25

Lets say we lost one byte `03` then we will get this string: `08 0c 12 07 80 e2 09 01 08 f4 <03> 80 e2 09 19`
Now my decoder will take embedded message `12 07 80 e2 09 01 08 f4 80` as a new decoding context and return such an error: ` While decode first_message.val2.val3 on byte 7 occurred error: Reached end of message while decoding'

But if we will not create a new decoding context the decoder can read `08 f4 80 e2 09` because msb will be zero only on `09` byte. And wireformat function will not detect reading out of bound. Yes we can detect it later when we will return to `decode` function but how can we say on which byte the error occured?

Totktonada: Thanks for the example!

Let's look into it from the single-context-and-length-argument approach perspective.

> Lets say we lost one byte `03` then we will get this string: `08 0c 12 07 80 e2 09 01 08 f4 <03> 80 e2 09 19`
>
> Now my decoder will take embedded message `12 07 80 e2 09 01 08 f4 80` as a new decoding context and return such an error: <...>

I'll sketchy picture a call tree.

* decode_outer_message(s: `08 0c 12 07 80 e2 09 01 08 f4 <03> 80 e2 09 19`, len: 14)
  * decode_int32(s: `08 0c`, len: 2)
  * decode_nested_message(s: `12 07 80 e2 09 01 08 f4 <03> 80`, len: 9)
    * decode_int32(s: `80 e2 09`, len: 3)
    * decode_int32(s: `08 f4 <03> 80`, len: 3)
      => boundary error

We can't continue reading bytes beyond the given length, so we get the same buffer boundary error.

In fact, both approaches are the same from the algorithmic point of view.

However, I have two reasons to consider the single-context-and-length-argument approach:

1. Imagine we write this code on C. We likely keep some pointer from-where-to-read-next-byte and a right boundary *for the given call frame*. We unlikely allocate a new buffer and copy a part of the source string on each call frame.

   Now we're in Lua and it is a bit non-trivial to eliminate copying. It is OK. However, I would like if the code is written in a way similar to how we write it in C. This way we can re-write parts using FFI or C without major changes in the code.
2. More important point is how another person recognizes typical programming patterns from your code.

   I see a resursive algorithm that pass some context over the functions. Usually a context is needed in recursive code to keep something across the call frames. If something is named as 'context' or 'ctx', I can't ever imagine that it is new for each frame. It is not a context, it is something completely different. Maybe 'frame'. Or, at least 'message-local-context'.

   If we need to pass frame-related information in a recursive algorithm, we have two straightforward ways:

   * Pass it as an argument.
   * Use a stack.

   Using a partially-copied partially-global-but-partially-frame-local, hmm, context is not a direct way to do so. My pattern recognition fails here.

Col-Waltz: I got your point, now I pass bounds for decoding as a function parameter and context stays unchanged. I added mentioned above testcase into tests.

src/lua/protobuf.lua
```diff
@@ -500,6 +550,19 @@ local function validate_type(field_def, value, exp_type)
     return
 end
 
+validate_field_id = function(field_id, field_name)
```
ImeevMA: Why not move this function before the `message()` function?

Issue #9401, opened 2023-11-24
Anonymous replica doesn't receive synchro queue state during join


sergepetrenko:
**Bug description**

Anonymous replica doesn't receive the synchronous transaction queue state when joining to the master, and believes the queue is unclaimed. As a result, any subsequent request to a synchronous space or synchronous queue owner change is rejected by the replica due to queue owner mismatch.

**Steps to reproduce:**
```lua
-- Instance 1.

-- Step 1.
box.cfg{listen = 3301}
box.ctl.promote()
box.schema.user.grant('guest', 'replication')

-- Step 3.
box.ctl.demote()

-- Instance 2.
-- Step 2.
box.cfg{replication = 3301, replication_anon = true, read_only = true}

```

After issuing step 3, you'll notice the following error in instance 2 logs:
```
2023-11-24 13:46:54.817 [34832] main/110/applier/3301 txn_limbo.c:895 E> RAFT: rejecting RAFT_DEMOTE (32) request from origin_id 1 replica_id 1 term 3. Limbo owner mismatch, owner_id 0
2023-11-24 13:46:54.817 [34832] main/110/applier/3301 txn_limbo.c:897 E> ER_SPLIT_BRAIN: Split-Brain discovered: got a request from a foreign synchro queue owner
```
Replication will be stopped:
```
tarantool> box.info.replication
---
- 1:
    id: 1
    uuid: 54405f68-c58d-49e0-bb9f-6b5b2613d105
    lsn: 2
    upstream:
      peer: '3301'
      lag: 0.00021696090698242
      status: stopped
      idle: 29.93699
      message: 'Split-Brain discovered: got a request from a foreign synchro queue
        owner'
    name: null
...

```
`box.info.synchro` on instance 2:
```
tarantool> box.info.synchro
---
- queue:
    owner: 0
    term: 0
    len: 0
    busy: false
  quorum: 1
...

```
So, the anonymous replica didn't receive the synchronous queue state from the master when joining. Otherwise it would have at least `term = 2`.

PR #10314, opened 2024-07-25
[backport 2.11] Add stubs for file FETCH_SNAPSHOT

Serpentian:
The last commit https://github.com/tarantool/tarantool/commit/62c49367a2eab6d8008da2d7201ffa03e002c31e will not be backported since it adds a feature with index 10, in 2.11 we have only 4 of them. Moreover, it bumps iproto version to 8, but all previous feature will not be supported.

It's not crucial, server just won't reply, that it has such feature.

Changelogs added in this backport must be marked experimental!!!  But only if they're features

### DIFF
--- a/changelogs/unreleased/gh-9401-anon-replica-synchro.md
+++ b/changelogs/unreleased/gh-9401-anon-replica-synchro.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed anonymous replicas not receiving the synchronous transaction queue state
+  during join (gh-9401).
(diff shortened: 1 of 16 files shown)


changelogs/unreleased/gh-9401-anon-replica-synchro.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/replication
```
Serpentian, line 1: Since this is bugfix, I suppose, we can merge it as is, without `experimental` title

andreyaksenov (approved):

sergepetrenko (approved):
Thanks!

changelogs/unreleased/gh-9401-anon-replica-synchro.md
sergepetrenko: Sure.

PR #10317, opened 2024-07-26
Finish the CE part of the logarithmic offset RFC

### LINKED ISSUE #8204: Optimize select with offset option

mkostoevr:
This patchset completes the CE part of the logarithmic offset RFC:
1. accelerate the `select` with `offset` in memtx;
2. introduce the `offset` parameter of `index:pairs` method;
3. introduce the `index:offset_of` method.

Closes #8204

### DIFF
--- a/changelogs/unreleased/gh-8204-pairs-offset.md
+++ b/changelogs/unreleased/gh-8204-pairs-offset.md
@@ -1,0 +1,4 @@
+## feature/lua
+
+* Introduced a new `offset` parameter in the `index:pairs` method. It allows to
+  skip the first tuples of the iterator (gh-8204).
--- a/changelogs/unreleased/gh-8204-select-offset-count-perf.md
+++ b/changelogs/unreleased/gh-8204-select-offset-count-perf.md
@@ -1,0 +1,10 @@
+## feature/memtx
+
+* Improved performance of tree index methods: `select()` with the `offset`
+  specified and `count()` method. The underlying algorithm for these methods is
+  changed: the old one's time complexity was `O(n)`, where `n` is the value of
+  `offset` or the amount of counted tuples. The new algorithm's complexity is
+  `O(log(size))`, where `size` is the number of tuples stored in the index. Now
+  it does not depend on the `offset` value or the amount of tuples to count. It
+  is safe to use these functions with arbitrary big offset values and tuple
+  count (gh-8204).
--- a/extra/exports
+++ b/extra/exports
@@ -64,7 +64,7 @@ box_index_count
 box_index_get
 box_index_id_by_name
 box_index_iterator
-box_index_iterator_after
+box_index_iterator_with_offset
 box_index_len
 box_index_max
 box_index_min
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -495,9 +495,10 @@ box_index_count(uint32_t space_id, uint32_t index_id, int type,
 /* {{{ Iterators ************************************************/
 
 box_iterator_t *
-box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
-			 const char *key, const char *key_end,
-			 const char *packed_pos, const char *packed_pos_end)
+box_index_iterator_with_offset(uint32_t space_id, uint32_t index_id, int type,
+			       const char *key, const char *key_end,
+			       const char *packed_pos,
+			       const char *packed_pos_end, uint32_t offset)
 {
 	assert(key != NULL && key_end != NULL);
 	mp_tuple_assert(key, key_end);
@@ -541,8 +542,8 @@ box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
 	struct txn_ro_savepoint svp;
 	if (txn_begin_ro_stmt(space, &txn, &svp) != 0)
 		return NULL;
-	struct iterator *it = index_create_iterator_after(index, itype, key,
-							  part_count, pos);
+	struct iterator *it = index_create_iterator_with_offset(
+		index, itype, key, part_count, pos, offset);
 	txn_end_ro_stmt(txn, &svp);
 	if (it == NULL)
 		return NULL;
@@ -558,8 +559,8 @@ box_iterator_t *
 box_index_iterator(uint32_t space_id, uint32_t index_id, int type,
 		   const char *key, const char *key_end)
 {
-	return box_index_iterator_after(space_id, index_id, type, key, key_end,
-					NULL, NULL);
+	return box_index_iterator_with_offset(space_id, index_id, type, key,
+					      key_end, NULL, NULL, 0);
 }
 
 int
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -981,6 +981,34 @@ generic_index_create_iterator(struct index *base, enum iterator_type type,
 }
 
 
+struct iterator *
+generic_index_create_iterator_with_offset(struct index *base,
+					  enum iterator_type type,
+					  const char *key, uint32_t part_count,
+					  const char *pos, uint32_t offset)
+{
+	/* Create a reguar iterator. */
+	struct iterator *it = index_create_iterator_after(
+		base, type, key, part_count, pos);
+	if (it == NULL)
+		return NULL;
+
+	/* Skip the required amount of tuples. */
+	for (size_t i = 0; i < offset; i++) {
+		if (box_check_slice() != 0)
+			goto fail;
+		struct tuple *skipped_tuple;
+		if (iterator_next(it, &skipped_tuple) != 0)
+			goto fail;
+		if (skipped_tuple == NULL)
+			return it;
+	}
+	return it;
+fail:
+	iterator_delete(it);
+	return NULL;
+}
+
 struct index_read_view *
 generic_index_create_read_view(struct index *index)
 {
--- a/src/box/index.h
+++ b/src/box/index.h
@@ -575,6 +575,13 @@ struct index_vtab {
 					    const char *key,
 					    uint32_t part_count,
 					    const char *pos);
+	/** The same as create_iterator but skip first @offset tuples. */
+	struct iterator *(*create_iterator_with_offset)(struct index *index,
+							enum iterator_type type,
+							const char *key,
+							uint32_t part_count,
+							const char *pos,
+							uint32_t offset);
 	/** Create an index read view. */
 	struct index_read_view *(*create_read_view)(struct index *index);
 	/** Introspection (index:stat()) */
@@ -919,6 +926,15 @@ index_replace(struct index *index, struct tuple *old_tuple,
 }
 
 static inline struct iterator *
+index_create_iterator_with_offset(struct index *index, enum iterator_type type,
+				  const char *key, uint32_t part_count,
+				  const char *pos, uint32_t offset)
+{
+	return index->vtab->create_iterator_with_offset(
+		index, type, key, part_count, pos, offset);
+}
+
+static inline struct iterator *
 index_create_iterator_after(struct index *index, enum iterator_type type,
 			    const char *key, uint32_t part_count,
 			    const char *pos)
@@ -1079,6 +1095,11 @@ struct iterator *
 generic_index_create_iterator(struct index *base, enum iterator_type type,
 			      const char *key, uint32_t part_count,
 			      const char *pos);
+struct iterator *
+generic_index_create_iterator_with_offset(struct index *base,
+					  enum iterator_type type,
+					  const char *key, uint32_t part_count,
+					  const char *pos, uint32_t offset);
 int generic_index_build_next(struct index *, struct tuple *);
 void generic_index_end_build(struct index *);
 int
--- a/src/box/index.h
+++ b/src/box/index.h
@@ -256,14 +256,16 @@ box_tuple_extract_key(box_tuple_t *tuple, uint32_t space_id,
 /** \endcond public */
 
 /**
- * Allocate and initialize iterator for space_id, index_id. If packed_pos is
- * not NULL, iterator will start right after tuple with position, described by
- * this argument. A returned iterator must be destroyed by box_iterator_free().
+ * Allocate and initialize iterator for space_id, index_id and then skip @a
+ * offset tuples. If packed_pos is not NULL, iterator will start right after
+ * tuple with position, described by this argument. A returned iterator must
+ * be destroyed by box_iterator_free().
  */
 box_iterator_t *
-box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
-			 const char *key, const char *key_end,
-			 const char *packed_pos, const char *packed_pos_end);
+box_index_iterator_with_offset(uint32_t space_id, uint32_t index_id, int type,
+			       const char *key, const char *key_end,
+			       const char *packed_pos,
+			       const char *packed_pos_end, uint32_t offset);
 
 /**
  * A helper for position extractors. Get packed position of tuple in
@@ -342,7 +344,7 @@ struct iterator {
 	/**
 	 * Pointer to a buffer which was allocated for start position with
 	 * runtime_alloc. Will be freed on iterator_delete.
-	 * Needed for box_index_iterator_after.
+	 * Needed for box_index_iterator_with_offset.
 	 */
 	char *pos_buf;
 };
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -49,10 +49,12 @@ ffi.cdef[[
     typedef struct iterator box_iterator_t;
 
     box_iterator_t *
-    box_index_iterator_after(uint32_t space_id, uint32_t index_id, int type,
-                             const char *key, const char *key_end,
-                             const char *packed_pos,
-                             const char *packed_pos_end);
+    box_index_iterator_with_offset(uint32_t space_id, uint32_t index_id,
+                                   int type, const char *key,
+                                   const char *key_end,
+                                   const char *packed_pos,
+                                   const char *packed_pos_end,
+                                   uint32_t offset);
     int
     box_iterator_next(box_iterator_t *itr, box_tuple_t **result);
     void
@@ -2037,8 +2039,12 @@ end
 
 local function check_pairs_opts(opts, key_is_nil, level)
     local iterator = check_iterator_type(opts, key_is_nil, level and level + 1)
+    local offset = 0
     local after = nil
     if opts ~= nil and type(opts) == "table" then
+        if opts.offset ~= nil then
+            offset = opts.offset
+        end
         if opts.after ~= nil then
             after = opts.after
             if after ~= nil and type(after) ~= "string" and type(after) ~= "table"
@@ -2047,7 +2053,7 @@ local function check_pairs_opts(opts, key_is_nil, level)
             end
         end
     end
-    return iterator, after
+    return iterator, after, offset
 end
 
 box.internal.check_pairs_opts = check_pairs_opts
@@ -2458,16 +2464,16 @@ base_index_mt.pairs_ffi = function(index, key, opts)
     local ibuf = cord_ibuf_take()
     local pkey, pkey_end = tuple_encode(ibuf, key, 2)
     local svp = builtin.box_region_used()
-    local itype, after = check_pairs_opts(opts, pkey + 1 >= pkey_end, 2)
+    local itype, after, offset = check_pairs_opts(opts, pkey + 1 >= pkey_end, 2)
     local ok = iterator_pos_set(index, after, ibuf, 2)
     local keybuf = ffi.string(pkey, pkey_end - pkey)
     cord_ibuf_put(ibuf)
     local cdata
     if ok then
         local pkeybuf = ffi.cast('const char *', keybuf)
-        cdata = builtin.box_index_iterator_after(
+        cdata = builtin.box_index_iterator_with_offset(
                 index.space_id, index.id, itype, pkeybuf, pkeybuf + #keybuf,
-                iterator_pos[0], iterator_pos_end[0])
+                iterator_pos[0], iterator_pos_end[0], offset)
     end
     builtin.box_region_truncate(svp)
     if cdata == nil then
@@ -2479,11 +2485,11 @@ end
 base_index_mt.pairs_luac = function(index, key, opts)
     check_index_arg(index, 'pairs', 2)
     key = keify(key)
-    local itype, after = check_pairs_opts(opts, #key == 0, 2)
+    local itype, after, offset = check_pairs_opts(opts, #key == 0, 2)
     local keymp = msgpack.encode(key)
     local keybuf = ffi.string(keymp, #keymp)
     local cdata = internal.iterator(index.space_id, index.id, itype, keymp,
-        after, 2);
+        after, offset, 2);
     return fun.wrap(iterator_gen_luac, keybuf,
         ffi.gc(cdata, builtin.box_iterator_free))
 end
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -2515,6 +2515,23 @@ base_index_mt.count_luac = function(index, key, opts)
     return internal.count(index.space_id, index.id, itype, key);
 end
 
+-- 0-based iterator-relative offset of the first matching tuple. If such tuple
+-- does not exist, returns the offset at which it would be located if existed.
+--
+-- For an existing tuple, if index:offset_of(key, {iterator = it}) == N, then
+-- index:pairs(nil, {iterator = it, offset = N}) will start the iterator from
+-- the same tuple as index:pairs(key, {iterator = it})
+base_index_mt.offset_of = function(index, key, opts)
+    check_index_arg(index, 'offset_of', 2)
+    local itype = check_iterator_type(opts, #key == 0, 2)
+    if itype == box.index.EQ then
+        opts.iterator = box.index.GE
+    elseif itype == box.index.REQ then
+        opts.iterator = box.index.LE
+    end
+    return index:len() - index:count(key, opts)
+end
+
 base_index_mt.get_ffi = function(index, key)
     if builtin.box_read_ffi_is_disabled then
         return base_index_mt.get_luac(index, key)
@@ -2729,6 +2746,14 @@ space_mt.count = function(space, key, opts)
         return 0 -- empty space without indexes, return 0
     end
     return pk:count(key, opts)
+end
+space_mt.offset_of = function(space, key, opts)
+    check_space_arg(space, 'offset_of', 2)
+    local pk = space.index[0]
+    if pk == nil then
+        return 0 -- empty space without indexes, return 0
+    end
+    return pk:offset_of(key, opts)
 end
 space_mt.bsize = function(space)
     check_space_arg(space, 'bsize', 2)
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -2392,6 +2392,8 @@ static const struct index_vtab memtx_tree_disabled_index_vtab = {
 	/* .get = */ generic_index_get,
 	/* .replace = */ disabled_index_replace,
 	/* .create_iterator = */ generic_index_create_iterator,
+	/* .create_iterator_with_offset = */
+	generic_index_create_iterator_with_offset,
 	/* .create_read_view = */ generic_index_create_read_view,
 	/* .stat = */ generic_index_stat,
 	/* .compact = */ generic_index_compact,
@@ -2455,6 +2457,8 @@ get_memtx_tree_index_vtab(void)
 				 memtx_tree_index_replace<USE_HINT>,
 		/* .create_iterator = */
 			memtx_tree_index_create_iterator<USE_HINT>,
+		/* .create_iterator_with_offset = */
+		generic_index_create_iterator_with_offset,
 		/* .create_read_view = */
 			memtx_tree_index_create_read_view<USE_HINT>,
 		/* .stat = */ generic_index_stat,
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -314,6 +314,8 @@ struct tree_iterator {
 	enum iterator_type type;
 	struct memtx_tree_key_data<USE_HINT> after_data;
 	struct memtx_tree_key_data<USE_HINT> key_data;
+	/** The amount of tuples to skip after the iterator start. */
+	uint32_t offset;
 	/**
 	 * Data that was fetched last, needed to make iterators stable.
 	 * Contains NULL as pointer to tuple only if there was no data fetched.
@@ -916,13 +918,13 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 	struct memtx_tree_key_data<USE_HINT> start_data =
 		it->after_data.key != NULL ? it->after_data : it->key_data;
 	enum iterator_type type = it->type;
-	size_t unused;
+	size_t curr_offset;
 	/* The flag is true if the found tuple equals to the key. */
 	bool equals;
 	struct memtx_tree_data<USE_HINT> *initial_elem;
 	if (!memtx_tree_lookup(tree, &start_data, it->after_data,
 			       &type, region, &it->tree_iterator,
-			       &unused, &equals, &initial_elem))
+			       &curr_offset, &equals, &initial_elem))
 		return 0;
 
 	/*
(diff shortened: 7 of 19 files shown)


src/box/memtx_tx.c
```diff
@@ -3341,36 +3458,55 @@ memtx_tx_nearby_gap_item_new(struct txn *txn, enum iterator_type type,
...
+	item->until = until;
+	item->until_hint = until_hint;
+
+	/*
+	 * The @a until tuple may be invisible to the counting transaction and
+	 * may be deleted later on, let's ref it to prevent its deletion until
+	 * this gap item is deleted.
+	 */
+	if (item->until)
+		tuple_ref(item->until);
```
mkostoevr, line 3494: This must be revised after the TX management is documented cause the tuple won't be deleted until it's story is, but the lifetime of stories is not clear enough to me yet.

mkostoevr: The tuple is dirty, so there's a story hilding it from the deletion. And the story won't be deleted until the last gap item that uess it is deleted, but since the count tracking is performed after the ruple had been clarified, the tuple stories has `GAP_INPLACE` items in their `read_gaps`, which will only be deleted after all transaction's gap items are removed (including this count).

So it's OK to not reference the tuple here.

src/box/memtx_tree.cc
```diff
@@ -932,25 +975,59 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
...
+	if (it->skip != 0 && res != NULL) {
+		size_t need_skip_more_visible =
+			memtx_tx_index_invisible_count_matching_until(
+				in_txn(), space, index_base, type,
+				start_data.key, start_data.part_count,
+				res->tuple, res->hint);
+
+		/* Skip more visible tuples. */
+		memtx_tree_iterator_t<USE_HINT> *iterator = &it->tree_iterator;
+		while (need_skip_more_visible && res != NULL) {
```
CuriousGeorgiy, line 1002: Nit: `need_skip_more_visible != 0`.

src/box/memtx_tx.c
```diff
@@ -3309,6 +3403,30 @@ memtx_tx_inplace_gap_item_new(struct txn *txn)
 	return item;
 }
 
+/*
+ * Saves the given @a key in the @a short_key buffer of size @a short_key_size
+ * if it fits or allocates a new one on the @a txn's region. Returns the key
+ * length in the @a key_len output parameter.
+ */
+static char *
+memtx_tx_save_key(struct txn *txn, const char *key, uint32_t part_count,
```
CuriousGeorgiy, line 3412: Nit: looks like this can be introduced as a refactoring of gap items in a separate patch.

src/box/memtx_tree.cc
```diff
@@ -965,23 +1042,60 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
 	}
 
 /********MVCC TRANSACTION MANAGER STORY GARBAGE COLLECTION BOUND START*********/
+	if (it->skip != 0) {
```
CuriousGeorgiy, line 1045: This is out of scope of this PR, but:

Doesn't this kill the optimization? I understand there is no another way with our current isolation level, but I would argue this at least one example where one could benefit from weaker isolation levels.

src/box/memtx_tree.cc
```diff
@@ -932,25 +975,59 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
...
+		}
+	}
+
+	bool is_eq = type == ITER_EQ || type == ITER_REQ;
+
 	/* If we skip tuple, flag equals is not actual - need to refresh it. */
-	if (it->after_data.key != NULL && res != NULL &&
-	    (type == ITER_EQ || type == ITER_REQ)) {
+	if (((it->after_data.key != NULL && is_eq) || it->skip != 0) &&
+	    res != NULL) {
```
CuriousGeorgiy, line 1018: Nit: you do some microrefactoring across this patch, like replacing `(type == ITER_EQ || type == ITER_REQ)` with `is_eq `. It would be a little simpler to review if you would try to factor them out into separate patches.

src/box/memtx_tree.cc
```diff
@@ -965,23 +1042,60 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
...
 	/*
 	 * Since MVCC operates with `key_def` of index but `start_data` can
 	 * contain key extracted with `cmp_def`, we should crop it by passing
 	 * `part_count` not greater than `key_def->part_count`.
 	 */
 	uint32_t key_part_count = index->base.def->key_def->part_count;
-	if (!key_is_full ||
-	    ((type == ITER_GE || type == ITER_LE) && !equals) ||
-	    (type == ITER_GT || type == ITER_LT))
+	if (it->skip == 0 &&
```
CuriousGeorgiy, line 1095: Nit: maybe add a `goto` instead of `it->skip == 0`?

src/box/memtx_tree.cc
```diff
@@ -932,25 +975,59 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
...
 
+	/*
+	 * It we have skipped raw tuples, some of them may be invisible to the
+	 * current transaction, let's skip further in this case.
+	 */
+	struct txn *txn = in_txn();
+	if (it->skip != 0 && res != NULL) {
+		size_t need_skip_more_visible =
+			memtx_tx_index_invisible_count_matching_until(
+				in_txn(), space, index_base, type,
```
CuriousGeorgiy, line 996: Nit: `in_txn()` -> `txn`.

CuriousGeorgiy (approved):
Only reviewed `memtx: accelerate select with offset in tree` — LGTM.

changelogs/unreleased/gh-8204-pairs-offset.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Introduced a new `offset` parameter in the `index:pairs` method. It allows to
+  skip first tuples of the iterator (gh-8204).
```
lenkis, line 4:
```suggestion
  skip the first tuples of the iterator (gh-8204).
```

src/box/lua/schema.lua
```diff
@@ -2509,6 +2518,23 @@ base_index_mt.count_luac = function(index, key, opts)
     return internal.count(index.space_id, index.id, itype, key);
 end
 
+-- 1-based iterator-relative index of the first matching tuple. If the such a
```
lenkis, line 2521:
```suggestion
-- 1-based iterator-relative index of the first matching tuple. If such a
```

src/box/lua/schema.lua
```diff
@@ -2509,6 +2518,23 @@ base_index_mt.count_luac = function(index, key, opts)
...
+-- tuple does not exist, returns the index it would be located at if existed.
```
lenkis, line 2522:
```suggestion
-- tuple does not exist, returns the index that would be located at if existed.
```

src/box/memtx_tx.h
```diff
@@ -318,6 +318,57 @@ memtx_tx_track_gap(struct txn *txn, struct space *space, struct index *index,
...
+		     const char *key, uint32_t part_count)
+{
+	if (!memtx_tx_manager_use_mvcc_engine)
+		return 0;
+	return memtx_tx_track_count_until_slow(txn, space, index, type, key,
+					       part_count, NULL, HINT_NONE);
+}
+
+/**
+ * Record in TX manager that a transaction @a txn have counted @a index from @a
```
lenkis, line 351:
```suggestion
 * Record in TX manager that a transaction @a txn has counted @a index from @a
```

changelogs/unreleased/gh-8204-select-offset-count-perf.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/box
+
+* Significantly improved the performance of `select` requests with the
+  `offset` specified and `count` requests. This prevents DoS conditions
```
lenkis, line 4:
```suggestion
  `offset` specified and with `count` requests. This prevents DoS conditions
```

lenkis (approved):

changelogs/unreleased/gh-8204-pairs-offset.md
mkostoevr: Fixed, thanks!

src/box/lua/schema.lua
mkostoevr: Fixed, thanks!

src/box/lua/schema.lua
mkostoevr: Reworded with "the index at which it would be located" to make it clear. Please let me know if you're OK with this.

changelogs/unreleased/gh-8204-select-offset-count-perf.md
mkostoevr: The "`select` requests with the `offset` specified" and "`count` requests" are separated entities, performance is improved for each of them. It does not seem "with" can be applied here.

Please feel free to correct me.

src/box/memtx_tree.cc

src/box/memtx_tx.c
mkostoevr: Done.

src/box/memtx_tree.cc
mkostoevr: It does in case of a big number of concurrent statements. We've discussed offline with @alyapunov and @drewdzzz about this issue and it's been desided to prefer consistency over performance with minimal amount of changes until we have real MVCC users reporting about performance problems.

But I aggree with your point in general, MVCC-oblivious count would hurt much less in some cases.

src/box/memtx_tree.cc
mkostoevr: Done.

src/box/memtx_tree.cc

src/box/memtx_tree.cc

src/box/memtx_tree.cc
CuriousGeorgiy: > But I aggree with your point in general, MVCC-oblivious count would hurt much less in some cases.

It’s not just about count, it’s also about select with offset.

Could you please open a ticket about allowing to lower the isolation level?

mkostoevr: Not sure I understood the your main concern. There're two aspects of this:
1. The line you refer to creates a global `GAP_COUNT` item, which slows down any concurrent inserts and deletes until the counting transaction is finished. Do you mean to introduce an ability to prevent count gap tracking? This way a transaction that once counted items in an index may have an inconsistent value received next time.
2. It's also possible (in theory), count raw tuples without traversing existing stories (which may be slow if we have a big amount of ones) to substract invisible ones. This way the transaction will potentially see uncommitted concurrent writes and subjects for garbage collection.

mkostoevr: Not sure I understood the idea. There're two aspects of this:
1. The line you refer to creates a global `GAP_COUNT` item, which slows down any concurrent inserts and deletes until the counted (selected with offset) transaction is finished. Do you mean to introduce an ability to prevent count gap tracking? This way a transaction that once counted items in an index may have an inconsistent value received next time. It's worse for select, where it can return different sets of values on each indentical request, for example:

   ```lua
   -- Index: [1, 3, 5, 7, 9]
   --[[TX1]]: select(nil, {offset = 3}) -- [7, 9]
   --[[TX2]]: insert(2)                 -- Now index: [1, 2, 3, 5, 7, 9].
   -- If TX1 doesn't create count gap item, TX2 doesn't conflict with it, so no read view for TX1.
   --[[TX1]]: select(nil, {offset = 3}) -- [5, 7, 9]
   ```

2. It's also possible (in theory), count raw tuples without traversing existing stories to substract invisible ones. This way the transaction will potentially count uncommitted concurrent writes and subjects for garbage collection. It may be OK for `index:len()`, but not for select with offset where we must only count the offset for visible tuples. The result for select is the same as above in this case.

CuriousGeorgiy: I was rather referring to 2, because I don’t like the additional traversing. But giving it a thought, I guess it doesn’t worsen the situation drastically, compared to the existing MVCC overheads.

mkostoevr:
Rebased to the current master after since the base PR (#10139) had been merged.

src/box/memtx_tx.c
```diff
@@ -3027,12 +3075,37 @@ memtx_tx_tuple_clarify_slow(struct txn *txn, struct space *space,
...
+		if (until != NULL &&
+		    !memtx_tx_tuple_is_before(cmp_def, story->tuple,
+					      until, until_hint, type))
+			continue;
+
+		/* All tuples in the story chain share the same key. */
+		if (!memtx_tx_tuple_matches(key_def, story->tuple,
+					    type, key, part_count))
+			continue;
+
```
locker, line 3108: Nit: It'd be nice if this was done in one helper function because this is the second place where you call them one after another.

src/box/memtx_tree.cc
```diff
@@ -314,6 +314,8 @@ struct tree_iterator {
 	enum iterator_type type;
 	struct memtx_tree_key_data<USE_HINT> after_data;
 	struct memtx_tree_key_data<USE_HINT> key_data;
+	/** The amount of tuples to skip after the iterator start. */
+	uint32_t skip;
```
locker, line 318: Nit: Better name it `offset` - to match the iterator parameter.

src/box/memtx_tree.cc
```diff
@@ -886,10 +913,26 @@ memtx_tree_lookup(memtx_tree_t<USE_HINT> *tree,
...
+	/*
+	 * Skip the amount of tuples required. Here raw tuples are skipped,
+	 * but in case of multiple concurrent transactions, some of these
+	 * tuples might be invisible to the current one, this is handled on
+	 * the caller site.
+	 */
+	if (skip != 0) {
+		memtx_tree_skip(tree, iterator, offset, skip,
+				iterator_type_is_reverse(*type));
+	}
```
locker, line 935: IMO better do it outside `memtx_tree_lookup` - in `tree_iterator_start`. Also, I'd inline `memtx_tree_skip` since it's small, used in one place, and having it closer to the rest of the "skip" logic would make the code easier to follow.

extra/exports
```diff
@@ -65,6 +65,7 @@ box_index_get
 box_index_id_by_name
 box_index_iterator
 box_index_iterator_after
```
locker, line 67: Looks like it's not used anymore. Please drop it.

src/box/lua/schema.lua
```diff
@@ -2046,8 +2049,14 @@ local function check_pairs_opts(opts, key_is_nil, level)
                 box.error(box.error.ITERATOR_POSITION, level and level + 1)
             end
         end
+        if opts.offset ~= nil then
+            offset = opts.offset
+            if type(offset) ~= "number" then
+                box.error(box.error.INVALID_OFFSET, level and level + 1)
+            end
+        end
```
locker, line 2057: We don't have a separate error code for this in `select`. I don't see a point in adding one for `pairs`. `INVALID_ARG` should do.

src/box/lua/schema.lua
```diff
@@ -2518,6 +2518,23 @@ base_index_mt.count_luac = function(index, key, opts)
     return internal.count(index.space_id, index.id, itype, key);
 end
 
+-- 1-based iterator-relative index of the first matching tuple. If such a tuple
+-- does not exist, returns the index at which it would be located if existed.
+--
+-- For an existing tuple, if index:index_of(key, {iterator = it}) == N, then
+-- index:pairs(nil, {iterator = it, offset = N - 1}) will start the iterator
+-- from the same tuple as index:pairs(key, {iterator = it})
+base_index_mt.index_of = function(index, key, opts)
```
locker, line 2527: Maybe name it `offset_of` - for consistency with the `offset` parameter? Also, may be make it 0-based so that it can be passed to `pairs` with a decrement?

src/box/memtx_tx.c
```diff
@@ -3369,16 +3473,31 @@ memtx_tx_nearby_gap_item_new(struct txn *txn, enum iterator_type type,
...
+	struct memtx_tx_mempool *pool = &txm.count_gap_item_mempool;
+	struct count_gap_item *item = memtx_tx_xmempool_alloc(txn, pool);
+	gap_item_base_create(&item->base, GAP_COUNT, txn);
+
+	item->type = type;
+	item->part_count = part_count;
+	item->key = memtx_tx_save_key(txn, key, part_count, item->short_key,
+				      sizeof(item->short_key), &item->key_len);
+	item->until = until;
```
drewdzzz, line 3498: The tuple should be referenced here.

src/box/memtx_tree.cc
```diff
@@ -978,13 +979,12 @@ tree_iterator_start(struct iterator *iterator, struct tuple **ret)
...
 	 */
-	uint32_t key_part_count = index->base.def->key_def->part_count;
 	if (!key_is_full ||
 	    ((type == ITER_GE || type == ITER_LE) && !equals) ||
 	    (type == ITER_GT || type == ITER_LT))
 		memtx_tx_track_gap(txn, space, index_base, successor, type,
-				   start_data.key,
-				   MIN(start_data.part_count, key_part_count));
+				   start_data.key, MIN(start_data.part_count,
+				   index_base->def->key_def->part_count));
```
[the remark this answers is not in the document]
mkostoevr: This commit makes the next one less invasive (and thus, easier to read). I can squash it once the review is completed.

src/box/memtx_tx.c
```diff
@@ -3339,19 +3363,8 @@ memtx_tx_nearby_gap_item_new(struct txn *txn, enum iterator_type type,
...
-		item->key = NULL;
-	} else if (item->key_len <= sizeof(item->short_key)) {
-		item->key = item->short_key;
-	} else {
-		item->key = memtx_tx_xregion_alloc(txn, item->key_len,
-						   MEMTX_TX_ALLOC_TRACKER);
-	}
-	memcpy((char *)item->key, key, item->key_len);
+	item->key = memtx_tx_save_key(txn, key, part_count, item->short_key,
+				      sizeof(item->short_key), &item->key_len);
```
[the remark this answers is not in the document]
mkostoevr: See the [comment above](https://github.com/tarantool/tarantool/pull/10317#discussion_r1741899121).

changelogs/unreleased/gh-8204-select-offset-count-perf.md
```diff
@@ -0,0 +1,5 @@
...
+  previously possible (gh-8204).
```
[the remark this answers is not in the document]
mkostoevr: Updated the changelog and added documentation to the commit regarding the [`index_object:count()`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/count/), [`index_object:select()`](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/select/) and the [`difference between memtx and vinyl storage engines`](https://www.tarantool.io/en/doc/latest/platform/engines/memtx_vinyl_diff/).

src/box/memtx_tx.c
mkostoevr: Done.

src/box/memtx_tree.cc

extra/exports
mkostoevr: Done.

src/box/lua/schema.lua
mkostoevr: Done, the error is catched on the Lua side now (as it is for the regular select).

src/box/lua/schema.lua
mkostoevr: This was the first thig I suggested in the RFC, but it's [been decided](https://www.notion.so/select-offset-N-count-key-log-n-6fead8b03db6409e9d09f56dd58f1006?d=028a099c41a64d249da33a06c526bfef&pvs=4#aa664802eeca45b39f24878ee9bd01a2) to stuck with this solution.

src/box/memtx_tree.cc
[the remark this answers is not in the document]
locker: It isn't for me, actually. Without the context (which is in a different commit), I don't understand why this is done - this change by itself doesn't make the code any better or worse.

src/box/lua/schema.lua
locker: I read the comments in the doc. AFAIU the argument for `index_of` was to avoid confusion with `offset`, which is 0-based, while `index_of` was suggested to be 1-based because it's common in the Lua world. But `offset` is already 0-based so I don't understand why we should bother making `index_of` 1-based, thus introducing a different term.

src/box/memtx_tree.cc
locker: This function can't be used in the read view iterator as is because we use different types for read views. I guess we'll have to duplicate the code anyway.

src/box/memtx_tree.cc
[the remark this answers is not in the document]
mkostoevr: Squashed the commits into one ("memtx: accelerate select with offset in tree") since all the reviewers have watched the changes.

src/box/memtx_tree.cc
mkostoevr: This is a matter of overloading a bunch of functions: see the [corresponding commit in the read-view PR](https://github.com/tarantool/tarantool/pull/10478/commits/a83f5e3c42aaf27ebfdd8b3850b9df475b48eb34).

This may make code bigger, but it creates a single point of responsibility for the lookup behavior.

The same trick is done for the read-view [count implementation](https://github.com/tarantool/tarantool/pull/10357/commits/9316f0643992f428d47c216b212f83550240f7f3), so it's generic too.

src/box/lua/schema.lua
mkostoevr: Here're the discussion stages:
1. Initially the function was `offset_of` and returned a 0-based value.
2. Then it's been said we better to return 1-based values in the Lua world.
3. Then it's been said it's better to be called `index_of` instead of `offset_of` if it's 1-based.

I aggree with your point but I'm afraid of doing anything against the [RFC](https://www.notion.so/tarantool/select-offset-N-count-key-log-n-6fead8b03db6409e9d09f56dd58f1006), feel free to add a comment there.

Important thing to be noted: the product document required the [`indexof` method returning a 1-based value](https://www.notion.so/tarantool/offset-n-count-key-log-n-4aba908b48674123892474957420c85f?pvs=4#1981fb9b6757424cb961e0c1e574b1bc) (the link to the exact block may not work as expected, see the very last block of the document).

src/box/memtx_tx.c
mkostoevr: Given the following statements:
1. The `until` tuple is [clarified](https://github.com/tarantool/tarantool/blob/3d984f6e0421c2442b396ddbd35ff7d4b1c1c8e6/src/box/memtx_tree.cc#L1022), so it's dirty and there's a story [hilding](https://github.com/tarantool/tarantool/blob/218fc46f5f44e6e27be73c6dceaa58f1eb377f46/src/box/memtx_tx.c#L967) it from the deletion.
   1. The tuple story has `GAP_INPLACE` item in its read_gaps.
2. And the story [won't](https://github.com/tarantool/tarantool/blob/218fc46f5f44e6e27be73c6dceaa58f1eb377f46/src/box/memtx_tx.c#L1627-L1632) be garbage collected until the last gap item that ues it is deleted.
4. The gap item will only be deleted after all gap items of the transaction are removed (including this count).

So if the count gap item is alive (and we can access its `until` member):
1. The transaction that created it is still alive.
2. The`until` tuple story is not garbage collected since it has the `GAP_INPLACE` of the transaction in it.
3. The tuple is not deleted since it has the story referencing it.

So it's OK to not reference the tuple here.

changelogs/unreleased/gh-8204-pairs-offset.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
```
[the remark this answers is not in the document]
mkostoevr: Fixed, thanks!

src/box/index.h
```diff
@@ -575,6 +575,13 @@ struct index_vtab {
 					    const char *key,
 					    uint32_t part_count,
 					    const char *pos);
+	/** The same as create_iterator but skip first @offset tuples. */
+	struct iterator *(*create_iterator_with_offset)(struct index *index,
+							enum iterator_type type,
+							const char *key,
+							uint32_t part_count,
+							const char *pos,
+							uint32_t offset);
```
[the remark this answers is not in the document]
mkostoevr: Seems like a way to go if we don't want the slice check in the index virtual methods. The only alternative discussed in the RFC was introducing a new `uint32_t offset` parameter to the existing method.

I'd implement the way you suggested if @locker has no objections.

src/box/memtx_tx.c
mkostoevr: Given the following statements:
1. The `until` tuple is [clarified](https://github.com/tarantool/tarantool/blob/3d984f6e0421c2442b396ddbd35ff7d4b1c1c8e6/src/box/memtx_tree.cc#L1022), so it's dirty and there's a stor
y [hilding](https://github.com/tarantool/tarantool/blob/218fc46f5f44e6e27be73c6dceaa58f1eb377f46/src/box/memtx_tx.c#L967) it from the deletion.
   1. The tuple story has `GAP_INPLACE` item in its read_gaps.
2. And the story [won't](https://github.com/tarantool/tarantool/blob/218fc46f5f44e6e27be73c6dceaa58f1eb377f46/src/box/memtx_tx.c#L1627-L1632) be garbage collected until the last gap item that ues it is deleted.
4. The gap item will only be deleted after all gap items of the transaction are [removed](https://github.com/tarantool/tarantool/blob/218fc46f5f44e6e27be73c6dceaa58f1eb377f46/src/box/memtx_tx.c#L3505-L3511) (including this count).

So if the count gap item is alive (and we can access its `until` member):
1. The transaction that created it is still alive.
2. The`until` tuple story is not garbage collected since it has the `GAP_INPLACE` of the transaction in it.
3. The tuple is not deleted since it has the story referencing it.

So it's OK to not reference the tuple here.

src/box/lua/schema.lua
locker: I talked to @unera. He says that he suggested the name `index_of` in the original design document because it's named so in Redis. He doesn't mind naming it `offset_of`. He says that making it 0-based would be better because it would be consistent with the `offset` parameter of `select`. Making it 1-based would require the user to subtract 1 from it before using it with `select`.

Another reason I don't like the name `index_of` because `box.index.foo.index_of(...)` looks weird ("index" here is used twice but has completely different meanings).

src/box/memtx_tree.cc
locker: I don't think it's a good idea because tree and its view are handled differently. For example, we don't need to do any kind of tracking in a view in contrast to a tree. If the tx manager was implemented at a different level (not in the index), it could make sense but otherwise it will probably make the code convoluted IMO. Besides, the view is implemented in EE and it should stay there for now.

mkostoevr: Inlined the iterator skip into the `tree_iterator_start`.

src/box/lua/schema.lua
mkostoevr: Renamed to `offset_of` and made it return 0-based values. Also added a no-index case to the test suite.

src/box/index.h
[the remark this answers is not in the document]
locker: > Let's add a new argument `uint32_t *offset` to `create_iterator` and each index will decrease `offset` value if it can do it better than N advancements. For example, tree index will always return 0 here since it can do offset in `O(log n)`, and hash index won't touch this argument at all so we will advance the iterator manually in `box_select`.

This would obscure the function contract. I see nothing wrong in adding a new virtual method - it looks clearer to me. The situation with pagination is different because an index implementation either supports it or not so it either succeeds or raises an error.

src/box/memtx_tx.c
drewdzzz: When DDL happens, all the stories are deleted, including one referencing the tuple. So, even if the transaction is aborted by DDL, it's still actually alive (waits for rollback) so the count gap and the deleted tuple still can be accessed.

src/box/index.cc
```diff
@@ -981,6 +981,32 @@ generic_index_create_iterator(struct index *base, enum iterator_type type,
...
+					  const char *key, uint32_t part_count,
+					  const char *pos, uint32_t offset)
+{
+	/* Create a reguar iterator. */
+	struct iterator *it = index_create_iterator_after(
+		base, type, key, part_count, pos);
+
+	/* Skip the required amount of tuples. */
+	for (size_t i = 0; i < offset; i++) {
+		if (box_check_slice() != 0)
```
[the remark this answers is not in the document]
drewdzzz: Discussed with Sergey Petrenko, he's agree with you, so I resolve this comment.

src/box/index.h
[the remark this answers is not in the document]
drewdzzz: OK, we decided to check slices in index, so there is no need to implement offset for other indexes outside `create_index`.
What's about separate methods - I still don't like it, but you've already discussed it in RFC, so I resolve the comment.

src/box/memtx_tx.c
mkostoevr: Haven't considered the DDL machinery. Thanks!

```diff
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -340,6 +340,12 @@ memtx_tx_count_gap_item_new(struct txn *txn, enum iterator_type type,
                            const char *key, uint32_t part_count,
                            struct tuple *until, hint_t until_hint);
 
+/**
+ * Destroy the count gap item.
+ */
+static void
+memtx_tx_count_gap_item_destroy(struct gap_item_base *base);
+
 /**
  * Helper structure for searching for point_hole_item in the hash table,
  * @sa point_hole_item_pool.
@@ -3165,6 +3165,7 @@ memtx_tx_delete_gap(struct gap_item_base *item)
                pool = &txm.nearby_gap_item_mempoool;
                break;
        case GAP_COUNT:
+               memtx_tx_count_gap_item_destroy(item);
                pool = &txm.count_gap_item_mempool;
                break;
        case GAP_FULL_SCAN:
@@ -3473,11 +3474,7 @@ memtx_tx_nearby_gap_item_new(struct txn *txn, enum iterator_type type,
 }
 
 /**
- * Allocate and create count gap item. The @a until tuple's story must have
- * a gap item from the @a txn transaction or be tracked by it, so the story
- * is not deleted by the garbage collector and the tuple is not deleted (if
- * it's not NULL).
- *
+ * Allocate and create count gap item.
  * Note that in_read_gaps base member must be initialized later.
  */
 static struct count_gap_item *
@@ -3497,10 +3494,20 @@ memtx_tx_count_gap_item_new(struct txn *txn, enum iterator_type type,
                                      sizeof(item->short_key), &item->key_len);
        item->until = until;
        item->until_hint = until_hint;
+       if (item->until != NULL)
+               tuple_ref(item->until);
 
        return item;
 }

+/**
+ * Destroy the count gap item.
+ */
+static void
+memtx_tx_count_gap_item_destroy(struct gap_item_base *base)
+{
+       struct count_gap_item *item = (struct count_gap_item *)base;
+       if (item->until != NULL)
+               tuple_unref(item->until);
+}
+
 /**
  * Allocate and create full scan gap item.
  * Note that in_read_gaps base member must be initialized later.
```

drewdzzz (approved):
Thanks for the patch, LGTM!

mkostoevr:
@lenkis, please take a look at the new version of the [changelogs/unreleased/gh-8204-select-offset-count-perf.md](https://github.com/tarantool/tarantool/pull/10317/files#diff-a6ec629ca26766374a8f6b8d77a7ea1643986628f0f45b6133123a24c3ed2e5e).

src/box/memtx_tx.c
mkostoevr: @drewdzzz rebased and dropped these changes since the #10160 (which removes all the global gaps including `GAP_COUNT` on DDL along with stories) is merged.

mkostoevr:
Rebased and dropped the GAP_COUNT destruction since the #10160 (which removes all the global gaps including GAP_COUNT on DDL along with stories) is merged.

changelogs/unreleased/gh-8204-select-offset-count-perf.md
```diff
@@ -0,0 +1,10 @@
+## feature/memtx
+
+* Improved performance of tree index methods: `select()` with the `offset`
+  specified and `count()` method. The underlying algorithm for these is
+  changed: the old one's time complexity was `O(n)` where `n` is the value of
+  `offset` or the amount of counted tuples. The new algorithm's complexity is
+  `O(log(size))` where `size` is the number of tuples stored in the index. So
+  now it does not depend on the value of `offset` or the amount of tuples to
+  count and it's safe to use these functions with arbitrary big offset values
+  and tuple count (gh-8204).
```
xuniq, line 10:
```suggestion
* Improved performance of tree index methods: `select()` with the `offset`
  specified and `count()` method. The underlying algorithm for these methods is
  changed: the old one's time complexity was `O(n)`, where `n` is the value of
  `offset` or the amount of counted tuples. The new algorithm's complexity is
  `O(log(size))`, where `size` is the number of tuples stored in the index. Now it does not depend on the `offset` value or the amount of tuples to
  count. It is safe to use these functions with arbitrary big offset values
  and tuple count (gh-8204).
```

xuniq (approved):

PR #10319, opened 2024-07-29
connpool: support filtering by `sharding.roles`

### LINKED ISSUE #10318: connpool: support filtering by `sharding.roles`
Sometimes it's convenient to have a way to filter out all router or storage replicasets in connpool. The role of the replicaset is specified by the `sharding.role` option, which is not supported in `connpool`.

Need to support filtering by a `sharding_roles` option in `connpool.filter()`.

georgiy-belyanin:
Support filtering replicasets by the configured sharding role by adding `sharding_roles` into the `connpool.filter()` and `connpool.call()` options.

Closes #10318

### DIFF
--- a/changelogs/unreleased/gh-10318-connpool-sharding-roles-filter.md
+++ b/changelogs/unreleased/gh-10318-connpool-sharding-roles-filter.md
@@ -1,0 +1,4 @@
+## feature/connpool
+
+- The `connpool.filter()` and `connpool.call()` functions now support
+  filtering by the `sharding.roles` option (gh-10318).
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -172,6 +172,8 @@ local function is_candidate_match_static(names, opts)
            is_replicaset_match(opts.replicasets, names.replicaset_name) and
            is_instance_match(opts.instances, names.instance_name) and
            is_roles_match(opts.roles, config:get('roles', get_opts)) and
+           is_roles_match(opts.sharding_roles,
+                          config:get('sharding.roles', get_opts)) and
            is_labels_match(opts.labels, config:get('labels', get_opts))
 end
 
@@ -196,6 +198,7 @@ local function filter(opts)
         instances = '?table',
         labels = '?table',
         roles = '?table',
+        sharding_roles = '?table',
         mode = '?string',
     })
     opts = opts or {}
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/gh-10318-connpool-sharding-roles-filter.md
```diff
@@ -0,0 +1,4 @@
+## feature/connpool
+
+- `connpool.filter()` now supports filtering by a `sharding.roles`
```
xuniq, line 3:
```suggestion
- The `connpool.filter()` function now supports filtering by the `sharding.roles`
```

xuniq (approved):

changelogs/unreleased/gh-10318-connpool-sharding-roles-filter.md
georgiy-belyanin: Thank you! Applied this change.

src/box/lua/connpool.lua
```diff
@@ -196,19 +218,33 @@ local function filter(opts)
         instances = '?table',
         labels = '?table',
         roles = '?table',
+        sharding_roles = '?table',
```
Totktonada, line 221: I would expect that all the `connpool.filter` filters are supported by `connpool.call`.

src/box/lua/connpool.lua
```diff
@@ -196,19 +218,33 @@ local function filter(opts)
...
+            if sharding_role ~= 'storage' and
+               sharding_role ~= 'rebalancer' and
+               sharding_role ~= 'router' then
+               local msg = 'Unknown sharding role \"%s\" in '..
+                           'connpool.filter() call. Expected one of the '..
+                           '\"storage\", \"router\", \"rebalancer\"'
+               error(msg:format(sharding_role), 0)
+            end
+        end
+    end
```
Totktonada, line 240: I would exclude `rebalancer`, because it is not very trivial and I guess that it is not really needed by users.

The default `sharding.rebalancer_mode` is `'auto'`, which means that the rebalancer chooses an appropriate instance on its own (some of the RW storage instances). It means that the option is not always static and, moreover, usually dynamic. We can track where the rebalancer dynamically (like we do for RO/RW mode), but I think that it doesn't worth the effort.

At the same time, supporting of the static rebalancer configuration without supporting the dynamic one would look confusing, as for me. I would just raise an error that `sharding_role = "rebalancer"` is not supported (this error should be different from the unknown sharding role error).

src/box/lua/connpool.lua
```diff
@@ -149,6 +149,26 @@ local function is_roles_match(expected_roles, present_roles)
...
+    for _, present_role_name in pairs(present_roles) do
+        roles[present_role_name] = true
+    end
+    for _, expected_role_name in pairs(expected_roles) do
+        if roles[expected_role_name] == nil then
+            return false
+        end
+    end
+    return true
+end
```
Totktonada, line 170: It is a copy of `is_roles_match`. I would just call `is_roles_match`.

georgiy-belyanin: Thank you. Applied.

src/box/lua/connpool.lua
georgiy-belyanin: Sounds. Fixed this strange inconsistency.

georgiy-belyanin:
I have some unexpected breaks when trying to evaluate the tests locally. Made the PR draft until I've managed with the problems.

src/box/lua/connpool.lua
Totktonada: The implementation looks OK, but, please, add a test case for `connpool.call` with the `sharding_roles` option.

Totktonada (approved):
I see no flaws visually. LGTM.

Please, let me know when this PR becomes ready.

src/box/lua/connpool.lua
georgiy-belyanin: Added the test.

Totktonada:
Pushed to `master`, a future `3.2.0` release.

Issue #1803, opened 2016-09-30
Cross-engine transactions


kostja:
Allow to mix Vinyl and MemTX operations in a single transactions.

dyu:
:+1:

rohitjoshi:
:1

CuriousGeorgiy:
@alyapunov

The recent [_Efficiently Making Cross-Engine Transactions Consistent_](https://sigmodrecord.org/publications/sigmodRecord/2303/pdfs/08_skeena-zhang.pdf) paper by Tianzheng Wang et al. may be a good reference.

unera:
@locker Let's do it.

@Totktonada, do You have any suggestions/notes for?

PR #10325, opened 2024-07-29
txn: first step towards cross-engine transactions

Gumix:
This patch replaces `txn::engine` with `txn::engines[3]`, so that a multi-statement transaction can involve up to 3 storage engines. However, for now the `ENGINE_SUPPORTS_CROSS_ENGINE_TX` flag is set only for `memtx` engine (when MVCC is off), so there are no user-visible changes.

Needed for #1803
Needed for tarantool/tarantool-ee#823

### DIFF
--- a/src/box/engine.h
+++ b/src/box/engine.h
@@ -40,6 +40,20 @@
 extern "C" {
 #endif /* defined(__cplusplus) */
 
+enum {
+	/**
+	 * For simplicity, assume that the total engine count can't exceed
+	 * the value of this constant.
+	 */
+	MAX_ENGINE_COUNT = 10,
+	/**
+	 * Max number of engines involved in a multi-statement transaction.
+	 * This value must be greater than any `engine::id' of an engine
+	 * without `ENGINE_BYPASS_TX' flag.
+	 */
+	MAX_TX_ENGINE_COUNT = 3,
+};
+
 struct engine;
 struct engine_read_view;
 struct txn;
@@ -281,6 +295,10 @@ enum {
 	 * Engine setting this flag must support read views.
 	 */
 	ENGINE_JOIN_BY_MEMTX = 1 << 3,
+	/**
+	 * Set if the engine supports cross-engine transactions.
+	 */
+	ENGINE_SUPPORTS_CROSS_ENGINE_TX = 1 << 4,
 };
 
 struct engine {
@@ -338,7 +356,7 @@ struct engine_join_ctx {
 	void **data;
 };
 
-/** Register engine engine instance. */
+/** Register engine instance. */
 void engine_register(struct engine *engine);
 
 /** Call a visitor function on every registered engine. */
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -614,6 +614,7 @@ memtx_engine_prepare(struct engine *engine, struct txn *txn)
 	if (memtx_tx_manager_use_mvcc_engine) {
 		struct txn_stmt *stmt;
 		stailq_foreach_entry(stmt, &txn->stmts, next) {
+			assert(stmt->engine == engine);
 			assert(stmt->space->engine == engine);
 			memtx_tx_history_prepare_stmt(stmt);
 		}
@@ -631,10 +632,11 @@ memtx_engine_commit(struct engine *engine, struct txn *txn)
 	struct txn_stmt *stmt;
 	stailq_foreach_entry(stmt, &txn->stmts, next) {
 		if (memtx_tx_manager_use_mvcc_engine) {
+			assert(stmt->engine == engine);
 			assert(stmt->space->engine == engine);
 			memtx_tx_history_commit_stmt(stmt);
 		}
-		if (stmt->engine_savepoint != NULL) {
+		if (stmt->engine == engine && stmt->engine_savepoint != NULL) {
 			struct space *space = stmt->space;
 			struct tuple *old_tuple = stmt->rollback_info.old_tuple;
 			if (space->upgrade != NULL && old_tuple != NULL)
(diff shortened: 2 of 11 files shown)


mkostoevr (approved):
LGTM.

It's an interesting decision to address engines in the `struct txn` by their IDs.

src/box/txn.c
```diff
@@ -553,18 +555,33 @@ txn_begin_in_engine(struct engine *engine, struct txn *txn)
...
+	}
+	txn->engines_count++;
+	txn->engines[engine->id] = engine;
+
+	if (txn->engines_count > 1) {
+		for (size_t i = 0; i < MAX_TX_ENGINE_COUNT; i++) {
+			struct engine *engine = txn->engines[i];
+			if (engine != NULL &&
+			    !(engine->flags &
+			      ENGINE_SUPPORTS_CROSS_ENGINE_TX)) {
```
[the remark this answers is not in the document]
Gumix: It’s not enough. If we insert a second engine which supports cross-engine transaction, we have to validate the first engine. But if there is only one engine, it’s ok if it doesn’t support cross tx.

src/box/memtx_engine.cc
```diff
@@ -614,7 +614,9 @@ memtx_engine_prepare(struct engine *engine, struct txn *txn)
 	if (memtx_tx_manager_use_mvcc_engine) {
 		struct txn_stmt *stmt;
 		stailq_foreach_entry(stmt, &txn->stmts, next) {
-			assert(stmt->space->engine == engine);
+			assert(stmt->space->engine == stmt->engine);
+			if (stmt->engine != engine)
+				continue;
```
locker, line 619: If memtx mvcc is enabled, there won't be any alien statements on the list, no?

src/box/memtx_engine.cc
```diff
@@ -631,7 +633,9 @@ memtx_engine_commit(struct engine *engine, struct txn *txn)
 	struct txn_stmt *stmt;
 	stailq_foreach_entry(stmt, &txn->stmts, next) {
 		if (memtx_tx_manager_use_mvcc_engine) {
-			assert(stmt->space->engine == engine);
+			assert(stmt->space->engine == stmt->engine);
+			if (stmt->engine != engine)
+				continue;
```
locker, line 638: There shouldn't be any alien tuples in this case AFAIU. At the same time, I don't think we should call `memtx_space_upgrade_untrack_tuple` below for alien tuples.

src/box/txn.c
```diff
@@ -553,18 +555,33 @@ txn_begin_in_engine(struct engine *engine, struct txn *txn)
...
+				diag_set(ClientError,
+					 ER_CROSS_ENGINE_TRANSACTION,
+					 engine->name);
+				goto error;
+			}
+		}
 	}
```
locker, line 577: I don't like this check. Let's rework.
1. Add two boolean flags to `txn_flag`: `TXN_IS_STARTED` and `TXN_SUPPORTS_MULTI_ENGINE` (the names might need revising). `TXN_IS_STARTED` is initially false; `TXN_SUPPORTS_MULTI_ENGINE` is initially true.
2. If we add a new engine and `TXN_IS_STARTED` is false, set it to true, and clear `TXN_SUPPORTS_MULTI_ENGINE` if the engine doesn't support multi-engine transactions.
3. If we add a new engine that doesn't support multi-engine transactions and `TXN_IS_STARTED` is set, abort.
4. If we add a new engine and `TXN_SUPPORTS_MULTI_ENGINE` is unset, abort.

The we wouldn't need to iterate over all engines here.

locker (changes requested):

src/box/memtx_engine.cc
Gumix: Yes, this change is unnecessary at this point.

src/box/memtx_engine.cc
Gumix: Changed to:
```diff
@@ -631,10 +632,11 @@ memtx_engine_commit(struct engine *engine, struct txn *txn)
        struct txn_stmt *stmt;
        stailq_foreach_entry(stmt, &txn->stmts, next) {
                if (memtx_tx_manager_use_mvcc_engine) {
+                       assert(stmt->engine == engine);
                        assert(stmt->space->engine == engine);
                        memtx_tx_history_commit_stmt(stmt);
                }
-               if (stmt->engine_savepoint != NULL) {
+               if (stmt->engine == engine && stmt->engine_savepoint != NULL) {
                        struct space *space = stmt->space;
                        struct tuple *old_tuple = stmt->rollback_info.old_tuple;
                        if (space->upgrade != NULL && old_tuple != NULL)
```

src/box/txn.c
Gumix: OK, reworked this way.

src/box/engine.h
```diff
@@ -296,6 +314,13 @@ struct engine {
 	struct rlist link;
 };
 
+/** True if `engine' supports cross-engine transactions. */
+static inline bool
+engine_supports_cross_engine_tx(const struct engine *engine)
+{
+	return engine->flags & ENGINE_SUPPORTS_CROSS_ENGINE_TX;
+}
+
```
locker, line 323: Nit: We don't like helper functions for other engine flags so I wouldn't introduce a flag for this one, either. You could use a local variable to avoid long lines.

src/box/txn.c
```diff
@@ -553,17 +559,40 @@ txn_begin_in_engine(struct engine *engine, struct txn *txn)
...
+		assert(txn->engines[engine->id] == engine);
+		return 0;
+	}
+	if (!txn_has_flag(txn, TXN_IS_STARTED_IN_ENGINE)) {
+		txn_set_flags(txn, TXN_IS_STARTED_IN_ENGINE);
+		assert(txn_has_flag(txn, TXN_SUPPORTS_MULTI_ENGINE));
+		if (!engine_supports_cross_engine_tx(engine))
+			txn_clear_flags(txn, TXN_SUPPORTS_MULTI_ENGINE);
+		goto begin_in_engine;
+	}
```
locker, line 573: `engine_begin()` may fail, in which case the transaction flags will be wrong. You should update the flags only on success.

locker (changes requested):

src/box/engine.h
Gumix: Removed.

src/box/txn.c
Gumix: Fixed.

PR #10329, opened 2024-07-30
Fixes in CMake modules

ligurio:


Buristan (approved):
Hi, Sergey!
Thanks for the patchset!
I will proceed with the review per-patch.

---

**[PATCH 1/3] cmake: fix warning about missed DOWNLOAD_EXTRACT_TIMESTAMP**

As an alternative approach, we may set the `CMP0135` policy to `NEW` at the top level of the CMake to be sure that this warning won't return in the future. But since new a "ExternalProject" with the `URL` field isn't added frequently, it is OK to leave it as is.

The patch itself LGTM, after fixing a bunch of insignificant nits below regarding to the commit message.

Commit title width is more than 50 symbols.

> CMake 3.24+ produce a warning about missed option

Typo: s/produce/produces/
Typo: s/missed/the missed/

> The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy
CMP0135 is not set.

It is better to format this part with `|` like [the following](https://github.com/tarantool/tarantool/commit/ed2d260f2972eaebb6f96ded9d0b01916ac13a83), to make it clear that this is a warning message.

Also, it looks like the `[2]` link is missed in the message itself.

> CMake 3.23 and below sets the timestamps of the extracted contents

Typo: s/sets/set/

> CMake 3.24 and above prefers to set the timestamps of all

Typo: s/perfers/prefer/

---

**[PATCH 2/3] cmake: fix building statically by Ninja**

LGTM, after fixing some nits regarding the commit message.

> On attempt to build Tarantool statically by Ninja (with enabled

Typo: s/attempt/an attempt/
Typo: s/enabled/the enabled/

> CMake option BUILD_STATIC_WITH_BUNDLED_LIBS) error below is

Typo: s/) error/), error/

---

**[PATCH 3/3] cmake: fix a warning**

LGTM.

Minor: I would reformulate this like the following:
`test/fuzz: fix the CMake warning`
Feel free to ignore.

It is better to separate the warning message itself from the commit message with `|` or ` ``` `.

ligurio:
>  Thanks for the patchset!

@Buristan Sergey, thanks for review! Fixed all issues and force-pushed to the branch.

ImeevMA (approved):

Issue #9376, opened 2023-11-16
Premature flush of limbo during promote


funbringer:
**Bug description**

While node is being promoted, `txn_limbo_write_promote` writes a `RAFT_PROMOTE` request to its wal and immediately proceeds to flush the limbo in `txn_limbo_read_promote`:

https://github.com/tarantool/tarantool/blob/4b88d5fd9833b1c236572e7c557646909ec415eb/src/box/txn_limbo.c#L576-L579

https://github.com/tarantool/tarantool/blob/4b88d5fd9833b1c236572e7c557646909ec415eb/src/box/txn_limbo.c#L1196-L1198

https://github.com/tarantool/tarantool/blob/4b88d5fd9833b1c236572e7c557646909ec415eb/src/box/txn_limbo.c#L591-L593

However, according to the raft paper (3.6.2),

> As described in Section 3.5, a leader knows that an entry from its current term is committed once that entry is stored on a majority of the servers. If a leader crashes before committing an entry, future leaders will attempt to finish replicating the entry. However, **a leader cannot immediately conclude that an entry from a previous term is committed once it is stored on a majority of servers.**
> ...
> To eliminate problems like the one in Figure 3.7, Raft never commits log entries from previous terms by counting replicas. Only log entries from the leader’s current term are committed by counting replicas; once an entry from the current term has been committed in this way, then all prior entries are committed indirectly because of the Log Matching Property. There are some situations where a leader could safely conclude that an older log entry is committed (for example, if that entry is stored on every server), but Raft takes a more conservative approach for simplicity.

Until the newly elected leader has **committed** a new raft entry in the current term, it is still possible to rewrite the history using an entry with a newer term. If we consider that rewinding a limbo flush is not possible, this means that the current behavior is not correct. Furthermore, this causes problems in production when the current leader writes `RAFT_PROMOTE` but fails to deliver it to other nodes on time, causing another "split brain" scenario.

TL;DR: We should only apply the side effects of promote after a quorum has been reached for a new entry with the current term (e.g. the `RAFT_PROMOTE` itself).

sergepetrenko:
See also: #8095

PR #10334, opened 2024-07-30
feat(qsync): PROMOTE now requires quorum & CONFIRM

funbringer:
This patchset fundamentally changes the behavior of the protocol message `IPROTO_RAFT_PROMOTE`.

# The problem

Previously, we'd first win raft elections becoming a raft leader, then write the message unconditionally **w/o any quorum**. This works fine under many circumstances, but things become more complicated once we introduce network partitions & lags and cluster node hiccups (remember that the node might slow down significantly or become unresponsive due to heavy load or insufficient fiber context switches, etc). Once we start writing `IPROTO_RAFT_PROMOTE`, there is no going back -- if we limp and fail to deliver it in time, any other node that succeeds in becoming a new leader will cause a split brain check to go off. This has been described previously in https://github.com/tarantool/tarantool/issues/9376.

# The proposed solution

To mitigate this problem, the algorithm becomes a little more complicated:
1. Start the elections using `election_mode = "candidate"` or `box.ctl.promote()`.
2. Win the elections, becoming a raft leader.
3. Claim the limbo using `IPROTO_RAFT_PROMOTE`:
	1. Write an `IPROTO_RAFT_PROMOTE` (promote request) in the current term...
	2. **and link it to the preceding unconfirmed promote requests (or `limbo->promote_greatest_term` if none) using `PREV_TERM`** _(more on that later)_.
	3. Add it to the queue `limbo->pending_promotes`.
	4. Gather quorum on the promote request by counting ACKs from fellow nodes.
	5. Once the quorum is gathered, write an `IPROTO_RAFT_CONFIRM` (confirm request) pointing to the promote request (aka confirm-for-promote).
	6. Now everyone reads the confirm request and applies the promote request **and its dependencies** (`PREV_TERM`) in order.
4. Success, now the leader owns the limbo and the cluster is R/W.

It's impossible for the raft leader to add new transactions to the limbo until the ownership transition has completed.
	
Every promote request is uniquely identified by its term (per raft), so we can store all of them in a list sorted by term. Until the leader gathers quorum on **its own promote request written in the current term** (the active request, see `txn_limbo_get_active_promote`), it may not assume limbo ownership or commit/rollback regular transactions. Once the quorum is gathered, all promote requests reachable from the active one via `PREV_TERM` are committed as well, thus becoming parts of history retroactively.

Whenever a node receives a foreign promote request, it may not know whether it has gathered quorum or not -- only its author is responsible for counting ACKs, but the author might die suddenly and never write a confirm-for-promote again. The promote request in question is now known as the pending promote request. If a new leader suddenly comes to power, it has to deal with the consequences. It may not drop the pending promote request, because it might get confirmed eventually (now or later). This is exactly why we add it as a dependency (a prerequisite in our path to owning the limbo) via `PREV_TERM`. This situation might repeat several times in a row, so we have to introduce a queue for all pending promote requests.

The algorithm above makes sure that failures at steps 3.4 and 3.5 -- including the inability to complete them in time -- do not cause a split brain or a permanent cluster desynchronization. A newly elected leader should always link all promote requests is has witnessed, so any other node should eventually come to the same conclusions regarding the limbo ownership.

If the leader has reached step 3.1 but failed to complete the write in its term, the promote request will not gather quorum and will be nopified upon arrival. Same goes for the corresponding confirm-for-promote at step 3.5 -- either the promote request reaches the quorum and **will eventually become a part of history** with the help of a more recent promote request (`PREV_TERM`), or it never gathered quorum and won't get a corresponding confirm-for-promote. Also keep in mind that if the message gathered a quorum, any future leader must have it.

# Regarding the upgrades

The patch adds two new iproto keys: `PREV_TERM` (0x71) and `WAIT_ACK` (0x72). We have observed an upgrade problem in clusters with tarantool prior to https://github.com/tarantool/tarantool/commit/ee0660b8316966e4f9b4c79b18b6a9837e83ac89. This is something to keep in mind. Luckily, the problem does not affect 2.11.2.

Speaking of the upgrade routine itself: there are no guard rails per se, but there is a limited compatibility mode for legacy promote requests (the ones that don't have `WAIT_ACK` set to `true`). We apply those immediately w/o quorum, which means that we can easily read old snapshots and even respect the newly-elected leaders running the pre-patch tarantool version. However, once the upgraded instance emits a new-style promote request, it will eventually cause a transient "split brain" error in all pre-patch nodes that happen to be running. The error will go away once the node is restarted with a post-patch tarantool binary.

Thus, the effective way to upgrade is the following:
1. Upgrade all replicas, but not the current leader L.
2. Promote any upgraded replica to leader, causing a transient "split brain" in L.
3. Upgrade L's binary and restart it, completing the upgrade.

# The patchset structure

Currently, the commits may lack elaborate commit messages, yet each has its own meaning nonetheless. Certain undecided features (e.g. everything `box.ctl.wait`-related) have been extracted to their own commits. We really shouldn't squash them prematurely so as not to inhibit quick removal of poor ideas.

# Questions

## Should `box.ctl.promote()` wait for the promote request to take effect?

Currently, `box.ctl.promote()` **does not wait for the promote to take effect**. However, it does block until a new promote is written and queued. In order to wait for completion, one is expected to use `box.ctl.wait_rw()`. Sadly, we can't say the same about `election_mode = "candidate"`, because it is very vague and implicit.

In other words, the recipe is:
* `box.ctl.promote()` => `box.ctl.wait_rw()`.
* `election_mode = "candidate"` => `while box.info.synchro.queue.term < box.info.election.term do fiber.sleep(0) end` (used in a few tests).

For now, this seems to be the perfect spot: not too blocking (no need to spawn a fiber most of the time) and not too irritating (can be dealt with). However, the author has no strong opinions on this matter and is willing to discuss the alternatives.

## What about `IPROTO_RAFT_DEMOTE`?

Oh, this is a tricky one. Given that `IPROTO_RAFT_DEMOTE` only becomes relevant when `election_mode = 'off'` and serves as a "synchro escape hatch", we pretty much don't care about this mode in Picodata. Thus, I left a non-quorum implementation in place just to make the tests pass. However, if you **really feel** like we should implement quorum demote for the sake of the upstream, I won't argue.

## What about on-disk data format (snapshots)?

We deliberately decided not to change the snapshot format and `IPROTO_JOIN_META`. Instead, we disallow both snapshotting and adding new nodes while the limbo ownership transition is taking place.

## What about tests?

Considering that `box.ctl.promote()` does not wait for the promote to take effect, we had to add a dozen `box.ctl.wait_rw()` calls here and there. There are a few tests that didn't make it -- we had to drop them out of inability to properly fix them. Hopefully, the maintainers will give a better clue.

Furthermore, we have added a new test suite (`qpromote`) for this feature specifically -- among others, it contains a few test cases which routinely cause "split brain" in pre-patch versions (many thanks to @LizardWizzard).

---

Original patch: https://git.picodata.io/picodata/tarantool/-/merge_requests/175


Gerold103:
This is huge and very valuable work, absolutely gorgeous research and tremendous effort. That is probably the biggest and most useful patch made outside of Tarantool VK team.

> If a new leader suddenly comes to power, it has to deal with the consequences. It may not drop the pending promote request, because it might get confirmed eventually (now or later). This is exactly why we add it as a dependency (a prerequisite in our path to owning the limbo) via PREV_TERM. This situation might repeat several times in a row, so we have to introduce a queue for all pending promote requests.

What happens if the old leader wrote PROMOTE, replicated it to one other instance, then died. And then a third instance wins next term and write a PROMOTE too. Would it successfully receive that first PROMOTE and put it before its own one? 

> Promote any upgraded replica to leader, causing a transient "split brain" in L.

Would it work to shut down L, then let one of the replicas become a new leader, then restart L with the new binary? Looks like this wouldn't cause a split-brain, right?

> Should box.ctl.promote() wait for the promote request to take effect?

I am fine with how it works now. I.e. let it emit a promote, but do not wait to become writable. I am also fine with making it blocking (wait until become fully established leader or a new term happens). If we want to do less work, then I saw lets leave it as is.

> What about IPROTO_RAFT_DEMOTE?

I would say it must be synchronous too. It seems like it would be easier to make it work same as PROMOTE instead of making an exception for DEMOTE. Until I had a closer in-depth look at the code, can't say more.

> we disallow both snapshotting and adding new nodes while the limbo ownership transition is taking place.

Hm? Even if the limbo is empty? JOIN and snapshotting can't add non-committed synchro txns, but if the limbo is empty, they should work fine. The promote queue doesn't go into the snapshot anyway.

src/box/txn_limbo.c
```diff
@@ -1006,43 +1450,141 @@ txn_limbo_filter_confirm_rollback(struct txn_limbo *limbo,
...
+	struct pending_promote *item;
+	struct rlist *queue = &limbo->pending_promotes;
+	rlist_foreach_entry_reverse(item, queue, link) {
+		if (req->term == item->req.term) {
+			say_error("%s. Duplicate term detected",
+				  reject_str(req));
+			diag_set(ClientError, ER_SPLIT_BRAIN,
+				 "Replication",
+				 "PROMOTE with a duplicate term");
+		}
```
Gerold103, line 1606: Is `return -1` missing here?

src/box/txn_limbo.h
```diff
@@ -156,6 +175,11 @@ struct txn_limbo {
 	 * limbo and raft are in sync and the terms are the same.
 	 */
 	uint64_t promote_greatest_term;
```
Gerold103, line 200: Doesn't this now become "promote greatest confirmed term"? AFAIU, it gets updated only when a PROMOTE gets confirmed. Not when it is just written to WAL. Otherwise imagine this:
1. Node 1 write PROMOTE and hangs.
2. Node 2 bumps term a few times and write PROMOTE too.
3. Now Node-2-PROMOTE gets delivered to Node-1. It will pass all filters until `txn_limbo_filter_promote()`. Here `req->prev_term == limbo->promote_greatest_term` would be `false` and `rlist_foreach_entry_reverse()` won't find the preceeding PROMOTE. Because Node-1 has PROMOTE say term 10 and Node-2's PROMOTE can be 20.

But it seems to work fine when `promote_greatest_term` is updated only once a new PROMOTE gets confirmed. And that looks so in the code too.

Did I get it right? Should we also update this comment then? Right now it says that the term is updated right after PROMOTE WAL write.

src/box/txn_limbo.c
```diff
@@ -363,6 +415,11 @@ void
 txn_limbo_checkpoint(const struct txn_limbo *limbo, struct synchro_request *req,
 		     struct vclock *vclock)
 {
+	/* For simplicity we prohibit checkpoints during leadership changes. */
+	if (!rlist_empty((struct rlist *)&limbo->pending_promotes))
+		say_warn("limbo checkpoint: cannot capture pending PROMOTEs");
```
Gerold103, line 432: 1. That looks like a warning which won't prohibit the snapshot, no? If it is prohibited on an upper level, then shouldn't this be an assert?
2. Why do we prohibit it? What happens if the limbo is empty, we make a snapshot with a non-empty promote queue, and restart? Wouldn't the node then just re-enter the elections gracefully? I.e. bump term, start new elections, etc.

src/box/txn_limbo.c
```diff
@@ -393,44 +466,116 @@ synchro_request_write(const struct synchro_request *req)
...
+		/* Process all confirmed PROMOTE requests in order. */
+		rlist_foreach_entry_safe(item, queue, link, tmp) {
+			struct synchro_request *req = &item->req;
+			txn_limbo_apply_promote(limbo, req, req->origin_id);
+			rlist_del_entry(item, link);
+			free(item);
+
+			if (item == confirmed)
+				break;
+		}
```
Gerold103, line 576: Can't find where do you sort the promotes. Are they sorted by term? The only place where promotes are being added to the queue is `txn_limbo_make_promote()`. Can it happen, than node-1 makes a PROMOTE, then hangs, then node-2 bumps term, makes PROMOTE too. Then node-1 unhangs and node-2 receives an older PROMOTE from node-1. Looks like it should be placed in the beginning of the queue, but from the code it would be placed in the end, violating the order. I am 99% sure I've missed something. Could you help?

Gerold103 (commented):
Reading this patch is a pleasure, fun, and very insightful 💘. How your patch was evolving from the initial versions of sending promote queues over the network and now to this relatively easy to understand and elegant PR is amazing.

I will be back with more comments. Will write them in portions when have time.

src/box/txn_limbo.c
```diff
@@ -553,50 +751,215 @@ txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
...
+		? rlist_last_entry(queue, typeof(*last), link)
+		: NULL;
+
+	uint64_t greatest_term = limbo->promote_greatest_term;
+	*req = (struct synchro_request) {
 		.type = IPROTO_RAFT_PROMOTE,
-		.replica_id = limbo->owner_id,
 		.origin_id = instance_id,
-		.lsn = lsn,
+		.wait_ack = true,
```
Gerold103, line 779: You don't seem to need `wait_ack` field. It is always equal to `type == IPROTO_RAFT_PROMOTE`, from what I see. Why did you add it?

src/box/txn_limbo.c
```diff
@@ -553,50 +751,215 @@ txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
...
+int
+txn_limbo_write_promote(struct txn_limbo *limbo, int64_t lsn, uint64_t term)
+{
+	assert(latch_is_locked(&limbo->promote_latch));
+	int rc = -1;
+
+	struct synchro_request req;
+	txn_limbo_make_promote(limbo, lsn, term, &req);
+
+	limbo->is_writing_promote = true;
```
Gerold103, line 802: Would you add an assert here, that `is_writing_promote` is false right now?

src/box/txn_limbo.c
```diff
@@ -553,50 +751,215 @@ txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
...
+
+int
+txn_limbo_wait_promote_attempts(struct txn_limbo *limbo,
+				uint64_t count, double timeout)
+{
+	struct fiber_cond *cond = &limbo->write_promote_cond;
+	while (limbo->write_promote_count < count)
+		if (fiber_cond_wait_timeout(cond, timeout) != 0)
+			return -1;
 	return 0;
```
Gerold103, line 842: This won't hold if the calling fiber would be simply woken up manually via `fiber.wakeup()`. Why do we need to wait for it this way? I can only imagine places which don't need to wait at all or need to wait for a timeout, not for a count.

Gerold103 (commented):
Here is the next portion of questions. I am glad that I can learn from this patchset.

funbringer:
Thanks for the review! I'll return to this asap.

PR #10335, opened 2024-07-30
Introduce protobuf parser

Col-Waltz:
Introducing protobuf parser.
Parser can process .proto format files and return a protocol
similar to that returned by the `protobuf:protocol()` function.

Part of #9844

@TarantoolBot document
Title: Protobuf parser
Product: Tarantool
Root document: -

Protobuf parser API

Introducing protobuf parser. The only argument of the function is the
path to the .proto file. This parser supports only 'proto3' syntax
and can`t process any other syntax levels and editions.

Function `protobuf:protocol()` supports a limited set of protobuf
constructions therefore parser does the same. For example: keyword
`oneof` can't be parsed, because encoder does not support 'Oneof'
construction.

Parser also provides validation to field and enum ids and tracks
names and types declaration. Thus return protocol can be used in
`protobuf:encode()` function without additional validation.

Simple example of .proto file:

```
syntax = 'proto3';

message Employee {
    string name = 1;
    int32 id = 2;
    bool has_email = 3;
    DomainType email_domain = 4;
}

enum DomainType {
    mail = 0;
    gmail = 1;
    yandex = 2;
}
```


changelogs/unreleased/protobuf_parser.md
```diff
@@ -0,0 +1,3 @@
+## feature/lua
+
+* Introduced lua parser for .proto files (gh-9844).
```
andreyaksenov, line 3:
```suggestion
* Introduced a Lua parser for `.proto` files (gh-9844).
```

andreyaksenov (approved):

Issue #9722, opened 2024-02-21
Subdue remaining system fibers on graceful shutdown


nshy:
In the scope of #8423 we started to implement graceful Tarantool shutdown. In the process of shutdown we finish client and system fibers in some order so that we don't access already shutdown subsystems. But yet some system fibers are remain unfinished or we just do not care whether they finished or not. These fibers seems to work well even with shutdown subsystems thus they are not tamed in scope of #8423.

Yet we need to care for them too. On Tarantool shutdown we free fibers stacks thus we'd better finish all fibers to avoid FP memleaks when suppressions will finally be removed.

The list of untamed fibers:
```
src/box/memtx_engine.cc:        memtx->gc_fiber = fiber_new_system("memtx.gc", memtx_>
src/lib/core/fiber_pool.c:                      f = fiber_new_system(cord_name(cord()>
src/box/lua/net_box.c:  transport->worker = fiber_new_system(name, netbox_worker_f);
src/lib/swim/swim.c:    swim->event_handler = fiber_new_system("SWIM event handler",
src/box/alter.cc:                       fiber = fiber_new_system("synchro_filter_enab>
src/box/alter.cc:                       fiber = fiber_new_system("synchro_filter_disa>
src/box/raft.c:         box_raft_worker = fiber_new_system("raft_worker",
```
```
src/lua/fiber.lua:fiber._internal.set_system = fiber_set_system
src/lua/init.lua:    fiber._internal.set_system(fiber.self())
```

We probably will not finish fiber that called `os.exit()` (the second in the second list). For this fiber we can just don't free the stack.

PR #10353, opened 2024-08-02
box: drop iproto connections before client fibers on shutdown

### LINKED ISSUE #10313: `cannot gracefully shutdown client fibers` error in tests
After commit e513d4ce51d25ae56ca356183e6999ad6c298867 some tests start failing with the following error in logs:
```
2024-07-25 11:56:56.914 [624966] main/103/on_shutdown F> cannot gracefully shutdown client fibers
```
The most obvious one is `replication/replicaset_ro_mostly.test.lua`, but I've also seen `replication/election_basic.test.lua` and `replication/election_qsync_stress.test.lua` fail with the same error.

Probably there are more tests that started failing like this, but I just haven't caught them yet.

drewdzzz:
Currently we have such order of shutting down: client fibers, then
iproto subsystem and then fiber pool (details can be found in https://github.com/drewdzzz/tarantool/commit/e513d4ce51d25ae56ca356183e6999ad6c298867).
The problem is iproto produces new fibers in `tx_fiber_pool` when
processing new requests so when Tarantool is shutting down client
fiber, a new one can appear and it won't be handled. Moreover, there is
no point in handling such fibers because one can appear even after all
client fibers are finished.

Also, we cannot just stop `tx_fiber_pool` so that it won't produce new
workers while shutting down client fibers because it's also needed for
vinyl. But since only iproto can produce new client (non-system)
long-living fibers when shutting down, let's simply drop all iproto
connections before finishing client fibers. From user's point of view,
one thing will be changed - he will receive `Peer closed` error message
instead of `fiber is cancelled` for hanging requests. I don't think it's
bad because the new message better reflects that Tarantool is being
stopped and that the error is not retriable.

By the way, we already disallow accepting new iproto connections right
before shutdown (`iproto_on_shutdown_f`) but the problem is connected
client can keep sending requests. And that's why our replication tests
became flaky - if shutdown happens between JOIN and SUBSCRIBE, fiber
pool produces new client fiber after we cancelled all such fibers.

Follow-up #9722
Closes #10313


src/box/box.cc
```diff
@@ -6165,6 +6165,14 @@ box_shutdown(void)
...
 	 */
 	box_watcher_shutdown();
+	/*
+	 * Iproto connections should be dropped before client fibers because
+	 * they can produce new ones in `tx_fiber_pool`.
+	 */
+	if (iproto_drop_connections(box_shutdown_timeout) != 0) {
+		diag_log();
+		panic("cannot gracefully shutdown iproto requests");
+	}
```
locker, line 6175: > Currently we have such order of shutting down: client fibers, then iproto subsystem and then fiber pool (details can be found in https://github.com/tarantool/tarantool/commit/e513d4ce51d25ae56ca356183e6999ad6c298867). The problem is iproto produces new fibers in `tx_fiber_pool` when processing new requests so when Tarantool is shutting down client fiber, a new one can appear and it won't be handled. Moreover, there is no point in handling such fibers because one can appear even after all client fibers are finished.

A proper IPROTO connection should stop sending any requests and exit as soon as it returns a shutdown notification so no new fibers should appear from IPROTO when we proceed to cancelling client fibers. Can you describe in more details what happened in the failing tests?

drewdzzz: It seems that replication connection is not proper then - replica can send `IPROTO_SUBSCRIBE` even when master is shutting down, that's why it produces the new client fiber after all such fibers were cancelled. We could try to make it "proper", but it seems to be much more complicated than simply drop all iproto connections before finishing client fibers - then all connections will be handled correctly.

locker (approved):

locker:
@drewdzzz ASAN workflows fail. Please fix.

drewdzzz:
I used `retrying` helper to wait for server to shut down (default value is 5 second) and ASAN tests failed because it wasn't enough. I checked with 20 sec timeout and it worked. Given that I delay the shutdown for about 5 seconds manually in the test and ASAN build works slow, 5 second timeout was just too little.

Now I rewrote the test with `fiber:join` to get rid of timeout at all.
<details>

<summary>diff</summary>

```diff
diff --git a/test/box-luatest/shutdown_test.lua b/test/box-luatest/shutdown_test.lua
index 3d8e7a0aa5..0a23041a03 100644
--- a/test/box-luatest/shutdown_test.lua
+++ b/test/box-luatest/shutdown_test.lua
@@ -248,11 +248,10 @@ g.test_shutdown_with_active_connection = function(cg)
     local socket = socket_connect(cg.server)
 
     -- Start stopping server
-    local stopped = false
-    fiber.create(function()
+    local f = fiber.create(function()
         cg.server:stop()
-        stopped = true
     end)
+    f:set_joinable(true)
 
     -- Wait until Tarantool will cancel client fibers
     local log = fio.pathjoin(cg.server.workdir, cg.server.alias .. '.log')
@@ -267,9 +266,8 @@ g.test_shutdown_with_active_connection = function(cg)
     write_eval(socket, expr)
 
     -- Check if server is successfully stopped
-    t.helpers.retrying({}, function()
-        t.assert(stopped)
-    end)
+    local ok, err = f:join()
+    t.assert(ok, err)
 end
 
 local g_idle_pool = t.group('idle pool')
```

</details>

locker (approved):

PR #10355, opened 2024-08-02
box: don't require functional index function to be sandboxed

### LINKED ISSUE #4765: You can not return box.NULL from functional index
Tarantool version:
2.2.1-10-g1aeaa9e

OS version:
Ubuntu 18.10

Bug description:
box.NULL is not accessible from functional index

Steps to reproduce:

Preparation
```
S = box.schema.create_space('S', {format = {{name='k', type='string'}, {name='x', type='integer'}, {name='y', type='integer'}}})

S:create_index('pk', {parts={1}})

fi_code = [[
    function(tup)
        local xx = tup[2]
        local yy = tup[3]
        if yy <= 5 then
            return {{xx, yy}}
        end
        return {{xx, box.NULL}}
    end
]]

box.schema.func.create('fi_fun', {body = fi_code, is_deterministic = true, is_sandboxed = true, opts = {is_multikey = true}})

S:create_index( 'FI', {
        parts = { {1, 'integer', is_nullable = true}, {2, 'integer', is_nullable = true} },
        func = 'fi_fun',
        unique = false
    } )
```

Reproduction:
```
tarantool> S:insert({'a',1,1})
---
- ['a', 1, 1]
...

tarantool> S:insert({'b',1,2})
---
- ['b', 1, 2]
...

tarantool> S:insert({'c',2,1})
---
- ['c', 2, 1]
...

tarantool> S:insert({'d',1,7})
---
- error: 'Failed to build a key for functional index ''FI'' of space ''S'': [string
    "return     function(tup)..."]:7: attempt to index global ''box'' (a nil value)'
...
```

### LINKED ISSUE #10349: Do not require functional index function to be sandboxed
There's a very strict requirement for a function that can be used in a functional index - apart from being deterministic it also must be executable in a sandbox environment:

https://github.com/tarantool/tarantool/blob/09bfdd473ffd23169dbe8d4e5d0510e8c70b11ec/src/box/alter.cc#L287-L304

A sandbox environment is limited to the following built-in functions:

https://github.com/tarantool/tarantool/blob/09bfdd473ffd23169dbe8d4e5d0510e8c70b11ec/src/box/lua/call.c#L860-L864

This means that a functional index function can't call deterministic functions exported by modules. For example, one can't use `msgpack.encode` for indexing arbitrary-typed fields, as illustrated in the code snippet below.

<details><summary>Example of indexing arbitrary-typed fields</summary>

```lua
local msgpack = require('msgpack')
local varbinary = require('varbinary')

box.cfg{}

box.schema.func.create('test', {
    is_deterministic = true,
    body = [[function(tuple)
        local msgpack = require('msgpack')
        local varbinary = require('varbinary')
        return {varbinary.new(msgpack.encode(tuple.value))}
    end]]
})

box.schema.space.create('test', {
    format = {
        {'key', 'unsigned'},
        {'value', 'any'},
    },
})
box.space.test:create_index('primary')
box.space.test:create_index('value', {
    func = 'test',
    parts = {{1, 'varbinary'}},
})

box.space.test:insert({1, {foo = 'bar'}})
box.space.test:insert({2, {fuzz = 'buzz'}})

local function get(value)
    return box.space.test.index.value:get(varbinary.new(msgpack.encode(value)))
end

print(get({foo = 'bar'}))
print(get({fuzz = 'buzz'}))

os.exit(0)
```

</details>

Another example is #4765: one can't even return `box.NULL` from a functional index function.

We could introduce a whitelist of safe functions and modules that could be called from a functional index function, but maintaining this list up-to-date would be complicated and it wouldn't allow to use user-defined functions so we'd probably end up making this list extendable from Lua, which sounds like a lot of work. So the suggestion is to let the user use a functional index function without a sandbox at their own risk. BTW in Tarantool Enterprise Edition, we [don't require](https://github.com/tarantool/tarantool-ee/blob/1784466e6a8b1a568ca60bedc3b00df73f84cc43/src/box/space_upgrade.c#L436-L446) a space upgrade function to be sandboxed, and we haven't heard any complaints about it so far.

locker:
The requirement for a function used in a functional index to be sandboxed is way too strict because a sandbox environment includes only a handful of Lua built-ins, like `math` or `table`. This makes it impossible for a functional index function to call built-in deterministic functions defined in modules (like `msgpack.encode` or `json.encode`) while this could be useful in certain scenarios (see #10349 for an example). Let's allow users to create functional indexes using non-sandboxed functions at their own risk.

Closes #4765
Closes #10349

### DIFF
--- a/changelogs/unreleased/gh-10349-func-index-without-sandbox.md
+++ b/changelogs/unreleased/gh-10349-func-index-without-sandbox.md
@@ -1,0 +1,4 @@
+## feature/box
+
+* A functional index can now use a function that does not have the flag
+  `is_sandboxed` (gh-10349).
--- a/test/box-luatest/gh_10349_functional_index_without_sandbox_test.lua
+++ b/test/box-luatest/gh_10349_functional_index_without_sandbox_test.lua
@@ -1,0 +1,57 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new()
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.after_each(function(cg)
+    cg.server:exec(function()
+        if box.space.test ~= nil then
+            box.space.test:drop()
+        end
+        box.schema.func.drop('test', {if_exists = true})
+    end)
+end)
+
+g.test_func_index_without_sandbox = function(cg)
+    cg.server:exec(function()
+        local msgpack = require('msgpack')
+        local varbinary = require('varbinary')
+        box.schema.func.create('test', {
+            is_deterministic = true,
+            body = [[function(tuple)
+                local msgpack = require('msgpack')
+                local varbinary = require('varbinary')
+                return {varbinary.new(msgpack.encode(tuple.value))}
+            end]]
+        })
+        local s = box.schema.space.create('test', {
+            format = {
+                {'key', 'unsigned'},
+                {'value', 'any'},
+            },
+        })
+        s:create_index('primary')
+        s:create_index('value', {
+            func = 'test',
+            parts = {{1, 'varbinary'}},
+        })
+        s:insert({1, {foo = 'bar'}})
+        s:insert({2, {fuzz = 'buzz'}})
+        local function get(value)
+            local v = varbinary.new(msgpack.encode(value))
+            return box.space.test.index.value:get(v)
+        end
+        t.assert_equals(get({foo = 'bar'}), {1, {foo = 'bar'}})
+        t.assert_equals(get({fuzz = 'buzz'}), {2, {fuzz = 'buzz'}})
+        t.assert_equals(get({1, 2, 3}), nil)
+    end)
+end
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -293,11 +293,12 @@ static int
 func_index_check_func(struct func *func) {
 	assert(func != NULL);
 	if (func->def->language != FUNC_LANGUAGE_LUA ||
-	    func->def->body == NULL || !func->def->is_deterministic ||
-	    !func->def->is_sandboxed) {
-		diag_set(ClientError, ER_WRONG_INDEX_OPTIONS,
-			 "referenced function doesn't satisfy "
-			 "functional index function constraints");
+	    func->def->body == NULL || !func->def->is_deterministic) {
+		const char *errmsg = tt_sprintf(
+			"function '%s' doesn't meet functional index "
+			"function criteria (stored, deterministic, written in Lua)",
+			func->def->name);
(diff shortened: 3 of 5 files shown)


test/box-luatest/gh_10349_functional_index_without_sandbox_test.lua
```diff
@@ -0,0 +1,71 @@
...
+        local msgpack = require('msgpack')
+        local varbinary = require('varbinary')
+        local function get(value)
+            local v = varbinary.new(msgpack.encode(value))
+            return box.space.test.index.value:get(v)
+        end
+        t.assert_equals(get({foo = 'bar'}), {1, {foo = 'bar'}})
+        t.assert_equals(get({fuzz = 'buzz'}), {2, {fuzz = 'buzz'}})
+        t.assert_equals(get({1, 2, 3}), {3, {1, 2, 3}})
+    end)
```
Totktonada, line 70: What is the idea behind this part of the test case?

I understood it this way: the part above checks that a functional index works correctly for values written during a lifetime of the given process. The quoted part of the test splits writes into a space with a functional index into two parts: the first is written into a snapshot, the second is written into WAL. This way it checks that the functional index is operational for values read from a snapshot or WAL, not only for ones that are written by the process during its lifespan.

OTOH, I don't know, why it needs attention. I guess it is possible that local recovery may construct functional index in a separate code path that has different checks, and the test verifies that both paths allows a non-sandboxed function.

However, it is just guess. I would like if you drop a comment regarding an idea behind it.

Totktonada (approved):
I'm not familiar with the implementation of the functional indexes, so I can't spot some pitfall or a non-trivial peculiar, if any.

The changes looks valid and meaningful and I've no objections.

test/box-luatest/gh_10349_functional_index_without_sandbox_test.lua
locker: I don't know why I added this. Just in case, I guess. Functional indexes are created in the same code path both during recovery and normal operation. We don't have any extra checks that would differ so the code is fully covered by the pre-restart test part. I dropped this confusing part from the test.

Totktonada: OK, I don't mind.

changelogs/unreleased/gh-10349-func-index-without-sandbox.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* A functional index may now use a function that does not have the flag
```
xuniq, line 3:
```suggestion
* A functional index can now use a function that does not have the flag
```

xuniq (approved):

changelogs/unreleased/gh-10349-func-index-without-sandbox.md
locker: Fixed, thanks!

PR #10357, opened 2024-08-02
Make it possible to implement the fast count in the memtx tree read view

mkostoevr:
Part of https://github.com/tarantool/tarantool-ee/issues/858


mkostoevr:
Let's make all this in the #10478.

Issue #10306, opened 2024-07-24
internal: grant access to a global Lua function without writing to a database


Totktonada:
## Context

Tarantool 3.0.0 offers a way to grant access to all or particular global Lua functions without registering them in the database using `box.schema.func.create()` (see #8803 and #9360).

All global Lua functions (except built-in ones):

```lua
box.schema.user.create('alice')
box.schema.user.grant('alice', 'execute', 'lua_call')
```

A particular global Lua function:

```lua
box.schema.user.create('alice')
box.schema.user.grant('alice', 'execute', 'lua_call', 'my_func')
```

However, it still requires the database to be writable[^1] and be upgraded to the last schema (see #7149)[^2].

[^1]: `box.schema.{user,role}.grant()` and `box.schema.{user,role}.revoke()` performs a write to the `_func` system space.
[^2]: A write to a system space (including `_func`) is considered as DDL.

## Problem

It is not always convenient. See examples below.

<details>
<summary>Can't call <code>box.ctl.make_bootstrap_leader</code></summary>

`replication.bootstrap_strategy` = `supervised` requires to call the `box.ctl.make_bootstrap_leader` function to bootstrap a database, but before the database bootstrap it is possible to do only from an admin console and impossible to do via iproto.

We can't command an instance to boostrap a replicaset from an external cluster management tool, for example, the [failover coordinator](https://www.tarantool.io/en/doc/latest/concepts/replication/supervised_failover/#repl-supervised-failover).

This scenario is complicated by a lack of any users before a database bootstrap: a remote agent can only be connected as `guest`.
</details>

<details>
<summary>Can't call <code>failover.execute</code></summary>

`replication.failover` = `supervised` requires the `failover.execute` function to be callable from the failover coordinator via iproto to make an instance writable. However, we need a writable instance to add a permission to call this function. Uroboros.

It makes impossible to start tarantool 3.x on 2.x snapshots with the failover coordinator without preliminary granting the needed permissions on 2.x. It complicates the upgrade scenario for a user.
</details>

## Proposal

Add an internal API to grant a permission to call a global Lua function to a user without a database write.

The API may be, for example, the following:

```lua
box.internal.grant_lua_call('alice')
box.internal.grant_lua_call('alice', 'my_func')

box.internal.revoke_lua_call('alice')
box.internal.revoke_lua_call('alice', 'my_func')
```

A public API for this functionality will be covered by a separate issue.

Part of TNTP-201.

Issue #10310, opened 2024-07-24
config: grant access to call a global Lua function in runtime


Totktonada:
The idea is to allow to grant access to call a global Lua function even if the database is read-only or has an old schema version if the configuration allows it.

In other words, if the configuration has the following snippet:

```yaml
credentials:
  users:
    alice:
      privileges:
        - permissions: [execute]
          lua_call: [my_func]
```

The `my_func` function can be called by the `alice` user via iproto disregadring the database mode/status.

Let's also support `lua_call: [all]` variant in the same way: allow to call all the global Lua functions except built-in ones disregarding the database mode/status.

> [!NOTE]
> I propose to keep writing the privileges to the database when possible (when the database is RW and schema versions is the newest[^1]) for compatibility and consistency with other privileges.

The motivation is to break cycles like the following:

* A bootstrapped database is needed to grant access to a function that assigns a bootstrap leader.
* The RW database mode is needed on some instance in a replicaset to grant access to a function that makes an instance RW.

This task depends on #10304 and #10306.

Part of TNTP-201.

[^1]: DDL is allowed only when the schema version is the newest, see #7149.

sergepetrenko:
> The motivation is to break cycles like the following:
> 
>     * A bootstrapped database is needed to grant access to a function that assigns a bootstrap leader.
> 
>     * The RW database mode is needed on some instance in a replicaset to grant access to a function that makes an instance RW.

Maybe we can make the functions you mention always accessible? IOW, create some whitelist for a specific Tarantool version.

Totktonada:
> Maybe we can make the functions you mention always accessible? IOW, create some whitelist for a specific Tarantool version.

It seems, it is a possible solution. However, I have some doubts regarding it:

* A user may have its own bootstrap leader assignment, RO/RW management, informational functions.
* A user may want to control these privileges from the configuration (including ones for the built-in tarantool functions).
* A user may have application functions that are not related to the database anyhow.
* It is a logical step toward runtime privileges: in my opinion, if we make all the config's privileges runtime from the scratch, it simplifies a lot of things for us and for a user.

PR #10358, opened 2024-08-02
box: add runtime lua_call privileges

mandesero:
This patch adds internal API methods for granting and revoking Lua function access.

New internal API methods:

-  `box.internal.lua_call_runtime_priv_grant(<user-name>, <function-name>)` grants access to the specified function for the specified user.
-  `box.internal.lua_call_runtime_priv_grant(<user-name>, '')` grants universal access (excluding built-in functions) for the specified user.
- `box.internal.lua_call_runtime_priv_reset()` revokes all previously granted function execution permissions for all users.

Closes https://github.com/tarantool/tarantool/issues/10306
Part of https://github.com/tarantool/tarantool/issues/10310

### DIFF
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -6290,6 +6290,7 @@ box_free(void)
 	port_free();
 	iproto_constants_free();
 	mempool_destroy(&sync_trigger_data_pool);
+	box_lua_call_runtime_priv_reset();
 	/* schema_module_free(); */
 	/* session_free(); */
 }
--- a/src/box/call.c
+++ b/src/box/call.c
@@ -45,10 +45,13 @@
 #include "small/rlist.h"
 #include "tt_static.h"
 #include "box/mp_box_ctx.h"
+#include "salad/grp_alloc.h"
 
 struct rlist box_on_call = RLIST_HEAD_INITIALIZER(box_on_call);
 
 static const struct port_vtab port_msgpack_vtab;
+
+static struct mh_strnstrnptr_t *user_rt_access = NULL;
 
 void
 port_msgpack_create_with_ctx(struct port *base, const char *data,
@@ -160,6 +163,10 @@ box_run_on_call(enum iproto_type type, const char *expr, int expr_len,
 	trigger_run(&box_on_call, &ctx);
 }
 
+static bool
+box_lua_call_runtime_priv_is_granted(const char *uname, uint32_t uname_len,
+				     const char *fname, uint32_t fname_len);
+
 int
 access_check_lua_call(const char *name, uint32_t name_len)
 {
@@ -181,10 +188,16 @@ access_check_lua_call(const char *name, uint32_t name_len)
 			return 0;
 	}
 	struct user *user = user_find(cr->uid);
-	if (user != NULL)
+	if (user != NULL) {
+		const char *uname = user->def->name;
+		uint32_t uname_len = strlen(uname);
+		if (box_lua_call_runtime_priv_is_granted(uname, uname_len, name,
+							 name_len))
+			return 0;
 		diag_set(AccessDeniedError, priv_name(PRIV_X),
 			 schema_object_name(SC_FUNCTION),
 			 tt_cstr(name, name_len), user->def->name);
+	}
 	return -1;
 }
 
@@ -272,4 +285,94 @@ box_process_eval(struct call_request *request, struct port *port)
 	int rc = box_lua_eval(expr, expr_len, &args, port);
 	port_msgpack_destroy(&args);
 	return rc;
+}
+
+static struct mh_strnstrnptr_key_t*
+mh_strnstrnptr_key_alloc(const char *s1, uint32_t s1_len,
+			 const char *s2, uint32_t s2_len)
+{
+	struct mh_strnstrnptr_key_t *node;
+	struct grp_alloc all = grp_alloc_initializer();
+	grp_alloc_reserve_data(&all, sizeof(*node) + s1_len + s2_len);
+	grp_alloc_use(&all, xmalloc(grp_alloc_size(&all)));
+	void *p = grp_alloc_create_data(&all, sizeof(*node));
+	node = (struct mh_strnstrnptr_key_t *)p;
+	node->s1 = grp_alloc_create_data(&all, s1_len);
+	memcpy((void *)node->s1, s1, s1_len);
+	node->s1_len = s1_len;
+	if (s2_len != 0) {
+		node->s2 = grp_alloc_create_data(&all, s2_len);
+		memcpy((void *)node->s2, s2, s2_len);
+	} else {
+		node->s2 = NULL;
+	}
+	node->s2_len = s2_len;
+	assert(grp_alloc_size(&all) == 0);
+	return node;
+}
+
+void
+box_lua_call_runtime_priv_reset(void)
+{
+	if (user_rt_access == NULL)
+		return;
+
+	mh_int_t i;
+	mh_foreach(user_rt_access, i) {
+		free(mh_strnstrnptr_node(user_rt_access, i)->val);
+	}
+	mh_strnstrnptr_delete(user_rt_access);
+	user_rt_access = NULL;
+}
+
+void
+box_lua_call_runtime_priv_grant(const char *uname, uint32_t uname_len,
+				const char *fname, uint32_t fname_len)
+{
+	if (user_rt_access == NULL)
+		user_rt_access = mh_strnstrnptr_new();
+
+	/* If the grant exists, do nothing. */
+	uint32_t hash = mh_strnstrnptr_hash(uname, uname_len, fname, fname_len);
+	struct mh_strnstrnptr_key_t access_key =
+		{ uname, uname_len, fname, fname_len, hash };
+	if (fname_len == 0)
+		access_key.s2 = NULL;
+	mh_int_t pos = mh_strnstrnptr_find(user_rt_access, &access_key, NULL);
+	if (pos != mh_end(user_rt_access))
+		return;
+
+	/* Grant access to lua_call function. */
+	struct mh_strnstrnptr_key_t *node =
+		mh_strnstrnptr_key_alloc(uname, uname_len, fname, fname_len);
+	const struct mh_strnstrnptr_node_t access_node =
+		{ node->s1, uname_len, node->s2, fname_len, hash, node };
+	mh_strnstrnptr_put(user_rt_access, &access_node, NULL, NULL);
+}
+
+static bool
+box_lua_call_runtime_priv_is_granted(const char *uname, uint32_t uname_len,
+				     const char *fname, uint32_t fname_len)
+{
+	if (user_rt_access == NULL)
+		return false;
+
+	uint32_t universe_hash = mh_strnstrnptr_hash(uname, uname_len, NULL, 0);
+	struct mh_strnstrnptr_key_t access_universe_key =
+		{ uname, uname_len, NULL, 0, universe_hash };
+	mh_int_t pos = mh_strnstrnptr_find(user_rt_access, &access_universe_key,
+					   NULL);
+
+	/* Check for universe access. Don't allow to call built-ins. */
+	if (pos != mh_end(user_rt_access) &&
+	    !tarantool_lua_is_builtin_global(fname, fname_len))
+		return true;
+
+	uint32_t hash = mh_strnstrnptr_hash(uname, uname_len, fname, fname_len);
+	struct mh_strnstrnptr_key_t access_node =
+		{ uname, uname_len, fname, fname_len, hash };
+	pos = mh_strnstrnptr_find(user_rt_access, &access_node, NULL);
+	if (pos == mh_end(user_rt_access))
+		return false;
+	return true;
 }
--- a/src/box/lua/call.c
+++ b/src/box/lua/call.c
@@ -1268,6 +1268,41 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
 	return 0;
 }
 
+static int
+lbox_box_lua_call_runtime_priv_reset(struct lua_State *L)
+{
+	if (lua_gettop(L) != 0) {
+		diag_set(IllegalParams,
+			 "Usage: box.internal.lua_call_runtime_priv_reset()");
+		return luaT_error(L);
+	}
+
+	box_lua_call_runtime_priv_reset();
+	return 0;
+}
+
+static int
+lbox_box_lua_call_runtime_priv_grant(struct lua_State *L)
+{
+	if (lua_gettop(L) != 2 || lua_type(L, 1) != LUA_TSTRING ||
+	    lua_type(L, 2) != LUA_TSTRING) {
+		diag_set(IllegalParams,
+			 "Usage: box.internal.lua_call_runtime_priv_grant(user, func)");
+		return luaT_error(L);
+	}
+
+	size_t grantee_name_len;
+	const char *grantee_name = luaL_checklstring(L, 1, &grantee_name_len);
+
+	size_t func_name_len;
+	const char *func_name = luaL_checklstring(L, 2, &func_name_len);
+
+	box_lua_call_runtime_priv_grant(grantee_name,
+					(uint32_t)grantee_name_len,
+					func_name, (uint32_t)func_name_len);
+	return 0;
+}
+
 static void
 call_serializer_update_options(void)
 {
@@ -1291,6 +1326,8 @@ static const struct luaL_Reg boxlib_internal[] = {
 	{"call_loadproc",  lbox_call_loadproc},
 	{"module_reload", lbox_module_reload},
 	{"func_call", lbox_func_call},
+	{"lua_call_runtime_priv_grant", lbox_box_lua_call_runtime_priv_grant},
+	{"lua_call_runtime_priv_reset", lbox_box_lua_call_runtime_priv_reset},
 	{NULL, NULL}
 };
 
--- a/src/lib/core/assoc.h
+++ b/src/lib/core/assoc.h
@@ -209,6 +209,69 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
 	return mh_strnu32_find(h, &key, NULL);
 }
 
+/*
+ * Map: (const char *, uint32_t, const char *, uint32_t) => (void *)
+ */
+
+#define mh_name _strnstrnptr
+/**
+ * Key of `mh_strnstrnptr_node_t` hash table.
+ */
+struct mh_strnstrnptr_key_t {
+	/* First string. */
+	const char *s1;
+	/* First string length. */
+	uint32_t s1_len;
+	/* Second string. */
+	const char *s2;
+	/* Second string length. */
+	uint32_t s2_len;
+	/* Key hash calculated using `mh_strnstrnptr_hash`. */
+	uint32_t hash;
+};
+
+/**
+ * Node of `mh_strnstrnptr_node_t` hash table.
+ */
+struct mh_strnstrnptr_node_t {
+	/* First string. */
+	const char *s1;
+	/* First string length. */
+	uint32_t s1_len;
+	/* Second string. */
+	const char *s2;
+	/* Second string length. */
+	uint32_t s2_len;
+	/* Key hash calculated using `mh_strnstrnptr_hash`. */
+	uint32_t hash;
+	/* Mapped value. */
+	void *val;
+};
+
+#define mh_key_t struct mh_strnstrnptr_key_t *
+#define mh_node_t struct mh_strnstrnptr_node_t
+
+#define mh_arg_t void *
+#define mh_hash(a, arg) ((a)->hash)
+#define mh_hash_key(a, arg) ((a)->hash)
+#define mh_cmp(a, b, arg) ((a)->s1_len != (b)->s1_len || \
+			   (a)->s2_len != (b)->s2_len || \
+			   memcmp((a)->s1, (b)->s1, (a)->s1_len) || \
+			   memcmp((a)->s2, (b)->s2, (a)->s2_len))
+#define mh_cmp_key(a, b, arg) mh_cmp(a, b, arg)
+#include "salad/mhash.h"
+
+static inline uint32_t
+mh_strnstrnptr_hash(const char *s1, uint32_t s1_len,
+		    const char *s2, uint32_t s2_len)
+{
+	uint32_t h = MH_STRN_HASH_SEED;
+	uint32_t carry = 0;
+	PMurHash32_Process(&h, &carry, s1, s1_len);
+	PMurHash32_Process(&h, &carry, s2, s2_len);
+	return PMurHash32_Result(h, carry, s1_len + s2_len);
+};
+
 #if defined(__cplusplus)
 } /* extern "C" */
 #endif /* defined(__cplusplus) */
--- a/test/box-luatest/gh_10306_lua_call_runtime_access_test.lua
+++ b/test/box-luatest/gh_10306_lua_call_runtime_access_test.lua
@@ -1,0 +1,152 @@
+local t = require('luatest')
+local cbuilder = require('luatest.cbuilder')
+local cluster = require('test.config-luatest.cluster')
+
+local g = t.group()
+
+local config = cbuilder:new()
+    :add_instance('i-001', {})
+    :set_global_option('credentials.users.alice')
+    :set_global_option('credentials.users.alice.password', 'ALICE')
+    :config()
+
+g.before_all(cluster.init)
+g.after_each(cluster.drop)
+g.after_all(cluster.clean)
+
+local function access_error_msg(user, func)
+    local templ = 'Execute access to function \'%s\' is denied for user \'%s\''
+    return string.format(templ, func, user)
+end
+
+local function define_access_error_msg_function(server)
+    server:exec(function(access_error_msg)
+        rawset(_G, 'access_error_msg', loadstring(access_error_msg))
+    end, {string.dump(access_error_msg)})
+end
+
+g.test_lua_call_runtime_priv_grant = function(g)
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    define_access_error_msg_function(cluster['i-001'])
+    cluster['i-001']:exec(function()
+        local netbox = require('net.box')
+        local config = require('config')
+        local uri = config:instance_uri().uri
+        local con = netbox.connect(uri, {user='alice', password='ALICE'})
+
+        rawset(_G, 'foo', function() return true end)
+        rawset(_G, 'bar', function() return true end)
+
+        -- Try to call 'foo' for user `alice`. Check that user `alice` does not
+        -- have privileges to call `foo`.
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'foo'),
+                                  function() con:call('foo') end)
+
+        -- Grant access to call function `foo` for user `alice`.
+        box.internal.lua_call_runtime_priv_grant('alice', 'foo')
+        t.assert(con:call('foo'))
+
+        -- Check that user `alice` does not have permission to call functions
+        -- other than `foo`.
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'bar'),
+                                  function() con:call('bar') end)
+    end)
+end
+
+g.test_lua_call_runtime_priv_reset = function(g)
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    define_access_error_msg_function(cluster['i-001'])
+    cluster['i-001']:exec(function()
+        local netbox = require('net.box')
+        local config = require('config')
+        local uri = config:instance_uri().uri
+        local con = netbox.connect(uri, {user='alice', password='ALICE'})
+
+        rawset(_G, 'foo', function() return true end)
+        rawset(_G, 'bar', function() return true end)
+
+        -- Grant access to call functions `foo` and `bar` for user `alice`.
+        box.internal.lua_call_runtime_priv_grant('alice', 'foo')
+        box.internal.lua_call_runtime_priv_grant('alice', 'bar')
+
+        t.assert(con:call('foo'))
+        t.assert(con:call('bar'))
+
+        -- Reset access to call lua functions. Verify that user `alice` does not
+        -- have privileges to call any functions.
+        box.internal.lua_call_runtime_priv_reset()
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'foo'),
+                                  function() con:call('foo') end)
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'bar'),
+                                  function() con:call('bar') end)
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'baz'),
+                                  function() con:call('baz') end)
+    end)
+end
+
+g.test_grant_universe_lua_call = function(g)
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    define_access_error_msg_function(cluster['i-001'])
+    cluster['i-001']:exec(function()
+        local netbox = require('net.box')
+        local config = require('config')
+        local uri = config:instance_uri().uri
+        local con = netbox.connect(uri, {user='alice', password='ALICE'})
+
+        rawset(_G, 'foo', function() return true end)
+        rawset(_G, 'bar', function() return true end)
+
+        -- Grant universe access to user `alice`.
+        box.internal.lua_call_runtime_priv_grant('alice', '')
+
+        -- Universe access gives privileges to call any non-builtins functions.
+        t.assert(con:call('foo'))
+        t.assert(con:call('bar'))
+
+        -- Universe access does not allow calling built-in functions.
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'box.info'),
+                                  function() con:call('box.info') end)
+    end)
+end
+
+g.test_grant_builtins_lua_call = function(g)
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    define_access_error_msg_function(cluster['i-001'])
+    cluster['i-001']:exec(function()
+        local netbox = require('net.box')
+        local config = require('config')
+        local uri = config:instance_uri().uri
+        local con = netbox.connect(uri, {user='alice', password='ALICE'})
+
+        t.assert_error_msg_equals(_G.access_error_msg('alice', 'box.info'),
+                                  function() con:call('box.info') end)
+
+        -- Grant access to call builtins for user `alice`.
+        box.internal.lua_call_runtime_priv_grant('alice', 'box.info')
+
+        -- Check that user `alice` can call built-in functions.
+        t.assert_equals(con:call('box.info'), box.info())
+    end)
+end
+
+g.test_lua_call_runtime_priv_invalid_usage = function(g)
+    local cluster = cluster.new(g, config)
+    cluster:start()
+    define_access_error_msg_function(cluster['i-001'])
+    cluster['i-001']:exec(function()
+        local msg
+        msg = 'Usage: box.internal.lua_call_runtime_priv_grant(user, func)'
+        t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_grant)
+        t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_grant,
+                                  'alice')
+        t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_grant,
+                                  5, 'foo')
+        msg = 'Usage: box.internal.lua_call_runtime_priv_reset()'
+        t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_reset,
+                                  'alice')
+    end)
+end
--- a/src/box/call.h
+++ b/src/box/call.h
@@ -80,6 +80,23 @@ box_process_eval(struct call_request *request, struct port *port);
 int
 access_check_lua_call(const char *name, uint32_t name_len);
 
+/**
+ * Reset the user runtime privileges.
+ */
+void
+box_lua_call_runtime_priv_reset(void);
+
+/**
+ * Grant a access to lua_call function for a user.
+ * @param uname Name of the user.
+ * @param uname_len Length of the user name.
+ * @param fname Name of the function.
+ * @param fname_len Length of the function name.
+ */
+void
+box_lua_call_runtime_priv_grant(const char *uname, uint32_t uname_len,
+				const char *fname, uint32_t fname_len);
+
 #if defined(__cplusplus)
 } /* extern "C" */
 #endif /* defined(__cplusplus) */


src/box/call_rt_user_privilege.c
```diff
@@ -0,0 +1,167 @@
...
+		struct mh_i32ptr_node_t repl_id_node = {0, NULL};
+		struct mh_i32ptr_node_t *prepl_id_node = &repl_id_node;
+		mh_i32ptr_put(user_runtime_universe, &id_node, &prepl_id_node,
+			      NULL);
+		return;
+	}
+
+	/*  Don't allow to grant built-ins.  */
+	if (tarantool_lua_is_builtin_global(func_name, func_name_len))
+		return;
```
Totktonada, line 41: It is OK to grant a built-in function. The idea is that granting all the functions (using `'all'` in config or `''` in Lua) shouldn't give rights to execute built-in ones.

Quote from PR #9373:

> If specified, the privilege will be applied only to the given function name, which may be a built-in <...>

mandesero: Now grant_lua_call(<user>, '') allow to call all the global Lua functions except built-in ones.
The only way to call a built-in function is to explicitly grant access to that function.

changelogs/unreleased/gh_10306_lua_call_runtime_access.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* Add the ability to grant and revoke access to any `lua_call` function in the
+  internal API (gh-10306).
```
xuniq, line 4:
```suggestion
* It is now possible to grant and revoke access to any `lua_call` function in the
  internal API (gh-10306).
```

xuniq (approved):

src/lib/core/assoc.h
```diff
@@ -189,6 +189,76 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
...
+	const char *fname;
+	/* Key function name length. */
+	uint32_t fname_len;
+	/* Key hash calculated using `mh_access_hash`. */
+	uint32_t hash;
+	/* Mapped value. */
+	int val;
+};
+
+#define mh_node_t struct mh_access_node_t
```
Totktonada, line 241: It seems, we don't need a separate key/node structs. It is enough for us to have a set, without a mapped value.

At least, the following change seems to work.

```diff
diff --git a/src/box/rt_lua_call_access.c b/src/box/rt_lua_call_access.c
index d17fc63eb6..0fc486bf83 100644
--- a/src/box/rt_lua_call_access.c
+++ b/src/box/rt_lua_call_access.c
@@ -50,11 +50,9 @@ grant_rt_access(const char *uname, uint32_t uname_len,
 	/* Grant access to lua_call function. */
 	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
 	const struct mh_access_node_t access_node =
-		{ uname, uname_len, fname, fname_len, hash, RT_ACCESS };
-	struct mh_access_node_t repl_access_node = { NULL, 0, NULL, 0, 0, 0 };
-	struct mh_access_node_t *prepl_access_node = &repl_access_node;
-	mh_access_put(user_rt_access_cache, &access_node, &prepl_access_node,
-		      NULL);
+		{ uname, uname_len, fname, fname_len, hash };
+	struct mh_access_node_t *ignored;
+	mh_access_put(user_rt_access_cache, &access_node, &ignored, NULL);
 }
 
 void
@@ -72,7 +70,7 @@ revoke_rt_access(const char *uname, uint32_t uname_len,
 		/* Revoke access to lua_call function. */
 		uint32_t hash = mh_access_hash(uname, uname_len, fname,
 					       fname_len);
-		struct mh_access_key_t key =
+		struct mh_access_node_t key =
 			{ uname, uname_len, fname, fname_len, hash };
 		mh_int_t pos = mh_access_find(user_rt_access_cache, &key, NULL);
 		if (pos != mh_end(user_rt_access_cache))
@@ -92,7 +90,7 @@ check_rt_access(const char *uname, uint32_t uname_len,
 		return RT_ACCESS;
 
 	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
-	struct mh_access_key_t key =
+	struct mh_access_node_t key =
 		{ uname, uname_len, fname, fname_len, hash };
 	pos = mh_access_find(user_rt_access_cache, &key, NULL);
 	if (pos == mh_end(user_rt_access_cache))
diff --git a/src/lib/core/assoc.h b/src/lib/core/assoc.h
index 2b2b0ec310..5f63a47776 100644
--- a/src/lib/core/assoc.h
+++ b/src/lib/core/assoc.h
@@ -190,7 +190,7 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
 }
 
 /*
- * Map: (const char *, uint32_t, const char *, uint32_t) => (void *)
+ * Set: (const char *, uint32_t, const char *, uint32_t)
  */
 
 static inline uint32_t
@@ -202,24 +202,6 @@ mh_access_hash(const char *uname, uint32_t uname_len,
 
 #define mh_name _access
 
-/**
- * Key of `mh_access_node_t` hash table.
- */
-struct mh_access_key_t {
-	/* Key user name. */
-	const char *uname;
-	/* Key user name length. */
-	uint32_t uname_len;
-	/* Key function name. */
-	const char *fname;
-	/* Key function name length. */
-	uint32_t fname_len;
-	/* Key hash calculated using `mh_access_hash`. */
-	uint32_t hash;
-};
-
-#define mh_key_t struct mh_access_key_t *
-
 /**
  * Node of `mh_access_node_t` hash table.
  */
@@ -234,10 +216,9 @@ struct mh_access_node_t {
 	uint32_t fname_len;
 	/* Key hash calculated using `mh_access_hash`. */
 	uint32_t hash;
-	/* Mapped value. */
-	int val;
 };
 
+#define mh_key_t struct mh_access_node_t *
 #define mh_node_t struct mh_access_node_t
 
 #define mh_arg_t void *
@@ -255,7 +236,7 @@ mh_access_find_str(struct mh_access_t *h, const char *uname, uint32_t uname_len,
 		   const char *fname, uint32_t fname_len)
 {
 	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
-	struct mh_access_key_t key = {uname, uname_len, fname, fname_len, hash};
+	struct mh_access_node_t key = {uname, uname_len, fname, fname_len, hash};
 	return mh_access_find(h, &key, NULL);
 }
 
```

src/box/rt_lua_call_access.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <zlib.h>
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
```
Totktonada, line 8: `zlib` is not used here, Lua function too. And there are no booleans. Please, drop the unused includes.

Also, I see that the `.c` file doesn't use anything like `diag_set`, so `diag.h` include is redundant too.

```diff
diff --git a/src/box/rt_lua_call_access.c b/src/box/rt_lua_call_access.c
index 0fc486bf83..1bf4904713 100644
--- a/src/box/rt_lua_call_access.c
+++ b/src/box/rt_lua_call_access.c
@@ -1,5 +1,4 @@
 #include "rt_lua_call_access.h"
-#include "diag.h"
 #include "assoc.h"
 #include "user.h"
 #include "session.h"
diff --git a/src/box/rt_lua_call_access.h b/src/box/rt_lua_call_access.h
index f5fc371aec..c427f29c11 100644
--- a/src/box/rt_lua_call_access.h
+++ b/src/box/rt_lua_call_access.h
@@ -1,11 +1,6 @@
 #pragma once
 
 #include <stdint.h>
-#include <stdbool.h>
-#include <zlib.h>
-#include <lua.h>
-#include <lauxlib.h>
-#include <lualib.h>
 
 #if defined(__cplusplus)
 extern "C" {
```

src/box/lua/call.c
```diff
@@ -1268,6 +1269,56 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
...
+	size_t func_name_len;
+	const char *func_name = luaL_checklstring(L, 2, &func_name_len);
+
+	grant_rt_access(grantee_name,  (uint32_t)grantee_name_len,
+			func_name, (uint32_t)func_name_len);
+	return 0;
+}
+
+static int
+lbox_revoke_rt_access(struct lua_State *L)
```
Totktonada, line 1298: AFAIU, it is not used in the final version of the #10392 patch. It means that is it a dead code. I suggest to drop it.

At the same time #10392 adds `rt_lua_call_access_reset()`. It seems logical to add it in this patch instead.

src/box/lua/call.c
```diff
@@ -1268,6 +1269,56 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
...
+	if (lua_gettop(L) != 2)
+		return luaL_error(L, "internal.grant_lua_call bad arguments. "
+				  "Expected exactly two: user name and "
+				  "function name.");
+
+	struct credentials *grantor = effective_user();
+	struct user *grantor_user = user_find(grantor->uid);
+	if (grantor->uid != ADMIN && is_super(grantor_user) == 0)
+		return luaL_error(L, "internal.grant_lua_call only ADMIN or "
+				  "user with role SUPER can grant access.");
```
Totktonada, line 1284: This check is inconsistent with permissions to change permissions in the `_priv` system space.

https://github.com/tarantool/tarantool/blob/8947cb04f59423e2944d48b8a1effec2fb11b1db/src/box/alter.cc#L3955-L3961

I don't really know, whether it is good or bad to forbid such actions for the `super` role, but I'm for consistency.

src/box/CMakeLists.txt
```diff
@@ -275,6 +275,7 @@ set(box_sources
     sql_stmt_cache.c
     wal.c
     call.c
+    rt_lua_call_access.c
```
Totktonada, line 278: How about just `rt_access.c`? It seems, it fits good and it is concise.

src/box/rt_lua_call_access.h
```diff
@@ -0,0 +1,70 @@
...
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+#define is_super(u) ((u)->roles.m[0] == 20u)
+#define is_universe(s, l) ((l) == 0 || (s)[0] == '\0')
+#define RT_ACCESS 1
+#define RT_NO_ACCESS 0
```
Totktonada, line 17: AFAIS, we have two general approaches in tarantool:

1. Use `bool` return value.
2. Use `int`, where 0 is success and -1 (with setting a diagnostics area) is an error.

Since we don't set a diag here, I would use booleans.

src/box/rt_lua_call_access.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Grant a access to lua_call function for a user.
+ * @param uname Name of the user.
+ * @param uname_len Length of the user name.
+ * @param fname Name of the function.
+ * @param fname_len Length of the function name.
+ */
+void
+grant_rt_access(const char *uname, uint32_t uname_len,
+		const char *fname, uint32_t fname_len);
```
Totktonada, line 43: We generally prepent a function name with its compilation unit name.

changelogs/unreleased/gh-10306-lua-call-runtime-access.md
```diff
@@ -0,0 +1,4 @@
+## feature/box
+
+* It is now possible to grant and revoke access to any `lua_call` function
+  in the internal API (gh-10306).
```
Totktonada, line 4: I think we shouldn't advertize a use of an internal API anyhow, including information in the release notes. Let's drop the changelog entry.

src/lib/core/assoc.h
```diff
@@ -189,6 +189,76 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
...
+
+#define mh_arg_t void *
+#define mh_hash(a, arg) ((a)->hash)
+#define mh_hash_key(a, arg) ((a)->hash)
+#define mh_cmp(a, b, arg) ((a)->uname_len != (b)->uname_len || \
+			    (a)->fname_len != (b)->fname_len || \
+			    strncmp((a)->uname, (b)->uname, (a)->uname_len) || \
+			    strncmp((a)->fname, (b)->fname, (a)->fname_len))
```
Totktonada, line 249: We shouldn't assume that the strings are ended with `'\0'`, because there is a provided length. Let's use `memcmp()` instead.

src/lib/core/assoc.h
```diff
@@ -189,6 +189,76 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
...
+ */
+
+static inline uint32_t
+mh_access_hash(const char *uname, uint32_t uname_len,
+	       const char *fname, uint32_t fname_len)
+{
+	return mh_strn_hash(uname, uname_len) ^ mh_strn_hash(fname, fname_len);
+}
+
+#define mh_name _access
```
Totktonada, line 203: In my understanding the `assoc.h` file provides general-purpose containers. Taking it into account, I would name the hash and the related functions/structures with something derived from the 'two strings with length' idea, without any relation to the accesses.

Something like the following (I use `strn_strn` name here):

```diff
diff --git a/src/box/rt_lua_call_access.c b/src/box/rt_lua_call_access.c
index 1bf4904713..d0f43c4bfc 100644
--- a/src/box/rt_lua_call_access.c
+++ b/src/box/rt_lua_call_access.c
@@ -4,7 +4,7 @@
 #include "session.h"
 #include "../lua/init.h" /* tarantool_lua_is_builtin_global */
 
-static struct mh_access_t *user_rt_access_cache = NULL;
+static struct mh_strn_strn_t *user_rt_access_cache = NULL;
 static struct mh_strnptr_t *user_rt_universe_access_cache = NULL;
 
 const char *
@@ -19,14 +19,14 @@ user_name_by_id(uint32_t uid, uint32_t *name_len)
 void
 rt_lua_call_access_init(void)
 {
-	user_rt_access_cache = mh_access_new();
+	user_rt_access_cache = mh_strn_strn_new();
 	user_rt_universe_access_cache = mh_strnptr_new();
 }
 
 void
 rt_lua_call_access_destroy(void)
 {
-	mh_access_delete(user_rt_access_cache);
+	mh_strn_strn_delete(user_rt_access_cache);
 	mh_strnptr_delete(user_rt_universe_access_cache);
 }
 
@@ -39,19 +39,18 @@ grant_rt_access(const char *uname, uint32_t uname_len,
 		uint32_t hash = mh_strn_hash(uname, uname_len);
 		const struct mh_strnptr_node_t uid_node =
 			{ uname, uname_len, hash, NULL };
-		struct mh_strnptr_node_t repl_uid_node = { NULL, 0, 0, NULL };
-		struct mh_strnptr_node_t *prepl_uid_node = &repl_uid_node;
+		struct mh_strnptr_node_t *ignored;
 		mh_strnptr_put(user_rt_universe_access_cache, &uid_node,
-			       &prepl_uid_node, NULL);
+			       &ignored, NULL);
 		return;
 	}
 
 	/* Grant access to lua_call function. */
-	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
-	const struct mh_access_node_t access_node =
+	uint32_t hash = mh_strn_strn_hash(uname, uname_len, fname, fname_len);
+	const struct mh_strn_strn_node_t access_node =
 		{ uname, uname_len, fname, fname_len, hash };
-	struct mh_access_node_t *ignored;
-	mh_access_put(user_rt_access_cache, &access_node, &ignored, NULL);
+	struct mh_strn_strn_node_t *ignored;
+	mh_strn_strn_put(user_rt_access_cache, &access_node, &ignored, NULL);
 }
 
 void
@@ -67,13 +66,13 @@ revoke_rt_access(const char *uname, uint32_t uname_len,
 				       NULL);
 	} else {
 		/* Revoke access to lua_call function. */
-		uint32_t hash = mh_access_hash(uname, uname_len, fname,
+		uint32_t hash = mh_strn_strn_hash(uname, uname_len, fname,
 					       fname_len);
-		struct mh_access_node_t key =
+		struct mh_strn_strn_node_t key =
 			{ uname, uname_len, fname, fname_len, hash };
-		mh_int_t pos = mh_access_find(user_rt_access_cache, &key, NULL);
+		mh_int_t pos = mh_strn_strn_find(user_rt_access_cache, &key, NULL);
 		if (pos != mh_end(user_rt_access_cache))
-			mh_access_del(user_rt_access_cache, pos, NULL);
+			mh_strn_strn_del(user_rt_access_cache, pos, NULL);
 	}
 }
 
@@ -88,10 +87,10 @@ check_rt_access(const char *uname, uint32_t uname_len,
 	    !tarantool_lua_is_builtin_global(fname, fname_len))
 		return RT_ACCESS;
 
-	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
-	struct mh_access_node_t key =
+	uint32_t hash = mh_strn_strn_hash(uname, uname_len, fname, fname_len);
+	struct mh_strn_strn_node_t key =
 		{ uname, uname_len, fname, fname_len, hash };
-	pos = mh_access_find(user_rt_access_cache, &key, NULL);
+	pos = mh_strn_strn_find(user_rt_access_cache, &key, NULL);
 	if (pos == mh_end(user_rt_access_cache))
 		return RT_NO_ACCESS;
 	return RT_ACCESS;
diff --git a/src/lib/core/assoc.h b/src/lib/core/assoc.h
index 5f63a47776..a30f2c36d5 100644
--- a/src/lib/core/assoc.h
+++ b/src/lib/core/assoc.h
@@ -194,50 +194,50 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
  */
 
 static inline uint32_t
-mh_access_hash(const char *uname, uint32_t uname_len,
-	       const char *fname, uint32_t fname_len)
+mh_strn_strn_hash(const char *s1, uint32_t s1_len,
+		  const char *s2, uint32_t s2_len)
 {
-	return mh_strn_hash(uname, uname_len) ^ mh_strn_hash(fname, fname_len);
+	return mh_strn_hash(s1, s1_len) ^ mh_strn_hash(s2, s2_len);
 }
 
-#define mh_name _access
+#define mh_name _strn_strn
 
 /**
- * Node of `mh_access_node_t` hash table.
+ * Node of `mh_strn_strn_node_t` hash table.
  */
-struct mh_access_node_t {
-	/* Key user name. */
-	const char *uname;
-	/* Key user name length. */
-	uint32_t uname_len;
-	/* Key function name. */
-	const char *fname;
+struct mh_strn_strn_node_t {
+	/* First string. */
+	const char *s1;
+	/* First string length. */
+	uint32_t s1_len;
+	/* Second string. */
+	const char *s2;
 	/* Key function name length. */
-	uint32_t fname_len;
-	/* Key hash calculated using `mh_access_hash`. */
+	uint32_t s2_len;
+	/* Key hash calculated using `mh_strn_strn_hash`. */
 	uint32_t hash;
 };
 
-#define mh_key_t struct mh_access_node_t *
-#define mh_node_t struct mh_access_node_t
+#define mh_key_t struct mh_strn_strn_node_t *
+#define mh_node_t struct mh_strn_strn_node_t
 
 #define mh_arg_t void *
 #define mh_hash(a, arg) ((a)->hash)
 #define mh_hash_key(a, arg) ((a)->hash)
-#define mh_cmp(a, b, arg) ((a)->uname_len != (b)->uname_len || \
-			    (a)->fname_len != (b)->fname_len || \
-			    strncmp((a)->uname, (b)->uname, (a)->uname_len) || \
-			    strncmp((a)->fname, (b)->fname, (a)->fname_len))
+#define mh_cmp(a, b, arg) ((a)->s1_len != (b)->s1_len || \
+			    (a)->s2_len != (b)->s2_len || \
+			    memcmp((a)->s1, (b)->s1, (a)->s1_len) || \
+			    memcmp((a)->s2, (b)->s2, (a)->s2_len))
 #define mh_cmp_key(a, b, arg) mh_cmp(a, b, arg)
 #include "salad/mhash.h"
 
 static inline mh_int_t
-mh_access_find_str(struct mh_access_t *h, const char *uname, uint32_t uname_len,
-		   const char *fname, uint32_t fname_len)
+mh_strn_strn_find_str(struct mh_strn_strn_t *h, const char *s1, uint32_t s1_len,
+		      const char *s2, uint32_t s2_len)
 {
-	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
-	struct mh_access_node_t key = {uname, uname_len, fname, fname_len, hash};
-	return mh_access_find(h, &key, NULL);
+	uint32_t hash = mh_strn_strn_hash(s1, s1_len, s2, s2_len);
+	struct mh_strn_strn_node_t key = {s1, s1_len, s2, s2_len, hash};
+	return mh_strn_strn_find(h, &key, NULL);
 }
 
 #if defined(__cplusplus)
```

src/box/rt_lua_call_access.h
```diff
@@ -0,0 +1,70 @@
...
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+#define is_super(u) ((u)->roles.m[0] == 20u)
+#define is_universe(s, l) ((l) == 0 || (s)[0] == '\0')
```
Totktonada, line 15: I would use an in-place `length == 0` check without any extra data lookups.

src/box/rt_lua_call_access.c
```diff
@@ -0,0 +1,101 @@
+#include "rt_lua_call_access.h"
+#include "diag.h"
+#include "assoc.h"
+#include "user.h"
+#include "session.h"
+#include "../lua/init.h" /* tarantool_lua_is_builtin_global */
+
+static struct mh_access_t *user_rt_access_cache = NULL;
+static struct mh_strnptr_t *user_rt_universe_access_cache = NULL;
```
Totktonada, line 9: These hash maps are not caches, they are the primary source of information about the runtime privileges. I think that the current `*_cache` name may be confusing.

src/box/rt_lua_call_access.c
```diff
@@ -0,0 +1,101 @@
...
+{
+	/* Grant universe access. */
+	if (is_universe(fname, fname_len)) {
+		uint32_t hash = mh_strn_hash(uname, uname_len);
+		const struct mh_strnptr_node_t uid_node =
+			{ uname, uname_len, hash, NULL };
+		struct mh_strnptr_node_t repl_uid_node = { NULL, 0, 0, NULL };
+		struct mh_strnptr_node_t *prepl_uid_node = &repl_uid_node;
+		mh_strnptr_put(user_rt_universe_access_cache, &uid_node,
+			       &prepl_uid_node, NULL);
```
Totktonada, line 46: May be simplified to:

```c
struct mh_strnptr_node_t *ignored = NULL;
mh_strnptr_put(<...>, &ignored, <...>);
```

And the same below.

src/lib/core/assoc.h
```diff
@@ -189,6 +189,76 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
...
+#define mh_cmp(a, b, arg) ((a)->uname_len != (b)->uname_len || \
+			    (a)->fname_len != (b)->fname_len || \
+			    strncmp((a)->uname, (b)->uname, (a)->uname_len) || \
+			    strncmp((a)->fname, (b)->fname, (a)->fname_len))
+#define mh_cmp_key(a, b, arg) mh_cmp(a, b, arg)
+#include "salad/mhash.h"
+
+static inline mh_int_t
+mh_access_find_str(struct mh_access_t *h, const char *uname, uint32_t uname_len,
+		   const char *fname, uint32_t fname_len)
```
Totktonada, line 255: The `<...>_find_str()` suffix looks confusing. We're searching for two strings, it is different from `mh_strnptr_find_str()` in this regard.

Totktonada: Hm. AFAIS, this function is not used at all. Let's drop it?

src/box/rt_lua_call_access.c
```diff
@@ -0,0 +1,101 @@
...
+		return;
+	}
+
+	/* Grant access to lua_call function. */
+	uint32_t hash = mh_access_hash(uname, uname_len, fname, fname_len);
+	const struct mh_access_node_t access_node =
+		{ uname, uname_len, fname, fname_len, hash, RT_ACCESS };
```
Totktonada, line 53: It may be convenient to have a `mh_strn_strn_node_create()` function that calculates the hash:

```c
struct mh_strn_strn_node_t access_node;
mh_strn_strn_node_create(&access_node, uname, uname_len, fname, fname_len);
```

src/lib/core/assoc.h
mandesero: Dropped.

src/box/rt_lua_call_access.c
mandesero: Done.

src/box/rt_lua_call_access.c
mandesero: Removed postfix `*_cache`.

src/box/rt_lua_call_access.h
```diff
@@ -0,0 +1,70 @@
...
+#define RT_ACCESS 1
+#define RT_NO_ACCESS 0
+
+const char *
+user_name_by_id(uint32_t uid, uint32_t *name_len);
```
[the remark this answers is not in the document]
mandesero: Moved to `src/box/user.[h;cc]`

src/box/rt_lua_call_access.h
mandesero: Done.

src/lib/core/assoc.h
mandesero: Renamed to `*_strn_strn_*`

src/lib/core/assoc.h
mandesero: Done.

changelogs/unreleased/gh-10306-lua-call-runtime-access.md
mandesero: Changelog removed.

src/box/rt_lua_call_access.h
mandesero: renamed.

src/box/rt_lua_call_access.h
mandesero: I decided not to use `RT_ACCESS` and others, but to use `bool` directly in `rt_access_check`.

src/box/CMakeLists.txt
mandesero: renamed.

src/box/lua/call.c
mandesero: now only `admin` can grant/revoke access.

src/box/lua/call.c
mandesero: `rt_lua_call_access_reset` added, `lbox_revoke_rt_access` removed.

src/box/rt_lua_call_access.h
mandesero: Done.

src/lib/core/assoc.h
mandesero: Done.

src/box/lua/call.c
```diff
@@ -1268,6 +1269,40 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
...
 
+static int
+lbox_rt_access_reset(struct lua_State *L)
+{
+	if (lua_gettop(L) != 0)
+		return luaL_error(L, "internal.reset_lua_call takes no "
+				  "arguments, but %d given.", lua_gettop(L));
+	rt_access_reset();
+	return 0;
+}
```
Totktonada, line 1280: It seems, we should do the same `uid == ADMIN` check here as in `grant_lua_call`.

src/box/rt_access.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif /* defined(__cplusplus) */
+
+const char *
+user_name_by_id(uint32_t uid, uint32_t *name_len);
```
Totktonada, line 11: It seems, you forgot to remove the declaration. It duplicates one from `user.h`.

src/box/rt_lua_call_access.c
Totktonada: I see no problem (if we do it after all the needed declarations):

```diff
diff --git a/src/box/rt_access.c b/src/box/rt_access.c
index acfd29a2c4..a1417039d6 100644
--- a/src/box/rt_access.c
+++ b/src/box/rt_access.c
@@ -5,19 +5,6 @@
 static struct mh_strn_strn_t *user_rt_access = NULL;
 static struct mh_strnptr_t *user_rt_universe_access = NULL;
 
-void
-mh_strn_strn_node_create(struct mh_strn_strn_node_t *node,
-			 const char *s1, uint32_t s1_len,
-			 const char *s2, uint32_t s2_len)
-{
-	uint32_t hash = mh_strn_strn_hash(s1, s1_len, s2, s2_len);
-	node->s1 = s1;
-	node->s1_len = s1_len;
-	node->s2 = s2;
-	node->s2_len = s2_len;
-	node->hash = hash;
-}
-
 void
 rt_access_init(void)
 {
diff --git a/src/lib/core/assoc.h b/src/lib/core/assoc.h
index 1e91b19794..bc53d360ed 100644
--- a/src/lib/core/assoc.h
+++ b/src/lib/core/assoc.h
@@ -231,6 +231,19 @@ struct mh_strn_strn_node_t {
 #define mh_cmp_key(a, b, arg) mh_cmp(a, b, arg)
 #include "salad/mhash.h"
 
+static inline void
+mh_strn_strn_node_create(struct mh_strn_strn_node_t *node,
+			 const char *s1, uint32_t s1_len,
+			 const char *s2, uint32_t s2_len)
+{
+	uint32_t hash = mh_strn_strn_hash(s1, s1_len, s2, s2_len);
+	node->s1 = s1;
+	node->s1_len = s1_len;
+	node->s2 = s2;
+	node->s2_len = s2_len;
+	node->hash = hash;
+}
+
 #if defined(__cplusplus)
 } /* extern "C" */
 #endif /* defined(__cplusplus) */
```

(We can also move `mh_strn_strn_hash` below to place all the `*_strn_strn_*` functions together.)

src/box/lua/call.c
mandesero: Added.

src/box/rt_access.h
mandesero: Removed.

src/box/rt_access.c
```diff
@@ -0,0 +1,83 @@
...
+		const char *fname, uint32_t fname_len)
+{
+	/* Grant universe access. */
+	if (fname_len == 0) {
+		uint32_t hash = mh_strn_hash(uname, uname_len);
+		const struct mh_strnptr_node_t uid_node =
+			{ uname, uname_len, hash, NULL };
+		struct mh_strnptr_node_t *ignored = NULL;
+		mh_strnptr_put(user_rt_universe_access, &uid_node,
+			       &ignored, NULL);
```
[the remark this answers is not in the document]
mandesero: I add copying username string with `user_name` structure which is allocated with `gap_alloc` + `xmalloc`, also we need to save this structure in node so I add `mh_strn_strn_node/key_t` again.

src/box/rt_lua_call_access.c
mandesero: I removed `mh_strn_strn_node_create` because it is now used in one place and I don't think it is necessary.

src/box/user.cc
```diff
@@ -670,6 +670,16 @@ user_find_by_name(const char *name, uint32_t len)
 	return NULL;
 }
 
+/** Return user name by id */
+const char *
+user_name_by_id(uint32_t uid, uint32_t *name_len)
+{
+	struct user *user = user_find(uid);
+	const char *name = user->def->name;
```
Totktonada, line 678: It seems, we get a NULL dereference here if the user doesn't exist.

mandesero: Check for `NULL` added.

src/box/rt_access.c
```diff
@@ -0,0 +1,83 @@
...
+		return;
+	}
+
+	/* Grant access to lua_call function. */
+	struct mh_strn_strn_node_t access_node;
+	mh_strn_strn_node_create(&access_node, uname, uname_len,
+				 fname, fname_len);
+	struct mh_strn_strn_node_t *ignored = NULL;
+	mh_strn_strn_put(user_rt_access, &access_node, &ignored, NULL);
```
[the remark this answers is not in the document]
Totktonada: The function name still not copied in the new version of the patch.

src/box/rt_access.c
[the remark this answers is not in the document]
Totktonada: Hm. Really, it is how `access_lua_call_registry` works.

It seems unnecessarily complicated, but that's another question. I'll resolve this discussion and will reach you if I have ideas around simplicity of this code.

src/box/user.cc
```diff
@@ -670,6 +670,20 @@ user_find_by_name(const char *name, uint32_t len)
 	return NULL;
 }
 
+/** Return user name by id */
+const char *
+user_name_by_id(uint32_t uid, uint32_t *name_len)
+{
+	struct user *user = user_find(uid);
+	if (user == NULL) {
+		diag_set(ClientError, ER_NO_SUCH_USER, int2str(uid));
```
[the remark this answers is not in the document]
mandesero: `diag_set` removed.

src/box/rt_access.c
```diff
@@ -0,0 +1,111 @@
...
+	node->str = grp_alloc_create_str(&all, uname, uname_len);
+	node->len = uname_len;
+	assert(grp_alloc_size(&all) == 0);
+
+	/* Grant universe access. */
+	if (fname_len == 0) {
+		uint32_t hash = mh_strn_hash(node->str, uname_len);
+		const struct mh_strnptr_node_t uid_node =
+			{ node->str, uname_len, hash, node };
+		mh_strnptr_put(user_rt_universe_access, &uid_node, NULL, NULL);
```
Totktonada, line 81: If the node exists, it is replaced by the new node. It seems, we loss the memory allocated by `grp_alloc` above in the case.

The same for `mh_strn_strn_put()` below.

mandesero: Fixed.

Totktonada: `rt_access_check` may return `true` if the permission is granted for universe, while the `rt_access_grant` is called for a particular function.

Since there is no revoke function it works valid, but I would prefer a solution that would skip putting a value in the hashmap only if the value actually present in *this* hashmap.

I think that it would be less error-prone this way.

src/box/rt_access.c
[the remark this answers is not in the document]
mandesero: Added copying of function name string. I also decided to move the allocation logic up, but chose not to place it in assoc.*, as it seems to me these are local changes only relevant to rt_access.

src/box/box.cc
```diff
@@ -6225,6 +6227,7 @@ box_free(void)
 	tuple_free();
 	port_free();
 	iproto_constants_free();
+	rt_access_destroy();
```
Totktonada, line 6230: From the [style guide](https://www.tarantool.io/en/doc/latest/contributing/c_style_guide/):

> For function naming we have a convention is to use:
>
> * new/delete for functions which allocate + initialize and destroy + deallocate an object,
> * create/destroy for functions which initialize/destroy an object but do not handle memory management,
> * init/free for functions which initialize/destroy libraries and subsystems.

Here we deinitialize the subsystem, so it should be named as `<...>_free()`.

Totktonada, line 6230: The order of the deinitialization is generally the reverse of the initialization order. It is not so in the given case. If there is a reason, why it should be placed here, please comment. Otherwise, please, adjust the order.

src/box/rt_access.c
```diff
@@ -0,0 +1,158 @@
+#include "rt_access.h"
+#include "assoc.h"
+#include "../lua/init.h" /* tarantool_lua_is_builtin_global */
+#include "salad/grp_alloc.h"
+
+static struct mh_strn_strn_t *user_rt_access = NULL;
+static struct mh_strnptr_t *user_rt_universe_access = NULL;
+
+struct mh_strnptr_key_t*
+mh_strnptr_key_alloc(const char *str, uint32_t str_len)
```
Totktonada, line 10: It seems, it should be declared as `static` (and the same for `mh_strn_strn_key_alloc` below).

src/box/rt_access.c
```diff
@@ -0,0 +1,158 @@
...
+{
+	if (h == NULL)
+		return;
+
+	mh_int_t i;
+	mh_foreach(h, i) {
+		struct mh_strnptr_node_t *node = mh_strnptr_node(h, i);
+		if (node != NULL)
+			free(node->val);
+	}
```
Totktonada, line 56: `free()` must accept the NULL pointer according to the [POSIX standard](https://pubs.opengroup.org/onlinepubs/007904975/functions/free.html).

So, we can reduce this code snippet down to:

```c
mh_foreach(h, i)
	free(mh_strnptr_node(h, i)->val);
```

(The same for `mh_strn_strn_free` below.)

src/box/box.cc
mandesero: Renamed.

mandesero: Order changed.

src/box/rt_access.c
mandesero: Done.

src/box/rt_access.c
```diff
@@ -0,0 +1,158 @@
...
+	node->s1 = grp_alloc_create_str(&all, s1, s1_len);
+	node->s1_len = s1_len;
+	node->s2 = grp_alloc_create_str(&all, s2, s2_len);
+	node->s2_len = s2_len;
+	assert(grp_alloc_size(&all) == 0);
+	return node;
+}
+
+void
+mh_strnptr_free(struct mh_strnptr_t *h)
```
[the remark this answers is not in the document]
mandesero: Done.

src/box/rt_access.c
mandesero: Done.

Totktonada (approved):
Thank you for the patch and the updates!

I'm comfortable with the implementation. LGTM.

I've some ideas how to make this code more concise, but I think I shouldn't make this activity a part of the review process.

Totktonada:
@locker Please, look over this patch if you have a spare time. I would appreciate your glance and suggestions.

src/box/lua/call.c
```diff
@@ -1268,6 +1269,46 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
 	return 0;
 }
 
+static int
+lbox_rt_access_reset(struct lua_State *L)
+{
+	if (lua_gettop(L) != 0)
+		return luaL_error(L, "internal.reset_lua_call takes no "
+				  "arguments, but %d given.", lua_gettop(L));
```
locker, line 1277: I don't think we need to bother raising human-readable errors on invalid arguments for these functions because they're internal so it should be fine to use `luaL_checkstring` without checking `lua_top`, but if you do please:

1. Use box errors.
2. Change the error message to "Usage: ...".
3. Check argument types with `lua_type` and raise box errors for them as well.
4. Cover these checks by tests.

I mean something like this:

```C
if (lua_top(L) != 0) {
        diag_set(IllegalParams, "Usage: box.internal.lua_call_runtime_priv_reset()");
        return luaT_error(L);
}

if (lua_top(L) != 2 || lua_type(L, 1) != LUA_TSTRING || lua_type(L, 2) != LUA_TSTRING) {
        diag_set(IllegalParams, "Usage: box.internal.lua_call_runtime_priv_grant(user, func)");
        return luaT_error(L);
}
```

src/lib/core/assoc.h
```diff
@@ -189,6 +189,69 @@ mh_strnu32_find_str(struct mh_strnu32_t *h, const char *str, uint32_t len)
...
 
+/*
+ * Map: (const char *, uint32_t, const char *, uint32_t) => (void *)
+ */
+
+#define mh_name _strn_strn
+/**
+ * Key of `mh_strn_strn_node_t` hash table.
+ */
+struct mh_strn_strn_key_t {
```
locker, line 200: I don't like this new hash table type (primarily because of its name). I'd prefer to introduce a new hash table type exclusively for storing runtime Lua call privileges. However, if you're sure you want to add it, please:

1. Do it in a separate commit.
2. Name it consistently with other generic hash tables. I can't suggest a nice name but `mh_strn_strn` sounds like it's mapping `strn` to `strn`, not `strn+strn` to `ptr`. I guess, to be consistent with our other generic hash tables, it should be named `mh_strnstrnptr`.

src/box/lua/call.c
```diff
@@ -1268,6 +1269,46 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
...
+
+	struct credentials *cr = effective_user();
+	if (cr->uid != ADMIN)
+		return luaL_error(L, "internal.reset_lua_call only ADMIN "
+				  "can grant access.");
```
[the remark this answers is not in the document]
mandesero: I believe this check is necessary for the following reasons. When using `require('net.box').connect`, even if the user has the privileges to execute Lua functions `lua_call: [all]`, this does not grant them privileges to execute built-in functions. In this case, `box.session.su` would be inaccessible.

src/box/rt_access.h
```diff
@@ -0,0 +1,53 @@
...
+ * Initialize the user rt access.
+ */
+void
+rt_access_init(void);
+
+/**
+ * Destroy the user rt access.
+ */
+void
+rt_access_free(void);
```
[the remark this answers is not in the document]
mandesero: I think we still need to call free during deinitialization since we're using `xmalloc`. It's possible to remove `init` and handle initialization when the first privilege is granted, but if we remove `free`, we must ensure that `reset` is called. The question is, where would be the best place to do this?

src/box/lua/call.c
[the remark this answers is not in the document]
locker: > I believe this check is necessary for the following reasons. When using `require('net.box').connect`, even if the user has the privileges to execute Lua functions `lua_call: [all]`, this does not grant them privileges to execute built-in functions. In this case, `box.session.su` would be inaccessible.

These functions should be added to the built-in function list (along with all other box functions) so granting `lua_call: [all]` wouldn't grant the user access to it.

> In addition, @Totktonada and I have decided that this API should only be used by the `admin`, and granting other users the rights to use this API should not be allowed.

We don't do access checks for other internal functions. I don't see why these internal functions should be different.

src/lib/core/assoc.h
mandesero: Renamed to `mh_strnstrnptr` and moved it to a separate commit.

src/box/lua/call.c
mandesero: Replaced with box errors. Added invalid usage test case.

src/box/rt_access.c
```diff
@@ -0,0 +1,154 @@
+#include "rt_access.h"
+#include "assoc.h"
+#include "../lua/init.h" /* tarantool_lua_is_builtin_global */
+#include "salad/grp_alloc.h"
+
+static struct mh_strn_strn_t *user_rt_access = NULL;
+static struct mh_strnptr_t *user_rt_universe_access = NULL;
```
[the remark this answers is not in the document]
mandesero: Redesigned to use a single table.

src/box/rt_access.h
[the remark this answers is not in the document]
mandesero: Moved to `src/box/call.c`.

src/box/call.c
```diff
@@ -173,6 +174,13 @@ access_check_lua_call(const char *name, uint32_t name_len)
 	access &= ~universe.access_lua_call[cr->auth_token].effective;
 	if (access == 0 && !tarantool_lua_is_builtin_global(name, name_len))
 		return 0;
+
+	/* Check for runtime access. */
+	uint32_t uname_len = 0;
+	const char *uname = user_name_by_id(cr->uid, &uname_len);
+	if (uname != NULL && rt_access_check(uname, uname_len, name, name_len))
+		return 0;
+
```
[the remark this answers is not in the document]
mandesero: Removed `user_name_by_id`. Updated check location.

src/box/rt_access.h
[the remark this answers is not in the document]
locker: I think it's okay to call `reset` at exit (where you call `free` now).

src/box/lua/call.c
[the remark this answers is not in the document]
Totktonada: The `lua_call: [all]` point actually doesn't work: it doesn't give an access to these internal functions, because they're in the `box` namespace. Also, I see no other ways to grant access to call these functions accidentally without granting everything.

It seems, there are no strict point for and against the check. They all about consistency.

We can look on the consistency here from two points of view.

1. It is a Lua code and it doesn't usually perform access checks. This is consistency with other functions implemented in Lua.
2. It has the same effect as granting `lua_call` using `box.schema.user.grant()` or a direct write to `_priv` -- the call becomes allowed. These actions are [forbidden](https://github.com/tarantool/tarantool/pull/10358#discussion_r1728984204) for all except the admin using a low level code -- the `_priv` trigger in `alter.cc`. It seems consistent that actions that have the same effect are allowed for the same group of users.

I'm for consistency and I'm not really insist on consistency of the second kind. I think that the effect is more important than how it is implemented, but I can find a logic in the first type of consistency too.

If @locker insist on the first type of consistency, OK.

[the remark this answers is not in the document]
locker: Since everything will work just fine even if we don't add this access check, I think we shouldn't add it.

src/box/rt_access.h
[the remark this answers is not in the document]
mandesero: Done.

src/box/lua/call.c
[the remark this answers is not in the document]
mandesero: Check for `ADMIN` dropped.

src/box/call.c
```diff
@@ -160,6 +163,10 @@ box_run_on_call(enum iproto_type type, const char *expr, int expr_len,
 	trigger_run(&box_on_call, &ctx);
 }
 
+bool
+box_lua_call_runtime_priv_check(const char *uname, uint32_t uname_len,
```
locker, line 167: Nit:
- Please make this function `static`.
- Let's name it `box_lua_call_runtime_priv_is_granted`. A function called `FOO_check` usually returns 0 or -1 in Tarantool.

test/box-luatest/gh_10306_lua_call_runtime_access_test.lua
```diff
@@ -0,0 +1,157 @@
...
+        t.assert_error_msg_equals(msg, function()
+            box.internal.lua_call_runtime_priv_grant("alice")
+        end)
+        t.assert_error_msg_equals(msg, function()
+            box.internal.lua_call_runtime_priv_grant(5, "foo")
+        end)
+        msg = 'Usage: box.internal.lua_call_runtime_priv_reset()'
+        t.assert_error_msg_equals(msg, function()
+            box.internal.lua_call_runtime_priv_reset("alice")
+        end)
```
locker, line 155: Nit: You could write:

```lua
t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_grant)
t.assert_error_msg_equals(msg, box.internal.lua_call_runtime_priv_grant, 'alice')
...
```

locker (approved):

src/box/call.c
mandesero: Renamed.

test/box-luatest/gh_10306_lua_call_runtime_access_test.lua
mandesero: Done.

PR #10362, opened 2024-08-03
doc: update config storage doc

DerekBum:
Update `config.storage` README to include `tarantool-ee` changes from this PR [1].

Note that these changes are for EE only.

[1] https://github.com/tarantool/tarantool-ee/pull/857

Part of https://github.com/tarantool/tarantool-ee/issues/849

### DIFF
--- a/doc/examples/config/config-storage/README.md
+++ b/doc/examples/config/config-storage/README.md
@@ -71,6 +71,15 @@ The data corresponding to the path contains the following fields:
 * `value`: any string.
 * `mod_revision`: unsigned number, last `revision` at which this
 value modification occurred.
+* An optional time until expiration. After this time mark, the data is
+  considered as deleted (since Tarantool 3.2).
+
+Since Tarantool version 3.2 keys could be added to the storage with the
+specified TTL. The keys guaranteed not to expire early, but they could live a
+bit longer than expected due to communications between replicas and master.
+
+Each expired key generates a delete event. This means that each expiration bumps
+the storage revision and wakes up watchers.
 
 ### Operations
 
@@ -89,10 +98,13 @@ means the operation will be performed for all values.
 #### Put
 
 To put the value by the path use
-`config.storage.put(path, value)`, where:
+`config.storage.put(path, value, opts)`, where:
 
 * `path`: string, path.
 * `value`: string.
+* `opts.ttl`: number, TTL in seconds. By default, keys will not expire.
+  This option is only supported for Tarantool versions >= 3.2. For older
+  versions, this option is ignored.
 
 ```lua
 config.storage.put('/foo/bar', 'v1')
@@ -108,6 +120,10 @@ config.storage.put('/foo/bar', 'v2')
 config.storage.put('/foo/bar', 'v2')
 ---
 - revision: 3
+...
+config.storage.put('/foo/bar', 'v3', {ttl = 60})
+---
+- revision: 4
 ...
 ```
 
@@ -225,6 +241,7 @@ Each operation is a list that contains `cmd` string as the
 first element and arguments. For example:
 ```lua
 {'put', '/foo/bar', 'v1'}
+{'put', '/foo/bar', 'v2', {ttl = 60}} -- Since version 3.2.
 {'delete', '/a/'}
 ```
 
@@ -313,6 +330,13 @@ config.storage.txn({
     is_success: false
   revision: 10
 ...
+
+-- Specified TTL for a key. Since version 3.2.
+config.storage.txn({
+    predicates = {{'revision', '==', revision}},
+    on_success = {{'put', '/a', 'v2', {ttl = 60}}}
+    on_failure = {{'put', '/b', 'v1', {ttl = 30}}}
+})
 ```
 
 ### Watchers


doc/examples/config/config-storage/README.md
```diff
@@ -111,6 +111,18 @@ config.storage.put('/foo/bar', 'v2')
 ...
 ```
 
+Only in Enterprise Edition a TTL option is available. To set a TTL in seconds
```
oleg-jukovec, line 114: The config storage available only in Tarantool EE. I see no reason for this sentence.
```suggestion
To set a TTL in seconds
```

oleg-jukovec (commented):
I think we need to update just the lines:

https://github.com/tarantool/tarantool/blob/09bfdd473ffd23169dbe8d4e5d0510e8c70b11ec/doc/examples/config/config-storage/README.md?plain=1#L91-L95

And maybe add some example with `txn`.

doc/examples/config/config-storage/README.md
DerekBum: Updated.

oleg-jukovec (approved):
Thank you.

doc/examples/config/config-storage/README.md
```diff
@@ -89,10 +89,11 @@ means the operation will be performed for all values.
 #### Put
 
 To put the value by the path use
-`config.storage.put(path, value)`, where:
+`config.storage.put(path, value, opts)`, where:
 
 * `path`: string, path.
 * `value`: string.
+* `opts.ttl`: number, TTL in seconds. By default, keys will not expire.
```
Totktonada, line 96: I would add information about tarantool version since which it is supported and that the older versions ignore this option.

doc/examples/config/config-storage/README.md
```diff
@@ -313,6 +319,11 @@ config.storage.txn({
     is_success: false
   revision: 10
 ...
+config.storage.txn({
+    predicates = {{'revision', '==', revision}},
+    on_success = {{'put', '/a', 'v2', {ttl = 60}}}
+    on_failure = {{'put', '/b', 'v1', {ttl = 30}}}
+})
```
Totktonada, line 326: Two previous examples are accompanied with a comment about an idea of the exapmle. Let's do it here too.

doc/examples/config/config-storage/README.md
Totktonada, line 96: I would add an information about guarantees behind the TTL option. The Data model section looks suitable for it.

Totktonada (approved):
Thanks for the document update! It is OK for me.

doc/examples/config/config-storage/README.md
DerekBum: Thanks, updated.

doc/examples/config/config-storage/README.md
DerekBum: Updated.

doc/examples/config/config-storage/README.md
DerekBum: Updated.

doc/examples/config/config-storage/README.md
```diff
@@ -71,6 +71,12 @@ The data corresponding to the path contains the following fields:
 * `value`: any string.
 * `mod_revision`: unsigned number, last `revision` at which this
 value modification occurred.
+* `deadline`: unsigned number, time until expiration. This field is present
+  since Tarantool version 3.2.
```
Totktonada, line 75: Sorry for nitpicking, but all the other fields (`path`, `value` and `mod_revision`) are present in the `config.storage.get` response. It may give a false assumption that the deadline is present in this response are well.

Since the deadline doesn't shown in any response, I wouldn't mention it here, at least not in this form.

We can say that a TTL may be associated with the data, but we shouldn't give any certain name. For example:

> * `path`: <...>
> * `value`: <...>
> * `mod_revision`: <...>
> * An optional time to live, after which the data is considered as deleted (since Tarantool 3.2).

DerekBum: Updated.

Totktonada: Thanks!

doc/examples/config/config-storage/README.md
```diff
@@ -71,6 +71,12 @@ The data corresponding to the path contains the following fields:
 * `value`: any string.
 * `mod_revision`: unsigned number, last `revision` at which this
 value modification occurred.
+* An optional time until expiration. After this time mark, the data is
+  considered as deleted (since Tarantool 3.2).
+
+Since Tarantool version 3.2 keys could be added to the storage with the
+specified TTL. The keys guaranteed not to expire early, but they could live a
+bit longer than expected due to communications between replicas and master.
```
Totktonada, line 79: (Sorry that I can't send all the comments at once: some thoughts arrive into the head after some delay.)

I would highlight that the expiration bumps the storage revision and wake ups watchers.

It is similar to what etcd docs [states](https://etcd.io/docs/v3.4/learning/api/#lease-api) regarding the lease mechanism:

> Each expired key generates a delete event in the event history.

DerekBum: Updated.

PR #10364, opened 2024-08-05
upgrade: fix upgrading from schema 1.6.9

### LINKED ISSUE #10180: Fix upgrading from version 1.6 to 1.10 and above
**Bug description**
After upgrading tarantool 1.6 to 1.10, issuing `box.schema.upgrade()`, `box.snapshot()` and restarting the replica, it started failing like this:
```
2024-06-27 16:31:10.963 [34] main/101/autogen.lua I> recovering from `/var/lib/tarantool/snaps/00000000000000000317.snap'
2024-06-27 16:31:10.963 [34] main/101/autogen.lua I> cluster uuid 10da52ae-5102-4ee9-9996-8196f15a1126
2024-06-27 16:31:10.965 [34] main/101/autogen.lua memtx_space.c:317 E> rollback: Tuple field 5 type does not match one required by operation: expected map
2024-06-27 16:31:10.965 [34] main/101/autogen.lua key_def.h:672 E> ER_FIELD_TYPE: Tuple field 5 type does not match one required by operation: expected map
2024-06-27 16:31:10.965 [34] main/101/autogen.lua F> can't initialize storage: Tuple field 5 type does not match one required by operation: expected map
```
It turned out the problem was with the following request (lsn 87) in the snapshot:
Снимок экрана 2024-06-27 в 18 19 57

Note how it has field 5 encoded as a msgpack array rather than a map.
This was allowed back in 1.6 days (field 5 of space `_user` had type 'any'):
https://github.com/tarantool/tarantool/blob/e14d0b4eabf506f94f94853a11377cb893e3d035/src/box/lua/upgrade.lua#L338-L340

But during upgrade to 1.7.5 this was changed, and field 5 started to expect a map.
https://github.com/tarantool/tarantool/blob/e14d0b4eabf506f94f94853a11377cb893e3d035/src/box/lua/upgrade.lua#L615-L616

At the same time user creation was fixed: in 1.6 empty user credentials were encoded as an array (erroneously):
https://github.com/tarantool/tarantool/blob/d36ba2796839db0edd3c62e9048cd41b1ee08d1e/src/box/lua/schema.lua#L1251-L1256
And in 1.7.something it became:
https://github.com/tarantool/tarantool/blob/24b70de108673da4ebc3e8124d64b874ee6d6d8a/src/box/lua/schema.lua#L1332-L1337

Unfortunately this bug in empty table encoding was not taken into account in the upgrade script: it left "empty array" fields as is without converting them to `map`:
https://github.com/tarantool/tarantool/blob/e14d0b4eabf506f94f94853a11377cb893e3d035/src/box/lua/upgrade.lua#L563-L574

Let's fix this to ease our users migration from 1.6 to 1.10 and above.
I think this should be fixed in 1.10 as well, because a direct upgrade from 1.6 to 2.x is only possible with a downtime. The only way to perform a rolling upgrade is taking an intermediate step at 1.10

Note that the fix should be applied to the triggers which allow to recover 1.6 schema on 2.x as well. The triggers were copied from the old upgrade scripts and hence contain the same error:
https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/lua/upgrade.lua#L191-L199
https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/lua/upgrade.lua#L255-L261

A more correct predicate for updating `tuple[5]` contents would be
`tuple[5] == nil or type(tuple[5] ~= table) or table_is_empty(tuple[5])`

Serpentian:
This commit fixes some cases of upgrading schema from 1.6.9:

1. Fix updating empty password for users. In 1.6 credentials were array
   in _user, in 1.7.5 they became map.

2. Automatically update the format of user spaces. Format of system
   spaces have been properly fixed during upgrade to 1.7.5. However,
   commit https://github.com/tarantool/tarantool/commit/519bc82e4fb4250b7d337a03680cdef355555037 ("Parse and validate space formats") introduced
   strict checking of format field in 1.7.6. So, the format of user
   spaces should be also fixed.

Back in 1.6 days, it was allowed to write anything in space format.
This commit only fixes valid uses of format:
    {name = 'a', type = 'number'}
    {'a', type = 'number'}
    {'a', 'num'}
    {'a'}

Invalid use of format (e.g. {{}}, or {{5, 'number'}} will cause error
anyway. User has to fix the format on old version and only after that
start a new one.

This commit also introduces the test, which checks, that we can
properly upgrade from 1.6.9 to the latest versions, at least in basic
cases.

Closes https://github.com/tarantool/tarantool/issues/10180


locker (approved):

changelogs/unreleased/gh-10180-upgrade-from-1-6.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/upgrade
+
+* Automatically fix incorrect format of user defined spaces and empty password
+  for users during upgrade to 1.7.5 (gh-10180).
```
lenkis, line 4:
```suggestion
* Automatically fix incorrect format of user-defined spaces and empty password
  for users during upgrade to 1.7.5 (gh-10180).
```

test/box-luatest/gh_10180_upgrade_from_1_6_8_test.lua
```diff
@@ -0,0 +1,28 @@
...
+g.test_upgrade = function(cg)
+    cg.server:exec(function()
+        t.assert_equals(box.space._schema:get{'version'}, {'version', 1, 6, 8})
+        t.assert(pcall(box.schema.upgrade))
+        local version = require('internal.mkversion').get()
+        t.assert_equals(box.space._schema:get{'version'},
+            {'version', version.major, version.minor, version.patch})
+    end)
+    -- Restart the server and check, that we properly recover even if snapshot
+    -- have not been taken. Luatest automatically calls wait_until_ready().
```
lenkis, line 26:
```suggestion
    -- Restart the server and check that we recover properly even if the snapshot
    -- has not been taken. Luatest automatically calls wait_until_ready().
```

lenkis (approved):

src/box/lua/upgrade.lua
```diff
@@ -314,9 +315,25 @@ space_formats_1_7_5._vuser = space_formats_1_7_5._user
...
+    end
+    if space_formats_1_7_5[tuple[3]] then
+        -- Update format of system spaces.
+        if not table.equals(space_formats_1_7_5[tuple[3]], tuple[7]) then
+            tuple = tuple:update{{'=', 7, space_formats_1_7_5[tuple[3]]}}
+            log.info("Update space '%s' format: new format %s", tuple[3],
+                json.encode(tuple[7]))
+        end
+    elseif tuple[7] and type(tuple[7]) == 'table' and next(tuple[7]) ~= nil and
+           tuple[7][1]['name'] == nil then
```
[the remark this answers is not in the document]
Serpentian: Sorry, I didn't get you. If user specified the format as `{}`, then we have nothing to do. This code block updates format from `{<some-name>, type = <type>}` to `{name = <some-name>, type = <type>}`. We should not update empty formats and this is checked with `next(tuple[7]) ~= nil`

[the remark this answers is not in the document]
sergepetrenko: > We should not update empty formats and this is checked with next(tuple[7]) ~= nil

Why? Aren't we fixing exactly this problem for user passwords? When they were erroneously encoded as an empty `MP_ARRAY` and not an empty `MP_MAP`.

Can the same happen with space formats?

IOW, I suggest overwriting empty tables, no matter how they are encoded (`MP_ARRAY` or `MP_MAP`)

src/box/lua/upgrade.lua
```diff
@@ -314,9 +315,34 @@ space_formats_1_7_5._vuser = space_formats_1_7_5._user
...
+    if tuple == nil then
+        return nil
+    end
+    local is_changed = false
+    if space_formats_1_7_5[tuple[3]] and
        not table.equals(space_formats_1_7_5[tuple[3]], tuple[7]) then
+        -- Update format of system spaces.
         tuple = tuple:update{{'=', 7, space_formats_1_7_5[tuple[3]]}}
+        is_changed = true
+    elseif tuple[7] and type(tuple[7]) == 'table' and next(tuple[7]) ~= nil and
```
Gerold103, line 327: Ditto, `type(tuple[7]) == 'table'` already filters out `nils`, so the previous check of `elseif tuple[7]` isn't needed.

test/box-luatest/upgrade/1.6.8/gen.lua
```diff
@@ -0,0 +1,61 @@
+--
+-- This script creates a snapshot on Tarantool 1.6.9-94-ga5da60bdd.
```
Gerold103, line 2: But the folder is called `upgrade/1.6.8`. Shouldn't it use 1.6.8 then?

Gerold103 (commented):
Thanks! Clear and understandable. I only have a question to the general idea - does it make sense to guarantee such a long upgrade-jump from 1.6.8 to 3? AFAIU, each major version is potentially breaking. So users were always expected to go through at least version 2, before they jump to 3, if they are having some breaking points, or?

src/box/lua/upgrade.lua
```diff
@@ -253,7 +253,8 @@ local function upgrade_to_1_7_5()
 end
 
 local function user_trig_1_7_5(_, tuple)
-    if tuple and not tuple[5] then
+    if tuple and
+       (tuple[5] == nil or type(tuple[5]) ~= 'table' or not next(tuple[5])) then
```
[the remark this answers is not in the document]
Serpentian: Agree, fixed

src/box/lua/upgrade.lua
Serpentian: Fixed

test/box-luatest/upgrade/1.6.8/gen.lua
Serpentian: Renamed the dir name to 1.6.9)

Serpentian:
> I only have a question to the general idea - does it make sense to guarantee such a long upgrade-jump from 1.6.8 to 3? AFAIU, each major version is potentially breaking. So users were always expected to go through at least version 2, before they jump to 3, if they are having some breaking points, or?

Since we will release new 1.10 version anyway, I suppose, we should do our best to ease the upgrade process for our users, it costs us nothing. 

We guarantee, that user can make upgrade without downtime 1.6 -> 1.10 -> 2.X -> 3.X. With downtime it's said in docs, that we support direct upgrade from 1.6 -> 2.X ([docs](https://www.tarantool.io/en/doc/latest/admin/upgrades/1.6-1.10/)). There're no incompatible changes in direct upgrade 1.6 -> 3.X, at least for now (the test shows that), and I hope it'll remain the same.

src/box/lua/upgrade.lua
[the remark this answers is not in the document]
Serpentian: Fixed

src/box/lua/upgrade.lua
```diff
@@ -314,9 +314,34 @@ space_formats_1_7_5._vuser = space_formats_1_7_5._user
...
+        -- {name = 'a', type = 'number'}.
+        local format = tuple[7]
+        for _, f in ipairs(format) do
+            f['name'] = f[1]
+            f[1] = nil
+            is_changed = true
+        end
+        tuple = tuple:update{{'=', 7, format}}
+    elseif tuple[7] == nil
+           or type(tuple[7]) ~= 'table' or not next(tuple[7]) then
```
Gerold103, line 339: `tuple[7] == nil` is included into `type(tuple[7]) ~= 'table'`.

test/box-luatest/gh_10180_upgrade_from_1_6_9_test.lua
```diff
@@ -0,0 +1,29 @@
...
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.test_upgrade = function(cg)
+    cg.server:exec(function()
+        t.assert_equals(box.space._schema:get{'version'}, {'version', 1, 6, 8})
```
Gerold103, line 19: The test is called 'upgrade from 1.6.9', which means the schema must be at least 1.6.9, not 1.6.8, no? Or did 1.6.9 not have any own changes since 1.6.8 and hence didn't have its own schema version?

Gerold103 (commented):
Only nits are left 🙏!

test/box-luatest/gh_10180_upgrade_from_1_6_9_test.lua
Serpentian: > Or did 1.6.9 not have any own changes since 1.6.8 and hence didn't have its own schema version?

Yep:

```lua
./src/tarantool: version 1.6.9-94-ga5da60bdd
tarantool> box.cfg{}
2024-10-15 17:10:21.604 [143921] main/101/interactive C> version 1.6.9-94-ga5da60bdd
2024-10-15 17:10:21.604 [143921] main/101/interactive C> log level 5
2024-10-15 17:10:21.604 [143921] main/101/interactive I> mapping 1073741824 bytes for tuple arena...
2024-10-15 17:10:21.606 [143921] main/101/interactive I> initializing an empty data directory
2024-10-15 17:10:21.607 [143921] snapshot/101/main I> creating `./00000000000000000000.snap.inprogress'
2024-10-15 17:10:21.607 [143921] snapshot/101/main I> saving snapshot `./00000000000000000000.snap.inprogress'
2024-10-15 17:10:21.607 [143921] snapshot/101/main I> done
2024-10-15 17:10:21.612 [143921] main/101/interactive I> ready to accept requests
---
...

tarantool> box.space._schema:get('version')
---
- ['version', 1, 6, 8]
...

```

src/box/lua/upgrade.lua
Serpentian: Fixed

src/box/lua/upgrade.lua
```diff
@@ -314,9 +314,33 @@ space_formats_1_7_5._vuser = space_formats_1_7_5._user
...
+        -- {'a', type = 'number'}, and not it looks like that:
+        -- {name = 'a', type = 'number'}.
+        local format = tuple[7]
+        for _, f in ipairs(format) do
+            f['name'] = f[1]
+            f[1] = nil
+            is_changed = true
+        end
+        tuple = tuple:update{{'=', 7, format}}
+    elseif type(tuple[7]) ~= 'table' or not next(tuple[7]) then
```
Gerold103, line 338: I am wondering why here it is `not next(tuple[...])` and in the seemingly same places above it is `next(tuple[...]) == / != nil`? If this can be `box.NULL`, then shouldn't we always explicitly compare with `nil`?

Gerold103 (approved):
Thanks! Top 💪.

src/box/lua/upgrade.lua
Serpentian: I suppose, it was just added on different iterations of review and this is the result.  Let's explicitly use comparison with `nil` everywhere, since we don't wanna see `{box.NULL}` arrays in the format anyway.

changelogs/unreleased/gh-10180-upgrade-from-1-6.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/upgrade
```
sergepetrenko, line 1: > 2. Start to automatically change format of user spaces: previously it
   was {'a', type = 'number'}, now it's {name = 'a', type = 'number'}.
   Make sure, that space format is encoded as array.

Are you sure it wasn't possible to set format like `{{'a', 'number'}}`? I see this syntax is supported on new tarantool versions, but the format is converted to `{{name='a', type='number'}}` internally.

Also I don't understand this sentence: `Make sure, that space format is encoded as array.`

sergepetrenko (commented):
Thanks for the patch!

changelogs/unreleased/gh-10180-upgrade-from-1-6.md
Serpentian: > Are you sure it wasn't possible to set format like {{'a', 'number'}}? I see this syntax is supported on new tarantool versions, but the format is converted to {{name='a', type='number'}} internally.

Back in 1.6 it was also possible to use both variants. The problem there, is that it have not been converted:

```lua
src/tarantool: version 1.6.9-94-ga5da60bdd
<...>
tarantool> box.schema.space.create('test', {format = {{name = 'a', type = 'number'}}})
<...>
arantool> box.schema.space.create('test2', {format = {{'a', type = 'number'}}})
<...>
tarantool> box.space._space:select()
<...>
  - [512, 1, 'test', 'memtx', 0, {}, [{'name': 'a', 'type': 'number'}]]
  - [513, 1, 'test2', 'memtx', 0, {}, [{1: 'a', 'type': 'number'}]]
```

Without changing the format of user spaces, we fail here:

https://github.com/tarantool/tarantool/blob/6ae595f5da606478223d69b641213444766bca85/src/box/field_def.c#L512-L516

```text
"object_id","type":"unsigned"},{"name":"privilege","type":"unsigned"}]
[001] server | 2024-10-31 13:18:51.814 [19517] main/104/server_instance.lua/box.upgrade I> Update space '_cluster' format: new format [{"name":"id","type":"unsigned"},{"name":"uuid","type":"string"}]
[001] server | 2024-10-31 13:18:51.814 [19517] main/104/server_instance.lua memtx_engine.cc:450 E> error at request: {type: 'INSERT', replica_id: 0, lsn: 33, space_id: 280, index_id: 0, tuple: [514, 1, "format", "memtx", 0, {}, [{1: "a", "type": "number"}, {1: "b", "type": "num"}, {1: "c", "type": "array"}, {1: "d", "type": "array"}]]}
[001] server | 2024-10-31 13:18:51.814 [19517] main/104/server_instance.lua field_def.c:514 E> ER_WRONG_SPACE_FORMAT: Wrong space format field 1: expected a map with string keys
[001] server | 2024-10-31 13:18:51.814 [19517] main/104/server_instance.lua F> can't initialize storage: Wrong space format field 1: expected a map with string keys
[001] server | 2024-10-31 13:18:51.814 [19517] main/104/server_instance.lua F> can't initialize storage: Wrong space format field 1: expected a map with string keys
[001] not ok 1	box-luatest.gh_10180_upgrade_from_1_6_9.test_upgrade
[001] #   ...ng/tnt/tarantool/test-run/lib/luatest/luatest/server.lua:539: Process is terminated when waiting for "server is ready" condition for server (alias: server, workdir: server-8qNUtZhveKJA, pid: 19517)
[001] #   stack traceback:
[001] #   	...ol/test/box-luatest/gh_10180_upgrade_from_1_6_9_test.lua:10: in function <...ol/test/box-luatest/gh_10180_upgrade_from_1_6_9_test.lua:6>
[001] #   	...
[001] # Ran 1 tests in 0.334 seconds, 0 succeeded, 1 errored
[001] 
```

sergepetrenko: Thanks for addressing this issue!

sergepetrenko (approved):

Gerold103 (approved):

sergepetrenko:
@Serpentian, please prepare a backport to 1.10 as well

PR #10376, opened 2024-08-07
doc: update config storage doc

DerekBum:
Update `config.storage` README to include new changes to `config.storage` API:
Any request with specified TTL used on older version of schema will end with an error.

Part of tarantool/tarantool-ee#849

NO_DOC=doc_changes
NO_TEST=doc_changes
NO_CHANGELOG=doc_changes


doc/examples/config/config-storage/README.md
```diff
@@ -78,6 +78,18 @@ Since Tarantool version 3.2 keys could be added to the storage with the
 specified TTL. The keys guaranteed not to expire early, but they could live a
 bit longer than expected due to communications between replicas and master.
 
+If the schema used by the user is not upgraded, any request with specified TTL
+will end with an error. To check, if the schema needs an upgrade, use
+`box.internal.schema_needs_upgrade()`. To check if TTL requests are supported,
```
Totktonada, line 83: Please, don't suggest to use an internal API.

There is a warning in logs. I would like to add an alert into `box.info.config.alerts`, but there is no one at the moment.

Can we omit the 'to check <...>, use <...>' sentence, at least for now?

doc/examples/config/config-storage/README.md
```diff
@@ -78,6 +78,18 @@ Since Tarantool version 3.2 keys could be added to the storage with the
...
+use newly added field `config.storage.info.features.ttl`:
+```lua
+local info = conn.call('config.storage.info')
+if not info.features.ttl then
```
Totktonada, line 87: It seems, we should suggest `if not (info.features or {}).ttl then` or `if not info.features or not info.features.ttl then` -- otherwise this code may fail on an attempt to index `nil` value.

doc/examples/config/config-storage/README.md
DerekBum: Updated.

doc/examples/config/config-storage/README.md
DerekBum: Updated.

doc/examples/config/config-storage/README.md
Totktonada: Quote:

> If the schema used by the user is not upgraded, any request with specified TTL
will end with an error. To check if TTL requests are supported, use newly added
field `config.storage.info.features.ttl`:
> ```lua
> local info = conn.call('config.storage.info')
> if info.features == nil or not info.features.ttl then
>     -- Upgrade the schema.
> end
> ```
> To upgrade the schema, use `box.schema.upgrade()` and `box.snapshot()`.

It mixes an application/module developer scenario, when a client service needs to know capabilities of the server side service, and an administration scenario, when an administrator examines what is going wrong with the server side service.

A developer may actually use a `net.box` connection and the feature check to prepare a client code to work with different server versions. For example, lack of a necessary feature may be reported as an error.

The service health is the work of an administrator. He/she doesn't use `net.box` (usually), because he/she can enter into the admin console and invoke needed administrative functions directly. We shouldn't suggest writting some code using `net.box` to the administrator, because it unnecessarily complicates a simple thing. Moreover, this instruction may be confusing in the sense that it is not clear, where the upgrade should be run: on the client or on the server.

IOW, I propose to split these cases.

> To check if TTL requests are supported, use newly added field `config.storage.info.features.ttl`:
>
> ```lua
> local info = conn.call('config.storage.info')
> if info.features == nil or not info.features.ttl then
>     error('TTL is not supported on the config storage server')
> end
> ```
>
> Tarantool 3.2 needs the database schema upgrade to enable the TTL feature. Use `box.schema.upgrade()` and `box.snapshot()`. See also the [upgrade guides](https://www.tarantool.io/en/doc/latest/admin/upgrades/).

If you really want to give a recipe how to check whether the instance is upgraded, OK. I think that it is redundant, but there are some:

* Look over the `Your schema version is <...> while Tarantool <...> requires a more recent schema version. Please, consider using box.schema.upgrade().` message in logs.
* Run `box.atomic(function() box.space._schema:replace({'x'}) box.space._schema:delete({'x'}) end)` from the console. If the instance is not upgraded, an error is shown.

DerekBum: Updated. I've considered to omit instructions on how to check if the instance is upgraded.

Totktonada (approved):

Totktonada:
It is a README file update, so it doesn't require `full-ci`. I'm going to proceed with the patch.

PR #10379, opened 2024-08-07
vinyl: do not abort unrelated transactions on DDL

### LINKED ISSUE #10375: DDL aborts unrelated Vinyl transactions
**Bug description**

Any DDL operation aborts **all** transactions, even those that wouldn't be otherwise affected by it.

```
Tarantool 3.2.0-entrypoint-200-gd9b643b0b198
Target: Linux-x86_64-Debug
Compiler: GNU-13.2.0
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/vlad/src/tarantool/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror -g -ggdb -O0
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/vlad/src/tarantool/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror -g -ggdb -O0
```

**Steps to reproduce**

Run the following script:

```Lua

box.cfg{log_level = 'warn'}

s:create_index('pk')

box.begin()
s:insert({1})

local f = fiber.new(box.schema.space.create, 'test2')
f:set_joinable(true)
f:join()

box.commit()

os.exit(0)
```

**Actual behavior**

The Vinyl transaction fails to commit:

```
2024-08-07 10:58:50.707 [129565] main/114/lua memtx_tx.c:698 W> Transaction committing DDL (id=392) has aborted another TX (id=391)
2024-08-07 10:58:50.707 [129565] main test.lua:15 E> ER_TRANSACTION_CONFLICT: Transaction has been aborted by conflict
2024-08-07 10:58:50.707 [129565] main F> fatal error, exiting the event loop
```

**Expected behavior**

The Vinyl transaction commits successfully because it's unaffected by the DDL operation.

A DDL operation should abort only those transactions that actually read from or wrote to the space modified by the DDL operation, not all active transactions. The bug was introduced by commit 8f4be3227635ef606581dda7e85f254a96347c0e. I fixed a similar issue in Vinyl a while back, see commit d3e12369565151ba8c0632a84adab4bf90eff23f. We need to do something like it in memtx.

locker:
Since commit 8f4be3227635 ("txm: disallow yields after DDL operation in TX"), any DDL operation aborts **all** active transactions, even those that wouldn't be affected by it anyway, see `memtx_engine_prepare()`, `memtx_tx_abort_all_for_ddl()`. Actually, there's no need to do that in Vinyl because it properly handles concurrent DDL operations, see commit d3e123695651 ("vinyl: abort affected transactions when space is removed from cache"). Let's skip Vinyl transactions from consideration by marking the Vinyl engine with a special flag.

Closes #10375

### DIFF
--- a/changelogs/unreleased/gh-10375-vy-do-not-abort-unrelated-tx-on-ddl.md
+++ b/changelogs/unreleased/gh-10375-vy-do-not-abort-unrelated-tx-on-ddl.md
@@ -1,0 +1,3 @@
+## bugfix/vinyl
+
+* Fixed a bug when any DDL operation aborted unrelated transactions (gh-10375).
--- a/src/box/engine.h
+++ b/src/box/engine.h
@@ -299,6 +299,12 @@ enum {
 	 * Set if the engine supports cross-engine transactions.
 	 */
 	ENGINE_SUPPORTS_CROSS_ENGINE_TX = 1 << 4,
+	/**
+	 * Set if the engine's transaction manager properly handles
+	 * concurrent DDL operations. A DDL operation will abort all
+	 * transactions for engines that don't have this flag set.
+	 */
+	ENGINE_TXM_HANDLES_DDL = 1 << 5,
 };
 
 struct engine {
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -690,6 +690,8 @@ memtx_tx_abort_all_for_ddl(struct txn *ddl_owner)
 	rlist_foreach_entry(to_be_aborted, &txm.all_txs, in_all_txs) {
 		if (to_be_aborted == ddl_owner)
 			continue;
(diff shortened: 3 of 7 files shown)


Gumix (approved):

xuniq (approved):

locker:
Cherry-picked to 3.1 and 2.11.

PR #10383, opened 2024-08-07
Add a test for the vclock_to_string and vclock_from_string and update a whitelist in test_engine.lua

ligurio:


sergepetrenko (approved):
Thanks for the patch!

test/fuzz/vclock_from_string_fuzzer.c
```diff
@@ -0,0 +1,27 @@
...
+	if (!buf)
+		return 0;
+	memcpy(buf, data, size);
+	buf[size] = '\0';
+
+	struct vclock vclock;
+	vclock_create(&vclock);
+	size_t res = vclock_from_string(&vclock, buf);
+	const char *vclock_str = vclock_to_string(&vclock);
+	if (size == res) {
```
sergepetrenko, line 20: It seems `vclock_from_string` returns 0 when everything's good and amount of successfully parsed bytes otherwise:
https://github.com/tarantool/tarantool/blob/f7bb3fc7a1222107753dc420675ecda043b0a5d2/src/lib/vclock/vclock.c#L150-L161

sergepetrenko: Also by looking at the code of `vclock_from_string` it seems it accepts vclocks like `{    1  : 2}` (with many spaces in between), but this input string won't match what is output from `vclock_to_string`.

sergepetrenko: I think making sure input and output strings are the same would be quite a hustle.

You could check that `vclock_from_string(intput)` is the same as `vclock_from_string(vclock_to_string(input))` instead

ligurio: Updated version doesn't use returned value.

Buristan (approved):
Hi, Sergey!
Thanks for the patch set!
I'll proceed with a review per patch below.

---

[PATCH 1/3] test/fuzz: add vclock_from_string_fuzzer

LGTM!

---

[PATCH 2/3] test/fuzz: ignore new error messages

LGTM, except a minor nit below:

> New error was introduced by commit 09bfdd473ffd

Typo: s/New error/The new error/
or
Typo: s/New error/A new error/

---

[PATCH 3/3] test/fuzz: use custom Lua headers

LGTM!

PR #10324, opened 2024-07-29
config: grant access to specific function in config

mandesero:
This patch adds the capability to grant access to specific Lua functions in the declarative configuration. You can now grant access by assigning the function name to `credentials.users.<user>.privileges.lua_call`.

For example, setting
	`credentials.users.alice.privileges.lua_call = ['foo']`
allows user 'alice' to execute the `foo` function.

Closes #10304

PR #10392, opened 2024-08-09
config: grant runtime access to call_lua functions from declarative configuration

### LINKED ISSUE #10310: config: grant access to call a global Lua function in runtime

mandesero:
This patch introduces the ability to grant execution privileges for Lua functions through declarative configuration, even when the database is in read-only mode or has an outdated schema version.

Users can specify  `lua_call: [<func_name>]`, enabling the execution of specified Lua functions (e.g., `failover:execute()` when all instance are in read-only mode). The `lua_call: [all]` option is also supported, allowing access to all global Lua functions except built-in ones, regardless of database mode or status.

Privileges are still written to the database when possible for consistency and compatibility.

Closes #10310


mandesero:
Waiting for
- [ ] #10324
- [ ] #10358

src/box/lua/config/applier/runtime_priv.lua
```diff
@@ -0,0 +1,64 @@
...
+                    end
+                end
+            end
+            for _, role_name in ipairs(user_roles) do
+                for call, _ in pairs(role_lua_call_priv[role_name] or {}) do
+                    privs[user_or_role_name][call] = true
+                end
+            end
+        end
+    end
```
Totktonada, line 39: This way we can miss privileges granted by a role that is granted for other role.

src/box/lua/config/applier/runtime_priv.lua
```diff
@@ -0,0 +1,64 @@
...
+
+    local role_lua_call_priv = {}
+    local user_lua_call_priv = {}
+
+    local function process_lua_call(privs, user_or_role_name, user_or_role_data)
+        privs[user_or_role_name] = {}
+        local user_roles = user_or_role_data.roles or {}
+
+        if user_or_role_data.privileges then
+            for _, privilege in pairs(user_or_role_data.privileges) do
```
Totktonada, line 15: Nit: AFAIR, it is an array-like table, we can use `ipairs` here.

src/box/lua/call.c
```diff
@@ -1269,6 +1269,15 @@ lbox_func_new_or_delete(struct trigger *trigger, void *event)
...
 }
 
+static int
+lbox_rt_lua_call_access_reset(struct lua_State *L)
+{
+	if (lua_gettop(L) != 0)
+		return luaL_error(L, "internal.clear_lua_call bad arguments.");
+	rt_lua_call_access_reset();
+	return 0;
+}
```
Totktonada, line 1279: It is better to move it into the patch, where the corresponding grant function is added.

src/box/lua/config/applier/runtime_priv.lua
```diff
@@ -0,0 +1,64 @@
...
+    credentials.users = credentials.users or {}
+
+    local role_lua_call_priv = {}
+    local user_lua_call_priv = {}
+
+    local function process_lua_call(privs, user_or_role_name, user_or_role_data)
+        privs[user_or_role_name] = {}
+        local user_roles = user_or_role_data.roles or {}
+
+        if user_or_role_data.privileges then
```
Totktonada, line 14: Still may have roles with `lua_call` inside.

src/box/lua/config/applier/runtime_priv.lua
```diff
@@ -0,0 +1,64 @@
+local function apply_rt_access(config_module)
+    local configdata = config_module._configdata
+    local credentials = configdata:get('credentials') or {}
+    credentials.roles = credentials.roles or {}
+    credentials.users = credentials.users or {}
```
Totktonada, line 5: It modifies the configuration table, so other callers of `config:get()` may receive the modified result. I think it is better to avoid it.

src/box/lua/config/applier/runtime_priv.lua
mandesero: Fixed. Also added role transitivity to the tests.

src/box/lua/config/applier/runtime_priv.lua
mandesero: Done.

src/box/lua/call.c
mandesero: Done.

src/box/lua/config/applier/runtime_priv.lua
mandesero: Done.

src/box/lua/config/applier/runtime_priv.lua
mandesero: `lua_call` from `privileges` and `lua_call` we get from `roles` are separated now

changelogs/unreleased/gh-10310-runtime-grant-lua-call-via-config.md
```diff
@@ -0,0 +1,5 @@
+## feature/box
```
Totktonada, line 1: Nit: This is config's feature (from a user perspective). Let's use the `## feature/config` category.

src/box/lua/config/applier/runtime_priv.lua
Totktonada: On a brief glance I found no test case for a user without `privileges`, but with a role that grants `lua_call`. It worth to add, I think.

changelogs/unreleased/gh-10310-runtime-grant-lua-call-via-config.md
mandesero: Replaced with `feature/config`

xuniq (approved):

src/box/lua/config/applier/runtime_priv.lua
Totktonada, line 1: This algorithm seems generally OK for me. However, I'm in a little doubt about the following points:

* It can go into an infinite recursion if a role depends on itself. It would be nice to detect such a situation.
* We reset the privileges and them start transforming the configuration. I would perform the transformation first and then, when we know that everything is good, reset+grant all the needed privileges.

I've experimented with this code a bit. My result: [code](https://github.com/Totktonada/tarantool/blob/experiment-with-runtime-priv-lua-1/src/box/lua/config/applier/runtime_priv.lua), [diff](https://github.com/Totktonada/tarantool/commit/02b260a9009d602983f2f37689e87c6a4dd1dd31).

The experiment also attempts to improve the readability (including brief comments), but, of course, this is a thing of a taste and feelings. Feel free to use and feel free to decline.

Totktonada (approved):
Looks generally OK for me. I've shared some thoughts above.

I'll return here back a bit later to look across the test cases.

test/box-luatest/gh_10310_grant_rt_access_from_config_test.lua
Totktonada, line 1: I think that the `test/config-luatest` test suite fits better here. (Also, I would use the `config:` category prefix in the commit subject instead of `box:`.)

Also, since this a new feature, we can omit the `gh_xxxxx` prefix in the test. Something like `test/config-luatest/runtime_priv_test.lua` would look better, IMHO.

src/box/lua/config/applier/runtime_priv.lua
Totktonada: I see now, there is `no_exec` role. It verifies this case.

Totktonada: Ah, no. Sorry.

test/box-luatest/gh_10310_grant_rt_access_from_config_test.lua
Totktonada, line 1: I think that we can decouple testing mechanics from testing scenarios more. This way it should be much simpler to add more cases if needed. My experiment: [code](https://github.com/Totktonada/tarantool/blob/6150b1d847927eb864a0646ec138d63f6ad2c430/test/box-luatest/gh_10310_grant_rt_access_from_config_test.lua), [diff](https://github.com/Totktonada/tarantool/commit/6150b1d847927eb864a0646ec138d63f6ad2c430).

Aside of the main goal (move mechanical actions out of the scenarios) it is shorter and even faster (no restart to go to RO, just config reload).

src/box/lua/config/applier/runtime_priv.lua
mandesero: Test case added.

src/box/lua/config/applier/runtime_priv.lua
mandesero: Thank you! I find your suggestions valuable. I agree with improvements and have incorporated them

test/box-luatest/gh_10310_grant_rt_access_from_config_test.lua
mandesero: prefix `box` replaced with `config`. Also tests relocated to `test/config-luatest/runtime_priv_test.lua`

src/box/lua/config/applier/runtime_priv.lua
mandesero: I've made some minor changes, specifically - I believe that when an error is detected, we should not terminate the process using `error(...)`, as this causes a failure in cases like `config-luatest.credentials_applier.test_remove_excessive_privs_for_default_user` ([see job](https://github.com/tarantool/tarantool/actions/runs/11008586283/job/30566506868?pr=10392)). I suggest logging the error and halting the runtime privileges without performing grant/reset operations.

test/box-luatest/gh_10310_grant_rt_access_from_config_test.lua
mandesero: Sounds great, I used it. I also added the test case mentioned above.

src/box/lua/config/applier/runtime_priv.lua
Totktonada: This is a mistake in the recursion detection: it misinterprets a diamond-shape graph as a resursion. I would transform the BFS to a recursively implemented DFS and with tracking of the nodes on the road. My experiment (seems working): [code](https://github.com/Totktonada/tarantool/blob/5273fd9a327c2d85ea3ee747f1cfdc4472cfd45f/src/box/lua/config/applier/runtime_priv.lua#L25-L51), [diff](https://github.com/Totktonada/tarantool/commit/5273fd9a327c2d85ea3ee747f1cfdc4472cfd45f).

test/config-luatest/runtime_priv_test.lua
georgiy-belyanin, line 1: AFAIU, "needs upgrade" and RO mode behave quite similar right now.
Though, I suggest adding a test on an outdated schema to ensure runtime priv still works, because one of the main reasons to implement this is a way to perform calls before upgrading.

georgiy-belyanin (changes requested):

georgiy-belyanin (approved):

test/config-luatest/runtime_priv_test.lua
georgiy-belyanin: @mandesero I've changed the review status to "approved" in case of implementing such a test is too hard or you find it unnecessary.

Totktonada: Such a test case likely will not test anything after #10520. I see no much reason to bother with it unless you really want (in this case I don't mind).

mandesero: Since
>  a test case likely will not test anything after #10520.

I think we can resolve this conversation.

src/box/lua/config/applier/runtime_priv.lua
Totktonada: Looks solved now. Thanks for the update!

PR #10403, opened 2024-08-13
failover: revoke leader RW in supervised mode [WIP]

### LINKED ISSUE #10419: config: make bootstrap leader go RO mode after bootstrapping
The bootstrap leader has to go into the read-write mode right after bootstrapping to let all the triggers proceed. It may cause some problems. For example, ithe supervised failover right now assumes it's the only one instance allowed to manage other replica modes.

In order to deal with this, it's impossible to go RO straight after the bootstrapping. Some of the necessary triggers might not finish. The suggested way is to make the bootstrap leader go read-only mode after some safe amount of time that is also suitable for the failover.

It's also will be useful to add an internal way to detect if the replicaset was bootstrapped.

georgiy-belyanin:
Closes #10419

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -496,12 +496,28 @@ local function apply(config)
         --
         -- The RO/RW mode remains unchanged on reload.
         local am_i_bootstrap_leader = false
+        local is_bootstrapped =
+            snapshot.get_path(configdata._iconfig_def) == nil
+
         if is_startup then
             local instance_name = configdata:names().instance_name
+
             am_i_bootstrap_leader =
-                snapshot.get_path(configdata._iconfig_def) == nil and
+                is_bootstrapped and
                 instance_name == configdata:bootstrap_leader_name()
+
             box_cfg.read_only = not am_i_bootstrap_leader
+        end
+
+        local leader_rw_interval =
+            instance_config:get(configdata._iconfig_def,
+                                'failover.lease_interval')
+
+        if is_bootstrapped then
+            -- This variable is needed to determine the cases when
+            -- the RW mode was set not by the coordinator but by the
+            -- replicaset.
+            config._made_leader_rw_for_interval = leader_rw_interval
         end
 
         -- It is possible that an instance with the minimal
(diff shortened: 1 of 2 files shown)


src/box/lua/config/applier/box_cfg.lua
```diff
@@ -496,14 +496,35 @@ local function apply(config)
...
+                recently_bootstrapped and
                 instance_name == configdata:bootstrap_leader_name()
+
             box_cfg.read_only = not am_i_bootstrap_leader
         end
 
+        if recently_bootstrapped then
+            local renew_interval =
+                instance_config:get(configdata._iconfig_def,
+                                    'failover.renew_interval')
```
Totktonada, line 517: The coordinator grants RW for `failover.lease_interval`. I would do it here too.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -496,14 +496,35 @@ local function apply(config)
...
+
+            box.internal.recently_bootstrapped = recently_bootstrapped
+
+            fiber.new(function()
+                fiber.sleep(renew_interval)
+                box.cfg({read_only = true})
+                box.internal.recently_bootstrapped = nil
+            end)
```
Totktonada, line 525: I suggest to set a fiber name in the first statement of the fiber function.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -496,14 +496,35 @@ local function apply(config)
...
+
             box_cfg.read_only = not am_i_bootstrap_leader
         end
 
+        if recently_bootstrapped then
+            local renew_interval =
+                instance_config:get(configdata._iconfig_def,
+                                    'failover.renew_interval')
+
+            box.internal.recently_bootstrapped = recently_bootstrapped
```
Totktonada, line 519: Can we do it in some our namespace? For example, assign it somewhere to `configdata`.

Totktonada: Thought about it a bit more: maybe `configdata` is not a nice variant to place it, because a new object is created on each reconfiguration. It would be better to place it into some global/singleton object. Maybe `config` module.

src/box/lua/config/applier/box_cfg.lua
georgiy-belyanin: Applied.

src/box/lua/config/applier/box_cfg.lua
georgiy-belyanin: You're right. Thank you. Done.

src/box/lua/config/applier/box_cfg.lua
georgiy-belyanin: Probably temporary added it to the config in `config._recently_bootstrapped`. All of the discussed ideas don't seem good enough.

georgiy-belyanin:
Closing this PR as the issue for which it's been intended has been implemented in tarantool/tarantool-ee#863.

PR #10407, opened 2024-08-14
[backport 1.10] upgrade: fix upgrading from schema 1.6.8

### LINKED ISSUE #10180: Fix upgrading from version 1.6 to 1.10 and above

Serpentian:
Looks like a new release of 1.10


Serpentian:
Closing for now, hope we want go this way

PR #10418, opened 2024-08-14
config: fix replication.sync_timeout default value

### LINKED ISSUE #10280: Config wrongly sets default value of the `replication.sync_timeout`
Compat options are set after `instance_config.lua` is executed, so `replication_sync_timeout` is always 0, regardless of the compat option set.

I start the tarantool instance with the following config:

```yaml
compat:
     box_cfg_replication_synchro_timeout: 'new'
     box_cfg_replication_sync_timeout: 'old'

credentials:
    users:
        guest:
            roles: [super]

groups:
  group001:
    replicasets:
      replicaset001:
        instances:
          instance001:
            database:
              mode: rw
            iproto:
              listen:
              - uri: '127.0.0.1:3301'
```

`sync_timeout` must be 300, if compat option is `old`, however it's 0:

```
localhost:3301> box.cfg.replication_sync_timeout
---
- 0
...

localhost:3301> c = require('compat')
---
...

localhost:3301> c
---
- - box_cfg_replication_synchro_timeout: new
  - box_cfg_replication_sync_timeout: old
  - sql_priv: default (new)
  - binary_data_decoding: default (new)
  - fiber_channel_close_mode: default (new)
  - sql_seq_scan_default: default (new)
  - box_error_serialize_verbose: default (old)
  - box_error_unpack_type_and_code: default (old)
  - yaml_pretty_multiline: default (new)
  - box_info_cluster_meaning: default (new)
  - box_tuple_new_vararg: default (new)
  - c_func_iproto_multireturn: default (new)
  - box_tuple_extension: default (new)
  - box_session_push_deprecation: default (old)
  - json_escape_forward_slash: default (new)
  - box_space_max: default (new)
  - fiber_slice_default: default (new)
  - box_space_execute_priv: default (new)
  - console_session_scope_vars: default (old)
...
```

`box.NULL` must be used instead of that if condition in `default` value:

https://github.com/tarantool/tarantool/blob/a3ef8fb60382e747eed2cb521ebb3329dd4058f7/src/box/lua/config/instance_config.lua#L1473-L1478

mandesero:
This patch fixes a bug where the `box.replication_sync_timeout` field was being assigned a value that did not match the schema scenario. The problem was caused by the `schema` object being initialized before the fields were assigned values from the configuration, causing the field value to be calculated incorrectly.

Closes #10280

NO_DOC=bugfix
NO_CHANGELOG=bugfix


src/box/lua/config/applier/compat.lua
```diff
@@ -1,6 +1,44 @@
 local compat = require('compat')
 local log = require('internal.config.utils.log')
 
+-- Define a set of fields whose values cannot be directly assigned in
+-- `instance_config.lua` because they require validation based on the
+-- configuration state. These fields depend on certain conditions that
+-- need to be checked after the main configuration has been created.
```
Totktonada, line 7: I'm a bit worried about this approach. Here we transform the configuration data, but this transformation is not based on some schema annotation. If you're looking into the schema, it has no information that would suggest that the data are implicitly changed this way.

Moreover, the data is changed in the process of applying. It means that some code see the original variant of the configuration data and some code see the transformed one.

I propose to implement it in a more straightforward way. Let's set the `default` annotation to `box.NULL` to highlight that it is determined dynamically. And determine it in `apply()` in `applier/box_cfg`.

See how `replication.election_mode` is handled:

https://github.com/tarantool/tarantool/blob/12169c290b6cbed20ab60567b29da36bd8e5e6cf/src/box/lua/config/instance_config.lua#L1494-L1504

https://github.com/tarantool/tarantool/blob/12169c290b6cbed20ab60567b29da36bd8e5e6cf/src/box/lua/config/applier/box_cfg.lua#L417-L421

----

If you want to implement a general mechnism for effective defaults, it should be implemented on the `experimental.config.utils.schema` level.

Here we have different ways to achieve this goal. We can support `default` as a function or have a separate `effective_default` annotation as a function. It makes the schema less declarative, because this way it contains more Lua code (imperative, unserializable in the general case, can't be interpreted on different languages). OTOH, it is acceptable as a fast solution: we have `validate` implemented this way.

Some DSL for declarative effective default annotations would be nice, but it seems a way more complicated in implementation.

mandesero: Okay, let's focus on just the `replication.sync_timeout`.

Totktonada: I'm OK with fixing just the problem and look toward a general mechanism separately.

Current implementation replaces a user provided value with a default value, it is incorrect. Also, I've found that we can eliminate some extra code from the applier.

I would like to describe how I use this schema+applier machinery. Let's imagine that there is no `replication.sync_timeout` option and we're going to add it.

## Take 1

A straightforward way to do so is the following:

```lua
-- instance_config.lua
return schema.new('instance_config', schema.record({
    <...>
    replication = schema.record({
        <...>
        sync_timeout = schema.scalar({
            type = 'number',
        }),
    }),
    <...>
}, <...>), <...>)

-- applier/box.cfg.lua; apply() function
local sync_timeout = configdata:get('replication.sync_timeout',
    {use_default = true})
if sync_timeout == nil then -- nil or box.NULL
    sync_timeout = <..calculate default value..>
end
box_cfg.sync_timeout = sync_timeout
```

> [!NOTE]
> `{use_default = true}` has no effect, because there is no default value for this option, but I prefer to use it everywhere for consistency and to reduce mistakes when we forgot about the default. I think I should make it default behavior of `configdata:get()` sooner or later.

There are existing options that are implemented this way: for example, `log`.

## Take 2

Now we can observe that some of other options have the `box_cfg` annotation. This annotation allows to don't write the `box_cfg.foo_bar = configdata:get('foo.bar', <...>)` code for each option, but collect them in one loop:

https://github.com/tarantool/tarantool/blob/1923078b4f65649f03fdd6f789f29ba635b836b7/src/box/lua/config/applier/box_cfg.lua#L328-L332

Let's use the annotation and eliminate the `configdata:get()` call. We just need to handle the default value in the applier code.

```lua
-- instance_config.lua
return schema.new('instance_config', schema.record({
    <...>
    replication = schema.record({
        <...>
        -- An effective default is determined by the
        -- box_cfg applier, because it depends on compat
        -- options.
        sync_timeout = schema.scalar({
            type = 'number',
            box_cfg = 'replication_sync_timeout',         -- !! added
        }),
    }),
    <...>
}, <...>), <...>)

-- applier/box.cfg.lua; apply() function
if box_cfg.sync_timeout == nil then -- nil or box.NULL    -- !! changed
    box_cfg.sync_timeout = <..calculate default value..>  -- !! changed
end
```

`replication.peers` is handled this way.

## Take 3

Let's take a third step. We can look into the `box.cfg()` implementation and check, whether it has a proper default.

`box.cfg()` defaults are in the `default_cfg` table in `src/box/lua/load_cfg.lua`. In this case, the table is updated by a compat option action, see the `compat_options` table and the `compat.add_option` call below.

On the first glance, we can just drop the code that handles the default from the box_cfg applier:

```lua
-- !! This is incorrect example !!

-- instance_config.lua
return schema.new('instance_config', schema.record({
    <...>
    replication = schema.record({
        <...>
        -- An effective default is determined by the
        -- box.cfg() call itself.
        sync_timeout = schema.scalar({              -- !! unchanged
            type = 'number',
            box_cfg = 'replication_sync_timeout',
        }),
    }),
    <...>
}, <...>), <...>)

-- applier/box.cfg.lua; apply() function
<..no new code is needed..>                         -- !! dropped the new code
```

But there is a nuance: idempotence.

For example, we set `replication.sync_timeout = 5` in the YAML configuration, start the instance. Then we want to use the default value: we remove the option from the YAML config and call `config:reload()`.

The effect is that `box_cfg.replication_sync_timeout` is `nil`. The `box.cfg(box_cfg)` call interprets it as an instruction to leave the option unchanged.

We expect that the option is set to 0, but it remains to be 5. (If we restart the instance, the option is correctly set to 0. It is what I call a violation of the idempotence principle in this context.)

However, we can instruct `box.cfg()` to take the default without calculating it by itself:

https://github.com/tarantool/tarantool/blob/1923078b4f65649f03fdd6f789f29ba635b836b7/src/box/lua/load_cfg.lua#L841-L846

If `v` is `nil`, it doesn't appear in the loop. However, if it is `''` or `box.NULL`, a default value is set. (Don't be confused by the `v == nil` check: `box.NULL == nil`.)

It means that we can set the default in the schema to `''` or `box.NULL` and get the expected behavior. I use `box.NULL` for this purpose.

Finally, the result is the following:

```lua
-- instance_config.lua
return schema.new('instance_config', schema.record({
    <...>
    replication = schema.record({
        <...>
        sync_timeout = schema.scalar({
            type = 'number',
            box_cfg = 'replication_sync_timeout',
            -- box.cfg() chooses a default depending on the
            -- box_cfg_replication_sync_timeout compat option.
            default = box.NULL,                                   -- !! added
        }),
    }),
    <...>
}, <...>), <...>)

-- applier/box.cfg.lua; apply() function
<..no new code is needed..>
```

## Remarks

One can observe that we can set most of the default values to `box.NULL`, because the defaults are exist in the `default_cfg` table.

Technically, that's true.

However, the configuration schema should be the primary source of truth about the configuration format, constraints, defaults and so on. The schema may be used in various tools. For example, some code in `tt` is [generated](https://github.com/tarantool/tt/pull/806) using it. There are users that enable autocompletion for tarantool configuration files in an IDE using a JSON Schema [generated](https://gist.github.com/Totktonada/5a310dfee015e8ac68fb01c066f8c222) from it.

IOW, if we have more detailed schema that can be used outside tarantool's code itself, we have extra gains.

Taking it into account, I decided that the schema should *duplicate* defaults from `load_cfg.lua` where possible (and a test should verify that the values correspond each other).

Maybe we'll eliminate the duplication by adjusting `box.cfg()` to take defaults from `instance_config.lua` (where possible) in a future.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -343,6 +344,12 @@ local function apply(config)
         box_cfg.replication = peer_uris(configdata)
     end
 
+    local replication = configdata:get('replication', {use_default = true})
+    if replication ~= nil then
+        replication['sync_timeout'] =
+            compat.box_cfg_replication_sync_timeout:is_old() and 300 or 0
+    end
```
Totktonada, line 351: I've described all the details in the previous comment. Here I'll summarize problems and changes that I expect.

1. This code replaces a value that is written by a user in the YAML configuration with a default.
2. It seems, just `default = box.NULL` works correctly -- `load_cfg.lua` has a logic to determine the default depending on the compat option. We can drop this block of code from the applier.

mandesero: Removed code from `box_cfg.lua`. I also added a test that checks that `replication.sync_timeout` is changed to the default value after `config:reload()` (case from "Example 3:").

Totktonada (approved):

src/box/lua/config/instance_config.lua
```diff
@@ -1,6 +1,5 @@
 local schema = require('experimental.config.utils.schema')
 local tarantool = require('tarantool')
-local compat = require('compat')
```
Totktonada, line 3: > NO_CHANGELOG=bugfix

We have 'Bugs fixed' section in the release notes and this particular change is a user-visible one. I propose to add a changelog entry.

From a user perspective it looks like the `box_cfg_replication_sync_timeout` = `old` has no effect when a tarantool instance is started from a YAML configuration.

mandesero: Added.

changelogs/unreleased/gh-10280-fix-replication-sync-timeout-default-value.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/config
+
+* Fixed a bug where the default value for `box.replication_sync_timeout` was
+  not being set correctly.
```
lenkis, line 4:
```suggestion
  not set correctly.
```

src/box/lua/config/instance_config.lua
```diff
@@ -1473,8 +1472,9 @@ return schema.new('instance_config', schema.record({
         sync_timeout = schema.scalar({
             type = 'number',
             box_cfg = 'replication_sync_timeout',
-            default = compat.box_cfg_replication_sync_timeout:is_old()
-                and 300 or 0,
+            -- The effective default is determined depending of
```
lenkis, line 1475:
```suggestion
            -- The effective default is determined depending on
```

test/config-luatest/instance_config_schema_test.lua
```diff
@@ -1350,6 +1350,10 @@ g.test_box_cfg_coverage = function()
         -- The effective default is determined depending of
         -- the replication.failover option.
         election_mode = true,
+
+        -- The effective default is determined depending of
```
lenkis, line 1354:
```suggestion
        -- The effective default is determined depending on
```

lenkis (approved):

changelogs/unreleased/gh-10280-fix-replication-sync-timeout-default-value.md
Totktonada, line 4: Please, add ` (gh-xxxxx).` at the end.

mandesero: Added.

Totktonada:
Pushed to `master`, a future `3.2.0` release.

Backported to `release/3.1` (a future `3.1.2` release) in PR #10457.

Issue #10430, opened 2024-08-16
Implement compact field map


alyapunov:
Now field map consist of int32_t offsets. That seems to much for small tuples. It should be int8_t if that's enough.

PR #10434, opened 2024-08-16
Field map improvements

### LINKED ISSUE #10429: Store format of field map in each tuple
Tuple format is a complex thing that serves several purposes:
* Engine virtual table. Contains pointer to engine-specific tuple_free function.
* Tuple check and create: constraints (field count, field type, constraints etc), field defaults. These things are used only when a tuple is created and actually should not be stored in each tuple. That should be a part of space or similar.
* Field name dictionary. That a two-way converter field_name <=> field_no. Essential part of a tuple. By design this part in not thread-safe now so it must not be used in threads other than TX.
* Tuple layout. Now that part of tuple format only describes tuple field map. Perhaps it should also include information about non-common fields (like lsn in vinyl tuples).

Currently tuple format (with all these parts) is stored in each tuple by uint16_t format_id. In memtx engine in read this format_id could be overwritten by free list pointer and must not be used.

Union of these part into one tuple format has certain benefits, at least compact 16bit representation. But splitting it into parts has benefits too: for every part a special optimization can be applied. Also, for example, the part about tuple layout may be placed in the part of tuple that is not overwritten by free list pointer in memtx engine and thus can be used in other threads.

So I propose a small step: store format that describes field map right in field map, if the field map is not empty. In future this format id could be changed to some other id or more complex thing. Also, since field map is always located right before msgpack data, the only pointer to msgpack inside a tuple will be enough to get all required for getting tuple field or tuple comparison; perhaps it could simplify tuple API in future.

Of course it would cost additional 2 bytes in tuple with field map, but together with compact field map implementation it would affect only big tuples in which it would be negligible.

alyapunov:
Implement compact field map and make possible to use field map in different from TX threads.

Closes #10429 #10430


src/box/tuple.h
```diff
@@ -789,6 +789,29 @@ tuple_format(struct tuple *tuple)
 	return format;
 }
 
+static inline struct tuple_format *
+tuple_field_map_format(const char *tuple_data)
+{
+	field_map_format_id_t fmf_id;
+	if ((intptr_t)tuple_data % 4 == 2)
+		fmf_id = 0;
```
mkostoevr, line 797: It would be nice to have something like `FORMAT_ID_DEFAULT = 0` for the sake of maintainability.

mkostoevr (approved):
LGTM, just one nit.

mkostoevr:
The development is continued in the #10497.

PR #10438, opened 2024-08-16
connpool: make `call()` execution faster

### LINKED ISSUE #10596: connpool: support getting only alive connections in `filter()`
Currently the `connpool.filter()` method if the mode is unspecified checks only the static properties. Though getting only the alive instances may be useful.

The suggested solution is to implement 2 more different modes in addition to `ro/rw` in `filter()`:
- `alive` for filtering out unavailable instances.
- `any` (default) for no filtering by the instance status.

The same way `call()` should use the `alive` filtering mode by default.

Jira task: [TNTP-223](https://jira.vk.team/browse/TNTP-223)

### LINKED ISSUE #10598: connpool: make `call()` execute faster
Currently `connpool.call()` method connects to the instances through calling `filter()` internally and then iterating through them. The filter meanwhile tries to access all of the instances matching the requirements.

This call candidate selection can be done smarter if you connect to all of the instances in parallel and try to wait for the first one available matching the requirements.

It should match both static and dynamic (correct mode) requirements.

If one of the prioritized modes (`prefer_ro`, `prefer_rw`) are chosen now it's suggested to fetch all of the instances matching static requirements and then selecting the most suitable one (though, it can be done smarter, if we can be sure that some of the fetched instances has the highest priority possible).

Jira task: [TNTP-399](https://jira.vk.team/browse/TNTP-399)

georgiy-belyanin:
This patch consist of two changes in the connection pool. For sure, they should be two different PRs, but I've accidentally made one of them as a minor change in terms of the other not knowing there is an issue on it.

### Add support for filtering available instances using `connpool.filter()`.

The `connpool.filter()` now by default returns only available instances. This check can be skipped using a new `connpool.filter()` boolean option `skip_connection_check` (by default it equals `false`).

Filtering by `ro`/`rw` mode requires the connection check to be enabled.

Examples:
```lua
-- 'instance-001', 'instance-002' are available, 'instance-003' is down.

-- The table will contain 'instance-001' and 'instance-002'.
let alive = connpool.filter()

-- The table will contain 'instance-001', 'instance-002' and
-- 'instance-003'.
let all = connpool.filter({ skip_connection_check = true })
```

Closes #10596
Jira task: [TNTP-223](https://jira.vk.team/browse/TNTP-223)

### Speed up the execution of the `connpool.call()` method in default and `ro`, `rw` modes.
The connection pool `call()` function used to connect to all of the instances matching some specified static requirements and synchronously select one of the active. It was slower a lot times slower than its possible because the connpool only actually needs one active connection to perform the call.

This patch makes the connection pool `call()` function work a bit smarter. Now if the priorities (`prefer_ro`/`prefer_rw` call modes) aren't used within the `connpool.call()` the pool will wait for any connection matching the requirements as a candidate to perform the call.

Closes #10598
Jira task: [TNTP-399](https://jira.vk.team/browse/TNTP-399) 

Requires #10287 to be merged.

### DIFF
--- a/changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
+++ b/changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
@@ -1,0 +1,5 @@
+## feature/connpool
+
+* Now `connpool.filter()` returns only alive instances by default.
+  To use the old behaviour and acquire all instances the new option
+  `skip_connection_check` can be used (gh-10596).
--- a/changelogs/unreleased/gh-10598-make-connpool-call-faster.md
+++ b/changelogs/unreleased/gh-10598-make-connpool-call-faster.md
@@ -1,0 +1,4 @@
+## feature/connpool
+
+* `connpool.call()` method now works faster in general when it works in
+  modes different from `prefer_ro`/`prefer_rw` (gh-10598).
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -94,7 +94,11 @@ local function is_candidate_checked(candidate)
            conn.state == 'closed'
 end
 
+-- This method connects to all of the specified instances
+-- and returns the set of successfully connected ones.
 local function connect_to_candidates(candidates)
+    if next(candidates) == nil then return {} end
+
     local delay = WATCHER_DELAY
     local connect_deadline = clock.monotonic() + WATCHER_TIMEOUT
 
@@ -232,12 +236,22 @@ local function filter(opts)
         roles = '?table',
         sharding_roles = '?table',
         mode = '?string',
+        skip_connection_check = '?boolean',
     })
     opts = opts or {}
+
     if opts.mode ~= nil and opts.mode ~= 'ro' and opts.mode ~= 'rw' then
         local msg = 'Expected nil, "ro" or "rw", got "%s"'
         error(msg:format(opts.mode), 0)
     end
+
+    if opts.skip_connection_check and opts.mode ~= nil then
+        local msg = 'Filtering by mode "%s" requires the connection ' ..
+                    'check but it\'s been disabled by the ' ..
+                    '"skip_connection_check" option'
+        error(msg:format(opts.mode), 0)
+    end
+
     if opts.sharding_roles ~= nil then
         for _, sharding_role in ipairs(opts.sharding_roles) do
             if sharding_role == 'rebalancer' then
@@ -271,12 +285,16 @@ local function filter(opts)
             table.insert(static_candidates, instance_name)
         end
     end
-    -- Return if retrieving dynamic information is not required.
-    if next(static_candidates) == nil or next(dynamic_opts) == nil then
+
+    -- Return if the connection check isn't needed.
+    if opts.skip_connection_check then
         return static_candidates
     end
 
     -- Filter the remaining candidates after connecting to them.
+    --
+    -- The connect_to_candidates() call returns quickly if it
+    -- receives empty table as an argument.
     local connected_candidates = connect_to_candidates(static_candidates)
     local dynamic_candidates = {}
     for _, instance_name in pairs(connected_candidates) do
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -78,6 +78,15 @@ local function connect(instance_name, opts)
     return conn
 end
 
+-- Run connecting to the instance, which if it is not already
+-- connected or in process of connecting.
+local function connect_in_background(instance_name)
+    pcall(connect, instance_name, {
+        wait_connected = false,
+        connect_timeout = WATCHER_TIMEOUT
+    })
+end
+
 local function is_candidate_connected(candidate)
     local conn = connections[candidate]
     return conn and conn.state == 'active' and conn:mode() ~= nil
@@ -94,9 +103,12 @@ local function is_candidate_checked(candidate)
            conn.state == 'closed'
 end
 
--- This method connects to all of the specified instances
--- and returns the set of successfully connected ones.
-local function connect_to_candidates(candidates)
+-- This method connects to the specified instances and returns
+-- the set of successfully connected ones.
+--
+-- opts.any could be used to specify a condition /w
+local function connect_to_candidates(candidates, opts)
+    if opts == nil then opts = {} end
     if next(candidates) == nil then return {} end
 
     local delay = WATCHER_DELAY
@@ -117,11 +129,16 @@ local function connect_to_candidates(candidates)
             :filter(is_candidate_connected)
             :totable()
 
-        local all_checked = fun.iter(candidates)
-            :all(is_candidate_checked)
+        local stop
+        if opts.any then
+            assert(type(opts.any) == 'function')
+            stop = fun.iter(candidates):any(opts.any)
+        else
+            stop = fun.iter(candidates):all(is_candidate_checked)
+        end
 
-        if all_checked then
-            return connected_candidates
+        if stop then
+            break
         end
 
         connection_mode_update_cond:wait(delay)
@@ -224,6 +241,21 @@ end
 
 local function is_candidate_match_dynamic(instance_name, opts)
     assert(opts ~= nil and type(opts) == 'table')
+
+    local conn = connections[instance_name]
+
+    -- No connection -> reconnect in background, exclude the candidate.
+    if conn == nil or conn.state == 'error' or conn.state == 'closed' then
+        connect_in_background(instance_name)
+        return false
+    end
+
+    -- In progress -> exclude the candidate.
+    if not (conn.state == 'active' and conn:mode() ~= nil) then
+        return false
+    end
+
+    -- Connected -> check using the provided filters.
     return is_mode_match(opts.mode, instance_name)
 end
 
@@ -237,8 +269,24 @@ local function filter(opts)
         sharding_roles = '?table',
         mode = '?string',
         skip_connection_check = '?boolean',
+        -- Internal option on how filter() behaves.
+        -- Used internally in the call() method.
+        --
+        -- The value is represented by enum (the values are sorted
+        -- from the fastest to the slowest):
+        -- * `no wait` -- only return already polled instances.
+        -- * `any` -- wait for any instance satisfying requirements
+        --   to be polled.
+        -- * `wait` (default) -- wait for all instances to be
+        --   polled.
+        _wait_mode = '?string',
     })
     opts = opts or {}
+
+    local wait_mode = opts._wait_mode
+    if wait_mode == nil then
+        wait_mode = 'wait'
+    end
 
     if opts.mode ~= nil and opts.mode ~= 'ro' and opts.mode ~= 'rw' then
         local msg = 'Expected nil, "ro" or "rw", got "%s"'
@@ -291,11 +339,28 @@ local function filter(opts)
         return static_candidates
     end
 
-    -- Filter the remaining candidates after connecting to them.
-    --
-    -- The connect_to_candidates() call returns quickly if it
-    -- receives empty table as an argument.
-    local connected_candidates = connect_to_candidates(static_candidates)
+    local connected_candidates
+    if wait_mode == 'wait' then
+        -- Filter the remaining candidates after connecting to them.
+        --
+        -- The connect_to_candidates() call returns quickly if it
+        -- receives empty table as an argument.
+        connected_candidates = connect_to_candidates(static_candidates)
+    elseif wait_mode == 'any' then
+        -- Try to connect to the candidates until at least one
+        -- satisfying the dynamic requirements is found.
+        connected_candidates = connect_to_candidates(static_candidates, {
+            any = function(instance_name)
+                return is_candidate_match_dynamic(instance_name, dynamic_opts)
+            end,
+        })
+    elseif wait_mode == 'no wait' then
+        -- Only continue with already polled instances.
+        connected_candidates = static_candidates
+    else
+        assert(false, 'Unexpected _wait_mode value.')
+    end
+
     local dynamic_candidates = {}
     for _, instance_name in pairs(connected_candidates) do
         if is_candidate_match_dynamic(instance_name, dynamic_opts) then
@@ -319,16 +384,36 @@ local function get_connection(opts)
         sharding_roles = opts.sharding_roles,
         mode = mode,
     }
-    local candidates = filter(candidates_opts)
+
+    -- If the mode is not prefer_*, try to find already connected
+    -- candidates.
+    local candidates
+    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
+        candidates = filter(candidates_opts)
+    else
+        -- Try to find already connected candidates.
+        candidates_opts._wait_mode = 'no wait'
+        candidates = filter(candidates_opts)
+
+        -- Try to find at least one connected otherwise.
+        if next(candidates) == nil then
+            candidates_opts._wait_mode = 'any'
+            candidates = filter(candidates_opts)
+        end
+
+        -- Last resort: try to connect to all the candidates.
+        if next(candidates) == nil then
+            candidates_opts._wait_mode = 'wait'
+            candidates = filter(candidates_opts)
+        end
+    end
+
     if next(candidates) == nil then
         return nil, "no candidates are available with these conditions"
     end
 
     -- Initialize the weight of each candidate.
     local weights = {}
-    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
-        candidates = connect_to_candidates(candidates)
-    end
     for _, instance_name in pairs(candidates) do
         weights[instance_name] = 0
     end
--- a/test/config-luatest/rpc_test.lua
+++ b/test/config-luatest/rpc_test.lua
@@ -993,3 +993,174 @@ end
 g.after_test('test_call_works_in_parallel', function(g)
     stop_stub_servers(g)
 end)
+
+g.test_call_picks_any = function(g)
+    local dir = treegen.prepare_directory({}, {})
+    local config = [[
+    credentials:
+      users:
+        guest:
+          roles: [super]
+        myuser:
+          password: "secret"
+          roles: [replication]
+          privileges:
+          - permissions: [execute]
+            universe: true
+
+    iproto:
+      listen:
+        - uri: 'unix/:./{{ instance_name }}.iproto'
+      advertise:
+        peer:
+          login: 'myuser'
+
+    groups:
+      group-001:
+        replicasets:
+          replicaset-001:
+            instances:
+              instance-001:
+                database:
+                  mode: rw
+              instance-003: {}
+          replicaset-002:
+            instances:
+              instance-002: {}
+    ]]
+    treegen.write_file(dir, 'config.yaml', config)
+
+    local opts = {
+        env = {LUA_PATH = os.environ()['LUA_PATH']},
+        config_file = 'config.yaml',
+        chdir = dir,
+    }
+    g.server_1 = server:new(fun.chain(opts, {alias = 'instance-001'}):tomap())
+    g.server_3 = server:new(fun.chain(opts, {alias = 'instance-003'}):tomap())
+
+    g.server_1:start({wait_until_ready = false})
+    g.server_3:start({wait_until_ready = false})
+    start_stub_servers(g, dir, { 'instance-002' })
+
+    g.server_1:wait_until_ready()
+    g.server_3:wait_until_ready()
+
+    g.server_1:exec(function()
+        local connpool = require('experimental.connpool')
+        local clock = require('clock')
+
+        -- Connect to the local instance to avoid possible race
+        -- conditions when connecting to non-local instance could
+        -- be faster than connecting to a local one.
+        connpool.connect('instance-001')
+
+        -- The connection timeout for filter() and call() methods
+        -- is hardcoded in connpool and equals 10 seconds.
+        local CONNECT_TIMEOUT = 10
+
+        -- call() should wait for any available instance matching
+        -- the requirements and not wait for the unavailable instance
+        -- connection attempts to be timeouted.
+        --
+        -- By default it calls the local one.
+        -- The maximum  interval is taken with a great extra time
+        -- (approx. 9 sec) to avoid flaky tests.
+        local timestamp_before_call = clock.monotonic()
+        t.assert_equals(connpool.call('box.info', nil, {}).name,
+                        'instance-001')
+        t.assert_lt(clock.monotonic() - timestamp_before_call,
+                    CONNECT_TIMEOUT * 0.9)
+
+        -- The same quick behaviour is applied to 'ro/rw' modes.
+        local timestamp_before_call = clock.monotonic()
+        t.assert_equals(connpool.call('box.info', nil, { mode='ro' }).name,
+                        'instance-003')
+        t.assert_lt(clock.monotonic() - timestamp_before_call,
+                    CONNECT_TIMEOUT * 0.9)
+    end)
+end
+g.after_test('test_call_picks_any', function(g)
+    stop_stub_servers(g)
+end)
+
+g.test_call_connects_to_all_if_prioritized = function(g)
+    local dir = treegen.prepare_directory({}, {})
+    local config = [[
+    credentials:
+      users:
+        guest:
+          roles: [super]
+        myuser:
+          password: "secret"
+          roles: [replication]
+          privileges:
+          - permissions: [execute]
+            universe: true
+
+    iproto:
+      listen:
+        - uri: 'unix/:./{{ instance_name }}.iproto'
+      advertise:
+        peer:
+          login: 'myuser'
+
+    groups:
+      group-001:
+        replicasets:
+          replicaset-001:
+            instances:
+              instance-001:
+                database:
+                  mode: rw
+              instance-003: {}
+          replicaset-002:
+            instances:
+              instance-002: {}
+              instance-004: {}
+    ]]
+    treegen.write_file(dir, 'config.yaml', config)
+
+    local opts = {
+        env = {LUA_PATH = os.environ()['LUA_PATH']},
+        config_file = 'config.yaml',
+        chdir = dir,
+    }
+    g.server_1 = server:new(fun.chain(opts, {alias = 'instance-001'}):tomap())
+    g.server_3 = server:new(fun.chain(opts, {alias = 'instance-003'}):tomap())
+
+    g.server_1:start({wait_until_ready = false})
+    g.server_3:start({wait_until_ready = false})
+    start_stub_servers(g, dir, { 'instance-002', 'instance-004' })
+
+    g.server_1:wait_until_ready()
+    g.server_3:wait_until_ready()
+
+    g.server_1:exec(function()
+        local connpool = require('experimental.connpool')
+        local clock = require('clock')
+
+        -- The connection timeout for filter() and call() methods
+        -- is hardcoded in connpool and equals 10 seconds.
+        local CONNECT_TIMEOUT = 10
+
+        -- call() with some of the prioritized modes (e.g "prefer_ro" or
+        -- "prefer_rw") should try to access all of the instances
+        -- parallelly.
+        local opts = { mode = 'prefer_ro' }
+        local timestamp_before_call = clock.monotonic()
+        t.assert_equals(connpool.call('box.info', nil, opts).name,
+                        'instance-003')
+        local elapsed_time = clock.monotonic() - timestamp_before_call
+
+        -- If connpool works as intended it should only wait a
+        -- single CONNECT_TIMEOUT.
+        --
+        -- The maximum  interval is taken with a great extra time
+        -- (approx. 9 sec) to avoid flaky tests.
+        t.assert_gt(elapsed_time, CONNECT_TIMEOUT * 0.9)
+        t.assert_lt(elapsed_time, CONNECT_TIMEOUT * 1.9)
+    end)
+end
+g.after_test('test_call_connects_to_all_if_prioritized', function(g)
+    stop_stub_servers(g)
+end)
--- a/test/config-luatest/rpc_test.lua
+++ b/test/config-luatest/rpc_test.lua
@@ -30,7 +30,7 @@ end
 
 local function stop_stub_servers(g)
     for _, server in pairs(g.stub_servers) do
-        server:close()
+        pcall(server.close, server)
     end
 end
 
@@ -411,6 +411,16 @@ g.test_filter_mode = function(g)
         local opts = {mode = 'rw'}
         t.assert_items_equals(connpool.filter(opts), exp)
 
+        exp = {"instance-001", "instance-002", "instance-003",
+               "instance-004"}
+        opts = {}
+        t.assert_items_equals(connpool.filter(opts), exp)
+
+        exp = {"instance-001", "instance-002", "instance-003",
+               "instance-004", "instance-005"}
+        opts = {skip_connection_check = true}
+        t.assert_items_equals(connpool.filter(opts), exp)
+
         exp = {"instance-002", "instance-004"}
         opts = {mode = 'ro'}
         t.assert_items_equals(connpool.filter(opts), exp)
@@ -429,6 +439,12 @@ g.test_filter_mode = function(g)
 
         local exp_err = 'Expected nil, "ro" or "rw", got "something"'
         opts = {mode = 'something'}
+        t.assert_error_msg_equals(exp_err, connpool.filter, opts)
+
+        local exp_err = 'Filtering by mode "rw" requires the connection ' ..
+                        'check but it\'s been disabled by the ' ..
+                        '"skip_connection_check" option'
+        opts = {mode = 'rw', skip_connection_check = true}
         t.assert_error_msg_equals(exp_err, connpool.filter, opts)
     end
 


src/box/lua/connpool.lua
```diff
@@ -202,8 +271,10 @@ local function filter(opts)
         mode = '?string',
     })
     opts = opts or {}
-    if opts.mode ~= nil and opts.mode ~= 'ro' and opts.mode ~= 'rw' then
-        local msg = 'Expected nil, "ro" or "rw", got "%s"'
+    if opts.mode ~= nil and opts.mode ~= 'alive' and opts.mode ~= 'all' and
+        opts.mode ~= 'ro' and opts.mode ~= 'rw' then
+        local msg = 'Incorrect filter mode "%s", allowed values: '..
```
georgiy-belyanin, line 276: It's suggested to rename the `alive` to `any`.

georgiy-belyanin:
Missing tests. It's not so hard to test the changes.

src/box/lua/connpool.lua
georgiy-belyanin: Done.

test/config-luatest/rpc_test.lua
```diff
@@ -780,3 +803,104 @@ g.test_call_mode = function(g)
...
+    end)
+    treegen.write_file(dir, 'one.lua', role)
+
+    local opts = {
+        env = {LUA_PATH = os.environ()['LUA_PATH']},
+        config_file = 'config.yaml',
+        chdir = dir,
+    }
+    g.server_1 = server:new(fun.chain(opts, {alias = 'instance-001'}):tomap())
+    g.server_2 = server:new(fun.chain(opts, {alias = 'instance-002'}):tomap())
```
georgiy-belyanin, line 869: Probably this call isn't needed.

changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
```diff
@@ -0,0 +1,4 @@
+## feature/connpool
+
+- Now `connpool.filter()` supports filtering only alive instances by
+  specifying `'alive'` mode (gh-10496).
```
Totktonada, line 4: The issue number seems wrong.

src/box/lua/connpool.lua
```diff
@@ -125,6 +125,42 @@ local function connect_to_candidates(candidates)
...
+        local ok, conn = pcall(connect, instance_name, {
+            wait_connected = false,
+            connect_timeout = WATCHER_TIMEOUT
+        })
+
+        if ok then
+            conn:on_connect(function()
+                acquire_any_cond:broadcast()
+            end)
+        end
```
Totktonada, line 148: `connect` may return a cached (already connected) connection. At least, we don't need the `on_connect` callback in the case.

Also, I'm wonder should we unregister the callback when it is unneeded? I guess it is saved and called on a reconnection. So, if we don't unregister it, it is present forever (if my guess it right). Maybe it is not a real problem, just noted to don't forget to think about it.

src/box/lua/connpool.lua
```diff
@@ -125,6 +125,42 @@ local function connect_to_candidates(candidates)
...
 
+local acquire_any_cond = nil
+local function acquire_any_candidate(instance_names, opts)
+    assert(type(opts) == 'table' and opts.mode)
+    assert(opts.mode == 'ro' or opts.mode == 'rw' or opts.mode =='alive')
+    local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
+
+    if not acquire_any_cond then
+        acquire_any_cond = fiber.cond()
+    end
```
Totktonada, line 136: We have a similar `connection_mode_update_cond`, which is signaled by the `connect` function. This cond seems mostly same, but it is about a different phase. Can we implement them in a similar way: signal both from `connect`?

Also, it seems that `:wait_state()` that uses an internal cond may fit our needs for waiting `alive` (and even better -- it returns immediately on other terminal states). And we can reuse `connection_mode_update_cond` when waiting for a particular RO/RW mode.

What do you think?

src/box/lua/connpool.lua
```diff
@@ -271,46 +316,102 @@ local function filter(opts)
             table.insert(static_candidates, instance_name)
         end
     end
-    -- Return if retrieving dynamic information is not required.
-    if next(static_candidates) == nil or next(dynamic_opts) == nil then
+
+    if mode == 'any' then
```
Totktonada, line 320: The comment seems useful, especially because a general 'if there are dynamic opts' check now looks as a `mode` check. Maybe is even worth to explicitly define some `has_dynamic_filters` predicate.

Why the `static_candidates` check for emptiness is removed?

src/box/lua/connpool.lua
```diff
@@ -271,46 +316,102 @@ local function filter(opts)
...
         return static_candidates
     end
 
-    -- Filter the remaining candidates after connecting to them.
-    local connected_candidates = connect_to_candidates(static_candidates)
+    local connected_instances = acquire_all_candidates(static_candidates)
+
```
Totktonada, line 325: This comment also seems useful: it marks the point where the actual network activity is performed. Helpful when you're not deeply in context of this code.

src/box/lua/connpool.lua
```diff
@@ -271,46 +316,102 @@ local function filter(opts)
...
+    if opts.mode ~= 'prefer_rw' and opts.mode ~= 'prefer_ro' then
+        if prefer_local then
+            for _, instance_name in ipairs(static_candidates) do
+                if instance_name ~= local_instance_name then
+                    goto continue
+                end
+
+                local conn = connect(local_instance_name,
+                                     { wait_connected = true })
+                assert(conn.state == 'active')
```
Totktonada, line 382: TBH, I don't get why it have to succeed.

src/box/lua/connpool.lua
```diff
@@ -125,6 +125,42 @@ local function connect_to_candidates(candidates)
...
+            end)
+        end
+    end
+
+    while clock.monotonic() < time_connect_end do
+        for _, instance_name in ipairs(instance_names) do
+            local conn = connections[instance_name]
+            if is_connection_valid(conn, {}) and
+               (opts.mode == 'alive' or opts.mode == conn:mode()) then
+                return conn
```
Totktonada, line 156: We generally return connections that're already have `mode`: we have to do it if we're going to use dynamic filter later. We don't do it here for the `opts.mode == 'alive'` case.

This function is specific: its only caller doesn't use the `mode` information. I think we need to explicitly clarify this nuance.

src/box/lua/connpool.lua
```diff
@@ -271,46 +316,102 @@ local function filter(opts)
...
+    end
+
+    -- In case there are specified prioritizing of matching
+    -- instances we fetch all matched static candidates.
+    -- Technically, it's possible to speedup fetching the instances
+    -- if there is already a connection to a top-priority connection
+    -- (e.g. one RO instance is established and the mode is
+    -- 'prefer_ro')
+    local candidates = acquire_all_candidates(static_candidates)
     if next(candidates) == nil then
```
Totktonada, line 409: It seems meaningful to exit earlier if there are no static candidates. Now the check is done only after connecting. The reason is that `acquire_all_candidates` is no-op on an empty list of candidates?

test/config-luatest/rpc_test.lua
```diff
@@ -16,7 +16,7 @@ local function skip_if_no_vshard()
 end
 
 g.test_connect = function(g)
-    local dir = treegen.prepare_directory({}, {})
+    local dir = treegen.prepare_directory(g, {}, {})
```
Totktonada, line 19: It seems it is an artifact of rebasing after bd27df009c403e89c003d5b66763c0f0bbf08440.

test/config-luatest/rpc_test.lua
```diff
@@ -780,3 +803,104 @@ g.test_call_mode = function(g)
...
+              instance-001:
+                database:
+                  mode: rw
+              instance-003: {}
+          replicaset-002:
+            instances:
+              instance-002:
+                iproto:
+                  listen:
+                    - uri: 9.0.0.0:8888
```
Totktonada, line 844: [Bad](https://github.com/tarantool/tarantool/pull/10287#discussion_r1773819957) idea.

src/box/lua/connpool.lua
Totktonada: I guess, `all` to `any`. OK.

test/config-luatest/rpc_test.lua
```diff
@@ -780,3 +803,104 @@ g.test_call_mode = function(g)
...
+
+        rawset(_G, 'f', f)
+
+        return {
+            stop = function() end,
+            apply = function() end,
+            validate = function() end,
+        }
+    end)
+    treegen.write_file(dir, 'one.lua', role)
```
Totktonada, line 861: Is it to acquire `box.info.name`? Can we just call `box.info` and get it from it?

test/config-luatest/rpc_test.lua
Totktonada: Can't spot the idea.

test/config-luatest/rpc_test.lua
Totktonada, line 1: Should we add some `filter({mode = 'alive'})` test case that would give a result different from `filter({mode = 'any'})`?

test/config-luatest/rpc_test.lua
```diff
@@ -780,3 +803,104 @@ g.test_call_mode = function(g)
...
+        t.assert(clock.monotonic() - timestamp_before_call <
+                 CONNECT_TIMEOUT / 2)
+
+        -- call() with some of the prioritized modes (e.g "prefer_ro" or
+        -- "prefer_rw") should try to access all of the instances.
+        opts = { prefer_local = true, mode = 'prefer_ro' }
+        timestamp_before_call = clock.monotonic()
+        t.assert_equals(connpool.call('f', nil, opts), 'instance-003')
+        t.assert(clock.monotonic() - timestamp_before_call >
+                 CONNECT_TIMEOUT / 2)
```
Totktonada, line 902: Maybe it worth to make this test case a separate one, because an implementation may re-use an information about the recent connection attempt.

If you're worrying about amount of boilerplate code, check out my `cbuilder` and `cluster` helpers. Not an ad :)

Totktonada:
I don't like that this module tends to be more and more complex, but the idea of the changes is clear for me (thanks!).

I hope that we'll reduce a complexity of the filtering code when the status information is collected in background -- in fact, all the filters will be static after it (except, maybe, a first access to the given instances).

Anyway, we're not going to implement the background state tracking within this pull request, because it is a significant task on its own.

I'm generally OK with the changes, but I would glance on it once again after the fixes.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Thank you. We've discussed the idea of creating a separate test case in #10287. 
`cbuilder` and `cluster` are great, though I avoided them for the consistency with the other tests.

I'm gonna write a separate issue on adding a timeout option for `call()` and `filter()` if you ok with it.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Since there is a stub TCP server creation here, this comment is no longer related.

changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
georgiy-belyanin: Thank you. Fixed.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you. 

I've added the comment line on skipping connection phase if the dynamic props not needed back. 

For some reason I though that `acquire_all_candidates()` will just pass if it receives an empty array. It didn't work so. 

I've left the `static_candidates` emptiness check removed but made `acquire_all_candidates()` just pass if it receives an empty array of instances. It seems for me this is a little bit more consistent.

test/config-luatest/rpc_test.lua
Totktonada: Here I meant a separate test case (function), not a separate file. The reason is not an execution time, but a possible re-using of past connect/filter information in a future implementation, so it will be much faster than expected.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Fixed this problems by starting stub tcp servers. Thank you.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Actually, there is a test on lines `414-424` in `config_luatest/rpc_test.lua`

test/config-luatest/rpc_test.lua
georgiy-belyanin: Fixed. Thank you.

src/box/lua/connpool.lua
georgiy-belyanin: Since I've made `acquire_all_candidates()` return empty list in case there is no parameters it works ok now.

We get to that line in case of `mode = prefer_*` and if there is no candidates the `acquire_all_candidates()` instantly returns.

src/box/lua/connpool.lua
georgiy-belyanin: Oh, didn't think of it. I wrapped it with a pcall and now check if everything is ok. Also, I've added a comment that the connect is still `wait_connected` since the call is almost always successfull.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you.
I think you're right that the callback should be dropped. Added dropping the callback inside the callback.
Also, now the hook isn't set in case of the connection was already checked.

src/box/lua/connpool.lua
georgiy-belyanin: Unfortunately, it's not possible to use here without using extra fibers.

test/config-luatest/rpc_test.lua
georgiy-belyanin: Oh, I'm sorry. Thanks. Changed it and implemented like this.

test/config-luatest/rpc_test.lua
georgiy-belyanin: It'll be a lot simpler and shorter. Implemented it like this. Thank you!

src/box/lua/connpool.lua
georgiy-belyanin: Oh, I see. Thank you. Probably, it's guaranteed to be available in some obvious way. Though, I didn't manage to make sure. Added the comments describing the behaviour of the `acquire_***` methods and the corresponding note for specifying possible absence of the `mode()` method.

changelogs/unreleased/gh-10598-make-connpool-call-faster.md
```diff
@@ -0,0 +1,3 @@
+## feature/connpool
+
+- `connpool.call()` method now works faster (gh-10598).
```
grafin, line 3: I don't like that there are two features in one commit. Furthermore I don't agree that 'works faster' is a feature at all. And as far as I understood the changes, it will only work faster in certain mode, while this changelog isn't clear about that.

I propose we merge this changelog with the other one, just adding there, that this 'alive' mode is faster than the others.

src/box/lua/connpool.lua
```diff
@@ -94,11 +94,15 @@ local function is_candidate_checked(candidate)
            conn.state == 'closed'
 end
 
-local function connect_to_candidates(candidates)
+-- This method connects to all of the specified instances
+-- and returns the set of successfully connected ones.
+local function acquire_all_candidates(instance_names)
```
grafin, line 99: For me the renaming of this function is not needed. Even in the comment you say that this function 'connects'.

src/box/lua/connpool.lua
```diff
@@ -94,11 +94,15 @@ local function is_candidate_checked(candidate)
...
+    if next(instance_names) == nil then return {} end
+
     local delay = WATCHER_DELAY
     local connect_deadline = clock.monotonic() + WATCHER_TIMEOUT
 
-    for _, instance_name in pairs(candidates) do
+    for _, instance_name in pairs(instance_names) do
```
grafin, line 105: Again, what is the reason behind renaming the parameter from `candidates` to `instance_names`?

src/box/lua/connpool.lua
```diff
@@ -125,6 +129,51 @@ local function connect_to_candidates(candidates)
...
+-- The method starts connecting to the specified set of
+-- instances and returns the first one available matching
+-- the specified dynamic requirements.
+--
+-- Note: if the specified mode is 'alive' the connection
+-- returned may not have the `mode()` method available.
+local function acquire_any_candidate(instance_names, opts)
+    assert(type(opts) == 'table' and opts.mode)
+    assert(opts.mode == 'ro' or opts.mode == 'rw' or opts.mode =='alive')
+    local time_connect_end = clock.monotonic() + WATCHER_TIMEOUT
```
grafin, line 143: Lets name it something like `connect_deadline` or something similar.

src/box/lua/connpool.lua
```diff
@@ -271,46 +329,111 @@ local function filter(opts)
...
     return dynamic_candidates
 end
 
+-- Subroutine for the `call()` function. Tries to find an active
+-- connection matching the requirements and considering the
+-- specified priorities.
 local function get_connection(opts)
-    local mode = nil
+    assert(opts.mode == 'alive' or opts.mode == 'ro' or opts.mode == 'rw' or
+           opts.mode == 'prefer_ro' or opts.mode == 'prefer_rw')
```
grafin, line 356: Why don't we separate the `is_alive` option from the mode? Then we could have something like `opts = {mode = 'rw', is_alive = true}` or `opts = {mode = 'prefer_ro', is_alive = true}`. For me this makes it a lot more flexible for the user and a lot more easy to understand.

grafin (changes requested):
I see some major problems with the current approach, which I've commented. I didn't review the tests and the inner logic, because I think that the API should be changed which would lead to major refactoring of the patches. Feel free to contact me to have a discussion.

src/box/lua/connpool.lua
Totktonada: `mode` = `rw` has no sense if `is_alive` = `false`. The same for `ro`, `prefer_*` -- all the mode values. So, I'm looking on it this way:

* `mode` = `any` -- no constraints
* `mode` = `alive` -- at least alive
* `mode` = `ro`/`rw` -- alive and has the given mode
* `mode` - `prefer_ro`/`prefer_rw` -- the same as previous, but with the give priority

Totktonada: We discussed it with @grafin and @georgiy-belyanin today. A summary is below.

* `connpool.call()` doesn't require any changes. We agreed to don't introduce any new options or option's values to it.
* The `connpool.filter()` behavior may be changed to return all alive instances by default. The experimenal status of the module allows it and we don't expect problems from this change. Moreover, we guess that this behavior is likely what early adopters actually expect.
* A separate option to return all instances disregarding its connection status is to be introduced for `connpool.filter()`.
* We considered several options:
  * Add it to the `mode` enumeration.
  * Add it into a newly created `ready` enumeration (to add more interesting readiness checks later to it).
  * Add a separate boolean option.
* The latter option looks the most safe and straightforward. We agreed on it.
* The name may be `require_connected` or `require_alive`[^1]. Not necessary some ideal name, because later we can revisit it in context of a more general readiness checking solution -- and deprecate this option.

[^1]: After the discussion I thought that it is more intuitive, when the default is `false`, so maybe it worth to consider `allow_disconnected` or `skip_connection_check`.

src/box/lua/connpool.lua
Totktonada: > Also, now the hook isn't set in case of the connection was already checked.

OK.

> I think you're right that the callback should be dropped. Added dropping the callback inside the callback.

It seems, you tried to drop the trigger and hit #10622. OK to don't do it so.

src/box/lua/connpool.lua
Totktonada: OK regarding the `:wait_state()` proposal: it is per-connection, but we need one condvar. I missed it.

But the proposal regarding consistency with `connection_mode_update_cond` still actual. What do you think about it?

test/config-luatest/rpc_test.lua
Totktonada: I see now, thanks!

src/box/lua/connpool.lua
```diff
@@ -125,6 +129,51 @@ local function connect_to_candidates(candidates)
...
+            end
+
+            conn:on_connect(on_connect_notify)
+        end
+    end
+
+    while clock.monotonic() < time_connect_end do
+        for _, instance_name in ipairs(instance_names) do
+            local conn = connections[instance_name]
+            if is_connection_valid(conn, {}) and
```
Totktonada, line 167: Just curious: if a connection in in a terminal state (for example, `error`), it means that `is_connection_valid` == `false`, but there is nothing to wait. Should we also check `is_connection_checked`?

src/box/lua/connpool.lua
Totktonada: OK, this problem is resolved. I'm not sure that we should handle it this way, but I'll open a separate comment.

src/box/lua/connpool.lua
```diff
@@ -271,46 +329,111 @@ local function filter(opts)
...
+                -- This connect is almomst always quick and successful,
+                -- so it's being waited. Though it can fail e.g. due to
+                -- the FD limit. That's why it still should be checked.
+                local ok, conn = pcall(connect, local_instance_name,
+                                       { wait_connected = true })
+
+                if not ok then break end
+
+                if is_candidate_match_dynamic(local_instance_name,
+                                              dynamic_opts) then
```
Totktonada, line 404: AFAIR, `wait_connected = true` doesn't guarantee that `:mode()` already has ro/rw state to filter on it. So the `is_candidate_match_dynamic` may return a wrong result (false negative).

It doesn't look as a problem, in fact (we'll just use another connection), but at least worth to comment.

However, I would better skip the connection explicitly if the mode is not received yet.

src/box/lua/connpool.lua
Totktonada: Opened https://github.com/tarantool/tarantool/pull/10438#discussion_r1779330155 and https://github.com/tarantool/tarantool/pull/10438#discussion_r1779330694.

src/box/lua/connpool.lua
georgiy-belyanin: This renaming has been made because I've wanted `acquire_***` function names to be more consistent internally.

src/box/lua/connpool.lua
georgiy-belyanin: Applied. Thank you.

changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
```diff
@@ -0,0 +1,4 @@
+## feature/connpool
+
+- Now `connpool.filter()` supports filtering only alive instances by
+  specifying `'alive'` mode (gh-10596).
```
lenkis, line 4:
```suggestion
- Now `connpool.filter()` supports a new filtering mode, `'alive'`, for filtering only 
  alive instances (gh-10596).
```

lenkis (approved):

changelogs/unreleased/gh-10596-connpool-filter-support-alive-filtering.md
georgiy-belyanin: Thank you. I've had to change the whole message due to the PR update. Please, review it one more time.

src/box/lua/connpool.lua
```diff
@@ -271,46 +329,111 @@ local function filter(opts)
...
+                    return conn
+                else
+                    break
+                end
```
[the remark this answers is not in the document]
georgiy-belyanin: Hm, the latter seems more logical. Though, I suppose it's ok to try to establish the connection if it doesn't exist. This should be quite fast and likely to be done only once. I see no huge difference between these approaches for a user except for the first call (that can execute just a little bit slower, in case of the local connection isn't established plus if the network is fast enough to transmit the call to an existing connection quicker than establishing a new local).

src/box/lua/connpool.lua
georgiy-belyanin: Now, I see what you meant. I am sorry I wasn't able to understand it earlier. Applied it. Nice point! Thanks.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you for writing down the points and options we've discussed. I've implemented the suggested behaviour using the `skip_connection_check` option.

changelogs/unreleased/gh-10598-make-connpool-call-faster.md
georgiy-belyanin: Thank you. I've splitted the patch int two distinct commits. Changed the changelog into a bit more specific that the `call()` is now faster only in some specific cases.

src/box/lua/connpool.lua
georgiy-belyanin: Actually, I've couldn't just leave the naming as it was. 

For the consistency I've changed these names:
- `connect_to_candidates` -> `acquire_all_instances` (vs `acquire_any_candidate`). The parameter is renamed due to the function name no longer talks about candidates at all.
- `get_connection`  -> `select_connection`

src/box/lua/connpool.lua
georgiy-belyanin: I've added the explicit check and the comment explaining the problem. Thank you!

src/box/lua/connpool.lua
georgiy-belyanin: Hmm, you're right. I've added implemented the behaviour. Thank you.

src/box/lua/connpool.lua
Totktonada: Now the code reuses `connection_mode_update_cond`, but we only need the connected state here, AFAIR.

My comment was just about implementing these two condvars in a similar way, not about merging them into one.

src/box/lua/connpool.lua
```diff
@@ -232,12 +237,21 @@ local function filter(opts)
...
     if opts.mode ~= nil and opts.mode ~= 'ro' and opts.mode ~= 'rw' then
         local msg = 'Expected nil, "ro" or "rw", got "%s"'
         error(msg:format(opts.mode), 0)
     end
+
+    if opts.skip_connection_check and opts.mode ~= nil then
+        local msg = 'Filtering by mode "%s" requires the connection ' ..
+                    'check disabled by the "skip_connection_check" option'
+        error(msg:format(opts.mode), 0)
+    end
```
Totktonada, line 297: I guess you meant the opposite: the connection checking functonality is required to perform filtering by the mode, so it can't be disabled in this case.

src/box/lua/connpool.lua
Totktonada: I usually don't mind some amount of minor readability changes without an explicit separation from the main change. However, here we have more lines of renames/refactorings than lines of the main change.

We have a commit (the first one), which states that it is about the connection liveness check, but I need to read over all these renames to find the 9 lines of code that actually do the job.

Also, we can't see from the commit that the renames are for consistency: the second function with this naming pattern is added in another commit now.

I don't care about a name of this function and a name of its argument (the old and the new variants are meaningful for me both). However, please, separate them from the logic change.

I would even split it more:

* The renames.
* `if next(static_candidates) == nil or next(dynamic_opts) == nil then` -> `if next(instance_names) == nil then return {} end`
* The `skip_connection_check` option and the default behavior change.

Totktonada: The `next(dynamic_opts) == nil` check is actually needed to implement the new behavior. So, it is OK to include it into the `skip_connection_check` commit. Initially I interpreted it as a change in the code that doesn't change any user-visible behavior.

Totktonada: BTW, `acquire_{all,any}_instance{s,}` doesn't make an impression that it performs connecting. `connect_to_candidates` does.

`connect_all_candidates` and `connect_any_candidate` would be more intuitive.

Or `connect_all_instances` and `connect_any_instance`.

src/box/lua/connpool.lua
```diff
@@ -348,9 +429,12 @@ local function get_connection(opts)
...
     -- Increase weight of local candidate.
     if opts.prefer_local ~= false then
+        local local_instance_name = box.info.name
+
         local weight_local = 1
-        if weights[box.info.name] ~= nil then
-            weights[box.info.name] = weights[box.info.name] + weight_local
+        if weights[local_instance_name] ~= nil then
+            weights[local_instance_name] = weights[local_instance_name] +
+                                           weight_local
```
Totktonada, line 437: Please, don't complicate understanding of a logic behind your patches by such irrelevant changes. At least move them to its own commits.

This particular name now can be replaced to `config:info('v2').hierarchy.instance` (in a separate PR). So, I wouldn't touch it.

src/box/lua/connpool.lua
```diff
@@ -383,6 +467,60 @@ local function get_connection(opts)
...
+    }
+
+    -- At first we want to acquire the instances matching the
+    -- static (configuration) requirements.
+    local static_candidates = filter(filter_static_opts)
+
+    -- In case we don't need to prioritize the matching instances
+    -- we can wait for any connection available.
+    local need_priorities =
+        opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro'
```
Totktonada, line 509: What was not obvious for me: why we don't perform the dynamic filtering if the instances are to be prioritized. Then I looked over all the dynamic filters and understood that they don't shrink the list of candidates in these `prefer_local`, `prefer_ro`, `prefer_rw` cases.

But I would highlight it explicitly, because it looks like we forgot to perform the dynamic filtering.

src/box/lua/connpool.lua
Totktonada: Don't know what occurs here, but it seems the old name still there.

src/box/lua/connpool.lua
Totktonada: Now this code looks so:

```lua
-- To use the local connection, the connect() call
-- should be ok, if the mode is specified the
-- connection mode should exist (thus, this select
-- may be false-negative) and the candidate should
-- match dynamic options.
if ok and (opts.mode == nil or conn:mode() ~= nil) and
   is_candidate_match_dynamic(local_instance, opts) then
    return conn
end
```

To be honest, I wouldn't understand what you want to say here: `conn:mode()` may be `nil` during some period after `connect` above, so `is_candidate_match_dynamic` may discard the connection.

src/box/lua/connpool.lua
Totktonada: In fact, I don't mind this extra waiting for `mode`, it simplifies the code a bit.

I would just comment that the given `:wait()` call may wait for an extra time to obtain the mode even if it is not used later anyhow (for example, there are no dynamic filters).

src/box/lua/connpool.lua
Totktonada, line 1: I'm worrying that we handle many situations differently: preferring a local instance, priorities and so on. These are different branches and the code looks like the logic is completely different. However, in fact, all the branches are doing something similar and, I guess, can be unufued as steps on the same road.

I've experimented here a bit and created a PoC patch ([branch](https://github.com/Totktonada/tarantool/commits/connpool-make-call-wait-for-any-experiment-1/)) that pass the tests from this patch.

It looks simpler, at least for me. What do you think? Is there any cons?

<details>
<summary>My PoC</summary>

```diff
diff --git a/src/box/lua/connpool.lua b/src/box/lua/connpool.lua
index e81cfe856a..8b634e0327 100644
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -78,6 +78,15 @@ local function connect(instance_name, opts)
     return conn
 end
 
+-- Run connecting to the instance, which if it is not already
+-- connected or in process of connecting.
+local function connect_in_background(instance_name)
+    pcall(connect, instance_name, {
+        wait_connected = false,
+        connect_timeout = WATCHER_TIMEOUT
+    })
+end
+
 local function is_candidate_connected(candidate)
     local conn = connections[candidate]
     return conn and conn.state == 'active' and conn:mode() ~= nil
@@ -96,7 +105,7 @@ end
 
 -- This method connects to all of the specified instances
 -- and returns the set of successfully connected ones.
-local function connect_to_candidates(candidates)
+local function connect_to_candidates(candidates, opts)
     if next(candidates) == nil then return {} end
 
     local delay = WATCHER_DELAY
@@ -117,11 +126,17 @@ local function connect_to_candidates(candidates)
             :filter(is_candidate_connected)
             :totable()
 
-        local all_checked = fun.iter(candidates)
-            :all(is_candidate_checked)
+        -- Stop condition.
+        local stop = false
+        if opts.any then
+            assert(type(opts.any) == 'function')
+            stop = fun.iter(candidates):any(opts.any)
+        else
+            stop = fun.iter(candidates):all(is_candidate_checked)
+        end
 
-        if all_checked then
-            return connected_candidates
+        if stop then
+            break
         end
 
         connection_mode_update_cond:wait(delay)
@@ -228,6 +243,21 @@ end
 
 local function is_candidate_match_dynamic(instance_name, opts)
     assert(opts ~= nil and type(opts) == 'table')
+
+    local conn = connections[instance_name]
+
+    -- No connection -> reconnect in background, exclude the candidate.
+    if conn == nil or conn.state == 'error' or conn.state == 'closed' then
+        connect_in_background(instance_name)
+        return false
+    end
+
+    -- In progress -> exclude the candidate.
+    if not (conn.state == 'active' and conn:mode() ~= nil) then
+        return false
+    end
+
+    -- Connected -> check using the provided filters.
     return is_mode_match(opts.mode, instance_name)
 end
 
@@ -240,7 +270,8 @@ local function filter(opts)
         roles = '?table',
         sharding_roles = '?table',
         mode = '?string',
-        skip_connection_check = '?boolean'
+        skip_connection_check = '?boolean',
+        _wait_mode = '?string',
     })
     opts = opts or {}
 
@@ -300,7 +331,14 @@ local function filter(opts)
     --
     -- The connect_to_candidates() call returns quickly if it
     -- receives empty table as an argument.
-    local connected_candidates = connect_to_candidates(static_candidates)
+    local connected_candidates = static_candidates
+    if opts._wait_mode ~= 'no wait' then
+        connected_candidates = connect_to_candidates(static_candidates, {
+            any = opts._wait_mode == 'any' and function(instance_name)
+                return is_candidate_match_dynamic(instance_name, dynamic_opts)
+            end or nil,
+        })
+    end
 
     -- Filter the remaining candidates after connecting to them.
     local dynamic_candidates = {}
@@ -326,16 +364,36 @@ local function get_connection(opts)
         sharding_roles = opts.sharding_roles,
         mode = mode,
     }
-    local candidates = filter(candidates_opts)
+
+    -- If the mode is not prefer_*, try to find already connected
+    -- candidates.
+    local candidates
+    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
+        candidates = filter(candidates_opts)
+    else
+        -- Try to find already connected candidates.
+        candidates_opts._wait_mode = 'no wait'
+        candidates = filter(candidates_opts)
+
+        -- Try to find at least one connected otherwise.
+        if next(candidates) == nil then
+            candidates_opts._wait_mode = 'any'
+            candidates = filter(candidates_opts)
+        end
+
+        -- Last resort: try to connect to all the candidates.
+        if next(candidates) == nil then
+            candidates_opts._wait_mode = nil
+            candidates = filter(candidates_opts)
+        end
+    end
+
     if next(candidates) == nil then
         return nil, "no candidates are available with these conditions"
     end
 
     -- Initialize the weight of each candidate.
     local weights = {}
-    if opts.mode == 'prefer_rw' or opts.mode == 'prefer_ro' then
-        candidates = connect_to_candidates(candidates)
-    end
     for _, instance_name in pairs(candidates) do
         weights[instance_name] = 0
     end
```
</details>

src/box/lua/connpool.lua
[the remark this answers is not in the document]
Totktonada: TBH, I don't like to handle the `prefer_local` option in many places. There is code that take it into account on sorting the candidates. In my opinion, we don't need any other code to support the option.

Also, I don't like that the local connection is handled somehow specifically. More logic to take into account, at least.

Also, I'm not so sure that this connection is always/likely successful and that it is always fast. There may be no configured iproto port on the given instance. There may be a DNS problem that prevents some hostnames from resolving (and it takes some visible time) -- and the URI of the local instance can use a symbolic hostname. And so on.

I'm not really against, but it seems for me that we can handle this connection exactly like others, except sorting. And since we can, I think, it worth to do.

src/box/lua/connpool.lua
georgiy-belyanin: Fixed. Thank you. I've meant something like `Filtering by mode "%s" requires the connection check [that has been] disabled by the "skip...`. I've added the lacking words from the square brackets.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you. I've applied the PoC almost completely except I've called `filter()` default mode as `wait()` and handled them the same way but expressed in less ternary logic:

<details><summary>See partial diff here</summary>
<p>

```diff
@@ -237,9 +269,25 @@ local function filter(opts)
         sharding_roles = '?table',
         mode = '?string',
         skip_connection_check = '?boolean',
+        -- Internal option on how filter() behaves.
+        -- Used internal in the call() method.
+        --
+        -- The value is represented by enum (the values are sorted
+        -- from the fastest to the slowest):
+        -- * `no wait` -- only return already polled instances.
+        -- * `any` -- wait for any instance satisfying requirements
+        --   to be polled.
+        -- * `wait` (default) -- wait for all instances to be
+        --   polled.
+        _wait_mode = '?string',
     })
     opts = opts or {}

+    local wait_mode = opts._wait_mode
+    if wait_mode == nil then
+        wait_mode = 'wait'
+    end
+
     if opts.mode ~= nil and opts.mode ~= 'ro' and opts.mode ~= 'rw' then
         local msg = 'Expected nil, "ro" or "rw", got "%s"'
         error(msg:format(opts.mode), 0)
@@ -291,11 +339,28 @@ local function filter(opts)
         return static_candidates
     end

-    -- Filter the remaining candidates after connecting to them.
-    --
-    -- The connect_to_candidates() call returns quickly if it
-    -- receives empty table as an argument.
-    local connected_candidates = connect_to_candidates(static_candidates)
+    local connected_candidates
+    if wait_mode == 'wait' then
+        -- Filter the remaining candidates after connecting to them.
+        --
+        -- The connect_to_candidates() call returns quickly if it
+        -- receives empty table as an argument.
+        connected_candidates = connect_to_candidates(static_candidates)
+    elseif wait_mode == 'any' then
+        -- Try to connect to the candidates until at least one
+        -- satisfying the dynamic requirements is found.
+        connected_candidates = connect_to_candidates(static_candidates, {
+            any = function(instance_name)
+                return is_candidate_match_dynamic(instance_name, dynamic_opts)
+            end,
+        })
+    elseif wait_mode == 'no wait' then
+        -- Only continue with already polled instances.
+        connected_candidates = static_candidates
+    else
+        assert(false, 'Unexpected _wait_mode value.')
+    end
+
```

</p>
</details>

georgiy-belyanin:
@Totktonada @grafin 

Please, take a loot at the updated patch. I've applied Alexander PoC (see here https://github.com/tarantool/tarantool/pull/10438#discussion_r1862483309). Now the patch does a very few changes to the original connpool logic and no renames. 

The only problem I've found is that technically it's possible for `call()` with `prefer_local=true` to execute the function not on the local instance.

It may happen when the `connpool.call()` is called for the first time. There is no connection to a local instance and theoretically it's possible for non-local instance to be polled first.

I've came across this problem when I've been testing it locally, though I don't think it's critical since it seems likely not to happen on the real net and it's only about the first `call()` execution and it's about prioritized execution so the code using it must be ready to work on non-local instance.

src/box/lua/connpool.lua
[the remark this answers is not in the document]
georgiy-belyanin: I've dropped these changes after applying your PoC. See here https://github.com/tarantool/tarantool/pull/10438#discussion_r1862483309.

src/box/lua/connpool.lua
georgiy-belyanin: I've dropped such unrelated changes after applying @Totktonada PoC. See here https://github.com/tarantool/tarantool/pull/10438#discussion_r1862483309.

src/box/lua/connpool.lua
georgiy-belyanin: I've dropped such unrelated changes after applying Alexander PoC. See here https://github.com/tarantool/tarantool/pull/10438#discussion_r1862483309.

src/box/lua/connpool.lua
georgiy-belyanin: I think this is no longer related after https://github.com/tarantool/tarantool/pull/10438#discussion_r1862483309.

grafin (approved):
Thanks for working on this!

Totktonada (approved):

changelogs/unreleased/gh-10598-make-connpool-call-faster.md
Totktonada: Seems resolved, so I'm going to close this discussion.

src/box/lua/connpool.lua
Totktonada: Seems resolved, so I'm going to close this discussion.

src/box/lua/connpool.lua
Totktonada: Seems resolved, so I'm going to close this discussion.

src/box/lua/connpool.lua
Totktonada: Seems resolved, so I'm going to close this discussion.

src/box/lua/connpool.lua
Totktonada: Nice, thanks!

PR #10449, opened 2024-08-20
Change cmake variable BACKUP_STORAGE to option

filonenko-mikhail:
There is BACKUP_STORAGE var to use any source to download dependencies. Convert it to option to make it possible to use it from command line.

NO_DOC=build
NO_TEST=build
NO_CHANGELOG=build


ylobankov (approved):

locker (approved):

sergepetrenko (approved):
Thanks for the patch!

Issue #10396, opened 2024-08-09
Memtx's MVCC `count({}, {iterator = 'ALL'})` by index with exclude_null returned -1 without error


d-enk:
**Bug description**

* OS: Linux 6.8.0-39-generic

Tarantool 3.1.1-0-gb65f805
Target: Linux-x86_64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/tmp/tarantool_install946144238/static-build/build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
Compiler: GNU-13.2.0
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2  -fmacro-prefix-map=/tmp/tarantool_install946144238=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -O2 -g -DNDEBUG -ggdb -O2
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2  -fmacro-prefix-map=/tmp/tarantool_install946144238=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -O2 -g -DNDEBUG -ggdb -O2

**Steps to reproduce**

```lua
box.cfg({
  memtx_use_mvcc_engine = true,
  log_level = "error",
})

local space = box.schema.space.create("TEST", {
  engine = "memtx",
  format = {
    { name = "ID", type = "unsigned", is_nullable = false },
    { name = "FLAG", type = "boolean", is_nullable = true },
  },
})
space:create_index("ID", {
  unique = true,
  type = "TREE",
  parts = { { field = "ID", type = "unsigned" } },
})
space:create_index("FLAG", {
  unique = false,
  type = "TREE",
  parts = {
    {
      field = "FLAG",
      type = "boolean",
      exclude_null = true,
      is_nullable = true,
    },
  },
})

require("fiber").new(box.atomic, { txn_isolation = "read-committed" }, function()
  assert(space.index.FLAG:max({ true }))

  space:update({ 1 }, { { "=", "FLAG", box.NULL } })
  space:insert({ 2, box.NULL })
end)

box.atomic(space.insert, space, { 1, true })

-- set box.error.last
[...]

PR #10456, opened 2024-08-21
memtx: skip excluded tuples in index count with MVCC enabled

drewdzzz:
Excluded tuples actually have their own history chains in MVCC - such chains consist of only one `memtx_story` containing excluded tuple itself. Such chains should be skipped when counting invisible tuples because they are not inserted to the index - that's what the commit does.

Closes https://github.com/tarantool/tarantool/issues/10396

### DIFF
--- a/changelogs/unreleased/gh-10396-memtx-mvcc-exclude-null-count.md
+++ b/changelogs/unreleased/gh-10396-memtx-mvcc-exclude-null-count.md
@@ -1,0 +1,5 @@
+## bugfix/memtx
+
+* Fixed a bug when `index:count()` could return the wrong number, raise the
+  last error, or fail with the `IllegalParams` error if the index has
+  the `exclude_null` attribute and MVCC is enabled (gh-10396).
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -2876,6 +2876,17 @@ memtx_tx_index_invisible_count_slow(struct txn *txn,
 		}
 		assert(link->newer_story == NULL);
 
+		/*
+		 * Excluded tuples have their own chains consisting of the only
+		 * excluded story. Such stories must be skipped since they are
+		 * not actually inserted to index.
+		 */
+		if (tuple_key_is_excluded(story->tuple, index->def->key_def,
+					  MULTIKEY_NONE)) {
+			assert(link->older_story == NULL);
+			continue;
+		}
+
 		struct tuple *visible = NULL;
 		bool is_prepared_ok = detect_whether_prepared_ok(txn);
 		bool unused;
--- a/test/box-luatest/gh_9954_mvcc_with_exclude_null_test.lua
+++ b/test/box-luatest/gh_9954_mvcc_with_exclude_null_test.lua
@@ -107,3 +107,53 @@ g.test_mvcc_with_exclude_null_space_drop = function()
         box.space.test:drop()
     end)
 end
+
+-- gh-10396
+g.test_mvcc_with_exclude_null_count = function()
+    g.server:exec(function()
+        local space = box.schema.space.create("TEST", {
+            format = {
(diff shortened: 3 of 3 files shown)


mkostoevr (approved):
LGTM. Nice fix!

test/box-luatest/gh_9954_mvcc_with_exclude_null_test.lua
```diff
@@ -107,3 +107,53 @@ g.test_mvcc_with_exclude_null_space_drop = function()
         box.space.test:drop()
     end)
 end
+
+-- gh-10396
```
CuriousGeorgiy, line 111: Nit: up to you, but we usually create separate files for each issue nowadays.

CuriousGeorgiy (approved):

CuriousGeorgiy:
> Excluded tuples actually have their own history chains in MVCC - such chains consist of only one `memtx_story` containing excluded tuple itself.

@drewdzzz @mkostoevr I would also evaluate whether we can get rid of this redundancy in the future.

src/box/memtx_tx.c
```diff
@@ -2876,6 +2876,17 @@ memtx_tx_index_invisible_count_slow(struct txn *txn,
 		}
 		assert(link->newer_story == NULL);
 
+		/*
+		 * Excluded tuples have their own chains consisting of the only
+		 * excluded story. Such stories must be skipped since they are
+		 * not actually inserted to index.
+		 */
+		if (tuple_key_is_excluded(story->tuple, index->def->key_def,
```
CuriousGeorgiy, line 2884: Ideally, these tuples should be filtered out by the check above (`link->in_index == NULL`), since they are not physically present in the index. See my comment above.

drewdzzz: I agree, but on the other hand we have an invariant that top story of every chain is physically in index, so lots of assertions check that `link->in_index != NULL` at the top story.

As a simplest solution, we could add a flag to `link` whether the story is excluded from the index, so we could replace assertions with `assert(link->in_index != NULL || link->is_excluded)`. But, probably, that's not the best solution and we can come up with something better.

Let's create another issue describing this problem and merge this fix as is. Are you okay with that?

test/box-luatest/gh_9954_mvcc_with_exclude_null_test.lua
drewdzzz: Yep, but this file is dedicated to mvcc with `exclude_null`, so I decided that it belongs here. I would say, it's actually follow up of `9954`.

changelogs/unreleased/gh-10396-memtx-mvcc-exclude-null-count.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/memtx
+
+* Fixed a bug when `index:count()` could return wrong number, raise the
+  last error or fail with `IllegalParams` error if the index has
+  `exclude_null` attribute and MVCC is enabled (gh-10396).
```
andreyaksenov, line 5:
```suggestion
* Fixed a bug when `index:count()` could return the wrong number, raise the
  last error, or fail with the `IllegalParams` error if the index has
  the `exclude_null` attribute and MVCC is enabled (gh-10396).
```

andreyaksenov (approved):

src/box/memtx_tx.c
CuriousGeorgiy: Yeah.

drewdzzz: OK, I'll file an issue.

sergepetrenko:
Backports
3.1 https://github.com/tarantool/tarantool/pull/10467
2.11 https://github.com/tarantool/tarantool/pull/10468

PR #10464, opened 2024-08-22
test: fix flaky gh_9918_synchro_queue_additional_info_test.lua

### LINKED ISSUE #10463: Test replication-luatest/gh_9918_synchro_queue_additional_info_test.lua is flaky
**Bug description**
The test `replication-luatest/gh_9918_synchro_queue_additional_info_test.lua` started flaky failing like this:
```console
[010] not ok 1	replication-luatest.gh_9918_synchro_queue_additional_info.test_age_field
[010] #   Can't modify data on a read-only instance - it is an orphan
[010] #   stack traceback:
[010] #   	...n-luatest/gh_9918_synchro_queue_additional_info_test.lua:27: in function <...n-luatest/gh_9918_synchro_queue_additional_info_test.lua:8>
[010] #   	...
[010] #   	[C]: in function 'xpcall'
[010] #   artifacts:
[010] #   	server2 -> /tmp/t/010_replication-luatest/artifacts/rs-2pZiWi1diM7q/server2-85RRMqavsQ6R
[010] #   	server1 -> /tmp/t/010_replication-luatest/artifacts/rs-2pZiWi1diM7q/server1-mpjZi3MR3PaT
```
```console
[004] not ok 3	replication-luatest.gh_9918_synchro_queue_additional_info.test_confirm_lag_field_with_prepared_async_tx
[004] #   Can't modify data on a read-only instance - it is an orphan
[004] #   stack traceback:
[004] #   	...n-luatest/gh_9918_synchro_queue_additional_info_test.lua:27: in function <...n-luatest/gh_9918_synchro_queue_additional_info_test.lua:8>
[004] #   	...
[004] #   	[C]: in function 'xpcall'
[004] #   artifacts:
[004] #   	server2 -> /tmp/t/004_replication-luatest/artifacts/rs-AnwxvLBSqi7L/server2-j8o9yw5xxQG0
[004] #   	server1 -> /tmp/t/004_replication-luatest/artifacts/rs-AnwxvLBSqi7L/server1-Xxe9zdEpUaKL
```

Astronomax:
This patch eliminates flaking in the `gh_9918_synchro_queue_additional_info_test.lua` test. The problem was that the test did not wait for the connection between master and replica to be established, and therefore master node was in the "orphan" state at the time `box.ctl.promote()` was called. Thus, it turned out that the master node became the owner of the limbo, but was still read only.

Closes #10463

### DIFF
--- a/test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
+++ b/test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
@@ -24,6 +24,7 @@ g.before_each(function(cg)
         box_cfg = box_cfg,
     }
     cg.replica_set:start()
+    cg.replica_set:wait_for_fullmesh()
     cg.leader:exec(function()
         box.ctl.promote()
         local s = box.schema.space.create('s', {is_sync = true})
@@ -31,7 +32,6 @@ g.before_each(function(cg)
         local as = box.schema.space.create('as', {is_sync = false})
         as:create_index('pk')
     end)
-    cg.replica_set:wait_for_fullmesh()
     cg.leader:wait_for_downstream_to(cg.replica)
 end)
 


test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
```diff
@@ -26,6 +26,7 @@ g.before_each(function(cg)
     cg.replica_set:start()
     cg.leader:exec(function()
         box.ctl.promote()
+        box.ctl.wait_rw()
```
CuriousGeorgiy, line 29: > The test expects `box.ctl.promote()` to be synchronous, i.e. that the
caller will wait until his node actually becomes the leader. But this is
not the case, the user must manually call `box.ctl.wait_rw()` to ensure
that he can actually write.

This test does not use elections. The instances are read-write by default, and AFAIC, the instance owns the limbo after `box.ctl.promote` returns:
https://github.com/tarantool/tarantool/blob/850673db5a69df2c7250d174ab15305624b2634a/src/box/box.cc#L3086

So I don't see any reason why the `leader` node should not be writable after `box.ctl.promote`.

It is not clear why the `leader` instance is orphan from the logs you presented in the issue. My guess is that it is because the replica set has not synced. If this is the case, then AFAIC, the right fix would be to move the `cg.replica_set:wait_for_fullmesh()` right after `cg.replica_set:start()`.

CuriousGeorgiy (changes requested):

test/replication-luatest/gh_9918_synchro_queue_additional_info_test.lua
Astronomax: Yes, you are right. Fixed.

CuriousGeorgiy (approved):
Please update the commit message.

sergepetrenko (approved):
Thanks for the fix!

sergepetrenko:
Backport to 3.2: https://github.com/tarantool/tarantool/pull/10642

PR #10469, opened 2024-08-22
vinyl: enable exact match optimization for unique secondary indexes

### LINKED ISSUE #10442: Skip disk lookup if found statement in memory for unique secondary index
**Reproducer**

(reproduced with Tarantool 3.2.0-entrypoint-234-gc3ed71f37ca1)

```Lua

local function csw()
    return fiber.self():info().csw
end

box.cfg{
    vinyl_cache = 0,
    log_level = 'warn',
}

s:create_index('pk')
s:create_index('sk', {parts = {2, 'unsigned'}})

s:insert{1, 10}
box.snapshot()
s:delete{1}

local c

c = csw()
assert(s.index.pk:get(1) == nil)
print('Primary index lookup', csw() - c)

c = csw()
assert(s.index.sk:get(10) == nil)
print('Secondary index lookup', csw() - c)

os.exit(0)
```

**Actual behavior**

```
Primary index lookup    0
Secondary index lookup  1
```

**Expected behavior**

Since the secondary index is unique, there's no need to read the disk if we found a `DELETE` statement in memory.

Actually, we have an optimization for lookup of an exact key but it doesn't work for unique secondary index unless it includes primary key parts. We could probably fix it for this case.

https://github.com/tarantool/tarantool/blob/f6e65fe59b7d21762563d2adef74c5bf6b3889f8/src/box/vy_read_iterator.c#L191-L212

https://github.com/tarantool/tarantool/blob/f6e65fe59b7d21762563d2adef74c5bf6b3889f8/src/box/vy_read_iterator.c#L493-L502

Reported by @palage4a @RunsFor.

locker:
If the iterator type is EQ/REQ/LE/GE and the search key is exact (that is, there may be at most one tuple matching the key in the index), there's no need to scan disk levels if we found a statement for this key in the memory level. We've had this optimization for ages but it worked only for full keys in terms `cmp_def` (key definition extended with primary key parts). Apparently, a lookup in a secondary index performed by the user wouldn't match these criteria unless the secondary index explicitly included all primary key parts.

This commit improves on that. Now, we enable the optimization if the search key is **exact**. We consider a key **exact** if either of the following conditions is true:

 - The key statement is a tuple (tuple has all key parts).
 - The key statement is a full key in terms of `cmp_def`.
 - The key statement is a full key in terms of `key_def`, it doesn't contain nulls, and the index is unique. The check for nulls is necessary because even a unique nullable index may contain more than one equal key with nulls.

Note, this patch slightly refactors the optimization, adding a few comments and hopefully making it more understandable. In particular, we remove the one-result-tuple optimization for exact EQ/REQ from `vy_read_iterator_advance` and put it in `vy_read_iterator_evaluate_src` instead. This way the whole optimization resides in one place.

Closes #10442

### DIFF
--- a/changelogs/unreleased/gh-10442-vy-exact-match-optimization.md
+++ b/changelogs/unreleased/gh-10442-vy-exact-match-optimization.md
@@ -1,0 +1,4 @@
+## bugfix/vinyl
+
+* Eliminated an unnecessary disk read when a key that was recently updated or
+  deleted was accessed via a unique secondary index (gh-10442).
--- a/src/box/vy_read_iterator.c
+++ b/src/box/vy_read_iterator.c
@@ -52,6 +52,11 @@ struct vy_read_src {
 	};
 	/** Set if the iterator was started. */
 	bool is_started;
+	/**
+	 * Set if this is the last (deepest) source that may store tuples
+	 * matching the search criteria.
+	 */
+	bool is_last;
 	/** See vy_read_iterator->front_id. */
 	uint32_t front_id;
 	/** History of the key the iterator is positioned at. */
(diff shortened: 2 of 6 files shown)


mkostoevr (approved):
LGTM. Interesting idea to check for existing nulls even if the key is nullable.

lenkis (approved):

locker:

PR #10472, opened 2024-08-26
[release/3.1] Generate changelog for 3.1.2

sergepetrenko:
Also, remote unreleased/ entries and add a new downgrade version to downgrade versions list.

NO_DOC=changelog
NO_TEST=changelog
NO_CHANGELOG=changelog

@Totktonada, I see some entries which duplicate the ones fixed in 2.11.4 (https://github.com/tarantool/tarantool/releases/tag/2.11.4). I'm not sure if we should remove or keep them in this changelog.

### DIFF
--- a/changelogs/3.1.2.md
+++ b/changelogs/3.1.2.md
@@ -1,0 +1,84 @@
+# 3.1.2
+
+Date: 2024-08-26
+Tag: 3.1.2
+
+## Overview
+
+3.1.2 is the 3rd [stable][release_policy] version of the 3.1 release
+series. It resolves 14 bugs since 3.1.1.
+
+The "stable" label means that we have all planned features implemented and we
+see no high-impact issues. However, if you encounter an issue, feel free to
+[report it][issues] on GitHub.
+
+[release_policy]: https://www.tarantool.io/en/doc/latest/release/policy/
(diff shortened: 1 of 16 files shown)


Totktonada:
> @Totktonada, I see some entries which duplicate the ones fixed in 2.11.4 (https://github.com/tarantool/tarantool/releases/tag/2.11.4). I'm not sure if we should remove or keep them in this changelog.

It seems OK to keep them to highlight a difference from `3.1.1`. Deleting has more sense for a feature release, where we generally tell about new functionality.

As said, I think that the ideal solution would be splitting the changes to some 'new' and 'backported' sections instead of just deleting/keeping the backported changes. We can start something like this from some future release.

Totktonada (approved):
I've no objections.

lenkis (approved):

sergepetrenko:
Added the extra changelog entry from 6376ffbc6b5b3c336b1711a31c1ac082bb9c15ee

sergepetrenko:
> As said, I think that the ideal solution would be splitting the changes to some 'new' and 'backported' sections instead of just deleting/keeping the backported changes. We can start something like this from some future release.

Yep, sounds good to me.

PR #10478, opened 2024-08-26
Make it possible to accelerate the select with offset and implement count in the memtx tree read view

mkostoevr:
Part of tarantool/tarantool-ee#858

### DIFF
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -936,6 +936,31 @@ generic_index_count(struct index *index, enum iterator_type type,
 	return count;
 }
 
+ssize_t
+generic_index_read_view_count(struct index_read_view *rv,
+			      enum iterator_type type, const char *key,
+			      uint32_t part_count)
+{
+	struct index_read_view_iterator it;
+	if (index_read_view_create_iterator(rv, type, key,
+					    part_count, &it) != 0)
+		return -1;
+	int rc = 0;
+	size_t count = 0;
+	struct read_view_tuple tuple;
+	while ((rc = index_read_view_iterator_next_raw(&it, &tuple)) == 0 &&
+	       tuple.data != NULL) {
+		rc = box_check_slice();
+		if (rc != 0)
+			break;
+		++count;
+	}
+	index_read_view_iterator_destroy(&it);
+	if (rc < 0)
+		return rc;
+	return count;
+}
+
 int
 generic_index_get_internal(struct index *index, const char *key,
 			   uint32_t part_count, struct tuple **result)
--- a/src/box/index.cc
+++ b/src/box/index.cc
@@ -1014,6 +1014,33 @@ fail:
 	return NULL;
 }
 
+int
+generic_index_read_view_create_iterator_with_offset(
+	struct index_read_view *rv, enum iterator_type type,
+	const char *key, uint32_t part_count,
+	const char *pos, uint32_t offset,
+	struct index_read_view_iterator *it)
+{
+	if (index_read_view_create_iterator_after(rv, type, key, part_count,
+						  pos, it) != 0)
+		return -1;
+
+	/* Skip the required amount of tuples. */
+	for (size_t i = 0; i < offset; i++) {
+		if (box_check_slice() != 0)
+			goto fail;
+		struct read_view_tuple skipped_tuple;
+		if (index_read_view_iterator_next_raw(it, &skipped_tuple) != 0)
+			goto fail;
+		if (skipped_tuple.data == NULL)
+			return 0;
+	}
+	return 0;
+fail:
+	index_read_view_iterator_destroy(it);
+	return -1;
+}
+
 struct index_read_view *
 generic_index_create_read_view(struct index *index)
 {
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -2300,6 +2300,14 @@ tree_read_view_free(struct index_read_view *base)
 # include "memtx_tree_read_view.cc"
 #else /* !defined(ENABLE_READ_VIEW) */
 
+template<bool USE_HINT>
+static ssize_t
+tree_read_view_count(struct index_read_view *rv, enum iterator_type type,
+		     const char *key, uint32_t part_count)
+{
+	return generic_index_read_view_count(rv, type, key, part_count);
+}
+
 template <bool USE_HINT>
 static int
 tree_read_view_get_raw(struct index_read_view *rv,
@@ -2458,6 +2466,7 @@ memtx_tree_index_create_read_view(struct index *base)
 {
 	static const struct index_read_view_vtab vtab = {
 		.free = tree_read_view_free<USE_HINT>,
+		.count = tree_read_view_count<USE_HINT>,
 		.get_raw = tree_read_view_get_raw<USE_HINT>,
 		.create_iterator = tree_read_view_create_iterator<USE_HINT>,
 		.create_iterator_with_offset =
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -2351,16 +2351,18 @@ static int
 tree_read_view_iterator_start(struct tree_read_view_iterator<USE_HINT> *it,
 			      enum iterator_type type,
 			      const char *key, uint32_t part_count,
-			      const char *pos)
+			      const char *pos, uint32_t offset)
 {
 	assert(type == ITER_ALL);
 	assert(key == NULL);
 	assert(part_count == 0);
 	assert(pos == NULL);
+	assert(offset == 0);
 	(void)type;
 	(void)key;
 	(void)part_count;
 	(void)pos;
+	(void)offset;
 	struct tree_read_view<USE_HINT> *rv =
 		(struct tree_read_view<USE_HINT> *)it->base.index;
 	it->base.next_raw = tree_read_view_iterator_next_raw<USE_HINT>;
@@ -2405,14 +2407,13 @@ tree_read_view_iterator_position_func(struct index_read_view_iterator *it,
 						pos, size);
 }
 
-/** Implementation of create_iterator index_read_view callback. */
+/** Implementation of create_iterator_with_offset index_read_view callback. */
 template <bool USE_HINT>
 static int
-tree_read_view_create_iterator(struct index_read_view *base,
-			       enum iterator_type type,
-			       const char *key, uint32_t part_count,
-			       const char *pos,
-			       struct index_read_view_iterator *iterator)
+tree_read_view_create_iterator_with_offset(
+	struct index_read_view *base, enum iterator_type type, const char *key,
+	uint32_t part_count, const char *pos, uint32_t offset,
+	struct index_read_view_iterator *iterator)
 {
 	struct tree_read_view_iterator<USE_HINT> *it =
 		(struct tree_read_view_iterator<USE_HINT> *)iterator;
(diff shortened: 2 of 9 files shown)


drewdzzz (changes requested):

src/box/index.cc
```diff
@@ -1010,6 +1014,56 @@ generic_index_create_iterator_with_offset(struct index *base,
...
+{
+	if (index_read_view_create_iterator_after(rv, type, key, part_count,
+						  pos, it) != 0)
+		return -1;
+
+	/* Skip the required amount of tuples. */
+	for (size_t i = 0; i < offset; i++) {
+		/* The fiber slice is checked in the next function. */
+		struct tuple *skipped_tuple;
+		if (box_index_read_view_iterator_next(it, &skipped_tuple) != 0)
```
locker, line 1073: Since you skip tuples anyway, you can use `next_raw` here. Then you wouldn't need to include `index_read_view.h`, nor define the function under `ENABLE_READ_VIEW`.

src/box/index.cc
```diff
@@ -936,6 +936,45 @@ generic_index_count(struct index *index, enum iterator_type type,
...
+			      uint32_t part_count)
+{
+	struct index_read_view_iterator it;
+	if (index_read_view_create_iterator(rv, type, key,
+					    part_count, &it) != 0)
+		return -1;
+	int rc = 0;
+	size_t count = 0;
+	struct tuple *tuple = NULL;
+	while ((rc = box_index_read_view_iterator_next(&it, &tuple)) == 0 &&
```
locker, line 953: Same comment as to `generic_index_read_view_create_iterator_with_offset` - no need to use `box_index_read_view_iterator_next` here, `next_raw` should be enough.

Also, I'd merge this commit with the one introducing `create_iterator_with_offset`.

src/box/memtx_tx.c
```diff
@@ -3659,6 +3663,55 @@ memtx_tx_snapshot_clarify_slow(struct memtx_tx_snapshot_cleaner *cleaner,
...
+
+		/* In-index tuple, but does not match with key and border. */
+		if (!memtx_tx_tuple_matches_until(key_def, entry->from, type,
+						  key, part_count, cmp_def,
+						  until, until_hint))
+			continue;
+
+		/* Count the invisible tuple. */
+		if (entry->to == NULL ||
+		    memtx_tx_snapshot_clarify_slow(cleaner, entry->to) == NULL)
```
locker, line 3709: I fail to understand why you both iterate over *all* hash table entries and do lookups along the way.

src/box/memtx_tx.h
```diff
@@ -539,6 +539,59 @@ memtx_tx_snapshot_clarify(struct memtx_tx_snapshot_cleaner *cleaner,
 	return memtx_tx_snapshot_clarify_slow(cleaner, tuple);
 }
 
+/** Helper of memtx_tx_snapshot_invisible_count_matching. */
+size_t
+memtx_tx_snapshot_invisible_count_matching_until_slow(
```
locker, line 544: It's a pity that we add dead code. Can we cover it with tests in CE? If not, can we move it to EE? Maybe just put this code under `ifdef ENABLE_READ_VIEW`?

src/box/index.cc
```diff
@@ -43,6 +43,10 @@
 #include "base64.h"
 #include "scoped_guard.h"
 
+#if defined(ENABLE_READ_VIEW)
+# include "index_read_view.h"
```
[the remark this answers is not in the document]
mkostoevr: The change had been dropped from the PR.

src/box/memtx_tree.cc
```diff
@@ -2299,6 +2299,19 @@ tree_read_view_free(struct index_read_view *base)
...
+template<bool USE_HINT>
+static ssize_t
+tree_read_view_count(struct index_read_view *rv, enum iterator_type type,
+		     const char *key, uint32_t part_count)
+{
+	(void)rv;
+	(void)type;
+	(void)key;
+	(void)part_count;
+	unreachable();
```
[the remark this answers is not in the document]
mkostoevr: Made it call the generic impelemtation.

src/box/memtx_tx.h
```diff
@@ -539,6 +539,59 @@ memtx_tx_snapshot_clarify(struct memtx_tx_snapshot_cleaner *cleaner,
...
+ * Detect whether the key of @a tuple from an index with the given @a key_def
+ * is visible in the snapshot with the given @a cleaner.
+ */
+static inline bool
+memtx_tx_snapshot_tuple_key_is_visible(
+	struct memtx_tx_snapshot_cleaner *cleaner,
+	struct key_def *key_def, struct tuple *tuple)
+{
+	return memtx_tx_snapshot_clarify(cleaner, tuple) != NULL &&
+	       !tuple_key_is_excluded(tuple, key_def, MULTIKEY_NONE);
```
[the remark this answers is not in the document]
mkostoevr: Cleaner is space-specific, not index-specific, and the PK of the space contains all tuples, including the ones possibly excluded from secondary keys.

src/box/memtx_tree.cc
```diff
@@ -2417,14 +2419,13 @@ tree_read_view_iterator_position_func(struct index_read_view_iterator *it,
...
-/** Implementation of create_iterator index_read_view callback. */
+/** Implementation of create_iterator_with_offset index_read_view callback. */
 template <bool USE_HINT>
 static int
-tree_read_view_create_iterator(struct index_read_view *base,
-			       enum iterator_type type,
-			       const char *key, uint32_t part_count,
-			       const char *pos,
-			       struct index_read_view_iterator *iterator)
+tree_read_view_create_iterator_with_offset(
```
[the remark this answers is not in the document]
mkostoevr: Made it raise an error if the CE read view iterator is created with the offset given.

src/box/memtx_tree.cc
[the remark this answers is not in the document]
mkostoevr: Good idea, merged. Thanks.

src/box/index.cc
mkostoevr: Right, used the raw iterator step here and in the `count()` generic implementation. Thanks.

src/box/index.cc
mkostoevr: Done.

Regarding merging the commits. If you have no objections, I keep these two: it seems to me like they should be separated since they introduce unrelated read view methods.

src/box/memtx_tx.c
mkostoevr: That's done cause the tuple clarification routine clarifies tuples in a loop. IDK what was the original point in this loop, but after introducing alter statements it started to make sense cause the altered tuple might not point directly to the cleared tuple, so we may need one more iteration. We don't care about alter statements here, so I removed the call. Also only handled the alter stamtements in no-MVCC mode, since it's not required in this case, see the "memtx: only store in-index tuples in the cleaner" commit.

Thanks.

src/box/memtx_tx.h
mkostoevr: Wrapped declarations in the header with the `ENABLE_READ_VIEW` guard.

src/box/memtx_tx.c
```diff
@@ -3628,13 +3630,15 @@ memtx_tx_snapshot_cleaner_create(struct memtx_tx_snapshot_cleaner *cleaner,
 		struct memtx_tx_snapshot_cleaner_entry entry;
 		entry.from = tuple;
 		entry.to = clean;
+		entry.in_pk = story->link[0].in_index != NULL;
```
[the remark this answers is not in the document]
mkostoevr: We don't actually, since the cleaner is only used for lookup of tuples received from the index. Made the hash table only cotain entries for in-indx tuples. See the [comment below](https://github.com/tarantool/tarantool/pull/10478#discussion_r1774917489).

src/box/memtx_tx.c
```diff
@@ -3628,13 +3630,15 @@ memtx_tx_snapshot_cleaner_create(struct memtx_tx_snapshot_cleaner *cleaner,
...
 		mh_snapshot_cleaner_put(ht,  &entry, NULL, 0);
 	}
 	struct space_alter_stmt *alter_stmt;
 	rlist_foreach_entry(alter_stmt, &space->alter_stmts, link) {
 		struct memtx_tx_snapshot_cleaner_entry entry;
 		entry.from = alter_stmt->new_tuple;
 		entry.to = alter_stmt->old_tuple;
+		entry.in_pk = false;
```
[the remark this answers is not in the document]
mkostoevr: This was a mistake, they actually are accessible, thanks. Also, the change is removed from the PR, see the [comment below](https://github.com/tarantool/tarantool/pull/10478#discussion_r1774917489).

> Why do we need them at all, shouldn't be memtx_stories of system spaces enough?

We don't have transaction isolation without MVCC, and all the changes in system spaces are visible immediately. Even if the change is a long-lasting yielding DDL operation. So these alter starements are required to exclude such contents of system spaces from the snapshot until the DDL has successfully completed.

src/box/memtx_tx.h
```diff
@@ -539,6 +539,63 @@ memtx_tx_snapshot_clarify(struct memtx_tx_snapshot_cleaner *cleaner,
...
+}
+
+#endif
```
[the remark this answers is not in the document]
mkostoevr: Fixed the comment and Included the `memtx_tx_read_view.c` into `memtx_tx.c` under `#ifdef` and created the file in the EE. This way it can use the `memtx_tx_tuple_matches[_uintil]` utility functions from the `memtx_tx.c` and be only defined in EE.

I'd also include a new `memtx_tx_read_view.h` into `memtx_tx.h` instead of having the declaration and implementation scattered over different repositories. Let me know what you think about it, I can't see anything similar in CE now.

src/box/memtx_tree.cc
```diff
@@ -2362,6 +2362,10 @@ tree_read_view_iterator_start(struct tree_read_view_iterator<USE_HINT> *it,
 	(void)pos;
 	struct tree_read_view<USE_HINT> *rv =
 		(struct tree_read_view<USE_HINT> *)it->base.index;
+	if (offset != 0) {
+		diag_set(UnsupportedIndexFeature, rv->base.def,
+			 "read view select with offset");
+	}
```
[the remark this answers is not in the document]

src/box/memtx_tx.h
```diff
@@ -539,6 +539,63 @@ memtx_tx_snapshot_clarify(struct memtx_tx_snapshot_cleaner *cleaner,
 	return memtx_tx_snapshot_clarify_slow(cleaner, tuple);
 }
 
+#if defined(ENABLE_READ_VIEW)
```
locker, line 542: Let's move the declarations to the EE repo as well and include `memtx_tx_read_view.h` here under `ifdef`.

Also, let's put the `include` directives after all declarations (in the end of the file).

mkostoevr: Done.

locker (approved):

src/box/memtx_tree.cc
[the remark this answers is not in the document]
mkostoevr: Seems this is a dead code, asserted `offset == 0` as it's done for other parameters.

The generic implementation performs a regular iterator skip instead of throwing anerror BTW.

src/box/memtx_tx.c
```diff
@@ -3625,17 +3625,43 @@ memtx_tx_snapshot_cleaner_create(struct memtx_tx_snapshot_cleaner *cleaner,
 		if (clean == tuple)
```
drewdzzz, line 3625: Unrelated to your patch, but while we are here. Is it ok that we clarify tuples to `prepared` ones, not `committed`?

src/box/memtx_tx.c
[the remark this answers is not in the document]
drewdzzz: Awesome comment you dropped in the code here! Helps to figure it out literally from scratch 😄

drewdzzz (commented):
Everything looks nice, just the last question: why do we clarify tuple with `is_prepared_ok` attribute in snapshot cleaner? Shouldn't we clean tuples to committed ones?

src/box/memtx_tx.c
mkostoevr: Can't say for sure, it's been this way ever since the MVCC snapshots are introduced.

drewdzzz: It's OK actually - we firstly freeze all prepared tuples, then we do `wal_sync` and `txn_limbo_confirm` - in other words, wait while all prepared tuples will become committed. So we must clarify prepared tuples here.

drewdzzz (approved):

PR #10488, opened 2024-08-28
Forbid non-integers in datetime.new, :set() and parse()

### LINKED ISSUE #10391: datetime strange behavior with non-integer fields
Reproducer:
```Lua
tarantool> require('datetime').new({min = 1.111, sec = 1.111, nsec = 111111111}).timestamp
---
- 67.882111111
...
```

Since I set `nsec`, I expect the value after the point to be `111111111`.

It seems like the best way is to only allow integer values ​​for all these fields to avoid this weird behavior. Not sure if this is possible though.

ligurio:
Fixes #10391

### DIFF
--- a/changelogs/unreleased/gh-10391-forbid-non-integers-datetime-new.md
+++ b/changelogs/unreleased/gh-10391-forbid-non-integers-datetime-new.md
@@ -1,0 +1,3 @@
+## bugfix/datetime
+
+- Forbid non-integers in `datetime.new()` (gh-10391).
--- a/changelogs/unreleased/gh-10391-forbid-non-integers-datetime-set.md
+++ b/changelogs/unreleased/gh-10391-forbid-non-integers-datetime-set.md
@@ -1,0 +1,3 @@
+## bugfix/datetime
+
+- Forbid non-integers in `:set()` and `datetime.parse()` (gh-10391).
--- a/changelogs/unreleased/gh-10391-forbid-non-integers-sql-datetime.md
+++ b/changelogs/unreleased/gh-10391-forbid-non-integers-sql-datetime.md
@@ -1,0 +1,3 @@
+## bugfix/datetime
+
+- Forbid non-integers in SQL's `CAST({}) AS datetime` (gh-10391).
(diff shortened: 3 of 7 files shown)


p7nov (approved):

ImeevMA (commented):
Hi! Thank you for the PR! I have no problems with the code itself, but I have some questions about commit-message.

1. I am not sure that we should use "Fixes #..." here since the problem is still present in SQL:
```Lua
tarantool> box.execute([[select cast({'year' : 1.1, 'month' : 1.1, 'week' : 1.1, 'day' : 1.1, 'hour' : 1.1, 'min' : 1.1, 'sec' : 1.1} as datetime);]]) 
---
- metadata:
  - name: COLUMN_1
    type: datetime
  rows:
  - ['0001-01-01T01:07:08Z']
...
```

I suggest to change in both patches to "Part of #...". You can leave the SQL part to me, I will fix it a bit later.

2. The second patch says that it fixes problems with `datetime.parse()`. So is this an expected behavior?
```Lua
tarantool> dt.parse('0001-01-01T01:01:02Z')
---
- 0001-01-01T01:01:02Z
- 20
...

tarantool> dt.parse('0001-01-01T01:01:02.5Z')
---
- 0001-01-01T01:01:02.500Z
- 22
...

tarantool> dt.parse('0001-01-01T01:01.5:02.5Z')
---
- 0001-01-01T01:01:00Z
- 16
...

tarantool> dt.parse('0001-01-01T01.5:01.5:02.5Z')
---
- 0001-01-01T01:00:00Z
- 13
...
```

ligurio:
>  I am not sure that we should use "Fixes #..." here since the problem is still present in SQL:

The issue was about non-integers in datetime Lua API, I propose to submit a separate issue about non-integers in SQL.
Is it ok for you?

> The second patch says that it fixes problems with datetime.parse().

The proposed patch fixes non-integers values in custom `tzoffset`.

> So is this an expected behavior?
 
The first and the second examples parsed correctly, third and fourth examples were parsed partially due to incorrect values.

ImeevMA (approved):
> The issue was about non-integers in datetime Lua API, I propose to submit a separate issue about non-integers in SQL.
> Is it ok for you?

Yes, it's ok for me.

LGTM!

ImeevMA (approved):

Issue #10274, opened 2024-07-18
info: big tuples are not displayed in box.slab.info()


DifferentialOrange:
**Bug description**

When working with big tuples, `box.slab.info()` info declares that there a lot of free space even though it isn't. (`box.slab.stats()` seems to do the same.)

* OS: Ubuntu, EPEL
* OS Version:  Ubuntu 22.04, EPEL7
* Architecture: amd64

Tarantool 2.11.3, 3.1.0.

**Steps to reproduce**

```lua
Tarantool 3.1.0-0-g96f6d88
type 'help' for interactive help
tarantool> box.cfg{memtx_max_tuple_size = 32 * 1024 * 1024, memtx_memory = 128 * 1024 * 1024}
2024-07-18 17:52:03.252 [1608429] main/104/interactive I> Tarantool 3.1.0-0-g96f6d88 Linux-x86_64-RelWithDebInfo
2024-07-18 17:52:03.252 [1608429] main/104/interactive I> log level 5 (INFO)
2024-07-18 17:52:03.252 [1608429] main/104/interactive I> wal/engine cleanup is paused
2024-07-18 17:52:03.253 [1608429] main/104/interactive I> mapping 134217728 bytes for memtx tuple arena...
2024-07-18 17:52:03.253 [1608429] main/104/interactive I> Actual slab_alloc_factor calculated on the basis of desired slab_alloc_factor = 1.044274
2024-07-18 17:52:03.253 [1608429] main/104/interactive I> mapping 134217728 bytes for vinyl tuple arena...
2024-07-18 17:52:03.254 [1608429] main/104/interactive/box.upgrade I> Recovering snapshot with schema version 3.1.0
2024-07-18 17:52:03.256 [1608429] main/104/interactive I> update replication_synchro_quorum = 1
2024-07-18 17:52:03.256 [1608429] main/104/interactive I> instance uuid 733cb0b9-05bd-45fe-8ceb-641c0e8e7e93
2024-07-18 17:52:03.256 [1608429] main/104/interactive I> instance vclock {}
2024-07-18 17:52:03.257 [1608429] main/104/interactive I> recovery start
2024-07-18 17:52:03.257 [1608429] main/104/interactive I> recovering from `./00000000000000000000.snap'
2024-07-18 17:52:03.257 [1608429] main/104/interactive I> replicaset uuid 562cfc81-c985-48f3-bc7d-241c3faccf48
[...]

PR #10503, opened 2024-09-02
Update c API

SEChudinov:
Add struct slab_info and fuction box_slab_info for use in lbox_slab_info and export in c API

### DIFF
--- a/src/box/box.h
+++ b/src/box/box.h
@@ -788,6 +788,22 @@ box_iproto_override(uint32_t req_type, iproto_handler_t handler,
 API_EXPORT int64_t
 box_info_lsn(void);
 
+/*! Codes for request memtx status information for box.slab.info. */
+enum box_slab_info_type {
+	BOX_SLAB_INFO_ITEMS_SIZE = 0, /*!< Allocated only for tuples. */
+	BOX_SLAB_INFO_ITEMS_USED = 1, /*!< Used only for tuples. */
+	BOX_SLAB_INFO_ARENA_SIZE = 2, /*!< Allocated for  tuples and indexes. */
+	BOX_SLAB_INFO_ARENA_USED = 3, /*!< Used for both tuples and indexes. */
+	BOX_SLAB_INFO_QUOTA_SIZE = 4, /*!< Memory limit for slab allocator. */
+	BOX_SLAB_INFO_QUOTA_USED = 5, /*!< Used by slab allocator. */
+};
+
+/**
+ * Get memtx status information for box.slab.info.
+ */
+API_EXPORT uint64_t
+box_slab_info(enum box_slab_info_type type);
+
 /** \endcond public */
 
 /**
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -52,6 +52,7 @@
 #include "small/small.h"
 #include "small/quota.h"
 #include "memory.h"
+#include "box/box.h"
 #include "box/engine.h"
 #include "box/memtx_engine.h"
 #include "box/allocator.h"
@@ -134,38 +135,28 @@ lbox_slab_stats(struct lua_State *L)
 static int
 lbox_slab_info(struct lua_State *L)
 {
-	struct memtx_engine *memtx;
-	memtx = (struct memtx_engine *)engine_by_name("memtx");
+	uint64_t items_size = box_slab_info(BOX_SLAB_INFO_ITEMS_SIZE);
+	uint64_t items_used = box_slab_info(BOX_SLAB_INFO_ITEMS_USED);
+	uint64_t arena_size = box_slab_info(BOX_SLAB_INFO_ARENA_SIZE);
+	uint64_t arena_used = box_slab_info(BOX_SLAB_INFO_ARENA_USED);
+	uint64_t quota_size = box_slab_info(BOX_SLAB_INFO_QUOTA_SIZE);
+	uint64_t quota_used = box_slab_info(BOX_SLAB_INFO_QUOTA_USED);
 
-	struct allocator_stats stats;
-	memset(&stats, 0, sizeof(stats));
-
-	/*
-	 * List all slabs used for tuples and slabs used for
-	 * indexes, with their stats.
-	 */
 	lua_newtable(L);
-	allocators_stats(&stats);
-	struct mempool_stats index_stats;
-	mempool_stats(&memtx->index_extent_pool, &index_stats);
-
-	double ratio;
 	char ratio_buf[32];
-
-	ratio = 100 * ((double) (stats.small.used + stats.sys.used)
-		/ ((double) (stats.small.total + stats.sys.total) + 0.0001));
+	double ratio = 100 * ((double)items_used / (items_size + 0.0001));
 	snprintf(ratio_buf, sizeof(ratio_buf), "%0.2lf%%", ratio);
 
 	/** How much address space has been already touched */
 	lua_pushstring(L, "items_size");
-	luaL_pushuint64(L, stats.small.total + stats.sys.total);
+	luaL_pushuint64(L, items_size);
 	lua_settable(L, -3);
 	/**
 	 * How much of this formatted address space is used for
 	 * actual data.
 	 */
 	lua_pushstring(L, "items_used");
-	luaL_pushuint64(L, stats.small.used + stats.sys.used);
+	luaL_pushuint64(L, items_used);
 	lua_settable(L, -3);
 
 	/*
@@ -180,16 +171,6 @@ lbox_slab_info(struct lua_State *L)
 	/** How much address space has been already touched
 	 * (tuples and indexes) */
 	lua_pushstring(L, "arena_size");
-	/*
-	 * We could use totals.total + index_stats.total here,
-	 * but this would not account for slabs which are sitting
-	 * in slab cache or in the arena, available for reuse.
-	 * Make sure a simple formula:
-	 * items_used_ratio > 0.9 && arena_used_ratio > 0.9 &&
-	 * quota_used_ratio > 0.9 work as an indicator
-	 * for reaching Tarantool memory limit.
-	 */
-	size_t arena_size = memtx->arena.used;
 	luaL_pushuint64(L, arena_size);
 	lua_settable(L, -3);
 	/**
@@ -197,12 +178,10 @@ lbox_slab_info(struct lua_State *L)
 	 * data (tuples and indexes).
 	 */
 	lua_pushstring(L, "arena_used");
-	/** System allocator does not use arena. */
-	luaL_pushuint64(L, stats.small.used + index_stats.totals.used);
+	luaL_pushuint64(L, arena_used);
 	lua_settable(L, -3);
 
-	ratio = 100 * ((double) (stats.small.used + index_stats.totals.used)
-		       / (double)(arena_size + 1));
+	ratio = 100 * ((double)arena_used / (arena_size + 1));
 	snprintf(ratio_buf, sizeof(ratio_buf), "%0.1lf%%", ratio);
 
 	lua_pushstring(L, "arena_used_ratio");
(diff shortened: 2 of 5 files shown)


src/box/box.h
```diff
@@ -788,6 +788,22 @@ box_iproto_override(uint32_t req_type, iproto_handler_t handler,
 API_EXPORT int64_t
 box_info_lsn(void);
 
+typedef struct slab_info
+{
+	uint64_t items_size;
+	uint64_t items_used;
+	double items_used_ratio;
```
Gumix, line 795: IMHO there is no sense in providing `*_ratio` values, anyone can calculate them on the user side.

src/box/box.h
```diff
@@ -788,6 +788,22 @@ box_iproto_override(uint32_t req_type, iproto_handler_t handler,
 API_EXPORT int64_t
 box_info_lsn(void);
 
+typedef struct slab_info
```
Gumix, line 791: Some kind of extensibility needs to be considered here.
Probably the `box.slab.info()` output will be changed/extended in scope of:
- https://github.com/tarantool/tarantool/issues/10274#issuecomment-2242827465

So the C API should be able to support the new output without breaking the binary compatibility.

locker: @Totktonada WDYT about the suggested C API? Maybe, we should make the new structure extendable (reserve some padding for future extensions or allocate it on the region)? Another alternative is to introduce a function that would return the value of a particular memory usage counter, for example:

```C
enum box_slab_info_type {
    BOX_SLAB_INFO_ITEMS = 0,
    BOX_SLAB_INFO_INDEX = 1,
    /* ... */
};

uint64_t
box_slab_info(enum box_slab_info_type type);
```

Totktonada: We definitely need this API be extendable. Padding is a valid option, but I think it is not the best option for informational API -- we may want to add many values to it.

The function that returns a specific counter looks better. However, I'm not a big fan of `enum`'s in the public API for such purposes. It obligates a user to use the new header version to have ability to use new metrics. Or define this enum on its side.

I think that string IDs fit better here. `strcmp()` on a few-bytes-long string is not a heavy thing. But it allows to compile against an old header version and get new metrics if actual tarantool version in runtime is the new one.

If this API is designing for logging/metrics reporting (or other way to give them to a human), it may be useful to get all the metrics disregarding whether they're known on the user side. We can return an array of key-value pairs `{.name = "items_size", .value = 42}`. This way getting many values at once doesn't require many calls. We can add two functions: for integer values and for double ones. Or return a type in the array:

```c
struct box_slab_info_item {
	const char *name;
	const char *type;
	union {
		uint64_t u64;
		double d;
	};
};

/*
 * Returns an array of metrics:
 *
 * [
 *     {"items_size", "uint64_t", 42},
 *     {"items_used", "uint64_t", 42},
 *     {"items_ratio", "double", 0.5},
 *     <...>
 * ]
 *
 * The array size is written to @a size.
 *
 * Allocated on box region, see box_region_used() and
 * box_region_truncate().
 */
struct box_slab_info_item *
box_slab_info(size_t *size);
```

src/box/box.h
Totktonada: It also may simplify the API if we're going to follow my proposal below to return the values in an array of key-value pairs. If we have only integral values, we don't need to return a type of the value.

p7nov (approved):

src/box/box.h
locker: > I think that string IDs fit better here. `strcmp()` on a few-bytes-long string is not a heavy thing.

The problem with this approach is that fetching a particular counter value is inefficient and inconvenient because it requires scanning of all returned items. AFAIU they're planning to use this function to get the actual memory usage on every batch insertion.

0x501D: The option with enum suits us better. We need the fastest possible API.

src/box/lua/slab.cc
```diff
@@ -134,38 +135,35 @@ lbox_slab_stats(struct lua_State *L)
...
 	 */
 	lua_newtable(L);
-	allocators_stats(&stats);
-	struct mempool_stats index_stats;
-	mempool_stats(&memtx->index_extent_pool, &index_stats);
-
-	double ratio;
 	char ratio_buf[32];
+	double ratio = 100 * ((double)(items_used)
+			/ ((double)(items_size) + 0.0001));
```
locker, line 152: Please don't split lines if they fit in 80 characters. Please remove extra parentheses and conversions.

```diff
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index 2b62813be162..7a197fb420ce 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -148,11 +148,8 @@ lbox_slab_info(struct lua_State *L)
 	 */
 	lua_newtable(L);
 	char ratio_buf[32];
-	double ratio = 100 * ((double)(items_used)
-			/ ((double)(items_size) + 0.0001));
-
-	snprintf(ratio_buf, sizeof(ratio_buf),
-		 "%0.2lf%%", ratio);
+	double ratio = 100 * ((double)items_used / (items_size + 0.0001));
+	snprintf(ratio_buf, sizeof(ratio_buf), "%0.2lf%%", ratio);
 
 	/** How much address space has been already touched */
 	lua_pushstring(L, "items_size");
@@ -199,11 +196,8 @@ lbox_slab_info(struct lua_State *L)
 	luaL_pushuint64(L, arena_used);
 	lua_settable(L, -3);
 
-	ratio = 100 * ((double)(arena_used) /
-		(double)(arena_size + 1));
-
-	snprintf(ratio_buf, sizeof(ratio_buf),
-		 "%0.1lf%%", ratio);
+	ratio = 100 * ((double)arena_used / (arena_size + 1));
+	snprintf(ratio_buf, sizeof(ratio_buf), "%0.1lf%%", ratio);
 
 	lua_pushstring(L, "arena_used_ratio");
 	lua_pushstring(L, ratio_buf);
@@ -231,11 +225,8 @@ lbox_slab_info(struct lua_State *L)
 	 * factor, it's the quota that give you OOM error in the
 	 * end of the day.
 	 */
-	ratio = 100 * ((double)(quota_used) /
-		(double)(quota_size + 0.0001));
-
-	snprintf(ratio_buf, sizeof(ratio_buf),
-		 "%0.2lf%%", ratio);
+	ratio = 100 * ((double)quota_used / (quota_size + 0.0001));
+	snprintf(ratio_buf, sizeof(ratio_buf), "%0.2lf%%", ratio);
 
 	lua_pushstring(L, "quota_used_ratio");
 	lua_pushstring(L, ratio_buf);
```

src/box/lua/slab.cc
```diff
@@ -134,38 +135,35 @@ lbox_slab_stats(struct lua_State *L)
...
+	uint64_t items_used = box_slab_info(BOX_SLAB_INFO_ITEMS_USED);
+	uint64_t arena_size = box_slab_info(BOX_SLAB_INFO_ARENA_SIZE);
+	uint64_t arena_used = box_slab_info(BOX_SLAB_INFO_ARENA_USED);
+	uint64_t quota_size = box_slab_info(BOX_SLAB_INFO_QUOTA_SIZE);
+	uint64_t quota_used = box_slab_info(BOX_SLAB_INFO_QUOTA_USED);
 
 	/*
 	 * List all slabs used for tuples and slabs used for
 	 * indexes, with their stats.
 	 */
```
locker, line 148: Apparently this comment was copied from `lbox_slab_stats` long time ago. It doesn't make any sense here. Let's remove it while we are here.

src/box/lua/slab.cc
```diff
@@ -189,7 +187,7 @@ lbox_slab_info(struct lua_State *L)
 	 * quota_used_ratio > 0.9 work as an indicator
 	 * for reaching Tarantool memory limit.
 	 */
-	size_t arena_size = memtx->arena.used;
+
```
locker, line 190: The comment above doesn't make any sense now. Let's move it to `box_slab_info()`.

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 8fa8272cfca0..44b283cc1644 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4254,8 +4254,17 @@ box_slab_info(enum box_slab_info_type type)
 	case BOX_SLAB_INFO_ITEMS_USED:
 		return stats.small.used + stats.sys.used;
 	case BOX_SLAB_INFO_ARENA_SIZE:
+		/*
+		 * We could use stats.small.used + index_stats.total.used
+		 * here, but this would not account for slabs which are
+		 * sitting in slab cache or in the arena, available for reuse.
+		 * Make sure a simple formula: items_used_ratio > 0.9 &&
+		 * arena_used_ratio > 0.9 && quota_used_ratio > 0.9 work as
+		 * an indicator for reaching Tarantool memory limit.
+		 */
 		return memtx->arena.used;
 	case BOX_SLAB_INFO_ARENA_USED:
+		/** System allocator does not use arena. */
 		return stats.small.used + index_stats.totals.used;
 	case BOX_SLAB_INFO_QUOTA_SIZE:
 		return quota_total(&memtx->quota);
diff --git a/src/box/lua/slab.cc b/src/box/lua/slab.cc
index 2b62813be162..7b38eb41d58a 100644
--- a/src/box/lua/slab.cc
+++ b/src/box/lua/slab.cc
@@ -178,16 +178,6 @@ lbox_slab_info(struct lua_State *L)
 	/** How much address space has been already touched
 	 * (tuples and indexes) */
 	lua_pushstring(L, "arena_size");
-	/*
-	 * We could use totals.total + index_stats.total here,
-	 * but this would not account for slabs which are sitting
-	 * in slab cache or in the arena, available for reuse.
-	 * Make sure a simple formula:
-	 * items_used_ratio > 0.9 && arena_used_ratio > 0.9 &&
-	 * quota_used_ratio > 0.9 work as an indicator
-	 * for reaching Tarantool memory limit.
-	 */
-
 	luaL_pushuint64(L, arena_size);
 	lua_settable(L, -3);
 	/**
@@ -195,7 +185,6 @@ lbox_slab_info(struct lua_State *L)
 	 * data (tuples and indexes).
 	 */
 	lua_pushstring(L, "arena_used");
-	/** System allocator does not use arena. */
 	luaL_pushuint64(L, arena_used);
 	lua_settable(L, -3);
 
```

locker (changes requested):

src/box/lua/slab.cc
```diff
@@ -31,6 +31,7 @@
 #include "trivia/util.h"
 
 #include "box/lua/slab.h"
+#include "box/box.h"
```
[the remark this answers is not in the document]
SEChudinov: Fix it

src/box/lua/slab.cc
SEChudinov: Fix it

src/box/lua/slab.cc
SEChudinov: Fix it

src/box/lua/slab.cc
SEChudinov: Fix it

Totktonada (approved):
I've no objections.

locker (approved):

locker:
Clang CI workflows failed to build because `box_slab_info` lacked `API_EXPORT` in the C file. I added it.

```diff
diff --git a/src/box/box.cc b/src/box/box.cc
index 48960f7fbdef..b3198568a906 100644
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4235,7 +4235,7 @@ box_info_lsn(void)
 /**
  * Get memtx status information for box.slab.info
  */
-uint64_t
+API_EXPORT uint64_t
 box_slab_info(enum box_slab_info_type type)
 {
 	struct memtx_engine *memtx;
```

</details>

Serpentian:
Backport to 3.2: https://github.com/tarantool/tarantool/pull/10723

PR #10509, opened 2024-09-02
say: replace panic macro with a function

### LINKED ISSUE #5365: Turn panic macro into a function
There is already defined `panic` macro in [say.h](https://github.com/tarantool/tarantool/blob/master/src/lib/core/say.h) header. There is also LuaJIT internal `panic` routine on Lua global state. Their mix might lead to ridiculous preprocessor results: the LuaJIT `panic` call (i.e. `g->panic(L)`) is misexpanded and compilation fails with the corresponding error.

@Gerold103 proposed in [his review](https://lists.tarantool.org/pipermail/tarantool-patches/2020-September/019777.html) to turn panic macro into a function.

mandesero:
This patch refactors the panic macro into a function to prevent preprocessor conflicts. Previously, the panic macro in `src/lib/core/say.h` could clash with LuaJIT's internal panic routine, causing macro expansion issues and compilation failures. By using a function instead, the patch resolves these conflicts and ensures correct handling of panic calls.

Closes #5365

NO_DOC=refactoring
NO_TEST=refactoring
NO_CHANGELOG=refactoring

### DIFF
--- a/src/lib/core/say.c
+++ b/src/lib/core/say.c
@@ -73,6 +73,19 @@ void
 enum say_format log_format = SF_PLAIN;
 enum { SAY_SYSLOG_DEFAULT_PORT = 512 };
 
+NORETURN void
+panic(const char *format, ...)
+{
+	char buf[1024];
+	va_list args;
+
+	va_start(args, format);
+	vsnprintf(buf, sizeof(buf), format, args);
+	va_end(args);
+
+	panic_status(EXIT_FAILURE, "%s", buf);
+}
+
 /** List of logs to rotate */
 static RLIST_HEAD(log_rotate_list);
 
--- a/src/lib/core/say.h
+++ b/src/lib/core/say.h
@@ -418,11 +418,13 @@ _say_strerror(int errnum);
 	lsan_turn_off(); \
 	exit(status); \
 })
-#define panic(...)			panic_status(EXIT_FAILURE, __VA_ARGS__)
 #define panic_syserror(...)		({ \
 	say(S_FATAL, tt_strerror(errno), __VA_ARGS__); \
 	exit(EXIT_FAILURE); \
 })
+
+NORETURN void
+panic(const char *format, ...);
 
 /**
  * Log a message once.
--- a/src/lua/utils.c
+++ b/src/lua/utils.c
@@ -1097,15 +1097,6 @@ tarantool_lua_utils_init(struct lua_State *L)
 	return 0;
 }
 
-/*
- * XXX: There is already defined <panic> macro in say.h header
- * (included in diag.h). As a result the call below is misexpanded
- * and compilation fails with the corresponding error. To avoid
- * this error the macro is undefined since it's not used anymore
- * in scope of this translation unit.
- */
-#undef panic
-
 /**
  * This routine encloses the checks and actions to be done when
  * the running fiber yields the execution.


ligurio (commented):
Max, thanks for teh patch!

in commit message:

- please replace prefix (`c`) to `say`
- usually we add an example of error that patch fixes, please add

mandesero:
> Max, thanks for teh patch!
> 
> in commit message:
> 
> * please replace prefix (`c`) to `say`
> * usually we add an example of error that patch fixes, please add

Sergey, thanks for feedback!

I’ve replaced the prefix (`c`) with `say` in the commit message and head of this PR. I’ve also added an example of the error that this patch fixes.

ligurio (approved):

mkokryashkin (changes requested):
If we look at this issue from a broader perspective, we'll see a more significant issue: our build process is broken and it leaks tarantool's definitions into a subproject.

We need to resolve this by changing the CMakeLists and not the sources.

mandesero:
> If we look at this issue from a broader perspective, we'll see a more significant issue: our build process is broken and it leaks tarantool's definitions into a subproject.
> 
> We need to resolve this by changing the CMakeLists and not the sources.

AFAIS, changing the CMakeLists either isn't feasible in this case or would require restructuring the project globally, because the `say.h` header is included in the `api_headers` in `src/CMakeLists.txt`.

https://github.com/tarantool/tarantool/blob/bf091358806ed17bf44efd2cf382a43c0ba49fe0/src/CMakeLists.txt#L249-L254

Hiding this macro would require setting the `PRIVATE` option, but the `panic` macro is used in many parts of the project, and hiding it only for specific modules seems odd (especially since there are modules that use both the `panic` macro and the `G(L)->panic` function (see https://github.com/tarantool/tarantool/blob/bf091358806ed17bf44efd2cf382a43c0ba49fe0/src/lua/utils.c)).

Therefore, I see a few possible solutions:

1. Use `#undef panic` when the macro is not needed in that part of the code;

https://github.com/tarantool/tarantool/blob/bf091358806ed17bf44efd2cf382a43c0ba49fe0/src/lua/utils.c#L1087-L1094

2. Use parentheses to call the function, like this: `(G(L)->panic)(L)`;
3. Replace the macro with a function.

However, points 1 and 2 require developers to constantly keep this subtle issue in mind. Option 3 would eliminate this "headache."

I may be missing something, as I don't have a deep understanding of the topic. If you already have ideas about what might be wrong in the CMakeLists, please share them.

-----

<details>
<summary>Example of error:</summary>


Let’s assume we have the following situation in `say.h`:

```c
// say.h
...
#define panic(msg) fprintf(stderr, "Panic: %s\n", msg)
...
```

And we have the following code in `another.c` that includes this header:

```c
// another.c
...
{
    global_State *g = G(L);
    ...
    // Try to call g->panic.
    // The preprocessor will replace g->panic with the panic macro, which will cause an error.
    g->panic("An error occurred!");
    ...
}
...
```

When you try to compile this code, you will get the following
compilation error:

```bash
error: ‘struct global_State’ has no member named ‘fprintf’
   xxx |     g->panic("An error occurred!");
```

</details>

igormunkin:
> If we look at this issue from a broader perspective, we'll see a more significant issue: our build process is broken and it leaks tarantool's definitions into a subproject.

This is not about the build process, but rather the hell named `src/lua/utils.c`, we ought to support.

> We need to resolve this by changing the CMakeLists and not the sources.

I see no reason for `panic` being macro instead of vararg function: there is nothing about performance or "metaprogramming". I would be glad to finally see this changeset, reducing tweaks and `XXX` comments over the codebase, in the trunk. 

P.S. @mandesero, please drop the comment and `#undef` you've mentioned above. There is no need for it anymore.

igormunkin (commented):
> If we look at this issue from a broader perspective, we'll see a more significant issue: our build process is broken and it leaks tarantool's definitions into a subproject.

This is not about the build process, but rather the hell named `src/lua/utils.c`, we ought to support.

> We need to resolve this by changing the CMakeLists and not the sources.

I see no reason for `panic` being macro instead of vararg function: there is nothing about performance or "metaprogramming". I would be glad to finally see this changeset, reducing tweaks and `XXX` comments over the codebase, in the trunk. 

P.S. @mandesero, please drop the comment and `#undef` you've mentioned above. There is no need for it anymore.

mandesero:
> I see no reason for `panic` being macro instead of vararg function: there is nothing about performance or "metaprogramming". I would be glad to finally see this changeset, reducing tweaks and `XXX` comments over the codebase, in the trunk.

@igormunkin, thanks for the review!

I’ve updated the patch, but I believe it’s still worth discussing with @mkokryashkin. He wanted to figure out how he could make changes to the build process without changing the source code.

PR #10513, opened 2024-09-03
Bump msgpuck submodule

### LINKED ISSUE #9965: Ability to insert duplicate keys in unique index
If not using the most efficient representation we can insert duplicate keys in unique index:

Tarantool 2.11.2-0-g1bac2d2

```lua
tarantool> s = box.schema.space.create('test')
---
...

tarantool> s:create_index('pk')
---
- unique: true
  parts:
  - type: unsigned
    is_nullable: false
    fieldno: 1
  hint: true
  id: 0
  type: TREE
  space_id: 512
  name: pk
...

tarantool> msgpack = require('msgpack')
---
...

tarantool> s:insert{1}
---
- [1]
...

tarantool> o = msgpack.object_from_raw('\x91\xcf\x00\x00\x00\x00\x00\x00\x00\x01')
---
...

tarantool> o:decode()
---
- [1]
...

tarantool> s:insert(o)
---
- [1]
...

tarantool> s:select()
---
- - [1]
  - [1]
...

```

Gerold103:
This update pulls the following commit:

* Fix mp_compare_uint()

Closes #9965

NO_DOC=bugfix
NO_TEST=tested in the submodule

### DIFF
--- a/changelogs/unreleased/gh-9965-wrong-uint-comparison.md
+++ b/changelogs/unreleased/gh-9965-wrong-uint-comparison.md
@@ -1,0 +1,5 @@
+## bugfix/core
+
+- Fixed a bug in the unsigned numbers comparison in tuple keys. It allowed to
+  insert the same key multiple times into a unique index, and sometimes wouldn't
+  allow to find an existing key in an index (gh-9965).


Gumix:
> Closes https://github.com/tarantool/tarantool/issues/9965

What about the case when one positive number is encoded as "signed integer", while another same number as "unsigned integer"?

changelogs/unreleased/gh-9965-wrong-uint-comparison.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+- Fixed a bug in the unsigned numbers comparison in tuple keys. It allowed to
+  insert the same key multiple times into a unique index, and sometimes wouldn't
+  allow to find an existing key in an index (gh-9965).
```
locker, line 5: I'm not sure it fixes the issue for hash index. AFAIR we don't decode integers to compute the hash.

src/lib/msgpuck
locker, line 1: Even though its unit-tested in msgpuck, I'd still add a functional test for Tarantool checking that it isn't possible to insert two different numbers in a unique index anymore.

p7nov (approved):

changelogs/unreleased/gh-9965-wrong-uint-comparison.md
Gerold103: Thanks for noticing! Indeed, hash index is also broken 🤯. We do decode, but only for indexes having a single key part with unsigned type. Otherwise fully broken.

I am currently investigating how to fix it with the smallest perf decrease.

Gerold103:
Suppressed by #10883.

Issue #10459, opened 2024-08-21
Add support for automatic rollback of asynchronous transactions from an old term when synchronous replication is used


CuriousGeorgiy:
Using asynchronous replication with synchronous replication leads to split-brain when a new leader gets appointed without knowledge of all the async transactions of the former leader. Such cases are quite common, it's impossible to guarantee that async transactions from the old leader reach the new leader in time. For example, the old leader might be unreachable, or simply die before replication takes place.

The user has a choice here: either not to mix async replication with synchronous replication at all, or face spurious split-brain cases each time the situation above occurs.

The reason for raising a split-brain error in a situation described above is that we can't keep the async transactions from an old term - they might overwrite something written by the new leader in the new term, and we can't roll them back, because we don't have the necessary info to do so.

Fixing this error requires manual intervention and a rebootstrap of the stray node. The conflicting transactions are also lost.

In order to simplify administration, let's add support for automatic rollback of asynchronous transactions from an old term by keeping asynchronous transactions in the synchronous queue until they get confirmed on a quorum of replicas.

The feature will be controlled by a new `replication_split_brain_handling_mode` enumeration option. For now, the only available mode will be `rollback`, which automatically rolls back conflicting asynchronous transactions in the event of a split-brain situation.

This feature can lead to data loss, so in order to make it safe, we will allow the user to intervene by adding a `box.ctl.on_replication_split_brain_rollback` event in the scope of #10460.

sergepetrenko:
Superseded by #10583

PR #10530, opened 2024-09-05
box: make asynchronous transactions bump the limbo's confirmed LSN

### LINKED ISSUE #10528: Asynchronous transactions must bump the confirmation boundary of the synchronous queue
Currently, we cannot detect a split-brain situation on the old leader when a promote arrives, and the old leader has committed asynchronous transactions that are not confirmed by the promote. We can only detect this on the new leader if he sees asynchronous transactions replicated to him from the old leader.

This is because asynchronous transactions do not bump the confirmation boundary of the synchronous queue, which is used to detect such a split-brain situation. Even though asynchronous transactions can completely bypass the synchronous queue, they must still bump the confirmation boundaries as if they were instantly confirmed (since asynchronous transactions are _per se_ synchronous transactions with quorum 1).

CuriousGeorgiy:
This patch set makes asynchronous transactions bump the synchronous queue's confirmed LSN.

Closes #10528
Needed for #10459

### DIFF
--- a/changelogs/unreleased/gh-10528-async-tx-bump-confirmation-boundary.md
+++ b/changelogs/unreleased/gh-10528-async-tx-bump-confirmation-boundary.md
@@ -1,0 +1,6 @@
+## feature/replication
+
+* Made asynchronous transactions bump the confirmation boundary of the
+  synchronous queue allowing to detect a split-brain situation on an old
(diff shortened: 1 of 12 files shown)


CuriousGeorgiy:
Superseded by #10599.

PR #10533, opened 2024-09-06
test: enable remote console for debug

ligurio:
The remote console is enabled for debugging purposes. It can be helpful when Tarantool is stuck and need an access to console to gather an additional information about state.

NO_CHANGELOG=testing
NO_DOC=testing

### DIFF
--- a/test/fuzz/lua/test_engine.lua
+++ b/test/fuzz/lua/test_engine.lua
@@ -12,6 +12,7 @@ testing if it exists.
 Usage: tarantool test_engine.lua
 ]]
 
+local console = require('console')
 local fiber = require('fiber')
 local fio = require('fio')
 local fun = require('fun')
@@ -1371,6 +1372,10 @@ end
 
 local function run_test(num_workers, test_duration, test_dir,
                         engine_name, verbose_mode)
+
+    local socket_path = ('unix/:/%s/console.sock'):format(fio.abspath(test_dir))
+    console.listen(socket_path)
+    log.info(('console listen on %s'):format(socket_path))
 
     if fio.path.exists(test_dir) then
         cleanup_dir(test_dir)


test/fuzz/lua/test_engine.lua
```diff
@@ -1372,6 +1373,10 @@ end
 local function run_test(num_workers, test_duration, test_dir,
                         engine_name, verbose_mode)
 
+    local socket_path = ('unix/:/tmp/tarantool_%s.sock'):format(seed)
```
locker, line 1376: I think it's better to create the socket in the test directory. And we should delete it at exit:

```lua
test_dir = fio.abspath(test_dir)
...
local sock = console.listen(fio.pathjoin(test_dir, 'console.sock'))
...
sock:close()
```

ligurio: Fixed a path:

```diff
--- a/test/fuzz/lua/test_engine.lua
+++ b/test/fuzz/lua/test_engine.lua
@@ -1373,7 +1373,8 @@ end
 local function run_test(num_workers, test_duration, test_dir,
                         engine_name, verbose_mode)
 
-    local socket_path = ('unix/:/tmp/tarantool_%s.sock'):format(seed)
+    local socket_path =
+        ('unix/:/%s/console_%s.sock'):format(test_dir_name, seed)
     console.listen(socket_path)
     log.info(('console listen on %s'):format(socket_path))
 ```

>  And we should delete it at exit:

I'm afraid that we will destroy a socket and tarantool not exit successfully after this...

Comment (locker) [reply test/fuzz/lua/test_engine.lua]: You don't need to use a seed to generate the file name. After restart you can listen on the same socket. I think using `console.sock` without a seed would be more convenient.

Also, if the test directory is relative, this won't work, as expected.

I suggest using `fio.abspath` and `fio.pathjoin` here. If the path starts with `/', there's no need to prepend it with `unix:/`.

```lua
test_dir = fio.abspath(test_dir)
...
console.listen(fio.pathjoin(test_dir, 'console.sock'))
```

locker (approved):

locker (approved):

locker:
Cherry-picked to 3.2.

PR #10535, opened 2024-09-06
vinyl: use VERBOSE level for logging ranges

### LINKED ISSUE #10524: Don't print vinyl key ranges to log
When a key range is dumped or compacted, Vinyl logs the range boundaries. For example, run the following script:

```Lua
local digest = require('digest')

box.cfg{}

s:create_index('pk', {page_size = 128, range_size = 1024})

while true do
    for i = 1, 100 do
        s:replace({math.random(1000 * 1000), digest.urandom(1000)})
        box.snapshot()
    end
end
```

In the log you'll see messages like:

```
2024-09-05 16:41:05.933 [481595] main/109/vinyl.scheduler I> 512/0: split range ([80972]..[114462]) by key [97320]
2024-09-05 16:41:05.936 [481595] main/109/vinyl.scheduler I> 512/0: started compacting range ([80972]..[97320]), runs 2/2
2024-09-05 16:41:05.936 [481595] main/109/vinyl.scheduler I> 512/0: started compacting range ([97320]..[114462]), runs 2/2
```

With numeric keys these messages look fine but with binary string keys they get really annoying. Let's replace range boundaries with range ids.

locker:
Whenever a range is compacted, split, or coalesced, we log the range boundaries. This gets really annoying if there's an index that has a lot of key parts or contains binary strings. Let's lower the level used for logging these events down to VERBOSE so that they are not shown by default but can be enabled if needed.

Closes #10524

### DIFF
--- a/changelogs/unreleased/gh-10524-use-verbose-level-for-logging-vinyl-ranges.md
+++ b/changelogs/unreleased/gh-10524-use-verbose-level-for-logging-vinyl-ranges.md
@@ -1,0 +1,4 @@
+## bugfix/vinyl
+
+* Lowered the level used for logging range dump, compaction, split, and
+  coalesce events from `INFO` to `VERBOSE` (gh-10524).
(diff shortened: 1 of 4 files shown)


p7nov (approved):

nshy (approved):

locker:
Cherry-picked to 2.11 and 3.2.

PR #10547, opened 2024-09-09
test/fuzz: fix socket path in an engine test

ligurio:
The function `console.listen` does not accept a path with duplicate slashes, so absolute pathes were rejected. The patch fixes that by building a path to Unix socket using `fio.pathjoin` that avoid double slashes. Follows up the commit 5d18b84e1e76 ("test: enable remote console for debug").

NO_CHANGELOG=testing
NO_DOC=testing
NO_TEST=testing


locker (approved):

locker:
Cherry-picked to 3.2.

Issue #10550, opened 2024-09-10
Properly enable LUAJIT_USE_ASAN for Tarantool ASan build 


Buristan:
Since #3071 is closed, it will be nice to enable [the corresponding build flag](https://github.com/tarantool/tarantool/blob/e88d3325ee42aa6265f7f49853adaa02be852274/cmake/luajit.cmake#L69-L82).

PR #10554, opened 2024-09-10
cmake: turn on sysmalloc with ASAN build

mandesero:
This patch enables `LUAJIT_USE_SYSMALLOC` in ASAN build when also GC64 mode is enabled (to avoid allocator and pointer bitness incompatibility).

Closes https://github.com/tarantool/tarantool/issues/10550
Follows up https://github.com/tarantool/tarantool/issues/3071

NO_TEST=internal
NO_DOC=internal
NO_CHANGELOG=internal

### DIFF
--- a/cmake/luajit.cmake
+++ b/cmake/luajit.cmake
@@ -66,17 +66,21 @@ if(ENABLE_VALGRIND)
         "Valgrind support" FORCE)
 endif()
 
-# Enable LuaJIT ASan support. The internal LuaJIT memory allocator
-# is not instrumented yet unfortunately, so to find any memory
-# faults it's worth building LuaJIT with system provided memory
-# allocator (i.e. enable LUAJIT_USE_SYSMALLOC option). However,
-# Tarantool doesn't finalize Lua universe the right way (see the
-# comments near <tarantool_lua_free>), so running Tarantool
-# testing routine with LUAJIT_USE_SYSMALLOC enabled generates
-# false-positive LSan leaks. Return back here to enable
-# LUAJIT_USE_SYSMALLOC, when the issue below is resolved.
-# https://github.com/tarantool/tarantool/issues/3071
+# Enable LuaJIT ASan support.
 if(ENABLE_ASAN)
+    if(LUAJIT_ENABLE_GC64)
+        set(LUAJIT_USE_SYSMALLOC ON CACHE BOOL
+            "System provided memory allocator (realloc/malloc)" FORCE)
+    else()
+        message(WARNING
+            "The internal LuaJIT memory allocator is not instrumented yet,"
+            " so to find any memory faults, it's worth building LuaJIT with"
+            " the system provided memory allocator. This requires enabling"
+            " the LUAJIT_ENABLE_GC64 option, as sysmalloc can only be used"
+            " in GC64 mode. Without it, sysmalloc cannot be used due to"
+            " allocator and pointer bitness incompatibility."
+        )
+    endif()
     set(LUAJIT_USE_ASAN ON CACHE BOOL
         "Build LuaJIT with AddressSanitizer" FORCE)
 endif()


cmake/luajit.cmake
```diff
@@ -69,14 +69,10 @@ endif()
...
-# allocator (i.e. enable LUAJIT_USE_SYSMALLOC option). However,
-# Tarantool doesn't finalize Lua universe the right way (see the
-# comments near <tarantool_lua_free>), so running Tarantool
-# testing routine with LUAJIT_USE_SYSMALLOC enabled generates
-# false-positive LSan leaks. Return back here to enable
-# LUAJIT_USE_SYSMALLOC, when the issue below is resolved.
-# https://github.com/tarantool/tarantool/issues/3071
+# allocator (i.e. enable LUAJIT_USE_SYSMALLOC option).
 if(ENABLE_ASAN)
+    set(LUAJIT_USE_SYSMALLOC ON CACHE BOOL
```
Buristan, line 74: For non-GC64 builds we can't use sysalloc since the bitness of the allocator and pointers should be the same. See [the following build failure](https://github.com/tarantool/tarantool/actions/runs/10793652440/job/29936087657?pr=10554#step:8:3760).

I suppose we should enable this flag only when the `LUAJIT_ENABLE_GC64` is given as well.

Buristan (changes requested):
Hi, Maxim!
Thanks for the patch!

Please reference #10550 to be closed. 
Also, please consider my comment below.

cmake/luajit.cmake
```diff
@@ -69,14 +69,15 @@ endif()
...
-# testing routine with LUAJIT_USE_SYSMALLOC enabled generates
-# false-positive LSan leaks. Return back here to enable
-# LUAJIT_USE_SYSMALLOC, when the issue below is resolved.
-# https://github.com/tarantool/tarantool/issues/3071
+# allocator (i.e. enable LUAJIT_USE_SYSMALLOC option). Also we need
+# to set LUAJIT_ENABLE_GC64 because for non-GC64 builds we can't
+# use sysmalloc since the bitness of the allocator and pointers should
+# be the same.
 if(ENABLE_ASAN)
+    set(LUAJIT_ENABLE_GC64 ON CACHE BOOL
```
Buristan, line 77: I suppose we shouldn't silently enable GC64 when the corresponding flag isn't set.
I think we should enable `LUAJIT_USE_SYSMALLOC` only for GC64 build and raise a _warning_ otherwise.

mandesero: Okay, `LUAJIT_USE_SYSMALLOC` is set only when `GC64=ON`.

cmake/luajit.cmake
```diff
@@ -66,17 +66,21 @@ if(ENABLE_VALGRIND)
...
-# comments near <tarantool_lua_free>), so running Tarantool
-# testing routine with LUAJIT_USE_SYSMALLOC enabled generates
-# false-positive LSan leaks. Return back here to enable
-# LUAJIT_USE_SYSMALLOC, when the issue below is resolved.
-# https://github.com/tarantool/tarantool/issues/3071
+# Enable LuaJIT ASan support.
 if(ENABLE_ASAN)
+    if (LUAJIT_ENABLE_GC64)
+        set(LUAJIT_USE_SYSMALLOC ON CACHE BOOL
+            "System provided memory allocator (realloc)" FORCE)
```
ligurio, line 73: I would remove `(realloc)` or replace with `(realloc/malloc)`.

ligurio (approved):
LGTM after fixing comments from Sergey K.

cmake/luajit.cmake
mandesero: Replaced with `(realloc/malloc)`.

cmake/luajit.cmake
```diff
@@ -66,17 +66,21 @@ if(ENABLE_VALGRIND)
...
+        set(LUAJIT_USE_SYSMALLOC ON CACHE BOOL
+            "System provided memory allocator (realloc/malloc)" FORCE)
+    else()
+        message(WARNING
+            "The internal LuaJIT memory allocator is not instrumented yet,"
+            " so to find any memory faults, it's worth building LuaJIT with"
+            " the system provided memory allocator. This requires enabling"
+            " the LUAJIT_ENABLE_GC64 option, as sysmalloc can only be used"
+            " in GC64 mode. Without it, sysmalloc cannot be used due to"
+            " allocator and pointer bitness incompatibility"
```
Buristan, line 81: Minor: Missed dot at the end of the sentence.

cmake/luajit.cmake
```diff
@@ -66,17 +66,21 @@ if(ENABLE_VALGRIND)
...
-# allocator (i.e. enable LUAJIT_USE_SYSMALLOC option). However,
-# Tarantool doesn't finalize Lua universe the right way (see the
-# comments near <tarantool_lua_free>), so running Tarantool
-# testing routine with LUAJIT_USE_SYSMALLOC enabled generates
-# false-positive LSan leaks. Return back here to enable
-# LUAJIT_USE_SYSMALLOC, when the issue below is resolved.
-# https://github.com/tarantool/tarantool/issues/3071
+# Enable LuaJIT ASan support.
 if(ENABLE_ASAN)
+    if (LUAJIT_ENABLE_GC64)
```
Buristan, line 71: Typo: s/if (/if(/

Buristan (changes requested):
Hi, Maxim!
Thanks for the fixes!
LGTM, after fixing a couple of nits below.

> This patch enables `LUAJIT_USE_SYSMALLOC` in ASAN build after
resolution of #3071.
> 
> NO_TEST=internal
NO_DOC=internal
NO_CHANGELOG=internal

I would rephrase it like the following (to accent that this is done only for GC64 mode):

> This patch enables `LUAJIT_USE_SYSMALLOC` in ASAN build when also GC64 mode is enabled (to avoid allocator and pointer bitness incompatibility). 
> 
> Closes #10550
> Follows up #3071

cmake/luajit.cmake
mandesero: Fixed.

cmake/luajit.cmake
mandesero: Fixed.

Buristan (approved):

cmake/luajit.cmake
ligurio: Thanks!

PR #10584, opened 2024-09-17
Add `encode_key_order` option to json encoder

### LINKED ISSUE #10606: json: support map serialization in particular order
Similar to #9747 but for json.

Part-of [TNTP-196](https://jira.vk.team/browse/TNTP-196).

## Use cases (see #9747)

Main reason is to be able to order keys in json serialization in log messages with `log_format = 'json'`. It is crucial to have such feature to be able to setup custom log context and the log messages still printing the information in predictable order. 

## Implementation

We have a serializer class in [src/lua/serializer.c](https://github.com/tarantool/tarantool/blob/ea6062964828015e5e1f376c6ec2f0b7829b40c2/src/lua/serializer.c) and [src/lua/serializer.h](https://github.com/tarantool/tarantool/blob/ea6062964828015e5e1f376c6ec2f0b7829b40c2/src/lua/serializer.h), which can accept options as second argument in `encode()` and `decode()` calls. Lets add an `encode_key_order` option, which can accept a table (array) of ordered keys, in the order we want them present in the encoded string.

It should work something like:

```tarantool
tarantool> json = require('json')
---
...

tarantool> map = {a = 1, c = 100, b = 2}
---
...

tarantool> json.encode(map)
---
- '{"b":2,"a":1,"c":100}'
...

tarantool> json.encode(map, {encode_key_order = {'a', 'b', 'c'}})
---
- '{"a":1,"b":2,"c":100}'
...

tarantool> json.encode(map, {encode_key_order = {'a', 'b'}})
---
- '{"a":1,"b":2,"c":100}'
...

tarantool> json.encode(map, {encode_key_order = {'a', 'b', 'd'}})
---
- '{"a":1,"b":2,"c":100}'
...
```

For already created serializer object we should be able to create a persistent ordering:
```tarantool
tarantool> json = require('json')
---
...

tarantool> map_1 = {a = 1, c = 100, b = 2}
---
...

tarantool> j = json.new()
---
...

tarantool> j.cfg({encode_key_order = {'a', 'b'}})
---
...

tarantool> j.encode(map_1)
---
- '{"a":1,"b":2,"c":100}'
...

tarantool> j.encode(map_1, {encode_key_order = {'c', 'b'}})
---
- '{"c":100,"b":2,"a":1}'
...

tarantool> j.encode(map_1)
---
- '{"a":1,"b":2,"c":100}'
...
```

grafin:
This option allows to pass an array of keys that will be used to sort keys in maps during serialization to json. The order of the provided keys in the resulting string will be the same as in the `encode_key_order` array. Keys, present in maps and not included in `encode_key_order` will be serialized after the sorted ones in arbitrary order.

Closes #10606

### DIFF
--- a/changelogs/unreleased/gh-10606-json-encode-key-order.md
+++ b/changelogs/unreleased/gh-10606-json-encode-key-order.md
@@ -1,0 +1,4 @@
+## feature/lua
+
+* Introduced a new `encode_key_order` option for the JSON encoder. It allows
+  you to pass the order of keys in the resulting JSON string (gh-10606).
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -77,6 +77,7 @@ static struct {
 	OPTION(LUA_TBOOLEAN, encode_use_tostring, 0),
 	OPTION(LUA_TBOOLEAN, encode_invalid_as_nil, 0),
 	OPTION(LUA_TBOOLEAN, encode_error_as_ext, 1),
+	OPTION(LUA_TTABLE,   encode_key_order, 0),
 	OPTION(LUA_TBOOLEAN, decode_invalid_numbers, 1),
 	OPTION(LUA_TBOOLEAN, decode_save_metatables, 1),
 	OPTION(LUA_TNUMBER,  decode_max_depth, 128),
@@ -88,8 +89,25 @@ luaL_serializer_create(struct luaL_serializer *cfg)
 {
 	rlist_create(&cfg->on_update);
 	for (int i = 0; OPTIONS[i].name != NULL; i++) {
-		int *pval = (int *) ((char *) cfg + OPTIONS[i].offset);
-		*pval = OPTIONS[i].defvalue;
+		switch (OPTIONS[i].type) {
+		case LUA_TBOOLEAN:
+		case LUA_TNUMBER: {
+			int *pval = (int *)((char *)cfg + OPTIONS[i].offset);
+			*pval = OPTIONS[i].defvalue;
+			break;
+		}
+		case LUA_TTABLE: {
+			if (strcmp(OPTIONS[i].name, "encode_key_order") != 0) {
+				unreachable();
+			}
+
+			char **p = (char **)((char *)cfg + OPTIONS[i].offset);
+			*p = NULL;
+			break;
+		}
+		default:
+			unreachable();
+		}
 	}
 }
 
@@ -97,7 +115,113 @@ void
 luaL_serializer_copy_options(struct luaL_serializer *dst,
 			     const struct luaL_serializer *src)
 {
+	luaL_serializer_options_delete(dst);
 	memcpy(dst, src, offsetof(struct luaL_serializer, end_of_options));
+
+	/*
+	 * Do a deep copy of the encoder key order. Not doing so would lead to
+	 * double free.
+	 */
+	size_t encode_key_order_len = 0;
+	while (src->encode_key_order != NULL &&
+	       src->encode_key_order[encode_key_order_len] != NULL) {
+		encode_key_order_len++;
+	}
+
+	if (encode_key_order_len > 0) {
+		dst->encode_key_order = xcalloc(
+			encode_key_order_len + 1, sizeof(char *));
+
+		for (size_t i = 0; i < encode_key_order_len; i++) {
+			size_t len = strlen(src->encode_key_order[i]);
+			dst->encode_key_order[i] = xcalloc(len + 1,
+							   sizeof(char));
+			memcpy(dst->encode_key_order[i],
+			       src->encode_key_order[i],
+			       len);
+		}
+	} else {
+		dst->encode_key_order = NULL;
+	}
+}
+
+/**
+ * Free memory allocated for encode_key_order.
+ * @param cfg Serializer configuration.
+ */
+static void
+luaL_serializer_free_encode_key_order(struct luaL_serializer *cfg)
+{
+	if (cfg->encode_key_order == NULL) {
+		return;
+	}
+
+	for (size_t i = 0; cfg->encode_key_order[i] != NULL; i++) {
+		free(cfg->encode_key_order[i]);
+	}
+
+	free(cfg->encode_key_order);
+	cfg->encode_key_order = NULL;
+}
+
+void
+luaL_serializer_options_delete(struct luaL_serializer *cfg)
+{
+	assert(cfg);
+	luaL_serializer_free_encode_key_order(cfg);
+	/* This is a partial TRASH(cfg), only for options. */
+	memset(cfg, '#', offsetof(struct luaL_serializer, end_of_options));
+}
+
+/**
+ * Destroy the serializer object, freeing all allocated memory.
+ * @param L Lua stack.
+ * @retval 0.
+ */
+static int
+luaL_serializer_destroy(struct lua_State *L)
+{
+	struct luaL_serializer *cfg = lua_touserdata(L, 1);
+	luaL_serializer_options_delete(cfg);
+	return 0;
+}
+
+/**
+ * Parse encode_key_order table from stack into serializer config.
+ * @param L Lua stack.
+ * @param cfg Serializer configuration.
+ */
+static void
+luaL_serializer_parse_encode_key_order(struct lua_State *L,
+				       struct luaL_serializer *cfg)
+{
+	assert(lua_istable(L, -1) || lua_isnil(L, -1) || luaL_isnull(L, -1));
+
+	luaL_serializer_free_encode_key_order(cfg);
+
+	size_t key_order_len = lua_objlen(L, -1);
+	if (key_order_len > 0) {
+		cfg->encode_key_order = xcalloc(
+			key_order_len + 1, sizeof(char *));
+	}
+
+	for (size_t i = 0, j = 0; i < key_order_len; i++) {
+		lua_pushinteger(L, i + 1);
+		lua_gettable(L, -2);
+
+		/* Skip keys that are not strings or numbers. */
+		if (!lua_isstring(L, -1)) {
+			lua_pop(L, 1);
+			continue;
+		}
+
+		size_t len = 0;
+		const char *key = lua_tolstring(L, -1, &len);
+		cfg->encode_key_order[j] = xcalloc(len + 1, sizeof(char));
+		memcpy(cfg->encode_key_order[j], key, len);
+		lua_pop(L, 1);
+		j++;
+	}
 }
 
 /**
@@ -113,7 +237,7 @@ static int *
 luaL_serializer_parse_option(struct lua_State *L, int i,
 			     struct luaL_serializer *cfg)
 {
-	lua_getfield(L, 2, OPTIONS[i].name);
+	lua_getfield(L, -1, OPTIONS[i].name);
 	if (lua_isnil(L, -1))
 		return NULL;
 	/*
@@ -128,6 +252,12 @@ luaL_serializer_parse_option(struct lua_State *L, int i,
 	case LUA_TNUMBER:
 		*pval = lua_tointeger(L, -1);
 		break;
+	case LUA_TTABLE:
+		if (strcmp(OPTIONS[i].name, "encode_key_order") == 0) {
+			luaL_serializer_parse_encode_key_order(L, cfg);
+			break;
+		}
+		unreachable();
 	default:
 		unreachable();
 	}
@@ -173,6 +303,33 @@ luaL_serializer_cfg(struct lua_State *L)
 	return 0;
 }
 
+void
+luaL_push_encode_key_order(struct lua_State *L, struct luaL_serializer *cfg)
+{
+	if (cfg->encode_key_order == NULL) {
+		lua_pushnil(L);
+		return;
+	}
+
+	lua_newtable(L);
+	for (size_t i = 0; cfg->encode_key_order[i] != NULL; i++) {
+		lua_pushinteger(L, i + 1);
+		lua_pushstring(L, cfg->encode_key_order[i]);
+		lua_settable(L, -3);
+	}
+}
+
+struct luaL_serializer *
+luaL_newserializer_config(struct lua_State *L)
+{
+	struct luaL_serializer *serializer = (struct luaL_serializer *)
+		lua_newuserdata(L, sizeof(*serializer));
+	luaL_getmetatable(L, LUAL_SERIALIZER);
+	lua_setmetatable(L, -2);
+	luaL_serializer_create(serializer);
+	return serializer;
+}
+
 struct luaL_serializer *
 luaL_newserializer(struct lua_State *L, const char *modname,
 		   const luaL_Reg *reg)
@@ -183,11 +340,7 @@ luaL_newserializer(struct lua_State *L, const char *modname,
 	lua_newtable(L);
 
 	/* Create new configuration */
-	struct luaL_serializer *serializer = (struct luaL_serializer *)
-			lua_newuserdata(L, sizeof(*serializer));
-	luaL_getmetatable(L, LUAL_SERIALIZER);
-	lua_setmetatable(L, -2);
-	luaL_serializer_create(serializer);
+	struct luaL_serializer *serializer = luaL_newserializer_config(L);
 
 	for (; reg->name != NULL; reg++) {
 		/* push luaL_serializer as upvalue */
@@ -214,6 +367,12 @@ luaL_newserializer(struct lua_State *L, const char *modname,
 		case LUA_TNUMBER:
 			lua_pushinteger(L, *pval);
 			break;
+		case LUA_TTABLE:
+			if (strcmp(OPTIONS[i].name, "encode_key_order") == 0) {
+				luaL_push_encode_key_order(L, serializer);
+				break;
+			}
+			unreachable();
 		default:
 			unreachable();
 		}
@@ -431,6 +590,158 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
 	lua_replace(L, idx);
 	lua_settop(L, top);
 	return luaL_tofield(L, cfg, idx, field);
+}
+
+/**
+ * Get the next field from the leftovers table in unsorted order.
+ * @param L Lua stack.
+ * @retval 0 if there are no more fields in the table.
+ */
+static int
+luaL_next_field_unsorted(struct lua_State *L)
+{
+	assert(lua_gettop(L) >= 4);
+	assert(lua_isnil(L, -2));
+	assert(lua_istable(L, -3));
+	assert(lua_istable(L, -4));
+
+	/* const table, table, nil, key */
+	int r = lua_next(L, -3);
+
+	if (r == 0) {
+		/*
+		 * No more elements in the table. Delete the copy of the table.
+		 * const table, table, nil
+		 */
+		lua_pop(L, 2);
+		/* const table */
+		return 0;
+	}
+
+	return r;
+}
+
+/**
+ * Get the next field from the initial table in unsorted order.
+ * @param L Lua stack.
+ * @retval 0 if there are no more fields in the table.
+ */
+static int
+luaL_next_field_fallback(struct lua_State *L)
+{
+	assert(lua_gettop(L) >= 3);
+	assert(lua_isnil(L, -2));
+	assert(lua_istable(L, -3));
+
+	/* const table, nil, key */
+	int r = lua_next(L, -3);
+
+	if (r == 0) {
+		/* const table, nil */
+		lua_pop(L, 1);
+		/* const table */
+		return 0;
+	}
+
+	/* const table, nil, key, value */
+	return r;
+}
+
+int
+luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg)
+{
+	/* const table, (table ?), key_idx, key */
+	assert(lua_gettop(L) >= 3);
+	assert(lua_isnil(L, -2) || lua_isnumber(L, -2));
+	assert(lua_istable(L, -3));
+
+	/* No need to sort anything */
+	if (cfg->encode_key_order == NULL) {
+		return luaL_next_field_fallback(L);
+	}
+
+	/*
+	 * Both key_idx and key are nil, first call, copy the table
+	 * so that we can remove printed elements from its copy, leaving
+	 * the initial table intact.
+	 */
+	if (lua_isnil(L, -1) && lua_isnil(L, -2)) {
+		/* const table, nil, nil */
+		lua_pop(L, 2);
+		lua_newtable(L);
+		/* const table, table */
+
+		/* Copy the table. */
+		lua_pushnil(L);
+		/* const table, table, nil */
+		while (lua_next(L, -3)) {
+			/* const table, table, key, value */
+			lua_pushvalue(L, -2);
+			/* const table, table, key, value, key */
+			lua_pushvalue(L, -2);
+			/* const table, table, key, value, key, value */
+			lua_settable(L, -5);
+			/* const table, table, key, value */
+			lua_pop(L, 1);
+			/* const table, table, key */
+		}
+		/* const table, table */
+		lua_pushnil(L);
+		lua_pushnil(L);
+		/* const table, table, nil, nil */
+	}
+
+	/* Going through the unsorted part */
+	if (!lua_isnil(L, -1) && lua_isnil(L, -2)) {
+		return luaL_next_field_unsorted(L);
+	}
+
+	/**
+	 * We are iterating through sorted part of the table
+	 * don't need to know the key.
+	 * const table, table, key_idx, key
+	 */
+	lua_pop(L, 1);
+	/* const table, table, key_idx */
+
+	assert(lua_isnil(L, -1) || lua_isnumber(L, -1));
+
+	/* Start with the first key */
+	lua_Integer key_idx = 0;
+	if (lua_isnumber(L, -1)) {
+		/* Get next key index if it is a number */
+		key_idx = lua_tointeger(L, -1) + 1;
+		assert(key_idx >= 0);
+	}
+	/* const table, table, key_idx */
+	lua_pop(L, 1);
+	/* const table, table */
+
+	/* Find the key in the key order. */
+	char *key = cfg->encode_key_order[key_idx];
+
+	/* encode_key_order end, continue iterating over the unsorted part. */
+	if (key == NULL) {
+		/* const table, table */
+		lua_pushnil(L);
+		lua_pushnil(L);
+		/* const table, table, nil, nil */
+		return luaL_next_field_unsorted(L);
+	}
+
+	/* Push current key_idx, key and value. */
+	/* const table, table */
+	lua_pushinteger(L, key_idx);
+	lua_pushstring(L, key);
+	lua_getfield(L, -3, key);
+	/* const table, table, key_idx, key, value */
+
+	/* Remove the key from the table. */
+	lua_pushnil(L);
+	lua_setfield(L, -5, key);
+
+	/* const table, table, key_idx, key, value */
+	return 5;
 }
 
 int
@@ -803,6 +1114,7 @@ int
 tarantool_lua_serializer_init(struct lua_State *L)
 {
 	static const struct luaL_Reg serializermeta[] = {
+		{"__gc", luaL_serializer_destroy},
 		{NULL, NULL},
 	};
 	luaL_register_type(L, LUAL_SERIALIZER, serializermeta);
--- a/src/lua/serializer.h
+++ b/src/lua/serializer.h
@@ -136,6 +136,8 @@ struct luaL_serializer {
 	int encode_invalid_as_nil;
 	/** Encode error object as MP_ERROR extension (MsgPack only). */
 	int encode_error_as_ext;
+	/** Array of keys to sort map serialization. */
+	char **encode_key_order;
 
 	/** Enables decoding NaN and Inf numbers */
 	int decode_invalid_numbers;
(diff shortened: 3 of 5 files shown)


changelogs/unreleased/gh-10606-json-encode-key-order.md
```diff
@@ -0,0 +1,4 @@
+## feature/lua
+
+* Introduced a new `encode_key_order` option to json encoder. It allows to pass
+  the order of keys in the resulting JSON string (gh-10606).
```
xuniq, line 4:
```suggestion
* Introduced a new `encode_key_order` option for the JSON encoder. It allows you to pass
  the order of keys in the resulting JSON string (gh-10606).
```

xuniq (approved):

changelogs/unreleased/gh-10606-json-encode-key-order.md
grafin: Thanks! Fixed.

grafin:
I have no idea, why `publish-module-api-doc / build-api-doc (pull_request)` fails. If anyone can help me - would appreciate.

Totktonada:
@grafin It seems, it is due to the `ubuntu-latest` image change: https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

src/lua/serializer.c
```diff
@@ -433,6 +552,129 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
...
+			lua_settable(L, -5);
+			/* const table, table, key, value */
+			lua_pop(L, 1);
+			/* const table, table, key */
+		}
+		/* const table, table */
+		lua_pushnil(L);
+		lua_pushnil(L);
+		/* const table, table, nil, nil */
+	}
```
georgiy-belyanin, line 618: Probably, I'm wrong, but there's no need to do a full copy of the table if the encode key order isn't specified and the method `luaL_next_field_unsorted()` doesn't delete the table copy.

I think extra table deletion can be done after going through the whole unsorted part.

Making an unnecessary full copy is quite expensive.

georgiy-belyanin (changes requested):
Hello, thank you for the patch!

Please, consider my notes.

src/lua/serializer.c
grafin: You are right that this might be unnecessary, but then the lau stack will be different for different options of the call. I've tried to avoid it, because I didn't want to deal with it) will look more into it too.

src/lua/serializer.c
```diff
@@ -98,6 +116,79 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
 			     const struct luaL_serializer *src)
 {
 	memcpy(dst, src, offsetof(struct luaL_serializer, end_of_options));
+
+	/**
```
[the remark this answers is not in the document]
grafin: Thanks! Fixed.

src/lua/serializer.c
```diff
@@ -433,6 +552,129 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
 	return luaL_tofield(L, cfg, idx, field);
 }
 
+/**
+ * Get the next field from the table in unsorted order.
+ * @param L Lua stack.
+ * @param cfg Serializer configuration.
```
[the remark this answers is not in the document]
grafin: Thanks for noticing, fixed.

src/lua/serializer.c
```diff
@@ -433,6 +552,129 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
...
+}
+
+int
+luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg)
+{
+	/* At least: const table, key_idx, key */
+	assert(lua_gettop(L) >= 3);
+	assert(lua_istable(L, -3));
+
+	/**
```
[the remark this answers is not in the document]
grafin: Fixed.

src/lua/serializer.c
```diff
@@ -433,6 +552,129 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
...
+static int
+luaL_next_field_unsorted(struct lua_State *L)
+{
+	assert(lua_istable(L, -3));
+
+	/* const table, table, nil, key */
+	int r = lua_next(L, -3);
+
+	if (r == 0) {
+		/**
```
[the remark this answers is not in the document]
grafin: Fixed.

src/lua/serializer.c
```diff
@@ -98,6 +116,79 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
...
+	/**
+	 * Do a deep copy of the encoder key order. Not doing so would lead to
+	 * double free.
+	 */
+	size_t encode_key_order_len;
+	for (encode_key_order_len = 0;
+	     src->encode_key_order != NULL &&
+	     src->encode_key_order[encode_key_order_len] != NULL;
+	     encode_key_order_len++) {
+	}
```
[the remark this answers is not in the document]
grafin: Sure, while loop is shorter and cleaner here, thanks for your suggestion.

src/lua/serializer.c
```diff
@@ -98,6 +116,79 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
...
+ */
+static void
+luaL_serializer_parse_encode_key_order(struct lua_State *L,
+				       struct luaL_serializer *cfg)
+{
+	assert(lua_istable(L, -1) || lua_isnil(L, -1));
+
+	free(cfg->encode_key_order);
+	cfg->encode_key_order = NULL;
+
```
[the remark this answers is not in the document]
grafin: Fixed.

src/lua/serializer.c
grafin: Ok. I've looked into this a bit more, and don't see an easy way to fix it, without complicating the logic and adding more functions similar to `luaL_next_field_unsorted()`.

I use `luaL_next_field_unsorted()` in two conditions:

1. There is no sorting order specified at all.
2. We have finished going through sorted part of the array and only need to go through what is left in arbitrary order.

The second branch obviously needs the copied array, from which we removed elements during iterating over the sorted part.

The first branch I was able to optimize only by introducing a new function, similar to `luaL_next_field_unsorted()`. If you or @Totktonada sees any better solution - feel free to propose, I would gladly improve this part.

<details>

<summary>
Full diff
</summary>

diff --git a/src/lua/serializer.c b/src/lua/serializer.c
index a892d067a5..424c807701 100644
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -551,14 +551,17 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
 }

 /**
- * Get the next field from the table in unsorted order.
+ * Get the next field from the leftovers table in unsorted order.
  * @param L Lua stack.
  * @retval 0 if there are no more fields in the table.
  */
 static int
 luaL_next_field_unsorted(struct lua_State *L)
 {
+       assert(lua_gettop(L) >= 4);
+       assert(lua_isnil(L, -2));
        assert(lua_istable(L, -3));
+       assert(lua_istable(L, -4));

        /* const table, table, nil, key */
        int r = lua_next(L, -3);
@@ -576,13 +579,47 @@ luaL_next_field_unsorted(struct lua_State *L)
        return r;
 }

+/**
+ * Get the next field from the initial table in unsorted order.
+ * @param L Lua stack.
+ * @retval 0 if there are no more fields in the table.
+ */
+static int
+luaL_next_field_fallback(struct lua_State *L)
+{
+       assert(lua_gettop(L) >= 3);
+       assert(lua_isnil(L, -2));
+       assert(lua_istable(L, -3));
+
+       /* const table, nil, key */
+       int r = lua_next(L, -3);
+
+       if (r == 0) {
+               /*
+                * const table, nil
+                */
+               lua_pop(L, 1);
+               /* const table */
+               return 0;
+       }
+
+       /* const table, nil, key, value */
+       return r;
+}
+
 int
 luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg)
 {
-       /* At least: const table, key_idx, key */
+       /* const table, (table ?), key_idx, key */
        assert(lua_gettop(L) >= 3);
+       assert(lua_isnil(L, -2) || lua_isnumber(L, -2));
        assert(lua_istable(L, -3));

+       /* No need to sort anything */
+       if (!cfg->encode_key_order) {
+               return luaL_next_field_fallback(L);
+       }
+
        /*
         * Both key_idx and key are nil, first call, copy the table
         * so that we can remove printed elements from its copy, leaving
@@ -614,11 +651,6 @@ luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg)
                /* const table, table, nil, nil */
        }

-       /* No need to sort anything */
-       if (!cfg->encode_key_order) {
-               return luaL_next_field_unsorted(L);
-       }
-
        /* Going through the unsorted part */
        if (!lua_isnil(L, -1) && lua_isnil(L, -2)) {
                return luaL_next_field_unsorted(L);
diff --git a/third_party/lua-cjson/lua_cjson.c b/third_party/lua-cjson/lua_cjson.c
index 70032f2c5e..c03d7f6787 100644
--- a/third_party/lua-cjson/lua_cjson.c
+++ b/third_party/lua-cjson/lua_cjson.c
@@ -308,9 +308,9 @@ static void json_append_object(lua_State *l, struct luaL_serializer *cfg,
     while (luaL_next_field(l, cfg) != 0) {
         /* Skip nil values */
         if (lua_isnil(l, -1)) {
-            /* const table, table, key_idx, key, value */
+            /* const table, (table ?), key_idx, key, value */
             lua_pop(l, 1);
-            /* const table, table, key_idx, key */
+            /* const table, (table ?), key_idx, key */
             continue;
         }

@@ -341,10 +341,10 @@ static void json_append_object(lua_State *l, struct luaL_serializer *cfg,
             break;
         }

-        /* const table, table, key_idx, key, value */
+        /* const table, (table ?), key_idx, key, value */
         json_append_data(l, cfg, current_depth, json);
         lua_pop(l, 1);
-        /* const table, table, key_idx, key */
+        /* const table, (table ?), key_idx, key */
     }
     /* const table */

</details>

grafin:
Fixed memory copying/releasing and assertion on encode_key_order members being strings or numbers (lets just skip values that are not strings or numbers).

<details>

<summary>
diff:
</summary>

```diff
--- a/src/lua/serializer.c
+++ b/src/lua/serializer.c
@@ -130,8 +130,15 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
        if (encode_key_order_len > 0) {
                dst->encode_key_order = xcalloc(
                        encode_key_order_len + 1, sizeof(char *));
-               memcpy(dst->encode_key_order, src->encode_key_order,
-                      encode_key_order_len * sizeof(char *));
+
+               for (size_t i = 0; i < encode_key_order_len; i++) {
+                       size_t len = strlen(src->encode_key_order[i]);
+                       dst->encode_key_order[i] = xcalloc(len + 1,
+                                                          sizeof(char));
+                       memcpy(dst->encode_key_order[i],
+                              src->encode_key_order[i],
+                              len);
+               }
        } else {
                dst->encode_key_order = NULL;
        }
@@ -144,6 +151,12 @@ luaL_serializer_free_options(struct luaL_serializer *cfg)
                return;
        }

+       if (cfg->encode_key_order) {
+               for (size_t i = 0; cfg->encode_key_order[i] != NULL; i++) {
+                       free(cfg->encode_key_order[i]);
+               }
+       }
+
        free(cfg->encode_key_order);
        cfg->encode_key_order = NULL;
 }
@@ -183,8 +196,16 @@ luaL_serializer_parse_encode_key_order(struct lua_State *L,
        for (size_t i = 0; i < key_order_len; i++) {
                lua_pushinteger(L, i + 1);
                lua_gettable(L, -2);
-               assert(lua_isstring(L, -1));
-               cfg->encode_key_order[i] = (char *)lua_tostring(L, -1);
+
+               /* Skip keys that are not strings or numbers. */
+               if (!lua_isstring(L, -1)) {
+                       continue;
+               }
+
+               size_t len = 0;
+               const char *key = lua_tolstring(L, -1, &len);
+               cfg->encode_key_order[i] = xcalloc(len + 1, sizeof(char));
+               memcpy(cfg->encode_key_order[i], key, len);
                lua_pop(L, 1);
        }
 }
```

</details>

src/lua/serializer.c
Totktonada, line 1: I'm a bit afraid of amount of memory management code in this Lua-centric patch.

I've exprimented a bit with it. It is possible to leave the memory management to Lua GC: we can use an upvalue of the encode method to access the Lua table with options. We can use a Lua table on the Lua stack to access call-local options.

I've created a [PoC](https://github.com/Totktonada/tarantool/commit/e467943d5869a677a716e55d2c4b9c8bdedf9086) based on your patch.

Cons:

* A separate mechnanism for options stored in Lua, not fully integrated into on-update-trigger and copy-options mechanisms of serializers (though, they're only used for msgpack serializer).

Pros:

* No manual memory management.

I'm really not satisfied how it looks, but maybe the idea itself is not so bad. For now, I just want to share it and hear opinions.

src/lua/serializer.c
```diff
@@ -98,6 +116,100 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
...
+ * @param L Lua stack.
+ * @param cfg Serializer configuration.
+ */
+static void
+luaL_serializer_parse_encode_key_order(struct lua_State *L,
+				       struct luaL_serializer *cfg)
+{
+	assert(lua_istable(L, -1) || lua_isnil(L, -1));
+
+	luaL_serializer_free_options(cfg);
```
Totktonada, line 188: If we have several options to free, this frees all of them.

src/lua/serializer.c
```diff
@@ -803,6 +1097,7 @@ int
 tarantool_lua_serializer_init(struct lua_State *L)
 {
 	static const struct luaL_Reg serializermeta[] = {
+		{"__gc", luaL_destroy_serializer},
```
Totktonada, line 1100: It seems correct, but dangerous. I think we should explicitly write the rules for managing the serializer objects, because now there are not-very-obvious constraints.

The best C-Lua interoperability is reached for objects that have a reference counter. It can be allocated from C, pushed to Lua then (many times) and it can be destroyed as from C as well as from Lua (when the refcounter reached zero). The lifetime of the object in C and Lua is arbitrary in the case: just increment the refcounter, use and decrement it. Increment on push to Lua and decrement in the Lua's GC handler.

We're in another situation here. We can't create a serializer object from C and then push it to Lua as a userdata object. We can't create a serializer object in Lua and use it in C after its lifespan in Lua ends.

In fact, we have two separate kinds of serializer objects:

* Serializer objects created in C in an automatic or a dynamic storage and to be destroyed in C. The destroy function (`luaL_serializer_free_options`) must be called on them (including objects in the automatic storage).
* Serializer objects created in Lua in the storage provided by a userdata object. Destroyed automatically, when the userdata is collected. Can be used in C only during a call of a function that references this object (usually using a upvalue) or if the serializer is a kind of always-loaded module.

The convertion from a Lua-citizen to a C-citizen serializer (or vice versa) is possible, but it requires to use the explicit copy constructor: `luaL_serializer_copy_options`.

src/lua/serializer.h
```diff
@@ -243,6 +252,15 @@ struct luaL_field {
...
 };
 
+/**
+ * Get the next field from the table in sorted order.
+ * @param L Lua stack.
+ * @param cfg Serializer configuration.
+ * @retval 0 if there are no more fields in the table.
+ */
+int
+luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg);
```
Totktonada, line 262: Nit: [lua_next](https://pgl.yoyo.org/luai/i/lua_next)'s description explicitly tells what is popped from the stack and what is pushed to it. I find it useful and I would do it here as well.

src/lua/serializer.h
```diff
@@ -184,6 +186,13 @@ void
 luaL_serializer_copy_options(struct luaL_serializer *dst,
 			     const struct luaL_serializer *src);
 
+/**
+ * Free memory allocated for serializer options.
+ * @param cfg Serializer configuration.
+ */
+void
+luaL_serializer_free_options(struct luaL_serializer *cfg);
```
Totktonada, line 194: Nit: It seems, `luaL_serializer_destroy` name fits good here (see also the bottom of the [chapter about naming](https://www.tarantool.io/en/doc/latest/contributing/c_style_guide/#chapter-4-naming) in our C Style Guide). (If we free the options, but continue to use the serializer object, something is definitely wrong.) I would also zap the structure using the `TRASH` macro from `trivia/util.h` to reveal memory management/ownership problems.

src/lua/serializer.c
```diff
@@ -433,6 +573,160 @@ lua_field_tostring(struct lua_State *L, struct luaL_serializer *cfg, int idx,
...
+{
+	/* const table, (table ?), key_idx, key */
+	assert(lua_gettop(L) >= 3);
+	assert(lua_isnil(L, -2) || lua_isnumber(L, -2));
+	assert(lua_istable(L, -3));
+
+	/* No need to sort anything */
+	if (!cfg->encode_key_order) {
+		return luaL_next_field_fallback(L);
+	}
```
Totktonada, line 644: Quote from https://www.tarantool.io/en/doc/latest/contributing/c_style_guide/#chapter-16-other:

> We don’t apply `!` operator to non-boolean values. It means, to check if an integer is not 0, you use `!= 0`. To check if a pointer is not NULL, you use `!= NULL`. The same for `==`.

src/lua/serializer.c
Totktonada: From algorithmic point of view, it is possible. We need to transform the `encode_key_order` array into a hashmap from a key to its ordinal or a next key.

<details>
<summary>Illustration</summary>

```lua
#!/usr/bin/env tarantool

local start_marker = {}
local finish_marker = {}
local next_keys_registry = setmetatable({
    unsorted = {
        [start_marker] = finish_marker,
    },
}, {__mode = 'k'})

local function buf_new()
    return {}
end

local function buf_add(buf, s)
    table.insert(buf, s)
end

local function buf_del(buf)
    table.remove(buf)
end

local function buf_get(buf)
    return table.concat(buf)
end

local function prepare_opts(opts)
    local key_order = opts.encode_key_order
    if key_order == nil or next(key_order) == nil then
        return {
            next_keys = next_keys_registry.unsorted,
        }
    end
    local next_keys = next_keys_registry[key_order]
    if next_keys == nil then
        next_keys = {}
        for i = 1, #key_order do
            local k1 = key_order[i]
            local k2 = key_order[i + 1] or finish_marker
            next_keys[k1] = k2
        end
        next_keys[start_marker] = key_order[1]
        next_keys_registry[key_order] = next_keys
    end
    return {
        next_keys = next_keys,
    }
end

local function encode_kv(buf, key, value)
    buf_add(buf, '"')
    buf_add(buf, key)
    buf_add(buf, '":"')
    buf_add(buf, value)
    buf_add(buf, '"')
end

local function encode(data, opts)
    local prepared_opts = prepare_opts(opts)
    local next_keys = prepared_opts.next_keys

    local buf = buf_new()
    buf_add(buf, '{')

    -- Sorted part.
    local key = next_keys[start_marker]
    while key ~= finish_marker do
        local value = data[key]
        if type(value) ~= 'nil' then
            encode_kv(buf, key, value)
            buf_add(buf, ',')
        end
        key = next_keys[key]
    end

    -- Unsorted part.
    for key, value in pairs(data) do
        if type(next_keys[key]) == 'nil' then
            encode_kv(buf, key, value)
            buf_add(buf, ',')
        end
    end

    buf_del(buf)
    buf_add(buf, '}')

    return buf_get(buf)
end

local data = {a = 'a', b = 'b', c = 'c', d = 'd'}
local opts = {
    encode_key_order = {'a', 'b'},
}
print(encode(data, opts))
```
</details>

From technical point of view, I wouldn't like to review such a code written using the Lua C API.

src/lua/serializer.c
```diff
@@ -98,6 +116,100 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
...
+		dst->encode_key_order = NULL;
+	}
+}
+
+void
+luaL_serializer_free_options(struct luaL_serializer *cfg)
+{
+	if (!cfg) {
+		return;
+	}
```
Totktonada, line 152: Nit: I would assume that the pointer points to a valid serializer object: we usually do so and we do it for serializers in other functions. We can add an assertion.

src/lua/serializer.c
Totktonada: Well, the upvalue+Lua stack value approach seems tricky. I'm OK to stick with all these mallocs. They're more straightforward, at least.

Totktonada (approved):
For me, it still seems complicated. On the other hand, this implementation is quite straightforward: so if we have a related problem, a fix is likely straightforward.

If we'll revisit C-Lua integration around the serializers in a future, the things hopefully become simpler. However, investigating it surely is not what we should do in this pull request.

I'm OK with the patch (left some minor remarks). Thanks!

src/lua/serializer.c
georgiy-belyanin: Thank you for working on it. I'm ok with the solution. Sorry, for the delay.

src/lua/serializer.c
```diff
@@ -98,6 +116,100 @@ luaL_serializer_copy_options(struct luaL_serializer *dst,
...
+
+	if (cfg->encode_key_order) {
```
georgiy-belyanin, line 154: Nit: the comment [on making the NULL comparison explicit](https://github.com/tarantool/tarantool/pull/10584/files#r1817411283) can be applied here too.

georgiy-belyanin (approved):
Thank you for working on it. Approve except 1 more nit. Sorry for the delay.

src/lua/serializer.h
grafin: I've added a detailed comment for the `next_field` function:
```c
/**
 * @brief Get the next field from the table in sorted order.
 *
 * Pops a key and a 'key index' from the stack, and pushes a key, 'key-index',
 * value for the next element in a table provided as a third argument on stack.
 *
 * If cfg->encode_key_order is set, this function iterates over the table in the
 * oreder defined by this array. During this process the 'key-index' is the
 * index of the current key in cfg->encode_key_order array. If there are no more
 * elements in the cfg->encode_key_order array this function iterates over the
 * unvisited elements of the table in an arbitrary order.
 *
 * During the iterating this function may push a copy (which may be modified by
 * the iterator) of the provided table on the stack which will be popped when
 * there are no more elements left in the table.
 *
 * When the iterator reaches the last element of the table this function returns
 * 0 and doesn't push anything on the stack, leaving the original table.
 *
 * A typical traversal looks like this:
 *
 * // cfg->encode_key_order is an array with the order of the keys
 * // table is on the top of the stack
 * lua_pushnil(L);  // first key index
 * lua_pushnil(L);  // first key
 * while (luaL_next_field(l, cfg) != 0) {
 *     printf("%s - %s\n",
 *            lua_typename(L, lua_type(L, -2)),  // key
 *            lua_typename(L, lua_type(L, -1))); // value
 *     lua_pop(L, 1); // remove value, keep key and key index for next iteration
 * }
 *
 * @param L Lua stack.
 * @param cfg Serializer configuration.
 * @retval 0 if there are no more fields in the table.
 */
int
luaL_next_field(struct lua_State *L, struct luaL_serializer *cfg);
```

src/lua/serializer.h
grafin: I've tried to apply the naming guide, and decided to name this one `luaL_serializer_delete`. Because it doesn't really destroys the serialezer it frees memory, used for it and it is also now used in options copy function:
```c
void
luaL_serializer_copy_options(struct luaL_serializer *dst,
			     const struct luaL_serializer *src)
{
	luaL_serializer_delete(dst);
	memcpy(dst, src, offsetof(struct luaL_serializer, end_of_options));
...
```

It would be strange to destroy the object to which we are going to copy something.

src/lua/serializer.c
grafin: Thanks, fixed.

src/lua/serializer.c
grafin: Thanks, fixed.

grafin:
Looks like some of my fixes broke the memory managment in CI (strangely enough, I ran the tests locally before pushhing several times and it went fine). I will investigate WTF and call everyone for one more time, ones I'm done)

src/lua/serializer.c
Totktonada: At least, now we have some brief suggestions. Thanks!

src/lua/serializer.h
Totktonada: > oreder

Thanks for the update!

src/lua/serializer.h
Totktonada: As for me, `_delete` for a variable on a stack looks strange. My expectation is that I can't access this storage anymore after `_delete`, just like after the `delete` keyword in C++.

grafin:
> Looks like some of my fixes broke the memory managment in CI

I've found the issue:
third_party/lua-cjson/lua_cjson.c:
```diff
@@ -439,7 +439,7 @@ static int json_encode(lua_State *l) {
     struct luaL_serializer *cfg = luaL_checkserializer(l);
 
     if (lua_gettop(l) == 2) {
-        struct luaL_serializer user_cfg;
+        struct luaL_serializer user_cfg = {};
         luaL_serializer_copy_options(&user_cfg, cfg);
         luaL_serializer_parse_options(l, &user_cfg);
         lua_pop(l, 1);
```

Didn't reproduce locally because of ASAN build, I guess ASAN does some extra memset to 0 under the hood...

I've also update the `luaL_serializer_copy_options` comment to mention this.
```c
/**
 * @brief Copy all option fields of @a src into @a dst.
 * Other fields, such as triggers, are not touched.
 *
 * This will try to free memory allocated for the options in @a dst. If @a dst
 * has been just allocated, it should be initialized with 0s.
 *
 * This may allocate memory on heap for some of the options, your @a dst
 * is created in C code, don't forget to call luaL_serializer_options_delete
 * when the serializer object is not needed anymore to prevent memory leaks.
 *
 * If the serializer object was created in lua, this will be done during garbage
 * collection automatically.
 *
 * @param dst Destination serializer to copy options into.
 * @param src Source serializer to copy options from.
 */
void
luaL_serializer_copy_options(struct luaL_serializer *dst,
			     const struct luaL_serializer *src);
```

Fixed the 'oreder' typo.

And renamed `luaL_serializer_delete` to `luaL_serializer_options_delete`.

src/lua/serializer.h
Totktonada: > And renamed luaL_serializer_delete to luaL_serializer_options_delete.

OK for me. Resolving.

third_party/lua-cjson/lua_cjson.c
```diff
@@ -423,10 +439,13 @@ static int json_encode(lua_State *l) {
...
 
     if (lua_gettop(l) == 2) {
-        struct luaL_serializer user_cfg = *cfg;
+        struct luaL_serializer user_cfg;
+        luaL_serializer_create(&user_cfg);
+        luaL_serializer_copy_options(&user_cfg, cfg);
         luaL_serializer_parse_options(l, &user_cfg);
         lua_pop(l, 1);
         json_append_data(l, &user_cfg, 0, &encode_buf);
+        luaL_serializer_options_delete(&user_cfg);
```
Totktonada, line 448: Sorry for the late comment, but JSON encoding (as well as decoding) can raise a Lua error. It seems, in this case we can get a memory leak.

As a simple fixup I can propose to allocate this serializer object as a userdata Lua object and keep it on the Lua stack.

src/lua/serializer.c
```diff
@@ -88,16 +89,139 @@ luaL_serializer_create(struct luaL_serializer *cfg)
...
+	memset(cfg, '#', offsetof(struct luaL_serializer, end_of_options));
+}
+
+/**
+ * Destroy the serializer object, freeing all allocated memory.
+ * @param L Lua stack.
+ * @retval 0.
+ */
+static int
+luaL_destroy_serializer(struct lua_State *L)
```
Totktonada, line 182: Nit: All the other functions start from `luaL_serializer_`.

grafin: Fixed.

third_party/lua-cjson/lua_cjson.c
grafin: Thanks for pointing that out! I've fixed it by creating the `user_cfg` on Lua stack with `__gc` method, and poping it right away, so that the GC will clean everything up even in case encoding/decoding fails. I've also added a test for that, that will fail under ASAN test.

Totktonada: We should keep it on the stack till the end of the function execution. Otherwise it may be collected on some Lua-land allocation (for example, `lua_newtable` or `lua_pushstring`). Let's push it on the 1st Lua stack slot (change with the value to encode using `lua_insert(L, -2);`).

grafin: I've fixed it) there were some problems with it, because some C functions in the trace were calling `lua_` functions and were counting the stack positions from the bottom. I've also added some test that will fail under ASAN with the previous approach.

third_party/lua-cjson/lua_cjson.c
```diff
@@ -423,10 +439,22 @@ static int json_encode(lua_State *l) {
...
-        struct luaL_serializer user_cfg = *cfg;
-        luaL_serializer_parse_options(l, &user_cfg);
+        /*
+         * user_cfg is per-call local version of serializer instance
+         * options: it is used if a user passes custom options to
+         * :encode() method within a separate argument. In this case
+         * it is required to avoid modifying options of the instance.
+         * Life span of user_cfg is restricted by the scope of
+         * :encode() so it is ok to push it onto the Lua stack and
+         * pop it right away. This way we avoid memory leak, as it will
```
Totktonada, line 449: > ok to push it onto the Lua stack and pop it right away.

We don't do it now (and it is not right).

third_party/lua-cjson/lua_cjson.c
Totktonada: Now it looks valid, thanks!

third_party/lua-cjson/lua_cjson.c
grafin: Thanks, fixed (in both places).

PR #10589, opened 2024-09-18
iproto: make iproto resistant to misusage

### LINKED ISSUE #10155: Tarantool is extremely easy to crash by IPROTO misusage
By mistake, I've sent `IPROTO_OK` after Tarantool sent me the snapshot via `IPROTO_FETCH_SNAPSHOT` request (Tarantool doesn't expect it) and then I sent `IPROTO_SUBSCRIBE`. Such sequence of requests suddenly crashed Tarantool. So I wrote a simple test with misusages of anonymous replication protocol. This protocol can be used not only by Tarantool, but by users manually (CDC, for example), so such misusage is possible in real world. Four cases have discovered five different assertion fails and three different crashes.

<details>
<summary> Test itself </summary>

Should be run in parallel (for example, `./test-run.sh $(yes iproto_crash | head -n 100)`)

```lua
local server = require('luatest.server')
local t = require('luatest')
local msgpack = require('msgpack')
local uri = require('uri')
local lsocket = require('socket')
 
local type = box.iproto.type
local key = box.iproto.key
 
local timeout = 60
 
local function socket_connect(server)
    local u = uri.parse(server.net_box_uri)
    local s = lsocket.tcp_connect(u.host, u.service)
    t.assert_not_equals(s, nil)
    -- Skip the greeting.
    s:read(box.iproto.GREETING_SIZE, timeout)
    return s
end
 
local function socket_write(s, header, body)
    return s:write(box.iproto.encode_packet(header, body))
end
 
local function socket_read(s)
    local size_mp = s:read(5, timeout)
    t.assert_equals(#size_mp, 5)
    local size = msgpack.decode(size_mp)
    local response = s:read(size, timeout)
    t.assert_equals(#response, size)
    return box.iproto.decode_packet(size_mp .. response)
end
 
local function socket_restart(g)
    g.s:close()
    g.s = socket_connect(g.server)
end
 
local function encode_map(map)
    return msgpack.object(setmetatable(map, {__serialize = 'map'}))
end
 
local function setmap(map)
    return setmetatable(map, {__serialize = 'map'})
end
 
local function write_fetch_snapshot(s)
    local header = {
        [key.REQUEST_TYPE] = box.iproto.type.FETCH_SNAPSHOT,
        [key.SYNC] = 1,
    }
    local body = setmap({})
    return socket_write(s, header, body)
end
 
local function write_subscribe(s, uuid, replicaset_uuid, is_anon)
    local header = {
        [key.REQUEST_TYPE] = box.iproto.type.SUBSCRIBE,
        [key.SYNC] = 1,
    }
    local body = {
        [key.REPLICASET_UUID] = replicaset_uuid,
        [key.INSTANCE_UUID] = uuid,
        [key.VCLOCK] = encode_map({}),
        [key.REPLICA_ANON] = is_anon,
    }
    return socket_write(s, header, body)
end
 
local function write_ok(s, body)
    local header = {
        [key.REQUEST_TYPE] = box.iproto.type.OK,
        [key.SYNC] = 1,
    }
    if body == nil then
        body = {}
    end
    return socket_write(s, header, body)
end
 
local function parse_data_stream(g)
    local h, b = socket_read(g.s)
    local request_type = h[key.REQUEST_TYPE]
    while request_type == type.INSERT or request_type == type.RAFT_PROMOTE do
        h, b = socket_read(g.s)
        request_type = h[key.REQUEST_TYPE]
    end
    return h, b
end
 
local g = t.group()
 
g.before_each(function(g)
    g.server = server:new()
    g.server:start()
    g.server:exec(function()
        for i = 1, 100 do
            box.space.test:replace{i}
        end
    end)
    g.s = socket_connect(g.server)
end)
 
g.after_each(function(g)
    g.s:close()
    g.server:stop()
end)
 
-- The case sends a dummy OK before subscribe
g.test_iproto_crash_on_subscribe = function(g)
    local uuid = require('uuid').str()
    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
 
    -- Write OK but do not read response
    write_ok(g.s)
 
    write_subscribe(g.s, uuid, replicaset_uuid, true)
    socket_read(g.s)
end
 
-- The case sends spams dummy OKs before subscribe
g.test_iproto_crash_on_subscribe_spam_ok = function(g)
    local uuid = require('uuid').str()
    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
 
    -- Write OK but do not read response
    for i = 1, 100 do
        write_ok(g.s)
    end
 
    write_subscribe(g.s, uuid, replicaset_uuid, true)
    socket_read(g.s)
end
 
-- The case simulates a situtaion where the user of anonymous replication
-- simply sent IPROTO_OK back after FETCH_SNAPSHOT by mistake (did not know
-- that Tarantool doesn't expect a reply on FETCH_SNAPSHOT)
g.test_iproto_crash_fetch_snapshot_subscribe = function(g)
    local uuid = require('uuid').str()
    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
 
    write_fetch_snapshot(g.s)
    -- Read OK response
    socket_read(g.s)
    -- Read all data sent in response
    local h, b = parse_data_stream(g)
    t.assert_equals(h.REQUEST_TYPE, type.OK)
 
    write_ok(g.s)
 
    write_subscribe(g.s, uuid, replicaset_uuid, true)
    socket_read(g.s)
end
 
-- The same as above, but additionally forgot to pass is_anon option to
-- subscribe
g.test_iproto_crash_fetch_snapshot_subscribe_non_anon = function(g)
    local uuid = require('uuid').str()
    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
 
    write_fetch_snapshot(g.s)
    -- Read OK response
    socket_read(g.s)
    -- Read all data sent in response
    local h, b = parse_data_stream(g)
    t.assert_equals(h.REQUEST_TYPE, type.OK)
 
    write_ok(g.s)
 
    write_subscribe(g.s, uuid, replicaset_uuid)
    socket_read(g.s)
end
```

</details>

<details>
<summary> Backtrace 1 </summary>

```sh
[013] #   #1  0x102550324 in crash_collect+200
[013] #   #2  0x1025501e0 in crash_signal_cb+76
[013] #   #3  0x18ca13584 in _sigtramp+56
[013] #   #4  0x102575b74 in iostream_writev+160
[013] #   #5  0x102575050 in coio_flush+64
[013] #   #6  0x102574d14 in coio_writev_timeout+360
[013] #   #7  0x102252008 in coio_writev(iostream*, iovec*, int, unsigned long)+56
[013] #   #8  0x102251f5c in coio_write_xrow+188
[013] #   #9  0x1023eab6c in relay_send(relay*, xrow_header*)+292
[013] #   #10 0x1023ec834 in relay_send_tx(relay*)+384
[013] #   #11 0x1023e8314 in relay_process_row(xstream*, xrow_header*)+788
[013] #   #12 0x1023d714c in xstream_write+40
[013] #   #13 0x1023d5bdc in recover_xlog(recovery*, xstream*, vclock const*)+812
[013] #   #14 0x1023d5300 in recover_remaining_wals(recovery*, xstream*, vclock const*, bool)+568
[013] #   #15 0x1023eb630 in relay_process_wal_event(wal_watcher*, unsigned int)+92
[013] #   #16 0x1023fec5c in wal_watcher_notify_perform+68
[013] #   #17 0x102568808 in cmsg_deliver+52
[013] #   #18 0x102569700 in cbus_process+192
[013] #   #19 0x102568f90 in cbus_pair+272
[013] #   #20 0x1023febc4 in wal_set_watcher+308
[013] #   #21 0x1023e91a4 in relay_subscribe_f(char*)+708
```

</details>

<details>
<summary> Backtrace 2 </summary>

```sh
[013] #   #1  0x100bc8324 in crash_collect+200
[013] #   #2  0x100bc81e0 in crash_signal_cb+76
[013] #   #3  0x18ca13584 in _sigtramp+56
[013] #   #4  0x100bedb74 in iostream_writev+160
[013] #   #5  0x100bed050 in coio_flush+64
[013] #   #6  0x100becd14 in coio_writev_timeout+360
[013] #   #7  0x1008ca008 in coio_writev(iostream*, iovec*, int, unsigned long)+56
[013] #   #8  0x1008c9f5c in coio_write_xrow+188
[013] #   #9  0x100a2299c in box_process_subscribe(iostream*, xrow_header const*)+4276
[013] #   #10 0x1008be688 in tx_process_replication(cmsg*)+684
[013] #   #11 0x100be0808 in cmsg_deliver+52
[013] #   #12 0x100be2b60 in fiber_pool_f+832
[013] #   #13 0x1008b2590 in fiber_cxx_invoke(int (*)(char*), char*)+32
[013] #   #14 0x100bd3e70 in fiber_loop+268
[013] #   #15 0x100fd90b0 in coro_startup+8
```

</details>

<details>
<summary> Backtrace 3 </summary>

```sh
[2024-06-21 14:28:30.394620] #   #1  0x100b14324 in crash_collect+200
[2024-06-21 14:28:30.394838] #   #2  0x100b141e0 in crash_signal_cb+76
[2024-06-21 14:28:30.395050] #   #3  0x18ca13584 in _sigtramp+56
[2024-06-21 14:28:30.395266] #   #4  0x100b381dc in iostream_read+160
[2024-06-21 14:28:30.395494] #   #5  0x100b37e88 in coio_read_ahead_timeout+364
[2024-06-21 14:28:30.395712] #   #6  0x100b38564 in coio_readn_ahead_timeout+56
[2024-06-21 14:28:30.395928] #   #7  0x100815da4 in coio_breadn_timeout(iostream*, ibuf*, unsigned long, double)+156
[2024-06-21 14:28:30.396120] #   #8  0x1008159b0 in coio_read_xrow_timeout_xc+104
[2024-06-21 14:28:30.396332] #   #9  0x1009ac6d4 in relay_reader_f(char*)+624
[2024-06-21 14:28:30.396542] #   #10 0x1007fe590 in fiber_cxx_invoke(int (*)(char*), char*)+32
[2024-06-21 14:28:30.398392] #   #11 0x100b1fe70 in fiber_loop+268
[2024-06-21 14:28:30.399069] #   #12 0x100f250b0 in coro_startup+8
```

</details>

<details>
<summary> A bunch of failed assertions </summary>

Reproduced on Linux x86-64:
```sh
./third_party/libev/ev_epoll.c:134: epoll_modify: Assertion `("libev: I/O watcher with invalid fd found in epoll_ctl", errno != EBADF && errno != ELOOP && errno != EINVAL)' failed.
```

Reproduced on MacOS 14 (Apple M2 Pro):
```sh
Assertion failed: ((io)->owner == NULL), function iostream_read, file iostream.h, line 198
```

```sh
Assertion failed: ((io)->owner == cord()), function iostream_read, file iostream.h, line 200
```

```sh
Assertion failed: ((io)->owner == NULL), function iostream_writev, file iostream.h, line 226.
```

Reproduced on both platforms:
```sh
Assertion failed: (! ev_is_active(&con->input)), function net_end_subscribe, file iproto.cc, line 3279.
```

</details>

Tarantool version: Tarantool 3.1.0-entrypoint-362-gd2240bf7c (current master)

Astronomax:
Fixed an issue when the tarantool could be easily crashed using iproto incorrectly, not according to the protocol.

The problem seems to be unsynchronized multi-threaded communication with the socket. Hence the large number of different backtraces and assertions. To reproduce this, instead of `IPROTO_OK` you can use anything whose route looks like ->main->iproto (looks like all messages are going this way now), for example, I checked this with `IPROTO_PING`. A bad execution scenario is shown in the picture:
iproto_misusage_championship
More specifically, the problem is that when `tx_process_replication` is executed in `main`, it does not know anything about the c-messages that are queued at the "net_x" endpoint in `iproto`. In `iproto_enqueue_batch` we call `ev_io_stop` on both `input`/`output` watchers, which prevents the `EV_READ`/`EV_WRITE` events from firing (and also clears pending events), but this does not help against `ev_feed_event`, which is called in the cmsg callback queued in the `iproto` thread.

Closes #10155

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh_10155_make_iproto_resistant_to_misusage.md
+++ b/changelogs/unreleased/gh_10155_make_iproto_resistant_to_misusage.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when a server could crash if a client sent an IPROTO replication
+  request without waiting for pending requests to complete (gh-10155).
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -920,6 +920,8 @@ struct iproto_connection
 	struct cmsg cancel_msg;
 	/** Set if connection is accepted in TX. */
 	bool is_established;
+	/** Number of iproto requests in flight. */
+	size_t request_count;
 };
 
 /** Returns a string suitable for logging. */
@@ -984,6 +986,8 @@ iproto_check_msg_max(struct iproto_thread *iproto_thread)
 static inline void
 iproto_msg_delete(struct iproto_msg *msg)
 {
+	assert(msg->connection->request_count > 0);
+	msg->connection->request_count--;
 	struct iproto_thread *iproto_thread = msg->connection->iproto_thread;
 	mempool_free(&msg->connection->iproto_thread->iproto_msg_pool, msg);
 	iproto_resume(iproto_thread);
@@ -1009,6 +1013,7 @@ iproto_msg_new(struct iproto_connection *con)
 	msg->stream = NULL;
 	msg->fiber = NULL;
 	rmean_collect(con->iproto_thread->rmean, IPROTO_REQUESTS, 1);
+	con->request_count++;
 	return msg;
 }
 
@@ -1019,6 +1024,7 @@ static inline void
 iproto_connection_feed_input(struct iproto_connection *con)
 {
 	assert(con->state == IPROTO_CONNECTION_ALIVE);
+	assert(!con->is_in_replication);
 	if (!ev_is_active(&con->input) && rlist_empty(&con->in_stop_list))
 		ev_feed_event(con->loop, &con->input, EV_CUSTOM);
 }
@@ -1030,6 +1036,7 @@ static inline void
 iproto_connection_feed_output(struct iproto_connection *con)
 {
 	assert(con->state == IPROTO_CONNECTION_ALIVE);
+	assert(!con->is_in_replication);
 	if (!ev_is_active(&con->output))
 		ev_feed_event(con->loop, &con->output, EV_CUSTOM);
 }
@@ -1283,6 +1290,13 @@ iproto_connection_input_buffer(struct iproto_connection *con)
 	return new_ibuf;
 }
 
+static inline bool
+iproto_is_flushed(struct iproto_connection *con)
+{
+	return con->wpos.obuf == con->wend.obuf &&
+		   con->wpos.svp.used == con->wend.svp.used;
+}
+
 /**
  * Check if message belongs to stream (stream_id != 0), and if it
  * is so create new stream or get stream from connection streams
@@ -1479,6 +1493,7 @@ iproto_connection_on_input(ev_loop *loop, struct ev_io *watcher,
 	assert(con->state == IPROTO_CONNECTION_ALIVE);
 	assert(rlist_empty(&con->in_stop_list));
 	assert(loop == con->loop);
+	assert(!con->is_in_replication);
 	/*
 	 * Throttle if there are too many pending requests,
 	 * otherwise we might deplete the fiber pool in tx
@@ -1562,6 +1577,11 @@ iproto_flush(struct iproto_connection *con)
 		return 0;
 	}
 	assert(begin->used < end->used);
+
+	ERROR_INJECT(ERRINJ_IPROTO_FLUSH_DELAY, {
+		return IOSTREAM_WANT_WRITE;
+	});
+
 	struct iovec iov[SMALL_OBUF_IOV_MAX+1];
 	struct iovec *src = obuf->iov;
 	int iovcnt = end->pos - begin->pos + 1;
@@ -1611,6 +1631,7 @@ iproto_connection_on_output(ev_loop *loop, struct ev_io *watcher,
 {
 	struct iproto_connection *con = (struct iproto_connection *) watcher->data;
 	assert(con->state == IPROTO_CONNECTION_ALIVE);
+	assert(!con->is_in_replication);
 	int rc;
 	while ((rc = iproto_flush(con)) <= 0) {
 		if (rc != 0) {
@@ -1675,6 +1696,7 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	con->tx.is_push_pending = false;
 	con->tx.is_push_sent = false;
 	rmean_collect(iproto_thread->rmean, IPROTO_CONNECTIONS, 1);
+	con->request_count = 0;
 	return con;
 }
 
@@ -1787,9 +1809,11 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
 {
 	uint64_t stream_id;
 	uint32_t type;
+	bool is_replication_request;
 	bool request_is_not_for_stream;
 	bool request_is_only_for_stream;
-	struct iproto_thread *iproto_thread = msg->connection->iproto_thread;
+	struct iproto_connection *con = msg->connection;
+	struct iproto_thread *iproto_thread = con->iproto_thread;
 	mh_i32_t *handlers = iproto_thread->req_handlers;
 	mh_int_t handler;
 	struct cmsg_hop *route;
@@ -1819,14 +1843,27 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
 		goto error;
 	}
 
-	msg->connection->is_in_replication = type == IPROTO_JOIN ||
-					     type == IPROTO_FETCH_SNAPSHOT ||
-					     type == IPROTO_REGISTER ||
-					     type == IPROTO_SUBSCRIBE;
+	is_replication_request = msg->header.type == IPROTO_JOIN ||
+				 msg->header.type == IPROTO_FETCH_SNAPSHOT ||
+				 msg->header.type == IPROTO_REGISTER ||
+				 msg->header.type == IPROTO_SUBSCRIBE;
+
+	if (is_replication_request) {
+		/**
+		 * Before processing a replication request, ensure that all
+		 * writing requests have been completely flushed to obuf.
+		 */
+		if (con->request_count > 1 || !iproto_is_flushed(con)) {
+			diag_set(ClientError, ER_PROTOCOL, "Can't process "
+				 "join/subscribe while there are pending requests");
+			goto error;
+		}
+		con->is_in_replication = true;
+	}
 
 	handler = mh_i32_find(handlers, type, NULL);
 	if (handler != mh_end(handlers)) {
-		assert(!msg->connection->is_in_replication);
+		assert(!con->is_in_replication);
 		cmsg_init(&msg->base, iproto_thread->override_route);
 		return;
 	}
--- a/src/lib/core/errinj.h
+++ b/src/lib/core/errinj.h
@@ -103,6 +103,7 @@ struct errinj {
 	_(ERRINJ_IPROTO_DISABLE_ID, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_DISABLE_WATCH, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_FLIP_FEATURE, ERRINJ_INT, {.iparam = -1}) \
+	_(ERRINJ_IPROTO_FLUSH_DELAY, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_PROCESS_REPLICATION_DELAY, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_SET_VERSION, ERRINJ_INT, {.iparam = -1}) \
 	_(ERRINJ_IPROTO_TX_DELAY, ERRINJ_BOOL, {.bparam = false}) \
--- a/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua
+++ b/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua
@@ -1,0 +1,241 @@
+local lsocket = require('socket')
+local msgpack = require('msgpack')
+local server = require('luatest.server')
+local t = require('luatest')
+local uri = require('uri')
+
+local key = box.iproto.key
+local type = box.iproto.type
+
+local g = t.group('make-iproto-resistant-to-misusage')
+--
+-- gh-10155: make iproto resistant to misusage.
+--
+local wait_timeout = 60
+
+local function iproto_error_type(error_type)
+    return bit.bor(box.iproto.type.TYPE_ERROR, error_type)
+end
+
+local function socket_connect(server)
+    local u = uri.parse(server.net_box_uri)
+    local s = lsocket.tcp_connect(u.host, u.service)
+    t.assert_not_equals(s, nil)
+    -- Skip the greeting
+    s:read(box.iproto.GREETING_SIZE, wait_timeout)
+    return s
+end
+
+local function encode_map(map)
+    return msgpack.object(setmetatable(map, {__serialize = 'map'}))
+end
+
+local function setmap(map)
+    return setmetatable(map, {__serialize = 'map'})
+end
+
+local function socket_write(s, header, body)
+    return s:write(box.iproto.encode_packet(header, body))
+end
+
+local function write_ok(s, body)
+    local header = {
+        [key.REQUEST_TYPE] = box.iproto.type.OK,
+        [key.SYNC] = 1,
+    }
+    if body == nil then
+        body = {}
+    end
+    return socket_write(s, header, body)
+end
+
+local function write_eval(s, expr)
+    local key = box.iproto.key
+    local header = {
+        [key.REQUEST_TYPE] = box.iproto.type.EVAL,
+        [key.SYNC] = 1,
+    }
+    local body = {
+        [key.EXPR] = expr,
+        [key.TUPLE] = {},
+    }
+    return s:write(box.iproto.encode_packet(header, body))
+end
+
+local function write_fetch_snapshot(s)
+    local header = {
+        [key.REQUEST_TYPE] = box.iproto.type.FETCH_SNAPSHOT,
+        [key.SYNC] = 1,
+    }
+    local body = setmap({})
+    return socket_write(s, header, body)
+end
+
+local function write_subscribe(s, uuid, replicaset_uuid, is_anon)
+    local header = {
+        [key.REQUEST_TYPE] = box.iproto.type.SUBSCRIBE,
+        [key.SYNC] = 1,
+    }
+    local body = {
+        [key.REPLICASET_UUID] = replicaset_uuid,
+        [key.INSTANCE_UUID] = uuid,
+        [key.VCLOCK] = encode_map({}),
+        [key.REPLICA_ANON] = is_anon,
+    }
+    return socket_write(s, header, body)
+end
+
+local function socket_read(s)
+    local size_mp = s:read(5, wait_timeout)
+    t.assert_equals(#size_mp, 5)
+    local size = msgpack.decode(size_mp)
+    local response = s:read(size, wait_timeout)
+    t.assert_equals(#response, size)
+    return box.iproto.decode_packet(size_mp .. response)
+end
+
+-- Read all data sent in FETCH_SNAPSHOT response
+local function read_snapshot(s)
+    local h, _ = socket_read(s)
+    local request_type = h[key.REQUEST_TYPE]
+    while request_type == type.INSERT or request_type == type.RAFT_PROMOTE do
+        h, _ = socket_read(s)
+        request_type = h[key.REQUEST_TYPE]
+    end
+    t.assert_equals(h.REQUEST_TYPE, type.OK)
+end
+
+g.before_each(function(g)
+    g.server = server:new()
+    g.server:start()
+    g.server:exec(function()
+        box.schema.space.create('test')
+        box.space.test:create_index('pk')
+        for i = 1, 100 do
+            box.space.test:replace{i}
+        end
+    end)
+    g.s = socket_connect(g.server)
+end)
+
+g.after_each(function(g)
+    g.s:close()
+    g.server:stop()
+end)
+
+-- The case sends a dummy OK before subscribe
+g.test_iproto_crash_on_subscribe = function(g)
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    -- Write OK but do not read response before writing subscribe
+    write_ok(g.s)
+    write_subscribe(g.s, uuid, replicaset_uuid, true)
+    local request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert_equals(request_type,
+        iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+end
+
+-- The test is designed to check the behavior when, at the time of receiving a
+-- replication request, the responses to all previous requests are written to
+-- obuf but not flushed to the socket
+g.test_iproto_crash_on_subscribe_flush_delay = function(g)
+    t.tarantool.skip_if_not_debug()
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    -- You can't use lustest.server:exec here because it will block
+    write_eval(g.s, "box.error.injection.set(" ..
+        "'ERRINJ_IPROTO_FLUSH_DELAY', true)")
+    -- Several requests in a row to pretend to be a non-blocking client and
+    -- provoke iproto_connection_feed_input, so that subscribe is immediately
+    -- read
+    for _ = 1, 3 do
+        write_ok(g.s)
+    end
+    -- A short pause so that the packages do not end up in one batch and all
+    -- responses have time to be written in obuf
+    require('fiber').sleep(0.1)
+    write_subscribe(g.s, uuid, replicaset_uuid, true)
+    -- A short pause so that the tx thread has time to write the response to
+    -- the subscribe request to the socket
+    require('fiber').sleep(0.1)
+    g.server:exec(function()
+        box.error.injection.set('ERRINJ_IPROTO_FLUSH_DELAY', false)
+    end)
+    -- Read IPROTO_EVAL response
+    local request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK)
+    for _ = 1, 3 do
+        request_type = socket_read(g.s)[key.REQUEST_TYPE]
+        t.assert_equals(request_type,
+            iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    end
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+end
+
+-- The case sends dummy OKs before subscribe
+g.test_iproto_crash_on_subscribe_spam_ok = function(g)
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    -- Write OKs but do not read responses before writing subscribe
+    for _ = 1, 100 do
+        write_ok(g.s)
+    end
+    write_subscribe(g.s, uuid, replicaset_uuid, true)
+    local request_type
+    for _ = 1, 100 do
+        request_type = socket_read(g.s)[key.REQUEST_TYPE]
+        t.assert_equals(request_type,
+            iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    end
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+end
+
+-- The case simulates a situation where the user of anonymous replication
+-- simply sent IPROTO_OK back after FETCH_SNAPSHOT by mistake (did not know
+-- that Tarantool doesn't expect a reply on FETCH_SNAPSHOT)
+g.test_iproto_crash_fetch_snapshot_subscribe = function(g)
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    write_fetch_snapshot(g.s)
+    local request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert_equals(request_type, type.OK)
+    read_snapshot(g.s)
+    -- Write OK but do not read response before writing subscribe
+    write_ok(g.s)
+    write_subscribe(g.s, uuid, replicaset_uuid, true)
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert_equals(request_type,
+        iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+end
+
+-- The same as above, but additionally forgot to pass is_anon option to
+-- subscribe
+g.test_iproto_crash_fetch_snapshot_subscribe_not_anon = function(g)
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    write_fetch_snapshot(g.s)
+    local request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert_equals(request_type, type.OK)
+    read_snapshot(g.s)
+    -- Write OK but do not read response before writing subscribe
+    write_ok(g.s)
+    -- Subscribe as not anon replica
+    write_subscribe(g.s, uuid, replicaset_uuid)
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert_equals(request_type,
+        iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type ==
+        iproto_error_type(box.error.TOO_EARLY_SUBSCRIBE) or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+end


changelogs/unreleased/gh_10155_make_iproto_resistant_to_misusage.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/iproto
+
+* Fixed an issue when the tarantool could be easily crashed using iproto
+  incorrectly, not according to the protocol (gh-10155).
```
xuniq, line 4:
```suggestion
* Fixed an issue when Tarantool could crash if iproto was used incorrectly, not according to the protocol (gh-10155).
```

xuniq (approved):

src/box/iproto.cc
```diff
@@ -1013,24 +1013,25 @@ iproto_msg_new(struct iproto_connection *con)
...
 /**
- * Signal output unless it's blocked on I/O.
+ * Signal output unless it's blocked on I/O or is in replication.
  */
 static inline void
 iproto_connection_feed_output(struct iproto_connection *con)
 {
 	assert(con->state == IPROTO_CONNECTION_ALIVE);
-	if (!ev_is_active(&con->output))
+	if (!ev_is_active(&con->output) && !con->is_in_replication)
```
sergepetrenko, line 1034: Does it mean that after a replication request is enqueued in the iproto thread all the pending message responses will be lost?

Actually I think this is fine, but I still want to bring this to attention of other reviewers.

Why I think this is fine: a user should either wait for all the pending request responses prior to issuing a replication request in a single connection or have a separate connection for all the requests he wants to perform alongside the replication stream.

sergepetrenko (approved):
Thanks for the fix! I like that the solution turned out even simpler than the one we've discussed f2f previously.

Please solicit another review from @Gerold103.

src/box/iproto.cc
locker: Maybe, better be explicit about this and raise an error if a replication request is received while there are active requests in tx?

sergepetrenko: Sounds good to me. Although it seems the existing `tx.requests_in_progress` metric wouldn't be enough for this.
We'd have to count already decoded but not yet processed requests on the iproto side.

changelogs/unreleased/gh_10155_make_iproto_resistant_to_misusage.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/iproto
+
+* Fixed an issue when Tarantool could easily crash if iproto was used
+  incorrectly, not according to the protocol (gh-10155).
```
locker, line 4: - We never used "iproto" section. Better use "core" for this bug.
- I'd be more specific about what happens here.

```suggestion
## bugfix/core

* Fixed a bug when a server could crash if a client sent an IPROTO replication
  request without waiting for pending requests to complete (gh-10155).
```

src/box/iproto.cc
```diff
@@ -3221,11 +3224,10 @@ net_send_msg(struct cmsg *m)
...
 	con->wend = msg->wpos;
 
-	if (con->state == IPROTO_CONNECTION_ALIVE) {
+	if (con->state == IPROTO_CONNECTION_ALIVE)
 		iproto_connection_feed_output(con);
-	} else if (iproto_connection_is_idle(con)) {
+	else if (iproto_connection_is_idle(con))
 		iproto_connection_close(con);
-	}
 	iproto_msg_delete(msg);
```
locker, line 3231: Please avoid unrelated changes. They make the patch difficult to review.

src/box/iproto.cc
Astronomax: > Does it mean that after a replication request is enqueued in the iproto thread all the pending message responses will be lost?

No, it shouldn't. What do you call "pending message responses"? If you mean responses that were written to `obuf *iproto_connection::obuf`, but have not yet been written to the socket `iostream iproto_connection::io`, then they will not be lost anywhere, and will be written after the processing of the replication-request is completed.
If you mean requests (`iproto_msg`'s from "tx" thread) that are queued at `cbus_endpoint` "net{iproto thread->id}", then everything is fine with them too. The corresponding data has already been written to `obuf` in the "tx" thread. They will be processed in the "iproto" thread, they simply will not cause data to be written from `obuf` to the socket (`iproto_connection_feed_output` won't do anything). But this data will be written to the socket later after the processing of the replication-request is completed.

Astronomax: In an good way, we should either cancel all pending requests **that the user has not yet read** or send them all without loss. But we will not be able to cancel all pending requests that the user has not yet read, because `obuf *iproto_connection:::obuf` is flushed to the socket asynchronously. That is, at the moment when we parse the replication request in "iproto" thread, some random number of responses could be written to the socket, and the user does not know how much he needs to subtract before starting to read the response to the replication request. Moreover, it may turn out that the last request that we tried to write to the socket was partially written. And if we right now become the owner of the socket in the “tx” thread right now and write a response to a replication request to it, we will get invalid msgpuck. Therefore, in my opinion, the best solution would be to either force it to wait for all previous requests to be written to the socket, or make it the user's responsibility (the user is required to wait for responses to all previous requests before sending a replication request).

sergepetrenko: Ok, I see, thanks for the explanation.

>  If you mean responses that were written to obuf *iproto_connection::obuf, but have not yet been written to the socket iostream iproto_connection::io

Yes, this is what I meant. Given your explanation I still think the responces are "lost" in a sense that a client first receives a responce to his replication request and only then receives a responce to his previous request (BTW, I'm not sure it's possible to continue normal iproto operation after replication is being processed).

I think we should return an error then if we receive a replication request before we have sent responces to all the previous user's requests.

Another option would be to pause replication request processing (in iproto thread) until all responces to previous requests are written.

@locker, which option do you prefer?

locker: > I think we should return an error then if we receive a replication request before we have sent responces to all the previous user's requests.
> 
> Another option would be to pause replication request processing (in iproto thread) until all responces to previous requests are written.

I prefer the first option (rejecting replication request) because it should be easier to implement AFAICS. Nobody should send a replication request before processing all pending responses.

Astronomax: I have implemented it, please have a look.

changelogs/unreleased/gh_10155_make_iproto_resistant_to_misusage.md
Astronomax: Fixed.

src/box/iproto.cc
Astronomax: Fixed.

src/box/iproto.cc
```diff
@@ -1283,6 +1287,35 @@ iproto_connection_input_buffer(struct iproto_connection *con)
...
+		if (begin->used == obuf_end.used) {
+			con->wpos.obuf = con->wend.obuf;
+			obuf_svp_reset(begin);
+		}
+	}
+	return con->wpos.obuf;
+}
+
+static inline bool
+iproto_flushed(struct iproto_connection *con)
```
locker, line 1309: Do we really need this refactoring? Can't we just check that both output buffers are empty?

Astronomax:
diff after last review request
```diff
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index c53f9455e6..b94967db89 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -1413,8 +1413,8 @@ iproto_enqueue_batch(struct iproto_connection *con, struct ibuf *in)
                /**
                 * We want to ensure that the replication request will not be
                 * processed until all writing requests have been completely
-                * flushed to obuf. It so happened that all requests that come here,
-                * except for some replication ones, write something to obuf.
+                * flushed to obuf. It so happened that all replication requests
+                * that come here write something to obuf.
                 */
                if (!con->is_in_replication)
                        con->active_writing_requests++;
@@ -1847,7 +1847,7 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
        if (is_replication_request) {
                /**
                 * Before processing a replication request, ensure that all
-                * writing requests have been completely flushed to obuf.
+                * write requests have been completely flushed to obuf.
                 */
                if (con->active_writing_requests > 0 || !iproto_flushed(con)) {
                        diag_set(ClientError, ER_UNABLE_TO_PROCESS_REPLICATION);
```

src/box/iproto.cc
```diff
@@ -1811,23 +1865,20 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
 		 type == IPROTO_ROLLBACK);
 
 	if (stream_id != 0 && request_is_not_for_stream) {
+		assert(!con->is_in_replication);
```
locker, line 1869: Why did you add these assertions? Why did you move the check for replication request above? This assertion can actually fail if someone tries to set `stream_id` for a replication request.

Astronomax:
diff
```diff
diff --git a/src/box/errcode.h b/src/box/errcode.h
index 719bf6a06f..3a65a3f075 100644
--- a/src/box/errcode.h
+++ b/src/box/errcode.h
@@ -434,7 +434,6 @@ struct errcode_record {
 	_(ER_WAL_QUEUE_FULL, 287,		"The WAL queue is full") \
 	_(ER_INVALID_VCLOCK, 288,		"Invalid vclock", "value", STRING) \
 	_(ER_SYNC_QUEUE_FULL, 289,		"The synchronous transaction queue is full") \
-	_(ER_UNABLE_TO_PROCESS_REPLICATION, 290, "Replication request couldn't be processed, responses to previous requests were not read") \
 	TEST_ERROR_CODES(_) /** This one should be last. */
 
 /*
diff --git a/src/box/iproto.cc b/src/box/iproto.cc
index a6ff5381ad..6995de383b 100644
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -845,8 +845,6 @@ struct iproto_connection
 	 * This field is accesable only from iproto thread.
 	 */
 	struct mh_i64ptr_t *streams;
-	/** Count of active requests that write something to obuf (gh-10155). */
-	size_t active_writing_requests;
 	/**
 	 * Kharon is used to implement box.session.push().
 	 * When a new push is ready, tx uses kharon to notify
@@ -1287,33 +1285,16 @@ iproto_connection_input_buffer(struct iproto_connection *con)
 	return new_ibuf;
 }
 
-static struct obuf *
-iproto_connection_output_buffer(struct iproto_connection *con)
-{
-	struct obuf_svp obuf_end = obuf_create_svp(con->wpos.obuf);
-	struct obuf_svp *begin = &con->wpos.svp;
-	if (con->wend.obuf != con->wpos.obuf) {
-		/*
-		 * Flush the current buffer before
-		 * advancing to the next one.
-		 */
-		if (begin->used == obuf_end.used) {
-			con->wpos.obuf = con->wend.obuf;
-			obuf_svp_reset(begin);
-		}
-	}
-	return con->wpos.obuf;
-}
-
 static inline bool
 iproto_flushed(struct iproto_connection *con)
 {
-	struct obuf *obuf = iproto_connection_output_buffer(con);
+	struct obuf *obuf = con->wpos.obuf;
 	struct obuf_svp *begin = &con->wpos.svp;
-	struct obuf_svp end = con->wend.svp;
+	struct obuf_svp *end = &con->wend.svp;
+
 	if (con->wend.obuf != obuf)
-		end = obuf_create_svp(obuf);
-	return (begin->used == end.used);
+		return (begin->used == obuf->used) && (end->used == 0);
+	return (begin->used == end->used);
 }
 
 /**
@@ -1410,15 +1391,6 @@ iproto_enqueue_batch(struct iproto_connection *con, struct ibuf *in)
 		con->input_msg_count[msg->p_ibuf == &con->ibuf[1]]++;
 
 		iproto_msg_prepare(msg, &pos, reqend);
-		/**
-		 * We want to ensure that the replication request will not be
-		 * processed until all writing requests have been completely
-		 * flushed to obuf. It so happened that all requests that come
-		 * here, except for some replication ones, write something to
-		 * obuf.
-		 */
-		if (!con->is_in_replication)
-			con->active_writing_requests++;
 		if (iproto_msg_start_processing_in_stream(msg)) {
 			cpipe_push_input(&con->iproto_thread->tx_pipe, &msg->base);
 			n_requests++;
@@ -1580,25 +1552,46 @@ error:;
 static int
 iproto_flush(struct iproto_connection *con)
 {
-	struct obuf *obuf = iproto_connection_output_buffer(con);
+	struct obuf *obuf = con->wpos.obuf;
+	struct obuf_svp obuf_end = obuf_create_svp(obuf);
 	struct obuf_svp *begin = &con->wpos.svp;
-	struct obuf_svp end = con->wend.svp;
-	if (con->wend.obuf != obuf)
-		end = obuf_create_svp(obuf);
-
-	assert(begin->used <= end.used);
-	if (begin->used == end.used) {
+	struct obuf_svp *end = &con->wend.svp;
+	if (con->wend.obuf != obuf) {
+		/*
+		 * Flush the current buffer before
+		 * advancing to the next one.
+		 */
+		if (begin->used == obuf_end.used) {
+			obuf = con->wpos.obuf = con->wend.obuf;
+			obuf_svp_reset(begin);
+		} else {
+			end = &obuf_end;
+		}
+	}
+	if (begin->used == end->used) {
 		/* Nothing to do. */
 		return 1;
 	}
 	if (!con->can_write) {
 		/* Receiving end was closed. Discard the output. */
-		*begin = end;
+		*begin = *end;
 		return 0;
 	}
+	assert(begin->used < end->used);
+
+	ERROR_INJECT_COUNTDOWN(ERRINJ_IPROTO_FLUSH_DELAY_COUNTDOWN, {
+		struct errinj *e =
+			errinj(ERRINJ_IPROTO_FLUSH_DELAY, ERRINJ_BOOL);
+		e->bparam = true;
+
+	});
+	ERROR_INJECT(ERRINJ_IPROTO_FLUSH_DELAY, {
+		return IOSTREAM_WANT_WRITE;
+	});
+
 	struct iovec iov[SMALL_OBUF_IOV_MAX+1];
 	struct iovec *src = obuf->iov;
-	int iovcnt = end.pos - begin->pos + 1;
+	int iovcnt = end->pos - begin->pos + 1;
 	/*
 	 * iov[i].iov_len may be concurrently modified in tx thread,
 	 * but only for the last position.
@@ -1606,14 +1599,14 @@ iproto_flush(struct iproto_connection *con)
 	memcpy(iov, src + begin->pos, iovcnt * sizeof(struct iovec));
 	sio_add_to_iov(iov, -begin->iov_len);
 	/* *Overwrite* iov_len of the last pos as it may be garbage. */
-	iov[iovcnt - 1].iov_len = end.iov_len - begin->iov_len * (iovcnt == 1);
+	iov[iovcnt-1].iov_len = end->iov_len - begin->iov_len * (iovcnt == 1);
 
 	ssize_t nwr = iostream_writev(&con->io, iov, iovcnt);
 	if (nwr >= 0) {
 		/* Count statistics */
 		rmean_collect(con->iproto_thread->rmean, IPROTO_SENT, nwr);
-		if (begin->used + nwr == end.used) {
-			*begin = end;
+		if (begin->used + nwr == end->used) {
+			*begin = *end;
 			return 0;
 		}
 		size_t offset = 0;
@@ -1622,7 +1615,7 @@ iproto_flush(struct iproto_connection *con)
 		begin->used += nwr;             /* advance write position */
 		begin->iov_len = advance == 0 ? begin->iov_len + offset: offset;
 		begin->pos += advance;
-		assert(begin->pos <= end.pos);
+		assert(begin->pos <= end->pos);
 		return IOSTREAM_WANT_WRITE;
 	} else if (nwr == IOSTREAM_ERROR) {
 		/*
@@ -1633,7 +1626,7 @@ iproto_flush(struct iproto_connection *con)
 		 */
 		diag_log();
 		con->can_write = false;
-		*begin = end;
+		*begin = *end;
 		return 0;
 	}
 	return nwr;
@@ -1675,7 +1668,6 @@ iproto_connection_new(struct iproto_thread *iproto_thread)
 	struct iproto_connection *con = (struct iproto_connection *)
 		xmempool_alloc(&iproto_thread->iproto_connection_pool);
 	con->streams = mh_i64ptr_new();
-	con->active_writing_requests = 0;
 	con->iproto_thread = iproto_thread;
 	con->input.data = con->output.data = con;
 	con->loop = loop();
@@ -1840,23 +1832,6 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
 	type = msg->header.type;
 	stream_id = msg->header.stream_id;
 
-	is_replication_request = msg->header.type == IPROTO_JOIN ||
-				 msg->header.type == IPROTO_FETCH_SNAPSHOT ||
-				 msg->header.type == IPROTO_REGISTER ||
-				 msg->header.type == IPROTO_SUBSCRIBE;
-
-	if (is_replication_request) {
-		/**
-		 * Before processing a replication request, ensure that all
-		 * writing requests have been completely flushed to obuf.
-		 */
-		if (con->active_writing_requests > 0 || !iproto_flushed(con)) {
-			diag_set(ClientError, ER_UNABLE_TO_PROCESS_REPLICATION);
-			goto error;
-		}
-		con->is_in_replication = true;
-	}
-
 	request_is_not_for_stream =
 		((type > IPROTO_TYPE_STAT_MAX &&
 		 type != IPROTO_PING) || type == IPROTO_AUTH);
@@ -1866,17 +1841,34 @@ iproto_msg_prepare(struct iproto_msg *msg, const char **pos, const char *reqend)
 		 type == IPROTO_ROLLBACK);
 
 	if (stream_id != 0 && request_is_not_for_stream) {
-		assert(!con->is_in_replication);
 		diag_set(ClientError, ER_UNABLE_TO_PROCESS_IN_STREAM,
 			 iproto_type_name(type));
 		goto error;
 	} else if (stream_id == 0 && request_is_only_for_stream) {
-		assert(!con->is_in_replication);
 		diag_set(ClientError, ER_UNABLE_TO_PROCESS_OUT_OF_STREAM,
 			 iproto_type_name(type));
 		goto error;
 	}
 
+	is_replication_request = msg->header.type == IPROTO_JOIN ||
+				 msg->header.type == IPROTO_FETCH_SNAPSHOT ||
+				 msg->header.type == IPROTO_REGISTER ||
+				 msg->header.type == IPROTO_SUBSCRIBE;
+
+	if (is_replication_request) {
+		/**
+		 * Before processing a replication request, ensure that all
+		 * writing requests have been completely flushed to obuf.
+		 */
+		if (mempool_count(&iproto_thread->iproto_msg_pool) > 1 ||
+		    !iproto_flushed(con)) {
+			diag_set(ClientError, ER_PROTOCOL, "Can't process "
+				 "join/subscribe while there are pending requests");
+			goto error;
+		}
+		con->is_in_replication = true;
+	}
+
 	handler = mh_i32_find(handlers, type, NULL);
 	if (handler != mh_end(handlers)) {
 		assert(!con->is_in_replication);
@@ -3264,8 +3256,6 @@ net_send_msg(struct cmsg *m)
 	struct iproto_msg *msg = (struct iproto_msg *) m;
 	struct iproto_connection *con = msg->connection;
 
-	if (!con->is_in_replication)
-		con->active_writing_requests--;
 	iproto_msg_finish_processing_in_stream(msg);
 	if (msg->len != 0) {
 		/* Discard request (see iproto_enqueue_batch()). */
diff --git a/src/lib/core/errinj.h b/src/lib/core/errinj.h
index c512bba598..be84283c6a 100644
--- a/src/lib/core/errinj.h
+++ b/src/lib/core/errinj.h
@@ -102,6 +102,8 @@ struct errinj {
 	_(ERRINJ_IPROTO_DISABLE_ID, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_DISABLE_WATCH, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_FLIP_FEATURE, ERRINJ_INT, {.iparam = -1}) \
+	_(ERRINJ_IPROTO_FLUSH_DELAY, ERRINJ_BOOL, {.bparam = false}) \
+	_(ERRINJ_IPROTO_FLUSH_DELAY_COUNTDOWN, ERRINJ_INT, {.iparam = -1}) \
 	_(ERRINJ_IPROTO_SET_VERSION, ERRINJ_INT, {.iparam = -1}) \
 	_(ERRINJ_IPROTO_TX_DELAY, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_IPROTO_WRITE_ERROR_DELAY, ERRINJ_BOOL, {.bparam = false})\
diff --git a/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua b/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua
index b712d98190..f06bc38501 100644
--- a/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua
+++ b/test/box-luatest/gh_10155_make_iproto_resistant_to_misusage_test.lua
@@ -120,7 +120,47 @@ g.test_iproto_crash_on_subscribe = function(g)
         iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
     request_type = socket_read(g.s)[key.REQUEST_TYPE]
     t.assert(request_type == type.OK or request_type ==
-        iproto_error_type(box.error.UNABLE_TO_PROCESS_REPLICATION))
+        iproto_error_type(box.error.PROTOCOL))
+end
+
+-- The test is designed to check the behavior when, at the time of receiving a
+-- replication request, the responses to all previous requests are written to
+-- obuf but not flushed to the socket
+g.test_iproto_crash_on_subscribe_flush_delay = function(g)
+    local uuid = require('uuid').str()
+    local replicaset_uuid = g.server:eval('return box.info.replicaset.uuid')
+    g.server:exec(function()
+        -- Skip one iproto_flush call and disable obuf flushing
+        box.error.injection.set('ERRINJ_IPROTO_FLUSH_DELAY_COUNTDOWN', 1)
+    end)
+    -- Several requests in a row to pretend to be a non-blocking client and
+    -- provoke iproto_connection_feed_input, so that subscribe is immediately
+    -- read
+    for _ = 1, 3 do
+        write_ok(g.s)
+    end
+    -- A short pause so that the packages do not end up in one batch and all
+    -- responses have time to be written in obuf
+    require('fiber').sleep(0.1)
+    write_subscribe(g.s, uuid, replicaset_uuid, true)
+    -- A short pause so that the tx thread has time to write the response to
+    -- the subscribe request to the socket
+    require('fiber').sleep(0.1)
+    g.server:exec(function()
+        box.error.injection.set('ERRINJ_IPROTO_FLUSH_DELAY', false)
+    end)
+    local request_type
+    for _ = 1, 3 do
+        request_type = socket_read(g.s)[key.REQUEST_TYPE]
+        t.assert_equals(request_type,
+            iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
+    end
+    request_type = socket_read(g.s)[key.REQUEST_TYPE]
+    t.assert(request_type == type.OK or request_type ==
+        iproto_error_type(box.error.PROTOCOL))
+    g.server:exec(function()
+        box.cfg{net_msg_max = box.NULL}
+    end)
 end
 
 -- The case sends dummy OKs before subscribe
@@ -140,7 +180,7 @@ g.test_iproto_crash_on_subscribe_spam_ok = function(g)
     end
     request_type = socket_read(g.s)[key.REQUEST_TYPE]
     t.assert(request_type == type.OK or request_type ==
-        iproto_error_type(box.error.UNABLE_TO_PROCESS_REPLICATION))
+        iproto_error_type(box.error.PROTOCOL))
 end
 
 -- The case simulates a situation where the user of anonymous replication
@@ -161,7 +201,7 @@ g.test_iproto_crash_fetch_snapshot_subscribe = function(g)
         iproto_error_type(box.error.UNKNOWN_REQUEST_TYPE))
     request_type = socket_read(g.s)[key.REQUEST_TYPE]
     t.assert(request_type == type.OK or request_type ==
-        iproto_error_type(box.error.UNABLE_TO_PROCESS_REPLICATION))
+        iproto_error_type(box.error.PROTOCOL))
 end
 
 -- The same as above, but additionally forgot to pass is_anon option to
@@ -183,5 +223,5 @@ g.test_iproto_crash_fetch_snapshot_subscribe_not_anon = function(g)
     request_type = socket_read(g.s)[key.REQUEST_TYPE]
     t.assert(request_type ==
         iproto_error_type(box.error.TOO_EARLY_SUBSCRIBE) or request_type ==
-        iproto_error_type(box.error.UNABLE_TO_PROCESS_REPLICATION))
+        iproto_error_type(box.error.PROTOCOL))
 end
diff --git a/test/box/error.result b/test/box/error.result
index 12031f8fcb..506118065e 100644
--- a/test/box/error.result
+++ b/test/box/error.result
@@ -500,7 +500,6 @@ t;
  |   287: box.error.WAL_QUEUE_FULL
  |   288: box.error.INVALID_VCLOCK
  |   289: box.error.SYNC_QUEUE_FULL
- |   290: box.error.UNABLE_TO_PROCESS_REPLICATION
  | ...
 
 test_run:cmd("setopt delimiter ''");
```

src/box/iproto.cc
```diff
@@ -845,6 +845,8 @@ struct iproto_connection
 	 * This field is accesable only from iproto thread.
 	 */
 	struct mh_i64ptr_t *streams;
+	/** Count of active requests that write something to obuf (gh-10155). */
+	size_t active_writing_requests;
```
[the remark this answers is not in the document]
Astronomax: AFAIU `mempool_count` counts all `iproto_msg`'s for all connections, but I'm only interested in messages associated with a specific connection. Messages from the other connection don't prevent me from accepting the replication request in another connection.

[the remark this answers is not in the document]
locker: True, my mistake.

src/box/iproto.cc
Astronomax: I had a misunderstanding, thank you.

src/box/iproto.cc
Astronomax: Fixed.

src/box/errcode.h
```diff
@@ -434,6 +434,7 @@ struct errcode_record {
 	_(ER_WAL_QUEUE_FULL, 287,		"The WAL queue is full") \
 	_(ER_INVALID_VCLOCK, 288,		"Invalid vclock", "value", STRING) \
 	_(ER_SYNC_QUEUE_FULL, 289,		"The synchronous transaction queue is full") \
+	_(ER_UNABLE_TO_PROCESS_REPLICATION, 290, "Replication request couldn't be processed, responses to previous requests were not read") \
```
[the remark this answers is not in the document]
Astronomax: Fixed.

src/box/iproto.cc
```diff
@@ -1283,6 +1290,18 @@ iproto_connection_input_buffer(struct iproto_connection *con)
 	return new_ibuf;
 }
 
+static inline bool
+iproto_flushed(struct iproto_connection *con)
```
locker, line 1294: Please rename to `iproto_is_flushed`.

src/box/iproto.cc
```diff
@@ -1283,6 +1290,18 @@ iproto_connection_input_buffer(struct iproto_connection *con)
...
+{
+	struct obuf *obuf = con->wpos.obuf;
+	struct obuf_svp *begin = &con->wpos.svp;
+	struct obuf_svp *end = &con->wend.svp;
+
+	if (con->wend.obuf != obuf)
+		return (begin->used == obuf->used) && (end->used == 0);
+	return (begin->used == end->used);
```
locker, line 1302: AFAIU if both buffers are empty then `wpos` and `wend` must point to the same buffer. That is, the following check should be enough, no?

```C
return con->wpos.obuf == con.wend.obuf &&
       con->wpos.svp.used == con->wend.svp.used;
```

Astronomax: Yes, it's true. Thanks!

locker (approved):

sergepetrenko (approved):
Thanks for the fix!

sergepetrenko:
@Astronomax, please, prepare PRs with backports of this patch to 2.11 and 3.2 (3.3 wasn't released yet when I merged the patch, so backport to 3.3 isn't necessary).

Astronomax:
@sergepetrenko, prepared:
https://github.com/tarantool/tarantool/pull/10892
https://github.com/tarantool/tarantool/pull/10893

PR #10599, opened 2024-09-22
box: support asynchronous commit modes for synchronous transactions

### LINKED ISSUE #10583: Support asynchronous `wait` modes for committing synchronous transactions
Some users may want to overcome the throughput limitation of synchronous transactions and commit them using asynchronous `wait` modes (2c66624). This will allow them to get the same consistency guarantees with `read-confirmed` and `linearizable` isolation levels without paying the throughput penalty and also be able to see changes committed this way with the `read-committed` isolation level. This will essentially enable early ack transactions as described in the [Async in the presence of sync](https://www.notion.so/tarantool/Async-in-the-presence-of-sync-9456c50afefc4b2dbe5514d8e1582803?pvs=4) RFC.

CuriousGeorgiy:
This patch adds support for asynchronous commit modes to synchronous transactions.

# ‼️ Please do not review the first 2 commits, they belong to https://github.com/tarantool/tarantool/pull/9869 ‼️

Closes #10583


src/box/txn.c
```diff
@@ -1233,6 +1229,12 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 			txn_limbo_ack(&txn_limbo, txn_limbo.owner_id,
 				      limbo_entry->lsn);
```
Gerold103, line 1230: `wait_mode='none'` guarantees that there won't be any WAL writes. No yields. But `txn_limbo_ack()` can make a `CONFIRM` WAL write. We need to make is async for `wait_mode='none'`.

test/box-luatest/gh_10583_support_async_wait_modes_for_sync_tx_test.lua
```diff
@@ -0,0 +1,206 @@
...
+        box.atomic({wait = wait_mode}, function() box.space.s:replace{0} end)
+        t.assert(box.info.synchro.queue.len, 1)
+
+        box.begin{txn_isolation = 'read-committed'}
+        t.assert_equals(box.space.s:get{0}, {0})
+        box.commit()
+        box.begin{txn_isolation = 'read-confirmed'}
+        t.assert_equals(box.space.s:get{0},
+                        _G.are_txns_not_isolated and {0} or nil)
+        box.commit()
```
Gerold103, line 53: Before and after this function the counter `fiber.self():csw()` must remain unchanged when `wait_mode='none'`.

test/replication-luatest/commit_wait_test.lua
```diff
@@ -1,58 +0,0 @@
-local t = require('luatest')
```
Gerold103, line 1: Lets keep the original test file name. The rules are that we use `gh_....` prefix only for bugfixes. For features the test file names don't mention tickets in their title. Usually. Here it seems most appropriate because it covers multiple feature-tickets concerning the same feature - `wait_mode`.

Gerold103 (commented):
Thanks, that was really quick 🔥, I thought it would actually take us much longer to make sync txns non-blocking.

changelogs/unreleased/gh-10583-support-async-wait-modes-for-sync-tx.md
```diff
@@ -0,0 +1,5 @@
+## feature/box
+
+* Added support for asynchronous wait modes (`box.commit{wait = ...}`) to
+  synchronous transactions. Changes committed this way can be observed with the
+  `read-committed` isolation level (gh-10583).
```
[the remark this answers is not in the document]
CuriousGeorgiy: We actually have `read-committed` and `read-confirmed` isolation levels instead of `read-prepared` and `read-confirmed`😁.

lenkis (approved):

src/box/txn.c
```diff
@@ -1201,15 +1195,17 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
...
 		fiber_set_txn(fiber(), txn);
 		goto rollback_io;
 	}
-	if (wait_mode != TXN_COMMIT_WAIT_MODE_COMPLETE) {
+	if ((!txn_has_flag(txn, TXN_WAIT_SYNC) ||
+	     txn_limbo.owner_id != instance_id) &&
+	    wait_mode != TXN_COMMIT_WAIT_MODE_COMPLETE) {
 		if (txn_has_flag(txn, TXN_IS_DONE))
 			goto finish_done;
 		txn->fiber = NULL;
```
Serpentian, line 1211: Two lines below there's an explicit yield in the `none` commit mode, if tx is synchronous. 

```C
	while (!req->is_complete)
		fiber_yield();
```

Why don't we explicitly forbid using sync tx with `none` mode?

Astronomax:
A transaction (vinyl or mvcc) can call `yield`. Is it ok to continue processing transaction in `wait='none'` mode? Maybe we should immediately abort the transaction when it calls `yield` for the first time? In addition, `txn_prepare` in `txn_commit_impl` runs `box.before_commit` event triggers, which also allow yield https://github.com/tarantool/tarantool/pull/9384.

Gerold103:
> Is it ok to continue processing transaction in wait='none' mode? 

@Astronomax, `wait='none'` is specified in `box.commit()`. There is no *"continue"* after this point. User can't yield the txn after it its submission for commit.

Astronomax:
@Gerold103, oh yes, I got it. But what about triggers called from `box.commit`?

Gerold103:
@Astronomax they can't yield. They are invoked inside of the scheduler-fiber, any yield in there causes panic and abortion.

src/box/txn.c
```diff
@@ -1133,12 +1133,19 @@ txn_commit_nop(struct txn *txn)
  * write completes.
  */
 static int
-txn_add_limbo_entry(struct txn *txn, const struct journal_entry *req)
+txn_add_limbo_entry(struct txn *txn, const struct journal_entry *req,
```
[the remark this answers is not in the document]
CuriousGeorgiy: Please read https://github.com/tarantool/tarantool/pull/10461#issuecomment-2356150239.

src/box/txn.c
CuriousGeorgiy: Fixed: after #10628 we can immediately return here.

test/box-luatest/gh_10583_support_async_wait_modes_for_sync_tx_test.lua
CuriousGeorgiy: Added checks for this.

src/box/txn.c
CuriousGeorgiy: Fixed: after #10628 we can return immediately after journal request submission.

test/replication-luatest/commit_wait_test.lua
CuriousGeorgiy: Fixed.

Astronomax:
@Gerold103, but the original [PR](https://github.com/tarantool/tarantool/pull/9384) says the following:
>
`box.before_commit` trigger-function restrictions:

Yield is allowed;
Allowed to write into database spaces;
If the function raises an error, the transaction is rolled back.
`box.on_commit` and `box.on_rollback` trigger-function restrictions:

Yield is forbidden (it will crash Tarantool);
The function should not access any database spaces;
If the function raises an error, the error is simply logged.

Also, I checked this manually on an engine (vinyl) that allows yield (otherwise on memtx without mvcc the transaction is aborted with "- error: Transaction has been aborted by a fiber yield", triggers obviously don't even run):
```console
tarantool> trigger = require('trigger')
---
...

tarantool> trigger.set('box.before_commit.space.test_vinyl', 't', function()
require('fiber').yield()
print('yield in before_commit trigger!')
print(require('fiber').name())
end)
---
- 'function: 0x40be6028'
...

tarantool> box.space.test_vinyl:insert{1}
yield in before_commit trigger!
interactive
---
- [1]
...

tarantool> 
```
But `yield()` in `on_commit` causes segfault:
```console
tarantool> trigger.set('box.on_commit.space.test_vinyl', 't', function()
require('fiber').yield()
print('yield in on_commit trigger!')
end)
---
- 'function: 0x41452470'
...

tarantool> box.space.test_vinyl:insert{2}
yield in before_commit trigger!
interactive
Segmentation fault
  code: SEGV_MAPERR
  addr: 0xb0
  context: 0x5928b9c36100
  siginfo: 0x5928b9c36230
  rax      0x0                0
  rbx      0x5928b7c234a0     98031416521888
  rcx      0x5928b9d6b010     98031451418640
  rdx      0x3ff0000000000000 4607182418800017408
  rsi      0x73ae2fb3eac0     127191961823936
  rdi      0x73ae2fb3eac0     127191961823936
  rsp      0x7fff04f75950     140733276707152
  rbp      0x7fff04f75980     140733276707200
  r8       0x5928b7549508     98031409337608
  r9       0x5928ba053549     98031454467401
  r10      0x0                0
  r11      0xfffffffc         4294967292
  r12      0x5928b7c23560     98031416522080
  r13      0x7fff04f76278     140733276709496
  r14      0x419420a0         1100226720
  r15      0x5928b7440bcd     98031408253901
  rip      0x5928b7478f35     98031408484149
  gs       0x0                0
  fs       0x0                0
  cr2      0xb0               176
  err      0x4                4
  oldmask  0x0                0
Current time: 1730620443
this may fail as well
#1  0x5928b746ef25 in crash_collect+256
#2  0x5928b746f961 in crash_signal_cb+100
#3  0x73ae2e842520 in __sigaction+80
#4  0x5928b7478f35 in cord_reset_slice+263
#5  0x5928b747a41a in fiber_yield_impl+262
#6  0x5928b747a586 in fiber_yield+18
#7  0x5928b747a7f3 in fiber_yield_timeout+314
#8  0x5928b747a9fa in fiber_sleep+244
#9  0x5928b7440bee in lbox_fiber_yield+33
#10 0x5928b74d9a77 in lj_BC_FUNCC+70
#11 0x5928b74e75ec in lua_pcall+968
#12 0x5928b744b018 in luaT_call+45
#13 0x5928b742db06 in func_adapter_lua_call+448
#14 0x5928b72e0899 in func_adapter_call+51
#15 0x5928b72e1561 in run_triggers_general+248
#16 0x5928b72e166b in run_triggers_of_single_space+163
#17 0x5928b72e19a9 in run_triggers_of_spaces+162
#18 0x5928b72e1a1b in run_triggers+51
#19 0x5928b72e1ad2 in txn_event_on_commit_run_triggers+58
#20 0x5928b72d6ab0 in txn_complete_success+728
#21 0x5928b72d7038 in txn_on_journal_write+1364
#22 0x5928b733e16a in journal_async_complete+102
#23 0x5928b733ebd8 in tx_schedule_queue+54
#24 0x5928b733ed64 in Аварийный останов
```

Astronomax:
With the current implementation, I have to wait until the `box.before_commit` trigger completes execution:
```console
tarantool> trigger = require('trigger')
---
...

tarantool> trigger.set('box.before_commit.space.test_vinyl', 't', function()                               
require('fiber').sleep(10)
end)                                    
---
- 'function: 0x402fa6d0'
...

tarantool> box.begin() box.space.test_vinyl:insert{1} box.commit({wait = 'none'})
---
...
```
For example, here the user gets control back only after 10 seconds.

Gerold103:
@Astronomax ah, sorry, yeah, I thought you mean only `box.on_commit`. Yes, `box.before_commit()` apparently can yield. Indeed it looks strange then, that `box.commit(wait=none)` would still yield, if we want to promise no yields. Apparently, we need to reword the guarantees that it won't yield unless some before-commit triggers do, and it 100% won't wait for WAL write or even yield to make it.

CuriousGeorgiy:
@Astronomax I think this is out of the scope of this PR and affect the `wait_mode='none'` in general. Could you please open a separate ticket in `tarantool/doc`?

Astronomax:
Another issue (or maybe not an issue either). We still support `replication_synchro_timeout` if `compat.replication_synchro_timeout == 'old'`, but it doesn't work if `wait == 'none'`. Simply because the fiber in which the commit is executed is also responsible for rollback due to `replication_synchro_timeout`. We exit the `txn_commit_impl` before `txn_limbo_wait_complete` is called. Maybe we’ll move the `replication_synchro_timeout` rollback handling out of the fiber (out of the `txn_limbo_wait_complete`), for example, into a callback of some timer, or in any other way make it asynchronous?

CuriousGeorgiy:
@Astronomax I don't think you care about the result of the transaction with `wait='submit'` or `wait=none`. You don't know if it got committed or rolled back in the case of asynchronous transactions too.

I don't think it makes sense to complicate the code for an old backward compatibility option, if it is avoidable. If anyone complains about this behaviour, we can always change it later.

src/box/txn_limbo.c
```diff
@@ -237,6 +237,8 @@ txn_limbo_on_remove(struct txn_limbo *limbo,
 	assert(limbo->size >= 0);
 	limbo->len--;
 	assert(limbo->len >= 0);
+	/* Wake up all fiber waiting to add a new limbo entry. */
+	fiber_cond_broadcast(&limbo->wait_cond);
```
Gerold103, line 241: It is going to be quite expensive to do that on each popped limbo entry. Because it would wakeup all the limbo-waiting fibers, including the `txn_limbo_wait_complete()`, and those we might have thousands. They would all get a spurious wakeup. This cond should only be used by rare events, like when the limbo parameters change.

A not too intrusive alternative would be to invoke it only when `limbo->len == limbo->max_size` (before the decrement). It means the queue was full and became not full. A state change, right? We are not interested in the cases, when the queue was full and remains full somehow. Or was not full and remains not full. Still a lot of spurious wakeups, but now only when the queue max size is reached, which is supposed to be rare.

Another way would be in `txn_add_limbo_entry()` to subscribe on the first limbo txn's on_commit trigger and the last txn's on_rollback trigger. In those triggers you would wake your fiber, and not any other fiber. And you would get notified when anything leaves the limbo. OTOH, if we get a lot of waiters and they all subscribe on the same border txns, they would all be woken up. Thus again spurious wakeups.

Unless you could think of any better ways, I would then suggest the hack above about `len == max_size` before decrement.

test/replication-luatest/commit_wait_test.lua
```diff
@@ -1,58 +1,207 @@
...
+            end)
+        end)
+        atomic_f:set_joinable(true)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+
+        box.cfg{replication_synchro_quorum = 1}
+        t.assert(sync_f:join())
+        box.cfg{replication_synchro_quorum = 2}
+        t.assert_equals(box.space.s:get{0}, {0})
+        t.assert_equals(box.info.synchro.queue.len, 1)
```
Gerold103, line 184: Why is it guaranteed 1? You had a moment after `box.cfg{replication_synchro_quorum = 1}` when there was a yield. Technically the second transaction during that yield could reach WAL, get written, produce a `CONFIRM`, and get committed. All before you restore `box.cfg{replication_synchro_quorum = 2}`.

Gerold103 (commented):
Thank you! I am happy to see how simple is this patch after the ground work with the async `CONFIRM` is done. Good work 🤝.

src/box/txn.c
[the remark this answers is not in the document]
Gerold103: I still didn't understand from that comment what is the intention. Do we keep async with elections are enabled or not? If we do, then we still have a problem with the split brain. AFAIU, the intention was that with the new wait_mode support for sync txns we can make all the transactions synchronous when the limbo is owned. What is the design vision, @sergepetrenko?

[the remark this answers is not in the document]
sergepetrenko: @Gerold103, I think we'll gradually come to a situation when async replication isn't used in mix with synchro.
With the patch Georgiy is working on we provide the users with all the functionality to use only synchronous replication. Now we should advertise this setup as the only valid one for raft/syncrho.

But disabling async replication completely when synchro is enabled needs to be thought through separately. Synchronous transactions with `wait_mode='submit'` still provide a different semantic compared to asynchronous transactions, so replacing async with sync would be a breaking change.

Besides I'm not sure we have thought of all the possible use cases for async + sync. Probably we should make the "only synchro" mode the default one but still let the users use async + sync if they really want to.

So, let's design the "only synchronous" transition separately, out of scope of this task.

src/box/txn_limbo.c
CuriousGeorgiy: I guess the ideal robust solution would be to do it similarly to `journal_queue` (refactoring it to some `fiber_write_queue` abstraction), but I would prefer going for a simpler solution that you mentioned for now.

> I would then suggest the hack above about `len == max_size` before decrement.

Done.

test/replication-luatest/commit_wait_test.lua
CuriousGeorgiy: I see your concern, but the fiber responsible for the second transaction will have to yield after submitting the journal request and AFAIC it is guaranteed that it will be scheduled later than the fiber running the test code. I have also ran many test instances concurrently with a large number of test-run jobs, and it doesn't fail.

src/box/txn.c
```diff
@@ -1150,12 +1150,19 @@ txn_commit_nop(struct txn *txn)
...
 {
 	uint32_t origin_id = req->rows[0]->replica_id;
 	if (origin_id == 0 && txn_limbo.size >= txn_limbo.max_size) {
-		diag_set(ClientError, ER_SYNC_QUEUE_FULL);
-		return -1;
+		if (wait_mode == TXN_COMMIT_WAIT_MODE_NONE) {
+			diag_set(ClientError, ER_SYNC_QUEUE_FULL);
+			return -1;
+		}
+		while (txn_limbo.size >= txn_limbo.max_size) {
```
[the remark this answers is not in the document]
CuriousGeorgiy: Added a `txn_limbo_wait_for_space` helper.

src/box/txn_limbo.c
```diff
@@ -451,14 +451,16 @@ txn_limbo_wait_complete(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
...
 		e->txn->signature = TXN_SIGNATURE_QUORUM_TIMEOUT;
 		e->txn->limbo_entry = NULL;
 		txn_limbo_abort(limbo, e);
 		txn_clear_flags(e->txn, TXN_WAIT_SYNC | TXN_WAIT_ACK);
 		txn_limbo_complete(e->txn, false);
 		if (e == entry)
 			break;
-		fiber_wakeup(e->txn->fiber);
+		if (needs_wakeup)
+			fiber_wakeup(e->txn->fiber);
```
Serpentian, line 478: Do we really need `fiber_wakeup` here? `txn_complete_fail` will call `txn_free_or_wakeup`, which will wake up the fiber, if it's not `NULL`. All tests seem to pass for me without `wakeup`

Serpentian (commented):
This is probably the last comment I have. Looks great!

changelogs/unreleased/gh-10853-async-commit-async-tx-after-sync-tx-synchro-timeout-crash.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/box
+
+* Fixed a crashed caused by asynchronously committing an asynchronous
+  transaction after a synchronous transaction with a timeout, the latter
+  timing out (gh-10853).
```
Serpentian, line 5: Nit: It's probably the "Fixed ~~a crashed~~ the crash ...". And the meaning of the changelog eludes me, we probably should say it easier, smth like: "Fixed the crash caused by asynchronously committing an asynchronous transaction following a synchronous timed out transaction"

Serpentian (approved):
Thank you! LGTM

Gerold103 (approved):
Thank you, @CuriousGeorgiy. I am happy to see your fix making the code more flexible, and that this patchset managed to fix a crash, and that we no longer panic on confirmation failure. It was an old and untrivial piece of tech debt.

test/replication-luatest/commit_wait_test.lua
Gerold103: Hm. So the test entirely depends on the fiber scheduling order. That looks quite unstable, but I don't mind fixing it later, when it starts flakying.

test/box-luatest/gh_10853_async_commit_async_tx_after_sync_tx_synchro_timeout_crash_test.lua
```diff
@@ -0,0 +1,39 @@
...
+
+-- Test that asynchronously committing an async transaction after a sync
+-- transaction that later times out on `replication_synchro_timeout` works
+-- correctly.
+g.test_async_commit_async_async_tx_after_sync_tx_synchro_timeout = function(cg)
+   cg.s:exec(function()
+       local f = require('fiber').create(function() box.space.s:replace{0} end)
+       f:set_joinable(true)
+       box.atomic({wait = 'none'}, function() box.space.a:replace{0} end)
+       t.assert_equals(box.info.synchro.queue.len, 2)
```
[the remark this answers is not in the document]
CuriousGeorgiy: > It might happen before you reach the length check.

But we don't yield until line 34, right? So how can the timeout fire before we yield? Anyway, applied your suggestion to make it clear.

changelogs/unreleased/gh-10853-async-commit-async-tx-after-sync-tx-synchro-timeout-crash.md
CuriousGeorgiy: Fixed.

Issue #10520, opened 2024-09-05
Allow DDL, when upgrading from 2.X to 3.X 


Serpentian:
In the scope of https://github.com/tarantool/tarantool/issues/7149 DDL was fully forbidden until the schema is upgraded. However, In order to simplify the process of tarantool upgrade it was decided to partially allow it on the old schema. 

- DDL remains forbidden, when schema is way too old (less than 2.11.1).
- DDL is mostly allowed, when 3.X is started on top of schema >= 2.11.1. However, creating of persistent trigger is still forbidden. Inserting names is also explicitly forbidden until `box.schema.upgrade` is called (it's impossible to insert names  to `_cluster`, `_schema`, such attempt returns an error, which says user to upgrade schema) (names are under question). 
- When new schema versions appear, it'll be the responsibility of calling code (e.g. `box.schema.func.create`, `box.schema.space.create` and so on) to check that it's executed on a fresh enough schema version and throw an error otherwise.

In the scope of this issue, we must also verify, that Tarantool properly works before the scheme is upgraded.

Issue #10521, opened 2024-09-05
Introduce 


Serpentian:
I propose to introduce a new Tarantool module `version`, which will allow to comfortably work with versions. Currently user must manually parse `version` field from `_schema` and than write functions to compare versions. This is not very convenient.

API of the `version` module: TBD.

Internally it might use `internal.mkversion`. I suppose, the module should allow conveniently work with `_TARANTOOL` variable and with `version` from `_schema`. The inspiration could be taken from [vshard](https://github.com/tarantool/vshard/blob/master/vshard/version.lua)

Totktonada:
See also https://github.com/orgs/tarantool/discussions/7006 for other tarantool version parsing implementations (some of them are outdated, though).

I can't understand from the description, whether the module incapsulates the database version schema information or a user should know about `_schema` in order to use it for schema version checks?

PR #10618, opened 2024-09-26
schema: broadcast persistent schema version in box.status

### LINKED ISSUE #10546: Add `schema_version` to `box.status` broadcast
https://github.com/tarantool/tarantool/issues/10520 will allow users to apply the majority of the DDL during upgrade without any problems: user will have to wait for rw status (via watcher or some config role function) and create their spaces and indexes. The problem there, that some of the DDL will still be forbidden, until box.schema.upgrade is called (e.g. creating persistent triggers) and we should give users a convenient way to use such features.

Currently they will have to check schema version in `_schema`, then `on_replace` trigger for `_schema` will have to be set, and users will have to unregister that trigger after its execution. Moreover, this trigger will check if instance is rw. Not very convenient.

---

Let's  add new field in `box.status`: `schema_version`. In order to apply DDL after Tarantool upgrade, user will create watcher, which checks rw status and schema_version.

This will simplify the realization of https://github.com/tarantool/tarantool/issues/10538

--- 

The alternative of introducing completely new watcher, which includes only `schema_version`, was rejected as Tarantool doesn't allow to create watcher on several events. It will be difficult to synchronize two events: `box.status`, which includes `read_only`, and `box.schema_version`.

Serpentian:
https://github.com/tarantool/tarantool/issues/10520 will allow DDL before schema is upgraded
in order to simplify the upgrade process. However, some DDL will
remain forbidden until box.schema.upgrade is called, e.g. creating
persistent triggers when upgrading to 3.0.0.

In order to make writing upgrade triggers easier this commit adds
new field to the box.status system event: dd_version. This should
not be confused with schema.version value, as the latter shows, whether
the database schema (e.g. creating space) was updated.
box.status.dd_version on the other side is the value of the
persistent schema from box.space._schema['version'].

Without box.status.dd_version user has to set on_replace trigger for
box.space._schema, which will check version and rw state, after which
trigger will send it further. The problem with this approach is the fact,
that it requires synchronization of two changing states: dd schema
version and rw state, which is difficult to achieve.

With box.status.dd_version user has to just set watcher, which
checks !status.is_ro and status.dd_version >= needed one.

Closes https://github.com/tarantool/tarantool/issues/10546

### DIFF
--- a/changelogs/unreleased/gh-10546-schema-version-broadcast.md
+++ b/changelogs/unreleased/gh-10546-schema-version-broadcast.md
@@ -1,0 +1,3 @@
+## feature/core
+
+* Added the `dd_version` field to the `box.status` system event (gh-10546).
--- a/src/box/alter.cc
+++ b/src/box/alter.cc
@@ -4152,6 +4152,7 @@ on_commit_dd_version(struct trigger *trigger, void * /* event */)
 	struct fiber *fiber = data->fiber;
 	if (fiber != NULL)
 		fiber_wakeup(fiber);
+	box_broadcast_status();
 	return 0;
 }
 
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -297,12 +297,6 @@ static void
 box_storage_init(void);
 
 /**
- * Broadcast the current instance status
- */
-static void
-box_broadcast_status(void);
-
-/**
  * A timer to broadcast the updated vclock. Doing this on each vclock update
  * would be too expensive.
  */
@@ -6022,18 +6016,20 @@ box_broadcast_id(void)
 	assert((size_t)(w - buf) < sizeof(buf));
 }
 
-static void
+void
 box_broadcast_status(void)
 {
 	char buf[1024];
 	char *w = buf;
-	w = mp_encode_map(w, 3);
+	w = mp_encode_map(w, 4);
 	w = mp_encode_str0(w, "is_ro");
 	w = mp_encode_bool(w, box_is_ro());
 	w = mp_encode_str0(w, "is_ro_cfg");
 	w = mp_encode_bool(w, cfg_geti("read_only"));
 	w = mp_encode_str0(w, "status");
 	w = mp_encode_str0(w, box_status());
+	w = mp_encode_str0(w, "dd_version");
+	w = mp_encode_str0(w, version_id_to_string(dd_version_id));
 
 	box_broadcast("box.status", strlen("box.status"), buf, w);
 
--- a/test/box-luatest/upgrade_to_3_0_0_test.lua
+++ b/test/box-luatest/upgrade_to_3_0_0_test.lua
@@ -1,30 +1,46 @@
 local server = require('luatest.server')
+local replica_set = require('luatest.replica_set')
 local t = require('luatest')
 
 local g = t.group("Upgrade from 2.11.0")
 
 g.before_all(function(cg)
-    cg.server = server:new({
-        datadir = 'test/box-luatest/upgrade/2.11.0',
+    cg.replica_set = replica_set:new{}
+    cg.master = cg.replica_set:build_and_add_server({
+        alias = 'master',
+        datadir = 'test/box-luatest/upgrade/2.11.0/replicaset/instance-001'
     })
-    cg.server:start()
+    cg.replica = cg.replica_set:build_and_add_server({
+        alias = 'replica',
+        datadir = 'test/box-luatest/upgrade/2.11.0/replicaset/instance-002',
+        box_cfg = {
+            read_only = true,
+            replication = cg.master.net_box_uri,
+        },
+    })
+    cg.replica_set:start()
 end)
 
 g.after_all(function(cg)
-    cg.server:drop()
+    cg.replica_set:drop()
 end)
 
 g.after_each(function(cg)
-    cg.server:exec(function()
+    cg.master:exec(function()
         box.schema.downgrade('2.11.0')
+        box.snapshot()
+    end)
+    cg.replica:wait_for_vclock_of(cg.master)
+    cg.replica:exec(function()
         box.snapshot()
     end)
 end)
 
 g.test_new_replicaset_uuid_key = function(cg)
-    cg.server:exec(function()
+    cg.master:exec(function()
         t.assert_equals(box.space._schema:get{'version'}, {'version', 2, 11, 0})
         box.schema.upgrade()
+
         local _schema = box.space._schema
         t.assert_equals(_schema:get{'cluster'}, nil)
         t.assert_equals(_schema:get{'replicaset_uuid'}.value,
@@ -33,7 +49,7 @@ g.test_new_replicaset_uuid_key = function(cg)
 end
 
 g.test_ddl_is_forbidden = function(cg)
-    cg.server:exec(function()
+    cg.master:exec(function()
         t.assert_equals(box.space._schema:get{'version'}, {'version', 2, 11, 0})
         -- DDL is prohibited before 2.11.1.
         local msg = "DDL_BEFORE_UPGRADE requires schema version 2.11.1"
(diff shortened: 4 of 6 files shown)


xuniq (approved):

src/box/box.cc
```diff
@@ -299,7 +299,7 @@ box_storage_init(void);
 /**
  * Broadcast the current instance status
  */
-static void
+void
 box_broadcast_status(void);
```
locker, line 303: You can remove this now, as you have the declaration in the header.

src/box/box.cc
```diff
@@ -5963,18 +5963,20 @@ box_broadcast_id(void)
...
-	w = mp_encode_map(w, 3);
+	w = mp_encode_map(w, 4);
 	w = mp_encode_str0(w, "is_ro");
 	w = mp_encode_bool(w, box_is_ro());
 	w = mp_encode_str0(w, "is_ro_cfg");
 	w = mp_encode_bool(w, cfg_geti("read_only"));
 	w = mp_encode_str0(w, "status");
 	w = mp_encode_str0(w, box_status());
+	w = mp_encode_str0(w, "schema_version");
+	w = mp_encode_str0(w, version_id_to_string(dd_version_id));
```
locker, line 5979: Wouldn't it be better to return the version as an array, e.g. `{3, 1, 0}`? It'd be easier for comparisons.

src/box/box.cc
```diff
@@ -5963,18 +5963,20 @@ box_broadcast_id(void)
...
 	char *w = buf;
-	w = mp_encode_map(w, 3);
+	w = mp_encode_map(w, 4);
 	w = mp_encode_str0(w, "is_ro");
 	w = mp_encode_bool(w, box_is_ro());
 	w = mp_encode_str0(w, "is_ro_cfg");
 	w = mp_encode_bool(w, cfg_geti("read_only"));
 	w = mp_encode_str0(w, "status");
 	w = mp_encode_str0(w, box_status());
+	w = mp_encode_str0(w, "schema_version");
```
locker, line 5978: Wouldn't it be better to name the new field `dd_version` so as not to confuse with the schema version we use in IPROTO?

src/box/box.cc
Serpentian: WDYT of returning `dd_version_id` itself, as unsigned integer? In the scope of https://github.com/tarantool/tarantool/issues/10521 `version` module will be introduced and I'm intending on adding `version.from_id(info.dd_version_id)` method there, so it'll be easy to use, even better IMHO than calling smth like `version.new(info.version[1], info.version[2], info.version[3])`.

P.S. RFC for version module is ready for review

locker: Once we add the version id to the public API, we won't be able to change its representation. In particular, using the version module as is would limit the max version number to 255:

https://github.com/tarantool/tarantool/blob/7cf41a1bb9b3f0a2eac82c296a764688f5d8b35c/src/box/lua/mkversion.lua#L42-L47

So I'd prefer returning an array or a string instead. If we export the version module, using strings would probably be fine.

src/box/box.cc
Totktonada: I agree that it is better to use different names, but I think that `dd_version` is closer to DD itself rather than its schema/format/layout. I can propose `dd_layout` here.

src/box/alter.cc
```diff
@@ -4150,6 +4150,7 @@ on_commit_dd_version(struct trigger *trigger, void * /* event */)
 	struct fiber *fiber = data->fiber;
 	if (fiber != NULL)
 		fiber_wakeup(fiber);
+	box_broadcast_status();
```
Totktonada, line 4153: Is it issued on downstreams too or only on an instance that performs the schema upgrade?

I guess that the former is true. If it is not complicated, I would add a test case that makes it clear.

test/box-luatest/upgrade_to_3_0_0_test.lua
```diff
@@ -20,9 +16,29 @@ end)
...
+        t.assert_equals(version, mkversion.new(2, 11, 0))
+
+        box.cfg{read_only = false}
+        box.schema.upgrade()
+
         local _schema = box.space._schema
         t.assert_equals(_schema:get{'cluster'}, nil)
         t.assert_equals(_schema:get{'replicaset_uuid'}.value,
                         box.info.replicaset.uuid)
+        t.assert_equals(version, mkversion.get_latest())
```
Totktonada, line 42: Nit: It is mixed with another test case (regarding names). I think it would be more clear if we split them.

src/box/box.cc
Totktonada: I think that a string or an array is fine. Can't say what is better -- so, I'm OK with any of these variants.

src/box/box.cc
Serpentian: Indeed, removed

src/box/box.cc
Serpentian: I like string better, let's use it, since nobody cares

src/box/box.cc
Serpentian: I find name `dd_layout` strange, I suppose, we should be consistent with internal variable naming (`dd_version_id`). Let's use `dd_version` in `box.info`. @Totktonada, what do you think?

src/box/alter.cc
Serpentian: It's executed everywhere, on every instance in the replicaset. Added test for that

test/box-luatest/upgrade_to_3_0_0_test.lua
Serpentian: Sure, I separated it from schema checking test.

src/box/box.cc
Totktonada: If it is to follow an existing naming (that doen't not interfere with 'schema version'), I'm OK. Well, I think that naming could be better if we implement it from scratch, but the consistency seems more important for me.

Am I understand right: we're going to add `box.info.dd_version` and `dd_version` field into `box.status` event? If so, I'm OK.

changelogs/unreleased/gh-10546-schema-version-broadcast.md
```diff
@@ -0,0 +1,3 @@
+## feature/core
+
+* Added the `schema_version` field to the `box.status` system event (gh-10546).
```
Totktonada, line 3: Same, should be `dd_version`.

Totktonada (approved):
No objections from me.

src/box/box.cc
Serpentian: > we're going to add box.info.dd_version

Do we really need that? User can get it from `_schema` space. We already duplicate that, but in `box.status` it's for convenience of upgrade process. Why do we want to add it to the `box.info`?

Totktonada: Reading a specific value from a system space is not something that we should normally suggest to a user. A system space layout/content is a kind of public API due to historical reasons, at least for some spaces. But if something is going to be widely used/too visible, I would add it into some 'undoubtedly public' API.

Also, this idea appears in the version module RFC discussion and it seems meaningful for me.

I don't insist, but I think that it is a good idea.

changelogs/unreleased/gh-10546-schema-version-broadcast.md
Serpentian: My bad, fixed

src/box/box.cc
```diff
@@ -5963,18 +5957,20 @@ box_broadcast_id(void)
...
 	char *w = buf;
-	w = mp_encode_map(w, 3);
+	w = mp_encode_map(w, 4);
 	w = mp_encode_str0(w, "is_ro");
 	w = mp_encode_bool(w, box_is_ro());
 	w = mp_encode_str0(w, "is_ro_cfg");
 	w = mp_encode_bool(w, cfg_geti("read_only"));
 	w = mp_encode_str0(w, "status");
 	w = mp_encode_str0(w, box_status());
+	w = mp_encode_str0(w, "dd_version");
```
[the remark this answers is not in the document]
Serpentian: Fixed everything)

src/version.h
```diff
@@ -75,6 +75,7 @@ version_id_to_string(uint32_t version_id)
 			  version_id_minor(version_id),
 			  version_id_patch(version_id));
 }
+
```
locker, line 78: Nit: This belongs to the previous commit.

locker (approved):

Serpentian:
Rebased

PR #10619, opened 2024-09-26
[backport 2.11] schema: allow _cluster update after join

### LINKED ISSUE #10549: [2.11 only] Any update of an existing record to `box.space._cluster` fails, even if instance id/uuid is untouched.
When I try to update an existing entry in space `_cluster` and add a new field to it, I receive an error: `error: The local instance id 1 is read-only`.

Happens due to `replica_check_id()`:
https://github.com/tarantool/tarantool/blob/8cebbf2cad5dec84d669ab24cf8c0f6ec8856538/src/box/replication.cc#L235-L251

```
tarantool> box.space._cluster:fselect{}
---
- - +-----+--------------------------------------+
  - ǀ id  ǀ                 uuid                 ǀ
  - +-----+--------------------------------------+
  - ǀ  1  ǀ"a8f6b359-d54d-4010-90ea-20e4ce680d04"ǀ
  - +-----+--------------------------------------+
...

tarantool> box.space._cluster:update({1}, {{'=', 3, 'my_name'}})
---
- error: The local instance id 1 is read-only
...

tarantool> tup=box.space._cluster:get{1}
---
...

tarantool> tup = tup:update{{'=', 3, 'my_name'}}
---
...

tarantool> tup
---
- [1, 'a8f6b359-d54d-4010-90ea-20e4ce680d04', 'my_name']
...

tarantool> box.space._cluster:replace(tup)
---
- error: The local instance id 1 is read-only
...
```

This was fixed in 3.0+, because instance names introduction demanded this change, so now we only have to fix 2.11.4 for the sake of upgrade procedure easening in scope of https://github.com/tarantool/tarantool/issues/10311

Serpentian:
The function replica_check_id() is called on any change in
_cluster: insert, delete, update. It was supposed to check if the
replica ID is valid - not nil, not out of range (VCLOCK_MAX).

But it was also raising an error when the ID matched this
instance's ID unless the instance was joining. That happened even
if a _cluster tuple was updated without changing the ID at all.
For example, if one would just do
_cluster:replace(_cluster:get(box.info.id)).

Better do the check in the only place where the mutation can
happen - on deletion. Since replica ID is a primary key in
_cluster, it can't be updated there. Only inserted or deleted.

This commit is backported to 2.11, since we want to allow using
persistent names as early as we can in order to simplify the upgrade
process. We also bump the schema version in the following commit in
order to distinguish this version from overs 2.11.X, where persistent
names doesn't work.

Closes https://github.com/tarantool/tarantool/issues/10549

NO_DOC=bugfix and refactoring
NO_CHANGELOG=cannot happen without touching system spaces
NO_TEST=too insignificant for an own test

(cherry picked from commit https://github.com/tarantool/tarantool/commit/cb8f4715c58aeaa5008888b66a328f200198a93c)


src/box/alter.cc
Totktonada, line 1: I would rewrite the commit message in context of our reason to backport it to 2.11. The original message is misleading, when it is out of the original patch series.

See d8460466680f07f7cd75c4a1b273f20a18a2f3a5 or 46be3c9942c99fd7e8211b143bdd27bb892353bc for example.

Totktonada (approved):
No objections from my side. Thanks!

src/box/alter.cc
Serpentian: Fixed the commit msg

sergepetrenko (approved):
Thanks!
LGTM.

PR #10626, opened 2024-09-28
txn: ack limbo owner WAL writes to the limbo asynchronously

### LINKED ISSUE #10628: Acknowledge synchronous queue owner WAL writes to the synchronous queue asynchronously
Currently, a synchronous transaction must always wait for its WAL write to finish, acknowledge itself to the synchronous queue, and potentially wait for a synchronous CONFIRM request WAL write.

In the scope of #10583, we want to support asynchronous wait modes for synchronous transactions. To enable the `none` mode, we need to acknowledge our own WAL writes for the synchronous queue asynchronously by:
1. Moving the acknowledgment to the callback for finishing transaction WAL write;
2. Detaching the CONFIRM request WAL writing from the acknowledgment, and making it asynchronous.

CuriousGeorgiy:
This patch set enables asynchronous acknowledgment of limbo owner WAL writes to the limbo.

Closes #10628


src/box/txn_limbo.c
```diff
@@ -387,9 +387,8 @@ txn_limbo_checkpoint(const struct txn_limbo *limbo, struct synchro_request *req,
 	req->confirmed_vclock = vclock;
 }
 
-/** Write a request to WAL. */
-static void
-synchro_request_write(const struct synchro_request *req)
+static int
+synchro_request_write_without_error_handling(const struct synchro_request *req)
```
Gerold103, line 468: Lets just leave it with the old name? If a function returns an `int`, it usually means it didn't do any final error handling anyway. This is a normal behaviout. And the panic-function we could rename to `synchro_request_write_or_panic()`. How does that sound?

src/box/txn_limbo.c
```diff
@@ -40,6 +40,69 @@
...
+	    limbo->volatile_confirmed_lsn == limbo->confirmed_lsn)
+		return 0;
+	if (limbo->is_in_rollback)
+		return -1;
+	if (txn_limbo_write_confirm(limbo) != 0) {
+		diag_log();
+		return -1;
+	}
+
+	limbo->confirmed_lsn = limbo->volatile_confirmed_lsn;
```
Gerold103, line 82: What if during WAL write `volatile_confirmed_lsn` was bumped again? Then you would save it as written even though it is not? Or do I miss something?

src/box/txn_limbo.c
```diff
@@ -40,6 +40,69 @@
...
+txn_limbo_worker_f(va_list args)
+{
+	(void)args;
+	struct txn_limbo *limbo = fiber()->f_arg;
+	assert(limbo == &txn_limbo);
+	while (!fiber_is_cancelled()) {
+		fiber_check_gc();
+		ERROR_INJECT_YIELD(ERRINJ_TXN_LIMBO_WORKER_DELAY);
+		if (txn_limbo_worker_bump_confirmed_lsn(limbo) != 0)
+			fiber_sleep(0);
```
Gerold103, line 99: Lets make the sleep non-zero. Otherwise in case of any long lasting WAL problems this will go into 100% CPU. Something like 1 second hardcoded would be enough. And in debug-build could make it 0.1 sec for faster tests.

src/box/txn_limbo.c
```diff
@@ -530,8 +594,8 @@ txn_limbo_read_confirm(struct txn_limbo *limbo, int64_t lsn)
 static void
 txn_limbo_confirm_lsn(struct txn_limbo *limbo, int64_t confirm_lsn)
 {
-	txn_limbo_write_confirm(limbo, confirm_lsn);
-	txn_limbo_read_confirm(limbo, confirm_lsn);
+	limbo->volatile_confirmed_lsn = confirm_lsn;
```
Gerold103, line 597: Lets add here and in other volatile LSN updates an assertion that the new LSN is > the old one.

src/box/txn_limbo.c
```diff
@@ -1163,6 +1227,10 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
 	case IPROTO_RAFT_PROMOTE:
 	case IPROTO_RAFT_DEMOTE: {
 		assert(limbo->svp_confirmed_lsn == -1);
+		/*
+		 * Do not update the `volatile_confirmed_lsn` here, this is
+		 * handled by `limbo->is_in_rollback`.
+		 */
```
Gerold103, line 1233: Hm? Does it mean in the other places the volatile LSN can go down when you assign it? I have a feeling we are going to get nasty bugs when we allow that LSN to go down in any random place and not just in one single place like here.

Gerold103 (commented):
That is a nice patch, much simpler than had I expected it to be 🔥!

My original thinking was that we would do all the limbo-control work in that worker fiber, including PROMOTE writes. It would nicely serialize everything via one fiber and we would have guaranteed zero races, maybe even without `promote_latch`. But I didn't think of trying to just have CONFIRMs there. Good idea.

src/box/txn_limbo.c
```diff
@@ -40,6 +40,69 @@
...
+		return -1;
+	if (txn_limbo_write_confirm(limbo) != 0) {
+		diag_log();
+		return -1;
+	}
+
+	limbo->confirmed_lsn = limbo->volatile_confirmed_lsn;
+	vclock_follow(&limbo->confirmed_vclock, limbo->owner_id,
+		      limbo->confirmed_lsn);
+	txn_limbo_read_confirm(limbo, limbo->confirmed_lsn);
```
Serpentian, line 85: Why do we need assigning `confirmed_lsn` value and bumping `confirmed_vclock` here, when it's all done in `txn_limbo_read_confirm`. We just need to pass value of `volatile_confirmed_lsn` before `journal_write` to it

src/box/txn_limbo.h
```diff
@@ -162,13 +162,20 @@ struct txn_limbo {
 	 */
 	struct latch promote_latch;
 	/**
-	 * Maximal LSN gathered quorum and either already confirmed in WAL, or
-	 * whose confirmation is in progress right now. Any attempt to confirm
-	 * something smaller than this value can be safely ignored. Moreover,
-	 * any attempt to rollback something starting from <= this LSN is
-	 * illegal.
+	 * Maximal LSN that gathered quorum and has already been persisted in
```
Serpentian, line 165: This assumption is broken, when we're trying to write `PROMOTE/DEMOTE`. You'll have not yet persisted state in `confirmed_vclock`, as `req_prepare` will change it

src/box/txn_limbo.c
Serpentian: Shouldn't we update `volatile_confirmed_vclock` here and then update `confirmed_vclock` in the `read_promote/demote`?

Serpentian (commented):
Thank you for the patch!

src/box/txn_limbo.h
```diff
@@ -482,6 +482,13 @@ txn_limbo_has_owner(struct txn_limbo *limbo)
 	return limbo->owner_id != REPLICA_ID_NIL;
 }
 
+/** Return whether limbo is owned by current instance. */
+static inline bool
+txn_limbo_is_owned_by_current_instance(const struct txn_limbo *limbo)
```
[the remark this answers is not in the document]
CuriousGeorgiy: See this discussion: https://github.com/tarantool/tarantool/pull/10461#discussion_r1731808895.

src/box/txn_limbo.c
CuriousGeorgiy: As I see it, the worker fiber should not interfere when the `limbo->is_in_rollback`. 

> Hm? Does it mean in the other places the volatile LSN can go down when you assign it? I have a feeling we are going to get nasty bugs when we allow that LSN to go down in any random place and not just in one single place like here.

I didn't get your point. If `limbo->is_in_rollback`, it does not accept new transactions, and we make sure the confirm is not written concurrently.

Technically, `volatile_confirmed_lsn` can go down after reading promote/demote, just like `confirmed_lsn` can, since we will start looking at another instance's transactions.

LSN cannot go down in any random place, this patch does not change the logic of promote/demote.

> Shouldn't we update `volatile_confirmed_vclock` here and then update `confirmed_vclock` in the read_promote/demote?

What's the point of introducing a `volatile_confirmed_vclock`, if we only want to asynchronously write confirms? I.e., the LSNs of transactions coming from a fixed instance.

src/box/txn_limbo.c
CuriousGeorgiy: `synchro_request_write_xc` means it throws an exception, but it doesn't. We don't have an established naming scheme for panicking functions, so I would go for @Gerold103's version.

src/box/txn_limbo.c
CuriousGeorgiy: I was looking at this code:
https://github.com/tarantool/tarantool/blob/4a866f64d64c610a3c8441835fee3d8dda5eca71/src/box/raft.c#L128-L142

I guess you are right that we should back off in this case. But let's not have configuration-dependent code.

src/box/txn_limbo.c
CuriousGeorgiy: I overlooked it, thanks!

src/box/txn_limbo.h
CuriousGeorgiy: See https://github.com/tarantool/tarantool/pull/10626#discussion_r1788125077.

src/box/txn_limbo.c
CuriousGeorgiy: Thanks! Fixed and added a test case.

src/box/txn_limbo.c
```diff
@@ -40,6 +40,70 @@
...
+
+static void
+txn_limbo_read_confirm(struct txn_limbo *limbo, int64_t lsn);
+
+/**
+ * Write a confirmation entry to the WAL. After it's written all the
+ * transactions waiting for confirmation may be finished.
+ */
+static int
+txn_limbo_write_confirm(struct txn_limbo *limbo)
```
Serpentian, line 54: Nit: it seems better to introduce `txn_limbo_write_synchro` and `txn_limbo_write_synchro_or_panic`. Since in the future not only confirm may become repeatable and we won't introduce the same function for each request.

src/box/txn_limbo.c
```diff
@@ -63,6 +128,12 @@ txn_limbo_create(struct txn_limbo *limbo)
 	limbo->confirm_lag = 0;
 	limbo->max_size = 0;
 	limbo->size = 0;
+	limbo->worker = fiber_new_system("txn_limbo_worker",
```
Serpentian, line 131: `fiber_new_system_xc`

src/box/txn.c
```diff
@@ -1237,15 +1241,6 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 	if (txn_has_flag(txn, TXN_WAIT_SYNC)) {
 		struct txn_limbo_entry *limbo_entry = txn->limbo_entry;
 		assert(limbo_entry->lsn > 0);
-		/*
```
Serpentian, line 1240: I don't like, that we change the behavior in this "refactoring" patch. We didn't introduce `wait_mode` yet.

Before this patch tx waited for `CONFIRM` to be written to disk, I suppose, the behavior should remain the same for `COMPLETE` mode for now, we should explicitly wait for `CONFIRM` to be written in `txn_commit_impl`. Now the behavior is incorrect, `req->is_complete` loop in `commit_impl` will exit, when we'll say a limbo worker to write a CONFIRM, but we won't wait for it to do so.

Serpentian (changes requested):
Looks great! It's easier to read the limbo code with your additions now. But I strongly don't like, that `complete` wait mode doesn't wait for `confirm` to be written, seems it won't be difficult to do in the current patchset.

src/box/txn.c
CuriousGeorgiy: I don't get your comment. We still have `txn_limbo_wait_complete`, which waits for the limbo entry confirmation.

Serpentian: You're right, my bad. No problem here, we're blocked until `confirm` is written

src/box/box.cc
```diff
@@ -440,7 +440,7 @@ box_check_writable(void)
 			error_set_uuid(e, "queue_owner_uuid", &r->uuid);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/txn_limbo.h
```diff
@@ -499,6 +499,13 @@ txn_limbo_has_owner(struct txn_limbo *limbo)
...
+{
+	return limbo->owner_id == instance_id;
```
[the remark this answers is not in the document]
CuriousGeorgiy: Fixed.

src/box/txn_limbo.c
CuriousGeorgiy: This is a `.c` file.

src/box/txn_limbo.c
CuriousGeorgiy: Done.

src/box/txn_limbo.c
```diff
@@ -1236,10 +1316,8 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
 	case IPROTO_RAFT_PROMOTE:
 	case IPROTO_RAFT_DEMOTE: {
 		assert(limbo->svp_confirmed_lsn == -1);
-		limbo->svp_confirmed_lsn = limbo->confirmed_lsn;
-		limbo->confirmed_lsn = req->lsn;
-		vclock_reset(&limbo->confirmed_vclock, limbo->owner_id,
-			     req->lsn);
```
Gerold103, line 1226: Why do we no longer reset `confirmed_vclock` here and below?

src/box/txn_limbo.c
```diff
@@ -40,6 +40,67 @@
...
+txn_limbo_worker_f(va_list args)
+{
+	(void)args;
+	struct txn_limbo *limbo = fiber()->f_arg;
+	assert(limbo == &txn_limbo);
+	while (!fiber_is_cancelled()) {
+		fiber_check_gc();
+		ERROR_INJECT_YIELD(ERRINJ_TXN_LIMBO_WORKER_DELAY);
+		if (txn_limbo_worker_bump_confirmed_lsn(limbo) != 0)
+			fiber_sleep(1);
```
Gerold103, line 97: Lets make it `0.01` when `#if TEST_BUILD`? (It is set in `trivial/config.h` for all non-packaged builds, including Release build used in CI.)

test/replication-luatest/gh_5295_split_brain_test.lua
```diff
@@ -161,6 +161,10 @@ g.test_bad_confirm_old_term = function(cg)
     reconnect_and_check_split_brain(cg)
 end
 
+g.before_test('test_good_confirm_old_term', function(cg)
+    cg.main:update_box_cfg{replication_synchro_timeout = 1000}
+end)
+
 -- Obsolete sync transaction confirmation might be fine when it doesn't
 -- contradict local history.
 g.test_good_confirm_old_term = function(cg)
```
Gerold103, line 170: Wouldn't it be easier to just move this single line to the beginning of the test's body? Same for the 'after test' callback. I mean, since the appearance of luatest in our life I am wondering what is the sense of these callbacks, if it is always easier to just write them directly in the test body. No exceptions. What is your opinion?

Gerold103 (commented):
Thanks for the fixes, excellent ⚡️.

src/box/txn_limbo.c
CuriousGeorgiy: I don't like the idea of having the system behave differently under debug and release build types.

test/replication-luatest/gh_5295_split_brain_test.lua
CuriousGeorgiy: I personally use the callbacks to clarify that this is setup code not related to the test logic so that a reader can focus on the test logic rather than the test setup.

src/box/txn_limbo.c
CuriousGeorgiy: Because:
1) It is always updated in `txn_limbo_read_confirm` as part of `txn_limbo_read_promote` during `txn_limbo_req_commit`. If I leave the reset here, I will end up updating it to the same value, which will trigger an assertion. 
2) AFAIC, an LSN coming from a PROMOTE/DEMOTE request must always be `vclock_follow`ed rather than `vclock_reset`.

Gerold103: > AFAIC, an LSN coming from a PROMOTE/DEMOTE request must always be vclock_followed rather than vclock_reset, since it can only monotonically grow

It grows and shrinks the same as `confirmed_vclock`. If it would only grow, then you would be able to leave the assertion that the confirmed vclock always only grows. But evidently it doesn't. No?

Besides, from what I see, PROMOTE only has `confirmed_vclock` set in `synchro_request` on checkpoint. On non-checkpoint PROMOTE this field is NULL. Which means that being sent to a remote host it won't have this vclock set, and `txn_limbo_req_commit()` won't do `vclock_copy()`. Would you please investigate why does it pass the tests, when you deleted `vclock_reset()`? That doesn't look right. Either this isn't covered by any tests, or vclock-copy happens somewhere somehow, or you should be able to leave `confirmed_vclock` also monotonically growing.

src/box/txn_limbo.c
Gerold103: That will work both in Release and Debug. `TEST_BUILD` doesn't depend on the build type. It is set for local and CI builds allowing to make them faster.

test/replication-luatest/gh_5295_split_brain_test.lua
Gerold103: The setup is related to the test logic. If the timeout is small, the test wouldn't work, would it? Which makes it a part of the test which is necessary to know. Anyway, thanks for the explanation.

Gerold103 (commented):
See my answers above ^^^.

src/box/txn_limbo.c
```diff
@@ -1236,10 +1316,8 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
 	case IPROTO_RAFT_PROMOTE:
 	case IPROTO_RAFT_DEMOTE: {
 		assert(limbo->svp_confirmed_lsn == -1);
-		limbo->svp_confirmed_lsn = limbo->confirmed_lsn;
-		limbo->confirmed_lsn = req->lsn;
-		vclock_reset(&limbo->confirmed_vclock, limbo->owner_id,
-			     req->lsn);
+		limbo->svp_confirmed_lsn = limbo->volatile_confirmed_lsn;
+		limbo->volatile_confirmed_lsn = req->lsn;
```
Serpentian, line 1320: You bump `volatile_confirmed_lsn`, which may be dangerous. I'm not sure, it's possible but let's anyway make sure, it cannot happen. Consider the following situation:

1. You failed to write `CONFIRM` in `txn_limbo_worker_f`. It'll cause `fiber_sleep(1)` and retry.
2. During `fiber_sleep` we try to write promote, which in `req_prepare` bumps `volatile_confirmed_lsn`. It's being written.
3. `txn_limbo_worker_f` wakes up and tries to write `CONFIRM` with lsn from the `promote`. There's no space in the journal queue, we're waiting.
4. `PROMOTE` is rolled back, we failed to write it, now there's enough space in queue, `CONFIRM` goes there and successfully written. It's incorrect `CONFIRM`.

The question here, whether 4th point can happen. But anyway, this bumping of `volatile_confirmed_lsn`  in `req_prepare` looks like a slippery slope to me

CuriousGeorgiy: What you described is impossible. We will never try to write CONFIRM concurrently to a PROMOTE request:
https://github.com/CuriousGeorgiy/tarantool/blob/f02774fad25d7cb67a619141a6a8cafe78887785/src/box/txn_limbo.c#L70-L71

src/box/txn_limbo.c
CuriousGeorgiy: I forgot to leave an update to my comment: I discussed this with @sergepetrenko and we came to a conclusion that `vclock_reset` here is needed solely for the case when `req->lsn` is the same as `confirmed_lsn` (`vclock_follow` prohibits passing a value equal to the current one).

So, these two snippets would actually be equivalent in `txn_limbo_req_prepare`:
```c
limbo->confirmed_lsn = req->lsn;
vclock_reset(&limbo->confirmed_vclock, limbo->owner_id, req->lsn);
```

```c
if (limbo->confirmed_lsn < req_lsn) {
    limbo->confirmed_lsn = req->lsn;
    vclock_follow(&limbo->confirmed_vclock, limbo->owner_id, req->lsn);
}
```

The second snippet is exactly what `txn_limbo_read_confirm` does:
https://github.com/tarantool/tarantool/blob/5325abd3441ecb4b589799c32ec181d88724b8a8/src/box/txn_limbo.c#L551-L554

So all I did is delayed the `confirmed_lsn` and `confirmed_vclock` updates to `txn_limbo_read_confirm`.

> It grows and shrinks the same as confirmed_vclock. If it would only grow, then you would be able to leave the assertion that the confirmed vclock always only grows. But evidently it doesn't. No?

It never shrinks. It monotonically grows — not _strictly_ monotonically though. The problem is with the lack of strictness, since a promote request with `req->lsn` equal to the `confirmed_lsn` and equal to the `confirmed_vclock` component could arrive.

> Besides, from what I see, PROMOTE only has confirmed_vclock set in synchro_request on checkpoint.

That's not correct, it is updated in `txn_limbo_read_confirm`, as I have pointed out above.

> Would you please investigate why does it pass the tests, when you deleted `vclock_reset()`?

I hope my explanation above is convincing enough. At least I managed to convince @sergepetrenko.

src/box/txn_limbo.c
CuriousGeorgiy: Sorry, my bad, I misread your comment, thought you were just pushing for this once again https://github.com/tarantool/tarantool/pull/10626#discussion_r1781690963.

Applied your suggestion.

src/box/txn_limbo.c
Serpentian: Thank you for explanation 💪

Serpentian (approved):
LGTM! Seems I have no more objections. Solid patch 💪

src/box/txn_limbo.c
```diff
@@ -37,9 +37,75 @@
...
+	assert(!limbo->is_in_rollback);
+	return txn_limbo_write_synchro(limbo, IPROTO_RAFT_CONFIRM, lsn, 0,
+				       NULL);
+}
+
+static int
+txn_limbo_worker_bump_confirmed_lsn(struct txn_limbo *limbo)
+{
+	assert(limbo->is_in_rollback ||
+	       limbo->volatile_confirmed_lsn >= limbo->confirmed_lsn);
```
Gerold103, line 68: How is the LSN comparison related to `is_in_rollback`? How does rollback being set allow `volatile_confirmed_lsn` to become `< confirmed_lsn `?

I applied this diff and the replication-luatest/ + box-luatest/ tests passed. Did I miss something here?
```Diff
index e1a01ec6c8..ce41b35c1a 100644
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -64,8 +64,7 @@ txn_limbo_write_confirm(struct txn_limbo *limbo, int64_t lsn)
 static int
 txn_limbo_worker_bump_confirmed_lsn(struct txn_limbo *limbo)
 {
-	assert(limbo->is_in_rollback ||
-	       limbo->volatile_confirmed_lsn >= limbo->confirmed_lsn);
+	assert(limbo->volatile_confirmed_lsn >= limbo->confirmed_lsn);
 	while (txn_limbo_is_owned_by_current_instance(limbo) &&
 	       limbo->volatile_confirmed_lsn > limbo->confirmed_lsn) {
 		if (limbo->is_in_rollback)
@@ -80,8 +79,7 @@ txn_limbo_worker_bump_confirmed_lsn(struct txn_limbo *limbo)
 		ERROR_INJECT_YIELD(ERRINJ_TXN_LIMBO_WORKER_DELAY);
 		txn_limbo_read_confirm(limbo, volatile_confirmed_lsn);
 	}
-	assert(limbo->is_in_rollback ||
-	       limbo->volatile_confirmed_lsn >= limbo->confirmed_lsn);
+	assert(limbo->volatile_confirmed_lsn >= limbo->confirmed_lsn);
 	return 0;
 }
 ```

**Review by Gerold103 — COMMENTED** [review]
I am so so sorry, but I have to leave one more question 😁. I am glad that you are working on this, it looks really good! I am just probably misunderstanding something again 🤦.

File: src/box/txn_limbo.c (hunk 14 above)
Comment (Gerold103) [reply src/box/txn_limbo.c:1226]: Aaah, so previously `confirmed_lsn` was already "volatile". This is why it could legally go back, and the vclock too. But now it is always updated only on disk. So it is either monotonic, or we get a split brain. Beautiful!

> That's not correct, it is updated in txn_limbo_read_confirm, as I have pointed out above.

I didn't speak about update. I was wondering about `synchro_request`. In this object it is only set on checkpoint `PROMOTE`. Which means `confirmed_vclock` is never sent around besides on join probably. But it is irrelevant now. I understand. Thanks for the explanation.

Comment (Gerold103) [reply src/box/txn_limbo.c:1226]: Btw, does this all mean `svp_confirmed_lsn` is not needed anymore? For example, look at this:
```C
	case IPROTO_RAFT_PROMOTE:
	case IPROTO_RAFT_DEMOTE: {
		assert(limbo->svp_confirmed_lsn == -1);
		limbo->svp_confirmed_lsn = limbo->volatile_confirmed_lsn;
		limbo->volatile_confirmed_lsn = req->lsn;
		txn_limbo_update_system_spaces_is_sync_state(
			limbo, req, /*is_rollback=*/false);
		break;
	}
```
We could as well completely drop `svp_confirmed_lsn`, and on promotion failure we could restore `limbo->volatile_confirmed_lsn = limbo->confirmed_lsn`.

CuriousGeorgiy: > Btw, does this all mean svp_confirmed_lsn is not needed anymore? For example, look at this

No, imagine the following scenario the current limbo owner:
```
Initially: volatile_confirmed_lsn = 1, confirmed_lsn = 1
TX commit: -> volatile_confirmed = 2, confirmed_lsn = 1
Limbo fiber: doesn't get woken up for a while
Promote request from a new leader: volatile_confirmed_lsn = 1, confirmed_lsn = 1
Promote request fails: volatile_confirmed_lsn = confirmed_lsn = 1, confirmed_lsn = 1
```

But after the promote request fails `volatile_confirmed_lsn` is supposed to return to the value 2.

src/box/txn_limbo.c
CuriousGeorgiy: I originally thought that the `volatile_confirmed_lsn` could possibly go lower than `confirmed_lsn` during the promote/demote request execution, but after our discussion it seems like the confirmed lsns can only monotonically grow, so the invariant `volatile_confirmed_lsn >= confirmed_lsn` always holds.

Applied your diff.

Gerold103 (approved):

PR #10630, opened 2024-09-30
memtx: itroduce `exhaust_after` and `step_count` tree iterator members

mkostoevr:
The EQ/REQ tree iterators in the memtx read view perform comparison of the search key with each next tuple using the `key_data` member. The thing is, since a read view of a tree is immutable, we can just perform the second lookup to find the last tuple and only compare tuple pointers on each step. This will reduce the total overhead of iteration over tuples with huge field count, because the equal key comparison is expensive for them in read views now since we can't use the tuple format.

The first step towards this is introducing a new `exhaust_after` field to be used instead of the `key_data` in case of partial key search, and `step_count` for heuristics on whether to optimize the iteration or not.

Part of tarantool/tarantool-ee#921

NO_DOC=internal
NO_TEST=internal
NO_CHANGELOG=internal

### DIFF
--- a/src/box/memtx_tree.cc
+++ b/src/box/memtx_tree.cc
@@ -44,6 +44,7 @@
 #include "trivia/util.h"
 #include "tt_sort.h"
 #include <small/mempool.h>
+#include <math.h>
 
 /**
  * Struct that is used as a key in BPS tree definition.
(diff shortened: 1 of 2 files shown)


src/box/memtx_tree.cc
```diff
@@ -2271,6 +2271,10 @@ struct tree_read_view_iterator {
 	 * is exhausted - pagination relies on it.
 	 */
 	struct memtx_tree_data<USE_HINT> *last;
+	/** The last element of the iteration for EQ and REQ iterators. */
+	struct memtx_tree_data<USE_HINT> *exhaust_after;
```
locker, line 2275: `end` may be a better name (but then we'd have to make it point to the next tuple after the last one to match the STL naming conventions)

locker (approved):

PR #10605, opened 2024-09-23
ci: bump clang version to 19

ligurio:
Docker image: https://github.com/tarantool/infra/pull/206
Tarantool EE: https://github.com/tarantool/tarantool-ee/pull/922
Depends on https://github.com/tarantool/tarantool/issues/10631

NO_CHANGELOG=ci
NO_DOC=ci
NO_TEST=ci

Issue #10228, opened 2024-07-11
Rework integration with Undefined Behavior sanitizer


ligurio:
The commit 366cb668fb5c596fba522003a2e45aefdcd0d599 introduced integration with Undefined Behavior sanitizer [^1].
However, many UBSan checks were disabled globally for a whole project [^2]: 

> The patch makes it possible to turn the sanitizer on and catch
> UBs. The only supported UB so far is alignment check. Other types
> can be added gradually, along with fixing bugs which they find.

Due to global suppressions, we missed a bug #10143 reported by OSS Fuzz project, but this issue can be found by our unit test `swim_proto.test` with fewer efforts, see repro steps [^3].

I propose to enable all supported UBsan checks (see a full list in [^1]) and fix found issues or disable UBSan checks for particular functions with `__attribute__((no_sanitize("nonnull-attribute")))`.

[^1]: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
[^2]: https://github.com/tarantool/tarantool/blob/b0b32bff022caeffd64ca687baab908f25fa2fcb/cmake/compiler.cmake#L224-L270
[^3]: https://github.com/tarantool/tarantool/issues/10143#issuecomment-2222668971

Depends on:

- [x] https://github.com/tarantool/tarantool/issues/10699
- [x] https://github.com/tarantool/tarantool/issues/10631

ligurio:
How we did it in LuaJIT - https://github.com/tarantool/luajit/commit/b9ff5ae8ea60516f630a380948bfd140c237385a

ligurio:
https://github.com/tarantool/tarantool/issues/10742
https://github.com/tarantool/tarantool/issues/10741
https://github.com/tarantool/tarantool/issues/10740

PR #10632, opened 2024-09-30
cmake: fix build with GCC's AddressSanitizer

ligurio:
GNU GCC compiler has AddressSanitizer support since 4.8.0 [1], but it was unsupported in tarantool's build. The patch fixes a build by GNU GCC with enabled AddressSanitizer.

1. https://gcc.gnu.org/gcc-4.8/changes.html

NO_CHANGELOG=build
NO_DOC=build
NO_TEST=build


grafin (approved):
Thanks for the patch! LGTM, I've checked it with gcc 13.3.0, it works. Would be great if you could fix UBSAN for gcc too).

Currently it fails with:

```
CMake Error at cmake/compiler.cmake:228 (message):
  Undefined behaviour sanitizer only available for clang
  src/CMakeLists.txt:4 (enable_tnt_compile_flags)
```

But as far as I googled - UBSAN works with gcc starting with version 4.9.x.

ligurio:
> Thanks for the patch! LGTM, I've checked it with gcc 13.3.0, it works. Would be great if you could fix UBSAN for gcc too).

Sure, I'll do the same for UBSan but later and in another PR (in scope of #10228 and after merging of #10605).

Buristan (commented):
Hi, Sergey!
Thanks for the patch!
LGTM with the one question below:

I see the following lines regarding static build and Clang 16:
https://github.com/tarantool/tarantool/blob/47295fdebc1910b990c391c26fe92918a9c0c001/cmake/CheckDependencies.cmake#L33-L36

Are they Clang-specific, or should we update them for static build with GCC too?

ligurio:
> Are they Clang-specific, or should we update them for static build with GCC too?

No, seems `libresolv` is not needed. But `libasan` must be added to `ALLOWLIST`:

```diff
--- a/cmake/CheckDependencies.cmake
+++ b/cmake/CheckDependencies.cmake
@@ -34,6 +34,9 @@ elseif(UNIX)
     if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND ENABLE_ASAN)
         set(ALLOWLIST ${ALLOWLIST} libresolv)
     endif()
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND ENABLE_ASAN)
+        set(ALLOWLIST ${ALLOWLIST} libasan)
+    endif()
 else()
     message(FATAL_ERROR "Unknown platform")
 endif()
```

Otherwise build has failed.

Buristan (approved):
Thanks for the updates!
LGTM!

Issue #10508, opened 2024-09-02
Iproto request for batch insertions in Arrow format


Gumix:
For the nodes to take advantage of batch insertions during WAL recovery and replication, batch insertions should be represented differently in WAL: the request should store the tuples in columnar format instead of our usual row-based format.

At the moment the target engine is `memcs`, but the new request could be supported by `memtx` and `vinyl` too.

See also https://github.com/tarantool/tarantool-ee/issues/820

PR #10638, opened 2024-10-01
third_party: fix `nanoarrow` install directory

Gumix:
Currently, in case of in-source build, the nanoarrow library is compiled in `<tarantool_git>/third_party/nanoarrow/`, although it is better to compile it in `<tarantool_git>/build/nanoarrow/`, because the `build` directory is ignored by `.gitignore`.

Follow-up #10508


ImeevMA (approved):

Totktonada (approved):
I just made the same fixup, verified locally and was about to send it. Opened the issue and found the backlink to this PR. Thanks!

PR #10639, opened 2024-10-02
test: disable fiber slice check in replica_apply_order test

locker:
Since commit e19bca5a74e8 ("box: check fiber slice in generic implementation of index count"), Vinyl's version of `index.count()` checks the fiber slice. As a result, the test may fail if it runs under a heavy load:

```
| @@ -94,6 +94,7 @@
|      end
|  end;
|   | ---
| + | - error: fiber slice is exceeded
|   | ...
|  -- Verify that at any moment max index is corresponding to amount of tuples,
|  -- which means that changes apply order is correct
```

Let's set the max fiber slice to a big value to avoid that.

### DIFF
--- a/test/replication/replica_apply_order.result
+++ b/test/replication/replica_apply_order.result
@@ -75,6 +75,10 @@ test_run:cmd("switch replica1")
  | ---
  | - true
  | ...
+-- index.count() checks fiber slice (gh-10553)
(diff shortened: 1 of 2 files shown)


nshy (approved):

locker:
Cherry-picked to 2.11 and 3.2.

PR #10644, opened 2024-10-03
index_def: make index_def_new non-failing

drewdzzz:
Function `index_def_new` can fail only due to failure of identifier
validation and due to failure of `key_def_merge` helper. The helper
`key_def_merge` can fail only because of OOM or if multikey validation
has failed.

Firstly, we never need to check identifier in `index_def_new` because we
always do it earlier actually (except for cases we use string literals
for internal indexes, but we don't need to check them anyway), so let's
remove this check. Secondly, `key_def_merge` cannot fail because of
multikey validation in `index_def_new` because the secondary index is
already validated and primary index cannot have multikey path at all.
So here it can fail only because of allocation failure - let's simply
make `key_def_merge` panic on OOM so that `index_def_new` will become
non-failing.

Closes tarantool/security#130

### DIFF
--- a/src/box/index_def.c
+++ b/src/box/index_def.c
@@ -103,10 +103,6 @@ index_def_new(uint32_t space_id, uint32_t iid, const char *name,
 	/* Use calloc to make index_def_delete() safe at all times. */
 	struct index_def *def = xcalloc(1, sizeof(*def));
 	def->name = xstrndup(name, name_len);
-	if (identifier_check(def->name, name_len)) {
-		index_def_delete(def);
-		return NULL;
-	}
 	if (space_name != NULL)
 		def->space_name = xstrdup(space_name);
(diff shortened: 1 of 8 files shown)


src/box/index_def.c
```diff
@@ -120,10 +120,6 @@ index_def_new(uint32_t space_id, uint32_t iid, const char *name,
 		if (opts->is_unique)
 			def->cmp_def->unique_part_count =
 				def->key_def->part_count;
-		if (def->cmp_def == NULL) {
-			index_def_delete(def);
-			return NULL;
```
locker, line 125: The `identifier_check` call above can't fail either because the identifier must have already been checked in `index_def_new_from_tuple()`. Let's also drop that check, update the comment to `index_def_new()` to state that it never fails, and drop `NULL` checks from all call sites while we are at it.

locker:
Cherry-picked to 3.2.

PR #10645, opened 2024-10-03
memtx: do not pass NULL to memcpy when creating gap item in MVCC

drewdzzz:
According to the C standard, passing `NULL` to `memcpy` is UB, even if it copies nothing (number of bytes to copy is 0). The commit fixes such situation in memtx MVCC.

**NB**: This problem is [already fixed](https://github.com/tarantool/tarantool/blob/37bf64b7d7f76ea3330909655da90d64d0add558/src/box/memtx_tx.c#L3343) in `master`, so the PR is going to be merged into `release/3.2`. Also, it should be cherry-picked to `2.11`.

Closes tarantool/security#129


locker (approved):

locker:
Cherry-picked to 2.11.

PR #10651, opened 2024-10-05
ci: remove Coverity workflow

Totktonada:
It doesn't work since 2023-11-18. The uploading succeeds, but the website says:

> The Coverity Build tool version is no longer supported. Please
> download the latest version for your platform from
> https://scan.coverity.com/download...

It seems, some specific toolset is installed in the `tarantool/testing:debian-buster` image and it was deprecated 11 months ago.

Recently the CI workflow starts to fail due to use of the old image with an old CMake in it:

> [  2%] Performing configure step for 'bundled-nanoarrow-project'
> -- Building using CMake version: 3.13.4
> -- Configuring incomplete, errors occurred!
> CMake Error at CMakeLists.txt:19 (cmake_minimum_required):
>   CMake 3.14 or higher is required.  You are running version 3.13.4

It is likely due to commit 49c160c28c97 ("third_party: initial import of nanoarrow").

Since nobody complains that there are no scanning results for almost a year, I think that it is OK to remove the workflow.

If someone wants to maintain Coverity scanning, I'm OK. I suggest to revisit the existing implementation: get rid of the custom container image with preinstalled Coverity toolset and use something that looks more convenient/maintainable: say, [this][1] GitHub Action.

Personally, I would better look for some open-source code analysis tools with a machine-readable output (to implement notifications and time series statistics) and a human-readable output (static HTML). This way we're not depend on some vendor and can setup everything in a way we like.

When we don't like tools we're using, it goes to be unmaintained and becomes broken soon.

[1]: https://github.com/marketplace/actions/unofficial-coverity-scan


Totktonada:
We discussed it with @ligurio recently. He said that Coverity is a really useful tool and it would be nice to have its feedback for our project. I also remembered that I regularly share something interesting from these regular scans with the team (on Saturday), but once emails gone I completely forget about it. We agreed that it has no sense to keep the non-working workflow, but it seems worthful to pay some time to attempt to fix it.

Let's consider this PR on hold for now.

Totktonada:
Sent #10673 that seems to fix the workflow.

PR #10654, opened 2024-10-07
httpc: replace ibuf_alloc with xibuf_alloc

ligurio:
There is no check for NULL for a value returned by `ibuf_alloc`, the NULL will be passed to `memcpy()` if the aforementioned function will return a NULL. The patch fixes that by replacing `ibuf_alloc` with macros `xibuf_alloc` that never return NULL.

Found by Svace.

NO_CHANGELOG=codehealth
NO_DOC=codehealth
NO_TEST=codehealth

### DIFF
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -601,7 +601,7 @@ httpc_request_io_read(struct httpc_request *req, char *buf, size_t len,
 
 		if (recv_len > remain) {
 			const size_t tocopy = recv_len - remain;
-			char *ptr = ibuf_alloc(&req->io_recv, tocopy);
+			char *ptr = xibuf_alloc(&req->io_recv, tocopy);
 			memcpy(ptr, recv + remain, tocopy);
 		}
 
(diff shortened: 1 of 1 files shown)


src/httpc.c
```diff
@@ -304,7 +304,7 @@ int
 httpc_set_body(struct httpc_request *req, const char *body, size_t size)
 {
 	ibuf_reset(&req->send);
-	char *chunk = ibuf_alloc(&req->send, size);
+	char *chunk = xibuf_alloc(&req->send, size);
 	if (chunk == NULL) {
```
oleg-jukovec, line 308: As far as I understand, we don't need the `if (chunk == NULL)` check after the change.

ligurio: Right, reverted:

```diff
diff --git a/src/httpc.c b/src/httpc.c
index c774be57af..a7abadf7b3 100644
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -304,7 +304,7 @@ int
 httpc_set_body(struct httpc_request *req, const char *body, size_t size)
 {
 	ibuf_reset(&req->send);
-	char *chunk = xibuf_alloc(&req->send, size);
+	char *chunk = ibuf_alloc(&req->send, size);
 	if (chunk == NULL) {
 		diag_set(OutOfMemory, size, "ibuf", "http request body");
 		return -1;

```

oleg-jukovec (approved):

Totktonada (approved):

PR #10661, opened 2024-10-08
bps: improve tree view tests

nshy:
On COW matras make a copy of extent while API requires only copy a block. So bps tree may miss COW requests for its block but the block is copied accidentally and the test passes. To avoid this issue let's make extent and block the same size.

The same improvement we done for bps vector in https://github.com/tarantool/tarantool-ee/pull/932.


mkostoevr (approved):
LGTM, good to see the fix here too.

locker (approved):

PR #10655, opened 2024-10-07
ctest prerequisites 1

ligurio:
Needed by https://github.com/tarantool/tarantool/pull/10216

PR #10663, opened 2024-10-08
[Backport 2.11] ctest prerequisites 1

ligurio:
Orig PR: https://github.com/tarantool/tarantool/pull/10655


test/app-luatest/gh_7479_bug_fix_uri_lib_test.lua
Totktonada, line 1: We still have `tarantoolctl` in 2.11. I think we shouldn't drop the test case in this branch.

Totktonada (approved):

PR #10664, opened 2024-10-08
box: log error that caused initial checkpoint failure

locker:
Currently, we just panic without providing any additional information if we failed to create the initial checkpoint on bootstrap. This complicates trouble shooting. Let's replace `panic()` with `say_error()` and raise the exception that caused the failure. The exception will be caught by by `box_cfg()`, which will log it and then panic.


nshy (approved):

locker:
Cherry-picked to 3.2 and 2.11.

PR #10621, opened 2024-09-26
third_party: update libcurl from 8.8.0 to 8.10.1

Col-Waltz:
Updates curl module to the version 8.10.1. The new version brings several new options, bug fixes and vulnerability fixes and improve overall stability.

All new options are set in our build script. Here is the list of added options with information about the purpose of each option and the commit hash in the curl in which this option was added:

New options from 8.8.0 to 8.10.1

CURL_USE_RUSTLS - Enables Rustls for SSL/TLS. Added in commit 
    curl/curl@ed76a23fccc1 ("cmake: add rustls")
CURL_USE_WOLFSSH - Option to use wolfSSH. Added in commit
    curl/curl@0d8fdd1c7421 ("cmake: add wolfSSH support")
CURL_USE_GSASL - Option to use libgsasl. Added in commit
    curl/curl@66bf995d1cfc ("cmake: add CURL_USE_GSASL option with detection + CI test")
CURL_DISABLE_SHA512_256 - Disables SHA-512/256 hash algorithm. Added in commit
    curl/curl@33629949488c ("build: add options to disable SHA-512/256 hash algo")
CURL_USE_LIBUV - Use libuv for event-based tests. Added in commit
    curl/curl@f7d5f47059c3 ("cmake: add support for `CURL_USE_LIBUV` option")

Old options that were not included in previous curl bumps

BUILD_EXAMPLES - Build libcurl examples. Added in curl 8.8.0 release in commit
    curl/curl@dfdd978f7c60 ("cmake: add `BUILD_EXAMPLES` option to build examples")
USE_ECH - Enables ECH support. Added in curl 8.8.0 release in commit
    curl/curl@a362962b7289 ("TLS: add support for ECH (Encrypted Client Hello)")
USE_HTTPSRR - Enables HTTPS RR support for ECH. Added in curl 8.8.0 release in commit
    curl/curl@a362962b7289 ("TLS: add support for ECH (Encrypted Client Hello)")
BUILD_STATIC_CURL - Builds curl executable with static libcurl. Added in curl 8.3.0 release in commit
    curl/curl@1199308dbc90 ("cmake: support building static and shared libcurl in one go")
CURL_DISABLE_NEGOTIATE_AUTH - Disables negotiate authentication. Added in curl 8.3.0 release in commit 
    curl/curl@e92edfbef644 ("lib: add ability to disable auths individually")
CURL_DISABLE_SRP - Disables TLS-SRP support. Added in curl 8.4.0 release in commit
[...]

PR #10672, opened 2024-10-10
build: don't depend on libidn2

Totktonada:
The recent libcurl update in commit 0919f390802f ("third_party: update libcurl from 8.8.0 to 8.10.1") reveals an old problem in our libcurl's build script: it attempts to disable the libidn2 dependency using the `-DHAVE_LIBIDN2=OFF` option instead of `-DUSE_LIBIDN2=OFF`.

It doesn't work after a recent curl building machinery change, see https://github.com/curl/curl/pull/14555

This commit fixes the problem by using `-DUSE_LIBIDN2=OFF`.

Reported in #10621


cmake/BuildLibCURL.cmake
```diff
@@ -150,8 +150,7 @@ macro(curl_build)
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_RUSTLS=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_WOLFSSH=OFF")
     list(APPEND LIBCURL_CMAKE_FLAGS "-DCURL_USE_GSASL=OFF")
-    list(APPEND LIBCURL_CMAKE_FLAGS "-DHAVE_LIBIDN2=OFF")
```
ligurio, line 153: Should we put `DHAVE_LIBIDN2` to `cmake/curl-excluded-options.txt`?

ligurio (approved):
Thanks for the patch! LGTM with a minor comment.

cmake/BuildLibCURL.cmake
Totktonada: I've verified that it is not in the `cmake -LHA` output, at least on curl `8.10.1`. It seems, now there is no point to place it neither to the passed options, nor to the ignored ones.

Col-Waltz: I agree with this changes, `USE_LIBIDN2` was set to ON by default in `curl/CMakeLists.txt` so I didn't know that we don't depend on it.

Col-Waltz (approved):
Thanks for the patch!

I agree with this changes, `USE_LIBIDN2` was set to ON by default in `curl/CMakeLists.txt` so I didn't know that we don't depend on it.

cmake/BuildLibCURL.cmake
ligurio: Thanks for explanation!

Totktonada:
Backported to `release/3.2` in PR #10676 and to `release/2.11` in PR #10677.

PR #10684, opened 2024-10-14
say: fix NULL pointer dereference in log_syslog_init

nshy:
If `opts.identity` is `NULL` and `strdup()` is failed we do `NULL` pointer dereference when reporting the error. Let's just panic if `strdup()` failed. While at it replace another `strdup()` with `xstrdup()` in this function. Our current approach is to panic on runtime OOM.

Closes tarantool/security#128


locker (approved):

locker:
Cherry-picked to 3.2 and 2.11.

PR #10688, opened 2024-10-15
config: expose configured names

### LINKED ISSUE #10308: config: expose configured instance/replicaset/group names
## Context

Any instance that is started from the YAML configuration has instance name, replicaset name and group name. However, there is no API to obtain all of them in all the situations.

<details>
<summary>Details</summary>

----

`box.info.name` (https://github.com/tarantool/doc/issues/3468) and `box.info.replicaset.name` (https://github.com/tarantool/doc/issues/3467) *usually* show configured instance and replicaset names. However, it is not so during an upgrade from an 2.x snapshot (before calling `box.schema.upgrade()`), because these values are about names written to the database. The names are written to the database after the upgrade.

A group name from the YAML configuration is not shown anywhere. It can be deduced from `config:instances()` (https://github.com/tarantool/doc/issues/4105) using an instance name.

----
</details>

There is an internal API that allows to obtain the names:

```yaml
# tarantool> config._configdata_applied:names()
---
- instance_uuid: null
  replicaset_uuid: null
  instance_name: instance-001
  replicaset_name: replicaset-001
  group_name: group-001
...
```

It shows the names from the configuration and works before `box.schema.upgrade()`.

## Proposal

Expose the names from the public API:

```yaml
# tarantool> config:info()
# or
# tarantool> config:info('v2')
# or
# tarantool> box.info.config
---
- <..status, alerts, meta..>
  hierarchy:
    instance_name: instance-001
    replicaset_name: replicaset-001
    group_name: group-001
...
```

## Motivation

A seamless upgrade sometimes requires a knowledge about instance/replicaset names of the current instance.

For example, the `crud` module [finds](https://github.com/tarantool/crud/blob/98b120ef7095fa34525ef9d335a1458a2edf0cca/crud/common/vshard_utils.lua#L7-L27) the replicaset in `vshard`'s configuration to [determine](https://github.com/tarantool/crud/blob/98b120ef7095fa34525ef9d335a1458a2edf0cca/crud/common/vshard_utils.lua#L58-L77) an appropriate username to grant permissions to the storage functions (at least to ensure that they're already exist).

----

Also tracked in TNTP-347.

Totktonada:
An instance in the YAML configuration always has a name and always corresponds to a particular replicaset and a particular group. This is forced by the hierarchical configuration format.

These names are now reported in `config:info()` and `box.info.config`.

```yaml
# tarantool> config:info()
# or
# tarantool> config:info('v2')
# or
# tarantool> box.info.config
---
- <..status, alerts, meta..>
  hierarchy:
    group: group-001
    replicaset: replicaset-001
    instance: instance-001
...
```

The `hierarchy` field is an empty table if tarantool is started with a Lua script or in the interactive mode without a script/a config.

Usually `box.info.name` equals to `box.info.config.hierarchy.instance`, but there are situations, when only one of them is reported.

`box.info.config.hierarchy.instance` corresponds to the *configured* instance name and it is reported when the YAML configuration is in use. This instance name is written to the database as well, but the write operation is postponed if the database schema version is lower then needed to support the persistent names (see https://github.com/tarantool/doc/issues/4592 for details).

A user can meet this situation while running tarantool 3.x on data from tarantool 2.x, before the upgrading procedure is performed.

On the other hand, a user may use the persistent names feature (see https://github.com/tarantool/doc/issues/3468) together with the classic approach, when tarantool is run without the configuration file, using a Lua script. In this case, `box.info.config.hierarchy` is empty, but `box.info.name` shows the name passed to `box.cfg({instance_name = <...>})` (and written to the database).

The same is applicable to `box.info.replicaset.name` and `box.info.config.hierarchy.replicaset`.

Fixes #10308


changelogs/unreleased/config-info-hierarchy.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* `config:info()` and its alias `box.info.config` now has the `hierarchy` field
+  that shows configured instance, replicaset and group names (gh-10308).
```
p7nov, line 4:
```suggestion
* `config:info()` and its alias `box.info.config` now has the `hierarchy` field
  that shows the configured instance, replica set, and group names (gh-10308).
```

p7nov (approved):

test/config-luatest/config_info_hierarchy_test.lua
```diff
@@ -0,0 +1,153 @@
...
+end)
+
+g.before_test('test_interactive', function(g)
+    g.it = it.new()
+end)
+
+g.after_test('test_interactive', function(g)
+    g.it:close()
+    g.it = nil
+end)
```
georgiy-belyanin, line 31: Nit: As I see in other tests (e.g in `enterprise-luatest`), such `before_/after_` calls surround the test on which they're applied.

```lua
g.before_test('a')
g.a = function()
g.after_test('b')
```

IMO it's a lot simpler to read the tests like this.

test/config-luatest/config_info_hierarchy_test.lua
```diff
@@ -0,0 +1,153 @@
...
+            replicaset = 'r-001',
+            instance = 'i-001',
+        }
+
+        t.assert_equals(config:info().hierarchy, exp)
+        t.assert_equals(config:info('v1').hierarchy, exp)
+        t.assert_equals(config:info('v2').hierarchy, exp)
+        t.assert_equals(box.info.config.hierarchy, exp)
+    end)
+end
```
georgiy-belyanin, line 48: Nit: probably, it's worth use something like `assert_config_info_hierarchy` accepting the server and the expected value. This function seems more versatile.

The expected value could be optional with the default value equal to existing local `exp` in `assert_configured_names`

test/config-luatest/config_info_hierarchy_test.lua
```diff
@@ -0,0 +1,153 @@
...
+-- config:info() if tarantool is started from a Lua script,
+-- without the declarative configuration.
+g.test_script = function(g)
+    g.server:exec(function()
+        local config = require('config')
+
+        t.assert_equals(config:info().hierarchy, {})
+        t.assert_equals(config:info('v1').hierarchy, {})
+        t.assert_equals(config:info('v2').hierarchy, {})
+        t.assert_equals(box.info.config.hierarchy, {})
```
georgiy-belyanin, line 140: Nit: such `assert_config_info_hierarchy(g.server, {})` could be used here, for example.

test/config-luatest/config_info_hierarchy_test.lua
georgiy-belyanin, line 1: Minor: IMO it's worth testing out the values different from `g-001/r-001/i-001` (e.g. non-master replica, second replicaset), since I can easily imagine the wrong implementation of this using the code from the config. There area lot of different variables meaning different instances and replicasets and after the modification it can be broken. 

Probably, this is too much for such a small patch though.

georgiy-belyanin (changes requested):
Thank you for the patch! Such a nice and useful field. Please, consider my comments below

test/config-luatest/config_info_hierarchy_test.lua
Totktonada: The idea behind splitting these hooks from the main testing code is that setup/teardown code in not very interesting, when looking into a test case logic. For example, we have a preamble that defines a standard setup/teardown hooks for `cluster`:

```lua
g.before_all(cluster.init)                                                                                       
g.after_each(cluster.drop)                                                                                       
g.after_all(cluster.clean)  
```

And we just forget about all this machinery, when writing/looking a test case. I think that's nice.

OTOH, I understand that if you see tightly coupled things splitted over the file, it distracts a bit.

I don't like a direct solution: placing the hooks in-between the test cases like these are objects of the same kind.

However, I can write it in a more structured way:

```lua
-- Add a test case with before/after_test hooks in a convenient
-- way.
--
-- g:add_test_case('test_foo', {
--     before = function(g)
--         <...>
--     end,
--     function(g)
--     end,
--     after = function(g)
--         <...>
--     end,
-- })
g.add_test_case = function(g, name, funcs)
    if funcs.before ~= nil then
        g.before_test(name, funcs.before)
    end
    assert(funcs[1] ~= nil)
    g[name] = funcs[1]
    if funcs.after ~= nil then
        g.after_test(name, funcs.after)
    end
end

<...>

-- Verify that there are no instance/replicaset/group names in
-- config:info() if tarantool is started from a Lua script,
-- without the declarative configuration.
g:add_test_case('test_script', {
    before = function(g)
        g.server = server:new({alias = 'classic_server'})
        g.server:start()
    end,
    function(g)
        g.server:exec(function()

            t.assert_equals(config:info().hierarchy, {})
            t.assert_equals(config:info('v1').hierarchy, {})
            t.assert_equals(config:info('v2').hierarchy, {})
            t.assert_equals(box.info.config.hierarchy, {})
        end)
    end,
    after = function(g)
        g.server:stop()
        g.server = nil
    end,
})
```

I like it more.

OTOH, I can just define one `after_each` and start needed processes from within a test case:

```lua
diff --git a/test/config-luatest/config_info_hierarchy_test.lua b/test/config-luatest/config_info_hierarchy_test.lua
index 6cd333ac25..562c47c858 100644
--- a/test/config-luatest/config_info_hierarchy_test.lua
+++ b/test/config-luatest/config_info_hierarchy_test.lua
@@ -11,23 +11,15 @@ g.before_all(cluster.init)
 g.after_each(cluster.drop)
 g.after_all(cluster.clean)
 
-g.before_test('test_script', function(g)
-    g.server = server:new({alias = 'classic_server'})
-    g.server:start()
-end)
-
-g.after_test('test_script', function(g)
-    g.server:stop()
-    g.server = nil
-end)
-
-g.before_test('test_interactive', function(g)
-    g.it = it.new()
-end)
-
-g.after_test('test_interactive', function(g)
-    g.it:close()
-    g.it = nil
+g.after_each(function(g)
+    if g.server ~= nil then
+        g.server:stop()
+        g.server = nil
+    end
+    if g.it ~= nil then
+        g.it:close()
+        g.it = nil
+    end
 end)
 
 local function assert_configured_names(server)
@@ -96,11 +88,6 @@ g.test_2_11_1 = function(g)
     --
     -- Otherwise it is not possible to connect to the instance
     -- from the test and the ready check fails on the startup.
-    --
-    -- NB: If this sequence of steps is failed in the middle (so
-    -- we don't reach g.it:close()), the g.it handle is collected
-    -- by GC and the process is killed. At worst, it is performed
-    -- on luatest termination.
     g.it = it.new()
     g.it:roundtrip("fio = require('fio')")
     g.it:roundtrip(('fio.chdir(%q)'):format(data_dir))
@@ -131,6 +118,9 @@ end
 -- config:info() if tarantool is started from a Lua script,
 -- without the declarative configuration.
 g.test_script = function(g)
+    g.server = server:new({alias = 'classic_server'})
+    g.server:start()
+
     g.server:exec(function()
 
@@ -144,6 +134,7 @@ end
 -- Same as previous, but for tarantool started in the interactive
 -- mode, without a YAML config or a Lua script.
 g.test_interactive = function(g)
+    g.it = it.new()
     g.it:roundtrip("config = require('config')")
 
     t.assert_equals(g.it:roundtrip('config:info().hierarchy'), {})
```

I usually use this way. Don't know why I forgot about it this time.

I'm going to fix it this way.

changelogs/unreleased/config-info-hierarchy.md
Totktonada: Thanks! Fixed.

test/config-luatest/config_info_hierarchy_test.lua
Totktonada: Done.

test/config-luatest/config_info_hierarchy_test.lua
Totktonada: Implemented.

test/config-luatest/config_info_hierarchy_test.lua
Totktonada: Good idea. I've implemented it.

test/config-luatest/config_info_hierarchy_test.lua
Totktonada: I wouldn't bother, but since someone worrying, OK. Added as a separate test case.

georgiy-belyanin: It looks a lot more complicated than I've imagined. Though, I suppose it's good. Thanks!

src/box/lua/config/init.lua
```diff
@@ -498,15 +498,30 @@ function methods.reload(self)
...
+    if self._configdata_applied == nil then
+        return {}
+    end
+
+    local names = self._configdata_applied:names()
+    return {
+        group = names.group_name,
+        replicaset = names.replicaset_name,
+        instance = names.instance_name,
+    }
```
grafin, line 511: nit: right now `_config_applied:names()` will always return a table, is there a test, that future changes will never break this assumption? I'm mostly worried, because `_configdata_applied` is an internal structure and in future someone may change its behavior in a way that will break this code. We could just add something like `if names == nil then names = {} end` here and it shouldn't break ever)

grafin (approved):
Thanks for the patch! I don't see any problems with it, except one nit, not really directly connected with this patch itself.

test/config-luatest/config_info_hierarchy_test.lua
georgiy-belyanin: Thank you. I haven't thought of this since I consider even the initialization code as the closely related to the test as different configurations can be passed. You've suggested a nice approach, hope I won't forget about it when I'll need it. A new one way to write it is also good.

src/box/lua/config/init.lua
Totktonada: It would be dead code and it would create a false impression about the `:names()` contract.

An instance always have group/replicaset/instance names by construction. If `<configdata object>:names()` breaks this assumption, we have to change `applier/box_cfg.lua` and `applier/sharding.lua`, which also lean on it.

I would leave it as is.

test/config-luatest/config_info_hierarchy_test.lua
```diff
@@ -0,0 +1,192 @@
...
+        'g1-r2-i2',
+        'g2-r1-i1',
+        'g2-r1-i2',
+        'g2-r2-i1',
+        'g2-r2-i2',
+    }) do
+        assert_hierarchy(cluster[instance], {
+            group = instance:match('(g[12])%-r[12]%-i[12]'),
+            replicaset = instance:match('(g[12]%-r[12])%-i[12]'),
+            instance = instance,
```
georgiy-belyanin, line 104: Beautiful one.

georgiy-belyanin (approved):
Thank you for the fixes! LGTM

src/box/lua/config/init.lua
grafin: Thanks for the explanation!

PR #10692, opened 2024-10-15
build: bump clang version to 19

ligurio:
The patch bumps the Clang version from 16 to 19 in workflows that are intended to use the most modern Clang version. Many things were changed since version 16, see release notes in [1], [2], [3].

The `-Ofast` command-line option has been deprecated in Clang 19, see [3]. It is advised to switch to `-O3 -ffast-math` if the use of non-standard math behavior is intended, and -O3 otherwise. The patch replaces `-Ofast` with `-O3 -ffast-math` to suppress a message below in a build log:

NO_WRAP
clang-19: warning: argument '-Ofast' is deprecated; use '-O3 -ffast-math' for
	  the same behavior, or '-O3' to enable only conforming optimizations [-Wdeprecated-ofast]
NO_WRAP

The bump had needed an update of Dockerfile used for building the Docker image used in the aforementioned workflows, see pull request in [4].

1. https://releases.llvm.org/17.0.1/tools/clang/docs/ReleaseNotes.html
2. https://releases.llvm.org/18.1.6/tools/clang/docs/ReleaseNotes.html
3. https://releases.llvm.org/19.1.0/tools/clang/docs/ReleaseNotes.html
4. https://github.com/tarantool/infra/pull/206

NO_CHANGELOG=build
NO_DOC=build
NO_TEST=build

(cherry picked from commit 11145f6a82b1753e81baf8a439b29962dfef4221)


ligurio:
Found mem leak: https://github.com/tarantool/tarantool/issues/10698

Issue #10691, opened 2024-10-15
config: separate schema code and data validators


mandesero:
### Overview

Currently, validation annotations (i.e. `validate` functions) are mixed in with the configuration schema (that defines the structure and types of the data). While it can be useful to view both the schema and validation logic together, more often than not, the validation code takes up a significant amount of space and makes the schema itself difficult to read.

#### Separate Validator Functions

Create a dedicated `validators.lua` file where validation functions are defined separately:

```lua
-- validators.lua
local M = {}

M['iproto.advertise.client'] = function(data, w)
	-- <...>
end

return M
```

#### Proposal 1: Use `schema.new` to inject validators

In this proposal, the `schema.new` function will be modified to automatically inject the necessary `validate` annotations from the `validators` module into the schema nodes during the schema preprocessing step (i.e., when copying the schema nodes).

```lua
-- instance_config.lua
local validators = require('internal.config.validators')

return schema.new('instance_config', {
	-- <..schema without validate = function() end..>
}, {
	validators = validators,
})
```

#### Proposal 2: Use a wrapper to add validators

Once [#10636](https://github.com/tarantool/tarantool/issues/10636) is resolved, we can reuse the wrapper that adds descriptions to schema fields and extend it to also set validators. This approach will apply changes at the `instance_config` level without modifying the `schema` implementation:

```lua
-- instance_config.lua
local validators = require('internal.config.validators')

local function set_validator(schema_obj, ctx)
	-- <...>
	schema_obj.validate = validators[<..field_path..>]
	-- <...>
end

local function <..wrapper..>(schema, ctx)
	set_description(schema, ctx)
	set_validator(schema, ctx)
	-- <...>
end

return schema.new('instance_config', <..wrapper..>({
	-- <..schema without validate = function() end..>
}), <...>)
```

-----
Jira: [TNTP-263](https://jira.vk.team/browse/TNTP-263)

georgiy-belyanin:
I'd like to discuss the suggested approach a bit.

Implicit relations in the code make it harder to read. One of the strongest parts of the existing schema definition is that it's easy to see what we expect in the field and how it's validated.

E.g. take a look at the way we define `iproto.advertise.client`:
```diff
             client = schema.scalar({
                 type = 'string',
                 default = box.NULL,
-                validate = validate_uri_field(false, true),		
             }),
```

With explicit validator field it's obvious we expect URI here. Without it's unclear from quite a few variables what kind of string/number we expect.

[...]

PR #10694, opened 2024-10-15
config: separate schema code and data validators

mandesero:
This patch separates the instance config schema and data validators. The validation logic is moved to `box/lua/config/validators.lua`, improving schema readability by decoupling validation annotations from the schema structure.

The `validators.lua` file now contains the validation logic, as demonstrated in the following snippet:

```lua
-- validators.lua
local M = {}

<...>

M['app'] = function(app, w)
	if app.file ~= nil and app.module ~= nil then
		w.error('Fields file and module cannot appear at the same time')
	end
end

<...>

return M
```

This patch uses **Proposal 1** from #10691. 

----

Closes https://github.com/tarantool/tarantool/issues/10691


src/box/lua/config/instance_config.lua
```diff
@@ -2512,19 +2168,7 @@ return schema.new('instance_config', add_descriptions(schema.record({
...
-    --
-    -- * Missed fields are not validated.
-    -- * The outmost instance config record is marked with the
-    --   'scope' annotation (when the instance config is part of
-    --   the cluster config), but this annotation is not easy to
-    --   reach from the 'validate' function of a nested schema
-    --   node.
-    validate = validate_outmost_record,
-})), {
+}, {})), {
```
Totktonada, line 2171: `, {}` is redundant, just drop it.

src/box/lua/config/instance_config.lua
```diff
@@ -1092,21 +948,7 @@ return schema.new('instance_config', add_descriptions(schema.record({
...
-                    if data.password ~= nil and data.login == nil then
-                        w.error('Password cannot be set without setting login')
-                    end
-                    -- If a params is set, an uri must also be specified.
-                    if data.params ~= nil and data.uri == nil then
-                        w.error('Params cannot be set without setting uri')
-                    end
-                end,
-            }),
+            }, {}),
```
Totktonada, line 951: Same, `, {}` is redundant, we can just drop the argument.

Totktonada: The same for other such places.

mandesero:
There are several duplicate functions in `validators.lua`, `schema.lua`, and `instance_config.lua`. I suggest considering the creation of an additional `utils` module to reorganize the code and adhere to the DRY.

One idea I had was to collect the helper methods within `validators.lua` like this:

```lua
-- validators.lua

<...>

return {
    validators = M,
    methods = methods,
}
```

However, I've started to dislike this approach for several reasons. One key issue is that it limits the usability of these methods in other modules, such as `schema.lua`.

I'm also thinking about moving most of the methods from `instance_config.lua` to a new module. That way `schema.lua` can stay focused solely on schema-related code. @Totktonada, what do you think about this?

src/box/lua/config/instance_config.lua
mandesero: Dropped.

src/box/lua/config/instance_config.lua
mandesero: Dropped.

src/box/lua/config/instance_config.lua
```diff
@@ -119,32 +92,21 @@ end
...
     -- Enterprise Edition.
     --
     -- This order is consistent with data type validation, which
     -- is also performed before the EE check.
-    schema_node.validate = chain2(
-        schema_node.validate,
-        enterprise_edition_validate)
     schema_node.apply_default_if = chain2(
-        schema_node.apply_default_if,
+        schema_node.applygit_default_if,
```
Totktonada, line 104: Typo?

mandesero: Oops, looks like some console input slipped in by accident.
Fixed it, thanks!

Totktonada:
We discussed it a bit with @mandesero and it seems that we can add new modules to `src/box/lua/config/utils`: for example, `funcutils` and `network`.

src/box/lua/config/instance_config.lua
```diff
@@ -98,127 +73,24 @@ local function enterprise_edition_apply_default_if(_data, _w)
...
     -- check, whether the option is to be used with Tarantool
     -- Enterprise Edition.
     --
     -- This order is consistent with data type validation, which
     -- is also performed before the EE check.
-    schema_node.validate = chain2(
-        schema_node.validate,
-        enterprise_edition_validate)
-    schema_node.apply_default_if = chain2(
+    schema_node.apply_default_if = funcutils.chain2(
```
Totktonada, line 87: I like the idea of splitting this patchset to three commits, but if we do, we should move the function (and fix the caller code) in the appropriate commits (the first one for `chain2` and the second one for `uri_is_suitable_to_connect`).

src/box/lua/config/utils/funcutils.lua
```diff
@@ -0,0 +1,21 @@
+local methods = {}
```
Totktonada, line 1: Sorry for nitpicking, but I'm trying to use the term 'method' in context of Lua, when referring to a function that accepts a particular object as the first argument (`self`), and this function stored inside a table or its `__index` table that allows to call it using a colon (`obj:mymethod(<...>)`).

There are several frequent styles how a Lua module is organized.

```lua
local M = {}

function M.foo(<...>)
    <...>
end

return M
```

Here `M` stands for 'module', not 'methods'. At least, in my understanding.

Another valid style:

```lua
local mymodule = {}

function mymodule.foo(<...>)
    <...>
end

return mymodule
```

And the third one:

```lua
local function foo(<...>)
    <...>
end

return {
    foo = foo,
}
```

I have no strict preference here. If all the other modules in this directory use the same style, let's use it here as well. If not, choose any of them.

But I think that 'methods' doesn't fit here good.

mandesero: > Sorry for nitpicking

No problem, I refactored it to use the third option.

src/box/lua/config/instance_config.lua
mandesero: > but if we do, we should move the function (and fix the caller code) in the appropriate commits

Yes, it should have been done that way. I've fixed it.

src/box/lua/config/instance_config.lua
```diff
@@ -103,65 +76,21 @@ end
...
 
     -- Perform a domain specific validation first and only then
     -- check, whether the option is to be used with Tarantool
     -- Enterprise Edition.
     --
     -- This order is consistent with data type validation, which
     -- is also performed before the EE check.
-    schema_node.validate = funcutils.chain2(
-        schema_node.validate,
-        enterprise_edition_validate)
```
Totktonada, line 131: The comment seems to contradict the code: the order of the checks is changed. Before: type, validate, validate_ee. After: type, validate_ee, validate.

Sadly, it seems that we have no test case that targets a situation with an incorrect EE option, when we're run the config validation on CE.

Also, the comment was right before the relevant code, now it seems separated from it.

Next, we unconditionally set `schema_node.validate` assuming that the schema has no such a field at this point of execution. I think it would be good to add an assertion for it, at least.

However, it seems that we should add the `enterprise_edition_validate` at the end of the validation functions chain.

I can't propose some good solution here right now.

src/box/lua/config/instance_config.lua
```diff
@@ -2029,30 +1736,10 @@ return schema.new('instance_config', schema.record({
...
-                if data == nil or scope == nil then
-                    return
-                end
-                if scope ~= 'global' then
-                    w.error('sharding.rebalancer_disbalance_threshold should '..
-                            'be a defined in global scope')
-                end
-            end,
         }),
         rebalancer_max_receiving = schema.scalar({
```
Totktonada, line 1744: There are a couple of options in the `sharding` section, where the validation function remains in `instance_config.lua`. For example, `sharding.rebalancer_max_receiving`.

src/box/lua/config/instance_config.lua
Totktonada: Hm. My guess was incorrect. You add a validation function to the beginning of the chain:

```lua
schema.validate = funcutils.chain2(validate, schema.validate)
```

However, it means that if we have a validation function for `schema.set`, the instance_config schema provided function is executed before `validate_no_repeat`.

Totktonada: > Sadly, it seems that we have no test case that targets a situation with an incorrect EE option, when we're run the config validation on CE.

This is not right.

`test_config_etcd_no_prefix` from `test/config-luatest/instance_config_schema_test.lua` detects it.

Totktonada: OK, let's structurize thoughts.

I don't like that we place the external validation function in the middle of the validation chain:

```lua
schema.validate = funcutils.chain2(validate, schema.validate)
```

This way type checks are performed first, then the external function, then one that is set right in the schema.

It seems logical for me to place the external validator at the end.

However, if we do it, we have to solve the order of domain-specific-validate and enterprise-edition-validate functions. There are several ways to do so.

1. Support `enterprise_edition` (boolean) annotation in the schema.lua module.

   From the schema.lua side it seems too specific. We need it in tarantool, but I see no sense to provide it for modules/application. So, it looks more like instance_config's deal.
2. Include the `enterprise_edition_validate` function into `validators.lua` in a straightforward way:

   ```lua
   M['config.etcd'] = funcutils.chain2(function(data, w)
       <...>
   end, enterprise_edition_validate)
   ```

   ```lua
   M['audit_log'] = enterprise_edition_validate
   ```

   I don't like that we lost connection to the `enterprise_edition` annotation and the `apply_default_if` annotations. If we go this way, we need to add some test that verifies it. However, it is something that should be expressed in the code in some quite obvious way, I think.
3. Assign a default validation function to all the nodes, but check, whether the `enterprise_edition` annotation is in the schema node inside the validation function (check `w.schema.enterprise_edition`).

   At first, it seems inaccurate to assign a validation function, when it is known that it always succeeds for the given node.

   And it also likely break our `schema.validate == validate_no_repeat` check that we use in the JSON Schema generation (at least, needs additional logic here).
4. Extend the `validators` table API: instead of just a table, expose a resolver function from it, where we can analyze the schema node and decide about needed validation functions.

   ```lua
   -- validators.lua
   local registry = {}

   registry['config.etcd'] = function(data, w)
       <...>
   end

   <...>

   return {
       gen = function(w)
           local path = table.concat(w.path, '.')
           local v = funcutils.chain2(w.schema.validate, registry[w.path])
           if w.schema.enterprise_edition then
               v = funcutils.chain2(v, enterprise_edition_validate)
           end
           return v
       end,
   }
   ```

   This way it is not like the `validators` table, but I would name it somehow like `validate_gen`.

   Seems like a good variant. We increase a power of the `schema.lua` API in a general way to cover our real needs.
5. Postprocess instance config schema based on the `enterprise_edition` annotation right in `instance_config.lua`.

   <details>
   <summary>PoC</summary>

   ```diff
   diff --git a/src/box/lua/config/instance_config.lua b/src/box/lua/config/instance_config.lua
   index 7254b839c7..824d94ad5e 100644
   --- a/src/box/lua/config/instance_config.lua
   +++ b/src/box/lua/config/instance_config.lua
   @@ -76,18 +76,6 @@ end
    -- Available only in Tarantool Enterprise Edition.
    local function enterprise_edition(schema_node)
        schema_node.enterprise_edition = true
   -    schema_node.validate = enterprise_edition_validate
   -
   -    -- Perform a domain specific validation first and only then
   -    -- check, whether the option is to be used with Tarantool
   -    -- Enterprise Edition.
   -    --
   -    -- This order is consistent with data type validation, which
   -    -- is also performed before the EE check.
   -    schema_node.apply_default_if = funcutils.chain2(
   -        schema_node.apply_default_if,
   -        enterprise_edition_apply_default_if)
   -
        return schema_node
    end
    
   @@ -321,7 +309,43 @@ local function feedback_apply_default_if(_data, _w)
        return box.internal.feedback_daemon ~= nil
    end
    
   -return schema.new('instance_config', schema.record({
   +local function add_enterprise_edition_checks(schema)
   +    local res = table.copy(schema)
   +
   +    if res.enterprise_edition then
   +        -- Perform a domain specific validation first and only
   +        -- then check, whether the option is to be used with
   +        -- Tarantool Enterprise Edition.
   +        --
   +        -- This order is consistent with data type validation,
   +        -- which is also performed before the EE check.
   +        res.validate = funcutils.chain2(res.validate,
   +            enterprise_edition_validate)
   +        res.apply_default_if = funcutils.chain2(
   +            res.apply_default_if,
   +            enterprise_edition_apply_default_if)
   +    end
   +
   +    -- luacheck: ignore 542 empty if branch
   +    if res.type == 'record' then
   +        local new_fields = {}
   +        for k, v in pairs(res.fields) do
   +            new_fields[k] = add_enterprise_edition_checks(v)
   +        end
   +        res.fields = new_fields
   +    elseif res.type == 'map' then
   +        res.key = add_enterprise_edition_checks(res.key)
   +        res.value = add_enterprise_edition_checks(res.value)
   +    elseif res.type == 'array' then
   +        res.items = add_enterprise_edition_checks(res.items)
   +    else
   +        -- Scalar -- no recursion.
   +    end
   +
   +    return res
   +end
   +
   +local internal_schema = schema.new('instance_config_internal', schema.record({
        config = schema.record({
            reload = schema.enum({
                'auto',
   @@ -2183,6 +2157,13 @@ return schema.new('instance_config', schema.record({
            default = false,
        }),
    }), {
   +    _extra_annotations = {
   +        validators = validators,
   +    },
   +})
   +
   +internal_schema = add_enterprise_edition_checks(internal_schema.schema)
   +return schema.new('instance_config', internal_schema, {
        methods = {
            instance_uri = instance_uri,
            apply_vars = apply_vars,
   @@ -2191,6 +2172,5 @@ return schema.new('instance_config', schema.record({
        },
        _extra_annotations = {
            descriptions = descriptions.instance_descriptions,
   -        validators = validators,
        }
    })
   diff --git a/src/box/lua/config/utils/schema.lua b/src/box/lua/config/utils/schema.lua
   index fbf880fb40..8f4d73a883 100644
   --- a/src/box/lua/config/utils/schema.lua
   +++ b/src/box/lua/config/utils/schema.lua
   @@ -1879,11 +1879,13 @@ local function set_schema_annotations(schema, annotations)
        local function set_validator(schema, ctx)
            local field_path = table.concat(ctx.path, '.')
            local validate = validators[field_path]
   -        -- The schema object may already have a validate function set by
   -        -- `enterprise_edition` or provided by `schema.set`. To preserve
   -        -- the existing validate function, use `chain2` to combine the
   -        -- new validator with the one that is already set.
   -        schema.validate = funcutils.chain2(validate, schema.validate)
   +        -- The schema object may already have a validate function,
   +        -- defined by a user using the `vaidate` annotation
   +        -- directly or indirectly provided by `schema.set`. To
   +        -- preserve the existing validate function, use `chain2`
   +        -- to combine the already set function with the new
   +        -- validator.
   +        schema.validate = funcutils.chain2(schema.validate, validate)
        end
    
        -- Set annotations for the given schema node based
   ```
   </details>

   I don't like that we need to create an internal schema to run the post-processing with assigning validators from the table, then extract a schema node and call instance_config's post-processing with assigning `enterprise_edition_validate`.

   It seems like too many technical (boilerplate) code to express a simple logic.

I think it worth to look toward 4th variant. What do you think?

@georgiy-belyanin I would appreciate your glance as well.

src/box/lua/config/utils/schema.lua
```diff
@@ -1872,15 +1874,27 @@ local function set_schema_annotations(schema, annotations)
...
+    -- Set the validator for the given schema node based on the field
+    -- path accumulated in ctx.path.
+    local function set_validator(schema, ctx)
+        local field_path = table.concat(ctx.path, '.')
+        local validate = validators[field_path]
+        -- The schema object may already have a validate function set by
+        -- `enterprise_edition` or provided by `schema.set`. To preserve
+        -- the existing validate function, use `chain2` to combine the
+        -- new validator with the one that is already set.
+        schema.validate = funcutils.chain2(validate, schema.validate)
```
Totktonada, line 1886: I think that the order of the chaining should different: the existing function first, then the external one. See more detailed thoughts in https://github.com/tarantool/tarantool/pull/10694#discussion_r1871693053.

src/box/lua/config/validators.lua
```diff
@@ -0,0 +1,550 @@
...
+    if data == nil or scope == nil then
+        return
+    end
+    if scope ~= 'global' then
+        w.error('sharding.rebalancer_enabled must be defined in ' ..
+                'the global scope.')
+    end
+end
+
+M['sharding.roles'] = nil
```
Totktonada, line 491: Seems unneded.

Totktonada:
Thank you for the laborious work! I appreciate how the validation code is now organized: much easier to navigate.

(TBH, I even prepared some commits that simplify cluster configuration scope checks. I'll send them after this patchset lands.)

I looked over the changes and I see no problems like a missed check or a behavior change (including how the JSON Schema is generated).

I have one objection regarding the EE options checking, it is described in a thread above. OK from me after resolving it.

src/box/lua/config/instance_config.lua
georgiy-belyanin: Thank you for mentioning. The 2nd option seems the most straightforward and it doesn't require extending schema API. The 4th one is good too.

If we'd like not to spend many time on the refactoring the fourth option seems quite OK.

Though, they all seem too me not so good since they're make straightforward code harder to understand. I've written the concerns in #10691

mandesero: Thanks for the suggestions, I think they'll be useful in the future, for example if we decide to implement @georgiy-belyanin's idea with validator chains or something similar.

To sum it up, in a chat with @Totktonada  and @georgiy-belyanin  it was decided that the simplest approach at the moment is to move the implementation of the validation functions to `validators.lua` and replace the function implementations in `instance_config.lua` with the corresponding references to `validators[<..path..>]`.

src/box/lua/config/instance_config.lua
mandesero: Thank you! I missed it.

src/box/lua/config/utils/schema.lua
mandesero: This code has been removed, so it is no longer relevant.

Totktonada (approved):

PR #10711, opened 2024-10-17
memtx: clarify tuples against given index in memtx_tx_snapshot_cleaner

drewdzzz:
Currently, we create `memtx_tx_snapshot_cleaner` for each index in read view. However, we somewhy clarify all tuples against primary index in all cleaners. As a result, secondary indexes work incorrectly in read view when MVCC is enabled, we may even get a tuple with one key, but a tuple with another key will be returned because it is clarified against primary index and repsects its order - that's wrong because all indexes have its own orders.

Community Edition is not affected at all since it uses read view only for making a snapshot - we use only primary indexes there.

Part of https://github.com/tarantool/tarantool-ee/issues/939

EE part: https://github.com/tarantool/tarantool-ee/pull/948

### DIFF
--- a/src/box/memtx_hash.cc
+++ b/src/box/memtx_hash.cc
@@ -639,7 +639,7 @@ memtx_hash_index_create_read_view(struct index *base)
 	index_read_view_create(&rv->base, &vtab, base->def);
 	struct space *space = space_by_id(base->def->space_id);
 	assert(space != NULL);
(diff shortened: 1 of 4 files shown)


mkostoevr (approved):

locker (approved):

drewdzzz:
Cherry-pick to 2.11 and 3.2?

PR #9461, opened 2023-12-09
config: support config schema evolution

Totktonada:
This PR is in very draft state.

See #9452 for details.

PR #10717, opened 2024-10-18
version: introduce Lua module

### LINKED ISSUE #10521: Introduce

Serpentian:
Commit aeb53caa95f8 ("upgrade: allow DDL before box.schema.upgrade is called") allowed using DDL before schema upgrade. However, some features remain blocked. In order to make it convenient for user to write application code, which uses such features and works even before schema upgrade, we introduced `dd_version` in `box.status` broadcast (commit dde4602807cf ("schema: broadcast persistent schema version in box.status")).

However, user still needs a convenient way to compare schema versions. This commit introduces such module.

Closes #10521

@TarantoolBot document
Title: Module version
Since: Tarantool 3.3.0

Overview

The version module is intended to create and compare version objects. For now it supports only schema version (see `dd_version`), full release versions (e.g. `_TARANTOOL`) are not supported yet.

API Reference

Below is list of all `version` function and members:
* version.new() - Create a version object
* version.fromstr() - Get a converted from string version object

1. version.new(major, minor, patch)

```lua
-- Get current schema
tarantool> t = box.space._schema:get('version')
tarantool> v = version.new(t[2], t[3], t[4])
tarantool> v
---
- patch: 0
  minor: 3
  _id: 197376
  major: 3
...
```

Parameters:
    unsigned integers
Return:
    version object
Return type:
    table

2. version.__call() - same as version.new().

3. version.fromstr(version_str) Creates version object from string. Note, that all excess symbols in the string are ignored.

```lua
tarantool> version.fromstr('1.2.1-some-string')
---
- patch: 1
  minor: 2
  _id: 66049
  major: 1
...
```

4. version_object supports comparing and `tostring` method. It is the table, which has `major`, `minor`, `patch` fields.


src/box/lua/mkversion.lua
Totktonada, line 1: Comparing with a value of different type is possible for cdata objects.

```lua
tarantool> ffi = require('ffi')
tarantool> ffi.cdef('struct foo { int x; };')
tarantool> ffi.metatype('struct foo', {__lt = function(self, x) return self.x < x end})
tarantool> ffi.new('struct foo', {x = 1}) < 5
---
- true
...

tarantool> ffi.new('struct foo', {x = 6}) < 5
---
- false
...
```

I'm not sure that I would like to use cdata here, but anyway it worth to ask/discuss points about it: why we are/aren't going this way?

src/box/lua/mkversion.lua
```diff
@@ -25,16 +28,17 @@ local function mkversion_new(major, minor, patch)
...
-                      patch)
+    self._id = bit.bor(bit.lshift(bit.bor(bit.lshift(major, 8), minor), 8),
+                       patch)
     return self
 end
 
 -- Parse string with version in format 'A.B.C' to version object.
 local function mkversion_from_string(version)
-    local major, minor, patch = version:match('^(%d+)%.(%d+)%.(%d+)$')
+    local major, minor, patch = version:match('^(%d+)%.(%d+)%.(%d+)[^%.]*$')
```
Totktonada, line 38: Just comes into my head: it means that the following code that returns true now:

```lua
local v = require('version')

return v.fromstr('1.2.3.4') == v.fromstr('1.2.3.5')
```

...would return `false` if we dedice to support more components in a future.

And it means that we can't extend it due to compatibility concerns.

src/box/lua/mkversion.lua
Totktonada, line 1: I just remember that we have the `internal.config.utils.expression` module that supports three component version comparisons:

Usage example:

https://github.com/tarantool/tarantool/blob/c10c62927e4cba52cfc44103fc1e21c9d6606fb2/src/box/lua/config/applier/sharding.lua#L20-L23

It was implemented to support a config schema evolution: #9461.

Do you find it convenient?

Totktonada:
Just some questions and thoughts at this point. I see no problems in the implementation, but want to re-check our decisions made during the discussion of the RFC document.

changelogs/unreleased/gh-10521-introduce-version-module.md
```diff
@@ -0,0 +1,6 @@
+## feature/core
+
+* Introduced `version` Lua module, which supports creating and
+  comparing schema version (aka `dd_version` in `box.status`, aka
+  `box.space._schema:get('version')`). Intended to ease the use
+  of blocked features before `box.schema.upgrade` (gh-10521).
```
p7nov, line 6:
```suggestion
* Introduced the `version` Lua module, which enables creation and
  comparison of schema version objects (values of `dd_version` in
  `box.status` or `box.space._schema:get('version')`). The module
  eases the use of features blocked before `box.schema.upgrade`
  (gh-10521).
```

p7nov (approved):

src/box/lua/mkversion.lua
locker: Let's support full version specification right now? It shouldn't be difficult AFAIU so why put it off?

src/box/lua/mkversion.lua
```diff
@@ -25,16 +28,17 @@ local function mkversion_new(major, minor, patch)
...
     if major == nil then
-        error('version should be in format A.B.C')
+        local err = "error during parsing version string '%s'"
+        box.error(box.error.PROC_LUA, string.format(err, version))
```
locker, line 41: I think it's better be `box.error.ILLEGAL_PARAMS`.

src/box/lua/version.lua
locker, line 1: I think the `version` module should be located in `src/lua/version.lua` and it should be independent (i.e. better move `mkversion` parts there than make it `require('mkversion')`). As for box-specific methods (like `get_latest_version`), they'd better be in `box.internal`. So some refactoring maybe in place. I don't insist though.

src/box/lua/mkversion.lua
Serpentian: > Let's support full version specification right now? It shouldn't be difficult AFAIU so why put it off?

But it will be independent from the C `version` module we have. Are we ok with that? 

Because I won't touch `src/version.c` module. It's used in replication, there're a lot of version dependent checks in `relay` and `applier`. If the way, we create `version_id` will be changed in C, then we'll have to introduce new flags in replication, which will say which version of the `version` module replica uses. I
don't wanna deal with that for now and see no reason for that actually (except of the code cleanliness). 

P.S. I'll completely get rid of the `mkversion`, we'll have two independent versions of the `version` module: public one (in the current patch), and C internal (`src/version`)

locker: > > Let's support full version specification right now? It shouldn't be difficult AFAIU so why put it off?
> 
> But it will be independent from the C `version` module we have. Are we ok with that?

I think it's okay as long as the Lua `version` module is compatible with it, i.e. it can parse strings generated with the C `version`. In the Lua module we could implement comparisons field-by-field, without using the version id. We don't need to deal with version labels (like alpha, beta, entrypoint) in replication or schema so the version representation would remain the same in C.

src/box/lua/mkversion.lua
Serpentian: It's convenient, when we can compare with strings. Let's use `cdata`, thank you)

src/box/lua/mkversion.lua
Serpentian: > Let's support full version specification right now?

Done

src/box/lua/mkversion.lua
Serpentian: I suppose it's better to use explicit version objects, as everyone already got used to in external modules

src/box/lua/mkversion.lua
Serpentian: Done

src/box/lua/version.lua
Serpentian: In `lua/version` now. `mkversion` is removed, all needed functions are in `box.internal`

src/box/lua/mkversion.lua
Totktonada: I'm not so convinced, but I don't think that these things can't exist in parallel some day. Let's go your road today.

src/box/lua/config/applier/credentials.lua
```diff
@@ -681,13 +681,12 @@ end
 
 local function on_schema_replace_trigger(_, new)
     assert(on_schema_replace_trigger_is_set)
-
-    if new == nil then
+    if new == nil or new[1] ~= 'version' then
```
Totktonada, line 684: This is the only change in the logic I see within this commit. However, it doesn't look related to the removal of the `mkversion` module.

Can you clarify it in the commit message or separate into another commit?

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -307,7 +307,8 @@ local function names_apply(config, missing_names, schema_version)
 
     -- Wait for rw state. If schema version is nil, bootstrap is
     -- going to be done.
-    if schema_version and schema_version < mkversion.get_latest() then
+    if schema_version and schema_version <
+       box.internal.version_schema_latest_get()then
```
Totktonada, line 311: Nit: no whitespace before `then`.

src/box/lua/config/applier/credentials.lua
Serpentian: Yep, it's the bug, on replace trigger tried to create version object, when e.g. `replicaset_name` is set

src/lua/version.lua
```diff
@@ -0,0 +1,243 @@
...
+    end
+
+    rel_type = rel_type or ''
+    rel_num = rel_num or 0
+    commit = commit or 0
+    check_param(major, 'major', 'number')
+    check_param(minor, 'minor', 'number')
+    check_param(patch, 'patch', 'number')
+    check_param(rel_type, 'release type', 'string')
+    check_param(rel_num, 'release number', 'number')
```
Totktonada, line 72: It is about a pre-release and have nothing related to a release. I find these names confusing.

Also, I don't like that these fields have two-word names unlike all the others.

semver.org calls this part `<pre-release identifier>` in a grammar and just `prerelease` in a regexp. Even just `suffix` would make less confusion, IMHO.

Are we really need to split `rc1` to `rc` and `1` in some use scenario? If we don't, I would union them and call `prerelease`. This is simple and straightforward.

src/lua/version.lua
```diff
@@ -0,0 +1,243 @@
...
+--
+ffi.cdef([[
+    struct version {
+        /** Major number. */
+        uint32_t major;
+        /** Minor number. */
+        uint32_t minor;
+        /** Patch number. */
+        uint32_t patch;
+        /** Pre-release type: e.g. entrypoint, apha. */
```
Totktonada, line 43: Typo: apha -> alpha.

src/box/lua/upgrade.lua
```diff
@@ -1454,30 +1454,31 @@ end
...
+    {version = mkversion.new(2, 7, 1), func = upgrade_to_2_7_1},
+    {version = mkversion.new(2, 9, 1), func = upgrade_to_2_9_1},
+    {version = mkversion.new(2, 10, 1), func = upgrade_to_2_10_1},
+    {version = mkversion.new(2, 10, 4), func = upgrade_to_2_10_4},
+    {version = mkversion.new(2, 10, 5), func = upgrade_to_2_10_5},
+    {version = mkversion.new(2, 11, 0), func = upgrade_to_2_11_0},
+    {version = mkversion.new(2, 11, 1), func = upgrade_to_2_11_1},
+    {version = mkversion.new(3, 0, 0), func = upgrade_to_3_0_0},
+    {version = mkversion.new(3, 1, 0), func = upgrade_to_3_1_0},
+    {version = mkversion.new(3, 3, 0), func = upgrade_to_3_3_0},
```
Totktonada, line 1478: These changes look unneeded (and the similar ones below).

Serpentian: The default is `version.fromstr` (see metastable od the module), it seems it much more convenient to create objects from strings and not manually with `version.new`

Totktonada: OK, got it. Thanks for the clarification!

src/box/lua/upgrade.lua
Totktonada, line 1: GitHub marks this file as conflicting on the pull request's webpage. It seems, it needs a rebase.

changelogs/unreleased/gh-10521-introduce-version-module.md
```diff
@@ -0,0 +1,5 @@
+## feature/core
```
locker, line 1: Looks more like `feautre/lua` to me.

src/box/lua/version.lua
```diff
@@ -0,0 +1,45 @@
...
+end
+
+local function version_schema_latest_get()
+    local version = builtin.box_latest_dd_version_id()
+    return version_from_id(version)
+end
+
+box.internal.version_from_tuple = version_from_tuple
+box.internal.version_schema_get = version_schema_get
+box.internal.version_schema_latest_get = version_schema_latest_get
```
locker, line 44: I'd name them `box.internal.dd_version` and `box.internal.latest_dd_version` for consistency with the C functions.

src/lua/version.lua
locker: I think we could rename `commit_num` to `commit`. As for `release_num`, I don't like it, either. I agree that having just one field `prerelease` (alpha, rc1, rc2) would look better. We just need to be careful when comparing these identifiers.

src/box/lua/config/applier/box_cfg.lua
Serpentian: Added one)

src/lua/version.lua
Serpentian: Now there's only `prerelease` field, which returns the concatenated `prerelease_name` and `prerelease_num`. User still can get separated values via `_prerelease_name` and `_prerelease_num` if he'll look into sources

src/lua/version.lua
Serpentian: Fixed

src/box/lua/upgrade.lua
Serpentian: Rebased

changelogs/unreleased/gh-10521-introduce-version-module.md
Serpentian: Sure, done

src/box/lua/version.lua
Serpentian: No problem, don't care a lot how to name them, renamed. Just tried to start with `version_` prefix, `dd` functions are not prefixed but are consistent with C

src/lua/version.lua
```diff
@@ -0,0 +1,248 @@
...
+            box.error(box.error.ILLEGAL_PARAMS,
+                      string.format(msg, prerelease_name))
+        end
+    end
+
+    return ffi.new('struct version', {
+        major = major,
+        minor = minor,
+        patch = patch,
+        _prerelease_name_pointer = prerelease_name,
```
locker, line 87: `prerelease_name` may be garbage-collected while the version object is still in use.

```lua
local ffi = require('ffi')
ffi.cdef([[struct foo { const char *bar; };]])
local foo = ffi.new('struct foo', {string.rep('x', 1024 * 1024)})
collectgarbage('collect')
print(ffi.string(foo.bar))
```

<details><summary>Segmentation fault</summary>

```
Segmentation fault
  code: SEGV_MAPERR
  addr: 0x40bdc020
  context: 0x5a59a16f6340
  siginfo: 0x5a59a16f6470
  rax      0x20               32
  rbx      0x407a953c         1081775420
  rcx      0x100              256
  rdx      0x40bdc020         1086177312
  rsi      0x40016008         1073831944
  rdi      0x40bdc020         1086177312
  rsp      0x74ec18c80be8     128557376867304
  rbp      0x407a9378         1081774968
  r8       0x40db8628         1088128552
  r9       0x100              256
  r10      0x0                0
  r11      0x203              515
  r12      0x0                0
  r13      0x0                0
  r14      0x407aa0a0         1081778336
  r15      0x5a599f5416e5     99340971677413
  rip      0x74ec1bb8b75d     128557426194269
  eflags   0x10283            66179
  gs       0x0                0
  fs       0x0                0
  cr2      0x40bdc020         1086177312
  err      0x4                4
  oldmask  0x0                0
Current time: 1731495081
this may fail as well
#1  0x5a599f483b61 in crash_collect+256
#2  0x5a599f48459d in crash_signal_cb+100
#3  0x74ec1ba45320 in __sigaction+80
#4  0x74ec1bb8b75d in __nss_database_lookup+14205
#5  0x5a599f54196d in lj_cf_ffi_string+648
#6  0x5a599f4f25c7 in lj_BC_FUNCC+70
#7  0x5a599f50010b in lua_pcall+969
#8  0x5a599f45ef03 in luaT_call+45
#9  0x5a599f45046c in lua_main+283
#10 0x5a599f451086 in run_script_f+3073
#11 0x5a599f1a7183 in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*)+34
#12 0x5a599f490757 in fiber_loop+219
#13 0x5a599f8c0f18 in coro_init+120
Aborted
```

</details>

src/box/lua/config/applier/credentials.lua
locker: Can you add a test?

Serpentian: > Can you add a test?

What for? The bug does not exist before replacing `mkversion` with `version`: `mkversion.from_tuple` returned nil, if smth goes wrong, with `version` module used `internal.version_from_tuple` either returns version object or throws an error.  And after replacing `mkversion` with `version` a bunch of tests doesn't work.

src/lua/version.lua
Serpentian: I didn't manage to reproduce the segfault with version module itself. Tried to run the following script:

<details><summary>Details</summary>
<p>

```lua
local v = require('version')
local a = v('1.1.1-entrypoint1-45')
collectgarbage('collect')
print(a.prerelease)

```

</p>
</details>

However, I anyway changed implementation so that now it doesn't have any strings in cdef structure out of harm's way. It's anyway better to save weights and not names themselves, this way it's easier to compare, the code even looks cleaner, IMHO

src/lua/version.lua
```diff
@@ -0,0 +1,248 @@
...
+        return version_fromstr(version)
+    end
+    return nil
+end
+
+local version_mt = {
+    __index = function(v, field)
+        if field == 'prerelease' then
+            return v._prerelease_name .. v._prerelease_num
+        elseif field == '_prerelease_name' then
```
locker, line 194: Do we really need to export `_prerelease_name` here? Users should only use `prerelease` while for comparisons and internal operations `_prerelease_id` and `_prerelease_num` should be enough.

src/box/lua/config/applier/credentials.lua
Totktonada: Seems solved now, thanks!

src/lua/version.lua
```diff
@@ -0,0 +1,248 @@
...
+        uint32_t commit;
+    };
+]])
+
+local prerelease_name_to_weight = {
+    entrypoint = 10,
+    alpha = 20,
+    beta = 30,
+    rc = 40,
+}
```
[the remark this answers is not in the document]
Serpentian: Done, ids are used instead

src/lua/version.lua
```diff
@@ -0,0 +1,248 @@
...
+        return ver1._prerelease_num - ver2._prerelease_num
+    end
+    if ver1.commit ~= ver2.commit then
+        return ver1.commit - ver2.commit
+    end
+    return 0
+end
+
+local version_ctype
+local function version_cast_if_needed(version)
```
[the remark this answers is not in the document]
Serpentian: Renamed to `version_cast`

src/lua/version.lua
```diff
@@ -0,0 +1,248 @@
...
+        if not l or not r then
+            return false
+        end
+        return version_cmp(l, r) == 0
+    end,
+    __lt = function(l, r)
+        l = version_cast_if_needed(l)
+        r = version_cast_if_needed(r)
+        if not l or not r then
+            box.error(box.error.ILLEGAL_PARAMS, "Cannot cast to version object")
```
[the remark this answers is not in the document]
Serpentian: Good catch! Done

src/lua/version.lua
Serpentian: Nah, not really, we can make it a function, it was just convenient to use

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+for id, name in ipairs(prerelease_names) do
+    prerelease_name_to_id[name] = id
+end
+
+local function version_new(major, minor, patch, prerelease, commit)
+    commit = commit or 0
+    check_param(major, 'major', 'number')
+    check_param(minor, 'minor', 'number')
+    check_param(patch, 'patch', 'number')
+    check_param(commit, 'commit', 'number')
```
locker, line 76: 1. Please add the level argument to all `check_param` calls. Should be 2 to point to the calling code.
2. Please add a test checking errors raised by these functions.

test/app-luatest/version_module_test.lua
```diff
@@ -0,0 +1,300 @@
...
+
+        t.assert(ver <= v.ver, ('%d'):format(i))
+        t.assert(ver <= v.str, ('%d'):format(i))
+        t.assert(v.str <= ver, ('%d'):format(i))
+        t.assert(v.ver <= ver, ('%d'):format(i))
+
+        t.assert(ver >= v.ver, ('%d'):format(i))
+        t.assert(ver >= v.str, ('%d'):format(i))
+        t.assert(v.str >= ver, ('%d'):format(i))
+        t.assert(v.ver >= ver, ('%d'):format(i))
```
locker, line 241: Please use `t.assert_equals`, `t.assert_lt`, `t.assert_ge`, etc instead of comparing the result in the assertion. They yield a more verbose output in case of a failure.

src/box/CMakeLists.txt
Totktonada, line 1: Sorry, but I'm going to push [one more](https://github.com/tarantool/tarantool/pull/10849) mkversion usage.

BTW, I've a plan to review your patchset today. Sorry for the delay.

test/app-luatest/version_module_test.lua
Serpentian: @locker I intentionally used `t.assert` and not its variations, as I wanna be sure, that luatest doesn't do some smart stuff (like for example for comparison of the tables [here](https://github.com/tarantool/luatest/blob/7dc5cb7d581312f48c97d8f394674975d5a45fd8/luatest/comparator.lua#L6-L25)). Moreover, `assert_lt/ge` accepts only numbers, see [here](https://github.com/tarantool/luatest/blob/7dc5cb7d581312f48c97d8f394674975d5a45fd8/luatest/assertions.lua#L365-L368), so comparison of cdata objects is not possible.

src/box/CMakeLists.txt
Serpentian: > Sorry, but I'm going to push https://github.com/tarantool/tarantool/pull/10849 mkversion usage.

And this patch already requires EE version, since we merged https://github.com/tarantool/tarantool-ee/pull/1004. So, nevermind)

test/app-luatest/version_module_test.lua
Totktonada: AFAIR, I saw a discussion or a comment about it before. It seems worthful to include a note into the code as a comment, so this idea is clear for a reader and the question doesn't raise again.

src/lua/version.lua
Serpentian: Thank you for noticing. Fixed `check_param` and addded tests for them. Note, that the trace is automatically tested, when error is checked ([luatest](https://github.com/tarantool/luatest/blob/7dc5cb7d581312f48c97d8f394674975d5a45fd8/luatest/assertions.lua#L143)). So, I just added tests for arguments

src/box/CMakeLists.txt
Serpentian: Rebased and added EE version https://github.com/tarantool/tarantool-ee/pull/1009

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+            local msg = "Unknown release type '%s'"
+            box.error(box.error.ILLEGAL_PARAMS,
+                      string.format(msg, prerelease_name), 2)
+        end
+    else
+        -- Greatest priority, actual release.
+        prerelease_id = 0
+    end
+
+    assert(prerelease_id <= MAX_PRERELEASE_NAME_WEIGHT)
```
Totktonada, line 100: Looks like the `MAX_PRERELEASE_NAME_WEIGHT` variable is used only for this boundary check. And it has `weight` name that seems a trace from a previous version of the patch. And we don't check boundaries for other fields.

I would either do the boundary check for all the fields or drop this one.

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+    check_param(commit, 'commit', 'number', 2)
+
+    local prerelease_id
+    local prerelease_num = 0
+    if prerelease then
+        local prerelease_name
+        check_param(prerelease, 'prerelease', 'string', 2)
+        prerelease_name, prerelease_num = prerelease:match('^(%a+)(%d+)')
+        if not prerelease_name then
+            prerelease_name = prerelease:match('^(%a+)')
```
Totktonada, line 85: Should we raise an error if the pre-release marker has unexpected format? For example,

```lua
tarantool> version.new(3, 3, 0, 'rc.5')
---
- 3.3.0-rc
...
```

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+    check_param(minor, 'minor', 'number', 2)
+    check_param(patch, 'patch', 'number', 2)
+    check_param(commit, 'commit', 'number', 2)
+
+    local prerelease_id
+    local prerelease_num = 0
+    if prerelease then
+        local prerelease_name
+        check_param(prerelease, 'prerelease', 'string', 2)
+        prerelease_name, prerelease_num = prerelease:match('^(%a+)(%d+)')
```
Totktonada, line 83: Here we also accept something unexpected:

```lua
tarantool> version.new(3, 3, 0, 'rc5.pre1')
---
- 3.3.0-rc5
...
```

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+    if not major or not minor or not patch then
+        local msg = "Error during parsing version string '%s'. " ..
+                    "At least X.Y.Z is required."
+        box.error(box.error.ILLEGAL_PARAMS, string.format(msg, version_str), 2)
+    end
+    major = tonumber(major)
+    minor = tonumber(minor)
+    patch = tonumber(patch)
+    -- Cut to 'type<num>-<num>-g<commit>'.
+    pos = version_str:find('-')
```
Totktonada, line 129: This `pos` does not necessarily correspond to the `<major>.<minor>.<patch>` triplet end.

```lua
tarantool> version('3.3.0HAHAHAHAHA-entrypoint-202-gedf82e996d')
---
- 3.3.0-entrypoint-202
...
```

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+        local ok, v = pcall(version_fromstr, version)
+        return ok and v
+    end
+    return nil
+end
+
+local function version_get_prerelease(v)
+    if v._prerelease_id == 0 then
+        return ''
+    end
```
Totktonada, line 199:
```lua
tarantool> do local prerelease = nil return version.new(3, 3, 0, prerelease).prerelease == prerelease end
---
- false
...

tarantool> do local prerelease = '' return version.new(3, 3, 0, prerelease).prerelease == prerelease end
---
- error: Unknown release type 'nil'
...
```

Shouldn't one of these calls return the `prerelease` value back and the other be forbidden?

Totktonada (approved):
I've no objections, looks good. Thank you for the patchset!

I left some remarks regarding parsing, but I don't know, whether it is important or some things are made non-strict for simplicity. Feel free to solve them in a way you like more (including keeping as is).

src/lua/version.lua
Serpentian: This one is not needed anymore, it was used only in assertion. Dropped it

src/lua/version.lua
Serpentian: Now prerelease type is strictly checked, it must be "^(name)(num)$", arbitrary symbols in the end are not allowed anymore

src/lua/version.lua
Serpentian: See https://github.com/tarantool/tarantool/pull/10717#discussion_r1855700695

src/lua/version.lua
Serpentian: Fixed, now `nil` is returned instead of empty string. `''` prerelease type is forbidden (also fixed the error msg, `'nil`' name won't be shown anymore).

src/lua/version.lua
Serpentian: I've decided to strictify the checks, since it's not difficult to do. Now arbitrary symbols between version parts in string are forbidden. Also, I've forbidden to add arbitrary info in the end of the version string, if it's not prefixed with `-` symbol: `3.3.0-rc1dev` is forbidden, `3.3.0-rc1-dev` is allowed.

src/lua/version.lua
```diff
@@ -0,0 +1,259 @@
...
+    if not pos then
+        goto finish
+    end
+    version_str = version_str:sub(pos + 1)
+
+    -- Part 2 - prerelease name and number, might be absent.
+    prerelease = version_str:match('^(%a+%d*)')
+    -- Cut to '<num>-g<commit>'.
+    if prerelease then
+        pos = version_str:find('-')
```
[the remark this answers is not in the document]
Serpentian: See https://github.com/tarantool/tarantool/pull/10717#discussion_r1855703983

Serpentian:
Had to rebase for EE part, so here's the diff:

<details><summary>Details</summary>
<p>

```diff
diff --git a/src/lua/version.lua b/src/lua/version.lua
index d824b5af27..96d859b03a 100644
--- a/src/lua/version.lua
+++ b/src/lua/version.lua
@@ -18,18 +18,12 @@
 -- * Build metadata is denoted by appending hyphen, not plus sign. Not
 --   x.y.z-typeN+commit-ghash, but x.y.z-typeN-commit-ghash.
 --
--- Note, that ghash itself and everything after it (e.g. -dev suffix) is
--- ignored and doesn't affect comparing of the versions.
---

 local ffi   = require('ffi')
 local utils = require('internal.utils')

 local check_param = utils.check_param

--- Weight of the empty prerelease name.
-local MAX_PRERELEASE_NAME_WEIGHT = 255
-
 --
 -- Cdata is used in order to allow comparison of the version object with
 -- strings. Since relational metamethods do not support mixed types in Lua,
@@ -80,24 +74,23 @@ local function version_new(major, minor, patch, prerelease, commit)
     if prerelease then
         local prerelease_name
         check_param(prerelease, 'prerelease', 'string', 2)
-        prerelease_name, prerelease_num = prerelease:match('^(%a+)(%d+)')
+        prerelease_name, prerelease_num = prerelease:match('^(%a+)(%d+)$')
         if not prerelease_name then
-            prerelease_name = prerelease:match('^(%a+)')
+            prerelease_name = prerelease:match('^(%a+)$')
         else
             prerelease_num = tonumber(prerelease_num)
         end
         prerelease_id = prerelease_name_to_id[prerelease_name]
         if not prerelease_id then
-            local msg = "Unknown release type '%s'"
+            local msg = "Unknown prerelease type '%s'"
             box.error(box.error.ILLEGAL_PARAMS,
-                      string.format(msg, prerelease_name), 2)
+                      string.format(msg, prerelease_name or prerelease), 2)
         end
     else
         -- Greatest priority, actual release.
         prerelease_id = 0
     end

-    assert(prerelease_id <= MAX_PRERELEASE_NAME_WEIGHT)
     return ffi.new('struct version', {
         major = major,
         minor = minor,
@@ -110,47 +103,70 @@ end

 local function version_fromstr(version_str)
     check_param(version_str, 'version_str', 'string', 2)
+    local str = version_str
     --  x.x.x-name<num>-<num>-g<commit>
     -- \____/\________/\_____/
     --   P1      P2      P3
-    local major, minor, patch, prerelease, commit, pos
+    local major, minor, patch, prerelease, commit, pos, end_pos, tmp_pos, _

     -- Part 1 - version ID triplet. Compulsory fields.
-    major, minor, patch = version_str:match('^(%d+)%.(%d+)%.(%d+)')
+    _, end_pos, major, minor, patch = str:find('^(%d+)%.(%d+)%.(%d+)')
     if not major or not minor or not patch then
-        local msg = "Error during parsing version string '%s'. " ..
-                    "At least X.Y.Z is required."
-        box.error(box.error.ILLEGAL_PARAMS, string.format(msg, version_str), 2)
+        goto error
     end
     major = tonumber(major)
     minor = tonumber(minor)
     patch = tonumber(patch)
     -- Cut to 'type<num>-<num>-g<commit>'.
-    pos = version_str:find('-')
+    pos = str:find('-')
     if not pos then
         goto finish
     end
-    version_str = version_str:sub(pos + 1)
+    if pos ~= end_pos + 1 then
+        -- Do not allow arbitary symbols between version triplet and prerelease:
+        -- e.g. forbid 3.3.0Hello-rc1.
+        goto error
+    end
+    str = str:sub(pos + 1)

     -- Part 2 - prerelease name and number, might be absent.
-    prerelease = version_str:match('^(%a+%d*)')
+    _, tmp_pos, prerelease = str:find('^(%a+%d*)')
     -- Cut to '<num>-g<commit>'.
     if prerelease then
-        pos = version_str:find('-')
+        end_pos = tmp_pos
+        pos = str:find('-')
         if not pos then
             goto finish
         end
-        version_str = version_str:sub(pos + 1)
+        if pos ~= end_pos + 1 then
+            -- Do not allow arbitary symbols between prerelease and commit:
+            -- e.g. forbid 3.3.0-rc1.Hello-20
+            goto error
+        end
+        str = str:sub(pos + 1)
     end

     -- Part 3 - commit count since the latest release, might be absent.
-    commit = version_str:match('^(%d+)')
+    _, tmp_pos, commit = str:find('^(%d+)')
     if commit then
+        end_pos = tmp_pos
         commit = tonumber(commit)
     end

 ::finish::
-    return version_new(major, minor, patch, prerelease, commit)
+    -- Note, that ghash itself and everything after it (e.g. -dev suffix) is
+    -- ignored and doesn't affect comparing of the versions. Only
+    -- x.y.z-typeN-commit part is parsed and validated. However, if anything
+    -- is present after version string it must be prefixed with '-' symbol.
+    str = str:sub(end_pos + 1)
+    if str ~= '' and not str:match('^(-.*)$') then
+        goto error
+    end
+    do return version_new(major, minor, patch, prerelease, commit) end
+
+::error::
+    local msg = "Error during parsing version string '%s'"
+    box.error(box.error.ILLEGAL_PARAMS, string.format(msg, version_str), 2)
 end

 local function version_cmp(ver1, ver2)
@@ -195,7 +211,7 @@ end

 local function version_get_prerelease(v)
     if v._prerelease_id == 0 then
-        return ''
+        return nil
     end
     local name = prerelease_names[v._prerelease_id]
     if v._prerelease_num ~= 0 then
@@ -239,7 +255,7 @@ local version_mt = {
     __tostring = function(v)
         local str = v.major .. '.' .. v.minor .. '.' .. v.patch
         local prerelease = version_get_prerelease(v)
-        if prerelease ~= '' then
+        if prerelease then
             str = str .. '-' .. prerelease
         end
         if v.commit ~= 0 then
diff --git a/test/app-luatest/version_module_test.lua b/test/app-luatest/version_module_test.lua
index 4cba3a9f5e..3c6fcb7101 100644
--- a/test/app-luatest/version_module_test.lua
+++ b/test/app-luatest/version_module_test.lua
@@ -265,30 +265,35 @@ g.test_order = function()
     end
 end

-g.test_error = function()
+g.test_fromstr_error = function()
     local msg = "Error during parsing version string"
+    -- Check, that version triplet is mandatory.
     t.assert_error_msg_contains(msg, version.fromstr, 'bad version')
     t.assert_error_msg_contains(msg, version.fromstr, '1.x.x')
     t.assert_error_msg_contains(msg, version, '1.2.x')
+    -- Forbid arbitary symbols between triplet and prerelease.
+    t.assert_error_msg_contains(msg, version, '1.2.7Hey-rc4-46-')
+    -- Forbid arbitary symbols between prerelease and commit num.
+    t.assert_error_msg_contains(msg, version, '1.2.7-rc4Hey-46-')
+    -- Forbid arbitary symbols not prefixed with symbol '-'.
+    t.assert_error_msg_contains(msg, version, '1.2.7.')
+    t.assert_error_msg_contains(msg, version, '1.2.7Hey')
+    t.assert_error_msg_contains(msg, version, '1.2.7-rc1-20Hey')
+    msg = "Unknown prerelease type 'rcHey'"
+    t.assert_error_msg_contains(msg, version, '1.2.7-rcHey')
+end

-    msg = "Cannot cast to version object"
-    t.assert_error_msg_contains(msg, function()
-        if version('1.2.3') <= {} then t.assert(true) end
-    end)
-    t.assert_not_equals(version('1.1.1'), {})
-
-    ffi.cdef('struct foo { int bar; };')
-    local cdata = ffi.new('struct foo', {bar = 1})
-    t.assert_error_msg_contains(msg, function()
-        if version('1.2.3') < cdata then t.assert(true) end
-    end)
-    t.assert_not_equals(version('1.1.1'), cdata)
-
-    msg = "Unknown release type 'gam'"
-    t.assert_error_msg_contains(msg, version, '1.2.7-gam4ma3-46-')
+g.test_prerelease_argument = function()
+    local msg = "Unknown prerelease type '%s'"
+    -- Note, that empty prerelease type is also forbidden.
+    local cases = { 'rc.5', 'rc5.pre1', '' }
+    for _, c in ipairs(cases) do
+        t.assert_error_msg_contains(msg:format(c), version.new, 1, 1, 1, c)
+    end
+end

-    -- Check the types of the arguments.
-    msg = "should be a number"
+g.test_arguments_type = function()
+    local msg = "should be a number"
     t.assert_error_msg_contains(msg, version.new, 'a', 1, 1)
     t.assert_error_msg_contains(msg, version.new, 1, 'a', 1)
     t.assert_error_msg_contains(msg, version.new, 1, 1, 'a')
@@ -296,12 +301,23 @@ g.test_error = function()
     msg = "should be a string"
     t.assert_error_msg_contains(msg, version.new, 1, 1, 1, 1)
     t.assert_error_msg_contains(msg, version, 1)
+end

+g.test_cast_error = function()
+    local msg = "Cannot cast to version object"
+    t.assert_error_msg_contains(msg, function()
+        if version('1.2.3') <= {} then t.assert(true) end
+    end)
+    t.assert_error_msg_contains(msg, function()
+        if version('1.2.3') <= 'a.a.a.' then t.assert(true) end
+    end)

-    -- This one is valid, 1.2.4-2 is parsed, -2 is commit number. Everything
-    -- after it is skipped.
-    local a = version('1.2.4-2-rc0-dev')
-    t.assert_equals(tostring(a), '1.2.4-2')
+    ffi.cdef('struct foo { int bar; };')
+    local cdata = ffi.new('struct foo', {bar = 1})
+    t.assert_error_msg_contains(msg, function()
+        if version('1.2.3') < cdata then t.assert(true) end
+    end)
+    t.assert_not_equals(version('1.1.1'), cdata)
 end

 g.test_fields = function()
@@ -315,4 +331,6 @@ g.test_fields = function()

     t.assert_equals(type(v.major), 'number')
     t.assert_equals(type(v.prerelease), 'string')
+
+    t.assert_equals(version.new(1, 2, 5, nil).prerelease, nil)
 end

```

</p>
</details>

locker (approved):

Totktonada:
Thanks for the updates! Seems OK for me.

PR #10718, opened 2024-10-18
config: fix scaling a replicaset down to one node

### LINKED ISSUE #10716: config: scaling a replicaset down to one instance doesn't work
For example, we have the following configuration with two instances in a replicaset:

```yaml
credentials:
  users:
    replicator:
      password: secret
      roles: [replication]

iproto:
  listen:
  - uri: 'unix/:./{{ instance_name }}.iproto'
  advertise:
    peer:
      login: 'replicator'

replication:
  failover: manual

groups:
  g-001:
    replicasets:
      r-001:
        leader: i-001
        instances:
          i-001: {}
          i-002: {}
```

1. Start the replicaset:

   ```sh
   $ tarantool -i --config config.yaml --name i-001
   ```

   ```sh
   $ tarantool -i --config config.yaml --name i-002
   ```

2. Drop `i-002` from the config.
3. Reload configuration on `i-001`:

   ```yaml
   i-001> config = require('config')
   i-001> config:reload()
   ```

   It is expected that `i-002` is removed from `box.cfg.replication`.

   Actual result:

   ```yaml
   i-001> box.cfg.replication
   ---
   - - password: secret
       uri: unix/:./i-001.iproto
       login: replicator
     - password: secret
       uri: unix/:./i-002.iproto
       login: replicator
   ...
   ```
4. Stop `i-002`.
5. Observe warnings in `box.info.replication`:

   ```yaml
   i-001> box.info.replication
   ---
   - 1:
       id: 1
       uuid: 3a19e457-5984-4e70-b7a5-5645e962d8f3
       lsn: 7
       name: i-001
     2:
       id: 2
       uuid: ca16036d-7d51-4c82-aa18-ca1d7f1d1844
       lsn: 0
       upstream:
         peer: replicator@unix/:./i-002.iproto
         lag: 0.00017929077148438
         status: disconnected
         idle: 122.068509099
         message: 'connect to unix/:./i-002.iproto, called on fd 25, aka unix/:(socket):
           No such file or directory'
         system_message: No such file or directory
       name: i-002
       downstream:
         status: stopped
         message: 'unexpected EOF when reading from socket, called on fd 25, aka unix/:(socket),
           peer of unix/:(socket): Broken pipe'
         system_message: Broken pipe
   ...
   ```
6. Fix it:

   ```lua
   i-001> box.cfg{replication = box.cfg.replication[1]}
   ```
7. Observe less warnings in `box.info.replication`:

   ```yaml
   i-001> box.info.replication
   ---
   - 1:
       id: 1
       uuid: 3a19e457-5984-4e70-b7a5-5645e962d8f3
       lsn: 7
       name: i-001
     2:
       id: 2
       uuid: ca16036d-7d51-4c82-aa18-ca1d7f1d1844
       lsn: 0
       downstream:
         status: stopped
         message: 'unexpected EOF when reading from socket, called on fd 25, aka unix/:(socket),
           peer of unix/:(socket): Broken pipe'
         system_message: Broken pipe
       name: i-002
   ...
   ```

The reason is `return nil` in the following code:

https://github.com/tarantool/tarantool/blob/b33f17b25de6bcbe3ebc236250976e4a0250e75e/src/box/lua/config/applier/box_cfg.lua#L8-L12

This `nil` is passed to `box.cfg({replication = <...>})` and interpreted by `box.cfg()` as an instruction to leave the parameter as is. We should pass `box.NULL` instead.

Also tracked in TNTP-243.

Totktonada:
Before this patch there was a mistake in the YAML configuration processing, when an instance removal is not taken into account in generation of a list of upstreams. It occurs when the updated YAML configuration has only one instance in the given replicaset.

Fixes #10716

### DIFF
--- a/test/config-luatest/downscaling_test.lua
+++ b/test/config-luatest/downscaling_test.lua
@@ -1,0 +1,128 @@
+local t = require('luatest')
+local cbuilder = require('luatest.cbuilder')
+local cluster = require('test.config-luatest.cluster')
+
+local g = t.group()
+
+g.before_all(cluster.init)
+g.after_each(cluster.drop)
+g.after_all(cluster.clean)
+
+-- Verify that if an instance is removed from the declarative
+-- configuration, it is removed from upstreams in the box-level
+-- configuration (box.cfg.replication) after config:reload().
+--
+-- It effectively stops the data flow from the removed instance to
+-- others.
+g.test_size_3_to_2 = function(g)
+    local config = cbuilder:new()
+        :set_replicaset_option('replication.failover', 'manual')
+        :set_replicaset_option('leader', 'i-001')
+        :add_instance('i-001', {})
+        :add_instance('i-002', {})
+        :add_instance('i-003', {})
+        :config()
+
+    local cluster = cluster.new(g, config)
+    cluster:start()
+
+    -- Verify a test case prerequisite: all the instances have all
+    -- the instances in upstreams.
+    cluster:each(function(server)
+        server:exec(function()
+            local config = require('config')
+
+            t.assert_equals(box.cfg.replication, {
+                config:instance_uri('peer', {instance = 'i-001'}),
+                config:instance_uri('peer', {instance = 'i-002'}),
+                config:instance_uri('peer', {instance = 'i-003'}),
+            })
+        end)
+    end)
+
+    -- Remove i-003 from the configuration and write it to the
+    -- file.
+    local config_2 = cbuilder:new(config)
+        :set_replicaset_option('instances.i-003', nil)
+        :config()
+    cluster:sync(config_2)
+
+    -- Define a helper method.
+    cluster.each_except = function(self, exclusion, f)
+        self:each(function(server)
+            if server.alias ~= exclusion then
+                f(server)
+            end
+        end)
+    end
+
+    -- Reload the configuration on i-001 and i-002.
+    cluster:each_except('i-003', function(server)
+        server:exec(function()
+            local config = require('config')
+
+            config:reload()
+        end)
+    end)
+
+    -- Verify that i-003 is not in the configured upstreams on
+    -- i-001 and i-002.
+    cluster:each_except('i-003', function(server)
+        server:exec(function()
+            local config = require('config')
+
+            t.assert_equals(box.cfg.replication, {
+                config:instance_uri('peer', {instance = 'i-001'}),
+                config:instance_uri('peer', {instance = 'i-002'}),
+            })
+        end)
+    end)
+end
+
+-- Same as previous, but scales the replicaset down from two
+-- instances to one.
+--
+-- This scenario was broken before gh-10716.
+g.test_size_2_to_1 = function(g)
+    local config = cbuilder:new()
+        :set_replicaset_option('replication.failover', 'manual')
+        :set_replicaset_option('leader', 'i-001')
+        :add_instance('i-001', {})
+        :add_instance('i-002', {})
+        :config()
+
+    local cluster = cluster.new(g, config)
+    cluster:start()
+
+    -- Verify a test case prerequisite: both instances are in the
+    -- upstream list on i-001.
+    cluster['i-001']:exec(function()
+        local config = require('config')
+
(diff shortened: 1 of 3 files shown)


test/config-luatest/downscaling_test.lua
```diff
@@ -0,0 +1,128 @@
...
+    cluster:sync(config_2)
+
+    -- Define a helper method.
+    cluster.each_except = function(self, exclusion, f)
+        self:each(function(server)
+            if server.alias ~= exclusion then
+                f(server)
+            end
+        end)
+    end
```
georgiy-belyanin, line 57: I suppose it's better to change the sync method of the cluster not to iterate in each through the instances removed with `sync()`

Seems like, we should create a separate issue on it. What do you think?

test/config-luatest/downscaling_test.lua
```diff
@@ -0,0 +1,128 @@
...
+
+        config:reload()
+    end)
+
+    -- Verify that there is no i-002 in the upstreams. In fact,
+    -- when the only instance in the upstreams list is the
+    -- current instance, the empty upstreams list is generated.
+    cluster['i-001']:exec(function()
+        t.assert_equals(box.cfg.replication, {})
+    end)
```
georgiy-belyanin, line 127: Nit: probably, we can rewrite it a bit briefer:
```suggestion
    cluster['i-001']:exec(function()
        -- Reload the configuration on i-001.
        require('config'):reload()

        -- Verify that there is no i-002 in the upstreams. In
        -- fact, when the only instance in the upstreams list is
        -- the current instance, the empty upstreams list is 
        -- generated.
        t.assert_equals(box.cfg.replication, {})
    end)
```

test/config-luatest/downscaling_test.lua
```diff
@@ -0,0 +1,128 @@
...
+    cluster:each_except('i-003', function(server)
+        server:exec(function()
+            local config = require('config')
+
+            t.assert_equals(box.cfg.replication, {
+                config:instance_uri('peer', {instance = 'i-001'}),
+                config:instance_uri('peer', {instance = 'i-002'}),
+            })
+        end)
+    end)
```
georgiy-belyanin, line 79: Dito:
```
    cluster:each_except('i-003', function(server)
        server:exec(function()
            -- Reload the configuration on i-001 and i-002.
            config:reload()

            -- Verify that i-003 is not in the configured
            -- upstreams on i-001 and i-002.
            t.assert_equals(box.cfg.replication, {
                config:instance_uri('peer', {instance = 'i-001'}),
                config:instance_uri('peer', {instance = 'i-002'}),
            })
        end)
    end)
```

FYI: I've tested it, the test still passes.

georgiy-belyanin (approved):
Thank you for the patch! Strange I haven't come across these problems yet. I am generally OK. But still have a few notes.

test/config-luatest/downscaling_test.lua
Totktonada: I agree that it would be convenient to have a support for temporary and permanent removing of instances from the cluster in the `cluster` testing helper. We should save the server objects into a separate field in the first case (to stop/drop them later), but we possibly should stop/drop the server object immediately in the second case. Also, a permanent removal means that we should go to a master instance and drop the corresponding `box.space._cluster` entry.

IOW, I'm in doubt about a proper API here, so postponed it. Maybe `:sync()` is not really what we want to have at the end and several more explicit methods are needed.

I prefer to return here back on demand, that's just our testing helper. I don't think that it needs an issue like a defect or a functionality in tarantool.

I hope it'll spin for a while in my head and I'll return back with some proposal.

test/config-luatest/downscaling_test.lua
Totktonada: I like that the test case looks as a sequence of steps on the outmost level. If we mix them, sure, it is less code. But I doubt that it become more clear.

georgiy-belyanin: Hm, ok. If it's done intentionally I am ok with it.

grafin (approved):
Thanks for fixing this! I see no problems.

Issue #10551, opened 2024-09-10
Crash on OOM on TX rollback 


nshy:
The script:

```lua
local fiber = require('fiber')

box.cfg{}
for i=1,10 do
    local s = box.schema.create_space('test' .. i)
    s:create_index('pk')
    s:insert({1})
end

local errinj = box.error.injection
errinj.set('ERRINJ_WAL_DELAY', true)
for i=1,10 do
    fiber.create(function()
        local s = box.space['test' .. i]
        s:insert({2})
    end)
end
fiber.create(function()
    box.snapshot()
end)
errinj.set('ERRINJ_INDEX_ALLOC', true)
errinj.set('ERRINJ_WAL_WRITE_DISK', true)
errinj.set('ERRINJ_WAL_DELAY', false)
os.exit()
```

crashes with backtrace:
```c
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007adcaeea5463 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2  0x00007adcaee4c120 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007adcaee334c3 in __GI_abort () at abort.c:79
#4  0x00005cfbcb319df2 in crash_signal_cb (signo=11, siginfo=0x5cfbfcc7e330, context=0x5cfbfcc7e200) at /home/shiny/dev/tarantool/src/lib/core/crash.c:204
#5  <signal handler called>
#6  0x00005cfbcb083591 in NS_USE_HINT::bps_tree_memtx_tree_delete_from_leaf (tree=0x5cfbfccf83a8, leaf_path_elem=0x7fffa6254360)
    at /home/shiny/dev/tarantool/src/lib/salad/bps_tree.h:3745
#7  0x00005cfbcb0885b0 in NS_USE_HINT::bps_tree_memtx_tree_process_delete_leaf (tree=0x5cfbfccf83a8, path=0x7fffa62543a0, leaf_path_elem=0x7fffa6254360)
    at /home/shiny/dev/tarantool/src/lib/salad/bps_tree.h:5423
#8  0x00005cfbcb08a7fa in NS_USE_HINT::bps_tree_memtx_tree_delete (t=0x5cfbfccf83a8, elem=..., if_identical=false, deleted_elem=0x0, offset=0x0)
[...]

Issue #727, opened 2015-02-17
RTREE index safety if out of memory


alyapunov:
Now if memory allocation fails, rtree index panic.
We should handle OOM more carefully. Ideally any memory fails must be logged and system must rolback to previous correct state (that it'll be hard to implement due to rtree specific features).
I've added a disabled test for this case (test/box/rtree_errinj.test.lua, CR pending now)

kostja:
See gh-625

kostja:
How to fix: 
- put aside a set of blocks which are sufficient to cover insert-on-delete needs
- if out of main index memory, switch to using the reserve, and disable the tree until the reserve is refilled
- to calculate the necessary amount of the set-aside buffer, use a probablistic model, assuming page splits of pages are independent events, and simulation with stress testing.

PR #10734, opened 2024-10-23
memtx: use the matras's new reservation machinery

mkostoevr:
A long time ago the it's been found that in some circumstances the
RTREE index could fail to rollback because of OOM. This is the case
because removal of a key from such an index can sometimes lead to
allocation of new blocks that could fail. So, in order to guarantee
a successful rollback of an insertion into the index, the memtx
engine extent reservation machinery had been introduced (see the
issue #727).

But some time ago it's beed found out that the reservation of memory
prior to insertion can be not enough to guarantee a successful roll
back of the statements, so it's been decided to use a regular malloc
for failed allocations performed during rollback (see #10551). This
also fixed the RTREE issue for which the extent reservation had been
introduced in the first place.

But this functionality allows to simplify code of data structures,
such as BPS tree, so that we can reserve the max amount of extents
that could be required to perform an operation and, if reservation
succeeds, be sure that we're safe from any kind of OOM up until we
have finished the operation.

But it's only known to specific data structures which amount of memory
will be required for them to perform their operations, and all current
memtx structures (BPS tree, light hash, rtree, bitset) use matras to
deal with memory allocations, so it would be nice to make them able to
reserve exactly as much extents as they need themselves instead of
reserving max sane amount of ones as it's done now (see the constants:
`RESERVE_EXTENTS_BEFORE_REPLACE` and `RESERVE_EXTENTS_BEFORE_DELETE`).

The problem is that if we reserve extents in each matras instance we
will possibly be wasting a lot of memory in each index, so it would
be much better to share the reserved extents as it's done in the memtx
engine now.

So, in order to allow indexes reserve memoy for themselves and share
the reserved memory amoug all indexes, let's introduce a new entity:
`matras_allocator`.

The entity is to be shared amoug all memtx indexes and it allows to
reserve memory for indexes for various purposes. So instead of matras
holding the reserved extents, it's the deal of the new shared entity.

And the entity is exactly what the small submodule bump introduces.

Needed for tarantool/tarantool-ee#822

NO_DOC=submodule bump
NO_TEST=submodule bump
NO_CHANGELOG=submodule bump

### DIFF
--- a/perf/light.cc
+++ b/perf/light.cc
@@ -7,6 +7,8 @@
 
 #include <benchmark/benchmark.h>
 
+#include <small/matras.h>
+
 // This test contains benchmarks for Light - data structure implementing
 // Tarantool HASH index and being hash table under the hood.
 //
@@ -195,26 +197,21 @@ namespace {
 	};
 
 	static constexpr std::size_t light_extent_size = 16 * KB;
-	static std::size_t extents_count = 0;
 
 	inline void *
-	light_malloc_extend(void *ctx)
+	light_malloc_extend(struct matras_allocator *allocator)
 	{
-		std::size_t *p_extents_count = (size_t *) ctx;
-		assert(p_extents_count == &extents_count);
-		++*p_extents_count;
+		(void)allocator;
 		return malloc(light_extent_size);
 	}
 
 	inline void
-	light_free_extend(void *ctx, void *p)
+	light_free_extend(struct matras_allocator *allocator, void *p)
 	{
-		size_t *p_extents_count = (size_t *) ctx;
-		assert(p_extents_count == &extents_count);
-		--*p_extents_count;
+		(void)allocator;
 		free(p);
 	}
-}; // namespace
+}; /* namespace */
 
 #define LIGHT_NAME
 #define LIGHT_DATA_TYPE const TupleRaw *
@@ -476,12 +473,14 @@ class Light {
 public:
 	Light()
 	{
-		light_create(&ht, 0, light_extent_size, light_malloc_extend,
-			     light_free_extend, &extents_count, nullptr);
+		matras_allocator_create(&allocator, light_extent_size,
+					light_malloc_extend, light_free_extend);
+		light_create(&ht, 0, &allocator, nullptr);
 	}
 	~Light()
 	{
 		light_destroy(&ht);
+		matras_allocator_destroy(&allocator);
 	}
 
 	// Functions return smth in order to suppress "error: invalid use of void expression".
@@ -519,8 +518,7 @@ public:
 	clear()
 	{
 		light_destroy(&ht);
-		light_create(&ht, 0, light_extent_size, light_malloc_extend,
-			     light_free_extend, &extents_count, nullptr);
+		light_create(&ht, 0, &allocator, nullptr);
 	}
 
 	void
@@ -541,6 +539,7 @@ public:
 	}
 private:
 	struct light_core ht;
+	struct matras_allocator allocator;
 };
 
 using USet = std::unordered_set<TupleRef, Hash::TupleHash, TupleEqual>;
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -2173,22 +2173,6 @@ memtx_index_extent_free(struct matras_allocator *matras_allocator, void *extent)
 	mempool_free(&memtx->index_extent_pool, extent);
 }
 
-/**
- * Reserve num extents in pool.
- * Ensure that next num extent_alloc will succeed w/o an error
- */
-int
-memtx_index_extent_reserve(struct memtx_engine *memtx, int num)
-{
-	ERROR_INJECT(ERRINJ_INDEX_ALLOC, {
-		/* same error as in mempool_alloc */
-		diag_set(OutOfMemory, MEMTX_EXTENT_SIZE,
-			 "mempool", "new slab");
-		return -1;
-	});
-	return matras_allocator_reserve(&memtx->index_extent_allocator, num);
-}
-
 bool
 memtx_index_def_change_requires_rebuild(struct index *index,
 					const struct index_def *new_def)
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -92,6 +92,13 @@ memtx_tuple_new_raw_impl(struct tuple_format *format, const char *data,
 static void
 memtx_engine_run_gc(struct memtx_engine *memtx, bool *stop);
 
+static void *
+memtx_index_extent_alloc(struct matras_allocator *matras_allocator);
+
+static void
+memtx_index_extent_free(struct matras_allocator *matras_allocator,
+			void *extent);
+
 template<class ALLOC>
 static void
 memtx_alloc_init(void)
@@ -220,12 +227,7 @@ memtx_engine_free(struct engine *engine)
 	mempool_destroy(&memtx->iterator_pool);
 	if (mempool_is_initialized(&memtx->rtree_iterator_pool))
 		mempool_destroy(&memtx->rtree_iterator_pool);
-	void *p = memtx->reserved_extents;
-	while (p != NULL) {
-		void *next = *(void **)p;
-		memtx_index_extent_free(memtx, p);
-		p = next;
-	}
+	matras_allocator_destroy(&memtx->index_extent_allocator);
 	mempool_destroy(&memtx->index_extent_pool);
 	slab_cache_destroy(&memtx->index_slab_cache);
 	mh_ptr_delete(memtx->malloc_extents);
@@ -1771,11 +1773,13 @@ memtx_engine_new(const char *snap_dirname, bool force_recovery,
 	slab_cache_create(&memtx->index_slab_cache, &memtx->arena);
 	mempool_create(&memtx->index_extent_pool, &memtx->index_slab_cache,
 		       MEMTX_EXTENT_SIZE);
+	matras_allocator_create(&memtx->index_extent_allocator,
+				MEMTX_EXTENT_SIZE,
+				memtx_index_extent_alloc,
+				memtx_index_extent_free);
 	matras_stats_create(&memtx->index_extent_stats);
 	mempool_create(&memtx->iterator_pool, cord_slab_cache(),
 		       MEMTX_ITERATOR_SIZE);
-	memtx->num_reserved_extents = 0;
-	memtx->reserved_extents = NULL;
 	memtx->malloc_extents = mh_ptr_new();
 
 	memtx->state = MEMTX_INITIALIZED;
@@ -2117,17 +2121,12 @@ create_memtx_tuple_format_vtab(struct tuple_format_vtab *vtab)
 /**
  * Allocate a block of size MEMTX_EXTENT_SIZE for memtx index
  */
-void *
-memtx_index_extent_alloc(void *ctx)
+static void *
+memtx_index_extent_alloc(struct matras_allocator *matras_allocator)
 {
-	struct memtx_engine *memtx = (struct memtx_engine *)ctx;
-	if (memtx->reserved_extents) {
-		assert(memtx->num_reserved_extents > 0);
-		memtx->num_reserved_extents--;
-		void *result = memtx->reserved_extents;
-		memtx->reserved_extents = *(void **)memtx->reserved_extents;
-		return result;
-	}
+	struct memtx_engine *memtx = container_of(matras_allocator,
+						  struct memtx_engine,
+						  index_extent_allocator);
 	ERROR_INJECT(ERRINJ_INDEX_ALLOC, { goto fail; });
 	void *ret;
 	while ((ret = mempool_alloc(&memtx->index_extent_pool)) == NULL) {
@@ -2159,10 +2158,12 @@ fail:
 /**
  * Free a block previously allocated by memtx_index_extent_alloc
  */
-void
-memtx_index_extent_free(void *ctx, void *extent)
+static void
+memtx_index_extent_free(struct matras_allocator *matras_allocator, void *extent)
 {
-	struct memtx_engine *memtx = (struct memtx_engine *)ctx;
+	struct memtx_engine *memtx = container_of(matras_allocator,
+						  struct memtx_engine,
+						  index_extent_allocator);
 	mh_int_t p = mh_ptr_find(memtx->malloc_extents, extent, NULL);
 	if (p != mh_end(memtx->malloc_extents)) {
 		mh_ptr_del(memtx->malloc_extents, p, NULL);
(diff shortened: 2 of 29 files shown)


src/box/vy_cache.c
```diff
@@ -52,6 +52,24 @@ enum {
...
 
+static void *
+vy_cache_tree_page_alloc(struct matras_allocator *allocator)
+{
+	(void)allocator;
+	void *ret = malloc(VY_CACHE_TREE_EXTENT_SIZE);
+	if (ret == NULL)
+		diag_set(OutOfMemory, VY_CACHE_TREE_EXTENT_SIZE, "malloc",
+			 "ret");
+	return ret;
```
mkostoevr, line 63: The checkpatch complains about this malloc call, which is copied from below to allow specify the functions on the matras allocator construction.

Replace with xmalloc?

src/box/vy_mem.c
```diff
@@ -126,6 +130,8 @@ void
 vy_mem_delete(struct vy_mem *index)
 {
 	index->env->tree_extent_size -= index->tree_extent_size;
+	/* TODO: Destroy the index->tree? */
+	matras_allocator_destroy(&index->matras_allocator);
```
mkostoevr, line 134: Shouldn't we destroy the mem tree here?

locker: There's no need, because the `lsregion` allocator doesn't support freeing memory blocks. Destroying the tree would result in unnecessary work.

src/box/memtx_engine.h
```diff
@@ -266,21 +259,21 @@ extern struct tuple *
...
-memtx_index_extent_free(void *ctx, void *extent);
+memtx_index_extent_free(struct matras_allocator *allocator, void *extent);
 
 /**
  * Reserve num extents in pool.
  * Ensure that next num extent_alloc will succeed w/o an error
  */
 int
-memtx_index_extent_reserve(struct memtx_engine *memtx, int num);
+memtx_index_extent_reserve(struct memtx_engine *memtx, unsigned num);
```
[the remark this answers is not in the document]
mkostoevr: Made the methods static.

<details>

<summary>diff</summary>

```diff
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -92,6 +92,13 @@ memtx_tuple_new_raw_impl(struct tuple_format *format, const char *data,
 static void
 memtx_engine_run_gc(struct memtx_engine *memtx, bool *stop);
 
+static void *
+memtx_index_extent_alloc(struct matras_allocator *matras_allocator);
+
+static void
+memtx_index_extent_free(struct matras_allocator *matras_allocator,
+                       void *extent);
+
 template<class ALLOC>
 static void
 memtx_alloc_init(void)
@@ -2114,7 +2121,7 @@ create_memtx_tuple_format_vtab(struct tuple_format_vtab *vtab)
 /**
  * Allocate a block of size MEMTX_EXTENT_SIZE for memtx index
  */
-void *
+static void *
 memtx_index_extent_alloc(struct matras_allocator *matras_allocator)
 {
        struct memtx_engine *memtx = container_of(matras_allocator,
@@ -2151,7 +2158,7 @@ memtx_index_extent_alloc(struct matras_allocator *matras_allocator)
 /**
  * Free a block previously allocated by memtx_index_extent_alloc
  */
-void
+static void
 memtx_index_extent_free(struct matras_allocator *matras_allocator, void *extent)
 {
        struct memtx_engine *memtx = container_of(matras_allocator,
@@ -2171,7 +2178,7 @@ memtx_index_extent_free(struct matras_allocator *matras_allocator, void *extent)
  * Ensure that next num extent_alloc will succeed w/o an error
  */
 int
-memtx_index_extent_reserve(struct memtx_engine *memtx, unsigned num)
+memtx_index_extent_reserve(struct memtx_engine *memtx, int num)
 {
        ERROR_INJECT(ERRINJ_INDEX_ALLOC, {
                /* same error as in mempool_alloc */
diff --git a/src/box/memtx_engine.h b/src/box/memtx_engine.h
index ecc3620647..b13ed9d4b6 100644
--- a/src/box/memtx_engine.h
+++ b/src/box/memtx_engine.h
@@ -254,26 +254,12 @@ extern struct tuple *
 (*memtx_tuple_new_raw)(struct tuple_format *format, const char *data,
                       const char *end, bool validate);
 
-/**
- * Allocate a block of size MEMTX_EXTENT_SIZE for memtx index
- * @ctx must point to memtx engine
- */
-void *
-memtx_index_extent_alloc(struct matras_allocator *allocator);
-
-/**
- * Free a block previously allocated by memtx_index_extent_alloc
- * @ctx must point to memtx engine
- */
-void
-memtx_index_extent_free(struct matras_allocator *allocator, void *extent);
-
 /**
  * Reserve num extents in pool.
  * Ensure that next num extent_alloc will succeed w/o an error
  */
 int
-memtx_index_extent_reserve(struct memtx_engine *memtx, unsigned num);
+memtx_index_extent_reserve(struct memtx_engine *memtx, int num);
 
 /**
  * Generic implementation of index_vtab::def_change_requires_rebuild,
```

</details>

Regarding the reservation method, it seems more natural to have it in the engine, cause the memtx index extent allocator is under the memtx engine's responsibility.

src/box/vy_cache.h
```diff
@@ -119,6 +119,8 @@ struct vy_cache_env {
 	struct rlist cache_lru;
 	/** Common mempool for vy_cache_node struct */
 	struct mempool cache_node_mempool;
+	/** Common matras allocator for vy_cache_node struct */
```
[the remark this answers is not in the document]
mkostoevr: Fixed.

```diff
--- a/src/box/vy_cache.h
+++ b/src/box/vy_cache.h
@@ -119,7 +119,7 @@ struct vy_cache_env {
        struct rlist cache_lru;
        /** Common mempool for vy_cache_node struct */
        struct mempool cache_node_mempool;
-       /** Common matras allocator for vy_cache_node struct */
+       /** Common matras extent allocator. */
        struct matras_allocator allocator;
        /** Size of memory occupied by cached tuples */
```

src/box/sequence.c
```diff
@@ -85,11 +85,12 @@ sequence_data_equal_key(struct sequence_data data, uint32_t id)
 
 static struct light_sequence_core sequence_data_index;
 static struct mempool sequence_data_extent_pool;
+static struct matras_allocator allocator;
```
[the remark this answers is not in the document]
mkostoevr: Renamed.

src/box/memtx_engine.cc
```diff
@@ -2177,33 +2171,15 @@ memtx_index_extent_free(void *ctx, void *extent)
  * Ensure that next num extent_alloc will succeed w/o an error
  */
 int
-memtx_index_extent_reserve(struct memtx_engine *memtx, int num)
+memtx_index_extent_reserve(struct memtx_engine *memtx, unsigned num)
```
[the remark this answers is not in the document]
mkostoevr: Rolled back the type change.

perf/light.cc
```diff
@@ -198,32 +213,19 @@ namespace {
...
 	}
 
 	inline void
-	light_free_extend(void *ctx, void *p)
+	light_free_extend(struct matras_allocator *allocator, void *p)
 	{
-		size_t *p_extents_count = (size_t *) ctx;
-		assert(p_extents_count == &extents_count);
-		--*p_extents_count;
+		--extents_count;
```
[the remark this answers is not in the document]
mkostoevr: Done.

perf/light.cc
```diff
@@ -189,7 +189,22 @@ namespace {
...
+#define LIGHT_DATA_TYPE const TupleRaw *
+#define LIGHT_KEY_TYPE Key_t
+#define LIGHT_CMP_ARG_TYPE int
+#define LIGHT_EQUAL(a, b, arg) tuple_equals(a, b)
+#define LIGHT_EQUAL_KEY(a, b, arg) key_equals(a, b)
+
+#include "salad/light.h"
+
+/**
+ * The `struct matras_allocator` is unknown up to this point,
```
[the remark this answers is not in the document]
mkostoevr: Done.

src/box/vy_mem.c
mkostoevr: So should I drop the matras allocator deinitialization too? It only frees the reserved blocks.

src/box/vy_cache.c
mkostoevr: Done.

```diff
--- a/src/box/vy_cache.c
+++ b/src/box/vy_cache.c
@@ -56,11 +56,7 @@ static void *
 vy_cache_tree_page_alloc(struct matras_allocator *allocator)
 {
        (void)allocator;
-       void *ret = malloc(VY_CACHE_TREE_EXTENT_SIZE);
-       if (ret == NULL)
-               diag_set(OutOfMemory, VY_CACHE_TREE_EXTENT_SIZE, "malloc",
-                        "ret");
-       return ret;
+       return xmalloc(VY_CACHE_TREE_EXTENT_SIZE);
 }
 
 static void
```

src/box/vy_mem.c
locker: I guess so. Let's add a comment explaining why we don't call them.

mkostoevr: Fixed.

```diff
--- a/src/box/vy_mem.c
+++ b/src/box/vy_mem.c
@@ -130,8 +130,13 @@ void
 vy_mem_delete(struct vy_mem *index)
 {
        index->env->tree_extent_size -= index->tree_extent_size;
-       /* TODO: Destroy the index->tree? */
-       matras_allocator_destroy(&index->matras_allocator);
+       /*
+        * It would be expected to destroy the index->matras_allocator and
+        * the index->tree here, but they simply free allocated and reserved
+        * memory blocks, and the lsregion used for allocation of the memory
+        * does not support freeing arbitrary blocks, so they're effectively
+        * no-ops. Let's omit them.
+        */
        tuple_format_unref(index->format);
        fiber_cond_destroy(&index->pin_cond);
        TRASH(index);
```

mkostoevr: Fixed.

```diff
--- a/src/box/vy_mem.c
+++ b/src/box/vy_mem.c
@@ -130,8 +130,13 @@ void
 vy_mem_delete(struct vy_mem *index)
 {
        index->env->tree_extent_size -= index->tree_extent_size;
-       /* TODO: Destroy the index->tree? */
-       matras_allocator_destroy(&index->matras_allocator);
+       /*
+        * It would be expected to destroy the index->matras_allocator and
+        * the index->tree here, but they simply free allocated and reserved
+        * memory blocks, and the lsregion used for allocation of the memory
+        * does not support freeing arbitrary blocks, so they're effectively
+        * no-ops. Let's omit them.
+        */
        tuple_format_unref(index->format);
        fiber_cond_destroy(&index->pin_cond);
        TRASH(index);
```

src/box/vy_mem.c
```diff
@@ -114,9 +115,12 @@ vy_mem_new(struct vy_mem_env *env, struct key_def *cmp_def,
...
 	index->format = format;
 	tuple_format_ref(format);
+	matras_allocator_create(&index->matras_allocator,
+				VY_MEM_TREE_EXTENT_SIZE,
+				vy_mem_tree_extent_alloc,
+				vy_mem_tree_extent_free);
 	vy_mem_tree_create(&index->tree, cmp_def,
-			   vy_mem_tree_extent_alloc,
-			   vy_mem_tree_extent_free, index, NULL);
+			   &index->matras_allocator, NULL);
```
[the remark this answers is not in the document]
locker: Technically, I think we could use the global generation counter for allocations from a `vy_mem` because only the newest `vy_mem` is active (the ones that use older generations are sealed for modifications). Moreover, allocating a block with an older generation counter would trigger an assertion in `lsregion_alloc`. I'll see what we can do about it. For now, I guess we can leave it as is.

BTW we must never reserve blocks with this allocator because it's garbage-collected externally. Seems fine because we don't use matras read views in Vinyl.

src/box/memtx_engine.cc
```diff
@@ -2185,25 +2186,7 @@ memtx_index_extent_reserve(struct memtx_engine *memtx, int num)
...
-			diag_set(OutOfMemory, MEMTX_EXTENT_SIZE,
-				 "mempool", "new slab");
-			return -1;
-		}
-		*(void **)ext = memtx->reserved_extents;
-		memtx->reserved_extents = ext;
-		memtx->num_reserved_extents++;
-	}
-	return 0;
+	return matras_allocator_reserve(&memtx->index_extent_allocator, num);
```
locker, line 2189: The patch looks good to me, but I'd like to see block reservation done in indexes (rtree, bps, light) to make sure the new matras allocator API fits. Let's try it if it doesn't take too long.

src/lib/salad/rtree.c
```diff
@@ -1055,6 +1055,33 @@ rtree_remove(struct rtree *tree, const struct rtree_rect *rect, record_t obj)
...
+	/*
+	 * The data structure has no OOM checks, so we have to resreve the
+	 * amount of matras block allocations potentially enough to perform
+	 * the operations required. It's based on the amount of extents we
+	 * currently reserve in the memtx space prior to replace according
+	 * to the smallest page size: 512 (so we sum-up the reserved memory
+	 * and divide it into 512 and got 512 pages).
+	 */
+	if (matras_alloc_reserve(&tree->mtab, 512) != 0)
+		return -1;
```
locker, line 1071: I don't think it's a good idea to refer to the code we're deleting. What's the page size in rtree?

src/lib/salad/rtree.h
```diff
@@ -284,6 +284,20 @@ rtree_insert(struct rtree *tree, struct rtree_rect *rect, record_t obj);
...
+ * @return 0 on success, -1 on memory error
+ * @param tree - pointer to a tree
+ * @param old_rect - rectangle of the record to delete
+ * @param old_obj - record to delete
+ * @param new_rect - rectangle to insert
+ * @param new_obj - record to insert
+ * @param removed - set to true if the old_obj is removed
+ */
+int
+rtree_replace(struct rtree *tree, struct rtree_rect *old_rect, record_t old_obj,
```
locker, line 298: Please re-split. The first commit should introduce this method to the salad library and add a unit test. The second commit should patch the memtx engine to use this method instead of `rtree_insert` + `rtree_delete` + reservation.

src/lib/salad/rtree.c
mkostoevr: 512 for 1-dimensional tree, 8192 for 20-dimentional tree.

We reserve 16 extents prior to replace, size of an extent is 16kb, so in total we reserve 16 * 16kb = 262144 bytes.

This is either 512 1-dimentional tree pages or 32 20-dimentional tree pages.

So I took the higher value, now we have

locker: AFAIK the most common tree is 2-dimensional, not 1-dimensional.

I took a quick look at the rtree source code and found the constant `RTREE_MAX_HEIGHT` equal to 16. Can you deduce the reservation count from it? Wouldn't `2 * max height` be always enough?

src/lib/salad/bps_tree.h
```diff
@@ -814,38 +814,47 @@ bps_tree_insert_get_offset(struct bps_tree *tree, bps_tree_elem_t new_elem,
...
- * @param elem - the element tot delete
- * @return - 0 on success or -1 if the element was not found in tree
+ * @param elem - the element to delete
+ * @param deleted - pointer to the deleted element or NULL
+ * @return - 0 on success, -1 on memory error
  */
 static inline int
-bps_tree_delete(struct bps_tree *tree, bps_tree_elem_t elem);
+bps_tree_delete(struct bps_tree *tree, bps_tree_elem_t elem,
+		bps_tree_elem_t **deleted);
```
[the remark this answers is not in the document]
locker: Also, the commit message should start with "salad: ...", not "bps: ...".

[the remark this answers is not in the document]
mkostoevr: Changed the commit message and fixed the `bps_vec_delete` signature in a separated commit.

src/box/memtx_engine.cc
mkostoevr: Discussed f2f, it's been decided to finish with RTREE in the future: #10831.

src/lib/salad/rtree.c
mkostoevr: [Deferred](https://github.com/tarantool/tarantool/pull/10734#discussion_r1847764459) this part of the patch.

src/lib/salad/rtree.h
mkostoevr: [Deferred](https://github.com/tarantool/tarantool/pull/10734#discussion_r1847764459) this part of the patch.

locker (approved):

Issue #10756, opened 2024-10-29
config: support transformation of composite types in <schema object>:map()


Totktonada:
The `experimental.config.utils.schema` module has a `<schema object>:map()` method, which allows to transform scalar values:

https://github.com/tarantool/tarantool/blob/40fa44c2b1205642cb36cdee2bd85169606e0c0c/src/box/lua/config/utils/schema.lua#L1358-L1359

However, we can't transform a record, a map or an array value using it.

The proposal is to call the user-provided transformation function for these composite types too. I think that the pre-order traversal fits good here (and it is simpler to implement).

Since the `:apply_default()` method is implemented using the `:map()` method, it has the same constraint: defaults are applied only for scalar values. If we adjust the `:map()` method, the `:apply_default()` method will work for composite types too.

Lack of this functionality causes a couple of problems in config's code:

* #10728
* tarantool/tarantool-ee#963
* tarantool/tarantool-ee#964

I'm going to workaround these problems separately, without using the proposed `:map()` enhancement. However, it is good to implement the feature later and use it to eliminate the workarounds.

Jira issue: TNTP-243.

PR #10759, opened 2024-10-29
config: fix handling of audit_log.spaces removal

Totktonada:
If the option is removed from the YAML configuration, `config:reload()` should add an alert that a non-dynamic option is changed and it will be applied after a restart. This patch works the problem around. The root of the problem will be fixed in #10756.

Fixes tarantool/tarantool-ee#964

### DIFF
--- a/changelogs/unreleased/config-audit-log-spaces-removal.md
+++ b/changelogs/unreleased/config-audit-log-spaces-removal.md
@@ -1,0 +1,3 @@
+## bugfix/config
+
+* Fixed handling of `audit_log.spaces` option removal (ghe-964).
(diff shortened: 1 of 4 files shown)


Totktonada:
NB: I would like to revert this workaround in 3.3 (without backporting to `release/3.2`). Draft PR is here: #10761.

xuniq (approved):

georgiy-belyanin (approved):
Thank you for the patch. Everything seems to work.

Totktonada:
Rebased on top of a fresh `master`.

Totktonada:
Pushed to `master`, a future `3.3.0` release.

Backported to `release/3.2` (a future `3.2.2` release) in PR #10791.

Issue #10776, opened 2024-11-01
config: temporary exclude an instance from the replicaset


Totktonada:
The proposal is to add an option to temporary exclude an instance from the replicaset.

```yaml
groups:
  g:
    replicasets:
      r:
        instances:
          i-001: {}
          i-002: {}
          i-003: {}
          i-004:
            replication:
              disabled: true
```

When the option is enabled:

1. The instance can't be bootstrapped (a local snapshot is required to start).
2. `i-004` has an empty upstreams list (`box.cfg.replication`).
3. Other replicaset members have no `i-004` in the upstreams list.

The effect of the option is mostly equivalent to removal of the instance from the config, but it is sometimes more convenient to add such a marker instead. Useful for repairing activities.

Jira: TNTP-364

Totktonada:
It is too easy to feed stale data to a client this way (see https://github.com/tarantool/tarantool/pull/10780#discussion_r1826607048). I'll come up with some more consistent proposal.

PR #10780, opened 2024-11-01
config: add replication.disabled option

Totktonada:
The commit adds the main logic behind the `replication.disabled` option. It is useful to temporary exclude an instance from its replica set.

The instance with `replication.disabled = true` disconnects from all the upstreams.

All the other instances in the replica set refuse to fetch data from the disabled instance.

If the instance is non-anonymous it is kept in the `_cluster` system space, so the replication from/to it can be enabled back.

*(One bullet from #10776 is not solved here: refuse to bootstrap if the option is set. This functionality is to be implemented in a separate patch.)*

Part of #10776

### DIFF
--- a/changelogs/unreleased/config-replication-disabled.md
+++ b/changelogs/unreleased/config-replication-disabled.md
@@ -1,0 +1,4 @@
+## feature/config
+
+* Added the `replication.disabled` option to temporary exclude an instance from
+  its replica set (gh-10776).
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -26,7 +26,6 @@ local function peer_uris(configdata)
     local uris = {}
     for _, peer_name in ipairs(peers) do
         local iconfig_def = configdata._peers[peer_name].iconfig_def
-        local is_anon = instance_config:get(iconfig_def, 'replication.anon')
         -- Don't use anonymous replicas as upstreams.
         --
         -- An anonymous replica can't be an upstream for a
@@ -38,7 +37,14 @@ local function peer_uris(configdata)
         --
         -- A user may configure a custom data flow using
         -- `replication.peers` option.
-        if not is_anon then
+        local is_anon = instance_config:get(iconfig_def, 'replication.anon')
+        -- Don't replicate data from a disabled instance, because
+        -- it is likely disabled with a reason: an administrator
+        -- is performing some repairing work on it. We shouldn't
+        -- see this intermediate database state on other peers.
+        local is_disabled = instance_config:get(iconfig_def,
+            'replication.disabled')
+        if not is_anon and not is_disabled then
             local uri = instance_config:instance_uri(iconfig_def, 'peer',
                 {log_prefix = "replicaset dataflow configuration: "})
             if uri == nil then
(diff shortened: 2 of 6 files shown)


src/box/lua/config/applier/box_cfg.lua
```diff
@@ -341,9 +347,29 @@ local function apply(config)
         box_cfg.listen = box.NULL
     end
 
-    -- Construct box_cfg.replication.
-    if box_cfg.replication == nil then
+    -- luacheck: ignore 542 empty if branch
+    if configdata:get('replication.disabled', {use_default = true}) then
```
Totktonada, line 351: Let's save the configuration value into a variable to use in the second patch (where we forbid to start without a local snapshot).

Totktonada: Done.

xuniq (approved):

georgiy-belyanin (changes requested):
Thank you for the patch! Please consider my comments:

src/box/lua/config/instance_config.lua
```diff
@@ -1386,6 +1386,10 @@ return schema.new('instance_config', schema.record({
         }),
     }),
     replication = schema.record({
+        disabled = schema.scalar({
+            type = 'boolean',
+            default = false,
+        }),
```
Totktonada, line 1392: We already have `console.enabled`, `fiber.top.enabled`, `feedback.enabled`, `flightrec.enabled`. I think `replication.enabled` would be more consistent with other options.

changelogs/unreleased/config-replication-disabled.md
Totktonada, line 1: I'm still in some doubt regarding ability to have a traffic from clients on an instance that is marked this way. There are several reasons:

* Nothing forbids to mark a current replicaset leader this way by a mistake. So, replicas will not receive data updates. The replicas may feed stale data to clients.
* Even if the marked instance is a replica, nothing prevents a client traffic to reach it. This way the clients may receive stale data.

A few thoughts around:

* I don't like to mix RO/RW mode and disabled mode. One may need to have the disabled instance in the RW mode to perform some repairing work. We can move it to RO once and then let the admin use the low-level `box.cfg({read_only = true/false})` API, but it seems better to just do nothing.
* I think that it would be a good idea to stop processing iproto requests at all, when the instance is marked as disabled. My initial doubt was that some administration tools may use iproto (over `tt connect`). However, there is always ability to reach an admin console socket and as far as I see, admins usually use it.

If we agree that iproto requests shouldn't be served when the instance is marked as disabled, we need to decide how. Just `box.cfg({listen = box.NULL})` doesn't break current connections. It seems, we need an API to call `iproto_drop_connections` from `src/box/iproto.cc`.

changelogs/unreleased/config-replication-disabled.md
```diff
@@ -0,0 +1,4 @@
+## feature/config
+
+* Added the `replication.disabled` option to temporary exclude an instance from
+  its replica set (gh-10776).
```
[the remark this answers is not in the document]
Totktonada: I agree that it is better to reduce the scope of the option to one instance.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -341,9 +347,31 @@ local function apply(config)
...
+            'instance %q: the data is not replicated from or to this ' ..
+            'instance'):format(instance_name)
+        config._aboard:set({type = 'warn', message = warning})
+    elseif box_cfg.replication == nil then
+        -- Construct box_cfg.replication.
         box_cfg.replication = peer_uris(configdata)
+    else
+        -- Nothing to do: the peer URIs are set using the
+        -- replication.peers option. It is already in the box_cfg
+        -- table.
```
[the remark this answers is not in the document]
Totktonada: It is likely due to the safe startup mode (c80b121544f60740fd2605dc2e15f86c84480c59), where we can call the `applier/box_cfg` code twice for a configured leader during applying of one configuration.

My guess was that everything is OK with the alerts, because they're cleaned up before reloading. The fix should be to just set a `key` option with some unique alert identifier for the `<aboard object>:set()` call.

Totktonada:
Well, sometimes I need to write some code to turn the head on.

I think it is too easy to feed stale data to clients this way and we should actually disable processing of incoming iproto requests in such a 'disabled instance' mode.

I'm going to close this pull request and return here back with something more safe to use.

Issue #10796, opened 2024-11-07
config: add isolated mode for an instance


Totktonada:
The proposal is to add an option to temporary isolate an instance to perform replicaset repairing activities: debug a problem on the isolated instance (without affecting the non-isolated part) or extract some data from the isolated instance and apply them on the non-isolated part of the replicaset.

```yaml
groups:
  g:
    replicasets:
      r:
        instances:
          i-001: {}
          i-002: {}
          i-003: {}
          i-004:
            isolated: true
```

In this mode the instance doesn't perform processing of iproto requests, doesn't perform background data modifications and refuses to accept updates from other replicaset members. Also, other replicaset members don't replicate data from the isolated instance.

Technically speaking, when the option is enabled, the following actions are performed:

1. iproto stops listening for new connections.
2. iproto drops all the current connections.
3. The instance goes to read-only mode.
4. The instance is disconnected from all the replication upstreams.
5. Other replicaset members exclude the isolated instance from the replication upstreams.

One new startup constraint is added in the isolated mode:

1. The instance can't be bootstrapped (a local snapshot is required to start).

*(This is a continuation of ideas from #10776 -- but here I additionally propose to stop processing of iproto requests to reduce a chance to have user traffic on a stalled or broken instance.)*

Jira: TNTP-364
Documentation request: https://github.com/tarantool/doc/issues/4632

## Progress

- [x] https://github.com/tarantool/tarantool/pull/10798
- [x] https://github.com/tarantool/tarantool/pull/10801
- [x] https://github.com/tarantool/tarantool/pull/10808
- [x] https://github.com/tarantool/tarantool/pull/10811
- [x] https://github.com/tarantool/tarantool/pull/10813
- [x] https://github.com/tarantool/tarantool/pull/10816
- [ ] https://github.com/tarantool/tarantool/pull/10821

PR #10798, opened 2024-11-08
iproto: add Lua API to drop connections (internal)

Totktonada:
This low-level API is needed to prevent execution of user's requests over iproto, when an instance is marked as disabled/isolated. See details about this mode in #10796.

Part of #10796

### DIFF
--- a/src/box/lua/iproto.c
+++ b/src/box/lua/iproto.c
@@ -241,6 +241,29 @@ lbox_iproto_session_new(struct lua_State *L)
 }
 
 /**
+ * Drop all current connections.
+ *
+ * Accepts a timeout (in seconds).
+ *
+ * Returns nothing on success, raises an error on a failure.
+ *
+ * See iproto_drop_connections() for details.
+ */
+static int
+lbox_iproto_drop_connections(struct lua_State *L)
+{
+	int n_args = lua_gettop(L);
+	if (n_args != 1 || lua_type(L, 1) != LUA_TNUMBER)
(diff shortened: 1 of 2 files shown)


src/box/lua/iproto.c
```diff
@@ -240,6 +240,29 @@ lbox_iproto_session_new(struct lua_State *L)
...
+	if (n_args != 1 || lua_type(L, 1) != LUA_TNUMBER)
+		return luaL_error(L, "Usage: box.iproto.internal."
+				  "drop_connections(timeout)");
+	double timeout = lua_tonumber(L, 1);
+	int rc = iproto_drop_connections(timeout);
+	if (rc < 0)
+		return luaT_error(L);
+	return 0;
+}
+
```
georgiy-belyanin, line 265: Rough idea to be implemented later: use some (not sure what) default value, since in general the timeout is not so important. 

At the same time, this internal handle seems useful for tarantool/tarantool-ee#971

georgiy-belyanin (approved):
Thank you for the patch (I wanna be better than Copilot AI)! LGTM.

src/box/lua/iproto.c
Totktonada: TBH, I don't understand why it may take some visible time. A couple of shutdown+close calls, plus inter-thread messages. No synchronous network transmissions, I guess.

So maybe it shouldn't report any error, shouldn't have a timeout and if something goes bad, should just log it.

I'm not ready to make this decision now, because I didn't investigate/test it against these questions.

PR #10799, opened 2024-11-08
Public C API for sending work to TX

Gerold103:
See #10635 and the commits.


src/box/lua/config/instance_config.lua
```diff
@@ -741,6 +741,11 @@ return schema.new('instance_config', schema.record({
             box_cfg = 'worker_pool_threads',
             default = 4,
         }),
+        tx_user_pool_size = schema.scalar({
+            type = 'number',
```
Totktonada, line 745: `'integer'` fits better here -- at least a better diagnostic is reported if a non-integral number is passed.

I see, there are options, where `'number'` is used in the similar situation, but these are just mistakes.

Feel free to ignore, this is a negligible point.

src/lib/core/cbus.h
```diff
@@ -217,12 +217,11 @@ cpipe_submit_flush(struct cpipe *pipe)
 static inline void
 cpipe_push(struct cpipe *pipe, struct cmsg *msg)
 {
-	assert(loop() == pipe->producer);
```
locker: AFAICS the producer can be either `loop()` or `NULL` - we never set any other value for it. So I wouldn't let the user pass an arbitrary loop. Instead, I'd introduce a new function for creating a "loop-less" pipe (`cpipe_create_noev`?). IMO that would be more explicit. Then we wouldn't need to patch all the places where `cpipe_create` is called. I'd also leave this assertion be.

changelogs/unreleased/tnt_tx_push.md
```diff
@@ -0,0 +1,7 @@
+## feature/box
```
locker, line 1: Should probably be `feature/core`.

I also wouldn't go too deep into details in the release notes. For example, I wouldn't mention the configuration parameter - it should go to the documentation.

test/app-tap/exports.test.lua
```diff
@@ -52,6 +52,8 @@ local check_symbols = {
     'tnt_mp_sizeof_error',
     'tnt_mp_sizeof_decimal',
     'tnt_mp_sizeof_uuid',
+    'tnt_tx_flush',
+    'tnt_tx_push',
```
locker, line 56: I think this belongs in the previous patch.

test/box-luatest/tx_user_fiber_pool_test.lua
locker, line 1: It's a pity that a pure application-server feature that doesn't require any box methods uses `box.cfg` for configuration and is tested in the `box-luatest` test suite.

Maybe, a tweak would be enough to configure the pool size for now? Do we actually expect anyone to tune this parameter? Maybe postpone introduction of the configuration option until we receive a feedback from users?

src/lib/core/cbus.h
Gerold103: I can't, unfortunately. In the tests I have a worker thread with `loop() != NULL` which creates a ev-less pipe having `producer == NULL`.

test/box-luatest/tx_user_fiber_pool_test.lua
Gerold103: Good observation.

Not exposing any option I don't like. We do know that `net_msg_max` is too small for many people. There is no one universal value working for everyone. Given that this PR feature's target usage is the same as IProto (handling network requests), it is doomed to face the same problem. Lets make it configurable from the start.

Speaking of how to configure it - I can suggest to make it configurable via YAML config + via `fiber.tx_user_pool_size(new_val) -- return old val`. Then `box.cfg` isn't needed, and people who don't write Lua code still can configure it using their YAML config.

I also thought about tweaks instead `fiber.tx_user_pool_size()`, but the tweaks aren't public, are they? This would force people to use either an internal API or the YAML config. Users with the legal usage of public Lua API would be deprived.

What do you think about the YAML + `fiber.tx_user_pool_size()` suggestion?

src/box/lua/config/instance_config.lua
Gerold103: Fixed, thanks for noticing!

changelogs/unreleased/tnt_tx_push.md
Gerold103: Done.

test/app-tap/exports.test.lua
Gerold103: Indeed, thanks for the notice, done.

test/box-luatest/tx_user_fiber_pool_test.lua
locker: > What do you think about the YAML + `fiber.tx_user_pool_size()` suggestion?

I like it. The fiber Lua function could be introduced in the same patch with the new C API methods while the last one would use it in the YaML config. I also think it's better to move the functional tests to the previous patch as well while the last patch (the one about the config) would just check that the configuration module properly forwards the value to `fiber.tx_user_pool_size()`.

sergepetrenko (approved):
Thanks for the patch!
Nicely done. I have no objections.

src/lib/core/tnt_thread.cc
```diff
@@ -0,0 +1,108 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#include "tnt_thread.h"
+
+#include "cbus.h"
+#include "fiber_pool.h"
+
+namespace {
```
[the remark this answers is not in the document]
Gerold103: In C++ the class methods can't be declared static like in C (`static` keyword makes them not require an object, but the symbol name is still global). They can only be hidden using an anon namespace. In this case I was protecting `TXPipe::TXPipe()` and `TXPipe::~TXPipe()`.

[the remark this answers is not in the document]
sergepetrenko: Ok, I see. Thanks for the explanation!

p7nov (approved):

test/box-luatest/tx_user_fiber_pool_test.lua
Gerold103: Done! Please, have a look.

Totktonada (approved):
Looked on the configuration part -- no objections.

locker (approved):

Issue #6281, opened 2021-08-01
Harness raft 'learner' feature


Gerold103:
With the current implementation of raft it is possible to disable all synchronous transactions by adding a new instance to the cluster until it is fully joined and up-to-date.

The simplest example: there is 1 node with the canonical (`replication_synchro_quorum = 'N/2 + 1'`). It works fine, but has lots of data. If now a second node is added, the first node won't be able to commit any synchronous data because the quorum became 2, but the second node is busy with long initial and then final join stages.

The same will happen for a cluster having multiple nodes and getting multiple new instances. For example, cluster of 3 nodes, and it gets 3 new nodes - the quorum becomes 4 and the synchronous replication is dead until at least one of the new nodes finished its join stages.

In etcd the problem is solved by having a special role for such instances: https://etcd.io/docs/v3.3/learning/learner/ - learner. They do not affect the quorum, can't vote, can't be leaders. But they can join and get up to date with the leader before entering the quorum.

Need to check if this issue actually exists in Tarantool (make a test, perhaps with an error injection). Then try to add the new role.

In the etcd approach, if I understood it correctly, I do not like that a user needs to set the role manually. In Tarantool it might be done automatically, I hope. Make all new nodes learners until they finish at least both join stages, and then turn of the actual `election_mode` specified in the config.

Serpentian:
The issue indeed exists in Tarantool. Here's the reproducer: https://github.com/Serpentian/tarantool/tree/gh-6281-learner-role.
It can be run with: `./test-run.py --builddir ../build replication-luatest/gh_6281_raft_learner_feature_test.lua`

PR #10805, opened 2024-11-11
config: add failover learners section

georgiy-belyanin:
This patch adds the support for `failover.replicasets.<replicaset-name>.learners` configuration section and corresponding checks for it.

This option is used to specify the list of instances from the specific replicaset that can't be selected as a master by the supervised failover coordinator.

The supervised coordinator is enterprise-edition feature.

EE issue: tarantool/tarantool-ee#991
EE PR: tarantool/tarantool-ee#992

Jira task: [TNTP-192](https://jira.vk.team/browse/TNTP-192)

### DIFF
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2325,6 +2325,13 @@ return schema.new('instance_config', schema.record({
                 type = 'string',
             }),
             value = schema.record({
+                -- Instances that can't be chosen as a master
+                -- by the supervised failover coordinator.
+                learners = schema.array({
+                    items = schema.scalar({
+                        type = 'string',
+                    }),
+                }),
                 -- Priorities for the supervised failover mode.
                 priority = schema.map({
                     key = schema.scalar({
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -592,6 +592,23 @@ local function validate_failover_config(instances, failover_config)
         replicasets[def.replicaset_name] = true
     end
 
+    local function verify_instance_in_replicaset(instance_name, replicaset_name)
+        if instances[instance_name] == nil then
+            error(('instance %s from replicaset %s specified in the '..
+                   'failover.replicasets section doesn\'t exist')
+                  :format(instance_name, replicaset_name), 0)
+        end
+
+        local instance_replicaset = instances[instance_name].replicaset_name
+        if instance_replicaset ~= replicaset_name then
+            error(('instance %s from replicaset %s is specified in ' ..
+                   'the wrong replicaset %s in the failover.replicasets ' ..
+                   'configuration section')
+                  :format(instance_name, instance_replicaset,
+                          replicaset_name), 0)
+        end
+    end
+
     for replicaset_name, replicaset in pairs(failover_config.replicasets) do
         if replicasets[replicaset_name] == nil then
             error(('replicaset %s specified in the failover configuration '..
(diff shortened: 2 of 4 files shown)


src/box/lua/config/instance_config.lua
```diff
@@ -2325,6 +2325,13 @@ return schema.new('instance_config', schema.record({
                 type = 'string',
             }),
             value = schema.record({
+                -- The instances that can't be appointed by the
+                -- supervised failover coordinator.
+                non_appointable = schema.array({
```
Totktonada, line 2330: Raw idea: it would look simpler if we call it just `read_only`.

georgiy-belyanin: Hm, sounds good, though the name could be easily misunderstood since the instance could be RW meanwhile marked as a `read_only`. E.g. right after bootstrapping.

Totktonada: Hm. Now I'm in doubt about this decision.

Quote:

> Though, it was problematic in case of all instances are marked as non-appointable. I don't want to forbid such configuration. Since bootstrap leader anyway goes RO after a while I suppose it's ok not to take the appointable status into account when choosing the bootstrap leader.

We don't have to explicitly forbid it. Just let all the instances wait a bootstrap leader. It would be a straightforward behavior.

Let's imagine that this `non_appointable` instance is expressed in config as an instance with `priority` = `-inf` and apply all the related logic.

TBH, I would think twice about naming to avoid `non_foo` options.

1. `read_only`
2. `learners`
3. `readers`

Something like this would sound good for me. `read_only` seems preferable, since it is consistent with other APIs.

Terminology questions are always hard, so I would appreciate if you ask for more opinions here.

Totktonada: The `learner` word is likely recognized by others, see #6281.

Also, `follower` seems quite friendly term too.

Maybe it is easier to use `learner`/`follower` in sentences comparing to `read_only`, which interferes with other parameters.

grafin: In cartridge similar option was called `electable` (https://github.com/tarantool/cartridge/commit/06147dd891b89f6ba90d9366892a0822b3fc995f), but I'm ok with `learner/follower`, I'm against `read_only` because it is kinda confusing with `box.cfg.read_only` and `non_appointable` because I don't like the negative prefix)

grafin (approved):
Thanks for the patch! I see no major problems here, except for the option name itself, commented under the corresponding conversation.

src/box/lua/config/instance_config.lua
georgiy-belyanin: Thank you for quite detailed discussion and experienced opinion on it. I've changed the option name to `learners`

Totktonada (approved):

Totktonada:
@georgiy-belyanin I think that we should exclude the learners from candidates to bootstrap leaders, here:

https://github.com/tarantool/tarantool/blob/f65de7e71ec1a1f80cf960ba653b672ba358c74a/src/box/lua/config/configdata.lua#L916-L937

We can solve it in a follow-up patch, though. I'm going to push the option, because I have no objections regarding the option itself.

georgiy-belyanin:
> @georgiy-belyanin I think that we should exclude the learners from candidates to bootstrap leaders, here:
> 
> https://github.com/tarantool/tarantool/blob/f65de7e71ec1a1f80cf960ba653b672ba358c74a/src/box/lua/config/configdata.lua#L916-L937
> 
> We can solve it in a follow-up patch, though. I'm going to push the option, because I have no objections regarding the option itself.

Filed an issue #10842

PR #10808, opened 2024-11-11
config: don't bootstrap database in isolated mode

Totktonada:
In the isolated mode an instance refuses to fetch any data from other replicaset members. An attempt to bootstrap a database on such an instance would produce its own database (for example, it assigns its own replicaset UUID).

Unlikely a user wants to place two instances into the same replicaset in the configuration and get two different databases.

Part of #10796

### DIFF
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -775,6 +775,35 @@ local function validate_misplacing(cconfig)
     end
 end
 
+-- Check startup conditions.
+local function validate_startup(instance_name, iconfig_def)
+    local is_startup = type(box.cfg) == 'function'
+    if not is_startup then
+        return
+    end
+
+    local no_snap = snapshot.get_path(iconfig_def) == nil
+    local isolated = instance_config:get(iconfig_def, 'isolated')
+
+    -- Forbid startup without a local snapshot in the isolated
+    -- mode.
+    if no_snap and isolated then
+        error(('Startup failure.\nThe isolated mode is enabled and the ' ..
+            'instance %q has no local snapshot. An attempt to bootstrap ' ..
+            'the instance would lead to the split-brain situation.'):format(
+            instance_name), 0)
+    end
+
+    -- TODO: There is a situation, which looks similar, but we
+    -- don't report an error in the case. It is a startup without
+    -- a local snapshot with replication.peers configured as an
+    -- empty list if there are other instances in the replicaset.
+    -- Are there cases, when it is OK? Maybe if the instance is
+    -- assigned as a bootstrap leader? Now we pass it over, but
+    -- maybe it worth to revisit it later and report an error in
+    -- some definitely/likely erroreous cases.
+end
+
 local function new(iconfig, cconfig, instance_name)
     -- Find myself in a cluster config, determine peers in the same
     -- replicaset.
@@ -920,6 +949,9 @@ local function new(iconfig, cconfig, instance_name)
             instance_uuid = instance_uuid,
         }, iconfig_def)
     end
+
+    -- A couple of checks that are only performed on startup.
+    validate_startup(instance_name, iconfig_def)
 
     return setmetatable({
         _iconfig = iconfig,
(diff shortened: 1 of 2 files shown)


src/box/lua/config/configdata.lua
```diff
@@ -775,6 +775,36 @@ local function validate_misplacing(cconfig)
...
+    local isolated = instance_config:get(iconfig_def, 'isolated')
+
+    -- Forbid startup without a local snapshot in the isolated
+    -- mode.
+    if no_snap and isolated then
+        error(('Startup failure.\nThe isolated mode is enabled and the ' ..
+            'instance %q has no local snapshot. An attempt to bootstrap ' ..
+            'the database would produce data that has nothing in common ' ..
+            'with data of other replicaset peers: this instance wouldn\'t ' ..
+            'able to join the replicaset later.'):format(instance_name), 0)
```
grafin, line 795: This message seems a bit to long. Maybe something like this will be better:
```
'Startup failure.\nThe isolated mode is enabled and the ' ..
'instance %q has no local snapshot. An attempt to bootstrap ' ..
'the instance would lead to split-brain.'
```

grafin (approved):
Thanks for the patch! LGTM, except for one small nit, feel free to ignore.

src/box/lua/config/configdata.lua
Totktonada: Hm. We usually say 'split-brain', when there is some transaction history in common and then it forks. Here we likely have ever different replicaset UUIDs (if it is not set explicitly in the configuration). And conflicting instance IDs. Wouldn't saying just 'split-brain' here be confusing?

I see nothing bad in a long message if it adds useful information and/or directions what to do. I would ever say that I'm tired of error messages that [tells nothing](https://stackoverflow.com/questions/36216437/tarantool-long-wal-write/) to a user.

Totktonada: Hoewver, I'm for the shorter message if it fed the same information. If you read my doubts and still prefer to just say 'split-brain', I would do it.

Totktonada: Discussed it a bit with @grafin and changed:

```diff
diff --git a/src/box/lua/config/configdata.lua b/src/box/lua/config/configdata.lua
index 587005c23c..7633144f4a 100644
--- a/src/box/lua/config/configdata.lua
+++ b/src/box/lua/config/configdata.lua
@@ -790,9 +790,8 @@ local function validate_startup(instance_name, iconfig_def)
     if no_snap and isolated then
         error(('Startup failure.\nThe isolated mode is enabled and the ' ..
             'instance %q has no local snapshot. An attempt to bootstrap ' ..
-            'the database would produce data that has nothing in common ' ..
-            'with data of other replicaset peers: this instance wouldn\'t ' ..
-            'able to join the replicaset later.'):format(instance_name), 0)
+            'the instance would lead to the split-brain situation.'):format(
+            instance_name), 0)
     end
 
     -- TODO: There is a situation, which looks similar, but we
diff --git a/test/config-luatest/isolated_mode_test.lua b/test/config-luatest/isolated_mode_test.lua
index 46f4a88bc7..d3a7cf6283 100644
--- a/test/config-luatest/isolated_mode_test.lua
+++ b/test/config-luatest/isolated_mode_test.lua
@@ -44,9 +44,8 @@ g.test_startup_no_snap = function()
     -- Verify the exit code and the error reported to stderr.
     local exp_err = 'Startup failure.\n' ..
         'The isolated mode is enabled and the instance "i-003" has no local ' ..
-        'snapshot. An attempt to bootstrap the database would produce data ' ..
-        'that has nothing in common with data of other replicaset peers: ' ..
-        'this instance wouldn\'t able to join the replicaset later.'
+        'snapshot. An attempt to bootstrap the instance would lead to the ' ..
+        'split-brain situation.'
     t.assert_covers(res, {
         exit_code = 1,
         stderr = ('LuajitError: %s\nfatal error, exiting the event loop')
```

Also, rebased the patch on top of the current master.

PR #10811, opened 2024-11-12
config: switch to RO in isolated mode

Totktonada:
The isolated instance shouldn't produce new transactions, including ones from a background fiber started by an application or a role. So, the isolated instance forcefully goes to the read-only mode disregarding any other configuration options.

Part of #10796

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -61,6 +61,39 @@ local function peer_uris(configdata)
     return uris
 end
 
+-- Modify box-level configuration values and perform other actions
+-- to enable the isolated mode (if configured).
+local function switch_isolated_mode(configdata, box_cfg)
+    -- If the isolated mode is not enabled, there is nothing to do.
+    if not configdata:get('isolated', {use_default = true}) then
+        return
+    end
+
+    -- An application or a role may perform background database
+    -- modification if the instance is in the RW mode: for
+    -- example, a role may perform eviction of stale records.
+    -- If the instance is in the isolated mode, it should be in RO
+    -- to don't produce any new transactions.
+    --
+    -- The reason is that these transactions will be sent to other
+    -- replicaset members and applied on them, when the instance
+    -- goes from the isolated mode. At the same time, the
+    -- non-isolated part of the replicaset may serve requests and
+    -- perform data modifications. An attempt to modify the same
+    -- data from two instances may break data integrity[^1].
+    --
+    -- It is recommended to extract all the needed data from the
+    -- isolated instance and perform the modifications on the
+    -- current leader (in the non-isolated part of the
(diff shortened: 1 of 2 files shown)


grafin (approved):
LTGM.

src/box/lua/config/applier/box_cfg.lua

georgiy-belyanin, line 1: Feels like the documentation on [https://github.com/tarantool/tarantool/blob/56e9c1bd2ae1cbc0b22cd8777f17d179592cd5cc/src/box/lua/config/applier/box_cfg.lua#L684-L688](https://github.com/tarantool/tarantool/blob/56e9c1bd2ae1cbc0b22cd8777f17d179592cd5cc/src/box/lua/config/applier/box_cfg.lua#L684-L688) now lies. I suppose it should include the information RO is enforced if the instance is isolated.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -754,6 +787,10 @@ local function apply(config)
             box_cfg.read_only = true
         end
 
+        -- RO may be enforced by the isolated mode, so we call the
+        -- function after all the other RO/RW logic is done.
+        switch_isolated_mode(configdata, box_cfg)
+
```
georgiy-belyanin, line 793: I don't like the same `switch_isolated_mode()` in two parts though it seems safer than any other approach. Also it brings kinda:
```
force_ro_if_some_condition()
forcier_ro_if_some_condition()
```

If resulting `needs_retry` can be `false` if the instance is isolated seems like there is no difference between startup and non startup scenarios and switch to ro can be done before all of these checks.
```
if isolated then <enforce ro> return end

if is_startup and <other conds> then 
  <old startup code>
end

<otherwise code>
```

Have you considered this idea?

georgiy-belyanin (approved):
Thank you for the patch! In general, I am OK but I have a few concerns. Please, take a look at them.

src/box/lua/config/applier/box_cfg.lua
Totktonada: In my vision the safe startup logic and the isolated mode logic are orthogonal. However, it doesn't block me from adding a remark:

```lua
-- NB: The read-only mode may be enforced due to other reasons
-- (such as enabled isolated mode) that are not specific to
-- the startup flow.
```

georgiy-belyanin: Thank you! I suppose you're right the logic is not so related, though it seems more consistent because the previous docs don't even cover that case at all.

The new approach is better.

src/box/lua/config/applier/box_cfg.lua
georgiy-belyanin: Thank you for refactoring and for a detailed description of the suggested approach.

PR #10813, opened 2024-11-13
config: drop iproto connections in isolated mode

Totktonada:
An isolated instance shouldn't process user traffic. This commit adds a logic to the isolated mode that stop listening for new connections and drop existing connections in iproto.

Part of #10796

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -5,6 +5,7 @@ local snapshot = require('internal.config.utils.snapshot')
 local schedule_task = fiber._internal.schedule_task
 local mkversion = require('internal.mkversion')
 local tarantool = require('tarantool')
+local clock = require('clock')
 
 local function peer_uris(configdata)
     local peers = configdata:peers()
@@ -63,7 +64,7 @@ end
 
 -- Modify box-level configuration values and perform other actions
 -- to enable the isolated mode (if configured).
-local function switch_isolated_mode(configdata, box_cfg)
+local function switch_isolated_mode_before_box_cfg(configdata, box_cfg)
     -- If the isolated mode is not enabled, there is nothing to do.
     if not configdata:get('isolated', {use_default = true}) then
         return
@@ -92,6 +93,123 @@ local function switch_isolated_mode(configdata, box_cfg)
     --
     -- TODO(gh-10404): Set ro_reason=isolated.
     box_cfg.read_only = true
+
+    -- Don't accept new iproto connections.
+    --
+    -- An instance in the isolated mode shouldn't process a
+    -- traffic from a user.
+    box_cfg.listen = box.NULL
+end
+
+-- Perform post-box-cfg actions to enable the isolated mode (if
+-- configured).
+local function switch_isolated_mode_after_box_cfg(config)
+    local configdata = config._configdata
+
+    -- If the isolated mode is not enabled, there is nothing to do.
+    if not configdata:get('isolated', {use_default = true}) then
+        return
+    end
+
+    -- Drop existing iproto connections.
+    --
+    -- An instance in the isolated mode shouldn't process a
+    -- traffic from a user.
+    --
+    -- Here we assume that when box.cfg() is finished, iproto
+    -- listening sockets are already closed and no new connections
+    -- may be created. So, the only thing to do is to close the
+    -- existing connections.
+    --
+    -- NB: It seems safe to start this fiber several times in
+    -- parallel, so there are no checks, whether such a fiber
+    -- already exists.
+    fiber.new(function()
+        local name = 'isolated_mode_drop_iproto_connections'
+        fiber.self():name(name, {truncate = true})
+
+        -- A note regarding box.iproto.internal.drop_connections.
+        --
+        -- It seems, the current behavior if connections are not
+        -- closed in time is to report the timeout error and
+        -- continue closing connection from the given generation
+        -- in background.
+        --
+        -- It means that the passed timeout doesn't affect the
+        -- behavior of the process of closing connections itself
+        -- and can be only used to let the caller know whether
+        -- and when the connections (existed at the moment of the
+        -- invocation) are closed.
+        --
+        -- Taking it into account, the timeout in this code is
+        -- not configurable and chosen more or less arbitrarily.
+        -- It only affects whether and when an alert is issued.
+        --
+        -- If we find a need in a future, we can make it
+        -- configurable.
+        local timeout = 10
+
+        -- Testing requires to redefine the timeout. Use an
+        -- environment variable for that.
+        local env_var_name = 'TT_CONFIG_DROP_CONNECTION_TIMEOUT'
+        local redefined_timeout = tonumber(os.getenv(env_var_name))
+        if redefined_timeout ~= nil then
+            log.warn('isolated mode: %s redefines the iproto connections ' ..
+                'drop timeout: %d', env_var_name, redefined_timeout)
+            timeout = redefined_timeout
+        end
+
+        local start_time = clock.monotonic()
+        local ok, err = pcall(box.iproto.internal.drop_connections, timeout)
+
+        -- An unique key is set here to don't duplicate the alert
+        -- if the box_cfg.apply function is called twice within
+        -- the same startup/reconfiguration process.
+        --
+        -- Currently, it is possible on startup due to the safe
+        -- startup process. See c80b121544f6 ("config: add safe
+        -- startup mode") for details.
+        --
+        -- Also, this key is used to drop the alert (if any) after
+        -- a successful `drop_connections` call.
+        local alert_key = 'isolated_mode_drop_iproto_connections_timeout'
+
+        if not ok then
+           local message = ('isolated mode: can\'t drop iproto connections ' ..
+                'during %d seconds (continued in background): %s'):format(
+                timeout, err)
+
+            -- There is no clear point, when to drop this alert.
+            --
+            -- Once appeared, it exists till the next
+            -- reconfiguration, because all the alerts are dropped
+            -- before applying the new configuration.
+            --
+            -- Also, dropped on a successful attempt to drop
+            -- connections.
+            config._aboard:set({
+                type = 'warn',
+                message = message,
+            }, {
+                key = alert_key,
+            })
+            return
+        end
+
+        -- If it appears that there were several reconfigurations
+        -- in row, each schedules a fiber to drop the connections,
+        -- but the fibers are executed after the reconfigurations,
+        -- and the first issued an alert, but the next
+        -- successfully dropped the connections, let's drop this
+        -- alert.
+        --
+        -- No-op if there is no alert with the given key.
+        config._aboard:drop(alert_key)
+
+        local duration = clock.monotonic() - start_time
+        log.info('isolated mode: successfully dropped iproto connections ' ..
+            '(took %d seconds)', duration)
+    end)
 end
 
 local function log_destination(log)
@@ -359,8 +477,52 @@ local function get_schema_version_before_cfg(config)
     return nil
 end
 
+-- See hooks_new().
+local hooks_mt = {
+    __index = {
+        add = function(self, hook, ctx)
+            local entry = {
+                hook = hook,
+                ctx = ctx,
+            }
+            table.insert(self._hooks, entry)
+        end,
+        run = function(self)
+            for _, entry in ipairs(self._hooks) do
+                entry.hook(entry.ctx)
+            end
+        end,
+    },
+}
+
+-- A little helper to collect some hooks and run them later.
+--
+-- Usage:
+--
+--  | local hooks = hooks_new()
+--  |
+--  | eat_pizza()
+--  | hooks:add(throw_away, {what = 'pizza box'})
+--  | hooks:add(wash_hands, {person = 'me'})
+--  |
+--  | drink_juice()
+--  | hooks:add(wash_glass, {glass = 'my glass'})
+--  |
+--  | finish_dinner()
+--  | hooks:run()
+local function hooks_new()
+    return setmetatable({
+        _hooks = {},
+    }, hooks_mt)
+end
+
 -- Returns nothing or {needs_retry = true}.
 local function apply(config)
+    -- There are some actions that logically correspond to some
+    -- logic written before the box.cfg() call, but technically
+    -- they should be run after box.cfg(). Place them here.
+    local post_box_cfg_hooks = hooks_new()
+
     local configdata = config._configdata
     local box_cfg = configdata:filter(function(w)
         return w.schema.box_cfg ~= nil
@@ -679,7 +841,8 @@ local function apply(config)
 
     -- RO may be enforced by the isolated mode, so we call the
     -- function after all the other logic that may set RW.
-    switch_isolated_mode(configdata, box_cfg)
+    switch_isolated_mode_before_box_cfg(configdata, box_cfg)
+    post_box_cfg_hooks:add(switch_isolated_mode_after_box_cfg, config)
 
     -- First box.cfg() call.
     --
@@ -797,6 +960,7 @@ local function apply(config)
 
         log.debug('box_cfg.apply (startup)')
         box.cfg(box_cfg)
+        post_box_cfg_hooks:run()
 
         -- NB: needs_retry should be true when force_read_only is
         -- true. It is so by construction.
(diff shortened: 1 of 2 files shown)


src/box/lua/config/applier/box_cfg.lua
```diff
@@ -92,6 +93,108 @@ local function switch_isolated_mode(configdata, box_cfg)
...
+        -- Testing requires to redefine the timeout. Use an error
+        -- injection for that.
+        local errinj_name = 'ERRINJ_CONFIG_DROP_CONNECTION_TIMEOUT'
+        local errinj_timeout = box.error.injection.get(errinj_name)
+        assert(type(errinj_timeout) == 'number')
+        if errinj_timeout >= 0 then
+            log.warn('isolated mode: %s redefines the iproto connections ' ..
+                'drop timeout: %d', errinj_name, errinj_timeout)
+            timeout = errinj_timeout
+        end
```
grafin, line 161: just a proposal: maybe we can add some "`internal`" option for this timeout, for me it just doesn't seem right to change the timeout via error injection.

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -92,6 +93,108 @@ local function switch_isolated_mode(configdata, box_cfg)
...
+        local ok, err = pcall(box.iproto.internal.drop_connections, timeout)
+        if not ok then
+            -- An unique key is set here to don't duplicate the
+            -- alert if the box_cfg.apply function is called twice
+            -- within the same startup/reconfiguration process.
+            --
+            -- Currently, it is possible on startup due to the
+            -- safe startup process. See c80b121544f6 ("config:
+            -- add safe startup mode") for details.
+            local key = 'isolated_mode_drop_iproto_connections_timeout'
```
grafin, line 173: It is stated that this fiber could be started several times, and if several of them fail we will loose some of the errors, may be we could concatenate the error messages, or generate a unique key for each fiber?

src/box/lua/config/applier/box_cfg.lua
Totktonada: I've considered to just add an environment variable, but then observed that there are error injections related to timeouts/delays:

```c
$ grep 'ERRINJ.*\(TIMEOUT\|DELAY\).*dparam' src/lib/core/errinj.h
	_(ERRINJ_BUILD_INDEX_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_FLIGHTREC_LOG_DELAY, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_RELAY_EXIT_DELAY, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_RELAY_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_SNAP_WRITE_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_TX_DELAY_PRIO_ENDPOINT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_VY_READ_PAGE_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_VY_RUN_WRITE_STMT_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_VY_SCHED_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
	_(ERRINJ_VY_SQUASH_TIMEOUT, ERRINJ_DOUBLE, {.dparam = 0}) \
```

In fact, anything that works (and is not in the public API) is OK for me.

src/box/lua/config/applier/box_cfg.lua
Totktonada: If the alerts have the same key, the last one replaces the previous one. So, we have one alert here at most. It seems meaningful, because they have the same message, wthout any additional (possibly different) details (such as a list of problematic connections).


However, it seems that I can discard the alert after a successful call. I'll do.

src/box/lua/config/applier/box_cfg.lua
grafin: I've didn't check all of them but the use seems different:
```c
	ERROR_INJECT_DOUBLE(ERRINJ_SNAP_WRITE_TIMEOUT, inj->dparam > 0,
			    thread_sleep(inj->dparam));
```

Making them more like an inner delay for some actions, rather then a timeout for some inner call. ENV variable maybe ok to me, but again it doesn't seem SO internal as an internal option.)

Totktonada: You're right in regards to the current error injections usage.

However, some hidden option in the YAML configuration seems problematic: we need to explicitly take it into account in various places. I wouldn't do it.

Also, I don't like to expose functions just for testing: it feels inaccurate (and also requires to be able to call the function that is not always possible -- for example, before first `box.cfg()`).

Next, if the `tweaks` mechanism can be used purely from Lua (need to check) and can be used using an env variable (to set the value on startup), I would use it.

Also, I can attempt to use `debug.getupvalue()` to access the internal variable, like I do in the `odict` testing.

https://github.com/tarantool/tarantool/blob/0656a9231149663a0f13c4be7466d4776ccb0e66/src/box/lua/config/utils/odict.lua#L58

https://github.com/tarantool/tarantool/blob/0656a9231149663a0f13c4be7466d4776ccb0e66/test/config-luatest/odict_test.lua#L226-L235

But it is problematic to access on startup. I can do it using an override module, though. Like here:

https://github.com/tarantool/tarantool/blob/0656a9231149663a0f13c4be7466d4776ccb0e66/test/app-luatest/gh_8613_centralized_config_options_test.lua#L99-L130

It would be unnecessarily tricky, IMHO.

As the last resort I can implement something like tweaks/errinj on the config module level. However, I would prefer to use/extend some tarantool mechanism.

I'll look here around.

grafin: I didn't think it would be so difficult. If there is no "better" easy solution, lets just leave the error injection.

Totktonada: A tweak needs a definition in C code. No problems with it technically, but I don't want to spread relevant code over the codebase.

There is an interesting idea to reuse `tweaks.box_shutdown_timeout`: it is 3 seconds by default and it is used to call the same `iproto_drop_connections` function on tarantool's shutdown.

However, this value is used to hang tarantool in CI in case of a problem in shutdown.

https://github.com/tarantool/tarantool/blob/0656a9231149663a0f13c4be7466d4776ccb0e66/src/box/box.cc#L263-L273

Again, it doesn't block me technically, but it is not what I expect from some looking-as-a-hardcoded-constant value.

Also, I need some code (an override mode, -e option, TT_PRELOAD code) to set a tweak on startup.

So, I'm going to just use an env variable with a long enough name.

In fact, I already added one undocumented env variable before: `TT_CONSOLE_HIDE_SHOW_PROMPT`. The purpose was different, though. I wanted to have a chance to leave a suggestion to a user how to disable the new functionality if someting goes wrong with it.

So, I think an env undocumented variable for testing purposes is OK too.

src/box/lua/config/applier/box_cfg.lua
Totktonada: Am I answered your question? Is the answer OK for you?

----

> However, it seems that I can discard the alert after a successful call. I'll do.

I've implemented it. I didn't add a test case, because it seems complicated to perform so fast reconfigurations to get such fibers actually executed in parallel and, moreover, one should go over the timeout, but another should succeed.

Anyway, `aboard:drop(key)` is just no-op if there is no alert. I think it is completely OK to just drop the alert if the action/condition related to it passed successfully.

```diff
diff --git a/src/box/lua/config/applier/box_cfg.lua b/src/box/lua/config/applier/box_cfg.lua
index 711d541295..ecf2724c4c 100644
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -161,35 +161,51 @@ local function switch_isolated_mode_after_box_cfg(config)
 
         local start_time = clock.monotonic()
         local ok, err = pcall(box.iproto.internal.drop_connections, timeout)
-        if not ok then
-            -- An unique key is set here to don't duplicate the
-            -- alert if the box_cfg.apply function is called twice
-            -- within the same startup/reconfiguration process.
-            --
-            -- Currently, it is possible on startup due to the
-            -- safe startup process. See c80b121544f6 ("config:
-            -- add safe startup mode") for details.
-            local key = 'isolated_mode_drop_iproto_connections_timeout'
 
-            local message = ('isolated mode: can\'t drop iproto connections ' ..
+        -- An unique key is set here to don't duplicate the alert
+        -- if the box_cfg.apply function is called twice within
+        -- the same startup/reconfiguration process.
+        --
+        -- Currently, it is possible on startup due to the safe
+        -- startup process. See c80b121544f6 ("config: add safe
+        -- startup mode") for details.
+        --
+        -- Also, this key is used to drop the alert (if any) after
+        -- a successful `drop_connections` call.
+        local alert_key = 'isolated_mode_drop_iproto_connections_timeout'
+
+        if not ok then
+           local message = ('isolated mode: can\'t drop iproto connections ' ..
                 'during %d seconds (continued in background): %s'):format(
                 timeout, err)
 
             -- There is no clear point, when to drop this alert.
-            -- So, don't bother about it much.
             --
             -- Once appeared, it exists till the next
             -- reconfiguration, because all the alerts are dropped
             -- before applying the new configuration.
+            --
+            -- Also, dropped on a successful attempt to drop
+            -- connections.
             config._aboard:set({
                 type = 'warn',
                 message = message,
             }, {
-                key = key,
+                key = alert_key,
             })
             return
         end
 
+        -- If it appears that there were several reconfigurations
+        -- in row, each schedules a fiber to drop the connections,
+        -- but the fibers are executed after the reconfigurations,
+        -- and the first issues an alert, but the next
+        -- successfully dropped the connections, let's drop this
+        -- alert.
+        --
+        -- No-op if there is no alert with the given key.
+        config._aboard:drop(alert_key)
+
         local duration = clock.monotonic() - start_time
         log.info('isolated mode: successfully dropped iproto connections ' ..
             '(took %d seconds)', duration)
```

src/box/lua/config/applier/box_cfg.lua
Totktonada: Updated with the `TT_CONFIG_DROP_CONNECTION_TIMEOUT` environment variable.

```diff
diff --git a/src/box/lua/config/applier/box_cfg.lua b/src/box/lua/config/applier/box_cfg.lua
index 172fa726e9..711d541295 100644
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -149,15 +149,14 @@ local function switch_isolated_mode_after_box_cfg(config)
         -- configurable.
         local timeout = 10
 
-        -- Testing requires to redefine the timeout. Use an error
-        -- injection for that.
-        local errinj_name = 'ERRINJ_CONFIG_DROP_CONNECTION_TIMEOUT'
-        local errinj_timeout = box.error.injection.get(errinj_name)
-        assert(type(errinj_timeout) == 'number')
-        if errinj_timeout >= 0 then
+        -- Testing requires to redefine the timeout. Use an
+        -- environment variable for that.
+        local env_var_name = 'TT_CONFIG_DROP_CONNECTION_TIMEOUT'
+        local redefined_timeout = tonumber(os.getenv(env_var_name))
+        if redefined_timeout ~= nil then
             log.warn('isolated mode: %s redefines the iproto connections ' ..
-                'drop timeout: %d', errinj_name, errinj_timeout)
-            timeout = errinj_timeout
+                'drop timeout: %d', env_var_name, redefined_timeout)
+            timeout = redefined_timeout
         end
 
         local start_time = clock.monotonic()
diff --git a/src/lib/core/errinj.h b/src/lib/core/errinj.h
index 8dd4722db6..49c2414b50 100644
--- a/src/lib/core/errinj.h
+++ b/src/lib/core/errinj.h
@@ -87,7 +87,6 @@ struct errinj {
 	_(ERRINJ_CHECK_FORMAT_DELAY_COUNTDOWN, ERRINJ_INT, {.iparam = -1}) \
 	_(ERRINJ_COIO_SENDFILE_CHUNK, ERRINJ_INT, {.iparam = -1}) \
 	_(ERRINJ_COIO_WRITE_CHUNK, ERRINJ_BOOL, {.bparam = false}) \
-	_(ERRINJ_CONFIG_DROP_CONNECTION_TIMEOUT, ERRINJ_DOUBLE, {.dparam = -1}) \
 	_(ERRINJ_DYN_MODULE_COUNT, ERRINJ_INT, {.iparam = 0}) \
 	_(ERRINJ_ENGINE_JOIN_DELAY, ERRINJ_BOOL, {.bparam = false}) \
 	_(ERRINJ_FIBER_MADVISE, ERRINJ_BOOL, {.bparam = false}) \
diff --git a/test/config-luatest/isolated_mode_test.lua b/test/config-luatest/isolated_mode_test.lua
index acbe1158f4..68669ba9cf 100644
--- a/test/config-luatest/isolated_mode_test.lua
+++ b/test/config-luatest/isolated_mode_test.lua
@@ -253,12 +253,8 @@ g.test_iproto_stop_failure = function(g)
     cluster:start()
 
     -- Imitate a timeout error on dropping iproto connections.
-    --
-    -- NB: Error injections work as in Debug as well as in
-    -- RelWithDebInfo build, when used from the Lua API.
     cluster['i-001']:exec(function()
-        local errinj_name = 'ERRINJ_CONFIG_DROP_CONNECTION_TIMEOUT'
-        box.error.injection.set(errinj_name, 0)
+        os.setenv('TT_CONFIG_DROP_CONNECTION_TIMEOUT', '0')
     end)
 
     -- Connect to the server's console and verify that it works.
@@ -319,9 +315,6 @@ end
 -- The test case mostly to verify that the alert is not
 -- duplicated.
 g.test_iproto_stop_failure_on_startup = function(g)
-    t.tarantool.skip_if_not_debug(
-        'errinj based test cases only work in the Debug build')
-
     local config = cbuilder:new()
         :add_instance('i-001', {})
         :config()
@@ -330,10 +323,7 @@ g.test_iproto_stop_failure_on_startup = function(g)
         env = {
             -- Imitate a timeout error on dropping iproto
             -- connections.
-            --
-            -- Setting an error injection using an environment
-            -- variable works only in the Debug build.
-            ['ERRINJ_CONFIG_DROP_CONNECTION_TIMEOUT'] = 0,
+            ['TT_CONFIG_DROP_CONNECTION_TIMEOUT'] = '0',
         },
     })
     cluster:start()
```

src/box/lua/config/applier/box_cfg.lua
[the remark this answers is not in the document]
Totktonada: Done.

src/box/lua/config/applier/box_cfg.lua
grafin: Yeah, thanks for the explanation!

src/box/lua/config/applier/box_cfg.lua
grafin: Thanks for looking into this!

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -789,23 +907,27 @@ local function apply(config)
 
         -- RO may be enforced by the isolated mode, so we call the
         -- function after all the other RO/RW logic is done.
-        switch_isolated_mode(configdata, box_cfg)
+        switch_isolated_mode_before_box_cfg(configdata, box_cfg)
 
         log.debug('box_cfg.apply (startup)')
         box.cfg(box_cfg)
 
+        switch_isolated_mode_after_box_cfg(config)
```
Totktonada, line 915: Needs some update after [changes](https://github.com/tarantool/tarantool/pull/10811#discussion_r1850563921) in the previous patch.

Totktonada: Updated.

georgiy-belyanin (approved):
Thank you for working on it! LGTM.

Are you going to document `TT_CONFIG_DROP_CONNECTION_TIMEOUT` option? Feels like it's not really useful.

Totktonada:
@georgiy-belyanin Thanks for the review!

> Are you going to document `TT_CONFIG_DROP_CONNECTION_TIMEOUT` option? Feels like it's not really useful.

Nope. Just like `TT_CONSOLE_HIDE_SHOW_PROMPT` it is for testing and for support (in case of very unexpected problems).

If we need to make it configurable, I would better add a configuration option (if we will see a need).

PR #10815, opened 2024-11-14
config: add failover.stateboard.enabled option

georgiy-belyanin:
Now `failover.stateboard.enabled` boolean option can be used to enable/disable the failover coordinator stateboard. Default value equals to `true`.

`failover.stateboard.enabled: true` requires for one of the compatible remote config sources to be set in the `config` section. Otherwise the failover coordinator doesn't start.

The option is enterprise-only since it's a part of the supervised failover.

Needed for EE tarantool/tarantool-ee#996.

### DIFF
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2300,6 +2300,22 @@ return schema.new('instance_config', schema.record({
         -- are configured in the `config.etcd` or the
         -- `config.storage` section.
         stateboard = schema.record({
+            -- Stateboard enabled/disabled status.
+            -- A coordinator having stateboard enabled doesn't
+            -- start without specifying compatible remote config
+            -- storage in the config.
+            --
+            -- Beware: disabling stateboard leads to the
+            -- following consequences:
+            -- * Failover commands aren't supported.
+            -- * Failover state monitoring isn't provided.
+            -- * Multiple coordinators over the same replicaset
+            --   behave in an unpredictable way. Two RW instances
+            --   are possible.
+            enabled = schema.scalar({
+                type = 'boolean',
+                default = true,
+            }),
             -- How often information in the stateboard is updated.
             renew_interval = schema.scalar({
                 type = 'number',
--- a/test/config-luatest/cluster_config_schema_test.lua
+++ b/test/config-luatest/cluster_config_schema_test.lua
@@ -326,6 +326,7 @@ g.test_defaults = function()
             lease_interval = 30,
             renew_interval = 10,
             stateboard = {
+                enabled = true,
                 renew_interval = 2,
                 keepalive_interval = 10,
             },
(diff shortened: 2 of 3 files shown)


src/box/lua/config/instance_config.lua
Totktonada, line 1: In my view, the key point is that now the coordinator can't start without the stateboard unless it is requested explicitly. However, it is not requested to be included into the documentation for the option.

(And also I hardly understand the idea behind two documentation requests for the same feature -- another one is in the EE commit -- so, I don't know, where I should comment it.)

src/box/lua/config/instance_config.lua
```diff
@@ -2300,6 +2300,19 @@ return schema.new('instance_config', schema.record({
...
         stateboard = schema.record({
+            -- Stateboard enabled/disabled status.
+            --
+            -- Beware: disabling stateboard leads to the
+            -- following consequences:
+            -- * Failover commands aren't supported.
+            -- * Failover state monitoring isn't provided.
+            -- * Multiple coordinators over the same replicaset
+            --   behave in an unpredictable way. Two RW instances
+            --   are possible.
```
Totktonada, line 2311: Same here: for me the key point behind this option is that now the coordinator doesn't start by default without a configured remote config/state storage and that one who needs it should use the option. We tell about consequences of disabling, but I think the main idea should be explained first.

Totktonada (approved):
OK for me ~except some remarks how to express the idea better in comments/docs (fixed now)~.

src/box/lua/config/instance_config.lua
georgiy-belyanin: The idea behind two distinct PRs is that one of them is supposed to update the configuration reference meanwhile the other one is for the supervised failover coordinator.

I am linking them after merging.

georgiy-belyanin: I've made the doc request include the information that `failover.stateboard.enabled: true` requires the remote config source to be set.

Thank you. That's a nice point.

src/box/lua/config/instance_config.lua
georgiy-belyanin: Dito. I've changed the comment here. Now it's started from the main idea that by default one of the remote configuration sources should be set.

grafin (approved):
I see no problems here. Thanks for the patch!

grafin (changes requested):
I've just noticed that in the commit message has a typo AFAIU https://github.com/tarantool/tarantool-ee/pull/966 is mentioned twice, looks like https://github.com/tarantool/tarantool-ee/pull/996 is intended

georgiy-belyanin:
> I've just noticed that in the commit message has a typo AFAIU [tarantool/tarantool-ee#966](https://github.com/tarantool/tarantool-ee/pull/966) is mentioned twice, looks like [tarantool/tarantool-ee#996](https://github.com/tarantool/tarantool-ee/issues/996) is intended

Thank you for noticing! Fixed.

PR #10816, opened 2024-11-15
config: don't replicate data from/to isolated instance

Totktonada:
The isolated mode can be enabled to stop data modifications on the given instance, including ones from the replication: it may help to debug a problem or extract some needed data. So, it seems worthful to stop replication from the non-isolated part of the replicaset to the instance.

The isolated instance goes to RO and drops iproto connections -- it should effectively stop replication from it *to* the non-isolated part of the replicaset. There is no sense to retry an attempt to connect to it and flood logs with errors. Let's just remove the isolated instance from the upstreams list.

Part of #10796

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -99,6 +99,13 @@ local function switch_isolated_mode_before_box_cfg(configdata, box_cfg)
     -- An instance in the isolated mode shouldn't process a
     -- traffic from a user.
     box_cfg.listen = box.NULL
+
+    -- Don't replicate data from other replicaset members.
+    --
+    -- The isolated mode can be enabled to stop data modifications
+    -- on the given instance, including ones from the replication.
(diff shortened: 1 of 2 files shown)


src/box/lua/config/applier/box_cfg.lua

grafin (approved):
Thanks for the patch! There is a small typo in one of the commit messages 'connections -- it' - looks like there is an extraneous - here. Other than this - everything looks good to me.

Totktonada:
@grafin Thanks for the review!

Two hyphens are here deliberately -- this is how I represent a dash in texts, where I don't want to go beyond ASCII. However, I'm not sure that it fits good with the English grammar in the first place. I'll look around here, when rebase the patchset next time.

src/box/lua/config/applier/box_cfg.lua
Totktonada: Fixed.

georgiy-belyanin (approved):
Thank you for the patch. Everything is clear and documented rather well.

PR #10821, opened 2024-11-15
config: issue an alert on isolated instance

### LINKED ISSUE #10776: config: temporary exclude an instance from the replicaset

### LINKED ISSUE #10796: config: add isolated mode for an instance

Totktonada:
This way it is harder to overlook that the instance is isolated. It islikely important information for a person who performs administration tasks.

Closes #10796

### DIFF
--- a/src/box/lua/config/applier/box_cfg.lua
+++ b/src/box/lua/config/applier/box_cfg.lua
@@ -71,11 +71,25 @@ end
 
 -- Modify box-level configuration values and perform other actions
 -- to enable the isolated mode (if configured).
-local function switch_isolated_mode_before_box_cfg(configdata, box_cfg)
+local function switch_isolated_mode_before_box_cfg(config, box_cfg)
+    local configdata = config._configdata
+
     -- If the isolated mode is not enabled, there is nothing to do.
     if not configdata:get('isolated', {use_default = true}) then
         return
     end
+
+    -- Issue a warning to highlight the unusual instance status to
+    -- the administrator.
+    local key = 'isolated_mode_enabled'
+    local message = ('The isolated mode is set for the instance %q'):format(
+        config._instance_name)
+    config._aboard:set({
+        type = 'warn',
+        message = message,
+    }, {
+        key = key,
+    })
 
     -- An application or a role may perform background database
     -- modification if the instance is in the RW mode: for
@@ -855,7 +869,7 @@ local function apply(config)
 
     -- RO may be enforced by the isolated mode, so we call the
     -- function after all the other logic that may set RW.
-    switch_isolated_mode_before_box_cfg(configdata, box_cfg)
+    switch_isolated_mode_before_box_cfg(config, box_cfg)
     post_box_cfg_hooks:add(switch_isolated_mode_after_box_cfg, config)
 
     -- First box.cfg() call.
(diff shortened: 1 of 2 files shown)


src/box/lua/config/applier/box_cfg.lua

grafin (approved):
LGTM. Thanks a lot for breaking this patchset into several PRs!

src/box/lua/config/applier/box_cfg.lua
Totktonada: Done.

georgiy-belyanin (approved):
Thank you for the work you've done for easy patch review. Once again, everything is quite clear and IMO ok.

georgiy-belyanin:
I am not really into Tarantool administration process: probably it's also worth creating an alert on non-isolated instances that there is an isolated instance in the replicaset and that's why the data isn't replicated?

Totktonada:
> I am not really into Tarantool administration process: probably it's also worth creating an alert on non-isolated instances that there is an isolated instance in the replicaset and that's why the data isn't replicated?

I'm here on the more-or-less same side: probably :)

This idea comes to me too and I was in doubt, whether it is too many warnings (all the instances of the replicaset will be marked as yellow in TCM, for example). I've no decision here, so I didn't do it.

Issue #9261, opened 2023-10-13
Timeouts slip if there are no yields for a long time


nshy:
Suppose current fiber does not yield for a time `T1`. If it arms a timer using `ev_timer` to trigger in time `T2` then timer will trigger in 
`T2 - T1` time instead of `T2` in the best case (fiber yields right after arming the timer and there are no other fibers). The issue is ev updates its current time (`mn_now`) only when when it grabs control back. So in the above situation at the time of arming the current time is `T1` behind "real time".

There is `ev_now_update` to update `ev` current time at any moment.

Totktonada:
Looks like a reincarnation of #2737.

PR #10826, opened 2024-11-18
txn: propagate transaction timeout to `box.commit`

### LINKED ISSUE #10181: Support transaction timeout for synchronous transactions
After deprecation of the `IPROTO_RAFT_ROLLBACK` and `replication_synchro_timeout`, introduced in https://github.com/tarantool/tarantool/pull/9869, entries in limbo will live indefinitely and won't be automatically rolled back on timeout. 

We need to support `box.begin({timeout = <some-value>}` for synchronous transactions. The expected behavior: after some timeout passes, `box.commit` returns error, transaction is not rolled back, it remains in limbo

Astronomax:
New `compat.box_begin_timeout_meaning` option controls the behavior of the timeout `box.begin{timeout=`.

When the option is set to 'old' the behavior of `box.begin{timeout=` is no different from the behavior that was before this patch. This timeout can only abort the transaction until `box_commit()` is called. If the transaction is synchronous and `compat.replication_synchro_timeout='old'`, then immediately after the transaction is written to WAL, `replication_synchro_timeout` will start working. If the transaction didn't have time to gather a quorum within the timeout, it is rolled back.

~~The new behavior can only be enabled if `compat.replication_synchro_timeout` is set to 'new', that is, when `replication_synchro_timeout` is not running.~~ With the new behavior, `replication_synchro_timeout` does not work
regardless of what value is set in `compat.box_cfg_replication_synchro_timeout`.
In the new behavior, the timeout does not stop working after `box_commit()` is called, it covers the entire process of committing the transaction:

 * If the timeout expires before `box.commit()` is called **or while `before_commit` triggers are executing**, the transaction is aborted. The user receives an error "Transaction has been aborted by timeout".
 * If the timeout expires later, the user receives an error "Transaction timeout has expired during commit. It continues to commit." with field `.expired_at` with following possible values: "WAL write", "quorum collection", "CONFIRM write" depending on at what point of fiber clocks the timeout expired. As can be seen from the error text, the transaction is not rolled back, it continues to be committed, but the user's fiber is ready to execute something else.

Closes #10181

### DIFF
--- a/changelogs/unreleased/gh-10181-propagate-transaction-timeout-to-box-commit.md
+++ b/changelogs/unreleased/gh-10181-propagate-transaction-timeout-to-box-commit.md
@@ -1,0 +1,7 @@
+## feature/box
+
+* A new compat option `compat.box_begin_timeout_meaning` has been added.
+  Controls the behavior of the timeout `box.begin{timeout=`. If set to 'new',
+  the transaction timeout is propagated further to `box.commit`. If set to
+  'old', the behavior is no different from what it was before this patch
+  appeared (gh-10181).
--- a/src/box/errcode.h
+++ b/src/box/errcode.h
@@ -438,6 +438,7 @@ struct errcode_record {
 	_(ER_REPLICA_GC, 291,			"Cannot clean up replica's resources", "uuid", STRING, "details", STRING) \
 	_(ER_ALIEN_ENGINE, 292,			"Snapshot contains alien space engine row", "engine", STRING) \
 	_(ER_MVCC_UNAVAILABLE, 293,		"MVCC is unavailable for storage engine '%s' so it cannot be used in the same transaction with '%s', which supports MVCC", "engine_without_mvcc", STRING, "engine_with_mvcc", STRING) \
+	_(ER_COMMIT_TIMEOUT, 294,		"Transaction timeout has expired during commit. It's not aborted and continues to commit in background.", "expired_at", STRING) \
 	TEST_ERROR_CODES(_) /** This one should be last. */
 
 /*
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -545,6 +545,7 @@ txn_begin(void)
 	txn->fiber = NULL;
 	txn->timeout = TIMEOUT_INFINITY;
 	txn->rollback_timer = NULL;
+	txn->begin_time = fiber_clock();
 	fiber_set_txn(fiber(), txn);
 	trigger_create(&txn->fiber_on_yield, txn_on_yield, NULL, NULL);
 	trigger_add(&fiber()->on_yield, &txn->fiber_on_yield);
@@ -1034,7 +1035,7 @@ txn_prepare(struct txn *txn)
 	if (txn_check_can_continue(txn) != 0)
 		return -1;
 
-	if (txn->rollback_timer != NULL) {
+	if (!box_begin_timeout_new_meaning && txn->rollback_timer != NULL) {
 		ev_timer_stop(loop(), txn->rollback_timer);
 		txn->rollback_timer = NULL;
 	}
@@ -1048,6 +1049,11 @@ txn_prepare(struct txn *txn)
 	 */
 	if (txn_check_can_continue(txn) != 0)
 		return -1;
+
+	if (box_begin_timeout_new_meaning && txn->rollback_timer != NULL) {
+		ev_timer_stop(loop(), txn->rollback_timer);
+		txn->rollback_timer = NULL;
+	}
 
 	assert(txn->psn == 0);
 	/* psn must be set before calling engine handlers. */
@@ -1187,8 +1193,20 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 		txn->fiber = NULL;
 		return 0;
 	}
-	while (!req->is_complete)
-		fiber_yield();
+	while (!req->is_complete) {
+		if (box_begin_timeout_new_meaning) {
+			double timeout = txn->begin_time +
+					 txn->timeout - fiber_clock();
+			if (fiber_yield_timeout(timeout)) {
+				diag_set(ClientError, ER_COMMIT_TIMEOUT,
+					 "WAL write");
+				txn->fiber = NULL;
+				return -1;
+			}
+		} else {
+			fiber_yield();
+		}
+	}
 	if (req->res < 0) {
 		diag_set_journal_res(req->res);
 		goto rollback_io;
@@ -1198,7 +1216,8 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 		assert(limbo_entry->lsn > 0);
 		int rc = txn_limbo_wait_complete(&txn_limbo, limbo_entry);
 		if (rc < 0) {
-			if (fiber_is_cancelled()) {
+			if (fiber_is_cancelled() ||
+			    box_begin_timeout_new_meaning) {
 				txn->fiber = NULL;
 				return -1;
 			} else {
@@ -1741,6 +1760,7 @@ txn_on_yield(struct trigger *trigger, void *event)
 		return 0;
 	}
 	if (txn->rollback_timer == NULL && txn->timeout != TIMEOUT_INFINITY) {
+		assert(txn->begin_time == fiber_clock());
 		size_t size;
 		txn->rollback_timer =
 			tx_region_alloc_object(txn, TX_OBJECT_EV_TIMER, &size);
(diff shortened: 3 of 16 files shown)


changelogs/unreleased/gh-10181-propagate-transaction-timeout-to-box-commit.md
```diff
@@ -0,0 +1,7 @@
+## bugfix/replication
+
+* A new compat option `compat.box_begin_timeout_meaning` has been added.
+  Controls the behavior of the timeout `box.begin{timeout=`. When set to 'new',
+  the transaction timeout is propagated further to `box.commit`. If 'old' is
+  set, the behavior is no different from what it was before this patch
```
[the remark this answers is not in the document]
Astronomax: Thanks! Fixed.

Serpentian:
Related: https://github.com/tarantool/tarantool/issues/9261:

```
tarantool> a = 17 for i = 0, 1000000000 do a = (a * a) % 31 end box.begin({timeout=5}) fiber.sleep(0) fiber.sleep(0) box.commit()
---
- error: Transaction has been aborted by timeout
```

It was decided not to deal with this in the scope of the current patch

src/box/txn.c
```diff
@@ -1030,7 +1031,7 @@ txn_prepare(struct txn *txn)
 	if (txn_check_can_continue(txn) != 0)
 		return -1;
 
-	if (txn->rollback_timer != NULL) {
+	if (!box_begin_timeout_new_meaning && txn->rollback_timer != NULL) {
```
Serpentian, line 1034: This looks confusing. Why with the old compat it works here and with the new one, several lines below. What is the reason for that? Is it a bug? If so, we should just move it below and fix in the separate commit

src/box/txn.c
```diff
@@ -1183,8 +1189,20 @@ txn_commit_impl(struct txn *txn, enum txn_commit_wait_mode wait_mode)
 		txn->fiber = NULL;
 		return 0;
 	}
-	while (!req->is_complete)
-		fiber_yield();
+	while (!req->is_complete) {
+		if (box_begin_timeout_new_meaning) {
```
Serpentian, line 1193: Nit: no sense in doing that in `while (!req->is_complete)` cycle. We cannot retry it anyway

src/box/txn_limbo.c
```diff
@@ -427,26 +433,29 @@ txn_limbo_wait_complete(struct txn_limbo *limbo, struct txn_limbo_entry *entry)
...
 	struct txn_limbo_entry *e;
-	bool is_first_waiting_entry = true;
-	rlist_foreach_entry(e, &limbo->queue, in_queue) {
-		if (e == entry)
-			break;
-		if (txn_has_flag(e->txn, TXN_WAIT_ACK) &&
-		    e->txn->fiber != NULL) {
-			is_first_waiting_entry = false;
-			break;
+	if (!box_begin_timeout_new_meaning &&
```
Serpentian, line 436: I didn't get, why do you need to make that conditional

src/box/txn.c
Astronomax: In the old behavior, the timeout did not cover `before_commit` triggers.
In the new behavior, the timeout covers the `before_commit` trigger and the entire commit after it. And when discussing the design, we decided that the transaction would be rolled back if the timeout expired while the triggers were executing.

src/box/errcode.h
```diff
@@ -436,6 +436,7 @@ struct errcode_record {
 	_(ER_SYNC_QUEUE_FULL, 289,		"The synchronous transaction queue is full") \
 	_(ER_KEY_PART_VALUE_OUT_OF_RANGE, 290,	"The value of key part exceeds the supported range for type", "partno", UINT, "field_type", STRING, "value", MSGPACK, "min", MSGPACK, "max", MSGPACK) \
 	_(ER_REPLICA_GC, 291,			"Cannot clean up replica's resources", "uuid", STRING, "details", STRING) \
+	_(ER_COMMIT_TIMEOUT, 292,		"Transaction timeout has expired during commit. It continues to commit.", "expired_at", STRING) \
```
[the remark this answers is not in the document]
Astronomax: Yes, it's more informative.

src/box/txn.c
Astronomax: AFAIU it's possible to observe spurious wakeup while using fiber (I don't see any execution sequence to reproduce, but apparently it's possible). Therefore, in all such places, we wrap `fiber_yield` in a loop with a condition check.
This is also recommended even when using `fiber_cond_wait` and `fiber_cond_wait_timeout`:
https://github.com/tarantool/tarantool/blob/222770d5242e23b91ba5bcf3d9ef7c919a04af40/src/lib/core/fiber_cond.h#L110-L130

src/box/txn_limbo.c
Astronomax: As we have decided (perhaps temporarily), if `compat.box_cfg_replication_synchro_timeout = 'old'` and `compat.box_begin_timeout_meaning = 'new'`, then we prefer `box_begin({timeout=...})` (`replication_synchro_timeout` not actually starts). And that's just what I do here.
Here we handle the rollback by `replication_synchro_timeout`. And we know that the `replication_synchro_timeout` timer was only started `if (!box_begin_timeout_new_meaning && replication_synchro_timeout_rollback_enabled)`:
```cpp
if (box_begin_timeout_new_meaning) {
	double timeout = txn->begin_time +
		txn->timeout - fiber_clock();
	rc = fiber_cond_wait_timeout(
		&limbo->wait_cond, timeout);
} else if (replication_synchro_timeout_rollback_enabled) {
	double timeout = start_time +
		replication_synchro_timeout - fiber_clock();
	rc = fiber_cond_wait_timeout(
		&limbo->wait_cond, timeout);
} else {
	rc = fiber_cond_wait(&limbo->wait_cond);
}
```

Astronomax:
It was required that the option `box_begin_timeout_meaning` could be switched to 'new' only if the compat option `replication_synchro_timeout` already was set to 'new'.
Many approaches have been considered: https://github.com/Astronomax/tarantool/commit/c29825f251099abfe0e0ac1b2e86ef7e19b358eb, https://github.com/Astronomax/tarantool/commit/6701529077f87bb8e195f916d0f3965d0553d25e, https://github.com/Astronomax/tarantool/commit/432276c3c19d0e65a0e83e2db56ad07d3e3a2341, but it was decided that they all complicate the implementation, while not being particularly useful for the user. We believe that manually switching compat options is no longer so relevant for the average user; now there is a config for this. Therefore, we will allow user to manually set 'new' in `box_begin_timeout_meaning` and 'old' in `replication_synchro_timeout`, but we will assume that in this case `box_begin_timeout_meaning` has the highest priority, so only it will work. But we will still check when setting options through the config, and we will prohibit such a combination of values.

Serpentian (approved):
Seems, I've no comments about the patch itself, let's ask @Totktonada totake a look at the config related part

src/lua/compat.lua
```diff
@@ -7,6 +7,8 @@ local OLD = false
 local ffi = require('ffi')
 local tweaks = require('internal.tweaks')
 
+local compat = { }
```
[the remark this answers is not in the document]
Astronomax: Fixed.

changelogs/unreleased/gh-10181-propagate-transaction-timeout-to-box-commit.md
```diff
@@ -0,0 +1,7 @@
+## bugfix/replication
+
+* A new compat option `compat.box_begin_timeout_meaning` has been added.
+  Controls the behavior of the timeout `box.begin{timeout=`. If set to 'new',
+  the transaction timeout is propagated further to `box.commit`. If set to
+  'old', the behavior is no different from what it was before this patch
+  appeared (gh-10181).
```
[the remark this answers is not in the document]
Astronomax: Ok. I left the details only in the doc request.

Astronomax:
A different approach was chosen: https://github.com/tarantool/tarantool/pull/11264.

It was decided to leave the `box.begin({timeout = ...})` transaction timeout behavior alone, but to introduce a new `txn_synchro_timeout` configuration option that would detach the user fiber from the synchronous transaction if it takes too long to gather a quorum.

Astronomax:
A different approach was chosen: https://github.com/tarantool/tarantool/pull/11264.

It was decided to leave the `box.begin({timeout = ...})` transaction timeout behavior alone, but to introduce a new `txn_synchro_timeout` configuration option that would detach the user fiber from the synchronous transaction if it takes too long to gather a quorum.

PR #10828, opened 2024-11-18
schema: fix the index:offset_of parameter handling

### LINKED ISSUE #10827: A problem with `index:offset_of` parameter variations
The function does not handle the `nil` as an `opts` argument and numeric keys.

```
tarantool> box.space.s1:offset_of({10})
---
- error: 'builtin/box/schema.lua:2530: attempt to index local ''opts'' (a nil value)'
...

tarantool> box.space.s1:offset_of(10)
---
- error: 'builtin/box/schema.lua:2528: attempt to get length of local ''key'' (a number
    value)'
...
```

Follows-up #10317.

mkostoevr:
The function assumes the given key is a Lua table and the given opts is not nil. Let's fix that.

Closes #10827

### DIFF
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -2525,13 +2525,14 @@ end
 -- the same tuple as index:pairs(key, {iterator = it})
 base_index_mt.offset_of = function(index, key, opts)
     check_index_arg(index, 'offset_of', 2)
+    key = keify(key)
     local itype = check_iterator_type(opts, #key == 0, 2)
     if itype == box.index.EQ then
-        opts.iterator = box.index.GE
+        itype = box.index.GE
     elseif itype == box.index.REQ then
-        opts.iterator = box.index.LE
+        itype = box.index.LE
     end
-    return index:len() - index:count(key, opts)
+    return index:len() - index:count(key, itype)
 end
 
 base_index_mt.get_ffi = function(index, key)
--- a/test/box-luatest/gh_8204_fast_offset_test.lua
+++ b/test/box-luatest/gh_8204_fast_offset_test.lua
@@ -376,6 +376,72 @@ g_generic.test_offset_of = function()
     end)
 end
 
+-- Checks the 'offset_of' function parameters.
+g_generic.test_offset_of_params = function()
+    g_generic.server:exec(function()
+        -- Create and fill a space.
+        local s = box.schema.space.create('test')
+        s:create_index('pk', {parts = {{1, 'unsigned'}, {2, 'unsigned'}}})
+        s:insert({1, 1})
+        s:insert({1, 2})
+        s:insert({2, 1})
+        s:insert({2, 2})
+        s:insert({3, 1})
+        s:insert({3, 2})
+
+        -- No arguments.
+        t.assert_equals(s:offset_of(), 0)
+
+        -- Empty key.
+        t.assert_equals(s:offset_of({}), 0)
+
+        -- Numeric key.
+        t.assert_equals(s:offset_of(2), 2)
+
+        -- Regular key.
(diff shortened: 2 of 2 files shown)


src/box/lua/schema.lua
```diff
@@ -2525,6 +2525,10 @@ end
 -- the same tuple as index:pairs(key, {iterator = it})
 base_index_mt.offset_of = function(index, key, opts)
     check_index_arg(index, 'offset_of', 2)
+    key = keify(key)
+    if opts == nil then
```
drewdzzz, line 2529: More compact version:
```lua
opts = opts or {}
```

src/box/lua/schema.lua
```diff
@@ -2525,6 +2525,10 @@ end
...
+        opts = {}
+    end
```
drewdzzz, line 2531: What's about invalid `opts`? Let's use `check_param_table` helper.

drewdzzz (changes requested):

src/box/lua/schema.lua
mkostoevr: Applied.

src/box/lua/schema.lua
mkostoevr: Introduced the invalid option check.

drewdzzz (approved):

src/box/lua/schema.lua
```diff
@@ -2525,6 +2525,9 @@ end
 -- the same tuple as index:pairs(key, {iterator = it})
 base_index_mt.offset_of = function(index, key, opts)
     check_index_arg(index, 'offset_of', 2)
+    key = keify(key)
+    opts = opts or {}
+    check_param_table(opts, {iterator = 'string, number'}, 2)
```
drewdzzz, line 2530: Already approved the PR, but then joined a live discussion. We use `check_param_table` only on DDL (probably, for the sake of performance). Let's not use it here, just as in other DML ops?

(I was the one who suggested to use it - sorry 😕)

mkostoevr: Removed the opts check.

I don't think we should check for any type except for `nil` now: using other types violates the function interface. So the only `opts = opts or {}` should be sufficient as a valid use case check.

drewdzzz: We can do something like
```diff
diff --git a/src/box/lua/schema.lua b/src/box/lua/schema.lua
index 2555998125..e395b5fa6c 100644
--- a/src/box/lua/schema.lua
+++ b/src/box/lua/schema.lua
@@ -2525,13 +2525,15 @@ end
 -- the same tuple as index:pairs(key, {iterator = it})
 base_index_mt.offset_of = function(index, key, opts)
     check_index_arg(index, 'offset_of', 2)
+    key = keify(key)
+    opts = opts or {}
     local itype = check_iterator_type(opts, #key == 0, 2)
     if itype == box.index.EQ then
-        opts.iterator = box.index.GE
+        itype = box.index.GE
     elseif itype == box.index.REQ then
-        opts.iterator = box.index.LE
+        itype = box.index.LE
     end
-    return index:len() - index:count(key, opts)
+    return index:len() - index:count(key, itype)
 end
 
 base_index_mt.get_ffi = function(index, key)
```
No extra checks, and it works in the same way as count. But without it:
```lua
tarantool> box.space._space:count(nil, 'EQ')
---
- 27
...

tarantool> box.space._space:offset_of(nil, 'EQ')
---
- error: 'builtin/box/schema.lua:2532: attempt to index local ''opts'' (a string value)'
...
```

mkostoevr: Nice solution, applied.

drewdzzz: No objections anymore, thanks for the fixes!

locker (approved):

Issue #10541, opened 2024-09-06
config: wait for database schema 3.0.0 instead of the latest one to write instance/replicaset names


Totktonada:
The [`applier/box_cfg`](https://github.com/tarantool/tarantool/blob/9dbaa6a9bc0d65984b417f8a76aa8373b6125d16/src/box/lua/config/applier/box_cfg.lua) code waits for the latest database schema to write the instance/replicaset names. After #10520 it is possible to wait for the 3.0.0 schema, not the latest one. Let's do it.

If #10549 bumps the schema version to 2.11.5, we can also allow to write names on 2.11.5+.

See also #10539.

It is neither feature, not a bugfix. Just a companion task for #10520.

Also tracked in TNTP-361
Part of TNTP-41

Issue #10539, opened 2024-09-06
config: create user/role/privileges on a new enough database schema


Totktonada:
Tarantool 3.3 will allow most of DDL operations on a non-upgraded instance if the database schema is 2.11.1 or newer (#10520).

Let's issue the DDL operations on a non-upgraded instance from the config's [credentials applier](https://github.com/tarantool/tarantool/blob/9dbaa6a9bc0d65984b417f8a76aa8373b6125d16/src/box/lua/config/applier/credentials.lua) if the database schema version is 2.11.1 or newer. Now it waits for the schema to be fully upgraded.

I propose to consider this task neither bugfix, not feature, but rather a companion activity to #10520.

Also tracked in TNTP-359
Part of TNTP-41

PR #10829, opened 2024-11-18
Apply credentials and persistent names earlier, without waiting for `schema.upgrade`

Serpentian:
Closes https://github.com/tarantool/tarantool/issues/10541
Closes https://github.com/tarantool/tarantool/issues/10539


test/config-luatest/names_upgrade_test.lua
```diff
@@ -101,31 +100,59 @@ local function assert_before_upgrade()
...
     g.instance_1:exec(function()
         box.ctl.wait_rw()
-        box.schema.upgrade()
+        t.assert_equals(box.space._schema:get{'version'}, {'version', 2, 11, 0})
+        -- Pretend, that we upgrade to 2.11.5 version, since it's
+        -- the first 2.11 release, which properly works with names.
+        box.internal.run_schema_upgrade(function()
+            box.space._schema:delete("max_id")
+            box.space._schema:replace{'version', 2, 11, 5}
+        end)
```
Totktonada, line 121: I don't want to be picky about test cases, but there are some thoughts around.

In the previous version of the testing scenario we verify a user journey: upgrading from 2.11.0 to the latest version.

The relaxation of the schema requirements that is implemented in this patch opens the second user scenario: running a non-upgraded 3.x. instance on the 2.11.5 schema version.

The test case now verifies a partial upgrade scenario (from 2.11.0 to 2.11.5), but it is not accessible for a user. However, two *user* scenarios above are not directly verified.

Should we write separate test cases for upgrading 2.11.0 to latest and for starting on 2.11.5?

src/box/lua/config/applier/box_cfg.lua
```diff
@@ -220,10 +220,10 @@ local function names_schema_upgrade_on_replace(old, new)
         return
     end
 
-    local latest_version = mkversion.get_latest()
+    local version = mkversion(2, 11, 5)
```
Totktonada, line 223: Nit: `required_version`, `wanted_version`, `needed_version`, `expected_version`, `waited_version` -- I would use something like this to make it clear that it is not some current schema verison.

test/config-luatest/upgrade_test.lua
```diff
@@ -118,7 +121,14 @@ g.test_upgrade = function()
         for _, alert in pairs(require("config"):info().alerts) do
             table.insert(messages, alert.message)
         end
+        t.assert_items_exclude(messages, {exp})
```
Totktonada, line 124: `exp` is a table, so I guess the `messages` array never contains it.

Totktonada (approved):
The changes seems quite straightforward and I don't remember anything that would prevent it from working as expected. LGTM.

test/config-luatest/names_upgrade_test.lua
Serpentian: Done. Do we want the same for credentials? I don't see much sense in testing upgrade from 2.11.5 in case of names and from 2.11.1 in case of credentials, since it's already covered by upgrade from 2.11.0, in which we go through the intermediate step of upgrading to 2.11.1 and 2.11.5

test/config-luatest/upgrade_test.lua
Serpentian: Good catch. Didn't notice, that the initial `exp` is changed afterwards. Fixed

test/box-luatest/upgrade/2.11.5/replicaset/instance-001/gen.lua
```diff
@@ -0,0 +1,16 @@
...
+    instance_uuid = '22188ab6-e0e5-484c-89b6-5dfe2b969b7b',
+    replication = {3301, 3302},
+    listen = 3301,
+}
+
+box.schema.user.grant('guest', 'super')
+while not box.info.replication[2] or
+      not box.info.replication[2].downstream or
+      box.info.replication[2].downstream.status ~= 'follow' do
+    require('fiber').yield(0.1)
```
Totktonada, line 13: A little typo: `fiber.yield()` doesn't accept an argument: `fiber.sleep()` does,

https://github.com/tarantool/tarantool/blob/9a17476508ce0eb2562dfff1357281e25e274e15/src/lua/fiber.c#L728-L734

test/config-luatest/names_upgrade_test.lua
Totktonada: I'm a bit worrying about such 'stop upgrading' steps that use knowledge about internals. It feels fragile.

I'm OK, but I'll reach you (in the good sense) if something goes wrong with it :)

Taking into account that the upgrading tests are not easy to write, I think it is nice that we have some.

test/box-luatest/upgrade/2.11.5/replicaset/instance-001/gen.lua
Totktonada: That's really minor. And this is the only comment. I'll close it and proceed.

Totktonada (approved):

Totktonada:
@Serpentian Thank you for the decent work!

PR #10832, opened 2024-11-19
sql: do not use raw index for count

### LINKED ISSUE #10825: SQL count doesn't start a transaction in engine
Consider a script:
```lua
-- Cleanup and start
os.execute('rm 000*')
os.execute('rm -r 512')
box.cfg{}

-- Create a vinyl space
box.schema.space.create('test', {engine = 'vinyl'})
box.space.test:format{{'field1', 'unsigned'}}

local json = require('json')
local ch = fiber.channel(1)

local function do_count()
    return json.encode(box.execute([[SELECT COUNT() FROM SEQSCAN test]]))
    -- return box.space.test:count()
end

-- Open a transaction and count tuples
box.begin()
print(do_count())

-- Make a concurrent insert
fiber.create(function()
    box.space.test:insert{1}
    ch:put(true)
end)

-- Wait for write to be committed and count tuples again
ch:get()
print(do_count())
box.commit()
```

Output:
```
{"metadata":[{"name":"COLUMN_1","type":"integer"}],"rows":[[0]]}
{"metadata":[{"name":"COLUMN_1","type":"integer"}],"rows":[[1]]}
```

Output if `do_count` uses Lua function instead of SQL:
```
0
0
```

The problem is SQL [uses](https://github.com/tarantool/tarantool/blob/e5aea49b6b22c38ebdb4c03b3ddd39838106e3a3/src/box/sql.c#L251) raw `index_count` that doesn't manage transactions - `box_index_count` should be used here. Also, the problem can lead to other side-effects (for example, aborted transaction can still use SQL count since the check is done in `box_index_count`). The reproducer uses vinyl because any transaction is registered in memtx (which is certainly an encapsulation violation, but that's another problem) so the count statement doesn't have to start the transaction in memtx.

drewdzzz:
Currently, we use raw index for count operation instead of `box_index_count`. As a result, we skip a check if current transaction can continue and we don't begin transaction in engine if needed. So, if count statement is the first in a transaction, it won't be tracked by MVCC since it wasn't notified about the transaction. The commit fixes the mistake. Also, the commit adds a check if count was successful and covers it with a test.

Closes #10825

NO_DOC=bugfix


locker (approved):

p7nov (approved):

locker:
Cherry-picked to 3.2.

@drewdzzz I failed to cherry-pick this change to 2.11. Please prepare a PR.

PR #10833, opened 2024-11-19
memtx: do not abort unrelated transactions on DDL

### LINKED ISSUE #10377: Do not abort all yielding memtx transactions when DDL is committed
Currently, any DDL operation aborts **all** yielding memtx transactions, even those that are not directly affected by it:

```lua

box.cfg{
    log_level = 'warn',
    memtx_use_mvcc_engine = true,
}

local s = box.schema.space.create('test')
s:create_index('pk')

box.begin()
s:insert({1})

local f = fiber.new(box.schema.space.create, 'test2')
f:set_joinable(true)
f:join()

box.commit()

os.exit(0)
```

Output:

```
2024-08-07 12:49:27.531 [143562] main/114/lua memtx_tx.c:700 W> Transaction committing DDL (id=392) has aborted another TX (id=391)
2024-08-07 12:49:27.532 [143562] main test.lua:18 E> ER_TRANSACTION_CONFLICT: Transaction has been aborted by conflict
2024-08-07 12:49:27.532 [143562] main F> fatal error, exiting the event loop
```

This is caused by the following code:

https://github.com/tarantool/tarantool/blob/625afe418151de1015483ed24acaa57e1f85efb3/src/box/memtx_engine.cc#L623-L624

In contrast to memtx, Vinyl aborts only affected transactions when a DDL operation is committed, see commit d3e12369565151ba8c0632a84adab4bf90eff23f. We should do something similar in memtx.

drewdzzz:
See the commits for details.

Closes #10377

### DIFF
--- a/changelogs/unreleased/gh-10377-do-not-abort-unrelated-on-ddl.md
+++ b/changelogs/unreleased/gh-10377-do-not-abort-unrelated-on-ddl.md
@@ -1,0 +1,4 @@
+## feature/memtx
+
+* Now DDL operations abort only transactions related to the space that
+  they are modifying (gh-10377).
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -192,6 +192,8 @@ struct point_hole_item {
 	size_t key_len;
 	/** Storage for short key. @key may point here. */
 	char short_key[16];
+	/** Saved space id. */
+	uint32_t space_id;
 	/** Flag that the hash tables stores pointer to this item. */
 	bool is_head;
 };
@@ -728,47 +730,10 @@ memtx_tx_acquire_ddl(struct txn *tx)
 }
 
 /**
- * Fully temporary and remote transactions cannot possibly conflict each other
- * by definition, so we can filter them out when aborting transactions for DDL.
- * Temporary space transactions are a always nop, since they never have WAL rows
- * associated with them.
- */
-static bool
-memtx_tx_filter_temporary_and_remote_txs(struct txn *txn1, struct txn *txn2)
-{
-	return (txn_is_fully_temporary(txn1) && txn_is_fully_remote(txn2)) ||
-	       (txn_is_fully_remote(txn1) && txn_is_fully_temporary(txn2));
-}
-
-/**
  * Clean and clear all read lists of @a txn.
  */
 static void
 memtx_tx_clear_txn_read_lists(struct txn *txn);
-
-void
-memtx_tx_abort_all_for_ddl(struct txn *ddl_owner)
-{
-	struct txn *to_be_aborted;
-	rlist_foreach_entry(to_be_aborted, &txm.all_txs, in_all_txs) {
-		if (to_be_aborted == ddl_owner)
-			continue;
-		if (txn_has_flag(to_be_aborted, TXN_HANDLES_DDL))
-			continue;
-		if (to_be_aborted->status != TXN_INPROGRESS &&
-		    to_be_aborted->status != TXN_IN_READ_VIEW)
-			continue;
-		if (memtx_tx_filter_temporary_and_remote_txs(ddl_owner,
-							     to_be_aborted))
-			continue;
-		to_be_aborted->status = TXN_ABORTED;
-		txn_set_flags(to_be_aborted, TXN_IS_CONFLICTED);
-		memtx_tx_clear_txn_read_lists(to_be_aborted);
-		say_warn("Transaction processing DDL (id=%lld) has aborted "
-			 "another TX (id=%lld)", (long long) ddl_owner->id,
-			 (long long) to_be_aborted->id);
-	}
-}
 
 /**
  * Fix position of @a txn in global read view list to preserve the list to
@@ -808,6 +773,7 @@ memtx_tx_adjust_position_in_read_view_list(struct txn *txn)
 static void
 memtx_tx_abort_with_conflict(struct txn *victim)
 {
+	assert(victim->psn == 0);
 	if (victim->status == TXN_ABORTED)
 		return;
 	if (victim->status == TXN_IN_READ_VIEW)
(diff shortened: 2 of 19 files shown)


changelogs/unreleased/gh-10377-do-not-abort-unrelated-on-ddl.md
```diff
@@ -0,0 +1,4 @@
+## feature/memtx
+
+* Now DDL operations abort only transactions related to a modified
+  space (gh-10377).
```
p7nov, line 4:
```suggestion
* Now DDL operations abort only transactions related to the space that
  they are modifying (gh-10377).
```

p7nov (approved):

src/box/memtx_tx.c
```diff
@@ -2996,13 +2961,87 @@ memtx_tx_delete_gap(struct gap_item_base *item)
...
+memtx_tx_abort_space_schema_readers(struct space *space, struct txn *ddl_owner)
 {
+	/* Abort all story and gap-with-sucessor readers. */
+	struct memtx_story *story;
+	rlist_foreach_entry(story, &space->memtx_stories, in_space_stories) {
+		struct tx_read_tracker *tracker, *tmp;
+		rlist_foreach_entry_safe(tracker, &story->reader_list,
+					 in_reader_list, tmp) {
+			if (tracker->reader != ddl_owner)
+				memtx_tx_abort_with_conflict(tracker->reader);
```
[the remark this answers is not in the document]
drewdzzz: We could use more fine-grained approach, but it would be more fragile. Imagine altering a primary index. Despite the fact your non-unique secondary index wasn't touched by user, it is actually rebuilt since it uses primary key as a tie-breaker. Are there more pitfalls like this? I believe we would miss one or another.

So let's simply abort all space readers. DDLs are rare actually, so it doesn't seem to be a problem.

src/box/memtx_tx.c
```diff
@@ -2996,13 +2961,87 @@ memtx_tx_delete_gap(struct gap_item_base *item)
...
+			continue;
+		struct txn_stmt *stmt;
+		stailq_foreach_entry(stmt, &txn->stmts, next) {
+			if (stmt->space == space) {
+				memtx_tx_abort_with_conflict(txn);
+				break;
+			}
+		}
+		struct point_hole_item *hole_item;
+		rlist_foreach_entry(hole_item, &txn->point_holes_list,
```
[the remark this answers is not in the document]
drewdzzz: Test `g.test_ddl_aborts_related_point_readers` covers point holes readers - they are used for `get` that hasn't read anything.

src/box/space.c
```diff
@@ -852,6 +852,9 @@ generic_space_swap_index(struct space *old_space, struct space *new_space,
 {
 	SWAP(old_space->index_map[old_index_id],
 	     new_space->index_map[new_index_id]);
+	/* Swap read gaps back - they belong to the old space. */
+	rlist_swap(&old_space->index_map[old_index_id]->read_gaps,
```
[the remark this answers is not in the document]
drewdzzz: Elaborated in commit message:
```
Note that we swap indexes on DDL before the old space is invalidated, so
all index read gaps are actually moved to the new space. It's actually
incorrect but didn't break anything - we delete all transaction's
trackers when it is prepared or aborted, and we used to abort all
in-progress transactions on DDL. But with the new approach, we need
those gaps in the old space in order to find all space readers, so the
commit swaps read gaps back when moving space indexes.
```

src/box/memtx_tx.c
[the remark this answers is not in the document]
drewdzzz: Created a separate parametrized test that checks that DDL does not abort itself and that we don't abort unrelated RO transactions as well.

test/engine-luatest/gh_10375_ddl_does_not_abort_unrelated_transactions_test.lua
```diff
@@ -63,3 +62,92 @@ g.test_ddl_aborts_related_transactions = function(cg)
...
+            box.space.test1:create_index('secondary', {parts = {2, 'unsigned'}})
+        end)
+        f:set_joinable(true)
+        t.assert_equals({f:join()}, {true})
+        t.assert_error_covers({
+            type = 'ClientError',
+            code = box.error.TRANSACTION_CONFLICT,
+        }, box.commit)
+    end, {cg.params.engine})
+end
```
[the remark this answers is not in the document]
drewdzzz: Created a separate parametrized test that checks that DDL does not abort itself and that we don't abort unrelated RO transactions as well.

CuriousGeorgiy (approved):

src/box/space.c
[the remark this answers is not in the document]
drewdzzz: Done:
```c
/*
 * Swap read gaps back - they belong to the old space. It is important
 * since the old space is going to be invalidated and transactional
 * metadata is required to do it correctly.
 */
```

src/box/memtx_tx.c
locker, line 2977: Shouldn't we skip prepared transactions?

src/box/memtx_tx.c
```diff
@@ -2996,13 +2961,87 @@ memtx_tx_delete_gap(struct gap_item_base *item)
...
+				    in_point_holes_list) {
+			if (space->def->id == hole_item->space_id) {
+				memtx_tx_abort_with_conflict(txn);
```
locker, line 3024: We must abort transactions writing to the space because otherwise, they'd crash on commit (`txn_stmt` stores a reference to a space). Do we really need to abort readers? We don't do that in Vinyl.

src/box/memtx_tx.c
drewdzzz: Oh, thanks!
Only active transactions can have trackers - when transaction is prepared or rolled back, all its trackers are deleted. So it makes sense only for iteration over `all_txs`. Fixed:
```diff
diff --git a/src/box/memtx_tx.c b/src/box/memtx_tx.c
index 6ce2cfbde2..98411acbea 100644
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -773,6 +773,7 @@ memtx_tx_adjust_position_in_read_view_list(struct txn *txn)
 static void
 memtx_tx_abort_with_conflict(struct txn *victim)
 {
+	assert(victim->psn == 0);
 	if (victim->status == TXN_ABORTED)
 		return;
 	if (victim->status == TXN_IN_READ_VIEW)
@@ -2963,8 +2964,10 @@ memtx_tx_delete_gap(struct gap_item_base *item)
 
 /**
  * Abort all concurrent readers of the old schema of the space.
+ * Note that only active transaction can have trackers since when a transaction
+ * is rolled back or prepared all its trackers are deleted.
  */
-void
+static void
 memtx_tx_abort_space_schema_readers(struct space *space, struct txn *ddl_owner)
 {
 	/* Abort all story and gap-with-sucessor readers. */
@@ -3008,6 +3011,9 @@ memtx_tx_abort_space_schema_readers(struct space *space, struct txn *ddl_owner)
 	 */
 	struct txn *txn;
 	rlist_foreach_entry(txn, &txm.all_txs, in_all_txs) {
+		if (txn->status != TXN_INPROGRESS &&
+		    txn->status != TXN_IN_READ_VIEW)
+			continue;
 		if (txn == ddl_owner)
 			continue;
 		struct txn_stmt *stmt;
```

The new assertion covers the problem.

locker: Would be nice to have a test using `ERRINJ_WAL_DELAY` that would ensure this. There's a test for this in Vinyl:

https://github.com/tarantool/tarantool/blob/017700dccdb04edd8470af620c69f3e7c572895b/test/vinyl/errinj_ddl.test.lua#L156-L176

src/box/memtx_tx.c
drewdzzz: 1. Some trackers store keys that are compared to tuples with `key_def`. It's dangerous to compare old keys with a new `key_def` - for example, an `!is_nullable` optimization could be enabled after schema change - old nullable key could lead to a crash in this case. By the way, isn't vinyl affected by this problem too? Doesn't it store keys in read sets of transactions?
2. On space invalidation we remove all the stories and insert tuples according to what `active_ddl` observes. If some RO transactions were not aborted, we would need to think if we don't violate their isolation.

locker: > 1. Some trackers store keys that are compared to tuples with `key_def`. It's dangerous to compare old keys with a new `key_def` - for example, an `!is_nullable` optimization could be enabled after schema change - old nullable key could lead to a crash in this case. By the way, isn't vinyl affected by this problem too? Doesn't it store keys in read sets of transactions?

Vinyl read sets are stored in indexes (`vy_lsm::read_set`), i.e. a read set is dropped along with the index and its key definition. Doesn't it work the same way in memtx?

> 2. On space invalidation we remove all the stories and insert tuples according to what `active_ddl` observes. If some RO transactions were not aborted, we would need to think if we don't violate their isolation.

AFAIU letting read-only transactions proceed wouldn't break transaction serializability - we could assume that they were executed before the DROP/ALTER statement. The code would be so much simpler then and consistent with Vinyl.

src/box/memtx_tx.c
drewdzzz: Yeah, sounds right. Done.

src/box/memtx_tx.c
drewdzzz: Discussed f2f - decided to abort all RO transactions in memtx.

locker (approved):

drewdzzz:
Decided to backport to 3.2 but not to 2.11.

locker:
Cherry-picked to 3.2.

PR #10834, opened 2024-11-19
box: expose box_lua_find via tnt_internal_symbol

mkokryashkin:
In some cases, one might want to access a deeply nested Lua function from C, which is exactly the functionality implemented by the `box_lua_find`. Since the function in question is quite complex to re-implement, it is beneficial to expose it for internal use.

NO_DOC=internal
NO_CHANGELOG=internal


Totktonada (approved):
No objections from me.

PR #10849, opened 2024-11-22
config: alert on a non-latest database schema version

### LINKED ISSUE #10542: config: alert on a non-latest database schema version
I propose to show an alert in `box.info.config.alerts` if the database schema is not the latest one. Otherwise, it may be unnoticed.

Let's also drop the alert when the database schema is upgraded to the latest version.

Also tracked in TNTP-362
Part of TNTP-41

grafin:
A persistent alert will be shown in `box.info.config.alerts` while the database schema is not the latest available. This alert wil be dropped as soon as the database schema is updated.

Closes #10542

NO_DOC=minor and not user configurable
NO_CHANGELOG=minor


src/box/lua/config/applier/box_status.lua
```diff
@@ -0,0 +1,56 @@
...
+        box_status_state.watcher:unregister()
+    end
+
+    box_status_state.config = config
+    check_schema_version()
+    box_status_state.watcher = box_status_watcher()
+end
+
+return {
+    name = 'box_schema',
```
Totktonada, line 54: Nit: I guess `box_status` is expected here.

test/config-luatest/upgrade_test.lua
```diff
@@ -101,6 +102,32 @@ g.test_upgrade = function()
             box.space._schema:replace{'version', 2, 11, 1}
         end)
 
+        local function alert_exists(alert)
+            local alerts = box.info.config.alerts
+            for _, a in pairs(alerts) do
```
Totktonada, line 107: Nit: I would use `ipairs` to highlight that it is an array-like table.

test/config-luatest/upgrade_test.lua
```diff
@@ -101,6 +102,32 @@ g.test_upgrade = function()
...
+        local err_msg = (
+            'The schema version %s is outdated, the latest ' ..
+            'version is %s. Please, consider using box.schema.upgrade().')
+            :format(mkversion(2, 11, 0), mkversion.get_latest())
+
+        -- Checks the alert for the outdated schema version exists.
+        t.assert(alert_exists({type = 'warn', message = err_msg}),
+                 'Outdated schema version alert exists')
+
+        box.schema.upgrade()
```
Totktonada, line 124: Seems like it breaks the idea of the privileges upgrade test below. We want to verify that the privileges are granted on a non-lates schema.

There is `box.schema.upgrade()` at end of the test case. Let's place the alert check after it.

Hm. The alert check is already there :)

src/box/lua/config/applier/box_status.lua
grafin: Yep, thanks for noticing!

test/config-luatest/upgrade_test.lua
grafin: Oh, I've left this extraneous `box.schema.upgrade()` here while resolving a merge conflict, thanks for pointing this out! Also I removed the redundant check for the alert not being present after schema was upgraded.

grafin: I've reworked the test changes.

src/box/lua/config/applier/box_status.lua
```diff
@@ -0,0 +1,56 @@
...
+    return box.watch('box.status', function(key)
+        assert(key == 'box.status')
+        check_schema_version()
+    end)
+end
+
+local function apply(config)
+    if box_status_state.watcher ~= nil then
+        box_status_state.watcher:unregister()
+    end
```
Totktonada, line 46: How about registering it on demand? Re-registering each time seems redundant.

grafin: I rewrote the `apply` method, now I don't store the config itself, requiring it when I need access to the aboard itself only. And register the watcher only one time (the first time the `apply` is called.

Totktonada (approved):

PR #10850, opened 2024-11-22
config: add failover.log option options

grafin:
Needed for EE tarantool/tarantool-ee#925.

NO_CHANGELOG=changelog is a part of tarantool/tarantool-ee#925 NO_TEST=will be tested as a part of tarantool/tarantool-ee#925

@TarantoolBot document
Title: New configuration options `failover.log`

Now `failover.log.to` and `failover.log.file` options can be used to configure failover logger output.

Those options behave in similar way to `log.to` and `log.file` options.

`failover.log.to` defines the location for failover logs. This option accepts the following values:
- `stderr`: write logs to the standard error stream.
- `file`: write logs to a file defined in `failover.log.file`.

`failover.log.file` specifies a file for logs destination. It is mandatory to set this option if `failover.log.to` is set to `file`. If `failover.log.to` is set to anything else, this option is ignored.

The option is enterprise-only since it's a part of the supervised failover coordinator.

### DIFF
--- a/src/box/lua/config/instance_config.lua
+++ b/src/box/lua/config/instance_config.lua
@@ -2394,6 +2394,23 @@ return schema.new('instance_config', schema.record({
                 }),
             }),
         }),
+        log = schema.record({
+            to = schema.enum({
+                'stderr',
+                'file',
+            }, {
+                default = 'stderr',
+            }),
+            file = schema.scalar({
+                type = 'string',
+            }),
+        }, {
+            validate = function(data, w)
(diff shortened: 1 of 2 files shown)


Totktonada (approved):
I've no objections, but, please

1) Fix the instance config test (it expects a new default value).
2) Adjust the linked issue number (as we discussed, another issue is more related here).

georgiy-belyanin (approved):
Everything looks good!

Issue #10727, opened 2024-10-21
Tarantool sends `GROUP_LOCAL` rows via replication


Serpentian:
Raft state always has `GROUP_LOCAL` id, even though it's sent to other replicas.

https://github.com/tarantool/tarantool/blob/caf08d41c013a0b022289a28d04b959ef1cf70ba/src/box/xrow.c#L1445

Rows with `GROUP_LOCAL` should not be sent over network, there's no reason for raft to be the exception

PR #10861, opened 2024-11-27
relay: fix sending raft requests with GROUP_DEFAULT id

Serpentian:
Commit 7aac790ef3e28 ("relay: send raft requests with GROUP_DEFAULT id") tried to send all raft rows with GROUP_DEFAULT id. However, relay didn't know the version_id of the instance during initial JOIN, so raft state was sent with GROUP_LOCAL id. The test didn't properly work.

This commit fixes relay by setting proper version_id during initial JOIN. The test is also fully rewritten.

Follows up #10727


sergepetrenko (approved):
Thanks for the fix!

PR #10865, opened 2024-11-27
config: add on_event callback for roles

### LINKED ISSUE #10538: config: on_rw callback in a role
## Context

`cartridge` [passes](https://github.com/tarantool/cartridge/blob/7d99ecf639e8bc979fe7fc1b7cc94ae8530e1620/cartridge/roles.lua#L457-L459) an `opts.is_master` flag into the role's `apply_config` function. Also, it calls the `apply_config` function on transition from RO to RW. (See also https://github.com/tarantool/doc/issues/4280.)

Tarantool 3.x roles are different in this regard. `apply` is called on a reconfiguration -- when a new config arrives. It is not called on a transition from RO to RW.

After #10520 an RW status will be suitable to perform (most of) DDL operations. We need a convenient way to write role's database schema initialization code.

A user needs to register/unregister a watcher now:

```lua
-- myrole.lua
local function schema_init()
    box.once('myrole_init', function()
        box.schema.space.create('myrole_space')
        box.space.myrole_space:create_index('pk')
    end)
end

local box_status_watcher

local function apply(role_cfg)
    if box_status_watcher == nil then
        box_status_watcher = box.watch('box.status', function(_, status)
            if not status.is_ro then
                schema_init()
                box_status_watcher:unregister()
                box_status_watcher = nil
            end
        end)
    end
    <...>
end

return {
    <...>,
    apply = apply,
}
```

Or spawn a fiber with `box.ctl.wait_rw()`.

Both variants are a bit too wordy for such a typical scenario.

## Proposal

I propose to add an `on_rw` callback into the role next to the `apply` and others:

```lua
-- myrole.lua

<...>

return {
    <...>,
    apply = apply,
    on_rw = schema_init,
}
```

This way config's code takes all the complexity around waiting RW status and give a role author a simple API for that.

Details:

* `on_rw` is called on startup after `apply` if the instance is started in the RW mode.
* `on_rw` callbacks from different roles are started in the same order as their `apply` functions (as written in `roles` in the YAML configuration with reordering, where one role depends on another).
* Call `apply` for all the roles and then `on_rw` for all the roles. It seems more consistent with a situation, when the RW mode is assigned after a startup.
* The callback should also be called if the database schema version is changed (i.e. `box.schema.upgrade()` is called) and the status of the instance is RW. It allows to issue a DDL that uses a new feature only when the database is ready for it. We should pass the schema version into the callback (a user may use #10521 to compare verisons).

  ```lua
  local version = require('version')

  local function on_rw(status)
      init_schema_base()

      if status.version >= version.fromstring('3.4.5') then
          init_schema_3_4_5()
      end
  end

  return {
      <...>
      on_rw = on_rw,
  }
  ```

Also tracked in TNTP-225
Part of TNTP-41

grafin:
config: add on_event callback for roles

This patch add `on_event` callback for roles.

Closes #10538
Closes TNTP-225

@TarantoolBot document
Title: `on_event` callback for roles

Roles can now set a `on_event` callback, which will be called
every time a `box.status` system event is broadcasted or the
configuration has been updated (after the `apply` has finished).

`on_event` callbacks are called one after another in the order
the roles are applied (according to their dependencies). Each
event (`box.status` or configuration change) triggers a series
of callbacks, next series (for a new event) will be triggered
after the previous finishes.

`on_event` callbacks are executed inside a `pcall` and if error is
raised, it is logged with 'error' level and the series execution
continues.

It is guaranteed that all of the `on_event` triggers are executed as
part of the configuration process (and 'ready' or 'check_warnings'
status is reached only after all of them are done).

When the callback is called, three arguments are provided,
`on_event(config, key, value)` where:
`config` is the current configuration,
`key` is `'config.apply'` if the callback was trigger by configuration
update, or `'box.status'` if it was triggered by `box.status` system
event.
`value` is the same as in `box.status` system event.

Example usage:

```lua
return {
    name = 'my_role',
    validate = ...,
    apply = ...,
    stop = ...,
    on_event = function(config, key, value)
        local log = require('log')

        log.info('on_event is triggered by ' .. key)
        log.info('is_ro: ' .. value.is_ro)

        local roles_cfg = config:get('roles_cfg')
        roles_cfg['my_role']...
    end,
}
```

### DIFF
--- a/src/box/lua/config/applier/roles.lua
+++ b/src/box/lua/config/applier/roles.lua
@@ -1,16 +1,94 @@
 local log = require('internal.config.utils.log')
+local fiber = require('fiber')
 
-local last_loaded = {}
-local last_roles_ordered = {}
+local roles_state = {
+    -- Loaded roles, where the key is the role name and the value is the
+    -- role object.
+    last_loaded = {},
+
+    -- Ordered list of roles, where the order is determined by the
+    -- dependencies between roles.
+    last_roles_ordered = {},
+
+    -- The box.status watcher, used to call on_event callbacks.
+    box_status_watcher = nil,
+
+    -- Last values received from box.status watcher.
+    last_box_status_value = nil,
+
+    -- Last applied config.
+    last_config = nil,
+}
+
+-- Function decorator that is used to prevent call_on_event_callbacks() from
+-- being called concurrently by watcher and post_apply().
+local lock = fiber.channel(1)
+local function locked(f)
+    return function(...)
+        lock:put(true)
+        local status, err = pcall(f, ...)
+        lock:get()
+        if not status then
+            error(err, 0)
+        end
+    end
+end
+
+-- Call all the on_event callbacks from roles in order.
+local call_on_event_callbacks = locked(function(config, key, value)
+    local roles_ordered = roles_state.last_roles_ordered
+    local roles = roles_state.last_loaded
+    for _, role_name in ipairs(roles_ordered) do
+        local role = roles[role_name]
+        if role.on_event ~= nil then
+            log.verbose(('roles.on_event: calling callback for role ' ..
+                         '"%s"'):format(role_name))
+            local ok, err = pcall(role.on_event,
+                                  config, key, value)
+            if not ok then
+                log.error(('roles.on_event: callback for role ' ..
+                           '"%s" failed: %s'):format(role_name, err))
+            end
+        end
+    end
+end)
+
+local function box_status_watcher()
+    return box.watch('box.status', function(key, value)
+        assert(key == 'box.status')
+
+        -- Store the last value to be able to call on_event callbacks
+        -- when the config is changed.
+        roles_state.last_box_status_value = table.deepcopy(value)
+
+        -- Config is not fully applied yet, don't call on_event callbacks.
+        -- We will call them later in post_apply.
+        if roles_state.last_config == nil then
+            return
+        end
+
+        call_on_event_callbacks(roles_state.last_config, 'box.status', value)
+    end)
+end
 
 local function apply(_config)
-    log.verbose('roles.apply: do nothing')
+    if roles_state.box_status_watcher == nil then
+        roles_state.box_status_watcher = box_status_watcher()
+
+        local deadline = fiber.time() + 10
+        while roles_state.last_box_status_value == nil do
+            fiber.sleep(0.01)
+            if fiber.time() > deadline then
+                error('Timeout reached while waiting for box_status_watcher', 0)
+            end
+        end
+    end
 end
 
 local function stop_roles(roles_to_skip)
     local roles_to_stop = {}
-    for id = #last_roles_ordered, 1, -1 do
-        local role_name = last_roles_ordered[id]
+    for id = #roles_state.last_roles_ordered, 1, -1 do
+        local role_name = roles_state.last_roles_ordered[id]
         if roles_to_skip == nil or roles_to_skip[role_name] == nil then
             table.insert(roles_to_stop, role_name)
         end
@@ -20,7 +98,7 @@ local function stop_roles(roles_to_skip)
     end
     local deps = {}
     for role_name in pairs(roles_to_skip or {}) do
-        local role = last_loaded[role_name] or {}
+        local role = roles_state.last_loaded[role_name] or {}
         -- There is no need to check transitive dependencies for roles_to_skip,
         -- because they were already checked when roles were started, i.e. if
         -- role A depends on role B, which depends on role C, and we stop
@@ -49,7 +127,7 @@ local function stop_roles(roles_to_skip)
     end
     for _, role_name in ipairs(roles_to_stop) do
         log.verbose('roles.post_apply: stop role ' .. role_name)
-        local ok, err = pcall(last_loaded[role_name].stop)
+        local ok, err = pcall(roles_state.last_loaded[role_name].stop)
         if not ok then
             error(('Error stopping role %s: %s'):format(role_name, err), 0)
         end
@@ -134,7 +212,7 @@ local function post_apply(config)
     -- Load roles.
     local loaded = {}
     for _, role_name in ipairs(roles_ordered) do
-        local role = last_loaded[role_name]
+        local role = roles_state.last_loaded[role_name]
         if not role then
             log.verbose('roles.post_apply: load role ' .. role_name)
             role = require(role_name)
@@ -180,8 +258,17 @@ local function post_apply(config)
         end
     end
 
-    last_loaded = loaded
-    last_roles_ordered = roles_ordered
+    roles_state.last_loaded = loaded
+    roles_state.last_roles_ordered = roles_ordered
+    roles_state.last_config = table.deepcopy(config)
+
+    -- box.status watcher values should be already set, because the watcher
+    -- has been registered during apply.
+    assert(roles_state.last_box_status_value ~= nil)
+
+    -- Call on_event callbacks after the config is fully applied.
+    call_on_event_callbacks(config, 'config.apply',
+                            roles_state.last_box_status_value)
 end
 
 return {
--- a/changelogs/unreleased/gh-10538-add-on-event-callback-to-roles.md
+++ b/changelogs/unreleased/gh-10538-add-on-event-callback-to-roles.md
@@ -1,0 +1,3 @@
+## feature/config
+
+* Introduced `on_event` callback for roles (gh-10538).
--- a/test/config-luatest/roles_test.lua
+++ b/test/config-luatest/roles_test.lua
@@ -636,3 +636,249 @@ g.test_role_is_not_a_table = function()
             'expected table, got number',
     })
 end
+
+-- Ensure that on_event callback is called correctly.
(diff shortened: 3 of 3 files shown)


lenkis (approved):

changelogs/unreleased/gh-10538-add-on-box-status-callback-to-roles.md
```diff
@@ -0,0 +1,3 @@
+## feature/config
+
+* Introduced `on_box_status` callback for roles (gh-10538).
```
Totktonada, line 3: We agreed that it would look more accurate to call it `on_event` and pass either `box.status` or something about config (`'config.apply'`?[^1]) into the `key` argument.

I still think that we should pass role's config as the first argument, so the same function can be assigned to `apply` and `on_event`.

[^1]: We have `config.info` event, but it is called at slightly different point. Maybe it is not a god idea to mix them.

src/box/lua/config/applier/roles.lua
```diff
@@ -1,16 +1,78 @@
...
 local function apply(_config)
-    log.verbose('roles.apply: do nothing')
+    if roles_state.box_status_watcher == nil then
+        roles_state.box_status_watcher = box_status_watcher()
+
+        local deadline = fiber.time() + 10
+        while roles_state.last_box_status_value == nil do
+            fiber.sleep(0.01)
+            if fiber.time() > deadline then
+                error('Timeout reached while waiting for box_status_watcher')
```
Totktonada, line 66: Nit: I report all the error from config's code using `error(<...>, 0`, because this way a user don't see some internal files/lines in the report.

If it is a user side error, for example, from config validation. Here I don't mind, just highlighted my personal preference. Feel free to ignore.

src/box/lua/config/applier/roles.lua
```diff
@@ -1,16 +1,78 @@
...
+    last_box_status_value = nil,
+
+    -- Last applied config.
+    last_config = nil,
+}
+
+-- Call all the 'on_box_status' callbacks from roles in order.
+local function call_on_box_status_callbacks(config, value)
+    for _, role_name in ipairs(roles_state.last_roles_ordered) do
+        local role = roles_state.last_loaded[role_name]
```
Totktonada, line 26: I guess a reconfiguration may occur during the iteration, changing these values.

I would copy these two tables before the cycle.

src/box/lua/config/applier/roles.lua
```diff
@@ -180,8 +242,16 @@ local function post_apply(config)
...
+    roles_state.last_loaded = loaded
+    roles_state.last_roles_ordered = roles_ordered
+    roles_state.last_config = table.deepcopy(config)
+
+    -- box.status watcher values should be already set, because the watcher
+    -- has been registered during apply.
+    assert(roles_state.last_box_status_value ~= nil)
+
+    -- Call on_box_status callbacks after the config is fully applied.
+    call_on_box_status_callbacks(config, roles_state.last_box_status_value)
```
Totktonada, line 254: If `last_config` is non-nil, then the watcher may call the callbacks. If some of them yield for different (say, random) time intervals, it is possible that `on_box_status` is called before first `apply` for the same role.

I would prefer explicit latch variables: don't call `on_box_status` from the watcher until some `roles_state.post_apply_finished` (and set it after calling the callbacks from here).

src/box/lua/config/applier/roles.lua
```diff
@@ -1,16 +1,78 @@
...
+
+    -- Last values received from box.status watcher.
+    last_box_status_value = nil,
+
+    -- Last applied config.
+    last_config = nil,
+}
+
+-- Call all the 'on_box_status' callbacks from roles in order.
+local function call_on_box_status_callbacks(config, value)
```
Totktonada, line 24: I would pass an event key here and pass it to `on_box_status` (or `on_event` if we rename it). I would pass a different key for the reconfiguration event. It gives a bit more information for a user and costs us nothing.

(Maybe it also helps in testing -- we can differentiate an origin of the call.)

src/box/lua/config/applier/roles.lua
Totktonada, line 254: I would think about guarantees behind this API that we offer to a user.

Let's assume that we have roles `foo`, `bar`, `baz` placed in the configuration is this order.

We have the following order of initialization (loading) the roles: `foo`, `bar`, `baz`.

Let's assume (for simplicity) that the roles are independent or dependent as `baz` depends on `bar`, `bar` depends on `foo`. So, the reordering gives the same order: `foo`, `bar`, `baz`.

We have the following order of callbacks:

* `foo.validate`
* `bar.validate`
* `baz.validate`
* `foo.apply`
* `bar.apply`
* `baz.apply`

It is guaranteed that the first one ends its execution before the next is called. Also, it is guaranteed that all of them are executed as part of the configuration process (and we can't reach `ready` or `check_warnings` status until all of them are finished).

The `on_event` is different: it occurs on some arbitrary point of time, out of the configuration process.

However, we guarantee the order of calling them and that the callbacks that are triggered by one event are executed one after the previous finished.

We also guarantee that neither of the `on_event` callbacks are called before `apply` for.. one given role?.. for all the roles?

At least, it gives us some predictable behaviour.

It is still not clear, how one roles should notify another one that it finishes its runtime/database initialization. Maybe some service registry or just a suggestion about proper using of globals. It seems like we should leave it out of scope here.

Anyway, I would like to see some description about guarantees behind this API, so a user wouldn't feel itself on a vague ground. Can you verify my wording, adjust/rephrase/shorten if needed and include it into the documentation request?

Totktonada: Another point for the documentation is how the errors are handled. Unlike `apply`, an error from the `on_event` callback is ignored (with logging a message on the `error` level). Even if the callback is called because the configuration is changed, its error is just logged and doesn't fail the (re)configuration process.

Next, it would be good to make it clear, whether the `on_event` callbacks are executed out of the (re)configuration process in terms of time: does it waiti for finishing them. IOW, whether the configuration is successfully applied only after execution of these callbacks (if called due to the configuration change) or it is performed in background just like as when the callbacks are called due to a `box.status` event.

I don't know, which behavior is more expected, TBH. It is OK to perform some long-running activity from a usual watcher callback? I guess, no. So, let's keep the synchronous execution as part of the (re)configuration process.

(In some future we need alerts about long execution of role's function or application initialization code. An internal applier execution.)

Totktonada:
Seems nice for me, thanks! I'm waiting for a change in the callback name, passing different event keys on (re)configuration and `box.status` event, more info in the documentation request. And we can go.

src/box/lua/config/applier/roles.lua
Totktonada: Also, we discussed with @grafin idea to copy `locked` pattern from `load_cfg.lua` and use it for the `call_on_box_status_callbacks` function: this way we have guaratees similar as for watchers disregarding that we have two source of events: reconfigration and box.status:

> A watcher callback never runs in parallel with itself: if the key to which a watcher is subscribed is updated while the watcher callback is running, the callback will be invoked again with the new value as soon as it returns.

(From the watcher's [doc](https://github.com/tarantool/doc/issues/2407).)

changelogs/unreleased/gh-10538-add-on-box-status-callback-to-roles.md
grafin: Renamed it to `on_event`, `key = 'box.status` if the callback was triggered by `box.status` system event, and `key = 'config.apply'` if it was triggered because of configuration update.

src/box/lua/config/applier/roles.lua
grafin: Fixed!

src/box/lua/config/applier/roles.lua
grafin: Thanks for noticing! Fixed.

src/box/lua/config/applier/roles.lua
grafin: Thanks for helping in finding this solution! Implemented.

src/box/lua/config/applier/roles.lua
grafin: Done. `key = 'box.status'` if the callback was triggered by `box.status` system event, and `key = 'config.apply'` if it was triggered because of configuration update.

src/box/lua/config/applier/roles.lua
grafin: Added a more detailed doc, describing the order of callback execution as well as what happens if a callback raises an error.

grafin:
@lenkis I've expanded the doc request, can you please take a second look.

src/box/lua/config/applier/roles.lua
Totktonada: I see now: no need to copy, these tables are reassigned to the new ones each time and not modified. However, `last_loaded` could be changed, so my comment was partially relevant.

So, storing the references before the cycle is actually enough. Thanks!

Totktonada:
@grafin @lenkis I'm going to proceed with the patch tonight. I hope nobody mind to discuss details about the documentation within the corresponding documentation team task.

Totktonada (approved):

PR #10866, opened 2024-11-27
cmake: fix configuration error in benchmark module

ligurio:
Depends on https://github.com/tarantool/small/pull/101
CI in Tarantool repository: https://github.com/tarantool/tarantool/pull/10868

----

CMake option `DOWNLOAD_EXTRACT_TIMESTAMP` is supported since 3.24 [1], when option is used in ExternalProject_Add and CMake version below 3.24 is used on configuration stage an error is produced:

```
CMake Error at /usr/share/cmake-3.22/Modules/ExternalProject.cmake:2787 (message):
  URL_MD5 is set to

    21a2604efeded8b4cbabc72f3e1c7a2a;DOWNLOAD_EXTRACT_TIMESTAMP;TRUE

  but must be a hex string.
  /usr/share/cmake-3.22/Modules/ExternalProject.cmake:3716 (_ep_add_download_command)
  cmake/BuildBenchmark.cmake:22 (ExternalProject_Add)
  perf/CMakeLists.txt:19 (include)
```

The patch fixes that by moving `DOWNLOAD_EXTRACT_TIMESTAMP` to the end of ExternalProject_Add arguments.

1. https://cmake.org/cmake/help/latest/module/ExternalProject.html#url

NO_CHANGELOG=build
NO_DOC=build
NO_TEST=build


Buristan (approved):
Hi, Sergey!
Thanks for the patches!
LGTM!

PR #10867, opened 2024-11-27
test: parallelize config-luatest/compat_test

Totktonada:
It spawns a lot of processes and sometimes hits a test timeout limit on my machine, when run in parallel with other tests.

Also, if I run the single test with the parallelization, the parallel execution reduces time from 45 to 25 seconds.


mandesero (approved):

Totktonada:
`perf_micro` is to be fixed in #10866.

Totktonada:
Backported to `release/3.2` in PR #10894.

PR #10872, opened 2024-11-28
config: add alerts namespaces

grafin:
WIP

NO_DOC=WIP
NO_CHANGELOG=WIP

Closes TNTP-226


Totktonada (approved):
Seems nice! Thank you for the patch!

I've only one optional remark: I often use a string deduplication key to report the same alert multiple times and be sure that only one (the last) is reported to the user. It is useful, when you react on several events and the same alerting condition is hold in these several calls. We can name a variant of a method to add such an alert `:set()` and require a deduplication key to be provided. Please, prefix the key name for the `aboard` machinery with the namespace name.

changelogs/unreleased/gh-10873-alerts
Totktonada, line 1: Please, add `.md` to the file name. Otherwise `tools/gen-release-notes` doesn't see it.

grafin: Fixed!

grafin:
I've added a `set(key, alert)` method. Maybe an `unset` is needed too?

Issue #10132, opened 2024-06-14
Define msgpack MP_INT type behavior


ImeevMA:
Currently Tarantool assumes in some places that MP_INT only contains negative integers, but this is different from the msgpack definition of MP_INT, where MP_INT can contain both negative and positive integers. It seems like we should either mention in the documentation that we don't support non-negative MP_INT values, or add appropriate support for them.

oleg-jukovec:
> It seems like we should either mention in the documentation that we don't support non-negative MP_INT values

Then we will lose compatibility with MessagePack for no reason.

sergos:
> > It seems like we should either mention in the documentation that we don't support non-negative MP_INT values
> 
> Then we will suddenly lose compatibility with MessagePack for no good reason. At least we could cast `MP_INT > 0` -> `MP_UINT` at the encoding/decoding stage.

This will impose additional overhead on the precious TX, especially since it should be applied to all fields in the msgpack: one can insert a lot of data and thereafter try to introduce an index (the really broken part) on an arbitrary field.

So I would not name it as 'lost of compatibility' since it was not there from day one (or so) rather introduce a dedicated rules to use the storage in the only correct way.

oleg-jukovec:
> This will impose additional overhead on the precious TX, especially since it should be applied to all fields in the msgpack: one can insert a lot of data and thereafter try to introduce an index (the really broken part) on an arbitrary field.

I didn't mean to check all incoming data and transform it, but just make an internal typecast MP_INT > 0 to MP_UINT in the internal structs on the decoding stage if needed.

ahmadhabibi14:
Any news for update ?

Gerold103:
In scope of this ticket must be resolved some specific places, which rely on `mp_decode_int()` returning negative values:
* `mem_cmp_msgpack()` saves `mp_decode_int()` result as `MEM_TYPE_INT` which is strictly negative. It breaks in many places, from what it looks like.
* `mem_from_mp_ephemeral()` - same.
* `port_c_get_vdbemem()` - same.

Issue #10084, opened 2024-06-03
SQL requests decode msgpack's int64 incorrectly


oleg-jukovec:
**Bug description**

* OS: Linux
* OS Version: Archlinux
* Architecture: amd64

```
$ tarantool --version
Tarantool 3.1.0-0-g96f6d8859
Target: Linux-x86_64-Release
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=TRUE
Compiler: GNU-14.1.1
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ozi/src/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -O3 -DNDEBUG 
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ozi/src/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -O3 -DNDEBUG
```

**Steps to reproduce**

The reproducer script (`test.lua`):

```Lua
#!/usr/bin/env tarantool

box.cfg{}

local s = box.schema.space.create('q', {
    if_not_exists = true,
    format = {
        {name = "id", type = "integer"},
    },
})
s:create_index('primary', {
    type = 'tree',
    parts = {1, 'int'},
    if_not_exists = true
})

s:replace(require('msgpack').object_from_raw(string.fromhex("91d30000000000000001")))
s:replace(require('msgpack').object_from_raw(string.fromhex("91d300000000665c9ec4")))

print("space:select():")
print(require('json').encode(s:select({}, {limit = 10})))
print("\nSQL select:")
print(require('json').encode(box.execute("SELECT * FROM seqscan q;")))

os.exit()
```

How to run:

Just type `tarantool test.lua`.

**Actual behavior**

The output:
```
space:select():
[[1],[1717345988]]

SQL select:
{"metadata":[{"name":"id","type":"integer"}],"rows":[[-31],[-28]]}
```

**Expected behavior**

```
space:select():
[...]

PR #10883, opened 2024-12-02
Suboptimal msgpuck support

### LINKED ISSUE #9965: Ability to insert duplicate keys in unique index

Gerold103:
Not including some problems of `MP_INT` vs `MP_UINT` in SQL.
Closes #9965.
Related to #10132, #10084.

Here is a related discussion: https://github.com/orgs/tarantool/discussions/11017.


src/box/tuple_bloom.c
```diff
@@ -217,7 +217,18 @@ tuple_bloom_maybe_has(const struct tuple_bloom *bloom, struct tuple *tuple,
 	uint32_t h = HASH_SEED;
 	uint32_t carry = 0;
 	uint32_t total_size = 0;
-
+	if (bloom->version == TUPLE_BLOOM_VERSION_V2) {
```
locker, line 220: The function looks weird with the two loops. IMO it'd be better to fold the version check in the loop body.

src/box/tuple_hash.cc
```diff
@@ -285,9 +371,10 @@ key_def_set_hash_func(struct key_def *key_def) {
 	key_def->key_hash = key_hash_slowpath;
 }
 
-uint32_t
-tuple_hash_field(uint32_t *ph1, uint32_t *pcarry, const char **field,
-		 enum field_type type, struct coll *coll)
+template<tuple_hash_version version>
```
locker, line 374: Do we really want to make this function support two formats? Since it's used only by the bloom filter, wouldn't it be better to copy the legacy implementation to `tuple_bloom.c`?

Gerold103: Nope, I didn't want it to support 2 formats. But I couldn't find a good solution here. The motivation is that either it is like this, or we are going to have some hashing functions duplicated with no test coverage, or we would need to introduce test coverage for them.

I decided the template would be the least intrusive solution while still keeping those legacy hash functions at least partially covered by tests.

Do you think it is better to duplicate? Do we need to also duplicate the tests then?

locker: I would prefer to just copy the old implementation to `tuple_bloom.c`. Code duplication is fine in this case IMO because we'll eventually drop them. I don't think we need to test these functions specifically because we aren't ever going to extend/patch them so the tests checking the legacy bloom filter functionality in Vinyl should be enough. Let's just add a few more indexes with some data to `gh_9965_data/gen.lua`  (double, number, string) to formally cover the code computing the legacy hash. For example, create an index storing number from 1 to 10 with step 0.5 and check that `get` works as expected.

src/box/tuple_bloom.c
Gerold103: That would mean the check happens multiple times instead of one. I didn't measure performance of this place, but that surely would be suboptimal. Should we keep it like this then? Or do you still think that better check it on each iteration?

Gumix (approved):

lenkis (approved):

Gerold103:
Don't look yet, I accidentally re-requested a review. Having a weird fail in unit/msgpack.cc test in release build with clang, debugging it.

test/box-luatest/gh_9965_non_compact_mp_test.lua
```diff
@@ -0,0 +1,224 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group('gh-9965-non-compact-mp', {
+    {engine = 'memtx'}, {engine = 'vinyl'}
```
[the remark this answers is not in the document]
Gerold103: Moved!

src/box/tuple_hash.cc
Gerold103: Done! I copied the functions and added more cases to gen.lua file.

locker (approved):

Gumix (approved):

Gerold103:
During rebase I did a change in the vinyl 9965 test. I turned off the vinyl cache, and noticed that in the old run files we won't be able to find a double key, if the search-key is float. And vice versa. This is matching the old behaviour I guess, which we intentionally keep for the old run-files. I caught this flaky test in static-build job, which somewhy had vinyl cache set to zero and managed to catch the problem. Now I made the cache zero in all 9965 tests.

Gerold103:
I will backport it manually.

PR #10885, opened 2024-12-03
box: drop long select warning

locker:
There's a better mechanism for detecting stalled fibers - fiber slice. It first logs a warning if a fiber doesn't yield for too long, then kills the fiber. The long select warning was added before the fiber slice feature was introduced. Initially, it was pretty annoying so we had to improve it over the time. For example, we disabled it for system spaces and introduced the "fullscan" option to suppress it.

Recently, select with offset became fast, see commit 421d36f2e144 ("memtx: accelerate select with offset in tree") but we still log the warning if the offset is > 1000. Instead of fixing this feature once again, let's accept that it's useless and finally drop it.

This commit drops the Lua log ratelimiter along the way because it isn't used anywhere else or covered by any tests, but doesn't remove "fullscan" option usage from tests - the option is silently ignored now.

This commit effectively rolls back the following commits:
 - commit 9b7467d1e05e5 ("lua: introduce `Ratelimit` class")
 - commit 7aa02f302d819 ("box: log dangerous `select` calls")
 - commit 29654ffe3638e ("box: improve check for dangerous `select` calls")
 - commit 4221a983ffea2 ("box: refactor `check_select_safety` and add it to `box.internal`")


sergos (approved):
LGTM.
Although it'll be good to mention there's dedicated test on timeslice (gh_6085_limit_iteration_in_space_test.lua).

Totktonada (approved):
No objections from me.

lenkis (approved):

PR #10899, opened 2024-12-05
Tuning perf_micro workflow

ligurio:
- `pull_request_target` was tested on PR https://github.com/sqaunderhood/tarantool/pull/1
- CI in main repository https://github.com/tarantool/tarantool/pull/10954

### DIFF
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -63,8 +63,8 @@ jobs:
     timeout-minutes: 60
 
     env:
-        PERF_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
-        PERF_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
+      PERF_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
+      PERF_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
 
     steps:
       - name: Prepare checkout
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -8,6 +8,8 @@ on:
     tags:
       - '**'
   pull_request:
+  pull_request_target:
+    types: [labeled]
   workflow_dispatch:
   workflow_call:
     inputs:
@@ -44,11 +46,16 @@ concurrency:
 jobs:
   perf_micro:
     # Run on push to the 'master' and release branches of tarantool/tarantool
-    # or on pull request if the 'notest' label is not set.
+    # or on pull request if the 'notest' label is not set
+    # or on pull request in the main repository if the
+    # 'performance' label is set.
     if: github.repository_owner == 'tarantool' &&
-        ( github.event_name != 'pull_request' ||
+        ( ( github.event_name != 'pull_request' &&
+            github.event_name != 'pull_request_target' ) ||
           ( github.event_name == 'pull_request' &&
-            !contains(github.event.pull_request.labels.*.name, 'notest') ) )
+            !contains(github.event.pull_request.labels.*.name, 'notest') ) ||
+          ( github.event_name == 'pull_request_target' &&
+            contains(github.event.pull_request.labels.*.name, 'performance') ) )
 
     # 'performance' label _must_ be set only for the single runner
     # to guarantee that results are not dependent on the machine.
@@ -76,7 +83,7 @@ jobs:
           fetch-depth: 0
           submodules: recursive
           repository: tarantool/tarantool
-          ref: ${{ inputs.submodule && 'master' || github.ref }}
+          ref: ${{ inputs.submodule && 'master' || env.PERF_COMMIT }}
       - uses: ./.github/actions/environment
       - name: Install deps
         uses: ./.github/actions/install-deps-debian
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -19,6 +19,8 @@ on:
         description: Git revision from submodule repository
         required: true
         type: string
+  schedule:
+    - cron: '0 0 * * *'  # Once a day at midnight.
 
 concurrency:
   # Update of a developer branch cancels the previously scheduled workflow
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -7,6 +7,14 @@ on:
       - 'release/**'
     tags:
       - '**'
+    paths:
+      - '.github/workflows/perf_micro.yml'
+      - 'CMakeLists.txt'
+      - 'cmake/**'
+      - 'patches/**'
+      - 'perf/**'
+      - 'src/**'
+      - 'third_party/**'
   pull_request:
   pull_request_target:
     types: [labeled]


Buristan (approved):
Hi, Sergey!
Thanks for the patch set!
I'll proceed with the review per patch below.

---

**[PATCH 1/5] ci: fix indentation**

LGTM.

---

**[PATCH 2/5] ci: schedule perf_micro every day midnight**

LGTM, except a few nits regarding the commit message.

> ci: schedule perf_micro every day midnight

Typo: s/midnight/at midnight/

> master branch will be rarely.

Typo: s/master/the master/
Typo: s/rarely/rare/

.github/workflows/perf_micro.yml
```diff
@@ -76,7 +84,7 @@ jobs:
           fetch-depth: 0
           submodules: recursive
           repository: tarantool/tarantool
-          ref: ${{ inputs.submodule && 'master' || github.ref }}
+          ref: ${{ inputs.submodule && 'master' || env.PERF_COMMIT }}
```
Buristan, line 94: Side note: don't quite get this change.
Is it necessary for the target_pull_request, or is this an unrelated fix?

.github/workflows/perf_micro.yml
```diff
@@ -58,7 +65,8 @@ jobs:
       - self-hosted
       - Linux
       - x86_64
-      - ${{ (github.event_name != 'pull_request' ||
+      - ${{ ((github.event_name != 'pull_request' ||
```
Buristan, line 75: I am a little confused by this change:

Looks like the condition 
```
(github.event_name != 'pull_request' || github.event_name != 'pull_request_target')
```
is always true (`pull_request_target != pull_request` and `pull_request != pull_request_target`).

I am not sure that we can use `github.event.pull_request.head.repo.full_name`, when the workflow is triggered by the `pull_request_target` event.

I suppose we can drop this change since the check for `pull_request_target` is made above.

Buristan (changes requested):
**[PATCH 3/5] ci: allow to run perf_micro in main repository**

I like the changes generally, but have a comment and a few nits regarding the commit message.

---
 
I suggest the following rewording for the commit title:
```
ci: allow run perf_micro from fork by label
```
Feel free to ignore.

> To be able to detect performance regressions before merge we

Typo: s/merge/merging/

>    and visualize them.

Typo: s/visualize/visualizing/

> The machine for running benchmark is valuable resource and it can

Typo: s/benchmark/benchmarks/
Typo: s/valuable resource/a valuable resource,/

> every push like we do with other workflows. To avoid this th

Typo: s/To avoid this th/To avoid this, the/

> patch introduce a label 'performance', `perf_micro` workflow will

Typo: s/introduce/introduces/

Side note: Maybe it is better to say "uses" instead since the label already exists and is not introduced by the patch itself.

> The second item requires an access to a secret token which is not

Typo: s/an access/access/
Typo: s/which/that/

> available in workflows executed in pull request. To workaround

Typo: s/in pull request/in the pull request/

> event does. Running workflow in context of the base allows to use

Typo: s/Running workflow in context/Running the workflow in the context/

Buristan (commented):
**[PATCH 4/5] ci: remove 'performance' once job is started**

---

I am not sure that this will be convenient for reviewers:

Once the label is gone, it is easy to forget that it was ever set, since it may be lost in the discussions inside of the PR (even if we add the specific comment about it, it has the same problem).

Also, IINM, `pull_request_target` is triggered only when the [label](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#label) is added:

> If you want to run your workflow when a label is added to or removed from an issue, pull request, or discussion, use the `labeled` or `unlabeled` activity types for the `issues`, `pull_request`, `pull_request_target`, or `discussion` events instead.

So, I see no reason to remove it.

.github/workflows/perf_micro.yml
```diff
@@ -7,6 +7,13 @@ on:
       - 'release/**'
     tags:
       - '**'
+    paths:
+      - 'cmake/**'
+      - 'CMakeLists.txt'
+      - '.github/workflows/perf_micro.yml'
+      - 'perf/**'
+      - 'src/**'
+      - 'third_party/**'
```
Buristan, line 16: I would also add `patches/**` here.

Minor: please sort entries alphabetically.

Buristan (approved):
**[PATCH 5/5] ci: reduce a number of perf_micro runs**

---

LGTM, with a minor comment below.

.github/workflows/perf_micro.yml
ligurio: > I would also add patches/** here.

Added.

> Minor: please sort entries alphabetically.

What's wrong with a current order?

.github/workflows/perf_micro.yml
ligurio: Without hunk, workflow will be executed in base branch commit.

screen-1734426556
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

ligurio:
> [PATCH 4/5] ci: remove 'performance' once job is started
> I am not sure that this will be convenient for reviewers:

ok, removed

.github/workflows/perf_micro.yml
ligurio: reverted:

```diff
--- a/.github/workflows/perf_micro.yml
+++ b/.github/workflows/perf_micro.yml
@@ -73,8 +73,7 @@ jobs:
       - self-hosted
       - Linux
       - x86_64
-      - ${{ ((github.event_name != 'pull_request' ||
-              github.event_name != 'pull_request_target') ||
+      - ${{ (github.event_name != 'pull_request' ||
           github.event.pull_request.head.repo.full_name == github.repository) &&
           'performance' || 'regular' }}
 ```

.github/workflows/perf_micro.yml
Buristan: I suppose that `.github/...` should be before `[Cc]make**`, feel free to ignore.

opomuc (approved):

mkokryashkin (approved):
I'll proceed with per-patch review below.

[ci: reduce a number of perf_micro runs](https://github.com/tarantool/tarantool/pull/10899/commits/d8e1219e35c18c6b105577ab49ab81acae5ae8f9)

I believe, the name is a bit unaligned with the extended description of the commit. It implies some kind of a counter of runs being present there, which is clearly not the case. It should be something like "ci: add paths filter to perf_micro workflow". Feel free to ignore.

[ci: allow to run perf_micro in main repository](https://github.com/tarantool/tarantool/pull/10899/commits/edebdf9a7e5c45f34487948d49e55709bbf66d21)
The same thing with the commit title. Something like "ci: run perf_micro workflow on performance label" is more suitable. Feel free to ignore.

Other patches LGTM.

ligurio:
> It should be something like "ci: add paths filter to perf_micro workflow". Feel free to ignore.

Updated.

> The same thing with the commit title. Something like "ci: run perf_micro workflow on performance label" is more suitable. Feel free to ignore.

Updated.

PR #10902, opened 2024-12-06
Fix iproto index name to ID mapping

### LINKED ISSUE #9923: Invalid IPROTO_INDEX_NAME mapping on the server side.
**Bug description**

The server could not correctly map an index from a string passed with IPROTO_INDEX_NAME to a number.
I found the problem with the `tarantool/go-tarantool` connector, but I'm able to reproduce it with `net.box`.

* OS: Arhclinux
* Architecture: amd64

Place `tarantool --version` output here.

```
$ tarantool --version
Tarantool 3.1.0-entrypoint-214-ge048a5021
Target: Linux-x86_64-Debug
Compiler: GNU-13.2.1
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ozi/src/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -Werror -g -ggdb -O0 
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ozi/src/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror -g -ggdb -O0
```

**Steps to reproduce**

1. Run the script with `tarantool -i`:
```Lua
box.cfg({
    listen = "localhost:3013",
})
box.once("init", function()
    local s = box.schema.space.create('test', {
        id = 621,
        temporary = true,
        if_not_exists = true,
        field_count = 3,
        format = {
            {name = "id", type = "unsigned"},
            {name = "name", type = "string"},
            {name = "arr1", type = "array"},
        },
    })
    s:create_index('primary', {
        type = 'tree',
        unique = true,
        parts = {1, 'unsigned'},
        if_not_exists = true
    })
    s:create_index('secondary', {
        id = 5, type = 'tree',
        unique = false,
        parts = {2, 'string'},
        if_not_exists = true
    })

    box.schema.user.create('test', {password = 'test'})
    box.schema.user.grant('test', 'execute', 'universe')
    box.schema.user.grant('test', 'read,write', 'space', 'test')
end)
```
2. Run a Tarantool in the second console and execute:
```
tarantool> conn = require('net.box').connect("localhost:3013", {user = 'test', password = 'test', fetch_schema = false})
---
...

tarantool> conn.space.test.index.secondary:select()
---
- error: 'No index #1 is defined in space ''test'''
...

tarantool> conn.space.test.index[5]:select()
---
- []
...
```

**Actual behavior**

The server does not correctly map an index id by an index name passed with `IPROTO_INDEX_NAME`.

**Expected behavior**

The server selects data from an index with a name passed with `IPROTO_INDEX_NAME`.

rorororom:
iproto: fix IPROTO index name to ID mapping

Previously, index names were resolved to dense IDs, which caused
failures when accessing indexes by name. This happened because the
index_map lookup expected user-visible IDs (index->def->iid), not dense IDs.

The fix replaces dense_id with index->def->iid to ensure correct mapping
of index names to their numeric IDs, aligning with the expected behavior
when accessing indexes by name.

Closes #9923

### DIFF
--- a/changelogs/unreleased/gh-9923-fix-invalid-iproto-name-mapping-on-the-server-side.md
+++ b/changelogs/unreleased/gh-9923-fix-invalid-iproto-name-mapping-on-the-server-side.md
@@ -1,0 +1,3 @@
+## bugfix/core
+
+ * Fixed a bug when IPROTO_INDEX_NAME was mapped into a wrong index identifier (gh-9923).
--- a/test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
+++ b/test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
@@ -1,0 +1,74 @@
+local net = require('net.box')
+local server = require('luatest.server')
+local t = require('luatest')
+
+local g = t.group()
+
+g.before_all(function()
+    g.server = server:new({alias = 'master'})
+    g.server:start()
+    g.server:exec(function()
+        box.schema.create_space('test', {})
+
+        box.space.test:create_index('primary', {
+            parts = {1, 'unsigned'},
+        })
+
+        box.space.test:create_index('secondary', {
+            id = 5,
+            parts = {2, 'string'},
+        })
+    end)
+    -- fetch schema = false to resolve index on server instead of client
+    g.conn = net.connect(g.server.net_box_uri, {fetch_schema = false})
+end)
+
+g.after_all(function()
+    g.conn:close()
+    g.server:drop()
+end)
+
+g.before_each(function()
+    g.server:exec(function()
+        box.space.test:truncate()
+    end)
+end)
+
+g.test_index_select = function()
+    local space = g.conn.space.test
+
+    space:insert({1, 'F'})
+
+    local res1 = space.index.secondary:select('F')
+    local res2 = space.index[5]:select('F')
+
+    t.assert_equals(res1, {{1, 'F'}})
+    t.assert_equals(res1, res2)
+end
+
+g.test_index_delete = function()
+    local space = g.conn.space.test
+
+    space:insert({2, 'A'})
+
+    local deleted_by_name = space.index.secondary:delete('A')
+    t.assert_equals(deleted_by_name, {2, 'A'})
+
+    local check_deleted_by_name = space.index.secondary:get('A')
+    t.assert_is(check_deleted_by_name, nil)
+
+    local check_deleted_from_secondary = space:get(2)
+    t.assert_is(check_deleted_from_secondary, nil)
+end
+
+g.test_index_get = function()
+    local space = g.conn.space.test
+
+    space:insert({4, 'C'})
+
+    local fetched_by_name = space.index.secondary:get('C')
+    t.assert_equals(fetched_by_name, {4, 'C'})
+
+    local check_fetched_by_id = space:get(4)
+    t.assert_equals(check_fetched_by_id, {4, 'C'})
+end
--- a/src/box/iproto.cc
+++ b/src/box/iproto.cc
@@ -2462,7 +2462,7 @@ tx_resolve_space_and_index_name(struct request *dml)
 				 space->def->name);
 			return -1;
 		}
-		dml->index_id = idx->dense_id;
+		dml->index_id = idx->def->iid;
 	}
 	return 0;
 }


test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,45 @@
...
+        })
+
+        local idx_name_result = s.index.secondary:select()
+        t.assert_equals(#idx_name_result, 0)
+
+        local idx_id_result = s.index.primary:select()
+        t.assert_equals(#idx_id_result, 0)
+
+    end)
+end
```
drewdzzz, line 45: The test doesn't cover anything - it works fine on master. Please, check that the new tests actually fail before the fix is applied.

Also:
1. The commit fixes IPROTO resolver - here you use indexes locally.
2. The problem was caused by sparse index id. Here are no spare index ids.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,45 @@
...
+g.before_all(function()
+    g.server = server:new({alias = 'master'})
+    g.server:start()
+end)
+
+g.after_all(function()
+    g.server:stop()
+end)
+
+g.test_index_by_name = function()
```
drewdzzz, line 15: The helper that you fixed is used in `tx_process1` - let's cover it with a case as well. Using `space.index.secondary:delete` would be sufficient.

drewdzzz (changes requested):
Thanks for the patch! See my comments below.

Regarding commit message:

Header:
1. Do not use capital letter after the category name (`Correct` -> `correct`)
2. Let's replace word "correct" with "fix" - "correct" is often used as an adjective and "fix" is always a verb.

Body:
1. Please, do not use multiple spaces between sentences.
2. I would rephrase it in such way: explain the problem (that Tarantool resolved name to dense id but Tarantool requests actually expect id observed by user - `index->def->iid`) and after that something like "The commit fixed this problem". What's about the current message - it's still not clear why you replaced `dense_id` with `index->def->iid`.
3. "fail due to incorrect index ID resolution when fetch_schema was false" - did you check it? Does it really works with `fetch_schema = true`?

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.test_index_access_errors = function()
+    -- fetch schema = false to resolve index on server instead of client
+    local conn = net.connect(g.server.net_box_uri, {fetch_schema = false})
+    t.assert_is_not(conn, nil)
```
drewdzzz, line 52: Let's move connection creation to `before_each` helper. You can do it in this way:
```lua
g.conn = net.connect(...)
```

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+
+    local space = conn.space.test
+    t.assert_is_not(space, nil)
+
+    local res1 = space.index.secondary:select()
+    local res2 = space.index[5]:select()
+    t.assert_equals(res1, res2)
+
+    conn:close()
```
drewdzzz, line 61: Ditto - let's move it to `after_each`.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+g.test_index_delete = function()
+    local conn = net.connect(g.server.net_box_uri, {fetch_schema = false})
+    t.assert_is_not(conn, nil)
+
+    local space = conn.space.test
+    t.assert_is_not(space, nil)
+
+    space:insert({2, 'A', {1, 2, 3}})
+    space:insert({4, 'B', {4, 5, 6}})
+    space:insert({8, 'C', {7, 8, 9}})
```
drewdzzz, line 73: Since you insert some tuples for test, let's do `space:truncate` in `after_each` hook so that these insertions won't affect other cases.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+            if_not_exists = true,
+        })
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.test_index_access_errors = function()
```
drewdzzz, line 49: The second case checks access errors too. Let's rename this case to reflect that we check `select` here.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+    space:insert({4, 'B', {4, 5, 6}})
+    space:insert({8, 'C', {7, 8, 9}})
+
+    local deleted_by_name = space.index.unique_secondary:delete('A')
+    t.assert_equals(deleted_by_name, {2, 'A', {1, 2, 3}})
+
+    local check_deleted_by_name = space:get(2)
+    t.assert_is(check_deleted_by_name, nil)
+
+    local deleted_by_id = space.index[6]:delete('C')
```
drewdzzz, line 81: Why do we need deletion by id here?

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+            if_not_exists = true,
+        })
+        box.space.test:create_index('secondary', {
+            id = 5,
+            type = 'tree',
+            unique = false,
+            parts = {2, 'string'},
+            if_not_exists = true,
+        })
+        box.space.test:create_index('unique_secondary', {
```
drewdzzz, line 35: Why do you need both unique and non-unique indexes here? Let's just add `unique = true` to `secondary`.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+    cg.server:exec(function()
+        box.schema.create_space('test', {
+            id = 621,
+            temporary = true,
+            if_not_exists = true,
+            field_count = 3,
+            format = {
+                {name = "id", type = "unsigned"},
+                {name = "name", type = "string"},
+                {name = "arr1", type = "array"},
```
drewdzzz, line 19: Do you actually need this `arr1` field? Let's drop it to simplify the test.
Also, let's drop `field_count` and `temporary` attributes and explicit id of the space - they have nothing in common with bug we are fixing.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,88 @@
...
+local g = t.group()
+
+g.before_all(function(cg)
+    cg.server = server:new({alias = 'master'})
+    cg.server:start()
+    cg.server:exec(function()
+        box.schema.create_space('test', {
+            id = 621,
+            temporary = true,
+            if_not_exists = true,
```
drewdzzz, line 14: Since you create the space and indexes in `before_all`, you don't need the `if_not_exists` attribute. Let's drop all `if_not_exists` attributes.

drewdzzz (changes requested):
Thanks for the fixes!
I have no objections regarding the fix, commit message and changelog. Let's just slightly improve the test.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,72 @@
...
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.before_each(function(cg)
+    cg.server:exec(function()
+        box.schema.create_space('test', {
+            id = 621,
```
drewdzzz, line 19: Let's drop this id - it's not actually needed.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,72 @@
...
+            parts = {1, 'unsigned'},
+        })
+        box.space.test:create_index('secondary', {
+            id = 5,
+            type = 'tree',
+            unique = true,
+            parts = {2, 'string'},
+        })
+    end)
+    cg.conn = net.connect(cg.server.net_box_uri, {fetch_schema = false})
```
drewdzzz, line 37: Why did you drop the comment why we need `fetch_schema = false`?

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,72 @@
...
+    end)
+end)
+
+g.test_index_select = function(cg)
+    local space = cg.conn.space.test
+    t.assert_is_not(space, nil)
+
+    local res1 = space.index.secondary:select()
+    local res2 = space.index[5]:select()
+    t.assert_equals(res1, res2)
```
drewdzzz, line 56: Let's insert some tuples to the space along the way - there is no point to compare results since they are empty.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,72 @@
...
+    cg.server:start()
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)
+
+g.before_each(function(cg)
+    cg.server:exec(function()
+        box.schema.create_space('test', {
```
drewdzzz, line 18: Let's create the space and indexes in `before_all` - in `after_each` hook you can call `space:truncate()` to drop all tuples.

drewdzzz (changes requested):
Thanks for the fixes! Looks good, just a few nitpicks.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+    cg.server:drop()
+end
+
+g.before_each = function(cg)
+    cg.server:exec(function()
+        box.space.test:truncate()
+    end)
+end
+
+g.after_each = function(cg)
```
drewdzzz, line 47: These hooks are not actually executed because they are defined in this way:
```lua
g.after_each = function() ... end
```

Please, define them in this way:
```lua
g.after_each(function() ... end)
```

Applicable to all hooks: `before_all`, `after_all`, `before_each`, `after_each`.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+        box.space.test:create_index('secondary', {
+            id = 5,
+            type = 'tree',
+            unique = true,
+            parts = {2, 'string'},
+        })
+    end)
+
+    -- fetch schema = false to resolve index on server instead of client
+    cg.conn = net.connect(cg.server.net_box_uri, {fetch_schema = false})
```
drewdzzz, line 33: You should create the connection in `before_each` hook instead of `before_all` since you drop it in `after_each`. Since `after_each` is not actually executed (see another comment), it doesn't break anything, but it will after the fix.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+    cg.conn:close()
+end
+
+g.test_index_select = function(cg)
+    local space = cg.conn.space.test
+    t.assert_is_not(space, nil)
+
+    space:insert({1, 'F'})
+
+    local res1 = space.index.secondary:select()
```
drewdzzz, line 57: Let's do `index:select('F')` to check if select was actually done by a string field.

drewdzzz (changes requested):
Thanks for the fixes! Almost perfect, only a few remarks.

drewdzzz (approved):
Thanks for the patch, LGTM!

changelogs/unreleased/gh-9923-fix-invalid-iproto-name-mapping-on-the-server-side.md
```diff
@@ -0,0 +1,3 @@
+## bugfix/core
+
+ * Fixed a bug when IPROTO_INDEX_NAME was mapped incorrectly into index id (gh-9923).
```
CuriousGeorgiy, line 3: Nit:

```suggestion
 * Fixed a bug when IPROTO_INDEX_NAME was mapped into a wrong index identifier (gh-9923).
```

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+
+g.test_index_select = function()
+    local space = g.conn.space.test
+    t.assert_is_not(space, nil)
+
+    space:insert({1, 'F'})
+
+    local res1 = space.index.secondary:select('F')
+    local res2 = space.index[5]:select('F')
+    t.assert_equals(res1, res2)
```
CuriousGeorgiy, line 59: Please add another assert that one of the results is equal to `{{1, f}}`. Otherwise, `res1` and `res2` could both return wrong results and you would not catch it.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+end
+
+g.test_index_delete = function()
```
CuriousGeorgiy, line 62: And where is `test_index_get`? Actually, I believe it is enough to test only one method, since they all go through the same code path, but if you want to cover everything, then please cover everything.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+    local space = g.conn.space.test
+    t.assert_is_not(space, nil)
+
+    space:insert({2, 'A'})
+    space:insert({4, 'B'})
+    space:insert({8, 'C'})
```
CuriousGeorgiy, line 68: What's the point of inserting other values here? Please keep your tests concise and straight to the point, so that a reviewer and future reader of your test does not spend time guessing why this is needed.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+
+    local deleted_by_name = space.index.secondary:delete('A')
+    t.assert_equals(deleted_by_name, {2, 'A'})
+
+    local check_deleted_by_name = space:get(2)
+    t.assert_is(check_deleted_by_name, nil)
```
CuriousGeorgiy, line 74: Please also add a check for the `get` from the secondary index by name.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,75 @@
...
+end)
+
+g.after_all(function()
+    g.server:drop()
+end)
+
+g.before_each(function()
+    -- fetch schema = false to resolve index on server instead of client
+    g.conn = net.connect(g.server.net_box_uri, {fetch_schema = false})
+    t.assert_is_not(g.conn, nil)
```
CuriousGeorgiy, line 40: This check is redundant. Your connection can never be `nil`. If an error occurs during the connection construction, an exception is thrown. Since you do not specify the `wait_connected` option (which is totally fine for a test), you are by default blocked until the connection is established [^1], so you do not have to care about this:
> wait_connected: a connection timeout. By default, the connection is blocked until the connection is established, but if you specify wait_connected=false, the connection returns immediately. If you specify this timeout, it will wait before returning (wait_connected=1.5 makes it wait at most 1.5 seconds).

[^1]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#lua-function.net_box.connect

CuriousGeorgiy (approved):
Hey Maria, nice to meet you! Great first patch to start your Tarantool journey!

I have left some comments below regarding polishing of the test. These comments are mostly nitpicks, lifehacks, tricks and conventional Tarantool wisdom.

Please expect a lot of them in the beginning of your journey — they do not in any way diminish your hard skills, their goal is rather to calibrate you to our development traditions and expectations.

xuniq (approved):

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
CuriousGeorgiy: Look's like you forgot to push the change.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,87 @@
...
+    end)
+end)
+
+g.after_all(function()
+    g.server:drop()
+end)
+
+g.before_each(function()
+    -- fetch schema = false to resolve index on server instead of client
+    g.conn = net.connect(g.server.net_box_uri, {fetch_schema = false})
```
CuriousGeorgiy, line 39: Nit: do you really need to recreate the connection after each test? Maybe do it once in `before_all`/`after_all`?

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,87 @@
...
+
+g.before_all(function()
+    g.server = server:new({alias = 'master'})
+    g.server:start()
+    g.server:exec(function()
+        box.schema.create_space('test', {
+            format = {
+                {name = "id", type = "unsigned"},
+                {name = "name", type = "string"},
+            },
```
CuriousGeorgiy, line 15: Nit: do you actually rely on the space format in the test? I think you can safely drop this as redundant.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,87 @@
...
+        })
+
+        box.space.test:create_index('primary', {
+            type = 'tree',
+            unique = true,
```
CuriousGeorgiy, line 20: Nit: `type = 'tree'` and `unique = true` are defaults that everyone recognizes, so you can safely drop them as redundant.

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
```diff
@@ -0,0 +1,87 @@
...
+    local deleted_by_name = space.index.secondary:delete('A')
+    t.assert_equals(deleted_by_name, {2, 'A'})
+
+    local check_deleted_by_name = space:get(2)
+    t.assert_is(check_deleted_by_name, nil)
+end
+
+g.test_index_get = function()
+    local space = g.conn.space.test
+    t.assert_is_not(space, nil)
```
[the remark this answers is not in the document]
rorororom: fix

test/box-luatest/gh_9923_invalid_iproto_index_name_mapping_on_the_server_side_test.lua
rorororom: fix

sergepetrenko:
Backports:
3.3 https://github.com/tarantool/tarantool/pull/10982
3.2 https://github.com/tarantool/tarantool/pull/10983

Issue #9894, opened 2024-04-01
Put all `_func` options to `opts`


locker:
Historically, some function options (e.g. `is_deterministic`, `is_null_key`) are stored as tuple fields of the `_func` space while others (`is_multikey`, `takes_raw_args`) are stored as map entries in the `_func.opts` field. This looks confusing and ugly. Let's move all function options to `_func.opts`.

```yaml
tarantool> box.space._func:fselect()
---
- |-
  +-----+-----+---------+------+--------+---------+---------+---------+-------+---------+---------+---------+---------+---------+---------+-----+-------+---------+--------+
  | id  |owner|  name   |setuid|language|  body   |routine_t|param_lis|returns|aggregate|sql_data_|is_determ|is_sandbo|is_null_c| exports |opts |comment| created |last_alt|
  +-----+-----+---------+------+--------+---------+---------+---------+-------+---------+---------+---------+---------+---------+---------+-----+-------+---------+--------+
  |  1  |  1  |"box.sche|  0   | "LUA"  |   ""    |"function|   []    | "any" | "none"  | "none"  |  false  |  false  |  true   | ["LUA"] | {}  |  ""   |"2023-03-|"2023-03|
  | 65  |  1  |  "LUA"  |  0   | "LUA"  |"function|"function|["string"| "any" | "none"  | "none"  |  false  |  false  |  true   |["LUA","S| {}  |  ""   |"2023-03-|"2023-03|
  +-----+-----+---------+------+--------+---------+---------+---------+-------+---------+---------+---------+---------+---------+---------+-----+-------+---------+--------+
...
```

PR #10905, opened 2024-12-08
`box.ctl.on_replication_split_brain_rollback` event

### LINKED ISSUE #10460: Add a `box.ctl.on_replication_split_brain_rollback` event to make #10459 safe
In the scope of #10459 we are going to allow to automatically rollback asynchronous transactions from an old term. This can lead to data loss, so in order to make this feature safe, let's allow the user to intervene by adding a `box.ctl.on_replication_split_brain_rollback` event.

### LINKED ISSUE #11053: Record in `_gc_consumers` space is lost after PROMOTE followed by an instance restart
**Steps to reproduce**

1. Setup two replicas, replica 1 being master.
```lua
-- replica 1
console = require('console')

box.cfg{
    listen = 3301,
    replication = {3301, 3302},
    replication_timeout = 0.1,
    replication_synchro_timeout = 600,
    replication_synchro_quorum = 4,
}

box.once('bootstrap', function()

    box.schema.create_space('s', {is_sync = true}):create_index('p')

    box.ctl.promote()
end)

console.start()

os.exit()

-- replica 2
console = require('console')

box.cfg{
    listen = 3302,
    read_only = true,
    replication = {3301, 3302},
    replication_synchro_quorum = 4,
    replication_timeout = 0.1,
}

console.start()

os.exit()
```

2. Simulate network partitioning of replica 2 from replica 1 on replica 2.
```lua
-- replica 2

box.cfg{replication = ''}
```

3. Execute a synchronous transaction on replica 1 (master) while being partitioned from replica 2.
```lua
-- replica 1

box.atomic({wait = 'submit'}, function() box.space.s:replace{0} end)
```

4. Setup a third replica.
```lua
-- replica 3
console = require('console')

box.cfg{
    listen = 3303,
    read_only = true,
    replication = {3301, 3302},
    replication_synchro_quorum = 4,
    replication_timeout = 0.1,
}

console.start()

os.exit()
```

5. Check `_gc_consumers` space on replica 1 and observe new record for replica 3.
```lua
-- replica 1

box.space._gc_consumers:select{}
```

6. Promote replica 2.
```lua
-- replica 2

box.ctl.promote()
```

7. Wait until replica 1 receives the promote request from replica 2.

8. Restart replica 1.

9. Check `_gc_consumers` space on replica 1 and observe loss of record for replica 3.
```lua
box.space._gc_consumers:select{}
```

**Actual behavior**

The record in `_gc_consumers` space for replica 3 is lost after PROMOTE followed by an instance restart.

**Expected behavior**

The record in `_gc_consumers` space for replica 3 is not lost after PROMOTE followed by an instance restart.

CuriousGeorgiy:
This patch set adds a new `box.ctl.on_replication_split_brain_rollback` event that occurs when an asynchronously committed synchronous transaction gets rolled back by a PROMOTE request to prevent a split brain.

Closes #10460
Closes #11053

### DIFF
--- a/changelogs/unreleased/gh-10460-on_replication_split_brain_rollback-event.md
+++ b/changelogs/unreleased/gh-10460-on_replication_split_brain_rollback-event.md
@@ -1,0 +1,6 @@
+## feature/box
+
+* Added a new `box.ctl.on_replication_split_brain_rollback` event that occurs
+  when asynchronously committed transaction stored in the synchronous
+  transaction queue get rolled back by a PROMOTE request to prevent split brain
+  (gh-10460).
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1444,6 +1444,17 @@ txn_set_xrow_flags(uint8_t xrow_flags)
 	txn->flags |= flags_map[xrow_flags & IPROTO_FLAG_WAIT_ACK];
 }
 
+void
+txn_recovery_set_limbo_policy(void)
+{
+	struct txn *txn = in_txn();
+	if (!txn)
+		return;
+	if (!txn_has_flag(txn, TXN_WAIT_SYNC) &&
+	    !txn_has_flag(txn, TXN_WAIT_ACK))
+		txn_set_flags(txn, TXN_FORCE_ASYNC);
+}
+
 /** Wait for a linearization point for a transaction. */
 static int
 txn_make_linearizable(struct txn *txn)
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1421,6 +1421,12 @@ box_txn_set_timeout(double timeout)
 void
 box_txn_make_sync(void)
 {
+	txn_set_xrow_flags(IPROTO_FLAG_WAIT_ACK);
+}
+
+void
+txn_set_xrow_flags(uint8_t xrow_flags)
+{
 	struct txn *txn = in_txn();
 	/*
 	 * Do nothing if transaction is not started,
@@ -1429,7 +1435,13 @@ box_txn_make_sync(void)
 	if (!txn)
 		return;
 
-	txn_set_flags(txn, TXN_WAIT_ACK);
+	const uint8_t flags_map[] = {
+		[IPROTO_FLAG_WAIT_SYNC] = TXN_WAIT_SYNC,
+		[IPROTO_FLAG_WAIT_ACK] = TXN_WAIT_ACK,
+	};
+
+	txn->flags |= flags_map[xrow_flags & IPROTO_FLAG_WAIT_SYNC];
+	txn->flags |= flags_map[xrow_flags & IPROTO_FLAG_WAIT_ACK];
 }
 
 /** Wait for a linearization point for a transaction. */
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -38,8 +38,34 @@
 #include "tt_static.h"
 #include "session.h"
 #include "trivia/config.h"
+#include "func_adapter.h"
+#include "port.h"
+#include "tuple.h"
+#include "mpstream/mpstream.h"
 
 struct txn_limbo txn_limbo;
+
+/**
+ * Data used to create txn iterators.
+ */
+struct txn_limbo_entry_iterator_data {
+	/** First statement of the transaction. */
+	struct txn_stmt *first_stmt;
+};
+
+/**
+ * Iterator over transaction statements used in triggers.
+ */
+struct txn_limbo_entry_port_c_iterator {
+	/** Iterator next. */
+	port_c_iterator_next_f next;
+	/** Current statement of the transaction. */
+	struct txn_stmt *stmt;
+};
+
+static_assert(
+	sizeof(struct txn_limbo_entry_port_c_iterator) <= sizeof(struct port_c_iterator),
+	"The implementation should fit into abstract instance");
 
 static int
 txn_limbo_write_synchro(struct txn_limbo *limbo, uint16_t type, int64_t lsn,
@@ -134,6 +160,9 @@ txn_limbo_create(struct txn_limbo *limbo)
 		panic("failed to allocate synchronous queue worker fiber");
 	limbo->worker->f_arg = limbo;
 	fiber_set_joinable(limbo->worker, true);
+	limbo->on_split_brain_rollback =
+		event_get("box.ctl.on_replication_split_brain_rollback", true);
+	event_ref(limbo->on_split_brain_rollback);
 }
 
 void
@@ -153,6 +182,7 @@ txn_limbo_destroy(struct txn_limbo *limbo)
 		txn_free(entry->txn);
 	}
 	fiber_cond_destroy(&limbo->wait_cond);
+	event_unref(limbo->on_split_brain_rollback);
 	TRASH(limbo);
 }
 
@@ -1306,6 +1336,132 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
 	system_spaces_update_is_sync_state(is_promote);
 }
 
+static int
+txn_limbo_entry_iterator_next(struct port_c_iterator *it, struct port *out,
+			      bool *is_eof)
+{
+	struct txn_limbo_entry_port_c_iterator *entry_it =
+		(struct txn_limbo_entry_port_c_iterator *)it;
+	struct txn_stmt *stmt = entry_it->stmt;
+	if (stmt == NULL) {
+		*is_eof = true;
+		return 0;
+	}
+
+	*is_eof = false;
+	struct xrow_header *row = stmt->row;
+	port_c_create(out);
+	struct region *region = &fiber()->gc;
+	size_t region_svp = region_used(region);
+	struct mpstream stream;
+	mpstream_init(&stream, region, region_reserve_cb, region_alloc_cb,
+		      mpstream_panic_cb, NULL);
+	mpstream_encode_map(&stream, 2);
+	mpstream_encode_str(&stream, "meta");
+	mpstream_encode_map(&stream, 3);
+	mpstream_encode_str(&stream, "replica_id");
+	mpstream_encode_uint(&stream, row->replica_id);
+	mpstream_encode_str(&stream, "lsn");
+	mpstream_encode_uint(&stream, row->lsn);
+	mpstream_encode_str(&stream, "term");
+	mpstream_encode_uint(&stream, txn_limbo.promote_greatest_term);
+	mpstream_encode_str(&stream, "statement");
+	struct request request;
+	VERIFY(xrow_decode_dml(row, &request,
+			       dml_request_key_map(row->type)) == 0);
+	size_t map_size = 2 + (request.key != NULL) +
+				     (request.tuple != NULL) +
+				     (request.ops != NULL) +
+				     (stmt->old_tuple != NULL) +
+				     (stmt->new_tuple != NULL);
+	mpstream_encode_map(&stream, map_size);
+	mpstream_encode_str(&stream, "op_type");
+	mpstream_encode_str(&stream, iproto_type_name(row->type));
+	mpstream_encode_str(&stream, "space_id");
+	mpstream_encode_uint(&stream, space_id(stmt->space));
+#define encode_request_field(field_name)				    \
+do {									    \
+	if (request.field_name != NULL) {				    \
+		mpstream_encode_str(&stream, #field_name);		    \
+		size_t len = request.field_name##_end - request.field_name; \
+		mpstream_memcpy(&stream, request.field_name, len);	    \
+	}								    \
+}									    \
+while (0)
+	encode_request_field(key);
+	encode_request_field(tuple);
+	encode_request_field(ops);
+#undef encode_request_field
+#define encode_stmt_tuple(tuple_name)				\
+do {								\
+	if (stmt->tuple_name != NULL) {				\
+		mpstream_encode_str(&stream, #tuple_name);	\
+		tuple_to_mpstream(stmt->tuple_name, &stream);	\
+	}							\
+}								\
+while (0)
+	encode_stmt_tuple(old_tuple);
+	encode_stmt_tuple(new_tuple);
+#undef encode_stmt_tuple
+	mpstream_flush(&stream);
+	size_t obj_size = region_used(region) - region_svp;
+	char *obj = xregion_join(region, obj_size);
+	port_c_add_mp(out, obj, obj + obj_size);
+	entry_it->stmt = stailq_next_entry(stmt, next);
+	return 0;
+}
+
+static void
+txn_limbo_entry_iterator_create(void *base_data, struct port_c_iterator *it)
+{
+	struct txn_limbo_entry_port_c_iterator *entry_it =
+		(struct txn_limbo_entry_port_c_iterator *)it;
+	entry_it->next = txn_limbo_entry_iterator_next;
+	struct txn_limbo_entry_iterator_data *data = base_data;
+	entry_it->stmt = data->first_stmt;
+}
+
+static int
+on_split_brain_rollback_run_triggers(struct txn_limbo *limbo, int64_t lsn)
+{
+	struct txn_limbo_entry *last_rollback =
+		txn_limbo_last_rollback_entry(limbo, lsn);
+	if (last_rollback == NULL)
+		return 0;
+	struct txn_limbo_entry *e;
+	struct region *region = &fiber()->gc;
+	rlist_foreach_entry_reverse(e, &limbo->queue, in_queue) {
+		if (e->txn->fiber != NULL)
+			continue;
+		struct txn_limbo_entry_iterator_data data = {
+			.first_stmt = txn_first_stmt(e->txn),
+		};
+		struct port args;
+		port_c_create(&args);
+		port_c_add_iterable(&args, &data,
+				    txn_limbo_entry_iterator_create);
+		struct event_trigger_iterator it;
+		event_trigger_iterator_create(&it,
+					      limbo->on_split_brain_rollback);
+		struct func_adapter *trigger = NULL;
+		const char *unused;
+		while (event_trigger_iterator_next(&it, &trigger, &unused)) {
+			size_t region_svp = region_used(region);
+			int rc = func_adapter_call(trigger, &args, NULL);
+			region_truncate(region, region_svp);
+			if (rc != 0) {
+				diag_set(ClientError, ER_SPLIT_BRAIN,
+					 "box.ctl.on_split_brain_rollback "
+					 "trigger failed");
+				return -1;
+			}
+		}
+		if (e == last_rollback)
+			break;
+	}
+	return 0;
+}
+
 int
 txn_limbo_req_prepare(struct txn_limbo *limbo,
 		      const struct synchro_request *req)
@@ -1332,6 +1488,32 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
 	if (txn_limbo_filter_request(limbo, req) < 0) {
 		limbo->is_in_rollback = false;
 		return -1;
+	}
+	if (recovery_state == FINISHED_RECOVERY &&
+	    txn_limbo_is_owned_by_current_instance(limbo) &&
+	    /*
+	     * A demote request cannot rollback anything, and a rollback request
+	     * on the limbo owner can only be processed during recovery, so we
+	     * only care about promote requests.
+	     */
+	    req->type == IPROTO_RAFT_PROMOTE) {
+		struct txn *txn = txn_begin();
+		if (txn == NULL)
+			return -1;
+		/*
+		 * It is safe to allow the user to write to fully local spaces
+		 * during the execution of the trigger, bypassing the limbo. We
+		 * check that the transaction is fully local during preparation
+		 * for committing the transaction.
+		 */
+		txn_set_flags(txn, TXN_FORCE_ASYNC);
+		if (on_split_brain_rollback_run_triggers(limbo,
+							 req->lsn + 1) != 0 ||
+		    box_txn_commit() != 0) {
+			VERIFY(box_txn_rollback() == 0);
+			limbo->is_in_rollback = false;
+			return -1;
+		}
 	}
 	/* Prepare for request execution and fine-grained filtering. */
 	switch (req->type) {
--- a/src/box/txn_limbo.c
+++ b/src/box/txn_limbo.c
@@ -682,13 +682,12 @@ txn_limbo_write_rollback(struct txn_limbo *limbo, int64_t lsn)
 	limbo->is_in_rollback = false;
 }
 
-/** Rollback all the entries >= @a lsn. */
-static void
-txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
+static struct txn_limbo_entry *
+txn_limbo_last_rollback_entry(struct txn_limbo *limbo, int64_t lsn)
 {
 	assert(limbo->owner_id != REPLICA_ID_NIL || txn_limbo_is_empty(limbo));
 	assert(limbo == &txn_limbo);
-	struct txn_limbo_entry *e, *tmp;
+	struct txn_limbo_entry *e;
 	struct txn_limbo_entry *last_rollback = NULL;
 	rlist_foreach_entry_reverse(e, &limbo->queue, in_queue) {
 		if (!txn_has_flag(e->txn, TXN_WAIT_ACK))
@@ -697,8 +696,18 @@ txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
 			break;
 		last_rollback = e;
 	}
+	return last_rollback;
+}
+
+/** Rollback all the entries >= @a lsn. */
+static void
+txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
+{
+	struct txn_limbo_entry *last_rollback =
+		txn_limbo_last_rollback_entry(limbo, lsn);
 	if (last_rollback == NULL)
 		return;
+	struct txn_limbo_entry *e, *tmp;
 	rlist_foreach_entry_safe_reverse(e, &limbo->queue, in_queue, tmp) {
 		txn_limbo_abort(limbo, e);
 		txn_clear_flags(e->txn, TXN_WAIT_ACK);
--- a/test/replication-luatest/gh_11053_gc_consumer_record_lost_after_promote_and_restart_test.lua
+++ b/test/replication-luatest/gh_11053_gc_consumer_record_lost_after_promote_and_restart_test.lua
@@ -1,0 +1,80 @@
+local t = require('luatest')
+local replica_set = require('luatest.replica_set')
+local server = require('luatest.server')
+
+local g = t.group()
+
+g.before_each(function(cg)
+    cg.replica_set = replica_set:new{}
+    cg.box_cfg = {
+        replication = {
+            server.build_listen_uri('master', cg.replica_set.id),
+            server.build_listen_uri('replica', cg.replica_set.id),
+        },
+        replication_timeout = 0.1,
+        replication_synchro_timeout = 120,
+        replication_synchro_quorum = 4,
+    }
+    cg.master = cg.replica_set:build_and_add_server{
+        alias = 'master',
+        box_cfg = cg.box_cfg,
+    }
+    cg.replica = cg.replica_set:build_and_add_server{
+        alias = 'replica',
+        box_cfg = cg.box_cfg,
+    }
+    cg.replica_set:start()
+    cg.replica_set:wait_for_fullmesh()
+    cg.master:exec(function()
+        box.schema.space.create('s', {is_sync = true}):create_index('p')
+        box.ctl.promote()
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+end)
+
+g.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+g.test_gc_consumer_record_not_lost_after_promote_and_restart = function(cg)
+    -- Simulate network partitioning.
+    cg.replica:update_box_cfg{replication = ''}
+    -- Submit a synchronous transaction while being partitioned from replica.
+    cg.master:exec(function()
+        box.atomic({wait = 'submit'}, function() box.space.s:replace{0} end)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+        t.assert_equals(box.space._gc_consumers:len(), 1)
+    end)
+    -- Join a new replica.
+    cg.new_replica = cg.replica_set:build_and_add_server{
+        alias = 'new_replica',
+        box_cfg = cg.box_cfg,
+    }
+    -- The join cannot be processed completely because we have a pending
+    -- transaction in the synchronous queue.
+    cg.new_replica:start({wait_until_ready = false})
+    -- Wait for the new replica's `_gc_consumer` record to appear on the master.
+    cg.master:exec(function()
+        t.helpers.retrying({timeout = 120}, function()
+            t.assert_equals(box.space._gc_consumers:len(), 2)
+        end)
+    end)
+    -- Promote the partitioned replica.
+    cg.replica:exec(function()
+        box.ctl.promote()
+    end)
+    -- Wait for the old master to receive promote from the partitioned replica.
+    cg.master:exec(function(replica_id)
+        t.helpers.retrying({timeout = 120}, function()
+            t.assert_equals(box.info.synchro.queue.owner, replica_id)
+        end)
+        t.assert_equals(box.info.synchro.queue.len, 0)
+    end, {cg.replica:get_instance_id()})
+    -- Restart the old master.
+    cg.master:restart()
+    -- Check that the new replica's `_gc_consumer` record is still present on
+    -- the old master.
+    cg.master:exec(function()
+        t.assert_equals(box.space._gc_consumers:len(), 2)
+    end)
+end
--- a/test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
+++ b/test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
@@ -1,0 +1,503 @@
+local t = require('luatest')
+local server = require('luatest.server')
+local replica_set = require('luatest.replica_set')
+
+local g = t.group()
+
+g.before_each(function(cg)
+    cg.replica_set = replica_set:new{}
+    cg.box_cfg = {
+        replication = {
+            server.build_listen_uri('master', cg.replica_set.id),
+            server.build_listen_uri('replica', cg.replica_set.id),
+        },
+        replication_timeout = 0.1,
+        replication_synchro_timeout = 120,
+        replication_synchro_quorum = 3,
+    }
+    cg.master = cg.replica_set:build_and_add_server{
+        alias = 'master',
+        box_cfg = cg.box_cfg,
+    }
+    cg.replica = cg.replica_set:build_and_add_server{
+        alias = 'replica',
+        box_cfg = cg.box_cfg,
+    }
+    cg.replica_set:start()
+    cg.replica_set:wait_for_fullmesh()
+    cg.master:exec(function()
+        rawset(_G, 'trigger', require('trigger'))
+        rawset(_G, 'trigger_ok', false)
+        local event = 'box.ctl.on_replication_split_brain_rollback'
+        local body = 'function() rawset(_G, "trigger_called", true) end'
+        box.schema.func.create('f', {body = body, trigger = event})
+        box.schema.space.create('a', {is_sync = false}):create_index('p')
+        box.schema.space.create('s', {is_sync = true}):create_index('p')
+        box.schema.space.create('l', {is_local = true}):create_index('p')
+        box.ctl.promote()
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
+end)
+
+g.after_each(function(cg)
+    cg.replica_set:drop()
+end)
+
+local function partition_server(s)
+    s:exec(function()
+        box.cfg{replication = ''}
+    end)
+end
+
+local function cause_split_brain(master, replica)
+    partition_server(replica)
+    master:exec(function()
+        box.atomic({wait = 'submit'}, function() box.space.s:replace{0} end)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+    end)
+    replica:exec(function()
+        box.ctl.promote()
+    end)
+end
+
+local function check_promote_from_partitioned_server(partitioned, master,
+                                                     replication)
+    partitioned:exec(function()
+        box.ctl.promote()
+    end)
+    master:exec(function()
+        t.helpers.retrying({timeout = 120}, function()
+            t.assert_equals(box.info.synchro.queue.len, 0)
+        end)
+        t.assert(_G.trigger_ok)
+    end)
+    partitioned:exec(function(replication)
+        box.cfg{replication = replication}
+    end, {replication})
+    master:exec(function()
+        box.ctl.promote()
+    end)
+    master:wait_for_downstream_to(partitioned)
+end
+
+local function test_stmt(cg, stmt)
+    partition_server(cg.replica)
+    cg.master:exec(function(stmt)
+        box.atomic({wait = 'submit'}, function()
+            box.space.s[stmt.request](box.space.s, stmt.arg, stmt.ops)
+        end)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+        local event = 'box.ctl.on_replication_split_brain_rollback'
+        _G.trigger.set(event, 't', _G.generate_trigger_test_case(stmt))
+    end, {stmt})
+    check_promote_from_partitioned_server(cg.replica, cg.master,
+                                          cg.box_cfg.replication)
+end
+
+local function test_several_stmts(cg)
+    partition_server(cg.replica)
+    cg.master:exec(function()
+        box.atomic({wait = 'submit'}, function()
+            box.space.s:replace{0}
+            box.space.s:replace{0}
+        end)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+        local event = 'box.ctl.on_replication_split_brain_rollback'
+        local stmt = {request = 'replace', arg = {0}, ops = nil,
+                      old_tuple = nil, new_tuple = {0}}
+        _G.trigger.set(event, 't', _G.generate_trigger_test_case(stmt, 2))
+    end)
+    check_promote_from_partitioned_server(cg.replica, cg.master,
+                                          cg.box_cfg.replication)
+end
+
+g.before_test('test_different_stmts', function(cg)
+    cg.master:exec(function()
+        rawset(_G, 'generate_trigger_test_case',
+            function(stmt, n_stmts)
+                return function(it)
+                    n_stmts = n_stmts ~= nil and n_stmts or 1
+                    local counter = 1
+                    for obj in it() do
+                        _G.trigger_ok = pcall(function()
+                            local meta = obj.meta
+                            local rid = meta.replica_id
+                            local lsn = meta.lsn
+                            local term = meta.term
+                            local sid = obj.statement.space_id
+                            local op_type = obj.statement.op_type
+                            local arg
+                            if stmt.request == "replace" or
+                               stmt.request == "insert" or
+                               stmt.request == "upsert" then
+                                arg = obj.statement.tuple
+                                t.assert_equals(obj.statement.key, nil)
+                            else
+                                arg = obj.statement.key
+                            end
+                            local ops
+                            if stmt.request == "update" then
+                                ops = obj.statement.tuple
+                                t.assert_equals(obj.statement.ops, nil)
+                            else
+                                ops = obj.statement.ops
+                            end
+                            local old_tuple = obj.statement.old_tuple
+                            local new_tuple = obj.statement.new_tuple
+
+                            t.assert_equals(term, box.info.synchro.queue.term)
+                            t.assert_equals(rid, box.info.id)
+                            if counter == n_stmts then
+                               t.assert_equals(lsn, box.info.lsn)
+                            end
+                            t.assert_equals(sid, box.space.s.id)
+                            t.assert_equals(op_type, stmt.request:upper())
+                            t.assert_equals(arg, stmt.arg)
+                            if stmt.ops ~= nil then
+                                t.assert_equals(ops, stmt.ops)
+                            end
+                            if counter == 1 then
+                                t.assert_equals(old_tuple, stmt.old_tuple)
+                            else
+                                t.assert_equals(old_tuple, arg)
+                            end
+                            t.assert_equals(new_tuple, stmt.new_tuple)
+                            counter = counter + 1
+                        end)
+                    end
+                    t.assert_equals(n_stmts, counter - 1)
+                end
+            end)
+        box.cfg{replication_synchro_quorum = 'N/2 + 1'}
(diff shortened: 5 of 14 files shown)


Gerold103 (commented):
Thanks for the patch! While the coding work is solid, I suggest we step a bit back, and see the thing from afar, trying to address a few design concerns.

src/box/txn.c
```diff
@@ -978,6 +978,17 @@ txn_journal_entry_new(struct txn *txn)
...
+		assert(txn_limbo.is_in_rollback);
+		if (!txn_is_fully_local(txn)) {
+			diag_set(IllegalParams,
+				 "Transactions from the "
+				 "box.ctl.on_replication_split_brain_rollback "
+				 "event trigger must be fully local");
+			return NULL;
+		}
+		txn_set_flags(txn, TXN_FORCE_ASYNC);
+	}
```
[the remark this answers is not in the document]
CuriousGeorgiy: > Given the uncertainty and depth of the problem I would say this patch needs a design before implementation. Is there a document I could read? Otherwise I suggest we make one, to discuss it.

AFAIC, this patch was already designed and discussed in the scope of the repair queue theme: https://www.notion.so/tarantool/WIP-Automatic-handling-of-async-transactions-from-old-terms-in-CE-and-EE-1882542d61a54e69b41c74f96436ad34?pvs=4#065efdadd8bd4ea6a91c13ef9329230f (didn't have time yet to move it to Confluence).

I agree that the requirement to run an arbitrary number of transactions seems too generic and unnecessary. I would prefer having only one transaction available as you described in `Alternative 2`. I don't think that always starting a transaction will introduce a noticeable overhead — moreover, this code is not the critical path.

However, the behavior will be less transparent and case-specific compared to the current solution.

src/box/txn_limbo.c
```diff
@@ -1520,11 +1698,16 @@ void
 txn_limbo_init(void)
 {
 	txn_limbo_create(&txn_limbo);
+	on_split_brain_rollback =
+		event_get("box.ctl.on_replication_split_brain_rollback", true);
+	event_ref(on_split_brain_rollback);
```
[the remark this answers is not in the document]
CuriousGeorgiy: I was just following the current naming conventions. This was also discussed in the RFC.

src/box/txn_limbo.c
```diff
@@ -1331,6 +1493,22 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
 		limbo->is_in_rollback = false;
 		return -1;
 	}
+	limbo->is_running_on_split_brain_rollback_triggers = true;
+	if (recovery_state == FINISHED_RECOVERY &&
+	    txn_limbo_is_owned_by_current_instance(limbo) &&
+	    /*
+	     * A demote request cannot rollback anything, and a rollback request
```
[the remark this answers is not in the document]
CuriousGeorgiy: Because it can't rollback anything from the current instance's synchronous queue, just as PROMOTE cannot.

src/box/txn.c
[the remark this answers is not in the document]
CuriousGeorgiy: /cc @sergepetrenko @Serpentian

src/box/txn_limbo.c
```diff
@@ -1304,6 +1338,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+	 *  3. The ID of the transaction;
+	 *  4. The ID of the limbo owner or 0 for local rows that are part of a
+	 *  global transaction;
+	 *  5. The LSN of the request.
+	 *  6. The ID of the space to which the request was applied.
+	 *  7. The request type.
+	 *  8. The argument of the request depending on the operation type
+	 *  9. An optional argument for the UPDATE/UPSERT request, may be nil.
+	 *  10. The tuple which was replaced by the operation, may be nil.
+	 *  11. The tuple created by the operation, may be nil.
```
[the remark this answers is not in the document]
CuriousGeorgiy: This design mirrors the design of `on_commit`/`on_rollback` iterators [^1]:
> The iterator has:
> 
> * an ordinal request number
> * the old value of the tuple before the request (nil for an insert request)
> * the new value of the tuple after the request (nil for a delete request)
> * the ID of the space

[^1]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_txn_management/on_commit/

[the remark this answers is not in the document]
CuriousGeorgiy: /cc @sergepetrenko @Serpentian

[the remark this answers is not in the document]
Gerold103: Yeah, but in case of on_commit/rollback it is 4 values, while in the new API it is 11 values. This is quite a big difference. We can't expose this kind of multireturn, it is too complex. Even in case of on_commit/rollback I would now say it should have been an object, because now we can't add stuff to it (I know that I did those triggers, but I didn't know any better back then 😭).

src/box/txn.c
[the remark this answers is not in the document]
Gerold103: > I don't think that's really a problem. Why not allow everyone to run fully local transactions during this time?

Because it would mean a bunch of TXNs would slip past the limbo, without being related to the limbo state, but abusing it to ignore the quorum.

Alternatively we could say, that if the limbo is in the rollback state, then it could be treated as "empty", which allows local and async txns to get committed without a quorum indeed. The problem though is what happens if the limbo rollback fails, and some stuff unrelated to the limbo problems was already committed?

That is very edgy. Thanks for the design link btw! Somehow I overlooked in that document, that it would be so intrusive in the code. Specifically the highlighted code chunk that we are discussing. It is a big signal we are doing something wrong, when such an intrusion becomes necessary.

src/box/txn_limbo.c
```diff
@@ -682,13 +682,10 @@ txn_limbo_write_rollback(struct txn_limbo *limbo, int64_t lsn)
...
 }
 
-/** Rollback all the entries >= @a lsn. */
-static void
-txn_limbo_read_rollback(struct txn_limbo *limbo, int64_t lsn)
+static struct txn_limbo_entry *
+txn_limbo_last_rollback_entry(struct txn_limbo *limbo, int64_t lsn)
 {
-	assert(limbo->owner_id != REPLICA_ID_NIL || txn_limbo_is_empty(limbo));
-	assert(limbo == &txn_limbo);
```
sergepetrenko, line 690: nit: why don't you want to leave these assertions here?

src/box/txn_limbo.c
```diff
@@ -1304,6 +1338,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+static int
+on_split_brain_rollback_run_triggers(struct txn_limbo *limbo, int64_t lsn)
+{
+	struct txn_limbo_entry *last_rollback =
+		txn_limbo_last_rollback_entry(limbo, lsn);
+	if (last_rollback == NULL)
+		return 0;
+	struct txn_limbo_entry *e;
+	rlist_foreach_entry_reverse(e, &limbo->queue, in_queue) {
+		if (!txn_has_flag(e->txn, TXN_WAIT_ACK) ||
```
sergepetrenko, line 1441: What about the `TXN_WAIT_SYNC` transactions? It seems we could put them into the repair queue as well. Would this cause some inconsistency?

src/box/txn.c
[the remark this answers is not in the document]
sergepetrenko: > **Alternative 2**
> 
> We start a transaction ourselves unconditionally, force-async. Then user's callback gets executed. The user knows it is already having a txn, and the user's code is supposed to just do whatever operations they need. Then the callback is done, and we try to commit the txn if it is not empty. If the user rolls it back and starts another one inside of the callback, then it won't be force-async, and won't be able to break anything.
> 
> It seems better than the previous solution in a sense that the user doesn't have to leave an open txn hanging. But it is worse in a sense that the user has no choice whether a txn is needed.

I like this variant. It will suit our needs with repair queue just fine and won't introduce any crutches like the current approach. Besides, this trigger is rather low-level, it describes a rarely needed state (similar to `box.ctl.on_schema_init`), so I think this is fine that its semantics differ a bit from other transaction triggers. We could even name this event "internal" to emphasize that it's not intended for general use (although we won't forbid such use, of course).

Besides, we could forbid user-called `box.commit()/box.rollback()` from the trigger similarly to how it was done in de80e0264f7deb58ea86ef85b37b92653a803430

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: Снимок экрана 2024-12-17 в 12 34 27

Just for the record, here's the format of repair_queue space we've agreed upon. Only `term` is missing, and it's needed, obviously.

From what I can see, `id` and `transaction_id` may be calculated by the trigger itself, no need to pass them,
replica_id, lsn, space_id, operation_type, key/tuple, ops_array, old_tuple, new_tuple may all be passed in a format similar to xlog iterator. Here's the output of `xlog.pairs` when `wal_ext` is enabled:
```
tarantool> for k,v in xlog.pairs('00000000000000000000.xlog') do json = require'json' print(json.encode(k)) print(json.encode(v)) end
1
{"HEADER":{"lsn":1,"replica_id":1,"type":"INSERT","timestamp":1734430078.2496},"BODY":{"space_id":280,"new_tuple":[512,1,"test","memtx",0,{},[]],"tuple":[512,1,"test","memtx",0,{},[]]}}
2
{"HEADER":{"lsn":2,"replica_id":1,"type":"INSERT","timestamp":1734430082.6448},"BODY":{"space_id":288,"new_tuple":[512,0,"pk","tree",{"unique":true},[[0,"unsigned"]]],"tuple":[512,0,"pk","tree",{"unique":true},[[0,"unsigned"]]]}}
3
{"HEADER":{"lsn":3,"replica_id":1,"type":"INSERT","timestamp":1734430089.2591},"BODY":{"space_id":512,"new_tuple":[1,"asdfb"],"tuple":[1,"asdfb"]}}
4
{"HEADER":{"lsn":4,"replica_id":1,"type":"INSERT","timestamp":1734430094.6888},"BODY":{"space_id":512,"new_tuple":[2,"bsdfa"],"tuple":[2,"bsdfa"]}}
5
{"HEADER":{"lsn":5,"replica_id":1,"type":"REPLACE","timestamp":1734430098.0942},"BODY":{"space_id":512,"old_tuple":[2,"bsdfa"],"new_tuple":[2],"tuple":[2]}}
6
{"HEADER":{"lsn":6,"replica_id":1,"type":"UPDATE","timestamp":1734430107.2436},"BODY":{"new_tuple":[1,"asdfb","bababa"],"old_tuple":[1,"asdfb"],"space_id":512,"index_base":1,"key":[1],"tuple":[["=",3,"bababa"]]}}
---
...
```

So I propose the following format:
```yaml
obj = {
    HEADER = {
        lsn = ...,
        replica_id = ...,
        type = ...,
        term = ...
    },
    BODY = {
        space_id = ...,
        key = ...,
        tuple = ...,
        old_tuple = ...,
        new_tuple = ...,
    }
}
```

src/box/txn.c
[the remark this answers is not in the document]
Gerold103: I would still recommend to reconsider if this should be an event or a trigger. Shame on me I didn't pay enough attention to it in the document 😣. Events semantics is to notify after the fact. Not to change or prevent or piggyback something. The semantics we need here is clearly a trigger - something that can affect the outcome.

For instance, of `box.status` event callback fails, does it prevent the status from changing? Nope. Then why reacting to this `box.ctl.thisnewlongname` event should change the state of anything?

Another example - `space:on_replace` is not an event, it is a trigger, because it can prevent the replace or add more statements to the same txn. Certainly not an event, right?

Does this make sense?

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: I am super super sorry 🙏, but how about we also change the repair queue format to store rows as objects? And only put into the space fields the things like `id`? The indexes anyway can be by json paths, if we need to index here anything. Otherwise I am really scared for extendibility of the format.

Speaking of the iteration: do we really want to make it so binary like `HEADER` and `BODY`? We are not iterating over a file here. And it doesn't reflect the binary structure anyway, because we don't store `term` in xrows, but above it looks like stored in the header.

But I like the rows being objects much more than multireturns 🔥! A couple of suggestions based on the things I wrote above:
- Perhaps flatten the object? So the header and body keys are all in one level? The motivation is what I said about the iterator being over txn statements, not xlogs.
- If not flatten, then perhaps add the term at the same level as `HEADER` and `BODY` keys? Otherwise it isn't what xlog iterator would return.

Thanks for the feedback and the discussion, @sergepetrenko and @CuriousGeorgiy , and apologies once again that I didn't realize those things during the RFC review. I hope not much would need change in the code.

src/box/txn.c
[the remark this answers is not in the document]
sergepetrenko: It seems there is some misunderstanding here. From what I see it is implemented like a trigger. So is there a problem with naming or something else?

Sure thing `space:on_replace` is a trigger, but it can be set via
```lua
local trigger = require('trigger')
trigger.set('box.space[512].on_replace', 'my_name.on_replace_trigger', function(...) ... end)
```
And here we refer to `box.space[512].on_replace` as an event upon which a trigger is fired. At least I found this terminology  here: https://github.com/tarantool/doc/issues/3988

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: > I am super super sorry 🙏, but how about we also change the repair queue format to store rows as objects? And only put into the space fields the things like id? The indexes anyway can be by json paths, if we need to index here anything. Otherwise I am really scared for extendibility of the format.

I agree, this makes sense not to introduce huge space formats with lots of fields. We even have this ticket regarding a similar problem in space `_func`: https://github.com/tarantool/tarantool/issues/9894

I suggest the following format then:
`[id : unsigned, transaction_id : unsigned, row_meta : map, statement : map]`
`row_meta` would contain `replica_id`, `lsn`, `term`.
`statement` would contain the rest: `op_type`, `space_id`, `key`, `tuple`, `old_tuple`, `new_tuple`.

> Perhaps flatten the object? So the header and body keys are all in one level? The motivation is what I said about the iterator being over txn statements, not xlogs.

I'd rather not mix row meta and operation details. It looks strange to pile everything in one level IMO.

> If not flatten, then perhaps add the term at the same level as HEADER and BODY keys? Otherwise it isn't what xlog iterator would return.

Maybe make the  object format similar to space format I propose, then? In this case it would return objects almost in the same format in which they will be stored in the space (`id` and `transaction_id` will be missing, since we agreed the user could calculate them himself). Does it make sense?
```
obj = {
    row_meta = {
        replica_id = ...,
        lsn = ...,
        term = ...,
    },
    statement = {
        op_type = ...,
        space_id = ...,
        key = ...,
        tuple = ...,
        old_tuple = ...,
        new_tuple = ...,
    }
}
```

lenkis (approved):

src/box/txn.c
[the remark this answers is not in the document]
Gerold103: Aaaaaah, omg, I was looking at `struct event *on_split_brain_rollback` and was thinking why is it an event like for `box.broadcast()`! Sorry, didn't notice it, then all is ok!

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: Thanks for considering it! Yes, looks way better. I would even suggest to shorten `row_meta` to just `meta`, otherwise top 💪.

src/box/txn_limbo.c
CuriousGeorgiy: Thanks for noticing, this change was not intended. Returned the assertions.

src/box/txn_limbo.c
```diff
@@ -1304,6 +1338,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+	 *  5. The LSN of the request.
+	 *  6. The ID of the space to which the request was applied.
+	 *  7. The request type.
+	 *  8. The argument of the request depending on the operation type
+	 *  9. An optional argument for the UPDATE/UPSERT request, may be nil.
+	 *  10. The tuple which was replaced by the operation, may be nil.
+	 *  11. The tuple created by the operation, may be nil.
+	 */
+	port_c_add_number(out, entry_it->req_num++);
+	port_c_add_number(out, (uint32_t)txn_limbo.promote_greatest_term);
```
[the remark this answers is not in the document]
CuriousGeorgiy: Thanks for noticing, this as a typo — changed the cast to `double`.

test/replication-luatest/gh_10460_on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,430 @@
+local t = require('luatest')
```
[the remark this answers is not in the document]
CuriousGeorgiy: Sorry, we used to name all tests this way in core team — I will start following this style from now. Fixed.

src/box/txn_limbo.c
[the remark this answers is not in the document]
CuriousGeorgiy: I see, thanks for pointing this out — fixed.

changelogs/unreleased/gh-10460-on_replication_split_brain_rollback-event.md
```diff
@@ -0,0 +1,5 @@
+## feature/box
+
+* Added a new `box.ctl.on_replication_split_brain_rollback` event that occurs
```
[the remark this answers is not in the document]
CuriousGeorgiy: I was referring to how individual entries are pushed onto the stack:
https://github.com/tarantool/tarantool/blob/ec65c54af009086b6e2fd603cc2095ce3587b220/src/box/lua/xlog.c#L123-L140

src/box/txn_limbo.c
CuriousGeorgiy: I agree, let's enable this for all asynchronously committed transactions stored in the limbo.

src/box/txn.c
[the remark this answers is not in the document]
Serpentian: I'm ok with the second variant too

src/box/txn_limbo.c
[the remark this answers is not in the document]
Serpentian: Agree with the solution, Sergey proposed

src/box/txn.c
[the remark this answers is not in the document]
CuriousGeorgiy: Implemented second alternative, as we have discussed.

src/box/txn_limbo.c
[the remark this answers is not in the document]
CuriousGeorgiy: > I suggest the following format then:
[id : unsigned, transaction_id : unsigned, row_meta : map, statement : map]

This is out of the scope of this ticket and PR and is more related to the repair queue, but since we started discussing here: AFAIU, transaction identifiers are not stable (i.e., they are affected by snapshots) — I suggest the transaction identifier you are referring  to here should be a monotonically increasing value in this space, rather than `txn->id`.

Implemented @sergepetrenko's variant (`row_meta` -> `meta`) using a MsgPack object.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,430 @@
...
+        rawset(_G, 'trigger_ok', false)
+        local event = 'box.ctl.on_replication_split_brain_rollback'
+        local body = 'function() rawset(_G, "trigger_called", true) end'
+        box.schema.func.create('f', {body = body, trigger = event})
+        box.schema.space.create('a', {is_sync = false}):create_index('p')
+        box.schema.space.create('s', {is_sync = true}):create_index('p')
+        box.schema.space.create('l', {is_local = true}):create_index('p')
+        box.ctl.promote()
+    end)
+    cg.master:wait_for_downstream_to(cg.replica)
```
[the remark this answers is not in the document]
CuriousGeorgiy: Sorry, I forgot to remove your assignment, and you were looking at a WIP version. Should be fine now — does not flake on my machine.

CuriousGeorgiy:
Checkpatch output seems to FP for my case:
```console
ERROR:MACRO_ARG_PRECEDENCE: Macro argument 'field_name' may be better as '(field_name)' to avoid precedence issues
#217: FILE: src/box/txn_limbo.c:1383:
+#define encode_request_field(field_name)				    \
+do {									    \
+	if (request.field_name != NULL) {				    \
+		mpstream_encode_str(&stream, #field_name);		    \
+		size_t len = request.field_name##_end - request.field_name; \
+		char *buf = mpstream_reserve(&stream, len);		    \
+		memcpy(buf, request.field_name, len);			    \
+		mpstream_advance(&stream, len);				    \
+	}								    \
+}									    \
+while (0)
```

I couldn't figure out how to make it happy.

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: > AFAIU, transaction identifiers are not stable (i.e., they are affected by snapshots) — I suggest the transaction identifier you are referring to here should be a monotonically increasing value in this space, rather than txn->id.

Yes, indeed.

src/box/txn.c
```diff
@@ -1416,6 +1416,12 @@ box_txn_set_timeout(double timeout)
 
 void
 box_txn_make_sync(void)
+{
+	box_txn_set_xrow_flags(IPROTO_FLAG_WAIT_ACK);
+}
+
+void
+box_txn_set_xrow_flags(uint8_t xrow_flags)
```
Gerold103, line 1431: The function `box_txn_make_sync()` has `box_` prefix because it is public, available in module.h. For internal txn methods lets better use just `txn_`, without `box_`. It would then be `txn_set_xrow_flags()`. Does that make sense?

src/box/txn.c
```diff
@@ -995,7 +995,8 @@ txn_journal_entry_new(struct txn *txn)
 				return NULL;
 			}
 			txn_set_flags(txn, TXN_WAIT_SYNC | TXN_WAIT_ACK);
-		} else if (!txn_limbo_is_empty(&txn_limbo)) {
+		} else if (recovery_state == FINISHED_RECOVERY &&
+			   !txn_limbo_is_empty(&txn_limbo)) {
```
Gerold103, line 999: > During recovery and replication, the `TXN_WAIT_SYNC` must not be set for such transactions.

Why isn't `TXN_FORCE_ASYNC` enough? I don't get this part. No matter recovery or not, if a txn didn't have these flags during WAL write, then it won't get them. Otherwise if it had them, then they should stay. There doesn't seem to be any need to clear this flag on recovery, if it was present before recovery.

src/box/txn_limbo.c
Gerold103, line 690: In response to Sergey you said you returned the assertions. But you moved them to `txn_limbo_read_rollback()`. Why? Is there a case where they aren't true for `txn_limbo_last_rollback_entry()`?

src/box/txn_limbo.c
```diff
@@ -38,9 +38,36 @@
 #include "tt_static.h"
 #include "session.h"
 #include "trivia/config.h"
+#include "func_adapter.h"
+#include "port.h"
+#include "tuple.h"
+#include "mpstream/mpstream.h"
+#include "lua/tuple.h"
```
Gerold103, line 45: Hm. Making limbo depend on Lua is a bold move 😁. And we absolutely can't do that, unfortunately or not. Lets take care of the dependencies, so they don't get bloated and entangled too much. The future-us will say "thanks" to the past-us.

I see that you use it for `tuple_to_mpstream()` function. But it doesn't seem to depend on Lua at all. I suggest to follow the Boy-Scout Rule ("leave code better than it was before you"), and move this function to box/tuple.h(.c).

src/box/txn_limbo.c
```diff
@@ -1306,6 +1337,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+	}								\
+}									\
+while (0)
+	encode_stmt_tuple(old_tuple);
+	encode_stmt_tuple(new_tuple);
+#undef encode_stmt_tuple
+	mpstream_flush(&stream);
+	size_t obj_size = region_used(region) - region_svp;
+	char *obj = xregion_join(region, obj_size);
+	port_c_add_mp_object(out, obj, obj + obj_size, NULL);
```
Gerold103, line 1412: Why using an mp object instead of just `port_c_add_mp()`? People won't be able to insert this directly to any space (it is a map, not an array). Which means they will have to decode it anyway, to get the individual fields, check if it wasn't inserted already (the callback can be called more than once for the same txns if PROMOTE fails).

Is there a specific reason we use mpobject here?

src/box/txn_limbo.c
```diff
@@ -1306,6 +1337,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+		char *buf = mpstream_reserve(&stream, len);		    \
+		memcpy(buf, request.field_name, len);			    \
+		mpstream_advance(&stream, len);				    \
+	}								    \
+}									    \
+while (0)
+	encode_request_field(key);
+	encode_request_field(tuple);
+	encode_request_field(ops);
+#undef encode_request_part
```
Gerold103, line 1397: `encode_request_part` is not defined anywhere. Did you mean `encode_request_field`?

src/box/txn_limbo.c
```diff
@@ -1306,6 +1337,134 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+	mpstream_encode_str(&stream, "space_id");
+	mpstream_encode_uint(&stream, space_id(stmt->space));
+#define encode_request_field(field_name)				    \
+do {									    \
+	if (request.field_name != NULL) {				    \
+		mpstream_encode_str(&stream, #field_name);		    \
+		size_t len = request.field_name##_end - request.field_name; \
+		char *buf = mpstream_reserve(&stream, len);		    \
+		memcpy(buf, request.field_name, len);			    \
+		mpstream_advance(&stream, len);				    \
```
Gerold103, line 1390: Wouldn't `mpstream_memcpy()` work shorter?

src/box/txn_limbo.c
```diff
@@ -1333,6 +1492,32 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
...
+		if (txn == NULL)
+			return -1;
+		/*
+		 * It is safe to allow the user to write to fully local spaces
+		 * during the execution of the trigger, bypassing the limbo. We
+		 * check that the transaction is fully local during preparation
+		 * for committing the transaction.
+		 */
+		txn_set_flags(txn, TXN_FORCE_ASYNC);
+		if (on_split_brain_rollback_run_triggers(limbo,
```
Gerold103, line 1513: What happens if users would yield in their trigger? Also `box_txn_commit()` yields. AFAIR, `txn_limbo_req_prepare()` was never ready to yields. For example, some actions it does in the end seem important to be done fast. I suggest to move this code block to the end of the function. Or even to `txn_limbo_req_commit()`.

Gerold103 (commented):
Thanks for the fixes, @CuriousGeorgiy. It is a huge work, and I am thankful you've addressed my comments so diligently and patiently 🙏. Especially about the iteration format and when to start the txns.

src/box/txn_limbo.c
CuriousGeorgiy: MsgPack objects are more user-friendly than raw MsgPack strings with indexing APIs, and they are already extensively used in our user APIs:
1. Persistent function `takes_raw_args` option to pass arguments as MsgPack objects [^1]
2. net.box `return_raw` option to return results as MsgPack objects [^2]
3. `box.space.test.on_recovery_replace` and `box.space.test.before_recovery_replace` trigger arguments [^3]
4. `box.iproto.decode_packet` return values [^4]

Moreover, MsgPack objects support lazy decoding, so one can skip decoding transactions that are not interesting, as in the case when the trigger is called more than once for the same transactions when PROMOTE fails.

[^1]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/func_create/#lua-data.function_options.takes_raw_args
[^2]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#lua-function.conn.request
[^3]: https://github.com/tarantool/doc/issues/3988
[^4]: https://github.com/tarantool/doc/issues/3827

src/box/txn.c
CuriousGeorgiy: Yes, I followed this logic, however `txn_` functions accept a `txn` parameter (since they are like a method), while I want this function to handle the case when there is no `txn` transparently. Would you prefer to have a `txn_` function to which we pass `NULL` in some cases? I think it would look weird.

src/box/txn_limbo.c
CuriousGeorgiy: You, are right — judging by this code, I had a false impression that `mpstream_memcpy` does not reserve space. Fixed.

src/box/txn_limbo.c
CuriousGeorgiy: Sorry, I didn't pay enough attention to this move — thanks for highlighting, I completely agree. Fixed.

src/box/txn_limbo.c
CuriousGeorgiy: I just wanted to make the least amount of changes for clarity, but I see how it raises more questions — AFAIC, the assertions always hold, moved them to `txn_limbo_last_rollback_entry`.

src/box/txn.c
CuriousGeorgiy: > No matter recovery or not, if a txn didn't have these flags during WAL write, then it won't get them.

Since we set them during recovery based on the limbo state, it will actually get them. See https://github.com/tarantool/tarantool/issues/11053 (WIP changing this commit from refactoring to bugfix).

src/box/txn_limbo.c
CuriousGeorgiy: > AFAIR, txn_limbo_req_prepare() was never ready to yields.

What do you mean by `ready for yields`? This patchset includes tests that run transactions which make WAL write, hence, yielding — the tests work fine. 

> For example, some actions it does in the end seem important to be done fast.

Looking at this code, I don't see why is has to be done `fast`:
https://github.com/tarantool/tarantool/blob/b24328cd57b7cdf9e425394d1aa86e2be450f658/src/box/txn_limbo.c#L1333-L1341

> I suggest to move this code block to the end of the function.

I don't think this would be good: in case of failure this would require rolling back the actions done in the code block above.

> Or even to `txn_limbo_req_commit()`.

This definitely won't work. We cannot fail in `txn_limbo_req_commit`.

src/box/txn.c
Gerold103: I think we can safely call it `txn_` and not require an explicit txn parameter. Subjectively to me it feels like in this case the method prefixes are mostly about their visibility and applicability scope. We do already have `txn_detach()`, `txn_begin_ro_stmt()` (it uses `in_txn()` internally), `txn_savepoint_release()`. As much as I would like us to pass txn always explicitly, when it comes to fiber-local data, we seem to be ok.

src/box/txn.c
Gerold103: Omg 😱, thanks for noticing and reporting this! Is there a chance we could make this in limbo.c file somehow? Since this behaviour is caused by the limbo, it seems natural if we could also work it around inside the limbo, hm? For example, when add_entry happens or somewhere there 🤔?

src/box/txn.c
CuriousGeorgiy: > We do already have `txn_detach()`, `txn_begin_ro_stmt()` (it uses `in_txn()` internally), `txn_savepoint_release()`.

Forgot about those — then I completely agree, fixed.

src/box/txn.c
CuriousGeorgiy: > Is there a chance we could make this in limbo.c file somehow?

I don't think so: how are we going to distinguish between a transaction for which the xrow has the `wait_sync` flag versus a transaction for which we set the `TXN_WAIT_SYNC` because the limbo was not empty? The problem is essentially that we set the `TXN_WAIT_SYNC` flag too aggressively.

> Since this behaviour is caused by the limbo, it seems natural if we could also work it around inside the limbo, hm?

I don't think this a problem, TBH: we have the logic for setting the synchronous replication flags in `txn_journal_entry_new`, so it's reasonable to keep it all there.

src/box/txn_limbo.c
CuriousGeorgiy: > So when we return port_c_add_mp() to Lua code, it stays as binary strings? I had impression that they get decoded into Lua tables 🫣.

Sorry, I was wrong on that part, they are indeed decoded into Lua tables:
https://github.com/tarantool/tarantool/blob/43aa0bf45ef18f76b312fabf3d3842d81a970bae/src/box/lua/misc.cc#L319-L329

AFAIC, one of the benefits for why MsgPack objects are used in the APIs I mentioned above is that one can insert them into a space if they store an array [^1]:
```lua
box.space.bands:insert(msgpack.object({1, 'The Beatles', 1960}))
```

But in our case we store a map, so we cannot benefit from this. I guess the argument about lazy decoding is not so relevant, as this code is not on the critical path of execution, so I agree — let's just use `port_c_add_mp()`, as you suggested.

[^1]: https://www.tarantool.io/ru/doc/latest/reference/reference_lua/msgpack/#lua-object.msgpack.msgpack_object

src/box/txn.c
Gerold103: Thanks for explaining, I see the problem now. You managed to elaborate on that very well, and I think I might have a new idea how we can make it in another way.

Recovery is being processed in box.cc in `wal_stream_apply_dml_row()`. Can we just check those things there and set the force_async flag just there? Before calling `txn_commit_submit()` the function could check if the txn must be force-async.

I want to explain a bit why I am afraid of expanding `txn_journal_entry_new()` - to be completely honest with you, the setting of flags in this function was a sacrifice made for perf, which I am not even sure is worth it anymore. I needed a way how to check whether a txn is sync. For that one has to walk all the statements. And txn-commit function has exactly one place where it happens - `txn_journal_entry_new()` 😣. So I piggybacked the flag setting into this iteration over the statements. Ideally this function should have never changed the txn. Its name is literally "journal entry new", it must only create the journal entry. This was a bad idea I think, from my side. Perhaps I should have better walked the txn statement list twice or at least renamed the function.

However what we could say until now is that at least this function only looks at the txn itself and the limbo. Only 1 global state is involved. The limbo.

With the new commit this function operates on 2 global states: limbo and recovery state. Moreover, it checks the recovery state even after the recovery is surely done. This just looks somehow not right.

And I am trying to find a way to keep the recovery out of it. Seems like all the recovery txns are coming from `wal_stream_apply_dml_row()`. For the xlogs. For the snapshot I think we have something similar, but don't remember where. Perhaps in the snapshot case all is fine anyway.

So, would making this same check in `wal_stream_apply_dml_row()` be enough? If necessary, we could walk the txn statements in that function.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+        {'replace', {0}, nil, nil, {0}},
+        {'replace', {777, 1}, nil, {777, 0}, {777, 1}},
+        {'insert', {0}, nil, nil, {0}},
+        {'delete', {0}, nil, nil, nil},
+        {'delete', {777}, nil, {777, 0}, nil},
+        {'update', {777}, {{'=', 2, 1}}, {777, 0}, {777, 1}},
+        {'update', {777}, {{'=', 2, 1}, {'=', 3, 1}}, {777, 0}, {777, 1, 1}},
+        {'upsert', {777}, {{'=', 2, 1}}, {777, 0}, {777, 1}},
+        {'upsert', {0}, {{'=', 2, 1}}, nil, {0}},
+    }
```
Gerold103, line 195: Have you thought about maybe having them as key-value tables? Like
```
local stmts = {
    {type = 'replace', key = ..., tuple = ... },
    ...
}
```
Tbh, it is quite hard to read at the moment, and you have to pass all those items individually in the code above, while could pass each statement as a key-value table. I feel like this would simplify the code above quite a bit.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+        t.assert_not(rawget(_G, 'trigger_called'))
+    end)
+end
+
+g.test_async_committed_async_tx_in_limbo = function(cg)
+    partition_server(cg.replica)
+    cg.master:exec(function()
+        require('fiber').create(function()
+            box.space.s:replace{0}
+        end)
```
Gerold103, line 226: You must be able to use `submit` with synchronous txns too now. I think it was merged to master already.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+end
+
+-- Test that the trigger does not fire for synchronously committed synchronous
+-- transactions.
+g.test_sync_committed_sync_tx = function(cg)
+    partition_server(cg.replica)
+    cg.master:exec(function()
+        require('fiber').create(function()
+            box.space.s:replace{0}
+        end)
```
Gerold103, line 248: Ditto.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+        t.assert(_G.trigger_called)
+    end, {cg.replica:get_instance_id()})
+end
+
+g.before_test('test_non_fully_local_txns', function(cg)
+    cg.master:exec(function()
+        local event = 'box.ctl.on_replication_split_brain_rollback'
+        _G.trigger.set(event, 't', function()
+            local msg = "Can't modify data on a read-only instance"
+            _G.trigger_ok= pcall(function()
```
Gerold103, line 336: Whitespace missing after `trigger_ok`.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+    cg.master:exec(function(replica_id)
+        t.helpers.retrying({timeout = 120}, function()
+            local upstream = box.info.replication[replica_id].upstream
+            t.assert_equals(upstream.status, 'loading')
+            local msg = 'A rollback for a synchronous transaction is received'
+            t.assert_equals(upstream.message, msg)
+        end)
+        t.assert_equals(box.info.synchro.queue.len, 1)
+    end, {cg.replica:get_instance_id()})
+end
```
Gerold103, line 496: I have to say - the test coverage is very impressive, I wouldn't even think about some of those cases myself.

Gerold103 (commented):
Thank you for the fixes! Now I had a closer look at the tests, given that the main patch is basically almost perfect.

src/box/txn_limbo.c
Gerold103: > What do you mean by ready for yields?

I mean we never intended for this function to yield. Who knows what happens if some other fiber can observe `is_in_rollback` being set, nothing else being changed, and `PROMOTE` in progress?

Anyway, nevermind, it is just me being paranoid again. And you are completely right about everything else, like that it is covered with tests; it has to be done before anything else because it can fail; and that we would have to revert things if move this to the end of `txn_limbo_req_prepare()`.

Thank you for checking!

src/box/txn_limbo.c
Gerold103: Thanks for the investiation!

test/replication-luatest/gh_11053_gc_consumer_record_lost_after_promote_and_restart_test.lua
```diff
@@ -0,0 +1,80 @@
...
+    }
+    cg.replica = cg.replica_set:build_and_add_server{
+        alias = 'replica',
+        box_cfg = cg.box_cfg,
+    }
+    cg.replica_set:start()
+    cg.replica_set:wait_for_fullmesh()
+    cg.master:exec(function()
+        box.schema.space.create('s', {is_sync = true}):create_index('p')
+        box.ctl.promote()
```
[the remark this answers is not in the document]
CuriousGeorgiy: Please note that the elections are disabled by default and that I do not enable them for tests — a promote request does not need quorum.

src/box/txn_limbo.c
```diff
@@ -1306,6 +1336,132 @@ txn_limbo_update_system_spaces_is_sync_state(struct txn_limbo *limbo,
...
+{
+	struct txn_limbo_entry *last_rollback =
+		txn_limbo_last_rollback_entry(limbo, lsn);
+	if (last_rollback == NULL)
+		return 0;
+	struct txn_limbo_entry *e;
+	struct region *region = &fiber()->gc;
+	rlist_foreach_entry_reverse(e, &limbo->queue, in_queue) {
+		if (e->txn->fiber != NULL)
+			continue;
```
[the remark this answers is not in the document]
CuriousGeorgiy: Because we only care about asynchronously committed transactions (which don't have a fiber attached to them), for which the user cannot get the result of the commit. For synchronously committed transactions (i.e., `wait_mode = 'complete'`) the user will get the result of the commit directly.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
```diff
@@ -0,0 +1,496 @@
...
+                            t.assert_equals(new, new_tuple)
+                            counter = counter + 1
+                        end)
+                    end
+                    t.assert_equals(n_stmts, counter - 1)
+                end
+            end)
+        box.cfg{replication_synchro_quorum = ''}
+        box.space.s:replace{777, 0}
+        box.cfg{replication_synchro_quorum = 3}
```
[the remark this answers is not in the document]
CuriousGeorgiy: Because we need a tuple to exist in the database to test `update` and `upsert` statements.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
CuriousGeorgiy: Thanks, applied your suggestion, it indeed looks better now.

I guess the lesson of this patch series for me is to stop passing so much stuff around 😁

src/box/txn_limbo.c
```diff
@@ -1333,6 +1489,32 @@ txn_limbo_req_prepare(struct txn_limbo *limbo,
...
+		/*
+		 * It is safe to allow the user to write to fully local spaces
+		 * during the execution of the trigger, bypassing the limbo. We
+		 * check that the transaction is fully local during preparation
+		 * for committing the transaction.
+		 */
+		txn_set_flags(txn, TXN_FORCE_ASYNC);
+		if (on_split_brain_rollback_run_triggers(limbo,
+							 req->lsn + 1) != 0 ||
+		    box_txn_commit() != 0) {
```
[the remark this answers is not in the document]
CuriousGeorgiy: > What if there are multiple independent triggers? AFAIU, the event supports that. Is it intentional that they will all run in the same txn? 

Yes, it is intentional.

> What if there were a lot of async rows committed - the txn might get quite large.

So what's the problem that you see? От перестановки мест слагаемых сумма не меняется.

> Do we maybe want to run it per each invocation of the callback?

What if one of the triggers fails, while we commit the transactions for other triggers — how we are going to handle that? I think the current semantics is the most predictable and well-defined.

> And whatever is decided, lets mention it in the doc.

Reworded the commit message to explicitly use event trigger_s_ execution.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
CuriousGeorgiy: I need to synchronously commit this transaction.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
CuriousGeorgiy: Ditto.

src/box/txn.c
CuriousGeorgiy: Fixed, removed the check from `txn_journal_entry_new`, now we set `TXN_FORCE_ASYNC` during recovery of all transactions that did not go through the limbo at commit time.

test/replication-luatest/on_replication_split_brain_rollback_event_test.lua
CuriousGeorgiy: Fixed.

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: But wait, the user can get async commits results too. With `on_commit` trigger. And in `on_rollback` trigger (if the user would set them), the user has no way to see if it was due to split brain. Shouldn't it be all unified regardless of `wait_mode`?

src/box/txn.c
Gerold103: Thank you 🙏.

test/replication-luatest/gh_11053_gc_consumer_record_lost_after_promote_and_restart_test.lua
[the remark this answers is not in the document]
Gerold103: Omg, I forgot how horrible the promotion without elections is 😂.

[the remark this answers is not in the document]
Gerold103: But then why do you even specify the quorum at all, if its value doesn't matter? It does look confusing in the test.

src/box/txn_limbo.c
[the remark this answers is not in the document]
Gerold103: The repair queue AFAIU was mainly for async transactions, which were committed without a quorum. Not for non-blocking synchro txns. Those anyway can't be confirmed and then rolled back. This type of split-brain handling only works for the "fake split brain" which is caused by async txns.

If we get a split-brain on actual synchronous txns, it only means the quorum wasn't right, and this can't be repaired. Because such txns most likely aren't even in the limbo already.

Gerold103:
@sergepetrenko and @Serpentian , please, join the review. I have no major comments left. We are just discussing last nits with Georgy.

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: I agree, let's discuss this verbally. And also summon @Serpentian.

src/box/txn_limbo.c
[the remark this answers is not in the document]
sergepetrenko: I think we should ban issuing`box.begin()/box.commit()` from the trigger callback.
Or at least mention in the doc that these explicit box.begin()/box.commit() will break the user's expectations about the trigger atomicity.

Other than that - I don't see a problem in having a single large txn for everything done in the trigger.

sergepetrenko (commented):
Thanks for the patch!

What bothers me is
1. the trigger name
2. that the trigger is ran only on PROMOTE, but not on usual ROLLBACK (although there the user also doesn't know the state of a wait='submit' transaction).
3. that some transactions (committed synchronously) are not pushed to the trigger

Let's discuss this on a call.

changelogs/unreleased/gh-10460-on_replication_split_brain_rollback-event.md
```diff
@@ -0,0 +1,6 @@
...
```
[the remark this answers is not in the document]
Gerold103: Good discussion, and I am very sorry that we have to bring this all up at such a late stage 😭, Georgy did a wonderful job in the code, I hope we won't have to change much.

Speaking of names - this is indeed not really a split-brain, when we rollback not committed synchronous transactions. So I like more with the `promote` suffix. Or even deeper - `on_leader_change_tail_rollback`.

PR #10940, opened 2024-12-13
box.sequence: sequence_data recovery after reset

### LINKED ISSUE #9871: Sequence state is not saved after `reset` and tarantool restart.
**Bug description**

Sequence state is not saved after `reset` and tarantool restart.

* OS: Linux
* OS Version: Ubuntu 23.10
* Architecture: amd64

> tarantool --version
> Tarantool 2.11.2-0-g1bac2d257
> Target: Linux-x86_64-RelWithDebInfo
> Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=TRUE
> Compiler: GNU-13.2.0
> C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/psergee/work/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type
> CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -fopenmp -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/psergee/work/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type

**Steps to reproduce**

1. start tarantool
2. create a sequence, space and the index with the sequence set.
3. insert some records
4. reset the sequence
5. restart tarantool
6. insert new record

**Actual behavior**

The sequence is not reset and continues after the tarantool new start.
Here is an example:
```lua
tarantool> box.schema.sequence.create('id_seq', { if_not_exists = true })
box.schema.create_space('test', {
    format = {
        {'id', type='unsigned', is_nullable=false},
        {'name', type='string', is_nullable=false},
    },
    if_not_exists = true,
})
box.space.test:create_index('primary', {
    sequence = 'id_seq',
    if_not_exists = true
})
---
...

tarantool> box.space.test:insert{nil, 'aaa'}
---
- [1, 'aaa']
...

tarantool> box.space.test:insert{nil, 'bbb'}
---
- [2, 'bbb']
...

tarantool> box.sequence.id_seq:reset()
---
...

tarantool> 

$ tarantool
Tarantool 2.11.2-0-g1bac2d257
tarantool> box.cfg{}
. . . 
---
...

tarantool> box.space.test:insert{nil, 'ccc'}
---
- [3, 'ccc']
...
```

**Expected behavior**

The sequence state is reset after tarantool stop and start.
This is how it works without restart after `reset`:
```lua
tarantool> box.space.test:insert{nil, 'aaa'}
---
- [1, 'aaa']
...

tarantool> box.space.test:insert{nil, 'bbb'}
---
- [2, 'bbb']
...

tarantool> box.sequence.id_seq:reset()
---
...

tarantool> box.space.test:insert{nil, 'ccc'}
---
- error: Duplicate key exists in unique index "primary" in space "test" with old tuple
    - [1, "aaa"] and new tuple - [1, "ccc"]
...

tarantool>
```

LevKats:
There are two sources of sequence state in tarantool: `_sequence_data` table and it's light version used in `box/sequence.c` which is stored in a hashmap. The reason for this decision was lack of cross-engine transactions. The first is used in lua API, the second -- internally to process memtx index updates. Now lua `reset()` call always produce WAL entry and we don't rely on memtx tables content during the initial recovery since the latest hasmap state is stored into snapshot directly.

Fixes #9871
NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh_9871_sequence_data_recovery_after_reset.md
+++ b/changelogs/unreleased/gh_9871_sequence_data_recovery_after_reset.md
@@ -1,0 +1,4 @@
+## bugfix/core
+
+* Fixed a bug when the `sequence:reset()` call result was not recovered
+  after the server restart (gh-9871).
--- a/src/box/box.cc
+++ b/src/box/box.cc
@@ -4258,13 +4258,43 @@ box_sequence_set(uint32_t seq_id, int64_t value)
 API_EXPORT int
 box_sequence_reset(uint32_t seq_id)
 {
+	bool was_in_txn = box_txn();
+	box_txn_savepoint_t *savepoint = NULL;
+	if (was_in_txn) {
+		savepoint = box_txn_savepoint();
+		if (savepoint == NULL)
+			return -1;
+	} else if (!was_in_txn && box_txn_begin() != 0) {
+		return -1;
+	}
 	struct sequence *seq = sequence_cache_find(seq_id);
 	if (seq == NULL)
-		return -1;
+		goto error;
 	if (access_check_sequence(seq) != 0)
-		return -1;
+		goto error;
+	int64_t sequence_value;
+
+	if (sequence_get_value(seq, &sequence_value) == 0) {
+		/*
+		 * Without the update here sequence_data_delete call result
+		 * below may not be recovered due to the specifics of how
+		 * on_replace trigger work in the absence of old_tuple
+		 */
+		if (sequence_data_update(seq_id, sequence_value) != 0)
+			goto error;
+	}
 	sequence_reset(seq);
-	return sequence_data_delete(seq_id);
+	if (sequence_data_delete(seq_id) != 0)
+		goto error;
+	if (!was_in_txn && box_txn_commit() != 0)
+		goto error;
+	return 0;
+error:
+	if (was_in_txn)
+		box_txn_rollback_to_savepoint(savepoint);
+	else
+		box_txn_rollback();
+	return -1;
 }
 
 API_EXPORT int
--- a/src/box/request.c
+++ b/src/box/request.c
@@ -240,9 +240,15 @@ request_handle_sequence(struct request *request, struct space *space,
 		 * auto increment request never tries to insert a
 		 * value that is already in the space. Note, this
 		 * code is also invoked on final recovery to restore
-		 * the sequence value from WAL.
+		 * the sequence value from WAL. However, during the
+		 * initial recovery, proper sequence values are stored
+		 * in snapshots, and one should use these because
+		 * otherwise, sequence values would depend on
+		 * the order of recovered spaces which may be
+		 * misleading.
 		 */
-		if (likely(mp_read_int64(&key, &value) == 0))
+		if (likely(mp_read_int64(&key, &value) == 0 &&
+			   recovery_state != INITIAL_RECOVERY))
 			return sequence_update(seq, value);
 	}
 	request_update_header(request, request->header, region);
--- a/test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
+++ b/test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
@@ -1,0 +1,79 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+local group_config = {
+    {engine = 'memtx', use_txn = true},
+    {engine = 'memtx', use_txn = false},
+    {engine = 'vinyl', use_txn = false}
+}
+local g = t.group('storage', group_config)
+
+g.before_all(function(cg)
+    cg.server = server:new()
+    cg.server:start()
+end)
+
+g.before_each(function(cg)
+    cg.server:exec(function(engine, use_txn)
+        box.schema.sequence.create('id_seq')
+        local s = box.schema.create_space('test', {
+            format = {
+                {'id', type='unsigned', is_nullable=false},
+                {'name', type='string', is_nullable=false},
+            },
+            engine = engine,
+            if_not_exists = true,
+        })
+        s:create_index('primary', {
+            sequence = 'id_seq',
+            if_not_exists = true
+        })
+        if use_txn then
+            box.begin()
+        end
+        s:insert{nil, 'aaa'}
+        s:insert{nil, 'bbb'}
+        s:insert{nil, 'ccc'}
+        local seq = box.sequence.id_seq
+        seq:reset()
+        s:delete(1)
+        if use_txn then
+            box.commit()
+        end
+    end, {cg.params.engine, cg.params.use_txn})
+end)
+
+local function commit()
+    local s = box.space.test
+    s:insert{nil, 'ddd'}
+    local ddd = s:get(1)
+    t.assert_equals(ddd[2], 'ddd')
+end
+
+g.test_no_recovery = function(cg)
+    cg.server:exec(commit)
+end
+
+g.test_snapshot_recovery = function(cg)
+    cg.server:exec(function()
+        box.snapshot()
+    end)
+    cg.server:restart()
+    cg.server:exec(commit)
+end
+
+g.test_wal_recovery = function(cg)
+    cg.server:restart()
+    cg.server:exec(commit)
+end
+
+g.after_each(function(cg)
+    cg.server:exec(function()
+        box.space.test:drop()
+        box.sequence.id_seq:drop()
+    end)
+end)
+
+g.after_all(function(cg)
+    cg.server:drop()
+end)


test/box-luatest/gh_9871_sequence_restart_after_reset_test.lua
```diff
@@ -0,0 +1,69 @@
+local server = require('luatest.server')
+local t = require('luatest')
```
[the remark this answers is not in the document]
locker: If you make the test multi-engine, please put it in the `engine-luatest` suite.

src/box/box.cc
```diff
@@ -4263,6 +4263,16 @@ box_sequence_reset(uint32_t seq_id)
...
 	if (access_check_sequence(seq) != 0)
 		return -1;
+	int64_t sequence_value;
+	if (sequence_get_value(seq, &sequence_value) == 0) {
+		/*
+		 * Our goal here is to guarantee the existence of an
+		 * entry corresponding to that seq_id so it's deletion
+		 * would affect WAL contents.
+		 */
+		if (sequence_data_update(seq_id, sequence_value) != 0)
```
locker, line 4273: I think `sequence_data_update` and `sequence_data_reset` should be performed in the same transaction, otherwise, there's a chance that the sequence is reset and updated between the two operations.

src/box/request.c
```diff
@@ -242,7 +242,8 @@ request_handle_sequence(struct request *request, struct space *space,
 		 * code is also invoked on final recovery to restore
 		 * the sequence value from WAL.
 		 */
-		if (likely(mp_read_int64(&key, &value) == 0))
+		if (likely(mp_read_int64(&key, &value) == 0) &&
+		    recovery_state != INITIAL_RECOVERY)
```
locker, line 246: This looks correct, but I don't understand why you change this code now. Do some tests fail without this check?

changelogs/unreleased/gh_9871_sequence_data_recovery_after_reset.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Fixed a bug that resulted in `sequence:reset()` call result not saved
+  in WAL and snapshot and thus not recovered after the server restart
+  (gh-9871).
```
lenkis, line 5:
```suggestion
* Fixed a bug when the `sequence:reset()` call result was not saved
  in WAL and snapshot and thus was not recovered after the server restart
  (gh-9871).
```

lenkis (approved):

test/box-luatest/gh_9871_sequence_restart_after_reset_test.lua
```diff
@@ -0,0 +1,69 @@
...
+    g.server:restart()
+    g.server:exec(common)
+end
+
+g.after_each(function(g)
+    g.server:exec(function()
+        box.space.test:drop()
+    end)
+end)
+
```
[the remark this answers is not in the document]
LevKats: Fixed

test/box-luatest/gh_9871_sequence_restart_after_reset_test.lua
```diff
@@ -0,0 +1,69 @@
...
+end
+
+g.test_no_recovery = function(g)
+    g.server:exec(common)
+end
+
+g.test_snapshot_recovery = function(g)
+    g.server:exec(function()
+        box.snapshot()
+        local fiber = require('fiber')
```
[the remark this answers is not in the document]
LevKats: Yes indeed. Fixed

test/box-luatest/gh_9871_sequence_restart_after_reset_test.lua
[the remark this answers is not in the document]
LevKats: Done. Now it also test transactions in memtx.

src/box/box.cc
```diff
@@ -4263,6 +4263,16 @@ box_sequence_reset(uint32_t seq_id)
...
+			return -1;
+	}
```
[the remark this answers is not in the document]
LevKats: I haven't done anything with triggers to replay such entries but now transactions are used. The comment is also fixed as far as I understand.

src/box/box.cc
LevKats: Done

changelogs/unreleased/gh_9871_sequence_data_recovery_after_reset.md
LevKats: Thanks! Fixed.

src/box/request.c
LevKats: Yes, `test_snapshot_recovery` is failing without this check. Actually it seems that the second condition should be unlikely too since it is irrelevant after the recovery. The problem with snapshots is that spaces, that depend on sequence may be recovered after sequences, so values from that spaces would be applied to sequences, which may be incorrect, when reset is the last operation before the checkpoint for example.

test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
```diff
@@ -0,0 +1,78 @@
...
+        local seq = box.sequence.id_seq
+        seq:reset()
+        s:delete(1)
+        if use_txn then
+            box.commit()
+        end
+    end, {cg.params.engine, cg.params.use_txn})
+end)
+
+local common = function()
```
Serpentian, line 46: Nit: please, let's define helper functions as `local function commit`, otherwise it's easy to confuse it with a test)

Serpentian (commented):
Only nits are left from my side. I'm ok with the current patch, since it won't be easy to allow `on_replace`, when there's no `old_tuple` and there's no much sense in it.

src/box/box.cc
```diff
@@ -4258,13 +4258,33 @@ box_sequence_set(uint32_t seq_id, int64_t value)
...
+		 */
+		if (sequence_data_update(seq_id, sequence_value) != 0)
+			goto exit_commit;
+	}
 	sequence_reset(seq);
-	return sequence_data_delete(seq_id);
+	rc = sequence_data_delete(seq_id);
+
+exit_commit:
+	if (!was_in_txn && box_txn_commit() != 0)
```
locker, line 4285: You should call `box_txn_rollback_to_savepoint` or `box_txn_rollback` on failure instead of `box_txn_commit`.

test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
```diff
@@ -0,0 +1,78 @@
...
+            format = {
+                {'id', type='unsigned', is_nullable=false},
+                {'name', type='string', is_nullable=false},
+            },
+            engine = engine,
+            if_not_exists = true,
+        })
+        s:create_index('primary', {
+            sequence = 'id_seq',
+            if_not_exists = true
```
locker, line 29: `if_not_exists` seems unnecessary because you drop the space at the end of the test.

Also, I think you should drop the sequence there as well.

src/box/request.c
locker: Putting the whole `if` condition under `likely` makes sense. It may also be worth updating the comment preceding the check.

src/box/box.cc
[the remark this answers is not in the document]
LevKats: Fixed the comment in `src/box/box.cc` so WAL is not mentioned there anymore

src/box/request.c
LevKats: Fixed both the condition and the comment above

changelogs/unreleased/gh_9871_sequence_data_recovery_after_reset.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/core
+
+* Fixed a bug when the `sequence:reset()` call result was not saved
+  in WAL and snapshot and thus was not recovered after the server restart
```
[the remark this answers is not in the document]
LevKats: Done. Now the commit message and changelog don't appeal to the WAL entries directly.

test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
LevKats: Fixed

src/box/box.cc
LevKats: Fixed. Take a look

test/engine-luatest/gh_9871_sequence_restart_after_reset_test.lua
LevKats: Done

src/box/box.cc
locker: You shouldn't rollback the whole outer transaction on failure. Please take a look at `box_txn_savepoint()` and `box_txn_rollback_to_savepoint()`.

LevKats: For some reason I thought that resetting is so important that we should break the entire transaction. Now savepoints are used for external transactions.

Serpentian (approved):

locker:
Cherry-picked to 3.3, 3.2, 2.11.

PR #10952, opened 2024-12-17
log: fix truncating log message to 1024 bytes in 'json' log

### LINKED ISSUE #10918: Remove the "message" size limit if log_format=json, currently 1Kb
Some customers ask to remove the message size limit when setting the "log_format=json" parameter. Now messages are cut off at 1Kb

grafin:
Before this patch, when log format was set to 'json' and log entry was written with format-string API (`require('log').info(<format string>, args...)`), in the output the message was truncated to 1024 bytes.

This patch fixes this behavior, and now the message will only be truncated if the full log entry (with all the context) doesn't fit into the 16KiB buffer. The message will be truncated in a way, so all the context is present.

Closes #10918

NO_DOC=bugfix

### DIFF
--- a/test/app-tap/logger.test.lua
+++ b/test/app-tap/logger.test.lua
@@ -3,7 +3,7 @@
 local log = require('log')
 
 local test = require('tap').test('log')
-test:plan(119)
+test:plan(142)
 
 local function test_invalid_cfg(cfg_method, cfg, name, expected)
     local _, err = pcall(cfg_method, cfg)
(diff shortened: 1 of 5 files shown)


src/lib/core/tt_static.h
```diff
@@ -56,6 +59,17 @@ tt_static_buf(void)
 					     sizeof(intptr_t));
 }
 
+/**
+ * Return a thread-local statically allocated temporary buffer of
+ * size @a TT_STATIC_BUF_BIG_LEN.
+ */
+static inline char *
+tt_static_buf_big(void)
```
locker, line 67: I don't quite like the fact that we have two kinds of static buffers now. I think it's possible to format and escape the message right in the `say_buf`. To achieve that, we could:

1. Make `json_escape` work in-place (i.e. with the same `buf` and `data`). AFAIU this should be fairly easy to do: first, scan the input string to count its final size, then escape characters while scanning the input string backwards.
2. Patch `say_format_json` so that it works as follows:
   - Format the string with an empty message.
   - After formatting the log record, `memmove` everything after the message to the end of the buffer.
   - Next, format the message in the resulting gap and escape it in-place.
   - Finally, `memmove` the message tail after the message.

What do you think?

lenkis (approved):

src/lib/core/tt_static.h
grafin: Thanks for the suggestions! I did rewrite it, like you proposed.

src/lib/core/tt_static.h
locker, line 1: Looks like with the new approach we don't need to patch this file.

locker (approved):

src/lib/core/tt_static.h
grafin: Yeah, I've overlooked this. Thanks for pointing it out!

grafin:
I would like to notice, that after this patch `json_escape(char *buf, int size, const char *data)` is no longer used anywhere. I'm not sure we need to delete it, but we might as well, to reduce the amount of dead code. @locker @Totktonada, whats your opinion on that?

locker:
> I would like to notice, that after this patch `json_escape(char *buf, int size, const char *data)` is no longer used anywhere. I'm not sure we need to delete it, but we might as well, to reduce the amount of dead code. @locker @Totktonada, whats your opinion on that?

I don't mind dropping it. Then it may be better to use the old name (`json_escape`) for the new function (called `json_escape_inplace` in this patch).

grafin:
Nevermind, I've overlooked that `json_escape` is still used as an argument for `SNPRINT` macro.

test/box/session_settings.test.lua
```diff
@@ -133,7 +133,7 @@ settings.sql_default_engine = true
 settings.sql_full_metadata = 'false'
 settings.sql_parser_debug = 'string'
 
-str = string.rep('a', 20 * 1024)
+str = string.rep('a', 37 * 1024)
```
Totktonada, line 136: I guess this change is not needed after reverting the static buffer size change.

test/app-tap/logger.test.lua
Totktonada, line 1: How about corner case tests?

* A message with length around the max capacity (plus 1, plus 2, minus 1, minus 2, exact) -- or just verify everything in between [16k-100; 16k+100].
* A message that ends with a symbol to be escaped, but this +1 symbol is out of buffer capacity (we shouldn't end the message with dangling backslash).
* Empty string message.

Are there such cases? If they're exist, we should update the testing message sizes.

Should we verify both `format == "json"` and the opposite scenarios? There is a condition regarding it in the changed code.

How about a new luatest-based test?

Totktonada (approved):
I've looked on the code and it seems correct. Thanks!

I'm in some doubt, whether we test it enough. Left a comment on this topic. Otherwise, OK.

grafin:
I've added more tap tests (luatest tests will be developed in scope of reworking the logger module task). Current solution works fine for `log.<level>("Long message format string...")` cases. But it fails to provide a valid JSON output in case of `log.<level>({message = "Long message...", ...})`, it was broken before, it is still broken, I don't see a reasonably quick solution now.

test/box/session_settings.test.lua
grafin: Fixed.

test/app-tap/logger.test.lua
Totktonada: Seems solved.

grafin: Thanks for suggesting this! I thinks I've covered all the mentioned cases with tests now, and found several bugs while doing so)

PR #10960, opened 2024-12-19
test: parallelize config-luatest/log_wrapper_test

Totktonada:
The test has 172 test cases and spawns a lot of processes. It takes ~30 seconds (+/-10 seconds) on my laptop in the sequential and in the parallel run both. The difference seems within the noise level. The parallel execution seems slightly longer, approximately +5 seconds.

However, if we run the test cases separately, the test timeout limit (110 seconds by default) is applied to each particular test case. It may be important for testing runs on a slow machine or when a slow instrumented build (such as ASAN) is used.

### DIFF
--- a/test/config-luatest/log_wrapper_test.lua
+++ b/test/config-luatest/log_wrapper_test.lua
@@ -1,3 +1,5 @@
+-- tags: parallel
+
 local t = require('luatest')
 local treegen = require('luatest.treegen')
 local justrun = require('luatest.justrun')


ligurio:
Why have you changed only one test in commit and one test in a previous commit (174f7b39e9aa4ced1e8c18865d1af282a5c2b09c)? I suspect we have more than two tests that could be run in parallel.

Totktonada:
> Why have you changed only one test in commit and one test in a previous commit ([174f7b3](https://github.com/tarantool/tarantool/commit/174f7b39e9aa4ced1e8c18865d1af282a5c2b09c))? I suspect we have more than two tests that could be run in parallel.

The change itself is tiny, but I perform execution time measurements, stability checks and it takes some time. I don't want to add new problems (like flaky tests), so I propose these parallelization changes only after some adequate verification.

I can't do it for all the tests that may gain from the parallelization at once. So, I'm going step-by-step.

Totktonada:
As a side effect of separating such changes I have more atomic changes with good descriptions. In fact, I like it.

I remember that several months ago I've tried to parallelize several tests, found some flaky one and didn't finished this work.

grafin (approved):

grafin:
Thanks for looking into this!

PR #10964, opened 2024-12-19
gitignore: add DartConfiguration.tcl

Totktonada:
Seems like this file is generated in the build directory for the ancient DART testing tool. When the build directory is the same as the source directory, git shows it as a new untracked file. That's irritating.


Totktonada:
Thanks for the glance!

> I would find a full list of files (like this [^1]) autogenerated by CMake and add it to `.gitignore` in a single commit
>
> [^1]: https://github.com/github/gitignore/blob/main/CMake.gitignore

Seems like it is already there. At least, my working directory is clean after the in-source build.

https://github.com/tarantool/tarantool/blob/dd5039a1da2fdc8dc07f4de93015da43ef6bf9fe/.gitignore#L24-L29

https://github.com/tarantool/tarantool/blob/dd5039a1da2fdc8dc07f4de93015da43ef6bf9fe/.gitignore#L65

https://github.com/tarantool/tarantool/blob/dd5039a1da2fdc8dc07f4de93015da43ef6bf9fe/.gitignore#L85

> Or generate `.gitignore` automatically by CMake [^2].
>
> [^2]: https://cmake.org/cmake/help/latest/command/file.html#generate

You know how I 'like' CMake things :)

I don't mind if someone else add some dev tooling that is convenient for him/her, but personally I prefer to just edit `.gitignore`.

Totktonada:
Seems like `full-ci` would be redundant here: I didn't change neither code, nor tests.

PR #10981, opened 2024-12-24
test: speedup and stabilize the gh_7605_qsort_recovery test

### LINKED ISSUE #10961: The test `box-luatest/gh_7605_qsort_recovery_test.lua` flaky fails
Sometimes the test fails like that:
```
[001] box-luatest/gh_7605_qsort_recovery_test.lua                                                                       [ fail ]
[001] Test failed! Output from reject file /tmp/t/rejects/box-luatest/gh_7605_qsort_recovery.reject:
[001] Tarantool version is 3.4.0-entrypoint-27-g2afead5ecc
[001] TAP version 13
[001] 1..1
[001] # Started on Thu Dec 19 17:18:15 2024
[001] # Starting group: box-luatest.gh_7605_qsort_recovery
[001] not ok 1	box-luatest.gh_7605_qsort_recovery.test_qsort_recovery
[001] #   fiber slice is exceeded
[001] #   stack traceback:
[001] #   	builtin/box/schema.lua:1944: in function '(for generator)'
[001] #   	...e/sptnt/test/box-luatest/gh_7605_qsort_recovery_test.lua:65: in function <...e/sptnt/test/box-luatest/gh_7605_qsort_recovery_test.lua:60>
[001] #   	...e/sptnt/test/box-luatest/gh_7605_qsort_recovery_test.lua:60: in function 'box-luatest.gh_7605_qsort_recovery.test_qsort_recovery'
[001] #   artifacts:
[001] #   	default -> /private/tmp/t/001_box-luatest/artifacts/default-Lk0Tx4_rYafC
[001] # Ran 1 tests in 106.874 seconds, 0 succeeded, 1 errored
[001] 
```

Also, even when it succeeds, it runs for around 150-180 seconds on my computer with Tarantool built in debug (like in coverage workflow), although on commit which introduced the test (e1d96170850) it runs for only ~60 seconds and there are no flaky failures.

Sampling shows that on newer tarantool versions at least one third of time is taken by `fiber_on_gc_alloc` introduced in commit 19abfd2a39205de270836adda7f3a733476acaa4.

The test becomes too long and unstable with fiber allocation checks, so it's time to make it shorter.

sergepetrenko:
The test used to take around 180 seconds in debug build and was quite flaky.

When the test was introduced in commit e1d961708504 ("Fix a bug in qsort") it was stated that such a long test with 2 million keys was needed to reproduce the bug.

Since the test introduction our sort implementation was rewritten in commit ec34bf7ba3ea ("core: introduce sample sort algorithm") and commit 4f617b702f0f ("box: introduce memtx_sort_threads config parameter").

Now multithread sort is used as long as there are more than a 1024 keys to sort, so such a huge test seems unnecessary.

Let's reduce test size to 20000 keys from 2000000 and make the test code yield every 1000 iterations instead of 10000, which is a more common practice.

When running the test a lot times in parallel the fiber slice check may still fail during index correctness check, so set fiber slice there as well.

Now the test runs around 6 seconds in debug build and may be removed from long-run list.

The test still spots the original problem in case it appears which can be verified by reverting the fix from the original commit.

Closes #10961

NO_CHANGELOG=test
NO_DOC=test

### DIFF
--- a/test/box-luatest/gh_7605_qsort_recovery_test.lua
+++ b/test/box-luatest/gh_7605_qsort_recovery_test.lua
@@ -34,15 +34,13 @@ end)
 
 g.test_qsort_recovery = function()
     g.server:exec(function()
-        local PACK = 10000
-        local TOTAL = 2000000
+        local PACK = 1000
+        local TOTAL = 20000
         local uuid = require "uuid"
         local fiber = require "fiber"
 
+        fiber.set_max_slice(600)
         for i = 1, TOTAL / PACK do
-            if fiber.set_slice then
-                fiber.set_slice(600)
-            end
             box.begin()
             for k = 1, PACK do
                 box.space.test:replace{(i - 1) * 1000 + k, uuid.str()}
@@ -58,10 +56,12 @@ g.test_qsort_recovery = function()
 
     -- check secondary index correctness
     g.server:exec(function()
-        local PACK = 10000
+        local PACK = 1000
         local prev_tuple = nil
         local fiber = require "fiber"
         local i = 0
+
+        fiber.set_max_slice(600)
         for _,tuple in box.space.test.index.i:pairs() do
             if prev_tuple ~= nil then
                 t.assert(prev_tuple[2] < tuple[2], "Unordered!")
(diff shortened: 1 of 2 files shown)


locker (approved):

sergepetrenko:
Backports:
3.3 https://github.com/tarantool/tarantool/pull/10991
3.2 https://github.com/tarantool/tarantool/pull/10992
2.11https://github.com/tarantool/tarantool/pull/10993

PR #10986, opened 2024-12-25
log: encode table arguments in json

### LINKED ISSUE #8611: JSONify tables at logging
Let's look on an example:

```lua
tarantool> log.info('foo: %s', {bar = 'baz'})
```

What does it print?

```
foo: table: 0x41419f28
```

It seems, I should encode my data into JSON myself. Okay...

```lua
tarantool> json = require('json')
tarantool> log.info('foo: %s', json.encode({bar = 'baz'}))
foo: {"bar":"baz"}
```

Not bad. But there are two problems:

1. If the given log level is disabled, such code spends CPU cycles on unnecessary JSON encoding.
2. If the json encoder instance is not configured as `.cfg({encode_use_tostring = true})`, the encoding may fail.

And, to be honest, it is just less convenient.

<details>
<summary>My workaround</summary>

(A bit stripped code from my actual wrapper.)

(Doesn't handle tables with `__serialize` or `__tostring` metamethods well.)

```lua
-- Logger wrapper with a few enhancements.
--
-- 1. Encode tables into JSON.

local json_noexc = require('json').new()
json_noexc.cfg({encode_use_tostring = true})

local logger_name = '<..my module name..>'
local log = require('log').new(logger_name)

local func2level = {
    [log.error] = 2,
    [log.warn] = 3,
    [log.info] = 5,
    [log.verbose] = 6,
    [log.debug] = 7,
}

local str2level = {
    ['error'] = 2,
    ['warn'] = 3,
    ['info'] = 5,
    ['verbose'] = 6,
    ['debug'] = 7,
}

local function say_closure(log_f)
    return function(fmt, ...)
        -- Skip logging based on the log level.
        --
        -- log.cfg is the main logger configuration. See gh-8610.
        local level = log.cfg.modules and log.cfg.modules[logger_name] and
            log.cfg.modules[logger_name] or log.cfg.level
        level = str2level[level] or level
        assert(type(level) == 'number')
        if func2level[log_f] > level then
            return
        end

        -- Micro-optimization: don't create a temporary table if
        -- it is not needed.
        local argc = select('#', ...)
        if argc == 0 then
            log_f(fmt)
            return
        end

        -- Encode tables into JSON.
        --
        -- Ignores presence of __serialize and __tostring in the
        -- metatatable. It is suitable for the module needs.
        local args = {...}
        for i = 1, argc do
            if type(args[i]) == 'table' then
                args[i] = json_noexc.encode(args[i])
            end
        end

        -- Pass the result to the logger function.
        log_f(fmt, unpack(args, 1, argc))
    end
end

return {
    error = say_closure(log.error),
    warn = say_closure(log.warn),
    info = say_closure(log.info),
    verbose = say_closure(log.verbose),
    debug = say_closure(log.debug),
}
```
</details>

locker:
This is more convenient than calling `json.encode` manually. It is also more efficient in case logging is skipped due to the log level.

How it used to work:

```
tarantool> require('log').info('%s', {foo = 'bar'})
table: 0x41f9d2e8
```

How it works now:

```
tarantool> require('log').info('%s', {foo = 'bar'})
{"foo":"bar"}
```

Note, JSON encoding is enabled only for tables that don't have the `__tostring` meta-method:

```
tarantool> require('log').info('%s', setmetatable({foo = 'bar'}, {__tostring = function() return 'foobar' end}))
foobar
```

Closes #8611

### DIFF
--- a/changelogs/unreleased/gh-8611-log-tables-in-json.md
+++ b/changelogs/unreleased/gh-8611-log-tables-in-json.md
@@ -1,0 +1,3 @@
+## feature/lua
+
+* Built-in logger now encodes table arguments in the JSON format (gh-8611).
--- a/src/lua/log.lua
+++ b/src/lua/log.lua
@@ -166,9 +166,19 @@ local function say(self, level, fmt, ...)
     local type_fmt = type(fmt)
     local format = "%s"
     local msg
-    if select('#', ...) ~= 0 then
+    local argc = select('#', ...)
+    if argc ~= 0 then
+        local args = {...}
+        for i = 1, argc do
+            if type(args[i]) == 'table' then
+                local mt = getmetatable(args[i])
+                if mt == nil or mt.__tostring == nil then
+                    args[i] = json.encode(args[i])
+                end
+            end
+        end
         local stat
-        stat, msg = pcall(string.format, fmt, ...)
+        stat, msg = pcall(string.format, fmt, unpack(args, 1, argc))
         if not stat then
             error(msg, 3)
         end
--- a/test/app-luatest/gh_8611_log_tables_in_json_test.lua
+++ b/test/app-luatest/gh_8611_log_tables_in_json_test.lua
@@ -1,0 +1,28 @@
+local t = require('luatest')
+local treegen = require('luatest.treegen')
+local justrun = require('luatest.justrun')
+
+local g = t.group()
+
+g.test_log_tables_in_json = function()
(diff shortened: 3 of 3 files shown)


Totktonada (approved):
Seems OK. Thanks!

grafin (approved):
Thanks for fixing this! Looks good.

lenkis (approved):

PR #10990, opened 2024-12-25
config: migrate extras module definitions to EE

georgiy-belyanin:
This patch migrates config.extras module definitions to Tarantool EE.
Now its sources are defined there.

Also, it makes Tarantool validate extras module during a config
initialization phase. It is changed this way to achieve possibility to
load configuration extras after loading the basic config module to
simplify the order of modules loading.

Previous loading order:
* Load `config.extras`.
* Load `config`.
* Load a few more `config.extras`.

New loading order:
* Load `config`.
* Load `config.extras`.

The simplification comes with a cost of less checks in no-configuration
mode though they're not so important since the scenario when we add
`config.extras` and work in no-configuration mode sounds quite strange.

Needed for tarantool/tarantool-ee#1024

### DIFF
--- a/src/box/CMakeLists.txt
+++ b/src/box/CMakeLists.txt
@@ -78,10 +78,7 @@ lua_source(lua_sources lua/config/utils/funcutils.lua      config_utils_funcutil
 lua_source(lua_sources lua/config/utils/network.lua        config_utils_network_lua)
 
 if (ENABLE_CONFIG_EXTRAS)
-    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/etcd.lua config_source_etcd_lua)
-    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/storage/init.lua config_storage_init_lua)
-    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/storage.lua config_source_storage_lua)
-    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/extras.lua config_extras_lua)
+    lua_multi_source(lua_sources ${CONFIG_EXTRAS_LUA_SOURCES})
 endif()
 # }}} config
 
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -35,7 +35,7 @@ local function load_extras()
     return extras
 end
 
-local extras = load_extras()
+local extras
 
 -- {{{ Helpers
 
@@ -105,6 +105,8 @@ function methods._register_applier(self, applier)
 end
 
 function methods._initialize(self)
+    extras = load_extras()
+
     -- The sources are synchronized in the order of registration:
     -- env, file, etcd (present in Tarantool EE), env for
     -- defaults.
--- a/src/box/lua/config/extras.h
+++ b/src/box/lua/config/extras.h
@@ -1,0 +1,16 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2010-2024, Tarantool AUTHORS, please see AUTHORS file.
+ */
+#pragma once
+
+#include "trivia/config.h"
+
+#if defined(ENABLE_CONFIG_EXTRAS)
+#include "lua/config/extras_impl.h"
+#else /* !defined(ENABLE_CONFIG_EXTRAS) */
+
+#define CONFIG_EXTRAS_LUA_MODULES
+
+#endif /* !defined(ENABLE_CONFIG_EXTRAS) */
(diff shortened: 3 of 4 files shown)


src/box/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ lua_source(lua_sources lua/config/utils/textutils.lua      config_utils_textutil
 if (ENABLE_CONFIG_EXTRAS)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/etcd.lua config_source_etcd_lua)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/storage/init.lua config_storage_init_lua)
+    lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/storage_client.lua config_storage_client_lua)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/source/storage.lua config_source_storage_lua)
     lua_source(lua_sources ${CONFIG_EXTRAS_DIR}/extras.lua config_extras_lua)
 endif()
```
Totktonada, line 83: Maybe it is the good time to finally move the list of these EE config modules to the EE sources? See `ENABLE_FAILOVER` below for example. It is tracked in TNTP-280.

src/box/lua/config/init.lua
```diff
@@ -57,7 +58,16 @@ end
 
 local methods = {}
 local mt = {
-    __index = methods,
+    __index = function(_, index)
```
Totktonada, line 61: Can we avoid a function here? It breaks tab-completion.

If we can't, can we also define a tab-completion hook (`__autocomplete`)?

georgiy-belyanin: Well, I've changed the code a bit:

```
--- a/src/box/lua/config/init.lua
+++ b/src/box/lua/config/init.lua
@@ -1,4 +1,5 @@
 local fio = require('fio')
+local fun = require('fun')
 local instance_config = require('internal.config.instance_config')
 local cluster_config = require('internal.config.cluster_config')
 local configdata = require('internal.config.configdata')
@@ -37,6 +38,7 @@ local function load_extras()
 end
 
 local extras = load_extras()
+local extra_methods = extras ~= nil and extras.export or {}
 
 -- {{{ Helpers
 
@@ -58,16 +60,11 @@ end
 
 local methods = {}
 local mt = {
-    __index = function(_, index)
-        if extras ~= nil and extras.export ~= nil and
-           extras.export[index] ~= nil then
-            return extras.export[index]
-        end
-
-        if methods[index] ~= nil then
-            return methods[index]
-        end
+    __index = setmetatable(methods, {__index = extra_methods}),
+    __autocomplete = function(self)
+        return fun.tomap(fun.chain(fun.iter(methods), fun.iter(extra_methods)))
     end,
+
     __serialize = filter_out_private_fields,
 }
```

I've tried to "merge" extra_methods and methods to use them as a table though it was problematic due to complicated loading process thus I've overridden the `__autocomplete`

What do you think now?

Totktonada: Can't we just do something like this in the EE code?

```lua
local function initialize(config_module)
    <...>
    local mt = getmetatable(config_module)
    local methods = mt.__index
    methods.foo = foo
end
```

Or add `:register_method()` into CE code and add the following into EE?

```lua
local function initialize(config_module)
    <...>
    config_module:register_method('foo', foo)
end
```

It seems that we can write something simple here, because the problem doesn't look complicated.

georgiy-belyanin: I don't like that approach since it requires config to be provided to access any methods added in EE (since `:initialize()` should be called)

Technically, I can add extra hook `pre_init()` that is called independently without config initialization if you insist. Though, I don't see any explicit pros comparing to the existing in it.

Totktonada: I see now, the `extras.initialize(config_module)` is called only in the YAML config startup flow. OK.

I still think that your way to add a couple of methods into a metatable involves too much programming :)

Can't we have a module in EE that does something like the following?

```lua
-- my_ee_module.lua

local mt = getmetatable(config)
local methods = mt.__index

function methods.foo(self, <...>)
    <...>
end
```

No metatable of a metatable field, no table merging, no `__autocomplete` metamethod.

The only thing we need to do to make it work is to load `my_ee_module` after the `config` module. The order of loading is determined by the `lua_sources` array in `src/box/lua/init.c`. There is no problem to achieve it.

src/box/CMakeLists.txt
georgiy-belyanin: After discussing, I've changed this PR purpose to migrate extras to EE.
Thus, the comment is applied.
Please, take a look now.

Totktonada (approved):
Thanks!

PR #11006, opened 2024-12-27
lib/bit: introduce bit_copy_range()

mkostoevr:
This function copies a range of bits between non-overlapping memory chunks. Fixed invalid address access (next byte availability check) and scanning NULLs instead of values in some cases by removing masks previously used to read source bits.

Needed for tarantool/tarantool-ee#1063

NO_DOC=internal
NO_CHANGELOG=internal


src/lib/bit/bit.h
```diff
@@ -303,6 +303,220 @@ bit_set_range(void *data, size_t pos, size_t count, bool val)
...
+ * @brief Copy \a count bits from memory chunk \a src starting from bit \a src_i
+ *  into \a dst starting from bit \a dst_i.
+ * @param dst - the memory chunk to copy bits into.
+ * @param dst_i - the position to copy bits into.
+ * @param count - the amount of bits to copy.
+ * @param src - the memory chunk to copy bits from.
+ * @param src_i - the position to copy bits from.
+ */
+static inline void
+bit_copy_range(uint8_t *restrict dst, size_t dst_i, size_t count,
```
Gumix, line 316: What if `src` and `dst` point to overlapping memory, but source and destination chunks are not overlapping because of `src_i` and `dst_i`? It sounds legal. I mean:
```
[ ... src ... dst ... src[src_i] ... src[src_i+count] ... dst[dst_i] ... dst[dst_i+count] ... ]
```
In such case the `restrict` keywords could cause UB, so we should either forbid this or remove `restrict`.

mkostoevr: Yes, this function is to be used on scanning (copying data from leaf mask into a buffer) and batched insertion (copying data from the batch validity to leaf masks).

It can't be used for rebalancing (moving bits within a single leaf mask) even if we allow that, so I suggest to restrict that for the sake of performance.

Added an assertion for that:

```diff
--- a/src/lib/bit/bit.h
+++ b/src/lib/bit/bit.h
@@ -316,6 +316,8 @@ static inline void
 bit_copy_range(uint8_t *restrict dst, size_t dst_i, size_t count,
               const uint8_t *restrict src, size_t src_i)
 {
+       assert(dst != src);
+
        if (count == 0)
                return;
 
```

Gumix: > I suggest to restrict that for the sake of performance.

Please add a comment to the function description.

> `assert(dst != src);`

A more precise check would look like:
```c
assert(src > dst + DIV_UP(dst_i + count, CHAR_BIT) || dst > src + DIV_UP(src_i + count, CHAR_BIT));
```

mkostoevr: <details>

<summary>Done.</summary>

```diff
--- a/src/lib/bit/bit.h
+++ b/src/lib/bit/bit.h
@@ -311,12 +311,15 @@ bit_set_range(void *data, size_t pos, size_t count, bool val)
  * @param count - the amount of bits to copy.
  * @param src - the memory chunk to copy bits from.
  * @param src_i - the position to copy bits from.
+ * @pre the bit buffers do not overlap.
  */
 static inline void
 bit_copy_range(uint8_t *restrict dst, size_t dst_i, size_t count,
               const uint8_t *restrict src, size_t src_i)
 {
-       assert(dst != src);
+       /* Assure the buffers are non-overlapping. */
+       assert(src > dst + DIV_ROUND_UP(dst_i + count, CHAR_BIT) ||
+              dst > src + DIV_ROUND_UP(src_i + count, CHAR_BIT));
 
        if (count == 0)
                return;
```

</details>

src/lib/bit/bit.h
```diff
@@ -303,6 +303,225 @@ bit_set_range(void *data, size_t pos, size_t count, bool val)
...
+	/*
+	 *    dst_i_bit
+	 *        |
+	 * Dst: - D D D - - - -
+	 *        \___/
+	 *          |
+	 *    dst_head_mask (but in most cases the head is until end of byte)
+	 */
+	size_t dst_head_mask = ((1ULL << dst_head_size) - 1) << dst_i_bit;
+	size_t dst_tail_mask = (1ULL << dst_tail_size) - 1;
```
nshy, line 357: Nit. Shift size here is below 8, so ULL is overkill.

src/lib/bit/bit.h
```diff
@@ -303,6 +303,225 @@ bit_set_range(void *data, size_t pos, size_t count, bool val)
 	 */
 }
```
nshy, line 304: Nit. I guess we don't need to mention bugfix here, only in EE.

nshy (approved):

src/lib/bit/bit.h
```diff
@@ -303,6 +303,225 @@ bit_set_range(void *data, size_t pos, size_t count, bool val)
...
+ * @param dst - the memory chunk to copy bits into.
+ * @param dst_i - the position to copy bits into.
+ * @param count - the amount of bits to copy.
+ * @param src - the memory chunk to copy bits from.
+ * @param src_i - the position to copy bits from.
+ * @pre the bit buffers do not overlap.
+ */
+static inline void
+bit_copy_range(uint8_t *restrict dst, size_t dst_i, size_t count,
+	       const uint8_t *restrict src, size_t src_i)
```
nshy, line 318: I'd also make `count` the last argument, for symmetry.

Gumix (approved):

src/lib/bit/bit.h
mkostoevr: Done.

PR #11007, opened 2024-12-27
memtx: finalize prepare of all transactions in memtx MVCC

### LINKED ISSUE #10715: Assertion `victim->psn == 0' failed in memtx_tx_handle_conflict()
**Bug description**

* OS: Linux
* OS Version: Ubuntu 24.04
* Architecture: amd64

Tarantool 3.3.0-entrypoint-161-gbf09135880
Target: Linux-x86_64-Debug
Compiler: Clang-18.1.3
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -fsanitize=address -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ubuntu/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror -g -ggdb -O0 
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -fno-common -msse2 -fsanitize=address -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/home/ubuntu/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -Werror -g -ggdb -O0 

**Steps to reproduce**

`./build/src/tarantool ./test/fuzz/lua/test_engine.lua --workers 1800 --test_duration 60 --engine vinyl --verbose --seed 17326`

**Actual behavior**

```
<snipped>

2024-10-18 06:57:18.913 [582307] main/1183/WRK #1068/..test.fuzz.lua.test_engine I> ERROR: opname "INDEX_UPDATE_OP", err "failed to read from file: Input/output error", args [["431abd63-9771-432f-b1c3-579f5d5de22c",463690488686,971084909917.45],{"unique":true,"parts":[{"fieldno":3,"sort_order":"asc","type":"uuid","exclude_null":false,"is_nullable":false},{"fieldno":2,"sort_order":"asc","type":"unsigned","exclude_null":false,"is_nullable":false},{"fieldno":1,"sort_order":"asc","type":"double","exclude_null":false,"is_nullable":false}],"options":{"page_size":2048,"run_count_per_level":3,"run_size_ratio":5,"bloom_fpr":0.5,"range_size":1073741824},"id":0,"space_id":512,"type":"TREE","name":"idx_1"},[["-",1,714444347063.12],["-",2,581701618565],["=",3,"8d1f420e-d91c-4111-9acb-877f955d048e"]]]
2024-10-18 06:57:18.913 [582307] main/1183/WRK #1068/..test.fuzz.lua.test_engine I> ERROR: opname "INDEX_CREATE_OP", err "Storage engine 'memtx' does not support cross-engine transactions", args []
2024-10-18 06:57:18.914 [582307] main/1112/WRK #997/..test.fuzz.lua.test_engine I> ERROR: opname "INDEX_GET_OP", err "failed to read from file: Input/output error", args [{"unique":true,"parts":[{"fieldno":3,"sort_order":"asc","type":"uuid","exclude_null":false,"is_nullable":false},{"fieldno":2,"sort_order":"asc","type":"unsigned","exclude_null":false,"is_nullable":false},{"fieldno":1,"sort_order":"asc","type":"double","exclude_null":false,"is_nullable":false}],"options":{"page_size":2048,"run_count_per_level":3,"run_size_ratio":5,"bloom_fpr":0.5,"range_size":1073741824},"id":0,"space_id":512,"type":"TREE","name":"idx_1"},["893fe638-e1d9-442e-8256-2fad2c59ca4d",848212856838,55347079295.925]]
2024-10-18 06:57:18.915 [582307] main/1881/WRK #1766/..test.fuzz.lua.test_engine I> ERROR: opname "UPSERT_OP", err "failed to read from file: Input/output error", args [[313939472593.49,981808514589,"a4b5c1fc-1165-4650-b1db-7b40ee8d9232",226319329138,"bvbarvrjnc",322030354274,false,"683264593497","2024-10-18T06:57:18Z",[1,2,3,4,5,6,7,8,9,10,11,12,13,14],{"4":4,"8":8,"1":1,"5":5,"9":9,"6":6,"3":3,"7":7,"2":2}],[["-",1,468861014578.57]]]
tarantool: src/box/memtx_tx.c:839: void memtx_tx_handle_conflict(struct txn *, struct txn *): Assertion `victim->psn == 0' failed.
```

log and tarantool binary [gh-10715.zip](https://github.com/user-attachments/files/17429725/gh-10715.zip)

**Expected behavior**

no assertion violation

drewdzzz:
We have a special `sysview` engine that bypasses transaction management and reads right from memtx indexes. The problem is memtx indexes use memtx MVCC by themselves, so despite `sysview` bypasses transaction management process, it actually uses memtx MVCC. So the problem here is that transaction can actually read something using `sysview`, a tracker in MVCC will be allocated, but `txn->engine[memtx->id]` still will be `NULL`, so the preparation of the transaction won't be finalized by memtx_tx and the tracker won't be deleted when transaction is prepared. On conflict, MVCC will try to abort that already prepared transaction and it will lead to a crash.

Actually, crashing scenarios are rather synthetic and break only our engine fuzzer. Here is an example of such scenario without manual use of system spaces:
1. Open a transaction and do an operation in a vinyl space.
2. Try to alter the space in the same transaction - it will fail with cross-engine transaction error.
3. Commit the transaction, but do not wait for WAL write (internally, such state is called `prepared`).
4. Alter the space in another transaction and commit it before the previous transaction was written to WAL. Here MVCC will try to conflict that already prepared transaction and it will break Tarantool.

To solve the problem, we could try to write quite a large patch to handle such transactions differently in memtx MVCC. But instead, let's simply finalize prepare of all transaction in MVCC. The current design already lacks encapsulation - all transactions (even non-memtx) are registered there, so such solution not only simple but also consistent with other MVCC parts. Note that memtx MVCC still can abort such non-memtx transactions when they are in-progress, but it seems that such behavior doesn't break other yielding engines (`vinyl` is the only one).

In order not to do unnecessary function call when MVCC is disabled, function `memtx_tx_prepare_finalize` is `static inline` now and uses a `*_slow` helper when MVCC is actually enabled - such approach is common in this subsystem.

Closes #10715

### DIFF
--- a/changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
+++ b/changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
@@ -1,0 +1,5 @@
+## bugfix/memtx
+
+* Fixed a crash when memtx MVCC tried to abort an already committed
+  non-memtx transaction if it used a system space view or tried to
+  perform a DDL operation (gh-10715).
--- a/src/box/txn.c
+++ b/src/box/txn.c
@@ -1069,6 +1069,12 @@ txn_prepare(struct txn *txn)
 			return -1;
 		}
 	}
+	/*
+	 * Non-memtx transactions can use memtx MVCC (for example, transactions
+	 * reading from sysview), so we need to finalize prepare of all
+	 * transactions, not only memtx ones.
+	 */
+	memtx_tx_prepare_finalize(txn);
 
 	trigger_clear(&txn->fiber_on_stop);
 	trigger_clear(&txn->fiber_on_yield);
--- a/test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
+++ b/test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
@@ -1,0 +1,78 @@
+local server = require('luatest.server')
+local t = require('luatest')
+
+-- The original issue is about non-memtx transactions, but let's
+-- test memtx along the way.
+local g = t.group(nil, {{engine = 'memtx'}, {engine = 'vinyl'}})
+
+g.before_all(function(g)
+    g.server = server:new{
+        box_cfg = {memtx_use_mvcc_engine = true}
+    }
+    g.server:start()
+end)
+
+g.after_all(function(g)
+    g.server:drop()
+end)
+
+g.before_each(function(g)
+    g.server:exec(function(engine)
+        local s = box.schema.space.create('s', {engine = engine})
+        s:create_index('pk')
+    end, {g.params.engine})
+end)
+
+g.after_each(function(g)
+    g.server:exec(function()
+        box.space.s:drop()
+    end)
+end)
+
+g.test_conflict_with_prepared_sysview_transaction = function(g)
+    g.server:exec(function()
+        local fiber = require('fiber')
+
+        local ch = fiber.channel(1)
+        local f = fiber.create(function()
+            box.begin()
+            -- Read from sysview over _space to conflict this
+            -- transaction with future DDL.
+            box.space._vspace:select{}
+            box.space.s:replace{1}
+            ch:put(true)
+            box.commit()
+        end)
+        f:set_joinable(true)
+
+        -- The crash happened because we conflicted a prepared non-memtx
+        -- transaction. Here we check that we don't conflict such
+        -- transactions anymore.
+        t.assert_equals({ch:get()}, {true})
+        box.space.s:format({{'f1', 'unsigned'}})
+        t.assert_equals({f:join()}, {true})
+    end)
+end
+
+-- Along the way, make sure that conflict of in-progress transaction
+-- caused by sysview read works well.
+g.test_conflict_with_in_progress_sysview_transaction = function(g)
+    g.server:exec(function()
+        local fiber = require('fiber')
+
+        box.begin()
+        -- Read from sysview over _space to conflict this
+        -- transaction with future DDL.
+        box.space._vspace:select{}
+        box.space.s:replace{1}
+
+        local f = fiber.create(function()
+            box.space.s:format({{'f1', 'unsigned'}})
+        end)
+        f:set_joinable(true)
+        t.assert_equals({f:join()}, {true})
+
+        t.assert_error_msg_equals(
+            'Transaction has been aborted by conflict', box.commit)
+    end)
+end
--- a/src/box/memtx_engine.cc
+++ b/src/box/memtx_engine.cc
@@ -626,7 +626,6 @@ memtx_engine_prepare(struct engine *engine, struct txn *txn)
 			assert(stmt->space->engine == engine);
 			memtx_tx_history_prepare_stmt(stmt);
 		}
-		memtx_tx_prepare_finalize(txn);
 	}
 	return 0;
 }
--- a/src/box/memtx_tx.c
+++ b/src/box/memtx_tx.c
@@ -2687,7 +2687,7 @@ memtx_tx_history_prepare_stmt(struct txn_stmt *stmt)
 }
 
 void
-memtx_tx_prepare_finalize(struct txn *txn)
+memtx_tx_prepare_finalize_slow(struct txn *txn)
 {
 	/* Just free all other lists - we don't need 'em anymore. */
 	memtx_tx_clear_txn_read_lists(txn);
--- a/src/box/memtx_tx.h
+++ b/src/box/memtx_tx.h
@@ -225,14 +225,24 @@ void
 memtx_tx_history_prepare_stmt(struct txn_stmt *stmt);
 
 /**
+ * Helper of `memtx_tx_prepare_finalize`.
+ */
+void
+memtx_tx_prepare_finalize_slow(struct txn *txn);
+
+/**
  * Finish preparing of a transaction.
  * Must be called for entire transaction after `memtx_tx_history_rollback_stmt`
  * was called for each transaction statement.
  *
  * NB: can trigger story garbage collection.
  */
-void
-memtx_tx_prepare_finalize(struct txn *txn);
+static inline void
+memtx_tx_prepare_finalize(struct txn *txn)
+{
+	if (memtx_tx_manager_use_mvcc_engine)
+		memtx_tx_prepare_finalize_slow(txn);
+}
 
 /**
  * @brief Commit statement in history.


changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
```diff
@@ -0,0 +1,5 @@
+## bugfix/memtx
```
CuriousGeorgiy, line 1: > To solve the problem, we could try to write quite a large patch to handle
such transactions differently in memtx MVCC. But instead, let's simply
finalize prepare of all transaction in MVCC. The current design already
lacks encapsulation - all transactions (even non-memtx) are registered
there, so such solution not only simple but also consistent with other
MVCC parts. Note that memtx MVCC still can abort such non-memtx
transactions when they are in-progress, but it seems that such behavior
doesn't break other yielding engines (`vinyl` is the only one).

I really dislike the motivation for this solution, it's totally a crudge. What if tomorrow the sysview engine (or one of the other transaction management bypassing engines) starts using other engines' metadata?

What's wrong with the following solutions?
1. Add a new helper around the `memtx_use_mvcc_engine` flag that will also check for the  `ENGINE_BYPASS_TX` flag set on a transaction (or check that transaction actually uses the memtx engine).
2. Rework the original `ENGINE_BYPASS_TX` flag solution: AFAIC, the engines that use it are actually read-only, so mixing them with other engines shouldn't break anything. We could introduce some `ENGINE_READ_ONLY` flag and avoid raising `ER_CROSS_ENGINE_TRANSACTION` if the transaction uses only read-only engines.

test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
```diff
@@ -0,0 +1,54 @@
...
+            box.commit()
+        end)
+        f:set_joinable(true)
+
+        -- The crash happened because we conflicted a prepared non-memtx
+        -- transaction. Here we check that we don't conflict such
+        -- transactions anymore.
+        t.assert_equals({ch:get()}, {true})
+        box.space.s:format({{'f1', 'unsigned'}})
+        t.assert_equals({f:join()}, {true})
```
CuriousGeorgiy, line 52: Nit: AFAIC, the curly braces around the values in `assert_equals` here and above are redundant.

test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
```diff
@@ -0,0 +1,54 @@
...
+        local ch = fiber.channel(1)
+        local f = fiber.create(function()
+            box.begin()
+            -- Read from sysview over _space to conflict this
+            -- non-memtx transaction with future DDL.
+            box.space._vspace:select{}
+            box.space.s:replace{1}
+            ch:put(true)
+            box.commit()
+        end)
```
CuriousGeorgiy, line 44: Nit: AFAIC, since you are using `fiber.create`, you will get control back only after the transaction is prepared and yields for the WAL write, so this `fiber.channel` is redundant.

test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
```diff
@@ -0,0 +1,54 @@
...
+end)
+
+g.after_all(function()
+    g.server:drop()
+end)
+
+g.before_each(function()
+    g.server:exec(function()
+        -- Create vinyl space to test non-memtx transactions
+        local s = box.schema.space.create('s', {engine = 'vinyl'})
```
CuriousGeorgiy, line 20: Please create a test matrix to also test memtx.

CuriousGeorgiy (changes requested):

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
drewdzzz: > Add a new helper around the `memtx_use_mvcc_engine` flag that will also check for the `ENGINE_BYPASS_TX` flag set on a transaction (or check that transaction actually uses the memtx engine).

Do you want to mark all transactions bypassed transaction management to finalize their preparation later? It seems more natural to me to simply finalize preparation of all of them - we register every transaction in memtx MVCC, let's finalize every one as well.
Actually, we should register only memtx transactions in memtx MVCC. I tried to do so in https://github.com/tarantool/tarantool/pull/10787, but we decided that we won't do that. Also, probably it would lead to some problems if we didn't collect conflicts for `sysview` engine in memtx MVCC.

> Rework the original `ENGINE_BYPASS_TX` flag solution: AFAIC, the engines that use it are actually read-only, so mixing them with other engines shouldn't break anything.

Engine `service` (space `_session_settings`) supports update, so actually it is not read-only.

> What if tomorrow the sysview engine starts using other engines' metadata?

Sounds as a bad idea. For example, `sysview` over `vinyl` would yield and implicitly abort `memtx` transactions. I think there are enough pitfalls, and my solution doesn't add new ones.

I had another idea - we could handle `sysview` differently. It wouldn't bypass transaction management, but it would be registered as an underlying engine (only `memtx` for now). But the patch would be more difficult and it transaction management subsystem would be more complicated.

test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
drewdzzz: Since vinyl is allowed to yield on replace, I added the channel to make sure that the transaction is actually prepared.

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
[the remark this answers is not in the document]
drewdzzz: In another comment.

test/box-luatest/gh_10715_memtx_mvcc_conflict_with_prepared_sysview_txn_test.lua
drewdzzz: Done.

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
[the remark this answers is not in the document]
drewdzzz: Done.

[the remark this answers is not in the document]
drewdzzz: Done.

locker: Can we make the sysview engine bypass memtx MVCC completely, i.e. query directly from memtx system spaces without tracking reads? I don't think we need to track reads because DDL operations (most of them) don't yield and read prepared (unconfirmed) statements.

src/box/txn.c
```diff
@@ -1069,6 +1069,11 @@ txn_prepare(struct txn *txn)
 			return -1;
 		}
 	}
+	/*
+	 * Since every transaction is registered in memtx_tx (see `txn_begin`),
+	 * every transaction should finalize its prepare there as well.
+	 */
+	memtx_tx_prepare_finalize(txn);
```
locker, line 1076: The new comment looks rather confusing because `memtx_tx_prepare_finalize` has nothing to do with `memtx_tx_register_txn` AFAICS.

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
CuriousGeorgiy: > Do you want to mark all transactions bypassed transaction management to finalize their preparation later? It seems more natural to me to simply finalize preparation of all of them - we register every transaction in memtx MVCC, let's finalize every one as well.

No, I just want all MVCC related API to check for the `ENGINE_BYPASS_TX` flag and exit early without doing anything, just like we do with `memtx_use_mvcc_engine`.

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
```diff
@@ -0,0 +1,5 @@
...
+
+* Fixed a crash when memtx MVCC tried to abort an already committed
+  non-memtx transaction if it used a system space view or tried to do
+  a DDL operation (gh-10715).
```
lenkis, line 5:
```suggestion
  non-memtx transaction if it used a system space view or tried to
  perform a DDL operation (gh-10715).
```

lenkis (approved):

changelogs/unreleased/gh-10715-memtx-mvcc-conflict-with-prepared-sysview-txn.md
drewdzzz: > No, I just want all MVCC related API to check for the `ENGINE_BYPASS_TX` flag

MVCC helpers are called under the hood by memtx tree methods and it has access only to the active transaction. So the question is - how can you check the `ENGINE_BYPASS_TX` flag if it belongs to an engine, not transaction? Read statements don't even produce `txn_stmt`, so we don't know which engine is reading data right now.

To exit early, we could somehow mark transactions that actually were "begun" in memtx. But in this case, sysview would bypass memtx MVCC until the first memtx statement and wouldn't bypass it after - I don't like this inconsistency.

Moreover, ignoring non-memtx transactions in memtx MVCC would produce quite large amount of code and complicate it even more. I don't like it as well.

So I propose this solution. It's really simple - we initialize all transactions in memtx MVCC, so let's uninitialize them all as well.

CuriousGeorgiy (approved):

src/box/txn.c
drewdzzz: Updated the comment:
```
/*
  * Non-memtx transactions can use memtx MVCC (for example, transactions
  * reading from sysview), so we need to finalize prepare of all
  * transactions, not only memtx ones.
  */
```

locker (approved):

Issue #10330, opened 2024-07-30
connpool: make `filter()` and `call()` execute faster


georgiy-belyanin:
Now `connpool.filter()` and `connpool.call()` try to establish the connection if the instance is not available. This leads to quite a big timeout wait in case when some of the instances are not available at all.

Another existing problem is that there may be too many unused connections established. Connection pool should be able to close the unused connections.

The suggested solution of both of these problems relies on #10352 on the connection time-to-live parameter. 

Connections established by the `connpool.connect` method should have a time to live automatically bumped if the connection is used. If the connection breaks the connection pool automatically marks the instance as unavailable and ignores it when evaluating `filter()` or `call()` speeding up their execution by avoiding attempts to connect once again.

`connect()` is also used internally by the `filter()` and `call()` functions and such implementation will make them use the same cache for free.

The deadline should be a configurable option in the new `connpool` configuration section.
```yaml
connpool:
  no_activity_timeout: 300
```

Also, closing unused connections solves some problems with the too many opened connections related issues occurring because of the file descriptor/memory limitations.

PR #11013, opened 2024-12-30
connpool: add internal pool class

georgiy-belyanin:
connpool: add internal pool class

This patch introduces an internal pool abstraction in
`experimental.connpool` module. The idea is quite similar to a cartridge
pool module [1] used in RPC [2]. We need to have a way to work with a
set of connections. But instead of URIs we use instance names now.

This abstraction is useful for #10330. The issue is about making the
connpool check the connections in the background. It includes closing
unused connections automatically to free a few file descriptors. We want
to make `filter()` and `call()` methods open connections that'd be
closed if they remain unused. These functions use a public `connect()`
method internally. Thereby straightforward implementation requires
extending public `connect()` method with an option making the connection
non-persistent.

This patch moves the basic `connect()` logic into an
internal pool function allowing to keep the "non-persistent"
completely internal.

This patch only involves the most straightforward implementation
of the pool as an key-value connection map with a `connect()` method
which is likely to be extended in the future. The `connect()` method
behaves the same way as it's been implemented in a public API except
instead of throwing an exception in case of an error it returns nil as
the first return value and an error message as the second return value.

Also, the pool introduces a `connect_to_multiple()` call providing the
same behavior as `connect_to_candidates()` i.e. connect to all of the
candidates and wait until the timeout or until all of them is reached,
an `any` option can be supplied to make the method wait for at least one
connection satisfying the constraints.

Needed for #10330

[^1] https://github.com/tarantool/cartridge/blob/master/cartridge/pool.lua
[^2] https://github.com/tarantool/cartridge/blob/master/cartridge/rpc.lua

### DIFF
--- a/src/box/lua/connpool.lua
+++ b/src/box/lua/connpool.lua
@@ -8,7 +8,10 @@ local netbox = require('net.box')
 local WATCHER_DELAY = 0.1
 local WATCHER_TIMEOUT = 10
 
-local connections = {}
+-- {{{ Basic instance connection pool
+
+local pool_methods = {}
+local pool_mt = {__index = pool_methods}
 
 local function is_connection_valid(conn, opts)
     if conn == nil or conn.state == 'error' or conn.state == 'closed' then
@@ -22,25 +25,25 @@ local function is_connection_valid(conn, opts)
     return true
 end
 
-local connection_mode_update_cond = nil
-local function connect(instance_name, opts)
-    if not connection_mode_update_cond then
-        connection_mode_update_cond = fiber.cond()
-    end
-
-    checks('string', {
+--- Connect to an instance or receive a cached connection by
+--- name.
+---
+--- The method returns a connection as the first return value
+--- and an error message as the second one.
+function pool_methods.connect(self, instance_name, opts)
+    checks('table', 'string', {
         connect_timeout = '?number',
         wait_connected = '?boolean',
         fetch_schema = '?boolean',
     })
     opts = opts or {}
 
-    local conn = connections[instance_name]
+    local conn = self._connections[instance_name]
     if not is_connection_valid(conn, opts) then
         local uri = config:instance_uri('peer', {instance = instance_name})
         if uri == nil then
             local err = 'No suitable URI provided for instance %q'
-            error(err:format(instance_name), 0)
+            return nil, err:format(instance_name)
         end
 
         local conn_opts = {
@@ -51,10 +54,10 @@ local function connect(instance_name, opts)
         local ok, res = pcall(netbox.connect, uri, conn_opts)
         if not ok then
             local msg = 'Unable to connect to instance %q: %s'
-            error(msg:format(instance_name, res.message), 0)
+            return nil, msg:format(instance_name, res.message)
         end
         conn = res
-        connections[instance_name] = conn
+        self._connections[instance_name] = conn
         local function mode(conn)
             if conn.state == 'active' then
                 return conn._mode
@@ -64,7 +67,7 @@ local function connect(instance_name, opts)
         conn.mode = mode
         local function watch_status(_key, value)
             conn._mode = value.is_ro and 'ro' or 'rw'
-            connection_mode_update_cond:broadcast()
+            self._connection_mode_update_cond:broadcast()
         end
         conn:watch('box.status', watch_status)
     end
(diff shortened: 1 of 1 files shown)


src/box/lua/connpool.lua
Totktonada, line 1: A private API is still an API and it should be convenient to use. I propose to look on it from this point of view.

We had the following:

* `connect` -- has all the options
* `connect_in_background` -- no wait, no exceptions, discard result, no options

Now we have the following:

* `connect_to_instance` -- has all the options except waiting
* `connect` -- as before
* `connect_in_background` -- as before

First of all, `connect_nowait`/`connect_async` or something like this would fit better for the new function.

Next, it is not clear, why `_nowait` is needed if there is `wait_connected` option. As a shortcut?

Next, if we have three connecting functions, we possibly should have the same options for them (otherwise it is easy to mix them up). Or rename `connect_in_background` to something like `mark2connect`.

Next, I don't understand how extracting a part of `connect` into a helper function allows to 'extend the internal logic with new handles and not to touch the public method at the same time'. If we change the extracted part -- we change `connect`. If we touch the left part -- we change `connect`. It seems, I need more details here to understand the idea behind.

georgiy-belyanin: You're right. The changes weren't clear at the moment.
After discussing the patch in the PM, please take a look at the new implementation of the connpool involving a new pool abstraction.

The motivation is in the commit message.

georgiy-belyanin:
I've did a few force-pushes since I've found a few issues with the previous variants.

src/box/lua/connpool.lua
```diff
@@ -8,7 +8,17 @@ local netbox = require('net.box')
...
+
+local pool_methods = {}
+local pool_mt = {
+    __index = function(self, key)
+        if self._connections[key] ~= nil then
+            return self._connections[key]
+        end
+        return pool_methods[key]
+    end
+}
```
Totktonada, line 21: Accessing the connection using `pool[instance_name]` breaks the pool's abstraction, in my opinion. Moreover, our code only calls `:connect()` to obtain a connection. All the other accesses are to obtain some properties: whether the connection is ready (`is_candidate_connected`), whether it is in a terminal state (`is_candidate_connected`), get the database mode (ro/rw). I propose to access these properties using methods, without direct access to the connection object.

(Also I found connected/checked terminology confusing and I don't mind if you change it.)

There are two other reasons why I don't like this `pool[instance_name]` shortcut (but these are more minor). First, it makes impossible to have an instance named `connect` or the same as any other pool's method. Second, it breaks tab-completion in console (that is good for instrospection/debugging).

Totktonada (commented):
Sorry for the long delay in review. I've a few comments here. Please, take a look.

src/box/lua/connpool.lua
```diff
@@ -78,10 +83,35 @@ local function connect(instance_name, opts)
     return conn
 end
 
+function pool_methods._wait_connection_status_update(self, timeout)
+    self._connection_mode_update_cond:wait(timeout)
+end
```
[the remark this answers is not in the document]
georgiy-belyanin: Applied. Thank you.

src/box/lua/connpool.lua
georgiy-belyanin: Thank you for this note. After seeing it I've decided to change the approach on refactoring a bit.

* I've changed the `pool.connect()` method in that way it never throws an error. Instead it returns `nil, errmsg` if something bad occurs. It reminds previous `connect_in_background()` except it also accept a `wait_connected` parameter.
* I've started to reuse the `connect()` method within all `connpool` methods and removed the `pool[instance_name]` index method.
* `connect_to_candidates()` call is now a part of the pool (and is now called `connect_to_multiple`), since the method itself is quite low-level and uses the conditional variable. By the way it makes exporting the conditional variable unnecessary.
```lua
--- This method connects to the specified instances and returns
--- the set of successfully connected ones.
---
--- Specifying a callback accepting an instance name and returning
--- a boolean value in `opts.any` changes the behavior to connect
--- to multiple instances until there is at least one satisfying
--- the condition or the timeout occurs.
function pool_methods.connect_to_multiple(self, instances, opts)
    checks('table', 'table', {
        any = '?function',
        timeout = '?number',
    })
    -- ...
end
```
* This changes make it possible to drastically simplify the `is_candidate_match_dynamic()`. It no longer needs `connect_in_background` call since evaluating `:connect()` on the desired instance schedules connecting to it automatically.
```diff
@@ -234,28 +255,19 @@ local function is_mode_match(mode, instance_name)
     if mode == nil then
         return true
     end
-    local conn = connections[instance_name]
-    assert(conn ~= nil)
-    return conn:mode() == mode
+    local conn = connections:connect(instance_name, {wait_connected = false})
+    assert(conn ~= nil and conn:mode() ~= nil)
+    return conn:mode() == mode or mode == nil
 end
 
 local function is_candidate_match_dynamic(instance_name, opts)
     assert(opts ~= nil and type(opts) == 'table')
 
-    local conn = connections[instance_name]
-
-    -- No connection -> reconnect in background, exclude the candidate.
-    if conn == nil or conn.state == 'error' or conn.state == 'closed' then
-        connect_in_background(instance_name)
-        return false
-    end
-
-    -- In progress -> exclude the candidate.
-    if not (conn.state == 'active' and conn:mode() ~= nil) then
-        return false
+    local conn = connections:connect(instance_name, {wait_connected = false})
+    if not conn or conn:mode() == nil then
+        return
     end
 
-    -- Connected -> check using the provided filters.
     return is_mode_match(opts.mode, instance_name)
 end
```
* The instantiated pool is now called `pool` instead of `connections`.

Please, take a look now and leave your opinion on the suggested changes.

src/box/lua/connpool.lua
```diff
@@ -73,77 +76,95 @@ local function connect(instance_name, opts)
...
-local function connect_to_candidates(candidates, opts)
-    if opts == nil then opts = {} end
-    if next(candidates) == nil then return {} end
+    -- Checks whether the candidate has responded with success or
+    -- with an error.
+    local function is_instance_checked(instance_name)
+        local conn = self._connections[instance_name]
+        return is_instance_connected(instance_name) or
+               conn.state == 'error' or conn.state == 'closed'
+    end
```
Totktonada, line 113: I'm a bit worrying about lack of the `conn == nil` condition here.

I understand that we can't have `conn == nil` after `self:connect()` even after a yield in the current implementation.

However, how about updating the configuration that we want to implement later? Or autoclosing of unused connections?

I think that the functions performs a very small job and we shouldn't think about outside code (whether the `nil` is possible) while looking on it. I would handle this case too.

src/box/lua/connpool.lua
Totktonada: I like your ideas. Thanks!

I think that `pool:mode(instance_name)` that doesn't return a connection (and doesn't start connection establishing in background) can be useful for filters, but I see that it requires some more work. Let's skip this idea for now.

Totktonada (approved):

Issue #10766, opened 2024-10-30
Non-blocking index build blocks synchro queue


drewdzzz:
In a replica's applier, we simply [do](https://github.com/tarantool/tarantool/blob/40fa44c2b1205642cb36cdee2bd85169606e0c0c/src/box/applier.cc#L313) `box_process_rw` for each request and then asynchronously commit them. The `box_process_rw` returns when a replace is done and all its `on_replace` triggers are fired.

If the request is an index build, applier will actually block on it - yes, it will yield to share CPU with other fibers, but the applier will wait until the index is completely built before applying further requests since the build happens in a system `on_replace` trigger and `box_process_rw` waits for it. Since it happens for all replicas, the master of synchronous cluster has to wait for a quorum to build the index. The problem leads to:
1. A commit latency spike - master cannot commit new requests while a quorum hasn't build the index;
2. Build of huge index and further requests can fail due to quorum timeout;
3. Synchro queue overflow can happen.

I wrote a [reproducer](https://gist.github.com/drewdzzz/018de967f40b90c7844ebe131875956a) with `luatest` (for easier cluster management and launch). The first test, imitating a huge index build with error injection, fails with quorum timeout error. The second one actually builds a huge index - it crashes a slave replica! I don't know if the crash is related to the index build only or I just triggered another problem. Crash backtrace:
```c++
replica | Segmentation fault
replica |   code: SEGV_ACCERR
replica |   addr: 0x5050505050505070
replica |   context: 0x12812fc18
replica |   siginfo: 0x12812fbb0
replica | Current time: 1730278554
replica | Please file a bug at https://github.com/tarantool/tarantool/issues
replica | Attempting backtrace... Note: since the server has already crashed,
replica | this may fail as well
replica | #1  0x102e10548 in crash_collect+200
replica | #2  0x102e103fc in crash_signal_cb+92
replica | #3  0x1994f6584 in _sigtramp+56
replica | #4  0x102cbef68 in xlog_write_entry+180
replica | #5  0x102cbe400 in wal_write_to_disk+1088
replica | #6  0x102e28b8c in cmsg_deliver+52
replica | #7  0x102e29a88 in cbus_process+192
replica | #8  0x102e29e1c in cbus_loop+28
replica | #9  0x102cbae18 in wal_writer_f+208
replica | #10 0x102ae6c88 in fiber_cxx_invoke(int (*)(char*), char*)+32
replica | #11 0x102e1c1ec in fiber_loop+264
[...]

PR #11016, opened 2025-01-07
replication: fix crash on non-empty limbo exit

Gerold103:
Normally it couldn't happen with user-originated transactions, because they were always attached to a fiber, and the shutdown code would panic if any user fibers couldn't be cancelled and joined fast enough.

However it could always happen with applier txns (async, no fiber) and now is possible with `wait='none'` or `'submit'` commit modes.

When this would happen, at least 2 situations were possible:

- The txn was not synchro and was just ignored, deleted together with the memory allocator.

- The txn was synchro and waiting for ACKs. It was "freed" (put back into the txn cache) without completion, and its triggers would leak.

- The txn was synchro and being written to WAL right now. Then it was also freed (including its region memory removal), and the WAL thread would sometimes crash on use-after-free.

The last case was a problem.

The solution is to stop freeing the txns in the limbo. They do not belong to the limbo anyway, only live there temporarily, and in case of actual completion they would have to be "completed" (txn_complete_fail() and txn_complete_success()) anyway, not just freed.

Those orphan txns need to be ignored then. They will never complete.

Another potential problem was that the limbo was created before WAL, but destroyed after WAL. Although in this particular case it didn't matter.

Note, that freeing the txns isn't possible even after the WAL thread is stopped. Because their freeing would try to unref the txn stmt tuples, which are already destroyed. Because the engine was destroyed before the limbo, and memtx allocators destruction frees all the memtx tuple memory, regardless of their refs.

Another approach would be to init the limbo after the engine and destroy it before the engine. But the txns objects still leak. Their cache is never freed. It would need a deeper rework if ever necessary.

Part of #10766

NO_DOC=bugfix

### DIFF
--- a/changelogs/unreleased/gh-10766-synchro-crash-on-exit.md
+++ b/changelogs/unreleased/gh-10766-synchro-crash-on-exit.md
@@ -1,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a crash which could happen while an instance was shut down
+  having an unfinished synchronous transaction (gh-10766).
--- a/src/box/wal.c
+++ b/src/box/wal.c
@@ -1168,7 +1168,7 @@ wal_write_to_disk(struct cmsg *msg)
 	 */
 
 	struct xlog *l = &writer->current_wal;
-
+	ERROR_INJECT_SLEEP_FOR(ERRINJ_WAL_DELAY_DURATION);
 	/*
 	 * Iterate over requests (transactions)
 	 */
@@ -1194,7 +1194,6 @@ wal_write_to_disk(struct cmsg *msg)
 		err_code = JOURNAL_ENTRY_ERR_IO;
 		goto done;
 	}
-
 	writer->checkpoint_wal_size += rc;
 	last_committed = stailq_last(&wal_msg->commit);
 	vclock_merge(&writer->vclock, &vclock_diff);
--- a/test/replication-luatest/gh_10766_shutdown_non_empty_limbo_test.lua
+++ b/test/replication-luatest/gh_10766_shutdown_non_empty_limbo_test.lua
@@ -1,0 +1,110 @@
+local t = require('luatest')
+local treegen = require('luatest.treegen')
+local justrun = require('luatest.justrun')
+local server = require('luatest.server')
+local fiber = require('fiber')
+
+local g = t.group()
+
+g.after_each(function(cg)
+    if cg.master then
+        cg.master:drop()
+    end
+end)
+
+g.test_non_empty_limbo_shutdown_wait_mode = function()
+    t.tarantool.skip_if_not_debug()
+    local script = [[
+        box.cfg{}
+        local s = box.schema.space.create('test', {is_sync = true})
+        s:create_index('pk')
+        box.ctl.promote()
+        box.cfg{replication_synchro_quorum = 2}
+        box.error.injection.set('ERRINJ_WAL_DELAY_DURATION', 0.2);
+
+        box.begin()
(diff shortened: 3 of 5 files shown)


src/box/wal.c
```diff
@@ -1194,7 +1194,12 @@ wal_write_to_disk(struct cmsg *msg)
 		err_code = JOURNAL_ENTRY_ERR_IO;
 		goto done;
 	}
-
+#if !NDEBUG
+	struct errinj *inj;
+	inj = errinj(ERRINJ_WAL_WRITE_DURATION, ERRINJ_DOUBLE);
+	if (inj->dparam > 0)
```
Serpentian, line 1200: I suppose it's better to introduce `ERROR_INJECT_SLEEP_FOR` macros in `errinj.h` in order to make it more reusable. Or we can modify the existing `ERROR_INJECT_SLEEP`

test/replication-luatest/gh_10766_shutdown_non_empty_limbo_test.lua
```diff
@@ -0,0 +1,110 @@
+local t = require('luatest')
```
Serpentian, line 1: The commit msg says:

> - The txn was synchro and being written to WAL right now. Then it
      was also freed (including its region memory removal), and
      the WAL thread would sometimes crash on use-after-free.

However, without the patch I didn't manage to reproduce the crash. In the majority of cases the test passes. The first test always passes, the second one sometimes hangs. I was testing on debug build. Am I doing smth wrong or it's reproduced very rarely? If it's the latter, are we ok with so rarely reproducible test?

Serpentian (commented):
Thank you! I'm ok with the patch itself, but I'm unsure about the test

test/replication-luatest/gh_10766_shutdown_non_empty_limbo_test.lua
Gerold103: I tried to catch it in Ubuntu (in docker) x86, Ubuntu ARM (also docker), Mac M3 - only the last one crashes reliably.

But I tried to improve it. I've moved the WAL delay a bit earlier, so after it is released, it will try to access more of the region data. On Mac both the test cases crash without the patch. Can you try again on your machine?

Serpentian: Yep, it's fails reliably now, thank you

Serpentian (approved):
Thank you for your patch! No more objections

changelogs/unreleased/gh-10766-synchro-crash-on-exit.md
```diff
@@ -0,0 +1,4 @@
+## bugfix/replication
+
+* Fixed a crash which could happen when an instance was being shut down while
+  having an unfinished synchronous transaction (gh-10766).
```
lenkis, line 4:
```suggestion
* Fixed a crash which could happen while an instance was shut down
  having an unfinished synchronous transaction (gh-10766).
```

lenkis: I just suggested more lightweight phrasing: the first "while" implies that we're talking about the shutdown *process* (so we can get rid of "was being shut down" in favor of "was shut down") + the second "while" is not much needed here.

lenkis (approved):

sergepetrenko (approved):
Thanks for the patch!
LGTM. Although, I'd like @nshy to take a look at this.

sergepetrenko:
`txn_limbo_free` was added in 6fd45447401, and was never backported to 3.2 or below. So backporting this PR to 3.3 only
Backport 3.3:

